PR #218, opened 2016-11-09
test_ipagetkeytab: use system-wide IPA CA cert location in tests

martbab:
Since /etc/ipa/ca.crt should be always present on the test runner, we should
use it in bind method tests and not rely on its presence in user conf dir.

https://fedorahosted.org/freeipa/ticket/6409


pspacek:
Jenkins tests now pass.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/3ecda74d14066f6609d72422041bcc0c6499de77

PR #222, opened 2016-11-09
Fix ipa-replica-install when upgrade from ca-less to ca-full

flo-renaud:
When ipa-replica-prepare is run on a master upgraded from CA-less to
CA-full, it creates the replica file with a copy of the local /etc/ipa/ca.crt.
This causes issues if this file hasn't been updated with ipa-certupdate,
as it contains the external CA that signed http/ldap certs, but not
the newly installed IPA CA.
As a consequence, ipa-replica-install fails with "Could not find a CA cert".

The fix consists in retrieving the CA certificates from LDAP instead of
the local /etc/ipa/ca.crt.

https://fedorahosted.org/freeipa/ticket/6375

### DIFF
--- a/ipaserver/install/ipa_replica_prepare.py
+++ b/ipaserver/install/ipa_replica_prepare.py
@@ -34,6 +34,7 @@ from six.moves.configparser import SafeConfigParser
 
 from ipaserver.install import certs, installutils, bindinstance, dsinstance
 from ipaserver.install.replication import enable_replication_version_checking
+from ipaserver.install.server.replicainstall import install_ca_cert
 from ipaserver.install.bindinstance import (
     add_zone, add_fwd_rr, add_ptr_rr, dns_container_exists)
 from ipapython import ipautil, admintool
@@ -356,6 +357,7 @@ class ReplicaPrepare(admintool.AdminTool):
             if options.setup_pkinit:
                 self.copy_pkinit_certificate()
 
+            self.retrieve_ca_certs()
             self.copy_misc_files()
 
             self.save_config()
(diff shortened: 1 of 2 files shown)


ipaserver/install/ipa_replica_prepare.py
```diff
@@ -448,12 +450,23 @@ def copy_pkinit_certificate(self):
...
+        ca_certs = certstore.get_ca_certs(api.Backend.ldap2, api.env.basedn,
+                                          api.env.realm, False)
+        certs = (c[0] for c in ca_certs if c[2] is not False)
+        dest = os.path.join(self.dir, "ca.crt")
+        try:
+            x509.write_certificate_list(certs, dest)
+        except Exception as e:
+            raise admintool.ScriptError(
+                "File creation {0} failed: {1}".format(dest, e))
+
```
stlaz: This looks a lot like `install_ca_cert()` from replicainstall.py. This function changed a bit (but not a whole lot) during installer-refactoring (see https://github.com/jcholast/freeipa/tree/installer-refactoring if you want). I would propose to wait for it to get pushed (this Friday) and then rather reuse its code instead of having it copy pasted here :)

flo-renaud: Hi,
thanks for the heads-up. I will wait for the installer refactoring and reuse its code when it's ready.

tomaskrizek:
Works just like expected.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/044d887e81d433b43c33b076a21fd1054796786e

PR #223, opened 2016-11-09
LDAP refactoring: remove admin_conn

tomaskrizek:
This first commit removes the admin_conn alias for api.Backend.ldap2 that was previously used in services.

When trying to get rid of it, I found some legacy code in ipa-server-upgrade. The second commit improves ldap connection management in upgrade and removes useless start and stops of directory server at random places.

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -25,7 +25,6 @@ import ipalib.errors
 from ipaplatform import services
 from ipaplatform.tasks import tasks
 from ipapython import ipautil, sysrestore, version, certdb
-from ipapython import ipaldap
 from ipapython.ipa_log_manager import root_logger
 from ipapython import certmonger
 from ipapython import dnsutil
@@ -1606,9 +1605,6 @@ def upgrade_configuration():
     remove_ds_ra_cert(subject_base)
     ds.start(ds_serverid)
 
-    # Force enabling plugins via LDAPI and external bind
-    ds.ldapi = True
-    ds.autobind = ipaldap.AUTOBIND_ENABLED
     ds.fqdn = fqdn
     ds.realm = api.env.realm
     ds.suffix = ipautil.realm_to_suffix(api.env.realm)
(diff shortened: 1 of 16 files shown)


tomaskrizek:
Bump for review.

I forgot about this PR, it's a final change from the refactoring effort.

mbasti-rh:
Works for me, I'll check code tomorrow

mbasti-rh:
ACK: removing admin_conn connections
- just for future, python doesn't like to have too many dereference from performance point fo view. So you should in future use ldap = api.Backend.ldap2 and then just use ldap.find_entries()
- it is not issue with this patch as it was done in this way before, but can be improved in future

NACK: upgrade commit
- you turned DS on earlier than it actually needs to be. In one of first steps of LDAP upgrade is DS turned off, so I don't see a reason for a such change

tomaskrizek:
Fixed the issue + rebased. Only the second commit has changed.

ipaserver/install/server/upgrade.py
```diff
@@ -1606,24 +1604,14 @@ def upgrade_configuration():
     remove_ds_ra_cert(subject_base)
     ds.start(ds_serverid)
 
-    # Force enabling plugins via LDAPI and external bind
-    ds.ldapi = True
-    ds.autobind = ipaldap.AUTOBIND_ENABLED
     ds.fqdn = fqdn
     ds.realm = api.env.realm
     ds.suffix = ipautil.realm_to_suffix(api.env.realm)
-    ds.principal = "ldap/%s@%s" % (ds.fqdn, ds.realm)
```
tomaskrizek: Discovered this after rebase, this was probably a regression from installer refactoring. `principal` is now a property.

mbasti-rh: Please write this to installer refactoring ticket

mbasti-rh: OR remove it in separate patch

tomaskrizek:
Depends on #262

mbasti-rh:
LGTM and Works for me, but I have to make sure that things I wrote inline didn't happen

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/68295bf8cfd57333deb50f58df1b336a4b48ffe7
https://fedorahosted.org/freeipa/changeset/0914fc6a6043846159f6d1c4bb433dcfe9ee3f46

PR #226, opened 2016-11-10
Build refactoring phase 5

pspacek:
This PR fixes IPA_VERSION_IS_GIT_SNAPSHOT option and vendor version passing from SPEC to configure. At also contains minor cleanup and srpm target which is used by Coverity.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,8 @@ build/
 compile
 test-driver
 freeipa-*.tar.gz
+.tarball_name
+.version
 
 # Python compilation
 *.pyc
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,8 +4,7 @@ MOSTLYCLEANFILES = ipasetup.pyc ipasetup.pyo \
 		   ignore_import_errors.pyc ignore_import_errors.pyo \
 		   ipasetup.pyc ipasetup.pyo \
 		   lite-server.pyc lite-server.pyo \
-		   pylint_plugins.pyc pylint_plugins.pyo \
-		   $(TARBALL)
+		   pylint_plugins.pyc pylint_plugins.pyo
 
 # user-facing scripts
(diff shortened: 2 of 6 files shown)


pspacek:
I've added missing files to .gitignore.

pspacek:
Rebased on top of current master.

tomaskrizek:
I don't understand Makefiles, but I tested building the git snapshots and srpms and it works. Just a few notes:

- `make clean` removes only the most recently created tarball
- there is not much of a time difference when building with `IPA_VERSION_IS_GIT_SNAPSHOT` - it takes about 1m50s instead of 1m35s when it's turned off. Is this expected?

pspacek:
- I've fixed the tarball removal so it deletes all of them.
- The slow-down ~ 15 % is a nice surprise, I was under impression that it will be even slower.
- I forgot to enable it by default as it was before the refactoring, this is fixed as well.

Please re-review.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/394edf5f055766fa0cdf70afab6d263f75d0d065
https://fedorahosted.org/freeipa/changeset/f6f5708a5ac56392f7a4b82f63c5e16cc1f1fd99
https://fedorahosted.org/freeipa/changeset/3dc5d2c6f9a0fc134616c615634ae505ef753f77
https://fedorahosted.org/freeipa/changeset/a691b7d1837595fecd37bf88a875cb0753f7e698
https://fedorahosted.org/freeipa/changeset/0023fb59240cb53a541763a89db038c186312154
https://fedorahosted.org/freeipa/changeset/8c81c6c5b8ea62addbc175308a4e357c75d65ef0
https://fedorahosted.org/freeipa/changeset/961773bd0455e6fc723dbbed4f53e4b483360c32

PR #227, opened 2016-11-10
cert-request: match names against principal aliases

frasertweedale:
Currently we do not check Kerberos principal aliases when validating
a CSR.  Enhance cert-request to accept the following scenarios:

- for hosts and services: CN and SAN dnsNames match a principal
  alias (realm and service name must be same as nominated principal)

- for all principal types: UPN or KRB5PrincipalName othername match
  any principal alias.

Fixes: https://fedorahosted.org/freeipa/ticket/6295

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -649,11 +649,13 @@ class cert_request(Create, BaseCertMethod, VirtualCommand):
         cn = cns[-1].value  # "most specific" is end of list
 
         if principal_type in (SERVICE, HOST):
-            if cn.lower() != principal.hostname.lower():
-                raise errors.ACIError(
-                    info=_("hostname in subject of request '%(cn)s' "
-                        "does not match principal hostname '%(hostname)s'")
-                        % dict(cn=cn, hostname=principal.hostname))
+            if not _dns_name_matches_principal(cn, principal, principal_obj):
+                raise errors.ValidationError(
+                    name='csr',
+                    error=_(
+                        "hostname in subject of request '%(cn)s' does not "
+                        "match name or aliases of principal '%(principal)s'"
+                        ) % dict(cn=cn, principal=principal))
         elif principal_type == USER:
             # check user name
             if cn != principal.username:
@@ -686,26 +688,32 @@ class cert_request(Create, BaseCertMethod, VirtualCommand):
             generalnames = x509.process_othernames(ext_san.value)
         for gn in generalnames:
             if isinstance(gn, cryptography.x509.general_name.DNSName):
+                if principal.is_user:
+                    raise errors.ValidationError(
+                        name='csr',
+                        error=_(
+                            "subject alt name type %s is forbidden "
+                            "for user principals") % "DNSName"
+                    )
+
                 name = gn.value
-                alt_principal = None
+
+                if _dns_name_matches_principal(name, principal, principal_obj):
+                    continue  # nothing more to check for this alt name
+
+                # no match yet; check for an alternative principal with
+                # same realm and service type as subject principal.
+                components = list(principal.components)
+                components[-1] = name
+                alt_principal = kerberos.Principal(components, principal.realm)
                 alt_principal_obj = None
                 try:
                     if principal_type == HOST:
-                        alt_principal = kerberos.Principal(
-                            (u'host', name), principal.realm)
-                        alt_principal_obj = api.Command['host_show'](name, all=True)
+                        alt_principal_obj = api.Command['host_show'](
+                            name, all=True)
                     elif principal_type == SERVICE:
-                        alt_principal = kerberos.Principal(
-                            (principal.service_name, name), principal.realm)
                         alt_principal_obj = api.Command['service_show'](
                             alt_principal, all=True)
-                    elif principal_type == USER:
-                        raise errors.ValidationError(
-                            name='csr',
-                            error=_(
-                                "subject alt name type %s is forbidden "
-                                "for user principals") % "DNSName"
-                        )
                 except errors.NotFound:
                     # We don't want to issue any certificates referencing
                     # machines we don't know about. Nothing is stored in this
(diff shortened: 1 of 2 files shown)


martbab:
@frasertweedale What is the intended semantics of the checks against principal aliases in SAN? If the requestor can use only the aliases belonging to the entry of the recieving principal, then it should be enough to retrieve the entry by searching for the 'krbprincipalname' from --principal option, retrieve it, and then checking whether all values of dnsName/KRB5PrincipalName are a subset of Kerberos principal aliases.

martbab:
Also, the current execution flow of the command is very confusing (retrieving objects based on intended principal types etc.). As a part of the ticket I was planning to do a sneaky refactoring of the flow which IMHO should look like this:

1.) you search entries by krbprincipalname extracted from 'principal' option (or from bind principal)

2.) If not found, you error out that such entry could not be found

3.) due to syntax overrides in ipaldap, all returned principals will be converted to Principal objects so *after you retrieve the entry and ensure that it exists* you can test whether it is service, user, etc.

4.) for values in SAN, you check whether the value is already container in the entries principals (as you do in this PR). If the principal is not there, you can try to retrieve the entry from ldap and either error out if not found, or check CA ACLs against it when present.

5.) if all is OK, forward the request to RA backend and issue the certificate.

Do you think that this would extend the scope of the ticket too much? If yes, I can open a separate ticket for this cleanup and do it on top of your work.

martbab:
Also one of the tests in caacl_profile_enforcement suite fails:
https://paste.fedoraproject.org/481011/12920714/

frasertweedale:
@martbab thanks for review; I will revisit this some time in next week (hopefully)

frasertweedale:
@martbab 

Semantics:

0. *Subject principal* is looked up by `--principal` option, via `{PRINCIPAL_TYPE}_show` command.  If you think this should be extended to allow `--principal` to use an alias, I am cool with that.
1. For host and service principals, CN must match[dns] (described below) a principal alias.
2. For host and service principals, SAN dnsNames must match[dns] a principal alias, **or** match an alternative principal.
3. For all principals, SAN KRB5PrincipalName and UPN values must match[exact] a principal alias.

**match[dns]**: iterate principal aliases.  Matches if: alias has same realm as `--principal` **and** alias has same service name as `--principal` **and** alias hostname equals (case insensitively) the SAN dnsName value.  (If we generalise `--principal` to search all aliases then I would recommend restricting the search to principals with same realm and service name as the `krbcanonicalname` of the returned principal).

-----

w.r.t. test failure, I cannot reproduce with this patch rebased on latest master.

frasertweedale:
@martbab I agree with doing the refactor you propose, but I deem it out of scope for this ticket.

Doing that refactor entails a cleanup of how we handle the LDAP `;binary` transfer encoding option, because this is currently not consistent across user, host and service plugins.

So... do we want to look at getting this merged soon?

martbab:
I agree that it is out of scope of this work and will open a separate ticket for the `cert-request` refactoring. You may open another ticket for the handling of binary transfer option as I feel these effort can be done independently.

martbab:
There are small issues I found in the patch. Please address those.

@apophys can quickly you review the new test cases? they pass (except for one pointed out in review) so they should be valid and OK.

ipatests/test_xmlrpc/test_caacl_profile_enforcement.py
```diff
@@ -483,12 +478,79 @@ def test_prepare_caacl_profile(self, santest_subca_acl):
...
+
+    The CSR includes a DNS name that shall correspond to a Kerberos
+    alias of the service.  Request should succeed.
+
+    """
+    def test_add_principal_alias(
+            self, santest_service_host_1):
+        api.Command.service_add_principal(
+            santest_service_host_1.name,
+            'srv/santest-host-2.ipa.local')
```
frasertweedale: Thanks for clarifying.  I was wonder how the test was failing.  It is obvious now thanks to your explanation :)

ipatests/test_xmlrpc/test_caacl_profile_enforcement.py
```diff
@@ -483,12 +478,79 @@ def test_prepare_caacl_profile(self, santest_subca_acl):
...
+                    api.Command.cert_request(
+                        santest_csr,
+                        principal=santest_service_host_1.name,
+                        cacn=santest_subca.name
+                    )
+
+
+@pytest.mark.tier1
+class TestPrincipalAliasForSubjectAltNameDnsName(SubjectAltNameOneServiceBase):
+    """Sign certificate request with an invalid SAN dnsName.
```
frasertweedale: Yep, thanks. Corrected in updated patch.

ipaserver/plugins/cert.py
```diff
@@ -787,6 +800,62 @@ def execute(self, csr, all=False, raw=False, **kw):
...
+
+
+def _collect_principal_aliases(principal_obj):
+    """
+    Return all Kerberos ``Principal`` aliases for given principal
+    (including the canonical name).
+
+    """
+    aliases = set(principal_obj.get('krbprincipalname', []))
+    for alias in principal_obj.get('krbcanonicalname', []):
```
frasertweedale: OK, thanks.  I've removed function `_collect_princpial_aliases()` and just replaced with `principal_obj.get('krbprincipalname', [])` at call sites.

ipaserver/plugins/cert.py
```diff
@@ -787,6 +800,62 @@ def execute(self, csr, all=False, raw=False, **kw):
...
+    """
+    for alias in _collect_principal_aliases(principal_obj):
+        # we can only compare them if both subject principal and
+        # the alias are service or host principals
+        if not (alias.is_service and principal.is_service):
+            continue
+
+        # ignore aliases with different realm or service name from
+        # subject principal
+        if alias.realm != principal.realm:
```
frasertweedale: I'm going to leave it in anyway.  It does no harm and is one less potential bug if we do ever support aliases with different realms, somehow.

martbab: Ok

ipaserver/plugins/cert.py
```diff
@@ -787,6 +800,50 @@ def execute(self, csr, all=False, raw=False, **kw):
...
+    Ensure that a stringy principal name (e.g. from UPN
+    or KRB5PrincipalName OtherName) matches the given principal.
+
+    """
+    try:
+        principal = kerberos.Principal(name)
+    except ValueError:
+        return False
+
+    return principal in principal_obj.get('krbprincpialname', [])
```
martbab: Please fix typo in attribute name.

frasertweedale: hah, silly fat fingers.  push incoming

apophys:
The tests look good to me.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/dfbdb5323863e6c3d681c1b33b1eb9d2efefd6c7

PR #229, opened 2016-11-10
Remove the renewal lock file upon uninstall

flo-renaud:
Make sure that the file /var/run/ipa/renewal.lock is deleted upon
uninstallation, in order to avoid subsequent installation issues.

Part of the refactoring effort, certificates sub-effort.

https://fedorahosted.org/freeipa/ticket/6433

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -4,6 +4,7 @@
 
 from __future__ import print_function
 
+import errno
 import os
 import pickle
 import shutil
@@ -1117,6 +1118,14 @@ def uninstall(installer):
                           'These may be untracked by executing\n'
                           ' # getcert stop-tracking -i <request_id>\n'
                           'for each id in: %s' % ', '.join(ids))
+
+    # Remove the cert renewal lock file
+    try:
+        os.remove(paths.IPA_RENEWAL_LOCK)
+    except OSError as e:
+        if e.errno != errno.ENOENT:
+            root_logger.warning("Failed to remove file %s: %s",
+                                paths.IPA_RENEWAL_LOCK, e)
 
     print("Removing IPA client configuration")
     try:
--- a/install/restart_scripts/renew_ra_cert
+++ b/install/restart_scripts/renew_ra_cert
@@ -30,7 +30,6 @@ import traceback
 from ipapython import ipautil
 from ipalib import api
 from ipaserver.install import certs, cainstance, krainstance
-from ipaplatform import services
 from ipaplatform.paths import paths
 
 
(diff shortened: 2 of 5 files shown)


jcholast:
The file is owned by the server, not the client, so it should be deleted in `ipa-server-install --uninstall`, not in `ipa-client-install --uninstall`.

flo-renaud:
You are right, I updated the PR to put the code at the end of server uninstallation.

frasertweedale:
Works as expected.

ipapython/certmonger.py
```diff
@@ -307,17 +307,30 @@ def request_and_wait_for_cert(
     The method also waits for the certificate to be available.
     """
     reqId = request_cert(nssdb, nickname, subject, principal,
-                         passwd_fname, dns, ca, profile,
-                         pre_command, post_command)
-    state = wait_for_request(reqId, timeout=60)
+                         passwd_fname, dns, ca, profile)
+    state = wait_for_request(reqId, timeout=180)
```
jcholast: Could you please put the timeout increase into a separate commit?

ipaserver/install/server/install.py
```diff
@@ -1130,6 +1130,12 @@ def uninstall(installer):
                           ' # getcert stop-tracking -i <request_id>\n'
                           'for each id in: %s' % ', '.join(ids))
 
+    # Remove the cert renewal lock file
+    try:
+        os.remove(paths.IPA_RENEWAL_LOCK)
+    except Exception:
+        pass
```
jcholast: This would be better (off the top of my head):
```python
    try:
        os.remove(paths.IPA_RENEWAL_LOCK)
    except OSError as e:
        if e.errno != errno.ENOENT:
            root_logger.warning("Failed to remove %s: %s", paths.IPA_RENEWAL_LOCK, e)
```

ipapython/certmonger.py
```diff
@@ -307,17 +307,30 @@ def request_and_wait_for_cert(
...
     ca_error = get_request_value(reqId, 'ca-error')
     if state != 'MONITORING' or ca_error:
         raise RuntimeError("Certificate issuance failed")
+
+    # Add presave and postsave commands
+    # This is not done earlier to avoid running the cmds
+    # during the initial request
```
jcholast: Instead of this, I would rather not restart the services in `renew_ra_cert`, `restart_httpd` and `restart_dirsrv` if they were not previously running.

ipapython/certmonger.py
jcholast: Maybe it might not even be necessary with the other fixes.

ipapython/certmonger.py
jcholast: Speaking of which, I don't think the restart in `renew_ra_cert` is necessary at all - httpd itself does not use the RA cert in any way, it is used only in the framework.

flo-renaud:
Hi,
I implemented @jcholast suggestions and finally found the origin of the lock.

mbasti-rh:
Works for me on both domain levels, I'd ACK this if nobody is against

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/198cd5fab3937fd8948bea4b4949e30db4e490a4

PR #231, opened 2016-11-10
Do not log DM password in ca/kra installation logs

stlaz:
We can merge this after refactoring merges not to mess the rebases.

https://fedorahosted.org/freeipa/ticket/6461

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -576,7 +576,10 @@ class CAInstance(DogtagInstance):
 
         self.backup_state('installed', True)
         try:
-            DogtagInstance.spawn_instance(self, cfg_file)
+            DogtagInstance.spawn_instance(
+                self, cfg_file,
+                nolog_list=(self.dm_password, self.admin_password)
+            )
         finally:
             os.remove(cfg_file)
 
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -150,19 +150,13 @@ class DogtagInstance(service.Service):
         return os.path.exists(os.path.join(
             paths.VAR_LIB_PKI_TOMCAT_DIR, self.subsystem.lower()))
 
-    def spawn_instance(self, cfg_file, nolog_list=None):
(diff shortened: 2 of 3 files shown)


martbab:
I would rather hide the password by default in the `spawn_instance` method in the same manner as is done for admin_password, see https://git.fedorahosted.org/cgit/freeipa.git/tree/ipaserver/install/dogtaginstance.py?id=f183f70e0183e51d569ada972bd3ec73cad76a30#n166

stlaz:
I must have misclicked "close" when viewing this PR on my phone. I believe we may rather add admin and DM passwords to the nolog_list at the point where the disclosed credentials file is created so that we avoid problems like this one in the future.

martbab:
Well that is what I was pointing at, by adding both DM and admin passwords to the parent method's default `nolog_list`, you are future-proofing the code because all spawn-instance calls will be safe. But maybe I am missing something.

tomaskrizek:
I agree. We need to re-add `self.dm_password` to `nolog_list`, just like it was before I removed it [here](https://github.com/freeipa/freeipa/commit/9340a1417acf120fed3e9ffbe9d658d3456743a1#diff-36dfe273a301d6b5ea2bbcf89c7cd661L167)

There is no reason to change it. I originally removed the line, because I thought I could remove `dm_password` from `DogtagInstance` all together, but that turned out not to be the case.

stlaz:
@martbab Oh, I thought you wanted me to re-add `dm_password` to DogtagInstance as @tomaskrizek which does not seem right as DogtagInstance is in no position to decide what to log and what not as it does not really know what's in that cfg_file it's getting.
Will get it passed from the actual caller of `spawn_instance` which is either cainstance or krainstance.

tomaskrizek:
I didn't notice `dm_password` is no longer in `DogtagInstance`, I re-added it elsewhere. In that case, as @stlaz said, passing it to `spawn_instance()` seems like the proper way to do it.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/e617f895e70e6812836870f504af6e22a5dc7def

PR #233, opened 2016-11-11
Build phase 6: %install cleanup

pspacek:
This patch set is based on phase 5 - PR #226.

Now all the installation steps (except Python2/3) are handled by make install.

Python 2/3 support will deserve separate PR.

### DIFF
--- a/install/Makefile.am
+++ b/install/Makefile.am
@@ -19,12 +19,11 @@ SUBDIRS =			\
 	$(NULL)
 
 install-exec-local:
-	mkdir -p $(DESTDIR)$(localstatedir)/lib/ipa/sysrestore
-	chmod 700 $(DESTDIR)$(localstatedir)/lib/ipa/sysrestore
-	mkdir -p $(DESTDIR)$(localstatedir)/lib/ipa/sysupgrade
-	chmod 700 $(DESTDIR)$(localstatedir)/lib/ipa/sysupgrade
-	mkdir -p $(DESTDIR)$(localstatedir)/lib/ipa/pki-ca
-	chmod 755 $(DESTDIR)$(localstatedir)/lib/ipa/pki-ca
+	$(INSTALL) -d -m 700 $(DESTDIR)$(IPA_SYSCONF_DIR)/custodia
+	$(INSTALL) -d -m 700 $(DESTDIR)$(localstatedir)/lib/ipa/backup
+	$(INSTALL) -d -m 700 $(DESTDIR)$(localstatedir)/lib/ipa/sysrestore
+	$(INSTALL) -d -m 700 $(DESTDIR)$(localstatedir)/lib/ipa/sysupgrade
+	$(INSTALL) -d -m 755 $(DESTDIR)$(localstatedir)/lib/ipa/pki-ca
 
 uninstall-local:
 	-rmdir $(DESTDIR)$(localstatedir)/lib/ipa/sysrestore
--- a/install/html/Makefile.am
+++ b/install/html/Makefile.am
@@ -16,3 +16,17 @@ EXTRA_DIST =                            \
 MAINTAINERCLEANFILES =                  \
         *~                              \
         Makefile.in
+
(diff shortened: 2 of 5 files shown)


pspacek:
Rebased on top of current master.

mbasti-rh:
Build failed
```
Failed to open: 'freeipa.spec.in', not a valid spec file.
```

install/Makefile.am
```diff
@@ -19,6 +19,10 @@ SUBDIRS =			\
 	$(NULL)
 
 install-exec-local:
+	mkdir -p $(DESTDIR)$(IPA_SYSCONF_DIR)/custodia
+	chmod 700 $(DESTDIR)$(IPA_SYSCONF_DIR)/custodia
```
tiran: Please use ```$(INSTALL) -d -m 700 -o root -g root``` here.

install/html/Makefile.am
```diff
@@ -16,3 +16,16 @@ EXTRA_DIST =                            \
...
+
+# Default user-modifiable HTML files are installed into /etc.
+# /usr points to these modifiable files in /etc
+# This is ugly but we do not have time to change it right now.
+# Relative paths must be used to ensure that symlinks created in buildroot
+# work after installation.
+htmldatadir = $(datarootdir)/ipa/html
+install-data-hook:
+	$(INSTALL) -d -D -m 755 $(DESTDIR)$(htmldatadir)
+	RELATIVE_PATH=$$(python -c "import os.path; print(os.path.relpath('$(appdir)', '$(htmldatadir)'))");	\
```
tiran: You don't need Python here. ln has ```-r``` for ```create symbolic links relative to link location```.

install/ui/Makefile.am
```diff
@@ -34,3 +34,6 @@ MAINTAINERCLEANFILES =                  \
         *~                              \
         Makefile.in
 	$(NULL)
+
+install-data-hook:
+	$(INSTALL) -d -D -m 755 $(DESTDIR)$(appdir)/js/plugins
```
tiran: Please make sure you install files and directories as root:root.

pspacek: make install cannot install files as root (or any other user) because make intall is typically called as unprivileged user by the rpmbuild.

install/Makefile.am
pspacek: I wanted to be consistent but good point, I will change it to $(INSTALL). `-o root ` cannot be used as explained below. Changing file ownership is job of package manager.

install/html/Makefile.am
pspacek: Nice! I will use that.

pspacek:
> Failed to open: 'freeipa.spec.in', not a valid spec file.

Damn it! I added the last commit with comment in SPEC file and did not run tests on that. Surprise surprise, RPM is PARSIN TEXT INSIDE COMMENTS and if you put %install into comment, guess what: It will blow up! Grrrr. (Another joke is that DNF builddep is throwing away error descrption from RPM SPEC file parser.)

pspacek:
This version fixes the fixable issues, i.e. everything mentioned above except changing file ownership.

pspacek:
@tiran Please re-review and set review status accordingly. Thanks!

pspacek:
I've fixed incorrect use of `-D` in `install` calls above.

mbasti-rh:
@tiran do you agree with changes?

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/636aaa7dbc649e685233f382cb8dd424345bebd3
https://fedorahosted.org/freeipa/changeset/20918579acb43391d5d04ee8050b37142a55df76
https://fedorahosted.org/freeipa/changeset/1fa0ed954bb45b6e3858c1c54470b1d16ab204d9
https://fedorahosted.org/freeipa/changeset/5a5373464fa67289c9a178b1c0569f585dc6dc34

PR #235, opened 2016-11-11
Remove unused Knob function

mbasti-rh:
`Knob` function is outdated and was replaced by `knob`. Make explicit
note in code about this.

### DIFF
--- a/ipapython/install/core.py
+++ b/ipapython/install/core.py
@@ -8,21 +8,18 @@ The framework core.
 
 import abc
 import collections
-import enum
 import functools
 import itertools
-import re
 import sys
 
 import six
 
 from ipapython.ipa_log_manager import root_logger
-from ipapython.ipautil import CheckedIPAddress
 
-from . import util, typing
+from . import util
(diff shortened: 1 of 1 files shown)


stlaz:
ACK, there should be note about this deprecation somewhere. Deleting Knob might be worth a ticket as well.

mbasti-rh:
I would wait with ACK, I realized that Knobs with capital K are not used anymore, so we can remove it instead of deprecating

stlaz:
From our offline discussion I got the impression the Knob function was still used somewhere, therefore the ACK. I'm not sure what was the reason of keeping Knob there even if unused, you may need checking with @jcholast.

mbasti-rh:
@jcholast ping

jcholast:
There was no reason, I just forgot, so go ahead.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/55b14abcb561422cf48755dae6b0638656535fe5

PR #236, opened 2016-11-11
Build phase 7: cleanup

pspacek:
Depends on PR #233.
- Clean-up ancient leftovers and clean minor bugs here and there.
- Support --enable-silent-rules and V=0 variable for make to make the build less verbose and warnings more visible.


client/Makefile.am
```diff
@@ -30,26 +27,22 @@ AM_CPPFLAGS =							\
 	$(SASL_CFLAGS)						\
 	$(POPT_CFLAGS)						\
 	$(WARN_CFLAGS)						\
-	$(INI_CFLAGS)						\
-	$(NULL)
+	$(INI_CFLAGS)
```
lslebodn: The reason why is $(NULL) is at the end of list is simpler diff when adding new entries.

If there is not NULL at the end and you want to add new entry there then diff will contain 1 removed line + 2 added lines

If there *is* NULL at the end and  you want to add new entry there  then diff will contain just 1 added line.

lslebodn:
You should the opposite of the last patch "Build: clean-up spurious NULL variables from Makefile.am files". The NULL should be added to each list.

NACK to the last patch. It does not simplify anything and it makes diff more complicated for adding new entries

client/Makefile.am
```diff
@@ -2,13 +2,10 @@
 
 AUTOMAKE_OPTIONS = 1.7 subdir-objects
 
-NULL =
-
```
lslebodn: These lines can be removed because NULL needn't be *defined*.

pspacek:
Hi Lukas. Given there is no technical justification to have it I'm going to remove these. Simple is better than complex.

lslebodn:
>Hi Lukas. Given there is no technical justification to have it I'm going to remove these. Simple is better than complex.
I am sorry I do not agree. The technical justification was explained in previous comments few time. The $(NULL) at the end of list makes patches much simpler and easier to read. The purpose of refactoring is make code simpler but also easier to **maintain/review**

makerpms.sh
```diff
@@ -4,7 +4,26 @@ set -o errexit
 pushd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 test ! -x "configure" && autoreconf -i
-test ! -f "Makefile" && ./configure "$@"
+# run configure with the same parameters as RPM build
+# this makes it easy to tweak files locally and use make install
+test ! -f "Makefile" && ./configure --enable-silent-rules \
+	--host=$(rpm -E %{_host}) \
```
tiran: NACK

rpm should be an optional dependency to support platforms without rpm command.

lslebodn: @tiran can you explain how a script with the name **makerpms.sh** cannot be rpm specific?

pspacek:
@lslebodn I've dropped the controversial patch which removes NULLs to allow you to send PR which adds it everywhere as you proposed. I'm going to wait till deadline at end of this working week. If your PR is not revied by then I'm going to push the removal so things are at least consistent.

lslebodn:
Thank you very much for removing problematic patch. Do not forget to remove it from "Build phase 8" :-) as well

A) there are few conflicts => patch need to be rebased

B) I do not think that @tiran objection about rpm in makerpms.sh is valid but he should reply.

Otherwise ACK after rebase
Nice work and thank you very much for it.

BTW is there a plan to have "make dictcheck" functional?

pspacek:
Rebased.

pspacek:
@lslebodn There is currently no plan to support distcheck: Python setuptools do not support VPATH builds as AFAIK it is impossible to do that without patching setuptools heavily.

pspacek:
ACKing on behalf of Lukas.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/27e7a89a6289d0d3009f5f7feb9802b7db171a15
https://fedorahosted.org/freeipa/changeset/46b6b9e3091d08412912c37f46af497ddd0b8afb
https://fedorahosted.org/freeipa/changeset/1cbd823990da0e931b666c4bc5c72f10d9de8115
https://fedorahosted.org/freeipa/changeset/d5683726d290b71eb44ab3b3150381f062e74df1
https://fedorahosted.org/freeipa/changeset/e2060e8e5562ed6a4fe760eba1babb5c1761576a
https://fedorahosted.org/freeipa/changeset/6b9977f04199bf161d7171aedae9f97648c415c8

PR #243, opened 2016-11-16
Don't modify redhat_system_units

tiran:
ipaplatform.fedora.services used to modify the redhat_system_units dict.
It now creates a proper shallow copy.


martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/94a9dfb9d72ecd25a01316febbf2ffec50912e2e

PR #245, opened 2016-11-16
Allow full customisability of IPA CA subject DN

frasertweedale:
This patchset adds full customisability of CA subject DN apart from
subject base, via the ipa-server-install `--ca-subject` option.  It
also renames ipa-server-install `--subject` option to
`--subject-base`, and adds `--ca-subject` and `--subject-base`
options to ipa-ca-install.

Earlier version of this patchset was previously reviewed by
@jcholast on freeipa-devel:
https://www.redhat.com/archives/freeipa-devel/2016-August/msg00570.html

All review items have been addressed, except for item 9.  The
suggestion will not work because a fair bit of code besides
what's in `ipaserver.install.ca` requires knowing the CA Subject DN
and/or subject base.  So the defaults must be applied close to the
"entry points".

I also carved a few smaller commits out of the main patch (but it is
still pretty big and hairy to review).

https://fedorahosted.org/freeipa/ticket/2614

### DIFF
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -80,12 +80,14 @@ def parse_options():
                       default=None,
                       help=(
                           "The certificate subject base "
-                          "(default O=<realm-name>)"))
+                          "(default O=<realm-name>).  "
+                          "RDNs are in LDAP order (most specific RDN first)."))
     parser.add_option("--ca-subject", dest="ca_subject",
                       default=None,
                       help=(
                           "The CA certificate subject DN "
-                          "(default CN=Certificate Authority,O=<realm-name>)"))
+                          "(default CN=Certificate Authority,O=<realm-name>). "
+                          "RDNs are in LDAP order (most specific RDN first)."))
 
     options, args = parser.parse_args()
     safe_options = parser.get_safe_opts(options)
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -122,6 +122,11 @@ def get_dirman_password():
 
 
 def install_replica(safe_options, options, filename):
+    if options.ca_subject:
+        sys.exit("--ca-subject cannot be used when installing a CA replica")
+    if options.subject_base:
+        sys.exit("--subject-base cannot be used when installing a CA replica")
+
     if options.promote:
         if filename is not None:
             sys.exit("Too many parameters provided. "
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -81,6 +81,11 @@ def parse_options():
                       help=(
                           "The certificate subject base "
                           "(default O=<realm-name>)"))
+    parser.add_option("--ca-subject", dest="ca_subject",
+                      default=None,
+                      help=(
+                          "The CA certificate subject DN "
+                          "(default CN=Certificate Authority,O=<realm-name>)"))
 
     options, args = parser.parse_args()
     safe_options = parser.get_safe_opts(options)
@@ -177,7 +182,6 @@ def install_replica(safe_options, options, filename):
     options.domain_name = config.domain_name
     options.dm_password = config.dirman_password
     options.host_name = config.host_name
-    options.subject_base = config.subject_base
     if os.path.exists(cafile):
         options.ca_cert_file = cafile
     else:
@@ -206,6 +210,18 @@ def install_master(safe_options, options):
 
     if not options.subject_base:
         options.subject_base = installutils.default_subject_base(api.env.realm)
+    if not options.ca_subject:
+        options.ca_subject = installutils.default_ca_subject_dn(
+            options.subject_base)
+
+    try:
+        ca.subject_validator(ca.VALID_SUBJECT_BASE_ATTRS, options.subject_base)
+    except ValueError as e:
+        sys.exit("Subject base: {}".format(e.message))
+    try:
+        ca.subject_validator(ca.VALID_SUBJECT_ATTRS, options.ca_subject)
+    except ValueError as e:
+        sys.exit("CA subject: {}".format(e.message))
 
     ca.install_check(True, None, options)
     ca.install(True, None, options)
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -172,7 +172,7 @@ def install_replica(safe_options, options, filename):
     options.domain_name = config.domain_name
     options.dm_password = config.dirman_password
     options.host_name = config.host_name
-    options.subject = config.subject_base
+    options.subject_base = config.subject_base
     if os.path.exists(cafile):
         options.ca_cert_file = cafile
     else:
@@ -201,7 +201,7 @@ def install_master(safe_options, options):
     options.domain_name = api.env.domain
     options.dm_password = dm_password
     options.host_name = api.env.host
-    options.subject = subject_base
+    options.subject_base = subject_base
 
     ca.install_check(True, None, options)
     ca.install(True, None, options)
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -76,6 +76,11 @@ def parse_options():
                       help="Signing algorithm of the IPA CA certificate")
     parser.add_option("-P", "--principal", dest="principal", sensitive=True,
                       default=None, help="User allowed to manage replicas")
+    parser.add_option("--subject-base", dest="subject_base",
+                      default=None,
+                      help=(
+                          "The certificate subject base "
+                          "(default O=<realm-name>)"))
 
     options, args = parser.parse_args()
     safe_options = parser.get_safe_opts(options)
@@ -194,14 +199,13 @@ def install_master(safe_options, options):
         if dm_password is None:
             sys.exit("Directory Manager password required")
 
-    config = api.Command['config_show']()['result']
-    subject_base = config['ipacertificatesubjectbase'][0]
-
     options.realm_name = api.env.realm
     options.domain_name = api.env.domain
     options.dm_password = dm_password
     options.host_name = api.env.host
-    options.subject_base = subject_base
+
+    if not options.subject_base:
+        options.subject_base = installutils.default_subject_base(api.env.realm)
 
     ca.install_check(True, None, options)
     ca.install(True, None, options)
--- a/install/tools/man/ipa-ca-install.1
+++ b/install/tools/man/ipa-ca-install.1
@@ -46,6 +46,12 @@ Type of the external CA. Possible values are "generic", "ms-cs". Default value i
 \fB\-\-external\-cert\-file\fR=\fIFILE\fR
 File containing the IPA CA certificate and the external CA certificate chain. The file is accepted in PEM and DER certificate and PKCS#7 certificate chain formats. This option may be used multiple times.
 .TP
+\fB\-\-ca\-subject\fR=\fISUBJECT\fR
+The CA certificate subject DN (default CN=Certificate Authority,O=REALM.NAME)
+.TP
+\fB\-\-subject\-base\fR=\fISUBJECT\fR
+The subject base for certificates issued by IPA (default O=REALM.NAME)
+.TP
 \fB\-\-ca\-signing\-algorithm\fR=\fIALGORITHM\fR
 Signing algorithm of the IPA CA certificate. Possible values are SHA1withRSA, SHA256withRSA, SHA512withRSA. Default value is SHA256withRSA. Use this option with --external-ca if the external CA does not support the default signing algorithm.
 .TP
--- a/install/tools/man/ipa-ca-install.1
+++ b/install/tools/man/ipa-ca-install.1
@@ -47,10 +47,10 @@ Type of the external CA. Possible values are "generic", "ms-cs". Default value i
 File containing the IPA CA certificate and the external CA certificate chain. The file is accepted in PEM and DER certificate and PKCS#7 certificate chain formats. This option may be used multiple times.
 .TP
 \fB\-\-ca\-subject\fR=\fISUBJECT\fR
-The CA certificate subject DN (default CN=Certificate Authority,O=REALM.NAME)
+The CA certificate subject DN (default CN=Certificate Authority,O=REALM.NAME).  RDNs are in LDAP order (most specific RDN first).
 .TP
 \fB\-\-subject\-base\fR=\fISUBJECT\fR
-The subject base for certificates issued by IPA (default O=REALM.NAME)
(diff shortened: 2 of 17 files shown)


install/tools/man/ipa-ca-install.1
```diff
@@ -46,6 +46,12 @@ Type of the external CA. Possible values are "generic", "ms-cs". Default value i
 \fB\-\-external\-cert\-file\fR=\fIFILE\fR
 File containing the IPA CA certificate and the external CA certificate chain. The file is accepted in PEM and DER certificate and PKCS#7 certificate chain formats. This option may be used multiple times.
 .TP
+\fB\-\-ca\-subject\fR=\fISUBJECT\fR
+The CA certificate subject DN (default CN=Certificate Authority,O=REALM.NAME)
+.TP
+\fB\-\-subject\-base\fR=\fISUBJECT\fR
+The subject base for certificates issued by IPA (default O=REALM.NAME)
+.TP
```
tiran: Please add a comment that the order of RDNs is in LDAP order, not in X.509 order. It might cause some confusing. Users (like me) expect the order of RDNs like in OpenSSL or Firefox.

https://fedorahosted.org/freeipa/ticket/6455

install/tools/man/ipa-server-install.1
```diff
@@ -129,8 +129,11 @@ Name of the Kerberos KDC SSL certificate to install
 \fB\-\-ca\-cert\-file\fR=\fIFILE\fR
 File containing the CA certificate of the CA which issued the Directory Server, Apache Server and Kerberos KDC certificates. The file is accepted in PEM and DER certificate and PKCS#7 certificate chain formats. This option may be used multiple times. Use this option if the CA certificate is not present in the certificate files.
 .TP
-\fB\-\-subject\fR=\fISUBJECT\fR
-The certificate subject base (default O=REALM.NAME)
+\fB\-\-ca\-subject\fR=\fISUBJECT\fR
+The CA certificate subject DN (default CN=Certificate Authority,O=REALM.NAME)
+.TP
+\fB\-\-subject\-base\fR=\fISUBJECT\fR
+The subject base for certificates issued by IPA (default O=REALM.NAME)
```
tiran: dito

ipaserver/install/ca.py
```diff
@@ -35,18 +35,41 @@
...
-                       'uid', 'postaladdress', 'postalcode', 'postofficebox',
-                       'houseidentifier', 'e', 'street', 'pseudonym',
-                       'incorporationlocality', 'incorporationstate',
-                       'incorporationcountry', 'businesscategory']
+VALID_SUBJECT_BASE_ATTRS = {
+    'st', 'o', 'ou', 'dnqualifier', 'c', 'serialnumber', 'l', 'title', 'sn',
+    'givenname', 'initials', 'generationqualifier', 'dc', 'mail', 'uid',
+    'postaladdress', 'postalcode', 'postofficebox', 'houseidentifier', 'e',
+    'street', 'pseudonym', 'incorporationlocality', 'incorporationstate',
+    'incorporationcountry', 'businesscategory',
```
tiran: OpenSSL has slightly different short names (email instead of e).

frasertweedale: I don't think it's an issue.  I didn't change any of them, anyhow - so if it is a problem we can address in another ticket / commit.

frasertweedale:
Added a new commit to add DNs-in-ldap-order comments.

ipaserver/install/ca.py
tiran: I wonder if we should mention the RDNs in the man page. After all OpenSSL is a de-facto standard.

tiran:
flake8 violation: ```./ipaserver/plugins/migration.py:750:80: E501 line too long (83 > 79 characters)```

frasertweedale:
@tiran I haven't a clue about that pep8 error... my commits don't even touch that file.

tiran:
@frasertweedale I don't have a clue, either. Let's ignore the message.

install/tools/ipa-ca-install
```diff
@@ -71,6 +78,16 @@ def parse_options():
                       help="Signing algorithm of the IPA CA certificate")
     parser.add_option("-P", "--principal", dest="principal", sensitive=True,
                       default=None, help="User allowed to manage replicas")
+    parser.add_option("--subject-base", dest="subject_base",
+                      default=None,
+                      help=(
+                          "The certificate subject base "
+                          "(default O=<realm-name>)"))
```
jcholast: IMO this deservers a separate "add missing --subject-base to ipa-ca-install", because the option should have been there since `ipa-ca-install` gained the ability to install CA master.

frasertweedale: As you wish.  PR updated.

ipaserver/install/ca.py
```diff
@@ -57,9 +80,14 @@ def install_check(standalone, replica_config, options):
...
     host_name = options.host_name
-    subject_base = options.subject
 
-    if replica_config is not None:
+    if replica_config is None:
+        ca_subject = options.ca_subject
+        subject_base = options.subject_base
+    else:
+        ca_subject = lookup_ca_subject(options.realm_name)
+        subject_base = replica_config.subject_base
```
jcholast: Rather than duplicating this below in `install_step_0()` and `install_step_1()`, it would be better to store the values in `options._ca_subject` and `options._subject_base` and reuse them in `install_step_0()` and `install_step_1()`.

frasertweedale:
@jcholast: new tickets pertaining to subject_base / certmap.conf config:

- **do not update ipaCertificateSubjectBase and certmap.conf in CA-less mode** 
  - https://fedorahosted.org/freeipa/ticket/6556
- **do not set (or look up) subject_base in sysupgrade file**
  - https://fedorahosted.org/freeipa/ticket/6557

Other review comments will be addressed in due course.  Thanks for reviewing.

ipaserver/install/server/install.py
```diff
@@ -255,7 +255,9 @@ def set_subject_in_config(realm_name, dm_password, suffix, subject_base):
                                  "on %s" % realm_name)
             raise e
         entry_attrs = conn.get_ipa_config()
-        if 'ipacertificatesubjectbase' not in entry_attrs:
+        cur_subject_base = entry_attrs.get('ipacertificatesubjectbase', [None])
+        if not cur_subject_base[0] \
+                or DN(cur_subject_base[0]) != DN(subject_base):
             entry_attrs['ipacertificatesubjectbase'] = [str(subject_base)]
             conn.update_entry(entry_attrs)
```
frasertweedale: That is nicer; will do.

ipaserver/install/server/replicainstall.py
```diff
@@ -704,6 +709,10 @@ def install_check(installer):
         raise RuntimeError("CA cert file is not available. Please run "
                            "ipa-replica-prepare to create a new replica file.")
 
+    # look up CA subject name (needed for DS certmap.conf)
+    options.subject = unicode(
+        DN(x509.load_certificate_from_file(cafile).subject))
```
frasertweedale: Yes; thanks.

ipaserver/install/dsinstance.py
```diff
@@ -1067,7 +1076,9 @@ def add_ca_cert(self, cacert_fname, cacert_name=''):
         self.stop()
 
         dirname = config_dirname(installutils.realm_to_serverid(self.realm))
-        certdb = certs.CertDB(self.realm, nssdir=dirname, subject_base=self.subject_base)
+        certdb = certs.CertDB(
+                self.realm, nssdir=dirname,
+                subject_base=self.subject_base, ca_subject=self.subject)
```
frasertweedale: Agreed.  (Disagree with first approach though; what if we rename var `certdb`, module `certs` or class `CertDB` and length changes?  You don't want to realign `N` extra lines for all the arguments :)

ipaserver/install/cainstance.py
```diff
@@ -330,7 +330,8 @@ def configure_instance(self, host_name, dm_password, admin_password,
                            pkcs12_info=None, master_host=None, csr_file=None,
                            cert_file=None, cert_chain_file=None,
                            master_replication_port=None,
-                           subject_base=None, ca_signing_algorithm=None,
+                           subject_base=None, subject=None,
```
frasertweedale: OK.

install/tools/ipa-ca-install
```diff
@@ -106,6 +125,33 @@ def get_dirman_password():
...
+    with open(certmap_filename) as f:
+        certmap_conf = f.read()
+    certmap_conf = re.sub(
+        r'^(certmap ipaca\s+)(.*)$',
+        r'\1{}'.format(ca_subject),
+        certmap_conf,
+        0,
+        re.MULTILINE)
+    with open(certmap_filename, 'w') as f:
+        f.write(certmap_conf)
```
frasertweedale: OK, no worries cobber.

install/tools/ipa-ca-install
```diff
@@ -106,6 +125,33 @@ def get_dirman_password():
         "Directory Manager (existing master)", confirm=False, validate=False)
 
 
+def reset_subject_base_config(dm_password, realm, subject_base):
+    """(Re)set CA subject base config in all relevant places."""
+    # LDAP
+    set_subject_base_in_config(realm, dm_password, subject_base)
+
+    # sysupgrade
+    sysupgrade.set_upgrade_state('certmap.conf', 'subject_base', subject_base)
```
frasertweedale: I'm wary of the greater scope of that change (hence I filed a separate ticket for it).

I really had no choice but to remove the `certmap.conf` lookup from `find_subject_base()` for this ticket though.

install/tools/ipa-ca-install
```diff
@@ -169,12 +215,17 @@ def install_replica(safe_options, options, filename):
...
     if os.path.exists(cafile):
         options.ca_cert_file = cafile
     else:
         options.ca_cert_file = None
 
+    # look up CA subject name (needed for DS certmap.conf)
+    ipa_ca_nickname = get_ca_nickname(config.realm_name)
+    db = certs.CertDB(config.realm_name, nssdir=paths.IPA_NSSDB_DIR)
+    cert = db.get_cert_from_db(ipa_ca_nickname)
+    options.ca_subject = six.text_type(DN(x509.load_certificate(cert).subject))
```
frasertweedale: Yep, it was done here to calculate `ca_subject` for `reset_subject_config()`.  Per your other review comment, this will be moved to `ca.install()` so this code can go away.

ipaserver/install/ca.py
```diff
@@ -36,18 +36,41 @@
...
+                raise ValueError("invalid attribute: \"%s\"" % rdn.attr)
+    except ValueError as e:
+        raise ValueError("invalid subject base format: %s" % e)
+
+
+def lookup_ca_subject(realm_name):
+    ipa_ca_nickname = certdb.get_ca_nickname(realm_name)
+    db = certs.CertDB(realm_name, nssdir=paths.IPA_NSSDB_DIR)
+    cert = db.get_cert_from_db(ipa_ca_nickname)
+    return unicode(DN(x509.load_certificate(cert).subject))
```
frasertweedale: OK.

ipaserver/install/ca.py
frasertweedale: Yep, that works fine.

ipaserver/install/ca.py
frasertweedale: Hmm, this gets looked up during `install_check`, before replica DS is created, so we would have to use a `remote_api`.  Now, what if the master is pre freeipa-4.2 thus lacks the `ca` plugin?

I don't think your suggested approach works in this case.  Shall we stay with reading the CA subject DN from the IPA CA cert, or do you have another suggestions?

install/tools/ipa-ca-install
jcholast: IMO the scope is not big (only 4 places including this one need an update), but OK.

ipaserver/install/ca.py
jcholast: Using `remote_api` here should be OK. If the master does not have `ca_show`, the default CA subject should be used (hence the try-except in the snippet).

ipaserver/install/ca.py
frasertweedale: Hm, actually this does not work fine :(

I cannot set the knob value directly because, in the replica_install case, the `ca_subject` and `subject_base` knobs have been replaced by a `removed` property.  (`ipalib/install/service.py:51`).

I cannot set `options._ca_subject` (or `._subject_base`) and then read it back later because the `options.__getattr__` raises `AttributeError` if the attr name does not correspond to a knob name (`ipapython/install/core.py:571`).

I find the installer/knob code quite hard to follow... do you have a suggestion?  (I certainly agree with your original point that it is preferable to compute the values once only, and cache them).

frasertweedale: (Previous comment is incorrect; please disregard)

HonzaCholasta:
* `--subject-base` and `--ca-subject` are not validated in `ipa-ca-install`.
* Please squash "{ds,ca}instance: rename 'subject' to 'ca_subject'" into "Allow full customisability of IPA CA subject DN".
* Please use the correct ticket URL in "Add sanity checks for use of --ca-subject and --subject-base".

frasertweedale:
@HonzaCholasta PR updated.  Re ticket URL, I think 2614 is the correct one for that commit.

HonzaCholasta:
@frasertweedale, the ticket *number* is correct, but the URL points to Dogtag Trac.

frasertweedale:
@HonzaCholasta whups!  Thanks for clarifying; fixed.

HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/324183cd63aeadbaa9678d610ba59e1295a606fe
https://fedorahosted.org/freeipa/changeset/db6674096c598918ea6b12ca33a96cf5e617a434
https://fedorahosted.org/freeipa/changeset/c6db493b06320455a2366945911939a605df2a73
https://fedorahosted.org/freeipa/changeset/6f3eb85c302f54bec561337e6627c89144b589ff
https://fedorahosted.org/freeipa/changeset/46bf0e89ae054b34adc66d08f205a5155e6f3fd6
https://fedorahosted.org/freeipa/changeset/f54df62abae4a15064bf297634558eb9be83ce33
https://fedorahosted.org/freeipa/changeset/09a65df6842411d42966111e50924df3de0b7031
https://fedorahosted.org/freeipa/changeset/3d01ec14c6e36fa962d0c54b2e08df0ecd401bd6
https://fedorahosted.org/freeipa/changeset/3f5660973251fe4b178e6486b6b86fbdd162d4d6
https://fedorahosted.org/freeipa/changeset/0c95a00147b1dd508736dacc847873ddddafb504

PR #246, opened 2016-11-16
Build: ignore rpmbuild for lint target

tomaskrizek:
Exclude rpmbuild from pylint checks when make lint is executed.
Clean up the current find expression.


mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/28c5e128c823f32787a4bde87e6b248928a2cd0a

PR #248, opened 2016-11-16
Fix the naming of ipa-dnskeysyncd service principal

martbab:
https://fedorahosted.org/freeipa/ticket/6405


mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/6ca96b3db03d4f3c5dbf465ca3d36bd563771c47

PR #251, opened 2016-11-17
Add rebuild rule for template files

tiran:
CONFIG_STATUS_DEPENDENCIES ensure that 'make' will re-run configure
after any of the template files (freeipa.spec.in, ipasetup.py.in...)
have been altered.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -30,6 +30,13 @@ EXTRA_DIST = .mailmap \
 	     pylintrc \
 	     pytest.ini
 
+# re-run configure after templates have been altered
+CONFIG_STATUS_DEPENDENCIES = \
+	     daemons/ipa-version.h.in \
+	     freeipa.spec.in \
+	     ipapython/version.py.in \
+	     ipasetup.py.in
+
 clean-local:
 	rm -rf "$(RPMBUILD)"
 	rm -rf "$(top_builddir)/dist"


pspacek:
NACK:
- the variable is defined incorrectly, it should be in configure.ac (see Automake manual chapter [16 Rebuilding Makefiles](https://www.gnu.org/software/automake/manual/html_node/Rebuilding.html))
- file `install/ui/src/libs/loader.js` is missing in the list
- please reference build system refactoring ticket in the commit message
- even more importantly, there is a conceptual problem:
I'm not even sure that this is the right approach. CONFIG_STATUS_DEPEDENCIES is intended for re-building Makefiles and related infrastructure, not individual files generated from templates. With this patch it will re-run configure and config.status for the whole tree. The correct way to do this is to re-run config.status only for particular file. (config.status is the script which is doing actual variable substitution into templates.)

Fixing this might require moving these particular template substitutions from `AC_CONFIG_FILES` to Makefile.am in particular directories.

tiran:
I created a ticket for you, https://fedorahosted.org/freeipa/ticket/6498 . 

This **exact (!)** AC_SUBST works for me. Additional newlines cause CONFIG_STATUS_DEPENDENCIES to be an empty rule.
```
# re-run configure after templates have been altered
AC_SUBST([CONFIG_STATUS_DEPENDENCIES],
     ['$(top_srcdir)/daemons/ipa-version.h.in $(top_srcdir)/freeipa.spec.in $(top_srcdir)/ipapython/version.py.in $(top_srcdir)/ipasetup.py.in $(top_srcdir)/install/ui/src/libs/loader.js.in'])
```

pspacek:
I'm going to provide a proper fix as agreed on meeting today.

PR #252, opened 2016-11-17
Use namespace-aware meta importer for ipaplatform

tiran:
Instead of symlinks and build-time configuration the ipaplatform module
is now able to auto-detect platforms on import time. The meta importer
uses the platform 'ID' from /etc/os-releases. It falls back to 'ID_LIKE'
on platforms like CentOS, which has ID=centos and ID_LIKE="rhel fedora".

The meta importer is able to handle namespace packages and the
ipaplatform package has been turned into a namespace package in order to
support external platform specifications.

https://fedorahosted.org/freeipa/ticket/6474

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,8 +83,3 @@ freeipa2-dev-doc
 
 /ipapython/version.py
 
-/ipaplatform/__init__.py
-/ipaplatform/constants.py
-/ipaplatform/paths.py
-/ipaplatform/services.py
-/ipaplatform/tasks.py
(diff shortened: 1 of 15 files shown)


martbab:
Is this PR valid given that we discussed to remove ipaplatform dependency from client-side modules?

stlaz:
It is not AFAIK. I noted that in https://fedorahosted.org/freeipa/ticket/6474 comment and there's also discussion about this in https://github.com/freeipa/freeipa/pull/271. Closing this PR.

PR #254, opened 2016-11-18
Replace LooseVersion with pkg_resource.parse_version

tiran:
pylint is having a hard time with distutils.version in tox's virtual
envs. virtualenv uses some tricks to provide a virtual distutils
package, pylint can't cope with.

https://github.com/PyCQA/pylint/issues/73 suggests to use pkg_resources
instead. pkg_resources' version parser has some more benefits, e.g. PEP
440 conformity.

### DIFF
--- a/ipaclient/remote_plugins/compat.py
+++ b/ipaclient/remote_plugins/compat.py
@@ -2,7 +2,6 @@
 # Copyright (C) 2016  FreeIPA Contributors see COPYING for license
 #
 
-from distutils.version import LooseVersion
 import importlib
 import os
 import re
@@ -12,6 +11,7 @@ import six
 
 from ipaclient.frontend import ClientCommand, ClientMethod
 from ipalib.frontend import Object
+from ipapython.ipautil import APIVersion
 
 if six.PY3:
     unicode = str
@@ -58,7 +58,7 @@ def get_package(server_info, client):
         server_info['version'] = server_version
         server_info.update_validity()
 
-    server_version = LooseVersion(server_version)
+    server_version = APIVersion(server_version)
 
     package_names = {}
     base_name = __name__.rpartition('.')[0]
@@ -66,15 +66,14 @@ def get_package(server_info, client):
     for name in os.listdir(base_dir):
         package_dir = os.path.join(base_dir, name)
         if name.startswith('2_') and os.path.isdir(package_dir):
-            package_version = name.replace('_', '.')
+            package_version = APIVersion(name.replace('_', '.'))
             package_names[package_version] = '{}.{}'.format(base_name, name)
 
     package_version = None
-    for version in sorted(package_names, key=LooseVersion):
-        if (package_version is None or
-                LooseVersion(package_version) < LooseVersion(version)):
+    for version in sorted(package_names):
+        if package_version is None or package_version < version:
             package_version = version
-        if LooseVersion(version) >= server_version:
+        if version >= server_version:
             break
 
     package_name = package_names[package_version]
--- a/ipalib/capabilities.py
+++ b/ipalib/capabilities.py
@@ -25,7 +25,7 @@ This module defines a dict, ``capabilities``, that maps feature names to API
 versions they were introduced in.
 """
 
-from distutils import version
+from ipapython.ipautil import APIVersion
 
 VERSION_WITHOUT_CAPABILITIES = u'2.51'
 
@@ -64,6 +64,6 @@ def client_has_capability(client_version, capability):
     :param client_version: The API version string reported by the client
     """
 
-    version_tuple = version.LooseVersion(client_version)
+    version = APIVersion(client_version)
 
-    return version_tuple >= version.LooseVersion(capabilities[capability])
+    return version >= APIVersion(capabilities[capability])
--- a/ipalib/frontend.py
+++ b/ipalib/frontend.py
@@ -20,12 +20,10 @@
 """
 Base classes for all front-end plugins.
 """
-
-from distutils import version
-
 import six
 
 from ipapython.version import API_VERSION
+from ipapython.ipautil import APIVersion
 from ipapython.ipa_log_manager import root_logger
 from ipalib.base import NameSpace
 from ipalib.plugable import Plugin, APINameSpace
@@ -770,16 +768,19 @@ class Command(HasParam):
         If the client minor version is less than or equal to the server
         then let the request proceed.
         """
-        server_ver = version.LooseVersion(API_VERSION)
-        ver = version.LooseVersion(client_version)
-        if len(ver.version) < 2:
-            raise VersionError(cver=ver.version, sver=server_ver.version, server= self.env.xmlrpc_uri)
-        client_major = ver.version[0]
+        server_apiver = APIVersion(self.api_version)
+        try:
+            client_apiver = APIVersion(client_version)
+        except ValueError:
+            raise VersionError(cver=client_version,
+                               sver=self.api_version,
+                               server=self.env.xmlrpc_uri)
 
-        server_major = server_ver.version[0]
-
-        if server_major != client_major:
-            raise VersionError(cver=client_version, sver=API_VERSION, server=self.env.xmlrpc_uri)
+        if (client_apiver.major != server_apiver.major
+                or client_apiver > server_apiver):
+            raise VersionError(cver=client_version,
+                               sver=self.api_version,
+                               server=self.env.xmlrpc_uri)
 
     def run(self, *args, **options):
         """
--- a/ipalib/plugable.py
+++ b/ipalib/plugable.py
@@ -24,8 +24,6 @@ The classes in this module make heavy use of Python container emulation. If
 you are unfamiliar with this Python feature, see
 http://docs.python.org/ref/sequence-types.html
 """
-
-from distutils.version import LooseVersion
 import operator
 import sys
 import threading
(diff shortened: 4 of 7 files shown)


tiran:
The PR is related to integration improvements and pip-installable effort, https://fedorahosted.org/freeipa/ticket/6468.

martbab:
I have a sneaky suspicion that the parse_version couldn't cope correctly with some of the downstream versioning schemes like RHEL z-stream releases and such.

That should be OK with API version, but client/server versions can pose a problem. There is are some examples in the ipatests/test_ipaserver/test_version_comparison.py I made when solving a bug that was caused by incorrect version comparison. Can you check if parse_version can handle them correctly?

tiran:
sigh, you are right, parse_version does not handle one version comparison as we expect: ```4.2.0-15.el7 < 4.2.0-15.el7_2.3``` 

```
from __future__ import print_function
import operator
import pkg_resources


version_strings = [
    ("3.0.0-1.el6", "3.0.0-2.el6", operator.lt),
    ("3.0.0-1.el6_8", "3.0.0-1.el6_8.1", operator.lt),
    ("3.0.0-42.el6", "3.0.0-1.el6", operator.gt),
    ("3.0.0-1.el6", "3.0.0-42.el6", operator.lt),
    ("3.0.0-42.el6", "3.3.3-1.fc20", operator.lt),
    ("4.2.0-15.el7", "4.2.0-15.el7_2.3", operator.lt),
    ("4.2.0-15.el7_2", "4.2.0-15.el7_2.3", operator.lt),
    ("4.2.0-15.el7_2.3", "4.2.0-15.el7_2.3", operator.eq),
    ("4.2.0-15.el7_2.3", "4.2.0-15.el7_2.2", operator.gt),
    ("4.2.0-1.fc23", "4.2.1-1.fc23", operator.lt),
    ("4.2.3-alpha1.fc23", "4.2.3-2.fc23", operator.lt),
    ("4.3.90.201601080923GIT55aeea7-0.fc23", "4.3.0-1.fc23", operator.gt)
]

for v1, v2, op in version_strings:
    v1 = pkg_resources.parse_version(v1)
    v2 = pkg_resources.parse_version(v2)
    if not op(v1, v2):
        print("failure: ", v1, op.__name__, v2)
```

```
$ python v.py 
failure:  4.2.0-15.el7 lt 4.2.0-15.el7_2.3
```

tiran:
In case you wonder what is going on, LooseVersion is both loose and dumb.

```
>>> ('el7',) < ('el7_2',)
True
```

The legacy version parser parses the version strings differently:
```
(-1, ('00000004', '00000002', '*final-', '00000015', '*el', '00000007', '*final'))
(-1, ('00000004', '00000002', '*final-', '00000015', '*el', '00000007', '*_', '00000002', '00000003', '*final'))
```

martbab:
Yes that's why we resorted to a direct CFFI call to RPM libs during server version check in upgrade. We simply could not win aside from re-implementing parsing of Z-stream versions etc. from scratch.

tiran:
I reworked my PR to use ```tasks.parse_ipa_version()``` for all version number checks. Of course I ran into the import cycle of death. I broke the cycle with a local import. It's not perfect but a completely fine solution. The function isn't called regularly.

martbab:
I was thinking about this a bit, and was wondering whether the platform-specific idiosyncracies of the version handling could be safely confined to the platform-specific code. I.E ipaplatform.base would define version comparisons via standard `pkg_resources` parser and fedora/redhat would override this with their platform-specific quirks.

The one thing that could be broken by this would be scenarios like Fedora clients talking to RHEL masters etc., but I think those sceniarios are not handled correctly by the current implementation anyway.

Another thing I was thinking about is whether we could use some proxy object in ipalib/ipaclient/ipapython libs which would use version comparison from ipaplatform if present, and if not use standard Python algorithms.

What I am aiming at that we should reduce the dependency of the PyPI candidate code on ipaplatform madness as much as is humanly possible. IMHO if we really want to use these modules in PyPI, then they ideally should not depend on ipaplatform at all.

martbab:
Upon closer inspection of the affected code it seems that all the code in ipalib/ipaclient/ipapython actually parses and compares API versions, which are sane and parseable by `pkg_resources`. Since client code should be concerned by API versions and not by platform-specific package versions (that is server's job) it seems that you do not even need to use ipaplatform hammer in this case.

tiran:
> I was thinking about this a bit, and was wondering whether the platform-specific idiosyncracies of the version handling could be safely confined to the platform-specific code. I.E ipaplatform.base would define version comparisons via standard pkg_resources parser and fedora/redhat would override this with their platform-specific quirks.

Somebody used a time machine and implemented your proposal already. ```tasks.parse_ipa_version()``` is a generic version parsing function. On RPM platforms it returns an object that uses ```librpm```.

IMHO it's ok to use ```tasks.parse_ipa_version()``` everywhere.

martbab:
Oh right, ok. My code reading skills are sub-par today.

mbasti-rh:
> > I was thinking about this a bit, and was wondering whether the platform-specific idiosyncracies of the version handling could be safely confined to the platform-specific code. I.E ipaplatform.base would define version comparisons via standard pkg_resources parser and fedora/redhat would override this with their platform-specific quirks.

> Somebody used a time machine and implemented your proposal already. tasks.parse_ipa_version() is a generic version parsing function. On RPM platforms it returns an object that uses librpm.
> 

Yeah, I did.


> IMHO it's ok to use tasks.parse_ipa_version() everywhere.

As Martin said, we have two versions: package released version and API version.

For released package version we need RPM/platform specific parser, but API version is just 2 numbers and standard python function can be used. It is the same accross platforms. 

API version is less important now, because we have versions of commads that scales better, and one day we may drop this overall API version completely. If you want to avoid importing platform then is fine to use standard python functions to compare API versions. Actually that rpmlib scares me.

mbasti-rh:
Thank you for PY3 fix, it actually belongs to this ticket https://fedorahosted.org/freeipa/ticket/6473

tiran:
Back to  ```parse_version```!

martbab:
It seems that your changes broke IPA upgrade:

```
Done configuring the web interface (httpd).
Applying LDAP updates
Upgrading IPA:
  [1/9]: stopping directory server
  [2/9]: saving configuration
  [3/9]: disabling listeners
  [4/9]: enabling DS global lock
  [5/9]: starting directory server
  [6/9]: upgrading server
ipa         : ERROR    Upgrade failed with 'SetuptoolsVersion' object has no attribute 'version'
  [error] RuntimeError: 'SetuptoolsVersion' object has no attribute 'version'
  [cleanup]: stopping directory server
  [cleanup]: restoring configuration
ipa.ipapython.install.cli.install_tool(CompatServerMasterInstall): ERROR    Update failed: 'SetuptoolsVersion' object has no attribute 'version'
ipa.ipapython.install.cli.install_tool(CompatServerMasterInstall): ERROR    The ipa-server-install command failed. See /var/log/ipaserver-install.log for more information
11-21 18:49 ipadocker.cli ERROR    Command ipa-server-install -U --domain ipa.test --realm IPA.TEST -p Secret123 -a Secret123 --setup-dns --auto-forwarders failed (exit code 1)
```
Traceback in ipaserver-install.log:

```
# tail -n 50  /var/log/ipaserver-install.log 
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 481, in __runner
    exc_handler(exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 510, in _handle_execute_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 500, in _handle_exception
    six.reraise(*exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 471, in __runner
    step()
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 468, in <lambda>
    six.reraise(*exc_info)
    value = gen.send(prev_value)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 705, in _configure
    next(executor)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 481, in __runner
    exc_handler(exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 510, in _handle_execute_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 568, in _handle_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 500, in _handle_exception
    six.reraise(*exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 565, in _handle_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 500, in _handle_exception
    six.reraise(*exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 471, in __runner
    step()
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 468, in <lambda>
    six.reraise(*exc_info)
    value = gen.send(prev_value)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/__init__.py", line 575, in main
    master_install(self)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/install.py", line 265, in decorated
    func(installer)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/install.py", line 851, in install
    ds.apply_updates()
  File "/usr/lib/python2.7/site-packages/ipaserver/install/dsinstance.py", line 693, in apply_updates
    raise RuntimeError("Update failed: %s" % e)

2016-11-21T17:49:45Z DEBUG The ipa-server-install command failed, exception: RuntimeError: Update failed: 'SetuptoolsVersion' object has no attribute 'version'
2016-11-21T17:49:45Z ERROR Update failed: 'SetuptoolsVersion' object has no attribute 'version'
2016-11-21T17:49:45Z ERROR The ipa-server-install command failed. See /var/log/ipaserver-install.log for more information
```

tiran:
setuptool's version parser does not support slicing. I need to find another solution for ```verify_client_version()```.

tiran:
@martbab more fun, the doc string of ```verify_client_version``` deviates from the actual implementation. The code does not implement the minor version check. It only compares major version numbers.

https://github.com/freeipa/freeipa/blob/master/ipalib/frontend.py#L764

martbab:
LGTM but please add the relevant ticket number into the commit message.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/2cbaf156045769b54150e4d4c3c1071f164a16fb

PR #257, opened 2016-11-18
Don't ship install subpackages with wheels

tiran:
The install subpackages of ipaclient, ipalib and ipapython contain
helper code for installers such as ipa-client-install. They also depend
on external modules that are not available on PyPI, e.g. SSSDConfig.
Since PyPI wheel packages do not support client installation, the
install subpackages contain dead and unsupported code.

The custom build_py plugin removes the subpackages from bdist_wheel
builds. It's not enough to just remove 'ipaclient.install' from the
'packages' list. Surplus files have to be removed from build/lib, too.

### DIFF
--- a/ipasetup.py.in
+++ b/ipasetup.py.in
@@ -14,8 +14,41 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
+from distutils import log
 import os
 import sys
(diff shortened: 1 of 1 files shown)


mbasti-rh:
ACK, please rebase

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/526bcea705d04895aa6b09bce996ac340783d1d0

PR #258, opened 2016-11-18
Break ipaplatform / ipalib import cycle of hell

tiran:
Here is an attempt to break the import cycle of hell between ipaplatform
and ipalib. All services now pass an ipalib.api object to
services.service(). RedHatServices.__init__() still needs to do a local
import because it initializes its wellknown service dict with service
instances.

### DIFF
--- a/ipaplatform/base/services.py
+++ b/ipaplatform/base/services.py
@@ -27,10 +27,10 @@ import os
 import json
 import time
 import collections
+import warnings
 
 import six
 
-import ipalib
 from ipapython import ipautil
 from ipaplatform.paths import paths
 
@@ -63,7 +63,6 @@ class KnownServices(collections.Mapping):
     instances as its own attributes on first access (or instance creation)
     and cache them.
     """
-
     def __init__(self, d):
         self.__d = d
 
@@ -93,9 +92,17 @@ class PlatformService(object):
 
     """
 
-    def __init__(self, service_name, api=ipalib.api):
+    def __init__(self, service_name, api=None):
         self.service_name = service_name
-        self.api = api
+        if api is not None:
+            self.api = api
+        else:
+            import ipalib  # FixMe: break import cycle
+            self.api = ipalib.api
+            warnings.warn(
+                "{s.__class__.__name__}('{s.service_name}', api=None) "
+                "is deprecated.".format(s=self),
+                RuntimeWarning, stacklevel=2)
 
     def start(self, instance_name="", capture_output=True, wait=True,
         update_service_list=True):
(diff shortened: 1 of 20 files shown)


ipaplatform/base/services.py
```diff
@@ -93,9 +92,17 @@ class PlatformService(object):
...
 
-    def __init__(self, service_name, api=ipalib.api):
+    def __init__(self, service_name, api=None):
         self.service_name = service_name
-        self.api = api
+        if api is not None:
+            self.api = api
+        else:
+            import ipalib  # FixMe: break import cycle
+            self.api = ipalib.api
```
tiran: This import is for backwards compatibility with external code. It can be removed in a future version of FreeIPA.

install/share/copy-schema-to-ca.py
```diff
@@ -104,7 +104,7 @@ def restart_pki_ds():
     """Restart the CA DS instance to pick up schema changes
     """
     root_logger.info('Restarting CA DS')
-    services.service('dirsrv').restart(SERVERID)
+    services.service('dirsrv', api).restart(SERVERID)
```
mbasti-rh: This is really special code, according docs, this code must be copied from RHEL7 to RHEL6 and then executed on RHEL6. S this change will break RHEL6. IMO this code should be set to stone as is :)

tiran:
thx @mbasti-rh . I took care of it.

ipaplatform/base/services.py
mbasti-rh: Maybe would be better to make this comment to code, I'm not sure if we should have `else` branch there. We never declared this as a public stable API.

mbasti-rh:
LGTM, except the inline comment I made,  I'll test it tomorrow

ipaplatform/base/services.py
tiran: The else block is required to keep b/w compatibility with scripts like ```copy-schema-to-ca.py```.

jcholast:
The original code is broken by design IMO. The API object is used only to get the configured service startup timeout and to guess our DS instance name. None of this is platform specific, so I would prefer if we removed this from `ipaplatform` altogether instead of "just" fixing the import issue.

Anyway, given that the current plan is to make `ipaclient` _not_ depend on `ipaplatform`, is this change still necessary?

tiran:
It's no longer a priority, but I still like to fix the imports No matter what the cyclic imports and cross-package dependencies are a mess. A wrong order of imports can trigger an import error. Even with this PR, ipapython.dogtag triggers an import of ipalib.api.

In the long run neither ipaplatform nor ipapython should have an import dependency on ipalib or a global ipalib.api object. Instead the API object should be passed.

mbasti-rh:
ACK from me if @jcholast is not against it

jcholast:
I'm OK with it.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/6409abf1a60f3548203e6607a2b157ff72af2c89

PR #260, opened 2016-11-21
Build: fix path in ipa-ods-exporter.socket unit file

pspacek:
This fixes regression caused by incorrect
daemons/dnssec/ipa-ods-exporter.socket.in path template introduced
in commit 312e780041fc9025ca3c189e6c9fcb54c7340714.

https://fedorahosted.org/freeipa/ticket/6495


mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/5862eaa1a0e3fbced79d6c209016c1138e692888

PR #261, opened 2016-11-21
upgrade: do not set HTTP and DS principals explicitly

martbab:
In ipa-server-upgrade, HTTP and DS principals are explicitly constructed from
realm, fqdn, and service prefix. This is no neither required nor desirable,
since the principal is now a read-only property which constructs the principal
name in the same way.

https://fedorahosted.org/freeipa/ticket/6500


martbab:
Closing this as duplicate of https://github.com/freeipa/freeipa/pull/262

PR #266, opened 2016-11-23
ipapython: simplify Env object initialization

jcholast:
Fully initialize Env objects in Env() instead of having to call their
private methods to complete the initialization later.

Do not use custom Env instance to determine the debug level to use for the
IPA API object - the IPA API object can properly determining the
configured debug level on its own.

Remove locking and related code from Env as it is never used.

https://fedorahosted.org/freeipa/ticket/6408


jcholast:
Only now I have noticed that this won't actually help fixing [ticket 6482](https://fedorahosted.org/freeipa/ticket/6482).

Nevermind this PR then.

pvoborni:
I don't understand the "Nevermind this PR then.".

fixing ticket 6482 is good but fixing ticket 6408 is required event more for  @tiran work, right?

stlaz:
From offline discussion I got that the PR should actually work in the end. I'll make the review.

jcholast:
Yes, my above comment is wrong (sorry).

stlaz:
This PR breaks almost all tests in test_ipalib/test_crud.py with `AttributeError: 'API' object has no attribute 'env'`. This error can be observed in some other tests:
http://pastebin.com/8EjE2QVS (please disregard the DNS tests failures).

martbab:
@stlaz @HonzaCholasta any progress on this PR or should we mark it as postponed and return to it later?

PR #267, opened 2016-11-23
ipa-replica-conncheck: do not close listening ports until required

tomaskrizek:
Previously, a separate thread would be created for each socket used
for conncheck. It would also time out after one second, after which it
would be closed and reopened again. This caused random failures of
conncheck.

Now all sockets are handled in a single thread and once the server
starts to listen on a port, it does not close that connection until the
script finishes.

Only IPv6 socket is used for simplicity, since it can handle both IPv6
and IPv4 connections. This requires IPv6 kernel support, which is
required by other parts of IPA anyway.

https://fedorahosted.org/freeipa/ticket/6487


install/tools/ipa-replica-conncheck
```diff
@@ -281,32 +277,86 @@ def configure_krb5_conf(realm, kdc, filename):
...
+                              dict(port=port, proto=proto))
+
+    def _bind_to_port(self, port, socket_type):
+        # Use IPv6 socket as it is able to accept both IPv6 and IPv4
+        # connections. Since IPv6 kernel module is required by other
+        # parts of IPA, it should always be available.
+        family = socket.AF_INET6
+
+        host = '::'   # all available interfaces
+        proto = 'tcp' if socket_type == socket.SOCK_STREAM else 'udp'
```
pspacek: What about `dict[proto]`? It will catch error cases where protocol is neither tcp nor udp for free...

install/tools/ipa-replica-conncheck
```diff
@@ -281,32 +277,86 @@ def configure_krb5_conf(realm, kdc, filename):
 
     krbconf.newConf(filename, opts)
 
+
 class PortResponder(threading.Thread):
 
-    def __init__(self, port, port_type, socket_timeout=1):
+    def __init__(self, ports):
```
pspacek: I would appreciate doc string saying what `ports` is expected to contain. Thanks!

install/tools/ipa-replica-conncheck
```diff
@@ -281,32 +277,86 @@ def configure_krb5_conf(realm, kdc, filename):
...
+        host = '::'   # all available interfaces
+        proto = 'tcp' if socket_type == socket.SOCK_STREAM else 'udp'
+
+        try:
+            sock = socket.socket(family, socket_type)
+
+            # Make sure IPv4 clients can connect to IPv6 socket
+            sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
+
+            if socket_type == socket.SOCK_STREAM:
```
pspacek: The code around is using either `if socket_type` or `if proto`. It would be good to use the same condition everywhere so it is easier to read.

install/tools/ipa-replica-conncheck
```diff
@@ -548,6 +597,7 @@ def main():
             time.sleep(3600)
             print_info("Connection check timeout: terminating listening program")
 
+
 if __name__ == "__main__":
     try:
         sys.exit(main())
```
pspacek: LOL, please fix this as well.

install/tools/ipa-replica-conncheck
```diff
@@ -281,32 +277,92 @@ def configure_krb5_conf(realm, kdc, filename):
...
+            root_logger.debug('%(port)d %(proto)s: Started listening' %
+                              dict(port=port, proto=proto))
+        except socket.error as e:
+            root_logger.warning('%(port)d %(proto)s: Failed to bind\n%(exc)s' %
+                                dict(port=port, proto=proto,
+                                     exc=traceback.format_exc(e)))
+        else:
+            self._sockets.append(sock)
+
+    def _respond(self, sock):
```
pspacek: Nitpick: Feel free to ignore this.
In general it would be better to split `_respond` into `_respond_tcp` and `_respond_udp`. The two ifs have basically nothing in common and it would remove one more nested if and could share logging (if logging is done in the calling function).

install/tools/ipa-replica-conncheck
```diff
@@ -342,7 +398,9 @@ def port_check(host, port_list):
         raise RuntimeError("Port check failed! Inaccessible port(s): %s" \
                 % ", ".join(msg_ports))
 
+
 def main():
+    global RESPONDER
```
pspacek: Nitpick: Feel free to ignore this.

Yuck. It would be better to split the huge `main` into smaller functions and pass `RESPONDER` between those. Anyway, this is mainly relict of the old code so I do not insist on getting rid of it.

install/tools/ipa-replica-conncheck
```diff
@@ -384,11 +442,8 @@ def main():
         # create listeners
         print_info("Start listening on required ports for remote master check")
 
-        for port in required_ports:
-            root_logger.debug("Start listening on port %d (%s)" % (port.port, port.description))
-            responder = PortResponder(port.port, port.port_type)
-            responder.start()
-            RESPONDERS.append(responder)
+        RESPONDER = PortResponder(required_ports)
```
pspacek: The design should have Implementation section mentioning why it is done in this particular way (i.e. explain why we have one thread vs. other options).

install/tools/ipa-replica-conncheck
```diff
@@ -548,18 +603,17 @@ def main():
...
 if __name__ == "__main__":
     try:
         sys.exit(main())
-    except SystemExit as e:
-        sys.exit(e)
     except KeyboardInterrupt:
         print_info("\nCleaning up...")
         sys.exit(1)
-    except RuntimeError as e:
-        sys.exit(e)
```
pspacek: Sorry, my bad, except RuntimeError must stay there. The program is using RuntimeError to pass error messages to the user.

install/tools/ipa-replica-conncheck
```diff
@@ -281,32 +277,92 @@ def configure_krb5_conf(realm, kdc, filename):
...
+
+            # Make sure IPv4 clients can connect to IPv6 socket
+            sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
+
+            if socket_type == socket.SOCK_STREAM:
+                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+
+            sock.bind((host, port))
+            if socket_type == socket.SOCK_STREAM:
+                sock.listen(1)
```
pspacek: Please add rationale for using value 1 to a comment/commit message/design.

tomaskrizek:
I've created a separate [ticket](https://fedorahosted.org/freeipa/ticket/6522) and PR #284 for the change discussed offline, since it seemed out of the scope for this ticket.

mbasti-rh:
needs rebase

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/af0ba661889c2e2c9a35d4cff9681c2abab73649

PR #268, opened 2016-11-23
Build system must regenerate file when template changes

pspacek:
Proper fix for https://fedorahosted.org/freeipa/ticket/6498.

This PR obsoletes #251.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -127,11 +127,11 @@ endif WITH_JSLINT
 lint: acilint apilint $(POLINT_TARGET) $(PYLINT_TARGET) $(JSLINT_TARGET)
 
 .PHONY: acilint
-acilint:
+acilint: $(top_builddir)/ipapython/version.py
 	cd $(srcdir); ./makeaci --validate
 
 .PHONY: apilint
-apilint:
+apilint: $(top_builddir)/ipapython/version.py
 	cd $(srcdir); ./makeapi --validate
 
 .PHONY: polint
@@ -142,8 +142,11 @@ polint:
 # folders rpmbuild, freeipa-* and dist. Skip (match, but don't print) .*,
 # *.in, *~. Finally print all python files, including scripts that do not
 # have python extension.
-.PHONY: pylint
-pylint:
+.PHONY: pylint $(top_builddir)/ipapython/version.py
+$(top_builddir)/ipapython/version.py:
+	(cd $(top_builddir)/ipapython && make version.py)
+
+pylint: $(top_builddir)/ipapython/version.py
 	FILES=`find $(top_srcdir) \
 		-type d -exec test -e '{}/__init__.py' \; -print -prune -o \
 		-path './rpmbuild' -prune -o \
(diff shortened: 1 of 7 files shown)


mbasti-rh:
works for me

tiran:
I don't like the approach and prefer ```AC_CONFIG_FILE``` over manual sed for templating. You only have to add a couple of rules like 

```
# Makefile.python.am
$(top_builddir)/ipasetup.py: $(top_builddir)/config.status $(top_builddir)/ipasetup.py.in
        $(MAKE) -C $(top_builddir) $(@F)
```

pspacek:
I already described problems with `AC_CONFIG_FILE` in https://github.com/freeipa/freeipa/pull/251#issuecomment-261470338 a week ago, including envisioned move from `AC_CONFIG_FILE` to `Makefile.am`.

Please propose a solution which does not have problems mentioned in https://github.com/freeipa/freeipa/pull/251#issuecomment-261470338 so we can consider it.


For the record, this sed replacement is nothing unusual. The sed replacement is what [Autoconf v2.69 manual chapter 4.8.2 Installation Directory Variables](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Installation-Directory-Variables.html) recommends and is already used all over the place in the build system (init directory, daemons/ipa-otpd, and elsewhere).

It is pitty that you did not comment on envisioned direction a week ago, nor a three days ago when first version of this PR was published.

tiran:
You gave a good reason to not use ```CONFIG_STATUS_DEPEDENCIES``` and I agree with your reasoning. I don't see a case against ```AC_CONFIG_FILE```, though. ```config.status``` substitution feature is more powerful than manual sed rules. I'm worried that we are going to run into problems in the future. It's surprising that some files can use all ```@VAR@``` substitutions and some only a limited subset.

Your patch already introduces proper dependencies for ```ipasetup.py``` and ```version.py```. Why not introduce a build rule for these files in ```Makefile.python.am```?

pspacek:
Oh, you are right, I was mixing `CONFIG_STATUS_DEPEDENCIES` and `AC_CONFIG_FILES`. Sorry!

So please let me explain the problem with `AC_CONFIG_FILES`:
`AC_CONFIG_FILES` properly substitutes variables only in Makefiles, as explained in [Autoconf v2.69 manual chapter 4.8.2 Installation Directory Variables](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Installation-Directory-Variables.html).

Yes, we can use `AC_CONFIG_FILES` so all variables can used for substitution, but in that case only subset of all usable variables will be substituted correctly. I do not think that it is right approach.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/ba6ae666acaf8b930d18f45efc7c9c9faad3526b
https://fedorahosted.org/freeipa/changeset/6857de02f3a9c2d7e99e33863be3c65f71fa0d58
https://fedorahosted.org/freeipa/changeset/89739a6c910461a3cac3abc1bf2ff162c7c5bc82
https://fedorahosted.org/freeipa/changeset/6fcfe689f47a02df023de69f62c889d9b4dc26fe
https://fedorahosted.org/freeipa/changeset/6aa360775a781bee5a2fdd884cbfa33b545fcbb4
https://fedorahosted.org/freeipa/changeset/a89f63c5a62c4a02fc248a095f539a099a9c28c5

PR #270, opened 2016-11-24
Test: uniqueness of certificate renewal master

ofayans:
https://fedorahosted.org/freeipa/ticket/6504

### DIFF
--- a/ipatests/test_integration/test_replica_promotion.py
+++ b/ipatests/test_integration/test_replica_promotion.py
@@ -441,3 +441,45 @@ class TestWrongClientDomain(IntegrationTest):
                                      raiseonerr=False)
         assert(result1.returncode == 0), (
             'Failed to promote the client installed with the upcase domain name')
+
+
+class TestRenewalMaster(IntegrationTest):
+
+    topology = 'star'
+    num_replicas = 1
+
+    @classmethod
+    def uninstall(cls, mh):
+        super(TestRenewalMaster, cls).uninstall(mh)
+
+    def test_replica_not_marked_as_renewal_master(self):
+        """
+        https://fedorahosted.org/freeipa/ticket/5902
+        """
+        master = self.master
+        replica = self.replicas[0]
+        result = master.run_command(["ipa", "config-show"]).stdout_text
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_renewal_master.py
```diff
@@ -0,0 +1,45 @@
...
+    @classmethod
+    def uninstall(cls, mh):
+        super(TestRenewalMaster, cls).uninstall(mh)
+
+    def test_replica_not_marked_as_renewal_master(self):
+        """
+        https://fedorahosted.org/freeipa/ticket/5902
+        """
+        master = self.master
+        suffix = ",".join(["dc=%s" % i for i in master.domain.name.split('.')])
```
martbab: You can call `ipa config-show` and search for the value of `IPA CA renewal master:` to get renewal master name.

ipatests/test_integration/test_renewal_master.py
```diff
@@ -0,0 +1,45 @@
...
+        assert("numEntries: 1" in result.stdout_text), (
+            "Unexpected number of renewal master entries in"
+            " the command stdout: %s" % result.stdout_text)
+        # Test that after master uninstallation, replica overtakes the cert
+        # renewal master role
+        tasks.uninstall_master(master)
+        result1 = self.replicas[0].run_command([
+            "ldapsearch", "-Y", "GSSAPI", "-b",
+            "cn=masters,cn=ipa,cn=etc,%s" % suffix,
+            "(ipaConfigString=caRenewalMaster)"
```
martbab: The same here, you just assert that the value now equals to replica instead of master.

martbab:
I have suggested some improvements in your code inline. Also, can we actually do this test as a part of other test suite or is it necessary to create a new one?

Also the commit message does not really state the purpose clearly.

flo-renaud:
Hi,
you may also want to perform the same test after changing the renewal master with _ipa config-mod --ca-renewal-master-server newrenewalmaster.example.com_.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/fad87a9962ee33cfebc4fa59aba589e98b076cea

PR #273, opened 2016-11-25
Build: workaround bug while calling parallel make from rpmbuild

pspacek:


martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/132b475c2586f3ced68724355e9c45722dccf604

PR #275, opened 2016-11-28
Enhance __repr__ method of Principal

martbab:
`__repr__` now returns more descriptive string containing the actual principal
name while keeping the ability to reconstruct the object from it.

This makes principal names visible in debug logs, easing troubleshooting a
bit.

https://fedorahosted.org/freeipa/ticket/6505

### DIFF
--- a/ipapython/kerberos.py
+++ b/ipapython/kerberos.py
@@ -181,3 +181,7 @@ class Principal(object):
             principal_string = u'@'.join([principal_string, realm])
 
         return principal_string
+
+    def __repr__(self):
+        return "{0.__module__}.{0.__name__}('{1}')".format(
+            self.__class__, self)
--- a/ipatests/test_ipapython/test_kerberos.py
+++ b/ipatests/test_ipapython/test_kerberos.py
@@ -82,6 +82,8 @@ def test_principals(valid_principal):
         assert getattr(princ, name) == value
 
     assert unicode(princ) == principal_name
+    assert repr(princ) == "ipapython.kerberos.Principal('{}')".format(
+        principal_name)
 
 
 def test_multiple_unescaped_ats_raise_error():


ipapython/kerberos.py
```diff
@@ -181,3 +181,7 @@ def __str__(self):
             principal_string = u'@'.join([principal_string, realm])
 
         return principal_string
+
+    def __repr__(self):
+        return "{}.{}('{}')".format(
+            self.__module__, self.__class__.__name__, self.__str__())
```
tiran: * does it actually work? Usually instances don't have a ```__module__``` attribute, only classes.
* don't call ```self.__str__()```
* where is the unit test? :)

Something like this should work
```
    def __repr__(self):
        return "{0.__module__}.{0.__name__}('{1}')".format(
            self.__class__, self)
```

martbab:
Sorry I somehow botched that, but it worked nevertheless. I have re-worked the PR according to your comments.

```
In [1]: import ipapython.kerberos
In [2]: p = ipapython.kerberos.Principal(u"HTTP/replica1.ipa.test")
In [3]: p
Out[3]: ipapython.kerberos.Principal('HTTP/replica1.ipa.test')
In [5]: r = eval('p')
In [6]: r
Out[6]: ipapython.kerberos.Principal('HTTP/replica1.ipa.test')
In [7]: r == p
Out[7]: True
```

tiran:
Can you please add a test to```ipatests/test_ipapython/test_kerberos.py``` ```test_principals```? Something along the line ```assert repr(princ) == "ipapython.kerberos.Principal('{}')".format(principal_name)``` should do the trick (untested).

martbab:
That sound like a good idea. Added such assert to the unit tests.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/38cc40ddb5bf965801500bb4f66fd965b12e3c88

PR #276, opened 2016-11-28
replica-conncheck: improve error msg + logging

tomaskrizek:
Replica conncheck may fail for other reasons then network
misconfiguration. For example, an incorrect admin password might be
provided. Since conncheck is ran as a separate script in quiet mode,
no insightful error message can be displayed.

User is instead pointed to the logs, which were also improved to contain
the usual on screen messages.

https://fedorahosted.org/freeipa/ticket/6497

### DIFF
--- a/install/tools/ipa-replica-conncheck
+++ b/install/tools/ipa-replica-conncheck
@@ -63,7 +63,7 @@ class SshExec(object):
     def __call__(self, command, verbose=False):
         # Bail if ssh is not installed
         if self.cmd is None:
-            print("WARNING: ssh not installed, skipping ssh test")
+            root_logger.warning("WARNING: ssh not installed, skipping ssh test")
             return ('', '', 0)
 
         tmpf = tempfile.NamedTemporaryFile()
@@ -110,11 +110,6 @@ BASE_PORTS = [
                 CheckedPort(80, SOCK_STREAM, "HTTP Server: Unsecure port"),
                 CheckedPort(443, SOCK_STREAM, "HTTP Server: Secure port"),
              ]
-
-
-def print_info(msg):
-    if not QUIET:
-        print(msg)
 
 
 def parse_options():
@@ -210,10 +205,6 @@ def parse_options():
     if not options.hostname:
         options.hostname = socket.getfqdn()
 
-    if options.quiet:
-        global QUIET
-        QUIET = True
-
     return safe_options, options
 
 
@@ -223,7 +214,8 @@ def logging_setup(options):
     if os.getegid() == 0 and options.log_to_file:
         log_file = paths.IPAREPLICA_CONNCHECK_LOG
 
-    standard_logging_setup(log_file, debug=options.debug)
+    standard_logging_setup(log_file, verbose=(not options.quiet),
+                           debug=options.debug, console_format='%(message)s')
 
 
 def sigterm_handler(signum, frame):
@@ -400,13 +392,14 @@ def port_check(host, port_list):
             else:
                 ports_failed.append(port)
                 result = "FAILED"
-        print_info("   %s (%d): %s" % (port.description, port.port, result))
+        root_logger.info("   %s (%d): %s" % (port.description, port.port, result))
 
     if ports_udp_warning:
-        print("The following UDP ports could not be verified as open: %s" \
-                % ", ".join(str(port.port) for port in ports_udp_warning))
-        print("This can happen if they are already bound to an application")
-        print("and ipa-replica-conncheck cannot attach own UDP responder.")
+        root_logger.warning(
+            ("The following UDP ports could not be verified as open: %s\n"
+             "This can happen if they are already bound to an application\n"
+             "and ipa-replica-conncheck cannot attach own UDP responder.")
+            % ", ".join(str(port.port) for port in ports_udp_warning))
 
     if ports_failed:
         msg_ports = []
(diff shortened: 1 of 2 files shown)


install/tools/ipa-replica-conncheck
```diff
@@ -362,29 +355,34 @@ def main():
...
+                         % options.master)
         tcp_ports = [ port for port in required_ports if port.port_type == SOCK_STREAM ]
         udp_ports = [ port for port in required_ports if port.port_type == SOCK_DGRAM ]
         port_check(options.master, tcp_ports)
 
         if udp_ports:
-            print_info("\nThe following list of ports use UDP protocol and would need to be")
-            print_info("checked manually:")
+            root_logger.info(("\nThe following list of ports use UDP protocol"
+                              "and would need to be"))
```
stlaz: I guess the double parentheses are just a typo, right?

install/tools/ipa-replica-conncheck
```diff
@@ -362,29 +355,34 @@ def main():
...
+            root_logger.info("checked manually:")
```
stlaz: Also, you can squash this into previous info() call.

install/tools/ipa-replica-conncheck
```diff
@@ -362,29 +355,34 @@ def main():
...
+                root_logger.info("   %s (%d): %s"
+                                 % (port.description, port.port, result))
 
-        print_info("\nConnection from replica to master is OK.")
+        root_logger.info("\nConnection from replica to master is OK.")
 
         # create listeners
-        print_info("Start listening on required ports for remote master check")
+        root_logger.info(("Start listening on required ports for remote "
+                         "master check"))
```
stlaz: Double parentheses again? Is there really some reason to it?

install/tools/ipa-replica-conncheck
```diff
@@ -509,57 +507,65 @@ def main():
...
                 if result.returncode != 0:
-                    print('Could not SSH into remote host. Error output:')
-                    for line in result.error_output.splitlines():
-                        print('    %s' % line)
-                    raise RuntimeError('Could not SSH to remote host.')
+                    root_logger.debug(result.error_output)
+                    raise RuntimeError((
+                        'Could not SSH to remote host.\n'
+                        'See /var/log/ipareplica-conncheck.log for more '
+                        'information.'))
```
stlaz: Here as well.

install/tools/ipa-replica-conncheck
```diff
@@ -509,57 +507,65 @@ def main():
...
                         " ".join(remote_check_opts))
                 returncode = result.returncode
                 stderr = result.error_output
-                print_info(result.output)
+                root_logger.info(result.output)
             if returncode != 0:
-                raise RuntimeError("Remote master check failed with following error message(s):\n%s" % stderr)
+                raise RuntimeError(
+                    ("Remote master check failed with following "
+                     "error message(s):\n%s" % stderr))
```
stlaz: ^^

install/tools/ipa-replica-conncheck
```diff
@@ -509,57 +507,65 @@ def main():
...
-            print_info("Listeners are started. Use CTRL+C to terminate the listening part after the test.")
-            print_info("")
-            print_info("Please run the following command on remote master:")
-
-            print_info("/usr/sbin/ipa-replica-conncheck " + " ".join(remote_check_opts))
+            root_logger.info(("Listeners are started. Use CTRL+C to terminate "
+                              "the listening part after the test."))
+            root_logger.info("")
+            root_logger.info(("Please run the following command on "
+                              "remote master:"))
```
stlaz: ^^

install/tools/ipa-replica-conncheck
tomaskrizek: To join the multiline string.

tomaskrizek: Which are not needed when the string is an argument for a function call :) I'll remove them.

install/tools/ipa-replica-conncheck
```diff
@@ -509,57 +507,64 @@ def main():
...
-            print_info("")
-            print_info("Please run the following command on remote master:")
+            root_logger.info(
+                "Listeners are started. Use CTRL+C to terminate the listening "
+                "part after the test.\n\n"
+                "Please run the following command on remote master:")
 
-            print_info("/usr/sbin/ipa-replica-conncheck " + " ".join(remote_check_opts))
+            root_logger.info("/usr/sbin/ipa-replica-conncheck " +
+                             " ".join(remote_check_opts))
```
stlaz: This could be merged into the previous message.

install/tools/ipa-replica-conncheck
```diff
@@ -328,13 +320,14 @@ def port_check(host, port_list):
...
     if ports_udp_warning:
-        print("The following UDP ports could not be verified as open: %s" \
-                % ", ".join(str(port.port) for port in ports_udp_warning))
-        print("This can happen if they are already bound to an application")
-        print("and ipa-replica-conncheck cannot attach own UDP responder.")
+        root_logger.warning(
+            ("The following UDP ports could not be verified as open: %s\n"
+             "This can happen if they are already bound to an application\n"
+             "and ipa-replica-conncheck cannot attach own UDP responder.")
+            % ", ".join(str(port.port) for port in ports_udp_warning))
```
stlaz: NOTE: using str.format() would help you remove the double parentheses here as well but you don't have to do it as the old '%' way is used in most of the other messages here.

install/tools/ipa-replica-conncheck
```diff
@@ -509,57 +507,64 @@ def main():
...
 
-                print_info("Check SSH connection to remote master")
+                root_logger.info("Check SSH connection to remote master")
                 result = ssh('echo OK', verbose=True)
                 if result.returncode != 0:
-                    print('Could not SSH into remote host. Error output:')
-                    for line in result.error_output.splitlines():
-                        print('    %s' % line)
-                    raise RuntimeError('Could not SSH to remote host.')
+                    root_logger.debug(result.error_output)
```
stlaz: Is there a reason not to print the output to stdout as it used to be?

install/tools/ipa-replica-conncheck
stlaz: Also, use str.format() here instead of '+'.

install/tools/ipa-replica-conncheck
```diff
@@ -509,57 +507,64 @@ def main():
...
         sys.exit(main())
     except SystemExit as e:
         sys.exit(e)
     except KeyboardInterrupt:
-        print_info("\nCleaning up...")
+        root_logger.info("\nCleaning up...")
         sys.exit(1)
     except RuntimeError as e:
-        sys.exit(e)
+        root_logger.error('ERROR: ' + str(e))
```
stlaz: Don't use '+' in strings, use str.format().

install/tools/ipa-replica-conncheck
tomaskrizek: It is purely a debug information (~50-100 lines), so I don't see the reason why it should be displayed to the user. Since previously the function lacked any reasonable logging, I think it was decided to simply print this debug info out, as the rest of the messages.

Since this patch adds logging to the function, I this the proper place for this information is the log. If the command fails, the user is prompted to check the log where he can find this info.

stlaz: Right, that's the explanation I was looking for, thanks.

install/tools/ipa-replica-conncheck
tomaskrizek: Fiixed

install/tools/ipa-replica-conncheck
tomaskrizek: Fixed

stlaz:
Needs rebase.

install/tools/ipa-replica-conncheck
```diff
@@ -585,56 +583,62 @@ def main():
...
 if __name__ == "__main__":
     try:
         sys.exit(main())
     except KeyboardInterrupt:
-        print_info("\nCleaning up...")
+        root_logger.info("\nCleaning up...")
         sys.exit(1)
     except RuntimeError as e:
-        sys.exit(e)
+        root_logger.error('ERROR: {ex}'.format(ex=e))
```
stlaz: NOTE: The 'ERROR' prefix was not here before, however it increases awareness that there was an error so I'm fine with keeping it.

stlaz:
Seems to work fine, ACK.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/eb6905bbb45d246f9beddeea69da2236cf67bc68
https://fedorahosted.org/freeipa/changeset/de981d348efed6dc58b2e355e65244853f06ebc1

PR #278, opened 2016-11-28
Restore the original functionality of `env` and `plugins` commands

martbab:
This reverts commit 1166fbc4946596fcc2ed51a1ec6990fc7dae8964 "Add 'ipa
localenv' subcommand" and instead fixes the command to be executed locally
unless `--server` option is given.

Note than `plugins` command fails due to
https://fedorahosted.org/freeipa/ticket/6513 but now at least it fails either
locally or on server-side :).

https://fedorahosted.org/freeipa/ticket/6490

### DIFF
--- a/ipalib/misc.py
+++ b/ipalib/misc.py
@@ -1,0 +1,124 @@
+#
+# Copyright (C) 2016  FreeIPA Contributors see COPYING for license
+#
+
+import re
+from ipalib import LocalOrRemote, _, ngettext
+from ipalib.output import Output, summary
+from ipalib import Flag
+from ipalib.plugable import Registry
+
+register = Registry()
+
+# FIXME: We should not let env return anything in_server
+# when mode == 'production'.  This would allow an attacker to see the
+# configuration of the server, potentially revealing compromising
+# information.  However, it's damn handy for testing/debugging.
+
+
+class env(LocalOrRemote):
+    __doc__ = _('Show environment variables.')
+
+    msg_summary = _('%(count)d variables')
+
+    takes_args = (
+        'variables*',
+    )
+
+    takes_options = LocalOrRemote.takes_options + (
+        Flag('all',
+            cli_name='all',
+            doc=_('retrieve and print all attributes from the server. Affects command output.'),
+            exclude='webui',
+            flags=['no_option', 'no_output'],
+            default=True,
+        ),
+    )
+
+    has_output = (
+        Output('result',
+            type=dict,
+            doc=_('Dictionary mapping variable name to value'),
+        ),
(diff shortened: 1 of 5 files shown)


ipalib/misc.py
```diff
@@ -0,0 +1,124 @@
...
+        result = {}
+        for namespace in self.api:
+            for plugin in self.api[namespace]():
+                cls = type(plugin)
+                key = '{}.{}'.format(cls.__module__, cls.__name__)
+                result.setdefault(key, []).append(namespace)
+
+        return dict(
+            result=result,
+        )
```
stlaz: I think this code should rather be in ipalib.plugins

ipalib/misc.py
```diff
@@ -26,26 +26,31 @@ class env(LocalOrRemote):
     )
 
     takes_options = LocalOrRemote.takes_options + (
-        Flag('all',
+        Flag(
```
stlaz: Wouldn't it be better to go like this?
```python
Flag('all',
     cli_name='all',
     ...
     )
```
Otherwise it looks a bit ridiculous as you're saving about 1-2 spaces here :)

martbab: If you look at some other newly written plugins (e.g. `ipaserver/plugins/server.py`) you will se that I have used the option formatting constistent with them and I would rather keep it  like that.

ipalib/misc.py
martbab: Why should it be in ipalib.plugins? can you explain your reasons better?

ipalib/misc.py
stlaz: Oh, alright.

ipalib/misc.py
stlaz: It was originally a plugin and still keeps plugin-like behavior so I thought it should rather be there but there's no other reason.

martbab: Well since these particular classes must behave as both server and client plugins at the same time, you would then have this code copy-pasted in two places. I rather opted to keep implementation in one place and a) registering it on server, and b) overriding it on client.

stlaz:
`env` works as expected, `plugins` seems to fail as expected. ACK.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/42307ae2dc4d9be06d3fd9b7e5af2e0761b8d6ad
https://fedorahosted.org/freeipa/changeset/0ae7bebb7690620b69fd9c84b7a385c2086c1ee1
https://fedorahosted.org/freeipa/changeset/64a4be26febf19e427760115912a858a804208a0

PR #281, opened 2016-11-28
Accept server host names resolvable only using /etc/hosts

pspacek:
Apparently "files" implementation of hosts NSS database cannot deal with
trailing period in host names.

Previously name server.example.com which is was resolvable neither using
dns nor myhostname NSS modules were rejected by installer
(despite having matching line in /etc/hosts).

These names which are resolvable purely using "files" database are now
accepted.

The problem is that I had to remove trailing period from names passed
to getaddrinfo() function. This effectivelly enables search list processing.
This means that items from the search list might be silently appended to
the query and we might get an IP address for totally different names
than we asked for.

Unfortunatelly I see no way around this while keeping ability
to use names from NSS hosts database.

https://fedorahosted.org/freeipa/ticket/6518

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -57,7 +57,6 @@ from ipaserver.install import certs, service, sysupgrade
 from ipaplatform import services
 from ipaplatform.paths import paths
 from ipaplatform.tasks import tasks
-from ipapython import dnsutil
 
 if six.PY3:
     unicode = str
@@ -474,9 +473,9 @@ def resolve_ip_addresses_nss(fqdn):
     :returns:
         list of IP addresses as UnsafeIPAddress objects
     """
-    # make sure the name is fully qualified
-    # so search path from resolv.conf does not apply
-    fqdn = str(dnsutil.DNSName(fqdn).make_absolute())
+    # it would be good disable search list processing from resolv.conf
+    # to avoid cases where we get IP address for an totally different name
+    # but there is no way to do this using getaddrinfo parameters
     try:
         addrinfos = socket.getaddrinfo(fqdn, None,
                                        socket.AF_UNSPEC, socket.SOCK_STREAM)


martbab:
So can you imagine some scenario where this behavior may cause issues? Some exotic DNS setup maybe?

pspacek:
This entierly depens on configuration. Imagine following imaginary company setup:
- public part of DNS tree is `example.com.`
- private part of DNS tree is `corp.`
- resolv.conf contains `corp` in search list

Now an admin is going to install IPA instance for publicly available services at server `srv1.ipa.example.com.`.  The name `srv1.ipa.example.com.` is not resolvable as --setup-dns option is used. Now, the `dns` module invoked by NSS will try to lookup `srv1.ipa.example.com.`. It might (depending on configuration) fallback to `srv1.ipa.example.com.corp.` which may accidentally exist (as an IPA server for company internal purposes).

This is purely hypotetical, I'm just trying to show that the code is subtly broken.

martbab:
I see. I guess we can live with the fact that we may break such eccentric DNS topologies.
I think we cannot really handle all the corner cases associated with guessing/setting hostname by ourselves anyway (yes I am not a big fan of FreeIPA stepping onto provisioning system's toes).

martbab:
Thinking of this some more, shouldn't be `--no-host-dns` option used and advertised if you want to set unresolvable hostname during install?

pspacek:
`--no-host-dns` disables all checks (theoretically) so it should be used only in special cases. Given it acts as kind of force switch, we should not advertise it. In either case the user will have to provide `--ip-address` option. Also, the user is asked for IP address in interactive mode so IMHO we are sufficiently covered.

martbab:
Ok I am fine with this.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/0e093f938d8126f11fed920b7381ba6e3d07da5b
ipa-4-4:
https://fedorahosted.org/freeipa/changeset/47ee2870d83eeb9b07137c765d3feb41da8b02c7

PR #284, opened 2016-11-29
ipautil: check for open ports on all resolved IPs

tomaskrizek:
When a hostname is provided to host_port_open, it should check if
ports are open for ALL IPs that are resolved from the hostname, instead
of checking whether the port is reachable on at least one of the IPs.

https://fedorahosted.org/freeipa/ticket/6522

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -55,6 +55,12 @@ GEN_PWD_LEN = 22
 GEN_TMP_PWD_LEN = 12  # only for OTP password that is manually retyped by user
 
 
+PROTOCOL_NAMES = {
+    socket.SOCK_STREAM: 'tcp',
+    socket.SOCK_DGRAM: 'udp'
+}
+
+
 class UnsafeIPAddress(netaddr.IPAddress):
     """Any valid IP address with or without netmask."""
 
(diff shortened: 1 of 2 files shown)


ipapython/ipautil.py
```diff
@@ -884,15 +900,26 @@ def host_port_open(host, port, socket_type=socket.SOCK_STREAM, socket_timeout=No
...
+
+                # Do not log udp failures as errors (to be consistent with
+                # the rest of the code that checks for open ports)
+                if socket_type == socket.SOCK_DGRAM:
+                    root_logger.debug(msg)
+                else:
+                    root_logger.error(msg)
         finally:
             if s:
                 s.close()
```
pspacek: The socket is not closed properly, I would say. `s.close()` is outside of the cycle but the cycle can open many sockets...

tomaskrizek: The `finally` block is inside the cycle, so all opened sockets should be closed properly.

pspacek: You are right, I'm blind.

ipapython/ipautil.py
```diff
@@ -884,15 +896,27 @@ def host_port_open(host, port, socket_type=socket.SOCK_STREAM, socket_timeout=No
...
+            if log_errors:
+                msg = ('Failed to connect to port %(port)d %(proto)s on '
+                       '%(addr)s' % dict(port=port,
+                                         proto=PROTOCOL_NAMES[socket_type],
+                                         addr=sa[0]))
+
+                # Do not log udp failures as errors (to be consistent with
+                # the rest of the code that checks for open ports)
+                if socket_type == socket.SOCK_DGRAM:
+                    root_logger.debug(msg)
```
pspacek: I would like to see this on WARNING level so it is consistent with messages printed by conncheck (and actually visible to user).

ipapython/ipautil.py
```diff
@@ -884,15 +896,27 @@ def host_port_open(host, port, socket_type=socket.SOCK_STREAM, socket_timeout=No
...
+            if log_errors:
+                msg = ('Failed to connect to port %(port)d %(proto)s on '
+                       '%(addr)s' % dict(port=port,
+                                         proto=PROTOCOL_NAMES[socket_type],
+                                         addr=sa[0]))
+
+                # Do not log udp failures as errors (to be consistent with
+                # the rest of the code that checks for open ports)
+                if socket_type == socket.SOCK_DGRAM:
+                    root_logger.warning(msg)
```
pspacek: I just noticed that FreeIPA logging is so botched that it does not show WARNING messages by default. Please leave the code as it is and I will ack it.

mbasti-rh:
needs rebase

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/a24cd01304aaef77b66d0e178585c9ec8bbce9b5

PR #289, opened 2016-11-30
Require python-gssapi >= 1.2.0

tiran:
The PyPI package for python-gssapi 1.1.x has a packaging bug. It depends on
enum34 for Python 3 although it is only required for 2.7. 1.2.0 is the
oldest version that has been tested at length by QE. It's know to work.

Bump up in freeipa.spec is not required for technical reasons. The
packaging bug only affects PyPI packages. It's policy to keep
requirements in sync.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -115,7 +115,7 @@ BuildRequires:  samba-python
 BuildRequires:  python-setuptools
 # 0.6: serialization.load_pem_private_key, load_pem_public_key
 BuildRequires:  python-cryptography >= 0.6
-BuildRequires:  python-gssapi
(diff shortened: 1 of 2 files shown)


martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/8559791e0d520f4a3503e35d1975ac31448b1390

frozencemetery:
We (the python-gssapi team) do not believe that is correct.  This problem with enum34 is fixed in the latest 1.1.z release (1.1.4).

We also do have CI that runs on every commit, so every released version should be stable, though 1.2.0 is also a great version.

PR #290, opened 2016-11-30
Require python-cryptography >= 1.3.1

tiran:
python-cryptography versions < 1.3 no longer compile with recent OpenSSL
1.0.2 versions. In order to build wheels, a more recent version of
cryptography is required. 1.3.1 is the oldest well tested version (RHEL
7.3) that is known to work with FreeIPA.

Bump up in freeipa.spec is not required for technical reasons. The
problem only affects PyPI packages. It's policy to keep
requirements in sync.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -113,8 +113,8 @@ BuildRequires:  python-cffi
 %if 0%{?with_lint}
 BuildRequires:  samba-python
 BuildRequires:  python-setuptools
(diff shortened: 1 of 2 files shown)


martbab:
Please rebase the PR so we can do clean merge, it should be simple conflict resolution.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/289982e02fa6bef700fe2c1900ddbed864876faa

PR #291, opened 2016-11-30
replica install: track the RA agent certificate again

jcholast:
During the rebase of commit 822e1bc82af3a6c1556546c4fbe96eeafad45762 on top
of commit 808b1436b4158cb6f926ac2b5bd0979df6ea7e9f, the call to track the
RA agent certificate with certmonger was accidentally removed from
ipa-replica-install.

Put the call back so that the certificate is tracked after replica install.


jcholast:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/4221266562778806f02748fee2dfbd814261f2b4

PR #292, opened 2016-11-30
Increase the timeout waiting for certificate issuance in installer

flo-renaud:
During the server installation, the installer requests certificates
through certmonger. The current timeout is 60s and is too low.
Increase this timeout to 120s (the code checks the status each 5s
up to the timeout value).

https://fedorahosted.org/freeipa/ticket/6433

### DIFF
--- a/ipalib/install/certmonger.py
+++ b/ipalib/install/certmonger.py
@@ -30,6 +30,7 @@ import dbus
 import shlex
 import subprocess
 import tempfile
+from ipalib import api
 from ipapython.ipa_log_manager import root_logger
 from ipaplatform.paths import paths
 from ipaplatform import services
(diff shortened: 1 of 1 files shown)


martbab:
@jcholast would it make sense to move this value into `ipalib.constants` and make it configurable?

mbasti-rh:
+1

flo-renaud:
@martbab @mbasti-rh: I checked the code and some parts already use api.env.startup_timeout for certmonger requests (in ipa_certupdate.py or ipa_cacert_manage.py for instance). Is it OK for you if I replace my hardcoded value with api.env.startup_timeout?

martbab:
IMHO anything is better than hardcoded values so you have my blessing.

ipalib/install/certmonger.py
```diff
@@ -309,7 +310,8 @@ def request_and_wait_for_cert(
     reqId = request_cert(nssdb, nickname, subject, principal,
                          passwd_fname, dns, ca, profile,
                          pre_command, post_command)
-    state = wait_for_request(reqId, timeout=60)
+    timeout = api.env.startup_timeout + 60
```
mbasti-rh: I still see hardcoded value (60)

AFAIK startup_timeout is 5 minutes yb default, so there is no need for additional 60 seconds

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/9e3c17c6ded868b4261aa76137c703a4fb866578

PR #293, opened 2016-12-01
Run out-of-tree tests in Travis CI

martbab:
This PR implements out-of-tree tests in Travis CI using
https://github.com/martbab/ipa-docker-test-runner

There is couple of issues with this PR we may discuss:

* Currently ipa-docker-test-runner is pulled and installed directly from git
  master branch.
  It will probably be better if I release some stable version to PyPI so that
  changes in master do not break CI. This would also allow us to cache the
  package and its dependencies between builds, saving some time

* We have currently two concurrent jobs that run most tests split evenly
  between them. This cuts the time from 42 or so minutes to around half an
  hour. Any ideas to further speed the tests are welcome.

* DNS tests are flaky in Travis CI. Some of them are ignored because they tend
  to fail non-deterministically. I will try to determin the root causes and
  work around them.

### DIFF
--- a/.test_runner_config.yaml
+++ b/.test_runner_config.yaml
@@ -1,0 +1,50 @@
+container:
+  detach: true
+  hostname: master.ipa.test
+  working_dir: /freeipa
+host:
+  binds:
+  - /sys/fs/cgroup:/sys/fs/cgroup:ro
+  - /dev/urandom:/dev/random:ro
+  privileged: true
+  security_opt:
+  - label:disable
+  tmpfs:
+  - /tmp
+  - /run
+server:
+  domain: ipa.test
+  password: Secret123
+  realm: IPA.TEST
+steps:
+  build:
+  - make V=0 ${make_target}
+  builddep:
+  - rm -rf /var/cache/dnf/*
+  - "dnf makecache fast || :"
+  - dnf builddep -y ${builddep_opts} --spec freeipa.spec.in --best --allowerasing
+  cleanup:
+  - chown -R ${uid}:${gid} ${container_working_dir}
+  configure:
+  - ./autogen.sh
+  install_packages:
+  - dnf install -y ${container_working_dir}/dist/rpms/*.rpm --best --allowerasing
+  install_server:
+  - ipa-server-install -U --domain ${server_domain} --realm ${server_realm} -p ${server_password}
+    -a ${server_password} --setup-dns --auto-forwarders
+  - ipa-kra-install -p ${server_password}
+  lint:
+  - make V=0 lint
+  prepare_tests:
+  - echo ${server_password} | kinit admin && ipa ping
+  - cp -r /etc/ipa/* ~/.ipa/
+  - echo ${server_password} > ~/.ipa/.dmpw
+  - echo 'wait_for_dns=5' >> ~/.ipa/default.conf
+  run_tests:
+  - ipa-run-tests ${tests_ignore} -k-test_dns_soa ${tests_verbose} ${path}
+tests:
+  ignore:
+  - test_integration
+  - test_webui
+  - test_ipapython/test_keyring.py
+  verbose: true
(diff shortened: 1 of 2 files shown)


.test_runner_config.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - /run
+server:
+  domain: ipa.test
+  password: Secret123
+  realm: IPA.TEST
+steps:
+  build:
+  - make V=0 ${make_target}
+  builddep:
+  - rm -rf /var/cache/dnf/*
```
tiran: dnf clean all?

.test_runner_config.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - "dnf makecache fast || :"
+  - dnf builddep -y ${builddep_opts} --spec freeipa.spec.in --best --allowerasing
+  cleanup:
+  - chown -R ${uid}:${gid} ${container_working_dir}
+  configure:
+  - autoreconf -i && ./configure
```
tiran: ./autogen.sh

.test_runner_config.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - dnf install -y ${container_working_dir}/dist/rpms/*.rpm --best --allowerasing
+  install_server:
+  - ipa-server-install -U --domain ${server_domain} --realm ${server_realm} -p ${server_password}
+    -a ${server_password} --setup-dns --auto-forwarders
+  - ipa-kra-install -p ${server_password}
+  lint:
+  - make V=0 lint
+  prepare_tests:
+  - echo ${server_password} | kinit admin && ipa ping
+  - cp -r /etc/ipa/* ~/.ipa/.
```
tiran: What's with the extra dot at the end of the line?

.travis.yml
```diff
@@ -1,16 +1,39 @@
...
+            TESTS_TO_RUN="test_cmdline
+            test_install
+            test_ipalib
+            test_ipapython
+            test_ipaserver
+            test_pkcs10
+            test_xmlrpc/test_[l-z]*.py"
 before_install:
     - pip install pep8
+    - pip3 install --user git+https://github.com/martbab/ipa-docker-test-runner
```
tiran: In Travis you can safely install to the global site-packages directory. The image gets destroyed after each run anyway.

.test_runner_config.yaml
martbab: rm -rf /var/cache/dnf/* && dnf makecache fast are a workaround around deadlocks/synchronization issues observed when you run DNF too fast :). I will open a ticket against DNF for this.

tiran: Computers ... :( I don't want to know how much time got wasted to find the bug.

.travis.yml
```diff
@@ -1,16 +1,44 @@
...
+            test_ipalib
+            test_ipapython
+            test_ipaserver
+            test_pkcs10
+            test_xmlrpc/test_[l-z]*.py"
 before_install:
     - pip install pep8
+    - >
+      pip3 install
+      git+https://github.com/martbab/ipa-docker-test-runner@release-0-2-0
```
mbasti-rh: We moved this under FreeIPA organization on github, this link should be updated

stlaz:
Since I recently run into issues with ipa-server-install and low entropy somewhere around creation of kdb proxy which drastically increased install time, would it make sense to install haveged atop of our rpms to possibly mitigate the problem?

martbab:
We bind mount (non-blocking) /dev/urandom as /dev/random inside the container so we should actually never get stuck on low entropy in there.

stlaz:
Good. I see the tests pass now and both @tiran's nitpicks and @mbasti-rh's comment have been resolved, so an ACK is in order.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/5ecaea6bc4f49c2665597ca38fc52f4fae8a9d24
https://fedorahosted.org/freeipa/changeset/6d6fbc010ec2b607a11e0ff69c8cbdcd3c1d47d9

PR #300, opened 2016-12-02
WebUI: Add support for custom table pagination size

pvomacka:
New customization button opens dialog with field for setting the number of lines
in tables. After saving the new value there is new topic which starts refreshing
current table facet (if shown) and set all other facets expired. Therefore all
tables are immediately regenerated.

https://fedorahosted.org/freeipa/ticket/5742

### DIFF
--- a/install/ui/src/freeipa/facet.js
+++ b/install/ui/src/freeipa/facet.js
@@ -25,12 +25,14 @@ define([
         'dojo/_base/declare',
         'dojo/_base/lang',
         'dojo/dom-construct',
+        'dojo/topic',
         'dojo/on',
         'dojo/Stateful',
         'dojo/Evented',
         './_base/Singleton_registry',
         './_base/construct',
         './builder',
+        './config',
         './ipa',
         './jquery',
         './navigation',
@@ -43,8 +45,8 @@ define([
         './dialog',
         './field',
         './widget'
-       ], function(declare, lang, construct, on, Stateful, Evented,
-                   Singleton_registry, construct_utils, builder, IPA, $,
+    ], function(declare, lang, construct, topic, on, Stateful, Evented,
+                   Singleton_registry, construct_utils, builder, config, IPA, $,
                    navigation, phases, reg, rpc, su, text, ActionDropdownWidget) {
 
 /**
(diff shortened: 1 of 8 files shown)


install/ui/src/freeipa/facet.js
```diff
@@ -2359,6 +2361,14 @@ exp.table_facet = IPA.table_facet = function(spec, no_init) {
             selectable: that.selectable && !that.read_only
         });
 
+        topic.subscribe("change-pagination", function() {
+            that.table.page_length = config.get('table_page_size');
+
+            if (that.is_shown) that.refresh();
+
+            that.set_expired_flag();
```
pvoborni: Should the expiration flag be set first and then issuing the refresh?

install/ui/src/freeipa/field.js
```diff
@@ -962,6 +962,39 @@ field.validator = IPA.validator = function(spec) {
 };
 
 /**
+ * Javascript integer validator
+ *
+ * It allows to insert only integer numbers which can be safely represented by
+ * Javascript.
+ *
+ * @class
+ * @alternateClassName IPA.metadata_validator
```
pvoborni: alternate name is IPA.integer_validator

ipaserver/plugins/internal.py
```diff
@@ -250,6 +250,10 @@ class i18n_messages(Command):
             "update": _("Update"),
             "view": _("View"),
         },
+        "customization": {
+            "customization": _("Customization"),
+            "table_pagination": _("Table Pagination Size"),
```
pvoborni: What about using only "Pagination size". "Table" won't tell much to users. I.e. "what table?"

pvomacka:
@pvoborni Thank you for review. Proposed changes fixed.

martbab:
master:

* 7b699105a52d4d8c26a73044ba182d752b4a9833 Add javascript integer validator
* f78cc8932626de667c6e3a4461141a10a5d9c2e6 Make singleton from config module
* e1dfc51e48050ac1ad431d56003dc26e17ca653e Add support for custom table pagination size

PR #280, opened 2016-11-28
Set explicit confdir option for global contexts

tiran:
Some API contexts are used to modify global state (e.g. files in /etc
and /var). These contexts do not support confdir overrides. Initialize
the API with an explicit confdir argument to paths.ETC_IPA.

The special contexts are:

* backup
* cli_installer
* installer
* ipctl
* renew
* restore
* server
* updates

The patch also corrects the context of the ipa-httpd-kdcproxy script to
'server'.

https://fedorahosted.org/freeipa/ticket/6389

Signed-off-by: Christian Heimes <cheimes@redhat.com>

PR #301, opened 2016-12-02
scripts, tests: explicitly set confdir in the rest of server code

HonzaCholasta:
Commit 1e6a204b4372bbbfb722a00370a5ce4e34406b9f added explicit confdir
setting to api.bootstrap() calls of a randomly selected portion of
server-side scripts and tests. This commit adds it to the rest of
server-side code for consistency.

https://fedorahosted.org/freeipa/ticket/6389

### DIFF
--- a/doc/guide/wsgi.py.txt
+++ b/doc/guide/wsgi.py.txt
@@ -1,3 +1,4 @@
+from ipaplatform.paths import paths
 from ipalib import api
 from ipalib.config import Env
 from ipalib.constants import DEFAULT_CONFIG
@@ -6,11 +7,12 @@ from ipalib.constants import DEFAULT_CONFIG
 # by reading in the configuration file(s). The server always reads
 # default.conf and will also read in `context'.conf.
 env = Env()
-env._bootstrap(context='server', log=None)
+env._bootstrap(context='server', log=None, confdir=paths.ETC_IPA)
 env._finalize_core(**dict(DEFAULT_CONFIG))
 
 # Initialize the API with the proper debug level
-api.bootstrap(context='server', debug=env.debug, log=None) (ref:wsgi-app-bootstrap)
+api.bootstrap(context='server', confdir=paths.ETC_IPA,
+              debug=env.debug, log=None) (ref:wsgi-app-bootstrap)
 try:
     api.finalize() (ref:wsgi-app-finalize)
 except Exception as e:
--- a/install/tools/ipa-compat-manage
+++ b/install/tools/ipa-compat-manage
@@ -100,7 +100,10 @@ def main():
         if dirman_password is None:
             sys.exit("Directory Manager password required")
 
-    api.bootstrap(context='cli', in_server=True, debug=options.debug)
+    api.bootstrap(context='cli',
+                  in_server=True,
+                  debug=options.debug,
+                  confdir=paths.ETC_IPA)
     api.finalize()
     api.Backend.ldap2.connect()
 
--- a/install/tools/ipa-csreplica-manage
+++ b/install/tools/ipa-csreplica-manage
@@ -413,7 +413,11 @@ def main():
         api_env['log'] = None # turn off logging for non-root
 
     api.bootstrap(
-        context='cli', in_server=True, verbose=options.verbose, **api_env
+        context='cli',
+        in_server=True,
+        verbose=options.verbose,
+        confdir=paths.ETC_IPA,
+        **api_env
     )
     api.finalize()
 
--- a/install/tools/ipa-managed-entries
+++ b/install/tools/ipa-managed-entries
@@ -24,6 +24,7 @@ import re
 import sys
 from optparse import OptionParser  # pylint: disable=deprecated-module
 
+from ipaplatform.paths import paths
 from ipapython import config
 from ipaserver.install import installutils
 from ipalib import api, errors
@@ -72,7 +73,7 @@ def main():
         sys.exit("Unrecognized action [" + args[0] + "]")
     standard_logging_setup(None, debug=options.debug)
 
-    api.bootstrap(context='cli', debug=options.debug)
+    api.bootstrap(context='cli', debug=options.debug, confdir=paths.ETC_IPA)
     api.finalize()
     api.Backend.ldap2.connect(bind_pw=options.dirman_password)
 
--- a/ipaserver/advise/base.py
+++ b/ipaserver/advise/base.py
@@ -25,6 +25,7 @@ from textwrap import wrap
 from ipalib import api
 from ipalib.plugable import Plugin, API
 from ipalib.errors import ValidationError
+from ipaplatform.paths import paths
 from ipapython import admintool
 from ipapython.ipa_log_manager import log_mgr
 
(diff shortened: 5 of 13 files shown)


install/tools/ipa-managed-entries
```diff
@@ -72,7 +73,7 @@ def main():
         sys.exit("Unrecognized action [" + args[0] + "]")
     standard_logging_setup(None, debug=options.debug)
 
-    api.bootstrap(context='cli', debug=options.debug)
+    api.bootstrap(context='cli', debug=options.debug, confdir=paths.ETC_IPA)
```
stlaz: This is an ugly old script, therefore this should be in try-except block.

tiran:
I'll review the patch by the end of the week. Some changes are not required.

install/tools/ipa-managed-entries
stlaz: Nvm, the script behaves the same in other error cases, disregard the previous comment.

stlaz:
@tiran I find all the changes actually required. I think ACK is in order unless you spell out those which you think are not necessary.

doc/guide/wsgi.py.txt
```diff
@@ -6,11 +7,12 @@ from ipalib.constants import DEFAULT_CONFIG
...
 # default.conf and will also read in `context'.conf.
 env = Env()
-env._bootstrap(context='server', log=None)
+env._bootstrap(context='server', log=None, confdir=paths.ETC_IPA)
 env._finalize_core(**dict(DEFAULT_CONFIG))
 
 # Initialize the API with the proper debug level
-api.bootstrap(context='server', debug=env.debug, log=None) (ref:wsgi-app-bootstrap)
+api.bootstrap(context='server', confdir=paths.ETC_IPA,
+              debug=env.debug, log=None) (ref:wsgi-app-bootstrap)
```
tiran: NACK

Documentation should explain the common case (no confdir override). Users may want to use the IPA_CONFDIR var here. I'm ok with a comment or second example.

install/tools/ipa-compat-manage
```diff
@@ -100,7 +100,10 @@ def main():
         if dirman_password is None:
             sys.exit("Directory Manager password required")
 
-    api.bootstrap(context='cli', in_server=True, debug=options.debug)
+    api.bootstrap(context='cli',
+                  in_server=True,
+                  debug=options.debug,
+                  confdir=paths.ETC_IPA)
```
tiran: NACK, the script works fine with IPA_CONFDIR. It does neither modify local file nor requires root permission to run.

install/tools/ipa-csreplica-manage
```diff
@@ -413,7 +413,11 @@ def main():
         api_env['log'] = None # turn off logging for non-root
 
     api.bootstrap(
-        context='cli', in_server=True, verbose=options.verbose, **api_env
+        context='cli',
+        in_server=True,
+        verbose=options.verbose,
+        confdir=paths.ETC_IPA,
```
tiran: NACK

Same reason as for ```ipa-compat-manage```

install/tools/ipa-managed-entries
tiran: NACK

Same reason as for ```ipa-compat-manage```

ipaserver/advise/base.py
```diff
@@ -235,9 +236,13 @@ def print_advice(self, keyword):
...
 
-        api.bootstrap(in_server=False, context='cli')
+        api.bootstrap(in_server=False,
+                      context='cli',
+                      confdir=paths.ETC_IPA)
         api.finalize()
-        advise_api.bootstrap(in_server=False, context='cli')
+        advise_api.bootstrap(in_server=False,
+                             context='cli',
+                             confdir=paths.ETC_IPA)
```
tiran: NACK

Same reason as for ```ipa-compat-manage```

ipaserver/dnssec/ldapkeydb.py
```diff
@@ -414,7 +415,8 @@ def zone_keypairs(self):
     log = ipa_log_manager.root_logger
 
     # IPA framework initialization
-    ipalib.api.bootstrap(in_server=True, log=None)  # no logging to file
+    # no logging to file
+    ipalib.api.bootstrap(in_server=True, log=None, confdir=paths.ETC_IPA)
```
tiran: NACK

Same reason as for ```ipa-compat-manage```

This is just some debugging code in ```if __name__ == '__main__'``` block. There is no harm to support IPA_CONFDIR

ipatests/test_ipaserver/test_ldap.py
```diff
@@ -111,7 +112,7 @@ def test_Backend(self):
         # a client-only api. Then we register in the commands and objects
         # we need for the test.
         myapi = create_api(mode=None)
-        myapi.bootstrap(context='cli', in_server=True)
+        myapi.bootstrap(context='cli', in_server=True, confdir=paths.ETC_IPA)
```
tiran: NACK

Tests use their own confdir for testing. Let's handle the env var in the test runner.

ipatests/test_ipaserver/test_serverroles.py
```diff
@@ -469,7 +470,9 @@ def teardown_data(self):
 @pytest.fixture(scope='module')
 def mock_api(request):
     test_api = create_api(mode=None)
-    test_api.bootstrap(in_server=True, ldap_uri=api.env.ldap_uri)
+    test_api.bootstrap(in_server=True,
+                       ldap_uri=api.env.ldap_uri,
+                       confdir=paths.ETC_IPA)
```
tiran: NACK

same reason as above

install/tools/ipa-compat-manage
stlaz: You're still required to enter Directory Manager password which is why I would expect the script not to be influenced by an environment variable.

jcholast: It does not matter at all that the script works with IPA_CONFDIR - there is no use case for this and it only introduces additional maintenance burden for us.

doc/guide/wsgi.py.txt
jcholast: This file should be kept in sync with `install/share/wsgi.py`. This change does just that, there is no other intention behind it.

install/tools/ipa-compat-manage
tiran: I don't agree. There is no additonal maintenance burden.

HonzaCholasta: Yes, there is - we will have to test this and make sure we don't break it, which is not worth doing for a "feature" nobody wants and will ever use.

stlaz: +1

stlaz:
This PR implements the stuff that was agreed on in later comments in https://github.com/freeipa/freeipa/pull/280 and actually requested by @pvoborni.
Currently, I do not see the reason why this PR should not be accepted, if IPA_CONFDIR is required in either of these scripts for some reason, we can implement it later and add the justification to it.

HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/fe6f2b6f6effcf9f3c58e1e3f6d0874609c10c25

tiran:
My philosophy is: _Don't fix it it it ain't broken._

PR #303, opened 2016-12-02
Add python-pyasn1-modules into dependencies

pvomacka:
Python-pyasn1-modules is required by python-ldap package, but it would be
good to not rely on another package and rather say explicitely, that
this package is necessary.

https://fedorahosted.org/freeipa/ticket/6398


tiran:
Good idea, please add BuildRequires for the package, too.

pvomacka:
Added, I also added more information into commit message.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/a8b7dbff8ac660a28faf7ef43c7a0952171423b8

PR #304, opened 2016-12-02
Relax check for .git to support freeipa in submodules

tiran:
Let's relax the check for .git from directory to exists in order to
support freeipa in a git submodule. Submodules have a .git file with
content like

    gitdir: ../.git/modules/freeipa


martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/cac0c2d951e10d49372a038c73f796dc3beb62b9

PR #307, opened 2016-12-05
Lowered the version of gettext

pvomacka:
The lower version is needed while building on RHEL.
Also regenerated Rules-quot file.


pspacek:
If the file `Rules-quot` is generated by `autoreconf -i`, please remove it completely and add it into `.gitignore`.

pspacek:
@pvomacka Pavel, you did not remove the `po/Rules-quot` file. Adding it to `.gitignore` is not enough.

NACK (sorry for messing with the label, too fat fingers)

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/e5686c0ccb4a71fa0ab460fd45280d9adba91cca

PR #302, opened 2016-12-02
Use env var IPA_CONFDIR to get confdir for 'cli' context 

pvoborni:
This is continuation of PR #182 . It contains patch from #182 and additional one which addresses the issue raised in review.

PR #308, opened 2016-12-05
Add 'env_confdir' to constants

martbab:
Env confdir is always populated so it should be listed among variables
set during a call to `Env._bootstrap()`.

https://fedorahosted.org/freeipa/ticket/6389


tiran:
```env_confdir ``` was added in PR #302. This PR is required to make test passing.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/1300381d45a23073261e62cb031cf4285a34f641

PR #309, opened 2016-12-05
ipa-replica-conncheck: fix race condition

tomaskrizek:
When the thread that opens ports would execute notify() before the
original thread could call wait(), the original thread would wait
indefinitely for a notify() call.

https://fedorahosted.org/freeipa/ticket/6487


mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/a44974cdf8a6c53f2e86dac92aa579ba45f666e4

PR #310, opened 2016-12-06
WIP: CLI testing

mirielka:
Here is basic part of CLI testing for you to take a look at and provide feedback, before it's all done and polished up.

How it works: so far it's only tuned to run stageuser tests, so use `./make-tests ipatests/test_xmlrpc/test_stageuser_plugin.py --cli` to run in CLI mode, or of course without the --cli option to run original API tests. Note that last three tests are expected to fail in CLI mode as group tracker has not been modified for CLI testing yet.

What gets changed: 
- '--cli' option is added
- base tracker: changes that should ensure the basic functionality to run existing tests in CLI mode, this includes way how to execute commands and mapping of defined API options and output values to the CLI style (the idea of the mapping was to convert the text output from CLI to API style output so that more extensive methods to compare CLI output with expected values do not have to be created)
- specific trackers: changes are needed because CLI calls have different output than API calls (some messages are different, some items of API output are not present in CLI). So far I've modified stageuser tracker fully (i.e. that one should be final) and user tracker partly (just so that stageuser tests that use user tracker work properly, but user tests will not work in CLI mode yet).
- xmlrpc_test.py: unfortunately I failed to find a way how to raise the exception from subprocess so that it would fit the exception comparison using raises_exact method, so I chose to raise ExecutionError when it occurs in subprocess and compare just the error message in stderr. Since I expect the CLI tests to be run along with API tests, any change in type of raised error should be cought by API tests.

So far I observed:
- positives: no tests need to be changed, all the necessary changes are in trackers (which was the intention)
- negatives: only works for tracker based tests, i.e. not for declarative tests or tests that use just api.Command style; tests run longer that API tests

So please, if you have any comment and suggestions, I'll be glad to hear them so that I can polish this and finish the modifications.

### DIFF
--- a/ipatests/test_xmlrpc/tracker/base.py
+++ b/ipatests/test_xmlrpc/tracker/base.py
@@ -14,6 +14,12 @@ from ipapython.dn import DN
 from ipapython.version import API_VERSION
 from ipatests.util import Fuzzy
 
+import subprocess
+import os
+import re
+import pytest
+from ipalib.cli import to_cli
+
 
 class Tracker(object):
     """Wraps and tracks modifications to a plugin LDAP entry object
@@ -70,6 +76,22 @@ class Tracker(object):
     create_keys = None
     update_keys = None
 
+    # Mapping of API style options and output to CLI style options and output
+    mapping_options = {
+        'addattr': 'addattr',
+        'all': 'all',
+        'setattr': 'setattr',
+        'raw': 'raw',
+        'version': 'version',
+        }
+    mapping_output = {}
+
+    # Indicator for skipping errors that are not handled during CLI testing
+    skip_error = False
+
+    # List of options that do not take value
+    novalue = ['all', 'raw']
+
     _override_me_msg = "This method needs to be overridden in a subclass"
 
     def __init__(self, default_version=None):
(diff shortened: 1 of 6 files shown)


ipatests/test_xmlrpc/tracker/base.py
```diff
@@ -111,26 +133,105 @@ def filter_attrs(self, keys):
             raise RuntimeError('The tracker instance has no attributes.')
         return {k: v for k, v in self.attrs.items() if k in keys}
 
+    def cli_command(self, name, *args, **options):
+        """
+        Prepare CLI command for testing
+        """
+        cmd = "ipa {}".format(to_cli(name))
+        for item in args:
+            cmd += " {}".format(item)
```
mbasti-rh: did you mean?
```
cmd = cmd + ' '.join(args)
```

ipatests/test_xmlrpc/tracker/base.py
```diff
@@ -111,26 +133,105 @@ def filter_attrs(self, keys):
...
+        for key in options:
+            if key in self.novalue and options[key]:
+                cmd += " --{0}".format(self.mapping_options[key])
+            elif key in self.novalue and not options[key]:
+                continue
+            else:
+                if type(options[key]) is unicode and ' ' in options[key]:
```
mbasti-rh: you cannot use `is unicode` it is not in Py3, so in Py3 is should be

```
if six.PY3:
    unicode = str
```

mbasti-rh: or even better please use `isinstance()` instead of `type()`

ipatests/test_xmlrpc/tracker/base.py
mirielka: Ah, silly me, I forgot how to python. Will implement both your suggestions, thank you!

pvoborni:
Marking as postponed. We cannot expect the changes to be addressed by @mirielka any time soon. And CLI testing might  need design discussion.

pvoborni:
This PR is pretty outdated and no development is happening. Therefore closing but also adding to: https://www.freeipa.org/page/Postponed

PR #311, opened 2016-12-06
bindinstance: use LDAP to determine configuration status

martbab:
Instead of checking sysrestore status which leads to incorrect
evaluation of DNS configuration status during 4.2 -> 4.4 upgrade, use
server roles API to query DNS server role status on the master.

Use the default implementation only if LDAP connection is unavailable or
stale.

https://fedorahosted.org/freeipa/ticket/6503


martbab:
OK I will then close this PR and create a separate one against ipa-4-4

PR #312, opened 2016-12-06
bindinstance: use data in named.conf to determine configuration status

martbab:
Instead of checking sysrestore status which leads to incorrect
evaluation of DNS configuration status during 4.2 -> 4.4 upgrade, look
into named.conf to see whther it was already modified by IPA installer.

https://fedorahosted.org/freeipa/ticket/6503


mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/f0e09c42b76f229486e5dea097cd2b6602999943
ipa-4-4:
https://fedorahosted.org/freeipa/changeset/bf28d79afeff4575adc9ba0618b5acbf0cf51009

PR #313, opened 2016-12-06
ipaclient.plugins: Use api_version from internally called commands

dkupka:
https://fedorahosted.org/freeipa/ticket/6539

### DIFF
--- a/ipaclient/plugins/automount.py
+++ b/ipaclient/plugins/automount.py
@@ -60,6 +60,10 @@ class automountlocation_tofiles(MethodOverride):
 
     NO_CLI = classproperty(__NO_CLI_getter)
 
+    @property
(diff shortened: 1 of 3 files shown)


mbasti-rh:
Can you please provide better commit description in commit message? What, why and how it fixes? Think about our future us when we will be doing git archaeology

mbasti-rh:
Works for me

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/d841a79dc104521f736469eff7154c2f4266082b
ipa-4-4:
https://fedorahosted.org/freeipa/changeset/6ef666ed12fd73026f0f1d49faba152ae27d6082

PR #318, opened 2016-12-07
server install: fix external CA install

jcholast:
Replace the dual definitions of domain_name, dm_password and admin_password
knobs in server install with single definitions using the original names
without the 'new_' prefix.

This fixes the options read from the installer option cache in step 2 of
external CA install to use the correct knob names.


flo-renaud:
Works as expected.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/4fff09978eab520d130d87c0112b5caac907e651

PR #315, opened 2016-12-06
[ipa-4-4] gracefully handle setting replica bind dn group on old masters

martbab:
https://fedorahosted.org/freeipa/ticket/6532

This PR is for ipa-4-4 branch only. I will prepare separate PR for master since
the replication code was changed quite a b it during installer refactoring.

PR #319, opened 2016-12-07
[master] gracefully handle setting replica bind dn group on old masters

martbab:
This is the version of https://github.com/freeipa/freeipa/pull/315 for master
branch only

https://fedorahosted.org/freeipa/ticket/6532

### DIFF
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -436,6 +436,34 @@ class ReplicationManager(object):
         return DN(('cn', 'replica'), ('cn', self.db_suffix),
                   ('cn', 'mapping tree'), ('cn', 'config'))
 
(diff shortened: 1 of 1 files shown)


martbab:
PR #322 was closed, it was agreed that this fix is better.

martbab:
@flo-renaud I was able to reproduce the issue you observed when trying to test the patch. I have opened an upstream ticket for it https://fedorahosted.org/freeipa/ticket/6549. I think it is a regression caused by installer refactoring and is unrelated to this PR.

flo-renaud:
Hi,
thanks for the patch. It works as expected.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/95e602598a481f9c4a3b69ce8a861bf3816aa8ba

PR #320, opened 2016-12-07
add missing attribute to ipaca replica during CA topology update

martbab:
The previous fix for missing 'nsds5replicabinddngroupcheckinterval' fails when
the first CA master is being set up. The attribute addition from update file
has to be moved to the update plugin with a proper logic that determines the
presence of o=ipaca replica entry.

https://fedorahosted.org/freeipa/ticket/6508

### DIFF
--- a/install/share/ca-topology.uldif
+++ b/install/share/ca-topology.uldif
@@ -12,4 +12,3 @@ default: cn: ca
 
 dn: cn=replica,cn=o\3Dipaca,cn=mapping tree,cn=config
 onlyifexist: nsds5replicabinddngroup: cn=replication managers,cn=sysaccounts,cn=etc,$SUFFIX
-add: nsds5replicabinddngroupcheckinterval: 60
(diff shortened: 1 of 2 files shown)


mbasti-rh:
IMO #322 this might be related, @martbab can you please check it?

martbab:
@tbordaz mentioned that this fix is better and that we should discard https://github.com/freeipa/freeipa/pull/322. Please continue review.

martbab:
@mbasti-rh also this PR solves a bit different issue that the others (https://github.com/freeipa/freeipa/pull/315 and https://github.com/freeipa/freeipa/pull/319) and that is that the upgrade path for CA replicas was broken. It supplements the two PR so that everything is set up correctly also for o=ipaca suffix, not only the domain one.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/6086a6dbad21d93ed584d508f9844d73f64a4542
https://fedorahosted.org/freeipa/changeset/6d0e450c8226a8e23d88cf21487a77db66a2968b
ipa-4-4:
https://fedorahosted.org/freeipa/changeset/0ae9cd75cd8495a85c78f9781c7fab721b464230
https://fedorahosted.org/freeipa/changeset/b0acb23ff378d3c4fd8b54123696a8b48e6cd472

PR #321, opened 2016-12-08
certdb: fix PKCS#12 import with empty password

jcholast:
Since commit f919ab4ee0ec26d77ee6978e75de5daba4073402, a temporary file is
used to give passwords to pk12util. When a password is empty, the temporary
will be empty as well, which pk12util does not like.

Add new line after the password in the temporary file to please pk12util.

https://fedorahosted.org/freeipa/ticket/6541


dkupka:
Works for me.

jcholast:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/a35f5181c0af459e9f054838805628f31316cbe9

PR #324, opened 2016-12-09
Check for conflict entries before raising domain level

tbordaz:
Checking of conflicts is not only done in topology container as
tests showed it can occurs elsewhere

https://fedorahosted.org/freeipa/ticket/6534

### DIFF
--- a/ipaserver/plugins/domainlevel.py
+++ b/ipaserver/plugins/domainlevel.py
@@ -48,6 +48,30 @@ def get_domainlevel_range(master_entry):
         return DomainLevelRange(0, 0)
 
 
+def check_conflict_entries(ldap, api, desired_value):
+    """
+    Check if conflict entries exist in topology subtree
+    """
+
+    container_dn = DN(
+        ('cn', 'ipa'),
+        ('cn', 'etc'),
+        api.env.basedn
+    )
+    conflict="(nsds5replconflict=*)"
+    subentry="(|(objectclass=ldapsubentry)(objectclass=*))"
+    try:
+        ldap.get_entries(
+            filter="(& %s %s)" % (conflict, subentry),
+            base_dn=container_dn,
+            scope=ldap.SCOPE_SUBTREE)
+        message = _("Domain Level cannot be raised to {0}, "
+                    "existing replication conflicts have to be resolved."
+                    .format(desired_value))
+        raise errors.InvalidDomainLevelError(reason=message)
+    except errors.NotFound:
+        pass
+
 def get_master_entries(ldap, api):
     """
     Returns list of LDAPEntries representing IPA masters.
@@ -130,6 +154,10 @@ class domainlevel_set(Command):
                             "does not support it."
                             .format(desired_value, master['cn'][0]))
                 raise errors.InvalidDomainLevelError(reason=message)
+
+        # Check if conflict entries exist in topology subtree
+        # should be resolved first
+        check_conflict_entries(ldap, self.api, desired_value)
 
         current_entry.single_value['ipaDomainLevel'] = desired_value
         ldap.update_entry(current_entry)


ipaserver/plugins/domainlevel.py
```diff
@@ -48,6 +48,30 @@ def get_domainlevel_range(master_entry):
...
+        api.env.basedn
+    )
+
+    try:
+        conflicts, _dummy = ldap.find_entries(
+            filter="(&(nsds5replconflict=*)(|(objectclass=ldapsubentry)(objectclass=*)))",
+            base_dn=container_dn,
+            scope=ldap.SCOPE_SUBTREE)
+    except errors.EmptyResult:
+        conflicts = None
```
martbab: I would rewrite the body of this function like this:

```
+    try:
+        ldap.get_entries(
+            filter="(&(nsds5replconflict=*)(|(objectclass=ldapsubentry)(objectclass=*)))",
+            base_dn=container_dn,
+            scope=ldap.SCOPE_SUBTREE)
+            return True
+    except errors.NotFound:
+        return False
```

`get_entries` raises NotFound if the result is empty. This makes the code little more pythonic IMHO.

ipaserver/plugins/domainlevel.py
```diff
@@ -48,6 +48,30 @@ def get_domainlevel_range(master_entry):
...
+
+    if conflicts:
+        return True
+    else:
+        return False
+
```
martbab: I would rather rewrite the code to raise exception directly from this function:

```
+    try:
+        ldap.get_entries(
+            filter="(&(nsds5replconflict=*)(|(objectclass=ldapsubentry)(objectclass=*)))",
+            base_dn=container_dn,
+            scope=ldap.SCOPE_SUBTREE)
+
+        raise errors.InvalidDomainLevelError(...)
+    except errors.NotFound:
+        return
```

I use get_entreis here because it raises NotFound when no conflicts are found.

ipaserver/plugins/domainlevel.py
```diff
@@ -131,6 +155,13 @@ def execute(self, *args, **options):
                             .format(desired_value, master['cn'][0]))
                 raise errors.InvalidDomainLevelError(reason=message)
 
+        # Check if conflict entries exist in topology subtree, should be resolved first
+        if check_conflict_entries(ldap, self.api):
```
martbab: You can then just plainly call `check_conflict_entries(ldap, desired_value, self.api)` here.

martbab:
I have some comments regarding the logic of the check. These will also fix the warning reported by pylint checker.

ipaserver/plugins/domainlevel.py
```diff
@@ -48,6 +48,26 @@ def get_domainlevel_range(master_entry):
...
+        ('cn', 'etc'),
+        api.env.basedn
+    )
+
+    try:
+        ldap.get_entries(
+            filter="(&(nsds5replconflict=*)(|(objectclass=ldapsubentry)(objectclass=*)))",
+            base_dn=container_dn,
+            scope=ldap.SCOPE_SUBTREE)
+        return True
```
martbab: Please raise the InvalidDomainLevel here directly, there is no point in returning a boolean and then checking it in the caller to raise exception.

This leads to cleaner code uncluttered with extraneous boolean checks.

martbab:
I am still not happy with the the PR, see review comments.

martbab:
There were some PEP8 errors but I fixed them and pushed to:

ipa-4-4:
https://fedorahosted.org/freeipa/changeset/d028d23c5f0c3e1b18c15fad67a0893870f5d27c
master:
https://fedorahosted.org/freeipa/changeset/26bd7ebfa27d15221e5d3fa1e3871a0085c31e0f

PR #325, opened 2016-12-12
WebUI: Hide incorrectly shown buttons on hosts tab in ID Views

pvomacka:
There was missing default value for evaluator adapter.
In that case the adapter variable could be undefined and
it crashes on building adapter. Therefore it did not
evaluate all evaluators. That is the reason why 'Delete'
and 'Add' buttons were incorrectly shown.
Default value is now set to empty object.

https://fedorahosted.org/freeipa/ticket/6546


pvoborni:
Works for me

pvoborni:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/17392b0ef754781775a10973b2fee8a6d1697f5d
ipa-4-4:
https://fedorahosted.org/freeipa/changeset/cd62808c0e62013f594b0a31acf5bc65f4ce0683

PR #326, opened 2016-12-12
adtrust: remove FILE: prefix from 'dedicated keytab file' in smb.conf

abbra:
Samba 4.5 does not allow to specify access mode for the keytab (FILE: or
WRFILE:) from external sources. Thus, change the defaults to a path
(implies FILE: prefix) while Samba Team fixes the code to allow the
access mode prefix for keytabs.

On upgrade we need to replace 'dedicated keytab file' value with the
path to the Samba keytab that FreeIPA maintains. Since the configuration
is stored in the Samba registry, we use net utility to manipulate the
configuration:

    net conf setparm global 'dedicated keytab file' /etc/samba/samba.keytab

Fixes https://fedorahosted.org/freeipa/ticket/6551


ipaserver/install/server/upgrade.py
```diff
@@ -1653,6 +1674,7 @@ def upgrade_configuration():
 
     cleanup_kdc(fstore)
     cleanup_adtrust(fstore)
+    upgrade_atrust_config()
```
martbab: Typo here, should be `upgrade_adtrust_config`

martbab:
Pylint complaints about undefined function because you made a typo.

abbra:
Thanks, fixed it.

martbab:
The branch needs a rebase. Otherwise works as expected and has no PEP8 errors despite what crazy Travis CI claims.

abbra:
Rebased to git master.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/38cc01b1c92da36653e0ce4d8f7066282fd1d102

martbab:
Rebased and pushed to ipa-4-4

https://fedorahosted.org/freeipa/changeset/fd8c17252fbc3efccad66cdcd373195c0fc53ce3

PR #327, opened 2016-12-12
WebUI: RPC refactoring

pvomacka:
Moved from ML ( https://www.redhat.com/archives/freeipa-devel/2016-November/msg00338.html ) to PR.


pvomacka:
The last comment from pvoborni: 
"patch 84:

Looks good, works fine, it just needed rebase(I could provide that).

Idea, but that doesn't have to be implemented, or sometime in future,
right now it is not useful: What about providing the rpc object in the
event, and having unique id for each rpc call so that we could track all
rpc which are executed.


patch 101:

1. It's event name but the property name looks like that it contains a text:
   that.change_text = 'change-activity-text';

Should it be rather: that.change_text_event.

Or even, why does it compare previous text? Does it matter? Wouldn't be
better to have 'set-activity' event. And then the handler would call
something new set_text method:

set_text(new_activity)
  that.dots = 0
  that.text = new_activity
  that.make_step()


-- 
Petr Vobornik"

pvomacka:
Patch 84: Yes, that is really good idea, but as you said - we don't have usecase for it right now. But I created a ticket to not forget about it. https://fedorahosted.org/freeipa/ticket/6553

Patch 101: I changed the name of event to 'set-activity-event', now it accepts one parameter which is new text of activity widget. But it can be extended in the future to accept more parameters and set more attributes. 

I think that creating new method "that.set_text" or something similar can lead to calling that method instead of using topics (events). That's the reason why I left the setting of text in anonymous function (event listener).

pvoborni:
works for me, the travis failure is invalid̈́ - web ui is not related to the tests and pylint passes

pvoborni:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/5a950aeb29963ed22a2c3c1b80723589ac4097de
https://fedorahosted.org/freeipa/changeset/be7865bf4f9b6774a17f31380e96b76d0473f982

PR #328, opened 2016-12-12
fix: regression in API version comparison

mbasti-rh:
Commint 2cbaf156045769b54150e4d4c3c1071f164a16fb introduced a regression
in API version comparison. In case that newer client is trying to call
older server an error is returned, but it should work. This commit fixes
it.


mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/0663faf2585be4af3501965934703da0ede41610

PR #329, opened 2016-12-13
experiment: did pull/177 break ci?

frasertweedale:
This PR reverts the commits from pull/177 to test the hypothesis
that something in these commits broke CI.


frasertweedale:
Yes, it looks like I broke CI.  Feel free to merge this PR if I don't find a fix quickly enough.

PR #330, opened 2016-12-13
Build: forbid builds in working directories containing white spaces

pspacek:
Spaces are causing problems in libtool, makefiles, autoconf itself, gettextize
framework etc. so this issue cannot be easily fixed.

Return on investment is too small to invest into this. Let's detect the
whitespace early and error out with descriptive error message.

https://fedorahosted.org/freeipa/ticket/6537

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5,6 +5,14 @@ AC_INIT([freeipa],
         IPA_VERSION,
         [https://hosted.fedoraproject.org/projects/freeipa/newticket])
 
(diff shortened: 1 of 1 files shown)


martbab:
I agree that a safeguard that fails early is sufficient for this corner-case unless someone proves us otherwise. PRs are welcome in that case.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/19aba7c555edf065b9f2fa95142da81b92396264

PR #331, opened 2016-12-13
WebUI: don't change casing of Auth Indicators values

pvomacka:
All values were previously converted to lowercase which was not
coresponding with CLI behaviour. Now they stay as they are
inserted. I also have to change the strings to lowercase because
the otp and radius should be inserted as lowercase words.

https://fedorahosted.org/freeipa/ticket/6308

### DIFF
--- a/install/ui/src/freeipa/widget.js
+++ b/install/ui/src/freeipa/widget.js
@@ -2502,6 +2502,8 @@ IPA.custom_checkboxes_widget = function(spec) {
 
     var that = IPA.checkboxes_widget(spec);
 
+    that.set_value_to_lowercase = spec.set_value_to_lowercase || false;
+
     that.add_dialog_title = spec.add_dialog_title ||
                             "@i18n:dialogs.add_custom_value";
     that.add_field_label = spec.add_field_label ||
@@ -2619,7 +2621,7 @@ IPA.custom_checkboxes_widget = function(spec) {
 
             if (!value || value === '') continue;
 
-            value = value.toLowerCase();
+            if (that.set_value_to_lowercase) value = value.toLowerCase();
             that.values.push(value);
         }
 
--- a/install/ui/src/freeipa/aci.js
+++ b/install/ui/src/freeipa/aci.js
@@ -546,6 +546,9 @@ aci.attributes_widget = function(spec) {
     spec.add_field_label = spec.add_field_label ||
                             '@i18n:objects.permission.attribute';
 
+    spec.set_value_to_lowercase = spec.set_value_to_lowercase === undefined ?
+                    true : spec.set_value_to_lowercase;
+
     var that = IPA.custom_checkboxes_widget(spec);
 
     that.object_type = spec.object_type;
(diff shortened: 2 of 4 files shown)


ipaserver/plugins/internal.py
```diff
@@ -201,10 +201,10 @@ class i18n_messages(Command):
             "auth_indicator": _("Authentication indicator"),
             "config_tooltip": _("<p>Implicit method (password) will be used if no method is chosen.</p><p><strong>Password + Two-factor:</strong> LDAP and Kerberos allow authentication with either one of the authentication types but Kerberos uses pre-authentication method which requires to use armor ccache.</p><p><strong>RADIUS with another type:</strong> Kerberos always use RADIUS, but LDAP never does. LDAP only recognize the password and two-factor authentication options.</p>"),
             "custom_auth_ind_title": _("Add Custom Authentication Indicator"),
-            "otp": _("OTP"),
+            "otp": _("otp"),
             "type_otp": _("Two factor authentication (password + OTP)"),
             "type_password": _("Password"),
-            "type_radius": _("RADIUS"),
+            "type_radius": _("radius"),
```
pvoborni: Changing this shows then invalid lowercase value on IPA Server/Configuration page for auth types and also on user details page.

It would be better to create separate:
```
authindicator_otp: _("otp"),
authindicator_radius _("radius"),
```
And use those for auth indicators checkboxes to avoid such side effects.

install/ui/src/freeipa/widget.js
```diff
@@ -2509,6 +2509,9 @@ IPA.custom_checkboxes_widget = function(spec) {
 
     var that = IPA.checkboxes_widget(spec);
 
+    that.set_value_to_lowercase = spec.set_value_to_lowercase === undefined
+                                    ? true : spec.set_value_to_lowercase;
```
pvoborni: Would be better if the default behavior was 'no lower casing' -> change to false and then adjust usages of the widget.

Reasoning: no formatting seems more nutural

ipaserver/plugins/internal.py
pvoborni: or maybe even better, auth indicators should not use translated labels. So labels should be the same value as value and they can be hardcoded.

pvoborni:
LGTM (reading code).

install/ui/src/freeipa/widget.js
```diff
@@ -2509,6 +2509,8 @@ IPA.custom_checkboxes_widget = function(spec) {
 
     var that = IPA.checkboxes_widget(spec);
 
+    that.set_value_to_lowercase = spec.set_value_to_lowercase || false;
```
pvoborni: When the default is changed, it needs to be changed also in the other widget where it was used: `aci.attributes_widget` 

Without it, following command will break Web UI: `ipa permission-mod AA_foo --attrs=businessCategory`

pvoborni:
ACK but I've find out that the change is not enough because of existing bug. See pr #554

dkupka:
master:

* 0220fc8986e4fef017185bde675dc9cf0f90afd8 WebUI: Allow disabling lowering text in custom_checkbox_widget
* ad3451067ad474ea52872913d6789b1652f9a9c4 WebUI: don't change casing of Auth Indicators values

PR #332, opened 2016-12-13
Fix regression in test suite

frasertweedale:
32b1743e5fb318b226a602ec8d9a4b6ef2a25c9d introduced a regression in
test_serverroles.py, caused by ca_find attempting to log into the
Dogtag REST API.  (ca_find is called by cert_find which is caused by
server_del during cleanup).

Avoid logging into Dogtag in cert_find unless something actually
needs to be retrieved.

Fixes: https://fedorahosted.org/freeipa/ticket/6178

### DIFF
--- a/ipaserver/plugins/ca.py
+++ b/ipaserver/plugins/ca.py
@@ -161,15 +161,21 @@ class ca(LDAPObject):
     }
 
 
-def set_certificate_attrs(entry, options, always_include_cert=True):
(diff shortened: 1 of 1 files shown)


martbab:
Thanks for pinpointing the issue, looks like Travis is happy with your fix. 👍

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/74b8cf2c4a8dd36577d76c35a9ef08352ef025b7

PR #338, opened 2016-12-14
 password policy: Add explicit default password policy for hosts and services

dkupka:


dkupka:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/6f1d927467e7907fd1991f88388d96c67c9bff61
https://fedorahosted.org/freeipa/changeset/b1a20599c4f9fdcd208998694185b65460126703

PR #341, opened 2016-12-14
certprofile-mod: correctly authorise config update

mbasti-rh:
Certificate profiles consist of an FreeIPA object, and a
corresponding Dogtag configuration object.  When updating profile
configuration, changes to the Dogtag configuration are not properly
authorised, allowing unprivileged operators to modify (but not
create or delete) profiles.  This could result in issuance of
certificates with fraudulent subject naming information, improper
key usage, or other badness.

Update certprofile-mod to ensure that the operator has permission to
modify FreeIPA certprofile objects before modifying the Dogtag
configuration.

https://fedorahosted.org/freeipa/ticket/6560

### DIFF
--- a/ipaserver/plugins/certprofile.py
+++ b/ipaserver/plugins/certprofile.py
@@ -310,6 +310,11 @@ class certprofile_mod(LDAPUpdate):
             raise errors.ProtectedEntryError(label='certprofile', key=keys[0],
                 reason=_('Certificate profiles cannot be renamed'))
         if 'file' in options:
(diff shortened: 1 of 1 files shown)


mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/fec4c32ff15a96736740cf7d2f713a21af0b227e
ipa-4-4:
https://fedorahosted.org/freeipa/changeset/c12a52f0d78b30931713a3548b22e799d41f3622

PR #342, opened 2016-12-14
[4.3] password policy: Add explicit default password policy for hosts and services

dkupka:
Set explicitly krbPwdPolicyReference attribute to all hosts (entries in
cn=computers,cn=accounts), services (entries in cn=services,cn=accounts) and
Kerberos services (entries in cn=$REALM,cn=kerberos). This is done using DS's
CoS so no attributes are really added.

The default policies effectively disable any enforcement or lockout for hosts
and services. Since hosts and services use keytabs passwords enforcements
doesn't make much sense. Also the lockout policy could be used for easy and
cheap DoS.


dkupka:
Fixed upstream
ipa-4-3:
https://fedorahosted.org/freeipa/changeset/42263a5a729096135702c0b974f255a058c0cdaf

PR #343, opened 2016-12-14
[4.3] certprofile-mod: correctly authorise config update

mbasti-rh:
Certificate profiles consist of an FreeIPA object, and a
corresponding Dogtag configuration object.  When updating profile
configuration, changes to the Dogtag configuration are not properly
authorised, allowing unprivileged operators to modify (but not
create or delete) profiles.  This could result in issuance of
certificates with fraudulent subject naming information, improper
key usage, or other badness.

Update certprofile-mod to ensure that the operator has permission to
modify FreeIPA certprofile objects before modifying the Dogtag
configuration.

https://fedorahosted.org/freeipa/ticket/6560


mbasti-rh:
Fixed upstream
ipa-4-3:
https://fedorahosted.org/freeipa/changeset/278d7cf4708f1c6ac05d5fcb21db582d9aa7bab3

PR #345, opened 2016-12-15
ipa-kdb: search for password policies globally

abbra:
With the CoS templates now used to create additional password policies
per object type that are placed under the object subtrees, DAL driver
needs to search for the policies in the whole tree.

Individual policies referenced by the krbPwdPolicyReference attribute
are always searched by their full DN and with the base scope. However,
when KDC asks a DAL driver to return a password policy by name, we don't
have any specific base to search. The original code did search by the
realm subtree.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1404910

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb_pwdpolicy.c
+++ b/daemons/ipa-kdb/ipa_kdb_pwdpolicy.c
@@ -163,7 +163,7 @@ krb5_error_code ipadb_get_pwd_policy(krb5_context kcontext, char *name,
     }
 
     kerr = ipadb_simple_search(ipactx,
-                               ipactx->realm_base, LDAP_SCOPE_SUBTREE,
+                               ipactx->base, LDAP_SCOPE_SUBTREE,
                                src_filter, std_pwdpolicy_attrs, &res);
     if (kerr) {
         goto done;


martbab:
I have taken the liberty to fix the commit message to reference https://fedorahosted.org/freeipa/ticket/6561 instead to BZ and pushed it to:

ipa-4-3:
https://fedorahosted.org/freeipa/changeset/b9b919e127c453eda02ea142d7cd80c16aa5ca31
ipa-4-4:
https://fedorahosted.org/freeipa/changeset/84f6df6349b5c412467746777e905d9e4f8792ca
master:
https://fedorahosted.org/freeipa/changeset/73f33569c8893610e246b2f44a7aeaec872b37e6

simo5:
I know this is already closed but NACK.
The problem here is in searching "base"
this means ending up serhing also in things like slapi-nis.
We need to change the code to search in cn=REALM, and, if that fails, search again in cn=accounts.

I do not know if we should revert or just patch on top.

abbra:
NACK to @simo5 concerns. We are not affected by slapi-nis on searches from KDC.

PR #346, opened 2016-12-15
Minimal test suite for kadmin.local

martbab:
This ensures that we do not break it (much) in the
future like we did recently

https://fedorahosted.org/freeipa/ticket/6561

### DIFF
--- a/ipatests/test_ipaserver/test_kadmin.py
+++ b/ipatests/test_ipaserver/test_kadmin.py
@@ -1,0 +1,125 @@
+#
+# Copyright (C) 2016 FreeIPA Contributors see COPYING for license
+#
+
+"""
+Test suite for creating principals via kadmin.local and modifying their keys
+"""
+
+import os
+import pytest
+import tempfile
+
+from ipalib import api
+
+from ipaserver.install import installutils
+
+
+@pytest.yield_fixture()
+def keytab():
+    fd, keytab_path = tempfile.mkstemp(suffix='.keytab')
+    os.close(fd)
+
+    try:
+        yield keytab_path
+    finally:
+        try:
+            os.remove(keytab_path)
+        except OSError:
+            pass
+
+
+@pytest.fixture()
+def service_in_kerberos_subtree(request):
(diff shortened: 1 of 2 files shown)


ipatests/test_ipaserver/test_kadmin.py
```diff
@@ -0,0 +1,125 @@
...
+@pytest.mark.skipif(
+    os.getuid() != 0, reason="kadmin.local is accesible only to root")
+class TestKadmin(object):
+    def assert_success(self, command, *args):
+        """
+        Since kadmin.local returns 0 also when internal errors occur, we have
+        to catch the command's stderr and check that it is empty
+        """
+        result = command(*args)
+        assert not result.error_output
```
mbasti-rh: Shouldnt be here tested rather return code than stderr?

stlaz: In the last case, there was a Kerberos internal database error but the return code was still 0 (also pointed out in the comment). I wonder, could this cause issues should Kerberos throw some warnings on stderr or do we assume no warnings should be present at all at any time?

martbab: Warnings are also thrown in stderr (e.g. when no password policy was specified) but they are not tested in this case as they are part of setup. The problem here is, as Standa pointed out, that if the query fails, kardmin still returns 0, so testing for the presence of stderr is still the sanest way to assert command success. But I can re-implement the test to assert success by expecting certain messages in stdout if you wish.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/f59673506493e0199c17000538adb7c80b477aa0
https://fedorahosted.org/freeipa/changeset/d95bdbbfd505dde1348413fc7a1233ac834ce344
ipa-4-4:
https://fedorahosted.org/freeipa/changeset/f0f48ec14f3ff55852393927533ffd253cb5a04b
https://fedorahosted.org/freeipa/changeset/e02323c1c3b3c3dadd57d9f1885ec1af046718de

PR #347, opened 2016-12-16
Improvements in {get|set}_directive functions

martbab:
These should fix https://fedorahosted.org/freeipa/ticket/6460 and future-proof
the code a bit

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -388,11 +388,14 @@ def set_directive(filename, directive, value, quotes=True, separator=' ',
 
     This has only been tested with nss.conf
 
-   :param directive: directive name
-   :param value: value of the directive
-   :param quotes: whether to quote `value` in `quote_char`. If true, then
-        the `quote_char` are first escaped to avoid unparseable directives
-   :param quote_char: the character used for quoting `value`
+    :param filename: input filename
+    :param directive: directive name
+    :param value: value of the directive
+    :param quotes: whether to quote `value` in `quote_char`. If true, then
+        the `quote_char` are first escaped to avoid unparseable directives.
+    :param separator: character serving as separator between directive and
+        value
+    :param quote_char: the character used for quoting `value`
     """
 
     def format_directive(directive, value, separator, quotes, quote_char):
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -436,16 +436,31 @@ def set_directive(filename, directive, value, quotes=True, separator=' ',
     fd.close()
     os.chown(filename, st.st_uid, st.st_gid) # reset perms
 
+
 def get_directive(filename, directive, separator=' '):
     """
     A rather inefficient way to get a configuration directive.
+
+    :param filename: input filename
+    :param directive: directive name
+    :param separator: separator between directive and value
+    :param quote_char: the characters that are used in this particular config
+        file to quote values. This character will be stripped and unescaped
+        from the raw value.
+
+    :returns: The (unquoted) value if the directive was found, None otherwise
     """
     fd = open(filename, "r")
     for line in fd:
         if line.lstrip().startswith(directive):
             line = line.strip()
-            result = line.split(separator, 1)[1]
-            result = result.strip('"')
+
+            (directive, sep, value) = line.partition(separator)
+            if not sep or not value:
+                raise ValueError("Malformed directive: {}".format(line))
+
+            result = value.strip().strip(quote_char)
+            result = ipautil.unescape_seq(quote_char, result)[0]
             result = result.strip(' ')
             fd.close()
             return result
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -380,8 +380,38 @@ def update_file(filename, orig, subst):
         return 1
 
 
-def set_directive(filename, directive, value, quotes=True, separator=' ',
-                  quote_char='\"'):
+def quote_directive_value(value, quote_char):
+    """Quote a directive value
+    :param value: string to quote
+    :param quote_char: character which is used for quoting. All prior
+        occurences will be escaped before quoting to avoid unparseable value.
+    :returns: processed value
+    """
+    if value.startswith(quote_char) and value.endswith(quote_char):
+        return value
+
+    return "{quote}{value}{quote}".format(
+        quote=quote_char,
+        value="".join(ipautil.escape_seq(quote_char, value))
+    )
+
+
+def unquote_directive_value(value, quote_char):
+    """Unquote a directive value
+    :param value: string to unquote
+    :param quote_char: character to strip. All escaped occurences of
+        `quote_char` will be uncescaped during processing
+    :returns: processed value
+    """
+    unescaped_value = "".join(ipautil.unescape_seq(quote_char, value))
+    if (unescaped_value.startswith(quote_char) and
+            unescaped_value.endswith(quote_char)):
+        return unescaped_value[1:-1]
+
+    return unescaped_value
+
+
+def set_directive(filename, directive, value, quotes=True, separator=' '):
     """Set a name/value pair directive in a configuration file.
 
     A value of None means to drop the directive.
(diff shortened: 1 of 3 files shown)


ipapython/ipautil.py
```diff
@@ -1294,15 +1294,17 @@ def unescape_seq(seq, *args):
 
 def escape_seq(seq, *args):
     """
-    escape (prepend '\\') all occurences of sequence in input strings
+    escape (prepend '\\') all occurences of sequence in input strings unless it
+    is already escaped.
```
pspacek: I think that the function called `escape_seq` should have no say in what is being escaped. Given that next commit modified `get_directive` to be inverse of `set_directive`, why is this change necessary at all?

ipaserver/install/installutils.py
```diff
@@ -436,16 +436,27 @@ def format_directive(directive, value, separator, quotes, quote_char):
...
     A rather inefficient way to get a configuration directive.
+
+    :param filename: input filename
+    :param directive: directive name
+    :param separator: separator between directive and value
+    :param quote_char: the characters that are used in this particular config
+        file to quote values. This character will be stripped and unescaped
+        from the raw value.
+
+    :returns: The (unquoted) value if the directive was found, None otherwise
```
pspacek: I know this is not originally your code but KeyError exception souds more appropriate to me (for non-existing directives)...

ipaserver/install/installutils.py
```diff
@@ -436,16 +436,27 @@ def format_directive(directive, value, separator, quotes, quote_char):
...
     """
     fd = open(filename, "r")
     for line in fd:
         if line.lstrip().startswith(directive):
             line = line.strip()
             result = line.split(separator, 1)[1]
-            result = result.strip('"')
+            result = result.strip(quote_char)
```
pspacek: I think that this usage of `strip` will cause problems if the text in config file is like `\"abc\"`. It will remove only the last `"` (which is incorrect anyway because `"` are part of the string and not quotation).

pspacek: BTW this usage of `strip` does not remove any whitespace (unless `quote_char` is a whitespace ...) so the commit message needs correction as well.

pspacek:
Please see my in-line comments.

ipapython/ipautil.py
martbab: Probably not, I was just thinking about future proofing the code. You are probably right that the function should be as dumb as possible.

ipaserver/install/installutils.py
martbab: I would rather not alter the function's contract in this way, who know how many callers are relying on this particular behavior. If you open a ticket for this I can do it in separate PR.

ipaserver/install/installutils.py
martbab: Regarding stripping the quote_char, I realized that the correct thing to do is actually to first unescape all quote chars and then do the strip as this is the inverse operation of set_directive. Do you agree with such a change?.

martbab:
Please see my in-line replies to your in-line comments :).

ipaserver/install/installutils.py
pspacek: No problem, I do not insist on this change.

ipaserver/install/installutils.py
pspacek: Hmm, that depends. How is `quote_char` supposed to behave? Can it be used just once in each value like `    "this is my value"    `? Or is it possible to use it as in C strings like `    "first"  " part" "of " "string" ` producing string value `first part of string`?

Speaking of whitespace, the functions should be really inversible. Is there a reason to mock with the whitespace when reading values? (Just wondering out loud.)

pspacek: Thinking about it a bit more, I would argue that first use of `strip()` on line https://github.com/freeipa/freeipa/pull/347#discussion-diff-92979573R456 is incorrect anyway. The code first does `strip()` and then `split(separator, 1)[1]`.

What if result of `split(separator, 1)[1]` contains whitespace?

martbab:
Can somebody take over the review of this PR please?

ipaserver/install/ipa_server_certinstall.py
```diff
@@ -134,14 +134,18 @@ def install_http_cert(self):
         dirname = certs.NSS_DIR
 
         old_cert = installutils.get_directive(paths.HTTPD_NSS_CONF,
-                                              'NSSNickname')
+                                              'NSSNickname', quote_char="'")
```
tomaskrizek: Why is it safe to assume the `NSSNickname` will always be enclosed in single quotes?

mod_nss allows values to be enclosed in double quotes. If double quotes were indeed used, they'd be interpreted as a part of the value, which is not intended. It would be the exact same issue as ticket 6460, except for double quotes instead of single quotes.

tomaskrizek:
Please see my feedback in in-line comments.

ipaserver/install/ipa_server_certinstall.py
martbab: You are right. I think that I will delegate the quoting/unquoting transformations to separate functions. That will be more robust IMHO.

ipaserver/install/installutils.py
```diff
@@ -380,33 +380,55 @@ def update_file(filename, orig, subst):
...
+
+def unquote_directive_value(value, quote_char):
+    """Unquote a directive value
+    :param value: string to unquote
+    :param quote_char: character to strip. All escaped occurences of
+        `quote_char` will be uncescaped during processing
+    :returns: processed value
+    """
+    unescaped_value = "".join(ipautil.unescape_seq(quote_char, value))
+    return unescaped_value.strip(quote_char)
```
tomaskrizek: Using `strip()` causes issues when the actual value starts or ends with the same type of quote as is used to escape it, e.g. `'cn=\'ABC\''` gets stripped to `cn='ABC`.

ipaserver/install/installutils.py
```diff
@@ -380,33 +380,55 @@ def update_file(filename, orig, subst):
...
-   :param directive: directive name
-   :param value: value of the directive
-   :param quotes: whether to quote `value` in `quote_char`. If true, then
-        the `quote_char` are first escaped to avoid unparseable directives
-   :param quote_char: the character used for quoting `value`
+    :param filename: input filename
+    :param directive: directive name
+    :param value: value of the directive
+    :param quotes: whether to quote `value` in `quote_char`. If true, then
+        the `quote_char` are first escaped to avoid unparseable directives.
```
tomaskrizek: Please remove reference to quote_char from comment.

ipaserver/install/installutils.py
```diff
@@ -416,33 +438,43 @@ def format_directive(directive, value, separator, quotes, quote_char):
...
 
+
 def get_directive(filename, directive, separator=' '):
     """
     A rather inefficient way to get a configuration directive.
+
+    :param filename: input filename
+    :param directive: directive name
+    :param separator: separator between directive and value
+    :param quote_char: the characters that are used in this particular config
```
tomaskrizek: Update doc here as well.

tomaskrizek:
I still managed to find a an issue for certain edge cases. See inline comments for more info.

martbab:
Thanks, I have fixed the docstrings. I have also made directive unquoting less silly.

tomaskrizek:
I wasn't able to find any more issues with the quoting of certificate names. The directive quoting seems to work properly now.

martbab:
Thank's, let's hope that all this code will be replaced by some proper configuration parsing mechanism in the future.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/e1ed8b5eff40331ba532d37f3fb08814d8a55b77
https://fedorahosted.org/freeipa/changeset/517d43e78b8d8ea0b796a6ff6a379236eaae21df
https://fedorahosted.org/freeipa/changeset/2831b30e9a9de947481c058d8d32e174f951b1c0
https://fedorahosted.org/freeipa/changeset/86f4a93fb3aeb6742acab5abaa1c17b525ea4223

PR #348, opened 2016-12-19
ca: fix ca-find with --pkey-only

HonzaCholasta:
Since commit 32b1743e5fb318b226a602ec8d9a4b6ef2a25c9d, ca-find will fail
with internal error if --pkey-only is specified, because the code to
look up the CA certificate and certificate chain assumes that the ipaCAId
attribute is always present in the result.

Fix this by not attempting to lookup the certificate / chain at all when
--pkey-only is specified.

https://fedorahosted.org/freeipa/ticket/6178

### DIFF
--- a/ipaserver/plugins/ca.py
+++ b/ipaserver/plugins/ca.py
@@ -162,7 +162,10 @@ class ca(LDAPObject):
 
 
 def set_certificate_attrs(entry, options, want_cert=True):
-    ca_id = entry['ipacaid'][0]
+    try:
+        ca_id = entry['ipacaid'][0]
+    except KeyError:
+        return
     full = options.get('all', False)
     want_chain = options.get('chain', False)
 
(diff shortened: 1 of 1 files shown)


frasertweedale:
IMO the current change is fine, but I would also implement a defensive guard within `set_certificate_attrs` in case this somehow happens in some other command.

```python
def set_certificate_attrs(entry, options, want_cert=True):
    if 'ipacaid' not in entry:
        return
    ca_id = entry['ipacaid'][0]
    ...
```

frasertweedale:
LGTM

HonzaCholasta:
@frasertweedale, is that an ACK? :-)

frasertweedale:
It is an ACK.  I don't have perms to add the label tho :)

mbasti-rh:
@frasertweedale your permissions have been upgraded :)

frasertweedale:
Thanks @mbasti-rh !

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/ceb26f5ac428cdbed8ec1fa89e9ed6f1d903a5a0

PR #349, opened 2016-12-19
spec file: do not define with_lint inside a comment

jcholast:
RPM expands macros even inside comments in spec files, so the with_lint
macro is unintentionally always defined.

Escape the percent sign in '%global' in the comment to prevent this.


martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/1b85e59ceeb115dfb24e9c6bacb665b935f9543c

PR #350, opened 2016-12-19
spec file: revert to the previous Release tag

HonzaCholasta:
Revert from the current Release tag value `upstream` to the previously used
`0%{?dist}`, because:

* `0` sorts before `1`, which is usually used as the initial release number
  in downstream packages,

* the information provided by `%{?dist}` is useful, as packages built on
  one OS are not always installable on another OS.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -41,7 +41,7 @@
 
 Name:           freeipa
 Version:        %{IPA_VERSION}
-Release:        upstream
+Release:        0%{?dist}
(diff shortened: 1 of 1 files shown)


stlaz:
I also appreciate the dist information, ACK.

HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/eb1f05d598d821f8e7eb5b8cfe606f570052f263

PR #351, opened 2016-12-19
[fedora-26] named.conf template: update API for bind 9.11

tomaskrizek:
Please **do not merge** this patch upstream, we need to have BIND 9.11 available before we do, otherwise it will break DNS installation. This patch is intended for Fedora 26 downstream and I'm only posting it for review.

This patch only fixes DNS for new IPA installations. Another patch for fixing existing named configs is necessary. This will most likely be fixed in bind-dyndb-ldap upstream.

---

Use the new API for bind 9.11. Removed deprecated "serial_autoincrement"
and updated the rest of configuration to conform to the new format.

https://fedorahosted.org/freeipa/ticket/6565

### DIFF
--- a/install/share/bind.named.conf.template
+++ b/install/share/bind.named.conf.template
@@ -43,13 +43,11 @@ zone "." IN {
 include "$RFC1912_ZONES";
 include "$ROOT_KEY";
 
-dynamic-db "ipa" {
-	library "ldap.so";
-	arg "uri ldapi://%2fvar%2frun%2fslapd-$SERVER_ID.socket";
-	arg "base cn=dns, $SUFFIX";
-	arg "server_id $FQDN";
-	arg "auth_method sasl";
-	arg "sasl_mech GSSAPI";
-	arg "sasl_user DNS/$FQDN";
-	arg "serial_autoincrement yes";
+dyndb "ipa" "$BIND_LDAP_SO" {
+	uri "ldapi://%2fvar%2frun%2fslapd-$SERVER_ID.socket";
+	base "cn=dns, $SUFFIX";
+	server_id "$FQDN";
+	auth_method "sasl";
+	sasl_mech "GSSAPI";
+	sasl_user "DNS/$FQDN";
 };
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -371,18 +371,11 @@ Summary: IPA integrated DNS server with support for automatic DNSSEC signing
 Group: System Environment/Base
 BuildArch: noarch
 Requires: %{name}-server = %{version}-%{release}
-Requires: bind-dyndb-ldap >= 10.0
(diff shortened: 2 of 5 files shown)


tomaskrizek:
The version of BIND in `freeipa.spec.in` is also subject to change. There is currently a bug that affects `named-pkcs11`, but not `named`.

I also have a [COPR repo](https://copr.fedorainfracloud.org/coprs/tkrizek/bind-9.11/) with bind 9.11 for F25. I'm going to add bind-dyndb-ldap v11.0 there as well.

I'll also try to build for Fedora 24.

install/share/bind.named.conf.template
```diff
@@ -43,13 +43,11 @@ zone "." IN {
...
 
-dynamic-db "ipa" {
-	library "ldap.so";
-	arg "uri ldapi://%2fvar%2frun%2fslapd-$SERVER_ID.socket";
-	arg "base cn=dns, $SUFFIX";
-	arg "server_id $FQDN";
-	arg "auth_method sasl";
-	arg "sasl_mech GSSAPI";
-	arg "sasl_user DNS/$FQDN";
-	arg "serial_autoincrement yes";
```
mbasti-rh: why serial autoincrement was removed?

mbasti-rh:
How do you solve upgrades F25->F26?

tomaskrizek:
This fix only applies to new IPA installations.

Upgrade of `named.conf` will be handled separately by bind-dyndb-ldap. When a new version will be installed, a postinstall scriptet will run a script to transform `named.conf` to the new format.

install/share/bind.named.conf.template
tomaskrizek: It's an obsolote option that is no longer supported.

tomaskrizek:
Required version of BIND is a subject to change. When a version with fixed `named-pkcs11` issue ([BZ 1410433](https://bugzilla.redhat.com/show_bug.cgi?id=1410433)) is released, I will update it. Patch should not be merged until then.

tomaskrizek:
I've updated the version of BIND and the patch should be complete. I suggest we do a review (you can use https://copr.fedorainfracloud.org/coprs/tkrizek/bind-9.11/ for F24/F25), but delay merging this patch so we do not have to use the COPR for our upstream development until necessary.

MartinBasti:
Tested manually

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/c26dd805bdb020b12346d8cb66638883c1f46b9e
https://fedorahosted.org/freeipa/changeset/e8a2abd548b594e6f22f38445ee32bcaa7f27303
https://fedorahosted.org/freeipa/changeset/5de7065fe5769e5c3d90205b0ecc963d96f4db58

PR #354, opened 2016-12-20
Fix DL1 replica installation in CA-less topology

frasertweedale:
Commit dbb98765d73519289ee22f3de1a5ccde140f6f5d changed certmonger
requests for DS and HTTP certificates during installation to raise
on error (https://fedorahosted.org/freeipa/ticket/6514).
This introduced a regression in DL1 replica installation in CA-less
topology.  A certificate was requested, but prior to the
aforementioned commit this would fail silently and installation
continued, whereas now installation fails.

Guard the certificate request with a check that the topology is
CA-ful.

Fixes: https://fedorahosted.org/freeipa/ticket/6573

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -1338,7 +1338,7 @@ def install(installer):
         # Always try to install DNS records
         install_dns_records(config, options, remote_api)
 
-        if promote:
+        if promote and ca_enabled:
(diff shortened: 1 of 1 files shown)


martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/4028ad73e74fe62bd4871e842dbb69ff660125f9

PR #41, opened 2016-08-31
Postpone enabling LDAPS in replica promotion

tomaskrizek:
Fixes a bug that prevented ipa-replica-install with CA, because
LDAPS was configured before the SSL cerificate was assigned.

https://fedorahosted.org/freeipa/ticket/6226

PR #355, opened 2016-12-20
Set up DS TLS on replica in CA-less topology

frasertweedale:
Fixes: https://fedorahosted.org/freeipa/ticket/6226

### DIFF
--- a/ipaserver/install/dsinstance.py
+++ b/ipaserver/install/dsinstance.py
@@ -390,7 +390,9 @@ class DsInstance(service.Service):
 
         self.step("creating DS keytab", self._request_service_keytab)
         if self.promote:
-            if self.ca_is_configured:
+            if self.pkcs12_info:
+                self.step("configuring ssl for ds instance", self.__enable_ssl)
+            else:
                 self.step("retrieving DS Certificate", self.__get_ds_cert)
             self.step("restarting directory server", self.__restart_instance)
 
--- a/ipaserver/install/dsinstance.py
+++ b/ipaserver/install/dsinstance.py
@@ -278,7 +278,7 @@ class DsInstance(service.Service):
         self.step("creating indices", self.__create_indices)
         self.step("enabling referential integrity plugin", self.__add_referint_module)
         if enable_ssl:
-            self.step("configuring ssl for ds instance", self.__enable_ssl)
+            self.step("configuring TLS for DS instance", self.__enable_ssl)
         self.step("configuring certmap.conf", self.__certmap_conf)
         self.step("configure new location for managed entries", self.__repoint_managed_entries)
         self.step("configure dirsrv ccache", self.configure_dirsrv_ccache)
@@ -351,7 +351,7 @@ class DsInstance(service.Service):
     def enable_ssl(self):
         self.steps = []
 
-        self.step("configuring ssl for ds instance", self.__enable_ssl)
+        self.step("configuring TLS for DS instance", self.__enable_ssl)
         self.step("restarting directory server", self.__restart_instance)
         self.step("adding CA certificate entry", self.__upload_ca_cert)
 
@@ -391,7 +391,7 @@ class DsInstance(service.Service):
         self.step("creating DS keytab", self._request_service_keytab)
         if self.promote:
             if self.pkcs12_info:
-                self.step("configuring ssl for ds instance", self.__enable_ssl)
+                self.step("configuring TLS for DS instance", self.__enable_ssl)
             else:
                 self.step("retrieving DS Certificate", self.__get_ds_cert)
             self.step("restarting directory server", self.__restart_instance)
@@ -559,9 +559,9 @@ class DsInstance(service.Service):
         root_logger.debug("calling setup-ds.pl")
         try:
             ipautil.run(args)
-            root_logger.debug("completed creating ds instance")
+            root_logger.debug("completed creating DS instance")
         except ipautil.CalledProcessError as e:
-            raise RuntimeError("failed to create ds instance %s" % e)
+            raise RuntimeError("failed to create DS instance %s" % e)
 
         # check for open port 389 from now on
         self.open_ports.append(389)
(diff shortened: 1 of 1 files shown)


jcholast:
This is basically the same as 89de60c5d8ba64d619101a7498b8c4469b6e50ae which had to be reverted because it is not the proper fix.

I would rather wait for the proper fix (#41), which has not been merged yet because it is blocked by https://bugzilla.redhat.com/show_bug.cgi?id=1377413.

tomaskrizek:
89de60c was reveted because while it fixed this particular use case, it broke others. IIRC it broke regular replica promotion with CA.

The proper fix is not yet ready, nor on the IPA side (#41 is a step in the right direction, but it also requires some more code fixes, especially properly closing some ad hoc LDAP connections), nor on the NSS side (ETA unknown).

If this patch works and doesn't break other use cases, I would merge it and keep the ticket open. After the NSS bug is fixed, we can fix this properly.

frasertweedale:
FWIW, this one does not break CA-ful replica promotion.

tomaskrizek:
I've tested the following use cases:

- CA-less replica promotion domlvl1: *ldapssl running*; but the following behaviour is present: If `ipa-ca-install` is executed on replica, it finishes. But next `ipa-ca-install`, i.e. on master, will fail with CA did not start after 300 seconds. Relevant parts of pki and dirsrv logs:
```
[21/Dec/2016:12:43:46][localhost-startStop-1]: SSL handshake happened
Could not connect to LDAP server host vm-058-045.abc.idm.lab.eng.brq.redhat.com port 636 Error netscape.ldap.LDAPException: Authentication failed (48)
---
[21/Dec/2016:12:43:46.640540945 +0100] conn=4 fd=66 slot=66 SSL connection from 10.34.58.45 to 10.34.58.45
[21/Dec/2016:12:43:46.653170560 +0100] conn=4 TLS1.2 128-bit AES
[21/Dec/2016:12:43:46.665708312 +0100] conn=4 op=0 BIND dn="" method=sasl version=3 mech=EXTERNAL
[21/Dec/2016:12:43:46.667668986 +0100] conn=4 op=0 RESULT err=48 tag=97 nentries=0 etime=0
```
The same behavior is present when `ipa-ca-install` is first installed on master and then on replica. Basically, the second `ipa-ca-install` will fail. Running `ipa-certupdate` on the second server fixes the issue. This seems to be a separate issue, so I will file a bug for this.
- CA-full replica promotion domlvl1: *lpadssl running*
- CA-less replica installation domlvl0: *ldapssl running*
- CA-full replica installation domlvl0: *ldapssl running*

The fix seems to properly start the ldapssl both with CA-less and CA-full, therefore I'd accept this as a proper fix for the issue. Please address the minor improvement I suggested inline.

flo-renaud:
@tomaskrizek FYI, the current documentation states that ipa-certupdate must be run after ipa-ca-install (see https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Linux_Domain_Identity_Authentication_and_Policy_Guide/CA-less-to-CA.html).

mbasti-rh:
> @tomaskrizek FYI, the current documentation states that ipa-certupdate must be run after ipa-ca-install (see https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Linux_Domain_Identity_Authentication_and_Policy_Guide/CA-less-to-CA.html).

Bad UX, please open a RFE ticket for ipa-ca-install to execute certupdate automatically when needed

jcholast:
@mbasti-rh, `ipa-certupdate` has to be run on *all* systems in the domain after installing a CA. How do you propose we do that from `ipa-ca-install`? Anyway, the behavior @tomaskrizek is observing happens if you don't run `ipa-certupdate` *before* `ipa-ca-install` *on replica* and is caused by `ipa-ca-install` using local files rather than LDAP when looking for CA certificates.

pvoborni:
Running `ipa-certupdate` on all systems after `ipa-ca-install` is problematic. But we can at least make sure that `ipa-ca-install` on replica will get whatever is possible automatically at beginning (e.g. run equivalent of `ipa-certupdate`) and also that it have usable state after finishing `ipa-ca-install`, i.e., it will run IPA whatever IPA calls it needs and possible.  Anyway it is for other ticket. Maybe it already exists.

mbasti-rh:
@jcholast anyway I still see ways how to improve UX
- print big fat message to user at the end of ipa-ca-install to run ipa-certupdate everywhere when needed (instead of finding solution in the darkest corners in docs)
- automatically run it at least on the current replica

Still worth ticket IMO

@pvoborni +1

ipaserver/install/dsinstance.py
```diff
@@ -390,7 +390,9 @@ def create_replica(self, realm_name, master_fqdn, fqdn,
 
         self.step("creating DS keytab", self._request_service_keytab)
         if self.promote:
-            if self.ca_is_configured:
+            if self.pkcs12_info:
+                self.step("configuring ssl for ds instance", self.__enable_ssl)
```
Akasurde: Can you change case for `ssl` and `ds` ?

frasertweedale: There are other occurrences of lower-case "ssl" and "ds" within strings in this file.  (The string above is a copy-paste of another).  I created a new ticket: https://fedorahosted.org/freeipa/ticket/6586.  I will add a commit to this PR to address it.

ipaserver/install/dsinstance.py
```diff
@@ -392,6 +392,8 @@ def create_replica(self, realm_name, master_fqdn, fqdn,
         if self.promote:
             if self.ca_is_configured:
                 self.step("retrieving DS Certificate", self.__get_ds_cert)
+            elif self.pkcs12_info:
```
frasertweedale: @tomaskrizek this was implemented as requested.  Bump for review.

tomaskrizek:
I re-tested the most recent change in domlvl1. ldapssl is turned on both for CA-less replica install and CA-full replica install. 

I also created a ticket for the above mentioned behavior. https://fedorahosted.org/freeipa/ticket/6577

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/6f7d982fe2e2d2f042e85710b8d8d59167e5796f
https://fedorahosted.org/freeipa/changeset/a5fb5f2da1be158cde585a087aaf97eca6218dd7

mbasti-rh:
Please provide PR for ipa-4-4 too

frasertweedale:
ipa-4-4 PR: #371

PR #358, opened 2016-12-20
Use the tar Posix option for tarballs

simo5:
This is necessary to be able to successfully build archives in
environments controlled by an IPA domain which may have large uidNumbers
for user accounts.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -15,7 +15,7 @@ esac
 
 AC_CONFIG_HEADERS([config.h])
 
-AM_INIT_AUTOMAKE([foreign 1.9 tar-ustar])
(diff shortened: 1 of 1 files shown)


pspacek:
Please add a note *why* it is necessary. Something like "ustar format we used before could not handle files with > 65535" or so. This will help a lot with Git archaeology.

pspacek:
CondACK if the message is corrected.

simo5:
Amended

pspacek:
Thanks, ACK!

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/2bc01ec5b4a91a805912bdada429a91ab08ed196

PR #361, opened 2016-12-21
This PR implements a number of improvements for our Travis CI:

martbab:
* split out the test runner part into a standalone script, .travis.yml should
  now only define test matrix, set environment variables and process output 
  after failure

* mark the project as Python one, implement support for running builds using
  different python version (future-proofing against incoming Python2/3 CI) and
  cache job dependencies

* use separate job for pep8/linters. This shaves off ca 6-8 minutes from
  overall build time. You should get CI results in 26 min compared to previous
  33 min

### DIFF
--- a/.travis_run_task.sh
+++ b/.travis_run_task.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
-
+#
+# Copyright (C) 2017 FreeIPA Contributors see COPYING for license
+#
 # NOTE: this script is intended to run in Travis CI only
 
 PYTHON="/usr/bin/python${TRAVIS_PYTHON_VERSION}"
--- a/.travis_run_task.sh
+++ b/.travis_run_task.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 # NOTE: this script is intended to run in Travis CI only
-set -ev
 
+PYTHON="/usr/bin/python${TRAVIS_PYTHON_VERSION}"
 test_set=""
 developer_mode_opt="--developer-mode"
 
@@ -29,6 +29,7 @@ docker pull $TEST_RUNNER_IMAGE
 ipa-docker-test-runner -l $CI_RESULTS_LOG \
     -c $TEST_RUNNER_CONFIG \
     $developer_mode_opt \
+    --container-environment "PYTHON=$PYTHON" \
     --container-image $TEST_RUNNER_IMAGE \
     --git-repo $TRAVIS_BUILD_DIR \
     $TASK_TO_RUN $test_set
--- a/.travis_run_task.sh
+++ b/.travis_run_task.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# NOTE: this script is intended to run in Travis CI only
+set -ev
+
+test_set=""
+developer_mode_opt="--developer-mode"
+
+if [[ "$TASK_TO_RUN" == "lint" ]]
+then
+    if [[ "$TRAVIS_EVENT_TYPE" == "pull_request" ]]
+    then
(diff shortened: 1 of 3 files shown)


martbab:
Bump for review.

stlaz:
I assume the licence headers did not break the automember tests so this could be pushed.

Just a brief question: would it be reasonable to get the line number of "========= FAILURES =========" and tail the "$CI_TRAVIS_LOG" from the end to it?

martbab:
@stlaz thanks for an idea, that will actually make the log output much more useful. I will try to think about it.

stlaz:
@martbab My naive solution is to do something like
```bash
LINE=`grep -n -m 1 $CI_TRAVIS_LOG -e "=== FAILURES ===" | cut -d: -f1`
LINES=`wc -l $CI_TRAVIS_LOG`
tail -n `expr $LINES - $LINE` $CI_TRAVIS_LOG
```

martbab:
@stlaz I have implemented a simple log trimming which keeps only pytest failures if present. The original behavior is kept as a fallback for the case if the setup fails.

stlaz:
The change LGTM, ACK, we'll see how it works :)

.travis_run_task.sh
```diff
@@ -8,6 +8,17 @@ PYTHON="/usr/bin/python${TRAVIS_PYTHON_VERSION}"
 test_set=""
 developer_mode_opt="--developer-mode"
 
+function truncate_log_to_test_failures() {
+    # chop off everything in the CI_RESULTS_LOG preceding pytest error output
+    # if there are pytest errors in the log
+    $error_fail_regexp='\(=== ERRORS ===\)\|\(=== FAILURES ===\)'
```
stlaz: Should `$error_fail_regexp` miss "$"?

.travis_run_task.sh
```diff
@@ -8,6 +8,17 @@ PYTHON="/usr/bin/python${TRAVIS_PYTHON_VERSION}"
...
+
+    if grep -e "$error_fail_regexp" $CI_RESULTS_LOG
```
stlaz: Please add > /dev/null

.travis_run_task.sh
martbab: Oops sorry I forgot how to bash.

stlaz:
I have no more remarks on this, hopefully final ACK.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/d86cae7748a8a629c942f1eafc0a0267f2c9611e
https://fedorahosted.org/freeipa/changeset/758731088eee0294af59812dbd1976db89b9dda0
https://fedorahosted.org/freeipa/changeset/aff4e684e1f13d7da4248d17c0b8b2adf2e37033
https://fedorahosted.org/freeipa/changeset/1267e3e72305ee6bda0dd348ae1737b6f68f4371
https://fedorahosted.org/freeipa/changeset/149d86de14b00b73f625fefe73c2322a2fffac06
https://fedorahosted.org/freeipa/changeset/b8423492f5dce32183b34d718e4619fe3ca8bfef
https://fedorahosted.org/freeipa/changeset/b6216756f6c7a950e9bf2afe56a582dd8195c513
https://fedorahosted.org/freeipa/changeset/f48d6fc168253209bed3f1dd5a543f15d1f54669
https://fedorahosted.org/freeipa/changeset/4abd3f554a436e6446ba59c75c09fb0ff8b7fe4a
https://fedorahosted.org/freeipa/changeset/0ef55a91ef9c591cee3a7e1ff0e391cdc32423c3

PR #352, opened 2016-12-19
Clarify meaning of --domain and --realm in installers

pspacek:
Man pages need bigger overhaul. Take this as hot-fix for FAQ.

https://fedorahosted.org/freeipa/ticket/6574

PR #362, opened 2017-01-02
Clarify meaning of --domain and --realm in installers

stlaz:
This is my take on original https://github.com/freeipa/freeipa/pull/352. I hope I fixed all the mentioned issues + I added some missing articles.


frasertweedale:
All of my comments from #352 were addressed.

@stlaz you were the only other person to review #352 and request changes, so I assume
you have addressed those too, in which case: ACK.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/25a6ddcce8e7b9effaf19431c421dc5b3497fa22

PR #363, opened 2017-01-03
ipaclient: schema cache: Handle malformed server info data gracefully

dkupka:
As a part of CLI schema cache some data about each previously contacted server
are stored in simple JSON file. The file may get corrupted and became
undecodable for various reasons (parallel access, file system error,
tampering). Since the data are not necessary we should just warn an continue.

https://fedorahosted.org/freeipa/ticket/6578

### DIFF
--- a/ipaclient/remote_plugins/__init__.py
+++ b/ipaclient/remote_plugins/__init__.py
@@ -41,8 +41,14 @@ class ServerInfo(collections.MutableMapping):
         try:
             with open(self._path, 'r') as sc:
                 self._dict = json.load(sc)
-        except EnvironmentError as e:
-            if e.errno != errno.ENOENT:
+        except Exception as e:
+            if (isinstance(e, EnvironmentError) and
+                    e.errno == errno.ENOENT):  # pylint: disable=no-member
+                # ignore non-existent file, this happens when the cache was
+                # erased or the server is contacted for the first time
+                pass
+            else:
+                # warn that the file is unreadable, probably corrupted
                 logger.warning('Failed to read server info: {}'.format(e))
 
     def _write(self):


ipaclient/remote_plugins/__init__.py
```diff
@@ -40,9 +40,14 @@ def __init__(self, api):
     def _read(self):
         try:
             with open(self._path, 'r') as sc:
-                self._dict = json.load(sc)
-        except EnvironmentError as e:
-            if e.errno != errno.ENOENT:
+                self._dict = json.loads(sc)
```
HonzaCholasta: `sc` is a file object, so `load()` was correct, wasn't it?

dkupka: Yes, stupid me. I was curious how the file looks like so I had there:
data = sc.read()
json.loads(data)
and printed the data when the error occurred. Obviously, I haven't changed it back properly.
Thanks!

HonzaCholasta:
Can the corruption also happen for schema files?

dkupka:
@HonzaCholasta Yes, it can happen but it is already handles correctly: https://github.com/freeipa/freeipa/blob/master/ipaclient/remote_plugins/schema.py#L387

ipaclient/remote_plugins/__init__.py
```diff
@@ -41,8 +41,13 @@ def _read(self):
         try:
             with open(self._path, 'r') as sc:
                 self._dict = json.load(sc)
-        except EnvironmentError as e:
-            if e.errno != errno.ENOENT:
+        except (EnvironmentError, ValueError) as e:
```
HonzaCholasta: I guess you should 'except Exception as e:' here, same as in https://github.com/freeipa/freeipa/blob/master/ipaclient/remote_plugins/schema.py#L387

HonzaCholasta:
@dkupka, ok, see inline comment.

ipaclient/remote_plugins/__init__.py
dkupka: Ok, I will update it for the sake of consistency. But there will be no difference in behaviour.

HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/d15ccde20fcc97a597180255ee9f5eb38caa206c

PR #364, opened 2017-01-03
Client-only builds with --disable-server

tiran:
https://fedorahosted.org/freeipa/ticket/6517

Simple approach to reduce dependencies for client only builds. I simply wrapped all daemon, init and install dependencies in AM_COND_IF(). I'm open to suggestions. Let the bike shedding begin!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2,12 +2,16 @@ ACLOCAL_AMFLAGS = -I m4
 
 if ENABLE_SERVER
     SERVER_SUBDIRS = daemons init install ipaserver
-else
-    SERVER_SUBDIRS =
 endif
+
+if WITH_IPATESTS
+    IPATESTS_SUBDIRS = ipatests
+endif
+
 IPACLIENT_SUBDIRS = ipaclient ipalib ipapython
+
 SUBDIRS = asn1 util client contrib po \
-	$(IPACLIENT_SUBDIRS) ipaplatform ipatests $(SERVER_SUBDIRS)
+	$(IPACLIENT_SUBDIRS) ipaplatform $(IPATESTS_SUBDIRS) $(SERVER_SUBDIRS)
 
 
 MOSTLYCLEANFILES = ipasetup.pyc ipasetup.pyo \
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,14 @@
 ACLOCAL_AMFLAGS = -I m4
 
+if ENABLE_SERVER
+    SERVER_SUBDIRS = daemons init install ipaserver
+else
+    SERVER_SUBDIRS =
+endif
 IPACLIENT_SUBDIRS = ipaclient ipalib ipapython
-SUBDIRS = asn1 util client contrib daemons init install $(IPACLIENT_SUBDIRS) ipaplatform ipaserver ipatests po
+SUBDIRS = asn1 util client contrib po \
+	$(IPACLIENT_SUBDIRS) ipaplatform ipatests $(SERVER_SUBDIRS)
+
 
 MOSTLYCLEANFILES = ipasetup.pyc ipasetup.pyo \
 		   ignore_import_errors.pyc ignore_import_errors.pyo \
--- a/configure.ac
+++ b/configure.ac
@@ -27,13 +27,20 @@ AC_HEADER_STDC
 PKG_PROG_PKG_CONFIG
 
 AC_ARG_ENABLE([server],
-[  --disable-server        Disable server support],
-[case "${enableval}" in
-  yes) enable_server=true ;;
-  no)  enable_server=false ;;
-  *) AC_MSG_ERROR([bad value ${enableval} for --disable-server]) ;;
-esac],[enable_server=true])
-AM_CONDITIONAL([ENABLE_SERVER], [test x$enable_server = xtrue])
+    [AC_HELP_STRING([--disable-server], [Disable server support])],
+    [case "${enableval}" in
+         yes) enable_server=yes ;;
+          no) enable_server=no ;;
+           *) AC_MSG_ERROR([bad value ${enableval} for --disable-server]) ;;
+     esac],
+    [enable_server=yes])
+AM_CONDITIONAL([ENABLE_SERVER], [test x$enable_server = xyes])
+
+AC_ARG_WITH([ipatests],
+    [AC_HELP_STRING([--without-ipatests], [Build without ipatests])],
+    [with_ipatests=${withval}],
+    [with_ipatests=yes])
+AM_CONDITIONAL([WITH_IPATESTS], [test x"$with_ipatests" = xyes])
 
 AM_CONDITIONAL([HAVE_GCC], [test "$ac_cv_prog_gcc" = yes])
 
@@ -515,4 +522,11 @@ AM_COND_IF([ENABLE_SERVER], [
 ], [
     echo "\
         build mode:               client only"
+])
+AM_COND_IF([WITH_IPATESTS], [
+    echo "\
+        with ipatests:            yes"
+], [
+    echo "\
+        with ipatests:            no"
 ])
--- a/configure.ac
+++ b/configure.ac
@@ -24,6 +24,17 @@ LT_INIT
 
 AC_HEADER_STDC
 
+PKG_PROG_PKG_CONFIG
+
+AC_ARG_ENABLE([server],
+[  --disable-server        Disable server support],
+[case "${enableval}" in
+  yes) enable_server=true ;;
+  no)  enable_server=false ;;
+  *) AC_MSG_ERROR([bad value ${enableval} for --disable-server]) ;;
+esac],[enable_server=true])
+AM_CONDITIONAL([ENABLE_SERVER], [test x$enable_server = xtrue])
+
 AM_CONDITIONAL([HAVE_GCC], [test "$ac_cv_prog_gcc" = yes])
 
 dnl ---------------------------------------------------------------------------
@@ -33,37 +44,10 @@ PKG_CHECK_MODULES([NSPR], [nspr])
 PKG_CHECK_MODULES([NSS], [nss])
 
 dnl ---------------------------------------------------------------------------
-dnl - Check for DS slapi plugin
-dnl ---------------------------------------------------------------------------
-
-# Need to hack CPPFLAGS to be able to correctly detetct slapi-plugin.h
-SAVE_CPPFLAGS=$CPPFLAGS
-CPPFLAGS=$NSPR_CFLAGS
-AC_CHECK_HEADER(dirsrv/slapi-plugin.h)
-if test "x$ac_cv_header_dirsrv_slapi-plugin_h" = "xno" ; then
-	AC_MSG_ERROR([Required 389-ds header not available (389-ds-base-devel)])
-fi
-AC_CHECK_HEADER(dirsrv/repl-session-plugin.h)
-if test "x$ac_cv_header_dirsrv_repl_session_plugin_h" = "xno" ; then
-	AC_MSG_ERROR([Required 389-ds header not available (389-ds-base-devel)])
-fi
-CPPFLAGS=$SAVE_CPPFLAGS
-
-if test "x$ac_cv_header_dirsrv_slapi_plugin_h" = "xno" ; then
-	AC_MSG_ERROR([Required DS slapi plugin header not available (fedora-ds-base-devel)])
-fi
-
-dnl ---------------------------------------------------------------------------
 dnl - Check for KRB5
 dnl ---------------------------------------------------------------------------
 
 PKG_CHECK_MODULES([KRB5], [krb5])
-AC_CHECK_HEADER(krad.h, [], [AC_MSG_ERROR([krad.h not found])])
-AC_CHECK_LIB(krad, main, [], [AC_MSG_ERROR([libkrad not found])])
-KRAD_LIBS="-lkrad"
-krb5rundir="${localstatedir}/run/krb5kdc"
-AC_SUBST(KRAD_LIBS)
-AC_SUBST(krb5rundir)
 
 AC_CHECK_HEADER(kdb.h, [], [AC_MSG_ERROR([kdb.h not found])])
 AC_CHECK_MEMBER(
@@ -105,11 +89,6 @@ dnl ---------------------------------------------------------------------------
 PKG_CHECK_MODULES([CRYPTO], [libcrypto])
 
 dnl ---------------------------------------------------------------------------
-dnl - Check for UUID library
-dnl ---------------------------------------------------------------------------
-PKG_CHECK_MODULES([UUID], [uuid])
-
-dnl ---------------------------------------------------------------------------
 dnl - Check for Python
 dnl ---------------------------------------------------------------------------
 
@@ -120,69 +99,6 @@ AM_PATH_PYTHON(2.7)
 if test "x$PYTHON" = "x" ; then
   AC_MSG_ERROR([Python not found])
 fi
-
-dnl ---------------------------------------------------------------------------
-dnl Check for ndr_krb5pac and other samba libraries
-dnl ---------------------------------------------------------------------------
-
-PKG_PROG_PKG_CONFIG()
-PKG_CHECK_MODULES([TALLOC], [talloc])
-PKG_CHECK_MODULES([TEVENT], [tevent])
-PKG_CHECK_MODULES([NDRPAC], [ndr_krb5pac])
-PKG_CHECK_MODULES([NDRNBT], [ndr_nbt])
-PKG_CHECK_MODULES([NDR], [ndr])
-PKG_CHECK_MODULES([SAMBAUTIL], [samba-util])
-SAMBA40EXTRA_LIBPATH="-L`$PKG_CONFIG --variable=libdir samba-util`/samba -Wl,-rpath=`$PKG_CONFIG --variable=libdir samba-util`/samba"
-AC_SUBST(SAMBA40EXTRA_LIBPATH)
-
-bck_cflags="$CFLAGS"
-CFLAGS="$NDRPAC_CFLAGS"
-AC_CHECK_MEMBER(
-    [struct PAC_DOMAIN_GROUP_MEMBERSHIP.domain_sid],
-    [AC_DEFINE([HAVE_STRUCT_PAC_DOMAIN_GROUP_MEMBERSHIP], [1],
-               [struct PAC_DOMAIN_GROUP_MEMBERSHIP is available.])],
-    [AC_MSG_NOTICE([struct PAC_DOMAIN_GROUP_MEMBERSHIP is not available])],
-                 [[#include <ndr.h>
-                   #include <gen_ndr/krb5pac.h>]])
-
-CFLAGS="$bck_cflags"
-
-LIBPDB_NAME=""
-AC_CHECK_LIB([samba-passdb],
-             [make_pdb_method],
-             [LIBPDB_NAME="samba-passdb"; HAVE_LIBPDB=1],
-             [LIBPDB_NAME="pdb"],
-             [$SAMBA40EXTRA_LIBPATH])
-
-if test "x$LIB_PDB_NAME" = "xpdb" ; then
-  AC_CHECK_LIB([$LIBPDB_NAME],
-               [make_pdb_method],
-               [HAVE_LIBPDB=1],
-               [AC_MSG_ERROR([Neither libpdb nor libsamba-passdb does have make_pdb_method])],
-               [$SAMBA40EXTRA_LIBPATH])
-fi
-
-AC_SUBST(LIBPDB_NAME)
-
-AC_CHECK_LIB([$LIBPDB_NAME],[pdb_enum_upn_suffixes],
-             [AC_DEFINE([HAVE_PDB_ENUM_UPN_SUFFIXES], [1], [Ability to enumerate UPN suffixes])],
-             [AC_MSG_WARN([libpdb does not have pdb_enum_upn_suffixes, no support for realm domains in ipasam])],
-             [$SAMBA40EXTRA_LIBPATH])
-
-dnl ---------------------------------------------------------------------------
-dnl Check for libunistring
-dnl ---------------------------------------------------------------------------
-AC_CHECK_HEADERS([unicase.h],,AC_MSG_ERROR([Could not find unicase.h]))
-AC_CHECK_LIB([unistring],
-             [ulc_casecmp],
-             [UNISTRING_LIBS="-lunistring"],
-             [AC_MSG_ERROR([libunistring does not have ulc_casecmp])])
-AC_SUBST(UNISTRING_LIBS)
-
-dnl ---------------------------------------------------------------------------
-dnl Check for libverto
-dnl ---------------------------------------------------------------------------
-PKG_CHECK_MODULES([LIBVERTO], [libverto])
 
 dnl ---------------------------------------------------------------------------
 dnl - Check for cmocka unit test framework http://cmocka.cryptomilk.org/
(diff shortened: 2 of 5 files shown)


tomaskrizek:
Could you please extend [V4/Build system refactoring](http://www.freeipa.org/page/V4/Build_system_refactoring) to include steps describing how to perform client-only build?

Also, is this supposed to build the freeipa-client rpm package? I wasn't able to build the rpm (``` *** No rule to make target 'distdir'.  Stop.```).

Or does it only support make + make install? When I tried to do this it also produced ipaserver and ipatests directories. It was probably a misconfiguration on my side.

Please provide instructions on how to do a proper client-only build.

tiran:
The PR only affects ```make install```, Python packaging and integration efforts. The goal is to reduce the amount of necessary build dependencies for installation and packaging of Python wheels or for developers that are only interested to build ipa-client locally.

At the moment it is not possible to build FreeIPA without Samba, talloc, tevent, 389 DS, systemd and a couple of more packages. The dependency tree is rather heavy. These dependencies are not relevant for clients, though.

I haven't touched RPM builds deliberately. RPM spec can be adjusted in a subsequent PR. It's not relevant for me.

tomaskrizek:
The extra dependencies are indeed not necessary with this change.

However, `make install` produces  directories like `/usr/lib/python2.7/site-packages/ipaserver`, `/usr/lib/python2.7/site-packages/ipatests`, ... I don't think these should be present when doing a client-only build.

tiran:
Nit-pick: A build does not produce any files outside the build environment. Of course **make install** produces the files -- unless you change the prefix with ```./configure --prefix```.

tomaskrizek:
I'm not really experienced with autotools, so I do not want to ack this PR without someone else taking a look. I'm also not sure about the best practices in this area. Perhaps @lslebodn could share his opinion on this change?

lslebodn:
I think it would be simpler to read the code without to many `AM_COND_IF()`. IMHO it would be simpler to move C-related part to separate file (e.g. `server_daemons.m4`) an conditionally include  the file with `m4_include`.

I checked it very briefly; I might miss something. How do you want to handle python server part?

tiran:
@lslebodn I like the idea to move the server related header and lib detection to a separate m4 file.

In server-less mode, I plan to ignore the Python server part completely.

lslebodn:
I fine with ignoring python related parts; but it should be documented. But you might ask other freeIPA developers. (maybe on freeipa-devel)

tiran:
@lslebodn I think you misunderstood me. The PR adds a new build flavor that ignores and skips the any server-related steps of the build process. The fact that this PR ignores the server part is actually a a **feature**. :)

The last version of the patch now skips ```ipaserver``` subdir completely, too.

lslebodn:
`make dist` failed if configure was executed with `--disable-server`

```
make[3]: Leaving directory '/workdir/freeipa/po'
make[2]: Leaving directory '/workdir/freeipa/po'
 (cd daemons && make  top_distdir=../freeipa-4.4.90.dev201702081623+git9da17b545 distdir=../freeipa-4.4.90.dev201702081623+git9da17b545/daemons \
     am__remove_distdir=: am__skip_length_check=: am__skip_mode_fix=: distdir)
make[2]: Entering directory '/workdir/freeipa/daemons'
make[2]: *** No rule to make target 'distdir'.  Stop.
make[2]: Leaving directory '/workdir/freeipa/daemons'
make[1]: *** [Makefile:707: distdir] Error 1
make[1]: Leaving directory '/workdir/freeipa'
make: *** [Makefile:806: dist] Error 2
```

Do we care?

Makefile.am
```diff
@@ -1,7 +1,11 @@
 ACLOCAL_AMFLAGS = -I m4
 
 IPACLIENT_SUBDIRS = ipaclient ipalib ipapython
-SUBDIRS = asn1 util client contrib daemons init install $(IPACLIENT_SUBDIRS) ipaplatform ipaserver ipatests po
+SUBDIRS = asn1 util client contrib $(IPACLIENT_SUBDIRS) ipaplatform ipatests po
+
+if ENABLE_SERVER
+    SUBDIRS += daemons init install ipaserver
+endif
```
lslebodn: I think you can move `ipatests` as well because they require server.

lslebodn:
See inline comment and issue above; otherwise LGTM.

Makefile.am
tiran: Only some parts of the test suite require ipaserver. I have a branch that can execute just the client tests with tox.

lslebodn: `ipatests/test_xmlrpc/` and `ipatests/test_integration/` requires them

tomaskrizek:
@tiran Do you need `make dist` for anything?

I'm not aware of any plans to release client-only IPA sources, so I don't think it's needed.

tiran:
@tomaskrizek @lslebodn Although I don't need ```make dist```, you made me aware of a bug in ```Makefile.am```. automake and ```+=``` do not mix well. I moved the list into a new var ```SERVER_SUBDIRS```.

Makefile.am
```diff
@@ -1,7 +1,14 @@
...
 
+if ENABLE_SERVER
+    SERVER_SUBDIRS = daemons init install ipaserver
+else
+    SERVER_SUBDIRS =
+endif
+
 IPACLIENT_SUBDIRS = ipaclient ipalib ipapython
-SUBDIRS = asn1 util client contrib daemons init install $(IPACLIENT_SUBDIRS) ipaplatform ipaserver ipatests po
+SUBDIRS = asn1 util client contrib $(IPACLIENT_SUBDIRS) ipaplatform ipatests po $(SERVER_SUBDIR)
```
tomaskrizek: Probably a typo here, missing `S` at the end of `$SERVER_SUBDIR`.

configure.ac
```diff
@@ -500,56 +411,6 @@ AC_CONFIG_FILES([
...
-    install/ui/build/Makefile
-    install/ui/build/dojo/Makefile
-    install/ui/build/freeipa/Makefile
-    install/tools/Makefile
-    install/tools/man/Makefile
-    install/updates/Makefile
-    install/restart_scripts/Makefile
-    install/wsgi/Makefile
-    install/oddjob/Makefile
     ipaclient/Makefile
```
lslebodn: This change caused me troubles after fixing typo in `SERVER_SUBDIR`.
Why did you move them?

lslebodn:
```
  autoreconf -if
  ./configure --disable-server
  make rpms V=0
  make dist
```

tiran: I moved the makefiles because they use autoconf vars that are not available in ```--disable-server``` mode. I see two ways to solve the problem:

* still generate all makefiles although some are potentially broken
* set ```DIST_SUBDIRS = $(SUBDIRS)```

https://www.gnu.org/software/automake/manual/html_node/SUBDIRS-vs-DIST_005fSUBDIRS.html

tomaskrizek:
Server build works now, but there's still the `make dist` issue discussed above.

tiran:
I followed @lslebodn advice and changed the PR a bit. I now generate all Makefiles again to fix the ```make dist``` issue. Some of the Makefile are not working correctly because some vars are declared empty (e.g. header locations, libs and so on). Since they are not included in ```SUBDIRS```, they are not used in ```make```. ```make dist``` uses ```DIST_SUBDIRS``` and does not depend on the missing vars.

pvoborni:
I still think that this use case needs to be documented in http://www.freeipa.org/page/V4/Build_system_refactoring#How_to_Use . 

IMHO `make dist` can fail with --disable-server. Use case for `make dist` is releasing and there is no point to do release without server bitw. But make sure to document it and test that `make dist` works without it ;) .

lslebodn:
On (10/02/17 06:11), Petr Vobornik wrote:
>I still think that this use case needs to be documented in http://www.freeipa.org/page/V4/Build_system_refactoring#How_to_Use . 
>
>IMHO `make dist` can fail with --disable-server. Use case for `make dist` is releasing and there is no point to do release without server bitw. But make sure to document it and test that `make dist` works without it ;) .
>


Petr FYI: make dist should include all files into tarball
even though configure was invoked with parameter --disable-server.

LS

tiran:
In the latest version, ```--disable-server``` does not affect ```make dist```. It only changes the components that are built by ```make``` and installed by ```make install```.

```--disable-server ``` is already documented at http://www.freeipa.org/page/V4/Build_system_refactoring#Configuration . After the PR has landed, I'll update the table.

**Comment by lslebodn** [comment]
It works quite good with the following change

```
diff --git a/Makefile.am b/Makefile.am
index 311f6121f..13e5a87b0 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,13 +1,13 @@
 ACLOCAL_AMFLAGS = -I m4
 
 if ENABLE_SERVER
-    SERVER_SUBDIRS = daemons init install ipaserver
+    SERVER_SUBDIRS = daemons init install ipaserver ipatests
 else
     SERVER_SUBDIRS =
 endif
 IPACLIENT_SUBDIRS = ipaclient ipalib ipapython
 SUBDIRS = asn1 util client contrib po \
-       $(IPACLIENT_SUBDIRS) ipaplatform ipatests $(SERVER_SUBDIRS)
+       $(IPACLIENT_SUBDIRS) ipaplatform $(SERVER_SUBDIRS)
 
 
 MOSTLYCLEANFILES = ipasetup.pyc ipasetup.pyo \
```

**Comment by tiran** [comment]
@lslebodn it works even better without your proposed changes. Parts ```ipatests``` work fine for ```--disable-server``` builds. I need the package to run tests.

**Comment by lslebodn** [comment]
On (13/02/17 01:25), Christian Heimes wrote:
>@lslebodn it works even better without your proposed changes. Parts ```ipatests``` work fine for ```--disable-server``` builds. I need the package to run tests.
>

The old version (4.4) of `CLIENT_ONLY` build did not package
ipatests.

Could you describe a reason/use-case for installing `ipatests`
without server?

LS

**Comment by tiran** [comment]
Packaging is a different issue. The PR does not provide RPM packaging for client-only build. It merely implements configuration and building without server components.

For client-only builds I need ipatests to run part of the test suite to verify client code. Test suites ```test_ipapython, test_ipalib, test_pkcs10``` without ```test_ipalib.test_rpc``` work without ```ipaserver```.

**Comment by lslebodn** [comment]
On (13/02/17 03:08), Christian Heimes wrote:
>Packaging is a different issue. The PR does not provide RPM packaging for client-only build. It merely implements configuration and building without server components.
>

I mentioned old version of `CLIENT_ONLY` build because I consider it
as a referential implementation. And `ipa tests` were not installed in 4.4
for client only build.


>For client-only builds I need ipatests to run part of the test suite to verify client code. Test suites ```test_ipapython, test_ipalib, test_pkcs10``` without ```test_ipalib.test_rpc``` work without ```ipaserver```.
>

I expected a little bit more details.

Do you need to run `make install` and then run tests in installed directory?
Or how do you want to "run part of the test suite".

Because if you needn't run "make install" for your use-case then
my proposed patch would work.

BTW `ipatests` will still be part of tarball and/or git. You can run them
even thought they will not be installed with `make install`

LS

**Comment by tiran** [comment]
Two reasons

1. ```make install```
2. I need ipatests to be part of the build process in order to get a Python package for tox later.

**Comment by lslebodn** [comment]
On (13/02/17 03:56), Christian Heimes wrote:
>Two reasons
>
>1. ```make install```
>2. I need ipatests to be part of the build process in order to get a Python package for tox later.
>

OK, thank you for explanation.

Then we should install just tests from directory `ipatests`
which does not require daemon for execution.

LS

**Comment by tiran** [comment]
No, the test runner should either detect missing packages and skip tests automatically, or should grow an option to load and execute client tests only. It's a separate issue.

**Comment by lslebodn** [comment]
On (13/02/17 04:32), Christian Heimes wrote:
>No, the test runner should either detect missing packages and skip tests automatically, or should grow an option to load and execute client tests only. It's a separate issue.
>

I have a different opinion. It is not a separate issue.
For me, the name of configure option is crystall clear.
It should not install anything related to daemon part; even thought it is test.

Maybe we can add another option to install tests (--with-tests?? +default yes)
It would work for your use-case and still allow old `CLIENT_ONLY` build
(equivalent to 4.4)

Or you can propose another compromise.

LS

**Comment by tiran** [comment]
I'm following the development principals of **minimum viable product**. This PR solves a critical use case for me. With the PR I can build FreeIPA client packages in a lean and clean build container. Without the ```--disable-server``` flag I'm forced to bloat my build env with lots of additional dependencies and then throw away all the extra stuff.

My changes don't solve https://fedorahosted.org/freeipa/ticket/6517 to its full extend. The PR provides enough of https://fedorahosted.org/freeipa/ticket/6517 to enable me to finish some time critical as soon as possible. RPM packaging changes and ipatests improvements for client-only builds can be implemented another time. I consider these changes sugar coating (aka stretch goals).

**Comment by lslebodn** [comment]
On (13/02/17 05:05), Christian Heimes wrote:
>I'm following the development principals of **minimum viable product**. This PR solves a critical use case for me. With the PR I can build FreeIPA client packages in a lean and clean build container. Without the ```--disable-server``` flag I'm forced to bloat my build env with lots of additional dependencies and then throw away all the extra stuff.
>

My comments are about semantic of this option.
`--disable-server` should disable all parts which depends on server.

I know that your use case is a little bit different but I do not like
misusing of `--disable-server` for different use-cases (from semantic POV)

That's the reason why I proposed compromise/alternative solution
for installing `ipatests` which needn't be tight together
with `--disable-server`.

>My changes don't solve https://fedorahosted.org/freeipa/ticket/6517 to its full extend. The PR provides enough of https://fedorahosted.org/freeipa/ticket/6517 to enable me to finish some time critical as soon as possible. RPM packaging changes and ipatests improvements for client-only builds can be implemented another time. I consider these changes sugar coating (aka stretch goals).
>

One more time; it will be solved with my proposed change to `ipatests`
+ small tweak to spec file (due to python2/3 changes)

That is exactly way how I tested it. A little bit hacky way but works
for testing: https://paste.fedoraproject.org/556868/48699519

LS

**Comment by tiran** [comment]
I'm following a different design and development philosophy. In my experience an iterative approach with small, incremental improvements is often better and faster than striving for 100% perfect PRs. Large and feature complete PRs take more time than evolutionary steps.

Your object regarding semantics is valid for the ticket, but not necessarily valid for this PR as this PR only addresses a part of the problem. I don't dispute that your proposed changes to the spec file are necessary. However I argue for a separate PR. I'm not an expert in RPM packaging and I'd rather let somebody else figure out the appropriate way to deal with client-only packaging. ipatests is yet another problem that should be solved in a third PR. Ticket https://fedorahosted.org/freeipa/ticket/6517 does not, in fact it should not be solved in one PR.

Please review this PR under three viewpoints:

* Does it contribute to resolving ticket https://fedorahosted.org/freeipa/ticket/6517 ?
* Does it enable future changes to solve the ticket?
* Does it break any code or feature that is currently present? [1]

[1] client-only packaging is currently not available

**Comment by lslebodn** [comment]
On (14/02/17 02:29), Christian Heimes wrote:
>I'm following a different design and development philosophy. In my experience an iterative approach with small, incremental improvements is often better and faster than striving for 100% perfect PRs. Large and feature complete PRs take more time than evolutionary steps.
>

I have never wrote anythig against this philosophy.

All small chages can make sense from semantical point of view.
Misussing names/options for different use-case just create a big mess
and confuse other people.

>Please review this PR under three viewpoints:
>
>* Does it contribute to resolving ticket https://fedorahosted.org/freeipa/ticket/6517 ?

client only build and --disable-server is the same thing
(at least from "make install" POV)
I have never required changes to spec file.

>* Does it enable future changes to solve the ticket?

If you will not install ipatests (if there is a way to not
install ipatest) then it will enable future changes to solve the ticket.

Because solving ticket6517 would be just writing right spec file.

ATM it does not enable future changes to solve the ticket.

>* Does it break any code or feature that is currently present? [1]

Yes, it install server related options even though they should not be installed

Summary: You should realize that the name of PR is
"Client-only builds with --disable-server" and your use-case is not pure
client only build.

LS

**Comment by tiran** [comment]
You are missing the point. Obviously tests are an important part of building. I can't test the client bits when ipatests is not available.

Let's do small, incremental improvements. I need the client-only builds ASAP (EOW, next week tops) for some container stuff. Client-only RPMs can be implemented later in beta phase for 4.5.

**Comment by lslebodn** [comment]
On (16/02/17 00:57), Christian Heimes wrote:
>You are missing the point. Obviously tests are an important part of building. I can't test the client bits when ipatests is not available.
>

You missed that the name of this PR is
"Client-only builds with --disable-server"

So this PR *MUST* implement client-only build.

Your use-case is different and you need to realize that
freeIPA does not have a *unit* test for client bits.
All client parts are tested by integration tests
which require server.

>Let's do small, incremental improvements. I need the client-only builds ASAP (EOW, next week tops) for some container stuff. Client-only RPMs can be implemented later in beta phase for 4.5.
>

I have never requested anything related to "Client-only RPMs"
I always mentioned "make install"

Let's do the correct change from semantic POV.

If you want to install tests with client-only build
then please add new configure time option for this purpose
And do not misuse option "--disable-server"

Misusing options is a bad/hacky approch
and we need to clean hacks from freeIPA and not create new one.

LS

**Comment by tiran** [comment]
Lukas, you are wasting both my and your precious time with a needless bike-shedding discussion about semantics. The ```--disable-server``` option skips all parts of the build process that are only relevant for server and not relevant for client. ```ipatests``` is relevant for both, therefore it stays.

I told you that I need  ```ipatests``` to run tests as part of my build process. It is not yet part of the upstream FreeIPA build process. I also told you that I will provide another PR that will take care of it and add client-only tests. This PR acts as a foundation for both my container build processes and the future PR.

**Comment by pvoborni** [comment]
I think it is OK to keep the behavior of the patch and go with it provided that the behavior is properly document in the design page after push. 

The only reason to block it would be that it would be difficult to change it later or if it breaks any existing functionality. But AFAIK it is not the case here.

**Comment by tomaskrizek** [comment]
ACK, I'm fine with pushing this PR.

`make install` does install ipatests for client-only build, other server-related packages are omitted.

Server build work like before and isn't affected by this PR.

**Comment by lslebodn** [comment]
On (16/02/17 02:30), Christian Heimes wrote:
>Lukas, you are wasting both my and your precious time with a needless bike-shedding discussion about semantics. The ```--disable-server``` option skips all parts of the build process that are only relevant for server and not relevant for client. ```ipatests``` is relevant for both, therefore it stays.
>

It's not bikeshadig. You are adding HACKs to freeipa build system.

If this PR will be pushed with cuurent state then I will need to send PR
to fix client-only build and it will not install ipatests with
--disable-server.

I proposed you a compromise few times.
Thank you very much for ignoring it.

LS

**Comment by tiran** [comment]
I have one very important question: Without ipatests, how are you going to automatically test client-only builds?

**Comment by martbab** [comment]
Well you won't, and they were not tested automatically before AFAIK (checked with FreeIPA 4.2.0 sources). We can agree that they *should* be, but since this eventuality was never considered before the build refactoring happened I guess we should first restore the original behavior, and then decide whether the client-only build should run those few unit tests during build.

**Comment by tiran** [comment]
I have to validate client-only Python wheel packages any way. We can just reuse the same infrastructure to test client-only RPMs later. See PR #475.

By the way, validation of Python wheels is the main reason I must have ```ipatests``` in the build.

**Comment by pvoborni** [comment]
I still fail to see why we should care about `make dist` with `configure --disable-server` this is not a combination of options which should be used together, there is no point in it except theoretical exercise.

> then I will need to send PR to fix client-only build and it will not install ipatests with --disable-server.

Why?

**Comment by lslebodn** [comment]
On (17/02/17 01:12), Petr Vobornik wrote:
>I still fail to see why we should care about `make dist` with `configure --disable-server` this is not a combination of options which should be used together, there is no point in it except theoretical exercise.
>

`make dist` works

>> then I will need to send PR to fix client-only build and it will not install ipatests with --disable-server.
>
>Why?

The main problem with `client-only` build and `ipatests`
is that there are not test for it.

* all integration test assume that server will be installed first
  and not that server is already available somewhere.
* unit test should be run as part of build (e.g. `make check`)

So there is not any reason to install `ipatests` for client only build.

Client only build was tested just in downstream :-(

LS

**Comment by lslebodn** [comment]
It looks like @tiran does not want to move this PR forward.

So here is a link to patch which does not misuse client-only
build and add configure time option to install ipatests.

https://paste.fedoraproject.org/paste/~u7iDljnMTvinxmLFoc25V5M1UNdIGYhyRLivL9gydE=/

As part of this work I found out that 4 dependencies can be simply
moved to server only build: `libuuid`, `DIRSRV`, `libsss_idmap`,
`libsss_nss_idmap`.

Result => Client only build require less C-dependencies.

LS

**Comment by tiran** [comment]
Thanks for your contribution. I added your patch to my PR. On my system I ran into a minor issue. Some C99 types like ```uint8_t``` were not defined and I had to include ```stdint.h```.

By the way I'm just going to ignore your snidely and snarky comment.

**Comment by tiran** [comment]
Now ```-without-ipatests``` argument for @lslebodn

**Comment by HonzaCholasta** [comment]
@tiran, not just for @lslebodn, `--without-ipatests` will be very useful to me for RHEL and Arch Linux packaging as well 😉.

**Comment by lslebodn** [comment]
>Thanks for your contribution. I added your patch to my PR. On my system I ran into a minor issue. >Some C99 types like uint8_t were not defined and I had to include stdint.h.

This change is not enough; there is still warning:
```
ipa_pwd_ntlm.c: In function 'encode_nt_key':
ipa_pwd_ntlm.c:58:5: warning: implicit declaration of function 'strlen' [-Wimplicit-function-declaration]
     il = strlen(newPasswd);
     ^
ipa_pwd_ntlm.c:58:10: warning: incompatible implicit declaration of built-in function 'strlen'
     il = strlen(newPasswd);
          ^
```

>By the way I'm just going to ignore your snidely and snarky comment.

No problem. I am going to forget that my proposal for compromise was ignored for 12 days.

The latest version is a small improvement; but there are still problems small issues because this PR was not create with intention to use tox. I know you are busy. So I wrote client-only implementation from scratch.

This PR is superseded by #494

tomaskrizek:
The PR works and the `--without-ipatests` option omits the ipatests directory.

However, #494 doesn't install extra dependencies with `mock --without=server`.

simo5:
So this is the reasoning and why I am approving this PR and not #494.

When you build all components, including server bits, tests are installed, therefore when we build just client bits tets that are relevant to client bits also need to be installed for consistency.

Any switch should default to the same behavior regardless of whether server build is enabled. It is confusing if the --with[out]-[ipa]tests switch changes default based on a different switch passed to configure.

As far as I understand this PR maintains the same default for either server or client only builds, so it gets my approval.

lslebodn:
On (22/02/17 06:16), Simo Sorce wrote:
>So this is the reasoning and why I am approving this PR and not #494.
>
>When you build all components, including server bits, tests are installed, therefore when we build just client bits tets that are relevant to client bits also need to be installed for consistency.
>
>Any switch should default to the same behavior regardless of whether server build is enabled. It is confusing if the --with[out]-[ipa]tests switch changes default based on a different switch passed to configure.
>
>As far as I understand this PR maintains the same default for either server or client only builds, so it gets my approval.
>

Neither of python packages which I mention in #494 package unit test in fedora.

So there is not a reason to package them by default for client only build.
And integration tests require server therefore must not be installed
by default with client-only build.

Result:
This PR has wrong default for instalation of ipatests with client-only build.

LS

pvoborni:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/70554938d4f9ba5b347cd4bc8001428e905198e4
https://fedorahosted.org/freeipa/changeset/41d7ae54fafc6deb602e1a990eaec37c6ae4880b
https://fedorahosted.org/freeipa/changeset/20c1eb9844223d892da47da1ea10662d37953ff8
https://fedorahosted.org/freeipa/changeset/2747f2ad782c7640ecc6949098f0d43411182255

pvoborni:
Also I added section to FreeIPA wiki: http://www.freeipa.org/page/V4/Build_system_refactoring#Packager_-_client_only_build

PR #365, opened 2017-01-03
Silence pylint import errors of ipaserver in ipalib and ipaclient

tiran:
In client-only installations the ipaserver package is not available.
Additional guards prevent pylint to complain about missing ipaserver
package.


mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/987d24f784e05e911bf4e87bd1156abb1dd56210

PR #366, opened 2017-01-03
Use pytest conftest.py 

tiran:
Let's replace some ugly hacks with proper pytest conftest.py hooks.
Test initialization of ipalib.api is now handled in
pytest_cmdline_main(). Pytest plugins, markers and ignores are also
moved into conftest.py. Additional guards make it possible to run tests
without ipaserver installed.


mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/1e06a5195bafe0224d77371987f2509f5508ca2f

PR #367, opened 2017-01-04
Remove nsslib from IPA

stlaz:
This batch of patches removes NSSConnection along with the whole ipapython.nsslib from IPA and replaces it with more standard httplib.HTTPSConnection.

NSSConnection was causing a lot of trouble in the past because it  is apparently very fragile when it comes to nss library initialization. On top of that, when NSSConnection is used to set up an HTTPS connection in FIPS, it always requires a password to NSS database as NSS apparently tries to create a temporary private key and store it to the database even though client authentication is not required in the SSL connection.

TODO (will require changes in certmonger/dogatg.c):
- [x] remove NSSConnection from client modules
- [x] remove NSSConnection from server modules where it's used to connect to the certificate server
- [x] remove the nsslib library completely
- [x] we may probably remove ipaCert from /etc/httpd/alias and stop tracking it with certmonger
- [x] once ^- is done, track /var/lib/ipa/ra-agent.pem in certmonger instead


tiran:
* Ticket 5695 looks wrong, it's about ```FreeIPA on FIPS enabled systems```.
* Are customers fine with the fact that FreeIPA clients will no longer very CRLs? OpenSSL does not automatically download and verify CRLs. OCSP is not yet supported by Python's ssl module.

ipalib/util.py
```diff
@@ -187,6 +194,72 @@ def normalize_zone(zone):
...
+        # remove the slice of negating protocol options according to options
+        min_idx = tls_cutoff.index(self.tls_cutoff_map[tls_version_min])
+        max_idx = tls_cutoff.index(self.tls_cutoff_map[tls_version_max])
+        del(tls_cutoff[min_idx:max_idx+1])
+
+        # official Python documentation states that the best option to get
+        # TLSv1 and later is to setup SSLContext with PROTOCOL_SSLv23
+        # and then negate the insecure SSLv2 and SSLv3
+        ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+        ctx.options |= (ssl.OP_NO_SSLv2 | ssl.OP_NO_SSLv3)
```
tiran: ```()``` are not needed

ipalib/util.py
```diff
@@ -187,6 +194,72 @@ def normalize_zone(zone):
...
+        ]
+        # remove the slice of negating protocol options according to options
+        min_idx = tls_cutoff.index(self.tls_cutoff_map[tls_version_min])
+        max_idx = tls_cutoff.index(self.tls_cutoff_map[tls_version_max])
+        del(tls_cutoff[min_idx:max_idx+1])
+
+        # official Python documentation states that the best option to get
+        # TLSv1 and later is to setup SSLContext with PROTOCOL_SSLv23
+        # and then negate the insecure SSLv2 and SSLv3
+        ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
```
tiran: You want to set secure settings for cipher suites, too. The default settings are rather bad. In fact it would be better if you use ```ssl.create_default_context``` instead of ```ssl.SSLContext```.

ipalib/util.py
```diff
@@ -187,6 +194,72 @@ def normalize_zone(zone):
...
+        for version in tls_cutoff:
+            ctx.options |= version
+
+        if cafile is not None:
+            ctx.verify_mode = ssl.CERT_REQUIRED
+            ctx.check_hostname = True
+            ctx.load_verify_locations(cafile)
+
+        if client_certfile is not None:
+            ctx.load_cert_chain(client_certfile)
```
tiran: Are we using unencrypted client certs? Python's ssl module can handle separate key/cert files and encrypted private keys.

stlaz:
You're right, I should probably write some design. The current implementation does not check CRL or OSCP, so we're "fine" with this change. There is a plan on doing CRL check in certmonger, though.

ipalib/util.py
stlaz: - alright
- I was thinking explicit is better than implicit but I can remove it if you insist

ipalib/util.py
stlaz: I think we dropped `create_default_context` as it was loading system-wide certs which we do not want.

I did not find any mentions in the official documentation about create_default_context setting different cipher suites from SSLContext, can you elaborate more on which should be used?

ipalib/util.py
stlaz: We are. I guess the question here is whether we'd be able to track it with certmonger.

rcritten:
Did you open a bug against NSS or python-nss regarding the PIN requirement?

stlaz:
@rcritten I spoke to the NSS people who assured me it's the intended behavior. But thanks for the remainder, I will open a Bugzilla for that as well, I was considering it before Christmas.

ipaserver/plugins/dogtag.py
```diff
@@ -1238,29 +1238,16 @@ def _parse_dogtag_error(body):
...
         else:
-            self.sec_dir = paths.HTTPD_ALIAS_DIR
-            self.pwd_file = paths.ALIAS_PWDFILE_TXT
-        self.noise_file = self.sec_dir + os.sep + '.noise'
-        self.ipa_key_size = "2048"
-        self.ipa_certificate_nickname = "ipaCert"
-        self.ca_certificate_nickname = "caCert"
-        self._read_password()
+            self.ca_cert = paths.IPA_CA_CRT
+        self.client_certfile = paths.RA_AGENT_PEM
```
stlaz: Note to myself: RA_AGENT_PEM will need to move to `dot_ipa` as well. Testing guide on Wiki will need updating, too.

ipalib/util.py
tiran: ```create_default_context``` does not load the system certificates if you pass in a cafile, capath or cadata argument. Does it really makes sense to ignore the system's trust store? How are we going to validate certificates from 3rd parties, when NSSDB is gone? FreeIPA supports 3rd party certificates for FreeIPA UI, e.g. Lets Encrypt.

Contrary to ```SSLContext```, ```create_default_context``` sets additional options like better selection and order of cipher suites. (https://docs.python.org/3/library/ssl.html#ssl.create_default_context).
> The settings are: PROTOCOL_TLS, OP_NO_SSLv2, and OP_NO_SSLv3 with high encryption cipher suites without RC4 and without unauthenticated cipher suites.

ipalib/util.py
tiran: I guess so:

```
Usage: ipa-getcert request [options]

Required arguments:
* If using an NSS database for storage:
  -d DIR        NSS database for key and cert
  -n NAME       nickname for NSS-based storage (only valid with -d)
  -t NAME       optional token name for NSS-based storage (only valid with -d)
* If using files for storage:
  -k FILE       PEM file for private key
  -f FILE       PEM file for certificate (only valid with -k)
* If keys are to be encrypted:
  -p FILE       file which holds the encryption PIN
  -P PIN        PIN value
```

ipalib/util.py
tiran: keep it :)

ipalib/util.py
```diff
@@ -187,6 +194,72 @@ def normalize_zone(zone):
...
+        # TLSv1 and later is to setup SSLContext with PROTOCOL_SSLv23
+        # and then negate the insecure SSLv2 and SSLv3
+        ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+        ctx.options |= (ssl.OP_NO_SSLv2 | ssl.OP_NO_SSLv3)
+        # pylint: enable=no-member
+        for version in tls_cutoff:
+            ctx.options |= version
+
+        if cafile is not None:
+            ctx.verify_mode = ssl.CERT_REQUIRED
```
tiran: I don't like the fact that no cafile implicitly means no validation. I'd rather have an explicit argument like ```verify=False``` with default setting ```verify=True```. By the way is there any use case to have an unverified connection in FreeIPA?

ipalib/util.py
stlaz: AFAIK we are ignoring system's trust store now so I do not think this should be changed. As for 3rd party certificates, I think all these are store in /etc/ipa/ca.crt so we should be safe there (see ipa_certupdate.py:update_client()).

I am not opposed to using `create_default_context()`, although I thought we may want to have the SSLContext set up exactly according to our needs and in the same way in all Python versions should this ever change.

ipalib/util.py
stlaz: Thanks for the hint, this may work.

tiran:
```
ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
ctx.options = ssl.OP_ALL | ssl.OP_NO_COMPRESSION | ssl.OP_SINGLE_DH_USE | ssl.OP_SINGLE_ECDH_USE | ssl.OP_NO_SSLv2 | ssl.OP_NO_SSLv3
try:
    # use Fedora crypto policy
    # https://fedoraproject.org/wiki/Changes/CryptoPolicy
    ctx.set_ciphers("PROFILE=SYSTEM")
except ssl.SSLError:
    # high ciphers without RC4, MD5, TripleDES, pre-shared key and secure remote password
    ctx.set_ciphers("HIGH:!aNULL:!eNULL:!MD5:!RC4:!3DES:!PSK:!SRP")
```

stlaz:
In the last update I added  SSLv2 support in IPAHTTPSConnection for backward compatibility (https://goo.gl/images/gqh2D9).
I also removed the Fedora crypto policies ciphers as we are not supporting that right now and if we did, we should do that on server as well. There would perhaps be a ticket required.
Also added a ticket to "Move RA agent certificate file export to a different location" as it fixes an issue with missing /etc/httpd/alias/kra-agent.pem as well.

ipalib/util.py
stlaz: I was thinking about this - do we get anything by encrypting the client cert (ra-agent.pem)? The file is root-readable only, if we encrypt it we're only adding more overhead to reading it as the same user can get to the encryption PIN anyway.

ipalib/util.py
```diff
@@ -187,6 +200,88 @@ def normalize_zone(zone):
...
+        # and then negate the insecure SSLv2 and SSLv3
+        ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+        ctx.options |= (
+            ssl.OP_ALL | ssl.OP_NO_COMPRESSION | ssl.OP_SINGLE_DH_USE |
+            ssl.OP_SINGLE_ECDH_USE | ssl.OP_NO_SSLv3
+        )
+
+        # high ciphers without RC4, MD5, TripleDES, pre-shared key
+        # and secure remote password
+        ctx.set_ciphers("HIGH:!aNULL:!eNULL:!MD5:!RC4:!3DES:!PSK:!SRP")
```
tiran: How about you move the hard coded constants to a config option, e.g. ```tls_ciphers```? It makes it much easier to change the option later or set Fedora crypto policy in ```/etc/ipa/default.conf```.

stlaz:
I created the design for this effort: http://www.freeipa.org/page/V4/Replace_NSS_with_OpenSSL

ipalib/util.py
stlaz: I believe that the settings in config options are meant to be user-friendly but I don't think a string such as this one is. I would just set the healthy defaults here and have an admin of the IPA server decide which ciphers should be used by setting these in the config files that are already available.

tiran: No argue on user friendly-ness here. Yes, the cipher string is awkward. The cipher string is used all over the place, e.g. Apache mod_ssl, nginx and tons of other libraries that use OpenSSL. Admins are familiar with it.

```ctx.set_ciphers()``` overrides any default settings in other config files, e.g. system policies. In order to make the ciphers configurable, you have to offer an API.

**Comment by rcritten** [comment]
Wait, you added support for SSLv2? Please remove it, it isn't needed even for backwards compatibility and would not be considered a regression.

**Comment by tiran** [comment]
@rcritten I wonder if we need to support any version except TLS 1.2 at all. Are there any versions of FreeIPA stack that do not have TLS 1.2 support?

**Comment by tiran** [comment]
Let's not make @stlaz jump through more bike-shedding hoops. How about we let him finish this PR, and then address TLS versions, ciphers and other simplifications in another PR?

**Comment by stlaz** [comment]
@rcritten `tls_version_min/max` could have been set to "ssl2" just as well as "ssl3" but perhaps it's for the best to remove them. I will try to do the certmonger part and will remove this with it.

**Comment by rcritten** [comment]
SSLv2 should not be supported, period.

Not that it would work anyway because most SSL libs have completely removed this support, but it is just a terrible idea to even try and allow it.

The rest I'm flexible on.

**Comment by stlaz** [comment]
In the latest patchset, the "ipaCert" is removed from the "/etc/httpd/alias/" NSSDB and all the machinery around the certificate is moved accordingly.
I am addressing support of old SSL protocol versions in https://github.com/freeipa/freeipa/pull/396, although that one currently requires some changes.

**Comment by stlaz** [comment]
In the last update I renamed the proposed config option `ca_certfile` to `cacert_store` and made a requirement for it to be absolute path. This was done with possible future changes to it (thanks @HonzaCholasta for pointing that out).

If the tests pass then this should be ready for review.

**Comment by stlaz** [comment]
Rebased on current master.

File: install/restart_scripts/renew_ra_cert [hunk 7]
```diff
@@ -28,14 +28,14 @@ import shutil
 import traceback
 
 from ipalib.install.kinit import kinit_keytab
-from ipalib import api
-from ipaserver.install import certs, cainstance, dogtaginstance
+from ipalib import api, x509
+from ipaserver.install import certs, cainstance
 from ipaplatform.paths import paths
 
+from cryptography.hazmat.primitives import serialization
```
HonzaCholasta: Nitpick: please place 3rd party module imports between standard module imports and IPA module imports.

ipapython/dogtag.py
```diff
@@ -146,6 +147,7 @@ def https_request(host, port, url, cafile, client_certfile,
 
     def connection_factory(host, port):
         conn = IPAHTTPSConnection(host, port, cafile, client_certfile,
+                                  client_keyfile,
```
HonzaCholasta: Nitpick: could you please use keyword arguments for the files?

ipaserver/install/cainstance.py
```diff
@@ -64,10 +65,11 @@
 from ipaserver.install import ldapupdate
 from ipaserver.install import replication
 from ipaserver.install import sysupgrade
-from ipaserver.install.dogtaginstance import (
-    DogtagInstance, export_ra_agent_pem)
+from ipaserver.install.dogtaginstance import DogtagInstance
 from ipaserver.plugins import ldap2
 
+from cryptography.hazmat.primitives import serialization
```
HonzaCholasta: Nitpick: please place 3rd party module imports between standard module imports and IPA module imports.

ipaserver/install/cainstance.py
```diff
@@ -874,6 +876,11 @@ def __request_ra_certificate(self):
                 except OSError:
                     pass
 
+        ipaapi_gid = grp.getgrnam(ipalib.constants.IPAAPI_GROUP).gr_gid
+        for fname in (paths.RA_AGENT_PEM, paths.RA_AGENT_KEY):
+            os.chown(fname, -1, ipaapi_gid)
+            os.chmod(fname, 0o440)
```
HonzaCholasta: I guess also `restorecon` is in order.

ipaserver/install/ipa_replica_prepare.py
```diff
@@ -427,7 +427,7 @@ def copy_httpd_certificate(self):
             self.log.info("Creating SSL certificate for the Web Server")
             self.export_certdb("httpcert", passwd_fname)
 
-            self.log.info("Exporting RA certificate")
+            self.log.info("Copying RA certificate")
```
HonzaCholasta: Nitpick: IMO the original message is better, copying is an implementation detail of the exporting.

ipaserver/install/server/upgrade.py
```diff
@@ -878,77 +883,50 @@ def certificate_renewal_update(ca, ds, http):
...
+                'cert-database': paths.PKI_TOMCAT_ALIAS_DIR,
+                'cert-nickname': certnick,
+                'ca': 'dogtag-ipa-ca-renew-agent',
+                'cert-presave-command': template % 'stop_pkicad',
+                'cert-postsave-command':
+                    (template % 'renew_ca_cert "{}"').format(certnick),
+            }
+        )
+    # set profile for caSigningCert
+    requests[3]['template-profile'] = 'ipaCACertRenewal'
```
HonzaCholasta: Nitpick: I think a plain list is more readable than the above.

ipaserver/install/server/upgrade.py
```diff
@@ -1380,22 +1349,56 @@ def fix_trust_flags():
...
+        os.chmod(pem_file, 0o440)
+
+    if os.path.exists('/etc/httpd/alias/kra-agent.pem'):
+        os.remove('/etc/httpd/alias/kra-agent.pem')
+
+    httpdb = certs.CertDB(api.env.realm)
+    oldcert = httpdb.get_cert_from_db(ra_nick)
+    if oldcert:
+        certmonger.stop_tracking(paths.HTTPD_ALIAS_DIR, nickname=ra_nick)
+        httpdb.delete_cert(ra_nick)
```
HonzaCholasta: This sort of duplicates the `update_ra_cert_store` plugin's functionality, doesn't it?

ipaserver/plugins/dogtag.py
```diff
@@ -1241,9 +1241,13 @@ def __init__(self, api):
         self.ca_cert = api.env.cacert_store
         if api.env.in_tree:
             self.client_certfile = os.path.join(
-                api.env.dot_ipa, 'ra-agent.pem')
+                api.env.dot_ipa, 'dogtag-agent.pem')
+
+            self.client_keyfile = os.path.join(
+                api.env.dot_ipa, 'dogtag-agent.key')
```
HonzaCholasta: Why the rename from `ra-agent` to `dogtag-agent`?

ipaserver/install/ca.py
```diff
@@ -317,15 +316,15 @@ def install_step_1(standalone, replica_config, options):
         cadb = certs.CertDB(realm_name, subject_base=subject_base)
         dsdb = certs.CertDB(realm_name, nssdir=dirname, subject_base=subject_base)
         trust_flags = dict(reversed(cadb.list_certs()))
-        trust_chain = cadb.find_root_cert('ipaCert')[:-1]
-        for nickname in trust_chain[:-1]:
-            cert = cadb.get_cert_from_db(nickname, pem=False)
+        trust_chain = certstore.get_ca_certs(
+            api.Backend.ldap2, api.env.basedn, api.env.realm, True)
+        for cert, nickname, _trusted, _ext_key_usage in trust_chain[:-1]:
```
HonzaCholasta: This change might have unintentional side effects, I would rather play it safe look for the CA cert chain in Dogtag's NSS database.

ipaserver/plugins/dogtag.py
```diff
@@ -2019,8 +2019,11 @@ def get_client(self):
             str(self.kra_port),
             'kra')
 
-        connection.set_authentication_cert(paths.RA_AGENT_PEM,
-                                           paths.RA_AGENT_KEY)
+        connection.session.cert = (paths.RA_AGENT_PEM, paths.RA_AGENT_KEY)
+        # uncomment the following when this commit makes it to release
+        # https://git.fedorahosted.org/cgit/pki.git/commit/?id=71ae20c
+        # connection.set_authentication_cert(paths.RA_AGENT_PEM,
+        #                                    paths.RA_AGENT_KEY)
```
HonzaCholasta: Why does it need to be reverted? What's the status on pki side?

ipaserver/install/krainstance.py
```diff
@@ -307,7 +307,7 @@ def __create_kra_agent(self):
             usertype=["undefined"],
             userCertificate=[cert_data],
             description=['2;%s;%s;%s' % (
-                cert.serial,
+                cert.serial_number,
```
HonzaCholasta: This is not related to this PR, could you please post it separately?

ipalib/config.py
```diff
@@ -496,6 +496,17 @@ def _bootstrap(self, **overrides):
         if 'nss_dir' not in self:
             self.nss_dir = self._join('confdir', 'nssdb')
 
+        if 'cacert_store' not in self:
+            self.cacert_store = self._join('confdir', 'ca.crt')
```
stlaz: Sure

HonzaCholasta:
Besides what I wrote in inline comments, we need to get rid of `/var/lib/ipa/radb` now that it's unused.

ipalib/util.py
HonzaCholasta: This can (and IMO should) be done in a subsequent PR.

ipaserver/install/cainstance.py
```diff
@@ -874,6 +876,11 @@ def __request_ra_certificate(self):
                 except OSError:
                     pass
 
+        ipaapi_gid = grp.getgrnam(ipalib.constants.IPAAPI_GROUP).gr_gid
+        for fname in (paths.RA_AGENT_PEM, paths.RA_AGENT_KEY):
+            os.chown(fname, -1, ipaapi_gid)
```
stlaz: Use chgrp instead

ipaserver/install/ca.py
stlaz: => look for ca-signing-cert in /etc/pki/pki-tomcat/alias and use certdb.py:get_ca_nickname() to get the nickname with which it should be stored.

stlaz:
First set of fixes to comments arrived, throwing it to Travis.

ipaserver/install/krainstance.py
stlaz: Ok

stlaz:
Some more fixes for Travis to check.

ipaserver/plugins/dogtag.py
stlaz: This is a hack to the commented solution, pki-base.10.3.5-11 available on Fedora still does not provide the method I modified in PKI. We'd better use the public methods of objects we work with instead of injecting their attributes in case their implementation changes.

ipaserver/install/ca.py
stlaz: Done

stlaz:
Hopefully all issues were addressed + `radb` removed. If the Travis check passes then this is ready for review again.

stlaz:
The issues from the previous build should be resolved now, can be reviewed, hopefully the build passes.

ipalib/util.py
```diff
@@ -187,6 +198,137 @@ def normalize_zone(zone):
...
+    # pylint: enable=no-member
+    # disable all TLS versions not in tls_span
+    for version in TLS_VERSIONS:
+        if version not in tls_span:
+            ctx.options |= tls_cutoff_map[version]
+
+    # make sure the required TLS versions are available if Python decides
+    # to modify the default TLS flags
+    for version in tls_span:
+        ctx.options &= ~tls_cutoff_map[version]
```
HonzaCholasta: Nitpick: it looks like this can be done in a single `for` loop rather than two:
```python
    for version in TLS_VERSIONS:
        if version in tls_spawn:
            ctx.options &= ~tls_cutoff_map[version]
        else:
            ctx.options |= ~tls_cutoff_map[version]
```

ipalib/util.py
```diff
@@ -187,6 +198,137 @@ def normalize_zone(zone):
...
+    tls_span = get_proper_tls_version_span(tls_version_min, tls_version_max)
+
+    # official Python documentation states that the best option to get
+    # TLSv1 and later is to setup SSLContext with PROTOCOL_SSLv23
+    # and then negate the insecure SSLv2 and SSLv3
+    # pylint: disable=no-member
+    ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+    ctx.options |= (
+        ssl.OP_ALL | ssl.OP_NO_COMPRESSION | ssl.OP_SINGLE_DH_USE |
+        ssl.OP_SINGLE_ECDH_USE | ssl.OP_NO_SSLv2 | ssl.OP_NO_SSLv3
```
HonzaCholasta: Nitpick: aren't `OP_NO_SSLv2` and `OP_NO_SSLv3` added by the for loop below?

ipaserver/install/ca.py
```diff
@@ -177,14 +177,14 @@ def install_check(standalone, replica_config, options):
...
-        cadb = certs.CertDB(realm_name, subject_base=options._subject_base)
+        cadb = certs.CertDB(realm_name, paths.HTTPD_ALIAS_DIR,
+                            subject_base=options._subject_base)
         dsdb = certs.CertDB(
             realm_name, nssdir=dirname, subject_base=options._subject_base)
 
         for db in (cadb, dsdb):
             for nickname, _trust_flags in db.list_certs():
-                if nickname in (certdb.get_ca_nickname(realm_name),
-                                'ipaCert'):
```
HonzaCholasta: Few lines below is a subject name check, you should remove the RA agent from there as well.

ipaserver/install/cainstance.py
```diff
@@ -271,6 +273,30 @@ def create_ca_user():
...
+    """
+    This is a decorator which helps to backup RA agent files that could be
+    left in the system from previous installations.
+    It sets the right permissions to the newly received files.
+    """
+    def wrapper():
+        # first back up the RA agent files
+        for fname in (paths.RA_AGENT_PEM, paths.RA_AGENT_KEY):
+            if os.path.exists(fname):
+                os.rename(fname, fname + '.bcp')
```
HonzaCholasta: IMO we should not backup the files, they are in `/var/lib/ipa` and noone except IPA itself has any bussiness touching them.

ipaserver/install/cainstance.py
```diff
@@ -271,6 +273,30 @@ def create_ca_user():
...
+
+        # retrieve the certificate and its private key
+        ra_cert_retrieval()
+
+        # set the permissions for the files
+        ipaapi_gid = grp.getgrnam(ipalib.constants.IPAAPI_GROUP).gr_gid
+        for fname in (paths.RA_AGENT_PEM, paths.RA_AGENT_KEY):
+            os.chown(fname, -1, ipaapi_gid)
+            os.chmod(fname, 0o440)
+            tasks.restore_context(fname)
```
HonzaCholasta: Note to self: check if the permissions are intact after renewal.

ipaserver/install/cainstance.py
```diff
@@ -749,7 +789,7 @@ def __get_ca_chain(self):
 
     def __import_ca_chain(self):
         # Backup NSS trust flags of all already existing certificates
-        certdb = certs.CertDB(self.realm)
+        certdb = certs.CertDB(self.realm, self.cert_db_dir)
```
HonzaCholasta: This method is supposed to import the CA chain into the RA agent database. Now that there is no RA agent database, it should either update `/etc/ipa/ca.crt` or be removed, but it should definitely not update `/etc/httpd/alias`.

ipaserver/install/ipa_backup.py
```diff
@@ -158,7 +157,7 @@ class Backup(admintool.AdminTool):
         paths.SMB_CONF,
         paths.SAMBA_KEYTAB,
         paths.DOGTAG_ADMIN_P12,
-        paths.KRA_AGENT_PEM,
+        paths.RA_AGENT_PEM,
```
HonzaCholasta: Also `paths.RA_AGENT_KEY`.

ipaserver/install/kra.py
```diff
@@ -69,15 +69,18 @@ def install_check(api, replica_config, options):
...
         dm_password = options.dm_password
         host_name = api.env.host
         subject_base = dsinstance.DsInstance().find_subject_base()
 
         pkcs12_info = None
         master_host = None
-        ra_only = not options.setup_kra
         promote = False
+    elif not replica_config.setup_kra:
+        return
```
HonzaCholasta: Nitpick: please move this inside the `else` below for better readibility.

ipaserver/install/krainstance.py
```diff
@@ -30,17 +30,17 @@
...
 from ipaserver.install import cainstance
 from ipaserver.install import installutils
 from ipaserver.install import ldapupdate
-from ipaserver.install.dogtaginstance import (export_kra_agent_pem,
-                                              DogtagInstance)
+from ipaserver.install.dogtaginstance import DogtagInstance
 from ipaserver.plugins import ldap2
 from ipapython.ipa_log_manager import log_mgr
 
+from cryptography.hazmat.primitives import serialization
```
HonzaCholasta: Nitpick: please put 3rd party module imports betweed standard module imports and IPA module imports.

ipaserver/install/ipa_replica_prepare.py
```diff
@@ -601,7 +601,8 @@ def export_certdb(self, fname, passwd_fname, is_kdc=False):
                 api.env.realm, nssdir=self.dir, subject_base=subject_base)
             db.create_passwd_file()
             ca_db = certs.CertDB(
-                api.env.realm, host_name=api.env.host,
+                api.env.realm, paths.HTTPD_ALIAS_DIR,
+                host_name=api.env.host,
                 subject_base=subject_base)
             db.create_from_cacert()
             db.create_server_cert(nickname, hostname, ca_db)
```
HonzaCholasta: Rather than using `/etc/httpd/alias` as `ca_db` here, update `CertDB.create_server_cert()` not to require the `other_certdb` argument.

ipaserver/install/krbinstance.py
```diff
@@ -346,8 +346,9 @@ def __create_host_keytab(self):
         self.move_service_to_host(host_principal)
 
     def setup_pkinit(self):
-        ca_db = certs.CertDB(self.realm, host_name=self.fqdn,
-                                subject_base=self.subject_base)
+        ca_db = certs.CertDB(self.realm, paths.HTTPD_ALIAS_DIR,
+                             host_name=self.fqdn,
+                             subject_base=self.subject_base)
```
HonzaCholasta: Rather than using `/etc/httpd/alias` here, please move `install_pem_from_p12` out of `CertDB`.

ipaserver/install/server/install.py
```diff
@@ -774,15 +775,11 @@ def install(installer):
                           if n in options.__dict__}
             write_cache(cache_vars)
 
-        # Create RA DB
-        radb = certs.CertDB(realm_name, nssdir=paths.IPA_RADB_DIR,
-                            user=IPAAPI_USER, group=IPAAPI_GROUP,
-                            truncate=True)
+        # create HTTPD NSS db
+        certs.CertDB(realm_name, nssdir=paths.HTTPD_ALIAS_DIR,
+                     group=constants.HTTPD_GROUP, truncate=True)
```
HonzaCholasta: This should be removed, the database already exists.

ipaserver/install/plugins/update_ra_cert_store.py
```diff
@@ -2,74 +2,92 @@
...
+        certdb.export_pkcs12(ra_nick, p12file)
+
+        # get the private key and certificate from the file
+        with open(p12file, 'r') as f:
+            data = f.read()
+            cert = crypto_x509.load_pem_x509_certificate(
+                data, default_backend())
+            key = serialization.load_pem_private_key(
+                data, password=None, backend=default_backend())
+        os.remove(p12file)
```
HonzaCholasta: A PKCS#12 file is neither an X.509 certificate file nor a private key file. Use `openssl pkcs12` to extract the certificate and private key from the PKCS#12 file.

ipaserver/install/server/upgrade.py
HonzaCholasta: Unless you want to quickly find something and have to look at multiple places (the for loop and this line) to get the information you are looking for.

ipaserver/plugins/dogtag.py
HonzaCholasta: OK.

ipalib/util.py
stlaz: Right, I need to fix this.

ipalib/util.py
stlaz: They might be, I was also wondering whether to add them here or not. I chose to do that for explicitness but may remove them.

ipaserver/install/ca.py
stlaz: Will do.

ipaserver/install/cainstance.py
stlaz: Ok, will just remove them.

ipaserver/install/cainstance.py
stlaz: Ok, I'll see what I can do here.

ipaserver/install/ipa_backup.py
stlaz: Good catch, I forgot about this.

ipaserver/install/kra.py
stlaz: Alright.

ipaserver/install/krainstance.py
stlaz: Did I really leave it there? Will fix.

ipaserver/install/ipa_replica_prepare.py
stlaz: Ok.

ipaserver/install/krbinstance.py
stlaz: Ok.

ipaserver/install/plugins/update_ra_cert_store.py
stlaz: Alright.

ipaserver/install/server/install.py
stlaz: Does it really? Shouldn't a new be created here, backing up the old one?

ipaserver/install/server/upgrade.py
stlaz: I have to say I didn't like having it on two places very much as well. Could have done it similar to what's in `CAInstance.tracking_reqs` but did not like that very much either. Will keep it the old way with a minor improvement of having the requests there straight away.

stlaz:
All the raised issues should've been addressed in the latest PR. Except for the NSS DB creation, please answer the question in `ipaserver/install/server/install.py`

ipaserver/install/plugins/update_ra_cert_store.py
stlaz: I was actually able to simplify the code very much since the same as here is what happens during DL0 replica installation therefore `CAInstance.import_ra_cert()` was already prepared for this.

ipaserver/install/server/install.py
HonzaCholasta: Really, it's created in `HTTPInstance.__setup_ssl()`.

stlaz:
NSS DB creation removed from server install, did not realize it does not matter anymore.

ipaserver/install/plugins/update_ra_cert_store.py
```diff
@@ -2,74 +2,61 @@
...
+        certdb = NSSDatabase(nssdir=paths.HTTPD_ALIAS_DIR)
+        if not certdb.has_nickname(ra_nick):
             # Nothign to do
             return False, []
 
-        newdb = certdb.NSSDatabase(nssdir=paths.IPA_RADB_DIR)
-        if os.path.exists(paths.IPA_RADB_DIR):
-            if newdb.has_nickname('ipaCert'):
+        if os.path.exists(paths.RA_AGENT_PEM):
+            if certdb.has_nickname(ra_nick):
```
HonzaCholasta: This if statement is superfluous - the condition will always be true, because it is verified to be true above.

ipaserver/install/ca.py
```diff
@@ -177,14 +177,14 @@ def install_check(standalone, replica_config, options):
     if standalone:
         dirname = dsinstance.config_dirname(
             installutils.realm_to_serverid(realm_name))
-        cadb = certs.CertDB(realm_name, subject_base=options._subject_base)
+        cadb = certs.CertDB(realm_name, paths.HTTPD_ALIAS_DIR,
+                            subject_base=options._subject_base)
```
HonzaCholasta: The same NSS database as below in `install()` should be used here, i.e. `paths.PKI_TOMCAT_ALIAS_DIR`.

ipaserver/install/httpinstance.py
```diff
@@ -420,7 +419,8 @@ def __setup_ssl(self):
 
     def __import_ca_certs(self):
         # first for the RA DB
-        db = certs.CertDB(self.realm, subject_base=self.subject_base)
+        db = certs.CertDB(self.realm, paths.HTTPD_ALIAS_DIR,
+                          subject_base=self.subject_base)
         self.import_ca_certs(db, self.ca_is_configured)
```
HonzaCholasta: I don't think this needs to be done twice.

ipaserver/plugins/rabase.py
```diff
@@ -44,8 +44,8 @@ def __init__(self, api):
             self.sec_dir = api.env.dot_ipa + os.sep + 'alias'
             self.pwd_file = self.sec_dir + os.sep + '.pwd'
         else:
-            self.sec_dir = paths.IPA_RADB_DIR
-            self.pwd_file = os.path.join(paths.IPA_RADB_DIR, 'pwdfile.txt')
+            self.sec_dir = paths.HTTPD_ALIAS_DIR
+            self.pwd_file = os.path.join(paths.HTTPD_ALIAS_DIR, 'pwdfile.txt')
```
HonzaCholasta: This does not make sense anymore, the same as in `RestClient.__init__()` should be done here.

ipaserver/secrets/store.py
```diff
@@ -46,7 +46,7 @@ def PKI_TOMCAT_password_callback():
 
 
 def HTTPD_password_callback():
```
HonzaCholasta: Isn't this unused now?

HonzaCholasta:
Upgrade from 4.3 fails with:
```
2017-02-28T07:07:18Z DEBUG Starting external process
2017-02-28T07:07:18Z DEBUG args=/usr/bin/pk12util -d /etc/httpd/alias -o (6, '/etc/httpd/alias/tmpFNEJrK') -n ipaCert -k /etc/httpd/alias/pwdfile.txt
2017-02-28T07:07:18Z DEBUG Process execution failed
2017-02-28T07:07:18Z DEBUG Destroyed connection context.ldap2_139873144635088
2017-02-28T07:07:18Z ERROR Upgrade failed with coercing to Unicode: need string or buffer, tuple found
2017-02-28T07:07:18Z DEBUG Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/ipaserver/install/upgradeinstance.py", line 219, in __upgrade
    self.modified = (ld.update(self.files) or self.modified)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/ldapupdate.py", line 911, in update
    self._run_updates(all_updates)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/ldapupdate.py", line 883, in _run_updates
    self._run_update_plugin(update['plugin'])
  File "/usr/lib/python2.7/site-packages/ipaserver/install/ldapupdate.py", line 859, in _run_update_plugin
    restart_ds, updates = self.api.Updater[plugin_name]()
  File "/usr/lib/python2.7/site-packages/ipalib/frontend.py", line 1470, in __call__
    return self.execute(**options)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/plugins/update_ra_cert_store.py", line 47, in execute
    certdb.export_pkcs12(ra_nick, p12file)
  File "/usr/lib/python2.7/site-packages/ipapython/certdb.py", line 232, in export_pkcs12
    ipautil.run(args)
  File "/usr/lib/python2.7/site-packages/ipapython/ipautil.py", line 442, in run
    preexec_fn=preexec_fn)
  File "/usr/lib64/python2.7/subprocess.py", line 390, in __init__
    errread, errwrite)
  File "/usr/lib64/python2.7/subprocess.py", line 1024, in _execute_child
    raise child_exception
TypeError: coercing to Unicode: need string or buffer, tuple found

2017-02-28T07:07:18Z DEBUG Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/ipaserver/install/service.py", line 423, in start_creation
    run_step(full_msg, method)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/service.py", line 413, in run_step
    method()
  File "/usr/lib/python2.7/site-packages/ipaserver/install/upgradeinstance.py", line 227, in __upgrade
    raise RuntimeError(e)
RuntimeError: coercing to Unicode: need string or buffer, tuple found

2017-02-28T07:07:18Z DEBUG   [error] RuntimeError: coercing to Unicode: need string or buffer, tuple found
```

HonzaCholasta:
CA-less to CA-full `ipa-ca-install` fails with:
```
2017-02-28T07:24:47Z DEBUG   File "/usr/lib/python2.7/site-packages/ipaserver/install/installutils.py", line 892, in run_script
    return_value = main_function()

  File "/sbin/ipa-ca-install", line 304, in main
    promote(safe_options, options, filename)

  File "/sbin/ipa-ca-install", line 270, in promote
    install_master(safe_options, options)

  File "/sbin/ipa-ca-install", line 235, in install_master
    ca.install(True, None, options)

  File "/usr/lib/python2.7/site-packages/ipaserver/install/ca.py", line 204, in install
    install_step_1(standalone, replica_config, options)

  File "/usr/lib/python2.7/site-packages/ipaserver/install/ca.py", line 325, in install_step_1
    config_ipa=True, config_compat=True)

  File "/usr/lib/python2.7/site-packages/ipalib/install/certstore.py", line 410, in put_ca_cert_nss
    config_ipa, config_compat)

  File "/usr/lib/python2.7/site-packages/ipalib/install/certstore.py", line 233, in put_ca_cert
    config_ipa=config_ipa, config_compat=config_compat)

  File "/usr/lib/python2.7/site-packages/ipalib/install/certstore.py", line 160, in update_ca_cert
    subject, issuer_serial, public_key = _parse_cert(dercert)

  File "/usr/lib/python2.7/site-packages/ipalib/install/certstore.py", line 39, in _parse_cert
    raise ValueError("failed to decode certificate: %s" % e)

2017-02-28T07:24:47Z DEBUG The ipa-ca-install command failed, exception: ValueError: failed to decode certificate: Unable to load certificate
```

HonzaCholasta:
`ipa-replica-install` with `--setup-ca` fails with:
```
2017-02-28T07:38:41Z DEBUG   File "/usr/lib/python2.7/site-packages/ipapython/admintool.py", line 172, in execute
    return_value = self.run()
  File "/usr/lib/python2.7/site-packages/ipapython/install/cli.py", line 336, in run
    cfgr.run()
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 328, in run
    self.execute()
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 352, in execute
    for _nothing in self._executor():
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 394, in __runner
    exc_handler(exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 423, in _handle_execute_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 413, in _handle_exception
    six.reraise(*exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 384, in __runner
    step()
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 381, in <lambda>
    six.reraise(*exc_info)
    value = gen.send(prev_value)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 618, in _configure
    next(executor)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 394, in __runner
    exc_handler(exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 423, in _handle_execute_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 481, in _handle_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 413, in _handle_exception
    six.reraise(*exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 478, in _handle_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 413, in _handle_exception
    six.reraise(*exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 384, in __runner
    step()
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 381, in <lambda>
    six.reraise(*exc_info)
    value = gen.send(prev_value)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/__init__.py", line 595, in main
    replica_install(self)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/replicainstall.py", line 398, in decorated
    func(installer)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/replicainstall.py", line 1455, in install
    ca.install(False, config, options)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/ca.py", line 203, in install
    install_step_0(standalone, replica_config, options)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/ca.py", line 282, in install_step_0
    use_ldaps=standalone)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/cainstance.py", line 478, in configure_instance
    self.start_creation(runtime=210)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/service.py", line 423, in start_creation
    run_step(full_msg, method)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/service.py", line 413, in run_step
    method()
  File "/usr/lib/python2.7/site-packages/ipaserver/install/cainstance.py", line 289, in wrapper
    ra_cert_retrieval(cls, *args)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/cainstance.py", line 729, in __import_ra_key
    custodia.import_ra_key(self.master_host)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/custodiainstance.py", line 119, in import_ra_key
    cli.fetch_key('ra/ipaCert')
  File "/usr/lib/python2.7/site-packages/ipaserver/secrets/client.py", line 100, in fetch_key
    r.raise_for_status()
  File "/usr/lib/python2.7/site-packages/requests/models.py", line 844, in raise_for_status
    raise HTTPError(http_error_msg, response=self)

2017-02-28T07:38:41Z DEBUG The ipa-replica-install command failed, exception: HTTPError: 404 Client Error: Not Found for url: https://vm-226.abc.idm.lab.eng.brq.redhat.com/ipa/keys/ra/ipaCert?type=kem&value=eyJhbGciOiJSU0EtT0FFUCIsImVuYyI6IkEyNTZDQkMtSFM1MTIiLCJraWQiOm51bGx9.UbDCRVumiqN3YDwxdfHGvkzDakB0Isbq3dTZ9tMVe5NK_wuvGmtBYGUfO46IQmBsqto0N28WJUcselfuY8Q3uSOEPb0HximzbmnJm-S1TrF9KMsHgozwhNjXDAXapXSmiFqyKiTPAvxLzx0OKq052oIbnUsprk3s6R_mQ5AWrP52DsVqM9EovOHa6RM3t4xasamhJ_at1eR647TGqS8pamulRkSilK-kxyPQnN7Rphz_cyr0yIlG93xnfQTsCJ9WcodWmVPrPgP3PvKh8OaINHqUsfT-gUR0IR3hXkN5slAYSxOdmNYwLVR-wtc6Yh3jf4LpRgbHhfmuNx8rTU3nuw.-xfhG2UGKLBU7D6UzO4d5g.zG5YCUsskbqiYRYEFGqDwbx0JghL2Yo-oioFM8KBVoxvfVNtrUzN96TU-aQn08WXS2GFvAdXo7-EpOMtrFwPEGiWfXhLDjYAhVCAmu69YYjptCDokSEY8PK1HtUJanVTb0LtcPlp78yNyM-ZGC42-PfMiiG66rlWMMpWtAWPHugxDa8EcV8AlTFdqtqFHwzRYxISXUbuiwGD_h8pht2irYGdSeJ6Aa6Fwk54ZQdQshb24njjBt-MIrgy1YKlTkF4nPhuPhH0o70IWFoQSQ24R7GVfofMc3xnoUCtPlv-QLnaOucnvCDZPUBx33JGRJP5Y7Acpp_MJTkWqmSWvmXlhNouWXtpe3oTakKNYqqurNieJtHaGmIyuyn0yzQcv4w1re8aEn9Zv2TIZvSfq5qxMqUvlhiyhej6ZCRQ1FdLaslPbFuullik95Ik-pF7BBtvn3d5LKnZJgYnQg2n3yzi73zdMdync_rovg5abmWKLAM_SrRVgeoJ9-TQrJ18HvuViIoz1n14-TQyGKaw8hsbNGJZE0vrPx0gVTl4-HJK9PLP8M1jUylSxtVRC-Xv1bny4LkLQotJuV0wbZDKHg77gk3xolAUAZ6ZrWD8xUVprdJlYW423NYyX_t-2-c6HUQBlmZqkv_lYrgaZ5WZ5mw2U1MwVxf_KU6SnwYM3kPQm3a13KkeoK9IdmnZv3YAvrKOjtekmUwCjaItX74FAg2IJA.eoL7I4gO67k_Ejt9ttgKFnw29RC3Uzz1Ykn6-mA4DsM
```

ipaserver/install/httpinstance.py
stlaz: Lol 💩

ipaserver/install/ca.py
stlaz: Sure, missed that.

ipaserver/install/plugins/update_ra_cert_store.py
stlaz: I will make it go away.

ipaserver/plugins/rabase.py
stlaz: Ok

ipaserver/secrets/store.py
stlaz: Did not find it anywhere, will remove it.

stlaz:
The issues should hopefully be fixed

stlaz:
Fixed another issue with CA-less to CA-full upgrade.

HonzaCholasta:
Upgrade from 4.4.3 asks for a PKCS#12 file password and then fails:
```
  Cleanup     : freeipa-server-common-4.4.3-1.fc25.noarch                                                                                                                                                    14/16 
  Cleanup     : freeipa-client-common-4.4.3-1.fc25.noarch                                                                                                                                                    15/16 
  Cleanup     : freeipa-common-4.4.3-1.fc25.noarch                                                                                                                                                           16/16 
Enter password for PKCS12 file: 
Re-enter password: 
IPA server upgrade failed: Inspect /var/log/ipaupgrade.log and run command ipa-server-upgrade manually.
Unexpected error - see /var/log/ipaupgrade.log for details:
NetworkError: cannot connect to 'ldapi://%2fvar%2frun%2fslapd-ABC-IDM-LAB-ENG-BRQ-REDHAT-COM.socket': 
The ipa-server-upgrade command failed. See /var/log/ipaupgrade.log for more information
  Verifying   : freeipa-client-4.4.90.dev201703010721+git5bb660e-0.fc25.x86_64                                                                                                                                1/16 
  Verifying   : freeipa-client-common-4.4.90.dev201703010721+git5bb660e-0.fc25.noarch                                                                                                                         2/16 
  Verifying   : freeipa-common-4.4.90.dev201703010721+git5bb660e-0.fc25.noarch                                                                                                                                3/16 
```

stlaz:
This should now be fixed. In my endless naivety I had thought passing no password to `export_pkcs12()` would actually mean no password will be set.

ipaserver/install/cainstance.py
HonzaCholasta: Conclusion: they are.

HonzaCholasta:
CA-less to CA-ful conversion still fails:
```
2017-03-01T09:14:40Z DEBUG Starting external process
2017-03-01T09:14:40Z DEBUG args=/usr/sbin/pkispawn -s CA -f /tmp/tmpgj_Ue4
2017-03-01T09:14:40Z DEBUG Process finished, return code=1
2017-03-01T09:14:40Z DEBUG stdout=Log file: /var/log/pki/pki-ca-spawn.20170301101440.log
Loading deployment configuration from /tmp/tmpgj_Ue4.
Installing CA into /var/lib/pki/pki-tomcat.
Storing deployment configuration into /etc/sysconfig/pki/tomcat/pki-tomcat/ca/deployment.cfg.

Installation failed: Directory '/etc/pki/pki-tomcat' already exists!


2017-03-01T09:14:40Z DEBUG stderr=pkispawn    : ERROR    ....... Directory '/etc/pki/pki-tomcat' already exists!

2017-03-01T09:14:40Z CRITICAL Failed to configure CA instance: Command '/usr/sbin/pkispawn -s CA -f /tmp/tmpgj_Ue4' returned non-zero exit status 1
2017-03-01T09:14:40Z CRITICAL See the installation logs and the following files/directories for more information:
2017-03-01T09:14:40Z CRITICAL   /var/log/pki/pki-tomcat
2017-03-01T09:14:40Z DEBUG Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/ipaserver/install/service.py", line 423, in start_creation
    run_step(full_msg, method)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/service.py", line 413, in run_step
    method()
  File "/usr/lib/python2.7/site-packages/ipaserver/install/cainstance.py", line 611, in __spawn_instance
    nolog_list=(self.dm_password, self.admin_password)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/dogtaginstance.py", line 144, in spawn_instance
    self.handle_setup_error(e)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/dogtaginstance.py", line 391, in handle_setup_error
    raise RuntimeError("%s configuration failed." % self.subsystem)
RuntimeError: CA configuration failed.

2017-03-01T09:14:40Z DEBUG   [error] RuntimeError: CA configuration failed.
```
Not sure if it's caused by the PR or not, but either way it can be fixed later.

HonzaCholasta:
`ipa-replica-install --setup-ca` still fails with the same error though.

stlaz:
@HonzaCholasta I saw this issue as well, once you hit it on a VM no `pkispawn` will run correctly. I am not sure if it's caused by this PR, my guess is it shouldn't be as `pkispawn` was not touched at all but I can't be sure.

HonzaCholasta:
OK. Let's fix it later.

HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/dfd560a190cb2ab13f34ed9e21c5fb5c6e793f18
https://fedorahosted.org/freeipa/changeset/2a1494c9aef2e2b5c06e427e689787e5a2c4dc7f
https://fedorahosted.org/freeipa/changeset/1e89d28aaf3a0a4b48fc09a5d98262f1000c52a3
https://fedorahosted.org/freeipa/changeset/6b074ad833a12acbd4643795b2150fa7f019d6b2
https://fedorahosted.org/freeipa/changeset/0a54fac02cecad3b9e3bf8ad0c8a44df3b701857
https://fedorahosted.org/freeipa/changeset/afea026a5c45ce24f3bf6da499b4d334eea3ca78
https://fedorahosted.org/freeipa/changeset/2a9d1fb7d9dda0299c6f7cd75a715182d15e04df
https://fedorahosted.org/freeipa/changeset/76e8d7b35d110e5cf5494898950ab3607799c031
https://fedorahosted.org/freeipa/changeset/595f9b64e31dc9e4f035119e834db7e6cb152dce
https://fedorahosted.org/freeipa/changeset/51a2b1372936106ff95d5a45afc813f146653ae4
https://fedorahosted.org/freeipa/changeset/24b134c633390343ba76e4091fa612650976280a
https://fedorahosted.org/freeipa/changeset/5ab85b365ae886558b1f077b0d039a0d24bebfa7

PR #368, opened 2017-01-04
WebUI: fix incorrect behavior of ESC button on combobox     

pvomacka:
When combobox is opened then ESC key should close it. There was a bug
that ESC key closed also the dialog. It was caused by bad keyboard event
handling. The CB was closed by keydown event and the dialog by keyup.
    
Therefore the propagating of keyup and keydown event is stopped when CB
is opened (when the event is fired on CB element).
    
https://fedorahosted.org/freeipa/ticket/6388

### DIFF
--- a/install/ui/src/freeipa/add.js
+++ b/install/ui/src/freeipa/add.js
@@ -36,6 +36,8 @@ IPA.entity_adder_dialog = function(spec) {
 
     var that = IPA.dialog(spec);
 
+    that.on_cancel = that.close;
+
     IPA.confirm_mixin().apply(that);
 
     /** @property {string} method="add" API method for add command */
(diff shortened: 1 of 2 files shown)


pvoborni:
Code LGTM, but I did not tests the behavior, so cannot give ACK now.

MartinBasti:
Works for me

pvoborni:
ACK given that Martin did functional testing

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/1a96e7f9e737941480436269668ccde0a50395f9
https://fedorahosted.org/freeipa/changeset/6c6c68df544ac1046741d91dfdc59ef8d96b863c

PR #369, opened 2017-01-04
Catch ValueError raised by pytest.config.getoption()

tiran:
pytest.config.getoption() can raise ValueError for unknown options, too.

I ran into the issue while I was working on PR #366


mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/3387734e6c6d47a756b5e914e7e515d2610a424f

PR #370, opened 2017-01-05
ci: send build log to paste.fedoraproject.org

frasertweedale:
<strike>This commit is just to see if we can ship our build logs off travis
to a pastebin.

If we can, we can refine the approach to only ship logs when the build
broke, provide better output about where to find them, etc.</strike>

No longer experimental; ready for review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,5 +33,20 @@ install:
 script:
     - travis_wait 50 ./.travis_run_task.sh
 after_failure:
+    - >
+      echo "Sending test runner output to paste.fedoraproject.org"
+      && sudo apt-get -qq update && sudo apt-get install -y jq
+      && gzip < ci_results_${TRAVIS_BRANCH}.log > ci_results.log.gz
+      && ls -l ci_results.log.gz
+    - >
+      PASTE_ID=$(curl https://paste.fedoraproject.org/~freeipa.ci/ -H Expect:
+      --data api_submit=true
+      --data mode=json
+      --data paste_lang=text
+      --data paste_expire=$(expr 86400 '*' 28)
+      --data-urlencode paste_data@ci_results.log.gz
+      | jq --raw-output .result.id)
+      && echo "Download gzipped logfile from: https://paste.fedoraproject.org/$PASTE_ID/raw/"
+      || echo "Failed to submit paste!"
     - echo "Test runner output:"; tail -n $CI_BACKLOG_SIZE $CI_RESULTS_LOG
     - echo "PEP-8 errors:"; cat $PEP8_ERROR_LOG


frasertweedale:
OK, so we can ship a paste off but paste.fedoraproject.org does not like the file size (~1.8M).
In this case the HTTP response is 200 OK and the response body is the HTML frontpage.  The
paste does not succeed.

Experimentally: a paste of < 512K succeeds, but a paste of ~1M fails.

Now, fpaste is happy enough accepting binary data, e.g. a gzipped file

    curl https://paste.fedoraproject.org/520077/raw/ | zless

The downsides to doing that are:

    1. Cannot view in browser
    2. Inefficiency of percent-encoding (compressed data will inflate by ~2.5x for transfer)
        - base64url-encoding the compressed data will avoid percent-encoding and limit inflation to 1.33x

But the upside is of course that we can get these files off so developers can get at them, so I think we should do that.

frasertweedale:
Additional notes about paste.fedoraproject.org projects:

- seems that only names consisting entirely of alpha chars work (thus ruling out `freeipa-ci` or similar)
- pastes to a project namespace appear in *both* the project archive, and the main archive.
- example command:
  ```shell
  curl -v https://paste.fedoraproject.org/~freeipa/ -H Expect: \
    -d api_submit=true \
    -d mode=json \
    -d paste_lang=text \
    -d paste_data=hello+world \
    -d paste_expire=300
  ```
- paste can be accessed via top name space or project (or any *other*, too)

frasertweedale:
Note: a new fedora pastebin is forthcoming.  Staging instance: https://modernpaste.stg.fedoraproject.org/

martbab:
LGTM, it would be nice if you could temporarily sneak in some breaking commit so that we can see how the paste looks like.

frasertweedale:
@martbab the paste looks like gobbledygook; it's gzipped.

We will see it in action soon enough :)

frasertweedale:
So... any blocker on merging this?

martbab:
We were discussing yout PR with @HonzaCholasta yesterday and he suggested that we could pack the test runner log, HTTP and dirsrv logs and server installer log into an archive and upload it to https://transfer.sh/ a very lightweight file hosting site. What do you think?

HonzaCholasta:
Right, I suggested https://transfer.sh, because uploading a file there is as easy as:
```bash
curl --upload-file ./file https://transfer.sh/file
```
BTW I would not limit ourselves to the few logs @martbab suggested, but upload as much as possible: `/var/log/ipa*`, `/var/log/httpd`, `/var/log/dirsrv`, `/var/log/pki/pki-tomcat`, `journalctl` dump for all of our services, ...

martbab:
@HonzaCholasta I have measured the size of the whole gzipp'ed /var/log directory from the CI run and it has around 6 megs. We may thus paste the whole archived directory there (including journal and stuff) and have all the bases covered. I am working on a POC patch and will see how it will turn out.

frasertweedale:
:+1: sounds good.  Take what's there and run with it :)

HonzaCholasta:
@martbab, I would rather not include irrelevant stuff, it's just noise.

martbab:
@HonzaCholasta ok in that case I will archive: `/var/log/httpd`, `/var/log/dirsrv/`, `/var/log/pki/pki-tomcat`, `ipaserver/client-install.log` and a dump from systemd journal. Should be enough.

frasertweedale:
Superseded by https://github.com/freeipa/freeipa/pull/449 ; closing.

PR #371, opened 2017-01-05
[4.4] Set up DS TLS on replica in CA-less topology

frasertweedale:
Fixes: https://fedorahosted.org/freeipa/ticket/6226


tomaskrizek:
I re-tested CA-less and CA-full use cases in both domlvl0 and domlvl1. They all seem to work and ldapssl is running.

Thanks for fixing the issue!

mbasti-rh:
Fixed upstream
ipa-4-4:
https://fedorahosted.org/freeipa/changeset/cdb6ffb779b7e1e563494eb3234b2441ba74d692

PR #372, opened 2017-01-05
Restore IPA 3.0 compatibility of copy-schema-to-ca.py

tiran:
Apparently ipaplatform.paths is not available on IPA 3.x.

https://fedorahosted.org/freeipa/ticket/6540

### DIFF
--- a/install/share/copy-schema-to-ca.py
+++ b/install/share/copy-schema-to-ca.py
@@ -17,7 +17,6 @@ import shutil
 
 from hashlib import sha1
 
-from ipaplatform.paths import paths
 from ipapython import ipautil
 from ipapython.ipa_log_manager import root_logger, standard_logging_setup
 from ipaserver.install.dsinstance import schema_dirname
@@ -33,6 +32,14 @@ except ImportError:
     # oh dear, this is an old IPA (3.0+)
     from ipaserver.install.dsinstance import DS_USER   #pylint: disable=E0611
     from ipaserver.install.cainstance import PKI_USER  #pylint: disable=E0611
+
+try:
+    from ipaplatform.paths import paths
+    USR_SHARE_IPA_DIR = paths.USR_SHARE_IPA_DIR
+    ETC_IPA = paths.ETC_IPA
+except (ImportError, AttributeError):
+    USR_SHARE_IPA_DIR = "/usr/share/ipa/"
+    ETC_IPA = "/etc/ipa"
 
 try:
     from ipaplatform import services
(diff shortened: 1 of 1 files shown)


stlaz:
Is there a reason not to stick with the original `ipautil.SHARE_DIR` and without setting `confdir`? This script won't be run on servers that either need `confdir` set or have `ipaplatform.paths`, will it (I know I acked the latter, did not realize there would be trouble)?

mbasti-rh:
this script must work only with IPA3.x, so I wouldn't add there anything from 4.4/master code. As I pointed out several times I don't think that this code even should be in master branch, as we are always just fixing regressions there, but rather as separate script, or script in IPA3.x. Unfortunately I haven't got enough support for my idea.

HonzaCholasta:
I agree with @mbasti-rh. IMO we should remove all 4.0+ specific code from the script, add a version check at the beginning and disable all failing pylint checks. Maybe also add a comment at the top saying that the file is not to be modified anymore.

stlaz:
+1, that was actually my original point. Just revert the change done to the file in https://git.fedorahosted.org/cgit/freeipa.git/commit/?id=d6b755e3fcaf32158f4ee36d45e3344b4a03fbc2, don't add `confdir` option to api.bootstrap() and let the script die in peace.

tiran:
How about we remove the file entirely and just post it on the wiki or something?

martbab:
IIRC @mbasti-rh proposed to maintain the script separately and serve it to users via external repo or something but the idea was rejected.

mbasti-rh:
I proposed 2 ideas:
- move it to ipa-3-3 branch (or)
- extract that script from freeipa repo and allow to download that script from freeipa.org (and access.redhat.com)

tiran:
So with *separate script* you meant a separate downloadable version of the script. Got it! :)

It seems we have a consent. @mbasti-rh I second your proposal to move it to freeipa.org (that what I meant with wiki) and access.redhat.com.

stlaz:
+1, we need to fix the script first, though.

tiran:
Or we just grab a working and tested version from an old release.

mbasti-rh:
@tiran +1, but first this has to be generally approved :) topic for meeting maybe (today?)

tiran:
I have updated the ticket https://fedorahosted.org/freeipa/ticket/6540#comment:5 with the result of this discussion. I'm going to close the PR. Let's start a new one to remove it and update ```ipaserver/install/cainstance.py``` plus builds.

PR #376, opened 2017-01-06
client install: correctly report all failures

HonzaCholasta:
In commit 5249eb817efbb5708d097173a8d5f1e322fb201e, the client install code
was converted to use exception handling instead of return codes. However,
some return statements were not converted to raise statements and as a
result, ipa-client-install will report success in some error conditions.

Convert the return statements to raise statements to fix the issue.

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -2715,13 +2715,13 @@ def _install(options):
             except errors.PublicError as e2:
                 root_logger.warning(
                     "Second connect with delegate=True also failed: %s", e2)
-                root_logger.error(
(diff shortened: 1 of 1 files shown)


stlaz:
I suspect we are suffering the same "always return 0" error as we've already got reported in other installers, right?

stlaz:
Alright, that's a known issue from the refactorings. Other than that the patch is fine. ACK.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/26630db9d0fb1d9c8a02840b71b3fb3e8bdf3e0d

PR #377, opened 2017-01-09
dogtaginstance: track server certificate with our renew agent

HonzaCholasta:
This patchset is intended to make @simo5's life easier when changing the RA agent certificate location in #314.

**renew agent: handle non-replicated certificates**

In addition to replicated certificates (Dogtag certificates, RA
certificate), handle non-replicated certificates in
dogtag-ipa-ca-renew-agent as well.

**dogtaginstance: track server certificate with our renew agent**

Track Dogtag's server certificate with dogtag-ipa-ca-renew-agent instead of
dogtag-ipa-renew-agent.

**cainstance: do not configure renewal guard**

Do not configure renewal guard for dogtag-ipa-renew-agent, as it is not
used in IPA anymore.

https://fedorahosted.org/freeipa/ticket/5959

### DIFF
--- a/install/certmonger/dogtag-ipa-ca-renew-agent-submit
+++ b/install/certmonger/dogtag-ipa-ca-renew-agent-submit
@@ -108,6 +108,15 @@ def is_renewable():
         return x509.is_self_signed(cert) or is_lightweight_ca()
 
 
+def is_replicated():
+    return not get_nickname()
+
+
+def is_renewal_master():
+    ca = cainstance.CAInstance(host_name=api.env.host)
+    return ca.is_renewal_master()
+
+
 @contextlib.contextmanager
 def ldap_connect():
     conn = None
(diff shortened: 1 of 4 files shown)


abbra:
Looks very good to me. ACK from my side.

stlaz:
Works fine.

stlaz:
I made a patch that makes is_renewal_master and set_renewal_master classmethods on @tiran recommendation. Feel free to push it along or leave it, don't let it slow us down if you don't like it.
http://pastebin.com/dCDTAtnS
ACK

stlaz:
This PR can be safely pushed, an unknown upstream contributor with the same github nick as me will later create a PR with the proposed classmethod change.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/d5af11f65cc2a2d6860579a63a173b67cb12bcf3
https://fedorahosted.org/freeipa/changeset/ad49bda907b3c2ec5b98946a2c4000bb6edaf835
https://fedorahosted.org/freeipa/changeset/926fe2049a1839fd7e68c9fa55f64154ee83c841

PR #379, opened 2017-01-09
Packaging: Add IPA commands package

tiran:
The ipacommands package contains ipa-getkeytab and ipa-rmkeytab for
installation in a virtual env. The programs are compiled with distutils
/ setuptools.

The ipa and freeipa packages are placeholders to prevent PyPI squashing
attacks and reserve the names for future use. `pip install ipa` installs
ipaclient.

https://fedorahosted.org/freeipa/ticket/6484

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,3 +112,9 @@ freeipa2-dev-doc
 /ipapython/version.py
 /ipapython/.DEFAULT_PLUGINS
 
+/pypi/ipacommands/COPYING
+/pypi/ipacommands/Contributors.txt
+/pypi/ipacommands/asn1
+/pypi/ipacommands/client
+/pypi/ipacommands/ipasetup.py
+/pypi/ipacommands/util
--- a/Makefile.am
+++ b/Makefile.am
@@ -285,6 +285,8 @@ bdist_wheel: $(WHEELDISTDIR)
 	for dir in $(IPA_WHEEL_PACKAGES) ipatests; do \
 	    $(MAKE) $(AM_MAKEFLAGS) -C $${dir} $@ || exit 1; \
 	done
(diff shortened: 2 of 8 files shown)


tiran:
The ```ipa``` and ```freeipa``` packages are necessary to prevent typo squatting or name squatting attacks, e.g. http://arstechnica.com/security/2016/06/college-student-schools-govs-and-mils-on-perils-of-arbitrary-code-execution/ . We want to make sure that a developer gets FreeIPA when he does ```pip install freeipa```.

I already reserved the names on PyPI. It is necessary to upload new packages for ```ipa``` and ```freeipa``` regularly. Otherwise PyPI considers our packages obsolete and may remove them. See https://www.python.org/dev/peps/pep-0541/

pvoborni:
I thought that I understand why this PR is needed bud in fact I don't.  Ticket #6484 is closed. Why is it attached to it?

How will the pypi packaging change if ipacommands package is not there? Would it be used for anything? How it should be used?

MartinBasti:
We need placeholder package for sure, this PR should be splitted into 2, but I'm still not endorsed to have ipa-getkeytab installable by pip

tiran:
I don't mind to maintain my own copy of ipacommands with ```ipa-getkeytab``` until we agree on a permanent solution.

pvoborni:
If there is reason it can  be maintained in IPA, but what is the reason?

tiran:
I have postponed the ```ipacommands``` part. Placeholders are covered by #472.

tomaskrizek:
Closing for inactivity. If this PR is still relevant, please re-open it.

tiran:
This PR will probably never land.

PR #380, opened 2017-01-09
Travis CI: actually return non-zero exit status when the test job fails

martbab:
Thanks to @stlaz for catching this.


stlaz:
It works but for some reason there are many extra newlines in the failure log

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/9b5b7131502a73fa24dc56c72a9648528c5aceee

PR #381, opened 2017-01-09
disable hostname canonicalization by Kerberos library

martbab:
By default, Kerberos client library attempts to canonicalize service
hostname in TGS requests. This can fail e.g. if hosts file on the client
machine references short names before FQDNs. In this case the short name
is used in TGS_REQ which KDC fails to resolve.

Since we do not (yet) support referencing hosts by their short names it
is safe to just disable this behavior in krb5.conf and use supplied
FQDNs.

https://fedorahosted.org/freeipa/ticket/6584

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -676,6 +676,7 @@ def configure_krb5_conf(
         ])
     libopts.extend([
         krbconf.setOption('rdns', 'false'),
+        krbconf.setOption('dns_canonicalize_hostname', 'false'),
         krbconf.setOption('ticket_lifetime', '24h'),
         krbconf.setOption('forwardable', 'true'),
         krbconf.setOption('udp_preference_limit', '0')


tiran:
One Travis job was failing, I restarted it.

tomaskrizek:
Works as expected.

martbab:
Thanks for ACK. I would like to ask @simo5  if this change is okay from krb5 point of view and does not pose any security problem for clients.

simo5:
@martbab this change actually improves security by avoiding a DNS lookup that could be manipulated by an attacker, however it also means some setups may break, because they depend on canonicalization to actually get the correct name, and should be documented in release notes.

pvoborni:
To not forget to update the release notes later at release, @martbab could you update the respected fields in both ticket and BZ when the patch is pushed.

martbab:
@pvoborni will do.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/566c86a782bfd7d50938866e9f89faf56cea773f

PR #382, opened 2017-01-09
[Py3] ipa-server-install fixes (working NTP, DS, CA install steps)

MartinBasti:
This PR should allow to install server with py3

Up to "Done configuring certificate server (pki-tomcatd)."

I decided to split this effort as it looke that there will be too many patches

### DIFF
--- a/ipaclient/plugins/vault.py
+++ b/ipaclient/plugins/vault.py
@@ -969,7 +969,7 @@ class vault_retrieve(Local):
         json_vault_data = decoding_ctx.cipher_op(wrapped_vault_data)\
             + decoding_ctx.digest_final()
 
-        vault_data = json.loads(json_vault_data)
+        vault_data = json.loads(json_vault_data.decode('utf-8'))
         data = base64.b64decode(vault_data[u'data'].encode('utf-8'))
 
         encrypted_key = None
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -1101,7 +1101,8 @@ class JSONServerProxy(object):
         )
 
         try:
-            response = json_decode_binary(json.loads(response.decode('ascii')))
+            response = json_decode_binary(
+                json.loads(response.decode('utf-8')))
         except ValueError as e:
             raise JSONError(error=str(e))
 
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -1252,8 +1252,9 @@ class LDAPClient(object):
                 value = u'\\'.join(
                     value[i:i+2] for i in six.moves.range(-2, len(value), 2))
             else:
-                value = value_to_utf8(value)
+                value = six.text_type(value)
                 value = ldap.filter.escape_filter_chars(value)
+
             if not exact:
                 template = '%s'
                 if leading_wildcard:
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -887,7 +887,8 @@ class LDAPClient(object):
         elif isinstance(val, tuple):
             return tuple(self.encode(m) for m in val)
         elif isinstance(val, dict):
-            dct = dict((self.encode(k), self.encode(v)) for k, v in val.items())
+            # key in dict must be str not bytes
+            dct = dict((k, self.encode(v)) for k, v in val.items())
             return dct
         elif isinstance(val, datetime.datetime):
             return val.strftime(LDAP_GENERALIZED_TIME_FORMAT)
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -657,13 +657,12 @@ class CAInstance(DogtagInstance):
         Used when setting up replication
         """
         # Add the new RA cert to the database in /etc/httpd/alias
-        (agent_fd, agent_name) = tempfile.mkstemp()
-        os.write(agent_fd, self.dm_password)
-        os.close(agent_fd)
-        try:
-            import_pkcs12(rafile, agent_name, self.ra_agent_db, self.ra_agent_pwd)
-        finally:
-            os.remove(agent_name)
+        with tempfile.NamedTemporaryFile(mode="w") as agent_file:
+            agent_file.write(self.dm_password)
+            agent_file.flush()
+
+            import_pkcs12(
+                rafile, agent_file.name, self.ra_agent_db, self.ra_agent_pwd)
 
         self.configure_agent_renewal()
 
@@ -759,10 +758,9 @@ class CAInstance(DogtagInstance):
 
         ca_dn = DN(self.ca_subject)
         for cert in certlist:
-            try:
-                chain_fd, chain_name = tempfile.mkstemp()
-                os.write(chain_fd, cert)
-                os.close(chain_fd)
+            with tempfile.NamedTemporaryFile(mode="w") as chain_file:
+                chain_file.write(cert)
+                chain_file.flush()
                 (_rdn, subject_dn) = certs.get_cert_nickname(cert)
                 if subject_dn == ca_dn:
                     nick = get_ca_nickname(self.realm)
@@ -772,10 +770,8 @@ class CAInstance(DogtagInstance):
                     trust_flags = ',,'
                 self.__run_certutil(
                     ['-A', '-t', trust_flags, '-n', nick, '-a',
-                     '-i', chain_name]
+                     '-i', chain_file.name]
                 )
-            finally:
-                os.remove(chain_name)
 
         # Restore NSS trust flags of all previously existing certificates
         for nick, trust_flags in cert_backup_list:
@@ -783,13 +779,15 @@ class CAInstance(DogtagInstance):
 
     def __request_ra_certificate(self):
         # create a temp file storing the pwd
-        (agent_fd, agent_pwdfile) = tempfile.mkstemp(dir=paths.VAR_LIB_IPA)
-        os.write(agent_fd, self.admin_password)
-        os.close(agent_fd)
+        agent_file = tempfile.NamedTemporaryFile(
+            mode="w", dir=paths.VAR_LIB_IPA, delete=False)
+        agent_file.write(self.admin_password)
+        agent_file.close()
 
         # create a temp pem file storing the CA chain
-        (chain_fd, chain_file) = tempfile.mkstemp(dir=paths.VAR_LIB_IPA)
-        os.close(chain_fd)
+        chain_file = tempfile.NamedTemporaryFile(
+            mode="w", dir=paths.VAR_LIB_IPA, delete=False)
+        chain_file.close()
 
         chain = self.__get_ca_chain()
         data = base64.b64decode(chain)
@@ -799,17 +797,17 @@ class CAInstance(DogtagInstance):
              "-inform",
              "DER",
              "-print_certs",
-             "-out", chain_file,
+             "-out", chain_file.name,
              ], stdin=data, capture_output=False)
 
         agent_args = [paths.DOGTAG_IPA_CA_RENEW_AGENT_SUBMIT,
                       "--dbdir", self.agent_db,
                       "--nickname", "ipa-ca-agent",
-                      "--cafile", chain_file,
+                      "--cafile", chain_file.name,
                       "--ee-url", 'http://%s:8080/ca/ee/ca/' % self.fqdn,
                       "--agent-url",
                       'https://%s:8443/ca/agent/ca/' % self.fqdn,
-                      "--sslpinfile", agent_pwdfile]
+                      "--sslpinfile", agent_file.name]
         helper = " ".join(agent_args)
 
         # configure certmonger renew agent to use temporary agent cert
@@ -842,8 +840,11 @@ class CAInstance(DogtagInstance):
             certmonger.modify_ca_helper(
                 ipalib.constants.RENEWAL_CA_NAME, old_helper)
             # remove the pwdfile
-            os.remove(agent_pwdfile)
-            os.remove(chain_file)
+            for f in (agent_file, chain_file):
+                try:
+                    os.remove(f.name)
+                except OSError:
+                    pass
 
     def __setup_sign_profile(self):
         # Tell the profile to automatically issue certs for RAs
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -33,6 +33,7 @@ import syslog
 import time
 import tempfile
 
+import six
 # pylint: disable=import-error
 from six.moves.configparser import ConfigParser, RawConfigParser
 # pylint: enable=import-error
@@ -500,7 +501,7 @@ class CAInstance(DogtagInstance):
         # Directory server
         config.set("CA", "pki_ds_ldap_port", "389")
         config.set("CA", "pki_ds_password", self.dm_password)
-        config.set("CA", "pki_ds_base_dn", self.basedn)
+        config.set("CA", "pki_ds_base_dn", six.text_type(self.basedn))
         config.set("CA", "pki_ds_database", "ipaca")
 
         if self.use_ldaps:
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -599,7 +599,7 @@ class CAInstance(DogtagInstance):
         config.set("Tomcat", "pki_ajp_host", "::1")
 
         # Generate configuration file
-        with open(cfg_file, "wb") as f:
+        with open(cfg_file, "w") as f:
             config.write(f)
 
         self.backup_state('installed', True)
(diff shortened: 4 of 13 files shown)


ipapython/ipaldap.py
```diff
@@ -886,7 +886,8 @@ def encode(self, val):
         elif isinstance(val, tuple):
             return tuple(self.encode(m) for m in val)
         elif isinstance(val, dict):
-            dct = dict((self.encode(k), self.encode(v)) for k, v in val.items())
+            # key in dict must be str not bytes
+            dct = dict((unicode(k), self.encode(v)) for k, v in val.items())
```
HonzaCholasta: This should be just `k`, `LDAPEntry` already guarantees the keys are unicode.

ipaserver/install/cainstance.py
```diff
@@ -499,7 +500,7 @@ def __spawn_instance(self):
         # Directory server
         config.set("CA", "pki_ds_ldap_port", "389")
         config.set("CA", "pki_ds_password", self.dm_password)
-        config.set("CA", "pki_ds_base_dn", self.basedn)
+        config.set("CA", "pki_ds_base_dn", six.text_type(self.basedn))
```
HonzaCholasta: Isn't the real issue here that `DN` does not have a proper `__str__()`?

ipaserver/install/dsinstance.py
```diff
@@ -582,14 +582,15 @@ def __update_dse_ldif(self):
             'dse.ldif'
         )
 
-        with tempfile.NamedTemporaryFile(delete=False) as new_dse_ldif:
+        with tempfile.NamedTemporaryFile(
+                mode='w', delete=False) as new_dse_ldif:
```
HonzaCholasta: The default `mode` is `w+b`, so I guess if you just want to change from binary to text mode, `mode` should be `w+t`.

ipaserver/install/httpinstance.py
```diff
@@ -314,9 +315,8 @@ def create_cert_db(self):
 
         # Create the password file for this db
         password = ipautil.ipa_generate_password()
-        f = os.open(pwd_file, os.O_CREAT | os.O_RDWR)
-        os.write(f, password)
-        os.close(f)
+        with io.open(pwd_file, 'w') as f:
```
HonzaCholasta: Nitpick, but could you be more explicit and use `wt` as mode?

ipaserver/install/service.py
```diff
@@ -209,7 +209,7 @@ def _ldap_mod(self, ldif, sub_dict=None, raise_on_err=True,
 
         if dm_password:
             [pw_fd, pw_name] = tempfile.mkstemp()
-            os.write(pw_fd, dm_password)
+            os.write(pw_fd, dm_password.encode())
```
HonzaCholasta: Could you make also this code more Pythonic and replace `mkstemp()` with `TemporaryFile()`?

ipaserver/install/dsinstance.py
tiran: LDIFWriter never reads from its input file. There is no point opening new_dse_ldif for read and write.

ipaserver/install/httpinstance.py
tiran: 'wt' is an uncommon way to spell out write in text mode. 'w' is the canonical way.

HonzaCholasta: Explicit is better than implicit, except when it isn't, okay.

ipaserver/install/dsinstance.py
HonzaCholasta: OK.

ipaserver/install/service.py
mbasti-rh: I can

ipapython/ipaldap.py
mbasti-rh: ok

ipaserver/install/cainstance.py
mbasti-rh: I don't see difference between __str__ implementation in `DNSName` and `DN`, I will investigate OptionParser, maybe it changed in Py3 to enforce strings

ipaserver/install/service.py
tiran: +1 for getting rid of ```mkstemp()``` and ```os.write()```. grep found a couple of more places where a raw FD and ```os.write()```. Feel free to address them in another PR.

ipaserver/install/cainstance.py
```diff
@@ -839,8 +838,11 @@ def __request_ra_certificate(self):
             certmonger.modify_ca_helper(
                 ipalib.constants.RENEWAL_CA_NAME, old_helper)
             # remove the pwdfile
-            os.remove(agent_pwdfile)
-            os.remove(chain_file)
+            for f in (agent_file, chain_file):
+                try:
+                    os.remove(f.name)
+                except IOError:
```
tiran: os.remove() raises an OSError, not an IOError.

ipaserver/install/cainstance.py
```diff
@@ -769,24 +768,24 @@ def __import_ca_chain(self):
...
             certdb.trust_root_cert(nick, trust_flags)
 
     def __request_ra_certificate(self):
         # create a temp file storing the pwd
-        (agent_fd, agent_pwdfile) = tempfile.mkstemp(dir=paths.VAR_LIB_IPA)
-        os.write(agent_fd, self.admin_password)
-        os.close(agent_fd)
+        agent_file = tempfile.NamedTemporaryFile(
+            mode="w", dir=paths.VAR_LIB_IPA, delete=False)
+        agent_file.write(self.admin_password)
```
tiran: Please use with statement here.

ipaserver/install/cainstance.py
```diff
@@ -769,24 +768,24 @@ def __import_ca_chain(self):
...
+        agent_file.close()
 
         # create a temp pem file storing the CA chain
-        (chain_fd, chain_file) = tempfile.mkstemp(dir=paths.VAR_LIB_IPA)
-        os.close(chain_fd)
+        chain_file = tempfile.NamedTemporaryFile(
```
tiran: with statement

ipaserver/install/service.py
```diff
@@ -208,10 +208,11 @@ def _ldap_mod(self, ldif, sub_dict=None, raise_on_err=True,
...
         if dm_password:
-            [pw_fd, pw_name] = tempfile.mkstemp()
-            os.write(pw_fd, dm_password)
-            os.close(pw_fd)
-            auth_parms = ["-x", "-D", "cn=Directory Manager", "-y", pw_name]
+            with tempfile.NamedTemporaryFile(
+                    mode='w', delete=False) as pw_file:
+                pw_name = pw_file.name
+                pw_file.write(dm_password)
+                auth_parms = ["-x", "-D", "cn=Directory Manager", "-y", pw_name]
```
tiran: Please move the auth_param part out of the context manager. I got confused by it for a sec. It looked like you omitted ```pw_file.flush()``` like in other places. After a closer look it turns out the file name is used after the file has been closed.

```
with tempfile.NamedTemporaryFile(mode='w', delete=False) as pw_file:
    pw_file.write(dm_password)
pw_name = pw_file.name
auth_parms = ["-x", "-D", "cn=Directory Manager", "-y", pw_name]
```

ipaserver/install/cainstance.py
mbasti-rh: To use `with` I have to indent or split a lot of code in that functions, is it worth it?

ipaserver/install/cainstance.py
mbasti-rh: the same as above

ipaserver/install/cainstance.py
mbasti-rh: fixed, thanks

ipaserver/install/service.py
mbasti-rh: done

mbasti-rh:
@HonzaCholasta @tiran bump for review

ipaserver/install/cainstance.py
HonzaCholasta: It's just a single function, so IMO it's OK.

install/certmonger/dogtag-ipa-ca-renew-agent-submit
```diff
@@ -135,7 +135,7 @@ def call_handler(_handler, *args, **kwargs):
         cookie = os.environ.pop('CERTMONGER_CA_COOKIE', None)
         if cookie is not None:
             try:
-                context = json.loads(cookie)
+                context = json.loads(ipautil.decode_json(cookie))
```
HonzaCholasta: This is not necessary as `cookie` is always an instance of `str` here in Python 3.

ipaclient/remote_plugins/schema.py
```diff
@@ -522,7 +522,7 @@ def _write_schema(self, fingerprint):
 
     def _read(self, path):
         with zipfile.ZipFile(self._file, 'r') as zf:
-            return json.loads(zf.read(path).decode('utf-8'))
+            return json.loads(decode_json(zf.read(path)))
```
HonzaCholasta: We don't need to guess the encoding here, the original code is correct.

ipalib/rpc.py
```diff
@@ -1101,7 +1101,8 @@ def __request(self, name, args):
         )
 
         try:
-            response = json_decode_binary(json.loads(response.decode('ascii')))
+            response = json_decode_binary(
+                json.loads(ipautil.decode_json(response)))
```
HonzaCholasta: `response` is always `bytes`, so we don't need to guess the encoding here.

ipaclient/plugins/vault.py
```diff
@@ -969,7 +970,7 @@ def forward(self, *args, **options):
         json_vault_data = decoding_ctx.cipher_op(wrapped_vault_data)\
             + decoding_ctx.digest_final()
 
-        vault_data = json.loads(json_vault_data)
+        vault_data = json.loads(decode_json(json_vault_data))
```
HonzaCholasta: It looks like `json_vault_data` is always an instance of `bytes`, so instead of guessing the encoding, we should always decode it as UTF-8. Ditto for encoding `json_vault_data` in `vault_archive.forward()`.

ipapython/ipaldap.py
```diff
@@ -1244,11 +1245,11 @@ def make_filter_from_attr(
             ]
             return cls.combine_filters(flts, rules)
         elif value is not None:
-            if isinstance(value, bytes):
-                if six.PY3:
-                    value = value.decode('raw_unicode_escape')
+            if isinstance(value, six.binary_type):
```
HonzaCholasta: I would prefer if you kept `bytes` here, it's the same as `six.binary_type`.

ipaserver/rpcserver.py
```diff
@@ -493,7 +493,7 @@ def marshal(self, result, error, _id=None,
 
     def unmarshal(self, data):
         try:
-            d = json.loads(data)
+            d = json.loads(ipautil.decode_json(data))
```
HonzaCholasta: `data` is read from `wsgi.input` and according to PEP 3333, reading from `wsgi.input` returns `bytes`, so we don't need to guess here and can always decode as UTF-8.

tiran: You are assuming that bytes always means UTF-8 encoded text. Although it's the common case for JSON, it is not necessarily correct. JSON can also be encoded as UTF-16 LE/BE or UTF-32 LE/BE.

install/certmonger/dogtag-ipa-ca-renew-agent-submit
mbasti-rh: ok

ipaclient/remote_plugins/schema.py
mbasti-rh: bytes doesn't say anything about encoding, but AFAIK it is part of IPA so we can assume it is always  'utf-8' encoded

ipalib/rpc.py
mbasti-rh: I'm quite curious why it is only ascii in original code

ipapython/ipaldap.py
mbasti-rh: ok

mbasti-rh:
We had discussion with @HonzaCholasta, and IPA framework only expects that everything is UTF-8 only, so even in case we parse UTF-32 properly, framework will answer by UTF-8 encoding. Maybe we should rather validate if input is utf-8 compatible earlier and send proper public error.

ipaserver/rpcserver.py
HonzaCholasta: We currently support only UTF-8:
```
$ echo '{"method": "ping", "params": [[], {}]}' | iconv -f utf-8 -t utf-16 | curl https://$HOSTNAME/ipa/json --cacert /etc/ipa/ca.crt --negotiate -u : -e https://$HOSTNAME/ipa/json -H 'Content-Type: application/json' --data-binary @-
{
    "error": {
        "code": 909, 
        "data": {
            "error": "No JSON object could be decoded"
        }, 
        "message": "Invalid JSON-RPC request: No JSON object could be decoded", 
        "name": "JSONError"
    }, 
    "id": null, 
    "principal": "admin@ABC.IDM.LAB.ENG.BRQ.REDHAT.COM", 
    "result": null, 
    "version": "4.4.90.dev201701171151+git7aee0d4"
}
```
Adding support for other encodings is not in the scope of this effort, and should be based on HTTP headers rather than guesswork anyway.

mbasti-rh: @tiran do you agree to hardcode everything into 'utf-8' for now? (we can open a ticket to support more encodings)

ipaclient/remote_plugins/schema.py
HonzaCholasta: Yep, that's right.

tiran:
Let's reiterate. It's obviously wrong to assume that request data such as JSON are encoded as UTF-8. It can be just any encoding. Outside the Western world, JSON and XML are often encoded as UTF-16.

That doesn't mean we have to support other encodings than UTF-8 right now. It's fine to restrict requests and responses to UTF-8 as only supported encoding. The check should be performed early in the WSGI layer. A client sends can send the request type as part of the content type. The framework should check for the presence of an encoding hint and refuse encodings that are ```encoding.lower() not in {'utf8', 'utf-8'}```.

MartinBasti:
I left json_decode() only in places where external JSON request are coming, all other internal usages of JSON should be in utf-8 encoding.

Other requests are out of scope of this PR and should be resolved in separate tickets/PRs

tiran:
I have opened ticket https://fedorahosted.org/freeipa/ticket/6624 to track the matter.

ipaserver/install/cainstance.py
MartinBasti: It looks that it really takes only string in py3

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/23239bccc18f2fdc10431134a1c6a777f450704e
https://fedorahosted.org/freeipa/changeset/e0641092770530e3b93c00de415172751d031210
https://fedorahosted.org/freeipa/changeset/63b5d4a8594c5c6bc9ade69996fbbc1bcf19a2bf
https://fedorahosted.org/freeipa/changeset/bbe8849a654ed0764e1834f24d1837df41a79881
https://fedorahosted.org/freeipa/changeset/7ae5e5f66919141821fdffbd6f8683c6d7afddd7
https://fedorahosted.org/freeipa/changeset/0d4074b4f1a57ed6545d819aa5a48e4b35237568
https://fedorahosted.org/freeipa/changeset/2547bca8df69e6c4d5f4c67a63fbc3c06ccc95c6
https://fedorahosted.org/freeipa/changeset/232ceed5bbfb0afa45078d8e95b84dabe4d7cafd
https://fedorahosted.org/freeipa/changeset/c0b5c6709d9e3a51117994fc8b605ba54e6263d7
https://fedorahosted.org/freeipa/changeset/51578882fc8456788d69a57de1a1d45ead58ba14
https://fedorahosted.org/freeipa/changeset/0a1d7f2e01819ad6e4a19d0416b3a01883dea7d0
https://fedorahosted.org/freeipa/changeset/4b148c8ca3d022020fa6caccf02729c090c8dbcb
https://fedorahosted.org/freeipa/changeset/746d4ffc583a847834a592150644fa4270486c89
https://fedorahosted.org/freeipa/changeset/1e0f98a146ecedf84b8e3e07fbd41a897ddd399d
https://fedorahosted.org/freeipa/changeset/0eb5a0e0ec2d232d2921ae5f9e8d0885146a5610
https://fedorahosted.org/freeipa/changeset/18337bf7f7c31a47fe0c7280f82fca043b548bd5

PR #383, opened 2017-01-09
Remove duplicated step from DS install

mbasti-rh:
"Adding SASL mappings.." is duplicated step in __common_setup in DS
instance and should be removed.


stlaz:
The tests passed, ACK.

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/083b4241d287a731e2cf7fed5c61b30da52a8e37

PR #384, opened 2017-01-10
Add fix for user prompt in dnsrecord-add

Akasurde:
Fix added to skip optional parameter in dnsrecord-add
interactive prompts

Fixes https://fedorahosted.org/freeipa/ticket/6457


HonzaCholasta:
I'm afraid this is not a proper fix, as it introduces a regression in CLI behavior.

A proper fix would be to use correct argument names - in the trace in the ticket it says `a_part_create_reverse`, but it should be `a_extra_create_reverse`.

PR #385, opened 2017-01-10
Generate sha256 ssh pubkey fingerprints for hosts

stlaz:
Replace md5 with sha256 for host ssh pubkey fingerprints. MD5 is disabled in FIPS mode, newer versions of OpenSSH print SHA256 public key fingeprint anyway.

### DIFF
--- a/ipalib/util.py
+++ b/ipalib/util.py
@@ -301,7 +301,7 @@ def convert_sshpubkey_post(entry_attrs):
         except (ValueError, UnicodeDecodeError):
             continue
 
-        fp = pubkey.fingerprint_hex_md5()
+        fp = pubkey.fingerprint_hex_sha256()
         comment = pubkey.comment()
         if comment:
             fp = u'%s %s' % (fp, comment)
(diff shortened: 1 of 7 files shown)


tiran:
What's the migration plan for existing SSHFP records? Are there any supported versions of OpenSSH or other SSH client that do not support SSHFP with SHA256? Would it make sense to run a hybrid mode for a while (SHA256 and MD5 records unless FIPS is enabled)?

stlaz:
@tiran Which SSHFP records do you mean?

tiran:
Your change influenced the value of ```entry_attrs['sshpubkeyfp']``` in ```convert_sshpubkey_post```. Is the value only used in UI or does it affect data in LDAP like DNS SSHFP records? I tracked down some code paths and it looks like the pubkey fingerprint isn't stored in LDAP. The DNS plugin uses different method to calculate SSHFP records. Am I right to assume that this change only affects UI?

stlaz:
@tiran Yes, exactly, this is only a UI thing.

tiran:
@stlaz I'm sorry, go ahead and ignore what I said! :)

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/721105c53de6fbc0abc7799ec7f48920e02089bd

PR #386, opened 2017-01-10
Tests: Add tree root domain role in legacy client tests

gkaihorodova:
The errors that are raised in the legacy client tests are because it inherits test cases that use tree root domain role from trust test, but in legacy client tests the optional_extra_roles does not contain the ad_treedomain.

https://fedorahosted.org/freeipa/ticket/6600


martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/822a119100f8ab93aacdb14b982609f1dc69531d
ipa-4-4:
https://fedorahosted.org/freeipa/changeset/52527d6323eec1a2ae4bf01dd64412a3822c516d

PR #388, opened 2017-01-11
py3: enable py3 pylint

mbasti-rh:
We should run pylint in both python2 and python3 versions

### DIFF
--- a/.test_runner_config.yaml
+++ b/.test_runner_config.yaml
@@ -40,7 +40,8 @@ steps:
     -a ${server_password} --setup-dns --auto-forwarders
   - ipa-kra-install -p ${server_password}
   lint:
-  - make V=0 lint
+  - PYTHON=/usr/bin/python2 make V=0 lint
+  - PYTHON=/usr/bin/python3 make V=0 pylint
   prepare_tests:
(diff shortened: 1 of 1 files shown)


.test_runner_config.yaml
```diff
@@ -40,7 +40,8 @@ steps:
     -a ${server_password} --setup-dns --auto-forwarders
   - ipa-kra-install -p ${server_password}
   lint:
-  - make V=0 lint
+  - PYTHON=/usr/bin/python2 make V=0 lint
+  - PYTHON=/usr/bin/python3 make V=0 lint
```
tiran: please make the python3 target ```pylint``` only. I just realized that lint also runs the i18n and JS linters.

mbasti-rh: I don't think that running those linters twice is an issue, but as you wish I updated PR

tiran:
It's just a minor performance improvements to speed up CI.

mbasti-rh:
Lint part of travis-CI passed

mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/d648c6a6925298a1db0c61381d72b6c4d0500c10

PR #389, opened 2017-01-11
Fix build in mock

lslebodn:


tiran:
Thanks, your PR fixes some concerns of my ticket https://fedorahosted.org/freeipa/ticket/6604.

lslebodn:
I updated doc help do jslint in latest version

lslebodn:
@tiran or @pvomacka, @tomaskrizek  It would be good if you could test/ack the latest version of the patch.
Because currently it is not possible to build freeIPA with upstream spec file in mock; which blocks reasonable testing with static analysers

tomaskrizek:
Thanks for the fix and explanation!

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/5c18feaa206bbaee692fc3640b7b79c8d9d6a638
https://fedorahosted.org/freeipa/changeset/3f91469f327d8d9f3b27e0b67c54a4f47ad845c1
https://fedorahosted.org/freeipa/changeset/b82d285a4a75e11cc9291ecca12d2fcc26f43ed1

PR #391, opened 2017-01-12
ipapython: Add dependencies on version.py

tiran:
install-exec and bdist_wheel also depend on version.py. Let's ensure
that version.py is correctly generated when installing or building
packages.

Yes, make is clever and correctly merges dependencies with rules from
included make files.


mbasti-rh:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/504f4417070a308ef54b8f98ff25d02c6604a6f6

PR #393, opened 2017-01-12
[Py3] allow to run wsgi - part1

MartinBasti:
With these patches we can run commands with server running on py3

Note: to use py3 install module `python3-mod_wsgi` that enables py3 wsgi automatically

Note: this may or may not depend (I haven't tested) on my PR #382 so it contains all patches, will be rebased when PR #382 merged

first WSGI related commit is 
- py3: session.py decode server name to str ￼…			7ff4b83

### DIFF
--- a/ipalib/x509.py
+++ b/ipalib/x509.py
@@ -251,13 +251,22 @@ def normalize_certificate(rawcert):
 
     rawcert = strip_header(rawcert)
 
-    if util.isvalid_base64(rawcert):
-        try:
-            dercert = base64.b64decode(rawcert)
-        except Exception as e:
-            raise errors.Base64DecodeError(reason=str(e))
-    else:
+    try:
+        if isinstance(rawcert, bytes):
+            # base64 must work with utf-8, otherwise it is raw bin certificate
+            decoded_cert = rawcert.decode('utf-8')
+        else:
+            decoded_cert = rawcert
+    except UnicodeDecodeError:
         dercert = rawcert
+    else:
+        if util.isvalid_base64(decoded_cert):
+            try:
+                dercert = base64.b64decode(decoded_cert)
+            except Exception as e:
+                raise errors.Base64DecodeError(reason=str(e))
+        else:
+            dercert = rawcert
 
     # At this point we should have a DER certificate.
     # Attempt to decode it.
--- a/ipalib/x509.py
+++ b/ipalib/x509.py
@@ -85,12 +85,16 @@ def strip_header(pem):
     """
     Remove the header and footer from a certificate.
     """
-    s = pem.find("-----BEGIN CERTIFICATE-----")
-    if s >= 0:
-        e = pem.find("-----END CERTIFICATE-----")
-        pem = pem[s+27:e]
-
-    return pem
+    regexp = (
+        u"^-----BEGIN CERTIFICATE-----(.*?)-----END CERTIFICATE-----"
+    )
+    if isinstance(pem, bytes):
+        regexp = regexp.encode('ascii')
+    s = re.search(regexp, pem, re.MULTILINE | re.DOTALL)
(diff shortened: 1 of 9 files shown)


tiran:
@MartinBasti cert tests are failing. I have restarted the failing job. Let's see if the error persists or was just caused by a Travis hick up.

MartinBasti:
@tiran we found the issue that caues random test fails, @HonzaCholasta will provide PR with fix, that should be pushed before these commits

MartinBasti:
Rebased, ready to be reviewed

ipaserver/session.py
```diff
@@ -21,6 +21,7 @@
 import os
 import re
 import time
+import io
```
tiran: Is there any particular reason you are using the io module here? In Python 3 ```io.open is open```. In Python 2 the io module's main use is to auto-encode UTF-8 text. It's not wrong, I'm just curious.

ipalib/x509.py
```diff
@@ -85,9 +85,16 @@ def strip_header(pem):
     """
     Remove the header and footer from a certificate.
     """
-    s = pem.find("-----BEGIN CERTIFICATE-----")
+    if isinstance(pem, bytes):
```
tiran: How about a regular expression?
```
def strip_header(pem):
    regexp = u"^-----BEGIN CERTIFICATE-----\s+(.*?)\s+^-----END CERTIFICATE-----"
    if isinstance(pem, bytes):
        regexp = regexp.encode('ascii')
    s = re.search(regexp, pem, re.MULTILINE | re.DOTALL)
    if s is not None:
        return s.group(1)
    else:
        return pem
```

ipaserver/session.py
MartinBasti: I wanted to use the same

ipalib/x509.py
MartinBasti: Good idea, thanks

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/a9fec1de1aa2b3c0f4c4ec6eff25ff2e75c774b0
https://fedorahosted.org/freeipa/changeset/9739d0354a8ac5fd357f7d131b3f75aa05df058b
https://fedorahosted.org/freeipa/changeset/7e8eb533752bbf5e2f05ec6bfb0ffefa8e9dcddf
https://fedorahosted.org/freeipa/changeset/35e135c4e3a7f0bf21ed4c838b8f76b43701a047
https://fedorahosted.org/freeipa/changeset/cca9aa43e146f15e235eee1197209d0ca88eb39c
https://fedorahosted.org/freeipa/changeset/aa036e5f332ef0b1ebbff6b824e236b1eeaf076e
https://fedorahosted.org/freeipa/changeset/dd3d9f1ca61946ea5d7daa17ba1d8a883922d526
https://fedorahosted.org/freeipa/changeset/49333058c869dd4bd654a7974e6e144ffd3f0dc3
https://fedorahosted.org/freeipa/changeset/b37d18288d40b4ec0b5a8df676456e09ae5f26c1
https://fedorahosted.org/freeipa/changeset/deaf9ae2473833dacb64c4961db3ae9f7c570ebd
https://fedorahosted.org/freeipa/changeset/1023cfebff99af165212dee94290a05754297270
https://fedorahosted.org/freeipa/changeset/d5ab0637fe89cbcb61491fe08b7376aeaf7ccdb8
https://fedorahosted.org/freeipa/changeset/47e76e16ef2e5d714881f3cce204611a95b4e5c8
https://fedorahosted.org/freeipa/changeset/b8d6524d43dd0667184aebc79fb77a9b8a46939a
https://fedorahosted.org/freeipa/changeset/980c8a5f9e4ccbcd3c11def9cab33d0e61e945ae

PR #395, opened 2017-01-13
Configure PKI ajp redirection to use "localhost" instead of "::1"

flo-renaud:
When ipa-server-install configures PKI, it provides a configuration file
with the parameter pki_ajp_host set to ::1. This parameter is used to configure
Tomcat redirection in /etc/pki/pki-tomcat/server.xml:
    <Connector port="8009"
            protocol="AJP/1.3"
            redirectPort="8443"
            address="::1" />
ie all requests to port 8009 are redirected to port 8443 on address ::1.

If the /etc/hosts config file does not define ::1 for localhost, then AJP
redirection fails and replica install is not able to request a certificate
for the replica. Using "localhost" instead works with IPv4 or IPv6.

https://fedorahosted.org/freeipa/ticket/6575

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -246,8 +246,8 @@ Requires(post): systemd-units
 Requires: selinux-policy >= %{selinux_policy_version}
 Requires(post): selinux-policy-base >= %{selinux_policy_version}
 Requires: slapi-nis >= %{slapi_nis_version}
-Requires: pki-ca >= 10.3.5-6
-Requires: pki-kra >= 10.3.5-6
+Requires: pki-ca >= 10.3.5-11
+Requires: pki-kra >= 10.3.5-11
 Requires(preun): python systemd-units
 Requires(postun): python systemd-units
 Requires: policycoreutils >= 2.1.12-5
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -594,10 +594,6 @@ class CAInstance(DogtagInstance):
             config.set("CA", "pki_external_ca_cert_chain_path", cert_chain_file.name)
             config.set("CA", "pki_external_step_two", "True")
 
-        # PKI IPv6 Configuration
-        config.add_section("Tomcat")
-        config.set("Tomcat", "pki_ajp_host", "::1")
-
         # Generate configuration file
         with open(cfg_file, "w") as f:
             config.write(f)


tomaskrizek:
The fix solves [ticket #6575](https://fedorahosted.org/freeipa/ticket/6575), but I once again encountered [ticket #4291](https://fedorahosted.org/freeipa/ticket/4291): In pure IPv6 environments, the CA fails to start at the end of IPA server installation. I'm not sure why that happens, since `localhost` does correctly resolve to `::1` when using `getent hosts localhost`.

Flo is investigating and trying to reproduce the issue.

tiran:
Bad news, you are out of luck.

Dogtag uses its own LDAP connector, which in turn uses JSS (NSS bindings for Java) to provide TLS for LDAP. SSLSocket from ```org.mozilla.jss``` does not support ```AF_INET6``` and is therefore limited to IPv4 connections, https://hg.mozilla.org/projects/jss/file/1a96a08e6f3d/org/mozilla/jss/ssl/SSLSocket.c#l443

The experimental branch of JSS has IPv6 support, https://hg.mozilla.org/projects/jss/file/c76470016016/org/mozilla/jss/ssl/SSLSocket.c#l593, though.

tiran:
I have created ticket https://fedorahosted.org/pki/ticket/2575 to track the issue.

pvoborni:
Btw our goal is not to make pure IPv6 working - this was not tested even before the regression.

pvoborni:
Possible ways how to fix upgrades are outlined in https://bugzilla.redhat.com/show_bug.cgi?id=1398600#c48 and comment 49.

tomaskrizek:
@flo You're right, I'm able to install the IPA server in IPv6 env now. I probably forgot some configuration beforehand.

@tiran That's odd. If JSS does not support IPv6 at all, why does `::1` work when IPv6 is enabled?

flo-renaud:
Please wait before merging this PR.

@pvoborni Endi suggests 2 possible strategies for the upgrade fix: either in IPA or in PKI. We need to pick one, and depending on the choice modify this PR accordingly.

tiran:
@tomaskrizek Upstream doesn't support IPv6. @nkinder told me that our Fedora packages contan several patches. The patches are not yet in upstream. One of the patches adds IPv6 support. I have not checked if the patch provides IPv6 for both client and server sockets or just for SSLSocket.

flo-renaud:
This PR has been modified to be consistent with PKI fix for [2570](https://fedorahosted.org/pki/ticket/2570). PKI now defines by default the AJP redirection to "localhost", meaning that we do not need any more to override this setting. Upgrade is also handled by PKI.

stlaz:
Removed the ACK label since this is not yet reviewed.

tomaskrizek:
Since the bug is completely fixed on the PKI side, shouldn't we bump the `Requires` to require the fixed version of PKI?

Installation in IPV6-only environment will not work without the updated PKI, since 127.0.0.1 was used as a default before [3a49b9b3738befc03914b0a96aad61f9650fb935](https://git.fedorahosted.org/cgit/pki.git/commit/?id=3a49b9b3738befc03914b0a96aad61f9650fb935).

pvoborni:
Yes, but in different patch please.  PKI with the fix was not released yet. So it should not block review of this patch. We can leave the ticket open until it is bumped.

tomaskrizek:
**ACK** for z-stream with the patched PKI.

Waiting for the PKI release and bump of `Requires` to ack and merge upstream.

HonzaCholasta:
@pvoborni, there is no benefit in bumping `Requires` in a separate patch, as this patch is blocked by it anyway and cannot be pushed as is. Please update the current patch once the fixed `pki-core` is released.

flo-renaud:
Hi,
PR updated with dependency on pki 10.3.5-11 (note that this package is currently available in fedora updates-testing only).

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/eaa87c75b9f57500265b2dc9480b996b2b92e1e3

MartinBasti:
Please create a backport PR for IPA 4.4.x

PR #396, opened 2017-01-13
Explicitly remove support of SSLv2

stlaz:
It was possible to set tls_version_min/max to 'ssl2', even though
newer versions of NSS will fail to set this as a valid TLS version.
This patch explicitly checks for deprecated TLS versions prior to
creating a TLS connection.

https://fedorahosted.org/freeipa/ticket/6607

### DIFF
--- a/ipalib/config.py
+++ b/ipalib/config.py
@@ -41,8 +41,11 @@ from six.moves.configparser import RawConfigParser, ParsingError
 
 from ipapython.dn import DN
 from ipalib.base import check_name
-from ipalib.constants import CONFIG_SECTION
-from ipalib.constants import OVERRIDE_ERROR, SET_ERROR, DEL_ERROR
+from ipalib.constants import (
+    CONFIG_SECTION,
+    OVERRIDE_ERROR, SET_ERROR, DEL_ERROR,
+    TLS_VERSIONS
+)
 from ipalib import errors
 
 if six.PY3:
@@ -577,6 +580,26 @@ class Env(object):
                 self.server = parsed.netloc
 
         self._merge(**defaults)
+
+        # set the best known TLS version if min/max versions are not set
+        if 'tls_version_min' not in self:
+            self.tls_version_min = TLS_VERSIONS[-1]
+        elif self.tls_version_min not in TLS_VERSIONS:
+            raise errors.EnvironmentError(
+                "Unknown TLS version '{ver}' set in tls_version_min."
+                .format(ver=self.tls_version_min))
+
+        if 'tls_version_max' not in self:
+            self.tls_version_max = TLS_VERSIONS[-1]
+        elif self.tls_version_max not in TLS_VERSIONS:
+            raise errors.EnvironmentError(
+                "Unknown TLS version '{ver}' set in tls_version_max."
+                .format(ver=self.tls_version_max))
+
+        if self.tls_version_max < self.tls_version_min:
+            raise errors.EnvironmentError(
+                "tls_version_min is set to a higher TLS version than "
+                "tls_version_max.")
 
     def _finalize(self, **lastchance):
         """
--- a/ipapython/nsslib.py
+++ b/ipapython/nsslib.py
@@ -23,6 +23,8 @@ from __future__ import print_function
 import getpass
 import socket
 from ipapython.ipa_log_manager import root_logger
+from ipapython.ipa_log_manager import log_mgr
+from ipalib.constants import TLS_VERSIONS, TLS_VERSION_MINIMAL
 
 from nss.error import NSPRError
 import nss.io as io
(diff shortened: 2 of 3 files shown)


tiran:
* What is the point of supporting SSL 3.0, TLS 1.0 and TLS 1.1 on the client side these days? How about we remove ancient and potentially dangerous TLS versions completely?
* Would it be possible to validate the values during API initialization?

stlaz:
- I think we may need to discuss the support on Monday meeting, generally I think SSL 3.0 and TLS 1.0 should not be supported but there might be troubles with connectivity to legacy IPA servers
- Yes, although in that case we would have to fail instead of falling back to "reasonable defaults" as Env object attribute values cannot be changed once set

ipalib/config.py
```diff
@@ -497,6 +500,17 @@ def _bootstrap(self, **overrides):
...
 
+        if self.tls_version_min not in TLS_VERSIONS:
+            raise errors.EnvironmentError(
+                "Unknown TLS version '{ver}' set in tls_version_min."
+                .format(self.tls_version_min))
+
+        if self.tls_version_min not in TLS_VERSIONS:
+            raise errors.EnvironmentError(
+                "Unknown TLS version '{ver}' set in tls_version_min."
+                .format(self.tls_version_min))
```
HonzaCholasta: I think you meant `tls_version_max`.

ipapython/nsslib.py
```diff
@@ -129,6 +131,45 @@ def client_auth_data_callback(ca_names, chosen_nickname, password, certdb):
...
+        raise InvocationError("tls_version_max ('{val}') is not a known "
+                              "TLS version.".format(val=tls_version_max))
+
+    if min_version_idx > max_version_idx:
+        raise InvocationError("tls_version_min is higher than "
+                              "tls_version_max.")
+
+    if min_version_idx < min_allowed_idx:
+        min_version_idx = min_allowed_idx
+        root_logger.warning("tls_version_min set too low ('{old}'),"
```
HonzaCholasta: Please use a module-specific logger rather than the root logger:
```python

logger = log_mgr.get_logger(__name__)
```

ipapython/nsslib.py
```diff
@@ -129,6 +131,45 @@ def client_auth_data_callback(ca_names, chosen_nickname, password, certdb):
...
+def get_proper_tls_version_span(tls_version_min, tls_version_max):
+    # every tls version from `tls_versions` prior to min_allowed_idx
+    # is deprecated in IPA
+    min_allowed_idx = TLS_VERSIONS.index(TLS_VERSION_MINIMAL)
+
+    try:
+        min_version_idx = TLS_VERSIONS.index(tls_version_min)
+    except ValueError:
+        raise InvocationError("tls_version_min ('{val}') is not a known "
+                              "TLS version.".format(val=tls_version_min))
```
HonzaCholasta: `InvocationError` is not right. I think `RuntimeError` should be OK.

ipalib/config.py
stlaz: Yup

ipapython/nsslib.py
stlaz: Ok

ipapython/nsslib.py
stlaz: Shall I replace all appearances of `root_logger` in the module, then?

stlaz:
Did not realize merging to Env from default constants was happening in the end of `_finalize_core()`, moved the checks in config.py accordingly.
Also, for some reason, github shows `root_logger` issue as solved but it's not - should all `root_logger` appearances be replaces by a module-own logger?

HonzaCholasta:
@stlaz, you don't have to replace `root_logger` in old code, but don't use it in new code.

stlaz:
Done. Also added a docstring to the `get_proper_tls_version_span()` function.

HonzaCholasta:
LGTM.

tomaskrizek:
Please update the commit title and description to make it clear that it also removes support of SSLv3.

ipapython/nsslib.py
```diff
@@ -129,6 +134,55 @@ def client_auth_data_callback(ca_names, chosen_nickname, password, certdb):
...
+        min_version_idx = TLS_VERSIONS.index(tls_version_min)
+    except ValueError:
+        raise ValueError("tls_version_min ('{val}') is not a known "
+                         "TLS version.".format(val=tls_version_min))
+
+    try:
+        max_version_idx = TLS_VERSIONS.index(tls_version_max)
+    except ValueError:
+        raise ValueError("tls_version_max ('{val}') is not a known "
+                         "TLS version.".format(val=tls_version_max))
```
stlaz:
```Python
max_version_idx = get_proper_tls_version(
        max_version_idx, "max_version_idx", min_version_idx=min_version_idx)
```
I would rather avoid such overkill.

tomaskrizek: ok

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/ac6f573a3014aa09811ca1559d470afe75eadbec

PR #397, opened 2017-01-17
Improve wheel building and provide ipaserver wheel for local testing

tiran:
The PR improve wheel bundle building and allows ipaserver bundles for local testing
with instrumented build of Python. Debug builds and instrumented builds can have a different binary interface (ABI). For example it is useful for dtrace or test installations in a virtual env. ipaplatform and ipaserver will not be uploaded to PyPI, though.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -256,6 +256,17 @@ endif  # WITH_JSLINT
 WHEELDISTDIR = $(top_builddir)/dist/wheels
 WHEELBUNDLEDIR = $(top_builddir)/dist/bundle
 
+@MK_IFEQ@ ($(IPA_SERVER_WHEELS),1)
+    IPA_WHEEL_PACKAGES @MK_ASSIGN@ $(IPACLIENT_SUBDIRS) ipaplatform ipaserver
+    IPA_OMIT_INSTALL @MK_ASSIGN@ 0
+@MK_ELSE@
+    IPA_WHEEL_PACKAGES @MK_ASSIGN@ $(IPACLIENT_SUBDIRS)
+    IPA_OMIT_INSTALL @MK_ASSIGN@ 1
+@MK_ENDIF@
+
+# additional wheels for bundle, e.g. IPA_EXTRA_WHEELS="ipatests[webui] pylint"
+IPA_EXTRA_WHEELS=
+
 $(WHEELDISTDIR):
 	mkdir -p $(WHEELDISTDIR)
 
@@ -263,19 +274,22 @@ $(WHEELBUNDLEDIR):
 	mkdir -p $(WHEELBUNDLEDIR)
 
 bdist_wheel: $(WHEELDISTDIR)
-	for dir in $(IPACLIENT_SUBDIRS); do \
+	rm -f $(foreach item,$(IPA_WHEEL_PACKAGES) ipatests,$(WHEELDISTDIR)/$(item)-*.whl)
+	export IPA_OMIT_INSTALL=$(IPA_OMIT_INSTALL); \
+	for dir in $(IPA_WHEEL_PACKAGES) ipatests; do \
 	    $(MAKE) $(AM_MAKEFLAGS) -C $${dir} $@ || exit 1; \
 	done
 
 wheel_bundle: $(WHEELBUNDLEDIR) bdist_wheel .wheelconstraints
-	rm -f $(foreach item,$(IPACLIENT_SUBDIRS),$(WHEELBUNDLEDIR)/$(item)-*.whl)
-	$(PYTHON) -m pip wheel \
+	rm -f $(foreach item,$(IPA_WHEEL_PACKAGES) ipatests,$(WHEELBUNDLEDIR)/$(item)-*.whl)
+	@# dbus-python sometimes fails when MAKEFLAGS is set to -j2 or higher
+	MAKEFLAGS= $(PYTHON) -m pip wheel \
 	    --disable-pip-version-check \
 	    --constraint .wheelconstraints \
 	    --find-links $(WHEELDISTDIR) \
 	    --find-links $(WHEELBUNDLEDIR) \
 	    --wheel-dir $(WHEELBUNDLEDIR) \
-	    $(IPACLIENT_SUBDIRS)
+	    $(IPA_WHEEL_PACKAGES) $(IPA_EXTRA_WHEELS)
 
 wheel_placeholder: $(WHEELDISTDIR)
 	for dir in $(IPA_PLACEHOLDERS); do \
--- a/ipaserver/plugins/caacl.py
+++ b/ipaserver/plugins/caacl.py
@@ -2,12 +2,10 @@
 # Copyright (C) 2015  FreeIPA Contributors see COPYING for license
 #
 
-import pyhbac
 import six
 
 from ipalib import api, errors, output
 from ipalib import Bool, Str, StrEnum
-from ipalib.constants import IPA_CA_CN
 from ipalib.plugable import Registry
 from .baseldap import (
     LDAPObject, LDAPSearch, LDAPCreate, LDAPDelete, LDAPQuery,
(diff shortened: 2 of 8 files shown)


ipaserver/plugins/hbactest.py
```diff
@@ -210,7 +215,7 @@
 
 register = Registry()
 
-def convert_to_ipa_rule(rule):
+def _convert_to_ipa_rule(rule):
```
MartinBasti: This change is not related to pyhbac, should be in separate patch

tiran: I made the function a private function because it uses pyhbac internally. Since it's only used internally, I figured out that it makes more sense to make it private than to add a check for ```pyhbac is not None``` here, too.

tiran: I included a comment in the commit message.

pvoborni:
@tiran  I have very vague idea how this is helpful. You have mentioned it during post-devconf "API meeting".  But I no longer remember it and description of this PR is very general.  

In order to move all the pypi patches forward, we need to document(maybe design) the whole pypi workflow.  This is not mentioned in http://www.freeipa.org/page/V4/Build_system_refactoring nor in  http://www.freeipa.org/page/V4/Integration_Improvements  I.e. how FreeIPA project will work/supply packages to PYPI and what are actually the requirements for these packages.  What is expected to work and what not (like everything related to pyhbac).

Right now I have no idea what are the missing blocker parts and what are just nice-to-have things. 

Also I don't really like the part that the patches use custom repo of python-nss. But I'm glad that you are working with @jdennis to improve it. @stlaz, with PR #367 what are the remaining usages of python-nss? Could we actually get rid of python-nss completely?

tiran:
@pvoborni The main reason for this PR is explained in the initial PR message. I like to run an IPA framework server with specially instrumented Python builds for profiling or for debugging. The special builds are powerful and incredible useful tools to find bugs or hot spots.

Profile and debug builds have a different ABI than standard builds. Therefore I have to compile all C extensions myself to make them compatible with the new ABI. It is much easier than it sounds, because distutils, setuptools and pip just take care of all the complicated bits and pieces. But this works only for native Python packages. SSSD uses its own build system and has no packages on PyPI. It would take too much time and effort to change SSSD now. Commits 1f195bb4 and c69c30c2 make pyhbac and other SSSD components optional.

Commit 905118a1 allows me to build all ipaserver wheel and full ipaclient wheels with ```install``` subpackage for local testing. These packages are not meant to be uploaded to PyPI. They are really just for local testing.

Last but not least 5710587f is a workaround for a python-nss packaging issue. @jdennis is aware of the problem and will address it in due time. We can't get rid of python-nss. Dogtag PKI's Python modules depend on python-nss.

tiran:
To clarify and emphasis, this PR has nothing to do with the PyPI packaging effort. Zero. Zip. Nada. Nilch!

 The sole intent of this PR is debugging and profiling. It gives us tools to find bugs, to increase performance and to reduce memory usage.

HonzaCholasta:
I can't say I agree with this approach. If this is just for testing, surely you can work around the missing `pyhbac` in some isolated spot rather than make the import optional all over the place, when it is in fact required. Maybe inject an empty module into `sys.modules` if it's missing? Or reach out to the SSSD guys and help them add `pyhbac` to PyPI?

ipaserver/setup.py
```diff
@@ -76,4 +70,12 @@
                 'IPASecStore = ipaserver.secrets.store:IPASecStore',
             ],
         },
+        extras_require={
+            # These packages are currently not available on PyPI.
+            "caacl": ["pyhbac"],
```
HonzaCholasta: The missing `pyhbac` does not actually affect `caacl`, but `cert` - `cert-request` is the only place where the `acl_evaluate` function is called.

tiran: fixed, thx

tiran:
@HonzaCholasta FreeIPA has conditional imports for SSSD modules in several places, e.g. in the trust plugin. 96f614e closes the gap and applies the same technique to the last unconditional import from SSSD.

HonzaCholasta:
The trust plugin and other trust bits are optional. The cert plugin, which depends on `pyhbac`, is *not* optional, so you can't apply the same logic to it.

An acceptable compromise would be to skip the cert plugin entirely if `pyhbac` is not available:
```python
try:
    import pyhbac
except ImportError:
    raise errors.SkipPluginModule(reason=_('pyhbac is not installed'))
```

tiran:
I didn't know about the ```SkipPluginModule``` feature. I agree with you, your solution is more elegant.

tiran:
@jdennis released python-nss 1.0.1. I removed my workaround.

MartinBasti:
needs rebase

ipaserver/plugins/caacl.py
```diff
@@ -17,6 +16,11 @@
 from ipalib import _, ngettext
 from ipapython.dn import DN
 
+try:
+    import pyhbac
+except ImportError:
+    pyhbac = None
```
HonzaCholasta: `SkipPluginModule` should be raised here as well. Alternatively, you could move `acl_evaluate` and this import to cert.py, since that the only place where it's used, and raise `SkipPluginModule` there.

tiran:
I've moved the code to cert.py and raise SkipPluginModule from there.

tiran:
PR is blocked by #613

MartinBasti:
Please rebase, it is ok to me, I see potential for future server unit testing. I will test when rebased. If somebody is against this please say it now.

HonzaCholasta:
LGTM.

tiran:
Thanks @MartinBasti 

I rebased the PR and added a small workaround for ```dbus-python```. The package uses make to compile some of its internal dependencies. It looks like there is a bug in ```dbus-python```'s makefile. It sometimes fails to compile with my ```MAKEFLAGS=-j4``` env var. ```Makefile.am``` line 253 sets MAKEFLAGS to empty value for ```pip wheel```.

Makefile.am
```diff
@@ -227,26 +227,40 @@ endif  # WITH_JSLINT
...
+@MK_IFEQ@ ($(IPA_SERVER_WHEELS),1)
+    IPA_WHEEL_PACKAGES @MK_ASSIGN@ $(IPACLIENT_SUBDIRS) ipaplatform ipaserver
+    IPA_OMIT_INSTALL @MK_ASSIGN@ 0
+@MK_ELSE@
+    IPA_WHEEL_PACKAGES @MK_ASSIGN@ $(IPACLIENT_SUBDIRS)
+    IPA_OMIT_INSTALL @MK_ASSIGN@ 1
+@MK_ENDIF@
+
+# additional wheels for bundle, e.g. IPA_EXTRA_WHEELS="ipatests[webui] pylint"
+IPA_EXTRA_WHEELS=
```
tiran: I added these two lines to document the intent of IPA_EXTRA_WHEELS.

Makefile.am
```diff
@@ -227,26 +227,40 @@ endif  # WITH_JSLINT
...
+	for dir in $(IPA_WHEEL_PACKAGES) ipatests; do \
 	    $(MAKE) $(AM_MAKEFLAGS) -C $${dir} $@ || exit 1; \
 	done
 
 wheel_bundle: $(WHEELBUNDLEDIR) bdist_wheel .wheelconstraints
-	rm -f $(foreach item,$(IPACLIENT_SUBDIRS),$(WHEELBUNDLEDIR)/$(item)-*.whl)
-	$(PYTHON) -m pip wheel \
+	rm -f $(foreach item,$(IPA_WHEEL_PACKAGES) ipatests,$(WHEELBUNDLEDIR)/$(item)-*.whl)
+	@# dbus-python sometimes fails when MAKEFLAGS is set to -j2 or higher
+	MAKEFLAGS= $(PYTHON) -m pip wheel \
```
tiran: That's the workaround for ```dbus-python```.

MartinBasti:
Build failed:

```
make wheel_bundle IPA_SERVER_WHEELS=1
...
  checking for DBUS... no
  configure: error: Package requirements (dbus-1 >= 1.6) were not met:
  
  No package 'dbus-1' found
  
  Consider adjusting the PKG_CONFIG_PATH environment variable if you
  installed software in a non-standard prefix.
  
  Alternatively, you may set the environment variables DBUS_CFLAGS
  and DBUS_LIBS to avoid the need to call pkg-config.
  See the pkg-config man page for more details.
    File "<string>", line 1, in <module>
    File "/tmp/pip-build-l97uxR/dbus-python/setup.py", line 106, in <module>
      'build_ext': BuildExt,
    File "/usr/lib64/python2.7/distutils/core.py", line 151, in setup
      dist.run_commands()
    File "/usr/lib64/python2.7/distutils/dist.py", line 953, in run_commands
      self.run_command(cmd)
    File "/usr/lib64/python2.7/distutils/dist.py", line 972, in run_command
      cmd_obj.run()
    File "/usr/lib/python2.7/site-packages/wheel/bdist_wheel.py", line 199, in run
      self.run_command('build')
    File "/usr/lib64/python2.7/distutils/cmd.py", line 326, in run_command
      self.distribution.run_command(command)
    File "/usr/lib64/python2.7/distutils/dist.py", line 972, in run_command
      cmd_obj.run()
    File "/tmp/pip-build-l97uxR/dbus-python/setup.py", line 62, in run
      cwd=builddir)
    File "/usr/lib64/python2.7/subprocess.py", line 186, in check_call
      raise CalledProcessError(retcode, cmd)
  subprocess.CalledProcessError: Command '['/tmp/pip-build-l97uxR/dbus-python/configure', '--disable-maintainer-mode', 'PYTHON=/usr/bin/python', '--prefix=/tmp/pip-build-l97uxR/dbus-python/build/temp.linux-x86_64-2.7/prefix']' returned non-zero exit status 1
  
  ----------------------------------------
  Failed building wheel for dbus-python
  Running setup.py clean for dbus-python
  Running setup.py bdist_wheel for MarkupSafe ... done
  Stored in directory: /tmp/freeipa/dist/bundle
  Running setup.py bdist_wheel for pycparser ... done
  Stored in directory: /tmp/freeipa/dist/bundle
  Running setup.py bdist_wheel for configparser ... done
  Stored in directory: /tmp/freeipa/dist/bundle
Successfully built cryptography python-yubico pyusb python-nss pyldap netifaces gssapi MarkupSafe pycparser configparser
Failed to build dbus-python
ERROR: Failed to build one or more wheels
Makefile:1222: recipe for target 'wheel_bundle' failed

```

tiran:
You need dbus-devel package.

I opened https://pagure.io/freeipa/issue/6842 to track lack of documentation.

MartinBasti:
So put it into specfile to `with_wheels` section

MartinBasti:
And document in `BUILD.txt` how to build wheels

tiran:
@MartinBasti ```dbus-devel``` is in the ```with_wheels``` section. Documentation is part of https://pagure.io/freeipa/issue/6842 .

MartinBasti:
@tiran sorry, but then something doesn't work as expected

```
$ dnf builddep -b -D "with_wheels 1" --spec freeipa.spec.in
<everything already installed>

$ make wheel_bundle IPA_SERVER_WHEELS=1
...
  Failed building wheel for dbus-python

$ git grep -in  dbus-devel freeipa.spec.in 
freeipa.spec.in:149:BuildRequires:  dbus-devel

$ rpm -q dbus-devel
package dbus-devel is not installed
```

Probably because `%global with_wheels 0` is defined in the spec file, so it always overrides my CLI settings

MartinBasti:
Missing dependency `dbus-glib-devel` otherwise it works

tiran:
Sigh, I got it.

MartinBasti:
Can you rebase please? I see Ben's commits in this PR

MartinBasti:
master:

* 3064b890e24a5056c77f19c3951cfb59d49366f8 Conditionally import pyhbac
* 7c9df35d3dd812543cf537c0762c6db728379919 Add extra_requires for additional dependencies
* ae1c2086db3efb51341025ae25f9b39060868ac1 Add an option to build ipaserver wheels
* 40a60675f3feb118af70562582399afefe97214d Don't hard-code with_wheels

PR #398, opened 2017-01-18
Support for Certificate Identity Mapping

flo-renaud:
See design http://www.freeipa.org/page/V4/Certificate_Identity_Mapping

https://fedorahosted.org/freeipa/ticket/6542

### DIFF
--- a/install/share/73certmap.ldif
+++ b/install/share/73certmap.ldif
@@ -1,0 +1,14 @@
+## IPA Base OID:
+##
+## Attributes:          2.16.840.1.113730.3.8.22.1.x
+## ObjectClasses:       2.16.840.1.113730.3.8.22.2.y
+##
+dn: cn=schema
+attributeTypes: (2.16.840.1.113730.3.8.22.1.1 NAME 'ipaCertMapPromptUsername' DESC 'Prompt for the username when multiple identities are mapped to a certificate' EQUALITY booleanMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.7 SINGLE-VALUE X-ORIGIN 'IPA v4.5' )
+attributeTypes: (2.16.840.1.113730.3.8.22.1.2 NAME 'ipaCertMapMapRule' DESC 'Certificate Mapping Rule' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 SINGLE-VALUE X-ORIGIN 'IPA v4.5' )
+attributeTypes: (2.16.840.1.113730.3.8.22.1.3 NAME 'ipaCertMapMatchRule' DESC 'Certificate Matching Rule' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 SINGLE-VALUE X-ORIGIN 'IPA v4.5' )
+attributeTypes: (2.16.840.1.113730.3.8.22.1.4 NAME 'ipaCertMapData' DESC 'Certificate Mapping Data' EQUALITY caseIgnoreMatch SUBSTR caseIgnoreSubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 X-ORIGIN 'IPA v4.5' )
+attributeTypes: (2.16.840.1.113730.3.8.22.1.5 NAME 'ipaCertMapPriority' DESC 'Rule priority' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 SINGLE-VALUE X-ORIGIN 'IPA v4.5' )
+objectClasses: (2.16.840.1.113730.3.8.22.2.1 NAME 'ipaCertMapConfigObject' DESC 'IPA Certificate Mapping global config options' AUXILIARY MAY ipaCertMapPromptUsername X-ORIGIN 'IPA v4.5' )
+objectClasses: (2.16.840.1.113730.3.8.22.2.2 NAME 'ipaCertMapRule' DESC 'IPA Certificate Mapping rule' SUP top STRUCTURAL MUST cn MAY ( description $ ipaCertMapMapRule $ ipaCertMapMatchRule $ associatedDomain $ ipaCertMapPriority $ ipaEnabledFlag ) X-ORIGIN 'IPA v4.5' )
+objectClasses: (2.16.840.1.113730.3.8.22.2.3 NAME 'ipaCertMapObject' DESC 'IPA Object for Certificate Mapping' AUXILIARY MAY ipaCertMapData X-ORIGIN 'IPA v4.5' )
--- a/install/updates/73-certmap.update
+++ b/install/updates/73-certmap.update
@@ -1,0 +1,23 @@
+# Configuration for Certificate Identity Mapping
+dn: cn=certmap,$SUFFIX
+default:objectclass: top
+default:objectclass: nsContainer
+default:objectclass: ipaCertMapConfigObject
+default:cn: certmap
+default:ipaCertMapPromptUsername: FALSE
+
+dn: cn=certmaprules,cn=certmap,$SUFFIX
+default:objectclass: top
+default:objectclass: nsContainer
+default:cn: certmaprules
+
+# Certificate Identity Mapping Administrators
+dn: cn=Certificate Identity Mapping Administrators,cn=privileges,cn=pbac,$SUFFIX
+default:objectClass: top
+default:objectClass: groupofnames
+default:objectClass: nestedgroup
+default:cn: Certificate Identity Mapping Administrators
+default:description: Certificate Identity Mapping Administrators
+
+dn: $SUFFIX
+add:aci: (targetattr = "ipacertmapdata")(targattrfilters="add=objectclass:(objectclass=ipacertmapobject)")(version 3.0;acl "selfservice:Users can manage their own X.509 certificate identity mappings";allow (write) userdn = "ldap:///self";)
--- a/ipapython/dn.py
+++ b/ipapython/dn.py
@@ -1155,8 +1155,14 @@ class DN(object):
     def _get_rdn(self, rdn):
         return self.RDN_type(*rdn, **{'raw': True})
 
-    def __str__(self):
+    def ldap_text(self):
         return dn2str(self.rdns)
+
+    def x500_text(self):
+        return dn2str(reversed(self.rdns))
+
+    def __str__(self):
+        return self.ldap_text()
 
     def __repr__(self):
         return "%s.%s('%s')" % (self.__module__, self.__class__.__name__, self.__str__())
--- a/ipaserver/plugins/baseuser.py
+++ b/ipaserver/plugins/baseuser.py
@@ -19,14 +19,17 @@
 
 import six
 
-from ipalib import api, errors
-from ipalib import Flag, Int, Password, Str, Bool, StrEnum, DateTime, Bytes
+from ipalib import api, errors, x509
+from ipalib import (
+    Flag, Int, Password, Str, Bool, StrEnum, DateTime, Bytes, DNParam)
 from ipalib.parameters import Principal
 from ipalib.plugable import Registry
 from .baseldap import (
     DN, LDAPObject, LDAPCreate, LDAPUpdate, LDAPSearch, LDAPDelete,
-    LDAPRetrieve, LDAPAddAttribute, LDAPRemoveAttribute, LDAPAddMember,
-    LDAPRemoveMember, LDAPAddAttributeViaOption, LDAPRemoveAttributeViaOption)
+    LDAPRetrieve, LDAPAddAttribute, LDAPModAttribute, LDAPRemoveAttribute,
+    LDAPAddMember, LDAPRemoveMember,
+    LDAPAddAttributeViaOption, LDAPRemoveAttributeViaOption,
+    add_missing_object_class)
 from ipaserver.plugins.service import (
    validate_certificate, validate_realm, normalize_principal)
 from ipalib.request import context
@@ -134,7 +137,7 @@ class baseuser(LDAPObject):
     object_class_config = 'ipauserobjectclasses'
     possible_objectclasses = [
         'meporiginentry', 'ipauserauthtypeclass', 'ipauser',
-        'ipatokenradiusproxyuser'
+        'ipatokenradiusproxyuser', 'ipacertmapobject'
     ]
     disallow_object_classes = ['krbticketpolicyaux']
     permission_filter_objectclasses = ['posixaccount']
@@ -146,7 +149,8 @@ class baseuser(LDAPObject):
         'memberofindirect', 'ipauserauthtype', 'userclass',
         'ipatokenradiusconfiglink', 'ipatokenradiususername',
         'krbprincipalexpiration', 'usercertificate;binary',
-        'krbprincipalname', 'krbcanonicalname'
+        'krbprincipalname', 'krbcanonicalname',
+        'ipacertmapdata'
     ]
     search_display_attributes = [
         'uid', 'givenname', 'sn', 'homedirectory', 'krbcanonicalname',
@@ -359,6 +363,13 @@ class baseuser(LDAPObject):
             cli_name='certificate',
             label=_('Certificate'),
             doc=_('Base-64 encoded user certificate'),
+        ),
+        Str(
+            'ipacertmapdata*',
+            cli_name='certmapdata',
+            label=_('Certificate mapping data'),
+            doc=_('Certificate mapping data'),
+            flags=['no_create', 'no_update', 'no_search'],
         ),
     )
 
@@ -728,3 +739,154 @@ class baseuser_remove_cert(LDAPRemoveAttributeViaOption):
         self.obj.convert_usercertificate_post(entry_attrs, **options)
 
         return dn
+
+
+class ModCertMapData(LDAPModAttribute):
+    attribute = 'ipacertmapdata'
+    takes_options = (
+        DNParam(
+            'issuer?',
+            cli_name='issuer',
+            label=_('Issuer'),
+            doc=_('Issuer of the certificate'),
+            flags=['virtual_attribute']
+        ),
+        DNParam(
+            'subject?',
+            cli_name='subject',
+            label=_('Subject'),
+            doc=_('Subject of the certificate'),
+            flags=['virtual_attribute']
+        ),
+        Bytes(
+            'certificate*', validate_certificate,
+            cli_name='certificate',
+            label=_('Certificate'),
+            doc=_('Base-64 encoded user certificate'),
+            flags=['virtual_attribute']
+        ),
+    )
+
+    @staticmethod
+    def _build_mapdata(subject, issuer):
+        return u'X509:<I>{issuer}<S>{subject}'.format(
+            issuer=issuer.x500_text(), subject=subject.x500_text())
+
+    @classmethod
+    def _convert_options_to_certmap(cls, entry_attrs, issuer=None,
+                                    subject=None, certificates=()):
+        """
+        Converts options to ipacertmapdata
+
+        When --subject --issuer or --certificate options are used,
+        the value for ipacertmapdata is built from extracting subject and
+        issuer,
+        converting their values to X500 ordering and using the format
+        X509:<I>issuer<S>subject
+        For instance:
+        X509:<I>O=DOMAIN,CN=Certificate Authority<S>O=DOMAIN,CN=user
+        A list of values can be returned if --certificate is used multiple
+        times, or in conjunction with --subject --issuer.
+        """
+        data = []
+        data.extend(entry_attrs.get(cls.attribute, list()))
+
+        if issuer or subject:
+            data.append(cls._build_mapdata(subject, issuer))
+
+        for dercert in certificates:
+            cert = x509.load_certificate(dercert, x509.DER)
+            issuer = DN(cert.issuer)
+            subject = DN(cert.subject)
+            if not subject:
+                raise errors.ValidationError(
+                    name='certificate',
+                    error=_('cannot have an empty subject'))
+            data.append(cls._build_mapdata(subject, issuer))
+
+        entry_attrs[cls.attribute] = data
+
+    def get_args(self):
+        # ipacertmapdata is not mandatory as it can be built
+        # from the values subject+issuer or from reading certificate
+        for arg in super(ModCertMapData, self).get_args():
+            if arg.name == 'ipacertmapdata':
+                yield arg.clone(required=False, alwaysask=False)
+            else:
+                yield arg.clone()
+
+    def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys,
+                     **options):
+        # The 3 valid calls are
+        # ipa user-add-certmapdata LOGIN --subject xx --issuer yy
+        # ipa user-add-certmapdata LOGIN [DATA] --certificate xx
+        # ipa user-add-certmapdata LOGIN DATA
+        # Check that at least one of the 3 formats is used
+
+        try:
+            certmapdatas = keys[1] or []
+        except IndexError:
+            certmapdatas = []
+        issuer = options.get('issuer')
+        subject = options.get('subject')
+        certificates = options.get('certificate', [])
+
+        # If only LOGIN is supplied, then we need either subject or issuer or
+        # certificate
+        if (not certmapdatas and not issuer and not subject and
+                not certificates):
+            raise errors.RequirementError(name='ipacertmapdata')
+
+        # If subject or issuer is provided, other options are not allowed
+        if subject or issuer:
+            if certificates:
+                raise errors.MutuallyExclusiveError(
+                    reason=_('cannot specify both subject/issuer '
+                             'and certificate'))
+            if certmapdatas:
+                raise errors.MutuallyExclusiveError(
+                    reason=_('cannot specify both subject/issuer '
+                             'and ipacertmapdata'))
+            # If subject or issuer is provided, then the other one is required
+            if not subject:
+                raise errors.RequirementError(name='subject')
+            if not issuer:
+                raise errors.RequirementError(name='issuer')
+
+        # if the command is called with --subject --issuer or --certificate
+        # we need to add ipacertmapdata to the attrs_list in order to
+        # display the resulting value in the command output
+        if 'ipacertmapdata' not in attrs_list:
+            attrs_list.append('ipacertmapdata')
+
+        self._convert_options_to_certmap(
+            entry_attrs,
+            issuer=issuer,
+            subject=subject,
+            certificates=certificates)
+
+        return dn
+
+
+class baseuser_add_certmapdata(ModCertMapData, LDAPAddAttribute):
+    __doc__ = _("Add one or more certificate mappings to the user entry.")
+    msg_summary = _('Added certificate mappings to user "%(value)s"')
+
+    def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys,
+                     **options):
+
+        dn = super(baseuser_add_certmapdata, self).pre_callback(
+            ldap, dn, entry_attrs, attrs_list, *keys, **options)
+
+        # The objectclass ipacertmapobject may not be present on
+        # existing user entries. We need to add it if we define a new
+        # value for ipacertmapdata
+        add_missing_object_class(ldap, u'ipacertmapobject', dn)
+
+        return dn
+
+
+class baseuser_remove_certmapdata(ModCertMapData,
+                                  LDAPRemoveAttribute):
+    __doc__ = _("Remove one or more certificate mappings from the user entry.")
+    msg_summary = _('Removed certificate mappings from user "%(value)s"')
--- a/ipaserver/plugins/certmap.py
+++ b/ipaserver/plugins/certmap.py
@@ -1,0 +1,391 @@
+# Authors:
+#   Florence Blanc-Renaud <flo@redhat.com>
+#
+# Copyright (C) 2017  Red Hat
+# see file 'COPYING' for use and warranty information
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import six
+
+from ipalib import api, errors
+from ipalib.parameters import Bool, DNSNameParam, Flag, Int, Str
+from ipalib.plugable import Registry
+from .baseldap import (
+    LDAPCreate,
+    LDAPDelete,
+    LDAPObject,
+    LDAPQuery,
+    LDAPRetrieve,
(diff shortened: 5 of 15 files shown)


ipaserver/plugins/baseuser.py
```diff
@@ -18,15 +18,17 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import six
+from ldap.dn import dn2str, str2dn
```
MartinBasti: Respectively implement reverse() to `ipapython.dn.DN` if it is not already there

ipaserver/plugins/baseuser.py
```diff
@@ -694,3 +703,161 @@ class baseuser_remove_principal(LDAPRemoveAttribute):
...
+    the value for ipacertmapdata is built from extracting subject and issuer,
+    converting their values to X500 ordering and using the format
+    X509:<I>issuer<S>subject
+    For instance:
+    X509:<I>O=DOMAIN,CN=Certificate Authority<S>O=DOMAIN,CN=user
+    A list of values can be returned if --usercertificate is used multiple
+    times, or in conjunction with --subject --issuer.
+    """
+    data = []
+    if 'ipacertmapdata' in options:
```
MartinBasti: This can be shorter:

```
      for item in options.get('ipacertmapdata', ()):
            data.append(item)
```

ipaserver/plugins/baseuser.py
```diff
@@ -694,3 +703,161 @@ class baseuser_remove_principal(LDAPRemoveAttribute):
...
+        for item in options['ipacertmapdata']:
+            data.append(item)
+
+    if 'issuer' in options or 'subject' in options:
+        issuer = options.get('issuer')
+        subject = options.get('subject')
+        data.append(_build_mapdata(subject, issuer))
+
+    if 'usercertificate' in options:
```
MartinBasti: you can use get(key, default) too here to save one level of indentation

ipaserver/plugins/baseuser.py
```diff
@@ -694,3 +703,161 @@ class baseuser_remove_principal(LDAPRemoveAttribute):
...
+
+
+    def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys,
+                     **options):
+        # The 3 valid calls are
+        # --subject xx --issuer yy
+        # --certificate
+        # --data
+        # Check that at least one of the 3 formats is used
+        if 'issuer' not in options and \
```
MartinBasti: this is  ugly :)

more pythonic way is to use `all()`

```
# from top of my head untested
if all(key not in options for key in ('issuer', subject', ....)):
```

ipaserver/plugins/baseuser.py
```diff
@@ -694,3 +703,161 @@ class baseuser_remove_principal(LDAPRemoveAttribute):
...
+        for option in super(baseuser_remove_certmap, self).get_options():
+            if option.name in ['ipacertmapdata']:
+                yield option.clone(required=False, alwaysask=False)
+            else:
+                yield option.clone()
+
+
+    def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys,
+                     **options):
+        if 'issuer' not in options and \
```
MartinBasti: `all()` can be use here too

ipaserver/plugins/certmap.py
```diff
@@ -0,0 +1,345 @@
...
+
+    permission_filter_objectclasses = ['ipacertmaprule']
+    managed_permissions = {
+        'System: Add Certmap Rules': {
+            'replaces_global_anonymous_aci': True,
+            'ipapermright': {'add'},
+            'default_privileges': {
+                'Certificate Identity Mapping Administrators'},
+        },
+        'System: Read Certmap Rules': {
```
MartinBasti: I have a question related to ACI, if `System: Read Certmap Configuration` is allowed for `ldap://all`, shouldn't be this `System: Read Certmap Rules` accessible to `ldap://all`? What/who is the consumer of this rules? SSSD?

ipaserver/plugins/user.py
```diff
@@ -366,6 +368,13 @@ class user(baseuser):
             },
             'default_privileges': {'PassSync Service'},
         },
+        'System: Manage User Certificate Mappings': {
```
MartinBasti: IMHO the similar permission should be added to stageuser implementation, something like `System: Manage Stage User Certificate Mappings`

MartinBasti:
I put some inline commets, @flo-renaud if you don't know where to register OIDs feel free to ping me

ipaserver/plugins/user.py
flo-renaud: The permissions for Users are finer-grained than the permissions for stage users. For instance, for users, there is "System: Modify Users" (which does not grant write on usercertificate) and "System: Manage User Certificates".
For stage user, there is only "System: Modify stage User" which applies to all the attributes.

I applied the same strategy (finer-grained for users) but if it is a mistake I can create a similar permission for cert mapping on stage users.

ipaserver/plugins/baseuser.py
flo-renaud: Agree

ipaserver/plugins/baseuser.py
flo-renaud: Agree

ipaserver/plugins/baseuser.py
flo-renaud: Agree

ipaserver/plugins/baseuser.py
flo-renaud: Agree

ipaserver/plugins/baseuser.py
flo-renaud: I will implement reverse() in ipapython.dn.DN and add unit tests.

install/share/73certmap.ldif
```diff
@@ -0,0 +1,17 @@
+## IPA Base OID:
+##
+## Attributes:          2.16.840.1.113730.3.8.22.x
```
flo-renaud: Sure, my plan was to wait for final agreement on the PR before registering the OIDs.

ipaserver/plugins/user.py
MartinBasti: Ok

ipaserver/plugins/baseuser.py
```diff
@@ -694,3 +703,159 @@ class baseuser_remove_principal(LDAPRemoveAttribute):
...
+    ),
+)
+
+
+def _convert_to_x500(name):
+    """
+    Converts a (ipa) DN into a string representation following X500 order
+    """
+    if name:
+        name.reverse()
```
MartinBasti: I'm always scared by in-place methods. Do you count with that the object where the original name reference was pointing to, changed also outside the `_convert_to_x500` method?

maybe we should rather implement get_reverse() method that returns a new DN and keeps the original reference untouched

ipapython/dn.py
```diff
@@ -1419,6 +1419,15 @@ def rindex(self, pattern, start=None, end=None):
...
         return i
 
+    def get_reverse(self):
+        '''
+        Return a reverse copy of the DN. For instance, the reverse of
+        uid=user,dc=example,dc=com is dc=com,dc=example,uid=user
+        '''
+        reverse_dn = DN(self)
+        reverse_dn.rdns.reverse()
+        return reverse_dn
```
HonzaCholasta: There is no such thing as reversed DN - it's only the the textural representation which might be in both directions, but the underlying DN is the same. I would prefer if you instead added two new methods `ldap_text()` and `x500_text()` which return textual representation of the DN in the LDAP and X.500 order, respectively, and use them as appropriate:
```python
    def ldap_text(self):
        return dn2str([[self.to_openldap()]])

    def x500_text(self):
        return dn2str([[reversed(self.to_openldap()]])

    def __str__(self):
        return self.ldap_text()
```

ipaserver/plugins/baseuser.py
```diff
@@ -360,6 +362,13 @@ class baseuser(LDAPObject):
             label=_('Certificate'),
             doc=_('Base-64 encoded user certificate'),
         ),
+        Str(
+            'ipacertmapdata*',
+            cli_name='data',
```
HonzaCholasta: The CLI name should be `certmapdata`, `data` is far too generic and disconnected from certificate mapping.

ipaserver/plugins/baseuser.py
```diff
@@ -694,3 +703,158 @@ class baseuser_remove_principal(LDAPRemoveAttribute):
...
+        subject = options.get('subject')
+        data.append(_build_mapdata(subject, issuer))
+
+    for dercert in options.get('usercertificate', ()):
+        cert = x509.load_certificate(dercert, x509.DER)
+        issuer = DN(cert.issuer)
+        subject = DN(cert.subject)
+        data.append(_build_mapdata(subject, issuer))
+
+    return data
```
HonzaCholasta: Instead of a bunch of global definitions used in both `baseuser_add_certmap` and `baseuser_remove_certmap`, create a base class (e.g. `ModCertMapData`) which `baseuser_add_certmap` and `baseuser_remove_certmap` inherit from and encapsulate the common stuff inside it.

ipaserver/plugins/baseuser.py
```diff
@@ -694,3 +703,158 @@ class baseuser_remove_principal(LDAPRemoveAttribute):
...
+
+
+class baseuser_add_certmap(LDAPAddAttributeViaOption):
```
HonzaCholasta: This should be called `baseuser_add_certmapdata`, because the attribute is called `certmapdata` (actually it is called `data` now, but see my comment above). Ditto for `baseuser_remove_certmap`.

ipaserver/plugins/baseuser.py
```diff
@@ -694,3 +703,158 @@ class baseuser_remove_principal(LDAPRemoveAttribute):
...
+    msg_summary = _('Added certificate mappings to user "%(value)s"')
+
+    attribute = 'ipacertmapdata'
+    takes_options = certmap_options
+
+    def get_options(self):
+        # ipacertmapdata is not mandatory as it can be built
+        # from the values subject+issuer or from reading usercertificate
+        for option in super(baseuser_add_certmap, self).get_options():
+            if option.name in ['ipacertmapdata']:
```
HonzaCholasta: I think `option.name == 'ipacertmapdata'` would be preferrable.

ipaserver/plugins/baseuser.py
```diff
@@ -694,3 +703,158 @@ class baseuser_remove_principal(LDAPRemoveAttribute):
...
+            raise errors.RequirementError(name='data')
+
+        # The objectclass ipacertmapobject may not be present on
+        # existing user entries. We need to add it if we define a new
+        # value for ipacertmapdata
+        if 'objectclass' not in entry_attrs:
+            entry_attrs_old = ldap.get_entry(dn, ['objectclass'])
+        objclasses_lc = [x.lower() for x in entry_attrs_old['objectclass']]
+        if 'ipacertmapobject' not in objclasses_lc:
+            entry_attrs['objectclass'] = ['ipacertmapobject']
```
HonzaCholasta: Overwriting the `objectclass` is wrong and won't work, you should append `ipacertmapobject` to the existing value. Or, you could use `baseldap.add_missing_object_class()` which does all of this for you.

ipaserver/plugins/baseuser.py
```diff
@@ -694,3 +703,158 @@ class baseuser_remove_principal(LDAPRemoveAttribute):
...
+
+        entry_attrs[self.attribute] = _convert_options_to_certmap(options)
```
HonzaCholasta: IMO it would be better to implement this using `default_from` on the `ipacertmapdata`, `issuer` and `subject` parameters.

ipaserver/plugins/certmap.py
```diff
@@ -0,0 +1,357 @@
...
+        ),
+        Str(
+            'ipacertmapmatchrule?',
+            cli_name='matchrule',
+            label=_('Matching rule'),
+            doc=_('Rule used to check if a certificate can be used for'
+                  ' authentication'),
+        ),
+        Str(
+            'associateddomain*',
```
HonzaCholasta: I *think* this should be a `DNSNameParam`.

ipaserver/plugins/baseuser.py
flo-renaud: I initially considered using default_from, but it would be able to build ipacertmapdata only if both issuer and subject parameters are present (when one key is missing, it does not get called). The ipacertmapdata value can also be extracted from the certificate parameter, and I did not find any method in the framework allowing to take input either from (subject+issuer) or certificate. Would you have any suggestion?

flo-renaud:
Hi @HonzaCholasta,
PR updated with most of your comments, except the suggestion to use default_from. Please see my answer inline for this one.

ipaserver/plugins/baseuser.py
HonzaCholasta: By "common stuff" I meant all of `certmap_options`, `_build_mapdata` and `_convert_options_to_certmap`.

HonzaCholasta:
@flo-renaud, nevermind the `default_from` suggestion, I was wrong - if e.g. both `--certmapdata` and `--certificate` are specified, we want to use both, not throw away `--certificate`, which is exactly what would happen if `--certmapdata` had default derived from `--certificate`.

One more issue, I think the `--certmapdata` option in `user-add-certmapdata` and friends should actually be a positional argument, as that would be more consistent with existing commands. The common pattern is that positional arguments are used to specify the literal value of the attribute (such as principal name in `user-add-principal`), but options need some preprocessing (such as conversion from UID to DN in `group-add-member`). Currently the only exception to this scheme is `user-add-cert` and friends, but that's only because the original intent was to add a certificate file positional argument, but it never happened.

flo-renaud:
Hi @HonzaCholasta 
PR updated with `ipa user-add-certmapdata` using positional arg for CERTMAPDATA

install/updates/73-certmap.update
```diff
@@ -0,0 +1,27 @@
...
+default:objectclass: nsContainer
+default:cn: certmaprules
+
+# Certificate Identity Mapping Administrators
+dn: cn=Certificate Identity Mapping Administrators,cn=privileges,cn=pbac,$SUFFIX
+default:objectClass: top
+default:objectClass: groupofnames
+default:objectClass: nestedgroup
+default:cn: Certificate Identity Mapping Administrators
+default:description: Certificate Identity Mapping Administrators
```
HonzaCholasta: Note that we usually put new privileges into `40-delegation.update`. I'm not saying it's wrong to have it here, just curious why have you done it this way.

ipaserver/plugins/baseuser.py
```diff
@@ -728,3 +739,138 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
...
+
+class ModCertMapData(LDAPModAttribute):
+    attribute = 'ipacertmapdata'
+    takes_options = (
+        DNParam(
+            'issuer?',
+            cli_name='issuer',
+            label=_('Issuer'),
+            doc=_('Issuer of the certificate'),
+            flags=['virtual_attribute', 'no_create', 'no_update']
```
HonzaCholasta: Note that the `no_create` and `no_update` flags are meaningless outside of object plugin params.

ipaserver/plugins/baseuser.py
```diff
@@ -728,3 +739,138 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
...
+            doc=_('Base-64 encoded user certificate'),
+        ),
+    )
+
+    @staticmethod
+    def _build_mapdata(subject, issuer):
+        return u'X509:{issuer}{subject}'.format(
+            issuer='<I>{}'.format(issuer.x500_text()) if issuer else '',
+            subject='<S>{}'.format(subject.x500_text()) if subject else '',
+            )
```
HonzaCholasta: Why not just:
```python
    return u'X509:<I>{issuer}<S>{subject}'.format(issuer=issuer, subject=subject)
```
?

install/updates/73-certmap.update
flo-renaud: Hi @HonzaCholasta 
I simply had a look at install/updates/60-trusts.update and did the same, i.e. define new containers and privileges in the same file.
I will keep your note for future reference though.

ipaserver/plugins/baseuser.py
flo-renaud: Hi,
issuer or subject can be None. If for instance issuer=None we should return X509:<S>cn=subject but not X509:<I>None<S>cn=subject

ipaserver/plugins/baseuser.py
```diff
@@ -728,3 +739,138 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
...
+        # ipa user-add-certmapdata LOGIN --subject xx --issuer yy
+        # ipa user-add-certmapdata LOGIN [DATA] --certificate xx
+        # ipa user-add-certmapdata LOGIN DATA
+        # Check that at least one of the 3 formats is used
+
+        # If only LOGIN is supplied, then we need either subject or issuer or
+        # certificate
+        if len(keys) < 2:
+            valid_options = ('issuer', 'subject', 'usercertificate')
+            if all(key not in options for key in valid_options):
```
HonzaCholasta: I'm pretty sure just checking for presence of the options is not good enough, you should also check that none of the options are `None`. IMO the easiest way to achieve this would be to add this at the beginning of `pre_callback`:
```python
        try:
            certmapdatas = args[1] or []
        except IndexError:
            certmapdatas = []
        issuer = options.get('issuer')
        subject = options.get('subject')
        certificates = options.get('usercertificate') or []
```
and work with these variables further in the code.

ipaserver/plugins/baseuser.py
```diff
@@ -728,3 +739,138 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
...
+        A list of values can be returned if --usercertificate is used multiple
+        times, or in conjunction with --subject --issuer.
+        """
+        data = []
+        data.extend(certmapdata)
+
+        if 'issuer' in options or 'subject' in options:
+            issuer = options.get('issuer')
+            subject = options.get('subject')
+            data.append(ModCertMapData._build_mapdata(subject, issuer))
```
HonzaCholasta: IMO it would be better to declare `_convert_options_to_certmap()` as class method, so you can `cls._build_mapdata()` instead of `ModCertMapData._build_mapdata()`.

ipaserver/plugins/baseuser.py
```diff
@@ -728,3 +739,138 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
...
+
+        # if the command is called with --subject --issuer or --certificate
+        # we need to add ipacertmapdata to the attrs_list in order to
+        # display the resulting value in the command output
+        if 'ipacertmapdata' not in attrs_list:
+            attrs_list.append('ipacertmapdata')
+
+        data = entry_attrs.get(self.attribute, list())
+        entry_attrs[self.attribute] = self._convert_options_to_certmap(
+            data, options)
```
HonzaCholasta: IMO it would be better to hide all this inside `_convert_options_to_certmap()` and have it called like this here:
```python
        self._convert_options_to_certmap(entry_attrs, options)
```

ipaserver/plugins/baseuser.py
```diff
@@ -728,3 +739,138 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
...
+        DNParam(
+            'subject?',
+            cli_name='subject',
+            label=_('Subject'),
+            doc=_('Subject of the certificate'),
+            flags=['virtual_attribute', 'no_create', 'no_update']
+        ),
+        Bytes(
+            'usercertificate*', validate_certificate,
+            cli_name='certificate',
```
HonzaCholasta: Actually, can we name the param `certificate` instead of `usercertificate`? It does not refer to the LDAP attribute after all. Speaking of that, it should have the `virtual_attribute` flag.

ipaserver/plugins/baseuser.py
HonzaCholasta: When they can be `None`? I don't see when this can be useful.

ipaserver/plugins/certmap.py
```diff
@@ -0,0 +1,355 @@
...
+
+    msg_summary = _('Added Certificate Identity Mapping Rule "%(value)s"')
+
+
+@register()
+class certmaprule_mod(LDAPUpdate):
+    __doc__ = _('Modify a Certificate Identity Mapping Rule.')
+
+    msg_summary = _('Modified Certificate Identity Mapping Rule "%(value)s"')
+
```
HonzaCholasta: I think we should validate that `associateddomain` is either the IPA domain or one of the trusted domains in `certmaprule_add` and `certmaprule_mod`.

ipaserver/plugins/baseuser.py
flo-renaud: Hi @HonzaCholasta 
for instance if the user calls `ipa user-add-certmapdata LOGIN --subject cn=subject` then issuer=None.

ipaserver/plugins/certmap.py
flo-renaud: @HonzaCholasta 
I discussed this point offline with Sumit and his answer was that any domain name should be accepted. If SSSD cannot make sense of one of the domains it will ignore it and log a message

flo-renaud:
@HonzaCholasta 
PR updated according to your comments. Thanks for the detailed review!

ipaserver/plugins/certmap.py
HonzaCholasta: I can't say I agree, IMHO it does not make sense to allow rules for domains not related to IPA in any way. Also, with validation, we would be able to catch typos.

ipaserver/plugins/baseuser.py
HonzaCholasta: I get that, but what is the use case? The subject name is meaningless without context, and the issuer name provides that context. The only use case I can think of is that there is only a single CA (forever), so the issuer name is implied, but I believe that's a corner case which does not need to have special support outside of the generic `ipa user-add-certmapdata LOGIN VALUE`.

flo-renaud: @HonzaCholasta 
I'm afraid I disagree here. AD supports both (see [Map a Certificate to a User Account](https://technet.microsoft.com/en-us/library/cc754866(v=ws.11).aspx)) and it is easy for us to provide a function that supports only issuer / only subject / issuer+subject, so why make it more difficult for the administrator? It is more user-friendly to use the issuer or subject options rather than specify the format X509:....

flo-renaud:
PR updated with the check on domain in certmaprule-add/mod.

ipaserver/plugins/baseuser.py
HonzaCholasta: Because it has additional costs for us (in maintenance, testing, ...) for very questionable benefit. Nobody requested it and I seriously doubt anybody would. What does it matter if it's user friendly when there's no user to use it that way?

AD does not support issuer-only AFAICT as it is non-sensical, but it supports [other variants](https://blogs.msdn.microsoft.com/spatdsg/2010/06/18/howto-map-a-user-to-a-certificate-via-all-the-methods-available-in-the-altsecurityidentities-attribute/) which this PR does not implement, so clearly supporting everything AD supports is not the goal here, and I don't think it ever was in the first place.

install/updates/73-certmap.update
```diff
@@ -0,0 +1,27 @@
+# Configuration for Certificate Identity Mapping
+dn: cn=certmap,cn=ipa,cn=etc,$SUFFIX
+default:objectclass: top
+default:objectclass: nsContainer
+default:objectclass: ipaConfigObject
+default:objectclass: ipaCertMapContainer
+default:objectclass: ipaCertMapConfigObject
+default:cn: certmap
+default:ipaconfigstring: CertMapVersion 1
+default:ipacertmapversion: 1
```
HonzaCholasta: What is the point of the version? Why is it specified both using `ipaConfigString` and `ipaCertMapVersion`?

install/updates/73-certmap.update
```diff
@@ -0,0 +1,27 @@
+# Configuration for Certificate Identity Mapping
+dn: cn=certmap,cn=ipa,cn=etc,$SUFFIX
+default:objectclass: top
+default:objectclass: nsContainer
+default:objectclass: ipaConfigObject
+default:objectclass: ipaCertMapContainer
+default:objectclass: ipaCertMapConfigObject
```
HonzaCholasta: You can't have 2 structural object classes (`nsContainer` and `ipaCertMapConfigObject`) in an entry. [RFC 4512, section 2.4.2](https://tools.ietf.org/html/rfc4512#section-2.4.2) says:
> An object or alias entry is characterized by precisely one structural object class superclass chain which has a single structural object class as the most subordinate object class. This structural object class is referred to as the structural object class of the entry.

I would define `ipaCertMapConfigObject` as auxiliary to fix this.

ipaserver/plugins/baseuser.py
```diff
@@ -728,3 +739,145 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
...
+        issuer = options.get('issuer')
+        subject = options.get('subject')
+        certificates = options.get('certificate', [])
+
+        # If only LOGIN is supplied, then we need either subject or issuer or
+        # certificate
+        if not certmapdatas:
+            valid_options = (issuer, subject, certificates)
+            if all(not option for option in valid_options):
+                raise errors.RequirementError(name='CERTMAPDATA')
```
HonzaCholasta: Nitpick, but IMO this would be a little bit more readable:
```python
        if not certmapdatas and not issuer and not subject and not certificates:
            raise errors.RequirementError(name='ipacertmapdata')
```
(BTW you should use the name of the param in `RequirementError`, like above.)

ipaserver/plugins/baseuser.py
```diff
@@ -728,3 +739,145 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
...
+                    reason=_('cannot specify both subject/issuer '
+                             'and CERTMAPDATA'))
+
+        # if the command is called with --subject --issuer or --certificate
+        # we need to add ipacertmapdata to the attrs_list in order to
+        # display the resulting value in the command output
+        if 'ipacertmapdata' not in attrs_list:
+            attrs_list.append('ipacertmapdata')
+
+        self._convert_options_to_certmap(entry_attrs, options)
```
HonzaCholasta: Nitpick, but I think it would be better to pass `certmapdatas`, `issuer`, `subject` and `certificates` as arguments rather than `options`, so you can avoid the redundant `options.get()` etc. in `_convert_options_to_certmap()`.

ipaserver/plugins/certmap.py
```diff
@@ -0,0 +1,396 @@
...
+    :param associateddomain: domains to be checked
+
+    :raises: ValidationError if the domain is neither IPA domain nor trusted
+    """
+    domains = options.get('associateddomain')
+    if domains:
+        trust_suffix_namespace = set()
+        trust_suffix_namespace.add(api_inst.env.domain.lower())
+
+        trust_objects = api_inst.Command.trust_find(u'', sizelimit=0)['result']
```
HonzaCholasta: Is the empty string necessary? It shouldn't be.

ipaserver/plugins/certmap.py
```diff
@@ -0,0 +1,396 @@
...
+            },
+            'default_privileges': {
+                'Certificate Identity Mapping Administrators'},
+        },
+    }
+
+    def get_dn(self, *keys, **options):
+        rulename = keys[-1]
+        dn = super(certmaprule, self).get_dn(rulename, **options)
+        return dn
```
HonzaCholasta: It looks like this method does effectively nothing. What is its purpose?

ipaserver/plugins/baseuser.py
```diff
@@ -728,3 +739,145 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
...
+        ),
+        DNParam(
+            'subject?',
+            cli_name='subject',
+            label=_('Subject'),
+            doc=_('Subject of the certificate'),
+            flags=['virtual_attribute']
+        ),
+        Bytes(
+            'certificate*', validate_certificate,
```
HonzaCholasta: Outside of basic certificate validation, we should also validate that the certificate has non-empty subject.

install/updates/73-certmap.update
```diff
@@ -0,0 +1,24 @@
+# Configuration for Certificate Identity Mapping
+dn: cn=certmap,cn=ipa,cn=etc,$SUFFIX
+default:objectclass: top
+default:objectclass: nsContainer
+default:objectclass: ipaConfigObject
```
HonzaCholasta: Now that the `ipaConfigString` is gone, `ipaConfigObject` is not necessary anymore.

ipaserver/plugins/baseuser.py
```diff
@@ -728,3 +739,154 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
...
+    )
+
+    @staticmethod
+    def _build_mapdata(subject, issuer):
+        return u'X509:<I>{issuer}<S>{subject}'.format(
+            issuer=issuer.x500_text(), subject=subject.x500_text())
+
+    @classmethod
+    def _convert_options_to_certmap(cls, entry_attrs, issuer=None,
+                                    subject=None, certificates=[]):
```
MartinBasti: Please don't use mutable python types as default value in options. Yous should use `certificates=()`

I know that this is on many places in code but we are trying to get rid of it.

HonzaCholasta:
LGTM. @flo-renaud, don't forget to register the new OIDs.

sumit-bose:
It looks like the ACis on the latest version do not allow hosts to access the rules. When I do 'kinit -k' on the IPA server or a client and call

    ldapsearch -H ldap://ipa-server.ipa.devel '(&(objectClass=ipaCertMapRule)(ipaEnabledFlag=TRUE))' -Y GSSAPI

I do not get any results. When I call 'kinit admin' and use the same ldapsearch I get my rule returned. Can you confirm this or is my test system broken?

flo-renaud:
Hi @sumit-bose ,
I am not able to reproduce this issue:
`[root@vm-161 ~]# kinit -k
[root@vm-161 ~]# klist
Ticket cache: KEYRING:persistent:0:krb_ccache_h6XRpeK
Default principal: host/vm-161.example.com@DOM-161.EXAMPLE.COM

Valid starting       Expires              Service principal
02/22/2017 21:30:10  02/23/2017 21:30:10  krbtgt/DOM-161.EXAMPLE.COM@DOM-161.EXAMPLE.COM
[root@vm-161 ~]# ldapsearch -H ldap://vm-161 '(&(objectClass=ipaCertMapRule)(ipaEnabledFlag=TRUE))' -Y GSSAPI -LLL
SASL/GSSAPI authentication started
SASL username: host/vm-161.example.com@DOM-161.EXAMPLE.COM
SASL SSF: 56
SASL data security layer installed.
dn: cn=rule1,cn=certmaprules,cn=certmap,dc=dom-161,dc=example,dc=com
objectClass: ipacertmaprule
objectClass: top
cn: rule1
description: d1
ipaEnabledFlag: TRUE
`
Do you have the ACI "permission:System: Read Certmap Rules" defined on dn: cn=certmaprules,cn=certmap,$BASEDN? It should grant access to ldap:///all

sumit-bose:
Ok, sorry for the noise, I tested on a fresh install again and now it is working as expected. I guess I shouldn't have tried to update from an older version of your patch to a newer one.

dkupka:
Works for me.

dkupka:
master:

* 9e24918c89f30a6d7064844dc0dd848bb35140df Support for Certificate Identity Mapping

PR #400, opened 2017-01-18
WebUI: Certificate Mapping

pvomacka:
Add WebUI for certificate mapping

https://fedorahosted.org/freeipa/ticket/6601

### DIFF
--- a/install/ui/src/freeipa/navigation/menu_spec.js
+++ b/install/ui/src/freeipa/navigation/menu_spec.js
@@ -151,7 +151,21 @@ var nav = {};
                     ]
                 },
                 { entity: 'otptoken' },
-                { entity: 'radiusproxy' }
+                { entity: 'radiusproxy' },
+                {
+                    entity: 'certmaprule',
+                    facet: 'search',
+                    children: [
+                        {
+                            entity: 'certmaprule',
+                            facet: 'search'
+                        },
+                        {
+                            entity: 'certmapconfig',
+                            facet: 'details'
+                        }
+                    ]
+                }
             ]
         },
         {
--- a/install/ui/src/freeipa/plugins/certmap.js
+++ b/install/ui/src/freeipa/plugins/certmap.js
@@ -1,0 +1,381 @@
+//
+// Copyright (C) 2017  FreeIPA Contributors see COPYING for license
+//
+
+
+define([
+        'dojo/_base/lang',
+        'dojo/_base/declare',
+        'dojo/Evented',
+        'dojo/on',
+        '../navigation',
+        '../field',
+        '../ipa',
+        '../phases',
+        '../reg',
+        '../widget',
+        '../text',
+        '../util',
+        // plain imports
+        '../search',
+        '../entity'],
+            function(lang, declare, Evented, on, navigation, mod_field, IPA,
+                     phases, reg, widget_mod, text, util) {
+/**
+ * Certificate map module
+ * @class
+ */
+var certmap = IPA.certmap = {
+
+    search_facet_group: {
+        facets: {
+            certmaprule_search: 'certmaprule_search',
+            certmapconfig: 'certmapconfig_details'
+        }
+    }
+};
+
+var make_certmaprule_spec = function() {
+return {
+    name: 'certmaprule',
+    facets: [
+        {
+            $type: 'search',
+            always_request_members: true,
+            details_facet: 'details',
+            facet_groups: [certmap.search_facet_group],
+            facet_group: 'search',
+            row_enabled_attribute: 'ipaenabledflag',
+            columns: [
+                'cn',
+                {
+                    name: 'ipaenabledflag',
+                    label: '@i18n:status.label',
+                    formatter: 'boolean_status'
+                },
+                'description'
+            ],
+            actions: [
+                'batch_disable',
+                'batch_enable'
+            ],
+            control_buttons: [
+                {
+                    name: 'disable',
+                    label: '@i18n:buttons.disable',
+                    icon: 'fa-minus'
+                },
+                {
+                    name: 'enable',
+                    label: '@i18n:buttons.enable',
+                    icon: 'fa-check'
+                }
+            ]
+        },
+        {
+            $type: 'details',
+            disable_facet_tabs: true,
+            facet_groups: [certmap.search_facet_group],
+            facet_group: 'search',
+            actions: [
+                'enable',
+                'disable',
+                'delete'
+            ],
+            header_actions: ['enable', 'disable', 'delete'],
+            state: {
+                evaluators: [
+                    {
+                        $factory: IPA.enable_state_evaluator,
+                        field: 'ipaenabledflag'
+                    }
+                ]
+            },
+            sections: [
+                {
+                    name: 'details',
+                    fields: [
+                        'cn',
+                        {
+                            $type: 'textarea',
+                            name: 'description'
+                        },
+                        {
+                            name: 'ipacertmapmaprule',
+                            tooltip: {
+                                title: '@mc-opt:certmaprule_add:ipacertmapmaprule:doc'
+                            }
+                        },
+                        {
+                            name: 'ipacertmapmatchrule',
+                            tooltip: {
+                                title: '@mc-opt:certmaprule_add:ipacertmapmatchrule:doc'
+                            }
+                        },
+                        {
+                            $type: 'multivalued',
+                            name: 'associateddomain',
+                            tooltip: {
+                                title: '@mc-opt:certmaprule_add:associateddomain:doc'
+                            }
+                        },
+                        {
+                            name: 'ipacertmappriority',
+                            tooltip: {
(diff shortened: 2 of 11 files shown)


install/ui/src/freeipa/plugins/certmap.js
```diff
@@ -0,0 +1,261 @@
...
+     */
+    that.create_add_options = function() {
+        var options = {};
+        var certmapdata = that.adder_dialog.get_field('ipacertmapdata').value;
+
+        options['ipacertmapdata'] = certmapdata;
+
+        return options;
+    };
+
```
flo-renaud: Hi @pvomacka,
thanks for the patch. Would it be possible to have a specific Add dialog window, which would allow to add Certificate mapping data with different input formats? Currently, the dialog takes the input and calls the equivalent of user-add-certmap --data, but it would also be interesting to allow the user to provide a certificate and call user-add-certmap --certificate, or allow the user to provider a subject + issuer and call user-add-certmap --subject --issuer.

pvomacka: Hi @flo-renaud,
Yes, it is possible and not hard to implement. I can add it.

dkupka: @flo-renaud After some thinking I would rather remove --issuer, --subject and --certificate from user-add-certmap command. I don't like the fact that single command does several different things and I think it's confusing.
I think the idea is great but I don't know how to do it in straightforward (for the user) way.

pvomacka: @dkupka @flo-renaud I would say that it would be kind of hard (especially in CLI) for users to understand which options should be used together and which can't be used together. From point of view of WebUI, there is no problem to create dialog, which allows only correct combinations of options.

install/ui/src/freeipa/plugins/certmap.js
```diff
@@ -0,0 +1,394 @@
...
+            policies: [
+                {
+                    $factory: IPA.multiple_choice_section_policy,
+                    widget: 'type'
+                }
+            ]
+        };
+
+        that.adder_dialog = certmap.custom_adder_dialog(spec);
+        that.adder_dialog.create_button({
```
pvoborni: Could we use confirmation mixin for the adder dialog? So that you would have to define the buttons? Or if it is not possible then define a dialog class for custom_command_multivalued_widget

install/ui/src/freeipa/navigation/menu_spec.js
```diff
@@ -151,7 +151,21 @@ var nav = {};
                     ]
                 },
                 { entity: 'otptoken' },
-                { entity: 'radiusproxy' }
+                { entity: 'radiusproxy' },
+                {
+                    entity: 'certmaprule',
```
pvoborni: Maybe more a question for UX person: is the position of the menu item on good place? Right now the order is: 
* Certificates
* OTP Tokens
* RADIUS Servers
* Certificate Identity Mapping Rules

Following order would be more natural to me:
* Certificates
* Certificate Identity Mapping Rules
* OTP Tokens
* RADIUS Servers

But I don't insist on it.

install/ui/test/data/ipa_init.json
```diff
@@ -315,6 +315,18 @@
                             "view_certificate": "Certificate for ${entity} ${primary_key}",
                             "view_certificate_btn": "View Certificate"
                         },
+                        "certmap": {
```
pvoborni: If I'm correct, some of your other PRs no longer update this file. We've talk to remove the offline mode. So what about removing it? Or is this still used for unit tests?

install/ui/src/freeipa/plugins/certmap.js
```diff
@@ -0,0 +1,394 @@
...
+                    ]
+                }
+            ]
+        }
+    ],
+    adder_dialog: {
+        $factory: certmap.rule_adder_dialog,
+        fields: [
+            'cn',
+            'ipacertmapissuer',
```
pvoborni: In your testing vm, ipacertmapissuer option doesn't exist.

install/ui/src/freeipa/plugins/certmap.js
```diff
@@ -0,0 +1,394 @@
...
+                    name: 'details',
+                    fields: [
+                        'cn',
+                        {
+                            $type: 'textarea',
+                            name: 'description'
+                        },
+                        'ipacertmapissuer',
+                        'ipacertmapmaprule',
+                        'ipacertmapmatchrule',
```
pvoborni: Would be good if mapping and matching rules had examples written near the fields. Without reading any docs I don't have a clue what should be written there. Same in details page.

install/ui/src/freeipa/user.js
```diff
@@ -218,6 +218,15 @@ return {
                             label: '@i18n:objects.cert.certificates'
                         },
                         {
+                            $type: 'certmap_multivalued',
```
pvoborni: Would be good if it had the question mark hint to describe what is the purpose of this field. What problem can be solved by it. I believe it is not self explanatory.

install/ui/src/freeipa/plugins/certmap.js
```diff
@@ -0,0 +1,394 @@
...
+                    child_spec: {
+                        $type: 'textarea'
+                    }
+                },
+                {
+                    name: 'issuer'
+                },
+                {
+                    name: 'subject'
+                }
```
pvoborni: Example in issuer and subject would be also helpful.

install/ui/test/data/ipa_init.json
pvomacka: Yes, you're right. I will remove this change from current PR and will create new PR for removing offline WebUI.

install/ui/src/freeipa/plugins/certmap.js
```diff
@@ -0,0 +1,394 @@
...
+            sections: [
+                {
+                    name: 'details',
+                    fields: [
+                        'cn',
+                        {
+                            $type: 'textarea',
+                            name: 'description'
+                        },
+                        'ipacertmapissuer',
```
flo-renaud: You can remove the ipacertmapissuer attribute as the new design doesn't define this attr any more.

install/ui/src/freeipa/plugins/certmap.js
```diff
@@ -0,0 +1,394 @@
...
+    spec.fields = [
+        {
+            $type: 'multivalued',
+            name: 'ipacertmapdata',
+            label: '@i18n:objects.certmap.data_label',
+            widget: 'type.ipacertmapdata'
+        },
+        {
+            $type: 'multivalued',
+            name: 'usercertificate',
```
flo-renaud: this should be 'certificate'

install/ui/src/freeipa/plugins/certmap.js
```diff
@@ -0,0 +1,394 @@
...
+
+    spec.widgets = [
+        {
+            $type: 'multiple_choice_section',
+            name: 'type',
+            choices: [
+                {
+                    name: 'data',
+                    label: '@i18n:objects.certmap.data_label',
+                    fields: ['ipacertmapdata', 'usercertificate'],
```
flo-renaud: This should be 'certificate'

install/ui/src/freeipa/plugins/certmap.js
```diff
@@ -0,0 +1,394 @@
...
+                }
+            ],
+            widgets: [
+                {
+                    $type: 'multivalued',
+                    name: 'ipacertmapdata'
+                },
+                {
+                    $type: 'multivalued',
+                    name: 'usercertificate',
```
flo-renaud: This should be 'certificate'

pvomacka:
Hello @flo-renaud and @pvoborni 

thank you for reviews, all proposed changes are done in last commits, please look at them. Thank you very much.

flo-renaud:
Hi @pvomacka 
Thank you for the updated PR.
I probably wongly advised you to replace 'usercertificate' with 'certificate' in one extra place where it was not needed, because now the "Certificates" field of the user details page does not display any more the full certificates. My bad...
Apart from that, everything works as expected. Thanks!

pvomacka:
Hi @flo-renaud 
Thank you for review. 

The issue about certificates is different and here is the fix: https://github.com/freeipa/freeipa/pull/519

pvomacka:
In last update I changed just line 33 in certmap.js file.

flo-renaud:
Hi @pvomacka 
thank you, LGTM.

pvomacka:
@pvoborni Thanks for review. I removed the space :)

tomaskrizek:
master:

* 27027bbc9cf7faa29c3c94686635559cbcbde98a WebUI: Add possibility to set field always writable
* fba318b83337b71ccb3421690071a130171fbdfe WebUI: Create non editable row widget for mutlivalued widget
* d3700275c1b63aeeab13c7dd9e09249bc2c8e4d7 WebUI: Add Custom command multivalued adder dialog
* 19426f32ff99feb7c64a4174728cd2b6b946a49a WebUI: Add certmap module

PR #404, opened 2017-01-19
tests: Add LDAP URI to ldappasswd explicitly

davidkupka:
Test should always respect api.env.* values.

https://fedorahosted.org/freeipa/ticket/6622

### DIFF
--- a/ipatests/test_integration/util.py
+++ b/ipatests/test_integration/util.py
@@ -21,6 +21,7 @@ import time
 import re
 
 from ipaplatform.paths import paths
+from ipalib import api
 from ipalib.constants import DEFAULT_CONFIG
 
 def run_repeatedly(host, command, assert_zero_rc=True, test=None,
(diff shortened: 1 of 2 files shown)


tiran:
```ipatests/test_integration/util.py``` calls ldappasswd without ```-H``` option, too. Related to the issue at hand, ```ipaserver/install/service.py``` has a similar issue with ldapmodify.

**Comment by dkupka** [comment]
@tiran Nice catch, I've added it to ipaserver/install/service.py, too. but I would rather not extend this outside of tests.

**Comment by dkupka** [comment]
@tiran After a closer look I believe that it's handled correctly in ipaserver/install/service.py
https://github.com/freeipa/freeipa/blob/master/ipaserver/install/service.py#L205:L208

**Comment by tiran** [comment]
I stand corrected! Thanks David.

File: ipatests/test_integration/util.py [hunk 1]
```diff
@@ -86,5 +87,5 @@ def ldappasswd_user_change(user, oldpw, newpw, master):
     userdn = "uid={},{},{}".format(user, container_user, basedn)
 
     args = [paths.LDAPPASSWD, '-D', userdn, '-w', oldpw, '-a', oldpw,
-            '-s', newpw, '-x']
+            '-s', newpw, '-x', '-H', api.env.ldap_uri]
```
MartinBasti: Have you tested this?

This code is executed on controller machine, thus api.env may be undefined, because there is no installed IPA.
I'm not sure if ldap_uri may be diferent on each (I assume than ldapi URI is the same on all machines, but not ldaps)

martbab: @MartinBasti I agree with your concerns. I guess constructed LDAP URI from passed in master parameter sounds as a safer bet.

PR #405, opened 2017-01-20
ipa-restore must stop tracking PKINIT cert in the preparation phase

flo-renaud:
ipa-restore calls certmonger to stop tracking the PKI certs, HTTP and DS certs.
It must also stop tracking the newly introduced PKINIT cert (stored in
/var/kerberos/krb5kdc/kdc.crt).

Otherwise the restore operation ends up with PKINIT cert tracked twice and
uninstallation fails.

https://fedorahosted.org/freeipa/ticket/6570


martbab:
Thanks, the patch looks ok and backup-restore tests passed.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/ceec512b09002e8cf9388873418644ec584db30a

PR #406, opened 2017-01-20
_resolve_records: fix assert, nameserver_ip can be none

MartinBasti:


martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/ccea23138ba6e9b54c08d472341ddbd64ffc45df

PR #408, opened 2017-01-23
ipaldap: properly escape raw binary values in LDAP filters

HonzaCholasta:
Manually escape each byte in the value, do not use
ldap.filter.escape_filter_chars() as it does not work with bytes in
Python 3.

### DIFF
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -19,6 +19,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+import binascii
 import time
 import datetime
 from decimal import Decimal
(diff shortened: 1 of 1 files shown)


ipapython/ipaldap.py
```diff
@@ -1245,11 +1245,10 @@ def make_filter_from_attr(
             return cls.combine_filters(flts, rules)
         elif value is not None:
             if isinstance(value, bytes):
-                if six.PY3:
-                    value = value.decode('raw_unicode_escape')
+                value = u''.join(
+                    u'\\{:02x}'.format(i) for i in six.iterbytes(value))
```
tiran: format() with hex is rather inefficient and slow. An implementation with hexlify() is five to ten times faster.

```
def enc_hexlify(s):
    hexstring = binascii.hexlify(s).decode('ascii')
    # [-2:0] is empty string for the initial \
    return u'\\'.join(hexstring[i:i+2] for i in six.moves.range(-2, len(s)+2, 2))
```

HonzaCholasta: This is not performance critical code, but OK (I like the empty initial string hack :+1:).

MartinBasti:
Works for me.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/84a9611cb885f04c72cd657c3a3e7bc4aff39d93

PR #414, opened 2017-01-28
SPEC: Update SELinux file context of ipa-otpd

lslebodn:
The ipa-otpd binary was moved but SELinux was not updated.
This is a workaround for developers who develop on platforms
with older SELinux policy.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -243,6 +243,7 @@ Requires(pre): systemd-units
 Requires(post): systemd-units
 Requires: selinux-policy >= %{selinux_policy_version}
 Requires(post): selinux-policy-base >= %{selinux_policy_version}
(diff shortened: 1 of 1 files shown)


martbab:
@lslebodn CI complains that your changes produce an invalid specfile. Can you please fix this. A better question, is there a demand to have this workaround in the spec file?

lslebodn:
On (10/03/17 00:28), Martin Babinsky wrote:
>@lslebodn CI complains that your changes produce an invalid specfile. Can you please fix this. A better question, is there a demand to have this workaround in the spec file?
>
It could be fixed; but there are more SELinux bugs now.
So it does not worth IMHO.

Feel free to close.

LS

martbab:
I agree with closing this, we will need more substantial policy update

PR #415, opened 2017-01-30
ca-del: require CA to already be disabled

frasertweedale:
Currently ca-del disables the target CA before deleting it.
Conceptually, this involves two separate permissions: modify and
delete.  A user with delete permission does not necessarily have
modify permission.

As we move toward enforcing IPA permissions in Dogtag, it is
necessary to decouple disablement from deletion, otherwise the
disable operation would fail if the user does not have modify
permission.  Although it introduces an additional step for
administrators, the process is consistent, required permissions are
clear, and errors are human-friendly.

freeipa-devel discussion: https://www.redhat.com/archives/freeipa-devel/2017-January/msg00435.html

### DIFF
--- a/ipaserver/plugins/ca.py
+++ b/ipaserver/plugins/ca.py
@@ -286,7 +286,12 @@ class ca_del(LDAPDelete):
 
         ca_id = self.api.Command.ca_show(keys[0])['result']['ipacaid'][0]
         with self.api.Backend.ra_lightweight_ca as ca_api:
-            ca_api.disable_ca(ca_id)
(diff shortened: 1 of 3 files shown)


apophys:
Could you please extend the tests with the invalid order of the commands on a ca entry?

frasertweedale:
@apophys done; PR updated.

frasertweedale:
Shelving this PR for now.  It might get resurrected later.  Discussion:
https://www.redhat.com/archives/freeipa-devel/2017-February/msg00150.html

PR #417, opened 2017-01-30
private_ccache: yield ccache name

frasertweedale:
When using private_ccache, yield 'path' from the context manager.
This is cleaner than inspecting 'os.environ['KRB5CCNAME']' within
the context.


frasertweedale:
Build failure is unrelated to patch.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/caca181d3b73c045abd72e464a195c6b61c251c7

PR #418, opened 2017-01-30
replica install: do not log host OTP

HonzaCholasta:
Do not log the value of the --password option of ipa-client-install when it
is run from ipa-replica-install before replica promotion.

https://fedorahosted.org/freeipa/ticket/6633

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -876,6 +876,7 @@ def ensure_enrolled(installer):
 
         args = [paths.IPA_CLIENT_INSTALL, "--unattended", "--no-ntp"]
         stdin = None
+        nolog = []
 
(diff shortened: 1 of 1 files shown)


ipaserver/install/server/replicainstall.py
```diff
@@ -876,6 +876,7 @@ def ensure_enrolled(installer):
 
         args = [paths.IPA_CLIENT_INSTALL, "--unattended", "--no-ntp"]
         stdin = None
+        nolog = []
```
Akasurde: LGTM

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/054c1e013aee6fdbee2e9966c32df02d91f0c2c1

PR #419, opened 2017-01-30
ipa-ca-install: do not fail without --subject-base and --ca-subject

HonzaCholasta:
When --subject-base and --ca-subject are not specified in ipa-ca-install,
default values are used. DN objects are used as the default values in
ipa-ca-install, but the CA installer expects the values to be strings. This
causes ipa-ca-install to fail unless both --subject-base and --ca-subject
are specified.

Convert the DN objects to strings to fix the issue.

https://fedorahosted.org/freeipa/ticket/2614

### DIFF
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -216,10 +216,11 @@ def install_master(safe_options, options):
     options.host_name = api.env.host
 
     if not options.subject_base:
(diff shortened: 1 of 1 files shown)


HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/87400cdec1054971f50f90a0c63f18ab045f3833

PR #420, opened 2017-01-30
Allow login to WebUI using Kerberos aliases/enterprise principals

martbab:
The logic of the extraction/validation of principal from the request and
subsequent authentication was simplified and most of the guesswork will be done
by KDC during kinit. This allows for using principal aliases/enterprise
principals to log into WebUI and also lays foundation for enabling logons from
users from trusted domains.

https://fedorahosted.org/freeipa/ticket/6343

### DIFF
--- a/ipaserver/rpcserver.py
+++ b/ipaserver/rpcserver.py
@@ -50,13 +50,12 @@ from ipalib.errors import (PublicError, InternalError, JSONError,
 from ipalib.request import context, destroy_context
 from ipalib.rpc import (xml_dumps, xml_loads,
     json_encode_binary, json_decode_binary)
-from ipalib.util import normalize_name
 from ipapython.dn import DN
 from ipaserver.plugins.ldap2 import ldap2
 from ipalib.backend import Backend
 from ipalib.krb_utils import (
-    krb5_format_principal_name,
     get_credentials_if_valid)
+from ipapython import kerberos
 from ipapython import ipautil
 from ipaplatform.paths import paths
 from ipapython.version import VERSION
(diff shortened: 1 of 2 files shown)


dkupka:
LGTM and works as expected. Not ACKing only because it's marked as WIP.

ipaserver/rpcserver.py
```diff
@@ -977,7 +960,7 @@ def __call__(self, environ, start_response):
         # Get the ccache we'll use and attempt to get credentials in it with user,password
         ipa_ccache_name = get_ipa_ccache_name()
         try:
-            self.kinit(user, self.api.env.realm, password, ipa_ccache_name)
+            self.kinit(unicode(user_principal), password, ipa_ccache_name)
```
dkupka: Nitpick: unicode is not necessary, kinit() does the conversion anyway.

martbab: Oh okay I missed that.

martbab:
I would wait with ACKing/pushing this PR until https://github.com/freeipa/freeipa/pull/314 is pushed.

martbab:
Now that privilege separation was implemented I have rebased the PR and request a proper review of this patch.

martbab:
@abbra can you also have a quick look at this PR if it is OK from the trusted user login perspective?

ipaserver/rpcserver.py
```diff
@@ -872,33 +871,15 @@ def __call__(self, environ, start_response):
...
-            # username is of the form NetBIOS\user
-            return self.unauthorized(environ, start_response, '', 'denied')
-
-        else:
+        try:
+            user_principal = kerberos.Principal(user)
+        except Exception:
             # username is of the form user or of some wild form, e.g.
-            # user@REALM1@REALM2 or NetBIOS1\NetBIOS2\user (see normalize_name)
+            # user@REALM1@REALM2 or NetBIOS1\NetBIOS2\user
```
abbra: The comment is not true because ipapython.kerberos.Principal does not expect NetBIOS1\user, it only expects normal Kerberos principals.

ipaserver/rpcserver.py
```diff
@@ -872,33 +871,15 @@ def __call__(self, environ, start_response):
...
+            return self.unauthorized(environ, start_response, '', 'denied')
 
-            # wild form username will fail at kinit, so nothing needs to be done
-            pass
+        if not (user_principal.is_user or user_principal.is_enterprise):
```
abbra: This code actually prevents services to contact us. you should also allow user_principal.is_service.

martbab: Does it make sense for a service to use WebUI? Or is the `login_kerberos` handler used anywhere else? AFAIK it handles only WebUI logins currently.

ipaserver/rpcserver.py
martbab: That's true I will fix that.

martbab:
@abbra I have a question regarding one of your comments, please review.

ipaserver/rpcserver.py
abbra: login_kerberos can be used by any API user, e.g. any Kerberos principal. I'd prefer to not remove this possibility.

martbab: Thanks for catching this, I have not realized that. I will then remove all restrictions apart from being parseable principal name.

abbra:
Thanks. LGTM and works for me with IPA user, IPA host principal, and AD user. The latter cannot yet actually use Web UI but that is a separate PR.

dkupka:
master:

* f8d7e37a091c1df4c989b80b8d19e12ab35533c8 Allow login to WebUI using Kerberos aliases/enterprise principals

PR #422, opened 2017-01-31
Fix reference before assignment

frasertweedale:
In 'store_session_cookie', if the server does not set the session
cookie for some reason, the 'session_cookie' variable does not get
assigned, resulting in UnboundLocalError.  Set an initial value of
'None'.

Fixes: https://fedorahosted.org/freeipa/ticket/6636


MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/924794f62b9d3d0f46ca18e4f9338eaed865c03e

PR #423, opened 2017-01-31
dns-update-system-records: add support for nsupdate output format

MartinBasti:
Option --out does the trick


ipaclient/plugins/dns.py
```diff
@@ -417,6 +418,63 @@ def interactive_prompt_callback(self, kw):
...
+    def _nsupdate_output_file(self, file, result, labels):
+        """Store data in nsupdate format in file"""
+        def parse_rname_rtype(record):
+            """Get rname and rtype from textual representation of record"""
+            l = record.split(' ', 5)
+            return l[0], l[3]
+
+        already_removed = set()
+        for key in self.record_groups:
+            if result.get(key):
```
tomaskrizek: Is it possible to use `key in result` here? It would improve readability.

ipaclient/plugins/dns.py
```diff
@@ -417,6 +418,63 @@ def interactive_prompt_callback(self, kw):
...
+                    # delete old first
+                    r_name_type = parse_rname_rtype(val)
+                    if r_name_type not in already_removed:
+                        # remove it only once
+                        already_removed.add(r_name_type)
+                        file.write("update delete {rname} {rtype}\n".format(
+                            rname=r_name_type[0], rtype=r_name_type[1]
+                        ))
+                    # add new
+                    file.write("update add {}\n".format(val))
```
tomaskrizek: What happens if `update add ...` is ran twice for the same record?

ipaclient/plugins/dns.py
```diff
@@ -431,11 +489,12 @@ def output_for_cli(self, textui, output, *args, **options):
...
-                textui.print_indented(u'{}:'.format(labels[key]), indent=1)
-                for val in sorted(result[key]):
-                    textui.print_indented(val, indent=2)
-                textui.print_line(u'')
+
+        self._standard_output(textui, result, labels)
+
+        out = options.get('out')  # output to file
+        if out:
+            with open(out, "w") as f:
```
tomaskrizek: I think it'd be nice if we had try-catch block here to handle potential errors with file operations.

tomaskrizek:
I added some in-line comments/questions.

ipaclient/plugins/dns.py
MartinBasti: ok, I'll use the same code as cert.py does for certificates

ipaclient/plugins/dns.py
MartinBasti: This shouldn't happen

ipaclient/plugins/dns.py
MartinBasti: fixed

ipaclient/plugins/dns.py
```diff
@@ -417,6 +418,63 @@ def interactive_prompt_callback(self, kw):
...
+                textui.print_indented(u'{}:'.format(labels[key]), indent=1)
+                for val in sorted(result[key]):
+                    textui.print_indented(val, indent=2)
+                textui.print_line(u'')
+
+    def _nsupdate_output_file(self, file, result, labels):
+        """Store data in nsupdate format in file"""
+        def parse_rname_rtype(record):
+            """Get rname and rtype from textual representation of record"""
+            l = record.split(' ', 5)
```
MartinBasti: fixed

ipaclient/plugins/dns.py
```diff
@@ -417,6 +418,63 @@ def interactive_prompt_callback(self, kw):
...
+    record_groups = ('ipa_records', 'location_records')
+
+    def get_options(self):
+        for option in super(dns_update_system_records, self).get_options():
+            yield option
+        yield Str(
+            'out?',
+            include='cli',
+            doc=_('file to store DNS records in nsupdate format')
+        )
```
HonzaCholasta: Not that this won't work, but you can use `takes_options` as usual here, `MethodOverride` will tak care of properly merging the remote and local options.

ipaclient/plugins/dns.py
```diff
@@ -431,11 +489,15 @@ def output_for_cli(self, textui, output, *args, **options):
...
+
+        self._standard_output(textui, result, labels)
+
+        out = options.get('out')  # output to file
+        if out:
+            try:
+                with open(out, "w") as f:
+                    self._nsupdate_output_file(f, result, labels)
+            except (OSError, IOError) as e:
+                raise errors.FileError(reason=unicode(e))
```
HonzaCholasta: This block should be moved to `forward()`, `output_for_cli()` is supposed to just format the result on the CLI, not write stuff to files.

ipaclient/plugins/dns.py
tomaskrizek: So the command `update delete ...` will remove for example **all** A records for a domain and that's why you're checking that it won't run twice. However, when the records are added, they should be all different (in case of A record, different IPv4). Is that correct?

pvoborni:
I've added acceptance criteria and user story to the related FreeIPA ticket.   

I miss a "how to use part" - a specific example. This should be in FreeIPA.org wiki, e.g. in design page (rest of the design page can be copied user story and empty), but the how to use section with both auth methods is a critical part.

ipaclient/plugins/dns.py
MartinBasti: yes IPA doesn't return duplicated records

ipaclient/plugins/dns.py
MartinBasti: I did exactly the same as was used in user.py plugin

ipaclient/plugins/dns.py
MartinBasti: not fixed, reverting. I did that on purpose to not generate empty location records

```
if key in result and result[key]:
```
can be replaced by
```
if result.get(key):
```

MartinBasti:
@pvoborni http://www.freeipa.org/page/Howto/Updating_FreeIPA_system_DNS_records_on_a_remote_DNS_server

Still WIP, but can be reviewed if format fulfill expectations.

tomaskrizek:
Please update the ticket in trac/JIRA to mentiond the command does not support stdout. LGTM otherwise.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/7eb2ef61905a5c6ddf04237f0aa84e7585e1186d
https://fedorahosted.org/freeipa/changeset/5bd82174233095a3cccfbbf8524622440c31b10c

PR #424, opened 2017-01-31
Tests: fix wait_for_replication task

MartinBasti:
DS changed a format of replication status attribute. Now it is with
prefix "Error (x)" where x is the error code.

Both formats were kept to allow tests run on older and new
versions of DS.


MartinBasti:
bump for review

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/ad1a5551d5ec716dc745f39e82d38cc634229cb0

PR #425, opened 2017-01-31
ipa-kra-install must create directory if it does not exist

flo-renaud:
ipa-kra-install creates an admin cert file in
/root/.dogtag/pki-tomcat/ca_admin.cert but does not check that the
parent directory exists. This situation can happen when uninstall + restore
has been run.

The fix creates the directory if not present.

https://fedorahosted.org/freeipa/ticket/6606


MartinBasti:
Pushed: 

master:

066f5b7c904208d0fd79862dfaa7166fff42fd30 ipa-kra-install must create directory if it does not exist

PR #426, opened 2017-01-31
DNSSEC: forwarders validation improvement

MartinBasti:
Some DNS servers behaves oddly and instead sending result without RRSIG records
don't reply at all when DNSSEC flag is enabled (timeout). Instead of
hard error IPA should this handle as DNSSEC error and continue with
installation/adding forwarders.

### DIFF
--- a/ipalib/util.py
+++ b/ipalib/util.py
@@ -670,8 +670,7 @@ def validate_dnssec_global_forwarder(ip_addr, log=None, timeout=10):
                               timeout=timeout)
     except DNSException as e:
         _log_response(log, e)
-        raise UnresolvableRecordError(owner=owner, rtype=rtype, ip=ip_addr,
-                                      error=e)
+        raise DNSSECSignatureMissingError(owner=owner, rtype=rtype, ip=ip_addr)
 
     try:
         ans.response.find_rrset(


ipalib/util.py
```diff
@@ -670,8 +670,7 @@ def validate_dnssec_global_forwarder(ip_addr, log=None, timeout=10):
                               timeout=timeout)
     except DNSException as e:
         _log_response(log, e)
-        raise UnresolvableRecordError(owner=owner, rtype=rtype, ip=ip_addr,
-                                      error=e)
+        raise DNSSECSignatureMissingError(owner=owner, rtype=rtype, ip=ip_addr)
```
tomaskrizek: Please update the docstring since the function no longer raises `UnresolvableRecordError`.

tomaskrizek:
I think the same issue can also occur in `validate_dnssec_zone_forwarder_step2()`.

ipalib/util.py
MartinBasti: It still can due `_validate_edns0_forwarder()` call

MartinBasti:
I was thinking about it and I disagree. Checking forwards zone behaves differently than check for global forwarders. The validate `_dnssec_zone_forwarder_step2` is called aginst an IPA DNS server, so there shouln't be that issue as we know how bind is supposed to work.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/387a1513bb9dc0dc546753bfaa8a59aae8f30b83

PR #427, opened 2017-01-31
[Py3] WSGI part 2

MartinBasti:
with this PR:
* server can be installed with python3-mod_wsgi
* any xmlrpc test can be executed to find a new py3 issues (still a lot of them there)

### DIFF
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -308,7 +308,7 @@ def json_encode_binary(val, version):
             encoded = encoded.decode('ascii')
         return {'__base64__': encoded}
     elif isinstance(val, Decimal):
-        return {'__base64__': base64.b64encode(str(val))}
+        return unicode(val)
     elif isinstance(val, DN):
         return str(val)
     elif isinstance(val, datetime.datetime):
--- a/ipaserver/plugins/baseldap.py
+++ b/ipaserver/plugins/baseldap.py
@@ -1036,7 +1036,7 @@ last, after all sets and adds."""),
                     except ValueError:
                         if isinstance(delval, bytes):
                             # This is a Binary value, base64 encode it
-                            delval = unicode(base64.b64encode(delval))
+                            delval = base64.b64encode(delval).decode('ascii')
                         raise errors.AttrValueNotFound(attr=attr, value=delval)
 
         # normalize all values
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -849,7 +849,7 @@ class LDAPClient(object):
                               Principal)):
             return six.text_type(val).encode('utf-8')
         elif isinstance(val, DNSName):
-            return val.to_text()
+            return val.to_text().encode('ascii')
         elif isinstance(val, bytes):
             return val
         elif isinstance(val, list):
--- a/ipaserver/plugins/ca.py
+++ b/ipaserver/plugins/ca.py
@@ -4,8 +4,6 @@
 
 import base64
 
-import six
-
 from ipalib import api, errors, output, Bytes, DNParam, Flag, Str
 from ipalib.constants import IPA_CA_CN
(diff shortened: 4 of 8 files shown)


ipaserver/plugins/baseldap.py
```diff
@@ -1036,8 +1037,8 @@ def process_attr_options(self, entry_attrs, dn, keys, options):
                     except ValueError:
                         if isinstance(delval, bytes):
                             # This is a Binary value, base64 encode it
-                            delval = unicode(base64.b64encode(delval))
-                        raise errors.AttrValueNotFound(attr=attr, value=delval)
+                            delval = base64.b64encode(delval).decode('ascii')
+                            raise errors.AttrValueNotFound(attr=attr, value=delval)
```
tiran: Why did you indent the line?

MartinBasti: I have no idea

ipaserver/plugins/baseldap.py
```diff
@@ -255,7 +255,8 @@ def entry_to_dict(entry, **options):
                 value = list(entry.raw[attr])
                 for (i, v) in enumerate(value):
                     try:
-                        value[i] = v.decode('utf-8')
+                        if isinstance(v, bytes):
+                            value[i] = v.decode('utf-8')
```
HonzaCholasta: `entry.raw[]` should always return a list of bytes. Why the `if`?

ipalib/rpc.py
```diff
@@ -308,7 +308,7 @@ def json_encode_binary(val, version):
             encoded = encoded.decode('ascii')
         return {'__base64__': encoded}
     elif isinstance(val, Decimal):
-        return {'__base64__': base64.b64encode(str(val))}
+        return {'__base64__': base64.b64encode(str(val).encode('ascii'))}
```
HonzaCholasta: The value must be text, but `base64.b64encode()` returns binary data, so you should also decode its result.

ipaserver/plugins/baseldap.py
MartinBasti: will be fixed in ipaldap

ipalib/rpc.py
MartinBasti: ok

ipalib/rpc.py
```diff
@@ -308,7 +308,8 @@ def json_encode_binary(val, version):
             encoded = encoded.decode('ascii')
         return {'__base64__': encoded}
     elif isinstance(val, Decimal):
-        return {'__base64__': base64.b64encode(str(val))}
+        return {'__base64__': base64.b64encode(
+            str(val).encode('ascii')).decode('ascii')}
```
HonzaCholasta: Could we please get rid of this non-sense and return `unicode(val)`? (It is a backward compatible change.)

MartinBasti: as you wish

HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/caa560ca79e4038b161b27d11e3f144606dbbcdb
https://fedorahosted.org/freeipa/changeset/a93b2bea5ce88a934ba2ab39bdaa518fb55064c4
https://fedorahosted.org/freeipa/changeset/a3d3b0ad2537c9d11d9c6108c31e079f0dfcf31c
https://fedorahosted.org/freeipa/changeset/03d0a55e8a21a334ca4dc625527cae93633a7314
https://fedorahosted.org/freeipa/changeset/a584758cfb87567a9c640ae107903b0f6c9fec30
https://fedorahosted.org/freeipa/changeset/ab53d80883320060769b7bfada2a813b345b9e4a
https://fedorahosted.org/freeipa/changeset/4c84341b8bc14cd19a4e2c2df4c13b95ff7eeb05

PR #429, opened 2017-01-31
[py3] ipactl restart: log httplib failues as debug

MartinBasti:
With python3 there are several excerptions ConnectionRefusedError raised
before ipactl is able to connect to dogtag after restart. These
exception should be logged on debug level until timeout is reached.

### DIFF
--- a/ipapython/dogtag.py
+++ b/ipapython/dogtag.py
@@ -209,7 +209,7 @@ def _httplib_request(
         http_body = res.read()
         conn.close()
     except Exception as e:
-        root_logger.exception("httplib request failed:")
+        root_logger.debug("httplib request failed:", exc_info=True)
         raise NetworkError(uri=uri, error=str(e))
 
     root_logger.debug('response status %d',    http_status)


ipapython/dogtag.py
```diff
@@ -209,7 +210,8 @@ def _httplib_request(
         http_body = res.read()
         conn.close()
     except Exception as e:
-        root_logger.exception("httplib request failed:")
+        root_logger.debug("httplib request failed")
```
tiran: Instead of ```traceback.format.exc()```, use  ```root_logger.debug("httplib request failed",  exc_info=True)``` to log message with traceback.

MartinBasti: fixed

MartinBasti:
This happens with python2.7 too, I reproduced it today

tiran:
Yeah, I reported the issue as https://fedorahosted.org/freeipa/ticket/6674 . Feel free to close it as duplicate.

stlaz:
I don't see what this has to do with Py3. The issue is the same on Py2. Swap the ticket for the one of @tiran and I'll ack this. If this gets triaged for 4.4 as well we can backport it later.

MartinBasti:
Ticket corrected, commit msg ammended

stlaz:
Thanks, ACK.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/53c8e9a53f026d83d5328896d1ea0cf72690cf24

PR #430, opened 2017-02-01
[py3] tests_xmlrpc: do not call str() on bytes

MartinBasti:
Calling str() on bytes causes undesired side effect: it adds prefix "b"
to the result of conversion. The method decode() should be used instead.


MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/5de70e31999eb219bd47aa81b0c003a6c15cf748

PR #431, opened 2017-02-02
py3: ldapupdate: fix logging str(bytes) issue

MartinBasti:
bytes as argument of str() gives unexpected result by adding prefix "b"
there.

Also add missing safe_option() call to logging (it will fix another
str(bytes) issue)

Other byteswarnings are from https://github.com/etingof/pyasn1/issues/14


MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/b24787a67fd8b19b9222979a963a8f28b22153ee

PR #432, opened 2017-02-02
build: Add missing dependency on libxmlrpc{,_util}

dkupka:
Change in libxmlrpc packaging uncovered missing linking  dependency in our
build system.

https://fedorahosted.org/freeipa/ticket/6637

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -224,7 +224,7 @@ PKG_CHECK_MODULES([SASL], [libsasl2])
 dnl ---------------------------------------------------------------------------
 dnl - Check for XMLRPC-C
 dnl ---------------------------------------------------------------------------
-PKG_CHECK_MODULES([XMLRPC], [xmlrpc_client])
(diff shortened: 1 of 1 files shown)


tiran:
ACK

ipa-join uses functions from ```libxmlrpc.so``` (e.g. ```xmlrpc_string_new```) and from ```libxmlrpc_util.so``` (e.g. ```xmlrpc_env_init```). In the past it was no problem because ```libxmlrpc_client.so``` depends on both libraries and pulled the function in. Nowadays indirect linking triggers a DSO error. All libraries must be linked directly.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/f4088b3a00b3cbd1a0133ac90cba85e501573f76

MartinBasti:
Needs separate PR for ipa-4-4 branch

PR #434, opened 2017-02-04
csrgen: Automate full cert request flow

LiptonB:
Adds `--autogenerate` flag to `ipa cert-request` command. It no longer
requires a CSR passed on the command line, instead it creates a config
(bash script) with `cert-get-requestdata`, then runs it to build a CSR,
and submits that CSR.

Example usage (NSS database):
$ ipa cert-request --autogenerate --principal blipton --profile-id userCert --database /tmp/certs

Example usage (PEM private key file):
$ ipa cert-request --autogenerate --principal blipton --profile-id userCert --private-key /tmp/key.pem

### DIFF
--- a/ipaclient/plugins/cert.py
+++ b/ipaclient/plugins/cert.py
@@ -19,6 +19,11 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import subprocess
+from tempfile import NamedTemporaryFile as NTF
+
+import six
+
 from ipaclient.frontend import MethodOverride
 from ipalib import errors
 from ipalib import x509
@@ -27,16 +32,85 @@ from ipalib.parameters import File, Flag, Str
 from ipalib.plugable import Registry
 from ipalib.text import _
 
+if six.PY3:
+    unicode = str
+
 register = Registry()
 
 
 @register(override=True, no_fail=True)
 class cert_request(MethodOverride):
+    takes_options = (
+        Str(
+            'database?',
+            label=_('Path to NSS database'),
+            doc=_('Path to NSS database to use for private key'),
+        ),
+        Str(
+            'private_key?',
+            label=_('Path to private key file'),
+            doc=_('Path to PEM file containing a private key'),
+        ),
+    )
+
     def get_args(self):
         for arg in super(cert_request, self).get_args():
             if arg.name == 'csr':
-                arg = arg.clone_retype(arg.name, File)
+                arg = arg.clone_retype(arg.name, File, required=False)
             yield arg
+
+    def forward(self, csr=None, **options):
+        database = options.pop('database', None)
+        private_key = options.pop('private_key', None)
+
+        if csr is None:
+            if database:
+                helper = u'certutil'
+                helper_args = ['-d', database]
+            elif private_key:
+                helper = u'openssl'
+                helper_args = [private_key]
+            else:
+                raise errors.InvocationError(
+                    message=u"One of 'database' or 'private_key' is required")
+
+            with NTF() as scriptfile, NTF() as csrfile:
+                profile_id = options.get('profile_id')
+
+                self.api.Command.cert_get_requestdata(
+                    profile_id=profile_id,
+                    principal=options.get('principal'),
+                    out=unicode(scriptfile.name),
+                    helper=helper)
+
+                helper_cmd = [
+                    'bash', '-e', scriptfile.name, csrfile.name] + helper_args
+
+                try:
+                    subprocess.check_output(helper_cmd)
+                except subprocess.CalledProcessError as e:
+                    raise errors.CertificateOperationError(
+                        error=(
+                            _('Error running "%(cmd)s" to generate CSR:'
+                              ' %(err)s') %
+                            {'cmd': ' '.join(helper_cmd), 'err': e.output}))
+
+                try:
+                    csr = unicode(csrfile.read())
+                except IOError as e:
+                    raise errors.CertificateOperationError(
+                        error=(_('Unable to read generated CSR file: %(err)s')
+                               % {'err': e}))
+                if not csr:
+                    raise errors.CertificateOperationError(
+                        error=(_('Generated CSR was empty')))
+        else:
+            if database is not None or private_key is not None:
+                raise errors.MutuallyExclusiveError(reason=_(
+                    "Options 'database' and 'private_key' are not compatible"
+                    " with 'csr'"))
+
+        return super(cert_request, self).forward(csr, **options)
 
 
 @register(override=True, no_fail=True)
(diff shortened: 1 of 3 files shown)


MartinBasti:
- pylint:
```
************* Module ipaclient.plugins.cert
ipaclient/plugins/cert.py:102: [W1612(unicode-builtin), cert_request.forward] unicode built-in referenced)
ipaclient/plugins/cert.py:127: [W1612(unicode-builtin), cert_request.forward] unicode built-in referenced)
ipaclient/plugins/cert.py:99: [W0612(unused-variable), cert_request.forward] Unused variable 'requestdata')
```

for unicode you can use etiher `six.string_type()` or
```
if six.PY3:
    unicode = str
```

-  pep8 errors
- failing test expects DN object instead of String

ipaclient/plugins/cert.py
```diff
@@ -32,12 +36,96 @@
...
+
     def get_args(self):
         for arg in super(cert_request, self).get_args():
             if arg.name == 'csr':
-                arg = arg.clone_retype(arg.name, File)
+                arg = arg.clone_retype(arg.name, File, required=False)
             yield arg
 
+    def forward(self, *keys, **options):
+        autogenerate = options.pop('autogenerate', False)
```
HonzaCholasta: I don't think `autogenerate` needs to (or should) be a separate option:
```python
    def forward(self, csr=None, **options):
        autogenerate = csr is None
        ...
```

ipaclient/plugins/cert.py
```diff
@@ -32,12 +36,96 @@
...
+        Str(
+            'database?',
+            label=_('Path to NSS database'),
+            doc=_('Path to NSS database to use for private key'),
+        ),
+        Str(
+            'private_key?',
+            label=_('Path to private key file'),
+            doc=_('Path to PEM file containing a private key'),
+        ),
```
HonzaCholasta: You will most certainly need a password (file) as well for the NSS database or private key.

ipaclient/plugins/cert.py
```diff
@@ -32,12 +36,96 @@
...
+        if autogenerate:
+            if database:
+                helper = u'certutil'
+                helper_args = ['-d', database]
+            elif private_key:
+                helper = u'openssl'
+                helper_args = [private_key]
+            else:
+                raise errors.InvocationError(
+                    format="One of 'database' or 'private_key' is required")
```
HonzaCholasta: You should use `message=` here.

ipaclient/plugins/cert.py
```diff
@@ -32,12 +36,96 @@
...
+
+            scriptfile = tempfile.NamedTemporaryFile(delete=False)
+            scriptfile.close()
+            csrfile = tempfile.NamedTemporaryFile(delete=False)
+            csrfile.close()
+            csrfilename = csrfile.name
```
HonzaCholasta: It would be better to put these into a `with` statement so that they are deleted automatically after use:
```python
    with NamedTemporaryFile() as scriptfile, NamedTemporaryFile() as csrfile:
        ...
```

ipaclient/plugins/cert.py
```diff
@@ -32,12 +36,96 @@
...
+
+            # profile_id is optional for cert_request, but not for
+            # cert_get_requestdata, so pass the default explicitly when
+            # necessary
+            profile_id = options.get('profile_id')
+            if profile_id is None:
+                profile_id = self.get_default_of('profile_id')
```
HonzaCholasta: I know I suggested this to you, but I didn't realize it would require hard-coding the default profile name into `profile_id` in server-side `cert_request`. I would rather not do that and instead use `dogtag.DEFAULT_PROFILE` as the default value of `profile_id` in `cert_get_requestdata`. I'm aware that this goes against the idea of having the server decide what is the default profile, but `cert_get_requestdata` will be moved to the server, so this is only temporary.

ipaclient/plugins/cert.py
```diff
@@ -32,12 +36,96 @@
...
+
+            helper_cmd = ['bash', '-e', scriptfile.name, csrfilename] + helper_args
+
+            try:
+                subprocess.check_call(helper_cmd)
+            except subprocess.CalledProcessError:
+                raise errors.CertificateOperationError(
+                    error=(
+                        _('Error running "%(cmd)s" to generate CSR') %
+                        {'cmd': ' '.join(helper_cmd)}))
```
HonzaCholasta: It would be nice to at least log the original error, or use it as part of the new error message. Ditto for the `IOError` below.

ipaclient/plugins/cert.py
```diff
@@ -32,12 +36,96 @@
...
+            if not csr:
+                raise errors.CertificateOperationError(
+                    error=(_('Generated CSR was empty')))
+
+            rv = super(cert_request, self).forward(unicode(csr), **options)
+            os.remove(csrfilename)
+            return rv
+        else:
+            if database is not None or private_key is not None:
+                raise errors.OptionError(format=_(
```
HonzaCholasta: This should be a `MutuallyExclusiveError`.

LiptonB:
Thanks for the comments, and sorry about submitting this with lint errors. I think I've followed all of your suggestions, let me know what you think.

ipaclient/plugins/cert.py
```diff
@@ -27,17 +32,109 @@
...
+                helper = u'openssl'
+                helper_args = [private_key]
+                if password_file:
+                    helper_args += ['-passin', 'file:%s' % password_file]
+            else:
+                raise errors.InvocationError(
+                    message=u"One of 'database' or 'private_key' is required")
+
+            with tempfile.NamedTemporaryFile(
+                    ) as scriptfile, tempfile.NamedTemporaryFile() as csrfile:
```
HonzaCholasta: Nitpick: if you `from tempfile import NamedTemporaryFile`, the you would be able to fit this into a single line.

ipaclient/plugins/cert.py
```diff
@@ -27,17 +32,109 @@
...
+                    raise errors.CertificateOperationError(
+                        error=(_('Generated CSR was empty')))
+
+                return super(cert_request, self).forward(csr, **options)
+        else:
+            if database is not None or private_key is not None:
+                raise errors.MutuallyExclusiveError(reason=_(
+                    "Options 'database' and 'private_key' are not compatible"
+                    " with 'csr'"))
+            return super(cert_request, self).forward(csr, **options)
```
HonzaCholasta: Nitpick: if you move this call one intendation level to the left, you will be able to remove the forward call above.

HonzaCholasta:
Thank you. LGTM, but please squash the fixup commit.

LiptonB:
@HonzaCholasta thanks, updated!

ipaclient/plugins/cert.py
LiptonB: Had to shorten the name more to keep the descriptive variable names, is this ok?

HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/39a5d9c5aae77687f67d9be02457733bdfb99ead
https://fedorahosted.org/freeipa/changeset/4350dcdea22fd2284836315d0ae7d38733a7620e
https://fedorahosted.org/freeipa/changeset/ada91c20588046bb147fc701718d3da4d2c080ca

PR #435, opened 2017-02-06
py3: fix replica install regression

MartinBasti:
Principal object must be created from string not from bytes

https://fedorahosted.org/freeipa/ticket/6640


stlaz:
Works for me.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/91ab650ac42d34d4958e33da7ef0641842511a89

PR #436, opened 2017-02-06
x509: allow leading text in PEM files

HonzaCholasta:
This fixes a regression introduced in commit
b8d6524d43dd0667184aebc79fb77a9b8a46939a.

### DIFF
--- a/ipatests/test_ipalib/test_x509.py
+++ b/ipatests/test_ipalib/test_x509.py
@@ -69,6 +69,17 @@ class test_x509(object):
         x509.load_certificate((newcert,))
         x509.load_certificate([newcert])
 
+        # Load a good cert with headers and leading text
+        newcert = (
+            'leading text\n-----BEGIN CERTIFICATE-----' +
+            goodcert +
+            '-----END CERTIFICATE-----')
+        x509.load_certificate(newcert)
+
+        # Should handle list/tuple
+        x509.load_certificate((newcert,))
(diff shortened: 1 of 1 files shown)


tiran:
NACK

The ^ is correct because the regular expression must search for a line that starts with ```-----BEGIN CERTIFICATE-----```. I cannot reproduce the issue locally. The regexp matches a cert with leading text:

```
>>> import re
>>> regexp = u"^-----BEGIN CERTIFICATE-----(.*?)-----END CERTIFICATE-----"
>>> pem = u"leading line\n-----BEGIN CERTIFICATE-----\nabcd\n-----END CERTIFICATE-----\ntrailing text"
>>> re.search(regexp, pem, re.MULTILINE | re.DOTALL)
<_sre.SRE_Match object at 0x7f667778d0a8>
>>> re.search(regexp, pem, re.MULTILINE | re.DOTALL).group(1)
u'\nabcd\n'
```

HonzaCholasta:
Oops, didn't realize that `^` matches beginning of each line in multiline mode. I think we can keep the test, though.

tiran:
Yes, please keep the test. It should pass with the current regular expression, too.

HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/89dfbab3ca076812590f371c21abcb51b350170b

PR #437, opened 2017-02-06
FIPS: replica install check

tomaskrizek:
PR depends on the rest of the FIPS patches.

### DIFF
--- a/ipalib/config.py
+++ b/ipalib/config.py
@@ -47,6 +47,10 @@ from ipalib.constants import (
     TLS_VERSIONS
 )
 from ipalib import errors
+try:
+    from ipaplatform.tasks import tasks
+except ImportError:
+    tasks = None
 
 if six.PY3:
     unicode = str
@@ -442,6 +446,10 @@ class Env(object):
         self.script = path.abspath(sys.argv[0])
         self.bin = path.dirname(self.script)
         self.home = os.environ.get('HOME', None)
+
+        # Set fips_mode only if ipaplatform module was loaded
+        if tasks is not None:
+            self.fips_mode = tasks.is_fips_enabled()
 
         # Merge in overrides:
         self._merge(**overrides)
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -512,6 +512,13 @@ def promote_openldap_conf(hostname, master):
 
 
 def check_remote_version(api):
+    """
+    Perform a check to verify remote server's version
+
+    :param api: remote API
+
+    :raises: ``ScriptError`` if the checks fails
+    """
     client = rpc.jsonclient(api)
     client.finalize()
 
@@ -524,7 +531,7 @@ def check_remote_version(api):
     remote_version = parse_version(env['version'])
     api_version = parse_version(api.env.version)
     if remote_version > api_version:
-        raise RuntimeError(
+        raise ScriptError(
             "Cannot install replica of a server of higher version ({}) than"
             "the local version ({})".format(remote_version, api_version))
 
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -4,6 +4,7 @@
 
 from __future__ import print_function
 
+import contextlib
 import dns.exception as dnsexception
 import dns.name as dnsname
 import dns.resolver as dnsresolver
@@ -511,29 +512,37 @@ def promote_openldap_conf(hostname, master):
         root_logger.info("Failed to update {}: {}".format(ldap_conf, e))
 
 
-def check_remote_version(api):
+@contextlib.contextmanager
+def rpc_client(api):
     """
-    Perform a check to verify remote server's version
+    Context manager for JSON RPC client.
 
-    :param api: remote API
-
-    :raises: ``ScriptError`` if the checks fails
+    :param api: api to initiate the RPC client
     """
     client = rpc.jsonclient(api)
     client.finalize()
-
     client.connect()
+
     try:
-        env = client.forward(u'env', u'version')['result']
+        yield client
     finally:
         client.disconnect()
 
+
+def check_remote_version(client, local_version):
+    """
+    Verify remote server's version is not higher than this server's version
+
+    :param client: RPC client
+    :param local_version: API version of local server
+    :raises: ScriptError: if the checks fails
+    """
+    env = client.forward(u'env', u'version')['result']
     remote_version = parse_version(env['version'])
-    api_version = parse_version(api.env.version)
-    if remote_version > api_version:
+    if remote_version > local_version:
         raise ScriptError(
             "Cannot install replica of a server of higher version ({}) than"
-            "the local version ({})".format(remote_version, api_version))
+            "the local version ({})".format(remote_version, local_version))
 
 
 def common_check(no_ntp):
@@ -745,6 +754,7 @@ def install_check(installer):
                          ldap_uri=ldapuri)
     remote_api.finalize()
     installer._remote_api = remote_api
+
     conn = remote_api.Backend.ldap2
     replman = None
     try:
@@ -1083,7 +1093,8 @@ def promote_check(installer):
     remote_api.finalize()
     installer._remote_api = remote_api
 
-    check_remote_version(remote_api)
+    with rpc_client(remote_api) as client:
+        check_remote_version(client, api.env.version)
 
     conn = remote_api.Backend.ldap2
     replman = None
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -529,6 +529,27 @@ def rpc_client(api):
         client.disconnect()
 
 
+def check_remote_fips_mode(client, local_fips_mode):
+    """
+    Verify remote server's fips-mode is the same as this server's fips-mode
+
+    :param client: RPC client
+    :param local_fips_mode: boolean indicating whether FIPS mode is turned on
+    :raises: ScriptError: if the checks fails
+    """
+    env = client.forward(u'env', u'fips_mode')['result']
+    remote_fips_mode = env.get('fips_mode', False)
+    if local_fips_mode != remote_fips_mode:
+        if local_fips_mode:
+            raise ScriptError(
+                "Cannot join FIPS-enabled replica into existing topology: "
+                "FIPS is not enabled on the master server.")
+        else:
+            raise ScriptError(
+                "Cannot join replica into existing FIPS-enabled topology: "
+                "FIPS has to be enabled locally first.")
+
+
 def check_remote_version(client, local_version):
     """
     Verify remote server's version is not higher than this server's version
@@ -1095,6 +1116,7 @@ def promote_check(installer):
 
     with rpc_client(remote_api) as client:
         check_remote_version(client, api.env.version)
+        check_remote_fips_mode(client, api.env.fips_mode)
 
     conn = remote_api.Backend.ldap2
     replman = None
--- a/ipatests/test_ipalib/test_config.py
+++ b/ipatests/test_ipalib/test_config.py
@@ -562,6 +562,7 @@ class test_Env(ClassChecker):
 
         # Test using DEFAULT_CONFIG:
         defaults = dict(constants.DEFAULT_CONFIG)
+        defaults['fips_mode'] = object
         (o, home) = self.finalize_core(None, **defaults)
         assert list(o) == sorted(defaults)
         for (key, value) in defaults.items():


ipaserver/install/server/replicainstall.py
```diff
@@ -508,22 +508,44 @@ def promote_openldap_conf(hostname, master):
...
 
+    # Check version compatibility
     remote_version = parse_version(env['version'])
     api_version = parse_version(api.env.version)
     if remote_version > api_version:
-        raise RuntimeError(
-            "Cannot install replica of a server of higher version ({}) than"
-            "the local version ({})".format(remote_version, api_version))
+        raise ScriptError(
+            _("Cannot install replica of a server of higher version "
```
MartinBasti: We don't use ugettext in installers, only in plugins

ipaserver/install/server/replicainstall.py
```diff
@@ -508,22 +508,44 @@ def promote_openldap_conf(hostname, master):
...
+              "(%(remote_version)s) than the local version (%(api_version)s)")
+            % dict(remote_version=remote_version, api_version=api_version))
+
+    # Check FIPS mode compatibility
+    remote_fips_mode = env['fips_mode']
+    fips_mode = tasks.is_fips_enabled()
+    if fips_mode != remote_fips_mode:
+        if fips_mode:
+            raise ScriptError(
+                _("Cannot join FIPS-enabled replica into existing topology: "
```
MartinBasti: Same here

ipaserver/install/server/replicainstall.py
```diff
@@ -508,22 +508,44 @@ def promote_openldap_conf(hostname, master):
...
+                  "FIPS is not enabled on the master server."))
+        else:
+            raise ScriptError(
+                _("Cannot join replica into existing FIPS-enabled topology: "
```
MartinBasti: and here

MartinBasti:
I'm still afraid that users may want to create a FIPS replica from the non-FIPS master, even if it is not recommended due security. How can be this achieved?

tomaskrizek:
@MartinBasti Since this check is performed only during installation, the user could simply install non-FIPS replica and then turn FIPS on afterwards. There might be issues with this approach and thus it is neither recommended nor supported, as stated in the [documentation](https://www.freeipa.org/page/V4/FreeIPA-on-FIPS#Multiple_servers_in_topology).

MartinBasti:
@tomaskrizek on current versions of RHEL and fedora IPA doesn't start in FIPS, but upgrading first and then enabling FIPS might be the way

ipalib/config.py
```diff
@@ -44,6 +44,7 @@
 from ipalib.constants import CONFIG_SECTION
 from ipalib.constants import OVERRIDE_ERROR, SET_ERROR, DEL_ERROR
 from ipalib import errors
+from ipaplatform.tasks import tasks
```
HonzaCholasta: Importing from `ipaplatform` to `ipalib` is forbidden, this will have to be worked around.

ipaserver/install/server/replicainstall.py
```diff
@@ -508,22 +508,44 @@ def promote_openldap_conf(hostname, master):
...
     remote_version = parse_version(env['version'])
     api_version = parse_version(api.env.version)
     if remote_version > api_version:
-        raise RuntimeError(
-            "Cannot install replica of a server of higher version ({}) than"
-            "the local version ({})".format(remote_version, api_version))
+        raise ScriptError(
+            "Cannot install replica of a server of higher version "
+            "(%(remote_version)s) than the local version (%(api_version)s)"
+            % dict(remote_version=remote_version, api_version=api_version))
```
HonzaCholasta: NACK, `.format()` is the preferred way of formatting strings.

ipaserver/install/server/replicainstall.py
```diff
@@ -1077,7 +1099,7 @@ def promote_check(installer):
     remote_api.finalize()
     installer._remote_api = remote_api
 
-    check_remote_version(remote_api)
+    check_remote_compatibility(remote_api)
```
HonzaCholasta: This check is executed only in replica promotion. There should be a check for classic replica install as well; IMO it should reject install when FIPS is enabled.

ipalib/config.py
```diff
@@ -497,6 +498,10 @@ def _bootstrap(self, **overrides):
         if 'plugins_on_demand' not in self:
             self.plugins_on_demand = (self.context == 'cli')
 
+        # Set fips_mode:
+        if 'fips_mode' not in self:
+            self.fips_mode = tasks.is_fips_enabled()
```
stlaz: Do we want the user to be able to set env.fips_mode during api bootstrapping? I do not think so. This should be set amongst other run-time variables.

ipaserver/install/server/replicainstall.py
```diff
@@ -508,22 +508,44 @@ def promote_openldap_conf(hostname, master):
...
-            "Cannot install replica of a server of higher version ({}) than"
-            "the local version ({})".format(remote_version, api_version))
+        raise ScriptError(
+            "Cannot install replica of a server of higher version "
+            "(%(remote_version)s) than the local version (%(api_version)s)"
+            % dict(remote_version=remote_version, api_version=api_version))
+
+    # Check FIPS mode compatibility
+    remote_fips_mode = env['fips_mode']
+    fips_mode = tasks.is_fips_enabled()
```
stlaz: I think you could rather pass your api.env.fips_mode to `check_remote_compatibility` instead.

ipaserver/install/server/replicainstall.py
```diff
@@ -508,22 +508,44 @@ def promote_openldap_conf(hostname, master):
...
-        raise RuntimeError(
-            "Cannot install replica of a server of higher version ({}) than"
-            "the local version ({})".format(remote_version, api_version))
+        raise ScriptError(
+            "Cannot install replica of a server of higher version "
+            "(%(remote_version)s) than the local version (%(api_version)s)"
+            % dict(remote_version=remote_version, api_version=api_version))
+
+    # Check FIPS mode compatibility
+    remote_fips_mode = env['fips_mode']
```
stlaz: `env` does not necessarily contain "fips_mode" key.

ipaserver/install/server/replicainstall.py
stlaz: Yes, that's what happens now with the `is_fips_enabled()` guards. Since FIPS is not available on DL0, it should fail.

martbab:
@tomaskrizek since you added a new key to the Env object, you will have to fix `test_ipalib/test_config.py` to account for this change, see https://travis-ci.org/freeipa/freeipa/jobs/198916106#L443

pvoborni:
@MartinBasti I'm not sure from your comment if you would like to provide a way to change non-FIPS server into a FIPS server or just brainstorming ways how it can be worked around. In any case this path is not a goal and actually should be discouraged. http://www.freeipa.org/page/V4/FreeIPA-on-FIPS#Design

MartinBasti:
@pvoborni more or less brainstorming, as I'm almost sure that people will want to migrate current deployments to FIPS mode

tomaskrizek:
Thanks for the feedback. Hopefully I addressed all the concerns above in the update.

ipalib/config.py
```diff
@@ -440,6 +444,9 @@ def _bootstrap(self, **overrides):
         self.bin = path.dirname(self.script)
         self.home = os.environ.get('HOME', None)
 
+        if tasks is not None:
+            self.fips_mode = tasks.is_fips_enabled()
```
stlaz: This needs an explanation in a comment.

ipaserver/install/server/replicainstall.py
```diff
@@ -508,22 +509,60 @@ def promote_openldap_conf(hostname, master):
...
+    Verify remote server's fips-mode is the same as this server's fips-mode
+
+    :param client: RPC client
+    :param local_fips_mode: boolean indicating whether FIPS mode is turned on
+
+    :raises: ``ScriptError`` if the checks fails
+    """
+    env = client.forward(u'env', u'fips_mode')['result']
+    remote_fips_mode = env.get('fips_mode', False)
+    fips_mode = tasks.is_fips_enabled()
```
stlaz: I believe you left some mess here.

ipaserver/install/server/replicainstall.py
```diff
@@ -508,22 +509,60 @@ def promote_openldap_conf(hostname, master):
...
+                "Cannot join replica into existing FIPS-enabled topology: "
+                "FIPS has to be enabled locally first.")
+
+
+def check_remote_version(client, local_version):
+    """
+    Verify remote server's version is not higher than this server's version
+
+    :param client: RPC client
+
```
stlaz: Please fill this empty line with the other parameter documentation. Don't leave a blank line before `:raises`, it's ugly and it's not usually done this way [pythonhosted:Sphinx](https://pythonhosted.org/an_example_pypi_project/sphinx.html).

ipaserver/install/server/replicainstall.py
```diff
@@ -508,22 +509,60 @@ def promote_openldap_conf(hostname, master):
...
+    :raises: ``ScriptError`` if the checks fails
```
stlaz: Should be `:raises: ScripError: blahblah`

ipaserver/install/server/replicainstall.py
```diff
@@ -747,6 +786,10 @@ def install_check(installer):
                          ldap_uri=ldapuri)
     remote_api.finalize()
     installer._remote_api = remote_api
+
+    with rpc_client(remote_api) as client:
+        check_remote_fips_mode(client, api.env.fips_mode)
```
stlaz: Why are you checking fips mode twice in DL1? Also, if this is supposed to address the issue "Explode if trying to install in FIPS on DL0", this won't work.

stlaz: The DL1 twice checking remark is false, thought this code was in `common_check()`. The other part still applies, though.

stlaz:
LGTM

ipalib/constants.py
```diff
@@ -215,6 +215,7 @@
     ('script', object),  # sys.argv[0]
     ('bin', object),  # The directory containing the script
     ('home', object),  # $HOME
+    ('fips_mode', object),  # FIPS-mode of the server
```
HonzaCholasta: Won't this cause `api.env.fips_mode` to be `object` rather than unset when `ipaplatform` is not installed?

tomaskrizek: You're right. I didn't notice these settings are global, not just test-specific. I removed this bit and fixed the test instead.

ipaserver/install/server/replicainstall.py
```diff
@@ -508,22 +509,58 @@ def promote_openldap_conf(hostname, master):
...
+    :param local_version: API version of local server
+    :raises: ScriptError: if the checks fails
+    """
+    env = client.forward(u'env', u'version')['result']
     remote_version = parse_version(env['version'])
-    api_version = parse_version(api.env.version)
-    if remote_version > api_version:
-        raise RuntimeError(
+    if remote_version > local_version:
+        raise ScriptError(
```
HonzaCholasta: Note that `ScriptError` is really only necessary when you need the script to fail with a specific exit code. If you don't need that and are fine with exit code 1, `RuntimeError` is all you need.

ipatests/test_ipalib/test_config.py
```diff
@@ -563,7 +563,7 @@ def test_finalize_core(self):
         # Test using DEFAULT_CONFIG:
         defaults = dict(constants.DEFAULT_CONFIG)
         (o, home) = self.finalize_core(None, **defaults)
-        assert list(o) == sorted(defaults)
+        assert set(o).issuperset(set(defaults))
```
HonzaCholasta: I would rather keep the original check, but make sure that `o` does not contain `fips_mode` before doing it.

tomaskrizek: I'm not sure how you'd like to make sure `o` doesn't contain `fips_mode`. `Env` object does not allow deletion and the only way I can think of would be to violate the object's "private" properties.

I propose we either:

- always set fips_mode: to `True`, `False` or `None`
- accept that `Env` can now have optional attributes and use the test proposed above

martbab: You can just add the `fips_mode` into ipalib/constants and set its value to `object`. We had to do this workaround when @tiran added `nss_dir` to the config (see https://github.com/freeipa/freeipa/blob/master/ipalib/constants.py#L228). The test in the original from will then just work(R).

tomaskrizek: @martbab I tried to do that, but that also sets the `fips_mode` key for the env object, doesn't it? We want the `fips_mode` key to not be present if ipaplatform wasn't imported.

HonzaCholasta: I think this should work:
```python
    defaults = dict(constants.DEFAULT_CONFIG)
    defaults['fips_mode'] = None
    (o, home) = self.finalize_core(None, **defaults)
    assert list(o) == sorted(defaults)
```

tomaskrizek: Ah, of course! fixed

ipaserver/install/server/replicainstall.py
```diff
@@ -738,6 +775,7 @@ def install_check(installer):
                          ldap_uri=ldapuri)
     remote_api.finalize()
     installer._remote_api = remote_api
+
```
HonzaCholasta: Why was the "FIPS is not allowed in domain level 0" check removed?

stlaz: It was not removed but rather moved to `install_check()` in https://github.com/freeipa/freeipa/pull/445/files as this is a DL-specific check from now on.

stlaz: It is still there: https://github.com/freeipa/freeipa/pull/437/files#diff-4fc382c22d987062a45dddeaac8bb461L648

tomaskrizek: https://github.com/tomaskrizek/freeipa/blob/28c57e48423e77756eb152068db44933cd0540f7/ipaserver/install/server/replicainstall.py#L685

ipatests/test_ipalib/test_config.py
```diff
@@ -562,6 +562,7 @@ def test_finalize_core(self):
 
         # Test using DEFAULT_CONFIG:
         defaults = dict(constants.DEFAULT_CONFIG)
+        defaults['fips_mode'] = object
```
HonzaCholasta: `object`. Huh?

HonzaCholasta:
LGTM.

ipatests/test_ipalib/test_config.py
tomaskrizek: Yes, to skip the value check (few lines below - https://github.com/freeipa/freeipa/pull/437/files/28c57e48423e77756eb152068db44933cd0540f7#diff-c5fcc8900faeb561ad5f2a58fef94c85L568 ). If we set it to `None`, the test will check whether `fips_mode` value is `None` - which is not the desired behavior.

stlaz:
3 LGTM + tests passing seems like a good enough reason for ACK to me.

MartinBasti:
Needs rebase

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/3372ad2766c0d182fa88c8bc28cf43477dc4cb3b
https://fedorahosted.org/freeipa/changeset/7292890042677ae40faa44753ebf570db6c19e7c
https://fedorahosted.org/freeipa/changeset/62e884ff7f037a28a15d61cc9fa9c46e5c40cda5
https://fedorahosted.org/freeipa/changeset/397ca71e897b42a23ed4ef294fca367c1542a2aa
https://fedorahosted.org/freeipa/changeset/cf25ea7e300cdada57bd964acb4393cc11ad333e

PR #438, opened 2017-02-07
ipaldap: preserve order of values in LDAPEntry._sync()

HonzaCholasta:
In Python 2, the order was preserved by accident.

This change makes sure the order is preserved in both Python 2 and 3.


MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/e920ae22525d34e1f524e2e59159ac50c603bc8c

PR #441, opened 2017-02-08
Print test env information

tiran:
Print api.env, uname, euid/egid, cwd and Python version when tests are
run with -v (e.g. ipa-run-tests -v).


MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/b20f6fb29478de6b4f25741bc4fd975a5e0be671

PR #442, opened 2017-02-08
Add option to run tests in-tree and out-of-tree mode

tiran:
By default ipa-run-tests and pytest auto-detect the presence of
../ipasetup.py.in and run tests in-tree mode when the file exists. The
option can be overriden with ipa-run-tests --in-tree=true/false.


tiran:
Not useful or relevant any more.

PR #443, opened 2017-02-08
Stronger check for DM password during server install

stlaz:
DM password is used as an NSS database password during server installation, therefore it must comply to NSS database requirements for password in FIPS.

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -54,6 +54,22 @@ NoneType = type(None)
 SYSRESTORE_DIR_PATH = paths.SYSRESTORE
 
 
+def check_password_fips_nssdb_compatible(password):
+    """
+    Check whether the given password can be used for NSSDB setup in FIPS mode
+    """
+    gotnumeric = any(c.isdigit() for c in password[:-1])
+    gotupper = any(c.isupper() for c in password[1:])
+    gotlower = any(c.islower() for c in password)
+    gotspecial = not password.isalnum()
+    classes = sum([gotnumeric, gotupper, gotlower, gotspecial])
+    if classes < 3:
+        raise ValueError("Password must contain at least one character "
+                         "from each of three out of these four character "
+                         "classes: numeric, uppercase letters, lowercase "
+                         "letters and special symbols.")
+
+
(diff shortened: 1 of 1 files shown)


pvoborni:
Function check_password_fips_nssdb_compatible  looks like a great candidate for unit test.

ipaserver/install/server/install.py
```diff
@@ -54,6 +54,22 @@
...
 
 
+def check_password_fips_nssdb_compatible(password):
+    """
+    Check whether the given password can be used for NSSDB setup in FIPS mode
+    """
+    gotnumeric = any(c.isdigit() for c in password[:-1])
+    gotupper = any(c.isupper() for c in password[1:])
+    gotlower = any(c.islower() for c in password)
+    gotspecial = password.isalnum()
```
pvoborni: This line should be negated.

stlaz: Thanks, was editing this right before pushing but forgot how to logic :)

ipaserver/install/server/install.py
```diff
@@ -54,6 +54,22 @@
 SYSRESTORE_DIR_PATH = paths.SYSRESTORE
 
 
+def check_password_fips_nssdb_compatible(password):
+    """
+    Check whether the given password can be used for NSSDB setup in FIPS mode
+    """
+    gotnumeric = any(c.isdigit() for c in password[:-1])
```
tomaskrizek: Perhaps we could use a comment that the `:-1` and `1:` is a NSSDB-specific requirement.

HonzaCholasta:
IMHO you got it backwards - DM password may not comply to NSS requirements for passwords, therefore it must not be used as a password for any NSS database during server install.

stlaz:
@HonzaCholasta: +1, you're right, I should investigate more on how to change this behavior, either we or Dogtag don't behave correctly here.
@pvoborni, @tomaskrizek: out of curiousity, do we have a design/guideline on how to write unit tests for FreeIPA? Did not find any.

stlaz:
Closing as REJECTED, this will be sorted out in another way.

PR #446, opened 2017-02-08
Add password file to certutil calls in ipapython.certdb module

stlaz:
With this patchset, ipa-client-install should not ask for NSS database password.

Prerequisite for it to work in FIPS:
https://github.com/freeipa/freeipa/pull/367

**edit:** This was a part of a bigger branch and might be missing some parts.

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -2284,18 +2284,8 @@ def install_check(options):
 
 def create_ipa_nssdb():
     db = certdb.NSSDatabase(paths.IPA_NSSDB_DIR)
-    pwdfile = os.path.join(db.secdir, 'pwdfile.txt')
-
-    ipautil.backup_file(pwdfile)
-    ipautil.backup_file(os.path.join(db.secdir, 'cert8.db'))
-    ipautil.backup_file(os.path.join(db.secdir, 'key3.db'))
-    ipautil.backup_file(os.path.join(db.secdir, 'secmod.db'))
-
-    with open(pwdfile, 'w') as f:
-        f.write(ipautil.ipa_generate_password())
-    os.chmod(pwdfile, 0o600)
-
-    db.create_db(pwdfile)
+    db.create_db(backup=True)
+    os.chmod(db.pwd_file, 0o600)
     os.chmod(os.path.join(db.secdir, 'cert8.db'), 0o644)
     os.chmod(os.path.join(db.secdir, 'key3.db'), 0o644)
     os.chmod(os.path.join(db.secdir, 'secmod.db'), 0o644)
@@ -2667,8 +2657,7 @@ def _install(options):
             for cert in ca_certs
         ]
         try:
-            pwd_file = ipautil.write_tmp_file(ipautil.ipa_generate_password())
-            tmp_db.create_db(pwd_file.name)
+            tmp_db.create_db()
 
             for i, cert in enumerate(ca_certs):
                 tmp_db.add_cert(cert, 'CA certificate %d' % (i + 1), 'C,,')
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -17,7 +17,6 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-import binascii
 import os
 import io
 import pwd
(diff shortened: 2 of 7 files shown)


ipapython/certdb.py
```diff
@@ -83,7 +83,8 @@ class NSSDatabase(object):
     # got too tied to IPA server details, killing reusability.
     # BaseCertDB is a class that knows nothing about IPA.
     # Generic NSS DB code should be moved here.
-    def __init__(self, nssdir=None):
+    def __init__(self, nssdir=None, password_filename=None):
+        self.password_filename = password_filename
```
HonzaCholasta: Could we default to `os.path.join(nssdir, 'pwdfile.txt')`?

ipapython/certdb.py
```diff
@@ -104,14 +105,18 @@ def __exit__(self, type, value, tb):
...
         return ipautil.run(new_args, stdin, **kwargs)
 
     def create_db(self, password_filename):
         """Create cert DB
 
         :param password_filename: Name of file containing the database password
         """
-        self.run_certutil(["-N", "-f", password_filename])
+        # run_certutil will use self.password_filename to setup the db
+        self.password_filename = password_filename
```
HonzaCholasta: Now that the filename is specified via `__init__()` argument, I see very little point in keeping the argument in `create_db()` as well.

ipaserver/install/custodiainstance.py
```diff
@@ -159,7 +159,7 @@ def __get_keys(self, ca_host, cacerts_file, cacerts_pwd, data):
                              '-w', pk12pwfile])
 
             # Add CA certificates
-            tmpdb = CertDB(self.realm, nssdir=tmpnssdir)
+            tmpdb = NSSDatabase(tmpnssdir, password_filename=nsspwfile)
```
HonzaCholasta: Why is this necessary?

stlaz: The path to CertDB password file in this case would be $tmpnssdir"/pwdfile.txt". This file would be then passed to CertDB's inner NSSDatabase - but there is no such file. As the author of custodiainstance creates the password file in $tmpnssdir"/pk12pwfile" it was my assumption that they never meant to use IPA-specific NSS database representation.

ipapython/certdb.py
stlaz: We might but password file does not need to exist in general. I think this assumption is made in CertDB instead.
I can add it nonetheless.

ipapython/certdb.py
stlaz: Ok

stlaz: As I started doing this I realized that this change actually has to be done along with the one proposed on line 87, will do both, then.

ipaclient/install/client.py
```diff
@@ -2288,26 +2288,26 @@ def install_check(options):
 
 
 def create_ipa_nssdb():
-    db = certdb.NSSDatabase(paths.IPA_NSSDB_DIR)
-    pwdfile = os.path.join(db.secdir, 'pwdfile.txt')
+    db = certdb.NSSDatabase(paths.IPA_NSSDB_DIR,
+                            os.path.join(paths.IPA_NSSDB_DIR, 'pwdfile.txt'))
```
stlaz: I could do this change properly by replacing `NSSDatabase` with `CertDB` but that would mean having `certs.py` moved to `ipalib/install` so that it's in a reasonable spot which I am not sure is worth it.

stlaz:
NSSDatabase now defaults its `.password_file` to `.sec_dir + 'passwd.txt'`. It's necessary to create a pwdfile.txt in system-wide cert store so that actions like CA renew function properly even with FIPS.

ipaserver/install/server/upgrade.py
```diff
@@ -1521,6 +1521,11 @@ def upgrade_configuration():
             api.env.realm, certs.NSS_DIR, host_name=api.env.host)
     ca_running = ca.is_running()
 
+    # create passswd.txt file in PKI_TOMCAT_ALIAS_DIR if it does not exist
+    # this file will be required on most actions over this NSS DB in FIPS
+    if not os.path.exists(paths.PKI_TOMCAT_ALIAS_DIR, 'passwd.txt'):
```
HonzaCholasta: It's `pwdfile.txt`.

stlaz: Seems like I got somehow obsessed with `passwd.txt` for some reason. Fixed.

stlaz:
This patchset seems more like a cleanup after the privilege separation one, although adding a password to certutil calls is still the main topic here.

ipaserver/install/server/upgrade.py
```diff
@@ -1543,6 +1543,11 @@ def upgrade_configuration():
             api.env.realm, paths.IPA_RADB_DIR, host_name=api.env.host)
     ca_running = ca.is_running()
 
+    # create passswd.txt file in PKI_TOMCAT_ALIAS_DIR if it does not exist
+    # this file will be required on most actions over this NSS DB in FIPS
+    if not os.path.exists(paths.PKI_TOMCAT_ALIAS_DIR, 'pwdfile.txt'):
```
stlaz: That was the original intention, thanks.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/ca457eb5ce12291f555f1bf771114d6d7d191987
https://fedorahosted.org/freeipa/changeset/b20b0489ea06931bfa7d46bdbd6623bc3f09219b

PR #448, opened 2017-02-08
Tests: Basic coverage with tree root domain

gkaihorodova:
Tests: Basic coverage with tree root domain
Extend existing legacy client tests to cover test cases with tree root domain.
https://fedorahosted.org/freeipa/ticket/6489

### DIFF
--- a/ipatests/test_integration/test_legacy_clients.py
+++ b/ipatests/test_integration/test_legacy_clients.py
@@ -58,6 +58,8 @@ class BaseTestLegacyClient(object):
     testuser_gid_regex = None
     subdomain_testuser_uid_regex = None
     subdomain_testuser_gid_regex = None
+    treedomain_testuser_uid_regex = None
+    treedomain_testuser_gid_regex = None
 
     # To allow custom validation dependent on the trust type
     posix_trust = False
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_legacy_clients.py
```diff
@@ -326,6 +328,91 @@ def test_login_disabled_subdomain_ad_user(self):
...
 
+    def test_getent_treedomain_ad_user(self):
+        if not self.ad_treedomain:
+            raise nose.SkipTest('AD tree root domain is not available.')
+
+        self.clear_sssd_caches()
+        testuser = 'treetestuser@{0}'.format(self.ad_treedomain)
+        result = self.legacy_client.run_command(['getent', 'passwd', testuser])
+
+        testuser_regex = "treetestuser@%s:\*:%s:%s:"\
```
martbab: Backslashes in multiline strings are discouraged by PEP-8, please enclose the whole string literal in parentheses and leave out backslashes.

Also please try to use new-style string formatting in new code.

martbab:
I have quickly skimmed through code and have one comment. Also, I have noticed the extreme code triplication of the test cases. I think that this warrants some refactoring first before adding tree-root domain tests.

ipatests/test_integration/test_legacy_clients.py
gkaihorodova: Thank you for review. I'll try to rewrite that part using new-style string formatting.

gkaihorodova:
Can you be a little bit more specific about "triplication of the test cases ", please. 
Because, to be honest, I'm having hard time trying to navigate myself there.

martbab:
If you look at the test cases (e.g. test_login_ipa-user, test_login_ad_user, test_login_subdomain_user are the 'best: examples) you can see that the function body is the same code copy-pasted with slight alterations so that it works for the new case. Your patch adds a *fourth* level of copy-pasta to the code, which is something that grieves me greatly.

Clearly, you can group the common code into a private method that can be only called with the use-case specific parameters for each test case. Or you can expand the existing mixing hierarchy to achieve this. Then it would also be simpler to extend the test cases for tree-root domains.

gkaihorodova:
Thank you for explanation and tips. I noticed it as well and I agree that it (and not only that) worth refactoring. Yes, my PR is more or less copy-paste, because I was following existing pattern in the code.
Also I think PR can be pushed and at the same time feel free to open ticket/request or whatever is more suitable for refactoring  task and signed it to me. 
Please, don t be grieve, it makes me sad

martbab:
Well you still have some issues to fix, notably the failing Travis CI and the not-so nice multiline-string literal.

gkaihorodova:
Yes, sure I'll work on these issues

gkaihorodova:
Bump for review

martbab:
The patch looks ok, let's hope that our CI will play nice with it.

gkaihorodova:
Thanks you for review. Let's hope for the best .

martbab:
master:

* 10494b1bb34b6ff9c1b810cc0739c761b017202c Tests: Basic coverage with tree root domain

PR #449, opened 2017-02-08
Travis CI: Upload the logs from failed jobs to transfer.sh

martbab:
When a non-lint job fails, all the relevant logs from the test runner
will be gzipped and uploaded to https://transfer.sh file sharing
service. The download link will then be displayed at the very end of the
Travis build log.

You can see the output of a failed job here:
https://travis-ci.org/martbab/freeipa/jobs/199647801

Just go all the way to the log bottom and expand the last statement. I have not
found any other way to make the link more visible unfortunately.


HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/91341f4035e0d78b0adbe9a09ba69e1fd35ec26d

PR #450, opened 2017-02-08
Add FIPS-token password of HTTPD NSS database

stlaz:
This change is required for httpd to function properly in FIPS

### DIFF
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -329,14 +329,19 @@ class HTTPInstance(service.Service):
         This is the format of mod_nss pin files.
         """
         pwd_conf = paths.HTTPD_PASSWORD_CONF
-
         ipautil.backup_file(pwd_conf)
-        f = open(pwd_conf, "w")
-        f.write("internal:")
-        pwdfile = open(os.path.join(paths.HTTPD_ALIAS_DIR, 'pwdfile.txt'))
-        f.write(pwdfile.read())
-        f.close()
-        pwdfile.close()
+
+        passwd_fname = os.path.join(paths.HTTPD_ALIAS_DIR, 'pwdfile.txt')
+        with open(passwd_fname, 'r') as pwdfile:
+            password = pwdfile.read()
+
+        with open(pwd_conf, "w") as f:
+            f.write("internal:")
+            f.write(password)
+            f.write("\nNSS FIPS 140-2 Certificate DB:")
+            f.write(password)
+            # make sure other processes can access the file contents ASAP
+            f.flush()
         pent = pwd.getpwnam(constants.HTTPD_USER)
         os.chown(pwd_conf, pent.pw_uid, pent.pw_gid)
         os.chmod(pwd_conf, 0o400)


ipaserver/install/certs.py
```diff
@@ -504,7 +504,10 @@ def create_password_conf(self):
         f = open(self.pwd_conf, "w")
         f.write("internal:")
         pwdfile = open(self.passwd_fname)
-        f.write(pwdfile.read())
```
MartinBasti: If you are touching this, you can change thsi code to use `with` statement

ipaserver/install/certs.py
```diff
@@ -501,13 +501,19 @@ def create_password_conf(self):
...
+            password = pwdfile.read()
+
+        with open(self.pwd_conf, "w") as f:
+            f.write("internal:")
+            f.write(password)
+            f.write("\nNSS FIPS 140-2 Certificate DB:")
+            f.write(password)
+            # make sure other processes can access the file contents ASAP
+            f.flush()
+        if os.path.exists(self.pwd_conf):
```
MartinBasti: Why do you need to check presence of file that is actually created a few lines above?

stlaz: Sorry, I was doing something else and left some mess there.

rcritten:
I guess this is one approach to fix the problem. Would it be cleaner to pass in, or detect, FIPS mode, and only write out the token that will actually be used?

stlaz:
That was my original approach to it but we had offline talk with @HonzaCholasta and got to the point that it might be better to do it this way.
From my point of view it's more fool-proof for the people who would install FreeIPA in non-FIPS mode but then thought it'd be cool to turn FIPS on. Anyone reading this in the future - that is **NOT SUPPORTED**. There would probably be more different issues, let this not be one.

ipaserver/install/certs.py
```diff
@@ -501,12 +501,17 @@ def create_password_conf(self):
...
-        pwdfile.close()
+
+        with open(self.passwd_fname) as pwdfile:
+            password = pwdfile.read()
+
+        with open(self.pwd_conf, "w") as f:
+            f.write("internal:")
+            f.write(password)
+            f.write("\nNSS FIPS 140-2 Certificate DB:")
+            f.write(password)
```
HonzaCholasta: No `\n` here?

HonzaCholasta:
LGTM. I guess we don't have to bother with upgrade, given that you can turn on FIPS post-install, right?

ipaserver/install/certs.py
stlaz: I did not run into any issues with it not being there. It was not there when there was only the `internal` token so I assume there's no need for it here as well.

stlaz:
You shouldn't turn FIPS on post-install (is what I think you mean), correct.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/0b9b6b52d7f2e64a52ef8fd570839711311fa254

PR #451, opened 2017-02-08
certdb: remove unused keysize property

tomaskrizek:
Keysize property is no longer used anywhere in the code. It was
originally introduced for the request_cert function, which was later
refactored to use a function argument instead.

---

The value of this property caught my eye, because I don't think we should be using 1024bit keys. Fortunately, I discovered this bit of code is obsolete and we actually use 2048bit key length by default.

Commit that originally introduced the property: 158b4e8ff4704b967d4049e2a16f9b32fbb33b80
Commit that removed the usage of the property: 9182c10b03a7841c9318ad64ae6c5deda77d93d1

### DIFF
--- a/ipaserver/install/certs.py
+++ b/ipaserver/install/certs.py
@@ -109,7 +109,6 @@ class CertDB(object):
             raise RuntimeError("Unable to determine the current directory: %s" % str(e))
 
         self.cacert_name = get_ca_nickname(self.realm)
-        self.valid_months = "120"
 
         # We are going to set the owner of all of the cert
         # files to the owner of the containing directory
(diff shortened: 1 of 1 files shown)


stlaz:
If you want to remove them, you may want to check for other properties as well (I see `valid_months` at least).

tomaskrizek:
Commit that removed valid_moths property: e736e75ce9724ae8298a5b69d093313cd6e62b60

frasertweedale:
Conditional ACK: just fix the type `s/moths/months/` in the commit message.

tomaskrizek:
@frasertweedale Fixed.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/47565c0fc75721f457e87b1c3e3325fff6a3b3ae
https://fedorahosted.org/freeipa/changeset/36f46a5301ce62b5549899e5d693fca0b88946fb

PR #452, opened 2017-02-08
[ WIP] ipa-run-tests: allow to run tests with server-api

MartinBasti:
This allow to test server-api with ipa-run-tests. It is useful because
internal error tracebacks are printed to test output and it is handy to
use it with python -bb option to check BytesWarnings

- I havent tested option parsing in pytest yet, only the code around that allows to run server_api
- This can be useful with Travis to see tracebacks directly in test output
- tests may be faster
- we should really rename test_xmlrpc to something like test_api
- I will use this for python -bb testing

### DIFF
--- a/ipatests/conftest.py
+++ b/ipatests/conftest.py
@@ -71,9 +71,31 @@ def pytest_configure(config):
     config.option.doctestmodules = True
 
 
+def pytest_addoption(parser):
+    def truefalse(arg):
+        if arg.lower() == 'true':
+            return True
+        if arg.lower() == 'false':
+            return False
+        return arg  # triggers an error later
+
+    group = parser.getgroup("IPA integration tests")
+    group.addoption(
+        '--in-server',
+        dest="ipa_in_server",
+        type=truefalse,
+        choices=(True, False),
+        default=False,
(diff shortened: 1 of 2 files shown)


ipatests/conftest.py
```diff
@@ -71,9 +71,31 @@ def pytest_configure(config):
...
+            return False
+        return arg  # triggers an error later
+
+    group = parser.getgroup("IPA integration tests")
+    group.addoption(
+        '--in-server',
+        dest="ipa_in_server",
+        type=truefalse,
+        choices=(True, False),
+        default=False,
```
tiran: You don't need to go the complicated route like in PR #442. Please replace type, choices and default with a simple ```action='store_true```, https://docs.python.org/3/library/argparse.html#action

MartinBasti: sure, but there are more issues I have to solve than this one :) still WIP

MartinBasti:
There are too many issues, it is not possible to run tests with server api easily, too many changes required.

tiran:
It's going to get simpler when privilege separation patch has landet.

PR #453, opened 2017-02-09
Cleanup certdb

tiran:
* use with statement to open/close files
* prefer fchmod/fchown when a file descriptor is available
* set permission before data is written to file
* remove chdir() hack with proper cwd argument to ipautil.run()

Do not ever change the working directory of a program. It's a really bad
idea. Just consider what is going to happen if two threads or two
different parts of a process decide to own control over the working
directory?

### DIFF
--- a/ipaserver/install/certs.py
+++ b/ipaserver/install/certs.py
@@ -117,10 +117,12 @@ class CertDB(object):
         self.host_name = host_name
         self.ca_subject = ca_subject
         self.subject_base = subject_base
+
         try:
-            self.cwd = os.getcwd()
+            self.cwd = os.path.abspath(os.getcwd())
         except OSError as e:
-            raise RuntimeError("Unable to determine the current directory: %s" % str(e))
+            raise RuntimeError(
+                "Unable to determine the current directory: %s" % str(e))
 
         self.cacert_name = get_ca_nickname(self.realm)
 
@@ -163,6 +165,8 @@ class CertDB(object):
     def __del__(self):
         if self.reqdir is not None:
             shutil.rmtree(self.reqdir, ignore_errors=True)
+            self.reqdir = None
+        self.nssdb.close()
         try:
             os.chdir(self.cwd)
         except OSError:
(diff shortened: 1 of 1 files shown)


rcritten:
I'm pretty sure the chdir() hack was due to SELinux issues, be sure to test in enforcing mode. It may no longer be required.

tiran:
Thx Rob, I use ```ipautil.run(cwd=...)``` to change the working directory just for the subprocess instead of the entire parent process.

tiran:
@stlaz You did most work with NSS and certdb recently. Can you have a look at this collection of fixes. Iis it useful for you or do you plan to rip out the module soonish? Either way please feel free to merge or close this PR.

stlaz:
@tiran Thanks for reminding me. I was waiting for some of my fixes to get pushed as well, I will go through your PR first thing tomorrow.

ipaserver/install/certs.py
```diff
@@ -162,48 +156,45 @@ def setup_cert_request(self):
...
-            pent = pwd.getpwnam(uid)
-            os.chown(fname, pent.pw_uid, pent.pw_gid)
-        else:
-            os.chown(fname, self.uid, self.gid)
+    def set_perms(self, fname, write=False):
         perms = stat.S_IRUSR
         if write:
             perms |= stat.S_IWUSR
-        os.chmod(fname, perms)
+        if hasattr(fname, 'fileno'):
```
stlaz: Scratch that, didn't realize you are passing files to it as well.

tiran: Yes, it's both more efficient and more secure to operate on FDs.

ipaserver/install/certs.py
```diff
@@ -142,10 +138,8 @@ def passwd_fname(self):
     def __del__(self):
         if self.reqdir is not None:
             shutil.rmtree(self.reqdir, ignore_errors=True)
-        try:
-            os.chdir(self.cwd)
-        except OSError:
-            pass
+            self.reqdir = None
+        self.nssdb.close()
```
tiran: Personally I find it cleaner to always call close. You never know of NSSDatabase is going to need close() for other reasons.

But I'm not going to fight you on this.

stlaz: Alright, you can keep it there, I don't mind.

ipaserver/install/certs.py
```diff
@@ -360,14 +360,14 @@ def create_server_cert(self, nickname, hostname, subject=None):
...
-        self.issue_server_cert(self.certreq_fname, self.certder_fname)
-        self.import_cert(self.certder_fname, nickname)
-        fd = open(self.certder_fname, "r")
-        dercert = fd.read()
-        fd.close()
-
-        os.unlink(self.certreq_fname)
-        os.unlink(self.certder_fname)
+        try:
+            cdb.issue_server_cert(self.certreq_fname, self.certder_fname)
```
stlaz: The behavior here changed a bit, there's no need to ask another database to issue server cert for us since the location of Dogtag agent certificates is know.
Do `s/cdb/self/`.

ipaserver/install/certs.py
```diff
@@ -183,22 +187,20 @@ def setup_cert_request(self):
         self.certreq_fname = self.reqdir + "/tmpcertreq"
         self.certder_fname = self.reqdir + "/tmpcert.der"
 
-        # When certutil makes a request it creates a file in the cwd, make
-        # sure we are in a unique place when this happens
-        os.chdir(self.reqdir)
```
stlaz: This changes original functionality, please, add these lines back.

stlaz:
The changes are fine. Please, squash the two commits together so that we can push it.

MartinBasti:
master:

* 22d7492c94837342a559c368454c223f566490ac Cleanup certdb

PR #457, opened 2017-02-10
adtrustinstance: use LDAPI/EXTERNAL to retrieve CIFS keytab

martbab:
In order to be able to function as a part of composite installer, samba
configuration code must be able to work without admin credentials. This
requires changes in the CIFS principal key retrieval method so that it is not
bound to the presence of privileged user ccache. This is achieved by slightly
altering and re-using the recently developed code for service keytab retrieval.

https://fedorahosted.org/freeipa/ticket/6638

### DIFF
--- a/ipaserver/install/adtrustinstance.py
+++ b/ipaserver/install/adtrustinstance.py
@@ -530,27 +530,18 @@ class ADTRUSTInstance(service.Service):
             api.Backend.ldap2, self.smb_dn, "member",
             [self.cifs_agent, self.host_princ])
 
-    def __setup_principal(self):
-        try:
-            api.Command.service_add(unicode(self.principal))
-        except errors.DuplicateEntry:
-            # CIFS principal already exists, it is not the first time
-            # adtrustinstance is managed
-            # That's fine, we we'll re-extract the key again.
-            pass
-        except Exception as e:
-            self.print_msg("Cannot add CIFS service: %s" % e)
-
+    def clean_previous_keytab(self, keytab=None):
+        """
+        Purge old CIFS keys from samba and clean up samba ccache
+        """
         self.clean_samba_keytab()
         installutils.remove_ccache(paths.KRB5CC_SAMBA)
 
-        try:
-            ipautil.run(["ipa-getkeytab", "--server", self.fqdn,
-                                          "--principal", self.principal,
-                                          "-k", self.keytab])
-        except ipautil.CalledProcessError:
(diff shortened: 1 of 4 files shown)


ipaserver/install/adtrustinstance.py
```diff
@@ -530,27 +530,18 @@ def __setup_group_membership(self):
...
-            # CIFS principal already exists, it is not the first time
-            # adtrustinstance is managed
-            # That's fine, we we'll re-extract the key again.
-            pass
-        except Exception as e:
-            self.print_msg("Cannot add CIFS service: %s" % e)
-
+    def _clean_previous_keytab(self):
+        """
+        Purge old CIFS keys from /etc/krb5.keytab and clean up samba ccache
```
abbra: Samba does not use /etc/krb5.keytab. Instead, Samba has own keytab in /etc/samba/samba.keytab.

ipaserver/install/adtrustinstance.py
```diff
@@ -530,27 +530,18 @@ def __setup_group_membership(self):
...
-        try:
-            ipautil.run(["ipa-getkeytab", "--server", self.fqdn,
-                                          "--principal", self.principal,
-                                          "-k", self.keytab])
-        except ipautil.CalledProcessError:
-            root_logger.critical("Failed to add key for %s"
-                                 % self.principal)
+    def _set_keytab_owner(self):
+        """
+        Do not re-set ownership of /etc/krb5.keytab
```
abbra: Same here -- Samba uses /etc/samba/samba.keytab. That keytab file should be root:root.

martbab:
Sorry I just got confused for a bit. Fixed the docstring now refer only to 'samba keytab'.

martbab:
I have added some commits to cope with changes made during privliege spearation work

martbab:
Bump for review.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/af998c4d30175fb3ecc148e1b3a7aca03ef9239a
https://fedorahosted.org/freeipa/changeset/6c0baa6208c2bf97b5ed7ea6e9836963dced64b0
https://fedorahosted.org/freeipa/changeset/ce3baf28ce81458e1c5bf57188858d3d120ec3dd
https://fedorahosted.org/freeipa/changeset/8bac62b7f5d01ceb20388599e8549b1b222f283e

PR #459, opened 2017-02-13
Faster JSON encoder/decoder

tiran:
Improve performance of FreeIPA's JSON serializer and deserializer.

* Don't indent and sort keys. Both options trigger a slow path in
  Python's json package. Without indention and sorting, encoding
  mostly happens in optimized C code.
* Replace O(n) type checks with O(1) type lookup and eliminate
  the use of isinstance().
* Check each client capability only once for every conversion.
* Use decoder's obj_hook feature to traverse the object tree once and
  to eliminate calls to isinstance().

Closes: https://fedorahosted.org/freeipa/ticket/6655

### DIFF
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -51,7 +51,7 @@ from six.moves import urllib
 from ipalib.backend import Connectible
 from ipalib.constants import LDAP_GENERALIZED_TIME_FORMAT
 from ipalib.errors import (public_errors, UnknownError, NetworkError,
-    KerberosError, XMLRPCMarshallError, JSONError, ConversionError)
+    KerberosError, XMLRPCMarshallError, JSONError)
 from ipalib import errors, capabilities
 from ipalib.request import context, Connection
 from ipapython.ipa_log_manager import root_logger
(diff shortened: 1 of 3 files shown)


pvoborni:
Is there a way(I did not read changes thoroughly) to enable sorting and indentation, e.g. for testing purposes?

tiran:
Why would you want to sort or indent the raw output? The extra verbose output of ```ipa``` just loads and dumps the output a second time. It's less efficient but who cares about minor efficiency issues of a debug feature? For browser testing, any web developer tool will give you nicely formatted JSON, too.

abbra:
Right, as long as ipa CLI is capable to print formatted debug output, that's enough.

pvoborni:
It's usually quicker to read raw response in browser than the folded "preview" because everything is visible and no clicking is required. Same for curl testing. But for curl I can imagine piping it to some tool.

tiran:
```curl url | python -m json.tool```

**Comment by pvoborni** [comment]
As mention on meeting, if rpcserver prettyprints into output in debug mode then it is fine.

**Comment by tiran** [comment]
@pvoborni I have modified the PR and added a pretty_print option. JSON is now pretty printed for verbose level 2 and higher.

The old implementation converted all list to tuples. With ```obj_hook```, only lists in a JSON objects are converted at the moment. Nested lists are not fully converted, which causes a test failure. I wonder why we decided to convert lists to tuples in the first place? Can we drop the conversion and just use lists here?

File: ipalib/rpc.py [hunk 1]
```diff
@@ -274,102 +274,171 @@ def xml_dumps(params, version, methodname=None, methodresponse=False,
...
+    * DNSName -> {'__dns_name__': unicode}
+
+    The _ipa_obj_hook() functions unserializes the marked JSON objects to
+    bytes, datetime and DNSName.
+
+    :see: _ipa_obj_hook
+    """
+    __slots__ = ('version', '_cap_datetime', '_cap_dnsname')
+
+    _identity = object()
```
HonzaCholasta: Could you please add a comment explaining that `_identity = object()` with `if func is _identity then arg else func(arg)` is faster than `_identity = lambda x: x` with `func(arg)` so that no one comes in the future and replaces the former with the latter because it's duplicate code?

ipalib/rpc.py
```diff
@@ -274,102 +274,171 @@ def xml_dumps(params, version, methodname=None, methodresponse=False,
...
+def json_decode_binary(val):
+    """Convert serialized JSON string back to Python data structure
+
+    :param val: JSON string
+    :type val: str, bytes
+    :return: Python data structure
+    :see: _ipa_obj_hook, _JSONPrimer
+    """
+    if isinstance(val, bytes):
+        val = val.decode('utf-8')
```
HonzaCholasta: IMO decoding the value should be the caller's responsibility.

MartinBasti:
LGTM

ipalib/rpc.py
```diff
@@ -274,102 +274,171 @@ def xml_dumps(params, version, methodname=None, methodresponse=False,
...
+    """Serialize a Python object structure to JSON
+
+    :param object val: Python object structure
+    :param str version: client version
+    :param bool pretty_print: indent and sort JSON (warning: slow!)
+    :return: text
+    :note: pretty printing triggers a slow path in Python's JSON module. Only
+           use pretty_print in debug mode.
+    """
+    result = _JSONPrimer(version).convert(val)
```
HonzaCholasta: Wouldn't it be a little bit faster if we kept the type - encoder mapping in a global object initialized once rather than reconstructing it in every `json_encode_binary()` call?

ipalib/rpc.py
tiran: It should be handled by the JSON library if possible. The approach allows us to replace the JSON library with a more efficient one without breaking API.

ipalib/rpc.py
tiran: Yes, it might be a tiny bit more efficient, but also more complicated. I would have to find another way to handle client versions difference efficiently. It's a constant startup cost any way. ```__hash__``` and ```__eq__``` are efficient.

HonzaCholasta: OK.

ipalib/rpc.py
HonzaCholasta: OK.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/8159c2883bf66980582d1227c364df4e592bdd7e
https://fedorahosted.org/freeipa/changeset/b12b1e4c0b19a84ccffcc702ab608d818382a697
https://fedorahosted.org/freeipa/changeset/3cac0378e94efc2ee1070eff2984eb1147bcf463
https://fedorahosted.org/freeipa/changeset/2ff07b958079e5a8972b2e7a06881521361746cc
https://fedorahosted.org/freeipa/changeset/1d7fcfe15d279e50d9ac29464a30f8e594db1802

PR #462, opened 2017-02-14
[WIP] pylint: add custom check for forbidden imports

MartinBasti:
Some modules of FreeIPA should not be imported to some other FreeIPA
modules, like ipalib into ipapython

This is WIP, it misses a lot of rules and I had hard time with naming variables, feedback more than welcome.

### DIFF
--- a/pylint_plugins.py
+++ b/pylint_plugins.py
@@ -9,10 +9,9 @@ import sys
 
 from astroid import MANAGER
 from astroid import scoped_nodes
-
-
-def register(linter):
-    pass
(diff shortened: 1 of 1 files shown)


tiran:
Can you turn module matching into a regular expression? We need bit more advanced checks, e.g. ```ipalib``` should not import from ```ipaplatform``` except for modules in ```ipalib.install```.

MartinBasti:
@HonzaCholasta has some WIP patches for this in his drawer which may be better than this, so closing this PR in favor of Honza's patches

PR #463, opened 2017-02-14
pylint_plugins: add forbidden import checker

HonzaCholasta:
Add new pylint AST checker plugin which implements a check for imports
forbidden in IPA. Which imports are forbidden is configurable in pylintrc.

Provide default forbidden import configuration and disable the check for
existing forbidden imports in our code base.

Supersedes @MartinBasti's PR #462.

### DIFF
--- a/pylint_plugins.py
+++ b/pylint_plugins.py
@@ -5,14 +5,18 @@
 from __future__ import print_function
 
 import copy
+import os.path
 import sys
 
 from astroid import MANAGER
 from astroid import scoped_nodes
+from pylint.checkers import BaseChecker
+from pylint.checkers.utils import check_messages
+from pylint.interfaces import IAstroidChecker
 
 
 def register(linter):
-    pass
+    linter.register_checker(IPAChecker(linter))
 
 
 def _warning_already_exists(cls, member):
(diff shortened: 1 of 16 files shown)


pylint_plugins.py
```diff
@@ -255,3 +259,79 @@ def fix_ipa_classes(cls):
...
 
 MANAGER.register_transform(scoped_nodes.Class, fix_ipa_classes)
+
+
+class IPAChecker(BaseChecker):
+    __implements__ = IAstroidChecker
+
+    name = 'ipa'
+    msgs = {
+        'E9901': (
```
MartinBasti: Could we use Warning instead of error?

pylint_plugins.py
```diff
@@ -255,3 +259,79 @@ def fix_ipa_classes(cls):
...
+            'Forbidden import %s',
```
MartinBasti: Maybe would be nice to write whihc rule has been applied `Forbidden import %s (%s should not import %s)`

pylint_plugins.py
HonzaCholasta: Sure.

pylint_plugins.py
HonzaCholasta: I agree.

MartinBasti:
> Can you turn module matching into a regular expression? We need bit more advanced checks, e.g. ipalib should not import from ipaplatform except for modules in ipalib.install.

How can be the issue mentioned by @tiran solved in this PR? should regexp be used or allow rules added?

HonzaCholasta:
@MartinBasti, this issue is already solved in the PR without using regular expressions. See `pylintrc` for example.

MartinBasti:
Ok, this will not work if ipaclient/submodule allows to import any module, but seems OK for me now, can be improved when needed

HonzaCholasta:
I don't know what you mean, could you give me an example?

MartinBasti:
In this case:
```
   ipaclient/:ipaclient.install:ipalib.install:ipaplatform:ipaserver,
   ipaclient/install/:ipaserver,
```

`ipaclient/install` allows all import everything but `ipaserver`, but I cannot currently specify a rule that allows `ipaclient/install` import everything (with `ipaserver`)

But as I said this is a corner case, should be done when needed

HonzaCholasta:
You can, using:
```
    ipaclient/install/
```

MartinBasti:
Awesome then

HonzaCholasta:
The format could be nicer though - suggestions are welcome.

martbab:
@MartinBasti any progress in reviewing this PR?

HonzaCholasta:
@martbab, I haven't incorporated @MartinBasti's suggestions in yet.

HonzaCholasta:
I have now incorporated the suggestions.

MartinBasti:
master:

* 5d489ac5604ca959cfe439c0594b8739073f3cea pylint_plugins: add forbidden import checker

PR #465, opened 2017-02-14
Tests: search for disabled users

MartinBasti:
Add tests for searching disabled/enabled users.
   
XFAIL: newly created users has no 'nsaccountlock' attribute set and
user-find doesn't return them as active users. This should be fixed.

Partially tests: #444

### DIFF
--- a/ipatests/test_xmlrpc/test_user_plugin.py
+++ b/ipatests/test_xmlrpc/test_user_plugin.py
@@ -240,6 +240,47 @@ class TestFind(XMLRPC_test):
         result = command()
         user.check_find(result, pkey_only=True)
 
+    @pytest.mark.xfail(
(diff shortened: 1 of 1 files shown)


stlaz:
Is there a ticket for the xfail scenario?

MartinBasti:
No because according @HonzaCholasta this is expected framework behavior

stlaz:
Does that mean that `user-find` no longer works?

MartinBasti:
@stlaz That means the any *-find command may work unexpectedly with non-mandratory attribute. For this case you must get all active users by `user-find --disabled=false` + `user-find --disabled=`

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/79b3fbf97d66adb1f5c960e5473b90f85cbe145a

PR #466, opened 2017-02-15
pkinit: make sure to have proper dictionary for Kerberos instance on upgrade

abbra:
When running PKINIT upgrade we need to make sure full substitution
dictionary is in place or otherwise executing LDAP updates will fail to
find proper objects because $SUFFIX, $DOMAIN, and other variables
will not be substituted.

Fixes https://fedorahosted.org/freeipa/ticket/6670


MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/14d84daf29543978c6383da10f4f2d913346f013

PR #467, opened 2017-02-15
ipaclient: schema cache: Write all schema files in concurrent-safe way

dkupka:
https://fedorahosted.org/freeipa/ticket/6668

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -282,6 +282,41 @@ def write_tmp_file(txt):
 
     return fd
 
+
+@contextmanager
+def concurrent_open(filename, mode='r'):
+    """Ensure that complete file is read/written
+
+    Works only for r, rb, w, w+, wb, wb+ modes. Default is r.
+
+    In read mode behaves the same as build-in open. In write mode all data are
+    written to the temporary file first and then moved to target path.
+    This approach ensures that reading will be performed on complete file and
+    writting will result in complete or none file written at all.
+    """
+    if mode in ('w', 'w+', 'wb', 'wb+'):
+        directory, basename = os.path.split(os.path.abspath(filename))
+        with tempfile.NamedTemporaryFile(
+                mode=mode, dir=directory, prefix=basename, delete=False
+        ) as temp_file:
+            try:
+                yield temp_file
+            finally:
+                temp_file.flush()
+                os.fsync(temp_file.fileno())
+                try:
+                    os.rename(temp_file.name, filename)
+                except EnvironmentError:
+                    os.unlink(temp_file.name)
+                    raise
+    elif mode in ('r', 'rb'):
+        with open(filename, mode) as f:
+            yield f
+    else:
+        raise ValueError(u"mode string must be one of 'r', 'rb', 'w', 'w+', "
+                         u"'wb', 'wb+'")
+
+
 def shell_quote(string):
     if isinstance(string, str):
         return "'" + string.replace("'", "'\\''") + "'"
--- a/ipaclient/remote_plugins/__init__.py
+++ b/ipaclient/remote_plugins/__init__.py
@@ -14,6 +14,8 @@ from . import schema
 from ipaclient.plugins.rpcclient import rpcclient
 from ipapython.dnsutil import DNSName
 from ipapython.ipa_log_manager import log_mgr
+from ipapython.ipautil import concurrent_open
+
 
 logger = log_mgr.get_logger(__name__)
 
@@ -58,7 +60,8 @@ class ServerInfo(collections.MutableMapping):
             except EnvironmentError as e:
                 if e.errno != errno.EEXIST:
                     raise
-            with open(self._path, 'w') as sc:
+
+            with concurrent_open(self._path, 'w') as sc:
                 json.dump(self._dict, sc)
         except EnvironmentError as e:
             logger.warning('Failed to write server info: {}'.format(e))
--- a/ipaclient/remote_plugins/schema.py
+++ b/ipaclient/remote_plugins/schema.py
@@ -5,8 +5,6 @@
 import collections
 import contextlib
 import errno
-import fcntl
-import io
 import json
 import os
 import sys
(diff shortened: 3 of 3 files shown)


ipapython/ipautil.py
```diff
@@ -282,6 +282,37 @@ def write_tmp_file(txt):
...
+    Works only for r, rb, w, w+, wb, wb+ modes. Default is r.
+
+    In read mode behaves the same as build-in open. In write mode all data are
+    written to the temporary file first and then moved to target path.
+    This approach ensures that reading will be performed on complete file and
+    writting will result in complete or none file written at all.
+    """
+    if mode in ('w', 'w+', 'wb', 'wb+'):
+        with tempfile.NamedTemporaryFile(
+                mode=mode, prefix=path, delete=False) as temp_file:
```
tiran: prefix is not enough. You have to include the ```dir``` arg to ensure that the file is created in the same directory as the target dir. Otherwise both files may end up on two diferent file systems and os.rename() is no longer atomic.

https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp

You also can't just yield the temp file or its file attribute, because ```with concurrent_open(...)``` will call ```TemporaryFile.__exit__```, which closes the fd and breaks ```fsync()```.

```
class RenameContext(object):
    def __init__(self, temp_file, destpath):
        self.temp_file = temp_file
        self.destpath = destpath

    def __enter__(self):
        return self.temp_file.file  # raw file

    def __exit__(self, exc, value, tb):
        f = self.temp_file
        f.flush()
        os.fsync(f.fileno())
        f.close()
        try:
            os.rename(f.name, self.destpath)
        except OSError:
            os.unlink(f.name)  # only Windows developers call this remove :)
            raise
```

And then

```
filename = os.path.abspath(filename)
directory, prefix = os.path.splitpath(filename)
with tempfile.NamedTemporaryFile(mode=mode, dir=directory, prefix=prefix, delete=False) as f:
    return RenameContext(f, filename)
```

ipapython/ipautil.py
```diff
@@ -282,6 +282,37 @@ def write_tmp_file(txt):
 
     return fd
 
+
+@contextmanager
+def concurrent_open(path, mode='r'):
```
dkupka: build-in open uses 'name' in python 2 and 'file' in python 3. So why not? I can combine it to 'filename'.

ipapython/ipautil.py
```diff
@@ -282,6 +282,37 @@ def write_tmp_file(txt):
...
+    """
+    if mode in ('w', 'w+', 'wb', 'wb+'):
+        with tempfile.NamedTemporaryFile(
+                mode=mode, prefix=path, delete=False) as temp_file:
+            yield temp_file
+            temp_file.flush()
+            os.fsync(temp_file.fileno())
+            try:
+                os.rename(temp_file.name, path)
+            except EnvironmentError:
```
dkupka: AFAIK EnvironmentError in python 2 is super class for IOError and OSError, in python3 EnvironmentError and IOError are aliases for OSError. So I believe this is correct way of catching any OS/FS-related error in both python 2 and python 3.

ipapython/ipautil.py
dkupka: Regarding the prefix you're right, I haven't thought about relative names. I know that rename works only inside single FS otherwise raises error.

Regarding yielding `NamedTemporaryFile` instance from concurrent_open I'm not sure I understand you. Do you want to say that `NamedTemporaryFile.__exit__` is called when `yield temp_file` is executed on line 300? I don't agree with that because the context haven't finished yet. The problem I see not with line 300 is that I should enclose it with `try: finally:`.

tiran: Ah, I didn't see that you wrapped ```concurrent_open``` in a ```@contextmanager```. In that case the ```__exit__``` function of ```NamedTemporaryFile``` is not called because the function doesn't return the ```NamedTemporaryFile``` instance directly but rather a wrapped instance.

try/except OSError is fine. You don't want to call ```os.unlink()``` when ```os.rename()``` was successfully called.

ipapython/ipautil.py
tiran: OSError is sufficient. os.rename() can never raise IOError.

ipapython/ipautil.py
tiran: PS: Without ```dir``` argument, the tempfile module uses the system's temp file directory. (TMPDIR, TEMP, TMP, /tmp, /var/tmp). These days temp dir are on a different file system, partly for performance, partly for security reasons.

ipapython/ipautil.py
dkupka: Since I've first read PEP 3151 (expecially section about Confusing OS-related exceptions [1])  I like to play it safely and always catch EnvironmentError because (like to PEP says):
In fact, it is hard to think of any situation where OSError should be caught but not IOError, or the reverse.

[1] https://www.python.org/dev/peps/pep-3151/#confusing-set-of-os-related-exceptions

ipapython/ipautil.py
```diff
@@ -282,6 +282,41 @@ def write_tmp_file(txt):
...
+            try:
+                yield temp_file
+            finally:
+                temp_file.flush()
+                os.fsync(temp_file.fileno())
+                try:
+                    os.rename(temp_file.name, filename)
+                except EnvironmentError:
+                    os.unlink(temp_file.name)
+                    raise
```
tiran: The new try/finally also overrides the destination file when there is an exception. How about:

```
            try:
                yield temp_file
            except Exception:
                raise
            else:
                temp_file.flush()
                os.fsync(temp_file.fileno())
                temp_file.close()
                os.rename(temp_file.name, filename)
            finally:
                os.unlink(temp_file.name)
```

The os.unlink() function is always executed as last statement, https://www.python.org/dev/peps/pep-0341/

tiran: correction, unlink doesn't override destination.

```
            try:
                yield temp_file
            except Exception:
                raise
            else:
                temp_file.flush()
                os.fsync(temp_file.fileno())
                temp_file.close()
                os.rename(temp_file.name, filename)
            finally:
                try:
                    os.unlink(temp_file.name)
                except OSError as e:
                    if e.errno != errno.ENOENT:
                        raise
```

dkupka:
superseded by #488

PR #468, opened 2017-02-15
Remove non-sensical kdestroy on https stop

simo5:
This kdestroy runs as root and wipes root's own ccachs ...
this is totally inappropriate.

https://fedorahosted.org/freeipa/ticket/6673

### DIFF
--- a/install/share/ipa-httpd.conf.template
+++ b/install/share/ipa-httpd.conf.template
@@ -1,7 +1,7 @@
 # Do not edit. Created by IPA installer.
 
 [Service]
+Environment=KRB5CCNAME=$KRB5CC_HTTPD
 Environment=GSS_USE_PROXY=yes
 Environment=KDCPROXY_CONFIG=$KDCPROXY_CONFIG
 ExecStartPre=$IPA_HTTPD_KDCPROXY
-ExecStopPost=$POST
--- a/ipaplatform/redhat/tasks.py
+++ b/ipaplatform/redhat/tasks.py
@@ -458,7 +458,7 @@ class RedHatTaskNamespace(BaseTaskNamespace):
             dict(
                 KDCPROXY_CONFIG=paths.KDCPROXY_CONFIG,
                 IPA_HTTPD_KDCPROXY=paths.IPA_HTTPD_KDCPROXY,
-                POST='-{kdestroy} -A'.format(kdestroy=paths.KDESTROY)
+                KRB5CC_HTTPD=paths.KRB5CC_HTTPD,
             )
         )
 
--- a/ipaplatform/base/paths.py
+++ b/ipaplatform/base/paths.py
@@ -351,5 +351,6 @@ class BasePathNamespace(object):
     IPA_GETKEYTAB = '/usr/sbin/ipa-getkeytab'
     EXTERNAL_SCHEMA_DIR = '/usr/share/ipa/schema.d'
     GSSPROXY_CONF = '/etc/gssproxy/10-ipa.conf'
+    KRB5CC_HTTPD = '/tmp/krb5cc-httpd'
 
 path_namespace = BasePathNamespace
--- a/ipaplatform/debian/paths.py
+++ b/ipaplatform/debian/paths.py
@@ -89,7 +89,6 @@ class DebianPathNamespace(BasePathNamespace):
     VAR_OPENDNSSEC_DIR = "/var/lib/opendnssec"
     OPENDNSSEC_KASP_DB = "/var/lib/opendnssec/db/kasp.db"
     IPA_ODS_EXPORTER_CCACHE = "/var/lib/opendnssec/tmp/ipa-ods-exporter.ccache"
-    KRB5CC_HTTPD = "/var/run/apache2/ipa/krbcache/krb5ccache"
     IPA_CUSTODIA_SOCKET = "/run/apache2/ipa-custodia.sock"
     IPA_CUSTODIA_AUDIT_LOG = '/var/log/ipa-custodia.audit.log'
 


ipaplatform/redhat/tasks.py
```diff
@@ -451,8 +451,7 @@ def configure_httpd_service_ipa_conf(self):
             dict(
                 KRB5CC_HTTPD=paths.KRB5CC_HTTPD,
                 KDCPROXY_CONFIG=paths.KDCPROXY_CONFIG,
-                IPA_HTTPD_KDCPROXY=paths.IPA_HTTPD_KDCPROXY,
-                POST='-{kdestroy} -A'.format(kdestroy=paths.KDESTROY)
+                IPA_HTTPD_KDCPROXY=paths.IPA_HTTPD_KDCPROXY
```
tiran: please leave the trailing comma

martbab:
I would rather keep `kdestroy` there, but only really purge the apache ccache explicitly:

```diff
--- a/ipaplatform/redhat/tasks.py
+++ b/ipaplatform/redhat/tasks.py
@@ -452,7 +452,8 @@ class RedHatTaskNamespace(BaseTaskNamespace):
                 KRB5CC_HTTPD=paths.KRB5CC_HTTPD,
                 KDCPROXY_CONFIG=paths.KDCPROXY_CONFIG,
                 IPA_HTTPD_KDCPROXY=paths.IPA_HTTPD_KDCPROXY,
-                POST='-{kdestroy} -A'.format(kdestroy=paths.KDESTROY)
+                POST='-{kdestroy} -c {ccache}'.format(
+                    kdestroy=paths.KDESTROY, ccache=paths.KRB5CC_HTTPD)
             )
         )
```

Otherwise we will bump into regressions caused by stale HTTPD ccaches left over when backing up/restoring IPA installation. I would demonstrate it on a failing backup/restore tests but they are completely messed up right now.

tiran:
Why do we back up ccache in the first place?

martbab:
We do not backup ccache, we back up apache keytab.

During restore into installer server we back up old Kerberos keys, but without any mechanism to purge the new apache ccache acquired during the installation of new server you would end up with key mismatch and nothing would work until the ccache expires.

As to why a) we backup Kerberos keys, and b) support restoring into running IPA server that is beyond me.

martbab:
@rcritten can you please re-read my comment very slowly? I wrote that we *do* backup keytabs.

martbab:
And indeed I can reproduce the original failure reported in https://fedorahosted.org/freeipa/ticket/5296 with this PR.

If I manually remove apache ccache (kdestroy -c /var/run/httpd/ipa/krbcache/krb5ccache) I can use the framework again.

rcritten:
Rudeness is not necessary.

You said:

"As to why a) we backup Kerberos keys, and b) support restoring into running IPA server that is beyond me."

The reason for a) is to restore an exact copy of what was backed up. 

As for b, the idea of restoring into a running IPA server to replace the existing install with a new one is something I discuss in some detail at http://www.freeipa.org/page/V3/Backup_and_Restore and outline the ton of problems associated with it. It was never intended to be supported due to these issues.

tiran:
I'm with @rcritten .

If we need to clean up / remove some files during a restore, then these clean-ups should be handled by ```ipa-restore```. The service files are IMHO the wrong place.

pvoborni:
And AFAIK b) is not supported.  @martbab , does something indicate otherwise?

martbab:
@rcritten I apologize for sounding rude. I misread your comment and interpreted it differently than intended.

That said, if the restore to a running IPA server is not intended to be supported, why do we have a number of tests for this scenario?  I have tried to find some discussion in the design page you posted but did not find any discussion of restore into running server, only the steps taken.

@tiran I tend to agree with you now. It seemed like a good idea to purge ccaches in the unit file when we switched from KEYRING: to FILE: for apache. However the restore use-case is not the only one which can result into stale ccache, I can also think about requesting new Apache keytab, restarting the service and be left with a stale ccache and key mismatch again.

MartinBasti:
@pvoborni this is the way how it this tested by QA, so that's why I added this kind of test to upstream. I disagree that `b)` is not supported. It is just emulation fo case when user ruined kerberos keytabs and service configuration and the one needs to restore backup on the installed server.

simo5:
@MartinBasti the unit files are the wrong place to destroy ccaches, especially given they run as a different user (root) and may not have access to destroy stuff when we start using KCM.
If we need clear ccaches then we need a different plan, please reopen the original bug, and push this PR to fix the impeding issue.

MartinBasti:
@simo5 any ideas how this should be fixed? We cannot push this patch without additional fix of removing outdated ccache because it will cause permanent fail of CI for backup/restore and it will mask real issues.

simo5:
If this is about backup/restore, add a kdestroy ccache in the restore scripts, making sue it su - apache first

MartinBasti:
how about @martbab comment?

https://github.com/freeipa/freeipa/pull/468#issuecomment-280056786

> However the restore use-case is not the only one which can result into stale ccache, I can also think about requesting new Apache keytab, restarting the service and be left with a stale ccache and key mismatch again.

simo5:
If you request a new keytab you should clean up the cacche ?
If we have a way to run the post exec command as the right user and with the right /tmp (httpd unit file uses namepaced /tmp) we could keep this code in the unit file I guess, although it would be wasteful in most cases when ccache does not change...

martbab:
Could we use just keep the post command as "kdestroy -c {apache_ccache_path}"? Or is everything chrooted into name-spaced /tmp and we can not access the ccache file from within the unit file?

abbra:
Yes, when namespaced /tmp is used, unit file does not have any view into that.

tiran:
How about we use systemd PrivateTmp for temporary files? It is not only more secure but it also automatically removes all temporary files when the service is stopped: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#PrivateTmp=

abbra:
@tiran we do use PrivateTmp already. This is not about PrivateTmp, though, because we don't store credentials caches in a private tmp.

tiran:
That's my point. Why is the ccache file not stored in ```PrivateTmp```? The ccache can be removed at any time. It doesn't have to be retained.

 ```PrivateTmp``` solves the issue for us. We just have to figure out how to combine ```tmpfiles.d``` with ```PrivateTmp``` to create ```/var/run/apache``` in private temp.

simo5:
I guess we can simply set KRB5CCNAME=/tmp/krb5_httpd in the unit file and we should be ok then.
@martbab or @mbasti, can you try that ?
If it solves your scenario we can change this PR to replace the POST with that Env in the unit file.

simo5:
Uhm I just tried setting KRB5CCNAME=/tmp/krb5_httpd in my install and ... I found out we do not actually generate an httpd ccache, so why are we trying to destroy the ccache again ?
Anyway, I am going to add this Environment line to the unit file just in case, so that we can address the issue if we ever need a ccache.

install/share/ipa-httpd.conf.template
```diff
@@ -1,7 +1,7 @@
 # Do not edit. Created by IPA installer.
 
 [Service]
+Environment=KRB5CCNAME=/tmp/krb5-httpd
```
martbab: I would rather revive  `paths.KRB5CC_HTTPD` and substitute it to the template than having magic strings.

I also noticed that `KRB5CC_HTTPD` was left over in ipaplatform.debian.paths module.

simo5: ok I'll do that

martbab: Thank you let's wait for Travis CI.

martbab:
I have also noticed that the ccache is not created there, strange. However I think it is better to explicitly specify file-based ccache anyway just to be one the safe side. Otherwise everything seems to work as expected, even `ipa-restore to live server` scenario.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/b8f304c66994ae82ea484a4e8bd057d4ccf1e6bd

martbab:
@pvoborni should the fix go also into 4-4 branch? see https://fedorahosted.org/freeipa/ticket/6673#comment:3

PR #469, opened 2017-02-15
Ignore unlink error in ipa-otpd.socket

tiran:
Don't fail in case the file does not exist.

### DIFF
--- a/daemons/ipa-otpd/ipa-otpd.socket.in
+++ b/daemons/ipa-otpd/ipa-otpd.socket.in
@@ -3,7 +3,7 @@ Description=ipa-otpd socket
 
 [Socket]
 ListenStream=@krb5rundir@/DEFAULT.socket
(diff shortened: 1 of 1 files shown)


martbab:
LGTM, but do we require this fix also in 4-4 branch?

HonzaCholasta:
This will ignore all errors, not just file does not exist. Are we OK with that?

tiran:
Would you rather use ```/usr/bin/rm -f``` to only ignore missing files but propagate permission errors? I'm not sure why unlink was used in favor of rm. @simo5 ?

simo5:
@tiran I do not know, @npmccallum may know.

npmccallum:
We shouldn't use either. We should use RemoveOnStop= now.

npmccallum:
This PR can be closed.

PR #470, opened 2017-02-15
WebUI: Size limit warning on details pages fixed

pvomacka:
Entity select fields accepted globally set size limit and in situations when
there were more entries than global size limit allows then the "Truncated" warning
shows up. Also only subset of items was shown.
All entity select widgets now uses find methods with sizelimit set to 0
which says get all entries.

This setting is configurable using search_all_entries attribute.

https://fedorahosted.org/freeipa/ticket/6618

### DIFF
--- a/install/ui/src/freeipa/rpc.js
+++ b/install/ui/src/freeipa/rpc.js
@@ -72,6 +72,12 @@ rpc.command = function(spec) {
     that.options = $.extend({}, spec.options || {});
 
     /**
+     * @property {Array} suppress_warnings array of message codes which
+     * are suppressed
+     */
+    that.suppress_warnings = spec.suppress_warnings || [];
+
+    /**
      * Success handler
      * @property {Function}
      * @param {Object} data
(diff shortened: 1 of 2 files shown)


pvoborni:
Would it be better to suppress the warning and use sensible size limit. I.e. the entity select doesn't need to show all entries. I'm afraid that it might have negative performance impact.

install/ui/src/freeipa/rpc.js
```diff
@@ -219,6 +225,7 @@ rpc.command = function(spec) {
 
         for (var i=0,l=msgs.length; i<l; i++) {
             var msg = lang.clone(msgs[i]);
+            if (that.suppress_warnings.indexOf(msg.code) > -1) break;
```
pvoborni: Shouldn't it be continue instead of break? So that it will process also other possible msgs which you don't want to suppress.

pvoborni:
Code looks good and works fine, ACK.

pvomacka:
ipa-4-5:

* 422c9058d9a6be69db4eab7db654b9184ae5eab6 WebUI: Add support for suppressing warnings
* 697a5779b377a5d76c1cb212514b6feb46326f71 WebUI: suppress truncation warning in select widget
master:

* 7b3a10da7001d7ee394cd891d926def66d0f2546 WebUI: Add support for suppressing warnings
* b9e6ad1967ba24c7ebe5181da1ebe32d30e7b28f WebUI: suppress truncation warning in select widget

PR #471, opened 2017-02-16
Fix some privilege separation regressions

HonzaCholasta:
**client install: create /etc/ipa/nssdb with correct mode**

The NSS database directory is created with mode 640, which causes the IPA
client to fail to connect to any IPA server, because it is unable to read
trusted CA certificates from the NSS database.

Create the directory with mode 644 to fix the issue.

**server upgrade: fix upgrade in CA-less**

Use /etc/httpd/alias instead of /var/lib/ipa/radb in upload_cacrt, as
/var/lib/ipa/radb is not populated in CA-less.

Do not migrate ipaCert from /etc/httpd/alias to /var/lib/ipa/radb in
CA-less, as it might be an incorrect certificate from previous CA-ful
install, and is not necessary anyway.

**server upgrade: fix upgrade from pre-4.0**

update_ca_renewal_master uses ipaCert certmonger tracking information to
decide whether the local server is the CA renewal master or not. The
information is lost when migrating from /etc/httpd/alias to
/var/lib/ipa/radb in update_ra_cert_store.

Make sure update_ra_cert_store is executed after update_ca_renewal_master
so that correct information is used.

**server upgrade: always upgrade KRA agent PEM file**

Before the KRA agent PEM file is exported in server upgrade, the sysupgrade
state file is consulted. This causes the KRA agent PEM file not to be
exported to the new location if the upgrade was executed in the past.

Do not consult the sysupgrade state file to decide whether to upgrade the
KRA agent PEM file or not, the existence of the file is enough to make this
decision.

https://fedorahosted.org/freeipa/ticket/5959
https://fedorahosted.org/freeipa/ticket/6675

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -79,14 +79,12 @@ def uninstall_ipa_memcached():
     We can't use the full service uninstaller because that will attempt
     to stop and disable the service which by now doesn't exist. We just
     want to clean up sysrestore.state to remove all references to
-    ipa_kpasswd.
+    ipa_memcached.
     """
     ipa_memcached = service.SimpleServiceInstance('ipa_memcached')
 
-    enabled = not ipa_memcached.restore_state("enabled")
+    ipa_memcached.uninstall()
 
-    if enabled is not None and not enabled:
-        ipa_memcached.remove()
 
 def backup_file(filename, ext):
     """Make a backup of filename using ext as the extension. Do not overwrite
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -1386,7 +1386,9 @@ def fix_trust_flags():
 def export_kra_agent_pem():
     root_logger.info('[Exporting KRA agent PEM file]')
 
-    if sysupgrade.get_upgrade_state('http', 'export_kra_agent_pem'):
+    sysupgrade.remove_upgrade_state('http', 'export_kra_agent_pem')
+
+    if os.path.exists(paths.KRA_AGENT_PEM):
         root_logger.info("KRA agent PEM file already exported")
         return
 
@@ -1395,8 +1397,7 @@ def export_kra_agent_pem():
         return
 
     krainstance.export_kra_agent_pem()
-
-    sysupgrade.set_upgrade_state('http', 'export_kra_agent_pem', True)
+    installutils.remove_file(paths.OLD_KRA_AGENT_PEM)
 
 
 def update_mod_nss_protocol(http):
--- a/install/updates/05-pre_upgrade_plugins.update
+++ b/install/updates/05-pre_upgrade_plugins.update
@@ -8,4 +8,3 @@ plugin: update_referint
 plugin: update_uniqueness_plugins_to_new_syntax
 
 # last
-plugin: update_ra_cert_store
--- a/install/updates/90-post_upgrade_plugins.update
+++ b/install/updates/90-post_upgrade_plugins.update
@@ -15,6 +15,8 @@ plugin: update_idrange_type
 plugin: update_pacs
 plugin: update_service_principalalias
 plugin: update_upload_cacrt
+# update_ra_cert_store has to be executed after update_ca_renewal_master
+plugin: update_ra_cert_store
 
 # last
 # DNS version 1
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -2284,7 +2284,7 @@ def install_check(options):
 
 def create_ipa_nssdb():
     db = certdb.NSSDatabase(paths.IPA_NSSDB_DIR)
-    db.create_db(backup=True)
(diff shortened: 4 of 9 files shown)


ipaserver/install/server/upgrade.py
```diff
@@ -1386,7 +1386,9 @@ def fix_trust_flags():
 def export_kra_agent_pem():
     root_logger.info('[Exporting KRA agent PEM file]')
 
-    if sysupgrade.get_upgrade_state('http', 'export_kra_agent_pem'):
+    sysupgrade.remove_upgrade_state('http', 'export_kra_agent_pem')
+
+    if os.path.exists(paths.KRA_AGENT_PEM):
```
tiran: By the way who takes care of renewal of ```kra-agent.pem``` file? certmonger only tracks the cert in NSSDB.

```
Request ID '20170215101245':
        status: MONITORING
        stuck: no
        key pair storage: type=NSSDB,location='/var/lib/ipa/radb',nickname='ipaCert',token='NSS Certificate DB',pinfile='/var/lib/ipa/radb/pwdfile.txt'
        certificate: type=NSSDB,location='/var/lib/ipa/radb',nickname='ipaCert',token='NSS Certificate DB'
        CA: IPA
        issuer: CN=Certificate Authority,O=IPA.EXAMPLE
        subject: CN=IPA RA,O=IPA.EXAMPLE
        expires: 2019-02-03 15:26:21 UTC
        key usage: digitalSignature,nonRepudiation,keyEncipherment,dataEncipherment
        eku: id-kp-serverAuth,id-kp-clientAuth
        pre-save command: 
        post-save command: 
        track: yes
        auto-renew: yes

```

tiran: It's a separate issue. I created https://fedorahosted.org/freeipa/ticket/6680 to track it.

stlaz:
LGTM

stlaz:
Upgrade still fails when run for the first time during `dnf update`:
http://pastebin.com/H4kt6hVb
When I run it by hand after this failure, it gets a bit further, but NSSConnection fails in the `[Migrating certificate profiles to LDAP]` step:
http://pastebin.com/8tBjYjkU

tiran:
@HonzaCholasta, we got merge conflicts.

HonzaCholasta:
@stlaz, not sure what's going on there, but not my fault, these failures happen even without this PR.

ipaserver/install/server/upgrade.py
```diff
@@ -79,11 +79,12 @@ def uninstall_ipa_memcached():
...
     want to clean up sysrestore.state to remove all references to
-    ipa_kpasswd.
+    ipa_memcached.
     """
     ipa_memcached = service.SimpleServiceInstance('ipa_memcached')
 
     enabled = not ipa_memcached.restore_state("enabled")
+    ipa_memcached.restore_state("running")
 
     if enabled is not None and not enabled:
```
stlaz: Totally OT: `enabled` will never be None so the first part of the condition is redundant.

ipaserver/install/server/upgrade.py
```diff
@@ -79,11 +79,12 @@ def uninstall_ipa_memcached():
     We can't use the full service uninstaller because that will attempt
     to stop and disable the service which by now doesn't exist. We just
     want to clean up sysrestore.state to remove all references to
-    ipa_kpasswd.
+    ipa_memcached.
     """
     ipa_memcached = service.SimpleServiceInstance('ipa_memcached')
 
     enabled = not ipa_memcached.restore_state("enabled")
+    ipa_memcached.restore_state("running")
```
stlaz: Does the same problem exist with `kpasswd` which is where this copy-paste code has its origin?

stlaz:
Note that `KRA_AGENT_PEM` will not be moved to the correct folder if KRA is not installed but that's fine with me.
`/bin/systemctl status  ipa_memcached.service` still shows the service as `running` although there's the strange line `Loaded: not-found (Reason: No such file or directory)`. That does not seem ok, should we stop the service as well?

ipaserver/install/server/upgrade.py
HonzaCholasta: No one complained since the code was added in 2012, so I guess it doesn't.

stlaz:
The raised issues seem to have been fixed. ACK.

HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/b4fa354f500bcf3ac23ee3805f2c166c6a635b92
https://fedorahosted.org/freeipa/changeset/ba8a10fbdb39cab672038e1a6dc9c7507070cdf9
https://fedorahosted.org/freeipa/changeset/97e838e10da3b42e3605d230e0b8e01b9148876f
https://fedorahosted.org/freeipa/changeset/0862e320916e0123df7e8505ba61229db0cb1e4a
https://fedorahosted.org/freeipa/changeset/6d34c2169fcd520cc726e58e01d008ae3637aad4

PR #472, opened 2017-02-16
Packaging: Add placeholder packages

tiran:
The ipa and freeipa packages are placeholders to prevent PyPI squashing
attacks and reserve the names for future use. `pip install ipa` installs
ipaclient.

The new PR provides just the two placeholder packages from PR #379.


MartinBasti:
Shouldn't we have build dependency on `python[3]-wheel` without it `bdist_wheel` target is not working

MartinBasti:
I'm not an expert about how PyPI is working, but shouldn't be there also placeholder packages for:
- ipaserver
- ipaplatform
- ipatests


How about [free]ipa-server and friends, can a bad person use this for an attack when somebody uses `pip install freeipa-server` ?

configure.ac
```diff
@@ -577,6 +577,9 @@ AC_CONFIG_FILES([
     ipaserver/Makefile
     ipatests/Makefile
     ipatests/man/Makefile
+    packaging/Makefile
```
MartinBasti: can we rename this to `wheel_packaging` or `pypi_packaging` to be clear what is the purpose, because it is only for `bdist_wheel`

tiran: Sure! how about just pypi?

tiran:
At the moment wheels are not required for RPM building. python-wheel is not available on RHEL, but I can work around it. Should the RPM spec file only contain dependencies for RPM packaging or also dependencies for general development and general packaging?

I can create placeholder modules for ipaserver, ipaplatform and ipatests, too. It's going to be a bit more tricky. Let's see...

configure.ac
MartinBasti: or `pypi_placeholders`

HonzaCholasta:
Is this really the right thing to do? IMO it does not make much sense to have placeholders for every `ipa*` package, as it does not scale at all - nothing is preventing a potential attacker to register their own `ipa*` package, which will confuse PyPI users all the same and will prevent us to use that name ourselves in the future, should we want to.

MartinBasti:
We want to prevent others to have packages in PyPI with the same names as used for IPA. This is reasonable for protecting users to get attacker code from PyPI and rewrite working modules installed from rpms. In case that somebody install `ipamodulefromhell` we really cannot help this user

tiran:
Yes, it is the right thing to do. You can trust in the expert with a decade of experience with Python packaging (formerly known as cheese shop).

MartinBasti:
I talked with Honza how to handle the build dependency for pypi, and we may to remove the commit that adds python-wheel or add new option to specfile that will install pypi related packages  `with_pypi` or so.

Do you plan to have more dependencies related only to pypi?

tiran:
You requsted a dependency in the first place :)

If you are going to add a special build or dependency flavor for PyPI packaging, please also add ```python[23]-twine```. It's the uploader tool we are going to use to upload packages to PyPI.

tiran:
OK, you got ```with_wheels``` in ```freeipa.spec.in``` now. ```with_wheels``` is more logical than ```with_pypi``` because wheels have more uses than just PyPI upload.

MartinBasti:
Thank you.

I see errors reported by pylint
```
************* Module ipaserver.install.installutils
ipaserver/install/installutils.py:1209: [E1101(no-member), store_version] Module 'ipaplatform' has no 'NAME' member)
ipaserver/install/installutils.py:1221: [E1101(no-member), check_version] Module 'ipaplatform' has no 'NAME' member)
ipaserver/install/installutils.py:1224: [E1101(no-member), check_version] Module 'ipaplatform' has no 'NAME' member)
```

MartinBasti:
LGTM, please rebase and I will test it.

tiran:
@MartinBasti I have rebased the branch and added wheel + placeholder building to make check. The pylint violations have disappeared.

MartinBasti:
Is really needed to have this is make check? It makes build dependencies of wheel mandratory not optional

```
error: invalid command 'bdist_wheel'
make[5]: *** [bdist_wheel] Error 1
Makefile:591: recipe for target 'bdist_wheel' failed
make[5]: Leaving directory '/freeipa/rpmbuild/BUILD/freeipa-4.4.90.dev201702221837+git07cd377/pypi/freeipa'
Makefile:1192: recipe for target 'wheel_placeholder' failed
make[4]: *** [wheel_placeholder] Error 1
make[4]: *** Waiting for unfinished jobs....
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help
```

tiran:
I'm open to suggestions here, but I like to have automatic validation of packaging.

lslebodn:
I agree with @MartinBasti it is not a unit test.
IMHO better approach is to test it in CI/travis/...

tiran:
That's not the point here. We are arguing about a new build dependency (python-wheel).

lslebodn:
FYI: At the moment, `make check` run just C-based unit test and all of them are optional.
If required dependency is not found at configure time then test is not build/executed.

The reason is that required dependencies are not in some downstream distributions and IIRC python-wheel is not there either.

pvoborni:
Some distros like RHEL doesn't have python-wheel packaged. It can be disabled by downstream patch, but better would be to remove it or make it configurable.

lslebodn:
NACK for downstream patch. The intentin of build system refactoring was make packaging in downstream simpler.

MartinBasti:
@tiran I thought we agreed on having `with_wheels` in specfile and install dependencies only when you want to build wheel packages, what is not the case of RHEL. So what is the issue with python-wheel?

My only concern it to not run wheel build in make check

tiran:
@MartinBasti I dropped the last commit. make check no longer checks wheel packages.

I'm going to open a new ticket for @martbab and ask him to add to add a proper test for wheel building to his awesome container magic.

MartinBasti:
The old pylint issue is back
```
************* Module ipaserver.install.installutils
ipaserver/install/installutils.py:1209: [E1101(no-member), store_version] Module 'ipaplatform' has no 'NAME' member)
ipaserver/install/installutils.py:1221: [E1101(no-member), check_version] Module 'ipaplatform' has no 'NAME' member)
ipaserver/install/installutils.py:1224: [E1101(no-member), check_version] Module 'ipaplatform' has no 'NAME' member)
```

Makefile.am
```diff
@@ -224,6 +223,15 @@ bdist_wheel: $(WHEELDISTDIR)
...
 
+wheel_placeholder: $(WHEELDISTDIR)
+	for dir in $(IPA_PLACEHOLDERS); do \
+	    $(MAKE) $(AM_MAKEFLAGS) -C $(top_srcdir)/pypi/$${dir} bdist_wheel || exit 1; \
+	done
+
+pypi_packages: bdist_wheel wheel_placeholder
+	@echo -e "\n\nTo upload packages to PyPI, run:\n"
+	@echo -e "    twine upload $(WHEELDISTDIR)/*-$(VERSION)-py2.py3-none-any.whl\n"
+
```
lslebodn: Is there a reason why top-level make need to be polluted with this new build targets?
They can be just in  `pypi/Makefile.am`.
It will simplify `IPA_PLACEHOLDERS`; you will iterate over all `SUBDIRS` there.

tiran:
Commit 1f8326aa fixes an issue in ```Makefile.python.am```. I think the issue caused ```ipaplatform``` and ```pypi/ipaplatform``` to cross streams.

Makefile.am
tiran: Yes, there are. If you look closely you will notice that the new target depends on bdist_wheel and wheel_placeholder and the placeholder parts creates wheels in dist/wheels. It doesn't maek sense to have the target just in pypi.

My first version iterated over subdirs in ./pypi. This doesn't work for ipacommands (next PR) because ipacommands must be built as sdist, not bdist_wheel.

lslebodn: You can move all following make targets `bdist_wheel wheel_bundle wheel_placeholder pypi_packages` into `pypi/Makefile.am`
(including all dependencies.

The top level make file will be simpler. The same as before wheel/pypy effort.
And *Makefile.am* in sub directory *pypy* will be python only.

tiran: I respectfully decline your suggestion.

* It's much more important to make package building and testing easy to use than to keep the top level Makefile simple.
* The top level Makefile is the public interface. Sublevel makefiles are helpers. I don't like to have targets in sublevel makefiles that are not usable from the top level Makefile.
* The targets are easily discoverable, .e.g. with ```make <tab>```.

**```</bikeshedding>```**

Makefile.python.am
```diff
@@ -20,7 +19,7 @@ all-local: $(top_builddir)/ipasetup.py
 		    --build-base "$(abs_builddir)/build"
 
 install-exec-local: $(top_builddir)/ipasetup.py
-	if [ "x$(pkginstall)" = "xtrue" ]; then \
+	if [ "x$(pkginstall)" != "xfalse" ]; then \
```
MartinBasti: I miss explanation why in commit message

Makefile.am
MartinBasti: I like to have `bdist_wheel` accessible from the top makefile too

lslebodn: that could be solved with
```
bdist_wheet:
    $(MAKE) -C $(top_builddir)/pypi
```

It would make the main Makefile nicer+simpler but that's just nice to have

The main problem is not in this PR but in structure of directories in freeipa git.
There are mixed directories for Python code/C-code/javascript code.
As a result of this there are many *.*SUBDIR* variables in top level makefile to workaround this directory structure.

I thought that moving some *.*SUBDIR* variables to `pypi/Makefile.am` would simplify it.

 Sorry for a little noise.

Makefile.python.am
tiran: > Default to pkginstall=true without duplicated definitions

automake was complaining about duplicated definition. I dropped ```pkginstall=true``` from ```Makefile.python.am``` and only set it in special ```Makefile.am```

MartinBasti:
master:

* 2e784336b0fe99baa47cf3e024f744ed56dc12ec Packaging: Add placeholder packages
* e2b9ea2fd58b98edbb8d6aec97aadeea7cf11dcb Add python-wheel as build requirement
* acdd1f59782bb836d6c4c255689918368adb8dab Add placeholders for ipaplatform, ipaserver and ipatests
* b4c1bf1c7d1a63e802abe6334bd1112d2d468513 Add with_wheels global to install wheel and PyPI packaging dependencies
* ab9f42d6eeefeaca2e4a5a9acfbb07b428be4616 Python build: use --build-base everywhere
* 60cfacc54167b7b94b63874ade62740d980e3746 pylint: ignore pypi placeholders
* bc1f60b3ba74032cb0895e154e02971aa380a6b3 Default to pkginstall=true without duplicated definitions

PR #473, opened 2017-02-16
Fix session/cookie related issues introduced with the privilege separation patches

simo5:
Fixes two bugs opened recently about double cookies being returned and ccache removal


abbra:
LGTM

ipaserver/plugins/session.py
```diff
@@ -21,7 +20,9 @@ def execute(self, *args, **options):
         ccache_name = getattr(context, 'ccache_name', None)
         if ccache_name is None:
             self.debug('session logout command: no ccache_name found')
+        else:
+            delattr(context, 'ccache_name')
 
-        logout(ccache_name)
+        setattr(context, 'logout_cookie', '')
```
tiran: why not ```context.logout_cookie = ''```? setattr is only needed when the first or second argument is variable.

simo5: for consistency, all manipulations of context happen via setattr/getattr/delattr

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/b895f4a34bcbd0b1787d2bfc1db25f34c3584b9c
https://fedorahosted.org/freeipa/changeset/d0642bfa55e9c24429675f623bc0e35824bc9fb0

PR #476, opened 2017-02-17
vault: cache the transport certificate on client

HonzaCholasta:
https://fedorahosted.org/freeipa/ticket/6652

### DIFF
--- a/ipaclient/plugins/vault.py
+++ b/ipaclient/plugins/vault.py
@@ -20,30 +20,38 @@
 from __future__ import print_function
 
 import base64
+import collections
+import errno
 import getpass
 import io
 import json
 import os
 import sys
+import tempfile
 
 from cryptography.fernet import Fernet, InvalidToken
 from cryptography.hazmat.backends import default_backend
-from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
 from cryptography.hazmat.primitives.asymmetric import padding
 from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
 from cryptography.hazmat.primitives.padding import PKCS7
 from cryptography.hazmat.primitives.serialization import (
     load_pem_public_key, load_pem_private_key)
-from cryptography.x509 import load_der_x509_certificate
 
 from ipaclient.frontend import MethodOverride
+from ipalib import x509
+from ipalib.constants import USER_CACHE_PATH
 from ipalib.frontend import Local, Method, Object
 from ipalib.util import classproperty
 from ipalib import api, errors
 from ipalib import Bytes, Flag, Str
 from ipalib.plugable import Registry
 from ipalib import _
+from ipapython.dnsutil import DNSName
+from ipapython.ipa_log_manager import log_mgr
+
+logger = log_mgr.get_logger(__name__)
 
 
 def validated_read(argname, filename, mode='r', encoding=None):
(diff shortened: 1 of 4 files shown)


ipaclient/plugins/vault.py
```diff
@@ -813,7 +861,46 @@ def forward(self, *args, **options):
...
+                transport_cert = get_cached_transport_cert(api.env.domain)
+            except KeyError:
+                config = self.api.Command.vaultconfig_show()['result']
+                transport_cert_der = config['transport_cert']
+                transport_cert = (
+                    x509.load_certificate(transport_cert_der, x509.DER))
+                set_cached_transport_cert(api.env.domain, transport_cert)
+            else:
+                transport_cert_der = (
+                    transport_cert.public_bytes(serialization.Encoding.DER))
```
tiran: Please move this functionality to a plugin or some other place. I like to fetch and cache the KRA transport cert every time Custodia is started. The API should have an optional argument to force a refresh.

ipaclient/plugins/vault.py
```diff
@@ -169,6 +180,58 @@ def decrypt(data, symmetric_key=None, private_key=None):
...
+    filename = os.path.join(TRANSPORT_CERT_CACHE_PATH, basename)
+
+    try:
+        try:
+            return x509.load_certificate_from_file(filename)
+        except EnvironmentError as e:
+            if e.errno != errno.ENOENT:
+                raise
+    except Exception as e:
+        logger.warning("Failed to load %s: %s", filename, e)
```
tiran: Please use ```exc_info=True``` everywhere in order to get a full traceback. It makes debugging much more pleasant.

ipaclient/plugins/vault.py
```diff
@@ -928,29 +1015,51 @@ def forward(self, *args, **options):
...
-        response = self.api.Command.vault_retrieve_internal(*args, **options)
+            try:
+                response = self.api.Command.vault_retrieve_internal(*args,
+                                                                    **options)
+            except errors.NotFound:
+                raise
+            except (errors.InternalError,
+                    errors.ExecutionError,
+                    errors.GenericError):
+                del_cached_transport_cert(api.env.domain)
```
tiran: This removes the cached key in any case of error. That's OK for a PoC. For the final version we should be more clever and only delete the certificate when it is not recognized by the KRA. I'll figure out the exact error message.

tiran: KRA subsystem doesn't report invalid KRA transport cert correctly, https://fedorahosted.org/pki/ticket/2597

HonzaCholasta: The cache is deleted only in case of an unexpected error, which is why there is a list of them, not in case of any error.

The KRA bug should definitely be fixed, but we should still cache the cert properly even for older servers.

ipaclient/plugins/vault.py
HonzaCholasta: This is not what was requested in the ticket...

ipaclient/plugins/vault.py
HonzaCholasta: OK.

ipaclient/plugins/vault.py
HonzaCholasta: Anyway, what's the point? Is there a design or PoC for Custodia that I can take a look at?

tiran: It was obvious to me that the certificate is retrieved and cached in the main process at start-up. It's faster and avoids so many issues with concurrent reads and writes in concurrent processes. In fact it was so obvious to me that I did not consider any other approach when I wrote the ticket.

HonzaCholasta: Added in-memory caching in the last update.

ipaclient/plugins/vault.py
```diff
@@ -568,14 +580,119 @@ def forward(self, *args, **options):
...
+
+        name = self.name + '_internal'
+        try:
+            return self.api.Command[name](*args, **options)
+        except errors.NotFound:
+            raise
+        except (errors.InternalError,
+                errors.ExecutionError,
+                errors.GenericError):
+            raise _TransportCertInvalid(sys.exc_info())
```
tiran: ```sys.exc_info()``` is problematic. It returns a stack frame object, which usually creates a reference cycle that keeps all local objects alive. If any object has ```__del__``` method, then Python 2 is unable to break the reference cycle. It's rather bad for long running processes because it is a kind of memory leak.

Here are two reasonable well written blog postings that I just found
* http://engineering.hearsaysocial.com/2013/06/16/circular-references-in-python/
* https://cosmicpercolator.com/2016/01/13/exception-leaks-in-python-2-and-3/

Proposal:

```
class _TransportCertInvalid(Exception):
    def __init__(self, exception):
        self.inner = exception
```

and

```
        except (errors.InternalError,
                     errors.ExecutionError,
                     errors.GenericError) as e:
            root_logger.debug("some message", exc_info=True)
            raise _TransportCertInvalid(e)
```

later:

```
six.reraise(type(e.inner), e.inner, None)
```

ipaclient/plugins/vault.py
```diff
@@ -568,14 +580,119 @@ def forward(self, *args, **options):
...
+
+    def internal(self, mechanism, session_key, *args, **options):
+        """
+        Calls the internal counterpart of the command.
+        """
+        domain = self.api.env.domain
+        dirname = TRANSPORT_CERT_CACHE_PATH
+        basename = DNSName(domain).ToASCII() + '.pem'
```
tiran: What do you think about ```DNSName(domain).ToASCII() + '-kra-transport.pem'``` to make the purpose of the cert clear?

ipaclient/plugins/vault.py
```diff
@@ -568,14 +580,119 @@ def forward(self, *args, **options):
...
+        if transport_cert_der is not None:
+            try:
+                return self._do_internal(mechanism,
+                                         session_key,
+                                         transport_cert_der,
+                                         *args, **options)
+            except _TransportCertInvalid:
+                _transport_cert_cache.pop(domain, None)
+                try:
+                    os.remove(filename)
```
tiran: Minor nit pick: ```os.unlink```. Unix developers don't remove files. We unlink names from inodes. :)

HonzaCholasta: Right 😄

ipaclient/plugins/vault.py
HonzaCholasta: Seems a bit redundant to me, since the full path is `~/.cache/ipa/kra-transport-certs/$DOMAIN.pem`.

ipaclient/plugins/vault.py
HonzaCholasta: There is no local variable which would hold the reference to the `sys.exc_info()` result or the `_TransportCertInvalid` exception. Is raising the exception enough to create a reference to it?

ipaclient/plugins/vault.py
tiran: An in-memory cache does not work for Custodia. It has to be an API call that ignores any caches (file) and downloads the KRA transport cert from remote server.

tiran:
I didn't see your comment regarding the in-memory cache because github hid the section after your most recent push.

The in-memory cache doesn't work for Custodia because Custodia is a forking webserver. Requests are handled in one-shot client processes. I must be able to forcefully download the certificate in the main process, before it starts listening on incoming requests.

HonzaCholasta:
Calling `vaultconfig_show` now refreshes the cache, you can call it to download and cache the certificate in the main process.

MartinBasti:
needs rebase

MartinBasti:
Works for me

MartinBasti:
master:

* 98bb5397c535e5e1a6c5ade9f0fb918be1d282c3 vault: cache the transport certificate on client

PR #477, opened 2017-02-17
Use RemoveOnStop to cleanup systemd sockets

npmccallum:


npmccallum:
This PR supersedes PR #469.

tiran:
RemoveonStop was added in systemd-214. Let me figure which version is on RHEL.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/d05d1115e409962fee3576a4bfc5cecfacef4fd3

PR #478, opened 2017-02-17
[4.4] Do not configure PKI ajp redirection to use "::1"

flo-renaud:
When ipa-server-install configures PKI, it provides a configuration file
with the parameter pki_ajp_host set to ::1. This parameter is used to configure
Tomcat redirection in /etc/pki/pki-tomcat/server.xml:
    <Connector port="8009"
        protocol="AJP/1.3"
        redirectPort="8443"
        address="::1" />
ie all requests to port 8009 are redirected to port 8443 on address ::1.

If the /etc/hosts config file does not define ::1 for localhost, then AJP
redirection fails and replica install is not able to request a certificate
for the replica.

Since PKI has been fixed (see PKI ticket 2570) to configure by default the AJP
redirection with "localhost", FreeIPA does not need any more to override
this setting.
The code now depends on pki 10.3.5-11 which provides the fix in the template
and the upgrade.

https://fedorahosted.org/freeipa/ticket/6575

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -161,8 +161,8 @@ Requires(post): systemd-units
 Requires: selinux-policy >= %{selinux_policy_version}
 Requires(post): selinux-policy-base >= %{selinux_policy_version}
 Requires: slapi-nis >= %{slapi_nis_version}
-Requires: pki-ca >= 10.3.5-6
-Requires: pki-kra >= 10.3.5-6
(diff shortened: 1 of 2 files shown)


martbab:
Fixed upstream
ipa-4-4:
https://fedorahosted.org/freeipa/changeset/4a30e9d53475d60fb76242a098f1d969d6b19f75

PR #479, opened 2017-02-17
Merge AD trust installer into composite ones

martbab:
This PR implements setup of Samba/Winbind as a part of server/replica install.
I will update installation tests in a separate PR in order not to inflate an
already sizeable amount of code touched in this one.

I also updated man pages of ipa-server/replica-install, but since the entries
are a bit chatty, it may be a good idea to write a more terse option
descriptions and provide a link to `ipa-adtrust-install` for more thorough
explanation.

The commits from https://github.com/freeipa/freeipa/pull/457 are on the bottom
of the branch in order to provide working AD trust installer in cases where
admin password is not provided.


martbab:
Bump for review.

martbab:
I have added a basic integration tests for the built-in AD trust installation, you can run them on 3 machines (master + 2 replicas) by running 
```bash
# ipa-run-tests --verbose /usr/lib/python2.7/site-packages/ipatests/test_integration/test_installation.py -k TestADTrustInstall
```
and having a properly configured test config.

ipaserver/install/server/__init__.py
```diff
@@ -331,6 +336,13 @@ def dirsrv_config_file(self, value):
     )
     pkinit_cert_name = prepare_only(pkinit_cert_name)
 
+    # TODO: --add-agents makes no sense on server install (since it is a first
+    # master
```
HonzaCholasta: That's already resolved, isn't it?

ipaserver/install/server/__init__.py
```diff
@@ -548,6 +560,10 @@ def dm_password(self, value):
     def admin_password(self, value):
         validate_admin_password(value)
 
+    # always run sidgen task and do not allow adding agents on first master
+    add_sids = True
```
HonzaCholasta: AFAICT `add_sids` is only relevant when upgrading from an old server which did not have the `sidgen` plugin enabled, so it in fact should always be `False` on first master.

martbab: The problem is that (for some reason) the 389-ds is populated by several default groups during bootstrap. One of these groups is `cn=editors` which is POSIX one has no SID assigned when I hardcoded `add_sids` to False (`cn=admins` have SID assinged directly by CIFS installer since it has to have special RID of 512).

@rcritten is there any reason why are these groups created during bootstrap and not during LDAP update phase?

MartinBasti:
Works for me, except,  ipa-server-install --setup-adtrust works even without freeipa-server-trust-ad package. Please fix this in a new PR in way how DNS is done.

martbab:
I have noticed that the check for installed dependencies is buggy, I will have to fix it before pushing.

Also we would need to move the 'editors' group addition to the LDAP update phase since it remains with missing SID during ipa-server-install when `add_sids` knob is set to False. @abbra @rcritten is that ok with you? Please see inline comment for more details.

abbra:
If you can differentiate how the installer is being run, then for composite installer always run add_sids.

martbab:
@abbra I think that I am confused by the way sidgen plugin works. During LDAP configuration I can see that sidgen/extdom plugins are activated. e.g:

```
...
  [43/47]: enabling compatibility plugin
  [44/47]: activating sidgen plugin
  [45/47]: activating extdom plugin
...
```
Yet unless I install AD trust related bits, there are no SIDs generated on entries I am added (user or groups). When the AD trust installer is run, I see that the sidgen task is activated:

```
...
 [13/21]: activating sidgen task
 [14/21]: configuring smbd to start on boot
...
```

The admin user now has SID added by installer, yet the existing POSIX groups (editors) have no SIDs associated with them, only the new user I add afterwards.

Do we have a documentation about the semantics of different sidgen-related operations somewhere? If not, can you please explain the behavior I am seeing here?

abbra:
Unless you specified --add-sids to ipa-adtrust-install (or `add_sids=True` in ADTrustInstance.setup() call), no task would be run. 'Activating sidgen task' only adds configuration to allow the task to be run.

martbab:
OK I will then hard-code `add_sids=True` in ipa-server-install

martbab:
I have added a commit that fixes the choeck for missing dependencies in composite installers.

MartinBasti:
Please rebase

PR #480, opened 2017-02-18
Hide request_type doc string in cert-request help

Akasurde:
Fix hides description of request_type argument in cert-request
command help

Fixes https://pagure.io/freeipa/issue/6494
Fixes https://pagure.io/freeipa/issue/5734

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -479,7 +479,7 @@ class certreq(BaseCertObject):
             'request_type',
             default=u'pkcs10',
             autofill=True,
-            flags={'no_update', 'no_update', 'no_search'},
+            flags={'no_option', 'no_update', 'no_update', 'no_search'},
         ),
         Str(
             'profile_id?', validate_profile_id,


MartinBasti:
Ticket is `Enumerate all available request type options in ipa cert-request help` but your commit doesn't enumerate all possible certtypes

frasertweedale:
I would like to NACK this.  We instead want to hide or remove the option, because
we only support PKCS #10 and this is unlikely to change any time soon.

There is already a ticket for that: https://fedorahosted.org/freeipa/ticket/5734

Akasurde:
@frasertweedale What do you recommend to hide this option ? does removing this option has detrimental effect on `cert-request` command ?

frasertweedale:
@Akasurde if we just want to hide it, I think you use a client override for the `cert_request` command
and filter out the option.  @HonzaCholasta can confirm.

OTOH if we just want to remove it altogether, that is straightforward.  It will break any clients that
explicitly pass the option.  I suspect it's unlikely that there are such clients out there, but we cannot
know for sure, so as much as I'd like to remove it, I'm hesitant.

Akasurde:
@HonzaCholasta ping

HonzaCholasta:
I agree with @frasertweedale, but we can't remove the option from the server plugin altogether, as pre-4.4 clients always send it. You can hide the option by adding the `no_option` flag in the param definition.

Akasurde:
@HonzaCholasta Thanks I will use `no_option` flag and update PR.

Akasurde:
Bump for review.

MartinBasti:
master:

* a1bb442054936113369a88b49483e914664712e7 Hide request_type doc string in cert-request help


ipa-4-5:

* 535e8610c556ab1a0eb83e9798e7e182355d8396 Hide request_type doc string in cert-request help

Akasurde:
@MartinBasti @frasertweedale  @HonzaCholasta Thanks

MartinBasti:
@Akasurde you are welcome

PR #482, opened 2017-02-20
Remove MD5 certificate fingerprints

stlaz:
To be "backward compatible" we cannot remove `md5_fingerprint` so we at least supply the reason why it can't be counted.

edit: For this to work https://github.com/freeipa/freeipa/pull/437 needs to be pushed first.


rcritten:
In service.py the error isn't wrapped in _(). You should use the same message in both.

Given the different messages I'm surprised this didn't pop up as a test failure.

MartinBasti:
I don't think that this is a good way how to handle backward compatibility. With FIPS mode enabled there is no md5 backward compatibility and users should adapt their automation. In case that  IPA API is used directly it will contain a garbage and it may not be catched faster enough by any automation on user side. We should not provide anything related to md5 under FIPS mode and let any possible automation using IPA API to fail early on missing values.

tomaskrizek:
@rcritten Currently, the tests fail because we need #437 merged. It would be caught.

@MartinBasti The only other option I see is to provide `None`. We can't remove the md5 fingerprint from API - or can we?

stlaz:
I am fine with not providing `md5_fingerprint` at all but that would require the tests to be fixed as well and I am not sure how to easily do that in this case.

tomaskrizek:
Actually, we don't need to provide `md5_fingerprint` at all in FIPS, since the attribute is marked as `vritual_attribute`.

stlaz:
@rcritten thanks for noticing the discrepancy in the previous version of the commit, it was a leftover from previous implementation.
I reworked the commit to remove MD5 certificate fingerprints completely, leaving just SHA1. Is a requirement to accept this patch to announce this on freeipa-devel/users?

tomaskrizek:
@stlaz I think it'd be good to discuss this change on freeipa-devel. Also, since we're removing md5, I'd consider adding sha256.

MartinBasti:
+1 for sha256

tomaskrizek:
Btw, I think sha256 can be added in a separate PR. Let's just wait if there are any concerns about removing md5 on the freeipa-devel.

stlaz:
I don't have a problem adding it in this PR, it needs to be resolved according to the outcome of the freeipa-devel thread anyway.

tomaskrizek:
ACK, there is no disagreement on the freeipa-devel.

I'm already working on replacing SHA1 with SHA256 given the [recent events](https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html).

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/e2d1b21c5049f68d0336dcaf3f8657b214a34e2b

PR #483, opened 2017-02-20
lite-server: validate LDAP connection and cache schema

tiran:
The LDAP schema cache makes the lite-server behave more like mod_wsgi.

See https://fedorahosted.org/freeipa/ticket/6679

### DIFF
--- a/contrib/lite-server.py
+++ b/contrib/lite-server.py
@@ -51,10 +51,12 @@ import os
 import optparse  # pylint: disable=deprecated-module
 import ssl
 import sys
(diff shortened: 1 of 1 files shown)


MartinBasti:
Code looks good to me, as far as this is just code for developers I assume that @tiran tested it enough, so ACK.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/dcb618152572ca013a447336e13d24399b5f7960

PR #485, opened 2017-02-20
Fix session logout

simo5:
There were 2 issues with session logouts, one is that the logout_cookie
was checked and acted on in the wrong place, the other is that the wrong
value was set in the IPASESSION header.

Fixes https://fedorahosted.org/freeipa/ticket/6685

### DIFF
--- a/ipaserver/plugins/session.py
+++ b/ipaserver/plugins/session.py
@@ -23,6 +23,6 @@ class session_logout(Command):
         else:
             delattr(context, 'ccache_name')
 
-        setattr(context, 'logout_cookie', '')
+        setattr(context, 'logout_cookie', 'MagBearerToken=')
 
         return dict(result=None)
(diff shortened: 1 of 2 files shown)


ipaserver/plugins/session.py
```diff
@@ -23,6 +23,6 @@ def execute(self, *args, **options):
         else:
             delattr(context, 'ccache_name')
 
-        setattr(context, 'logout_cookie', '')
+        setattr(context, 'logout_cookie', 'MagBearerToken=')
```
simo5: MagBearerToken is the value that matter sin the cookie generated by mod_auth_gssapi. Since the privsep partches removed our session code and use mod_auth_gssapi + mod_session instead, then that's what we need to use.

This patch is a fix for 6685, the part about blacklist is just a suggestion for an aditional improvement but it is not required to fix the issue.

stlaz: Thank you for the explanation.

stlaz:
Works as expected.

martbab:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/908d2eaba46f5f123b49af400a8b696545c62b54

PR #486, opened 2017-02-20
Migrate OTP import script to python-cryptography

npmccallum:
https://fedorahosted.org/freeipa/ticket/5192

### DIFF
--- a/ipaserver/install/ipa_otptoken_import.py
+++ b/ipaserver/install/ipa_otptoken_import.py
@@ -29,10 +29,14 @@ import struct
 from lxml import etree
 import dateutil.parser
 import dateutil.tz
-import nss.nss as nss
 import gssapi
 import six
 from six.moves import xrange
+
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.kdf import pbkdf2
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+from cryptography.hazmat.backends import default_backend
 
 from ipapython import admintool
 from ipalib import api, errors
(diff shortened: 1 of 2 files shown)


npmccallum:
This is an old patch I found on my system that doesn't appear to be merged.

tiran:
Thanks Indiana Nathaniel, good code archaeology. The ticket aligns nicely with https://fedorahosted.org/freeipa/ticket/6650

ipaserver/install/ipa_otptoken_import.py
```diff
@@ -229,27 +212,32 @@ class XMLDecryptor(object):
...
-        slot = nss.get_best_slot(mech)
-        key = nss.import_sym_key(slot, mech, nss.PK11_OriginUnwrap, nss.CKA_ENCRYPT, self.__key)
-        iv = nss.param_from_iv(mech, nss.SecItem(data[0:ivlen//8]))
-        ctx = nss.create_context_by_sym_key(mech, nss.CKA_DECRYPT, key, iv)
-        out = ctx.cipher_op(data[ivlen // 8:])
-        out += ctx.digest_final()
+        iv = data[:algo.block_size / 8]
+        data = data[len(iv):]
+
+        cipher = Cipher(algo(self.__key), mode(iv))
```
stlaz: You're missing `backend` here.

ipaserver/install/ipa_otptoken_import.py
```diff
@@ -169,50 +173,29 @@ def __init__(self, enckey):
...
-        for i in range(1, blocks + 1):
-            # Set initial values.
-            last = self.salt + struct.pack('>I', i)
-            hash = [0] * mac.digest_size
+        self.kdf = pbkdf2.PBKDF2HMAC(
+            algorithm=hmod,
+            length=klen,
+            salt=salt,
+            iterations=itrs
+        )
```
stlaz: `backend` is missing here as well.

stlaz:
Thanks for the patch, less `nss` is always good. It seems that python-cryptography might have added the `backend` attribute to some constructors since the patch was created, our tests found two of such spots, if you could perhaps add it there.
I personally don't care much for the pep8 errors, IMHO the code reads better this way.

stlaz:
Fixed in https://github.com/freeipa/freeipa/issues/509

PR #488, opened 2017-02-20
Speed up client schema cache

tiran:
It's inefficient to open a zip file over and over again. By loading all
members of the schema cache file at once, the ipa CLI script starts
about 25 to 30% faster for simple cases like help and ping.

Before:

```
$ time for i in {1..20}; do ./ipa ping >/dev/null; done

real    0m13.608s
user    0m10.316s
sys     0m1.121s
```

After:

```
$ time for i in {1..20}; do ./ipa ping >/dev/null; done

real    0m9.330s
user    0m7.635s
sys     0m1.146s
```

https://fedorahosted.org/freeipa/ticket/6690

### DIFF
--- a/ipaclient/remote_plugins/schema.py
+++ b/ipaclient/remote_plugins/schema.py
@@ -3,13 +3,11 @@
 #
 
 import collections
-import contextlib
 import errno
-import fcntl
-import io
 import json
 import os
 import sys
+import tempfile
 import types
 import zipfile
 
@@ -374,7 +372,6 @@ class Schema(object):
         self._dict = {}
         self._namespaces = {}
         self._help = None
-        self._file = six.BytesIO()
 
         for ns in self.namespaces:
             self._dict[ns] = {}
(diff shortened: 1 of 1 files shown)


dkupka:
@tiran Currently the file is first copied into BytesIO and then all reading is done from it. Your modification IMO supersedes the need for the BytesIO copy because everything is read into memory at once. Could you remove it?

tiran:
@dkupka Makes sense, I dropped the temporary buffer and replaced the file locking logic with tempfile + os.rename.

dkupka:
The speedup I see is smaller (10-15%) [1] than what you're reporting but that might be caused by the fact that I store the cache on really slow file system (NFS mount). Anyway the changes makes sense, thanks.

[1] https://da.gd/7r75A

tiran:
It looks like your IPA server is about half as fast (26sec / 13sec for 20 pings). In absolute numbers, it's still ~2.5 sec faster. In your case, performance probably dominated by server latency.

MartinBasti:
master:

* 332dbab1ff09eb719eb9e0a7a90bbf5b6e69ddc9 Speed up client schema cache
* 3be696c92f6948ea0ced9784920600b73703e414 Drop in-memory copy of schema zip file

PR #489, opened 2017-02-21
Fix error in ca_cert_files validator

stlaz:
ClientInstall expects a single ca_cert_file as a string but the
framework gives it a list.

https://fedorahosted.org/freeipa/ticket/6694

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -3555,6 +3555,10 @@ class ClientInstall(ClientInstallInterface,
 
     @ca_cert_files.validator
     def ca_cert_files(self, value):
+        if not isinstance(value, list):
+            raise ValueError("Expected list, got {!r}".format(value))
+        # this is what init() does
+        value = value[-1]
         if not os.path.exists(value):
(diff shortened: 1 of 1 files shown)


ipaclient/install/client.py
```diff
@@ -3555,6 +3555,10 @@ class ClientInstall(ClientInstallInterface,
 
     @ca_cert_files.validator
     def ca_cert_files(self, value):
+        if not isinstance(value, list):
```
tiran: Does it makes sense to check for len(value) == 1, too?

```
if not isinstance(value, list) or len(value) != 1:
    raise ValueError("Expected list of length 1, got {!r}".format(value))
```

tiran:
tentative ack, see comment

ipaclient/install/client.py
stlaz: This check is performed in the option parser itself.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/0fffeabe0249d9c3c11e522fccf22ddeb1197b64

PR #490, opened 2017-02-21
certdb: use certutil and match_hostname for cert verification

HonzaCholasta:
Use certutil and ssl.match_hostname calls instead of python-nss for
certificate verification.

### DIFF
--- a/ipalib/x509.py
+++ b/ipalib/x509.py
@@ -35,6 +35,7 @@ from __future__ import print_function
 import binascii
 import datetime
 import ipaddress
+import ssl
 import base64
 import re
 
@@ -49,6 +50,7 @@ from ipalib import api
 from ipalib import util
 from ipalib import errors
 from ipapython.dn import DN
+from ipapython.dnsutil import DNSName
 
 if six.PY3:
     unicode = str
@@ -406,6 +408,27 @@ def process_othernames(gns):
             yield gn
 
 
+def _pyasn1_get_san_general_names(cert):
+    tbs = decoder.decode(
+        cert.tbs_certificate_bytes,
+        asn1Spec=rfc2459.TBSCertificate()
+    )[0]
+    OID_SAN = univ.ObjectIdentifier('2.5.29.17')
+    # One would expect KeyError or empty iterable when the key ('extensions'
+    # in this particular case) is not pressent in the certificate but pyasn1
+    # returns None here
+    extensions = tbs['extensions'] or []
+    gns = []
+    for ext in extensions:
+        if ext['extnID'] == OID_SAN:
+            der = decoder.decode(
+                ext['extnValue'], asn1Spec=univ.OctetString())[0]
+            gns = decoder.decode(der, asn1Spec=rfc2459.SubjectAltName())[0]
+            break
+
+    return gns
+
+
 def get_san_general_names(cert):
     """
     Return SAN general names from a python-cryptography
@@ -430,22 +453,7 @@ def get_san_general_names(cert):
     and should go away.
 
     """
-    tbs = decoder.decode(
-        cert.tbs_certificate_bytes,
-        asn1Spec=rfc2459.TBSCertificate()
-    )[0]
-    OID_SAN = univ.ObjectIdentifier('2.5.29.17')
-    # One would expect KeyError or empty iterable when the key ('extensions'
-    # in this particular case) is not pressent in the certificate but pyasn1
-    # returns None here
-    extensions = tbs['extensions'] or []
-    gns = []
-    for ext in extensions:
-        if ext['extnID'] == OID_SAN:
-            der = decoder.decode(
-                ext['extnValue'], asn1Spec=univ.OctetString())[0]
-            gns = decoder.decode(der, asn1Spec=rfc2459.SubjectAltName())[0]
-            break
+    gns = _pyasn1_get_san_general_names(cert)
 
     GENERAL_NAME_CONSTRUCTORS = {
         'rfc822Name': lambda x: cryptography.x509.RFC822Name(unicode(x)),
@@ -504,6 +512,17 @@ def _pyasn1_to_cryptography_oid(oid):
     return cryptography.x509.ObjectIdentifier(str(oid))
 
 
+def get_san_a_label_dns_names(cert):
+    gns = _pyasn1_get_san_general_names(cert)
+    result = []
+
+    for gn in gns:
+        if gn.getName() == 'dNSName':
+            result.append(unicode(gn.getComponent()))
+
+    return result
+
+
 def chunk(size, s):
     """Yield chunks of the specified size from the given string.
 
@@ -543,3 +562,23 @@ def format_datetime(t):
     if t.tzinfo is None:
         t = t.replace(tzinfo=UTC())
     return unicode(t.strftime("%a %b %d %H:%M:%S %Y %Z"))
+
+
+def match_hostname(cert, hostname):
+    match_cert = {}
+
+    match_cert['subject'] = match_subject = []
+    for rdn in cert.subject.rdns:
+        match_rdn = []
+        for ava in rdn:
+            if ava.oid == cryptography.x509.oid.NameOID.COMMON_NAME:
+                match_rdn.append(('commonName', ava.value))
+        match_subject.append(match_rdn)
+
+    values = get_san_a_label_dns_names(cert)
+    if values:
+        match_cert['subjectAltName'] = match_san = []
+        for value in values:
+            match_san.append(('DNS', value))
+
+    ssl.match_hostname(match_cert, DNSName(hostname).ToASCII())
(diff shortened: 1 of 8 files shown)


ipalib/x509.py
```diff
@@ -543,3 +544,35 @@ def format_datetime(t):
...
+    match_cert['subject'] = match_subject = []
+    for rdn in cert.subject.rdns:
+        match_rdn = []
+        for ava in rdn:
+            if ava.oid == cryptography.x509.oid.NameOID.COMMON_NAME:
+                name = 'commonName'
+            else:
+                name = ava.dotted_string
+            match_ava = (name, ava.value)
+            match_rdn.append(match_ava)
```
tiran: Surplus ```()``` in ```match_ava = (name, ava.value)```. How about you collapse both lines into one? ``` match_rdn.append((name, ava.value))```.

ipalib/x509.py
```diff
@@ -543,3 +544,35 @@ def format_datetime(t):
...
+
+    try:
+        san = cert.extensions.get_extension_for_class(
+                cryptography.x509.SubjectAlternativeName)
+    except cryptography.x509.ExtensionNotFound:
+        pass
+    else:
+        match_cert['subjectAltName'] = match_san = []
+        for value in san.get_values_for_type(cryptography.x509.DNSName):
+            match_dnsname = ('DNS', str(value))
```
tiran: Why do you use ```str()``` here? It's either unnecessary (if value is already a str) or wrong (if value is bytes or DNSName instance). Since ```get_values_for_type``` returns the value attribute of DNSName instances and DNSName.value is guaranteed to be text, the call is not necessary.

ipalib/x509.py
```diff
@@ -543,3 +544,35 @@ def format_datetime(t):
...
+            match_san.append(match_dnsname)
+
+    try:
+        ssl.match_hostname(match_cert, hostname)
+    except ssl.SSLError as e:
```
tiran: match_hostname raises a subclass of ```ValueError```. It's part of the API. There is no need to catch ```SSLError```.

ipalib/x509.py
```diff
@@ -543,3 +544,35 @@ def format_datetime(t):
...
+    match_cert = {}
+
+    match_cert['subject'] = match_subject = []
+    for rdn in cert.subject.rdns:
+        match_rdn = []
+        for ava in rdn:
+            if ava.oid == cryptography.x509.oid.NameOID.COMMON_NAME:
+                name = 'commonName'
+            else:
+                name = ava.dotted_string
```
tiran: You can safely omit other attribute value assertion. Python's match_hostname will both ignore other AVAs and never support AVAs OID in dotted string notation.

tiran:
Do we ensure that the function is always called with an IDN A-Label encoded hostname? ```ssl.match_hostname``` assumes that all parts are A-labels, not U-labels.

ipalib/x509.py
HonzaCholasta: Makes sense.

ipalib/x509.py
HonzaCholasta: I see.

ipalib/x509.py
HonzaCholasta: Sure, why not.

HonzaCholasta:
@tiran, how do I ensure that?

tiran:
The hostname must be ASCII text. Something like ```hostname.encode('ascii')``` should catch non-ASCII text and Python 3 bytes.

ipalib/x509.py
HonzaCholasta: I don't remember TBH - this patch has been laying on my HDD for quite a while.

ipalib/x509.py
```diff
@@ -543,3 +545,28 @@ def format_datetime(t):
...
+
+    try:
+        san = cert.extensions.get_extension_for_class(
+                cryptography.x509.SubjectAlternativeName)
+    except cryptography.x509.ExtensionNotFound:
+        pass
+    else:
+        match_cert['subjectAltName'] = match_san = []
+        for value in san.value.get_values_for_type(cryptography.x509.DNSName):
+            value = DNSName(value).ToASCII()
```
tiran: You must not our DNSName class here. It's an invalid and therefore unsafe conversion. I fear we cannot convert back safely. In general IDNA is not a bijective encoding. This is especially true for different kinds of IDNA. PyCA cryptography uses IDNA 2008 UTS46, DNSName uses IDNA 2003.

Consider this example:

```
>>> from ipapython.dnsutil import DNSName
>>> import idna
>>> asn1_string = 'xn--strae-oqa.de'
>>> pyca_dnsname = idna.decode(asn1_string)
>>> DNSName(pyca_dnsname).ToASCII()
u'strasse.de'
```

I talked to @reaperhulk about the issue. Paul pointed me to https://github.com/pyca/cryptography/issues/3357.

Yes, hostname verification is a mess. I got it wrong multiple times, caused at least one CVE and found multiple other CVEs. :/

HonzaCholasta:
I think this PR is ready now.

stlaz:
@tiran Could you please finish the review? I guess we can omit the change in `.spec.in` for the review time being.

stlaz:
I tried to use the wonderful github tool to resolve conflicts to make this more review-friendly but I guess it kind of missed the magic, it's ready for review anyway, please, finish it.

tiran:
github magic is bad magic :/ It still shows up as 'conflicting' for me.

I'll try to find time to review the issue tomorrow, Thursday latest.

ipalib/x509.py
```diff
@@ -543,3 +562,23 @@ def format_datetime(t):
...
+    for rdn in cert.subject.rdns:
+        match_rdn = []
+        for ava in rdn:
+            if ava.oid == cryptography.x509.oid.NameOID.COMMON_NAME:
+                match_rdn.append(('commonName', ava.value))
+        match_subject.append(match_rdn)
+
+    values = get_san_a_label_dns_names(cert)
+    if values:
+        match_cert['subjectAltName'] = match_san = []
```
tiran: As of RFC 6125 section 6.4.4. you also have to take SRV-ID and URI-ID into account. Since ssl.match_hostname supports IP addresses, you also need to include iPAddress fields or validate that hostname never looks like an IP address.

This will cause a security bug as soon as https://bugs.python.org/issue28196 is fixed.

ipalib/x509.py
```diff
@@ -406,6 +408,27 @@ def process_othernames(gns):
             yield gn
 
 
+def _pyasn1_get_san_general_names(cert):
+    tbs = decoder.decode(
+        cert.tbs_certificate_bytes,
+        asn1Spec=rfc2459.TBSCertificate()
+    )[0]
+    OID_SAN = univ.ObjectIdentifier('2.5.29.17')
```
tiran: Please use SUBJECT_ALTERNATIVE_NAME from cryptography.x509.oid.

tiran:
Your PR is going to remove the last import from python-nss. Awesome!

Please remove the requirement from ```ipapython/setup.py``` and ```freeipa.spec.in```, too.

ipalib/x509.py
```diff
@@ -543,3 +562,23 @@ def format_datetime(t):
     if t.tzinfo is None:
         t = t.replace(tzinfo=UTC())
     return unicode(t.strftime("%a %b %d %H:%M:%S %Y %Z"))
+
+
+def match_hostname(cert, hostname):
+    match_cert = {}
+
+    match_cert['subject'] = match_subject = []
+    for rdn in cert.subject.rdns:
```
HonzaCholasta: What RFC 5280 actually states is that the subject might be empty, not that the subject field is optional.

Empty subject has an empty list of RDNs:
```
>>> from cryptography.hazmat.backends import default_backend
>>> from cryptography.x509 import load_pem_x509_certificate
>>> load_pem_x509_certificate(open('no-subject.crt', 'rb').read(), default_backend()).subject.rdns
[]
```
i.e. no extra check is necessary.

ipalib/x509.py
HonzaCholasta: `ssl.match_hostname` in Python 2 supports only DNS SANs, so I didn't bother including support for anything else.

I guess adding support for other SAN types won't hurt, that is if we don't mind the different behavior in Python 2 vs 3.

HonzaCholasta:
Awesome indeed!

As for your suggestions to improve the validation, I completely agree with them, but the focus of this PR is to refactor the current validation not to use python-nss, which it delivers. Could you please file a ticket for the improvements, so that it gets more visibility and can be properly tracked?

MartinBasti:
master:

* 9183cf2a7505624235b255b1406702cdaa65bb38 certdb: use certutil and match_hostname for cert verification
* 2b33230f669ca22d6948a4a351b4c92ba15222ab setup, pylint, spec file: drop python-nss dependency

PR #491, opened 2017-02-21
Don't prepend option names with additional '--'

stlaz:
The options now have '--' prepended by their names already, don't
add it.

The issue example:
running `ipa-client-install --ca-cert-file /home/cartman/nonexistent_file` gives
```
Usage: ipa-client-install [options]

ipa-client-install: error: option ----ca-cert-file: '/home/slaznick/pokus' is not a valid certificate file
```


MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/9ac068ad04a2323192f9447986a3d1c5431f1e50

PR #113, opened 2016-09-24
ipalib.constants: Remove default domain, realm, basedn, xmlrpc_uri, ldap_uri

pspacek:
Domain, realm, basedn, xmlrpc_uri, ldap_uri do not have any reasonable default.
This patch removes hardcoded default so the so the code which depends
on these values blows up early and does not do crazy stuff
with default values instead of real ones.

This should help to uncover issues caused by improper ipalib
initialization.

It will surely break something but right now, at the beginning of devel cycle, is IMHO the right time to do change like this and to remove some old cruft.

PR #492, opened 2017-02-21
config: remove meaningless defaults

HonzaCholasta:
**ipalib.constants: Remove default domain, realm, basedn, xmlrpc_uri, ldap_uri**

Domain, realm, basedn, xmlrpc_uri, ldap_uri do not have any reasonable default.
This patch removes hardcoded default so the so the code which depends
on these values blows up early and does not do crazy stuff
with default values instead of real ones.

This should help to uncover issues caused by improper ipalib
initialization.

**config: provide defaults for `xmlrpc_uri`, `ldap_uri` and `basedn`**

Derive the default value of `xmlrpc_uri` and `ldap_uri` from `server`.
Derive the default value of `basedn` from `domain`.

This supersedes @pspacek's PR #113.

### DIFF
--- a/ipalib/config.py
+++ b/ipalib/config.py
@@ -579,6 +579,20 @@ class Env(object):
         if 'log' not in self:
             self.log = self._join('logdir', '%s.log' % self.context)
 
+        if 'basedn' not in self and 'domain' in self:
+            self.basedn = DN(*(('dc', dc) for dc in self.domain.split('.')))
+
+        # Derive xmlrpc_uri from server
+        # (Note that this is done before deriving jsonrpc_uri from xmlrpc_uri
+        # and server from jsonrpc_uri so that when only server or xmlrpc_uri
+        # is specified, all 3 keys have a value.)
+        if 'xmlrpc_uri' not in self and 'server' in self:
+            self.xmlrpc_uri = 'https://{}/ipa/xml'.format(self.server)
+
+        # Derive ldap_uri from server
+        if 'ldap_uri' not in self and 'server' in self:
+            self.ldap_uri = 'ldap://{}'.format(self.server)
+
         # Derive jsonrpc_uri from xmlrpc_uri
         if 'jsonrpc_uri' not in self:
             if 'xmlrpc_uri' in self:
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -134,8 +134,9 @@ DEFAULT_CONFIG = (
     # Following values do not have any reasonable default.
     # Do not initialize them so the code which depends on them blows up early
     # and does not do crazy stuff with default values instead of real ones.
+    # ('server', 'localhost'),
     # ('xmlrpc_uri', 'http://localhost:8888/ipa/xml'),
-    # jsonrpc_uri is set in Env._finalize_core()
+    # ('jsonrpc_uri', 'http://localhost:8888/ipa/json'),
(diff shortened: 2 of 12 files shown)


tiran:
https://github.com/HonzaCholasta/freeipa/blob/4ebf4b907213c9951eb9cbd276e0460552563fb1/ipalib/config.py#L579 initializes server from jsonrpc_uri. Does it make sense move this block before your new code?

HonzaCholasta:
@tiran, not really, the order does not matter here.

tiran:
It does matter. In the current version ```if 'server' not in self:``` is checked and ```self.server``` is checked a couple of lines after ```if 'ldap_uri' not in self and 'server' in self:```.

HonzaCholasta:
I stand corrected, but it does not make sense to reorder the code as you suggested anyway, as it would change the current default of `server` when only `xmlrpc_uri` is specified in the configuration from "use the hostname from `xmlrpc_uri`" do "do not set a default value".

tiran:
Can you add a comment to explain the order of checks and assignments? Without explanation, it's going to confuse the next poor developer.

HonzaCholasta:
Sure.

tiran:
It's probably easier to always define options like ```'ldap_uri``` but use ```None``` as default.

```
cd .; ./makeaci --validate
./makeaci: ipaserver/plugins/dogtag.py:244: ignoring ImportError: No module named backports_abc
./makeaci: ipaserver/plugins/dogtag.py:244: ignoring ImportError: No module named rnc2rng
  File "./makeaci", line 134, in <module>
    main(options)
  File "./makeaci", line 107, in main
    api.finalize()
  File "/freeipa/ipalib/plugable.py", line 747, in finalize
    self._get(plugin)
  File "/freeipa/ipalib/plugable.py", line 776, in _get
    instance = self.__instances[plugin] = plugin(self)
  File "/freeipa/ipaserver/plugins/ldap2.py", line 72, in __init__
    ldap_uri = api.env.ldap_uri
AttributeError: 'Env' object has no attribute 'ldap_uri'
Exception AttributeError: "'ldap2' object has no attribute
'id'" in <bound method ldap2.__del__ of ipaserver.plugins.ldap2.ldap2()> ignored
make: *** [acilint] Error 1
Makefile:1108: recipe for target 'acilint' failed
```

HonzaCholasta:
I took the hard way and removed the URI argument from `ldap2.__init__()`.

martbab:
Regardless of what branch status widget shows the PR seems to be in need of rebase.

MartinBasti:
master:

* e9cb74fd27f4015ad980781785d95bd4107b6f40 user, migration: use LDAPClient for ad-hoc LDAP connections
* 935fcaea2edc6bb189a6fe4f9aefc2998695b74d {ca,kra}instance: drop redundant URI argument from ad-hoc ldap2 connections
* 8f849a77f5b26e087a078376c2fb17db49d9cdcf test_ldap: drop redundant URI argument
* 4736fef6bb30bc07592dc5acaac05d193d660d5e ldap2: remove URI argument from ldap2 constructor
* 3f6411a49c49da7013341ff8feae3a63e75e0fbf ipalib.constants: Remove default domain, realm, basedn, xmlrpc_uri, ldap_uri
* ba3963b4dc445fe6979bf321617e8c397107bd02 config: provide defaults for `xmlrpc_uri`, `ldap_uri` and `basedn`

PR #493, opened 2017-02-21
Update Contributors.txt

mkosek:
Update mailmap with the new mistyped authors and generate a new
Contributors list.

### DIFF
--- a/Contributors.txt
+++ b/Contributors.txt
@@ -18,13 +18,13 @@ Developers:
 	Thierry Bordaz
 	Sumit Bose
 	François Cami
+	Petr Čech
 	Xiao-Long Chen
 	Jan Cholasta
 	Yuri Chornoivan
 	Brian Cook
 	Rob Crittenden
 	Frank Cusack
-	Petr Čech
 	Nalin Dahyabhai
 	Don Davis
 	John Dennis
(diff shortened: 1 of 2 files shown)


Contributors.txt
```diff
@@ -18,13 +18,13 @@ Developers:
 	Thierry Bordaz
 	Sumit Bose
 	François Cami
+	Petr Čech
```
tomaskrizek: Seems like the lexicographic ordering isn't quite right, `Č` should come after `C`

mkosek: When sorting, I use English collator:

```
collator = icu.Collator.createInstance(icu.Locale('en_US.UTF-8'))
authors.sort(cmp=collator.compare, key=lambda x: x.split()[-1].lower())
```

I could change it to Czech collator, but then for example surnames beginning with "Ch" are sorted after "H", which would confuse English speakers.

So I would rather stick with current working version (already applied to names like Špaček) to make the sorting clear for wide population.

stlaz:
I don't give two poops and a popsicle about the order of names in the Contributors.txt file. ACK.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/b367c3a622c8c8b96777e4ce50334d2a4477bbe7

MartinBasti:
If script for generating contributors file was published it may be documented in releasing page and it may be part of each release.

PR #495, opened 2017-02-22
Fix ipa-server-upgrade

stlaz:
I was to eager to ACK https://github.com/freeipa/freeipa/pull/471.

Running ipa-server-upgrade would fail to stop ipa_memcached if
it's already uninstalled.

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -83,7 +83,8 @@ def uninstall_ipa_memcached():
     """
     ipa_memcached = service.SimpleServiceInstance('ipa_memcached')
 
-    ipa_memcached.uninstall()
+    if ipa_memcached.is_configured():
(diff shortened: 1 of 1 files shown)


tiran:
Looks totally reasonable.

I checked, ```SimpleServiceInstance('ipa_memcached')``` does not raise an exception if systemd has no service file for IPA memcached at all.

MartinBasti:
Does this belong to any ticket which caused this regression?

HonzaCholasta:
@MartinBasti, https://fedorahosted.org/freeipa/ticket/5959.

stlaz:
I see where this is going, added the ticket to the commit message.

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/32076df10231b381a80c9ef850c2c31d7a25feb8

PR #497, opened 2017-02-23
added more meaningful help for --external-ca-type option

tscherf:


MartinBasti:
This is already fixed in master, however it missed what is the default value

```
  --external-ca-type={generic,ms-cs}
                        Type of the external CA
```

MartinBasti:
Replaced by #499

PR #498, opened 2017-02-23
compat: fix `Any` params in `batch` and `dnsrecord`

HonzaCholasta:
The `methods` argument of `batch` and `dnsrecords` attribute of `dnsrecord`
were incorrectly defined as `Str` instead of `Any`.

https://fedorahosted.org/freeipa/ticket/6647


MartinBasti:
Works for me

MartinBasti:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/19060db1b8fa9d1d3e8f3ac3fcd1f387e9a39c94

PR #501, opened 2017-02-24
C compilation fixes and hardening

tiran:
Fix "implicit declaration of function ‘strlen’" in ipa_pwd_ntlm.c,
credits to Lukas.

Add -Werror=implicit-function-declaration to CFLAGS to point developers
to missing includes. It causes compilation to fail when a developer
forgets to add a required include. The problem is no longer hidden in a
massive wall of text from make.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -24,6 +24,9 @@ LT_INIT
 
 AC_HEADER_STDC
 
+dnl fail hard when includes statements are missing
+CFLAGS+=" -Werror=implicit-function-declaration"
+
 PKG_PROG_PKG_CONFIG
 
 AC_ARG_ENABLE([server],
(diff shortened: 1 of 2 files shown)


stlaz:
I agree that C compilation should be hardened for FreeIPA, seeing warnings is nothing unusual here. This builds fine. ACK.

MartinBasti:
master:

* 2828a2b92b89932d66b640e5047161448d522e2e C compilation fixes and hardening

lslebodn:
FYI; it is far far away from best practice to modify `CFLAGS` and configure time; unless you test compiler options. Such change should be in makefile `AM_CFLAGS`

In the future, try to address comments in PR  (#364) rather then ignoring them and then fixing in new PR

tiran:
@lslebodn feel free to open a new PR. This PR and #364 are closed.

lslebodn:
I know it's closed.

It was just a kindly reminder to address obvious problems as part of review process.

PR #506, opened 2017-02-24
Use IPA CA cert in Custodia secrets client

tscherf:


tiran:
Why do you propose to change the settings? By default python-requests enforces certificate validation. Without additional settings, it uses the system trust store. The IPA root CA is injected into the system trust store.

HonzaCholasta:
We don't want to trust certificates issued by random internet CAs, this is how it should have been from the beginning. A commit message would be nice though.

@tscherf, please add this ticket URL to the commit message: https://fedorahosted.org/freeipa/ticket/6686

tiran:
Please change the title of the commit, too. It's implies that we did not verify certs in the past.

In the future please don't call the system trust store a random collection of CAs. It's diminishing and vilifying the hard work of the security team to provide a secure selection of CA certs. This change is purely an attempt to harden IPA and use the same selection of CAs everywhere.

tscherf:
When the system wide trust store is supposed to be used here, then something else must be broken somewhere in the verification code. Without explicitly using the IPA trust anchor stored in IPA_CA_CRT, the installer failed with an "[SSL: CERTIFICATE_VERIFY_FAILED]" error. We have seen this in CA-less and chained CA setups.

tscherf:
Sorry, closed this by mistake.

tiran:
LGTM, but I want @simo5 to give the final ACK.

Since Custodia is only used during replica installation on an enrolled system, ipa-client-install has already provided the certificate. I don't see any issue in the proposed fix.

```ipaserver.secrets.client``` does not yet use Custodia's own client library. I'll keep the problem in mind once we have updated to recent Custodia version.

simo5:
Works for me.

frasertweedale:
@tiran FYI custodia is also used for Lightweight CA key replication, at any time a new LWCA gets created, to propagate its signing key among replicas.  So this is a useful change.

HonzaCholasta:
Fixed upstream
master:
https://fedorahosted.org/freeipa/changeset/16dac0252e52c8de07fd8a6a86ec0896074cbe9d

PR #507, opened 2017-02-24
Use https to get security domain from Dogtag

tiran:

### DIFF
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -51,7 +51,11 @@ def get_security_domain():
     Get the security domain from the REST interface on the local Dogtag CA
     This function will succeed if the local dogtag CA is up.
     """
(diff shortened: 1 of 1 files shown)


pvoborni:
What is a context of this patch? Is something broken only in master. Or also 4.4, Fedora, RHEL,...?

pvoborni:
I.e. I want to know if something needs to be or should be backported.

tiran:
The patch hardens the installer a bit. It would be a good idea to backport the patch to 4.4.

It's not critical since it's a read operation on localhost.

tomaskrizek:
master:

* d1c5d92897d3e262edd2e43295c1270590aebd3d Use https to get security domain from Dogtag

tomaskrizek:
If backport for 4.4 is needed, please open another PR against `ipa-4-4`. Thanks.

PR #508, opened 2017-02-24
Fix ipa.service unit re. gssproxy

flo-renaud:
ipa.service unit defines Requires=gssproxy. Because of this, during
ipa-server-upgrade, the restart of gssproxy triggers a restart of ipa unit
(hence stopping LDAP server and breaking the connection api.Backend.ldap2).
Calls using this connection after gssproxy restart fail and ipa-server-upgrade
exits on failure.
The fix defines Wants=gssproxy to avoid the restart of ipa.service

https://fedorahosted.org/freeipa/ticket/6705

### DIFF
--- a/init/systemd/ipa.service.in
+++ b/init/systemd/ipa.service.in
@@ -1,7 +1,7 @@
 [Unit]
 Description=Identity, Policy, Audit
 Requires=network.target
-Requires=gssproxy.service
+Wants=gssproxy.service
 After=network.target
 
 [Service]


simo5:
Should we also change the Requires on network.target ?
Do we really want to have a restart of IPa if someone restarts the network ?

abbra:
Good point. I think we shouldn't restart ourselves as we anyway are listening on all interfaces with 0.0.0.0.

flo-renaud:
@simo5 @abbra I agree but this should be tracked in a separate issue.

simo5:
Seemed worth fixing at the same time, but I won't insist.

MartinBasti:
master:

* 98e3b14a0477232054b02065c857fb1b16ce85a6 Fix ipa.service unit re. gssproxy

PR #510, opened 2017-02-25
Vault: port key wrapping to python-cryptography

tiran:
https://fedorahosted.org/freeipa/ticket/6650


tiran:
@simo5 Do I remember correctly that PKCS1v1.5 side channel attacks applies only to unpadding and not to padding?

tiran:
I can answer the question myself. The side channel attack on RSAEP PKCS1 v1.5 is a chosen-ciphertext attack Bleichenbacher attack. It applies to unpadding and RSA decryption, because its only a danger to the side that owns the private key. Therefore

stlaz:
Tested this and gone through the code, both was fine. ACK.

MartinBasti:
master:

* ed7a03a1af8b556247b929635e2972be4f2b32e4 Vault: port key wrapping to python-cryptography

PR #512, opened 2017-02-27
test_config: fix fips_mode key in Env

tomaskrizek:
Setting fips_mode to object would fail if ipaplatform.tasks module
wasn't present.


MartinBasti:
master:

* 770d4cda430803f8e020c57971c4dd8e802dc417 Env __setitem__: replace assert with exception
* 5055b34cefd6e3f9b707aed076a49ae97b38aa3c test_config: fix fips_mode key in Env

PR #513, opened 2017-02-27
certdb: Don't restore_context() of new NSSDB

tiran:
It's not necesary to restore the context of newly created files. SELinux
ensures that new files have the correct permission. An explicit
restore_context() is only required when either policies have changed or
the context was changed manually.

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -29,7 +29,6 @@ from cryptography.hazmat.primitives import serialization
 from nss import nss
 from nss.error import NSPRError
 
-from ipaplatform.tasks import tasks
 from ipapython.dn import DN
 from ipapython.ipa_log_manager import root_logger
 from ipapython import ipautil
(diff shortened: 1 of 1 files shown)


tiran:
I also dropped =1 check. http://man7.org/linux/man-pages/man2/chown.2.html

> If the owner or group is specified as -1, then that ID is not changed.

MartinBasti:
This is old code, initially added here 49b36583a50e7f542e0667f3e2432ab1aa63924e

But I failed to detect why restorecon call has been added for new databases.

LGTM, but I want to think more about it, why it was added there and what we can break with this commit

stlaz:
I don't see how this could break anything given that it's been used like that for ages prior to priv-sep patches.

tiran:
Maybe it was required back then. 7, 8 years is a long time. Nowadays new files are created with correct context:

```
# rm -f /etc/ipa/nssdb/testfile
# touch /etc/ipa/nssdb/testfile
# ls -laZ /etc/ipa/nssdb/testfile 
-rw-r--r--. 1 root root unconfined_u:object_r:cert_t:s0 0 Mar  1 09:08 /etc/ipa/nssdb/testfile
# restorecon /etc/ipa/nssdb/testfile 
# ls -laZ /etc/ipa/nssdb/testfile 
-rw-r--r--. 1 root root unconfined_u:object_r:cert_t:s0 0 Mar  1 09:08 /etc/ipa/nssdb/testfile
```

MartinBasti:
Ok, I agree

MartinBasti:
master:

* a163ad77b3d12f2da2b135de29f594c06190b41a certdb: Don't restore_context() of new NSSDB

PR #514, opened 2017-02-27
Limit sessions to 30 minutes by default

simo5:
When we changed the session handling code we unintentinally extended
sessions expiraion time to the whole ticket lifetime of 24h.

Related to https://fedorahosted.org/freeipa/ticket/5959

### DIFF
--- a/install/conf/ipa.conf
+++ b/install/conf/ipa.conf
@@ -67,6 +67,7 @@ WSGIScriptReloading Off
   Session On
   SessionCookieName ipa_session path=/ipa;httponly;secure;
   SessionHeader IPASESSION
+  SessionMaxAge 1800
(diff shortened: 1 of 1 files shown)


tiran:
Would it makes sense to add https://httpd.apache.org/docs/trunk/mod/mod_session.html#sessionexpiryupdateinterval and set it to a small value like 30 seconds?

> The SessionExpiryUpdateInterval directive allows sessions to avoid the cost associated with writing the session each request when only the expiry time has changed. This can be used to make a website more efficient or reduce load on a database when using mod_session_dbd.

simo5:
No, we do not store sessions in a session db, so that setting is not useful to us.

MartinBasti:
master:

* d5e7a57e5b25b9cecb7a65096487a65374ad860d Limit sessions to 30 minutes by default

PR #515, opened 2017-02-27
Re-add ipapython.config.config for backwards compatibilty

tiran:
IPAConfig, config and init_config were removed in rev 7b966e85. Ipsilon uses
ipapython.config to get realm, domain and server of an enrolled host. Re-add
a simplified version that reads settings from api.env. init_config() does
not perform DNS discovery.

Depends on PR #492 to get meaningful defaults.

https://fedorahosted.org/freeipa/ticket/6707

### DIFF
--- a/ipapython/config.py
+++ b/ipapython/config.py
@@ -24,6 +24,7 @@ from optparse import (
 from copy import copy
 
 from ipapython.dn import DN
+import ipalib
 
 
 class IPAFormatter(IndentedHelpFormatter):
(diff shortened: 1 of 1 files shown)


MartinBasti:
IIRC we agreed that there should be warning that this is deprecated and `api.env` should be used instead.

tiran:
I can add a deprecation warning after we have agreed upon a new API. What's the official way to get the values w/o requiring credentials?

HonzaCholasta:
Could we please revert to the original `IPAConfig` implementation rather than wrapping around `api.env`? I know I'm the one who suggested it, but I have given it some thought and I would rather not have to import from `ipalib` into `ipapython` and keep the original behavior intact.

tiran:
OK, I'm closing this PR then. 

@HonzaCholasta, please open another PR and revert 7b966e8.

PR #516, opened 2017-02-28
IdM Server: list all Employees with matching Smart Card

flo-renaud:
Implement a new IPA command allowing to retrieve the list of users matching the provided certificate.
The command is using SSSD Dbus interface, thus including users from IPA domain and from trusted domains. This requires sssd-dbus package to be installed on IPA server.

https://fedorahosted.org/freeipa/ticket/6646

### DIFF
--- a/ipaserver/plugins/certmap.py
+++ b/ipaserver/plugins/certmap.py
@@ -17,9 +17,14 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import base64
+import dbus
 import six
 
-from ipalib import api, errors
+from ipalib import api, errors, x509
+from ipalib import Bytes
+from ipalib.crud import Search
+from ipalib.frontend import Object
 from ipalib.parameters import Bool, DNSNameParam, Flag, Int, Str
 from ipalib.plugable import Registry
 from .baseldap import (
@@ -33,6 +38,7 @@ from .baseldap import (
     pkey_to_value)
 from ipalib import _, ngettext
 from ipalib import output
+from ipaserver.plugins.service import validate_certificate
 
 
 if six.PY3:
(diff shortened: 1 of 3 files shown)


flo-renaud:
Note: this PR is work in progress. It requires PR#398 Support for Certificate Identity Mapping and sssd patches not pushed yet.

abbra:
One thing I don't like is that SELinux policy requirements aren't mentioned. To allow ipaapi user to talk to SSSD dbus interface, you have to have a policy that allows this.

simo5:
Why do we need to talk to SSSD to do this?
Don't we have all the needed data in LDAP already ?

flo-renaud:
Hi @simo5 
The command must also be able to return matching entries coming from trusted domains, and SSSD is able to handle this part for us.

ipaserver/plugins/certmap.py
```diff
@@ -389,3 +393,115 @@ def execute(self, cn, **options):
...
+        """
+        sssd = _sssd(self.log)
+        result = dict()
+
+        cert = options['certificate']
+        users = sssd.list_users_by_cert(cert)
+        result['count'] = len(users)
+        result['uids'] = users
+        result['summary'] = u'{count} user{plural} matched'.format(
+            count=len(users), plural='' if len(users) == 1 else 's')
```
HonzaCholasta: This is not translatable. Instead of filling in `summary` yourself, you should specify the `msg_summary` class attribute and have it filled automatically.

ipaserver/plugins/certmap.py
```diff
@@ -389,3 +393,115 @@ def execute(self, cn, **options):
...
+
+@register()
+class certmap_match(Command):
+    __doc__ = _('Search for users matching the provided certificate.')
+
+    has_output = (
+        output.summary,
+        output.Output('uids', (list, tuple, type(None)), _('Matched uid')),
+        output.Output('count', int, _('Number of entries returned')),
+        )
```
HonzaCholasta: Please avoid using custom `has_output`, as it requires a client-side plugin. Define `certmap_match` as a search method of a virtual `certmap` object instead:
```python
from ipalib.crud import Search
from ipalib.frontend import Object

@register()
class certmap(Object):
    has_params = (
        Str(
            'uid',
            ...
            flags={'no_search'},
        ),
        Bytes(
            'certificate',
            ...
        ),
    )

@register()
class certmap_match(Search):
    def get_args(self):
        for arg in super(certmap_match, self).get_args():
            if arg.name == 'criteria':
                continue
            yield arg

    def execute(self, *args, **options):
        ...
        result['result'] = [{'uid': user} for user in users]
        result['count'] = len(users)
        result['truncated'] = False
        return result
```

Also, I don't think the result should be called `uid`, as that's usually the username without the domain qualifier. You can either rename it or split it into `uid` and `domain` (you can do that with the object plugin).

ipaserver/plugins/certmap.py
```diff
@@ -389,3 +393,115 @@ def execute(self, cn, **options):
...
+
+    takes_options = (
+        Bytes(
+            'certificate', validate_certificate,
```
HonzaCholasta: I guess this should be a positional argument, since it is the only argument and is required.

ipaserver/plugins/certmap.py
HonzaCholasta: Actually, I think it would make sense to group the results by domain, in which case `certmap` should look like this:
```python
@register()
class certmap(Object):
    has_params = (
        DNSNameParam(
            'domain',
            ...
            flags={'no_search'},
        ),
        Bytes(
            'certificate',
            ...
        ),
        Str(
            'uid*',
            ...
            flags={'no_search'},
        ),
    )
```

simo5:
I am not sure we want to wait for replies from trusted domains, it may be very slow, and in some cases it will just not work right (one way trusts with strict access control on entries).
Active Directory forces users to provide a hint when logging into trusted domains with smart cards and does not query the remote domain. Have we considered this ?

sumit-bose:
Yes, a hint aka user name will be used during authentication. But this PR here is about to get an idea which user is allowed to authenticate based on the current certificate mapping configuration. Since the certificate mapping configuration requires remote domains to be added explicitly to admin can control which domains are included in the search.

flo-renaud:
@abbra ,
Thanks for your comment. Running in permissive mode I did not see any AVC logged in the journal.

@HonzaCholasta 
thanks for the tips re. writing API. I have followed your advice and made certificate a positional argument. The output will look like this:
```
---------------
2 users matched
---------------
  Domain: DOM-076.ABC.IDM.LAB.ENG.BRQ.REDHAT.COM
  Usernames: user1, user2
----------------------------
Number of entries returned 2
----------------------------
```

ipaserver/plugins/certmap.py
pvoborni: @HonzaCholasta it would be good to catch these issues in threads like http://www.redhat.com/archives/freeipa-devel/2017-February/msg00878.html which are meant especially for it to avoid  rewriting patches.

ipaserver/plugins/certmap.py
```diff
@@ -389,3 +395,148 @@ def execute(self, cn, **options):
...
+        Users.ListByCertificate.
+        SSSD does the lookup based on certificate mapping rules, using
+        FreeIPA domain and trusted domains.
+        :raise RemoteRetrieveError: if DBus returns an exception
+        """
+        sssd = _sssd(self.log)
+
+        cert = args[0]
+        users = sssd.list_users_by_cert(cert)
+        count = sum([len(l) for (_k, l) in users.items()])
```
HonzaCholasta: `count` should be the number of entries in the result, i.e. `len(result)`.

If you want to keep the "N users matched" summary intact, you can do it using a `get_summary_default` method:
```python
    def get_summary_default(self, output):
        count = sum(len(entry['uid']) for entry in output['result'])
        return self.msg_summary % dict(count=count)
```

ipaserver/plugins/certmap.py
```diff
@@ -389,3 +395,148 @@ def execute(self, cn, **options):
...
+    """
+    takes_params = (
+        DNSNameParam(
+            'domain',
+            label=_('Domain'),
+            flags={'no_search'},
+        ),
+        Str(
+            'uid*',
+            label=_('Usernames'),
```
HonzaCholasta: In the `user` plugin, `uid` is called "Login" - perhaps we should use the same (singular or plural) here as well?

HonzaCholasta:
@flo-renaud, please rebase.

flo-renaud:
Hi @HonzaCholasta 
thank you for your comments. Patch rebased.

ipaserver/plugins/certmap.py
HonzaCholasta: I see you added `get_summary_default`, but it seems you forgot to update the `count` line to `count = len(results)`.

flo-renaud:
Hi @HonzaCholasta 
sorry I overlooked the change for count. It's updated now, thank you for the review.

HonzaCholasta:
@flo-renaud, thanks, LGTM.

BTW Travis fails because there is no `sssd-dbus >= 1.15.1` - submitting a build to freeipa-master now.

dkupka:
@flo-renaud While playing with this command I've noticed one disturbing fact. Because we rely on SSSD and SSSD rely its cache we will likely return inaccurate result.
I'm thinking about use-case when admin calls certmap-match to list current users mapped to the certificate. Then he performs some changes and  calls certmap-match again to verify his changes. At that point SSSD may use cache and return obsolete result.
One possible solution would be expiring the cache on every certmap-match call but that can easily have serious performance impact.

flo-renaud:
Hi @dkupka 
As the goal of this command is to return exactly the same list of users as SSSD would consider for authentication, IMHO it is expected that we may have a cached list instead of an up-to-date list of results, because sssd authentication would have the same result.

dkupka:
@flo-renaud That's right but we should probably stress this somehow because it's not intuitive. Also we're returning what SSSD would return on master but we have no idea what it will return on some other host.

sumit-bose:
I agree, it would be good if the help text can mention that cached data is used and maybe even mention the sss_cache utility to invalidate the entry. If the doc team can add this to the official documentation it would be even better.

dkupka:
@sumit-bose I agree. If this is in help text we can also display it in WebUI.
@flo-renaud  Please add description and explanation of this behaviour into __doc__ for certmap_match. Otherwise the pull request looks good to me and works as expected.

flo-renaud:
@dkupka 
I added the following explanation in the doc for certmap_match:
    """
    Search for users matching the provided certificate.

    This command relies on SSSD to retrieve the list of matching users and
    may return cached data. For more information on purging SSSD cache,
    please refer to sss_cache documentation.
    """

dkupka:
@flo-renaud Thank you.

dkupka:
master:

* ea34e17a46a60efb9c4dc81dab919a1639dec73b IdM Server: list all Employees with matching Smart Card

HonzaCholasta:
I forgot to say that in the CLI, the certificate should be specified using a file. PR #557 implements this.

PR #521, opened 2017-02-28
Add nsaccountlock to user attributes when a new user is created

redhatrises:
This adds a the `nsaccountlock` attribute to a user upon account creation. This addresses newly created accounts; however, it does not address the issue of existing accounts. If `nsaccountlock` does not exist for a user, `ipa user-find --disabled=False` should return `Accounts disabled: False`; however, it returns nothing.

So, the question is how to deal with `nsaccountlock` missing in existing user accounts? I am not sure how to extend the framework to return `Accounts disabled: False` if `nsaccountlock` is NoneType. 

For more info, see @MartinBasti's post merge comments in #444


MartinBasti:
It looks that my proposal is not the right way, sorry. (See: #444)

PR #522, opened 2017-03-01
dogtag: remove redundant property definition

frasertweedale:
The dogtag `ra` backend defines a `ca_host` property, which is also
defined (identically) by the `RestClient` class, which recently
became a superclass of `ra`.  Remove the redundant property
definition.


MartinBasti:
master:

* 49f87f34be5f04f18a6d916276153e9ef1e5852c dogtag: remove redundant property definition

PR #525, opened 2017-03-01
Remove import nss from test_ldap

tiran:
test_ldap just imported nss.nss to call nss_init_nodb(). It should be
safe to remove the call. Let's see what CI has to say.


martbab:
looks like Travis did not mind at all

MartinBasti:
master:

* 79c0e6d355c9e7bcc7cacc37faaba8e999d56400 Remove import nss from test_ldap

PR #526, opened 2017-03-01
server install: do not attempt to issue PKINIT cert in CA-less

HonzaCholasta:
Require the user to provide the PKINIT cert with --pkinit-cert-file or
disable PKINIT with --no-pkinit in CA-less ipa-server-install,
ipa-replica-prepare and ipa-replica-install.

Do not attempt to issue the PKINIT cert in CA-less ipa-server-upgrade.

https://pagure.io/freeipa/issue/5678

### DIFF
--- a/ipaserver/install/ipa_replica_prepare.py
+++ b/ipaserver/install/ipa_replica_prepare.py
@@ -160,16 +160,21 @@ class ReplicaPrepare(admintool.AdminTool):
             self.option_parser.error("You cannot specify a --reverse-zone "
                 "option together with --no-reverse")
 
-        #Automatically disable pkinit w/ dogtag until that is supported
-        options.setup_pkinit = False
-
         # If any of the PKCS#12 options are selected, all are required.
         cert_file_req = (options.dirsrv_cert_files, options.http_cert_files)
         cert_file_opt = (options.pkinit_cert_files,)
+        if options.setup_pkinit:
+            cert_file_req += cert_file_opt
         if any(cert_file_req + cert_file_opt) and not all(cert_file_req):
             self.option_parser.error(
-                "--dirsrv-cert-file and --http-cert-file are required if any "
-                "PKCS#12 options are used.")
+                "--dirsrv-cert-file, --http-cert-file, and --pkinit-cert-file "
+                "or --no-pkinit are required if any key file options are used."
+            )
+        if not options.setup_pkinit and options.pkinit_cert_files:
+            self.option_parser.error(
+                "--no-pkinit and --pkinit-cert-file cannot be specified "
+                "together"
+            )
 
         if len(self.args) < 1:
             self.option_parser.error(
--- a/ipaserver/install/server/__init__.py
+++ b/ipaserver/install/server/__init__.py
@@ -347,10 +347,18 @@ class ServerInstallInterface(client.ClientInstallInterface,
         # If any of the key file options are selected, all are required.
         cert_file_req = (self.dirsrv_cert_files, self.http_cert_files)
         cert_file_opt = (self.pkinit_cert_files,)
+        if not self.no_pkinit:
+            cert_file_req += cert_file_opt
         if any(cert_file_req + cert_file_opt) and not all(cert_file_req):
             raise RuntimeError(
-                "--dirsrv-cert-file and --http-cert-file are required if any "
-                "key file options are used.")
+                "--dirsrv-cert-file, --http-cert-file, and --pkinit-cert-file "
+                "or --no-pkinit are required if any key file options are used."
+            )
+        if self.no_pkinit and self.pkinit_cert_files:
+            raise RuntimeError(
+                "--no-pkinit and --pkinit-cert-file cannot be specified "
+                "together"
+            )
 
         if not self.interactive:
             if self.dirsrv_cert_files and self.dirsrv_pin is None:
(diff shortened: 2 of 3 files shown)


abbra:
An idea behind the original solution was to always produce PKINIT certificate by certmonger in case of CA-less install to be able to have anonymous PKINIT supported. PKINIT cert should have specific attributes and in many cases they aren't issued by external CAs. However, the certificate is not really needed to be connected to existing CAs.

Admins can re-issue PKINIT cert afterwards but at least we can get anonymous PKINIT to wrap 2FA with.

So this pull request actually breaks CA-less deployment.

HonzaCholasta:
In CA-less there is no CA to request the certificate from, so there is a dangling failed certmonger request. This PR removes the broken request and thus fixes CA-less.

HonzaCholasta:
This is what you currently get in CA-less install:
```
# getcert list
Number of certificates and requests being tracked: 1.
Request ID '20170301121440':
	status: CA_UNREACHABLE
	ca-error: Server at https://vm-226.abc.idm.lab.eng.brq.redhat.com/ipa/xml failed request, will retry: -504 (libcurl failed to execute the HTTP POST transaction, explaining:  Failed to connect to vm-226.abc.idm.lab.eng.brq.redhat.com port 443: Connection refused).
	stuck: no
	key pair storage: type=FILE,location='/var/kerberos/krb5kdc/kdc.key'
	certificate: type=FILE,location='/var/kerberos/krb5kdc/kdc.crt'
	CA: IPA
	issuer: 
	subject: 
	expires: unknown
	pre-save command: 
	post-save command: 
	track: yes
	auto-renew: yes
# ls /var/kerberos/krb5kdc/kdc.crt
ls: cannot access '/var/kerberos/krb5kdc/kdc.crt': No such file or directory
```

abbra:
No, you are wrong. Certmonger has own local self-signed CA in all installs:

    # getcert list-cas
      ....
     CA 'local':
         is-default: no
         ca-type: EXTERNAL
         helper-location: /usr/libexec/certmonger/local-submit

This is what can and should be used for self-signed case for PKINIT.

abbra:
This was, perhaps, missed in the original commit, though. The idea was that in CA-less mode we change request to use Local CA.

HonzaCholasta:
In CA-less mode one has to provide all the certs manually. I don't see why the KDC cert should be an exception and why we should reinvent the wheel for it.

You can't use the local CA anyway, because it's not trusted by IPA. Even if you made it trusted on the local system, it would not be trusted globally - to do that you would have to either make every local CA on every server trusted globally, which does not scale well and would most likely cause more issues than solve, or provide a mechanism to synchronize the CA's private key between servers, which is non-trivial and out of the scope of the PKINIT effort.

If you think it is a good idea to support the local CA in addition to Dogtag, please file a RFE. Meanwhile, this PR fixes an obvious bug without implemeting any additional features.

abbra:
This PR does not handle upgrade case which is what Local CA considers. We don't need other systems trust the certificate and we don't need to synchronize anything because KDC cert in upgrade case is issued automatically and is used by privilege separation code on the same machine.

HonzaCholasta:
The local CA is in fact not used in CA-less upgrade. This is what you get after upgrade from 4.4.3 to current master:
```
# getcert list
Number of certificates and requests being tracked: 1.
Request ID '20170301142723':
	status: CA_UNREACHABLE
	ca-error: Server at https://vm-226.abc.idm.lab.eng.brq.redhat.com/ipa/xml failed request, will retry: 4001 (RPC failed at server.  CA is not configured).
	stuck: no
	key pair storage: type=FILE,location='/var/kerberos/krb5kdc/kdc.key'
	certificate: type=FILE,location='/var/kerberos/krb5kdc/kdc.crt'
	CA: IPA
	issuer: 
	subject: 
	expires: unknown
	pre-save command: 
	post-save command: 
	track: yes
	auto-renew: yes
# ls /var/kerberos/krb5kdc/kdc.crt
ls: cannot access '/var/kerberos/krb5kdc/kdc.crt': No such file or directory
```

Additionally, there is no mention of using the local CA to issue the cert in CA-less in any of the following designs:
* http://www.freeipa.org/page/V4/External_Authentication
* http://www.freeipa.org/page/V4/Kerberos_PKINIT

In other words, using the local CA is something a) not designed properly b) not implemented at all.

HonzaCholasta:
Updated the PR to also handle CA-less server upgrade.

@abbra, I'm not opposed to the idea of using the local CA to issue the KDC cert, but if we agree to use it, we should use it in both CA-less and CA-ful - if the CA does not need to be trusted as you say, using the IPA CA in CA-ful is meaningless and only adds unnecesary complexity.

abbra:
ACK for the patch. However, I'm not claiming that CA does not need to be trusted. What I'm saying is that for Anonymous PKINIT's use in privilege separation code we can issue certs using local CA because we can trust local CA on IPA masters. They would be all different local CAs, of course, but this was thought to be a stop-gap until admins can replace local certificates with the proper ones some time after upgrade.

Privilege separation code now supports several ways to kinit and falls back to a wrapping with HTTP/ipa.master credentials in case anonymous PKINIT is not available.

HonzaCholasta:
OK, thanks.

HonzaCholasta:
master:

* ba3c201a03cd0b224b43e45245147e48b7291f9f server install: do not attempt to issue PKINIT cert in CA-less

PR #527, opened 2017-03-01
Fix replica with --setup-ca issues

stlaz:
nolog argument of ipautil.run requires tuple, not a string.

I am a bad person.


HonzaCholasta:
master:

* 052de4308c64b126bee440e970be4cf8449c5ebc Fix replica with --setup-ca issues

PR #528, opened 2017-03-01
Fix CA-less upgrade

stlaz:
In CA-less mode there's no /etc/pki/pki-tomcat/password.conf so it
does not make sense to try to create a password file for an NSS
database from it (the NSS database does not exist either).

Thanks @HonzaCholasta for discovering this.


HonzaCholasta:
master:

* a7c8077ce8f72eee26e8f5d4362239313ffdae3d Fix CA-less upgrade

PR #529, opened 2017-03-01
installer: update time estimates

tomaskrizek:
Time estimates have been updated to be more accurate. Only
tasks that are estimated to take longer than 10 seconds have
the estimate displayed.

https://pagure.io/freeipa/issue/6596

### DIFF
--- a/ipaserver/install/krainstance.py
+++ b/ipaserver/install/krainstance.py
@@ -132,7 +132,7 @@ class KRAInstance(DogtagInstance):
         self.step("enabling KRA instance", self.__enable_instance)
 
         try:
-            self.start_creation(runtime=126)
+            self.start_creation(runtime=120)
         finally:
             self.clean_pkispawn_files()
 
--- a/ipaserver/install/service.py
+++ b/ipaserver/install/service.py
@@ -365,7 +365,7 @@ class Service(object):
         self.steps.append((message, method, run_after_failure))
 
     def start_creation(self, start_message=None, end_message=None,
-        show_service_name=True, runtime=-1):
+                       show_service_name=True, runtime=None):
         """
         Starts creation of the service.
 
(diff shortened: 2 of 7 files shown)


stlaz:
This will say:
```
Configuring certificate server (pki-tomcatd). Estimated time: 3 minutes
  [1/2]: configure certmonger for renewals
  [2/2]: Importing RA key
```
but the operation lasts ~5 seconds at most.

ipaserver/install/krainstance.py
```diff
@@ -134,8 +134,13 @@ def configure_instance(self, realm_name, host_name, dm_password,
 
             self.step("enabling KRA instance", self.__enable_instance)
 
+        if ra_only:
```
tomaskrizek: Not sure if this is necessary. Is kra install ever executed with `ra_only=True` option?

tomaskrizek:
@stlaz That estimate was a bit off :) Thanks for noticing!

ipaserver/install/krainstance.py
stlaz: Please rebase on current master

ipaserver/install/service.py
```diff
@@ -401,7 +401,7 @@ def start_creation(self, start_message=None, end_message=None,
                 else:
                     end_message = "Done configuring %s." % self.service_desc
 
-        if runtime > 0:
+        if runtime and runtime > 0:
```
stlaz: The correct condition is `if runtime is not None and and runtime > 0`

stlaz:
Please fix the little issue and we will push this.

ipaserver/install/service.py
MartinBasti: `SyntaxError: invalid syntax` :-P

stlaz: 🤕

MartinBasti:
master:

* 09c6b7578046fed0824fc0f0d9040be69c0f0eb6 installer: update time estimates

PR #531, opened 2017-03-02
httpinstance: disable system trust module in /etc/httpd/alias

HonzaCholasta:
Currently the NSS database in /etc/httpd/alias is installed with the system
trust module enabled. This is problematic for a number of reasons:

* IPA has its own trust store, which is effectively bypassed when the
  system trust module is enabled in the database. This may cause IPA
  unrelated CAs to be trusted by httpd, or even IPA related CAs not to be
  trusted by httpd.

* On client install, the IPA trust configuration is copied to the system
  trust store for third parties. When this configuration is removed, it may
  cause loss of trust information in /etc/httpd/alias
  (https://bugzilla.redhat.com/show_bug.cgi?id=1427897).

* When a CA certificate provided by the user in CA-less install conflicts
  with a CA certificate in the system trust store, the latter may be used
  by httpd, leading to broken https
  (https://www.redhat.com/archives/freeipa-users/2016-July/msg00360.html).

Disable the system trust module on install and upgrade to prevent the
system trust store to be used in /etc/httpd/alias and fix all of the above
issues.

https://pagure.io/freeipa/issue/6132

### DIFF
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -349,11 +349,25 @@ class HTTPInstance(service.Service):
         os.chown(pwd_conf, pent.pw_uid, pent.pw_gid)
         os.chmod(pwd_conf, 0o400)
 
+    def disable_system_trust(self):
+        name = 'Root Certs'
+        args = [paths.MODUTIL, '-dbdir', paths.HTTPD_ALIAS_DIR, '-force']
+
+        result = ipautil.run(args + ['-list', name],
+                             env={},
+                             capture_output=True)
+        if 'Status: Enabled' in result.output:
+            ipautil.run(args + ['-disable', name], env={})
+            return True
+
+        return False
+
     def __setup_ssl(self):
         db = certs.CertDB(self.realm, nssdir=paths.HTTPD_ALIAS_DIR,
                           subject_base=self.subject_base, user="root",
                           group=constants.HTTPD_GROUP,
                           truncate=(not self.promote))
+        self.disable_system_trust()
         if self.pkcs12_info:
             if self.ca_is_configured:
                 trust_flags = 'CT,C,C'
--- a/ipaplatform/base/paths.py
+++ b/ipaplatform/base/paths.py
@@ -165,6 +165,7 @@ class BasePathNamespace(object):
     BIN_KVNO = "/usr/bin/kvno"
     LDAPMODIFY = "/usr/bin/ldapmodify"
     LDAPPASSWD = "/usr/bin/ldappasswd"
+    MODUTIL = "/usr/bin/modutil"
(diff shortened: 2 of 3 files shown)


tiran:
Your solution is just a temporary solution. The file is re-added every time mod_nss is updated. The mod_nss package needs to be changed, too.

HonzaCholasta:
Sigh, did not notice that.

tiran:
```
$ rpm -qf /etc/httpd/alias/libnssckbi.so 
mod_nss-1.0.14-1.fc25.x86_64
````

https://src.fedoraproject.org/cgit/rpms/mod_nss.git/tree/mod_nss.spec#n158

```
%files
...
%{_sysconfdir}/httpd/alias/libnssckbi.so
```

**Comment by HonzaCholasta** [comment]
Updated to use `modutil -disable` which works even on mod_nss reinstall.

**Comment by rcritten** [comment]
IIRC on install all three existing db's get copied to <name>.orig, or something like that right? So uninstall would move those back into place effectively disabling this?

**Comment by HonzaCholasta** [comment]
Yes. That is, once https://pagure.io/freeipa/issue/4639 is fixed.

**Comment by rcritten** [comment]
Just FYI I'm opening an upstream discussion with the NSS team on this. It is very strange that there is a conflict like this, particularly between master and replica.

**Comment by HonzaCholasta** [comment]
The conflict between master and replica exists because on the master, client install is executed last, but on (domain level 1+) replica it is executed first, so on the master `/etc/httpd/alias` is populated first and `/etc/pki/ca-trust/source/ipa.p11-kit` is created later, but on (domain level 1+) replica it is done the other way around.

File: ipaserver/install/httpinstance.py [hunk 1]
```diff
@@ -349,11 +349,25 @@ def create_password_conf(self):
...
+        args = [paths.MODUTIL, '-dbdir', paths.HTTPD_ALIAS_DIR, '-force']
+
+        result = ipautil.run(args + ['-list', name],
+                             env={},
+                             capture_output=True)
+        if 'Status: Enabled' in result.output:
+            ipautil.run(args + ['-disable', name], env={})
+            return True
+
+        return False
```
stlaz: I don't see any use of the return value.

stlaz:
It seems to work fine for "mod_nss" reinstalls but `ipa-server-upgrade` is currently failing so I can't confirm that's ok.

stlaz:
I rebased your patchset on current master and put the uninstallation of `ipa_memcached` into a multipass block and all seems to work now.

MartinBasti:
master:

* f037bfa48356a5fb28eebdb76f9dbd5cb461c2d2 httpinstance: disable system trust module in /etc/httpd/alias

PR #532, opened 2017-03-02
Fix cookie with Max-Age processing

stlaz:

### DIFF
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -759,9 +759,11 @@ class KerbTransport(SSLTransport):
         session_cookie = None
         try:
             for cookie in cookie_header:
-                session_cookie = \
-                    Cookie.get_named_cookie_from_string(cookie, COOKIE_NAME,
-                                                        request_url)
+                session_cookie = (
+                    Cookie.get_named_cookie_from_string(
+                        cookie, COOKIE_NAME, request_url,
+                        timestamp=datetime.datetime.utcnow())
+                    )
                 if session_cookie is not None:
                     break
         except Exception as e:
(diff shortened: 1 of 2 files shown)


ipalib/rpc.py
```diff
@@ -861,7 +863,9 @@ def get_session_cookie_from_persistent_storage(self, principal):
 
         # Search for the session cookie within the cookie string
         try:
-            session_cookie = Cookie.get_named_cookie_from_string(cookie_string, COOKIE_NAME)
+            session_cookie = Cookie.get_named_cookie_from_string(
+                cookie_string, COOKIE_NAME,
+                timestamp=datetime.datetime.now())
```
tiran: ```datetime.datetime.utcnow()```

Our cookie library uses UTC but now() returns a naive localized time stamp.

stlaz: Thanks, fixed.

simo5:
Do we really care for calculating the expiration time ?
Should we just set timestamp to 0 or even remove the whole thing ?

stlaz:
If I read the code well, in a well-set-up cookie, during `store_session_cookie()` when `Cookie.get_named_cookie_from_string()` is called, the expiration gets normalized which basically means removing the `Max-Age` attribute and replacing it with the `Expires` attribute in the cookie string (see `Cookie.normalize_expiration()` and `Cookie.__str__()`). When later retrieving the cookie, it should not have the `Max-Age` attribute anymore but only `Expires`. Therefore we need to calculate it or change the way `normalize_expiration()` behaves.

simo5:
Ok, sorry for some reason I thought this was on the server side, where we do not care what the cookie looks like, but on the client side we indeed care.

apophys:
Hi, can this PR get little more attention? The issue seems to be a cause for a lot of failures in our integration tests. (I'm not 100% sure though)

simo5:
LGTM, please merge

HonzaCholasta:
master:

* 24eeb4d6a3be678d652247a4a862ffde037514da Fix cookie with Max-Age processing

PR #533, opened 2017-03-02
WebUI: Change structure of Identity submenu

pvomacka:
Previously there were 'User Groups', 'Host Groups' and 'Netgroups'
separately, now these three items are grouped into one named 'Groups'
which has sidebar with three items mentioned above.

This change allows us to move ID views into Identity submenu.

https://pagure.io/freeipa/issue/6717

### DIFF
--- a/install/ui/src/freeipa/group.js
+++ b/install/ui/src/freeipa/group.js
@@ -32,7 +32,17 @@ define([
         './entity'],
     function(on, IPA, $, phases, reg) {
 
-var exp = IPA.group = {};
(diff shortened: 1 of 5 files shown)


simo5:
I do not have enough insights on the .js side to say this is all correct, but having seen the mockups I want to give an ack from my side here.

tomaskrizek:
master:

* 070bc48dd6c9bce32caa0f0f2de8d44b4e5bbbb1 WebUI: Change structure of Identity submenu

PR #534, opened 2017-03-02
Move csrgen templates into ipaclient package

tiran:
csrgen broke packaging of ipaclient for PyPI. All csrgen related
resources are now package data of ipaclient package. Package data is
accessed with Jinja's PackageLoader() or through pkg_resources.

https://pagure.io/freeipa/issue/6714


tiran:
@LiptonB please have a look.

ipaclient/setup.py
```diff
@@ -56,5 +63,6 @@
         extras_require={
             "install": ["ipaplatform"],
             "otptoken_yubikey": ["yubico", "usb"]
-        }
+        },
+        zip_safe=False,
```
LiptonB: Is this needed? According to http://setuptools.readthedocs.io/en/latest/setuptools.html#setting-the-zip-safe-flag "And if the project uses pkg_resources for all its data file access, then C extensions and other data files shouldn’t be a problem at all." `zip_safe=true` *seems* to work when I build it locally, but I'm not sure if I'm testing it properly.

LiptonB:
Oops, sorry about the breakage. This seems fine to me, although I hadn't really been thinking of the templates and rules as data files. They're intended to be possible to modify, more like config files. (Come to think of it, `/usr/share` wasn't that appropriate for them either). So, that and the fact that they're now duplicated between `python2.*/site-packages` and `python3.*/site-packages` give me pause (especially if the user might edit them), but I don't have strong feelings about it.

tiran:
In my opinion, a user should never modify a file that managed by a package manager and not explicitly marked as a config file. Both files in ```/usr/share``` and ```site-packages``` are not config files.

How about http://jinja.pocoo.org/docs/2.9/api/#jinja2.ChoiceLoader and this idea?

```
loader = jinja2.ChoiceLoader(
    jinja2.FileSystemLoader(os.path.join(api.env.conf_dir, 'csrgen/templates')),
    jinja2.PackageLoader('ipaclient', 'csrgen/templates'),
)
```

This allows users to override the templates by copying them to ```/etc/ipa/csrgen/templates```. We'd need similar code for the JSON files, too.

ipaclient/setup.py
tiran: Ah, Jinja uses a pkg_resources ResourceManager to access files. So in theory, zip loading would work.

In practice the option and feature is longer relevant. Python eggs are a dying species and Python wheels no longer use zip deployment.

LiptonB:
I think this is a much better way to make it configurable than how I had it, and the implementation looks good to me. Thanks!

dkupka:
master:

* 80be18162921268be9c8981495c9e8a4de0c85cd Move csrgen templates into ipaclient package
* 177f07e163d6d591a1e609d35e0a6f6f5347551e Chain CSR generator file loaders

PR #535, opened 2017-03-02
add whoami command

abbra:
`ipa whoami` command allows to query details about currently
authenticated identity. The command returns following information:

  * object class name
  * function to call to get actual details about the object
  * arguments to pass to the function
  * options to pass to the function

There are five types of objects that could bind to IPA using their
credentials. `ipa whoami` call expects one of the following:

  * users
  * staged users
  * hosts
  * Kerberos services
  * ID user override from the default trust view

The latter category of objects is automatically mapped by SASL GSSAPI
mapping rule in 389-ds for users from trusted Active Directory forests.

The command is expected to be used by Web UI to define proper view for
the authenticated identity.

Below is an example of how communication looks like for an Active
Directory user which has ID override in 'Default Trust View'.

    $ ipa -vv whoami
    ipa: INFO: trying https://ipa.example.com/ipa/session/json
    ipa: INFO: Forwarding 'whoami/1' to json server 'https://ipa.example.com/ipa/session/json'
    ipa: INFO: Request: {
        "id": 0,
        "method": "whoami/1",
        "params": [
            [],
            {
                "version": "2.220"
            }
        ]
    }
    ipa: INFO: Response: {
        "error": null,
        "id": 0,
        "principal": "Administrator@AD.DOMAIN",
        "result": {
            "arguments": [
                "default trust view",
                "administrator@ad.domain"
            ],
            "details": "idoverrideuser_show/1",
            "object": "idoverrideuser",
            "options": []
        },
        "version": "<IPA VERSION>"
    }

Fixes https://pagure.io/freeipa/issue/6643

### DIFF
--- a/ipaserver/plugins/whoami.py
+++ b/ipaserver/plugins/whoami.py
@@ -1,0 +1,136 @@
+#
+# Copyright (C) 2017  FreeIPA Contributors see COPYING for license
+#
+
+import six
+from ipalib import api, Command, errors, output, Str
+from ipalib import _
+from ipapython.dn import DN
+from ipalib.plugable import Registry
+from .idviews import DEFAULT_TRUST_VIEW_NAME
+
+if six.PY3:
+    unicode = str
+
+__doc__ = _("""
+Return information about currently authenticated identity
+
+Who am I command returns information on how to get
+more details about the identity authenticated for this
+request. The information includes:
+
+ * type of object
+ * command to retrieve details of the object
+ * arguments and options to pass to the command
+
+The information is returned as a dictionary. Examples below use
+'key: value' output for illustrative purposes.
+
+EXAMPLES:
+
+ Look up as IPA user:
+   kinit admin
+   ipa console
+   >> api.Command.whoami()
+   ------------------------------------------
+   object: user
+   command: user_show/1
+   arguments: admin
+   ------------------------------------------
+
+ Look up as a user from a trusted domain:
+   kinit user@AD.DOMAIN
+   ipa console
+   >> api.Command.whoami()
+   ------------------------------------------
+   object: idoverrideuser
+   command: idoverrideuser_show/1
+   arguments: ('default trust view', 'user@ad.domain')
+   ------------------------------------------
+
+ Look up as a host:
+   kinit -k
+   ipa console
+   >> api.Command.whoami()
+   ------------------------------------------
+   object: host
+   command: host_show/1
+   arguments: ipa.example.com
+   ------------------------------------------
+
+ Look up as a Kerberos service:
+   kinit -k -t /path/to/keytab HTTP/ipa.example.com
+   ipa console
+   >> api.Command.whoami()
+   ------------------------------------------
+   object: service
+   command: service_show/1
+   arguments: HTTP/ipa.example.com
+   ------------------------------------------
+""")
+
+register = Registry()
+
+
+@register()
+class whoami(Command):
+    __doc__ = _('Describe currently authenticated identity.')
+
+    NO_CLI = True
+
+    output_params = (
+        Str('object', label=_('Object class name')),
+        Str('command', label= _('Function to get details')),
+        Str('arguments*', label=_('Arguments to details function')),
+    )
+
+    has_output = (
+        output.Output('object', unicode, _('Object class name')),
+        output.Output('command', unicode, _('Function to get details')),
+        output.Output('arguments', list, _('Arguments to details function')),
+    )
+
+    def execute(self, **options):
+        """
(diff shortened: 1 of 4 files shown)


abbra:
Design page: http://www.freeipa.org/page/V4/Who_Am_I_Command

stlaz:
I believe that in CLI `ipa whoami` should actually output the output of the command it found with the arguments and options it found since in WebUI this is eventually done as well.
I can go ahead and try to implement it if we can agree on such behavior.

abbra:
Uhm, no, I don't want that. It makes the command behaving differently depending on a context and that would be broken. For client-side plugin that would also be an abuse of interface, I'd say.

stlaz:
Ok. It just doesn't seem right to have a command which shows something that's not immediately useful to the user. I am not sure whether we should have it enabled for CLI.

abbra:
We can disable it for CLI, that's not a problem.

ipaserver/plugins/whoami.py
```diff
@@ -0,0 +1,141 @@
...
+            if container is None:
+                continue
+            # Adjust container for exception two-level objects
+            if o.name in exceptions:
+                container = exceptions[o.name][0] + container
+            if dn.find(container + api.env.basedn) == 1:
+                # We found exact container this DN belongs to
+                o_name = unicode(o.name)
+                o_args = [unicode(entry.single_value.get(o.primary_key.name))]
+                o_opts = []
```
stlaz: There is no reason to have `options` in the output since you never add anything there, please, remove it.

stlaz:
Please, disable the command for CLI since by itself it does not bring any valuable information to the user, in my opinion it can only confuse them. It'll be easier to turn it on if there's a demand for it rather than turning it off if it confuses some people.
Also, please rename the `details` part of output to `command` as that's what it really is, `details` is confusing.
Please, also remove the `options` from the output since it's not currently used. I hope it's safe to assume that new output parameters can be added at any time.
You don't need to bother much with the other comments I left, there're more important things to do now than fiddling with docs and minor performance issues.

ipaserver/plugins/whoami.py
```diff
@@ -0,0 +1,141 @@
...
+from ipalib.plugable import Registry
+from .idviews import DEFAULT_TRUST_VIEW_NAME
+
+if six.PY3:
+    unicode = str
+
+__doc__ = _("""
+Return a description of currently authenticated identity
+
+Who am I command returns information on who to get
```
abbra: Fixed.

ipaserver/plugins/whoami.py
```diff
@@ -0,0 +1,141 @@
...
+        output.Output('options', list, _('Options to details function')),
+    )
+
+    def execute(self, **options):
+        """
+        Retrieve the DN we are authenticated as to LDAP and find bindable IPA
+        object that handles the container where this DN belongs to. Then report
+        details about this object.
+        """
+        exceptions = dict(
```
abbra: Fixed.

ipaserver/plugins/whoami.py
```diff
@@ -0,0 +1,141 @@
...
+        if dn == DN('cn=Directory Manager'):
+            raise errors.NotFound(
+                    reason=_('Cannot query Directory Manager with API'))
+
+        entry = ldap.get_entry(dn)
+        o_name = None
+        o_func = None
+        o_args = []
+        o_opts = []
+        for o in api.Object():
```
abbra: There is no need to init it because api is fully initialized already at the point when any plugin command is run. api.Object() returns an iterator, it is *not* a class constructor.

ipaserver/plugins/whoami.py
abbra: On further thinking I removed it.

abbra:
Updated.

stlaz:
@abbra Thank you for the changes, the patch seems fine now. I tested the user/service/host scenarios and it worked fine. I couldn't test idviews since trusts are broken now but I assume it should work fine as well.
If you could only apply the following patch https://transfer.sh/IA7Ic/0001-improve-one-more-dict.patch which improves the last `dict()` behavior then I'll bless this patch :)
We may want to add some tests later so I will propose to leave the ticket open.

abbra:
Done. I've also updated the design page to reflect the changes.

stlaz:
Thank you, ACK. Please don't close the ticket, we still need tests.

MartinBasti:
master:

* 381c1c7a8fe63526d21cb65decb75fb5ffda676a add whoami command

MartinBasti:
Tests: https://pagure.io/freeipa/issue/6745

PR #536, opened 2017-03-03
ipa systemd unit should define Wants=network instead of Requires=network

flo-renaud:
The file ipa.service defines
    Requires=network.target
which means that ipa stack will be restarted each time the network stack
is restarted. This is not needed, and Wants=network.target will be sufficient.

https://fedorahosted.org/freeipa/ticket/6723


tomaskrizek:
master:

* f447489707812643ee918266f99ca1ac82a408af ipa systemd unit should define Wants=network instead of Requires=network

PR #539, opened 2017-03-06
Define errors_by_code in ipalib.errors

frasertweedale:
The errors_by_code mapping will soon be used in more places, as part
of the Dogtag GSS-API authentication work.  Move its definition to
ipalib.errors.

### DIFF
--- a/ipalib/errors.py
+++ b/ipalib/errors.py
@@ -1995,5 +1995,7 @@ class GenericError(PublicError):
 public_errors = tuple(sorted(
     messages.iter_messages(globals(), PublicError), key=lambda E: E.errno))
 
+errors_by_code = dict((e.errno, e) for e in public_errors)
+
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


HonzaCholasta:
The codes are only relevant for XML-RPC. Why do you need them?

frasertweedale:
@HonzaCholasta when Dogtag execute the existing cert-request validation logic (which will be extracted to a new function), if an exception gets raised Dogtag returns it in the response, and IPA reconstructs it, so that there is no change to the user experience.

HonzaCholasta:
Seems like an overkill for the ~2 types of errors which the code uses. Anyway, I would rather wait before making any kind of decision based on cert-request code until after it is refactored to be less insane (https://pagure.io/freeipa/issue/6531).

frasertweedale:
Righto.  I'll withdraw this PR for now and it will make a comeback closer to landing the gssapi work.

PR #540, opened 2017-03-06
rabase.get_certificate: make serial number arg mandatory

frasertweedale:
In rabase.get_certificate it does not make sense for the
serial_number argument to be optional.  Make it a mandatory
positional argument.


tomaskrizek:
master:

* 3ba0375c831eca673c2df146b565a32dbc03fdb3 rabase.get_certificate: make serial number arg mandatory

PR #541, opened 2017-03-06
We don't offer no quickies

stlaz:
It's not our main priority as developers to offer any forms of quickies
nor guides on how to perform them.

http://www.urbandictionary.com/define.php?term=quickie


dkupka:
master:

* 30d7c210a4d153fcb5007651a80d8d53512abba3 We don't offer no quickies

PR #543, opened 2017-03-06
Add options to allow ticket caching

simo5:
This new option (planned to land in gssproxy 0.7) we cache the ldap
ticket properly and avoid a ticket lookup to the KDC on each and every
ldap connection. (Also requires krb5 libs 1.15.1 to benefit from caching).

NOTE: It is safe to apply this to master, if gssproxy does not support this option it simply is ignored.

### DIFF
--- a/install/share/gssproxy.conf.template
+++ b/install/share/gssproxy.conf.template
@@ -4,6 +4,7 @@
   cred_store = keytab:$HTTP_KEYTAB
   cred_store = client_keytab:$HTTP_KEYTAB
   allow_protocol_transition = true
+  allow_client_ccache_sync = true
   cred_usage = both
   euid = $HTTPD_USER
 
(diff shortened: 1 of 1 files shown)


abbra:
LGTM. Here I'd also like to bump gssproxy and krb5 dependencies in the spec file. We need to ensure gssproxy is actually updated.

simo5:
Yes, I think we should add a new PR later once we release gssproxy 0.7

MartinBasti:
Bump

tiran:
@simo5 please resolve the merge conflict

simo5:
@MartinBasti can we push this ? It makes a big difference in framework performance and load on the KDC

martbab:
I think that we have all dependencies in spec already so I do not see a reason not to.

martbab:
master:

* 4ee7e4ee6d6500d8b8935c9033388adc4cdbe672 Add options to allow ticket caching

HonzaCholasta:
@martbab, the ticket says 4.5.1, but this was not pushed to ipa-4-5.

HonzaCholasta:
ipa-4-5:

* 62d39385e20b3e1b059466f37cc063833355551e Add options to allow ticket caching

PR #544, opened 2017-03-07
Don't use weak ciphers for client HTTPS connections

stlaz:
https://pagure.io/freeipa/issue/6730


tomaskrizek:
master:

* fda22c33441d3b2c541a272e411ac1503a20fb01 Don't use weak ciphers for client HTTPS connections

PR #546, opened 2017-03-07
Store session cookie in a ccache option

simo5:
Instead of using the kernel keyring, store the session cookie within the
ccache. This way kdestroy will really wipe away all crededntials.

Ticket: https://pagure.io/freeipa/issue/6661

### DIFF
--- a/ipapython/session_storage.py
+++ b/ipapython/session_storage.py
@@ -1,0 +1,197 @@
+#
+# Copyright (C) 2017  FreeIPA Contributors see COPYING for license
+#
+
+import ctypes
+
+
+KRB5_CC_NOSUPP = -1765328137
+
+
+try:
+    LIBKRB5 = ctypes.CDLL('libkrb5.so.3')
+except OSError as e:  # pragma: no cover
+    raise ImportError(str(e))
+
+
+class _krb5_context(ctypes.Structure):  # noqa
+    """krb5/krb5.h struct _krb5_context"""
+    _fields_ = []
+
+
+class _krb5_ccache(ctypes.Structure):  # noqa
+    """krb5/krb5.h struct _krb5_ccache"""
+    _fields_ = []
+
+
+class _krb5_data(ctypes.Structure):  # noqa
+    """krb5/krb5.h struct _krb5_data"""
+    _fields_ = [
+        ("magic", ctypes.c_int32),
+        ("length", ctypes.c_uint),
+        ("data", ctypes.c_char_p),
+    ]
+
+
+class krb5_principal_data(ctypes.Structure):  # noqa
+    """krb5/krb5.h struct krb5_principal_data"""
+    _fields_ = []
+
+
+class KRB5Error(Exception):
+    pass
+
+
+def krb5_errcheck(result, func, arguments):
+    """Error checker for krb5_error return value"""
+    if result != 0:
+        raise KRB5Error(result, func.__name__, arguments)
+
+
+krb5_principal = ctypes.POINTER(krb5_principal_data)
+krb5_context = ctypes.POINTER(_krb5_context)
+krb5_ccache = ctypes.POINTER(_krb5_ccache)
+krb5_data_p = ctypes.POINTER(_krb5_data)
+krb5_error = ctypes.c_int32
+
+krb5_init_context = LIBKRB5.krb5_init_context
+krb5_init_context.argtypes = (ctypes.POINTER(krb5_context), )
+krb5_init_context.restype = krb5_error
+krb5_init_context.errcheck = krb5_errcheck
+
+krb5_free_context = LIBKRB5.krb5_free_context
+krb5_free_context.argtypes = (krb5_context, )
+krb5_free_context.retval = None
+
+krb5_free_principal = LIBKRB5.krb5_free_principal
+krb5_free_principal.argtypes = (krb5_context, krb5_principal)
+krb5_free_principal.retval = None
+
+krb5_free_data_contents = LIBKRB5.krb5_free_data_contents
+krb5_free_data_contents.argtypes = (krb5_context, krb5_data_p)
+krb5_free_data_contents.retval = None
+
+krb5_cc_default = LIBKRB5.krb5_cc_default
+krb5_cc_default.argtypes = (krb5_context, ctypes.POINTER(krb5_ccache), )
+krb5_cc_default.restype = krb5_error
+krb5_cc_default.errcheck = krb5_errcheck
+
+krb5_cc_close = LIBKRB5.krb5_cc_close
+krb5_cc_close.argtypes = (krb5_context, krb5_ccache, )
+krb5_cc_close.retval = krb5_error
+krb5_cc_close.errcheck = krb5_errcheck
+
+krb5_parse_name = LIBKRB5.krb5_parse_name
+krb5_parse_name.argtypes = (krb5_context, ctypes.c_char_p,
+                            ctypes.POINTER(krb5_principal), )
+krb5_parse_name.retval = krb5_error
+krb5_parse_name.errcheck = krb5_errcheck
+
+krb5_cc_set_config = LIBKRB5.krb5_cc_set_config
+krb5_cc_set_config.argtypes = (krb5_context, krb5_ccache, krb5_principal,
+                               ctypes.c_char_p, krb5_data_p, )
+krb5_cc_set_config.retval = krb5_error
+krb5_cc_set_config.errcheck = krb5_errcheck
+
+krb5_cc_get_config = LIBKRB5.krb5_cc_get_config
+krb5_cc_get_config.argtypes = (krb5_context, krb5_ccache, krb5_principal,
+                               ctypes.c_char_p, krb5_data_p, )
+krb5_cc_get_config.retval = krb5_error
+krb5_cc_get_config.errcheck = krb5_errcheck
+
+
+def store_data(princ_name, key, value):
+    """
+    Stores the session cookie in a hidden ccache entry.
+    """
+    context = krb5_context()
+    principal = krb5_principal()
+    ccache = krb5_ccache()
+
+    try:
+        krb5_init_context(ctypes.byref(context))
+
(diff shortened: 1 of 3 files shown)


ipapython/ccache_storage.py
```diff
@@ -0,0 +1,234 @@
...
+            value = self.value
+            if value is None:
+                return None
+            elif not isinstance(value, str):
+                return value.decode('utf-8')
+            return value
+
+    class _krb5_context(ctypes.Structure):  # noqa
+        """krb5/krb5.h struct _krb5_context"""
+        __slots__ = ()
```
MartinBasti: Why `__slots__` are defined? Is it performance improvement?

ipapython/ccache_storage.py
```diff
@@ -0,0 +1,234 @@
...
+except OSError as e:  # pragma: no cover
+    LIBKRB5 = e
+else:
+    class c_text_p(ctypes.c_char_p):  # noqa
+        """A c_char_p variant that can handle UTF-8 text"""
+        @classmethod
+        def from_param(cls, value):
+            if value is None:
+                return None
+            if PY3 and isinstance(value, str):
```
MartinBasti: Can be this replaced with:
```
if not isinstance(value, bytes):
    return value.encode('utf-8')
```
?

ipapython/ccache_storage.py
```diff
@@ -0,0 +1,234 @@
...
+    krb5_cc_set_config.retval = krb5_error
+    krb5_cc_set_config.errcheck = krb5_errcheck
+
+    krb5_cc_get_config = LIBKRB5.krb5_cc_get_config
+    krb5_cc_get_config.argtypes = (krb5_context, krb5_ccache, krb5_principal,
+                                   ctypes.c_char_p, krb5_data_p, )
+    krb5_cc_get_config.retval = krb5_error
+    krb5_cc_get_config.errcheck = krb5_errcheck
+
+class session_store:
```
MartinBasti: Should be `class session_store(object):`

ipapython/ccache_storage.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, type, value, traceback):
+        if self.__context:
+            krb5_free_context(self.__context)
+            self.__context = None
+
+    def __del__(self):
```
MartinBasti: Don't define `__del__` method in python unless you have bulletproof evidence that it will work in your case.
https://docs.python.org/2/reference/datamodel.html#object.__del__

In python, if you want safely do cleanup, it must be done by contextmanager (what you have already implemented), don't use __del__ method

ipapython/ccache_storage.py
```diff
@@ -0,0 +1,234 @@
...
+                            ctypes.byref(principal))
+
+            ccache = krb5_ccache()
+            krb5_cc_default(self.__context, ctypes.byref(ccache))
+
+            data = _krb5_data()
+            krb5_cc_get_config(self.__context, ccache, principal,
+                               self._hidden_cred_name, ctypes.byref(data))
+
+            return str(data.data)
```
MartinBasti: What is `data.data` type? If bytes this will return unexpected string  in py3. `.decode()` should be used in case of bytes

ipapython/ccache_storage.py
```diff
@@ -0,0 +1,234 @@
...
+    pass
+
+
+PY3 = sys.version_info[0] == 3
+
+
+try:
+    LIBKRB5 = ctypes.CDLL('libkrb5.so.3')
+except OSError as e:  # pragma: no cover
+    LIBKRB5 = e
```
MartinBasti: Shouldn't we rather fail early and raise ImportError ?

MartinBasti:
Pylint failed and I have a few inline comments
```
************* Module ipapython.ccache_storage
ipapython/ccache_storage.py:234: [C0305(trailing-newlines), ] Trailing newlines)
ipapython/ccache_storage.py:32: [W1612(unicode-builtin), c_text_p.from_param] unicode built-in referenced)
ipapython/ccache_storage.py:45: [E1101(no-member), c_text_p.text] Class 'value' has no 'decode' member)
ipapython/ccache_storage.py:128: [C1001(old-style-class), session_store] Old-style class defined.)
ipapython/ccache_storage.py:132: [E0710(raising-non-exception), session_store.__init__] Raising a new style class which doesn't inherit from BaseException)
ipapython/ccache_storage.py:6: [W0611(unused-import), ] Unused import os)
```

ipapython/ccache_storage.py
```diff
@@ -0,0 +1,234 @@
...
+import sys
+
+import six
+
+
+class KRB5Error(Exception):
+    pass
+
+
+PY3 = sys.version_info[0] == 3
```
simo5: ok

ipapython/ccache_storage.py
simo5: I guess we could

ipapython/ccache_storage.py
simo5: turns out I am not actually using c_text_p anywhere in the end, so I'm removing all of this

ipapython/ccache_storage.py
simo5: it's defined empty, but I'll remove them altogether

rcritten:
Should this patch not also remove the keyring code?

Unit tests should be provided.

ipapython/ccache_storage.py
simo5: removed

ipapython/ccache_storage.py
simo5: it is decalred as c_char_p

simo5:
@rcritten the keyring stuff is still used for detection of keyring in other places, so I did not touch it as those uses are still vaild

simo5:
Not sure how to provide unit tests, these functions work only if you have a valid ccache in the name of the principal you are trying to store a session cookie for.

simo5:
Ok removed a bunch of code and made sure pylint passes.

simo5:
I also renamed the module and the class, makes more sense to me this way around.

ipapython/session_storage.py
```diff
@@ -0,0 +1,193 @@
...
+krb5_cc_get_config = LIBKRB5.krb5_cc_get_config
+krb5_cc_get_config.argtypes = (krb5_context, krb5_ccache, krb5_principal,
+                               ctypes.c_char_p, krb5_data_p, )
+krb5_cc_get_config.retval = krb5_error
+krb5_cc_get_config.errcheck = krb5_errcheck
+
+
+class ccache_store(object):
+    def __init__(self, name='X-IPA-Session-Cookie'):
+        self.__context = None
```
tiran: Please use single underscore here. Double underscore lead to name mangling, which makes debugging harder.

ipapython/session_storage.py
```diff
@@ -0,0 +1,193 @@
...
+        context = krb5_context()
+        krb5_init_context(ctypes.byref(context))
+        self.__context = context
+
+        self._hidden_cred_name = name
+
+    def __enter__(self):
+        return self
```
tiran: context should be initialized in ```__enter__```, not in ```__init__```

MartinBasti: Actually `ccache_store` is not used together with `with` statement, what is your goal @simo5?

simo5: I just copied example code from krbproxy, I do not have any special goal.
I could even not carry a global context but crate and destroy contexts at each call
Whatever is preferred.

simo5:
Ok I decide to do away with the whole class stuff, given we never really keep a round the class object for more than one operation at a time in actual use.
As @rcritten requested I also added a test, and I am glad it was asked as I found a failure case we need to handle (see the exception handling in remove_data()

ipatests/test_ipapython/test_session_storage.py
```diff
@@ -0,0 +1,40 @@
+#
+# Copyright (C) 2017  FreeIPA Contributors see COPYING for license
+#
+
+"""
+Test the `session_storage.py` module.
+"""
+
+from nose.tools import raises  # pylint: disable=E0611
```
MartinBasti: it is unused and instead of nose.raises please use pytest raises

MartinBasti:
```
************* Module ipapython.session_storage
ipapython/session_storage.py:187: [W1624(indexing-exception), remove_data] Indexing exceptions will not work on Python 3)
************* Module ipalib.rpc
ipalib/rpc.py:114: [E1120(no-value-for-parameter), read_persistent_client_session_data] No value for argument 'value' in function call)
************* Module ipatests.test_ipapython.test_session_storage
ipatests/test_ipapython/test_session_storage.py:39: [W0612(unused-variable), test_session_storage.test_03] Unused variable 'e')
ipatests/test_ipapython/test_session_storage.py:9: [W0611(unused-import), ] Unused raises imported from nose.tools)
ipatests/test_ipapython/test_session_storage.py:12: [W0611(unused-import), ] Unused import pytest)
```

simo5:
Oops sorry, forgot to run make pylint on my last iteration, should be all fixed now

ipatests/test_ipapython/test_session_storage.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def setup(self):
+        # TODO: set up test user and kinit to it
+        # tmpdir = tempfile.mkdtemp(prefix = "tmp-")
+        # os.environ['KRB5CCNAME'] = 'FILE:%s/ccache' % tmpdir
+        self.principal = 'admin'
+        self.key = 'X-IPA-test-session-storage'
+        self.data = 'Test Data'
+
+    def test_01(self):
```
MartinBasti: Tests has no description

MartinBasti: but they are somehow selfexplanatory

MartinBasti:
master:

* 7cab95955539b5f9596dcda5886ea3d9755fb193 Store session cookie in a ccache option

PR #547, opened 2017-03-07
Use GSS-SPNEGO if connecting locally

simo5:
GSS-SPNEGO allows us to negotiate a SASL bind with less roundtrips
therefore use it when possible.

We only enable it for local connections for now because we only
recently fixed Cyrus SASL to do proper GSS-SPNEGO negotiation. This
change means a newer and an older version are not compatible.

Restricting ourselves to the local host prevents issues with
incompatible services, and it is ok for us as we are only really
looking for speedups for the local short-lived connections performed
by the framework. Most other clients have longer lived connections,
so peformance improvements there are not as important.

### DIFF
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -52,6 +52,7 @@ if six.PY3:
 
 # Global variable to define SASL auth
 SASL_GSSAPI = ldap.sasl.sasl({}, 'GSSAPI')
+SASL_GSS_SPNEGO = ldap.sasl.sasl({}, 'GSS-SPNEGO')
 
 _debug_log_ldap = False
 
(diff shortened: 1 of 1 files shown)


abbra:
LGTM but I think we should also update Requires: in the spec file to use cyrus-sasl-2.1.26-29.fc26 or later.

simo5:
We actually do not need to put a strong require, this patch will work regardless, but won't provide any performance advantage on older versions.

You will add a stronger require when the GC work is done, so we can defer to that point to add it.

tomaskrizek:
The patch works with both `cyrus-sasl-2.1.26-26.2.fc24` and `cyrus-sasl-2.1.26-29.fc26`.

Since the newer version is not a hard dependency, we can add it later on, as @simo5 suggested.

tomaskrizek:
master:

* adf8aabf10a57383aa6216625921503b83575757 Use GSS-SPNEGO if connecting locally

PR #548, opened 2017-03-07
ipa-server-install: add --setup-kra option

MartinBasti:


ipaserver/install/server/replicainstall.py
```diff
@@ -836,14 +836,12 @@ def install_check(installer):
             root_logger.debug('No IPA DNS servers, '
                               'skipping forward/reverse resolution check')
 
-        kra_enabled = remote_api.Command.kra_is_enabled()['result']
```
tomaskrizek: If I understand this correctly, before this patch, KRA would be installed automatically if it was present on the master server. With this patch, `--setup-kra` would have to be explicitly passed to `ipa-replica-install`.

Is this the case or am I missing something? Is this behaviour change intended?

ipaserver/install/server/install.py
```diff
@@ -384,6 +384,8 @@ def install_check(installer):
     print("  * Create and configure an instance of Directory Server")
     print("  * Create and configure a Kerberos Key Distribution Center (KDC)")
     print("  * Configure Apache (httpd)")
+    if options.setup_kra:
```
tomaskrizek: Please update the man pages with `--setup-kra` option for both `ipa-server-install` and `ipa-replica-install`.

ipaserver/install/server/replicainstall.py
stlaz: No, you don't understand it. KRA would not be installed. See `kra.py:install()` for details.

HonzaCholasta:
NACK on the "KRA: run install and install_check only when KRA should be installed" commit. The end goal for all component installers is to make them isolated and handle their options themselves, so that they can be packaged separately (among other things). This commit takes the code in the opposite direction. Also it does not make the code more readable because it is inconsistent with the CA installer.

ipaserver/install/server/__init__.py
```diff
@@ -159,7 +159,6 @@ def domain_level(self, value):
         None,
         description="configure a dogtag KRA",
     )
-    setup_kra = enroll_only(setup_kra)
```
HonzaCholasta: Why did you remove this line? It must be here, otherwise `--setup-kra` will appear in `ipa-replica-prepare` once it's ported to the install framework.

ipaserver/install/server/replicainstall.py
MartinBasti: commit removed

ipaserver/install/server/install.py
MartinBasti: done

ipaserver/install/server/__init__.py
MartinBasti: I somehow thought that it will not work, reverted.

MartinBasti:
Given that there is no time, I dropped commit you NACKed as it unneeded for this PR, but please note my disagreement about a way how `kra.py` handles `--setup-kra` option for the future release.

dkupka:
master:

* 4006cbbc02c368ac9e5e3721613158decb34fd37 KRA: add --setup-kra to ipa-server-install
* 25fa2bb6c9fa1b498330b13c9a6116b646eb75ba tests: use --setup-kra in tests

PR #232, opened 2016-11-11
Installer refactoring

jcholast:
This PR contains the installer refactoring work we (@jcholast, @martbab, @mbasti-rh, @stlaz) did in the last 6 weeks.

What is included:
- Replica install workflows were updated to more closely resemble each other.
- Some classic replica installation and replica promotion code was merged.
- Client installer was turned into a module.
- Command line options shared between multiple installers are now defined in one place.

What is missing:
- Use GSSAPI authentication for initial replication on domain level 0 (jcholast/freeipa#22).
- Merge the rest of classic replica installation and replica promotion code.
- Call into the client installer module instead of executing `ipa-client-install` in server installers.
- Update the remaining installers (`ipa-replica-prepare`, `ipa-ca-install`, `ipa-kra-install`, `ipa-dns-install`) to make use of the single option definition.

Known issues:
- Ugly help in `ipa-server-install`, `ipa-replica-install` and `ipa-client-install`.

https://fedorahosted.org/freeipa/ticket/6392

Note that the commits in this PR were already reviewed and ACKed over at [jcholast/freeipa](https://github.com/jcholast/freeipa) ([list of PRs](https://github.com/jcholast/freeipa/pulls?q=is%3Aclosed+label%3Ainstaller-refactoring)).

If you find any issues, please file a Trac ticket or leave a comment in this PR.

PR #550, opened 2017-03-08
install: fix help

HonzaCholasta:
This PR fixes the known issue in the installer refactoring PR #232 and concludes https://pagure.io/freeipa/issue/6392.

**server install: remove duplicate -w option**

Remove duplicate -w alias of --admin-password in ipa-server-install and
ipa-replica-install.

**install: add missing space in realm_name description**

**server install: remove duplicate knob definitions**

Remove duplicate definitions of knobs already defined in client install.

**client install: split off SSSD options into a separate class**

Split off SSSD knob definitions from the ClientInstallInterface class into
a new SSSDInstallInterface class.

**install CLI: remove magic option groups**

Do not automatically create the "basic options" and "uninstall options"
option groups in the CLI code.

**install: re-introduce option groups**

Re-introduce option groups in ipa-client-install, ipa-server-install and
ipa-replica-install.

https://pagure.io/freeipa/issue/6392


ipapython/install/core.py
```diff
@@ -146,15 +150,19 @@ def validate(self, value):
...
-    if type_ is None:
-        type_ = NoneType
+_type = type
+
+
+def _knob(type=_missing, default=_missing, bases=_missing, _order=_missing,
+          sensitive=_missing, deprecated=_missing, description=_missing,
+          group=_missing, cli_names=_missing, cli_deprecated_names=_missing,
+          cli_metavar=_missing):
+    if type is None:
```
MartinBasti: Can we avoid shadowing of python builtins?

ipapython/install/core.py
```diff
@@ -189,6 +200,31 @@ def knob(type_=_missing, default=_missing, bases=_missing, sensitive=_missing,
...
+def knob(type, default=_missing, **kwargs):
+    return _knob(
+        type, default,
+        _order=next(_counter),
+        **kwargs
+    )
+
+
+def extend_knob(base, default=_missing, bases=_missing, group=_missing,
+                **kwargs):
```
stlaz: There shall be at least a short docstring for future humans who try to find out what happened here.

ipaserver/install/adtrust.py
```diff
@@ -451,6 +454,7 @@ class ADTrustInstallInterface(ServiceAdminInstallInterface):
         description="Add IPA masters to a list of hosts allowed to "
                     "serve information about users from trusted forests"
     )
+    add_agents = replica_install_only(add_agents)
```
stlaz: Is this fixing some kind of a regression? I don't mind if it's fixed in this commit, I would just like to know.

ipaclient/install/sssd.py
```diff
@@ -0,0 +1,52 @@
...
+
+    fixed_primary = knob(
+        None,
+        description="Configure sssd to use fixed server as primary IPA server",
+    )
+    fixed_primary = enroll_only(fixed_primary)
+
+    permit = knob(
+        None,
+        description="disable access rules by default, permit all access.",
```
HonzaCholasta: This is how it was cased before, I did not (intentionally) change any of the descriptions.

ipapython/install/core.py
HonzaCholasta: Because the `Knob` attribute name is `type`. IMHO shadowing would be an issue only if `type` was called in the function, which is not the case here. BTW `type` is used as an argument name even in the Python standard library, on numerous occasions.

ipapython/install/cli.py
```diff
@@ -136,22 +136,21 @@ def add_options(cls, parser, positional=False):
                 help="unattended (un)installation never prompts the user",
             )
 
-        basic_group = optparse.OptionGroup(parser, "basic options")
-
         groups = collections.OrderedDict()
-        groups[None] = basic_group
+        groups[None] = parser
```
HonzaCholasta: `None` does not have `.add_option()`.

ipapython/install/core.py
```diff
@@ -168,17 +176,20 @@ def knob(type_=_missing, default=_missing, bases=_missing, sensitive=_missing,
...
 
     class_dict = {}
-    class_dict['_order'] = next(_counter)
-    if type_ is not _missing:
-        class_dict['type'] = type_
+    if type is not _missing:
+        class_dict['type'] = type
     if default is not _missing:
         class_dict['default'] = default
+    if _order is not _missing:
```
HonzaCholasta: It will be inherited from the parent.

ipapython/install/core.py
HonzaCholasta: OK.

ipaserver/install/adtrust.py
HonzaCholasta: This is just a cosmetic fix - the knob was redefined in `ServerInstallInterface` only to decorate it with `replica_install_only`, which is a) completely unnecessary b) inconsistent with other component installers.

ipaclient/install/sssd.py
stlaz: I would consider changing it a small doc fix, similar to PEP-8 fixes for Python. But then again I don't mind this in the end, I would probably ACK this even if you rewrote all the help to Pirate English.

ipapython/install/cli.py
stlaz: Oh, I see, so this will create some set of basic options. Please, could you add a comment there explaining it so that it's very clear?

stlaz:
Works like a charm, ACK.

MartinBasti:
master:

* 00f49dd7bbf277757902c94990d33758fec56b23 server install: remove duplicate -w option
* 5efa55c88d73d9f5db77df4be9fedf03f9b323d1 install: add missing space in realm_name description
* 94f362d7b0b6c838752eb2f6674149e96d3ae95b server install: remove duplicate knob definitions
* 1cfe06c79eb0b98a0f4bd663165156596b59e85f client install: split off SSSD options into a separate class
* 774d8d0a5dc0ac175ab0cecc76001632c2a79744 install CLI: remove magic option groups
* 2fc9feddd02bb17c3a9eb7efde83277fcf93252c install: re-introduce option groups

PR #552, opened 2017-03-08
man: add missing --setup-adtrust option to manpage

MartinBasti:
ipa-server-install and ipa-replica-install manpages miss --setup-adtrust
options

https://pagure.io/freeipa/issue/6630


martbab:
master:

* 6c95f33d37a2c346fc56d9890d594f1e40029c77 man: add missing --setup-adtrust option to manpage

PR #553, opened 2017-03-08
Add check for removing last KRA server

stlaz:
This patchset adds a check for removal of a last KRA server + adds a message about there only being one KRA to WebUI.

### DIFF
--- a/install/ui/src/freeipa/topology.js
+++ b/install/ui/src/freeipa/topology.js
@@ -501,23 +501,34 @@ topology.servers_search_facet = function(spec, no_init) {
             // Do not show warning if there is only one master
             if (result.length <= 1) return;
 
-            var counter = 0;
+            var ca_counter = 0;
+            var kra_counter = 0;
 
             for (var i=0, l=result.length; i<l; i++) {
                 var current = result[i];
                 var roles = current.result.enabled_role_servrole;
                 for (var k=0, m=roles.length; k<m; k++) {
-                    if (roles[k] === 'CA server') counter++;
+                    if (roles[k] === 'CA server') ca_counter++;
+                    if (roles[k] === 'KRA server') kra_counter++;
                 }
             }
 
-            // Create dialog and show it only when there is only one CA server
-            if (counter != 1) return;
+            // Show a dialog when there is only one CA or KRA server
+            if (ca_counter != 1 && kra_counter != 1) return;
 
-            var message = text.get('@i18n:objects.servers.ca_warning_message');
+            var message = text.get(
+                '@i18n:objects.servers.svc_warning_message') + '<ul>';
+            if (ca_counter === 1) {
+                message += '<li>CA</li>\n';
+            }
+            if (kra_counter === 1) {
+                message += '<li>KRA</li>\n';
+            }
+            message += '</ul>';
+
             var dialog = IPA.dialog({
-                name: 'ca_warning',
-                title: '@i18n:objects.servers.ca_warning_title',
+                name: 'svc_warning',
+                title: '@i18n:objects.servers.svc_warning_title',
                 sections: [
                     {
                         show_header: false,
--- a/install/ui/test/data/ipa_init.json
+++ b/install/ui/test/data/ipa_init.json
@@ -556,8 +556,8 @@
                             "label_singular": "Server Role",
                         },
                         "servers": {
-                            "ca_warning_message": "It is strongly recommended to keep the CA services installed on more than one server.",
-                            "ca_warning_title": "Warning: Only One CA Server Detected",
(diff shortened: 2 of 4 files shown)


stlaz:
Hm, I forgot that KRA is the only IPA component that has a standalone uninstaller, this is therefore only a partial fix.

MartinBasti:
@stlaz I wrote it to ticket

stlaz:
@MartinBasti ah, sorry, I completely overlooked it. The current PR version implements your suggestion.

MartinBasti:
Please create a separate commit for KRA Uninstall

stlaz:
Split done.

MartinBasti:
JFTR: KRA uninstall commit is here #556

install/ui/src/freeipa/topology.js
```diff
@@ -497,23 +497,40 @@ topology.servers_search_facet = function(spec, no_init) {
...
                 for (var k=0, m=roles.length; k<m; k++) {
-                    if (roles[k] === 'CA server') counter++;
+                    if (roles[k] === 'CA server') ca_counter++;
+                    if (roles[k] === 'KRA server') kra_counter++;
                 }
             }
 
             // Create dialog and show it only when there is only one CA server
-            if (counter != 1) return;
+            if (ca_counter != 1 && kra_counter != 1) return;
```
MartinBasti: This will show error when only one CA is installed, it should be `<1` and no KRA

install/ui/src/freeipa/topology.js
```diff
@@ -497,23 +497,40 @@ topology.servers_search_facet = function(spec, no_init) {
...
+                    field: false,
+                    $type: 'html',
+                    html: text.get('@i18n:objects.servers.kra_warning_message')
+                });
 
-            var message = text.get('@i18n:objects.servers.ca_warning_message');
             var dialog = IPA.dialog({
-                name: 'ca_warning',
-                title: '@i18n:objects.servers.ca_warning_title',
+                name: 'dogtag_warning',
```
MartinBasti: I don't like to have just one common title

MartinBasti: Can you dynamically create that list? something like `Warning: Only One {roles} Server Detected`

install/ui/src/freeipa/topology.js
MartinBasti: Nevermind, this was broken before your PR arrived

MartinBasti: But in this case, I there should be both warnings for CA and KRA, so it is broken by your patch

MartinBasti:
Probably you we should fix this before we double number of alerts
https://pagure.io/freeipa/issue/6598

install/ui/src/freeipa/topology.js
```diff
@@ -497,23 +497,40 @@ topology.servers_search_facet = function(spec, no_init) {
...
+                    if (roles[k] === 'KRA server') kra_counter++;
                 }
             }
 
             // Create dialog and show it only when there is only one CA server
-            if (counter != 1) return;
+            if (ca_counter != 1 && kra_counter != 1) return;
+
+            var messages = [];
+            if (ca_counter == 1)
```
pvomacka: Please use '===' instead od '==', two equals might cause weird behavior. 

And for better reading of code please use code block { } around commands after if statement. It will work if you have only one command after the condition, but in case that the command is not onliner it is better to have it in curly braces.

install/ui/src/freeipa/topology.js
```diff
@@ -497,23 +497,40 @@ topology.servers_search_facet = function(spec, no_init) {
...
+                messages.push(
+                {
+                    field: false,
+                    $type: 'html',
+                    html: text.get('@i18n:objects.servers.ca_warning_message')
+                });
+            if (kra_counter == 1)
```
pvomacka: The same as above.

pvoborni:
Fix for 6598 in #566

install/ui/src/freeipa/topology.js
stlaz: `==` can't cause this, we know already that the types are equal as `ca_counter` is a local variable that is not assigned from any external source.

install/ui/src/freeipa/topology.js
stlaz: Same answer as above.

stlaz:
Reworked how the beg for a service replication worked.

MartinBasti:
master:

* 670f8fb1db109ec2c9ab7e5d2189325988220b23 Add check to prevent removal of last KRA
* 1e8db4b5c7a55dac0008ad9b9bf5802ba30e8c2a Add message about last KRA to WebUI Topology view

PR #554, opened 2017-03-08
webui: fixes normalization of value in attributes widget

pvoborni:
Fix is in checkboxes widget but the only affected one is attributes widget.

Reproduction:
 1. Add permission with attribute with uppercase character
   $ ipa permission-add aa_test --type=stageuser --attrs=businessCategory --right=read
 2. Check if it is correctly displayed in Web UI

Actual result:
 - businesscategory is not checked
Expected result:
 - businesscategory is checked


dkupka:
master:

* 56a2642af0a29328df4defef138b9fa65624335a webui: fixes normalization of value in attributes widget

PR #558, opened 2017-03-09
ipapython: fix DEFAULT_PLUGINS in version.py

HonzaCholasta:
Replace the placeholder with the actual value during build.

This fixes the client incorrectly assuming that the default version of all
plugins is 1.

https://pagure.io/freeipa/issue/6597


MartinBasti:
master:

* abf25d3cb6570e6ae7cd094ea6a5f4a1bd75d8a7 ipapython: fix DEFAULT_PLUGINS in version.py

PR #560, opened 2017-03-09
rpcserver: x509_login: Handle unsuccessful certificate login gracefully

dkupka:
When mod_lookup_identity is unable to match user by certificate (and username)
it unsets http request's user. mod_auth_gssapi is then unable to get Kerberos
ticket and doesn't set KRB5CCNAME environment variable.
x509_login.__call__ now returns 401 in such case to indicate that request was
not authenticated.

### DIFF
--- a/ipaserver/rpcserver.py
+++ b/ipaserver/rpcserver.py
@@ -834,6 +834,16 @@ class login_kerberos(KerberosLogin):
 class login_x509(KerberosLogin):
     key = '/session/login_x509'
 
+    def __call__(self, environ, start_response):
+        self.debug('WSGI login_x509.__call__:')
+
+        if 'KRB5CCNAME' not in environ:
+            return self.unauthorized(
+                environ, start_response, 'KRB5CCNAME not set',
+                'Authentication failed')
+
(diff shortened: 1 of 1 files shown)


ipaserver/rpcserver.py
```diff
@@ -832,6 +832,18 @@ class login_kerberos(KerberosLogin):
 class login_x509(KerberosLogin):
     key = '/session/login_x509'
 
+    def __call__(self, environ, start_response):
+        self.debug('WSGI login_x509.__call__:')
+
+        try:
```
MartinBasti: I assume that this operation called always, so from performance POW it should be faster. Please use:

```python
if 'KRB5CCNAME' not in environ:
     return ...
```

Exception handling is expensive.
But we are talking about miliseconds so I'm fine with this too

flo-renaud:
Hi,
the invalid cert login correctly returns 401.

MartinBasti:
master:

* 70889d4d5e7e2bd65ab1d4a28e5eda4a51c9b0c0 rpcserver: x509_login: Handle unsuccessful certificate login gracefully

PR #561, opened 2017-03-09
ldap2: fix crash in development mode

HonzaCholasta:
Do not set or delete attributes directly on the ldap2 instance, as that
raises an AttributeError in development mode because of ReadOnly locking.

Use the usual workaround of `object.__setattr__` and `object.__delattr__`
to fix the issue.

https://pagure.io/freeipa/issue/6625

### DIFF
--- a/.test_runner_config.yaml
+++ b/.test_runner_config.yaml
@@ -48,6 +48,8 @@ steps:
   install_server:
   - ipa-server-install -U --domain ${server_domain} --realm ${server_realm} -p ${server_password}
     -a ${server_password} --setup-dns --setup-kra --auto-forwarders
+  - sed -ri "s/mode = production/mode = development/" /etc/ipa/default.conf
+  - systemctl restart httpd.service
   lint:
   - PYTHON=/usr/bin/python2 make V=0 lint
   - PYTHON=/usr/bin/python3 make V=0 pylint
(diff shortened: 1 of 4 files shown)


tomaskrizek:
Seems to work all right, but the locking issue still affects other parts of the code. For example, `ipa cert-show` in development mode fails with:

```
AttributeError: locked: cannot set ra_lightweight_ca.cookie 
```

I'm not sure if that's in the ticket's scope or not.

HonzaCholasta:
Let's see what Travis detects.

HonzaCholasta:
Travis didn't detect anything else, so I think we are good to go. Shall we keep the `.test_runner_config.yaml` change? (@martbab?)

martbab:
Yes but please split it into a separate patch so that CI-related changes are tracked separately and do not pollute the changes in the code.

HonzaCholasta:
Turns out unsetting mode in default.conf is not good enough, one has to explicitly set it to "development" to enable development mode, because "production" is the default.

Let's see if Travis detects something this time.

HonzaCholasta:
OK everything is green now.

MartinBasti:
master:

* 8fdd7a9ffc263c1198afa5479cda41d319f11d91 backend plugins: fix crashes in development mode
* fe4489ede2b40902fb7d734d04a1f997c6df86fb Travis CI: run tests in development mode

PR #562, opened 2017-03-09
[ipa-4-4] server install: require IPv6 stack to be enabled

tomaskrizek:
Add checks to install and replica install to verify IPv6 stack
is enabled. IPv6 is required by some IPA parts (AD, conncheck, ...).

https://pagure.io/freeipa/issue/6608


tomaskrizek:
ipa-4-4:

* a572e61cb5153df8a040757eaba0c47531f0fe85 server install: require IPv6 stack to be enabled

PR #563, opened 2017-03-09
backup: backup anonymous keytab

MartinBasti:
Freeipa stops working without anon keytab

https://pagure.io/freeipa/issue/5959


martbab:
master:

* 8fb61a55fe32438752567bde8af73d6b8230a386 backup: backup anonymous keytab

PR #564, opened 2017-03-09
Reconfigure Kerberos library config as the last step of KDC install

martbab:
During KDC installation, we overwrite the existing `/etc/krb5.conf` file
from client version to use only local KDC for client requests. However,
this means that services such as certmonger may try to kinit against
local KDC before it is up and running, resulting in subtle but serious
bugs.

The file should be updated only when KDC is set up properly and running.

https://pagure.io/freeipa/issue/6739

### DIFF
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -139,6 +139,8 @@ class KrbInstance(service.Service):
             pass
 
     def __common_post_setup(self):
+        self.step("configuring Kerberos library to use local KDC",
+                  self.__configure_krb5_conf)
         self.step("starting the KDC", self.__start_instance)
         self.step("configuring KDC to start on boot", self.__enable)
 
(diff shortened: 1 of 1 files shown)


abbra:
LGTM.

simo5:
I do not think this is the correct fix/bug
What we want to do is to change kdc.conf to require certs only after we have installed them.
The KDC is already properly configured and running otherwise but fails to start on replica because certs are not there. We need it to not fail, not to allow certmonger to go oevr the network to other servers

martbab:
But the certs are requested by certmonger on replica which tries to kinit against *the very same KDC that is being configured and is not running yet* because it was told so by the Kerberos config that was updated before starting KDC.

abbra:
@simo5 KDC starts just fine with missing certs. It disables PKINIT if certs aren't reachable. However, if KDC is not running at all, certmonger cannot complete the cert request at all.

simo5:
@martbab @abbra see the pull request in #567

martbab:
Ah right this won't work because on master there would be no library configuration for KDC deployment (realm, etc) that's why server install in travis crashed. Closing this PR as #567 superseds it.

PR #566, opened 2017-03-09
webui: do not warn about CAs if there is only one master

pvoborni:
Web UI showed pop-up dialog which recommends to install additional CA in
topology section when only 1 CA existed even if there was only one master.

Though behind the pop-up is to prevent situation, where multiple replicas
are installed but neither with --setup-ca option and thus risking to loose
CA when original master is lost.

The warning was displayed also if only one IPA server exists. It is unnecessary
to annoy admin only about CA because the entire IPA is not duplicated.

Therefore the pop-up is now shown only one IPA server exists.

https://pagure.io/freeipa/issue/6598

### DIFF
--- a/install/ui/src/freeipa/topology.js
+++ b/install/ui/src/freeipa/topology.js
@@ -497,6 +497,10 @@ topology.servers_search_facet = function(spec, no_init) {
             on_success(data, text_status, xhr);
 
             var result = data.result.results;
+
+            // Do not show warning if there is only one master
+            if (result.length <= 1) return;
+
             var counter = 0;
 
             for (var i=0, l=result.length; i<l; i++) {


pvoborni:
Written in a way that it can be then easily extended with KRA check.

install/ui/src/freeipa/topology.js
```diff
@@ -507,6 +507,9 @@ topology.servers_search_facet = function(spec, no_init) {
                 }
             }
 
+            // Do not show warning if there is only one master
+            if (result.length <= 1) return;
```
stlaz: I'd expect this check to appear prior to the `for` cycle so no useless computations are done, other than that LGTM. I will test it when I'm fixing https://github.com/freeipa/freeipa/pull/553.

pvoborni: fixed

MartinBasti:
master:

* 6027a8111fa9ed7a058fb222f4f96b12039deb8b webui: do not warn about CAs if there is only one master

PR #567, opened 2017-03-09
Configure KDC to use certs after they are deployed

simo5:
Certmonger needs to access the KDC when it tries to obtain certs,
so make sure the KDC can run, then reconfigure it to use pkinit anchors
once certs are deployed.

### DIFF
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -68,6 +68,7 @@ class KrbInstance(service.Service):
         self.kdc_password = None
         self.sub_dict = None
         self.pkcs12_info = None
+        self.config_pkinit = None
 
     suffix = ipautil.dn_attribute_property('_suffix')
     subject_base = ipautil.dn_attribute_property('_subject_base')
@@ -140,12 +141,16 @@ class KrbInstance(service.Service):
 
     def __common_post_setup(self):
         self.step("starting the KDC", self.__start_instance)
+        if self.config_pkinit:
+            self.step("installing X509 Certificate for PKINIT",
+                      self.setup_pkinit)
         self.step("configuring KDC to start on boot", self.__enable)
 
     def create_instance(self, realm_name, host_name, domain_name, admin_password, master_password, setup_pkinit=False, pkcs12_info=None, subject_base=None):
         self.master_password = master_password
         self.pkcs12_info = pkcs12_info
         self.subject_base = subject_base
+        self.config_pkinit = setup_pkinit
 
         self.__common_setup(realm_name, host_name, domain_name, admin_password)
 
@@ -159,10 +164,6 @@ class KrbInstance(service.Service):
         self.step("creating anonymous principal", self.add_anonymous_principal)
 
         self.__common_post_setup()
-
-        if setup_pkinit:
-            self.step("installing X509 Certificate for PKINIT",
-                      self.setup_pkinit)
 
         self.start_creation(runtime=30)
 
@@ -178,14 +179,12 @@ class KrbInstance(service.Service):
         self.pkcs12_info = pkcs12_info
         self.subject_base = subject_base
         self.master_fqdn = master_fqdn
+        self.config_pkinit = setup_pkinit
 
         self.__common_setup(realm_name, host_name, domain_name, admin_password)
 
         self.step("configuring KDC", self.__configure_instance)
         self.step("adding the password extension to the directory", self.__add_pwd_extop_module)
-        if setup_pkinit:
-            self.step("installing X509 Certificate for PKINIT",
-                      self.setup_pkinit)
 
         self.__common_post_setup()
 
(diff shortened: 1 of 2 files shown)


simo5:
Still testing but this should be the way to go to fix the bug reported in #564

ipaserver/install/krbinstance.py
```diff
@@ -377,6 +381,15 @@ def setup_pkinit(self):
         # have any selinux issues with the file context
         shutil.copyfile(paths.IPA_CA_CRT, paths.CACERT_PEM)
 
+        # Now modify configuration to add pkinit anchors and restart KDC
+        self.sub_dict['NOPK'] = ''
+        self.__template_file(paths.KRB5KDC_KDC_CONF, chmod=None, backup=False)
+        try:
+            self.stop()
+            self.start()
```
martbab: Shouldn;t we use `self.restart` here?

ipaserver/install/krbinstance.py
```diff
@@ -377,6 +381,15 @@ def setup_pkinit(self):
...
+        except Exception:
+            root_logger.critical("krb5kdc service failed to restart")
```
martbab: If this is critical error, please re-raise to stop installation altogether. If it is a soft error, please log at error level not critical.

martbab:
I think we can avoid the copy-pasta by actually moving PKINIT requesting code into `__common_post_setup` like this:

```diff
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -142,10 +142,15 @@ class KrbInstance(service.Service):
         self.step("starting the KDC", self.__start_instance)
         self.step("configuring KDC to start on boot", self.__enable)
 
+        if self.setup_pkinit:
+            self.step("installing X509 Certificate for PKINIT",
+                      self.setup_pkinit)
+
     def create_instance(self, realm_name, host_name, domain_name, admin_password, master_password, setup_pkinit=False, pkcs12_info=None, subject_base=None):
         self.master_password = master_password
         self.pkcs12_info = pkcs12_info
         self.subject_base = subject_base
+        self.setup_pkinit = setup_pkinit
 
         self.__common_setup(realm_name, host_name, domain_name, admin_password)
 
@@ -160,10 +165,6 @@ class KrbInstance(service.Service):
 
         self.__common_post_setup()
 
-        if setup_pkinit:
-            self.step("installing X509 Certificate for PKINIT",
-                      self.setup_pkinit)
-
         self.start_creation(runtime=30)
 
         self.kpasswd = KpasswdInstance()
@@ -178,14 +179,12 @@ class KrbInstance(service.Service):
         self.pkcs12_info = pkcs12_info
         self.subject_base = subject_base
         self.master_fqdn = master_fqdn
+        self.setup_pkinit = setup_pkinit
 
         self.__common_setup(realm_name, host_name, domain_name, admin_password)
 
         self.step("configuring KDC", self.__configure_instance)
         self.step("adding the password extension to the directory", self.__add_pwd_extop_module)
-        if setup_pkinit:
-            self.step("installing X509 Certificate for PKINIT",
-                      self.setup_pkinit)
 
         self.__common_post_setup()
 
```
Yes we have now duplicated member assignment but still better that duplicate logic. Also I have some inline comments.

simo5:
Should have addressed all concerns in this push

martbab:
Codewise LGTM, but I get the following error on the replica being deployed regardless of whether I use `--setup-ca` or not. WebUI on replica is not working and throwing around auth erros:

```console
Request ID '20170313152814':
        status: CA_REJECTED
        ca-error: Server at https://replica1.ipa.test/ipa/xml failed request, will retry: -504 (libcurl failed to execute the HTTP POST transaction, explaining:  Failed to connect to replica1.ipa.test port 443: Connection refused).
        stuck: yes
        key pair storage: type=FILE,location='/var/kerberos/krb5kdc/kdc.key'
        certificate: type=FILE,location='/var/kerberos/krb5kdc/kdc.crt'
        CA: IPA
        issuer: 
        subject: 
        expires: unknown
        pre-save command: 
        post-save command: 
        track: yes
        auto-renew: yes
```

If I resubmit the request it goes to 'MONITORING' status. Whn I restart KDC I can use replica's WebUI again. I guess the problem is that certmonger contacts replica's own CA even if it is not configured and apache is not running yet, hence the error. It should contact remote master at all times.

martbab:
@simo5 actually I found multiple issues during review and concluded that setting up PKINIT on DL1 replica never worked correctly actually. Will open respective blocker tickets ASAP.

simo5:
Can you figure out exactly why certmonger is doing this ?

martbab:
@simo5 yes the whole PKINIT setup logic on replica is flawed and will probably need to be moved into a later point in master/replica install. Can I re-use your PR and prepare a new one that will fix it properly? I will keep you the author of this commit if you wish.

simo5:
Sure no prob

martbab:
@simo5 thank you

martbab:
Superseded by https://github.com/freeipa/freeipa/pull/584

PR #569, opened 2017-03-10
Remove copy-schema-to-ca.py from master branch

MartinBasti:
This script is used only for IPA <3.1, so it must be compatible with
ipa-3-0 branch, so it should be placed there

https://pagure.io/freeipa/issue/6540


MartinBasti:
The script doesn't work under ipa-3-0, but it works for RHEL6, so I just put it to contrib/ and marked as RHEL6

MartinBasti:
master:

* f4c7f1dd8a9ce530a8291219a904686ee47e59c7 Remove copy-schema-to-ca.py from master branch
* ca5b53adccdd581bc39233378c422ca448e6edd2 Add copy-schema-to-ca for RHEL6 to contrib/

PR #570, opened 2017-03-10
ipaserver/dcerpc.py: use arcfour_encrypt from samba

abbra:
Samba Python bindings provide samba.arcfour_encrypt(key, data). Instead
of implementing own wrapper, use Samba's.

In future Samba versions this wrapper will be FIPS 140-2 compatible.

Fixes https://pagure.io/freeipa/issue/6697


martbab:
master:

* 7657754e02a5fa62265327937a6c7fd19b381610 ipaserver/dcerpc.py: use arcfour_encrypt from samba

PR #573, opened 2017-03-13
Provide centralized management of user short name resolution

martbab:
This PR implement an initial version of AD user short name resolution
infrastructure consumable by SSSD.[1]

Most of the stuff described in the design page[2] is in-place except of hooks
that would refresh the domain resolution orders after trust domain removal or
disablement. I would like to do them in a separate PR.

Also some edge cases like specifying only separator (':') or an empty domain
('dom1::dom2') have no special treatment, the current code will just complain
about empty DNS labels. Should I improve this behavior?

[1] https://pagure.io/freeipa/issue/6372
[2] https://www.freeipa.org/page/V4/AD_User_Short_Names


ipapython/ldapattrs.py
```diff
@@ -0,0 +1,135 @@
+#
+# Copyright (C) 2016 FreeIPA Contributors see COPYING for license
```
MartinBasti: 2017 ehm ehm

ipapython/ldapattrs.py
```diff
@@ -0,0 +1,135 @@
...
+#
+
+"""
+classes implementing a more object-oriented handling of selected LDAP
+attributes in the framework
+"""
+
+from __future__ import print_function, unicode_literals
```
MartinBasti: `unicode_literals` doesn't sound right to me

MartinBasti:
ACIs? AFAIK SSSD should be able to read this

HonzaCholasta:
I would rather avoid the refactoring in 4.5 - this is fragile code you are touching and I'm afraid it might break in some cases (think different client / server version combinations, thin client vs fat client, etc.).

As for the edge case values, IMO we should allow `:` without complaining as a special case to support "no domains in the list" configuration, and otherwise require known domain names (like in `certmaprule-add`).

martbab:
Updated PR, added ACIs and fixed Py2/Py3 compatibility of doctests.

ipaserver/plugins/idviews.py
```diff
@@ -94,7 +95,7 @@ class idview(LDAPObject):
     container_dn = api.env.container_views
     object_name = _('ID View')
     object_name_plural = _('ID Views')
-    object_class = ['ipaIDView', 'top']
+    object_class = ['ipaIDView', 'ipaNameResolutionData', 'top']
```
MartinBasti: IMO `ipaNameResolutionData` must be in `possible_object_class`,  because idviews after upgrade doesn't have it

abbra:
I don't see ACI.txt regenerated.

ipaserver/plugins/idviews.py
abbra: Correct, it is auxilliary object class, so it should be in the `possible_object_class` and `default_attributes` should contain `ipaNameResolutionOrder` to be able to return the value if it is there.

ipaserver/plugins/config.py
```diff
@@ -250,6 +254,14 @@ class config(LDAPObject):
             label=_('IPA CA renewal master'),
             doc=_('Renewal master for IPA certificate authority'),
             flags={'virtual_attribute', 'no_create'}
+        ),
+        SymbolDelimitedIdentifiers(
+            'ipadomainresolutionorder?',
+            cli_name='domain_resolution_order',
+            label=_('Domain resolution order'),
+            doc=_('colon-separated list of domains used for short name'
```
pvoborni: nitpick: all other doc strings in this plugin start with capital letter

martbab:
@HonzaCholasta I agree, I have removed the commit which introduces special param handling and resorted to simple splitting in validator. I have also regenerated ACIs in the respective commits.

VERSION.m4
```diff
@@ -73,8 +73,8 @@ define(IPA_DATA_VERSION, 20100614120000)
 #                                                      #
 ########################################################
 define(IPA_API_VERSION_MAJOR, 2)
-define(IPA_API_VERSION_MINOR, 220)
-# Last change: Add whoami command
+define(IPA_API_VERSION_MINOR, 222)
```
HonzaCholasta: 220 + 1 != 222

install/share/60basev3.ldif
```diff
@@ -84,5 +85,6 @@ objectClasses: (2.16.840.1.113730.3.8.12.24 NAME 'ipaPublicKeyObject' DESC 'Wrap
 objectClasses: (2.16.840.1.113730.3.8.12.25 NAME 'ipaPrivateKeyObject' DESC 'Wrapped private keys' SUP top AUXILIARY MUST ( ipaPrivateKey $ ipaWrappingKey $ ipaWrappingMech ) X-ORIGIN 'IPA v4.1' )
 objectClasses: (2.16.840.1.113730.3.8.12.26 NAME 'ipaSecretKeyObject' DESC 'Wrapped secret keys' SUP top AUXILIARY MUST ( ipaSecretKey $ ipaWrappingKey $ ipaWrappingMech ) X-ORIGIN 'IPA v4.1' )
 objectClasses: (2.16.840.1.113730.3.8.12.34 NAME 'ipaSecretKeyRefObject' DESC 'Indirect storage for encoded key material' SUP top AUXILIARY MUST ( ipaSecretKeyRef ) X-ORIGIN 'IPA v4.1' )
+objectClasses: (2.16.840.1.113730.3.8.12.39 NAME 'ipaNameResolutionData' DESC 'Data used to resolve short names to fully-qualified form' SUP top AUXILIARY MAY ( ipaDomainResolutionOrder ) X-ORIGIN 'IPA v4.5')
```
HonzaCholasta: `ipaNameResolutionConfig` would make more sense to me (I know, I know).

ipaserver/plugins/config.py
```diff
@@ -266,6 +277,103 @@ def show_servroles_attributes(self, entry_attrs, **options):
...
+            forest_domains = command.trustdomain_find(forest_name)['result']
+
+            trusted_domains.update(
+                {
+                    dom['cn'][0]: dom['domain_enabled'][0]
+                    for dom in forest_domains if 'domain_enabled' in dom
+                }
+            )
+
+        return trusted_domains
```
HonzaCholasta: It would be nice to share code with `check_associateddomain_is_trusted`. Not a blocker.

install/share/60basev3.ldif
martbab: I have already registered this OID and name so we are out of luck I think.

ipaserver/plugins/config.py
martbab: Nice catch, I will try to think about plugging the trustdomain retrieval magic there.

martbab:
Server upgrade consists only from adding the objectclass to ipaConfig which is taken care of in the update file. The idview object schema is modified on-demand when the attribute is set. Is there something else I need to take care of?

HonzaCholasta:
IMO you should add the object class to all existing idviews on upgrade rather than add it on-demand.

MartinBasti:
@HonzaCholasta it will break in case when idview entry is created on older replica, so it is more safe to appending the objectclass dynamically

HonzaCholasta:
Ah, right.

martbab:
PR rebased, I have fixed bugs in ID view objectclass handling and re-used the trusted domain retrieval code in certmap plugin. This is a separate commit so it can be removed if necessary.

I have noticed that with current PR we can not add the domain resolution order to Default Trust View, as it is protected from both modification and removal. @abbra is this expected also in this case?

abbra:
Yes, it is expected too. Remember that 'Default Trust View' is a view that applies globally. You have already global setting to apply.

martbab:
Ok thanks for explanation.

MartinBasti:
master:

* 594c87daf873ceec0c0cf3464bcb1aadb9f2b92a Short name resolution: introduce the required schema
* 1b5f56d15455b6019dd532cb9635fa2c44cb0022 ipaconfig: add the ability to manipulate domain resolution order
* 544d66b7109300e570fb6849f0f9bab8020f3b66 idview: add domain_resolution_order attribute
* 4e5e3eebb223b7f2760e21f22e42775982104b0d Re-use trust domain retrieval code in certmap validators

PR #574, opened 2017-03-13
ipa-replica-prepare fix

stlaz:
A regression was introduced in https://github.com/freeipa/freeipa/commit/0a54fac02cecad3b9e3bf8ad0c8a44df3b701857. Fix + don't fail if either file was not created during server-cert creation.

### DIFF
--- a/ipaserver/install/certs.py
+++ b/ipaserver/install/certs.py
@@ -368,8 +368,11 @@ class CertDB(object):
             with open(self.certder_fname, "r") as f:
                 dercert = f.read()
         finally:
-            os.unlink(self.certreq_fname)
-            os.unlink(self.certder_fname)
+            for fname in (self.certreq_fname, self.certder_fname):
+                try:
+                    os.unlink(fname)
+                except OSError:
+                    pass
 
         return dercert
 
--- a/ipaserver/install/ipa_replica_prepare.py
+++ b/ipaserver/install/ipa_replica_prepare.py
@@ -603,7 +603,8 @@ class ReplicaPrepare(admintool.AdminTool):
 
         try:
             db = certs.CertDB(
-                api.env.realm, nssdir=self.dir, subject_base=subject_base)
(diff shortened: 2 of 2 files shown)


ipaserver/install/certs.py
```diff
@@ -368,8 +368,11 @@ def create_server_cert(self, nickname, hostname, subject=None):
             with open(self.certder_fname, "r") as f:
                 dercert = f.read()
         finally:
-            os.unlink(self.certreq_fname)
-            os.unlink(self.certder_fname)
+            for fname in (self.certreq_fname, self.certder_fname):
+                try:
+                    os.unlink(fname)
+                except Exception:
```
MartinBasti: Shame! Too broad exception, use OSError instead

MartinBasti:
Can be this caused by your patch?
```
error exporting Server certificate: Command '/usr/bin/openssl pkcs12 -export -name KDC-Cert -in /tmp/tmpmS5rCkipa/realm_info/kdc.pem -out /tmp/tmpmS5rCkipa/realm_info/pkinitcert.p12 -passout file:/tmp/tmpmS5rCkipa/realm_info/pkinit_pin.txt' returned non-zero exit status 1
```

stlaz:
Very unlikely but I'll investigate.

stlaz:
My wild guess is that it might be caused by ba3c201a but not by this patchset as it does not touch it.

stlaz:
Actually, this is most probably a privilege-separation issue since "kdc.pem" which we try to read here does not exist ever since.

stlaz:
@MartinBasti should be fixed in https://github.com/freeipa/freeipa/pull/580

MartinBasti:
master:

* 992e6ecd1ff33f4f872e8f174bd426507c55f5c4 Fix ipa-replica-prepare server-cert creation
* 8980f4098ebf6b62556e24f090718802d1e495d3 Don't fail more if cert req/cert creation failed

PR #575, opened 2017-03-13
IPA certauth plugin

sumit-bose:
This patch add a certauth plugin which allows the IPA server to support
PKINIT for certificates which do not include a special SAN extension which
contains a Kerberos principal but allow other mappings with the help of
SSSD's certmap library.

Related to https://pagure.io/freeipa/issue/4905

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,7 @@ freeipa2-dev-doc
 /daemons/dnssec/ipa-ods-exporter.socket
 /daemons/ipa-kdb/ipa_kdb_tests
 /daemons/ipa-kdb/tests/.dirstamp
+/daemons/ipa-kdb/ipa-certauth
 /daemons/ipa-otpd/ipa-otpd
 /daemons/ipa-otpd/ipa-otpd.socket
 /daemons/ipa-otpd/ipa-otpd@.service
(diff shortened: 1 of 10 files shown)


sumit-bose:
I updated the code to reflect the latest changes in the interface from https://github.com/krb5/krb5/pull/610.

abbra:
The code LGTM. Once updated SSSD is added to freeipa-master copr, let's see what CI says.

Authentication indicators' handling would need to be added in a separate PR once certmap rules would provide the indicator value.

dkupka:
I've tested the patches and it worked as expected. Once CI successfully finishes I'll ACK it.

dkupka:
master:

* da880decfedc66f9d0d2734dcb86c23a8866f603 ipa-kdb: add ipadb_fetch_principals_with_extra_filter()
* c4156041feb9c48598427ad59e43313b9c7327bb IPA certauth plugin
ipa-4-5:

* cfaaf4e821338dbc146dd49d3c22978165d2e329 ipa-kdb: add ipadb_fetch_principals_with_extra_filter()
* 5a1ce1fbaa6c7a85bd1bee2a70b8b22509ede7c7 IPA certauth plugin

PR #578, opened 2017-03-13
Coverity: fix bad use of null-like value in cert.py

tomaskrizek:


apophys:
Please do not post links to internal Red Hat resources in public.

stlaz:
Shame on you, @tomaskrizek

MartinBasti:
I would rather focus on why `principal_obj` is not defined there

stlaz:
if you look at steps 2, 4 and 13 in the Synopsis report (sorry, community!), you will see that this is a false positive. Anyone, please close this as "rejected" to confirm.

tomaskrizek:
Sorry about the link, I've removed it.

Nevertheless, it seems that's indeed a false positive, because `principal_type` is set to `USER`, but `principal_type == KRBTGT` on [L616](https://github.com/freeipa/freeipa/pull/578/files#diff-95cc6f5739d8923e9d470c2f686038f1R616) is evaluated as true instead of `principal_type == USER` at [L624](https://github.com/freeipa/freeipa/pull/578/files#diff-95cc6f5739d8923e9d470c2f686038f1R624) which would set `principal_obj`. There is no other assignment to `principal_type` in between.

Closing the PR, coverity error is a false positive.

MartinBasti:
But this is about `principal_obj`, I don't see it in step 2, but I see `principal_obj = None` in step 11

MartinBasti:
Ah I see it is really false positive

PR #579, opened 2017-03-14
csrgen: hide cert-get-requestdata in CLI

HonzaCholasta:
The CSR generation feature is supposed to be used from cert-request, hide
the internal cert-get-requestdata command in the CLI.

https://fedorahosted.org/freeipa/ticket/4899


HonzaCholasta:
master:

* 72de679eb445c975ec70cd265d37d4927823ce5b csrgen: hide cert-get-requestdata in CLI

PR #580, opened 2017-03-14
Fix KDC certificates export on DL0

stlaz:
I don't know since when this has been broken but my guess is - for a long time.


stlaz:
We should not care about KDC certificates at all on DL0.

PR #581, opened 2017-03-14
Backup KDC certificate pair

stlaz:
KDC certificate pair was added but is not included in backup which
might cause issues when restoring the IPA service.

https://pagure.io/freeipa/issue/6748

This probably does not fix the issue as a whole but I am not sure
if there's more that we can do on the IPA side.


MartinBasti:
master:

* ee6d031a6a0939c1f51a874b1f8f9b19ec727203 Backup KDC certificate pair

PR #582, opened 2017-03-14
Remove pkinit from ipa-replica-prepare

stlaz:
The PKINIT feature is not available on domain level 0 so any
options about pkinit are false.

https://pagure.io/freeipa/issue/6759

### DIFF
--- a/ipaserver/install/ipa_replica_prepare.py
+++ b/ipaserver/install/ipa_replica_prepare.py
@@ -88,9 +88,6 @@ class ReplicaPrepare(admintool.AdminTool):
         parser.add_option("--allow-zone-overlap", dest="allow_zone_overlap",
             action="store_true", default=False, help="create DNS "
             "zone even if it already exists")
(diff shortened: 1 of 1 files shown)


MartinBasti:
Works for me. @abbra @simo5 do you have any objections?

abbra:
LGTM.

HonzaCholasta:
The options were available since forever, so I guess you should just hide them instead of removing them.

The same options are still available in domain level 0 `ipa-server-install` - is this intentional?

abbra:
They were in DL0 in `ipa-server-install` for very long time and never worked. We left them there to make sure we can get them back to work sometime later. We did but in new design `ipa-replica-prepare` does not need to use these options, unlike `ipa-server-install`.

MartinBasti:
master:

* 46d4d534c08d14756b989e157e87a078d174ad5c Remove pkinit from ipa-replica-prepare

PR #584, opened 2017-03-14
Improve the implementation of PKINIT certificate retrieval

martbab:
The original PKINIT cert request code contained numerous defects, namely:

   * nearly absent handling of rejected requests and CA errors which resulted
     e.g. in an unusable WebUI after replica installation
     and
   * certificate request logic that was not consistent with the rest of the
     installers (DS, HTTP for example): what caused hard errors in their case
     went unnoticed in PKINIT setup

This PR consolidates this code so that errors arising from CA rejecting the
PKINIT cert request cause the installers to abort immediately. The PKINIT step
was also split into a separate method executed before LDAP updates. The name
was chosen to be `enable_ssl` in order to make the planned refactoring of
certificate requesting code (https://pagure.io/freeipa/issue/6429) easier: the
method name is not accurate but at least it is consistent with e.g. LDAP
installer co the common code can be grepper with greater ease.

https://pagure.io/freeipa/issue/6739

### DIFF
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -69,6 +69,7 @@ class KrbInstance(service.Service):
         self.sub_dict = None
         self.pkcs12_info = None
         self.master_fqdn = None
+        self.config_pkinit = None
 
     suffix = ipautil.dn_attribute_property('_suffix')
     subject_base = ipautil.dn_attribute_property('_subject_base')
@@ -147,6 +148,7 @@ class KrbInstance(service.Service):
         self.master_password = master_password
         self.pkcs12_info = pkcs12_info
         self.subject_base = subject_base
+        self.config_pkinit = setup_pkinit
 
         self.__common_setup(realm_name, host_name, domain_name, admin_password)
 
@@ -160,10 +162,6 @@ class KrbInstance(service.Service):
         self.step("creating anonymous principal", self.add_anonymous_principal)
 
         self.__common_post_setup()
-
-        if setup_pkinit:
-            self.step("installing X509 Certificate for PKINIT",
-                      self.setup_pkinit)
 
         self.start_creation()
 
@@ -179,14 +177,12 @@ class KrbInstance(service.Service):
         self.pkcs12_info = pkcs12_info
         self.subject_base = subject_base
         self.master_fqdn = master_fqdn
+        self.config_pkinit = setup_pkinit
 
         self.__common_setup(realm_name, host_name, domain_name, admin_password)
 
         self.step("configuring KDC", self.__configure_instance)
         self.step("adding the password extension to the directory", self.__add_pwd_extop_module)
-        if setup_pkinit:
-            self.step("installing X509 Certificate for PKINIT",
-                      self.setup_pkinit)
 
         self.__common_post_setup()
 
@@ -392,6 +388,20 @@ class KrbInstance(service.Service):
         # Finally copy the cacert in the krb directory so we don't
         # have any selinux issues with the file context
         shutil.copyfile(paths.IPA_CA_CRT, paths.CACERT_PEM)
+
+        try:
+            self.restart()
+        except Exception:
+            root_logger.critical("krb5kdc service failed to restart")
+            raise
+
+    def enable_ssl(self):
+        if self.config_pkinit:
+            self.steps = []
+            self.step("installing X509 Certificate for PKINIT",
+                      self.setup_pkinit)
+
+            self.start_creation()
 
     def get_anonymous_principal_name(self):
         return "%s@%s" % (ANON_USER, self.realm)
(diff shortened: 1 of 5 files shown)


install/share/kdc.conf.template
```diff
@@ -12,6 +12,6 @@
   dict_file = $DICT_WORDS
   default_principal_flags = +preauth
 ;  admin_keytab = $KRB5KDC_KADM5_KEYTAB
-  pkinit_identity = FILE:$KDC_CERT,$KDC_KEY
-  pkinit_anchors = FILE:$CACERT_PEM
+$NOPK  pkinit_identity = FILE:$KDC_CERT,$KDC_KEY
```
abbra: This is not really needed. krb5kdc will complain that pkinit_identity is not readable and PKINIT cannot be configured but the rest will continue working. We have `pkinit_identity` and `pkinit_anchors` in `install/share/kdc.conf.template` since 2010, they contained wrong values and missing certs, and KDC did work fine.

install/share/kdc.conf.template
```diff
@@ -12,6 +12,6 @@
...
+$NOPK  pkinit_anchors = FILE:$CACERT_PEM
```
abbra: Same here. Do not comment out this configuration option.

ipaserver/install/krbinstance.py
```diff
@@ -221,6 +220,7 @@ def __setup_sub_dict(self):
                              KRB5KDC_KADM5_ACL=paths.KRB5KDC_KADM5_ACL,
                              DICT_WORDS=paths.DICT_WORDS,
                              KRB5KDC_KADM5_KEYTAB=paths.KRB5KDC_KADM5_KEYTAB,
+                             NOPK=';',
```
abbra: no need for the comment out.

ipaserver/install/krbinstance.py
```diff
@@ -393,6 +394,15 @@ def setup_pkinit(self):
         # have any selinux issues with the file context
         shutil.copyfile(paths.IPA_CA_CRT, paths.CACERT_PEM)
 
+        # Now modify configuration to add pkinit anchors and restart KDC
+        self.sub_dict['NOPK'] = ''
```
abbra: no need to update kdc.conf here. Just do the restart.

ipaserver/install/krbinstance.py
```diff
@@ -393,6 +394,15 @@ def setup_pkinit(self):
...
+        self.__template_file(paths.KRB5KDC_KDC_CONF, chmod=None, backup=False)
+        try:
+            self.restart()
+        except Exception:
+            root_logger.critical("krb5kdc service failed to restart")
+            raise
+
```
abbra: I also think we should actually check that anonymous PKINIT works with new configuration -- since krb5kdc will start just fine with non-working PKINIT, you really need to try to 'kinit -n' to see if anonymous pkinit works.

MartinBasti:
Since I applied this PR, I cannot pass through failing client side installation:
```
Forwarding 'ping' to json server 'https://vm-024.abc.idm.lab.eng.brq.redhat.com/ipa/json'
Cannot connect to the server due to Kerberos error: No valid Negotiate header in server response. Trying with delegate=True
trying https://vm-024.abc.idm.lab.eng.brq.redhat.com/ipa/json
Forwarding 'ping' to json server 'https://vm-024.abc.idm.lab.eng.brq.redhat.com/ipa/json'
Second connect with delegate=True also failed: No valid Negotiate header in server response
Installation failed. As this is IPA server, changes will not be rolled back.
Cannot connect to the IPA server RPC interface: No valid Negotiate header in server response
The ipa-client-install command failed. See /var/log/ipaclient-install.log for more information
ipa.ipapython.install.cli.install_tool(CompatServerMasterInstall): ERROR    Configuration of client side components failed!
ipa.ipapython.install.cli.install_tool(CompatServerMasterInstall): ERROR    The ipa-server-install command failed. See /var/log/ipaserver-install.log for more information
```

martbab:
@simo5 are you OK with @abbra's inline suggestions (it is your commit after all :))?
@MartinBasti hmmm I will try to reproduce the issue with fresh VMs.

ipaserver/install/krbinstance.py
martbab: That sound like a really good idea that can prevent regressions like the one reported in the ticket to happen. I will add the anon pkinit check and fail with hard error if it does not work.

martbab:
I have reworked the PR quite a bit and added/changed a few checks due to replication race conditions affecting PKINIT requests from replica to master.

MartinBasti:
Works for me and code looks OK

MartinBasti:
master:

* 95768de06fbef78169329af12b29e4d65e4bf157 Make PKINIT certificate request logic consistent with other installers
* b5b23e073e59930e4dcf14ea8031c2c0441e6344 Request PKINIT cert directly from Dogtag API on first master
* bd18b5f91e3f98fa877def245c54c1cd33bd372e Move PKINIT configuration to a later stage of server/replica install
* 069948466e81d99a0dd48ffffa32af50351d0189 Make wait_for_entry raise exceptions
* 8f4abf7bc1607fc44f528b8a443b69cb82269e69 check that the master requesting PKINIT cert has KDC enabled
* b45629fc480e61464b402ac2fc52c6f9fc61df0e check for replica's KDC entry on master before requesting PKINIT cert
* a1686a90c0cc8c16c89ef1bada7f507729bf3252 Try out anonymous PKINIT after it is configured

PR #585, opened 2017-03-14
Remove allow_constrained_delegation from gssproxy.conf

pvomacka:
This change reverts option which breaks priviledge separation.


simo5:
Please change commit message to:

The Apache process *must* not allowed to use constrained delegation to contact services because it is already allowed to impersonate users to itself. Allowing it to perform constrained delegation would let it impersonate any user against the LDAP service without authentication.

MartinBasti:
master:

* f4cd61f3011877fc9cc2a809438059b07362b0aa Remove allow_constrained_delegation from gssproxy.conf

PR #587, opened 2017-03-14
Python 3: Fix session storage

tiran:
ctypes can only handle bytes, not text. Encode and decode all incoming
and outgoing text from UTF-8 to bytes.

### DIFF
--- a/ipapython/session_storage.py
+++ b/ipapython/session_storage.py
@@ -104,6 +104,13 @@ def store_data(princ_name, key, value):
     """
     Stores the session cookie in a hidden ccache entry.
     """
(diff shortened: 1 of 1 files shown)


tiran:
@simo5 is UTF-8 correct or are keys and values ASCII only?

simo5:
Technically principal names could use any encoding ... but we make the assumption they are utf-8 in freeIPA, so this should be ok.

MartinBasti:
master:

* 42bc778c0c1de91f0d8dc695dfee4e5aea4cc1f0 Python 3: Fix session storage

PR #588, opened 2017-03-14
CONFIGURE: Properly detect libpopt on el7

HonzaCholasta:
libpopt added pkg-config file in 1.16 but there are still distributions
which has older version of library (el6, el7). And new features from
libpopt are not used anywhere. Configure should try to detect as much as
possible and users should not use workarounds with explicitely enabled
variables as parameters e.g.
   ./configure POPT_LIBS="-lpopt "

This change originating from PR #494 is required to make building upstream IPA possible on RHEL/CentOS 7.


HonzaCholasta:
master:

* 4fe9166ac9f9a100d69ce37f19ae1ae971bb2ce1 CONFIGURE: Properly detect libpopt on el7

PR #590, opened 2017-03-15
Validate user input for cert-get-requestdata

Akasurde:
Fix adds validatation for Principal and CSR generation
tool values

Fixes https://pagure.io/freeipa/issue/6742

### DIFF
--- a/ipaclient/plugins/csrgen.py
+++ b/ipaclient/plugins/csrgen.py
@@ -77,6 +77,10 @@ class cert_get_requestdata(Local):
             util.check_writable_file(options['out'])
 
         principal = options.get('principal')
+        if principal is None:
+            raise errors.InvocationError(
+                message=_('Principal is required')
+            )
         profile_id = options.get('profile_id')
         if profile_id is None:
             profile_id = dogtag.DEFAULT_PROFILE


ipaclient/plugins/csrgen.py
```diff
@@ -75,10 +75,19 @@ def execute(self, *args, **options):
             util.check_writable_file(options['out'])
 
         principal = options.get('principal')
+        if principal is None:
+            raise errors.InvocationError(message=u"Principal is required")
```
tiran: Please include ```_()``` for i18n.

ipaclient/plugins/csrgen.py
```diff
@@ -75,10 +75,19 @@ def execute(self, *args, **options):
...
         profile_id = options.get('profile_id')
         if profile_id is None:
             profile_id = dogtag.DEFAULT_PROFILE
         helper = options.get('helper')
+        if helper is None:
+            raise errors.InvocationError(message=u"CSR generation tool is"
```
tiran: You are missing a trailing space in this line. You can make the code more readable by indenting the class' arguments:

```
            raise errors.InvocationError(
                message=_(u"CSR generation tool is required")
            )
```

ipaclient/plugins/csrgen.py
```diff
@@ -75,10 +75,19 @@ def execute(self, *args, **options):
...
+                                                 "required")
+        if helper.lower() not in ['openssl', 'certutil']:
+            raise errors.InvocationError(message=u"Allowed values for CSR "
```
tiran: same here

ipaclient/plugins/csrgen.py
Akasurde: OK

ipaclient/plugins/csrgen.py
Akasurde: OK.

ipaclient/plugins/csrgen.py
Akasurde: OK.

rcritten:
You are duplicating the list of helpers. It would have been better to have helper defined as a StrEnum. If it isn't too late to change (e.g. no release has shipped with that in the API) then perhaps a separate patch, then you wouldn't need this enforcement at all.

Akasurde:
@rcritten I don't know about backward compatibility of changing helper to StrEnum. @MartinBasti  @HonzaCholasta Can you please comment on this?

MartinBasti:
I have no context about how exactly certrequest is supposed to work, but IMO it was done in that way to allow dynamically adding more helpers as plugins, that's why it is Str and not SrEnum, but code doesn't look it may support that.

@LiptonB do you remember why Str param was used?

@Akasurde Right now there is no backward compatibility because 4.5 will be first release that contains this feature.

LiptonB:
I don't think one could really add a new helper without modifying the code, so there's probably no need to allow arbitrary strings. Given that, StrEnum seems appropriate.

For the record, https://github.com/freeipa/freeipa/pull/542 is going to be modified to remove the `helper` parameter of `cert-get-requestdata` entirely, though I haven't had a chance to make the change yet.

MartinBasti:
> For the record, #542 removes the helper parameter of cert-get-requestdata, and will be modified to remove the concept of different helpers entirely, though I haven't had a chance to make that change yet.

today is 4.5 release so you have to keep some level of backward compatibility in that PR

Akasurde:
@MartinBasti Should I wait for #542 to get merged?

Akasurde:
Bump for review.

tomaskrizek:
Closing for inactivity. If this PR is still relevant, please re-open it.

PR #591, opened 2017-03-15
spec file: add unconditional python-setuptools BuildRequires

HonzaCholasta:
python-setuptools is required not only for lint, but to make the build
possible at all.

Move the python-setuptools BuildRequires from the lint section to the main
section.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -79,8 +79,10 @@ BuildRequires:  libtool
 BuildRequires:  gettext
 BuildRequires:  gettext-devel
 BuildRequires:  python-devel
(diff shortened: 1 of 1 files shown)


HonzaCholasta:
master:

* 7ef4e9eb810063243fcc575434d856c854b14eee spec file: add unconditional python-setuptools BuildRequires

lslebodn:
Post push comment. This is a reason why there was configure time check in **rejected** PR#494.
Spec file change is not very upstream friendly.
https://github.com/freeipa/freeipa/pull/494/commits/a1321abbdb2cf510e0f36c65d2af0fb6329d2e23

PR #592, opened 2017-03-15
slapi plugins: fix CFLAGS

HonzaCholasta:
Add explicit NSPR_CFLAGS and NSS_CFLAGS where NSPR_LIBS and NSS_LIBS is
used.

Use DIRSRV_CFLAGS rather than hardcode -I/usr/include/dirsrv.

Append NSPR_CFLAGS to DIRSRV_CFLAGS in ./configure as slapi-plugin.h
includes nspr.h.


HonzaCholasta:
master:

* b7329e31f5c985b9721e3a21b1cd1bec6430129d slapi plugins: fix CFLAGS

PR #593, opened 2017-03-15
Add make devcheck for developers 

tiran:
Ticket 6604 makes pylint and jsl optional dependencies. The change
is controversal, because some developers prefer that pylint and jsl
should be required unless explicitly disabled.

`make patchcheck` is my answer to address the concerns. It's a superior
solution to `make lint` as pre-commit check. It combines several
additional checks under a single, easy rememberable and convenient make
target:

* build all
* acilint, apiclient, jslint, polint
* make check
* pylint under Python 2 and 3
* subset of unit test suite

https://fedorahosted.org/freeipa/ticket/6604

Depends on 

- [X] #475
- [X] #587 
- [X] #594
- [x] #636
- [x] #670

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -152,6 +152,35 @@ JSLINT_TARGET = jslint
 endif WITH_JSLINT
 lint: acilint apilint $(POLINT_TARGET) $(PYLINT_TARGET) $(JSLINT_TARGET)
 
+.PHONY: devcheck
+devcheck: all
+if ! WITH_POLINT
+	@echo "ERROR: polint not available"; exit 1
+endif
+if ! WITH_PYLINT
+	@echo "ERROR: pylint not available"; exit 1
+endif
+if ! WITH_JSLINT
+	@echo "ERROR: jslint not available"; exit 1
+endif
+if ! WITH_PYTHON2
+	@echo "ERROR: python2 not available"; exit 1
+endif
+	@ # run all linters, tests, and check with Python 2
+	PYTHONPATH=$(top_srcdir) $(PYTHON2) ipatests/ipa-run-tests \
+	    --ipaclient-unittests
+	$(MAKE) $(AM_MAKEFLAGS) acilint apilint polint jslint check
+	$(MAKE) $(AM_MAKEFLAGS) PYTHON=$(PYTHON2) pylint
+if WITH_PYTHON3
+	@ # just tests and pylint on Python 3
+	PYTHONPATH=$(top_srcdir) $(PYTHON3) ipatests/ipa-run-tests \
+	    --ipaclient-unittests
+	$(MAKE) $(AM_MAKEFLAGS) PYTHON=$(PYTHON3) pylint
+else
+	@echo "WARNING: python3 not available"
+endif
+	@echo "All tests passed."
+
 .PHONY: $(top_builddir)/ipapython/version.py
 $(top_builddir)/ipapython/version.py:
 	(cd $(top_builddir)/ipapython && make version.py)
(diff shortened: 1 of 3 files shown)


tiran:
Depends on PRs #475, #587, #594

tiran:
All dependencies have been merged. PR is ready for review.

ipatests/util.py
```diff
@@ -328,20 +328,20 @@ def assert_deepequal(expected, got, doc='', stack=tuple()):
     If the tests fails, it will raise an ``AssertionError`` with detailed
     information, including the path to the offending value.  For example:
 
-    >>> expected = [u'Hello', dict(world=u'how are you?')]
-    >>> got = [u'Hello', dict(world='how are you?')]
+    >>> expected = [u'Hello', dict(world=1)]
+    >>> got = [u'Hello', dict(world=1.0)]
```
tiran: The old test scenario does no longer work on Python 3 since ```u'how are you'``` and ```'how are you'``` have identical type. ```u'how are you' != b'how are you'```. It works with int / float on all Python versions.

ipatests/util.py
```diff
@@ -328,20 +328,20 @@ def assert_deepequal(expected, got, doc='', stack=tuple()):
...
     Traceback (most recent call last):
       ...
     AssertionError: assert_deepequal: type(expected) is not type(got).
       Testing my nested data
-      type(expected) = <type 'unicode'>
-      type(got) = <type 'str'>
-      expected = u'how are you?'
-      got = 'how are you?'
-      path = (0, 'world')
+      type(expected) = <... 'int'>
```
tiran: Python 2 has ```<type 'int'>``` while Python 3 uses ```<class 'int'>```.

Makefile.am
```diff
@@ -152,6 +152,35 @@ JSLINT_TARGET = jslint
 endif WITH_JSLINT
 lint: acilint apilint $(POLINT_TARGET) $(PYLINT_TARGET) $(JSLINT_TARGET)
 
+.PHONY: patchcheck
+patchcheck: all
```
stlaz: This is a [major nitpick](https://i.imgflip.com/1mam2x.jpg) but could it rather be called `develcheck`? The two `ch` in close succession don't quite roll of my keyboard, if that's not an intention.

tiran: Python uses patchcheck, so I'm used to it. But I don't mind to switch to devcheck, develcheck, commitcheck or similar.

By the way, make has tab completion :)

> Use the tab, Luke!

stlaz: `devcheck` seems fine to me :)

stlaz:
The changes to Makefile and configure.ac are just fine. I understand that changes in the `ipapython/session_storage.py` are done elsewhere so once that is pushed, we'll need a rebase.
I don't see the explanation why we're disabling the test in `ipatests/test_ipapython/test_session_storage.py `, that might need a different commit?

tiran:
```test_session_storage``` is not a unit test or functional test. It is an integration test that depends on a valid Kerberos configuration and session. Do you prefer a separate PR?

stlaz:
Whichever is ok with you, I don't mind if it's in the same PR if it is related to the same ticket.

tiran:
I split the changes to session storage tests into a separate commit. The other commit is in #670

stlaz:
Thanks, ACK.

MartinBasti:
@stlaz why is this ACKed when it depends on #670 ?

stlaz:
@MartinBasti #670 was ACKed already and the commit was originally a part of this.

MartinBasti:
Ah right the description hasn't been updated

MartinBasti:
Needs rebase

MartinBasti:
master:

* e357133fd7b276ccabfe1896ee948f2bb3541d94 Add make devcheck for developers
* 6c092c24b2bfbba0a3f263d88f7a0dbf83f24869 Skip test_session_storage in ipaclient unittest mode


ipa-4-5:

* 89ab24f1fbb58feb603d60503c685ebad41a4237 Add make devcheck for developers
* c80adf6e0d16f807f90479660af22540cd92d774 Skip test_session_storage in ipaclient unittest mode

PR #595, opened 2017-03-15
idviews: correctly handle modification of non-existent view

martbab:
the pre-callback in `idview-mod` did not correctly handle non-existent
object during objectclass check. It will now correctly report that the
object was not found instead on generic 'no such entry'.

https://pagure.io/freeipa/issue/6372


MartinBasti:
master:

* 1cdd5dee006426c996f67240b6cb2c1aa05e5168 idviews: correctly handle modification of non-existent view

PR #596, opened 2017-03-15
spec file: support client-only build

HonzaCholasta:
nspr-devel, nss-devel and openssl-devel are required for client-only build,
move their respective BuildRequires from the server-specific BuildRequires
section to the main BuildRequires section.

Pass --enable-server or --disable-server to ./configure based on the value
of %{ONLY_CLIENT}.

Remove the `make client-check` call from %check, as the client-check target
does not exist anymore. Always call `make check` instead.

Do not package the /usr/share/ipa directory in freeipa-client-common, as it
is not created in client-only build.

https://pagure.io/freeipa/issue/6517

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1,6 +1,11 @@
 # Define ONLY_CLIENT to only make the ipa-client and ipa-python
 # subpackages
 %{!?ONLY_CLIENT:%global ONLY_CLIENT 0}
+%if %{ONLY_CLIENT}
+    %global enable_server_option --disable-server
+%else
+    %global enable_server_option --enable-server
+%endif
 
 # Build with ipatests
 %if ! %{ONLY_CLIENT}
@@ -97,6 +102,9 @@ BuildRequires:  python3-setuptools
 BuildRequires:  systemd
 # systemd-tmpfiles which is executed from make install requires apache user
 BuildRequires:  httpd
+BuildRequires:  nspr-devel
+BuildRequires:  nss-devel
+BuildRequires:  openssl-devel
 BuildRequires:  libini_config-devel
 BuildRequires:  cyrus-sasl-devel
 %if ! %{ONLY_CLIENT}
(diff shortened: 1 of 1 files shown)


lslebodn:
You can also move few dependencies to server only build
```
BuildRequires:  libini_config-devel
BuildRequires:  cyrus-sasl-devel
```

Check spec file changes in **rejected** PR https://github.com/freeipa/freeipa/pull/494/files

otherwise LGTM

HonzaCholasta:
I can't, both are required to build client components (`ipa-getkeytab` specifically), moving them to the server section would break client-only RPM build.

tiran:
```ipa-getkeytab``` uses ```libini_config``` and ```libsasl2```.

**Comment by pvomacka** [comment]
Client only build does not work on Fedora. So NACK.

**Comment by lslebodn** [comment]
On (15/03/17 02:32), Jan Cholasta wrote:
>I can't, both are required to build client components (`ipa-getkeytab` specifically), moving them to the server section would break client-only RPM build.
>
Sorry for confusion. I misinterpreted changes in diff.

The real change was moving following part to server only build
```
 # %{_unitdir}, %{_tmpfilesdir}
 BuildRequires:  systemd
 # systemd-tmpfiles which is executed from make install requires apache user
 BuildRequires:  httpd
```

But git diff decided to show moving of other lines

LS

HonzaCholasta:
This PR now depends on PR #597.

pvomacka:
@HonzaCholasta Thank you for update - only client build on Fedora now works. So ACK once the travis pass.

pvomacka:
ipa server build without tests does not work, so NACK

HonzaCholasta:
Now that PR #502 was pushed this should finally work.

HonzaCholasta:
master:

* 417f1926c48b426b34b18edb28869f4f06824873 spec file: support client-only build

PR #600, opened 2017-03-15
CONFIGURE: Improve detection of xmlrpc_c flags

lslebodn:
The pkg-config files for xmlrpc_c libraries are shipped just
in fedora/rhel due to downstream patch. Debian does not have
pkg-config files for xmlrpc_c. Therefore we need to fallback to older
method of detection XMLRPC_*FLAGS which was reverted
by the commit 1e0143c159134337a00a91d4ae64e614f72da62e

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -171,7 +171,20 @@ PKG_CHECK_MODULES([SASL], [libsasl2])
 dnl ---------------------------------------------------------------------------
 dnl - Check for XMLRPC-C
 dnl ---------------------------------------------------------------------------
-PKG_CHECK_MODULES([XMLRPC], [xmlrpc xmlrpc_client xmlrpc_util])
+PKG_CHECK_MODULES([XMLRPC], [xmlrpc xmlrpc_client xmlrpc_util], [],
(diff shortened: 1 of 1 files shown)


lslebodn:
@tjaalton It should simplify you work on debian

abbra:
LGTM. Falling back to a standard check is fine.

MartinBasti:
Could you add ticket to commit message please https://fedorahosted.org/freeipa/ticket/6418

tjaalton:
yes, I just added a comment on the old PR, I need this on Debian, so +1 from me

MartinBasti:
Commit message amended before pushed

MartinBasti:
master:

* 2a4f7f2cfaf6ac5ffaf4cc2b43fa0e9b5fa3ebe4 CONFIGURE: Improve detection of xmlrpc_c flags

lslebodn:
On (15/03/17 08:14), MartinBasti wrote:
>Commit message amended before pushed
>
Thank you; I was busy with other tasks.

LS

PR #601, opened 2017-03-15
spec file: always provide python package aliases

HonzaCholasta:
Provide python-ipa* aliases for python2-ipa* subpackages when the
python_provide RPM macro is not available.


HonzaCholasta:
master:

* 990ce9eef314622440b2036742bbf34f57ba2699 spec file: always provide python package aliases

PR #608, opened 2017-03-16
tasks: run `systemctl daemon-reload` after httpd.service.d updates

HonzaCholasta:
Run `systemctl daemon-reload` after
`/etc/systemd/system/httpd.service.d/ipa.conf` is created or deleted,
otherwise systemd will not merge the file into httpd.service and therefore
required environment variables will not be set for httpd.

This fixes authentication failures ("No valid Negotiate header in server
response") due to missing `GSS_USE_PROXY=yes` in httpd environment.

https://pagure.io/freeipa/issue/6773

### DIFF
--- a/ipaplatform/redhat/tasks.py
+++ b/ipaplatform/redhat/tasks.py
@@ -483,6 +483,9 @@ class RedHatTaskNamespace(BaseTaskNamespace):
         os.chmod(paths.SYSTEMD_SYSTEM_HTTPD_IPA_CONF, 0o644)
         self.restore_context(paths.SYSTEMD_SYSTEM_HTTPD_IPA_CONF)
 
+        ipautil.run([paths.SYSTEMCTL, "--system", "daemon-reload"],
+                    raiseonerr=False)
+
     def configure_http_gssproxy_conf(self):
         ipautil.copy_template_file(
             os.path.join(paths.USR_SHARE_IPA_DIR, 'gssproxy.conf.template'),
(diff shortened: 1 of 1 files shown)


martbab:
Hmmm I just caught a following error during FreeIPA replica uninstall:

```console[root@replica1 ~]# ipa-server-install --uninstall -U
Updating DNS system records
--------------------------------------
Deleted IPA server "replica1.ipa.test"
--------------------------------------
Shutting down all IPA services
Unconfiguring ntpd
Configuring certmonger to stop tracking system certificates for KRA
Configuring certmonger to stop tracking system certificates for CA
Unconfiguring CA
Unconfiguring web server
ipa         : ERROR    Command '/bin/systemctl restart httpd.service' returned non-zero exit status 1
```

see the excerpt of the uninstall log here: https://paste.fedoraproject.org/paste/TcHWFTK-TwNhO0v6~BBMG15M1UNdIGYhyRLivL9gydE=/

It looks like we need another daemon-reload in the Apache uninstallation. Although I don't see any reason to add it, it looks like some restore operations make systemd unhappy.

HonzaCholasta:
@martbab, can I see httpd error_log? Uninstall works fine for me.

HonzaCholasta:
@martbab, and httpd journal as well.

martbab:
@HonzaCholasta I was not able to reproduce it any more so I guess that it was transient error. If I encounter it again I will file a separate ticket.

martbab:
master:

* 3de09709cc33f1d26f2d605bac82110fe73dde03 tasks: run `systemctl daemon-reload` after httpd.service.d updates
ipa-4-5:

* 62c41219acdd0e82201168aea5cb22879c655742 tasks: run `systemctl daemon-reload` after httpd.service.d updates

martbab:
If you need the fix in ipa-4-4 you need to file a rebased PR against that branch.

PR #609, opened 2017-03-16
[4.4] Fix cookie with Max-Age processing

stlaz:


tomaskrizek:
ipa-4-4:

* 40f3b8f8a3d33864528138e517ce3240da6c9a4a Fix cookie with Max-Age processing
* 5caade99127ff46141d2f6b7137f7aa62c0ff3bc Add debug log in case cookie retrieval went wrong

PR #610, opened 2017-03-16
[4.3] Fix cookie with Max-Age processing

stlaz:

### DIFF
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -693,8 +693,11 @@ class KerbTransport(SSLTransport):
 
         # Search for the session cookie
         try:
-            session_cookie = Cookie.get_named_cookie_from_string(cookie_header,
-                                                                 COOKIE_NAME, request_url)
+            session_cookie = (
+                    Cookie.get_named_cookie_from_string(
+                        cookie_header, COOKIE_NAME, request_url,
+                        timestamp=datetime.datetime.utcnow())
+            )
         except Exception as e:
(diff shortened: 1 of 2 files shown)


ipalib/rpc.py
```diff
@@ -788,8 +791,10 @@ def get_session_cookie_from_persistent_storage(self, principal):
 
         # Search for the session cookie within the cookie string
         try:
-            session_cookie = Cookie.get_named_cookie_from_string(cookie_string, COOKIE_NAME)
-        except Exception as e:
+            session_cookie = Cookie.get_named_cookie_from_string(
+                cookie_string, COOKIE_NAME,
+                timestamp=datetime.datetime.utcnow())
+        except Exception:
```
MartinBasti: Debug log here might be useful, right?

stlaz: Sure, why not. I was just porting the code not really looking that way or another.

MartinBasti: In that case, create another commit that can be applied to master and 4.5 branch as well

MartinBasti:
Please open a new ticket `Backport ...`. Ticket you used is closed in closed milestone

stlaz:
The ticket's already there, I just recycled the old commit message and forgot to replace it.

tomaskrizek:
ipa-4-3:

* 0d66046e501a4a1a09a0a74a96a499cb88ffb03b Fix cookie with Max-Age processing
* 71475e3153117e554d22a2a27d7882ba4f890be8 Add debug log in case cookie retrieval went wrong

PR #611, opened 2017-03-17
Add debug log in case cookie retrieval went wrong

stlaz:
When backporting the fix, @MartinBasti pointed out we could use a debug log.

https://pagure.io/freeipa/issue/6774


tomaskrizek:
ipa-4-5:

* c59729d783993f60582f5cc6ca018545231df22b Add debug log in case cookie retrieval went wrong
master:

* 0bb858ea770e081817dc243579d08ad1f113e825 Add debug log in case cookie retrieval went wrong

PR #612, opened 2017-03-17
[4.5] Add debug log in case cookie retrieval went wrong

stlaz:
https://pagure.io/freeipa/issue/6774


tomaskrizek:
ipa-4-5:

* c59729d783993f60582f5cc6ca018545231df22b Add debug log in case cookie retrieval went wrong

PR #613, opened 2017-03-17
Constrain wheel package versions

tiran:
The presence of IPA packages on PyPI revealed an interesting issue with
make wheel_bundle. pip gives final releases a higher precedence than our
development packages. make wheel_bundle downloads ipa 4.5.0 from PyPI
instead of using our own wheels.

Use a constraint file to enforce correct versions.

https://pagure.io/freeipa/issue/6468

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ freeipa2-dev-doc
 /rpmbuild/
 # Build
 /ipasetup.py
+/.wheelconstraints
 *.egg-info
 
 # Subdirectories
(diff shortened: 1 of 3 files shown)


tiran:
@MartinBasti please finish the review. The issue makes my daily work harder and more complicated.

MartinBasti:
Works for me

tomaskrizek:
master:

* fe17d187f9f2cbac28fe369cbcdd697d85105481 Constrain wheel package versions
ipa-4-5:

* 7c93a518c8b6fb0e3a85bc1ae0ee807c71168213 Constrain wheel package versions

PR #614, opened 2017-03-17
[4.5] Constrain wheel package versions

tiran:
The presence of IPA packages on PyPI revealed an interesting issue with
make wheel_bundle. pip gives final releases a higher precedence than our
development packages. make wheel_bundle downloads ipa 4.5.0 from PyPI
instead of using our own wheels.

Use a constraint file to enforce correct versions.

https://pagure.io/freeipa/issue/6468

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ freeipa2-dev-doc
 /rpmbuild/
 # Build
 /ipasetup.py
+/.wheelconstraints
 *.egg-info
 
 # Subdirectories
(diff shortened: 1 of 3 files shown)


tomaskrizek:
ipatool supports pushing to multiple branches, so there's no need to open multiple PRs if the commits are exactly the same and don't have to be rebased.

tomaskrizek:
ipa-4-5:

* 7c93a518c8b6fb0e3a85bc1ae0ee807c71168213 Constrain wheel package versions

MartinBasti:
@tomaskrizek make sure that both commits are the same

tomaskrizek:
@MartinBasti I checked, they are.

PR #617, opened 2017-03-17
Allow renaming of sudo and HBAC rules

stlaz:
This simple hack adds a rename option to client side sudorule-mod
command.

https://pagure.io/freeipa/issue/2466

### DIFF
--- a/ipaserver/plugins/baseldap.py
+++ b/ipaserver/plugins/baseldap.py
@@ -36,7 +36,7 @@ from ipalib.text import _
 from ipalib.util import json_serialize, validate_hostname
 from ipalib.capabilities import client_has_capability
 from ipalib.messages import add_message, SearchResultTruncated
-from ipapython.dn import DN
+from ipapython.dn import DN, RDN
 from ipapython.version import API_VERSION
 
 if six.PY3:
@@ -549,7 +549,7 @@ class LDAPObject(Object):
     rdn_attribute = ''
     uuid_attribute = ''
     attribute_members = {}
-    rdn_is_primary_key = False # Do we need RDN change to do a rename?
+    allow_rename = False
     password_attributes = []
     # Can bind as this entry (has userPassword or krbPrincipalKey)
     bindable = False
@@ -1384,7 +1384,7 @@ class LDAPUpdate(LDAPQuery, crud.Update):
     def get_options(self):
         for option in super(LDAPUpdate, self).get_options():
             yield option
-        if self.obj.rdn_is_primary_key:
+        if self.obj.allow_rename:
             yield self._get_rename_option()
 
     def execute(self, *keys, **options):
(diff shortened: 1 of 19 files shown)


abbra:
I don't like it is done on the client side. This will not work for Web UI, for example.
Additionally, no validation of cn={newname} is here to be a single value RDN. If we add this as --setattr, we probably want to return meaningful error, not a general --setattr error.

stlaz:
Thank you Alexander for your insight. Since this was a hack, I did not want to do it server-wise. I chose a different approach to the problem and reworked the original idea so the rename option is now worked with on server.
With this approach, we are able to white-list objects which we think may be allowed renaming even though their primary keys are not in their RDN.

Just for the record, the names of sudo rules are still not checked for CN compatibility since their primary key is not part of their DN, but that's how things have been since for ever, I am afraid (you can try `ipa sudorule-add bad,cn=rule`).

ipaserver/plugins/baseldap.py
```diff
@@ -550,6 +550,7 @@ class LDAPObject(Object):
     uuid_attribute = ''
     attribute_members = {}
     rdn_is_primary_key = False # Do we need RDN change to do a rename?
+    allow_rename = False
```
HonzaCholasta: Why introduce a new class attribute when `rdn_is_primary_key` does exactly the same?

stlaz: Because we want to rename objects whose primary key is not RDN. But you're right, perhaps the attribute is just ill-named.

HonzaCholasta: IMO the attribute is entirely ill, as it does 2 different things: it allows renaming *and* forces DN update on rename.

I think we should finally get rid of it and use
* the `allow_rename` attribute to allow renaming and
* decide whether a DN update is required or not based on the current DN.

HonzaCholasta: Note that the above should be done in a separate commit and linked to https://pagure.io/freeipa/issue/2466.

abbra:
I like the idea but please address @HonzaCholasta comments.

stlaz:
For the record, and I might be wrong, I did a bit of researching, the `rdn_is_primary_key` is actually misused in some cases, as RDN is the primary key for e.g. `pwpolicy` and `idrange` but these have this attribute set to `False`.
I believe in the above cases, `rdn_is_primary_key` might have been used this way just so that those objects do not show the `rename` (they are not allowed to change the primary key anyway). I thought we won't need `allow_rename` at all in the end but for these cases we will probably need to keep it.

stlaz:
The latest patch removes the `rdn_is_private_key` attribute, replaces it with `allow_rename` which actually says correctly what's happening. Also, the decision whether primary key is in RDN is decided on checking whether the primary key is in RDN rather than on anything else.
Also added a comment explaining that the `modrdn` operation is performed also when `setattr` is doing changes to the primary key + RDN because it was far from obvious in the code.

MartinBasti:
I like the `allow_rename` attribute, as it really explains what is happaning, Also I like the reworked check if primary key is in DN because original `self.obj.primary_key.name in entry_attrs` may return false positive results.

I'm afraid about one thing. This will basically break custom user plugins if they used `rdn_is_private_key`. Shall we do some backward compatibility magic?

abbra:
I haven't seen any custom plugin that used `rdn_is_private_key`. We can document the change in release notes.

MartinBasti:
Please provide tests, LGTM otherwise

stlaz:
Added the tests but did not test them so we may want to see what Travis has to say about that.

stlaz:
*sigh* there was a rogue space.

Split into three separate commits.

MartinBasti:
Please update release notes (changelog)

stlaz:
Changelogs were updated.

pvomacka:
ipa-4-5:

* 28db6cd40100c6301121e3f82c074624fe53729c Reworked the renaming mechanism
* 85f2a19f88eef94ff080a42246658f572b5275f4 Allow renaming of the HBAC rule objects
* 7d3229bfb88f0fdc559245c8741563faba716106 Allow renaming of the sudorule objects
master:

* 8e4408e6784f929b4c3d861f0dd509335238e951 Reworked the renaming mechanism
* 55424c8677ba7de464c820afd31260aa4a7678d0 Allow renaming of the HBAC rule objects
* 8c1409155e9a9a978d3d763045a84d1eac585dfd Allow renaming of the sudorule objects

PR #621, opened 2017-03-18
Add --password-expiration to allow an admin to force a password change

redhatrises:
- Allows an admin to easily force a user to expire their password forcing them to change it.


abbra:
I would prefer this to be an option in `ipa passwd`, e.g. `ipa passwd --force-reset` which instead of modifying a user password would modify krbPasswordExpiration value.

redhatrises:
@abbra why not have it in both `ipa user-mod` and `ipa passwd`?

abbra:
Hm. `ipa user-mod` has --random and also supports specifying --password, so yes, both interfaces should be provided.

HonzaCholasta:
I don't agree. There should be one and only one obvious way to do it. There is no real benefit in having this in multiple different places, it just adds unnecessary complexity. Let's not repeat mistakes of the past and put this solely into `passwd`.

HonzaCholasta:
Actually, maybe `user-mod` is a better place for the option, as it does LDAP modify operation, whereas `passwd` does LDAP password change extended operation.

redhatrises:
Okay, so since it will reside in one location, should it be `user-mod` (PR already uses `user-mod`) or `passwd`?

abbra:
Ok, let's go with `user-mod` as original request goes, based on the fact that we are not changing the password, we are changing its properties.

LGTM.

HonzaCholasta:
I have given this some thought over the night - maybe we should make the option more generic and allow the user to specify the expiration time rather than special case it for "now" time, i.e. `--password-expiration=2017-03-21T07:58:05Z` to expire the password at a specific time, `--password-expiration=now` to expire the password now, just like `--force-password-reset` does.

redhatrises:
@HonzaCholasta that's an interesting idea. Most of the time, a password reset is forced immediately, but that does provide more flexibility. I assume that the datetime input should match the `2017-03-21T07:58:05Z` format?

HonzaCholasta:
@redhatrises, do not handle the format yourself, use the `DateTime` param type. Note that you will need to extend it to correctly interpret the "now" value.

ipalib/parameters.py
```diff
@@ -1665,12 +1666,16 @@ class DateTime(Param):
...
         if isinstance(value, six.string_types):
-            for date_format in self.accepted_formats:
-                try:
-                    time = datetime.datetime.strptime(value, date_format)
-                    return time
-                except ValueError:
-                    pass
+            if value == u'now':
+                time = datetime.datetime.strptime(strftime("%Y%m%d%H%M%SZ", gmtime()), "%Y%m%d%H%M%SZ")
+                return time
```
HonzaCholasta: Please use `datetime.utctime()`.

redhatrises:
@HonzaCholasta used `datetime.utcnow()` as I couldn't find a reference for `datetime.utctime()`

HonzaCholasta:
@redhatrises, `datetime.utcnow()` is what I meant.

redhatrises:
> @redhatrises, datetime.utcnow() is what I meant.

Oh good. Ready for your review.

HonzaCholasta:
The `admin` user is not allowed to write to the attribute:
```
$ kinit admin
Password for admin@ABC.IDM.LAB.ENG.BRQ.REDHAT.COM: 
$ ipa user-mod jcholast --password-expiration=now
ipa: ERROR: Insufficient access: Insufficient 'write' privilege to the 'krbPasswordExpiration' attribute of entry 'uid=jcholast,cn=users,cn=accounts,dc=abc,dc=idm,dc=lab,dc=eng,dc=brq,dc=redhat,dc=com'.
```
Please update the "Admin can manage any entry" ACI in `install/updates/20-aci.update`.

redhatrises:
@HonzaCholasta updated "Admins can write passwords" ACI to contain 'krbPasswordExpiration' as the "Admin can manage any entry" ACI already had 'krbPasswordExpiration' added.

VERSION.m4
```diff
@@ -74,7 +74,7 @@ define(IPA_DATA_VERSION, 20100614120000)
 ########################################################
 define(IPA_API_VERSION_MAJOR, 2)
 define(IPA_API_VERSION_MINOR, 224)
```
MartinBasti: Bump minor version please

install/updates/20-aci.update
```diff
@@ -54,7 +54,7 @@ remove:aci:(targetattr != "userPassword || krbPrincipalKey || sambaLMPassword ||
 add:aci:(targetattr != "userPassword || krbPrincipalKey || sambaLMPassword || sambaNTPassword || passwordHistory || krbMKey || krbPrincipalName || krbCanonicalName || krbPasswordExpiration || krbPwdHistory || krbLastPwdChange || krbExtraData || krbLastSuccessfulAuth || krbLastFailedAuth || ipaUniqueId || memberOf || enrolledBy || ipaNTHash || ipaProtectedOperation")(version 3.0; acl "Admin can manage any entry"; allow (all) groupdn = "ldap:///cn=admins,cn=groups,cn=accounts,$SUFFIX";)
 # Write-only
 remove:aci:(targetattr = "userPassword || krbPrincipalKey || sambaLMPassword || sambaNTPassword || passwordHistory")(version 3.0; acl "Admins can write passwords"; allow (add,delete,write) groupdn="ldap:///cn=admins,cn=groups,cn=accounts,$SUFFIX";)
-add:aci:(targetattr = "userPassword || krbPrincipalKey || sambaLMPassword || sambaNTPassword || passwordHistory || ipaNTHash")(version 3.0; acl "Admins can write passwords"; allow (add,delete,write) groupdn="ldap:///cn=admins,cn=groups,cn=accounts,$SUFFIX";)
```
MartinBasti: Please leave the original ACI there, just replace s/`add`/`remove`/ to avoid unholy mess with multiple ACIs (and keep the new one there as well :) )

redhatrises: @MartinBasti not sure I fully understand you correctly, but set the original ACI to remove and added a new ACI one.

MartinBasti: yes, add rule to remove current one from LDAP

HonzaCholasta:
@redhatrises, the "Admin can manage any entry" ACI in fact contains a blacklist of attributes which admins aren't allowed to write. To actually fix the issue you must also remove `krbPasswordExpiration` from the "Admin can manage any entry" ACI.

redhatrises:
@HonzaCholasta I also removed `krbPasswordExpiration` from the "Admin can manage any entry" ACI.

HonzaCholasta:
Works for me. Thanks!

MartinBasti:
master:

* 274b0bcf5ff2408739d94ba1b1b4bca69f310dfc Add --password-expiration to allow admin to force user password expiration

ACI.txt
```diff
@@ -351,7 +351,7 @@ aci: (targetattr = "member")(target = "ldap:///cn=ipausers,cn=groups,cn=accounts
 dn: cn=users,cn=accounts,dc=ipa,dc=example
 aci: (targetfilter = "(objectclass=posixaccount)")(version 3.0;acl "permission:System: Add Users";allow (add) groupdn = "ldap:///cn=System: Add Users,cn=permissions,cn=pbac,dc=ipa,dc=example";)
 dn: cn=users,cn=accounts,dc=ipa,dc=example
-aci: (targetattr = "krbprincipalkey || passwordhistory || sambalmpassword || sambantpassword || userpassword")(targetfilter = "(&(!(memberOf=cn=admins,cn=groups,cn=accounts,dc=ipa,dc=example))(objectclass=posixaccount))")(version 3.0;acl "permission:System: Change User password";allow (write) groupdn = "ldap:///cn=System: Change User password,cn=permissions,cn=pbac,dc=ipa,dc=example";)
+aci: (targetattr = "krbpasswordexpiration || krbprincipalkey || passwordhistory || sambalmpassword || sambantpassword || userpassword")(targetfilter = "(&(!(memberOf=cn=admins,cn=groups,cn=accounts,dc=ipa,dc=example))(objectclass=posixaccount))")(version 3.0;acl "permission:System: Change User password";allow (write) groupdn = "ldap:///cn=System: Change User password,cn=permissions,cn=pbac,dc=ipa,dc=example";)
```
simo5: Why do you need to add krbpasswordexpiration to the change user password option ?
I think it should be a separate (higher privileged permission) to tamper with password expirations.
Frontline helpdesk may be allowed to change a user password, but they are not allowed to change expiration for example.

ipaserver/plugins/user.py
```diff
@@ -261,7 +261,7 @@ class user(baseuser):
             ],
             'ipapermdefaultattr': {
                 'krbprincipalkey', 'passwordhistory', 'sambalmpassword',
-                'sambantpassword', 'userpassword'
+                'sambantpassword', 'userpassword', 'krbpasswordexpiration'
```
simo5: See above

PR #622, opened 2017-03-20
replica prepare: fix wrong IPA CA nickname in replica file

HonzaCholasta:
Lookup IPA CA subject and pass it to CertDB when creating dscert.p12 and
httpcert.p12, otherwise a generic nickname will be used for the IPA CA
certificate instead of "$REALM IPA CA".

This fixes replica install on domain level 0 from a replica file created
using ipa-replica-install on IPA 4.5.

https://pagure.io/freeipa/issue/6777


martbab:
I ran into some issues with CA deployment, but they vanished after re-provisioning my test env. Otherwise it works.

PR #623, opened 2017-03-20
client install: do not assume /etc/krb5.conf.d exists

HonzaCholasta:
Add `includedir /etc/krb5.conf.d` to /etc/krb5.conf only if
/etc/krb5.conf.d exists.

This fixes client install on platforms which do not have /etc/krb5.conf.d.

https://pagure.io/freeipa/issue/6589

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -641,13 +641,17 @@ def configure_krb5_conf(
             'value': 'File modified by ipa-client-install'
         },
         krbconf.emptyLine(),
-        {
-            'name': 'includedir',
-            'type': 'option',
-            'value': paths.COMMON_KRB5_CONF_DIR,
-            'delim': ' '
-        }
     ]
+
+    if os.path.exists(paths.COMMON_KRB5_CONF_DIR):
+        opts.extend([
+            {
+                'name': 'includedir',
+                'type': 'option',
+                'value': paths.COMMON_KRB5_CONF_DIR,
+                'delim': ' '
+            }
+        ])
 
     # SSSD include dir
     if configure_sssd:
--- a/.gitignore
+++ b/.gitignore
@@ -77,7 +77,6 @@ freeipa2-dev-doc
 /daemons/dnssec/ipa-ods-exporter.socket
 /daemons/ipa-kdb/ipa_kdb_tests
 /daemons/ipa-kdb/tests/.dirstamp
-/daemons/ipa-kdb/ipa-certauth
 /daemons/ipa-otpd/ipa-otpd
 /daemons/ipa-otpd/ipa-otpd.socket
 /daemons/ipa-otpd/ipa-otpd@.service
--- a/daemons/ipa-kdb/Makefile.am
+++ b/daemons/ipa-kdb/Makefile.am
@@ -44,12 +44,6 @@ dist_noinst_DATA = ipa_kdb.exports
 
 if BUILD_IPA_CERTAUTH_PLUGIN
 ipadb_la_SOURCES += ipa_kdb_certauth.c
-
-
-krb5confdir = $(sysconfdir)/krb5.conf.d
-krb5conf_DATA = ipa-certauth
-else
-dist_noinst_DATA += ipa-certauth
 endif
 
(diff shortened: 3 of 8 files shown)


tiran:
I'd rather create ```/etc/krb5.conf.d``` than to make the line conditional.

HonzaCholasta:
There is no reason to, the directory is not owned by us and we don't use it for anything anyway (see ticket triage for relevant discussion).

puiterwijk:
Would you be upgrading the krb5.conf after people upgrade krb5-libs to include the new includedir then?
Since that's what would happen if you don't change the krb5.conf and people update to a krb5-libs that has the includedir.

I've had to help a lot of people that ended up with configuration files lacking krb5.conf.d due to ipa-client setups (and other company configs, but at least that's limited to people working at companies giving broken krb5 configs).

ipaclient/install/client.py
```diff
@@ -641,14 +641,18 @@ def configure_krb5_conf(
...
-        {
-            'name': 'includedir',
-            'type': 'option',
-            'value': paths.COMMON_KRB5_CONF_DIR,
-            'delim': ' '
-        }
     ]
 
+    if os.path.exists(paths.COMMON_KRB5_CONF_DIR):
+        opts += [
```
puiterwijk: Why using opts += here, while the rest of the code uses opts.extend?

HonzaCholasta:
@puiterwijk, upgrade will be handled by krb5 itself, see https://bugzilla.redhat.com/show_bug.cgi?id=1431198.

ipaclient/install/client.py
HonzaCholasta: Because I didn't notice that.

lslebodn:
FYI: `/etc/krb5.conf.d` is not default include directory it is fedora/el7 specific.

debian testing has MIT kerberos 1.15 and  `/etc/krb5.conf.d` does not exist there as is not included in /etc/krb5.conf.

So +1 for @HonzaCholasta approach.

tiran:
The ipa-certauth plugin now starts to rely on the existence of ```/etc/krb5.conf.d```:

```
%config(noreplace) %{_sysconfdir}/krb5.conf.d/ipa-certauth
```

**Practicality beats purity**, let's make ```/etc/krb5.conf.d``` part of the offical FreeIPA configuation settings on all IPA enrolled systems.

lslebodn:
On (28/03/17 04:08), Christian Heimes wrote:
>The ipa-certauth plugin now starts to rely on the existence of ```/etc/krb5.conf.d```:
>
>```
>%config(noreplace) %{_sysconfdir}/krb5.conf.d/ipa-certauth
>```
>

The upstream spec file is fedora/rhel spec files and fedora+rhel have
`%{_sysconfdir}/krb5.conf.d/`. I cannot see any problem.

>**Practicality beats purity**, let's make ```/etc/krb5.conf.d``` part of the offical FreeIPA configuation settings on all IPA enrolled systems.
>

But neither debian nor arch linux/opensuse have this directory(or any other)
included by default in `/etc/krb5.conf`.

I would like to see standard directory for krb5 snippet files.
But that should be solved in distribution. And just used by freeipa.

LS

tiran:
**Practicality beats purity**

Let's define ```/etc/krb5.conf.d``` as part of our API and don't waste more time on shaving yet another yak. @tjaalton (Debian/Ubuntu maintainer) said

> <tjaalton> fine by me

frozencemetery:
(Note: a standard directory in distributions that freeipa could use would be provided by the krb5 maintainer, not the freeipa maintainer.)

frozencemetery:
Adding on to my previous comment: I've talked with the Debian maintainers, and they plan to add the same `includedir` after the Stretch release.  So, eventually (for some values of eventually) we won't have to worry about this.

HonzaCholasta:
@frozencemetery, this is *not* for the sake of Debian. We will still have to worry about this for operating systems which are not Fedora- or Debian-based.

martbab:
master:

* d5fc0ddd874f6289c0fcf60a6a4147ca5b280700 install: do not assume /etc/krb5.conf.d exists

PR #624, opened 2017-03-20
Use connection keep-alive

tiran:
Do not forcefully close the connection after every request. This enables
HTTP connection keep-alive, also known as persistent TCP and TLS/SSL
connection. Keep-alive speed up consecutive HTTP requests by 15% (for
local, low-latency network connections to a fast server) to multiple
times (high latency connections or remote peers).

pache has a default keep alive timeout of 5 seconds. That's too low for
interactive commands, e.g. password prompts. 30 seconds sounds like a
good compromise.

https://pagure.io/freeipa/issue/6641

### DIFF
--- a/install/conf/ipa.conf
+++ b/install/conf/ipa.conf
@@ -1,5 +1,5 @@
 #
-# VERSION 24 - DO NOT REMOVE THIS LINE
+# VERSION 25 - DO NOT REMOVE THIS LINE
 #
 # This file may be overwritten on upgrades.
 #
@@ -19,6 +19,11 @@ DirectoryIndex index.html
 # Substantially increase the request field size to support MS-PAC
 # requests, ticket #2767. This should easily support a 64KiB PAC.
 LimitRequestFieldSize 100000
+
+# Increase connection keep alive time. Default value is 5 seconds, which is too
+# short for interactive ipa commands. 30 seconds is a good compromise.
+KeepAlive On
+KeepAliveTimeout 30
 
 # ipa-rewrite.conf is loaded separately
 
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -686,7 +686,7 @@ class KerbTransport(SSLTransport):
                 return self.parse_response(response)
         except gssapi.exceptions.GSSError as e:
             self._handle_exception(e)
-        finally:
+        except BaseException:
             self.close()
 
     if six.PY3:
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -79,6 +79,13 @@ except ImportError:
     from xmlrpc.client import (Binary, Fault, DateTime, dumps, loads, ServerProxy,
             Transport, ProtocolError, MININT, MAXINT)
 
+# pylint: disable=import-error
+if six.PY3:
+    from http.client import RemoteDisconnected
+else:
+    from httplib import BadStatusLine as RemoteDisconnected
+# pylint: enable=import-error
+
 
 if six.PY3:
     unicode = str
@@ -531,6 +538,7 @@ class SSLTransport(LanguageAwareTransport):
         host, self._extra_headers, _x509 = self.get_host_info(host)
 
         if self._connection and host == self._connection[0]:
+            root_logger.debug("HTTP connection keep-alive (%s)", host)
             return self._connection[1]
 
         conn = create_https_connection(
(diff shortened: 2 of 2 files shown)


tomaskrizek:
I examined this in wireshark. Without this patch, `ipa vault-add` would establish 7 TCP connections to apache, while it establishes only 3 with this patch. I wasn't able to track down where are the 2 rogue connections opened and why. The situation is similar for other commands.

The question is whether this improvement is good enough or whether we want to optimize the RPC to actually use just a single connection.

Also, please follow the development process next time and assign yourself to the ticket when you start working on it, so other don't have to invest time into solving the same issue.

install/conf/ipa.conf
```diff
@@ -20,6 +20,11 @@ DirectoryIndex index.html
 # requests, ticket #2767. This should easily support a 64KiB PAC.
 LimitRequestFieldSize 100000
 
+# Increase connection keep alive time. Default value is 5 seconds, which is too
+# short for interactive ipa commands. 30 seconds is a good compromise.
+KeepAlive On
+KeepAliveTimeout 30
```
tiran: Your assumption is incorrect. For ```ipa``` client, the connection is automatically closed by ```ipaclient.api.Backend.rpcclient.disconnect()``` or when the client process exits. The 30 seconds keepalive timeout optimizes both ```ipa``` and browser sessions. For browser session I would even increase the keepalive timeout for 60 seconds in order to reduce the load on the server.

Modern webservers like Apache use high performance socket handling features of modern operating systems. ```epoll``` can easily handle thousands of connections. Repeating TCP handshake and TLS handshake over and over again cost magnitudes more performance than watching a bunch of additional sockets for I/O. A longer keepalive period won't be a problem until we have to handle thousands of client connections simultaneously.

ipalib/rpc.py
```diff
@@ -686,8 +695,18 @@ def single_request(self, host, handler, request_body, verbose=0):
                 return self.parse_response(response)
         except gssapi.exceptions.GSSError as e:
             self._handle_exception(e)
-        finally:
```
tiran: There are comments and debug logging calls.

install/conf/ipa.conf
tomaskrizek: Ok, didn't notice that the connections are actually closed elsewhere in ipa client.

tomaskrizek:
The extra connections seem to come from the internals of `httplib` library. If the hostname resolves to both IPv4 and IPv6 address, one connection is established to IPv4 and two to IPv6. I wasn't able to find the reason for this, but it doesn't seem to be related to our code.

```
connect(4, {sa_family=AF_INET, sin_port=htons(443), sin_addr=inet_addr("10.0.0.1")}, 16) = 0
connect(4, {sa_family=AF_INET6, sin6_port=htons(443), inet_pton(AF_INET6, "dead:beef::cafe", &sin6_addr), sin6_flowinfo=htonl(0), sin6_scope_id=0}, 28) = 0
connect(4, {sa_family=AF_INET6, sin6_port=htons(443), inet_pton(AF_INET6, "dead:beef::cafe", &sin6_addr), sin6_flowinfo=htonl(0), sin6_scope_id=0}, 28) = 0
```

tiran:
This behavior could be caused by https://github.com/python/cpython/blob/master/Lib/socket.py#L688 . What's ```socket.getaddrinfo(host, 443, 0, socket.SOCK_STREAM)``` for your host?

tomaskrizek:
@tiran I checked that code as well, `getaddrinfo` returns both IPv6 and IPv4. That could explain two connections, but I'm not sure where the third one comes from.

tomaskrizek:
master:

* 7beb6d1cad7e2200208cb14be6c823a89abf0dc3 Use connection keep-alive
* b2bdd2e1a912573ae4a3e8e5f40831a800d972f7 Add debug logging for keep-alive
* 7f567286f6b89f3e981af02913e833d3e8ed5064 Increase Apache HTTPD's default keep alive timeout
ipa-4-5:

* 25cf4a2e76ff976fe15029f9da7e4e3555f203d4 Use connection keep-alive
* f78439439c3c2ef6491fd5275de9d40b4b40a9b7 Add debug logging for keep-alive
* 4b426fbfa2dc83f1f43abbc2b9396bd9f1b07f74 Increase Apache HTTPD's default keep alive timeout

PR #625, opened 2017-03-20
[RFC] remote plugins: add option to force compat plugins

HonzaCholasta:
Add a new `force_client_compat` env flag to force client API not to do any
RPC calls to initialize remote plugins in `.finalize()` and use the newest
compat plugins instead.

Setting the flag serves as a workaround for `api.finalize()` requiring
valid Kerberos credentials.

https://pagure.io/freeipa/issue/6408

### DIFF
--- a/ipaclient/remote_plugins/compat.py
+++ b/ipaclient/remote_plugins/compat.py
@@ -12,6 +12,7 @@ import six
 from ipaclient.frontend import ClientCommand, ClientMethod
 from ipalib.frontend import Object
 from ipapython.ipautil import APIVersion
+from ipapython.version import API_VERSION
 
 if six.PY3:
     unicode = str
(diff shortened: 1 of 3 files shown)


HonzaCholasta:
@tiran, please review this, I'm not sure whether this kind of workaround is good enough or not.

tiran:
I don't understand the implications of this change and the new flag:

* What are the benefits and drawbacks of ```force_client_compat=False```?
* What are the benefits and drawbacks of ```force_client_compat=True```?
* Why does FreeIPA have schema download and compat plugins at all?
* Why is this feature implemented as *either/or* option instead of *optimistic try/fallback*?

New feature is missing unit and integration tests.

HonzaCholasta:
* With `force_client_compat=False`, the benefit is the client API matches the remote server API, the drawback is `api.finalize()` does RPC calls and touches schema cache (i.e. the current behavior).
* With `force_client_compat=True`, the benefit is `api.finalize()` does no RPC calls nor does it touch schema cache, the drawback is that the client API is stuck at API version 2.164 (IPA 4.3.3).
* Schema download exists to support newer servers versions without having to update the client. Compat plugins exist to support older server versions which do not have schema support. (See http://www.freeipa.org/page/V4/API_Compatiblity.)
* *Optimistic try/fallback* is the current behavior which requires RPC calls in `api.finalize()` to detect the server's capabilities in order to reconstruct it's API locally. With this PR it's possible to skip this step and fall back to the behavior of IPA 4.3.3.

PR #629, opened 2017-03-20
adtrust: make sure that runtime hostname result is consistent with the configuration

abbra:
FreeIPA's `ipasam` module to Samba uses gethostname() call to identify
own server's host name. This value is then used in multiple places,
including construction of cifs/host.name principal. `ipasam` module
always uses GSSAPI authentication when talking to LDAP, so Kerberos
keys must be available in the /etc/samba/samba.keytab. However, if
the principal was created using non-FQDN name but system reports
FQDN name, `ipasam` will fail to acquire Kerberos credentials.
Same with FQDN principal and non-FQDN hostname.

Also host name and principal name must have the same case.

Report an error when configuring ADTrust instance with inconsistent
runtime hostname and configuration. This prevents errors like this:

    [20/21]: starting CIFS services
    ipa         : CRITICAL CIFS services failed to start

    where samba logs have this:

    [2017/03/20 06:34:27.385307,  0] ipa_sam.c:4193(bind_callback_cleanup)
      kerberos error: code=-1765328203, message=Keytab contains no suitable keys for cifs/ipatrust@EXAMPLE.COM
    [2017/03/20 06:34:27.385476,  1] ../source3/lib/smbldap.c:1206(get_cached_ldap_connect)
      Connection to LDAP server failed for the 16 try!

Fixes https://pagure.io/freeipa/issue/6786

### DIFF
--- a/ipaserver/install/adtrustinstance.py
+++ b/ipaserver/install/adtrustinstance.py
@@ -27,6 +27,7 @@ import uuid
 import string
 import struct
 import re
+import socket
 
 import six
 
@@ -689,6 +690,15 @@ class ADTRUSTInstance(service.Service):
         except Exception as e:
             root_logger.critical("Enabling nsswitch support in slapi-nis failed with error '%s'" % e)
 
+    def __validate_server_hostname(self):
+        hostname = socket.gethostname()
+        if hostname != self.fqdn:
+            raise ValueError("Host reports different name than configured: "
+                             "'%s' versus '%s'. Samba requires to have "
+                             "the same hostname or Kerberos principal "
+                             "'cifs/%s' will not be found in Samba keytab." %
+                             (hostname, self.fqdn, self.fqdn))
+
     def __start(self):
         try:
             self.start()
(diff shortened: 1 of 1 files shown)


MartinBasti:
```
************* Module ipaserver.install.adtrustinstance

ipaserver/install/adtrustinstance.py:693: [E0602(undefined-variable), ADTRUSTInstance.__validate_server_hostname] Undefined variable 'socket')

PEP-8 errors:
./ipaserver/install/adtrustinstance.py:695:77: E502 the backslash is redundant between brackets
./ipaserver/install/adtrustinstance.py:696:73: E502 the backslash is redundant between brackets
./ipaserver/install/adtrustinstance.py:697:72: E502 the backslash is redundant between brackets
./ipaserver/install/adtrustinstance.py:698:79: E502 the backslash is redundant between brackets
```

abbra:
Removed backslashes and also moved the check to be the first step when creating an instance.

MartinBasti:
master:

* 0d817ae63a4ad8ba7a29910a9342a78e15e89593 adtrust: make sure that runtime hostname result is consistent with the configuration


ipa-4-5:

* e430699024df06e1e6f819824548986eb0fa5fd2 adtrust: make sure that runtime hostname result is consistent with the configuration

PR #632, opened 2017-03-21
ipa-sam: create the gidNumber attribute in the trusted domain entry

flo-renaud:
When a trusted domain entry is created, the uidNumber attribute is created
but not the gidNumber attribute. This causes samba to log
	Failed to find a Unix account for DOM-AD$
because the samu structure does not contain a group_sid and is not put
in the cache.
The fix creates the gidNumber attribute in the trusted domain entry,
and initialises the group_sid field in the samu structure returned
by ldapsam_getsampwnam. This ensures that the entry is put in the cache.

Note that this is only a partial fix for 6660 as it does not prevent
_netr_ServerAuthenticate3 from failing with the log
	_netr_ServerAuthenticate3: netlogon_creds_server_check failed. Rejecting auth request from client VM-AD machine account dom-ad.example.com.

https://pagure.io/freeipa/issue/6827

### DIFF
--- a/daemons/ipa-sam/ipa_sam.c
+++ b/daemons/ipa-sam/ipa_sam.c
@@ -195,6 +195,7 @@ struct ipasam_privates {
 	char *trust_dn;
 	char *flat_name;
 	struct dom_sid fallback_primary_group;
+	char *fallback_primary_group_gid_str;
 	char *server_princ;
 	char *client_princ;
 	struct sss_idmap_ctx *idmap_ctx;
@@ -2419,6 +2420,9 @@ static NTSTATUS ipasam_set_trusted_domain(struct pdb_methods *methods,
 	if (entry == NULL || sid == NULL) {
 		smbldap_make_mod(priv2ld(ldap_state), entry, &mods,
 				 LDAP_ATTRIBUTE_UIDNUMBER, IPA_MAGIC_ID_STR);
+		smbldap_make_mod(priv2ld(ldap_state), entry, &mods,
+		                 LDAP_ATTRIBUTE_GIDNUMBER,
+				 ldap_state->ipasam_privates->fallback_primary_group_gid_str);
 	}
 
 	if (td->netbios_name != NULL) {
@@ -2829,6 +2833,7 @@ static bool init_sam_from_td(struct samu *user, struct pdb_trusted_domain *td,
 {
 	NTSTATUS status;
 	struct dom_sid *u_sid;
+	struct dom_sid *g_sid;
 	char *name;
 	char *trustpw = NULL;
 	char *trustpw_utf8 = NULL;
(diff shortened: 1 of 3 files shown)


flo-renaud:
I updated the commit message with a different issue number, related to the "Failed to find a unix account" message.

daemons/ipa-sam/ipa_sam.c
```diff
@@ -2419,6 +2419,8 @@ static NTSTATUS ipasam_set_trusted_domain(struct pdb_methods *methods,
 	if (entry == NULL || sid == NULL) {
 		smbldap_make_mod(priv2ld(ldap_state), entry, &mods,
 				 LDAP_ATTRIBUTE_UIDNUMBER, IPA_MAGIC_ID_STR);
+		smbldap_make_mod(priv2ld(ldap_state), entry, &mods,
+		                 LDAP_ATTRIBUTE_GIDNUMBER, IPA_MAGIC_ID_STR);
```
abbra: @flo-renaud I think we need to change this. Autogeneration with DNA plugin wouldn't work here because a generated group ID does not mean the group object actually will exist. Given that TDO is created in the sub-tree that is not handled by the Managed Entries plugin, we should set gidNumber to a value of existing group.

Luckily, we have SMB Default Group for this. This is our fallback POSIX group for cases like this. Please fetch gidNumber of the SMB Default Group and use it instead here.

```
dn: cn=Default SMB Group,cn=groups,cn=accounts,dc=example,dc=com
cn: Default SMB Group
description: Fallback group for primary group RID, do not add users to this group
objectClass: top
objectClass: ipaobject
objectClass: posixgroup
objectClass: ipantgroupattrs
ipaUniqueID: 19ff163e-e60a-11e6-8794-001a4a62eb77
gidNumber: 1536000003
ipaNTSecurityIdentifier: S-1-5-21-570121326-3336757064-1157332047-1003
```

daemons/ipa-sam/ipa_sam.c
```diff
@@ -2884,6 +2893,23 @@ static bool init_sam_from_td(struct samu *user, struct pdb_trusted_domain *td,
 	}
 	talloc_free(u_sid);
 
+	tmp_ctx= talloc_init("init_sam_from_td");
```
abbra: Do not create empty context here. Create it off `user` object. It is guaranteed to be allocated with talloc too.

flo-renaud:
Hi @abbra 
thank you for the review. PR updated following your comments, and with an upgrade plugin to handle existing trusted domain objects.

abbra:
Thanks. I read through the code and it looks good to me. I'm going to test it together with my work on ipasam_update_sam_account() tomorrow.

abbra:
LGTM.

`nltest /sc_verify:ipa.example.test` works thanks to this pull request:
```
C:\Users\Administrator>nltest /sc_query:ipa.example.test
Flags: 30 HAS_IP  HAS_TIMESERV
Trusted DC Name \\master.ipa.example.test
Trusted DC Connection Status Status = 0 0x0 NERR_Success
The command completed successfully
```

MartinBasti:
master:

* e052c2dce04f5ce147dc2b6804f44705fa4d69df ipa-sam: create the gidNumber attribute in the trusted domain entry
* 5405de5bc15941d71137af10aa66a6cf922d9e6d Upgrade: add gidnumber to trusted domain entry


ipa-4-5:

* 91d36941653476abfff6a54ba7cb5a9f2c12c22d ipa-sam: create the gidNumber attribute in the trusted domain entry
* eddd29f1d52d63ea702437b0dd2a2826df52bc26 Upgrade: add gidnumber to trusted domain entry

PR #633, opened 2017-03-22
Support 8192-bit RSA keys in default cert profile

frasertweedale:
Update the caIPAserviceCert profile to accept 8192-bit RSA keys.
Affects new installs only, because there is not yet a facility to
update included profiles.

Fixes: https://pagure.io/freeipa/issue/6319


tomaskrizek:
master:

* 1530758475c2e21dd732581ff6816e03ca74dede Support 8192-bit RSA keys in default cert profile
ipa-4-5:

* 9118c08455d42f4e7f43370be1a858595a60bc9a Support 8192-bit RSA keys in default cert profile

PR #634, opened 2017-03-22
cert: do not limit internal searches in cert-find

HonzaCholasta:
Instead, apply the limits on the combined result.

This fixes (absence of) `--sizelimit` leading to strange behavior, such as
`cert-find --users user` returning a non-empty result only with
`--sizelimit 0`.

https://pagure.io/freeipa/issue/6716

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -1324,7 +1324,7 @@ class cert_find(Search, CertMethod):
 
         return result, False, True
 
-    def _ca_search(self, all, raw, pkey_only, sizelimit, exactly, **options):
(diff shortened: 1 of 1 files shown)


stlaz:
The tests obviously fail as they expect the `cert-find` command to respect the `sizelimit` option.

HonzaCholasta:
The fix was incomplete, it should be OK now.

stlaz:
Works for me.

martbab:
master:

* 6de507c2cad255975665eca6dd6ef7c8f2458d51 cert: do not limit internal searches in cert-find
ipa-4-5:

* 6382f9eee335907362a5ccb44b892f59de7d3751 cert: do not limit internal searches in cert-find

PR #636, opened 2017-03-22
[Py3] Fix ipatests.util doc tests

tiran:
Doctests of ipatests.util fail under Python 3.

The old test scenario does no longer work on Python 3 since u'how are you'
and 'how are you' have identical type, but u'how are you' != b'how are you'.
It works with int / float on all Python versions.

Python 2 has <type 'int'> while Python 3 uses <class 'int'>.


MartinBasti:
master:

* 397e6716974f90168792ec0a6ad0b7b37c02eb87 Fix ipatests.util doc tests

PR #637, opened 2017-03-22
ldap2: use LDAP whoami operation to retrieve bind DN for current connection

abbra:
For external users which are mapped to some DN in LDAP server, we
wouldn't neccesary be able to find a kerberos data in their LDAP entry.
Instead of searching for Kerberos principal use actual DN we are bound
to because for get_effective_rights LDAP control we only need the DN
itself.

Fixes https://pagure.io/freeipa/issue/6797

### DIFF
--- a/ipaserver/plugins/ldap2.py
+++ b/ipaserver/plugins/ldap2.py
@@ -286,12 +286,11 @@ class ldap2(CrudBackend, LDAPClient):
 
         assert isinstance(dn, DN)
 
-        principal = getattr(context, 'principal')
-        entry = self.find_entry_by_attr("krbprincipalname", principal,
-            "krbPrincipalAux", base_dn=self.api.env.basedn)
+        bind_dn = self.conn.whoami_s()[4:]
+
         sctrl = [
             GetEffectiveRightsControl(
-                True, "dn: {0}".format(entry.dn).encode('utf-8'))
+                True, "dn: {0}".format(bind_dn).encode('utf-8'))
         ]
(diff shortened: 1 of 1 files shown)


ipaserver/plugins/ldap2.py
```diff
@@ -286,12 +286,16 @@ def get_effective_rights(self, dn, attrs_list):
...
 
-        principal = getattr(context, 'principal')
-        entry = self.find_entry_by_attr("krbprincipalname", principal,
-            "krbPrincipalAux", base_dn=self.api.env.basedn)
+        bind_dn = None
+        try:
+            bind_dn = self.conn.whoami_s()[4:]
+        finally:
+            if bind_dn is None:
+                return None
```
martbab: Why do we need a `try:..finally` block here? I would rather do `bind_dn = self.conn.whoami_s()[4:]` and let the exception propagate to the callers if something bad happens. Or you may catch it and re-raise some public and more descriptive error.

martbab:
LGTM but lint has some complains probably related to my in-line comment.

pvomacka:
Hi @abbra, thank you for patch, works for me.

abbra:
Removed try: finally: block, I agree that it is better to propagate error up the stack.

martbab:
ipa-4-5:

* 7d48fb841a23e9f036f3d449d80623d1225c820a ldap2: use LDAP whoami operation to retrieve bind DN for current connection
master:

* 7324451834ec03786fda947679f750fe2a72f29c ldap2: use LDAP whoami operation to retrieve bind DN for current connection

PR #639, opened 2017-03-22
WebUI: Login for AD Users

pvomacka:
Allows login as AD user. AD Users has its own menu specification as there is visible only its profile and list of active IPA users. 

https://pagure.io/freeipa/issue/3242

### DIFF
--- a/install/ui/src/freeipa/Application_controller.js
+++ b/install/ui/src/freeipa/Application_controller.js
@@ -31,6 +31,7 @@ define([
         './widgets/App',
         './widgets/FacetContainer',
         './ipa',
+        './rpc',
         './reg',
         './config',
         './widget',
@@ -41,7 +42,7 @@ define([
         './plugins/load_page'
        ],
        function(declare, array, Deferred, on, topic, query, dom_class, auth,
-            JSON, App_widget, FacetContainer, IPA, reg, config, widget_mod,
+            JSON, App_widget, FacetContainer, IPA, rpc, reg, config, widget_mod,
             Menu, Router, routing, menu_spec) {
 
     /**
@@ -156,7 +157,7 @@ define([
         /**
          * Turns off one item in user dropdown menu and remove its listener.
          * @param {string} name of the user menu item which should be disabled
-         * @param {Object} listener disable this listener
+         * @param {Object} listener disable di
          */
         disable_user_menu_item: function(name, listener) {
             this.app_widget.disable_user_menu_item(name);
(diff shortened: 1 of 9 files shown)


install/ui/src/freeipa/ipa.js
```diff
@@ -650,13 +664,13 @@ IPA.update_password_expiration = function() {
...
-        args: [IPA.whoami.uid[0]]
+        args: [IPA.whoami.data.uid[0]]
     });
     reset_dialog.succeeded.attach(function() {
         var command = IPA.get_whoami_command();
         var orig_on_success = command.on_success;
         command.on_success = function(data, text_status, xhr) {
-            orig_on_success.call(this, data, text_status, xhr);
+            orig_on_success.call(this, data.result, text_status, xhr);
             IPA.update_password_expiration();
```
pvoborni: ` IPA.update_password_expiration()` method wasn't update to new `IPA.whoami` object content

install/ui/src/freeipa/Application_controller.js
```diff
@@ -255,13 +261,36 @@ define([
...
+        },
+
+        update_logged_in: function(logged_in) {
             this.app_widget.set('logged', logged_in);
+
+            var whoami = IPA.whoami;
+            var fullname = '';
+            var entity = whoami.metadata.object;
+
+            if (entity === 'user') fullname = whoami.data.krbprincipalname[0];
```
pvomacka: I can call `user-show login` with all and change the behavior to try to use `cn` and in case that there is no `cn` I can use `krbprincipalname`. Would it be better?

pvomacka:
I implemented all comments which you proposed and I also changed menu of AD user selfservice - I removed User tab and renamed User ID override to Profile.

abbra:
LGTM and works just fine:

pvoborni:
The code changes looks good to me.  ACK given that it works fine (@abbra 's comment).

martbab:
master:

* 1dcdcd12f4336c98e7507fe0e7f0c0da2bc69eba WebUI: check principals in lowercase
* 2992e3c5d480567cfdc71b38365d5d74f009b4d2 WebUI: add method for disabling item in user dropdown menu
* ceedc3f7ecb1300ed5bfaf5db8ef1b1450c6288e WebUI: Add support for login for AD users
ipa-4-5:

* bee9c9f090e7808a2381054fa63c1d036743296c WebUI: check principals in lowercase
* 01a0a38bdf53821bc420f01dc98fae577f83eabb WebUI: add method for disabling item in user dropdown menu
* 228e039e7d718ced7dce7c32cca3a89404c0a16e WebUI: Add support for login for AD users

PR #640, opened 2017-03-22
Remove pkinit options from master/replica on DL0

stlaz:
This patchset removes the ability of setting pkinit options on domain level 0 for server/replica installs. Also fixes a usability issue with `--no-pkinit` I noticed and did not care creating ticket for.

### DIFF
--- a/ipaserver/install/server/__init__.py
+++ b/ipaserver/install/server/__init__.py
@@ -340,15 +340,15 @@ class ServerInstallInterface(ServerCertificateInstallInterface,
         cert_file_opt = (self.pkinit_cert_files,)
         if not self.no_pkinit:
             cert_file_req += cert_file_opt
-        if any(cert_file_req + cert_file_opt) and not all(cert_file_req):
-            raise RuntimeError(
-                "--dirsrv-cert-file, --http-cert-file, and --pkinit-cert-file "
-                "or --no-pkinit are required if any key file options are used."
-            )
         if self.no_pkinit and self.pkinit_cert_files:
             raise RuntimeError(
                 "--no-pkinit and --pkinit-cert-file cannot be specified "
                 "together"
+            )
+        if any(cert_file_req + cert_file_opt) and not all(cert_file_req):
+            raise RuntimeError(
+                "--dirsrv-cert-file, --http-cert-file, and --pkinit-cert-file "
+                "or --no-pkinit are required if any key file options are used."
             )
 
         if not self.interactive:
--- a/ipaserver/install/server/__init__.py
+++ b/ipaserver/install/server/__init__.py
@@ -332,8 +332,23 @@ class ServerInstallInterface(ServerCertificateInstallInterface,
         if not os.path.exists(value):
             raise ValueError("File %s does not exist." % value)
 
+    def _is_promote(self):
+        """
+        :returns: True if domain level options correspond to domain level > 0
+        """
+        raise NotImplementedError()
+
     def __init__(self, **kwargs):
         super(ServerInstallInterface, self).__init__(**kwargs)
+
+        # pkinit is not supported on DL0, don't allow related options
+        if not self._is_promote():
+            if (self.no_pkinit or self.pkinit_cert_files is not None or
+                    self.pkinit_pin is not None):
+                raise RuntimeError(
+                    "pkinit on domain level 0 is not supported. Please "
+                    "don't use any pkinit-related options.")
+            self.no_pkinit = True
 
         # If any of the key file options are selected, all are required.
         cert_file_req = (self.dirsrv_cert_files, self.http_cert_files)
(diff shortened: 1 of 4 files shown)


ipaserver/install/server/__init__.py
```diff
@@ -335,21 +335,29 @@ def dirsrv_config_file(self, value):
     def __init__(self, **kwargs):
         super(ServerInstallInterface, self).__init__(**kwargs)
 
+        if self.domain_level == constants.DOMAIN_LEVEL_0:
+            if (self.no_pkinit or self.pkinit_cert_files is not None or
+                    self.pkinit_pin is not None):
+                raise RuntimeError(
+                    "pkinit on domain level 0 is not supported. Please don't "
```
Akasurde: Can you please edit man page as well do describe this change ?

stlaz: Sure, I will put the behavior of `no-pkinit` on domain level 0 to the correct man pages.
The reasoning to why this certain change was done is to be found in the commit message.

martbab:
@abbra I believe these changes are in line with our recent discussion regarding pkinit availability on DL0. Do you agree?

ipaserver/install/server/__init__.py
```diff
@@ -335,6 +335,14 @@ def dirsrv_config_file(self, value):
     def __init__(self, **kwargs):
         super(ServerInstallInterface, self).__init__(**kwargs)
 
+        if self.domain_level == constants.DOMAIN_LEVEL_0:
+            if (self.no_pkinit or self.pkinit_cert_files is not None or
```
MartinBasti: I'd drop `self.no_pkinit` from expression.

stlaz: If we allow `--no-pkinit` to be specified on DL0 the user might think that not-using it may change something for them IMO.

abbra:
Good question. I think we should remove all mentioning of PKINIT options for DL0 and explicitly configure local CA there. On DL1 we already require to provide pkinit cert for CA-less setup. However, there we should treat --no-pkinit as use of local CA (certmonger's one).

ipaserver/install/server/__init__.py
MartinBasti: But it is contradicting, user specifies `--no-pkinit` and got error `pkinit on domain level 0 is not supported`

stlaz: That's why there's the part not to use any pkinit-related options.

MartinBasti:
```
ipa-replica-install --no-pkinit  (as negative test without master installed)

2017-03-27T17:04:09Z DEBUG Logging to /var/log/ipareplica-install.log
2017-03-27T17:04:09Z DEBUG   File "/usr/lib/python2.7/site-packages/ipapython/admintool.py", line 172, in execute
    return_value = self.run()
  File "/usr/lib/python2.7/site-packages/ipapython/install/cli.py", line 314, in run
    cfgr = transformed_cls(**kwargs)
  File "/usr/lib/python2.7/site-packages/ipapython/install/common.py", line 102, in __init__
    **kwargs)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/__init__.py", line 602, in __init__
    super(ServerReplicaInstall, self).__init__(**kwargs)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/__init__.py", line 338, in __init__
    if self.domain_level == constants.DOMAIN_LEVEL_0:
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 611, in __getattr__
    raise AttributeError(name)

2017-03-27T17:04:09Z DEBUG The ipa-replica-install command failed, exception: AttributeError: domain_level

```

stlaz:
Ah, right, replica does not have `domain_level` option 🙄

MartinBasti:
With this PR applied I cannot use webUI with DL0

stlaz:
@MartinBasti Even though this commit basically breaks the behavior, it's not in its scope to fix it, it's somehow intended to break it, actually. It will be fixed elsewhere.

I fixed the issue with running this on replica and removed one redundant check as well.

I also noticed that DL0 replica has a usability issue where it checks for either `*-cert-file` option and requires them all, once it has it, it will say that these options can't be used with replica file. I will not fix that here, though.

stlaz:
Pushed a cleaner version of the previous changes, thanks @HonzaCholasta for the suggestion.

martbab:
@MartinBasti WebUI not working in DL0/--no-pkinit is beyond the scope of this PR. I am working on fixing that in a separate PR.

MartinBasti:
master:

* 6cda1509a68d7a21578280d381a6b9e994fd4f49 Fix the order of cert-files check
* 9e3ae785ac9b62b8e0809a4aa56363c458316135 Don't allow setting pkinit-related options on DL0
* 8af884d0489d5d57895959d27ca6eb8815c6c922 replica-prepare man: remove pkinit option refs
* fe7cf1e854b7dc28861455011091df3cbe45abe9 Remove redundant option check for cert files


ipa-4-5:

* 497e766427b3ced865ff88a51cd0c2c96e8b24f9 Fix the order of cert-files check
* a1ad1ffa3540da4b5d5c1963b3818d9c9260e1a2 Don't allow setting pkinit-related options on DL0
* 85720b6bdc764b98dd471799ccc1045e1379709e replica-prepare man: remove pkinit option refs
* 8f7b6c349f4e81e88ef36f014e26de6b1f3f3e41 Remove redundant option check for cert files

PR #641, opened 2017-03-22
Set "KDC:Disable Last Success" by default

MartinBasti:
In big deployments enabled recording of the last sucesfull login
this creates a huge changelog on DS side and cause performance
issues even if this is excluded from replication.

Actually this is not used directly by FreeIPA so it is safe to remove
in new installations. User who need this must manually remove
"KDC:Disable Last Success" using `ipa config-mod` command or WebUI.

https://pagure.io/freeipa/issue/5313

### DIFF
--- a/install/share/bootstrap-template.ldif
+++ b/install/share/bootstrap-template.ldif
@@ -410,6 +410,7 @@ ipaUserObjectClasses: ipasshuser
 ipaDefaultEmailDomain: $DOMAIN
 ipaMigrationEnabled: FALSE
 ipaConfigString: AllowNThash
+ipaConfigString: KDC:Disable Last Success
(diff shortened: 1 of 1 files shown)


MartinBasti:
Bump for review

stlaz:
This change of default behavior seems to have removed a lot of write-load from DS so I believe we should go with it. However, add comment about it to the issue "changelog", please.

pvomacka:
ipa-4-5:

* fdcd5f486839d9279dcba74b74f7756ace5812fa Set "KDC:Disable Last Success" by default
master:

* eeaf428b1befc37489ed5ee14ae193b46cbd1db7 Set "KDC:Disable Last Success" by default

PR #644, opened 2017-03-23
extdom: improve certificate request

sumit-bose:
Certificates can be assigned to multiple user so the extdom plugin must use
sss_nss_getlistbycert() instead of sss_nss_getnamebycert() and return a
list of fully-qualified user names.

Due to issues on the SSSD side the current version of lookups by
certificates didn't work at all and the changes here won't break existing
clients.

Related to https://pagure.io/freeipa/issue/6646

Since I used the revers lookup for the domain separator in patch I added a
second patch which does this where needed in the reminder of the code as well
to be consistent. Allthough using @-signs in short names is not common practice
it might happen as can be see in https://pagure.io/SSSD/sssd/issue/3219.

The sss_nss_getlistbycert() call is added to SSSD in
https://github.com/SSSD/sssd/pull/207.

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom.h
+++ b/daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom.h
@@ -95,7 +95,8 @@ enum response_types {
     RESP_USER,
     RESP_GROUP,
     RESP_USER_GROUPLIST,
-    RESP_GROUP_MEMBERS
+    RESP_GROUP_MEMBERS,
+    RESP_NAME_LIST
 };
 
 struct extdom_req {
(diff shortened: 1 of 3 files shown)


abbra:
LGTM. I read the code but since SSSD counterpart is currently on review, travis fails the build.

dkupka:
Works for me.

tomaskrizek:
master:

* ee455f163d756a6b71db8e999365139cad46c6ad extdom: do reverse search for domain separator
* 8960398a57f69c124ec3105289dc355baa0d5b09 extdom: improve cert request
ipa-4-5:

* 8046f9baab1e93b8b8e11d05088c8cdabdd47281 extdom: do reverse search for domain separator
* a510a3d7e9f37e89acee84bed2363cb7f57fe88e extdom: improve cert request

PR #645, opened 2017-03-23
Create temporaty directories at the begining of uninstall

dkupka:
Since commit 38c6689 temporary directories are no longer created at package
install time. Instead they're created at server install time.
Some steps in uninstall also assume that temporary direcories exist. Creating
the directories in the begining of server uninstall ensure that the uninstall
will go through.

https://pagure.io/freeipa/issue/6715

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -1042,6 +1042,10 @@ def uninstall(installer):
 
     rv = 0
 
+    # further steps assumes that temporary directories exists so rather
+    # ensure they are created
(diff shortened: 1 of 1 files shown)


MartinBasti:
Not the nicest but good enough for now

dkupka:
master:

* 3dcd3426310ccacdb1564ad7fe83358110a044f6 Create temporaty directories at the begining of uninstall
ipa-4-5:

* c0a395776f3c9e4f4612fa16bb6af40646c3cdbf Create temporaty directories at the begining of uninstall

PR #650, opened 2017-03-24
CA-less installation fix

stlaz:
These patches fix the CA-less installation by guessing the names for CA and server-cert nicknames in /etc/httpd/alias. The fix is not very nice since it's guessing but I am not sure if there's anything else we can do at this point.

Also, `HTTPInstance.start/stop_tracking_certificates` would probably not need the guessing since it's only relevant for CA-full installations where we know the server-cert nickname is `Server-Cert` so I can replace it there if you think that'd be better.

### DIFF
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -118,6 +118,7 @@ class WebGuiInstance(service.SimpleServiceInstance):
     def __init__(self):
         service.SimpleServiceInstance.__init__(self, "ipa_webgui")
 
+
 class HTTPInstance(service.Service):
     def __init__(self, fstore=None, cert_nickname='Server-Cert',
                  api=api):
@@ -130,6 +131,7 @@ class HTTPInstance(service.Service):
             service_user=HTTPD_USER,
             keytab=paths.HTTP_KEYTAB)
 
+        self.cacert_nickname = None
         self.cert_nickname = cert_nickname
         self.ca_is_configured = True
         self.keytab_user = constants.GSSPROXY_USER
@@ -441,6 +443,9 @@ class HTTPInstance(service.Service):
             if not server_certs:
                 raise RuntimeError("Could not find a suitable server cert.")
 
+        # store the CA cert nickname so that we can publish it later on
+        self.cacert_nickname = db.cacert_name
+
     def __import_ca_certs(self):
         db = certs.CertDB(self.realm, nssdir=paths.HTTPD_ALIAS_DIR,
                           subject_base=self.subject_base)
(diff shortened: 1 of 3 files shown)


ipaserver/install/httpinstance.py
```diff
@@ -382,18 +385,20 @@ def __setup_ssl(self):
...
+            # this is CA-less and we don't know the server-cert nickname
+            if self.cert_nickname is None:
+                # We only handle one server cert
+                self.cert_nickname = server_certs[0][0]
             self.create_password_conf()
 
-            # We only handle one server cert
-            nickname = server_certs[0][0]
-            if nickname == 'ipaCert':
-                nickname = server_certs[1][0]
```
HonzaCholasta: Removing this is not safe, there are various reasons `ipaCert` might be present in /etc/httpd/alias even now (previous install of older IPA version, etc.)

ipaserver/install/httpinstance.py
```diff
@@ -441,7 +446,8 @@ def __import_ca_certs(self):
     def __publish_ca_cert(self):
         ca_db = certs.CertDB(self.realm, nssdir=paths.HTTPD_ALIAS_DIR,
                              subject_base=self.subject_base)
-        ca_db.publish_ca_cert(paths.CA_CRT)
+        ca_nickname = ca_db.find_root_cert(self.cert_nickname)[-1]
+        ca_db.export_pem_cert(ca_nickname, paths.CA_CRT)
```
HonzaCholasta: If you add this line at the end of `__setup_ssl()`:
```python
        self.cacert_name = db.cacert_name
```
You can replace this change and all of the other changes with this single line here, above `ca_db.publish_ca_cert()`:
```python
        ca_db.cacert_name = self.cacert_name
```

stlaz: Ah, right, I did not realize this will be set to the correct value during the initialization from pkcs12.

stlaz:
Fixed according to the comments, thanks.

HonzaCholasta:
@stlaz, `NSSDatabase.publish_ca_cert()` and `CertDB.publish_ca_cert()` become unused after your changes, could we remove them?

stlaz:
Huh, I did not notice, thanks. Removing it, I could have removed the big fat comment about why not to use it.

HonzaCholasta:
@stlaz, please rebase.

HonzaCholasta:
I found additional bugs in CA-less (replica) install, but with this PR, `publish_ca_cert` does not fail anymore.

HonzaCholasta:
Actually, there is a pylint failure introduced by this PR:
```
************* Module ipapython.certdb
ipapython/certdb.py:579: [C0305(trailing-newlines), ] Trailing newlines)
```

stlaz:
Sorry, must have screwed up the rebase.

HonzaCholasta:
@stlaz, please also provide a version of this PR rebased on ipa-4-5.

stlaz:
Done in https://github.com/freeipa/freeipa/pull/685

HonzaCholasta:
master:

* 8c87014e199b3dbe885c69d40a01d2723f813c3e Get correct CA cert nickname in CA-less
* aae9a918b68dc4f9a7b4fb9abf1bb4d26673109d Remove publish_ca_cert() method from NSSDatabase

PR #652, opened 2017-03-24
dogtag-ipa-ca-renew-agent-submit: fix the is_replicated() function

flo-renaud:
dogtag-ipa-ca-renew-agent-submit behaves differently depending on the
certificate it needs to renew. For instance, some certificates (such as IPA RA)
are the same on all the hosts and the renewal is actually done only on
the renewal master. On other nodes, the new cert is downloaded from LDAP.

The function is_replicated() is returning the opposite as what it should. If
the cert nickname is IPA RA, it should return that the cert is replicated but
it doesn't, and this leads to a wrong code path to renew the cert.

https://pagure.io/freeipa/issue/6813

### DIFF
--- a/install/certmonger/dogtag-ipa-ca-renew-agent-submit
+++ b/install/certmonger/dogtag-ipa-ca-renew-agent-submit
@@ -119,7 +119,7 @@ def is_renewable():
 
 
 def is_replicated():
-    return not get_nickname()
+    return bool(get_nickname())
 
 
 def is_renewal_master():


HonzaCholasta:
master:

* e934da09d5e738c735f874931dd1b54d79b3150b dogtag-ipa-ca-renew-agent-submit: fix the is_replicated() function
ipa-4-5:

* 8f738f1ea9f86a921e3dc0fd02e57419f3173ed9 dogtag-ipa-ca-renew-agent-submit: fix the is_replicated() function

PR #654, opened 2017-03-24
spec: update url and sources

tomaskrizek:
Point Source to upstream sources on pagure and add signature file.
Use https for project URL.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -69,8 +69,9 @@ Summary:        The Identity, Policy and Audit system
 
 Group:          System Environment/Base
 License:        GPLv3+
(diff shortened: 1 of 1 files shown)


MartinBasti:
LGTM

tomaskrizek:
It looks like the build somehow depends on the specfile, because it fails with

```
error: Bad file: /home/sharp/git/freeipa/rpmbuild/SOURCES/freeipa-4.5.90.dev201703241540+gitcec2cfb.tar.gz.asc: No such file or directory
```

I guess we can omit this in upstream.

Fedora spec files have been updated and their build works fine.

PR #655, opened 2017-03-27
httpinstance.disable_system_trust: Don't fail if module 'Root Certs' …

dkupka:
…is not available

Server installation failed when attmpting to disable module 'Root Certs' and
the module was not available in HTTP_ALIAS_DIR. When the module is not
available there's no need to disable it and the error may be treated as
success.

https://pagure.io/freeipa/issue/6803

### DIFF
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -355,9 +355,17 @@ class HTTPInstance(service.Service):
         name = 'Root Certs'
         args = [paths.MODUTIL, '-dbdir', paths.HTTPD_ALIAS_DIR, '-force']
 
-        result = ipautil.run(args + ['-list', name],
-                             env={},
-                             capture_output=True)
+        try:
+            result = ipautil.run(args + ['-list', name],
+                                 env={},
+                                 capture_output=True)
+        except ipautil.CalledProcessError as e:
+            if e.returncode == 29:  # ERROR: Module not found in database.
+                root_logger.debug(
+                    'Module %s not available, treating as disabled', name)
+                return False
+            raise
+
         if 'Status: Enabled' in result.output:
             ipautil.run(args + ['-disable', name], env={})
             return True


ipaserver/install/httpinstance.py
```diff
@@ -355,9 +355,15 @@ def disable_system_trust(self):
...
 
-        result = ipautil.run(args + ['-list', name],
-                             env={},
-                             capture_output=True)
+        try:
+            result = ipautil.run(args + ['-list', name],
+                                 env={},
+                                 capture_output=True)
+        except ipautil.CalledProcessError:
+            root_logger.debug(
```
tiran: This catches and ignores all call error. I would rather limit the exception to exit code 29:

```
$ mkdir /tmp/nssdb
$ certutil -N -d /tmp/nssdb/ --empty-password
$ modutil -dbdir /tmp/nssdb/ -force -list 'Root Certs'
ERROR: Module "Root Certs" not found in database.
$ echo $?
29
```

stlaz:
Hm, I believe the `-list` operation was there just to check whether the module is there. If `modutil` fails like this no matter the situation, it'd be better to just try to disable it whatever happens and go away with "meh, so what" if the disable fails.

stlaz:
For the record:
```bash
[slaznick@vm-066 ~]$ sudo modutil -dbdir nssdb/ -disable 'Root Certs' -force
ERROR: Module "Root Certs" not found in database.
[slaznick@machine ~]$ echo $?
29
[slaznick@machine ~]$ 
```

tiran:
@stlaz The broad except also catches and ignores typos in the command line or missing ```modutil``` binary.

stlaz:
@tiran I of course agree on narrowing the broad except down, my point is we should rather remove the whole `-list` part and just try to do `-disable`.

dkupka:
@tiran @stlaz Makes sense. I will update it accordingly. Thanks for suggestions.

HonzaCholasta:
@stlaz, you can't do just `-disable`, as that would break upgrade (note that the `-list` is there because `-disable` always reports success).

stlaz:
@HonzaCholasta You're right, I completely forgot about that one.

stlaz:
This fixes the mentioned issue. I did not test whether the actual disable works but I should hope so as I don't see how this could break it.

tomaskrizek:
ipa-4-5:

* 2a499551ca5ddf2596cc19a77f47c34e9f5c10c5 httpinstance.disable_system_trust: Don't fail if module 'Root Certs' is not available
master:

* 0128e805e591bc8ca5cea99739ad4cd7478df0b4 httpinstance.disable_system_trust: Don't fail if module 'Root Certs' is not available

PR #657, opened 2017-03-27
configure: fix --disable-server with certauth plugin

sumit-bose:
Resolves https://pagure.io/freeipa/issue/6816

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -231,6 +231,8 @@ dnl ---------------------------------------------------------------------------
 AM_CONDITIONAL([BUILD_IPA_CERTAUTH_PLUGIN],
                [test x$have_certauth_plugin = xyes -a x"$SSSCERTMAP_LIBS" != x])
 AM_COND_IF([BUILD_IPA_CERTAUTH_PLUGIN], [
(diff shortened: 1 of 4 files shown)


lslebodn:
`AM_CONDITIONAL` need to be executed every time.

So just `AM_CONDITIONAL([BUILD_IPA_CERTAUTH_PLUGIN]` need to be moved from sever.m4 -> configure.ac. The best would be to build somewhere after `m4_include(server.m4)`

tiran:
LGTM

I verified that the PR fixes

* --disable-server
* --enable-server with old version of krb5 that does not have ```krb5/certauth_plugin.h```

pvomacka:
ipa-4-5:

* 203d5416ce807f5cdcf9e2431feef84d49b3df61 configure: fix --disable-server with certauth plugin
* 8fde0b88d7c9360e16820d6086eba3e3ca0eee1e ipa-kdb: do not depend on certauth_plugin.h
master:

* 054f1bd78b04a79f765f524f829b34c0ee252a1b configure: fix --disable-server with certauth plugin
* 0ba0c0781367d8e2d4affca29e3cf5ab93c4c33a ipa-kdb: do not depend on certauth_plugin.h

PR #659, opened 2017-03-27
WebUI: Allow to add certs to certmapping with CERT LINES around

pvomacka:
The certificate to the certmapping might be inserted as
base64 encoded blob. This patch allows to also insert the certificate
blob with surrounding "-----BEGIN CERTIFICATE-----" and
"-----END CERTIFICATE-----" lines. This behavior is the same in
widget for assigning certificates to users, so the change helps
WebUI to be more consistent.

https://pagure.io/freeipa/issue/6772

### DIFF
--- a/install/ui/src/freeipa/plugins/certmap.js
+++ b/install/ui/src/freeipa/plugins/certmap.js
@@ -8,6 +8,7 @@ define([
         'dojo/_base/declare',
         'dojo/Evented',
         'dojo/on',
+        '../certificate',
         '../navigation',
         '../field',
         '../ipa',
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @pvomacka ,

thank you for the patch, it works as expected.

pvoborni:
Code LGTM, ACK give that it works for @flo-renaud

pvomacka:
ipa-4-5:

* eda23a9847197513555f6237b46c658365dfc12d WebUI: Allow to add certs to certmapping with CERT LINES around
master:

* 84b38b6793cbc45d36c39abf79893e22e90baac6 WebUI: Allow to add certs to certmapping with CERT LINES around

PR #660, opened 2017-03-27
rpcserver.login_x509: Actually return reply from __call__ method

dkupka:
__call__ didn't return causing internal error in wsgi application. Previously
this bug was hidden by some other error and the code worked even though it
shouldn't.


pvomacka:
Please change ticket to this one: https://pagure.io/freeipa/issue/6819

pvomacka:
ipa-4-5:

* c80941e98bfd00c1c6e530aa4a592354adff8d90 rpcserver.login_x509: Actually return reply from __call__ method
master:

* 7e1fdd2c5881893fd9540689045a11f9e88beef9 rpcserver.login_x509: Actually return reply from __call__ method

PR #661, opened 2017-03-28
git-commit-template: update ticket url to use pagure.io instead of fe…

flo-renaud:
…dorahosted.org

After the migration to pagure.io, tickets are accessed through another URL.

In order to use the commit template:
git config commit.template .git-commit-template

https://pagure.io/freeipa/issue/6822

### DIFF
--- a/.git-commit-template
+++ b/.git-commit-template
@@ -2,4 +2,4 @@ component: Subject
 
 Explanation
 
-https://fedorahosted.org/freeipa/ticket/XXXX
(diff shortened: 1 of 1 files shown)


stlaz:
ACK, stopping the tests as the change does not have anything to do with our codebase.

tomaskrizek:
master:

* f17460a34ce452b46d431850aa565efd6c7b23ba git-commit-template: update ticket url to use pagure.io instead of fedorahosted.org

PR #663, opened 2017-03-28
Generate PIN for PKI to help Dogtag in FIPS

stlaz:
Dogtag is currently unable to generate a PIN it could use for
an NSS database creation in FIPS. Generate it for them so that
we don't fail.

https://pagure.io/freeipa/issue/6824


tomaskrizek:
ipa-4-5:

* 39eac72faef5f44c9fb2cad943ad58d23fe60cf3 Generate PIN for PKI to help Dogtag in FIPS
master:

* e204d030fc4154800acb0b2b312188e72dd80f80 Generate PIN for PKI to help Dogtag in FIPS

PR #665, opened 2017-03-28
Allow erasing ipaDomainResolutionOrder attribute

fidencio:
Currently when trying to erase the ipaDomainResolutionOrder attribute we
hit an internal error as the split() method is called on a None object.

By returning early in case of empty string we now allow removing the
ipaDomainResolutionOrder attribute by both doing calling delattr or
setting its value to an empty string.

https://pagure.io/freeipa/issue/6825

Signed-off-by: Fabiano Fidêncio <fidencio@redhat.com>

### DIFF
--- a/ipaserver/plugins/config.py
+++ b/ipaserver/plugins/config.py
@@ -359,6 +359,11 @@ class config(LDAPObject):
 
         domain_resolution_order = entry_attrs[attr_name]
 
+        # setting up an empty string means that the previous configuration has
+        # to be cleaned up/removed. So, do nothing and let it pass
+        if not domain_resolution_order:
+            return
+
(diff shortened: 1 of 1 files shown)


MartinBasti:
LGTM

fidencio:
I have updated the commit message as I found a typo there: "both doing calling delattr or" -> "both calling delattr or"

MartinBasti:
master:

* e03056cf34de5ba0100d62f008d76e8c851c3ba7 Allow erasing ipaDomainResolutionOrder attribute


ipa-4-5:

* 08a921cc08b5b841260caa2e45653a704b88542c Allow erasing ipaDomainResolutionOrder attribute

PR #631, opened 2017-03-21
Upgrade: configure PKINIT after adding anonymous principal

martbab:
In order to set up PKINIT, the anonymous principal must already be
created, otherwise the upgrade with fail when trying out anonymous
PKINIT. Switch the order of steps so that this issue does not occur.

https://pagure.io/freeipa/issue/6792

PR #666, opened 2017-03-28
Fix anonymous principal handling in replica install

martbab:
This PR should unblock replica install against <4.5 masters if `--no-pkinit`
option is given. Be aware of the non-working WebUI after install, this will be
fixed once local PKINIT will be implemented.

Requires https://github.com/freeipa/freeipa/pull/631

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -1482,20 +1482,6 @@ def add_default_caacl(ca):
     sysupgrade.set_upgrade_state('caacl', 'add_default_caacl', True)
 
 
-def enable_anonymous_principal(krb):
-    princ_realm = krb.get_anonymous_principal_name()
-    dn = DN(('krbprincipalname', princ_realm), krb.get_realm_suffix())
-    try:
-        _ = api.Backend.ldap2.get_entry(dn)  # pylint: disable=unused-variable
-    except ipalib.errors.NotFound:
-        krb.add_anonymous_principal()
-
-    try:
-        api.Backend.ldap2.set_entry_active(dn, True)
-    except ipalib.errors.AlreadyActive:
-        pass
-
-
 def setup_pkinit(krb):
     root_logger.info("[Setup PKINIT]")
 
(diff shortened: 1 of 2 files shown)


ipaserver/install/server/upgrade.py
```diff
@@ -1499,15 +1499,14 @@ def enable_anonymous_principal(krb):
...
-        return
-
     if not api.Command.ca_is_enabled()['result']:
         root_logger.info("CA is not enabled")
         return
 
-    krb.setup_pkinit()
+    if not os.path.exists(paths.KDC_CERT):
+        root_logger.info("Requesting PKINIT certificate")
+        krb.setup_pkinit()
```
stlaz: You should probably stop the execution here in case PKINIT is already set up.

stlaz:
I actually did the review of https://github.com/freeipa/freeipa/pull/631 alongside this.
I do not think the order of adding the anonymous principal and setting up PKINIT matters that much. From what I saw in Kerberos guides, it's usually actually done after PKINIT setup since until then, the anonymous principal is pretty much unusable.
The problem was rather the testing of anonymous pkinit before the anonymous principal was added, that is just plainly weird and I'm glad that that's now fixed.
ACK since this fixes the issues mentioned in comments.

ipaserver/install/server/upgrade.py
stlaz: If you could perhaps add the outcome of the offline discussion here, please :)

MartinBasti:
needs rebase

MartinBasti:
master:

* 191668e85be0b53020a56df409731812e528d101 Always check and create anonymous principal during KDC install
* 2eabb0dab7b4dab1c45395f3e02d43676d91f4a2 Remove duplicate functionality in upgrade


ipa-4-5:

* ce94f7fa7b4eca296d2f9692d35c2558bfeddb46 Always check and create anonymous principal during KDC install
* 0fcd56533a00c28f9f8f800c77b8c2c580cb3a8f Remove duplicate functionality in upgrade

PR #667, opened 2017-03-28
idrange-add: properly handle empty --dom-name option

flo-renaud:
When idrange-add is called with --dom-name=, the CLI exits with
This happens because the code checks if the option is provided but does not
check if the value is None.

We need to handle empty dom-name as if the option was not specified.

https://pagure.io/freeipa/issue/6404

### DIFF
--- a/ipaserver/plugins/idrange.py
+++ b/ipaserver/plugins/idrange.py
@@ -411,7 +411,7 @@ class idrange_add(LDAPCreate):
 
         # This needs to stay in options since there is no
         # ipanttrusteddomainname attribute in LDAP
-        if 'ipanttrusteddomainname' in options:
+        if options.get('ipanttrusteddomainname'):
             if is_set('ipanttrusteddomainsid'):
                 raise errors.ValidationError(name='ID Range setup',
                     error=_('Options dom-sid and dom-name '


stlaz:
LGTM, except you're talking about `idrange-mod` in the commit message but are fixing `idrange-add` (`idrange-mod` does not have the option at all).

flo-renaud:
Hi @stlaz 
I fixed the commit message.

In contrary to what I told you offline, you need to configure an AD trust with ipa-adtrust-install and ipa trust-add ... in order to reproduce the original issue. My bad...

stlaz:
@flo-renaud That's completely OK :)
I thought we could probably add an assert to `CIDict.__contains__()` method since I realize the issue was somewhere else than this fixed check, but the current situation fails verbosely enough so that's probably fine.
ACK.

martbab:
@flo-renaud can you please add a test case for this to `ipatests/test_xmlrpc/test_range_plugin.py` so that we do not regress in the future?

tomaskrizek:
master:

* 70743c8c48db54309a09d510b3a5d8ae86c29e58 idrange-add: properly handle empty --dom-name option


ipa-4-5:

* 077a61524d79ac5ab6f0eb46450c82ad5594bd2b idrange-add: properly handle empty --dom-name option

flo-renaud:
@martbab 
thank you for the suggestion. The new test is available in PR #692

PR #669, opened 2017-03-29
server: make sure we test for sss_nss_getlistbycert

abbra:
Fixes https://pagure.io/freeipa/issue/6828


tiran:
AC_CHECK_LIB only checks for functions in libs. Compilation may still fail if header files and library are not in sync. IMHO we don't have to care about this broken case.

LGTM

abbra:
On the systems where pkg-config is available, positive result from pkg-config check means headers are available because pkg-config files are part of development sub-packages. Symbol check in a library is enough then.

tomaskrizek:
Works as expected.

tomaskrizek:
master:

* 67e5244cad72bef76de1c4df47a0c77a672fa861 server: make sure we test for sss_nss_getlistbycert
ipa-4-5:

* 8be6987da72dff0ebd4e02c946b45b5b1705d880 server: make sure we test for sss_nss_getlistbycert

PR #670, opened 2017-03-29
[Py3] session storage parameters must be bytes

tiran:
Fixes TypeError: bytes or integer address expected instead of str instance


MartinBasti:
master:

* d06315de6b1e951d6cce7d7d6495a32b44216274 session storage parameters must be bytes

PR #672, opened 2017-03-29
IPA-KDB: use relative path in ipa-certmap config snippet

sumit-bose:
Architecture specific paths should be avoided in the global Kerberos
configuration because it is read e.g. by 32bit and 64bit libraries they
are installed in parallel.

Resolves https://pagure.io/freeipa/issue/6833

### DIFF
--- a/daemons/ipa-kdb/Makefile.am
+++ b/daemons/ipa-kdb/Makefile.am
@@ -40,18 +40,16 @@ ipadb_la_SOURCES = 		\
 	ipa_kdb_audit_as.c	\
 	$(NULL)
 
+dist_noinst_DATA = ipa_kdb.exports
+
 if BUILD_IPA_CERTAUTH_PLUGIN
 ipadb_la_SOURCES += ipa_kdb_certauth.c
 
 
-%: %.in
-	sed \
-		-e 's|@plugindir@|$(plugindir)|g' \
-		'$(srcdir)/$@.in' >$@
-
 krb5confdir = $(sysconfdir)/krb5.conf.d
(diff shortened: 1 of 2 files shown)


tiran:
LGTM

For the recording: according to https://web.mit.edu/kerberos/krb5-1.12/doc/admin/conf_files/krb5_conf.html#plugins the plugin directive uses ```plugin_base_dir``` as base dir:

> module
> This tag may have multiple values. Each value is a string of the form modulename:pathname, which causes the shared object located at pathname to be registered as a dynamic module named modulename for the pluggable interface. If pathname is not an absolute path, it will be treated as relative to the plugin_base_dir value from [libdefaults].

> plugin_base_dir
> If set, determines the base directory where krb5 plugins are located. The default value is the krb5/plugins subdirectory of the krb5 library directory.

@sumit-bose What happens when the shared library is missing? Does 32bit kinit fail or work on a X86_64 system when 32bit ipadb.so is missing?

abbra:
> @sumit-bose What happens when the shared library is missing? Does 32bit kinit fail or work on a X86_64 system when 32bit ipadb.so is missing?

It is not about kinit. The module is for KDC, not client side. We guarantee it exists because we install it.

HonzaCholasta:
master:

* 6c2772dde52c84024d32533b29e6cbd04c69924a IPA-KDB: use relative path in ipa-certmap config snippet


ipa-4-5:

* fa46a01c37021e7b2b57fd3092383100e39792fb IPA-KDB: use relative path in ipa-certmap config snippet

PR #674, opened 2017-03-29
Replace hard-coded kdcproxy path with WSGI script

tiran:
mod_wsgi has no way to import a WSGI module by dotted module name. A new
kdcproxy.wsgi script is used to import kdcproxy from whatever Python
version mod_wsgi is compiled against. This will simplify moving FreeIPA
to Python 3 and solves an import problem on Debian.

Resolves: https://pagure.io/freeipa/issue/6834

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1256,6 +1256,7 @@ fi
 # END
 %dir %{_usr}/share/ipa
 %{_usr}/share/ipa/wsgi.py*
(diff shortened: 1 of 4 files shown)


MartinBasti:
master:

* 2cd6788c3f52a9b87f24b9b3e57d66a864397966 Replace hard-coded kdcproxy path with WSGI script

PR #675, opened 2017-03-29
[4.5, master] Fix PKCS11 helper

MartinBasti:
Slots in HSM are not assigned statically, we have to chose proper
slot from token label.

Softhsm i2.2.0 changed this behavior and now slots can change over
time (it is allowed by pkcs11 standard).

Changelog:
* created method get_slot() that returns slot number from
  used label
* replaces usage of slot in __init__ method of P11_Helper
  with label
* slot is dynamically detected from token label before
  session is opened
* pkcs11-util --init-token now uses '--free' instead '--slot'
  which uses first free slot (we don't care about slot numbers
  anymore)

https://pagure.io/freeipa/issue/6692

### DIFF
--- a/daemons/dnssec/ipa-dnskeysync-replica
+++ b/daemons/dnssec/ipa-dnskeysync-replica
@@ -15,6 +15,7 @@ import os
 import sys
 
 import ipalib
+from ipalib.constants import SOFTHSM_DNSSEC_TOKEN_LABEL
 from ipalib.install.kinit import kinit_keytab
 from ipapython.dn import DN
 from ipapython.ipa_log_manager import root_logger, standard_logging_setup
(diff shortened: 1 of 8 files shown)


stlaz:
`cffi.api.CDefError: cannot parse "typedef CK_RV (*CK_C_GetSlotList) (CK_BBOOL tokenPresent,`
-> you're using CK_BBOOL type before defining it.

MartinBasti:
In 50% cases DNSSEC works for me :-). Ready for review.

(The issue was unrelated to PKCS11, I had somehow broken machine probably)

stlaz:
I ran the integration test and the result was fine, ACK.

MartinBasti:
ipa-4-5:

* e6b2ed6b68589ff7ee39b95559836af54f39e2de Fix PKCS11 helper


master:

* e8f2a415b3dcba30b0c39cd542acd6b459f46957 Fix PKCS11 helper

PR #677, opened 2017-03-30
cert: defer cert-find result post-processing

HonzaCholasta:
Rather than post-processing the results of each internal search,
post-process the combined result.

This avoids expensive per-certificate searches on certificates which won't
even be included in the combined result when cert-find is executed with the
--all option.

https://pagure.io/freeipa/issue/6808

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -250,6 +250,11 @@ def normalize_pkidate(value):
     return datetime.datetime.strptime(value, PKIDATE_FORMAT)
 
 
+def convert_pkidatetime(value):
+    value = datetime.datetime.fromtimestamp(int(value) // 1000)
+    return x509.format_datetime(value)
+
+
 def validate_csr(ugettext, csr):
     """
     Ensure the CSR is base64-encoded and can be decoded by our PKCS#10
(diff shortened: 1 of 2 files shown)


ipaserver/plugins/cert.py
```diff
@@ -1469,18 +1451,17 @@ def _ldap_search(self, all, raw, pkey_only, no_members, **options):
                     try:
                         obj = result[key]
                     except KeyError:
-                        obj = self._get_cert_obj(cert, all, raw, pkey_only)
+                        obj = {}
```
stlaz: Indeed, this is a bug. I added a user certificate. Now, when I do `ipa cert-find --pkey-only`, the result is:
```
-----------------------
11 certificates matched
-----------------------

...

-----------------------------
Number of entries returned 11
-----------------------------
```
but only 10 certificates are shown.

I would also expect that when I specify `--pkey-only` only primary keys are shown, which is not true but I suspect it's behaved like this for quite some time - only few certificates are displayed with limited information with the previous implementation. The `--pkey-only` option does not seem to have any effect with this patch, though, except that some certificates are not displayed at all.

stlaz:
What worries me the most is that the tests are green even though this is potentially a serious problem.

stlaz:
The patched IPA works better than the current 4.4 and 4.5 branches in terms of options logic, that's good.
From the code I am not sure which searches we do miss, could you elaborate on that a bit, please?

stlaz:
We may need these changes in 4.5 and 4.4, too since `cert-find` is rather broken there, too.

HonzaCholasta:
That might require backporting [issue 6564](https://pagure.io/freeipa/issue/6564) as well.

HonzaCholasta:
master:

* eb6d4c3037d0cc269a7924745f1cbd8f647e6e1a cert: defer cert-find result post-processing


ipa-4-5:

* 49f9d799c171c7ae2ac546a33a353c2c40b4719c cert: defer cert-find result post-processing

PR #679, opened 2017-03-31
Make sure remote hosts have our keys

simo5:
In complex replication setups a replica may try to obtain CA keys from a
host that is not the master we initially create the keys against.
In this case race conditions may happen due to replication. So we need
to make sure the server we are contacting to get the CA keys has our
keys in LDAP. We do this by waiting to positively fetch our encryption
public key (the last one we create) from the target host LDAP server.

Fixes: https://pagure.io/freeipa/issue/6838

### DIFF
--- a/ipaserver/install/custodiainstance.py
+++ b/ipaserver/install/custodiainstance.py
@@ -1,6 +1,6 @@
 # Copyright (C) 2015 FreeIPa Project Contributors, see 'COPYING' for license.
 
-from ipaserver.secrets.kem import IPAKEMKeys
+from ipaserver.secrets.kem import IPAKEMKeys, KEMLdap
 from ipaserver.secrets.client import CustodiaClient
 from ipaplatform.paths import paths
 from ipaplatform.constants import constants
@@ -18,6 +18,7 @@ import shutil
 import os
 import stat
 import tempfile
+import time
 import pwd
 
 
@@ -122,12 +123,37 @@ class CustodiaInstance(SimpleServiceInstance):
         cli = self.__CustodiaClient(server=master_host_name)
         cli.fetch_key('dm/DMHash')
 
+    def __wait_keys(self, host, timeout=300):
+        ldap_uri = 'ldap://%s' % host
+        deadline = int(time.time()) + timeout
+        root_logger.info("Waiting up to {} seconds to see our keys "
+                         "appear on host: {}".format(timeout, host))
+
+        konn = KEMLdap(ldap_uri)
+        saved_e = None
+        while True:
+            try:
+                return konn.check_host_keys(self.fqdn)
+            except Exception as e:
+                # log only once for the same error
+                if not isinstance(e, type(saved_e)):
+                    root_logger.debug(
+                        "Transient error getting keys: '{err}'".format(err=e))
+                    saved_e = e
+                if int(time.time()) > deadline:
+                    raise RuntimeError("Timed out trying to obtain keys.")
+                time.sleep(1)
+
     def __get_keys(self, ca_host, cacerts_file, cacerts_pwd, data):
         # Fecth all needed certs one by one, then combine them in a single
         # p12 file
 
         prefix = data['prefix']
         certlist = data['list']
+
+        # Before we attempt to fetch keys from this host, make sure our public
+        # keys have been replicated there.
+        self.__wait_keys(ca_host)
 
         cli = self.__CustodiaClient(server=ca_host)
 
--- a/ipaserver/secrets/kem.py
+++ b/ipaserver/secrets/kem.py
@@ -24,6 +24,7 @@ import ldap
 
 IPA_REL_BASE_DN = 'cn=custodia,cn=ipa,cn=etc'
 IPA_KEYS_QUERY = '(&(ipaKeyUsage={usage:s})(memberPrincipal={princ:s}))'
+IPA_CHECK_QUERY = '(cn=enc/{host:s})'
 RFC5280_USAGE_MAP = {KEY_USAGE_SIG: 'digitalSignature',
                      KEY_USAGE_ENC: 'dataEncipherment'}
 
@@ -77,6 +78,17 @@ class KEMLdap(iSecLdap):
         jwk = self._parse_public_key(ipa_public_key)
         jwk['use'] = KEY_USAGE_MAP[usage]
         return json_encode(jwk)
+
+    def check_host_keys(self, host):
+        conn = self.connect()
+        scope = ldap.SCOPE_SUBTREE
+
+        ldap_filter = self.build_filter(IPA_CHECK_QUERY, {'host': host})
+        r = conn.search_s(self.keysbase, scope, ldap_filter)
+        if len(r) != 1:
+            raise ValueError("Incorrect number of results (%d) searching for"
+                             "public key for %s" % (len(r), host))
+        return True
 
     def _format_public_key(self, key):
         if isinstance(key, str):


simo5:
I haven't tested this yet ... but what could possibily go wrong? :-)

pvoborni:
Shouldn't the ticket number be: https://pagure.io/freeipa/issue/6838 ?

simo5:
Seem like both errors are the same problem.
Should we mark 6688 a duplicate of 6838 ?

simo5:
Nevermind they are not duplicates.
I'll fix the commit message.

stlaz:
Fails with
```2017-04-12T14:16:14Z DEBUG The ipa-replica-install command failed, exception: ValueError: Incorrect number of results (0) searching forpublic key for host/vm-225.abc.idm.lab.eng.brq.redhat.com@DOM-096.ABC.IDM.LAB.ENG.BRQ.REDHAT.COM
```
on first replica, every try.

ipaserver/install/custodiainstance.py
```diff
@@ -122,13 +124,34 @@ def import_dm_password(self, master_host_name):
...
+        deadline = int(time.time()) + timeout
+        root_logger.info("Waiting up to {} seconds to see our keys "
+                         "appear on host: {}".format(timeout, host))
+
+        konn = KEMLdap(ldap_uri)
+        while True:
+            try:
+                konn.get_key(KEY_USAGE_ENC, principal)
+                return
+            except Exception:
```
stlaz: Could you use a more specific exception? I noticed that should not keys be retrieved, we're getting `ValueError`. We don't want to be waiting here if we're unable to connect to remote master.

simo5: What do you suggest ?

simo5: note that if we do not wait we fail intallation, not sure we gain anything by not waiting, and we can deal with a transient connection error if we wait.

stlaz: I don't suggest not waiting, of course. I would have gone for just `ValueError` but you're probably right that the connection error might appear.
It might be worth to do this, at least:
```python
exc = None
while int(time.time()) < deadline:
    try:
        konn.get_key(KEY_USAGE_ENC, principal)
        break
    except ValueError as e:
        exc = e
        time.sleep(1)
    except Exception as e:
        root_logger.debug("Failed to get keys: '{err}'".format(err=e))
        exc = e
        time.sleep(1)
else:
    if exc is not None:
        raise exc
    else:
        raise RuntimeError("Unable to obtain keys in expected time.")
```

simo5: Ok I handled this (but differently) and rebased on top of current master.
Please check it out @stlaz

pvoborni:
What is this PR waiting for?

stlaz:
I was expecting some action about my previous comment:

> Fails with
>  2017-04-12T14:16:14Z DEBUG The ipa-replica-install command failed, exception: ValueError: Incorrect number of results (0) searching forpublic key for host/vm-225.abc.idm.lab.eng.brq.redhat.com@DOM-096.ABC.IDM.LAB.ENG.BRQ.REDHAT.COM
> on first replica, every try.

I did not see any change in code to fix this but I can try again.

stlaz:
Still fails.

simo5:
Can you please attach more of the logs before the failure ?

simo5:
@stlaz just FYI, I am sking this info because I cannot reproduce locally with a single replica.

simo5:
Nevermind I finally reproduced

simo5:
Turned out my master had some more relaxed permissions I added when developing the feature.
I now have added a new function to just check for the host keys without asking for data that cannot be read with the identity we have available.
This has been tested and seems to work correctly.
Please check @stlaz

stlaz:
@simo5 will check, sorry for not replying yesterday, I was no more at my machine.

ipaserver/secrets/kem.py
```diff
@@ -78,6 +79,17 @@ def get_key(self, usage, principal):
...
+    def check_host_keys(self, host):
+        conn = self.connect()
+        scope = ldap.SCOPE_SUBTREE
+
+        ldap_filter = self.build_filter(IPA_CHECK_QUERY, {'host': host})
+        r = conn.search_s(self.keysbase, scope, ldap_filter)
+        if len(r) != 1:
+            raise ValueError("Incorrect number of results (%d) searching for"
+                             "public key for %s" % (len(r), host))
+        return True
```
stlaz: The `return True` seems a bit redundant.

stlaz:
Seems to work fine against current master, but fails with
```
Configuring ipa-custodia
  [1/4]: Generating ipa-custodia config file
  [2/4]: Generating ipa-custodia keys
  [3/4]: starting ipa-custodia 
  [4/4]: configuring ipa-custodia to start on boot
Done configuring ipa-custodia.

ipa.ipapython.install.cli.install_tool(CompatServerReplicaInstall): ERROR    503 Server Error: Service Unavailable for url: https://vm-096.abc.idm.lab.eng.brq.redhat.com/ipa/keys/ca/caSigningCert%20cert-pki-ca?type=kem&value=eyJhbGciOiJSU0EtT0FFUCIsImVuYyI6IkEyNTZDQkMtSFM1MTIiLCJraWQiOm51bGx9.k6y2jmI8oxRIsieU93_RzG5mZU_u_DPW2XL2jjLukYPZ3oZOkLkufof0fBeH6LAR66aL9m5C9j26GmhlTqNsm2FUQT7Xql975rYR3veooDwLQlPx6k4X1J4CTEeSsf7RVj8KfLE5e4K-nW1hTyepsbm7RDAA_-tbLvWzEqCQ0I3bfpPEDmlML08FA9T_yuPb1FkT0-lSCLV5PHya4tOB3R2q5CHC2b6BpwZQtbVW8eohshEmJMTO2NMAyPlfJscgSHYmhi6oliToV_Dh90Ej1UH_S0UOkHLsvIV5IoW4EGeaGdeHwHo4GsSGHGN3exVxWk9GShhJ_WJ-dlXSGQ_9CA.SfWWO_VrqzKKX3EYSh3E1Q.n4GtjcFZOQSZmAG9MShIQVtfRv_N3jEQMS46rLGUU6xIS-BYBL0Xq1UWP6VFrZW-g96Iqe2PIBhv4m1FsuAzP_gzac1lCr2ghcVuj3rAUg81G5s8vPuYNl_Ur5UVlQ2LtWzGLc26s1z_43MF7qCl8iayvXqnweK8_kj54F1RUJ-Awp0--Z4mnK_FFrPU4BBW2_EjZ1tOR8dV7NnxnN2Gd2tiDFl6Kkbj91rf6Bo2f8telN5RJsX52PsNW2z-l78TOIAKY4qfHhSVz31RO3xgUbyu3yQ79sGIxD66hzmVisB_LnbpNHbIjCP1wKEXXSo-IPrDtXk7ZWZrEITtItzynbzBKddVLjcNMjoqGz-lhLWVNg8R8rdHEdUzhlkdM-kFfW6Fz57wSyOZnt4KvQ-lZxY62TLQB1gqJ7vhzUPUs1g7C9rsy4gTQPjuRxXnLRvqXSb3arQPkrUl_hLqRuAm8FL-ClYY9G38KVns81QTygKvkDC8E5LQBJfyzkg93AyTXNBcrdCxP8AGgaxLBlGyEX-ya0g3mVX5fz_Uj6gyKjtOS_x1AUHOMkAMRmVEzvixrz-krCMWYOQDmJi19OlNeNjb7-NUVDxPRryr7e6Po2OqSbSjP6kUSw_QbMZf8BCrqV4TUFOwndTmZ68n1TOrCqie-UO71TJnherD_3m60_t3-Li1uy6_WWX66BBEMCCtsZBJWP7OYj7c9CzWGuzUEI7g75i4TZwoM1z0SjuyoPE.ZbRawj1B943OeF6AD_W0Z3pfk13fs14rbj_Ab8n-ZXI
ipa.ipapython.install.cli.install_tool(CompatServerReplicaInstall): ERROR    The ipa-replica-install command failed. See /var/log/ipareplica-install.log for more information
```
against 4.4.4 master.

simo5:
I've seen this once but thought it was a fluke due to my "unclean" master, as the following times it did not happen.
Can you reproduce the error against 4.4.4 consistently ?

ipaserver/secrets/kem.py
simo5: I can remove it.

stlaz:
I was able to do it two times in a row with the same master, I can try to reinstall both the master and replica if you want. What do you mean "unclean"? It's a clean 4.4.4 master, no code changes, `/etc/httpd/alias` and `/etc/pki/pki-tomcat/alias` NSS databases seem fine, too.

simo5:
I meant my setup was unclean.
I will try to reproduce here.
Does master w/o this patch work properly against 4.4.4 ?

stlaz:
Not sure, I will try that.

stlaz:
It seems that replica install fails even without this patch so it's OK to go with it?

simo5:
We need to find why it breaks though, but yeah I think we can go forward with this patch of others agree.
Can you open a separate bug for the failure you got ?

stlaz:
Will do, ACKing this in the meantime.

stlaz:
Removing the ACK to retest on 4.4.4 with Fedora custodia version.

tomaskrizek:
ipa-4-5:

* 5f8d1119fe38807e86930af50d3680e28efe68eb Make sure remote hosts have our keys


master:

* 1f9f84a66d6cf9391b91ee4a13ac0f1119212578 Make sure remote hosts have our keys

PR #680, opened 2017-03-31
ipa-otpd.socket.in: Use a platform specific value for KDC service file

tjaalton:
https://pagure.io/freeipa/issue/6845


tomaskrizek:
I believe Christian's suggestion in the updated commit. Looks good to me.

tomaskrizek:
master:

* 076eb409a032cbd689f1d5e298c1009e80168e34 ipa-otpd.socket.in: Use a platform specific value for KDC service file

PR #681, opened 2017-04-02
Fix ipadiscovery

alex-zel:
Sort SRV records for LDAP/KRB based on priority.

### DIFF
--- a/ipaclient/install/ipadiscovery.py
+++ b/ipaclient/install/ipadiscovery.py
@@ -21,6 +21,7 @@ import socket
 
 import six
 
+from operator import attrgetter
 from ipapython.ipa_log_manager import root_logger
 from dns import resolver, rdatatype
 from dns.exception import DNSException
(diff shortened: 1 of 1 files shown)


tiran:
You can simplify your code a lot with the operator module and sorted(key) trick:

https://docs.python.org/3/library/operator.html#operator.attrgetter
https://docs.python.org/3/library/functions.html#sorted

```
import operator
```

```
answers = resolver.query(qname, rdatatype.SRV)
answers = sorted(answer, key=operator.attrgetter('priority'))
```

Please squash your changes into one commit.

martbab:
Hi Alex, a few comments:

1.) please see PEP8 guide for correct Python formatting https://www.python.org/dev/peps/pep-0008/ namely, do not use tabs but 4 spaces for indentation.

2.) I do not see much value in sorting TXT records. We are searching for _kerberos TXT record which should occur only once in DNS domain.

3.) please use a more concise sorting mechanism mentioned by @tiran, your way is very unpythonic and inefficient (list insertions).

ipaclient/install/ipadiscovery.py
```diff
@@ -521,7 +530,16 @@ def ipadnssearchkrbrealm(self, domain=None):
         root_logger.debug("Search DNS for TXT record of %s", qname)
 
         try:
-            answers = resolver.query(qname, rdatatype.TXT)
+            answers = []
+            dns_answers = resolver.query(qname, rdatatype.SRV)
```
MartinBasti: You changed `TXT` to `SRV` here. It should stay TXT RR and it has no priority attribute

PR #684, opened 2017-04-03
httpinstance: make sure NSS database is backed up

HonzaCholasta:
The NSS database at /etc/httpd/alias is not properly initialized and backed
up in CA-less replica promotion. This might cause the install to fail after
previous install and uninstall.

Make sure the NSS database is initialized and backed up even in CA-less
replica promotion to fix the issue.

https://pagure.io/freeipa/issue/4639

### DIFF
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -373,10 +373,11 @@ class HTTPInstance(service.Service):
         return False
 
     def __setup_ssl(self):
+        truncate = not self.promote or not self.ca_is_configured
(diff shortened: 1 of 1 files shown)


stlaz:
Without this patch, I encountered a different issue but with the same root cause. The patch fixes it, so ACK.

HonzaCholasta:
master:

* 5f5a3b29dba7cc736ba334aefb55484baeefeb76 httpinstance: make sure NSS database is backed up


ipa-4-5:

* 471dfcbe1cc3f319da788add3661cb6d63e3c0f0 httpinstance: make sure NSS database is backed up

PR #685, opened 2017-04-03
[4.5] Caless fix

stlaz:
Get correct CA cert nickname in CA-less
    
During CA-less installation, we initialize the HTTPD alias
database from a pkcs12 file. This means there's going to
be different nicknames to the added certificates. Store
the CA certificate nickname in HTTPInstance__setup_ssl()
to be able to correctly export it late

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -596,12 +596,3 @@ class NSSDatabase(object):
         finally:
             del certdb, cert
             nss.nss_shutdown()
-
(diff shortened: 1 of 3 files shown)


HonzaCholasta:
ipa-4-5:

* ebf24e783604952e59e557b5537c6d0de6146ce4 Get correct CA cert nickname in CA-less
* 99389748beb0158811505efa606c27e1e2e0bc7b Remove publish_ca_cert() method from NSSDatabase

PR #686, opened 2017-04-03
Stabilize make pypi_packages

tiran:
Parallel make or flags like IPA_OMIT_INSTALL and IPA_SERVER_WHEELS could
lead to bad packages for PyPI. Only build the packages we want with
correct flags.


MartinBasti:
master:

* d0c36b9c2eae1298604ec8ad4597e19e20365e11 Stabilize make pypi_packages

PR #688, opened 2017-04-03
Update get_attr_filter in LDAPSearch to handle nsaccountlock user searches

redhatrises:
- Update get_attr_filter in LDAPSearch to handle nsaccountlock by setting the default value for
nsaccountlock to false as well as update the filter to check for the default value

- https://pagure.io/freeipa/issue/6896


ipaserver/plugins/baseldap.py
```diff
@@ -1937,7 +1937,23 @@ def get_attr_filter(self, ldap, **options):
...
         search_kw['objectclass'] = self.obj.object_class
-        return ldap.make_filter(search_kw, rules=ldap.MATCH_ALL)
+        default_kw = self.get_default(**options)
+
+        filters = []
+        for name, value in search_kw.items():
+            if name == 'nsaccountlock' and value == False:
+                # If nsaccountlock is False, set the value to True and
+                # search for nsaccountlock != True. This way we can search
+                # for any users where nsaccountlock=False or nsaccountlock=None.
```
HonzaCholasta: No magic special cases please, the whole point of doing this in `baseldap` was that it would be generic. If the bit below doesn't work, we need to figure out how to make it work.

HonzaCholasta: OK, figured it out:
* The `nsaccountlock` param in `user` plugin does not have a default value specified.
* The check for default value below should actually look like this:
  ```python
              default = self.get_default_of(name, **options)
              if default is not None and value == default:
                  fltr = ldap.combine_filters([fltr, '(!({}=*))'.format(name)])
  ```

redhatrises:
@HonzaCholasta ready for your review.

redhatrises:
Bump for review

redhatrises:
Should this also go into the 4.5 branch?

HonzaCholasta:
I guess it should. Could you please file a ticket?

redhatrises:
@HonzaCholasta done. https://pagure.io/freeipa/issue/6896

HonzaCholasta:
Thanks!

HonzaCholasta:
Actually, please remove the change in `VERSION.m4`, as it is not necessary and prevents the patch from applying cleanly on top of ipa-4-5.

Also please add the ticket link to commit messages.

redhatrises:
@HonzaCholasta I removed the change to `VERSION.m4`

HonzaCholasta:
master:

* 38276d3473ecf2a4cc5b5e2a107347f046625626 Update get_attr_filter in LDAPSearch to handle nsaccountlock user searches


ipa-4-5:

* dc4d60c9665408666ab3dfab7023a578c34d65a2 Update get_attr_filter in LDAPSearch to handle nsaccountlock user searches

PR #690, opened 2017-04-04
server-install: remove broken no-pkinit check

stlaz:
Don't check for no-pkinit option in case pkinit cert file was
provided. Setting no-pkinit is prohibited in this case, so without
this fix we have an impossible option-check if we want to provide
an own pkinit certificate and private key.

https://pagure.io/freeipa/issue/6807


MartinBasti:
master:

* 1160dc5d8bacea42a7ada45a10bf1019a3af5aca server-install: remove broken no-pkinit check


ipa-4-5:

* 1eb681ec7d4f6f42e733463f29374f0fecee4e68 server-install: remove broken no-pkinit check

PR #691, opened 2017-04-05
Add force-join option to replica install

stlaz:
This patchset adds the force-join option to the replica installer. It also tries to improve the developer's experience by narrowing down the scope of originally an all-eating try-except block.

### DIFF
--- a/ipaserver/install/server/__init__.py
+++ b/ipaserver/install/server/__init__.py
@@ -166,7 +166,6 @@ class ServerInstallInterface(ServerCertificateInstallInterface,
     """
     description = "Server"
 
-    force_join = False
     kinit_attempts = 1
     fixed_primary = True
     ntp_servers = None
@@ -526,6 +525,7 @@ class ServerMasterInstall(ServerMasterInstallInterface):
     Server master installer
     """
 
+    force_join = False
     servers = None
     no_wait_for_dns = True
     host_password = None
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -908,52 +908,51 @@ def install_check(installer):
 
 
 def ensure_enrolled(installer):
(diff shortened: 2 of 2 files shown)


ipaserver/install/server/__init__.py
```diff
@@ -177,6 +176,9 @@ class ServerInstallInterface(ServerCertificateInstallInterface,
     preserve_sssd = False
     no_sssd = False
 
+    force_join = client.ClientInstallInterface.force_join
+    force_join = replica_install_only(force_join)
```
HonzaCholasta: This bit seems to be unnecessary, since you are masking `force_join` in `ServerMasterInstall` with a constant.

stlaz: Right. I meant to provide a more general approach but it probably does not make much sense here.

ipaserver/install/server/__init__.py
```diff
@@ -595,6 +595,8 @@ class ServerReplicaInstall(ServerReplicaInstallInterface):
     subject_base = None
     ca_subject = None
 
+    force_join = client.ClientInstallInterface.force_join
```
HonzaCholasta: I meant this bit as well, `force_join` is already present in `ServerReplicaInstall` through inheritance.

stlaz: Bah, I must be broken today, you're right, of course.

tomaskrizek:
master:

* 87051f51c695afa3e9ba072da7005cf1a4194668 Add the force-join option to replica install
* db84516d23d3a6e53e95881828d494fb80647602 replicainstall: better client install exception handling


ipa-4-5:

* 72f0ecde783be7d304044eff60c8c85e160d65d8 Add the force-join option to replica install
* 534df55ea5ae736db832e0885520a6dfbd09299a replicainstall: better client install exception handling

PR #692, opened 2017-04-05
tests: add non-reg for idrange-add

flo-renaud:
Add non regression test for issue 6404: when idrange-add is called with
empty dom-name, the command returns

https://pagure.io/freeipa/issue/6404


martbab:
master:

* 342dccea47f6cb14cda63f75789eab51070fb3f6 tests: add non-reg for idrange-add
ipa-4-5:

* ab2706721db217d55ae549d50a95ace571e65aa6 tests: add non-reg for idrange-add

PR #693, opened 2017-04-05
[tests] collect audit.log for easier selinux investigation

MartinBasti:
Audit log contains useful information about selinux issues


MartinBasti:
master:

* fd597f83aed53bf3281ce9ec6b94f601868cfc75 collect audit.log for easier selinux investigation

PR #694, opened 2017-04-05
RFC: implement local PKINIT deployment in server/replica install

martbab:
This PR implements a basic local PKINIT functionality for server install with
'--no-pkinit' specified, and replica install against older masters or with
'--no-pkinit'.

These patches unblock WebUI logins/password auths on masters/replicas in the
cases proper PKINIT was not configured for whatever reasons.

Nevertheless, there are following things lacking in this PR that I will either
push on top of this one or create a new PR:

- [x] removal of anonymous keytab, asi it is now useless (and always was)
- [x] upgrade and transitions between PKINIT configurations
- [x] reporting PKINIT state in LDAP
- [ ] API for querying the PKINIT status on all masters

http://www.freeipa.org/page/V4/Kerberos_PKINIT

### DIFF
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -38,6 +38,7 @@ from ipalib.constants import ANON_USER
 from ipalib.install import certmonger
 from ipapython.ipa_log_manager import root_logger
 from ipapython.dn import DN
+from ipapython.dogtag import KDC_PROFILE
 
 from ipaserver.install import replication
 from ipaserver.install import ldapupdate
@@ -354,61 +355,84 @@ class KrbInstance(service.Service):
             remote_ldap.gssapi_bind()
             replication.wait_for_entry(remote_ldap, kdc_dn, timeout=60)
 
-    def setup_pkinit(self):
-        if self.pkcs12_info:
-            certs.install_pem_from_p12(self.pkcs12_info[0],
-                                       self.pkcs12_info[1],
-                                       paths.KDC_CERT)
-            certs.install_key_from_p12(self.pkcs12_info[0],
-                                       self.pkcs12_info[1],
-                                       paths.KDC_KEY)
-        else:
-            subject = str(DN(('cn', self.fqdn), self.subject_base))
-            krbtgt = "krbtgt/" + self.realm + "@" + self.realm
-            certpath = (paths.KDC_CERT, paths.KDC_KEY)
-
-            try:
-                prev_helper = None
-                if self.master_fqdn is None:
-                    ca_args = [
-                        paths.CERTMONGER_DOGTAG_SUBMIT,
-                        '--ee-url', 'https://%s:8443/ca/ee/ca' % self.fqdn,
-                        '--certfile', paths.RA_AGENT_PEM,
-                        '--keyfile', paths.RA_AGENT_KEY,
-                        '--cafile', paths.IPA_CA_CRT,
-                        '--agent-submit'
-                    ]
-                    helper = " ".join(ca_args)
-                    prev_helper = certmonger.modify_ca_helper('IPA', helper)
-                else:
-                    self._wait_for_replica_kdc_entry()
-
-                certmonger.request_and_wait_for_cert(
-                    certpath,
-                    subject,
-                    krbtgt,
-                    dns=self.fqdn,
-                    storage='FILE',
-                    profile='KDCs_PKINIT_Certs')
-            except dbus.DBusException as e:
-                # if the certificate is already tracked, ignore the error
-                name = e.get_dbus_name()
-                if name != 'org.fedorahosted.certmonger.duplicate':
-                    root_logger.error("Failed to initiate the request: %s", e)
-                return
-            finally:
-                if prev_helper is not None:
-                    certmonger.modify_ca_helper('IPA', prev_helper)
-
-        # Finally copy the cacert in the krb directory so we don't
-        # have any selinux issues with the file context
-        shutil.copyfile(paths.IPA_CA_CRT, paths.CACERT_PEM)
+    def _call_certmonger(self, certmonger_ca='IPA'):
+        subject = str(DN(('cn', self.fqdn), self.subject_base))
+        krbtgt = "krbtgt/" + self.realm + "@" + self.realm
+        certpath = (paths.KDC_CERT, paths.KDC_KEY)
 
         try:
-            self.restart()
-        except Exception:
-            root_logger.critical("krb5kdc service failed to restart")
-            raise
+            prev_helper = None
+            # on the first CA-ful master without '--no-pkinit', we issue the
+            # certificate by contacting Dogtag directly
+            use_dogtag_submit = all(
+                [self.master_fqdn is None,
+                 self.pkcs12_info is None,
+                 self.config_pkinit])
+
+            if use_dogtag_submit:
+                ca_args = [
+                    paths.CERTMONGER_DOGTAG_SUBMIT,
+                    '--ee-url', 'https://%s:8443/ca/ee/ca' % self.fqdn,
+                    '--certfile', paths.RA_AGENT_PEM,
+                    '--keyfile', paths.RA_AGENT_KEY,
+                    '--cafile', paths.IPA_CA_CRT,
+                    '--agent-submit'
+                ]
+                helper = " ".join(ca_args)
+                prev_helper = certmonger.modify_ca_helper(certmonger_ca, helper)
+
+            certmonger.request_and_wait_for_cert(
+                certpath,
+                subject,
+                krbtgt,
+                ca=certmonger_ca,
+                dns=self.fqdn,
+                storage='FILE',
+                profile=KDC_PROFILE)
+        except dbus.DBusException as e:
+            # if the certificate is already tracked, ignore the error
+            name = e.get_dbus_name()
+            if name != 'org.fedorahosted.certmonger.duplicate':
+                root_logger.error("Failed to initiate the request: %s", e)
+            return
+        finally:
+            if prev_helper is not None:
+                certmonger.modify_ca_helper(certmonger_ca, prev_helper)
+
+    def issue_selfsigned_pkinit_certs(self):
+        self._call_certmonger(certmonger_ca="SelfSign")
+        # for self-signed certificate, the certificate is its own CA, copy it
+        # as CA cert
+        shutil.copyfile(paths.KDC_CERT, paths.CACERT_PEM)
+
+    def issue_ipa_ca_signed_pkinit_certs(self):
+        try:
+            self._call_certmonger()
+            # copy IPA CA bundle to the KDC's CA cert bundle
+            shutil.copyfile(paths.IPA_CA_CRT, paths.CACERT_PEM)
+        except RuntimeError as e:
+            root_logger.error("PKINIT certificate request failed: %s", e)
+            root_logger.error("Failed to configure PKINIT")
+            self.stop_tracking_certs()
+            self.issue_selfsigned_pkinit_certs()
+
+    def install_external_pkinit_certs(self):
+        certs.install_pem_from_p12(self.pkcs12_info[0],
+                                   self.pkcs12_info[1],
+                                   paths.KDC_CERT)
+        certs.install_key_from_p12(self.pkcs12_info[0],
+                                   self.pkcs12_info[1],
+                                   paths.KDC_KEY)
+        # copy IPA CA bundle to the KDC's CA cert bundle
+        # NOTE: this may not be the same set of CA certificates trusted by
+        # externally provided PKINIT cert.
+        shutil.copyfile(paths.IPA_CA_CRT, paths.CACERT_PEM)
+
+    def setup_pkinit(self):
+        if self.pkcs12_info:
+            self.install_external_pkinit_certs()
+        elif self.config_pkinit:
+            self.issue_ipa_ca_signed_pkinit_certs()
 
     def test_anonymous_pkinit(self):
         with ipautil.private_ccache() as anon_ccache:
@@ -418,6 +442,15 @@ class KrbInstance(service.Service):
                 raise RuntimeError("Failed to configure anonymous PKINIT")
 
     def enable_ssl(self):
+        """
+        generate PKINIT certificate for KDC. If `--no-pkinit` was specified,
+        only configure local self-signed KDC certificate for use as a FAST
+        channel generator for WebUI. Do not advertise the installation steps in
+        this case.
+        """
+        if self.master_fqdn is not None:
+            self._wait_for_replica_kdc_entry()
+
         if self.config_pkinit:
             self.steps = []
             self.step("installing X509 Certificate for PKINIT",
@@ -425,6 +458,14 @@ class KrbInstance(service.Service):
             self.step("testing anonymous PKINIT", self.test_anonymous_pkinit)
 
             self.start_creation()
+        else:
+            self.issue_selfsigned_pkinit_certs()
+
+        try:
+            self.restart()
+        except Exception:
+            root_logger.critical("krb5kdc service failed to restart")
+            raise
 
     def get_anonymous_principal_name(self):
         return "%s@%s" % (ANON_USER, self.realm)
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -48,6 +48,38 @@ from ipaplatform.constants import constants
 from ipaplatform.tasks import tasks
 from ipaplatform.paths import paths
 
+PKINIT_ENABLED = 'pkinitEnabled'
+
+
+def get_pkinit_request_ca():
+    """
+    Return the certmonger CA name which is serving the PKINIT certificate
+    request. If the certificate is not tracked by Certmonger, return None
+    """
+    pkinit_request_id = certmonger.get_request_id(
+        {'cert-file': paths.KDC_CERT})
+
+    if pkinit_request_id is None:
+        return
+
+    return certmonger.get_request_value(pkinit_request_id, 'ca-name')
+
+
+def is_pkinit_enabled():
+    """
+    check whether PKINIT is enabled on the master by checking for the presence
+    of KDC certificate and it's tracking CA
+    """
+
+    if os.path.exists(paths.KDC_CERT):
+        pkinit_request_ca = get_pkinit_request_ca()
+
+        if pkinit_request_ca != "SelfSign":
+            return True
+
+    return False
+
+
 class KpasswdInstance(service.SimpleServiceInstance):
     def __init__(self):
         service.SimpleServiceInstance.__init__(self, "kadmin")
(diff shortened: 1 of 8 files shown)


install/certmonger/ipa-pkinit-submit
```diff
@@ -0,0 +1,129 @@
...
+    syslog.syslog(syslog.LOG_NOTICE, "Forwarding request to {}".format(helper))
+
+    args = [paths.IPA_SUBMIT]
+    args.extend(sys.argv[1:])
+
+    result = ipautil.run(args, raiseonerr=False, env=os.environ,
+                         capture_output=True)
+    rc = result.returncode
+
+    if six.PY2:
```
MartinBasti: Why do you need `raw` output here?

install/certmonger/ipa-pkinit-submit
```diff
@@ -0,0 +1,129 @@
...
+        handler = request_pkinit_cert
+    else:
+        syslog.syslog(
+            syslog.LOG_ERR,
+            'Unrecognized pkinit state: {}'.format(pkinit_state))
+        return UNCONFIGURED
+
+    return handler()
+
+
```
MartinBasti: `if __name__ == '__main__':`  missing here?

install/certmonger/ipa-pkinit-submit
```diff
@@ -0,0 +1,129 @@
...
+try:
+    sys.exit(main())
+except Exception as e:
```
MartinBasti: `e` is unused

ipaserver/install/krbinstance.py
```diff
@@ -372,6 +373,12 @@ def _wait_for_replica_kdc_entry(self):
             replication.wait_for_entry(remote_ldap, kdc_dn, timeout=60)
 
     def setup_pkinit(self):
+        try:
+            root_logger.debug("Adding '%s' CA to certmonger", PKINIT_CA_NAME)
+            certmonger.add_ca(PKINIT_CA_NAME, paths.IPA_PKINIT_SUBMIT)
+        except ValueError:
```
MartinBasti: This looks weird, a new exception should be used, like `CAAlreadyConfiguredError`

ipaserver/install/krbinstance.py
```diff
@@ -449,6 +484,9 @@ def enable_ssl(self):
             self.step("testing anonymous PKINIT", self.test_anonymous_pkinit)
 
             self.start_creation()
+        else:
```
MartinBasti: This is duplication of `if` branch, can we have it without `if else` statement

MartinBasti:
`upgrade and transitions between PKINIT configurations` does this cover:

- CA-less to CA-full upgrade?
- installed 4.4.4 --- upgraded ---> 4.5.0 --- upgraded ----> 4.5.1
- installed 4.5.0 --- upgraded ---> 4.5.1

?

martbab:
@MartinBasti I haven't thought about CA-less -> CA-full but in this case you would have local PKINIT and should configure full PKINIT manually

All the other scenarios should be covered by the incoming code.

Regarding your comment on the certmonger helper/special CA, we (me and @HonzaCholasta ) decided to remove it and use a self-sign CA instead.

martbab:
I have re-worked the PR and implemented most of the missing steps (except for API for querying PKINIT status in topology). I have also removed the PKINIT-specific CA and helper. The installer will now call either `IPA` or self-sign CA depending on configuration and passed-in options. The PKINIT state recording was also changed to depend on the KDC certificate tracking status and CA that tracks it.

ipaserver/install/krbinstance.py
```diff
@@ -354,77 +435,111 @@ def _wait_for_replica_kdc_entry(self):
...
 
     def enable_ssl(self):
         if self.config_pkinit:
             self.steps = []
             self.step("installing X509 Certificate for PKINIT",
                       self.setup_pkinit)
-            self.step("testing anonymous PKINIT", self.test_anonymous_pkinit)
-
             self.start_creation()
+        else:
```
MartinBasti: why we need `if else` braches here if the code is almost the same

ipaserver/install/server/replicainstall.py
```diff
@@ -1525,6 +1525,10 @@ def install(installer):
         # remove the extracted replica file
         remove_replica_info_dir(installer)
 
+    if not options.no_pkinit:
```
MartinBasti: This looks weird, you always test it but only sometimes show text

MartinBasti:
Should be anon keytab removed by upgrade, are there any leftovers in LDAP to be removed during upgrade?

ipaserver/install/krbinstance.py
martbab: That's a question of UX we may discuss further. by default we inform the user that PKINIT certificate is being requested and installed. When he passes in `--no-pkinit` we pretend to not configure it and only use the local self-sign path. I thought that in this  case we shouldn't inform the user about this, otherwise we will have plenty complaints about an installer setting up undesired feature even when instructed not to.

ipaserver/install/server/replicainstall.py
martbab: The same point as above, we may discuss it separately.

ipaserver/install/krbinstance.py
MartinBasti: ok

martbab:
@MartinBasti  I can add some removal logic to upgrader if required.

ipaserver/install/krbinstance.py
```diff
@@ -47,6 +49,85 @@
...
+    else:
+        raise ValueError(
+            "Unrecognized certmonger CA: {}".format(pkinit_request_ca))
+
+
+def record_pkinit_state_ldap(_api):
+    """
+    record PKINIT state in the master's KDC entry in LDAP. The state is
+    recorded in ipaConfigString attribute as a 'pkinit_status:STATUS' where
+    STATUS is one of (local, external, full).
```
HonzaCholasta: Why not just `pkinit_enabled`? There is no difference between external and full for clients.

ipaserver/install/server/install.py
```diff
@@ -892,6 +892,10 @@ def install(installer):
     # Make sure the files we crated in /var/run are recreated at startup
     tasks.configure_tmpfiles()
 
+    if not options.no_pkinit:
+        print("Testing anonymous PKINIT")
+    krb.test_anonymous_pkinit()
```
HonzaCholasta: Why do we need to test anything? Are we not sure what are we doing?

ipaserver/install/krbinstance.py
HonzaCholasta: Ditto everywhere else, IMO the local-external-full tri-state is confusing, "local" is an implementation detail and does not tell the user anything useful, and "external" and "full" is just normal CA-less vs CA-full described in odd words.

ipaserver/install/server/upgrade.py
```diff
@@ -1485,14 +1485,20 @@ def add_default_caacl(ca):
...
+    if pkinit_state not in (krbinstance.PKINIT_FULL,
+                            krbinstance.PKINIT_EXTERNAL):
 
-    if not os.path.exists(paths.KDC_CERT):
-        root_logger.info("Requesting PKINIT certificate")
-        krb.setup_pkinit()
+        pkinit_setup_handler = krb.setup_local_pkinit
 
+        if api.Command.ca_is_enabled()['result']:
+            pkinit_setup_handler = krb.setup_full_pkinit
```
HonzaCholasta: If full / external pkinit was not enabled before, we should not enable it during upgrade, otherwise `--no-pkinit` will work only until the first upgrade.

ipaserver/install/krbinstance.py
HonzaCholasta: A comment in the code would be nice.

ipaserver/install/server/upgrade.py
martbab: that was agreed on with @abbra and @simo5 during designing the missing parts (see http://www.freeipa.org/page/V4/Kerberos_PKINIT). If you disagree we should discuss with them.

ipaserver/install/server/install.py
martbab: Testing PKINIT was added so that we fail early if the configuration is wrong. Otherwise you just get hard to debug issues such as login to WebUI failing suddenly. I can remove this if we are confident that the current code is robust enough.

ipaserver/install/krbinstance.py
martbab: This was agreed on during design discussion http://www.freeipa.org/page/V4/Kerberos_PKINIT @simo5  wanted to have clear distinction between individual PKINIT configurations.

ipaserver/install/server/upgrade.py
simo5: When we designed this, the idea was that on --no-pkinit we would create an "external" self-signed certificate. This would be like local, but would prevent upgrades because it is not tracked by a CA certmonger can request certs from.

If "external" can't be used, then you have to track somewhere that the admin explicitly passed the --no-pkinit flag and prevent upgrades.

In general we *do* want autometic upgrades unless the admin explicitly and unequivocably set a --no-pkinit flag.

The other option is to completely ignore --no-pkinit and mark those flags as deprecated and document that, if an dmin doesn't want pkinit at all they will have to replace the pkinit certs with certs of their own that work only locally.

ipaserver/install/krbinstance.py
martbab: ok I will add the comment.

ipaserver/install/krbinstance.py
martbab: @simo5 Although when I think about it, the external/full PKINIT distinction is indistinguishable from the client's POV so we can actually have binary disabled (local)/ enabled state without any harm even during upgrades, both external and full pkinit states mean that no further upgrade is required.

martbab:
I have added comments to the "hidden" PKINIT setup steps. I would also like @simo5 and @abbra to answer further comments regarding thid PR. It was on ice for nearly two weeks and we should really have it merged soon.

abbra:
I read through the code and I believe it addresses all use cases we have been discussing. LGTM.

martbab:
@abbra I received an interactive review from @HonzaCholasta today and he is not very keen on idea of having ternary (absent/local/external/full) PKINIT configuration. He suggests to only have it absent/off (local implementation)/on and thus drop differentiation between PKINIT configured with IPA CA issued or 3rd party certificates. The main concern here is that the 'local' PKINIT configuration is actually an implementation detail we should not leak to clients, they should be only able to tell if it is configured for them or not.

If you look into the design page, the two states (full/external) behave the same during replica installation and upgrade so the differentiation does not bring much new information to the users. So a simple on/off switch (something like pkinitStatus: off/on) could be enough and it could simplify the transition and UX. What do you think?

abbra:
I agree that it is internal detail whether we use local pkinit or not. However, we need to know that it is existing as oposed to not existing at all for older systems where we are going to perform upgrades. However, as you can derive this information by presence or lack of actual KDC certificate file in the file system during upgrade, this can be reduced, indeed.

One more detail: we already have pkinit plugin (`ipaserver/plugins/pkinit.py`) which has `ipa pkinit-anonymous enable/disable` command. This command cannot now be used because even for 'local' case we require anonymous PKINIT to be usable and this means we cannot disable the principal.

Perhaps, you can remove this command and add instead `ipa pkinit-status` command to show the status? It would show list of KDCs and their status.

martbab:
We can query that PKINIT was not configured at all by a) checking the presence of KDC keypair, b) checking the sysupgrade (no presence of pkinit flag implies no configuration is present), and c) querying LDAP (no presence of ipaConfigString) so we have multiple redundant ways to determine that PKINIT is not configured at all.

As for the removal of pkinit status, I intend to replace the existing command by `ipa pkinit-status` as a follow-up PR once this one is merged.

I will then update the design page to reflect this discussion and update the implementation in this PR.

abbra:
Yep. Then this PR can be merged once you removed distinction external/full.

ipaserver/install/krbinstance.py
```diff
@@ -47,6 +49,86 @@
...
+
+    if not os.path.exists(paths.KDC_CERT):
+        return
+
+    pkinit_request_ca = get_pkinit_request_ca()
+
+    if pkinit_request_ca is None or pkinit_request_ca == "IPA":
+        return PKINIT_ENABLED
+    elif pkinit_request_ca == "SelfSign":
+        return PKINIT_DISABLED
```
HonzaCholasta: Whether PKINIT is enabled or not is not a tri-state, it is a boolean. You are basically still exposing internal implementation detail to the user by making it a tri-state.

ipaserver/install/krbinstance.py
```diff
@@ -354,77 +436,116 @@ def _wait_for_replica_kdc_entry(self):
...
+            return
+        finally:
+            if prev_helper is not None:
+                certmonger.modify_ca_helper(certmonger_ca, prev_helper)
+
+    def issue_selfsigned_pkinit_certs(self):
+        self._call_certmonger(certmonger_ca="SelfSign")
+        # for self-signed certificate, the certificate is its own CA, copy it
+        # as CA cert
+        shutil.copyfile(paths.KDC_CERT, paths.CACERT_PEM)
```
HonzaCholasta: NACK, please explicitly use `paths.KDC_CERT` as an achor in `login_password`.

ipaserver/install/krbinstance.py
```diff
@@ -354,77 +436,116 @@ def _wait_for_replica_kdc_entry(self):
...
+
+    def issue_ipa_ca_signed_pkinit_certs(self):
+        try:
+            self._call_certmonger()
+            shutil.copyfile(paths.IPA_CA_CRT, paths.CACERT_PEM)
```
HonzaCholasta: `paths.IPA_CA_CRT` contains a set of CA certificates trusted for SSL servers, which may, but don't have to be the same as the set of CA certificates trusted for PKINIT. You don't have to fix this now, but at least a comment would be nice.

HonzaCholasta:
@martbab, this sounds like a typical instance of a we will do it later = we will do it never situation. IMO we should remove the superfluous check right away, as that would give us more incentive to actually implement the test.

ipaserver/install/krbinstance.py
martbab: @HonzaCholasta true, the selfsign/absent PKINIT information is actually required only locally during upgrade (that can be inferred from the presence of KDC certificate and its managing CA), server/replica install make decisions only based on provided options and state of topology.

It can thus be simplified further by reporting only 'pkinitEnabled' in LDAP and handling selfsign/absent only locally during upgrade. @simo5 is that ok with you?

ipaserver/install/krbinstance.py
martbab: Then I will need to conditionally select between KDC_CERT and CACERT_PEM based on the PKINIT status, since the KDC_CERT is not a vaild anchor in full PKINIT scenario. This means that I will need to perform extra D-BUS call to certmonger for each login attempt but I think that won't be a huge issue from performance POV.

ipaserver/install/krbinstance.py
martbab: OK I will add a comment here. Given this was the original behavior a fix will warant a separate PR.

martbab:
@HonzaCholasta Then  the best course of action is to remove the PKINIT check and raise the priority of the issue for test case.

ipaserver/install/krbinstance.py
simo5: If it still conforms to the cases we identified with the table in the design document it is fine by me.

martbab: @simo5 Yes it does, this information will just not be exposed to clients as it is our internal implementation detail.

simo5: I am in favor of not over-exposing internal sausage-making.

ipaserver/install/krbinstance.py
martbab: @HonzaCholasta and to determine PKINIT status we need to either:

- check PKINIT status in LDAP, which we can not do from login_password because we do not have a working LDAP connection (as I found out now)
-  Query certmonger's D-Bus interface about the tracking information of KDC certificate which ipaapi user is denied by the current DBus policy

So I am open to suggestions how to implement this in rpcserver layer.

martbab: Oh wait, with your proposal I can just check for present of PKINIT's CA bundle and not do any complex magicks. I will look into that.

martbab: After some offline discussion with @HonzaCholasta we concluded that there currently is no easy way to select KDC cert or IPA CA bundle base don PKINIT status. So we will roll with the original idea to copy KDC cert to IPA CA bundle to 'emulate' the correct anchor for now. We will think about a proper handling of this issue later.

ipaserver/install/krbinstance.py
```diff
@@ -47,6 +48,38 @@
...
+def is_pkinit_enabled():
+    """
+    check whether PKINIT is enabled on the master by checking for the presence
+    of KDC certificate and it's tracking CA
+    """
+
+    if os.path.exists(paths.KDC_CERT):
+        pkinit_request_ca = get_pkinit_request_ca()
+
+        if pkinit_request_ca is None or pkinit_request_ca == "IPA":
```
HonzaCholasta: I would maybe go even further and say `if pkinit_request_ca != 'SelfSign'`, I guess we should not care which CA is used as long as it isn't SelfSign.

HonzaCholasta:
LGTM.

martbab:
Any volunteer to do a functional review?

HonzaCholasta:
Works for me, ACK.

HonzaCholasta:
master:

* b1a1e104391c84cb9af7b0a7c8748c8652442ddb separate function to set ipaConfigString values on service entry
* fb52f7a1f328b126626525179d5250692daca2cd Allow for configuration of all three PKINIT variants when deploying KDC
* 86972299d937960bcb713fc73b447cddb4ea44bd API for retrieval of master's PKINIT status and publishing it in LDAP
* 3adb9ca875f8eb99e99a29e17a471a2b6f408a4a Use only anonymous PKINIT to fetch armor ccache
* 68c6a4d4e1340ce01bdc7ec5dd394604a3da7688 Stop requesting anonymous keytab and purge all references of it
* 2374b648d0dfd08ec4cfbcc35f7987fa8b8a6ffa Use local anchor when armoring password requests
* a194055c92c7ca4eba29323f990ec3b92026221b Upgrade: configure local/full PKINIT depending on the master status
* 960e361f68a3d7acd9bcf16ec6fe8f6d5376c4ae Do not test anonymous PKINIT after install/upgrade


ipa-4-5:

* 31a24436592304db6e84270e4a95df34d1e0af46 separate function to set ipaConfigString values on service entry
* b49e075c90a7ab43e82f422aa11dc7540e2fb2c0 Allow for configuration of all three PKINIT variants when deploying KDC
* a0e2a09292ffa2adbf97c2e7e4facc9693dbc311 API for retrieval of master's PKINIT status and publishing it in LDAP
* fca378c9a65f582ac3dcda4b6201e8847ed9e512 Use only anonymous PKINIT to fetch armor ccache
* 9fcc794dac6ffb1f1cc6c92a588ea0911be5ba14 Stop requesting anonymous keytab and purge all references of it
* 5031929b6d710336f6308d7f46779c9e8e98103a Use local anchor when armoring password requests
* 2452e6e5f3a7e7a25eadf5243a28da75a47f9d2c Upgrade: configure local/full PKINIT depending on the master status
* d497c4589cc7506ef9a88b691b8b1d97ad1f1009 Do not test anonymous PKINIT after install/upgrade

PR #701, opened 2017-04-07
ipa help doesn't always work

neffs:
`ipa help` will not work when calling it when no schema is cached.

### DIFF
--- a/ipaclient/remote_plugins/schema.py
+++ b/ipaclient/remote_plugins/schema.py
@@ -516,7 +516,9 @@ class Schema(object):
 
     def get_help(self, namespace, member):
         if isinstance(self._help, bytes):
-            self._help = json.loads(self._help.decode('utf-8'))
+            self._help = json.loads(
+                self._help.decode('utf-8')  # pylint: disable=no-member
+            )
 
         return self._help[namespace][member]
 
--- a/ipaclient/remote_plugins/schema.py
+++ b/ipaclient/remote_plugins/schema.py
@@ -383,6 +383,7 @@ class Schema(object):
 
         if fingerprint is None:
             fingerprint, ttl = self._fetch(client, ignore_cache=read_failed)
+            self._help = self._generate_help(self._dict)
             try:
                 self._write_schema(fingerprint)
             except Exception as e:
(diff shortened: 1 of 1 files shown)


martbab:
@neffs please fix pylint error reported in Travis CI:

```console
************* Module ipaclient.remote_plugins.schema

ipaclient/remote_plugins/schema.py:519: [E1101(no-member), Schema.get_help] Instance of 'dict' has no 'decode' member)

make: *** [pylint] Error 2

Makefile:1175: recipe for target 'pylint' failed
```

neffs:
Not sure why pylint is reporting this, it looks fine to me and I didn't change this function.
```python
        if isinstance(self._help, bytes):
            self._help = json.loads(self._help.decode('utf-8'))
```

maybe we can trick it with a change like this (similar to read_namespace_member):
```python
        value = self._help
        if isinstance(value, bytes):
            self._help = json.loads(value.decode('utf-8'))
```

neffs:
@martbab I pushed this additional change but pylint still reports this error which is clearly wrong because the line is only executed if it is an instance of bytes. Could you please provide some directions?
I only have the code on my mac and unfortunately can't run the tests locally.

dkupka:
@neffs Hi, thanks for the patch. As for the pylint issue it happens sometimes that pylint gets confused even in situations that seems perfectly clear. This will be probably the case because it complains about member of dict instance in branch when it's guaranteed to be bytes instance.

I propose following "fix". Just tell pylint to don't worry about this case.
``` python
 if isinstance(self._help, bytes):
            self._help = json.loads(
                self._help.decode('utf-8')  # pylint: disable=no-member
            )
```

dkupka:
@neffs Thanks for the fix. Could you please drop the second commit ("Fix pylint error in get_help function"), rebase and force-push? The commit gets completely reverted in the third one so it has no value.

dkupka:
Works for me, thanks. 

(Sorry for the close/reopen, it's too early in the morning here :-)

martbab:
master:

* d5bb541061e6c0952d2075a24d0a58c87455f233 Store help in Schema before writing to disk
* bf0ba9b36e95f2e2b14bb27059280027d8354c13 Disable pylint in get_help function because of type confusion.

PR #704, opened 2017-04-10
WebUI: cert login: Configure name of parameter used to pass username

dkupka:
Directive LookupUserByCertificateParamName tells mod_lookup_identity module the
name of GET parameter that is used to provide username in case certificate is
mapped to multiple user accounts.
Without this directive login with certificate that's mapped to multiple users
doesn't work.

https://pagure.io/freeipa/issue/6860

### DIFF
--- a/install/conf/ipa.conf
+++ b/install/conf/ipa.conf
@@ -117,6 +117,7 @@ Alias /ipa/session/cookie "/usr/share/ipa/gssapi.login"
   NSSVerifyClient require
   NSSUserName SSL_CLIENT_CERT
   LookupUserByCertificate On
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @dkupka 
thank you for the patch, everything works as expected with a single certificate mapped to multiple users.

pvomacka:
ipa-4-5:

* a9721e529e7a02eeb40d29cb7820e69cd86d9337 WebUI: cert login: Configure name of parameter used to pass username
master:

* 157831a287c64106eed4da4ace5228d7e369ae2f WebUI: cert login: Configure name of parameter used to pass username

PR #706, opened 2017-04-11
Fix CA-less to CA-full upgrade

stlaz:
CertDB would have always created a directory on initialization. This
behavior changes here by replacing the truncate argument with create
which will only create the database when really required.

https://pagure.io/freeipa/issue/6853

### DIFF
--- a/ipaserver/install/ca.py
+++ b/ipaserver/install/ca.py
@@ -183,6 +183,8 @@ def install_check(standalone, replica_config, options):
             realm_name, nssdir=dirname, subject_base=options._subject_base)
 
         for db in (cadb, dsdb):
+            if not db.exists():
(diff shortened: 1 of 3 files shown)


stlaz:
I am not a big fan of creating directories on object initialization, thus this change. Another way of doing this would be of course to check whether the pki-tomcat/alias dir exists in the first place and not create the CertDB there.

tomaskrizek:
master:

* 9ac56e47d78582fbc2911f67a7344bcce321842f Fix CA-less to CA-full upgrade


ipa-4-5:

* 7a6f78bab8f9f76bf37fb105ec2537676d889cc2 Fix CA-less to CA-full upgrade

PR #709, opened 2017-04-11
Fix s4u2self with adtrust

simo5:
When ADtrust is installed we add a PAC to all tickets, during protocol
transition we need to generate a new PAC for the requested user ticket,
not check the existing PAC on the requestor ticket.

https://pagure.io/freeipa/issue/6862

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb_mspac.c
+++ b/daemons/ipa-kdb/ipa_kdb_mspac.c
@@ -2117,6 +2117,7 @@ krb5_error_code ipadb_sign_authdata(krb5_context context,
     struct ipadb_context *ipactx;
     bool with_pac;
     bool with_pad;
+    bool make_ad = false;
     int result;
     krb5_db_entry *client_entry = NULL;
     krb5_boolean is_equal;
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @simo5,

I tested webUI authentication with a IPA user and it is working with this patch.

daemons/ipa-kdb/ipa_kdb_mspac.c
```diff
@@ -2165,7 +2166,14 @@ krb5_error_code ipadb_sign_authdata(krb5_context context,
                                   "currently not supported.");
     }
 
-    if (is_as_req && with_pac && (flags & KRB5_KDB_FLAG_INCLUDE_PAC)) {
+    /* we need to create a PAC if we are requested one and this is an AS REQ,
+     * or we are doing protocol transition (s4u2self) */
+    if ((is_as_req && (flags & KRB5_KDB_FLAG_INCLUDE_PAC)) ||
+        (flags & KRB5_KDB_FLAG_PROTOCOL_TRANSITION)) {
```
abbra: I think if should be `(flags & KRB5_KDB_FLAGS_S4U)` because both protocol transition and constraint delegation are in `KRB5_KDB_FLAGS_S4U` and `fetch_kdb_authdata()` does check against this constant when looking if it is included.

simo5: No this would force a rebuild of the PAC in s4u2proxy cases.
In that case we check and forward port the PAC *or* create it. And this case is already handled in pre-existing code.

abbra: Ok, the other case (constraint delegation) of `KRB5_KDB_FLAGS_S4U` flag is handled already.

pvomacka:
ipa-4-5:

* b5114070ae55bcc7ec1abe57b4c303cee4822930 Fix s4u2self with adtrust
master:

* e88d5e815ea440bcef4acdc5f8fcb3a29e6eaec9 Fix s4u2self with adtrust

PR #713, opened 2017-04-13
WebUI: fix showing required asterisk '*'

pvomacka:
There was a bug that when user switch between two facets where is
required field and in one of them is writable and in second one
is not writable, then the asterisk which marks required field is
not shown. i.e. admin vs. user details page or global_passwd_policy
vs. other_passwd_policy details page.

That was caused by incorrect evaluation of required state of field.
Evaluation works that way: evaluate old required state, then evaluate
current required state and if states has changed then emit change event.
The evaluation depends on writable and read_only state of field.
Those two states are set before evaluation of required state, but
their old values (for evaluating previous required stated) were
not stored anywhere.

This commit adds two attributes which stores old writable
and read_only states. The required asterisk is then shown correctly.

https://pagure.io/freeipa/issue/6849

### DIFF
--- a/install/ui/src/freeipa/field.js
+++ b/install/ui/src/freeipa/field.js
@@ -221,11 +221,25 @@ field.field = IPA.field = function(spec) {
     that.read_only = spec.read_only;
 
     /**
+     * Attribute for storing previous value of read_only attribute.
+     * It is set during changing read_only attribute.
+     * @property {boolean}
+     */
+    that.old_read_only = spec.read_only;
+
+    /**
      * Writable is set during load
      * @readonly
      * @property {boolean}
      */
     that.writable = true;
+
+    /**
+     * Attribute for storing previous value of writable attribute.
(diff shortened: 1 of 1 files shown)


pvoborni:
I was thinking about signature of `is_required(old)` method. It is not very clear what it does from reading the code if we forget context of this patch, especially if we see: `var old = that.is_required(true)`. There is probably similar thing in the code already implemented by me. But I was thinking if it should be rather changed into two methods: `is_required()` and `was_required`. Then the `readonly` or `writable` will use calls like `var old = that.was_required();` and ` var required = that.is_required();`.

pvomacka:
@pvoborni the change you are proposing makes sense, I changed the code that way. Thank you

tomaskrizek:
master:

* d5ef1a7fd0863b52d91fc44f0e9dbdb9d76a7bb1 WebUI: fix showing required asterisk '*'

PR #714, opened 2017-04-15
fix minor typo in ipa-adtrust-install.1

realsobek:
ipa-4-5 and ipa-4-4 are affected too. I ignore branches that are not listed at http://www.freeipa.org/page/Roadmap#Current_Efforts
As far as I can see I cannot put the changes in here. Do I have to create a pull request for each one?

PR #715, opened 2017-04-15
use correct option name

realsobek:


install/tools/man/ipa-adtrust-install.1
```diff
@@ -26,7 +26,7 @@ Adds all necessary objects and configuration to allow an IPA server to create a
 trust to an Active Directory domain. This requires that the IPA server is
 already installed and configured.
 
-Please note you will not be able to estabilish an trust to an Active Directory
+Please note you will not be able to establish an trust to an Active Directory
```
stlaz: This change is done in a different PR already: https://github.com/freeipa/freeipa/pull/714

po/uk.po
```diff
@@ -7998,7 +7998,9 @@ msgstr "Словник перекладених повідомлень"
 msgid ""
 "\n"
 "ID Views\n"
+"\n"
```
stlaz: The messages to be translated get autogenerated using the common [gettext](https://docs.python.org/2/library/gettext.html) practice, therefore it does not make sense changing it here but should be changed in the code if there's a need for it.

stlaz:
Since the changes here are part of https://github.com/freeipa/freeipa/pull/716, I am going to close this PR.

PR #717, opened 2017-04-16
csrgen: Finish NSS support

LiptonB:
I took the approach of generating a new key for each request, as keys already stored in a database are difficult to name precisely. I also had to add another hook to `CSRLibraryAdaptor` that is called after the cert is returned from the server, so that we could add the cert to the database as desired.


rcritten:
Closing due to inactivity.

PR #719, opened 2017-04-18
External CA fixes

stlaz:
External CA installation would have failed for 2 reasons:
- Trying to perform Kerberos install twice
- Rewriting the CA cert file with each consecutive certificate in the certificate chain instead of appending them

This patchset fixes that behavior.

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -762,11 +762,12 @@ def install(installer):
             options.subject_base, options.ca_subject, 1101, 1100, None)
 
     krb = krbinstance.KrbInstance(fstore)
-    krb.create_instance(realm_name, host_name, domain_name,
-                        dm_password, master_password,
-                        setup_pkinit=not options.no_pkinit,
-                        pkcs12_info=pkinit_pkcs12_info,
-                        subject_base=options.subject_base)
+    if not options.external_cert_files:
+        krb.create_instance(realm_name, host_name, domain_name,
+                            dm_password, master_password,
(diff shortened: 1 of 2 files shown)


ipaserver/install/server/install.py
```diff
@@ -761,12 +761,15 @@ def install(installer):
...
-                        subject_base=options.subject_base)
+    if not options.external_cert_files:
+        krb = krbinstance.KrbInstance(fstore)
+        krb.create_instance(realm_name, host_name, domain_name,
+                            dm_password, master_password,
+                            setup_pkinit=not options.no_pkinit,
+                            pkcs12_info=pkinit_pkcs12_info,
+                            subject_base=options.subject_base)
+    else:
+        krb = krbinstance.KrbInstance(fstore)
```
HonzaCholasta: This exact line is in both branches of the if statement. I think it would be better to move it before the if statement.

stlaz: You're right, this is silly, thanks.

HonzaCholasta:
master:

* 25a33ce8b1c77b0d957772143affd7085757bccb server-install: No double Kerberos install
* 7b8503173b253860c1059bd40858f2fdffb4ae33 ext. CA: correctly write the cert chain


ipa-4-5:

* 2144eaf25ef1148c9353dfb2680f8811fd8c21aa server-install: No double Kerberos install
* a6af0033a4d0af387eebdd6500eb1e74c5c29ce7 ext. CA: correctly write the cert chain

PR #723, opened 2017-04-20
Store GSSAPI session key in /var/run/httpd

MartinBasti:
Runtime data should be stored in /var/run instead of /etc/httpd/alias.
This change is also compatible with selinux policy.

### DIFF
--- a/install/conf/ipa.conf
+++ b/install/conf/ipa.conf
@@ -1,5 +1,5 @@
 #
-# VERSION 25 - DO NOT REMOVE THIS LINE
+# VERSION 26 - DO NOT REMOVE THIS LINE
 #
 # This file may be overwritten on upgrades.
 #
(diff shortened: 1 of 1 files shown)


HonzaCholasta:
Could we put the mod_auth_gssapi session key in `/var/run/ipa/session.key`? `/var/run/ipa` is where we store IPA-specific stuff, including mod_auth_gssapi ccaches.

MartinBasti:
Sure

tomaskrizek:
Functional ACK. There was a concern in the ticket's discussion about reboots - are we going to handle them?

MartinBasti:
This approach was agreed on devel meeting

martbab:
master:

* 2bab2d4963daa99742875f3633a99966bc56f5a3 Store GSSAPI session key in /var/run/ipa
ipa-4-5:

* b2aa3ed0bc9f5385ab6e8b1720d9f1d33136e5dc Store GSSAPI session key in /var/run/ipa

simo5:
This patch is wrong please revert

simo5:
As I noted in the ticket: "At most you may want to store it in /var/lib/ipa/somewhere, but we do not want to break sessions (there are people using APIs from non-interactive scripts) just because you needed to restart a service/server quickly.
These keys are considered long term keys, and should not be thrown away at each reboot."

Let me also add that:
1. the directory needs to be writable by the apache user as the key is created the first time the server is started
2. only the apache user must be able to read this key

simo5:
The current patch moved the key in a place where apache cannot write, resulting in an ephemeral key that is thrown away each time apache is restarted/reloaded.

MartinBasti:
The issue will be fixed on the SELinux side

PR #724, opened 2017-04-20
upgrade: adtrust update_tdo_gidnumber plugin must check if adtrust is…

flo-renaud:
… installed

During upgrade, the plugin update_tdo_gidnumber is launched in order to
add a gidnumber to the Trusted Domain Object.
This plugin should not be run when ad trust is not installed, otherwise an
error message is displayed.

https://pagure.io/freeipa/issue/6881

### DIFF
--- a/ipaserver/install/plugins/adtrust.py
+++ b/ipaserver/install/plugins/adtrust.py
@@ -329,6 +329,11 @@ class update_tdo_gidnumber(Updater):
     def execute(self, **options):
         ldap = self.api.Backend.ldap2
 
+        # First, see if trusts are enabled on the server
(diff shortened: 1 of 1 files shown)


abbra:
LGTM

MartinBasti:
master:

* 434d9e539d24fe0110c5d6bf4a4342daf40d15d5 upgrade: adtrust update_tdo_gidnumber plugin must check if adtrust is installed


ipa-4-5:

* c05bd60585fb80e061b8582a648a65204c709f51 upgrade: adtrust update_tdo_gidnumber plugin must check if adtrust is installed

PR #725, opened 2017-04-21
Fix certificate_out check in CertRetrieveOverride

Akasurde:
Fixes: https://pagure.io/freeipa/issue/6885


HonzaCholasta:
NACK:

```
$ ipa cert-show 1
ipa: ERROR: Filename is empty
```

PR #726, opened 2017-04-21
Add check for directory name

Akasurde:
Fix adds check to verify if user provided input is not
a directory when filename is required.

Fixes: https://pagure.io/freeipa/issue/6883

### DIFF
--- a/ipalib/util.py
+++ b/ipalib/util.py
@@ -170,6 +170,10 @@ def check_writable_file(filename):
     """
     if filename is None:
         raise errors.FileError(reason=_('Filename is empty'))
+
+    if not os.path.isfile(filename):
+        raise errors.FileError(reason=_('Expected file but %(fname)s is not a '
+                                        'regular file' % dict(fname=filename)))
     try:
         if os.path.exists(filename):
             if not os.access(filename, os.W_OK):


ipalib/util.py
```diff
@@ -170,6 +170,10 @@ def check_writable_file(filename):
     """
     if filename is None:
         raise errors.FileError(reason=_('Filename is empty'))
+
+    if os.path.isdir(filename):
+        raise errors.FileError(reason=_('Directory name found instead of '
+                                        'Filename'))
```
stlaz: The error does not say much and has `F` in "filename" uppercased for some reason, could you rather go with something like "Expected file but {fname} is a directory".format(fname=filename)?

Akasurde: OK.

ipalib/util.py
```diff
@@ -170,6 +170,10 @@ def check_writable_file(filename):
     """
     if filename is None:
         raise errors.FileError(reason=_('Filename is empty'))
+
+    if os.path.isdir(filename):
+        raise errors.FileError(reason=_('Expected file but {fname} is a '
+                                        'directory'.format(fname=filename))
```
stlaz: You are missing a parentheses.

ipalib/util.py
MartinBasti: It must be `%(fname)s` translation strings doesn't support format (in versions we need at least)

stlaz: I keep forgetting that, thank you.

tiran:
What about other types that might cause trouble, e.g. socket, fifo, device files, dangling symlinks?

Akasurde:
@tiran Do you think we should allow only files here ?

stlaz:
Obviously we can't push this until the tests pass.

PR #730, opened 2017-04-25
spec file: bump python-netaddr Requires

HonzaCholasta:
Bump python-netaddr Requires to the version which has correct private and
reserved IPv4 address ranges.

This fixes DNS server install failure when 0.0.0.0 is entered as a
forwarder.

https://pagure.io/freeipa/issue/6894

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -33,11 +33,15 @@
 
 %global alt_name ipa
 %if 0%{?rhel}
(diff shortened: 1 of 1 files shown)


MartinBasti:
master:

* 0784e53f7f8a323acafbbff26a9d1c0276a229b0 spec file: bump python-netaddr Requires

MartinBasti:
If this should go into 4.5 please open a new PR

PR #733, opened 2017-04-26
[4.5] Fix CA/server cert validation in FIPS

stlaz:
In FIPS, the NSS library needs to be passed passwords to perform
certificate validation. Should we not have passed it and the NSS
guys have not fixed this yet, we would get SEC_ERROR_BAD_SIGNATURE
which is completely different error than one would expect but
that's just how things are with NSS right now.

https://pagure.io/freeipa/issue/6897

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -77,6 +77,11 @@ def find_cert_from_txt(cert, start=0):
     return (cert, e)
 
 
+def get_file_cont(slot, token, filename):
+    with open(filename) as f:
+        return f.read()
+
+
 class NSSDatabase(object):
     """A general-purpose wrapper around a NSS cert database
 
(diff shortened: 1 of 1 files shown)


ipapython/certdb.py
```diff
@@ -77,6 +77,12 @@ def find_cert_from_txt(cert, start=0):
     return (cert, e)
 
 
+def get_file_cont(slot, token, filename):
+    with open(filename) as f:
```
Akasurde: LGTM

ipapython/certdb.py
```diff
@@ -77,6 +77,12 @@ def find_cert_from_txt(cert, start=0):
...
+        cont = f.read()
```
tiran: ```return f.read()```

stlaz:
Made a quickfix according to @tiran, the ACK can stay. Thanks, I was being paranoid.

MartinBasti:
ipa-4-5:

* 651d132b701b773b2bbeb41496d6c5ddbf6d19b3 Fix CA/server cert validation in FIPS

PR #734, opened 2017-04-26
kerberos session: use CA cert with full cert chain for obtaining cookie

pvoborni:
Http request performed in finalize_kerberos_acquisition doesn't use
CA certificate/certificate store with full certificate chain of IPA server.
So it might happen that in case that IPA is installed with externally signed
CA certificate, the call can fail because of certificate validation
and e.g. prevent session acquisition.

If it will fail for sure is not known - the use case was not discovered,
but it is faster and safer to fix preemptively.

https://pagure.io/freeipa/issue/6876

### DIFF
--- a/ipaserver/rpcserver.py
+++ b/ipaserver/rpcserver.py
@@ -602,7 +602,8 @@ class KerberosSession(HTTP_Status):
         try:
             target = self.api.env.host
             r = requests.get('http://{0}/ipa/session/cookie'.format(target),
-                             auth=NegotiateAuth(target, ccache_name))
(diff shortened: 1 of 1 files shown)


MartinBasti:
master:

* c19196a0d3fc0a38c4c83cb8a7fde56e6bc310af kerberos session: use CA cert with full cert chain for obtaining cookie


ipa-4-5:

* 82679c11f1fc0701d753433d1f2d14c3ee0279af kerberos session: use CA cert with full cert chain for obtaining cookie

PR #735, opened 2017-04-26
automount install: do not wait for sssd restart on uninstallation

pvoborni:
Change in 2d4d1a9dc0ef2bbe86751768d6e6b009a52c0dc9 no longer initializes
api in `ipa-client-automount --uninstallation` Which caused error in
wait_for_sssd which gets realm from initialized API.

In my opinion, there is no reason to check working sssd after uninstallation by running
id command. If anything depends on running sssd then it should do the check.

Also fix call of xxx_service_class_factory which requires api as param.

https://pagure.io/freeipa/issue/6861

### DIFF
--- a/client/ipa-client-automount
+++ b/client/ipa-client-automount
@@ -193,7 +193,7 @@ def configure_autofs_sssd(fstore, statestore, autodiscover, options):
     sssdconfig.write(paths.SSSD_CONF)
     statestore.backup_state('autofs', 'sssd', True)
 
-    sssd = services.service('sssd')
+    sssd = services.service('sssd', api)
     sssd.restart()
     print("Restarting sssd, waiting for it to become available.")
     wait_for_sssd()
@@ -281,7 +281,7 @@ def uninstall(fstore, statestore):
                         break
                 sssdconfig.save_domain(domain)
                 sssdconfig.write(paths.SSSD_CONF)
-                sssd = services.service('sssd')
+                sssd = services.service('sssd', api)
                 sssd.restart()
                 wait_for_sssd()
             except Exception as e:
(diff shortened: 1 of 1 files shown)


rcritten:
I guess I have more issues with the commit message than the patch content. What would you suggest ensure that sssd is up? A typical user wouldn't notice until an nss lookup failed which likely means a login was rejected. The wait for restart was there to ensure, in interactive sessions at least, that unconfiguring automount didn't hose the system.

To me is another example of how wrong it is to require a ticket to initialize an API.

pvoborni:
Thanks Rob, this reason for the wait didn't occurred to me. New patch changes api initialization so that it works for both install and uninstall even without user's Kerberos credentials and with cleared cache,

client/ipa-client-automount
```diff
@@ -448,7 +448,11 @@ def main():
             kinit_keytab(host_princ, paths.KRB5_KEYTAB, ccache_name)
             os.environ['KRB5CCNAME'] = ccache_name
         except gssapi.exceptions.GSSError as e:
-            sys.exit("Failed to obtain host TGT: %s" % e)
+            sys.exit("Failed to obtained host TGT: %s" % e)
```
rcritten: The original line is more correct.

pvoborni:
The error message was reverted to original (I was fixing the comment below and wondered why it was not fixed, now I know).

MartinBasti:
master:

* b4e447fa6fc7d659ae6a3b6285d4ddda0baa0be4 automount install: fix checking of SSSD functionality on uninstall


ipa-4-5:

* ff513d6b20ee0a2ca90b06b8c114386f1e5751d9 automount install: fix checking of SSSD functionality on uninstall

PR #736, opened 2017-04-26
Fixing the cert-request command comparing whole email address case-sensitively.

felipevolpone:
Now, the cert-request command compares the domain part of the email case-insensitively.

Fixes: [https://pagure.io/freeipa/issue/5919](https://pagure.io/freeipa/issue/5919)

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -798,7 +798,9 @@ class cert_request(Create, BaseCertMethod, VirtualCommand):
             # fail if any email addr from DN does not appear in ldap entry
             email_addrs = csr_obj.subject.get_attributes_for_oid(
                     cryptography.x509.oid.NameOID.EMAIL_ADDRESS)
-            if len(set(email_addrs) - set(principal_obj.get('mail', []))) > 0:
+            csr_emails = [attr.value for attr in email_addrs]
+            if not _emails_are_valid(csr_emails,
+                                     principal_obj.get('mail', [])):
                 raise errors.ValidationError(
                     name='csr',
                     error=_(
@@ -884,8 +886,8 @@ class cert_request(Create, BaseCertMethod, VirtualCommand):
                             "match requested principal") % gn.name)
             elif isinstance(gn, cryptography.x509.general_name.RFC822Name):
                 if principal_type == USER:
-                    if principal_obj and gn.value not in principal_obj.get(
-                            'mail', []):
+                    if not _emails_are_valid([gn.value],
+                                             principal_obj.get('mail', [])):
                         raise errors.ValidationError(
                             name='csr',
                             error=_(
@@ -951,6 +953,25 @@ class cert_request(Create, BaseCertMethod, VirtualCommand):
             result=result,
             value=pkey_to_value(int(result['request_id']), kw),
         )
+
+
+def _emails_are_valid(csr_emails, principal_emails):
+    """
+    Checks if any email address from certificate request does not
+    appear in ldap entry, comparing the domain part case-insensitively.
+    """
+
+    def lower_domain(email):
+        email_splitted = email.split('@', 1)
+        if len(email_splitted) > 1:
+            email_splitted[1] = email_splitted[1].lower()
+
+        return '@'.join(email_splitted)
+
+    principal_emails_lower = set(map(lower_domain, principal_emails))
+    csr_emails_lower = set(map(lower_domain, csr_emails))
+
+    return csr_emails_lower.issubset(principal_emails_lower)
 
 
 def principal_to_principal_type(principal):
--- a/ipatests/test_xmlrpc/test_cert_plugin.py
+++ b/ipatests/test_xmlrpc/test_cert_plugin.py
@@ -251,6 +251,29 @@ class test_cert(BaseCert):
         res = api.Command['service_find'](self.service_princ)
         assert res['count'] == 0
 
+    def test_00011_emails_are_valid(self):
+        """
+        Verify the different scenarios when checking if any email addr
+        from DN or SAN extension does not appear in ldap entry.
+        """
+
+        from ipaserver.plugins.cert import _emails_are_valid
+        email_addrs = [u'any@EmAiL.CoM']
+        result = _emails_are_valid(email_addrs, [u'any@email.com'])
+        assert True == result, result
+
+        email_addrs = [u'any@EmAiL.CoM']
+        result = _emails_are_valid(email_addrs, [u'any@email.com',
+                                                 u'another@email.com'])
+        assert True == result, result
+
+        result = _emails_are_valid([], [u'any@email.com'])
+        assert True == result, result
+
+        email_addrs = [u'invalidEmailAddress']
+        result = _emails_are_valid(email_addrs, [])
+        assert False == result, result
+
 
 @pytest.mark.tier1
 class test_cert_find(XMLRPC_test):


ipaserver/plugins/cert.py
```diff
@@ -860,6 +861,23 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
...
 
 
+def _emails_are_valid(cert_emails, principal_emails):
+    """
+    Checks if any email addr from DN does not appear in ldap entry,
+    comparing the domain part case-insensitively.
+    """
+
+    def lower_domain(email):
+        return email.split('@')[0] + '@' + email.split('@')[1].lower()
```
frasertweedale: The case where the input has multiple '@' characters should be accounted for.  Otherwise we
could end up accepting an invalid value.  We must also handle the case where input does not
have an '@' character.  Obviously this is not a valid email but the current code would raise
`IndexError`.

Also we should avoid doing the `split` multiple times.

ipaserver/plugins/cert.py
```diff
@@ -705,7 +705,8 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
             # fail if any email addr from DN does not appear in ldap entry
             email_addrs = csr_obj.subject.get_attributes_for_oid(
                     cryptography.x509.oid.NameOID.EMAIL_ADDRESS)
-            if len(set(email_addrs) - set(principal_obj.get('mail', []))) > 0:
+            if not _emails_are_valid(email_addrs,
+                                     principal_obj.get('mail', [])):
```
frasertweedale: This is checking the email addresses (if any) in the Subject DN.  We also need to check rfc822Names in the
Subject Alternative Name extension.

felipevolpone:
@frasertweedale I did the check in SAN extension. 

However, I'm not sure if these are valid situations:
Case 1)
The principal email is A@email.com
The email in the certificate is B@email.com
The emails in the SAN extensions are: A@email.com, C@email.com

or this:

Case 2)
The principal email is A@email.com
The email in the certificate is B@email.com, A@email.com
The email in the SAN extensions is: C@email.com

If the case 1 is valid, the check in line 799 (below) is not right, because it expects that all emails in SAN extension are in the principal.
```python
elif isinstance(gn, cryptography.x509.general_name.RFC822Name):
    if principal_type == USER:
        if principal_obj and gn.value not in principal_obj.get(
                'mail', []):
            raise errors.ValidationError(
                name='csr',
                error=_(
                    "RFC822Name does not match "
                    "any of user's email addresses")
            )
    else:
        raise errors.ValidationError(
            name='csr',
            error=_(
                "subject alt name type %s is forbidden "
                "for non-user principals") % "RFC822Name"
```

ipaserver/plugins/cert.py
```diff
@@ -860,6 +866,33 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
...
+
+    if not any(principal_emails):
+        return False
+
+    def lower_domain(email):
+        if '@' not in email:
+            return ''
+
+        email_splited = email.split('@')
+        return '{}@{}'.format(email_splited[0], email_splited[1].lower())
```
frasertweedale: This logic still needs to be fixed.  It is simple to construct a CSR with emails that have no `@` (or multiple).
(It probably possible to make a bogus LDAP attribute value, too).

So I'd suggest something like:

```python
email_split = email.split('@', 1)  # split only at first occurrence of '@'
if len(email_split) > 1:
    email_split[1] = email_split[1].lower()
return email_split
```

There is no need to "rejoin" the parts back into a string; we can compare on the resulting
lists.  If you end up using `set` then you can turn the lists into tuples so that you have a
hashable value.

felipevolpone: I did just a little bit different because the example above it doesn't work. The `split` method would return a list like `["fbarreto", "redhat.com"]`, so, it's necessary to join the two parts with a `@`.

ipaserver/plugins/cert.py
```diff
@@ -705,7 +705,13 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
             # fail if any email addr from DN does not appear in ldap entry
             email_addrs = csr_obj.subject.get_attributes_for_oid(
                     cryptography.x509.oid.NameOID.EMAIL_ADDRESS)
-            if len(set(email_addrs) - set(principal_obj.get('mail', []))) > 0:
+
+            san_email_addrs = csr_obj.extensions.get_extension_for_oid(
+                    cryptography.x509.oid.ExtensionOID
+                                     .SUBJECT_ALTERNATIVE_NAME)
+
+            if not _emails_are_valid(email_addrs, san_email_addrs,
```
felipevolpone: I prefer the option the option "a", since it seems more straightforward.

felipevolpone:
I hope it's fine now

ipaserver/plugins/cert.py
```diff
@@ -791,8 +793,10 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
                             "match requested principal") % gn.name)
             elif isinstance(gn, cryptography.x509.general_name.RFC822Name):
                 if principal_type == USER:
-                    if principal_obj and gn.value not in principal_obj.get(
-                            'mail', []):
+                    gn_value = (gn.value if isinstance(gn.value, list)
+                                         else [gn.value])
```
frasertweedale: I think `gn.value` will always be a string, so you can just put:

```python
if not _emails_are_valid([gn.value], principal_obj.get('mail', [])):
```

If I'm wrong about that, could you please explain?

ipaserver/plugins/cert.py
```diff
@@ -860,6 +864,27 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
...
+    """
+
+    if not any(principal_emails):
+        return False
+
+    def lower_domain(email):
+        email_splited = email.split('@', 1)
+        email_splited[1] = email_splited[1].lower()
+
+        return '@'.join(email_splited)
```
frasertweedale: The `join` is OK, but I think not necessary, because you will always be comparing values
returned from `lower_domain`, and lists will compare properly.  So joining is just a bit of
unneeded but harmless work.

ipaserver/plugins/cert.py
```diff
@@ -860,6 +864,27 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
...
+
+    principal_emails_lower = set(map(lower_domain, principal_emails))
+    csr_emails_lower = set(map(lower_domain, csr_emails))
+
+    return csr_emails_lower.issubset(principal_emails_lower)
```
frasertweedale: This looks correct :+1:

ipatests/test_xmlrpc/test_cert_plugin.py
```diff
@@ -253,6 +253,43 @@ def test_00010_cleanup(self):
...
+        Verify the different scenarios when checking if any email addr
+        from DN or SAN extension does not appear in ldap entry.
+        """
+
+        from ipaserver.plugins.cert import _emails_are_valid
+        from collections import namedtuple
+        NameAttr = namedtuple('NameAttr', 'value')
+
+        subject_addrs = [NameAttr(u'any@EmAiL.CoM')]
+        result = _emails_are_valid(subject_addrs, [], [u'any@email.com'])
```
frasertweedale: These tests need updating now that signature of `_emails_are_valid` changed.

ipaserver/plugins/cert.py
felipevolpone: In all tests that I did, `gn.value` was a string. However, since I don't know very much all about the code, I thought it would be a "safer" solution. Changed to your suggestion :)

ipaserver/plugins/cert.py
felipevolpone: Yeah, I prefer to stay with the '@' join to be more consistent with the method name and all the context. If you think that it's not relevant at all, I can just remove it :)

ipatests/test_xmlrpc/test_cert_plugin.py
felipevolpone: Done. I was waiting to we agree with the implementation and then change the tests.

ipaserver/plugins/cert.py
```diff
@@ -953,6 +955,27 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
...
 
 
+def _emails_are_valid(csr_emails, principal_emails):
+    """
+    Checks if any email address from certificate does not
+    appear in ldap entry, comparing the domain part case-insensitively.
+    """
+
+    if not any(principal_emails) or not any(csr_emails):
+        return False
```
frasertweedale: We should get rid of this.  The subject check is sufficient.  If there are no `csr_emails`, then
the test should be vacuously `True`, not `False`.  Similarly, if there are no `principal_emails`, then
the valid if and only if there are also no `csr_email`.  All of this is covered by the subset check,
so just remove these two lines.

ipaserver/plugins/cert.py
frasertweedale: Well, it's not needed, but it doesn't hurt, and I do not mind if we keep it.

ipatests/test_xmlrpc/test_cert_plugin.py
```diff
@@ -251,6 +251,29 @@ def test_00010_cleanup(self):
...
+        result = _emails_are_valid(email_addrs, [u'any@email.com'])
+        assert True == result, result
+
+        email_addrs = [u'any@EmAiL.CoM']
+        result = _emails_are_valid(email_addrs, [u'any@email.com',
+                                                 u'another@email.com'])
+        assert True == result, result
+
+        result = _emails_are_valid([], [u'any@email.com'])
+        assert False == result, result
```
frasertweedale: As discussed above, this case should be `True`.

ipaserver/plugins/cert.py
felipevolpone: Ok

ipaserver/plugins/cert.py
```diff
@@ -961,14 +961,12 @@ def _emails_are_valid(csr_emails, principal_emails):
...
     """
 
-    if not any(principal_emails) or not any(csr_emails):
-        return False
-
     def lower_domain(email):
-        email_splited = email.split('@', 1)
-        email_splited[1] = email_splited[1].lower()
+        email_splitted = email.split('@', 1)
+        if len(email_splitted) > 1:
```
felipevolpone: This became necessary since we don't have the if statement (in line 964) anymore.

ipaserver/plugins/cert.py
```diff
@@ -953,6 +955,25 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
         )
 
 
+def _emails_are_valid(csr_emails, principal_emails):
+    """
+    Checks if any email address from certificate does not
```
frasertweedale: Should say "certificate request"

felipevolpone:
Done! Thank you Fraser :)) :+1:

ipaserver/plugins/cert.py
```diff
@@ -884,8 +886,8 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
                             "match requested principal") % gn.name)
             elif isinstance(gn, cryptography.x509.general_name.RFC822Name):
                 if principal_type == USER:
-                    if principal_obj and gn.value not in principal_obj.get(
```
frasertweedale: @felipevolpone sorry mate, one more tiny thing that I missed.  We need to retain the `if principal_obj and ...` because there is a scenario where `principal_obj = None` - namely, the PKINIT KDC certificate use case.  While it's unlikely that the CSR in this use case will include an RFC822Name, we should make the code resilient to this case.

felipevolpone: Hi Fraser, ok, no problem. 
So, we should have this?
```python
 if principal_obj and not _emails_are_valid([gn.value], principal_obj.get('mail', [])):
```

frasertweedale: @felipevolpone actually, I was confused.  This whole bit is guarded by an `if principal_type == USER`, and if that is true, then `principal_obj` is set.  So we are actually good to go; the check that was there before your change was
superfluous.

MartinBasti:
master:

* d973168e89c7fb5e8c36919b3adb685371e1a3ab Fixing the cert-request comparing whole email address case-sensitively.

PR #737, opened 2017-04-26
Vault: Explicitly default to 3DES CBC

tiran:
The server-side plugin for IPA Vault relied on the fact that the default
oid for encryption algorithm is 3DES in CBC mode (DES-EDE3-CBC). Dogtag
10.4 has changed the default from 3DES to AES. Pass the correct
algorithm OID to KeyClient.archive_encrypted_data().

Closes: https://pagure.io/freeipa/issue/6899

### DIFF
--- a/ipaserver/plugins/vault.py
+++ b/ipaserver/plugins/vault.py
@@ -38,6 +38,14 @@ from ipapython.dn import DN
 if api.env.in_server:
     import pki.account
     import pki.key
+    # pylint: disable=no-member
+    try:
+        # pki >= 10.4.0
+        from pki.crypto import DES_EDE3_CBC_OID
+    except ImportError:
+        DES_EDE3_CBC_OID = pki.key.KeyClient.DES_EDE3_CBC_OID
+    # pylint: enable=no-member
+
 
 if six.PY3:
     unicode = str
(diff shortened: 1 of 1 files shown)


tiran:
* I haven't verified that the patch actually solves the problem
* Needs backport to at least 4.5
* Either needs backport to 4.4 or 4.4 must required Dogtag < 10.4

pvoborni:
Should go to 4.4.5 unless  pki-core-10.4.0-1 is removed from f25. Blocking new Dogtag update in 4.4 doesn't seem right to me.

tiran:
I talked to Matt. Dogtag 10.4 will not be pushed to F25 and F26, only rawhide/F27. Additionally, Ade will also address the bug in Dogtag. The next 10.4 release will have a fix, too.

frasertweedale:
Tested; fix makes it work again against Dogtag (where Dogtag does not contain Ade's fix).  ACK.

MartinBasti:
master:

* 5197422ef65e7239fc56c562ab87d99388a38a8d Vault: Explicitly default to 3DES CBC


ipa-4-5:

* e94a1d18653fe2e9558ac0b70bdf2ddd1f78d150 Vault: Explicitly default to 3DES CBC

PR #738, opened 2017-04-26
restore: restart gssproxy after restore

pvoborni:
So that gssproxy picks up new configuration and therefore related
usages like authentication of CLI against server works

https://pagure.io/freeipa/issue/6902

@simo5 btw, what is the proper name of gssproxy? Is it GSSAPI proxy, gss-proxy or gssproxy?

Note: if this patch is wrong, feel free to take over and abolish this PR.

### DIFF
--- a/ipaserver/install/ipa_restore.py
+++ b/ipaserver/install/ipa_restore.py
@@ -401,6 +401,9 @@ class Restore(admintool.AdminTool):
                 services.knownservices.pki_tomcatd.enable()
                 services.knownservices.pki_tomcatd.disable()
 
+                self.log.info('Restarting GSSAPI proxy')
+                gssproxy = services.service('gssproxy', api)
+                gssproxy.restart()
                 self.log.info('Starting IPA services')
                 run(['ipactl', 'start'])
                 self.log.info('Restarting SSSD')


simo5:
The name of the project is GSS-Proxy, the package name is gssproxy.

simo5:
will a "systemctl reload gssproxy" do the right thing @frozencemetery ?

frozencemetery:
The systemd-portable way to do this is as I understand it is `systemctl try-reload-or-restart gssproxy` (unless you want to start it if it's not running, at which point you'd do `systemctl reload-or-restart gssproxy`).  Then you're not dependent on how the distro maintainer decides to handle it (which varies between distros here due to reasons).

That is something of an orthogonal consideration though - digging further into the code, it doesn't look like there's a `reload`-style verb of any kind in the services abstraction layer?  (I see `is_running()` and `restart()`, but no `reload()` in services.py.)  I don't know how you want to address this; whether it makes sense to extend the abstraction, or whether since no one else uses it gssproxy can just be `restart`ed.

pvoborni:
PR #748 obsoletes this one - this PR was created badly and so I cannot force update it. New one uses reload-or-restart

PR #739, opened 2017-04-27
[4.5] spec file: bump krb5 Requires for certauth fixes

HonzaCholasta:
Bump krb5-* Requires to the version which includes the final version of
certauth support.

https://pagure.io/freeipa/issue/4905


martbab:
ipa-4-5:

* ec3a2a6063beb4ec96796b66abb82476a5c7bd0f spec file: bump krb5 Requires for certauth fixes

PR #741, opened 2017-04-27
6.9 -> 7.4 migration fixes

stlaz:
**Allow rewriting of cached properties**
    
Cached property should not be treated anyway special from a normal
property. If we need to rewrite/remove it, we should be able to do
just so.

**Refresh Dogtag RestClient.ca_host property**
    
Refresh the ca_host property of the Dogtag's RestClient class when
it's requested as a context manager.
    
This solves the problem which would occur on DL0 when installing
CA against an old master which does not have port 8443 accessible.
The setup tries to update the cert profiles via this port but
fail. This operation should be performed against the local instance
anyway.

https://pagure.io/freeipa/issue/6878

### DIFF
--- a/ipalib/util.py
+++ b/ipalib/util.py
@@ -34,7 +34,6 @@ import dns
 import encodings
 import sys
 import ssl
-from weakref import WeakKeyDictionary
 
 import netaddr
 from dns import resolver, rdatatype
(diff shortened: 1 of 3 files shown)


ipalib/util.py
```diff
@@ -520,10 +520,8 @@ def __get__(self, obj, cls):
         return self.store[obj]
 
     def __set__(self, obj, value):
-        raise AttributeError("can't set attribute")
+        self.store[obj] = value
```
tiran: If you are going to turn the cached property into a writeable and deletable property, then you can remove the complicated weakref stuff and replace it with a simple non-data descriptor:

```
class cachedproperty(object):
    slots = ('fget',)
    def __init__(self, fget):
        self.fget = fget
    def __get__(self, obj, objtype=None):
        if obj is None:
            return self
        value = self.fget(obj)
        setattr(obj, self.fget.__name__, value)
        return value
```

tiran: ah, and add ```self.__doc__ = fget.__doc__```.

stlaz: Decided to go with a simpler approach, pylint was a hater anyway.

stlaz:
For the record - the tests are passing on my machine, etwas stimmt hier nicht.

ipaserver/plugins/dogtag.py
```diff
@@ -1250,34 +1249,45 @@ def __init__(self, api):
...
         """
-        :return:   host
-                   as str
+        :returns: FQDN of a host hopefully providing a CA service
 
-        Select our CA host.
+        Select our CA host, cache it for the first time.
         """
+        if self._ca_host is not None:
+            return self._ca_host
```
tiran: yeah, that's simpler 👍

MartinBasti:
It failed to me
```
  [20/28]: Configure HTTP to proxy connections
  [21/28]: restarting certificate server
  [22/28]: migrating certificate profiles to LDAP
  [error] NetworkError: cannot connect to 'https://vm-058-166.abc.idm.lab.eng.brq.redhat.com:8443/ca/rest/account/login': [Errno 111] Connection refused
```

stlaz:
This was supposed to be fixed by the patch and worked for me, it seems that I may need to investigate it further.

stlaz:
Turns out I forgot to reorder the CA installation steps a bit.

MartinBasti:
master:

* 0d406fcb784924bfe685729f3156efb8c902b947 Refresh Dogtag RestClient.ca_host property
* 92313c9e9d37733feb79d1b1c825178f48d6c69c Remove the cachedproperty class


ipa-4-5:

* 32981a0f9d0ff699e3d16da8f5a37c112871ba3a Refresh Dogtag RestClient.ca_host property
* 9de343987e6d76d2edeba372c73c1060657aef59 Remove the cachedproperty class

PR #742, opened 2017-04-27
Revert "Store GSSAPI session key in /var/run/ipa"

martbab:
This reverts commit 2bab2d4963daa99742875f3633a99966bc56f5a3. It was
pointed out that apache has no access to /var/lib/ipa directry breaking
the session handling.


martbab:
master:

* 50f6883662e258b0335c8b3cb69946d6dcbf206c Revert "Store GSSAPI session key in /var/run/ipa"

PR #743, opened 2017-04-27
[ipa-4-5] Revert "Store GSSAPI session key in /var/run/ipa"

martbab:
This reverts commit 2bab2d4. It was
pointed out that apache has no access to /var/lib/ipa directry breaking
the session handling.


martbab:
ipa-4-5:

* a4e1ab6c893182b8b3610c0b45120194be4a0376 Revert "Store GSSAPI session key in /var/run/ipa"

PR #745, opened 2017-04-27
tests: add missing dependency iptables

MartinBasti:
KDC proxy tests are using iptables, but this is optional package in at
least Fedora cloud image, thus we must have it in dependencies


tomaskrizek:
master:

* 6c061b6836c13bf63553c6143b19e89658937e7e tests: add missing dependency iptables

MartinBasti:
AFAIK nobody complains about this issue except me, and I'm fine with master only

PR #746, opened 2017-04-27
KDC proxy URI records

MartinBasti:
Automatic creation of KDC proxy URI records
    
Enables creation of following KDC proxy URL records per each replica:
 ```   
 _kerberos.example.com. IN URI <prio> <weight> "krb5srv:M:kkdcp:https://ipaserver.example.com/KdcProxy"
 _kpasswd.example.com. IN URI <prio> <weight> "krb5srv:M:kkdcp:https://ipaserver.example.com/KdcProxy"
```
Records are created for each replica in topology as KDC proxy is enabled
by default. (If KDC proxy is manually disabled this record will be
created anyway)
    
URI records for kadmin discovery are not created because FreeIPA doesn't
support kadmin.
    
See: https://k5wiki.kerberos.org/wiki/Projects/KDC_Discovery
    
https://pagure.io/freeipa/issue/6337

### TODO
- [x] Add URI records for 88/UDP, 88/TCP with higher priority to keep client asking directly KDC first
- [ ] Add URI records for HTTPS only when kdc proxy is enabled on server (requires to update server roles with role attribute KDC proxy)
- [ ] Allows to set priorities of IPA system records by admins
- [ ] Update release notes
- `Postponed`: Fix https://pagure.io/freeipa/issue/6906 to allow enroll FreeIPA clients using HTTPS KDC proxy

### DIFF
--- a/ipaserver/dns_data_management.py
+++ b/ipaserver/dns_data_management.py
@@ -37,6 +37,23 @@ IPA_DEFAULT_MASTER_SRV_REC = (
     (DNSName(u'_kpasswd._udp'), 464),
 )
 
+IPA_DEFAULT_KDC_URI_REC = (
+    # URI record name, target
+    (DNSName(u'_kpasswd'), u'krb5srv:M:tcp:{server}'),
+    (DNSName(u'_kpasswd'), u'krb5srv:M:udp:{server}'),
+    (DNSName(u'_kerberos'), u'krb5srv:M:tcp:{server}'),
+    (DNSName(u'_kerberos'), u'krb5srv:M:udp:{server}'),
+)
+
+# URI records for KDCProxy must have lower priority than for KDC, clients must
+# prefer to connect directly to KDC
+IPA_KDCPROXY_PRIORITY_PENALIZATION = 10
+IPA_DEFAULT_KDCPROXY_URI_REC = (
+    # URI record name, target
+    (DNSName(u'_kpasswd'), u'krb5srv:M:kkdcp:https://{server}/KdcProxy'),
+    (DNSName(u'_kerberos'), u'krb5srv:M:kkdcp:https://{server}/KdcProxy'),
+)
+
 IPA_DEFAULT_ADTRUST_SRV_REC = (
     # srv record name, port
     (DNSName(u'_ldap._tcp.Default-First-Site-Name._sites.dc._msdcs'), 389),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: ipaserver/dns_data_management.py [hunk 1]
```diff
@@ -37,6 +37,23 @@
...
+    # URI record name, target
+    (DNSName(u'_kpasswd'), u'krb5srv:M:tcp:{server}'),
+    (DNSName(u'_kpasswd'), u'krb5srv:M:udp:{server}'),
+    (DNSName(u'_kerberos'), u'krb5srv:M:tcp:{server}'),
+    (DNSName(u'_kerberos'), u'krb5srv:M:udp:{server}'),
+)
+
+# URI records for KDCProxy must have lower priority than for KDC, clients must
+# prefer to connect directly to KDC
+IPA_KDCPROXY_PRIORITY_PENALIZATION = 10
```
tiran: Don't be so shy, give it a big boost of 100 or even 1000. It makes it a bit easier to manually tune relevance later. Priority is a 16bit unsigned int, https://tools.ietf.org/html/rfc7553#section-4.2 .

MartinBasti: I'm shy because with DNS locations we use +50, and users may get lost when I increase it too much

MartinBasti: And manual tune for IPA DNS records is currently unsupported, it will replace any custom change done by users during upgrade

tiran: Does the DNS location system add +50 for all off-site locations?

simo5:
I am not entirely sure we want to care for the cse where an admin disables KDC Proxy in an automatic fashion; otherwise we would also need to check if TCP or UDP are disabled and change that too.
FreeIPA as a product enables TCP/UDP and proxy and an admin that wants to change this by manually changing configurations should also take care of manually changing the URI records in DNS I think. Just like they would need to change records in DNS if either TCP or UDP protocols were disabled.
However if it is overly simple to detect and update records based on enabled protocols I am not against doing so.

MartinBasti:
@simo5 we don't support manual changes of IPA system records, it is regenerated automatically, so any manual changes are overwritten when: new replica is added/replica is removed/user executes manually `ipa dns-update-system-records`

simo5:
@MartinBasti In this case we need a way to tell the system what are the priorities and which protocols are enabled, priorities are important too, admins need to be able to change them as they see fit.

MartinBasti:
@simo5 not really a 4.5 material then

simo5:
We can probably defer.

PR #747, opened 2017-04-27
vault: piped input for ipa vault-add fails

flo-renaud:
An exception is raised when using echo "Secret123\n" | ipa vault-add myvault

This happens because the code is using (string).decode(sys.stdin.encoding)
and sys.stdin.encoding is None when the input is read from a pipe.
The fix is using the prompt_password method defined by Backend.textui,
which gracefully handles this issue.

https://bugzilla.redhat.com/show_bug.cgi?id=1445358

### DIFF
--- a/ipaclient/plugins/vault.py
+++ b/ipaclient/plugins/vault.py
@@ -21,11 +21,9 @@ from __future__ import print_function
 
 import base64
 import errno
-import getpass
 import io
 import json
 import os
-import sys
 import tempfile
 
 from cryptography.fernet import Fernet, InvalidToken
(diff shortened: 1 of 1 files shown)


ipaclient/plugins/vault.py
```diff
@@ -304,7 +279,8 @@ def forward(self, *args, **options):
                 password = password.rstrip('\n')
 
             else:
-                password = get_new_password()
+                password = self.api.Backend.textui.prompt_password(
```
Akasurde: LGTM

stlaz: Is this LGTM meant for the whole patch or just this one-liner? If the prior, please, don't add your LGTMs to single lines, if the latter, please, do complete reviews.

stlaz:
@Akasurde: Don't add ACK label when the PR is not OK!
@flo-renaud: You will need to specify a ticket for this PR.

flo-renaud:
@stlaz 
Thank you for the reminder. Commit msg updated with issue 6907

stlaz:
Thank you for the brief action taken. Re-adding the ACK label.

ipaclient/plugins/vault.py
Akasurde: LGTM was for complete review. I tried the patch and worked for me, so add ACK and LGTM.

MartinBasti:
master:

* d5c41ed4ad370c7d74296a830993a5bd3fd32e5f vault: piped input for ipa vault-add fails


ipa-4-5:

* c8ca0f89a68b5d57c56344fdeb12fd436976c726 vault: piped input for ipa vault-add fails

PR #748, opened 2017-04-28
restore: restart/reload gssproxy after restore

pvoborni:
So that gssproxy picks up new configuration and therefore related
usages like authentication of CLI against server works

https://pagure.io/freeipa/issue/6902

### DIFF
--- a/ipaplatform/base/services.py
+++ b/ipaplatform/base/services.py
@@ -154,6 +154,10 @@ class PlatformService(object):
 
         return
 
+    def reload_or_restart(self, instance_name="", capture_output=True,
+                          wait=True):
+        return
+
     def restart(self, instance_name="", capture_output=True, wait=True):
         return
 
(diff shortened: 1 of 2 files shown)


pvoborni:
Obsoletes PR #738

ipaplatform/base/services.py
```diff
@@ -298,14 +302,25 @@ def start(self, instance_name="", capture_output=True, wait=True):
             instance_name,
             update_service_list=update_service_list)
 
-    def restart(self, instance_name="", capture_output=True, wait=True):
-        ipautil.run([paths.SYSTEMCTL, "restart",
-                     self.service_instance(instance_name)],
+    def _restart_base(self, instance_name, operation, capture_output=True,
```
tomaskrizek: I'd either set `operation="restart"` as default or rename the method to something more descriptive.

pvoborni: I think it's better to pass explicit operation to the function to prevent surprise behavior.  The public functions are then more readable.

Feel free to suggest better name for the base function.

pvoborni: It was meant as a base class for: "reload", "restart", "try-restart", "reload-or-restart" and "try-reload-or-restart". The name could be `reload_restart_base`, but given that it is private/protect function then it can be renamed at any time in the future without any trouble when new methods are introduced as needed..

tomaskrizek: The method as is could be used for any systemctl command and I think the name should reflect that.

How about `_systemctl_exec` or `_systemctl_command`?

tomaskrizek: Nevermind, since the method also waits for open ports, the name shouldn't suggest it just executes a systemctl command.

tomaskrizek:
How is this patch going to work for Debian? Shouldn't we also implement `reload_or_restart` for `DebianSysvService`?

pvoborni:
Should work:

```
def debian_service_class_factory(name, api=None):
    if name == 'dirsrv':
        return redhat_services.RedHatDirectoryService(name, api)
    if name == 'domainname':
        return DebianNoService(name, api)
    if name == 'ipa':
        return redhat_services.RedHatIPAService(name, api)
    if name == 'messagebus':
        return DebianNoService(name, api)
    if name == 'ntpd':
        return DebianSysvService("ntp", api)
    return DebianService(name, api)
```
so it's `DebianService`

```
class DebianService(redhat_services.RedHatService):
    system_units = debian_system_units
```
then
```
class RedHatService(base_services.SystemdService):
```

I.e. it is not `DebianSysvService`

tomaskrizek:
Ok, everything looks good then.

tomaskrizek:
master:

* 3a4c8e39c3e38ec651cfcbb3cac59e0e92e04fe0 restore: restart/reload gssproxy after restore


ipa-4-5:

* 04ed1fa3acdf002ecc37dde4f5d226c0fbe5aa30 restore: restart/reload gssproxy after restore

PR #749, opened 2017-04-28
Added plugins directory to python2-ipaclient subpackage

olivergs:
Subpackage does not own that directory and could create conflicts if a plugin creates it on its onwn


MartinBasti:
Hello, could you please also add it for python3-ipaclient too?

olivergs:
Added :)

MartinBasti:
Thanks, could you please merge into one patch?

olivergs:
Created new PR #756

PR #752, opened 2017-05-02
upgrade: add missing DN suffix when enabling KDC proxy

tomaskrizek:
This issue prevented from upgrading from IPA 4.1.

I also discovered a missing python dependency when I was running the ipa-server-upgrade manually. For packagers: the Python version that has the required symbols in CentOS is 2.7.5-24

https://pagure.io/freeipa/issue/6920

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -643,6 +643,7 @@ Requires: python-gssapi >= 1.2.0
 Requires: gnupg
 Requires: keyutils
 Requires: pyOpenSSL
+Requires: python >= 2.7.9
 Requires: python-cryptography >= 1.6
 Requires: python-netaddr >= %{python_netaddr_version}
 Requires: python-libipa_hbac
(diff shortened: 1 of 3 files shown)


martbab:
This makes me wonder why don't we just use `self.api.env.basedn` in the ldap_enable/enable_kdcproxy/enable_pkinit methods but instead rely on 3 ways how to construct the root DN.

LGTM as a quick fix, but we need to overhaul this part service installers in the future.

MartinBasti:
Works for me

martbab:
A separate PR will be needed for ipa-4-5 branch.

MartinBasti:
master:

* 999706fcdfa7fd4206a2399aa578fb00753d9978 python2-ipalib: add missing python dependency
* 4b8ab77dd4800bd9c6b822502462ee649c88c663 installer service: fix typo in service entry
* ebefb281775d5bd5f32459ac597af78781d7dbf5 upgrade: add missing suffix to http instance

PR #754, opened 2017-05-03
ipa-server-install with external CA: fix pkinit cert issuance

flo-renaud:
ipa-server-install with external CA fails to issue pkinit certs.
This happens because the installer calls
krb = krbinstance.KrbInstance(fstore)
then
krb.enable_ssl()
and in this code path self.config_pkinit is set to None, leading to a wrong
code path.

The fix initializes the required fields of the krbinstance before calling
krb.enable_ssl.

https://pagure.io/freeipa/issue/6921

### DIFF
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -108,6 +108,14 @@ class KrbInstance(service.Service):
     suffix = ipautil.dn_attribute_property('_suffix')
     subject_base = ipautil.dn_attribute_property('_subject_base')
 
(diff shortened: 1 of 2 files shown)


stlaz:
LGTM, will test it.

tomaskrizek:
ipa-4-5:

* 8107125e177ac9f378d149d7b0fa1d3774c9be3a ipa-server-install with external CA: fix pkinit cert issuance


master:

* a24923066dd95a88ded329f1a558d46fbb9d8f81 ipa-server-install with external CA: fix pkinit cert issuance

PR #755, opened 2017-05-03
Use proper SELinux context with http.keytab

MartinBasti:
During upgrade keytab is moved to a new location using "move" operation.
This commit replaces move operation with "copy" and "remove" that
ensures a proper selinux context.

https://pagure.io/freeipa/issue/6924

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -1427,7 +1427,15 @@ def update_ipa_httpd_service_conf(http):
 def update_http_keytab(http):
     root_logger.info('[Moving HTTPD service keytab to gssproxy]')
     if os.path.exists(paths.OLD_IPA_KEYTAB):
-        shutil.move(paths.OLD_IPA_KEYTAB, http.keytab)
+        # ensure proper SELinux context by using copy operation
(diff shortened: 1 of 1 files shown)


ipaserver/install/server/upgrade.py
```diff
@@ -1427,7 +1427,15 @@ def update_ipa_httpd_service_conf(http):
...
     root_logger.info('[Moving HTTPD service keytab to gssproxy]')
     if os.path.exists(paths.OLD_IPA_KEYTAB):
-        shutil.move(paths.OLD_IPA_KEYTAB, http.keytab)
+        # ensure proper SELinux context by using copy operation
+        shutil.copy(paths.OLD_IPA_KEYTAB, http.keytab)
+        try:
+            os.remove(paths.OLD_IPA_KEYTAB)
+        except OSError as e:
+            root_logger.error(
+                'Cannot remove file %s (%s). Please remove file manually.',
```
martbab: nitpicky nit: there should be `the` before `file`.

martbab:
master:

* 7f4c2fbd975d09c01e6898a4eb70d7dfea1171b4 Use proper SELinux context with http.keytab
ipa-4-5:

* bda733db9ede3307595963a8c086e1b700c41e25 Use proper SELinux context with http.keytab

PR #757, opened 2017-05-03
ca, kra install: validate DM password

tomaskrizek:
Prevent CA and KRA installation from proceeding if provided DM password is invalid to avoid broken installations with no possibility to uninstall CA or KRA.

https://pagure.io/freeipa/issue/6892

### DIFF
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -117,17 +117,19 @@ def parse_options():
 
 
 def _get_dirman_password(password=None, unattended=False):
+    # sys.exit() is used on purpose, because otherwise user is advised to
+    # uninstall the component, even though it is not needed
     if not password:
         if unattended:
             sys.exit('Directory Manager password required')
-        try:
-            password = installutils.read_password(
-                "Directory Manager (existing master)", confirm=False,
-                validate=False)
-        except KeyboardInterrupt:
-            sys.exit(0)
-        if password is None:
-            sys.exit("Directory Manager password required")
+        password = installutils.read_password(
+            "Directory Manager (existing master)", confirm=False,
+            validate=False)
+    try:
+        installutils.validate_dm_password_ldap(password)
+    except ValueError:
+        sys.exit("Directory Manager password is invalid")
+
     return password
 
 
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -116,9 +116,19 @@ def parse_options():
     return safe_options, options, filename
 
 
-def get_dirman_password():
-    return installutils.read_password(
-        "Directory Manager (existing master)", confirm=False, validate=False)
+def _get_dirman_password(password=None, unattended=False):
+    if not password:
+        if unattended:
+            sys.exit('Directory Manager password required')
+        try:
+            password = installutils.read_password(
+                "Directory Manager (existing master)", confirm=False,
+                validate=False)
+        except KeyboardInterrupt:
+            sys.exit(0)
+        if password is None:
+            sys.exit("Directory Manager password required")
+    return password
 
 
 def install_replica(safe_options, options, filename):
@@ -142,16 +152,8 @@ def install_replica(safe_options, options, filename):
         check_creds(options, api.env.realm)
 
     # get the directory manager password
-    dirman_password = options.password
-    if not dirman_password:
-        if options.unattended:
-            sys.exit('Directory Manager password required')
-        try:
-            dirman_password = get_dirman_password()
-        except KeyboardInterrupt:
-            sys.exit(0)
-        if dirman_password is None:
-            sys.exit("Directory Manager password required")
+    dirman_password = _get_dirman_password(
+        options.password, options.unattended)
 
     if (not options.promote and not options.admin_password and
             not options.skip_conncheck and options.unattended):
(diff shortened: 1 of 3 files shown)


install/tools/ipa-ca-install
```diff
@@ -116,9 +116,22 @@ def parse_options():
...
+def _get_dirman_password(password=None, unattended=False):
+    if not password:
+        if unattended:
+            sys.exit('Directory Manager password required')
+        try:
+            password = installutils.read_password(
+                "Directory Manager (existing master)", confirm=False,
+                validate=False)
+        except KeyboardInterrupt:
+            sys.exit(0)
```
tiran: keyboard interrupt should not result in exit code 0. A common exit code is 130 (128 + SIGINT).

MartinBasti: I'm afraid that this is inconsistent with other parts of code. The rest will return error code 1 in case of SIGINT.

If you want to keep new error code, then manpage should be updated as well.

MartinBasti: Or even better, don't catch that exception and let `installutils.run_script()` to handle it in the same way for every script.

ipaserver/install/installutils.py
```diff
@@ -329,6 +330,19 @@ def _read_password_default_validator(password):
...
+def dm_password_validator(password):
+    """
+    Validate DM password by attempting to connect to LDAP. api.env has to
+    contain valid ldap_uri.
+    """
+    client = LDAPClient(api.env.ldap_uri, cacert=paths.IPA_CA_CRT)
+    try:
+        client.simple_bind(DIRMAN_DN, password)
+    except errors.ACIError:
+        raise ValueError("Invalid Directory Manager password")
```
MartinBasti: I'm sorry but I think you may missing following code there:

```
else:
    client.unbind()
```

install/tools/ipa-ca-install
```diff
@@ -126,8 +126,11 @@ def _get_dirman_password(password=None, unattended=False):
                 validate=False)
         except KeyboardInterrupt:
             sys.exit(130)
-        if password is None:
-            sys.exit("Directory Manager password required")
+    try:
+        installutils.dm_password_validator(password)
+    except ValueError:
+        sys.exit("Directory Manager password is invalid")
```
martbab: Please don't use sys.exit() explicitly here, either let the exception propagate up the stack to the caller or re-raise RuntimeError as you do in ipa-kra-install

ipaserver/install/installutils.py
martbab: Also I would rather use verbs to name functions/methods, like `validate_dm_password` (but I have a suspicion that one is already taken).

stlaz:
There will be no more sys.exits. This patchset shall not be ACKed until all have been removed.

tomaskrizek:
Thanks for the feedback, hopefully I addressed all the issues.

install/tools/ipa-ca-install
```diff
@@ -116,9 +116,19 @@ def parse_options():
     return safe_options, options, filename
 
 
-def get_dirman_password():
-    return installutils.read_password(
-        "Directory Manager (existing master)", confirm=False, validate=False)
+def _get_dirman_password(password=None, unattended=False):
+    if not password:
+        if unattended:
+            raise RuntimeError('Directory Manager password required')
```
MartinBasti: Please use script error to be consistent

```
ipaserver/install/server/install.py:            raise ScriptError("Directory Manager password required")
```

install/tools/ipa-ca-install
```diff
@@ -116,9 +116,19 @@ def parse_options():
...
+        password = installutils.read_password(
+            "Directory Manager (existing master)", confirm=False,
+            validate=False)
+    try:
+        installutils.validate_dm_password_ldap(password)
+    except ValueError:
+        raise RuntimeError("Directory Manager password is invalid")
```
MartinBasti: Please use `ScriptError` here as well.

ipaserver/install/ipa_kra_install.py
```diff
@@ -137,6 +137,13 @@ def ask_for_options(self):
     def run(self):
         super(KRAInstaller, self).run()
 
+        # Verify DM password. This has to be called after ask_for_options(),
+        # so it can't be placed in validate_options().
+        try:
+            installutils.validate_dm_password_ldap(self.options.password)
+        except ValueError:
+            raise RuntimeError("Directory Manager password is invalid")
```
MartinBasti: `ScriptError` please :)

tomaskrizek:
Implementing the tests shouldn't block us from pushing this fix. I opened a ticket for it: https://pagure.io/freeipa/issue/6941

MartinBasti:
We have to use `sys.exit()` in this case, because I forgot that CA still uses old style installer. Without `sys.exit()` ti will always suggest user to uninstall server:

```
```

We don't want to uninstall server due typo in password

stlaz:
You forgot an import in ipa-ca-install:
```
************* Module ipa-ca-install

install/tools/ipa-ca-install:37: [W0611(unused-import), ] Unused ScriptError imported from ipapython.admintool)
```

MartinBasti:
master:

* 80d61c2e01bdce0ab805037bfc1ce8e9543d2b10 ca install: merge duplicated code for DM password
* 7a4a368c5387569f6802b5b3ca0686ea20f1de7e installutils: add DM password validator
* 1b1bace75095cf7a26a56156ff537479ef4b9619 ca, kra install: validate DM password


ipa-4-5:

* 282fc0c86474bafcb28234eabbd807b99a98adec ca install: merge duplicated code for DM password
* 4c12b71717b2ca1d4af5018f77c07f8f4b4feca5 installutils: add DM password validator
* b8bcaa61ec6c9effcf029f82ca21685b692e0b7f ca, kra install: validate DM password

PR #760, opened 2017-05-03
[4.4] Run ipa-custodia under Python 2

tiran:
Closes: https://pagure.io/freeipa/issue/6926

### DIFF
--- a/init/systemd/ipa-custodia.service
+++ b/init/systemd/ipa-custodia.service
@@ -3,8 +3,7 @@ Description=IPA Custodia Service
 
 [Service]
 Type=simple
-
-ExecStart=/usr/sbin/custodia /etc/ipa/custodia/custodia.conf
+ExecStart=/usr/libexec/ipa/ipa-custodia /etc/ipa/custodia/custodia.conf
 PrivateTmp=yes
 Restart=on-failure
 RestartSec=60s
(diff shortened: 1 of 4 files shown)


init/systemd/ipa-custodia.service
```diff
@@ -3,8 +3,7 @@ Description=IPA Custodia Service
 
 [Service]
 Type=simple
-
-ExecStart=/usr/sbin/custodia /etc/ipa/custodia/custodia.conf
+ExecStart=@libexecdir@/ipa/ipa-custodia /etc/ipa/custodia/custodia.conf
```
stlaz: This macro does not get replaced resulting in the service not being properly configured.

stlaz:
Works for me. However, I do not see the reason to do `custodia > 0.2`, please, either provide some or remove it.

tiran:
The patch doesn't work with custodia 0.1. It needs at least 0.2.0-1.

stlaz:
Alright, thanks. ACK.

tomaskrizek:
@tiran Once this PR is merged, do you plan to remove the conflict with `freeipa-server-common < 4.5` from custodia or do we need to address something else as well?

tiran:
@tomaskrizek yes, custodia 0.5 will no longer have the restriction.

tomaskrizek:
Needs re-base for 4.4 and 4.5.

tomaskrizek:
ipa-4-4:

* 307c4bd62609c9ac58633e3ccc61d85e2caacbcc Run ipa-custodia under Python 2

lslebodn:
@tiran Do you plan to release custodia-0.2 in fedora 25?
Because this patch broke building of ipa-4.4 on fedora 25.

If you do not plan to upgrade custodia in f25 then this patch should be (partially) reverted

PR #762, opened 2017-05-04
fix managed-entries printing IPA not installed

stlaz:
ipa-managed-entries would print "IPA is not configured on this system."
even though this is not true if run as a normal user. Add check for
root running the script.

https://pagure.io/freeipa/issue/6928


tomaskrizek:
master:

* 6522c4a8378a22ffe82e8e845698ab104f611888 fix managed-entries printing IPA not installed

PR #763, opened 2017-05-04
Dogtag: fail if instance cannot be (re)started

stlaz:
**Make CA/KRA fail when they don't start**

Since all the services throw exceptions when we're unable to
start/restart them, CA/KRA should not be an exception to it.

**Fix wrong message on Dogtag instances stop**

https://pagure.io/freeipa/issue/6766

### DIFF
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -156,22 +156,10 @@ class DogtagInstance(service.Service):
                       ignore_errors=True)
 
     def restart_instance(self):
-        try:
-            self.restart('pki-tomcat')
-        except Exception:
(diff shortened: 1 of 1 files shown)


MartinBasti:
master:

* 1a7a1f955e327bf1a06faa53c517bdbffff22eba Make CA/KRA fail when they don't start
* aba384ddb535e81f81a518fa468a8ed095250ca1 Fix wrong message on Dogtag instances stop


ipa-4-5:

* 81f97cb89e17e63b3dcb8925a373970ac61764c2 Make CA/KRA fail when they don't start
* 1b44c4caa1e7a1f90b3b3537de9cc1529f0891e8 Fix wrong message on Dogtag instances stop

PR #766, opened 2017-05-05
ipa-kra-install: fix check_host_keys

flo-renaud:
ipa-kra-install on a replica checks that the keys are available before
going further to avoid race condition due to replication. The issue is
that the check_host_keys method expects to find exactly one key for
cn=env/host but 2 may exist: one below cn=custodia and one below
cn=dogtag,cn=custodia.
The fix is to check that at least one key exist (not exactly one key).

https://pagure.io/freeipa/issue/6934

### DIFF
--- a/ipaserver/secrets/kem.py
+++ b/ipaserver/secrets/kem.py
@@ -72,7 +72,7 @@ class KEMLdap(iSecLdap):
                                          'princ': principal})
         r = conn.search_s(self.keysbase, scope, ldap_filter)
         if len(r) != 1:
-            raise ValueError("Incorrect number of results (%d) searching for"
+            raise ValueError("Incorrect number of results (%d) searching for "
                              "public key for %s" % (len(r), principal))
         ipa_public_key = r[0][1]['ipaPublicKey'][0]
         jwk = self._parse_public_key(ipa_public_key)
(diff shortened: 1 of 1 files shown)


ipaserver/secrets/kem.py
```diff
@@ -85,8 +85,8 @@ def check_host_keys(self, host):
 
         ldap_filter = self.build_filter(IPA_CHECK_QUERY, {'host': host})
         r = conn.search_s(self.keysbase, scope, ldap_filter)
-        if len(r) != 1:
-            raise ValueError("Incorrect number of results (%d) searching for"
+        if len(r) < 1:
```
MartinBasti: Hello,

I think that `if not r:` is more pythonic and I'd also rephrase the error message, that "No public keys were found ...."

It is just nitpick, thanks.

martbab: Nitpicks for now, but AFAIK pylint 1.7.1 will report these checks as errors.

flo-renaud:
Hi @MartinBasti  @martbab 
thank you for the comment. PR updated with your suggestion.

MartinBasti:
master:

* 8983ce53e3fdee98926f81f3012146e33bb92d30 ipa-kra-install: fix check_host_keys


ipa-4-5:

* b90dce88e227174aa33270beee9b3d6ff51cce59 ipa-kra-install: fix check_host_keys

PR #768, opened 2017-05-09
Ticket#6854 caless

Rezney:
What was done?

~~~
1.) caless-create-pki

The script was kind of merged with https://github.com/freeipa/freeipa-tools/blob/master/makepki.sh. Standa took care of PKINIT certificates generation so that write_chain() function was introduced which handles cert chain in the pkcs12 files and also reverse chanin order for openssl command.

Then gen_pkinit_extensions() and gen_pkinit_cert() are handling the PKINIT certificate generation. See https://web.mit.edu/kerberos/krb5-1.13/doc/admin/pkinit.html for details.

2.) test_caless.py

As the tests are currently failing due to the pkinit option not provided "pkinit_pin, pkinit_pkcs12_exists and pkinit_pkcs12" parameters were added to both install_server() and prepare_replica methods and particular options are added to installator. Then copy_pkinit() is handling pkinit certs transfer.

TestPKINIT class contains test_server_replica_install_pkinit() test which checks both server and replica install with pkinit for a starter.

Eventually added "raiseonerr=False" to ipa_certs_cleanup() cause tests were failing there but that whole workaround for ticket 4639 will be removed in different commit.
~~~

What can be improved? (at least what I am aware of)

~~~
Currently pkinit certificates are not inside nss db so we copy it separately (we could also move it to certdir and copy as whole). Tried to put it there with pk12util but the certs were getting nicknames from openssl friendly names (I guess). Added -name parameter to "openssl pkcs12 -export" command and the nicknames were fine (e.g. "ca1/pkinit-server" after certuril -L) however after the "caless-create-pki" script was done all pkinit cert nicknames were just prefixed with "ca1/" (instead of ca1/ ca2/ etc.).
~~~

Issues found:

~~~
Replica install with pkinit is not failing anymore with "Certificate issuance failed (CA_UNREACHABLE)", however the ERROR message is still presented:


[ipa.ipatests.pytest_plugins.integration.host.Host.vm-021.cmd26]   [1/1]: installing X509 Certificate for PKINIT
[ipa.ipatests.pytest_plugins.integration.host.Host.vm-021.cmd26] ipa         : ERROR    PKINIT certificate request failed: Certificate issuance failed (CA_UNREACHABLE)
[ipa.ipatests.pytest_plugins.integration.host.Host.vm-021.cmd26] ipa         : ERROR    Failed to configure PKINIT
[ipa.ipatests.pytest_plugins.integration.host.Host.vm-021.cmd26] Done configuring Kerberos KDC (krb5kdc).
[ipa.ipatests.pytest_plugins.integration.host.Host.vm-021.cmd26] Applying LDAP updates
[ipa.ipatests.pytest_plugins.integration.host.Host.vm-021.cmd26] Upgrading IPA:. Estimated time: 1 minute 30 seconds
[ipa.ipatests.pytest_plugins.integration.host.Host.vm-021.cmd26]   [1/9]: stopping directory server
~~~


abbra:
PKINIT certificates are using by `krb5kdc` which uses OpenSSL. It means they cannot be placed in an NSSDB.

Rezney:
Ah, sorry I was not descriptive enough. I meant a temporary nssdb which is created by the script on the controller which is running the integration tests. However thanks for your input. Good to know this.

MartinBasti:
This PR is obsoleted by #769

PR #770, opened 2017-05-09
cert-show: writable files does not mean dirs

stlaz:
ipalib.util.check_writable_file didn't check whether the argument
is an actual file which is now fixed.

https://pagure.io/freeipa/issue/6883


tomaskrizek:
master:

* 33b3d7ad7ada45edbd178fe99f1257c40f39dcaa cert-show: writable files does not mean dirs


ipa-4-5:

* 2410023ce6ef3255ddbaaf8939a928e733297d62 cert-show: writable files does not mean dirs

PR #773, opened 2017-05-09
[WIP] Warn in cert-request if CSR doesn't contain SAN

felipevolpone:
The code is not "production-ready", however, I would like to know if I'm on the right path. 

AFAIK we should check if the SAN extension is provided and if it has DNSName info.

Fix: https://pagure.io/freeipa/issue/6663

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -771,6 +771,18 @@ class cert_request(Create, BaseCertMethod, VirtualCommand):
         cn = cns[-1].value  # "most specific" is end of list
 
         if principal_type in (SERVICE, HOST):
+
+            has_dns_in_san_ext = False
+            if ext_san:
+                for gn in x509.process_othernames(ext_san.value):
+                    if isinstance(gn, cryptography.x509.general_name.DNSName):
+                        has_dns_in_san_ext = True
+
+            if not ext_san or not has_dns_in_san_ext:
+                print('Warning: The SAN extension '
+                      'should be provided. Please, check the RFC 2818.')
+
+
             if not _dns_name_matches_principal(cn, principal, principal_obj):
                 raise errors.ValidationError(
                     name='csr',
--- a/install/ui/src/freeipa/dns.js
+++ b/install/ui/src/freeipa/dns.js
@@ -294,6 +294,11 @@ return {
         height: 300,
         sections: [
             {
+                name: 'dnszone_title',
+                label: 'Select the required zone type.',
+                fields: []
+            },
+            {
                 name: 'name',
                 layout: IPA.dnszone_name_section_layout,
                 fields: [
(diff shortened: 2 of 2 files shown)


frasertweedale:
Was there agreement that this should be implemented?  (I am personally
against it, because the next release should update the default profile to use
the new CommonNameToSanExtDefault profile component).

If we do implement this, IMO it should be a per-profile configuration, because there may
be legitimate use cases where SAN is not needed.

If we do pursue the current approach, we should further check not only that SAN
is present, but that it contains a DNSName.  Put another way, with the current patch,
SAN can be present, but it might contain only KRB5PrincipalName and no DNSName,
and therefore the warning will not show, but it probably should have warned.

HonzaCholasta:
@frasertweedale, I'm not aware of any agreement and I'm against this as well.

ipaserver/plugins/cert.py
```diff
@@ -771,6 +771,18 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
...
         if principal_type in (SERVICE, HOST):
+
+            has_dns_in_san_ext = False
+            if ext_san:
+                for gn in x509.process_othernames(ext_san.value):
+                    if isinstance(gn, cryptography.x509.general_name.DNSName):
+                        has_dns_in_san_ext = True
+
+            if not ext_san or not has_dns_in_san_ext:
+                print('Warning: The SAN extension '
```
stlaz: http://stackoverflow.com/questions/6918493/in-python-why-use-logging-instead-of-print

MartinBasti: Hi, I assume you want show warning to user not just in httpd log. For this case you have to use `self.add_message()` method in plugin, otherwise warning is not propagated to client.

stlaz: Ah, you were trying to show the message to the user, do that ^ instead.

pvoborni:
AFAIK, there was not an agreement not implementing this, otherwise the ticket would be closed. The ticket #6663 was created to warn until the change in profiles is implemented(#4970).  It was mentioned yesterday on IPA meeting that we want to warn - when discussing: https://bugzilla.redhat.com/show_bug.cgi?id=1445345 and https://bugzilla.redhat.com/show_bug.cgi?id=1445927

ipaserver/plugins/cert.py
felipevolpone: I used `print` to make clear that the code was not in its final version and that I was trying to show the message to the user; I just didn't know how to do it. I'll try to be more straightforward next time. ty!

pvoborni:
I don't think it makes sense to spend time on configuration of warning - that is larger change (ldap attr, schema, api...) and as such out of scope of 4.5.

Simple warning is IMO good, but it should be worded in a sense that SAN is not always needed. Probably mention in what general use cases it is needed e.g. web services/pages.

PR #774, opened 2017-05-10
Remove pkinit-anonymous command

stlaz:
Ever since from v4.5, FreeIPA expects at least some kind of
anonymous PKINIT to work. Deprecate the command which is
capable of turning this feature off.

https://pagure.io/freeipa/issue/6936

### DIFF
--- a/ipaserver/plugins/pkinit.py
+++ b/ipaserver/plugins/pkinit.py
@@ -1,51 +1,13 @@
-# Authors:
-#   Simo Sorce <ssorce@redhat.com>
 #
-# Copyright (C) 2010  Red Hat
-# see file 'COPYING' for use and warranty information
+# Copyright (C) 2017  FreeIPA Contributors see COPYING for license
 #
-# This program is free software; you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from ipalib import api, errors
-from ipalib import Str
-from ipalib import Object, Command
+from ipalib import Object
 from ipalib import _
 from ipalib.plugable import Registry
-from ipalib.constants import ANON_USER
-from ipapython.dn import DN
-
-__doc__ = _("""
-Kerberos pkinit options
-
-Enable or disable anonymous pkinit using the principal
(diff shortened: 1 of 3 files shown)


ipaserver/plugins/pkinit.py
```diff
@@ -17,36 +17,21 @@
...
- Enable anonymous pkinit:
-  ipa pkinit-anonymous enable
-
- Disable anonymous pkinit:
-  ipa pkinit-anonymous disable
-
-For more information on anonymous pkinit see:
-
-http://k5wiki.kerberos.org/wiki/Projects/Anonymous_pkinit
+This module is deprecated since FreeIPA 4.5.1
```
martbab: I would write that the whole module is deprecated (as there will probably be the `pkinit-status` command in this module, see https://github.com/freeipa/freeipa/pull/790. You can rather write that the pkinit-anonymous commands are deprecated and I will then re-write the module docstring in my PR.

ipaserver/plugins/pkinit.py
```diff
@@ -57,49 +42,22 @@ class pkinit(Object):
...
-    __doc__ = _('Enable or Disable Anonymous PKINIT.')
-
-    princ_name = '%s@%s' % (ANON_USER, api.env.realm)
-    default_dn = DN(('krbprincipalname', princ_name), ('cn', api.env.realm), ('cn', 'kerberos'), api.env.basedn)
+    __doc__ = _('Originally to enable or disable Anonymous PKINIT.\n'
+                'Deprecated since FreeIPA 4.5.1')
 
     takes_args = (
-        Str('action', valid_arg),
+        Str('action?'),
```
martbab: Why is `action` turned into a optional argument? is this required?

stlaz: I don't want users to be bothered by error messages that `action` is required or whatever is printed there should `action` not be provided, given that the output stays the same no matter what.

ipaserver/plugins/pkinit.py
stlaz: Ok.

ipaserver/plugins/pkinit.py
martbab: True if it is deprecated and no-op, then we should not bother users about required options. Do you agree @HonzaCholasta ?

HonzaCholasta: We should remove the command completely, as it never really worked.

abbra:
Just remove the command completely. FreeIPA prior to 4.5 never supported PKINIT operations and never allowed using anonymous PKINIT. Disabling/enabling it was left for admins that knew what they wanted. However, with FreeIPA 4.5 we require anonymous PKINIT to be enabled all time -- be it with a local self-signed cert or with some other certificate issued by a proper CA. An anonymous principal can only be used to create a FAST channel, nothing else.

stlaz:
Thanks, pruned the file a bit.

martbab:
master:

* 24099d0f806103d8ec57d69fc97e9b4ae061bfdd Remove pkinit-anonymous command

martbab:
ipa-4-5 needs a separate PR due to merge conflicts.

PR #781, opened 2017-05-11
Warn in cert-request if CSR doesn't contain SAN

felipevolpone:
The code is not "production-ready", however, I would like to know if I'm on the right path.

AFAIK we should check if the SAN extension is provided and if it has DNSName info.

Fix: https://pagure.io/freeipa/issue/6663

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -770,7 +770,22 @@ class cert_request(Create, BaseCertMethod, VirtualCommand):
                 error=_("No Common Name was found in subject of request."))
         cn = cns[-1].value  # "most specific" is end of list
 
+        principal_type = 3
         if principal_type in (SERVICE, HOST):
+
+            has_dns_in_san_ext = False
+            if ext_san:
+                for gn in x509.process_othernames(ext_san.value):
+                    if isinstance(gn, cryptography.x509.general_name.DNSName):
+                        has_dns_in_san_ext = True
+
+            if not ext_san or not has_dns_in_san_ext:
+                self.add_message(
+                    "Warning: Certificate request for a [{0}, {1}] "
+                    "certificate did not contain Subject Alternative Name, "
+                    "which is required for many use cases."
+                    .format(principal.hostname, principal.service_name))
+
             if not _dns_name_matches_principal(cn, principal, principal_obj):
                 raise errors.ValidationError(
                     name='csr',


felipevolpone:
Hi everyone, 

after a long long day, I did a great job deleting the branch from PR #773, then Github closed it.
If someone knows how to reopen it, great. Otherwise, I create a new branch and this new PR. 

Sorry about that :(

Notifying people that were following that thread: @stlaz @MartinBasti @frasertweedale @HonzaCholasta @pvoborni

ipaserver/plugins/cert.py
```diff
@@ -771,6 +771,17 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
...
         if principal_type in (SERVICE, HOST):
+
+            has_dns_in_san_ext = False
+            if ext_san:
+                for gn in x509.process_othernames(ext_san.value):
+                    if isinstance(gn, cryptography.x509.general_name.DNSName):
+                        has_dns_in_san_ext = True
+
+            if not ext_san or not has_dns_in_san_ext:
+                print('Warning: The SAN extension '
```
frasertweedale: I think we should say "Subject Alternative Name" instead of the abbreviation "SAN".

frasertweedale: Also recalling the review comments from https://github.com/freeipa/freeipa/pull/773#pullrequestreview-37222207 about sending the warning back to the client, and the `print` is just placeholder while final version
of patch is worked out.

ipaserver/plugins/cert.py
```diff
@@ -771,6 +771,17 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
...
+                      'should be provided. Please, check the RFC 2818.')
```
frasertweedale: I don't think we need to reference the RFC.

Ideally we should suggest to the user what to do, but because we don't know what tool created the CSR
(certmonger, certutil, openssl, ...?) it is a bit hard to give concrete steps to user about how to solve their problem.

I would say something like: "Certificate request for a {host,service} certificate did not contain Subject Alternative Name, which is required for many use cases."

This warning will be suprious if the profile uses the new CommonNameToSANDefault component (assuming the CN looks like a DNS name).   I have a PR to add this component to `caIPAserviceCert` (the default profile) for new installs only.

I'm just not sure if this won't create more support load than it avoids.  It is hard to guess what will happen.

felipevolpone: I changed the message, please, checks if it looks good.

felipevolpone:
@frasertweedale I did the changes you proposed, however I'm not sure if there is an agreement in the team that this ticket should be done.

frasertweedale:
@felipevolpone I've put the ticket up for re-triage.  Let's see what comes out of that.

felipevolpone:
Closing it. It was decided to do not have this feature/warning.

PR #782, opened 2017-05-11
[WIP] Improving GUI text in "Add DNS Zones" popup

felipevolpone:
Improving usability of the "Add DNS Zones" popup in Web UI.

Ticket: https://pagure.io/freeipa/issue/6687

### DIFF
--- a/install/ui/src/freeipa/dns.js
+++ b/install/ui/src/freeipa/dns.js
@@ -294,6 +294,11 @@ return {
         height: 300,
         sections: [
             {
+                name: 'dnszone_title',
+                label: 'Select the required zone type.',
+                fields: []
+            },
+            {
                 name: 'name',
                 layout: IPA.dnszone_name_section_layout,
                 fields: [
@@ -307,6 +312,7 @@ return {
                         $type: 'dnszone_name',
                         name: 'name_from_ip',
                         radio_name: 'dnszone_name_type',
+                        required: false,
                         validators: ['network']
                     }
                 ]
@@ -750,18 +756,12 @@ IPA.add_dns_zone_name_policy = function() {
             idnsname_w.input.prop('disabled', false);
             name_from_ip_w.input.prop('disabled', true);
 
-            idnsname_f.set_required(true);
-            name_from_ip_f.set_required(false);
-
             name_from_ip_f.reset();
         });
 
         name_from_ip_w.radio_clicked.attach(function() {
             idnsname_w.input.prop('disabled', true);
             name_from_ip_w.input.prop('disabled', false);
-
-            idnsname_f.set_required(false);
-            name_from_ip_f.set_required(true);
 
             idnsname_f.reset();
         });


pvoborni:
I'm not completely sure that the approach suggested in bug report is correct. That is why I suggested alternative in  https://bugzilla.redhat.com/show_bug.cgi?id=1419834#c2  

So before implementing it a small conversation could have happen to agree on the approach.

install/ui/src/freeipa/dns.js
```diff
@@ -294,6 +294,11 @@ return {
         height: 300,
         sections: [
             {
+                name: 'dnszone_title',
+                label: 'Select the required zone type.',
```
pvoborni: Except login, all strings in Web UI are localized. This method won't localize it.  Web UI has a text module to get a string from API, resp. from translations loaded on startup.  Search for `text.get('@i18n:`

pvomacka:
Hello @felipevolpone ,
Thank you for your patch. 
For adding arbitrary text into a dialog or details page is probably the most suitable IPA.html_widget (but it has a big disadvantage - described below in section A). You can put it into the section you created. It might look like this: 
```
{
  name: 'dnszone_title',
  show_header: false,
  fields: [
    {
      field: false,
      $type: 'html',
      name: 'info',
      html: "Select the required zone type."
    }
  ],
  layout: {
    $factory: IPA.fluid_layout,
    widget_cls: "col-sm-12 controls",
    label_cls: "hide"
  }
},
```
Layout attribute of the section might not be needed, but I would say that here it good to add it. It hides label of field and set width of the field to 100% of the dialog. 

(Simpler solutions below - B and C)
A) The html attribute contains text which will be displayed. Text there should be taken from translatable strings. It can be done by using `text.get('i18n:path.to.the.string')` and writing the string into ipaserver/internal.py. The main challenge here might be to find a place where the string has to be loaded. It has to be done before building the whole dialog and its sections. You will probably need to override `dialog_build_properites` attribute of entity specification and there change `$post_ops` operation which where is the function which builds adder dialog for entity (add there loading of translate string).

B) (not tested) Another solution would be to set text field instead of html one and turn off the field in the same way as above and then set it non-writable and read_only. Then hide the label and there the `text.get()` should work directly in field definition. (should not be needed to change behavior of building entity's adder dialog). 

C) Another solution will be to create new widget, which will work in the same way as `IPA.html_widget` but it will support translatable strings.

If you have any question feel free to ask. :)

pvomacka:
Sorry I haven't refresh the page so I didn't see @pvoborni comment before I sent mine. The suggestion which Petr wrote into Bugzilla should be discussed with @MartinBasti  and if I recall correctly he did not recommend it from point of view of DNS.

pvoborni:
Ok, when one field is not usuable because IP address or network address are also valid DNS zones, then the proper way is to follow patternfly design for this kind of workflows: http://www.patternfly.org/pattern-library/forms-and-controls/progressive-disclosure/

pvomacka:
Yes, this pattern should be used. We already have a widget for this (without hiding not-selected area) and it is used i.e. in certmapdata adder dialog which could be opened from user's details page. Try to look for `multiple_choice_section`.

felipevolpone:
I'm dropping this PR because I'm focused on other tickets for now

PR #784, opened 2017-05-12
ipa-replica-manage del (dl 0): remove server from defaultServerList

flo-renaud:
ipa-replica-manage del should remove the server from the entry
cn=default,ou=profile,$BASE
The entry contains an attribute
defaultServerList: srv1.domain.com srv2.domain.com srv3.domain.com

The code calls srvlist = ret.single_value.get('defaultServerList') which means
that srvlist contains a single value (string) containing all the servers
separated by a space, and not a list of attribute values. Because of that,
srvlist[0] corresponds to the first character of the value.
The fix splits srvlist and not srvlist[0].

https://pagure.io/freeipa/issue/6946

### DIFF
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -1336,17 +1336,17 @@ class ReplicationManager(object):
             dn = DN(('cn', 'default'), ('ou', 'profile'), self.suffix)
             ret = self.conn.get_entry(dn)
             srvlist = ret.single_value.get('defaultServerList', '')
-            srvlist = srvlist[0].split()
+            srvlist = srvlist.split()
             if replica in srvlist:
                 srvlist.remove(replica)
                 attr = ' '.join(srvlist)
-                mod = [(ldap.MOD_REPLACE, 'defaultServerList', attr)]
-                self.conn.modify_s(dn, mod)
+                ret['defaultServerList'] = attr
+                self.conn.update_entry(ret)
         except errors.NotFound:
             pass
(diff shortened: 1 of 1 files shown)


ipaserver/install/replication.py
```diff
@@ -1336,12 +1336,12 @@ def replica_cleanup(self, replica, realm, force=False):
...
             if replica in srvlist:
                 srvlist.remove(replica)
                 attr = ' '.join(srvlist)
-                mod = [(ldap.MOD_REPLACE, 'defaultServerList', attr)]
-                self.conn.modify_s(dn, mod)
+                ret['defaultServerList'] = attr
+                self.conn.update_entry(ret)
         except errors.NotFound:
             pass
         except ldap.NO_SUCH_ATTRIBUTE:
```
martbab: since you now call `update_entry` instead of raw `modify_s`, you can replace `ldap.NO_SUCH_ATTRIBUTE` and `ldap.TYPE_OR_VALUE_EXISTS` exceptions with `errors.MidairCollision` and `errors.EmptyModlist` (see `error_handler` method of `LDAPClient` in `ipapython/ipaldap.py` for more info about raw ldap -> ipalib.errors mapping)

flo-renaud:
Hi @martbab 
Thank you for the review. I updated the PR with your comments.

martbab:
master:

* 319a079f6decc81c7f0912ee5aef239768db554b ipa-replica-manage del (dl 0): remove server from defaultServerList

PR #785, opened 2017-05-12
otptoken-add-yubikey: When --digits not provided use default value

dkupka:
Since Thin client was introduced default values for options are not populated
in client side plugins. When option has default value and is needed in client
plugin it must be handled by explicitly.

https://pagure.io/freeipa/issue/6900


stlaz:
Works for me.

MartinBasti:
ipa-4-5:

* 749fc90d1fde0d012acb05ba64309f4a6ed63124 otptoken-add-yubikey: When --digits not provided use default value


master:

* e415da22f350fbda5b8b341bf2dc5f969cecb84a otptoken-add-yubikey: When --digits not provided use default value

PR #786, opened 2017-05-15
ipa-server-install: fix uninstall

flo-renaud:
ipa-server-install --uninstall fails to stop tracking the certificates
because it assigns a tuple to the variable nicknames, then tries to
call nicknames.append(). This is a regression introduced by 21f4cbf8.

Assignment should be done using nicknames = list(self.tracking_reqs) instead.

https://pagure.io/freeipa/issue/6950

### DIFF
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -319,7 +319,7 @@ class DogtagInstance(service.Service):
         services.knownservices.messagebus.start()
         cmonger.start()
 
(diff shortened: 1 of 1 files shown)


HonzaCholasta:
master:

* d9ed2573fd5b4dcdc8ea865f16d81325707e0f9d ipa-server-install: fix uninstall


ipa-4-5:

* 752e167497eca87632261dec7bbb352cd0e599c8 ipa-server-install: fix uninstall

PR #792, opened 2017-05-16
named.conf template: add modification warning

tomaskrizek:
Based on our discussion with Brian Atkinsson, I think we should add some warning about making undocumented changes to the config files.

---

Warn the user some modifications may break IPA setup or upgrade.
All changes the user makes should be explicitly supported and mentioned
in the documentation. Undocumented and unsupported changes, such as
renaming dyndb part from "ipa" will break IPA.

### DIFF
--- a/install/share/bind.named.conf.template
+++ b/install/share/bind.named.conf.template
@@ -43,6 +43,9 @@ zone "." IN {
 include "$RFC1912_ZONES";
 include "$ROOT_KEY";
 
+/* WARNING: This part of the config file is IPA-managed.
+ * Modifications may break IPA setup or upgrades.
+ */
 dyndb "ipa" "$BIND_LDAP_SO" {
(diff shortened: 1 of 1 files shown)


stlaz:
Aside from the tests failing because DNS fails to start with your patch, I don't think you can claim this easily without listing the supported modifications.

tomaskrizek:
- fixed the broken comment
- mentioned any changes should be documented in FreeIPA or Red Hat IdM documentation

I don't think going into more detail is necessary.

stlaz:
The IdM documentation does not list allowed operations on the file, only a note of what the file is most commonly used for.

tomaskrizek:
Fair enough, let's go with a simple warning that changes may break IPA setup/upgrades.

MartinBasti:
master:

* a924efe847e725d6a4dde209f57521a9a899ef23 named.conf template: add modification warning

PR #797, opened 2017-05-19
ipa-replica-conncheck: handle ssh not installed

flo-renaud:
When ipa-replica-conncheck is run but ssh is not installed, the tool exits
with a stack trace. Properly handle the error by raising an Exception.

https://pagure.io/freeipa/issue/6935

### DIFF
--- a/install/tools/ipa-replica-conncheck
+++ b/install/tools/ipa-replica-conncheck
@@ -59,12 +59,11 @@ class SshExec(object):
         self.user = user
         self.addr = addr
         self.cmd = distutils.spawn.find_executable('ssh')
-
-    def __call__(self, command, verbose=False):
         # Bail if ssh is not installed
         if self.cmd is None:
-            root_logger.warning("WARNING: ssh not installed, skipping ssh test")
-            return ('', '', 0)
+            raise RuntimeError("ssh not installed")
+
+    def __call__(self, command, verbose=False):
 
         tmpf = tempfile.NamedTemporaryFile()
         cmd = [
@@ -596,7 +595,11 @@ def main():
 
                 # Ticket 5812 Always qualify requests for admin
                 user = principal
-                ssh = SshExec(user, options.master)
+                try:
+                    ssh = SshExec(user, options.master)
+                except RuntimeError as e:
+                    root_logger.warning("WARNING: %s, skipping ssh test" % e)
+                    return 0
 
                 root_logger.info("Check SSH connection to remote master")
                 result = ssh('echo OK', verbose=True)


install/tools/ipa-replica-conncheck
```diff
@@ -63,8 +63,7 @@ class SshExec(object):
     def __call__(self, command, verbose=False):
         # Bail if ssh is not installed
         if self.cmd is None:
-            root_logger.warning("WARNING: ssh not installed, skipping ssh test")
-            return ('', '', 0)
+            raise RuntimeError("ssh not installed, skipping ssh test")
```
martbab: IMHO it would be a bit cleaner to move the whole check for `self.cmd` to the object constructor so that we can bail out as early as possible.

martbab:
I have a small inline comment. I also miss any processing of the raised exception in the caller. If we want to skip SSH check when ssh is not installed (and we may argue about that), the called needs to catch RuntimeError from SshExec and log at warning/error level.

flo-renaud:
Hi @martbab 
thank you for the review, I agree with your comments. My plan was to raise the RuntimeException so that ipa-replica-conncheck exits on error when ssh is not installed (same behavior as if ssh exits on error). The output would be:
    Retrying using SSH...
    ERROR: Remote master check failed with the following error message:
    Could not SSH to remote host, ssh not installed
and exit code would be 1.

martbab:
I was quite not sure what to do when SSH is not installer because the intent of the original code (at least from what I inferred) was that the absence of SSHd should trigger a warning message and skip SSH checks. Your commit changes the behavior so that this state is a hard error.

Shouldn't we first keep the original behavior?

flo-renaud:
Hi @martbab 
I initially thought that the code never worked and always triggered an exception, but you are right, it used to skip the check when ssh client is not installed. I updated the PR to revert to the original behavior.

install/tools/ipa-replica-conncheck
```diff
@@ -594,7 +593,11 @@ def main():
 
                 # Ticket 5812 Always qualify requests for admin
                 user = principal
-                ssh = SshExec(user, options.master)
+                try:
+                    ssh = SshExec(user, options.master)
+                except RuntimeError as e:
+                    root_logger.warning("WARNING: %s, skipping ssh test" % e)
+                    return
```
martbab: We should return 0 explicitly as in original code to fullfill the `main()` function's contract (returns integers to sys.exit which passed them as scripts return codes).

flo-renaud: Hi @martbab 
thank you for the comment. I updated the PR to return 0 explicitely.

martbab:
ipa-4-5:

* bacccb70a2e91efa22ee19aec9cca75bac94bd95 ipa-replica-conncheck: handle ssh not installed


master:

* f960450820c13284b52b4c5f420f0f1191a45619 ipa-replica-conncheck: handle ssh not installed

PR #800, opened 2017-05-19
Change python-cryptography to python2-cryptography

pvomacka:
Package name is python2-cryptography and even that it Provides
python-cryptography package, it causes problems during update of IPA
on RHEL - python2-cryptography is not updated. After changing required package
name to python2-cryptography upgrade on RHEL works well.

Fixes: https://pagure.io/freeipa/issue/6749

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -168,7 +168,7 @@ BuildRequires:  python3-wheel
 %if 0%{?with_lint}
 BuildRequires:  samba-python
 # 1.6: x509.Name.rdns (https://github.com/pyca/cryptography/issues/3199)
-BuildRequires:  python-cryptography >= 1.6
+BuildRequires:  python2-cryptography >= 1.6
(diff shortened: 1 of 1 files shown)


martbab:
Shouldn't this be a downstream-only patch, given that it affects only RHEL while Fedora resolves the dependencies correctly?

pvoborni:
Bumping the dep upstream later  without this change have potential to introduce the bug again in downstream if one is not careful.  This will also unify the usages as ipatests package uses python2-cryptography already.

MartinBasti:
Voting for explicit `python2-*` prefix

martbab:
@pvoborni thank you for explanation I am OK with this change then.

MartinBasti:
master:

* 9149f2d9c6e90f6dd974ea4317f71d17db3c869c Change python-cryptography to python2-cryptography

MartinBasti:
Please provide a PR for ipa-4-5 branch

PR #801, opened 2017-05-22
httpinstance: wait until the service entry is replicated

HonzaCholasta:
Wait until the local HTTP service entry is replicated to the remote master
before requesting the server certificate.

This prevents a replication conflict between the service entry added
locally and service entry added remotely when requesting the certificate.

https://pagure.io/freeipa/issue/6867

### DIFF
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -32,9 +32,11 @@ import six
 from augeas import Augeas
 
 from ipalib.install import certmonger
+from ipapython import ipaldap
 from ipapython.certdb import (IPA_CA_TRUST_FLAGS,
                               EXTERNAL_CA_TRUST_FLAGS,
                               TRUSTED_PEER_TRUST_FLAGS)
+from ipaserver.install import replication
 from ipaserver.install import service
 from ipaserver.install import certs
 from ipaserver.install import installutils
@@ -120,12 +122,15 @@ class HTTPInstance(service.Service):
 
     subject_base = ipautil.dn_attribute_property('_subject_base')
 
-    def create_instance(self, realm, fqdn, domain_name, pkcs12_info=None,
+    def create_instance(self, realm, fqdn, domain_name, dm_password=None,
+                        pkcs12_info=None,
                         subject_base=None, auto_redirect=True, ca_file=None,
-                        ca_is_configured=None, promote=False):
+                        ca_is_configured=None, promote=False,
+                        master_fqdn=None):
         self.fqdn = fqdn
         self.realm = realm
         self.domain = domain_name
+        self.dm_password = dm_password
         self.suffix = ipautil.realm_to_suffix(self.realm)
         self.pkcs12_info = pkcs12_info
         self.dercert = None
@@ -141,6 +146,7 @@ class HTTPInstance(service.Service):
         if ca_is_configured is not None:
             self.ca_is_configured = ca_is_configured
         self.promote = promote
+        self.master_fqdn = master_fqdn
 
         self.step("stopping httpd", self.__stop)
         self.step("setting mod_nss port to 443", self.__set_mod_nss_port)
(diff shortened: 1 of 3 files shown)


ipaserver/install/httpinstance.py
```diff
@@ -577,3 +581,17 @@ def start_tracking_certificates(self):
...
         db.track_server_cert(self.cert_nickname, self.principal,
                              db.passwd_fname, 'restart_httpd')
+
+    def request_service_keytab(self):
+        super(HTTPInstance, self).request_service_keytab()
+
+        if self.master_fqdn is not None:
+            service_dn = DN(('krbprincipalname', self.principal),
+                            ('cn', 'services'),
+                            ('cn', 'accounts'),
```
MartinBasti: IMHO `api.env.container_service` should be used

ipaserver/install/httpinstance.py
```diff
@@ -577,3 +581,17 @@ def start_tracking_certificates(self):
...
+                            self.suffix)
+
+            ldap_uri = ipaldap.get_ldap_uri(self.master_fqdn)
+            with ipaldap.LDAPClient(ldap_uri) as remote_ldap:
+                remote_ldap.gssapi_bind()
```
martbab: Will the SASL/GSSAPI bind work also in domain level 0? I am not quite sure we request any host TGT during replica install in this case. We should verify this case so that we do not break migration paths from older masters.

ipaserver/install/httpinstance.py
```diff
@@ -577,3 +581,17 @@ def start_tracking_certificates(self):
         db = certs.CertDB(self.realm, nssdir=paths.HTTPD_ALIAS_DIR)
         db.track_server_cert(self.cert_nickname, self.principal,
                              db.passwd_fname, 'restart_httpd')
+
+    def request_service_keytab(self):
+        super(HTTPInstance, self).request_service_keytab()
+
+        if self.master_fqdn is not None:
```
martbab: Maybe we can factor this check into a separate method and push it up to the parent class to always call it in `request_service_keytab`. We will thus have a generic solution for all service installers.

HonzaCholasta: Maybe we can, but not in 4.5, I don't want to risk breaking something else.

martbab: There is some risk involved so we can postpone this effort.

ipaserver/install/httpinstance.py
```diff
@@ -570,3 +576,16 @@ def start_tracking_certificates(self):
...
+            service_dn = DN(('krbprincipalname', self.principal),
+                            api.env.container_service,
+                            self.suffix)
+
+            ldap_uri = ipaldap.get_ldap_uri(self.master_fqdn)
+            with ipaldap.LDAPClient(ldap_uri) as remote_ldap:
+                remote_ldap.simple_bind(ipaldap.DIRMAN_DN, self.dm_password)
+                replication.wait_for_entry(remote_ldap, service_dn, timeout=60)
```
martbab: I am afraid this will not work in DL1 since directory manager password is randomly generated at the start of the replica and you cannot use it to connect to remote master (you will get ACIError) until the very end of replica install when Custodia pulls in the 'real' DM password from master.

IIRC http installer has `self.promote` attribute set during replica install, so you can use the good ole ugly promote hack:

```python
            with ipaldap.LDAPClient(ldap_uri) as remote_ldap:
                if self.promote:
                    remote_ldap.gssapi_bind()
                else:
                    remote_ldap.simple_bind(ipaldap.DIRMAN_DN, self.dm_password)
```

That should keep us covered for all cases (DL0, DL1, migrations). If you feel like doing a little bit of cleanup today then you can factor this part out to a separate method of parent class that returns the connection to master, but I do not insist on this improvement.

HonzaCholasta: Right. Also the simple bind should be done over TLS.

martbab:
master:

* ab71cd5a1693c221950bdfa9ffdfb99b9c317004 httpinstance: wait until the service entry is replicated


ipa-4-5:

* 9871bc08f8b8f51e2a05c4dfa18d844f9c141b8d httpinstance: wait until the service entry is replicated

PR #802, opened 2017-05-22
Improve cert messages some more + do that for KDC certs as well

stlaz:
Some of the previous error message handling would not work in
a locale different from English so we just output the error message we get
from `certutil`.

Also, since this was previously done, there is now kdc cert validation that is
different from the others so make the error messages more verbose there.

OpenSSL prints quite verbose messages so you may want to grab only a
part of it, let me know what your opinion is.

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -192,13 +192,17 @@ def verify_kdc_cert_validity(kdc_cert, ca_certs, realm):
 
         try:
             ipautil.run(
-                [OPENSSL, 'verify', '-CAfile', ca_file.name, kdc_file.name])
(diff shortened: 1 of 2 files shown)


stlaz:
I myself don't like having messages like `The server certificate in http.p12 is not valid: certutil: the certificate is invalid: <ERR>` as it sounds like `This office will not tolerate redundancy in this office here.` plus it mentions `certutil` which may other be good or bad, don't know how.
But I suppose it's safer to just put it over there and have users see what they can do, so alright.

martbab:
master:

* f827fe0f19596d29f9354368077fb43be2e16e8e cert-validate: keep all messages in cert validation
* bee3c1eccd44f7671a1455d12235bcbb910494b3 More verbose error message on kdc cert validation

PR #803, opened 2017-05-22
ipatests: add systemd journal collection for multihost tests

tomaskrizek:
Some messages are only logged in journal. Collection of journal
makes debugging failed tests from logs easier.

Fixes: https://pagure.io/freeipa/issue/6971

### DIFF
--- a/ipatests/pytest_plugins/integration/__init__.py
+++ b/ipatests/pytest_plugins/integration/__init__.py
@@ -24,6 +24,7 @@ from __future__ import print_function
 import os
 import tempfile
 import shutil
+import re
 
 import pytest
 from pytest_multihost import make_multihost_fixture
@@ -46,6 +47,14 @@ def pytest_addoption(parser):
         help="Directory to store integration test logs in.")
 
 
+def _get_logname_from_node(node):
+    name = node.nodeid
+    name = re.sub('\(\)/', '', name)      # remove ()/
+    name = re.sub('[()]', '', name)       # and standalone brackets
+    name = re.sub('(/|::)', '-', name)
+    return name
+
+
 def collect_test_logs(node, logs_dict, test_config):
     """Collect logs from a test
 
@@ -56,7 +65,7 @@ def collect_test_logs(node, logs_dict, test_config):
     :param test_config: Pytest configuration
     """
     collect_logs(
-        name=node.nodeid.replace('/', '-').replace('::', '-'),
+        name=_get_logname_from_node(node),
         logs_dict=logs_dict,
         logfile_dir=test_config.getoption('logfile_dir'),
         beakerlib_plugin=test_config.pluginmanager.getplugin('BeakerLibPlugin'),
--- a/ipatests/pytest_plugins/integration/__init__.py
+++ b/ipatests/pytest_plugins/integration/__init__.py
@@ -72,6 +72,36 @@ def collect_test_logs(node, logs_dict, test_config):
     )
 
 
+def collect_systemd_journal(node, hosts, test_config):
+    """Collect systemd journal from remote hosts
+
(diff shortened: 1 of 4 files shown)


ipatests/pytest_plugins/integration/__init__.py
```diff
@@ -55,13 +64,43 @@ def collect_test_logs(node, logs_dict, test_config):
...
+        # Get journal content
+        cmd = host.run_command(
+            ['journalctl', '--since', '-1h'],
+            log_stdout=False, raiseonerr=False)
+        if cmd.returncode:
+            log.error('An error occurred while collecting journal')
+            continue
+
+        # Write journal to file
+        with open(os.path.join(topdirname, "journal"), 'w') as f:
```
tomaskrizek: This puts the log to `test_name/hostname/journal`.
Other logs collected from the remote machine are in `test_name/hostname/var/log`.

Would it be better to place the journal to `test_name/hostname/var/log/messages`?

MartinBasti: No, unless you download the whole journal

tomaskrizek: Ok, my reasoning was the same.

Downloading the whole journal is definitely not going to happen. Even just the last hour is quite a lot of data (100KB when gzipped).

MartinBasti: I expected that you download only chosen services not everything. In this case it can be in /var/log/messages but basically I doesn't care if `messages` or `journal`.

ipatests/pytest_plugins/integration/__init__.py
```diff
@@ -55,13 +64,43 @@ def collect_test_logs(node, logs_dict, test_config):
...
+    for host in hosts:
+        log.info("Collecting journal from: %s", host.hostname)
+
+        topdirname = os.path.join(logfile_dir, name, host.hostname)
+        if not os.path.exists(topdirname):
+            os.makedirs(topdirname)
+
+        # Get journal content
+        cmd = host.run_command(
+            ['journalctl', '--since', '-1h'],
```
MartinBasti: We have tests running for more than 5 hours.

I don't like hard coded constants, could be this part of test config?

tomaskrizek: To clarify, the log collection is executed after each test class. While some test suites take 5 hours to finish, they are comprised of multiple test classes. One hour should be sufficient, especially considering that if the test fails, the failure will be among the most recent entries.

I'll investigate how difficult would it be to place this constant into test config.

tomaskrizek:
I've extracted the constant to test settings. I've also tried to customize it and it works.

ipatests/test_integration/test_testconfig.py
```diff
@@ -60,6 +61,8 @@
     "IPv6SETUP": "",
     "IPADEBUG": "",
     "DOMAINLVL": str(MAX_DOMAIN_LEVEL),
+    "SYSTEMD_PAGER": "cat",  # allows journalctl log collection
```
MartinBasti: You added this only into test, so test will fail because nothing was added to config code.

tomaskrizek: Removed, it works even without this env variable set.

MartinBasti:
needs rebase

tomaskrizek:
@MartinBasti Done.

MartinBasti:
master:

* 906c4c94590455c41603e5eca869b64c8d86d4f5 ipatests: change logdir naming pattern for multihost tests
* 48b7e83511b80b3a1e0416134496ac37244f1073 ipatests: add systemd journal collection for multihost tests

PR #805, opened 2017-05-22
Fix rare race condition with missing ccache file

simo5:
In some circumstances the ccache file may disappear while
mod_auth_gssapi still has a valid cookie and the client is performing a
json server call.

This may lead to credentials getting sourced from the keytab.
Make sure we enforce what GSS NAME we want to resolve so HTTP creds are
never mistakenly sourced.

Ticket: #6972

### DIFF
--- a/ipaserver/rpcserver.py
+++ b/ipaserver/rpcserver.py
@@ -777,8 +777,17 @@ class jsonserver_session(jsonserver, KerberosSession):
             self.debug('no ccache, need login')
             return self.need_login(start_response)
 
+        # If we have a ccache, make sure we have a GSS_NAME and use
+        # it to resolve the ccache name (Issue: 6972 )
+        principal = environ.get('GSS_NAME')
+        if principal is None:
+            self.debug('no GSS Name, need login')
+            return self.need_login(start_response)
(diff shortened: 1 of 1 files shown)


abbra:
LGTM

ipaserver/rpcserver.py
```diff
@@ -777,8 +777,17 @@ def __call__(self, environ, start_response):
             self.debug('no ccache, need login')
             return self.need_login(start_response)
 
+        # If we have a ccache, make sure we have a GSS_NAME and use
+        # it to resolve the ccache name (Issue: XXXX )
```
MartinBasti: Please, put proper issue number there.

simo5: Done, thanks for catching this

stlaz:
Readding the ACK based on @frozencemetery's approval and on the fact that this was just a comment issue.

MartinBasti:
master:

* 83619e804b53f2bfb8b0d94882c01abed23c4a9f Fix rare race condition with missing ccache file


ipa-4-5:

* 90b432b93d71c795eaa4f92d3a9e1d322dce1802 Fix rare race condition with missing ccache file

PR #810, opened 2017-05-24
user.py: replace user_mod with ldap.update_entry()

Tiboris:
Refactoring user_add class to use 'ldap.update_entry()' call
    instead of api call 'user_mod' when --noprivate option is used.

    https://pagure.io/freeipa/issue/5788

### DIFF
--- a/ipaserver/plugins/user.py
+++ b/ipaserver/plugins/user.py
@@ -598,22 +598,16 @@ class user_add(baseuser_add):
         except errors.AlreadyGroupMember:
             pass
 
-        # delete description attribute NO_UPG_MAGIC if present
-        if options.get('noprivate', False):
-            if not options.get('all', False):
-                desc_attr = ldap.get_entry(dn, ['description'])
-                entry_attrs.update(desc_attr)
-            if 'description' in entry_attrs and NO_UPG_MAGIC in entry_attrs['description']:
-                entry_attrs['description'].remove(NO_UPG_MAGIC)
-                kw = {'setattr': unicode('description=%s' % ','.join(entry_attrs['description']))}
-                try:
-                    self.api.Command['user_mod'](keys[-1], **kw)
-                except (errors.EmptyModlist, errors.NotFound):
-                    pass
-
         # Fetch the entry again to update memberof, mep data, etc updated
         # at the end of the transaction.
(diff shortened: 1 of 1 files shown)


ipaserver/plugins/user.py
```diff
@@ -603,11 +603,14 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
...
                 entry_attrs.update(desc_attr)
+                entry_attrs._orig['description'] = list(
+                    desc_attr['description'])  # The entry_attrs.update()
+                # does not change the entry_attrs._orig
             if 'description' in entry_attrs and NO_UPG_MAGIC in entry_attrs['description']:
                 entry_attrs['description'].remove(NO_UPG_MAGIC)
-                kw = {'setattr': unicode('description=%s' % ','.join(entry_attrs['description']))}
                 try:
-                    self.api.Command['user_mod'](keys[-1], **kw)
+                    self._exc_wrapper(
```
MartinBasti: is really needed to call _exc_wrapper?  ldap.update_entry(entry) should return the same exceptions. IMO it is too complex with _exc_wrapper.

ipaserver/plugins/user.py
```diff
@@ -603,11 +603,14 @@ def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
             if not options.get('all', False):
                 desc_attr = ldap.get_entry(dn, ['description'])
                 entry_attrs.update(desc_attr)
+                entry_attrs._orig['description'] = list(
```
MartinBasti: How about to move this whole block 20 lines below, where `newentry = ldap.get_entry(dn, ['*'])` is. IIUC that call will download the whole user_entry, so you can just delete `NO_UPG_MAGIC` if needed and you even save one more LDAP operation without ugly hacking of LDAPEntry.

stlaz:
+ please don't indent the commit message.

MartinBasti:
master:

* d73ec06cb3c8042f90d1520485aa9b0f2e7c8fbe user.py: replace user_mod with ldap.update_entry()


ipa-4-5:

* daeac31ba54d88159b46d61fc443ccd45902f8f2 user.py: replace user_mod with ldap.update_entry()

PR #811, opened 2017-05-24
[4.5] Remove pkinit-anonymous command

stlaz:
Ever since from v4.5, FreeIPA expects at least some kind of
anonymous PKINIT to work. The pkinit-anonymous command was supposed
to enable/disable anonymous pkinit by locking/unlocking the
anonymous principal. We can't allow this for FreeIPA to work
so we are removing the command as it was never supported anyway.

https://pagure.io/freeipa/issue/6936


martbab:
ipa-4-5:

* 4e878c3dc6f72cae4e7b4cb2ef45f2f4e91ac287 Remove pkinit-anonymous command

PR #812, opened 2017-05-24
Refactoring cert-find to use API call directly instead of using

felipevolpone:
Refactoring cert-find to use API calls directly instead of using raw LDAP search.

Upstream ticket: https://pagure.io/freeipa/issue/6948

I removed the raw LDAP search and used the API directly. In the old code, the call ` self.obj._owners()` returns `service, hots and user`. However, when testing the code, only the service was being used, so I made it only use the service API. 

If there another scenario where `user and host` are used, I thought to do something like:

```python
for owner in self.obj._owners():
    api_name = owner.name
    response = api.Command[api_name+'_find'](options[api_name])
    ...  # continues
```
Is that correct?

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -1090,8 +1090,8 @@ class cert(BaseCertObject):
                 param = param.clone(flags=param.flags - {'no_search'})
             yield param
 
-        for owner in self._owners():
-            yield owner.primary_key.clone_rename(
+        for owner, search_key in self._owners():
+            yield search_key.clone_rename(
                 'owner_{0}'.format(owner.name),
                 required=False,
                 multivalue=True,
@@ -1101,15 +1101,22 @@ class cert(BaseCertObject):
             )
 
     def _owners(self):
-        for name in ('user', 'host', 'service'):
-            yield self.api.Object[name]
+        for obj_name, search_key in [('user', None),
+                                     ('host', None),
+                                     ('service', 'krbprincipalname')]:
+            obj = self.api.Object[obj_name]
+            if search_key is None:
+                pkey = obj.primary_key
+            else:
+                pkey = obj.params[search_key]
+            yield obj, pkey
 
     def _fill_owners(self, obj):
         dns = obj.pop('owner', None)
         if dns is None:
             return
 
-        for owner in self._owners():
+        for owner, _search_key in self._owners():
             container_dn = DN(owner.container_dn, self.api.env.basedn)
             name = 'owner_' + owner.name
             for dn in dns:
@@ -1373,8 +1380,8 @@ class cert_find(Search, CertMethod):
                 option = option.clone(default=None, autofill=None)
             yield option
 
-        for owner in self.obj._owners():
-            yield owner.primary_key.clone_rename(
+        for owner, search_key in self.obj._owners():
+            yield search_key.clone_rename(
                 '{0}'.format(owner.name),
                 required=False,
                 multivalue=True,
@@ -1385,7 +1392,7 @@ class cert_find(Search, CertMethod):
                      owner.object_name_plural),
                 label=owner.object_name,
             )
-            yield owner.primary_key.clone_rename(
+            yield search_key.clone_rename(
                 'no_{0}'.format(owner.name),
                 required=False,
                 multivalue=True,
@@ -1504,7 +1511,7 @@ class cert_find(Search, CertMethod):
         ldap = self.api.Backend.ldap2
 
         filters = []
-        for owner in self.obj._owners():
+        for owner, search_key in self.obj._owners():
             for prefix, rule in (('', ldap.MATCH_ALL),
                                  ('no_', ldap.MATCH_NONE)):
                 try:
@@ -1520,7 +1527,7 @@ class cert_find(Search, CertMethod):
                     filters.append(filter)
 
                 filter = ldap.make_filter_from_attr(
-                    owner.primary_key.name,
+                    search_key.name,
                     value,
                     rule)
                 filters.append(filter)


simo5:
Won't this cause it to not find certificates associated to users ?
Currently that works, this change is not replicating the same functionality of the original code.

HonzaCholasta:
@simo5, it will. Same for hosts.

@felipevolpone, one possible correct fix would be to modify `cert._owners` to:
```python
    def _owners(self):
        for obj_name, pkey_name in [('user', None),
                                    ('host', None),
                                    ('service', 'krbprincipalname')]:
            obj = self.api.Object[obj_name]
            if pkey_name is None:
                pkey = obj.primary_key
            else:
                pkey = obj.params[pkey_name]
            yield obj, pkey
```
and replace all occurences of:
```python
    for owner in (...)._owners():
        ... owner.primary_key ...
```
with:
```python
    for owner, owner_pkey in (...)._owners():
        ... owner_pkey ...
```

stlaz:
```
************* Module ipaserver.plugins.cert
ipaserver/plugins/cert.py:1505: [C0303(trailing-whitespace), ] Trailing whitespace)
ipaserver/plugins/cert.py:1506: [W0312(mixed-indentation), ] Found indentation with tabs instead of spaces)
ipaserver/plugins/cert.py:1508: [C0303(trailing-whitespace), ] Trailing whitespace)
ipaserver/plugins/cert.py:1512: [C0303(trailing-whitespace), ] Trailing whitespace)
ipaserver/plugins/cert.py:1513: [W0312(mixed-indentation), ] Found indentation with tabs instead of spaces)
ipaserver/plugins/cert.py:1514: [W0312(mixed-indentation), ] Found indentation with tabs instead of spaces)
ipaserver/plugins/cert.py:1515: [W0312(mixed-indentation), ] Found indentation with tabs instead of spaces)
ipaserver/plugins/cert.py:1516: [W0312(mixed-indentation), ] Found indentation with tabs instead of spaces)
ipaserver/plugins/cert.py:1517: [W0312(mixed-indentation), ] Found indentation with tabs instead of spaces)
ipaserver/plugins/cert.py:1519: [W0312(mixed-indentation), ] Found indentation with tabs instead of spaces)
ipaserver/plugins/cert.py:1520: [W0312(mixed-indentation), ] Found indentation with tabs instead of spaces)
ipaserver/plugins/cert.py:1516: [E1101(no-member), cert_find._service_search] Instance of 'cert_find' has no '_get_cert_obj' member)
ipaserver/plugins/cert.py:1516: [E0602(undefined-variable), cert_find._service_search] Undefined variable 'raw')
ipaserver/plugins/cert.py:1504: [W0612(unused-variable), cert_find._service_search] Unused variable 'ldap')
```
```
PEP-8 errors:
./ipaserver/plugins/cert.py:1505:1: W293 blank line contains whitespace
./ipaserver/plugins/cert.py:1506:1: E101 indentation contains mixed spaces and tabs
./ipaserver/plugins/cert.py:1506:1: W191 indentation contains tabs
./ipaserver/plugins/cert.py:1507:1: E101 indentation contains mixed spaces and tabs
./ipaserver/plugins/cert.py:1508:1: W293 blank line contains whitespace
./ipaserver/plugins/cert.py:1512:1: W293 blank line contains whitespace
./ipaserver/plugins/cert.py:1513:1: E101 indentation contains mixed spaces and tabs
./ipaserver/plugins/cert.py:1513:1: W191 indentation contains tabs
./ipaserver/plugins/cert.py:1514:1: W191 indentation contains tabs
./ipaserver/plugins/cert.py:1514:2: E101 indentation contains mixed spaces and tabs
./ipaserver/plugins/cert.py:1515:1: W191 indentation contains tabs
./ipaserver/plugins/cert.py:1515:2: E101 indentation contains mixed spaces and tabs
./ipaserver/plugins/cert.py:1516:1: W191 indentation contains tabs
./ipaserver/plugins/cert.py:1516:2: E101 indentation contains mixed spaces and tabs
./ipaserver/plugins/cert.py:1517:1: W191 indentation contains tabs
./ipaserver/plugins/cert.py:1517:2: E101 indentation contains mixed spaces and tabs
./ipaserver/plugins/cert.py:1519:1: W191 indentation contains tabs
./ipaserver/plugins/cert.py:1519:2: E101 indentation contains mixed spaces and tabs
./ipaserver/plugins/cert.py:1520:1: W191 indentation contains tabs
```
https://i.imgflip.com/1pol0g.jpg
This does not even compile.

felipevolpone:
Talking with @simo5 on IRC, he proposed to we do not change the whole code, but only fix it doing something like:

```python
filter = ldap.make_filter_from_attr('krbprincipalname', value, rule)
```
in https://github.com/freeipa/freeipa/blob/master/ipaserver/plugins/cert.py#L1523

simo5:
Ok one thing was in the back of my mind and came up now, we need to keep in mind that krbprincipalname can be multivalued. It won't affect this case (I think) but if we are going to expose some new attribute we need to make sure users of the API are not tricked into thinking it is alays the service's unique name (that's what krbCanonicalName is for).

simo5:
So Iam for the very localized change still (to be clear)

martbab:
Remember taht you have to use 'exact=False' in the filter to perform substring search for krbPrincipalName given the fact that (except for services) the principal is constructed from primary key by appending realm (and prepending `host/` in the case of hosts). This, however, opens a range of possibilities for new bug to creep in (considering 'tuser' is the owner but we have 'tuser1' and 'tuser2' in LDAP, what will your search filter return?).

That's why I think this is not correct solution given we currently reference owners by primary keys and not by principals (krbPrincipalName != primary key in most cases except services without krbCanonicalName attribute). I am more inclined to @HonzaCholasta's solution as it seems cleaner to me. An alternative is to report principals as cert owners, which will break API, however.

felipevolpone:
@HonzaCholasta thanks for the help, but the idea isn't avoid calling the ldap directly, and instead of that call the APIs?

If we change _only_ the `_owners` method, we'll still have ldap calls here: https://github.com/freeipa/freeipa/blob/master/ipaserver/plugins/cert.py#L1515-L1545 . Am I missing something?

HonzaCholasta:
@felipevolpone, that is a bad idea. Calling the API instead of doing a direct LDAP search would degrade performace (currently everything is done in a single LDAP search, with API calls it will be *at least* one LDAP search per owner class) and offers less flexibility (the current code allows you to find *any* LDAP entry which refers to a certificate, with API calls you are limited to whatever is defined in the API).

The PR currently breaks the `--user` and `--host` options, because they no longer expect a user name and host name, but principal names (as @martbab already pointed out).

felipevolpone:
@HonzaCholasta thank you for the explanation, I misunderstood the ticket title. I did the changes that you suggested.

ipaserver/plugins/cert.py
```diff
@@ -1101,15 +1101,22 @@ def get_params(self):
...
+                pkey = obj.params[pkey_name]
+            yield obj, pkey
 
     def _fill_owners(self, obj):
         dns = obj.pop('owner', None)
         if dns is None:
             return
 
-        for owner in self._owners():
+        for owner, __ in self._owners():
```
felipevolpone: I'm used to use `_` to variables that are not used, but since `_` is a function, I used `__`. Not sure if this is the right approach.

martbab: @felipevolpone a common practice in FreeIPA is to just use normal variable name prefixed with '_', as in:

```python
for owner, _pkey in self._owners():
```

Pylint should not report these as unused vars.

ipaserver/plugins/cert.py
```diff
@@ -1101,15 +1101,22 @@ def get_params(self):
             )
 
     def _owners(self):
-        for name in ('user', 'host', 'service'):
-            yield self.api.Object[name]
+        for obj_name, pkey_name in [('user', None),
```
felipevolpone: Fair enough :+1:

felipevolpone:
Done. @frasertweedale if there is something wrong with the commit message, please tell me.
Thanks for reviewing :)

MartinBasti:
master:

* 44bd5e358b027f8956b730f250854efb5087f05e Changing cert-find to do not use only primary key to search in LDAP.


ipa-4-5:

* df1276e9daf9ee8db05538f47706855cb3d11e01 Changing cert-find to do not use only primary key to search in LDAP.

PR #813, opened 2017-05-25
Add Subject Key Identifier to CA cert validity check

frasertweedale:
CA certificates MUST have the Subject Key Identifier extension to
facilitiate certification path construction.  Not having this
extension on the IPA CA certificate will cause failures in Dogtag
during signing; it tries to copy the CA's Subject Key Identifier to
the new certificate's Authority Key Identifier extension, which
fails.

When installing an externally-signed CA, check that the Subject Key
Identifier extension is present in the CA certificate.

Fixes: https://pagure.io/freeipa/issue/6976

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -717,6 +717,12 @@ class NSSDatabase(object):
             raise ValueError("not a CA certificate")
 
         try:
+            cert.extensions.get_extension_for_class(
+                    cryptography.x509.SubjectKeyIdentifier)
+        except cryptography.x509.ExtensionNotFound:
+            raise ValueError("missing subject key identifier extension")
+
+        try:
             self.run_certutil(['-V', '-n', nickname, '-u', 'L'],
                               capture_output=True)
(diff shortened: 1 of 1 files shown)


stlaz:
I will review this but we also need a patch for 4.5 which unfortunately still uses python-nss for certificate validation.

stlaz:
Works for me, 4.5 patch is needed with ^-- in mind.

frasertweedale:
@stlaz I don't think backport to 4.5 is essential.  This issue is something that should be
rare in practice, i.e. if a proper CA implementation signs the CSR (e.g.
MS/AD-CS) the SKI extension will be there.

stlaz:
@frasertweedale I guess you're right. We should therefore not triage it for 4.5 since it's just a rare usability bug which will probably not appear in production. Thanks for the patch and reasoning :)

MartinBasti:
master:

* bc6d4995144505c45a62320c71f503b54f68a962 Add Subject Key Identifier to CA cert validity check

PR #814, opened 2017-05-25
Add new permission to grant 'add' on cas container

Tiboris:
https://pagure.io/freeipa/issue/6609

### DIFF
--- a/ipaserver/plugins/ca.py
+++ b/ipaserver/plugins/ca.py
@@ -138,6 +138,12 @@ class ca(LDAPObject):
             ],
             'default_privileges': {'CA Administrator'},
         },
+        'System: Grant add on CAs container': {
+            'ipapermtargetfilter': {'(objectclass=nscontainer)'},
+            'ipapermbindruletype': 'permission',
+            'ipapermright': {'add'},
+            'default_privileges': {'CA Administrator'},
+        },
         'System: Delete CA': {
             'ipapermright': {'delete'},
             'replaces': [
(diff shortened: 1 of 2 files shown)


ipaserver/install/plugins/update_managed_permissions.py
```diff
@@ -291,6 +291,14 @@
             'ipadomainlevel', 'objectclass',
         },
     },
+    'System: Grant add on CAS container': {
```
MartinBasti: IMO this should be kept together with other sub CA related permissions in `ipaserver/plugins/ca.py`

ipaserver/plugins/ca.py
```diff
@@ -138,6 +138,19 @@ class ca(LDAPObject):
             ],
             'default_privileges': {'CA Administrator'},
         },
+        'System: Grant add on CAS container': {
+            'ipapermtargetfilter': {'(objectclass=nscontainer)'},
+            'ipapermbindruletype': 'permission',
+            'ipapermright': {'add'},
+            'replaces':
```
MartinBasti: You can remove `replaces`, there is no old ACI that should be removed

ipaserver/plugins/ca.py
```diff
@@ -138,6 +138,19 @@ class ca(LDAPObject):
             ],
             'default_privileges': {'CA Administrator'},
         },
+        'System: Grant add on CAS container': {
```
MartinBasti: I prefer to keep consistency with other privileges, please call it 'System: Grant add on CAs container'

ipaserver/plugins/ca.py
```diff
@@ -138,6 +138,12 @@ class ca(LDAPObject):
             ],
             'default_privileges': {'CA Administrator'},
         },
+        'System: Grant add on CAs container': {
+            'ipapermtargetfilter': {'(objectclass=nscontainer)'},
+            'ipapermbindruletype': 'permission',
+            'ipapermright': {'add'},
+            'default_privileges': {'CA Administrator'},
+        },
```
frasertweedale: I think we do not want to add a new permission for this; the existing 'System: Add CA' permission
should be sufficient.

frasertweedale:
This problem is actually caused by a bug in a preliminary permission check during
`ca_add.pre_callback`, not an ACI issue.  I am looking into it, but this is not the
solution, so I am NACKing it.

frasertweedale:
New PR: https://github.com/freeipa/freeipa/pull/837

PR #816, opened 2017-05-25
only stop/disable simple service if it is installed

martbab:
The SimpleServiceInstance uninstaller assument that the service to
uninstall was always present on the system. This may not be valid in
some cases (e.g. containerized deployments) and thus we need to change
the service state only when we know that the unit file exists.

https://pagure.io/freeipa/issue/6977


martbab:
ipa-4-5:

* 6114150de20a7d8371c7383f619cd0fefe339cbf only stop/disable simple service if it is installed


master:

* 8b6f8ed7d47542b9bd8b7453a8a0e202ed1db97d only stop/disable simple service if it is installed

PR #817, opened 2017-05-25
[py3] Change ConfigParser to RawConfigParser

stlaz:
In case ipa_generate_password() generates a sequence containing
'%', ConfigParser.set() will fail because it would think it is a
string that should be interpolated.


martbab:
master:

* 35675ca2bbe9c044f115764a2daac45f7468be00 Change ConfigParser to RawConfigParser

PR #818, opened 2017-05-26
Avoid possible endless recursion in RPC call from client

stlaz:
This commit removes recursion which may lack end condition.

https://pagure.io/freeipa/issue/6796

===============================
This is my try to fix ^--. The methods and their arguments in the module are very poorly documented (they are not documented), so I just hope I can initialize the variables used in logs prior to the cycle.

I also think this actually relates more to https://pagure.io/freeipa/issue/6775 than the ticket mentioned here and that the person "fixing" https://pagure.io/freeipa/issue/6796 messed up the tickets but it's hopefully OK to fix it this way.

Aside from just moving everything into a cycle, I also improved logging a bit and had same error handling for different errors merged into one `except` block (`SSLError`, `socket.error`).

### DIFF
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -1088,50 +1088,63 @@ class RPCClient(Connectible):
         :param kw: Keyword arguments to pass to remote command.
         """
         server = getattr(context, 'request_url', None)
-        self.log.info("Forwarding '%s' to %s server '%s'",
-                      name, self.protocol, server)
         command = getattr(self.conn, name)
         params = [args, kw]
-        try:
-            return self._call_command(command, params)
-        except Fault as e:
-            e = decode_fault(e)
-            self.debug('Caught fault %d from server %s: %s', e.faultCode,
-                server, e.faultString)
-            if e.faultCode in errors_by_code:
-                error = errors_by_code[e.faultCode]
-                raise error(message=e.faultString)
-            raise UnknownError(
-                code=e.faultCode,
-                error=e.faultString,
-                server=server,
-            )
-        except SSLError as e:
-            raise NetworkError(uri=server, error=str(e))
-        except ProtocolError as e:
-            # By catching a 401 here we can detect the case where we have
-            # a single IPA server and the session is invalid. Otherwise
-            # we always have to do a ping().
-            session_cookie = getattr(context, 'session_cookie', None)
-            if session_cookie and e.errcode == 401:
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @stlaz,
thank you for the patch. The fix will avoid recursion in forward(), but I think that the same type of recursion could happen in create_connection().

stlaz:
Thanks for the good catch, @flo-renaud. While the recursion in the `forward()` method was quite easy to fix, it was not so in the `create_connection()`. I tried to do several improvements to the code on the way and I am submitting it to testing here.

stlaz:
The tests passed, I am just adding a minor comment typo fixup.

ipalib/rpc.py
```diff
@@ -999,77 +999,77 @@ def create_connection(self, ccache=None, verbose=None, fallback=None,
...
+                        # Unauthorized. Remove the session and try again.
+                        delattr(context, 'session_cookie')
+                        try:
+                            delete_persistent_client_session_data(principal)
+                        except Exception:
+                            # This shouldn't happen if we have a session but
+                            # it isn't fatal.
+                            pass
+                        # try once more with a new session
+                        break
```
flo-renaud: This break is related to the loop on urls. This means that when reached, we exit from the loop on urls and retry with the first url. This is not optimal if the first server is down as it will be retried many times. I would rather put the loops in the reverse order, i.e. when the session_cookie needs to be deleted, retry with the same url.

stlaz: I was following the original implementation order but you're right that it makes much more sense your proposed way. I'll improve it right away.

flo-renaud:
Hi @stlaz 
Thank you, the patch looks good to me.

MartinBasti:
master:

* 81a808caeb5676427610e113b5a259511c2835d6 Avoid possible endless recursion in RPC call
* 79d1752577e8fcb568b701509fe5b52f949d5e4b rpc: preparations for recursion fix
* e1f8684e858b4ae47b54acd0d76a844bc20ce443 rpc: avoid possible recursion in create_connection


ipa-4-5:

* a5b413b72e224120acde09d1c877be11b3f61b6b Avoid possible endless recursion in RPC call
* d8aab383a39a22cc613cf64e5d66ce69111d97df rpc: preparations for recursion fix
* cb6c93dad044c724ba2cedbff49bf71aea939418 rpc: avoid possible recursion in create_connection

PR #819, opened 2017-05-26
[ipa-4-5] Change python-cryptography to python2-cryptography

pvomacka:
Package name is python2-cryptography and even that it Provides
python-cryptography package, it causes problems during update of IPA
on RHEL - python2-cryptography is not updated. After changing required package
name to python2-cryptography upgrade on RHEL works well.

Fixes: https://pagure.io/freeipa/issue/6749


martbab:
ipa-4-5:

* 14ff94a0d481051613338a512260b6a473671538 Change python-cryptography to python2-cryptography

PR #820, opened 2017-05-26
Amend some regressions in backup/restore tests

martbab:
After investigating the backup_restore test failures in CI I found out that
most of them were caused by recently introduced changes in 4.5 development and
are quite easy to fix.

The last one, however, is caused by some bug in Vault and may not be as easy as
it seems.

https://pagure.io/freeipa/issue/6956


MartinBasti:
master:

* 2624cf2e4cc4b4b560fe64688ee55160f9d8fb80 test_backup_restore: do not fail on missing KrbLastSuccessfulAuth
* 1e5f55e791ac3665d4db0f6d14995491f11ef887 Do not delete DS and PKI users during backup/restore tests

PR #821, opened 2017-05-26
fix incorrect suffix handling in topology checks

martbab:
When trying to delete a partially removed master entry lacking
'iparepltopomanagedsuffix' attribute, the code that tries to retrieve
tha value for further computations passes None and causes unhandled
internal errors.

If the attribute is empty or not present, we should return empty list
instead as to not break calling cod attribute, the code that tries to
retrieve tha value for further computations passes None and causes
unhandled internal errors. We should return empty list instead.

https://pagure.io/freeipa/issue/6965

### DIFF
--- a/ipaserver/topology.py
+++ b/ipaserver/topology.py
@@ -72,12 +72,15 @@ def get_topology_connection_errors(graph):
 
 def map_masters_to_suffixes(masters):
     masters_to_suffix = {}
+    managed_suffix_attr = 'iparepltopomanagedsuffix_topologysuffix'
 
     for master in masters:
-        try:
-            managed_suffixes = master.get(
-                'iparepltopomanagedsuffix_topologysuffix')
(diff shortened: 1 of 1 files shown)


pvoborni:
can this happen?
```
>>> d = {'a': None}
>>> r = d.get('a', [])
>>> print r
None
```

And thus fail with the same error?

martbab:
@pvoborni it shouldn't but given how our framework sometimes (mis)-behaves the possibility is there.

felipevolpone:
Tested, works as advertised.

martbab:
ipa-4-5:

* d651a9877d0e2f9dd1b057630508b488678bb86e fix incorrect suffix handling in topology checks


master:

* 8ef4888af77f8e6fd8324297d26287b575b18163 fix incorrect suffix handling in topology checks

PR #824, opened 2017-05-29
ca-add: validate Subject DN name attributes

frasertweedale:
If the Subject DN is syntactically valid but contains unrecognised
name attributes, FreeIPA accepts it but Dogtag rejects it, returning
status 400 and causing the framework to raise RemoteRetrieveError.

Update the ca-add command to perform some additional validation on
the user-supplied Subject DN, making sure that we recognise all the
attributes.

Fixes: https://pagure.io/freeipa/issue/6987

### DIFF
--- a/ipaserver/plugins/ca.py
+++ b/ipaserver/plugins/ca.py
@@ -4,9 +4,12 @@
 
 import base64
 
+import six
+
 from ipalib import api, errors, output, Bytes, DNParam, Flag, Str
 from ipalib.constants import IPA_CA_CN
 from ipalib.plugable import Registry
+from ipapython.dn import ATTR_NAME_BY_OID
 from ipaserver.plugins.baseldap import (
     LDAPObject, LDAPSearch, LDAPCreate, LDAPDelete,
     LDAPUpdate, LDAPRetrieve, LDAPQuery, pkey_to_value)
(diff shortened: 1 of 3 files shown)


felipevolpone:
I tested and it fixes the error.

I'm not sure if it's relevant, but maybe the unrecognized attributes could be printed as the user wrote them. 
This:
```ipa: ERROR: invalid 'Subject DN': Unrecognized attributes: dn```
Could be:
```ipa: ERROR: invalid 'Subject DN': Unrecognized attributes: DN```

Good to go.

**Comment by frasertweedale** [comment]
@felipevolpone I did think about that.  I agree it would be nice but IMO it adds unnecessary
implementation complexity, for handling a probably-rare user error.

File: ipaserver/plugins/ca.py [hunk 1]
```diff
@@ -4,9 +4,12 @@
 
 import base64
 
+import six
+
 from ipalib import api, errors, output, Bytes, DNParam, Flag, Str
 from ipalib.constants import IPA_CA_CN
 from ipalib.plugable import Registry
+from ipapython.dn import _ATTR_NAME_BY_OID
```
stlaz: I suppose we should remove the prepending underscore as we're importing it.

frasertweedale: @stlaz fair enough, I'll update PR soon.

stlaz:
Thanks for addressing my comment. @felipevolpone seems to be OK with the rest of the PR so I am adding the official ACK.

MartinBasti:
master:

* 5f0e13ce9c3d1ead02de61a148de973fc6787b96 ca-add: validate Subject DN name attributes

PR #827, opened 2017-05-29
pylint: explicitly depends on python2-pylint

MartinBasti:
F26 defaults to python3 with pylint package, we have to explicitly ask
for python2 version of pylint

https://pagure.io/freeipa/issue/6986

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -170,7 +170,11 @@ BuildRequires:  samba-python
 # 1.6: x509.Name.rdns (https://github.com/pyca/cryptography/issues/3199)
 BuildRequires:  python2-cryptography >= 1.6
 BuildRequires:  python-gssapi >= 1.2.0
+%if 0%{?fedora} >= 26
+BuildRequires:  python2-pylint
+%else
 BuildRequires:  pylint >= 1.6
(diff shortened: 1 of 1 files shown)


MartinBasti:
Works only for F26. F25 doesn't have python2-pylint package

tiran:
```
%if 0%{?fedora} >= 26
BuildRequires:  python2-pylint
%else
BuildRequires:  python-pylint >= 1.6
%endif
```

MartinBasti:
@tiran yes it is possible, but given that F27 is a target platform for FreeIPA 4.6 and more package changes will be required, I'd drop F25 and continue with developing on F26 and avoid too many `if` in code

tiran:
@MartinBasti I'm still developing on F25 because I'm reluctant to upgrade my working machine to F26 until F26 has been released and stabilized. Can we compromise and keep master F25 compatibility for a while?

MartinBasti:
@tiran OK, given that pylint is broken on F26

MartinBasti:
master:

* be1415b6cc8f5dadc1ac3766305a33f370fdf9bb pylint: explicitly depends on python2-pylint

PR #829, opened 2017-05-30
client.py: Replace hardcoded 'admin' with options.principal

Tiboris:
Fixes: https://pagure.io/freeipa/issue/5406

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -2907,6 +2907,14 @@ def _install(options):
 
         # Check that nss is working properly
         if not options.on_master:
+            user = options.principal
+            if user is None:
+                user = "admin@%s" % cli_domain
+                root_logger.info("Principal is not set when enrolling with OTP"
+                                 "; using principal '%s' for 'getent passwd'",
+                                 user)
+            elif '@' not in user:
+                user = "%s@%s" % (user, cli_domain)
             n = 0
             found = False
             # Loop for up to 10 seconds to see if nss is working properly.
@@ -2915,16 +2923,15 @@ def _install(options):
             # Particulary, SSSD might take longer than 6-8 seconds.
             while n < 10 and not found:
                 try:
-                    ipautil.run(["getent", "passwd", "admin@%s" % cli_domain])
+                    ipautil.run(["getent", "passwd", user])
                     found = True
                 except Exception as e:
                     time.sleep(1)
                     n = n + 1
 
             if not found:
-                root_logger.error(
-                    "Unable to find 'admin' user with "
-                    "'getent passwd admin@%s'!" % cli_domain)
+                root_logger.error("Unable to find '%s' user with 'getent "
+                                  "passwd %s'!" % (user.split("@")[0], user))
                 if conf:
                     root_logger.info("Recognized configuration: %s", conf)
                 else:


ipaclient/install/client.py
```diff
@@ -2915,16 +2921,17 @@ def _install(options):
             # Particulary, SSSD might take longer than 6-8 seconds.
             while n < 10 and not found:
                 try:
-                    ipautil.run(["getent", "passwd", "admin@%s" % cli_domain])
+                    ipautil.run(["getent", "passwd", "%s@%s"
+                                % (principal, cli_domain)])
```
flo-renaud: The man page for ipa-client-install specifies that -p is used to provide a kerberos principal. This means that options.principal can have the format username or username@DOMAIN.COM.
If the latter format is provided, the code will call getent passwd username@DOMAIN.COM@domain.com and fail.

Tiboris: Sure, thanks.

Tiboris:
Hi @flo-renaud 
I am really sorry but I think issue [5060](https://pagure.io/freeipa/issue/5060) is related to the  ipa-client-automount and it is closed as fixed.

ipaclient/install/client.py
```diff
@@ -2907,6 +2907,14 @@ def _install(options):
 
         # Check that nss is working properly
         if not options.on_master:
+            user = options.principal
+            if user is None:
+                user = "admin"
+                root_logger.info("Principal is not set when enrolling with OTP"
+                                 "; using principal '%s' for 'getent passwd'",
+                                 principal)
```
abbra: Should it be 'user', not 'principal' here?

abbra:
See additional comments inline.

ipaclient/install/client.py
Tiboris: Yes, thanks.

flo-renaud:
Hi @Tiboris 
ticket 5060 was related to ipa-replica-install. You can see that ipa-client-automount is still using hardcoded "admin" value in line [88](https://pagure.io/freeipa/blob/master/f/client/ipa-client-automount#_88), but it is just used to ensure that SSSD is up. As ipa-client-automount does not prompt for any kerberos principal, I guess we can leave this part as is.

Tiboris:
Thank you, i missed it...

Tiboris:
Hello @flo-renaud 
do I have to create new ticket for it or the ipa-client-automount will remain as it is?

flo-renaud:
Hi @Tiboris 
I think we can leave ipa-client-automount as it is.

ipaclient/install/client.py
```diff
@@ -2907,6 +2907,14 @@ def _install(options):
...
         # Check that nss is working properly
         if not options.on_master:
+            user = options.principal
+            if user is None:
+                user = "admin"
+                root_logger.info("Principal is not set when enrolling with OTP"
+                                 "; using principal '%s' for 'getent passwd'",
+                                 user)
+            else:
+                user = user.split("@")[0]
```
flo-renaud: Hi @Tiboris ,
I still have a nitpick. With this call to split, the code is actually loosing some information. For instance, if the options.principal was myuser@bogusdomain.com we end up with user = myuser and run getent passwd myuser@ipadomain.com (ie we replace bogusdomain with ipadomain.com).
I would rather do the opposite, i.e make sure that the "user" variable contains a kerberos id with the format name@domain (if it contains only name, then add @cli_domain) or admin@cli_domain, and call "getent passwd user"  instead of "getent passwd user@cli_domain".

Tiboris: Thank you! 
I hopefully understand your concerns.

flo-renaud:
Hi @Tiboris 
thank you, works as expected.

Tiboris:
Hi @flo-renaud 
Thanks too.

MartinBasti:
master:

* eae8714026339e1e7926d01412916ad08be737b9 client.py: Replace hardcoded 'admin' with options.principal


ipa-4-5:

* 7bef8c7ff16c95a43312a3d162ff354625ecd198 client.py: Replace hardcoded 'admin' with options.principal

PR #830, opened 2017-05-30
custodia dep: require explictly python2 version

MartinBasti:
python-custodia matches python3-custodia, but for py2 installations we
need python2-custodia explicitly

https://pagure.io/freeipa/issue/6962


martbab:
master:

* a90a113b66fca620b04635442b135a5136ece7ba custodia dep: require explictly python2 version


ipa-4-5:

* 444107a00bf995aca62aba74ea02b52e577ab791 custodia dep: require explictly python2 version

PR #831, opened 2017-05-30
[4.4] custodia dep: require explictly python2 version

MartinBasti:
python-custodia matches python3-custodia, but for py2 installations we
need python2-custodia explicitly

https://pagure.io/freeipa/issue/6962


martbab:
ipa-4-4:

* a1276d550a1a5f28e1214ceb53cbe460428baef1 custodia dep: require explictly python2 version

PR #833, opened 2017-05-30
Fixes traceback in log and corrects console output

Tiboris:
There was traceback in error_log, when running trustdomain-find
command for trusted child domain and it results into ERROR.
Correct behavior is to show 0 entries found.

Fixes https://pagure.io/freeipa/issue/6477

### DIFF
--- a/ipaserver/plugins/baseldap.py
+++ b/ipaserver/plugins/baseldap.py
@@ -2007,7 +2007,14 @@ class LDAPSearch(BaseLDAPCommand, crud.Search):
         except IndexError:
             term = None
         if self.obj.parent_object:
-            base_dn = self.api.Object[self.obj.parent_object].get_dn(*keys)
+            try:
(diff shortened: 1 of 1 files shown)


felipevolpone:
Tested and it works.

Looks good to me.

HonzaCholasta:
NACK. Please don't "fix" the framework with workarounds for ill-behaved commands, this needs to be fixed in the trust plugin.

stlaz:
Yes, I believe that when base DN of a search is not found, we should not be generally treating this gently.

felipevolpone:
@Tiboris if I understood @HonzaCholasta correctly, the fix should be done in [trustdomain_find command](https://github.com/freeipa/freeipa/blob/master/ipaserver/plugins/trust.py#L1544).

stlaz:
Fixing this is not a trivial problem, it would take a lot of time which does not bring that much value. We (@Tiboris, @martbab and me) agreed to close this as rejected.

PR #835, opened 2017-05-31
kdc.key should not be visible to all

stlaz:
While the world certainly is interested in our privates, we
should not just go ahead and show it to them.

https://pagure.io/freeipa/issue/6973

### DIFF
--- a/ipalib/install/certmonger.py
+++ b/ipalib/install/certmonger.py
@@ -370,8 +370,8 @@ def request_cert(
         request_parameters['cert-postsave-command'] = post_command
 
     if perms:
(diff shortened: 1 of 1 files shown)


abbra:
LGTM. Nice that at least one pair of eyes noticed the mixup.

martbab:
master:

* 3b6892783ee6ed6dac9c4f328cc89ae030ce10a7 kdc.key should not be visible to all


ipa-4-5:

* 37be8e9ac3b46d6d31199227216b5a5a8d5d5614 kdc.key should not be visible to all

PR #836, opened 2017-05-31
Only warn when specified server IP addresses don't match intf

MartinBasti:
In containers local addresses differ from public addresses and we need
a way to provide only public address to installers.

https://pagure.io/freeipa/issue/2715
https://pagure.io/freeipa/issue/4317

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -276,7 +276,7 @@ def read_ip_addresses():
         if not ip:
             break
         try:
-            ip_parsed = ipautil.CheckedIPAddress(ip, match_local=True)
+            ip_parsed = ipautil.CheckedIPAddress(ip)
         except Exception as e:
             print("Error: Invalid IP Address %s: %s" % (ip, e))
             continue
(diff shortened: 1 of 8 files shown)


ipaserver/install/installutils.py
```diff
@@ -585,7 +585,7 @@ def get_server_ip_address(host_name, unattended, setup_dns, ip_addresses):
     if len(hostaddr):
         for ha in hostaddr:
             try:
-                ips.append(ipautil.CheckedIPAddress(ha, match_local=True))
+                ips.append(ipautil.CheckedIPAddress(ha, match_local=False))
```
tomaskrizek: Shouldn't this behaviour also change on line 279?

tomaskrizek:
Do we want to handle `ipa-client-install --ip-address` in the same way? I think the behaviour should be consistent.

ipaserver/install/installutils.py
MartinBasti: Indeed it should

MartinBasti:
I don't see reason why not, PR updated.

tomaskrizek:
Thanks for the updates, looks good now.

MartinBasti:
master:

* 6637980af6069623d944d9d592cbadba20b610a2 Only warn when specified server IP addresses don't match intf


ipa-4-5:

* 6206ac8bd23250bda0f8eb628f422671b9b99ad1 Only warn when specified server IP addresses don't match intf

PR #838, opened 2017-06-01
Explicitly ask for py2 dependencies in py2 packages

MartinBasti:
In future default package names can start to pointing to py3 instead of
py2. We have to explicitly ask for python2-* and python3-* packages.

This commit changes only dependencies that are available in both F25 and
F26


stlaz:
Please resolve conflicts. Otherwise LGTM + works on F26

MartinBasti:
Resolved

martbab:
master:

* a2147de6e2eb217163d6f106d3220c7b1e7570b5 Explicitly ask for py2 dependencies in py2 packages

PR #840, opened 2017-06-01
Add Role 'Enrollment Administrator'

Tiboris:
User with the 'Enrollment Administrator' role assigned to is
able to enroll host against a FreeIPA server as a client
using the ipa-client-install command.

The 'Enrollment Administrator' contains 'Host Enrollment' privilege only.

Points to: https://pagure.io/freeipa/issue/6852

### DIFF
--- a/install/updates/45-roles.update
+++ b/install/updates/45-roles.update
@@ -91,3 +91,12 @@ add:member: cn=Security Architect,cn=roles,cn=accounts,$SUFFIX
 dn: cn=Password Policy Administrator,cn=privileges,cn=pbac,$SUFFIX
 add:member: cn=Security Architect,cn=roles,cn=accounts,$SUFFIX
 
+dn: cn=Enrollment Administrator,cn=roles,cn=accounts,$SUFFIX
+default:objectClass: groupofnames
+default:objectClass: nestedgroup
+default:objectClass: top
+default:cn: Enrollment Administrator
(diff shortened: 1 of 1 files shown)


MartinBasti:
Off Topic: Shouldn't have "IT specialists" also 'Host Enrollment' privilege, because they have 'Host administrators' already and this should close the circle.

"Client administrator" sounds to me like too much as the role can only enroll client.  How about "Enrolling Administrator"/"Client Enrolling Administrator". But I'm not sure.

abbra:
I think "Enrollment Administrator" would be fine. On the "IT Specialists" and "Host Enrollment" -- why not, it seems logical.

Tiboris:
Thanks, changed.

abbra:
LGTM.

abbra:
The patch works but we also agreed to add 'Host Enrollment' privilege to 'IT Specialists'. Please add this as a separate commit mentioning the same ticket.

Tiboris:
On it.

Tiboris:
I think adding 'Host Enrollment' privilege to 'IT Specialists' is not necessary because the  'IT Specialists' role has already a 'Host Administrators' privilege assigned an this privilege contain all of the 'Host Enrollment' privilege permissions.

abbra:
Got it. Thanks.

martbab:
master:

* 468eb3c712140399ed2ec346ff4356bffd590e09 Add Role 'Enrollment Administrator'

PR #841, opened 2017-06-01
ipa-kdb: use canonical principal in certauth plugin

sumit-bose:
Currently the certauth plugin use the unmodified principal from the
request to lookup the user. This might fail if e.g. enterprise
principals are use. With this patch the canonical principal form the kdc
entry is used.

Resolves https://pagure.io/freeipa/issue/6993

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb_certauth.c
+++ b/daemons/ipa-kdb/ipa_kdb_certauth.c
@@ -274,7 +274,7 @@ static krb5_error_code ipa_certauth_authorize(krb5_context context,
         }
     }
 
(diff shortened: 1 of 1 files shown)


dkupka:
Works for me.

MartinBasti:
master:

* 117d6e9be0c386f134bd27eee3377e70df77f0f0 ipa-kdb: use canonical principal in certauth plugin


ipa-4-5:

* e8d8aab469ca634f4ec38b869767316806c739f1 ipa-kdb: use canonical principal in certauth plugin

PR #843, opened 2017-06-02
[WIP] Fixing test_installation.py tests

felipevolpone:
I've been working on the test_installation.py suite and figure out how to solve some of them.

The TestInstallWithCA1 have 9 tests failing; 6 of them can be fixed adding 
```bash
<ip> ipa-ca.$DOMAIN
```
into the master `/etc/hosts`. After that, three of them are still failing.
The log: https://paste.fedoraproject.org/paste/7n3CMEH5nhiHu~Vai8cObV5M1UNdIGYhyRLivL9gydE=. 

They are:
 * test_replica1_with_ca_install
 * test_replica2_with_ca_kra_install
 * test_replica1_ipa_kra_install

I've moved the tests 
 * test_replica2_with_ca_kra_install
 * test_replica1_ipa_kra_install

to a new class (TestInstallWithCA1_KRA1) and created a new install method, which use the `setup_kra=True` option in the install_master method. The tests are still failing, but for another reason, the logs: https://paste.fedoraproject.org/paste/ytzzIUDhh5ARcunpSfSubV5M1UNdIGYhyRLivL9gydE=

Ticket: https://pagure.io/freeipa/issue/6995

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -110,7 +110,6 @@ class ADTrustInstallTestBase(IntegrationTest):
 ##
 # Master X Replicas installation tests
 ##
-
 class TestInstallWithCA1(InstallTestBase1):
 
     @classmethod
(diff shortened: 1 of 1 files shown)


MartinBasti:
Tests are fine, but FreeIPA is broken with missing ipa-ca record, there is an ongoing work that should resolve this issue.

felipevolpone:
@MartinBasti you mean that they're fine after the code change, right? Because, before that they were failing due to `There is no KRA server in the domain, can't setup a KRA clone`.

MartinBasti:
IIRC that those tests used to pass, so it may be a new bug in IPA or it just suffer from domino effect caused by `ipa-ca` issues.

I actually don't understand what your commit fixes. This error `There is no KRA server in the domain, can't setup a KRA clone` shouldn't happen because installing KRA is part of test in `InstallTestBase1.test_replica0_ipa_kra_install` that precedes installation of `replica1` and `replica2`. So topology should have KRA server installed on `replica0`. IMO this step failed due `ipa-ca` failure and caused domino effect.

PR #844, opened 2017-06-02
py3: fix regression in schemaupdate

frasertweedale:
The python-ldap classes that process schema definitions require a
unicode string, not a byte string.  A recent py3 compatibility fix
(d89de4219d0e8ee33e81d6b6d1bc6c22ac9ffbaa) changed the constructor
argument to a unicode string to dispel a warning, but this broke
schema update.  Change it back to a bytestring.

Part of: https://fedorahosted.org/freeipa/ticket/4985


MartinBasti:
Thank you for this catch

MartinBasti:
master:

* 89eb162fcd60861ed4c628dab4e1aaf10c6160bb py3: fix regression in schemaupdate

PR #845, opened 2017-06-02
ipadiscovery: Return realm as a string

stlaz:
We don't have a use for realm as a bytes instance, return it as a
string, otherwise there's a use of str() on bytes in py3.

### DIFF
--- a/ipaclient/install/ipadiscovery.py
+++ b/ipaclient/install/ipadiscovery.py
@@ -514,7 +514,11 @@ class IPADiscovery(object):
         return servers
 
     def ipadnssearchkrbrealm(self, domain=None):
-        realm = None
+        """
+        :param domain: Domain to be searched in
+        :returns: string of a realm found in a TXT record
+                  None if no realm was found
+        """
         if not domain:
             domain = self.domain
         # now, check for a Kerberos realm the local host or domain is in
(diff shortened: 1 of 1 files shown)


MartinBasti:
A TXT record can contain anything in any encoding, it contains just bytes without predefined semantic, we should play safe here and catch UnicodeDecodeError.

Also currently we support only ASCII for realms, I'm not sure which character are allowed to be in realm outside IPA world, I haven't found a clear answer.

stlaz:
https://tools.ietf.org/html/rfc1464 suggests that only ASCII may be contained in a TXT record, although I like to go with the Internet rule: "Trust noone." and so I decided to decode it as UTF-8.
You want to go even further on the aforementioned rule which is OK with me, then, I will update the PR.

MartinBasti:
That is an experimental only RFC. By general it can contain any byte textually represented in octal form.

str() in python2 converts from bytes using ASCII codec, so by keeping compatibility and avoiding to parse weird realms it should be IMO ASCII. But I can live with utf-8 as well, but I haven't checked how is this handled later in code

MartinBasti:
LGTM, we anyway assume inside framework that everything is in utf-8

MartinBasti:
master:

* 1cab1e980c7cad65943e2984158a458a6472c9d2 ipadiscovery: Return realm as a string
* c8cc62564b09959c451522b1d7eb122566594867 Docstring+refactor of IPADiscovery.ipadnssearchkrbrealm()

PR #847, opened 2017-06-02
Turn off OCSP check

pvomacka:
The OCSP check was previously turned on but it introduced several
issues. Therefore the check will be turned off by default.

For turning on should be used ipa advise command with correct recipe.
The solution is tracked here: https://pagure.io/freeipa/issue/6982

Fixes: https://pagure.io/freeipa/issue/6981

### DIFF
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -34,8 +34,7 @@ from augeas import Augeas
 from ipalib.install import certmonger
 from ipapython import ipaldap
 from ipapython.certdb import (IPA_CA_TRUST_FLAGS,
-                              EXTERNAL_CA_TRUST_FLAGS,
-                              TRUSTED_PEER_TRUST_FLAGS)
+                              EXTERNAL_CA_TRUST_FLAGS)
 from ipaserver.install import replication
 from ipaserver.install import service
 from ipaserver.install import certs
@@ -66,6 +65,10 @@ NSS_CIPHER_SUITE = [
     '+rsa_aes_256_gcm_sha_384', '+rsa_aes_256_sha'
 ]
 NSS_CIPHER_REVISION = '20160129'
+
+OCSP_DIRECTIVE = 'NSSOCSP'
+
+NSS_OCSP_ENABLED = 'nss_ocsp_enabled'
 
 
 def httpd_443_configured():
@@ -156,7 +159,7 @@ class HTTPInstance(service.Service):
                   self.set_mod_nss_protocol)
         self.step("setting mod_nss password file", self.__set_mod_nss_passwordfile)
         self.step("enabling mod_nss renegotiate", self.enable_mod_nss_renegotiate)
-        self.step("enabling mod_nss OCSP", self.enable_mod_nss_ocsp)
+        self.step("disabling mod_nss OCSP", self.disable_mod_nss_ocsp)
         self.step("adding URL rewriting rules", self.__add_include)
         self.step("configuring httpd", self.__configure_http)
         self.step("setting up httpd keytab", self.request_service_keytab)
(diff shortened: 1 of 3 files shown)


martbab:
How did we resolve the issue of tracking nssocsp status in sysupgrade state? Shouldn't we record this so that we now it was disabled by our installer/upgrader?

ipaserver/install/httpinstance.py
```diff
@@ -272,21 +271,18 @@ def enable_mod_nss_ocsp(self):
...
-        elif ocsp_dir is None:
-            # Directive is missing and comment is missing
-            aug.set('{}/directive[last()+1]'.format(path), "NSSOCSP")
+        # there is NSSOCSP directive in nss.conf file, comment it
+        # otherwise just do nothing
+        if ocsp_dir is not None:
+            ocsp_state = aug.get('{}/directive[.="NSSOCSP"]/arg'.format(path))
+            aug.remove('{}/directive[.="NSSOCSP"]/arg'.format(path))
+            aug.rename('{}/directive[.="NSSOCSP"]'.format(path), '#comment')
+            aug.set('{}/#comment[.="NSSOCSP"]'.format(path),
```
martbab: Extracting the string literals into a separate variables named e.g. `ocsp_path`, `ocsp_arg` could really improve the readability of the method.

ipaserver/install/httpinstance.py
```diff
@@ -263,30 +264,36 @@ def enable_mod_nss_renegotiate(self):
         installutils.set_directive(paths.HTTPD_NSS_CONF, 'NSSRenegotiation', 'on', False)
         installutils.set_directive(paths.HTTPD_NSS_CONF, 'NSSRequireSafeNegotiation', 'on', False)
 
-    def enable_mod_nss_ocsp(self):
+    def disable_mod_nss_ocsp(self):
+        if not sysupgrade.get_upgrade_state('httpd', 'nss_oscp_enabled'):
```
martbab: I would reformulate the condition as `if sysupgrade.get_upgrade_state('httpd', 'nss_ocsp_enabled') is None` this correctly reflects the semantics that if the upgrade state is None, it was not touched by IPA before.

Also please move the 'nss_ocsp_enabled' literal to a top-level module constants named e.g. NSS_OCSP_ENABLED.

pvomacka:
@martbab thank you for review. I did changes which you proposed.

ipaserver/install/httpinstance.py
```diff
@@ -263,30 +266,34 @@ def enable_mod_nss_renegotiate(self):
         installutils.set_directive(paths.HTTPD_NSS_CONF, 'NSSRenegotiation', 'on', False)
         installutils.set_directive(paths.HTTPD_NSS_CONF, 'NSSRequireSafeNegotiation', 'on', False)
 
-    def enable_mod_nss_ocsp(self):
+    def disable_mod_nss_ocsp(self):
+        if sysupgrade.get_upgrade_state('httpd', NSS_OCSP_ENABLED) is None:
+            self.__disable_mod_nss_ocsp()
+            sysupgrade.set_upgrade_state('httpd', NSS_OCSP_ENABLED, False)
```
martbab: During testing I just noticed that in sysrestore Apache is referred to as 'httpd', but in sysupgrade it is referred to as 'http'. Can you please rename the section into 'http' in the sysupgrade checks?

Sorry that I did not notice it earlier, I again falled into a trap of assuming consistency in our code.

martbab:
ipa-4-5:

* 51b361f475b3e25ace982873beb05cafcba95808 Turn off OCSP check


master:

* 566361e63d4a670460df3dbb28b9d19f38eaea2d Turn off OCSP check

PR #849, opened 2017-06-02
session_storage: Correctly handle string/byte types

stlaz:
In session_storage.py, store_data() stores data as the bytes data
type but get_data() is returning a string. Have get_data() return
bytes as well.


martbab:
master:

* d665224a85610cccbe7d291e9ed41d2ce7e5b61c session_storage: Correctly handle string/byte types

PR #851, opened 2017-06-05
ipa-kdb: add pkinit authentication indicator in case of a successful certauth

abbra:
We automatically add 'otp' and 'radius' authentication indicators when
pre-authentication with OTP or RADIUS did succeed. Do the same for
certauth-based pre-authentication (PKINIT).

A default PKINIT configuration does not add any authentication
indicators unless 'pkinit_indicator = pkinit' is set in kdc.conf.
Unfortunately, modifying kdc.conf automatically is a bit more
complicated than modifying krb5.conf. Given that we have 'otp' and
'radius' authentication indicators also defined in the code not in the
kdc.conf, this change is following an established trend.

SSSD certauth interface does not provide additional information about
which rule(s) succeeded in matching the incoming certificate. Thus,
there is not much information we can automatically provide in the
indicator. It would be good to generate indicators that include some
information from the certmapping rules in future but for now a single
'pkinit' indicator is enough.

Fixes https://pagure.io/freeipa/issue/6736

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb_certauth.c
+++ b/daemons/ipa-kdb/ipa_kdb_certauth.c
@@ -267,6 +267,7 @@ static krb5_error_code ipa_certauth_authorize(krb5_context context,
     int ret;
     size_t c;
     char *principal = NULL;
+    char **auth_inds = NULL;
     LDAPMessage *res = NULL;
     krb5_error_code kerr;
     LDAPMessage *lentry;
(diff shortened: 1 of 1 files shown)


martbab:
master:

* e8a7e2e38ad7cea2964305247430e964d2b785b1 ipa-kdb: add pkinit authentication indicator in case of a successful certauth


ipa-4-5:

* ca02cea8dfd63290e4821833fc2ac7d457290e9f ipa-kdb: add pkinit authentication indicator in case of a successful certauth

PR #852, opened 2017-06-05
pkinit manage: introduce ipa-pkinit-manage

HonzaCholasta:
**server certinstall: update KDC master entry**

After the KDC certificate is installed, add the PKINIT enabled flag to the
KDC master entry.

**pkinit manage: introduce ipa-pkinit-manage**

Add the ipa-pkinit-manage tool to allow enabling / disabling PKINIT after
the initial server install.

**server upgrade: do not enable PKINIT by default**

Enabling PKINIT often fails during server upgrade when requesting the KDC
certificate.

Now that PKINIT can be enabled post-install using ipa-pkinit-manage, avoid
the upgrade failure by not enabling PKINIT by default.

https://pagure.io/freeipa/issue/7000

### DIFF
--- a/install/tools/man/ipa-pkinit-manage.1
+++ b/install/tools/man/ipa-pkinit-manage.1
@@ -1,0 +1,34 @@
+.\"
+.\" Copyright (C) 2017  FreeIPA Contributors see COPYING for license
+.\"
+.TH "ipa-pkinit-manage" "1" "Jun 05 2017" "FreeIPA" "FreeIPA Manual Pages"
+.SH "NAME"
+ipa\-pkinit\-manage \- Enables or disables PKINIT
+.SH "SYNOPSIS"
+ipa\-pkinit\-manage [options] <enable|disable|status>
+.SH "DESCRIPTION"
+Run the command with the \fBenable\fR option to enable PKINIT.
+
+Run the command with the \fBdisable\fR option to disable PKINIT.
+
+Run the command with the \fBstatus\fR to determine the current status of PKINIT.
+.SH "OPTIONS"
+.TP
+\fB\-\-version\fR
+Show the program's version and exit.
+.TP
+\fB\-h\fR, \fB\-\-help\fR
+Show the help for this program.
+.TP
+\fB\-v\fR, \fB\-\-verbose\fR
+Print debugging information.
+.TP
+\fB\-q\fR, \fB\-\-quiet\fR
+Output only errors.
+.TP
+\fB\-\-log\-file\fR=\fIFILE\fR
+Log to the given file.
+.SH "EXIT STATUS"
+0 if the command was successful
+
+1 if an error occurred
--- a/ipaserver/install/ipa_pkinit_manage.py
+++ b/ipaserver/install/ipa_pkinit_manage.py
@@ -1,0 +1,93 @@
+#
+# Copyright (C) 2017  FreeIPA Contributors see COPYING for license
+#
+
+from __future__ import print_function
+
+from ipalib import api
+from ipaplatform.paths import paths
+from ipapython.admintool import AdminTool
+from ipaserver.install.krbinstance import KrbInstance, is_pkinit_enabled
+
+
+class PKINITManage(AdminTool):
+    command_name = "ipa-pkinit-manage"
+    usage = "%prog <enable|disable|status>"
+    description = "Manage PKINIT."
+
+    def validate_options(self):
+        super(PKINITManage, self).validate_options(needs_root=True)
+
+        option_parser = self.option_parser
+
+        if not self.args:
+            option_parser.error("action not specified")
+        elif len(self.args) > 1:
+            option_parser.error("too many arguments")
+
+        action = self.args[0]
+        if action not in {'enable', 'disable', 'status'}:
+            option_parser.error("unrecognized action '{}'".format(action))
+
+    def run(self):
+        api.bootstrap(in_server=True, confdir=paths.ETC_IPA)
+        api.finalize()
+
+        api.Backend.ldap2.connect()
+        try:
+            action = self.args[0]
+            if action == 'enable':
+                self.enable()
+            elif action == 'disable':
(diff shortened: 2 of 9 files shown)


install/tools/man/ipa-pkinit-manage.1
```diff
@@ -0,0 +1,34 @@
+.\"
+.\" Copyright (C) 2017  FreeIPA Contributors see COPYING for license
+.\"
+.TH "ipa-pkinit-manage" "1" "Jun 05 2017" "FreeIPA" "FreeIPA Manual Pages"
+.SH "NAME"
+ipa\-pkinit\-manage \- Enables or disables PKINIT
```
martbab: I would add here a comment that this enables PKINIT on single IPA master only, not in the whole topology. Otherwise users may get confused.

ipaserver/install/krbinstance.py
```diff
@@ -451,6 +451,30 @@ def pkinit_enable(self):
         service.set_service_entry_config(
             'KDC', self.fqdn, [PKINIT_ENABLED], self.suffix)
 
+    def pkinit_disable(self):
+        """
+        unadvertise enabled PKINIT feature in master's KDC entry in LDAP
+        """
+        ldap = api.Backend.ldap2
+        dn = DN(('cn', 'KDC'),
+                ('cn', self.fqdn),
```
martbab: Yo can use `DN(('cn', 'KDC'), api.Object.master.get_dn(self.fqdn))` to construct service DN in a cleaner way using existing abstractions.

ipaserver/install/ipa_pkinit_manage.py
```diff
@@ -0,0 +1,93 @@
...
+
+        if bool(is_pkinit_enabled()) is not bool(setup_pkinit):
+            try:
+                krb.stop_tracking_certs()
+            except RuntimeError as e:
+                if ca_enabled:
+                    self.log.warning(
+                        "Failed to stop tracking certificates: %s", e)
+
+            krb.enable_ssl()
```
martbab: `enable_ssl` should do `pkinit_enabled` when configuring proper/external PKINIT. However, for readability's sake I am OK with leaving an explicit call below as long as it doesn't break anything.

ipaserver/install/ipa_pkinit_manage.py
```diff
@@ -0,0 +1,93 @@
...
+from ipalib import api
+from ipaplatform.paths import paths
+from ipapython.admintool import AdminTool
+from ipaserver.install.krbinstance import KrbInstance, is_pkinit_enabled
+
+
+class PKINITManage(AdminTool):
+    command_name = "ipa-pkinit-manage"
+    usage = "%prog <enable|disable|status>"
+    description = "Manage PKINIT."
```
martbab: "manage PKINIT on the master."

ipaserver/install/ipa_pkinit_manage.py
```diff
@@ -0,0 +1,93 @@
...
+        self._setup(True)
+
+    def disable(self):
+        self._setup(False)
+
+    def status(self):
+        if is_pkinit_enabled():
+            print("PKINIT is enabled")
+        else:
+            print("PKINIT is disabled")
```
martbab: Also here I would add master's FQDN to the message to make clear this is tied to the local master's config.

install/tools/man/ipa-pkinit-manage.1
HonzaCholasta: The same phrasing works for `ipa-compat-manage`, `ipa-managed-entries` and `ipa-nis-manage`, so I don't see why it should not work here as well.

ipaserver/install/krbinstance.py
HonzaCholasta: This is how the DN is constructed throughout installers (including `set_service_entry_config` which you yourself added recently), so I'm not changing it.

ipaserver/install/ipa_pkinit_manage.py
HonzaCholasta: Which master is "the master"?

ipaserver/install/ipa_pkinit_manage.py
HonzaCholasta: This way is more robust - note that `enable_ssl` is called here only when PKINIT has not already been enabled, whereas `pkinit_enable` is called *always*.

ipaserver/install/ipa_pkinit_manage.py
HonzaCholasta: Same as above. I prefer consistency over perfection.

ipaserver/install/ipa_pkinit_manage.py
martbab: Then maybe something like "Manage local PKINIT configuration on IPA server", but if the locality of the commands is not specified in other management tools then it is better to keep at least some semblance of consistency.

ipaserver/install/krbinstance.py
martbab: ok.

install/tools/man/ipa-pkinit-manage.1
martbab: I see, let's keep it like this, then.

martbab:
ipa-4-5:

* 1b62e5aac9d9668604e82879c020bff310fa549f server certinstall: update KDC master entry
* c072135340bc8e75f621e2b9163b1347b9eb528f pkinit manage: introduce ipa-pkinit-manage
* cb9353d6e0fbc0912dd20bf29e3835a7740d1af6 server upgrade: do not enable PKINIT by default


master:

* e131905f3e0fe9179c5f4a09da4e7a204012603a server certinstall: update KDC master entry
* 92276c1e8809f3ff6b59bd6124869f816627bac7 pkinit manage: introduce ipa-pkinit-manage
* 0772ef20b39b11950fddc913a350534988294c89 server upgrade: do not enable PKINIT by default

PR #853, opened 2017-06-06
x509,certdb: handle certificates as bytes

stlaz:
Certificates, both in PEM and DER format, should be handled as bytes
in Python 3.

=================================
This PR unblocks IPA with **external CA** installation under Python 3.

### DIFF
--- a/ipalib/x509.py
+++ b/ipalib/x509.py
@@ -235,12 +235,10 @@ def make_pem(data):
     Convert a raw base64-encoded blob into something that looks like a PE
     file with lines split to 64 characters and proper headers.
     """
-    if isinstance(data, bytes):
-        data = data.decode('ascii')
-    pemcert = '\r\n'.join([data[x:x+64] for x in range(0, len(data), 64)])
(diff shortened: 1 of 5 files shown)


stlaz:
To test this, you can use the script I provided at https://github.com/freeipa/freeipa-tools/pull/48, you just have to add `python3 ` before the `ipa-server-install` commands.

stlaz:
This will require some more refactoring, adding WIP label.

Basically, `Bytes` parameter only base64-decodes its value if it's of `unicode`/`str` instance (Python2/Python3), but in most cases, its value is checked with `validate_ceritifate()` of the `ipaserver.plugins.service` module, which requires the certificate to be in DER format. But the ceritificate won't be decoded to DER from PEM as it above decoding condition is not fulfilled.

stlaz:
Added a potential fixup to the tests to see whether they'd pass. It's not supposed to be pushed since it either proves that this PR should be worked on more thoroughly.

stlaz:
Now included in https://github.com/freeipa/freeipa/pull/878, closing this PR.

PR #854, opened 2017-06-06
server-side and client-side advises for configuring smart card auth

martbab:
These advises generate scripts which configure all the components
required for successful processing and performing of smart card auth requests on IPA server and enrolled clients.

https://pagure.io/freeipa/issue/6982

### DIFF
--- a/ipaserver/advise/plugins/smart_card_auth.py
+++ b/ipaserver/advise/plugins/smart_card_auth.py
@@ -1,0 +1,266 @@
+#
+# Copyright (C) 2017 FreeIPA Contributors see COPYING for license
+#
+
+from ipalib.plugable import Registry
+from ipaplatform.paths import paths
+from ipaserver.advise.base import Advice
+from ipaserver.install.httpinstance import NSS_OCSP_ENABLED
+
+register = Registry()
+
+
+@register()
+class config_server_for_smart_card_auth(Advice):
+    """
+    Configures smart card authentication via Kerberos (PKINIT) and for WebUI
+    """
+
+    description = ("Instructions for enabling Smart Card authentication on "
+                   " a single FreeIPA server. Includes Apache configuration, "
+                   "enabling PKINIT on KDC and configuring WebUI to accept "
+                   "Smart Card auth requests. To enable the feature in the "
+                   "whole topology you have to run the script on each master")
+
+    nss_conf = paths.HTTPD_NSS_CONF
+    nss_ocsp_directive = 'NSSOCSP'
+    nss_nickname_directive = 'NSSNickname'
+
+    def get_info(self):
+        self.log.exit_on_nonroot_euid()
+        self.check_ccache_not_empty()
+        self.check_hostname_is_in_masters()
+        self.resolve_ipaca_records()
+        self.enable_nss_ocsp()
+        self.mark_httpd_cert_as_trusted()
+        self.restart_httpd()
+        self.record_httpd_ocsp_status()
+        self.check_and_enable_pkinit()
+        self.enable_ok_to_auth_as_delegate_on_http_principal()
+
+    def check_ccache_not_empty(self):
+        self.log.comment('Check whether the credential cache is not empty')
+        self.log.exit_on_failed_command(
+            'klist',
+            [
+                "Credential cache is empty",
+                'Use kinit as privileged user to obtain Kerberos credentials'
+            ])
+
+    def check_hostname_is_in_masters(self):
+        self.log.comment('Check whether the host is IPA master')
+        self.log.exit_on_failed_command(
+            'ipa server-find $(hostname -f)',
+            ["This script can be run on IPA master only"])
+
+    def resolve_ipaca_records(self):
+        ipa_domain_name = self.api.env.domain
+
+        self.log.comment('make sure bind-utils are installed so that we can '
+                         'dig for ipa-ca records')
+        self.log.exit_on_failed_command(
+            'yum install -y bind-utils',
+            ['Failed to install bind-utils'])
+
+        self.log.comment('make sure ipa-ca records are resolvable, '
+                         'otherwise error out and instruct')
+        self.log.comment('the user to update the DNS infrastructure')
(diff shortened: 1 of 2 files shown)


abbra:
It looks good, thanks. One small comment is that we probably want to allow the script to work on non-RPM platforms as well (Debian or Ubuntu too) where `yum` does not exist. On the other hand, we also would probably want to eventually allow setting this up against `mod_ssl` too. These two changes can be added later.

martbab:
support for non-RPM platforms would require some more additions to the base Advice code to handle this systematically, or alternatively we may just test for the presence of required command and fail with instruction to install the missing package using platform-specific means.

While we may want to migrate to `mod_ssl` in the future, there will be much more work to do regarding switching TLS modules so the advise can be ported as a part of this effort.

abbra:
Yes, I'm not asking you to implement these improvements as part of this PR. That's fine to be done in a separate effort.

flo-renaud:
Hi @martbab 
thank you for the patch. Works great!
We could also enhance the script:
- check that it is run by the root user
- to enable NSSOCP, the regex does not match "##NSSOCSP off" but should. 
- the script must be run on each IPA server, maybe we should make it clearer.
- the script could also configure /etc/sssd.conf with
[pam]
pam_cert_auth = True

This could be done in a later commit, I can open a RFE if needed

martbab:
@flo thanks for your input, I will rework the PR tomorrow.

martbab:
@flo @abbra I have rebased PR and included also a recipe for client configuration for the sake of completeness.

abbra:
Thanks. Comments so far:
* client configuration does not make sure to ask for a removal of `pam_pkcs11` package
* client configuration does not call for the `authconfig` to configure PAM stack. It seems to me that we either need a second recipe that enables IPA client system for a remote logon or need to integrate it in the same recipe
* there are few places in the server plugin that do an explicit 'if ... ' command. This probably would be better to replace with `commands_on_predicate` like in the client case

martbab:
@abbra thanks for review. Is `pam_pkcs11` removal necessary for client? Also what option does the recipe need to pass to `authconfig` to properly configure smart card auth? isn't it enough to configure SSSD?

abbra:
It is all documented in https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7-Beta/html/Linux_Domain_Identity_Authentication_and_Policy_Guide/smart-cards.html (notice 7-Beta, this is a new material for RHEL 7.4 beta), section 22.5.1.3.

martbab:
That section[1] only instructs to configure `pam_cert_auth=true` in the SSSD's `pam` section which is already done on both server and client, see `enable_pam_auth_in_sssd` method. Am I missing something? 

[1] https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7-Beta/html/Linux_Domain_Identity_Authentication_and_Policy_Guide/idm-smart-cards.html

flo-renaud:
Hi @martbab 
I think @abbra was referring to this [section](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7-Beta/html/Linux_Domain_Identity_Authentication_and_Policy_Guide/smart-cards-advanced.html#sc-clients-config) running authconfig --enablesmartcard --smartcardmodule=sssd --updateall
This command will update sssd.conf with pam_cert_auth=True.

martbab:
@flo ah sorry I missed that. I will incorporate it into advise then.

martbab:
@flo regarding enabling Smart Card login ( add PKCS#11 module, configure SSSD and such), do we really need to setup this on server? I do not expect somebody logging directly to machine hosting FreeIPA server using smard card reader.

abbra:
@martbab, this actually makes full sense -- if you want to increase the security of your IPA masters, you might force using smart cards only to do ssh login. We are not there yet by default but I can see it being a mandate for organizations that enforce industry or governmental requirements.

abbra:
Note that "directly" may actually mean using a virtualized remote smart card access which is provided via virtualized USB pass-through done by your virtualization environment.

martbab:
Also I get the following error when running authconfig:

```console
authconfig: Authentication module /lib64/security/pam_pkcs11.so is missing. Authentication process might not work correctly.
```

It is understandable, since I have removed pam_pkcs11 package as per documentation, but it still puzzles me.

It may be that I have an old version of authconfig, as I am developing this on F25 where I have authconfig-6.2.10-14.fc25.x86_64.

abbra:
@martbab, definitely `authconfig` in fc25 is too old for this. On F26 I have version 7.0.1-1. It does announce support for SSSD smartcard enablement.

ipaserver/advise/plugins/smart_card_auth.py
```diff
@@ -0,0 +1,266 @@
...
+            ['Failed to install bind-utils'])
+
+        self.log.comment('make sure ipa-ca records are resolvable, '
+                         'otherwise error out and instruct')
+        self.log.comment('the user to update the DNS infrastructure')
+        self.log.command('ipaca_records=$(dig +short '
+                         'ipa-ca.{})'.format(ipa_domain_name))
+
+        self.log.exit_on_predicate(
+            '[ -z "ipaca_records" ]',
```
flo-renaud: This should be $ipaca_records (the $ is missing).

martbab: Oops sorry, should be fixed now.

flo-renaud:
Hi @martbab 
Thank you for the fix. Works as expected.

martbab:
master:

* 0569c02f17f853d97280f52f4a7fefecc72cf45d Extend the advice printing code by some useful abstractions
* e418e9a4ca747886c53d05ae80597834f1d3d021 Prepare advise plugin for smart card auth configuration


ipa-4-5:

* 7ea7ee4326679c098d3e4e4d6a2bc743707708ca Extend the advice printing code by some useful abstractions
* 84ca9761bd47f28b72581d1fe6bd8cfa824b6df3 Prepare advise plugin for smart card auth configuration

PR #855, opened 2017-06-06
Prevent issues with older clients

simo5:
Older clients have issues parsing cookies, and cannot handle well the MaxAge setting.
So the first patch is about removing it.

Unfortunately this means cookies will be valid for the duration of the authentication ticket which is set to 24h by default.
This is a bit high, so the second patch adds the ability to set the "kinit_lifetime" in /etc/api/default.conf so that users authenticating using username/password can have their tickets (and therefore their session) hard capped at whatever lifetime is set there.

Users that use HTTP negotiate can control their session duration by getting shorter lived tickets via kinit.

In all cases users can click on the logout button to blow away credentials.

### DIFF
--- a/install/conf/ipa.conf
+++ b/install/conf/ipa.conf
@@ -1,5 +1,5 @@
 #
-# VERSION 26 - DO NOT REMOVE THIS LINE
+# VERSION 27 - DO NOT REMOVE THIS LINE
 #
 # This file may be overwritten on upgrades.
 #
@@ -77,7 +77,9 @@ WSGIScriptReloading Off
   Session On
   SessionCookieName ipa_session path=/ipa;httponly;secure;
   SessionHeader IPASESSION
-  SessionMaxAge 1800
+  # Uncomment the following to have shorter sessions, but beware this may break
+  # old IPA client tols that incorrectly parse cookies.
+  # SessionMaxAge 1800
   GssapiSessionKey file:/etc/httpd/alias/ipasession.key
 
   GssapiImpersonate On
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -155,6 +155,7 @@ DEFAULT_CONFIG = (
     ('session_auth_duration', '20 minutes'),
     # How a session expiration is computed, see SessionManager.set_session_expiration_time()
     ('session_duration_type', 'inactivity_timeout'),
+    ('kinit_lifetime', None),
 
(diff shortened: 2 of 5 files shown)


simo5:
Fixes https://pagure.io/freeipa/issue/6774

simo5:
In my test setup I verified the cookie does not have the MaxAge setting, and that kinit_lifetime properly causes the session to expire after the lifetime indicated.

abbra:
Pylint failed:
```
************* Module ipaserver.rpcserver
ipaserver/rpcserver.py:973: [E1101(no-member), login_password.kinit] Class 'env' has no 'kinit_lifetime' member)
make: *** [pylint] Error 2
```

Do we want to set `kinit_lifetime` during upgrade if it is missing?

simo5:
I thought just defining it as None in the constants was enough ?
We do not want to set a kinit_lifetime entry in defaults.conf, I am ok with the default being None for now I think.

simo5:
Change to used the correct bug number:
https://pagure.io/freeipa/issue/7001

abbra:
Looks like pylint is unhappy with such definitions. For most of other options from default.conf we have them explicitly set in `Env._bootstrap()`.

pvoborni:
@MartinBasti is the correct way how to solve this pylint issue to extend pylint_plugins.py?

snippet. of similar vars: 
```python
fake_api_env = {'env': [
    'host',
    'realm',
    'session_auth_duration',
    'session_duration_type',
]}

# this is due ipaserver.rpcserver.KerberosSession where api is undefined
fake_api = {'api': [fake_api_env] + NAMESPACE_ATTRS}
```

MartinBasti:
@pvoborni yes, this is the way how to handle false positive missing members in pylint

simo5:
Ok I added it to pylint_plugins, hopefully it is addressed fully now

MartinBasti:
master:

* 77db574cca900eb03d8e0ab577568c2058ae4a86 Add code to be able to set default kinit lifetime
* c52ca92cdabf07da5b3b27344b8eaa8b161ea49b Revert setting sessionMaxAge for old clients


ipa-4-5:

* 0def2ec653ac29210414aa09499b309dd1c3ac7d Add code to be able to set default kinit lifetime
* 728e2f681a938f25d091a8480ec6db3961ebfc98 Revert setting sessionMaxAge for old clients

PR #856, opened 2017-06-07
adtrustinstance: pep8, py3 fixes

stlaz:
There was a bug in adtrustinstance where there was a comparison
of a number to a string which was tailored so that it would always
pass (or at least in the default case). Luckily for us, Python 3 is a bit
more clever so it was throwing exceptions there.

We may want a ticket different from https://pagure.io/freeipa/issue/4985
for that but I am keeping that one for now, will only create a new one if
requested.

### DIFF
--- a/ipaserver/install/adtrustinstance.py
+++ b/ipaserver/install/adtrustinstance.py
@@ -354,7 +354,7 @@ class ADTRUSTInstance(service.Service):
 
             if abs(self.rid_base - self.secondary_rid_base) < size:
                 self.print_msg("Primary and secondary RID base are too close. "
-                      "They have to differ at least by %d." % size)
+                               "They have to differ at least by %d." % size)
                 raise RuntimeError("RID bases too close.\n")
 
             # Modify the range
--- a/ipaserver/install/adtrustinstance.py
+++ b/ipaserver/install/adtrustinstance.py
@@ -509,23 +509,17 @@ class ADTRUSTInstance(service.Service):
             self.print_msg(UPGRADE_ERROR % dict(dn=targets_dn))
 
     def __write_smb_registry(self):
-        template = os.path.join(paths.USR_SHARE_IPA_DIR, "smb.conf.template")
-        conf = ipautil.template_file(template, self.sub_dict)
-        [tmp_fd, tmp_name] = tempfile.mkstemp()
-        os.write(tmp_fd, conf)
-        os.close(tmp_fd)
-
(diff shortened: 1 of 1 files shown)


ipaserver/install/adtrustinstance.py
```diff
@@ -345,9 +345,14 @@ def __add_rid_bases(self):
 
             # Abort if RID bases are too close
             local_range = ranges_with_no_rid_base[0]
-            size = local_range.single_value.get('ipaIDRangeSize')
+            try:
+                size = int(local_range.single_value.get('ipaIDRangeSize'))
```
MartinBasti: How about when `'ipaIDRangeSize'` is unset (`None`)? Should we rather skip comparison `if abs(self.rid_base - self.secondary_rid_base) < size:` than fail hard?

ipaserver/install/adtrustinstance.py
```diff
@@ -509,23 +509,16 @@ def __add_s4u2proxy_target(self):
...
-        args = [paths.NET, "conf", "import", tmp_name]
-
-        try:
-            ipautil.run(args)
-        finally:
-            os.remove(tmp_name)
+        template = os.path.join(paths.USR_SHARE_IPA_DIR, "smb.conf.template")
+        conf = ipautil.template_file(template, self.sub_dict)
+        with tempfile.NamedTemporaryFile(mode='w') as tmp_conf:
+            tmp_conf.write(conf)
```
MartinBasti: `flush()` may be needed because another process is opening that file.

MartinBasti:
Please open a new ticket+PR for non-working ID Range comparison. We need clear tracking in case that this will be backported.

stlaz:
This PR now depends on https://github.com/freeipa/freeipa/pull/860.

ipaserver/install/adtrustinstance.py
stlaz: I don't know what are the implications of `ipaIDRangeSize` missing or whether we'd even get here if it was not here.
I am judging that it's a hard requirement, `int(None)` will lead to the failure nonetheless, I suppose it's OK to hard fail here. @abbra might have a different opinion.

MartinBasti:
Needs rebase

MartinBasti:
master:

* a83b2583ab9eaeffc1b37ceb9009bdc7d5de084c adtrustinstance: pep8 fix
* 8311069d1890d6b742b5aac17dfaaf79a6e1bf62 adtrustinstance: write the conf as a string

PR #857, opened 2017-06-07
server.py: Removes dns-server configuration from ldap

Tiboris:
After invocation of the `ipa server-del <hostname>`
command there was still record in ldap if DNS
was installed on the `<hostname>` server.

Fixes: https://pagure.io/freeipa/issue/6572

### DIFF
--- a/ipaserver/plugins/server.py
+++ b/ipaserver/plugins/server.py
@@ -692,6 +692,12 @@ class server_del(LDAPDelete):
                     message=_("You may need to manually remove them from the "
                               "tree")))
 
+    def _cleanup_server_dns_config(self, hostname):
+        try:
+            self.api.Command.dnsserver_del(hostname)
+        except errors.NotFound:
+            pass
+
     def pre_callback(self, ldap, dn, *keys, **options):
         pkey = self.obj.get_primary_key_from_dn(dn)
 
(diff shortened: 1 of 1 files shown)


MartinBasti:
NACK

Please create a new method `_cleanup_server_dns_config` because this patch unrelated to DNS records but it is for DNS configuration per server.

ipaserver/plugins/server.py
```diff
@@ -676,6 +676,16 @@ def _cleanup_server_dns_records(self, hostname, **options):
                 **options):
             return
 
+        dn = self.api.Object.dnsserver.get_dn(hostname)
+        assert isinstance(dn, DN)
+
+        ldap = self.obj.backend
+
+        try:
+            ldap.delete_entry(dn)
```
MartinBasti: you can use internal `api.Command.dnsserver_del()` command instead.

Tiboris: OK

MartinBasti:
Actually NACK until you fix commit message, this is not related to baseldap.py

Tiboris:
It should be better now.

MartinBasti:
master:

* 063211d665d02fc343952f5b158fd8d89223fbc9 server.py: Removes dns-server configuration from ldap


ipa-4-5:

* 005c92868ce36770ce89e87ef3cdeae62d11ece4 server.py: Removes dns-server configuration from ldap

PR #859, opened 2017-06-07
Add CommonNameToSANDefault to default cert profile

frasertweedale:
The CommonNameToSANDefault component was added to Dogtag 10.4.  When
a profile is configured to use it, this profile copies the CN in the
certificate to the Subjet Alternative Name extension as a dNSName
(if and only if it does look like a DNS name).

It is desirable that the default service profile use this component.
Add it to the default profile, for new installations only.  For
existing installations, until a proper profile update mechanism is
implemented, administrators who wish to use it must configure it via
the 'certprofile-mod' command.

Part of: https://pagure.io/freeipa/issue/4970

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -303,8 +303,8 @@ Requires(post): systemd-units
 Requires: selinux-policy >= %{selinux_policy_version}
 Requires(post): selinux-policy-base >= %{selinux_policy_version}
 Requires: slapi-nis >= %{slapi_nis_version}
-Requires: pki-ca >= 10.3.5-11
-Requires: pki-kra >= 10.3.5-11
+Requires: pki-ca >= 10.4.0-1
+Requires: pki-kra >= 10.4.0-1
 Requires(preun): python systemd-units
 Requires(postun): python systemd-units
 Requires: policycoreutils >= 2.1.12-5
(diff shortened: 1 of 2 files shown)


frasertweedale:
COPR with 10.4 builds: https://copr.fedorainfracloud.org/coprs/mharmsen/Dogtag-10.4.0/packages/

pki-core 10.4 builds should be added to the `@freeipa/freeipa-master` too.

MartinBasti:
How are upgrades of cert profile handled? (if they are needed)

frasertweedale:
Latest pki 10.4 builds are in this COPR: https://copr.fedorainfracloud.org/coprs/g/pki/10.4/packages/

MartinBasti:
IMHO this should be documented in release notes (changelog field in ticket) and howto document should be provided how to manually upgrade profile for SAN extensions after upgrade to dogtag 10.4

frasertweedale:
ipa-4-5 PR: https://github.com/freeipa/freeipa/pull/863

HonzaCholasta:
master:

* 1a35a2e213b46f3c5bb91d0f1b7fa05e8f051d4a (HEAD) Add CommonNameToSANDefault to default cert profile

PR #860, opened 2017-06-07
adtrustinstance: fix ID range comparison

stlaz:
The ID range comparison was comparing numbers to a string or possibly
to `None` and was tailored in such a way that the check would always
pass although it went directly against the definition of the absolute
value of a substitution.

https://pagure.io/freeipa/issue/7002

### DIFF
--- a/ipaserver/install/adtrustinstance.py
+++ b/ipaserver/install/adtrustinstance.py
@@ -345,9 +345,14 @@ class ADTRUSTInstance(service.Service):
 
             # Abort if RID bases are too close
             local_range = ranges_with_no_rid_base[0]
-            size = local_range.single_value.get('ipaIDRangeSize')
+            try:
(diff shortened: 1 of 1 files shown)


MartinBasti:
Is 'None' a valid value for `ipaIDRangeSize` ?

stlaz:
I don't think it is.

MartinBasti:
@stlaz could you provide more evidence? :-)

stlaz:
No, because I have no idea what that does. @abbra might know more.

abbra:
ipaIDRangeSize is used by `sidgen` and `range-check` plugins in 389-ds and also by SSSD. It cannot be None. In fact, it has to be more than 0 (and less than UINT32_MAX) or otherwise it is considered wrong.

MartinBasti:
master:

* 440c61dc40353833cad3a5fc509821ce1f23757f adtrustinstance: fix ID range comparison

PR #861, opened 2017-06-07
[IPA 4.5] Bump version of python-gssapi

pvomacka:
Complete fixing of the bug requires fix on python-gssapi side.
That fix is included in version 1.2.0-5.

Fixes: https://pagure.io/freeipa/issue/6796


MartinBasti:
ipa-4-5:

* 15d5ddd417d801a2356dcb043feef1aed8f76a25 Bump version of python-gssapi

PR #862, opened 2017-06-07
dnsserver.py: dnsserver-find no longer returns internal server error

Tiboris:
Invocation of the ipa dnsserver-find command failed with
internal server error when there is no DNS server in topology.

Fixes: https://pagure.io/freeipa/issue/6571


MartinBasti:
master:

* 74d36a8af69a2946007ebd4d57c7bf0891d561db dnsserver.py: dnsserver-find no longer returns internal server error

PR #863, opened 2017-06-09
[ipa-4-5] Add CommonNameToSANDefault to default cert profile

frasertweedale:
The CommonNameToSANDefault component was added to Dogtag 10.4.  When
a profile is configured to use it, this profile copies the CN in the
certificate to the Subject Alternative Name extension as a dNSName
(if and only if it does look like a DNS name).

It is desirable that the default service profile use this component.
Add it to the default profile, for new installations only.  For
existing installations, until a proper profile update mechanism is
implemented, administrators who wish to use it must configure it via
the 'certprofile-mod' command.

Part of: https://pagure.io/freeipa/issue/4970


HonzaCholasta:
ipa-4-5:

* 33aa4c25a2c3d158e43978d8699c3776d0e06599 (HEAD) Add CommonNameToSANDefault to default cert profile

PR #865, opened 2017-06-09
ipatests: do not collect systemd journal when logfile_dir is missing

tomaskrizek:
If logs aren't collected to logfile_dir, skip collection of systemd
journal.

Related https://pagure.io/freeipa/issue/6971


MartinBasti:
master:

* 44e3496bd1a3004bc7a6497cbd212bba7910b2e3 ipatests: do not collect systemd journal when logfile_dir is missing

PR #866, opened 2017-06-12
Add a README to certificate profile templates directory

frasertweedale:
There have been several instances of people using the profile
configuration template files as actual profile configurations,
resulting in failures and support load.  Add a README to the profile
template directory to explain that these files should not be used
and advise of the recommend procedure.

Fixes: https://pagure.io/freeipa/issue/7014


MartinBasti:
master:

* d7e1ab8438b02db9250b0985be29ac3325c2d2dc Add a README to certificate profile templates directory

PR #867, opened 2017-06-12
trust-mod: allow modifying list of UPNs of a trusted forest

abbra:
There are two ways for maintaining user principal names (UPNs) in Active
Directory:
 - associate UPN suffixes with the forest root and then allow for each
   user account to choose UPN suffix for logon
 - directly modify userPrincipalName attribute in LDAP

Both approaches lead to the same result: AD DC accepts user@UPN-Suffix
as a proper principal in AS-REQ and TGS-REQ.

The latter (directly modify userPrincipalName) case has a consequence
that this UPN suffix is not visible via netr_DsRGetForestTrustInformation
DCE RPC call. As result, FreeIPA KDC will not know that a particular UPN
suffix does belong to a trusted Active Directory forest. As result, SSSD
will not be able to authenticate and validate this user from a trusted
Active Directory forest.

This is especially true for one-word UPNs which otherwise wouldn't work
properly on Kerberos level for both FreeIPA and Active Directory.

Administrators are responsible for amending the list of UPNs associated
with the forest in this case. With this commit, an option is added to
'ipa trust-mod' that allows specifying arbitrary UPN suffixes to a
trusted forest root.

As with all '-mod' commands, the change replaces existing UPNs when
applied, so administrators are responsible to specify all of them:

  ipa trust-mod ad.test --upns={existing.upn,another_upn,new}

Fixes: https://pagure.io/freeipa/issue/7015

### DIFF
--- a/API.txt
+++ b/API.txt
@@ -5772,11 +5772,12 @@ output: ListOfEntries('result')
 output: Output('summary', type=[<type 'unicode'>, <type 'NoneType'>])
 output: Output('truncated', type=[<type 'bool'>])
 command: trust_mod/1
-args: 1,9,3
+args: 1,10,3
 arg: Str('cn', cli_name='realm')
 option: Str('addattr*', cli_name='addattr')
 option: Flag('all', autofill=True, cli_name='all', default=False)
 option: Str('delattr*', cli_name='delattr')
+option: Str('ipantadditionalsuffixes*', autofill=False, cli_name='upn_suffixes')
 option: Str('ipantsidblacklistincoming*', autofill=False, cli_name='sid_blacklist_incoming')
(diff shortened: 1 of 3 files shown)


martbab:
LGTM, the only little nitpick I have is that the CLI option should be named `--upn-suffixes` as `--upns` implies that you can specify full User principal names which you don't. You only specify suffixes.

martbab:
Nevermind I fixed this for @abbra. Let's wait for Travis and then we can push it.

martbab:
ipa-4-5:

* 9a31b21bff7c83219a4973adf815c900628ab620 trust-mod: allow modifying list of UPNs of a trusted forest


master:

* abb638487580af99882b4751b64939d0aff0d38b trust-mod: allow modifying list of UPNs of a trusted forest

PR #872, opened 2017-06-14
Add IPA-specific bind unit file

stlaz:
During upgrade of Fedora 25 to 26, when FreeIPA is installed with
DNS, bind attempts to start before KDC which leads to a failed
start because it requires a ticket to connect to LDAP.

Add an own unit file with a dependency which sets bind to start
after the KDC service.

https://pagure.io/freeipa/issue/7018


ipaserver/install/server/upgrade.py
```diff
@@ -1592,6 +1593,28 @@ def disable_httpd_system_trust(http):
...
+    use the new instead
+    """
+    root_logger.info('[Making bind use FreeIPA-specific unit file]')
+
+    if sysupgrade.get_upgrade_state('named-pkcs11.service',
+                                    'ipa_unit_file'):
+        root_logger.info("Already using the IPA-specific unit file")
+        return
+
+    bind_old = SystemdService('named-pkcs11', 'named-pkcs11.service', api=api)
```
MartinBasti: A platform specific constant should be used for this

MartinBasti:
I have a few comments:

* named-pkcs11 should be masked in installer and upgrader to avoid issues when users start incorrect named service manually
* please update release notes, this is a quite big change that we changed name of service

stlaz:
The last update should sort the masking of the named-pkcs11 service along the previously handled named. I will update the release notes once we settle on the form of the patch.

ipaserver/install/bindinstance.py
```diff
@@ -747,20 +752,24 @@ def __enable(self):
...
+            try:
+                svc.stop()
+            except Exception as e:
+                root_logger.debug(
+                    "Unable to stop {name} ({err})"
+                    .format(name=svc_name, err=e))
+            try:
+                svc.mask()
+            except Exception as e:
+                root_logger.debug(
```
MartinBasti: this is not a debug message, it will require manual intervention.

ipaserver/install/server/upgrade.py
```diff
@@ -1592,6 +1592,40 @@ def disable_httpd_system_trust(http):
...
+        try:
+            bind.start()
+        except Exception as e:
+            root_logger.warning(
+                'Unable to start ipa-named-pkcs11 service (%s)', e)
+
+    try:
+        bind_old.mask()
+    except Exception as e:
+        root_logger.warning('Unable to mask named service (%s)', e)
```
MartinBasti: named-pkcs11 service

MartinBasti:
I checked BZ, this may not be worth fixing as those fails are just during upgrade but at the end named is working. For sure this huge change cannot go to ipa-4-4 or ipa-4-5

stlaz:
We agreed that the change is unnecessary.

PR #873, opened 2017-06-14
kra: promote: Get ticket before attempting to get KRA keys with custodia

dkupka:
When installing second (or consequent) KRA instance keys are retrieved using
custodia. Custodia checks that the keys are synchronized in master's directory
server and the check uses GSSAPI and therefore fails if there's no ticket in
ccache.

### DIFF
--- a/ipaserver/install/kra.py
+++ b/ipaserver/install/kra.py
@@ -10,6 +10,7 @@ import os
 import shutil
 
 from ipalib import api
+from ipalib.install.kinit import kinit_keytab
 from ipaplatform import services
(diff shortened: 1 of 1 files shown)


stlaz:
The patch works but:
a) you're missing a ticket
b) commit message lines are too long

martbab:
master:

* 342f72140f9bd8b8db19f469ae4c56cac7492901 kra: promote: Get ticket before calling custodia


ipa-4-5:

* 15076a1c2b0fb31dce3903e5f50cab9edf68ad07 kra: promote: Get ticket before calling custodia

PR #874, opened 2017-06-14
Changing cert-find to go through the proxy instead of using the port 8080

felipevolpone:
The cert-find command now uses the proxy to reach Dogtag, instead of using the port 8080. In order to accomplish that, it's necessary to change the proxy configuration adding the service URL.

Ticket: https://pagure.io/freeipa/issue/6966

### DIFF
--- a/install/conf/ipa-pki-proxy.conf
+++ b/install/conf/ipa-pki-proxy.conf
@@ -1,4 +1,4 @@
-# VERSION 10 - DO NOT REMOVE THIS LINE
+# VERSION 11 - DO NOT REMOVE THIS LINE
 
 ProxyRequests Off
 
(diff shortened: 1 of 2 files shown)


rcritten:
When modifying the Apache config files you need to bump VERSION so the change is applied to existing installs.

felipevolpone:
@rcritten I forget that. Thanks for reminding me. Done.

frasertweedale:
All good.  Nice work @felipevolpone :+1:

tomaskrizek:
master:

* 36532031cfef893aa2f95d709efd807729de79f7 Changing cert-find to go through the proxy instead of using the port 8080


ipa-4-5:

* 960b9a3f1b1637c0ea5b11d6d7f671a09a3e89da Changing cert-find to go through the proxy instead of using the port 8080

PR #875, opened 2017-06-15
Fix ip address checks

MartinBasti:
Fix various checks of IP address in installers, removal of some unneeded checks that are not working correctly,  and mainly causes only false positive errors.

This PR also fixes regressions caused by bf9886a84393d1d1546db7e49b102e08a16a83e7

https://pagure.io/freeipa/issue/4317

### DIFF
--- a/install/tools/ipa-dns-install
+++ b/install/tools/ipa-dns-install
@@ -47,7 +47,9 @@ def parse_options():
                       default=False, help="print debugging information")
     parser.add_option("--ip-address", dest="ip_addresses", metavar="IP_ADDRESS",
                       default=[], action="append",
-                      type="ip", ip_local=True, help="Master Server IP Address. This option can be used multiple times")
+                      type="ip",
+                      help="Master Server IP Address. This option can be used "
+                           "multiple times")
     parser.add_option("--forwarder", dest="forwarders", action="append",
                       type="ip", help="Add a DNS forwarder. This option can be used multiple times")
     parser.add_option("--no-forwarders", dest="no_forwarders", action="store_true",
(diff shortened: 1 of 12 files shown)


dkupka:
Changes look good to me. But it uncovers another unwanted behaviour: With --setup-dns the installer adds IP and FQDN into /etc/hosts. This results in all traffic from local system towards it's FQDN being routed over the external IP.
I believe this is not a good idea but don't have strong opinion about fixing it in this PR or separate one. It's time to stop touching /etc/hosts.

MartinBasti:
@dkupka it is not aim of this PR to fix what you mentioned.

dkupka:
@MartinBasti You're right.

dkupka:
master:

* 82ad586f6cbf6e707add3c866ed4e37ade69b045 Fix local IP address validation
* cb48a49c80f4a11d2d16511e0f1366867320f153 ipa-dns-install: remove check for local ip address
* 0b69e44f16fbba6ab7ddef5a3e55bdabcfd6a8a6 refactor CheckedIPAddress class
* 6024165101677c844dc3bbb337e290df2e66eaf1 CheckedIPAddress: remove match_local param
* f9cba7d161f788c32336b66ff7c641f4a1ed2754 Remove ip_netmask from option parser
* 1b8dc1131c9ca7218efb8fe16dcce97f9f960be9 replica install: add missing check for non-local IP address
* f3537297bee2890c6b839750bb7a0a2cf904cdf9 Remove network and broadcast address warnings

dkupka:
@MartinBasti please rebase for ipa-4-5

MartinBasti:
Rebased in #881

PR #876, opened 2017-06-16
python-netifaces: update to reflect upstream changes

MartinBasti:
python-netifaces now provides IPv6 netmask in format mask/prefix. It
breaks freeipa as it is unexpected format for python-netaddr. We must
split netmask and provide only prefix for netaddr.

https://pagure.io/freeipa/issue/7021

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -197,6 +197,7 @@ class CheckedIPAddress(UnsafeIPAddress):
         :return: InterfaceDetails named tuple or None if no interface has
         this address
         """
+        root_logger.debug("Searching for an interface of IP address: %s", self)
         if self.version == 4:
             family = netifaces.AF_INET
         elif self.version == 6:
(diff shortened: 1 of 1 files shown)


martbab:
Shouldn't we bump requires on python-netifaces so that we don't accidentally pull in the older version that can break this new code?

pvoborni:
Could the checked addresses(input from netifaces) be logged? It is easier to debug if something goes wrong.

MartinBasti:
@martbab should work with both versions, I don't want to bump requires for this
@pvoborni It could, I'll update PR

martbab:
master:

* 52b43c7168d02e84865e31faa478d8a93be7a913 python-netifaces: update to reflect upstream changes


ipa-4-5:

* 56d04b3dccc967630d869006dfbd0003fcfedabe python-netifaces: update to reflect upstream changes

PR #877, opened 2017-06-18
LDAP ObjectClasses are case-insensitive

seriv:
In the IPA servers of the previous versions new users were created with objectClass capitalizations like 'posixAccount'. 
They should be treated as matching 'posixaccount'.


felipevolpone:
Closing this PR due to lack of activity.

pvoborni:
for completeness, see #950

PR #878, opened 2017-06-19
Make certificate an object

stlaz:
This is a WIP patchset (you can tell it's WIP by the lack of the commit
messages) which implements certificates as objects. This should
very much simplify the way certificates are processed in the whole
project and is a first part of this effort. This makes us not care about
the certificate format (=> no encoding/decoding) until we really need
to write it somewhere.

In the second part of this effort, there should be a Certificate parameter
to the framework which should further improve the format-agnostic
behavior.

This is a part of Python 3 effort which showed that it's necessary to
improve how certificates are worked with.


ipalib/x509.py
```diff
@@ -101,47 +105,171 @@ def strip_header(pem):
...
+        crypto_x509.load_pem_x509_certificate(data, backend=default_backend())
+    )
+
 
-    if (datatype == PEM):
-        data = strip_header(data)
-        data = base64.b64decode(data)
+def load_der_x509_certificate(data):
+    """
+    Load an X.509 certificate in PEM format.
```
MartinBasti: Please choose one

```load_der_x509_certificate```

VS

```Load an X.509 certificate in PEM format.```

File: ipalib/x509.py [hunk 2]
```diff
@@ -274,73 +364,61 @@ def normalize_certificate(rawcert):
...
     to DER regardless, then back out to PEM.
     """
-    dercert = normalize_certificate(rawcert)
 
     try:
-        fp = open(filename, 'w')
-        fp.write(make_pem(base64.b64encode(dercert)))
-        fp.close()
+        with open(filename, 'wb') as fp:
+            fp.write(cert)
```
MartinBasti: This should be: `fp.write(cert.public_bytes(Encoding.PEM))`

ipapython/certdb.py
```diff
@@ -570,34 +567,33 @@ def import_files(self, files, import_keys=False, key_password=None,
...
-            try:
-                ipautil.run(args)
-            except ipautil.CalledProcessError as e:
-                raise RuntimeError(
-                    "No matching certificate found for private key from %s" %
-                    key_file)
+            with tempfile.NamedTemporaryFile(), tempfile.NamedTemporaryFile() \
+                    as (in_file, out_file):
+                x509.write_certificate_list(extracted_certs, in_file.name)
+                in_file.write('\n{pkey}'.format(pkey=extracted_key))
```
MartinBasti: By default `NamedTemporaryFile` opens in binary mode, use `b` for binary data

ipapython/certdb.py
```diff
@@ -570,34 +567,33 @@ def import_files(self, files, import_keys=False, key_password=None,
...
+                out_password = ipautil.ipa_generate_password()
+                out_pwdfile = ipautil.write_tmp_file(out_password)
+                args = [
+                    OPENSSL, 'pkcs12',
+                    '-export',
+                    '-in', in_file.name,
+                    '-out', out_file.name,
+                    '-passin', 'file:' + self.pwd_file,
+                    '-passout', 'file:' + out_pwdfile.name,
+                ]
```
MartinBasti: You should `flush()` in_file

ipapython/certdb.py
```diff
@@ -637,9 +629,9 @@ def has_nickname(self, nickname):
 
     def export_pem_cert(self, nickname, location):
         """Export the given cert to PEM file in the given location"""
-        cert = self.get_cert(nickname, pem=True)
-        with open(location, "w+") as fd:
-            fd.write(cert)
+        cert = self.get_cert(nickname)
+        with open(location, "wb+") as fd:
```
MartinBasti: Please remove `+`, we need only write operation here

ipaserver/install/cainstance.py
```diff
@@ -791,23 +791,21 @@ def __export_ca_chain(self):
...
         # We have all the certificates in certlist, write them to a PEM file
         for path in [paths.IPA_CA_CRT,
                      paths.KDC_CA_BUNDLE_PEM,
                      paths.CA_BUNDLE_PEM]:
-            with open(path, 'w') as ipaca_pem:
+            with open(path, 'wb') as ipaca_pem:
                 for cert in certlist:
-                    ipaca_pem.write(cert)
+                    ipaca_pem.write(cert.public_bytes(x509.Encoding.PEM))
                     ipaca_pem.write('\n')
```
MartinBasti: bytes required

stlaz:
Most of the comments should be fixed now but I made some minor new changes which will most probably break a lot of stuff but what can you do.

stlaz:
The PR is getting more or less stable. There are some more things I would like to do:
- Make x509 extensions objects
- Make _IPACertificate properties cached

I will probably do that in another PR.

Now I just need to make this reviewable and I believe we could eventually push it.

stlaz:
I rebased the fix from https://github.com/freeipa/freeipa/pull/853 to this PR so that it is now clear how Certmonger is blocking us from introducing Python 3 support in FreeIPA.

The PR is ready to be reviewed once the tests pass.

ipapython/certdb.py
```diff
@@ -99,7 +97,7 @@ def find_cert_from_txt(cert, start=0):
     if s < 0 or e < 0:
         raise RuntimeError("Unable to find certificate")
 
-    cert = cert[s:e]
+    cert = x509.load_pem_x509_certificate(cert[s:e])
```
stlaz: `cert[s:e]` needs to be encoded, I did not realize it.

ipaserver/install/installutils.py
```diff
@@ -1051,7 +1051,7 @@ def load_pkcs12(cert_files, key_password, key_nickname, ca_cert_files,
             if ca_cert is None:
                 ca_cert = cert
 
-            cert_obj = x509.load_certificate(cert, x509.DER)
+            cert_obj = x509.load_der_x509_certificate(cert)
```
stlaz: We get the cert object from the NSS db already so this would fail.

ipapython/certdb.py
```diff
@@ -180,12 +178,11 @@ def unparse_trust_flags(trust_flags):
 
 
 def verify_kdc_cert_validity(kdc_cert, ca_certs, realm):
-    pem_kdc_cert = kdc_cert.public_bytes(serialization.Encoding.PEM)
-    pem_ca_certs = '\n'.join(
-        cert.public_bytes(serialization.Encoding.PEM) for cert in ca_certs)
+    pem_ca_certs = ''.join(
```
stlaz: This should have been bytes, I will make this uses the `x509.write_certificate_list` function instead.

ipaserver/install/certs.py
```diff
@@ -321,30 +319,20 @@ def load_cacert(self, cacert_fname, trust_flags):
...
             result = self.run_certutil(args, capture_output=True)
-            cert = result.output
-            if pem:
-                return cert
-            else:
-                cert, _start = find_cert_from_txt(cert, start=0)
-                cert = x509.strip_header(cert)
-                dercert = base64.b64decode(cert)
-                return dercert
+            return x509.load_pem_x509_certificate(result.output)
```
stlaz: Needs `result.raw_output`

ipaclient/install/client.py
```diff
@@ -1648,7 +1648,7 @@ def get_ca_certs_from_ldap(server, basedn, realm):
         root_logger.debug("get_ca_certs_from_ldap() error: %s", e)
         raise
 
-    certs = [x509.load_certificate(c[0], x509.DER) for c in certs
+    certs = [x509.load_der_x509_certificate(c[0]) for c in certs
```
stlaz: Should be `(c[0] for c in certs if c[2] is not False)` instead.

MartinBasti:
I'm curious how clients will behave with the new `Certificate` param:
* how old old clients with hardcoded bytes will work
* how old thin clients without knowledge of Certificate object will work.
This must be tested, I'm afraid that those params must stay `Bytes` and Certificate object must be used only internally on server (as I proposed originally)

stlaz:
@HonzaCholasta showed me that there are still some changes to be done. Lucky he came to work for the last day today.
AFAIK old clients will be sending a base64 encoded strings, that's fine.
As for thin client, that one sends a string if it does not know the parameter, so we should be fine there as well.
I'll test the two scenarios.

stlaz:
So this PR is getting stable again.
I tested 4.4 client against a server with this version and adding and displaying of certificates worked fine. Thus older clients should work as well since they are sending the same information, but I'll rather double-check.

stlaz:
FTR: RHEL 6.9 client is ok with this as well, except there's a bug (this PR does not have any influence on it) which makes `cert-show` fail with `ValueError: cert_show.validate_output(): unexpected keys ['summary', 'value'] in ...`, but the values of the `certificate` field are ok there.

rcritten:
Why split load_certificate into separate functions?

stlaz:
To more resemble API of python-cryptography so that we may possibly one day remove all such calls and replace them should they implement all the things that are currently missing.

stlaz:
I am raising the priority of this PR. I would like to have a review when I get back on Wednesday so that I don't have to be blocked in the Python 3 porting effort.

frasertweedale:
I'll review it.

ipaplatform/redhat/tasks.py
```diff
@@ -302,14 +298,12 @@ def insert_ca_certs_into_systemwide_ca_store(self, ca_certs):
                 obj += "trusted: true\n"
             elif trusted is False:
                 obj += "x-distrusted: true\n"
-            obj += "%s\n\n" % cert
+            obj += "%s\n\n" % cert.public_bytes(x509.Encoding.PEM)
```
stlaz: Will do!

ipaserver/plugins/host.py
```diff
@@ -27,9 +27,10 @@
 import six
```
stlaz: I did notice this in user.py plugin but forgot to check with the other two. I'll fix it, thanks.

ipalib/x509.py
```diff
@@ -101,47 +104,353 @@ def strip_header(pem):
...
+    Only use this function when you can't be sure what kind of format does
+    your certificate have, e.g. input certificate files in installers
+
+    :returns: a ``_IPACertificate`` object.
+    :raises: ``ValueError`` if unable to load the certificate.
+    """
+    try:
+        return load_pem_x509_certificate(data)
+    except ValueError:
+        return load_der_x509_certificate(data)
```
stlaz: It does not seem we're supporting it now so I'd go with what we currently have.

frasertweedale: The old `x509.load_certificate(..., pem=True)` function does handle plain base64,
via `strip_headers()` which returns data unchanged if there are no headers.  The
stripped data then gets base64-decoded and parsed as DER.

I don't mind if we no longer support this, but neither will I be surprised when someone's
tests / deployment scripts / whatever breaks :)

stlaz:
I have fixed the issues with the `--out` option for host and certificate plugins and replace string interpolation to use the `str.format()` method instead.

I mean to keep the "TODO: REMOVE THIS" commit there since Certmonger still sends an ugly OpenSSL-unreadable PEM and I did not have yet time to fix it. I will reword the commit message since we probably don't want "TODO" messages in upstream git, I just wanted to make it very ugly so that people notice and I don't want this workaround in an upstream release. On the other hand, I don't want to wait with this PR to be pushed until the changes in Certmonger are done since it blocks Python 3 migration progression.

ipalib/x509.py
stlaz: I don't think this was used for the cases where `load_unknown_x509_certificate()` is but rather for reading certificates in the framework where the user is supposed to supply the certificate as a base64-encoded DER. This is now handled it in the `Certificate` parameter which converts anything coming to the framework to `_IPACertificate`.

This reminds me I wanted to rename `_IPACertificate` to `IPACertificate` -_-

frasertweedale:
On Wed, Jul 26, 2017 at 01:00:40AM -0700, Stanislav Láznička wrote:
> I have fixed the issues with the `--out` option for host and certificate plugins and replace string interpolation to use the `str.format()` method instead.
> 
> I mean to keep the "TODO: REMOVE THIS" commit there since Certmonger still sends an ugly OpenSSL-unreadable PEM and I did not have yet time to fix it. I will reword the commit message since we probably don't want "TODO" messages in upstream git, I just wanted to make it very ugly so that people notice and I don't want this workaround in an upstream release. On the other hand, I don't want to wait with this PR to be pushed until the changes in Certmonger are done since it blocks Python 3 migration progression.
> 
I agree we do not want "TODO" in the commit message.  I'd suggest
opening another issue to track the Certmonger thing (possibly
against Certmonger itself?) and reference that ticket URL in the
commit message.

stlaz:
Thanks for the suggestion. I actually did open the issue a month ago (https://pagure.io/certmonger/issue/76), I tried to bump it today. Also, I forgot to add it anywhere in the comments, it's now also there with a big fat doctext ;)

stlaz:
pep8 finally woke up in Travis and found some issues, I fixed those.

pvomacka:
master:

* 4375ef860fdd8221baeff23e88f9217b0cddc5ac Split x509.load_certificate() into PEM/DER functions
* b5732efda6d27f588adbc3f41259bc4511716f43 x509: Make certificates represented as objects
* 43c74d33337d4c65947816ba31de5eaa53e001ff Introduce load_unknown_x509_certificate()
* bf4dae70e0d163f4d485771d7d163169748fa6b3 tests: fix failing HTTPS connection
* 5ff1de84909dd65c44b9aa48000b9e73a7a93716 parameters: relax type checks
* 5a44ca638310913ab6b0c239374f4b0ddeeedeb3 Create a Certificate parameter
* 1521296297d99ea9f6e8f67b30d8e11c6bd56426 x509,certdb: handle certificates as bytes
* e1f88c844e704246e4a948dc74489513f66633d5 Fixup of not-so-good PEM certs

PR #879, opened 2017-06-19
FIPS mode and NT hashes

sumit-bose:
In FIPS mode NT hashes (aka md4) are not allowed. If FIPS more is detected we
disable NT hashes in the password plugin even is they are allowed by IPA
configuration.

Since ipa-sam is running as part of smbd is it safe to use the E_md4hash()
from Samba. This way ipa-sam does not depend on other crypto libraries
which might depend on other rules like e.g. FIPS mode.

### DIFF
--- a/daemons/ipa-sam/ipa_sam.c
+++ b/daemons/ipa-sam/ipa_sam.c
@@ -110,6 +110,7 @@ char *sid_string_dbg(const struct dom_sid *sid); /* available in libsmbconf.so *
 char *escape_ldap_string(TALLOC_CTX *mem_ctx, const char *s); /* available in libsmbconf.so */
 bool secrets_store(const char *key, const void *data, size_t size); /* available in libpdb.so */
 void idmap_cache_set_sid2unixid(const struct dom_sid *sid, struct unixid *unix_id); /* available in libsmbconf.so */
+bool E_md4hash(const char *passwd, uint8_t p16[16]); /* available in libcliauth-samba4.so */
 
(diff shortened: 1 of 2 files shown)


abbra:
LGTM. @stlaz, please test it in FIPS mode enabled.

pvoborni:
Ticket for the fix: https://pagure.io/freeipa/issue/7026

sumit-bose:
I just pushed a new version which include the upstream ticket reference in the commit message, there are no changes to the code.

stlaz:
@sumit-bose Thank you for your patch. I can positively say that the error no more breaks server installation in FIPS mode, I was also able to set up a trust against my own AD server.
I did see some errors but those turn out to be probably caused by some misconfiguration of my system since your patches most certainly don't affect them.
Therefore I am giving an ACK to this PR :)

pvomacka:
master:

* f169481b558bd7ec9102e02e40eb38df7867a694 ipa-sam: replace encode_nt_key() with E_md4hash()
* 1f0ca6aafd966e086692007a0df76e3c3276915f ipa_pwd_extop: do not generate NT hashes in FIPS mode


ipa-4-5:

* b63b6790efc82c87398c39ba4f55330756b7b3cf ipa-sam: replace encode_nt_key() with E_md4hash()
* 79a5f3bf321f15e4c120d16b8988ed0cdb0ae64c ipa_pwd_extop: do not generate NT hashes in FIPS mode

PR #25, opened 2016-08-26
Added install check before executing ipa-* command

Akasurde:
Fixes: https://fedorahosted.org/freeipa/ticket/6261

Signed-off-by: Abhijeet Kasurde akasurde@redhat.com

PR #880, opened 2017-06-19
Changing how commands handles error when it can't connect to IPA server

felipevolpone:
The commands that connects with IPA server can raise a `NetworkError` with the message: "ipa: ERROR: can't connect to `http://localhost:8888/ipa/json': [Errno 111] Connection refused`. Instead of that, this changes the message error in order to be more user-friendly.

I've used the `GenericError` because it inherits from `PublicError`and do not have a default message. So, I do not have to change the `run` method in `ipalib/cli.py` to handle a different exception/case.

Ticket: https://pagure.io/freeipa/issue/6261

### DIFF
--- a/ipalib/__init__.py
+++ b/ipalib/__init__.py
@@ -923,7 +923,7 @@ from ipalib.crud import Create, Retrieve, Update, Delete, Search
 from ipalib.parameters import DefaultFrom, Bool, Flag, Int, Decimal, Bytes, Str, IA5Str, Password, DNParam
 from ipalib.parameters import (BytesEnum, StrEnum, IntEnum, AccessTime, File,
                         DateTime, DNSNameParam)
-from ipalib.errors import SkipPluginModule
+from ipalib.errors import SkipPluginModule, GenericError, NetworkError
 from ipalib.text import _, ngettext, GettextFactory, NGettextFactory
 
 Registry = plugable.Registry
@@ -942,12 +942,17 @@ class API(plugable.API):
                 ipaserver.plugins,
             )
         else:
-            import ipaclient.remote_plugins
-            import ipaclient.plugins
-            result = (
-                ipaclient.remote_plugins.get_package(self),
-                ipaclient.plugins,
-            )
+            try:
+                import ipaclient.remote_plugins
+                import ipaclient.plugins
+                result = (
+                    ipaclient.remote_plugins.get_package(self),
+                    ipaclient.plugins,
+                )
+            except NetworkError:
+                #  instead of raising the default error connection message,
+                #  raising a more user-friendly one
+                raise GenericError('Cannot find IPA server to contact')
 
         if self.env.context in ('installer', 'updates'):
             # pylint: disable=import-error,ipa-forbidden-import


felipevolpone:
The ticket describe some commands that are not showing the right message. IMO we could split it into one ticket per command.

ipalib/__init__.py
```diff
@@ -942,12 +942,17 @@ def packages(self):
...
+                import ipaclient.remote_plugins
+                import ipaclient.plugins
+                result = (
+                    ipaclient.remote_plugins.get_package(self),
+                    ipaclient.plugins,
+                )
+            except NetworkError:
+                #  instead of raising the default error connection message,
+                #  raising a more user-friendly one
+                raise GenericError('Cannot find IPA server to contact')
```
stlaz: This does not help anything at all. It actually makes the situation much worse. You are basically only omitting useful information for no reason. Please don't do that, you yourself can probably tell that it's not by a smallest bit clever.

felipevolpone: any suggestion?

felipevolpone: what about something like: 
```python
except NetworkError as e:
    if 'localhost' in e.uri:
        raise GenericError('Cannot find IPA server to contact')
```

stlaz: I don't think that's correct, either. I agree that the ticket's formatting makes it tough to get the grasp of it, but it does not say that the message printed is incorrect, it mentions that it appears at wrong time - when IPA is not installed.

From the BZ:
```
Basically all the commands, if possible should check if ipa
server is installed in the first place and if not then it should display
message like 'IPA is not configured on this system" rather than displaying
Network error or any other message.
```
I believe that we indeed could check whether IPA is actually installed before doing anything, this is already implemented in some of our scripts (e.g. `ipa-ca-install`).

I am not sure, however, what's the scope of such a change, and whether such a change brings enough value given possible time spent to prioritize it.

You should probably also check the discussion at https://github.com/freeipa/freeipa/pull/25 (not sure whether you did already) for some more inputs on this.

felipevolpone: I wasn't trying to just change the message, but check if IPA is well configured. I thought the message `can't connect to http://localhost:8888/ipa/json` was saying that.

I have checked the PR#25; to fix independent scripts seems easier than apply the same  behavior to all commands. Since the changes could be done locally. 

I was thinking in to do something like in [ipalib/plugable.py](https://github.com/freeipa/freeipa/blob/master/ipalib/plugable.py#L706-L715):
```python
    from ipaserver.install import installutils

    def finalize(self):
        """
        Finalize the registration, instantiate the plugins.

        `API.bootstrap` will automatically be called if it hasn't been
        already.
        """
        
        installutils.check_server_configuration()

        self.__doing('finalize')
        self.__do_if_not_done('load_plugins')
```

However, it doesn't seems a good idea import ipaserver code in ipalib.

stlaz: That indeed is not a good idea since the api is finalized on client as well so it would most probably be always failing on clients.

I wonder whether we could always say that when `uri == http://localhost:8888/ipa/json`, the client is not installed. If so, there would perhaps be a way to do this check even earlier.
I think @MartinBasti might could give us pointer on the `uri` handling and whether `localhost` is special enough for us to treat it that way.

MartinBasti: localhost may be used for local in-tree testing

MartinBasti:
You may need https://github.com/freeipa/freeipa/pull/492 to reliably determine when ipa client is not configured

stlaz:
I see we have two PRs trying to solve the same thing (this and https://github.com/freeipa/freeipa/pull/25). Both are kind of stuck and I don't see them changed after https://github.com/freeipa/freeipa/pull/492 was pushed. I am closing both of them since neither seems to reflect the changes. Feel free to open a new one/reopen them when this is fixed or if you think your solution still applies.

PR #882, opened 2017-06-20
Py3 fixes

MartinBasti:


ipatests/test_ipapython/test_session_storage.py
```diff
@@ -23,13 +23,14 @@ def setup(self):
         self.principal = 'admin'
         self.key = 'X-IPA-test-session-storage'
         self.data = 'Test Data'
```
stlaz: `data` in `store_data` should be bytes anyway, why not just `b'Test Data'`?

MartinBasti: `self.data` are used as parameter, so I keep the same type

stlaz: The first thing `store_session` does is it converts it to bytes

stlaz:
Since the tests pass, I suppose we can push this. ACK.

MartinBasti:
master:

* 8416d5772d5e14829df58dabb24b30721c302c07 py3: Remove comparison >=2 of debnug log level
* 24eadd3a39f472669585355c5db00fc430acac1b py3: ipa_otptoken_import: fix lamba code inspection
* e53674e741fa8a2268e4a663c3ffbc00d891123c py3: ipa_otptoken_import: fix calling unicode on bytes
* 637d259361306bcbc4af86bbedf17a7ff6875716 py3: ipa_otptoken_import: fix hex decoding
* 902f736a2b288fac5d080b7f146f6a8b16f882f1 py3: test_otptoken_import: fix bytes usage

PR #883, opened 2017-06-20
Travis: check for BytesWarnings in httpd error_log

MartinBasti:
This check should prevent regressions in already py3 ported server plugins.
Later it may be extened to multiple logs.

### DIFF
--- a/.test_runner_config.yaml
+++ b/.test_runner_config.yaml
@@ -62,6 +62,7 @@ steps:
   - ipa-test-config --help
   - ipa-test-task --help
   - ipa-run-tests ${tests_ignore} -k-test_dns_soa ${tests_verbose} ${path}
+  - '! grep -n -C5 BytesWarning /var/log/httpd/error_log'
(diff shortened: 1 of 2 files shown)


.test_runner_config.yaml
```diff
@@ -62,6 +62,7 @@ steps:
   - ipa-test-config --help
   - ipa-test-task --help
   - ipa-run-tests ${tests_ignore} -k-test_dns_soa ${tests_verbose} ${path}
+  - '! grep -n -B10 BytesWarning /var/log/httpd/error_log'
```
stlaz: Wouldn't BytesWarning appear only in Python3?

MartinBasti: There is one check that is in py2 (I don't remember from top of my head), but majority of checks are in py3

MartinBasti:
master:

* b43dab83880b8ed99ffab3ed8b4b69e5da80177c Travis: check for BytesWarnings in httpd error_log

PR #885, opened 2017-06-20
NULL LDAP context in call to ldap_search_ext_s during search

tbordaz:
KDC crashes on quite random interval while trying to reach LDAP
    https://pagure.io/freeipa/issue/7017

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb.c
+++ b/daemons/ipa-kdb/ipa_kdb.c
@@ -465,6 +465,12 @@ int ipadb_get_connection(struct ipadb_context *ipactx)
     ret = ipadb_reinit_mspac(ipactx, false);
     if (ret && ret != ENOENT) {
         /* TODO: log that there is an issue with adtrust settings */
+        if (ipactx->lcontext == NULL) {
+            /* for some reason ldap connection was reset in ipadb_reinit_mspac
+             * and is no longer established => failure of ipadb_get_connection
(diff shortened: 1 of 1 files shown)


abbra:
We need to find out why ipadb->lcontext is wiped and not replaced.

tbordaz:
The connection handling looks complex even if I admit it is always a difficult part of a server.
I recorded a scenario where ipadb->lcontext got wiped without error reported to callers (https://pagure.io/freeipa/issue/7017#comment-445619).

The scenario means KDC was able to establish connections to DS but for some reasons consecutive search req were failing. Then KDC either failed to establish a new connection or bind or retrieve rootdse or config.

abbra:
I think we can merge this small fix. We still need to find out where and why ipa-db->lcontext is wiped but this fix is not worse than the previous one and at least addresses a potential issue. So, LGTM.

Please rebase before ACK.

tbordaz:
Thanks for the review and the ack.
Rebase pushed

abbra:
ACK.

pvomacka:
master:

* 0f0423cf531d5fab73516323750af6b5106ba070 NULL LDAP context in call to ldap_search_ext_s during search


ipa-4-5:

* 1cb93edd5bbfbca51362ca8dd3bf7c40d2020218 NULL LDAP context in call to ldap_search_ext_s during search

PR #886, opened 2017-06-21
*config-show: do not show empty roles/attributes

martbab:
Do not break the intended semantics of optional Params in *config-show commands which states that if the attribute is not present (or there is no read access to it) the attribute will not be sent in the JSON-RPC response.

Previously, server roles/attributes containing an empty tuple were sent and we should maintain the consistency with other API commands.

https://pagure.io/freeipa/issue/7029

### DIFF
--- a/ipaserver/plugins/config.py
+++ b/ipaserver/plugins/config.py
@@ -278,8 +278,7 @@ class config(LDAPObject):
 
         role_config = backend.config_retrieve(role_name)
         for key, value in role_config.items():
-            if value:
-                entry_attrs.update({key: value})
+            entry_attrs.update({key: value})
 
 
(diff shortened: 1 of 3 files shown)


MartinBasti:
master:

* f4d77533f5e0cb306da2aeb9548d891c5264c088 *config-show: Do not show empty roles/attributes
* e2e380e83be8bafd8cf23e0a395edf065b1ae961 Do not remove the old masters when setting the attribute fails


ipa-4-5:

* 2431c76775a1e314a3a03f608bb7aa776d3c8bf2 *config-show: Do not show empty roles/attributes
* 03a30c0fb2748d5724112e702567c71cbd19d624 Do not remove the old masters when setting the attribute fails

PR #888, opened 2017-06-21
Make py3 default for ported scripts

MartinBasti:


stlaz:
LGTM, I just need to test whether it's working correctly.

stlaz:
Seems to work, ACK.

pvomacka:
master:

* 17103e53cbbf6d4cfb42d727d3f2edb7b56c8e9c py3: temporary set dependencies to both py2 and py3 packages
* aa1c0cf3e8f2a77e27d548f73694d99aa0497459 py3: run already ported scripts under py3 by default

PR #890, opened 2017-06-22
Make sure we check ccaches in all rpcserver paths

simo5:
We need to verify the ccache is avcailable in all cases or finalize
will cause us to acquire creds with the keytab which is not what we
want.

### DIFF
--- a/ipaserver/rpcserver.py
+++ b/ipaserver/rpcserver.py
@@ -592,6 +592,41 @@ class KerberosSession(HTTP_Status):
     needing this do not share a common base class.
     '''
 
+    def need_login(self, start_response):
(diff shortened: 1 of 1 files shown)


stlaz:
Works for me. I created a ticket for the issue, if you could possibly add it to the commit message: https://pagure.io/freeipa/issue/7037

stlaz:
Thanks!

MartinBasti:
master:

* 0537ab07bad852381357df3b93d5ca7d30b1b5ec Make sure we check ccaches in all rpcserver paths


ipa-4-5:

* 89463c0c527142c18b44f5b367fa8b2c683df011 Make sure we check ccaches in all rpcserver paths

PR #891, opened 2017-06-23
Add SKI and AKI to CA certs in ca-less integration test

frasertweedale:
The IPA installer now checks that CA certs include the Subject Key
Identifier extension (which is required by Dogtag and RFC 5280).
But this broke our integration tests, which were not adding the
extension.

Update the caless-create-pki script to add these extensions.

The Subject Key Identifier and Authority Key Identifier values are
randomly chosen for each CA, and propagated down to the 'gen_cert()'
subroutine so that profiles have access to them.  Each profile can
choose how to use it.  For now, only the 'ca' profile uses them, but
for maximum correctness the 'server' profile (i.e. for leaf
certificates) could be updated to add the CA's SKI to the AKI
extension.  This is left for a later commit.

Fixes: https://pagure.io/freeipa/issue/7030

### DIFF
--- a/ipatests/test_integration/scripts/caless-create-pki
+++ b/ipatests/test_integration/scripts/caless-create-pki
@@ -3,8 +3,9 @@
 # Copyright (C) 2017  FreeIPA Contributors see COPYING for license
 #
 
-profile_ca_request_options=(-1 -2 -4)
-profile_ca_request_input="\$'0\n1\n5\n6\n9\ny\ny\n\ny\n1\n7\nfile://'\$(readlink -f \$dbdir)/\$ca.crl\$'\n-1\n-1\n-1\nn\nn\n'"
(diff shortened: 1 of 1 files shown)


Rezney:
Hi Fraser, thank you very much for taking a look into this. When I was opening case 7030 I was aiming for using Jan Cholasta`s makepki.py script. What do you think about this way of handling it? As for now there is no p12 but I will have prepared PR for this soon too.

https://github.com/freeipa/freeipa-tools/pull/54

Of course for the moment we can leave it as it is.

frasertweedale:
@Rezney sure, I don't mind if we switch to using the other script.  It seems to use python-cryptography
which is certainly more readable and maintainable than arcane certutil incantations.

frasertweedale:
Closing in favour of: https://github.com/freeipa/freeipa/pull/899

PR #892, opened 2017-06-23
Always check peer has keys before connecting

simo5:
When pulling the DM password we may have the same issues reported in
ticket #6838 for CA keys.
This commit makes sure we always check the peer has keys before any
client operation.

### DIFF
--- a/ipaserver/install/custodiainstance.py
+++ b/ipaserver/install/custodiainstance.py
@@ -13,7 +13,6 @@ from ipaserver.install import ldapupdate
 from ipaserver.install import sysupgrade
 from base64 import b64decode
 from jwcrypto.common import json_decode
-import functools
 import shutil
 import os
 import stat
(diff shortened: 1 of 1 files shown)


Rezney:
Run the test (TestLineTopologyWithoutCA) where I was having the issue 15x with this patch and did not hit it anymore.

stlaz:
Wouldn't this make CA uninstallable on a replica in a CA-less domain?

simo5:
I do not see how, the keys we are looking for are custodia keys and they are always created by us and then replicated, we are just making sure they are replicated to the other master before continuing.
custodia runs regardless of the fact the other server is or not a CA or whether a CA exists at all.

stlaz:
Ah, sorry, I got carried away while thinking about some other issues.

simo5:
@stlaz do we have anything blocking this one ?

stlaz:
Yes, the problem was that nobody was actually reviewing this, I am going to have a review morning tomorrow so I'll have a look.

stlaz:
Tested three different topologies and with CA-less. Seems to work fine, thanks! ACK.

pvomacka:
master:

* c565fa1781121cf284886e1f2d0691cfe10760d2 Always check peer has keys before connecting


ipa-4-5:

* 2dccb91792f9369a5fe0248f8d2a3e95081737c3 Always check peer has keys before connecting

PR #894, opened 2017-06-23
Fixing ipa-replica-install --setup-kra if it's the first KRA in topology

felipevolpone:
I'm trying to fix the ticket, but I'm not quite sure of how to do it. Until now, I removed the exception and called the api in kra to install it. However, I'm getting an exception:
```
bash-4.3$ sudo python /usr/sbin/ipa-replica-install -r DOM-116.ABC.IDM.LAB.ENG.BRQ.REDHAT.COM --setup-kra --setup-ca
WARNING: conflicting time&date synchronization service 'chronyd' will be disabled in favor of ntpd

IPA client is already configured on this system, ignoring the --domain, --server, --realm, --hostname, --password and --keytab options.

ipa.ipapython.install.cli.install_tool(CompatServerReplicaInstall): ERROR    Timed out trying to obtain keys.
ipa.ipapython.install.cli.install_tool(CompatServerReplicaInstall): ERROR    The ipa-replica-install command failed. See /var/log/ipareplica-install.log for more information
```

from /var/log/ipareplica-install.log  
```
2017-06-23T18:38:44Z DEBUG stderr=
2017-06-23T18:38:44Z DEBUG Destroyed connection context.ldap2_140135237350736
2017-06-23T18:38:44Z DEBUG Created connection context.ldap2_140135237350736
2017-06-23T18:38:44Z DEBUG raw: hostgroup_show(u'ipaservers', rights=True, all=True, version=u'2.228')
2017-06-23T18:38:44Z DEBUG hostgroup_show(u'ipaservers', rights=True, all=True, raw=False, version=u'2.228', no_members=False)
2017-06-23T18:38:44Z DEBUG flushing ldaps://vm-116.abc.idm.lab.eng.brq.redhat.com from SchemaCache
2017-06-23T18:38:44Z DEBUG retrieving schema for SchemaCache url=ldaps://vm-116.abc.idm.lab.eng.brq.redhat.com conn=<ldap.ldapobject.SimpleLDAPObject instance at 0x7f73c6769f38>
2017-06-23T18:38:44Z DEBUG Destroyed connection context.ldap2_140135237350736
2017-06-23T18:38:44Z DEBUG Created connection context.ldap2_140135237350736
2017-06-23T18:38:44Z DEBUG flushing ldaps://vm-116.abc.idm.lab.eng.brq.redhat.com from SchemaCache
2017-06-23T18:38:44Z DEBUG retrieving schema for SchemaCache url=ldaps://vm-116.abc.idm.lab.eng.brq.redhat.com conn=<ldap.ldapobject.SimpleLDAPObject instance at 0x7f73c6769c20>
2017-06-23T18:38:44Z DEBUG No IPA DNS servers, skipping forward/reverse resolution check
2017-06-23T18:38:44Z DEBUG Initializing principal host/vm-058-064.abc.idm.lab.eng.brq.redhat.com@DOM-116.ABC.IDM.LAB.ENG.BRQ.REDHAT.COM using keytab /etc/krb5.keytab
2017-06-23T18:38:44Z DEBUG using ccache /tmp/krbcc9omA2g/ccache
2017-06-23T18:38:44Z DEBUG Attempt 1/1: success
2017-06-23T18:38:44Z DEBUG Loading StateFile from '/var/lib/ipa/sysrestore/sysrestore.state'
2017-06-23T18:38:44Z DEBUG Loading Index file from '/var/lib/ipa/sysrestore/sysrestore.index'
2017-06-23T18:38:44Z INFO Waiting up to 300 seconds to see our keys appear on host: None
2017-06-23T18:38:45Z DEBUG Transient error getting keys: '{'desc': "Can't contact LDAP server"}'
2017-06-23T18:43:45Z DEBUG Destroyed connection context.ldap2_140135237350736
2017-06-23T18:43:45Z DEBUG   File "/usr/lib/python2.7/site-packages/ipapython/admintool.py", line 172, in execute
    return_value = self.run()
  File "/usr/lib/python2.7/site-packages/ipapython/install/cli.py", line 333, in run
    cfgr.run()
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 366, in run
    self.validate()
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 375, in validate
    for _nothing in self._validator():
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 434, in __runner
    exc_handler(exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 458, in _handle_validate_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 453, in _handle_exception
    six.reraise(*exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 424, in __runner
    step()
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 421, in <lambda>
    six.reraise(*exc_info)
    value = gen.send(prev_value)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 636, in _configure
    next(validator)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 434, in __runner
    exc_handler(exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 458, in _handle_validate_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 521, in _handle_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 453, in _handle_exception
    six.reraise(*exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 518, in _handle_exception
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 453, in _handle_exception
    six.reraise(*exc_info)
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 424, in __runner
    step()
  File "/usr/lib/python2.7/site-packages/ipapython/install/core.py", line 421, in <lambda>
    six.reraise(*exc_info)
    value = gen.send(prev_value)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/__init__.py", line 613, in main
    replica_promote_check(self)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/replicainstall.py", line 386, in decorated
    func(installer)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/replicainstall.py", line 408, in decorated
    func(installer)
  File "/usr/lib/python2.7/site-packages/ipaserver/install/server/replicainstall.py", line 1270, in promote_check
    raise ScriptError(e)

2017-06-23T18:43:45Z DEBUG The ipa-replica-install command failed, exception: ScriptError: Timed out trying to obtain keys.
2017-06-23T18:43:45Z ERROR Timed out trying to obtain keys.
2017-06-23T18:43:45Z ERROR The ipa-replica-install command failed. See /var/log/ipareplica-install.log for more information
```

There is a high chance that I'm getting the wrong path here, so if there is someone able to help me (pointing to some docs or explaining more details of it), it would be great.

Ticket https://pagure.io/freeipa/issue/7008

### DIFF
--- a/ipaserver/install/kra.py
+++ b/ipaserver/install/kra.py
@@ -16,6 +16,7 @@ from ipaplatform.paths import paths
 from ipapython import certdb
 from ipapython import ipautil
 from ipapython.install.core import group
+from ipapython.ipa_log_manager import root_logger
 from ipaserver.install import custodiainstance
 from ipaserver.install import cainstance
 from ipaserver.install import krainstance
@@ -26,7 +27,8 @@ from . import dogtag
 
 
 def install_check(api, replica_config, options):
-    if replica_config is not None and not replica_config.setup_kra:
+
+    if not options.setup_kra:
         return
 
     kra = krainstance.KRAInstance(api.env.realm)
@@ -47,9 +49,9 @@ def install_check(api, replica_config, options):
 
     if replica_config is not None:
         if not api.Command.kra_is_enabled()['result']:
-            raise RuntimeError(
-                "KRA is not installed on the master system. Please use "
-                "'ipa-kra-install' command to install the first instance.")
+            root_logger.info("KRA is not installed on the master system. "
+                             "KRA will be installed on the replica.")
+            return
 
         if options.promote:
             return
@@ -69,9 +71,13 @@ def install_check(api, replica_config, options):
 
 
 def install(api, replica_config, options):
-    if replica_config is None:
-        if not options.setup_kra:
-            return
+
+    if not options.setup_kra:
+        return
+
+    if not api.Command.kra_is_enabled()['result']:
+        # if kra is not installed on master, will install it for
+        # the first time in the replica
         realm_name = api.env.realm
         dm_password = options.dm_password
         host_name = api.env.host
@@ -81,8 +87,6 @@ def install(api, replica_config, options):
         master_host = None
         promote = False
     else:
-        if not replica_config.setup_kra:
-            return
         krafile = os.path.join(replica_config.dir, 'kracert.p12')
         if options.promote:
             with ipautil.private_ccache():
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -824,14 +824,14 @@ def install_check(installer):
             logger.debug('No IPA DNS servers, '
                          'skipping forward/reverse resolution check')
 
-        kra_enabled = remote_api.Command.kra_is_enabled()['result']
+        setup_kra = remote_api.Command.kra_is_enabled()['result']
 
         if ca_enabled:
             options.realm_name = config.realm_name
             options.host_name = config.host_name
             ca.install_check(False, config, options)
 
-        if kra_enabled:
+        if setup_kra:
             try:
                 kra.install_check(remote_api, config, options)
             except RuntimeError as e:
@@ -887,7 +887,7 @@ def install_check(installer):
             os.environ['KRB5CCNAME'] = ccache
 
     installer._ca_enabled = ca_enabled
-    installer._kra_enabled = kra_enabled
+    installer._setup_kra = setup_kra
     installer._ca_file = cafile
     installer._fstore = fstore
     installer._sstore = sstore
@@ -1249,20 +1249,13 @@ def promote_check(installer):
                 'KRA', conn, config.kra_host_name)
         if kra_host is not None:
             config.kra_host_name = kra_host
-            kra_enabled = True
-        else:
-            if options.setup_kra:
-                logger.error("There is no KRA server in the domain, "
-                             "can't setup a KRA clone")
-                raise ScriptError(rval=3)
-            kra_enabled = False
 
         if ca_enabled:
             options.realm_name = config.realm_name
             options.host_name = config.host_name
             ca.install_check(False, config, options)
 
-        if kra_enabled:
+        if options.setup_kra:
             try:
                 kra.install_check(remote_api, config, options)
             except RuntimeError as e:
@@ -1320,7 +1313,7 @@ def promote_check(installer):
                 os.environ['KRB5CCNAME'] = ccache
 
     installer._ca_enabled = ca_enabled
-    installer._kra_enabled = kra_enabled
+    installer._setup_kra = options.setup_kra
     installer._ca_file = cafile
     installer._fstore = fstore
     installer._sstore = sstore
@@ -1338,7 +1331,7 @@ def promote_check(installer):
 def install(installer):
     options = installer
     ca_enabled = installer._ca_enabled
-    kra_enabled = installer._kra_enabled
+    setup_kra = installer._setup_kra
     fstore = installer._fstore
     sstore = installer._sstore
     config = installer._config
@@ -1364,7 +1357,7 @@ def install(installer):
                 if conn.isconnected():
                     conn.disconnect()
                 os.environ['KRB5CCNAME'] = ccache
-        config.dirman_password = ipautil.ipa_generate_password()
+        config.dirman_password = get_dirman_password()
 
         # FIXME: allow to use passed in certs instead
         if ca_enabled:
@@ -1463,7 +1456,7 @@ def install(installer):
     service.print_msg("Applying LDAP updates")
     ds.apply_updates()
 
-    if kra_enabled:
+    if setup_kra:
         kra.install(api, config, options)
 
     service.print_msg("Restarting the KDC")
--- a/ipatests/pytest_plugins/integration/tasks.py
+++ b/ipatests/pytest_plugins/integration/tasks.py
@@ -388,6 +388,9 @@ def install_replica(master, replica, setup_ca=True, setup_dns=False,
         fix_apache_semaphores(replica)
         args.extend(['-r', replica.domain.realm])
 
+    if not stdin_text:
+        stdin_text = replica.config.dirman_password
+
     result = replica.run_command(args, raiseonerr=raiseonerr,
                                  stdin_text=stdin_text)
     if result.returncode == 0:


stlaz:
Just please ask anyone on our IRC, Brno developers should be most helpful. This includes (IRC nicks):

- jcholast
- martbab
- mbasti
- dkupka
- slaznick (me)
- tkrizek

For community people, you can find us on irc.freenode.net, channel #freeipa where there's also other developers to help.

As for the PR and why it's bad:
It's not quite OK to:

- install anything before checking it could be installed
- install KRA before the core IPA services are installed and running
- install KRA before CA

felipevolpone:
@stlaz I did the changes, however I'm still getting the timeout exception. On IRC you suggested to use `private_ccache`, but see that this is already done here: https://github.com/freeipa/freeipa/blob/master/ipaserver/install/kra.py#L88-L100

stlaz:
Hm, sorry, I did not realize what the actual problem is. The fact is - the keys will never be on the master since it's not a KRA master.

MartinBasti:
`kra.install` nor `ca.install` cannot be called from `promote_check` or `install_check` functions. Only `kra.install_check` and `ca.install_check` can be. Please don't mix checking if installation is possible with installation itself

felipevolpone:
Thanks for pointing that @MartinBasti. 

Now the installation goes on until https://github.com/freeipa/freeipa/blob/master/ipaserver/install/server/replicainstall.py#L1471 and then I get the error:
```bash
ipa.ipapython.install.cli.install_tool(CompatServerReplicaInstall): ERROR    404 Client Error: Not Found for url: https://vm-058-064.abc.idm.lab.eng.brq.redhat.com/ipa/keys/ca/auditSigningCert%20cert-pki-kra?type=kem&value=eyJhbGciOiJSU0EtT0FFUCIsImVuYyI6IkEyNTZDQkMtSFM1MTIiLCJraWQiOm51bGx9.ZOnewjJo0xVvaCVfTR1An3VKuCW_9t0LupEzW_BFXzQhrXGMUwfePZ5mf3zwugk02vbg8KuYr6YAQKSaMuq6Xw-AbSNP-sTDNSBnlMcstuW3YoJoiI62hgbnZ0EVWHZCssDyti79-SROsNmhG5NACujriIFCnUDZRyNca_agH1lKkZBYYBlbJ9oRHde7UV_ZEp4Xk7NN2yZsCDENE42Ax7hsZXIvJGMT9CEk15y6BCxv4GO_mfaf_0IfiLW1xJhYFvTMze334CYgoYHlmKPVqddCazmgKN5Y_O5EDUIUhtc1NsOjITUmXfTC2vlVux7QRXPQjdISyuW2ICI2BQ-hnQ.QCjTO8KRODTuQrum32HZ-A.B5_aB-IFMZ3ing60XoRUbWpkoxGpeEyqdOeZSgNYdcVGFJEvnF6vkQxOJqCETCWkf_vvGS8Mff-JYwhidfc38j7ST9e9zHMapD9V8TmZxgLfnU6PsPlJvH_sSVK-rtjQsF5jTuk1vC_-gY6UxIxYPIi4gwaTLZLUbWaLo77vntqC4mRsg6u3Pjn8Jx-6XimYPQejabKB-OVYeojebklhtssyQHLcwj7RHyyVPeryCm9cxh_NKLYIIXlGcEZTlpNCfd3t7oFvs1bIu_JDYuO3L_1aQoUfmnSk6AjM3eLj-8IPSnod6fXfhJ1HHSMx3OiEgQyFGLRzS-oUQLS-UDhEP8RuxWrpPvsjCuwmGly9S4hr5AbkrTBcrsaqllB4AEl-uNYP8ntfdCkLNZdjr4m9EjO17FBk-a3ahfQ3PCLUJb3364ocRyvlcjQwbE4_8sSpAdNMmTKa5PyY6mIKaJa0RTN0bl8IS-brzruDhw8y_MJMbEHRVwf34Km8opVuso18GZVgHHFDZVNvSAOKjCdepnDHJsMplaMnyG8T58lzG4QXbn3goxaKVNPpDGXmJQqV7pes9ETfA2t-70phQraoFC8wzvmjp3_bbKkJ5C0vR36318zAxLqOIhnZXnaz2cyVnM5i1ACbFrf02SDK69CmBhDShmXpkhwC3AiwJBTGj_HPlmQNuombe40XMt3CITpDQT2tM6BzDJC7g514_9BhWGm2XIBr3wYtTvzfmqZPZhGCI7mayCZ4ZXCblYT3zK6p.bODuczfTQ_Hi2ITboe0k6jeqDGO0pGB_MaHt8jtNCjY
ipa.ipapython.install.cli.install_tool(CompatServerReplicaInstall): ERROR    The ipa-replica-install command failed. See /var/log/ipareplica-install.log for more information
```

ipareplica-install.log: http://pastebin.test.redhat.com/498192

ipaserver/install/kra.py
```diff
@@ -45,15 +45,15 @@ def install_check(api, replica_config, options):
             raise RuntimeError(
                 "Dogtag CA is not installed.  Please install the CA first")
 
+    if options.promote:
```
stlaz: You should still rather keep this check in the `if replica_config is not None` branch

stlaz:
The problem is, as I realized a bit later yesterday, that you're passing `config` to `ca/kra.install` methods even if these are the first instances of such service in the topology. This makes replica install think you're installing clone of such service rather than a first instance, then the result is you end up in the branch which tries to get the keys from a master but the keys have never been there.

ipaserver/install/server/replicainstall.py
```diff
@@ -1252,11 +1252,9 @@ def promote_check(installer):
             config.kra_host_name = kra_host
             kra_enabled = True
         else:
-            if options.setup_kra:
-                root_logger.error("There is no KRA server in the domain, "
-                                  "can't setup a KRA clone")
-                raise ScriptError(rval=3)
             kra_enabled = False
```
simo5: This exception is right if you are installing a replica, why did you remove it ?

simo5:
Ok I see what's the issue here, let me try to explain how the CA/KRA (and in general all components) works.

When you run any install you have two phases: 1) install_check, 2) install

all install_check phases are run first for all components and they check that the preconditions to actually successfully install a component are met, then when all checks are done components are installed in the right order.

In the CA and KRA cases we never assumed that it was legal to pass --setup-ca/kra in a replica on a domain where no CA/KRA exists. So we actually check for this precondition in the install_check phase of the two subsystems.

Ie if you pass in a "replica_config" in the intal_check phase of the KRA (see ./ipaserver/install/kra.py:install_check() ) we will bail out and fail install.

So what you need is two things:
1. you need to make it legal (change install_check) to install a KRA service during a replica install
 To do this it is probably sufficient to "return" instead of raising in the part where the "replica_config is not None" branch checks if the kra_is_enabled() check is performed.
2. You have to alter the install because in there too is replica_config is not None, the install now *assumes* you are going to replicating an existing KRA.

Instead now you will have to change the check to key it on whether an existing KRA is available on the domain or not when replica_config is passed in.

So something like this in pseudo-code:
if KRA_is_installed is True and replica_config is not None:
   kra_install_is_replica = True
else if KRA_is_installed is false:
   kra_install_is_replica = False

and then execute the rest of the code conditional to "kra_install_is_replica" instead of "replica_config is not None".

You will probably have to do something similar for CA or check if there are similar issues in that path at any rate.
Note that you still need a check for:
if replica_config is not None and not replica_config.setup_kra:
   return
run this at the top of install.

HTH

felipevolpone:
@simo5 thank you very much for such a detailed explanation. I've changed the code and now I think it reflects how it should be. Tell me if I'm wrong.

However, I'm still getting an error in the first step of the installation:
`self.step("configuring KRA instance", self.__spawn_instance)` in [this line.](https://github.com/freeipa/freeipa/blob/master/ipaserver/install/krainstance.py#L284-L289)

The error:
```
ipa.ipaserver.install.krainstance.KRAInstance: CRITICAL See the installation logs and the following files/directories for more information:
ipa.ipaserver.install.krainstance.KRAInstance: CRITICAL   /var/log/pki/pki-tomcat
RuntimeError: RuntimeE...ailed.',)
> /usr/lib/python2.7/site-packages/ipaserver/install/krainstance.py(289)__spawn_instance()
-> tmp_agent_pwd)
(Pdb) c
  [error] RuntimeError: KRA configuration failed.
ipa.ipapython.install.cli.install_tool(CompatServerReplicaInstall): ERROR    KRA configuration failed.
ipa.ipapython.install.cli.install_tool(CompatServerReplicaInstall): ERROR    The ipa-replica-install command failed. See /var/log/ipareplica-install.log for more information
The program exited via sys.exit(). Exit status:  1
```
/var/log/ipareplica-install.log
http://pastebin.test.redhat.com/498684

/var/log/pki/pki-tomcat/localhost_access_log.2017-06-28.txt
```
2620:52:0:224e:21a:4aff:fe23:1660 - - [28/Jun/2017:20:55:40 +0200] "GET /ca/admin/ca/getStatus HTTP/1.1" 200 168
2620:52:0:224e:21a:4aff:fe23:1660 - - [28/Jun/2017:21:02:19 +0200] "POST /ca/admin/ca/getStatus HTTP/1.1" 200 168
2620:52:0:224e:21a:4aff:fe23:1660 - ipara [28/Jun/2017:21:02:28 +0200] "GET /ca/rest/account/login HTTP/1.1" 200 218
2620:52:0:224e:21a:4aff:fe23:1660 - ipara [28/Jun/2017:21:02:28 +0200] "GET /ca/rest/authorities/b1c1a3f0-5aaa-4193-aadf-98937b4692eb/cert HTTP/1.1" 200 992
2620:52:0:224e:21a:4aff:fe23:1660 - ipara [28/Jun/2017:21:02:28 +0200] "GET /ca/rest/account/logout HTTP/1.1" 204 -
2620:52:0:224e:21a:4aff:fe23:1660 - - [28/Jun/2017:21:02:31 +0200] "POST /ca/rest/certrequests?issuer-id=b1c1a3f0-5aaa-4193-aadf-98937b4692eb HTTP/1.1" 200 345
2620:52:0:224e:21a:4aff:fe23:1660 - - [28/Jun/2017:21:02:31 +0200] "POST /ca/agent/ca/displayBySerial HTTP/1.1" 200 12605
2620:52:0:224e:21a:4aff:fe23:1660 - - [28/Jun/2017:21:18:19 +0200] "GET /ca/rest/securityDomain/domainInfo HTTP/1.1" 200 694
2620:52:0:224e:21a:4aff:fe23:1660 - - [28/Jun/2017:21:18:20 +0200] "GET /ca/rest/account/login HTTP/1.1" 401 994
```

ipaserver/install/kra.py
```diff
@@ -69,9 +68,19 @@ def install_check(api, replica_config, options):
...
 
 def install(api, replica_config, options):
-    if replica_config is None:
-        if not options.setup_kra:
-            return
+
+    if replica_config is not None and not replica_config.setup_kra:
+        return
+
+    install_kra_replica = False
```
simo5: rename this to install_kra_as_replica

ipaserver/install/kra.py
```diff
@@ -69,9 +68,19 @@ def install_check(api, replica_config, options):
...
+    kra = krainstance.KRAInstance(api.env.realm)
+
+    if kra.is_installed() and replica_config is not None:
+        install_kra_replica = False
+    elif not kra.is_installed():
+        install_kra_replica = True
+
```
simo5: So actually thinking about this check more I am not sure why we need to check if replica_config is None.
Either a kra is installed, in which case we need to make a clone (replica) or the kra is not installed in which case we need to create one from scratch.
however this check is wrong, because "is_installed" checks if a KRA is installed locally, not in the domain.

So I think this check is completely useless. Remove it.

ipaserver/install/kra.py
```diff
@@ -69,9 +68,19 @@ def install_check(api, replica_config, options):
...
+    if install_kra_replica:
```
simo5: Here replace this with:
if not api.Command.kra_is_enabled()['result']:

and later on after the else: remove the check to return if setup_kra is not set as we already checked for it atthe start of the function.

felipevolpone: I should change because this `api.Command.kra_is_enabled()['result']` checks if KRA is installed on master? The `kra.is_installed()` checks locally, right?

ipaserver/install/kra.py
felipevolpone: ack

ipaserver/install/kra.py
felipevolpone: Isn't it relevant to keep in replica_config check in case of domain level 0?

ipaserver/install/kra.py
felipevolpone: Done

ipaserver/install/kra.py
simo5: I think you can completely remove the install_kra_as_replica "middleman" variable and directly check only the api.Command below here.

simo5: I am not sure what difference checking replica_config would make at the moment.
It was used previously because we did not allow to install a new KRA from scratch on a replica, but now we are so it is not really relevant to check I think.

ipaserver/install/kra.py
```diff
@@ -47,9 +48,7 @@ def install_check(api, replica_config, options):
 
     if replica_config is not None:
         if not api.Command.kra_is_enabled()['result']:
-            raise RuntimeError(
-                "KRA is not installed on the master system. Please use "
-                "'ipa-kra-install' command to install the first instance.")
+            return
```
simo5: We may want to log a (info level) comment here that no KRA was found so we are going to install a new one from scratch.

ipaserver/install/kra.py
```diff
@@ -69,9 +68,16 @@ def install_check(api, replica_config, options):
 
 
 def install(api, replica_config, options):
-    if replica_config is None:
-        if not options.setup_kra:
-            return
+
+    if replica_config is not None and not replica_config.setup_kra:
+        return
+
```
simo5: About the above, although in the server install code we calll kra.install only if options.setup_kra is true, it feels like we may just want to check:
if not options.setup_kra:
    return

This will catch either replica or server install cases should we change how installs are called in future.
In fact we might even go to the server install code and skip the check there and let the install return.

Note that to do this we'll need to change replicainstall.py to set:
installer.setup_kra = options.setup_kra
instead of:
installer._kra_enabled = kra_enabled

ipaserver/install/server/replicainstall.py
```diff
@@ -1252,11 +1253,9 @@ def promote_check(installer):
...
+            if options.setup_kra:
+                kra_enabled = True
```
simo5: wee comment about options.setup_kra for the other file, we can completely remove the kra_enabled variable everywhere and these 2 lines and simply set installer.setup_kra = options.setup_kra at the end of the check phase (note you have to do this for both classic install and promotion, and replace checks of kra_enabled with options.setup_kra throughout this file)

felipevolpone:
@simo5 I've done all the changes you proposed (and I agree with them), however the error that I pointed in comment above is still happening. Any idea what is it?

simo5:
ERROR:  Unable to access security domain: 401 Client Error: Unauthorized for url: https://vm-058-064.abc.idm.lab.eng.brq.redhat.com:443/ca/rest/account/login

This seems the relevant issue, it seem we are missing a certificate needed for accessing the URL ?
We need the other logs perhaps to find out what cert if any was used, but it seems like a missing step, you should probably compare the code used to setup CA/KRA in server install vs replica install and see if any critical step is missing in the replicainstall side. We should actually move any such step in the CA/KRA install code if we are missing it.

simo5:
Can you also see if we have clues in http error log ?

ipaserver/install/server/replicainstall.py
```diff
@@ -1321,7 +1315,7 @@ def promote_check(installer):
                 os.environ['KRB5CCNAME'] = ccache
 
     installer._ca_enabled = ca_enabled
-    installer._kra_enabled = kra_enabled
+    installer._kra_enabled = options.setup_kra
```
simo5: This should be: installer.setup_kra as well and any use of _kra_enabled fixed through the code.
Keep in mind that the variable "installer" here ends up being "options" in the install() function, which is very confusing I know, we should fix that too eventually.

felipevolpone:
> This seems the relevant issue, it seem we are missing a certificate needed for accessing the URL ?

How could I check this? There is some documentation about which certificates are used for each URL/service? 

> We need the other logs perhaps to find out what cert if any was used, but it seems like a missing step, you should probably compare the code used to setup CA/KRA in server install vs replica install and see if any critical step is missing in the replicainstall side. We should actually move any such step in the CA/KRA install code if we are missing it.

In general, I've not noticed any major difference between the server installation code and what we're doing in the replica. I've tried some different things, like adding some one these calls:

```
krb.restart()
tasks.restore_context(paths.CACHE_IPA_SESSIONS)
```
but it did not changed the final result.

I tried to add this: `ca.set_subject_base_in_config(options.subject_base)`, but it raised an exception as well.

The logs:
/var/log/httpd/access_log
https://pastebin.com/bJz8PAmS

/var/log/httpd/error_log
https://pastebin.com/XKH6ZX8q

/var/log/ipareplica-install.log
https://pastebin.com/DAJTxtq3

/var/log/pki/pki-kra-spawn.20170630153654.log
https://pastebin.com/48hT3kEk

I also have checked the logs from a server installation (using KRA) against the logs above, but I didn't not found something significant (I'm not sure where I should look for something specific).

ipaserver/install/server/replicainstall.py
```diff
@@ -1250,20 +1250,13 @@ def promote_check(installer):
...
-                                  "can't setup a KRA clone")
-                raise ScriptError(rval=3)
-            kra_enabled = False
 
         if ca_enabled:
             options.realm_name = config.realm_name
             options.host_name = config.host_name
             ca.install_check(False, config, options)
 
-        if kra_enabled:
```
stlaz: Last Fall we were trying to refactor installators so that same stuff is handled the same way. You break this for KRA and CA installation, although these are basically the same. Please, don't.

ipaserver/install/kra.py
```diff
@@ -47,9 +49,9 @@ def install_check(api, replica_config, options):
 
     if replica_config is not None:
         if not api.Command.kra_is_enabled()['result']:
-            raise RuntimeError(
-                "KRA is not installed on the master system. Please use "
-                "'ipa-kra-install' command to install the first instance.")
+            root_logger.info("KRA is not installed on the master system. "
```
stlaz: Please remove this logging. Note that `root_logger` should generally not be used (although it is unfortunately used very much throughout IPA a lot).
I also believe that a user does not care about this message.

stlaz:
Does this work in both domain levels?

ipaserver/install/server/replicainstall.py
simo5: I asked for this, I would agree to change the ca part to be like the new fra, not the other way around.
It's kinda useless and confusing to use additional variables when the one that exist (nca can be easily grepped through the code, work just fine.

ipaserver/install/kra.py
simo5: I also asked for this, it allows debugging to figure out what happened.
Perhaps we should log ad the debug level, not info, but I definitely would like to have an early indication of what the installer thinks is going on.

simo5:
@felipevolpone not sure wht touching the krb service of the HTTP sessions would make any difference ?
What is failing here is that the KRA is trying to use a certificate that is tied to an admin accunt in order to bootstrap itself (afaict). Somehow this cert is not being install, I suspect that may be because the replica install code assume that to be already available in the domain. You need to look at the difference between the code paths used to install CA/KRA on the first master and on a replica and see if there are operations done differently.
Because although you are installing on a replica from the POV of KRA ins this case you are acting like if you are installing the first master, so any precondition we fullfill in that case needs to be fullfilled in this case too.
Note that this may also be a dead end, in that a certificate may be asked as a fallback due to a different missing precondition and it is not wat we need. So you really need to compare KRA initial install to what gos on in a replica to really figure out where is the difference.

stlaz:
Ah, I forgot about the questions. Soooo...

@simo5: +1 about that krb service.
As for the certificate, we have a client certificate for Dogtag which is to be found at `/var/lib/ipa/ra-agent.crt` which is created during CA setup. This certificate is used for interacting with Dogtag, if you're missing it at the point of installing KRA, you've done something very wrong.

ipaserver/install/kra.py
stlaz: Hm, there's a point to that. I propose moving this message to `KRAInstance.configure_instance` but as you say, it should be a debug message instead.

ipaserver/install/server/replicainstall.py
stlaz: Yes, to deepen my point - both should be handled the same way so I don't care one is being changed but the other has to be, too.

felipevolpone:
My theory is that `pki-spawn` failing due to a lack of the /root/ca-agent.p12 file on the system. 

```
$ pkispawn -f /tmp/tmpLiHsXG -s KRA
Log file: /var/log/pki/pki-kra-spawn.20170703201453.log
Loading deployment configuration from /tmp/tmpLiHsXG.
ERROR:  Unable to access security domain: 401 Client Error: 
Unauthorized for url: https://vm-058-064.abc.idm.lab.eng.brq.redhat.com:443/ca/rest/account/login
```

The cfg file: https://pastebin.com/S3YCQ0Et
The log: https://pastebin.com/dNTTyey4

Maybe @frasertweedale knows something about it this.

simo5:
As far as I can tell /root/ca-agent.p12 aka paths.DOGTAG_ADMIN_P12 is generated (at least that's the only place where it is referenced) by the CAInstance in  __spawn_instance(), but not in the case where ra_only is True.
ra_only is True when replica_config.setup_ca is True (see install_step_0).

So in the case you are installing the KRA for the first time but not the CA (your case) I think you will have to find a way to generate a valid ca-agent.p12
You cannot transfer it because in dom lvl0 we do not copy it in the replica file, and in dom lvl 1 custodia does not have a way to retrieve it (unless it is also copied in one of the nss databases like the RA cert).

simo5:
I think you should run ipa-kra-install on a replica server ina  domain where no kra has been installed and see if it fails there. If not you need to find out what is the difference between what your patch does and a plain ipa-kra-install does on a replica, because they should behave the same way.

simo5:
Note that one issue may be the DM password. During the replica install process at lvl1 we do not use the "domain" DM password, but we generate one for the local install and at the end of the process we replace it in DS via Custodia, by sucking out the hashed password out of the master we are installing from. So if that temporary DM password is used to stand up the KRA (and I remember that in ipa-kra-install we explicitly ask for the DM password) then that may be the issue. In that case a --setup-kra may have to prompt for the actual DM password (undesirable) or maybe @frasertweedale or @vakwetu know of a way to install KRA w/o using a DM password, or by creating a temporary admin user in the domain.

felipevolpone:
> So in the case you are installing the KRA for the first time but not the CA (your case) I think you will have to find a way to generate a valid ca-agent.p12

If I understood correctly, our case ir setup ca + setup kra, right? 

> I think you should run ipa-kra-install on a replica server ina domain where no kra has been installed and see if it fails there

It's working properly.

> Note that one issue may be the DM password. During the replica install process at lvl1 we do not use the "domain" DM password, but we generate one for the local install and at the end of the process we replace it in DS via Custodia, by sucking out the hashed password out of the master we are installing from [...]

That is exactly the problem. I've changed [this line](https://github.com/freeipa/freeipa/blob/master/ipaserver/install/server/replicainstall.py#L1371) to the line below (**only for testing purposes**) and the installation works.
```python
# config.dirman_password = ipautil.ipa_generate_password()
config.dirman_password = u'12345678'  # this is password I've setup during the installation
```
However, I'm not sure how to continue from here. Should we ask for the password or there is a way to use the dm password without asking for it?

simo5:
Please hunt down @frasertweedale and/or @vakwetu, they will know if it is posible to install KRA w/o requiring the DM password.

felipevolpone:
@frasertweedale or @vakwetu, could you help us?

frasertweedale:
@felipevolpone I'll investigate on Wednesday.

frasertweedale:
@felipevolpone sorry I did not have time for detailed investigation.  I will continue tomorrow.  Apologies if I am blocking you!

felipevolpone:
@frasertweedale it's ok, no problem. thanks for status

felipevolpone:
Hi @frasertweedale  could you check this PR, please? It's blocking the upstream tests.

felipevolpone:
Last commit fixes the issue, however I'm not sure if it's the right approach to ask for the DM password.

simo5:
Unless dogtag people give you a way to make KRA work w/o owning the DM password I do not think we have another way for now.
It would be nice if CA/KRA could be configured to use ldapi or at least a cert to connect to LDAP instead of insisting on using the DM password, or maybe use a separate account... perhaps we should open a ticket in dogtag for that.

frasertweedale:
Let's ask for DM password for now (it is tantamount to `ipa-replica-install` followed by `ipa-kra-install`, after all, and this currently requires DM password).  LDAPI is not a bad idea and we already have an upstream ticket for that: https://pagure.io/dogtagpki/issue/1585

stlaz:
We'll close it for now, please, add this requirement to the man pages.

PR #895, opened 2017-06-26
Ensure network is online prior to an upgrade

stlaz:
Fedora system upgrades don't necessary require the network to be up.
However, we do, so we have to make sure the network is online before
performing an upgrade.

https://pagure.io/freeipa/issue/7039

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1017,6 +1017,10 @@ fi
 python2 -c "import sys; from ipaserver.install import installutils; sys.exit(0 if installutils.is_ipa_configured() else 1);" > /dev/null 2>&1
 
 if [  $? -eq 0 ]; then
+    # This is necessary for Fedora system upgrades which by default
+    # work with the network being offline
+    /bin/systemctl start network-online.target
+
     # This must be run in posttrans so that updates from previous
     # execution that may no longer be shipped are not applied.
     /usr/sbin/ipa-server-upgrade --quiet >/dev/null || :


tomaskrizek:
This change depends on https://github.com/systemd/systemd/pull/6065

stlaz:
We can push it nonetheless

tomaskrizek:
I disagree. If the systemd change doesn't go through, the problem isn't fixed.

stlaz:
That's not our problem. We need the network online so we make our best effort to make it online. We don't care what the guys in systemd do.
Once users start reporting this, we can easily say "Wow, we did not expect systemd to not provide us with an online network when we explicitly issue `systemctl start network-online.target`. Please, go and report this bug to them." That should effectively make the systemd guys prioritize fixing the issue on their side.

stlaz:
Systemd PR was merged.

tomaskrizek:
This should work fine now. I don't think it's necessary to bump the dependency in spec. It'll just start to work once the systemd is updated.

tomaskrizek:
If you think a release note is necessary, please update the ticket.

tomaskrizek:
master:

* 7ef08aa8cb6e504bc38850a6bf49cb2766b3c1df Ensure network is online prior to an upgrade


ipa-4-5:

* 3fd92a9c677f0486823c5218d087fa3703db5eab Ensure network is online prior to an upgrade

tomaskrizek:
I accidentally pushed this PR to a mirror repo instead of upstream. Sorry about that. I'm re-opening the PR and pushing it to the proper repo.

tomaskrizek:
master:

* a36f2aed637dad5bac6f2063c05a5ed64eb922ce Ensure network is online prior to an upgrade


ipa-4-5:

* 6ca6942fda4cea7a78569eb83031e7e6032ace47 Ensure network is online prior to an upgrade

PR #896, opened 2017-06-27
[py3] wsgi fixes

stlaz:
These patches fix WebUI for FreeIPA when using python3-mod_wsgi. To get it working, however, one also needs to run Fedora 26 with python3-pyldap with the following fixes:
https://github.com/pyldap/pyldap/pull/104
https://github.com/pyldap/pyldap/commit/60801268c56784f86847caa400dc0aad1699eb26

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,7 @@ env:
           TEST_RUNNER_CONFIG=".test_runner_config_py3_temp.yaml"
           TESTS_TO_RUN="test_cmdline
                 test_ipalib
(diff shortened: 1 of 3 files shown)


MartinBasti:
master:

* db4d0998fdd37422ed9dcb0d88ea5adf6ac98412 wsgi plugins: mod_wsgi expects bytes as an output
* d308abac2e6758014ebb39bcbd0fc3bd61a57db2 rpcserver: remove addition of str and bytes

PR #898, opened 2017-06-27
py3: ipa-dnskeysyncd: fix bytes issues

MartinBasti:
LDAP client returns values as bytes, thus ipa-dnskeysyncd must work with
bytes properly.

### DIFF
--- a/ipaserver/dnssec/bindmgr.py
+++ b/ipaserver/dnssec/bindmgr.py
@@ -11,6 +11,8 @@ import os
 import shutil
 import stat
 
+import six
+
 import ipalib.constants
 from ipapython.dn import DN
 from ipapython import ipautil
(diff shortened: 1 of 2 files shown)


flo-renaud:
Hi @MartinBasti,

Thank you for the patch. There are a few places left that need to be converted to bytes:
- in _get_signing_attr the default value should be [b'FALSE'] instead of ['FALSE']
- in __is_dnssec_enabled the comparison should use b'TRUE" instead of 'TRUE'
Otherwise it looks good to me.

MartinBasti:
Thank you @flo-renaud , I updated the PR

flo-renaud:
Thank you @MartinBasti for the PR update. I also noticed an issue in bindmgr.py in get_zone_dir_name():
```
        for label in zone:
            for char in label:
                c = ord(char)
```
In python3, char will be an int while in python 2 it is a string. Because of that, ord() will fail in python3.

stlaz:
@tomaskrizek to kindly finish this PR to get more familiar to DNSSEC

tomaskrizek:
Superseded by #1011

PR #899, opened 2017-06-28
test_caless: introduce new python makepki + fix SKI extension issue

Rezney:
Change makepki.sh for new makepki.py which should be more
readable, maintainable and extendable than the old script.
In this test we use it as a module and import create_pki().

The new makepki adds SKI and AKI extensions for correct
cert validation.

Other minor changes needed as we do not use NSS to store our
certificates on the test controller.

https://pagure.io/freeipa/issue/7030

Signed-off-by: Michal Reznik <mreznik@redhat.com>


Rezney:
Changes in test_caless.py are as follows:

- Got rid of old bash script execution
- Introduced new makepki
- Fixed some paths due to nssdb now missing
- Changed pkinit default name and removed copy_pkinit_cert() as we put pkinit certs to files_to_copy now
- Reworked export_pkcs12() to generate p12 file using whole CA chain
- Added prepare_pem() method to copy --ca-cert-file to pem_filename
- Changed get_pem() to cat to output instead of using certutil

WIP:

- prepare_pem() and get_pem() might be done better. At least we should choose either .crt or .pem and not combine.

Rezney:
WIP. Going to fix the lint issues...

ipatests/test_integration/create_caless_pki.py
```diff
@@ -0,0 +1,557 @@
...
+
+
+def create_pki(pki_vars):
+
+    # as original "makepki.py" from freeipa-tools uses mostly constant global
+    # variables in the code here we want to set what we get from ca_less test
+    # suite as global variables too
+
+    module = sys.modules[__name__]
+    for name, value in pki_vars.items():
```
stlaz: It might be better to just set these variables from the calling script.

ipatests/test_integration/test_caless.py
```diff
@@ -111,9 +116,6 @@ class CALessBase(IntegrationTest):
     def install(cls, mh):
         cls.cert_dir = tempfile.mkdtemp(prefix="ipatest-")
         cls.pem_filename = os.path.join(cls.cert_dir, 'root.pem')
-        scriptfile = os.path.join(os.path.dirname(__file__),
-                                  'scripts',
-                                  'caless-create-pki')
         cls.cert_password = cls.master.config.admin_password
 
         cls.crl_path = os.path.join(cls.master.config.test_dir, 'crl')
```
stlaz: There's no such directory being created

ipatests/test_integration/test_caless.py
```diff
@@ -126,26 +128,25 @@ def install(cls, mh):
...
+        create_caless_pki.create_pki()
 
         for host in cls.get_all_hosts():
             tasks.apply_common_fixes(host)
 
             # Copy CRLs over
-            base = os.path.join(cls.cert_dir, 'nssdb')
             host.transport.mkdir_recursive(cls.crl_path)
-            for source in glob.glob(os.path.join(base, '*.crl')):
+            for source in glob.glob(os.path.join(cls.crl_path, '*.crl')):
```
stlaz: If the module works the same as in freeipa-tools, then I believe the CRL files are to be found in `cls.cert_dir`

ipatests/test_integration/test_caless.py
```diff
@@ -306,21 +303,53 @@ def get_replica_filename(self, replica):
...
+        certchain_fname = os.path.join(cls.cert_dir, nickname + '.pem')
+
+        nick_chain = nickname.split(os.sep)
+
+        # to construct whole chain e.g "ca1 - ca/sub - ca1/sub/server"
+        for index, _value in enumerate(nick_chain):
+            cert_path = '/'.join(nick_chain[:index+1])
+            fname_chain.append(os.path.join(cls.cert_dir, cert_path + '.crt'))
+        # to simulate unknown CA we drop ca3 from the chain
+        if nickname.startswith('ca3/'):
```
stlaz: Please remove this and just append the files to `fname_chain` if they exist. `ca3.crt` is not generated as a part of `crreate_caless_pki` (or it was not originally, hopefully you did not change this).

ipatests/test_integration/test_caless.py
```diff
@@ -306,21 +303,53 @@ def get_replica_filename(self, replica):
...
-                     '-d', 'nssdb',
-                     '-K', cls.cert_password,
-                     '-W', password], cwd=cls.cert_dir)
+
+        fname_chain = []
+
+        key_fname = os.path.join(cls.cert_dir, nickname + '.key')
+        certchain_fname = os.path.join(cls.cert_dir, nickname + '.pem')
+
+        nick_chain = nickname.split(os.sep)
```
stlaz: This is actually wrong, the nicknames are (in our case) always separated by '/'. I lead you to this, but I was wrong, sorry.

ipatests/test_integration/test_caless.py
```diff
@@ -306,21 +303,53 @@ def get_replica_filename(self, replica):
...
+            cert_path = '/'.join(nick_chain[:index+1])
+            fname_chain.append(os.path.join(cls.cert_dir, cert_path + '.crt'))
+        # to simulate unknown CA we drop ca3 from the chain
+        if nickname.startswith('ca3/'):
+            fname_chain.pop(0)
+
+        with open(certchain_fname, 'w') as chain:
+                for cert_fname in fname_chain:
+                    with open(cert_fname) as cert:
+                        chain.write(cert.read())
```
stlaz: Instead of having a separate file you can just create the `fname_chain` as a list of arguments to `ipautil.run`. So instead of having the `-in` argument to `openssl`, you extend the argument list by a list which should look somehow like this:
```python
['-certfile', 'ca1.crt', '-certfile', 'ca1/server.crt']
```
and so on.
I am doing the very same in https://github.com/freeipa/freeipa-tools/pull/46/files except that's in bash. You can just rewrite it to python.

ipatests/test_integration/test_caless.py
```diff
@@ -306,21 +303,53 @@ def get_replica_filename(self, replica):
...
+                     "-name", nickname], cwd=cls.cert_dir,
+                    )
+
+    @classmethod
+    def prepare_cacert(cls, nickname):
+        """ Prepare pem file for root_ca_file/ca-cert-file option """
+        # create_caless_pki saves certificates with ".crt" extension by default
+        suffix = '.crt'
+        fname_from_nick = os.path.join(cls.cert_dir, nickname + suffix)
+        ipautil.run(['cp', fname_from_nick, cls.pem_filename],
```
stlaz: `shutil.copy`
Brrr, creepy.

ipatests/test_integration/test_caless.py
```diff
@@ -306,21 +303,53 @@ def get_replica_filename(self, replica):
...
+                    cwd=cls.cert_dir)
 
     @classmethod
-    def get_pem(cls, nickname):
+    def get_cacert(cls, nickname):
+        """ Return cert as base64 encoded ascii for TestIPACommands """
+        suffix = '.crt'
+        fname_from_nick = os.path.join(cls.cert_dir, nickname + suffix)
         result = ipautil.run(
```
stlaz:
```python
cacert_fname = '{}.crt'.format(os.path.join(cls.cert_dir, nickname))
with open(cacert_fname, 'r') as f:
    return f.read()
```
This is not bash :(

ipatests/test_integration/test_caless.py
```diff
@@ -306,21 +303,53 @@ def get_replica_filename(self, replica):
...
+    def prepare_cacert(cls, nickname):
+        """ Prepare pem file for root_ca_file/ca-cert-file option """
+        # create_caless_pki saves certificates with ".crt" extension by default
+        suffix = '.crt'
+        fname_from_nick = os.path.join(cls.cert_dir, nickname + suffix)
+        ipautil.run(['cp', fname_from_nick, cls.pem_filename],
+                    cwd=cls.cert_dir)
 
     @classmethod
-    def get_pem(cls, nickname):
```
stlaz: Why rename it?

ipatests/test_integration/create_caless_pki.py
```diff
@@ -0,0 +1,548 @@
...
+# purpose with or without fee is hereby granted, provided that the above
+# copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+# ANY SPECIAL, cert_dirECT, INcert_dirECT, OR CONSEQUENTIAL DAMAGES OR ANY
+# DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+# ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+# OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
```
stlaz: You really should change the header here.

ipatests/test_integration/test_caless.py
```diff
@@ -309,23 +306,45 @@ def get_replica_filename(self, replica):
...
+
+        fname_chain = []
+
+        key_fname = os.path.join(cls.cert_dir, nickname + '.key')
+        cert_fname = os.path.join(cls.cert_dir, nickname + '.crt')
+
+        nick_chain = nickname.split('/')
+
+        # to construct whole chain e.g "ca1 - ca1/sub - ca1/sub/server"
+        for index, _value in enumerate(nick_chain):
```
stlaz: Usually, you'd probably just do `for idx in range(0, len(nick_chain))` but I don't mind this staying here.

ipatests/test_integration/test_caless.py
```diff
@@ -309,23 +306,45 @@ def get_replica_filename(self, replica):
...
+        fname_from_nick = '{}.crt'.format(os.path.join(cls.cert_dir, nickname))
+        shutil.copy(fname_from_nick, cls.pem_filename)
 
     @classmethod
     def get_pem(cls, nickname):
-        result = ipautil.run(
-            [paths.CERTUTIL, '-L', '-d', 'nssdb', '-n', nickname, '-a'],
-            cwd=cls.cert_dir, capture_output=True)
-        return result.output
+        """ Return cert as base64 encoded ascii for TestIPACommands """
```
stlaz: Kudos for a docstring, except the returned string is a PEM certificate rather than just a base64 representation of something.

ipatests/test_integration/test_caless.py
Rezney: Unfortunately this does not work.

ipatests/test_integration/create_caless_pki.py
Rezney: ouch

ipatests/test_integration/test_caless.py
Rezney: enumerate is preferable I guess. Even our code quantifier says so :) Moreover it looks better

stlaz:
The tests are passing, should anything fail, we can fit that later. Thank you @Rezney for coping up with me in this PR!
ACK.

tomaskrizek:
master:

* 64375ba65bfc56694a425b1e39bd2081e604f777 test_caless: introduce new python makepki + fix SKI extension issue

PR #901, opened 2017-07-03
Passdb private state

abbra:
This is a set of patches to allow FreeIPA to be built against Samba 4.7 when it will be released. The code should work just fine against Samba 4.6 already.

### DIFF
--- a/daemons/ipa-sam/ipa_sam.c
+++ b/daemons/ipa-sam/ipa_sam.c
@@ -4532,8 +4532,12 @@ static NTSTATUS pdb_init_ipasam(struct pdb_methods **pdb_method,
 				      uri, false, NULL, NULL,
 				      &ipasam_state->ldap_state);
 		if (NT_STATUS_IS_OK(status)) {
+#ifdef HAVE_SMBLDAP_SET_BIND_CALLBACK
+			smbldap_set_bind_callback(ipasam_state->ldap_state, bind_callback, ipasam_state);
+#else
 			ipasam_state->ldap_state->bind_callback = bind_callback;
 			ipasam_state->ldap_state->bind_callback_data = ipasam_state;
+#endif
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


MartinBasti:
I tried samba packages from https://copr.fedorainfracloud.org/coprs/pviktori/samba-py3/package/samba/  and build failed:

```
ipa_sam.c: In function 'priv2ld':
ipa_sam.c:349:25: error: dereferencing pointer to incomplete type 'struct smbldap_state'
  return priv->ldap_state->ldap_struct;
                         ^~
ipa_sam.c:350:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
Makefile:601: recipe for target 'ipa_sam.lo' failed
```

abbra:
Thanks, I missed one place there. However, this patch is not going to work aginst that samba build because py3 copr misses the patch I sent today to samba-technical@. Hopefully, Andreas will get it reviewed before 4.7.0-rc1 comes out tomorrow.

abbra:
I've updated patches and tested them with Andreas `asn/samba_ad_dc` COPR repository. They now pass both compile and simple tests for me.

abbra:
I've tested on F26. We need these patches in Rawhide because an update to Samba 4.7 is already there.

abbra:
Note that rawhide build currently fails due to https://bugzilla.redhat.com/show_bug.cgi?id=1467526 (gcc bug on ppc64le). We still need to commit these changes upstream to make it possible to test with `asn/samba_ad_dc` COPR repo.

MartinBasti:
master:

* 11d43a16035d9ce7970ddf757f17025289ec4854 ipa-sam: use own private structure, not ldapsam_privates
* 3ab6a68e91a4fad5af037a910fd4f6b6d8f7611f ipa-sam: use smbldap_set_bind_callback for Samba 4.7 or later

PR #903, opened 2017-07-05
Warning the user when using a loopback IP as forwarder

felipevolpone:
Now, the user can pass a loopback IP in the --forwarder option.
Previously, an error would be raised, now we just show a warning message.

Fixes: https://pagure.io/freeipa/issue/5801

### DIFF
--- a/install/tools/ipa-dns-install
+++ b/install/tools/ipa-dns-install
@@ -54,7 +54,7 @@ def parse_options():
                       help="Master Server IP Address. This option can be used "
                            "multiple times")
     parser.add_option("--forwarder", dest="forwarders", action="append",
-                      type="ip", help="Add a DNS forwarder. This option can be used multiple times")
+                      type="ip_with_loopback", help="Add a DNS forwarder. This option can be used multiple times")
     parser.add_option("--no-forwarders", dest="no_forwarders", action="store_true",
                       default=False, help="Do not add any DNS forwarders, use root servers instead")
     parser.add_option("--auto-forwarders", dest="auto_forwarders",
--- a/ipapython/config.py
+++ b/ipapython/config.py
@@ -23,6 +23,7 @@ from optparse import (
 # pylint: enable=deprecated-module
 from copy import copy
 import socket
+import functools
 
 from dns import resolver, rdatatype
 from dns.exception import DNSException
@@ -33,6 +34,7 @@ from six.moves.urllib.parse import urlsplit
 # pylint: enable=import-error
 
 from ipapython.dn import DN
+from ipapython.ipautil import CheckedIPAddress, CheckedIPAddressLoopback
 
 try:
     # pylint: disable=ipa-forbidden-import
@@ -65,13 +67,16 @@ class IPAFormatter(IndentedHelpFormatter):
             ret += "%s %s\n" % (spacing, line)
         return ret
 
-def check_ip_option(option, opt, value):
-    from ipapython.ipautil import CheckedIPAddress
 
+def check_ip_option(option, opt, value, allow_loopback=False):
     try:
-        return CheckedIPAddress(value)
+        if allow_loopback:
+            return CheckedIPAddressLoopback(value)
+        else:
+            return CheckedIPAddress(value)
     except Exception as e:
-        raise OptionValueError("option %s: invalid IP address %s: %s" % (opt, value, e))
+        raise OptionValueError("option {}: invalid IP address {}: {}"
+                               .format(opt, value, e))
 
 def check_dn_option(option, opt, value):
     try:
@@ -95,9 +100,11 @@ class IPAOption(Option):
     security-sensitive such as passwords.
     """
     ATTRS = Option.ATTRS + ["sensitive", "constructor"]
-    TYPES = Option.TYPES + ("ip", "dn", "constructor")
+    TYPES = Option.TYPES + ("ip", "dn", "constructor", "ip_with_loopback")
     TYPE_CHECKER = copy(Option.TYPE_CHECKER)
     TYPE_CHECKER["ip"] = check_ip_option
+    TYPE_CHECKER["ip_with_loopback"] = functools.partial(check_ip_option,
+                                                         allow_loopback=True)
     TYPE_CHECKER["dn"] = check_dn_option
     TYPE_CHECKER["constructor"] = check_constructor
 
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -245,6 +245,25 @@ class CheckedIPAddress(UnsafeIPAddress):
         self._net = ifnet
 
 
+class CheckedIPAddressLoopback(CheckedIPAddress):
+    """IPv4 or IPv6 address with additional constraints with
+    possibility to use a loopback IP.
+    Reserved or link-local addresses are never accepted.
+    """
+    def __init__(self, addr, parse_netmask=True, allow_multicast=False):
+
+        super(CheckedIPAddressLoopback, self).__init__(
+                addr, parse_netmask=parse_netmask,
+                allow_multicast=allow_multicast,
+                allow_loopback=True)
+
+        if self.is_loopback():
+            # print is being used instead of a logger, because at this
+            # moment, in execution process, there is no logger configured
+            print("WARNING: You are using a loopback IP: {}".format(addr),
+                  file=sys.stderr)
+
+
 def valid_ip(addr):
     return netaddr.valid_ipv4(addr) or netaddr.valid_ipv6(addr)
 
--- a/ipaserver/install/dns.py
+++ b/ipaserver/install/dns.py
@@ -490,7 +490,7 @@ class DNSInstallInterface(hostname.HostNameInstallInterface):
 
     forwarders = knob(
         # pylint: disable=invalid-sequence-index
-        typing.List[ipautil.CheckedIPAddress], None,
+        typing.List[ipautil.CheckedIPAddressLoopback], None,
         description=("Add a DNS forwarder. This option can be used multiple "
                      "times"),
         cli_names='--forwarder',
--- a/ipapython/install/cli.py
+++ b/ipapython/install/cli.py
@@ -16,7 +16,8 @@ import six
 
 from ipapython import admintool
 from ipapython.ipa_log_manager import standard_logging_setup
-from ipapython.ipautil import CheckedIPAddress, private_ccache
+from ipapython.ipautil import (CheckedIPAddress, CheckedIPAddressLoopback,
+                               private_ccache)
 
 from . import core, common
 
@@ -166,6 +167,8 @@ class ConfigureTool(admintool.AdminTool):
                 kwargs['type'] = 'int'
             elif knob_scalar_type is long:
                 kwargs['type'] = 'long'
+            elif knob_scalar_type is CheckedIPAddressLoopback:
+                kwargs['type'] = 'ip_with_loopback'
             elif knob_scalar_type is CheckedIPAddress:
                 kwargs['type'] = 'ip'
             elif issubclass(knob_scalar_type, enum.Enum):


felipevolpone:
Even using `root_logger.warning`, the message did not show up on the console neither in /var/log/ipaserver-install.log. Should I use `print`?

abbra:
No, don't use print in this helper function. Given that all we want is to allow `--forwarder` with a loopback IP, I think it is fine just to have the message in the `/var/log/ipaserver-install.log` alone.

As to why it is not appearing in the logs, this needs to be investigated.

MartinBasti:
Our loggers for installers are set up to log only error+ messages on console, however it should log everything in log file.

MartinBasti:
Instead of 
```python
 if opt == '--forwarder':
```

I prefer to create a new method:

`check_ip_with_loopback_option`

In a following way
```python
def _check_ip_option(option, opt, value, allow_loopback=False):
#
# Code from original check_ip_option updated with allow_loopback kw introduced here
#

def check_ip_option(option, opt, value):
    _check_ip_option(option, opt, value)

def check_ip_with_loopback_option(option, opt, value):
    _check_ip_option(option, opt, value, allow_loopback=True)
```

and then update
```diff
TYPE_CHECKER["ip"] = check_ip_option
+ TYPE_CHECKER["ip_with_loopback"] = check_ip_with_loopback_option
```

and replace in installer knob `ip` with `ip_with_loopback` for forwarders

Also please note that ipa-dns-install still uses own parses outside knobs so please update it there as well

MartinBasti:
IMHO my solution is longer but easier for future maintenance as it is clear from option specification what is expected instead of *hardcoded* option somewhere in validators

stlaz:
You don't need to create new functions to wrap around the old one, as this is a special case, the cleaner solution is to use `functools.partial` and just add an argument to the old function for `TYPE_CHECKER["ip_with_loopback"]`.
Also, it's not cool to warn a user that they are doing something when they are really not.

felipevolpone:
I've done the changes, however either standa's proposals did not work:

* `typing.List[functools.partial(ipautil.CheckedIPAddress, allow_loopback=True)],` raise this error:
```
TypeError: Parameters to generic types must be types. Got <functools.partial object at 0x7f53065b0cb0>
```

* `TypeError: issubclass() arg 1 must be a class` raise this error
```
functools.partial(typing.List[ipautil.CheckedIPAddress], allow_loopback=True),
```

stlaz:
No, you're not missing something, I thought this might happen but am not that experienced with the `typing` library so I was hoping it wouldn't. You'll probably need to find another way around this.

ipapython/config.py
```diff
@@ -65,14 +65,20 @@ def format_usage(self, usage):
             ret += "%s %s\n" % (spacing, line)
         return ret
 
-def check_ip_option(option, opt, value):
+def __check_ip_option(option, opt, value, allow_loopback=False):
```
stlaz: Let me quote myself here:
>You don't need to create new functions to wrap around the old one, as this is a special case, the cleaner solution is to use functools.partial and just add an argument to the old function for TYPE_CHECKER["ip_with_loopback"].

ipaserver/install/dns.py
```diff
@@ -487,6 +488,16 @@ def zonemgr(self, value):
             error = unicode(e).encode(stderr_encoding)
             raise ValueError(error)
 
+    forwarder_loopback = knob(
```
stlaz: DO NOT ADD a new know, modify the old one.

ipapython/config.py
```diff
@@ -31,7 +31,7 @@
 from six.moves.configparser import SafeConfigParser
 from six.moves.urllib.parse import urlsplit
 # pylint: enable=import-error
-
+from ipapython.ipa_log_manager import root_logger
```
stlaz: You left this here.

felipevolpone:
I've changed the knob to use `CheckedIPAddressLoopback`, however now I'm getting the error:
```
[root@server vagrant]# ipa-server-install --forwarder 127.0.0.11 [...]
ipa-server-install: error: option --forwarder: failed to detect a valid IP address from 'None'`
```

If I tried to pass the argument using `=` the error changes to:
```
[root@server vagrant]# ipa-server-install --forwarder=127.0.0.11 [...]
ipa-server-install: error: --forwarder option does not take a value
``` 
Checking the [optparse source code](https://pastebin.com/UD790ehG) (line 1489)` the error is raised 
 due to `option.takes_value()` returns `None`. And the code of method `takes_value` (line 752) is `return self.type is not None`.

Soooo, the problem seems to be in the `option.type` property, for some reason it's not being passed to `Knob`.

Also, the `addr` parameter in `CheckedIPAddressLoopback __init__` is None.

install/tools/ipa-dns-install
```diff
@@ -50,6 +50,10 @@ def parse_options():
                       type="ip",
                       help="Master Server IP Address. This option can be used "
                            "multiple times")
+    parser.add_option("--forwarder-loopback", dest="forwarders_loopback",
+                      action="append", type="ip_with_loopback",
+                      help="Add a DNS forwarder, being able to use a loopback"
+                      " IP. This option can be used multiple times")
     parser.add_option("--forwarder", dest="forwarders", action="append",
                       type="ip", help="Add a DNS forwarder. This option can be used multiple times")
```
MartinBasti: use `type="ip_with_loopback"` here

install/tools/ipa-dns-install
```diff
@@ -50,6 +50,10 @@ def parse_options():
                       type="ip",
                       help="Master Server IP Address. This option can be used "
                            "multiple times")
+    parser.add_option("--forwarder-loopback", dest="forwarders_loopback",
```
MartinBasti: Don't add a new option, use the --forwarder option

ipapython/ipautil.py
```diff
@@ -243,6 +243,21 @@ def set_ip_net(self, ifnet):
...
+    Reserved or link-local addresses are never accepted.
+    """
+    def __init__(self, addr, parse_netmask=True, allow_multicast=False):
+
+        super(CheckedIPAddressLoopback, self).__init__(
+                addr, parse_netmask=parse_netmask, allow_loopback=True,
+                allow_multicast=allow_multicast)
+
+        if self.is_loopback():
+            root_logger.error("you're using a loopback ip")
```
MartinBasti: use `root_logger.warning` please

felipevolpone: ok

install/tools/ipa-dns-install
felipevolpone: ok

MartinBasti:
in `ipapython.install.cli`:
```
            elif knob_scalar_type is CheckedIPAddress:
                kwargs['type'] = 'ip'
```
this should be extended with
```
            elif knob_scalar_type is CheckedIPAddressLoopback:
                kwargs['type'] = 'ip_with_loopback'
```

MartinBasti:
@HonzaCholasta ^^ this looks suboptimal to me, can it be extracted somewhere to make it more visible, it took me really long time until I was able to find that in cli.py is missing definition for OptionParser

ipapython/ipautil.py
```diff
@@ -243,6 +243,21 @@ def set_ip_net(self, ifnet):
...
+    Reserved or link-local addresses are never accepted.
+    """
+    def __init__(self, addr, parse_netmask=True, allow_multicast=False):
+
+        super(CheckedIPAddressLoopback, self).__init__(
+                addr, parse_netmask=parse_netmask, allow_loopback=True,
+                allow_multicast=allow_multicast)
+
+        if self.is_loopback():
+            root_logger.warning("you're using a loopback ip")
```
MartinBasti: This changed recently to just `logger`, please rebase and fix

MartinBasti: Actually, I want to see the IP Address in warning msg (there might be multiple addresses in IPv4 world)

MartinBasti:
I'm almost sure that `Fixing ipa-replica-install --setup-kra if it's the first KRA in topology` is not a right description of changes made here

MartinBasti:
`logger` only as I said in inline comment

```
************* Module ipapython.ipautil
ipapython/ipautil.py:260: [E0602(undefined-variable), CheckedIPAddressLoopback.__init__] Undefined variable 'root_logger')
```

ipapython/config.py
```diff
@@ -31,7 +31,6 @@
 from six.moves.configparser import SafeConfigParser
 from six.moves.urllib.parse import urlsplit
 # pylint: enable=import-error
-
```
stlaz: Please, keep the newline, it separates categories of imports.

ipapython/config.py
```diff
@@ -71,7 +70,17 @@ def check_ip_option(option, opt, value):
     try:
         return CheckedIPAddress(value)
     except Exception as e:
-        raise OptionValueError("option %s: invalid IP address %s: %s" % (opt, value, e))
+        raise OptionValueError("option %s: invalid IP address %s: %s" %
```
stlaz: When you're changing this, why not to change it to the newer `str.format()` way right away?

ipapython/config.py
```diff
@@ -71,7 +70,17 @@ def check_ip_option(option, opt, value):
...
+                               (opt, value, e))
+
+def check_ip_option_with_loopback(option, opt, value):
```
stlaz: Did the `functools.partial` solution not work for you?

ipapython/ipautil.py
```diff
@@ -245,6 +245,21 @@ def set_ip_net(self, ifnet):
...
 
+class CheckedIPAddressLoopback(CheckedIPAddress):
+    """IPv4 or IPv6 address with additional constraints with
+    possibility to use a loopback IP.
+    Reserved or link-local addresses are never accepted.
+    """
+    def __init__(self, addr, parse_netmask=True, allow_multicast=False):
+
+        super(CheckedIPAddressLoopback, self).__init__(
+                addr, parse_netmask=parse_netmask, allow_loopback=True,
```
stlaz: Personally, if I were to change one specific option here, I would probably add it as a last argument, but that's up to you, I guess.

felipevolpone: Actually, I would like to remove the parameters from `__init__` and have something like: 
```python
    def __init__(self, addr):

        super(CheckedIPAddressLoopback, self).__init__(
                addr, 
                parse_netmask=parse_netmask, 
                allow_multicast=allow_multicast,
                allow_loopback=True)
```

Would this break something?

ipapython/config.py
felipevolpone: TBH I think that doing this way is more straightforward. IMHO `functools.partial` adds an unnecessary complexity to the code.

ipapython/config.py
felipevolpone: Nice, I'll change it.

ipapython/config.py
felipevolpone: ack

ipapython/config.py
stlaz: It's more straightforward to code, not to maintain. Once someone comes saying he does not like the error message, you go and change it on one spot, forget about the other, this goes to Bugzilla, QA will push it back to you, and you have to fix it once more.
Having `functools.partial` here would allow you to prevent this circle.
But if you don't want it there, fine, I won't hold this PR just because of that.

ipapython/ipautil.py
```diff
@@ -245,6 +245,22 @@ def set_ip_net(self, ifnet):
...
+    """
+    def __init__(self, addr, parse_netmask=True, allow_multicast=False):
+
+        super(CheckedIPAddressLoopback, self).__init__(
+                addr, parse_netmask=parse_netmask,
+                allow_multicast=allow_multicast,
+                allow_loopback=True)
+
+        if self.is_loopback():
+            logger.warning("You are using a loopback ip: %s", addr)
```
stlaz: IP should be in capitals.

stlaz:
I was testing this and was kind of surprised to see that the warning is never printed. Actually, I was surprised at first, upset later. It would seem that the class is never set up like that but instead just the checker is triggered.
I originally made the following change:
```python
 def check_ip_option(option, opt, value, allow_loopback=False):
     try:
        if allow_loopback:
            return CheckedIPAddressLoopback(value)
        else:
            return CheckedIPAddress(value)
     except Exception as e:
         raise OptionValueError("option {}: invalid IP address {}: {}"
                                .format(opt, value, e))
```
but that lead to `No handlers could be found for logger "ipapython.ipautil"` being printed at the beginning of the `ipa-server-install`.

I am not sure how this could be resolved. It would seem that it would be easiest not to print the warning until the installer refactoring is finished and the type check is performed by initializing the appropriate type of the knob with the supplied value. In that case, a new ticket for printing the warning would be appropriate. Otherwise you should probably dig for why the logger handler warning is printed.

felipevolpone:
The loggers have been refactored by @HonzaCholasta, maybe something changed that we are not aware of. 

@HonzaCholasta could you help us here? 

Debugging the code, I discovered that when `logger.warning` is called for its first time, the message is raised: `No handlers could be found for logger "ipapython.ipautil"`. If I call it once again, the exception is not raised.

As you can see here:
```bash
[vagrant@server ~]$ sudo python -m pdb /usr/sbin/ipa-server-install --domain $DOMAIN --realm $REALM --setup-dns --ip-address $IP  --ssh-trust-dns --hostname `hostname` --no-host-dns --forwarder 127.0.0.11 -p $PASSWORD -a $PASSWORD -U

(Pdb) tbreak ipapython/ipautil.py:56
Breakpoint 1 at /usr/lib/python2.7/site-packages/ipapython/ipautil.py:56
(Pdb) c
Deleted breakpoint 1
> /usr/lib/python2.7/site-packages/ipapython/ipautil.py(56)<module>()
-> logger = logging.getLogger(__name__)
(Pdb) n
> /usr/lib/python2.7/site-packages/ipapython/ipautil.py(59)<module>()
-> TMP_PWD_ENTROPY_BITS = 128
(Pdb) logger.info('bla')
(Pdb) logger.warning("bla")
No handlers could be found for logger "ipapython.ipautil"
(Pdb) logger.warning("bla")
(Pdb)
```

tomaskrizek:
Since this PR is still in progress, I'm dropping the prioritized flag, because it won't make it into 4.5.3 release.

felipevolpone:
Getting deeper into the code and the changes made in the log, I think I found the problem:
In `ipapython/admintool.py::AdminTool::main` method (that will validate the input options), 
`self._setup_logging` was not called yet, so there is no logging handler to be used.

That would explain why the exception is not raised in the second time logger.warning is called: after calling logger.* (warning, info, error, debug, etc) for the first time (without setting a handler), the logging module calls `logging.basicConfig`. 

I was able to fix the problem adding `standard_logging_setup(None, debug=True)` as first line to `ipapython/admintool.py::AdminTool::main` method. Not sure if it's the right approach. If it is, I can create another PR to fix it.

felipevolpone:
To this PR works, it depends on #937

stlaz:
Needs rebase on current master.

felipevolpone:
Done

ipapython/ipautil.py
```diff
@@ -245,6 +245,24 @@ def set_ip_net(self, ifnet):
...
+
+        super(CheckedIPAddressLoopback, self).__init__(
+                addr, parse_netmask=parse_netmask,
+                allow_multicast=allow_multicast,
+                allow_loopback=True)
+
+        if self.is_loopback():
+            # print is being used instead of a logger, because at this
+            # moment, in execution process, there is no logger configured
+            print("You are using a loopback IP: {}".format(addr))
```
stlaz: Should be at least printed to stderr. The bad thing is it's not going to appear in our logs but I guess we should be able to spot a loopback IP ourselves.

felipevolpone: done

ipapython/ipautil.py
```diff
@@ -245,6 +245,25 @@ def set_ip_net(self, ifnet):
...
+
+        super(CheckedIPAddressLoopback, self).__init__(
+                addr, parse_netmask=parse_netmask,
+                allow_multicast=allow_multicast,
+                allow_loopback=True)
+
+        if self.is_loopback():
+            # print is being used instead of a logger, because at this
+            # moment, in execution process, there is no logger configured
+            print("You are using a loopback IP: {}".format(addr),
```
stlaz: Please make this a "WARNING: You are using a ..."

felipevolpone:
Can we have an ack on this PR?

stlaz:
Yes, sorry, this slipped from my usual PR checking loop.

felipevolpone:
master:

* 405da071d109ec683676d56fac3bccfc4606535e Warning the user when using a loopback IP as forwarder

PR #908, opened 2017-07-10
WebUI: Actions array in search facet spec does not have 'undefined' value anymore

pvomacka:
In case that there are no actions specified in spec object, the concatenation
of arrays causes that 'undefined' item is added into the array.

https://pagure.io/freeipa/issue/7052

### DIFF
--- a/install/ui/src/freeipa/search.js
+++ b/install/ui/src/freeipa/search.js
@@ -63,7 +63,7 @@ exp.search_facet_control_buttons_pre_op = function(spec, context) {
     var default_actions = ['refresh', 'batch_remove', 'add'];
     var merged_actions = override_actions(spec.custom_actions, default_actions);
 
-    spec.actions = merged_actions.concat(spec.actions);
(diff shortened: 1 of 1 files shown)


pvoborni:
The code works even though it can be written as: 
```javascript
spec.actions = merged_actions.concat(spec.actions || []);
```
If you want to keep the patch as is and CI is green then it is an ACK.

pvoborni:
Unrelated comment, a thought for future changes: I was thinking if the merging mechanism should be allowed by default for all facets. And e.g. the fuction `override_actions` be refactored into `merge_spec_array` and be part of e.g. construct.js or Spec_mod.js to be reusable elsewhere.

pvomacka:
@pvoborni about your thoughts for future, that's a good idea to do it even more general, but I wouldn't do it right now (as it is not related to this ticket) and wait till some other facet will require it.

tomaskrizek:
master:

* 76f217b2893b0604ed0adf372422c685c2febbe8 Fixes bug in actions creating for search facet

PR #911, opened 2017-07-11
WebUI: fix for negative number in pagination size settings

pvomacka:
This PR adds new validator which allows to input only positive numbers into fiels. The new validator is then used in field for setting the length of tables (number of rows) on search facets.

https://pagure.io/freeipa/issue/6980


felipevolpone:
I've tested and it works properly :+1:

martbab:
master:

* 3cac851498ab85e6b0abf74bc3ff4eea2f960983 WebUI: Add positive number validator
* cfa157c1e554e0cd113a6fb9fd3af9fbad4529ef WebUI: change validator of page size settings

PR #916, opened 2017-07-13
Py3 adtrust

MartinBasti:


abbra:
I think you also can explicitly raise samba dependency version to 4.7.0 or above. Nothing else provides python3-samba subpackage.

MartinBasti:
We will need to bump samba to even higher version anyway to a newer version because not everything what we need is accessible in py3 (`samba.dcerpc` for example)

stlaz:
The troubled imports do not show on my Fedora 26 machine. Could we speed up the review of https://github.com/freeipa/freeipa/pull/992 to finally get f26 into our testing containers?
I suppose this might be some kind of pylint screw up which might be fixed in later versions.

stlaz:
Since Travis build is constantly failing, I tried to work around that including this in another PR: https://github.com/freeipa/freeipa/pull/1052

PR #918, opened 2017-07-14
dcerpc: support Python 3

abbra:
Make 'ipa trust-add' work under Python 3. One needs patches from
https://lists.samba.org/archive/samba-technical/2017-July/121746.html
to Samba too.

Marked as WIP until Samba patches are released as part of 4.7.0-RC2.

Since we haven't switched whole ipa server environment to Python 3 yet,
following packages need to be installed to make trust code working under
Python 3:

 - python3-libsss_nss_idmap
 - python3-sss
 - python3-samba
 - python3-mod_wsgi

### DIFF
--- a/ipaserver/dcerpc.py
+++ b/ipaserver/dcerpc.py
@@ -188,7 +188,7 @@ class ExtendedDNControl(LDAPControl):
         self.integerValue = 1
 
     def encodeControlValue(self, value=None):
-        return '0\x03\x02\x01\x01'
+        return b'0\x03\x02\x01\x01'
 
 
 class DomainValidator(object):
@@ -820,6 +820,8 @@ class DomainValidator(object):
 
 
 def string_to_array(what):
+    if six.PY3 and isinstance(what, bytes):
+        return [v for v in what]
     return [ord(v) for v in what]
 
 
@@ -940,6 +942,14 @@ class TrustDomainInstance(object):
         search_result = None
         try:
             _objtype, res = conn.search_s('', _ldap.SCOPE_BASE)[0]
+            for o in res.keys():
+                if isinstance(res[o], list):
+                    t = res[o]
+                    for z, v in enumerate(t):
+                        if isinstance(v, bytes):
+                            t[z] = v.decode('utf-8')
+                elif isinstance(res[o], bytes):
+                    res[o] = res[o].decode('utf-8')
             search_result = res['defaultNamingContext'][0]
             self.info['dns_hostname'] = res['dnsHostName'][0]
         except _ldap.LDAPError as e:


ipaserver/dcerpc.py
```diff
@@ -938,6 +940,15 @@ def retrieve_anonymously(self, remote_host,
         search_result = None
         try:
             _objtype, res = conn.search_s('', _ldap.SCOPE_BASE)[0]
+            for o in res.keys():
+                if isinstance(res[o], (list, tuple)):
+                    t = res[o]
+                    for z in range(len(t)):
+                        if isinstance(t[z], bytes):
+                            t[z] = t[z].decode('utf-8')
+                    res[o] = t
```
MartinBasti: This is not needed, python keeps reference

abbra:
Using ipaldap here is not reasonable. We only need to perform a single anonymous rootDSE request with a special control to retrieve a SID. Thus, the code here is very limited.

I removed tuple handling because it is not needed and left only a list. Now `enumerate()` works for me without problems.

ipaserver/dcerpc.py
```diff
@@ -820,6 +820,8 @@ def __retrieve_trusted_domain_gc_list(self, domain):
 
 
 def string_to_array(what):
+    if isinstance(what, bytes):
```
tiran: Under Python 2 ``str is bytes``, so this will return a list of strings and not a list of ints.

ipaserver/dcerpc.py
```diff
@@ -940,6 +942,14 @@ def retrieve_anonymously(self, remote_host,
         search_result = None
         try:
             _objtype, res = conn.search_s('', _ldap.SCOPE_BASE)[0]
+            for o in res.keys():
+                if isinstance(res[o], list):
+                    t = res[o]
+                    for z, v in enumerate(t):
+                        if isinstance(v, bytes):
+                            t[z] = v.decode('utf-8')
```
tiran: Under Python 2, this will return unicode. Is this intended?

abbra: Yes, it is intended to be unicode here in py2 case.

ipaserver/dcerpc.py
abbra: I added 'if six.PY3 and isinstance ...' as discussed on IRC.

abbra:
Rebased and fixed one check as per discussion with @tiran

stlaz:
Works for me, ACK. Thanks!

pvomacka:
Hi @abbra, 

could you please add related ticket link into the commit message? 

Thank you.

stlaz:
@abbra https://pagure.io/freeipa/issue/4985 should be just fine for this.

abbra:
Pushed an update that adds ticket 4985 to the commit message.

pvomacka:
master:

* 928374ca7de5c0bec54e848a35b6bacda9f55343 dcerpc: support Python 3

PR #922, opened 2017-07-18
logging: make sure logging level is set to proper value

tomaskrizek:
During py.test initialization, the value 'debug' is passed instead
of logging.DEBUG.

### DIFF
--- a/ipalib/plugable.py
+++ b/ipalib/plugable.py
@@ -461,21 +461,7 @@ class API(ReadOnly):
             if not match:
                 continue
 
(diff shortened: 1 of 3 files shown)


tomaskrizek:
@dkupka Could you please take a look at this PR? We need to merge this so we can run PR CI on master branch.

martbab:
master:

* ba5f7afedca282e79fdfde0dd359028fc4111f90 logging: make sure logging level is set to proper value

PR #923, opened 2017-07-18
py3: fixing zonemgr_callback

felipevolpone:
Previously, `zonemgr_callback` was expecting unicode, but getting bytes.

https://pagure.io/freeipa/issue/5990

### DIFF
--- a/ipaserver/install/bindinstance.py
+++ b/ipaserver/install/bindinstance.py
@@ -423,7 +423,11 @@ def zonemgr_callback(option, opt_str, value, parser):
             encoding = getattr(sys.stdin, 'encoding', None)
             if encoding is None:
                 encoding = 'utf-8'
-            value = value.decode(encoding)
+
+            # value is of a string type in both py2 and py3
+            if not isinstance(value, unicode):
+                value = value.decode(encoding)
+
             validate_zonemgr_str(value)
         except ValueError as e:
             # FIXME we can do this in better way
@@ -433,7 +437,7 @@ def zonemgr_callback(option, opt_str, value, parser):
             if stderr_encoding is None:
                 stderr_encoding = 'utf-8'
             error = unicode(e).encode(stderr_encoding)
-            parser.error("invalid zonemgr: " + error)
+            parser.error(b"invalid zonemgr: " + error)
 
     parser.values.zonemgr = value
 
--- a/ipalib/util.py
+++ b/ipalib/util.py
@@ -379,7 +379,7 @@ def validate_domain_name(domain_name, allow_underscore=False, allow_slash=False)
 
 def validate_zonemgr(zonemgr):
     assert isinstance(zonemgr, DNSName)
-    if any('@' in label for label in zonemgr.labels):
+    if any(b'@' in label for label in zonemgr.labels):
         raise ValueError(_('too many \'@\' characters'))
 
 


felipevolpone:
I've tested these scenarios (in py2 and py3): 
```
ipa-dns-install --zonemgr jélirãsó@mail.com
ipa-dns-install --zonemgr jélirã@só@mail.com # here the exception is raised properly        
```
I'm not sure if there is other case to test it.

ipaserver/install/bindinstance.py
```diff
@@ -423,17 +423,15 @@ def zonemgr_callback(option, opt_str, value, parser):
             encoding = getattr(sys.stdin, 'encoding', None)
             if encoding is None:
                 encoding = 'utf-8'
-            value = value.decode(encoding)
+            if not isinstance(value, unicode):
```
stlaz: I don't think this is ever going to be bytes.

ipaserver/install/bindinstance.py
```diff
@@ -423,17 +423,15 @@ def zonemgr_callback(option, opt_str, value, parser):
...
         except ValueError as e:
             # FIXME we can do this in better way
             # https://fedorahosted.org/freeipa/ticket/4804
             # decode to proper stderr encoding
-            stderr_encoding = getattr(sys.stderr, 'encoding', None)
-            if stderr_encoding is None:
-                stderr_encoding = 'utf-8'
-            error = unicode(e).encode(stderr_encoding)
-            parser.error("invalid zonemgr: " + error)
+            parser.error("invalid zonemgr: " + unicode(e))
```
stlaz: According to ttps://fedorahosted.org/freeipa/ticket/4804, this is a wrong fix.

ipaserver/install/bindinstance.py
stlaz: I had some more time and I see now why you did this here. `OptionParser` behaves differently in Python 2/3 so it actually makes sense to do this. If you could probably drop a comment there.
`value` will never be of `bytes` type, though, I believe you did not express yourself well in the commit message. In Python 3, `value` is of `str` type which does not have `decode()`. Unfortunately, it's also of `str` type in Python 2 which does has the `decode()` method.

ipaserver/install/bindinstance.py
stlaz: I suppose the original purpose of this encoding excercise was to do
`parser.error(b"invalid zonemgr: " + error)`

ipaserver/install/bindinstance.py
felipevolpone: > OptionParser behaves differently in Python 2/3 so it actually makes sense to do this

Yep :)
Good point, I'll change the commit message and add a code comment there as well. What do you think it would be a good comment that express what you're saying? 

Something like this? 
```python
...
# value is str in py2 and py3
# in py3 isinstance(value, unicode) returns True
if not isinstance(value, unicode):
...
```

stlaz: Good!
I'd go just with the comment: `# value is of a string type in both py2 and py3`

ipaserver/install/bindinstance.py
felipevolpone: If I understood correctly what you're saying, this would not help if `error = unicode(e)`, because in python3 is not possible to "add" str +  bytes.

*Python 3*:  
```python
[root@server vagrant]# python3.5
Python 3.5.3 (default, May 10 2017, 15:05:55)
>>> b"a" + str("b")
TypeError: can't concat bytes to str
>>> b"a" + "b"
TypeError: can't concat bytes to str
```

*Python 2.7*
```python
[root@server vagrant]# python2.7
Python 2.7.13 (default, May 10 2017, 20:04:28)
>>> b"a" + unicode("b")
u'ab'
>>> b"a" + str("b")
'ab'
>>> b"a" + "b"
'ab'
```

This `parser.error("invalid zonemgr: " + unicode(e))` black magic works (to py3 and py2), because at the begging of file we have (line 62):
```python
if six.PY3:
    unicode = str
```

I've tested with the error described in ticket 4804, and with this new implementation it works:
```bash
[root@server vagrant]# python2  /usr/sbin/ipa-dns-install --zonemgr=TÃªko@example.com
ipa-dns-install: error: invalid zonemgr: domain name 'TÃªko.example.com' should be normalized to: tãako.example.com\

[root@server vagrant]# python3  /usr/sbin/ipa-dns-install --zonemgr=TÃªko@example.com
ipa-dns-install: error: invalid zonemgr: domain name 'TÃªko.example.com' should be normalized to: tãako.example.com
```

felipevolpone:
@stlaz I've changed the commit message, please check if it looks good to you.

ipaserver/install/bindinstance.py
stlaz: I meant you should keep the original code and just do:
`parser.error(b"invalid zonemgr: " + error)`
since `.encode(whatever)` will give you bytes.

felipevolpone: Ah ok, got it

ipaserver/install/bindinstance.py
```diff
@@ -423,17 +423,18 @@ def zonemgr_callback(option, opt_str, value, parser):
...
+
+            # value is of a string type in both py2 and py3
+            if not isinstance(value, unicode):
+                value = value.decode(encoding)
+
             validate_zonemgr_str(value)
         except ValueError as e:
-            # FIXME we can do this in better way
-            # https://fedorahosted.org/freeipa/ticket/4804
-            # decode to proper stderr encoding
```
stlaz: I don't see why this comment should be removed.

stlaz:
LGTM, ACK.

pvomacka:
master:

* 75d26e1f0121f875bdb017b0636c02a6f5660e8a py3: fixing zonemgr_callback

PR #925, opened 2017-07-20
Move x509.py from ipalib to ipapython

stlaz:
x509.py does not really contain any code which is connected to
IPA API, move it to the ipapython module.
This change fixes several cases of import cycles.

-----------------
This PR depends on https://github.com/freeipa/freeipa/pull/878 and contains all of its commits so that I don't have to do the same change during rebase. Only the latest commit is related to this PR.


stlaz:
Weird, seems like `_subject_base` was actually somehow needed although it does not seem to be used anywhere in the code.

stlaz:
Huh, the requested certificate eventually gets in the `MONITORING` state yet this is not propagated to FreeIPA. I will look some more into that when I arrive back on Wednesday.

stlaz:
Hm, I did not notice the `ipalib.errors` import on the top of the file. One would expect that `errors` module could also be separated from `ipalib` but noooo. I won't deal with this, closing the PR, I may get to it later.

PR #928, opened 2017-07-21
WebUI: fix jslint error

pvomacka:
jslint warned about parsing string to integer without explicit radix.
This error was introduced in commit 3cac851.

The PR is without ticket as it is only one liner.


abbra:
ACK.

martbab:
master:

* 0b8d46019172e83bdbacf7200759de640c19a2d4 WebUI: fix jslint error

PR #931, opened 2017-07-26
ipactl: log check_version exception

tomaskrizek:
When version is mismatched and ipa-server-upgrade is required,
log the version mismatch properly in journal.

### DIFF
--- a/install/tools/ipactl
+++ b/install/tools/ipactl
@@ -138,7 +138,8 @@ def version_check():
     try:
         installutils.check_version()
     except (installutils.UpgradeMissingVersionError,
-            installutils.UpgradeDataOlderVersionError):
+            installutils.UpgradeDataOlderVersionError) as exc:
+        emit_err("IPA version error: %s" % exc)
         emit_err("Upgrade required: please run ipa-server-upgrade command")
         raise IpactlError("Aborting ipactl")
     except installutils.UpgradeVersionError as e:


tomaskrizek:
This is just a drive-by case I wanted to fix, because such error message would be helpful for debugging. It's a simple fix, but I can create a ticket if needed.

install/tools/ipactl
```diff
@@ -138,7 +138,8 @@ def version_check():
     try:
         installutils.check_version()
     except (installutils.UpgradeMissingVersionError,
-            installutils.UpgradeDataOlderVersionError):
+            installutils.UpgradeDataOlderVersionError) as exc:
+        emit_err("IPA version error: %s" % exc)
```
MartinBasti: Please use .`str.format()`, it is preferred

tomaskrizek: This is an attempt to be consistent within the function. In some other places of the file, there are also str(e) usages... I'm either keeping it this way or I'm fixing the entire file. Which would you prefer?

MartinBasti: I'm fine with just changing this particular line, I don't think that in this case is needed to keep a consistency.

pvomacka:
master:

* fc7e2565bdb2b2a40380ffb343d8a7295ac21cd7 ipactl: log check_version exception

PR #932, opened 2017-07-26
replica-conncheck: log when failed to RPC connect

stlaz:
It's nearly impossible to find out what happened when doing
replica connection check and it fails during the RPC phase.
The error is now logged.


pvomacka:
master:

* 06fbf4b312777751085f3dcfeff456b037ee2737 replica-conncheck: log when failed to RPC connect

PR #933, opened 2017-07-26
Checks if Directory Server is installed and running before installation

felipevolpone:
In cases when IPA is installed in two steps (external CA), it's necessary to check (in the second step) if Directory Server is running and if it's installed before continue with the IPA installation.

Fixes: https://pagure.io/freeipa/issue/6611

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -618,6 +618,14 @@ def install_check(installer):
         # check addresses here, dns module is doing own check
         no_matching_interface_for_ip_address_warning(ip_addresses)
 
+    instance_name = "-".join(realm_name.split("."))
+    dirsrv = services.knownservices.dirsrv
+    if (options.external_cert_files
+           and dirsrv.is_installed(instance_name)
+           and not dirsrv.is_running(instance_name)):
+        logger.debug('Starting Directory Server')
+        services.knownservices.dirsrv.start(instance_name)
+
     if options.setup_adtrust:
         adtrust.install_check(False, options, api)
 
--- a/ipaplatform/redhat/services.py
+++ b/ipaplatform/redhat/services.py
@@ -121,6 +121,10 @@ class RedHatDirectoryService(RedHatService):
 
         return True
 
+    def is_installed(self, instance_name):
+        file_path = "{}/{}-{}".format(paths.ETC_DIRSRV, "slapd", instance_name)
+        return os.path.exists(file_path)
+
     def restart(self, instance_name="", capture_output=True, wait=True,
                 ldapi=False):
     # We need to explicitly enable instances to install proper symlinks as


ipaserver/install/server/install.py
```diff
@@ -336,6 +336,12 @@ def install_check(installer):
             "Please uninstall it before configuring the IPA server, "
             "using 'ipa-client-install --uninstall'")
 
+    instance_name = "-".join(options.realm_name.split("."))
+    if (is_ipa_configured() and
```
stlaz: The correct way to check whether this is a 2nd step of external-CA installation is on (original) line 729:
```python
if not options.external_cert_files:
    ... code ...
```
I would prefer all such code is to be found at that very spot if it's not really necessary to do otherwise and I don't believe it is in this case.

I am not sure whether DS should be started in this case or not, if it were me, I would just probably have it started since it's a requirement for the installation.

felipevolpone: Well, if the right way to check is with `if not options.external_cert_files`, I think I could put the code in line [406](https://github.com/freeipa/freeipa/blob/master/ipaserver/install/server/install.py#L406), so.

I think it makes more sense to do the check in the `install_check` method instead of in `install`, what do you think?

> I am not sure whether DS should be started in this case or not, if it were me, I would just probably have it started since it's a requirement for the installation.

In my POV, since this is a corner case, I think we should not re-start DS because this is already done in the first step. Warn/raise an error is enough to the user take some action.

pvoborni: If it is not started by the installer, then one needs to do it manually. Context from the original bug:
```
This causes problems especially in unattended container installations when completely new container is run for the second phase and the dirsrv is thus not up.
```
From this perspective it looks better to start it. 

From top of my head I cannot tell if there can be similar issue with other service we might depend on.

felipevolpone:
rebasing commit to trigger PR CI

felipevolpone:
triggering travis

frasertweedale:
@felipevolpone could you rebase this?  I will review it.

ipaserver/install/server/install.py
```diff
@@ -618,6 +618,14 @@ def install_check(installer):
         # check addresses here, dns module is doing own check
         no_matching_interface_for_ip_address_warning(ip_addresses)
 
+    instance_name = "-".join(options.realm_name.split("."))
```
frasertweedale: If realm name is not explicitly given as an argument to `--ipa-server-install`, this causes
installation to fail with:
```
The ipa-server-install command failed, exception: AttributeError: 'NoneType' object has no attribute 'split'
```

It should just be `realm_name.split` instead of `options.realm_name.split`, I think.

frasertweedale: I made this change and tested;  once you make this little fix, it is an ACK from me.

ipaserver/install/server/install.py
```diff
@@ -618,6 +618,14 @@ def install_check(installer):
...
+    dirsrv = services.knownservices.dirsrv
+    if (options.external_cert_files and
+       dirsrv.is_installed(instance_name) and not
+       dirsrv.is_running(instance_name)):
```
frasertweedale: I would rather see it indented per this example from PEP 8:
```python
# Add some extra indentation on the conditional continuation line.
if (this_is_one_thing
        and that_is_another_thing):
    do_something()
```

But it is not a blocker for me, just a nit-pick :)

ipaserver/install/server/install.py
felipevolpone: ack, thanks.

ipaserver/install/server/install.py
felipevolpone: Looks more readable to me too :)

felipevolpone:
Thanks for reviewing the patch Fraser. 

I did the changes, waiting for PR CI.

tomaskrizek:
master:

* bf0b74bec4dc8b75a71d9e8d1374755a81d8b1df Checks if Dir Server is installed and running before IPA installation

PR #935, opened 2017-07-27
ca: TypeError fix

stlaz:
We were trying to load a certificate as a string even though it
was an object already.

------------
Please push this before it gets to our integration test suite.


pvomacka:
master:

* 0c848b791df7368a32434b3ac87a45728abc2d0f ca: TypeError fix

PR #936, opened 2017-07-27
cainstance: read cert file as bytes

stlaz:

------
Another small thing that slipped through my fingers, happens during external CA installation.


stlaz:
ACKing under one-line rule.

pvomacka:
master:

* c95617e71495fff4bf2352fdeb5089e3e02bd79b cainstance: read cert file as bytes

PR #937, opened 2017-07-27
Configuring log handlers during the validation of input parameters

felipevolpone:
Previously, a log handler would be configured only after all the input parameters be validated, as can be checked in `ipapython/admintool.py::AdminTool::main`. So, any call to `logger.[warning,info,error,debug]`, during that phase, doesn't work and it also raises an exception. 

Now, log handlers are setup before the input parameters validation phase.

Fixes: https://pagure.io/freeipa/issue/7071

### DIFF
--- a/ipapython/admintool.py
+++ b/ipapython/admintool.py
@@ -136,6 +136,8 @@ class AdminTool(object):
         :param argv: Command-line arguments.
         :return: Command exit code
         """
+        standard_logging_setup(None, verbose=True)
+
         if cls not in cls._option_parsers:
             # We use cls._option_parsers, a dictionary keyed on class, to check
             # if we need to create a parser. This is because cls.option_parser
--- a/ipapython/ipa_log_manager.py
+++ b/ipapython/ipa_log_manager.py
@@ -157,6 +157,7 @@ def standard_logging_setup(filename=None, verbose=False, debug=False,
 
     root_logger = logging.getLogger()
     root_logger.setLevel(logging.DEBUG)
+    root_logger.handlers = []
 
     # File output is always logged at debug level
     if filename is not None:


felipevolpone:
To test this PR you can apply the patch from PR #903.

felipevolpone:
Rebasing to get PR CI

rcritten:
Can you provide an example here and/or the ticket of when this may happen?

felipevolpone:
Sure. First, you need to apply PR 903. 
Then, you apply this patch.
After that, all you need to do is call `ipa-server-install` using a loopback ip in the `--forwarder` option.

Like this:
```
ipa-server-install --domain $DOMAIN --realm $REALM --setup-dns --ip-address $IP  --ssh-trust-dns --hostname `hostname` --no-host-dns --forwarder 127.0.0.11 -p $PASSWORD -a $PASSWORD -U 
```

You should get a log output with the message:
```
[root@example vagrant]# ipa-server-install --domain $DOMAIN --realm $REALM --setup-dns --ip-address $IP  --ssh-trust-dns --hostname `hostname` --no-host-dns --forwarder 127.0.0.11 -p $PASSWORD -a $PASSWORD -U
ipapython.ipautil: WARNING  You are using a loopback IP: 127.0.0.11
``` 

Without this PR, a message would be printed: `No handlers could be found for logger "ipapython.ipautil"` 

PS: You "need" to apply PR 903, because it has a log call during validation phase.

frasertweedale:
@felipevolpone maybe this need to be rebased to make the CI happy?

frasertweedale:
OK, so.  I just did an `ipa-server-install` with this patch applied (among others).

In the client install step, all the log messges appear twice:
```
  [9/9]: starting directory server                                                      
Done.                                                                                   
Restarting the KDC                                                                      
Please add records in this file to your DNS system: /tmp/ipa.system.records.2fefdza8.db 
Configuring client side components                                                      
Using existing certificate '/etc/ipa/ca.crt'.                                           
Using existing certificate '/etc/ipa/ca.crt'.                                           
Client hostname: f26-1.ipa.local                                                        
Client hostname: f26-1.ipa.local                                                        
Realm: IPA.LOCAL                                                                        
Realm: IPA.LOCAL                                                                        
DNS Domain: ipa.local                                                                   
DNS Domain: ipa.local                                                                   
IPA Server: f26-1.ipa.local                                                             
IPA Server: f26-1.ipa.local                                                             
BaseDN: dc=ipa,dc=local                                                                 
BaseDN: dc=ipa,dc=local                                                                 
... and so on
```

I guess that this patch is the cause?  I which case, NACK until fixed.

felipevolpone:
PR updated with the proper fix.

I've also checked the outputs from commands `ipa-server-install, ipa-server-upgrade, ipa-kra-install` and the logs `ipaserver-install, ipaserver-uninstall, ipaclient-install, ipaclient-uninstall`. They're still the same after this patch being applied.

ipapython/admintool.py
```diff
@@ -136,6 +136,8 @@ def main(cls, argv):
         :param argv: Command-line arguments.
         :return: Command exit code
         """
+        standard_logging_setup(None, verbose=True)
```
stlaz: This is called later in the execution.
Your so-called proper fix is no proper fix at all, you should have realized that something's wrong when @frasertweedale showed you the doubled output.
Please fix this properly and don't call `standard_logging_setup()` twice.

stlaz: I've been a bit prickly in my previous comment, sorry. Please, find another way to do this.

felipevolpone: Hi @stlaz, thanks for reviewing this. Do you think that is so wrong call `standard_logging_setup` another time? IMHO it's harmless. Actually, it is already called twice*, my patch would call it a third time.

Maybe, a better fix would be call `self._setup_logging(no_file=True)`, however `main` is a class method, so it's not possible call `_setup_logging`.

* It's called twice because in `execute` method, it calls `self._setup_logging(no_file=True)` and some lines bellow `self.setup_logging()`. Both methods calls `standard_logging_setup`.

stlaz: It is wrong and what you’re pointing out is a hack. I’ll post more tomorrow when I’m not in a cinema :)

stlaz: Here I go as promised.

As you can see, previously, we did not allow logging during the handling of the options. That's there for a good reason - some options are able to influence how and where to log. Therefore the only issue you're seeing in your PR https://github.com/freeipa/freeipa/pull/903 is nothing but the one you introduced yourself.

This is rather magical and is not documented anywhere. That's wrong. I did not realize that sooner either.

I would propose fixing https://github.com/freeipa/freeipa/pull/903 to print warning in runtime since we're not able to "fix" this properly (I don't think it's actually broken, it may only be way too magical).

felipevolpone: Hmm ok, now I understand your reasons. Ok, so I updated 903 and I'll close this as rejected, ok?

felipevolpone: Just bringing more info to the table (to future discussions) as was discussed on IRC:

The getLogger method will always return only one instance of the log, it will not create another object, as we can see in the docs [1]: "*All calls to this function with a given name return the same logger instance. This means that logger instances never need to be passed between different parts of an application.*" 

So, every call to `standard_logging_setup` would reinitialize the config was already done, due to `root_logger.handlers = []`. This can lead to an overload, but it seems harmless IMHO. 

In the future, we can try to find the options are able to influence how and where to log (as standa said) and make them independent.

[1] https://docs.python.org/2/library/logging.html#logging.getLogger

stlaz:
Closing, thanks!

PR #939, opened 2017-07-27
Changing how commands handles error when it can't connect to IPA server

felipevolpone:
Checking if env has the server attribute. If it doesn't, it means that an IPA server was not configured.

https://pagure.io/freeipa/issue/6261

### DIFF
--- a/client/ipa-client-automount
+++ b/client/ipa-client-automount
@@ -46,6 +46,7 @@ from ipaclient.install import ipachangeconf, ipadiscovery
 from ipalib import api, errors
 from ipalib.install import sysrestore
 from ipalib.install.kinit import kinit_keytab
+from ipalib.util import check_client_configuration
 from ipapython import ipautil
 from ipapython.ipa_log_manager import standard_logging_setup
 from ipapython.dn import DN
@@ -53,6 +54,8 @@ from ipaplatform.constants import constants
 from ipaplatform.tasks import tasks
 from ipaplatform import services
 from ipaplatform.paths import paths
+from ipapython.admintool import ScriptError
+
 
 logger = logging.getLogger(os.path.basename(__file__))
 
@@ -376,11 +379,13 @@ def configure_nfs(fstore, statestore):
                      rpcgssd.service_name, str(e))
 
 def main():
+    try:
+        check_client_configuration()
+    except ScriptError as e:
+        sys.exit(e)
 
     fstore = sysrestore.FileStore(paths.IPA_CLIENT_SYSRESTORE)
     statestore = sysrestore.StateFile(paths.IPA_CLIENT_SYSRESTORE)
-    if not fstore.has_files() and not os.path.exists(paths.IPA_DEFAULT_CONF):
-        sys.exit('IPA client is not configured on this system.\n')
 
     options, _args = parse_options()
 
--- a/ipalib/cli.py
+++ b/ipalib/cli.py
@@ -55,7 +55,9 @@ from ipalib.constants import CLI_TAB, LDAP_GENERALIZED_TIME_FORMAT
 from ipalib.parameters import File, Str, Enum, Any, Flag
 from ipalib.text import _
 from ipalib import api  # pylint: disable=unused-import
+from ipalib.util import check_client_configuration
 from ipapython.dnsutil import DNSName
+from ipapython.admintool import ScriptError
 
 import datetime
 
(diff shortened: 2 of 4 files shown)


felipevolpone:
Old PR #880

flo-renaud:
Hi @felipevolpone,

Some of the client commands already perform similar checks to ensure IPA client is configured (for instance ipa-client-automount and ipa-certupdate), by checking if /etc/ipa/default.conf exists and fstore.has_files(). It would be better to keep the consistency and use the same checks.

You could create a method check_client_configuration() that would be called by all the tools needing to check that ipa client is configured.

ipalib/cli.py
```diff
@@ -1347,9 +1348,7 @@ def run(api):
     try:
         (_options, argv) = api.bootstrap_with_global_options(context='cli')
 
-        if not hasattr(api.env, 'server'):
-            logger.error('IPA is not configured on this system.')
-            return
+        check_client_configuration()
```
felipevolpone: Hi @flo-renaud,

I've changed the code to use the same check that ipa-client-automount does, but then the ipa help commands does not work anymore.

flo-renaud:
Hi @felipevolpone 
I believe it's expected that ipa help fails if IPA is not configured on the system, because with the introduction of thin client, the client must fetch the schema from the server to build the help content. So if the system is not configured, there is no server defined and it is not possible to fetch the schema.
In this case I think it's OK to print that IPA client is not configured, even for ipa help commands.
Note that ipa --help still works.

felipevolpone:
@flo-renaud ah nice, I didn't know that.
So, I've done the changes you proposed and also changed the scripts were doing the check.

ipalib/util.py
```diff
@@ -1069,6 +1072,15 @@ def ensure_krbcanonicalname_set(ldap, entry_attrs):
     entry_attrs.update(old_entry)
 
 
+def check_client_configuration():
+    """
+    Check if IPA client is configured on the system.
+    """
+    fstore = sysrestore.FileStore(paths.IPA_CLIENT_SYSRESTORE)
+    if not fstore.has_files() and not os.path.exists(paths.IPA_DEFAULT_CONF):
+        sys.exit('IPA client is not configured on this system.\n')
```
flo-renaud: It would be better to use raise admintools.ScriptError instead of sys.exit and handle the exception when needed.

felipevolpone: Thanks for pointing that :) Done

tomaskrizek:
@felipevolpone please re-base for PR CI tests

client/ipa-client-automount
```diff
@@ -376,11 +377,10 @@ def configure_nfs(fstore, statestore):
                      rpcgssd.service_name, str(e))
 
 def main():
+    check_client_configuration()
```
flo-renaud: Here you need to handle the exception with
```
try:
  check_client_configuration()
except ScriptError as e:
  sys.exit(e)
```

otherwise the tool exits with a traceback.

ipalib/util.py
```diff
@@ -1069,6 +1074,15 @@ def ensure_krbcanonicalname_set(ldap, entry_attrs):
     entry_attrs.update(old_entry)
 
 
+def check_client_configuration():
+    """
+    Check if IPA client is configured on the system.
+    """
+    fstore = sysrestore.FileStore(paths.IPA_CLIENT_SYSRESTORE)
+    if not fstore.has_files() and not os.path.exists(paths.IPA_DEFAULT_CONF):
+        raise ScriptError('IPA client is not configured on this system.\n')
```
flo-renaud: You can remove the \n

ipalib/cli.py
```diff
@@ -1346,6 +1347,9 @@ def run(api):
     error = None
     try:
         (_options, argv) = api.bootstrap_with_global_options(context='cli')
+
+        check_client_configuration()
+
```
flo-renaud: Here also you should handle the exception and call sys.exit(e) otherwise the cli prints the traceback.

flo-renaud:
Hi @felipevolpone 
Thank you for the patch, everything works as expected now.

tomaskrizek:
master:

* cac3475a0454b730d6e5b2093c2e63d395acd387 Changing how commands handles error when it can't connect to IPA server

tomaskrizek:
@felipevolpone Please backport this patch for `ipa-4-5`.

PR #940, opened 2017-07-28
replica: fix SetuptoolsVersion comparison

stlaz:
Python 3 does not allow comparing SetuptoolsVersion and str
instances.

-----------------------
Note that this does not allow full ipa-replica-install completion since that's now blocked by Custodia: https://bugzilla.redhat.com/show_bug.cgi?id=1476150


stlaz:
The tests passed and this is a one-liner which is easy to verify, thus ACKing it with one-liner rule.

pvomacka:
master:

* 76904ba84dce182ea003d0f88e66fc09778f51eb replica: fix SetuptoolsVersion comparison

PR #942, opened 2017-07-28
otptoken_yubikey.py: Removed traceback when package missing.

Tiboris:
IPA should suggest user to install dependent packages instead
of throwing traceback. To work with IPA and Yubikey, packages
libyubikey(not in official RHEL repo) and libusb are required.

Resolves: https://pagure.io/freeipa/issue/6979

### DIFF
--- a/ipaclient/plugins/otptoken_yubikey.py
+++ b/ipaclient/plugins/otptoken_yubikey.py
@@ -131,6 +131,9 @@ class otptoken_add_yubikey(Command):
             raise NotFound(reason="No YubiKey found: %s" % e.strerror)
         except yubico.yubikey.YubiKeyError as e:
             raise NotFound(reason=e.reason)
+        except ValueError as e:
+            raise NotFound(reason=str(e) + ". Please install 'libyubikey' "
+                           "and 'libusb' packages first.")
 
         assert yk.version_num() >= (2, 1)
 


flo-renaud:
Hi @Tiboris 

Thank you for your patch. I was able to reproduce the issue on RHEL 7.4 but not on fedora 26. On RHEL74, the issue happens when libusbx is missing (if I try to remove this package in fedora, dnf would also try to remove the protected package "dnf", this is why I could not reproduce on fedora).

Your fix properly fixes the issue.

flo-renaud:
Hi @Tiboris 

Thank you for your patch. I was able to reproduce the issue on RHEL 7.4 but not on fedora 26. On RHEL74, the issue happens when libusbx is missing (if I try to remove this package in fedora, dnf would also try to remove the protected package "dnf", this is why I could not reproduce on fedora).

Your fix properly fixes the issue.

Tiboris:
Hi @flo-renaud 
I was not able to reproduce this behavior on fedora 25+ as well, but anyway, Bugzilla related to this issue has been filled against RHEL and is hopefully properly fixed with this commit.
Thanks!

tomaskrizek:
master:

* 76357283ec522aef8aaf3d3b8eea7155263e3517 otptoken_yubikey.py: Removed traceback when package missing.

PR #945, opened 2017-07-31
DNS update: reduce timeout for CA records

MartinBasti:
Timeout 120 seconds is quite long and it makes uninstallation too long
for. Given that this is non critical operation and may be executed
manually later, waiting 120 seconds is too much. Usually waiting longer
will not help at all to resolve missing record.

30 seconds is long enough 🕯

### DIFF
--- a/ipaserver/dns_data_management.py
+++ b/ipaserver/dns_data_management.py
@@ -55,6 +55,8 @@ IPA_DEFAULT_NTP_SRV_REC = (
     (DNSName("_ntp._udp"), 123),
 )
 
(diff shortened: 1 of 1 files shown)


tomaskrizek:
This should be linked with https://pagure.io/freeipa/issue/6176

stlaz:
Well, link it during push, I doubt Martin Basti is ever going to do that now.

stlaz:
master:

* dffddbd2c0dae7d646bdb2be23ac5ebe46f19d7a DNS update: reduce timeout for CA records

PR #947, opened 2017-08-01
test_caless: add SAN dNSName extensions for wildcard tests

Rezney:
It may happen that FQDN does not match with the domain mapped to
the host. In this case we add wildcard for both domains.

### DIFF
--- a/ipatests/test_integration/create_caless_pki.py
+++ b/ipatests/test_integration/create_caless_pki.py
@@ -162,7 +162,7 @@ def profile_ca(builder, ca_nick, ca):
 
 def profile_server(builder, ca_nick, ca,
                    warp=datetime.timedelta(days=0), dns_name=None,
-                   badusage=False):
+                   badusage=False, wildcard=False):
     now = datetime.datetime.utcnow() + warp
 
     builder = builder.not_valid_before(now)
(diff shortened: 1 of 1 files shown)


Rezney:
E.g. in a lab environment there may be hostname vm-007.ipa.test and domain dom-007.ipa.test.

ipatests/test_integration/create_caless_pki.py
```diff
@@ -204,6 +204,21 @@ def profile_server(builder, ca_nick, ca,
...
+                     x509.DNSName(u'*.' + domain),
+                     x509.DNSName(u'*.' + server_domain),
+                ]),
+                critical=False,
+            )
+        else:
+            builder = builder.add_extension(
+                x509.SubjectAlternativeName([x509.DNSName(u'*.' + domain)]),
+                critical=False,
+            )
```
frasertweedale: I think the logic is fine, but a few comments:

1. partially build the GeneralName list, then append to it if the condition matches (eliminate duplicate code)
2. if for some reason there is no `.` in `domain`, we don't want to crash when indexing the split result at `[1]`.

So, something like:

```python
if wildcard:
  names = [x509.DNSName(u'*.' + domain)]
  server_split = server1.split('.', 1)
  if len(server_split) == 2 and domain != server_split[1]
    names.append(x509.DNSName(u'*.' + server_split[1])
  builder = builder.add_extension(x509.SubjectAlternativeName(names), critical=False)
```

frasertweedale:
Logic looks fine, code looks fine.  I haven't tested it but it's obviously not getting
in the way or regressing anything else, so ACK.

tomaskrizek:
master:

* a3c99367bfe1071073cd93237660d783459b25e2 test_caless: add SAN dNSName extensions for wildcard tests

PR #948, opened 2017-08-01
Test script for ipa-custodia

tiran:
You may find my test script for ipa-custodia useful for testing and debugging issues like https://bugzilla.redhat.com/show_bug.cgi?id=1476150

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -933,6 +933,7 @@ install/tools/ipa-cacert-manage
 install/tools/ipa-compat-manage
 install/tools/ipa-csreplica-manage
 install/tools/ipa-custodia
+install/tools/ipa-custodia-check
 install/tools/ipa-dns-install
 install/tools/ipa-httpd-kdcproxy
 install/tools/ipa-kra-install
@@ -1322,6 +1323,7 @@ fi
 %{_libexecdir}/certmonger/ipa-server-guard
 %dir %{_libexecdir}/ipa
 %{_libexecdir}/ipa/ipa-custodia
+%{_libexecdir}/ipa/ipa-custodia-check
 %{_libexecdir}/ipa/ipa-dnskeysyncd
 %{_libexecdir}/ipa/ipa-dnskeysync-replica
 %{_libexecdir}/ipa/ipa-ods-exporter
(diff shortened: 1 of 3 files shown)


tiran:
@stlaz suggested to add the script to freeipa tools rather than freeipa's contrib directory. I'm happy with either.

stlaz:
Looks good to me. Perhaps we may keep it here. If you please rebase the PR on current master, @tiran, so that the tests pass and we can push it, that's be nice :)

Tiboris:
Hi @tiran,

rebase to master could help pr-ci tests to pass. Could you please try it?

flo-renaud:
Hi @tiran,

Thank you for this tool!

When ipa-custodia-check is run with target server=replica which does not have the CA installed, it reports errors trying to retrieve ca/* keys. To me this should not be considered as an error because the replica does not store the keys. The tool could add a check on the server role "CA" and try to retrieve ca/* keys only if CA role is present, or turn the error into a warning.

An error also happens if run with --store from a replica without CA with target server = CA, because the tool needs /etc/pki/pki-tomcat/password.conf in order to store the certs, and the directory /etc/pki/pki-tomcat does not exist if the replica doesn't have the CA role.

tiran:
By default the test script attempts to retrieve all possible secrets. IMO it's the correct answer, too. After all the remote server is not able to offer these secrets and therefore cannot be used to create a CA replica. You can limit the keys with command line options, e.g. ``ipa-custodia-check dm/DMHash``.

The store option is rather dangerous. I've changed the script to suppress the option. It's still there but ``--help`` no longer lists the argument, ``--store`` should only be used if you really know what you are doing.

rcritten:
It doesn't handle running on an unconfigured server:

```
# ipa-custodia-check `hostname`
  File "/home/rcrit/ipa-custodia-check", line 282, in <module>
    main()
  File "/home/rcrit/ipa-custodia-check", line 276, in main
    tester = IPACustodiaTester(parser, args)
  File "/home/rcrit/ipa-custodia-check", line 105, in __init__
    self.realm = api.env.realm
AttributeError: 'Env' object has no attribute 'realm'
```

tiran:
I've added a check for ``is_ipa_configured``.

rcritten:
This looks ok. Before I git it the ack is there a reason you are installing into libexec rather than bin/sbin?

I'm assuming it is because it is a diagnostic and not meant for regular use but just want to confirm first.

tiran:
Yes, you are correct. I don't want to have a diagnostic tool in sbin. It's not useful for a general audience. The script is designed for rare occasion of a failing replica installation.

tiran:
master:

* 38b17e1c792312a7fa10bf13601ff2614ff039ad Test script for ipa-custodia

PR #949, opened 2017-08-01
host, service: fix adding host/svc with a cert

stlaz:
ipaldap.LDAPEntry expects that entry attributes, if multi-valued,
are lists.

The recent cert refactoring made it possible to pass certificate
values from options directly to LDAPEntry instance but options
store multi-values as tuples. This caused crashes since the
whole tuple was stuck in a list instead of the values of the
tuples to be extracted in the list.

https://pagure.io/freeipa/issue/7077

### DIFF
--- a/ipaserver/plugins/host.py
+++ b/ipaserver/plugins/host.py
@@ -691,7 +691,6 @@ class host_add(LDAPCreate):
             # save the password so it can be displayed in post_callback
             setattr(context, 'randompassword', entry_attrs['userpassword'])
 
-        entry_attrs['usercertificate'] = options.get('usercertificate', [])
         entry_attrs['managedby'] = dn
         entry_attrs['objectclass'].append('ieee802device')
         entry_attrs['objectclass'].append('ipasshhost')
--- a/ipaserver/plugins/service.py
+++ b/ipaserver/plugins/service.py
@@ -623,8 +623,6 @@ class service_add(LDAPCreate):
 
         self.obj.validate_ipakrbauthzdata(entry_attrs)
 
-        entry_attrs['usercertificate'] = options.get('usercertificate', [])
-
         if not options.get('force', False):
             # We know the host exists if we've gotten this far but we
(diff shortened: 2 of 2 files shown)


ipaserver/plugins/host.py
```diff
@@ -691,7 +691,8 @@ def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys, **options):
             # save the password so it can be displayed in post_callback
             setattr(context, 'randompassword', entry_attrs['userpassword'])
 
-        entry_attrs['usercertificate'] = options.get('usercertificate', [])
+        entry_attrs['usercertificate'] = list(
+            options.get('usercertificate', []))
```
flo-renaud: unless I'm mistaken, this line is not needed at all. The entry_attrs already contains the usercertificate attribute as a list (because the class inherits from LDAPCreate).

ipaserver/plugins/service.py
```diff
@@ -623,7 +623,8 @@ def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys, **options):
 
         self.obj.validate_ipakrbauthzdata(entry_attrs)
 
-        entry_attrs['usercertificate'] = options.get('usercertificate', [])
+        entry_attrs['usercertificate'] = list(
+            options.get('usercertificate', []))
```
flo-renaud: same comment as above.

ipaserver/plugins/host.py
stlaz: That is a great comment, I haven't realized this. Thanks!

flo-renaud:
Hi @stlaz 
Thank you! The code looks good to me, and the CLI works as expected. Let's wait for PR-CI status.

flo-renaud:
As the checks have passes, acking this PR

tomaskrizek:
master:

* 7d217c8c9b6f9703b46a42ecbd55b4ca32868795 host, service: fix adding host/svc with a cert

PR #951, opened 2017-08-02
server plugin: pass bytes to ldap.modify_s

stlaz:
The server-del command passes str instance instead of bytes to
ldap.modify_s which results in the target server not being
removed properly.


felipevolpone:
LGTM

pvomacka:
master:

* d147948fccd7fda1909a809a691f86722cb6de6b server plugin: pass bytes to ldap.modify_s

PR #952, opened 2017-08-02
tasks: remove str on bytes

stlaz:
This enables ipa-certupdate to be run in Python 3


stlaz:
I forgot I already fixed this in another PR 🙄

PR #953, opened 2017-08-02
[master] WebUI: Turn on pagination on certificate page

pvomacka:
Almost all other search tables use pagination. Only this one does not.
This change makes WebUI more consistent.

https://pagure.io/freeipa/issue/6079

### DIFF
--- a/install/ui/src/freeipa/certificate.js
+++ b/install/ui/src/freeipa/certificate.js
@@ -1647,7 +1647,11 @@ return {
             facet_groups: [exp.facet_group],
             facet_group: 'certificates',
             additional_navigation_arguments: [ 'cacn' ],
-            pagination: false,
+            command_options: { all: true },
+            dont_call_show_method: true,
+            search_all_entries: true,
+            sort_enabled: false,
+            show_values_with_dup_key: true,
             no_update: true,
             columns: [
                 {
--- a/install/ui/src/freeipa/facet.js
+++ b/install/ui/src/freeipa/facet.js
@@ -2180,7 +2180,7 @@ exp.table_facet = IPA.table_facet = function(spec, no_init) {
             return;
         }
 
-        if (that.search_all_entries) {
+        if (that.search_all_entries && !that.pagination) {
             // map contains the primary keys and the complete records
             that.load_records(records_map.values);
             return;


frasertweedale:
It seems really slow, but I guess the pagination actually improves it... and it all works as
expected so ACK.

pvomacka:
@frasertweedale yes, that's true, that it is really slow.. I can try to make it better somehow.

pvoborni:
What is the performance impact? cert-find  does 3 internal searches. With sizelimit=0 and --all which getting cert from Dogtag by ra. get_certificate(serial_num) call. This might mean thousands of calls.

With what number of certs it was tried?

frasertweedale:
@proborni it was quite slow with 120 certs, but not unbearable, and limited now by the fact that pagination is 20 at a time.  I don't think it's a regression; just something I observed during testing.

frasertweedale:
Does this need a rebase to make the CI happy?

Modulo tests passing, I'm happy to put the ACK back because I don't think the performance
problem is a regression.

stlaz:
Hm, looks like it does, I'll try to rebase it for Pavel.

stlaz:
More about the potential performance regression:
There's actually only two internal searches, not three (there's one which looks like a search but in fact, it's not). One of the searches is at CA, one is in LDAP, both are performed unconditionally with no limits in size and time.
For these reasons, it would make sense to cache certificates even for paginated pages, but then again, that would probably mean to make a special exception just for this one case 🙄

pvoborni:
Fraser, you might be right, but the argument is not very convincing. Some environments have thousands of certificates in DogTag. Doing performance decisions based on personal feeling with 200 cert is a flaky statement. 

What is the complexity of the operation? 

I find code of this command poorly documented/commented.  Which makes it hard to understand what individual subsearch does or why they are there. But let's look at it.

When looking at cert_find command. I see there iteration with 3 types of searches:

```python
for sub_search in (self._cert_search,
                   self._ca_search,
                   self._ldap_search):
    sub_result, sub_truncated, sub_complete = sub_search(
        all=all,
        raw=raw,
        pkey_only=pkey_only,
        no_members=no_members,
        **options)
```
* `_ldap_search` seems to do 1 LDAP search, get matching certificates, process it and add it to result list.
* `_ca_search` gets Sub CAs' certs?  i.e. also one LDAP search
* `_ cert_search` parses cert and returns it. 

OK, sot not a big issue - max 2 searches so far  as Standa wrote.

But here, we can see that with `all` option, for every IPA issued cert we do an additional API call to dogtag. Which changes complexity form O(1) at least to O(n).

So with combination of `sizelimit=0` and `all=true` will do at least 2000 ldap searches  - probably more because I have a hunch that `ra.get_certificate(str(serial_number)` does more than one search.

```
if not pkey_only:
    ca_objs = {}
    ra = self.api.Backend.ra

    for key, obj in six.iteritems(result):
        if all and 'cacn' in obj:
            _issuer, serial_number = key
            cacn = obj['cacn']

            try:
                ca_obj = ca_objs[cacn]
            except KeyError:
                ca_obj = ca_objs[cacn] = (
                    self.api.Command.ca_show(cacn, all=True)['result'])

            obj.update(ra.get_certificate(str(serial_number)))
```

This is a hard NACK unless I'm wrong.

pvomacka:
@pvoborni Thank you for investigating how exactly searches work. In that case we should close this PR and close the ticket (or at least move it into Future Releases with a comment) until performance improvements in cert-find (or proper server-side pagination) are implemented.

frasertweedale:
I'm not going to die on this hill, but if this change is a improvement (to UI/UX), without
introducing regressions (performance or other), why not merge it?

The existence of performance issues with cert-find is indeed a problem
(I filed issue: https://pagure.io/freeipa/issue/7178); but IMO it is a separate
problem, which is not made worse by (possibly even mitigated by) pagination.

pvoborni:
lines
```
command_options: { all: true },
search_all_entries: true,
```
triggers: the performance degradation conditions `sizelimit=0` and `all=true` so I'd say it is made worse by this patch.

frasertweedale:
@pvoborni right; I had overlooked that previously `all` was not given.

But I'm not sure I see a reason why it needed to be given... I don't see any other change
that would seem to need the extra attributes returned by 'all'.

Maybe I am missing something (I am not familiar with WebUI implementation so it's quite likely!)

tiran:
@pvomacka Please rebase the PR to master to kick off another CI run.

pvoborni:
When I find some time I'll look what if we can change the patch not to use --all.

pvoborni:
Tried variant without `--all`. It doesn't work. 

But IMO it should be possible to make it work if we change cert_find plugin a bit. 

I also think that there is a bug in cert_find plugin. I had a test setup where a user has a user certificate from some other CA with serial number matching serial number of a cert issued by other IPA CA - in this case it was #5. Instead of returning 2 distinct records it returns IPA cert with serial num 5 twice.

pvoborni:
I don't have time to work on both parts anytime soon. Pavel won't update it. So I'm closing the PR. It was referenced in the ticket for future purposes. https://pagure.io/freeipa/issue/6079

PR #954, opened 2017-08-02
py3 conncheck + service plugin fixes

stlaz:
commit 09942d0268f02ed15df5f7f3aad3220196c5a41c (HEAD -> py3-conncheck, private/py3-conncheck)
Author: Stanislav Laznicka <slaznick@redhat.com>
Date:   Wed Aug 2 16:05:16 2017 +0200

    conncheck: fix progression on failure
    
    traceback.format_exc() does not take exception object as an argument.
    This made Python 3 get stuck amid ipa-replica-conncheck, probably
    because it was waiting for a thread to finish.
    
commit fe820cbc1f3469150ab90af401b119d5d316f3ab
Author: Stanislav Laznicka <slaznick@redhat.com>
Date:   Wed Aug 2 15:59:39 2017 +0200

    kerberos: fix sorting Principal objects
    
    When service-find was issued under Python 3, the command fails
    because it tried to sort a list of Principal objects which was not
    possible.
    
### DIFF
--- a/install/tools/ipa-replica-conncheck
+++ b/install/tools/ipa-replica-conncheck
@@ -346,9 +346,9 @@ class PortResponder(threading.Thread):
                 sock.listen(1)
 
             logger.debug('%d %s: Started listening', port, proto)
-        except socket.error as e:
+        except socket.error:
             logger.warning('%d %s: Failed to bind', port, proto)
-            logger.debug("%s", traceback.format_exc(e))
+            logger.debug("%s", traceback.format_exc())
         else:
             self._sockets.append(sock)
 
--- a/ipapython/kerberos.py
+++ b/ipapython/kerberos.py
@@ -93,6 +93,18 @@ class Principal(object):
     def __ne__(self, other):
         return not self.__eq__(other)
 
+    def __lt__(self, other):
+        return unicode(self) < unicode(other)
+
+    def __le__(self, other):
+        return self.__lt__(other) or self.__eq__(other)
+
+    def __gt__(self, other):
+        return not self.__le__(other)
+
+    def __ge__(self, other):
+        return self.__gt__(other) or self.__eq__(other)
+
     def __hash__(self):
         return hash(self.components + (self.realm,))
 


felipevolpone:
Hi @stlaz,

I was not able to reproduce the bug (that commit fe820cb fixes) using ipa master: 
```
[root@localhost freeipa]# ipa --version
VERSION: 4.5.90.dev201708052005+gitd147948, API_VERSION: 2.229
[root@localhost freeipa]# python3 /usr/bin/ipa service-find
------------------
5 services matched
------------------
  Principal name: ldap/example.test@EXAMPLE.TEST
  Principal alias: ldap/example.test@EXAMPLE.TEST
  ...
----------------------------
Number of entries returned 5
----------------------------
```
It works even using `--principal` option.

Am I missing something?

stlaz:
Most probably it's your server that's not running in Python 3. Follow the steps I described on freeipa-devel just now and you should be able to reproduce it.

install/tools/ipa-replica-conncheck
```diff
@@ -346,9 +346,9 @@ class PortResponder(threading.Thread):
                 sock.listen(1)
 
             logger.debug('%d %s: Started listening', port, proto)
-        except socket.error as e:
+        except socket.error:
             logger.warning('%d %s: Failed to bind', port, proto)
-            logger.debug("%s", traceback.format_exc(e))
+            logger.debug("%s", traceback.format_exc())
```
felipevolpone: Hi standa,

I think we could use just `logger.debug(traceback.format_exc())` here. AFAIK there is no need to use `"%s"` in this case.
However, in the entire code base we're using `logger.debug("%s", traceback.format_exc())`, so even though I prefer the first option, it's up to you; just let me know.

felipevolpone:
Tested and it's works properly.

install/tools/ipa-replica-conncheck
stlaz: Since the entire code uses `%s`, I would just go with it. I suppose that it is actually how standard python logging should be used, although I am not that sure about that.

So I would just go with it.

stlaz:
@felipevolpone Do you require some changes or is it ok to push this, then?

felipevolpone:
ack

pvomacka:
master:

* dbeb41efd6ad2067f24a099f27c02cab88a8d84b kerberos: fix sorting Principal objects
* 31a5cf588e1171bcfd58d6a08a76867552dfe811 conncheck: fix progression on failure

PR #955, opened 2017-08-02
host_port_open: revert to old behavior where one iface is sufficient

pvoborni:
Commit a24cd01304aaef77b66d0e178585c9ec8bbce9b5

Changed behavior of host_port_open to require all discovered interfaces to
listed on the port.

But usage of host_port_open function in wait_for_open_ports function which is
indirectly used from service.start might be still ok with only one interface.

Requiring all interfaces might then cause issue(waiting till timeout) in IPA upgrader in specific DNS
or network setups.

https://pagure.io/freeipa/issue/7083

### DIFF
--- a/install/tools/ipa-replica-conncheck
+++ b/install/tools/ipa-replica-conncheck
@@ -380,7 +380,8 @@ def port_check(host, port_list):
         try:
             port_open = ipautil.host_port_open(
                 host, port.port, port.port_type,
-                socket_timeout=CONNECT_TIMEOUT, log_errors=True)
+                socket_timeout=CONNECT_TIMEOUT, log_errors=True,
+                check_all_ifaces=True)
         except socket.gaierror:
             raise RuntimeError("Port check failed! Unable to resolve host name '%s'" % host)
         if port_open:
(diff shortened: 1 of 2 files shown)


tomaskrizek:
I think we should implement the fix differently. The `host_port_open()` shouldn't be changed. It is correct that it checks for the open port on all resolved IPs. If a hostname is resolved to multiple IPs, it should be reachable on all of them.

The problem is the incorrect usage of `host_port_open()`, or rather `wait_for_open_ports()`. If you want to check only for internal ports, you can use `host='localhost'` to determine if it is reachable on all IPs the localhost resolves to (usually both IPv4 and IPv6).

In case you absolutely need to check just one specific IP, you can specify it in the check as well, e.g. `host=127.0.0.1`.

TL;DR `host_port_open()` behaves as it should, the usage is incorrect and should be fixed

pvoborni:
The call already uses host=`localhost`. What was wrong in the environment when the issue happened was that it was not listening on ::1 even though the address was resolved. This is an issue in env configuration of IPA host.  IMO we should put this info into a log to help with investigation if it happens.

PR #956, opened 2017-08-03
Log progress of wait_for_open_ports

pvoborni:
### control logging of host_port_open from caller

host_port_open copied logging behavior of ipa-replica-conncheck utility
which doesn't make it much reusable.

Now log level can be controlled from caller so other callers might use
other logging level without host_port_open guessing what was the
intention.

### log progress of wait_for_open_ports

To know what to focus on when some check fail. E.g. to detect that
IPv6 address or its resolution for localhost is misconfigured.

Also gradually increases wait time so that it won't log 600 same
messages when wait time is 600.

https://pagure.io/freeipa/issue/7083

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -1224,14 +1224,19 @@ def wait_for_open_ports(host, ports, timeout=0):
     op_timeout = time.time() + timeout
 
     for port in ports:
+        logger.debug('waiting for port: %s', port)
+        log_error = True
         while True:
-            port_open = host_port_open(host, port)
+            port_open = host_port_open(host, port, log_errors=log_error)
+            log_error = False  # Log only first err so that the log is readable
 
             if port_open:
+                logger.debug('SUCCESS: port: %s', port)
                 break
             if timeout and time.time() > op_timeout: # timeout exceeded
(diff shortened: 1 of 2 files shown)


ipapython/ipautil.py
```diff
@@ -986,19 +987,12 @@ def host_port_open(host, port, socket_type=socket.SOCK_STREAM,
...
                                          proto=PROTOCOL_NAMES[socket_type],
                                          addr=sa[0]))
-
-                # Do not log udp failures as errors (to be consistent with
-                # the rest of the code that checks for open ports)
-                if socket_type == socket.SOCK_DGRAM:
-                    logger.warning('%s', msg)
-                else:
-                    logger.error('%s', msg)
+                logger.log(log_level, '%s', msg)
```
tomaskrizek: `msg` is already a string, please use `logger.log(log_level, msg)`

tomaskrizek:
master:

* cc72db67e2eaede577c3129d572b85e9c2ba593c control logging of host_port_open from caller
* 038d1920657c6fd349f8414ed173a9c97681a602 log progress of wait_for_open_ports

tomaskrizek:
@pvoborni Please backport this to `ipa-4-5`.

PR #957, opened 2017-08-03
Pass ipa-ca-agent credentials as PEM files

rcritten:
In Fedora 27 curl is proposing to switch to using OpenSSL as
the crypto backend instead of NSS. This requires a new set of
arguments to certmonger to bootstrap fetching the IPA RA cert.

Debian/Ubuntu are still currently using an NSS-backed curl
so support both methods for now.

https://pagure.io/freeipa/issue/7076

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -462,9 +462,6 @@ class CAInstance(DogtagInstance):
         os.close(cfg_fd)
         pent = pwd.getpwnam(self.service_user)
         os.chown(cfg_file, pent.pw_uid, pent.pw_gid)
-        self.tmp_agent_db = tempfile.mkdtemp(
-                prefix="tmp-", dir=paths.VAR_LIB_IPA)
-        self.tmp_agent_pwd = ipautil.ipa_generate_password()
 
         # Create CA configuration
         config = RawConfigParser()
@@ -484,9 +481,6 @@ class CAInstance(DogtagInstance):
                 ipautil.format_netloc(api.env.domain)))
 
         # Client security database
-        config.set("CA", "pki_client_database_dir", self.tmp_agent_db)
-        config.set("CA", "pki_client_database_password", self.tmp_agent_pwd)
-        config.set("CA", "pki_client_database_purge", "False")
         config.set("CA", "pki_client_pkcs12_password", self.admin_password)
 
         # Administrator
@@ -614,8 +608,7 @@ class CAInstance(DogtagInstance):
                 self, cfg_file,
                 nolog_list=(self.dm_password,
                             self.admin_password,
-                            pki_pin,
-                            self.tmp_agent_pwd)
+                            pki_pin)
             )
         finally:
             os.remove(cfg_file)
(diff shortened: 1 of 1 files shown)


stlaz:
Why try to feed Certmonger with an NSS database? Why not to just extract the agent certificate from `/root/ca-agent.p12`, use them instead all the time and be done with it?

I did a similar thing about a month ago (extracted the certs from the NSS db and try to use them) but Certmonger still failed so I was guessing something needs to be changed there as well.

Point is - we've been using PEM files for authentication against dogtag with Certmonger for about a whole version now and there haven't been any problems doing that so I would just pass files to Certmonger here all the time and be done with it. Guessing `curl`'s crypto backend is something nobody should have to do.

stlaz:
Please note that ever since https://bugzilla.redhat.com/show_bug.cgi?id=1476150, it might be troublesome to get the keys from the `cacert.p12` file using the `openssl` tool, I did not find a way to do it yet.

stlaz:
For the record, this patch works, but I'd very much rather do it the "always pass the certificates" way as it seems more sane.

If we need this patch rather sooner, then instead of trying to extract the required keys from `/root/cacert.p12` we can get a .p12 from the NSS database with the content we need and which you actually already use. It would be possible to fix it later once we get the info from the Dogtag guys about how to extract what we need. I just hope I'll catch them before going on a PTO.

rcritten:
The cert I need is in ca-agent.p12 but it is also in a temporary NSS database at this point in the installation so I use that directly.

I kept the NSS code because in Fedora/RHEL we have the nss-pem module to handle PEM files in NSS_linked curl but not all distributions have that (Debian does not for sure).

I submitted the 4.5.0 patch first because this is where the breakage is. My goal was to get things running in F-27 first. I can abandon this and submit a new one against master if you'd prefer.

rcritten:
Updated patch to remove the PKCS#12 creation step since we already have ca-agent.p12.

stlaz:
I tried to test this today but forgot that you still got this PR against IPA 4.5. Could you please cherry-pick your commit on current master, push the result into this PR and change this PR to be against the `master` branch? You do the last step by using the `Edit` button on top of the page.
I know you probably know this, but the easiest way to do all this is to do the following:
```bash
git checkout master
git pull
git checkout -b openssl_master
git cherry-pick openssl
git push --force <PRIVATE_REMOTE> openssl_master:openssl
```
and then you click `Edit` on top of the page and choose `master` as a branch to push this PR into.

The `git pull` on FreeIPA master will also allow our PRCI to perform its checks against your PR so that we fully follow our current review process.

In the meantime, I will test this cherry-picked on our current master so that we can speed the review up should all other tests pass.

stlaz:
With the changes I described in the above post, `ipa-server-install` passes on rawhide.

stlaz:
Hm, I would really need to see PRCI and travis outputs so that we can be sure we don't break those. I know Travis was actually passing, @tomaskrizek, is there a way to run PRCI on this PR?

stlaz:
I am giving this a conditional ACK, if the tests pass, it's OK to be pushed.

tomaskrizek:
It seems like Travis CI wasn't executed on this PR, I'm not sure why.

stlaz:
It's because of the branch change. Since we rebased prior to it, Travis would not run as there were merge conflicts. After the branch change, Travis would not start because apparently that's a change not good enough for it (or so it seems).
For future generations it might be worth noting that changing branch of the PR should be done prior to the rebase, which I did not realize.
Anyway, Travis was previously passing and this PR does not affect the plugins much, it only should affect the installation, and that's passing, too. My ACK therefore stands.

tomaskrizek:
master:

* 9c1ab3ca5015317091f40ac8c352823a75849cef Pass ipa-ca-agent credentials as PEM files

PR #959, opened 2017-08-07
schema cache: Generate help after fetching schema from server

davidkupka:
When schema is not in cache and is fetched from server help must be generated
otherwise `ipa help` call fails with traceback.

https://pagure.io/freeipa/issue/7092


flo-renaud:
Hi @pvomacka, @davidkupka 
Isn't this issue a duplicate of the issue [6999](https://pagure.io/freeipa/issue/6999) fixed by [PR 701](https://github.com/freeipa/freeipa/pull/701)? It seems the fix wasn't backported to rhel 7.4, but is present on ipa master branch.

davidkupka:
@flo-renaud you're right, thank you!

PR #961, opened 2017-08-09
Vault testcase improvement

Akasurde:
* Collect logs for install KRA

fixes: https://pagure.io/freeipa/issue/7098


tomaskrizek:
The KRA log is properly collected now, the directory structure looks like [this](https://fedorapeople.org/groups/freeipa/prci/jobs/aceb34ba-7db2-11e7-b23e-001a4a2316f2/), which is not ideal, because some logs are duplicated, but I don't see any other good way to address this with the current testing framework.

@Akasurde Please link the ticket with the entire URL in the commit message, since the issue isn't on github and wouldn't be linked properly. Also, I'd use keyword Related instead of Fixed, because the ticket isn't resolved yet and should be kept open.

tomaskrizek:
master:

* 8d3924dc98ba5a3873ae21c3c60de97510eebc88 Vault testcase improvement

PR #962, opened 2017-08-09
tests: fix external_ca test suite failing due to missing SKI

Rezney:
external_ca test suite is failing during installation due to
missing SKI extension.

https://pagure.io/freeipa/issue/7099

### DIFF
--- a/ipatests/test_integration/test_external_ca.py
+++ b/ipatests/test_integration/test_external_ca.py
@@ -17,9 +17,14 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import os
+import base64
 
 from ipatests.pytest_plugins.integration import tasks
 from ipatests.test_integration.base import IntegrationTest
+
+
+EXTERNAL_CA_KEY_ID = base64.b64encode(os.urandom(64))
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_external_ca.py
```diff
@@ -17,11 +17,16 @@
...
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import os
+import base64
 
 from ipatests.pytest_plugins.integration import tasks
 from ipatests.test_integration.base import IntegrationTest
 
 
+AKI_ID = base64.b64encode(os.urandom(64))
+SKI_ID = base64.b64encode(os.urandom(64))
```
frasertweedale: I think some clearer names would be helpful to people because e.g. we have
`AKI_ID` being used as a subject key ID below (for the external CA).

Perhaps `EXTERNAL_CA_KEY_ID` and `IPA_CA_KEY_ID` would be better names?

tomaskrizek:
master:

* 4caabb140e31218c72846877da0c985c6a178d7a tests: fix external_ca test suite failing due to missing SKI

ipatests/test_integration/test_external_ca.py
Rezney: Makes sense, fixed.

PR #964, opened 2017-08-10
Restore old version of caIPAserviceCert for upgrade only

frasertweedale:
The latest version of caIPAserviceCert profile includes a feature
that is not available before Dogtag 10.4, and this version of the
profile is intended for new installs only (otherwise, problems will
arise in topologies containing CA replicas at an earlier version).
But IPA versions before v4.2 did not use LDAP-based profiles, so the
new version of the profile gets imported when upgrading from
pre-v4.2 to v4.5 or later.

We do not yet have a proper version- and topology-aware profile
update mechanism, so to resolve this issue, ship the older version
of the profile alongside the newer version, and make sure we use the
older version when importing the profile in an upgrade context.

https://pagure.io/freeipa/issue/7097


flo-renaud:
Hi @frasertweedale 
Thank you for the patch. I tested with an upgrade from rhel 7.1 to rhel 7.4, ensured that ipa-server-upgrade completed successfully and that the caIPAserviceCert profile does not contain the constraint commonNameToSANDefaultImpl. Works as expected!

pvomacka:
master:

* 79955189217fec328f2d561a4a1a23ddb29eac44 Restore old version of caIPAserviceCert for upgrade only

PR #965, opened 2017-08-10
[ipa-4-5] Restore old version of caIPAserviceCert for upgrade only

frasertweedale:
The latest version of caIPAserviceCert profile includes a feature
that is not available before Dogtag 10.4, and this version of the
profile is intended for new installs only (otherwise, problems will
arise in topologies containing CA replicas at an earlier version).
But IPA versions before v4.2 did not use LDAP-based profiles, so the
new version of the profile gets imported when upgrading from
pre-v4.2 to v4.5 or later.

We do not yet have a proper version- and topology-aware profile
update mechanism, so to resolve this issue, ship the older version
of the profile alongside the newer version, and make sure we use the
older version when importing the profile in an upgrade context.

https://pagure.io/freeipa/issue/7097


pvomacka:
ipa-4-5:

* 87393daba6b414e3afe6e22e77c9b20e561e5302 Restore old version of caIPAserviceCert for upgrade only

PR #970, opened 2017-08-11
[ipa-4-5] backport #6999 to ipa-4-5

pvoborni:
Such simple fix should be in stabilization release.

https://pagure.io/freeipa/issue/7092 or https://pagure.io/freeipa/issue/6999 one needs to be opened on push.

### DIFF
--- a/ipaclient/remote_plugins/schema.py
+++ b/ipaclient/remote_plugins/schema.py
@@ -516,7 +516,9 @@ class Schema(object):
 
     def get_help(self, namespace, member):
         if isinstance(self._help, bytes):
-            self._help = json.loads(self._help.decode('utf-8'))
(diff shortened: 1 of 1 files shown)


pvoborni:
Previous PR: #701

pvoborni:
@rcritten I see that you "approved these changes" which I would understand as an ACK. Was it an ACK?

rcritten:
Yes, ACK

pvomacka:
ipa-4-5:

* 9308e63fde736ea4274a280f18c83a4bc2935c81 Store help in Schema before writing to disk
* 7352b731169e4f995b3ffe254b9e0cae6b82a763 Disable pylint in get_help function because of type confusion.

PR #972, opened 2017-08-14
build: checkout *.po files at the end of makerpms.sh

tomaskrizek:
*.po are re-generated during build. These changes shouldn't be
comitted to git unless translation have been updated (during
release).

Fixes https://pagure.io/freeipa/issue/6605

### DIFF
--- a/makerpms.sh
+++ b/makerpms.sh
@@ -26,4 +26,8 @@ test ! -f "Makefile" && ./configure --enable-silent-rules \
 	"$@"
 make rpms
 
+# Workaround to ignore re-generated *.po files in git repo
+# See https://pagure.io/freeipa/issue/6605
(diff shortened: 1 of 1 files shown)


pvomacka:
As @stlaz mentioned this won't work in case that someone would download tarball and then run ./makerpms.sh, but as the script is our tool for building rpms during development I would live with that. So, ACK. If someone will face this issue in the future, we can add check whether we are in Git repo or not later.

pvomacka:
master:

* f32784f4cb0c89ceae95f5bbc7c7224ed139aeb0 build: checkout *.po files at the end of makerpms.sh

pvomacka:
@tomaskrizek please rebase this commit on ipa-4-5 branch and create separate PR.

PR #977, opened 2017-08-15
WebUI: fix incorrectly shown links in association tables

pvomacka:
Previously all columns was shown as links, that was caused by setting
link attribute of each column to true. This true value was there because
of possibility to turn off links in whole table on self-service pages.

Now only column which is primary key is set to be shown as link.

https://pagure.io/freeipa/issue/7066


felipevolpone:
Tested and it works properly. LGTM

pvomacka:
master:

* ed7de96648daf95868f34107890700124836b69d WebUI: fix incorrectly shown links in association tables

PR #978, opened 2017-08-16
Fixing and porting dogtag-ipa-ca-renew-agent-submit

stlaz:
This PR fixes several issues with the `dogtag-ipa-ca-renew-agent-submit` script that were introduced in the recent certificate refactoring and eventually ports the script and its scriptlets to Python 3.

https://pagure.io/freeipa/issue/4985

To properly test the port, you will need https://github.com/freeipa/freeipa/pull/975

PR #979, opened 2017-08-16
cacert_manage: fix CA cert renewal

stlaz:
Since IPACertificate object is passed to CAInstance's
update_people_entry() and update_authority_entry(),
these are not be able to load this object as a certificate again and
thus would fail. This commit fixes that.

Depends on https://github.com/freeipa/freeipa/pull/978

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -1323,23 +1323,23 @@ def backup_config():
             "Dogtag must be stopped when creating backup of %s" % path)
     shutil.copy(path, path + '.ipabkp')
 
-def __update_entry_from_cert(make_filter, make_entry, dercert):
+def __update_entry_from_cert(make_filter, make_entry, cert):
     """
     Given a certificate and functions to make a filter based on the
     cert, and make a new entry based on the cert, update database
     accordingly.
 
-    ``make_filter``
+    :param make_filter:
         function that takes a certificate in DER format and
         returns an LDAP search filter
 
-    ``make_entry``
+    :param make_entry:
         function that takes a certificate in DER format and an
         LDAP entry, and returns the new state of the LDAP entry.
         Return the input unchanged to skip an entry.
 
-    ``dercert``
-        An X509.3 certificate in DER format
+    :param cert:
+        An IPACertificate object
 
     Logging is done via syslog.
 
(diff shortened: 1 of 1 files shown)


ipaserver/install/cainstance.py
```diff
@@ -1421,37 +1420,35 @@ def make_filter(dercert):
...
             ldap2.ldap2.MATCH_ALL)
 
-    def make_entry(dercert, entry):
-        cert = x509.load_der_x509_certificate(dercert)
+    def make_entry(cert, entry):
         serial_number = cert.serial_number
         subject = DN(cert.subject)
         issuer = DN(cert.issuer)
-        entry['usercertificate'].append(dercert)
+        entry['usercertificate'].append(cert)
```
flo-renaud: Here cert is a IPACertificate object, shouldn't you use instead cert.public_bytes(x509.Encoding.DER)?

ipaserver/install/cainstance.py
```diff
@@ -1421,37 +1420,35 @@ def make_filter(dercert):
...
         subject = str(DN(cert.subject))
         return ldap2.ldap2.make_filter(
             dict(objectclass='authority', authoritydn=subject),
             rules=ldap2.ldap2.MATCH_ALL,
         )
 
-    def make_entry(dercert, entry):
-        cert = x509.load_der_x509_certificate(dercert)
+    def make_entry(cert, entry):
+        cert = x509.load_der_x509_certificate(cert)
```
stlaz: Thank you for this catch, I must have left this here by an accident

ipaserver/install/cainstance.py
stlaz: ever since 5a44ca638310913ab6b0c239374f4b0ddeeedeb3, `ipaldap` should be capable of dealing with the `cryptography.x509.Certificate` objects as well :)

flo-renaud: OK, thank you for the explanation.

flo-renaud:
Hi @stlaz 
there is a PEP8 complaint:
./ipaserver/install/cainstance.py:1343:1: E302 expected 2 blank lines, found 1

but apart from that, it looks good to me.

pvomacka:
master:

* 6fc7a96e11b3f5de88f971478dc9365bf91c069e cacert_manage: fix CA cert renewal

PR #981, opened 2017-08-17
[ipa-4-5] WebUI: fix incorrectly shown links in association tables

pvomacka:
Previously all columns was shown as links, that was caused by setting
link attribute of each column to true. This true value was there because
of possibility to turn off links in whole table on self-service pages.

Now only column which is primary key is set to be shown as link.

https://pagure.io/freeipa/issue/7066


felipevolpone:
It works and LGTM

tomaskrizek:
ipa-4-5:

* 705351c27092292d0ddc6c099bf060f68a3c8616 WebUI: fix incorrectly shown links in association tables

PR #983, opened 2017-08-17
makerpms.sh: make git checkout optional

tomaskrizek:
In case someone is using the script from tarball, outside of git,
the git checkout command shouldn't fail the script.

Related https://pagure.io/freeipa/issue/6605


tomaskrizek:
master:

* f7d1a10f22548fb2c1eba99a3001c6da49605edd makerpms.sh: make git checkout optional

PR #984, opened 2017-08-17
ipautil: port host_port_open() to python 3

stlaz:
socket.send() expects `Bytes` instance, not string

This causes issues with `ipa-replica-conncheck` when installing a replica.


stlaz:
Nobody cared for a review, the tests pass and the fix is a simple enough one-liner. ACK.

tomaskrizek:
master:

* b1fbdbefdb201fc036e41e84a25e3e0725d8e3d2 ipautil: port host_port_open() to python 3

PR #985, opened 2017-08-18
py3: fix caless to CA promotion on replica

stlaz:

---------

Encountered this problem when trying to promote CA-less replica to CA-full, may happen on CA-less master as well although I did not investigate that.


pvomacka:
master:

* 2151ab02c19d0a74d90d3784a1ac1b14877a7b96 py3: fix caless to CA promotion on replica

PR #986, opened 2017-08-18
py3: Depend on newer pyldap for server-upgrade

stlaz:


abbra:
LGTM, let's see how builders react.

stlaz:
Modified to be `Requires(pre)` dependencies as @abbra pointed out since we need the deps for our upgrade run in the upgrade scriptlets.
It would seem that our PRCI machine run out of space, unfortunately ☹️

tomaskrizek:
master:

* 7bf6eb7e19ecdb24be185ab2a4d0a549029b9cf0 py3: Depend on newer pyldap for server-upgrade

PR #989, opened 2017-08-18
Removing part of circular dependency of ipalib in ipaplatform

felipevolpone:
After commit  cac3475a0454b730d6e5b2093c2e63d395acd387, ipa-backup is broken due to circular dependencies. This fixes it, removing circular dependency of ipalib. The ipalib.constants.IPAAPI_USER is now passed as parameter to the functions that use it.

https://pagure.io/freeipa/issue/7108

### DIFF
--- a/ipaplatform/redhat/tasks.py
+++ b/ipaplatform/redhat/tasks.py
@@ -47,10 +47,6 @@ from ipaplatform.paths import paths
 from ipaplatform.redhat.authconfig import RedHatAuthConfig
 from ipaplatform.base.tasks import BaseTaskNamespace
 
-# pylint: disable=ipa-forbidden-import
-from ipalib.constants import IPAAPI_USER
-# pylint: enable=ipa-forbidden-import
-
 logger = logging.getLogger(__name__)
 
 _ffi = FFI()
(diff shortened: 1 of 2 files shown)


ipaplatform/redhat/tasks.py
```diff
@@ -518,9 +514,9 @@ def _create_tmpfiles_dir(self, name, mode, uid, gid):
         os.chmod(name, mode)
         os.chown(name, uid, gid)
 
-    def create_tmpfiles_dirs(self):
+    def create_tmpfiles_dirs(self, ipaapi_user):
```
tiran: How about you replace the code with ``systemd-tmpfiles`` and fold it into ``configure_tmpfiles``? You can get rid of duplicated configuration, too.

```
check_call(['systemd-tmpfiles', '--create', os.path.basename(ETC_TMPFILESD_IPA)]
```

felipevolpone: Hi @tiran, thanks for reviewing this.
That sounds a good idea to me, but in some places only `create_tmpfiles_dirs` is called, like:
* https://github.com/freeipa/freeipa/blob/master/ipaserver/install/server/install.py#L1055
* https://github.com/freeipa/freeipa/blob/master/ipaserver/install/server/replicainstall.py#L1359

Which leads to: Is harmful to change this calls to `configure_tmpfiles` (after adding `systemd-tmpfiles` to the method)?

tiran: You may copy the config file one time more than necessary. I don't see any harm in that.

tomaskrizek:
master:

* 0b7d9c5a7b271f0e8b296dac86525b9ce948084a Removing part of circular dependency of ipalib in ipaplaform

tomaskrizek:
@felipevolpone Please backport manually against `ipa-4-5`. In that version `IPAAPI_USER` is also used of the tmpfiles function.

PR #990, opened 2017-08-22
Remove e.message and a relative import

stlaz:
Partially fixes:
- https://pagure.io/freeipa/issue/4985
- https://pagure.io/freeipa/issue/6874

### DIFF
--- a/ipaserver/plugins/migration.py
+++ b/ipaserver/plugins/migration.py
@@ -28,7 +28,7 @@ from ipalib import api, errors, output
 from ipalib import Command, Password, Str, Flag, StrEnum, DNParam, Bool
 from ipalib.cli import to_cli
 from ipalib.plugable import Registry
-from .user import NO_UPG_MAGIC
+from ipaserver.plugins.user import NO_UPG_MAGIC
 from ipalib import _
 from ipapython.dn import DN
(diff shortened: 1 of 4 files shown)


ipaserver/plugins/migration.py
```diff
@@ -28,7 +28,7 @@
 from ipalib import Command, Password, Str, Flag, StrEnum, DNParam, Bool
 from ipalib.cli import to_cli
 from ipalib.plugable import Registry
-from .user import NO_UPG_MAGIC
+from ipaserver.plugins.user import NO_UPG_MAGIC
```
stlaz: While @tiran did not think this change necessary, since this would be the only spot to have such a relative import and also by the "explicit is better than implicit" (maybe explicitly explicit in this case, I agree) I would keep it. It fixes `bad-python3-import` Pylint anyway and might be cleaner than just putting a pylint-disable here.

stlaz:
This is a no-brainer, adding prioritized flag since it's been hanging here for 6 days already.

pvomacka:
master:

* e6a9de8a2ea453843e71ff717bee7dd74383db64 py3: remove Exception.message appearances
* 2bc5b7f044435921b9fa1d5728731a843919102e py3: remove relative import

PR #994, opened 2017-08-23
py3: fix schema response for py2 server with py3 client

frasertweedale:
The schema RPC response can include a bytes value for the
'topic_topic' field (when the server is running in Python 2).
In this case, a Python 3 client fails to initialise the API
when a schema fetch is required, because it cannot serialise
the bytes values in the schema object to JSON.

Ensure the 'topic_topic' value is unicode.


stlaz:
Works for me, thanks!

pvomacka:
master:

* b1e1109679c0a95aadbd63ce2587561b58d7a050 py3: fix schema response for py2 server with py3 client

PR #995, opened 2017-08-23
py3: handle bytes in schema response

frasertweedale:
If a server is running under py2 it could return a bytes value for
the 'topic_topic' field in the schema response.  A py3 client fails
to handle this (in one place it applies 'str' to it, which raises
BytesWarning; in other places it tries to serialise the schema to
JSON which fails because of the bytes value).

Handle the case where 'topic_topic' is not unicode, and handle bytes
values when serialising the schema to JSON.

### DIFF
--- a/ipaclient/remote_plugins/schema.py
+++ b/ipaclient/remote_plugins/schema.py
@@ -64,6 +64,12 @@ _PARAMS = {
 }
 
 
+def json_default(obj):
+    if isinstance(obj, bytes):
+        return obj.decode('utf-8')
+    raise TypeError
+
+
 class _SchemaCommand(ClientCommand):
     pass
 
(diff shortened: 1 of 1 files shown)


stlaz:
Thank you for this fix, I was trying to untangle the schema plugin spaghetti some time ago but failed. This should also fix https://pagure.io/freeipa/issue/6809 I believe.

frasertweedale:
@stlaz yep, that looks like the same issue; I'll update commit message to include it.

ipaclient/remote_plugins/schema.py
```diff
@@ -64,6 +64,13 @@
 }
 
 
+class BytesEncoder(json.JSONEncoder):
+    def default(self, obj):
```
tiran: Instead of using a subclass, ``json.dumps()`` also takes a ``default`` function kwarg. The function takes one argument and should raise TypeError in case of an unsupported type.

frasertweedale: @tiran is there something undesirable about using the class (apart from a tiny bit more code and another layer of indirection?)

tiran: Both solutions are fine. The default function argument is a few lines less code. The json module doesn't have to create a new encoder class for every call and can use the default encoder instance instead.

stlaz:
Tested with current and RHEL 6.9 client and works well. Thank you for the fix, ACK!

pvomacka:
master:

* 947ac4bc1f6f4016cf5baf2ecb4577e893bc3948 py3: handle bytes in schema response

PR #996, opened 2017-08-23
ipa-pki-retrieve-key: ensure we do not crash

frasertweedale:
If ipa-pki-retrieve-key fails for some reason (which may be a
"legitimate" reason, e.g. the server it is attempting to contact
being offline), the program terminates with an uncaught exception,
resulting in crash report.

Catch all exceptions; if an exception gets raised, report the
traceback and exit with nonzero status.

Fixes: https://pagure.io/freeipa/issue/7115

### DIFF
--- a/install/tools/ipa-pki-retrieve-key
+++ b/install/tools/ipa-pki-retrieve-key
@@ -4,29 +4,39 @@ from __future__ import print_function
 
 import os
 import sys
+import traceback
 
 from ipalib import constants
 from ipalib.config import Env
 from ipaplatform.paths import paths
 from ipaserver.secrets.client import CustodiaClient
 
-env = Env()
-env._finalize()
 
-keyname = "ca_wrapped/" + sys.argv[1]
-servername = sys.argv[2]
+def main():
+    env = Env()
+    env._finalize()
 
-service = constants.PKI_GSSAPI_SERVICE_NAME
-client_keyfile = os.path.join(paths.PKI_TOMCAT, service + '.keys')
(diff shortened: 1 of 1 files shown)


stlaz:
Thanks for the patch. You'll need to rebase on current master for the tests to pass.

If there's a way I can run the script on my own, could you please post it so that I can check whether nothing breaks?

frasertweedale:
@stlaz thanks, I rebased.

To run the script:
```
sudo -u pkiuser /usr/libexec/ipa/ipa-pki-retrieve-key 'caSigningCert cert-pki-ca' f25-0.ipa.local
```

Substitute whatever certificate nickname or target server hostname you want.
Obviously the program needs to fail so you can refer to a server that does not exist,
or one that exists but is not available.

Run the program and make sure that abrt does not log an unhandled python exception
to the journal.  Note: we use `traceback` to print a traceback and we exit nonzero, so
the actual behaviour of the program is difficult to distinguish from a true unhandled exception,
so you must check whether abrt logs an unhandled exception or not (it should not).

Hope that helps; thanks for reviewing!

stlaz:
@frasertweedale Thanks for the patch. I tried this even with the `abrt` tool and it works well.

I wish I got to this PR earlier, I knew the change was quite obvious, sorry for the delay.

ACK.

stlaz:
master:

* 09f746f56823ec6120437ba625f0db9b5d704e3e ipa-pki-retrieve-key: ensure we do not crash

frasertweedale:
@stlaz no worries; thanks for reviewing!

PR #997, opened 2017-08-23
Uninstall: fix BytesWarning exception

stlaz:
When uninstalling, if server does not have all roles, exception
is thrown as the role is not found. `member_principal` variable
has to be string here, otherwise we're using str on bytes.


stlaz:
You most probably see this a couple times a day, anyone can test it, it's pretty easy.

frasertweedale:
Does what it says on the tin.  Here have an ACK :)

pvomacka:
master:

* 6ca8787cc7f4865cfce20241440f17d3a41bef78 Uninstall: fix BytesWarning exception

PR #998, opened 2017-08-23
Unify storing certificates in LDAP

stlaz:
Recent certificate refactoring left the system in a state where
the certificates are somewhere converted to DER format, somewhere
directly sent to ipaldap as IPACertificate objects. The latter
is the desirable way, make sure it's the one commonly used.


frasertweedale:
Tested; works.  (I _hope_ I did not miss any codepaths :S)

pvomacka:
master:

* 31142ead830b441c81ee1439e81210426792cac5 Unify storing certificates in LDAP

PR #1000, opened 2017-08-24
Fix CA-less installation

frasertweedale:
Fix CA-less installation failure due to an incorrect use of with
statement.

Fixes: https://pagure.io/freeipa/issue/7118

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -579,8 +579,8 @@ class NSSDatabase(object):
             self.add_cert(cert, nickname, EMPTY_TRUST_FLAGS)
 
         if extracted_key:
-            with tempfile.NamedTemporaryFile(), tempfile.NamedTemporaryFile() \
-                    as (in_file, out_file):
+            with tempfile.NamedTemporaryFile() as in_file, \
+                    tempfile.NamedTemporaryFile() as out_file:
                 x509.write_certificate_list(extracted_certs, in_file.name)
                 in_file.write(extracted_key)
                 in_file.flush()
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -581,7 +581,8 @@ class NSSDatabase(object):
         if extracted_key:
             with tempfile.NamedTemporaryFile() as in_file, \
                     tempfile.NamedTemporaryFile() as out_file:
-                x509.write_certificate_list(extracted_certs, in_file.name)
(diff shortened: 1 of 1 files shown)


stlaz:
Thanks for fixing my doodoo, I tend to struggle with the `with` statement when multiple arguments appear.

frasertweedale:
@stlaz I think there's another regression there (the same example from the ticket now fails for a different reason - it cannot find a cert for they key).  But I'll push forward on that tomorrow :)

stlaz:
@frasertweedale Yes, i tried the fix and got to it, too. So I went ahead to check modulus of my private key + certificate but they matched so it would seem like a regression, unfortunately :(
I usually perform CA-less installation with PKCS12 files instead of just cert/key pairs and that installation passes for me but since I haven't tried this use-case up until now I did not notice the regressions.

stlaz:
The following will get it working. We opened the same file at a different spot but the original file position indicator would not be moved. The result is that the certificate gets rewritten by the private key 🙄 
```diff
diff --git a/ipapython/certdb.py b/ipapython/certdb.py
index 33f32a4c9..56c97936d 100644
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -581,7 +581,8 @@ class NSSDatabase(object):
         if extracted_key:
             with tempfile.NamedTemporaryFile(), tempfile.NamedTemporaryFile() \
                     as (in_file, out_file):
-                x509.write_certificate_list(extracted_certs, in_file.name)
+                for cert in extracted_certs:
+                    in_file.write(cert.public_bytes(x509.Encoding.PEM))
                 in_file.write(extracted_key)
                 in_file.flush()
                 out_password = ipautil.ipa_generate_password()
```

frasertweedale:
@stlaz I added another commit to this PR with your fix (and correct authorship info :) )

frasertweedale:
I think I might just ACK this because @stlaz reviewed my commit and I reviewed his change...

stlaz:
Yes, the ACK is correct :)

pvomacka:
master:

* 477b3dca802349ac4e8d60cfcabc7fe31d0ab9ef Fix incorrect 'with' statement in CA-less installation
* 7a86ff5d9bf5d9bf0f5970298addcd5b0f21d728 Fix cert file creation during CA-less installation

PR #1003, opened 2017-08-25
py3: Fix searching for yubikeys

stlaz:
Bumping the requirements for the python-yubikey package. This is
unfortunately most probably fixed only in Fedora because there was no
upstream release for about a year and a half now. That also means
bumping the version in ipasetup.py.in would be pointless.

https://pagure.io/freeipa/issue/7121

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -730,7 +730,8 @@ Requires: python3-qrcode-core >= 5.0.0
 Requires: python3-pyasn1
 Requires: python3-pyasn1-modules
 Requires: python3-dateutil
-Requires: python3-yubico >= 1.2.3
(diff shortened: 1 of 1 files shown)


stlaz:
This will break PRCI because the package is not yet in Fedora. Feel free to add karma in bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2017-63975ea815

stlaz:
python-yubico was pushed to stable on f26, @tomaskrizek, please, add it to our copr and re-run the tests, please.

pvomacka:
master:

* 0be9a1721133ad512a7420f6508249d4e8453d90 py3: Fix searching for yubikeys

PR #1004, opened 2017-08-25
cert: fix application of 'str' to bytes when formatting otherName

frasertweedale:


stlaz:
Seems pretty straightforward, ACK :)

PR #1005, opened 2017-08-25
Fixing how sssd.conf is updated when promoting a client to replica

felipevolpone:
When promoting a client to a replica we have to change sssd.conf, deleting _srv_ part from 'ipa_server' property and setting 'ipa_server_mode' to true.

Previously, the wrong domain could be updated since the ipa_domain variable was not being used properly.

Ticket: https://pagure.io/freeipa/issue/7127

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -434,30 +434,27 @@ def promote_sssd(host_name):
     sssdconfig.import_config()
     domains = sssdconfig.list_active_domains()
 
-    ipa_domain = None
-
     for name in domains:
         domain = sssdconfig.get_domain(name)
         try:
             hostname = domain.get_option('ipa_hostname')
             if hostname == host_name:
-                ipa_domain = domain
+                break
         except SSSDConfig.NoOptionError:
             continue
-
-    if ipa_domain is None:
-        raise RuntimeError("Couldn't find IPA domain in sssd.conf")
     else:
-        domain.set_option('ipa_server', host_name)
-        domain.set_option('ipa_server_mode', True)
-        sssdconfig.save_domain(domain)
-        sssdconfig.write()
+        raise RuntimeError("Couldn't find IPA domain in sssd.conf")
 
-        sssd = services.service('sssd', api)
-        try:
-            sssd.restart()
-        except CalledProcessError:
-            logger.warning("SSSD service restart was unsuccessful.")
+    domain.set_option('ipa_server', host_name)
+    domain.set_option('ipa_server_mode', True)
+    sssdconfig.save_domain(domain)
+    sssdconfig.write()
+
+    sssd = services.service('sssd', api)
+    try:
+        sssd.restart()
+    except CalledProcessError:
+        logger.warning("SSSD service restart was unsuccessful.")
 
 
 def promote_openldap_conf(hostname, master):
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -1863,10 +1863,14 @@ def upgrade_configuration():
         cainstance.ensure_ipa_authority_entry()
 
     set_sssd_domain_option('ipa_server_mode', 'True')
+    set_sssd_domain_option('ipa_server', api.env.host)
 
     sssdconfig = SSSDConfig.SSSDConfig()
     sssdconfig.import_config()
     sssd_enable_service(sssdconfig, 'ifp')
+
+    sssd = services.service('sssd', api)
+    sssd.restart()
 
     krb = krbinstance.KrbInstance(fstore)
     krb.fqdn = fqdn


pvoborni:
There might be a bunch of existing replicas with incorrect settings. I believe similar config fix should be run also on upgrade.

stlaz:
Removing the prioritized flag so that it does not consume PR CI resources first since the issue with this PR seems to be deterministic. Please, fix the actual issue first.

felipevolpone:
I just rebased the code and the tests are fine now.

ipaserver/install/server/upgrade.py
```diff
@@ -1863,11 +1863,15 @@ def upgrade_configuration():
...
 
     set_sssd_domain_option('ipa_server_mode', 'True')
+    set_sssd_domain_option('ipa_server', api.env.host)
 
     sssdconfig = SSSDConfig.SSSDConfig()
     sssdconfig.import_config()
     sssd_enable_service(sssdconfig, 'ifp')
 
+    sssd = services.service('sssd', api)
+    sssd.restart()
```
stlaz: Why is the restart necessary?

abbra: Yes, otherwise sssd wouldn't apply its configuration change.

stlaz: Thanks for the explanation

stlaz:
Personally, I think we should not try to fix anything in the upgrade. If we do what's proposed, the other IPA domain config still remains broken even after the upgrade but it may not be so apparent.

abbra:
@stlaz this is so wrong. We have a replica originally promoted before 4.5 is broken after upgrade because we add PKINIT enablers during upgrade. As result, in CAless and CA external cases we should ensure we are using exactly the KDC located on the very same machine. However, even in other cases before the new replica promotion mechanism came in, we had relied on the fact that on IPA masters/replicas we always use local services: LDAPI for LDAP access, explicitly defined KDC in krb5.conf and so on. Replica promotion approach wasn't fully implemented because it did not convert properly already existing client to an IPA master in this regard. We do not have other means to fix this up than in upgrade phase.

stlaz:
@abbra If I understand this problem correctly, this happens when the machine is a client to two different IPA domains. In that case, if we correct the settings for the first one during upgrade, the other one will still be broken because the `ipa_server` setting will still be pointing to this machine, although it definitely should not since this is a server to a different domain.
My point is, as long as the configuration is broken and IPA is visibly broken, the user knows they should fix something. If we fix this and make it seem like it's working, it may not strike anyone that the configuration is actually still broken.

stlaz:
What I did not realize is that a client to two IPA domains is not supported (thanks for @abbra to pointing that out) so it's fine to fix the supported solution and keep the upgrade fix as is.

ipaserver/install/server/replicainstall.py
```diff
@@ -448,9 +448,9 @@ def promote_sssd(host_name):
     if ipa_domain is None:
         raise RuntimeError("Couldn't find IPA domain in sssd.conf")
     else:
-        domain.set_option('ipa_server', host_name)
-        domain.set_option('ipa_server_mode', True)
-        sssdconfig.save_domain(domain)
+        ipa_domain.set_option('ipa_server', host_name)
+        ipa_domain.set_option('ipa_server_mode', True)
+        sssdconfig.save_domain(ipa_domain)
```
stlaz: IMHO this is a better fix:
```diff
diff --git a/ipaserver/install/server/replicainstall.py b/ipaserver/install/server/replicainstall.py
index 9d251b5e9..e80584cf0 100644
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -437,2 +436,0 @@ def promote_sssd(host_name):
-    ipa_domain = None
-
@@ -444 +442 @@ def promote_sssd(host_name):
-                ipa_domain = domain
+                break
@@ -447,3 +444,0 @@ def promote_sssd(host_name):
-
-    if ipa_domain is None:
-        raise RuntimeError("Couldn't find IPA domain in sssd.conf")
@@ -451,4 +446 @@ def promote_sssd(host_name):
-        domain.set_option('ipa_server', host_name)
-        domain.set_option('ipa_server_mode', True)
-        sssdconfig.save_domain(domain)
-        sssdconfig.write()
+        raise RuntimeError("Couldn't find IPA domain in sssd.conf")
@@ -456,5 +448,10 @@ def promote_sssd(host_name):
-        sssd = services.service('sssd', api)
-        try:
-            sssd.restart()
-        except CalledProcessError:
-            logger.warning("SSSD service restart was unsuccessful.")
+    domain.set_option('ipa_server', host_name)
+    domain.set_option('ipa_server_mode', True)
+    sssdconfig.save_domain(domain)
+    sssdconfig.write()
+
+    sssd = services.service('sssd', api)
+    try:
+        sssd.restart()
+    except CalledProcessError:
+        logger.warning("SSSD service restart was unsuccessful.")
```
It removes both redundant variable and an `else` branch but if you don't like it, I can ACK this as well.

felipevolpone: I liked your proposal, so squashing commits and re-basing.

rcritten:
This looks good to me but I'll let @stlaz give final ack.

stlaz:
Thanks, ACK!

stlaz:
master:

* 05acd0965f94f818073f0acc002427f704c0d922 Fixing how sssd.conf is updated when promoting a client to replica

stlaz:
@felipevolpone A manual backport for `ipa-4-5` is needed.

felipevolpone:
Backported to ipa-4-5 in PR #1075

PR #1006, opened 2017-08-28
py3: fix vault public key decoding

frasertweedale:
Part of: https://pagure.io/freeipa/issue/7033

This commit fixes one issue in FreeIPA related to vault.  It does not fully
resolve #7033 because there are changes needed in Dogtag too.


pvomacka:
master:

* 9cead4da2ee4a1b1abbb60e28dee1a78cda83bf5 py3: fix vault public key decoding

PR #1007, opened 2017-08-28
py3: minor fixes

stlaz:
These are some bugs that were introduced by an accident and showed while testing with RHEL 6.9 client.


felipevolpone:
LGTM

stlaz:
master:

* e09b6c260288b955c1f7d63209ff6b932f363ad1 rpc: don't encode bytes
* 1b78f79283e633abc5dd901ca4db99cea36aca1a cert: fix wrong assumption of cert-show result type

PR #1008, opened 2017-08-28
Fix ipa-server-upgrade: This entry already exists

flo-renaud:
ipa-server-upgrade fails when running the ipaload_cacrt plugin. The plugin
finds all CA certificates in /etc/httpd/alias and uploads them in LDAP
below cn=certificates,cn=ipa,cn=etc,$BASEDN.
The issue happens because there is already an entry in LDAP for IPA CA, but
with a different DN. The nickname in /etc/httpd/alias can differ from
$DOMAIN IPA CA.

To avoid the issue:
1/ during upgrade, run a new plugin that removes duplicates and restarts ldap
(to make sure that uniqueness attr plugin is working after the new plugin)
2/ modify upload_cacert plugin so that it is using $DOMAIN IPA CA instead of
cn=$nickname,cn=ipa,cn=etc,$BASEDN when uploading IPA CA.

https://pagure.io/freeipa/issue/7125

### DIFF
--- a/install/updates/90-post_upgrade_plugins.update
+++ b/install/updates/90-post_upgrade_plugins.update
@@ -15,6 +15,7 @@ plugin: update_ca_renewal_master
 plugin: update_idrange_type
 plugin: update_pacs
 plugin: update_service_principalalias
+plugin: update_fix_duplicate_cacrt_in_ldap
 plugin: update_upload_cacrt
 # update_ra_cert_store has to be executed after update_ca_renewal_master
 plugin: update_ra_cert_store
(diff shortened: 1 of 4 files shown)


stlaz:
master:

* 69bda6b440d6b84f042ff74b9ce708d963616eda Fix ipa-server-upgrade: This entry already exists

stlaz:
@flo-renaud please create a PR against 4-5 as well.

flo-renaud:
Hi @stlaz 
I opened PR https://github.com/freeipa/freeipa/pull/1023

PR #1010, opened 2017-08-28
Adds whoami DS plugin in case that plugin is missing

pvomacka:
When first installation of IPA has been done when whoami
plugin was not enabled in DS by default and then IPA was
upgraded to newer versions, then after upgrade to IPA 4.5
WebUI stops working. This is caused by new requirement on
whoami DS plugin which is used to obtain information about
logged in entity.

This fix adds check whether the plugin is enabled and if not
then IPA enables the plugin during upgrade.

https://pagure.io/freeipa/issue/7126

### DIFF
--- a/install/updates/20-whoami.update
+++ b/install/updates/20-whoami.update
@@ -1,0 +1,14 @@
+dn: cn=whoami,cn=plugins,cn=config
+default:objectClass: top
+default:objectClass: nsSlapdPlugin
+default:objectClass: extensibleObject
+default:cn: whoami
+default:nsslapd-plugin-depends-on-type: database
+default:nsslapd-pluginDescription: whoami extended operation plugin
+default:nsslapd-pluginEnabled: on
+default:nsslapd-pluginId: whoami-plugin
+default:nsslapd-pluginInitfunc: whoami_init
+default:nsslapd-pluginPath: libwhoami-plugin
+default:nsslapd-pluginType: extendedop
+default:nsslapd-pluginVendor: 389 Project
+default:nsslapd-pluginVersion: 1.0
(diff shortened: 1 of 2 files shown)


rcritten:
Seems to me this would be cleaner as an update plugin. It would also run as part of the update but would be less code to maintain (no changes to the installer).

It needs to be a plugin rather than an update file so it can signal that a 389-ds restart is needed.

pvomacka:
Hi @rcritten thank you for the review. I created update file with whoami plugin instead of doing my own check.

rcritten:
This update file needs to be added to install/updates/Makefile.am

I'm not sure this will work as a standalone update since a restart of 389-ds is needed. I think that during a major upgrade a restart is done anyway to re-enable the ports so it may be ok, I just need to do further testing.

rcritten:
Otherwise it works as an update file for our purposes as far as I can tell.

I tested by:
* Installing 4.5 from git
* edit whoami.py to remove NO_CLI, restart httpd and remove schema
* ipa -vv whoami (it errors because the output format isn't setup, but it's fine because the data is ok)
* systemctl stop dirsrv@EXAMPLE-COM
* edit dse.ldif to remove whoami plugin
* systemctl start dirsrv@EXAMPLE-COM
* ipa -vv whoami (fails because no DS plugin as expected)
* ipa-server-upgrade
* ipa -vv whoami (succeeds again, plugin added)

So you just need to add it to Makefile.am so it gets installed.

pvomacka:
@rcritten Thank you for testing and another review. I just updated install/updates/Makefile.am file to include whoami.update file.

Tiboris:
Thanks Pavel and Rob!

Works for me.

rcritten:
+1

tomaskrizek:
master:

* 45bd31b436ffea298b12034e8a9b829569d19d2c Adds whoami DS plugin in case that plugin is missing

PR #999, opened 2017-08-23
dnssec: fix localhsm.py utility script

tomaskrizek:
See e6b2ed6b68589ff7ee39b95559836af54f39e2de for details.

Fixes https://pagure.io/freeipa/issue/7116

Signed-off-by: Tomas Krizek <tkrizek@redhat.com>

PR #973, opened 2017-08-15
test_dnssec: re-add named-pkcs11 workarounds

tomaskrizek:
DNSSEC tests starrted to fail again, probably due to a bug in
some underlaying component.

This reverts commit 8bc677512296a7e94c29edd0c1a96aa7273f352a
and makes the xfail test check less strict - it will no longer
mark the test suite red if it passes.

Related https://pagure.io/freeipa/issue/5348

PR #1011, opened 2017-08-28
py3: dnssec

tomaskrizek:
This PR is a partial fix that should allow DNSSEC installation for master. Keys will not be distributed to replicas. With my limited DNSSEC/IPA knowledge, I wasn't able to verify the data stored in LDAP are actually correct. In case they are not, this would prevent installation of DNSSEC replicas in the future. 

Our DNSSEC tests are not passing, thus we can't use them to verify this PR.  Given these circumstances, I propose to officially discourage DNSSEC installation in 4.6.0.

This PR supersedes #898. For review, it is highly recommended to rebase on #999.


tomaskrizek:
As we agreed on the devel mtg, we'll keep DNSSEC in Python 2 for now.

stlaz:
You'll need to rebase on the current master for the tests to pass.

stlaz:
Seems to have been closed by an accident.

ipaserver/dnssec/bindmgr.py
```diff
@@ -141,17 +145,21 @@ def get_zone_dir_name(self, zone):
...
-                    escaped += chr(c)
+                if six.PY2:
+                    # PY3 char is already int
+                    char = ord(char)
+                if (
+                    (char >= 0x30 and char <= 0x39) or  # digit
+                    (char >= 0x41 and char <= 0x5A) or  # uppercase
+                    (char >= 0x61 and char <= 0x7A) or  # lowercase
+                    char == 0x2D or                     # hyphen
+                    char == 0x5F                        # underscore
```
tiran: ETOOMUCHC

```
if six.PY3:
    char = chr(char)
if char.isalnum() or char in "-_":
    escaped += char.lower()
else:
    escaped += "%%%02X" % ord(char)
```

tomaskrizek:
If someone wants to take over this PR, feel free to do so. I think DNSSEC test suite should be fixed first (#973) so we can at least somehow verify the functionality.

tiran:
Closed in favor of new PR #1289

PR #1012, opened 2017-08-29
csrgen: support openssl 1.0 and 1.1

abbra:
Support both openssl 1.0 and 1.1 APIs where sk_* functions got prefixed
with OPENSSL_ in the latter version.

Since referencing a symbol from a dynamically loaded library generates
exception, use the AttributeError exception to catch it and fall back to
the older method.

Fixes https://pagure.io/freeipa/issue/7110


frasertweedale:
Tested, works with f25 (openssl-1.0) and f26 (openssl-1.1).  ACK.

tomaskrizek:
master:

* 79378c90512a1cdd5f3d5ec6482e434caea06e01 csrgen: support openssl 1.0 and 1.1

PR #1018, opened 2017-08-30
Python3: Fix winsync replication agreement

flo-renaud:
When configuring a winsync replication agreement, the tool performs a search
on AD for defaultNamingContext. The entry contains the value as a bytes, it
needs to be decoded otherwise subsequent calls to
DN(WIN_USER_CONTAINER, self.ad_suffix) will fail.

### DIFF
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -1089,7 +1089,8 @@ class ReplicationManager(object):
                                    ['defaultNamingContext'])
             for dn,entry in res:
                 if dn == "":
-                    self.ad_suffix = entry['defaultNamingContext'][0]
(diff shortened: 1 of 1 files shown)


stlaz:
You'll need to rebase on current master for the tests to pass.

stlaz:
I am sorry about this, should have noticed earlier - ticket 4985 has already been closed, we need to use https://pagure.io/freeipa/issue/7131 instead.

flo-renaud:
Hi @stlaz,
I have update the commit message with the right ticket number. Thank you for reminding me!

stlaz:
master:

* 93be966dafab8cb50c94a5248f9db3826c26d95b Python3: Fix winsync replication agreement

PR #1024, opened 2017-08-31
py3: turn all shebangs to python3

stlaz:
This patchset turns all shebangs in IPA scripts to Python 3. Note that this may seem like going against what we agreed on some two months ago but the outcome to turn everything to `/usr/bin/python` to use default python version would be a setback in the Fedora planning: https://fedoraproject.org/wiki/FinalizingFedoraSwitchtoPython3 and there is no reason not to use the up-till-now long running Python 3 effort.

This also required to regenerate the `API.txt` file so that our checks don't fail to validate for python2/3 type discrepancies.

### DIFF
--- a/client/ipa-certupdate
+++ b/client/ipa-certupdate
@@ -1,4 +1,4 @@
-#! /usr/bin/python2 -E
+#!/usr/bin/python3 -E
 # Authors: Jan Cholasta <jcholast@redhat.com>
 #
 # Copyright (C) 2014  Red Hat
--- a/client/ipa-client-automount
+++ b/client/ipa-client-automount
@@ -1,4 +1,4 @@
-#!/usr/bin/python2 -E
+#!/usr/bin/python3 -E
 #
 # Authors:
 #   Rob Crittenden <rcritten@redhat.com>
--- a/client/ipa-client-install
+++ b/client/ipa-client-install
@@ -1,4 +1,4 @@
-#! /usr/bin/python2 -E
+#!/usr/bin/python3 -E
 # Authors: Simo Sorce <ssorce@redhat.com>
 #          Karl MacMillan <kmacmillan@mentalrootkit.com>
 #
--- a/contrib/copy-schema-to-ca-RHEL6.py
+++ b/contrib/copy-schema-to-ca-RHEL6.py
@@ -1,4 +1,4 @@
-#! /usr/bin/python2
+#!/usr/bin/python2
 
 """Copy the IPA schema to the CA directory server instance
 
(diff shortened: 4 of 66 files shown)


tiran:
linting is failing with a rather strange error:
``An exception has occured when running command: 'utf-8' codec can't decode bytes in position 32766-32767: unexpected end of data``

stlaz:
I believe this had something to do with the translation, I will see to that.

stlaz:
So the problem appears only in the Chinese translation where if the `s` in a format string like `something %(host)s something else` is followed directly by a Chinese letter, the translation does not consider the formatting substituter (in this case `host`) as valid and reports violation. I did not find why this is yet, though.

tiran:
The regular expression for ``%(foo)s`` is ``r'%\(\w+\)[srduoxf]\b'``. The ``\b`` suffix means:

https://docs.python.org/3/library/re.html#regular-expression-syntax
> \b
>    Matches the empty string, but only at the beginning or end of a word. A word is defined as a sequence of word characters. Note that formally, \b is defined as the boundary between a \w and a \W character (or vice versa), or between \w and the beginning/end of the string. This means that r'\bfoo\b' matches 'foo', 'foo.', '(foo)', 'bar foo baz' but not 'foobar' or 'foo3'.

Python 3's unicode matching code may be more strict and not consider a Chinese character as a valid boundary.

stlaz:
That would probably be it, it does not make sense to make an exception for a Chinese sign. We'll probably need to update the translations.

tiran:
The regular expression assumes that a formatting must either occur at the end of a string or a non-alnum character must come next. But it looks like Chinese doesn't require spaces there. GNU gettext doesn't require a space here, too. For Python strings it just has to be a valid Python expression. Python happily accepts ``"%(foo)sbar" % {'foo': 'bar'}``.

Let's remove ``\b`` from the three regular expressions in https://github.com/freeipa/freeipa/blob/master/ipatests/i18n.py#L70-L85

stlaz:
This change led to most of the translation errors disappear, so +1 there. We are now only dealing with some errors that may need to reword the original message so that they can safely be translated even in languages that leave out spaces.

tiran:
I'd like to land PR #1580 first so ``makeapi`` generates the same output under Python 2 and 3.

tiran:
#1585 has landed, please rebase. Hopefully #1585 and #1580 were enough to fix all outstanding issues of your PR.

tiran:
```
$ grep -R '#!/usr/bin/python2' .
./contrib/copy-schema-to-ca-RHEL6.py:#!/usr/bin/python2
./install/tools/ipa-custodia-check:#!/usr/bin/python2
```

The first line is correct. The second one is a bug. Let's fix it in another PR. I want to close this one ASAP.

tiran:
master:

* f31797c70ac6d1fe4a651def1f87350bec16d194 Have all the scripts run in python 3 by default

PR #1029, opened 2017-08-31
pkinit: fix sorting dicts

stlaz:
Python 3 won't allow this.

### DIFF
--- a/ipaserver/plugins/pkinit.py
+++ b/ipaserver/plugins/pkinit.py
@@ -121,6 +121,7 @@ class pkinit_status(Search):
         if server is not None:
             self.api.Object.server_role.ensure_master_exists(server)
 
-        result = sorted(self.get_pkinit_status(server, status))
+        result = sorted(self.get_pkinit_status(server, status),
+                        key=lambda d: d.get('server_server'))
 
         return dict(result=result, count=len(result), truncated=False)


ipaserver/plugins/pkinit.py
```diff
@@ -121,6 +121,6 @@ def execute(self, *keys, **options):
         if server is not None:
             self.api.Object.server_role.ensure_master_exists(server)
 
-        result = sorted(self.get_pkinit_status(server, status))
+        result = self.get_pkinit_status(server, status)
```
flo-renaud: The issue seems rather to be that on ligne 96, pkinit_server_server key can be absent from ipa_master_config.

stlaz: This is fixing this error which happens when you call `sorted` on a type which does not implement comparisons:
```
2380] ipa: ERROR: non-public: TypeError: '<' not supported between instances of 'dict' and 'dict'
2380] Traceback (most recent call last):
2380]   File "/usr/lib/python3.6/site-packages/ipaserver/rpcserver.py", line 370, in wsgi_execute
2380]     result = command(*args, **options)
2380]   File "/usr/lib/python3.6/site-packages/ipalib/frontend.py", line 450, in __call__
2380]     return self.__do_call(*args, **options)
2380]   File "/usr/lib/python3.6/site-packages/ipalib/frontend.py", line 478, in __do_call
2380]     ret = self.run(*args, **options)
2380]   File "/usr/lib/python3.6/site-packages/ipalib/frontend.py", line 800, in run
2380]     return self.execute(*args, **options)
2380]   File "/usr/lib/python3.6/site-packages/ipaserver/plugins/pkinit.py", line 124, in execute
2380]     result = sorted(self.get_pkinit_status(server, status))
2380] TypeError: '<' not supported between instances of 'dict' and 'dict'
```

Example:
```
python3 -c "{'milan': 'jednicka'} > {'milan': 'dvojka'}"
  File "<string>", line 1, in <module>
TypeError: '>' not supported between instances of 'dict' and 'dict'
```

Is there possibly one other error that would happen should the key was absent? Could it really be absent?

flo-renaud: Hi,
I had the issue with an IPA 4.4 server upgraded to 4.5:
  File "/usr/lib/python3.6/site-packages/ipaserver/rpcserver.py", line 370, in wsgi_execute
    result = command(*args, **options)
  File "/usr/lib/python3.6/site-packages/ipalib/frontend.py", line 450, in __call__
    return self.__do_call(*args, **options)
  File "/usr/lib/python3.6/site-packages/ipalib/frontend.py", line 478, in __do_call
    ret = self.run(*args, **options)
  File "/usr/lib/python3.6/site-packages/ipalib/frontend.py", line 800, in run
    return self.execute(*args, **options)
  File "/usr/lib/python3.6/site-packages/ipaserver/plugins/pkinit.py", line 124, in execute
    result = sorted(self.get_pkinit_status(server, status))
  File "/usr/lib/python3.6/site-packages/ipaserver/plugins/pkinit.py", line 96, in get_pkinit_status
    pkinit_servers = ipa_master_config['pkinit_server_server']
KeyError: 'pkinit_server_server'

This is a different error, though.

pvoborni:
There are two issues:

* doesn't work in Python3
* sorting doesn't work

It's fine if the PR fixes just the Python3 issue though it might be better if it would actually sort as it was probably initially intended. 

In any case, the commit message should have  been better. It doesn't describe what it fixes.

stlaz:
I was actually trying to do a pre-release hotfix but it's perhaps good it did not get in. I'll address both the issues here.
@flo-renaud it's pretty awesome you were able to find out about the upgrade issue, it would otherwise be very hard to discover.

stlaz:
I decided to fix the other issue in a separate PR since it needs its own ticket and may have different preferences as to to which branches it should be pushed: https://github.com/freeipa/freeipa/pull/1046

flo-renaud:
Hi @stlaz 
thank you for the updated PR. both python3 and the sorting now work.

stlaz:
master:

* fcc2c5da9724a78f9ea11436a09684db935ea085 pkinit: fix sorting dictionaries

PR #1030, opened 2017-09-01
Python3 - depend on python3-* packages by default in Fedora

stlaz:
This patchset is an easier way for a more proper solution in https://github.com/freeipa/freeipa/pull/1024. We should investigate the proper solution more to see why it fails but there's currently not enough time before release.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -932,7 +932,7 @@ install/tools/ipactl
 ipa
 '
 for P in $PY3_SUBST_PATHS; do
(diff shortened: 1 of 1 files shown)


tomaskrizek:
master:

* 3aff58c804145716a34b38a0d7009df62be17d0d spec: remove python3 workaround
* 263217ff46f7ec260629c2ace5640bd281be9914 spec: have the scripts depend on py3 packages
* aa969da461a2bbdfc2fc2de49c3f21b6de23e9e1 spec: remove strict options from shebangs

PR #1033, opened 2017-09-01
 Fixing internal error in param-{find,show}

felipevolpone:
Fixing internal error in param-{find,show} with nonexistent object and showing properly error message.

Since PR #1013 probably won't be updated anymore, I created this one with previous work plus some additional changes.

https://pagure.io/freeipa/issue/7134

### DIFF
--- a/ipatests/test_xmlrpc/test_schema_plugin.py
+++ b/ipatests/test_xmlrpc/test_schema_plugin.py
@@ -1,0 +1,121 @@
+#
+# Copyright (C) 2017  FreeIPA Contributors see COPYING for license
+#
+
+"""
+Test the `ipaserver/plugins/schema.py` module.
+"""
+
+import pytest
+
+from ipalib import errors
+from ipatests.test_xmlrpc.tracker.base import Tracker
+from ipatests.test_xmlrpc.xmlrpc_test import XMLRPC_test
+
+
+@pytest.mark.tier1
+class TestParamFindAndShowCommand(XMLRPC_test):
+    """Test functionality of the ipa param-{find,show} command"""
+
+    tracker = Tracker()
+
+    def test_param_find(self):
+        """Test param-find command"""
+
+        # right command without criteria
+        result = self.tracker.run_command('param_find', u'user-add')
+        assert len(result['result']) != 0, result
+        assert result['result'][0]['name'] == 'uid', result
+        assert result['result'][0]['cli_name'] == 'login', result
+        assert result['result'][0]['label'] == 'User login', result
+
+        # right command, right criteria
+        criteria = u'postalcode'
+        result = self.tracker.run_command('param_find', u'user-add', criteria)
+        assert len(result['result']) != 0, result
+        for item in result['result']:
+            assert (criteria in item['name'].lower() or
+                    criteria in item['doc'].lower()), item
+
+        # right command, wrong criteria
+        result = self.tracker.run_command('param_find', u'user-add', u'fake')
+        assert len(result['result']) == 0, result
+
+        # wrong command, wrong criteria
+        result = self.tracker.run_command('param_find', u'fake', u'fake')
+        assert len(result['result']) == 0, result
+
+    def test_param_show(self):
+        """Test param-show command"""
+
+        # right command, right criteria
+        criteria = u'uid'
+        result = self.tracker.run_command('param_show', u'user-add', criteria)
+        assert result['result'] is not None, result
+        assert result['result']['name'] == 'uid', result
+        assert result['result']['cli_name'] == 'login', result
+        assert result['result']['label'] == 'User login', result
+
+        # right command without criteria
+        with pytest.raises(errors.RequirementError):
+            self.tracker.run_command('param_show', u'user-add')
+
+        # right command, wrong criteria
+        with pytest.raises(errors.NotFound):
+            self.tracker.run_command('param_show', u'user-add', u'fake')
+
+        # wrong command, wrong criteria
+        with pytest.raises(errors.NotFound):
+            self.tracker.run_command('param_show', u'fake', u'fake')
+
+
+class TestOutputFindAndShowCommand(XMLRPC_test):
+    """Test functionality of the ipa output-{find,show} command"""
+    tracker = Tracker()
+
+    def test_output_find(self):
+        """Test output-find command"""
+
+        # right command without criteria
+        result = self.tracker.run_command('output_find', u'user-add')
+        assert len(result['result']) != 0, result
+        assert result['result'][0]['name'] == 'summary', result
+        assert result['result'][0]['doc'] == \
+            'User-friendly description of action performed', result
+
+        # right command, right criteria
+        criteria = u'result'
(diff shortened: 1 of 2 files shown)


frasertweedale:
NACK.  The `param-{find,show}` commands work for me now, but `output-{find,show}` still
error in some cases:

```
[f25-2:~/dev/freeipa] [ pull/1033● ] ftweedal% ipa output-find fake         
[f25-2:~/dev/freeipa] [ pull/1033● ] ftweedal% ipa output-show fake fake
[f25-2:~/dev/freeipa] [ pull/1033● ] ftweedal% ipa output-show user-mod fake
ipa: ERROR: fake: output not found
[f25-2:~/dev/freeipa] [ pull/1033● ] ftweedal% ipa output-show user-mod value
  Name: value
  Documentation: The primary_key value of the entry, e.g. 'jdoe' for a user
  Type: str
```

frasertweedale:
@felipevolpone the latest commit has fixed the failure cases from my last comment but I still found a couple cases that caused internal error:

```
[f25-2:~/dev/freeipa] [ pull/1033● ] ftweedal% ipa output-find fake fake   
[f25-2:~/dev/freeipa] [ pull/1033● ] ftweedal% ipa param-find user-mod last
```

stlaz:
Travis is having issues with their new environment for builds. I pushed a temporary fix to `master`, feel free to rebase for the tests to pass properly.

rcritten:
A test should be added in ipatests/test_xmlrpc as well to ensure that this continues to function as expected, including both positive (e.g. user_show) and negative (e.g. fake) tests.

felipevolpone:
Tests added and patch updated to cover the behavior Fraser pointed out.

ipatests/test_xmlrpc/test_schema_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
rcritten: We use a standard contributors header now:

Copyright (C) 2017  FreeIPA Contributors see COPYING for license

ipatests/test_xmlrpc/test_schema_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+@pytest.mark.tier1
+class TestParamFindAndShowCommand(XMLRPC_test):
+    """Test functionality of the ipa param-{find,show} command"""
+    tracker = Tracker()
+
+    def test_param_find(self):
+        """Test param-find command"""
+
+        # right command without criteria
+        criteria = u'postalcode'
```
rcritten: criteria is not used here

ipatests/test_xmlrpc/test_schema_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # right command, right criteria
+        criteria = u'postalcode'
+        result = self.tracker.run_command('param_find', u'user-add', criteria)
+        assert len(result['result']) != 0, result
+        for item in result['result']:
+            assert (criteria in item['name'].lower() or
+                    criteria in item['doc'].lower()), item
+
+        # wrong command, right criteria
```
rcritten: I guess you were on a roll but is there right criteria for a wrong command ;-)

ipatests/test_xmlrpc/test_schema_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        result = self.tracker.run_command('param_find', u'fake', criteria)
+        assert len(result['result']) == 0, result
+
+        # right command, wrong criteria
+        result = self.tracker.run_command('param_find', u'user-add', u'fake')
+        assert len(result['result']) == 0, result
+
+        # wrong command, wrong criteria
```
rcritten: Same here. Maybe this is legitimate but I'd think that once the command is determined to be bad it doesn't matter what else was passed, right? I mean, how can one determine the "right" criteria for a command without having the command defined? In this case, postalcode makes sense for users but not for groups, so isn't correct in all cases anyway.

ipatests/test_xmlrpc/test_schema_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        with pytest.raises(errors.NotFound):
+            result = self.tracker.run_command('param_show', u'user-add', u'fake')
+
+        # wrong command, right criteria
+        with pytest.raises(errors.NotFound):
+            result = self.tracker.run_command('param_show', u'fake', u'uid')
+
+        # wrong command, wrong criteria
+        with pytest.raises(errors.NotFound):
+            result = self.tracker.run_command('param_show', u'fake', u'fake')
```
rcritten: Two more similar criteria-related. I think a single test is sufficient.

ipatests/test_xmlrpc/test_schema_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class TestOutputFindAndShowCommand(XMLRPC_test):
+    """Test functionality of the ipa output-{find,show} command"""
+    tracker = Tracker()
+
```
rcritten: Here is a philosophical question: is it worth comparing the output to something known or is it safe to assume that any output is ok? I'm not sure how much introspective is required and unless we want to test every single possible command's output it will still be hit or miss.

felipevolpone:
@rcritten I updated the PR with changes you proposed. Also, I added a simple check, to each command, in order to ensure it returns at least one right thing.

ipatests/test_xmlrpc/test_schema_plugin.py
```diff
@@ -0,0 +1,121 @@
...
+    def test_param_find(self):
+        """Test param-find command"""
+
+        # right command without criteria
+        result = self.tracker.run_command('param_find', u'user-add')
+        assert len(result['result']) != 0, result
+        expected = {'cli_name': 'login', 'name': 'uid', 'label': 'User login'}
+        assert set(expected.items()).issubset(set(result['result'][0].items()))
```
felipevolpone: @rcritten if you think this line is too complicated; I can change it with separate comparisons, like:
```python
assert result['result'][0]['cli_name'] == 'login'
assert result['result'][0]['name'] == 'uid'
```

rcritten: Ideally it would be broken out so in case of failure it would be clear which value failed the assertion.

rcritten:
Works great.

Tiboris:
master:

* 7ab49dda6db52c0960f50f1744aa19dc70d3c62f schema: Fix internal error in param-{find,show} with nonexistent object
* 3822120077d5bd0f951560d3181654d1f28fca4e Fixing param-{find,show} and output-{find,show} commands

PR #1034, opened 2017-09-01
browser config: cleanup after removal of Firefox extension

pvoborni:
Firefox extension which served for configuring Kerberos auth in Firefox
until version which banned self-signed extensions was removed in commit
6c53765ac1746ea3cb82554775a37fe43af062e8.

Given that configure.jar, even older Firefox config tool, was removed
sometime before that, there is no use for signtool tool. It is good
because it is removed from Fedora 27 anyway. So removing last unused
function which calls it.

The removal of FF extension was not exactly clean so removing also
browserconfig.html which only purpose was to use the extension. Therefore
also related JS files are removed. This removal requires unauthorized.html
to be updated so that it doesn't point to non-existing page. And given that
it now points only to single config page, we can change link in UI login page
to this page (ssbrowser.html). While at it, improving buttons in ssbrowser.html.

Btw, commit 6c53765ac1746ea3cb82554775a37fe43af062e8 removed also generation of
krb.js. It had one perk - with that info ssbrowser.html could display real
Kerberos domain instead of only 'example.com'.  I don't have time to revert this
change so removing traces of krb.js as well.

https://pagure.io/freeipa/issue/7135


tomaskrizek:
master:

* b0184d10aba3bd0302b4ef7bdcecb513ae00c54d browser config: cleanup after removal of Firefox extension

PR #1035, opened 2017-09-04
WebUI: ID user override causes traceback in httpd/error_log

pvomacka:
The 'ipaanchoruuid' is link widget and it had set 'other_entity'
attribute to 'user'. That means that WebUI automatically checks
whether the created link is available (whether there is an existing
user with the name set as value of the 'ipaanchoruuid' field).
The user with that name is not obviously found because it tries to
find a user while the value is ipaanchoruuid (AD user).

Removing 'other_entity' from specification tells WebUI to not
check the link.

The link widget is even set to non-clicable state. So the check does not
makes sense now.

https://pagure.io/freeipa/issue/7139

### DIFF
--- a/install/ui/src/freeipa/idviews.js
+++ b/install/ui/src/freeipa/idviews.js
@@ -257,8 +257,7 @@ return {
                         {
                             $type: 'link',
                             name: 'ipaanchoruuid',
-                            label: '@i18n:objects.idoverrideuser.anchor_label',
-                            other_entity: 'user'
+                            label: '@i18n:objects.idoverrideuser.anchor_label'
                         },
                         {
                             $type: 'textarea',


pvoborni:
ID overrides can be defined also for IPA users - basically it is in every ID View other than "default trust view". So for that use case the link makes sense and this change can be seen as a regression.

Also I believe that `ipa user-show nonexistantuser` is a valid use case and there is no reason why Web UI should avoid using a similar API call.

So the questions are:

* is it OK that ipa user-show returns a traceback in debug mode?
* if not, where to fix it on API side?
* if yes, then shouldn't we close the bug?

pvomacka:
@pvoborni Ok, so then we should close this PR and rather think whether we should fix traceback in log or not.

PR #1042, opened 2017-09-05
Use TLS for the cert-find operation

rcritten:
Switch from using urllib.request to dogtag.https_request

https://pagure.io/freeipa/issue/7027

### DIFF
--- a/ipaserver/plugins/dogtag.py
+++ b/ipaserver/plugins/dogtag.py
@@ -248,7 +248,6 @@ import time
 import contextlib
 
 import six
-from six.moves import urllib
 
 from ipalib import Backend, api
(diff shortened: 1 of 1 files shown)


stlaz:
If you rebase your PR onto https://github.com/freeipa/freeipa/pull/1043, we should be able to find out what's wrong.

pvoborni:
I miss  explanation why the change is being done in the commit message.

stlaz:
You should probably remove the unused variable. Also, PR CI does not seem to like you for some reason 😞

stlaz:
master:

* fa6181293a1bf3207c4da1494d9d6cf9d715377b Use TLS for the cert-find operation

PR #1044, opened 2017-09-06
WebUI: Fix calling undefined method during reset passwords

pvomacka:
When calling reset password the whoami command is not called in batch
command, therefore the result is different then in calling
during reset password operation. That needs to be handled to properly
set entity_show method which needs to be called after to gather
data about logged in entity.

https://pagure.io/freeipa/issue/7143

### DIFF
--- a/install/ui/src/freeipa/ipa.js
+++ b/install/ui/src/freeipa/ipa.js
@@ -191,7 +191,7 @@ var IPA = function () {
             }
         }));
 
-        batch.add_command(that.get_whoami_command(true));
+        batch.add_command(that.get_whoami_command());
 
         batch.add_command(rpc.command({
             method: 'env',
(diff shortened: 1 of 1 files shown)


abbra:
Could you please split the commit in two: one for `get_whoami_command()` signature change and another one for the actual `data.result || data` change.

pvomacka:
@abbra thanks for review, I splitted changes into two commits.

abbra:
LGTM.

AdamWill:
FWIW I've tested (by applying the patch to the Fedora package and re-running the openQA tests) that the fix does indeed work. 👍

stlaz:
master:

* 755a50044cd2eefef3d365cc42c76a6ed4715afe WebUI: remove unused parameter from get_whoami_command
* a7ab63b84cb07c2a09a271b2bb598fcbf0d144f9 WebUI: Fix calling undefined method during reset passwords

PR #1045, opened 2017-09-06
Fix ipa-server-upgrade with server cert tracking

flo-renaud:
ipa-server-upgrade fails with Server-Cert not found, when trying to
track httpd/ldap server certificates. There are 2 issues in the upgrade:
- the certificates should be tracked only if they were issued by IPA CA or
IPA lighweight subCA (it is possible to have CA configured but 3rd part certs)
- the certificate nickname can be different from Server-Cert

The fix provides methods to find the server crt nickname for http and ldap,
and a method to check if the server certs are issued by IPA and need to be
tracked by certmonger.

https://pagure.io/freeipa/issue/7141

### DIFF
--- a/ipaserver/install/certs.py
+++ b/ipaserver/install/certs.py
@@ -42,6 +42,7 @@ from ipapython.certdb import get_ca_nickname, find_cert_from_txt, NSSDatabase
 from ipapython.dn import DN
 from ipalib import pkcs10, x509, api
 from ipalib.errors import CertificateOperationError
+from ipalib.install import certstore
 from ipalib.text import _
 from ipaplatform.paths import paths
 
@@ -658,6 +659,31 @@ class CertDB(object):
                                              principal=principal,
                                              subject=host,
                                              passwd_fname=self.passwd_fname)
+
+    def is_ipa_issued_cert(self, api, nickname):
+        """
+        Return True if the certificate contained in the CertDB with the
+        provided nickname has been issued by IPA.
+
+        Note that this method can only be executed if api has been initialized
+        """
+        # This method needs to compare the cert issuer (from the NSS DB
+        # and the subject from the CA (from LDAP), because nicknames are not
+        # always aligned.
+
+        cacert_subject = certstore.get_ca_subject(
+            api.Backend.ldap2,
+            api.env.container_ca,
+            api.env.basedn)
+
+        # The cert can be issued directly by IPA. In this case, the cert
+        # issuer is IPA CA subject.
+        cert = self.get_cert_from_db(nickname)
+        if cert is None:
+            raise RuntimeError("Could not find the cert %s in %s"
+                               % (nickname, self.secdir))
+
+        return DN(cert.issuer) == cacert_subject
 
 
 class _CrossProcessLock(object):
--- a/ipaserver/install/dsinstance.py
+++ b/ipaserver/install/dsinstance.py
@@ -1033,22 +1033,59 @@ class DsInstance(service.Service):
                 logger.error(
                     'Unable to restart DS instance %s: %s', ds_instance, e)
 
+    def get_server_cert_nickname(self, serverid=None):
+        """
+        Retrieve the nickname of the server cert used by dirsrv.
+
+        The method directly reads the dse.ldif to find the attribute
+        nsSSLPersonalitySSL of cn=RSA,cn=encryption,cn=config because
+        LDAP is not always accessible when we need to get the nickname
+        (for instance during uninstall).
+        """
+        if serverid is None:
+            serverid = self.get_state("serverid")
+        if serverid is not None:
+            dirname = config_dirname(serverid)
+            config_file = os.path.join(dirname, "dse.ldif")
+            rsa_dn = "cn=RSA,cn=encryption,cn=config"
+            with open(config_file, "r") as in_file:
+                parser = upgradeinstance.GetEntryFromLDIF(
+                    in_file,
+                    entries_dn=[rsa_dn])
+                parser.parse()
+                try:
+                    config_entry = parser.get_results()[rsa_dn]
+                    nickname = config_entry["nsSSLPersonalitySSL"][0]
+                    return nickname.decode('utf-8')
+                except (KeyError, IndexError):
+                    logger.error("Unable to find server cert nickname in %s",
+                                 config_file)
+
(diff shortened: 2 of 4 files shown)


ipaserver/install/certs.py
```diff
@@ -659,6 +660,38 @@ def request_service_cert(self, nickname, principal, host):
...
+        # From the NSSDB, retrieve the chain that issued the server cert
+        # (retrieve the whole chain because the cert may be
+        # issued by an IPA lighweight CA or IPA CA)
+        # and extract the root CA subject
+        try:
+            db_ca_chain = self.find_root_cert(nickname)[:-1]
+            db_ca_nickname = db_ca_chain[-1]
+            db_cacert = self.get_cert_from_db(db_ca_nickname)
+            db_cacert_subject = DN(db_cacert.subject)
+        except Exception:
```
stlaz: I'd rather avoid using bare `Exception`. Would `CalledProcessError` do here?
Also, IMO it would make sense to take either the first two or three steps and have them in a separated function (`get_issuer_db_nickname`/`get_issuer_cert_from_db`).

ipaserver/install/certs.py
```diff
@@ -659,6 +660,38 @@ def request_service_cert(self, nickname, principal, host):
                                              subject=host,
                                              passwd_fname=self.passwd_fname)
 
+    def cert_issued_by_ipa(self, nickname):
```
stlaz: This is rather a nitpick but I'd maybe rather call it `is_ipa_issued_cert`.

flo-renaud:
I simplified the PR to only handle certs issued by IPA CA (not by IPA subCAs).
If we want to track certs issued by a subCA, then it should be part of a specific RFE and PR (the current code of ipa-server-certinstall does not track certs issued by IPA subCAs).

ipaserver/install/dsinstance.py
```diff
@@ -1032,22 +1032,52 @@ def uninstall(self):
...
+            serverid = self.get_state("serverid")
+        if serverid is not None:
+            dirname = config_dirname(serverid)
+            try:
+                dsdb = certs.CertDB(self.realm, nssdir=dirname)
+                nickname = dsdb.find_server_certs()[0][0]
+                return nickname
+            except Exception as e:
+                logger.error("Unable to find server cert nickname in %s (%s)",
+                             dirname, e)
```
stlaz: This will probably print "index out of range" in case no server certs were found. Not sure whether we want that, although I could live with it.

ipaserver/install/dsinstance.py
```diff
@@ -1032,22 +1032,52 @@ def uninstall(self):
...
         dsdb = certs.CertDB(self.realm, nssdir=dirname)
-        dsdb.track_server_cert(self.nickname, self.principal,
-                               dsdb.passwd_fname,
-                               'restart_dirsrv %s' % serverid)
+        if dsdb.is_ipa_issued_cert(nickname):
+            dsdb.track_server_cert(nickname, self.principal,
+                                   dsdb.passwd_fname,
+                                   'restart_dirsrv %s' % serverid)
+        else:
+            logger.debug("Will not track server certificate %s as it is "
```
stlaz: I'd maybe point out we're talking DS server cert: "Will not track DS server certificate ..."

ipaserver/install/httpinstance.py
```diff
@@ -578,12 +583,17 @@ def uninstall(self):
...
     def start_tracking_certificates(self):
         db = certs.CertDB(self.realm, nssdir=paths.HTTPD_ALIAS_DIR)
-        db.track_server_cert(self.cert_nickname, self.principal,
-                             db.passwd_fname, 'restart_httpd')
+        nickname = self.get_mod_nss_nickname()
+        if db.is_ipa_issued_cert(nickname):
+            db.track_server_cert(nickname, self.principal,
+                                 db.passwd_fname, 'restart_httpd')
+        else:
+            logger.debug("Will not track server cert %s as it is not issued "
```
stlaz: Similarly to DS, it might we worth mentioning we're talking HTTP server cert.

flo-renaud:
Hi @stlaz 
thank you for your comments. I updated the PR with your suggestions.

ipaserver/install/dsinstance.py
```diff
@@ -1033,22 +1033,52 @@ def uninstall(self):
...
                     'Unable to restart DS instance %s: %s', ds_instance, e)
 
+    def get_server_cert_nickname(self, serverid=None):
+        """
+        Retrieve the nickname of the server cert used by dirsrv.
+
+        The method reads the NSS DB instead of using the attribute
+        nssslpersonalityssl of cn=RSA,cn=encryption,cn=config because
+        LDAP is not always accessible when we need to get the nickname
+        (for instance during uninstall).
```
frasertweedale: Is there a reason we cannot read the `nsSSLPersonalitySSL` attribute directly out of
`/etc/dirsrv/slapd-REALM-NAME/dse.ldif` ?

During uninstall, `DsInstance.stop_tracking_certificates` gets called before `remove_ds_instance`,
so `dse.ldif` should still be available to read directly.

flo-renaud: @frasertweedale 
I initially thought that this would be using the internal implementation of the directory server (cn=config stored in dse.ldif) but other parts of the code also directly read this file, so your idea is better.
I updated the PR with both your suggestions, thank you for the review!

frasertweedale:
Tested, works as expected.

There is one corner case not handled - what if the http/ldap service cert is issued by a Sub-CA of the IPA CA.

I think it is fine to not address this case now.  I'll file a new ticket where we can discuss it.

So, here is your ACK @flo-renaud  :)

frasertweedale:
Related issue for dealing with the case when HTTP/DS service cert is issued by
an IPA lightweight CA: https://pagure.io/freeipa/issue/7160

stlaz:
master:

* 87540fe1ef8a191e521ddf1584b4cbebb7dece94 Fix ipa-server-upgrade with server cert tracking

stlaz:
@frasertweedale thanks for the review!
@flo-renaud please, create a manual backport for `ipa-4-5`

flo-renaud:
@standa, PR https://github.com/freeipa/freeipa/pull/1097 created for ipa 4-5

PR #1046, opened 2017-09-06
pkinit: don't fail when no pkinit servers found

stlaz:
If we issue pkinit-status after an upgrade from a pre-4.5 ipa
version, it would have failed with KeyError since the
pkinit_server_server of IPA config was never initialized.

https://pagure.io/freeipa/issue/7144


flo-renaud:
Hi @stlaz 
Thank you for the patch, works as expected!

stlaz:
master:

* 16909a128b766f00dadff9e48a4bd4643134d32f pkinit: don't fail when no pkinit servers found

PR #1047, opened 2017-09-06
Use the user-provided CA chain file in connections & check for file existence

rcritten:
The user now may provide their own CA chain to the make API commands but it isn't honored. This is in effect a regression from when the crypto library was switched from NSS to OpenSSL. It is very useful if you want to be able to create a connection to an IPA server without enrolling as an IPA client. This is used in OpenStack to bootstrap some entries and create an OTP used later during enrollment.

The value is also not checked for existence throwing a generic "no such file" error rather than "file <foo> doesn't exist"

### DIFF
--- a/ipatests/test_ipalib/test_rpc.py
+++ b/ipatests/test_ipalib/test_rpc.py
@@ -28,8 +28,10 @@ import six
 # pylint: disable=import-error
 from six.moves.xmlrpc_client import Binary, Fault, dumps, loads
 # pylint: enable=import-error
+from six.moves import urllib
 
 from ipatests.util import raises, assert_equal, PluginTester, DummyClass
+from ipatests.util import Fuzzy
 from ipatests.data import binary_bytes, utf8_bytes, unicode_str
 from ipalib.frontend import Command
 from ipalib.request import context, Connection
@@ -341,3 +343,52 @@ class test_xml_introspection(object):
                 "command 'system.methodHelp' takes at most 1 argument")
         else:
             raise AssertionError('did not raise')
+
+
+class test_rpcclient_context(PluginTester):
+    """
+    Test the context in `ipalib.rpc.rpcclient` plugin.
+    """
+    def setup(self):
+        try:
+            api.Backend.rpcclient.connect(ca_certfile='foo')
+        except (errors.NetworkError, IOError):
+            raise nose.SkipTest('%r: Server not available: %r' %
+                                (__name__, api.env.xmlrpc_uri))
+
+    def teardown(self):
+        if api.Backend.rpcclient.isconnected():
+            api.Backend.rpcclient.disconnect()
+
+    def test_context_cafile(self):
+        """
+        Test that ca_certfile is set in `ipalib.rpc.rpcclient.connect`
+        """
+        ca_certfile = getattr(context, 'ca_certfile', None)
+        assert_equal(ca_certfile, 'foo')
+
+    def test_context_principal(self):
+        """
+        Test that principal is set in `ipalib.rpc.rpcclient.connect`
+        """
+        principal = getattr(context, 'principal', None)
+        assert_equal(principal, 'admin@%s' % api.env.realm)
+
+    def test_context_request_url(self):
+        """
+        Test that request_url is set in `ipalib.rpc.rpcclient.connect`
+        """
+        request_url = getattr(context, 'request_url', None)
+        assert_equal(request_url, 'https://%s/ipa/session/json' % api.env.host)
+
+    def test_context_session_cookie(self):
+        """
+        Test that session_cookie is set in `ipalib.rpc.rpcclient.connect`
+        """
+        fuzzy_cookie = Fuzzy('^ipa_session=MagBearerToken=[A-Za-z0-9+\/]+=*;$')
+
+        session_cookie = getattr(context, 'session_cookie', None)
+        # pylint-2 is incorrectly spewing Too many positional arguments
+        # pylint: disable=E1121
+        unquoted = urllib.parse.unquote(session_cookie)
+        assert(unquoted == fuzzy_cookie)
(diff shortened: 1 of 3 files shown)


stlaz:
Travis is having issues with their new environment for builds. I pushed a temporary fix to `master`, feel free to rebase for the tests to pass properly.

I wonder why PR CI did not pick up this PR to test.

ipatests/ipa-run-tests
```diff
@@ -33,13 +33,16 @@ import sys
 import pytest
 
 import ipalib
-import ipatests
+
+try:
+    import ipatests
+    HERE = os.path.dirname(os.path.abspath(ipatests.__file__))
+except ImportError:
+    HERE = os.getcwd()
```
frasertweedale: this (and in `conftest.py`)... belong to a `WIP: Make tests work locally for me.` commit?

Should be cleaned up... :)

install/tools/ipa-ca-install
```diff
@@ -253,6 +258,67 @@ def install(safe_options, options, filename):
...
+                      "uninstall procedure?", False):
+        ca_instance.print_msg("Aborting uninstall operation.")
+        sys.exit(0)
+
+    # Note that I'm completely ignoring the replication agreement so it
+    # doesn't matter what domain level this is. This is based on the
+    # (bad) assumption that this is only being executed to fix a blown
+    # install and not to remove the CA as a component.
+
+    # TODO: Figure out what is going on with serial # ranges
```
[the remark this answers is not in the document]
rcritten: This is for a CA uninstaller that was nacked. I'm carrying it around in my personal master in case I want to try to revive it.

ipatests/ipa-run-tests
rcritten: Yeah, I had a few extra local patches I forgot to remove before last pushing.

stlaz:
Be a  lamb and fix the following code issues, please:
```
************* Module ipatests.test_ipalib.test_rpc
ipatests/test_ipalib/test_rpc.py:352: [E0213(no-self-argument), test_rpcclient_context.setup] Method should have "self" as first argument)
ipatests/test_ipalib/test_rpc.py:359: [E0213(no-self-argument), test_rpcclient_context.teardown] Method should have "self" as first argument)
ipatests/test_ipalib/test_rpc.py:391: [E1121(too-many-function-args), test_rpcclient_context.test_context_session_cookie] Too many positional arguments for function call)
ipatests/test_ipalib/test_rpc.py:34: [W0611(unused-import), ] Unused assert_deepequal imported from ipatests.util)
```

stlaz:
The following test fails: `test_rpcclient_context.test_context_session_cookie`

ipatests/test_ipalib/test_rpc.py
```diff
@@ -341,3 +343,52 @@ def test_help_many_params(self):
...
+        Test that request_url is set in `ipalib.rpc.rpcclient.connect`
+        """
+        request_url = getattr(context, 'request_url', None)
+        assert_equal(request_url, 'https://%s/ipa/session/json' % api.env.host)
+
+    def test_context_session_cookie(self):
+        """
+        Test that session_cookie is set in `ipalib.rpc.rpcclient.connect`
+        """
+        fuzzy_cookie = Fuzzy('^ipa_session=MagBearerToken=[A-Za-z0-9+\/+]=+;$')
```
stlaz: Your regexp is wrong. It should be `Fuzzy('^ipa_session=MagBearerToken=[A-Za-z0-9+/]+=+;$')` (you misplaced the positive iteration)

ipatests/test_ipalib/test_rpc.py
```diff
@@ -341,3 +343,52 @@ def test_help_many_params(self):
...
+
+        session_cookie = getattr(context, 'session_cookie', None)
+        # pylint-2 is incorrectly spewing Too many positional arguments
+        # pylint: disable=E1121
+        unquoted = urllib.parse.unquote(session_cookie)
+        assert(unquoted == fuzzy_cookie)
```
stlaz: I believe `urllib.parser.unqoute` will return a `str` instance. By default, `Fuzzy` expects `unicode` type, you can either specify the type as in `Fuzzy(type=str)` or will need to do `assert(unicode(unquoted == fuzzy_cookie))`. The first approach would be preffered, though.

ipatests/test_ipalib/test_rpc.py
rcritten: Bah, that's what I guess for rushing. I also made trailing = to be *.

ipatests/test_ipalib/test_rpc.py
rcritten: It is a unicode on my system. Perhaps it maintains the current type.

stlaz: I did not test it, if it's unicode, it should be fine :)

ipatests/test_ipalib/test_rpc.py
stlaz: Ah, right, I did not realize that, good catch.

rcritten:
In addition to cafile being set it is probably worthwhile to check on permissions/existence as well because otherwise an OSError may be thrown which from experience is tricky to track down. Something like:

    if not os.path.isfile(cafile) or not os.access(cafile, os.R_OK):
        raise RuntimeError("cafile %s doesn't exist or is unreadable" % cafile)

I can do this as a separate PR if that is preferable.

rcritten:
Any other reservations about this one?

stlaz:
None, I just want to re-check this once more before ACKing

stlaz:
Alright, LGTM, can be pushed.

rcritten:
master:

* 17bda0b1a532ff2ac8503187e2d6e648f63d427f Use the CA chain file from the RPC context
* e8a26afb94e57ce08efad60ed9684bc58b9b1c58 Add test to ensure that properties are being set in rpcclient
* 01bfe2247e2297e9e0a55fb1baa0078525747c33 If the cafile is not present or readable then raise an exception

PR #1049, opened 2017-09-07
ldif: handle attribute names as strings

stlaz:
ldif.LDIFRecordList handles all attribute names as utf-8 strings
and all attribute values as bytes. If we take the attribute value
and try to search for it in the entry (= dictionary), if it contains
the attribute name as a key (which is a string), their hashes match.
However, even if hashes match, Python needs to make sure those two
are the same in case of a hash collision, so it tries to compare them.
This causes BytesWarning exception when running in strict mode
because `bytes` and `str` instances cannot be compared. KeyError
would be thrown in a non-strict mode.

Also, when later passing the attr to replace_value(), we need for it
to be `str` otherwise the modifications handler fails because it
tries to sort the attributes it's modifying but that's a bit less
poetic issue than the first one.

https://pagure.io/freeipa/issue/7129

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -1434,6 +1434,7 @@ class ModifyLDIF(ldif.LDIFParser):
 
             if "replace" in entry:
                 for attr in entry["replace"]:
+                    attr = attr.decode('utf-8')
                     try:
                         self.replace_value(dn, attr, entry[attr])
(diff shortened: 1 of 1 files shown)


stlaz:
master:

* 0f13e663cac0d29518daa2eb6f1c4da759b90d2a ldif: handle attribute names as strings

PR #1052, opened 2017-09-08
travis: run the same tests in python2/3

stlaz:
We missed running some tests in python3

------------

When meddling with the Travis CI settings, I noticed that we still don't run all the tests in both Python versions. I wish I noticed earlier.

### DIFF
--- a/ipatests/test_xmlrpc/tracker/certmapdata.py
+++ b/ipatests/test_xmlrpc/tracker/certmapdata.py
@@ -1,6 +1,8 @@
 #
 # Copyright (C) 2015  FreeIPA Contributors see COPYING for license
 #
+import base64
+
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend
 from nose.tools import assert_raises
@@ -50,11 +52,9 @@ class CertmapdataMixin(object):
                     certs = [certs]
 
             for cert in certs:
-                cert = x509.load_pem_x509_certificate(
-                    (b'-----BEGIN CERTIFICATE-----\n'
-                     b'{}-----END CERTIFICATE-----\n'
-                     .format(cert)),
-                    default_backend()
+                cert = x509.load_der_x509_certificate(
+                    base64.b64decode(cert),
+                    backend=default_backend()
                 )
                 issuer = DN(cert.issuer).x500_text()
                 subject = DN(cert.subject).x500_text()
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,67 +35,19 @@ env:
         - TASK_TO_RUN="run-tests"
           PYTHON=/usr/bin/python3
           TEST_RUNNER_CONFIG=".test_runner_config_py3_temp.yaml"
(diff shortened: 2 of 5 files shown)


stlaz:
I need to add some builds to our copr repo but I'd like to sort out some infrastructure matters first, namely I'd like to remove Fedora 25 from our PR CI environment.

ipatests/test_xmlrpc/tracker/certmapdata.py
```diff
@@ -50,11 +52,9 @@ def _data_from_options(self, **options):
                     certs = [certs]
 
             for cert in certs:
-                cert = x509.load_pem_x509_certificate(
-                    (b'-----BEGIN CERTIFICATE-----\n'
-                     b'{}-----END CERTIFICATE-----\n'
-                     .format(cert)),
-                    default_backend()
+                cert = x509.load_der_x509_certificate(
```
flo-renaud: The cert is built with a PEM format in ipatests/test_xmlrpc/test_certmap_plugin.py, line 148:  u'certificate': (u'MIICwzCCAau...'), so you probably don't need this change.

stlaz: In that case I would not understand why it was previously built like that. I can try to load `cert` as PEM right away, but applying `.format()` on bytes won't work so at least minor change is needed.

stlaz:
We're blocked by https://bugzilla.redhat.com/show_bug.cgi?id=1491137.

ipatests/test_xmlrpc/tracker/certmapdata.py
stlaz: Hmm, I see you write `u'certificate': (u'MIICwzCCAau...'),`. That, however, is not a PEM formatted cert AFAIK because it's missing the required header/footer. It's just a base64-encoded DER cert.

Do you propose not to convert the cert to cryptography `Certificate` at all? I suppose that as well might work.

stlaz:
Alexander helped with the samba issue and we are now unblocked :)

ipatests/test_xmlrpc/tracker/certmapdata.py
flo-renaud: Hi @stlaz,
I had missed the point about format() on bytes, sorry about that. You are right, your proposal from the PR should work (ie apply base64.d64decode).

flo-renaud:
Hi @stlaz 
thank you, works for me.

stlaz:
@flo-renaud Thank you!

I just realized when pushing that Martin Basti had an old ticket at his commit so I fixed it, the code was not changed.

stlaz:
master:

* 436d2de456c14c869c9970fb53c2931c7c8e67c8 ldap2: don't use decode() on str instance
* c8161fc40cf0ec38612bc346097d6a93972ba032 certmap testing: fix wrong cert construction
* 8be28145bf763ff8dfcf4d03cbd9b196d4482755 py3: set samba dependencies
* 537690ae44b5187721b518b7e304d2abe4fa619c travis: run the same tests in python2/3

PR #1056, opened 2017-09-11
Remove temporary workaround for Travis CI

stlaz:
Once Travis fixes their issue/will post resolution, we can remove the workaround.


stlaz:
https://github.com/travis-ci/travis-ci/issues/8363 is the possible cause of these failures.

tiran:
#1303 addresses the issue in master, 4.6 and 4.5.

PR #1057, opened 2017-09-12
py3: fix ipa cert-request --database ...

flo-renaud:
Fix bytes vs str issues in ipa cert-request

https://pagure.io/freeipa/issue/7148

### DIFF
--- a/ipaclient/csrgen.py
+++ b/ipaclient/csrgen.py
@@ -407,11 +407,11 @@ class OpenSSLAdaptor(object):
         self.password_filename = password_filename
 
     def key(self):
-        with open(self.key_filename, 'r') as key_file:
+        with open(self.key_filename, 'rb') as key_file:
             key_bytes = key_file.read()
         password = None
         if self.password_filename is not None:
-            with open(self.password_filename, 'r') as password_file:
+            with open(self.password_filename, 'rb') as password_file:
                 password = password_file.read().strip()
 
         key = load_pem_private_key(key_bytes, password, default_backend())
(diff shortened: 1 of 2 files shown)


ipaclient/csrgen.py
```diff
@@ -407,11 +407,11 @@ def __init__(self, key_filename, password_filename):
         self.password_filename = password_filename
 
     def key(self):
-        with open(self.key_filename, 'r') as key_file:
+        with open(self.key_filename, 'rb') as key_file:
             key_bytes = key_file.read()
         password = None
         if self.password_filename is not None:
-            with open(self.password_filename, 'r') as password_file:
+            with open(self.password_filename, 'rb') as password_file:
```
stlaz: Do we need to read passwords as bytes?

flo-renaud: Hi @stlaz 
thank you for your comment. Yes, according to https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/#pem the method load_pem_private_key expects bytes for key_bytes and password.

stlaz: Ah, thanks. I'll perform a sanity test and will ACK this afterwards :)

stlaz:
I noticed several other problems with csrgen, I'll get back to this PR after these have been dealt with.

The first problem is that `cert_request` expects csr as a string, there are some other problems with `csrgen_ffi` and generally the `ipalib.pkcs10` may need a similar rework to the one done in the `ipalib.x509` module.

rcritten:
This fixes the immediate problem but I still get no cert and a traceback is thrown when there is no NSS database.

When I manually create a db in /tmp/nssdb I get this:

    ipa: ERROR: Required CSR generation rule syntaxSubject is missing data

stlaz:
It might help adding `--csr-profile-id userCert`, you'll overcome the problem. You'll encounter some more, though. I have a  🚧  man-at-work fix but it's not quite ready for anyone's eyes https://github.com/stlaz/freeipa/commit/62d0e74d4d883d740b56159933f91ee319268da6

stlaz:
I should hope https://github.com/freeipa/freeipa/pull/1110 fixes the Python 3 issues with the `csrgen` module. I included this fix there, we can reopen this PR if necessary.

PR #1062, opened 2017-09-12
dsinstance: Restore context after changing dse.ldif

abbra:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1490762

### DIFF
--- a/ipaserver/install/dsinstance.py
+++ b/ipaserver/install/dsinstance.py
@@ -596,6 +596,7 @@ class DsInstance(service.Service):
                 parser.parse()
             new_dse_ldif.flush()
         shutil.copy2(temp_filename, dse_filename)
+        tasks.restore_context(dse_filename)
         try:
             os.remove(temp_filename)
         except OSError as e:


rcritten:
What isn't obvious to me is what changed. This code has been around since 2015. Was it a fluke that it worked up to now?

abbra:
That's a good question. No, I don't have an answer. But since we have two bugs where this is manifested in two different SELinux contexts, there must be something different now in F27.

rcritten:
LGTM

stlaz:
It's working. I'd ACK it but we really need a ticket to decide which branches it should be pushed into.

lslebodn:
>It's working. I'd ACK it but we really need a ticket to decide which branches it should be pushed into.

Ideally to all branches which introduced this regression (from SELinux POV) :-)

lslebodn:
>It's working. I'd ACK it but we really need a ticket to decide which branches it should be pushed into.

Ideally to all branches which introduced this regression (from SELinux POV) :-)

stlaz:
I created an issue for this: https://pagure.io/freeipa/issue/7150
If you please add it to the ticket, I can at least push this to master + ipa-4-6 for now and we'll decide the other branches on the next devel meeting.

abbra:
Rebased and added upstream issue references.

stlaz:
master:

* 473ddbdb66e563d93a30ac51b1ac559adbd18190 dsinstance: Restore context after changing dse.ldif

PR #1071, opened 2017-09-13
client: fix retrieving certs from HTTP

stlaz:
We're applying bytes regex on the result of a command but were
using decoded stdout instead of raw.


flo-renaud:
Hi @stlaz 
Thank you for the patch, works as expected!

stlaz:
master:

* c4505f080479068db41c1a6ed99945b973cb0134 client: fix retrieving certs from HTTP

PR #1075, opened 2017-09-13
[backport][ipa-4-5] Fixing how sssd.conf is updated when promoting a client to replica

felipevolpone:
When promoting a client to a replica we have to change sssd.conf,
deleting _srv_ part from 'ipa_server' property and setting 'ipa_server_mode' to true.

Previously, the wrong domain could be updated since the ipa_domain variable was not being used properly.

https://pagure.io/freeipa/issue/7127


stlaz:
ACKing based on previous ACK.

stlaz:
ipa-4-5:

* 79570804b0ffaae14b5db11b2d6e45dceff51aba Fixing how sssd.conf is updated when promoting a client to replica

PR #1076, opened 2017-09-13
[backport][ipa-4-5] Changing idoverrideuser-* to treat objectClass case insensitively

felipevolpone:
This is import to avoid problems when migrating from olders
versions of IPA and using idoverrideuser-* commands.

https://pagure.io/freeipa/issue/7074


stlaz:
ACKing based on previous ACK.

stlaz:
ipa-4-5:

* 61e8b4936f1fd73f8d4c359348cf83f37da35fef Changing idoverrideuser-* to treat objectClass case insensitively

PR #1078, opened 2017-09-13
Increase dbus client timeouts during CA install

zultron:
In FreeIPA 4.5 and 4.6, the `ipa-server-install` program can fail during the "Configuring certificate server (pki-tomcatd)" stage on a heavily-loaded host.

The dogtag service is memory intensive, and on a memory-constrained system, can be unresponsive right after start-up, especially for expensive operations that generate new certificates.

These delays can cause failures when dbus clients time out while attempting to run
new certificate operations through certmonger.  This patch changes dbus client timeouts for some such operations to 120 seconds (from the default 25 seconds).

### DIFF
--- a/ipalib/install/certmonger.py
+++ b/ipalib/install/certmonger.py
@@ -31,6 +31,7 @@ import shlex
 import subprocess
 import tempfile
 from ipalib import api
+from ipalib.constants import CA_DBUS_TIMEOUT
 from ipapython.ipa_log_manager import root_logger
 from ipapython.dn import DN
 from ipaplatform.paths import paths
@@ -595,7 +596,9 @@ def modify_ca_helper(ca_name, helper):
         old_helper = ca_iface.Get('org.fedorahosted.certmonger.ca',
                                   'external-helper')
         ca_iface.Set('org.fedorahosted.certmonger.ca',
-                     'external-helper', helper)
+                     'external-helper', helper,
+                     # Give dogtag extra time to generate cert
+                     timeout=CA_DBUS_TIMEOUT)
         return old_helper
 
 
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -29,6 +29,7 @@ import pki.system
 
 from ipalib import api, errors
 from ipalib.install import certmonger
+from ipalib.constants import CA_DBUS_TIMEOUT
 from ipaplatform import services
 from ipaplatform.constants import constants
 from ipaplatform.paths import paths
@@ -261,7 +262,9 @@ class DogtagInstance(service.Service):
                 iface.add_known_ca(
                     name,
                     command,
-                    dbus.Array([], dbus.Signature('s')))
+                    dbus.Array([], dbus.Signature('s')),
+                     # Give dogtag extra time to generate cert
+                    timeout=CA_DBUS_TIMEOUT)
 
     def __get_pin(self):
         try:
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -281,6 +281,8 @@ IPA_CA_CN = u'ipa'
 IPA_CA_RECORD = "ipa-ca"
 IPA_CA_NICKNAME = 'caSigningCert cert-pki-ca'
 RENEWAL_CA_NAME = 'dogtag-ipa-ca-renew-agent'
+# How long dbus clients should wait for CA certificate RPCs [seconds]
+CA_DBUS_TIMEOUT = 120
 
 # regexp definitions
 PATTERN_GROUPUSER_NAME = '^[a-zA-Z0-9_.][a-zA-Z0-9_.-]*[a-zA-Z0-9_.$-]?$'


ipalib/install/certmonger.py
```diff
@@ -595,7 +595,7 @@ def modify_ca_helper(ca_name, helper):
         old_helper = ca_iface.Get('org.fedorahosted.certmonger.ca',
                                   'external-helper')
         ca_iface.Set('org.fedorahosted.certmonger.ca',
-                     'external-helper', helper)
+                     'external-helper', helper, timeout=120)
```
felipevolpone: It would be nice to have a code comment here, explaining briefly why we are setting the timeout to this value.

ipaserver/install/dogtaginstance.py
```diff
@@ -261,7 +261,7 @@ def configure_certmonger_renewal(self):
                 iface.add_known_ca(
                     name,
                     command,
-                    dbus.Array([], dbus.Signature('s')))
+                    dbus.Array([], dbus.Signature('s')), timeout=120)
```
felipevolpone: It would be nice to have a code comment here, explaining briefly why we are setting the timeout to this value.

ipalib/install/certmonger.py
zultron: Done.

ipaserver/install/dogtaginstance.py
zultron: Done.  Thanks for the review.

stlaz:
I actually don't think this change will change much. The crash we're seeing is pretty deterministic, really. Do go ahead and try:
```bash
domain=IPA.TEST; docker run -t --name freeipa-popelnice -h $HOSTNAME \
                         --tmpfs /run --tmpfs /tmp --security-opt seccomp:unconfined \
                         -v /dev/urandom:/dev/random:ro -v /opt/ipa-data/master:/data \
                         -v /sys/fs/cgroup:/sys/fs/cgroup:ro --cap-add=SYS_TIME \
                         -e DEBUG_TRACE=1 -e DEBUG_NO_EXIT=1 freeipa:fedora-26 \
                        --hostname $HOSTNAME --domain $domain --realm ${domain^^} \
                        -p milan_je_buh123 -a milan_je_buh123 --setup-dns --auto-forwarders --no-reverse -U
```
It should fail on `pkispawn` during a `pki-tomcat` start up. Dogtag seems to be trying to set up a kernel keyring but is denied.

I am currently trying to get some more info from the Dogtag guys.

zultron:
@stlaz, you're right, this PR doesn't address the problem you're pointing out.  (I work around that problem for now with `docker run --privileged`.)

This PR addresses dbus timeouts breaking the install.  This can happen on memory-constrained systems, where any swapping after dogtag startup can exceed the default 25 second dbus default timeout.

rcritten:
I don't see a downside in increasing the timeout whether it's for memory constrained systems, slow disks or whatever. Fast systems will never hit it so it's a moot point in that case. The 120 seconds seems arbitrary though, how did you come up with it? Should there be a define somewhere, DBUS_TIMEOUT, and use that instead? I think I'd want a comment about the default 25 seconds (or whatever it is) as well to leave a bread crumb for future devs.

zultron:
@rcritten, the 120 seconds is arbitrary, you're correct.  On a system that swaps pretty heavily during dogtag startup, the certmonger request was observed to take 61 seconds to complete, so I chose a value around 2x.  Do you have another suggestion for determining the timeout?

I'm fine with adding the additional comments and plumbing in a `DBUS_TIMEOUT` parameter of some sort.

Thanks for the review and suggestions.

stlaz:
I agree with @rcritten that this should be a constant. Personally, I would probably have it configurable as well.

rcritten:
It can get tricky to make this sort of thing configurable because normally you'd put this sort of value into /etc/ipa/default.conf but since there is no config yet...

I'd be ok with hardcoding it. There is precedence for that elsewhere.

Normally I'd like to see some sort of "Waiting" message somewhere in the case of log timeouts so users don't think something is hung up but in this case I don't see an obvious place to put it. I think for the average user this is fine given AFAIK this timeout hasn't come up before. On underpowered h/w I'd hope that the user would expect things to take forever.

zultron:
Please see the updated PR.

I admit to being a bit out of my depth here, and wasn't sure if `ipalib.constants.DEFAULT_CONFIG['dbus_certmonger_timeout']` placed near other timeouts would be better than e.g. `ipalib.constants.DBUS_CERTMONGER_TIMEOUT` by itself.

stlaz:
The `DEFAULT_CONFIG` dictionary serves for initializing the `api`. You want to put your constant among the others outside the `DEFAULT_CONFIG`.

zultron:
Thanks, @stlaz, PR updated.

rcritten:
LGTM, tested ok in master (after minor rebase). Will wait for add'l reviews before giving final ack.

stlaz:
If @rcritten tested this, it's ok with me. ACK.

tomaskrizek:
@stlaz Please make sure all the tests executed and green before acking :) PR CI wasn't triggered before.

The commit message is also missing an upstream ticket -- do we want to create one or do we just push this to master only?

zultron:
@tomaskrizek, I'd like this to go into a 4.6 release, if possible.  I went ahead and created [issue #7213] on pagure (with the assumption that that's where "upstream" is), and added that URL to the commit log (it's [pushed][2], but there seems to be some lag in updating this PR).

[1]:  https://pagure.io/freeipa/issue/7213
[2]: https://github.com/zultron/freeipa/tree/ipa-4-5-dbus-timeouts

stlaz:
We'll add this to the master and backport to the `ipa-4-6` branch. I did not notice tests were failing, sorry.

tomaskrizek:
@zultron Thanks for creating the issue and linking it!

tomaskrizek:
I've re-based the commit against `master` and opened #1170 for it. We'll also backport it to `ipa-4-6`. I'm closing this PR since it's opened against `ipa-4-5`.

Thanks for your contribution, I apologize for the bureaucracy :)

PR #1079, opened 2017-09-14
ipa-server-upgrade: fix the logic for tracking certs

flo-renaud:
ipa-server-upgrade needs to configure certmonger with the right options
in order to track PKI, HTTP and LDAP certs (for instance the RA agent cert
location has changed from older releases).
The upgrade code looks for existing tracking requests with the expected
options by using criteria (location of the NSSDB, nickname, CA helper...)
If a tracking request is not found, it means that it is either using wrong
options or not configured. In this case, the upgrade stop tracking
all the certs, reconfigures the helpers, starts tracking the certs so that
the config is up-to-date.

The issue is that the criteria is using the keyword 'ca' instead of
'ca-name' and this leads to upgrade believing that the config needs to be
updated in all the cases.

https://pagure.io/freeipa/issue/7151

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -925,7 +925,7 @@ def certificate_renewal_update(ca, ds, http):
         {
             'cert-database': paths.PKI_TOMCAT_ALIAS_DIR,
             'cert-nickname': 'auditSigningCert cert-pki-ca',
-            'ca': 'dogtag-ipa-ca-renew-agent',
+            'ca-name': 'dogtag-ipa-ca-renew-agent',
             'cert-presave-command': template % 'stop_pkicad',
             'cert-postsave-command':
                 (template % 'renew_ca_cert "auditSigningCert cert-pki-ca"'),
@@ -933,7 +933,7 @@ def certificate_renewal_update(ca, ds, http):
         {
             'cert-database': paths.PKI_TOMCAT_ALIAS_DIR,
             'cert-nickname': 'ocspSigningCert cert-pki-ca',
-            'ca': 'dogtag-ipa-ca-renew-agent',
+            'ca-name': 'dogtag-ipa-ca-renew-agent',
             'cert-presave-command': template % 'stop_pkicad',
             'cert-postsave-command':
                 (template % 'renew_ca_cert "ocspSigningCert cert-pki-ca"'),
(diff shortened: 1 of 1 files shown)


rcritten:
Are you sure about the property name? I see no ca-name defined in certmonger, but it does know about ca, in src/tdbus.h: #define CM_DBUS_PROP_CA "ca"

flo-renaud:
Yes, there is special code in https://pagure.io/freeipa/blob/master/f/ipalib/install/certmonger.py#_179 to handle the ca name.

rcritten:
I'm still seeing it fail to find the request for 'caSigningCert cert-pki-ca' and therefore reconfiguring the tracking again. It succesfully found a few others including subsystem and ocsp (and perhaps others, I wasn't paying that close attention).

flo-renaud:
Hi @rcritten 
I can't manage to reproduce what you are seeing. Which version did you install before launching the upgrade?

frasertweedale:
@rcritten @flo-renaud if there is a problem with the criteria for `caSigningCert cert-pki-ca`, IMO it can be treated as a separate issue from this one.

(OTOH a drive-by fix would be fine too, but since @flo-renaud doesn't have a reproducer yet I don't think we need to hold up this fix).

rcritten:
@flo-renaud I'm just re-running ipa-server-upgrade and every time it resets the certmonger tracking. Running from master in rawhide.

frasertweedale:
I think this patch has an undesirable interaction with #1045 (apart from the merge conflict).
If HTTP/DS service certs are not issued by IPA CA, we do not track them.  But the list of `requests` in `certificate_renewal_update` includes the HTTP and DS service certs.

So if they are (legitimately) not tracked, it does not find a tracking request, then all the tracking requests get removed and replaced.

rcritten:
@frasertweedale re: tackling 'caSigningCert cert-pki-ca' later, if the point is to not re-create the tracking with each upgrade then I think it's relevant. I can try to reproduce again with a clean install if desired.

rcritten:
I manually re-based and re-tested and it is working as expected on a stock install in rawhide.

@frasertweedale has a point about the Apache and 389-ds certs. @flo-renaud already has code to find nicknames from 87540fe1ef8a191e521ddf1584b4cbebb7dece94 . Perhaps that should be integrated. If the nickname is not Server-Cert then don't add it to the requests list.

flo-renaud:
@rcritten Thank you for your time re-testing.
@frasertweedale Thank you for the comment about service certs not issued by IPA CA. I added code to handle this case.

flo-renaud:
master:

* 73b20975698d449229e5f7ac2eecff7da135822d ipa-server-upgrade: fix the logic for tracking certs
* d87163c29096e04553b0e1fcf7c1753e8a6cf716 ipa-server-upgrade: do not add untracked certs to the request list

PR #1081, opened 2017-09-14
OTP import: support hash names with HMAC- prefix

abbra:
Refactor convertHashName() method to accept hash names prefixed with
HMAC- or any other prefix. Extending the method should be easier in
future.

Fixes https://pagure.io/freeipa/issue/7146

### DIFF
--- a/ipaserver/install/ipa_otptoken_import.py
+++ b/ipaserver/install/ipa_otptoken_import.py
@@ -95,7 +95,9 @@ def convertTokenType(value):
 def convertHashName(value):
     "Converts hash names to their canonical names."
 
-    return {
+    default_hash = u"sha1"
+    known_prefixes = ("", "hmac-",)
+    known_hashes = {
         "sha1":    u"sha1",
         "sha224":  u"sha224",
         "sha256":  u"sha256",
@@ -106,7 +108,24 @@ def convertHashName(value):
         "sha-256": u"sha256",
         "sha-384": u"sha384",
         "sha-512": u"sha512",
-    }.get(value.lower(), u"sha1")
+    }
+
+    if value is None:
+        return default_hash
+
+    v = value.lower()
+    for prefix in known_prefixes:
+        if prefix:
+            w = v[len(prefix):]
+        else:
+            w = v
+        result = known_hashes.get(w)
+        if result is not None:
+            break
+    else:
+        result = default_hash
+
+    return result
 
 
 def convertHMACType(value):
--- a/ipatests/test_ipaserver/test_otptoken_import.py
+++ b/ipatests/test_ipaserver/test_otptoken_import.py
@@ -22,6 +22,7 @@ import os
 import pytest
 
 from ipaserver.install.ipa_otptoken_import import PSKCDocument, ValidationError
+from ipaserver.install.ipa_otptoken_import import convertHashName
 
 basename = os.path.join(os.path.dirname(__file__), "data")
 
@@ -129,3 +130,21 @@ class test_otptoken_import(object):
                 'ipatokenotpdigits': 8,
                 'type': u'hotp',
             })]
+
+    def test_valid_tokens(self):
+        assert convertHashName('sha1') == u'sha1'
+        assert convertHashName('hmac-sha1') == u'sha1'
+        assert convertHashName('sha224') == u'sha224'
+        assert convertHashName('hmac-sha224') == u'sha224'
+        assert convertHashName('sha256') == u'sha256'
+        assert convertHashName('hmac-sha256') == u'sha256'
+        assert convertHashName('sha384') == u'sha384'
+        assert convertHashName('hmac-sha384') == u'sha384'
+        assert convertHashName('sha512') == u'sha512'
+        assert convertHashName('hmac-sha512') == u'sha512'
+
+    def test_invalid_tokens(self):
+        """The conversion defaults to sha1 on unknown hashing"""
+        assert convertHashName('something-sha256') == u'sha1'
+        assert convertHashName('') == u'sha1'
+        assert convertHashName(None) == u'sha1'


ipaserver/install/ipa_otptoken_import.py
```diff
@@ -106,7 +107,21 @@ def convertHashName(value):
         "sha-256": u"sha256",
         "sha-384": u"sha384",
         "sha-512": u"sha512",
-    }.get(value.lower(), u"sha1")
+    }
+
+    v = value.lower()
+    for prefix in known_prefixes:
+        if len(prefix) > 0:
```
rcritten: Just a nit, I think this is equivalent and more efficient:

if prefix:

abbra: Right, this is just cleaner as well. I've pushed an update.

abbra:
Somehow, simple replication test failed. I'm sure my changes aren't related to that one at all.

rcritten:
How about some test cases:

    #
    # Copyright (C) 2017  FreeIPA Contributors see COPYING for license
    #
    """
    Tests for the `ipaserver.install/ipa_otptoken_import` module.
    """
    
    from ipaserver.install import ipa_otptoken_import
    import pytest
    
    pytestmark = pytest.mark.tier0
    
    def test_valid_tokens():
        assert ipa_otptoken_import.convertHashName('sha1') == u'sha1'
        assert ipa_otptoken_import.convertHashName('hmac-sha1') == u'sha1'
        assert ipa_otptoken_import.convertHashName('sha224') == u'sha224'
        assert ipa_otptoken_import.convertHashName('hmac-sha224') == u'sha224'
        assert ipa_otptoken_import.convertHashName('sha256') == u'sha256'
        assert ipa_otptoken_import.convertHashName('hmac-sha256') == u'sha256'
        assert ipa_otptoken_import.convertHashName('sha384') == u'sha384'
        assert ipa_otptoken_import.convertHashName('hmac-sha384') == u'sha384'
        assert ipa_otptoken_import.convertHashName('sha512') == u'sha512'
        assert ipa_otptoken_import.convertHashName('hmac-sha512') == u'sha512'
    
    def test_invalid_tokens():
        """The conversion defaults to sha1 on unknown hashing"""
        assert ipa_otptoken_import.convertHashName('something-sha256') == u'sha1'
        assert ipa_otptoken_import.convertHashName('') == u'sha1'
        assert ipa_otptoken_import.convertHashName(None) == u'sha1'

Note that the last test fails because None is not allowed. It fails on v.lower(). Should it be allowed?

abbra:
That's a good part (last test) but failure in https://fedorapeople.org/groups/freeipa/prci/jobs/3764e65e-9965-11e7-9e9b-fa163eb0eb3c/report.html is test_integration/test_simple_replication.py::TestSimpleReplication::()::test_user_replication_to_master

rcritten:
Yeah our comments passed each other in the night ...

I'm not worried about the replication failure. I can't see how that could possibly be related to this change.

abbra:
I've updated a bit to explicitly check for None value and set out default hash together with other definitions. This should be enough to cover all cases. However, the tests you pointed to aren't even executed anywhere or otherwise asserts would caught up 'hmac-sha224' already. Below is an output from the original code, without my changes:

```python
# python
Python 2.7.13 (default, Jun 26 2017, 10:20:05) 
[GCC 7.1.1 20170622 (Red Hat 7.1.1-3)] on linux2
>>> from ipaserver.install import ipa_otptoken_import
>>> ipa_otptoken_import.convertHashName('hmac-sha224')
u'sha1'
>>> def test_valid_tokens():
...     assert ipa_otptoken_import.convertHashName('sha1') == u'sha1'
...     assert ipa_otptoken_import.convertHashName('hmac-sha1') == u'sha1'
...     assert ipa_otptoken_import.convertHashName('sha224') == u'sha224'
...     assert ipa_otptoken_import.convertHashName('hmac-sha224') == u'sha224'
...     assert ipa_otptoken_import.convertHashName('sha256') == u'sha256'
...     assert ipa_otptoken_import.convertHashName('hmac-sha256') == u'sha256'
...     assert ipa_otptoken_import.convertHashName('sha384') == u'sha384'
...     assert ipa_otptoken_import.convertHashName('hmac-sha384') == u'sha384'
...     assert ipa_otptoken_import.convertHashName('sha512') == u'sha512'
...     assert ipa_otptoken_import.convertHashName('hmac-sha512') == u'sha512'
... 
>>> def test_invalid_tokens():
...     """The conversion defaults to sha1 on unknown hashing"""
...     assert ipa_otptoken_import.convertHashName('something-sha256') == u'sha1'
...     assert ipa_otptoken_import.convertHashName('') == u'sha1'
...     assert ipa_otptoken_import.convertHashName(None) == u'sha1'
... 
>>> test_valid_tokens()
  File "<stdin>", line 5, in test_valid_tokens
AssertionError
>>> test_invalid_tokens()
  File "<stdin>", line 5, in test_invalid_tokens
  File "/usr/lib/python2.7/site-packages/ipaserver/install/ipa_otptoken_import.py", line 105, in convertHashName
    }.get(value.lower(), u"sha1")
AttributeError: 'NoneType' object has no attribute 'lower'
>>> 
```

rcritten:
def test_valid_tokens():
            assert ipa_otptoken_import.convertHashName('sha1') == u'sha1'
            assert ipa_otptoken_import.convertHashName('hmac-sha1') == u'sha1'
            assert ipa_otptoken_import.convertHashName('sha224') == u'sha224'
    >       assert ipa_otptoken_import.convertHashName('hmac-sha224') == u'sha224'
    E       AssertionError: assert 'sha1' == 'sha224'
    E         - sha1
    E         + sha224

It fails for me as expected without your patch.

As for None, I could go either way on whether you want to drop that test or add a None test in the code.

abbra:
I did add handling of None it the last push, so now it should be covering all the cases.

abbra:
As discussed on IRC, I added tests created by @rcritten  above to this changeset. Hopefully, we wouldn't regress anymore.

rcritten:
The None test needs to test value not v:

    >       if v is None:
    E       UnboundLocalError: local variable 'v' referenced before assignment

stlaz:
master:

* 8661611d3e1d63a01fdc250610a87b3813e5125f OTP import: support hash names with HMAC- prefix

PR #1090, opened 2017-09-18
Don't write p11-kit EKU extension object if no EKU

stlaz:
b5732efd introduced a regression because it tries to write EKU
that's actually in the CA cert instead of using the LDAP information.
However, when no EKU is available,
IPACertificate.extended_key_usage_bytes still returned at least
EKU_PLACEHOLDER OID to keep the behavior the same as in previous
versions. This caused the EKU_PLACEHOLDER to be written in the
ipa.p11-kit file which made Firefox report FreeIPA Web UI as
improperly configured.

https://pagure.io/freeipa/issue/7119

### DIFF
--- a/ipalib/x509.py
+++ b/ipalib/x509.py
@@ -295,8 +295,11 @@ class IPACertificate(object):
 
     @property
     def extended_key_usage_bytes(self):
+        eku = self.extended_key_usage
+        if eku is None:
+            return
+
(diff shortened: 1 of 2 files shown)


stlaz:
master:

* e537686bcc8248bc0216ce634ae7707fa65e70ba Don't write p11-kit EKU extension object if no EKU

PR #1097, opened 2017-09-19
Backport 4-5: Fix ipa-server-upgrade with server cert tracking

flo-renaud:
ipa-server-upgrade fails with Server-Cert not found, when trying to
track httpd/ldap server certificates. There are 2 issues in the upgrade:
- the certificates should be tracked only if they were issued by IPA CA
(it is possible to have CA configured but 3rd part certs)
- the certificate nickname can be different from Server-Cert

The fix provides methods to find the server crt nickname for http and ldap,
and a method to check if the server certs are issued by IPA and need to be
tracked by certmonger.

https://pagure.io/freeipa/issue/7141

Reviewed-By: Fraser Tweedale <ftweedal@redhat.com>

### DIFF
--- a/ipaserver/install/certs.py
+++ b/ipaserver/install/certs.py
@@ -42,6 +42,7 @@ from ipapython.certdb import get_ca_nickname, find_cert_from_txt, NSSDatabase
 from ipapython.dn import DN
 from ipalib import pkcs10, x509, api
 from ipalib.errors import CertificateOperationError
+from ipalib.install import certstore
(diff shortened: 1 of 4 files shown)


stlaz:
ACKing based on a previous ACK.

stlaz:
ipa-4-5:

* 52853875e298e38a1e5a9a56c02aac9e30916044 Backport 4-5: Fix ipa-server-upgrade with server cert tracking

PR #1099, opened 2017-09-19
rpc: don't decode cookie_string if it's None

stlaz:
This removes an ugly debug message from client installation

### DIFF
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -901,6 +901,8 @@ class RPCClient(Connectible):
         # (possibly with more than one cookie).
         try:
             cookie_string = read_persistent_client_session_data(principal)
+            if cookie_string is None:
+                return
             cookie_string = cookie_string.decode('utf-8')
         except Exception as e:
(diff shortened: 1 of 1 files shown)


frasertweedale:
@stlaz is there a ticket associated with this patch?  What are the step to reproduce
this issue that this patch addresses?

stlaz:
I did not create a ticket as the change seemed small to me, although it might have been worth it.

I did see this in client installation when `ipa-server-install` was run (message like "'NoneType' has no 'decode` member."). Note that the failure is non-fatal due to the bare Exception, it's just that it probably should not appear in the logs as it may sound like something bad could have actually happened.

frasertweedale:
@stlaz thanks, I see the error in the client install log.

When I have confirmed that this patch avoids it, you shall have my ACK.

tomaskrizek:
master:

* 5acd4840903c7f290c1af7d6442e90bf4c0d19dd rpc: don't decode cookie_string if it's None

PR #1103, opened 2017-09-20
ipatests: keep default log level for 389-ds

tomaskrizek:
The log level of 8192 for 389-ds is very noisy and conceals useful debug
messages. The default log level 0 should be sufficient for our purposes.

Part of: https://pagure.io/freeipa/issue/7162

### DIFF
--- a/ipatests/pytest_plugins/integration/tasks.py
+++ b/ipatests/pytest_plugins/integration/tasks.py
@@ -236,14 +236,14 @@ def restore_hostname(host):
         host.run_command(['rm', backupname])
 
 
-def enable_replication_debugging(host):
-    logger.info('Enable LDAP replication logging')
+def enable_replication_debugging(host, log_level=0):
+    logger.info('Set LDAP debug level')
     logging_ldif = textwrap.dedent("""
         dn: cn=config
         changetype: modify
         replace: nsslapd-errorlog-level
-        nsslapd-errorlog-level: 8192
-        """)
+        nsslapd-errorlog-level: {log_level}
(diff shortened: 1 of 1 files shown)


tomaskrizek:
@tbordaz Could you please review/ack? You can check any of the ds logs and verify the debug messages aren't there.

tomaskrizek:
I agree it's better to keep the code. I've changed to PR to use log_level of 0 by default, but left the option to change it.

rcritten:
LGTM too

tomaskrizek:
For some reason, it doesn't work properly. The replication messages are still in the ds error log, for example, see https://fedorapeople.org/groups/freeipa/prci/jobs/2df3b8ee-9eaa-11e7-8c70-fa163e566bda/test_integration-test_simple_replication.py-TestSimpleReplication--test_user_replication_to_replica/master.ipa.test/var/log/dirsrv/slapd-IPA-TEST/errors.gz

Firstyear:
Hmmm, maybe there is a test elsewhere that is causing this change? It looks like the loglevel is set correctly and the server initial is not logging, but suddenly it changes. You could use the DS audit log to try and track who triggered the change of the loglevel, but I think this patch and what it is "now" is good to merge.

tomaskrizek:
Based on comments above, I'm re-adding the ack. I'll keep the ticket open until the replication messages are no longer in ds log.

tomaskrizek:
master:

* be6f1a67cbf3e08ee0749dc1fe158f585e5546b7 ipatests: set default 389-ds log level to 0

PR #1109, opened 2017-09-22
Collect group membership without a size limit

rcritten:
If the # of group memberships exceeded the search size limit
then SizeLimitExceeded was raised. Being in too many groups
should not cause a *_show to fail.

https://pagure.io/freeipa/issue/7112

### DIFF
--- a/ipaserver/plugins/baseldap.py
+++ b/ipaserver/plugins/baseldap.py
@@ -715,7 +715,9 @@ class LDAPObject(Object):
             result = self.backend.get_entries(
                 self.api.env.basedn,
                 filter=filter,
-                attrs_list=[''])
+                attrs_list=[''],
+                size_limit=-1,  # paged search will get everything anyway
+                paged_search=True)
         except errors.NotFound:
             result = []
 


ipaserver/plugins/baseldap.py
```diff
@@ -715,7 +715,7 @@ def get_memberofindirect(self, entry):
             result = self.backend.get_entries(
                 self.api.env.basedn,
                 filter=filter,
-                attrs_list=[''])
+                attrs_list=[''], size_limit=0)
```
stlaz: A few lines above this the very same is done using `size_limit=-1, paged_search=True` so I'd rather that we stayed consistent.

rcritten:
Nice catch, done.

ipaserver/plugins/baseldap.py
```diff
@@ -715,7 +715,9 @@ def get_memberofindirect(self, entry):
             result = self.backend.get_entries(
                 self.api.env.basedn,
                 filter=filter,
-                attrs_list=[''])
+                attrs_list=[''],
+                size_limit=-1, # paged search will get everything anyway
```
stlaz: `./ipaserver/plugins/baseldap.py:719:31: E261 at least two spaces before inline comment`
For some reason Travis did not report this, I may need to check why.

stlaz:
Works great, please fix the pep8 issue and I will ACK this.

stlaz:
Thanks, ACK!

pvomacka:
master:

* 418421d94182f03aa9af38f7ab9ed141828f1eb2 Collect group membership without a size limit

PR #1110, opened 2017-09-25
csrgen: Introduce CertificateSigningRequest parameter

stlaz:
This pull request tries to introduce a new parameter for CSRs so that these are handled properly in Python 3 and are backward-compatible with the older API which will try to send strings to the server.

**NOTE**: tests for the new parameter are currently not included, it's here for the sole purpose of a review of the code and should not be pushed until the tests arrive.

- [x] do the code
- [x] do the tests

-----------------------
This PR includes code @flo-renaud's fix from https://github.com/freeipa/freeipa/pull/1057


flo-renaud:
Hi @stlaz,
Thank you for the patch. I noticed the following issues when using IDM web GUI:
- when generating a new cert for a user or a service (Users > Actions > New certificate), the operation fails with IPA Error 4015: Base64DecodeError - Base64 decoding failed: Incorrect padding
- when generating a new cert for a host (Hosts > Actions > New certificate), the operation fails with IPA Error 4301: CertificateOperationError - Certificate operation cannot be completed: Failure decoding Certificate Signing Request: Unable to load request

When using the cli:
- ipa cert-request --principal <principal> test.csr fails with ipa: ERROR: Certificate operation cannot be completed: Failure decoding Certificate Signing Request: Unable to load request
- ipa cert-request --principal <principal> --database /tmp/nssdb shows a warning:
/usr/lib/python3.6/site-packages/ipaclient/csrgen.py:464: BytesWarning: str() on a bytes instance
  ['certutil', '-S', '-n', self.nickname, '-s', 'CN=%s' % temp_cn,
(because temp_cn is bytes and conflict with %s)
- ipa cert-request --principal <principal> --private-key displays:
/usr/lib/python3.6/site-packages/ipaclient/csrgen_ffi.py:209: UserWarning: implicit cast from 'char *' to a different pointer type: will be forbidden in the future (check that the types are as you expect; use an explicit ffi.cast() if they are correct)
  subj, rdn_type, MBSTRING_UTF8, v.value, -1, -1, mval):

stlaz:
Thank you for your review, I'll fix this ASAP.

stlaz:
Thank you for the review!
This latest version should fix the WebUI CSR signing, I did not realize we were actually sending to the API what one would expect we would be sending, that's great.
I'll see about the other errors.

flo-renaud:
Hi @stlaz,
your latest version fixes the web gui issue, thank you.

stlaz:
This last update should fix all the issues mentioned except for the `UserWarning`, I need to see about that.

Previously I made some silly changes which I reverted, and I forgot to pick one of the changes you did in your patch for which I'm sorry.

The `UserWarning` was not introduced as a part of these patches but should be fixed here, too.

stlaz:
The `UserWarning` should now also be fixed.

flo-renaud:
Hi @stlaz 
the patch works for me. I am waiting for the new tests before giving the ACK.

stlaz:
Added the tests for the new parameter.

flo-renaud:
Hi @stlaz 
thank you for the tests, works for me. simple_replication tests are failing but it is completely unrelated to your commit (same failure seen with different PRs), so giving a ACK.

stlaz:
Unfortunately, it seems that in the case of replica installation, the failure is caused by this PR as certmonger sends us a request with a CSR in the form of a base64-encoded DER (without sending the information it's base64-encoded data as is usual in cases bytes are being passed to the API) so I have to scratch what I said in https://github.com/freeipa/freeipa/pull/1110#issuecomment-332132123 about API being sent what one would expect ☹️
I will need to fix this on the FreeIPA side and add a test about that I'm afraid.

rcritten:
What should certmonger be sending that it isn't? Is the API not enforcing some requirement?

stlaz:
So, from the point of view of the previous implementation, what `certmonger` sends us is kind of right. 

Previously, someone (and I am not pointing fingers, I don't care) thought that it would be a good idea to have the CSR argument as a `Str`. As far as I understand, it would be because on the client-side, we're passing it a `File` parameter which can't read files as bytes but is rather trying to decode them right away. This blocks us from passing DER-formatted files to `cert-request` and similar commands - for eternity (unless, you know, we introduce a new parameter for byte files which will confuse the users completely).

Now to my point about unexpected data - `Str` representation is not correct when you're sending us base64-encoded bytes, `Bytes` would be the way to go. In that case, you would be sending us something like:
```
"params": [
        [],
        {
            "csr": {
                "__base64__": "<THE_BASE64_ENCODED_DER_CERT>"
            },
            "version": "2.229"
        }
    ]
}
```
but that collides with the old `Str` implementation, I believe. It might be OK to just receive the whole PEM-formatted cert as a `Str`, we'd be able to work with that.

Can I work around this behavior? Yes.
Do I want to? No.
Do I have to? Most probably yes. I will implement this the way it was working previously, just chipping away the header and footer from whatever we receive and will try to base64-decode the rest.

A fun fact, btw: the very same code works on master but fails on replica at this step:
```
Configuring directory server (dirsrv)
  [1/3]: configuring TLS for DS instance
  [error] RuntimeError: Certificate issuance failed (CA_UNREACHABLE)
```
I am not going to be spelunking for the cause, though, and will just learn to live with it.

stlaz:
This latest patchset should fix the issue during replica installation. Note the changes in https://github.com/freeipa/freeipa/pull/1110/commits/cd8f0213ca259a3222e54342d6a1b53a534d44ea and addition of a testing data in https://github.com/freeipa/freeipa/pull/1110/commits/2d7299ea3c3ab4b277d84fe5182a82f525a8a5b0.

ipaserver/install/certs.py
```diff
@@ -85,6 +85,23 @@ def export_pem_p12(pkcs12_fname, pkcs12_pwd_fname, nickname, pem_fname):
                  "-passout", "file:" + pkcs12_pwd_fname])
 
 
+def strip_csr_header(csr):
+    """
```
flo-renaud: I did not find a place where this function is used.

stlaz: Ah, you're right, I removed it but it got back in a rebase accident, I will remove it.

stlaz:
Removed the appearance of a redundant function + rebased on more current master.

ipaserver/install/certs.py
```diff
@@ -412,9 +412,6 @@ def issue_server_cert(self, certreq_fname, cert_fname):
         with open(certreq_fname, "r") as f:
             csr = f.read()
 
-        # We just want the CSR bits, make sure there is nothing else
-        csr = pkcs10.strip_header(csr)
-
```
flo-renaud: If you remove this part, Dogtag will error out failing to parse the CSR.

stlaz:
The latest run was affected by the KDC errors we've seen previously, logs available at: https://fedorapeople.org/groups/freeipa/prci/jobs/99319500-b4de-11e7-93a6-fa163eb69f7e/
Re-running.

flo-renaud:
Hi @stlaz ,
thank you for the patch. Works as expected, but you need to rebase and resolve conflicts with ipatests/test_pkcs10/test_pkcs10.py.

stlaz:
Rebased on current master.

flo-renaud:
The CI ests exited on timeout, thus I'm adding the re-run label.
Apart from that, everything works as expected. Waiting for PR CI to finish successfully in order to give the ACK label

flo-renaud:
All tests are green, ACK.

stlaz:
ipatool did not like the rebase as it was missing some commits from upstream master, rebasing again so that the PR can be pushed.

stlaz:
I had to push this manually, for some reason 0008-Remove-pkcs10-module-contents.patch would appear as empty to git, even though it wasn't. Closing.

PR #1111, opened 2017-09-25
ipalib.x509 autumn cleanup

stlaz:
This is a cleanup of the x509 module which removes an ugly hack that prevents a circular dependency and the `strip_header()` function which is properly replaced by the `load_unknown_x509_certificate()` function where needed.

### DIFF
--- a/ipaclient/plugins/cert.py
+++ b/ipaclient/plugins/cert.py
@@ -209,6 +209,7 @@ class cert_find(MethodOverride):
                 raise errors.MutuallyExclusiveError(
                     reason=_("cannot specify both raw certificate and file"))
             if 'certificate' not in options and 'file' in options:
-                options['certificate'] = x509.strip_header(options.pop('file'))
+                options['certificate'] = x509.load_unknown_x509_certificate(
+                                            options.pop('file'))
 
         return super(cert_find, self).forward(*args, **options)
--- a/ipaclient/plugins/certmap.py
+++ b/ipaclient/plugins/certmap.py
@@ -40,7 +40,7 @@ class certmap_match(MethodOverride):
                 raise errors.MutuallyExclusiveError(
                     reason=_("cannot specify both raw certificate and file"))
             if args:
-                args = [x509.strip_header(args[0])]
+                args = [x509.load_unknown_x509_certificate(args[0])]
             elif 'certificate' in options:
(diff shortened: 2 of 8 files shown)


Tiboris:
Thanks @stlaz !

rcritten:
No ticket(s)?

x509 shouldn't be moved from ipalib to ipapython until there is a discussion about how/if the API is supported.

stlaz:
This is a code cleanup, the changes are explained in the commit messages, I don't think it needs a ticket.

`x509` is not moved from `ipalib` to `ipapython` here at all. I have another branch for that where I do that rather poorly (it does not work) but I can make it better and make it work if we wanted to do that switch. It would seem only natural because our `x509` module does not really need to depend on `ipalib` at all. That should not be discussed here, though, if you want, we can talk about that on IRC, get a ticket, discuss it on devel meeting and get it done.

Also, if you don't want this pushed, please, remove the ACK label.

rcritten:
You should remove the references to moving the code from the commit message. It isn't relevant to this change.

It was my understanding that policy was that all changes required a ticket in order for proper planning, scoping and to enhance the history of what changed and why.

The changes themselves seem ok otherwise.

stlaz:
I will remove the reference, thanks for pointing that out, it shouldn't really be there.

I believe that some changes, like cleanup actions, should not require a ticket if the changes are explained in the commit messages clearly. That's the best way of keeping the history of the changes, anyway.
Please bear in mind that we already have 1400+ tickets for about 10 developers. That is a clear message that we fail in a very horrible and gruesome manner as far as planning goes, however sad this information is (to me).

rcritten:
The use of tickets is probably better for a broader audience but they are important for planning, scoping and just knowing what others are working on. This doesn't mean unplanned work cannot be done but it needs to be agreed upon work otherwise something else could slip. This will become more important once sprints are used.

The number of opened issues is not relevant to the question of whether a commit should be associated to one.

stlaz:
I asked the team to either close or ACK this last week. They did neither, so I am re-adding the original ACK and pushing the change.

stlaz:
master:

* c9265a7b0528b7f2495bf15e79cd9dd96af578aa x509: remove the strip_header() function
* 4cc94512719be2addf3c90f7a4913709a4ed0f0d x509: remove subject_base() function

PR #1115, opened 2017-09-28
ipa-kdb: reinit trusted domain data for enterprise principals

sumit-bose:
While processing enterprise principals the information about trusted domains
might not be up-to-date. With this patch ipadb_reinit_mspac() is called if an
unknown domain is part of the enterprise principal.

Resolves https://pagure.io/freeipa/issue/7172


tomaskrizek:
master:

* fe1aad767948f4636c2251b0cf7b6a25f9bf7145 ipa-kdb: reinit trusted domain data for enterprise principals

PR #1116, opened 2017-09-28
Enable ephemeral KRA requests

rcritten:
Enabling ephemeral KRA requests will reduce the amount of LDAP
write operations and improve overall performance.

https://pagure.io/freeipa/issue/6703

NOTE: I'm not 100% sure on the upgrade for existing instances. My logic is that tomcat is always stopped and then within that block the CA (if any) will be updated. Given that the KRA runs in the same service that is why I stuck that update code there. It worked in my testing.

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -11,19 +11,8 @@ import shutil
 import pwd
 import fileinput
 import sys
-
 from augeas import Augeas
 import dns.exception
-
-import six
-# pylint: disable=import-error
-if six.PY3:
-    # The SafeConfigParser class has been renamed to ConfigParser in Py3
-    from configparser import ConfigParser as SafeConfigParser
-else:
-    from ConfigParser import SafeConfigParser
-# pylint: enable=import-error
-
 from ipalib import api
 from ipalib.install import certmonger, sysrestore
 import SSSDConfig
@@ -44,6 +33,7 @@ from ipaserver.install import ntpinstance
 from ipaserver.install import bindinstance
 from ipaserver.install import service
 from ipaserver.install import cainstance
+from ipaserver.install import krainstance
 from ipaserver.install import certs
 from ipaserver.install import otpdinstance
 from ipaserver.install import schemaupdate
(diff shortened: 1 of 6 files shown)


rcritten:
CI seems to be failing with DNS issues. Re-running again to see if I'll get lucky.

abbra:
@frasertweedale  any news with this review?

ipaserver/install/server/upgrade.py
```diff
@@ -1692,11 +1695,26 @@ def upgrade_configuration():
             )
         upgrade_pki(ca, fstore)
 
+        if kra.is_configured():
+            logger.info('[Ensuring ephemeralRequest is enabled in KRA]')
+            cainstance.backup_config(path=paths.KRA_CS_CFG_PATH)
```
frasertweedale: I think it would be more straightforward to have `krainstance.backup_config()` instead of `cainstance.backup_config(new_arg_with_default_value)`.  The common bits can be extracted to a method in the shared `DogtagInstance` base class.

ipaserver/install/server/upgrade.py
```diff
@@ -1692,11 +1695,26 @@ def upgrade_configuration():
...
+            else:
+                logger.info('ephemeralRequest is already enabled')
+
     # several upgrade steps require running CA.  If CA is configured,
     # always run ca.start() because we need to wait until CA is really ready
     # by checking status using http
     if ca.is_configured():
         ca.start('pki-tomcat')
+    if kra.is_configured() and not kra.is_running():
+        kra.start('pki-tomcat')
```
frasertweedale: The CA and KRA (when configured) are both run under `pki-tomcatd@pki-tomcat.service` systemd service.  So the `ca.start('pki-tomcat')` will also start the KRA.  So I think this addition is redundant.

frasertweedale:
Works for me; a couple of small suggestions / questions.

ipaserver/install/server/upgrade.py
rcritten: I laid the groundwork for this in a separate patch. The changes are small but required significant reformatting to get past the pep8 checks.

ipaserver/install/server/upgrade.py
rcritten: I realize this, I did it for future-proofing in case the KRA service ever runs independently. I've left it in for now, I can remove it if you insist.

It should basically always fail.

frasertweedale: OK, I'm happy for it to be left in, but please mention the above-stated reason in the commit message.

frasertweedale:
ACK on this.  But note that pull/1347 also deals with the `set_directive` long lines
to deal with PEP-8 violations, and whichever of these lands first will mean a rebase
for the other.

If #1347 gets merged first I'll do the rebase of this one, if you like.

tiran:
@rcritten Please rebase this branch on latest master. I don't want to run into another issue was stale tests again. Thx

tiran:
This PR will cause massive merge conflicts with PR #1379. I'm holding off until #1379 has been merged.

tiran:
I created PR #1405 to fix the merge conflict.

PR #1118, opened 2017-09-29
tests_py3: decode get_file_contents() result

Rezney:
When running tests in python3 we get bytes object instead of
bytestring and when passing it to multihost run_command we later
fail on concatenation in shell_quote().

### DIFF
--- a/ipatests/pytest_plugins/integration/tasks.py
+++ b/ipatests/pytest_plugins/integration/tasks.py
@@ -228,7 +228,7 @@ def restore_files(host):
 def restore_hostname(host):
     backupname = os.path.join(host.config.test_dir, 'backup_hostname')
     try:
-        hostname = host.get_file_contents(backupname)
+        hostname = host.get_file_contents(backupname, encoding='utf-8')
     except IOError:
         logger.debug('No hostname backed up on %s', host.hostname)
     else:


felipevolpone:
@Rezney hi, how can I validate this patch? There is some case where an exception is raised or something like that? It fixes which scenario? Thanks

Rezney:
@felipevolpone You can run "test_client_reenrollment" test suite to verify this. The issue happens during the teardown:

`test_integration/base.py:79: in uninstall
    tasks.uninstall_master(cls.master)
pytest_plugins/integration/tasks.py:714: in uninstall_master
    unapply_fixes(host)
pytest_plugins/integration/tasks.py:194: in unapply_fixes
    restore_hostname(host)
pytest_plugins/integration/tasks.py:236: in restore_hostname
    host.run_command(['hostname', hostname.strip()])
../pytest_multihost/host.py:243: in run_command
    command.stdin.write(shell_quote(arg))
`

I also adjusted the commit message a little.

abbra:
LGTM

stlaz:
master:

* 3b5e97933432dc8a1685cb34a51aa7c14fc13625 tests_py3: decode get_file_contents() result

PR #1119, opened 2017-10-02
ipa-cacert-manage renew: switch from ext-signed CA to self-signed

flo-renaud:
The scenario switching from externally signed CA to self-signed CA is
currently failing because the certmonger helper goes through the wrong
code path when the cert is not self-signed.

When the cert is not self-signed but the admin wants to switch to self-signed
a new cert needs to be requested, not retrieved from LDAP.

https://pagure.io/freeipa/issue/7173


frasertweedale:
Works for me.

tomaskrizek:
master:

* 49cf5ec64b1b7a7437ca285430353473c215540e ipa-cacert-manage renew: switch from ext-signed CA to self-signed

PR #1120, opened 2017-10-03
tests: Mark failing tests as failing

celestian:
Some tests from installation suite assume that it is possible
to --setup-kra, respectively --setup-ca if there is no kra,
respectively ca on IPA master.

This patch marks those tests as failing.

Signed-off-by: Petr Čech <pcech@redhat.com>

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -111,6 +111,7 @@ class ADTrustInstallTestBase(IntegrationTest):
 # Master X Replicas installation tests
 ##
 
+@pytest.mark.xfail(reason="FreeIPA ticket 7008")
 class TestInstallWithCA1(InstallTestBase1):
 
     @classmethod
@@ -133,6 +134,7 @@ class TestInstallWithCA1(InstallTestBase1):
         super(TestInstallWithCA1, self).test_replica2_ipa_dns_install()
 
 
+@pytest.mark.xfail(reason="FreeIPA ticket 7008")
 class TestInstallWithCA2(InstallTestBase2):
 
     @classmethod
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_installation.py
```diff
@@ -165,13 +168,15 @@ def test_replica0_ipa_kra_install(self):
         tasks.install_kra(self.replicas[0], first_instance=False)
 
 
+@pytest.mark.xfail(reason="FreeIPA ticket 7008")
```
Tiboris: Here installs only master. According to [7008](https://pagure.io/freeipa/issue/7008) should not be an issue.

ipatests/test_integration/test_installation.py
```diff
@@ -229,6 +235,7 @@ def test_replica0_ipa_kra_install(self):
...
```
Tiboris: Same as above, test installs only master. Should not be an issue.

celestian:
So, thanks for review. I fixed it and I run it again in my PR CI, you can see progress here:
https://github.com/celestian/freeipa/pull/2

celestian:
I pushed new version of the patch. Unfortunately I am not able to run my PR CI with installation suite today. I will run it as soon as possible.

Tiboris:
Thanks @celestian.

Could you please check inline comment? There is still one unresolved yet.

celestian:
All test from test suite are green now:
https://github.com/celestian/freeipa/pull/8

Tiboris:
Great! Thank you @celestian.

Tiboris:
master:

* cd80036b6b4fdfc843aac52126f01fdb1fc1067d tests: Mark failing tests as failing

PR #1121, opened 2017-10-03
Remove the `message` attribute from exceptions

stlaz:
This is causing python2 tests print ugly warnings about the
deprecation of the `message` attribute in python2.6.

====================
I had some time between reviews when the tests were running and noticed some ugly strings in python2 test results so I'm posting this.

### DIFF
--- a/ipalib/errors.py
+++ b/ipalib/errors.py
@@ -369,7 +369,7 @@ class ServerCommandError(PublicError):
     For example:
 
     >>> e = CommandError(name='foobar')
-    >>> raise ServerCommandError(error=e.message, server='https://localhost')
+    >>> raise ServerCommandError(error=str(e), server='https://localhost')
     Traceback (most recent call last):
       ...
     ServerCommandError: error on server 'https://localhost': unknown command 'foobar'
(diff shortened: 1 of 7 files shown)


stlaz:
@flo-renaud The easiest way to reproduce the issue I am solving here is to run:
`ipa-run-tests test_ipalib/test_errors.py`
You should be able to see errors similar to https://travis-ci.org/freeipa/freeipa/jobs/282742357 (check the `Test runner output:` section).

flo-renaud:
Hi @stlaz 
thank you for the patch and the procedure to reproduce the issue, works correctly. There is one occurrence of e.message that was missed, in ipatests/test_webui/test_user.py line 386:
```
        with pytest.raises(AssertionError) as e:
            self.add_record(user.ENTITY, user.DATA3)
        assert e.value.message == u'...'
```

stlaz:
@flo-renaud thanks for noticing!

flo-renaud:
@stlaz,
thank you, works for me.

stlaz:
master:

* af1b8513abc8e5be74e2886f9048c747235f1a14 Remove the `message` attribute from exceptions

PR #1122, opened 2017-10-03
 travis: make tests fail if pep8 does not pass 

stlaz:
Currently, pep8 does not seem to fail even though there may be errors, let's see if this helps it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,7 @@ install:
 script:
     - mkdir -p $CI_RUNNER_LOGS_DIR
     - travis_wait 50 ./.travis_run_task.sh
+    - test -z "`cat $PEP8_ERROR_LOG`"
 after_failure:
     - echo "Test runner output:"; tail -n $CI_BACKLOG_SIZE $CI_RESULTS_LOG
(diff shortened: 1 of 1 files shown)


rcritten:
I gather this means success given all the Travis tests failed? The other failures are unrelated AFAICT.

stlaz:
The Travis tests failed because the command I put there was not recognized, should have probably used `test` directly. I'll need to do that differently, though, as `git diff master -U0 | pep8 --diff &>$LOG` will still write EOL to the $LOG file.
Probably a `cat` of the log file should suffice.

rcritten:
Ah, I see it now:

[-s pep8_errors.log]: command not found

You may want to switch the test to -z if the file will always be touched to avoid using !.

stlaz:
pep8 now fails correctly, let's just test that breaking only tests will also make travis fail.

stlaz:
Travis correctly failed in that case as well, can be safely reviewed now.

stlaz:
master:

* 209bb2771260589d2dba690aa3eab750069d6803 travis: make tests fail if pep8 does not pass

PR #1125, opened 2017-10-03
Check if replica-s4u2proxy.ldif should be applied

felipevolpone:
Now, before applying replica-s3u2proxy.ldif, we check if the values are already there. The values can be
there, if a replica installation was done in the past and some info was left behind.

https://pagure.io/freeipa/issue/7174

### DIFF
--- a/ipaserver/install/dsinstance.py
+++ b/ipaserver/install/dsinstance.py
@@ -958,7 +958,24 @@ class DsInstance(service.Service):
         self._ldap_mod("replica-prevent-time-skew.ldif", self.sub_dict)
 
     def __setup_s4u2proxy(self):
-        self._ldap_mod("replica-s4u2proxy.ldif", self.sub_dict)
+
+        def __add_principal(last_cn, principal, self):
+            dn = DN(('cn', last_cn), ('cn', 's4u2proxy'),
+                    ('cn', 'etc'), self.suffix)
+
+            value = '{principal}/{fqdn}@{realm}'.format(fqdn=self.fqdn,
+                                                        realm=self.realm,
+                                                        principal=principal)
+
+            entry = api.Backend.ldap2.get_entry(dn, ['memberPrincipal'])
+            try:
+                entry['memberPrincipal'].append(value)
+                api.Backend.ldap2.update_entry(entry)
+            except errors.EmptyModlist:
+                pass
+
+        __add_principal('ipa-http-delegation', 'HTTP', self)
+        __add_principal('ipa-ldap-delegation-targets', 'ldap', self)
 
     def __create_indices(self):
         self._ldap_mod("indices.ldif")


rcritten:
What about the off-chance that only one of the values exist? Is that possible?

felipevolpone:
To be honest, I don't know :) I was hoping that someone could tell me that. If it is possible, we should also split the ldif file in two.

Also, will be an output with `enabling S4U2Proxy delegation` even if nothing will be done. Is that a problem?

rcritten:
I'm not sure there is a need to do this via LDIF. It may be possible to do with direct LDAP commands and catch and ignore DuplicateEntry errors.

pvoborni:
Could we consider an alternative approach? Identify several objects which are created when replica exists  and are removed on server-del. On next replica installation check for those in early stages and if the objects exists exit with error message stating that one must run `ipa-replica-manage del $replica --force --clean` for DL 0 and `ipa server-del --force` for DL 1.

rcritten:
That works it just isn't very user-friendly, having the command halted. It is certainly a safer way to go though as it will require the user to examine the current state, and a similar test is already being done for host existence (which is why I know it isn't user friendly ;-) )

So it's a choice of:

1. Force the user to do proper clean-up
2. Do the clean-up on our own (we can internally call these I think)
3. Handle only this one case of lack-of-cldanup.

felipevolpone:
TBH I would go with the case we have now. We know how many different ways a user can have a broken environment...

pvoborni:
Some background of the issue. This might happen in situation:
* `ipa-replica-install ...`
* `ipa-replica-install --uninstall`
*  `ipa-replica-install ...`

I.e. explicit `ipa server-del` or `ipa-replica-manage del` are missing in the process.

 `ipa server-del` is run by  `ipa-replica-install --uninstall` but often fails to replica the changes because it effectively cut's a branch under itself (destroys replication credentials) therefore there are left overs which then causes issues like this. 

After looking at it again, +1 to the approach and if the patch works, we can go with it. But I'm afraid it won't solve the situation as the similar issue but elsewhere in the code might happen - different LDAP record not deleted.

pvoborni:
Btw we cannot properly cleanup in uninstaller yet because uninstaller runs as directory manager on itself(LDAPI) but only with host credentials on other replica (GSSAPI). Host credentials doesn't have the permissions to delete every record.

tomaskrizek:
@felipevolpone Could you please rebase the PR against current master? There's no conflict, but the ipatool fails (probably due to a 3-way merge).

stlaz:
Author or reviewers ( @rcritten @pvoborni  ): please clean up your mess (`install/share/replica-s4u2proxy.lidf`)

pvoborni:
@stlaz Could you write it more constructively and avoid using words like mess? From your comment I have no idea what should be cleaned or why.

stlaz:
@pvoborni of course. Please remove the aforementioned file which is not used anywhere anymore as a result of this fix.

felipevolpone:
PR https://github.com/freeipa/freeipa/pull/1240 removes the file.

PR #1112, opened 2017-09-25
ipatests: use python3 if built with python3

stlaz:
Change the default python version for test scripts

https://pagure.io/freeipa/issue/7131

PR #1130, opened 2017-10-04
prci: bump f26 template to 0.1.5

tomaskrizek:


stlaz:
I'll check this template with https://github.com/freeipa/freeipa/pull/1112 since that's the reason we're adding it, anyway.

stlaz:
Fails with unknown option `html`, the package added in 0.1.4 was removed in 0.1.5.

stlaz:
We have newer templates now, will be added in https://github.com/freeipa/freeipa/pull/1112

PR #1131, opened 2017-10-04
Less confusing message for PKINIT configuration during install

slaykovsky:
Pagure issue: https://pagure.io/freeipa/issue/7179

### DIFF
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -502,8 +502,17 @@ class KrbInstance(service.Service):
             self._install_pkinit_ca_bundle()
             self.pkinit_enable()
         except RuntimeError as e:
-            logger.error("PKINIT certificate request failed: %s", e)
-            logger.error("Failed to configure PKINIT")
+            logger.warning("PKINIT certificate request failed: %s", e)
+            logger.warning("Failed to configure PKINIT")
+
+            self.print_msg("Full PKINIT configuration did not succeed")
+            self.print_msg(
+                "The setup will only install bits "
+                "essential to the server functionality")
+            self.print_msg(
+                "You can enable PKINIT after the "
+                "setup completed using 'ipa-pkinit-manage'")
+
             self.stop_tracking_certs()
             self.issue_selfsigned_pkinit_certs()
 


ipaserver/install/krbinstance.py
```diff
@@ -502,8 +502,11 @@ def issue_ipa_ca_signed_pkinit_certs(self):
             self._install_pkinit_ca_bundle()
             self.pkinit_enable()
         except RuntimeError as e:
-            logger.error("PKINIT certificate request failed: %s", e)
-            logger.error("Failed to configure PKINIT")
+            self.print_msg("Full PKINIT configuration was not succeed")
+            logger.debug(str(e))
```
stlaz: I think `e` will be converted to string anyway, will it not? Also, I think it makes sense to have the error message as a part of the warning.

ipaserver/install/krbinstance.py
```diff
@@ -502,8 +502,11 @@ def issue_ipa_ca_signed_pkinit_certs(self):
...
+            self.print_msg(
+                "Local PKINIT has been configured with "
```
stlaz: Use `logger.warning()`.

s/has been/was/

ipaserver/install/krbinstance.py
```diff
@@ -502,10 +502,16 @@ def issue_ipa_ca_signed_pkinit_certs(self):
...
-            logger.error("PKINIT certificate request failed: %s", e)
-            logger.error("Failed to configure PKINIT")
+            logger.warning("PKINIT certificate request failed: %s", e)
+            logger.warning("Failed to configure PKINIT")
+
+            self.print_msg("Full PKINIT configuration did not succeed")
+
             self.stop_tracking_certs()
             self.issue_selfsigned_pkinit_certs()
+            self.print_msg(
```
stlaz: The feature will not be functional after this fallback. Could we maybe instead of this add a message "The setup will only install bits essential to the server functionality" after the "Full PKINIT configuration did not succeed". You can also add a hint on how to enable PKINIT after the setup completed.
The reason for this is that with the current message, I still get a feeling that PKINIT is available at least locally for me (whatever someone might think that means).

slaykovsky: Thanks!

stlaz:
ACK, thanks!

tomaskrizek:
master:

* 7a3da27816f62240cebbac72de5c98a96ace0794 Less confusing message for PKINIT configuration during install

tomaskrizek:
@slaykovsky Please backport this patch to `ipa-4-5` branch manually. There's been a logger refactoring in 4.6 version.

slaykovsky:
@tomaskrizek thanks!

PR #1133, opened 2017-10-05
[4.5] Use correct container for ipa-4-5 testing

stlaz:
@freeipa/freeipa-master no longer provides Fedora 25 dependencies
but is still used in the current container for testing ipa-4-5.
Since ipa-4-5 won't pass F26 pylint check we can't use the master
branch container, thus we'll use an ad hoc one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
 cache: pip
 env:
     global:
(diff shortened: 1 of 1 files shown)


stlaz:
ipa-4-5:

* 7544e71fdf5a250dc161c1a5755c0854360fdb35 Use correct container for ipa-4-5 testing

PR #1136, opened 2017-10-06
ipaclient.plugins.dns: Cast DNS name to unicode.

slaykovsky:
cmd.api.Command.dnsrecord_split_parts expects name to be unicode
string and instead gets ascii. It leads to an error:
ipa: ERROR: invalid 'name': must be Unicode text

This commit's change is casting name's type to unicode so
'ipa dnsrecord-mod' will not fail with error above.

https://pagure.io/freeipa/issue/7185

### DIFF
--- a/ipaclient/plugins/dns.py
+++ b/ipaclient/plugins/dns.py
@@ -62,7 +62,7 @@ def __get_part_param(rrtype, cmd, part, output_kw, default=None):
 def prompt_parts(rrtype, cmd, mod_dnsvalue=None):
     mod_parts = None
     if mod_dnsvalue is not None:
-        name = record_name_format % rrtype.lower()
+        name = record_name_format % unicode(rrtype.lower())
         mod_parts = cmd.api.Command.dnsrecord_split_parts(
             name, mod_dnsvalue)['result']
 


ipaclient/plugins/dns.py
```diff
@@ -62,7 +62,7 @@ def __get_part_param(rrtype, cmd, part, output_kw, default=None):
 def prompt_parts(rrtype, cmd, mod_dnsvalue=None):
     mod_parts = None
     if mod_dnsvalue is not None:
-        name = record_name_format % rrtype.lower()
+        name = unicode(record_name_format % rrtype.lower())
```
Tiboris: You should use `decode` method on rrtype to prevent:
_BytesWarning: str() on a bytes instance_
And you should also remove unicode() call.

slaykovsky: Thanks, done.

ipaclient/plugins/dns.py
```diff
@@ -62,7 +62,7 @@ def __get_part_param(rrtype, cmd, part, output_kw, default=None):
 def prompt_parts(rrtype, cmd, mod_dnsvalue=None):
     mod_parts = None
     if mod_dnsvalue is not None:
-        name = record_name_format % rrtype.lower()
+        name = record_name_format % rrtype.lower().decode("utf-8")
```
Tiboris: This will unfortunately throw an attribute error:
`AttributeError: 'str' object has no attribute 'decode'`

Tiboris: @stlaz mentioned he has already seen similar issue.

slaykovsky: @Tiboris 4.5 is python 2.7, right? So it will not throw any attribute error.

Tiboris: @slaykovsky That is true! Sorry.
Have you tried to reproduce this bug on master?

slaykovsky: @Tiboris on master it's okay :)

Tiboris:
I have restarted travis task... again DNS issues.

ipaclient/plugins/dns.py
Tiboris: Should not we be at least `.decode('utf-8')` entire `name` and not 'only' `rrtype`?

And one not very relevant but to stay consistent could you please use 'utf-8' used everywhere in code so far, instead of "utf-8" ? 😉

slaykovsky: > Should not we be at least .decode('utf-8') entire name and not 'only' rrtype?
After formatting, `name` will become utf-8 string, e.g.:
```
>>> a = "a %s"
>>> type(a)
<type 'str'>
>>> a % "b".decode("utf-8")
u'a b'
```

> use 'utf-8' instead of "utf-8"
Sure :)

stlaz: "4.5 is python 2.7" is a not-good-enough reasoning not to fix this properly (meaning compatible with python3 - such incompatibilities often show poor design choices).

ipaclient/plugins/dns.py
```diff
@@ -62,7 +62,7 @@ def __get_part_param(rrtype, cmd, part, output_kw, default=None):
 def prompt_parts(rrtype, cmd, mod_dnsvalue=None):
     mod_parts = None
     if mod_dnsvalue is not None:
-        name = record_name_format % rrtype.lower()
+        name = record_name_format % unicode(rrtype.lower())
```
Tiboris: As you discussed with Standa, this, is a proper fix. Thanks, sorry for confusion

slaykovsky: @Tiboris thanks, mate!

Tiboris:
ipa-4-5:

* c3f390d2264f71cf4b87123fdcabd007075704ab ipaclient.plugins.dns: Cast DNS name to unicode.

PR #1137, opened 2017-10-09
ipa-kdb: support KDB DAL version 7.0

frozencemetery:
krb5-1.16 includes DAL version 7, which changes the signature of
audit_as_req to include local and remote address parameters.

This patch just enables building against the new DAL version, but
doesn't use the new information for anything.

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb.c
+++ b/daemons/ipa-kdb/ipa_kdb.c
@@ -709,7 +709,9 @@ kdb_vftabl kdb_function_table = {
 };
 #endif
 
-#if (KRB5_KDB_DAL_MAJOR_VERSION == 6) && defined(HAVE_KDB_FREEPRINCIPAL_EDATA)
+#if ((KRB5_KDB_DAL_MAJOR_VERSION == 6) || \
+     (KRB5_KDB_DAL_MAJOR_VERSION == 7)) && \
+    defined(HAVE_KDB_FREEPRINCIPAL_EDATA)
 kdb_vftabl kdb_function_table = {
     .maj_ver = KRB5_KDB_DAL_MAJOR_VERSION,
     .min_ver = 1,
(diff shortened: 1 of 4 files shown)


abbra:
The code changes are good, I'm not sure what causes caless and simple_replication runs to fail though as KDC seems to work well there.

tomaskrizek:
I'm doing a build for rawhide with this patch: https://koji.fedoraproject.org/koji/taskinfo?taskID=22365738

I also had to bump the version to 7.0 in https://github.com/freeipa/freeipa/blob/master/freeipa.spec.in#L90 Should this be bumped in a separate patch later on, or in this one?

The simple_replication failure was most likely unrelated, it seems to be cause be the `kinit admin` issue. I've restarted the test.

frozencemetery:
Thanks Tomas!  I've added the spec bump to this patch.

tomaskrizek:
@frozencemetery I can't see the spec file in this patch. Are you sure you've added it?

frozencemetery:
Whoops, have now, thanks!

tomaskrizek:
Don't worry about the failing tests -- we need to add a new krb5 build to our master COPR and it should work afterwards. I hope to get to it soon.

tomaskrizek:
I've slightly [modified the spec](https://github.com/freeipa/freeipa/pull/1137/files#diff-021f8f69d8c0598dcead3341b439283aR89) so we can keep using the old kdb dal version in f26,27 and only require the 7.0 in rawhide. This should allow us to keep developing and testing with krb5-1.15.

frozencemetery:
Sure, this is fine and won't break the build for f27+.

tomaskrizek:
master:

* 9f8700fceead6e7b4947dc86f161e78dabb5d186 ipa-kdb: support KDB DAL version 7.0

PR #1132, opened 2017-10-05
[tests] CA-less testcase fixes

Akasurde:
* Do not copy filenames from negative testcases such as 'does-not-exist'
* Add interactive 'yes'

Fixes: https://pagure.io/freeipa/issue/7182

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

PR #1142, opened 2017-10-11
ipatests: Fix interactive prompt in ca_less tests

Akasurde:
This fix adds additional prompt which was missing previously
in test_interactive_missing_ds_pkcs_password and
test_interactive_missing_http_pkcs_password under CA-less integration
testsuite.

Fixes: https://pagure.io/freeipa/issue/7182

### DIFF
--- a/ipatests/test_integration/test_caless.py
+++ b/ipatests/test_integration/test_caless.py
@@ -65,6 +65,9 @@ def get_install_stdin(cert_passwords=()):
 
 def get_replica_prepare_stdin(cert_passwords=()):
     lines = list(cert_passwords)  # Enter foo.p12 unlock password
+    lines += [
+        'yes',  # Continue [no]?
+    ]
     return '\n'.join(lines + [''])
 
 


pvoborni:
How can this fix ticket 7182 if the same description is written in PR #1143 ? I assume it fixes only a part. So the commit message is incorrect. Also ticket 7182 doesn't really describe what needs to be fixed, so basically it has unlimited scope which makes he hard do asses if all is fixed.

Akasurde:
@pvoborni Actually, this PR was a part of #1132 
I thought rather than fixing 7182, in single PR, I created two PR. 

I filed 7182 as a tracking bug, as I didn't know how or what needs to be fixed with respect to this ticket. I will update 7182 with these two issues and then we can take this forward.

Akasurde:
@pvoborni I have update pagure issue 7182 with my findings.

Rezney:
@Akasurde Hi, your fix looks good. Just please lets get rid of those square brackets and make the commit header something like: "tests: fix interactive prompt stdin". Thanks!

Akasurde:
@Rezney Ok.

stlaz:
master:

* 222cef1a4d55f8d311d01a51bc3e01e1a64aa337 ipatests: Fix interactive prompt in ca_less tests

PR #1143, opened 2017-10-11
[tests] Fix negative tests in CA-less testsuite

Akasurde:
This fix adds check for non-existing files which are given
by negative testcases to copy from controller to system under tests.
This fixes test_nonexistent_ds_pkcs12_file and test_nonexistent_http_pkcs12_file
tests in CA-less testsuite.

Fixes: https://pagure.io/freeipa/issue/7182


Akasurde:
Does anyone know what is cause of failure in 2929.2 in travis-ci ?

rcritten:
This looks ok but where can I find where it was failing before so  I can see that this fixes it?

Rezney:
Closing this PR as CA-less integration tests are now passing in both master and 4.5.

PR #1144, opened 2017-10-11
Ignore ACI errors when pwpolicy-del fails in group-del

germanparente:
The definitive fix to deal with the DS CVE would be to remove the ipaPermTargetFilter: (objectclass=<objectclass value>) from the "Remove <Object>" permissions during ipa-server-upgrade, for instance, and re-run the task to generate the acl's (if this not done automatically when modifying the permission entry).

this has been also discussed in BZ1441262

https://bugzilla.redhat.com/show_bug.cgi?id=1441262

### DIFF
--- a/ipaserver/plugins/group.py
+++ b/ipaserver/plugins/group.py
@@ -367,7 +367,10 @@ class group_del(LDAPDelete):
         assert isinstance(dn, DN)
         try:
             api.Command['pwpolicy_del'](keys[-1])
-        except errors.NotFound:
+        # we catch ACI error because of CVE fixed in DS
+        # when an entry does not exist and we have no privilege reading it,
+        # there's err=50 instead of err=32
+        except (errors.NotFound, errors.ACIError):
             pass
 
         return True


rcritten:
Please reformat the commit message.

The first line should be something like: ignore ACI errors fwhen pwpolicy-del fails in group-del

The rest of the line(s) should be broken into segments no longer than 65-70 characters per-line rather than one single long line.

The ticket needs to be added to the commit message.

If you know which CVE fix caused this then that would be helpful to note.

pep8 error: ./ipaserver/plugins/group.py:356:32: E231 missing whitespace after ','

It might be nice to note that the pwpolicy will be orphaned by including:

self.add_message(...)

Using the GER call can_delete() may be another way to handle this as well rather than a try/except.

germanparente:
Thanks for your comments.
In fact, there's never an orphan password policy. The issue here is that we try to delete an associated password policy even if it does not exist. And when the entry does not exist, we don't have control access to that entry because the aci for permission checks objectclass of the entry as filter. So, the error sent is ACIError instead of not found.
can_delete could be an alternative but not sure if the function needs the entry to exist to check the effective controls.

rcritten:
There can be an orphan if the user has only group management:

$ kinit admin
$ ipa pwpolicy-add test --priority 2
$ ipa group-add test
$ kinit someuser
<someuser is already in group management privilege>
$ ipa group-del test
$ ipa pwpolicy-find

test pwpolicy still exists, user legitimately doesn't have permission.

What I'd do instead of adding the ACIerror as part of the NotFound exception add it on its own.

except NotFound:
    pass
except ACIError:
    detect_if_pwpolicy_exists
    if exists:
        self.add_message(messages.SomeNewType)

The detection can be either pwpolicy_show, GER can_delete, policy_dn = self.api.Object.pwpolicy.get_dn_if_exists(keys[-1]))

Last one is probably best way to do it though with all of these there is a bit of a race given the pwpolicy could be added at the moment of the delete.

germanparente:
Hi,
Yes, I agree. The granularity oif permissions can lead to unexpected situations like this one.
We give permissions to delete groups but not to delete the associated password policy and we could finish to have these inconsistencies. The user, when defining roles and policies, cannot know that having permissions to delete groups implies that, if he has not also permissions to delete policies, the deletion of a group can leave orphan ones.

When some entities are dependent, imho, we should not let define such granularity.

Rob, I let you commit the patch that you have written to solve this issue. Nice one !

Thanks !

rcritten:
:-)

I haven't really worked out all the details yet but we should take advantage of the messaging system since we can.

I also don't _think_ an orphaned password policy would be an issue, either in re-adding the group or in having a ghost policy enforced, but it's something to look at.

As a cynic, I imagine worst case where some group/pwpolicy is created and after a bit just the group is deleted, then the group re-added much later and magically some ancient pwpolicy is reincarnated to much surprise!

germanparente:
In that case, a try-delete of password policy associated could also be done at group creation. That will cleaned orphans. But the issue is having rights to create/delete groups and not having rights to delete password policy. That could provoke ... surprises. :-)

rcritten:
Surprises. An excellent point.

I wonder if instead IPA should refuse to delete a group if there is an associated password policy that can't be removed.

Let me bring this up to discussion on the team meeting to try to get some consensus on the best approach.

germanparente:
I think the problem of permissions could provoke something worse. You could not even know that a password policy exists ? The server, after the CVE fixed, will return always ACIError, not knowing if the entry exists or it does not exists and you have no rights to read it. What we need to fix in these cases, is that if two entities are related, we cannot have rights to read/write/delete one and not the other.

rcritten:
I haven't completely thought it through but it should be possible, just might be expensive.

We could, for example, we can try to get the pwpolicy entry see if it exists, then GER can_delete to see if it is removable, then make decisions from that. It adds two LDAP calls though which is not great.

germanparente:
Well, there's still the issue that the user that is deleting a group has no "Read Password Policies" permission. In that case, that user can never know if it exists. The ldapsearch will not return 32, just 50 (ACIError).

rcritten:
Yeah, so I brought this up with the other devs and we want to tackle this in two stages.

For this ticket/BZ use GER to determine if the user can delete both the group and the policy and if they can then do so. If not then log a detailed message and return an ACI error to the end-user. This is so we can avoid disclosure that the password policy exists at all. The detailed log message should help the administrator figure out what is going on.

I'll file a new ticket to look at rewriting password policy to use the managed entries plugin to make a tighter link between groups and policies.

rcritten:
The password policy redesign ticket is https://pagure.io/freeipa/issue/7216

PR #1146, opened 2017-10-11
rpcserver.py code cleaning

slaykovsky:
Hi!

I was going through `rpcserver` module code and just found out it's very hard to read. So I've made some style changes and string formatting related changes.

For my point of view, it's now more easy to read.

Any comments are welcomed.

Thanks.

### DIFF
--- a/ipaserver/rpcserver.py
+++ b/ipaserver/rpcserver.py
@@ -1,7 +1,7 @@
 # Authors:
 #   Jason Gerard DeRose <jderose@redhat.com>
 #
-# Copyright (C) 2008-2016  Red Hat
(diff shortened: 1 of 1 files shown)


rcritten:
Reformatting text outside the context of another change is discouraged because it ends up polluting git history.

slaykovsky:
Hi @rcritten , let's say then I've changed `%` string formatting style to `format` and as a side effect I've reformatted the whole module.

rcritten:
Probably rejected for unrelated changes.

reformatting makes things like git blame less useful.

Style changes could be brought up to the team to be decided as a whole.

slaykovsky:
@rcritten ok, will close.

PR #1147, opened 2017-10-12
tests: correct usage of host.hostname in logger in tasks

Akasurde:
This fix adds correct usage of host.hostname in logger.

Fixes: https://pagure.io/freeipa/issue/7190


rcritten:
Should have a ticket describing the original problem.

Drop the [] in the commit message and make the topic more specific, e.g. tests: correct usage of host.hostname in logger in tasks

The change itself looks reasonable.

Akasurde:
@rcritten OK. I will do that.

tomaskrizek:
master:

* c8dbd0cfbe583eead478a3fe073c1ca52b2bb7bd tests: correct usage of hostname in logger in tasks

PR #1148, opened 2017-10-12
Use namespace-aware meta importer for ipaplatform

tiran:
Instead of symlinks and build-time configuration the ipaplatform module
is now able to auto-detect platforms on import time. The meta importer
uses the platform 'ID' from /etc/os-releases. It falls back to 'ID_LIKE'
on platforms like CentOS, which has ID=centos and ID_LIKE="rhel fedora".

The meta importer is able to handle namespace packages and the
ipaplatform package has been turned into a namespace package in order to
support external platform specifications.

https://fedorahosted.org/freeipa/ticket/6474


rcritten:
I know we used the term fallback previously but I wonder if some would prefer a "preferred" option instead, such that the build release is inserted into the array in position 0.

Rather than hand-parsing os-release could you use configparser by pre-pending a fake section like:

from six.moves.configparser import RawConfigParser
from six.moves import StringIO

parser = RawConfigParser()
with open("/etc/os-release") as stream:
    stream = StringIO("[file]\n" + stream.read())
    parser.readfp(stream)

print (parser.get('file', 'id'))

tiran:
The configparser module isn't well suited for the job. I would still need to remove quotes. I'm now using a regular expression with backref.

stlaz:
Did not check the code yet, I am reading "how do we parse os-release", an obvious answer is http://augeas.net/docs/references/1.4.0/lenses/files/shellvars-aug.html.
Dunno if that helps, just posting it here.

tiran:
The format of ``/etc/os-release`` is well-defined and restricted. We only need the ``ID`` and ``ID_LIKE`` fields from ``/etc/os-release``. The fields can be trivially extracted with a straight forward regular expression.

At the moment only ipaserver's installer uses augeas. I don't want to make all clients and consumers of Python wheels require another library and extension module. Is augeas even supported on BSD, macOS, and Windows?

stlaz:
Once https://github.com/freeipa/freeipa/pull/919 is merged, python-augeas will become a dependency of the ipaclient package anyway. Windows probably does not support augeas, FreeBSD and MacOS should be able to.

tiran:
@stlaz No, python-augeas is not becoming a dependency of ipaclient wheel. It's only used in ``ipaclient.install`` subpackage, which is not part of the ``ipaclient`` wheel.

rcritten:
I'm having a tough time getting tox tests to pass. Right now they all fail.

One failure is related to importing ipaserver and here is why:

$ find .tox -name ipaserver
.tox/pypi/lib/python3.6/site-packages/ipaserver

rcritten:
Re-tested in a fresh tree and the tests are passing now. Waiting for PR-CI to re-check.

abbra:
LGTM. I'm adding ACK since there is a clear consensus.

tiran:
master:

* a48f6511f6f463488452c320dc1370b0ac620646 Use namespace-aware meta importer for ipaplatform

PR #1151, opened 2017-10-13
[ipa-4-5] WebUI: Fix calling undefined method during reset passwords

pvomacka:
When calling reset password the whoami command is not called in batch
command, therefore the result is different then in calling
during reset password operation. That needs to be handled to properly
set entity_show method which needs to be called after to gather
data about logged in entity.

https://pagure.io/freeipa/issue/7175

**PUSH ONLY TO ipa-4-5 BRANCH - IT IS BACKPORT**

### DIFF
--- a/install/ui/src/freeipa/ipa.js
+++ b/install/ui/src/freeipa/ipa.js
@@ -191,7 +191,7 @@ var IPA = function () {
             }
         }));
 
-        batch.add_command(that.get_whoami_command(true));
+        batch.add_command(that.get_whoami_command());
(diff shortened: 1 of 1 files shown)


tomaskrizek:
ipa-4-5:

* 4be73033442276ea518d950e3df0531b418252d8 WebUI: remove unused parameter from get_whoami_command
* f3615994c304c3157ddfb2a32bf0de3cc1a12598 WebUI: Fix calling undefined method during reset passwords

PR #1155, opened 2017-10-16
Time skew fixes during initial replication

abbra:
This patchset implements time skew fixes for initial replication as required in the bug https://bugzilla.redhat.com/show_bug.cgi?id=1493150. This approach allows creating or re-initializing replicas in the situation when there are time differences between data centers.

Note that by default 389-ds does not allow for time skew in replication, this is generally a good approach and should be our default state. This patchset makes possible to unlock an impasse with cases where one would get replication state out of control due to various external reasons like a failed attempt at a parallel IPA upgrade.

### DIFF
--- a/install/share/Makefile.am
+++ b/install/share/Makefile.am
@@ -38,6 +38,7 @@ dist_app_DATA =				\
 	default-trust-view.ldif		\
 	delegation.ldif			\
 	replica-acis.ldif		\
+	replica-prevent-time-skew.ldif  \
 	ds-nfiles.ldif			\
 	dns.ldif			\
 	dnssec.ldif			\
--- a/install/share/replica-prevent-time-skew.ldif
+++ b/install/share/replica-prevent-time-skew.ldif
@@ -1,0 +1,4 @@
+dn: cn=config
+changetype: modify
+replace: nsslapd-ignore-time-skew
+nsslapd-ignore-time-skew: $SKEWVALUE
--- a/install/tools/ipa-replica-manage
+++ b/install/tools/ipa-replica-manage
@@ -1235,8 +1235,14 @@ def force_sync(realm, thishost, fromhost, dirman_passwd, nolookup=False):
         repl = replication.ReplicationManager(realm, thishost, dirman_passwd)
         repl.force_sync(repl.conn, fromhost)
     else:
+        ds = dsinstance.DsInstance(realm_name=realm)
+        ds.ldapi = os.getegid() == 0
+        ds.replica_manage_time_skew(prevent=False)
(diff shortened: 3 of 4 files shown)


rcritten:
Untested yet but I like the approach. I wonder if a block comment would be appropriate in the dsinstance methods to describe the reasoning for ignoring/preventing time skew. I'm not usually a fan of embedding ticket #'s into code but I wonder if this is complex enough to warrant it, or if a text desc is enough.

abbra:
Thanks. I added a text comment in `dsinstance.py` and also fixed pylint errors (long lines)

pvoborni:
A Pagure ticket was cloned for it: https://pagure.io/freeipa/issue/7211

rcritten:
Force sync will fail because it does an add operation so there can be two values on a single-valued attribute.

I believe you can replace the add with replace in replica-ignore-time-skew.ldif

I'm a little unsure about when the value should be set back to off. You are doing so a little later than I would have expected because replication should be completed immediately after step 28 and you aren't resetting it until step 42.

abbra:
Thanks, I refactored it a bit to reuse `replica-prevent-time-skew.ldif` for both cases. By default, it prevents time skew, while in `ipa-replica-manage` both enabling and disabling the time skew ignore is done explicitly via an optional argument.

I moved the second call in the initial case right after the replication. This should address your comments.

rcritten:
My force-sync test was failing with a no such entry because it was looking on the wrong host for the wrong agreement. This is because you use agreement.dn in the wait_for_repl_init. I got it working by adding this above that call:

agreement = repl.get_replication_agreement(thishost)

I also wonder what happens here in the case of winsync. Does it matter at all or should the ignore/prevent timesync be limited to the IPA_REPLICA type? I'm not sure and don't have an AD to test against this week. The 389-ds guys may be able to tell us if this would blow up winsync.

abbra:
@rcritten I updated `ipa-replica-manage` part to only apply to `IPA_REPLICA` type and fixed agreement DN use.

rcritten:
Not to be too pedantic but is replica-ignore-time-skew.ldif needed since the setting is dynamic in replica_prevent_time_skew()? Could it be renamed to replica_manage_time_skew() or something? I'm not firm on this but it's an ldif used only once during replica install.

Otherwise looks good and tests were successful.

abbra:
master:

* 051786ce372cc89e53fbab02086c2d1246580762 ds: ignore time skew during initial replication step
* 620f9653ba736948d9ec9443728cd6c5c50576c8 ipa-replica-manage: implicitly ignore initial time skew in force-sync

PR #1156, opened 2017-10-16
p11-kit: add serial number in DER format

stlaz:
This causes Firefox to report our CA certificate as not-trustworthy.
We were previously doing this correctly, however it slipped as an
error due to certificate refactoring.

https://pagure.io/freeipa/issue/7210

### DIFF
--- a/ipalib/x509.py
+++ b/ipalib/x509.py
@@ -123,18 +123,21 @@ class IPACertificate(object):
         # some field types encode-decoding is not strongly defined
         self._subject = self.__get_der_field('subject')
         self._issuer = self.__get_der_field('issuer')
+        self._serial_number = self.__get_der_field('serialNumber')
 
     def __getstate__(self):
         state = {
             '_cert': self.public_bytes(Encoding.DER),
             '_subject': self.subject_bytes,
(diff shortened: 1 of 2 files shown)


ipalib/x509.py
```diff
@@ -216,6 +217,10 @@ def serial_number(self):
         return self._cert.serial_number
 
     @property
+    def serial_number_bytes(self):
+        return self._serial_number
```
tiran: Why do you cache serial number instead of calling ``self.__get_der_field('serialNumber')`` here?

stlaz: Because it's how it's done for the rest of `_bytes` properties. The original decision was to cache all the `_bytes` attributes since it's unnecessary couple more steps to get them and the certificate represented by the `IPACertificate` instance should not change once the instance is created (notice there's no method to swap the original cert we got by `load_xxx_x509_certificate()`).

abbra:
LGTM. It was also tested in Fedora OpenQA infra and demonstrated that the problem got solved.

tomaskrizek:
master:

* 9b8b7afeb406f042c8c6d46f84cbb04126ac5204 p11-kit: add serial number in DER format

PR #1158, opened 2017-10-17
Block PyOpenSSL to prevent SELinux execmem in wsgi

tiran:
Some dependencies like Dogtag's pki.client library and custodia use
python-requsts to make HTTPS connection. python-requests prefers
PyOpenSSL over Python's stdlib ssl module. PyOpenSSL is build on top
of python-cryptography which trigger a execmem SELinux violation
in the context of Apache HTTPD (httpd_execmem).

When requests is imported, it always tries to import pyopenssl glue
code from urllib3's contrib directory. The import of PyOpenSSL is
enough to trigger the SELinux denial.

Block any import of PyOpenSSL's SSL module in wsgi by raising an
ImportError. The block is compatible with new python-requests with
unbundled urllib3, too.

Fixes: https://pagure.io/freeipa/issue/5442
Fixes: RHBZ#1491508

### DIFF
--- a/install/share/wsgi.py
+++ b/install/share/wsgi.py
@@ -25,6 +25,18 @@ WSGI appliction for IPA server.
 """
 import logging
 import os
+import sys
+
+# Some dependencies like Dogtag's pki.client library and custodia use
+# python-requsts to make HTTPS connection. python-requests prefers
+# PyOpenSSL over Python's stdlib ssl module. PyOpenSSL is build on top
+# of python-cryptography which trigger a execmem SELinux violation
+# in the context of Apache HTTPD (httpd_execmem).
+# When requests is imported, it always tries to import pyopenssl glue
+# code from urllib3's contrib directory. The import of PyOpenSSL is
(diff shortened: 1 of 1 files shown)


abbra:
Does this fix still allow python-requestshttps://www.freeipa.org/page/V4/Keytab_Retrieval#Overview to use system-wide CA trust store as maintained in `/etc/pki/ca-trust` in Fedora?

abbra:
Answering to myself after a test -- yes, it seems to work just fine. LGTM.

tiran:
The patch does not affect how requests loads CA certs.

tiran:
The patch only affects the WSGI web server. It does not affect installer, ipa command and other CLI programs. They'll still use PyOpenSSL in case it's is installed.

tomaskrizek:
master:

* dea059d158efe82ba71fe4e4669adb7caf45bc9f Block PyOpenSSL to prevent SELinux execmem in wsgi

PR #1159, opened 2017-10-17
Backport 7151

flo-renaud:
ipa-server-upgrade needs to configure certmonger with the right options
in order to track PKI, HTTP and LDAP certs (for instance the RA agent cert
location has changed from older releases).
The upgrade code looks for existing tracking requests with the expected
options by using criteria (location of the NSSDB, nickname, CA helper...)
If a tracking request is not found, it means that it is either using wrong
options or not configured. In this case, the upgrade stop tracking
all the certs, reconfigures the helpers, starts tracking the certs so that
the config is up-to-date.

The issue is that the criteria is using the keyword 'ca' instead of
'ca-name' and this leads to upgrade believing that the config needs to be
updated in all the cases.

https://pagure.io/freeipa/issue/7151


stlaz:
ACK based on a previous ACK.

PR #1168, opened 2017-10-18
ipatests: Fix on logs collection

celestian:
If the function `install_kra` or `install_ca` fails
on call `host.run_command(command, raiseonerr=raiseonerr)`
then the logs are not collected.

This situation is not optimal because we need to see what happend
during the debbuging the tests.

So, this patch solves this situation and it adds try--finally
construction.

https://pagure.io/freeipa/issue/7214


Tiboris:
Hi @celestian, 
Thanks for the fix!
Could you please add at least something in PR description, ideally in commit message too eg. where or how did you fix it, so it will be more obvious than it is now?

celestian:
@Tiboris , addressed.

celestian:
This patch is used here:
https://github.com/celestian/freeipa/pull/2
and it really collects the logs.

Tiboris:
master:

* 3a0410267f6f21678a827c051aecea6796c5ccd5 ipatests: Fix on logs collection

PR #1169, opened 2017-10-18
[Backport][ipa-4-5] Less confusing message for PKINIT configuration during install

slaykovsky:
The message about an error during replica setup was causing the
users to think the installation gone wrong even though this was
an expected behavior when ipa-replica-install was ran without
--no-pkinit flag and CA somehow is not reachable which defines
that there is something wrong in a topology but does not lead
to failure of the replica's installation. So now installation
will not print error messages to stdout but rather will give a
recomendation to user and write the old error message to log
as a warning so it still will be easy to find if needed.

https://pagure.io/freeipa/issue/7179

### DIFF
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -494,8 +494,17 @@ class KrbInstance(service.Service):
             self._install_pkinit_ca_bundle()
             self.pkinit_enable()
         except RuntimeError as e:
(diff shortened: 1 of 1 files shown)


tomaskrizek:
ipa-4-5:

* c79cdccfa2361f013fb51e8225b4630abab8b557 Less confusing message for PKINIT configuration during install

PR #1174, opened 2017-10-18
[Backport][ipa-4-5] ipatests: Fix on logs collection

tomaskrizek:
Original PR: #1168 

---

If the function `install_kra` or `install_ca` fails
on call `host.run_command(command, raiseonerr=raiseonerr)`
then the logs are not collected.

This situation is not optimal because we need to see what happend
during the debbuging the tests.

So, this patch solves this situation and it adds try--finally
construction.

https://pagure.io/freeipa/issue/7214


tomaskrizek:
ipa-4-5:

* 653ed4a00b90335153cea39378bf32685b1df8d2 ipatests: Fix on logs collection

PR #1175, opened 2017-10-18
ipaclient.plugins.dns: Cast DNS name to unicode

slaykovsky:
cmd.api.Command.dnsrecord_split_parts expects name to be unicode
string and instead gets ascii. It leads to an error:
ipa: ERROR: invalid 'name': must be Unicode text

This commit's change is casting name's type to unicode so
'ipa dnsrecord-mod' will not fail with error above.

https://pagure.io/freeipa/issue/7185


Tiboris:
PR #1136 has been already pushed into ipa-4-5. Waiting for CI to pass.

Tiboris:
master:

* fad88b358b6bd40b96ea7db11c2d7ed148225d06 ipaclient.plugins.dns: Cast DNS name to unicode

PR #1178, opened 2017-10-18
WebUI: make Domain Resolution Order writable

pvomacka:
Objectclass which defines the Domain Resolution Order is added to
the object only after modification. Therefore before modification of
object the attributelevelrights does not contain the 'domainresolutionorder'
attribute and the WebUI evaluates field as not writable.

'w_if_no_aci' flag was designed to make writable those fields
for which we don't have attributelevelrights.

https://pagure.io/freeipa/issue/7169

### DIFF
--- a/install/ui/src/freeipa/idviews.js
+++ b/install/ui/src/freeipa/idviews.js
@@ -101,6 +101,7 @@ return {
                         'cn',
                         {
                             name: 'ipadomainresolutionorder',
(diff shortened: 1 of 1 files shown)


pvoborni:
LGTM

abbra:
master:

* 04da85625615bf080967defa846788a6e0e643f8 WebUI: make Domain Resolution Order writable

PR #1179, opened 2017-10-19
adtrust: filter out subdomains when defining our topology to AD

abbra:
When definining a topology of a forest to be visible over a cross-forest
trust, we set *.<forest name> as all-catch top level name already.

This means that all DNS subdomains of the forest will already be matched
by this TLN. If we add more TLNs for subdomains, Active Directory will
respond with NT_STATUS_INVALID_PARAMETER.

Filter out all subdomains of the forest root domain. All other realm
domains will be added with explicit TLN records.

Fixes https://pagure.io/freeipa/issue/6666

### DIFF
--- a/ipaserver/dcerpc.py
+++ b/ipaserver/dcerpc.py
@@ -47,6 +47,7 @@ import samba
 
 import ldap as _ldap
 from ipapython import ipaldap
+from ipapython.dnsutil import DNSName
 from dns import resolver, rdatatype
 from dns.exception import DNSException
 import pysss_nss_idmap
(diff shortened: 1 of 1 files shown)


stlaz:
`ipaserver/dcerpc.py:1606: [W0612(unused-variable), TrustDomainJoins.get_realmdomains] Unused variable 'tforest')`

abbra:
I turned the assignment into a comment. ;)

stlaz:
pep8 wants you to have a space after the start of the comment :) I did not get to do a more proper review yet.
```
PEP-8 errors:
./ipaserver/dcerpc.py:1607:9: E265 block comment should start with '# '
```

abbra:
;) I fixed this. Looked at other pep8 reports and there are none new.

abbra:
master:

* 443ecbc29e3342e9021f6f4ae72ce11d5b238ba6 adtrust: filter out subdomains when defining our topology to AD

PR #1182, opened 2017-10-20
Use os.path.isfile() and isdir()

tiran:
Replace custom file_exists() and dir_exists() functions with proper
functions from Python's stdlib.

The change also gets rid of pylint's invalid bad-python3-import error,
https://github.com/PyCQA/pylint/issues/1565


stlaz:
Tests passed, does the right thing, ACK.

stlaz:
master:

* b29db07c3b3d8937f53684fdbba985fec525d69d Use os.path.isfile() and isdir()

PR #1036, opened 2017-09-04
Fixing tox and pylint errors

felipevolpone:
Fixing import errors introduced by commit cac3475a0454b730d6e5b2093c2e63d395acd387.

https://pagure.io/freeipa/issue/7132

Output of tox commands:
<details>
<summary>tox -e py27</summary>

<p>

```
fbarreto@freeipa (fix-tox-imports) tox -e py27
py27 recreate: /home/fbarreto/projects/freeipa/.tox/py27
py27 installdeps: ipaclient[csrgen], ipatests
py27 installed: asn1crypto==0.22.0,cffi==1.10.0,cryptography==2.0.3,decorator==4.1.2,dnspython==1.15.0,enum34==1.1.6,gssapi==1.2.0,idna==2.6,ipaclient==4.6.90.dev201709041448+gitac6e4cb61,ipaddress==1.0.18,ipalib==4.6.90.dev201709041448+gitac6e4cb61,ipapython==4.6.90.dev201709041448+gitac6e4cb61,ipatests==4.6.90.dev201709041448+gitac6e4cb61,Jinja2==2.9.6,MarkupSafe==1.0,netaddr==0.7.19,netifaces==0.10.6,nose==1.3.7,polib==1.0.8,py==1.4.34,pyasn1==0.3.3,pyasn1-modules==0.1.1,pycparser==2.18,pytest==3.2.1,pytest-multihost==1.1.1,python-ldap==2.4.42,qrcode==5.3,six==1.10.0
py27 runtests: PYTHONHASHSEED='3471542700'
py27 runtests: commands[0] | /home/fbarreto/projects/freeipa/.tox/py27/bin/ipa --help
Usage: ipa [global-options] COMMAND [command-options]

Manage an IPA domain

Options:
  --version          show program's version number and exit
  -h, --help         Show this help message and exit
  -e KEY=VAL         Set environment variable KEY to VAL
  -c FILE            Load configuration from FILE.
  -d, --debug        Produce full debuging output
  --delegate         Delegate the TGT to the IPA server
  -v, --verbose      Produce more verbose output. A second -v displays the
                     XML-RPC request
  -a, --prompt-all   Prompt for ALL values (even if optional)
  -n, --no-prompt    Prompt for NO values (even if required)
  -f, --no-fallback  Only use the server configured in /etc/ipa/default.conf

See "ipa help topics" for available help topics.
See "ipa help <TOPIC>" for more information on a specific topic.
See "ipa help commands" for the full list of commands.
[...]

PR #1184, opened 2017-10-20
Drop import of ipaplatform to fix PyPI packages

tiran:
ipalib.util imports ipaplatform but ipaplatform package is not available
as wheel package on PyPI. Fixes import errors introduced by commits
cac3475a and 0b7d9c5.

https://pagure.io/freeipa/issue/7132

Original patch by Felipe Barreto Volpone <fbarreto@redhat.com>


tomaskrizek:
Implemented in PR #1036.

PR #1186, opened 2017-10-20
lint: disable no-name-in-module for py3 package

stlaz:
pylint mistakenly reports no-name-in-module when we're deciding
whether to use the package for the given python version.

### DIFF
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -82,12 +82,12 @@ except ImportError:
     from xmlrpc.client import (Binary, Fault, DateTime, dumps, loads, ServerProxy,
             Transport, ProtocolError, MININT, MAXINT)
 
-# pylint: disable=import-error
+# pylint: disable=import-error, no-name-in-module
 if six.PY3:
     from http.client import RemoteDisconnected
 else:
     from httplib import BadStatusLine as RemoteDisconnected
-# pylint: enable=import-error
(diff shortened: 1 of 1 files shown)


tomaskrizek:
I don't have this issue when using `python2-pylint-1.7.4-1`

stlaz:
I'll see what happens after the 1.7.4-1 version is downloadable from the mirror my computer synces with.

stlaz:
Still seeing this with 1.7.4-1:
```
Pylint on /usr/bin/python is running, please wait ...
************* Module ipalib.rpc
ipalib/rpc.py:87: [E0611(no-name-in-module), ] No name 'RemoteDisconnected' in module 'http.client')
```
```bash
$ dnf list installed | grep pylint
pylint.noarch                          1.7.4-1.fc26                @updates     
python2-pylint.noarch                  1.7.4-1.fc26                @updates     
python3-pylint.noarch                  1.7.4-1.fc26                @updates
```

tomaskrizek:
Can you reproduce this with a clean `fedora:26` container? I wasn't able to. I think this might be an issue in your environment.

I'm also removing your `trivial` clickbait :D

stlaz:
Hm, it seems to only be an issue for me, I'll close this PR and we'll reopen it should I find out why this happens/someone else encounters it.

PR #1187, opened 2017-10-23
Add a notice to restart ipa services after certs are installed

slaykovsky:
Adding notice for user to restart services after
ipa-server-certinstall.

https://pagure.io/freeipa/issue/7016

### DIFF
--- a/install/tools/man/ipa-server-certinstall.1
+++ b/install/tools/man/ipa-server-certinstall.1
@@ -28,7 +28,8 @@ PKCS#12 is a file format used to safely transport SSL certificates and public/pr
 
 They may be generated and managed using the NSS pk12util command or the OpenSSL pkcs12 command.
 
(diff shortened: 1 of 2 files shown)


abbra:
Looks good but I see a problem in that the other two methods (http and dirsrv) also don't restart corresponding services. We should do something about it.

rcritten:
You should add back the and/or.

flo-renaud:
Hi @slaykovsky,

pylint check is failing because of the print() function. You can add
`from __future__ import print_function`
to resolve the issue.

flo-renaud:
Hi @slaykovsky 
thank you for the patch, works as expected and successfully passes the lint checks.

Tiboris:
master:

* 59802d37d17a33fc2417aa68dd10b3373e1800bd Add a notice to restart ipa services after certs are installed

PR #1190, opened 2017-10-23
test_external_dns: add missing test cases

Rezney:
Add NTP, ipa-ca and ADTrust system records tests. Also test if
changes are being reflected when uninstalling a host.

The test cases are added as extension into test_dns_locations suite.

https://pagure.io/freeipa/issue/6091

### DIFF
--- a/ipatests/test_integration/test_dns_locations.py
+++ b/ipatests/test_integration/test_dns_locations.py
@@ -3,6 +3,8 @@
 #
 import logging
 import time
+import pytest
+import six
 import dns.resolver
 import dns.rrset
 import dns.rdatatype
@@ -11,6 +13,7 @@ import dns.rdataclass
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_plugins.integration import tasks
 from ipapython.dnsutil import DNSName
+from ipalib.constants import IPA_CA_RECORD
 
 logger = logging.getLogger(__name__)
 
@@ -25,15 +28,45 @@ IPA_DEFAULT_MASTER_SRV_REC = (
     (DNSName(u'_kpasswd._udp'), 464),
 )
 
+IPA_DEFAULT_ADTRUST_SRV_REC = (
+    # srv record name, port
+    (DNSName(u'_ldap._tcp.Default-First-Site-Name._sites.dc._msdcs'), 389),
+    (DNSName(u'_ldap._tcp.dc._msdcs'), 389),
+    (DNSName(u'_kerberos._tcp.Default-First-Site-Name._sites.dc._msdcs'), 88),
+    (DNSName(u'_kerberos._udp.Default-First-Site-Name._sites.dc._msdcs'), 88),
+    (DNSName(u'_kerberos._tcp.dc._msdcs'), 88),
+    (DNSName(u'_kerberos._udp.dc._msdcs'), 88),
+)
+
+# we keep NTP SRV record separated as NTP is "optional"
+IPA_DEFAULT_NTP_SRV_REC = (
+    # srv record name, port
+    (DNSName("_ntp._udp"), 123),
+)
+
+IPA_CA_A_REC = (
+    (DNSName(six.text_type(IPA_CA_RECORD))),
+)
+
 
 def resolve_records_from_server(rname, rtype, nameserver):
+    error = None
     res = dns.resolver.Resolver()
     res.nameservers = [nameserver]
-    res.lifetime = 10
-    logger.debug("Query: %s %s, nameserver %s", rname, rtype, nameserver)
-    ans = res.query(rname, rtype)
-    logger.debug("Answer: %s", ans.rrset)
-    return ans.rrset
+    res.lifetime = 30
+    logger.info("Query: %s %s, nameserver %s", rname, rtype, nameserver)
+    # lets try to query 3x
+    for _i in range(3):
+        try:
+            ans = res.query(rname, rtype)
+            logger.info("Answer: %s", ans.rrset)
+            return ans.rrset
+        except (dns.resolver.NXDOMAIN, dns.resolver.Timeout) as e:
+            error = e
+            time.sleep(10)
+
+    pytest.fail("Query: {} {}, nameserver {} failed due to {}".format(
+        rname, rtype, nameserver, error))
 
 
 def _gen_expected_srv_rrset(rname, port, servers, ttl=86400):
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_dns_locations.py
```diff
@@ -25,15 +27,47 @@
...
-    return ans.rrset
+    res.lifetime = 30
+    logger.info("Query: %s %s, nameserver %s", rname, rtype, nameserver)
+    # lets try to query 3x
+    for i in range(3):
+        try:
+            ans = res.query(rname, rtype)
+            break
+        except (dns.resolver.NXDOMAIN, dns.resolver.Timeout) as e:
+            ans = e
```
MartinBasti: I don't like storing error state into variable answer.  Could you please define `exc = None` or `error = None` before cycle and store exception there

ipatests/test_integration/test_dns_locations.py
```diff
@@ -25,15 +27,47 @@
...
+            time.sleep(10)
+            pass
+    try:
+        logger.info("Answer: %s", ans.rrset)
+        return ans.rrset
+    except AttributeError:
+        pytest.fail("Query: %s %s, nameserver %s failed due {}",
```
MartinBasti: if you use `return ans....` in the previous try block, you don't need this try/except block. Anything here is error so just fail:
```
pytest.fail("Query: %s %s, nameserver %s failed due {}", rname, rtype, nameserver, error)
```

(Also instead of catching AttributeError would be better to use `isinstance(ans, Exception)` to detect exception)

ipatests/test_integration/test_dns_locations.py
```diff
@@ -83,53 +125,83 @@ def install(cls, mh):
...
-                    server_ip, name_abs, expected, query))
+    @classmethod
+    def delete_update_records(cls, rec_list=IPA_DEFAULT_MASTER_SRV_REC):
+        filepath = '/tmp/ipa.nsupdate'
+
+        cls.master.run_command([
+            'ipa', 'dns-update-system-records', '--dry-run', '--out', filepath
+        ])
+
+        if rec_list == IPA_CA_A_REC:
```
MartinBasti: Please do not use such comparison. Remove that ifs completly please.

This method needs only `rname`, so ask only for rnames as parameter, not for the whole structure.

```
delete_update_records(rnames=(r[0] for r in IPA_DEFAULT_MASTER_SRV_REC))
delete_update_records(rnames=IPA_CA_A_REC)
```

This method should be probably called `delete_update_system_records` as it is updating only them.

ipatests/test_integration/test_dns_locations.py
```diff
@@ -83,53 +125,83 @@ def install(cls, mh):
...
+        # allow unauthenticates nsupdate (no need to testing authentication)
+        cls.master.run_command([
+            'ipa', 'dnszone-mod', str(cls.domain),
+            '--update-policy=grant * wildcard *;'
+        ], raiseonerr=False)
+
+        cls.master.run_command(['nsupdate', '-g', filepath])
+        time.sleep(15)
+
+    def _test_against_server(self, server_ip, domain, expected_servers,
```
MartinBasti: Please split this into 2 methods:
* _test_A_rec_....
* _test_SRV_rec_...

rather than having one big method splitted by ifs to two separated parts

ipatests/test_integration/test_dns_locations.py
```diff
@@ -287,3 +359,99 @@ def test_change_weight(self):
 
         for ip in (self.replicas[1].ip, self.master.ip):
             self._test_against_server(ip, domain_paris_loc, servers_paris_loc)
+
+    def test_restore_locations_and_weight(self):
+        """Restore locations and weight. Not just for test purposes but also
+        for the following tests"""
+
+        self.master.run_command([
+            'ipa', 'server-mod', self.replicas[0].hostname, '--location='''])
```
MartinBasti: Did you mean: `"--location=''"` ?

ipatests/test_integration/test_dns_locations.py
```diff
@@ -287,3 +359,99 @@ def test_change_weight(self):
...
+        self.master.run_command([
+            'ipa', 'server-mod', self.replicas[1].hostname, '--location='''])
+        self.master.run_command([
+            'ipa', 'server-mod', self.master.hostname, '--location='''])
+
```
MartinBasti:
```python
for hostname in (self.master.hostname, self.replica[0].hostname, self.replica[1].hostname):
    self.master.run_command(['ipa', 'server-mod', hostname, "--location=''"])
```

ipatests/test_integration/test_dns_locations.py
```diff
@@ -287,3 +359,99 @@ def test_change_weight(self):
...
+        tasks.restart_named(self.master, self.replicas[0], self.replicas[1])
+
+    def test_ipa_ca_records(self):
+        """ Test ipa-ca dns records with firstly removing the records and then
+        using the nsupdate generated by dns-update-system-records"""
+        self.delete_update_records(rec_list=IPA_CA_A_REC)
+
+        expected_servers = (self.master.ip, self.replicas[1].ip)
+
+        for ip in (self.master.ip,):
```
MartinBasti: Quite useless for cycle.

ipatests/test_integration/test_dns_locations.py
Rezney: Originally yes but I tried two variations like "--location=''" and '--location=""' but always got:

`(Pdb) self.stderr_text
"ipa: ERROR: '': location not found\n"`

or:

`(Pdb) self.stderr_text
'ipa: ERROR: "": location not found\n'
`

''' is actually working for me :)

ipatests/test_integration/test_dns_locations.py
Rezney: sure thing...

ipatests/test_integration/test_dns_locations.py
Rezney: makes sense...

Rezney:
Thank you very much for the review Martin. I hopefully covered all your insights.

ipatests/test_integration/test_dns_locations.py
```diff
@@ -25,15 +27,47 @@
...
 
 def resolve_records_from_server(rname, rtype, nameserver):
     res = dns.resolver.Resolver()
     res.nameservers = [nameserver]
-    res.lifetime = 10
-    logger.debug("Query: %s %s, nameserver %s", rname, rtype, nameserver)
-    ans = res.query(rname, rtype)
-    logger.debug("Answer: %s", ans.rrset)
-    return ans.rrset
+    res.lifetime = 30
```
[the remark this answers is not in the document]
Rezney: mostly leftover with playing with sleeps and etc. as I was experiencing some records missing exceptionally. Would not it help here?

MartinBasti:
LGTM

stlaz:
@MartinBasti approves and I did see the test pass, ACK.

tomaskrizek:
master:

* 02c1d0e8e52fdd8e5d4acd418a682fc4b191dceb test_external_dns: add missing test cases

tomaskrizek:
@Rezney please backport to `ipa-4-5` manually.

PR #1193, opened 2017-10-24
Prevent installation with single label domains

slaykovsky:
Adds validation to prevent user to install ipa with single label
domain.

https://pagure.io/freeipa/issue/7207

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -471,6 +471,11 @@ def install_check(installer):
     if not options.realm_name:
         realm_name = read_realm_name(domain_name, not installer.interactive)
         logger.debug("read realm_name: %s\n", realm_name)
+
+        try:
+            validate_domain_name(realm_name, entity="realm")
+        except ValueError as e:
+            raise ScriptError("Invalid realm name: {}".format(unicode(e)))
     else:
         realm_name = options.realm_name.upper()
 
--- a/ipalib/install/service.py
+++ b/ipalib/install/service.py
@@ -129,6 +129,10 @@ class ServiceInstallInterface(common.Installable,
         cli_names='--realm',
     )
 
+    @realm_name.validator
+    def realm_name(self, value):
+        validate_domain_name(value, entity="realm")
+
     host_name = knob(
         str, None,
         description="The hostname of this machine (FQDN). If specified, the "
--- a/ipalib/util.py
+++ b/ipalib/util.py
@@ -388,11 +388,18 @@ def validate_dns_label(dns_label, allow_underscore=False, allow_slash=False):
                            % dict(chars=chars, chars2=chars2))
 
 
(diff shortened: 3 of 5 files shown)


flo-renaud:
Hi @slaykovsky 
thank you for your patch, it works as expected:

- installation with --domain specifying a single label domain is forbidden and outputs the error:
`ipa-server-install: error: option --domain: single label domains are not supported`
- ipa realmdomains-mod --domain $DOMAIN --add-domain=$SINGLELABELDOMAIN fails with:
`ipa: ERROR: invalid 'add_domain': single label domains are not supported`
- ipa config-mod --domain-resolution-order=$DOMAIN:$SINGLELABELDOMAIN fails with:
`ipa: ERROR: invalid 'ipadomainresolutionorder': Invalid domain name '$SINGLELABELDOMAIN': single label domains are not supported`

It may be interesting to also add the same check for the realm value provided to `ipa-server-install --realm $REALM`, because we currently accept single label realms such as LOCALDOMAIN.

You can also add a test:
- in ipatests/test_xmlrpc/test_realmdomains_plugin.py checking that `realmdomains-mod --add-domain` fails for a SLD.
- in ipatests/test_xmlrpc/test_config_plugin.py checking that `ipadomainresolutionorder` does not accept SLD.

slaykovsky:
Hi @flo-renaud 
thanks for the review!

I've added tests and realm verification (which is basically domain verification).

flo-renaud:
Hi @slaykovsky 
you may want to rebase to the current master in order to pick the switch from `pep8` to `pycodestyle`, it will fix the travis ci issue.

ipaserver/install/server/install.py
```diff
@@ -474,6 +474,11 @@ def install_check(installer):
     else:
         realm_name = options.realm_name.upper()
 
+    try:
+        validate_domain_name(realm_name)
+    except ValueError as e:
+        raise ScriptError("Invalid realm name: {}".format(unicode(e)))
+
```
flo-renaud: This check is performed quite late in the installation. You may consider adding
```
    @realm_name.validator
    def realm_name(self, value):
        validate_domain_name(value)
```
into ipalib/install/service.py so that the validation is performed earlier (as for domain_name). Keep the validation here for prompted input, though.

It would also be more clear if the error message for realm mentioned 'single label realms are not supported' instead of 'single label domains'.

slaykovsky: Thank you so much for pointing this out. Didn't know about validators in `ipalib/install/service.py` :)

flo-renaud:
Hi @slaykovsky 
thank you for the updated PR, works as expected. I am just waiting for travis-ci before giving the ACK label.

Rezney:
master:

* 905ab93c958a539eb4af7d4b008a5aa02292ba12 Prevent installation with single label domains

PR #1195, opened 2017-10-24
Fix ipa-replica-conncheck when called with --principal

flo-renaud:
ipa-replica-conncheck can be called with --principal / --password or
with an existing Kerberos credential cache in order to supply the
authorized identity logging in to the master machine (in
auto-master-check mode).

In domain-level 0, the tool is called with --principal and password
and tries to obtain a TGT by performing kinit, but does not set the
env var KRB5CCNAME. Subsequent calls to IPA API do not use the
credential cache and fail. In this case, ipa-replica-conncheck falls
back to using SSH to check master connectivity instead of IPA API,
and the ssh check is less robust.

The code should set the KRB5CCNAME env var for IPA API to use the
credential cache.

Fixes:
https://pagure.io/freeipa/issue/7221

### DIFF
--- a/install/tools/ipa-replica-conncheck
+++ b/install/tools/ipa-replica-conncheck
@@ -530,6 +530,9 @@ def main():
                 if result.returncode != 0:
                     raise RuntimeError("Could not get ticket for master server: %s" %
                                         result.error_output)
+                # Now that the cred cache file is initialized,
+                # use it for the IPA API calls
+                os.environ['KRB5CCNAME'] = CCACHE_FILE
 
             try:
                 logger.info("Check RPC connection to remote master")


stlaz:
Please rebase on current master, Travis is switching `pep8` to `pycodestyle`

rcritten:
This looks like a reasonable thing to do but I'm not sure this addresses the reported issue.

Running ipa-replica-conncheck by itself on an un-configured machine, which is similar to domain level 0, still falls back to SSH because there is no api.env.domain set. Even if there were there is no krb5.conf to kinit against, nor /etc/ipa/ca.crt to use to connect to the remote server with. So I'm a bit confused. Am I testing this wrong?

Note that if that stuff is in place it seems to work.

flo-renaud:
The domain-level 0 scenario is not completely equivalent to running ipa-replica-conncheck on an un-configured machine.
In the domain-level 0 scenario, ipa-replica-conncheck [(line 882)](https://github.com/freeipa/freeipa/blob/master/ipaserver/install/server/replicainstall.py#L882) is called after create_ipa_conf() [(line 715)](https://github.com/freeipa/freeipa/blob/master/ipaserver/install/server/replicainstall.py#L715), meaning that /etc/ipa/default.conf is already configured. It does not require kerberos to be already set-up as ipa-replica-conncheck creates a temporary krb5.conf file just for acquiring the ticket [(line 498)](https://github.com/freeipa/freeipa/blob/master/install/tools/ipa-replica-conncheck#L498).

Do you think we should also address running ipa-replica-conncheck in a completely unconfigured machine? I did not consider this as the man page states 

> REPLICA MACHINE OPTIONS
>        This  set of options is used when the connection check is run on a **prepared** IPA replica machine

rcritten:
I'm ok with you telling me my test was bogus :-)

I think ipa-replica-conncheck will work on an unconfigured machine using SSH. It would be complex to do so otherwise.

So this change works for me.

tomaskrizek:
master:

* 7b149b3c7f7513fa21051ed26bc68704cb78e971 Fix ipa-replica-conncheck when called with --principal

tomaskrizek:
@flo-renaud Please backport to `ipa-4-5` manually.

flo-renaud:
Backport to 4-5: https://github.com/freeipa/freeipa/pull/1214

PR #1215, opened 2017-10-27
Add indexing to improve host-find performance

stlaz:
host-find <host_name> command performance gets deteriorated when
there's way too many hosts in the LDAP tree. We're adding indices
to try and mitigate this behavior.

https://pagure.io/freeipa/issue/6371

============================
Untested, our lab is down and my private virtual machine segfaults during the provisioning process due to lack of memory.

### DIFF
--- a/install/updates/20-indices.update
+++ b/install/updates/20-indices.update
@@ -70,8 +70,9 @@ default:cn: fqdn
 default:ObjectClass: top
 default:ObjectClass: nsIndex
 default:nsSystemIndex: false
-default:nsIndexType: eq
-default:nsIndexType: pres
+only:nsIndexType: eq
+only:nsIndexType: pres
+only:nsIndexType: sub
 
 dn: cn=macAddress,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
 default:cn: macAddress
--- a/install/updates/20-indices.update
+++ b/install/updates/20-indices.update
@@ -268,3 +268,43 @@ default: objectClass: nsIndex
 only: nsSystemIndex: false
 only: nsIndexType: eq
 only: nsIndexType: sub
+
+dn: cn=description,cn=index,cn=userroot,cn=ldbm database,cn=plugins,cn=config
+default: cn: description
+default: objectclass: top
+default: objectclass: nsindex
+default: nssystemindex: false
+default: nsindextype: eq
+default: nsindextype: sub
+
+dn: cn=l,cn=index,cn=userroot,cn=ldbm database,cn=plugins,cn=config
+default: cn: l
+default: objectclass: top
+default: objectclass: nsindex
+default: nssystemindex: false
+default: nsindextype: eq
+default: nsindextype: sub
+
+dn: cn=nsOsVersion,cn=index,cn=userroot,cn=ldbm database,cn=plugins,cn=config
+default: cn: nsOsVersion
+default: objectclass: top
+default: objectclass: nsindex
+default: nssystemindex: false
+default: nsindextype: eq
+default: nsindextype: sub
+
+dn: cn=nsHardwarePlatform,cn=index,cn=userroot,cn=ldbm database,cn=plugins,cn=config
+default: cn: nsHardwarePlatform
+default: objectclass: top
+default: objectclass: nsindex
+default: nssystemindex: false
+default: nsindextype: eq
+default: nsindextype: sub
+
+dn: cn=nsHostLocation,cn=index,cn=userroot,cn=ldbm database,cn=plugins,cn=config
+default: cn: nsHostLocation
+default: objectclass: top
+default: objectclass: nsindex
+default: nssystemindex: false
+default: nsindextype: eq
+default: nsindextype: sub
--- a/install/share/indices.ldif
+++ b/install/share/indices.ldif
@@ -108,6 +108,7 @@ cn: fqdn
 nsSystemIndex: false
 nsIndexType: eq
 nsIndexType: pres
+nsIndexType: sub
 
 dn: cn=macAddress,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
 changetype: add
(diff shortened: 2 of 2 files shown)


install/updates/20-indices.update
```diff
@@ -267,3 +267,43 @@ default: objectClass: nsIndex
 only: nsSystemIndex: false
 only: nsIndexType: eq
 only: nsIndexType: sub
+
+dn: cn=description,cn=index,cn=userroot,cn=ldbm database,cn=plugins,cn=config
+default: cn: description
+default: objectclass: top
+default: objectclass: nsindex
+only: nssystemindex: false
```
rcritten: I see that only is used in a lot of other places in this update file but I think default is more appropriate.

stlaz:
The same happens to me even though all the objects are in the LDIF and even in  the `cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config` tree. Maybe @tbordaz could help explaining why?
I also did not notice any significant performance boost after updating the DS with the supplied LDIF file, although I added the indexing after filling my DS with the data (I was finally able to get a machine in the lab that could handle this).

rcritten:
This is what logconv.pl -u shows (I added some line breaks to limit scrolling):
```
Unindexed Component #2 (notes=U)
  -  Date/Time:             02/Nov/2017:10:35:13
  -  Connection Number:     13
  -  Operation Number:      3
  -  Etime:                 0.0100798645
  -  Nentries:              0
  -  IP Address:            192.168.166.197
  -  Search Base:           cn=computers,cn=accounts,dc=example,dc=com
  -  Search Scope:          1 (one)
  -  Search Filter:         (&(|(fqdn=*foo*)(description=*foo*)(l=*foo*)(nshostlocation=*foo*)
(krbcanonicalname=*foo*)(krbprincipalname=*foo*)(nshardwareplatform=*foo*)(nsosversion=*foo*)
(managedby=*foo*))(&(objectclass=ipaobject)(objectclass=nshost)(objectclass=ipahost)
(objectclass=pkiuser)(objectclass=ipaservice)))
  -  Bind DN:               uid=admin,cn=users,cn=accounts,dc=example,dc=com
```

rcritten:
I suspect this is because fqdn doesn't have sub as an index type.

rcritten:
I added sub to fqdn and that fixed it. Search time for 1k hosts went from 1.2s to 0.3s.

I ran this past Ludwig and he is going to look at it and get back to me. There may be another, better way to handle this.

stlaz:
Thanks for the research, @rcritten, I would not know this. I'm only afraid we can't go changing the schema as we like, so I hope Ludwig will be able to figure something out.

rcritten:
It's not a schema change. We control the indices. The thing is there are trade-offs of performance and disk, CPU and RAM when it comes to an index. Ludwig sent me something out-of-band, I'll cc you on any future responses I make.

He pointed out initially that there is a 4000 id scan limit and after that it turns into an allid search. The test data you're likely using as I recall uses 4k hosts so you may never see an fully-indexed search unless you update nsslapd-idlistscanlimit as well as adding sub to fqdn.

stlaz:
Ah, for some reason I thought `sub` needs to be added to `fqdn` attribute type, dunno why I was thinking that, really.
I am testing this with 40k hosts, I'll just add the `sub` operation to `fqdn` index and see what it does.

install/updates/20-indices.update
stlaz: `default` won't update the entry as we want it should the entry with such an attribute exist. A good example is the `fqdn` indices entry - should we add `default:nsIndexType: sub`, the `sub` operation type does not get added, we have to use `only` for all the index types.

stlaz:
After the latest changes, `host-find` takes about 2-3 seconds rather than 27-30 seconds on 40k host entries.

install/updates/20-indices.update
rcritten: For _new_ indexes then "default" should be fine. For updating _existing_ indexes then "only" is the way we've managed it in the past ("add" should work as well). "only" is a bit of a hammer since it would eliminate any end-user changes to the index.

stlaz: This is quite ambivalent, there would be reason to keep `only` so that any existing configuration we did not create can get overwritten in the same manner as our managed configuration. By putting `default` there, we still consider any possible existing configuration inadequate but are deciding not do do anything with it, which does not destroy the user config yet won't work in the way we want 🤕 

Please, help me decide here. Do we really want to go with `default`?

rcritten: IPA tries to make good default decisions for users. If this is anything other than default then on every upgrade the index is doing to revert back. This could be a bad experience. We've done this in places where we had the original index types incorrect to force the matter but as general principal I'd prefer to leave it up to the user to decide if they want additional or no types.

This will affect their installation but it is theirs after all. I'd go with default.

stlaz: That's a good reasoning, updating. hopefully I did not forget anything.

install/updates/20-indices.update
```diff
@@ -267,3 +268,43 @@ default: objectClass: nsIndex
...
+default: nsindextype: sub
+
+dn: cn=nsHardwarePlatform,cn=index,cn=userroot,cn=ldbm database,cn=plugins,cn=config
+default: cn: nsHardwarePlatform
+default: objectclass: top
+default: objectclass: nsindex
+defaul: nssystemindex: false
+defaul: nsindextype: eq
+defaul: nsindextype: sub
+
```
rcritten: This is what is causing the tests to fail: defaul -> default.

rcritten:
There is a typo in the update file. defaul -> default.

install/updates/20-indices.update
stlaz: Oh god, oh crap, thanks 😞

stlaz:
master:

* c99b37737a43a5b101158cff82136a56f9f47af3 Add the sub operation for fqdn index config
* 5458bb506f70653846e2053fa0234cd63cc3f629 Add indexing to improve host-find performance

PR #1228, opened 2017-11-02
Fix ipa-restore (python2)

flo-renaud:

Fixes:


rcritten:
Looks good.

flo-renaud:
master:

* f4a208311aa6618336e1ba6ae277243c727ec1fc Fix ipa-restore (python2)

PR #1230, opened 2017-11-03
test_forced_client: decode get_file_contents() result

Rezney:
Decode get_file_contents() in order to not get bytes when running py3


stlaz:
LGTM.

tomaskrizek:
master:

* 59e136e8bad0612d132ddf8c38711a25fdf9b0c1 test_forced_client: decode get_file_contents() result

PR #1232, opened 2017-11-06
Making ipa-ca-install more resilient

frasertweedale:
, or: *Proactively run ipa-certupdate for great good!*

These commits fix a couple of issues that can occur after a deployment has been
promoted from CA-less to CA-ful, and the admin does not follow up with
`ipa-certupdate`.  (And why should they have to?)

```
a9ad3b5ab (Fraser Tweedale, 6 days ago)
   Run certupdate after promoting to CA-ful deployment

   After installing a CA in a CA-less installations (using ipa-ca-install),
   the new CA certificate is not installed in
   /etc/httpd/alias. This causes communication failure between IPA framework
   and Dogtag (it cannot verify the Dogtag server certificate).

   Perform a CertUpdate as the final step when promoting a CA-less deployment
   to CA-ful.

   Fixes: https://pagure.io/freeipa/issue/7230

21fbf7088 (Fraser Tweedale, 7 days ago)
   ipa-ca-install: run certupdate as initial step

   When installing a CA replica, perform a certupdate to ensure that the
   relevant CA cert is present.  This is necessary if the admin has just
   promoted the topology from CA-less to CA-ful but didn't manually run
   ipa-certupdate afterwards.

   Fixes: https://pagure.io/freeipa/issue/6577

9520781fb (Fraser Tweedale, 7 days ago)
   CertUpdate: make it easy to invoke from other programs

   The guts of ipa-certupdate are useful to execute as part of other programs
   (e.g. as a first step of ipa-ca-install).  Refactor
   ipa_certupdate.CertUpdate to make it easy to do that.  In particular, make
   it possible to use an already-initialised API object.

   Part of: https://pagure.io/freeipa/issue/6577
```

### DIFF
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -27,7 +27,7 @@ import tempfile
 from ipalib.install.kinit import kinit_keytab
 from ipapython import ipautil
 
-from ipaclient.install.ipa_certupdate import CertUpdate
+from ipaclient.install import ipa_certupdate
 from ipaserver.install import installutils
 from ipaserver.install.installutils import create_replica_config
 from ipaserver.install.installutils import check_creds, ReplicaConfig
@@ -178,7 +178,7 @@ def install_replica(safe_options, options, filename):
     # Run ipa-certupdate to ensure we have the CA cert.  This is
     # necessary if the admin has just promoted the topology from
     # CA-less to CA-ful, and ipa-certupdate has not been run yet.
-    CertUpdate.run_with_args(api)
+    ipa_certupdate.run_with_args(api)
 
     # CertUpdate restarts DS causing broken pipe on the original
     # connection, so reconnect the backend.
@@ -257,7 +257,7 @@ def install_master(safe_options, options):
     # Run ipa-certupdate to add the new CA certificate to
     # certificate databases on this server.
     logger.info("Updating certificate databases.")
-    CertUpdate.run_with_args(api)
+    ipa_certupdate.run_with_args(api)
 
 def install(safe_options, options, filename):
     options.promote = False
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -27,6 +27,7 @@ import tempfile
 from ipalib.install.kinit import kinit_keytab
 from ipapython import ipautil
 
+from ipaclient.install.ipa_certupdate import CertUpdate
 from ipaserver.install import installutils
 from ipaserver.install.installutils import create_replica_config
 from ipaserver.install.installutils import check_creds, ReplicaConfig
@@ -173,6 +174,16 @@ def install_replica(safe_options, options, filename):
     if (not options.promote and not options.admin_password and
             not options.skip_conncheck and options.unattended):
         sys.exit('admin password required')
+
+    # Run ipa-certupdate to ensure we have the CA cert.  This is
+    # necessary if the admin has just promoted the topology from
+    # CA-less to CA-ful, and ipa-certupdate has not been run yet.
+    CertUpdate.run_with_args(api)
+
+    # CertUpdate restarts DS causing broken pipe on the original
+    # connection, so reconnect the backend.
+    api.Backend.ldap2.disconnect()
+    api.Backend.ldap2.connect()
 
     if options.promote:
         config = ReplicaConfig()
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -254,6 +254,10 @@ def install_master(safe_options, options):
     ca.install_check(True, None, options)
     ca.install(True, None, options)
 
+    # Run ipa-certupdate to add the new CA certificate to
+    # certificate databases on this server.
+    logger.info("Updating certificate databases.")
+    CertUpdate.run_with_args(api)
 
 def install(safe_options, options, filename):
     options.promote = False
--- a/ipaclient/install/ipa_certupdate.py
+++ b/ipaclient/install/ipa_certupdate.py
@@ -57,149 +57,147 @@ class CertUpdate(admintool.AdminTool):
         api.finalize()
 
         api.Backend.rpcclient.connect()
-        self.run_with_args(api)
+        run_with_args(api)
         api.Backend.rpcclient.disconnect()
 
-    @classmethod
-    def run_with_args(cls, api):
-        """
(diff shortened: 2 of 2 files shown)


rcritten:
I'm not sure you should be dinged on this but the previous CA-less certificates show up with very little detail in cert-find because it defers looking up details. I have an ugly hack to work around this for a pure CA-less but for this upgrade state I'm not sure what it should do.

  Serial number: 4096
  Serial number (hex): 0x1000

I'm not sure yet if this is related but I suspect it is.
1. single IPA CA-less master
2. ipa-ca-install
3. works
4. enroll ipa1.example.com as client, run ipa-replica-install

Fails with:

ACIError: Insufficient access: Principal 'ldap/replica1.example.com@EXAMPLE.COM' is not permitted to use CA 'ipa' with profile 'caIPAserviceCert' for certificate issuance.

tiran:
Please don't add classmethods unless it's an alternative constructor. staticmethods are just a silly idea. I know, I know, some linters and IDE suggest classmethods or staticmethods. Don't listen to them.

frasertweedale:
@tiran, what then shall I do... extract all of `run_with_api` to a top-level function (which is a noisy change but I am happy enough to do), or change to an instance method (which it does not need to be, thus for readability I would prefer not to do this).

IMO a class method is fine for behaviour that is "obviously part of this class", which doesn't need access to instance variables.

Anyhow, if you could explain why you do not think a classmethod is not appropriate and suggest the alternative approach you prefer, that would be helpful :)

tiran:
I see, the admin tools are not really designed to integrate with other code. I think the class can altered to work as required with some small modifications:

Change AdminTool so you can create an instance without options and args:

```
class AdminTool(object):
    def __init__(self, options, args):
        self.options = options
        self.args = args
        if options is not None:
            self.safe_options = self.option_parser.get_safe_opts(options)
```

Change CertUpdate to not initialize the API when it is already finalized:

```
class CertUpdate(admintool.AdminTool):
    ...
    def run(self):
        check_client_configuration()

        if not api.isdone('finalize'):
            api.bootstrap(context='cli_installer', confdir=paths.ETC_IPA)
            api.finalize()
```        

Then you should be able to run ``CertUpdate(None, None).run()`` easily.

frasertweedale:
@tiran I dunno mate, this seems more complicated than extracting a (class)method so that we can just give it an API value and have it execute the behaviour.

You have said "don't add classmethods" but you have not given reasons.  This seems like a valid use to me.  Please show me why I'm wrong.

tiran:
I'm not particularly fond of classmethods unless they act as an alternative constructor or modify a class object. It's not the case in this patch. Once you have turned all methods to class or static methods, it's going to be really tricky to refactor the class or to pass around instance arguments. What happens if we ever need to apply arguments?

I said that staticmethods are a silly idea and should not be used at all. Here is what [Guido](https://mail.python.org/pipermail/python-ideas/2012-May/014969.html) says about static methods:

> Ad we all know how limited static methods are. (They're basically an accident -- back in the Python 2.2 days when I was inventing new-style classes and descriptors, I meant to implement class methods but at first I didn't understand them and accidentally implemented static methods first. Then it was too late to remove them and only provide class methods.)

How about I make an alternative patch and you see how you like it?

frasertweedale:
@tiran if we need to add arguments to the command, we process the args in `run` and pass those args to `run_with_api`.  (This highlights that `run_with_api` is a bad name, so I will change it to something else like `run_with_args(cls, api)`).  So I'll push an updated patch tomorrow.

Regarding `classmethod`...

> I'm not particularly fond of classmethods unless they act as an alternative constructor or modify a class object. 

You're welcome to your feelings, but they do not amount to a valid argument.  I want to know
*reasons* why the use of `classmethod` in the current patch is problematic.

I grok the point on `staticmethod` but even so, I think it's a satisfactory way of associated related behaviour to a class.  But I am not using `staticmethod` in my patch so let us suspend debate about it.

pvoborni:
This PR seems stuck. What is the next step? It is not clear whether a new version authored by @frasertweedale or alternative by @tiran

frasertweedale:
The alternative has readability issues:

- The class constructor now gets calls multiple ways.  There are new branches of execution depending on whether some of the init args are `None` or not.  It is more confusing, and more complicated.

- Methods that do not access (and do not need to access) instance or class attributes are not marked as such.  Using `classmethod` and `staticmethod` makes it clear that these routines do not (and cannot) access instance and class attributes, respectively, which improves readability.

Reasons to avoid `classmethod` and `staticmethod` have not been put.

But I made a new commit that removes them anyway (I do not mind doing this).  The new commit (6dff7531b83c5a11dc30eac21846c2811d926659) is a trivial refactor that eliminates the use of `classmethod` and `staticmethod` and introduces no new logic, branches or other complexity.

rcritten:
I'll review this updated patch.

Incidentally I filed https://pagure.io/freeipa/issue/7293 to make some central utility function to update NSS databases and files with the CA chain. I found 4 or 5 sets of duplicate code to do the for/update loop (again, unrelated to this particular change).

rcritten:
Looks good and tests fine. I'm not sure if the last patch should be squashed into one of the others or not, though having it separate was nice for reviewing. The end result is fine it just shows some of the sausage making, not the end of the world.

I'll hold onto my ack in case you want to do that, otherwise I'll add it Monday.

frasertweedale:
Thanks @rcritten.  I'd like to keep the final patch separate.  It does show some of the sausage making, but it also addresses a separate concern from the other patches (desire to avoid class/staticmethod).

pvoborni:
Do we need to add some test to cover the scenario this PR improve?

frasertweedale:
@pvoborni yes, we should add tests.  It's a non-trivial integration test.  I filed a new issue to track it, with details about how it needs to be tested: https://pagure.io/freeipa/issue/7309

tiran:
master:

* 93d53e5cd02addf3b654f0a728bfac4d610a524f CertUpdate: make it easy to invoke from other programs
* 8960141adbd527ad249cd237f78a8ea2f7452e4e ipa-ca-install: run certupdate as initial step
* 97942a7c7aac61d657832709e2dcc45a4cb4e4df Run certupdate after promoting to CA-ful deployment
* 39fdc2d25063e9017baaa6ea1f79fc224b07f283 ipa_certupdate: avoid classmethod and staticmethod

PR #1233, opened 2017-11-06
CA-less integration tests minor log fixes

stlaz:
These changes should fix certain issues with debug logging in the CA-less tests, should we be able to get the debug logger working again.

Untested.


Rezney:
LGTM, tested and it passes...

rcritten:
I see your point about logging an object but how is the DER cert more useful? Wouldn't PEM be a better choice because then you could more easily decode what is there?

I also can't seem to find the output in the passing test anywhere to see what it actually looks like.

stlaz:
master:

* a3009b392a305fc286e43d51a72044baa0c3fd9d caless tests: make debug log of certificates sensible
* 983234c91bfe1f5a0cc6e1e35815da9f86da6136 caless tests: decode cert bytes in debug log

PR #1241, opened 2017-11-06
389-ds-base crashed as part of ipa-server-intall in ipa-uuid

tbordaz:
Bug Description:
	When adding an entry, ipa-uuid plugin may generate a unique value
	for some of its attribute.
	If the generated attribute is part of the RDN, the target DN
	is replaced on the fly and the previous one freed.
	Unfortunately, previous DN may be later used instead of
	the new one.

Fix Description:
	Make sure to use only the current DN of the operation

https://bugzilla.redhat.com/show_bug.cgi?id=1496226
https://pagure.io/freeipa/issue/7227


abbra:
LGTM.

stlaz:
master:

* 9345142c2b9938af4fa0df251489474de9f6826a 389-ds-base crashed as part of ipa-server-intall in ipa-uuid

PR #1244, opened 2017-11-07
Don't allow OTP or RADIUS in FIPS mode

stlaz:
RADIUS, which is also internally used in the process of OTP
authentication by ipa-otpd, requires MD5 checksums which
makes it impossible to be used in FIPS mode. Don't allow users
setting OTP or RADIUS authentication if in FIPS mode.

https://pagure.io/freeipa/issue/7168


abbra:
LGTM. Let's see how builds behave before ACKing.

stlaz:
master:

* 16a952a0a44a0ebee97029ea1d2f6b7593dd2622 Don't allow OTP or RADIUS in FIPS mode

PR #1254, opened 2017-11-08
Support sqlite NSSDB

tiran:
Prepare CertDB and NSSDatabase to support sqlite DB format. NSSDatabase
will automatically detect and use either old DBM or new SQL format. Old
databases are not migrated yet.

https://pagure.io/freeipa/issue/7049

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1606,9 +1606,14 @@ fi
 %ghost %attr(0644,root,apache) %config(noreplace) %{_sysconfdir}/ipa/default.conf
 %ghost %attr(0644,root,apache) %config(noreplace) %{_sysconfdir}/ipa/ca.crt
 %dir %attr(0755,root,root) %{_sysconfdir}/ipa/nssdb
+# old dbm format
 %ghost %config(noreplace) %{_sysconfdir}/ipa/nssdb/cert8.db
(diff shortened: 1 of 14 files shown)


tiran:
Test failure can be ignored. PEP 8 doesn't like the configparser import block in three files:
```
PEP-8 errors:
./ipaserver/install/ipa_backup.py:41:1: E402 module level import not at top of file
./ipaserver/install/ipa_replica_prepare.py:48:1: E402 module level import not at top of file
./ipaserver/install/ipa_restore.py:43:1: E402 module level import not at top of file
```

kaie:
Just FYI, we currently intend to keep the old default in F27. F28 will change the default

tiran:
FYI, Kai is investigating a safe upgrade path for our NSSDBs: https://bugzilla.mozilla.org/show_bug.cgi?id=1415912

rcritten:
This looks like a good start for supporting the new NSS db format. This patch shouldn't affect an install much at the moment other than backing up both styles of database. Otherwise it seems to more or less stay out of the way. Things will get a lot more interesting once the database conversion happens :-)

tiran:
master:

* 007174492908db3e3e7f45f768df1cebb79738a6 Support sqlite NSSDB

PR #1255, opened 2017-11-08
ipatests: fix circular import for collect_logs

tomaskrizek:
Move collect_logs function from util to avoid a circular import.


tomaskrizek:
master:

* c45a989506390a4bd159107566b6b2902ede5a30 ipatests: fix circular import for collect_logs

PR #1257, opened 2017-11-08
Backup ipa-custodia conf and keys

tiran:
https://pagure.io/freeipa/issue/7247

### DIFF
--- a/ipaserver/install/custodiainstance.py
+++ b/ipaserver/install/custodiainstance.py
@@ -30,8 +30,7 @@ class CustodiaInstance(SimpleServiceInstance):
     def __init__(self, host_name=None, realm=None):
         super(CustodiaInstance, self).__init__("ipa-custodia")
         self.config_file = paths.IPA_CUSTODIA_CONF
-        self.server_keys = os.path.join(paths.IPA_CUSTODIA_CONF_DIR,
-                                        'server.keys')
+        self.server_keys = paths.IPA_CUSTODIA_KEYS
         self.ldap_uri = None
         self.fqdn = host_name
         self.realm = realm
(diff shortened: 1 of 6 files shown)


ipaserver/install/custodiainstance.py
```diff
@@ -40,16 +39,18 @@ def __config_file(self):
...
+            UID=httpd_info.pw_uid,
+            GID=httpd_info.pw_gid
+        )
         conf = ipautil.template_file(template, sub_dict)
-        fd = open(self.config_file, "w+")
-        fd.write(conf)
-        fd.flush()
-        fd.close()
+        with open(self.config_file, "w") as f:
+            f.write(conf)
```
simo5: unless "with open" guarantees there is a flush you need to retain f.flush()

tiran: Yes, it does. close also flushes all buffers. Or did you mean to fsync / fdatasync the files from Kernel cache to disk?

simo5: I thought flush() causes a fsync/fdatasync, if it doesn't that's a mistake we need to fix here and elsewhere and add that.
We want to be sure keys are stored on disk before we continue here.

tiran: No, ``file.flush()`` is equivalent to ``fflush(3)``. It only flushes the internal buffer to the file descriptor. I can add fsync().

simo5: Yeah, I really think the flush() here has been used in error, and fsync() was meant to be used instead. That flush_fsync is super neat. Thanks.

rcritten:
Tested as working. Used the ipa-custodia-check command from https://github.com/freeipa/freeipa/pull/948/ to confirm that custodia was functioning properly (as well as confirming that the conf and keys were backed up and restored).

tomaskrizek:
master:

* 8bbeedc93fd442cbbb9bb70e5f446011e95211db Backup ipa-custodia conf and keys

PR #1259, opened 2017-11-08
[Backport][ipa-4-5]Fix ipa-restore (python2)

flo-renaud:

Fixes:


flo-renaud:
ipa-4-5:

* 229c84e9e6bf5db259e7768fc4da38d016b17069 Fix ipa-restore (python2)

PR #1260, opened 2017-11-08
[Backport][ipa-4-6]Fix ipa-restore (python2)

flo-renaud:
Backport of PR #1228 to ipa-4-6

Fixes:


flo-renaud:
ipa-4-6:

* 71c54ef03cabd598ff1c0032529793dbbcd59a9e Fix ipa-restore (python2)

PR #1262, opened 2017-11-09
ipa-restore: Set umask to 0022 while restoring

slaykovsky:
When some users are setting the umask to 0027 due to security
policies ipa-restore will result not working dirsrv.

So a fix is to temporary set umask to 0022 while ipa-restore is
running.

https://pagure.io/freeipa/issue/6844

### DIFF
--- a/ipaserver/install/ipa_restore.py
+++ b/ipaserver/install/ipa_restore.py
@@ -316,6 +316,9 @@ class Restore(admintool.AdminTool):
         os.chown(self.dir, pent.pw_uid, pent.pw_gid)
 
         cwd = os.getcwd()
+
+        logger.info("Temporary setting umask to 022")
+        old_umask = os.umask(0o022)
         try:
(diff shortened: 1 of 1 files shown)


felipevolpone:
Hi @slaykovsky, thanks for the patch.
It's a common practice assign the Pagure ticket to yourself once you start working on it. Also, when you have a PR, add the link to it on the ticket. You can post it as a comment or update the `on_review` field. Could you, please, do that? Thank you

slaykovsky:
Hi @felipevolpone!

The thing is I don't have permissions to do so yet :) But as soon as I will get them I will update the ticket.

Thanks!

tomaskrizek:
master:

* 28f7edaa08501fdb372ea2368b0c67aba509a55f ipa-restore: Set umask to 0022 while restoring

tiran:
The fix should be backported to 4.6, too.

PR #1266, opened 2017-11-10
test_caless: fix fix http.p12 is not valid and provide domain_level for replica tests

Rezney:


rcritten:
I don't understand what the http.p12 fix is supposed to do and the ticket holds no information on what was failing in those tests.

Why use badname with the http cert and not the dirsrv cert? Can you expand on the commit message as well?

Rezney:
I surely could have provided more info > ticket updated, however I fail to see where am I using badname with the http cert. Could you please point me to that? I will also improve the commit message afterwards. Thanks for checking this...

rcritten:
I had it reversed:

self.create_pkcs12('ca1/server-badname', filename='dirsrv.p12')

It isn't clear what this is for.

Rezney:
This is for testing the invalid_ds_cn. Bad name certificate is adding "not-" to CN e.g. not-server1.example.com. This is just a cosmetic change. I switched "replica" to "server-badname" as we have a cert for this purpose but both results to an invalid CN.

rcritten:
My point is that this isn't described in the commit message. It references a cosmetic change but this appears to have substance b/c the original cert wasn't, apparently, in fact invalid right?

Rezney:
The DS cert should always be invalid either with 'ca1/replica' or with ''ca1/server-badname'. Could you please suggest a sentence in commit message that would be passable? :)

rcritten:
I don't really know these tests so it's hard to comment in detail. I now understand the purpose of the change: using replica instead of service implicitly means that the name isn't going to match, whereas server-badname explicitly does that. So maybe something like:

Also use server-badname cert instead of cert for replica. This explicitly ensures a non-matching hostname/SAN rather than implicitly by using a certificate for the replica.

Rezney:
updated...

Rezney:
master:

* 495b85793c7cdaab58bf36d45d88e252612f150a test_caless: fix TypeError on domain_level compare
* cedd52d7f96f9d9db74b1b45e0c2a9e12532fa5c test_caless: fix http.p12 is not valid

PR #1267, opened 2017-11-10
[4.5] test_external_dns: add missing test cases

Rezney:
Add NTP, ipa-ca and ADTrust system records tests. Also test if
changes are being reflected when uninstalling a host.

The test cases are added as extension into test_dns_locations suite.

https://pagure.io/freeipa/issue/6091


tomaskrizek:
ipa-4-5:

* 200925696b39fdc396581e7a0137b053e7494f79 test_external_dns: add missing test cases

PR #1272, opened 2017-11-10
test_vault: increase WAIT_AFTER_ARCHIVE

Rezney:
Fixes failing "ipa vault-retrieve" on replica due to a vault
not already replicated. Increase from 30 to 45 seems to be enough.

### DIFF
--- a/ipatests/test_integration/test_vault.py
+++ b/ipatests/test_integration/test_vault.py
@@ -7,7 +7,7 @@ import time
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_plugins.integration import tasks
 
-WAIT_AFTER_ARCHIVE = 30  # give some time to replication
+WAIT_AFTER_ARCHIVE = 45  # give some time to replication
 
 
 class TestInstallKRA(IntegrationTest):


rcritten:
This was increased to 90 seconds two years ago (and reverted) to address some unspecified issue with vault. 45 seconds is a long time to wait merely for replication. Is there something else going on that increasing the value is going to hide? Is this indicative of performances issues (some of which IIRC we are aware of).

MartinBasti:
There used to be bug in vault when you asked dotag too early after archive operation. It says no vault found, even when vault archive operation returned success. That why it was increased temporarily to 90 seconds. Otherwise 30 seconds was enough for replication.

Shame on me I didn't add any dogtag ticket there.

tiran:
If 45 seconds sleep help, go for it.

By the way I don't like magic waits. They make tests slow and fragile.

Rezney:
I understand your point. If it was not "just" 15 sec increase I would rather open an issue for it. I made 2 runs with "45" and it was ok. I will watch the suite for some time and if it will be unstable will report it.

Rezney:
Just added ticket to the commit message...

Rezney:
master:

* d3a2a9be248a0b63f02a34ae889e6ec6dc75e2b0 test_vault: increase WAIT_AFTER_ARCHIVE

PR #1282, opened 2017-11-13
Log contents of config files when updating/writing them

rcritten:
This was originally going to be a PR to log the contents of default.conf for debugging purposes, mostly for replicas where in DL1 it gets written several times. Given that other config files like nsswitch.conf also use the same functions we get even more output as an extra benefit. The file sizes are currently not that big so it doesn't bloat the logs too much IMHO.

One patch adds the logging and the other patch changes the server installer to use the configuration writing too rather than manually creating the file.

### DIFF
--- a/ipaclient/install/ipachangeconf.py
+++ b/ipaclient/install/ipachangeconf.py
@@ -19,6 +19,7 @@
 #
 
 import fcntl
+import logging
 import os
 import shutil
 
@@ -26,6 +27,8 @@ import six
 
 if six.PY3:
     unicode = str
+
+logger = logging.getLogger(__name__)
 
 def openLocked(filename, perms):
     fd = -1
@@ -506,6 +509,8 @@ class IPAChangeConf(object):
                     f.close()
             except IOError:
                 pass
+        logger.debug("Updating configuration file %s", file)
+        logger.debug(output)
         return True
 
     def newConf(self, file, options, file_perms=0o644):
@@ -541,6 +546,8 @@ class IPAChangeConf(object):
                     f.close()
             except IOError:
                 pass
+        logger.debug("Writing configuration file %s", file)
+        logger.debug(output)
         return True
 
     @staticmethod
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -15,10 +15,11 @@ import textwrap
 
 import six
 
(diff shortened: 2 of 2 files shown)


tomaskrizek:
```
************* Module ipaserver.install.server.install
ipaserver/install/server/install.py:21: [W0611(unused-import), ] Unused format_netloc imported from ipapython.ipautil)
************* Module ipaclient.install.ipachangeconf
ipaclient/install/ipachangeconf.py:512: [W1202(logging-format-interpolation), IPAChangeConf.changeConf] Use % formatting in logging functions and pass the % parameters as arguments)
ipaclient/install/ipachangeconf.py:549: [W1202(logging-format-interpolation), IPAChangeConf.newConf] Use % formatting in logging functions and pass the % parameters as arguments)
```

rcritten:
Strange that I didn't see those lint errors when running tox locally. In any case they are fixed.

tomaskrizek:
This is odd, PR CI complains about conflict of samba-client during package installation, but this PR doesn't change any dependencies. 

@rcritten Could you please rebase to current master to see if it helps?

abbra:
We are building against older samba version (4.6.5) and trying to run with newer (4.7.0). There was ABI change between them. I think we need to upgrade our build vm template here, it seems.

tomaskrizek:
Please rebase to fix the testing jobs, it's caused by switching to F27 template.

tiran:
@rcritten Please rebase this one on latest master, too.

tiran:
I'm a bit nervous. Are we sure that we are not logging confidential informations like passphrases or PINs here?

rcritten:
I'll take a look after the new year. I'm nearly positive that IPA uses no passwords in files (all keytabs) but nearly isn't completely so I'll double-check.

rcritten:
Confirmed that no passwords are written. In my testing (full install w/CA and DNS + KRA) this will log /etc/ipa/default.conf, /etc/nsswitch.conf and /etc/openldap/ldap.conf

tiran:
master:

* 21b09522b5ff7a4a050f261dd2d42e3bdd222245 Don't manually generate default.conf in server, use IPAChangeConf
* 79432cbdc6546f66b2a641c5ac67da309f9b8d82 Log contents of files created or modified by IPAChangeConf

PR #1297, opened 2017-11-16
Remove Custodia keys on uninstall

tiran:
Keys are removed from disk and LDAP

https://pagure.io/freeipa/issue/7253

### DIFF
--- a/ipaserver/install/custodiainstance.py
+++ b/ipaserver/install/custodiainstance.py
@@ -64,10 +64,22 @@ class CustodiaInstance(SimpleServiceInstance):
                                                       realm=self.realm)
         sysupgrade.set_upgrade_state('custodia', 'installed', True)
 
+    def uninstall(self):
+        super(CustodiaInstance, self).uninstall()
+        keystore = IPAKEMKeys({
+            'server_keys': self.server_keys,
+            'ldap_uri': self.ldap_uri
+        })
+        keystore.remove_server_keys()
+        installutils.remove_file(self.config_file)
+        sysupgrade.set_upgrade_state('custodia', 'installed', False)
+
     def __gen_keys(self):
-        KeyStore = IPAKEMKeys({'server_keys': self.server_keys,
-                               'ldap_uri': self.ldap_uri})
-        KeyStore.generate_server_keys()
+        keystore = IPAKEMKeys({
+            'server_keys': self.server_keys,
+            'ldap_uri': self.ldap_uri
+        })
+        keystore.generate_server_keys()
 
     def upgrade_instance(self):
         if not sysupgrade.get_upgrade_state("custodia", "installed"):
--- a/ipaserver/secrets/kem.py
+++ b/ipaserver/secrets/kem.py
@@ -1,6 +1,8 @@
 # Copyright (C) 2015  IPA Project Contributors, see COPYING for license
 
 from __future__ import print_function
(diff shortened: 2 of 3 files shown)


ipaserver/secrets/kem.py
```diff
@@ -216,6 +235,24 @@ def generate_keys(self, servicename):
...
+        """
+        principal = '%s/%s@%s' % (servicename, self.host, self.realm)
+        ldapconn = KEMLdap(self.ldap_uri)
+        ldapconn.del_key(KEY_USAGE_SIG, principal)
+        ldapconn.del_key(KEY_USAGE_ENC, principal)
+        try:
+            os.unlink(self.config['server_keys'])
+        except OSError as e:
+            if e.errno != errno.ENOENT:
+                raise
```
rcritten: Might be more consistent to use installutils.remove_file(), perhaps extend that to grok ENOENT.

ipaserver/secrets/kem.py
```diff
@@ -124,6 +126,12 @@ def _format_public_key(self, key):
             encoding=serialization.Encoding.DER,
             format=serialization.PublicFormat.SubjectPublicKeyInfo)
 
+    def _get_dn(self, usage, principal):
+        servicename, host = principal.split('@')[0].split('/')
```
rcritten: I'd have sworn there used to be a helper to split out service, host, realm but I couldn't find it quickly.

ipaserver/install/custodiainstance.py
```diff
@@ -64,10 +65,22 @@ def create_instance(self):
...
         sysupgrade.set_upgrade_state('custodia', 'installed', True)
 
+    def uninstall(self):
+        keystore = IPAKEMKeys({
+            'server_keys': self.server_keys,
+            'ldap_uri': self.ldap_uri
+        })
+        keystore.remove_server_keys()
+        installutils.remove_file(self.config_file)
+        sysupgrade.set_upgrade_state('custodia', 'installed', False)
```
rcritten: Haven't tested yet but won't this potentially leave the service still enabled post-install?

ipaserver/secrets/kem.py
tiran: There is some code in ``ipapython.kerberos`` but it doesn't implement what I need.

rcritten: Looks like it was removed in 4.4 so this is ok.

ipaserver/secrets/kem.py
tiran: ``installutils`` is for ``ipaserver.install`` only. The function deliberately re-raises other errors, e.g. when the file cannot be removed.

ipaserver/install/custodiainstance.py
tiran: How do I make sure the service is disabled?

rcritten: My concern was removing the upgrade state but in looking further I think that is ok. I do think you should move the call to super(CustodiaInstance, self).uninstall() first so that the service gets stopped before you remove the keys.

tiran: Fixed

tiran:
master:

* 8700101d982bd3bbf08f32019567edd8f0952538 Remove Custodia keys on uninstall

PR #1301, opened 2017-11-17
Require UTF-8 fs encoding

tiran:
https://pagure.io/freeipa/issue/5887


abbra:
Changes look good. We should anyway force the server side to run under C.UTF-8 locale to prevent weird bugs like https://pagure.io/freeipa/issue/7238.

tiran:
master:

* e1bd827bbf56970ddd02ec174bf2317b64e75514 Require UTF-8 fs encoding

PR #1302, opened 2017-11-19
ipa-extdom-extop: refactor nsswitch operations

abbra:
Refactor nsswitch operations in ipa-extdom-extop plugin to allow use
of timeout-enabled nsswitch calls provided by libsss_nss_idmap.

Standard POSIX nsswitch API has no way to cancel requests which may
cause ipa-extdom-extop requests to hang far too long and potentially
exhaust LDAP server workers. In addition, glibc nsswitch API iterates
through all nsswitch modules one by one and with multiple parallel
requests a lock up may happen in an unrelated nsswitch module like
nss_files.so.2.

A solution to the latter issue is to directly load nss_sss.so.2 plugin
and utilize it. This, however, does not solve a problem with lack of
cancellable API.

With SSSD 1.16.1, libsss_nss_idmap provides a timeout-enabled variant of
nsswitch API that is directly integrated with SSSD client side machinery
used by nss_sss.so.2. As result, this API can be used instead of loading
nss_sss.so.2 directly.

To support older SSSD version, both direct loading of nss_sss.so.2 and
new timeout-enabled API are supported by this changeset. An API to
abstract both is designed to be a mix between internal glibc nsswitch
API and external nsswitch API that libsss_nss_idmap mimics. API does not
expose per-call timeout. Instead, it allows to set a timeout per
nsswitch operation context to reduce requirements on information
a caller has to maintain.

A choice which API to use is made at configure time.

In order to test the API, a cmocka test is updated to explicitly load
nss_files.so.2 as a backend. Since use of nss_sss.so.2 would always
depend on availablility of SSSD, predictable testing would not be
possible without it otherwise. Also, cmocka test does not use
nss_wrapper anymore because nss_wrapper overrides higher level glibc
nsswitch API while we are loading an individual nsswitch module
directly.

As result, cmocka test overrides fopen() call used by nss_files.so.2 to
load /etc/passwd and /etc/group. An overridden version changes paths to
/etc/passwd and /etc/group to a local test_data/passwd and
test_data/group. This way we can continue testing a backend API for
ipa-extdom-extop with the same data as with nss_wrapper.

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom.h
+++ b/daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom.h
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2017 Red Hat, Inc.
+ *
+ * This Program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 2 of the License.
+ *
+ * This Program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this Program; if not, write to the
+ *
+ *   Free Software Foundation, Inc.
+ *   59 Temple Place, Suite 330
+ *   Boston, MA 02111-1307 USA
+ *
+ */
+
+#ifndef BACK_EXTDOM_H
+#define BACK_EXTDOM_H
+#include <unistd.h>
+#include <pwd.h>
+#include <grp.h>
+
+/* Possible results of lookup using a nss_* function.
+ * Note: don't include nss.h as its path gets overriden by NSS library */
+enum nss_status {
+    NSS_STATUS_TRYAGAIN = -2,
+    NSS_STATUS_UNAVAIL,
+    NSS_STATUS_NOTFOUND,
+    NSS_STATUS_SUCCESS,
+    NSS_STATUS_RETURN
+};
+
+/* NSS backend operations implemented using either nss_sss.so.2 or libsss_nss_idmap API */
+struct nss_ops_ctx;
+
+int back_extdom_init_context(struct nss_ops_ctx **nss_context);
+void back_extdom_free_context(struct nss_ops_ctx **nss_context);
+void back_extdom_set_timeout(struct nss_ops_ctx *nss_context,
+                             unsigned int timeout);
+void back_extdom_evict_user(struct nss_ops_ctx *nss_context,
+                            const char *name);
+void back_extdom_evict_group(struct nss_ops_ctx *nss_context,
+                             const char *name);
+
+enum nss_status back_extdom_getpwnam(struct nss_ops_ctx *nss_context,
+                                     const char *name, struct passwd *pwd,
+                                     char *buffer, size_t buflen,
+                                     struct passwd **result,
+                                     int *lerrno);
+
+enum nss_status back_extdom_getpwuid(struct nss_ops_ctx *nss_context,
+                                     uid_t uid, struct passwd *pwd,
+                                     char *buffer, size_t buflen,
+                                     struct passwd **result,
+                                     int *lerrno);
+
+enum nss_status back_extdom_getgrnam(struct nss_ops_ctx *nss_context,
+                                     const char *name, struct group *grp,
+                                     char *buffer, size_t buflen,
+                                     struct group **result,
+                                     int *lerrno);
+
+enum nss_status back_extdom_getgrgid(struct nss_ops_ctx *nss_context,
+                                     gid_t gid, struct group *grp,
+                                     char *buffer, size_t buflen,
+                                     struct group **result,
+                                     int *lerrno);
+
+enum nss_status back_extdom_getgrouplist(struct nss_ops_ctx *nss_context,
+                                         const char *name, gid_t group,
+                                         gid_t *groups, int *ngroups,
+                                         int *lerrno);
+
+#endif /* BACK_EXTDOM_H */
--- a/daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
+++ b/daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
@@ -1,0 +1,276 @@
+/*
+ * Copyright 2013-2017 Red Hat, Inc.
+ *
+ * This Program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 2 of the License.
+ *
+ * This Program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this Program; if not, write to the
+ *
+ *   Free Software Foundation, Inc.
+ *   59 Temple Place, Suite 330
+ *   Boston, MA 02111-1307 USA
+ *
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <sys/types.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+#include <dlfcn.h>
+#include <errno.h>
+#include <pwd.h>
+#include <grp.h>
+#include <sys/param.h>
+#include "back_extdom.h"
+
+struct nss_ops_ctx {
+    void *dl_handle;
+    long int initgroups_start;
+
+    enum nss_status (*getpwnam_r)(const char *name, struct passwd *result,
+                                  char *buffer, size_t buflen, int *errnop);
+    enum nss_status (*getpwuid_r)(uid_t uid, struct passwd *result,
+                                  char *buffer, size_t buflen, int *errnop);
+    enum nss_status (*getgrnam_r)(const char *name, struct group *result,
+                                  char *buffer, size_t buflen, int *errnop);
+    enum nss_status (*getgrgid_r)(gid_t gid, struct group *result,
+                                  char *buffer, size_t buflen, int *errnop);
+    enum nss_status (*initgroups_dyn)(const char *user, gid_t group,
+                                      long int *start, long int *size,
+                                      gid_t **groups, long int limit,
+                                      int *errnop);
+};
+
+void back_extdom_free_context(struct nss_ops_ctx **nss_context)
+{
+    if ((nss_context == NULL) || (*nss_context == NULL)) {
+        return;
+    }
+
+    if ((*nss_context)->dl_handle != NULL) {
+        dlclose((*nss_context)->dl_handle);
+    }
+
+    free((*nss_context));
+    *nss_context = NULL;
+}
+
+int back_extdom_init_context(struct nss_ops_ctx **nss_context)
+{
+    struct nss_ops_ctx *ctx = NULL;
+
+    if (nss_context == NULL) {
+        return EINVAL;
+    }
+
+    ctx = calloc(1, sizeof(struct nss_ops_ctx));
+    if (ctx == NULL) {
+        return ENOMEM;
+    }
+    *nss_context = ctx;
+
+    ctx->dl_handle = dlopen("libnss_sss.so.2", RTLD_NOW);
+    if (ctx->dl_handle == NULL) {
+        goto fail;
+    }
+
+    ctx->getpwnam_r = dlsym(ctx->dl_handle, "_nss_sss_getpwnam_r");
+    if (ctx->getpwnam_r == NULL) {
+        goto fail;
+    }
+
+    ctx->getpwuid_r = dlsym(ctx->dl_handle, "_nss_sss_getpwuid_r");
+    if (ctx->getpwuid_r == NULL) {
+        goto fail;
+    }
+
+    ctx->getgrnam_r = dlsym(ctx->dl_handle, "_nss_sss_getgrnam_r");
+    if (ctx->getgrnam_r == NULL) {
+        goto fail;
+    }
+
+    ctx->getgrgid_r = dlsym(ctx->dl_handle, "_nss_sss_getgrgid_r");
+    if (ctx->getgrgid_r == NULL) {
+        goto fail;
+    }
+
+    ctx->initgroups_dyn = dlsym(ctx->dl_handle, "_nss_sss_initgroups_dyn");
+    if (ctx->initgroups_dyn == NULL) {
+        goto fail;
+    }
+
+    return 0;
+
+fail:
+    back_extdom_free_context(nss_context);
+
+    return EINVAL;
+}
+
+
+/* Following three functions cannot be implemented with nss_sss.so.2
+ * As result, we simply do nothing here */
+
+void back_extdom_set_timeout(struct nss_ops_ctx *nss_context,
+                             unsigned int timeout) {
+        /* no operation */
+}
+
+void back_extdom_evict_user(struct nss_ops_ctx *nss_context,
+                            const char *name) {
+        /* no operation */
+}
+
+void back_extdom_evict_group(struct nss_ops_ctx *nss_context,
+                             const char *name) {
+        /* no operation */
+}
+
+enum nss_status back_extdom_getpwnam(struct nss_ops_ctx *nss_context,
+                                     const char *name, struct passwd *pwd,
+                                     char *buffer, size_t buflen,
+                                     struct passwd **result,
+                                     int *lerrno) {
+    enum nss_status ret;
+
+    if (nss_context == NULL) {
+        return NSS_STATUS_UNAVAIL;
+    }
+
+    ret = nss_context->getpwnam_r(name, pwd,
+                                  buffer, buflen,
+                                  lerrno);
+
+    if ((ret == NSS_STATUS_SUCCESS) && (result != NULL)) {
+        *result = pwd;
+        *lerrno = 0;
+    }
+
+    return ret;
+}
+
+enum nss_status back_extdom_getpwuid(struct nss_ops_ctx *nss_context,
+                                     uid_t uid, struct passwd *pwd,
+                                     char *buffer, size_t buflen,
+                                     struct passwd **result,
+                                     int *lerrno) {
+    enum nss_status ret;
+
+    if (nss_context == NULL) {
+        return NSS_STATUS_UNAVAIL;
+    }
+
+    ret = nss_context->getpwuid_r(uid, pwd,
+                                  buffer, buflen,
+                                  lerrno);
+
+    if ((ret == NSS_STATUS_SUCCESS) && (result != NULL)) {
+        *result = pwd;
+        *lerrno = 0;
+    }
+
+    return ret;
+}
+
+enum nss_status back_extdom_getgrnam(struct nss_ops_ctx *nss_context,
+                                     const char *name, struct group *grp,
+                                     char *buffer, size_t buflen,
+                                     struct group **result,
+                                     int *lerrno) {
+    enum nss_status ret;
+
+    if (nss_context == NULL) {
+        return NSS_STATUS_UNAVAIL;
+    }
+
+    ret = nss_context->getgrnam_r(name, grp,
+                                  buffer, buflen,
+                                  lerrno);
+
+    if ((ret == NSS_STATUS_SUCCESS) && (result != NULL)) {
+        *result = grp;
+        *lerrno = 0;
+    }
+
+    return ret;
+}
+
+enum nss_status back_extdom_getgrgid(struct nss_ops_ctx *nss_context,
+                                     gid_t gid, struct group *grp,
+                                     char *buffer, size_t buflen,
+                                     struct group **result,
+                                     int *lerrno) {
+
+    enum nss_status ret;
+
+    if (nss_context == NULL) {
+        return NSS_STATUS_UNAVAIL;
+    }
+
+    ret = nss_context->getgrgid_r(gid, grp,
+                                  buffer, buflen,
+                                  lerrno);
+
+    if ((ret == NSS_STATUS_SUCCESS) && (result != NULL)) {
+        *result = grp;
+        *lerrno = 0;
+    }
+
+    return ret;
+}
+
+enum nss_status back_extdom_getgrouplist(struct nss_ops_ctx *nss_context,
+                                         const char *name, gid_t group,
+                                         gid_t *groups, int *ngroups,
+                                         int *lerrno) {
+
+    enum nss_status ret = NSS_STATUS_UNAVAIL;
+    long int tsize = MAX (1, *ngroups);
+    gid_t *newgroups = NULL;
+
+    if (nss_context == NULL) {
+        return NSS_STATUS_UNAVAIL;
+    }
+
+    newgroups = (gid_t *) calloc (tsize, sizeof (gid_t));
+    if (newgroups == NULL) {
+        *lerrno = ENOMEM;
+        return NSS_STATUS_TRYAGAIN;
+    }
+
+    newgroups[0] = group;
+    nss_context->initgroups_start = 1;
+
+    ret = nss_context->initgroups_dyn(name, group,
+                                      &nss_context->initgroups_start,
+                                      &tsize, &newgroups,
+                                      -1, lerrno);
+
+    (void) memcpy(groups, newgroups,
+                  MIN(*ngroups, nss_context->initgroups_start) * sizeof(gid_t));
+    free(newgroups);
+
+    if (*ngroups < nss_context->initgroups_start) {
+        ret = NSS_STATUS_TRYAGAIN;
+        *lerrno = ERANGE;
+    }
+
+    *ngroups = (int) nss_context->initgroups_start;
+
+    nss_context->initgroups_start = 0;
+
+    return ret;
+}
+
--- a/daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
+++ b/daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
@@ -1,0 +1,260 @@
+/*
+ * Copyright 2013-2017 Red Hat, Inc.
+ *
+ * This Program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 2 of the License.
+ *
+ * This Program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this Program; if not, write to the
+ *
+ *   Free Software Foundation, Inc.
+ *   59 Temple Place, Suite 330
+ *   Boston, MA 02111-1307 USA
+ *
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <sys/types.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+#include <errno.h>
+#include <pwd.h>
+#include <grp.h>
+#include "back_extdom.h"
+
+/* SSSD only exposes *_timeout() variants if the following symbol is defined */
+#define IPA_389DS_PLUGIN_HELPER_CALLS
+#include <sss_nss_idmap.h>
+
+struct nss_ops_ctx {
+    unsigned int timeout;
+};
+
+static enum nss_status __convert_sss_nss2nss_status(int errcode) {
+    switch(errcode) {
+    case 0:
+        return NSS_STATUS_SUCCESS;
+    case ENOENT:
+        return NSS_STATUS_NOTFOUND;
+    case ETIME:
+        /* fall-through */
+    case ERANGE:
+        return NSS_STATUS_TRYAGAIN;
+    case ETIMEDOUT:
+        /* fall-through */
+    default:
+        return NSS_STATUS_UNAVAIL;
+    }
+    return NSS_STATUS_UNAVAIL;
+}
+
+int back_extdom_init_context(struct nss_ops_ctx **nss_context)
+{
+    struct nss_ops_ctx *ctx = NULL;
+
+    if (nss_context == NULL) {
+        return EINVAL;
+    }
+
+    ctx = calloc(1, sizeof(struct nss_ops_ctx));
+
+    if (ctx == NULL) {
+        return ENOMEM;
+    }
+    *nss_context = ctx;
+    return 0;
+}
+
+void back_extdom_free_context(struct nss_ops_ctx **nss_context)
+{
+    if ((nss_context == NULL) || (*nss_context == NULL)) {
+        return;
+    }
+
+    free((*nss_context));
+    *nss_context = NULL;
+}
+
+
+void back_extdom_set_timeout(struct nss_ops_ctx *nss_context,
+                             unsigned int timeout) {
+    if (nss_context == NULL) {
+        return;
+    }
+
+    nss_context->timeout = timeout;
+}
+
+void back_extdom_evict_user(struct nss_ops_ctx *nss_context,
+                            const char *name) {
+    if (nss_context == NULL) {
+        return;
+    }
+
+    (void) sss_nss_getpwnam_timeout(name, NULL,
+                                    NULL, 0,
+                                    NULL,
+                                    SSS_NSS_EX_FLAG_INVALIDATE_CACHE,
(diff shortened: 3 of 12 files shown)


abbra:
It seems travis-ci PR requests fail constantly lately. I'm not sure what is causing it -- in one run I saw it just timing out. In this one I see it is crashing during execution of a cmocka test in ipa-extdom-extop plugin which is also executed in the fedora-26/build test. In the latter it passes just fine and I cannot reproduce this crash locally either. I need a help from anyone who maintains travis-ci/pr to understand why and how the crash happens.

abbra:
Finally, this PR completed all checks and can be reviewed/merged.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
```diff
@@ -0,0 +1,279 @@
...
+			  char *buffer, size_t buflen, int *errnop);
+	enum nss_status (*initgroups_dyn)(const char *user, gid_t group,
+			      long int *start, long int *size,
+			      gid_t **groups, long int limit,
+			      int *errnop);
+};
+
+void back_extdom_free_context(struct nss_ops_ctx **nss_context)
+{
+	if (nss_context == NULL) {
```
tiran: I think you should also add a check for *nss_context == NULL here

```
if ((nss_context == NULL) || (*nss_context == NULL)) {
    return;
}
```

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
```diff
@@ -0,0 +1,279 @@
...
+		return;
+	}
+
+	ctx = calloc(1, sizeof(struct nss_ops_ctx));
+
+	*nss_context = ctx;
+	if (ctx == NULL) {
+		return;
+	}
```
tiran: Minor style nit pick, this looks backwards. I think the check for NULL should come first, then the assignment. It doesn't change the program flow, though.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
```diff
@@ -0,0 +1,279 @@
...
+}
+
+enum nss_status back_extdom_getgrouplist(struct nss_ops_ctx *nss_context,
+					 const char *name, gid_t group,
+					 gid_t *groups, int *ngroups,
+					 int *lerrno) {
+
+	enum nss_status ret = NSS_STATUS_UNAVAIL;
+	long int tsize = MAX (1, *ngroups);
+	gid_t *newgroups = (gid_t *) malloc (tsize * sizeof (gid_t));
```
tiran: ``calloc(tsize, sizeof(gid_t))`` ?

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -406,35 +448,36 @@ int check_request(struct extdom_req *req, enum extdom_version version)
...
+                return LDAP_OPERATIONS_ERROR;
+            }
+            groups = new_groups;
+	}
 
-        ret = getgrouplist(name, gid, groups, &ngroups);
-        if (ret == -1) {
-            free(groups);
-            return LDAP_OPERATIONS_ERROR;
+        if (rc == NSS_STATUS_NOTFOUND) {
```
tiran: I think you should also handle ``lerrno == ENOMEM`` here. Otherwise you'd end up in an infinitive loop.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom.h
```diff
@@ -0,0 +1,80 @@
...
+
+/* NSS backend operations implemented using either nss_sss.so.2 or libsss_nss_idmap API */
+struct nss_ops_ctx;
+
+void back_extdom_init_context(struct nss_ops_ctx **nss_context);
+void back_extdom_free_context(struct nss_ops_ctx **nss_context);
+void back_extdom_set_timeout(struct nss_ops_ctx *nss_context,
+			     unsigned int timeout);
+void back_extdom_evict_user(struct nss_ops_ctx *nss_context,
+			    const char *name);
```
tiran: indent?

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom.h
```diff
@@ -0,0 +1,80 @@
...
+void back_extdom_evict_group(struct nss_ops_ctx *nss_context,
+			     const char *name);
```
tiran: indent?

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom.h
```diff
@@ -0,0 +1,80 @@
...
+  NSS_STATUS_UNAVAIL,
+  NSS_STATUS_NOTFOUND,
+  NSS_STATUS_SUCCESS,
+  NSS_STATUS_RETURN
+};
+
+/* NSS backend operations implemented using either nss_sss.so.2 or libsss_nss_idmap API */
+struct nss_ops_ctx;
+
+void back_extdom_init_context(struct nss_ops_ctx **nss_context);
```
tiran: Is there a reason to return void? I'd prefer if you make the function return an error indicator to handle ``ENOMEN`` etc.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom.h
abbra: nope, this is one of rendering issues for unified diffs when tabs and spaces are used like in this code.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom.h
abbra: same here.

abbra:
Pushed updated version.

abbra:
Finally, my code is passing all tests. However, travis failed tox-py36 and one of the integration tests. Tox tests failures are due to unrelated issues in a certificate tests. Integration tests failed due to known pyasn.1 issue in python-pyldap.

Given that both failures aren't related to this PR, I'm basically stuck.

tiran:
The IPv6 issues affects other PRs, too. Can you move https://github.com/freeipa/freeipa/pull/1302/commits/d8c3090195b824e7499c59c5612cab888b91c33f to a separate PR?

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
```diff
@@ -0,0 +1,278 @@
...
+				     struct passwd **result,
+				     int *lerrno) {
+	enum nss_status ret;
+
+	if (nss_context == NULL) {
+		return NSS_STATUS_UNAVAIL;
+	}
+
+	ret = (enum nss_status)
+		nss_context->getpwnam_r(name, pwd,
```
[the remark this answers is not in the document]
simo5: actually do not defined ret as enum nss_status at all.
Define it as the type returned naturally by getpwnam_r() and then let the compiler perform an implicit cast when it compares the result later on.
If you want to be fancy add a function that checks that ret is a valid NSS_STATUS value and returns the enum, and use it to return from this function.
something like
```
enum nss_status check_valid_status(int status) {
    switch(status) {
    case NSS_STATUS_TRYAGAIN:
        return NSS_STATUS_TRYAGAIN;
    [...]
    default:
        /* abort() ?
         * or define an invalid status and return something like */
        return NSS_STATUS_INVALID;
}

then return check_valid_status(ret);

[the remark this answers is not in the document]
Comment (simo5) [reply daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c]: Same ofr the rest of the functions ...

File: daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c [hunk 9]
```diff
@@ -0,0 +1,278 @@
...
+	return ret;
+}
+
+enum nss_status back_extdom_getgrouplist(struct nss_ops_ctx *nss_context,
+					 const char *name, gid_t group,
+					 gid_t *groups, int *ngroups,
+					 int *lerrno) {
+
+	enum nss_status ret = NSS_STATUS_UNAVAIL;
+	long int tsize = MAX (1, *ngroups);
```
simo5: Should we have a ceiling on how much memory we are willing to allocate ?
Do we need to check for negative numbers here ?
can we define ngroups as unsigned ?

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
```diff
@@ -0,0 +1,243 @@
...
+
+struct nss_ops_ctx {
+	unsigned int timeout;
+};
+
+static enum nss_status __convert_sss_nss2nss_status(int errcode) {
+	enum nss_status ret = NSS_STATUS_UNAVAIL;
+
+	if (errcode == 0) {
+		ret = NSS_STATUS_SUCCESS;
```
simo5: Please use a switch/case construct here (like the one I proposed above, with direct returns.
Properly handle the default case.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
```diff
@@ -0,0 +1,243 @@
...
+
+	return ret;
+}
+
+int back_extdom_init_context(struct nss_ops_ctx **nss_context)
+{
+	struct nss_ops_ctx *ctx = NULL;
+
+	if (nss_context == NULL) {
+		return -1;
```
simo5: Why are we returning -1 here ?
Can we use things like EINVAL ?

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
```diff
@@ -0,0 +1,243 @@
...
+void back_extdom_set_timeout(struct nss_ops_ctx *nss_context,
+			     unsigned int timeout) {
+	if (nss_context == NULL) {
+		return;
+	}
+
+	nss_context->timeout = timeout;
+}
+
+/* TODO: handle buffers and memory allocation in this function */
```
simo5: what buffers need to be handled ?

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
```diff
@@ -0,0 +1,243 @@
...
+void back_extdom_evict_user(struct nss_ops_ctx *nss_context,
+			    const char *name) {
+	if (nss_context == NULL) {
+		return;
+	}
+
+	(void) sss_nss_getpwnam_timeout(name, NULL,
```
simo5: I do not understand this call, are we overloading getpwnam semantics to delete users from cache by setting empy buffers ?
Or is this function just incomplete ?
Why don't we just have a sss_nss_evict_timeout() call ?

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -97,25 +98,48 @@ static int inc_buffer(size_t buf_max, size_t *_buf_len, char **_buf)
     return 0;
 }
 
-int getpwnam_r_wrapper(size_t buf_max, const char *name,
+int __nss_to_err(enum nss_status errcode)
+{
+    int ret = -1;
+
+    if (errcode == NSS_STATUS_SUCCESS) {
```
simo5: please use switch/case here

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -97,25 +98,48 @@ static int inc_buffer(size_t buf_max, size_t *_buf_len, char **_buf)
...
     buf = *_buf;
     buf_len = *_buf_len;
 
-    while (buf != NULL
-            && (ret = getpwnam_r(name, pwd, buf, buf_len, &result)) == ERANGE) {
-        ret = inc_buffer(buf_max, &buf_len, &buf);
-        if (ret != 0) {
-            if (ret == ERANGE) {
-                LOG("Buffer too small, increase ipaExtdomMaxNssBufSize.\n");
+    while (buf != NULL) {
```
simo5: while you are changing this, improve it this way:
for (buf = *_buf, buf_len = *_buf_len; buf != NULL;) {

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -97,25 +98,48 @@ static int inc_buffer(size_t buf_max, size_t *_buf_len, char **_buf)
...
+int __nss_to_err(enum nss_status errcode)
+{
+    int ret = -1;
+
+    if (errcode == NSS_STATUS_SUCCESS) {
+        ret = 0;
+    } else if (errcode == NSS_STATUS_NOTFOUND) {
+        ret = ENOENT;
+    } else if(errcode == NSS_STATUS_TRYAGAIN) {
+        ret = ERANGE;
```
simo5: This mapping seem wrong, why ERANGE ?
This sounds more like EAGAIN

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -97,25 +98,48 @@ static int inc_buffer(size_t buf_max, size_t *_buf_len, char **_buf)
...
 
-    while (buf != NULL
-            && (ret = getpwnam_r(name, pwd, buf, buf_len, &result)) == ERANGE) {
-        ret = inc_buffer(buf_max, &buf_len, &buf);
-        if (ret != 0) {
-            if (ret == ERANGE) {
-                LOG("Buffer too small, increase ipaExtdomMaxNssBufSize.\n");
+    while (buf != NULL) {
+        rc = back_extdom_getpwnam(ctx->nss_ctx, name, pwd, buf, buf_len, &result, &lerrno);
+        if (rc == NSS_STATUS_TRYAGAIN) {
```
simo5: wrap the previous call with _nss_to_err() and check for EAGAIN here.

simo5: As in:
```
ret = _nss_to_err(back_extdom...(...));
if (ret == EAGAIN) {
    ...

File: daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c [hunk 18]
```diff
@@ -97,25 +98,48 @@ static int inc_buffer(size_t buf_max, size_t *_buf_len, char **_buf)
...
+            ret = inc_buffer(ctx->max_nss_buf_size, &buf_len, &buf);
+            if (ret != 0) {
+                if (ret == ERANGE) {
+                    LOG("Buffer too small, increase ipaExtdomMaxNssBufSize.\n");
```
simo5: Move this into the error handler (after the label 'done')

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -97,25 +98,48 @@ static int inc_buffer(size_t buf_max, size_t *_buf_len, char **_buf)
...
+                }
+                goto done;
             }
-            goto done;
+        } else {
+            ret = __nss_to_err(rc);
+            break;
         }
     }
```
simo5: move the following line after the 'done' label if you make the changes above

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
simo5: meaning you just straight goto done on error

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -130,25 +154,31 @@ int getpwnam_r_wrapper(size_t buf_max, const char *name,
     return ret;
```
simo5: before returning handle the logging and setting ENOENT in a switch/case statement:
```
done:
    /* set buf/buf_len */
switch(ret) {
case 0:
    if (result == 0) ret = ENOENT;
    break;
case ERANGE:
    LOG("Buffer too small, increase ipaExtdomMaxNssBufSize.\n");
default:
    break;
}
return ret;

File: daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c (hunk 4 above)
Comment (simo5) [reply daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c]: already handled above, the break below will get out of the loop, but the structure o this cycle is bad, see my comments

File: daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c [hunk 21]
```diff
@@ -406,35 +448,39 @@ int check_request(struct extdom_req *req, enum extdom_version version)
...
 
-    ret = getgrouplist(name, gid, groups, &ngroups);
-    if (ret == -1) {
-        new_groups = realloc(groups, ngroups * sizeof(gid_t));
-        if (new_groups == NULL) {
-            free(groups);
-            return LDAP_OPERATIONS_ERROR;
+    do {
+        rc = back_extdom_getgrouplist(ctx->nss_ctx, name, gid, groups, &ngroups, &lerrno);
+        if (rc == NSS_STATUS_TRYAGAIN) {
```
simo5: I do not like this do() while loop, because it does not tell me what the terminating conditions really are.
It seem unlikely you want to loop forever on error, and it seem you check only for NSS_STATUS_TRYAGAIN, but again the problem is that I cannot be sure because of the coide structure (and no comments help either).

Can we do:
for(rc = NSS_STATUS_TRYAGAIN; rc == NSS_STATUS_TRYAGAIN;) {
instead ?
This will make use break out on any rc error that is *not* NSS_STATUS_TRYAGAIN

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -406,35 +448,39 @@ int check_request(struct extdom_req *req, enum extdom_version version)
...
+            new_groups = NULL;
+            if (lerrno == ERANGE) {
+                new_groups = realloc(groups, ngroups * sizeof(gid_t));
+            }
+            if ((new_groups == NULL) || (lerrno == ENOMEM)) {
+                free(groups);
+                return LDAP_OPERATIONS_ERROR;
+            }
+            groups = new_groups;
         }
```
simo5: when adding the above for loop add rc = NSS_STATUS_TRYAGAIN here

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_extop.c
```diff
@@ -202,6 +205,16 @@ static int ipa_extdom_init_ctx(Slapi_PBlock *pb, struct ipa_extdom_ctx **_ctx)
     }
     LOG("Maximal nss buffer size set to [%zu]!\n", ctx->max_nss_buf_size);
 
+
+    back_extdom_init_context(&ctx->nss_ctx);
+    timeout = slapi_entry_attr_get_uint(e,
+                                        "ipaExtdomMaxNssTimeout");
```
simo5: why line break ? it does not exceed 80 cols.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom.h
```diff
@@ -0,0 +1,79 @@
...
+
+/* Possible results of lookup using a nss_* function.
+ * Note: don't include nss.h as its path gets overriden by NSS library */
+enum nss_status {
+        NSS_STATUS_TRYAGAIN = -2,
+        NSS_STATUS_UNAVAIL,
+        NSS_STATUS_NOTFOUND,
+        NSS_STATUS_SUCCESS,
+        NSS_STATUS_RETURN
+};
```
simo5: bad indent again, this needs to be 4 spaces!

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
abbra: Because it is handled this way on nsswitch side. ERANGE is returned when insufficient buffer is supplied. There is no EAGAIN returned -- see getgrnam(3) or getpwnam(3).

abbra:
Updated the code according to @simo5 comments.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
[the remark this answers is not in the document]
simo5: Not handled in the rebase

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
abbra: Because this is what SSSD gives us.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
simo5: Not handled

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
simo5: Not handled

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
simo5: Not handled

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
simo5: there is another one as well

abbra: I removed this todo: after a rewrite of the initial version it became irrelevant.

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -97,135 +98,180 @@ static int inc_buffer(size_t buf_max, size_t *_buf_len, char **_buf)
...
 
-int getpwnam_r_wrapper(size_t buf_max, const char *name,
+int __nss_to_err(enum nss_status errcode)
+{
+    int ret = -1;
+
+    switch(errcode) {
+    case NSS_STATUS_SUCCESS:
+        ret = 0;
+        break;
```
simo5: please immediately return and eliminate the variable ret entirely

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -97,135 +98,180 @@ static int inc_buffer(size_t buf_max, size_t *_buf_len, char **_buf)
...
+        rc = back_extdom_getpwnam(ctx->nss_ctx, name, pwd, buf, buf_len, &result, &lerrno);
+        ret = __nss_to_err(rc);
+        if (ret == ERANGE) {
+            ret = inc_buffer(ctx->max_nss_buf_size, &buf_len, &buf);
+            if (ret != 0) {
+                goto done;
             }
-            goto done;
+        } else {
+            break;
```
simo5: the intention of my previous comment was to remove this else branch, sorry for not being clear

simo5: it seem like buf != NULL is not really the condition you are looking for in this cycle.
You want to break when ret != ERANGE, then maybe do that ?
The point is to express in the for loop what are the initial conditions and the exit condition, so that a future reader knows exactly what we care for and pay attention to that specific detail and not anciallry variables that just happen to allow us to do a loop:
```
buf = *_buf;
buf_len = *_buf_len;
for (ret = ERANGE; ret != ERANGE;) {
```

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
simo5: ok makes sense

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
abbra: ngroups is int in the nss interface -- see `getgrouplist(3)`. It is one of few areas where glibc and POSIX aren't consistent -- `initgroups_dyn()` API requires `long int` for number of groups while `getgrouplist()` which is an upper level API for `initgroups_dyn()` uses `int`.

abbra: We don't need to check on negative numbers because we do have `MAX(1, *ngroups)`, so we are guaranteed at least `1` here.

Also, the calls to `back_extdom_getgrouplist()` have no influence from external sources. Namely, `get_user_grouplist()` in ipa_extdom_common.c sets ngroups to 128 and does not depend on any user input.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
abbra: I cannot change existing SSSD API right now, especially for RHEL. @sumit-bose  was saying it is not as easy to pass through a new call in the sssd pipe, I think.

abbra:
Pushed new update which should address most if not all @simo5 comments now.

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -97,134 +98,137 @@ static int inc_buffer(size_t buf_max, size_t *_buf_len, char **_buf)
...
-            }
-            goto done;
+int getpwnam_r_wrapper(struct ipa_extdom_ctx *ctx, const char *name,
+                       struct passwd *pwd, char **buf, size_t *buf_len)
+{
+    int ret, lerrno = 0;
+    struct passwd *result = NULL;
+    enum nss_status rc;
+
+    for(rc = NSS_STATUS_TRYAGAIN; rc == NSS_STATUS_TRYAGAIN;) {
```
frozencemetery: I think this would be more clear as a `do/while` loop.  (Same suggestion in similar cases elsewhere.)

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
```diff
@@ -0,0 +1,241 @@
...
+    case ETIME:
+        /* fall-through */
+    case ERANGE:
+        return NSS_STATUS_TRYAGAIN;
+    case ETIMEDOUT:
+        /* fall-through */
+    default:
+        return NSS_STATUS_UNAVAIL;
+    }
+    return NSS_STATUS_UNAVAIL;
```
simo5: there should be no need for this last return, unless things like coverity complain, but it is ok to leave it in

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -97,134 +98,137 @@ static int inc_buffer(size_t buf_max, size_t *_buf_len, char **_buf)
...
     }
 
-    if (ret == 0 && result == NULL) {
-        ret = ENOENT;
-    }
-
 done:
-    *_buf = buf;
-    *_buf_len = buf_len;
+    switch(ret) {
```
frozencemetery: I think control flow is more clear here as an `if` - something like `if (ret == ERANGE) { ... } else if (result == NULL) { ... } return ret`.  Same suggestion for other nearby functions.

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_extop.c
```diff
@@ -202,6 +205,20 @@ static int ipa_extdom_init_ctx(Slapi_PBlock *pb, struct ipa_extdom_ctx **_ctx)
     }
     LOG("Maximal nss buffer size set to [%zu]!\n", ctx->max_nss_buf_size);
 
+
+    ret = back_extdom_init_context(&ctx->nss_ctx);
+    if(ret != 0) {
```
simo5: space after 'if'

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_cmocka_tests.c
```diff
@@ -31,24 +32,166 @@
...
+struct test_data {
+    struct extdom_req *req;
+    struct ipa_extdom_ctx *ctx;
+};
+
+/*
+ * redefine logging for mocks
+ */
+#ifdef __GNUC__
+    __attribute__((format(printf, 3, 4)))
```
frozencemetery: I recommend not adding these to the project and instead turning off the gcc warning during build instead.  (They do not exist elsewhere in the codebase as far as I can tell.)

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
```diff
@@ -0,0 +1,241 @@
...
+    switch(errcode) {
+    case 0:
+        return NSS_STATUS_SUCCESS;
+    case ENOENT:
+        return NSS_STATUS_NOTFOUND;
+    case ETIME:
+        /* fall-through */
+    case ERANGE:
+        return NSS_STATUS_TRYAGAIN;
+    case ETIMEDOUT:
```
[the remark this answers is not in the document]
abbra: Because we expect ETIMEDOUT to be returned. It is explicit code reading contract here. THink about yourself in 6 months.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
```diff
@@ -0,0 +1,273 @@
...
+                                     char *buffer, size_t buflen,
+                                     struct group **result,
+                                     int *lerrno) {
+    enum nss_status ret;
+
+    if (nss_context == NULL) {
+        return NSS_STATUS_UNAVAIL;
+    }
+
+    ret = (enum nss_status) nss_context->getgrnam_r(name, grp,
```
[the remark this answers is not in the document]
abbra: Yes, just missed it.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
```diff
@@ -0,0 +1,273 @@
...
+                                         gid_t *groups, int *ngroups,
+                                         int *lerrno) {
+
+    enum nss_status ret = NSS_STATUS_UNAVAIL;
+    long int tsize = MAX (1, *ngroups);
+    gid_t *newgroups = (gid_t *) calloc (tsize, sizeof (gid_t));
+
+
+    if (nss_context == NULL) {
+        return NSS_STATUS_UNAVAIL;
```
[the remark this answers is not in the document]
abbra: Thanks, moved further in.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    ret = sss_nss_getpwnam_timeout(name, pwd,
+                                   buffer, buflen,
+                                   result,
+                                   SSS_NSS_EX_FLAG_NO_FLAGS,
+                                   nss_context->timeout);
+
+    if (ret != 0 && lerrno != NULL) {
+        /* SSSD translates errno into return code */
```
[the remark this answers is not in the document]
abbra: I added a longer comment. A particular issue is that we don't have this documented anywhere else so one have to read both glibc and sssd code to know details of what is returned and how.

daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_nss_sss.c
```diff
@@ -0,0 +1,278 @@
...
+                            const char *name) {
+        /* no operation */
+}
+
+void back_extdom_evict_group(struct nss_ops_ctx *nss_context,
+                             const char *name) {
+        /* no operation */
+}
+
+#include <stdio.h>
```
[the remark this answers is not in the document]
abbra: was already removed. I used it for debugging getgrouplist() at some point.

abbra:
Re-pushed to force travis build

abbra:
master:

* 78ad1cfe4f8f2f9c51acbe8044a652e49ebaf1d7 ipa-extdom-extop: refactor nsswitch operations

PR #1305, opened 2017-11-20
libotp: add libraries after objects

tiran:
https://pagure.io/freeipa/issue/7189

Original patch by Rob Crittenden

### DIFF
--- a/daemons/ipa-slapi-plugins/libotp/Makefile.am
+++ b/daemons/ipa-slapi-plugins/libotp/Makefile.am
@@ -15,4 +15,4 @@ libotp_la_LIBADD = libhotp.la
 
 check_PROGRAMS = t_hotp
 TESTS = $(check_PROGRAMS)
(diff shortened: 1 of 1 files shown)


abbra:
LGTM.

stlaz:
The commit message does not say anything useful, please fix it.

tiran:
@stlaz There isn't anything useful to say. The title of the commit contains all relevant information.

rcritten:
Can we compromise and add something about linker and libotp to the first line?

If you want to expand on linker ordering or something that's fine too I guess. What I suspect happens is the default linker options were different for this guy's distribution hence this working for many years before this popping up.

stlaz:
@rcritten thanks. @tiran a commit which only tells you the original patch was made by someone else does not say you much when browsing history, this is much better.

tiran:
master:

* ba037a3551feb835b4e0880c4ba19d68214a1b96 libotp: add libraries after objects

PR #1306, opened 2017-11-20
Prevent installation of Py2 and Py3 mod_wsgi

tiran:
FreeIPA is either compatible with Python 2 mod_wsgi or Python 3
mod_wsgi. mod_wsgi can not coexist in the same Apache process as
mod_wsgi_python3. When both mod_wsgi and python3-mod_wsgi are installed,
the first loaded module wins and the other one is never loaded.

Add conflict on the other module to prevent installation of both
modules.

https://pagure.io/freeipa/issue/7161

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -304,14 +304,16 @@ Requires: httpd >= 2.4.6-31
 Requires(preun): python3
 Requires(postun): python3
 Requires: python3-gssapi >= 1.2.0-5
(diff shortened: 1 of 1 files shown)


stlaz:
LGTM.

tiran:
master:

* 57787f647e0f98ffd75dc18b41201fd2601a9160 Prevent installation of Py2 and Py3 mod_wsgi

PR #1293, opened 2017-11-16
Run tox tests for PyPI packages on Travis

tiran:
Obsoletes: #846

PR #1308, opened 2017-11-20
[Backport][ipa-4-5] Run tox tests for PyPI packages on Travis

tomaskrizek:
Original PR: #1293

### DIFF
--- a/.test_runner_config.yaml
+++ b/.test_runner_config.yaml
@@ -28,7 +28,7 @@ steps:
   builddep:
   - rm -rf /var/cache/dnf/*
   - "dnf makecache fast || :"
-  - dnf builddep -y ${builddep_opts} --spec freeipa.spec.in --best --allowerasing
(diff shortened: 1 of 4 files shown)


tiran:
``with_wheel`` section in the spec file is missing dependency on ``python2-tox``

tomaskrizek:
I see... In that case, perhaps we shouldn't add tox testing to `ipa-4-5` branch. I don't want to introduce another build dependency to stable version. Most patches go through `master` or `ipa-4-6` anyway, so this shouldn't be much of an issue.

@tiran What do you think?

tomaskrizek:
The `python-tox` dependency is used only in the `with_wheels` section, so it doesn't affect the build dependencies for rpm builds.

tomaskrizek:
We'd also have to backport at least 3a5b3be8b92a509d207d814c9fe294ee7b4e81c4 to get tox testing in `ipa-4-5`. Considering not much development is going on in this branch, I don't think it's worth it.

PR #1310, opened 2017-11-21
6577 ca replica install certupdate2

tiran:


rcritten:
I'm satisfied with Fraser's structured programming approach, can we close this?

PR #1312, opened 2017-11-21
ipalib/frontend.py output_for_cli loops optimization

slaykovsky:
Trivial fix which removes unnecessary for loops.

### DIFF
--- a/ipalib/frontend.py
+++ b/ipalib/frontend.py
@@ -1041,7 +1041,15 @@ class Command(HasParam):
 
         self.log_messages(output)
 
-        order = [p.name for p in self.output_params()]
+        order = []
+        labels = {}
+        flags = {}
+
+        for p in self.output_params():
+            order.append(p.name)
+            labels[p.name] = unicode(p.label)
+            flags[p.name] = p.flags
+
         if options.get('all', False):
             order.insert(0, 'dn')
             print_all = True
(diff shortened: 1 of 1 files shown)


ipalib/frontend.py
```diff
@@ -1041,18 +1043,23 @@ def output_for_cli(self, textui, output, *args, **options):
 
         self.log_messages(output)
 
-        order = [p.name for p in self.output_params()]
+        order = deque()
+        labels = dict()
```
tiran:
```
labels = {}
flags = {}
```

It's shorter, easier to read, and faster.

ipalib/frontend.py
```diff
@@ -1041,18 +1043,23 @@ def output_for_cli(self, textui, output, *args, **options):
...
+        flags = dict()
+
+        for p in self.output_params():
+            order.append(p.name)
+            labels[p.name] = unicode(p.label)
+            flags[p.name] = p.flags
+
         if options.get('all', False):
-            order.insert(0, 'dn')
+            order.appendleft('dn')
```
tiran: Unless it's performed in a loop, one ``insert(0, item)`` is ok. A deque is overkill. It requires more memory and doesn't amortize that fast.

ipalib/frontend.py
slaykovsky: Okay.

ipalib/frontend.py
slaykovsky: Will return list then :)

tiran:
Good catch! 🥇

tomaskrizek:
Please rebase to fix the testing jobs, it's caused by switching to F27 template.

felipevolpone:
@slaykovsky could you, please, rebase this PR to PR CI get it right?

tiran:
master:

* 197b5ca639ab31bc5f079c45eb36763c442bf74c ipalib/frontend.py output_for_cli loops optimization

PR #1320, opened 2017-11-22
  ipatest: replica install with existing entry on master 

mrizwan93:
replica install might fail because of existing entry for replica like
`cn=ipa-http-delegation,cn=s4u2proxy,cn=etc,$SUFFIX` etc. The situation
may arise due to incorrect uninstall of replica or ipa server-del is
not executed on master.

related bug : https://pagure.io/freeipa/issue/7174

Fixes: https://pagure.io/freeipa/issue/7276

### DIFF
--- a/ipatests/test_integration/test_replication_layouts.py
+++ b/ipatests/test_integration/test_replication_layouts.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2015  FreeIPA Contributors see COPYING for license
+# Copyright (C) 2017  FreeIPA Contributors see COPYING for license
 #
 
 import time
--- a/ipatests/test_integration/test_replica_promotion.py
+++ b/ipatests/test_integration/test_replica_promotion.py
@@ -3,6 +3,8 @@
 #
 
 import time
+from tempfile import NamedTemporaryFile
+import textwrap
 import pytest
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_plugins.integration import tasks
@@ -483,3 +485,45 @@ class TestRenewalMaster(IntegrationTest):
         assert("IPA CA renewal master: %s" % self.master.hostname in result), (
             "Master hostname not found among CA renewal masters"
         )
+
+
+class TestReplicaInstallWithExistingEntry(IntegrationTest):
+    """replica install might fail because of existing entry for replica like
+    `cn=ipa-http-delegation,cn=s4u2proxy,cn=etc,$SUFFIX` etc. The situation
+    may arise due to incorrect uninstall of replica.
+
+    https://pagure.io/freeipa/issue/7174"""
+
+    num_replicas = 1
+
+    def test_replica_install_with_existing_entry(self):
+        master = self.master
+        tasks.install_master(master)
+        replica = self.replicas[0]
+        tf = NamedTemporaryFile()
+        ldif_file = tf.name
+        base_dn = "dc=%s" % (",dc=".join(replica.domain.name.split(".")))
+        # adding entry for replica on master so that master will have it before
+        # replica installtion begins and creates a situation for pagure-7174
+        entry_ldif = textwrap.dedent("""
+            dn: cn=ipa-http-delegation,cn=s4u2proxy,cn=etc,{base_dn}
+            changetype: modify
+            add: memberPrincipal
+            memberPrincipal: HTTP/{hostname}@{realm}
+
+            dn: cn=ipa-ldap-delegation-targets,cn=s4u2proxy,cn=etc,{base_dn}
+            changetype: modify
+            add: memberPrincipal
+            memberPrincipal: ldap/{hostname}@{realm}""").format(
+            base_dn=base_dn, hostname=replica.hostname,
+            realm=replica.domain.name.upper())
+        master.put_file_contents(ldif_file, entry_ldif)
+        arg = ['ldapmodify',
+               '-h', master.hostname,
+               '-p', '389', '-D',
+               str(master.config.dirman_dn),   # pylint: disable=no-member
+               '-w', master.config.dirman_password,
+               '-f', ldif_file]
+        master.run_command(arg)
+
+        tasks.install_replica(master, replica)


ipatests/test_integration/test_bugzilla.py
```diff
@@ -0,0 +1,71 @@
+#
+# Copyright (C) 2015  FreeIPA Contributors see COPYING for license
```
tiran: It's 2017

ipatests/test_integration/test_bugzilla.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_replica_install_with_existing_entry(self):
+        result = tasks.install_master(self.master)
+        assert result.returncode == 0
+        master = self.master
+        replica = self.replicas[0]
+        tf = NamedTemporaryFile()
+        ldif_file = tf.name
+        base_dn = "dc=%s" % (",dc=".join(replica.domain.name.split(".")))
+        entry_ldif = """
```
tiran: The block is unreadable. Please indent / dedent the block correctly and use ```format()``` with kwargs:
```
    entry_ldif = textwrap.dedent("""
        dn: cn=ipa-http-delegation,cn=s4u2proxy,cn=etc,{base_dn}\n
        ...
    """).format(
        base_dn=base_dn, hostname=..., realm=...
    )
```

stlaz:
@mrizwan93 thank you for your first PR. Please modify your commit message (and the title of this PR) to reflect what's going on in here.
Please imagine that one day you may be browsing through the history of git changes and if you see `ipatest: automation for bz-1493145`, it won't help you much because you don't know what happened there in a Bugzilla somewhere, you may not necessarily have access to it and/or you just don't care enough and will try to see what's changed and whether that's a cause of what you need to fix or not.
All the above means one thing - you want a commit message clearly stating what the commit should be doing so that it won't slow you down in the future.

pvoborni:
Hi Rizwan, thanks for the patch. 

I'd like to comment on the general approach rather then the specifics to set a precedence.

The goal is to create a test which would prevent a bug from happening again. In this case it was a failure during replica installation where replica was not removed correctly.  From upstream perspective the fact that it was originally filed as bugzilla bug 1493145 is not really important.

1. Wrt linked issue:
I don't see much value in opening a issue https://pagure.io/freeipa/issue/7276 given following reasons:
* from my point of view not creating a new upstream issue for the test would be ok if the pull request would mention bug https://pagure.io/freeipa/issue/7174
* creating a new issue is also ok, but the issue is mentioning a bugzilla. If it should mention a bug it automates then it should be related upstream ticket. In this case https://pagure.io/freeipa/issue/7174 (available from the bz)
* the bug description doesn't help much as it doesn't say what is the use case it tests. It mentions the bz where one maybe could get the use case but in the BZ it doesn't seem obvious as well (this is a different problem). 
* if use case to test is not clear then it is an opportunity for conversation with fix author

2. Where the test lives

* please let us avoid having suites called: `test_bugzilla.py` Bugzillas are irrelevant from upstream PoV. But even having `test_issues.py` is not good IMO.  It is not about the issue itself. It is about use case. And the test should live in a suite related to the use case.  In this case it would be probably test_installation.py or something else related to replica installation.

ipatests/test_integration/test_bugzilla.py
```diff
@@ -0,0 +1,71 @@
...
+        memberPrincipal: HTTP/%s@%s\n
+
+        dn: cn=ipa-ldap-delegation-targets,cn=s4u2proxy,cn=etc,%s\n
+        changetype: modify\n
+        add: memberPrincipal\n
+        memberPrincipal: ldap/%s@%s""" % (
+        base_dn, replica.hostname, replica.domain.name.upper(),
+        base_dn, replica.hostname, replica.domain.name.upper())
+        master.put_file_contents(ldif_file, entry_ldif)
+        arg = ['ldapmodify', '-h', master.hostname, '-p', '389',
```
pvoborni: It also a deserves a comment what is the intention as the next reader will most likely not know. Somethink like:

This simulates a situation where `ipa server-del` is not called on master but only as part of replica uninstallation and where there deletions of `ipa-http-delegation,cn=s4u2proxy` entries are not propagated to master.

ipatests/test_integration/test_bugzilla.py
```diff
@@ -0,0 +1,71 @@
...
+        for r in self.replicas:
+            r.run_command(['ipa', 'user-show', test_user])
+
+
+class TestReplicaInstallWithExistingEntry(LayoutsBaseTest):
+    """Automation for:
+        https://bugzilla.redhat.com/show_bug.cgi?id=1493145
+        https://bugzilla.redhat.com/show_bug.cgi?id=1506913
+        ipa-replica-install might fail because of an already existing
+        entry cn=ipa-http-delegation,cn=s4u2proxy,cn=etc,$SUFFIX"""
```
pvoborni: I'd expect a comment that says something like: 
```
Tests that replica can be installed even if there are some remnant entries like 
`cn=ipa-http-delegation,cn=s4u2proxy,cn=etc,$SUFFIX` after incorrect replica 
uninstall (ipa server-del not called on master).

related bug: https://pagure.io/freeipa/issue/7174
```

ipatests/test_integration/test_bugzilla.py
```diff
@@ -0,0 +1,71 @@
...
+        # tests use custom installation
+        pass
+
+    def replication_is_working(self):
+        test_user = 'replication-testuser'
+        self.master.run_command(
+            ['ipa', 'user-add', test_user, '--first', 'test', '--last', 'user']
+        )
+
+        time.sleep(60)  # make sure the replication of user is done
```
pvoborni: Maybe Ok for now, but doesn't look very robust (it doesn't ensure anything, only hopes that 60s is enough).

ipatests/test_integration/test_bugzilla.py
pvoborni: @stlaz  or other, do you know if we already have some tasks which would do LDAP modify in tests? Seems like a common task to me.

ipatests/test_integration/test_replication_layouts.py
```diff
@@ -178,3 +181,47 @@ def test_2_connected_topology_with_ca_kra(self):
...
+        assert result.returncode == 0
+        master = self.master
+        replica = self.replicas[0]
+        tf = NamedTemporaryFile()
+        ldif_file = tf.name
+        base_dn = "dc=%s" % (",dc=".join(replica.domain.name.split(".")))
+        #adding entry for replica on master so that master will have it before
+        # replica installtion begins and creates a situation for pagure-7174
+        entry_ldif = textwrap.dedent("""
+        dn: cn=ipa-http-delegation,cn=s4u2proxy,cn=etc,{base_dn}\n
```
tiran: please indent the text block. You also don't need to add ``\n`` because triple-quoted strings retain line breaks.

ipatests/test_integration/test_replication_layouts.py
```diff
@@ -178,3 +181,47 @@ def test_2_connected_topology_with_ca_kra(self):
...
+        changetype: modify\n
+        add: memberPrincipal\n
+        memberPrincipal: HTTP/{hostname}@{realm}\n
+
+        dn: cn=ipa-ldap-delegation-targets,cn=s4u2proxy,cn=etc,{base_dn}\n
+        changetype: modify\n
+        add: memberPrincipal\n
+        memberPrincipal: ldap/{hostname}@{realm}""").format(
+        base_dn=base_dn, hostname=replica.hostname,
+        realm=replica.domain.name.upper())
```
tiran: This line and the line before also need to be indented by four spaces.

```
        entry_ldif = textwrap.dedent("""
            dn: cn=ipa-http-delegation,cn=s4u2proxy,cn=etc,{base_dn}
            changetype: modify
            add: memberPrincipal
            memberPrincipal: HTTP/{hostname}@{realm}
            dn: cn=ipa-ldap-delegation-targets,cn=s4u2proxy,cn=etc,{base_dn}
            changetype: modify
            add: memberPrincipal
            memberPrincipal: ldap/{hostname}@{realm}
        """).format(
            base_dn=base_dn, hostname=replica.hostname,
            realm=replica.domain.name.upper()
        )
```

ipatests/test_integration/test_replication_layouts.py
```diff
@@ -178,3 +181,47 @@ def test_2_connected_topology_with_ca_kra(self):
...
+        master.put_file_contents(ldif_file, entry_ldif)
+        arg = ['ldapmodify', '-h', master.hostname, '-p', '389',
+               '-D', str(master.config.dirman_dn), '-w',
+               master.config.dirman_password, '-f', ldif_file]
+        cmd = master.run_command(arg, raiseonerr=False)
+        assert cmd.returncode == 0, cmd.stdout_text
```
tiran: Why are you using ``raiseonerr=False`` and then checking the result for error? How about you use the default and let the command fail with a nice exception.

ipatests/test_integration/test_replication_layouts.py
```diff
@@ -178,3 +181,47 @@ def test_2_connected_topology_with_ca_kra(self):
...
+    `cn=ipa-http-delegation,cn=s4u2proxy,cn=etc,$SUFFIX` etc. The situation
+    may arise due to incorrect uninstall of replica.
+
+    https://pagure.io/freeipa/issue/7174"""
+
+    num_replicas = 1
+
+    def test_replica_install_with_existing_entry(self):
+        result = tasks.install_master(self.master)
+        assert result.returncode == 0
```
tiran: You don't need to check the return code yourself. ``install_master()`` has ``raiseonerr=True``.

ipatests/test_integration/test_replication_layouts.py
```diff
@@ -178,3 +181,47 @@ def test_2_connected_topology_with_ca_kra(self):
...
+        realm=replica.domain.name.upper())
+        master.put_file_contents(ldif_file, entry_ldif)
+        arg = ['ldapmodify', '-h', master.hostname, '-p', '389',
+               '-D', str(master.config.dirman_dn), '-w',
+               master.config.dirman_password, '-f', ldif_file]
+        cmd = master.run_command(arg, raiseonerr=False)
+        assert cmd.returncode == 0, cmd.stdout_text
+
+        cmd = tasks.install_replica(master, replica)
+        assert cmd.returncode == 0, cmd.stdout_text
```
tiran: Same here, ``install_replica()`` also has ``raiseonerr=True`` by default.

tiran:
pylint also complains:

```
ipatests/test_integration/test_replication_layouts.py:227: [C0305(trailing-newlines), ] Trailing newlines)
ipatests/test_integration/test_replication_layouts.py:218: [E1101(no-member), TestReplicaInstallWithExistingEntry.test_replica_install_with_existing_entry] Class 'config' has no 'dirman_dn' member)
```

ipatests/test_integration/test_replication_layouts.py
```diff
@@ -178,3 +181,47 @@ def test_2_connected_topology_with_ca_kra(self):
...
+
+    def test_replica_install_with_existing_entry(self):
+        result = tasks.install_master(self.master)
+        assert result.returncode == 0
+        master = self.master
+        replica = self.replicas[0]
+        tf = NamedTemporaryFile()
+        ldif_file = tf.name
+        base_dn = "dc=%s" % (",dc=".join(replica.domain.name.split(".")))
+        #adding entry for replica on master so that master will have it before
```
pvoborni: I've a feeling that not have a space after `#` will fail in pep8 check.

felipevolpone:
@mrizwan93 Please rebase to fix the testing jobs, it's caused by switching to F27 template.

mrizwan93:
@felipevolpone @tiran  I am seeing error in travis-ci like:
~~~~
1 )
11-28 05:30 ipadocker.command INFO     Executing command: dnf builddep -y -D "with_lint 1" --spec freeipa.spec.in --best --allowerasing
11-28 05:30 ipadocker.cli ERROR    Command dnf builddep -y -D "with_lint 1" --spec freeipa.spec.in --best --allowerasing failed (exit code 1)


2)
The downloaded packages were saved in cache until the next successful transaction.
You can remove cached packages by executing 'dnf clean packages'.
Error: Error downloading packages:
  Curl error (28): Timeout was reached for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f27&arch=x86_64 [Operation timed out after 30001 milliseconds with 0 out of 0 bytes received]
11-28 05:30 ipadocker.cli ERROR    Command dnf builddep -y -D "with_lint 1" --spec freeipa.spec.in --best --allowerasing failed (exit code 1)

3)
11-28 05:34 ipadocker.command INFO     Executing command: tox -e py27,py36,pypi,pylint3
11-28 05:42 ipadocker.cli ERROR    Command tox -e py27,py36,pypi,pylint3 failed (exit code 1)
~~~~

mrizwan93:
closing and reopeninig pr for triggering CI.

mrizwan93:
closing and reopeninig pr for triggering CI.

mrizwan93:
seeing following error in travis CI:
~~~~
The command "travis_wait 50 ./.travis_run_task.sh" exited with 1.

0.01s$ test -z "`cat $PEP8_ERROR_LOG`"

cat: pycodestyle_errors.log: No such file or directory
~~~~~

closing and reopeninig pr for triggering CI.

tiran:
The xmlrpc test case ``0036: dnsrecord_show: Show record 'testdnsres' in zone 'dnszone.test' with --structured and --all options`` has been failing on Python 3 in multiple PRs.

stlaz:
Could we maybe incorporate this in some other replica-install test so that we don't have to install a new master again just for this test?

mrizwan93:
@stlaz It is simulation of the scenario where ipa server-del didn't ran on master after replica uninstall or ipa-server-install --uninstall failed on replica and the entry for replica server exists in the directory server (i.e stale entry). This scenario makes a valid test case I believe, hence it is written separately and installing master as well.

stlaz:
Alright, fine, I suppose you can call it a separate scenario, I see it's being handled in our test suite like that anyway.
Just please move the test from `test_replication_layouts.py` to `test_replica_promotion.py`, you're not adding any new replication layout here.

tiran:
master:

* feee70d7bb90c05141ac15c058c2e26e2e0ab339 ipatest: replica install with existing entry on master

PR #1327, opened 2017-11-23
ensuring 389-ds plugins are enabled after install

akokshar:
To avoid problems caused by desabled plugins on 389-ds side
explicitly enable plugins required by IPA

https://pagure.io/freeipa/issue/7271

### DIFF
--- a/install/updates/20-enable_dirsrv_plugins.update
+++ b/install/updates/20-enable_dirsrv_plugins.update
@@ -1,0 +1,76 @@
+# 7-bit check, plugins, config
+dn: cn=7-bit check,cn=plugins,cn=config
+replace: nsslapd-pluginEnabled:off::on
+
+# Account Usability Plugin, plugins, config
+dn: cn=Account Usability Plugin,cn=plugins,cn=config
+replace: nsslapd-pluginEnabled:off::on
+
+# ACL Plugin, plugins, config
+dn: cn=ACL Plugin,cn=plugins,cn=config
+replace: nsslapd-pluginEnabled:off::on
+
+# ACL preoperation, plugins, config
+dn: cn=ACL preoperation,cn=plugins,cn=config
+replace: nsslapd-pluginEnabled:off::on
+
+# Auto Membership Plugin, plugins, config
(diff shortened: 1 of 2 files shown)


Firstyear:
Hey mate, this is great. I'm happy with this, because it means you can safely trust all plugins will work on upgrade, even if change 389 code.

I'm happy to ack this from the 389 side :) Thanks you so much for the fast patch given we only had this idea monday!

Firstyear:
PS: I think you can eliminate 7-bit check, , HTTP Client,, Internationalization Plugin, Schema Reload, and Views.

tiran:
Should we also ensure plugins in ``ipaserver.install.server.upgrade.upgrade_configuration()``?

Firstyear:
Yes @tiran that's a great idea.

akokshar:
@abbra, thank you. Will try to transform it into an update file.

This was one of my main question too. As I unerstood we want to make sure that IPA is not impacted in case 389-ds guys switch off some of the plugins which were previously enabled by default. So, 
- I interrupted installer at the point it just installed and started ldap
- checked what plugins are enabled 
- explicitly reenable them as a first step of the installer.

This way, we can be sure that all plugins we want (and might be more) are enabled.

abbra:
I think you can re-use updater's infrastructure (e.g. loading and applying update files) in installer too. This way we could have both goals achieved.

abbra:
@akokshar  please don't use merge to update your tree to master. Instead, use 'git rebase' and force push. Right now your commit 3227f6b includes a lot of unrelated changes.

akokshar:
@abbra, my bad. cleaned it already.

I did transform the patch to use update file. Could you take a look please?

abbra:
Looks good now. Thank you!

tiran:
@Firstyear @abbra The PR looks ready to me.

abbra:
I did LGTM already and as it passed all tests, ACK.

tiran:
master:

* fbb188976637ac64e9661fec6cc51fa7215290a4 ensuring 389-ds plugins are enabled after install

PR #1333, opened 2017-11-24
Fix ca less IPA install on fips mode

flo-renaud:
When ipa-server-install is run in fips mode and ca-less, the installer
fails when the keys are provided with --{http|dirsrv|pkinit}-cert-file
in a separate key file.

The installer transforms the key into PKCS#8 format using
openssl pkcs8 -topk8
but this command fails on a fips-enabled server, unless the options
-v2 aes256 -v2prf hmacWithSHA256
are also provided.

Fixes:
https://pagure.io/freeipa/issue/7280

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -605,9 +605,13 @@ class NSSDatabase(object):
                                 "Can't load private key from both %s and %s" %
                                 (key_file, filename))
 
+                        # the args -v2 aes256 -v2prf hmacWithSHA256 are needed
+                        # on OpenSSL 1.0.2 (fips mode). As soon as FreeIPA
+                        # requires OpenSSL 1.1.0 we'll be able to drop them
                         args = [
                             paths.OPENSSL, 'pkcs8',
                             '-topk8',
+                            '-v2', 'aes256', '-v2prf', 'hmacWithSHA256',
                             '-passout', 'file:' + self.pwd_file,
                         ]
                         if ((label != b'PRIVATE KEY' and key_password) or


tiran:
``-v2 aes256 -v2prf hmacWithSHA256`` are the default values in OpenSSL 1.1.0. Is this only required for OpenSSL 1.0.2?

tiran:
## man pkcs8(1) from OpenSSL 1.1.0g on Fedora 27
```
       -v2 alg
           This option sets the PKCS#5 v2.0 algorithm.

           The alg argument is the encryption algorithm to use, valid values include aes128, aes256 and des3. If this option isn't specified then aes256
           is used.

       -v2prf alg
           This option sets the PRF algorithm to use with PKCS#5 v2.0. A typical value value would be hmacWithSHA256. If this option isn't set then the
           default for the cipher is used or hmacWithSHA256 if there is no default.

           Some implementations may not support custom PRF algorithms and may require the hmacWithSHA1 option to work.
```

## man pkcs8(1) from OpenSSL 1.0.2k on RHEL 7.5
```
       -v2 alg
           This option enables the use of PKCS#5 v2.0 algorithms. Normally PKCS#8 private keys are encrypted with the password based encryption
           algorithm called pbeWithMD5AndDES-CBC this uses 56 bit DES encryption but it was the strongest encryption algorithm supported in PKCS#5 v1.5.
           Using the -v2 option PKCS#5 v2.0 algorithms are used which can use any encryption algorithm such as 168 bit triple DES or 128 bit RC2 however
           not many implementations support PKCS#5 v2.0 yet. If you are just using private keys with OpenSSL then this doesn't matter.

           The alg argument is the encryption algorithm to use, valid values include des, des3 and rc2. It is recommended that des3 is used.

       -v2prf alg
           This option sets the PRF algorithm to use with PKCS#5 v2.0. A typical value values would be hmacWithSHA256. If this option isn't set then the
           default for the cipher is used or hmacWithSHA1 if there is no default.
```

flo-renaud:
master:

* 19138c5ba3dd8fafcab766427cf025848d5d6016 Fix ca less IPA install on fips mode

PR #1347, opened 2017-11-29
Prevent set_directive from clobbering other keys

frasertweedale:
`set_directive` only looks for a prefix of the line matching the
given directive (key).  If a directive is encountered for which the
given key is prefix, it will be vanquished.

This occurs in the case of `{ca,kra}.sslserver.cert[req]`; the
`cert` directive gets updated after certificate renewal, and the
`certreq` directive gets clobbered.  This can cause failures later
on during KRA installation, and possibly cloning.

Match the whole directive to avoid this issue.

Fixes: https://pagure.io/freeipa/issue/7288

-----

Cause: corner case.

How to test:

1. ensure `ca.sslserver.certreq=<base64 CSR>` exists in `ca/CS.cfg`.
2. resubmit Certmonger tracking request for `Server-Cert cert-pki-ca` Dogtag system cert.
3. verify that `ca.sslserver.certreq=<base64 CSR>` still exists in `ca/CS.cfg`.

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -952,7 +952,7 @@ class CAInstance(DogtagInstance):
         installutils.set_directive(caconfig, 'ca.publish.rule.instance.FileCrlRule.enable', 'true', quotes=False, separator='=')
         installutils.set_directive(caconfig, 'ca.publish.rule.instance.FileCrlRule.mapper', 'NoMap', quotes=False, separator='=')
         installutils.set_directive(caconfig, 'ca.publish.rule.instance.FileCrlRule.pluginName', 'Rule', quotes=False, separator='=')
-        installutils.set_directive(caconfig, 'ca.publish.rule.instance.FileCrlRule.predicate=', '', quotes=False, separator='')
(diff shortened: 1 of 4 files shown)


flo-renaud:
Hi @frasertweedale ,
thank you for the patch, works as expected.
I tested multiple scenarios:
- CA-less install + ipa-ca-install + cert renewal + ipa-kra-install
- self-signed CA install + cert renewal + ipa-kra-install
- update of directives when the separator is a space (for instance NSSNickname in /etc/httpd/conf.d/nss.conf)

frasertweedale:
Thanks for reviewing @flo-renaud!

pvoborni:
Can it have a unit test?

frasertweedale:
@pvoborni yep, fair enough.  There ya go :)

stlaz:
Breaks tests.

frasertweedale:
@stlaz other recent PRs are failing with same errors.  Seems to be a regression on master or in the
CI image?  Any ideas what caused it?

frasertweedale:
Rebased; tests are now working.  Leaving the ACK off to make sure someone eyeballs the two
new commits :)

tiran:
I created PR #1379 on top of your PR, @frasertweedale

frasertweedale:
@tiran I like your work.  I'll close this PR and we can continue with #1379.

PR #1351, opened 2017-11-30
NSSDB: use preferred convert command

tiran:
After further testing, Kai Engert proposed to use -N with -f -@ to
convert a NSSDB from DBM to SQL format.

https://fedoraproject.org/wiki/Changes/NSSDefaultFileFormatSql#Upgrade.2Fcompatibility_impact

https://pagure.io/freeipa/issue/7049

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -347,6 +347,15 @@ class NSSDatabase(object):
 
     def convert_db(self, rename_old=True):
         """Convert DBM database format to SQL database format
+
+        **WARNING** **WARNING** **WARNING** **WARNING** **WARNING**
+
+        The caller must ensure that no other process or service is
+        accessing the NSSDB during migration. The DBM format does not support
+        multiple processes. If more than one process opens a DBM NSSDB for
+        writing, the database will become **irreparably corrupted**.
+
+        **WARNING** **WARNING** **WARNING** **WARNING** **WARNING**
         """
         if (self.dbtype == 'sql' or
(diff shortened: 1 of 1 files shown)


rcritten:
Given that the converted database needs be open writeable should  a comment be added that any services using the database should be shut down at time of conversion?

tiran:
I *think* the old DB is open read-only. The new SQL database can be safely opened writeable by multiple processes. Let's ask Kai.

kaie:
If another process is currently using the old DB, you shouldn't open it for conversion.
Parallel access of dbm databases is not recommended.
I don't know if the migration is guaranteed to open the old dbm read-only. Regardless of that, you that the process running in parallel has opened the database read-write, which might be sufficient for a bad migration.
I concur with Rob's suggestion to play safe and recommend to shutdown all other processes prior to migrating the dbm database.

tiran:
Thanks @kaie

I'm adding a warning to the function.

tiran:
@rcritten Now comes with a scary **WARNING** **WARNING** **WARNING** **WARNING** **WARNING**

rcritten:
I wonder if that is too subtle :-)

LGTM, will ack once CI passes

tiran:
@rcritten CI is passing after rebase

rcritten:
LGTM.

tiran:
master:

* 1505922c2b8cac82395909df6ee18e203776528d NSSDB: use preferred convert command

PR #1353, opened 2017-12-01
Add script to regenerate Custodia keys and config

tiran:
/usr/libexec/ipa/ipa-custodia-regen can be used to recover from broken
ipa-custodia keys and configuration. It removes existing config and
server key files, regenerates new files and uploads the new public keys
to LDAP.

https://pagure.io/freeipa/issue/6893


ipaserver/install/ipa_custodia_regen.py
```diff
@@ -0,0 +1,117 @@
+# Copyright (C) 2017  IPA Project Contributors, see COPYING for license
+"""Regenerate Custodia server.keys and config
+
+KEEP THIS SCRIPT COMPATIBLE WITH FREEIPA 4.5
```
simo5: This kind of notice usually does not work, do we have a way to check compatibility in some tests ?

ipaserver/install/ipa_custodia_regen.py
```diff
@@ -0,0 +1,117 @@
...
+class IPACustodiaRegen(admintool.AdminTool):
+    command_name = 'ipa-custodia-regen'
+
+    usage = "%prog [options]\n"
+
+    @classmethod
+    def add_options(cls, parser):
+        super(IPACustodiaRegen, cls).add_options(parser)
+        parser.add_option(
+            "-r", '--regenerate', action="store_true",
```
simo5: What is being regenreated with this option ?
The tool is called ipa-custodia-*regen* already, what should the user expect by running it with or without this option, why is it necessary ?

ipaserver/install/ipa_custodia_regen.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def validate_options(self):
+        super(IPACustodiaRegen, self).validate_options(needs_root=True)
+        try:
+            installutils.check_server_configuration()
+        except RuntimeError as e:
+            raise admintool.ScriptError(e)
+        if not self.options.regenerate:
+            raise admintool.ScriptError(
+                "Nothing to do, --regenerate option not given."
```
simo5: What's the point of requireing an option when that is the *only* option ?

ipaserver/install/ipa_custodia_regen.py
```diff
@@ -0,0 +1,117 @@
...
+            api.bootstrap(
+                in_server=True,
+                debug=self.options.verbose,
+                confdir=paths.ETC_IPA,
+            )
+            api.finalize()
+        if not api.Backend.ldap2.isconnected():
+            api.Backend.ldap2.connect()
+
+        self.regen()
```
simo5: maybe run a "consistency check" when --regenerate is not passed, that will tell you if running again with --regenerate is needed at all ?

ipaserver/install/ipa_custodia_regen.py
tiran: No, I don't see a way to check compatibility.

ipaserver/install/ipa_custodia_regen.py
tiran: The script does nothing without an option. It's a poor man's version of **read the documentation first**.

simo5:
After more explantations online I think this tool really should not exist and the --regenerate functionality belongs to the ipa-custodia-check tool.
Nack in this form.

tiran:
I have now combined checking and regeneration as one script. Since IPA uses optparse and optparse does not support positional arguments like argparse, the calling syntax of ipa-custodia-check has changed. It's now necassary to use ``ipa-custodia-check --check=hostname`` instead of ``ipa-custodia-check hostname``.

tiran:
With #1507, it's now possible to force key regeneration by removing the file and running ipa-server-upgrade. I'll change this PR to handle regeneration differently. I also need to take care of LWCA replication keys for @frasertweedale

PR #1354, opened 2017-12-04
test_backup_and_restore.py AssertionError fix

slaykovsky:
prefix in the backup function expects output to have
'ipa.ipaserver.install.ipa_backup.Backup:' and it's wrong. The right
one is 'ipaserver.install.ipa_backup:'.

### DIFF
--- a/ipatests/test_integration/test_backup_and_restore.py
+++ b/ipatests/test_integration/test_backup_and_restore.py
@@ -126,15 +126,13 @@ def backup(host):
 
     # Get the backup location from the command's output
     for line in result.stderr_text.splitlines():
-        prefix = ('ipa.ipaserver.install.ipa_backup.Backup: '
-                  'INFO: Backed up to ')
+        prefix = 'ipaserver.install.ipa_backup: INFO: Backed up to '
         if line.startswith(prefix):
             backup_path = line[len(prefix):].strip()
             logger.info('Backup path for %s is %s', host, backup_path)
             return backup_path
     else:
         raise AssertionError('Backup directory not found in output')
-
 
 
 class TestBackupAndRestore(IntegrationTest):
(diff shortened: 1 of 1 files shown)


abbra:
LGTM. Do we want to wait until PRCI starts taking the jobs again?

felipevolpone:
LGTM as well. I'm testing the changes [here](https://github.com/felipevolpone/freeipa/pull/66). Once it's green, I'll ack it

felipevolpone:
@slaykovsky How were you able to reproduce this error? Running tests with ipa-vagrant, in-tree, or what?
I'm asking because when running them [on pr ci](https://fedorapeople.org/groups/freeipa/prci/jobs/c1330528-d8f8-11e7-9568-0cc47a70b1ae/), I don't see this error message.

ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -126,7 +126,7 @@ def backup(host):
 
     # Get the backup location from the command's output
     for line in result.stderr_text.splitlines():
-        prefix = ('ipa.ipaserver.install.ipa_backup.Backup: '
+        prefix = ('ipaserver.install.ipa_backup: '
```
tiran: This code segment is broken and does not what you think it is doing. You are passing a tuple to ``startswith``. That's a multi-match. That's totally fine. Next up you are using the length of the **tuple** to strip the prefix from the line -- which is always 2 here. You want the length of the matching **string** instead.

tiran: Oh, it's not a tuple but a long string within parenthesis... Can you fit the entire string on one line?

slaykovsky: Sure, yes

ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -126,8 +126,7 @@ def backup(host):
 
     # Get the backup location from the command's output
     for line in result.stderr_text.splitlines():
-        prefix = ('ipa.ipaserver.install.ipa_backup.Backup: '
-                  'INFO: Backed up to ')
+        prefix = ('ipaserver.install.ipa_backup: INFO: Backed up to ')
```
tiran: Please drop the parenthesis, they are not necessary. Pylint / pep8 should warn about them, too. Did we disable the checker?

slaykovsky: Thanks, done.

felipevolpone:
Tested and it works properly. ack

tiran:
master:

* a70dcb1e1846d6c46d4bdadc048e61911ce57311 test_backup_and_restore.py AssertionError fix

PR #1357, opened 2017-12-05
Use correct version of Python in RPM scripts

frasertweedale:
Fixes: https://pagure.io/freeipa/issue/7299


frasertweedale:
@tiran yeah I hear ya.  Especially for testing whether FreeIPA server is installed, would be good to add an `ipactl` subcommand to do that.

PR #1359, opened 2017-12-06
install: report CA Subject DN and subject base to be used

frasertweedale:
Currently we do not report what Subject DN or subject base will be
used for the CA installation.  This leads to situations where the
administrator wants a different Subject DN later.  Display these
data as part of the "summary" prior to the final go/no-go prompt in
ipa-server-install and ipa-ca-install.

The go/no-go prompt in ipa-ca-install is new.  It is suppressed for
unattended installations.

Fixes: https://pagure.io/freeipa/issue/7246

### DIFF
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -18,6 +18,8 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+from __future__ import print_function
+
 import logging
 import sys
 import os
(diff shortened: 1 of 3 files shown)


install/tools/ipa-ca-install
```diff
@@ -241,6 +243,14 @@ def install_master(safe_options, options):
         sys.exit("CA subject: {}".format(e))
 
     ca.install_check(True, None, options)
+
+    if not options.unattended:
+        ca.print_ca_configuration(options)
```
tiran: Would you mind to print the CA configuration in both supervised and unattended mode?

frasertweedale: Sure, I'll push update tomorrow.  Thanks for reviewing!

tiran:
@frasertweedale FYI, ipatool was not able to merge this PR. I rebased the PR to upstream/master and forced pushed the rebase into your repository.

tiran:
master:

* 34f73b4a94ce9622380228825aa15d5b12c1bb6f install: report CA Subject DN and subject base to be used

PR #1364, opened 2017-12-06
Improve help message for ipa trust-add --range-type

flo-renaud:
Add the correct procedure for re-running ipa trust-add with a different
range type.

Fixes:
https://pagure.io/freeipa/issue/7308


tiran:
master:

* 891cced446db0514bfe85c94c014ade32deb6cae Improve help message for ipa trust-add --range-type

PR #1365, opened 2017-12-07
Add uniqueness constraint on CA ACL name

frasertweedale:
It is possible to add caacl entries with same "name" (cn).  The
command is supposed to prevent this but direct LDAP operations allow
it and doing that will cause subsequent errors.

Enable the DS uniqueness constraint plugin for the cn attribute in
CA ACL entries.

Fixes: https://pagure.io/freeipa/issue/7304

### DIFF
--- a/install/updates/10-uniqueness.update
+++ b/install/updates/10-uniqueness.update
@@ -92,3 +92,20 @@ add:uniqueness-across-all-subtrees: on
 dn: cn=ipaUniqueID uniqueness,cn=plugins,cn=config
 add:uniqueness-exclude-subtrees: cn=staged users,cn=accounts,cn=provisioning,$SUFFIX
 add:uniqueness-across-all-subtrees: on
+
+dn: cn=caacl name uniqueness,cn=plugins,cn=config
+default:objectClass: top
+default:objectClass: nsSlapdPlugin
+default:objectClass: extensibleObject
+default:cn: caacl name uniqueness
+default:nsslapd-pluginDescription: Enforce unique attribute values
+default:nsslapd-pluginPath: libattr-unique-plugin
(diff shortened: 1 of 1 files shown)


Firstyear:
Hey there,

Just want to comment on this one.

So the "uiduniq" feature of directory server has a couple of gotchas.

First, we don't actually check *existing* data for uniquness, only *new* data. So existing violations will go un-noticed.

Second, it's possible to have a race condition between two masters where they both add a record with unique properties to an object, then on replication, we can't reject *either* change, so the uniqueness is no longer upheld.

So a better idea to do this, is actually to have objects with a singleValued attr (preferably rdn). Then when you duplicate, we can use syntax level updates that will hold that only one value was used. One object or the other will survive, the other will be rejected. 

However, the issue there is that you don't know your object was rejected til some point in time after (once it becomes a conflict entry ...)

There are some unique challenges to async multi master systems ;) ensuring consistency like this is best effort, not a guarantee.

tiran:
IMHO the unique index on caacl CN is good enough for now. New CA ACLs are rarely created. I suggest to submit this PR and create another ticket to review all existing unique indexes later.

Firstyear:
Okay, if it's rare to update, the likelihood of this is low, so okay by me.

abbra:
The patch looks good too, and @Firstyear  explanation makes clear the risk is low, so ACK.

abbra:
I added a ticket https://pagure.io/freeipa/issue/7320 to record @Firstyear comment about potential collision risk.

tiran:
master:

* 0e9ce73a52c953f7e798150c6e720c41594a99b0 Add uniqueness constraint on CA ACL name

PR #1366, opened 2017-12-07
Re-enable some KRA installation tests

frasertweedale:
Some KRA installation tests were disabled due to failures caused by
security domain session replication lag.  This problem has been
addressed in Dogtag by introducing a default 5 second sleep after
security domain login, to give more time for session data to be
replicated to other hosts.  There is still a possibility for this
kind of failure, but the delay minimises it.

FreeIPA depends on the version of Dogtag that contains this change,
so remove the failing-test annotations.

Fixes: https://pagure.io/freeipa/issue/7220


tiran:
master:

* ba411b0f6d0559078dc99e201abc068b64637ad6 Re-enable some KRA installation tests

PR #1378, opened 2017-12-08
test_server_del: Added automation for bugzilla bz1506188.

ndehadrai:
Patch1:
  1) Added automation for verification steps for bz1506188.
  2) The code setusp up IPA-Master/ Replica and then checks for
     deletion of Replica using 'ipa server-del' command.
  3) It also checks for verification of 'ipa dnsserver-find' command
     output once the replica server details are removed.

 Patch2:
  1) Removed unsed assert statement inside code.
  2) Realigned code for doc-string information within
     80 character limit length.

Signed-off-by: Nikhil Dehadrai <ndehadra@redhat.com>

### DIFF
--- a/ipatests/pytest_plugins/integration/tasks.py
+++ b/ipatests/pytest_plugins/integration/tasks.py
@@ -1381,3 +1381,29 @@ def add_dns_zone(master, zone, skip_overlap_check=False,
                                     host.hostname + ".", '--a-rec', host.ip])
     else:
         logger.debug('Zone %s already added.', zone)
+
+
+def run_dnsserver_find(host, server_to_find, returncode=None):
+    """
+    Find DNS-server details
+    """
+    kinit_admin(host)
+    cmd = host.run_command(["ipa", "dnsserver-find", server_to_find],
+                           raiseonerr=False)
+    if cmd.returncode == returncode:
+        print cmd.stdout_text
+    else:
+        print cmd.stderr_text
+
+
+def run_server_find(host, server_to_find, returncode=None):
+    """
+    Find server details
+    """
+    kinit_admin(host)
+    cmd = host.run_command(["ipa", "server-find", server_to_find],
+                           raiseonerr=False)
+    if cmd.returncode == returncode:
+        print cmd.stdout_text
+    else:
+        print cmd.stderr_text
--- a/ipatests/test_integration/test_server_del.py
+++ b/ipatests/test_integration/test_server_del.py
@@ -300,3 +300,29 @@ class TestLastServices(ServerDelBase):
             self.replicas[0], self.master.hostname,
             ignore_last_of_role=True
         )
+
+
+class TestServerDelReplicaRemoval(IntegrationTest):
+    """
+    Test the checks for Removal of Replica during server-del
+    as part for verification of bz1506188
+    """
+    num_replicas = 1
+    domain_level = DOMAIN_LEVEL_1
+    topology = 'line'
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_topo(
+            cls.topology, cls.master, cls.replicas, [],
+            domain_level=cls.domain_level, setup_replica_cas=False)
+
+    def test_removal_of_replica(self):
+        """
+        test that removal of replica bz1506188.
+        """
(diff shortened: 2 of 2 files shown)


ipatests/pytest_plugins/integration/tasks.py
```diff
@@ -1381,3 +1381,29 @@ def add_dns_zone(master, zone, skip_overlap_check=False,
...
+
+def run_dnsserver_find(host, server_to_find, returncode=None):
+    """
+    Find DNS-server details
+    """
+    kinit_admin(host)
+    cmd = host.run_command(["ipa", "dnsserver-find", server_to_find],
+                           raiseonerr=False)
+    if cmd.returncode == returncode:
+        print cmd.stdout_text
```
tiran: You cannot use print statement because it's not compatible with Python 3. If you **have** to use print, then you must add ``from __future__ import print_function`` to the top of the file and use the print **function**. In general it's a bad idea to print in tests.

ipatests/test_integration/test_server_del.py
```diff
@@ -300,3 +300,29 @@ def test_forced_removal_of_master(self):
...
+        tasks.install_topo(
+            cls.topology, cls.master, cls.replicas, [],
+            domain_level=cls.domain_level, setup_replica_cas=False)
+
+    def test_removal_of_replica(self):
+        """
+        test that removal of replica bz1506188.
+        """
+        tasks.run_server_find(self.master, self.replicas[0].hostname, 0)
+        tasks.run_dnsserver_find(self.master, self.replicas[0].hostname, 0)
```
tiran: This test case will never fail because you are ignoring all errors. You have to check the return value of ``run_command`` or don't use ``raiseonerr=False`` in the helper functions.

ipatests/test_integration/test_server_del.py
```diff
@@ -300,3 +300,29 @@ def test_forced_removal_of_master(self):
             self.replicas[0], self.master.hostname,
             ignore_last_of_role=True
         )
+
+
+class TestServerDelReplicaRemoval(IntegrationTest):
+    """
+    Test the checks for Removal of Replica during server-del
+    as part for verification of bz1506188
```
pvoborni: It should mainly refer to the upstream ticket #6572 not the bugzilla.  The bug was that dnsserver records were deleted after removal of a replica.  The test case description doesn't mention it, IMO it should mention it. E.g. something like: "Test that DNS server records and Server records are deleted during server-del." The case can be extended in a future with more checks.

ipatests/pytest_plugins/integration/tasks.py
```diff
@@ -1381,3 +1381,29 @@ def add_dns_zone(master, zone, skip_overlap_check=False,
...
+    else:
+        print cmd.stderr_text
+
+
+def run_server_find(host, server_to_find, returncode=None):
```
pvoborni: This function has the same naming structure as other run_xx_xxx function but behaves differently which might confuse further reader or user. It is called run_server_find. One would expect that it runs it and returns output. But it fact it does a check to returncode and do nothing it even doesn't return the output. 

IMO it should either do some assertion but then should be named something like `def test_server_record_presence(host, server, should_exist). Or should return the output and be called `run_server_find`.

ipatests/pytest_plugins/integration/tasks.py
```diff
@@ -1381,3 +1381,29 @@ def add_dns_zone(master, zone, skip_overlap_check=False,
                                     host.hostname + ".", '--a-rec', host.ip])
     else:
         logger.debug('Zone %s already added.', zone)
+
+
+def run_dnsserver_find(host, server_to_find, returncode=None):
```
pvoborni: Same comment to function name and behavior as for `run_server_del`. Btw return code argument is not self-explanatory. It should e.g. `expected_returncode` or `dnsserver_present`.

pvoborni:
Commit message mentions 2 patches, but the PR has only 1 commit.

tiran:
@ndehadrai Your PR needs to be rebased on latest master.

tiran:
@ndehadrai Please address the review comments and rebase your PR.

rcritten:
Closing due to inactivity.

PR #1379, opened 2017-12-08
Prevent set_directive from clobbering other keys / safe directive setter

tiran:
This PR combines @frasertweedale PR #1347 with a safe directive setter (https://pagure.io/freeipa/issue/7312)

## Original PR message

`set_directive` only looks for a prefix of the line matching the
given directive (key).  If a directive is encountered for which the
given key is prefix, it will be vanquished.

This occurs in the case of `{ca,kra}.sslserver.cert[req]`; the
`cert` directive gets updated after certificate renewal, and the
`certreq` directive gets clobbered.  This can cause failures later
on during KRA installation, and possibly cloning.

Match the whole directive to avoid this issue.

Fixes: https://pagure.io/freeipa/issue/7288

-----

Cause: corner case.

How to test:

1. ensure `ca.sslserver.certreq=<base64 CSR>` exists in `ca/CS.cfg`.
2. resubmit Certmonger tracking request for `Server-Cert cert-pki-ca` Dogtag system cert.
3. verify that `ca.sslserver.certreq=<base64 CSR>` still exists in `ca/CS.cfg`.

## safe DirectiveSetter

The new context manager ``DirectiveSetter`` avoids several possible issues that can lead to destroyed configuration files.

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -452,34 +452,44 @@ def set_directive(filename, directive, value, quotes=True, separator=' '):
     :param separator: character serving as separator between directive and
         value.  Correct value required even when dropping a directive.
     """
-
-    new_directive_value = ""
-    if value is not None:
-        value_to_set = quote_directive_value(value, '"') if quotes else value
-
-        new_directive_value = "".join(
-            [directive, separator, value_to_set, '\n'])
-
-    valueset = False
     st = os.stat(filename)
-    fd = open(filename)
-    newfile = []
-    for line in fd:
-        if re.match(r'\s*{}'.format(re.escape(directive + separator)), line):
-            valueset = True
-            if value is not None:
-                newfile.append(new_directive_value)
-        else:
-            newfile.append(line)
-    fd.close()
-    if not valueset:
-        if value is not None:
-            newfile.append(new_directive_value)
+    with open(filename, 'r') as f:
+        lines = list(f)  # read the whole file
+        new_lines = set_directive_lines(
+            quotes, separator, directive, value, lines)
+    with open(filename, 'w') as f:
+        # don't construct the whole string; write line-wise
+        for line in new_lines:
+            f.write(line)
+    os.chown(filename, st.st_uid, st.st_gid)  # reset perms
 
-    fd = open(filename, "w")
-    fd.write("".join(newfile))
-    fd.close()
-    os.chown(filename, st.st_uid, st.st_gid) # reset perms
+
+def set_directive_lines(quotes, separator, k, v, lines):
+    """Set a name/value pair in a configuration (iterable of lines).
+
+    Replaces the value of the key if found, otherwise adds it at
+    end.  If value is ``None``, remove the key if found.
+
+    Takes an iterable of lines (with trailing newline).
+    Yields lines (with trailing newline).
+
+    """
+    new_line = ""
+    if v is not None:
+        v_quoted = quote_directive_value(v, '"') if quotes else v
+        new_line = ''.join([k, separator, v_quoted, '\n'])
+
+    found = False
+    for line in lines:
+        if re.match(r'\s*{}'.format(re.escape(k + separator)), line):
+            found = True
+            if v is not None:
+                yield new_line
+        else:
+            yield line
+
+    if not found and v is not None:
+        yield new_line
 
 
 def get_directive(filename, directive, separator=' '):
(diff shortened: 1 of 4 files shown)


ipaserver/install/installutils.py
```diff
@@ -436,6 +438,71 @@ def unquote_directive_value(value, quote_char):
...
+        # rename file and sync directory inode
+        os.rename(name, self.filename)
+        dirfd = os.open(directory, os.O_RDONLY | os.O_DIRECTORY)
+        try:
+            os.fsync(dirfd)
+        finally:
+            os.close(dirfd)
+
+    def set(self, directive, value, quotes=SENTINEL, separator=SENTINEL):
+        if quotes is SENTINEL:
```
frasertweedale: What's the advantage of a sentinel `object` over plain old `None`?

ipaserver/install/installutils.py
```diff
@@ -480,8 +549,9 @@ def set_directive_lines(quotes, separator, k, v, lines):
         new_line = ''.join([k, separator, v_quoted, '\n'])
 
     found = False
+    matcher = re.compile(r'\s*{}'.format(re.escape(k + separator)))
```
frasertweedale: :+1:

Obvious, in hindsight :)

ipaserver/install/installutils.py
```diff
@@ -436,6 +438,71 @@ def unquote_directive_value(value, quote_char):
...
+            os.close(dirfd)
+
+    def set(self, directive, value, quotes=SENTINEL, separator=SENTINEL):
+        if quotes is SENTINEL:
+            quotes = self.quotes
+        if separator is SENTINEL:
+            separator = self.separator
+        # materialize lines
+        # set_directive_lines() modify item, shrink or enlage line count
+        self.lines = list(set_directive_lines(
```
frasertweedale: This exposes another potential optimisation.  We can defer the construction of the `list` and just build
generator upon generator.  Force the `list` just before we write the file, and we will traverse the input
only once.

Not specifically recommending this, just pointing out the possibility :)

ipaserver/install/installutils.py
```diff
@@ -436,6 +438,71 @@ def unquote_directive_value(value, quote_char):
     return unescaped_value
 
 
+SENTINEL = object()
+
+
+class DirectiveSetter(object):
+    """Safe directive setter
+
```
frasertweedale:
```
PEP-8 errors:
./ipaserver/install/installutils.py:446:1: W293 blank line contains whitespace
```

ipaserver/install/installutils.py
tiran: You can pass in ``None`` to override the default. With ``None`` as default method argument, there is no way to distinguish between ``set(k, v)`` and ``set(k, v, quotes=None)``.

ipaserver/install/installutils.py
tiran: Another optimization idea: Modify the list in-place and use ``None`` as indicator for line removed.

tiran:
I had another idea to make code even more readable:

```
with DirectiveSetter(filename) as ds:
    ds.setlist([
        ('key1', 'value1'),
        ('key2', 'value2'),
        ...
        ('key99', 'value99'),
    ])
```

Also ``SENTINEL`` should rather be ``_SENTINEL``.

ipaserver/install/installutils.py
frasertweedale: But `None` is not a valid value for either `quotes` or `separator` parameters.
I'd rather just use `None` to indicate "existing or default setting" and instead
add check for `isinstance(quotes, bool)` and `isinstance(separator, str) and len(separator) > 0`.

frasertweedale:
@tiran yeah `setlist` is a nice idea.

flo-renaud:
Hi @tiran 
thank you for the patch, works for me.

tiran:
master:

* 2546ef6eb0c6321da810f42d0311b7053be43d62 Prevent set_directive from clobbering other keys
* 1b04718b3c57eed2131db018b9d4e46c8d7f7345 pep8: reduce line lengths in CAInstance.__enable_crl_publish
* c77f3a50d7a8d4b2f1e6fb7c95115a36a4ec6daa installutils: refactor set_directive
* f688b5d8a7ff340a4f358e99bfe219167832359e Add tests for installutils.set_directive
* f4001e1c53a263aa6c7f62385ed394631345a34c Add safe DirectiveSetter context manager

PR #1402, opened 2017-12-14
WebUI: make keytab tables on service and host pages writable

pvomacka:
There is no object class before adding the first item into tables,
therefore there are no ACI and WebUI is not able to figure out
whether table is writable or not. Adding flag 'w_if_no_aci'
tells "make it writable even if we have not ACIs and try to do
the API call.

https://pagure.io/freeipa/issue/7111


felipevolpone:
Tested and it works properly. However, to the functionality really works, it still depends on the fix of https://pagure.io/freeipa/issue/7324

pvoborni:
this issue, ticket #7111, is for IPA 4.5. Version where #7324 happens is still not confirmed, but seems that it is 4.6, but either way I don't think that we need to wait with pushing this.

AFAIK the fix does not introduce any new problem except that it makes more probable to encounter 7324 because now the value can be set in UI. But 7324 can be reproduced or fixed(value remove) also from CLI.  

So if 7324 is not in 4.5 that with pushing we add value. If it is in 4.5 then pushing won't add any value but also won't remove any.

tiran:
master:

* cf2d171d0c7ec4d3d212eef42007cea392265895 WebUI: make keytab tables on service and host pages writable

PR #1405, opened 2017-12-14
Enable ephemeral KRA requests

tiran:
@rcritten's PR #1116 rebased to latest master. I had to create a new PR because I don't have permission to fix the merge conflict.

## original message
Enabling ephemeral KRA requests will reduce the amount of LDAP
write operations and improve overall performance.

https://pagure.io/freeipa/issue/6703

NOTE: I'm not 100% sure on the upgrade for existing instances. My logic is that tomcat is always stopped and then within that block the CA (if any) will be updated. Given that the KRA runs in the same service that is why I stuck that update code there. It worked in my testing.

### DIFF
--- a/ipaserver/install/ca.py
+++ b/ipaserver/install/ca.py
@@ -336,7 +336,7 @@ def install_step_1(standalone, replica_config, options):
     ca.stop('pki-tomcat')
 
     # This is done within stopped_service context, which restarts CA
-    ca.enable_client_auth_to_db(paths.CA_CS_CFG_PATH)
+    ca.enable_client_auth_to_db()
 
     # Lightweight CA key retrieval is configured in step 1 instead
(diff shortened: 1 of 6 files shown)


tiran:
PR #1116 was originally acked by @frasertweedale 

The only difference between the original PR and my PR is in ``__enable_crl_publish()``, https://github.com/freeipa/freeipa/pull/1405/files#diff-98ceb4d25dfc81588cca4a0fbbe102cbR943

frasertweedale:
Yes, that's the right resolution, thanks @tiran.  ACKed.

tiran:
master:

* 10a847b682786690da3d38777a062dcebb4f1c35 Make the path to CS.cfg a class variable
* a7ae2dbc5ffa22d309d08ddb67b3e1ab24bc4cdc Enable ephemeral KRA requests

PR #1407, opened 2017-12-15
test_help: test "help" command without schema cache

Rezney:
This test case addresses upsteam ticket #6999, where "ipa help"
does not work if called when no schema is cached.

https://pagure.io/freeipa/issue/7325

### DIFF
--- a/ipatests/test_cmdline/test_help.py
+++ b/ipatests/test_cmdline/test_help.py
@@ -18,6 +18,9 @@
 #
 
 import sys
+import os
+import shutil
+import errno
 
 import six
 from six import StringIO
@@ -71,6 +74,27 @@ def test_ipa_help():
         return_value = api.Backend.cli.run(['help'])
     assert return_value == 0
     assert ctx.stderr == ''
+
+
+def test_ipa_help_without_cache():
+    """Test `ipa help` without schema cache"""
+    cache_dir = os.path.expanduser('~/.cache/ipa/schema/')
+    backup_dir = os.path.expanduser('~/.cache/ipa/schema.bak/')
+    shutil.rmtree(backup_dir, ignore_errors=True)
+    if os.path.isdir(cache_dir):
+        os.rename(cache_dir, backup_dir)
+    try:
+        with CLITestContext() as ctx:
+            return_value = api.Backend.cli.run(['help'])
+        assert return_value == 0
+        assert ctx.stderr == ''
+    finally:
+        shutil.rmtree(cache_dir, ignore_errors=True)
+        try:
+            os.rename(backup_dir, cache_dir)
+        except OSError as e:
+            if e.errno != errno.ENOENT:
+                raise
 
 
 def test_ipa_without_arguments():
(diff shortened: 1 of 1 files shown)


ipatests/test_cmdline/test_help.py
```diff
@@ -120,6 +122,16 @@ def test_command_help():
     assert unicode(user_add.doc) in help_ctx.stdout
 
 
+def test_command_help_without_cache():
+    cache_dir = os.path.expanduser('~/.cache/ipa/schema/')
+    if os.path.exists(cache_dir):
```
tiran: You should rather use ``os.path.isdir`` instead of ``os.path.exists``. Or even simpler, ``shutil.rmtree(cache_dir, ignore_errors=True)``

tiran:
I have another idea. FreeIPA supports ``XDG_CACHE_HOME`` env var. You could use the env var to run help with an alternative cache dir:

```
tmpdir = tempfile.mkdtemp()
oldcache = os.environ.get('XDG_CACHE_HOME')
os.environ['XDG_CACHE_HOME'] = tmpdir
try:
    ...
    assert os.path.isdir(os.path.join(tmpdir, 'ipa/schema'))
finally:
    if oldcache is None:
        del os.environ['XDG_CACHE_HOME']
    else:
        os.environ['XDG_CACHE_HOME'] = oldcache
    shutil.rmtree(tmpdir)
```

Rezney:
Ok. The second idea looks even better.

Rezney:
@tiran tried to make a test run, however it is failing on assert os.path.isdir... as it is not creating anything in the tmpdir. Even if try to put non-existent dir to 'XDG_CACHE_HOME' it is not complaining.

tiran:
You are running into https://pagure.io/freeipa/issue/6761 :(

The cache dir variable only works for new processes. You cannot change the settings from within a process.

Rezney:
I see, thanks for the info. Shall we use the actual patch then?

ipatests/test_cmdline/test_help.py
```diff
@@ -73,6 +75,25 @@ def test_ipa_help():
...
 
+def test_ipa_help_without_cache():
+    """Test `ipa help` without schema cache. Skip if no cache dir"""
+    cache_dir = os.path.expanduser('~/.cache/ipa/schema/')
+    backup_dir = os.path.expanduser('~/.cache/ipa/schema.bak/')
+    shutil.rmtree(backup_dir, ignore_errors=True)
+    if os.path.isdir(cache_dir):
+        os.rename(cache_dir, backup_dir)
+    else:
+        pytest.skip('Cache dir {} is not present'.format(cache_dir))
```
tiran: Why do you want to skip the entire test when the cache dir does not exist?

Rezney: Just this one test case,  however yes that is nonsense here. Just will need to avoid moving backup_dir to cache_dir when backup_dir does not exist.

Rezney: not to mention it is quite inconsistent with what I want to achieve here.

Rezney:
Should be ok now...

tiran:
master:

* 23d729e0de6c0bc7447327b434b2458ea85a529d test_help: test "help" command without cache

PR #1408, opened 2017-12-15
Fix pylint warnings inconsistent-return-statements

tiran:
Add consistent return to all functions and methods that are covered by
tox -e pylint[23]. I haven't checked if return None is always a good
idea or if we should rather raise an error.

See: https://pagure.io/freeipa/issue/7326

### DIFF
--- a/ipaclient/plugins/migration.py
+++ b/ipaclient/plugins/migration.py
@@ -77,3 +77,4 @@ can use their Kerberos accounts.''')
                                ldapuri)
             return 1
         textui.print_plain(unicode(self.pwd_migration_msg))
+        return None
--- a/ipaclient/plugins/vault.py
+++ b/ipaclient/plugins/vault.py
@@ -118,6 +118,8 @@ def encrypt(data, symmetric_key=None, public_key=None):
                 label=None
             )
         )
+    else:
+        return None
 
 
 def decrypt(data, symmetric_key=None, private_key=None):
@@ -150,6 +152,8 @@ def decrypt(data, symmetric_key=None, private_key=None):
         except ValueError:
             raise errors.AuthenticationError(
                 message=_('Invalid credentials'))
+    else:
+        return None
 
 
 @register(no_fail=True)
--- a/ipalib/backend.py
+++ b/ipalib/backend.py
@@ -135,20 +135,16 @@ class Executioner(Backend):
         destroy_context()
 
     def execute(self, _name, *args, **options):
-        error = None
         try:
             if _name not in self.Command:
                 raise CommandError(name=_name)
-            result = self.Command[_name](*args, **options)
-        except PublicError as e:
-            error = e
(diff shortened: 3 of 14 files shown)


ipaclient/plugins/migration.py
```diff
@@ -77,3 +77,4 @@ def output_for_cli(self, textui, result, ldapuri, **options):
                                ldapuri)
             return 1
         textui.print_plain(unicode(self.pwd_migration_msg))
+        return None
```
abbra: `cli` class expects that `output_for_cli` returns either error code or something that evaluates into a `False`. In the latter case we get the result silenced (`cli.run()` returns `0`). Probably better to change `return None` to `return 0` for consistency.

ipaclient/plugins/vault.py
```diff
@@ -118,6 +118,8 @@ def encrypt(data, symmetric_key=None, public_key=None):
                 label=None
             )
         )
+    else:
+        return None
```
abbra: `encrypt()` cannot return None because we call `base64.b64encode()` on the result unconditionally.

```
>>> import base64
>>> base64.b64encode(None)
  File "/usr/lib64/python3.6/base64.py", line 58, in b64encode
    encoded = binascii.b2a_base64(s, newline=False)
TypeError: a bytes-like object is required, not 'NoneType'
>>> 
```

I think for both `encrypt()` and `decrypt()` we can return `b''` or similar.

ipaclient/plugins/vault.py
```diff
@@ -150,6 +152,8 @@ def decrypt(data, symmetric_key=None, private_key=None):
         except ValueError:
             raise errors.AuthenticationError(
                 message=_('Invalid credentials'))
+    else:
+        return None
```
abbra: Same here -- probably better to return `b''`

ipaclient/plugins/vault.py
tiran: No, returning b'' is wrong and will hide bugs.

We have returned None in the past. This change just makes it obvious and explicit.

ipaclient/plugins/vault.py
tiran: Same here. The change just makes ``return None`` explicit.

ipaclient/plugins/vault.py
abbra: then I'd suggest to raise an error here. Since both `encode` and `decode` cannot function without symmetric or public/private keys, raising an exception is better than failing in base64.

abbra:
@tiran will make a separate PR for the exception raising for non-intended use of `decode/encode` in `vault.py`, so this PR is OK..

tiran:
master:

* 8cb756a2295d046c22a52fb7a51e7a8c17c7f116 Fix pylint warnings inconsistent-return-statements

PR #1409, opened 2017-12-18
Vault: Add argument checks to encrypt/decrypt

tiran:
Vault's encrypt and decrypt helper function take either symmetric or
public/private key. Raise an exception if either both or none of them
are passed down.

See https://pagure.io/freeipa/issue/7326


abbra:
LGTM. Let's wait for the CI results.

felipevolpone:
Adding ACK based on @abbra's comment.

tiran:
master:

* 1c9f0a4b1979a8a5076be0b434dc41a9f2a93317 Vault: Add argument checks to encrypt/decrypt

PR #1416, opened 2017-12-20
Do not allow users delete their last otp token

felipevolpone:
This adds a new verification on the ipa_otp_lasttoken 389 plugin, in order to do not allow users delete their
last otp token. The verification is done checking if the global configuration is set to otp in `cn=ipaConfig,cn=etc`

Fixes: [7012](https://pagure.io/freeipa/issue/7012)

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-otp-lasttoken/ipa_otp_lasttoken.c
+++ b/daemons/ipa-slapi-plugins/ipa-otp-lasttoken/ipa_otp_lasttoken.c
@@ -105,6 +105,48 @@ static bool sdn_is_only_enabled_token(Slapi_DN *target_sdn, const char *user_dn)
     return result;
 }
 
+static bool is_otp_enabled(const char *user_dn)
+{
+    char *attrs[] = { "ipaUserAuthType", NULL };
+    Slapi_Entry *entry = NULL;
+    Slapi_DN *sdn;
+    const Slapi_DN *base;
+    uint32_t authtypes;
+    int search_result = 0;
+    char *authConfigDN;
+
+    sdn = slapi_sdn_new_dn_byval(user_dn);
+    if (sdn == NULL) {
+        LOG_TRACE("File '%s' line %d: Unable to convert '%s' to Slapi_DN. ",
+                  __FILE__, __LINE__, user_dn);
+        return false;
+    }
+
+    base = slapi_get_suffix_by_dn(sdn);
+
+    authConfigDN = slapi_ch_smprintf("cn=ipaConfig,cn=etc,%s",
+                                     slapi_sdn_get_dn(base));
+    slapi_sdn_free(&sdn);
+    sdn = slapi_sdn_new_dn_byval(authConfigDN);
+
+    search_result = slapi_search_internal_get_entry(sdn, attrs, &entry,
+            otp_config_plugin_id(otp_config));
+    if (search_result != LDAP_SUCCESS) {
+        LOG_TRACE("File '%s' line %d: Unable to access LDAP entry '%s'. "
+                "Perhaps it doesn't exist? Error code: %d\n", __FILE__,
+                __LINE__, slapi_sdn_get_dn(sdn), search_result);
+    }
+
+    slapi_sdn_free(&sdn);
+    if (entry == NULL)
+        return false;
+
+    authtypes = otp_config_auth_types(otp_config, entry);
+    slapi_entry_free(entry);
+
+    return authtypes & OTP_CONFIG_AUTH_TYPE_OTP;
+}
+
 static bool is_pwd_enabled(const char *user_dn)
 {
     char *attrs[] = { "ipaUserAuthType", NULL };
@@ -158,6 +200,9 @@ static bool is_allowed(Slapi_PBlock *pb, Slapi_Entry *entry)
 
     if (!sdn_is_only_enabled_token(target_sdn, bind_dn))
         return true;
+
+    if (is_otp_enabled(bind_dn))
+        return false;
 
     if (is_pwd_enabled(bind_dn))
         return true;
--- a/ipatests/test_integration/test_otp.py
+++ b/ipatests/test_integration/test_otp.py
@@ -1,0 +1,54 @@
+#
+# Copyright (C) 2017  FreeIPA Contributors see COPYING for license
+#
+
+from ipatests.test_integration.base import IntegrationTest
+from ipatests.pytest_plugins.integration import tasks
+
+
+class TestOTPTokenCommand(IntegrationTest):
+    """Test functionality of the ipa otptoken-* commands"""
+
+    topology = 'line'
+
+    def test_delete_last_active_otp_token(self):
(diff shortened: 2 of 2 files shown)


tiran:
I think you have to check the user's settings, too. Shouldn't the user still be allowed to remove the last OTP token when she can also login with 1FA (password only)?

tscherf:
The issue only appears when the global default auth-type is set to 'otp'. When the auth-type is set on the user-account, it was always working as expected and the user was not able to delete the last token. Setting 'otp' and 'password' on the user account doesn't make too much sense IMHO.

abbra:
In https://www.freeipa.org/page/V4/OTP we lack clarity about a combination of global and user settings for last token. @npmccallum, would you please add clarification which cases are considered allowed to delete the last token given that user-defined config overrides the global one?

daemons/ipa-slapi-plugins/ipa-otp-lasttoken/ipa_otp_lasttoken.c
```diff
@@ -105,6 +105,41 @@ static bool sdn_is_only_enabled_token(Slapi_DN *target_sdn, const char *user_dn)
...
+    uint32_t authtypes;
+    int search_result = 0;
+    char *authConfigDN;
+
+    sdn = slapi_sdn_new_dn_byval(user_dn);
+    base = slapi_get_suffix_by_dn(sdn);
+
+    authConfigDN = slapi_ch_smprintf("cn=ipaConfig,cn=etc,%s",
+                                     slapi_sdn_get_dn(base));
+    sdn = slapi_sdn_new_dn_byval(authConfigDN);
```
tiran: You must ``slapi_sdn_free(&sdn)`` before you reassign it again.

daemons/ipa-slapi-plugins/ipa-otp-lasttoken/ipa_otp_lasttoken.c
```diff
@@ -105,6 +105,41 @@ static bool sdn_is_only_enabled_token(Slapi_DN *target_sdn, const char *user_dn)
...
+    char *attrs[] = { "ipaUserAuthType", NULL };
+    Slapi_Entry *entry = NULL;
+    Slapi_DN *sdn;
+    const Slapi_DN *base;
+    uint32_t authtypes;
+    int search_result = 0;
+    char *authConfigDN;
+
+    sdn = slapi_sdn_new_dn_byval(user_dn);
+    base = slapi_get_suffix_by_dn(sdn);
```
tiran: I don't see a ``slapi_sdn_free()`` call for base. Does ``slapi_get_suffix_by_dn`` return a borrowed reference?

daemons/ipa-slapi-plugins/ipa-otp-lasttoken/ipa_otp_lasttoken.c
```diff
@@ -105,6 +105,41 @@ static bool sdn_is_only_enabled_token(Slapi_DN *target_sdn, const char *user_dn)
...
+{
+    char *attrs[] = { "ipaUserAuthType", NULL };
+    Slapi_Entry *entry = NULL;
+    Slapi_DN *sdn;
+    const Slapi_DN *base;
+    uint32_t authtypes;
+    int search_result = 0;
+    char *authConfigDN;
+
+    sdn = slapi_sdn_new_dn_byval(user_dn);
```
tiran: You have to check the result of ``slapi_sdn_new_dn_byval`` for NULL. The value may be NULL in case DS is running out of memory or something else goes wrong

```
sdn = slapi_sdn_new_dn_byval(user_dn);
if (sdn == NULL) {
    LOG_TRACE("File '%s' line %d: Unable to convert '%s' to Slapi_DN. ", __FILE__, __LINE__, user_dn);
    return false;
}
```

daemons/ipa-slapi-plugins/ipa-otp-lasttoken/ipa_otp_lasttoken.c
tiran: And check sdn for NULL, too.

felipevolpone:
@tiran changes addressed, hopefully I got it right. I'll ping tbordaz to check if he had time to review this patch.

daemons/ipa-slapi-plugins/ipa-otp-lasttoken/ipa_otp_lasttoken.c
```diff
@@ -105,6 +105,48 @@ static bool sdn_is_only_enabled_token(Slapi_DN *target_sdn, const char *user_dn)
...
+    if (sdn == NULL) {
+      LOG_TRACE("File '%s' line %d: Unable to convert '%s' to Slapi_DN. ",
```
tiran: Small nit pick, the rest of the file seems to use four space identing.

flo-renaud:
Hi @felipevolpone 
I am not 100% comfortable with the behavior provided by the patch:
- When OTP is enabled globally, but a user-specific setting configures password-only, the removal of last token fails.
- When password only is configured globally, with "Disable per-user override", but a user-specific setting configures OTP, the removal of last token fails.
As @abbra said in his above comment, we should really clarify the expected behavior with global settings / user settings and the flag "Disable per-user override". Also if multiple auth methods are accepted (OTP and password), should we allow to delete the last token?

felipevolpone:
Hello @flo-renaud, I agree with you. The situation here is indeed very confusing.
However, we haven't received any feedback from @npmccallum as @abbra asked. So, I'm not sure how we should proceed.

npmccallum:
The user setting overrides the global setting unless OTP_CONFIG_AUTH_TYPE_DISABLED is set in the global setting. In that case, the user setting is ignored.

https://github.com/felipevolpone/freeipa/blob/ce787314d6eb0ba69509242ebd3d731bd5e62994/daemons/ipa-slapi-plugins/libotp/otp_config.c#L304

In order to solve the bootstrapping problem, if OTP_CONFIG_AUTH_TYPE_OTP is set but no tokens actually exist the user can log in with just a single factor. Thus, we need to prevent the user from removing the last token to intentionally downgrade their security (an admin should individually review all security downgrades).

The last token deletion prevention should work as follows. If the return value from otp_config_auth_types() contains OTP_CONFIG_AUTH_TYPE_PASSWORD, deletion of the last token should be allowed since the user can already log in with 1st factor only. Otherwise, it should be denied. Thus, this patch is incorrect.

I'm not sure what the actual bug is in the existing implementation, because it looks correct to me.

felipevolpone:
@npmccallum actually, the users are able to remove their last token even when the global config is set to otp. here is a simple demo:
output

npmccallum:
My previous comment "should work as follows" was only intended to describe how it **ought** to work. If it doesn't work that way (which seems obvious) then we have a bug. This means that either the `otp_config` object isn't receiving global configuration for some reason. Or there is a bug in the `otp_config` object internally. I think the former is more likely the case.

flo-renaud:
Closing this PR as PR #1582 is proposed instead.

PR #1418, opened 2017-12-20
Support out-of-tree builds

tiran:
Patch collection to enable out-of-tree build of FreeIPA

I had to add a distutils command to copy all Python files from srcdir to builddir and then build the package in builddir. We are auto-generating some files (e.g. ``ipapython/version.py``). There is no other way to include the file in the package while not writing into srcdir.

## how to test

```
$ autoreconf -i -f
$ make distclean (if you have a makefile)
$ mkdir buildroot
$ cd buildroot
$ ../configure
$ make
$ make pypi_packages
```


rcritten:
What is the purpose of this, to not pollute the tree and make things like git ignore easier to deal with?

tiran:
I'm closing the PR. We don't need OOT building at the moment. I'll cherry pick some of the useful fixes to another PR later.

PR #1419, opened 2017-12-20
test_cert_plugin: check if SAN is added with default profile

Rezney:
This test case addresses upstream case #7007.

https://pagure.io/freeipa/issue/7334

### DIFF
--- a/ipatests/test_xmlrpc/test_cert_plugin.py
+++ b/ipatests/test_xmlrpc/test_cert_plugin.py
@@ -224,16 +224,14 @@ class test_cert(BaseCert):
         assert 'valid_not_before' in res
         assert 'valid_not_after' in res
 
-    def test_00010_cleanup(self):
+    def test_00010_san_in_cert(self):
         """
-        Clean up cert test data
+        Test if SAN extension is automatically added with default profile.
         """
-        # Now clean things up
-        api.Command['host_del'](self.host_fqdn)
-
-        # Verify that the service is gone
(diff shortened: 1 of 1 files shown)


rcritten:
LGTM but I'm confused by the commit message. It references upstream case #7007 but I don't know what that is. I *think* it references ticket #7007. I don't think you need to include a pointer to that since it is in the referenced ticket. I'd just remove that reference, it is pretty self-descriptive otherwise.

ipatests/test_xmlrpc/test_cert_plugin.py
```diff
@@ -224,7 +224,16 @@ def test_0009_cert_find(self):
...
+    def test_00010_san_in_cert(self):
+        """
+        Test if SAN extension is automatically added with default profile.
+        """
+        csr = self.generateCSR(str(self.subject))
+        res = api.Command[
+            'cert_request'](csr, principal=self.service_princ)['result']
+        assert 'san_dnsname' in res
+
+    def test_00011_cleanup(self):
```
frasertweedale: It would make sense to bump this to `test_99999_cleanup(self)` so that in the future when new
tests get added, this name won't also need to be changed (resulting in smaller, cleaner patches).

Rezney:
Thanks for the review. Fixed.

tiran:
master:

* 65c592334a9c928eab7d2ce27a8357a5e81956e0 test_cert_plugin: check if SAN is added with default profile

PR #1420, opened 2017-12-21
Fix python-ldap byte warnings

tiran:


tiran:
I'll replace this PR by a better approach that will use a new python-ldap 3.0 API to ignore bytes warnings.

PR #1421, opened 2017-12-25
Match Common Name attribute in Subject

amitkumar50:
ipa cert_find command has an option called --subject. 
The option is documented as --subject=STR Subject. 
It is expected that a --subject option searches by X.509 subject field but it does not do so. 
It searches for CN not cert subject. Hence changing content of --subject help option.

Resolves: https://pagure.io/freeipa/issue/7322

### DIFF
--- a/ipaclient/remote_plugins/2_114/cert.py
+++ b/ipaclient/remote_plugins/2_114/cert.py
@@ -132,7 +132,7 @@ class cert_find(Command):
         parameters.Str(
             'subject',
             required=False,
-            label=_(u'Subject'),
+            label=_(u'Match cn attribute in subject'),
         ),
         parameters.Int(
             'revocation_reason',
--- a/ipaclient/remote_plugins/2_156/cert.py
+++ b/ipaclient/remote_plugins/2_156/cert.py
@@ -132,7 +132,7 @@ class cert_find(Command):
         parameters.Str(
             'subject',
             required=False,
-            label=_(u'Subject'),
+            label=_(u'Match cn attribute in subject'),
         ),
         parameters.Int(
             'revocation_reason',
--- a/ipaclient/remote_plugins/2_164/cert.py
+++ b/ipaclient/remote_plugins/2_164/cert.py
@@ -132,7 +132,7 @@ class cert_find(Command):
         parameters.Str(
             'subject',
             required=False,
-            label=_(u'Subject'),
+            label=_(u'Match cn attribute in subject'),
         ),
         parameters.Int(
             'revocation_reason',
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -1295,7 +1295,7 @@ class cert_find(Search, CertMethod):
     takes_options = (
         Str('subject?',
             label=_('Subject'),
-            doc=_('Subject'),
+            doc=_('Match cn attribute in subject'),
             autofill=False,
         ),
         Int('min_serial_number?',


pvoborni:
Hello, thanks for the patch.

I'm not confident that the patch fixes the issue in right way. It basically creates an API option with a description that it searches for CN. 

So let me start with a question: What is the desired behavior? The bug suggests renaming the option to CN which is basically equivalent of removal of the option and adding a new one. This is not exactly backwords compatible.

This fix is keeping the option but changing a label. It is consistent with behavior, is backwards compatible but creates inconsistency and will prevent from searching by subject in the future (the option will be taken).

Other option is to fix behavior to search the subject. I'm not familiar with dogtag backend but I see in https://github.com/freeipa/freeipa/blob/master/ipaserver/plugins/dogtag.py#L1857 that it explicitly searches in `commonName` field values when `subject` is passed. This might be the bug.  @rcritten, do you think that instead of `commonName` there should be `subjectName` (I see this attr in LDAP) or something else?

IMO when option is called subject and doc says it searches by subject then in needs to search by subject. A new otpion can be added for searching by cn.

rcritten:
I agree that the option is less-than-great but changing the meaning it would be a rather significant behavioral change in the API so should be done with some caution.

From what I can tell there is no search-by-subject option for searching through dogtag.

Changing the label in this way is not the way to fix this because label is used on output in the cli and the output is actually the full subject. A doc option could/should be added instead similar to the non-remote change.

frasertweedale:
IMO we should:

- leave the behaviour as-is;
- update the doc (and possibly the label) to reflect the behaviour
- add a `subject_dn` option for full search of subject DN (adding the option/behaviour to
  Dogtag first, if necessary)

akokshar:
I went through some rfcs and it looks to me like subject field is depricated.
It is still can be used but it shoud contain certain information only. Might it be that that is the reason dogtag is using commonName instead of subject?

frasertweedale:
@akokshar subject field is certainly not deprecated; one particular use of it is (using commonName for server DNS name validation).  The current behaviour and command naming seems to have been implemented with only this particular (now deprecated) use case in mind.

rcritten:
I like the suggestion from Alexander that I added to the ticket.

I have no memory why I linked --subject with only the CN. Maybe it is what dogtag supported, maybe I was lazy or just made a mistake. Who knows.

frasertweedale:
@rcritten I think FreeIPA is just reflecting Dogtag's behaviour/API.  It's not your fault :)

tiran:
``match cn field of certificate`` isn't self-explaining and ``cn`` is rather cryptic. The label should explain that it matches the common name of the cert's subject, perhaps something like ``Match the common name of the certificate's subject.``. I'll let @frasertweedale come up with proper English.

frasertweedale:
I'd say *Match Common Name attribute in Subject*

amitkumar50:
@tiran Done Thanks

tiran:
@amitkumar50 There seems to be a misunderstanding. You haven't updated the **labels** in the Python files. They are still ``label=_(u'match cn field of certificate'),``.

amitkumar50:
Done changes. Apologies for delay.

amitkumar50:
@rcritten Waiting for inputs

rcritten:
Sorry for the delay, especially since I have a nit request. Can you capitalize the m in match so the doc help will be more consistent with other strings? Otherwise this looks fine.

amitkumar50:
@rcritten Done Changes. Thanks.

rcritten:
master:

* 326fd6a70dc9fc9cb4b8bc45f76fe3f6092ee2a2 Match Common Name attribute in Subject

PR #1422, opened 2017-12-26
remove nsslapd-sasl-max-buffer-size override from 10-config.update

fcami:
Since https://pagure.io/389-ds-base/issue/47457 was fixed directly in 389 Directory Server, overriding  nsslapd-sasl-max-buffer-size at install time is not necessary anymore and it breaks overriding its value with --dirsrv-config-file.

The patch addresses:
https://bugzilla.redhat.com/show_bug.cgi?id=1527020
"nsslapd-sasl-max-buffer-size is hardcoded to '2097152' during
install even if another value was provided in an LDIF
( --dirsrv-config-file )"

Tested against RHEL 7.4, the nsslapd-sasl-max-buffer-size parameter
is still 2097152 after this change and the change allows overriding
its value using --dirsrv-config-file properly.

Fix suggested by Florence Blanc-Renaud.

Signed-off-by: François Cami <fcami@fedoraproject.org>

### DIFF
--- a/install/updates/10-config.update
+++ b/install/updates/10-config.update
@@ -58,12 +58,6 @@ addifnew:nsSaslMapPriority: 10
 dn: cn=Name Only,cn=mapping,cn=sasl,cn=config
 addifnew:nsSaslMapPriority: 10
 
-# Default SASL buffer size was too small and could lead for example to
-# migration errors
-# Can be removed when https://fedorahosted.org/389/ticket/47457 is fixed
-dn: cn=config
-only:nsslapd-sasl-max-buffer-size:2097152
-
 # Allow hashed passwords to be added by non-DM users. Without this
 # setting, password migration fails
 dn: cn=config


flo-renaud:
Hi @fcami ,
thank you for contributing! The patch correctly addresses the issue. I tested the following scenarii:
- fresh install without the --dirsrv-config-file: the value of nsslapd-sasl-max-buffer-size is not set in dse.ldif but a LDAP search correctly returns the default value 2097152
- fresh install with --dirsrv-config-file using a value > 2Mb: the value nsslapd-sasl-max-buffer-size is set in dse.ldif and correctly returned by ldapsearch
- fresh install with --dirsrv-config-file using a value < 2Mb: the value is set in dse.ldif but 389-ds overrides the value, ldapsearch returns 2097152

Just for the record, 389-ds forces a value of at least 2097152 = 2Mb (if dse.ldif does not contain any value or if the value is < 2Mb), meaning that upgrade from older versions will not cause any trouble either.

I cloned the BZ 1527020 into a pagure ticket (https://pagure.io/freeipa/issue/7341), that I assigned to you and for which I set the on_review field to this PR. Could you update the commit message and add the following line:

```
Fixes:
https://pagure.io/freeipa/issue/7341
```
This way, your patch will follow our commit template and I will be able to ACK. Thank you!

fcami:
Hi @flo-renaud 

Thanks for the review! I've updated the commit message as requested.

flo-renaud:
Hi @fcami 
thank you, the patch is OK for me. ACK

tiran:
master:

* 6f5042cd873a13cdae5ce41a6329a09dd7191da0 10-config.update: remove nsslapd-sasl-max-buffer-size override as https://pagure.io/389-ds-base/issue/47457 was fixed directly in 389 Directory Server.

PR #1423, opened 2017-12-27
IntegrationTests now collects logs from all test methods

felipevolpone:
It was missing the configuration and the proper use of the logfile_dir. Also, the dict of
logs to be collected should not be cleared. Now, all logs from all test methods will
be collected.

It's useful to say that it doesn't mean that all methods are changing the environment
or installing/reinstalling freeIPA. So, it's possible that some logs in different methods
would be the same.

Fixes: #7310 and #7335


felipevolpone:
The results can be checked here:
* https://fedorapeople.org/groups/freeipa/prci/jobs/334a12c8-eb0a-11e7-b030-001a4a231395/
* https://fedorapeople.org/groups/freeipa/prci/jobs/8be390be-eb03-11e7-b030-001a4a231395/
* https://fedorapeople.org/groups/freeipa/prci/jobs/fea1293e-eb05-11e7-abcd-001a4a23138d/

Rezney:
Tested and it looks like it is working fine.

ipatests/pytest_plugins/integration/__init__.py
```diff
@@ -179,8 +183,6 @@ def collect_logs(name, logs_dict, logfile_dir=None, beakerlib_plugin=None):
             else:
                 shutil.rmtree(topdirname)
 
-        logs_dict.clear()
```
stlaz: Your Jedi mind tricks ("the dict of logs to be collected should not be cleared") don't work on me. What was the original purpose of this clear? Why does not not apply now?

felipevolpone: Let me try to explain it.

To be honest, I do know what was the original purpose of the clear. The commit 29c28786e35ee8e7fced740e62e0b5ddaa9bb381 that introduces the code, does not explain it.

First, is important to know that:
- The `mh` fixture is called only once per class and it is there where we add data to the `logs_dict`. 
- The `integration_logs` is called in every test method.

So, running the code step by step: 
1. in the very first time, the `integration_logs` fixture is called and the `class_integration_log` is empty. It only does `yield class_integration_log` (an empty dict for now).
2. After that, the code goes to the `mh` fixture. [Here](https://github.com/felipevolpone/freeipa/blob/master/ipatests/pytest_plugins/integration/__init__.py#L239-L248) is where the name of the logs that should be collected are added. Some lines below, it calls `yield mh.install()` which start the process of installing ipa server and etc.
3. After the yield above, the flow goes back to the `integration_logs` fixture with the `class_integration_log` having all the logs to be collected. This happens because the `class_integration_log` and the `mh` fixture share the same object via parameter. They're pointing to the same object.
4. In `integration_logs` we call collect_test_logs` which expects to have the `logfile_dir` option to actually collect the logs described in the `logs_dict`.

And that's why we should not clear the dict. It is only built once per class, based on the topology.

I tried to make it more obvious, but this part of the code is quite confusing. If it's not clear, please, let me know.

felipevolpone:
@Rezney @stlaz does my comment/explanation address your doubts? If yes, I'll change the commit message to include it.

stlaz:
@felipevolpone Yes, I hadn't looked into this until last Friday afternoon, but I'll see about it today.

stlaz:
Do I understand it correctly that this PR basically does not try to fix the situation but introduce the `logfile_dir` as a possible option to the YAML config?

felipevolpone:
No. This PR is a fix for the problem and the use the `logfile_dir` is part of the fix.

felipevolpone:
@tiran changes addressed. If you think it's necessary, I can add my explanation to the commit message.

stlaz:
So from the first time I saw this PR, I thought there was some monkey business going on and turns out I was right.
The `logfile_dir` option which you're adding here and the name of which collides with an existing option that is fine by itself is completely unnecessary. The clearing of the dicts suffices. Remove your new option and explain why you're trying to hide arbitrary code behind something you claim is a fix.
Another thing is that with the patch as is, there's one extra directory being created in the collected logs folder. This one takes the name of the class. I don't mind it, although it's confusing. See if you can do something with it.

felipevolpone:
I'm not sure if I follow what you're saying.
I'm not adding any new option. `logfile_dir` is already there, we're just not using it.
Also, `request.config` is not `mh.config` so both are useful. `mh.config` is the `Config` class (that has the `logfile_dir` option) created in ` ipatests/pytest_plugins/integration/config.py` as you can see:

```
238         import pdb
239         pdb.set_trace()
240  ->     request.cls.fallback_config = mh.config
241
242         mh.domain = mh.config.domains[0]
243         [mh.master] = mh.domain.hosts_by_role('master')
244         mh.replicas = mh.domain.hosts_by_role('replica')
245         mh.clients = mh.domain.hosts_by_role('client')
(Pdb) request.config
<_pytest.config.Config object at 0x7fe541f442b0>
(Pdb) type(request.config)
<class '_pytest.config.Config'>
(Pdb) type(mh.config)
<class 'ipatests.pytest_plugins.integration.config.Config'>
(Pdb) request.config.logfile_dir
*** AttributeError: 'Config' object has no attribute 'logfile_dir'
(Pdb) mh.config.logfile_dir
'/root/'
```

felipevolpone:
@stlaz do we have some kind of agreement here?

stlaz:
@felipevolpone No. I can `pdb`, too, and if I add `pdb.set_trace()` right in front of `collect_logs()` in the `collect_test_logs()`, I can easily see that when I run `ipa-run-tests -v -s -r a --with-xunit test_integration/test_caless.py::TestServerInstall::test_http_san --logging-level=DEBUG --logfile-dir=/tmp/tests2` I will get:
```
-> collect_logs(
(Pdb) l
 63  	    :param node: The pytest collection node (request.node)
 64  	    :param logs_dict: Mapping of host to list of log filnames to collect
 65  	    :param test_config: Pytest configuration
 66  	    """
 67  	    import pdb; pdb.set_trace()
 68  ->	    collect_logs(
 69  	        name=_get_logname_from_node(node),
 70  	        logs_dict=logs_dict,
 71  	        logfile_dir=test_config.getoption('logfile_dir'),
 72  	        beakerlib_plugin=test_config.pluginmanager.getplugin('BeakerLibPlugin'),
 73  	    )
(Pdb) p test_config.getoption('logfile_dir')
'/tmp/tests2'
```
Would you kindly finally explain what you need your added `logfile_dir` for?

felipevolpone:
aha, now I see why we're not on the same page. The other config (the `fallback_config`) is that one with options declared in `ipa-tests.yml` file used by `IPATEST_YAML_CONFIG` env var.

stlaz:
Exactly. Please, remove it, it's not necessary.

felipevolpone:
No. The PR makes possible to use the option with the config file, which is a good (even minor) movement. I have no idea why you are against it.

stlaz:
Because it does not make sense.
Because the code is ugly.
Because the option is shadowed by command line option and thus may bring unforeseen puzzling problems when you're looking for the logs but they appear somewhere else.
Because you tried to sneak this change in as a fix but it's not a fix and therefore it lowers your credibility.

You can either remove the extra changes, create an RFE issue to add the option to config, and be done with this, or I will close this PR as rejected.

PR #1425, opened 2018-01-02
Documenting kinit_lifetime in /etc/ipa/default.conf

amitkumar50:
Describing the parameter kinit_lifetime that allows to limit the lifetime of ticket obtained by users authenticating to the WebGUI using login/password.

Resolves: https://pagure.io/freeipa/issue/7333


flo-renaud:
Hi @amitkumar50 ,

thank you for this patch. I did not check through the code but I believe that the parameters session_auth_duration and session_duration_type are not used anymore. Could you check if my assumption is correct, and in this case remove them from the man page in the same patch?

I would also add the following information:
- the expected format is a time duration string (for more info, see https://web.mit.edu/kerberos/krb5-1.12/doc/basic/date_format.html#duration)
- when the parameter is not set in default.conf, the ticket will have a duration inherited from the default value for kerberos clients, that can be set as ticket_lifetime in krb5.conf
- when the ticket lifetime has expired, the ticket is not valid anymore and the GUI will prompt to re-login with a message "Your session has expired. Please re-login."

rcritten:
Makes me wonder if this should spawn a ticket (or tickets) to completely eradicate session_auth_duration and session_duration_type from the source and to migrate existing values use kinit_lifetime on upgrade.

amitkumar50:
@flo-renaud Done changes. Thanks

client/man/default.conf.5
```diff
@@ -107,6 +107,9 @@ This is used in development and is generally a detected value. It means that the
 .B interactive <boolean>
 Specifies whether values should be prompted for or not. The default is True.
 .TP
+.B kinit_lifetime <time format>
+Controls the lifetime of ticket obtained by users authenticating to the WebGUI using login/password. The expected format is a time duration string (for more info, see https://web.mit.edu/kerberos/krb5-1.12/doc/basic/date_format.html#duration). When the parameter is not set in default.conf, the ticket will have a duration inherited from the default value for kerberos clients, that can be set as ticket_lifetime in krb5.conf. When the ticket lifetime has expired, the ticket is not valid anymore and the GUI will prompt to re-login with a message "Your session has expired. Please re-login."
```
flo-renaud: I'd rather avoid using a pointer to an external URL. You can use an example instead, as was done in session_auth_duration.

tiran:
SGTM, ACK after you have addressed Flo's comment.

flo-renaud:
@rcritten I have opened issue #7356 to track the removal of session_auth_duration (proposing upgrade to replace this parameter with kinit_lifetime).

@amitkumar50 Thank you for the new patch, works for me. ACK

flo-renaud:
master:

* 5c361f5450294e9d7b187112cbcb3b08bd037ae5 Documenting kinit_lifetime in /etc/ipa/default.conf

flo-renaud:
@amitkumar50 could you please create a manual backport for ipa-4-5? thanks

PR #1426, opened 2018-01-02
ACI: grant access to admins group instead of admin user

flo-renaud:
The ACI needed for staged users and deleted users were granted
only to the uid=admin user. They should rather be granted to
cn=admins group, to make sure that all members of the admins
group are able to call the command ipa user-del --preserve.

https://pagure.io/freeipa/issue/7342

### DIFF
--- a/install/updates/30-provisioning.update
+++ b/install/updates/30-provisioning.update
@@ -23,12 +23,14 @@ default: cn: deleted users
 # This is used for the admin to know if credential are set for stage users
 # We can do a query on a DN to see if an attribute exists or retrieve the value
 dn: cn=staged users,cn=accounts,cn=provisioning,$SUFFIX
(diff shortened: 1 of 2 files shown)


rcritten:
I'm not 100% sure using remove and add is going to do it in all cases. You'd need to see what failure to do the remove would do in terms of processing other directives. There is a replace directive which while clunky is perhaps more applicable.

flo-renaud:
Hi @rcritten ,
I tested:
- when the aci to remove is not present
- when the aci to remove is present
- when the aci to add is already present
- when the aci to add is not present
and all cases were working.

flo-renaud:
I also added non-regression test.

tiran:
Unrelated to the issue:
Is it a good idea to give admins read permission on passwords at all? AFAIK it allows admins to extract the raw password hash from LDAP. For standard users, the effective rights are ``swo`` (search, write, obliterate).

Tiboris:
master:

* d647072642bdf01d0bfee31a7b5615b145583da5 ACI: grant access to admins group instead of admin user

PR #1427, opened 2018-01-02
Skip zone overlap check with auto-reverse

justin-stephenson:
Skip the existing reverse zone overlap check during DNS installation
when both `--auto-reverse` and `--allow-zone-overlap` arguments are
provided together.

https://pagure.io/freeipa/issue/7239

### DIFF
--- a/ipaserver/install/bindinstance.py
+++ b/ipaserver/install/bindinstance.py
@@ -312,7 +312,7 @@ def read_reverse_zone(default, ip_address, allow_zone_overlap=False):
     return normalize_zone(zone)
 
 
-def get_auto_reverse_zones(ip_addresses):
+def get_auto_reverse_zones(ip_addresses, allow_zone_overlap=False):
     auto_zones = []
     for ip in ip_addresses:
         if ipautil.reverse_record_exists(ip):
@@ -320,12 +320,13 @@ def get_auto_reverse_zones(ip_addresses):
             logger.info("Reverse record for IP address %s already exists", ip)
             continue
         default_reverse = get_reverse_zone_default(ip)
-        try:
-            dnsutil.check_zone_overlap(default_reverse)
-        except ValueError:
-            logger.info("Reverse zone %s for IP address %s already exists",
-                        default_reverse, ip)
-            continue
+        if not allow_zone_overlap:
+            try:
+                dnsutil.check_zone_overlap(default_reverse)
+            except ValueError:
+                logger.info("Reverse zone %s for IP address %s already exists",
+                            default_reverse, ip)
+                continue
         auto_zones.append((ip, default_reverse))
     return auto_zones
 
@@ -503,7 +504,8 @@ def check_reverse_zones(ip_addresses, reverse_zones, options, unattended,
             ips_missing_reverse.append(ip)
 
     # create reverse zone for IP addresses that does not have one
-    for (ip, rz) in get_auto_reverse_zones(ips_missing_reverse):
+    for (ip, rz) in get_auto_reverse_zones(ips_missing_reverse,
+                                           options.allow_zone_overlap):
         if options.auto_reverse:
             logger.info("Reverse zone %s will be created", rz)
             checked_reverse_zones.append(rz)
--- a/ipatests/test_integration/test_caless.py
+++ b/ipatests/test_integration/test_caless.py
@@ -173,7 +173,7 @@ class CALessBase(IntegrationTest):
                        http_pin=_DEFAULT, dirsrv_pin=_DEFAULT, pkinit_pin=None,
                        root_ca_file='root.pem', pkinit_pkcs12_exists=False,
                        pkinit_pkcs12='server-kdc.p12', unattended=True,
-                       stdin_text=None):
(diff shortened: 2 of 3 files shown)


justin-stephenson:
Thank you for the review @tiran 

As this PR affects changes during the DNS setup/installer code, it looks like the integration test is the best method here. I read through existing integration tests but I am not clear what is the best way to setup a mock reverse DNS zone which the IPA server installation will check against during the integration test run.

For clarification I tested this fix in my virt-manager environment by installing IPA with `--setup-dns` on a system pointing to an AD DNS nameserver containing the existing/overlapping reverse zone. Due to replication, I could not simply use an IPA replica and an IPA server to test the creation of this reverse zone.

Any suggestions here is much appreciated.

justin-stephenson:
@tiran I added an integration test to this PR. 

I ran into some errors when running certain integration tests in my environment, therefore i'd like to ask if you or someone is able to review this new test for issues or mistakes.

Thanks in advance.

rcritten:
The underlying issue is that the reverse zone check was failing not because it existed but that in the reporter's environment the DNS request was failing altogether and the check considers any failure to be that the zone exists.

The code fix is good. These are the tests I came up with to force the issue.

```
diff --git a/ipatests/test_integration/test_installation.py b/ipatests/test_i
ntegration/test_installation.py
index 20a08f391..3acb03064 100644
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -9,6 +9,7 @@ installed.
 
 import pytest
 from ipalib.constants import DOMAIN_LEVEL_0
+from ipaplatform.paths import paths
 from ipatests.pytest_plugins.integration.env_config import get_global_config
@@ -222,6 +223,71 @@ class TestInstallWithCA_DNS2(InstallTestBase2):
         super(TestInstallWithCA_DNS2, self).test_replica2_ipa_kra_install()
 
 
+class TestInstallWithCA_DNS3(IntegrationTest):
+    """
+    Test an install with a bad DNS resolver configured to force a
+    timeout trying to verify the existing zones. In the case of a rev
erse
+    zone it is skipped unless --allow-zone-overlap is set regardless 
of
+    the value of --auto-reverse. Confirm that --allow-zone-overlap
+    lets the reverse zone be created.
+
+    ticket 7239
+    """
+
+    @classmethod
+    def install(cls, mh):
+
+        # Force a broken resolv.conf to simulate a bad response to
+        # reverse zone lookups
+        contents = "# Set as broken by ipatests\nnameserver 127.0.0.2
\n"
+        cls.master.put_file_contents(paths.RESOLV_CONF, contents)
+
+        tasks.install_master(cls.master, setup_dns=True,
+                             extra_args=['--allow-zone-overlap'])
+
+    def test_number_of_zones(self):
+        """There should be two zones: one forward, one reverse"""
+
+        result = self.master.run_command([
+            'ipa', 'dnszone-find'])
+
+        assert "in-addr.arpa." in result.stdout_text
+
+        assert "returned 2" in result.stdout_text
+
+
+class TestInstallWithCA_DNS4(IntegrationTest):
+    """
+    Test an install with a bad DNS resolver configured to force a
+    timeout trying to verify the existing zones. In the case of a rev
erse
+    zone it is skipped unless --allow-zone-overlap is set regardless 
of
+    the value of --auto-reverse. Confirm that without --allow-reverse
-zone
+    only the forward zone is created.
+
+    ticket 7239
+    """
+
+    @classmethod
+    def install(cls, mh):
+
+        # Force a broken resolv.conf to simulate a bad response to
+        # reverse zone lookups
+        contents = "# Set as broken by ipatests\nnameserver 127.0.0.2
\n"
+        cls.master.put_file_contents(paths.RESOLV_CONF, contents)
+
+
+    def test_number_of_zones(self):
+        """There should be one zone, a forward because rev timed-out"
""
+
+        result = self.master.run_command([
+            'ipa', 'dnszone-find'])
+
+        assert "in-addr.arpa." not in result.stdout_text
+
+        assert "returned 1" in result.stdout_text
+
+
 @pytest.mark.cs_acceptance
 class TestInstallWithCA_KRA_DNS1(InstallTestBase1):
```

rcritten:
I pushed an updated version of my proposed tests into this branch. Need someone else to take a look.

tiran:
pylint complains about a missing ``from __future__ import absolute_imports``:
```
************* Module ipatests.test_integration.test_installation
ipatests/test_integration/test_installation.py:12: [E0611(no-name-in-module), ] No name 'paths' in module 'ipatests.test_integration.ipaplatform.paths')
ipatests/test_integration/test_installation.py:12: [W0403(relative-import), ] Relative import 'ipaplatform.paths', should be 'ipatests.test_integration.ipaplatform.paths')
```

abbra:
With @rcritten  fixes I have restarted the PRCI. The whole set of patches looks better now.

rcritten:
master:

* 627cb490d2919daa8bd11310df987fced4bf9354 Extend CALessBase::installer_server to accept extra_args
* 00dceb434d0cd7aeddbd3c3eec04d5ac4efea61e Skip zone overlap check with auto-reverse

PR #1428, opened 2018-01-02
Check if replication exist before enable it

felipevolpone:
If the replication does not exist a custom exception is raised explaining the problem.

Fixes: [#7201](https://pagure.io/freeipa/issue/7201)

### DIFF
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -32,6 +32,7 @@ import ldap
 
 from ipalib import api, errors
 from ipalib.cli import textui
+from ipalib.text import _
 from ipapython import ipautil, ipaldap, kerberos
 from ipapython.admintool import ScriptError
 from ipapython.dn import DN
@@ -1547,6 +1548,11 @@ class ReplicationManager(object):
         Disable the replication agreement to hostname.
         """
         entry = self.get_replication_agreement(hostname)
+        if not entry:
+            raise errors.NotFound(reason=_(
+                "Replication agreement for %(hostname)s not found") % {
+                    'hostname': hostname
+                })
         entry['nsds5ReplicaEnabled'] = 'off'
 
         try:
@@ -1561,6 +1567,11 @@ class ReplicationManager(object):
         Note: for replication to work it needs to be enabled both ways.
         """
         entry = self.get_replication_agreement(hostname)
+        if not entry:
+            raise errors.NotFound(reason=_(
+                "Replication agreement for %(hostname)s not found") % {
+                    'hostname': hostname
+                })
         entry['nsds5ReplicaEnabled'] = 'on'
 
         try:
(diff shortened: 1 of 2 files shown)


rcritten:
The same needs to be done for disable_agreement.

Can you change commit msg to "If the replication agreement does not ..." ?

tiran:
AFAIK we are using translator ``_()`` everwhere else to i18n error strings.

Tiboris:
Works as described, should not we catch this exception and log it as warning ?

rcritten:
The replication lib dates back to the early days of the project. I think the PR as fine as it is. It is at least no worse than existing, similar code, and matches current style.

ipa-replica-manage is due a rewrite at some point at which time this library can be made more pythonic as well.

rcritten:
I guess my only issue is that now a NotFound will be raised but it won't be caught, generating a traceback in *replica-manage. The enable/disable_agreement calls should be wrapped in a try/except.

ipaserver/install/replication.py
```diff
@@ -1547,6 +1548,9 @@ def disable_agreement(self, hostname):
         Disable the replication agreement to hostname.
         """
         entry = self.get_replication_agreement(hostname)
+        if not entry:
+            raise errors.NotFound(
+                reason=_("Replication agreement for %s not found" % hostname))
```
MartinBasti: This violates rules of how to write translation strings

https://www.freeipa.org/page/Python_Coding_Style#Python_Internationalized_.28i18n.29_Strings

Also installers haven't been internationalized, only plugins/webUI/CLI. So I'd check if translation works for installers or additional changes may be needed.

ipaserver/install/replication.py
```diff
@@ -1561,6 +1565,9 @@ def enable_agreement(self, hostname):
         Note: for replication to work it needs to be enabled both ways.
         """
         entry = self.get_replication_agreement(hostname)
+        if not entry:
+            raise errors.NotFound(
+                reason=_("Replication agreement for %s not found" % hostname))
```
MartinBasti: Same here.

ipaserver/install/replication.py
MartinBasti: To be exact:
* is  better for translators to always have keyword substitutions (even just for one replacement)
* it should be consistent
* if you need to add additional substitution, you have to rewrite it anyway to keywords
* I think this will not work `_("%s" % test)`. You need `_("%s") % test`

Also please remove this substring `and it has more than one format substitution` from wiki

MartinBasti:
Nice! Thank you for saving translators!

However, I'm not sure if it's me, or the both disable/enable replication agreements methods are copy pasted with only one line changed. It may be worth it to deduplicate them, when you are touching this part of code.

rcritten:
I'm inclined to let the duplication go. If anyone insists I guess he could create a modify_replication_status() or something which does the actually LDAP modify and one passes in enable or disable, on/off or something.

It is probably a bit overkill though in this case given the functions sit right next to each other and are barely used.

MartinBasti:
Both ways are fine by me

tiran:
master:

* ece17cef17f1a84e92846ae288bee581533d753e Check if replication agreement exist before enable/disable it

PR #1429, opened 2018-01-03
Fix warnings and errors found by LGTM static code analyzer

tiran:
https://pagure.io/freeipa/issue/7344

### DIFF
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -223,25 +223,29 @@ def xml_unwrap(value, encoding='UTF-8'):
     :param value: The value to unwrap.
     :param encoding: The Unicode encoding to use (defaults to ``'UTF-8'``).
     """
-    if type(value) in (list, tuple):
+    if isinstance(value, (unicode, int, float, bool)):
+        # most common first
+        return value
+    elif value is None:
+        return value
+    elif isinstance(value, bytes):
+        return value.decode(encoding)
+    elif isinstance(value, (list, tuple)):
         return tuple(xml_unwrap(v, encoding) for v in value)
-    if type(value) is dict:
+    elif isinstance(value, dict):
         if '__dns_name__' in value:
             return DNSName(value['__dns_name__'])
         else:
             return dict(
                 (k, xml_unwrap(v, encoding)) for (k, v) in value.items()
             )
-    if isinstance(value, bytes):
-        return value.decode(encoding)
-    if isinstance(value, Binary):
+    elif isinstance(value, Binary):
         assert type(value.data) is bytes
         return value.data
-    if isinstance(value, DateTime):
+    elif isinstance(value, DateTime):
         # xmlprc DateTime is converted to string of %Y%m%dT%H:%M:%S format
         return datetime.datetime.strptime(str(value), "%Y%m%dT%H:%M:%S")
-    assert type(value) in (unicode, int, float, bool, type(None))
-    return value
+    raise TypeError(value)
 
 
 def xml_dumps(params, version, methodname=None, methodresponse=False,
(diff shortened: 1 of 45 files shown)


ipapython/install/common.py
```diff
@@ -63,15 +63,15 @@ def parent(self):
         raise AttributeError('parent')
 
     def _install(self):
-        for _nothing in self._installer(self.parent):
+        for unused in self._installer(self.parent):
```
frasertweedale: Might be going out on a limb here because the installer code is uberconfusing, but...
at a glance it looks like the only place `_installer` and `_uninstaller` are defined or
referenced is right here in this class, and they both yield a single `None`.

What happens if we delete `_installer` and `_uninstaller` and remove the `for` loop(s)?

ipalib/rpc.py
```diff
@@ -226,7 +226,7 @@ def xml_unwrap(value, encoding='UTF-8'):
     if type(value) in (list, tuple):
         return tuple(xml_unwrap(v, encoding) for v in value)
     if type(value) is dict:
-        if '__dns_name__' in value:
+        if '__dns_name__' in value:  # lgtm [py/member-test-non-container]
```
frasertweedale: This kind of type case must be very common.  I wonder... what happens if you change
the conditional
```python
if type(value) is dict:
```
to
```python
if isinstance(value, collections.Mapping):
```
does *LGTM* recognise the latter more idiomatic/robust construction and stop complaining?

.lgtm.yml
```diff
@@ -0,0 +1,7 @@
+# https://lgtm.com/projects/g/freeipa/freeipa
+
+queries:
+  # False positive unmatchable dollar in regular expression, e.g.
+  # installutils.update_file(certmap_filename, '$ISSUER_DN', str(ca_subject))
+  - exclude: py/regex/unmatchable-dollar
```
frasertweedale: Is there a way we can disable this check in just the one place rather than project-wide?

Also, it is a bug in *LGTM* (doesn't recognise that the datum flows through `re.escape`).
Will you report it to your acquaintant or suggest the best way to report the issue?

ipapython/install/common.py
tiran: Yeah, it's so überconfusing that I don't like to touch it. I *think* the functions are called by ``Installable._configure``.

ipalib/rpc.py
tiran: I'll try that now

.lgtm.yml
tiran: Yes, I figured it out now. I disabled the check globally because I thought it's all over the place. In fact it is just one place.

tiran:
@frasertweedale please review again

ipapython/install/common.py
frasertweedale: Righto, let's leave it alone :)

tiran:
master:

* cae2d99f89d940d2554a82f8ca1bbfeb1f46d336 LGTM: Silence unmatchable dollar
* 1ed4461f338e456837bb074602dd76e0f0bd5b10 LGTM: Use of exit() or quit()
* a4f36eec0ac741f0050ae15928f101932e8aff4a LGTM: Name unused variable in loop
* d3f43a673771aa40365ac1b34dc21e4e22d4a7cb LGTM: Membership test with a non-container
* de616888d5718532dbb8e66ed3f6c698cdf1e79c LGTM: Fix exception in permission_del
* dc599e0797572810a836b84f138f520672f33e94 LGTM: Remove redundant assignment
* 73ee9ff40e7026ce264d0445db96fa16a33ed589 LGTM: Fix multiple use before assignment
* f60b2c590684c8ffc2c6ce2efa7a99f0193f42a1 LGTM: raise handle_not_found()
* 5d02c6aaeb893bf382037d34dc67e6293365c702 LGTM: Use explicit string concatenation
* beb6d74b81eae9965ddc031db1a3826c01d59d30 LGTM: unnecessary else in for loop

PR #1430, opened 2018-01-03
Avoid to use non existent attributes

pilou-:
Closes: https://pagure.io/freeipa/issue/7345

2nd commit: about `path` used by `ServerInfo._read` and `ServerInfo._write`, I am not sure what value could be used here (with this patch both methods do nothing). Should the domain be used there (for example: `~/.cache/ipa/servers/ipa.test`) or network location extracted from `ipalib.request.context.request_url` ?

Related: 3f6411a49c49da7013341ff8feae3a63e75e0fbf.

### DIFF
--- a/ipaclient/remote_plugins/__init__.py
+++ b/ipaclient/remote_plugins/__init__.py
@@ -23,8 +23,6 @@ class ServerInfo(collections.MutableMapping):
     _DIR = os.path.join(USER_CACHE_PATH, 'ipa', 'servers')
 
     def __init__(self, api):
-        hostname = DNSName(api.env.server).ToASCII()
-        self._path = os.path.join(self._DIR, hostname)
         self._force_check = api.env.force_schema_check
         self._dict = {}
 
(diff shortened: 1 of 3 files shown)


rcritten:
IIRC server is mostly deprecated at this point except for some tricky cases when installing a replica so allowing that to be empty may be ok.

xmlrpc_uri is required by certmonger.

pilou-:
On this use case (Ansible dynamic inventory), `ipaclient` and `ipalib` are installed on a host where `certmonger` isn't running.

rcritten:
This is probably a valid use-case.

Have you tested what happens on an IPA client if *uri is missing in default.conf and the ipa tool is used and there are no SRV records defined?

pvoborni:
CI runs were not triggered by this PR. Propable cause is that the PR is not rebased on master. Please rebase the PR to trigger the CI runs.

stlaz:
@pvoborni that's not how our PRCI works. It needs a user to be in a whitelist and we're only whitelisting well-known contributors and team members. Adding the "re-run" label to trigger PRCI.

tiran:
I don't like the PR because it may hide configuration issues. I'd rather update ``ipalib.config.Env._finalize_core`` to derive ``api.env.server`` from either ``jsonrpc_uri`` or ``xmlrpc_uri`` very early. In case ``server`` can't be established outside of ``context == 'installer``, we should fail early.

rcritten:
Closing due to inactivity.

PR #1431, opened 2018-01-03
Fixing test_testconfig with proper asserts

felipevolpone:
When the `cls` in env_config.py is a WinHost, the `__init__` receives different
parameters. Now, it's adapted to all different kinds of hosts.

Also, it's necessary to add the host_type field to most of the domains created in the
test classes, because the field is returned by `pytest_multihost.Config`
in `pytest_plugins/integration/config.py::Config::to_dict`

Fixes: [#7346](https://pagure.io/freeipa/issue/7346)

The test result can be checked here:
https://fedorapeople.org/groups/freeipa/prci/jobs/75ff284a-f0ac-11e7-9671-001a4a23169f/


tiran:
master:

* f5c01c5e86336dd9746b56f511e64eaecf1ed656 Fixing test_testconfig with proper asserts

PR #1433, opened 2018-01-04
ipa-run-tests: replace chdir with plugin

tiran:
The ipa-run-tests command used os.chdir() to change into the ipatests/
directory. The approach works for simple cases but breaks some pytest
features. For example it makes it impossible to selects tests by their
fully qualified test name.

Further more, coverage statistics break because path and module names
get messed up by chdir.

A name plugin takes care of adjusting paths relative to ipatests and to
add ipatests as base. It's now possible to run tests with qualified test
names, e.g.

```
PYTHONPATH=$(pwd) ipatests/ipa-run-tests \
    ipatests/test_ipalib/test_base.py::test_ReadOnly::test_lock
```

### DIFF
--- a/ipatests/ipa-run-tests
+++ b/ipatests/ipa-run-tests
@@ -40,10 +40,41 @@ os.environ['IPATEST_XUNIT_PATH'] = os.path.join(os.getcwd(), 'nosetests.xml')
 
 HERE = os.path.dirname(os.path.abspath(ipatests.__file__))
 
-args = sys.argv[1:]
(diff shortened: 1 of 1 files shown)


slaykovsky:
Hello @tiran !

Thanks for a patch, I've just run one test and nothing did explode :)

tiran:
master:

* 93c158b05812b55383043e320cac89550993b8b2 ipa-run-tests: replace chdir with plugin

PR #1437, opened 2018-01-05
Idviews: fix objectclass violation on idview-add

flo-renaud:
When the option --domain-resolution-order is used with the command
ipa idview-add, the resulting LDAP object stores the value in
ipadomainresolutionorder attribute.
The issue is that the add command does not add the needed object
class (ipaNameResolutionData) because it is part of
possible_objectclasses but not of object_class.

The fix makes sure to add the objectclass when the option
--domain-resolution-order is used, and adds a non-regression test.

Note that idview-mod does not have any issue as it correctly handles
the addition of missing possible objectclasses.

Fixes:
https://pagure.io/freeipa/issue/7350

### DIFF
--- a/ipaserver/plugins/idviews.py
+++ b/ipaserver/plugins/idviews.py
@@ -22,10 +22,11 @@ import re
 import six
 
 from .baseldap import (LDAPQuery, LDAPObject, LDAPCreate,
-                                     LDAPDelete, LDAPUpdate, LDAPSearch,
-                                     LDAPAddAttributeViaOption,
-                                     LDAPRemoveAttributeViaOption,
-                                     LDAPRetrieve, global_output_params)
+                       LDAPDelete, LDAPUpdate, LDAPSearch,
+                       LDAPAddAttributeViaOption,
+                       LDAPRemoveAttributeViaOption,
+                       LDAPRetrieve, global_output_params,
+                       add_missing_object_class)
 from .hostgroup import get_complete_hostgroup_member_list
 from ipalib import (
     api, Str, Int, Flag, _, ngettext, errors, output
@@ -173,6 +174,12 @@ class idview_add(LDAPCreate):
     def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys, **options):
         self.api.Object.config.validate_domain_resolution_order(entry_attrs)
 
+        # The objectclass ipaNameResolutionData may not be present on
+        # the id view. We need to add it if we define a new
+        # value for ipaDomainResolutionOrder
+        if 'ipadomainresolutionorder' in entry_attrs:
+            add_missing_object_class(ldap, u'ipanameresolutiondata', dn,
+                                     entry_attrs, update=False)
         return dn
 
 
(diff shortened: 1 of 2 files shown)


ipaserver/plugins/idviews.py
```diff
@@ -173,6 +173,12 @@ class idview_add(LDAPCreate):
     def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys, **options):
         self.api.Object.config.validate_domain_resolution_order(entry_attrs)
 
+        # The objectclass ipaNameResolutionData may not be present on
+        # the id view. We need to add it if we define a new
+        # value for ipaDomainResolutionOrder
+        if 'ipadomainresolutionorder' in entry_attrs:
+            entry_attrs['objectclass'].append('ipanameresolutiondata')
```
tiran: What happens if  the entry already has the ipanameresolutiondata object class?
```
if ('ipadomainresolutionorder' in entry_attrs and not
        any(oc.lower() == 'ipanameresolutiondata' for oc in entry_attrs['objectclass'])):
    entry_attrs['objectclass'].append('ipanameresolutiondata')
```

flo-renaud:
Hi @tiran 
thank you for the review.
Re. your first comment, as the issue happens in the idview_add precallback, and it inherits from LDAPCreate, the entry contains only the object classes specified in object_class, i.e ['ipaIDView', 'top']. But the test cannot hurt, so I'm adding your suggestion.
Re. the second comment, good catch! I tested only with ipa-run-tests-3 and did not see the issue. Fixed now.

abbra:
@flo-renaud, we have a method to handle these cases: `ipaserver.plugins.baseldap.add_missing_object_class()` can be used to achieve what you want:

```
$ git grep add_missing_object_class
ipaserver/plugins/baseldap.py:def add_missing_object_class(ldap, objectclass, dn, entry_attrs=None, update=True):
ipaserver/plugins/baseuser.py:    add_missing_object_class)
ipaserver/plugins/baseuser.py:        add_missing_object_class(ldap, u'ipacertmapobject', dn)
ipaserver/plugins/host.py:                                     pkey_to_value, add_missing_object_class,
ipaserver/plugins/host.py:        add_missing_object_class(ldap, u'ipaallowedoperations', dn)
ipaserver/plugins/host.py:        add_missing_object_class(ldap, u'ipaallowedoperations', dn)
ipaserver/plugins/service.py:    add_missing_object_class,
ipaserver/plugins/service.py:        add_missing_object_class(ldap, u'ipaallowedoperations', dn)
ipaserver/plugins/service.py:        add_missing_object_class(ldap, u'ipaallowedoperations', dn)
```

flo-renaud:
@abbra 
thanks for the tip. I don't know why, I thought that the add_missing_objectclass method could only be used on existing LDAP entries, but the parameter update=False allows to use it on entries not created yet. PR updated with your suggestion.

abbra:
Thank you, @flo-renaud. The change looks good now. Let's wait for the tests to complete.

abbra:
ACK from my side. @tiran, are you satisfied too?

tiran:
Go for it!

tiran:
master:

* 830866d68a583c5d4ee679c63497a58b3fd31e25 Idviews: fix objectclass violation on idview-add

PR #1438, opened 2018-01-05
Make fastlint even faster

tiran:
- Check pycodestyle before pylint. pycodestyle takes seconds while
  pylint can easily take half a minute or more.
- Fix exit, needs two $
- Add some newlines to make output more readable


Tiboris:
tabs_vs_spaces
**But still patch works works for me :)**

tiran:
No, this is Makefile. It's correct to mix tabs and spaces here. The first tab is required by make, the spaces are for the part that is handled by shell.

tiran:
master:

* 6fe3228fb2f62d30604d9bc5507ea2e6b1642d82 Make fastlint even faster

PR #1439, opened 2018-01-05
Sort external schema files

tiran:
get_all_external_schema_files() now returns schema files sorted.

Fixes: https://pagure.io/freeipa/issue/7338


abbra:
For the record: we discussed merits of sorted(f) versus f.sort() on IRC and decided to go with the sorted(f) to better express what is intended here.
LGTM and ACK.

tiran:
master:

* dc4109c1a459919474dfb9c01996571d7fd5eae4 Sort external schema files

PR #1441, opened 2018-01-08
Removing ipaCert Private key & certificate after ipa upgrade

amitkumar50:
`# yum install ipa-server -y  {IPA 4.4}`
 `# ipa-server-install`
 `#yum update ipa-* -y`
 `# ls -ltr /etc/httpd/alias`

 Presently "-D" option only removes Cert.
 **`certutil ... -D -n ipaCert`**

 Using '-F' option to remove Private key and cert both
 **`certutil ... -F -n ipaCert`**

Resolves: https://pagure.io/freeipa/issue/7329

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -791,7 +791,7 @@ class NSSDatabase(object):
         self.run_certutil(args, stdin=cert.public_bytes(x509.Encoding.PEM))
 
     def delete_cert(self, nick):
-        self.run_certutil(["-D", "-n", nick])
+        self.run_certutil(["-F", "-n", nick])
 
     def verify_server_cert_validity(self, nickname, hostname):
         """Verify a certificate is valid for a SSL server with given hostname


stlaz:
LGTM. I believe we never claimed we're keeping old private keys in the databases, so it should be OK to change the method which is called in other cases, too. Still needs to be tested just for sure.

ipapython/certdb.py
```diff
@@ -791,7 +791,7 @@ def add_cert(self, cert, nick, flags):
         self.run_certutil(args, stdin=cert.public_bytes(x509.Encoding.PEM))
 
     def delete_cert(self, nick):
-        self.run_certutil(["-D", "-n", nick])
+        self.run_certutil(["-F", "-n", nick])
```
tiran: ``-F`` only deletes the private key. We want to delete both.

from man certutil
> When you delete keys, be sure to also remove any certificates associated with those keys from the certificate database, by using -D. Some smart cards do not let you remove a public key you have generated. In such a case, only the private key is deleted from the key pair. You can display the public key with the command certutil -K -h tokenname.

amitkumar50: @tiran Very Thanks for Review.

But I believe '-F' option deletes both keys & certificates. 

This is the test I did on local setup.
//Generate CA Key
    `# openssl genrsa -aes256 -out private/rootCA.key 4096;    chmod 400 private/rootCA.key`     

//Generate CA CERT
    `# openssl req -config openssl.cnf -key private/rootCA.key  -new -x509 -days 7300 -sha256 -extensions v3_ca -out certs/rootCA.crt`    

//Looking inside Certificate DB
`# mkdir nssdb-test;    cd nssdb-test;    certutil -N -d .`
`# ls -ltr`
total 60
-rw-------. 1 root root 16384 Jan  8 23:30 secmod.db
-rw-------. 1 root root 16384 Jan  8 23:30 key3.db
-rw-------. 1 root root 65536 Jan  8 23:30 cert8.db

//Create 1 File 'server.pfx' containing both Key and certificate
`# openssl pkcs12 -export     -out server.pfx     -inkey ../private/rootCA.key     -in ../certs/rootCA.crt`       
Enter pass phrase for ../private/rootCA.key:
Enter Export Password:
Verifying - Enter Export Password:
`#`

//Add pkcs12/*.pfx file to certificate DB
`# pk12util -i server.pfx     -d .`            
Enter password for PKCS12 file: 
pk12util: no nickname for cert in PKCS12 file.
pk12util: using nickname: Default Company Ltd
pk12util: PKCS12 IMPORT SUCCESSFUL

//Verifying Cert, Key is added to DB
`# certutil -L -d .`
Certificate Nickname        Trust Attributes
                                         SSL,S/MIME,JAR/XPI
Default Company Ltd           u,u,u                <<<<<<<<<Cert Added to DB
`# certutil  -K  -d.`          <<<<<<Key added to DB
certutil: Checking token "NSS Certificate DB" in slot "NSS User Private Key and Certificate Services"
< 0> rsa      9c962813d1b7619f1376e2fa5ea14203bf6fb4ca   Default Company Ltd

//Trying to use -D -F
`# certutil -F -D -n 'Default Company Ltd'     -d .`      
certutil: only one command at a time!              <<<<<<<<<<<<<<<<
You entered:  -D -F 
`#`

//Deleting Key, cert with -F
`# certutil -F -n 'Default Company Ltd'     -d .`

//Verifying Cert, Key are deleted from Certificate DB
`# certutil -L -d .`
Certificate Nickname                                         Trust Attributes
                                                             SSL,S/MIME,JAR/XPI
`# certutil -K -d .`
certutil: Checking token "NSS Certificate DB" in slot "NSS User Private Key and Certificate Services"
certutil: no keys found
`#`

As we can see -F option deletes key+associated certificate from Certificate DB.

Kindly do correct me.. If my interpretation is wrong...

tiran:
According to certutil's own documentation, ``-F`` does not delete the public cert. It's either a bug in the documentation of certutil or certutil does not guarantee to delete the cert with the private key.

amitkumar50:
@tiran very Thanks for review.
What to be done next.. Would you try above commands at your end.. 
Or to open bugzilla with nss-tools?

tiran:
Let's first start by providing a unit test for the change:

* unit test should be added to https://github.com/freeipa/freeipa/blob/master/ipatests/test_ipapython/test_certdb.py
* modify ``create_selfsigned()`` to take a certnick argument and use it as nickname and for the subject CN.
* add a new test that adds two self-signed certs, verifies that two certs and private keys are in NSSDB, removes one cert/key pair, verifies that only the other cert and private remain in NSSDB

kaie:
certutil -F calls PK11_DeleteTokenCertAndKey, which means it deletes both. It seems that currently there is no command to delete a key only. I agree the documentation of the -F parameter should be clarified to say that it will delete both key and certificate identified by the given nickname.

certutil -D calls SEC_DeletePermCertificate, which seams to only delete a certifiate.

kaie:
https://bugzilla.mozilla.org/show_bug.cgi?id=1429393

amitkumar50:
Anything pending from my end?

stlaz:
@amitkumar50 you were asked to add a unit test by @tiran in https://github.com/freeipa/freeipa/pull/1441#issuecomment-356565816

tiran:
You need to test three cases

1. both cert and key are present
2. only cert is present
3. only key is present

In any case, key and cert should be removed.

amitkumar50:
certutil -F -n test-cert -d /etc/pki/nssdb/
It is not deleting certificate only if present in Certificate Database.
But this command deletes file containing cert+Key if present in cert db.

rcritten:
So what are the next steps. Try -F first, then -D to ensure the cert is sure to go away?

Tests are still needed.

tiran:
The PR hasn't been updated in almost two months. Are you still working on the issue?

Tiboris:
Hello @amitkumar50 could you please have a look?

amitkumar50:
Apologies @Tiboris @tiran Will update with in a week, I was over occupied.

rcritten:
What do we want to do with this? We aren't actively working on branches < ipa-4-6 and the agent cert is not stored in the Apache NSS database in subsequent branches.

tiran:
I still like to see this issue addressed. A couple of places use ``delete_cert``. It's an easy fix, too.

```
try:
    # -F deletes cert and key if both are present
    self.run_certutil(["-F", "-n", nick])
    logger.debug("Deleting with -K failed, try cert only. (%s)", e)
    self.run_certutil(["-D", "-n", nick])
```

First test case creates cert + key, deletes it, and checks with ``list_keys()`` and ``list_certs``. Second test case imports a cert only, deletes it and checks again.

flo-renaud:
This PR did not move since April, hence closing and proposing a fix with tests in PR #3380.

PR #1446, opened 2018-01-09
Fixing test_backup_and_restore assert to do not rely on the order of returned data

felipevolpone:
Since we cannot assume that LDAP will return data in an ordered way,
the test should be changed to do not rely on that.

Instead of just comparing the output of the show-user command, this change
first order the groups returned in the 'Member of Group' field before
compare them.

https://pagure.io/freeipa/issue/7339

The result (green tests) can be checked here:
https://fedorapeople.org/groups/freeipa/prci/jobs/c43c46d0-f4aa-11e7-925b-001a4a2316ab/

This PR depends on the PR #1354

### DIFF
--- a/ipatests/test_integration/test_backup_and_restore.py
+++ b/ipatests/test_integration/test_backup_and_restore.py
@@ -65,6 +65,15 @@ def check_admin_in_ldap(host):
 def check_admin_in_cli(host):
     result = host.run_command(['ipa', 'user-show', 'admin'])
     assert 'User login: admin' in result.stdout_text, result.stdout_text
+
+    # LDAP do not guarantee any order, so the test cannot assume it. Based on
+    # that, the code bellow order the 'Member of groups' field to able to
+    # assert it latter.
+    data = dict(re.findall("\W*(.+):\W*(.+)\W*", result.stdout_text))
+    data["Member of groups"] = ', '.join(sorted(data["Member of groups"]
+                                                .split(", ")))
+    result.stdout_text = ''.join([' {}: {}\n'.format(k, v)
+                                  for k, v in data.items()])
     return result
 
 


tiran:
The magic constant may cause the test to fail if the order of fields is changed in the future. You could use https://docs.python.org/3/library/re.html#re.sub instead of magic constants to sort the fields:

```
def sortfield(mo):
    groups = mo.group(2).split(', ')
    return mo.group(1) + ', '.join(sorted(groups))

re.sub('(Member of groups:\s*)(.*)', sortfield, output)
```

felipevolpone:
I thought to go in that way, but then we would hardcode the "Member of groups" label. I'm not sure what is the best option

tiran:
IMO ``Member of groups`` is easier to understand and more searchable than ``9``. You could try to extract the string from ``user`` plugin. But I'm not sure if it's easily possible,

abbra:
It is actually in the `ipaserver.plugins.baseldap.global_output_params`, `Str('memberof_group?')`.

```
>>> from ipaserver.plugins.baseldap import global_output_params
>>> [i.label.msg for i in global_output_params if 'memberof_group' == i.name]
['Member of groups']
>>> 

```

tiran:
I think it's fine to hard-code ``Member of groups:`` here. You'd have to hard-code a constant for @abbra's approach, too.

slaykovsky:
Hello!

You can also check it this way:
```
def check_admin_in_cli(host):
    result = host.run_command(['ipa', 'user-show', 'admin'])
    assert 'User login: admin' in result.stdout_text, result.stdout_text
    # output = result.stdout_text.split('\n')

    # LDAP do not guarantee any order, so the test cannot assume it. Based on
    # that, the code bellow order the 'Member of groups' field to able to
    # assert it latter.
    data = dict(re.findall("\W*(.+):\W*(.+)\W*", result.stdout_text))
    data["Member of groups"] = sorted(data["Member of groups"].split(", "))
    result.stdout_text = str(data)

    return result
```

felipevolpone:
Thank you all for the suggestions :) I've updated the PR.

The downside of the patch as it is now is the fact that it changes the order of the resulted string, but since no other test depends on that, we're fine.

ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -65,6 +65,15 @@ def check_admin_in_ldap(host):
...
     result = host.run_command(['ipa', 'user-show', 'admin'])
     assert 'User login: admin' in result.stdout_text, result.stdout_text
+
+    # LDAP do not guarantee any order, so the test cannot assume it. Based on
+    # that, the code bellow order the 'Member of groups' field to able to
+    # assert it latter.
+    data = dict(re.findall("\W*(.+):\W*(.+)\W*", result.stdout_text))
+    data["Member of groups"] = ', '.join(sorted(data["Member of groups"]
+                                                .split(", ")))
+    result.stdout_text = ''.join([' {}: {}\n'.format(k,v)
```
slaykovsky: Whitespace is missing

ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -65,6 +65,15 @@ def check_admin_in_ldap(host):
...
+                                  for k,v in data.items()])
```
slaykovsky: Here also whitespace :)

Tiboris:
master:

* cd660d192231674d8e7d7bd85a8eec68eea29bc7 Fixing test_backup_and_restore assert to do not rely on the order

PR #1448, opened 2018-01-09
test_caless: add SAN extension to server certs

Rezney:
Currently when testing we are using SAN extension only in
KDC, wildcard certs and not in the valid http/ds certs.
During replica installation we then see a warning about certs
having no `subjectAltName`.
  
`[ipatests.pytest_plugins.integration.host.Host.replica.cmd1047]   [7/41]: adding default schema
[ipatests.pytest_plugins.integration.host.Host.replica.cmd1047] /usr/lib/python3.6/site-packages/requests/packages/urllib3/connection.py:340: SubjectAltNameWarning: Certificate for replica.ipa.test has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818. (See https://github.com/shazow/urllib3/issues/497 for details.)`

### DIFF
--- a/ipatests/pytest_plugins/integration/create_caless_pki.py
+++ b/ipatests/pytest_plugins/integration/create_caless_pki.py
@@ -384,14 +384,14 @@ def gen_server_certs(nick_base, hostname, org, ca=None):
                 x509.NameAttribute(NameOID.ORGANIZATION_NAME, org),
                 x509.NameAttribute(NameOID.COMMON_NAME, hostname)
              ]),
-             ca
+             ca, dns_name=hostname
              )
     gen_cert(profile_server, nick_base + u'-badname',
              x509.Name([
                 x509.NameAttribute(NameOID.ORGANIZATION_NAME, org),
                 x509.NameAttribute(NameOID.COMMON_NAME, u'not-' + hostname)
              ]),
-             ca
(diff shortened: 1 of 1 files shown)


tiran:
Could you please add SAN to the other certs, too? ``dns_name=u'not-' + hostname``, ``dns_name=u'alt-' + hostname``, etc. It may be a good idea to make dns_name a required keyword, too.

Rezney:
Thanks for checking this PR. I added SANs also to the other certs and run the changes on the suite. It passes. I would leave the "dns_name" option as it is as we may want to test also without SAN.

Rezney:
Spotted some failures when running whole ca-less test suite on this PR. Will need to check it. Removed ACK label for the moment.

tiran:
It was a CI hickup, nothing serious.

Rezney:
In this case "test_http/ds_san" tests were failing with:

`ScriptError: The server certificate in http.p12 is not valid: invalid for server master.ipa.test`

Because I changed "dns_name" from hostname to "alt-" + hostname and in this case there was nothing that would match the real hostname. Should be ok now.

tiran:
master:

* e55969f7e03a864fe59b21545577a2b12917ecbd test_caless: add SAN extension to other certs

PR #1449, opened 2018-01-09
Switch from mod_nss to mod_ssl

rcritten:
New installs using an IPA CA should work
Upgrades should work

Not tested and some known to not work:
- [x] CA-less install
- [x] promoting a replica
- [x] promoting a replica CA-less - DL1
- [x] promoting a replica CA-less - DL0
- [x] backup and restore (particularly edge cases like restoring from a mod_nss backup)
- [x] domain level 0 replica install with old master
- [x] domain level 0 replica install with new master

This PR is meant to to obtain status of current patches and as a jumping-off point to finish the rest of the transition.

### DIFF
--- a/ipalib/install/certmonger.py
+++ b/ipalib/install/certmonger.py
@@ -670,11 +670,3 @@ def wait_for_request(request_id, timeout=120):
         raise RuntimeError("request timed out")
 
     return state
-
-if __name__ == '__main__':
-    request_id = request_cert(paths.HTTPD_ALIAS_DIR,
-                              "cn=tiger.example.com,O=IPA",
-                              "HTTP/tiger.example.com@EXAMPLE.COM", "Test")
-    csr = get_request_value(request_id, 'csr')
-    print(csr)
-    stop_tracking(request_id)
--- a/ipalib/x509.py
+++ b/ipalib/x509.py
@@ -421,7 +421,7 @@ def load_unknown_x509_certificate(data):
         return load_der_x509_certificate(data)
 
 
-def load_certificate_from_file(filename, dbdir=None):
+def load_certificate_from_file(filename):
     """
     Load a certificate from a PEM file.
 
--- a/ipalib/x509.py
+++ b/ipalib/x509.py
@@ -31,6 +31,7 @@
 
 from __future__ import print_function
 
+import os
 import binascii
 import datetime
 import ipaddress
@@ -41,8 +42,9 @@ import re
 from cryptography import x509 as crypto_x509
 from cryptography import utils as crypto_utils
 from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.serialization import (
-    Encoding, PublicFormat
+    Encoding, PublicFormat, PrivateFormat, load_pem_private_key
 )
 from pyasn1.type import univ, char, namedtype, tag
 from pyasn1.codec.der import decoder, encoder
@@ -59,8 +61,12 @@ if six.PY3:
 PEM = 0
 DER = 1
 
-PEM_REGEX = re.compile(
+PEM_CERT_REGEX = re.compile(
     b'-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----',
+    re.DOTALL)
+PEM_PRIV_REGEX = re.compile(
+    b'-----BEGIN(?: ENCRYPTED)?(?: (?:RSA|DSA|DH|EC))? PRIVATE KEY-----.*?'
+    b'-----END(?: ENCRYPTED)?(?: (?:RSA|DSA|DH|EC))? PRIVATE KEY-----',
     re.DOTALL)
 
 EKU_SERVER_AUTH = '1.3.6.1.5.5.7.3.1'
@@ -431,7 +437,7 @@ def load_certificate_list(data):
 
     Return a list of python-cryptography ``Certificate`` objects.
     """
-    certs = PEM_REGEX.findall(data)
+    certs = PEM_CERT_REGEX.findall(data)
     return [load_pem_x509_certificate(cert) for cert in certs]
 
 
@@ -444,6 +450,35 @@ def load_certificate_list_from_file(filename):
     """
     with open(filename, 'rb') as f:
         return load_certificate_list(f.read())
+
+
+def load_private_key_list(data, password=None):
+    """
+    Load a private key list from a sequence of concatenated PEMs.
+
+    :param data: bytes containing the private keys
+    :param password: bytes, the password to encrypted keys in the bundle
+
+    :returns: List of python-cryptography ``PrivateKey`` objects
+    """
+    crypto_backend = default_backend()
+    priv_keys = []
+
+    for match in re.finditer(PEM_PRIV_REGEX, data):
+        if re.search(b"ENCRYPTED", match.group()) is not None:
+            if password is None:
+                raise RuntimeError("Password is required for the encrypted "
+                                   "keys in the bundle.")
+            # Load private key as encrypted
+            priv_keys.append(
+                load_pem_private_key(match.group(), password,
+                                     backend=crypto_backend))
+        else:
+            priv_keys.append(
+                load_pem_private_key(match.group(), None,
+                                     backend=crypto_backend))
+
+    return priv_keys
 
 
 def pkcs7_to_certs(data, datatype=PEM):
@@ -531,6 +566,24 @@ def write_certificate_list(certs, filename):
         with open(filename, 'wb') as f:
             for cert in certs:
                 f.write(cert.public_bytes(Encoding.PEM))
+    except (IOError, OSError) as e:
+        raise errors.FileError(reason=str(e))
+
+
+def write_pem_private_key(priv_key, filename):
+    """
+    Write a private key to a file in PEM format. Will force 0x600 permissions
+    on file.
+
+    :param priv_key: cryptography ``PrivateKey`` object
+    """
+    try:
+        with open(filename, 'wb') as fp:
+            os.fchmod(fp.fileno(), 0o600)
+            fp.write(priv_key.private_bytes(
+                Encoding.PEM,
+                PrivateFormat.TraditionalOpenSSL,
+                serialization.NoEncryption()))
     except (IOError, OSError) as e:
         raise errors.FileError(reason=str(e))
 
--- a/ipalib/x509.py
+++ b/ipalib/x509.py
@@ -543,8 +543,7 @@ def write_certificate(cert, filename):
     """
     Write the certificate to a file in PEM format.
 
-    The cert value can be either DER or PEM-encoded, it will be normalized
-    to DER regardless, then back out to PEM.
+    :param cert: cryptograpy ``Certificate`` object
     """
 
     try:
--- a/ipaplatform/base/paths.py
+++ b/ipaplatform/base/paths.py
@@ -51,6 +51,8 @@ class BasePathNamespace(object):
     HTTPD_IPA_CONF = "/etc/httpd/conf.d/ipa.conf"
     HTTPD_NSS_CONF = "/etc/httpd/conf.d/nss.conf"
     HTTPD_SSL_CONF = "/etc/httpd/conf.d/ssl.conf"
+    HTTPD_CERT_FILE = "/etc/pki/tls/certs/httpd.crt"
+    HTTPD_KEY_FILE = "/etc/pki/tls/private/httpd.key"
     # only used on Fedora
     HTTPD_IPA_WSGI_MODULES_CONF = None
     OLD_IPA_KEYTAB = "/etc/httpd/conf/ipa.keytab"
--- a/ipaplatform/base/paths.py
+++ b/ipaplatform/base/paths.py
@@ -51,8 +51,8 @@ class BasePathNamespace(object):
     HTTPD_IPA_CONF = "/etc/httpd/conf.d/ipa.conf"
     HTTPD_NSS_CONF = "/etc/httpd/conf.d/nss.conf"
     HTTPD_SSL_CONF = "/etc/httpd/conf.d/ssl.conf"
-    HTTPD_CERT_FILE = "/etc/pki/tls/certs/httpd.crt"
-    HTTPD_KEY_FILE = "/etc/pki/tls/private/httpd.key"
+    HTTPD_CERT_FILE = "/var/lib/ipa/certs/httpd.crt"
+    HTTPD_KEY_FILE = "/var/lib/ipa/certs/httpd.key"
     # only used on Fedora
     HTTPD_IPA_WSGI_MODULES_CONF = None
     OLD_IPA_KEYTAB = "/etc/httpd/conf/ipa.keytab"
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -593,6 +593,7 @@ class HTTPInstance(service.Service):
         self.configure_mod_ssl_certs()
         self.set_mod_ssl_protocol()
         self.set_mod_ssl_logdir()
+        self.__add_include()
 
         self.cert = x509.load_certificate_from_file(paths.HTTPD_CERT_FILE)
 
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -577,6 +577,7 @@ class HTTPInstance(service.Service):
                                        pk12_password,
                                        paths.HTTPD_KEY_FILE)
 
+        self.backup_ssl_conf()
         self.configure_mod_ssl_certs()
         self.set_mod_ssl_protocol()
         self.set_mod_ssl_logdir()
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -414,9 +414,6 @@ class HTTPInstance(service.Service):
                     "The public key of the issued HTTPD service certificate "
                     "does not match its private key.")
 
-        # store the CA cert nickname so that we can publish it later on
-        # self.cacert_nickname = db.cacert_name
-        # FIXME: figure this out too
         sysupgrade.set_upgrade_state('ssl.conf', 'migrated_to_mod_ssl', True)
 
     def configure_mod_ssl_certs(self):
@@ -432,7 +429,14 @@ class HTTPInstance(service.Service):
                                    paths.IPA_CA_CRT, False)
 
     def __publish_ca_cert(self):
-        self.export_ca_certs_file(paths.CA_CRT, self.ca_is_configured)
+        ca_subject = self.cert.issuer
+        certlist = x509.load_certificate_list_from_file(paths.IPA_CA_CRT)
+        ca_certs = [c for c in certlist if c.subject == ca_subject]
+        if not ca_certs:
+            raise RuntimeError("HTTPD cert was issued by an unknown CA.")
+        # at this time we can assume any CA cert will be valid since this is
+        # only run during installation
+        x509.write_certificate(ca_certs[0], paths.CA_CRT)
 
     def is_kdcproxy_configured(self):
         """Check if KDC proxy has already been configured in the past"""
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -212,7 +212,6 @@ class HTTPInstance(service.Service):
         # There is no safe way to co-exist since there is no safe port
         # to make mod_nss use, disable it completely.
         if os.path.exists(paths.HTTPD_NSS_CONF):
-            self.fstore.backup_file(paths.HTTPD_NSS_CONF)
             installutils.remove_file(paths.HTTPD_NSS_CONF)
 
     def set_mod_ssl_protocol(self):
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -326,27 +326,34 @@ class HTTPInstance(service.Service):
                 certmonger.stop()
 
     def __setup_ssl(self):
-        db = certs.CertDB(self.realm, nssdir=paths.HTTPD_ALIAS_DIR,
-                          subject_base=self.subject_base, user="root",
-                          group=constants.HTTPD_GROUP,
-                          create=True)
-
         if self.pkcs12_info:
-            # db.init_from_pkcs12(self.pkcs12_info[0], self.pkcs12_info[1],
-            #                    ca_file=self.ca_file,
-            #                    trust_flags=trust_flags)
-            server_certs = db.find_server_certs()
-            if len(server_certs) == 0:
+            p12_certs, p12_priv_keys = certs.pkcs12_to_certkeys(
+                *self.pkcs12_info)
+            keys_dict = {
+                k.public_key().public_numbers(): k
+                for k in p12_priv_keys
+            }
+            certs_keys = [
+                (c, keys_dict.get(c.public_key().public_numbers()))
+                for c in p12_certs
+            ]
+            server_certs_keys = [
+                (c, k) for c, k in certs_keys if k is not None
+            ]
+
+            if not server_certs_keys:
                 raise RuntimeError(
                     "Could not find a suitable server cert in import in %s"
                     % self.pkcs12_info[0]
                 )
 
             # We only handle one server cert
-            nickname = server_certs[0][0]
-            if nickname == 'ipaCert':
-                nickname = server_certs[1][0]
-            self.cert = x509.load_der_x509_certificate(paths.HTTPD_CERT_FILE)
+            self.cert = server_certs_keys[0][0]
+            x509.write_certificate(self.cert, paths.HTTPD_CERT_FILE)
+            x509.write_pem_private_key(
+                server_certs_keys[0][1],
+                paths.HTTPD_KEY_FILE
+            )
 
             if self.ca_is_configured:
                 self.start_tracking_certificates()
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -419,7 +419,7 @@ class HTTPInstance(service.Service):
                                    paths.IPA_CA_CRT, False)
 
     def __publish_ca_cert(self):
-        self.import_ca_certs_file(paths.CA_CRT, self.ca_is_configured)
+        self.export_ca_certs_file(paths.CA_CRT, self.ca_is_configured)
 
     def is_kdcproxy_configured(self):
         """Check if KDC proxy has already been configured in the past"""
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -116,6 +116,8 @@ class HTTPInstance(service.Service):
         self.step("stopping httpd", self.__stop)
         self.step("backing up ssl.conf", self.backup_ssl_conf)
         self.step("disabling nss.conf", self.disable_nss_conf)
+        self.step("configuring mod_ssl certificate paths",
+                  self.configure_mod_ssl_certs)
         self.step("setting mod_ssl protocol list to TLSv1.0 - TLSv1.2",
                   self.set_mod_ssl_protocol)
         self.step("configuring mod_ssl log directory",
@@ -198,6 +200,11 @@ class HTTPInstance(service.Service):
         tasks.configure_http_gssproxy_conf(IPAAPI_USER)
         services.knownservices.gssproxy.restart()
 
+    def get_mod_nss_nickname(self):
+        cert = installutils.get_directive(paths.HTTPD_NSS_CONF, 'NSSNickname')
+        nickname = installutils.unquote_directive_value(cert, quote_char="'")
+        return nickname
+
     def backup_ssl_conf(self):
         self.fstore.backup_file(paths.HTTPD_SSL_CONF)
 
@@ -225,7 +232,6 @@ class HTTPInstance(service.Service):
         if sysupgrade.get_upgrade_state('http', OCSP_ENABLED) is None:
             self.__disable_mod_ssl_ocsp()
             sysupgrade.set_upgrade_state('http', OCSP_ENABLED, False)
->>>>>>> 754c26ef6... Revisions for ocsp
 
     def __disable_mod_ssl_ocsp(self):
         aug = Augeas(flags=Augeas.NO_LOAD | Augeas.NO_MODL_AUTOLOAD)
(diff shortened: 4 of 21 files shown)


ipalib/install/certmonger.py
```diff
@@ -670,11 +670,3 @@ def wait_for_request(request_id, timeout=120):
         raise RuntimeError("request timed out")
 
     return state
-
```
tiran: Why did you remove the test block?

ipaplatform/base/paths.py
```diff
@@ -48,6 +48,8 @@ class BasePathNamespace(object):
     HTTPD_IPA_CONF = "/etc/httpd/conf.d/ipa.conf"
     HTTPD_NSS_CONF = "/etc/httpd/conf.d/nss.conf"
     HTTPD_SSL_CONF = "/etc/httpd/conf.d/ssl.conf"
+    HTTPD_CERT_FILE = "/etc/pki/tls/certs/httpd.crt"
```
tiran: Are these the default file names for mod_ssl? Although we highly recommend to install IPA on its own machine, people tend to mix installation. I wonder if we should consider or own file names, e.g. ``ipa-httpd``.

ipaserver/install/httpinstance.py
```diff
@@ -237,82 +191,65 @@ def configure_gssproxy(self):
...
+#            aug.save()
 
-    def __set_mod_nss_passwordfile(self):
-        installutils.set_directive(paths.HTTPD_NSS_CONF, 'NSSPassPhraseDialog', 'file:' + paths.HTTPD_PASSWORD_CONF)
 
     def __add_include(self):
         """This should run after __set_mod_nss_port so is already backed up"""
-        if installutils.update_file(paths.HTTPD_NSS_CONF, '</VirtualHost>', 'Include {path}\n</VirtualHost>'.format(path=paths.HTTPD_IPA_REWRITE_CONF)) != 0:
+        if installutils.update_file(paths.HTTPD_SSL_CONF, '</VirtualHost>', 'Include {path}\n</VirtualHost>'.format(path=paths.HTTPD_IPA_REWRITE_CONF)) != 0:
             print("Adding Include conf.d/ipa-rewrite to %s failed." % paths.HTTPD_NSS_CONF)
```
tiran: ``HTTPD_SSL_CONF``

ipaserver/install/httpinstance.py
```diff
@@ -436,41 +326,50 @@ def __setup_ssl(self):
...
+        """Configure the mod_ssl certificate directives"""
+        installutils.set_directive(paths.HTTPD_SSL_CONF,
+                                   'SSLCertificateFile',
+                                   paths.HTTPD_CERT_FILE, False)
+        installutils.set_directive(paths.HTTPD_SSL_CONF,
+                                   'SSLCertificateKeyFile',
+                                   paths.HTTPD_KEY_FILE, False)
+        installutils.set_directive(paths.HTTPD_SSL_CONF,
+                                   'SSLCACertificateFile',
+                                   paths.IPA_CA_CRT, False)
```
tiran: I don't see support for encrypted key file. Are you planning to add https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslpassphrasedialog later?

ipaserver/install/installutils.py
```diff
@@ -447,10 +447,11 @@ class DirectiveSetter(object):
     with DirectiveSetter('/path/to/conf') as ds:
         ds.set(key, value)
     """
-    def __init__(self, filename, quotes=True, separator=' '):
+    def __init__(self, filename, quotes=True, separator=' ', comment='#'):
```
tiran: I like it, but where is ``comment`` argument used? I don't see it.

ipalib/install/certmonger.py
rcritten: Because it was broken and generally unused AFAIK so I didn't bother to fix it.

I can fix it if desired, I found it useful during development, but it dated back to the very first commit virtually unchanged (2010).

ipaplatform/base/paths.py
rcritten: Sure, it's an easy fix and now is the time to doit.

ipaserver/install/httpinstance.py
rcritten: Gah, this is why we have reviews.

ipaserver/install/httpinstance.py
rcritten: I think this is something the SPIKE team can add. I was focusing on just getting it working at all.

ipaserver/install/installutils.py
rcritten: So this requires either further docs or a switch to Augeas.

A lot of the values in the default ssl.conf are commented out which means that new values are appended to the end of the file, outside of the VirtualHost. This can still work but the context is wrong and could really go badly if someone defined other VirtualHost with different trust chains.

I added the comment bit so that a new value that was otherwise commented out could be added near where it should have been. This code is a bit fragile and doesn't account for extra spaces, etc. but it worked in this one small use-case.

Augeas would be more robust.

tiran:
Test issues
```
________________ test_set_directive_lines.test_remove_directive ________________
self = <ipatests.test_ipaserver.test_install.test_installutils.test_set_directive_lines object at 0x7f5b86c352e8>
    def test_remove_directive(self):
        lines = installutils.set_directive_lines(
>           False, '=', 'foo', None, EXAMPLE_CONFIG)
E       TypeError: set_directive_lines() missing 1 required positional argument: 'comment'
```

lint issues (try ``make fastcheck``)

```
************* Module ipatests.test_ipaserver.test_install.test_installutils
ipatests/test_ipaserver/test_install/test_installutils.py:32: [E1120(no-value-for-parameter), test_set_directive_lines.test_remove_directive] No value for argument 'comment' in function call)
ipatests/test_ipaserver/test_install/test_installutils.py:37: [E1120(no-value-for-parameter), test_set_directive_lines.test_add_directive] No value for argument 'comment' in function call)
ipatests/test_ipaserver/test_install/test_installutils.py:42: [E1120(no-value-for-parameter), test_set_directive_lines.test_set_directive_does_not_clobber_suffix_key] No value for argument 'comment' in function call)
************* Module ipaserver.install.httpinstance
ipaserver/install/httpinstance.py:25: [W0611(unused-import), ] Unused import pwd)
ipaserver/install/httpinstance.py:26: [W0611(unused-import), ] Unused import re)
ipaserver/install/httpinstance.py:30: [W0611(unused-import), ] Unused import locale)
ipaserver/install/httpinstance.py:33: [W0611(unused-import), ] Unused import six)
ipaserver/install/httpinstance.py:34: [W0611(unused-import), ] Unused Augeas imported from augeas)
ipaserver/install/httpinstance.py:38: [W0611(unused-import), ] Unused EXTERNAL_CA_TRUST_FLAGS imported from ipapython.certdb)
ipaserver/install/httpinstance.py:38: [W0611(unused-import), ] Unused IPA_CA_TRUST_FLAGS imported from ipapython.certdb)
************* Module ipaserver.install.service
ipaserver/install/service.py:399: [W0612(unused-variable), Service.import_ca_certs_file] Unused variable 'ext_key_usage')
ipaserver/install/service.py:399: [W0612(unused-variable), Service.import_ca_certs_file] Unused variable 'nickname')
ipaserver/install/service.py:399: [W0612(unused-variable), Service.import_ca_certs_file] Unused variable 'trusted')
************* Module ipaserver.advise.plugins.smart_card_auth
ipaserver/advise/plugins/smart_card_auth.py:11: [E0611(no-name-in-module), ] No name 'NSS_OCSP_ENABLED' in module 'ipaserver.install.httpinstance')
```

ipaserver/install/service.py
```diff
@@ -377,10 +377,40 @@ def add_cert_to_service(self):
...
+
+        ca_certs = None
+        try:
+            ca_certs = certstore.get_ca_certs(
+                conn, self.suffix, self.realm, ca_is_configured)
+        except errors.NotFound:
+            pass
+        else:
+            with open(cafile, 'wb') as fd:
+                for cert, nickname, trusted, ext_key_usage in ca_certs:
```
stlaz: Maybe something like `x509.write_certificate_list([cert for cert, _nick, _trust, _eku in ca_certs], cafile)`?

Even if you don't choose that way, you'll probably need to prepend the variable names with "_". I did not check with pylint but I am pretty much sure it will yell at you for unused variables.

ipaserver/install/service.py
```diff
@@ -377,10 +377,40 @@ def add_cert_to_service(self):
...
+        Import the CA certificates stored in LDAP into an NSS database
+
+        :param db: the target NSS database
+        :param ca_is_configured: whether IPA is CA-less or not
+        :param conn: an optional LDAP connection to use
+        """
         if conn is None:
             conn = api.Backend.ldap2
 
+        ca_certs = None
```
stlaz: `ca_certs` will always be defined in the `else` branch.  Even if the universe goes wrong and it's not , this fix wouldn't make it better because `None` is not an iterable, so the `for` below would still throw an exception.

ipaserver/install/httpinstance.py
```diff
@@ -436,41 +326,50 @@ def __setup_ssl(self):
...
-                    post_command='restart_httpd')
+                    post_command='restart_httpd',
+                    storage='FILE',
+                )
             finally:
                 if prev_helper is not None:
                     certmonger.modify_ca_helper('IPA', prev_helper)
-
-            self.cert = db.get_cert_from_db(self.cert_nickname)
+                self.cert = x509.load_certificate_from_file(
```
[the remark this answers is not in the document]
rcritten: Yeah, not sure why I moved it inside, done.

ipaserver/install/service.py
rcritten: Fixed as part of the pylint tests fixes. Note that I'm purposely leaving one failing for now, the ipa-advise change, because I'm not sure why OCSP has to be disabled.

ipaserver/install/service.py
rcritten: Removed

ipaserver/install/httpinstance.py
```diff
@@ -345,79 +277,32 @@ def configure_certmonger_renewal_guard(self):
...
     def __setup_ssl(self):
         db = certs.CertDB(self.realm, nssdir=paths.HTTPD_ALIAS_DIR,
                           subject_base=self.subject_base, user="root",
                           group=constants.HTTPD_GROUP,
                           create=True)
-        self.disable_system_trust()
-        self.create_password_conf()
 
         if self.pkcs12_info:
-            if self.ca_is_configured:
```
stlaz: If you want, I can take care of the CA-less scenario here.

rcritten: That would be awesome and fits my plan to pull other devs in to complete the remaining tasks, thanks!

ipaserver/install/httpinstance.py
```diff
@@ -57,52 +53,6 @@
 HTTPD_USER = constants.HTTPD_USER
 KDCPROXY_USER = constants.KDCPROXY_USER
 
-# See contrib/nsscipersuite/nssciphersuite.py
-NSS_CIPHER_SUITE = [
```
stlaz: Should we port this into the format of the `SSLCipherSuite` directive?

stlaz:
Note to self: this should be tested on some system with FIPS-mode enabled.

ipaserver/install/httpinstance.py
tiran: Either we should no longer set a suite at all and leave it to the system policy. Or we should make it a platform decision and don't touch system policies for Fedora and RHEL.

stlaz:
@rcritten: Are we considering domain level 0 to be supported in an IPA version with this change, too?

ipaserver/install/httpinstance.py
stlaz: I guess it's fine to go with the system policies and possibly add cipher suites in the future in case the defaults are no longer good enough.

rcritten: Right, my goal was to use the system policy.

rcritten:
@stlaz I'm not aware of any reason why DL 0 shouldn't be supported.

stlaz:
@rcritten I ask that because you did not handle that case here yet and because it's not yet contained in your TODO list. We'll need to handle HTTPD certs export and import in `ipa-replica-prepare` but that should not be that much of a problem.

rcritten:
I think testing DL0 is an excellent thing to add to the TODO list!

stlaz:
@rcritten: +1
+TODO: update the `upload_cacrt.py` update plugin.

tiran:
linter is still complaining:
```
************* Module ipaserver.install.httpinstance
ipaserver/install/httpinstance.py:31: [W0611(unused-import), ] Unused Augeas imported from augeas)
************* Module ipaserver.advise.plugins.smart_card_auth
ipaserver/advise/plugins/smart_card_auth.py:11: [E0611(no-name-in-module), ] No name 'NSS_OCSP_ENABLED' in module 'ipaserver.install.httpinstance')
```

ipaserver/install/httpinstance.py
```diff
@@ -436,41 +321,50 @@ def __setup_ssl(self):
...
+                                   paths.HTTPD_KEY_FILE, False)
+        installutils.set_directive(paths.HTTPD_SSL_CONF,
+                                   'SSLCACertificateFile',
+                                   paths.IPA_CA_CRT, False)
 
     def __publish_ca_cert(self):
-        ca_db = certs.CertDB(self.realm, nssdir=paths.HTTPD_ALIAS_DIR,
-                             subject_base=self.subject_base)
-        ca_db.export_pem_cert(self.cacert_nickname, paths.CA_CRT)
+        self.import_ca_certs_file(paths.CA_CRT, self.ca_is_configured)
```
stlaz: Previously, this code would only publish the last CA cert in the chain, now it pastes the whole chain there. This is most probably used only in legacy clients and as a very improbable fallback to receiving the CA cert in `ipa-client-install`.
I wonder - could this behavior change have negative effect in such scenarios? Or should we have actually had the full CA chain there from the beginning?

tiran: See https://pagure.io/freeipa/issue/6454

A CA trust store should only contain trust anchors, not intermediate CA certs. Since most of our infrastructure doesn't support AIA chasing and AIA chasing has also a negative performance impact, intermediate CA certs must be part of the the cert chain in TLS handshake.

rcritten:
@tiran I can fix the Augeas issue but I purposely left the failure in the ipa-advise plugin as a reminder that this needs to be fixed properly. I don't know why OCSP has to be disabled in the web server TLS provider for smart cards to work.

rcritten:
Turns out I had it totally wrong. The advise script turns **on** OCSP which makes total sense.

I can investigate how to do this in mod_ssl. I think they may have added similar directives.

rcritten:
It occurs to me that I changed the upgrade directive for having disabled OCSP. Please give this some thought. It may not really be a problem since the previous update would have reverted in nss.conf which won't exist anymore, or at worst will be reverted to its original backup state.

OCSP is disabled b/c the AVA has ipa-ca built in which can cause issues when IPA DNS is not used.

stlaz:
I added 2 patches that should allow us to install freeIPA in CA-less mode, also 2 patches that fix minor code issues. I can create separate PRs for the latter 2 if required.

Please note that the fix of `upload_cacrt.py` is but a hotfix and would probably need to be addressed again in about half a year when 389-ds moves from NSS to SSL, too. I could fix this perhaps more properly by implementing something like `CertBundle` class that would allow us to keep the NSS workflows we grew dependent on, such a class could also encapsulate PEM/PKCS12 files with the actions I am performing on the PKCS#12 cert/keys in `HTTPInstance` in my patches here.

Let me know what you think.

stlaz:
Apparently there were some merge conflicts but `git am` did not notify me. I'll see what I can do about it.

rcritten:
I had some sort of odd merge patch at HEAD. I dropped it because it was all garbage, rebased against current master and force pushed. Should be saner now.

stlaz:
I did the rebase of your branch just yesterday (sorry). Not sure what happened here, but it seems like you took one of Felipe's patches and applied it on top 🤕 
I was looking at the history of the `upload_cacrt.py` file and I guess it's there because we did not always store the CA cert in the database so it's there just for the upgrade path from some very old FreeIPA versions.

stlaz:
Did the required changes + fixed a DL0 replica installation issue where we supply an empty password to extract the PKCS#12 file.

stlaz:
+TODO: rework `ipa-server-certinstall`

ipaserver/install/httpinstance.py
```diff
@@ -237,63 +191,53 @@ def configure_gssproxy(self):
...
-
-    def __disable_mod_nss_ocsp(self):
+    def backup_ssl_conf(self):
+        self.fstore.backup_file(paths.HTTPD_SSL_CONF)
+
+    def disable_nss_conf(self):
+        # There is no safe way to co-exist since there is no safe port
+        # to make mod_nss use, disable it completely.
+        if os.path.exists(paths.HTTPD_NSS_CONF):
+            installutils.remove_file(paths.HTTPD_NSS_CONF)
```
stlaz: You can’t just remove a file we don’t own. If the user decides to uninstall IPA, his system may be broken should they try to install something relying on mod_nss, whichever devil that is. Removing the file also breaks the scenario when you uninstall IPA and then try to downgrade the packages and install it again.

rcritten: It should be restored on uninstallation.

The problem is that there is a listen port defined in nss.conf and even if the engine is off Apache is going to listen on that port. So what port should be used? I punted.

What we could do instead is to move nss.conf to nss.conf.ipa in the same directory and create a zero-length nss.conf. If an admin wanted to restore it they could and we would let them deal with the port conflict. It just makes uninstallation quite a bit more complex.

stlaz: My point was exactly as you're saying - it should be restore on uninstallation. You can just push the file to `fstore` that the HTTPInstance uses and restore it at the point where all other files are being restored.

ipalib/x509.py
```diff
@@ -535,6 +568,22 @@ def write_certificate_list(certs, filename):
...
 
+def write_pem_private_key(priv_key, filename):
+    """
+    Write a private key to a file in PEM format.
+
+    :param priv_key: cryptography ``PrivateKey`` object
+    """
+    try:
+        with open(filename, 'wb') as fp:
+            fp.write(priv_key.private_bytes(
```
stlaz: This is potentially dangerous and would write the file as 0x644 if new. I'll be forcing 0x600 in the upcoming update.

stlaz:
The latest update fixes the access to written priv. keys + __hopefully__ correctly handles `ipa-certupdate` and `ipa-server-certinstall`

stlaz:
While restore succeeds, the WebUI is inaccessible due to a JS error:
```
ErrorcolumnNumber: 377
fileName: "https://masina.address.cool/ipa/ui/js/dojo/dojo.js?v=40690"
info: Array [ "js/freeipa/plugins.js?40690", error ]
lineNumber: 1
message: "scriptError"
src: "dojoLoader"
stack: "l@https://masina.address.cool/ipa/ui/js/dojo/dojo.js?v=40690:1:377\np.injectUrl/o<@https://masina.address.cool/ipa/ui/js/dojo/dojo.js?v=40690:1:12399\n"
__proto__: {…}constructor: function Error()message: ""
name: "Error"
stack: Getter & Setter
toSource: function toSource()
toString: function toString()
__proto__: Object { … }
dojo.js:1:13187
```
I am not a JS developer and have no idea what happens in that minimalized script.

stlaz:
The same happens even during a simple ugprade w/o backup-restore, will try to check whether that may be something in master.

pvoborni:
Not loading `js/freeipa/plugins.js` and thus breaking Web UI is caused by missing inclusion of redirection rules for https port. I.e. old `nss.conf` has line at the end `Include /etc/httpd/conf.d/ipa-rewrite.conf`, `ssl.conf` doesn't have this line. Adding the config line to `ssl.conf` makes Web UI work again.

stlaz:
The upgrade is fixed now. I'll go through the code as I noticed some FIXUP notes I did not address yet before submitting this for another review.

ipaserver/install/service.py
```diff
@@ -377,7 +379,36 @@ def add_cert_to_service(self):
             logger.critical("Could not add certificate to service %s entry: "
                             "%s", self.principal, str(e))
 
-    def import_ca_certs(self, db, ca_is_configured, conn=None):
+    def import_ca_certs_file(self, cafile, ca_is_configured, conn=None):
```
stlaz: I'm sorry, I just find the name of the method too confusing. Will call this couple `export_ca_certs_{file,nssdb}` in another fixup.

ipaserver/install/httpinstance.py
```diff
@@ -436,41 +367,50 @@ def __setup_ssl(self):
...
+            #      raise RuntimeError("Could not find a suitable server cert.")
 
         # store the CA cert nickname so that we can publish it later on
-        self.cacert_nickname = db.cacert_name
-
-    def __import_ca_certs(self):
-        db = certs.CertDB(self.realm, nssdir=paths.HTTPD_ALIAS_DIR,
-                          subject_base=self.subject_base)
-        self.import_ca_certs(db, self.ca_is_configured)
+        # self.cacert_nickname = db.cacert_name
```
stlaz: Isn't this done just OK in `__publish_ca_cert`?

stlaz: Never mind, it isn't. We mean to publish the lowest intermediate cert in the chain (at least in external CA case).

stlaz:
The latest patchset addresses the FIXME notes and removes some lines that were previously commented-out (that one patch should probably be used in a rebase as fixup).

@rcritten if you're aware of any change I forgot, please, let me know, otherwise this PR could be reviewed.

Once pushed we can think about encrypting the private keys with a password and moving HTTPD_{CERT,KEY}_FILE to an IPA-specifi location (or should the latter be done in this PR?).

stlaz:
I added a patch that moves the HTTPD cert/key files to a different location since otherwise a new PR for it would break IPA for any developers working on the master branch and it does not seem to be that much work anyway.

**WARNING:** Rebased on current master, please pull before force-pushing here.

Removing WIP tagging.

ipaserver/install/httpinstance.py
```diff
@@ -402,11 +402,14 @@ def __setup_ssl(self):
...
+                    f.read(), None, backend=x509.default_backend())
+
             # Verify we have a valid server cert
-            # FIXME: come up with openssl equivalent
-            #  server_certs = db.find_server_certs()
-            #  if not server_certs:
-            #      raise RuntimeError("Could not find a suitable server cert.")
+            if (priv_key.public_key().public_numbers()
+                    != self.cert.public_key().public_numbers()):
+                raise RuntimeError("Could not find a suitable server cert.")
```
rcritten: I'd prefer including a more specific error message like "private key does not match public key." IIRC the previous invocation would also verify the cert trust chain was correct but that may not be as easy to do with openssl which is why had the original FIXME. Comparing the keys would be a good first step and there are other restrictions in place elsewhere so this code may be sufficient.

ipaserver/install/httpinstance.py
```diff
@@ -429,7 +426,14 @@ def configure_mod_ssl_certs(self):
...
     def __publish_ca_cert(self):
-        self.export_ca_certs_file(paths.CA_CRT, self.ca_is_configured)
+        ca_subject = self.cert.issuer
+        certlist = x509.load_certificate_list_from_file(paths.IPA_CA_CRT)
+        ca_certs = [c for c in certlist if c.subject == ca_subject]
+        if not ca_certs:
+            raise RuntimeError("HTTPD cert was issued by an unknown CA.")
+        # at this time we can assume any CA cert will be valid since this is
+        # only run during installation
+        x509.write_certificate(ca_certs[0], paths.CA_CRT)
```
rcritten: Just curious why you went with a list here instead of a single certificate object? If there is more than one matching subject should things blow up anyway?

ipaserver/install/httpinstance.py
rcritten: I think the scenario I was thinking of was: you have IPA installed with mod_nss, you upgrade and things get converted to mod_ssl and mod_nss is backed up. You uninstall IPA. The restored mod_nss still has IPA configured.

ipaserver/install/httpinstance.py
stlaz: I am not sure there could be a case where there would be multiple certificates with the same subject in `paths.IPA_CA_CRT`. I assumed there would be only one, but the easiest way to get it from list of CA certs in `paths.IPA_CA_CRT` was a list comprehension which still results in a list. If you want, I could perhaps add a check that there's just one cert in the filtered-out list.

ipaserver/install/httpinstance.py
stlaz: Fixed the error message.

I did not find it anywhere in this path but if there was a cert trust chain verification, it must have happened in certmonger while it was adding the certificate to the NSS DB and thus I am assuming it would still be there.

The original `db.find_server_certs()` only returned a list of certificates the database kept private keys of.

rcritten:
The perms on httpd.crt are tighter than they need to be at 0600. Not a deal breaker.

Not sure the scenario matters yet but I installed 4.6.1-3, replaced Apache and 389-ds certs using ipa-server-certinstall, upgraded to master. Things worked ok but uninstall failed because apache couldn't restart. This is because:

httpd: Syntax error on line 353 of /etc/httpd/conf/httpd.conf: Syntax error on line 219 of /etc/httpd/conf.d/ssl.conf: Could not open configuration file /etc/httpd/conf.d/ipa-rewrite.conf: No such file or directory

rcritten:
I tested successfully (minus uninstall error mentioned already):

- New install using new bits
- Install on F27 using 4.6.1-3 upgrade to new bits
- Install on F27 CA-less upgrade to new bits
- Install on F27, ipa-server-certinstall, upgrade
- New install on F27, ipa-server-certinstall

I think just the uninstall is a potential blocker.

stlaz:
> Not sure the scenario matters yet but I installed 4.6.1-3, replaced Apache and 389-ds certs using ipa-server-certinstall, upgraded to master. Things worked ok but uninstall failed because apache couldn't restart. This is because:

> httpd: Syntax error on line 353 of /etc/httpd/conf/httpd.conf: Syntax error on line 219 of /etc/httpd/conf.d/ssl.conf: Could not open configuration file /etc/httpd/conf.d/ipa-rewrite.conf: No such file or directory

I was not able to reproduce this, could it be that you had a machine that might have possibly had unclean sysrestore or similars? During the uninstallation, HTTPD restores original files like ssl.conf and nss.conf and then tries to restart but the mentioned files get rewritten with their original copies and thus such problems should not occur.

stlaz:
Never mind, scratch my previous comment, just reproduced it while doing exactly what you said, previously I thought it was just the upgrade that broke it.
For some reason, the original `ssl.conf` was not backed up in this scenario. That did not cause uninstall failure in my case, but the `ssl.conf` file kept IPA-values. I need to investigate.

stlaz:
Found the bug, we actually really weren't storing the original `ssl.conf` in HTTPInstance mod_nss to mod_ssl migration code.

rcritten:
I'm seeing the same failure in mod_nss now:

httpd: Syntax error on line 353 of /etc/httpd/conf/httpd.conf: Syntax error on line 216 of /etc/httpd/conf.d/nss.conf: Could not open configuration file /etc/httpd/conf.d/ipa-rewrite.conf: No such file or directory

I think this is because the restored nss.conf is the one from the pre-upgrade state (not the pristine state).

ipaserver/install/httpinstance.py
tiran: Multiple subjects with the same subject and SKID are a valid use case. For example they are used to provide an alternative chain to an alternative root CA. For example Let's Encrypt is still cross signed by two CAs.

## letsencryptauthorityx3.pem
```
Issuer: C = US, O = Internet Security Research Group, CN = ISRG         
Subject: C = US, O = Let's Encrypt, CN = Let's Encrypt Authority X3
```

## lets-encrypt-x3-cross-signed.pem
```
Issuer: O = Digital Signature Trust Co., CN = DST Root CA X3
Subject: C = US, O = Let's Encrypt, CN = Let's Encrypt Authority X3
```

stlaz:
Hm, I was counting on the correctness of the docstring in `FileStore.backup_file()` that if a file is already backed up, it won't be again, but apparently it's wrong. I'll see what I can do with this.

stlaz:
I added a check that we're not already backing up the `nss.conf` in the 
"httpinstance: backup mod_nss conf instead of just removing it" commit. Untested.

rcritten:
I think one final issue, SSLCACertificateFile is getting added to ssl.conf twice. Once after the comment inside the VH and one appended to the end. I think it is because found is not being set.

rcritten:
This fixes it and another bug where when a comment match is found a line is lost:

diff --git a/ipaserver/install/installutils.py b/ipaserver/install/installuti
ls.py
index fe481f1c5..885549a80 100644
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -578,8 +578,10 @@ def set_directive_lines(quotes, separator, k, v, li
nes, comment):
             if v is not None:
                 yield new_line
         elif addnext:
+            found = True
             addnext = False
             yield new_line
+            yield line
         elif cmatcher.match(line):
             addnext = True
             yield line

stlaz:
Worked for me, was unsure how to push and keep authorship correct so I merged it into `Add value in set_directive after a commented-out version`

stlaz:
master:

* 805aea2443c902cd7c5525102701f116b43da575 Use mod_ssl instead of mod_nss for Apache TLS for new installs
* a0407f75f9b25c265008e907c7f97e2e4b0fa6e8 Remove main function from the certmonger library
* 5c64e28512be88f597cec0536576dbaba8b92878 Convert ipa-pki-proxy.conf to use mod_ssl directives
* 4596674481dbbeb9026f0ff793699479e1cfb09d Enable upgrades from a mod_nss-installed master to mod_ssl
* 5531c9f26b1910ec9f852d5ae0da7ddf078ec8fc Don't backup nss.conf on upgrade with the switch to mod_ssl
* 4d2c7a4a75a35d19a1ce069b45d3ac2d5ad80376 Add value in set_directive after a commented-out version
* fa135e6ef123c1ec48463d3389d01363395e2d8e Update smart_card_auth advise script for mod_ssl
* 7dc923cc4c2ec7c01156589f8314c78439dbc857 mod_ssl migration: fix upload_cacrt.py plugin
* 205675239a7ecd66ae77b01e90158b5a559b840e httpinstance: handle supplied PKCS#12 files in installation
* 8789afa18af968032e3f2159a48a6493f6755f47 x509: Remove unused argument of load_certificate_from_file()
* 60aa2c3b38e1f95c465821484f86657fa696e02b x509: Fix docstring of write_certificate()
* c85bf376f06cda17305a8723755f3635a3da275a certupdate: don't update HTTPD NSS db
* 9a7c3159841c10e353861519a1b25bd9b0ef773d Make ipa-server-certinstall store HTTPD cert in a file
* 92d91ed58be62e50aadeac5396602429a02f711e fixup: add ipa-rewrite.conf to ssl.conf on upgrade
* 0c388d1e8f9bdadf83c20ee4186470f0eee593a9 service: rename import_ca_certs_* to export_*
* dde62ff883dbd6f13763f789521d37fca125ea6d httpinstance: backup mod_nss conf instead of just removing it
* 8ea04ab3e3eef431d529c0d99e9d1ea5a65933e1 httpinstance: verify priv key belongs to certificate
* ee49947b6c258a7cff6fa126f06905f297a3e07b httpinstance: fix publishing of CA cert
* 1ca68ea7300daa125eb2fec9aa2de2122b3fd95e httpinstance fixup: remove commented-out lines
* b21941360cdfbcce9fc891befc160103b4678eec Move HTTPD cert/key pair to /var/lib/ipa/certs
* 75845733f86df5ddeba81d8ce5312be0eb511262 Backup ssl.conf when migrating from mod_nss

PR #1452, opened 2018-01-09
[Backport][ipa-4-6] Idviews: fix objectclass violation on idview-add

tiran:
When the option --domain-resolution-order is used with the command
ipa idview-add, the resulting LDAP object stores the value in
ipadomainresolutionorder attribute.
The issue is that the add command does not add the needed object
class (ipaNameResolutionData) because it is part of
possible_objectclasses but not of object_class.

The fix makes sure to add the objectclass when the option
--domain-resolution-order is used, and adds a non-regression test.

Note that idview-mod does not have any issue as it correctly handles
the addition of missing possible objectclasses.

Fixes:
https://pagure.io/freeipa/issue/7350

Reviewed-By: Alexander Bokovoy <abokovoy@redhat.com>

Manual backport of PR #1437

### DIFF
--- a/ipaserver/plugins/idviews.py
+++ b/ipaserver/plugins/idviews.py
@@ -22,10 +22,11 @@ import re
 import six
 
 from .baseldap import (LDAPQuery, LDAPObject, LDAPCreate,
-                                     LDAPDelete, LDAPUpdate, LDAPSearch,
-                                     LDAPAddAttributeViaOption,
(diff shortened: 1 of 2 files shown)


tiran:
ipa-4-6:

* 6c89b26623e896ccf7ad58d2301b509a55ca9c2e Idviews: fix objectclass violation on idview-add

PR #1458, opened 2018-01-10
Prepare migration of mod_nss NSSDB to sql format

tiran:
- Refactor CertDB to look up values from its NSSDatabase.
- Add run_modutil() helpers to support sql format. modutil does not
  auto-detect the NSSDB format.
- Add migration helpers to CertDB.
- Add explicit DB format to NSSCertificateDatabase stanza
- Restore SELinux context when migrating NSSDB.
- Add some debugging and sanity checks to httpinstance.

The actual database format is still dbm. Certmonger on Fedora 26 does
neither auto-detect DB format nor support SQL out of the box.

https://pagure.io/freeipa/issue/7354

### DIFF
--- a/.test_runner_config.yaml
+++ b/.test_runner_config.yaml
@@ -43,6 +43,7 @@ steps:
       systemd_journal.log
       `find daemons -name '*.log' -print`
   - chown ${uid}:${gid} ${container_working_dir}/var_log.tar
+  - ls -laZ /etc/dirsrv/slapd-*/ /etc/httpd/alias/ /etc/pki/pki-tomcat/alias/ || true
   configure:
   - ./autogen.sh
   install_packages:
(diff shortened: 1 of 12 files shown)


tiran:
This PR is required to make FreeIPA 4.6 working on rawhide / Fedora 28. PR #1449 will be 4.7-only.

rcritten:
This works for me if I start with a sqlite database.

It fails if I start with a dbm database with:
2018-01-10T18:01:37Z DEBUG args=/usr/bin/certutil -d sql:/etc/httpd/alias -N -f /etc/httpd/alias/pwdfile.txt -@ /etc/httpd/alias/pwdfile.txt
2018-01-10T18:01:37Z DEBUG Process finished, return code=255
2018-01-10T18:01:37Z DEBUG stdout=
2018-01-10T18:01:37Z DEBUG stderr=No password file "/etc/httpd/alias/pwdfile.txt" exists.
Invalid password.

And sure enough there is no pwdfile.txt. This is an assumption of IPA because by default the database has no pin.

tiran:
Fix is blocked by Certmonger bug https://bugzilla.redhat.com/show_bug.cgi?id=1533368

tiran:
My PR #1474 is the successor of this PR.

PR #1459, opened 2018-01-11
Make IntegrationTest fail if an error happened during uninstall

felipevolpone:
Before this change, if the uninstall process fails, the test would not fail, due to the raiseonerr=False.

Fixes: https://pagure.io/freeipa/issue/7357

The results can be checked here:
https://fedorapeople.org/groups/freeipa/prci/jobs/e0c64916-f6ff-11e7-baa4-001a4a23169a/
https://fedorapeople.org/groups/freeipa/prci/jobs/05d5be8a-f700-11e7-a97e-001a4a231699/


felipevolpone:
This would help us to prevent what happened with PR 1410.

tiran:
master:

* d945583cc77a723f7108c36555cdbe2b4da2d495 Make IntegrationTest fail if an error happened during uninstall

PR #1460, opened 2018-01-12
test to check if recommended replication agreement exceeds

mrizwan93:
Maximum recommended number of agreements per replica is 4.
This test checks if number of agreement exceeds the recommendation,
warning should be given to the user

related ticket : https://pagure.io/freeipa/issue/6533

### DIFF
--- a/ipatests/test_integration/test_replica_promotion.py
+++ b/ipatests/test_integration/test_replica_promotion.py
@@ -570,3 +570,24 @@ class TestReplicaInstallWithExistingEntry(IntegrationTest):
         master.run_command(arg)
 
         tasks.install_replica(master, replica)
+
+
+class TestRecomendedReplicationAgreement(IntegrationTest):
+    """Maximum recomended number of agreements per replica is 4.
+    This test checks if number of agreement exceeds the recomendation,
+    warnning should be given to the user
+
+    related ticket : https://pagure.io/freeipa/issue/6533"""
+
+    num_replicas = 5
+
+    def test_recomended_replication_agreement(self):
+        tasks.install_topo('star', self.master, self.replicas,
+                           [], setup_replica_cas=False)
+        arg = ['ipa', 'topologysuffix-verify', 'domain']
+        cmd = self.master.run_command(arg)
+        expected_str1 = ("Recommended maximum number of "
+                         "agreements per replica exceeded")
+        expected_str2 = "Maximum number of agreements per replica: 4"
+        assert (expected_str1 in cmd.stdout_text) and (
+               expected_str2 in cmd.stdout_text)


ipatests/test_integration/test_replica_promotion.py
```diff
@@ -570,3 +570,36 @@ def test_replica_install_with_existing_entry(self):
         master.run_command(arg)
 
         tasks.install_replica(master, replica)
+
+
+ass TestRecomendedReplicationAgreement(IntegrationTest):
```
Tiboris: You meant class? 😆

mrizwan93: oh, yes.

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -570,3 +570,36 @@ def test_replica_install_with_existing_entry(self):
...
+    def test_recomended_replication_agreement(self):
+        master = self.master
+        replica1 = self.replicas[0]
+        replica2 = self.replicas[1]
+        replica3 = self.replicas[2]
+        replica4 = self.replicas[3]
+        replica5 = self.replicas[4]
+
+        tasks.install_master(master)
+        tasks.install_replica(master, replica1)
```
Rezney: I guess you may also use something like: 

`tasks.install_topo('star', self.master, self.replicas, [],
                           setup_replica_cas=False)`

which would be much simpler...

mrizwan93: thanks, it is much simple.

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -570,3 +570,24 @@ def test_replica_install_with_existing_entry(self):
...
+
+    related ticket : https://pagure.io/freeipa/issue/6533"""
+
+    num_replicas = 5
+
+    def test_recomended_replication_agreement(self):
+        tasks.install_topo('star', self.master, self.replicas,
+                           [], setup_replica_cas=False)
+        arg = ['ipa', 'topologysuffix-verify', 'domain']
+        cmd = master.run_command(arg)
```
Tiboris: Undefined variable 'master' :)

Tiboris:
Leaving final ack on @Rezney

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -570,3 +570,24 @@ def test_replica_install_with_existing_entry(self):
...
+    related ticket : https://pagure.io/freeipa/issue/6533"""
+
+    num_replicas = 5
+
+    def test_recomended_replication_agreement(self):
+        tasks.install_topo('star', self.master, self.replicas,
+                           [], setup_replica_cas=False)
+        arg = ['ipa', 'topologysuffix-verify', 'domain']
+        cmd = self.master.run_command(arg)
+        expected_str1 = ("Recommended maximum number of " +
```
stlaz: No need to concat the strings.

mrizwan93: It is exceeding 80 characters if not concatenated.

Tiboris: You can delete '+' and it will be as one string.

stlaz: What I mean is that:
```
("Recommended maximum number of "
"agreements per replica exceeded")
```
is still valid and considered one string, thus more lightweight.

felipevolpone: @mrizwan93 there is no need to use the `+` when a string exceeds the limits of chars per line. You can do this way:
```python
string = ("a very big big big  big  big  big  big  big  big  big  big  big big "
          "_ big string")
print(string)
>>> a very big big big  big  big  big  big  big  big  big  big  big big _ big string
```

stlaz:
You would probably be able to simplify this test with appropriate injections of the required entries to your LDAP database. That way you wouldn't need to provision five replicas just to run one command.
It would however require some research about what entries you need to add to the database.

ipatests/test_integration/test_replica_promotion.py
mrizwan93: @Tiboris @stlaz @felipevolpone  thanks.

Rezney:
LGTM from my POV however would you please consider @stlaz approach too?

stlaz:
I believe that the test as written now is but a waste of resources and I can't imagine anyone to be willing to run it. Please, do consider reworking this in a more reasonable manner.

felipevolpone:
From PR CI point of view, I can say that we hardly will be able to run a test with 5 replicas. There is some way to change the test to use less replicas?

mrizwan93:
I am figuring out if I can inject entries into ldap to simulate a 5 replica entry into it as suggested by standa. However my suggestion is to align this test case with one of the testcase in test_replication_layout.py, as we are installing number of replicas anyway there.

Rezney:
@mrizwan93 currently our upstream "test_replication_layout.py" yaml config provides only 3 replicas. 

https://github.com/freeipa/freeipa-ci/blob/master/jenkins-job-builder/job-groups/repo-based-jobgroups.yml#L256

It would really be nice if we could avoid adding additional 2 replicas just for one testcase,

stlaz:
The sole purpose of `test_replication_layout.py` with >20 replicas is to test whether such scenarios would work. I remember only one time when those many replica tests were all run. Having 5 replicas to test just one command call is just unreasonable.

Please note that `topologysuffix_find` utilizes `server_find` and `topologysegment_find` (with the given suffix as an argument) to perform its checks. If you run the mentioned commands like `ipa <cmd_name> --all --raw` in your five-replicas topology, you should hopefully be able to see which entries you need to add.

mrizwan93:
@Rezney @stlaz 
Okay, thanks for considering it.

stlaz:
Adding the "Postponed" tag may seem to imply that this does not have to be finished which is not true. Please implement the missing functionality.

rcritten:
Closing due to inactivity.

PR #1461, opened 2018-01-12
test for second phase of ipa-server-install with --external-ca when dirsrv instance is stopped

mrizwan93:
When the dirsrv service, which gets started during the first
ipa-server-install --external-ca phase, is not running when the
second phase is run with --external-cert-file options, the
ipa-server-install command fail.

This test checks if second phase installs successfully when dirsrv
is stoped.

related ticket: https://pagure.io/freeipa/issue/6611

### DIFF
--- a/ipatests/test_integration/test_external_ca.py
+++ b/ipatests/test_integration/test_external_ca.py
@@ -95,6 +95,7 @@ def install_server_external_ca_step2(host, ipa_ca_cert, root_ca_cert):
 
 def service_control_dirsrv(host, function):
     """Function to control the dirsrv service i.e start, stop, restart etc"""
+
     dashed_domain = host.domain.realm.replace(".", '-')
     dirsrv_service = "dirsrv@%s.service" % dashed_domain
     cmd = host.run_command(['systemctl', function, dirsrv_service])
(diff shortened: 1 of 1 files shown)


stlaz:
Copy-pasta is bad for you. Please consider splitting the first and second external CA installation steps into separate functions.

stlaz:
Getting there, although you still have that copy-pasta there.

Now what you should do is to split this PR into two commits. In the first one, you'll create those functions that allow running separate steps of IPA with external CA installation and use them in the `TestExternalCA`. That way you'll get rid of the copy-pasted code. In the second commit, you'll introduce your new test.

Please be so kind and follow the steps above :)

mrizwan93:
@stlaz Done.

felipevolpone:
You need to rebase the PR go the build green :)

stlaz:
Tests seem to pass, ACK.

tiran:
master:

* f1f180985f27154efd900a95350d007d2dab6b7d When the dirsrv service, which gets started during the first ipa-server-install --external-ca phase, is not running when the second phase is run with --external-cert-file options, the ipa-server-install command fail.
* 1ad27076a53e15f7a06b68fd4879aabbf2d57770 Updated the TestExternalCA with the functions introduced for the steps of external CA installation.

PR #1468, opened 2018-01-15
test_gssproxy: test duplicated sections

Rezney:
Related to the issue where ipa-server-install failed because gssproxy
was not able to start due to a duplicated section

https://pagure.io/freeipa/issue/7363


Rezney:
Not sure if the current test location is correct. 
Could the "gssproxy" restart somehow affect other tests? Have it tested and it is not the case but just asking to be sure.

Rezney:
Closing as invalid...

PR #1470, opened 2018-01-16
RFE: ipa client should setup openldap for GSSAPI

amitkumar50:
The IPA client installer currently edits /etc/openldap/ldap.conf, setting up
the client to consume LDAP data from IPA.  It currently sets:
>URI
>BASE
>TLS_CACERT

This PR makes ipa-client to add these two AV pair:
>SASL_MECH GSSAPI
>TLS_REQCERT demand

Resolves: https://pagure.io/freeipa/issue/7366

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -522,8 +522,12 @@ def configure_openldap_conf(fstore, cli_basedn, cli_server):
         {
             'name': 'comment',
             'type': 'comment',
-            'value': '   URI, BASE and TLS_CACERT have been added if they '
-                     'were not set.'
+            'value': '   URI, BASE, TLS_CACERT and SASL_MECH'
+        },
+        {
+            'name': 'comment',
+            'type': 'comment',
+            'value': '   have been added if they were not set.'
         },
         {
             'name': 'comment',
@@ -572,6 +576,12 @@ def configure_openldap_conf(fstore, cli_basedn, cli_server):
             'name': 'TLS_CACERT',
             'type': 'option',
             'value': paths.IPA_CA_CRT
+        },
+        {
+            'action': 'addifnotset',
+            'name': 'SASL_MECH',
+            'type': 'option',
+            'value': 'GSSAPI'
         },
     ]
 


abbra:
@amitkumar50, unfortunately, it is not so simple.
We cannot set these parameters in the `ldap.conf` for all cases. For example, on IPA masters SSSD talks to AD DCs using GSSAPI and Active Directory LDAP implementation does not allow to mix TLS and SASL GSSAPI encryption/integrity protection.

https://msdn.microsoft.com/en-gb/library/cc223500.aspx

> While Active Directory permits SASL binds to be performed on an
> SSL/TLS-protected connection, it does not permit the use of SASL-layer
> encryption/integrity verification mechanisms on such a connection. While
> this restriction is present in Active Directory on Windows 2000 Server
> operating system, Windows Server 2003 operating system, Windows Server
> 2008 operating system, Windows Server 2008 R2 operating system, Windows
> Server 2012 operating system, Windows Server 2012 R2 operating system,
> Windows Server 2016 operating system, and Windows Server operating
> system, versions prior to Windows Server 2008 can fail to reject an LDAP
> bind that is requesting SASL-layer encryption/integrity verification
> mechanisms when that bind request is sent on a SSL/TLS-protected
> connection.

If we set these settings in `ldap.conf`, they would get applied to SSSD in all cases which might break communication in case of trust to AD. SSSD has own way to set SSSD-specific LDAP settings but unfortunately, not for AD subdomains, I think.

Could you please test this change in a real world setup?

ipaclient/install/client.py
```diff
@@ -573,6 +577,18 @@ def configure_openldap_conf(fstore, cli_basedn, cli_server):
...
+            'action': 'addifnotset',
+            'name': 'SASL_MECH',
+            'type': 'option',
+            'value': 'GSSAPI'
+        },
+        {
+            'action': 'addifnotset',
+            'name': 'TLS_REQCERT',
+            'type': 'option',
+            'value': 'demand'
```
tiran: openldap defaults to cert validation. There is no need to add the option:

```
./libraries/libldap/init.c:     {0, ATTR_TLS,   "TLS_REQCERT",          NULL,   LDAP_OPT_X_TLS_REQUIRE_CERT},
```

amitkumar50:
@tiran Done changes. Thanks.

amitkumar50:
@abbra Will check in real world env soon. Thanks.

amitkumar50:
@abbra 
Tested in IPA AD Trust.

On IPA Client and IPA server both

Can do userlookups using 'id,getnent'

with following config in ldap.conf
TLS_CACERTDIR /etc/openldap/cacerts
SASL_NOCANON    on
URI ldaps://rhel7u4-6.gsslab.pnq2.redhat.com
BASE dc=gsslab,dc=pnq2,dc=redhat,dc=com
TLS_CACERT /etc/ipa/ca.crt
SASL_MECH GSSAPI

amitkumar50:
@tiran Does this means SASL_MECH to be removed from commit

tiran:
@amitkumar50 According to OpenLDAP's documentation, SASL_MECH isn't supported as ldap.conf option. However @flo-renaud 's comment https://bugzilla.redhat.com/show_bug.cgi?id=1527215#c5 shows the contrary.

Should we add an option that is documented as non-supported?

abbra:
@tiran not sure where you have found it is 'not supported'. In my git clone I can see 'SASL_MECH' as part of the man page:
```
SASL OPTIONS
       If OpenLDAP is built with Simple Authentication and Security Layer support,
       there are more options you can specify.

       SASL_MECH <mechanism>
              Specifies the SASL mechanism to use.  This is a user-only option.
```

tiran:
> This is a user-only option.

is documented as

> Some options are user-only.  Such options are ignored if present in the ldap.conf (or file specified by LDAPCONF).

abbra:
The documentation is not really up to date on this. If you look into `openldap_ldap_init_w_conf()` function, it uses a table `attrs` to map attributes and ignore some of them in the system config:

```C
static const struct ol_attribute {
        int                     useronly;
        int                     type;
        const char *    name;
        const void *    data;
        size_t          offset;
} attrs[] = { ... };
```

In this table only `SASL_AUTHZID` and `SASL_AUTHCID` are ignored in the system-wide config for SASL configuration options:
```C
#ifdef HAVE_CYRUS_SASL
        {0, ATTR_STRING,        "SASL_MECH",            NULL,
                offsetof(struct ldapoptions, ldo_def_sasl_mech)},
        {0, ATTR_STRING,        "SASL_REALM",           NULL,
                offsetof(struct ldapoptions, ldo_def_sasl_realm)},
        {1, ATTR_STRING,        "SASL_AUTHCID",         NULL,
                offsetof(struct ldapoptions, ldo_def_sasl_authcid)},
        {1, ATTR_STRING,        "SASL_AUTHZID",         NULL,
                offsetof(struct ldapoptions, ldo_def_sasl_authzid)},
        {0, ATTR_SASL,          "SASL_SECPROPS",        NULL,   LDAP_OPT_X_SASL_SECPROPS},
        {0, ATTR_BOOL,          "SASL_NOCANON", NULL,   LDAP_BOOL_SASL_NOCANON},
#endif
```

tiran:
@flo-renaud figured that out, too.

The big question is: is it a documentation bug or a code bug?

abbra:
It is definitely a documentation bug because these defaults make sense: `SASL_AUTHCID` and `SASL_AUTHZID` are per-requestor while the rest can be mostly the same for the whole setup. You definitely would want to enforce `SASL_MECH`, `SASL_REALM`, `SASL_NOCANON` system-wide.

tiran:
I'm going to raise the issue with Howard Chu. If he agrees it's a documentation bug, then I'm fine with this change.

tiran:
I opened upstream bug http://www.openldap.org/its/index.cgi?findid=8818

flo-renaud:
Hi @amitkumar50 
if /etc/openldap/ldap.conf already contains a value for SASL_MECH (before calling the installer), this value is not modified by the installer. Is it intentional?

amitkumar50:
@flo-renaud not sure about it at the moment.
But are there any other changes required in PR?

flo-renaud:
Hi @amitkumar50 
the PR looks good to me. As the other changes to ldap.conf are also not overriding pre-existing settings, I think it's the right thing to use addifnotset.

tiran:
I agree, too.

tiran:
master:

* 07be3306c16c0c8eb729b980c5bd7fdba8343433 RFE: ipa client should setup openldap for GSSAPI

PR #1472, opened 2018-01-16
Enable and start oddjobd after ipa-restore if it's not running. 

slaykovsky:
If after ipa-restore the service oddjobd is not running,
domain-level1 replica installation will fail during
ipa-replica-conncheck because this step is using oddjob
to start the process ipa-replica-conncheck on the master.

This patch fixes it. Also added regression test.

https://pagure.io/freeipa/issue/7234

### DIFF
--- a/ipaserver/install/ipa_restore.py
+++ b/ipaserver/install/ipa_restore.py
@@ -423,6 +423,12 @@ class Restore(admintool.AdminTool):
                 logger.info('Restarting SSSD')
                 sssd = services.service('sssd', api)
                 sssd.restart()
+                logger.info('Restarting oddjobd')
+                oddjobd = services.service('oddjobd', api)
+                if not oddjobd.is_enabled():
+                    logger.info("Enabling oddjobd")
+                    oddjobd.enable()
+                oddjobd.start()
                 http.remove_httpd_ccaches()
                 # have the daemons pick up their restored configs
                 run([paths.SYSTEMCTL, "--system", "daemon-reload"])
--- a/ipatests/test_integration/test_backup_and_restore.py
+++ b/ipatests/test_integration/test_backup_and_restore.py
@@ -28,7 +28,9 @@ from ipapython.dn import DN
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_plugins.integration import tasks
 from ipatests.test_integration.test_dnssec import wait_until_record_is_signed
+from ipatests.test_integration.test_simple_replication import check_replication
 from ipatests.util import assert_deepequal
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


abbra:
Thanks. I'd like you to split out the patch that changes assert_deepequal into a separate one.
Builds in CI will be failing today as COPR infra is under maintenance and our repositories aren't accessible.

slaykovsky:
Thanks, @abbra , will change it.

ipaserver/install/ipa_restore.py
```diff
@@ -423,6 +423,9 @@ def run(self):
                 logger.info('Restarting SSSD')
                 sssd = services.service('sssd', api)
                 sssd.restart()
+                logger.info('Restarting oddjobd')
+                oddjobd = services.service('oddjobd', api)
+                oddjobd.restart()
```
flo-renaud: I'm afraid this won't be sufficient to just restart the service oddjobd. We need to make sure that it is also enabled (so that it starts after a reboot for instance), and there may also be some work related to sysrestore (to store if the service was running/enabled prior to the installation).

ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -412,3 +413,45 @@ class TestBackupReinstallRestoreWithKRA(BaseBackupAndRestoreWithKRA):
...
+
+        with restore_checker(self.master):
+            backup_path = backup(self.master)
+
+            logger.info("Backup path for %s is %s", self.master, backup_path)
+
+            self.master.run_command([
+                "ipa-server-install", "--uninstall", "-U"
+            ])
+
```
flo-renaud: In order to fully test the fix, you may want to disable or stop oddjobd before the call to ipa-restore.

ipaserver/install/ipa_restore.py
slaykovsky: Hello @flo-renaud :) Thank you, will change it.

ipatests/test_integration/test_backup_and_restore.py
slaykovsky: Okay, will do.

flo-renaud:
Hi @slaykovsky 
pylint found issues:
```
************* Module ipatests.util
ipatests/util.py:694: [W1303(missing-format-argument-key), DummyClass.__process] Missing keyword argument 'args' for format string)
ipatests/util.py:694: [W1303(missing-format-argument-key), DummyClass.__process] Missing keyword argument 'name' for format string)
ipatests/util.py:694: [W1303(missing-format-argument-key), DummyClass.__process] Missing keyword argument 'kwargs' for format string)
************* Module ipatests.test_integration.test_backup_and_restore
ipatests/test_integration/test_backup_and_restore.py:112: [W0612(unused-variable), restore_checker] Unused variable '_')
```

slaykovsky:
@flo-renaud thanks!

ipaserver/install/ipa_restore.py
```diff
@@ -423,6 +423,12 @@ def run(self):
                 logger.info('Restarting SSSD')
                 sssd = services.service('sssd', api)
                 sssd.restart()
+                logger.info('Restarting oddjobd')
+                oddjobd = services.service('oddjobd', api)
+                if not oddjobd.is_enabled():
+                    logger.info("Enabling oddjobd")
+                    oddjobd.enable()
+                oddjobd.restart()
```
flo-renaud: Hi @slaykovsky 
No need to restart oddjobd if it is already running. You can replace this call with oddjobd.start()

ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -421,3 +423,53 @@ class TestBackupReinstallRestoreWithKRA(BaseBackupAndRestoreWithKRA):
...
+            self.master.run_command([
+                "systemctl", "disable", "oddjobd"
+            ])
+
+            dirman_password = self.master.config.dirman_password
+            self.master.run_command(
+                ["ipa-restore", backup_path],
+                stdin_text=dirman_password + '\nyes'
+            )
+
```
tiran: How about you check the status of ``oddjobd`` here, too?

```
result = self.master.run_command([
    "systemctl", "status", "oddjobd"
])
assert "active (running)" in result.stdout_text
```

slaykovsky: I can add it but I think it's redundant as replica install will fail anyway if `oddjobd` is not running.

tiran: OK, SGTM.

tiran:
Awesome! I like your regression test.

tiran:
master:

* 93b7c40158313b9f0f125706265ea0ac07fb4c0e Enable and start oddjobd after ipa-restore if it's not running.

PR #1477, opened 2018-01-17
preventing ldap principal to be deleted

akokshar:
ipa-server-install --uninstall command is calling server-del to
delete replica. This scenario does not work since server-del
is also deleting all principals from and ldap breaking ldap
replication. As a result, only part of deletions are propagated
to the other replicals leaving a lot of orphaned data there.

https://pagure.io/freeipa/issue/7371

### DIFF
--- a/ipaserver/plugins/server.py
+++ b/ipaserver/plugins/server.py
@@ -659,10 +659,26 @@ class server_del(LDAPDelete):
         delete server kerberos key and all its svc principals
         """
         try:
+            # do not delete ldap principal if server-del command
+            # has been called on a machine which is being deleted
+            # since this will break replication.
+            # ldap principal to be cleaned later by topology plugin
+            # necessary changes to a topology plugin are tracked
+            # under https://pagure.io/freeipa/issue/7359
+            if master == self.api.env.host:
+                filter = (
+                    '(&(krbprincipalname=*/{}@{})'
+                    '(!(krbprincipalname=ldap/*)))'
+                    .format(master, self.api.env.realm)
+                )
+            else:
+                filter = '(krbprincipalname=*/{}@{})'.format(
+                    master, self.api.env.realm
+                )
+
             entries = ldap.get_entries(
-                self.api.env.basedn, ldap.SCOPE_SUBTREE,
-                filter='(krbprincipalname=*/{}@{})'.format(
-                    master, self.api.env.realm))
+                self.api.env.basedn, ldap.SCOPE_SUBTREE, filter=filter
+            )
 
             if entries:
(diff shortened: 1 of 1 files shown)


ipaserver/plugins/server.py
```diff
@@ -659,10 +659,24 @@ def _remove_server_host_services(self, ldap, master):
         delete server kerberos key and all its svc principals
         """
         try:
-            entries = ldap.get_entries(
-                self.api.env.basedn, ldap.SCOPE_SUBTREE,
+            # do not delete ldap principal if server-del command 
+            # has been called on a machine which is being deleted
+            # since this will break replication.
+            # ldap principal to be cleaned later by topology plugin
+            if master == self. api.env.host:
```
abbra: Please remove space between `self.` and `api`.

ipaserver/plugins/server.py
```diff
@@ -659,10 +659,24 @@ def _remove_server_host_services(self, ldap, master):
...
+                filter=(
```
abbra: In this filter, it is better to swap internal filters so that we first reduce the search space to the set of principals from the same host and then eliminate `ldap/..` principal from the list.

akokshar: done

abbra:
Build failed:
```
Sorry: IndentationError: unexpected indent (server.py, line 670)
```

akokshar:
looks like my hands.dll is corrupted :(

pvoborni:
I'd mention in the comment that deletion of ldap principal by topology plugin is still not implemented and a subject of https://pagure.io/freeipa/issue/7359. To let the reader know to not expect that behavior until it is implemented. Without mentioning it, somebody might open a bug that it doesn't behave as described.

If acked then add comment to ticket https://pagure.io/freeipa/issue/7359 to remove this comment when 7359 is implemented.

pvoborni:
The commit message should also mention that it won't fully fix the issue with left-over data but by this change, more data will be cleaned and only ldap principal will be left thus ending in a better state.

pvoborni:
Btw, I won't be doing a functional review. But wanted to check the soft aspects. 

Thanks for ammending the commit message but still I'd add the part that deleting LDAP principal by topology plugin is not implemented also in the code comment as I suggested in the first review comment so that it is clear from reading the code.

akokshar:
@abbra, could you please review this PR?

abbra:
LGTM.

tiran:
master:

* 1235f5958dd61e6f99191e512953acdf8b577fff preventing ldap principal to be deleted

PR #1479, opened 2018-01-17
Fixing WebUI Tests

felipevolpone:


ipatests/test_webui/test_hbac.py
```diff
@@ -138,29 +138,30 @@ def test_hbac_test(self):
...
         self.switch_to_facet('run_test')
+        self.wait_for_request(n=2)
         self.button_click('new_test')
         self.assert_facet('hbactest', 'user')
 
         # test pre-run validation and navigation to related facet
-        self.switch_to_facet('run_test')
-        self.button_click('run_test')
-        self.assert_dialog('message_dialog')
+        def __hbac_run_test(self):
```
pvoborni: I wonder if there can be a better name for the method. It switches to a tab and clicks on 'run_tests'. But it sounds like it was running some UI test.

rcritten:
Just some nits:

Patch 1. The commit message ends "it will". It will what?

Patch 2: Acording -> According
             Include a link to the geckodriver change?

Patch 6: Can you add a link to the ticket or commit that changed this behavior? I thought we had tracked that down.

felipevolpone:
As @pvoborni commented [here](https://github.com/freeipa/freeipa/pull/1479/commits/866a5551a22aa91a33fb6c87938b2842d3105fbb#r162281081), the `logout()` call is actually a workaround for something very weird that's happening in the `test_user.py::test_user::test_password_expiration_notification` case and that I cannot reproduce manually.

In the `test_password_expiration_notification` method, in the `#re-login as new user`, the `init_app` is failing because after calling `self.load()` (inside the init_app method), the application goes back to the user page logged as admin!

Debugging it:
```bash
test_webui/test_user.py::test_user::test_password_expiration_notification <- test_webui/ui_driver.py
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /usr/lib/python2.7/site-packages/ipatests/test_webui/test_user.py(314)test_password_expiration_notification()
-> self.logout()
(Pdb) self.driver.get_screenshot_as_file('logout.png')
True
(Pdb) l
309             self.navigate_to_record(user.PKEY, entity=user.ENTITY)
310             self.reset_password_action(pwd)
311
312             #re-login as new user
313             import pdb; pdb.set_trace()
314  ->         self.logout()
315             self.init_app(user.PKEY, pwd)
316
317             header = self.find('.navbar-pf', By.CSS_SELECTOR)
318             self.assert_text(
319                 '.header-passwordexpires',
(Pdb) self.driver.get_screenshot_as_file('logout.png')
True
(Pdb) n
> /usr/lib/python2.7/site-packages/ipatests/test_webui/test_user.py(315)test_password_expiration_notification()
-> self.init_app(user.PKEY, pwd)
(Pdb) self.driver.get_screenshot_as_file('after_logout.png')
True
(Pdb) s
--Call--
> /usr/lib/python2.7/site-packages/ipatests/test_webui/ui_driver.py(342)init_app()
-> def init_app(self, login=None, password=None):
(Pdb) n
> /usr/lib/python2.7/site-packages/ipatests/test_webui/ui_driver.py(346)init_app()
-> self.load()
(Pdb) args
self = <ipatests.test_webui.test_user.test_user object at 0x7ff2b22baf50>
login = itest-user
password = Secret.123
(Pdb) n
> /usr/lib/python2.7/site-packages/ipatests/test_webui/ui_driver.py(347)init_app()
-> self.wait(0.5)
(Pdb) self.driver.get_screenshot_as_file('after_load.png')
True
(Pdb) n
> /usr/lib/python2.7/site-packages/ipatests/test_webui/ui_driver.py(348)init_app()
-> self.login(login, password)
(Pdb) self.driver.get_screenshot_as_file('after_wait.png')
True
(Pdb) l
343             """
344             Load and login
345             """
346             self.load()
347             self.wait(0.5)
348  ->         self.login(login, password)
349             # metadata + default page
350             self.wait_for_request(n=5)
351
352         def load(self):
353             """
(Pdb) c
FAILED

=================================================================================================================== FAILURES ===================================================================================================================
_______________________________________________________________________________________________ test_user.test_password_expiration_notification ________________________________________________________________________________________________

args = (<ipatests.test_webui.test_user.test_user object at 0x7ff2b22baf50>,), self = <ipatests.test_webui.test_user.test_user object at 0x7ff2b22baf50>, name = '2018-01-18T20:49:05.272877_test_user_test_password_expiration_notification'

    @wraps(fn)
    def screenshot_wrapper(*args):
        try:
>           return fn(*args)

test_webui/ui_driver.py:99:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
test_webui/test_user.py:321: in test_password_expiration_notification
    header)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <ipatests.test_webui.test_user.test_user object at 0x7ff2b22baf50>, selector = '.header-passwordexpires', value = 'Your password expires in 6 days.'
parent = <selenium.webdriver.firefox.webelement.FirefoxWebElement (session="5cc2384f-4a55-4061-96bb-d02aaef19b1d", element="34586936-6f64-437a-b006-8de3e065eed1")>

    def assert_text(self, selector, value, parent=None):
        """
            Assert read-only text value in details page or in a form
            """
        text = self.get_text(selector, parent)
        text = text.strip()
        value = value.strip()
>       assert text == value, "Invalid value: '%s' Expected: %s" % (text, value)
E       AssertionError: Invalid value: '' Expected: Your password expires in 6 days.

test_webui/ui_driver.py:1595: AssertionError
```

The images:
logout.png
logout

after_logout.png
after_logout

after_load.png
after_load

after_wait.png
after_wait

pvoborni:
I see bug in:
```python
    def get_login_screen(self):
        """
        Get reference of login screen
        """
        return self.find('rcue-login-screen', 'id')
```

`'rcue-login-screen` doesn't exist. Therefore self.logged_in() might be buggy. Not sure yet if it might be a cause of that behaviour. Doesn't seem so. It might be replace by searching for login facet or by search for element with class `login-pf`. 

Would be interesting to see how login() behaves - e.g. if it fills anything in the first attempt of the test and not only at the cleanup phase.

pvoborni:
@felipevolpone do you know how is `logged_in()` evaluated in the `self.init_app(user.PKEY, pwd)` call.  Because, the effect, your patch did is that it logs in with a supplied password every time regardless of evaluation of this call. Before it, it may have skipped the login procedure if `logged_in()` was evaluated as true.  

I'd also advise avoiding using breakpoints in debugging as it may change behavior because of time aspects.  One theory could be that logout() only does clicking on `logout` button but it doesn't care about next step so there is missing any verification that logout was successful.  It then calls `self.init_app(user.PKEY, pwd)` which seems OK, but there is one step which is not good and that is a `self.driver.get(self.get_base_url())` call in `load`method. This doesn't simulate user behavior as the user would not enter new address in address bar.

felipevolpone:
Sorry for the long silence, but I was finally able to figure it out the logout problem and hopefully also addressed your requests. 

Now, only the vault test is failing, which is related to ticket [1]. There are some KRA tests being skipped, I already submitted a fix to PR CI to make them run.

[1] https://pagure.io/freeipa/issue/7373

felipevolpone:
Tests are green :) Once we agree in this PR, I will fix the pylint errors

ipatests/test_webui/test_navigation.py
```diff
@@ -106,12 +106,11 @@ def test_menu_navigation(self):
 
         # Identity
         # don't start by users (default)
-        self.navigate_by_menu('identity/group', False)
+        self.navigate_by_menu('identity/group_search', False)
         self.navigate_by_menu('identity/user_search', False)
         self.navigate_by_menu('identity/host', False)
-        self.navigate_by_menu('identity/hostgroup', False)
```
pvoborni: I'm not sure if removing the hostgroup completely is the right thing. The test failure was caused by: 070bc48dd6c9bce32caa0f0f2de8d44b4e5bbbb1 We might want to do something like:

```
self.navigate_by_menu('identity/group_search', False)
self.navigate_by_sidebar('Host Groups', False)
self.navigate_by_sidebar('Netgroups, False)
```
I'm not sure if we a methog like `navigate_by_sidebar` implemented though.

felipevolpone: good catch. We don't have `navigate_by_sidebar` but we do have `switch_to_facet` that looks for a tab and if it doesn't exist, it looks for a link in the sidebar.

pvoborni:
I did some changes described below in code they are at https://github.com/pvoborni/freeipa/pull/3

## WebUI Tests: fixing test_hbac

* ACK
## WebUI Tests: fixing test_group
Why is following removed?
```
-                ('radio', 'type', 'posix'),
-                ('callback', self.check_posix_enabled, True),
```

It tests if the text field is properly enabled when switching between various types. In this case for nonposix to posix. 

## WebUI Tests: fixing test_navigation 
As written in the previous comment we need to bring back testing of equivalent.

```
-        self.navigate_by_menu('identity/hostgroup', False)
-        self.navigate_by_menu('identity/netgroup', False)
```

## WebUI Tests: refactoring login method to be more readable

* ACK

## WebUI Tests: changing how the login screen is detected 

* ACK

## WebUI Tests: fixing test_range test case 

* I'd improve comments in the file, it is hard to read it. Also, usage of present continuous tense in comments doesn't seem right to me. 

## WebUI Tests: fixing test_user.py::test_test_noprivate_posix
I still don't understand why the 2s wait helps here. 2s is quite a lot. When looking at the code
I don't really agree with commit 6468a39850 which introduced this behavior of compobox but changing it is difficult (task for separate PR). It doesn't reflect what user would do.

The code does:
- click on "open" button
- fill textbox input
- switch to active element and press enter

It should do:
- fill a textbox

Or alternatively:
- click on "open" button to open the list and see that what we are looking for is not there
- fill a textbox
- click on "open" button to close it as it is in a way, thus no pressing of enter would happen
(easy change)

Thus wonder if replacing the wait(2) with `open_btn.click()` would solve it.


## WebUI Tests: Changing how the initial load process is done
* I'm missing an information why we are trying to click on the ipa_logo element. Or why the block is there.

E.g. there could be a comment: If a page is already loaded, e.g. in the case where we logged out,
do not issue GET request from the address bar. Thus allow to test behaviors like login->logout->login->...

## WebUI Tests: changing the ActionsChains.move_to_element to a new appr…
* ACK provided the fix in the previous patch

## Temporary commit: Adding WebUI Tests to PRCI
Will be dismissed, OK.

pvoborni:
I approve the changes. So let's fix the pylint errors, bring back PR-CI check and I think we will be ready for a merge.

pvoborni:
the temporary commit  still needs removal

felipevolpone:
@pvoborni Temporary commit removed. I summarized what is said in the link on commit  `
WebUI Tests: changing the ActionsChains.move_to_element to a new...` in the commit message as well.

tiran:
master:

* 83ed8d279210766d3d068a2a6daa0f8368c937e4 WebUI Tests: fixing test_hbac
* dae5bac39bb2cbea1219b320d05d2171f5f86e63 WebUI Tests: fixing test_group
* 3fa4378bc4f8b1d01c3f9844d605c174d0aa815f WebUI Tests: fixing test_navigation
* 7c3f9b79eb42252b4540e26267e5aa229343f392 WebUI Tests: refactoring login method to be more readable
* 49a17e98b0eb62636f4c4f0f43218a36fcea383d WebUI Tests: changing how the login screen is detected
* 12da43c54fb927b48677fcca50f5110c5e659b1e WebUI Tests: fixing test_range test case
* a072fe9718a4273aea0363667c728e6861c3f3b7 WebUI Tests: Changing how the initial load process is done
* 81fb7e5a321e2f4fa0e4112bea073c30dbe9d54e WebUI Tests: fixing test_user.py::test_test_noprivate_posix
* a5bd7bf7668092d9b329e640b656f23b8bf7bfe6 WebUI Tests: changing the ActionsChains.move_to_element to a new approach

PR #1480, opened 2018-01-18
IntegrationTests now collects logs from all test methods

felipevolpone:
`logs_dict` should not be cleared. It's filled once per class and it
should not be cleared after running the first test.

https://pagure.io/freeipa/issue/7310
https://pagure.io/freeipa/issue/7335

### DIFF
--- a/ipatests/pytest_plugins/integration/__init__.py
+++ b/ipatests/pytest_plugins/integration/__init__.py
@@ -179,8 +179,6 @@ def collect_logs(name, logs_dict, logfile_dir=None, beakerlib_plugin=None):
             else:
                 shutil.rmtree(topdirname)
 
-        logs_dict.clear()
-
 
 @pytest.fixture(scope='class')
 def class_integration_logs():


Rezney:
Thanks for this PR. It seems to be working fine, however I have two questions:

1. When I run an integration test in "testdir" I see a directory for every test_func but also a directory which looks like called after the class:

e.g

10:40 -TestSimpleReplication
10:39 -TestSimpleReplication--test_user_replication_to_replica

and

10:14 -TestSimpleReplication
10:12 -TestSimpleReplication--test_replica_removal
10:10 -TestSimpleReplication--test_user_replication_to_master
10:08 -TestSimpleReplication--test_user_replication_to_replica

Could you please elaborate on how/why it gets created. It looks like it is created always at the end.

2. Any idea why the "logs_dict.clear()" was there at the end? It still kind of troubles me that it would be there just for nothing.

Thank you in advance for casting light on this and sorry if you partially answered some of that in the previous PR and I missed that.

felipevolpone:
> Could you please elaborate on how/why it gets created. It looks like it is created always at the end.

The dir with the name of the class is created because of the `mh` decorator. It calls `collect_test_logs` in its last line, and once per class `request.node` (inside `collect_test_logs`) is the name of the test class. In other cases, `request.node` will be the name of the test method.

> Any idea why the "logs_dict.clear()" was there at the end? It still kind of troubles me that it would be there just for nothing.

To be honest, I don't know. I checked the commit messages and they don't tell anything about it. What I can suppose is that sometime in the past the decorators were running at different times. Now the `mh` decorator runs at the "beginning" of the class, it could be possible that sometime in the past it was running with `scope=method` or something like that. So, the results would be different. But, as I said, I'm just supposing.

encukou:
I think it was there to allow collecting logs in multiple passes, i.e. calling `collect_logs` in the middle of a test.
AFAIK that is not useful, and hopefully it's not used. Remove the `clear` if it works better.

Rezney:
@encukou Thanks for you input!

PR #1486, opened 2018-01-23
Fix pylint error in ipapython/dn.py

tiran:
```
ipapython/dn.py:1324: [R1710(inconsistent-return-statements), DN.__contains__]
Either all return statements in a function should return an expression, or none of them should.)
```


abbra:
LGTM.

tiran:
master:

* 7924dae6aed2ba1ba762464a21b5b4e7ac1c135b Fix pylint error in ipapython/dn.py

PR #1490, opened 2018-01-23
Give ODS socket a bit of time

tiran:
ipa-ods-exporter uses systemd socket activation. The script uses
select() to check if the socket is readable. A timeout of 0 is a bit too
aggressive. Sometimes select() doesn't consider the systemd socket as
readable. This causes ODS to fail silently

A timeout of one second seems to remove the problem. A proper error code
also signals that something went wrong.

Closes: https://pagure.io/freeipa/issue/7378

### DIFF
--- a/daemons/dnssec/ipa-ods-exporter
+++ b/daemons/dnssec/ipa-ods-exporter
@@ -449,17 +449,21 @@ def hex_set(s):
         out.add("0x%s" % hexlify(i))
     return out
 
+
 def receive_systemd_command():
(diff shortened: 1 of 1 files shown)


tiran:
The fix originates from PR https://github.com/freeipa/freeipa/pull/1289

tiran:
master:

* e1e3218270a008c2ebc0e74cbd54a8a218c3a80b Give ODS socket a bit of time

PR #1493, opened 2018-01-24
Fixing translation problems

slaykovsky:
ipa rpc server did set the LANG environment variable on each
request and it was not thread safe which led to the problems
which are described in the pagure ticket. Also, there were
mistakes regarding setting the Accept-Language HTTP header.

Now on each request we're setting the "languages" property
in the context thread local variable and client is setting the
Accept-Language HTTP header correctly.

Also, as the server is caching the schema and the schema can be
generated for several languages it's good to store different
schema fingerprint for each language separately.

pagure: https://pagure.io/freeipa/issue/7238

### DIFF
--- a/ipaclient/remote_plugins/__init__.py
+++ b/ipaclient/remote_plugins/__init__.py
@@ -30,9 +30,9 @@ class ServerInfo(collections.MutableMapping):
 
         # copy-paste from ipalib/rpc.py
         try:
-            self._language = (
-                 locale.setlocale(locale.LC_ALL, '').split('.')[0].lower()
-            )
+            self._language = locale.setlocale(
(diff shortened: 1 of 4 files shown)


stlaz:
Please, when writing a commit message, don't refer to a pagure ticket in the description of the fix. Pagure may seem like a persistent storage of information now but it does not necessarily have to be in the future. Just make a clear statement of what the failures caused by the code were.

slaykovsky:
@stlaz thanks!

Tiboris:
Thanks @slaykovsky, works well!

Tiboris:
master:

* 6c5a7464b26eaf5af890af3b0433bbb6e7bc5f3e Fixing translation problems

PR #1495, opened 2018-01-24
[Backport][ipa-4-6] Fixing vault-add-member to be compatible with py3

felipevolpone:
Changing from itervalues() to values() in order to be compatible with python3.

https://pagure.io/freeipa/issue/7373


tiran:
ipa-4-6:

* cf6aeae155d9121013c4189c84dd7146a1c47c1d Fixing vault-add-member to be compatible with py3

PR #1503, opened 2018-01-30
Update IPA CA issuer DN upon renewal

frasertweedale:
When renewing externally-signed CA or when switching from
externally-signed to self-signed CA, the Issuer DN can change.
Update the ipaCaIssuerDn field of the IPA CA entry upon renewal, to
keep it in sync.

Fixes: https://pagure.io/freeipa/issue/7316

### DIFF
--- a/ipaserver/install/ipa_cacert_manage.py
+++ b/ipaserver/install/ipa_cacert_manage.py
@@ -24,7 +24,7 @@ import os
 from optparse import OptionGroup  # pylint: disable=deprecated-module
 import gssapi
 
-from ipalib.constants import RENEWAL_CA_NAME, RENEWAL_REUSE_CA_NAME
+from ipalib.constants import RENEWAL_CA_NAME, RENEWAL_REUSE_CA_NAME, IPA_CA_CN
 from ipalib.install import certmonger, certstore
 from ipapython import admintool, ipautil
 from ipapython.certdb import (EMPTY_TRUST_FLAGS,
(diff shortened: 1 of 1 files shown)


tiran:
* Is it sufficient to update just ``ipaCaIssuerDN`` or should we also update ``ipaCaSubjectDN``?
* Who is taking care of entries in ``cn=CAcert,cn=ipa,cn=etc,$SUFFIX`` and ``cn=certificates,cn=ipa,cn=etc,$SUFFIX``?

frasertweedale:
@tiran thx for feedback.

We shouldn't need to update `ipaCaSubjectDN` because changing the CA subject DN should
be impossible and will certainly cause breakage.  If it is even possible, it implies a missing
sanity check (and would be dealt with as a separate PR).

For updating `cn=certificates,...` et al, this should already be happening as part of `ipa-cacert-manage`. If not, we should address as a separate ticket/PR.  The purpose of this PR is solely to fix the issue where issuer DN in the IPA `ipaca` entry from getting out of sync with the actual CA cert.

Cheers!

tiran:
Thanks for clarification! Your patch looks good and solves the documented issue.

frasertweedale:
Thanks @tiran.  Always good to have the probing questions :)

tiran:
@frasertweedale I have to drill deep to get to you, mate :)

@Tiboris Do you want to give the PR another review before I merge it?

Tiboris:
Yes patch looks good, @tiran  I wanted to try it but got stuck on something else.

tiran:
master:

* b07937d0b80c8ccc714ea62fafcc7090bbaecc23 Update IPA CA issuer DN upon renewal

PR #1505, opened 2018-01-30
Remove special characters in host_add random OTP generation

johnclarson:
Remove special characters in host_add random OTP generation

This fixes a regression in how random host-add OTP passwords are generated. Some shells try and interpret certain special characters when they confront them in unattended ipa-client-install where the OTP is used. Before 4.5.0, this was fixed by excluding an arbitrary list of special characters during OTP random password generation. This fix goes a step further and removes ALL special characters when the OTP is generated. In my opinion, the period of time between random OTP generation and use is small and the password is useless after the host is installed so removing special characters presents minimal security risk

https://pagure.io/freeipa/issue/7380

### DIFF
--- a/ipaserver/plugins/host.py
+++ b/ipaserver/plugins/host.py
@@ -686,7 +686,7 @@ class host_add(LDAPCreate):
                 entry_attrs['objectclass'].remove('krbprincipal')
         if options.get('random'):
             entry_attrs['userpassword'] = ipa_generate_password(
-                entropy_bits=TMP_PWD_ENTROPY_BITS)
+                entropy_bits=TMP_PWD_ENTROPY_BITS, special=None)
             # save the password so it can be displayed in post_callback
             setattr(context, 'randompassword', entry_attrs['userpassword'])
 


Tiboris:
Please edit title to be short and specific.
Update commit message with `git commit --amend` to have first line title, then the description of patch and at the bottom please add the Pagure ticket which you have created.
Feel free to take look on any PR for structure eg.: https://github.com/freeipa/freeipa/pull/1503
Thanks!

johnclarson:
Thanks for the feedback. I can see how the original was hard to read. I hope it is satisfactory now.

rcritten:
It is not safe to assume there is little time between when an OTP is generated and when it is used. This is the case in automated systems but OTPs can also be generated and handed out manually, or the automation can be backed up causing deployment to be slow by an indeterminate amount of time.

Granted that the previous revisions of this only used 7 special characters this would still have a positive impact on complexity.

I wonder if deployment systems should simply quote the password.

Tiboris:
Thanks @johnclarson it is more readable now!
But... there is no need for another commit to do such a change.
Please have a look at command `git commit --amend` again. With this command you are able to change last commit you created. Now your branch has 3 commits, and 2 of them are not necessary. You will need to reset git to commit 0de63aaba0a9614c56eca9fdd870a7b17055f31e then edit commit message with command mentioned above, and then you have to force push  these changes into your fork of project with `git push -f` ( or `--force` BUT please be carefull with this command's option in general) They will apear here.

johnclarson:
Thanks @rcritten I have tried to comment out the OTP in our userdata for the ipa-client-install and it does not seem to work for me for "&" characters. I will double check this, but I assumed that was the issue for the previous fix of limiting the special characters. For reference, here is the original bug and fix:

https://bugzilla.redhat.com/show_bug.cgi?id=827392

Not sure what has changed since then I guess.

johnclarson:
@Tiboris, thanks for the tip on git commit --amend. I will try and do as you say.

rcritten:
@johnclarson What enrollment system are you using? I have it in my head Foreman but I don't see a reference anywhere in the ticket.

The character set definitely changed (arguably a regression). There is some discussion among the other devs whether this is acceptable or not.

johnclarson:
@rcritten I am using Foreman with the snippet for freeipa registration included in my userdata. Let me take a closer look at how the OTP is processed and run some tests. I assumed the problem was this regression because it all worked every time before we upgraded to 4.5.0. I will get back to you ASAP.

johnclarson:
@rcritten here is a reference to the Foreman community template:

https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/snippet/freeipa_register.erb

The issue for us is that we are using AWS EC2 instances and there is an open bug with Foreman on that:

http://projects.theforeman.org/issues/17292

To get around that, I have the following in the ipa-client-install section of that snippet:

--------------------------------
# One-time password will be requested at install time. Otherwise, $HOST[OTP] is used as a placeholder value.
#/usr/sbin/ipa-client-install -w '<%= @host.otp || "$HOST[OTP]" %>'  --realm=<%= @host.realm %> -U $freeipa_mkhomedir $freeipa_opts $freeipa_server $freeipa_ssh
IPAOTP="<%= @host.otp %>" # this line isnt detemplated correctly unless we get it from curl
curl <%= foreman_url("user_data")%> | grep ^IPAOTP > /var/log/otp.log
.  /var/log/otp.log
# snip
/usr/sbin/ipa-client-install -w ${IPAOTP} $freeipa_mkhomedir $freeipa_opts --no-dns-sshfp --enable-dns-updates -f -U
-------------------------------------

The random OTP needs to be quoted in two places here, the IPAOTP variable declaration and the ipa-client-install invocation. I have added quotes to the latter and I have tested several builds until I saw a "&" character in the /var/log/otp.log. It seems that fixed the issue. the regression in FreeIPA made me assume that is where the problem was.

I will try and get the changes implemented for people on the Foreman side. It seems this is more a problem with Foreman than with FreeIPA.

Apologies for wasting your time and thanks for the help and thank you @Tiboris for the git help :)

rcritten:
@johnclarson I don't think you wasted my time :-)

You pointed out a change in behavior and I'm a bit on the fence about it.

On the one hand we actually did purposely restrict the character set in the past to avoid exactly this problem.

On the other hand invocations should probably quote things like this just in case.

Still, **we** changed the behavior so I think we have some responsibility in working out a solution that perhaps doesn't require every integrator to change their code as well.

johnclarson:
@rcritten Oops I closed the issue and updated the fix so people on Google can find it. Would you like to re-open it?

rcritten:
Done, thanks for reporting the issue and filing the PR! Hope it was a positive learning experience.

johnclarson:
@rcritten, no...thank YOU and your team. This is a solid product and I'd be honored to help out moving forward (as time permits). You guys (and gals) are awesome.

rcritten:
This looks good just some minor git nits...

Can you re-format the commit message? The format should be:

One line description, < 65 chars in length
<blank>
Multi-line detailed description, < 65 char line length
<blank>
ticket #<foo>

The message in your commit c920872c8a1774513249ba014db2abd3b599dc36 is closest to this, just exceeding the line limits on the description part.

Also, there are 3 commits including a merge commit. Can you rebase against current master and squash this all into a single commit? If you need some git-fu help feel free to ping me directly.

stlaz:
(Actually, it's 50 chars of commit message subject and 72 for the description)

Also, in your commits you're using an email address that's not associated with your Github account, not sure if that's intentional.

johnclarson:
Thanks @rcritten ...sent you an email for help on the git stuff.

rcritten:
master:

* eaa5be3eecb6a54cc4ff6469057901258cf736da Remove special characters in host_add random OTP generation

PR #1506, opened 2018-01-30
Adding more tests to PR CI

felipevolpone:
Adding the tests:
```
test_integration/test_topologies.py
test_integration/test_sudo.py
test_integration/test_kerberos_flags.py
test_integration/test_http_kdc_proxy.py
test_integration/test_forced_client_reenrollment.py
test_integration/test_advise.py
test_integration/test_testconfig.py
test_integration/test_service_permissions.py
test_integration/test_netgroup.py
test_integration/test_vault.py
```

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -23,7 +23,7 @@ jobs:
         git_refspec: '{git_refspec}'
         template: &ci-master-f27
           name: freeipa/ci-master-f27
-          version: 1.0.2
+          version: 1.0.3
         timeout: 1800
         topology: *build
 
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -62,3 +62,123 @@ jobs:
         template: *ci-master-f27
         timeout: 3600
         topology: *master_1repl
+
+  fedora-27/test_topologies:
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_topologies.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl
+
+  fedora-27/test_sudo:
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_sudo.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl_1client
+
+  fedora-27/test_kerberos_flags:
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_kerberos_flags.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl_1client
+
+  fedora-27/test_http_kdc_proxy:
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_http_kdc_proxy.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl_1client
+
+  fedora-27/test_forced_client_enrolment:
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_forced_client_reenrollment.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl_1client
+
+  fedora-27/test_advise:
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_advise.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl
+
+  fedora-27/test_testconfig:
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_testconfig.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl
+
+  fedora-27/test_service_permissions:
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_service_permissions.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl
+
+  fedora-27/test_netgroup:
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_netgroup.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl
+
+  fedora-27/test_vault:
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_vault.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl


stlaz:
I believe it would make much more sense to implement GitHub API caching proxy for PRCI rather than beating its limits with just one PR and adding a timeout to wait for eternity.
My 2c.

felipevolpone:
We already have a cache for the API. The timeouts that I increased are not related to the cache in any way.
They are related to the time that the test needs to run, some of then (like the installation one) can take hours.

stlaz:
Oh I see. I believe one of the original requirements for freeIPA PR CI was that testing of a PR can't exceed 1,5h. I believe that's a reasonable requirement so I'll have to ask you to lower the timeout back.

felipevolpone:
This is the list of the tests that I propose to run in each new PR. They are very stable and run under 1,5h.

.freeipa-pr-ci.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
         git_refspec: '{git_refspec}'
         template: &ci-master-f27
           name: freeipa/ci-master-f27
-          version: 1.0.2
+          version: 1.0.3
```
flo-renaud: General comment: on the [vagrant cloud page](https://app.vagrantup.com/freeipa/boxes/ci-master-f27) the box does not have any description, hence it's difficult to know why a specific version has any added value compared to the preceding one.

.freeipa-pr-ci.yaml
```diff
@@ -62,3 +62,123 @@ jobs:
...
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_http_kdc_proxy.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl_1client
```
flo-renaud: This test needs only 1 client and 0 replica

.freeipa-pr-ci.yaml
```diff
@@ -62,3 +62,123 @@ jobs:
...
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_kerberos_flags.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl_1client
```
flo-renaud: This test needs only 1 client and 0 replica.

.freeipa-pr-ci.yaml
```diff
@@ -62,3 +62,123 @@ jobs:
...
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_sudo.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl_1client
```
flo-renaud: This test needs only 1 client and 0 replica.

.freeipa-pr-ci.yaml
```diff
@@ -62,3 +62,123 @@ jobs:
...
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_topologies.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl
```
flo-renaud: It seems to me that the test_topologies test does not require any installation, so you could use a minimal topology with only one machine.

.freeipa-pr-ci.yaml
```diff
@@ -62,3 +62,123 @@ jobs:
...
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_advise.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl
```
flo-renaud: This test does not need any replica.

.freeipa-pr-ci.yaml
```diff
@@ -62,3 +62,123 @@ jobs:
...
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_testconfig.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl
```
flo-renaud: This test does not need any replica.

.freeipa-pr-ci.yaml
```diff
@@ -62,3 +62,123 @@ jobs:
...
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_service_permissions.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl
```
flo-renaud: This test does not need any replica.

.freeipa-pr-ci.yaml
```diff
@@ -62,3 +62,123 @@ jobs:
...
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_netgroup.py
+        template: *ci-master-f27
+        timeout: 3600
+        topology: *master_1repl
```
flo-renaud: This test does not need any replica.

felipevolpone:
@flo-renaud  raised an important subject here and I haven't explained properly, so here we go.
If we would have a different topology section (and a different runner size) for each topology our tests require, we would have something like:

only master
1vCPU/0.9 GB RAM controller
1vCPU/2.4 GB RAM master
total: 2vCPU/4 GB RAM

master + 1 client
1vCPU/0.9 GB RAM controller
1vCPU/2.4 GB RAM master
1vCPU/2.4 GB RAM client
total: 3vCPU/5.7 GB RAM

master + 1 repl
1vCPU/0.9 GB RAM controller
1vCPU/2.4 GB RAM master
1vCPU/2.4 GB RAM replica
total: 3vCPU/5.7 GB RAM

In the first case, we could create a runner that requires 2vCPU and 4 GB RAM and we would be fine. However, in the other cases, we don't have a template in OpenStack that provides a machine with ~6GB RAM and 3vCPU. The closest option is 8GB and 4GB (which also covers the case `1master_1replica_1client`). That's why I haven't created new topologies (in the `topologies` section).

So let's say we will create a runner with only `2vCPU and 4 GB RAM`.  On one hand that would be good, because we would use our resources in the best way possible, but on the other hand, if there are no tests of that size to run, the small runner would just sit and wait. 

If we don't create such kind of runner, a bigger (the `1master_1replica_1client` let's say) one would be able to run it, but it would not use the resources in the best way possible. Although, it would be able to run bigger tests. So, it's kind of a balance.

---
> What were your considerations when picking the new tests? Stability and duration only?

Yes, stability and duration. I put these tests to run like 10 or 15 times and they are stable. So that's the reason. 

> I would consider adding tests_integrations/test_upgrade.py (obviously only if the test is stable) as the upgrade proved to be a pain point.

Indeed. Can we go with this PR as it is and then I'll create another one to run tests_integrations/test_upgrade.py lot of times to we track if it is stable?

.freeipa-pr-ci.yaml
felipevolpone: Please see my comment below.

pvoborni:
We should explore a way to add OS flavors to fit our needs. But till we have the support of different sized of runners it doesn't matter much.  

But there is one other aspect which goes in favor of adding the templates. If I'm not wrong, vagrant will spawn a vm for all vms defined in the template. This takes time. If a test doesn't require a replica or a client then we can save some time by not launching them. 

I guess we can push this if it is stable. But we should optimize it a bit.

.freeipa-pr-ci.yaml
flo-renaud: Thanks, the description makes sense.

felipevolpone:
> We should explore a way to add OS flavors to fit our needs. But till we have the support of different sized of runners it doesn't matter much.

JFTR, As was talked on IRC, PR CI does support different sizes or runners.

flo-renaud:
I agree with @pvoborni we can push as it is today. But we should open a ticket to remember to optimize the topologies in a later PR.

slaykovsky:
Hi! I think it's not the right way to define topologies in the YAML config and define numerous Vagrant files one similar to another as topology information is already available inside tests.

felipevolpone:
> Hi! I think it's not the right way to define topologies in the YAML config and define numerous Vagrant files one similar to another as topology information is already available inside tests.

Hi @slaykovsky, do you have some proposal or alternative to that? Why do think it's that bad? We map it on the `.freeipa-pr-ci.yaml` just because was easy to develop at that time. An alternative could be: inspect the test file and discover the needed topology, but then I'm not sure how the runner would know which test it can run.

Either way, I don't see how this change would bring any value to the project at this moment. There are tons of other things to do (like fix all the broken tests, support run tests on RHEL, etc :)). It's more like something nice to have, I guess. What do you think?

slaykovsky:
Hi @felipevolpone 

As we need to have tests here which should be defined by our customers' needs it's very good to have a YAML which defines which tests we need to run. I talked about:
>inspect the test file and discover the needed topology

this. The needed amount of CPU and RAM can be calculated from the information which already exists in the tests.

Just the one example of the value: we will not have skipped tests because there are no resources configured.

Thanks.

felipevolpone:
> this. The needed amount of CPU and RAM can be calculated from the information which already exists in the tests.

indeed. 

> Just the one example of the value: we will not have skipped tests because there are no resources configured.

That would be a good benefit for sure. Let me try to rephrase myself here: do you think that it worth to spend time changing how we map the resources to this new way? 

Even though I see value in this new way, I don't think it worth. How it's implemented now may not be the best approach, but it works for its purpose. The tests being skipped due to lack of resources is mostly my fault of not paying attention to the tests needs when configuring our nightly PRs. However, once we have [nightly PRs definition in source tree](https://projects.engineering.redhat.com/browse/FREEIPA-868), we would probably get rid of this problem, because we would have more people looking at it (i.e: reviewing changes).

slaykovsky:
Well, I'm not the one who makes decisions :)

Just giving my thoughts here, because if we can automate something it's almost always worth to do so :)

felipevolpone:
> Well, I'm not the one who makes decisions :)

Well, me neither :)

stlaz:
There were commits + discussion after the ACK, removing the label.

flo-renaud:
IMHO, the goal of this PR is to add more tests to our PR-CI, so that we are confident when we perform code changes. @felipevolpone 's code changes address this goal.

The mechanism that PR-CI is using to pick a runner is not addressed in this PR, and if we want to make enhancements like automatic evaluation of the runner requirements, it is a really good idea that should be tracked in a separate ticket and in a separate PR. Let's not prevent/delay more tests just because we want a nicer mechanism.
My 2 cents...

felipevolpone:
I agree with @flo-renaud. IMHO the discussion does not stop us to have an ack here. Do you agree @slaykovsky ?

pvoborni:
I agree with @flo-renaud completely.

slaykovsky:
@flo-renaud @felipevolpone @pvoborni Hi! I'm not against any ACKs here because of the discussion.

As @stlaz wrote:
>There were **commits** + discussion after the ACK, removing the label

It's because there were more changes.

Thanks.

felipevolpone:
I rebased the commit to have tests run once again. So, leaving the final ack to somebody else than me.

flo-renaud:
Then I take full responsibility for this ACK :)

tiran:
master:

* 5ef49ffcbcc855239ed4e6c8ee9b362a6d7b56a0 Adding more tests to PR CI
* 9a6a90bbdfee7fac5426c898f03fa163e24e2598 prci: Bump ci-master-f27 template to 1.0.3

PR #1507, opened 2018-01-31
ipa-server-upgrade now checks custodia server keys

tiran:
The ipa-server-upgrade command now checks for presence of ipa-custodia's
config and server keys. In case any of the files is missing, it
re-creates both files.

Partly resolves https://pagure.io/freeipa/issue/6893. The upgrader does
not auto-detect broken or mismatching keys yet.

### DIFF
--- a/ipaserver/install/custodiainstance.py
+++ b/ipaserver/install/custodiainstance.py
@@ -92,7 +92,17 @@ class CustodiaInstance(SimpleServiceInstance):
         keystore.generate_server_keys()
 
     def upgrade_instance(self):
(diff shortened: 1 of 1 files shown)


frasertweedale:
Tested by deleting they keys, running `ipa-server-upgrade`, then creating LWCAs to
confirm that key replication works.  It does work, therefore ACK.

tiran:
@frasertweedale thanks for your review!

tiran:
master:

* 387ae9fd0f0afeecffb41ff8ffd6835ae66ea8ff ipa-server-upgrade now checks custodia server keys

PR #1514, opened 2018-01-31
Move Requires: pythonX-sssdconfig into conditional

rcritten:
With this change it looks like the client can be installed in python3-only.

https://pagure.io/freeipa/issue/5638


tiran:
master:

* aaf2eaabee889245fcb6c6d5a8fad1bf0a261033 Move Requires: pythonX-sssdconfig into conditional

PR #1516, opened 2018-02-02
ipa_tests: test subca key replication

Rezney:
Test if key replication is not failing.

https://pagure.io/freeipa/issue/7387

### DIFF
--- a/ipatests/test_integration/test_replica_promotion.py
+++ b/ipatests/test_integration/test_replica_promotion.py
@@ -11,9 +11,9 @@ from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_plugins.integration import tasks
 from ipatests.pytest_plugins.integration.tasks import (
     assert_error, replicas_cleanup)
-from ipalib.constants import DOMAIN_LEVEL_0
-from ipalib.constants import DOMAIN_LEVEL_1
-from ipalib.constants import DOMAIN_SUFFIX_NAME
+from ipalib.constants import (
+    DOMAIN_LEVEL_0, DOMAIN_LEVEL_1, DOMAIN_SUFFIX_NAME, IPA_CA_NICKNAME)
+from ipaplatform.paths import paths
 
 
 class ReplicaPromotionBase(IntegrationTest):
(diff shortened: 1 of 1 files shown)


Rezney:
Wondering whether we should create a new e.g. "test_certificates.py" module in order to not "pollute" "test_replica_promotion.py" with other sort of tests. Any comments?

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -570,3 +570,50 @@ def test_replica_install_with_existing_entry(self):
...
+
+    def test_sub_ca_key_replication(self):
+        master = self.master
+        replica = self.replicas[0]
+
+        result = master.run_command(['ipa', 'ca-add', self.SUBCA, '--subject',
+                                     self.SUBCA_CN])
+        assert result.returncode == 0
+
+        uuid = '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
```
tiran: Traditionally, regular expression constants and compiled re objects are suffixed with ``_re``, e.g. ``auth_id_re``. It's a bit confusing to have two variables with same name ``auth_id`` but different meaning.

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -570,3 +570,50 @@ def test_replica_install_with_existing_entry(self):
...
+        auth_id = re.compile('Authority ID: ({})'.format(uuid), re.IGNORECASE)
+        auth_id = "".join(re.findall(auth_id, result.stdout_text))
+
+        cert_nick = '{} {}'.format(IPA_CA_NICKNAME, auth_id)
```
tiran: Bonus points for new format operator 👍

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -570,3 +570,50 @@ def test_replica_install_with_existing_entry(self):
...
+
+        # give replication some time
+        time.sleep(30)
+
+        result = replica.run_command(['ipa-certupdate'])
+        assert result.returncode == 0
```
tiran: You don't need to clutter the code with ``assert result.returncode == 0``. pytest multihost's ``run_command`` method defaults to ``raiseonerr=True``, https://pagure.io/python-pytest-multihost/blob/master/f/pytest_multihost/host.py#_202

ipatests/test_integration/test_replica_promotion.py
Rezney: yep, makes sense...

ipatests/test_integration/test_replica_promotion.py
Rezney: keep forgetting about this or just like "asserts" :)

tiran:
master:

* e5a508a749364f6a8916676061185c62972ea4a9 ipa_tests: test subca key replication

PR #1517, opened 2018-02-04
Fix detection of KRA installation so upgrades can succeed

rcritten:
To determine whether KRA needs to be updated the standard
service.is_configured() was used. Since the KRA and CA share
the same service name (pki-tomcatd) this wasn't being detected
properly.

Add an additional check to see if the KRA CS.cfg is present.

https://pagure.io/freeipa/issue/7389

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -1710,7 +1710,7 @@ def upgrade_configuration():
             )
         upgrade_pki(ca, fstore)
 
-        if kra.is_configured():
+        if kra.is_installed():
             logger.info('[Ensuring ephemeralRequest is enabled in KRA]')
             kra.backup_config()
             value = installutils.get_directive(
@@ -1728,7 +1728,7 @@ def upgrade_configuration():
     # by checking status using http
     if ca.is_configured():
         ca.start('pki-tomcat')
-    if kra.is_configured() and not kra.is_running():
+    if kra.is_installed() and not kra.is_running():
         # This is for future-proofing in case the KRA is ever standalone.
         kra.start('pki-tomcat')
 
--- a/ipatests/test_integration/test_upgrade.py
+++ b/ipatests/test_integration/test_upgrade.py
@@ -1,0 +1,21 @@
+#
+# Copyright (C) 2018  FreeIPA Contributors see COPYING for license
+#
+
+"""
+Module provides tests to verify that the upgrade script works.
+"""
+
+from ipatests.test_integration.base import IntegrationTest
+from ipatests.pytest_plugins.integration import tasks
+
+
+class TestUpgrade(IntegrationTest):
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master, setup_dns=False)
+
+    def test_invoke_upgrader(self):
+        cmd = self.master.run_command(['ipa-server-upgrade'],
+                                      raiseonerr=False)
+        assert cmd.returncode == 0


stlaz:
I recently noticed this failure, too. When did it start to happen?

flo-renaud:
Hi @rcritten 
you could also replace the call to kra.is_configured() with kra.is_installed() since is_installed properly checks for the subsystem.
I quickly checked the tests and we currently have no test at all using ipa-server-upgrade. This fix would be a good opportunity to enhance the test suite and add a simple test calling install + ipa-server-upgrade.

stlaz:
Hm, must have been there from the start of the check given that the configuration of the service is decided on sstore content for the `service_name` of the caller class, and `service_name` for both `CAInstance` and `KRAInstance` is `pki-tomcatd`.
To do this properly, you should probably rather implement `is_configured` in `DogtagInstance` and base it on `cls.config_file`, where `config_file` would be set in each of the classes `__init__()` method.

+1 for the test

rcritten:
Implemented Flo's suggestions.

ipatests/test_integration/test_upgrade.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class TestUpgrade(IntegrationTest):
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master, setup_dns=False)
+
+    def test_invoke_upgrader(self):
+        # For a new version in to require the upgrade to run
+        sysupgrade.set_upgrade_state('ipa', 'data_version',
```
flo-renaud: Unless I'm mistaken, this command would be run on the controller from where the tests are launched (which can be different from the master) and will trigger an error.
Moreover, I don't think you need to tweak the data_version for ipa-server-upgrade to run, the upgrade can be executed even without changing the version.

ipaserver/install/server/upgrade.py
```diff
@@ -1710,7 +1710,7 @@ def upgrade_configuration():
             )
         upgrade_pki(ca, fstore)
 
-        if kra.is_configured():
+        if kra.is_installed():
```
stlaz: 18 lines below there's a very similar check, you should probably change that one, too.

ipatests/test_integration/test_upgrade.py
rcritten: Right, of course. My purpose was to be absolutely sure that the upgrade was successful. I can remove this and revisit in the future if need be.

ipaserver/install/server/upgrade.py
rcritten: Good catch, done.

ipatests/test_integration/test_upgrade.py
```diff
@@ -0,0 +1,22 @@
...
+# Copyright (C) 2018  FreeIPA Contributors see COPYING for license
+#
+
+"""
+Module provides tests to verify that the upgrade script works.
+"""
+
+from ipatests.test_integration.base import IntegrationTest
+from ipatests.pytest_plugins.integration import tasks
+from ipaserver.install import sysupgrade
```
stlaz: Unused import.

rcritten: Sigh, it's what I get for rushing. Fixed.

flo-renaud:
Hi @rcritten 
I tested the fix with an install of server without kra, followed by ipa-server-upgrade, and it proceeded smoothly.
I was also able to launch the integration test successfully.

I'm just waiting for the PR-CI success to give the ACK.

stlaz:
PR CI build fails with `No matching package to install: 'python2-ldap >= 2.4.15'`

felipevolpone:
Probably PR needs a rebase

flo-renaud:
Now that the tests are green, ACK

tiran:
master:

* 8821f7ae8e666b4ae42e232c672d616bf7fbffeb Fix detection of KRA installation so upgrades can succeed

PR #1521, opened 2018-02-05
ipa host-add: do not raise exception when reverse record not added

flo-renaud:
When ipa host-add --random is unable to add a reverse record (for instance
because the server does not manage any reverse zone), the command
adds the host but exits (return code=1) with an error without actually
outputing the random password generated.
With this fix, the behavior is modified. The commands succeeds (return code=0)
but prints a warning.

This commit also adds a unit test.

https://pagure.io/freeipa/issue/7374

### DIFF
--- a/ipatests/test_xmlrpc/test_host_plugin.py
+++ b/ipatests/test_xmlrpc/test_host_plugin.py
@@ -31,7 +31,7 @@ import base64
 import pytest
 
 from ipapython import ipautil
-from ipalib import api, errors
+from ipalib import api, errors, messages
 from ipapython.dn import DN
 from ipapython.dnsutil import DNSName
 from ipatests.test_util import yield_fixture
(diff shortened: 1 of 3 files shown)


ipatests/test_xmlrpc/test_host_plugin.py
```diff
@@ -100,6 +100,9 @@
 host_cert = get_testcert(DN(('CN', api.env.host), subject_base()),
                          'host/%s@%s' % (api.env.host, api.env.realm))
 
+missingrevzone = u'30.16.172.in-addr.arpa.'
```
rcritten: I don't understand the purpose of this and the subsequent attempt to delete it. Isn't the point that the reverse zone is missing? Or is it that the zone in fact does exist in the beginning of the test so you delete it? Is this because it would not be predictable to use some hardcoded value?

Will this have a potential impact on subsequent tests?

flo-renaud: The test needs to use a reverse zone that is not managed by IPA, and xmlrpc tests can be launched on an existing instance (on which I do not control the defined reverse zones), hence the deletion.

But I noted that this test fails on Travis because there is no DNS server answering for this reverse zone using a private IP address. I need to re-work this fix.

flo-renaud:
Hi,
I reworked the test so that it is not dependent on the master config.

tiran:
master:

* 4295df17a42897f6f59be21c25c5dd03984e35d3 ipa host-add: do not raise exception when reverse record not added

PR #1523, opened 2018-02-06
Replace wsgi package conflict with config file

tiran:
Instead of a package conflict, freeIPA now uses an Apache config file to
enforce the correct wsgi module. The workaround only applies to Fedora
since it is the only platform that permits parallel installation of
Python 2 and Python 3 mod_wsgi modules. RHEL 7 has only Python 2 and
Debian doesn't permit installation of both variants.

Fixes: https://pagure.io/freeipa/issue/7394

### DIFF
--- a/ipaplatform/base/tasks.py
+++ b/ipaplatform/base/tasks.py
@@ -211,6 +211,10 @@ class BaseTaskNamespace(object):
         """Remove configuration of httpd service of IPA"""
         raise NotImplementedError()
 
+    def configure_httpd_wsgi_conf(self):
+        """Configure WSGI for correct Python version"""
+        raise NotImplementedError()
+
     def is_fips_enabled(self):
         return False
 
--- a/ipaplatform/redhat/tasks.py
+++ b/ipaplatform/redhat/tasks.py
@@ -30,6 +30,7 @@ import os
 import socket
 import traceback
 import errno
+import sys
 
 from ctypes.util import find_library
 from functools import total_ordering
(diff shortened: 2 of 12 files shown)


ipaplatform/base/tasks.py
```diff
@@ -211,6 +211,10 @@ def remove_httpd_service_ipa_conf(self):
         """Remove configuration of httpd service of IPA"""
         raise NotImplementedError()
 
+    def configure_httpd_wsgi_conf(self):
+        """Configure WSGI for correct Python version"""
+        # only used on Fedora, NOP on other platforms
+
```
Rezney: `raise NotImplementedError()` missing

ipaplatform/redhat/tasks.py
```diff
@@ -484,6 +485,33 @@ def configure_http_gssproxy_conf(self, ipaapi_user):
         os.chmod(paths.GSSPROXY_CONF, 0o600)
         self.restore_context(paths.GSSPROXY_CONF)
 
+    def configure_httpd_wsgi_conf(self):
+        """Configure WSGI for correct Python version"""
+        conf = paths.HTTPD_IPA_WSGI_MODULES_CONF
+        if sys.version_info > (3, 0):
+            wsgi_module = constants.MOD_WSGI_PYTHON3
+        else:
+            wsgi_module = constants.MOD_WSGI_PYTHON3
```
Rezney: should not it be `constants.MOD_WSGI_PYTHON2` here?

ipaplatform/base/tasks.py
tiran: Nope, I did this on purpose. The function is not implemented on any platform except Fedora.

ipaplatform/redhat/tasks.py
tiran: yeah, got catch :)

tiran:
master:

* 1785a3e17b9196ac661074ae66c4774d720358fb Replace wsgi package conflict with config file

PR #1524, opened 2018-02-06
tests: ca-less to ca-full - remove certupdate

Rezney:
After commits 8960141 and 97942a7 we do not need to run
ipa-certupdate command anymore when switching to ca-full.

This patch removes the above mentioned commands in order to
properly test the scenario.

https://pagure.io/freeipa/issue/7309

### DIFF
--- a/ipatests/test_integration/test_caless.py
+++ b/ipatests/test_integration/test_caless.py
@@ -1572,18 +1572,24 @@ class TestServerReplicaCALessToCAFull(CALessBase):
     def test_server_ipa_ca_install(self):
         """Install CA on master"""
 
-        ca_master = tasks.install_ca(self.master)
-        assert ca_master.returncode == 0
-        cert_update_master = self.master.run_command(['ipa-certupdate'])
-        assert cert_update_master.returncode == 0
-        cert_update_replica = self.replicas[0].run_command(['ipa-certupdate'])
-        assert cert_update_replica.returncode == 0
+        tasks.install_ca(self.master)
+        # We are not calling ipa-certupdate on replica here since the next step
+        # installs CA clone there.
+
+        ca_show = self.master.run_command(['ipa', 'ca-show', 'ipa'])
+        assert 'Subject DN: CN=Certificate Authority,O={}'.format(
+            self.master.domain.realm) in ca_show.stdout_text
 
     def test_replica_ipa_ca_install(self):
         """Install CA on replica"""
 
-        ca_replica = tasks.install_ca(self.replicas[0])
-        assert ca_replica.returncode == 0
+        replica = self.replicas[0]
+
+        tasks.install_ca(replica)
+
+        ca_show = replica.run_command(['ipa', 'ca-show', 'ipa'])
+        assert 'Subject DN: CN=Certificate Authority,O={}'.format(
+            self.master.domain.realm) in ca_show.stdout_text
 
 
 class TestReplicaCALessToCAFull(CALessBase):


Rezney:
@frasertweedale Hi, could you please check if this is enough or any other additional check is needed? Thanks...

frasertweedale:
Hi @Rezney.  One more check is needed: on the master, immediately after CA installation we need
to check that CA-related IPA commands work.  `ipa cert-show 1` should be sufficient.

ipatests/test_integration/test_caless.py
```diff
@@ -1574,10 +1574,9 @@ def test_server_ipa_ca_install(self):
 
         ca_master = tasks.install_ca(self.master)
         assert ca_master.returncode == 0
-        cert_update_master = self.master.run_command(['ipa-certupdate'])
-        assert cert_update_master.returncode == 0
-        cert_update_replica = self.replicas[0].run_command(['ipa-certupdate'])
```
stlaz: AFAIK `ipa-certupdate` would only be run on the machine `ipa-ca-install` was run on, therefore you need to keep the latter certupdate there.

frasertweedale: @stlaz we don't want to keep the latter `ipa-certupdate`, because the next step is to do a CA replica install, and we want to check that `ipa-ca-install` does the `ipa-certupdate` behaviour as a first step.

stlaz: @frasertweedale ah, of course, I did not scroll down, my bad

Rezney: Cool, thanks to both for watching 👍

Rezney:
Done. Internal Jenkins job passing.

ipatests/test_integration/test_caless.py
frasertweedale: Maybe a comment explaining why we do not run `ipa-certupdate` would be useful.

Rezney: done...

ipatests/test_integration/test_caless.py
```diff
@@ -1574,10 +1574,11 @@ def test_server_ipa_ca_install(self):
...
-        assert cert_update_replica.returncode == 0
+        # ipa-certupdate command is not necessary anymore.
```
stlaz: This comment is unfortunately very confusing. If you're browsing the code, you think - what do you mean anymore? Was there a `ipa-certupdate` call before? Why was it there?
Please, replace it with something like "we're not calling ipa-certupdate on replica here since the next step installs CA clone there" and remove the pagure ticket reference.

Rezney: oki

Rezney:
@frasertweedale as per @stlaz question on irc, show we add also same check (ipa cert-show 1) on the replica?

frasertweedale:
@Rezney it will not hurt to add it, but I don't think it's essential.

ipatests/test_integration/test_caless.py
```diff
@@ -1574,10 +1574,11 @@ def test_server_ipa_ca_install(self):
...
         assert ca_master.returncode == 0
-        cert_update_master = self.master.run_command(['ipa-certupdate'])
-        assert cert_update_master.returncode == 0
-        cert_update_replica = self.replicas[0].run_command(['ipa-certupdate'])
-        assert cert_update_replica.returncode == 0
+        # We are not calling ipa-certupdate on replica here since the next step
+        # installs CA clone there.
+
+        ca_show = self.master.run_command(['ipa', 'ca-show', 'ipa'])
+        assert ca_show.returncode == 0
```
tiran: Could you please check the output of the command to verify that it prints the correct CA?

``assert ca_show.returncode == 0`` shouldn't be necessary because ``run_command`` raises an exception by default.

Rezney: sure, done...

Rezney:
@tiran @stlaz @frasertweedale is current state ok? Thx...

frasertweedale:
@Rezney the code is OK.  Will break if CLI labels, default Subject DN or subject base ever change,
but we only need to deal with that if it ever does happen.  @tiran @stlaz any objections to merging this now?

Rezney:
Was not running the second ca show on replica but on master. Fixed now...

tiran:
master:

* 2a50a7daf057860818f1595fe8156de52e9b6b9e tests: ca-less to ca-full - remove certupdate

PR #1529, opened 2018-02-06
ipa-kdb: update trust information in all workers

sumit-bose:
Currently there is already code to make sure that after trust is established an
AS-REQ of the local HTTP principal causes a refresh of the internal structures
holding the information about the trusted domains.

But this refreshes only the data of the current krb5kdc worker process on the
local host. Other workers and the KDCs on other hosts will update the data
eventually when a request with a principal from a trusted realm is handled.

During this phase, which might last quite long if remote principals are only
handled rarely, TGTs for local principals might or might not contain a PAC
because the decision if a PAC should be added or not is based on the
information about trusted domains. Since the PAC is needed to access services
on the AD side this access might fail intermittently depending which worker
process on which host is handling the request. This might e.g. affect SSSD
running on the IPA server with two-way trust.

To fix this this patch calls ipadb_reinit_mspac() whenever a PAC is needed but
without the 'force' flag so that the refresh will only happen if it wasn't
called recently (currently not more often than once a minute).

An alternative might be to do the refresh only when processing cross-realm TGT
requests. But this would be already too late because the local principal asking
for a cross-realm ticket would not have a PAC and hence the first attempt will
still fail due to the missing PAC. And injecting the PAC in the cross-realm TGT
while there is none in the requesting ticket does not sound right.

Related to https://pagure.io/freeipa/issue/7351

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb_mspac.c
+++ b/daemons/ipa-kdb/ipa_kdb_mspac.c
@@ -2121,6 +2121,7 @@ krb5_error_code ipadb_sign_authdata(krb5_context context,
     int result;
     krb5_db_entry *client_entry = NULL;
     krb5_boolean is_equal;
+    bool force_reinit_mspac = false;
 
 
     is_as_req = ((flags & KRB5_KDB_FLAG_CLIENT_REFERRALS_ONLY) != 0);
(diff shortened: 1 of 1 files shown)


tiran:
@abbra @simo5 please have a look.

simo5:
This will cause a reinit on each AS request, not to hot on that.
Can we think of a better way to handle it that is not so expensive ?
Keep in mind that clients are opretty aggressive with timeouts on requests and give up after 1 second on the first try (and then slowly back off), too much stuff in this "hot" path will basically cause a crappy experience and overload og the KDC.

sumit-bose:
@simo5, in general the reinit will "only" happen once a minute (current timeout in ipadb_reinit_mspac()), only a request with the HTTP principal of the local server will trigger a forced reinit.

As mentioned in the commit message I see no other suitable code path. The proper solution might be persistent searches or something similar which would require larger changes.

tiran:
master:

* 73f61ce214e784ab8176a1f7acac6a3dbf1474ae ipa-kdb: update trust information in all workers

PR #1536, opened 2018-02-07
Fix multiple uninstallation of server

tiran:
"ipa-server-install --uninstall" no longer fails with error message
"'Env' object has no attribute 'basedn'" when executed on a system that
has no freeIPA server installation.

Fixes: https://pagure.io/freeipa/issue/7063

### DIFF
--- a/.test_runner_config.yaml
+++ b/.test_runner_config.yaml
@@ -74,6 +74,8 @@ steps:
   - ipa-run-tests ${tests_ignore} -k-test_dns_soa ${tests_verbose} ${path}
   - '! grep -n -C5 BytesWarning /var/log/httpd/error_log'
   - ipa-server-install --uninstall -U
+  # second uninstall to verify that --uninstall without installation works
+  - ipa-server-install --uninstall -U
 tests:
   ignore:
   - test_integration
--- a/ipalib/config.py
+++ b/ipalib/config.py
@@ -576,6 +576,16 @@ class Env(object):
         if 'log' not in self:
             self.log = self._join('logdir', '%s.log' % self.context)
 
+        # Workaround for ipa-server-install --uninstall. When no config file
+        # is available, we set realm, domain, and basedn to RFC 2606 reserved
+        # suffix to suppress attribute errors during uninstallation.
+        if (self.in_server and self.context == 'installer' and
+                not getattr(self, 'config_loaded', False)):
+            if 'realm' not in self:
+                self.realm = 'UNCONFIGURED.INVALID'
+            if 'domain' not in self:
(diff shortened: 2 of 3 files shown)


tiran:
What's a good place for an integration test?

felipevolpone:
probably in test_integration/test_installation.py

rcritten:
I think ipa-client-automount should be extended to return 2 when it is not configured, like ipa-client-install does. Then this can avoid this when uninstalling a blown install (or maybe this should be treated separately):

```
Removing IPA client configuration
Unconfigured automount client failed: Command 'ipa-client-automount --uninstall --debug' returned non-zero exit status 1.
```

I guess this was suppressed in the past because there were default values defined previously?

tiran:
It's going to take some work to bend ``ipa-client-automount`` to our will. It's a rather old script that doesn't use the new framework yet. I opened https://pagure.io/freeipa/issue/7396

tiran:
@felipevolpone ping

simo5:
LGTM

felipevolpone:
The PR works and it fixes the problem. However, I still think the output is very confusing. If I install IPA and then run `ipa-server-install --uninstall` twice, the output is the same in the second time (check below). I don't think it makes sense print `If this server is the last instance of CA,
KRA, or DNSSEC master, uninstallation may result in data loss.` when the IPA is not installed.

```
[root@master rpms]# ipa-server-install --uninstall -U
...
[root@master rpms]# ipa-server-install --uninstall -U
WARNING:
IPA server is not configured on this system. If you want to install the
IPA server, please install it using 'ipa-server-install'.

WARNING: Failed to connect to Directory Server to find information about
replication agreements. Uninstallation will continue despite the possible
existing replication agreements.

If this server is the last instance of CA,
KRA, or DNSSEC master, uninstallation may result in data loss.
Shutting down all IPA services
Configuring certmonger to stop tracking system certificates for KRA
Configuring certmonger to stop tracking system certificates for CA
Removing IPA client configuration
IPA client is not configured on this system.
The ipa-client-install command failed. See /var/log/ipaclient-uninstall.log for more information
```

IMHO, the output should be just:
```
IPA server is not configured on this system. If you want to install the
IPA server, please install it using 'ipa-server-install'.
```

However, we can address this issue with another PR.

tiran:
I agree, the log output should be terse. It's still useful to go through all uninstallation steps. Sometimes uninstallation fails. A second uninstallation can remove installation artefacts.

tiran:
master:

* 0cc2a6cae01043f0aba32319ac8c6475780b6d7f Fix multiple uninstallation of server

PR #1537, opened 2018-02-07
ipa-kdb: use magic value to check if ipadb is used

sumit-bose:
The certauth plugin is configured in /etc/krb5.conf independently form
the database module. As a result the IPA certauth plugin can be added to
the configuration without the IPA DAL driver. Since the IPA certauth
plugin depends on the presence of the IPA DAL driver this patch adds a
magic value at the beginning of struct ipadb_context which can be
checked to see if the IPA DAL driver is properly initialized.

Resolves https://pagure.io/freeipa/issue/7261

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb.c
+++ b/daemons/ipa-kdb/ipa_kdb.c
@@ -522,6 +522,7 @@ static krb5_error_code ipadb_init_module(krb5_context kcontext,
     if (!ipactx) {
         return ENOMEM;
     }
+    ipactx->magic = IPA_CONTEXT_MAGIC;
 
     /* only check for unsupported 'temporary' value for now */
     for (i = 0; db_args != NULL && db_args[i] != NULL; i++) {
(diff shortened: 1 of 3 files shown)


tiran:
Don't mind, I was just curious. I love a good riddle.

tiran:
master:

* ec228f411e480e03108f0e0fec0f706e159ca61c ipa-kdb: use magic value to check if ipadb is used

simo5:
Not that it really matters but why do we need a magic value ?
Wouldn't a boolean is_dal_initialized set to true only by the DAL have been enough ?

sumit-bose:
It is important here that we have the context of the IPA DAL driver and not any other DAL driver which might accidentally have a boolean as first element of its internal context.

simo5:
Ok, I did not think you could get a wrong handle.

frozencemetery:
My understanding is that you'll never get the "wrong" handle - you just get whatever DAL is loaded, and if that's not the freeipa one when you expect it to be, badness ensues.

simo5:
@frozencemetery getting a non IPA DAL handle is the definition of wrong here :-)

PR #1538, opened 2018-02-07
Replace hard-coded paths with path constants

tiran:
Several run() calls used hard-coded paths rather than pre-defined paths
from ipaplatform.paths. The patch fixes all places that I was able to
find with a simple search.

### DIFF
--- a/client/ipa-client-automount
+++ b/client/ipa-client-automount
@@ -92,7 +92,7 @@ def wait_for_sssd():
     time.sleep(1)
     while n < 10 and not found:
         try:
-            ipautil.run(["getent", "passwd", "admin@%s" % api.env.realm])
+            ipautil.run([paths.GETENT, "passwd", "admin@%s" % api.env.realm])
             found = True
         except Exception:
             time.sleep(1)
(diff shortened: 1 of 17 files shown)


rcritten:
In principle it looks ok to me. I found a few additional changes:

ipatests/test_ipapython/test_ipautil.py has a slew of echo calls. I assume this is getting out to /usr/bin/echo.

ipaserver/install/krbinstance.py calls klist

ipaclient/install/client.py calls sshd and ipa-client-automount (there is another one you missed)

tiran:
@rcritten Thanks for your review. I have addressed your remarks.

rcritten:
Just one final request. IIRC this is related to porting IPA to work on SuSE. If that is true can you add a mention to the commit message before pushing?

tiran:
@rcritten ok

tiran:
master:

* 2391c75e3d7efcdc5c2f49defa5138fc7e6def06 Replace hard-coded paths with path constants

tiran:
Fix has landed in master. I'm not going to backport the fix since merge to ipa-4-6 does not apply cleanly and we don't need it in 4.6

PR #1544, opened 2018-02-08
[Backport][ipa-4-5] ipa-kdb: use magic value to check if ipadb is used

tiran:
Manual backport of PR #1537 

The certauth plugin is configured in /etc/krb5.conf independently form
the database module. As a result the IPA certauth plugin can be added to
the configuration without the IPA DAL driver. Since the IPA certauth
plugin depends on the presence of the IPA DAL driver this patch adds a
magic value at the beginning of struct ipadb_context which can be
checked to see if the IPA DAL driver is properly initialized.

Resolves https://pagure.io/freeipa/issue/7261

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb.c
+++ b/daemons/ipa-kdb/ipa_kdb.c
@@ -522,6 +522,7 @@ static krb5_error_code ipadb_init_module(krb5_context kcontext,
     if (!ipactx) {
         return ENOMEM;
     }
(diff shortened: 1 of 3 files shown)


tiran:
ipa-4-5:

* f7629b32a60514529f9ea41a3ab28447f57f4587 ipa-kdb: use magic value to check if ipadb is used

PR #1547, opened 2018-02-08
Update existing 389-DS cn=RSA,cn=encryption config

tiran:
389-DS >= 1.4.0 on Fedora 28 has a default entry for
cn=RSA,cn=encryption,cn=config. The installer now updates the entry in
case it already exists. This ensures that token and personality are
correct for freeIPA

Fixes: https://pagure.io/freeipa/issue/7393


tiran:
master:

* 939db89cacdd9450400093be33af891d17545c10 Update existing 389-DS cn=RSA,cn=encryption config

PR #1551, opened 2018-02-08
ipa host-add --ip-address: properly handle NoNameservers

flo-renaud:
When ipa host-add --ip-address is called but no DNS server is able to answer
for the reverse zone, get_reverse_zone raises a NoNameservers exception.
The exception is not managed by add_records_for_host_validation, and this
leads to the command exiting on failure with an InternalError:
    $ ipa host-add testhost.ipadomain.com --ip-address 172.16.30.22
A traceback is also logged in httpd error_log.

This commit properly handles the exception, and adds a test.

https://pagure.io/freeipa/issue/7397

### DIFF
--- a/ipaserver/plugins/dns.py
+++ b/ipaserver/plugins/dns.py
@@ -539,7 +539,14 @@ def get_reverse_zone(ipaddr):
     """
     ip = netaddr.IPAddress(str(ipaddr))
     revdns = DNSName(unicode(ip.reverse_dns))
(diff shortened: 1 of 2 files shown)


flo-renaud:
Pushed with fixed lint issues

flo-renaud:
master:

* 7364c268ebc241c2fa456951f0790ee5d83a7b3a ipa host-add --ip-address: properly handle NoNameservers

PR #1561, opened 2018-02-09
Use system-wide crypto-policies on Fedora

tiran:
HTTPS connections from IPA framework now uses system-wide
crypto-policies on Fedora.

The 'DEFAULT' crypto policy also includes unnecessary ciphers for
PSK, SRP, aDSS and 3DES. Since these ciphers are not used by freeIPA,
they are explicitly excluded.

Fixes: https://pagure.io/freeipa/issue/4853

### DIFF
--- a/install/share/bind.named.conf.template
+++ b/install/share/bind.named.conf.template
@@ -21,6 +21,9 @@ options {
 	bindkeys-file "$BINDKEYS_FILE";
 
 	managed-keys-directory "$MANAGED_KEYS_DIR";
+
+	/* crypto policy snippet on platforms with system-wide policy. */
+	$INCLUDE_CRYPTO_POLICY
 };
 
 /* If you want to enable debugging, eg. using the 'rndc trace' command,
--- a/ipaserver/install/bindinstance.py
+++ b/ipaserver/install/bindinstance.py
@@ -768,6 +768,13 @@ class BindInstance(service.Service):
             logger.debug("Unable to mask named (%s)", e)
 
     def __setup_sub_dict(self):
+        if paths.NAMED_CRYPTO_POLICY_FILE is not None:
+            crypto_policy = 'include "{}";'.format(
+                paths.NAMED_CRYPTO_POLICY_FILE
+            )
+        else:
+            crypto_policy = "// not available"
+
         self.sub_dict = dict(
             FQDN=self.fqdn,
             SERVER_ID=installutils.realm_to_serverid(self.realm),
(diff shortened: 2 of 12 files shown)


stlaz:
From Travis:
```
Configuring DNS (named)
  [1/11]: generating rndc key file
  [2/11]: adding DNS container
  [3/11]: setting up our zone
  [4/11]: setting up our own record
  [5/11]: setting up records for other masters
  [6/11]: adding NS record to the zones
  [7/11]: setting up kerberos principal
  [8/11]: setting up named.conf
  [error] KeyError: 'INCLUDE_CRYPTO_POLICY'
ipapython.admintool: ERROR    'INCLUDE_CRYPTO_POLICY'
ipapython.admintool: ERROR    The ipa-server-install command failed. See /var/log/ipaserver-install.log for more information
```

tiran:
Memo to me: don't change and test with the file in ``ipaserver/build/lib...``

install/share/bind.named.conf.template
```diff
@@ -21,6 +21,9 @@ options {
 	bindkeys-file "$BINDKEYS_FILE";
 
 	managed-keys-directory "$MANAGED_KEYS_DIR";
+
+	/* crypto policy snippet on platforms with system-wide policy. */
+	$INCLUDE_CRYPTO_POLICY
```
stlaz: I checked the man pages of `named.conf` but did not find a description of what happens when a file path is pasted to the config. Can you point me to docs that would explain this?

ipaserver/install/bindinstance.py
```diff
@@ -768,6 +768,13 @@ def __enable(self):
             logger.debug("Unable to mask named (%s)", e)
 
     def __setup_sub_dict(self):
+        if paths.NAMED_CRYPTO_POLICY_FILE is not None:
+            crypto_policy = 'include "{}";'.format(
```
stlaz: `named.conf` man pages don't describe the `include` statement 🤔

tiran: IPA uses include already, see https://github.com/freeipa/freeipa/blob/release-4-6-3/install/share/bind.named.conf.template#L42-L45

It's documented here, http://www.zytrax.com/books/dns/ch7/include.html

stlaz: Thanks.

rcritten:
LGTM, leaving final ack to Standa.

stlaz:
ACK, seems alright.

tiran:
master:

* 90a75f0d4300126f18dabfb9ca4df59cab4d97cb Use system-wide crypto-policies on Fedora
* aee0d2180c7119bef30ab7cafea81dc3df1170b7 Upgrade named.conf to include crypto policy
* 68caeb8b1921af5d2c872f24984865d471a7df3c Add mocked test for named crypto policy update

PR #1570, opened 2018-02-13
[Backport][ipa-4-6] - Help cache test

Rezney:
This test case addresses upsteam ticket #6999, where "ipa help"
does not work if called when no schema is cached.

https://pagure.io/freeipa/issue/7325


Rezney:
ipa-4-6:

* 225186d876cf26c591a24a553d56b40e4447b689 test_help: test "help" command without cache

PR #1571, opened 2018-02-13
Add workaround for netiface wheel bug

tiran:
Don't use binary wheel for netiface to workaround a bug in
netifaces-0.10.6-cp27-cp27mu-manylinux1_x86_64.whl. The shared library
uses UCS2 APIs although the wide unicode build should use UCS4 APIs.

See: https://github.com/al45tair/netifaces/issues/2

Alastair Houghton removed the bogus netiface wheels. This PR should do the trick in case the removal isn't sufficient.


tiran:
Tests are passing again without this workaround.

PR #1575, opened 2018-02-14
[Backport][ipa-4-5] - External ca py crypto backport

Rezney:
Switch external CA generation from certutil to python-cryptography
as this way of handling the certificates should be more readable,
maintainable and extendable (e.g. extensions handling).

Also as external CA is now a separate module we can import it and
use elsewhere.

https://pagure.io/freeipa/issue/7154


Rezney:
ipa-4-5:

* 38d5142aaf1fe0701e6ad66f716dd883f0517722 test_external_ca: switch to python-cryptography

PR #1582, opened 2018-02-14
User must not be able to delete his last active otp token

flo-renaud:
Fix and unit test for the issue. When an OTP token is the last active token, the user should not be allowed to delete its token if 'otp' is the only allowed authentication method.

Fixes:
https://pagure.io/freeipa/issue/7012

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-otp-lasttoken/ipa_otp_lasttoken.c
+++ b/daemons/ipa-slapi-plugins/ipa-otp-lasttoken/ipa_otp_lasttoken.c
@@ -50,6 +50,7 @@
 #define OTP_CONTAINER "cn=otp,%s"
 
 static struct otp_config *otp_config;
+void *ipa_otp_lasttoken_plugin_id;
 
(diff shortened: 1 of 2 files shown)


flo-renaud:
This PR replaces PR 1416.

flo-renaud:
@npmccallum 
Hi, could you have a look at this PR? Thanks.

npmccallum:
@flo-renaud This fix looks like the correct one to me. The existing logic is correct, we just aren't getting all the configuration because we call the API at the wrong time. Excellent find!

I can't comment on the correctness of the test code. But I approve the patch to the C code.

slaykovsky:
@flo-renaud thanks for a patch!

As @npmccallum reviewed C code, I took a look at the Python one :)

The test looks good and it passes so I will ack.

Thanks!

tiran:
master:

* 8b6506a5f1176ad768bb0e513436009906b8ff63 User must not be able to delete his last active otp token
* c701cd21d31e6bdf5f1078cdfca49e410e093e28 389-ds OTP lasttoken plugin: Add unit test

PR #1585, opened 2018-02-15
[Py3] Fix i18n test for Chinese translation

tiran:
Python 3's regular expression default to full range of unicode
characters. Restrict \w matches to ASCII and drop \b suffix check to fix
a problem with validation the Chinese translation zh_CN.

Co-Authored-By: Stanislav Laznicka <slaznick@redhat.com>


stlaz:
Works great, ACK.

tiran:
master:

* 0ee3a267112dfafe12726fb185a4ce260c67aff7 Fix i18n test for Chinese translation

PR #1587, opened 2018-02-15
Unified ldap_initialize() function

tiran:
Replace all ldap.initialize() calls with a helper function
ldap_initialize(). It handles cacert and cert validation correctly. It
also provides a unique place to handle python-ldap 3.0 bytes warnings in
the future.

Fixes: https://pagure.io/freeipa/issue/7411

### DIFF
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -36,6 +36,7 @@ from ipalib.text import _
 from ipapython import ipautil, ipaldap, kerberos
 from ipapython.admintool import ScriptError
 from ipapython.dn import DN
+from ipapython.ipaldap import ldap_initialize
 from ipaplatform.paths import paths
 from ipaserver.install import installutils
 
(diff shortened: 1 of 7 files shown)


tiran:
@abbra I also changed some code in AD related modules. Please verify.

ipaserver/install/replication.py
```diff
@@ -1077,13 +1078,10 @@ def setup_winsync_replication(self,
...
         try:
             # Validate AD connection
-            ad_conn = ldap.initialize('ldap://%s' % ipautil.format_netloc(ad_dc_name))
-            # the next one is to workaround bugs arounf opendalp libs+NSS db
-            # we need to first specify the OPT_X_TLS_CACERTFILE and _after_
-            # that initialize the context to prevent TLS connection errors:
-            # https://bugzilla.redhat.com/show_bug.cgi?id=800787
-            ad_conn.set_option(ldap.OPT_X_TLS_CACERTFILE, cacert)
-            ad_conn.set_option(ldap.OPT_X_TLS_NEWCTX, 0)
+            ad_conn = ldap_initialize(
```
abbra: This change modifies a behavior: we were ensuring no new context is used here but with ldap_initialize() we will now force OPT_X_TLS_NEWCTX if cacertfile is passed which is against what this code is intending.

tiran: I don't understand what you are trying to tell me. The code already creates a new context in ``set_option(ldap.OPT_X_TLS_NEWCTX, 0)``. The value ``0`` does not mean "no new context". It actually creates a new context. The documentation https://linux.die.net/man/3/ldap_set_option is misleading. Any non-NULL value creates a new context.

> LDAP_OPT_X_TLS_NEWCTX
> Instructs the library to create a new TLS library context. invalue must be const int *. A non-zero value pointed to by invalue tells the library to create a context for a server. 

A new context **must** be created whenever a TLS related option is changed on a connection. Otherwise OpenLDAP will silently ignore any TLS settings. This is one of the subtle bugs I was referring to.

tiran:
``set_option(ldap.OPT_X_TLS_NEWCTX, 0)`` creates a new context, see https://github.com/openldap/openldap/blob/OPENLDAP_REL_ENG_2_4_45/libraries/libldap/tls2.c#L799-L804

The code must create a new context when it changes any of the TLS parameters. Otherwise most TLS options are silently ignored.

abbra:
Got it. We need to run a winsync test to verify that nothing breaks against AD here but the code change then isn't introducing any issue so OK.

tiran:
I opened a documentation bug to clarify the option, http://www.openldap.org/its/index.cgi/Incoming?id=8805

tiran:
master:

* 1b0c55a3b362a8655d0ec6c55d30d6173fa88eee Unified ldap_initialize() function

PR #1591, opened 2018-02-15
ipa-custodia-checker now uses python3 shebang

tiran:
#1024 missed one script that was introduced later.


stlaz:
This is a new script which #1024 forgot about during its rebase. ACK.

tiran:
master:

* a349629fbad8bcb0179c1a1f6babe76464be89d2 ipa-custodia-checker now uses python3 shebang

PR #1596, opened 2018-02-16
webui:test realm domain add with DNS check

pvoborni:
Try adding and deleting with "Check DNS" (in html 'ok' button)

DNS check expects that the added domain will have DNS record:
    TXT _kerberos.$domain "$REALM"

When a new domain is added using dnszone-add it automatically adds
this TXT record and adds a realm domain. So in order to test without
external DNS we must get into state where realm domain is not added
(in order to add it) but DNS domain with the TXT record exists.

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -11,6 +11,10 @@ topologies:
     name: master_1repl_1client
     cpu: 4
     memory: 6700
+  ipaserver: &ipaserver
+    name: ipaserver
+    cpu: 1
+    memory: 2400
 
 jobs:
   fedora-27/build:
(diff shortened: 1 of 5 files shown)


pvoborni:
When somebody will do initial review and ack the approach, I'll remove the temp commit to trigger the usual PR-CI gating tests but till then I'll keep the results fo the Web UI test run here.

celestian:
ACK

tiran:
master:

* d7d13bc95043a85bd7621e364554df5989a7ce9d webui:tests: move DNS test data to separate file
* d73d49f3f69ed34cf8386422822a85a4707bf669 webui:tests: realm domain add with DNS check
* 6b214512b335998b424ef0c160b593b162b07c29 webui:tests: close big notifications in realm domains tests
* db2222fee4558004968900e8d1421abfb409f53a temp commit to run the affected tests

pvoborni:
This PR was acked bit soon, the temp commit should have been removed. It was reverted. Given that it changes only tests, there is low chance that it broke IPA (usualy PR-CI test were not run).

PR #1597, opened 2018-02-16
realm domains: improve doc text

pvoborni:
It is quite unclear how realm domains behave without reading source
code. New doc text describes its purpose and how it is managed.

### DIFF
--- a/ipaserver/plugins/realmdomains.py
+++ b/ipaserver/plugins/realmdomains.py
@@ -36,6 +36,16 @@ Realm domains
 
 Manage the list of domains associated with IPA realm.
 
+This list is useful for Domain Controllers from other realms which have
+established trust with this IPA realm. They need the information to know
+which request should be forwarded to KDC of this IPA realm.
+
+Automatic management: a domain is automatically added to the realm domains
+list when a new DNS Zone managed by IPA is created. Same applies for deletion.
+
+Externally managed DNS: domains which are not managed in IPA server DNS
+need to be manually added to the list using ipa realmdomains-mod command.
+
 EXAMPLES:
 
  Display the current list of realm domains:
@@ -118,10 +128,22 @@ class realmdomains(LDAPObject):
     )
 
 
-
 @register()
 class realmdomains_mod(LDAPUpdate):
-    __doc__ = _('Modify realm domains.')
+    __doc__ = _("""
+    Modify realm domains
+
+    DNS check: When manually adding a domain to the list, a DNS check is
+    performed by default. It ensures that the domain is associated with
+    the IPA realm, by checking whether the domain has a _kerberos TXT record
+    containing the IPA realm name. This check can be skipped by specifying
+    --force option.
+
+    Removal: when a realm domain which has a matching DNS zone managed by
+    IPA is being removed, a corresponding _kerberos TXT record in the zone is
+    removed automatically as well. Other records in the zone or the zone
+    itself are not affected.
+    """)
 
     takes_options = LDAPUpdate.takes_options + (
         Flag('force',


pvoborni:
@abbra  I was also looking at documentation: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html-single/windows_integration_guide/#trust-realmdomain and the first paragraph there seems quite misleading to me. Do you think the wording there is good?

abbra:
You are right that the first paragraph in the docs is not good. It should be removed completely, as well as the second one. I think a text you added in this PR should be used instead.

abbra:
LGTM

pvoborni:
Added a ticket to commit message so that it can be backported to 4.6.  https://pagure.io/freeipa/issue/7424

ipaserver/plugins/realmdomains.py
```diff
@@ -118,10 +128,22 @@ class realmdomains(LDAPObject):
...
 class realmdomains_mod(LDAPUpdate):
-    __doc__ = _('Modify realm domains.')
+    __doc__ = _("""
+    Modify realm domains
+
+    DNS check: When adding domain to the list manually, a DNS check is
+    performed by default. It checks whether the domain is associated
+    with the IPA realm. In practice it checks whether the domain has
+    _kerberos TXT record containing IPA realm name. This check can be
+    skipped by specifying --force option.
```
flo-renaud: nitpicking, but I would rephrase:
When manually adding a domain to the list, a DNS check is performed by default. It ensures that the domain is associated with the IPA realm, by checking whether the domain has a _kerberos TXT record containing the IPA realm name. This check can be skipped by specifying --force option.

ipaserver/plugins/realmdomains.py
```diff
@@ -36,6 +36,16 @@
 
 Manage the list of domains associated with IPA realm.
 
+This list is useful for Domain Controllers from other realms which have
+established trust with this IPA realm. They need the information to know
+which request should be forwarded to KDC of this IPA realm.
+
+Automatic management: a domain is added to realm domains list automatically
+when new DNS Zone managed by IPA is created. Same applies for deletion.
```
flo-renaud: nitpicking, but I would rephrase:
a domain is automatically added to the realm domains list when a new DNS Zone managed by IPA is created. Same applies for deletion.

ipaserver/plugins/realmdomains.py
```diff
@@ -36,6 +36,16 @@
...
+
+Externally managed DNS: domains which are not managed in IPA server DNS needs
+to be added to the list manually using `ipa realmdomains-mod` command.
```
flo-renaud: nitpicking, but I would rephrase:
domains which are not managed in IPA server DNS need to be manually added to the list using `ipa realmdomains-mod` command.

ipaserver/plugins/realmdomains.py
abbra: This got me to another point: Active Directory domains never have _kerberos TXT record associated, so the check is inadequate for that use case. We should use a different method by looking up for MSDCS entries in the domain and its parents. However, I'm not sure where we should stop. See https://technet.microsoft.com/en-us/library/cc759550(WS.10).aspx for details

pvoborni:
@flo-renaud thanks for the review, PR updated, all nitpicks should be addressed.

flo-renaud:
@pvoborni 
thanks for the update. The doc strings look good now.
As a follow-up, I will open a doc bug in order to fix Windows Integration guide and a pagure ticket linked to @abbra comment about not checking TXT records for AD.

pvoborni:
@flo-renaud , thanks for opening the bug

Tiboris:
master:

* b43e73143de66b0068cd01379c134d5ef6e304e9 realm domains: improve doc text

PR #1598, opened 2018-02-16
fastcheck: do not test context in pycodestyle

pvoborni:
`git diff` shows also context lines by default. When passed to pycodestyle
it can produce errors unrelated to changed lines. It prevents running of
subsequent checks.

Limiting context to 0 lines by `git diff -U0` enables to test only the
modified lines and allows to run subsequent checks.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -215,7 +215,7 @@ endif
 	    echo -e "Fast linting files:\n$${FILES}\n"; \
 	    echo "pycodestyle"; \
 	    echo "-----------"; \
-	    git diff $${MERGEBASE} | \
+	    git diff -U0 $${MERGEBASE} | \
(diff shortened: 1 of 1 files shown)


pvoborni:
Don't know if it should be also somewhere else, but looking at `.travis_run_task.sh` and there already is `-U0`

```
if [[ "$TASK_TO_RUN" == "lint" ]]
then
    if [[ "$TRAVIS_EVENT_TYPE" == "pull_request" ]]
    then
        git diff origin/$TRAVIS_BRANCH -U0 | pycodestyle --diff &> $PEP8_ERROR_LOG ||:
    fi 
fi
```

Tiboris:
master:

* f316eb83ddbdf270411a97460a6e72f8cc66b5dd fastcheck: do not test context in pycodestyle

PR #1599, opened 2018-02-19
Fix FileStore.backup_file() not to backup same file

stlaz:
FileStore.backup_file() docstring claimed not to store a
copy of the same file but the behavior of the method did not
match this description.

This commit makes the backed-up file filename derivation
deterministic by hashing its content by SHA-256, thus it
should not back up two files with the same filename and content.

### DIFF
--- a/ipalib/install/sysrestore.py
+++ b/ipalib/install/sysrestore.py
@@ -29,6 +29,8 @@ import os.path
 import shutil
 import random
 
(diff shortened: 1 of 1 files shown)


stlaz:
master:

* 364ffd5a0ff0848ec203f39397f00733a76d3350 Fix FileStore.backup_file() not to backup same file

PR #1612, opened 2018-02-22
[Backport][ipa-4-5]-  test_caless: introduce new python makepki + fix SKI extension issue

Rezney:
Change makepki.sh for new makepki.py which should be more
readable, maintainable and extendable than the old script.
In this test we use it as a module and import create_pki().

The new makepki adds SKI and AKI extensions for correct
cert validation.

Other minor changes needed as we do not use NSS to store our
certificates on the test controller.

https://pagure.io/freeipa/issue/7030

Signed-off-by: Michal Reznik <mreznik@redhat.com>


Rezney:
ipa-4-5:

* f84fc586b4adc48987bb85d2e38633b886185994 test_caless: introduce new python makepki + fix SKI extension issue

PR #1614, opened 2018-02-22
autoconf now prefers Python 3 over 2

tiran:
The configure script now looks for Python 3.6 or newer, then falls back to Python 2. All Makefile default to Python 3 if Python 3 is available.

Move logic for installing just the Python packages out of the spec file and into our root Makefile. It removes code duplication to simplify a spec file that supports building without Python 2

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,8 @@
 ACLOCAL_AMFLAGS = -I m4
 
 if ENABLE_SERVER
-    SERVER_SUBDIRS = daemons init install ipaserver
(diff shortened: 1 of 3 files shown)


tiran:
**ping** The PR is waiting for a review.

abbra:
ACK, looks good and passes tests.

tiran:
master:

* fd9ede5296afe7bfb52c4128d9dc1721a76f9add Simplify Python package installation
* fa94ef0412c4f24d49aa6dfbd15e3634d52d1ff5 autoconf prefers Python 3 over 2

PR #1615, opened 2018-02-22
Remove deprecated -p option from ipa-dns-install

tiran:
The option has been deprecated since at least freeIPA release 4.3.0 when
the installer was changed to use LDAPI.

See: https://pagure.io/freeipa/issue/4933

### DIFF
--- a/install/tools/ipa-dns-install
+++ b/install/tools/ipa-dns-install
@@ -25,8 +25,6 @@ import logging
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


tiran:
Even longer, I see a deprecation in 4.2.0, https://github.com/freeipa/freeipa/blob/release-4-2-0/install/tools/ipa-dns-install#L93-L95

tiran:
master:

* 2d8d5ad8d9a4c1fedce4ed4c6bcd1f5861ade982 Remove deprecated -p option from ipa-dns-install

PR #1616, opened 2018-02-22
Move DNS related files to server-dns package

tiran:
The freeipa-server package was shipping files that are only used by
freeipa-server-dns.

The systemd RPM macros are also available in RHEL 7.5.


tiran:
master:

* cfe4150b223355fc31d11955439f1704d9a9b416 Move DNS related files to server-dns package

PR #1617, opened 2018-02-22
[Backport][ipa-4-6] - test for second phase of ipa-server-install with --external-ca when dirsrv instance is stopped

Rezney:
When the dirsrv service, which gets started during the first
ipa-server-install --external-ca phase, is not running when the
second phase is run with --external-cert-file options, the
ipa-server-install command fail.

This test checks if second phase installs successfully when dirsrv
is stoped.


Rezney:
ipa-4-6:

* 1a3fc3b7a5f4f856e30567e3a0f71490ba015c86 When the dirsrv service, which gets started during the first ipa-server-install --external-ca phase, is not running when the second phase is run with --external-cert-file options, the ipa-server-install command fail.
* 522c681832259260fc1c75313e761ff1f1ba031c Updated the TestExternalCA with the functions introduced for the steps of external CA installation.

PR #1621, opened 2018-02-22
OTP FIPS mode fixes

npmccallum:
The following patches add various fixes for using OTP on a server running in FIPS mode.

### DIFF
--- a/daemons/ipa-slapi-plugins/libotp/hotp.c
+++ b/daemons/ipa-slapi-plugins/libotp/hotp.c
@@ -46,6 +46,7 @@
 #include <time.h>
 
 #include <nss.h>
+#include <blapit.h>
 #include <pk11pub.h>
 #include <hasht.h>
 #include <prnetdb.h>
@@ -65,6 +66,49 @@ static const struct {
     { "sha512", CKM_SHA512_HMAC },
     { }
 };
+
+static PK11SymKey *
+import_key(PK11SlotInfo *slot, CK_MECHANISM_TYPE mech, SECItem *key)
+{
+    uint8_t ct[(key->len / AES_BLOCK_SIZE + 1) * AES_BLOCK_SIZE];
+    uint8_t iv[AES_BLOCK_SIZE] = {};
+    SECItem ivitem = { .data = iv, .len = sizeof(iv), .type = siBuffer };
+    SECItem ctitem = { .data = ct, .len = sizeof(ct), .type = siBuffer };
+    PK11SymKey *ekey = NULL;
+    PK11SymKey *skey = NULL;
+
+    /* Try to import the key directly. */
+    skey = PK11_ImportSymKey(slot, mech, PK11_OriginUnwrap,
+                             CKA_SIGN, key, NULL);
+    if (skey)
+        return skey;
+
+    /* If we get here, we are probably in FIPS mode. Let's encrypt the key so
+     * that we can unseal it instead of loading it directly. */
+
+    /* Generate an ephemeral key. */
+    ekey = PK11_TokenKeyGenWithFlags(slot, CKM_AES_CBC_PAD, NULL,
+                                     AES_128_KEY_LENGTH, NULL,
+                                     CKF_ENCRYPT | CKF_UNWRAP,
+                                     PK11_ATTR_SESSION |
+                                     PK11_ATTR_PRIVATE |
+                                     PK11_ATTR_SENSITIVE, NULL);
+    if (!ekey)
+        goto egress;
+
+    /* Encrypt the input key. */
+    if (PK11_Encrypt(ekey, CKM_AES_CBC_PAD, &ivitem, ctitem.data, &ctitem.len,
+                     ctitem.len, key->data, key->len) != SECSuccess)
+        goto egress;
+
+    /* Unwrap the input key. */
+    skey = PK11_UnwrapSymKey(ekey, CKM_AES_CBC_PAD, &ivitem,
+                             &ctitem, mech, CKA_SIGN, key->len);
+
+egress:
+    PK11_FreeSymKey(ekey);
+    return skey;
+}
 
 /*
  * This code is mostly cargo-cult taken from here:
@@ -90,8 +134,7 @@ static bool hmac(SECItem *key, CK_MECHANISM_TYPE mech, const SECItem *in,
         }
     }
 
-    symkey = PK11_ImportSymKey(slot, mech, PK11_OriginUnwrap,
-                               CKA_SIGN, key, NULL);
+    symkey = import_key(slot, mech, key);
     if (symkey == NULL)
         goto done;
 
--- a/ipaserver/plugins/baseuser.py
+++ b/ipaserver/plugins/baseuser.py
@@ -31,7 +31,6 @@ from .baseldap import (
     LDAPAddAttributeViaOption, LDAPRemoveAttributeViaOption,
     add_missing_object_class)
 from ipaserver.plugins.service import (validate_realm, normalize_principal)
-from ipaserver.plugins.config import check_fips_auth_opts
 from ipalib.request import context
 from ipalib import _
 from ipalib.constants import PATTERN_GROUPUSER_NAME
@@ -481,7 +480,6 @@ class baseuser_add(LDAPCreate):
                             **options):
         assert isinstance(dn, DN)
         set_krbcanonicalname(entry_attrs)
-        check_fips_auth_opts(fips_mode=self.api.env.fips_mode, **options)
         self.obj.convert_usercertificate_pre(entry_attrs)
 
     def post_common_callback(self, ldap, dn, entry_attrs, *keys, **options):
@@ -605,7 +603,6 @@ class baseuser_mod(LDAPUpdate):
         assert isinstance(dn, DN)
         add_sshpubkey_to_attrs_pre(self.context, attrs_list)
 
-        check_fips_auth_opts(fips_mode=self.api.env.fips_mode, **options)
         self.check_namelength(ldap, **options)
 
         self.check_mail(entry_attrs)
--- a/ipaserver/plugins/config.py
+++ b/ipaserver/plugins/config.py
@@ -85,20 +85,6 @@ EXAMPLES:
 
 register = Registry()
 
-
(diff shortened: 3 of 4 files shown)


npmccallum:
@frozencemetery FYI

frozencemetery:
Thank you for including the log message :)

Tested a variety of tokens in FIPS mode.  Please note: this will work for all new OTPs.  It will also work for existing OTPs provided that they were not sha512 or sha384 (the default is sha1, so this probably won't affect many users).

daemons/ipa-slapi-plugins/libotp/hotp.c
```diff
@@ -66,6 +67,49 @@ static const struct {
...
+
+    /* Generate an ephemeral key. */
+    ekey = PK11_TokenKeyGenWithFlags(slot, CKM_AES_CBC_PAD, NULL,
+                                     AES_128_KEY_LENGTH, NULL,
+                                     CKF_ENCRYPT | CKF_UNWRAP,
+                                     PK11_ATTR_SESSION |
+                                     PK11_ATTR_PRIVATE |
+                                     PK11_ATTR_SENSITIVE, NULL);
+    if (!ekey)
+        goto egress;
```
rcritten: There should be a log message so we know it failed here along with the NSS error code.

daemons/ipa-slapi-plugins/libotp/hotp.c
```diff
@@ -66,6 +67,49 @@ static const struct {
...
+
+    /* Encrypt the input key. */
+    if (PK11_Encrypt(ekey, CKM_AES_CBC_PAD, &ivitem, ctitem.data, &ctitem.len,
+                     ctitem.len, key->data, key->len) != SECSuccess)
+        goto egress;
```
rcritten: There should be a log message so we know it failed here along with the NSS error code.

daemons/ipa-slapi-plugins/libotp/hotp.c
```diff
@@ -66,6 +67,49 @@ static const struct {
...
+
+    /* Unwrap the input key. */
+    skey = PK11_UnwrapSymKey(ekey, CKM_AES_CBC_PAD, &ivitem,
+                             &ctitem, mech, CKA_SIGN, key->len);
+
```
rcritten: What happens if the unwrap fails?

daemons/ipa-slapi-plugins/libotp/hotp.c
```diff
@@ -66,6 +67,49 @@ static const struct {
...
+egress:
+    PK11_FreeSymKey(ekey);
```
rcritten: If in the case where creating the encryption key fails will PK11_FreeSymKey() blow up if ekey is NULL?

daemons/ipa-slapi-plugins/libotp/hotp.c
npmccallum: The NSS error code is useless (SECFailure). You have to use GDB. This is also unlikely to fail except for OOM.

daemons/ipa-slapi-plugins/libotp/hotp.c
npmccallum: The NSS error code is useless (SECFailure). You have to use GDB. This is also unlikely to fail except for OOM.

daemons/ipa-slapi-plugins/libotp/hotp.c
npmccallum: Then you break out GDB.

daemons/ipa-slapi-plugins/libotp/hotp.c
npmccallum:
```
180	PK11_FreeSymKey(PK11SymKey *symKey)
181	{
182	    PK11SlotInfo *slot;
183	    PRBool freeit = PR_TRUE;
184	
185	    if (!symKey) {
186	        return;
187	    }
```

daemons/ipa-slapi-plugins/libotp/hotp.c
rcritten: PR_GetError()

daemons/ipa-slapi-plugins/libotp/hotp.c
rcritten: Shouldn't it at least be checked?  static analyzers are sure to whing.

daemons/ipa-slapi-plugins/libotp/hotp.c
rcritten: nice.

mkosek:
@rcritten, this is the Bugzilla with the context of this work:
https://bugzilla.redhat.com/show_bug.cgi?id=1544679
Merging Nathaniel's work should enable us to also revert the changes that were done in https://pagure.io/freeipa/issue/7168

tiran:
CI is failing:

```
ipa_kdb.c:639:13: warning: ‘ipadb_free’ defined but not used [-Wunused-function]
 static void ipadb_free(krb5_context context, void *ptr)
             ^~~~~~~~~~
ipa_kdb.c:634:14: warning: ‘ipadb_alloc’ defined but not used [-Wunused-function]
 static void *ipadb_alloc(krb5_context context, void *ptr, size_t size)
              ^~~~~~~~~~~
```

tiran:
No, that was just a red herring. The actual error is:

```
./.libs/libhotp.a(hotp.o): In function `hmac':
/freeipa/rpmbuild/BUILD/freeipa-4.6.90.dev201802221933+gitca42e8e/daemons/ipa-slapi-plugins/libotp/hotp.c:146: undefined reference to `slapi_log_error'
collect2: error: ld returned 1 exit status
```

pvoborni:
https://pagure.io/freeipa/issue/7243 seems to me like a good fit for these patches.

daemons/ipa-slapi-plugins/libotp/hotp.c
npmccallum: I'm not sure what the problem is here. If an error occurs in `import_key()`, we return `NULL`. This is true for all code paths. The `skey` variable is initialized to `NULL`, so static analyzers shouldn't complain.

daemons/ipa-slapi-plugins/libotp/hotp.c
npmccallum: It appears we can't log from this code. See the missing symbol error in the CI. I'm going to remove the patch which adds logging (at least temporarily).

npmccallum:
I've pushed a new set of patches, including reverting https://pagure.io/freeipa/issue/7168. It is very unfortunate that we can't log in this code path. But perhaps @frozencemetery can tackle this in a separate patch?

frozencemetery:
Is logging a blocker for merging this?  (Asking for my own prioritization.)

npmccallum:
IMHO, no.

npmccallum:
I pushed a new version. It only changes the text wrapping on the commit messages.

rcritten:
IMHO yes. The urgency is now so the patch should be made complete now. Pushing things off is usually equivalent to never doing them.

The problem if any of these fail is that we'll have no idea what at all has happened, things will just not work. We won't necessarily have a reproduction so what, we (or GSS) try to walk the user through a gdb session? Why would we even try that? What would point in that direction? If there is a log somewhere saying some call failed along with the NSS return code we at least get an inkling of where it is failing and hopefully why.

The final call, skey =..., never has the return value checked. So skey could be NULL and only the caller would find out. Is the caller expecting a NULL?

npmccallum:
Regarding the unchecked `skey`, yes the caller expects to receive NULL. See:
https://github.com/freeipa/freeipa/pull/1621/commits/a47b66a4ada7c28c59eb7085c40bbd621141389f#diff-21d729f4715e316ae0f969b07fb0ecbaR138

npmccallum:
Regarding logging, I see only one call that is likely to fail. I added logging for it, but then realized that this code doesn't actually link against 389ds and thus can't log (see the CI failure at the top of this PR). I removed that patch from this PR since it caused CI failures. All of the other calls are extremely likely to succeed and no logging exists for the existing crypto code in this file. Thus, I consider such logging an enhancement.

If FreeIPA maintainers disagree, then we need to do some more invasive changes to allow this code to support logging.

rcritten:
Can it log to syslog? That should be relatively straightforward. I know I have a difficult time debugging OTP in general and this is a black box without gdb.

npmccallum:
I honestly don't know the answer to that. Maybe @frozencemetery knows?

frozencemetery:
"Yes", it can.  It's a bit gross though - there's no state, so we never call `openlog()/closelog()`.  Messages end up in syslog like: `mar 01 14:15:05 freeipa.dbsplat.biz ns-slapd[1843]: message text`.

See the top commit on my branch over here: https://github.com/frozencemetery/freeipa/tree/logging

mkosek:
The added logs by @frozencemetery look promising. But let me CC here @tbordaz in case he has any advice for the FreeIPA DS plugin logging.

tbordaz:
I agree that currently libotp can not log. One possibility could be to change an libotp init function (otp_config_init ?) to pass a callback that will call slapi_log_error. Then libotp will use this callback.

frozencemetery:
Since this seems to be blocked on logging, what approach does the project want to see?  The raw `syslog()` I implemented?  API change?  Something else?

frozencemetery:
Bump - how does logging need to work for this to be merged?

rcritten:
Your syslog patch works for me.

I'm inclined to ack, will wait for others to comment before acking and pushing COB Mar 7.

frozencemetery:
Sorry to keep bumping, but can this be pushed, then?

tbordaz:
Using syslog() is good to me and avoid doing a complex fix to provide a logging function.
So I also ack this option

rcritten:
master:

* c9c58f2d35dfb8a6d1e187d260013812dcfdfd90 Fix OTP validation in FIPS mode
* a01a24ce5aded934095ea428d539ca30850f72e3 Increase the default token key size
* d498d7272d8de3e24afc442a3b001518fd98ebff Revert "Don't allow OTP or RADIUS in FIPS mode"

rcritten:
A manual backport to the ipa-4-5 branch is going to be required.

PR #1622, opened 2018-02-22
Don't return None on mismatched interactive passwords

rcritten:
This will cause the command to continue with no password set
at all which is not what we want.

We want to loop forever until the passwords match or the
user gives up and types ^D or ^C.

https://pagure.io/freeipa/issue/7383

Note that automated testing of this is difficult to impossible because the client code detects when data is being passed in via a pipe. so we can't pass in the mismatch passwords via stdin_text, it would have to be some much more complicated I fear.

### DIFF
--- a/ipalib/cli.py
+++ b/ipalib/cli.py
@@ -627,7 +627,6 @@ class textui(backend.Backend):
                     return pw1
                 else:
                     self.print_error(_('Passwords do not match!'))
-                    return None
         else:
             return self.decode(sys.stdin.readline().strip())
 


stlaz:
This would still return `None`.

tiran:
The ``while True`` loop will loop over until both passwords are equal.

stlaz:
Ah, you're right, I did not notice the loop. We should set max number of tries, then.

tiran:
I'd rather tell the user how to break out of the password prompt loop (CTRL+C).

tiran:
master:

* 0aaee0a97f5e26903474da10538e464a9f32dba7 Don't return None on mismatched interactive passwords

PR #1635, opened 2018-02-26
Encrypt httpd key stored on disk

stlaz:
This commit adds configuration for HTTPD to encrypt/decrypt its
key which we currently store in clear on the disc.

A password-reading script is added for mod_ssl. This script is
extensible for the future use of directory server with the
expectation that key encryption/decription will be handled
similarly by its configuration.

https://pagure.io/freeipa/issue/7421


install/tools/Makefile.am
```diff
@@ -37,4 +37,6 @@ dist_app_SCRIPTS =		\
 	ipa-custodia-check	\
 	ipa-httpd-kdcproxy	\
 	ipa-pki-retrieve-key	\
+	ipa-httppswd.sh		\
+	ipa-pwdreader.sh	\
```
tiran: Why did you add two scripts? It looks like an unnecessary complicated to me.

install/tools/ipa-pwdreader.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+HTTP_PASSWD_LOC="/var/lib/ipa/certs/httpd_passwd.txt"
+
+if [ "$(basename $0)" == "ipa-httppswd.sh" ] && \
+    [ -f "$HTTP_PASSWD_LOC" ]; then
+    cat "$HTTP_PASSWD_LOC"
```
tiran: [SSLPassPhraseDialog](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslpassphrasedialog) has context ``server_config``. It means that mod_ssl has only one option for the entire process. Although it is not supported, users tend to deploy additional services and virtual hosts on the same machine. Therefore this script must only return the httpd password for IPA's vhost.

install/tools/Makefile.am
stlaz: Explained in the commit message, one symlinks to the other.

install/tools/ipa-pwdreader.sh
stlaz: It does.

tiran: How is this suppose to work? There can only be one command for all vhosts. You have to check argv 1 and 2 to detect the correct vhost.

stlaz: Ah, the response only applies for the "Therefore this script must only return the httpd password for IPA's vhost."
I did not consider support for more because we did not support anything else running alongside IPA in the HTTPD service.

rcritten:
I'd drop the .sh extension. The other Apache password helpers don't include it so we should maintain the convention.

There only needs to be one script that looks like:
```
#!/bin/sh

case $1 in
    ipa.example.com:443) cat /var/lib/ipa/certs/httpd_passwd.txt;;

    # Other virtual hosts can be added by including hostname:port) echo $PASS or cat a file.
    # www.somewhere.com:443) echo "SecretPassword123";;
esac

exit 0
```

tiran:
Please take argument ``$2`` into account, too.

For SELinux we probably need to use a different directory for the pass phrases. We shouldn't allow the helper app to access certmonger's cert context.

rcritten:
Sure, you could do something like this to handle the other key types:
```
#!/bin/sh

case $2 in
    RSA) case $1 in
             ipa.example.com:443) cat /var/lib/ipa/certs/rsa_passwd.txt;;
         esac;;
    ECC) case $1 in
             ipa.example.com:443) cat /var/lib/ipa/certs/ecc_passwd.txt;;
        esac;;
    DSA) case $1 in
             ipa.example.com:443) cat /var/lib/ipa/certs/dsa_passwd.txt;;
        esac;;

esac

exit 0
```

Note I don't care what the filename is, I just used a different one for illustration purposes.

Good point about certmonger, that is a bit of a nasty problem. Seems like maybe certmonger could use the cojntents of the file and not the file directly. So when we create the tracking we read the file and pass that in as a raw PIN instead of as a file path to the file.

stlaz:
I updated the PR to reflect the comments here.

@tiran I am not sure what you mean by "certmonger SELinux context", the certificates are not stored within it nor have the context themselves:
```bash
$ ls -laZ /var/lib/ipa/certs/
total 20
drwxr-xr-x.  2 root root system_u:object_r:ipa_var_lib_t:s0     4096 Feb 28 08:40 .
drwxr-xr-x. 10 root root system_u:object_r:ipa_var_lib_t:s0     4096 Feb 28 08:27 ..
-rw-------.  1 root root system_u:object_r:ipa_var_lib_t:s0     2009 Feb 28 08:29 httpd.crt
-rw-------.  1 root root system_u:object_r:ipa_var_lib_t:s0     1874 Feb 28 08:29 httpd.key
-rw-------.  1 root root unconfined_u:object_r:ipa_var_lib_t:s0   41 Feb 28 08:25 vm-124.abc.idm.lab.eng.brq.redhat.com-443-RSA
```
Should we have a different context for the data here?

tiran:
@stlaz Are these SELinux contexts correct? AFAIK the F28 SELinux policy hasn't been modified for mod_ssl certs yet. Do you see AVCs?

stlaz:
@tiran The only AVC I see is the obvious:
```
time->Tue Feb 27 10:38:41 2018
type=AVC msg=audit(1519724321.598:389010): avc:  denied  { open } for  pid=373 comm="httpd" path="/usr/sbin/ipa-pwdreader" dev="dm-0" ino=145038 scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:object_r:nfs_t:s0 tclass=file permissive=1
```

stlaz:
So with the current version of F27 there were no AVCs (as you can see, the pwdreader script seem to have been mislabeled previously - `tcontext=system_u:object_r:nfs_t:s0`).

In this latest modification, I created 3 directories, one as a password storage, one is a cert storage, and one's a storage for the private keys.

tiran:
Awesome, good work @stlaz ! Let's try this. At worst there is a hidden *don't audit* rule that will prevent installation.

By the way I don't see any code that removes the password, key or cert file. Is Certmonger removing the files? In case it's leaving the files behind, please address the issue in another PR. I want to get this PR landed ASAP so QE is able to test it.

stlaz:
I already tried the installation and it works. The removal of key and cert was already handled in the uninstallation method, I just forgot about the password file there, thanks for reminding me.
I see there's some conflicts now, will resolve right away.

stlaz:
Should be alright now, hopefully.

tiran:
I'll give ACK after CI has passed.

stlaz:
tox seems to be failing for reasons not necessarily caused by this PR.

tiran:
master:

* 7cbd9bd429c97d335471b1c739f6d6d63c83c35d Encrypt httpd key stored on disk

PR #1636, opened 2018-02-26
Upgrade HTTPd's alias db to SQL

tiran:
https://pagure.io/freeipa/issue/7354

### DIFF
--- a/ipaserver/install/certs.py
+++ b/ipaserver/install/certs.py
@@ -687,22 +687,6 @@ class CertDB(object):
 
         return is_ipa_issued_cert(api, cert)
 
-    def needs_upgrade_format(self):
-        """Check if NSSDB file format needs upgrade
-
-        Only upgrade if it's an existing dbm database and default
-        database type is no 'dbm'.
-        """
-        return (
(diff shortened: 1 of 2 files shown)


rcritten:
Can you expand on the ticket and/or the commit message?

As I understand it this will have a very narrow focus: F-28 only until 4.7 ships and we switch to mod_ssl as the default crypto engine. So this really just applies to the ipa-4-6 branch right?

If not then this needs to be enhanced to recognize that the mod_ssl conversion is already done and not try to migrate the database.

rcritten:
The installation failed for me. Server-Cert was added to cert9.db but cert8.db still exists and the installer tries to look there for the cert.

The certmonger tracking doesn't explicitly have the NSS database directory prefix, not sure if that is on purpose or not.

rcritten:
Is this patch still needed since 4.6.0.pre1 was released to F28?

tiran:
After a brief internal discussion with @rcritten, I decided to close and reject this PR. The update from F27 / FreeIPA 4.6 to F28 / FreeIPA 4.7 is working without this patch. Let's not complicate the code and add another potential source of regression.

PR #1637, opened 2018-02-26
Don't try to backup CS.cfg during upgrade if CA is not configured

rcritten:
https://pagure.io/freeipa/issue/7409


frasertweedale:
Does the trick.  ACK.

rcritten:
master:

* 95a45a2b0942a9ac38d5418b23821f7da1ce28a3 Don't try to backup CS.cfg during upgrade if CA is not configured

PR #1640, opened 2018-02-28
Return a value if exceptions are raised in server uninstall

rcritten:
The server uninstall relies on the Continuous class which will
not re-raise exceptions, only log them.

The AdminTool class purports to "call sys.exit() with the return
value" but most of the run implementations returned no value, or
the methods they called returned nothing so there was nothing to
return, so this was a no-op.

The fix is to capture and bubble up the return values and to
return 1 if any exceptions are caught in Continuous, the exception
being SystemExit which is what is raised when sys.exit() is called.

This potentially affects other users in that when executing the
steps of an installer or uninstaller the highest return code
will be the exit value of that installer.

https://pagure.io/freeipa/issue/7330

### DIFF
--- a/ipapython/install/common.py
+++ b/ipapython/install/common.py
@@ -7,12 +7,11 @@ Common stuff.
 """
 
 import logging
-import traceback
 
 from . import core
 from .util import from_
 
-__all__ = ['step', 'Installable', 'Interactive', 'Continuous', 'installer',
+__all__ = ['step', 'Installable', 'Interactive', 'installer',
            'uninstaller']
 
 logger = logging.getLogger(__name__)
@@ -86,16 +85,6 @@ class Step(Installable):
 
 class Interactive(core.Configurable):
     interactive = core.Property(False)
-
-
-class Continuous(core.Configurable):
-    def _handle_execute_exception(self, exc_info):
-        try:
-            super(Continuous, self)._handle_execute_exception(exc_info)
-        except BaseException as e:
-            logger.debug("%s", traceback.format_exc())
-            if isinstance(e, Exception):
-                logger.error("%s", e)
 
 
 def installer(cls):
--- a/ipapython/install/common.py
+++ b/ipapython/install/common.py
@@ -109,7 +109,7 @@ def installer(cls):
 
 
 def uninstaller(cls):
-    class Uninstaller(Continuous, cls, Installable):
+    class Uninstaller(cls, Installable):
         def __init__(self, **kwargs):
             super(Uninstaller, self).__init__(uninstalling=True,
                                               **kwargs)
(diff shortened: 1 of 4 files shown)


ipapython/install/common.py
```diff
@@ -89,13 +89,24 @@ class Interactive(core.Configurable):
...
 class Continuous(core.Configurable):
+    def __init__(self, **kwargs):
+        super(Continuous, self).__init__(**kwargs)
+        self.exceptions_handled = 0
+
     def _handle_execute_exception(self, exc_info):
         try:
             super(Continuous, self)._handle_execute_exception(exc_info)
+        except SystemExit:
+            pass
```
stlaz: Please don't.

ipapython/install/common.py
```diff
@@ -89,13 +89,24 @@ class Interactive(core.Configurable):
...
         except BaseException as e:
             logger.debug("%s", traceback.format_exc())
             if isinstance(e, Exception):
                 logger.error("%s", e)
+            self.exceptions_handled = 1
+
+    def run(self):
+        super(Continuous, self).run()
+        return self.exceptions_handled
```
stlaz: That's fine for developers but won't values not in (0,1) confuse users?

ipapython/install/common.py
```diff
@@ -89,13 +89,24 @@ class Interactive(core.Configurable):
...
 
 
 def installer(cls):
```
stlaz: Why is uninstaller a subclass of the `Continuous` class anyway?

ipapython/install/common.py
rcritten: I could find no other way because of the way this is architected. sys.exit() is an exception unto itself if not caught will result in a nasty traceback.

ipapython/install/common.py
rcritten: I see your point. Sadly there is near zero documentation on how these classes are expected to work so I'm trying to tread gently. It is my assumption based on the class name Continuous and the fact that exceptions aren't fatal that this is expected to soldier in in most cases of failures so returning that number seems a reasonable thing to return. I'm not necessarily opposed to returning 0 or 1.

ipapython/install/common.py
rcritten: Yeah, exactly. Who knows. I can find no design discussion on the installer refactoring, the git commits are terse at best and there is near zero inline documentation.

My working assumption, as mentioned, is that it is because Continuous eats exceptions and expects the uninstaller to keep trying to uninstall stuff. It doesn't seem to do that in practice but that is outside the scope of what I was trying to fix.

tiran:
It sounds like``Continuous`` doesn't work as advertised. What happens if you remove ``Continuous`` from the list of parent classes?

stlaz:
Sorry, I forgot about this PR, I'll review it tomorrow. The removal of the `Continuous` class inheritance from uninstaller seems alright.

tiran:
master:

* 68c7b036893fcdad7cc2364b0fc2a841366493ef Return a value if exceptions are raised in server uninstall
* 64fca87a52e361718b46aaec8d1d5eea0f5ccdbc Remove the Continuous installer class, it is unused

PR #1646, opened 2018-03-01
Fixing cleanup process in test_caless

felipevolpone:
After commit bbe615e, if the uninstall process fails (in the test cleanup) the error is not hidden anymore.
That brought light to errors in the cleanup process on `TestReplicaInstall` test, like this:
```
RUN ['ipa-server-install', '--uninstall', '-U']
ipapython.admintool: ERROR    Server removal aborted:
Replication topology in suffix 'domain' is disconnected:
Topology does not allow server master.ipa.test to replicate with servers:
    replica0.ipa.test.
ipapython.admintool: ERROR    The ipa-server-install command failed
```

This commit changes the order of how a replica should be removed from the topology.

Other errors can be checked [here](https://fedorapeople.org/groups/freeipa/prci/jobs/1b27ac12-1bfe-11e8-9b66-fa163e97f492/).

### DIFF
--- a/ipatests/test_integration/test_caless.py
+++ b/ipatests/test_integration/test_caless.py
@@ -105,15 +105,14 @@ def replica_install_teardown(func):
             replica = args[0].replicas[0]
             master = args[0].master
             tasks.kinit_admin(master)
+            tasks.clean_replication_agreement(master, replica, cleanup=True,
+                                              raiseonerr=False)
+            master.run_command(['ipa', 'host-del', replica.hostname],
+                               raiseonerr=False)
             tasks.uninstall_master(replica, clean=False)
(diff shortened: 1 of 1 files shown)


abbra:
Do you still need the temporary commit?

felipevolpone:
@abbra The temporary commit shows that my PR actually fix the test (fedora-27/TestReplicaInstall). We don't run this test in PR CI nowadays because it takes a lot of time to run, so once the PR gets acked I'll remove the temporary commit.

pvoborni:
Change looks good to me. Remove the temp commit and I'll ack it.

tiran:
master:

* 2c05e42af6a8c7bccc1726bf05cf14be451a08fc Fixing cleanup process in test_caless

PR #1650, opened 2018-03-05
Remove unnecessary code form test_backup_restore

akokshar:
Check_admin_in_cli can keep only data necessary to vefiry
command result.

The other thing this code is not fully correct is that it
is breacking 'command' abstraction by setting property value
controlled by the 'command' object itself. The class instance
have to be as much independant and self-sufficient as
possible.

### DIFF
--- a/ipatests/test_integration/test_backup_and_restore.py
+++ b/ipatests/test_integration/test_backup_and_restore.py
@@ -73,11 +73,9 @@ def check_admin_in_cli(host):
     # that, the code bellow order the 'Member of groups' field to able to
     # assert it latter.
     data = dict(re.findall("\W*(.+):\W*(.+)\W*", result.stdout_text))
-    data["Member of groups"] = ', '.join(sorted(data["Member of groups"]
-                                                .split(", ")))
-    result.stdout_text = ''.join([' {}: {}\n'.format(k, v)
-                                  for k, v in data.items()])
-    return result
+    data["Member of groups"] = sorted(data["Member of groups"].split(', '))
+
+    return data
 
 
 def check_admin_in_id(host):
@@ -113,7 +111,7 @@ def check_custodia_files(host):
 
 CHECKS = [
     (check_admin_in_ldap, assert_entries_equal),
-    (check_admin_in_cli, assert_results_equal),
+    (check_admin_in_cli, assert_deepequal),
     (check_admin_in_id, assert_results_equal),
     (check_certs, assert_results_equal),
     (check_dns, assert_results_equal),


flo-renaud:
Hi @akokshar 
the test backup_restore is taking more than 1h to run and reaches a PR-CI timeout. You may consider splitting it in multiple tests to ensure it has enough time to finish.

pvoborni:
I don't understand what problem this PR solves. Return value of `check_admin_in_cli`  is processed by:

```
def assert_results_equal(a, b):
    def to_dict(r):
        return {
            'stdout': r.stdout_text,
            'stderr': r.stderr_text,
            'returncode': r.returncode,
        }
    assert_deepequal(to_dict(a), to_dict(b))
```

Which expect data with `stdout_text`, `stderr_text`, `returncode` attributes. Your change returns data dict which doesn't have those attributes. So seems to me that it should fail with `AttributeError: 'dict' object has no attribute 'stdout_text'`. Also, values under `Member of groups` key won't be used.

I checked nightly runs of backup and restore tests and they seem to pass.  

I could agree that parsing text, sorting and then joining might not be the best way but doesn't seem as something which needs fixing. 

I guess this PR might work  if just `assert_deepequal` would be used for processing the result instead of `assert_results_equal`.

akokshar:
you are right - I did forget to change assert function. If only we had well
working prci, I would discover this immediately.
Will update this PR.

Alexander

akokshar:
The problem this pr is intended to solve was discovered by QA some time ago - we suffered from issues caused by pytest-multihost and this code prevented us to use alternative tools.

The test is checking command output captured before backup with same command output captured after restore. The output is the only what have to be checked, so why do we store whole host.run_command object?
We do modify data (stdout_text) which is controlled by run_command object. Why do we need to delve into run_command object's guts which is an abstraction on remote command execution? I think it is good to keep objects as much independent as possible.

Exactly this attempt to set internal property on run_command prevented QA to use something else instead of pytest_multihost. This was an origin of that PR. Feel free to cancel it if you think it is not required at the moment.

pvoborni:
> If only we had well working prci, I would discover this immediately.

From this PR POV it works ok. The job just timed out.  Flo wrote what is required to do in order to run it properly. Another alternative is to increase timeout of the Job. How to do it was described by Felipe recently. Pasting the snipped here:

----
> 2. How can I add a test on my PR that is not being run by PR CI yet?

> If you would like to add a test in your PR that PR CI is not running
> on every PR, you just need to hack the .freeipa-pr-ci.yaml file. As an
> example, imagine that you would like to put the
> test_integration/test_http_kdc_proxy.py test to run.

> You just need to add this to the .freeipa-pr-ci.yaml file in your PR:

```yaml
fedora-27/test_http_kdc_proxy:
  requires: [fedora-27/build]
  priority: 50
  job:
    class: RunPytest
    args:
      build_url: '{fedora-27/build_url}'
      test_suite: test_integration/test_http_kdc_proxy.py
      template: *ci-master-f27
      timeout: 3600
```

> You may want to change the timeout value, because some tests takes more
> time to run. It's important to keep in mind that it may be needed to
> change the topology field as well (if the test needs more replicas
> or so on). If you check the nightly PRs, you can see more topologies
> there.

---

> The problem this pr is intended to solve was discovered by QA some time ago - we suffered from issues caused by pytest-multihost and this code prevented us to use alternative tools.
>
> The test is checking command output captured before backup with same command output captured after restore. The output is the only what have to be checked, so why do we store whole host.run_command object?
> We do modify data (stdout_text) which is controlled by run_command object. Why do we need to delve into run_command object's guts which is an abstraction on remote command execution? I think it is good to keep objects as much independent as possible.

I see, good point. But in that case this patch won't solve it. Because the same issues are in other checks:
```   
    (check_admin_in_cli, assert_results_equal),
    (check_admin_in_id, assert_results_equal),
    (check_certs, assert_results_equal),
    (check_dns, assert_results_equal),
    (check_kinit, assert_results_equal),
```

And maybe something similar is in other tests.
 
> Exactly this attempt to set internal property on run_command prevented QA to use something else instead of pytest_multihost. This was an origin of that PR.

> Feel free to cancel it if you think it is not required at the moment.

Up to you.

rcritten:
Closing due to inactivity.

PR #1653, opened 2018-03-06
vault: fix vault-retrieve to a file

stlaz:
`data` is bytes but we were opening the "--out" file as
a text.

https://pagure.io/freeipa/issue/7430

### DIFF
--- a/ipaclient/plugins/vault.py
+++ b/ipaclient/plugins/vault.py
@@ -1143,7 +1143,7 @@ class vault_retrieve(ModVaultData):
                 error=_('Invalid vault type'))
 
         if output_file:
-            with open(output_file, 'w') as f:
+            with open(output_file, 'wb') as f:
                 f.write(data)
 
         else:


abbra:
LGTM. For the record, this is how you can reproduce the issue:
```python
$ python3
Python 3.6.4 (default, Feb  8 2018, 14:42:51) 
[GCC 7.3.1 20180130 (Red Hat 7.3.1-2)] on linux
>>> c = b'oiuroingoirwhjoijrwt'
>>> with open("/tmp/myfile", "w") as f:
...     f.write(c)
... 
  File "<stdin>", line 2, in <module>
TypeError: write() argument must be str, not bytes
>>> with open("/tmp/myfile", "wb") as f:
...     f.write(c)
... 
20
>>> 
```

Using python2 both "w" and "wb" produce no difference when writing bytes object.

stlaz:
master:

* c9c41d2d9077202f68c16db53f532da1b7055c10 vault: fix vault-retrieve to a file

PR #1658, opened 2018-03-07
Misleading error message XXX-del for non-existent objects

amitkumar50:
ipa ca-del does not provide relevant log message when
ipa-server user tries deleting non-exitant CA.

`# ipa ca-del SubCA_9 SubCA_10 --continue`
`-------------`
Deleted CA ""
`-------------`
  Failed to remove: SubCA_9, SubCA_10

--continue option itself says Don't stop on errors.
So it will try deleting CAs anyhow.

I believe better would be to change text from
Deleted CA
to something:
Trying to Delete CA

Resolves: https://pagure.io/freeipa/issue/6106

### DIFF
--- a/ipaserver/plugins/ca.py
+++ b/ipaserver/plugins/ca.py
@@ -292,7 +292,7 @@ class ca_add(LDAPCreate):
 class ca_del(LDAPDelete):
     __doc__ = _('Delete a CA.')
 
-    msg_summary = _('Deleted CA "%(value)s"')
+    msg_summary = _('Trying to Delete CA "%(value)s"')
 
     def pre_callback(self, ldap, dn, *keys, **options):
         ca_enabled_check(self.api)


flo-renaud:
Hi @amitkumar50 

Travis-CI tests failed (TestCAbasicCRUD.test_delete). You would need to update the expected output in the test suite according to the changes you did in the ca-del command.

amitkumar50:
@flo-renaud shall i change this:
./ipatests/test_xmlrpc/tracker/ca_plugin.py
summary=u'Deleted CA "{}"'.format(self.name)
to
summary=u'Trying to Delete CA "{}"'.format(self.name)

flo-renaud:
Hi @amitkumar50 
you can perform the same logic as user-del, for instance look at https://pagure.io/freeipa/blob/master/f/ipaserver/plugins/user.py#_733:

```
$ ipa user-del m2 m3 --continue
-----------------
Deleted user "m3"
-----------------
  Failed to remove: m2
```

This means that your code needs to track 2 lists, one for the ids that could not be deleted and one for the ids that were successfully deleted, and return these lists in the execute() method in a similar way to:
```
            val = dict(result=dict(failed=failed), value=deleted)
            return val
```

If you want to run the xmlrpc tests for the ca plugin locally before submitting your PR, you will find instructions [here](https://www.freeipa.org/page/Testing#Full_test_run). You can use `./make-test ipatests/test_xmlrpc/test_ca_plugin.py` to run only the relevant test.

amitkumar50:
@flo-renaud Feature of ca-del is behaving properly just, message on NULL subCA cert is misleading.
That's why I beileve correction in message would be good.. Your thoughts..

flo-renaud:
Hi @amitkumar50 
I think we should have a consistent behavior for all commands providing the --continue option, and the current behavior is to report the ids for which the command succeeded and the ids for which the command failed.
You can ask for a second opinion on this if you disagree with me, no problem for me :) The mailing list freeipa-devel@lists.fedorahosted.org is a good place to start the discussion.

pvoborni:
When I search for `msg_summary = _('Deleted ` it returns me 49 matches. To do a consisitent change all these would need to be changed along with all releated tests. 

The original ticket suggeted to show something like: 
  `No Sub CA deleted` or `0 CA deleted`

But that doesn't seem right either.  To do such change LDAPDelete class would need to return artifical value as it is just pasted in summary in CLI. 
E.g. raw response from server looks like:
```json
    "result": {
        "result": {
            "failed": [
                "fbar",
                "fbar2",
                "fbar3"
            ]
        },
        "summary": "Deleted user \"\"",
        "value": []
    },
```

Also, there coud be a CA or other object named `No` or `0`.  People might also parse the results of the command automatically, changing it would break it or would make it harder.

Current situation is not the best, but proposed solution doesn't seem  better to me. Issue I find with msg `trying to delete` or `deleting` is that is implies and action which is happening but doesn't show a result. So e.g. a successfull delete of user wihout `--continue` would look like:

```
ipa user-del fbar3
-------------------------------------
Trying to Delete  user "fbar3"
--------------------------------------
```
The same for other objects. As a user I would probably be confused.

I'm thinking it might be better to close the Pagure issues as won't fix.

flo-renaud:
My bad, I though that ca-del was behaving badly and not showing the 2 lists. It is indeed showing the list of failed identities, and the list of successfully deleted identities.
So I agree with @pvoborni, changing the output from `Deleted CA ""` with an empty list to `No CA deleted` would break backward compatibility, and I'd rather close the ticket as won't fix.

rcritten:
Per above comments closing.

PR #1659, opened 2018-03-07
Do not check deleted files with `make fastlint`

Tiboris:
When any file from FreeIPA tree has been deleted there was
a failure like:
```
 pylint
 ------
 ************* Module ipaserver/install/ntpinstance.py
 ipaserver/install/ntpinstance.py:1: [F0001(fatal), ] No module named ipaserver/install/ntpinstance.py)
```
Adding --diff-filter to fastlint will not list deleted files
in git diff --names-only output to not include not existing
files to checklist.


stlaz:
Works great, thanks. ACK.

Tiboris:
master:

* 2b17a086d0cfa3be5e31fecfe7650004cb3de0eb Do not check deleted files with `make fastlint`

PR #1664, opened 2018-03-09
ipa-server-install: handle error when calling kdb5_util create

flo-renaud:
ipa-server-install creates the kerberos container by calling
kdb5_util create -s -r $REALM -x ipa-setup-override-restrictions

but does not react on failure of this command. The installer fails later
when trying to create a ldap principal, and it is difficult to diagnose the
root cause.

The fix raises a RuntimeException when kdb5_util fails, to make sure
that the installer exits immediately with a proper error message.

Note: no test added because there is no easy reproducer. One would need to
stop dirsrv just before calling kdb5_util to simulate a failure.

https://pagure.io/freeipa/issue/7438

### DIFF
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -332,8 +332,9 @@ class KrbInstance(service.Service):
         )
         try:
             ipautil.run(args, nolog=(self.master_password,), stdin=''.join(dialogue))
-        except ipautil.CalledProcessError:
-            print("Failed to initialize the realm container")
+        except ipautil.CalledProcessError as error:
+            logger.debug("kdb5_util failed with %s", error)
+            raise RuntimeError("Failed to initialize kerberos container")
 
     def __configure_instance(self):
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @frozencemetery 
Does it mean ACK? FreeIPA [process](https://www.freeipa.org/page/Contribute/Code#PR_Reviewer_responsibilities) requires the reviewer to add the ACK label to the PR (sorry, it may be a different use of pagure but we don't rely currently on the "Comment / Approve / Request changes"...)

frozencemetery:
Hi Flo, I'm not a developer so I can't actually ACK things (GitHub doesn't let me set labels).  If I understand the process correctly, this corresponds to a `Reviewed-by` (if you want it) without ACK.

flo-renaud:
Hi @frozencemetery 
thanks for the review and the confirmation!

flo-renaud:
master:

* 54ea4aade63e9f69de3efcefe9c8efec93372ac3 ipa-server-install: handle error when calling kdb5_util create

PR #1667, opened 2018-03-12
Error message while adding idrange with untrusted domain

amitkumar50:
While trying to add idrange with untrusted domain name error
message is misleading.

Changing the error message to:
invalid 'ID Range setup':Specified trusted domain
name could not be found.

Resolves: https://pagure.io/freeipa/issue/5078

### DIFF
--- a/ipaserver/plugins/idrange.py
+++ b/ipaserver/plugins/idrange.py
@@ -424,10 +424,10 @@ class idrange_add(LDAPCreate):
             if sid is not None:
                 entry_attrs['ipanttrusteddomainsid'] = sid
             else:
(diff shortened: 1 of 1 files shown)


pvoborni:
LGTM, but I'd encourage to remove "WebUI" from the commit title. The change is in API and applies both for Web UI and CLI. Having Web UI in a title will scare some reviewers off. :)

amitkumar50:
@pvoborni Done.Thanks.

tiran:
master:

* a3060b5238cd011b972909322491a3b784e66d0d Error message while adding idrange with untrusted domain

PR #1668, opened 2018-03-12
Backup HTTPD's mod_ssl config and cert-key pair

stlaz:
https://pagure.io/freeipa/issue/3757


pvoborni:
I was wondering whether you could create either a temp PR or temporary modify this PR with backup and restore tests to see if the change fixes some of the backup and restore tests which now currently fails in nightly testing on a check which doesn `ipa user-show admin` - returning an error about certificate verification.

rcritten:
The change looks good to me.

stlaz:
@pvoborni not a fan of temporary commits. If you want additional checks in PR CI on request, either implement a way to do that easily and without additional overhead (would require a UI) or I'm going to leave that up to the reviewer. If they want, they can add any commits in my PRs.

flo-renaud:
master:

* 7960352f1c3a0982ac87d061b9bef8418c2dae48 Backup HTTPD's mod_ssl config and cert-key pair

PR #1669, opened 2018-03-12
test_caless: adjust try/except to capture IOError

Rezney:
While testing on RHEL we are getting IOError instead of OSError.
Add also IOError.

This is mostly for compatibility reasons however should not cause
any issue as IOError is alias for OSError on Python3.


tiran:
@Rezney Where are you checking for ENOENT? I don't see it.

stlaz:
I forgot to check the commit message. Please, @Rezney, update accordingly.

Rezney:
@tiran @stlaz Sure, sorry for this. I fixed commit body but not header. Should be ok now...

tiran:
master:

* 5a04936f47de0c60867804391393d18f6e955169 test_caless: adjust try/except to capture also IOError

PR #1671, opened 2018-03-12
Log errors from NSS during FIPS OTP key import

frozencemetery:
This is the requested logging from #1621

### DIFF
--- a/daemons/ipa-slapi-plugins/libotp/hotp.c
+++ b/daemons/ipa-slapi-plugins/libotp/hotp.c
@@ -49,7 +49,9 @@
 #include <blapit.h>
 #include <pk11pub.h>
 #include <hasht.h>
+#include <prerror.h>
 #include <prnetdb.h>
+#include <syslog.h>
 
 struct digest_buffer {
     uint8_t buf[SHA512_LENGTH];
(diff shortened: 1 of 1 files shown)


frozencemetery:
(I think this also needs manual backport to ipa-4.5 for the same reasons.)

rcritten:
Correct, someone will need to manually backport this and submit a new PR.

tiran:
Writing to syslog may need a SELinux policy change.

rcritten:
Near as I can tell as long as syslog can write to the files and listen to the port there is no limitation on clients sending data to it.

As a test I augmented hotp.c to do a syslog in the hotp() function and built and ran the test program t_hotp.c in the source tree and saw all the logging in the syslog in enforcing. No AVCs were recorded. This is on F-27.

daemons/ipa-slapi-plugins/libotp/hotp.c
```diff
@@ -93,17 +95,27 @@ import_key(PK11SlotInfo *slot, CK_MECHANISM_TYPE mech, SECItem *key)
                                      PK11_ATTR_SESSION |
                                      PK11_ATTR_PRIVATE |
                                      PK11_ATTR_SENSITIVE, NULL);
-    if (!ekey)
+    if (!ekey) {
+        syslog(LOG_ERR, "libotp: in FIPS, PK11_TokenKeyGenWithFlags failed: %d",
```
tiran: Does 389-DS call ``openlog(3)`` to pass process and PID information to syslog?

frozencemetery: libotp can't - there's no library init call, as per [the base OTP/FIPS PR](https://github.com/freeipa/freeipa/pull/1621#issuecomment-369702955)

However, if we assume Linux, "The use of `openlog`() is optional; it will automatically be called by `syslog`() if necessary" per syslog(3).

rcritten:
master:

* bffcef6bbdc21315d79c783caf3fe3c74f2fe058 Log errors from NSS during FIPS OTP key import

PR #1679, opened 2018-03-14
WebUI tests: test_service

Rezney:
Extension of test_service WebUI tests.

### DIFF
--- a/ipatests/test_webui/test_service.py
+++ b/ipatests/test_webui/test_service.py
@@ -61,6 +61,14 @@ class sevice_tasks(UI_driver):
         pkey = '{}/{}@{}'.format(service, host, realm)
         return pkey
 
+    def add_host(self, hostname, dns_zone, force=False):
+        self.navigate_to_entity('host')
+        self.facet_button_click('add')
+        self.fill_textbox('hostname', hostname)
+        self.fill_textbox('dnszone', dns_zone)
+        if force:
+            self.check_option('force', 'checked')
+        self.dialog_button_click('add')
 
 @pytest.mark.tier1
 class test_service(sevice_tasks):
--- a/ipatests/test_webui/test_service.py
+++ b/ipatests/test_webui/test_service.py
@@ -54,10 +54,11 @@ class sevice_tasks(UI_driver):
             content = file_d.read()
         return content
 
-    def get_http_pkey(self):
-        host = self.config.get('ipa_server')
+    def get_service_pkey(self, service, host=None):
+        if not host:
+            host = self.config.get('ipa_server')
         realm = self.config.get('ipa_realm')
-        pkey = 'HTTP/%s@%s' % (host, realm)
+        pkey = '{}/{}@{}'.format(service, host, realm)
         return pkey
 
 
@@ -265,7 +266,7 @@ class test_service(sevice_tasks):
         Test Kerberos flags
         http://www.freeipa.org/page/V3/Kerberos_Flags
         """
-        pkey = self.get_http_pkey()
+        pkey = self.get_service_pkey('HTTP')
         name = 'ipakrbokasdelegate'
         mod = {'mod': [('checkbox', name, None)]}
         checked = ['checked']
--- a/ipatests/test_webui/test_service.py
+++ b/ipatests/test_webui/test_service.py
@@ -70,6 +70,31 @@ class sevice_tasks(UI_driver):
             self.check_option('force', 'checked')
         self.dialog_button_click('add')
 
+    def add_service(self, service,
+                    host=None,
+                    textbox=None,
+                    force=False,
+                    cancel=False,
+                    confirm=True):
+
+        if not host:
+            host = self.config.get('ipa_server')
+        self.navigate_to_entity(ENTITY)
+        self.facet_button_click('add')
+
+        self.select_combobox('service', service, combobox_input=textbox)
+        self.select_combobox('host', host)
+        if force:
+            self.wait(0.5)
+            self.check_option('force', 'checked')
+        if cancel:
+            self.dialog_button_click('cancel')
+            return
+        if not confirm:
+            return
+        self.dialog_button_click('add')
+        self.wait(0.3)
+        self.assert_no_error_dialog()
 @pytest.mark.tier1
 class test_service(sevice_tasks):
 
--- a/ipatests/test_webui/test_service.py
+++ b/ipatests/test_webui/test_service.py
@@ -112,8 +112,8 @@ class test_service(sevice_tasks):
         """
         Test service certificate actions
 
-        Requires to have CA installed and 'service_csr_path' configuration option
-        set.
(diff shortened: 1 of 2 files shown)


pvoborni:
Michal asked me to do a quick check. These are my initial thoughts. I did not run the tests. I also did not read the commits which are not mentioned. Will get to it maybe later today.

## ui_tests: select_combobox() fixes
LGTM

## ui_tests: test cancel and delete without button

Would it work if we replace the confirm/cancel functionality with confirm_btn='ok'.

So in usage one could use:

delete_record(..., confirm_btn='cancel')

Because if I get it right then "confirm" option is there only to skip the click on 'ok' button and is is useful only together with cancel=True, right?

##  ui_tests: make associations cancelable

Same comment. If the self.wait_for_request() is a problem then we could make this part configurable(skippable)

## ui_tests: add function to run cmd on UI host

`run_cmd_on_ui_host` we can specify what type of command to run. I.e. that it is a shell command and not e.g. a API command. I'm not sure if we should mention SSH in the comment. On one side it would be beneficial to get the meaning but it could then force some implementation and prevent future refactoring. Otherwise seems OK.

## ui_tests: add funcs to add/remove users public SSH key

Looks good. Maybe I'd consider making some parts like navigation to entity and clicking on update buttons configurable, as somebody might already be on the page or would like to add more stuff. But this can be extended later when needed.

## ui_tests: add assert_field_required()

Looks good. Maybe I'd just specify that where the error message is. E.g.

```
Assert we got 'Required field' error message in field validation
```

And maybe the command name could be something like:
`assert_field_validation_required`

As `assert_field_required` could also mean checking if field has '*' sign next to it.

## ui_tests: add assert_notification()

Would it work if the `find` doesn't find it and `assert_text` is set. Seems to me that it may fail that None doesn't have attribute text.

## ui_tests: run ipa-get/rmkeytab command on UI host

Would it work if `run_keytab_on_host` is run more than 1 time for different principals?

Also I'd rename pkey to principal and document possible actions in method comment.

ipatests/test_webui/ui_driver.py
```diff
@@ -1615,6 +1615,34 @@ def get_t_vals(t):
...
+
+        user (str): user to add the key to
+        ssh_key (str): public ssh key
+        """
+        self.navigate_to_entity('user')
+        self.navigate_to_record(user)
+
+        ssh_pub = 'div[name="ipasshpubkey"] button[name="add"]'
+        self.find(ssh_pub, By.CSS_SELECTOR).click()
+        self.wait_for_request(n=5, d=2)
```
pvoborni: Why wait for 5 requests with 2-second wait at the end? Clicking on add btw AFAIK doesn't cause any request. Simple `self.wait()` could be fine here.

Rezney:
@pvoborni 
Thanks for the first part of the review. Please see my actions upon it:

**ui_tests: test cancel and delete without button**:

Partially adjusted (confirm_btn), the other condition is needed for "key" testing.

**ui_tests: make associations cancelable**

Again adjusted only partially (confirm_btn). Later condition is needed in order to avoid assert error.

**ui_tests: add function to run cmd on UI host**

Fixed the docstring and also added a recommendation to use this sparsely.

**ui_tests: add funcs to add/remove users public SSH key**

Left this be for the moment as I will work with SSH keys in next suite.

**ui_tests: add assert_field_required()**

Fixed.

**ui_tests: add assert_notification()**

Fixed

**ui_tests: run ipa-get/rmkeytab command on UI host**

Changed to "pkey" to "principal".  
Could you please elaborate more on your worries about running "run_keytab_on_host" more than 1 time for different principals? 

Tried it with normal command and looks ok:

`[root@master ~]# ipa-getkeytab -p dogtag/master.ipa.test@IPA.TEST -k /tmp/test.keytab
Keytab successfully retrieved and stored in: /tmp/test.keytab
[root@master ~]# ipa-getkeytab -p HTTP/master.ipa.test@IPA.TEST -k /tmp/test.keytab
Keytab successfully retrieved and stored in: /tmp/test.keytab`

ipatests/test_webui/test_service.py
```diff
@@ -352,3 +361,461 @@ def test_kerberos_flags(self):
...
+        self.navigate_to_entity(ENTITY)
+        self.facet_button_click('add')
+        self.select_combobox('host', host, combobox_input='service')
+        self.dialog_button_click('add')
+        self.wait()
+        service_elem = self.find(".widget[name='service']", By.CSS_SELECTOR)
+        self.assert_field_validation_required(parent=service_elem)
+
+    @screenshot
+    def test_add_service_no_DNS_for_host(self):
```
pvoborni: This is the same tests has the same code as test_add_ldap_service - only different service name which is not important.

Rezney:
@pvoborni All your comments should be now hopefully addressed. Also many tests were squeezed into just two of them.

Also opened https://github.com/freeipa/freeipa-pr-ci/pull/173 to test some other certificate tests.

Rezney:
@pvoborni now all the new tests (test_provision_unprovision_keytab included) are passing. 

I guess I should get rid of the "enable WebUI tests" commit as we have still the one ongoing failure, right?

Please let me know if everything is ok.

pvoborni:
Yes, it looks OK. Code-wise removing the temp commit and giving ack is good to go. I did not and won't verify if the cases match test plan though.

Rezney:
Thx, should be ready for "ack" now. QE POV review will be done afterwards.

tiran:
master:

* d8cbd5d3ace7c3b032e9a1f2dc716b9feea1c554 ui_tests: change get_http_pkey() function
* 62a131aba0ec14eb5d15e92f7ed45172a2254669 ui_tests: add_host() support func in test_service
* 0f5084b9c4cfcb6ed25dd87dca7f58223994765a ui_tests: add_service() support func in test_service
* 735d48d82080e720d238519d233f6f333c2cf166 ui_tests: add more test cases to test_certification
* cd86fd21c574503eec2c20e8c9fe204b21e6a907 ui_tests: add more test cases
* 01fa54117de5e77f3ec94207e9f66bbfd177ac9e ui_tests: add assert_notification()
* 7fb4f755e9046d444bdd4e3e5362e2e15766686d ui_tests: add assert_field_required()
* 95de6f061cc1959030e25bd4d3e27bb1f962d3bb ui_tests: add funcs to add/remove users public SSH key
* 18e8c964f574ad9ad17b0dfdaba6b1cbd6def317 ui_tests: add function to run cmd on UI host
* 553183944ab808ba0b1b77dd4971c04dd3d3fc0f ui_tests: make associations cancelable
* 16083eb9b56a753fe1ba56fbbad826e7f4dc6e63 ui_tests: test cancel and delete without button
* bf1f2d1c3f16cea1800e773cbf7e8e59203bbbde ui_tests: select_combobox() fixes
* 5f87b9c3e502a33e50ad1f19bac3825729b9a7bb ui_tests: run ipa-get/rmkeytab command on UI host

PR #1689, opened 2018-03-15
ipa-replica-install: make sure that certmonger picks the right master

flo-renaud:
During ipa-replica-install, http installation first creates a service
principal for http/hostname (locally on the soon-to-be-replica), then
waits for this entry to be replicated on the master picked for the
install.
In a later step, the installer requests a certificate for HTTPd. The local
certmonger first tries the master defined in xmlrpc_uri (which is
pointing to the soon-to-be-replica), but fails because the service is not
up yet. Then certmonger tries to find a master by using the DNS and looking
for a ldap service. This step can pick a different master, where the
principal entry has not always be replicated yet.
As the certificate request adds the principal if it does not exist, we can
end by re-creating the principal and have a replication conflict.

The replication conflict later causes kerberos issues, preventing
from installing a new replica.

The proposed fix forces xmlrpc_uri to point to the same master as the one
picked for the installation, in order to make sure that the master already
contains the principal entry.

https://pagure.io/freeipa/issue/7041

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -197,7 +197,16 @@ def install_dns_records(config, options, remote_api):
                     'on master: %s', str(e))
 
 
-def create_ipa_conf(fstore, config, ca_enabled):
+def create_ipa_conf(fstore, config, ca_enabled, master=None):
+    """
+    Create /etc/ipa/default.conf master configuration
+    :param fstore: sysrestore file store used for backup and restore of
+                   the server configuration
+    :param config: replica config
+    :param ca_enabled: True if the topology includes a CA
+    :param master: if set, the xmlrpc_uri parameter will use the provided
+                   master instead of this host
+    """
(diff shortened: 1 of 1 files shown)


abbra:
Thanks for the patch, it looks good to me. Travis noted two failures. One is what I recorded as the issue 7447 yesterday:
Another one is actually a crash in libtcmalloc inside 389-ds:

abbra:
I looked a bit more into what this crash possibly could be and it looks like it is a crash caused by closing down nss database during shutdown of ns-slapd. Fedora's retrace has few similar crashes too -- see https://retrace.fedoraproject.org/faf/reports/1914474/ for example

flo-renaud:
@rcritten 
thanks for the review and the proposal. You are right, it's nicer to log the updated config. PR udpated with your suggestion.

flo-renaud:
master:

* 0f31564b35aac250456233f98730811560eda664 ipa-replica-install: make sure that certmonger picks the right master

PR #1692, opened 2018-03-15
Make fasttest pass without ~/.ipa/default.conf

tiran:
Some fast tests depend on an api.env with realm, domain, and host. On
machines without ~/.ipa/default.conf, the settings are not available.
Provide dummy values to make tests pass.

Closes: https://pagure.io/freeipa/issue/7432

### DIFF
--- a/ipatests/conftest.py
+++ b/ipatests/conftest.py
@@ -104,9 +104,14 @@ def pytest_addoption(parser):
 
 
 def pytest_cmdline_main(config):
-    api.bootstrap(
+    kwargs = dict(
         context=u'cli', in_server=False, in_tree=True, fallback=False
     )
+    if not os.path.isfile(os.path.expanduser('~/.ipa/default.conf')):
+        # dummy domain/host for machines without ~/.ipa/default.conf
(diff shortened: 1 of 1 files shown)


abbra:
LGTM.

pvoborni:
Thanks Christian for this patch, it was bugging me.

tiran:
@pvoborni I never noticed the issue because I had a super-old ``~/.ipa/default.conf`` :(

rcritten:
I'll want to double-check that lite-server.py continues to work. I always thought that the values passed in to bootstrap takes precedence. In the past I had executed some of the tests against lite-server to better see what was going on.

tiran:
Let's wait until @rcritten has confirmed that the patch doesn't break his workflow.

rcritten:
Latest patch works for me.

slaykovsky:
Hi @tiran 

Thank you for the patch, now `make fasttest` works fine without having `~/.ipa/default.conf`

tiran:
master:

* 75f7b7b5e697796fa5be3268e8332937aa15a5a2 Make fasttest pass without ~/.ipa/default.conf

PR #1695, opened 2018-03-15
Redirect CRL requests to the http port, not the https port

rcritten:
https://pagure.io/freeipa/issue/7433


tiran:
master:

* bfd11701188bb1d41bf0d15942b8bd2776cfa159 Redirect CRL requests to the http port, not the https port

PR #1696, opened 2018-03-15
Make WebUI unit tests to generate results as JUnit

slaykovsky:
Now WebUI unit tests are generating results in qunit format which
is not consumable well by Jenkins.

This patch adds NPM dependency for adding generation results in
JUnit XML format so it can be easily processed.


tiran:
master:

* c450e2dc8fac12a26f727ca4ecd0d3d7662e42bc Make WebUI unit tests to generate results as JUnit

PR #1697, opened 2018-03-16
Relax message check in test_create_host_with_ip

tiran:
On Travis CI, the DNS update in test case test_create_host_with_ip may fail
with different error messages. Relax the error message check and just
check that the test case is hitting a DNS update failure.

This fixes a flaky test case on CI.

Closes: https://pagure.io/freeipa/issue/7447


flo-renaud:
Hi @tiran 
I'm just annoyed I still don't understand what is different in Travis-CI env that would explain the different error msgs (I'm not DNS-fluent unfortunately...)
But I agree the proposed modification will make the test green, so ACK as soon as the test results are available.

abbra:
I agree. We are interested in seeing the first part of the message anyway.

flo-renaud:
Tests are green, giving ACK

tiran:
master:

* 69599560c16c8d3b26d18ccec5b54e54dfea9cab Relax message check in test_create_host_with_ip

PR #1700, opened 2018-03-16
Allow issuing certificates with IP addresses in subjectAltName

ipilcher:
Allow issuing certificates with IP addresses in the subject
alternative name (SAN), if all of the following are true.

* One of the DNS names in the SAN resolves to the IP address
  (possibly through a CNAME).
* All of the DNS entries in the resolution chain are managed by
  this IPA instance.
* The IP address has a (correct) reverse DNS entry that is managed
  by this IPA instance

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -19,14 +19,20 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+# When running make lint with Python 2, the dns submodule imports don't work
+# without this.
+from __future__ import absolute_import
+
 import base64
 import collections
 import datetime
+import itertools
 import logging
 from operator import attrgetter
 
 import cryptography.x509
 from cryptography.hazmat.primitives import hashes, serialization
+from dns import resolver, reversename
 import six
 
 from ipalib import Command, Str, Int, Flag
@@ -48,7 +54,7 @@ from .certprofile import validate_profile_id
 from ipalib.text import _
 from ipalib.request import context
 from ipalib import output
-from ipapython import kerberos
+from ipapython import dnsutil, kerberos
 from ipapython.dn import DN
 from ipaserver.plugins.service import normalize_principal, validate_realm
 
@@ -773,9 +779,12 @@ class cert_request(Create, BaseCertMethod, VirtualCommand):
                            "'userCertificate' attribute of entry '%s'.") % dn)
 
         # Validate the subject alt name, if any
-        generalnames = []
+        san_ipaddrs = set()
+        san_dnsnames = set()
         if ext_san is not None:
             generalnames = x509.process_othernames(ext_san.value)
+        else:
+            generalnames = []
         for gn in generalnames:
             if isinstance(gn, cryptography.x509.general_name.DNSName):
                 if principal.is_user:
@@ -787,6 +796,7 @@ class cert_request(Create, BaseCertMethod, VirtualCommand):
                     )
 
                 name = gn.value
+                san_dnsnames.add(name)
 
                 if _dns_name_matches_principal(name, principal, principal_obj):
                     continue  # nothing more to check for this alt name
@@ -858,10 +868,15 @@ class cert_request(Create, BaseCertMethod, VirtualCommand):
                             "subject alt name type %s is forbidden "
                             "for non-user principals") % "RFC822Name"
                     )
+            elif isinstance(gn, cryptography.x509.general_name.IPAddress):
+                san_ipaddrs.add(gn.value)
             else:
                 raise errors.ACIError(
                     info=_("Subject alt name type %s is forbidden")
                     % type(gn).__name__)
+
+        if san_ipaddrs:
+            _validate_san_ips(san_ipaddrs, san_dnsnames)
 
         # Request the certificate
         try:
@@ -1044,6 +1059,120 @@ def _principal_name_matches_principal(name, principal_obj):
         return False
 
     return principal in principal_obj.get('krbprincipalname', [])
+
+
+def _validate_san_ips(san_ipaddrs, san_dnsnames):
+    """
+    Check the IP addresses in a CSR subjectAltName.
+
+    Raise a ValidationError if the subjectAltName in a CSR includes
+    any IP addresses that do not match a DNS name in the SAN.  Matching means
+    the following:
+
+    * One of the DNS names in the SAN resolves (possibly via a single CNAME -
+      no CNAME chains allowed) to an A or AAAA record containing that
+      IP address.
+    * The IP address has a reverse DNS record pointing to that A or AAAA
+      record.
+    * All of the DNS records (A, AAAA, CNAME, and PTR) are managed by this IPA
+      instance.
+
+    :param san_ipaddrs: The IP addresses in the subjectAltName
+    :param san_dnsnames: The DNS names in the subjectAltName
+
+    :raises: errors.ValidationError if the SAN containes a non-matching IP
+        address.
+
+    """
+    san_dns_ips = set()
+    for name in san_dnsnames:
+        san_dns_ips.update(_san_dnsname_ips(name))
+    for ip in san_ipaddrs:
+        if unicode(ip) not in san_dns_ips:
+            raise errors.ValidationError(
+                name='csr',
+                error=_(
+                    "IP address in subjectAltName (%s) does not "
+                    "match any DNS name"
+                ) % name.value
+            )
+
+
+def _san_dnsname_ips(dnsname, dnsname_is_cname=False):
+    """
+    Resolve a DNS name to its IP address(es).
+
+    Returns a set of IP addresses, managed by this IPA instance,
+    that correspond to the DNS name (from the subjectAltName).
+
+    :param dnsname: The DNS name (text) for which to resolve the IP addresses
+    :param dnsname_is_cname: True when (recursively) resolving a CNAME (CNAME
+        chains are not followed)
+
+    :return: The set of IP addresses resolved from the DNS name
+
+    """
+    ips = set()
+    fqdn = dnsutil.DNSName(dnsname).make_absolute()
+    # This is a hack to avoid trying to find a DNS zone for unqualified DNS
+    # names.  Since no such zone exists, zone_for_name() will search all the
+    # way up to the root zone, which may take a while.
+    if len(fqdn) < 3:
+        logger.debug("Skipping IPs for %s: hostname too short", dnsname)
+        return ips
+    zone = dnsutil.DNSName(resolver.zone_for_name(fqdn))
+    name = fqdn.relativize(zone)
+    try:
+        result = api.Command['dnsrecord_show'](zone, name)['result']
+    except errors.NotFound as nf:
+        logger.debug("Skipping IPs for %s: %s", dnsname, nf)
+        return ips
+    for ip in itertools.chain(result.get('arecord', ()),
+                              result.get('aaaarecord', ())):
+        if _ip_rdns_ok(ip, fqdn):
+            ips.add(ip)
+    cnames = result.get('cnamerecord', ())
+    if cnames:
+        if dnsname_is_cname:
+            logger.debug("Skipping IPs for %s: chained CNAME", dnsname)
+        else:
+            for cname in cnames:
+                if not cname.endswith('.'):
+                    cname = u'%s.%s' % (cname, zone)
+                ips.update(_san_dnsname_ips(cname, True))
+    return ips
+
+
+def _ip_rdns_ok(ip, fqdn):
+    """
+    Check and IP address's reverse DNS record.
+
+    Determines whether the IP address has a reverse DNS entry (managed
+    by this IPA instance) that points to the FQDN.
+
+    :param ip: The IP address to check
+    :param fqdn: The FQDN (A/AAAA record) to which the reverse record should
+        point
+
+    :return: True if the IP address's reverse DNS record checks out, False if
+        it does not
+
+    """
+    rname = dnsutil.DNSName(reversename.from_address(ip))
+    zone = dnsutil.DNSName(resolver.zone_for_name(rname))
+    name = rname.relativize(zone)
+    try:
+        result = api.Command['dnsrecord_show'](zone, name)['result']
+    except errors.NotFound:
+        logger.debug("Skipping IP %s: reverse DNS record not found", ip)
+    # This assumes that the IP only has 1 A/AAAA record pointing to it (which
+    # really should be the case)
+    if any(ptr == fqdn.to_unicode() for ptr in result.get('ptrrecord', [])):
+        return True
+    else:
+        logger.debug("Skipping IP: %s: reverse DNS doesn't match FQDN %s",
+                     ip, fqdn)
+        return False
 
 
 @register()


frasertweedale:
A couple of preliminary comments:

It looks likes this change brings a dependency on `python2-dns` and the spec file needs updating accordingly.  See CI output for evidence (various failures to import module `dns`).

Please add the ticket URL to the commit message (`git commit --amend`).  I just filed
the upstream ticket: https://pagure.io/freeipa/issue/7451.

ipaserver/plugins/cert.py
```diff
@@ -773,9 +775,10 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
                            "'userCertificate' attribute of entry '%s'.") % dn)
 
         # Validate the subject alt name, if any
-        generalnames = []
+        san_has_ipaddrs = False
+        generalnames = ()
         if ext_san is not None:
-            generalnames = x509.process_othernames(ext_san.value)
+            generalnames = tuple(x509.process_othernames(ext_san.value))
```
tiran:
```
if ext_san is not None:
    generalnames = tuple(x509.process_othernames(ext_san.value))
else:
    generalnames = ()
```

ipaserver/plugins/cert.py
```diff
@@ -858,11 +861,16 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
                             "subject alt name type %s is forbidden "
                             "for non-user principals") % "RFC822Name"
                     )
+            elif isinstance(gn, cryptography.x509.general_name.IPAddress):
+                san_has_ipaddrs = True
```
frasertweedale: I see down below, in `_validate_san_ips`, we iterate the SAN names two more times, once to collect
dns names, and again to collect IP addresses.  I propose, then, that we build two sets during *this* iteration
of the SAN names.  So:
```python
elif isinstance(gn, cryptography.x509.general_name.DNSName):
    san_dnsnames.add(gn.value)  # collect dnsname for IPAddress validation
    ... # existing validation
...
elif isinstance(gn, cryptography.x509.general_name.IPAddress):
    san_ipaddrs.add(gn.value)  # collect ip addresses for IPAddress validation
```

Then later, in the call to `_validate_san_ips` you only need to pass in the sets of collected dnsnames and ipaddresses, so we don't need to traverse the `generalnames` again.
```python
if len(san_ipaddrs) > 0:
    _validate_san_ips(san_dnsnames, san_ipaddrs)
```

tiran: Like in ``DNSName`` case, you have prevent ``IPAddress`` SAN for user principals.

ipaserver/plugins/cert.py
```diff
@@ -1046,6 +1054,84 @@ def _principal_name_matches_principal(name, principal_obj):
...
+def _san_dnsname_ips(dnsname, dnsname_is_cname=False):
+    """
+    Returns a set of IP addresses, managed by this IPa instance,
+    that correspond to the DNS name (from the subjectAltName).
+
+    """
+    fqdn = dnsutil.DNSName(dnsname).make_absolute()
+    if fqdn.__len__() < 4:
+        logger.debug("Skipping IPs for %s: hostname too short" % dnsname)
+        return ()
```
frasertweedale: Please return empty `list` or empty `set` (probably the latter because you return `set` elsewhere in this subroutine).
`tuple` is semantically invalid here.

ipaserver/plugins/cert.py
```diff
@@ -858,11 +861,16 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
...
                     )
+            elif isinstance(gn, cryptography.x509.general_name.IPAddress):
+                san_has_ipaddrs = True
             else:
                 raise errors.ACIError(
                     info=_("Subject alt name type %s is forbidden")
                     % type(gn).__name__)
 
+        if san_has_ipaddrs:
+            _validate_san_ips(generalnames)
```
tiran: The ``validate_san_ips`` function iterates over general names again to construct a set of SAN dNSNames and IPAddresses. It would be easier to create both sets ``san_ipaddrs`` and ``san_dnsnames`` in this method and replace ``if san_has_ipaddrs`` with ``if san_ipaddrs``.

ipaserver/plugins/cert.py
```diff
@@ -1046,6 +1054,84 @@ def _principal_name_matches_principal(name, principal_obj):
     return principal in principal_obj.get('krbprincipalname', [])
 
 
+def _validate_san_ips(generalnames):
+    """
```
tiran: Please follow PEP 8 and https://www.python.org/dev/peps/pep-0257/ for doc strings.

ipaserver/plugins/cert.py
```diff
@@ -1046,6 +1054,84 @@ def _principal_name_matches_principal(name, principal_obj):
...
+
+
+def _san_dnsname_ips(dnsname, dnsname_is_cname=False):
+    """
+    Returns a set of IP addresses, managed by this IPa instance,
+    that correspond to the DNS name (from the subjectAltName).
+
+    """
+    fqdn = dnsutil.DNSName(dnsname).make_absolute()
+    if fqdn.__len__() < 4:
```
frasertweedale: this can be
```python
if len(fqdn) < 4:
```

Why is this required?  It is not clear to me why a name consisting of fewer than 4 components
cannot be resolved.

ipaserver/plugins/cert.py
```diff
@@ -1046,6 +1054,84 @@ def _principal_name_matches_principal(name, principal_obj):
...
+                raise errors.ValidationError(
+                    name='csr',
+                    error=_(
+                        "IP address in subjectAltName (%s) does not "
+                        "match any DNS name"
+                    ) % name.value
+                )
+
+
+def _san_dnsname_ips(dnsname, dnsname_is_cname=False):
```
tiran: The check takes at least two additional API calls for each IP addresses, which is rather expensive.

ipaserver/plugins/cert.py
```diff
@@ -1046,6 +1054,84 @@ def _principal_name_matches_principal(name, principal_obj):
...
+            ips.add(ip)
+    cnames = result.get('cnamerecord', ())
+    if cnames:
+        if dnsname_is_cname:
+            logger.debug("Skipping IPs for %s: chained CNAME" % dnsname)
+        else:
+            for cname in cnames:
+                if not cname.endswith('.'):
+                    cname = u'%s.%s' % (cname, zone)
+                ips.update(_san_dnsname_ips(cname, True))
```
frasertweedale: Do we need some protective measure here to prevent loops or deep CNAME paths?
(I think we do).  Maybe we can just punt on CNAME though, for now?

ipaserver/plugins/cert.py
```diff
@@ -1046,6 +1054,84 @@ def _principal_name_matches_principal(name, principal_obj):
...
+    """
+    rname = dnsutil.DNSName(reversename.from_address(ip))
+    zone = dnsutil.DNSName(resolver.zone_for_name(rname))
+    name = rname.relativize(zone)
+    try:
+        result = api.Command['dnsrecord_show'](zone, name)['result']
+    except errors.NotFound as nf:
+        logger.debug("Skipping IP %s: reverse DNS record not found" % ip)
+    # This assumes that the IP only has 1 A/AAAA record pointing to it.
+    if result['ptrrecord'][0] != fqdn.to_unicode():
```
frasertweedale: The `[0]` index operation seems unsafe here.  We can probably remove the restriction and make it safe
in one go anyhow.  Does the following, to replace lines 1127..1132, have the desired behaviour?
```python
return any(ptr == fqdn.to_unicode() for ptr in result.get('ptrrecord', []))
```

ipaserver/plugins/cert.py
frasertweedale: Yes, good pickup @tiran.

ipaserver/plugins/cert.py
frasertweedale: Haha, @tiran and I were both reviewing the PR at the same time and came up with mostly the same comments ^_^

ipaserver/plugins/cert.py
ipilcher: Yes, that was the intent.  I'm not familiar with this area at all, but wouldn't that be the right thing to do?  I.e., if a user can't "own" a DNS name, they can't own an IP address either.

If that's wrong, I'm going to have to rethink the logic for which IP addresses are considered acceptable for a particular certificate.

ipaserver/plugins/cert.py
ipilcher: Will do.  The doc strings in the current PR are just placeholders.

ipaserver/plugins/cert.py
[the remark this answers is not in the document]
ipilcher: Since I'm iterating over it multiple times, I wanted an immutable data structure.  (Java background showing.)

ipaserver/plugins/cert.py
ipilcher: Sounds reasonable.

My goal with the initial patch was to be as non-intrusive as possible to the existing flow when no IP addresses are present in the CSR.

ipaserver/plugins/cert.py
ipilcher: Will do.

ipaserver/plugins/cert.py
rcritten: I think that's right. I'd contain IPAddr entries to just hosts and services and if possible ensure that the IP address resolves to the hostname in question. This could be problematic for multi-homed hosts though. The managed_by attribute might be of use here assuming that all DNS hosts resolve into IPA hosts, in which case as long as a host/service requesting an IPAddr SAN is managed by the host that has the same name of the A/AAAA record then it is allowed.

ipaserver/plugins/cert.py
ipilcher: The goal is to avoid attempting to "resolve" relative DNS names and DNS names for top-level domains.  Keep in mind that the "root" element of a DNS name counts, so a name like www.foo.com has a length of 4.

ipaserver/plugins/cert.py
ipilcher: That's what setting dnsname_is_cname to True does.  We accept a single CNAME, but chains aren't allowed.

ipaserver/plugins/cert.py
ipilcher: Will do.

ipaserver/plugins/cert.py
frasertweedale: I don't think we can make this assumption.  With gTLDs a customer really could have a domain name of "nike." or "coke." or whatever.  And maybe they want an IPAddress certificate for it... seems unlikely but I just don't see any reason to impose this restriction, given that we're only use A/AAAA records managed by this IPA deployment for verification.

ipaserver/plugins/cert.py
frasertweedale: Thanks, I missed that.  Please add some commentary about that in the docstring.

ipaserver/plugins/cert.py
frasertweedale: Adding one statement to the DNSName case is an acceptable degree of intrusion :)

ipaserver/plugins/cert.py
frasertweedale: It's such a niche case, I wouldn't really worry about it.  If there's an obvious way to cut down the number of requests, for sure let's do it, but IMO this is not a blocker.

ipaserver/plugins/cert.py
ipilcher: OK.  So I can reduce the threshold to 3, which will allow names like "shoes.nike." (len = 3), while still ruling out completely unqualified names/TLDs.  Or I can completely remove the check.  IIRC, this will mean waiting for DNS timeouts when attempting to resolve the zone for unqualified names.

Any preference?

ipaserver/plugins/cert.py
ipilcher: If anything, I believe it's the ['ptrrecord'] that may fail.  It looks like the 1-line version will work, so I will go with that.  (Took me a bit to understand the syntax; C/Java background again.)

ipaserver/plugins/cert.py
```diff
@@ -1046,6 +1054,84 @@ def _principal_name_matches_principal(name, principal_obj):
...
+    """
+    Returns a set of IP addresses, managed by this IPa instance,
+    that correspond to the DNS name (from the subjectAltName).
+
+    """
+    fqdn = dnsutil.DNSName(dnsname).make_absolute()
+    if fqdn.__len__() < 4:
+        logger.debug("Skipping IPs for %s: hostname too short" % dnsname)
+        return ()
+    zone = dnsutil.DNSName(resolver.zone_for_name(fqdn))
```
frasertweedale: Instead of approaching it this way, where we potentially forward to external DNS servers that
may not respond promptly, perhaps we can perform `ipa dnszone-find` or direct LDAP searches to see
whether the IPA DNS serves a zone that includes the given name.

I propose an ldapsearch, otherwise we may have to repeat the `dnszone-find`, removing the leftmost domain component, until we either find a match or hit the root zone.

This ldapsearch can also take the place of the `dnsrecord-show` command.

Let me play around with the IPA DNS for a bit (I am not familiar with this subsystem) and see how we can do it.

ipaserver/plugins/cert.py
abbra: We do not allow single label deployments of IPA, at least not with trust to AD. I think having at least two labels limitation here is sensible.

frasertweedale: Someone will eventually want to deploy IPA for the gTLD.  Let's avoid the restriction here, even if it is currently imposed elsewhere.  After all, we will still check that IPA DNS has authority over the name and IP address (PTR record) in question.

ipaserver/plugins/cert.py
ipilcher: We could do that, but it's more work.  :-)

Seriously, this is getting into optimizing the error path.  We'll only hit DNS timeout issues in the case where the CSR SAN includes a hostname or IP that isn't managed by this IPA instance.  Is it worth spending effort on this case?

ipaserver/plugins/cert.py
frasertweedale: Hi @ipilcher , I'm back after a week of vacation and will be turning my attention back to this PR very soon (probably tomorrow).

ipilcher:
@frasertweedale 

> It looks likes this change brings a dependency on python2-dns and the spec file needs updating accordingly. See CI output for evidence (various failures to import module dns).

I'm struggling to figure this one out.  I can see the import errors, but I also see that both python2-dns and python3-dns are already required in several places in the SPEC file.  It isn't clear to me which packages are missing the requirement, whether it's a build or runtime requirement that missing, etc.

frasertweedale:
@ipilcher it looks to be caused by the import statement inferring a relative import (in python2 only).
I can reproduce with `make lint PYTHON=$(which python2)`.  This is because there is a `ipaserver.plugins.dns` module alongside the `ipaserver.plugins.cert` module where the import is performed.

I may suffice to change the import to:
```python
import dns.resolver
import dns.reversename
```
and chase it up at the call sites.

There are also some linter warnings to clean up:
```
ipaserver/plugins/cert.py:1087: [W1201(logging-not-lazy), _san_dnsname_ips] Specify string format arguments as logging function parameters)
ipaserver/plugins/cert.py:1094: [W1201(logging-not-lazy), _san_dnsname_ips] Specify string format arguments as logging function parameters)
ipaserver/plugins/cert.py:1104: [W1201(logging-not-lazy), _san_dnsname_ips] Specify string format arguments as logging function parameters)
ipaserver/plugins/cert.py:1125: [W1201(logging-not-lazy), _ip_rdns_ok] Specify string format arguments as logging function parameters)
ipaserver/plugins/cert.py:1128: [W1201(logging-not-lazy), _ip_rdns_ok] Specify string format arguments as logging function parameters)
ipaserver/plugins/cert.py:1124: [W0612(unused-variable), _ip_rdns_ok] Unused variable 'nf') 
```

frasertweedale:
@ipilcher actually my suggested solution also fails in the same way.  But this should do the trick:
```python
from __future__ import absolute_import
```

ipaserver/plugins/cert.py
frasertweedale: @ipilcher how does IPA currently issue a certificate with an unqualified name?  Does it not require creating a host or service principal with that exact (unqualified) name?  I cannot see any other way.

Is it a common practice to have hosts/services with unqualified names in the IPA deployment?  I suspect not.

ipilcher: As far as I can tell, it "just works."  For example, I've created a host for one of my switches - ``switch4.penurio.us``, and I'm able to issue a certificate with both ``switch4.penurio.us`` and ``switch4`` as DNS names in the subjectAltName.

I used this CSR:

```
-----BEGIN CERTIFICATE REQUEST-----
MIIBqzCCARQCAQAwMjETMBEGA1UECgwKUEVOVVJJTy5VUzEbMBkGA1UEAwwSc3dp
dGNoNC5wZW51cmlvLnVzMIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDE60MB
/fkVDwvAF3aSw5EkfucNloipohXBVD9T6dY01yEIVvUvUX1eraloGUPyZa+SUMyH
/qT6fOwzUen00hHl3fUHcWr68fCQHvzPXoYPSfMm2qHNRsAK0jRvyL75xhiBLTON
GtHvZ87xOoWKp3R/o0lsXyUU7uIubbJgzBeLoQIDAQABoDkwNwYJKoZIhvcNAQkO
MSowKDAmBgNVHREEHzAdghJzd2l0Y2g0LnBlbnVyaW8udXOCB3N3aXRjaDQwDQYJ
KoZIhvcNAQELBQADgYEAfyk6s0SZJ7vZsDnEf1wJG001FQnWnEvQz3KbLCaOaLbQ
RybUhz4EVFut+CgwgpKpZ+R8MST+ZLDBuToOkhdNEXWWktiyYriXRF8VXIVpRxKt
K5vHxMbPAcQwL9PK9EJSKFYKJUnAVSTpBCdOAu82WQB1SSi0O4sUNXs/GBn+OrA=
-----END CERTIFICATE REQUEST-----
```

And IPA (``python2-ipaserver-4.5.0-22.el7.centos.noarch``) issued this certificate:

```
-----BEGIN CERTIFICATE-----
MIIDszCCApugAwIBAgIEH/8AETANBgkqhkiG9w0BAQsFADA1MRMwEQYDVQQKEwpQ
RU5VUklPLlVTMR4wHAYDVQQDExVDZXJ0aWZpY2F0ZSBBdXRob3JpdHkwHhcNMTgw
NDExMTQwMTI5WhcNMjMwNDEyMTQwMTI5WjAyMRMwEQYDVQQKDApQRU5VUklPLlVT
MRswGQYDVQQDDBJzd2l0Y2g0LnBlbnVyaW8udXMwgZ8wDQYJKoZIhvcNAQEBBQAD
gY0AMIGJAoGBAMTrQwH9+RUPC8AXdpLDkSR+5w2WiKmiFcFUP1Pp1jTXIQhW9S9R
fV6tqWgZQ/Jlr5JQzIf+pPp87DNR6fTSEeXd9Qdxavrx8JAe/M9ehg9J8ybaoc1G
wArSNG/IvvnGGIEtM40a0e9nzvE6hYqndH+jSWxfJRTu4i5tsmDMF4uhAgMBAAGj
ggFQMIIBTDAfBgNVHSMEGDAWgBTeReJBnbTv7K95sQAnIHhqo24IXDA8BggrBgEF
BQcBAQQwMC4wLAYIKwYBBQUHMAGGIGh0dHA6Ly9pcGEtY2EucGVudXJpby51cy9j
YS9vY3NwMA4GA1UdDwEB/wQEAwIE8DAdBgNVHSUEFjAUBggrBgEFBQcDAQYIKwYB
BQUHAwIwdQYDVR0fBG4wbDBqoDKgMIYuaHR0cDovL2lwYS1jYS5wZW51cmlvLnVz
L2lwYS9jcmwvTWFzdGVyQ1JMLmJpbqI0pDIwMDEOMAwGA1UECgwFaXBhY2ExHjAc
BgNVBAMMFUNlcnRpZmljYXRlIEF1dGhvcml0eTAdBgNVHQ4EFgQUikM/FyUSAV90
BPMjevjTu7Vt8X8wJgYDVR0RBB8wHYISc3dpdGNoNC5wZW51cmlvLnVzggdzd2l0
Y2g0MA0GCSqGSIb3DQEBCwUAA4IBAQAZjAREXMp9rTxZYYSNvI8m05M+2Qq5tiSp
CKUOaoDZAe/ivQOJLBucCwipGVT4TlnPsV3jvbcHbENw/U+BQEK21OJvIkOPXKjq
p3PpDVHw3fY/f8VP2uZ5OBueYbZNQYFFN1Ad7G17CcvoWTwLbtxwqQ6O8h0G4bKf
M8+bbcMxpQQh6UwpY2LG+07fBjC/zOX0BmHsK8fgF7ggC44TvZnG7x93Ie8MDBPy
LUQOcPqhkL5dnYkLk8TTm1Lkx8O13zlPw2iqYj3IRWjueMFWi1a68rHtl3os4ah3
EdtwHvOsWxVQNXeCwTxbC+wyoMilMg4cdNI2j+jtfIjLECyzFWVs
-----END CERTIFICATE-----
```

frasertweedale: @ipilcher interesting, thanks.  I did not realise this - it is not an intentional part of `cert-request` behaviour, but a side-effect of `host-find`, which appends the domain to unqualified names.

Thinking about a principled design that does not impose the name "depth limit" (even if other parts of IPA do)... maybe we can do it this way:

1. (fact of current behaviour) for each SAN DNSName we locate a matching host or service principal.  An unqualified DNSName will either resolve to a principal entry whose qualified name we then know (because it appears in the entry), or the request will be rejected (because no matching host/service entry was found).

2. Therefore, when producing the set of IP addresses corresponding to the DNSNames (`_san_dnsname_ips`), instead of iterating the SAN DNSNames, iterate the qualified hostnames of the principal entries produced at (1).

This will ensure we only try to resolve IP addresses of qualified names, while allowing the possibility of a single-element qualified name (e.g. `nike.`).  (Again, never mind that the rest of IPA still has  to catch up with this reality.  But I do not want to perpetuate this bad assumption in new code).

ipilcher: That certainly makes sense (to the extent that I understand it).  I must admit that I have no clue how to actually implement it, though.

frasertweedale: @ipilcher I will be happy to implement this behaviour.  If you would push a new patch that fixes the tests and implements the other nits/suggestions already discussed, I will then implement and commentate the suggested behaviour and we can discuss and evaluate that implementation.

I do not mean to condescend or suggest that the approach cannot be explained so that you can implement it; rather let's collaborate on the code to un-stall your PR.

ipilcher: First, I want to apologize for being so unresponsive on this.  Conference prep, day job, kids ...  :-(
> I do not mean to condescend or suggest that the approach cannot be explained so that you can implement it; rather let's collaborate on the code to un-stall your PR.
Absolutely no offense taken.  I'd just like to get this feature working.
I believe that I've just pushed an updated PR that addresses all of the points that have been raised.
Thanks for your patience!

frasertweedale:
@ipilcher 

> First, I want to apologize for being so unresponsive on this. Conference prep, day job, kids ... :-(
> 
I know the feeling mate.  No apology needed :)

> I believe that I've just pushed an updated PR that addresses all of the points that have been raised.
> 
Thanks, I'll push an update in the next day or so.

frasertweedale:
@ipilcher some updates added in https://github.com/freeipa/freeipa/pull/1843.  Please review
the changes (which occur across a few commits).

One change not previously discussed is a generalisation of `_san_dnsname_ips` to allow
arbitrary cname depth.  IMO, the code is a bit clearer now, but I'm interested in your feedback on that.

I haven't tested my changes yet.  We'll need to add a bunch of tests for the new behaviour to the test suite.

tiran:
I'm closing this PR because @frasertweedale has cloned your PR as PR #1843. There is no point to keep both PRs open. Please follow up on the other PR.

PR #1703, opened 2018-03-19
Pylint 1.8.3 fixes

tiran:
Teach pylint more about the internals of API to fix various issues with
pylint 1.8.3.

### DIFF
--- a/pylint_plugins.py
+++ b/pylint_plugins.py
@@ -62,13 +62,22 @@ fake_backend = {'Backend': [
     {'wsgi_dispatch': ['mount']},
 ]}
 
-NAMESPACE_ATTRS = ['Command', 'Object', 'Method', fake_backend, 'Updater',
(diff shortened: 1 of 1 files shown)


pylint_plugins.py
```diff
@@ -90,13 +99,17 @@ def fake_class(name_or_class_obj, members=()):
     'ipalib.config.Env': [
         {'__d': ['get']},
         {'__done': ['add']},
-        'xmlrpc_uri',
```
stlaz: Is there a reason these members were here before? Why are they being removed? The commit message does not explain that.

stlaz:
LGMT and the tests pass -> ACK.

tiran:
master:

* ce8ec5028af0508fe2230a0834b775a0bc9d2cd9 Pylint 1.8.3 fixes

PR #1707, opened 2018-03-19
Make tox tests to generate results in JUnit XML

slaykovsky:
As our tox runs pytest it's great to have their results in JUnit
format for later processing.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,7 +17,7 @@ deps=
     ipatests
 commands=
     {envbindir}/ipa --help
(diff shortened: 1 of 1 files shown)


slaykovsky:
Hi @tiran

Thanks,  I've changed the file.

tiran:
thanks!

tiran:
master:

* b947296fe9bada50b17a42ad42794a35286ba3d4 Make tox tests to generate results in JUnit XML

PR #1708, opened 2018-03-19
Processing of server roles should ignore errors.EmptyResult

abbra:
When non-admin user issues a command that utilizes
api.Object.config.show_servroles_attributes(), some server roles might
return errors.EmptyResult, indicating that a role is not visible to this
identity.

Most of the callers to api.Object.config.show_servroles_attributes() do
not process errors.EmptyResult so it goes up to an API caller. In case
of Web UI it breaks retrieval of the initial configuration due to ipa
config-show failing completely rather than avoiding to show available
server roles.

Fixes: https://pagure.io/freeipa/issue/7452

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -1617,7 +1617,12 @@ def upgrade_configuration():
         AUTOREDIR='' if auto_redirect else '#',
         CRL_PUBLISH_PATH=paths.PKI_CA_PUBLISH_DIR,
         DOGTAG_PORT=8009,
-        CLONE='#'
+        CLONE='#',
+        WSGI_PREFIX_DIR=paths.WSGI_PREFIX_DIR,
+        GSSAPI_SESSION_KEY=paths.GSSAPI_SESSION_KEY,
+        FONTS_DIR=paths.FONTS_DIR,
+        IPA_CCACHES=paths.IPA_CCACHES,
+        IPA_CUSTODIA_SOCKET=paths.IPA_CUSTODIA_SOCKET
     )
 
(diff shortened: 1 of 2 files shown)


abbra:
I added one more patch related to F28 to fix F27->F28 upgrade failures reported by OpenQA.

ipaserver/install/server/upgrade.py
```diff
@@ -1651,6 +1651,7 @@ def upgrade_configuration():
 
         ds_dirname = dsinstance.config_dirname(ds_serverid)
 
+        sub_dict['WSGI_PREFIX_DIR'] = paths.WSGI_PREFIX_DIR
```
tiran: This belongs on line 1620 in the definition of ``sub_dict``.

abbra:
Moved `WSGI_PREFIX_DIR` definition to the `sub_dict` definition and rebased.

abbra:
During OpenQA tests a new set of issues were found -- basically, we need to add all variables from e6c707b168067ebb3705c21efc377acd29b23fff to the upgrade code. Adam Williamson did a fix downstream in Fedora and tested it in OpenQA. I updated upstream patch according to his findings.

stlaz:
Removing ACK due to code additions.

tiran:
ACK ACK

stlaz:
Tested and works.

Tiboris:
master:

* 48acb7d86583c54296bc8334517989ce5a2199ea Processing of server roles should ignore errors.EmptyResult
* 2da9a4ca6a060dea106f02269a961d9e6b8bebcf Update template directory with new variables when upgrading ipa.conf.template

PR #1705, opened 2018-03-19
More cleanup after uninstall

tiran:
Remove more files during ipaserver uninstallation:

* /etc/gssproxy/10-ipa.conf
* /etc/httpd/alias/*.ipasave
* /etc/httpd/conf/password.conf
* /etc/ipa/dnssec/softhsm2.conf
* /etc/systemd/system/httpd.service.d/
* /var/lib/ipa/dnssec/tokens

Fixes: https://pagure.io/freeipa/issue/7183
See: https://pagure.io/freeipa/issue/2694
Signed-off-by: Christian Heimes <cheimes@redhat.com>

PR #1712, opened 2018-03-20
https://pagure.io/freeipa/issue/7183

m3gat0nn4ge:
component: Uninstall cleanup
   
   Remove /etc/gssproxy/10-ipa.conf during uninstall
   
   https://pagure.io/freeipa/issue/7183

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,7 @@ freeipa2-dev-doc
 
 # Root directory
 /freeipa.spec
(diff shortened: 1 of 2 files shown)


stlaz:
Hello, thanks for your contribution. I believe we already have a PR which fixes this problem and adds even some more files to be cleaned up: https://github.com/freeipa/freeipa/pull/1705

PR #1714, opened 2018-03-20
use LDAP Whoami command when creating an OTP token

abbra:
ipa user-find --whoami is used by ipa otptoken-add to populate
ipaTokenOwner and managedBy attributes. These attributes, in turn are
checked by the self-service ACI which allows to create OTP tokens
assigned to the creator.

With 389-ds-base 1.4.0.6-2.fc28 in Fedora 28 beta there is a bug in
searches with scope 'one' that result in ipa user-find --whoami
returning 0 results.

Because ipa user-find --whoami does not work, non-admin user cannot
create a token. This is a regression that can be fixed by using LDAP
Whoami command.

Fixes: https://pagure.io/freeipa/issue/7456

### DIFF
--- a/ipaserver/plugins/otptoken.py
+++ b/ipaserver/plugins/otptoken.py
@@ -311,13 +311,12 @@ class otptoken_add(LDAPCreate):
         # If owner was not specified, default to the person adding this token.
         # If managedby was not specified, attempt a sensible default.
         if 'ipatokenowner' not in entry_attrs or 'managedby' not in entry_attrs:
-            result = self.api.Command.user_find(
-                whoami=True, no_members=False)['result']
-            if result:
-                cur_uid = result[0]['uid'][0]
+            cur_dn = DN(self.api.Backend.ldap2.conn.whoami_s()[4:])
+            if cur_dn:
+                cur_uid = cur_dn[0].value
                 prev_uid = entry_attrs.setdefault('ipatokenowner', cur_uid)
                 if cur_uid == prev_uid:
-                    entry_attrs.setdefault('managedby', result[0]['dn'])
+                    entry_attrs.setdefault('managedby', cur_dn.ldap_text())
 
         # Resolve the owner's dn
         _normalize_owner(self.api.Object.user, entry_attrs)


ipaserver/plugins/otptoken.py
```diff
@@ -311,13 +311,12 @@ def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys, **options):
         # If owner was not specified, default to the person adding this token.
         # If managedby was not specified, attempt a sensible default.
         if 'ipatokenowner' not in entry_attrs or 'managedby' not in entry_attrs:
-            result = self.api.Command.user_find(
-                whoami=True, no_members=False)['result']
-            if result:
-                cur_uid = result[0]['uid'][0]
+            cur_dn = DN(self.api.Backend.ldap2.conn.whoami_s()[4:])
```
flo-renaud: I would just add a comment explaining the [4:] (the result is a string starting with dn: uid=... and as we want to remove the "dn: " at the beginning, we need to remove the 4 first chars).

abbra: @flo-renaud we have this pattern in other places:
```
$ git grep conn.whoami_s
ipapython/ipaldap.py:        bind_dn = self.conn.whoami_s()[4:]
ipaserver/plugins/otptoken.py:            cur_dn = DN(self.api.Backend.ldap2.conn.whoami_s()[4:])
ipaserver/plugins/whoami.py:        dn = DN(ldap.conn.whoami_s()[4:])
```
I'm not sure it makes sense to add this info everywhere. It is documented in LDAP RFCs: https://tools.ietf.org/html/rfc4513 section 5.2.1.8. tells that authzid is
```
      authzId = dnAuthzId / uAuthzId

      ; distinguished-name-based authz id
      dnAuthzId =  "dn:" distinguishedName

      ; unspecified authorization id, UTF-8 encoded
      uAuthzId = "u:" userid
      userid = *UTF8 ; syntax unspecified
```

and "WHO AM I?" LDAP RFC https://tools.ietf.org/html/rfc4532 explicitly refers to RFC4513 for the format of the returned identity.

flo-renaud: Hi @abbra 
my point is that a short comment can sometimes save others time. We don't all have the same LDAP knowledge and experience with LDAP RFCs.
I don't ask to change the code everywhere but when touching a portion of code, it seems good practice to me to try to enhance the clarity. My 2 cents...

flo-renaud:
Tests are green, giving the ACK label.

tiran:
master:

* b47d6a3654ee05e5751bde8ab32934bd809a7347 use LDAP Whoami command when creating an OTP token

PR #1722, opened 2018-03-22
REDO: Fix installer CA port check to include 8080

m3gat0nn4ge:
component: installer - CA

The installer was not checking the availability of port 8080, cainstance.check_port() function has been renamed to cainstance.check_ports() and updated to check for port 8080 also.
The ipautil.host_port_open() function was being used incorrectly to test for a free port, I have added a new ipautil.host_port_free() function based on host_port_open() with different logic.
cainstance.check_port() now calls ipautil.host_port_free() instead of negating the result from ipautil.host_port_open().

https://pagure.io/freeipa/issue/7415


ipapython/ipautil.py
```diff
@@ -1037,6 +1037,53 @@ def host_port_open(host, port, socket_type=socket.SOCK_STREAM,
     return port_open
 
 
+def host_port_free(host, port, socket_type=socket.SOCK_STREAM,
```
tiran: Why are you introducing a new function. We already have a ``host_port_open`` helper That looks very similar to your code.

m3gat0nn4ge: Please see my comments here: https://pagure.io/freeipa/issue/7415

You cannot negate the result of host_port_open() to determine if a port is free

tiran: I left a comment on the ticket. We need a different approach to check if a port is bindable. Even the new check doesn't work properly.

m3gat0nn4ge:
host_port_free() has been rewritten... pls review

tiran:
Thanks for the updated patch. I have a couple of ideas how to simplify the checker and make it more robust.

* You don't have to overcomplicate logging. Just log with ``logger.debug()``. The logging framework will do the filtering for you.
* You have to wrap the socket code into a try/finally block. Otherwise you may end up with a dangling socket. It's really too bad that Python 2's socket class doesn't support with statement.
* You can share most of the code if you separate socket creation from binding. ``bind(('', port))`` will bind to any address even for IPv6.

This should do the trick. Feel free to adjust the code and improve the logging messages as you see fit.

```
try:
    s = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
except socket.error:
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    logger.debug("Checking for free port %i on IPv4 only", port)
else:
    logger.debug(...)

try:
    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, True)
    s.bind(('', port))  # '' means to any address IPv4 / IPv6
except socket.error as se:
    # failed, port is in use
    logger.debug(...)
    return False
else:
    # ok, port is not in use
    logger.debug(...)
    return True
finally:
   s.close()
```

m3gat0nn4ge:
Thanks @tiran ,

Regarding "**s.bind(('', port))  # '' means to any address IPv4 / IPv6"**
I read this yesterday from https://docs.python.org/3/library/socket.html:

"For IPv4 addresses, two special forms are accepted instead of a host address: the empty string represents INADDR_ANY, and the string '<broadcast>' represents INADDR_BROADCAST. This behavior is not compatible with IPv6, therefore, you may want to avoid these if you intend to support IPv6 with your Python programs."

That's why I went with the 2 separate host versions - '' for IPv4 and '::' for IPv6

Am I misinterpreting the docs?

m3gat0nn4ge:
Could I please get a little help interpreting the CI failure?.. thanks guys

tiran:
Please rebase, I just pushed #1729

tiran:
I think the warning in the docs is just about ``<broadcast>``, but I'm not sure. How about:

```
try:
    s = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
    anyaddr = '::'  # in6addr_any
except socket.error:
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    anyaddr = ''  # INADDR_ANY
...
    s.bind((anyaddr, port))
```

m3gat0nn4ge:
@tiran  .. ready for review, thnx

tiran:
Something went wrong with your rebase. I see several unrelated commits in your PR. Please squash your commits into one commit and force push the remaining commit. Hopefully that should get rid of the unrelated commits.

m3gat0nn4ge:
Squashed and pushed again... any idea what went wrong?

tiran:
No clue :(

Something went horribly wrong with your rebase. I don't understand what went wrong exactly.

felipevolpone:
@m3gat0nn4ge Hi, the workflow to rebase could be something like this (not sure what went wrong):
```
git clone https://github.com/m3gat0nn4ge/freeipa;
cd freeipa;
git checkout -b my_branch
<do some changes>
git push -u myorigin my_branch
git remote add upstream_code https://github.com/freeipa/freeipa;
git rebase upstream/master
git push -f myorigin my_branch;
```

m3gat0nn4ge:
Thanks Felipe,

I used 'git pull upstream master' assuming that 'pull.rebase=true' would do
the right thing... maybe it doesn't?

cheers

m3gat0nn4ge:
@tiran 

Just wondering what needs to be done to resolve this PR as it now has conflicting files... should I trash my repo, clone anew, then reapply my changes? .. or can this be fixed?

thanks

tiran:
I see, that you have committed your patches to your own master. That is not a good idea. In the future, create a branch, commit your changes to the branch and then push that branch as a pull request.

Since I don't have a simple solution to fix your branch, I have taken your patch and created a new PR with some slight modifications. You are listed as co author to give you appropriate credits.

Thanks for your hard work!

tiran:
The PR has been replaced by PR #1753. I'm closing this PR as reject and follow up on the other PR.

(``rejected`` is our internal label to track PRs that were not merged.)

PR #1723, opened 2018-03-22
replica-install: warn when there is only one CA in topology

frasertweedale:
For redundancy and security against catastrophic failure of a CA
master, there must be more than one CA master in a topology.
Replica installation is a good time to warn about this situation.
Print a warning at the end of ipa-replica-install, if there is only
one CA replica in the topology.

Fixes: https://pagure.io/freeipa/issue/7459

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -14,7 +14,9 @@ import dns.reversename as dnsreversename
 import os
 import shutil
 import socket
+import sys
 import tempfile
+import textwrap
 import traceback
 
 from pkg_resources import parse_version
@@ -1517,6 +1519,7 @@ def install(installer):
     if options.setup_adtrust:
         adtrust.install(False, options, fstore, api)
 
+    ca_servers = service.find_providing_servers('CA', api.Backend.ldap2, api)
     api.Backend.ldap2.disconnect()
 
     if not promote:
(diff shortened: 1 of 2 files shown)


ipaserver/install/server/replicainstall.py
```diff
@@ -1549,6 +1550,14 @@ def install(installer):
     # Everything installed properly, activate ipa service.
     services.knownservices.ipa.enable()
 
+    # Print a warning if CA role is only installed on one server
+    if len(ca_servers) == 1:
+        print()
+        print("WARNING: The CA service is only installed on one server ({})."
+                .format(ca_servers[0]))
```
tiran: ``E127 continuation line over-indented for visual indent``

ipaserver/install/server/replicainstall.py
```diff
@@ -1549,6 +1550,14 @@ def install(installer):
     # Everything installed properly, activate ipa service.
     services.knownservices.ipa.enable()
 
+    # Print a warning if CA role is only installed on one server
+    if len(ca_servers) == 1:
+        print()
```
tiran: The warning should be printed to stderr. You can use https://docs.python.org/3/library/textwrap.html#textwrap.dedent with triple quotes to create one string and then print it to ``file=sys.stderr``.

ipaserver/install/service.py
```diff
@@ -112,31 +112,42 @@ def add_principals_to_group(admin_conn, group, member_attr, principals):
...
 
-    Find a server that is a CA.
     """
     dn = DN(('cn', 'masters'), ('cn', 'ipa'), ('cn', 'etc'), api.env.basedn)
     query_filter = conn.make_filter({'objectClass': 'ipaConfigObject',
                                      'ipaConfigString': 'enabledService',
                                      'cn': svcname}, rules='&')
     try:
         entries, _trunc = conn.find_entries(filter=query_filter, base_dn=dn)
+        return [entry.dn[1].value for entry in entries]
```
tiran: Since you have to touch the patch for PEP8 violations any way, please move the return clause into a new ``else`` block (try/except/else).

frasertweedale:
@tiran thanks for review.  PR updated.

tiran:
Code looks good, but PR needs a rebase for fix #1729

tiran:
master:

* 5afbe1d26103909760239e62e252b09e5fb3b155 replica-install: warn when there is only one CA in topology

PR #1724, opened 2018-03-22
Test webui realm domains

celestian:
I prepare automation of some WebUI Realm Domains test.

This PR is continue of https://github.com/freeipa/freeipa/pull/1644

### DIFF
--- a/ipatests/test_webui/test_realmdomains.py
+++ b/ipatests/test_webui/test_realmdomains.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Authors:
 #   Petr Vobornik <pvoborni@redhat.com>
 #
@@ -19,12 +20,16 @@
 
 """
 Realm domains tests
+
+Update means Check DNS in WebUI.
+Force udpate means Force Update in WebUI.
 """
 
 from ipatests.test_webui.ui_driver import UI_driver
 from ipatests.test_webui.ui_driver import screenshot
 from ipatests.test_webui.data_dns import (
-    ZONE_ENTITY, ZONE_DATA, ZONE_PKEY, ZONE_DEFAULT_FACET
+    ZONE_ENTITY, FORWARD_ZONE_ENTITY, ZONE_DATA, FORWARD_ZONE_DATA,
+    ZONE_PKEY, FORWARD_ZONE_PKEY, ZONE_DEFAULT_FACET
 )
 import pytest
 
@@ -41,26 +46,10 @@ class test_realmdomains(UI_driver):
         self.wait_for_request()
         self.close_notifications()
 
-    @screenshot
-    def test_read(self):
+    def prepare_dns_zone(self, realmdomain):
         """
-        Realm domains mod tests
+        Prepare dns zone record for realmdomain
         """
-        self.init_app()
-        self.navigate_to_entity(ENTITY)
-
-        # add with force - skipping DNS check
-        self.add_multivalued('associateddomain', 'itest.bar')
-        self.facet_button_click('save')
-        self.dialog_button_click('force')
-        self.wait_for_request()
-        self.close_notifications()
-
-        # delete
-        self.del_realm_domain('itest.bar', 'force')
-        self.wait_for_request()
-
-        # Try adding and deleting with "Check DNS" (in html 'ok' button)
 
         # DNS check expects that the added domain will have DNS record:
         #    TXT _kerberos.$domain "$REALM"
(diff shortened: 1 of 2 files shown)


celestian:
@felipevolpone , I would like to be sure that I will not damge PR CI again. Could you please add `re-run` label after you check that my configuration is right? Thank you.

felipevolpone:
It is great now! Thank you! Just an extra info: if you want, you can change the line
```yml
test_suite: test_webui/
```
to
```yml
test_suite: test_webui/test_realmdomains.py
````

to only run the tests you're adding. This way the feedback will be faster.

**Comment by celestian** [comment]
@felipevolpone ,
I don't understand. PR CI removed `re-run` label, but how you can see, no results here. :-(

**Comment by felipevolpone** [comment]
Hi @celestian , you actually have a duplicated `ipaserver` section on it. Just remove it and it should be fine.

**Comment by celestian** [comment]
Fixed.

**Comment by pvoborni** [comment]
@celestian , based on two "failure" screenshots, I think that  2 or maybe all 3 failures are caused by a notification covering a space of element which needs some interaction. Basically same cause as in https://github.com/freeipa/freeipa/pull/1737 (there is more detailed description) solution might be the same.

**Comment by celestian** [comment]
Thanks, @pvoborni I fixed it with your proposal and force pushed it.

**Comment by celestian** [comment]
@Rezney It seems that it is GREEN now :-)
So, should I squash (is this the word?) my commits to.... 2, for example?

**Comment by celestian** [comment]
Last run (only on webUI realmd domains) passed. So I squashed it, I added support for all tests. And re-run it.

File: ipatests/test_webui/test_realmdomains.py [hunk 1]
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Authors:
 #   Petr Vobornik <pvoborni@redhat.com>
```
felipevolpone: Do we need to change this?

felipevolpone:
The code and the test cases LGTM, but I'll leave the final ack to @pvoborni .

ipatests/test_webui/test_realmdomains.py
celestian: We have https://github.com/freeipa/freeipa/blob/master/Contributors.txt
So... what about to delete it? @pvoborni , do you agree?

pvoborni:
The way how the ticket is created mean that it will go only to master branch. Most of Web UI patches are being pushed also to ipa-4-6 branch to allow testing on system which won't have master (4.7). I guess that this PR should go to 4.6 as well. Right Petr? If so then, as all other patches which needs to be backported, it needs a ticket and commit messages must have a link to the ticket - this is useful for tracking which patch is where even if the backported patch is slightly modified. The future backport PR will also then depend on PR #1765

celestian:
Rebased....

celestian:
There some problems with space on disk. There should be fix on it, so I try rerun.

celestian:
Timeout was to short.

celestian:
@felipevolpone   I prolonged timeout to 10800, but it is not nice solution. How does the yaml key `test_suite` work? Is it possible to set something like:
```
test_suite: test_webui/test_dns, test_webui/test_group
```
If not, another solution is to group some of the test suite to the sub-folder and create more targets for PR CI.

pvoborni:
`undo_multivalued` was already added to UI driver before this PR was completed so now pylint complains: 
```text
************* Module ipatests.test_webui.ui_driver
ipatests/test_webui/ui_driver.py:891: [E0102(function-redefined), UI_driver.undo_multivalued] method already defined line 841)
```

felipevolpone:
@celestian yes, it's possible to have such thing. Another solution is to do what @Rezney did in his PR #1919 .

celestian:
I removed duplicity in methods and I used .yaml configuration provided in #1919 .
Pushed again...

pvoborni:
In #1919 the new test definition doesn't work yet, as can be observed by the failing tasks. Therefore, it won't add value here either till it is fixed.

celestian:
@pvoborni. Thank you, I known that. Meanwhile we found another 3 tests which should be added. :-)

celestian:
Nice, the way how to run webui test works!
Now, I need to add two missing tests.

celestian:
Two missing tests added.
Only WebUI tests enabled.
Higher version of PR CI image.
Commit message needs update.

celestian:
Please, what is wrong with travis here?

celestian:
@felipevolpone  Could you please add ACK here?

celestian:
Just note, new tests were green :-)

Rezney:
master:

* 3c9810e9639095779e9b4f7c8bf37de3d4b8922c webui:tests: Add tests for realmd domains

PR #1725, opened 2018-03-22
upgrade: Run configuration upgrade under empty ccache collection

abbra:
Use temporary empty DIR-based ccache collection to prevent upgrade
failures in case KCM: or KEYRING: ccache type is used by default in
krb5.conf and is not available. We don't need any user credentials
during upgrade procedure but kadmin.local would attempt to resolve
default ccache and if that's not available, kadmin.local will fail.

This approach was successfully tested with OpenQA tests that upgrade from F27 to F28.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1558818

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -11,6 +11,8 @@ import shutil
 import pwd
 import fileinput
 import sys
+import tempfile
+from contextlib import contextmanager
 from augeas import Augeas
 import dns.exception
 from ipalib import api, x509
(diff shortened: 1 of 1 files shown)


ipaserver/install/server/upgrade.py
```diff
@@ -1931,6 +1933,33 @@ def upgrade_check(options):
...
+    old_path = os.environ.get('KRB5CCNAME')
+    try:
+        os.environ['KRB5CCNAME'] = kpath
+        yield
+    finally:
+        if old_path:
+            os.environ['KRB5CCNAME'] = old_path
+        else:
+            del os.environ['KRB5CCNAME']
+        for f in os.listdir(kpath_dir):
```
tiran: Use ``shutil.rmtree()``, it removes the directory and all its children.

ipaserver/install/server/upgrade.py
```diff
@@ -1931,6 +1933,33 @@ def upgrade_check(options):
...
+    # instead of whatever is a default in /etc/krb5.conf
+    #
+    # In Fedora 28 KCM: became a default credentials cache collection
+    # but if KCM daemon (part of SSSD) is not running, libkrb5 will fail
+    # to initialize. This causes kadmin.local to fail.
+    # Since we are in upgrade, we cannot kinit anyway (KDC is offline).
+    # Bug https://bugzilla.redhat.com/show_bug.cgi?id=1558818
+    kpath_dir = tempfile.mkdtemp(prefix="upgrade_ccaches",
+                                 dir=paths.IPA_CCACHES)
+    kpath = "DIR:{dir}s".format(dir=kpath_dir)
```
tiran: The extra ``s`` after the directory name looks like a mistake.

``kpath = "DIR:{}".format(kpath)`` is shorter and easier to read.

abbra:
Updated to use shutil.rmtree() and simplify the formatting.

tiran:
Thanks, LGTM to me!

flo-renaud:
Hi @abbra 
thank you for the patch, works for me. I tested by doing the following:
- install on fedora 27
- systemctl stop sssd-kcm.socket; systemctl disable sssd-kcm.socket
- ipa-server-upgrade => success (without the patch, this would be failing)

PR #1726, opened 2018-03-22
[RFE] IPA server installation with successful message

amitkumar50:
This PR updates IPA success installation message as this:
'============================================='
IPA server installation successful

You can access Web UI using https://ipaserver1.testrelm.test/ipa/ui

Obtain a Kerberos Ticket for issuing IPA command 'kinit admin'

Server is installed with following roles :

* Directory Server
* Kerbeors Server
* Httpd Server
* CA Server
* DNS Server

For troubleshooting and errors :
Please visit : https://www.redhat.com/en

For documentation :
Please visit :
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/linux_domain_identity_authentication_and_policy_guide/index

Resolves: https://pagure.io/freeipa/issue/6355

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -944,7 +944,26 @@ def install(installer):
 
     print("======================================="
           "=======================================")
-    print("Setup complete")
+    print("IPA server installation successful")
+    print("")
+    print("You can access Web UI using https://%s/ipa/ui" % (host_name))
+    print("")
+    print("Obtain a Kerberos Ticket for issuing IPA Command 'kinit admin'.")
+    print("")
+    print("Server is installed with following roles:")
+    print(" * 389 Directory Server")
+    print(" * Kerberos Server")
+    print(" * HTTP Server")
+    if setup_ca:
+        print(" * CA Server")
+    if options.setup_dns:
+        print(" * DNS Server")
+    print("")
+    print("For troubleshooting and errors:")
+    print("Please visit: https://www.freeipa.org/page/Troubleshooting")
+    print("")
+    print("For documentation:")
+    print("Please visit: https://www.freeipa.org/page/Documentation")
     print("")
     print("Next steps:")
     print("\t1. You must make sure these network ports are open:")
@@ -961,14 +980,9 @@ def install(installer):
     if not options.no_ntp:
         print("\t\t  * 123: ntp")
     print("")
-    print("\t2. You can now obtain a kerberos ticket using the command: "
-          "'kinit admin'")
-    print("\t   This ticket will allow you to use the IPA tools (e.g., ipa "
-          "user-add)")
-    print("\t   and the web user interface.")
 
-    if not services.knownservices.chronyd.is_running():
-        print("\t3. Kerberos requires time synchronization between clients")
+    if not services.knownservices.ntpd.is_running():
+        print("\t2. Kerberos requires time synchronization between clients")
         print("\t   and servers for correct operation. You should consider "
               "enabling chronyd.")
 


ipaserver/install/server/install.py
```diff
@@ -940,9 +940,24 @@ def install(installer):
...
+    print("IPA server installation successful")
     print("")
-    print("Next steps:")
+    print("You can access Web UI using https://%s/ipa/ui" % (host_name))
+    print("\nObtain a Kerberos Ticket for issuing IPA command 'kinit admin'.")
+    print("\nServer is installed with following roles :")
+    print("* Directory Server")
+    print("* Kerbeors Server")
+    print("* Httpd Server")
+    print("* CA Server")
```
tiran: CA and DNS are optional components. They might not have been installed.

ipaserver/install/server/install.py
```diff
@@ -940,9 +940,24 @@ def install(installer):
...
+    print("* DNS Server")
```
tiran: Optional

ipaserver/install/server/install.py
```diff
@@ -940,9 +940,24 @@ def install(installer):
...
+    print("\nFor troubleshooting and errors :")
+    print("Please visit : https://www.redhat.com/en")
+    print("\nFor documentation :")
+    print("Please visit :")
+    print("https://access.redhat.com/documentation/en-us/"
+    "red_hat_enterprise_linux/7/html/"
```
tiran: Please indent correctly.

ipaserver/install/server/install.py
```diff
@@ -940,9 +940,24 @@ def install(installer):
 
     print("======================================="
           "=======================================")
-    print("Setup complete")
+    print("IPA server installation successful")
     print("")
-    print("Next steps:")
+    print("You can access Web UI using https://%s/ipa/ui" % (host_name))
+    print("\nObtain a Kerberos Ticket for issuing IPA command 'kinit admin'.")
```
tiran: Either add ``\n`` to the previous line or add an empty print() statement. I prefer the latter.

ipaserver/install/server/install.py
```diff
@@ -940,9 +940,24 @@ def install(installer):
...
+    print("\nServer is installed with following roles :")
+    print("* Directory Server")
+    print("* Kerbeors Server")
+    print("* Httpd Server")
```
tiran: ``HTTP server``.

ipaserver/install/server/install.py
```diff
@@ -940,9 +940,24 @@ def install(installer):
...
           "=======================================")
-    print("Setup complete")
+    print("IPA server installation successful")
     print("")
-    print("Next steps:")
+    print("You can access Web UI using https://%s/ipa/ui" % (host_name))
+    print("\nObtain a Kerberos Ticket for issuing IPA command 'kinit admin'.")
+    print("\nServer is installed with following roles :")
+    print("* Directory Server")
+    print("* Kerbeors Server")
```
tiran: ``Kerberos``, not ``Kerbeors``

ipaserver/install/server/install.py
```diff
@@ -940,9 +940,24 @@ def install(installer):
...
     print("======================================="
           "=======================================")
-    print("Setup complete")
+    print("IPA server installation successful")
     print("")
-    print("Next steps:")
+    print("You can access Web UI using https://%s/ipa/ui" % (host_name))
+    print("\nObtain a Kerberos Ticket for issuing IPA command 'kinit admin'.")
+    print("\nServer is installed with following roles :")
+    print("* Directory Server")
```
tiran: IMO ``LDAP Server`` is better because it's self-explaining.

ipaserver/install/server/install.py
```diff
@@ -940,9 +940,24 @@ def install(installer):
...
+    print("* Kerbeors Server")
+    print("* Httpd Server")
+    print("* CA Server")
+    print("* DNS Server")
+    print("\nFor troubleshooting and errors :")
+    print("Please visit : https://www.redhat.com/en")
```
tiran: For FreeIPA, we should rather point users to ``freeipa.org`` and only RH IdM users to ``redhat.com``.

ipaserver/install/server/install.py
felipevolpone: Hi @amitkumar50 , the correct indentation that @tiran is saying is:
```python
print("https://access.redhat.com/documentation/en-us/"
      "red_hat_enterprise_linux/7/html/"
      "linux_domain_identity_authentication_and_policy_guide/index")
```

ipaserver/install/server/install.py
felipevolpone: Maybe `389 Directory Server (LDAP Server)`?

ipaserver/install/server/install.py
amitkumar50: Cannot understand this. Because something this sort is logged after install.

IPA server installation successful

You can access Web UI using https://rhel7u4-6.gsslab.pnq2.redhat.com/ipa/ui

Obtain a Kerberos Ticket for issuing IPA command 'kinit admin'.


There is already an line displayed between them.

tiran: My review comment was solely about the placement of ``\n`` in your patch.
Semantically the newline char ``\n`` belongs to the previous line, not the next line. This comes from the era of [typewriters](https://youtu.be/oSyivQLQfQw?t=43s), where a new line followed by a carriage return terminated a line. An example will probably help more:

Instead of:
```
    print("text")
    print("\nmore test")
```

it is better to use
```
    print("text\n")
    print("more test")
```

or even better

```
    print("text")
    print()
    print("more test")
```

All three variants will print out the same text. I prefer the last variant because it makes the code look like the actual output.

ipaserver/install/server/install.py
```diff
@@ -940,9 +940,22 @@ def install(installer):
 
     print("======================================="
           "=======================================")
-    print("Setup complete")
+    print("IPA server installation successful")
     print("")
-    print("Next steps:")
+    print("You can access Web UI using https://%s/ipa/ui" % (host_name))
+    print("\nObtain a Kerberos Ticket for issuing IPA command 'kinit admin'.")
+    print("\nServer is installed with following roles :")
```
tiran: In English, there is no space between before a colon. There are three other cases of ``text :`` below.

ipaserver/install/server/install.py
```diff
@@ -940,9 +940,22 @@ def install(installer):
...
+    print("* 389 Directory Server")
+    print("* Kerberos Server")
+    print("* HTTP Server")
```
tiran: You have removed CA and DNS completely. That's not what I meant. You should print a line for CA and DNS, when the user has selected CA or DNS in the installer.

amitkumar50:
@tiran Are there any other changes required?

ipaserver/install/server/install.py
```diff
@@ -940,7 +940,29 @@ def install(installer):
 
     print("======================================="
           "=======================================")
-    print("Setup complete")
+    print("IPA server installation successful")
+    print("")
+    print("You can access Web UI using https://%s/ipa/ui" % (host_name))
+    print("")
+    print("Obtain a Kerberos Ticket for issuing IPA command 'kinit admin'.")
```
rcritten: command should be plural

ipaserver/install/server/install.py
```diff
@@ -940,7 +940,29 @@ def install(installer):
...
+    print("")
+    print("Server is installed with following roles:")
+    print("* 389 Directory Server")
```
rcritten: If you insist on printing all roles they should be indented.

ipaserver/install/server/install.py
```diff
@@ -940,7 +940,29 @@ def install(installer):
...
+    print("Please visit : https://www.freeipa.org")
+    print("")
+    print("For documentation:")
+    print("Please visit:")
+    print("https://access.redhat.com/documentation/en-us/"
+          "red_hat_enterprise_linux/7/html/"
+          "linux_domain_identity_authentication_and_policy_guide/index")
     print("")
     print("Next steps:")
     print("\t1. You must make sure these network ports are open:")
```
rcritten: There is only one step left so either this should be dropped completely or make the CA root cert backup step 2. Of course in the case of CAless this would revert to being one step.

If you are going to point to troubleshooting should it not go to https://www.freeipa.org/page/Troubleshooting ?

This looks to be more than a screenful of data. All masters have the LDAP, KDC and HTTP "roles", is it worth displaying that?

Is there any assurance from the content people that the documentation link will never change?

amitkumar50: ||There is only one step left so either this should be dropped completely or make the CA root cert backup step 2. Of course in the case of CAless this would revert to being one step.
I believe we can drop it.

||This looks to be more than a screenful of data. All masters have the LDAP, KDC and HTTP "roles", is it worth displaying that?
||Is there any assurance from the content people that the documentation link will never change?
Yes this need to be clarified. But Providing link to troubleshooting is good idea. Customer can/should investigate before opening an Bug. Again This page need to be updated all times with latest information. If you want to point to some other page, more upto date and maintained. I have no objections.

ipaserver/install/server/install.py
amitkumar50: I cannot understand your comment.
On console it prints as:

Server is installed with following roles:
* 389 Directory Server
* Kerberos Server
* HTTP Server
* CA Server
* DNS Server

ipaserver/install/server/install.py
```diff
@@ -940,7 +940,29 @@ def install(installer):
...
           "=======================================")
-    print("Setup complete")
+    print("IPA server installation successful")
+    print("")
+    print("You can access Web UI using https://%s/ipa/ui" % (host_name))
+    print("")
+    print("Obtain a Kerberos Ticket for issuing IPA Command 'kinit admin'.")
+    print("")
+    print("Server is installed with following roles:")
+    print("* 389 Directory Server")
```
amitkumar50: I was not able to understand your indentation comment here.

ipaserver/install/server/install.py
rcritten: Honestly I'd prefer to send them to some freeipa.org page that links to the RH docs by version. It will be a significantly shorter URL as well. Perhaps https://www.freeipa.org/page/Documentation

amitkumar50:
So do you want the change:
#    if not services.knownservices.ntpd.is_running():
    if not services.knownservices.chronyd.is_running():
        print("\t3. Kerberos requires time synchronization between clients")
        print("\t   and servers for correct operation. You should consider "
                "enabling chronyd.")

amitkumar50:
@rcritten waiting for inputs

rcritten:
This is just a question, I don't really have a strong opinion yet:

The extra text has some value but it this adds 16+ lines to the output which scrolls off a typical 24-line screen (the existing message fits basically a full screen).

So will this provide much added value if it isn't visible by default?

nit: I'd replace Command with commands, or perhaps something about identity.

amitkumar50:
@rcritten Yes I think will provide added value, One can see the roles installed at 1st place.
Also links for documentation and troubleshooting would be handy, customer may resolve some of issues by themselves.

ipaserver/install/server/install.py
```diff
@@ -961,14 +980,9 @@ def install(installer):
...
-    print("\t2. You can now obtain a kerberos ticket using the command: "
-          "'kinit admin'")
-    print("\t   This ticket will allow you to use the IPA tools (e.g., ipa "
-          "user-add)")
-    print("\t   and the web user interface.")
-
-    if not services.knownservices.chronyd.is_running():
-        print("\t3. Kerberos requires time synchronization between clients")
+
+    if not services.knownservices.ntpd.is_running():
```
Tiboris: I believe that this will work only with ntpd service not chronyd.

tiran:
The PR has been dead for a while, closing.

PR #1729, opened 2018-03-23
Fix compatibility with latest pytest

tiran:
pytest removed copy() method from its Namespace class. Use the copy
module to make a copy of early options.


stlaz:
We should probably remove `py35` from tox.ini too since I believe the build changes made us require python 3.6+

stlaz:
I'll create another PR for that, ACK.

tiran:
master:

* 48fb6d2c87aee50eb81a1678a33cecd89a4ea3d6 Fix compatibility with latest pytest

PR #1731, opened 2018-03-23
HTTPD encrypted key upgrade and backup

stlaz:
During my recent work on HTTPD key encryption, I forgot to create the password to the encrypted key during upgrade and also to back it up during `ipa-backup`. This PR fixes that.

https://pagure.io/freeipa/issue/7421


ipaserver/install/httpinstance.py
```diff
@@ -613,9 +613,20 @@ def migrate_to_mod_ssl(self):
             certs.install_pem_from_p12(temp.name,
                                        pk12_password,
                                        paths.HTTPD_CERT_FILE)
+            with open(
+                os.path.join(
+                    paths.IPA_PASSWD_DIR,
+                    HTTPD_PASSWD_FILE_FMT.format(host=api.env.host)
+                    ), 'wb') as passwd_file:
+                os.fchmod(passwd_file.fileno(), 0o600)
+                passwd_fname = passwd_file.name
```
tiran: This looks backward. How about you set ``passwd_fname`` first and then use it in ``with open()`` and ``certs.install_key_from_p12()``.

ipaserver/install/certs.py
```diff
@@ -72,12 +72,20 @@ def install_pem_from_p12(p12_fname, p12_passwd, pem_fname):
                  "-passin", "file:" + pwd.name])
 
 
-def install_key_from_p12(p12_fname, p12_passwd, pem_fname):
+def install_key_from_p12(
+    p12_fname, p12_passwd, pem_fname, out_passwd_fname=None
```
tiran: This indention style is uncommon and hard to read. I'd prefer if you indent it by 8 and keep ``):`` on line 76.

ipaserver/install/ipa_backup.py
```diff
@@ -365,6 +366,11 @@ def add_instance_specific_data(self):
             if os.path.exists(file):
                 self.files.append(file)
 
+        self.files.append(
+            os.path.join(paths.IPA_PASSWD_DIR,
+                         HTTPD_PASSWD_FILE_FMT.format(host=api.env.host))
```
tiran: I wonder, why did you add ``HTTPD_PASSWD_FILE_FMT`` to ``ipaplatform.constants`` instead of ``ipaplatform.paths`` as absolute path?

ipaserver/install/httpinstance.py
stlaz: Yeah, I noticed that, too, I wrote this way too fast to think about such things :) Will fix.

ipaserver/install/certs.py
stlaz: Ok.

ipaserver/install/ipa_backup.py
stlaz: It changes with the virtual host name, that's how the password reader works.

tiran: I mean, why did you decide to split the directory and file name into two? Instead of
```
os.path.join(paths.IPA_PASSWD_DIR,
             HTTPD_PASSWD_FILE_FMT.format(host=api.env.host))
```

why not simply
```
paths.HTTPD_PASSWD_FILE_FMT.format(host=api.env.host)
```
?

stlaz: Hm, right, that's probably easier. I suppose I was thinking the IPA_PASSWD_DIR could be useful once but I can merge it to constants instead.

tiran:
master:

* 47cf159f118de42ba703ad5fa0b47f5fa0bbb894 Fix upgrading of FreeIPA HTTPD
* e707d974a396b9b57e753038e57f3de90f9cd37c ipa_backup: Backup the password to HTTPD priv key

PR #1732, opened 2018-03-23
[backport ipa-4-6] upgrade: Run configuration upgrade under empty ccache collection

abbra:
Manual backport for ipa-4-6 as KCM is also a default ccache type in Fedora 27

Use temporary empty DIR-based ccache collection to prevent upgrade
failures in case KCM: or KEYRING: ccache type is used by default in
krb5.conf and is not available. We don't need any user credentials
during upgrade procedure but kadmin.local would attempt to resolve
default ccache and if that's not available, kadmin.local will fail.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1558818

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -11,6 +11,8 @@ import shutil
 import pwd
 import fileinput
 import sys
+import tempfile
+from contextlib import contextmanager
 from augeas import Augeas
 import dns.exception
 from ipalib import api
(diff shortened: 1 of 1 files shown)


tiran:
This is a manual backport of #1725 to 4.6.

tiran:
ipa-4-6:

* 7645dc26b4cfd91cc4e1a9ebdb6f627b6da46a2a upgrade: Run configuration upgrade under empty ccache collection

PR #1733, opened 2018-03-23
Corrected the size of the user profile-menu in the utility navbar.

edward-of-clt:
Rather minor, but it has bothered me since day of using IPA. Pushing upstream.

Before:
image

After: 
image


pvoborni:
@rcritten this issue is that the bar with username is higher and thus a separator line on the left of username doesn't go till the bottom end of the bar.

Anway, upstream and Fedora uses a variant with a blue bar on the top. The screenshot has a red bar. So it suggests that either RHEL or CentOS is used. Given that there is "FreeIPA" and not "Identity management" I'd say CentOS.  But I don't see the issue on FreeIPA upstream master. So the issue is either in RHEL branding patch or CentOS branding patch.  Afaik RHEL's 'Identity management' logo has smaller height then "FreeIPA" logo CentOS uses and that might be the culprit. CentOS replaced logo but not branding CSS. I don't think this issue can be fixed here - upstream. It needs to be fixed on CentOS side.

Even though it doesn't apply here, I'll just describe how a CSS change would be done upstream:

The patternfly.css FreeIPA uses is built using `install/ui/util/make-patternfly.sh` it uses `install/ui/less/patternfly.less` which require patternfly, bootstrap and font-awesome sources codes in related directory.  Any change done outside of this procedure would be eventually lost.

FreIPA specific CSS modifications can be done in less files imported in `install/ui/less/ipa.less`

edward-of-clt:
Thanks for the clarification @pvoborni.

rcritten:
@pvoborni so what should we do with this PR?

pvoborni:
Closing based on previous comments.

PR #1742, opened 2018-03-26
Cleanup and remove more files on uninstall

tiran:
* /etc/nsswitch.conf.ipabkp
* /etc/openldap/ldap.conf.ipabkp
* /var/lib/ipa/sysrestore/*
* /var/named/dyndb-ldap/ipa/

See: https://pagure.io/freeipa/issue/2694

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -1150,6 +1150,10 @@ def uninstall(installer):
                      'but it may\n'
                      'mean your system hasn\'t be restored to its '
                      'pre-installation state.', SYSRESTORE_DIR_PATH)
+    else:
+        # sysrestore.state has no state left, remove it
+        sysrestore = os.path.join(SYSRESTORE_DIR_PATH, 'sysrestore.state')
+        installutils.remove_file(sysrestore)
 
     # Note that this name will be wrong after the first uninstall.
     dirname = dsinstance.config_dirname(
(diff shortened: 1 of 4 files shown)


ipaserver/install/server/install.py
```diff
@@ -1150,6 +1150,11 @@ def uninstall(installer):
                      'but it may\n'
                      'mean your system hasn\'t be restored to its '
                      'pre-installation state.', SYSRESTORE_DIR_PATH)
+    else:
+        # cleanup remaining sysrestore artifacts but keep the directory
+        for basedir, dirnames, filenames in os.walk(SYSRESTORE_DIR_PATH):
+            for filename in filenames:
+                os.unlink(os.path.join(basedir, filename))
```
rcritten: I'm unsure about this bit. A few lines up a line is printed that files remain and then they are removed. Ideally all files are removed/restored during uninstallation. If there are any left over then that is the bug and cleaning them up just covers it up.

tiran:
@rcritten I added a comment to the section.

The files are only removed when ``sysrestore.state`` contains no more states to restore. The loop is only executed, when ``has_state`` is false. In that case, ``sysrestore.state`` still exists, but only contains the hostname of the machine. 

In some cases, there might be other files left. For example, one test system had two copies of ``ntp.conf`` after first installation has failed.

rcritten:
I agree on the point of left-over files but I think we should find the root cause of the issue and not cover it up by cleaning up the mess afterward. On the plus side if this is indeed related to mostly ntpd config files that problem will solve itself soon enough :-)

tiran:
OK, you convinced me. The latest version just removes ``sysrestore.state`` after all state information has been removed from the file.

tiran:
master:

* 64438f861947febe5ea10eda56ff89fd8900e80a Cleanup and remove more files on uninstall

PR #1747, opened 2018-03-27
Enable SPAKE support where avilable

frozencemetery:
Because krb5 silently ignores unrecognized options, this is safe on
all versions.  It lands upstream in krb5-1.17; in Fedora, it was added
in krb5-1.6-17.

Upstream documentation can be found in-tree at
https://github.com/krb5/krb5/blob/master/doc/admin/spake.rst

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -627,6 +627,16 @@ def configure_krb5_conf(
         filename, client_domain, client_hostname, force=False,
         configure_sssd=True):
 
+    # First, write a snippet to krb5.conf.d.  Currently this doesn't support
+    # templating, but that could be changed in the future.
+    template = os.path.join(
+        paths.USR_SHARE_IPA_DIR,
+        os.path.basename(paths.KRB5_FREEIPA) + ".template")
+    contents = open(template).read()
+    open(paths.KRB5_FREEIPA, "w+").write(contents)
+    os.chmod(paths.KRB5_FREEIPA, 0x644)
+
+    # Then, perform the rest of our configuration into krb5.conf itself.
     krbconf = IPAChangeConf("IPA Installer")
     krbconf.setOptionAssignment((" = ", " "))
     krbconf.setSectionNameDelimiters(("[", "]"))
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -21,6 +21,7 @@ import gssapi
 import netifaces
 import os
 import SSSDConfig
+import shutil
 import socket
 import sys
 import tempfile
@@ -631,9 +632,9 @@ def configure_krb5_conf(
     # templating, but that could be changed in the future.
     template = os.path.join(
         paths.USR_SHARE_IPA_DIR,
-        os.path.basename(paths.KRB5_FREEIPA) + ".template")
-    contents = open(template).read()
-    open(paths.KRB5_FREEIPA, "w+").write(contents)
+        os.path.basename(paths.KRB5_FREEIPA) + ".template"
+    )
+    shutil.copy(template, paths.KRB5_FREEIPA)
     os.chmod(paths.KRB5_FREEIPA, 0x644)
 
     # Then, perform the rest of our configuration into krb5.conf itself.
--- a/install/share/Makefile.am
+++ b/install/share/Makefile.am
@@ -51,6 +51,7 @@ dist_app_DATA =				\
 	kdc_req.conf.template		\
 	krb5.conf.template		\
 	krb5.ini.template		\
+	freeipa.template		\
 	krb.con.template		\
 	krbrealm.con.template		\
 	smb.conf.template		\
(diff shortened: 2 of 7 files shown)


frozencemetery:
CC @npmccallum, @simo5 (for awareness)

simo5:
Very nice, but I wonder if we shouldn't also enable at last one of the NIST curves ?
Or are we confident all implementations will be able to suppor edwards25519 ?

frozencemetery:
edwards25519 is provided by krb5 itself, so it's always available.

simo5:
I am talking about third party clients :-)

abbra:
We need to add some code to upgrade path. We already have augeas lense there in `ipaserver/install/server/upgrade.py` to perform upgrade to PKINIT, may be it can be extended.

tiran:
Thomas and I pointed out some issues with Augeas in #919. It doesn't handle kerberos config correctly. Is it possible to configure spake in e.g. ``/etc/krb5.conf.d/spake.conf`` instead of changing ``/etc/krb5.conf``?

```
# /etc/krb5.conf.d/spake.conf
[kdcdefaults]
  spake_preauth_kdc_challenge = edwards25519
```

abbra:
@tiran no, kdc.conf does not include /etc/krb5.conf.d.

frozencemetery:
@simo5 Per section 4: 

> Clients and KDCs MUST implement the edwards25519 group, but MAY choose not to offer or accept it by   default.

simo5:
Ah thanks, I missed that it was a MUST.

simo5:
@abbra actually afaik the KDC merges krb5.conf and kdc.conf in a single profile, so I surely hope it read /etc/krb5.conf.d

npmccallum:
Have any other krb5 implementations discussed implementing this?

frozencemetery:
@npmccallum Not to my knowledge, though we probably wouldn't know until they'd done it.

abbra:
@simo5  checked kdc code and yes, it uses profile loading code with a list of files to load, by adding the base krb5 configuration files to the base kdc configuration files. So yes, resulting input to a profile loader is something like /etc/krb5.conf:/var/kerberos/krb5kdc/kdc.conf  which means if krb5.conf includes /etc/krb5.conf.d/ files, they will be visible to kdc.conf too.
This should simplify the configuration also for PKINIT code.

frozencemetery:
Configuring krb5.conf.d is possible, but there's no code to do that in the freeipa client right now.  I've added some, but I'm not very familiar with this part of the code.  In general, there's a lot of stuff in the client that we should move out to krb5.conf.d and keep the amount of changes we make to krb5.conf proper small.

I've added upgrade logic as well.  Should I be checking for the krb5.conf.d file there as well, or will that be handled elsewhere?

tiran:
Could we ship the config files with freeipa-client package and have them installed on ``dnf install freeipa-client`` time? Or are the settings too intrusive?

frozencemetery:
@tiran That would probably be fine, but some of them will be duplicates of what I have in /etc/krb5.conf already (in the libkrb5 package).  Not an issue as long as we expect the number of people with freeipa-client installed who haven't run `ipa-client-install` to be small.

My concern is for the other downstreams - we'd want it to be clear that they need to package this file, and I don't know how reasonable that is.

simo5:
I do not think changing krb5 behavior just because a package is installed is sensible, besides we need domain specific data, so we'd still need to add ad hoc config on top of static files. Better to use a tamplate but store the results in /etc/krb5.d

npmccallum:
What is the reason that we are trying to enable this specifically in FreeIPA? Can't we just let FreeIPA depend upon the OS's configuration? IMHO, that would be preferred for now since we don't require it for any specific FreeIPA feature.

frozencemetery:
For the server, there's no notion of OS configuration - we ship a default file, but it has to be customized for the realm.

For the client, I would love it if freeIPA would leave the OS configuration alone.  But right now it doesn't do that - anything I configure in krb5.conf as part of libkrb5 gets stomped on by `ipa-client-install`.

frozencemetery:
Updated.  If there are further changes you'd like to see, please let me know.

abbra:
LGTM. Thanks!

ipaclient/install/client.py
```diff
@@ -627,6 +627,16 @@ def configure_krb5_conf(
         filename, client_domain, client_hostname, force=False,
         configure_sssd=True):
 
+    # First, write a snippet to krb5.conf.d.  Currently this doesn't support
+    # templating, but that could be changed in the future.
+    template = os.path.join(
+        paths.USR_SHARE_IPA_DIR,
+        os.path.basename(paths.KRB5_FREEIPA) + ".template")
+    contents = open(template).read()
```
tiran: Use ``shutil.copyfile()`` to copy files.

npmccallum:
Congrats everyone! This has been a long time coming. Thanks to everyone who has been a part of this process!

tiran:
master:

* 792adebfabb456d154164387fb7e60acb30f4325 Enable SPAKE support using krb5.conf.d snippet
* 73c3495db2dc1df38407aa7b49fed2f222e85bbe Use shutil to copy file

PR #1748, opened 2018-03-27
Fix certificate retrieval in ipa-replica-prepare for DL0

rcritten:
The NSSDatabase attrs for the NSS database were NULL because the database hadn't been created yet.

### DIFF
--- a/ipaserver/install/certs.py
+++ b/ipaserver/install/certs.py
@@ -569,8 +569,16 @@ class CertDB(object):
         ])
 
     def create_from_cacert(self):
+        """
+        Ensure that a CA chain is in the NSS database.
+
+        If an NSS database already exists ensure that the CA chain
+        we want to load is in there and if not add it. If there is no
+        database then create an NSS database and load the CA chain.
+        """
         cacert_fname = paths.IPA_CA_CRT
-        if os.path.isfile(self.certdb_fname):
+
+        if self.nssdb.exists():
             # We already have a cert db, see if it is for the same CA.
             # If it is we leave things as they are.
             with open(cacert_fname, "r") as f:


ipaserver/install/certs.py
```diff
@@ -574,6 +574,7 @@ def export_pkcs12(self, pkcs12_fname, pkcs12_pwd_fname, nickname=None):
         ])
 
     def create_from_cacert(self):
+        self.nssdb.create_db()
         cacert_fname = paths.IPA_CA_CRT
         if os.path.isfile(self.certdb_fname):
```
tiran: ``if self.exists():`` is a better way to check if the NSSDB already exists.

tiran: I'm suggesting to (also) fix ``os.path.isfile(self.certdb_fname)``. The property ``self.certdb_fname`` redirects to ``self.nssdb.certdb``. The value ``certdb`` of a ``NSSDatabase`` instance is ``None``, in case the NSS DB doesn't exist yet. I had to change the implementation to support DBM and SQL databases. The attribute `certdb` is filled when either a DB already exists or a new DB is created using ``create_db``.

``os.path.isfile(self.certdb_fname)`` was the old way to check if ``cert8.db`` is present. The new way is ``self.exists()``. It makes use of DBM/SQL auto-detection and checks for the presence of all NSS DB related files (cert, key, secmod).

frasertweedale: I misread the code above, seeing `os.path.isfile(cacert_fname)` instead of what was actually written.

So, given we're now (as of this commit) *creating* the NSSDB, what is the point of retaining the code for looking up the cert.  The NSSDB will be empty.

So can't we just create the NSSDB and import the CA cert?  Or am I missing something...

tiran: In general it looks like a good idea to always create a DB here -- but only if the DB does not exist yet. The ``create_db`` method replaces any previous DB. The method deletes or renames all DB related files.

Therefore the ``create_db`` call must be protected by a ``if not self.exists()`` call.

frasertweedale: Right.  I agree.  @rcritten please guard the `create_db` in an `if not self.exists(): ...`.
Then the `if os.path.isfile(self.certdb_fname):` conditional can be removed and the
code in that block moved out one level.

Aside from that, I've tested the code and it does what it says on the tin :)

rcritten: This is never expected to work on an existing database, the name after all is create_from_cacert(). The idea is you are given a CA cert and you end up with an NSS database containing only that cert.

This method is only used once as well, in this one DL0 case.

Still, safety first. I can add a guard and a doc describing how this method is used.

ipaserver/install/certs.py
```diff
@@ -581,8 +581,16 @@ def export_pkcs12(self, pkcs12_fname, pkcs12_pwd_fname, nickname=None):
...
+        Ensure that a CA chain is in the NSS database.
+
+        If an NSS database already exists ensure that the CA chain
+        we want to load is in there and if not add it. If there is no
+        database then create an NSS database and load the CA chain.
+        """
         cacert_fname = paths.IPA_CA_CRT
-        if os.path.isfile(self.certdb_fname):
+
+        if self.nssdb.exists():
```
frasertweedale: Now we are missing the create, right?

Shouldn't it be something like:
```python
if not self.nssdb.exists():
  self.nssdb.create_db()
... # the rest
```

rcritten: It not self.nssdb.exists() it falls down to:

        self.create_passwd_file()
        self.create_certdbs()
        self.load_cacert(cacert_fname, IPA_CA_TRUST_FLAGS)

felipevolpone:
This PR fixes the problem that ticket https://pagure.io/freeipa/issue/7469 describes. `test_topology::TestReplicaManageDel::test_replica_managed_del_domlevel0` is now failing for other reasons (DNS-related)

rcritten:
Fraser was my reasoning ok? Are we good to push this?

frasertweedale:
@rcrit yeah I'm cool with the code, just got one integration test that's failing.  Unsure if related to the
change (probably not?)  Hitting it with the re-run hammer to see if it comes good...

felipevolpone:
Hi @rcritten, yes `test_replica_managed_del_domlevel0` is broken due to DNS issues. I have a PR (#1831 ) that fixes it. If you have some time to review it, we can move forward with this PR and the other one.

felipevolpone:
master:

* 573f13228da7d711d3b22fa7f78f4a78e199288b Fix certificate retrieval in ipa-replica-prepare for DL0

PR #1749, opened 2018-03-28
Fix upgrade when named.conf does not exist

frasertweedale:
Commit aee0d2180c7119bef30ab7cafea81dc3df1170b7 adds an upgrade step
that adds system crypto policy include to named.conf.  This step
omitted the named.conf existence check; upgrade fails when it does
not exist.  Add the existence check.

Part of: https://pagure.io/freeipa/issue/4853

### DIFF
--- a/ipatests/test_ipaserver/test_install/test_bindinstance.py
+++ b/ipatests/test_ipaserver/test_install/test_bindinstance.py
@@ -24,7 +24,6 @@ options {
 include "random/file";
 """
 
-
 EXPECTED_CONFIG = """
 options {
 \tdnssec-enable yes;
@@ -33,6 +32,12 @@ options {
 };
 
 include "random/file";
+"""
+
+# bindinstance.named_conf_exists() looks for a section like this
+IPA_DYNDB_CONFIG = """
+dyndb "ipa" "/usr/lib/bind/ldap.so" {
+};
 """
 
 POLICY_FILE = "/etc/crypto-policies/back-ends/bind.config"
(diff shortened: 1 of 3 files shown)


tiran:
The test will still fail. I sent you a patch for your patch.

ipatests/test_ipaserver/test_install/test_bindinstance.py
```diff
@@ -53,6 +62,7 @@ def test_add_crypto_policy(m_set, m_get, namedconf):
     m_get.return_value = False
     with open(namedconf, 'w') as f:
         f.write(TEST_CONFIG)
+        f.flush()
```
tiran: That shouldn't be necessary. The caches are flushed when the file is closed.

ipatests/test_ipaserver/test_install/test_bindinstance.py
```diff
@@ -33,6 +33,15 @@
 };
 
 include "random/file";
+
+dyndb "ipa" "/usr/lib/bind/ldap.so" {
```
tiran: You have to add ``dyndb "ipa"`` to ``TEST_CONFIG``, too. The test uses the value of ``TEST_CONFIG`` as mocked data. Check out inbox.

frasertweedale: Haah, thanks!  I wasn't writing the data.  Adding the `flush()` was just because I thought maybe it wasn't being flushed.  Silly me, I did not update all the right data :)

tiran:
Did you receive my patch file? I couldn't push to your branch, so I made a patch for you.

frasertweedale:
@tiran thanks, I'm incorporating your suggested changes.

tiran:
master:

* 421fc376ccb8668c07692d3a3394a5869dc97296 Fix upgrade when named.conf does not exist

tiran:
thanks, @frasertweedale

PR #1752, opened 2018-03-28
Allow anonymous access to parentID attribute

abbra:
Due to optimizations in 389-ds performed as result of
https://pagure.io/389-ds-base/issue/49372, LDAP search filter
is rewritten to include parentID information. It implies that parentID
has to be readable for a bound identity performing the search. This is
what 389-ds expects right now but FreeIPA DS instance does not allow it.

As result, searches with a one-level scope fail to return results that
otherwise are matched in a sub scope search.

While 389-ds developers are working on the fix for issue
https://pagure.io/389-ds-base/issue/49617, we can fix it by adding an
explicit ACI to allow reading parentID attribute at the suffix level.

Fixes: https://pagure.io/freeipa/issue/7466


tbordaz:
The patch looks good to me.
Note that granting parentid access should be enough but I have not tested it myself

abbra:
Pushed an update due to a syntax error (too many brackets) ;)

tiran:
master:

* 34d06b2be71823bc8898732f1ced0185f83afb01 Allow anonymous access to parentID attribute

PR #1753, opened 2018-03-28
Fix installer CA port check for port 8080

tiran:
The installer now checks that port 8080 is available and not in use by
any other application.

The port checker has been rewritten to use bind() rather than just
checking if a server responds on localhost. It's much more reliable and
detects more problems.

Original patch by m3gat0nn4ge.

Co-authored-by: Mega Tonnage <m3gat0nn4ge@gmail.com>
Fixes: https://pagure.io/freeipa/issue/7415

Replaces PR #1722

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -1037,6 +1037,55 @@ def host_port_open(host, port, socket_type=socket.SOCK_STREAM,
     return port_open
 
 
+def check_port_bindable(port, socket_type=socket.SOCK_STREAM):
+    """Check if a port is free and not bound by any other application
+
+    :param port: port number
+    :param socket_type: type (SOCK_STREAM for TCP, SOCK_DGRAM for UDP)
+
+    Returns True if the port is free, False otherwise
+    """
+    if socket_type == socket.SOCK_STREAM:
+        proto = 'TCP'
+    elif socket_type == socket.SOCK_DGRAM:
+        proto = 'UDP'
+    else:
+        raise ValueError(socket_type)
+
+    # Detect dual stack or IPv4 single stack
+    try:
+        s = socket.socket(socket.AF_INET6, socket_type)
+        anyaddr = '::'
+        logger.debug(
+            "check_port_bindable: Checking IPv4/IPv6 dual stack and %s",
+            proto
+        )
+    except socket.error:
+        s = socket.socket(socket.AF_INET, socket_type)
+        anyaddr = ''
+        logger.debug("check_port_bindable: Checking IPv4 only and %s", proto)
+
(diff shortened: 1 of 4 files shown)


tiran:
@m3gat0nn4ge

rcritten:
Would be nice to have a unit test for this. I hacked up something in a few minutes, feel free to improve on it: https://paste.fedoraproject.org/paste/W2xXEaPd2hpfI7qT72bhWQ

m3gat0nn4ge:
Well I can honestly say I have no experience writing unit tests... I tested
yours Rob, with the asserts changed to use ipautil.host_port_free(...) and
'make fasttests' was successful, and the logic looks good. I don't have any
suggestions to improve it...

tiran:
@rcritten thanks for the patch.

I have used your patch and changed it a bit to use pytest fixtures. During testing I actually found a bug in my implementation. The UDP port checker didn't work. Your patch helped me to figure out the root cause. ``SO_REUSEADDR`` makes no sense on UDP server sockets.

ipapython/ipautil.py
```diff
@@ -1037,6 +1037,55 @@ def host_port_open(host, port, socket_type=socket.SOCK_STREAM,
     return port_open
 
 
+def check_port_bindable(port, socket_type=socket.SOCK_STREAM):
+    """Check if a TCP port is free and not bound by any other application
```
rcritten: nit: drop TCP

ipatests/test_ipapython/test_ipautil.py
```diff
@@ -508,3 +509,40 @@ def test_run_stderr():
...
     assert "message" not in str(cm.value.output)
     assert "message" not in str(cm.value.stderr)
+
+
+@pytest.fixture(scope='function')
+def tcp_listen():
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    try:
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, True)
+        s.bind(('', 0))  # 0 means random port
```
rcritten: Will random port lead to random test failures? We can always accept it as-is and reconsider if we see periodic failures.

tiran: With port = 0, the OS will use a random, unused, and free port. It won't fail unless all ports > 1024 are exhausted. It's a useful to trick for tests.

ipapython/ipautil.py
tiran: OK

tiran:
master:

* 6aca027ecc5c1bbcd1a69deea10d2ff991c53f5e Fix installer CA port check for port 8080

PR #1754, opened 2018-03-28
Add commentary about PKI admin password

frasertweedale:
Add a note in cainstance.configure_instance that "admin_password" is
the password to be used for the PKI admin account, NOT the IPA admin
password.  In fact, it is set to the Directory Manager password.

This comment would have saved me some time during recent
investigation of a replica installation issue.


tiran:
master:

* 0176e1a68a5b5fc30cb753d6334ddeca8ce3f8c2 Add commentary about PKI admin password

tiran:
I short circuited CI. It doesn't make sense to burn countless CPU hours for a comment-only PR.

frasertweedale:
Thanks @tiran!

PR #1758, opened 2018-03-28
Fix TestSubCAkeyReplication providing the right path to pki log

felipevolpone:
The PKI debug log has its name in this format: `debug.<date>.log`. This commit changes the code to use this format, fixing the test.

Unfortunately, it's not possible to use some kind of regex (like debug.*.log) to get the file, because python multihost gets the path and tries to open (using the `open` python function) the file with that.


felipevolpone:
As usual, once this PR get an ack, I'll remove the temporary commit.

flo-renaud:
Hi @felipevolpone,
Thank you for the fix. You can link this PR with ticket [7095](https://pagure.io/freeipa/issue/7095):  [tracker] please rotate & compress /var/lib/pki/pki-tomcat/logs/ca/debug, since it is a consequence of the fix in dogtag.
Please do not forget to remove the temp commit.

felipevolpone:
Hi @flo-renaud, thank you for reviewing my PR :) I changed the commit message to include the ticket and removed the temporary commit. Please, check if it's ok.

flo-renaud:
Hi @felipevolpone 
Thanks for updating the PR. Temp commit removed, tests passed, ACK'ed, means that we can push now.

PR #1766, opened 2018-04-03
Change remove_segment to take a config object

rcritten:
This is to handle the case where there is no integration test
YAML provided where config.domains is an empty list.

https://pagure.io/freeipa/issue/7483

Note that I considered populating config with an empty set of domains and replicas but that would have required much more code and IMHO been more fragile. Given there is only a single caller to remove_segment I think this is fastest, safest route. If we end up with more callers needing it we can revisit.

The use-case for this is to:

- Install IPA
- Install the test pkgs
- mkdir ~/.ipa
- ln -s /etc/ipa/default.conf ~/.ipa/default.conf
- ln -s /etc/ipa/ca.crt ~/.ipa/ca.crt
- kinit admin
- ipa-run-tests

The simplest possible way to execute the tests in IPA.

### DIFF
--- a/ipatests/test_integration/test_topology.py
+++ b/ipatests/test_integration/test_topology.py
@@ -27,7 +27,7 @@ def find_segment(master, replica):
             return '-to-'.join(segment)
 
 
-def remove_segment(master, host1, host2):
+def remove_segment(config):
     """
     This removes a segment between host1 and host2 on master. The function is
     needed because test_add_remove_segment expects only one segment, but due to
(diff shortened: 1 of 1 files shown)


rcritten:
I'm wondering if a test for the YAML should be more generic. It affects other integration tests as well. It should probably be some top-level decorator in the integration tests that require it.

Something like:

config = get_global_config()
skip_no_integration = pytest.mark.skipif(len(config.domains) == 0,
                                       reason="Integration test configuration not set")

@skip_no_integration
class TestIntegrationSomething

rcritten:
I prefer Petr's method, closing this and the related ticket.

PR #1768, opened 2018-04-04
Freeipa correcting contributors

amitkumar50:


tiran:
The names are sorted alphabetically by last name. This is common in Western European countries.

rcritten:
Given the point of this PR was to re-sort the list in a a way we do not want I'm closing it.

PR #1771, opened 2018-04-04
Fix order of commands in test for removing topology segments

pvoborni:
this is an alternative approach to PR #1766 

test_topology_updated_on_replica_install_remove from the beginning used
invalid sequence of commands for removing a replica.

Proper order is:
  master$ ipa server-del $REPLICA
  replica$ ipa-server-install --uninstall

Alternatively usage of `ipa-replica-manage del $replica` instead of
`ipa server-del $replica` is possible. In essence ipa-replica-manage
calls the server-del command.

At some point there  was a plan to achieve uninstalation only through
`ipa-server-install --uninstall` but that was never achieved to this
date.

This change also removes the ugly wrapper which makes test collection
fail if no environment config is provided (i.e. replicas cannot be
indexed).
  $ pytest --collect-test ipatests/test_integration

https://pagure.io/freeipa/issue/6250

### DIFF
--- a/ipatests/test_integration/test_topology.py
+++ b/ipatests/test_integration/test_topology.py
@@ -27,26 +27,6 @@ def find_segment(master, replica):
             return '-to-'.join(segment)
 
 
-def remove_segment(master, host1, host2):
-    """
-    This removes a segment between host1 and host2 on master. The function is
-    needed because test_add_remove_segment expects only one segment, but due to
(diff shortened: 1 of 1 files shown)


tiran:
master:

* 7b1b0b35ea2c89e57c037ac57c453c322af42f0c Fix order of commands in test for removing topology segments

PR #1772, opened 2018-04-04
Check if ldap_uri is on /etc/ipa/default.conf

felipevolpone:
When removing a replica there are cases when ldap_uri is not on the
/etc/ipa/default.conf file anymore. So, before trying to get the value
of it, now the code checks if it's there first.

https://pagure.io/freeipa/issue/7474

This PR fixes the test `test_replica_promotion.py::TestReplicaPromotionLevel0::test_promotion_disabled `

Full log:
https://fedorapeople.org/groups/freeipa/prci/jobs/48fd7274-3162-11e8-8f04-fa163efc0cae/report.htm

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -195,3 +195,14 @@ jobs:
         timeout: 10800
         topology: *master_1repl
 
+  fedora-27/test_replica_promotion_TestReplicaPromotionLevel0:
+    requires: [fedora-27/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-27/build_url}'
+        test_suite: test_integration/test_replica_promotion.py::TestReplicaPromotionLevel0
+        template: *ci-master-f27
+        timeout: 8000
+        topology: *master_1repl
--- a/ipaserver/secrets/kem.py
+++ b/ipaserver/secrets/kem.py
@@ -213,7 +213,7 @@ class IPAKEMKeys(KEMKeysStore):
         if conf.read(ipaconf):
             self.host = conf.get('global', 'host')
             self.realm = conf.get('global', 'realm')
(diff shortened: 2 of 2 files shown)


simo5:
@tiran beats getting a python stack trace, but we may want to return an error in this case, not just leave ldap_uri empty, otherwise we'll get hard to diagnose issues later on if we try to access the variable expecting a value.

tiran:
Traceback:

```
  File "/usr/lib/python3.6/site-packages/ipaserver/install/server/__init__.py", line 589, in main
    uninstall(self)
  File "/usr/lib/python3.6/site-packages/ipaserver/install/server/install.py", line 252, in decorated
    func(installer)
  File "/usr/lib/python3.6/site-packages/ipaserver/install/server/install.py", line 1107, in uninstall
    custodiainstance.CustodiaInstance().uninstall()
  File "/usr/lib/python3.6/site-packages/ipaserver/install/custodiainstance.py", line 72, in uninstall
    'ldap_uri': self.ldap_uri
  File "/usr/lib/python3.6/site-packages/ipaserver/secrets/kem.py", line 217, in __init__
    self.ldap_uri = conf.get('global', 'ldap_uri', raw=True)
  File "/usr/lib64/python3.6/configparser.py", line 792, in get
    raise NoOptionError(option, section)
```

I'd rather change ``custodiainstance.CustodiaInstance().uninstall()`` to either use ldapi URI or not use LDAP at all. As far as I remember, 389-DS is already stopped when ``uninstall`` is called.

simo5:
If ldap_uri is not needed on uninstall, indeed just avoid setting it at all in that case.

tiran:
The uninstaller needs a keystore instance in order to delete the local key files. The LDAP URI isn't required but the ``IPAKEMKeys`` class requires it.

simo5:
Set it to a harmless value during uninstall() ?

PR #1773, opened 2018-04-04
ctypes.util.find_library() replacement

m3gat0nn4ge:
component: ipaclient/csrgen_ffi.py,  ipaserver/p11helper.py, ipaplatform/redhat/tasks.py
   
   Create a dumbed down replacement for ctypes.util.find_library().. parses ldconfig output
   
   https://pagure.io/freeipa/issue/6851

### DIFF
--- a/ipaplatform/base/fl.py
+++ b/ipaplatform/base/fl.py
@@ -18,26 +18,26 @@
 #
 
 '''
-This base module contains a find_library() replacement for 
+This base module contains a find_library() replacement for
  ctypes.util.find_library.
 '''
 
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import os
 import subprocess
 import logging
 
 logger = logging.getLogger(__name__)
 
+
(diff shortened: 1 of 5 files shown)


ipaplatform/base/fl.py
```diff
@@ -0,0 +1,52 @@
...
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def find_library(name):
+    logger.debug("Checking ldconfig output for: %s", name)
+    process = subprocess.Popen(
+          "ldconfig -N -p | awk '{print $1, $NF}' | grep '^lib%s\.so.*'"
+          % name, shell=True, stdout=subprocess.PIPE,
```
tiran: This code is slower and less secure than ``find_library`` from ctypes. You are executing at least four programs. ``find_library`` starts only one binary. ``shell=True`` is also a security issue and makes the code vulnerable to shell injection attacks.

PR #1775, opened 2018-04-04
Break out of teardown in test_replica_promotion.py if no config

rcritten:
These tests are all skipped if there is no YAML configuration
file passed but the teardown method is always called and since
there is a reference to the Config object this blows up if just
ipa-run-tests is executed.

Look at the config and break out if no domains are set.


rcritten:
To test run:

$ ipa-run-tests-3 test_integration/test_replica_promotion.py
============================= test session starts ==============================
platform linux -- Python 3.6.4, pytest-3.2.3, py-1.4.34, pluggy-0.4.0
rootdir: /home/rcrit/redhat/freeipa/dist/rpms, inifile:
plugins: sourceorder-0.5, multihost-1.1
collected 23 items                                                              

. sssssssssssssssssssssss

========================== 23 skipped in 0.09 seconds ==========================

Tiboris:
Works as described, thanks for the patch.

PR #1776, opened 2018-04-04
Provide ldap_uri in Custodia uninstaller

tiran:
Without ldap_uri, IPAKEMKeys parses /etc/ipa/default.conf. During
uninstallation, the file may no longer contain ldap_uri. This workaround
is required for test case
test_replica_promotion.py::TestReplicaPromotionLevel0::test_promotion_disabled

Fixes: https://pagure.io/freeipa/issue/7474
Co-authored-by: Felipe Barreto <fbarreto@redhat.com>

Alternative solution for #1772


frasertweedale:
Is this the agreed approach, @tiran @felipevolpone?  i.e. should #1772 be closed?

tiran:
@frasertweedale yes, that's my plan. @felipevolpone 's patch fixes the issue, too. But his approach may also hide a bug in the configuration.

frasertweedale:
Did an install and uninstallation, no worries.  All the tests are passing and the code looks fine so here's my ACK.

PR #1779, opened 2018-04-05
Defer import of ipaclient.csrgen

tiran:
The modules ipaclient.csrgen and ipaclient.csrgen_ffi are expensive to load,
but rarely used. On demand loading speeds up ipa CLI by about 200ms.

Fixes: https://pagure.io/freeipa/issue/7484

### DIFF
--- a/ipaclient/plugins/cert.py
+++ b/ipaclient/plugins/cert.py
@@ -23,7 +23,6 @@ import base64
 
 import six
 
-from ipaclient import csrgen
 from ipaclient.frontend import MethodOverride
 from ipalib import errors
 from ipalib import x509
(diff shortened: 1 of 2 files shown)


abbra:
LGTM.

ipaclient/plugins/cert.py
```diff
@@ -105,6 +104,10 @@ def get_args(self):
             yield arg
 
     def forward(self, csr=None, **options):
+        # Deferred import, because ipaclient.csrgen is expensive to load
+        # see https://pagure.io/freeipa/issue/7484
+        from ipaclient import csrgen
+
```
frasertweedale: I think we could defer this even further, to the `if csr is None` suite (about 10 lines down from here).  Then the import will only be incurred when the csrgen feature is actually being used, instead of on all `ipa cert-request` invocations.

tiran: Good catch!

tiran:
master:

* 395a68d20887d0ac010e480e68b225d6dfeff726 Defer import of ipaclient.csrgen

PR #1780, opened 2018-04-05
 ui_tests: workaround for data disappearing

Rezney:
Some data disappear from user details page after
the save action is performed.

https://pagure.io/freeipa/issue/5776


Rezney:
Closing this PR in favor of PR1782 and also because of the workaround was not reliable in cases where we were later asserting undo buttons which were gone after refresh.

PR #1781, opened 2018-04-05
 certdb: Move chdir into subprocess call

tiran:
According to a comment, certutil may create files in the current working
directory. Rather than changing the cwd of the current process,
FreeIPA's certutil wrapper now changes cwd for the subprocess only.

See: pagure.io/freeipa/issue/7416

## Remove os.chdir() from test_ipap11helper

test_ipap11helper no longer changes directory for the entire test suite.
The fix revealed a bug in another test suite. test_secrets now uses a
proper temporary directory.

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -297,7 +297,9 @@ class NSSDatabase(object):
         ]
         new_args.extend(args)
         new_args.extend(['-f', self.pwd_file])
-        return ipautil.run(new_args, stdin, **kwargs)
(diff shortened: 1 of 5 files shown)


tiran:
Tests are stable again. The full test suite passed two times without any flaky test.

tiran:
master:

* 807a5cbe7cc52690336c5095ec6aeeb0a4e8483c certdb: Move chdir into subprocess call
* 1b320ac3e7ab763d932512f2c497288711bc09e8 Remove os.chdir() from test_ipap11helper

PR #1782, opened 2018-04-05
webui: refresh complex pages after modification

pvoborni:
Details facet for user, hosts, service, user override entities require
complex reload as they gather information from multiple sources - e.g.
all of them do cert-find. On update only $entity-mod is execute and its
result doesn't have all information required for refresh of the page
therefore some fields are missing or empty.

This patch modifies the facets to do full refresh instead of default
load and thus the pages will have all required info.

https://pagure.io/freeipa/issue/5776

### DIFF
--- a/install/ui/src/freeipa/host.js
+++ b/install/ui/src/freeipa/host.js
@@ -518,6 +518,12 @@ IPA.host.details_facet = function(spec, no_init) {
         return that.entity.name+'_show_'+that.get_pkey();
     };
 
+    that.update_on_success = function(data, text_status, xhr) {
+        that.on_update.notify();
+        that.nofify_update_success();
+        that.refresh();
+    };
+
     if (!no_init) that.init_details_facet();
 
     return that;
--- a/install/ui/src/freeipa/idviews.js
+++ b/install/ui/src/freeipa/idviews.js
@@ -450,6 +450,12 @@ idviews.id_override_user_details_facet = function(spec) {
         return batch;
     };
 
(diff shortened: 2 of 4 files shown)


pvoborni:
I didn't like workaround for this issue in PR #1780 as it was for all entities and not only the ones which suffer from the issue. So this patch should make the PR obsolete. The plus side is that it will be fixed for everybody.  

PR-CI run with Web UI tests was executed at https://github.com/pvoborni/freeipa/pull/14

Rezney:
Tested manually in all 4 cases and LGTM. @felipevolpone would you be able to find some time and test from your side too? Thanks

pvoborni:
It would make sense to define a shared function in some parent facet so that the children can just use it so that we don't need to copy the same code 3 times. But in these three cases we are overriding behavior of `datails_facet` and there isn't any class in between. I did not want to create a switch of behavior `details_facet`'s implementation as it would complicate it without real benefit.

Maybe an introduction of new facet base class, e.g., `multi_details_facet` , which would inherit from details facet and these 3 would inherit from it would solve the situation. But then it would deserve a small refactoring in all these facets to simplify the `create_refresh_command` method to be part of base class as well given that it is the reason why this change was needed.

I don't think it is worth the effort.

felipevolpone:
Ok then, let's go with the simplest solution :) :+1:

Rezney:
master:

* b43c2f8ab4769a473ba81b23321acda7fd57e94c webui: refresh complex pages after modification

PR #1784, opened 2018-04-06
replica-install: pass --ip-address to client install

stlaz:
In replica DL1 installation, the --ip-address option was not passed
down to the ipa-client-install script (when not promoting client).
This resulted in creating DNS records for all of the host's interface
IP adresses instead of just those specified.

This patch passes all the --ip-address options down to the client
installation script.

https://pagure.io/freeipa/issue/7405

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -947,6 +947,10 @@ def ensure_enrolled(installer):
         args.append("--force-join")
     if installer.no_ntp:
         args.append("--no-ntp")
+    if installer.ip_addresses:
+        for ip in installer.ip_addresses:
+            # installer.ip_addresses is of type [CheckedIPAddress]
+            args.extend(("--ip-address", str(ip)))
 
     try:
         # Call client install script


pvoborni:
AFAIK, the same issue is also in `ipa-server-install` - we should modify also `ipaserver/install/server/install.py::install`

stlaz:
`ipa-server-install` as well as `ipa-replica-install` on DL0 don't perform the client-install code that updates the DNS records at all.
Actually, from reading the man, I am not sure about this - is it expected that only the specified IPs get their DNS records added? Did this behavior change or was it always so that the specified IPs were only added on top of all the other records?

rcritten:
Here is my quickie read on it.

For ipa-server-install:

From what I can tell if you pass in a set of addresses then /etc/hosts gets updated.

If you pass in a set AND setup bind then only those addresses are added to DNS (install_dns_records)

For ipa-replica-install:

If promoted then hosts is NOT updated but DNS will be if there is an IPA bind instance.

If full install then hosts IS updated and DNS is also updated if there is a bind instance.

Looking at the ipa-client-install log that was executed from within ipa-replica-install it added 4 addresses to DNS for the host (2 IPv4 and 2 IPv6) but the options for ipa-replica-install only called for adding two of them. So I think this patch is doing the right thing.

pvoborni:
I see, you are right.  I thought that on the master client also calls `update_dns` method, but that is part of `client_dns` which is called only if `on_master` is false.

Tiboris:
Works for me :) thanks for the patch!

Tiboris:
master:

* 8246d0cd5a9fc079ff2eb6bfd9fe841f9d416592 replica-install: pass --ip-address to client install

PR #1789, opened 2018-04-09
Require more recent glibc on F27

tiran:
On CPUs with AVX-512 instruction set, ntpd sometimes segfaults because
PTHREAD_STACK_MIN is too small. The bug has been fixed in
glibc-2.26-24.fc27.x86_64 or later.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1564527

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -879,6 +879,11 @@ Requires: python2-sssdconfig
 Requires: python2-cryptography >= 1.6
 Requires: iptables
 Requires: python2-mock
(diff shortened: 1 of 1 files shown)


slaykovsky:
Hello @tiran 

Thank you for this patch, tests are more stable now. I will ack it as fast as Travis will finish the run.

tiran:
master:

* 888d9861f86f27531753cc53644b098dc395098d Require more recent glibc on F27

PR #1559, opened 2018-02-09
test_dnssec: re-add named-pkcs11 workarounds

tiran:
DNSSEC tests starrted to fail again, probably due to a bug in
some underlaying component.

This reverts commit 8bc677512296a7e94c29edd0c1a96aa7273f352a
and makes the xfail test check less strict - it will no longer
mark the test suite red if it passes.

Related https://pagure.io/freeipa/issue/5348

Clone of @tomaskrizek PR #973

PR #1793, opened 2018-04-09
ipa tests: Set default TTL for the IPA zone to 1 second

abbra:
When running IPA tests, a default TTL for the zone should be set
very low to allow get rid of timeouts in the tests. Zone updates should
be propagated to the clients as soon as possible.

This is not something that should be used in production so the change is
done purely at install time within the tests. As zone information is
replicated, we only modify it when creating a master with integrated
DNS.

This change should fix a number of DNSSEC-related tests where default
TTL is longer than what a test expects and a change of DNSSEC keys
never gets noticed by the BIND. As result, DNSSEC tests never match
their expected output with what they received from the BIND.

### DIFF
--- a/ipatests/pytest_plugins/integration/tasks.py
+++ b/ipatests/pytest_plugins/integration/tasks.py
@@ -19,6 +19,7 @@
 
 """Common tasks for FreeIPA integration tests"""
 
+from __future__ import absolute_import
 import logging
 import os
 import textwrap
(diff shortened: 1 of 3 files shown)


tiran:
+1

What about the reverse zone? Do we have reverse zones in our test env?

abbra:
Yes, I think we have. Let us run this through PR CI to see if DNSSEC tests can indeed be fixed by the proposed changes and then I can add the reverse zone support too.

abbra:
Note that it would probably not help DNSSEC tests specifically as they add a test zone, so more commits will be needed.

ipatests/pytest_plugins/integration/tasks.py
```diff
@@ -268,6 +269,21 @@ def enable_replication_debugging(host, log_level=0):
                      stdin_text=logging_ldif)
 
 
+def set_default_ttl_for_ipa_dns_zone(host, raiseonerr=True):
+    args = [
+        'ipa dnszone-mod',
```
tiran: ``'ipa', 'dnszone-mod',``

abbra: yes, fixed. Found it the hard way :)

tiran:
Test is still failing. I'm seeing ``bug in zone_master_reconfigure(): default TTL cannot be changed at run-time`` in the journal.

abbra:
@tiran I think that particular message can be ignored:
```
        result = setting_update_from_ldap_entry("default_ttl", zone_settings,
                                                "DNSdefaultTTL", entry);
        if (result == ISC_R_SUCCESS)
                log_bug("default TTL cannot be changed at run-time");
        else if (result != ISC_R_IGNORE)
                goto cleanup;
```

tiran:
Your improvement has been landed as part of PR https://github.com/freeipa/freeipa/pull/1559 . Thanks! :100:

PR #1798, opened 2018-04-10
[Backport][ipa-4-5] replica-install: pass --ip-address to client install

stlaz:
In replica DL1 installation, the --ip-address option was not passed
down to the ipa-client-install script (when not promoting client).
This resulted in creating DNS records for all of the host's interface
IP adresses instead of just those specified.

This patch passes all the --ip-address options down to the client
installation script.

https://pagure.io/freeipa/issue/7405


tiran:
ipa-4-5:

* 55b313984a4d28bdd31c23250c28144eda58ea59 replica-install: pass --ip-address to client install

PR #1800, opened 2018-04-11
Add nsds5ReplicaReleaseTimeout to replica config

tiran:
The nsds5ReplicaReleaseTimeout setting prevents the monopolization of
replicas during initial or busy master-master replication. 389-DS
documentation suggets a timeout of 60 seconds to improve convergence of
replicas.

See: http://directory.fedoraproject.org/docs/389ds/design/repl-conv-design.html
Fixes: https://pagure.io/freeipa/issue/7488

### DIFF
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -474,20 +474,30 @@ class ReplicationManager(object):
 
         try:
             entry = conn.get_entry(dn)
+        except errors.NotFound:
+            pass
+        else:
             managers = {DN(m) for m in entry.get('nsDS5ReplicaBindDN', [])}
 
+            mods = []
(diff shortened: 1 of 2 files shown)


flo-renaud:
Hi @tiran 
thanks for the patch, works for me. I tested upgrade of a master, of a replica, as well as fresh master/replica install, and ipa-ca-install. The attribute is properly added in all the cases.

tiran:
Thanks @flo-renaud !

I'll try to create a test case for the fix, too. We don't have any existing tests for replication data in cn=config.

tiran:
master:

* afc0d4b62d043cd568ce87400f60e8fa8273495f Add nsds5ReplicaReleaseTimeout to replica config

PR #1805, opened 2018-04-12
Handle whitespace, add separator to regex in set_directive_lines

rcritten:
We added the separator to the regex in set_directive_lines to avoid
grabbing just a prefix. This doesn't allow for whitespace around
the separator.

For the Apache case we expected that the separator would be just
spaces but it can also use tabs (like Ubuntu 18). Add a special
case so that passing in a space separator is treated as whitespace
(tab or space).

https://pagure.io/freeipa/issue/7490

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -565,11 +565,15 @@ def set_directive_lines(quotes, separator, k, v, lines, comment):
         v_quoted = quote_directive_value(v, '"') if quotes else v
         new_line = ''.join([k, separator, v_quoted, '\n'])
 
+    # Special case: consider space as "white space" so tabs are allowed
+    if separator == ' ':
+        separator = '[ \t]+'
+
     found = False
     addnext = False  # add on next line, found a comment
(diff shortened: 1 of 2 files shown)


flo-renaud:
Hi @rcritten 
The code change looks good to me. The only issue I see is that the new test does not exhibit the issue (if you run it on the code without the fix, the test passes). You may want to add a test removing or modifying a directive that contains a tab.

rcritten:
Excellent catch, new test added

tjaalton:
seems to work here, thanks

tiran:
@rcritten ``set_directive`` looks ok, but what about ``get_directive``? It can't handle ``\t`` yet.

rcritten:
Good catch, I hadn't considered that. Currently looks like it throws an exception:

ValueError: Malformed directive: SSLCertificateFile     /etc/ssl/certs/ssl-cert-snakeoil.pem

flo-renaud:
@rcritten 
thanks for the updated PR, works for me. Let's wait for @tiran approval.

PR #1808, opened 2018-04-12
Fixes for ipatests.test_integration.test_caless.TestCertInstall

rcritten:
Don't change working directories when calling pk12util because this breaks passing in non-absolute filenames (regression)

Fix argument passed into load_pkcs12, s/realm/realm_name/

https://pagure.io/freeipa/issue/7489

NOTE: Don't push without removing the addition of the caless test to .freeipa-pr-ci.yaml

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -308,7 +308,7 @@ class NSSDatabase(object):
             "-d", '{}:{}'.format(self.dbtype, self.secdir)
         ]
         new_args.extend(args)
-        return ipautil.run(new_args, stdin, cwd=self.secdir, **kwargs)
+        return ipautil.run(new_args, stdin, **kwargs)
 
     def exists(self):
         """Check DB exists (all files are present)
--- a/ipaserver/install/ipa_server_certinstall.py
+++ b/ipaserver/install/ipa_server_certinstall.py
@@ -174,7 +174,7 @@ class ServerCertInstall(admintool.AdminTool):
     def replace_kdc_cert(self):
         # pass in `realm` to perform `NSSDatabase.verify_kdc_cert_validity()`
         cert, key, ca_cert = self.load_pkcs12(
-            ca_chain_fname=paths.CA_BUNDLE_PEM, realm=api.env.realm)
+            ca_chain_fname=paths.CA_BUNDLE_PEM, realm_name=api.env.realm)
 
         self.replace_key_cert_files(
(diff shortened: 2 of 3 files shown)


ipaserver/install/ipa_server_certinstall.py
```diff
@@ -174,7 +174,7 @@ def replace_http_cert(self):
     def replace_kdc_cert(self):
         # pass in `realm` to perform `NSSDatabase.verify_kdc_cert_validity()`
         cert, key, ca_cert = self.load_pkcs12(
-            ca_chain_fname=paths.CA_BUNDLE_PEM, realm=api.env.realm)
+            ca_chain_fname=paths.CA_BUNDLE_PEM, realm_name=api.env.realm)
```
stlaz: Darn, I did this :roll_eyes:

ipapython/certdb.py
```diff
@@ -308,7 +308,7 @@ def run_pk12util(self, args, stdin=None, **kwargs):
             "-d", '{}:{}'.format(self.dbtype, self.secdir)
         ]
         new_args.extend(args)
-        return ipautil.run(new_args, stdin, cwd=self.secdir, **kwargs)
+        return ipautil.run(new_args, stdin, **kwargs)
```
stlaz: We cannot make sure the paths being passed are absolute here, so LGTM.

stlaz:
The `test_caless` test suite is failing because it seems to be expecting relative path in the output, but it's getting absolute path instead.

rcritten:
I was seeing these path-related failures locally too but thought it might be related to my config. I'll see about generalizing that output or looking for another related change (it wasn't reported as failing this way last week).

slaykovsky:
Hello @rcritten 

The `test_caless` seems to be okay. Please remove your changes to `.freeipa_pr_ci.yaml` and I will ack PR.

Thanks!

rcritten:
master:

* 41352ef9388dbba408762d64b60a6a1f53048bcd Revert run_pk12util part of 807a5cbe7cc52690336c5095ec6aeeb0a4e8483c
* 138ae4abe7d4737ae2e1c6526476b8be009178db ipa-server-certinstall failing, unknown option realm
* 3384147ca1b69f64d197869577e39e7c056bcd51 Some PKCS#12 errors are reported with full path names
* 4919bd9dae0222509af6ebf4f50ff7602b48c334 Remove xfail from CALes test test_http_intermediate_ca

PR #1809, opened 2018-04-13
Fixes for test_server_del failures.

menonsudhir:
Fixes for test_server_del failures.
Modified assert text for the failures seen in TC

Signed-off-by: Sudhir Menon <sumenon@redhat.com>

### DIFF
--- a/ipatests/test_integration/test_server_del.py
+++ b/ipatests/test_integration/test_server_del.py
@@ -32,6 +32,8 @@ def check_master_removal(host, hostname_to_remove,
         assert "Ignoring topology connectivity errors." in result.stderr_text
 
     if ignore_last_of_role:
+        assert ("Deleting this server will leave your installation "
+                "without a DNS." in result.stderr_text)
         assert ("Ignoring these warnings and proceeding with removal" in
                 result.stderr_text)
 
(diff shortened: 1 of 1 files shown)


rcritten:
PEP-8 errors:
./ipatests/test_integration/test_server_del.py:35:80: E501 line too long (85 > 79 characters)

pvoborni:
Just skimmed through. It seems weird that in tests: `test_removal_of_master_raises_error_about_dnssec` and `test_removal_of_master_raises_error_about_last_ca` the assert text is being changed to "Deleting this server will leave your installation without a DNS".  I.e. one test should test CA role, the other DNSSec but assert is about DNS role. I.e. in effect the tests is testing something different and thus the change is not right.

pvoborni:
Btw, the reason why the test started to fail is probably be3ad1ed7a34e90c7107380bb2939f737306ba77 where the checks were reordered. `TestLastServices` relies on execution of tests in a specific order. So checks were changed but tests weren't. 

Given that master is install with DNS and CA and replica with anything and given that checks are in order: DNS, DNSSec, CA, KRA then the test should be something like:

1. install master (with DNS, CA)
2. install replica
3. test test_removal_of_master_raises_error_about_last_dns 
4. test_install_dns_on_replica1_and_dnssec_on_master (installing DNS and DNSSec will allow CA check)
5. test_removal_of_master_raises_error_about_dnssec
6. test_removal_of_master_raises_error_about_last_ca

We could remove `test_install_ca_on_replica1` as it doesn't add value. It was needed before for DNS checks.

pvoborni:
I just quickly put together the theory in my comment above with @felipevolpone 's PR CI recommendation. It's in PR #1913

pvoborni:
PR #1913 now obsoletes this PR.

rcritten:
Closing due to inactivity.

PR #1812, opened 2018-04-16
Fix upgrade (update_replica_config) in single master mode

frasertweedale:
Commit afc0d4b62d043cd568ce87400f60e8fa8273495f added an upgrade
step that add an attribute to a replica config entry.  The entry
only exists after a replica has been added, so upgrade was broken
for standalone server.  Catch and suppress the NotFound error.

Related to: https://pagure.io/freeipa/issue/7488


tiran:
master:

* 7c8fd5630da2de5d3c88cd5fec7787427259f123 Fix upgrade (update_replica_config) in single master mode

PR #1813, opened 2018-04-16
csrgen: fix when attribute shortname is lower case

frasertweedale:
OpenSSL requires attribute short names ("CN", "O", etc) to be in upper
case, otherwise it fails to add the attribute.  This can be triggered when
FreeIPA has been installed with --subject-base containing a lower-case
attribute shortname (e.g.
--subject-base="o=Red Hat").

Explicitly convert the attribute type string to an OID
(ASN1_OBJECT *).  If that fails, upper-case the type string and try again.

Add some tests for the required behaviour.

This PR also fixes a python3 str/bytes issue in the csrgen error handling.

Fixes: https://pagure.io/freeipa/issue/7496

### DIFF
--- a/ipaclient/csrgen.py
+++ b/ipaclient/csrgen.py
@@ -402,20 +402,31 @@ class CSRLibraryAdaptor(object):
 
 
 class OpenSSLAdaptor(object):
-    def __init__(self, key_filename, password_filename):
-        self.key_filename = key_filename
-        self.password_filename = password_filename
+    def __init__(self, key=None, key_filename=None, password_filename=None):
+        """
+        Must provide either ``key_filename`` or ``key``.
+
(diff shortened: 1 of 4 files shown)


tiran:
master:

* 53f87ee5cd9d19f6fb91a9a1eafc8ea798095954 py3: fix csrgen error handling
* 7633d62d858c14523a99143aa0ff36f76bb4ff68 csrgen: support initialising OpenSSL adaptor with key object
* 0ac1d3ea62efd9751fcc59cea46bcdafe1f11c37 csrgen: drive-by docstring
* 852618fd6529fbdd7b03077fae37c6fbbe45b51b csrgen: fix when attribute shortname is lower case

PR #1814, opened 2018-04-16
Change cronyd option from -q to -Q

Tiboris:
chronyd run with -q option is meant to start chronyd daemon
syncinng to server or pool provided as another option.
option -q also needs root privileges to star daemon.

For syncing time with another server than configured using
already running daemon is used option -Q.

Addresses: https://pagure.io/freeipa/issue/7487


Tiboris:
Ah i should have paid more attention when reading the manual.

PR #1815, opened 2018-04-16
upgrade: treat duplicate entry when updating as not an error

abbra:
When we attempt to update an entry during upgrade, it may have already
contain the data in question between the check and the update. Ignore
the change in this case and record it in the log.

Fixes: https://pagure.io/freeipa/issue/7450

### DIFF
--- a/ipaserver/install/ldapupdate.py
+++ b/ipaserver/install/ldapupdate.py
@@ -21,6 +21,7 @@
 
 # TODO
 # save undo files?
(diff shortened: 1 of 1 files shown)


rcritten:
So he was running updates simultaneously on multiple masters?

I thought this situation was supposed to raise MidairCollision.

abbra:
It raised DuplicateEntry instead and we didn't handle it.

PR #1821, opened 2018-04-17
[backport ipa-4-5] upgrade: treat duplicate entry when updating as not an error

abbra:
When we attempt to update an entry during upgrade, it may have already
contain the data in question between the check and the update. Ignore
the change in this case and record it in the log.

Fixes: https://pagure.io/freeipa/issue/7450
(cherry picked from commit 52c0d8c5bfc4d0c1fe8d8318edbf5909d139178c)

### DIFF
--- a/ipaserver/install/ldapupdate.py
+++ b/ipaserver/install/ldapupdate.py
@@ -21,6 +21,7 @@
 
 # TODO
 # save undo files?
+from __future__ import absolute_import
 
 import base64
 import sys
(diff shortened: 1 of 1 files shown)


abbra:
logging infra did change between 4.5 and master, adapt to older code.

tiran:
ipa-4-5:

* 2824c1755a0734df87afbf24c86ce1cadd6ccb54 upgrade: treat duplicate entry when updating as not an error

PR #1824, opened 2018-04-17
Adding test-cases for ipa-cacert-manage

amore17:
File        :       ipatests/test_integration/test_external_ca.py
    line        :       224 to 282

    Scenario1:  Manual renew external CA cert with invalid file
                when ipa-server is installed with external-ca
                and renew with invalid cert file the renewal
                should fail.

    Scenario2:  install CA cert manually
                Install ipa-server. Create rootCA, using
                ipa-cacert-manage install option install
                new cert from RootCA

### DIFF
--- a/ipatests/test_integration/test_external_ca.py
+++ b/ipatests/test_integration/test_external_ca.py
@@ -17,6 +17,7 @@
 
 from __future__ import absolute_import
 
+import os
 import re
 import time
 
@@ -25,6 +26,7 @@ from ipatests.test_integration.base import IntegrationTest
 from ipaplatform.paths import paths
 
 from itertools import chain, repeat
+from ipatests.pytest_plugins.integration.create_external_ca import ExternalCA
 
 IPA_CA = 'ipa_ca.crt'
 ROOT_CA = 'root_ca.crt'
@@ -220,3 +222,60 @@ class TestExternalCAdirsrvStop(IntegrationTest):
         tasks.kinit_admin(self.master)
         result = self.master.run_command(['ipa', 'user-show', 'admin'])
         assert 'User login: admin' in result.stdout_text
+
+
+class TestExternalCAInvalidCert(IntegrationTest):
+    """Manual renew external CA cert with invalid file"""
+
+    def test_external_ca(self):
+        # Step 1 of ipa-server-install.
+        install_server_external_ca_step1(self.master)
+
+        # Sign CA, transport it to the host and get ipa a root ca paths.
+        root_ca_fname, ipa_ca_fname = tasks.sign_ca_and_transport(
+            self.master, paths.ROOT_IPA_CSR, ROOT_CA, IPA_CA)
+
+        # Step 2 of ipa-server-install.
+        install_server_external_ca_step2(self.master, ipa_ca_fname,
+                                         root_ca_fname)
+
+        self.master.run_command([paths.IPA_CACERT_MANAGE, 'renew',
+                                 '--external-ca'])
+        result = self.master.run_command(['grep', '-v', 'CERTIFICATE',
+                                          ipa_ca_fname])
+        contents = result.stdout_text
+
+        BAD_CERT = 'bad_ca.crt'
+        invalid_cert = os.path.join(self.master.config.test_dir, BAD_CERT)
+        self.master.put_file_contents(invalid_cert, contents)
+        # Sign CA, transport it to the host and get ipa a root ca paths.
+        root_ca_fname, ipa_ca_fname = tasks.sign_ca_and_transport(
+            self.master, paths.IPA_CA_CSR, ROOT_CA, IPA_CA)
+        # renew CA with invalid cert
+        cmd = [paths.IPA_CACERT_MANAGE, 'renew', '--external-cert-file',
+               invalid_cert, '--external-cert-file', root_ca_fname]
+        result = self.master.run_command(cmd, raiseonerr=False)
+        assert result.returncode == 1
+
+
+class TestExternalCAInstall(IntegrationTest):
+    """install CA cert manually """
+
+    def test_install_master(self):
+        # step 1 install ipa-server
+
+        tasks.install_master(self.master)
+
+    def test_install_external_ca(self):
+        # Create root CA
+        external_ca = ExternalCA()
+        # Create root CA
+        root_ca = external_ca.create_ca()
+        root_ca_fname = os.path.join(self.master.config.test_dir, ROOT_CA)
+
+        # Transport certificates (string > file) to master
+        self.master.put_file_contents(root_ca_fname, root_ca)
+
+        # Install new cert
+        self.master.run_command([paths.IPA_CACERT_MANAGE, 'install',
+                                 root_ca_fname])


ipatests/test_integration/test_external_ca.py
```diff
@@ -218,3 +219,66 @@ def test_external_ca_dirsrv_stop(self):
...
         assert 'User login: admin' in result.stdout_text
+
+
+class TestExternalCAInvalidCert(IntegrationTest):
+    """Manual renew external CA cert with invalid file"""
+
+    def test_external_ca(self):
+        # Step 1 of ipa-server-install.
+        result = install_server_external_ca_step1(self.master)
+        assert result.returncode == 0
```
tiran: You don't need result object or the assert check here. ``install_server_external_ca_step1`` uses ``host.run_command()`` with default ``raiseonerr=True``

ipatests/test_integration/test_external_ca.py
```diff
@@ -218,3 +219,66 @@ def test_external_ca_dirsrv_stop(self):
...
+
+        # Sign CA, transport it to the host and get ipa a root ca paths.
+        root_ca_fname, ipa_ca_fname = tasks.sign_ca_and_transport(
+            self.master, paths.ROOT_IPA_CSR, ROOT_CA, IPA_CA)
+
+        # Step 2 of ipa-server-install.
+        result = install_server_external_ca_step2(
+            self.master, ipa_ca_fname, root_ca_fname)
+        assert result.returncode == 0
```
tiran: Same here

ipatests/test_integration/test_external_ca.py
```diff
@@ -218,3 +219,66 @@ def test_external_ca_dirsrv_stop(self):
...
+
+        result = self.master.run_command([paths.IPA_CACERT_MANAGE, 'renew',
+                                          '--external-ca'])
+        assert result.returncode == 0
```
tiran: Same here, ``run_command`` uses ``raiseonerr=True`` by default.

ipatests/test_integration/test_external_ca.py
```diff
@@ -218,3 +219,66 @@ def test_external_ca_dirsrv_stop(self):
...
+        result = self.master.run_command(['grep', '-v', 'CERTIFICATE', ipa_ca_fname])
+        contents = result.stdout_text
+        invalid_cert = os.path.join(self.master.config.test_dir, BAD_CERT)
+        self.master.run_command(['touch', invalid_cert])
```
tiran: Why is ``touch`` necessary here?

ipatests/test_integration/test_external_ca.py
```diff
@@ -218,3 +219,66 @@ def test_external_ca_dirsrv_stop(self):
...
+        self.master.put_file_contents(invalid_cert, contents)
+        # Sign CA, transport it to the host and get ipa a root ca paths.
+        root_ca_fname, ipa_ca_fname = tasks.sign_ca_and_transport(
+        self.master, paths.IPA_CA_CSR, ROOT_CA, IPA_CA)
+        # renew CA with invalid cert
+        result = self.master.run_command([paths.IPA_CACERT_MANAGE, 'renew',
```
tiran: You can make the code more readable with a different indentation and argument passing style.

```
        result = self.master.run_command(
            [paths.IPA_CACERT_MANAGE, 'renew',
             '--external-cert-file', invalid_cert,
             '--external-cert-file={}', root_ca_fname],
            raiseonerr=False
        )
```

ipatests/test_integration/test_external_ca.py
```diff
@@ -218,3 +219,66 @@ def test_external_ca_dirsrv_stop(self):
...
+
+
+class TestExternalCAInstall(IntegrationTest):
+    """install CA cert manually """
+
+    def test_install_master(self):
+        # step 1 install ipa-server
+
+        result = tasks.install_master(self.master)
+        assert result.returncode == 0
```
tiran: ``install_master()`` also uses ``raiseonerr=True``

ipatests/test_integration/test_external_ca.py
```diff
@@ -218,3 +219,66 @@ def test_external_ca_dirsrv_stop(self):
...
+        external_ca = ExternalCA()
+        # Create root CA
+        root_ca = external_ca.create_ca()
+        root_ca_fname = os.path.join(self.master.config.test_dir, ROOT_CA)
+
+        # Transport certificates (string > file) to master
+        self.master.put_file_contents(root_ca_fname, root_ca)
+
+        # Install new cert
+        result = self.master.run_command([paths.IPA_CACERT_MANAGE, 'install', format(root_ca_fname)], raiseonerr=False)
```
tiran: `` format(root_ca_fname)`` looks wrong to me. Did you mean to just pass ``root_ca_fname``?

Just rely on ``raiseonerr=True`` instead of combining ``raiseonerr=False`` with an assert.

ipatests/test_integration/test_external_ca.py
```diff
@@ -21,12 +21,13 @@
...
 from ipaplatform.paths import paths
-
+import os
 from itertools import chain, repeat
+from ipatests.pytest_plugins.integration.create_external_ca import ExternalCA
 
 IPA_CA = 'ipa_ca.crt'
 ROOT_CA = 'root_ca.crt'
-
+BAD_CERT = 'bad_ca.crt'
```
tiran: You don't need to define this as global constant, if you just use it inside one function.

ipatests/test_integration/test_external_ca.py
```diff
@@ -21,12 +21,13 @@
 from ipatests.pytest_plugins.integration import tasks
 from ipatests.test_integration.base import IntegrationTest
 from ipaplatform.paths import paths
-
+import os
```
tiran: Standard library imports belong in the top import section. Please put it before ``import re``.

ipatests/test_integration/test_external_ca.py
```diff
@@ -218,3 +217,62 @@ def test_external_ca_dirsrv_stop(self):
...
         result = self.master.run_command(['ipa', 'user-show', 'admin'])
         assert 'User login: admin' in result.stdout_text
+
+
+class TestExternalCAInvalidCert(IntegrationTest):
+    """Manual renew external CA cert with invalid file"""
+
+    def test_external_ca(self):
+        # Step 1 of ipa-server-install.
+        result = install_server_external_ca_step1(self.master)
```
tiran: Don't assign to ``result`` unless you need the object later. You can and should remove ``result = `` in five out of six occasions. IIRC the code linter won't accept unused assignments, too.

Rezney:
Apart from the Christian`s review also please address:

`PEP-8 errors:
./ipatests/test_integration/test_external_ca.py:244:80: E501 line too long (85 > 79 characters)
./ipatests/test_integration/test_external_ca.py:251:9: E122 continuation line missing indentation or outdented
./ipatests/test_integration/test_external_ca.py:257:80: E501 line too long (83 > 79 characters)
./ipatests/test_integration/test_external_ca.py:281:80: E501 line too long (119 > 79 characters)
./ipatests/test_integration/test_external_ca.py:284:1: W391 blank line at end of file`

tiran:
LGTM

Let's see if Travis likes your PR, too.

pvoborni:
The PR is not linked to any Pagure ticket. It means that it will be pushed only to master branch. I assume that we want it also in 4.6 branch. And therefore a pagure ticket is required (to track backports).

The commit message contains: 
```
Patch2:
	1)	As per review assert statements are removed
		on line 230,239,243,268
	2)	line 247 is removed
	3)	line 281 raiseonerr=True is used
	4) 	global constant is used in function
	5)	import os is moved to before import re

Patch3:
	1)	unused result object is removed
	2) 	removed blank line from the end of file

Patch4:
	1)	changes in line 247 with diffrent argument passing
```
There is only 1 patch in this PR so the message is confusing for future readers. It's very good that you wrote the changes you made. But it might be better to write them in pull request as commetns and keep only the end state and reasons in the commit message.


For scenario 2. What is the goal of the test? Is the goal to testing that the command finishes successfully or do we want to test also that the certificates are installed correctly and that certificate related operations will work? Right now, we the tests checks command run. It might be OK for this pull request but we should create the latter later as well.

amore17:
@pvoborni   as per your suggestion PR message is changed.
For scenario 2 the goal is to check install command is working or not,
In this PR I just added basic functionality for ipa-cacert-manage
for  certificates related functionality i'll add that later.

rcritten:
This just needs a rebase and then we can push it.

flo-renaud:
Given the PR has been rebased and all concerns addressed, I'm ACK'ing

PR #1825, opened 2018-04-17
install: validate AD trust-related options in installers

abbra:
We already validate that --setup-dns is specified when any of
DNS-related options provided by a user. Do the same for --setup-adtrust
case.

Fixes: https://pagure.io/freeipa/issue/7410


abbra:
I removed check for `add_sids` because we force it to `True` always in the installers.

abbra:
Removed `rid_base` and `secondary_rid_base` checks too. As I said, we already always set `add_sids` to `True` unconditionally and checking it makes no sense.

tiran:
master:

* 64ffd117d2b4938e0935f9ea00d5eb90c0e7b3e6 install: validate AD trust-related options in installers

PR #1826, opened 2018-04-17
Fix format string passed to pytest-multihost

Tiboris:
Integration trust test suit failed with error trying to
start chronyd because of bad formating of passed string

See: https://pagure.io/python-pytest-multihost/issue/15
Resolves: https://pagure.io/freeipa/issue/7487


tiran:
master:

* 692a9931da326d2de00255c494786f90929843eb Fix format string passed to pytest-multihost

PR #1829, opened 2018-04-18
ACL: Allow hosts to remove services they manage

abbra:
Allow hosts to delete services they own. This is an ACL that complements
existing one that allows to create services on the same host.

Add a test that creates a host and then attempts to create and delete a
service using its own host keytab.

Fixes: https://pagure.io/freeipa/issue/7486


rcritten:
master:

* 2de1aa27f981060acd6a867c89ef963a18cc27df ACL: Allow hosts to remove services they manage

PR #1830, opened 2018-04-18
certprofile: reject config with multiple profileIds

frasertweedale:
In certprofile-import if the config file contains two profileId directives
with different values, with the first matching the profile ID CLI argument
and the second differing, the profile gets imported under the second ID.
This leads to:

- failure to enable the profile
- failure to add the IPA "tracking" certprofile object
- inability to delete the misnamed profile from Dogtag (via ipa CLI)

To avert this scenario, detect and reject profile configurations where
profileId is specified multiple times (whether or not the values differ).

https://pagure.io/freeipa/issue/7503

### DIFF
--- a/ipaserver/plugins/certprofile.py
+++ b/ipaserver/plugins/certprofile.py
@@ -236,14 +236,25 @@ class certprofile_import(LDAPCreate):
         ca_enabled_check(self.api)
         context.profile = options['file']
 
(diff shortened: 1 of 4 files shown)


rcritten:
master:

* a7b18372ed0f6be95e382194ad599b8a35113351 certprofile: reject config with multiple profileIds
* 0f8593354d8008a501a6e7fa5f142d970face65a certprofile: add tests for config profileId scenarios

PR #1838, opened 2018-04-19
ui_tests: extend test_user suite

Rezney:
ui_tests: extend test_user suite
    
    Extend WebUI test_user suite with the following test cases:
    
    test_add_user_special
    test_user_misc
    test_ssh_keys
    test_add_delete_undo_reset
    test_disable_delete_admin
    test_login_without_username
    
https://pagure.io/freeipa/issue/7507


varunmylaraiah:
@pvoborni could you please also take a look?

ipatests/test_webui/ui_driver.py
```diff
@@ -1700,32 +1749,53 @@ def get_t_vals(t):
...
-        self.find(ssh_pub, By.CSS_SELECTOR).click()
-        self.wait()
-        self.driver.switch_to.active_element.send_keys(ssh_key)
-        self.dialog_button_click('update')
-        self.facet_button_click('save')
+        if type(ssh_keys) is not list:
+            ssh_keys = [ssh_keys]
+
+        if navigate:
+            self.navigate_to_entity('user')
```
pvoborni: I know it was in the original method. But it would be better if instead of string a constant would be used. I'm also not sure about navigation in this method, but it was here before this patch so could be kept here.

ipatests/test_webui/ui_driver.py
```diff
@@ -1700,32 +1749,53 @@ def get_t_vals(t):
             # add multiple at once and test table delete button
             self.add_table_associations(table, keys, delete=True)
 
-    def add_sshkey_to_user(self, user, ssh_key):
+    def add_sshkey_to_user(self, ssh_keys, user='admin', navigate=True,
+                           save=True):
```
pvoborni: `navigate` and `save` could be documented as well given that others are.

ipatests/test_webui/data_user.py
```diff
@@ -71,6 +122,8 @@
         ('textbox', 'givenname', 'Name3'),
         ('textbox', 'sn', 'Surname3'),
         ('checkbox', 'noprivate', None),
+        ('password', 'userpassword', 'Supersecret123'),
+        ('password', 'userpassword2', 'Supersecret123'),
```
pvoborni: DATA3 are used in method `test_noprivate_nonposix` why this test needs to set a password?

ipatests/test_webui/data_user.py
```diff
@@ -85,3 +138,105 @@
         ('combobox', 'gidnumber', '77777'),
     ]
 }
+
+PKEY5 = '1spe.cial_us-er$'
+DATA5 = {
```
pvoborni: I'm beginning to think whether we should stop using general `DATAX` names for data which serve for a specific test. Originally there was only a need for more records without a special meaning to test the simplest CRUD or add more records as members.

Here if somebody changes the data in future, we could easily miss its purpose - a test case. So e.g. instead of DATA, it can be DATA_SPECIAL_CHARS. Same with other DATAs.

ipatests/test_webui/test_user.py
```diff
@@ -376,6 +409,249 @@ def fill_password_dialog(self, password, current=None):
...
+        assert 'Username: Required field' in alert_e.text, 'Alert expected'
+        assert self.login_screen_visible()
+
+    @screenshot
+    def test_disable_delete_admin(self):
+        """
+        Test disabling/deleting admin is not allowed
+        """
+        self.init_app()
+        self.navigate_to_entity('user')
```
pvoborni: Let's use the ENTITY constant. Same elsewhere.

ipatests/test_webui/test_user.py
```diff
@@ -376,6 +409,249 @@ def fill_password_dialog(self, password, current=None):
...
+        # try to disable admin user
+        self.select_record('admin')
+        self.facet_button_click('disable')
+        self.dialog_button_click('ok')
+        self.assert_last_error_dialog(ERR_ADMIN_DEL, details=True)
+        self.dialog_button_click('ok')
+        self.assert_record('admin')
+
+        # try to delete admin user. Later we are
+        # confirming by keyboard to test also upstream ticket 4097
```
pvoborni: There is no need to use `upstream` here as all ticket references should be upstream unless a full path is provided.

ipatests/test_webui/test_user.py
```diff
@@ -376,6 +409,249 @@ def fill_password_dialog(self, password, current=None):
...
+        self.fill_textbox('uid', 'itest-user@')
+        self.assert_field_validation(ERR_INCLUDE)
+        self.fill_textbox('uid', 'itest-user*')
+        self.assert_field_validation(ERR_INCLUDE)
+        self.fill_textbox('uid', 'itest-user?')
+        self.assert_field_validation(ERR_INCLUDE)
+        self.dialog_button_click('cancel')
+
+        # Add an user with special chars
+        self.init_app()
```
pvoborni: Why `init_app` here? And below?

Rezney: Notes before fix: Those are leftovers from squashing the particular test cases in one test_method.

ipatests/test_webui/ui_driver.py
Rezney: In https://github.com/freeipa/freeipa/pull/1679 you suggested to make navigation configurable however I really do not mind to change it.

ipatests/test_webui/test_user.py
```diff
@@ -588,6 +864,164 @@ def test_life_cycles(self):
...
+
+        self.logout()
+        self.init_app(user.PKEY5, '!!!@@@###$$$')
+
+        self.navigate_to_record(user.PKEY2, entity=user.ENTITY)
+
+        show_ssh_key_btn = self.find('div.widget .btn[name="ipasshpubkey-0"]',
+                                     By.CSS_SELECTOR)
+        show_ssh_key_btn.click()
+        ssh_key_dialog = self.find('.modal-dialog textarea', By.CSS_SELECTOR)
```
pvoborni: The variable name is not precise. It is a textarea element and not a dialog. A bit of a nitpic: Maybe it could be:
ssh_key_element =  `self.find('textarea', By.CSS_SELECTOR, self.get_dialog())`

The `get_dialog` call should ensure that it uses consistent way to get the dialog and also it is the last dialog opened.

ipatests/test_webui/test_user.py
```diff
@@ -588,6 +864,164 @@ def test_life_cycles(self):
...
+
+        assert ssh_key_dialog.get_attribute('readonly') == 'true'
+        self.dialog_button_click('cancel')
+        self.logout()
+        self.init_app()
+
+        # cleanup
+        self.delete_record([user.PKEY2, user.PKEY5])
```
pvoborni: most likely won't work. Use `delete` method.

ipatests/test_webui/ui_driver.py
pvoborni: So let's keep it here and make the method more general to work for host and id views as we talked offline.

ipatests/test_webui/test_user.py
```diff
@@ -376,6 +409,249 @@ def fill_password_dialog(self, password, current=None):
...
+        self.assert_last_error_dialog(USR_EXIST.format('nsurname10'))
+        self.close_all_dialogs()
+
+        # try to modify user`s UID to -1 (should fail)
+        self.navigate_to_record(user.PKEY)
+        self.mod_record(
+            user.ENTITY, {'mod': [('textbox', 'uidnumber', '-1')]},
+            negative=True)
+        uid_e = self.find('.widget[name="uidnumber"]', By.CSS_SELECTOR)
+        self.assert_field_validation('Minimum value is 1', parent=uid_e)
```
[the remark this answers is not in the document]
Rezney: Lets leave this for another PR as discussed offline.

Rezney:
@pvoborni your comments should be hopefully covered. Before pushing I have been experiencing some issues in SSH keys part which I have to tune a little tomorrow so some fails are expected but anyway should be good for review.

@varunmylaraiah thanks for catching the comment.

ipatests/test_webui/test_misc_cases.py
```diff
@@ -0,0 +1,41 @@
+#
+# Copyright (C) 2018  FreeIPA Contributors see COPYING for license
+#
+#
+# This program is free software; you can redistribute it and/or modify
```
pvoborni: For new files, lets use the newer shorter code file header which is e.g. in `ipaserver/plugins/server.py` I.e. only the:

```
#
# Copyright (C) 2018  FreeIPA Contributors see COPYING for license
#
```

Without the license related text below.

ipatests/test_webui/ui_driver.py
```diff
@@ -815,6 +820,46 @@ def add_multivalued(self, name, value, parent=None):
...
+        clicked = False
+        for i in inputs:
+            val = i.get_attribute('value')
+            n = i.get_attribute('name')
+            if val == value:
+                s = "input[name='%s'] ~ .input-group-btn button[name=undo]" % n
+                link = self.find(s, By.CSS_SELECTOR, w, strict=True)
+                link.click()
+                self.wait()
+                clicked = True
```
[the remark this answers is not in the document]
pvoborni: Will the change below work for all cases? Multivalued widget with 2 modified fields will have 2 undo buttons and 2 undo all button visible. If we click on one undo button then the other one together with the `undo all` will be still displayed and the assert, as it checks for all undo buttons in the widget, will fail.

ipatests/test_webui/ui_driver.py
```diff
@@ -1700,34 +1751,86 @@ def get_t_vals(t):
...
+        Delete all ssh public keys of particular record
+
+        pkey (str): user/host/idview to add the key to
+        navigate (bool): whether we should navigate to record
         """
-        self.navigate_to_entity('user')
-        self.navigate_to_record(user)
+
+        if navigate:
+            self.navigate_to_record(pkey)
```
pvoborni: I'm not sure about this navigation change, I thought/expected that the change will be:

```
def delete_record_sshkeys(self, entity, pkey, navigate=False):
    ....
    if navigate:
            self.navigate_to_entity(entity)
            self.navigate_to_record(pkey)
```

to be consistent with other navigation code, otherwise seems OK. Don't know if to block the review on that.

PR #1851, opened 2018-04-25
Create users in server-common pre hook

tiran:
The ipaapi user was created in the server package but referenced by a
config file in the server-common package. The server-common package can
be installed without the server package. This caused an error

   Unknown user 'ipaapi'

with systemd-tmpfiles --create. The users are now created in the
server-common package.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1155,6 +1155,8 @@ if [ -e /usr/sbin/ipa_kpasswd ]; then
 # END
 fi
 
+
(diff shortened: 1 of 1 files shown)


slaykovsky:
Hello @tiran 

Thank you for the patch, acked.

tiran:
master:

* 236fa61eebc62ece1e17add0d0e5b82a35afeadb Create users in server-common pre hook

PR #1855, opened 2018-04-25
ipa-advise config-server-for-smart-card-auth: use mod-ssl

flo-renaud:
ipa-advise config-server-for-smart-card-auth produces a script that
was still using /etc/httpd/conf.d/nss.conf instead of
/etc/httpd/conf.d/ssl.conf for setting the Apache SSLOCSPEnable Directive.

The fix replaces references to nss.conf with ssl.conf.

https://pagure.io/freeipa/issue/7515


tiran:
master:

* 2d5245cebcb498cf7c3c98f38784ed4d7d641c38 ipa-advise config-server-for-smart-card-auth: use mod-ssl

PR #1860, opened 2018-04-26
Use single Custodia instance in installers

tiran:


simo5:
It would have been better to avoi gratuitous reformats, but other than that I like the change.

simo5:
Once you fix the test failures introduced by this PR we can ack.

rcritten:
The reformatting is most likely due to pep8 issues.

tiran:
@simo5 CI has passed successfully.

simo5:
LGTM

PR #1861, opened 2018-04-26
Hardening of topology plugin to prevent erronous deletion of a replic…

tbordaz:
…a agreement

When a segment is deleted, the underlying replica agreement is also deleted.
An exception to this is if the status of the deleted segment is "obsolete" (i.e. merged segments)
The status should contain only one value, but to be protected against potential
bugs (like https://pagure.io/389-ds-base/issue/49619) this fix checks if
"obsolete" is in the status values.

https://pagure.io/freeipa/issue/7461

### DIFF
--- a/daemons/ipa-slapi-plugins/topology/topology_post.c
+++ b/daemons/ipa-slapi-plugins/topology/topology_post.c
@@ -240,22 +240,26 @@ ipa_topo_post_del(Slapi_PBlock *pb)
         /* check if corresponding agreement exists and delete */
         TopoReplica *tconf = ipa_topo_util_get_conf_for_segment(del_entry);
         TopoReplicaSegment *tsegm = NULL;
-        char *status;
+        int obsolete_segment;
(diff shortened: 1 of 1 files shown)


daemons/ipa-slapi-plugins/topology/topology_post.c
```diff
@@ -241,21 +241,26 @@ ipa_topo_post_del(Slapi_PBlock *pb)
         TopoReplica *tconf = ipa_topo_util_get_conf_for_segment(del_entry);
         TopoReplicaSegment *tsegm = NULL;
         char *status;
```
tiran: ``status`` is no longer used.

tbordaz:
Good catch, thanks for reviewing it. I updated patch

tiran:
master:

* fc371b651e5512cf06dd5de932632444f519376c Hardening of topology plugin to prevent erronous deletion of a replica agreement

PR #1864, opened 2018-04-26
[Backport][ipa-4-6] Use single Custodia instance in installers

tiran:
Manual backport of PR #1860 

Installers now pass a single CustodiaInstance object around, instead of
creating new instances on demand. In case of replica promotion with CA,
the instance gets all secrets from a master with CA present. Before, an
installer created multiple instances and may have requested CA key
material from a different machine than DM password hash.

In case of Domain Level 1 and replica promotion, the CustodiaInstance no
longer adds the keys to the local instance and waits for replication to
other replica. Instead the installer directly uploads the new public
keys to the remote 389-DS instance.

Without promotion, new Custodia public keys are still added to local
389-DS over LDAPI.

Fixes: https://pagure.io/freeipa/issue/7518
Reviewed-By: Simo Sorce <ssorce@redhat.com>

### DIFF
--- a/install/tools/ipa-ca-install
+++ b/install/tools/ipa-ca-install
@@ -33,6 +33,7 @@ from ipaserver.install.installutils import create_replica_config
 from ipaserver.install.installutils import check_creds, ReplicaConfig
 from ipaserver.install import dsinstance, ca
 from ipaserver.install import cainstance, service
+from ipaserver.install import custodiainstance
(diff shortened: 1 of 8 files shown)


tiran:
ipa-4-6:

* a0cdeb68e21a932c9554b8a305e7dcbfcc77cd94 Use single Custodia instance in installers

PR #1866, opened 2018-04-27
Allow more flexible services

abbra:
This patchset makes service objects more usable in dynamic environments like Kubernetes. Kubernetes hosts could create service objects that have no corresponding host object in IPA and need to be able to retrieve keytabs for them.

Such keytabs would be used for client authentication against other resources in the same IPA realm. As such, applications running in containers on a Kubernetes host would not need to accept any Kerberos authentication from their users but instead they would use own keytabs to talk to database or file servers (or anything else). Since in Kerberos for client authentication there is no real requirement that a service name has a corresponding host (it just a string with one or more / in it), we can allow to skip a host object check when creating the service.

As result, Web UI needs to be changed to also allow editing a host part of the service name when adding a service. We considered making a more dynamic handling of the UI but that would require a complete re-factor of the entity_select widget to allow dynamic change of the contained input widgets.

A second patch adds services as members of groups. This would allow to add services to groups and, in turn, use groups to grant ability for services to retrieve keytabs  of other services. This approach reduces substantially amount of changes needed to support services as first class objects.


install/ui/src/freeipa/service.js
```diff
@@ -415,6 +416,11 @@ return {
                         name: 'force',
                         metadata: '@mc-opt:service_add:force'
                     }
```
tiran: Missing comma

abbra: fixed

abbra:
Fixed group tracker in the tests to make sure all 'unexpected' attributes handled properly

abbra:
Hopefully, fixed all tests now.

abbra:
Should be final: added actual switch to an RPC call service-add that Web UI would call when creating a service.

rcritten:
Is it intended to allow services to be members of POSIX groups?

abbra:
Yes. There is no harm in that. SSSD will not see them because they aren't getting under an (objectclass=posixAccount) filter. Deliberately limiting group types is not needed.

rcritten:
I realize that but I have no doubt that some user is going to complain that services are not visible in groups.

rcritten:
master:

* 9e8fb94e87339b9908ec05fe5274ca51df3a82cf service: allow creating services without a host to manage them
* e64286571739d27822b118c780fbb8825038ae1c group: allow services as members of groups

PR #1867, opened 2018-04-27
Load certificate files as binary data

tiran:
In Python 3, cryptography requires certificate data to be binary. Even
PEM encoded files are treated as binary content.

certmap-match and cert-find were loading certificates as text files. A
new BinaryFile type loads files as binary content.

Fixes: https://pagure.io/freeipa/issue/7520


tiran:
@flo-renaud I added additional code to handle ``IPACertificate`` and ``Certificate`` instances for the SSSD interface.

Not sure how this code was ever able to run. It must have been broken for a while.

ipaserver/plugins/certmap.py
```diff
@@ -432,18 +434,23 @@ def list_users_by_cert(self, cert):
         Call Users.ListByCertificate interface and return a dict
         with key = domain, value = list of uids
         corresponding to the users matching the provided cert
-        :param cert: DER cert
+
+        :param cert: DER cert, IPACertificate, Certificate
         :raise RemoteRetrieveError: if DBus error occurs
         """
-        try:
+        if isinstance(cert, (x509.IPACertificate, crypto_x509.Certificate)):
```
stlaz: You can only leave `crypto_x509.Certificate` here as `IPACertificate` objects are also its instance.

tiran: aah, I missed ``@crypto_utils.register_interface(crypto_x509.Certificate)``

stlaz:
The reported errors are gone. ACK.

tiran:
master:

* c925b44f4348baf674fa8637be21ba0addfe2487 Load certificate files as binary data

PR #1870, opened 2018-04-27
idoverrideuser-add: allow adding ssh key in web ui

abbra:
CLI already allows to pass public SSH key when creating an ID override
for a user. Web UI allows to add public SSH keys after the ID override
was created.

Add SSH key field to allow passing public SSH key in one go when
creating an ID override for a user.

Fixes: https://pagure.io/freeipa/issue/7519

### DIFF
--- a/install/ui/src/freeipa/idviews.js
+++ b/install/ui/src/freeipa/idviews.js
@@ -317,6 +317,10 @@ return {
                 $type: 'cert_textarea',
                 name: 'usercertificate'
             },
+            {
+                $type: 'sshkey',
+                name: 'ipasshpubkey'
+            },
             'loginshell',
             'homedirectory',
             {


install/ui/src/freeipa/idviews.js
```diff
@@ -317,6 +317,10 @@ return {
                 $type: 'cert_textarea',
                 name: 'usercertificate'
             },
+            {
+                $type: 'sshkey',
+                name: 'ipasshpubkey'
```
tiran: Should the field have ``label: '@i18n:objects.sshkeystore.keys'``, too?

abbra: I think it is not necessary because sshkey widget has 
```
text: text.get('@i18n:objects.sshkeystore.show_set_key')
```

tiran:
master:

* 2d6d768d1aaa958ae77b5941a4f268e12894c226 idoverrideuser-add: allow adding ssh key in web ui

PR #1871, opened 2018-04-27
Run PR-CI with Fedora 28

tiran:

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -13,7 +13,7 @@ topologies:
     memory: 6700
 
 jobs:
-  fedora-27/build:
+  fedora-28/build:
     requires: []
     priority: 100
     job:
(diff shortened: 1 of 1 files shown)


felipevolpone:
In PR #1887 it's running the nightly tests in f28

felipevolpone:
We should not change PR CI to test on f28 by default. As PR 1887 shows, we still have a lot of broken tests.

stlaz:
Is there anything else we need before pushing this?

felipevolpone:
PR #1887 is still kinda broken, so we need to get that fixed before merging this.

abbra:
I think python installer tools in 389-ds can be done separately from this PR. We should really move to testing on F28 for the rest.

stlaz:
Not sure if I understand correctly, python 389-ds installer is done in another PR.
I also don't at all understand why we should wait for https://github.com/freeipa/freeipa/pull/1887 to be green, if the basic suite that's run in every PR is green, we can fix the rest as we go and don't block other efforts in the meantime.

felipevolpone:
I'm not nacking this PR so feel free to merge it if you think it should be.
My point in my previous comment was that we still have failing tests on PR 1887 (in f28) and IMHO we should first fix those tests. However, it seems that we have more people that thinks the opposite, so let's ack this.

stlaz:
Thanks, Felipe. I really do believe that we should push it, let me do that, then :)

stlaz:
Pushed as 75e86f2f13cc2abef457593312cdd1b84d99733a

PR #1872, opened 2018-04-27
Log service start/stop/restart message

rcritten:
It wasn't apparent in the logs if a service stop or restart
was complete so in the case of a hang it wasn't obvious which
service was responsible. Including start here for completeness.

https://pagure.io/freeipa/issue/7436


tiran:
master:

* 6856a9f46cf7418b4188dbc27bc8c20841739542 Log service start/stop/restart message

PR #1884, opened 2018-04-30
Add absolute_import to test_authselect

stlaz:
This is to keep backward compatibility with Python 2


tiran:
master:

* d5e5bd501c8ed8df72562a9acdef922eaa00206e Add absolute_import to test_authselect

PR #1886, opened 2018-04-30
group-del: add a warning to logs when password policy could not be re…

abbra:
…moved

When a user with sufficient permissions creates a group using ipa
group-add and then deletes it again with group-del ipa gives an
Insufficient access error, but still deletes the group.

This is due to a need to remove an associaed password policy for the
group. However, a password policy might be inaccessible to the user
(created by a more powerful admin) and there is no way to check that it
exists with current privileges other than trying to remove it.

Seeing a Python exceptions in the Apache log without explanation is
confusing to many users, so add a warning message that explains what
happens here.

Fixes: https://pagure.io/freeipa/issue/6884

### DIFF
--- a/ipaserver/plugins/group.py
+++ b/ipaserver/plugins/group.py
@@ -20,6 +20,8 @@
 
 import six
 
+import logging
+
 from ipalib import api
 from ipalib import Int, Str, Flag
 from ipalib.constants import PATTERN_GROUPUSER_NAME
@@ -47,6 +49,8 @@ from ipapython.dn import DN
 
 if six.PY3:
     unicode = str
+
+logger = logging.getLogger(__name__)
 
 if api.env.in_server and api.env.context in ['lite', 'server']:
     try:
@@ -366,7 +370,16 @@ class group_del(LDAPDelete):
     def post_callback(self, ldap, dn, *keys, **options):
         assert isinstance(dn, DN)
         try:
+            # A user removing a group may have no rights to remove
+            # an associated policy. Make sure we log an explanation
+            # in the Apache logs for this.
             api.Command['pwpolicy_del'](keys[-1])
+        except errors.ACIError:
+            logger.warning(
+                "While removing group %s, user lacked permissions "
+                "to remove corresponding password policy. This is "
+                "not an issue and can be ignored.", keys[-1]
+            )
         except errors.NotFound:
             pass
 


rcritten:
LGTM. This is best solved once we get a notification system so we can notify an admin of an orphaned password policy but this is at least a step forward in both usability and notification.

pvoborni:
@rcritten we have a notification system, check `ipalib/messages.py`. Removed ACK label to think about this option.

Example of usage:
```python
# in warning.py:

class BrokenTrust(PublicMessage):
    """
    **13018** Trust for a specified domain is broken
    """

    errno = 13018
    type = "warning"
    format = _("Your trust to %(domain)s is broken. Please re-create it by "
               "running 'ipa trust-add' again.")


# elsewhere
   from ipalib.messages import (add_message,  BrokenTrust)


   add_message(
        options['version'],
        result,
       BrokenTrust(domain=entry.single_value['cn'])
   )
```

It is then displayed also in Web UI in a similar manner as e.g.  "User added" notification.

rcritten:
@pvoborni sorry, I mean some sort of administrative notification system.

There has been a long-standing request to be able to be notified of things like new users so directories can be created, etc. I've always pictured it as a sort of AMQP queue.

For this particular patch the user doing the delete may have no idea whether there is a password policy or not so alerting them to it wouldn't be much of an improvement over the insufficient access error IMHO. What would it say, "There might be an orphaned password policy due to this delete, alert your administrator?"

Logging is a good first step and an administrator could watch the logs for things like this and verify that any associate password policy has been deleted.

rcritten:
Are we all good to ack and push this?

ipaserver/plugins/group.py
```diff
@@ -366,9 +370,16 @@ def pre_callback(self, ldap, dn, *keys, **options):
     def post_callback(self, ldap, dn, *keys, **options):
         assert isinstance(dn, DN)
         try:
+            # A user removing a group may have no rights to remove
+            # an associated policy. Make sure we log an explanation
+            # in the Apache logs for this.
             api.Command['pwpolicy_del'](keys[-1])
-        except errors.NotFound:
-            pass
+        except (errors.NotFound, errors.ACIError) as e:
```
tiran: I'd prefer if you replace the isinstance check with two separate except blocks:

```
        try:
            ...
        except errors.ACIError as e:
            logger.warning(
                "While removing group %s, user lacked permissions "
                "to remove corresponding password policy. This is "
                "not an issue and can be ignored.", keys[-1]
            )
        except errors.NotFound:
            pass
```

(I also added a full stop to the end of the second sentence.)

abbra:
Updated the patch according to @tiran  suggestions.

tiran:
LGTM, let's wait for CI

tiran:
@rcritten @pvoborni Is there anything left to do here?

tiran:
master:

* 1adc941d1f1caeffa8cf490783b7819298e828ce group-del: add a warning to logs when password policy could not be removed

PR #1893, opened 2018-05-01
Disable message about log in ipa-backup if IPA is not configured

rcritten:
Introduce server installation constants similar to the client
but only tie in SERVER_NOT_CONFIGURED right now.

For the case of not configured don't spit out the "See <some log>
for more information" because no logging was actually done.

In the case of ipa-backup this could also be confusing if the
--log-file option was also passed in because it would not be
used.

https://pagure.io/freeipa/issue/6843

### DIFF
--- a/ipapython/admintool.py
+++ b/ipapython/admintool.py
@@ -32,6 +32,10 @@ from ipapython import version
 from ipapython import config
 from ipapython.ipa_log_manager import standard_logging_setup
 
+SUCCESS = 0
+SERVER_INSTALL_ERROR = 1
+SERVER_NOT_CONFIGURED = 2
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


ipaserver/install/installutils.py
```diff
@@ -76,6 +76,10 @@
 IPA_MODULES = [
     'httpd', 'kadmin', 'dirsrv', 'pki-tomcatd', 'install', 'krb5kdc', 'named']
 
+SUCCESS = 0
+SERVER_INSTALL_ERROR = 1
+SERVER_NOT_CONFIGURED = 2
```
tiran: Could you move the constants into ``ipapython.admintool``, please?

ipapython/admintool.py
```diff
@@ -301,7 +301,9 @@ def log_failure(self, error_message, return_value, exception, backtrace):
         if error_message:
             logger.error('%s', error_message)
         message = "The %s command failed." % self.command_name
-        if self.log_file_name:
+        if self.log_file_name and return_value != 2:
```
tiran: ``s/2/SERVER_NOT_CONFIGURED/``

tiran:
master:

* 648d7c0d38ef99196531e94cd9803c630e7fb05c Disable message about log in ipa-backup if IPA is not configured

PR #1894, opened 2018-05-02
install: fix reported external CA configuration

frasertweedale:
The installer reports the CA configuration that will be used,
including whether the CA is self-signed or externally-signed.

Installation with external CA takes two steps. The first step
correctly reports the externally signed configuration (like the
above), but the second step reports a self-signed configuration.

The CA *is* externally signed, but the configuration gets reported
incorrectly at step 2.  This could confuse the administrator.  Fix
the message.

Fixes: https://pagure.io/freeipa/issue/7523


tiran:
Thanks!

tiran:
master:

* 6659392a0a5064b25f8d0b5a2fe805a172b7e586 install: fix reported external CA configuration

PR #1899, opened 2018-05-02
Clean up Uninstall output

rcritten:
ipa-server-install did not configure a console logger. This caused odd output because ipa-client-install does and a normal server uninstall would end with "ipa-client-install command was successful"

Because there was no console logger there were print statements in a lot of places. Those were dropped because the equivalent logging statement was already there.

A number of info loggers were converted to debug to suppress their output. They were never shown in the past for the most part with the exception of the client installer and Forwarding and trying statements.

Adding the logger exposed two additional issues:

1. When deleting the last server the server-del command threw an LDAP error trying to write an empty attribute
2. Two ACI parsing issues related to subtypes

### DIFF
--- a/ipalib/aci.py
+++ b/ipalib/aci.py
@@ -25,7 +25,8 @@ import six
 # The Python re module doesn't do nested parenthesis
 
 # Break the ACI into 3 pieces: target, name, permissions/bind_rules
-ACIPat = re.compile(r'\(version\s+3.0\s*;\s*ac[li]\s+\"([^\"]*)\"\s*;\s*([^;]*);\s*\)', re.UNICODE)
+ACIPat = re.compile(r'\(version\s+3.0\s*;\s*ac[li]\s+\"([^\"]*)\"\s*;'
+                    r'\s*(.*);\s*\)', re.UNICODE)
 
 # Break the permissions/bind_rules out
 PermPat = re.compile(r'(\w+)\s*\(([^()]*)\)\s*(.*)', re.UNICODE)
--- a/ipapython/admintool.py
+++ b/ipapython/admintool.py
@@ -93,6 +93,7 @@ class AdminTool(object):
     log_file_name = None
     usage = None
     description = None
+    ignore_return_codes = ()
 
     _option_parsers = dict()
 
@@ -177,9 +178,13 @@ class AdminTool(object):
             self.setup_logging()
             return_value = self.run()
         except BaseException as exception:
+            if isinstance(exception, ScriptError):
+                # pylint: disable=no-member
+                if exception.rval and exception.rval > return_value:
+                    return_value = exception.rval  # pylint: disable=no-member
             traceback = sys.exc_info()[2]
             error_message, return_value = self.handle_error(exception)
-            if return_value:
+            if return_value and return_value not in self.ignore_return_codes:
                 self.log_failure(error_message, return_value, exception,
                     traceback)
                 return return_value
--- a/ipapython/install/cli.py
+++ b/ipapython/install/cli.py
@@ -58,7 +58,24 @@ def _get_usage(configurable_class):
 
 def install_tool(configurable_class, command_name, log_file_name,
                  debug_option=False, verbose=False, console_format=None,
-                 use_private_ccache=True, uninstall_log_file_name=None):
+                 use_private_ccache=True, uninstall_log_file_name=None,
+                 ignore_return_codes=()):
+    """
+    Some commands represent multiple related tools, e.g.
+    ``ipa-server-install`` and ``ipa-server-install --uninstall`` would be
+    represented by separate classes. Only their options are the same.
+
+    :param configurable_class: the command class for options
+    :param command_name: the command name shown in logs/output
+    :param log_file_name: if None, logging is to stderr only
+    :param debug_option: log level is DEBUG
+    :param verbose: log level is INFO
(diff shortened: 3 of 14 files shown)


ipaserver/plugins/server.py
```diff
@@ -619,7 +619,10 @@ def _remove_server_principal_references(self, master):
             if master in srvlist:
                 srvlist.remove(master)
                 attr = ' '.join(srvlist)
-                ret['defaultServerList'] = attr
+                if len(srvlist) == 0:
```
stlaz:
```python
if not srvlist:
    del ret['defaultServerList']
else:
    ret['defaultServerList'] = ' '.join(srvlist)
```
^-- removes the possibly unnecessary `join()` if the list is empty and the condition is a bit more Pythonic I believe :)

rcritten: +1, done

ipapython/install/cli.py
```diff
@@ -58,7 +58,24 @@ def _get_usage(configurable_class):
 
 def install_tool(configurable_class, command_name, log_file_name,
                  debug_option=False, verbose=False, console_format=None,
-                 use_private_ccache=True, uninstall_log_file_name=None):
+                 use_private_ccache=True, uninstall_log_file_name=None,
+                 ignore_return_codes=[]):
```
tiran: Please don't use mutable data types in argument lists. If you need an iterable, use a tuple as default argument ``ignore_return_codes=()``.

ipalib/aci.py
```diff
@@ -25,7 +25,8 @@
 # The Python re module doesn't do nested parenthesis
 
 # Break the ACI into 3 pieces: target, name, permissions/bind_rules
-ACIPat = re.compile(r'\(version\s+3.0\s*;\s*ac[li]\s+\"([^\"]*)\"\s*;\s*([^;]*);\s*\)', re.UNICODE)
+ACIPat = re.compile(r'\(version\s+3.0\s*;\s*ac[li]\s+\"([^\"]*)\"\s*;'
```
tiran: This change looks unrelated to the rest.

ipapython/admintool.py
```diff
@@ -93,6 +93,7 @@ class AdminTool(object):
     log_file_name = None
     usage = None
     description = None
+    ignore_return_codes = []
```
tiran: Same here, use a tuple ``()``.

ipatests/test_ipalib/test_aci.py
```diff
@@ -162,3 +162,15 @@ def test_aci_parsing_8():
 def test_aci_parsing_9():
     check_aci_parsing('(targetfilter = "(|(objectClass=person)(objectClass=krbPrincipalAux)(objectClass=posixAccount)(objectClass=groupOfNames)(objectClass=posixGroup))")(targetattr != "aci || userPassword || krbPrincipalKey || sambaLMPassword || sambaNTPassword || passwordHistory")(version 3.0; acl "Account Admins can manage Users and Groups"; allow (add, delete, read, write) groupdn = "ldap:///cn=admins,cn=groups,cn=accounts,dc=greyoak,dc=com";)',
         '(targetattr != "aci || userPassword || krbPrincipalKey || sambaLMPassword || sambaNTPassword || passwordHistory")(targetfilter = "(|(objectClass=person)(objectClass=krbPrincipalAux)(objectClass=posixAccount)(objectClass=groupOfNames)(objectClass=posixGroup))")(version 3.0;acl "Account Admins can manage Users and Groups";allow (add,delete,read,write) groupdn = "ldap:///cn=admins,cn=groups,cn=accounts,dc=greyoak,dc=com";)')
+
+
+def test_aci_parsing_10():
```
tiran: This also looks unrelated.

ipaserver/install/cainstance.py
```diff
@@ -1794,7 +1794,7 @@ def _create_dogtag_profile(profile_id, profile_data, overwrite):
         # import the profile
         try:
             profile_api.create_profile(profile_data)
-            logger.info("Profile '%s' successfully migrated to LDAP",
+            logger.debug("Profile '%s' successfully migrated to LDAP",
```
tiran: The second like with ``profile_id`` needs to be reindented, too.

tiran:
Thanks for cleanup up the logger! :+1:

ipalib/aci.py
rcritten: This and the change to test_aci.py are related because by enabling logging two ACI parsing errors were exposed by the LDAP updater.

ACI parsing is forgiving and errors are not treated as fatal so this didn't cause any operational issues but two rather nasty looking errors would appear while applying updates as a direct result of my output changes so I fixed this as well.

tiran:
```
************* Module ipaserver.install.custodiainstance
ipaserver/install/custodiainstance.py:87: [W0612(unused-variable), get_custodia_instance] Unused variable 'ldap_uri')
```

Tiboris:
master:

* b96906156be37a7b29ee74423b82f04070c84e22 Improve console logging for ipa-server-install
* 8ea227451f4d85dbca6a331a607d25744e85121b Drop attr defaultServerList if removing the last server
* 00ddb5dd53a2eb21cd7c97167b8a5f87ab728b07 server install: drop some print statements, change log level
* 036d51d5143a142e6a3070e6328a7bcd9b2125f0 Handle subyptes in ACIs

PR #1902, opened 2018-05-03
ipa-client package needs sssd-tool

tiran:
Commit ccec8c6c4193a204428b7ba0f93dac6f0eb26020 add a call to sssctl but
the providing package sssd-tools was not added to ipa-client package.
The tool is not need to build packages.

See: https://pagure.io/freeipa/issue/7376


abbra:
FYI, a reason why sssctl support warning was added in the original commit is to allow us working on the systems where sssd-tools is not installed.

tiran:
master:

* 1c1089c44df34c04d5ba4572191a49962e859a92 ipa-client package needs sssd-tool

PR #1904, opened 2018-05-03
[temp PR] f28 Patches and all nightly tests

felipevolpone:
Applying patches from PR #1871  and nightly tests


pvoborni:
Nightly tests already run with this setup. E.g. #1968 So closing.

PR #1906, opened 2018-05-04
mod_ssl: add SSLVerifyDepth for external CA installs

stlaz:
mod_ssl's limiting of client cert verification depth was causing
the replica installs to fail when master had been installed with
external CA since the SSLCACertificateFile was pointing to a file
with more than one certificate. This is caused by the default
SSLVerifyDepth value of 1. We set it to 5 as that should be
just about enough even for possible sub-CAs.

https://pagure.io/freeipa/issue/7530

### DIFF
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -319,3 +319,6 @@ USER_CACHE_PATH = (
 )
 
 SOFTHSM_DNSSEC_TOKEN_LABEL = u'ipaDNSSEC'
+# Apache's mod_ssl SSLVerifyDepth value (Maximum depth of CA
+# Certificates in Client Certificate verification)
+MOD_SSL_VERIFY_DEPTH = '5'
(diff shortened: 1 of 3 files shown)


ipalib/constants.py
```diff
@@ -319,3 +319,5 @@
 )
 
 SOFTHSM_DNSSEC_TOKEN_LABEL = u'ipaDNSSEC'
+# certificate verification depth of Apache's mod_ssl (SSLVerifyDepth)
```
tiran: Please mention that it affects client cert verification only. Perhaps just copy the short explanation from https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslverifydepth ?

``Maximum depth of CA Certificates in Client Certificate verification``

stlaz:
I also added a replica installation test to an existing simple external CA installation test.

rcritten:
Don't want to complicate things but...

Since we already have the chain why can't we calculate the depth?

tiran:
We don't know the verify depths in case a customer uses a nested lightweight sub CA.

stlaz:
master:

* 392f44a38a7e97242cfd2145592fbf6038191d09 mod_ssl: add SSLVerifyDepth for external CA installs

PR #1911, opened 2018-05-08
Validate the Directory Manager password in ipa-restore (take 2)

rcritten:
Validate the DM password before starting the actual restoration if /etc/ipa/default.conf exists AND dirsrv is running. If default.conf exists and dirsrv is not running raise an error.

Add two tests to validate this behavior.

https://pagure.io/freeipa/issue/7535


tiran:
master:

* 2256f9ef6af5c5eb780a0c5aeadb249ff882549f Validate the Directory Manager password before starting restore
* 59b3eb0433b85f554ba999487288ad18a2be33e0 Add tests for ipa-restore with DM password validation check

PR #1912, opened 2018-05-09
Allow user administrator to change user homedir

stlaz:
https://pagure.io/freeipa/issue/7427


stlaz:
master:

* a885f07d1388deb7c126e7f01b6c0db0085ccaf8 Allow user administrator to change user homedir

PR #1914, opened 2018-05-09
Fixing DNSSEC tests with restarting named

felipevolpone:
This commit fixes:
- TestInstallDNSSECLast::test_disable_reenable_signing_master
- TestInstallDNSSECLast::test_disable_reenable_signing_replica
- TestInstallDNSSECFirst::test_chain_of_trust

https://pagure.io/freeipa/issue/5670

### DIFF
--- a/ipatests/test_integration/test_dnssec.py
+++ b/ipatests/test_integration/test_dnssec.py
@@ -5,6 +5,7 @@
 from __future__ import absolute_import
 
 import logging
+import pytest
 
 import dns.dnssec
 import dns.resolver
@@ -144,6 +145,7 @@ class TestInstallDNSSECLast(IntegrationTest):
             self.master.ip, test_zone_repl, timeout=5
         ), "DNS zone %s is not signed (master)" % test_zone
 
+    @pytest.mark.xfail(reason='Ticket N 5670')
     def test_disable_reenable_signing_master(self):
 
         dnskey_old = resolve_with_dnssec(self.master.ip, test_zone,
@@ -191,6 +193,7 @@ class TestInstallDNSSECLast(IntegrationTest):
                                          rtype="DNSKEY").rrset
         assert dnskey_old != dnskey_new, "DNSKEY should be different"
 
+    @pytest.mark.xfail(reason='Ticket N 5670')
     def test_disable_reenable_signing_replica(self):
 
         dnskey_old = resolve_with_dnssec(self.replicas[0].ip, test_zone_repl,
@@ -307,6 +310,7 @@ class TestInstallDNSSECFirst(IntegrationTest):
             self.replicas[0].ip, root_zone, timeout=300
         ), "Zone %s is not signed (replica)" % root_zone
 
+    @pytest.mark.xfail(reason='Ticket N 5670')
     def test_chain_of_trust(self):
         """
         Validate signed DNS records, using our own signed root zone


tiran:
Thanks @felipevolpone 

Are restarts of named always necessary for some settings? We could add an oddjobd task to perform restarts automatically. Please open a bug.

felipevolpone:
> Are restarts of named always necessary for some settings? 

I'm not able to answer that, I have tried to figure it out, but I need some help.

Instead of restarting services on test, I tried to add it to the `dnszone_mod` command and check if the test results are the same.

I have added:
```diff
diff --git a/ipaserver/plugins/dns.py b/ipaserver/plugins/dns.py
index 016be0c6f..76ecb1606 100644
--- a/ipaserver/plugins/dns.py
+++ b/ipaserver/plugins/dns.py
@@ -2904,6 +2904,10 @@ class dnszone_mod(DNSZoneBase_mod):
         dn = super(dnszone_mod, self).post_callback(ldap, dn, entry_attrs,
                                                     *keys, **options)
         self.obj._rr_zone_postprocess(entry_attrs, **options)
+        named_regular = services.service('named-regular', api)
+        named_regular.restart()
         return dn
```
and put the tests to run. It fails with:
```
  File "/usr/lib/python3.6/site-packages/ipaplatform/base/services.py", line 319, in _restart_base
    skip_output=not capture_output)
  File "/usr/lib/python3.6/site-packages/ipapython/ipautil.py", line 572, in run
    p.returncode, arg_string, output_log, error_log
ipapython.ipautil.CalledProcessError: CalledProcessError(Command ['/bin/systemctl', 'restart', 'named-pkcs11.service'] returned non-zero exit status 1: 
"Failed to restart named-pkcs11.service: Interactive authentication required.
\\nSee system logs and 'systemctl status named-pkcs11.service' for details.\\n")
```
Full log: https://fedorapeople.org/groups/freeipa/prci/jobs/c29cfb2e-53d9-11e8-b324-fa163e8ca376/-TestInstallDNSSECLast--test_disable_reenable_signing_master/master.ipa.test/var/log/httpd/error_log.gz

I also have tried to reproduce this error running the commands manually on the same setup, but this problem doesn't happen. So, any help is welcome :)

felipevolpone:
@abbra @tiran  PR #1559 has the changes I did, but also includes more changes (decreasing ttl time, etc). Should we get this merged or go with 1559? Relevant to say that that patch has 5 restarts and this only 3.

An alternative would be: add `xfail` [1] to dnssec tests and merge PR 1559, given that this PR is a workaround and not exactly a fix. With `xfail` we could track when the tests would start to be green again. 

[1] http://pytest.readthedocs.io/en/reorganize-docs/new-docs/user/xfail.html

pvoborni:
I'd say we don't need to fix DNSSec to full support as it is not our immediate goal, but we need to put tests into a correct state so that it won't hide other issues and that we are able to detect if the situation which causes current failures changes. This patch is missing the detection. Original patch in https://pagure.io/freeipa/c/5567dff4b46cc05bf0ea44dd03afdd12645143a5 had something like it.

felipevolpone:
So what about we remove the `restart_named` and go only with `xfail`?

stlaz:
Yes, I won't ACK this as long as there are unexplained restarts.

felipevolpone:
I added `xfail` to the tests. Once it get acked I'll remove the temp commit

stlaz:
We'll go with the `xfail` way. I removed the temp commit so that this can be pushed.

flo-renaud:
Adding re-run label because of `fedora-28/external_ca — Process "vagrant up --no-provision --parallel" exited with error code 1`

stlaz:
master:

* f03df5fe41ed43e63ae7b7b63929140110bc85e0 Adding xfail to failing tests

stlaz:
Needs manual rebase.

PR #1915, opened 2018-05-09
Revoke existing certificate when a new one is issued by IPA CA

rcritten:


frasertweedale:
As I alluded in the ticket (https://pagure.io/freeipa/issue/7482) I think we should have a broader discussion about what is the correct revocation behaviour in `cert-request`.

There are four basic scenarios where `cert-request` will be revoked:

1. requesting a certificate for some new purpose.  the subject principal may already have certs for other purposes.  existing certificates should not be revoked.
2. renew an existing certificate, because the end of the validity period is near.  therefore there is little benefit to revoking.  furthermore, if there are multiple existing certificates, to decide which one is being replaced is non-trivial in ideal circumstances, and undecidable in general.
3. rekey or reissue an existing certificate, for any reason, when the expiry is not near.  the existing cert *should be revoked*.  But again, it is not simple to determine which certificate is being replaced.

Proposal:

- remove revocation behaviour from `cert-request`
  - revocation behaviour could be retained, but only activated if a speciifc argument is given e.g. `cert-request CSR --principal PRINC --revoke ISSUER:SERIAL` would revoke the certificate with `ISSUER:SERIAL` if (and only if) it appears in the subject principal's entry, and remove it from that entry.

- enhance the IPA certmonger helper to supply the appropriate options or invoke `ipa cert-revoke` separately, to revoke the certificate being renewed.  IIIRC the old certificate is provided in the certmonger helper's environment so this should be straightforward.

rcritten:
The revocation behavior is already gone as a regression. This is simply restoring the previous status-quo.

I'd prefer revoke to be opt-out if we add it at all but I think any wholesale changes about revocation will need a separate design discussion.

For services in particular the usage was designed to be very one-to-one: one IPA service per one Unix service so one cert per service.

Which means getting a new one means the old one is not needed, hence revoke it as superceded.

Why is it not needed? Who knows. Does it matter? Probably not.

You are right about determining which cert should be revoked is basically impossible if they are all issued by the IPA CA. But it doesn't matter because there should be a one-to-one mapping for IPA-issued certs. To flesh this scenario out I had to create a separate service, issue the cert, then use service-add-cert to stuff it into a different IPA service. So now I had 2 certs with the same subject and different serial numbers. Why in the world would someone legitimately do this?

And yes, both were revoked when I forced a renewal.

If for some reason a cert-add was done with an external cert then revocation won't even be attempted.

So for hosts and services I can see no need at all for having multiple certs therefore revocation is the right thing.

For users that is a different matter completely and is outside the scope of this PR.

ipaserver/plugins/cert.py
```diff
@@ -898,7 +898,7 @@ def execute(self, csr, all=False, raw=False, chain=False, **kw):
         store = profile['result']['ipacertprofilestoreissued'][0] == 'TRUE'
         if store and 'certificate' in result:
             cert = result.get('certificate')
-            kwargs = dict(addattr=u'usercertificate={}'.format(cert))
+            kwargs = dict(usercertificate=cert)
```
frasertweedale: I think that we should use `usercertificate=cert` only when `principal_type in [SERVICE, HOST]`,
and continue using `addattr` for user principals.  Replacing all existing `usercertificate` values doesn't seem right for user principals (even if `user_mod` does not revoke the removed certificates).

pvoborni:
The blog post pushed me in checking what were the reasons for changing IPA behavior. It led me to https://www.freeipa.org/page/V4/User_Certificates there is written:


> ## Changes to Certificate Bookkeeping

> This design changes the default expectation of current cert management API for hosts and services as it always expected userCertificate to be filled and used for certificate manipulation and bookkeeping (listing, revoking). Instead, some certificates may not be stored in the user entry at all and only stored in Dogtag (especially the short lived certificates).

> ## Revocation of the Certificates

> All older version of FreeIPA always revoked the certificate when an object (host, service) was deleted or disabled. With every revocation, the certificate has to be added to the revocation list and properly distributed in the CRL object. Big CRL revocation lists may cause issues with replication (#4048) or CRL processing.

> With this feature, objects are likely to have multiple certificates and the revocation list growth would increase even more. This design therefore plans to stop revoking certificates automatically as in most cases (deleting objects/certificates when not useful) the certificate does not have to be added to the revocation list and can be simply deleted and left for natural expiration. However, there are still cases when revocation is due (the key was compromised, user leaves organization and retained a copy of the private key) and FreeIPA needs to have ability to revoke these certificates. 

In other words, it was a consious decision and we were to hasty in calling it a regression. 

In practice, deleting (and thus also revoking) all host and service certs would hinder the multiple certificate support implemented Web UI and commands:
```
{host,service,user}-add-cert
{host,service,user}-host-remove-cert
```
As any newly requested cert would remove everything else added.

ipaserver/plugins/cert.py
rcritten: If the certificate isn't revoked then there is no reason to remove the others. I think a broader discussion is needed on freeipa-devel because this is going to lead to other side-effects, like certs never disappearing from usercertificate on expiration.

frasertweedale:
Let us decide what to do with this PR.  @pvoborni's archaeology has shown that *not revoking* is the
intended behaviour.  I have an AI to write up a design to formalise and justify (non-)revocation behaviours in FreeIPA.  (It will mostly be a formalisation of the [blog post](https://frasertweedale.github.io/blog-redhat/posts/2018-05-11-renewal-and-revocation.html).)

So, my vote is to close this PR and https://pagure.io/freeipa/issue/7482.

frasertweedale:
(ping @tiran since you filed https://pagure.io/freeipa/issue/7482)

rcritten:
Something needs to be opened because not revoking the certificate is a side-effect only. If we don't want to revoke the cert then we shouldn't even try, and ideally a code comment should be added.

I'm ok closing this PR and/or the ticket but something needs to track this work.

frasertweedale:
@rcritten OK, thanks.  As soon I have that design written up and published, and created a corresponding ticket, I'll reference them here, and in pagure ticket 7482, and close them.

frasertweedale:
Design: https://www.freeipa.org/page/V4/Certificate_revocation_behaviour_standardisation

Ticket: https://pagure.io/freeipa/issue/7580

PR #1919, opened 2018-05-10
ui_tests: checkbox click fix

Rezney:
We check a box with clicking on label by default however sometimes when a label is too short 
(1-2 letters) we are hitting an issue that the checkbox obscures the label.


Rezney:
@pvoborni added a ticket and removed the temp_commit for this ticket too...

Rezney:
master:

* ef3f0851f43e733c489a704ca973ba133c6532d4 ui_tests: checkbox click fix

PR #1920, opened 2018-05-11
ui_tests: improve "field_validation" method

Rezney:
Often when trying to check e.g. required field we pass the method another element as parent in order to narrow down a scope for validation. This way we can just pass "field" name to make the process easier.


varunmylaraiah:
This patch works for me

pvoborni:
The commit msg doesn't indicate to which branches it should be pushed. I assume that it is a prerequisite for some work which will go to 4.6. 

Otherwise looks good to me, providied that it works - I did not test.

Rezney:
@pvoborni added a ticket and removed the temp_commit...

Rezney:
master:

* 897f1cda930d33d32137de3d9a9742336be6531e ui_tests: improve "field_validation" method

PR #1922, opened 2018-05-14
Radius proxy multiservers fix

slaykovsky:
Now radius proxy plugin allows to add more then one radius server
into radius proxy but the first one from ldap response is being
parsed (you can see ./daemons/ipa-optd/parse.c).

So this kind of behaviour is a bug, as it was determined on IRC.

This patch removes possibility to add more then one radius server
into radius proxy.

Pagure: https://pagure.io/freeipa/issue/7542


abbra:
Thanks. Since you are changing API, you need to make sure API.txt is updated as well. Please run `make fastcheck` before the push -- it would tell you to run 'make makeapi`. ;)

slaykovsky:
@abbra oh, thanks, didn't know that :)

abbra:
LGTM. Let's see how that goes with tests.

rcritten:
master:

* b82af698828354fcfb14200ab0dd5a397185ad59 Radius proxy multiservers fix

PR #1924, opened 2018-05-14
tests: Fix on test_sudo_rule_restricted_to_one_hostmask_negative

celestian:
The error message was changed. This patch fix the test
to current situation.

https://pagure.io/freeipa/issue/7543

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -23,7 +23,7 @@ jobs:
         git_refspec: '{git_refspec}'
         template: &ci-master-f27
           name: freeipa/ci-master-f27
-          version: 1.0.3
+          version: 1.0.4
         timeout: 1800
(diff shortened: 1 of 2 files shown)


gkaihorodova:
Fix LGTM

celestian:
Wau, it seems that it wasn't bug at all.

celestian:
OK, we will try test it a little later. It seems that PR CI needs update. :-)

celestian:
Interesting fact!
With older PR CI image, this PR looks like bug. But with newer, this is correct PR which really fix something, respectively sudo error message in our test.

celestian:
This issue could be connected to:
https://pagure.io/SSSD/sssd/issue/3742

celestian:
This approach how to solve the issue doesn't seem to be valid any more.  So I am closing this PR.

PR #1925, opened 2018-05-14
ui_tests: extend test_selinuxusermap.py suite

Rezney:
Extend test_selinuxusermap.py suite with new test cases. Details in the ticket.
    
We also modify "add_table_associations" to handle "cancel" and "negative" in the way other methods works.
    
Lastly, we start using dialog_btn=None to test keyboard confirmation as we did use it incorrectly with "Negative=True" where it was already confirmed by "click".

### DIFF
--- a/ipatests/test_webui/test_selinuxusermap.py
+++ b/ipatests/test_webui/test_selinuxusermap.py
@@ -25,22 +25,29 @@ from ipatests.test_webui.ui_driver import UI_driver
 from ipatests.test_webui.ui_driver import screenshot
 import ipatests.test_webui.data_user as user
 import ipatests.test_webui.data_group as group
+import ipatests.test_webui.data_selinuxusermap as selinuxmap
 import ipatests.test_webui.data_hostgroup as hostgroup
+import ipatests.test_webui.data_hbac as hbac
 from ipatests.test_webui.test_host import host_tasks, ENTITY as HOST_ENTITY
 import pytest
 
-ENTITY = 'selinuxusermap'
-PKEY = 'itest-selinuxusermap'
-DATA = {
-    'pkey': PKEY,
-    'add': [
-        ('textbox', 'cn', PKEY),
-        ('textbox', 'ipaselinuxuser', 'user_u:s0'),
-    ],
-    'mod': [
-        ('textarea', 'description', 'itest-selinuxusermap desc'),
-    ],
-}
+try:
+    from selenium.webdriver.common.keys import Keys
+    from selenium.webdriver.common.action_chains import ActionChains
+except ImportError:
+    pass
+
+RULE_ALR_EXIST = 'SELinux User Map rule with name "{}" already exists'
+RULE_UPDATED = 'SELinux User Map {} updated'
+RULE_ADDED = 'SELinux User Map successfully added'
+INVALID_SEUSER = 'SELinux user {} not found in ordering list (in config)'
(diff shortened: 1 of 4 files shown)


ipatests/test_webui/test_selinuxusermap.py
```diff
@@ -89,9 +99,22 @@ def test_mod(self):
...
         self.mod_rule_tables(tables, categories, [])
 
+        # add associations then cancel
+        def get_t_vals(t):
+            table = t[0]
+            k = t[1]
+            e = []
+            if len(t) > 2:
+                e = t[2]
+            return table, k, e
```
pvoborni: This helper method seems redundant to me. Can be solved just by:

```
for t in tables:
     table = t[0]
     keys = t[1]        
     self.add_table_associations(table, [keys[0]], confirm_btn='cancel')
```

I also wonder why the same method in `mod_rule_tables` has the `e` in it. But that's another story.

ipatests/test_webui/test_selinuxusermap.py
```diff
@@ -25,22 +25,30 @@
 from ipatests.test_webui.ui_driver import screenshot
 import ipatests.test_webui.data_user as user
 import ipatests.test_webui.data_group as group
+import ipatests.test_webui.data_selinuxusermap as selinux
```
pvoborni: `selinux` is bit misleading. `selinuxmap` or `selinuxusermap` seems better

ipatests/test_webui/test_selinuxusermap.py
```diff
@@ -104,9 +127,245 @@ def test_actions(self):
...
+        # test selinux usermap undo button
+        self.add_record(selinux.ENTITY, selinux.DATA)
+        self.navigate_to_record(selinux.PKEY)
+        self.fill_fields(selinux.DATA['mod'])
+        self.assert_undo_button('description')
+        desc_undo_btn_e = '.widget [name="description"] .btn.undo'
+        desc_undo_btn = self.find(desc_undo_btn_e, By.CSS_SELECTOR)
+        desc_undo_btn.click()
+        self.assert_undo_button('description', visible=False)
+        self.verify_btn_action(mod_description)
```
pvoborni: these 4 'undo' lines look as a perfect candidate for a new ui_driver method. E.g. click_undo(field)

ipatests/test_webui/test_selinuxusermap.py
Rezney: Yep, makes sense. Fixed.

ipatests/test_webui/test_selinuxusermap.py
Rezney: Fixed.

ipatests/test_webui/test_selinuxusermap.py
Rezney: New method introduced.

varunmylaraiah:
@Rezney SelinuxUsermap test failed in "test disable enable multiple SELinux rules"

Rezney:
Aah, I see the problem now. The record is already there so it is firstly deleted and it is throwing one more notification which conflicts with expected one.

Rezney:
master:

* 8d508b8e954aca460949b5151422bf5f59f9eee8 ui_tests: extend test_selinuxusermap.py suite
* 0959c47676d11fa73ab10fb0559f8dc045d7f8dd ui_tests: add click_undo_button() func

PR #1931, opened 2018-05-16
Install shared templates for client setup too

abbra:
This patchset updates a spec file and our Make files to allow installing shared templates for the client build as well. So far we only have a single template for krb5.conf.d (freeipa.template) but there might be potentially more coming in.

This is a proper solution -- in Fedora 28/29 builds downstream I'm copying the files manually to avoid re-spinning a tarball with 4.7.0rc2. We still need to filter out client side templates from the server side in the spec file, so there is not a single place to define all files, unfortunately, but that's good enough for now as we have only a single file used by the client side.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2,7 +2,7 @@ ACLOCAL_AMFLAGS = -I m4
 
 if ENABLE_SERVER
     IPASERVER_SUBDIRS = ipaserver
-    SERVER_SUBDIRS = daemons init install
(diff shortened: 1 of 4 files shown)


abbra:
Pushed an update that should make tox targets work again.

rcritten:
Chose https://github.com/freeipa/freeipa/pull/1953 in favor of this one, closing.

PR #1935, opened 2018-05-17
ipa vault-archive overwrites an existing value without warning

amitkumar50:
Upstream ticket was raised for issuing an warning message
whenever data in ipa vault is overwritten.

In Bugzilla(1339129) its agreed upon that Current behavior is consistent
with other IPA commands. None of ipa mod commands asks for confirmation
and therefore it should be the same here.
But to document, that vault can contain only one value in ipa help vault.

This PR addresses the changes agreed in Bugzilla.

Resolves: https://pagure.io/freeipa/issue/5922

### DIFF
--- a/ipaserver/plugins/vault.py
+++ b/ipaserver/plugins/vault.py
@@ -55,7 +55,10 @@ Vaults
 """) + _("""
 Manage vaults.
 """) + _("""
-Vault is a secure place to store a secret.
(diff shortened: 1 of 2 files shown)


amitkumar50:
wrongly commited changes of another pr into into it. closing this Would be opening new one.

PR #1937, opened 2018-05-17
pylint3: workaround false positives reported for W1662

abbra:
Pylint3 falsely reports warning W1662: using a variable that was bound
inside a comprehension for the cases where the same name is reused for a
loop after the comprehension in question.

Rename the variable in a loop to avoid it.

If the code looks like the following:

  arr = [f for f in filters if callable(f)]
  for f in arr:
      result = result + f()

pylint3 would consider 'f' used outside of comprehension. Clearly, this
is a false-positive warning as the second 'f' use is completely
independent of the comprehension's use of 'f'.

### DIFF
--- a/ipaclient/csrgen.py
+++ b/ipaclient/csrgen.py
@@ -144,8 +144,8 @@ class Formatter(object):
                 for rule in field_mapping.data_rules]
 
             data_sources = []
-            for rule in field_mapping.data_rules:
-                data_source = rule.options.get('data_source')
+            for xrule in field_mapping.data_rules:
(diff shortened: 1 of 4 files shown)


tiran:
LGTM

Your patch is one good way to work around the issue. You could also turn ``[f for f in items]`` into a generator ``list(f for f in items)``. Generator expressions don't leak their inner var on Python 2.

rcritten:
master:

* c61151f6aa0c033834aed70561fc762c06176555 pylint3: workaround false positives reported for W1662

PR #1941, opened 2018-05-18
Extended UI test for selfservice permission.

mrizwan93:
Follwoing scenario added:
 - test_add_all_attr
 - test_add_and_add_another
 - test_add_and_edit
 - test_add_and_cancel
 - test_add_permission_undo
 - test_add_permission_reset
 - test_permission_negative
 - test_del_multiple_permission
 - test_permission_using_enter_key
 - test_reset_sshkey_permsission


ipatests/test_webui/test_selfservice.py
```diff
@@ -49,4 +59,230 @@ def test_crud(self):
...
+    @screenshot
+    def test_del_multiple_permission(self):
+        """
+        Try to delete multiple self service permission
+        """
+        self.init_app()
+        self.add_record(data_selfservice.ENTITY, data_selfservice.DATA1)
+        self.add_record(data_selfservice.ENTITY, data_selfservice.DATA)
+        self.delete_record([data_selfservice.DATA1['pkey'],
+                            data_selfservice.DATA['pkey']])
```
pvoborni: Would be good to add some asserts, e.g `assert_record(xy, negative=True) after `delete_record` call that they are deleted because `delete_record will happily not do anything if e.g. you are on a different page.

ipatests/test_webui/test_selfservice.py
```diff
@@ -49,4 +59,230 @@ def test_crud(self):
...
+        self.add_sshkey_to_record(user.SSH_RSA, user.PKEY, navigate=True)
+        self.assert_num_ssh_keys(1)
+        self.close_notifications()
+        self.logout()
+
+        self.init_app()
+        self.navigate_to_entity(data_selfservice.ENTITY)
+        self.navigate_to_record('Users can manage their own SSH public keys')
+        self.check_option('attrs', 'carlicense')
+        self.check_option('attrs', 'ipasshpubkey')
```
pvoborni: Would be good to add a small comment that it unchecks `ipasshpubkey`. And that checkingof carlicense is there to pass validator.

ipatests/test_webui/test_selfservice.py
```diff
@@ -49,4 +59,230 @@ def test_crud(self):
...
+        self.facet_button_click('save')
+        self.close_notifications()
+        self.logout()
+
+        # check if delete button is visible for ssh key
+        self.login(user.PKEY, password=pwd)
+        self.assert_disabled('[name="ipasshpubkey"]', negative=True)
```
pvoborni: This test is not right. `self.assert_disabled(.., negative=True)` basically test if it is enabled. This assert will pass in every situation regardless of previous steps. The test needs to check if a user can add or remove SSH Keys. 

There are two ways how to do it, maybe combination is correct:

- SSH Keys widget doesn't have, "add" and "delete" buttons visible
- After clicking on "Show/Set key", the textarea in the dialog is editable.

ipatests/test_webui/test_selfservice.py
```diff
@@ -49,4 +60,247 @@ def test_crud(self):
...
+        """
+        self.init_app()
+
+        # try to add duplicate entry
+        self.add_record(data_selfservice.ENTITY, data_selfservice.DATA1)
+        self.add_record(data_selfservice.ENTITY, data_selfservice.DATA1,
+                        negative=True, pre_delete=False)
+        self.assert_last_error_dialog(ENTRY_EXIST)
+        self.close_all_dialogs()
+        self.delete(data_selfservice.ENTITY, data_selfservice.DATA1)
```
Rezney: This will fail as delete() requires list here.

Rezney:
@pvoborni please let us know if your concerns were addressed...

ipatests/test_webui/test_selfservice.py
```diff
@@ -49,4 +60,245 @@ def test_crud(self):
...
+
+        # check if delete button is visible for ssh key
+        self.login(user.PKEY, password=pwd)
+        s = "button[name=remove]"
+        facet = self.get_facet()
+        btn = self.find(s, By.CSS_SELECTOR, facet, strict=True)
+        assert btn.is_displayed is not True
+
+    @screenshot
+    def test_del_leftover_permission(self):
```
Rezney: Could you please elaborate on this test? Why we are deleting all the permissions here?

.freeipa-pr-ci.yaml
```diff
@@ -27,183 +27,14 @@ jobs:
...
-    job:
-      class: RunPytest
-      args:
-        build_url: '{fedora-27/build_url}'
-        test_suite: test_integration/test_authselect.py
-        template: *ci-master-f27
-        timeout: 3600
-        topology: *master_1repl_1client
-
+        topology: *ipaserver
```
Rezney: Tests are currently not run as we are missing "ipaserver" declaration. Please add it to "topologies":

```
  ipaserver: &ipaserver
    name: ipaserver
    cpu: 1
```

Rezney: You can use "yaml" validator to check if everything is ok.

https://codebeautify.org/yaml-validator

ipatests/test_webui/test_selfservice.py
mrizwan93: I also had doubt about it that why we need to delete the default permissions. But it is listed in the TCMS plan that's why written it.

Rezney: Ah I see, well it does not make much sense for me. However my point here is that after every test we should restore IPA to previous state so other tests are not affected. Lets discard the last test as it does not bring any value. Thanks...

Rezney:
@mrizwan93 finally found the issue. Please change back class to "RunWebuiTests" from "RunPytest" in WebUI test blob of PRCI yaml file because this way tests are also run however without WebUI wrapper script...

Rezney:
@mrizwan93 Thanks for working on this PR. LGTM. I would maybe introduce something like "assert_button" in ui_driver instead of:

`
s = "button[name=remove]"
facet = self.get_facet()
btn = self.find(s, By.CSS_SELECTOR, facet, strict=True)
assert btn.is_displayed is not True
`

but let`s not block this PR by this for now.

ipatests/test_webui/test_selfservice.py
```diff
@@ -49,4 +60,231 @@ def test_crud(self):
...
+        self.check_option('attrs', 'ipasshpubkey')
+        self.facet_button_click('save')
+        self.close_notifications()
+        self.logout()
+
+        # check if delete button is visible for ssh key
+        self.login(user.PKEY, password=pwd)
+        s = "button[name=remove]"
+        facet = self.get_facet()
+        btn = self.find(s, By.CSS_SELECTOR, facet, strict=True)
```
pvoborni: This test assumes that there is no other widget with 'remove' button visible. E.g. if there would be any editable multivalued widget(e.g. telephone number) with some value set then this test would incorrectly fail as it could pick the 'remove' button from the other widget. Right now it is not a problem but might create hard to debug issue in the future. IMO the selector `s` should be limited to `ipasshpubkey` widget.

Rezney:
Tests passing. ACK

Rezney:
Aah, nope @mrizwan93 please remove the temp commit

rcritten:
master:

* 3927b0e7b1a6de1734d2e3b4e283dcc7d3f0a406 Extended UI test for selfservice permission.

PR #673, opened 2017-03-29
Conf template

tjaalton:
Move conf templates to a common location, make ipa.conf and named.conf portable.

PR #1949, opened 2018-05-21
Changes for Debian

tjaalton:
Here's a list of accumulated patches for Debian. Some have tickets attached, others not.


ipaplatform/debian/tasks.py
```diff
@@ -11,14 +11,20 @@
...
                                              was_sssd_configured):
-        # Debian doesn't use authconfig, nothing to restore
-        return True
+        ret = True
+        try:
+            ipautil.run(["pam-auth-update",
+                         "--package", "--remove", "mkhomedir"])
+        except ipautil.CalledProcessError:
+            ret = False
+        return ret
```
stlaz: A little nit:
```python
try:
    ...
except ipautil.CalledProcessError:
    return False
return True
```
The above looks nicer to me :)

ipaplatform/debian/tasks.py
```diff
@@ -27,8 +33,13 @@ def set_nisdomain(nisdomain):
...
     @staticmethod
     def modify_nsswitch_pam_stack(sssd, mkhomedir, statestore):
-        # Debian doesn't use authconfig, this is handled by pam-auth-update
-        return True
+        if mkhomedir:
+            try:
+                ipautil.run(["pam-auth-update", "--package", "--enable", "mkhomedir"])
+            except ipautil.CalledProcessError:
+                return False
+        else:
```
stlaz: I believe you did not mean to have this `else` here since it makes the function return `None` if `mkhomedir` resolves to `True` and the `ipautil.run()` passes.

ipaserver/install/bindinstance.py
```diff
@@ -805,6 +805,8 @@ def __setup_sub_dict(self):
             NAMED_VAR_DIR=paths.NAMED_VAR_DIR,
             BIND_LDAP_SO=paths.BIND_LDAP_SO,
             INCLUDE_CRYPTO_POLICY=crypto_policy,
+            DATA_DIR='' if ipaplatform.NAME == 'debian' else 'data/',
+            DISABLE_DEBIAN='//' if ipaplatform.NAME == 'debian' else '',
```
stlaz: I do not like platform-specific code outside the `ipaplatform` module.

ipaserver/install/httpinstance.py
```diff
@@ -235,12 +236,13 @@ def set_mod_ssl_protocol(self):
                                    '+TLSv1 +TLSv1.1 +TLSv1.2', False)
 
     def set_mod_ssl_logdir(self):
-        installutils.set_directive(paths.HTTPD_SSL_CONF,
-                                   'ErrorLog',
-                                   'logs/error_log', False)
-        installutils.set_directive(paths.HTTPD_SSL_CONF,
-                                   'TransferLog',
-                                   'logs/access_log', False)
+        if NAME is not 'debian':
```
stlaz: Could you please handle this once again in the `ipaplatform` module? Perhaps extend the tasks with something like `setup_httpd_logging()`.

ipaserver/install/bindinstance.py
```diff
@@ -805,6 +805,8 @@ def __setup_sub_dict(self):
             NAMED_VAR_DIR=paths.NAMED_VAR_DIR,
             BIND_LDAP_SO=paths.BIND_LDAP_SO,
             INCLUDE_CRYPTO_POLICY=crypto_policy,
+            DATA_DIR='' if NAME == 'debian' else 'data/',
+            DISABLE_DEBIAN='//' if NAME == 'debian' else '',
```
stlaz: Please handle platform-specific code in the `ipaplatform` module.

ipaserver/install/ldapupdate.py
```diff
@@ -328,12 +329,15 @@ def _identify_arch(self):
            etc.  Determine if a suffix is needed based on the current
            architecture.
         """
-        bits = platform.architecture()[0]
-
-        if bits == "64bit":
-            return "64"
+        if NAME is 'debian':
```
stlaz: Here the guilt is kind of on our side. Starting Fedora 28, the server code only supports 64-bit architectures (https://bugzilla.redhat.com/show_bug.cgi?id=1544386) so we could return `64` on Fedora and whatever is returned on Debian, have that to be returned there.

tjaalton: the problem is that what it returns is platform specific, so fedora dropping 32bit support doesn't help me at all

tjaalton: err, arch specific

ipaplatform/debian/tasks.py
tjaalton: sure

ipaplatform/debian/tasks.py
tjaalton: it needs to return True if mkhomedir is not set, but I've added another 'return True' before else: to match your suggestion for restore_pre_ipa_client_configuration above

ipaserver/install/bindinstance.py
tjaalton: ok, this was reviewed earlier in https://github.com/freeipa/freeipa/pull/673 so I just slapped it in here

ipaserver/install/ldapupdate.py
stlaz: Now it's been some time since I last played with `ipaplatform`, but wouln't it help if we add `LIBARCH='64'` to `ipaplatform/base/paths.py` and have the current `LIBARCH="/{0}".format(MULTIARCH)` in `ipaplatform/debian/paths.py`?
Also, is this a path or a constant? I believe it might be the latter.

stlaz: Or maybe really the other way around, have Fedora override the `"/{0}".format(MULTIARCH)` since it seems more accurate.

tjaalton: something like that yes

ipaplatform/debian/tasks.py
stlaz: Not sure whether I can't see the change or you did not push it yet, but anyway - just remove the `else:` and the padding in front of `return True` and you'll achieve the same if I am not mistaken.

ipaserver/install/ldapupdate.py
```diff
@@ -328,12 +329,15 @@ def _identify_arch(self):
...
+            return paths.LIBARCH
         else:
-            return ""
+            bits = platform.architecture()[0]
```
tiran: Please use ``sys.maxsize`` here: https://docs.python.org/3/library/sys.html#sys.maxsize

stlaz: I'd really rather go with my proposal above and get all platform/arch specific code to `ipaplatform`

tiran: That's a good idea, too.

ipaplatform/debian/tasks.py
tjaalton: I now pushed what I wrote, missed your latest comment

ipaplatform/debian/tasks.py
```diff
@@ -11,13 +11,19 @@
...
 
+from ipapython import ipautil
 
 class DebianTaskNamespace(RedHatTaskNamespace):
     @staticmethod
     def restore_pre_ipa_client_configuration(fstore, statestore,
                                              was_sssd_installed,
                                              was_sssd_configured):
-        # Debian doesn't use authconfig, nothing to restore
+        ret = True
```
stlaz: This is now unused.

ipaplatform/redhat/tasks.py
```diff
@@ -47,6 +47,7 @@
 from ipaplatform.paths import paths
 from ipaplatform.redhat.authconfig import get_auth_tool
 from ipaplatform.base.tasks import BaseTaskNamespace
+from ipaserver.install import installutils
```
stlaz: This causes the import the fail, I'll try to see if I can sort it out for you.

tjaalton: wow, that's a big hammer :)

stlaz:
Pushed a possible fix to the wrong import, untested. Let's see what happens.

tiran:
@tjaalton please rebase, @stlaz fixed the pycodestyle issue on master.

stlaz:
@tjaalton rebased for you and force-pushed into your branch, you may want to re-download it again should you want to do any changes to it :)

tjaalton:
Nah, I think this set of patches are enough for now, I'll prep another one after this has landed

stlaz:
If you could possibly fix these two errors:
```
************* Module ipaplatform.debian.tasks
ipaplatform/debian/tasks.py:21: [W0612(unused-variable), DebianTaskNamespace.restore_pre_ipa_client_configuration] Unused variable 'ret')
************* Module ipaserver.install.ldapupdate
ipaserver/install/ldapupdate.py:30: [W0611(unused-import), ] Unused import platform)
```
I'll find us someone to review our work :)

tjaalton:
I already did and pushed yesterday, but your changed overrode those..

stlaz:
Oh, darn, sorry, will fix it for you :smile:

stlaz:
Alright, hopefully all will be alright now, sorry about the mix-up.

tjaalton:
fingers crossed! :)

stlaz:
Should have expected that, did not like the move to `ipautil` myself. Will create a separate module for it so that I can still be viewed as a human being in the eyes of other developers :slightly_smiling_face:

tiran:
master:

* 0030118ddc81af2df215a31b3863e0f560332130 Create kadm5.acl if it doesn't exist
* 7a27651a0a7eaa8cedd15d05fe89116f7b64b3c0 constants: Fix HTTPD_GROUP for Debian
* a3a3d6da5bc88e0ec35420fb8f02aa0379248f3a paths: Fix some path definitions for Debian.
* 86ef31d76081bde396788b6585af96b09dbb3333 Add mkhomedir support for Debian
* c5ee8ae5297f1686f4af74e74c284860515c2dc6 named.conf: Disable duplicate zone on debian, and modify data dir
* ffdb20aeb31cf3b5f02f261aff11418ee3cf02d2 ldapupdate: Add support for Debian multiarch
* 8c0d7bb92ff3d0b13702b41faa47e96f697f222e Fix HTTPD SSL configuration for Debian.
* f47d86c719aa67f62f3d0c54f5270fc0fc8d1393 Move config directives handling code

PR #1951, opened 2018-05-22
Cosmetic changes for ipa pwpolicy command

amitkumar50:
If you try to change the value of the password policy to the
same value as it already has then you will get an error:

ipa: ERROR: no modifications to be performed.

PR changes this error message to:
ipa: ERROR: Stored value is same as provided

https://pagure.io/freeipa/issue/362

### DIFF
--- a/ipalib/errors.py
+++ b/ipalib/errors.py
@@ -1519,11 +1519,11 @@ class EmptyModlist(ExecutionError):
     >>> raise EmptyModlist()
     Traceback (most recent call last):
       ...
-    EmptyModlist: no modifications to be performed
+    EmptyModlist: Stored value is same as provided
     """
 
     errno = 4202
-    format = _('no modifications to be performed')
+    format = _('Stored value is same as provided')
 
 
 class DatabaseError(ExecutionError):


abbra:
The problem with the proposed solution is that it changes output for every other command as well because EmptyModlist is an exception used widely across IPA framework code. There are currently 118 references to EmptyModlist. So this is by no means a 'cosmetic change'.

amitkumar50:
Also without modifying EmptyModlist keeping 'EmptyModlist: no modifications to be performed' and only changing format = _('Stored value is same as provided'). We are able to get desired results
`# ipa pwpolicy-mod --minlength=9`
ipa: ERROR: stored value is same as provided

abbra:
I don't get that. What you change `format` in the exception definition, you *are* changing the EmptyModlist exception definition.

rcritten:
I think he means that the exception name hasn't changed (except EmptyModlist), just the text of the exception has. I don't know that we 

This is also raised other situations where the message is even less clear (e.g. ipaldap::move_entry)

I think this is one of those tickets that looks trivial and ends up being a very big onion (many layers and quite stinky).

amitkumar50:
yes agree @abbra EmptyModlist exception is called from many places. Do you suggest a better way for this particular use case.

abbra:
I don't think this ticket is still valid. Unless we analyse how we want to respond to these situations, I'd rather not change EmptyModlist to provide another message.

rcritten:
I'm inclined to agree, and to close the ticket as won't fix.

rcritten:
Closing PR per previous comments.

PR #1952, opened 2018-05-22
ipa-server-install: publish complete cert chain in /usr/share/ipa/html/ca.crt

flo-renaud:
When IPA is installed with an externally signed CA, the master installer does not publish the whole cert chain in /usr/share/ipa/html/ca.crt (but /etc/ipa/ca.crt contains the full chain).

If a client is installed with a One-Time Password and without the --ca-cert-file option, the client installer downloads the cert chain from http://master.example.com/ipa/config/ca.crt (=/usr/share/ipa/html/ca.crt). The client installation then fails. Note that when the client is installed by providing admin/password, installation succeeds because the cert chain is read from the LDAP server.

Implements the fix and associated integration test.

https://pagure.io/freeipa/issue/7526

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -61,7 +61,7 @@ jobs:
         test_suite: test_integration/test_external_ca.py::TestExternalCA test_integration/test_external_ca.py::TestSelfExternalSelf test_integration/test_external_ca.py::TestExternalCAInstall
         template: *ci-master-f28
         timeout: 3600
-        topology: *master_1repl
+        topology: *master_1repl_1client
 
   fedora-28/test_topologies:
     requires: [fedora-28/build]
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -432,7 +432,7 @@ class HTTPInstance(service.Service):
             raise RuntimeError("HTTPD cert was issued by an unknown CA.")
         # at this time we can assume any CA cert will be valid since this is
         # only run during installation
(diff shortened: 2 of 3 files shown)


flo-renaud:
Hi @rcritten 
yes, I tested the installation of a rhel 6 client (3.0.0-50.el6.1) with

```
(on master) ipa host-add $CLIENT --random
(on client) ipa-client-install --domain $DOMAIN --realm $REALM --password $OTP
(on client) kinit admin
(on client) ipa host-find
```
Is there anything else I should try on the client?

Travis failure seems to happen because the tests are run unordered even though they inherit from ipatests/test_integration/base.py which is using the @ordered annotation. I need to investigate this.

rcritten:
Perhaps with an external CA and/or with a 3rd party cert (manual testing is ok IMHO, these clients aren't going to be changing).

flo-renaud:
This update should fix the test ordering.
Tested with a rhel 6 client, enrolling to

- an externally signed CA master
- a CA-less master with ca > subca > master certificate chain

flo-renaud:
Adding the re-run label as fedora-27/test_sudo failed with"RunPytest Failed to publish artifacts "

tiran:
Patch doesn't apply any more and needs a rebase

flo-renaud:
Patch rebased, waiting for PR-CI results

tiran:
master:

* af99032d901d55e56bccdc272cfbf3617de05b53 ipa-server-install: publish complete cert chain in /usr/share/ipa/html/ca.crt
* 1d70ce850e965a2d5475895aa88668756a6810b3 Test for 7526

PR #1953, opened 2018-05-22
Move client templates to separate directory

tiran:
PR https://github.com/freeipa/freeipa/pull/1747 added the first template
for FreeIPA client package. The template file was added to server
templates, which broke client-only builds.

The template is now part of a new subdirectory for client package shared
data.


rcritten:
master:

* 992a5f482319369c231456e311bb316dd7747016 Move client templates to separate directory

PR #1954, opened 2018-05-22
Use GnuPG 2 rather than old GnuPG 1.4

tiran:
The /usr/bin/gpg command is old, legacy GnuPG 1.4 version. The
recommended version is GnuPG 2 provided by /usr/bin/gpg2. For our simple
uses-cases, gpg2 is a drop-in replacement for gpg.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1566019

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -757,6 +757,7 @@ Provides: python2-ipaplatform = %{version}-%{release}
 Requires: %{name}-common = %{version}-%{release}
 Requires: python2-gssapi >= 1.2.0-5
 Requires: gnupg
+Requires: gnupg2
 Requires: keyutils
 Requires: python2 >= 2.7.9
 Requires: python2-cryptography >= 1.6
@@ -810,6 +811,7 @@ Provides: python3-ipaplatform = %{version}-%{release}
 Requires: %{name}-common = %{version}-%{release}
 Requires: python3-gssapi >= 1.2.0
 Requires: gnupg
+Requires: gnupg2
 Requires: keyutils
 Requires: python3-cryptography >= 1.6
 Requires: python3-netaddr >= %{python_netaddr_version}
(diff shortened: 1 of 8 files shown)


rcritten:
Changes look ok. In DL0 I was able to install gpg2-signed replica file on a gpg-enabled replica and vice versa, so we maintain compatibility.

This needs to be wired into ipa-backup and ipa-restore as well.

I think we'll need to either require a certain named key or add an option to allow it to be overridden (or both). A simple one can be generated using:

 $ gpg2 --quick-gen-key freeipa

I tested encryption with:

 $ gpg2 --batch -o /tmp/out --default-recipient-self -e VERSION.m4

but I was never prompted which key to use when I had more than one which is odd. I'm not sure what "self" means in this case which is why I think a named key is safer (-R freeipa).

We'd need a BZ doc bug to update the new mechanism(s), including how to generate the key, what files to backup, etc. There is no need to worry about older backups because restores only work when the version matches exactly.

tiran:
For asymmetric encryption, GnuPG has two modes of operation:

* by default, GnuPG includes all recipient key ids and encryption subkey ids in the header of the encrypted blob.
* in hidden / anonymous recipient mode, GnuPG usually tries all private encryption keys until it is able to decrypt the header.

Therefore GnuPG does not have to prompt for key. It may prompt for all passphrases of all your keys, though.

stlaz:
```
>       agent.wait(5)
E       TypeError: wait() takes exactly 1 argument (2 given)
```

tiran:
Stupid Python 2

tiran:
Fixed upstream
master:
https://pagure.io/freeipa/c/dbc3788405b1d57c20946a98e40ca27c8ebac302
https://pagure.io/freeipa/c/8e165480ace76ab97e40e9396293eccff36497e0

PR #1957, opened 2018-05-23
Making nigthly test definition editable by FreeIPA contributors

felipevolpone:
Now the test definition of nightly tests will be on freeipa repo. The the definition that's used on every PR (previously as .freeipa-pr-ci.yaml) is in ipatests/prci_definitions/gating and the .freeipa-pr-ci.yaml file
is just a symlink to the real file.

In the same dir there is also nightly_master and nightly_rawhide, both to be used in nightly tests.

This PR is the result of the discussion on freeipa-devel mailing list [1].

[1] https://lists.fedoraproject.org/archives/list/freeipa-devel@lists.fedorahosted.org/message/4VAWJ4SFKKBFFICDLQCTXJWRRQHIYJLL/

PS: This PR depends on [this PR in freeipa-pr-ci](https://github.com/freeipa/freeipa-pr-ci/pull/201)


pvoborni:
> PS: This PR depends on this PR in freeipa-pr-ci

Isn't it the other way around? https://github.com/freeipa/freeipa-pr-ci/pull/201 changes the way how nightlies are created and depend on the definitions delivered in this PR. Gating will work regardless of freeipa/freeipa-pr-ci#201, won't' it?

tiran:
@flo-renaud @Rezney Can you take care of this PR when it is convenient to you?

tiran:
Superseded by PR https://github.com/freeipa/freeipa/pull/2139

PR #1961, opened 2018-05-25
Extended UI test for Certificates

mrizwan93:
UI test for certificates

### DIFF
--- a/ipatests/test_webui/test_cert.py
+++ b/ipatests/test_webui/test_cert.py
@@ -23,9 +23,97 @@ Cert tests
 
 from ipatests.test_webui.ui_driver import UI_driver
 from ipatests.test_webui.ui_driver import screenshot
+from datetime import date, timedelta
 import pytest
 
 ENTITY = 'cert'
+
+CERT_CSR = ("""-----BEGIN NEW CERTIFICATE REQUEST-----
+MIICcjCCAVoCAQAwLTERMA8GA1UEChMISVBBLlRFU1QxGDAWBgNVBAMTD21hc3Rl
+ci5pcGEudGVzdDCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAK9xZ/OT
+PC9vlwP78ACCYTOGy8C+Ho4OUF4p4PYPkns4Zdov6N/f0MlUcxY7cpAlKlAZT53b
+OphSKamc7nlNgUAzV1S6JTm1YXiTzG6mAOWWK/i3O25rNfiz0D+srlqS0ADsGPwG
+4vxuWJYwUEKcV4YgCYCJj78dD+yxoz5anVrNosN4tVGg6jfaPI9uu1T+FhsNM/AC
+EN7pa50bgeV7iBZs/pdZEXEsk18NO4W55yPAKQp5JFCL6eeBJcHTU/IuEb/YPCSr
+e873Iwzw4ZqW8dHbE10Za3VzdKPDUbtJp93rvU6imRavvifIAa3GgBuLYbpEXXcG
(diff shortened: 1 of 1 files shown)


ipatests/test_webui/data_cert.py
```diff
@@ -0,0 +1,18 @@
+ENTITY = 'cert'
```
stlaz: Why did you create this file?

mrizwan93: @stlaz I kept the data used in this testsuite in the mentioned file. As of now there is not much data in it, but as we will progress to add more testcases in it, its good to have separate file for data, isn't it?

ipatests/test_webui/test_cert.py
```diff
@@ -45,7 +92,404 @@ def test_read(self):
...
+        """
+        self.init_app()
+        self.navigate_to_entity(data_cert.ENTITY)
+
+        # add a new cert.
+        self.facet_button_click('request_cert')
+        self.fill_textbox('principal', 'HTTP/master.testrelm.test')
+        self.check_option('add', 'checked')
+        self.fill_textarea('csr', data_cert.CERT_CSR)
+        self.dialog_button_click('issue')
```
frasertweedale: It would be good, for readability, to break the setup (the cert request and revocation) into a separate method, preceding the others.

ipatests/test_webui/data_cert.py
frasertweedale: Yeah I agree with @stlaz, it would be more straightforward to just define `CERT_CSR` and `ENTITY` in the module where they're used.  If there ends up being a lot more data for other testcases in future, we can look at how to keep things tidy at that point in time.

mrizwan93:
adding re-run as pr-ci failed with "no description"

ipatests/test_webui/test_cert.py
```diff
@@ -23,10 +23,93 @@
...
+
+def check_space_error(self, string, option):
+    self.navigate_to_entity(ENTITY)
+    self.select('select[name=search_option]', option)
+    search_pkey(self, string)
+    self.assert_last_error_dialog(ERR_SPACE.format(option))
+    self.close_all_dialogs()
+
+
+def check_integer(self, string, option):
```
Rezney: I would recommend to add a docstring here as "check_integer" is not much descriptive from first look.

ipatests/test_webui/test_cert.py
```diff
@@ -49,3 +132,392 @@ def test_read(self):
...
+    @screenshot
+    def test_search_revocation_reason(self):
+        """
+        Try to search certificates by revocation reason
+        """
+        self.init_app()
+        self.navigate_to_entity(ENTITY)
+
+        # add a new cert
+        record = add_cert(self, 'HTTP/master.testrelm.test', CERT_CSR)
```
Rezney: We do not have "master.testrelm.test" host yet so the test will fail (and is failing in PRCI indeed)

Rezney: Looks like other 2 fails in PRCI are consequence of the first one.

Rezney:
Please rebase with latest master:

These should be already fixed:
```

test_service_test_add_service_using_enter
test_user_test_certificates.png
```

These seem to be new. First one looks like some timing issue.
Lets see if they reappear.
```

test_service_test_certificates
test_user_test_add_user_special
```

Rezney:
Tests are passing. Please remove the temp commit and we can push it. Thanks...

Rezney:
master:

* bdc3e3c58c94982e2c65c07823bd7e15041a87e8 Extended UI test for Certificates

PR #1965, opened 2018-05-28
Require python-ldap with fix for ref counting bug

tiran:
python-ldap 3.1.0 and python3-pyldap 2.4.37-4 fix a segfault caused by
a reference counting bug.

See: https://pagure.io/freeipa/issue/7324


tiran:
New package isn't available in F27, yet.

stlaz:
Please, just remove this Fedora version duality, I doubt that we're testing our package with `python3-pyldap` anyway.

tiran:
I have addressed your comment and added this PR to https://github.com/freeipa/freeipa/pull/1969

PR #1966, opened 2018-05-28
[Backport][ipa-4-6] Don't try to set Kerberos extradata when there is no principal

tiran:
This was causing ns-slapd to segfault in the password plugin.


abbra:
LGTM.

flo-renaud:
ipa-4-6:

* 92595cc36b89c06b419299d923e630fb94b4d42a Don't try to set Kerberos extradata when there is no principal

PR #1967, opened 2018-05-28
[Backport][ipa-4-5] Don't try to set Kerberos extradata when there is no principal

tiran:
This was causing ns-slapd to segfault in the password plugin.


abbra:
LGTM

flo-renaud:
ipa-4-5:

* 4065b9978d6da5dc950ed2f1ac27e35c36195162 Don't try to set Kerberos extradata when there is no principal

PR #1970, opened 2018-05-29
Print version string in installer

tiran:
The server, replica, and client installer now print the current version
number on the console, before the actual installer starts. It makes it
easier to debug problems with failed installations. Users typically post
the console output in a ticket.

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -27,7 +27,7 @@ from ipaclient.install.ipachangeconf import IPAChangeConf
 import ipaclient.install.timeconf
 from ipalib.install import certstore, sysrestore
 from ipalib.install.kinit import kinit_keytab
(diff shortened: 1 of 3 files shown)


rcritten:
Do we want a logger.debug for this as well?

tiran:
The version information is already in the debug logs (``ipa*-install.log``)

ipaserver/install/server/replicainstall.py
```diff
@@ -680,6 +680,10 @@ def install_check(installer):
     filename = installer.replica_file
     installer._enrollment_performed = False
 
+    print("This program will set up the FreeIPA replica.")
```
rcritten: nit: set up a FreeIPA replica instead of "the"

Otherwise LGTM

rcritten:
master:

* cf25823e997ef4f5036470413522d9d25d6b11a7 Print version string in installer

PR #1971, opened 2018-05-29
Disable Schema Compat plugin during server upgrade

rcritten:
If this is enabled it can cause a deadlock with SSSD trying
to look up entries and it trying to get data on AD users
from SSSD.

When reading the entry from LDIF try to get the camel-case
nsslapd-pluginEnabled and fall back to the all lower-case
nsslapd-pluginenabled if that is not found. It would be nice
if the fetch function was case sensitive but this is likely
overkill as it is, but better safe than blowing up.

Upon restoring it will always write the camel-case version.

https://pagure.io/freeipa/issue/6721

### DIFF
--- a/ipaserver/install/upgradeinstance.py
+++ b/ipaserver/install/upgradeinstance.py
@@ -37,6 +37,7 @@ from ipaserver.install import service
 logger = logging.getLogger(__name__)
 
 DSE = 'dse.ldif'
+COMPAT_DN = "cn=Schema Compatibility,cn=plugins,cn=config"
 
 
 class GetEntryFromLDIF(ldif.LDIFParser):
@@ -113,6 +114,7 @@ class IPAUpgrade(service.Service):
         self.step("saving configuration", self.__save_config)
         self.step("disabling listeners", self.__disable_listeners)
         self.step("enabling DS global lock", self.__enable_ds_global_write_lock)
+        self.step("disabling Schema Compat", self.__disable_schema_compat)
         self.step("starting directory server", self.__start)
         if self.schema_files:
             self.step("updating schema", self.__update_schema)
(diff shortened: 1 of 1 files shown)


ipaserver/install/upgradeinstance.py
```diff
@@ -161,6 +163,29 @@ def __save_config(self):
             else:
                 self.backup_state('nsslapd-global-backend-lock', global_lock)
 
+        with open(self.filename, "r") as in_file:
+            parser = GetEntryFromLDIF(in_file, entries_dn=[COMPAT_DN])
+            parser.parse()
+
+            try:
+                compat_entry = parser.get_results()[COMPAT_DN]
```
stlaz: I don't think you still need to have the file opened here.

ipaserver/install/upgradeinstance.py
```diff
@@ -161,6 +163,29 @@ def __save_config(self):
...
+            except KeyError:
+                pass
```
stlaz: Just add a `return` here and remove the else below.

ipaserver/install/upgradeinstance.py
```diff
@@ -161,6 +163,29 @@ def __save_config(self):
...
+            else:
+                try:
+                    schema_compat_enabled = compat_entry[
+                        'nsslapd-pluginEnabled'][0].decode('utf-8')
```
stlaz: Is the returned entry just a normal dictionary, does python-ldap not set case for all the keys, or doesn't python-ldap have their own caseless dictionary for these cases?

ipaserver/install/upgradeinstance.py
```diff
@@ -161,6 +163,29 @@ def __save_config(self):
...
+                    self.backup_state('schema_compat_enabled',
+                                      schema_compat_enabled)
+                except KeyError:
+                    try:
+                        schema_compat_enabled = compat_entry[
```
stlaz: You really should duplicate code less:
```python
schema_compat_enabled = compat_entry.get('nsslapd-pluginEnabled')
if schema_compat_enabled is None:
     schema_compat_enabled = compat_entry.get('nsslapd-pluginenabled')
if schema_compat_enabled:
    self.backup_state('schema_compat_enabled', schema_compat_enabled)
```

ipaserver/install/upgradeinstance.py
rcritten: Done.

rcritten:
This uses the same process to change the dse.ldif values that has been used for quite some time. I saw no AVCs in F28.

stlaz:
Went again through the code, all LGTM, I trust you there're no AVCs and that it works :stuck_out_tongue:

rcritten:
master:

* 2b3eb5c567e1e5d4fe7d945e36615cef3dd4d144 Disable Schema Compat plugin during server upgrade

PR #1974, opened 2018-05-30
Handle compressed responses from Dogtag

frasertweedale:
We currently accept compressed responses for some Dogtag resources,
via an 'Accept: gzip, deflate' header.  But we don't decompress the
received data.  Inspect the response Content-Encoding header and
decompress the response body according to its value.

The `gzip.decompress` function is only available on Python 3.2 or
later.  In earlier versions, it is necessary to use StringIO and
treat the compressed data as a file.  This commit avoids this
complexity.  Therefore it should only be included in Python 3 based
releases.

Fixes: https://pagure.io/freeipa/issue/7563

### DIFF
--- a/ipapython/dogtag.py
+++ b/ipapython/dogtag.py
@@ -19,6 +19,7 @@
 
 import collections
 import gzip
+import io
 import logging
 import xml.dom.minidom
 import zlib
@@ -61,6 +62,15 @@ INCLUDED_PROFILES = {
 
 DEFAULT_PROFILE = u'caIPAserviceCert'
 KDC_PROFILE = u'KDCs_PKINIT_Certs'
+
+
+if six.PY3:
+    gzip_decompress = gzip.decompress  # pylint: disable=no-member
+else:
+    # note: gzip.decompress available in Python >= 3.2
+    def gzip_decompress(data):
+        with gzip.GzipFile(fileobj=io.BytesIO(data)) as f:
+            return f.read()
 
 
 def error_from_xml(doc, message_template):
@@ -232,8 +242,7 @@ def _httplib_request(
 
     encoding = res.getheader('Content-Encoding')
     if encoding == 'gzip':
-        # note: gzip.decompress available in Python >= 3.2
-        http_body = gzip.decompress(http_body)  # pylint: disable=no-member
+        http_body = gzip_decompress(http_body)
     elif encoding == 'deflate':
         http_body = zlib.decompress(http_body)
 
--- a/ipapython/dogtag.py
+++ b/ipapython/dogtag.py
@@ -18,8 +18,10 @@
 #
 
 import collections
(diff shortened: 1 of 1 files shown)


rcritten:
Can't this be done more easily using mod_filter:
```
<LocationMatch "^/ca/rest/account/login|...>
   ...
    FilterChain gzinflate
</Location>
```
I'm just nore sure how to reproduce this in Fedora to test the theory

frasertweedale:
@rcritten what does that do?  Automatically inflate compressed response bodies on the server side?
Even after Apache itself does the deflate?

I'd say that the compression (of response, from server to client) is potentially desirable and we should just handle it in the client.  That said, I'd sooner remove the 'Accept: gzip, deflate' header from requests, and never worry about receiving compressed response bodies, than deploy more Apache config hacks :)

rcritten:
Apache doesn't do the deflate right now which is why the client (IPA) is receiving gzip'd data.

mod_filter is designed exactly for this type of use-case. This is no different in Apache than encrypting/decrypting data for TLS connections. It is just one more filter to pass through for compressed data so that when it hits mod_wsgi it is decompressed.

I honestly don't know why I configured the IPA rest client to accept compressed data but given the possible size of a cert-find it makes sense to compress that data. Remember that a find can be run on an IPA master without a CA so this data is not necessarily remaining only on the host.

ipapython/dogtag.py
```diff
@@ -228,8 +230,15 @@ def _httplib_request(
         logger.debug("httplib request failed:", exc_info=True)
         raise NetworkError(uri=uri, error=str(e))
 
+    encoding = res.getheader('Content-Encoding')
+    if encoding == 'gzip':
+        # note: gzip.decompress available in Python >= 3.2
+        http_body = gzip.decompress(http_body)  # pylint: disable=no-member
```
tiran: This will fail with an attribute error on Python 2. Debian still uses Python 2.7. This should do the trick:

```
import gzip
import io
import six

if six.PY3:
    gzip_decompress = gzip.decompress
else:
    def gzip_decompress(data):
        with GzipFile(fileobj=io.BytesIO(data)) as f:
            return f.read()
```

Code based on https://github.com/python/cpython/blob/3.6/Lib/gzip.py#L527

frasertweedale:
Thanks @tiran.  I'll update the PR tomorrow.

tiran:
@frasertweedale I updated your PR for you.

tiran:
master:

* 1da3eddf56450c4e41d92796fc303fdd0315690d Handle compressed responses from Dogtag
* 0a87de5ed523e0dc55d42a3e35882928c923e117 Backport gzip.decompress for Python 2

frasertweedale:
@tiran thanks for the Python 2 implementation and closing this out.

PR #1978, opened 2018-05-30
Add test for  --external-cert-file points to a non-existing file or invalid file.

amore17:


ipatests/test_integration/test_commands.py
```diff
@@ -133,3 +133,48 @@ def test_change_sysaccount_password_issue7561(self):
...
+    related to https://pagure.io/freeipa/issue/6985
+    """
+    num_replicas = 1
+
+    def test_install_ca_replica(self):
+
+        tasks.install_master(self.master)
+        replica = self.replicas[0]
+
+        # install ca with non-existing cert
```
flo-renaud: Hi @amore17 
in this test the master is installed with a self-signed CA. The CA installation on the replica will have the same self-signed CA and the argument --external-cert-file is ignored.
If the test goal is to check the behavior when --external-cert-file points to a non-existing file, I believe that the test should rather:
- install a CA_less master
- install the replication
- use the 2 step process with ipa-ca-install --external-ca / ipa-ca-install --external-cert-file to install an externally signed CA.

rcritten:
I don't understand why tests for two separate tickets are included in the same PR. The title reflects on the second change.

Please update the commit message to follow the format outlined at https://www.freeipa.org/page/Contribute/Code#Create_pull_request_on_Github

PR #1979, opened 2018-05-30
WebUI: Extend password policy tests with more scenarios

varunmylaraiah:
Extend WebUI test_pwpolicy suite with the following test cases
Added tests:
    krbpwdminlength: lower range integer
    krbmaxpwdlife: non-integer, abc
    krbmaxpwdlife: upper range integer,2147483648
    krbmaxpwdlife: lower range integer,-1
    krbminpwdlife: non-integer,edf
    krbminpwdlife: upper range integer,2147483648
    krbminpwdlife: lower range integer,-1
    krbpwdhistorylength: non-integer,HIJ
    krbpwdhistorylength: upper range integer,2147483648
    krbpwdhistorylength: lower range integer,-1
    krbpwdmindiffchars: noon-integer,3lm
    krbpwdmindiffchars: upper range integer,2147483648
    krbpwdmindiffchars: lower range integer, -1
    krbpwdminlength: non-integer, n0p
    krbpwdminlength: upper range integer,2147483648
    krbpwdminlength: lower range integer, -1
    cospriority: non-integer, abc
    cospriority: upper range integer,2147483648
    cospriority: lower range integer,-1
    krbpwdmaxfailure: non-integer
    krbpwdmaxfailure: upper range integer
    krbpwdmaxfailure: lower range integer
    krbpwdfailurecountinterval: non-integer
    krbpwdfailurecountinterval: upper range integer
    krbpwdfailurecountinterval: lower range integer
    krbpwdlockoutduration: non-integer
    krbpwdlockoutduration: upper range integer
    krbpwdlockoutduration: lower range integer
    deletePolicy_with various scenario
    MeasurementUnitAdded_Bug798363
    Delete global password policy
    add_Policy_adder_dialog_bug910463
    delete_Policy_deleter_dialog_bug910463
    test field: cospriority
    modifyPolicy(undo/refresh/reset)
    empty policy name
    upper bound of data range
    lower bound of data range
    non integer for policy priority

Details in the ticket https://pagure.io/freeipa/issue/7574

### DIFF
--- a/ipatests/test_webui/test_pwpolicy.py
+++ b/ipatests/test_webui/test_pwpolicy.py
@@ -23,26 +23,22 @@ Password policy tests
 
 from ipatests.test_webui.ui_driver import UI_driver
 from ipatests.test_webui.ui_driver import screenshot
+import ipatests.test_webui.data_group as group
+import ipatests.test_webui.data_pwpolicy as pwpolicy
 import pytest
 
-ENTITY = 'pwpolicy'
-DATA = {
-    'pkey': 'admins',
-    'add': [
-        ('combobox', 'cn', 'admins'),
-        ('textbox', 'cospriority', '364'),
-    ],
-    'mod': [
-        ('textbox', 'krbmaxpwdlife', '3000'),
-        ('textbox', 'krbminpwdlife', '1'),
-        ('textbox', 'krbpwdhistorylength', '0'),
-        ('textbox', 'krbpwdmindiffchars', '2'),
-        ('textbox', 'krbpwdminlength', '2'),
-        ('textbox', 'krbpwdmaxfailure', '15'),
-        ('textbox', 'krbpwdfailurecountinterval', '5'),
-        ('textbox', 'krbpwdlockoutduration', '3600'),
-    ],
-}
+try:
+    from selenium.webdriver.common.keys import Keys
+    from selenium.webdriver.common.action_chains import ActionChains
+except ImportError:
+    pass
+
+FIELDS = ['krbmaxpwdlife', 'krbminpwdlife', 'krbpwdhistorylength',
+          'krbpwdmindiffchars', 'krbpwdminlength', 'krbpwdmaxfailure',
+          'krbpwdfailurecountinterval', 'krbpwdlockoutduration',
+          'cospriority']
+EXPECTED_ERR = "invalid 'group': cannot delete global password policy"
+EXPECTED_MSG = 'Password Policy successfully added'
 
 
 @pytest.mark.tier1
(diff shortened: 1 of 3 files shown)


Rezney:
@pvoborni Could you please also take a look if you find free cycles?

ipatests/test_webui/test_pwpolicy.py
```diff
@@ -23,9 +23,17 @@
...
 
-ENTITY = 'pwpolicy'
+try:
+    from selenium.webdriver.common.by import By
+    from selenium.webdriver.common.keys import Keys
+    from selenium.webdriver.common.action_chains import ActionChains
+except ImportError:
+    pass
+
 DATA = {
```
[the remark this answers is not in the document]
varunmylaraiah: It was there from the beginning, So I left it there. Do you think we should move it?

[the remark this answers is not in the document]
varunmylaraiah: Moved to data_pwpolicy

ipatests/test_webui/test_pwpolicy.py
```diff
@@ -54,4 +51,214 @@ def test_crud(self):
...
+        # test add and edit record
+        self.add_record(pwpolicy.ENTITY, pwpolicy.DATA,
+                        dialog_btn='add_and_edit')
+
+        # test add password policy for special characters group
+        self.add_record(pwpolicy.ENTITY, pwpolicy.DATA_SPECIAL_CHAR,
+                        delete=True)
+
+        # empty group and priority (requires the field)
+        self.navigate_to_entity(pwpolicy.ENTITY)
```
pvoborni: is the navigation needed? Shouldn't we be already on the page?

ipatests/test_webui/test_pwpolicy.py
```diff
@@ -54,4 +51,214 @@ def test_crud(self):
...
+        self.facet_button_click('remove')
+        self.dialog_button_click('ok')
+        self.assert_last_error_dialog(EXPECTED_ERR, details=True)
+        actions = ActionChains(self.driver)
+        actions.send_keys(Keys.TAB)
+        actions.send_keys(Keys.ENTER).perform()
+        self.wait(0.5)
+        self.assert_record(pwpolicy.DEFAULT_POLICY)
+
+        # test add/delete Passwordpolicy confirming using ENTER key ticket-3200
```
pvoborni: a nit pick, wrt `ticket-3200`, we are not very consistent in source code when referencing tickets but this looks like yet another notation.  Could we stick to something already used? E.g. `ticket 3200` or `ticket #3200`.

ipatests/test_webui/test_pwpolicy.py
```diff
@@ -54,4 +51,214 @@ def test_crud(self):
...
+        self.assert_field_validation(minimum_value_expected_error,
+                                     field=field_priority)
+        # upper bound of data range
+        self.fill_input(field_priority, maximum_value)
+        self.assert_field_validation(expect_error='Maximum value is'
+                                                  ' 2147483647',
+                                     field=field_priority)
+        self.close_all_dialogs()
+
+    def modify_policy(self, pwdfield, expected_error, value):
```
pvoborni: the method name doesn't describe the purpose much. The intention here is not to modify the policy (the actual mod is not even executed) but rather test field validation after changing values. The method name should reflect that.

ipatests/test_webui/test_pwpolicy.py
```diff
@@ -54,4 +51,214 @@ def test_crud(self):
...
+                        dialog_btn='add_and_edit')
+
+        for field in FIELDS:
+            # bigger than max value
+            if field == 'krbmaxpwdlife':
+                self.modify_policy(field, 'Maximum value is 20000',
+                                   maximum_value)
+            elif field == 'krbpwdmindiffchars':
+                self.modify_policy(field, 'Maximum value is 5', maximum_value)
+            else:
```
pvoborni: This if, elif, else construction could be a bit misleading. Based on it, one would imagine that there are three fields but there is much more. Not sure if it can be enhanced just by comment what is the test or by changing of a structure or using a map.

ipatests/test_webui/test_pwpolicy.py
```diff
@@ -54,4 +51,214 @@ def test_crud(self):
...
+            modified_value = (pwpolicy.DATA_RESET['mod'][mod_field][2])
+            self.facet_button_click('revert')
+            self.verify_btn_action(field, modified_value)
+            mod_field + 1
+
+        # cleanup
+        self.navigate_to_entity(pwpolicy.ENTITY)
+        self.delete_record(pwpolicy.group.PKEY6)
+
+    def verify_btn_action(self, field, mod_value, negative=True):
```
pvoborni: This method has a specific purpose: check that value of field is equal or not equal to some provided value. But the name is quite generic, not describing what it does. So it could easily by `field_equals` or something like that.

ipatests/test_webui/test_pwpolicy.py
```diff
@@ -54,4 +51,214 @@ def test_crud(self):
...
+        current_value = self.get_field_value(field, element="input")
+        if negative:
+            assert current_value != mod_value
+        else:
+            assert current_value == mod_value
+
+    @screenshot
+    def test_verify_measurement_unit(self):
+        """
+        verifying measurement unit for password policy ticket-2437
```
pvoborni: again the ticket notation.

ipatests/test_webui/test_pwpolicy.py
```diff
@@ -54,4 +51,214 @@ def test_crud(self):
...
+        krbpwdlockoutduration = \
+            self.get_text('label[name="krbpwdlockoutduration"]')
+        assert "Max lifetime (days)" in krbmaxpwdlife
+        assert "Min lifetime (hours)" in krbminpwdlife
+        assert "History size (number of passwords)" in krbpwdhistorylen
+        assert "Failure reset interval (seconds)" in krbpwdfailurecountinterval
+        assert "Lockout duration (seconds)" in krbpwdlockoutduration
+
+        # cleanup
+        self.delete(group.ENTITY, [group.DATA, group.DATA2, group.DATA3,
```
pvoborni: this cleanup, or parts of it, should be in the respected methods so that the tests are self-containing. This way we are relying on specific tests execution order.

ipatests/test_webui/test_pwpolicy.py
varunmylaraiah: Thanks for catching this. I have removed.

ipatests/test_webui/test_pwpolicy.py
varunmylaraiah: This cleanup is related to pre_data

ipatests/test_webui/test_pwpolicy.py
pvoborni: The change doesn't reflect what I meant. The added comment is just repeating the code. That part was clear. 

I meant something like:
```python
# 'cn'("group") field is read-only and all other fields are integers without special validator
else:
                self.check_expected_error(field, 'Maximum value is 2147483647',
                                          maximum_value)
```

ipatests/test_webui/test_pwpolicy.py
pvoborni: The issue is that all the delete data is created in `test_misc` but cleaned in `test_verify_measurement_unit`. But those test are not related. Web UI tests don't inherit from IntegrationTest which has `@ordered` decorator to ensure test order. Also Web UI tests don't use this decorator. Thus the reason why it works could be purely random. There can be test execution where `test_verify_measurement_unit` is not the last and `test_misc` is not almost first. I've a feeling that it works only because the tests are sorted alphabetically. 

Each test should prepare it's own data and also clean them.

varunmylaraiah: Thank you @pvoborni for comment.
Will fix it.

pavelpicka:
code looks good, with all checks passed

tiran:
master:

* 39ac5f442a73aceb8690a5f07fcf08ea71893962 ui_tests: extend test_pwpolicy.py suite

PR #1980, opened 2018-05-30
Fix elements not being removed in otpd_queue_pop_msgid()

frozencemetery:
If the element being removed were not the queue head,
otpd_queue_pop_msgid() would not actually remove the element, leading
to potential double frees and request replays.


stlaz:
ACK. How nice it would be to have this unit-tested.

abbra:
ACK from my side too. Nice find.

rcritten:
master:

* a2e8d989a359690f493930461adf7f52a0cc745d Fix elements not being removed in otpd_queue_pop_msgid()

npmccallum:
@frozencemetery Great catch! Thanks!

PR #1987, opened 2018-06-01
Uninstall fix for named-pkcs11

slaykovsky:
Sometimes named-pkcs11 is not being stopped or reloaded during
uninstall and it causes a lot of problems while testing, for example,
backup and restore tests are failing because of ipa-server-install
fails on checking DNS step.

Also reverting 415578a19 as it's not needed anymore and additional
uninstall/install take a lot of time.

Fixes backup/restore tests runs. Maybe something else.

### DIFF
--- a/ipaserver/install/bindinstance.py
+++ b/ipaserver/install/bindinstance.py
@@ -1211,9 +1211,13 @@ class BindInstance(service.Service):
         # disabled by default, by ldap_enable()
         if enabled:
             self.enable()
+        else:
+            self.disable()
 
         if running:
             self.restart()
+        else:
+            self.stop()
 
         self.named_regular.unmask()
         if named_regular_enabled:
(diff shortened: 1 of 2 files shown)


rcritten:
IMHO you should do the revert as a separate commit.

I'm a little confused about the other change. It seems to suggest that the enabled/running commands don't always reflect reality. Enabled/disabled is a binary right? Same with running/stopped. Is there some deeper issue that this is going to cover up?

rcritten:
Ok, I get it now. The changes look ok, please just move the revert into a separate change which does a proper revert.

slaykovsky:
@rcritten Thank you very much! Will do.

tiran:
master:

* ec9ea73b630499a239ebfa8aef73f0e529001f3e Uninstall fix for named-pkcs11
* 283987c1dff716bcfa77569d53cf7809b3f0b98a Revert "Fixing TestBackupAndRestore::test_full_backup_and_restore_with_removed_users"

PR #1989, opened 2018-06-01
test for --external-cert-file points to a non-existing or invalid file

amore17:
Related to : https://pagure.io/freeipa/issue/6985
 Test for - ipa-ca-install command installs CA even if cert file is not specified with --external-cert-file option

### DIFF
--- a/ipatests/test_integration/test_caless.py
+++ b/ipatests/test_integration/test_caless.py
@@ -1181,6 +1181,36 @@ class TestReplicaInstall(CALessBase):
         if self.domain_level > DOMAIN_LEVEL_0:
             self.verify_installation()
 
+    @replica_install_teardown
+    def test_install_ca_replica(self):
+        # related to https://pagure.io/freeipa/issue/6985
+        replica = self.replicas[0]
+
+        self.create_pkcs12('ca1/replica', filename='server.p12')
+        result = self.prepare_replica(http_pkcs12='server.p12',
+                                      dirsrv_pkcs12='server.p12')
+        assert result.returncode == 0
+        if self.domain_level > DOMAIN_LEVEL_0:
+            self.verify_installation()
+
+        contents = (
+            '-----BEGIN CERTIFICATE-----\n'
+            'sdnmsdkfbsdifbsdbasdsdSDDDasdmnd\n'
+            '-----END CERTIFICATE-----')
+        cert1 = tempfile.mkdtemp(suffix='xyz.crt', dir=paths.TMP)
+        replica.put_file_contents(cert1, contents)
+        cert_files = [cert1]
+        # install ca directly with option --external-cert-file
+        result = tasks.install_ca(replica, cert_files=cert_files,
+                                  raiseonerr=False)
+        assert_error(result, 'First run the installer with --external-ca')
+
+        # install ca with invalid cert
+        tasks.install_ca(self.master, external_ca=True)
+        result = tasks.install_ca(replica, cert_files=cert_files,
+                                  raiseonerr=False)
+        assert_error(result, 'CA is not installed yet')
+
 
 class TestClientInstall(CALessBase):
     num_clients = 1
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -51,17 +51,29 @@ jobs:
         timeout: 3600
         topology: *master_1repl
 
-  fedora-28/external_ca:
+  fedora-28/external_ca_1:
     requires: [fedora-28/build]
     priority: 50
     job:
       class: RunPytest
       args:
         build_url: '{fedora-28/build_url}'
-        test_suite: test_integration/test_external_ca.py::TestExternalCA test_integration/test_external_ca.py::TestSelfExternalSelf test_integration/test_external_ca.py::TestExternalCAInstall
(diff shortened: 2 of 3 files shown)


ipatests/test_integration/test_caless.py
```diff
@@ -1181,6 +1181,36 @@ def test_no_ds_password(self):
...
+        result = tasks.install_ca(replica, cert_files=cert_files,
+                                  raiseonerr=False)
+        assert_error(result, 'First run the installer with --external-ca')
+
+        # install ca with invalid cert
+        tasks.install_ca(self.master, external_ca=True)
+        result = tasks.install_ca(replica, cert_files=cert_files,
+                                  raiseonerr=False)
+        assert_error(result, 'CA is not installed yet')
+
```
flo-renaud: Hi @amore17 
thank you for the test. If I understand correctly, it is testing the following scenario:

- install master CA-less
- install replica
- call ipa-ca-install --external-cert-file on the replica (without the first step ipa-ca-install --external-ca)
It is a valid test case, but I think that the bug 6985 was rather related to the following scenario:
- install master with CA
- install replica
- run ipa-ca-install --external-cert-file on the replica
In this case the replica should complain that CA instance is already configured on the master and external-cert-file will be ignored.

amore17: Thanks @flo-renaud for review I'll add that scenario in ipatests/test_integration/test_commands.py
should I remove this test-case ?

ipatests/test_integration/test_commands.py
```diff
@@ -133,3 +134,61 @@ def test_change_sysaccount_password_issue7561(self):
...
+        ipa-ca-install command installs CA on replica even if cert file
+        is not specified with --external-cert-file option. If executed
+        command with non-existing file, invalid file etc,
+        """
+        # related to https://pagure.io/freeipa/issue/6985
+        master = self.master
+
+        replica = self.replicas[0]
+        tasks.install_master(self.master)
+        tasks.install_replica(master, replica)
```
flo-renaud: Need to add the arg setup_ca=False otherwise ipa-replica-install will be called with --setup-ca

ipatests/test_integration/test_commands.py
```diff
@@ -133,3 +134,61 @@ def test_change_sysaccount_password_issue7561(self):
...
+        replica.put_file_contents(cert1, contents)
+        cert_files = [cert1]
+
+        # install ca with invalid cert
+        result = tasks.install_ca(replica, cert_files=cert_files,
+                                  raiseonerr=False)
+        assert result.returncode != 0
+        if result.returncode == 0:
+            tasks.uninstall_replica(master, replica)
+            tasks.install_replica(master, replica)
```
flo-renaud: Same comment as above, need to call tasks.install_replica(master,replica, setup_ca=False)

ipatests/test_integration/test_commands.py
```diff
@@ -133,3 +134,61 @@ def test_change_sysaccount_password_issue7561(self):
...
+        cmd1 = ["ipa-ca-install", "-U", "-p",
+                replica.config.dirman_password, "-P", 'admin',
+                "-w", replica.config.admin_password,
+                '--external-cert-file=']
+        result1 = replica.run_command(cmd1, raiseonerr=False)
+        assert result1.returncode != 0
+
+        if result1.returncode == 0:
+            tasks.uninstall_replica(master, replica)
+            tasks.install_replica(master, replica)
```
flo-renaud: same comment as above, need to call tasks.install_replica(master, replica, setup_ca=False)

ipatests/test_integration/test_commands.py
```diff
@@ -133,3 +134,61 @@ def test_change_sysaccount_password_issue7561(self):
...
+
+        # install ca with non-existent cert
+        cmd2 = ["ipa-ca-install", "-U", "-p",
+                replica.config.dirman_password, "-P", 'admin',
+                "-w", replica.config.admin_password,
+                '--external-cert-file=abc.crt']
+        result2 = replica.run_command(cmd2, raiseonerr=False)
+        assert result2.returncode != 0
```
flo-renaud: You can also test calling ipa-ca-install with --external-ca, it should complain that a CA instance is already installed and --external-ca would be ignored.

ipatests/test_integration/test_caless.py
flo-renaud: You can keep the test, it's a valid test case.

flo-renaud:
Hi @amore17 
there are some PEP-8 errors:
```
./ipatests/test_integration/test_replica_promotion.py:715:18: E225 missing whitespace around operator
./ipatests/test_integration/test_replica_promotion.py:727:16: E225 missing whitespace around operator
./ipatests/test_integration/test_replica_promotion.py:729:1: W391 blank line at end of file
```

flo-renaud:
Hi @amore17 
thank you for the patch. You need to revert the changes in .freeipa-pr-ci.yaml (PR #2017 will split external_ca test into 2 jobs), but apart from that, LGTM.

flo-renaud:
Adding re-run label because of `fedora-28/test_forced_client_enrolment — RunPytest timed out after 3600s `

amore17:
this should be added when issue is in fixed state,

PR #1990, opened 2018-06-01
Use replace instead of add to set new default ipaSELinuxUserMapOrder

rcritten:
The add was in effect replacing whatever data was already there
causing any custom order to be lost on each run of
ipa-server-upgrade.

https://pagure.io/freeipa/issue/6610


rcritten:
@flo-renaud yes, this is expected behavior. The intention was only to change the original default to a new one so any user-customizations should remain (whether they are sane or not).

flo-renaud:
@rcritten thanks for the explanation. ACK.

flo-renaud:
master:

* f976f6cfd8d3ca1f9bff822278086df1f999fdbe Use replace instead of add to set new default ipaSELinuxUserMapOrder

flo-renaud:
Hi @rcritten 
a manual backport is needed for ipa-4-5 and ipa-4-6, can you take care of this? Thanks.

PR #1997, opened 2018-06-06
Fix translation of python brace formatted messages

stanislavlevin:
Some of the messages cannot be translated because of a broken python brace formatting.

Simple reproducer:

Take a translatable message string from ipaserver/plugins/topology.py ```Topology management requires minimum domain level {0} ``` and put it on a test script with original formatting:
```python
#!/usr/bin/python

from ipalib.text import _
import os

os.environ['LC_ALL']='fr'
desired_value = '1'
msgtest = _('Topology management requires minimum domain level {0} '.format(desired_value))
import pdb;pdb.set_trace()
print msgtest
```

Out: 
```python
(Pdb) msgtest
Gettext('Topology management requires minimum domain level 1 ', domain='ipa', localedir=None)
(Pdb) print msgtest
Topology management requires minimum domain level 1
```
But French was expected.

Cause: The 'format' method should be called after a gettext one
to follow 'python-brace-format' rule:

```python
msgtest = unicode(_('Topology management requires minimum domain level {0} ')).format(desired_value)
```

with correct out:
```
(Pdb) msgtest
u'La gestion de la topologie n\xe9cessite le niveau de domaine minimal 1'
```

List of affected:
```
ipaserver/plugins/idrange.py:56
ipaserver/plugins/idrange.py:246
ipaserver/plugins/idrange.py:383
ipaserver/plugins/idrange.py:616
ipaserver/plugins/trust.py:688
ipaserver/plugins/topology.py:88
ipaserver/plugins/topology.py:257
ipaserver/plugins/topology.py:265
ipaserver/plugins/serverroles.py:82
ipaserver/plugins/serverroles.py:167
ipaserver/plugins/domainlevel.py:68
ipaserver/plugins/domainlevel.py:153
```

This patch should fix it.

### DIFF
--- a/ipaserver/plugins/idrange.py
+++ b/ipaserver/plugins/idrange.py
@@ -39,7 +39,7 @@ else:
     _dcerpc_bindings_installed = False
 
 
-ID_RANGE_VS_DNA_WARNING = """=======
+ID_RANGE_VS_DNA_WARNING = _("""=======
 WARNING:
 
 DNA plugin in 389-ds will allocate IDs based on the ranges configured for the
@@ -51,7 +51,7 @@ the new local range. Specifically, The dnaNextRange attribute of 'cn=Posix
 IDs,cn=Distributed Numeric Assignment Plugin,cn=plugins,cn=config' has to be
 modified to match the new range.
 =======
-"""
+""")
 
 __doc__ = _("""
 ID ranges
@@ -161,8 +161,7 @@ this domain has the SID S-1-5-21-123-456-789-1010 then 1010 id the RID of the
 user. RIDs are unique in a domain, 32bit values and are used for users and
 groups.
 
-{0}
-""".format(ID_RANGE_VS_DNA_WARNING))
+""") + ID_RANGE_VS_DNA_WARNING
 
 register = Registry()
 
@@ -243,8 +242,7 @@ class idrange(LDAPObject):
         StrEnum('iparangetype?',
                 label=_('Range type'),
                 cli_name='type',
-                doc=(_('ID range type, one of {vals}'
-                     .format(vals=', '.join(sorted(range_types))))),
+                doc=_('ID range type, one of allowed values'),
                 values=sorted(range_types),
                 flags=['no_update'],
                 )
@@ -400,8 +398,7 @@ class idrange_add(LDAPCreate):
 
     must be given to add a new range for a trusted AD domain.
 
-{0}
-""".format(ID_RANGE_VS_DNA_WARNING))
+""") + ID_RANGE_VS_DNA_WARNING
 
     msg_summary = _('Added ID range "%(value)s"')
 
@@ -615,8 +612,7 @@ class idrange_show(LDAPRetrieve):
 class idrange_mod(LDAPUpdate):
     __doc__ = _("""Modify ID range.
 
-{0}
-""".format(ID_RANGE_VS_DNA_WARNING))
+""") + ID_RANGE_VS_DNA_WARNING
 
     msg_summary = _('Modified ID range "%(value)s"')
 
--- a/ipatests/test_ipalib/test_text.py
+++ b/ipatests/test_ipalib/test_text.py
@@ -201,6 +201,12 @@ class test_Gettext(object):
         inst = self.klass('hello %(adj)s nurse', 'foo', 'bar')
         assert inst % dict(adj='tall', stuff='junk') == 'hello tall nurse'
 
+    def test_format(self):
+        inst = self.klass('{0} {adj} nurse', 'foo', 'bar')
+        posargs = ('hello', 'bye')
+        knownargs = {'adj': 'caring', 'stuff': 'junk'}
+        assert inst.format(*posargs, **knownargs) == 'hello caring nurse'
+
     def test_eq(self):
         inst1 = self.klass('what up?', 'foo', 'bar')
         inst2 = self.klass('what up?', 'foo', 'bar')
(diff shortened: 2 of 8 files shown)


tiran:
Thanks for your patch!

I don't like to have more ``unicode(_(""))`` in the code base. We'll try to get rid of Python 2. Instead IPA's translation objects should grow native support for ``format``.

- [ ] Add ``def format(self, *args, **kwargs)`` to ``ipalib.text.Gettext``, ``NGettext`` and ``ConcatenatedLazyText`` classes
- [ ] Add tests for ``format`` to ipatests.test_ipalib.test_text``. You can just copy and modify the ``test_mod`` tests for all three classes.
- [ ] Change your PR to use ``_("i18n").format()``

tiran:
I created ticket https://pagure.io/freeipa/issue/7586 to track the issue.

tiran:
Awesome! You even updated the doc examples. Thank you!

One very small nit pick: please add ``Fixes: https://pagure.io/freeipa/issue/7586`` to the git commits. Our tool chain uses the information to update and close tickets.

ipatests/test_ipalib/test_text.py
```diff
@@ -201,6 +201,12 @@ def test_mod(self):
         inst = self.klass('hello %(adj)s nurse', 'foo', 'bar')
         assert inst % dict(adj='naughty', stuff='junk') == 'hello naughty nurse'
 
+    def test_format(self):
+        inst = self.klass('{0} {adj} nurse', 'foo', 'bar')
+        posargs = ('hello', 'bye')
+        knownargs = {'adj': 'naughty', 'stuff': 'junk'}
+        assert inst.format(*posargs, **knownargs) == 'hello naughty nurse'
```
rcritten: I realize you are sticking to previous convention but I think a new nurse adjective should be chosen with a more positive message. Perhaps "compassionate". I'll see about changing the others.

rcritten: FWIW I'm going with the adjective "tall" in my patch.

stanislavlevin: 'caring nurse' is positive i think.

tiran: Sounds good to me!

stanislavlevin:
Please, pay attention that i've deleted several format() in ipaserver/plugins/trust.py and ipaserver/plugins/idrange.py due to translation at loading time instead of request one.

ipatests/test_ipalib/test_text.py
rcritten: Yeah, better than what I came up with :-)

tiran:
@abbra ``trust`` is your area of expertise. Please take a look.

ipaserver/plugins/idrange.py
```diff
@@ -243,8 +242,8 @@ class idrange(LDAPObject):
         StrEnum('iparangetype?',
                 label=_('Range type'),
                 cli_name='type',
-                doc=(_('ID range type, one of {vals}'
-                     .format(vals=', '.join(sorted(range_types))))),
+                doc=_('ID range type, one of ' +
+                      'ipa-local, ipa-ad-trust, ipa-ad-trust-posix'),
```
rcritten: Why drop the generated list in favor of a hardcoded one?

stanislavlevin: 1) format() triggers real translation. Such methods should be used for translatable objects at request time only. Otherwise message will be at server locale.

Concerning to Web UI.
On execution of json_metadata command idrange object is transformed into json by json_serialize. Which calls ```__json__``` method to Gettext / NGettext instances and the translation is performed. 
In case of _('{vals}').format - idrange stores already translated message and the carriage turns into a pumpkin.

2) unfortunately, it's impossible to specify 
```python
doc=_('ID range type, one of ' + ', '.join(sorted(range_types)))
```
output po template is 
```python
'ID range type, one of , '
```

IMO. the only solution for now.

abbra: Please replace it with
```
doc=(_('ID range type, one of {vals}').
         .format(vals=', '.join(sorted(range_types)))),
```
This way you'd have the string for translation with a formatting context and we'll fill up the values at runtime. You don't need to translate range types as they are option values.

stanislavlevin: I agree that range_types should be not translatable.
But as i said before it is not runtime context.
Please, take a look:

Even this message is here
```shell
msgunfmt /usr/share/locale/ru/LC_MESSAGES/ipa.mo | grep 'ID range type, one of {vals}' -A 1
msgid "ID range type, one of {vals}"
msgstr "Тип диапазона идентификаторов, одно из следующих значений: {vals}"
```
There is no translation in Web UI:
image

abbra: I see what you are getting to. I think we should just drop the second part here and use
```
doc=_("ID range type, one of allowed values")
```
In both Web UI API browser and in `ipa idrange-add --help` we get the list of allowed values anyway so there is no loss of context.

abbra:
In general, I like this change. Please split the last commit (use of all the changes) to separate commits per-file. We need this to allow easier backporting and maintenance.

stanislavlevin:
Hello, are there any other changes required?

abbra:
@stanislavlevin no, I think it is good to go in. ACK from my side.

tiran:
master:

* f4716b69910f082a2fe039338f4268d941792258 Add support for format method to translation objects
* 854597c411764b603ebbe15d97dbcadac321548d Use intended format() method of translation object
* 65414d1471fde54f41129d9641c0fec7160f0896 Fix formatted translations in domainlevel plugin
* 229f1608db868ced234e84e0bc7e949decd3e75e Fix translation of idrange_* commands description
* 6f245db8eacb9bdc1e248b4f6f25e5eb785478c4 Fix formatted translations in trust plugin
* 1dfdbfd8bf9657e4066a3ba5f847dbcb35228f0c Fix formatted translations of error messages in serverroles plugin
* 4b3bc490d3be109ea048f165a1929438760d8a54 Fix formatted translations of error messages in topology plugin

PR #1999, opened 2018-06-07
ui_tests: fixes for issues with sending key and focus on element

Rezney:
Fixes 2 issues in WebUI tests. One issue is that we are unable to
confirm a dialog by "Enter" keyboard - "actions.click()" helps
here to get focus on the page.

Second issue is probbaly related to screen resolution as we cannot
click to some of the action buttons (buttons which are having issue
varies).


pavelpicka:
Code looks good and checks passes, remove temp commit to ACK

Rezney:
master:

* b1f368c6829068bd3f5100829f9b36f479cc403c ui_tests: fixes for issues with sending key and focus on element

PR #2000, opened 2018-06-07
Test: when non responding forwarder is used, wrong error message is reported.

amore17:
related : https://pagure.io/freeipa/issue/6989

PR #2004, opened 2018-06-08
[Tests] Test for removal server with verbose option

gkaihorodova:
Test for removal server with verbose option
Related to https://bugzilla.redhat.com/show_bug.cgi?id=1480502

### DIFF
--- a/ipatests/pytest_plugins/integration/tasks.py
+++ b/ipatests/pytest_plugins/integration/tasks.py
@@ -717,7 +717,7 @@ def kinit_admin(host, raiseonerr=True):
 
 
 def uninstall_master(host, ignore_topology_disconnect=True,
-                     ignore_last_of_role=True, clean=True):
+                     ignore_last_of_role=True, clean=True, verbose=False):
     host.collect_log(paths.IPASERVER_UNINSTALL_LOG)
     uninstall_cmd = ['ipa-server-install', '--uninstall', '-U']
 
@@ -729,7 +729,12 @@ def uninstall_master(host, ignore_topology_disconnect=True,
     if ignore_last_of_role and host_domain_level != DOMAIN_LEVEL_0:
         uninstall_cmd.append('--ignore-last-of-role')
 
-    host.run_command(uninstall_cmd)
+    if verbose and host_domain_level != DOMAIN_LEVEL_0:
+        uninstall_cmd.append('-v')
+
+    result = host.run_command(uninstall_cmd)
+    assert "Traceback" not in result.stdout_text
+
     host.run_command(['pkidestroy', '-s', 'CA', '-i', 'pki-tomcat'],
                      raiseonerr=False)
     host.run_command(['rm', '-rf',
--- a/ipatests/test_integration/test_server_del.py
+++ b/ipatests/test_integration/test_server_del.py
@@ -185,7 +185,7 @@ class TestServerDel(ServerDelBase):
     def test_ignore_topology_disconnect_replica2(self):
         """
         tests that removal of replica2 with '--ignore-topology-disconnect'
-        destroys master for good
+        destroys master for good with verbose option for uninstallation
         """
         check_master_removal(
             self.client,
@@ -194,7 +194,7 @@ class TestServerDel(ServerDelBase):
         )
 
         # reinstall the replica
-        tasks.uninstall_master(self.replica2)
+        tasks.uninstall_master(self.replica2, verbose=True)
         tasks.install_replica(self.master, self.replica2, setup_ca=True)
 
     def test_removal_of_master_disconnects_both_topologies(self):


pvoborni:
Thanks for the patch, the test does not test the use case though. The bug was about `ipa-server-install --uninstall -U -v` but the test executes `ipa server-del -v ...` which does not run the uninstall.

gkaihorodova:
Thank you @pvoborni  for review. 
Change it for uninstall

rcritten:
PEP-8 errors:
./ipatests/pytest_plugins/integration/tasks.py:732:15: E272 multiple spaces before keyword

Can you add a link to the BZ in the commit message and explain the reasoning for this in the commit message?

I'm not sure what this is actually testing. It adds the -v option but doesn't ensure that the error message doesn't appear.

gkaihorodova:
Hello @rcritten ! 
I'll add more information to commit msg as you asked and fix pep8.
Just to give you more info: https://bugzilla.redhat.com/show_bug.cgi?id=1480502
After discussion with @pvoborni about automation for that specific test case we decided to have test for successful uninstallation with -v  to see that it won't produce a traceback and not limit it to testing specific error ("Bad file descriptor") as it is just a special case which might be rare occasion but other bugs can produce other traceback on errors and it would be good to catch them.

tiran:
hi @gkaihorodova,

code style check is still failing with ``./ipatests/pytest_plugins/integration/tasks.py:732:15: E272 multiple spaces before keyword``

gkaihorodova:
hi @tiran 
thanks, will fix it

rcritten:
I don't understand what this test does. The BZ was about a traceback in the output despite exiting successfully. This enables verbose output but doesn't check it for anything.

pvoborni:
+1 to Rob. Or to be more verbose I'd write the same comment here as in https://github.com/freeipa/freeipa/pull/2000#issuecomment-396632705 and https://github.com/freeipa/freeipa/pull/2000#issuecomment-396633879

Tiboris:
@gkaihorodova yes, i think that you need to add test like you wrote.
@rcritten ?

ipatests/test_integration/test_server_del.py
```diff
@@ -309,3 +309,12 @@ def test_forced_removal_of_master(self):
             self.replicas[0], self.master.hostname,
             ignore_last_of_role=True
         )
+
+    def test_removal_with_verbose_option(self):
+        """
+        Test remove master with verbose option
+        """
+        tasks.install_master(self.master)
+        result = tasks.uninstall_master(self.master, verbose=True)
```
Tiboris: Travis is failing because of uninstall_master does not have return we should change it :)
Assigning to function call which doesn't return

ipatests/pytest_plugins/integration/tasks.py
```diff
@@ -717,7 +717,7 @@ def kinit_admin(host, raiseonerr=True):
 
 
 def uninstall_master(host, ignore_topology_disconnect=True,
-                     ignore_last_of_role=True, clean=True):
+                     ignore_last_of_role=True, clean=True, verbose=True):
```
Tiboris: Aaaaand also maybe verbose should be set to False by default.

gkaihorodova:
@Tiboris  hello!
Thank you for review.
Trying to address all issues you mentioned, we came up to another solution for that specific use case. Instead of  writing separate test, which also required changes in uninstall task, we implemented small check inside of uninstall master task. That give us more wide coverage and various numbers of  uninstall scenarios, as each test is using master_uninstall task.

Tiboris:
As discussed with @celestian, the uninstall_master() should have verbose set to false as default. Having verbose set to true will cause that all tests using this task will have verbose output, thus could be not expected.

celestian:
Hello,

I went through the patch and I am afraid this patch is slowing slightly different issue we want to solve. 

The moethod `uninstall_master` is clean-up function. It is connected to the environment and shouldn't be used as a test. On the other hand, the method `run_server_del` is used in `test_server_del.py` which is our server deletion test suite.

I suppose, from the observation above, that the best solution is:
1. don't touch `uninstall_master`
2. use `run_server_del` for proper testing (defined `at ipatests/pytest_plugins/integration/tasks.py`)
3. check if `Traceback` is in output of `run_server_del` in functions in file `test_server_del.py` (tested issue has Traceback with return code 0)

celestian:
@Tiboris thanks. I was totally wrong, function `run_server_del` doesn't uninstall but deletes.

So, we are on the beginning.

Tiboris:
Hello @celestian is this edit acceptable?

celestian:
LGTM

celestian:
Hello @Tiboris , did you run the tests? :-)

ipatests/pytest_plugins/integration/tasks.py
```diff
@@ -729,7 +729,14 @@ def uninstall_master(host, ignore_topology_disconnect=True,
...
         uninstall_cmd.append('--ignore-last-of-role')
 
-    host.run_command(uninstall_cmd)
+    if verbose and host_domain_level != DOMAIN_LEVEL_0:
+        uninstall_cmd.append('-v')
+
+    result = host.run_command(uninstall_cmd)
+    if "Traceback" in result.stdout_text:
+        logger.error('Error: Traceback occurred during uninstallation')
+        assert False
```
rcritten: Would 'assert Traceback not in result.stdout_text be cleaner here? What does the logging buy us?

ipatests/pytest_plugins/integration/tasks.py
```diff
@@ -729,7 +729,14 @@ def uninstall_master(host, ignore_topology_disconnect=True,
     if ignore_last_of_role and host_domain_level != DOMAIN_LEVEL_0:
         uninstall_cmd.append('--ignore-last-of-role')
 
-    host.run_command(uninstall_cmd)
+    if verbose and host_domain_level != DOMAIN_LEVEL_0:
+        uninstall_cmd.append('-v')
```
rcritten: Why only for DL1+?

ipatests/pytest_plugins/integration/tasks.py
Tiboris: Okay

ipatests/pytest_plugins/integration/tasks.py
Tiboris: @gkaihorodova could you please help? :)

gkaihorodova: Because DL0 is not supported

Tiboris:
@rcritten  yes you are right, there is no reason to have it in run_server_del as well, so removed.

rcritten:
master:

* fd348773706b556bfe7660c31826b24b76c7f05e Add check for occuring traceback during uninstallation ipa master

PR #2013, opened 2018-06-11
[RFE] DNS package check should be called earlier in installation routine

amitkumar50:
In case the integrated DNS service should be used for IdM, either by using the
non-interactive '--setup-dns' option or by answering 'yes' to the interactive
'Do you want to configure integrated DNS (BIND)?' question, the check whether
the 'ipa-server-dns' package is installed, only comes after a lot of other information
have been evaluated ([server|dns|realm] name, DM and admin password).

This PR check, whether the 'ipa-server-dns' package is installed or not,
earlier in the installation routine.

Resolves: https://pagure.io/freeipa/issue/7577


pvoborni:
Hi, thanks for the patch. I'm not doing full review. That I'd leave for somebody else. But some points:

1. the PR needs rebase to not contain the 3 commits which are not in master - that was an incorrect push to Git Hub which was then overridden.
2. Would recommend to run `make fastlint` before creating the PR. The move does not contain all code so there is some indented code on the original place.
3. I have a suspicion that it won't work as it was moved before API is initialized and the dns.install_check method is using API.
4. DNS validation is quite complex and I'm not sure if the move can have unforeseen consequences. From this perspective, it might be safer to avoid a move of the entire method.

Maybe it would be simpler to
* move the package check to separate function in `ipaserver/install/dns.py`
* use that in the begging of `.../server/install.py` install check. But also use it in the dns install check as it needs to be check in the standalone installer `ipa-dns-install`

amitkumar50:
Thanks @pvoborni for review
>that was an incorrect push to Git Hub which was then overridden.

Yes I will generate a new one.
>Would recommend to run make fastlint before creating the PR. 

Will do. 
And will test moving to separate function. Also cover `ipa-dns-install`

amitkumar50:
`# yum whatprovides ipa-dns-install`
ipa-server-dns-4.2.0-15.el7_2.3.x86_64

So when we uninstall ipa-server-dns, ipa-dns-install will not be available.

Would it be still of use to move install check to dns.py

tiran:
Your code has a bug. It refers to a variable that is defined a bit later in the function. You have to rearrange the code block and move the DNS check **after** ``host_name`` is defined.

amitkumar50:
@tiran I will do that. But my point is, is it good to move check in separate function in dns.py?

tiran:
The installer needs a general overhaul and cleanup. For now, let's stick with the current coding style and leave the code inline.

amitkumar50:
I believe moving dns.install_check after retrieval of domain_name is good place to move the check up in installation routine.

......

else:
        domain_name = options.domain_name
    domain_name = domain_name.lower()

    if options.setup_dns:
        api.env.domain=domain_name
        dns.install_check(False, api, False, options, host_name)
        ip_addresses = dns.ip_addresses
    else:
        ip_addresses = get_server_ip_address(host_name,
                                             not installer.interactive, False,
                                             options.ip_addresses)

amitkumar50:
Would be generating new PR

PR #2014, opened 2018-06-11
Fedora 29 renamed fedora-domainname.service

tiran:
In Fedora 29, the fedora-domainname.service has been renamed to
nis-domainname.service like on RHEL. The ipaplatform service module for
Fedora now only renames the service, when it detects the presence of
fedora-domainname.service.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1588192
Fixes: https://pagure.io/freeipa/issue/7582

### DIFF
--- a/ipaplatform/fedora/services.py
+++ b/ipaplatform/fedora/services.py
@@ -24,14 +24,23 @@ Contains Fedora-specific service class implementations.
 
 from __future__ import absolute_import
 
+import os
+
 from ipaplatform.redhat import services as redhat_services
 
 # Mappings from service names as FreeIPA code references to these services
 # to their actual systemd service names
 fedora_system_units = redhat_services.redhat_system_units.copy()
 
-# Service that sets domainname on Fedora is called fedora-domainname.service
-fedora_system_units['domainname'] = 'fedora-domainname.service'
+# Fedora 28 and earlier have fedora-domainname.service. Starting from
+# Fedora 29, the service is called nis-domainname.service as defined in
+# ipaplatform.redhat.services.
+HAS_FEDORA_DOMAINNAME_SERVICE = os.path.isfile(
+    "/usr/lib/systemd/system/fedora-domainname.service"
+)
+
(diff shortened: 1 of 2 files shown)


tiran:
@rcritten First I looked into adding a version check for F29. This attempt would have required me to refactor some code in ``ipaplatform``. It was far simpler to just check for the presence of a file.

rcritten:
I have a change in my tree that I was going to just push as a patch for rawhide to just flip the name from fedora-domainname to nis-domainname while we work on it upstream. I may just hold off given this PR.

ipaplatform/fedora/services.py
```diff
@@ -24,14 +24,22 @@
...
 from ipaplatform.redhat import services as redhat_services
 
 # Mappings from service names as FreeIPA code references to these services
 # to their actual systemd service names
 fedora_system_units = redhat_services.redhat_system_units.copy()
 
-# Service that sets domainname on Fedora is called fedora-domainname.service
-fedora_system_units['domainname'] = 'fedora-domainname.service'
+# Fedora 28 and earlier have fedora-domainname.service. Starting from
+# Fedora 29, the service is called nis-domainname.service like on RHEL.
```
rcritten: The service is named rhel-domainname.service in RHEL. Just the comment is wrong.

rcritten:
Other than the comment this looks ok.

tiran:
@rcritten I have updated the PR and left an explanation on the bug tracker:

* The generic ``redhat`` service definition (parent of ``rhel`` and ``fedora`` platform definitions) now defines ``nis-domainname.service`` as default name for the domain name service.
* The ``fedora`` service definition overrides the name as ``fedora-domainname.service`` if and only if it detects the presence of the file ``/usr/lib/systemd/system/fedora-domainname.service``. The check will be removed when we drop support for F28.

tiran:
master:

* 907e1649580b8677d56da6207731addc178dca80 Fedora 29 renamed fedora-domainname.service

PR #2019, opened 2018-06-12
Use python3-lesscpy 0.13.0

tiran:
Require python-lesscpy 0.13. with Python 3 fix and use py3-lesscpy to
compile ipa.css.

python2-lesscpy was the last Python 2 dependency.

Fixes: https://pagure.io/freeipa/issue/7585

### DIFF
--- a/install/ui/util/make-css.sh
+++ b/install/ui/util/make-css.sh
@@ -22,5 +22,5 @@ set -o errexit
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-lesscpy -x $DIR/../less/ipa.less > $DIR/../css/ipa.css
+py3-lesscpy -x $DIR/../less/ipa.less > $DIR/../css/ipa.css
 exit $?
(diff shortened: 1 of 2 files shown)


hroncok:
> python2-lesscpy was the last Python 2 dependency.

\o/

install/ui/util/make-css.sh
```diff
@@ -22,5 +22,5 @@ set -o errexit
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-lesscpy -x $DIR/../less/ipa.less > $DIR/../css/ipa.css
+py3-lesscpy -x $DIR/../less/ipa.less > $DIR/../css/ipa.css
```
hroncok: See https://src.fedoraproject.org/rpms/python-lesscpy/pull-request/3

tiran: @hroncok That's just F29 or are you planning to land the change in F28, too?

hroncok: Nope, sorry. That would be too disturbing for stable release. You can just BR and use `/usr/bin/lesscpy` if you want "single-spec".

tiran:
master:

* 7d12bbb99bee40188168523ff685b7e8e573ba17 Use python3-lesscpy 0.13.0

PR #2029, opened 2018-06-15
Test Server-del fix

pavelpicka:
Adding and removing replication in test_ignore_topology_disconnect_replica to keep topology in good shape for next tests as reinstalling replica.

### DIFF
--- a/ipatests/test_integration/test_server_del.py
+++ b/ipatests/test_integration/test_server_del.py
@@ -167,8 +167,8 @@ class TestServerDel(ServerDelBase):
 
     def test_ignore_topology_disconnect_replica1(self):
         """
-        tests that removal of replica1 with '--ignore-topology-disconnect'
-        destroys master for good
+        tests that server-del of replica1 with '--ignore-topology-disconnect'
+        passes
         """
         check_master_removal(
             self.master,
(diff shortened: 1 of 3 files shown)


ipatests/test_integration/test_server_del.py
```diff
@@ -71,19 +71,31 @@ def check_removal_disconnects_topology(
...
         super(ServerDelBase, cls).install(mh)
-
-        cls.client = cls.clients[0]
         cls.replica1 = cls.replicas[0]
         cls.replica2 = cls.replicas[1]
 
 
+def remove_segment(master, left, right, suffix=DOMAIN_SUFFIX_NAME):
+    """
+    broke topology as in described in install method
```
pvoborni: A more precise description would be:
"""
Ensure that segment left-right or right-left don't exist
"""

As the method doesn't do anything based on the described topology.

ipatests/test_integration/test_server_del.py
```diff
@@ -103,20 +115,23 @@ def install(cls, mh):
...
-        segment_name_fmt = '{p[0].hostname}-to-{p[1].hostname}'
-        for domain_pair in permutations((cls.master, cls.replica2)):
-            tasks.destroy_segment(
-                cls.client, segment_name_fmt.format(p=domain_pair))
-
-        for ca_pair in permutations((cls.master, cls.replica1)):
-            tasks.destroy_segment(
-                cls.client, segment_name_fmt.format(p=ca_pair),
-                suffix=CA_SUFFIX_NAME)
+        # segment_name_fmt = '{p[0].hostname}-to-{p[1].hostname}'
```
pvoborni: The comment block seems like a leftover to remove.

pvoborni:
Also, it is beneficiation for future references to describe in a commit message why the changes were done - what root cause they address. An inspiration is in the previous commit messages. Also, the first line should make more clear what it does.

Good examples and reasoning are in https://chris.beams.io/posts/git-commit/ or https://wiki.openstack.org/wiki/GitCommitMessages

tiran:
Test is failing with ``OSError: [Errno 113] No route to host`` -- I'm re-running it.

PR #2032, opened 2018-06-15
Sort and shuffle SRV record by priority and weight

tiran:
On multiple occasions, SRV query answers were not properly sorted by
priority. Records with same priority weren't randomized and shuffled.
This caused FreeIPA to contact the same remote peer instead of
distributing the load across all available servers.

Two new helper functions now take care of SRV queries. sort_srv() sorts
SRV records. query_srv() combines SRV lookup with sort_srv().

Fixes: https://pagure.io/freeipa/issue/7475

### DIFF
--- a/ipaclient/install/ipadiscovery.py
+++ b/ipaclient/install/ipadiscovery.py
@@ -20,7 +20,6 @@
 from __future__ import absolute_import
 
 import logging
-import operator
 import socket
 
 import six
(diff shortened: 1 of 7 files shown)


tiran:
master:

* f90e137a173385f2ff7622c6430a6178bccd730c Sort and shuffle SRV record by priority and weight

PR #2034, opened 2018-06-18
Test that host can remove there own services

amore17:
A service may be decommissioned on the host and needs to be removed.
Related to : https://pagure.io/freeipa/issue/7486

### DIFF
--- a/ipatests/test_integration/test_service_permissions.py
+++ b/ipatests/test_integration/test_service_permissions.py
@@ -29,29 +29,29 @@ class TestServicePermissions(IntegrationTest):
     def test_service_as_user_admin(self):
         """Test that a service in User Administrator role can manage users"""
 
-        service_name = 'testservice/%s@%s' % (self.master.hostname,
-                                              self.master.domain.realm)
-        keytab_file = os.path.join(self.master.config.test_dir,
-                                   'testservice_keytab')
+        service_name1 = 'testservice1/%s@%s' % (self.master.hostname,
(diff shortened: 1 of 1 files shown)


tiran:
I don't think it is necessary to install a new master just for this test. Every master installation slows down test suite by about 10 minutes. When possible, test cases should reuse an existing master.

I suggest to add both your new test and ``test_service_access`` from ``TestServiceAuthenticationIndicators`` into ``TestServicePermissions`` class. Then choose a unique name for each test service and clean up test services after each test case.

ipatests/test_integration/test_service_permissions.py
```diff
@@ -78,61 +78,68 @@ def test_service_as_user_admin(self):
...
+                                 '-t', keytab_file2])
 
         # Verify access to service is rejected
-        result = self.master.run_command(['kvno', service_name],
+        result = self.master.run_command(['kvno', service_name2],
                                          raiseonerr=False)
         assert result.returncode > 0
+
+    def test_service_del(self):
+        """ Test that host can remove there own services.
```
rcritten: nit: replace "there" with "their" or "it's"

Not sure if you need/want to mention it here but this is also exercising that hosts can add their own services as well.

tiran:
Thanks @amore17

tiran:
master:

* 9ead70844e656b0b49433154e748769894cfb2ba Test that host can remove there own services

PR #2035, opened 2018-06-18
Installer: configure authselect with-sudo

flo-renaud:
authselect needs to be configured with the 'with-sudo' feature (except
when ipa-client-install is called with the option --no-sudo).

https://pagure.io/freeipa/issue/7562

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -2962,7 +2962,8 @@ def _install(options):
         tasks.modify_nsswitch_pam_stack(
             sssd=options.sssd,
             mkhomedir=options.mkhomedir,
-            statestore=statestore
+            statestore=statestore,
+            sudo=options.conf_sudo
         )
 
         logger.info("%s enabled", "SSSD" if options.sssd else "LDAP")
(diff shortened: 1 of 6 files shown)


flo-renaud:
Note: authselect is also working on a patch to properly handle upgrade of sudo configuration, see BZ [1582111](https://bugzilla.redhat.com/show_bug.cgi?id=1582111). A scratch build of authselect is available for testing, see [comment 7](https://bugzilla.redhat.com/show_bug.cgi?id=1577615#c7) in BZ 1577615 for the path to the scratch build.

rcritten:
Does this require a minimum version of authselect to work or is that already set?

flo-renaud:
Hi @rcritten ,
currently the master branch requires authselect >= 0.4-2. With this authselect version, my patch will make ipa installer properly configure sudo (client and server) for fresh installs, but does not handle client upgrade.
On the other hand, authselect-0.4-3.1 will be able to upgrade an ipa client and set `sudoers: files sssd` in /etc/nsswitch.conf when coming from a authconfig configuration or authselect 0.4.3 configuration.

rcritten:
Ok, thanks for the complete explanation. This looks good.

tiran:
master:

* eda831dba1e09e7f4660c64756343538042b48e0 Installer: configure authselect with-sudo

PR #2036, opened 2018-06-18
Replace some test case adjectives

rcritten:


rcritten:
@tiran next question is what branches do we want this for? Is master-only sufficient?

tiran:
master:

* 57fd79ffce8f655ae39f6e15b3f40632b4ddbd4b Replace some test case adjectives

tiran:
@rcritten Your patch merged cleanly into 4.6, so let's also fix 4.6.

PR #2037, opened 2018-06-18
ipaserver config plugin: Increase search records minimum limit

netoarmando:
Issue: https://pagure.io/freeipa/issue/6617

Check if the given search records value is greater than an arbitrary number that is not so close to zero.

The number 10 is arbitrary, I'm open for discussion.

The last test also works as a clean-up task, returning the setting back to _unlimited_.

### DIFF
--- a/ipaserver/plugins/config.py
+++ b/ipaserver/plugins/config.py
@@ -85,6 +85,18 @@ EXAMPLES:
 
 register = Registry()
 
+
+def validate_search_records_limit(ugettext, value):
+    """Check if value is greater than a realistic minimum.
+
+    Values 0 and -1 are valid, as they represent unlimited.
+    """
+    if value in {-1, 0}:
+        return
+    if value < 10:
+        return _('must be at least 10')
+
+
 @register()
 class config(LDAPObject):
     """
@@ -161,10 +173,10 @@ class config(LDAPObject):
             minvalue=-1,
         ),
         Int('ipasearchrecordslimit',
+            validate_search_records_limit,
             cli_name='searchrecordslimit',
             label=_('Search size limit'),
             doc=_('Maximum number of records to search (-1 or 0 is unlimited)'),
-            minvalue=-1,
         ),
         IA5Str('ipausersearchfields',
             cli_name='usersearch',
--- a/ipatests/test_xmlrpc/test_config_plugin.py
+++ b/ipatests/test_xmlrpc/test_config_plugin.py
@@ -227,4 +227,80 @@ class test_config(Declarative):
                         sl_domain),
             ),
         ),
+        dict(
+            desc='Set the number of search records to -1 (unlimited)',
+            command=(
+                'config_mod', [], {
+                    'ipasearchrecordslimit': u'-1',
+                },
+            ),
+            expected={
+                'result': lambda d: d['ipasearchrecordslimit'] == (u'-1',),
+                'summary': None,
+                'value': None,
+            },
+        ),
+        dict(
+            desc='Set the number of search records to greater than 10',
+            command=(
+                'config_mod', [], {
+                    'ipasearchrecordslimit': u'100',
+                },
+            ),
+            expected={
+                'result': lambda d: d['ipasearchrecordslimit'] == (u'100',),
+                'summary': None,
+                'value': None,
+            },
+        ),
+        dict(
+            desc='Set the number of search records to lower than -1',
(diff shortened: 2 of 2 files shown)


ipatests/test_xmlrpc/test_config_plugin.py
```diff
@@ -227,4 +227,42 @@ class test_config(Declarative):
...
+                'config_mod', [], {
+                    'ipasearchrecordslimit': u'0',
+                },
+            ),
+            expected={
+                'result': lambda d: d['ipasearchrecordslimit'] == (u'-1',),
+                'summary': None,
+                'value': None,
+            },
+        ),
```
rcritten: 0 is also translated into -1. Can you add that test as well while you are in here?

ipaserver/plugins/config.py
```diff
@@ -85,6 +85,13 @@
 
 register = Registry()
 
+
+def validate_search_records_limit(ugettext, value):
+    """Check if value in greater than a realistic minimum."""
+    if 0 < value < 10:
```
rcritten: This also allows < -1 which is not allowed. Only values >= -1 should be allowed.

I'd argue that 10 seconds is far too high. I might jump to 4 seconds but remember that for any one command there can be multiple searches performed so the total search limit is the accumulation of all searches. We tend to have an aggressive timeout so users won't break out of commands thinking something is wrong.

ipatests/test_xmlrpc/test_config_plugin.py
netoarmando: Actually it is what I did here. :-) Do you think I should add a test where the input is '-1'?

ipaserver/plugins/config.py
netoarmando: > This also allows < -1 which is not allowed. Only values >= -1 should be allowed.

Your are right, but those values will be rejected by the `minvalue=-1` argument. Checking for values >= -1 here will make it more explicitly for sure, but will make `minvalue` unnecessary, right? So it could be removed and this function remains as the only validation method, am I right?

> I'd argue that 10 seconds is far too high. I might jump to 4 seconds but remember that for any one command there can be multiple searches performed so the total search limit is the accumulation of all searches. We tend to have an aggressive timeout so users won't break out of commands thinking something is wrong.

I don't know if right, but I think this config sets the number of records returned by search (`ipasearchrecordslimit`), not the time which is defined by the `ipasearchtimelimit` command.

ipatests/test_xmlrpc/test_config_plugin.py
netoarmando: Added both cases.

netoarmando:
@rcritten I see, changing the expected result from an error to a real result.

netoarmando:
And what is happening to the tests? The output is displaying `timeout` errors during linter execution. Or am I missing something?

rcritten:
The test "Set the number of search records to greater than 10" failed (I didn't investigate the reason)

The tests are not independent so you just need to re-order it something like:

Set to -1
Set to 100
Set to 0

Mix in invalid tests in there if you want but if the -1 and 0 tests are next to each other then you'll get the no modifications error because a value of 0 is converted to -1 before saving.

netoarmando:
Now the tests are passing, it seems that I was using the default value (100) for the setting.

ipaserver/plugins/config.py
```diff
@@ -85,6 +85,18 @@
 
 register = Registry()
 
+
+def validate_search_records_limit(ugettext, value):
+    """Check if value in greater than a realistic minimum.
```
rcritten: nit: value is greater

ipatests/test_xmlrpc/test_config_plugin.py
```diff
@@ -227,4 +227,67 @@ class test_config(Declarative):
...
```
rcritten: So given the way the tests are run I wonder if we should reset it back to the default value of 100 at the end.

Leaving the limit as unlimited could have some subtle affect on tests that execute afterward.

netoarmando: agreed

rcritten:
Looks good, congrats on first patch!

rcritten:
master:

* de8d308196bab1dac5bba7a8a6a517a1e67e877f ipaserver config plugin: Increase search records minimum limit

PR #2044, opened 2018-06-19
[Fedora29]: fix dependency for *-domainname.service file

flo-renaud:
FreeIPA has a dependency on /usr/lib/systemd/system/*-domainname.service
file. In fedora <=28, this is provided by package 'initscripts'
but in fedora >= 29, this is provided by package 'hostname'.

Fixes:
https://pagure.io/freeipa/issue/7591


tiran:
master:

* 0b794cd43b0e8a18517cbccdec2f183ce4cb38ec fix dependency for *-domainname.service file

PR #2048, opened 2018-06-19
Always set ca_host when installing replica

tiran:
ipa-replica-install only set ca_host in its temporary
/etc/ipa/default.conf, when it wasn't installing a replica with CA. As a
consequence, the replica installer was picking a random CA server from
LDAP.

Always set the replication peer as ca_host. This will ensure that the
installer uses the same replication peer for CA. In case the replication
peer is not a CA master, the installer will automatically pick another
host later.


frasertweedale:
And my ACK!

Tiboris:
master:

* 8c3ff0308c95793a6809b8e0a3ed2a145ad3c8ea Always set ca_host when installing replica

PR #2053, opened 2018-06-21
Add ipa-advise plugin to set the CRL master

rcritten:
This might still be a bit rough around the edges but is ready for a review.

Note that for sudo to work you need to allow whatever use you are authenticating as to be able to run systemctl, grep and sed (not a very nice combination at all). I could easily drop this and require root only, I was trying to be nice.

I ran the script output through a bash linter and it is generally ok.

### DIFF
--- a/ipaserver/advise/base.py
+++ b/ipaserver/advise/base.py
@@ -408,6 +408,7 @@ class IpaAdvise(admintool.AdminTool):
     description = "Provides configuration advice for various use cases. To "\
                   "see the list of possible ADVICEs, run ipa-advise without "\
                   "any arguments."
+    shell = '/bin/sh'
 
     def __init__(self, options, args):
         super(IpaAdvise, self).__init__(options, args)
@@ -448,13 +449,13 @@ class IpaAdvise(admintool.AdminTool):
             for line in wrapped_description[1:]:
                 print("{off}{line}".format(off=' ' * len(prefix), line=line))
 
-    def print_header(self, header, print_shell=False):
+    def print_header(self, header, shell=None, print_shell=False):
         header_size = len(header)
 
         prefix = ''
         if print_shell:
             prefix = '# '
-            print('#!/bin/sh')
+            print('#!{}'.format(shell))
 
         # Do not print out empty header
         if header_size > 0:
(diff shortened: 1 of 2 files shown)


tiran:
```
PEP-8 errors:
./ipaserver/advise/plugins/crl_master.py:113:23: W605 invalid escape sequence '\$'
./ipaserver/advise/plugins/crl_master.py:127:25: W605 invalid escape sequence '\$'
```

To quote the escape sequence, you can turn the string into a raw some, e.g. ``r'awk "{ print \$3 }" | sed "s/@%s//")'``

tiran:
Tests are failing: ``'config_generic_linux_nss_pam_ldapd' object has no attribute 'shell'``

I have pushed a fix to your branch. Let's see if that solves the problem.

rcritten:
Some usage tips:

Generate the script:
`# ipa-advise set-crl-master > script.sh`

Set a new CRL generator:
`# script.sh replica.example.com`

Force setting one (one is not contactable, already master):
`# script.sh --force replica.example.com`

Use current principal to log into remote system and call sudo instead of logging in as root:
`# script.sh --sudo replica.example.com`

Log into each known master to determine who is the current master:
`# script.sh  --validate`

tiran:
On which machine do I have to execute the script? I couldn't get it to work properly.

The validate option requires an additional argument, but then ignores the argument:

```
# /tmp/script.sh 
Usage: /tmp/script.sh [--force] [--sudo] IPA_master
       /tmp/script.sh [--validate]
# /tmp/script.sh --validate
Usage: /tmp/script.sh [--force] [--sudo] IPA_master
       /tmp/script.sh [--validate]
# /tmp/script.sh --validate foo
Authenticating as root
Current setting on master.ipa.example: true
```

On the replica, it prompts for a password w/o saying what the password is required for:
```
# /tmp/script.sh --validate foo
Authenticating as root
Current setting on replica.ipa.example: false
Password: 
```

Turns out it's a ssh login on the master:
``/usr/bin/ssh root@replica.ipa.example grep ca.crl.MasterCRL.enableCRLUpdates /etc/pki/pki-tomcat/ca/CS.cfg``

rcritten:
The design was to do this as an ipa-advise plugin and after implementing it I think it was a mistake.

This is a rather large python module that generates a rather large bash script. It would be far more maintainable to create a standalone python script to perform the change, ipa-crl-master or something.

We could also include a man page for this which isn't possible for an advise script.

rcritten:
Closing this PR. This needs to be implemented as a standalone script.

PR #2057, opened 2018-06-22
Ensure that public cert and CA bundle are readable

tiran:
In CIS hardened mode, the process umask is 027. This results in some
files not being world readable. Ensure that write_certificate_list()
calls in client installer, server installer, and upgrader create cert
bundles with permission bits 0644.

Make CA bundles, certs, and cert directories world-accessible in
upgrader.

Fixes: pagure.io/freeipa/issue/7594

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -4,12 +4,14 @@
 
 from __future__ import print_function, absolute_import
 
+import errno
 import logging
 import re
 import os
 import shutil
 import pwd
 import fileinput
+import stat
 import sys
 import tempfile
 from contextlib import contextmanager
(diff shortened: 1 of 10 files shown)


tiran:
Note: I didn't change the default permission in ``write_certificate_list`` directly, because the function is also used in ipa CLI. On client side, it should obey umask.

ipaserver/install/server/upgrade.py
```diff
@@ -1700,6 +1702,34 @@ def migrate_to_authselect():
...
+        paths.KDC_CA_BUNDLE_PEM,
+        paths.IPA_CA_CRT,
+        paths.IPA_P11_KIT,
+    ]
+    for filename in candidates:
+        try:
+            s = os.stat(filename)
+        except OSError as e:
+            if e.errno != errno.ENOENT:
+                raise
```
Tiboris: Please correct me if I am wrong,
this bare raise will raise an OSError exception. This exception is not caught anywhere.
In ipaserver/install/ipa_server_upgrade.py:52 is invocation of upgrade() in try block catching "only" RuntimeError.

tiran: Yes, you are correct. It's intentional. The code block ignores ``ENOENT`` (no such file or directory), but will populate any other issue like file system or hard disk problems.

tiran: PS: Yes, the function re-raises the exception, because any other case than ENOENT is an unrecoverable error. So any unrecoverable problem will bubble up and eventually cause the installer to fail. This will only happen when the system is totally broken.

tiran:
I added a temporary commit that sets ``os.umask(0o077)`` in the ``ipaserver.install.server.install`` module. Tests were still passing.

ipaserver/install/server/upgrade.py
Tiboris: Yup this is what I meant thanks for explanation.

rcritten:
LGTM. I'm going to ack but leave pushing it in case @Tiboris wants to take another look at it.

Tiboris:
master:

* ba8cbb8c62d270772a9d70f5b2ca3bdab1e75d49 Ensure that public cert and CA bundle are readable
* 1434f2a203b123c8fedfa464f22ce13de89897de Always make ipa.p11-kit world-readable
* 89b2137dc257dfe3db0ff097e6c797223c2b5664 Make /etc/httpd/alias world readable & executable
* c2eb0f1612c920d03a8fd14863412c9cef275a3d Fix permission of public files in upgrader

Tiboris:
@tiran Manual backports are needed :(

PR #2051, opened 2018-06-20
Fix replication races in Dogtag admin code

tiran:
PR fixes two related issues:

# Fix replication races in Dogtag admin code

DogtagInstance.setup_admin and related methods have multiple LDAP
replication race conditions. The bugs can cause parallel
ipa-replica-install to fail.

The code from __add_admin_to_group() has been changed to use MOD_ADD
ather than search + MOD_REPLACE. The MOD_REPLACE approach can lead to
data loss, when more than one writer changes a group.

setup_admin() now waits until both admin user and group membership have
been replicated to the master peer.

Fixes: https://pagure.io/freeipa/issue/7593

# Improve and fix timeout bug in wait_for_entry()

replication.wait_for_entry() now can wait for an attribute value to
appear on a replica.

Fixed timeout handling caused by bad rounding and comparison. For small
timeouts, the actual time was rounded down. For example for 60 seconds
timeout and fast replica, the query accumulated to about 0.45 seconds
plus 60 seconds sleep. 60.45 is large enough to terminate the loop
"while int(time.time()) < timeout", but not large enough to trigger the
exception in "if int(time.time()) > timeout", because int(60.65) == 60.

Fixes: https://pagure.io/freeipa/issue/7595

PR #2059, opened 2018-06-22
[Backport][ipa-4-6] Fix replication races in Dogtag admin code

tiran:
Manual backport of PR #2051

PR fixes two related issues:

# Fix replication races in Dogtag admin code

DogtagInstance.setup_admin and related methods have multiple LDAP
replication race conditions. The bugs can cause parallel
ipa-replica-install to fail.

The code from __add_admin_to_group() has been changed to use MOD_ADD
ather than search + MOD_REPLACE. The MOD_REPLACE approach can lead to
data loss, when more than one writer changes a group.

setup_admin() now waits until both admin user and group membership have
been replicated to the master peer.

Fixes: https://pagure.io/freeipa/issue/7593

# Improve and fix timeout bug in wait_for_entry()

replication.wait_for_entry() now can wait for an attribute value to
appear on a replica.

Fixed timeout handling caused by bad rounding and comparison. For small
timeouts, the actual time was rounded down. For example for 60 seconds
timeout and fast replica, the query accumulated to about 0.45 seconds
plus 60 seconds sleep. 60.45 is large enough to terminate the loop
"while int(time.time()) < timeout", but not large enough to trigger the
exception in "if int(time.time()) > timeout", because int(60.65) == 60.

Fixes: https://pagure.io/freeipa/issue/7595

### DIFF
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -149,6 +149,8 @@ DEFAULT_CONFIG = (
     ('startup_timeout', 300),
     # How long http connection should wait for reply [seconds].
     ('http_timeout', 30),
+    # How long to wait for an entry to appear on a replica
+    ('replication_wait_timeout', 300),
 
     # Web Application mount points
     ('mount_ipa', '/ipa/'),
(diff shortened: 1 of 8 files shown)


tiran:
PR was ACKed manually because this is backport of PR #2051.

tiran:
ipa-4-6:

* 735763769855ef01cf34f355a57f35e058d91479 Improve and fix timeout bug in wait_for_entry()
* 39889882a7b0e5a3dc10208455728e9a678af606 Use common replication wait timeout of 5min
* 763951c7fd29a8032778b41018cf15d07a88c43b Fix replication races in Dogtag admin code

PR #2060, opened 2018-06-22
WebUI build update: use NodeJS instead of Rhino, use system UglifyJS, update Dojo

pvoborni:
## WebUI build: replace uglifyjs with system package

UgligyJS is packaged in Fedora and other OSes it is no longer required
to carry our own version. This will lower the maintenance burden - the
code doesn't need to be updated and it is less code to have in repo.

On some configuration usage of the bundled UglifyJS 1 produces
"JavaScript throw: java.lang.StackOverflowError" exception. Usage of more
recent version should fix it.

##  WebUI build: use NodeJS instead of Rhino

Rhino is no longer mainstream, nor is Nashorn. In addition it is quite
slow (about 10x) in comparison to NodeJS. Over the years NodeJS became
common part of OSes, thus one of the original reasons why use Rhino
went away.

The change in 01-Make-dojo-builder-buildable-by-itself.patch fixes
an incorrect change of the patch (it was not processing input options
well).

Removing configRhino.js and adding configNode.js are prerequisites
for Dojo Builder. These files are copied from Dojo project. Without
them it doesn̈́'t run. In long run, it would be good to replace Dojo
builder with something else but that is outside of this commit/PR.

Last changes are preparation for update to latest stable version of
Dojo 1. The updated Dojo and Dojo builder are in subsequent commit.

## Update Dojo and Dojo builder to 1.13.0

This is a result of the previous commits. Building the Dojo builder
was bit more complex as it was:
1. patched Dojo sources
2. built from Dojo builder sources.
3. moved to it's location in FreeIPA project
4. built by util/make-builder.sh (does minimazation and replaces
   itself)

Then Dojo layer is built by just:
1. util/make-dojo.sh

This process was documented some time ago at:

https://www.freeipa.org/page/V3/WebUI_build


pvoborni:
Btw, I haven't tested Web UI much, and there is a minor potential that the Dojo update can break something.

rcritten:
Should this include a temporary commit to run the UI tests?

netoarmando:
I think you also need to update [`prepare-dojo.sh#L79`](https://github.com/netoarmando/freeipa/blob/master/install/ui/util/prepare-dojo.sh#L79) script to point to the new branch `1.13`.

netoarmando:
@pvoborni added temporary tests to a PR against his fork: https://github.com/pvoborni/freeipa/pull/24

I also tried to built using nodejs and Rhino, the new builder is much faster, but the minified js is a little bit bigger (irrelevant but interesting). I haven't tested the UI, I'm assuming the automated tests are good enough.

netoarmando:
Failing test has been fixed in #2076.

netoarmando:
@pvoborni's initial set of tests (https://github.com/pvoborni/freeipa/pull/24) didn't include groups `a_to_d` and `s_to_v`.  
I've included them, and now some tests are failing.

pvoborni:
I did not run the with all Web UI tests in the separate PR because those tests were failing even without this PR. Seems that one batch was fixed. If failures in test_webui_s_to_v match the ones which we get in nightlies (i.e. no regression here) then we can remove the temp commit and push this.

netoarmando:
`fedora-28/test_webui` tests are passing on Nightly, so I'm going to rebase this to see if we can have all checks passing, but I think we are good either way.

tiran:
master:

* ca7cece13303b3610a71b9e2a0adcb462006382b WebUI build: replace uglifyjs with system package
* df95ba598313b8b03a492a3c89af151ef0faa08d WebUI build: use NodeJS instead of Rhino
* 19c3f173d9c824eaa099b07d2ea341acf42f12f2 Update Dojo and Dojo builder to 1.13.0

PR #2067, opened 2018-06-25
Prevent the creation on users and groups with numeric characters only

netoarmando:
Update regular expression validator to prevent user and group creation.

Issue: https://pagure.io/freeipa/issue/7572

The expression updated by this PR was changed only twice [since the beginging on project](https://github.com/freeipa/freeipa/blob/d1691eee88c5462ef1d015617fd5b65eec0319b9/ipalib/plugins/baseuser.py#L212):
- [first change](https://github.com/freeipa/freeipa/commit/37200806118d39ef8afe84ad5887a294d54e2659) removed the lenght limit
- [second change](https://github.com/freeipa/freeipa/commit/8f8e3d008f1de91337a83ea6d271662432209767) unified the expression used by user and group in the `constants.py` file

This PR doesn't prevent the creation of user or groups starting with a number. Related info: https://access.redhat.com/solutions/3103631

Also, I haven't updated the error message, should I do it? If so, possible wording:
```
may only include letters, numbers, _, -, ., $ and must contain a letter
```

It seems that _systemd_ has a [stricter set of rules](https://github.com/systemd/systemd/blob/master/src/basic/user-util.c#L569-L581), which I think it could affect FreeIPA in the future:
> - We don't allow any dots (this would break chown syntax which permits dots as user/group name separator)
> - We require that names fit into the appropriate utmp field
> - We don't allow empty user names

### DIFF
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -288,7 +288,9 @@ RENEWAL_REUSE_CA_NAME = 'dogtag-ipa-ca-renew-agent-reuse'
 CA_DBUS_TIMEOUT = 120
 
 # regexp definitions
-PATTERN_GROUPUSER_NAME = '^[a-zA-Z0-9_.][a-zA-Z0-9_.-]*[a-zA-Z0-9_.$-]?$'
+PATTERN_GROUPUSER_NAME = (
+    '(?!^[0-9]+$)^[a-zA-Z0-9_.][a-zA-Z0-9_.-]*[a-zA-Z0-9_.$-]?$'
+)
 
 # Kerberos Anonymous principal name
 ANON_USER = 'WELLKNOWN/ANONYMOUS'
--- a/ipatests/test_xmlrpc/test_group_plugin.py
+++ b/ipatests/test_xmlrpc/test_group_plugin.py
@@ -169,6 +169,26 @@ class TestGroup(XMLRPC_test):
                 error=u'may only include letters, numbers, _, -, . and $')):
             command()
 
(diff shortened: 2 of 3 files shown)


rcritten:
shadow-utils (which provides newuser and newgroup) has a name checker for users and groups which imposes its own length requirements plus (as a comment header):

```
User/group names must match gnu e-regex:
    [a-zA-Z0-9_.][a-zA-Z0-9_.-]{0,30}[a-zA-Z0-9_.$-]?

as a non-POSIX, extension, allow "$" as the last char for sake of Samba 3.x "add machine script"

Also do not allow fully numeric names or just "." or "..".
```

shadow-utils does the all numeric-check separately.

It looks to me like this regex matches the functionality of the shadow-utils name checker. 

Fedora 28 defines the max user and group name length as 32.

I wonder if we should also revisit the lengths. I forget why we chose to allow such long names. Length may be beyond the scope of this change given AD complexity (idviews). I'd be curious what @abbra thinks.

abbra:
Please do not limit name length for both groups and users. We are seeing IPA used for non-POSIX use cases too and there we have no need for limitations like that. 

Also, POSIX defines LOGIN_NAME_MAX for the maximum length of a login name and it is 256 on Fedora 28. There is also a minimum length for the login name which is _POSIX_LOGIN_NAME_MAX, set to 9 (including terminating null) in POSIX.

Having a match between shadow-utils and IPA regex is a good idea.

rcritten:
Ok, LGTM. The regex is the shadow-utils + not allowing all digits which is effectively what it does.

PR #2071, opened 2018-06-26
ipaserver/dcerpc.py: handle indirect topology conflicts

abbra:
When AD forest A has a trust with a forest B that claims ownership of a domain name (TLN) owned by an IPA forest, we need to build exclusion record for that specific TLN, not our domain name.

Use realmdomains to find a correct exclusion entry to build, not just using our own domain name.

Fixes: https://pagure.io/freeipa/issue/7370

### DIFF
--- a/ipaserver/dcerpc.py
+++ b/ipaserver/dcerpc.py
@@ -1137,22 +1137,41 @@ class TrustDomainInstance(object):
 
                 # Copy over the entries, extend with TLN exclusion
                 entries = []
+                is_our_record = False
                 for e in dominfo.entries:
                     e1 = lsa.ForestTrustRecord()
                     e1.type = e.type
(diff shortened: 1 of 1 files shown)


abbra:
Note that the original bug https://bugzilla.redhat.com/show_bug.cgi?id=1533803 has some unrelated misconfiguration issue which prevented tests to perform properly. However, its log actually showed this bug which I didn't notice originally due to that unrelated misconfiguration.

abbra:
@netoarmando thanks, I've updated the code to follow your suggestion and added few more comments.

netoarmando:
master:

* 81f36df7acdccac9b66d2e10adf9a04f8fb7fadd ipaserver/dcerpc.py: handle indirect topology conflicts

PR #2072, opened 2018-06-26
Fix login page translations

stanislavlevin:
Now JSON service is not available without authentication against IPA. But some of Web UI pages expect translations before or without Login process. 

1. i18n endpoint was added. This one serves requests for translations only.
2. translations JS module was implemented. This module is used to get translated messages via JSON request in a synchronous manner. To ensure translatability i18n messages should be initialized before any other JS code interacted with user is run.

Additionally, several, related to this issue parts of WebUI are fixed.

https://pagure.io/freeipa/issue/7559
Review is a highly appreciated.


install/ui/src/freeipa/translations.js
```diff
@@ -0,0 +1,82 @@
...
+    var result = false;
+
+    var jsondata = {
+        method: "i18n_messages",
+	params: [ [], { "version": window.ipa_loader.api_version } ]
+    };
+
+    var request = {
+        method: 'POST',
+        url: '/ipa/i18n_messages',
```
abbra: Please move the URL to config.js like any other URL so they can be customized

stanislavlevin: Thank you.

install/ui/src/freeipa/translations.js
```diff
@@ -0,0 +1,82 @@
+/*  Authors:
+ *    Petr Vobornik <pvoborni@redhat.com>
+ *
+ * Copyright (C) 2013 Red Hat
+ * see file 'COPYING' for use and warranty information
```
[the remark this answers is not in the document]
stanislavlevin: Thank you.

abbra:
Travis-CI broke due to something in phantomjs. I ran this PR locally and it succeeded. @stanislavlevin, once you'd address the other comments, I'll re-run the tests.

stanislavlevin:
I faced the same problem on local machine:
```
[D] ["phantomjs","onResourceRequested",{"headers":[{"name":"Accept","value":"application/json, text/javascript, */*; q=0.01"},{"name":"Origin","value":"file://"},{"name":"X-Requested-With","value":"XMLHttpRequest"}],"id":68,"method":"GET","time":"2018-06-26T21:15:46.767Z","url":"file:///ipa/i18n_messages"}]
[D] ["phantomjs","error.onError","NETWORK_ERR: XMLHttpRequest Exception 101: A network error occurred in synchronous requests."
```
I will check later.

stanislavlevin:
Something tells me that there should be more tests.

rcritten:
You could probably add a test somewhere in ipatests/test_xmlrpc to do some basic endpoint testing.

stanislavlevin:
Thank you ALL for a patience :-)
Now i'm familiar with "ipa-docker-test-runner". Awesome experience for me.

install/share/ipa.conf.template
```diff
@@ -100,6 +100,13 @@ WSGIScriptReloading Off
 # Target for login with internal connections
 Alias /ipa/session/cookie "/usr/share/ipa/gssapi.login"
 
+# Turn off Apache authentication for i18n messages
+<Location "/ipa/i18n_messages">
+  <RequireAll>
+    Require method POST
```
tiran: Why did you choose this way to change authentication and why only ``POST``? There are many more HTTP verbs that can modify data, e.g. ``DELETE`` or ``PATCH``.

ipaserver/rpcserver.py
```diff
@@ -784,6 +785,46 @@ def marshal(self, result, error, _id=None,
...
+        status = '401 Unauthorized'
+        headers = []
+        response = b''
+
+        logger.debug('jsonserver_i18n_messages: %s need login', status)
+        start_response(status, headers)
+        return [response]
+
+    def __call__(self, environ, start_response):
+        '''
```
tiran: We use ``"""`` for doc strings. Please write a doc string or remove it.

ipaserver/rpcserver.py
```diff
@@ -784,6 +785,46 @@ def marshal(self, result, error, _id=None,
...
 
+class jsonserver_i18n_messages(jsonserver):
+    """
+    JSON RPC server for i18n messages only.
+    """
+
+    key = '/i18n_messages'
+
+    def __init__(self, api):
+        super(jsonserver_i18n_messages, self).__init__(api)
```
tiran: You don't have to provide a super call if you just call the super method without additional call.

ipaserver/rpcserver.py
```diff
@@ -784,6 +785,46 @@ def marshal(self, result, error, _id=None,
...
+
+    def _on_finalize(self):
+        super(jsonserver_i18n_messages, self)._on_finalize()
```
tiran: Same here

ipaserver/rpcserver.py
```diff
@@ -784,6 +785,46 @@ def marshal(self, result, error, _id=None,
...
+        logger.debug('jsonserver_i18n_messages: %s need login', status)
+        start_response(status, headers)
+        return [response]
+
+    def __call__(self, environ, start_response):
+        '''
+        '''
+
+        logger.debug('WSGI jsonserver_i18n_messages.__call__:')
+        data = read_input(environ)
```
tiran: Related to your Apache configuration: You should refuse all but ``GET`` requests here.

ipatests/test_ipaserver/httptest.py
```diff
@@ -41,13 +42,15 @@ def send_request(self, method='POST', params=None):
         :param key When not None, overrides default app_uri
         """
         if params is not None:
+            if self.content_type == 'application/x-www-form-urlencoded':
             # urlencode *can* take two arguments
```
tiran: Indent the comment, too.

install/share/ipa.conf.template
stanislavlevin: If you mean why not:
```apache
<Location "/ipa/i18n_messages">
  Satisfy Any 
  Order Deny,Allow
  Allow from all 
</Location>
```
From docs (https://httpd.apache.org/docs/2.4/howto/access.html):
 ```apache
The Allow, Deny, and Order directives, provided by mod_access_compat, are deprecated and will go away in a future version. You should avoid using them, and avoid outdated tutorials recommending their use.
```
For this case the only allowed HTTP method is "POST" ( for strict configuration), all the others are refused. The goal of this endpoint is process of "i18n_messages" JSON requests via "POST".

File: ipaserver/rpcserver.py (hunk 5 above)
Comment (stanislavlevin) [reply ipaserver/rpcserver.py]: Perhaps, i misunderstand your thoughts. Excuse me, please.
super() is used for indirect reference as well as delegating method calls to some base class in the instance’s ancestor tree in case of further inheritance.

Comment (tiran) [reply ipaserver/rpcserver.py]: Since the ``__init__`` method doesn't perform additional work, you can drop the entire ``__init__`` method from the class. Python automatically calls the method from the parent class.

File: ipaserver/rpcserver.py (hunk 7 above)
Comment (stanislavlevin) [reply ipaserver/rpcserver.py]: Apache returns "403 Forbidden" as i use config access rules.

File: install/share/ipa.conf.template (hunk 3 above)
Comment (stanislavlevin) [reply install/share/ipa.conf.template]: I've changed apache config to allow all unconditionally.
```apache
<Location "/ipa/i18n_messages">
   Require all granted
</Location>
```

ipatests/test_ipaserver/httptest.py
stanislavlevin: Done.

ipaserver/rpcserver.py
stanislavlevin: Apache POST require rule was removed.

stanislavlevin:
Waiting for review.

install/ui/src/freeipa/ipa.js
```diff
@@ -76,13 +76,6 @@ var IPA = function () {
         processData: false
     };
 
-    /**
-     * i18n messages
-     * @deprecated
-     * @property {Object}
-     */
-    that.messages = {};
```
pvoborni: The patch set is touching basically all usages of `IPA.messages` except the ones in tests. This makes it good to finish the original idea and remove `IPA.messages` completely.  It will allow to not required `ipa` module in `translation` module and thus break a circular dependency and also make the codebase bit more cleaner.

stanislavlevin: I completely agree with you. Last "messages" have been removed.

abbra:
I built and ran this PR. Everything works as promised. ;) ACK from my side.

rcritten:
master:

* ab0835f91f83e295b6ef2875af6470056acdaf76 Add endpoint for serving i18n requests
* 86b57236c07e72907cda92c09a219b86200dff09 Disable authentication to endpoint for serving i18n requests
* de58b80891aab54ef0439449295846b7470724e6 Implement "translations" AMD
* 9492fb7f866c2e8cab269c4131d1a67ff6841af2 Add dependency to "translations" module
* c0c6b21ba19da4b0c609868c2f95433077341ef7 Stop fetching translations at metadata phase
* 5d8fde0ac1a43c8f3dbc53b44d69f3663a8b36fb Fix translations at LoginScreen widget
* 2a81ec3b94712e822f7d3a5452974626c4d5c42c Fix translations at login plugin
* 6bc37150459c779b46ded736a69637188fa95dd7 Fix translations at load_page plugin
* 7f9f59bae2a362ce945c49ad8342393b7a5c024f Fix translation of profile menu
* c4467aaeea6f255d181e286973bff2c1c6dc7280 Add static JSON dump of i18n_messages request
* b8607e24d610a240a2501fdfe0242233fa00e2e2 Fix Web UI 'get_entity_param' test
* 0dace623ab1c96b7af8fd64028c9e37d2b10f788 Add support for JSON request in HTTP test class
* 0908e80d62146c79c4312effb8e285c13294ec4b Add support for Accept-Language in HTTP test class
* f49fac7bda8150aee2086be9afdbe4eb81c3f18a Add tests for "i18n_messages" end point
* bb67eea1054c9fe80b8bba67c7dc2c0991a1b991 Fix Web UI "details lifecycle" test
* 4b2af2570aee6b108504e14330d61bf65a52905c Stop usage of "IPA.messages" in Web UI "utils" tests

PR #2073, opened 2018-06-27
Check if user permssions and umask 0022 is set when executing ipa-restore

mrizwan93:
This test checks if the access rights for user/group
is set to 644 on /var/lib/dirsrv/slapd-TESTRELM-TEST/ldif/*
and umask 0022 set while restoring.

related ticket: https://pagure.io/freeipa/issue/6844

### DIFF
--- a/ipatests/test_integration/test_backup_and_restore.py
+++ b/ipatests/test_integration/test_backup_and_restore.py
@@ -494,7 +494,7 @@ class TestBackupAndRestoreWithReplica(IntegrationTest):
         check_replication(self.master, replica, "testuser1")
 
 
-class TestUserrootFilesOwnership(IntegrationTest):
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -534,11 +540,12 @@ def test_userroot_ldif_files_ownership(self):
         cmd = self.master.run_command(arg)
         assert 'dirsrvdirsrv' in cmd.stdout_text
 
+        # also check of access rights are set to 644
         arg = ['stat',
-               '-c', '%U%G',
-               '/var/lib/dirsrv/slapd-' + dashed_domain + '/ldif/']
+               '-c', '%U%G%a',
```
tiran: Could you please change this to ``%U:%G:%a`` to make the check more readable?

PR #2076, opened 2018-06-27
ui_tests - stabilization fixes + test_config fix

Rezney:
This patch aims to fix the following tests which seems to be quite
    unstable recently:
    
    test_user::test_actions - closing notification and moving to element
    to have screenshot of current place.
    
    test_user::certificates - add wait() / close_notification
    
    test_config::test_size_limits - add wait() to fill_input()

### DIFF
--- a/ipatests/test_webui/test_config.py
+++ b/ipatests/test_webui/test_config.py
@@ -145,8 +145,11 @@ class test_config(UI_driver):
                                    'Must be an integer')
 
         # test field with negative value
-        self.assert_field_negative(size_limit_s, '-10',
-                                   'Minimum value is -1')
+        self.assert_field_negative(
+            size_limit_s, '-10',
+            "invalid 'searchrecordslimit': must be at least 10",
+            dialog=True,
+        )
 
         # test field with space
         self.assert_field_negative(size_limit_s, ' 11',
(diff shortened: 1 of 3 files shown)


rcritten:
FWIW the size limit tests are likely failing due to a change in the plugin that dropped the minimum value of -1 and effectively reset it to 10 via a validator function. The allowed values are -1, 0 and >= 10

netoarmando:
@Rezney I have a fix for the tests, should I commit on your branch? 

Like @rcritten said, the (now) expected behavior is a dialog containing the message: 
```
invalid 'searchrecordslimit': must be at least 10
```
Not the most user-friendly message, but the fix is basically:
```python
        # test field with negative value
        self.assert_field_negative(
            size_limit_s, '-10',
            "invalid 'searchrecordslimit': must be at least 10",
            dialog=True,
        )
``` 
Pagure ticket tracking this regression: https://pagure.io/freeipa/issue/7606

Rezney:
@netoarmando Thanks for the info. Yep I am already aware of this. Please feel free to commit to this branch.

Rezney:
Tests are passing. Dropping TEMP commit and original changes to test_config.

pavelpicka:
code looks good and tests are passing - ACK

tiran:
master:

* e140d198eaf6982a966b2d0bf3edbd091142a894 ui_tests: stabilization fixes
* 79391ad8e1e15af14b86167fb110c139d291a0a0 ui_tests: fix test_config::test_size_limits

PR #1956, opened 2018-05-23
Don't try to set Kerberos extradata when there is no principal

rcritten:
This was causing ns-slapd to segfault in the password plugin.

https://pagure.io/freeipa/issue/7561

Signed-off-by: Rob Crittenden <rcritten@redhat.com>

PR #2082, opened 2018-06-28
[Backport][ipa-4-6] Rename test class for testing simple commands, add test

rcritten:
Manual backport of the test from PR #1956 . Only the code was pushed to the ipa-4-6 branch.

The consensus in the review was that the name test_commands was
more generic than test_ipa_cli.

Add a test to change the password for sysaccount users using
using ldappasswd to confirm that a segfault fix does not regress.


rcritten:
ipa-4-6:

* 6652eb088e87cd3f74bd58be3ccda89ec0a46664 Rename test class for testing simple commands, add test

PR #2084, opened 2018-06-29
Catch ACIError instead of invalid credentials

tiran:
ipaldap's LDAPClient client turns INVALID_CREDENTIAL error into
ACIError. Catch the ACIError and wait until the user has been
replicated.

Apparently no manual or automated test ran into the timeout during
testing.

Fixes: Fixes: https://pagure.io/freeipa/issue/7593

### DIFF
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -472,7 +472,8 @@ class DogtagInstance(service.Service):
             time.sleep(1)
             try:
                 master_conn.simple_bind(self.admin_dn, self.admin_password)
-            except ldap.INVALID_CREDENTIALS:
+            except errors.ACIError:
+                # user not replicated yet
                 pass
             else:
                 logger.debug("Successfully logged in as %s", self.admin_dn)


abbra:
There are few more places where invalid credentials are expected to raise, they also need changes to ACIError:
```
$ git grep ldap.INVALID_CREDENTIALS
daemons/dnssec/ipa-dnskeysyncd:    except ldap.INVALID_CREDENTIALS as e:
ipapython/ipaldap.py:        except ldap.INVALID_CREDENTIALS:
ipaserver/install/dogtaginstance.py:            except ldap.INVALID_CREDENTIALS:
ipaserver/install/installutils.py:    if isinstance(error, ldap.INVALID_CREDENTIALS):
ipaserver/install/ldapupdate.py:    except ldap.INVALID_CREDENTIALS:
```

tiran:
* ``daemons/dnssec/ipa-dnskeysyncd`` uses a raw LDAP connection
* ``ipapython/ipaldap.py:        except ldap.INVALID_CREDENTIALS:`` is the line that maps the raw LDAP exception to ACIError
* ``ipaserver/install/dogtaginstance.py`` is addressed by this patch
* ``ipaserver/install/installutils.py`` needs investigation
* ``ipaserver/install/ldapupdate.py`` looks buggy, but should be addressed in another PR.

abbra:
The last one (`ldapupdate.py`) uses `ldap.LOCAL_ERROR` which is mapped to `ACIError` as well. The change happened at df4ed77962e906b2456c99e57084314000768569 but not all places were updated since then.

tiran:
@abbra Please open a separate issue for the problem. I don't want to mix the other fixes with this PR, because this PR has to land in 4.5 and 4.6 ASAP.

abbra:
I added https://pagure.io/freeipa/issue/7610 to track the latter.

tiran:
Thanks!

By the way, good catch :+1:

abbra:
Tests completed. ACK.

tiran:
master:

* 52cdd213b4320ef463c4d0053b436511e3f53709 Catch ACIError instead of invalid credentials

PR #2090, opened 2018-07-01
[RFE] DNS package check should be called earlier in installation routine

amitkumar50:
In case the integrated DNS service should be used for IdM, either by using the
non-interactive '--setup-dns' option or by answering 'yes' to the interactive
'Do you want to configure integrated DNS (BIND)?' question, the check whether
the 'ipa-server-dns' package is installed, only comes after a lot of other information
have been evaluated ([server|dns|realm] name, DM and admin password).

This PR check, whether the 'ipa-server-dns' package is installed or not,
earlier in the installation routine.

Resolves: https://pagure.io/freeipa/issue/7577

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -471,6 +471,15 @@ def install_check(installer):
 
     domain_name = domain_name.lower()
 
+    if options.setup_dns:
+        api.env.domain = domain_name
+        dns.install_check(False, api, False, options, host_name)
+        ip_addresses = dns.ip_addresses
+    else:
+        ip_addresses = get_server_ip_address(host_name,
+                                             not installer.interactive, False,
+                                             options.ip_addresses)
+
     if not options.realm_name:
         realm_name = read_realm_name(domain_name, not installer.interactive)
         logger.debug("read realm_name: %s\n", realm_name)
@@ -634,14 +643,6 @@ def install_check(installer):
         ca.install_check(False, None, options)
     if options.setup_kra:
         kra.install_check(api, None, options)
-
-    if options.setup_dns:
-        dns.install_check(False, api, False, options, host_name)
-        ip_addresses = dns.ip_addresses
-    else:
-        ip_addresses = get_server_ip_address(host_name,
-                                             not installer.interactive, False,
-                                             options.ip_addresses)
 
         # check addresses here, dns module is doing own check
         no_matching_interface_for_ip_address_warning(ip_addresses)


amitkumar50:
Will be checking travis build log

ipaserver/install/server/install.py
```diff
@@ -635,16 +643,8 @@ def install_check(installer):
...
-        ip_addresses = dns.ip_addresses
-    else:
-        ip_addresses = get_server_ip_address(host_name,
-                                             not installer.interactive, False,
-                                             options.ip_addresses)
-
-        # check addresses here, dns module is doing own check
-        no_matching_interface_for_ip_address_warning(ip_addresses)
+    # check addresses here, dns module is doing own check
+    no_matching_interface_for_ip_address_warning(ip_addresses)
```
tiran: The function call must stay in the second block.

rcritten:
The installation fails.

07-01 11:48 ipadocker.command INFO     Executing command: ipa-server-install -U --domain ipa.test --realm IPA.TEST -p Secret123 -a Secret123 --setup-dns --setup-kra --auto-forwarders
Checking DNS domain ipa.test, please wait ...
'Env' object has no attribute 'domain'

ipaserver/install/server/install.py
```diff
@@ -471,6 +471,15 @@ def install_check(installer):
 
     domain_name = domain_name.lower()
 
+    if options.setup_dns:
+        api.env.domain = domain_name
```
tiran: Why is this assignment necessary?

amitkumar50: inside install_check domain was not found inside api. api.env.domain.
So provided domain_name explicitly, and tested this does not break things.

tiran: I don't like this approach. It feels more like a dirty, fragile hack than a properly solution. Either the function shouldn't assume a fully configured api object or the ``api.env`` should be finalized.

amitkumar50: Thankyou @tiran I was also thinking the same.Will change, test and do the change.

rcritten: This assignment works only because the api hasn't been finalized. It makes me nervous. It makes me wonder if a separate pkg_check() is needed instead of that happening in install_check().

tiran: me, too.

I'm strongly against this hack. The ``dns.install_check()`` function does a lot of things. Some of the options also depend on a fully working and initialized api object. This needs more engineering work. Perhaps the check can be split into more steps?

abbra:
@amitkumar50 are you going to address the comments or should we close this PR?

tiran:
I have created a new PR https://github.com/freeipa/freeipa/pull/4696 that implements the check in a different way.

PR #1643, opened 2018-03-01
ipa-4-5: python2-pyasn1 workaround

Rezney:
initialize ASN.1 object by hand to make it compatible with ancient pyasn1

PR #2092, opened 2018-07-03
[IPA-4-6] - initialize ASN.1 object by hand.

Rezney:
Same workaround as for 7.5
https://github.com/freeipa/freeipa/pull/1643


Rezney:
Closing as RHEL specific. Will add the patch downstream.

PR #2093, opened 2018-07-03
ipa-client-install: Update how comments are added by ipachangeconf

netoarmando:
Due to how 'openldap-client' parses its configuration files this patch changes how comments are added, moving them to the line above instead of appending to the same line.

Issue: https://pagure.io/freeipa/issue/5202

### DIFF
--- a/ipatests/test_install/test_changeconf.py
+++ b/ipatests/test_install/test_changeconf.py
@@ -1,0 +1,50 @@
+# Copyright (C) 2018  FreeIPA Contributors see COPYING for license
+
+from __future__ import absolute_import
+
+import pytest
+from ipaclient.install.ipachangeconf import IPAChangeConf
+
+
+@pytest.fixture(scope='function')
+def config_filename(tmpdir):
+    filename = tmpdir.mkdir('data').join('config_file.conf')
+    filename.write('SOME_CONF /some/user/defined/path\n')
+    return filename
+
+
+def test_addifnotset_action(config_filename):
+    """Test if addifnotset action adds a comment about the modified conf.
+
+    IPA doesn't want to break existing configuration, if a value already exists
+    it adds a comment to the modified setting and a note about that on the line
+    above.
+
+    New settings will be added without any note.
+    """
+    ipa_conf = IPAChangeConf('IPA Installer Test')
+    ipa_conf.setOptionAssignment(' ')
+
+    opts = [
+        {
+            'action': 'addifnotset',
+            'name': 'SOME_CONF',
+            'type': 'option',
+            'value': '/path/defined/by/ipa',
+        },
+        {
+            'action': 'addifnotset',
+            'name': 'NEW_CONF',
+            'type': 'option',
+            'value': '/path/to/somewhere',
+        },
+    ]
+
+    ipa_conf.changeConf(str(config_filename), opts)
+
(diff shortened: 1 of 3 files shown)


netoarmando:
@rcritten thanks for adding some explanation to this patch! Something I failed to do.

> I added a separate test after yours to add a new option, VAR_NAME2, and it was not preceded by a comment "modified by IPA". Maybe my test was bad but I think an additional test like this is needed in any case.

The comment "modified by IPA" proceeds only variables that IPA would override. However, I can test the scenario you've mentioned to assert that the result wouldn't include the comment.

ipatests/test_install/test_changeconf.py
```diff
@@ -0,0 +1,50 @@
+# Copyright (C) 2018  FreeIPA Contributors see COPYING for license
+
+from __future__ import absolute_import
+
+import pytest
+from ipaclient.install.ipachangeconf import IPAChangeConf
+
+
+@pytest.fixture(scope='session')
+def config_file(tmpdir_factory):
```
tiran: ``tmpdir_factory`` was introduced in pytest 2.8, but RHEL 7.5 has 2.7.0-2.el7. You need to use another way to create a temporary file, e.g. ``tmpdir`` fixture with a ``function`` scope fixture.

ipatests/test_install/test_changeconf.py
```diff
@@ -0,0 +1,50 @@
...
+    file = tmpdir_factory.mktemp('data').join('config_file.conf')
```
tiran: ``file`` is the name of a Python 2 builtin. In other places, we use ``filename`` to refer to a file name.

ipatests/test_install/test_changeconf.py
```diff
@@ -0,0 +1,50 @@
...
+            'name': 'NEW_CONF',
+            'type': 'option',
+            'value': '/path/to/somewhere',
+        },
+    ]
+
+    ipa_conf.changeConf(str(config_file), opts)
+
+    assert config_file.readlines() == [
+        '#SOME_CONF modified by IPA\n',
```
tiran: ``#text`` is hard to read. Could you change the code to emit ``# text`` instead?

ipatests/test_install/test_changeconf.py
netoarmando: Nice observation, would pr-ci nightly catch this eventually?

ipatests/test_install/test_changeconf.py
netoarmando: My bad..

ipatests/test_install/test_changeconf.py
netoarmando: Agreed.

ipatests/test_install/test_changeconf.py
tiran: I don't think so. QE will eventually run into this and complain.

netoarmando:
@tiran  I've changed to `# text` format only in `# SOME_VAR modified by IPA` line.
I would update every comment to that format, I haven't done it because I was afraid it could break some script executed by administrators out there.

tiran:
master:

* 53c549664795f54dbc337e45aeba84fbff843109 ipa-client-install: Update how comments are added by ipachangeconf

PR #2097, opened 2018-07-04
[py37] Import ABCs from collections.abc

tiran:
Python 3 has moved all collection abstract base classes to
collections.abc. Python 3.7 started to deprecate the old aliases.

The whole import block needs to be protected with import-error and
no-name-in-module, because Python 2 doesn't have collections.abc module and
collections.abc.Mapping, while Python 3 doesn't have collections.Mapping.

Fixes: https://pagure.io/freeipa/issue/7609


tiran:
master:

* 198a2c61129675b928bb7a970d5fc4c63a032456 Import ABCs from collections.abc

PR #2098, opened 2018-07-04
ipa_tests: ipa-replica-prepare stuck on user input

Rezney:
TestOldReplicaWorksAfterDomainUpgrade is getting stuck while
    running "ipa-replica-prepare" as it is asking for user input:
    "Do you want to search for missing reverse zones?". Adding
    "--auto-reverse" in order to continue.

### DIFF
--- a/ipatests/pytest_plugins/integration/tasks.py
+++ b/ipatests/pytest_plugins/integration/tasks.py
@@ -379,7 +379,7 @@ def replica_prepare(master, replica, extra_args=(),
             '-p', replica.config.dirman_password,
             replica.hostname]
     if master_authoritative_for_client_domain(master, replica):
-        args.extend(['--ip-address', replica.ip])
+        args.extend(['--ip-address', replica.ip, '--auto-reverse'])
     args.extend(extra_args)
     result = master.run_command(args, raiseonerr=raiseonerr,
                                 stdin_text=stdin_text)


tiran:
@Rezney Should tests use ``--unattended`` by default, too?

Rezney:
This is flow before "ipa-replica-prepare":

```
    if master_authoritative_for_client_domain(master, replica):
        args.extend(['--ip-address', replica.ip, '--auto-reverse'])
```

This was introduced in [42d36442760](https://github.com/freeipa/freeipa/commit/42d364427606e39486645e4064ca16940b2f8837) 

In the test we get returncode 0 as we have both dnszone and reverse-zone. Though the function below does not seem to be testing for reverse zone:
```
def master_authoritative_for_client_domain(master, client):
    zone = ".".join(client.hostname.split('.')[1:])
    result = master.run_command(["ipa", "dnszone-show", zone],
                                raiseonerr=False)
    return result.returncode == 0
```

But later it looks like "ipa-replica-prepare" cannot find reverse zone and is offering a search:

GDB from ipa-replica-prepare process:

```
(gdb) py-bt
Traceback (most recent call first):
  <built-in method readinto of _io.FileIO object at remote 0x7f9fa08e13a8>
  <built-in method read1 of _io.BufferedReader object at remote 0x7f9fa09a39e8>
  <built-in method readline of _io.TextIOWrapper object at remote 0x7f9fa0950558>
  <built-in method input of module object at remote 0x7f9fa09a2638>
  File "/usr/lib/python3.6/site-packages/ipapython/ipautil.py", line 970, in user_input
    ret = input("%s [%s]: " % (prompt, choice))
  File "/usr/lib/python3.6/site-packages/ipaserver/install/bindinstance.py", line 91, in create_reverse
    True
  File "/usr/lib/python3.6/site-packages/ipaserver/install/bindinstance.py", line 472, in check_reverse_zones
    options.no_reverse = not create_reverse()
  File "/usr/lib/python3.6/site-packages/ipaserver/install/ipa_replica_prepare.py", line 272, in ask_for_options
    True)
  File "/usr/lib/python3.6/site-packages/ipapython/admintool.py", line 177, in execute
    self.ask_for_options()
  File "/usr/lib/python3.6/site-packages/ipapython/admintool.py", line 156, in main
    return command.execute()
  File "/usr/lib/python3.6/site-packages/ipapython/admintool.py", line 132, in run_cli
    sys.exit(cls.main(sys.argv))
  (frame information optimized out)
(gdb) quit
```

Is not my adding "--auto-reverse' covering any issue here?

Rezney:
@tiran Thanks for checking this PR. We usually do, however "ipa-replica-prepare" does not seem to have any "unattended" option. Could you please also check my previous comment?

tiran:
Your explanation sounds plausible and tests are now passing. Since ``ipa-replica-prepare`` doesn't support non-interactive installation and the problem is now gone, your patch looks fine to me. But I'm not 100% sure there are no additional ramifications.

Rezney:
@tiran Fair enough, could you please also add ACK label then. Cheers...

Rezney:
Added Pagure ticket and got rid of temp-commit.

tiran:
master:

* 417f74868276cf67580d39d82f5bbfe96c83d62c ipa_tests: ipa-replica-prepare stuck on user input

PR #2102, opened 2018-07-05
Delay enablement of services and DNS SRV entries

tiran:
### Query for server role IPA master

server_find and server_role plugin were hiding IPA master role
information. It's now possible to fetch IPA master role information and
to filter by IPA master role, e.g. to ignore servers that have some
services configured but not (yet) enabled.

### Only create DNS SRV records for ready server

When installing multiple replicas in parallel, one replica may create
SRV entries for other replicas, although the replicas aren't fully
installed yet. This may cause some services to connect to a server, that
isn't ready to serve requests.

The DNS IPASystemRecords framework now skips all servers that aren't
ready IPA masters.

### Delay enabling services until end of installer

Service entries in cn=FQDN,cn=masters,cn=ipa,cn=etc are no longer
created as enabled. Instead they are flagged as configuredService. At
the very end of the installer, the service entries are switched from
configured to enabled service.

- SRV records are created at the very end of the installer.
- Dogtag installer only picks fully installed servers
- Certmonger ignores all configured but not yet enabled servers.

Fixes: pagure.io/freeipa/issue/7566

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -901,14 +901,6 @@ def install(installer):
 
     if options.setup_dns:
         dns.install(False, False, options)
-    else:
-        # Create a BIND instance
-        bind = bindinstance.BindInstance(fstore)
-        bind.setup(host_name, ip_addresses, realm_name,
-                   domain_name, (), 'first', (),
-                   zonemgr=options.zonemgr,
-                   no_dnssec_validation=options.no_dnssec_validation)
-        bind.create_file_with_system_records()
 
     if options.setup_adtrust:
         adtrust.install(False, options, fstore, api)
@@ -940,6 +932,16 @@ def install(installer):
         print()
     except Exception:
         raise ScriptError("Configuration of client side components failed!")
+
+    # Enable configured services and update DNS SRV records
+    service.enable_services(host_name)
+    api.Command.dns_update_system_records()
+
+    if not options.setup_dns:
+        # After DNS and AD trust are configured and services are
+        # enabled, create a dummy instance to dump DNS configuration.
+        bind = bindinstance.BindInstance(fstore)
+        bind.create_file_with_system_records()
 
     # Everything installed properly, activate ipa service.
     services.knownservices.ipa.enable()
--- a/ipaserver/install/service.py
+++ b/ipaserver/install/service.py
@@ -27,6 +27,7 @@ import socket
 import time
 import traceback
 import tempfile
+import warnings
 
 import six
 
@@ -60,6 +61,10 @@ SERVICE_LIST = {
     'DNSSEC': ('ods-enforcerd', 100),
     'DNSKeySync': ('ipa-dnskeysyncd', 110),
 }
+
+CONFIGURED_SERVICE = u'configuredService'
+ENABLED_SERVICE = 'enabledService'
+
 
 def print_msg(message, output_fd=sys.stdout):
     logger.debug("%s", message)
@@ -122,7 +127,7 @@ def find_providing_servers(svcname, conn, api):
     """
     dn = DN(('cn', 'masters'), ('cn', 'ipa'), ('cn', 'etc'), api.env.basedn)
     query_filter = conn.make_filter({'objectClass': 'ipaConfigObject',
-                                     'ipaConfigString': 'enabledService',
+                                     'ipaConfigString': ENABLED_SERVICE,
                                      'cn': svcname}, rules='&')
     try:
         entries, _trunc = conn.find_entries(filter=query_filter, base_dn=dn)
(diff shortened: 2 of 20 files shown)


ipaserver/install/server/install.py
```diff
@@ -941,6 +933,18 @@ def install(installer):
...
+    api.Command.dns_update_system_records()
+    if not options.setup_dns:
+        # Create a BIND instance
+        bind = bindinstance.BindInstance(fstore)
+        bind.setup(host_name, ip_addresses, realm_name,
+                   domain_name, (), 'first', (),
+                   zonemgr=options.zonemgr,
+                   no_dnssec_validation=options.no_dnssec_validation)
+        bind.create_file_with_system_records()
+
```
rcritten: I think we need @abbra opinion on this. I'm guessing that IPA DNS needs to be configured before AD trust.

ipaserver/install/service.py
```diff
@@ -231,6 +235,47 @@ def set_service_entry_config(name, fqdn, config_values,
         raise e
 
 
+def enable_services(fqdn):
+    """Change all configured services to enabled server
```
rcritten: nit: I think just drop "server"

ipaserver/install/service.py
```diff
@@ -231,6 +235,47 @@ def set_service_entry_config(name, fqdn, config_values,
...
+
+    Server.ldap_enable() only marks a service as configured. Services
+    are enabled at the very end of installation.
```
rcritten: Can you expand on potential reasons why. I think the main purpose now is so that the services aren't advertised as available until the entire master is up and running.

ipaserver/install/service.py
```diff
@@ -578,7 +623,13 @@ def run_step(message, method):
         self.steps = []
 
     def ldap_enable(self, name, fqdn, dm_password=None, ldap_suffix='',
```
rcritten: Devil's advocate: this is changing the IPA API. Yes, it's internal but we don't know if any 3rd parties rely on it. Should the calls to ldap_enable be changed to a new method ldap_<something else>?

ipaserver/plugins/serverrole.py
```diff
@@ -129,6 +129,11 @@ class server_role_find(Search):
             minvalue=0,
             autofill=False,
         ),
+        Flag(
+            'include_master?',
+            label=_('Include Master'),
```
rcritten: I don't think a Flag needs a label

ipaserver/plugins/serverrole.py
```diff
@@ -151,8 +156,13 @@ def execute(self, *keys, **options):
...
 
-        result = [
-            r for r in role_status if r[u'role_servrole'] != "IPA master"]
+        if options.get('include_master') or role_name == "IPA master":
+            result = role_status
+        else:
+            result = [
+                r for r in role_status
+                if r[u'role_servrole'] != "IPA master"
+            ]
```
rcritten: Is it worth discussing this behavior in the command help?

ipaserver/install/server/install.py
tiran: ``create_file_with_system_records`` just dumps a file to file system.

rcritten: It is the bind.setup() that I'm most concerned about.

abbra: @rcritten yes, if IPA DNS is present, adtrust install will use it to set up DNS entries for `_msdcs`.

tiran:
The PR also fixes a problem with ``--setup-adtrust`` and without ``--setup-dns``. The ``/tmp/ipa.system.records.db`` now contains all AD entries, too.

```
_kerberos-master._tcp.ipa.example. 86400 IN SRV 0 100 88 master.ipa.example.
_kerberos-master._udp.ipa.example. 86400 IN SRV 0 100 88 master.ipa.example.
_kerberos._tcp.Default-First-Site-Name._sites.dc._msdcs.ipa.example. 86400 IN SRV 0 100 88 master.ipa.example.
_kerberos._tcp.dc._msdcs.ipa.example. 86400 IN SRV 0 100 88 master.ipa.example.
_kerberos._tcp.ipa.example. 86400 IN SRV 0 100 88 master.ipa.example.
_kerberos._udp.Default-First-Site-Name._sites.dc._msdcs.ipa.example. 86400 IN SRV 0 100 88 master.ipa.example.
_kerberos._udp.dc._msdcs.ipa.example. 86400 IN SRV 0 100 88 master.ipa.example.
_kerberos._udp.ipa.example. 86400 IN SRV 0 100 88 master.ipa.example.
_kerberos.ipa.example. 86400 IN TXT "IPA.EXAMPLE"
_kpasswd._tcp.ipa.example. 86400 IN SRV 0 100 464 master.ipa.example.
_kpasswd._udp.ipa.example. 86400 IN SRV 0 100 464 master.ipa.example.
_ldap._tcp.Default-First-Site-Name._sites.dc._msdcs.ipa.example. 86400 IN SRV 0 100 389 master.ipa.example.
_ldap._tcp.dc._msdcs.ipa.example. 86400 IN SRV 0 100 389 master.ipa.example.
_ldap._tcp.ipa.example. 86400 IN SRV 0 100 389 master.ipa.example.
ipa-ca.ipa.example. 86400 IN A 10.37.170.206
```

ipaserver/install/service.py
tiran: Good point, I kept the old ``ldap_enable`` call and changed all other places to use the new method ``ldap_configure``.

tiran:
master:

* 7c2ca14118f3396b2b4ca4ac4d4d986569349415 Query for server role IPA master
* 10457a01bf6077e6978b3672dbbd7dc86a170e91 Only create DNS SRV records for ready server
* 7284097eedef70dd556270732e6ab8e23501ce09 Delay enabling services until end of installer

PR #2104, opened 2018-07-05
Replace file.flush() calls with flush_sync() helper

netoarmando:
Calls to `os.fsync(f.fileno())` need to be accompained by `f.flush()`.

Commit 8bbeedc93fd442cbbb9bb70e5f446011e95211db introduces the helper
`ipapython.ipautil.flush_sync()`, which handles all calls in the right
order.

However, `flush_sync()` takes as parameter a file object with fileno
and name, where name must be a path to the file, this isn't possible
in some cases where file descriptors are used.

Issue: https://pagure.io/freeipa/issue/7251


ipaclient/install/client.py
```diff
@@ -1275,7 +1275,7 @@ def do_nsupdate(update_txt):
 
     update_fd = open(UPDATE_FILE, "w")
     update_fd.write(update_txt)
-    update_fd.flush()
+    ipautil.flush_sync(update_fd)
```
tiran: While you are on it, please also refactor and fix simple file operations to use a context manager:

```
with open(UPDATE_FILE, "w") as f:
    f.write(update_txt)
    ipautil.flush_sync(f)
```

Be careful when it's a temporary file. close() also deleted the file.

ipatests/test_cmdline/test_cli.py
```diff
@@ -317,7 +318,7 @@ def test_without_options():
     def test_certfind(self):
         with tempfile.NamedTemporaryFile() as f:
             f.write(goodcert_headers)
-            f.flush()
+            ipautil.flush_sync(f)
```
tiran: You don't have to add flush to all places, only when it's a critical permanent file. This code and other places in ``ipatests`` only create temporary files, that will be deleted any way. ``fsync`` is a costly operation, because it forces the Kernel and disk driver to bypass write caches.

netoarmando: This is also true outside `ipatests`, right? In various places a `NamedTemporaryFile` is created only to be used as an argument for some outside tool (e.g. `openssl verify -CAfile {temp_file}) and deleted after that.

tiran: Excellent analysis. Yes, it's only relevant for persistent files and not for places that use a short-lived temporary file. We mostly care about config files, certs, keys, passwords etc.

ipapython/certdb.py
```diff
@@ -365,7 +365,9 @@ def create_db(self, user=None, group=None, mode=None, backup=False):
                                  os.O_CREAT | os.O_WRONLY,
                                  pwdfilemode), 'w', closefd=True) as f:
                 f.write(ipautil.ipa_generate_password())
+                # flush and sync tempfile inode
                 f.flush()
+                os.fsync(f.fileno())
```
netoarmando: @tiran I call `fsync` here because `f` doesn't have a name, although I didn't sync the directory inode as in: https://github.com/freeipa/freeipa/blob/f47d86c719aa67f62f3d0c54f5270fc0fc8d1393/ipapython/directivesetter.py#L57-L67

PR #2106, opened 2018-07-06

abbra:

`cacert.p12`.

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-pwd-extop/common.c
+++ b/daemons/ipa-slapi-plugins/ipa-pwd-extop/common.c
@@ -422,6 +422,7 @@ done:
 
 int ipapwd_entry_checks(Slapi_PBlock *pb, struct slapi_entry *e,
                         int *is_root, int *is_krb, int *is_smb, int *is_ipant,
+                        int *is_memberof,
                         char *attr, int acc)
 {
     Slapi_Value *sval;
@@ -440,6 +441,10 @@ int ipapwd_entry_checks(Slapi_PBlock *pb, struct slapi_entry *e,
         }
     }
 
+    /* Default to not setting memberof flag: only set it for non-Kerberos principals
+     * when they have krbPrincipalAux but no krbPrincipalName */
+    *is_memberof = 0;
+
     /* Check if this is a krbPrincial and therefore needs us to generate other
      * hashes */
     sval = slapi_value_new_string("krbPrincipalAux");
@@ -449,6 +454,24 @@ int ipapwd_entry_checks(Slapi_PBlock *pb, struct slapi_entry *e,
     }
     *is_krb = slapi_entry_attr_has_syntax_value(e, SLAPI_ATTR_OBJECTCLASS, sval);
     slapi_value_free(&sval);
+
+    /* If entry has krbPrincipalAux object class but lacks krbPrincipalName and
+     * memberOf attributes consider this not a Kerberos principal object. In
+     * FreeIPA krbPrincipalAux allows to store krbPwdPolicyReference attribute
+     * which is added by a CoS plugin configuration based on a memberOf
+     * attribute value.
+     * Note that slapi_entry_attr_find() returns 0 if attr exists, -1 for absence
+     */
+    if (*is_krb) {
+        Slapi_Attr *attr_prname = NULL;
+        Slapi_Attr *attr_memberof = NULL;
+        int has_prname = slapi_entry_attr_find(e, "krbPrincipalName", &attr_prname);
+        int has_memberOf = slapi_entry_attr_find(e, "memberOf", &attr_memberof);
+        if ((has_prname == -1) && (has_memberOf == 0)) {
+            *is_memberof = 1;
+            *is_krb = 0;
+        }
+    }
 
     sval = slapi_value_new_string("sambaSamAccount");
     if (!sval) {
@@ -655,7 +678,7 @@ int ipapwd_getEntry(Slapi_DN *sdn, Slapi_Entry **e2, char **attrlist)
 
     if (sdn == NULL) {
         LOG_TRACE("No entry to fetch!\n");
-	return LDAP_PARAM_ERROR;
+        return LDAP_PARAM_ERROR;
     }
 
     local_sdn = slapi_sdn_dup(sdn);
--- a/daemons/ipa-slapi-plugins/ipa-pwd-extop/common.c
+++ b/daemons/ipa-slapi-plugins/ipa-pwd-extop/common.c
@@ -72,6 +72,7 @@ static struct ipapwd_krbcfg *ipapwd_getConfig(void)
     Slapi_Entry *config_entry = NULL;
     Slapi_Attr *a;
     Slapi_Value *v;
+    Slapi_DN *sdn = NULL;
     BerElement *be = NULL;
     ber_tag_t tag, tvno;
     ber_int_t ttype;
@@ -107,7 +108,9 @@ static struct ipapwd_krbcfg *ipapwd_getConfig(void)
     }
 
     /* get the Realm Container entry */
-    ret = ipapwd_getEntry(ipa_realm_dn, &realm_entry, NULL);
+    sdn = slapi_sdn_new_dn_byval(ipa_realm_dn);
+    ret = ipapwd_getEntry(sdn, &realm_entry, NULL);
+    slapi_sdn_free(&sdn);
     if (ret != LDAP_SUCCESS) {
         LOG_FATAL("No realm Entry?\n");
         goto free_and_error;
@@ -212,7 +215,9 @@ static struct ipapwd_krbcfg *ipapwd_getConfig(void)
     slapi_entry_free(realm_entry);
 
     /* get the Realm Container entry */
-    ret = ipapwd_getEntry(ipa_pwd_config_dn, &config_entry, NULL);
+    sdn = slapi_sdn_new_dn_byval(ipa_pwd_config_dn);
+    ret = ipapwd_getEntry(sdn, &config_entry, NULL);
+    slapi_sdn_free(&sdn);
     if (ret != LDAP_SUCCESS) {
         LOG_FATAL("No config Entry? Impossible!\n");
         goto free_and_error;
@@ -236,7 +241,9 @@ static struct ipapwd_krbcfg *ipapwd_getConfig(void)
     if (ipapwd_fips_enabled()) {
         LOG("FIPS mode is enabled, NT hashes are not allowed.\n");
     } else {
-        ret = ipapwd_getEntry(ipa_etc_config_dn, &config_entry, NULL);
+        sdn = slapi_sdn_new_dn_byval(ipa_etc_config_dn);
+        ret = ipapwd_getEntry(sdn, &config_entry, NULL);
+        slapi_sdn_free(&sdn);
         if (ret != LDAP_SUCCESS) {
             LOG_FATAL("No config Entry?\n");
             goto free_and_error;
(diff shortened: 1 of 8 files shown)


abbra:
Note that to test this according to the ticket 7181, we need to prepare a second replica so that password reuse for PKI CA admin becomes a reality for a password policy check. Instead, it might be easier to just try to change a password of sudo sysaccount. Thus, patch would be a WIP until a test could be added.

abbra:
Added tests.

ipatests/test_integration/test_commands.py
```diff
@@ -134,6 +134,84 @@ def test_change_sysaccount_password_issue7561(self):
...
+
+        # Now try to change password twice:
+        # as a user and as a cn=Directory Manager
+        # If ticket 7181 is not fixed, the second change will fail
+        tasks.ldappasswd_sysaccount_change(sysuser, original_passwd,
+                                           new_passwd, master,
+                                           use_dirman=False)
+        tasks.ldappasswd_sysaccount_change(sysuser, new_passwd,
+                                           new_passwd, master,
+                                           use_dirman=True)
```
rcritten: I don't know yet if it is the test or something else but I added in another call to tasks.ldappasswd_sysaccount_change() with use_dirman=False and the test passed when it should have failed.

abbra: We have three different paths, original patch only covered pre_mod. I added change password op and pre_add. We now always check policy but for dirman case we ignore returned constraint violation.

rcritten:
Subsequent password changes are still successful. Given that it should fail I added this to the test;

diff --git a/ipatests/test_integration/test_commands.py b/ipatests/test_integ
ration/test_commands.py
index 8fdb6b891..7475c3634 100644
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -211,6 +211,9 @@ class TestIPACommand(IntegrationTest):
         tasks.ldappasswd_sysaccount_change(sysuser, new_passwd,
                                            new_passwd, master,
                                            use_dirman=True)
+        tasks.ldappasswd_sysaccount_change(sysuser, new_passwd,
+                                           original_passwd, master,
+                                           use_dirman=False)

The password change is successful when it should raise an error about re-use AFAICT.

abbra:
With latest changes we properly process password changes for non-Kerberos users -- I verified it in manual tests. However, the test does not fail for an additional password change somehow. I'll look at that next week.

1. Password change as a user
```
DEBUG - start_tls - Received extended operation request with OID 1.3.6.1.4.1.1466.20037
DEBUG - start_tls - Start TLS extended operation request confirmed.
DEBUG - ipa-pwd-extop - Attempting OTP authentication for 'uid=forcedpolicy,cn=sysaccounts,cn=etc,$SUFFIX'.
DEBUG - PBKDF2_SHA256 - Comparing password
DEBUG - ipa-pwd-extop - no krbPrincipalName in user entry: uid=forcedpolicy,cn=sysaccounts,cn=etc,$SUFFIX
ipa-lockout-plugin - preop returning 0: success
PBKDF2_SHA256 - Comparing password
DEBUG - backend_passwdmod_extop - extopdn = uid=forcedpolicy,cn=sysaccounts,cn=etc,$SUFFIX
DEBUG - ipa-pwd-extop - Received extended operation request with OID 1.3.6.1.4.1.4203.1.11.1
DEBUG - PBKDF2_SHA256 - Comparing password
DEBUG - PBKDF2_SHA256 - Generated hash {PBKDF2_SHA256}some hash
DEBUG - ipa-pwd-extop - Not a kerberos user, ignore krb attributes
DEBUG - ipa-pwd-extop - success
```

2. Password change as `cn=Directory Manager`:
```
DEBUG - backend_passwdmod_extop - extopdn = uid=forcedpolicy,cn=sysaccounts,cn=etc,$SUFFIX
DEBUG - ipa-pwd-extop - Received extended operation request with OID 1.3.6.1.4.1.4203.1.11.1
DEBUG - PBKDF2_SHA256 - Generated hash {PBKDF2_SHA256}some hash
DEBUG - ipa-pwd-extop - Not a kerberos user, ignore krb attributes
DEBUG - ipa-pwd-extop - success
```

3. New password change as a user using old password, leads to rejection:
```
DEBUG - start_tls - Received extended operation request with OID 1.3.6.1.4.1.1466.20037
DEBUG - start_tls - Start TLS extended operation request confirmed.
DEBUG - ipa-pwd-extop - Attempting OTP authentication for 'uid=forcedpolicy,cn=sysaccounts,cn=etc,$SUFFIX'.
DEBUG - PBKDF2_SHA256 - Comparing password
DEBUG - ipa-pwd-extop - no krbPrincipalName in user entry: uid=forcedpolicy,cn=sysaccounts,cn=etc,$SUFFIX
DEBUG - PBKDF2_SHA256 - Comparing password
DEBUG - backend_passwdmod_extop - extopdn = uid=forcedpolicy,cn=sysaccounts,cn=etc,$SUFFIX
DEBUG - ipa-pwd-extop - Received extended operation request with OID 1.3.6.1.4.1.4203.1.11.1
DEBUG - PBKDF2_SHA256 - Comparing password
DEBUG - ipa-pwd-extop - Password reuse not permitted
```

daemons/ipa-slapi-plugins/ipa-pwd-extop/common.c
```diff
@@ -465,6 +465,23 @@ int ipapwd_entry_checks(Slapi_PBlock *pb, struct slapi_entry *e,
...
 
+    /* If entry has krbPrincipalAux object class but lacks krbPrincipalName and
+     * memberOf attributes consider this not a Kerberos principal object. In
+     * FreeIPA krbPrincipalAux allows to store krbPwdPolicyReference attribute
+     * which is added by a CoS plugin configuration based on a memberOf
+     * attribute value.
+     * Note that slapi_entry_attr_find() returns 0 if attr exists, -1 for absence
+     */
+    if (*is_krb) {
+	Slapi_Attr *attr_prname = NULL;
```
tiran: It looks like your patch uses TAB while the rest of the code uses 4 space indention.

ipatests/test_integration/test_commands.py
```diff
@@ -134,6 +135,95 @@ def test_change_sysaccount_password_issue7561(self):
...
+
+        result = master.run_command(
+            ["ipa", "pwpolicy-add", policy_group,
+                "--history=5", "--priority=1"],
+        )
+        assert 'History size: 5' in result.stdout_text
+
+        # Add a system account and add it to a group managed by the policy
+        base_dn = str(master.domain.basedn)  # pylint: disable=no-member
+        tf = NamedTemporaryFile()
```
tiran: I have seen this hack a couple of times and still don't like it. It creates a local temporary file and assumes, that it's also a valid name on a remote machine. How about a new helper function in ``ipatests.pytest_plugins.integration`` to properly create a temporary file on the remote host?

```
def upload_temp_contents(host, contents, encoding='utf-8'):
    """Upload contents to a temporary file

    :param host: Remote host instance
    :param contents: file content (str, bytes)
    :param encoding: file encoding
    :return: Temporary file name
    """
    result = host.run_command(['mktemp'])
    tmpname = result.stdout_text.strip()
    host.put_file_contents(tmpname, contents, encoding=encoding)
    return tmpname
```

abbra:
@tiran I have pushed an update that addresses your concerns. I still need to find out what's wrong with a test but I changed an assertion for a succeeding `ldappasswd` to ensure test fails if the last password change succeeds.

ipatests/test_integration/test_commands.py
```diff
@@ -134,6 +134,94 @@ def test_change_sysaccount_password_issue7561(self):
...
+                                           use_dirman=True)
+        try:
+            tasks.ldappasswd_sysaccount_change(sysuser, new_passwd,
+                                               new_passwd, master,
+                                               use_dirman=False)
+        except CalledProcessError as e:
+            if e.returncode != 1:
+                raise
+        else:
+            assert "Password change violating policy did not fail" is False
```
tiran: That's an unusual way to raise an assertion. ``"message" is False`` is evaluated first. I prefer ``pytest.fail("message")``.

abbra:
@tiran changed assert to fail(message).

abbra:
Note that the test still fails, so this PR has to stay yet.

rcritten:
@abbra Did you want to do any more work on this, it LGTM?

abbra:
@simo5, could you please look at this patchset? I'm puzzled why it doesn't do what it should -- namely, the test for intentional password reuse passes while it should fail.

ipatests/test_integration/test_commands.py
simo5: Don, non-krb users even have a password history ?

daemons/ipa-slapi-plugins/ipa-pwd-extop/ipa_pwd_extop.c
```diff
@@ -561,16 +561,21 @@ static int ipapwd_chpwop(Slapi_PBlock *pb, struct ipapwd_krbcfg *krbcfg)
...
+		 * even if it doesn't need to be applied for Directory Manager
+		 * or passsync managers, filter that error out */
+		if (pwdata.changetype != IPA_CHANGETYPE_DSMGR) {
+			errMesg = ipapwd_error2string(ret);
 			ret = ipapwd_to_ldap_pwpolicy_error(ret);
 			slapi_pwpolicy_make_response_control(pb, -1, -1, ret);
 			rc = LDAP_CONSTRAINT_VIOLATION;
+			goto free_and_return;
 		}
-		goto free_and_return;
```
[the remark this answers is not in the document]
abbra: This is what I did in the first variant of this patch but then I realized the point of ipapwd_CheckPolicy() is also to update the password history. If we are applying a password history for non-Kerberos accounts, we should call ipapwd_CheckPolicy() so that ipapwd_check_policy() would update the history.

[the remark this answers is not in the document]
simo5: Yeah then it's time we split Checks from Updates, it's hard to reason about an interface when it starts doing things it does not declare.

daemons/ipa-slapi-plugins/ipa-pwd-extop/common.c
```diff
@@ -465,6 +465,23 @@ int ipapwd_entry_checks(Slapi_PBlock *pb, struct slapi_entry *e,
...
+     * attribute value.
+     * Note that slapi_entry_attr_find() returns 0 if attr exists, -1 for absence
+     */
+    if (*is_krb) {
+        Slapi_Attr *attr_prname = NULL;
+        Slapi_Attr *attr_memberof = NULL;
+        int has_prname = slapi_entry_attr_find(e, "krbPrincipalName", &attr_prname);
+        int has_memberOf = slapi_entry_attr_find(e, "memberOf", &attr_memberof);
+        if ((has_prname == -1) && (has_memberOf == 0)) {
+            *is_krb = 0;
```
simo5: Ok I think this is the cause of password history being simply not set therefore never matching.
As this line here:
  https://github.com/freeipa/freeipa/blob/0aeccc08524931b0e37773b4836734bc32eca7f4/daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c#L1059
causes this code to bail out before ipapwd_setPasswordHistory() is called a few lines below.

tiran:
@abbra I rebased your PR and addressed all merge conflicts.

abbra:
I'm at the point where the code works locally but fails in PR CI. I'm going to redo my local test environment to make sure I'm replicating the same behavior as in PR CI.

rcritten:
rebased for temp commit

abbra:
Rebased to master, let's start from scratch with tests.

rcritten:
The test is failing because the forcedpolicy sysaccount user has no passwordHistory.

It has no history because policy isn't being applied, I'm assuming because the sysaccount user doesn't have a krbPwdPolicyReference set. It may be an issue with the class of service configuration.

rcritten:
I think the issue is that this is reported against a user in o=ipaca and using global policy and the test is using a sysaccount and creating its own policy. I modified it locally and it passed, need to do some more research on why.

abbra:
It should not matter -- global policy is applied when there is no any other referenced policy. The issue I saw was error on processing missing reference (because the object has no objectclass that allows the policy inferring).
I'm adding a change to the test to enable trace logging in errorlog. It would slowdown that specific test but we'll have logs collected on traces around ipa-pwd-extop.

abbra:
For the reference, here is resulting set of tracing logs from executing ``test_commands::TestIPACommand::test_change_sysaccount_password_issue7181` test in Azure Pipelines:
[test_integration-test_commands.py-TestIPACommand-test_change_sysaccount_password_issue7181.tar.gz](https://github.com/freeipa/freeipa/files/4359169/test_integration-test_commands.py-TestIPACommand-test_change_sysaccount_password_issue7181.tar.gz)
I haven't yet got through them.

rcritten:
I added a patch to not save password history for non-Kerberos users.

I also added a couple of test cases to ensure that password policy is enforced and that non-kerb users don't get a history.

rcritten:
Closing since we can't coax PR-CI into picking this up. Will re-open in another PR.

PR #2108, opened 2018-07-06
[Backport][ipa-4-6] [py37] Import ABCs from collections.abc

tiran:
Manual backport of PR #2097 to keep the 4.6 compatible with future versions of pylint.

Python 3 has moved all collection abstract base classes to
collections.abc. Python 3.7 started to deprecate the old aliases.

The whole import block needs to be protected with import-error and
no-name-in-module, because Python 2 doesn't have collections.abc module and
collections.abc.Mapping, while Python 3 doesn't have collections.Mapping.

Fixes: https://pagure.io/freeipa/issue/7609

### DIFF
--- a/ipaclient/remote_plugins/__init__.py
+++ b/ipaclient/remote_plugins/__init__.py
@@ -2,7 +2,6 @@
 # Copyright (C) 2016  FreeIPA Contributors see COPYING for license
 #
 
-import collections
 import errno
 import json
 import locale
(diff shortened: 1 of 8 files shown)


tiran:
auto ACK backport PR

tiran:
ipa-4-6:

* e2347faad2e36ab8e81a04b8db18bd683dc3dfe2 Import ABCs from collections.abc

PR #2111, opened 2018-07-06
Tune DS replication settings

tiran:
Tune 389-DS replication settings to improve performance and avoid
timeouts. During installation of a replica, the value of
nsDS5ReplicaBindDnGroupCheckInterval is reduced to 2 seconds. At the end
of the installation, the value is increased sensible production
settings. This avoids long delays during replication.

See: https://pagure.io/freeipa/issue/7617

### DIFF
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -75,6 +75,20 @@ STRIP_ATTRS = ('modifiersName',
                'internalModifiersName',
                'internalModifyTimestamp')
 
+# settings for cn=replica,cn=$DB,cn=mapping tree,cn=config
+# during replica installation
+REPLICA_CREATION_SETTINGS = {
+    "nsds5ReplicaReleaseTimeout": ["20"],
+    "nsds5ReplicaBackoffMax": ["3"],
+    "nsDS5ReplicaBindDnGroupCheckInterval": ["2"]
+}
+# after replica installation
+REPLICA_FINAL_SETTINGS = {
+    "nsds5ReplicaReleaseTimeout": ["60"],
+    "nsds5ReplicaBackoffMax": ["300"],  # default
+    "nsDS5ReplicaBindDnGroupCheckInterval": ["60"]
+}
+
 
 def replica_conn_check(master_host, host_name, realm, check_ca,
                        dogtag_master_ds_port, admin_password=None,
@@ -201,9 +215,13 @@ def wait_for_entry(connection, dn, timeout, attr=None, attrvalue='*',
 
 
 class ReplicationManager(object):
-    """Manage replication agreements between DS servers, and sync
-    agreements with Windows servers"""
-    def __init__(self, realm, hostname, dirman_passwd, port=PORT, starttls=False, conn=None):
+    """Manage replication agreements
+
+    between DS servers, and sync  agreements with Windows servers
+    """
+
+    def __init__(self, realm, hostname, dirman_passwd=None, port=PORT,
+                 starttls=False, conn=None):
         self.hostname = hostname
         self.port = port
         self.dirman_passwd = dirman_passwd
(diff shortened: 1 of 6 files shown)


ipaserver/install/replication.py
```diff
@@ -515,9 +524,8 @@ def replica_config(self, conn, replica_id, replica_binddn):
             nsds5flags=["1"],
             nsds5replicabinddn=[replica_binddn],
             nsds5replicabinddngroup=[self.repl_man_group_dn],
-            nsds5replicabinddngroupcheckinterval=["60"],
-            nsds5replicareleasetimeout=["60"],
             nsds5replicalegacyconsumer=["off"],
+            **REPLICA_CREATION_SETTINGS
         )
```
tbordaz: The overall patch looks good to me but there is something not clear to me.
ipa-replica-install creates two suffixes 'domain' and 'ipaca'. Those suffixes being shared between master and replica, ipa-replica-install also creates replication entry (cn=replica,cn=<suffix>,cn=mapping tree,cn=config).
It creates those entries on both master and replica.

My question, are those tuning applied on those entries on master host, replica host or on both ?

ipaserver/install/replication.py
```diff
@@ -75,6 +75,17 @@
...
+# during replica installation
+REPLICA_CREATION_SETTINGS = {
+    "nsds5ReplicaReleaseTimeout": ["20"],
+    "nsds5ReplicaBackoffMax": ["3"],
+    "nsDS5ReplicaBindDnGroupCheckInterval": ["2"]
+}
+# after replica installation
+REPLICA_FINAL_SETTINGS = REPLICA_CREATION_SETTINGS.copy()
+REPLICA_FINAL_SETTINGS["nsDS5ReplicaBindDnGroupCheckInterval"] = ["60"]
+
```
tbordaz: Regarding REPLICA_FINAL_SETTINGS, it reset those values when the command ipa-replica-install complete.
Could we have a scenario where two ipa-replica-install step on the toes of each other so that one sets the parallel install settings on master (20s, 3s, 2s) and the other reset the the master to final value (60) so the first one will do its installation with not tuned master.

tbordaz: Also the final settings are
nsds5ReplicaReleaseTimeout: 60
nsds5ReplicaBackoffMax: 300 (or better remove this attribute)
nsDS5ReplicaBindDnGroupCheckInterval: 60

ipaserver/install/replication.py
tbordaz: Thanks that is perfect :)

ipaserver/install/replication.py
tbordaz: Yes that make sense, IMHO it will work most of the time

tbordaz: So for me the remaining concern is the 3 values to restore at the end of the installation (not only nsDS5ReplicaBindDnGroupCheckInterval)

ipaserver/install/replication.py
```diff
@@ -75,6 +75,20 @@
...
+    "nsds5ReplicaBackoffMax": ["3"],
+    "nsDS5ReplicaBindDnGroupCheckInterval": ["2"]
+}
+# after replica installation
+REPLICA_FINAL_SETTINGS =  {
+    "nsds5ReplicaReleaseTimeout": ["60"],
+    "nsds5ReplicaBackoffMax": ["300"],  # default
+    "nsDS5ReplicaBindDnGroupCheckInterval": ["60"]
+}
+
```
tbordaz: Thanks. The patch looks good to me. You have my ack

Tiboris:
ACKed due to https://github.com/freeipa/freeipa/pull/2111#discussion_r200981570

tiran:
master:

* 811b0fdb4620938963f1a29d3fdd22257327562c Tune DS replication settings

PR #2112, opened 2018-07-06
[Backport][ipa-4-6] Delay enablement of services and DNS SRV entries

tiran:
Manual backport of PR #2102 to 4.6 branch.

### Query for server role IPA master

server_find and server_role plugin were hiding IPA master role
information. It's now possible to fetch IPA master role information and
to filter by IPA master role, e.g. to ignore servers that have some
services configured but not (yet) enabled.

### Only create DNS SRV records for ready server

When installing multiple replicas in parallel, one replica may create
SRV entries for other replicas, although the replicas aren't fully
installed yet. This may cause some services to connect to a server, that
isn't ready to serve requests.

The DNS IPASystemRecords framework now skips all servers that aren't
ready IPA masters.

### Delay enabling services until end of installer

Service entries in cn=FQDN,cn=masters,cn=ipa,cn=etc are no longer
created as enabled. Instead they are flagged as configuredService. At
the very end of the installer, the service entries are switched from
configured to enabled service.

- SRV records are created at the very end of the installer.
- Dogtag installer only picks fully installed servers
- Certmonger ignores all configured but not yet enabled servers.

Fixes: pagure.io/freeipa/issue/7566

### DIFF
--- a/API.txt
+++ b/API.txt
@@ -4421,9 +4421,10 @@ output: Entry('result')
 output: Output('summary', type=[<type 'unicode'>, <type 'NoneType'>])
 output: PrimaryKey('value')
 command: server_role_find/1
-args: 1,8,4
+args: 1,9,4
 arg: Str('criteria?')
 option: Flag('all', autofill=True, cli_name='all', default=False)
+option: Flag('include_master', autofill=True, default=False)
 option: Flag('raw', autofill=True, cli_name='raw', default=False)
 option: Str('role_servrole?', autofill=False, cli_name='role')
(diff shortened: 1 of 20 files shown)


tiran:
ipa-4-6:

* e70c9223fd983fb7d2c95c138cc8e807bda41be9 Query for server role IPA master
* ce8890836854e41d6d3766894979110529858936 Only create DNS SRV records for ready server
* 8fd206f4abdace679b5a852b867f1cf95426ebee Delay enabling services until end of installer

PR #2113, opened 2018-07-06
ipa client uninstall: clean the state store when restoring hostname

flo-renaud:
When ipa client was installed with the --hostname= option, it stores
[network]
hostname = (current hostname)
in /var/lib/ipa-client/sysrestore/sysrestore.state and changes the hostname
from (current hostname) to the value provided in --hostname.

During uninstall, the previous hostname is restored but the entry does
not get removed from sysrestore.state. As the uninstaller checks if all
entries from sysrestore.state have been restored, it warns that some
state has not been restored.

The fix calls statestore.restore_state() instead of statestore.get_state()
as this method also clears the entry.

https://pagure.io/freeipa/issue/7620

### DIFF
--- a/ipaplatform/redhat/tasks.py
+++ b/ipaplatform/redhat/tasks.py
@@ -392,7 +392,7 @@ class RedHatTaskNamespace(BaseTaskNamespace):
         statestore.backup_state('network', 'hostname', old_hostname)
 
     def restore_hostname(self, fstore, statestore):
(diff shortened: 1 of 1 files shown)


tiran:
master:

* e32cfd14a9559a126e29f8c0215e3e80bf3924f6 ipa client uninstall: clean the state store when restoring hostname

PR #2114, opened 2018-07-06
Fix CA topology warning

tiran:
Commit 7284097eedef70dd556270732e6ab8e23501ce09 kept
find_providing_servers('CA') call before enable_services(). Therefore the
list of known CA servers did not contain the current replica.
ipa-replica-install on the first replica with --setup-ca still printed
the CA topology warning.

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -1523,8 +1523,6 @@ def install(installer):
     if options.setup_adtrust:
         adtrust.install(False, options, fstore, api)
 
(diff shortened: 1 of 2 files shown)


tiran:
Fix up for PR #2102 

I also addressed a unicode problem that might cause trouble on Python 2.

abbra:
Looks good. Unfortunately, we had no test for the warning message so the change in behaviour wasn't obvious.

tiran:
PR only applies to master and 4.6. The 4.5 branch doesn't contain any warning code.

tiran:
master:

* 8fa767622f2e3f40109cbb85b7d3dccfd4e33a2e Fix CA topology warning

PR #2120, opened 2018-07-06
Fix race condition in get_locations_records()

tiran:
The method IPASystemRecords.get_locations_records() has a race condition.
The IPASystemRecords object creates a mapping of server names to server
data. get_locations_records() uses server_find() again to get a list of
servers, but then operates on the cached dict of server names.

In parallel replication case, the second server_find() call in
get_locations_records() can return additional servers. Since the rest of
the code operates on the cached data, the method then fails with a KeyError.

server_data is now an OrderedDict to keep same sorting as with
server_find().

### DIFF
--- a/ipaserver/dns_data_management.py
+++ b/ipaserver/dns_data_management.py
@@ -8,7 +8,7 @@ import logging
 
 import six
 
-from collections import defaultdict
+from collections import defaultdict, OrderedDict
 from dns import (
     rdata,
     rdataclass,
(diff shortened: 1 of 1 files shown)


Tiboris:
Hi @tiran,
thank you for the patch!
LGTM

Tiboris:
master:

* 199d50a4c8ac2dd96a8bca3af4a90e4a9c05adf9 Fix race condition in get_locations_records()

PR #2122, opened 2018-07-08
Auto-retry failing certmonger requests

tiran:
During parallel replica installation, a request sometimes fails with
CA_REJECTED or CA_UNREACHABLE. The error occur when the master is
either busy or some information haven't been replicated yet. Even
a stuck request can be recovered, e.g. when permission and group
information have been replicated.

A new function request_and_retry_cert() automatically resubmits failing
requests until it times out.

``ipa-client-install --request-cert`` now also waits until the cert has been requested and no longer silence errors.

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -824,6 +824,7 @@ def configure_certmonger(
     cmonger = services.knownservices.certmonger
     try:
         cmonger.enable()
+        cmonger.start()
     except Exception as e:
         logger.error(
             "Failed to configure automatic startup of the %s daemon: %s",
@@ -835,19 +836,24 @@ def configure_certmonger(
     subject = str(DN(('CN', hostname), subject_base))
     passwd_fname = os.path.join(paths.IPA_NSSDB_DIR, 'pwdfile.txt')
     try:
-        certmonger.request_cert(
+        certmonger.request_and_wait_for_cert(
             certpath=paths.IPA_NSSDB_DIR,
             storage='NSSDB',
             nickname='Local IPA host',
             subject=subject,
             dns=[hostname],
             principal=principal,
-            passwd_fname=passwd_fname
+            passwd_fname=passwd_fname,
+            resubmit_timeout=120,
         )
-    except Exception as ex:
-        logger.error(
-            "%s request for host certificate failed: %s",
-            cmonger.service_name, ex)
+    except Exception as e:
+        logger.exception("certmonger request failed")
+        raise ScriptError(
+            "{} request for host certificate failed: {}".format(
+                cmonger.service_name, e
+            ),
+            rval=CLIENT_INSTALL_ERROR
+        )
 
 
 def configure_sssd_conf(
--- a/ipalib/install/certmonger.py
+++ b/ipalib/install/certmonger.py
@@ -305,20 +305,56 @@ def add_subject(request_id, subject):
 def request_and_wait_for_cert(
         certpath, subject, principal, nickname=None, passwd_fname=None,
         dns=None, ca='IPA', profile=None,
-        pre_command=None, post_command=None, storage='NSSDB', perms=None):
-    """
-    Execute certmonger to request a server certificate.
+        pre_command=None, post_command=None, storage='NSSDB', perms=None,
+        resubmit_timeout=0):
+    """Request certificate, wait and possibly resubmit failing requests
 
-    The method also waits for the certificate to be available.
+    Submit a cert request to certmonger and wait until the request has
+    finished.
+
+    With timeout, a failed request is resubmitted. During parallel replica
+    installation, a request sometimes fails with CA_REJECTED or
+    CA_UNREACHABLE. The error occurs when the master is either busy or some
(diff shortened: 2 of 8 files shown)


ipalib/install/certmonger.py
```diff
@@ -317,10 +317,62 @@ def request_and_wait_for_cert(
...
+        )
+
     return reqId
 
 
+def request_and_retry_cert(
+        certpath, subject, principal, nickname=None, passwd_fname=None,
+        dns=None, ca='IPA', profile=None,
+        pre_command=None, post_command=None, storage='NSSDB', perms=None,
+        timeout=120):
```
stlaz: Why not just enhance `request_and_wait_for_cert()` with an option like `retry_on_unreachable()`?

ipalib/install/certmonger.py
```diff
@@ -317,10 +317,62 @@ def request_and_wait_for_cert(
...
+    yet. Even a stuck request can be recovered, e.g. when permission and
+    group information have been replicated.
+    """
+    req_id = request_cert(
+        certpath, subject, principal, nickname, passwd_fname, dns, ca,
+        profile, pre_command, post_command, storage, perms
+    )
+
+    deadline = time.time() + timeout
+    while True:
```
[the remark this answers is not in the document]
tiran: My first implementation looked like that. But I didn't want to define ``state`` and ``ca_error``.

ipalib/install/certmonger.py
tiran: That was in my first implementation, too. The code got too confusing and it was easier to rewrite the function from scratch. Since ``request_and_wait_for_cert`` isn't used any more, we could default to the timeout version.

ipalib/install/certmonger.py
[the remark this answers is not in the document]
tiran: Two more reasons why I prefer ``while True`` loop here:

* The current implementation is basically a ``do/while`` loop. It will run at least one time with ``timeout=0``.
* The loop has three exit conditions: success, timeout, and error. I find it confusing to have only one exit check in the loop header.

I can certainly add a comment and make the code flow more obvious.

ipaclient/install/client.py
```diff
@@ -835,19 +838,24 @@ def configure_certmonger(
...
             principal=principal,
-            passwd_fname=passwd_fname
+            passwd_fname=passwd_fname,
+            resubmit_timeout=120,
+        )
+    except Exception as e:
+        logger.exception("certmonger request failed")
+        raise ScriptError(
+            "{} request for host certificate failed: {}".format(
+                cmonger.service_name, e
```
Tiboris: ugly ugly

tiran: I already pushed a new build about half an hour ago.

Tiboris: already resolved LOL

ipaclient/install/client.py
```diff
@@ -823,7 +823,10 @@ def configure_certmonger(
 
     cmonger = services.knownservices.certmonger
     try:
-        cmonger.enable()
+        if not cmonger.is_enabled():
```
stlaz: What's the purpose of this? `cmonger.enable()` won't raise so you don't have to check for enablement.

ipaclient/install/client.py
```diff
@@ -823,7 +823,10 @@ def configure_certmonger(
...
+            cmonger.enable()
+        if not cmonger.is_running():
+            cmonger.start()
```
stlaz: Similarly to the above.

ipaclient/install/client.py
```diff
@@ -835,19 +838,24 @@ def configure_certmonger(
...
             storage='NSSDB',
             nickname='Local IPA host',
             subject=subject,
             dns=[hostname],
             principal=principal,
-            passwd_fname=passwd_fname
+            passwd_fname=passwd_fname,
+            resubmit_timeout=120,
+        )
+    except Exception as e:
```
stlaz: Could you please enhance this code? I do not like broad exceptions, especially when so many arguments are being passed.

Best thing to do would be to remove the broad exception here:
https://github.com/freeipa/freeipa/blob/199d50a4c8ac2dd96a8bca3af4a90e4a9c05adf9/ipalib/install/certmonger.py#L386-L388
and replace it with `RuntimeError, dbus.exceptions.DBusException`

tiran: This is really the case of *Turn all errors into a ScriptError*. The code logs the exception and throws a new exception with the stack context of the inner exception intact. Also I don't want to make the ``ipaclient`` package directly depend on dbus package.

ipaclient/install/client.py
tiran: Good to know, I'll remove the check.

ipaclient/install/client.py
stlaz: Alright

tiran:
master:

* 1fa2a7cd41095295ebee5cd3b280507580ba8fbb Auto-retry failed certmonger requests
* 2b669c52a566382eecbed1511640f647d54f5b55 Wait for client certificates

PR #2123, opened 2018-07-09
Fix link to browser configuration guide at Login page

stanislavlevin:
There is a mismatch between 'i18n' krb_auth_msg and 'LoginScreen'
widget kerberos_msg. The former links to "unauthorized.html", but the latter
to "ssbrowser.html". Both should link to "ssbrowser.html" page.


abbra:
Looks good to me.

abbra:
Could you please open a ticket too so that we could get it to 4.7?

stanislavlevin:
Of course! Pushed.

abbra:
ACKed. Will go in with the rest of changes related to Web UI.

PR #2128, opened 2018-07-10
Fix DNSSEC install regression

tiran:
7284097eedef70dd556270732e6ab8e23501ce09 introduced a regression in
DNSSEC master installation. For standalone and replica installation,
services have to be enabled before checking bind config.

Fixes: https://pagure.io/freeipa/issue/7635


Tiboris:
Thanks for the patch, @tiran. LGTM.

Tiboris:
master:

* 9222a08c288159a7a150923a2eb825a47da36a13 Fix DNSSEC install regression

PR #2129, opened 2018-07-10
Turn multihost config problems into errors

tiran:
The pytest multihost plugin skips tests, when there is a problem with a
test configuration. Configuration bugs like missing resources are not
considered a problem.

The IPA mh fixture now catches the skip exception and turns it into a
fatal error.

Fixes: https://pagure.io/freeipa/issue/7638

### DIFF
--- a/ipatests/pytest_plugins/integration/config.py
+++ b/ipatests/pytest_plugins/integration/config.py
@@ -23,6 +23,7 @@
 import logging
 import random
 
+import pytest
 import pytest_multihost.config
 
 from ipapython.dn import DN
@@ -98,6 +99,18 @@ class Config(pytest_multihost.config.Config):
     def to_env(self, **kwargs):
         from ipatests.pytest_plugins.integration.env_config import config_to_env
         return config_to_env(self, **kwargs)
+
+    def filter(self, descriptions):
+        """Destructively filters hosts and orders domains to fit description
+
+        By default make_multihost_fixture() skips a test case, when filter()
+        returns a FilterError. Let's turn FilterError into a fatal error
+        instead.
+        """
+        try:
+            super(Config, self).filter(descriptions)
+        except pytest_multihost.config.FilterError as e:
+            pytest.fail(str(e))
 
 
 class Domain(pytest_multihost.config.Domain):


tiran:
This PR turns problems like ``"Skipped: Not enough resources configured: Domain 0 not configured: {'type': 'IPA', 'hosts': {'master': 1, 'replica': 0, 'client': 1}}"`` for ``test_caless.py::TestClientInstall::()::test_client_install::setup`` into an actual error.

tiran:
On nightly PR-CI, several ``fedora-28/test_replica_promotion_*`` tests are not executed, too.

Tiboris:
@tiran good catch! 
Thanks for the patch :)

Tiboris:
Did we have an agreement with @gkaihorodova on how to proceed with this kind of behavior?

ipatests/pytest_plugins/integration/__init__.py
```diff
@@ -225,12 +225,15 @@ def mh(request, class_integration_logs):
...
-    )
+    try:
+        mh = make_multihost_fixture(
+            request,
+            domain_descriptions,
+            config_class=Config,
+            _config=get_global_config(),
+        )
+    except pytest.skip.Exception as e:
+        pytest.fail(str(e))
```
Tiboris: Can we replace this with skip?

gkaihorodova:
Hello @tiran, @Tiboris 

Skipped is a clear and separate state of the test, which means that test wasn't run. It doesn't mean that test failed itself. It means that it wasn't run by some reason, which needs to be solved. While we don't have a result of the test run - positive or negative, it's not right to mark it as "Failure", it can mislead.
Configuration test environment and test results are different things, I wouldn't go that way and mixed it.
Configuration issues should be solved separately  in my opinion

ipatests/pytest_plugins/integration/__init__.py
Tiboris: thanks for explanation

gkaihorodova:
@tiran 

 "A misconfigured or ill configured test should not pass as successful test" - it is not successful, it's just doesn't have result to be labeled  to success or failure- it's skipped, 
No problem with test, but with environment configuration to run test. Maybe, is better to think and try to resolve issue on the side of environment configuration? Some resource check  before test run? (Just thinking )

tiran:
> "A misconfigured or ill configured test should not pass as successful test" - it is not successful, it's just doesn't have result to be labeled to success or failure- it's skipped,

PR-CI results on GitHub have a binary state. There is only **successful** and **failed**. Skipped tests are not shown as skipped, but are mapped to success state, so any configuration issue is hidden.

> Maybe, is better to think and try to resolve issue on the side of environment configuration? Some resource check before test run? (Just thinking )

I fully agree with you. But in order to resolve configuration issues, we have to spot them first. This PR makes it easier to spot issue and to avoid similar problems in the future.

Rezney:
At the moment I tend to be more in favor of marking skipped tests as failed cause the default behavior can hide issues which we do not want to be hidden. E.g. when I was playing with paramiko upgrade 2 weeks ago during one of my tests upgrading the version caused all the tests to be skipped and eventually mark them as passed. I only knew there was an issue because I was watching the tests and spotted that they took just several minutes to be completed.

Can we run this patch upon our nightly tests? You can use the following yaml:

https://github.com/freeipa-pr-ci/freeipa/blob/a9e8857366fe1416c5ed987525c6a546a8fd86a1/.freeipa-pr-ci.yaml

tiran:
Thanks @Rezney 

I already identified a couple of jobs that are misconfigured. See second patch of https://github.com/freeipa/freeipa/pull/2139/commits

tiran:
## Known failing jobs

### fedora-28/test_caless_TestClientInstall

``pytest_multihost.config.FilterError: Domain 0 not configured: {'type': 'IPA', 'hosts': {'master': 1, 'replica': 0, 'client': 1}}``

#### fedora-28/test_replica_promotion_TestKRAInstall

#### fedora-28/test_replica_promotion_TestCAInstall

#### fedora-28/test_replica_promotion_TestReplicaManageCommands

#### fedora-28/test_replica_promotion_TestProhibitReplicaUninstallation

rcritten:
+1. The tests should be vocal when they are expected to execute but cannot, for whatever reason.

gkaihorodova:
> PR-CI results on GitHub have a binary state. There is only successful and failed. Skipped tests are not shown as skipped, but are mapped to success state, so any configuration issue is hidden.

@tiran for me it's still a bit radical solution, but I have to agree that in case of combination succeeded tests + skipped = green successful run it make sense.

tiran:
I found a better way to raise the error. The new patch no longer wraps ``make_multihost_fixture()``. Instead IPA ``Config.filter`` raises another exception. I'll push the new patch when PR-CI has finished.

```
--- a/ipatests/pytest_plugins/integration/config.py
+++ b/ipatests/pytest_plugins/integration/config.py
@@ -23,6 +23,7 @@
 import logging
 import random
 
+import pytest
 import pytest_multihost.config
 
 from ipapython.dn import DN
@@ -99,6 +100,18 @@ class Config(pytest_multihost.config.Config):
         from ipatests.pytest_plugins.integration.env_config import config_to_env
         return config_to_env(self, **kwargs)
 
+    def filter(self, descriptions):
+        """Destructively filters hosts and orders domains to fit description
+
+        By default make_multihost_fixture() skips a test case when filter()
+        returns a FilterError. Let's turn FilterError into fatal error
+        instead pytest.skip().
+        """
+        try:
+            super(Config, self).filter(descriptions)
+        except pytest_multihost.config.FilterError as e:
+            pytest.fail(str(e))
+
 
 class Domain(pytest_multihost.config.Domain):
     """Configuration for an IPA / AD domain"""

```

tiran:
master:

* 59ef5371e1f34a25408fafcc8aca740a1a2ef724 Turn multihost config problems into errors

PR #2130, opened 2018-07-10
Server-del test suite fix

pavelpicka:
Install topology manually as do not need uninstall from annotated methods
and remove replica manually


tiran:
The test case ``test_server_del.py::TestLastServices::()::test_forced_removal_of_master::teardown`` is failing with error message ``Server removal aborted: Deleting this server is not allowed as it would leave your installation without a CA..``

tiran:
You are almost there! ``test_server_del.py::TestLastServices::()::test_forced_removal_of_master::teardown`` fails with message

```
Server removal aborted: Deleting this server is not allowed as it would leave your installation without a CA.
```

PR #2131, opened 2018-07-10
Mark all expected failures as strict

tiran:
With strict=True, xfail() fails when the test case passes unexpectably.
This allows us to spot passing tests that are expected to fail.

Fixes: https://pagure.io/freeipa/issue/7613


tiran:
master:

* 0c1010d6f7585095402f47c111b8bb6cde3068b3 Mark all expected failures as strict
* ec65590c9f2e00d0e527e7bbf521a0b99997788c Fix XPASS in test_installation

PR #2132, opened 2018-07-10
Handle races in replica config

tiran:
When multiple replicas are installed in parallel, two replicas may try
to create the cn=replica entry at the same time. This leads to a
conflict on one of the replicas. replica_config() now handles the
conflict.


ipaserver/install/replication.py
```diff
@@ -476,61 +476,69 @@ def set_replica_binddngroup(self, r_conn, entry):
...
-            nsds5flags=["1"],
-            nsds5replicabinddn=[replica_binddn],
-            nsds5replicabinddngroup=[self.repl_man_group_dn],
-            nsds5replicalegacyconsumer=["off"],
-            **REPLICA_CREATION_SETTINGS
-        )
-        conn.add_entry(entry)
+        self._set_replica_binddngroup(conn, entry)
+
+        return entry
```
tbordaz: This part of the code looks good to me but I think it will fail at the next step: setup_changelog.

If two ipa-replica-install are doing the setup_promote_replication at the same time. The patch prevents a failure at replica_config (catching the replica already existing). So one did a ADD and the other will check the binddngroup. Next step is setup_changelog. We should also harden setup_changelog from running in parallel.

tiran: ``setup_changelog`` already handles add conflicts and catches the duplicate entry error.

https://github.com/freeipa/freeipa/blob/b21c3f9e88ea0b54140be02ba72c7076f05c8209/ipaserver/install/replication.py#L543-L563

tbordaz: Right. One place I found fragile is ensure_replication_managers when it calls _add_replication_managers. The replication managers entry can already exist

https://github.com/freeipa/freeipa/blob/b21c3f9e88ea0b54140be02ba72c7076f05c8209/ipaserver/install/replication.py#L1671-1693

tiran: Memo to me: Look into ``ldap.TYPE_OR_VALUE_EXISTS``

ipaserver/plugins/ldap2.py
```diff
@@ -419,7 +419,8 @@ def add_entry_to_group(self, dn, group_dn, member_attr='member', allow_same=Fals
                 modlist = [(a, b, self.encode(c))
                            for a, b, c in modlist]
                 self.conn.modify_s(str(group_dn), modlist)
-        except errors.DatabaseError:
+        except errors.DuplicateEntry:
+            # TYPE_OR_VALUE_EXISTS
             raise errors.AlreadyGroupMember()
```
tbordaz: indeed it was a weird translation DB_ERROR -> AlreadyGroupMember

The patch looks good to me. ACK

PR #2135, opened 2018-07-10
Fix pylint 2.0 return-related violations

netoarmando:
Aiming to support pylint 2.0 some functions and methods must have their
return statements updated in order to fix two new violations:

- `useless-return` (R1711):
  Useless return at end of function or method Emitted when a single
  "return" or "return None" statement is found at the end of function
  or method definition. This statement can safely be removed because
  Python will implicitly return None

- `inconsistent-return-statements` (R1710):
  Either all return statements in a function should return an
  expression, or none of them should. According to PEP8, if any return
  statement returns an expression, any return statements where no value
  is returned should explicitly state this as return None, and an
  explicit return statement should be present at the end of the
  function (if reachable)

### DIFF
--- a/ipaserver/dnssec/syncrepl.py
+++ b/ipaserver/dnssec/syncrepl.py
@@ -42,6 +42,7 @@ class SyncReplConsumer(ReconnectLDAPObject, SyncreplConsumer):
             return cookie
         else:
             logger.debug('Current cookie is: None (not received yet)')
+            return None
 
     def syncrepl_set_cookie(self, cookie):
         logger.debug('New cookie is: %s', cookie)
--- a/ipaserver/install/ipa_cacert_manage.py
+++ b/ipaserver/install/ipa_cacert_manage.py
@@ -126,6 +126,8 @@ class CACertManage(admintool.AdminTool):
                 return self.renew()
             elif command == 'install':
                 return self.install()
+            else:
+                raise NotImplementedError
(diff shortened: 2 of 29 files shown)


ipaserver/dnssec/syncrepl.py
```diff
@@ -42,6 +42,7 @@ def syncrepl_get_cookie(self):
             return cookie
         else:
             logger.debug('Current cookie is: None (not received yet)')
+        return None
```
tiran: Please indent this line
```
else:
    logger.debug('Current cookie is: None (not received yet)')
    return None
```

ipaserver/install/ipa_cacert_manage.py
```diff
@@ -128,6 +128,7 @@ def run(self):
                 return self.install()
         finally:
             api.Backend.ldap2.disconnect()
+        return None
```
tiran: ``validate_option`` ensures, that this line is not reachable. Please either add a comment or change the fix to something like:

```
             if command == 'renew':
                 return self.renew()
             elif command == 'install':
                 return self.install()
             else:
                 # Unreachable
                 raise NotImplementedError
```

tiran:
master:

* b4ad0d19a20cc66690e7f4e9c4327afeedff2ab2 Fix pylint 2.0 return-related violations

PR #2136, opened 2018-07-11
[Backport][ipa-4-6] Fix pylint 2.0 return-related violations

tiran:
Manual backport of PR #2135 to 4.6 branch.

Aiming to support pylint 2.0 some functions and methods must have their
return statements updated in order to fix two new violations:

- `useless-return` (R1711):
  Useless return at end of function or method Emitted when a single
  "return" or "return None" statement is found at the end of function
  or method definition. This statement can safely be removed because
  Python will implicitly return None

- `inconsistent-return-statements` (R1710):
  Either all return statements in a function should return an
  expression, or none of them should. According to PEP8, if any return
  statement returns an expression, any return statements where no value
  is returned should explicitly state this as return None, and an
  explicit return statement should be present at the end of the
  function (if reachable)


ipaserver/install/ipa_server_certinstall.py
```diff
@@ -225,6 +225,7 @@ def install_kdc_cert(self):
                         profile='KDCs_PKINIT_Certs')
         except RuntimeError as e:
             raise admintool.ScriptError(str(e))
+        return None
```
stlaz: wft?

stlaz: Ah, I see, improper rebase. Please fix.

tiran: Good catch! Fixed

stlaz:
LGTM

PR #2140, opened 2018-07-11
Add tab completion and history to ipa console

tiran:
ipa console is a useful tool to use FreeIPA's API in an interactive
Python console. The patch adds readline tab completion and history
support.

The PR also fixes a bug in ``ipa console filename`` case and adds a test case.

### DIFF
--- a/ipalib/cli.py
+++ b/ipalib/cli.py
@@ -992,12 +992,18 @@ class console(frontend.Command):
         )
         if filename:
             try:
-                script = open(filename)
+                with open(filename) as f:
+                    source = f.read()
             except IOError as e:
                 sys.exit("%s: %s" % (e.filename, e.strerror))
             try:
-                with script:
-                    exec(script, globals(), local)
+                compiled = compile(
+                    source,
+                    filename,
+                    'exec',
+                    flags=print_function.compiler_flag
+                )
+                exec(compiled, globals(), local)
             except Exception:
                 traceback.print_exc()
                 sys.exit(1)
--- a/ipalib/cli.py
+++ b/ipalib/cli.py
@@ -22,6 +22,7 @@ Functionality for Command Line Interface.
 """
 from __future__ import print_function
 
+import atexit
 import importlib
 import logging
 import textwrap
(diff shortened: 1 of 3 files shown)


abbra:
Thank you, @tiran, looks very useful. Do we have any of tests for interactive CLI? I guess not because for example `ipa console <filename>` fails:

```
# cat f.py 
print(api)

# ipa console f.py 
  File "/usr/lib/python3.6/site-packages/ipalib/cli.py", line 961, in run
    exec(script, globals(), local)
TypeError: exec() arg 1 must be a string, bytes or code object

```

abbra:
Perhaps, we should add a test that includes something like below where `f.py` could be more than just valid Python code -- it could include keystrokes too to retrieve something from a history.
```
# cat f.py |ipa console
(Custom IPA interactive Python console)
>>> <ipalib.API object at 0x7f6859f89208>
>>> 
now exiting InteractiveConsole...
```

tiran:
Today's build ``freeipa-client.x86_64 4.6.90.pre2.dev201807110728+git9222a08c2-0.fc28`` works fine for me:

```
# echo "api.env.version" | ipa console
(Custom IPA interactive Python console)
>>> '4.6.90.pre2.dev201807110728+git9222a08c2'
>>> 
now exiting InteractiveConsole...
# cat f.py | ipa console
(Custom IPA interactive Python console)
>>> 4.6.90.pre2.dev201807110728+git9222a08c2
>>> >>> 
now exiting InteractiveConsole...
# cat f.py 
print(api.env.version)

```

tiran:
ah, the ``ipa console filename`` case was broken. I added a test for both scenarios.

abbra:
@tiran all looks good except `upload_temp_contents()` -- please take a separate commit from PR#2106 so that we don't get them conflicting.

tiran:
@abbra It's a separate commit now.

abbra:
LGTM. Thanks!

tiran:
master:

* 10de2f37a25e1781026014ef1ee7180b17e8f422 Add tab completion and history to ipa console
* 5affc9b982f6314fc6d1c63fb687acee20f5144b Create helper function to upload to temp file
* 87904b8f6b7ebe4900572e45199449d8b7e47cc0 Fix ipa console filename

PR #2141, opened 2018-07-11
Fix pylint 2.0 conditional-related violations

netoarmando:
In order to support pylint 2.0 the following violations must be fixed:

- `chained-comparison` (R1716):
  Simplify chained comparison between the operands This message is
  emitted when pylint encounters boolean operation like
  "a < b and b < c", suggesting instead to refactor it to "a < b < c".

- `consider-using-in` (R1714):
  Consider merging these comparisons with "in" to %r To check if a
  variable is equal to one of many values,combine the values into a
  tuple and check if the variable is contained "in" it instead of
  checking for equality against each of the values.This is faster
  and less verbose.

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -2081,7 +2081,7 @@ def install_check(options):
             rval=CLIENT_INSTALL_ERROR
         )
 
-    if (hostname == 'localhost') or (hostname == 'localhost.localdomain'):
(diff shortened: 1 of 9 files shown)


tiran:
master:

* ba954efafdb3e430c76dfc327d2b683ac0e117eb Fix pylint 2.0 conditional-related violations

PR #2144, opened 2018-07-12
Unify and simplify LDAP service discovery

tiran:
Move LDAP service discovery and service definitions from
ipaserver.install to ipaserver. Simplify and unify different
implementations in favor of a single implementation.

Remark: The first iterations of this PR had active service discovery. This turns out to be too slow and unstable.

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -44,6 +44,7 @@ from ipaserver.install.installutils import (
     ReplicaConfig, load_pkcs12, is_ipa_configured)
 from ipaserver.install.replication import (
     ReplicationManager, replica_conn_check)
+from ipaserver.masters import find_providing_servers, find_providing_server
 import SSSDConfig
 from subprocess import CalledProcessError
 
@@ -1021,9 +1022,10 @@ def promote_check(installer):
         if subject_base is not None:
             config.subject_base = DN(subject_base)
 
-        # Find if any server has a CA
-        ca_host = service.find_providing_server(
-                'CA', conn, config.ca_host_name)
+        # Find any server with a CA
+        ca_host = find_providing_server(
+            'CA', conn, [config.ca_host_name]
+        )
         if ca_host is not None:
             config.ca_host_name = ca_host
             ca_enabled = True
@@ -1044,14 +1046,16 @@ def promote_check(installer):
                              "custom certificates.")
                 raise ScriptError(rval=3)
 
-        kra_host = service.find_providing_server(
-                'KRA', conn, config.kra_host_name)
+        # Find any server with a KRA
+        kra_host = find_providing_server(
+            'KRA', conn, [config.kra_host_name]
+        )
         if kra_host is not None:
             config.kra_host_name = kra_host
             kra_enabled = True
         else:
             if options.setup_kra:
-                logger.error("There is no KRA server in the domain, "
+                logger.error("There is no active KRA server in the domain, "
                              "can't setup a KRA clone")
                 raise ScriptError(rval=3)
             kra_enabled = False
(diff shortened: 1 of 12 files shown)


ipaserver/install/server/replicainstall.py
```diff
@@ -1555,7 +1566,7 @@ def install(installer):
     # Enable configured services and update DNS SRV records
     service.enable_services(config.host_name)
     api.Command.dns_update_system_records()
-    ca_servers = service.find_providing_servers('CA', api.Backend.ldap2, api)
+    ca_servers = find_providing_servers('CA', api.Backend.ldap2, api)
```
rcritten: Does it make more sense to make find_providing_servers() public or use the server_role mechanism to see if there is more than one CA installed?

rcritten:
A couple of the host tests are raising an Internal Error but this seems to be on the right track.

pvoborni:
Current server picking mechanism which is used also in vault plugin produces a bug https://pagure.io/freeipa/issue/7691 . Comment https://pagure.io/freeipa/issue/7691#comment-535233 analyses the situation. In short, by default, standard users don't have read permissions for service information in etc=masters subtree. Thus method like `find_active_server` will return `None` later causing the probably bad behavior of get_client in kra backend.

It makes sense to think about the context of this bug also in this refactoring, though the bug might be fixed separately. 

Also, just by looking at the code, it seems that this refactoring aims to partly resolve: https://fedorahosted.org/freeipa/ticket/4557

tiran:
`` RunPytest Failed to publish artifacts from ci-large-26``, rerun

fcami:
Full review done & this looks good to me (without actually running the code).

@tiran @rcritten indeed maybe find_providing_server(s) would be best in serverroles (serverroles.py) alongside server_role_search and friends?

Proposed changes (error messages and comments) live in:
https://github.com/fcami/freeipa/tree/active_server

fcami:
@tiran @rcritten - replying to myself I'd vote for merging this as-is and open another issue to investigate whether moving find_providing_server(s)  to serverroles is worthwhile. 

@tiran could you remove 'LDAP' from the commit message, as your work is about any kind of service discovery?

ipaserver/masters.py
```diff
@@ -0,0 +1,119 @@
...
+    # unique list of host names, DNS is case insensitive
+    servers = list(set(entry.dn[1].value.lower() for entry in entries))
+    # shuffle the list like DNS SRV would randomize it
+    random.shuffle(servers)
+    # Move preferred hosts to front
+    for host_name in reversed(preferred_hosts):
+        host_name = host_name.lower()
+        try:
+            servers.remove(host_name)
+        except ValueError:
```
rcritten: If a server is in the preferred list but isn't in the available list then it will be silently dropped. Should this be handled in some way?

rcritten:
I thought about the suggested move to servroles.py as well but it doesn't quite fit there. I'm not super-keen on a new file in ipaserver but I don't know where else it might go other than ipaserver/install.

ipaserver/masters.py
tiran: The case may occur if a server has an explicit ``ca_host`` in its ``default.conf`` but the machine has been decommissioned. It's an uncommon case. I'd rather not fail here if IPA can recover automatically. How about ``log.warn``?

tiran:
@fcami It's service discovery **from** LDAP, not **of** LDAP. For most services, IPA uses DNS SRV based discovery. LDAP service discovery is only used in some special cases, e.g. because IPA doesn't have SRV records for CA and KRA.

@rcritten yeah, ``ipaserver.install`` feels wrong. Modules are cheap and I prefer to put related code into one module rather than creating another 3k ``misc``, ``stuff``, or ``util`` module.

ipaserver/masters.py
rcritten: Yes, that'd be fine.

rcritten:
I'm ok with putting it into ipaserver. I'm trying to be cautious about moving code around in case others are using the internal API (yeah, unsupported, etc but it does break things from time to time).

rcritten:
This change LGTM for master.

rcritten:
Works in my testing, ack.

tiran:
master:

* 8decef33d310926c766f9b80ce78655b553a247d Unify and simplify LDAP service discovery

PR #2151, opened 2018-07-13
Fix translation of "unauthorized" and "ssbrowser" WebUI pages

stanislavlevin:
This is a simple fix for  https://pagure.io/freeipa/issue/7640.
But it depends on PR #2072 .


pvoborni:
Thanks for the PR. I ponder how to approach it. The PR introduces a translation string which overrides the same text in noscript. As it looks functionally OK, I'm afraid that the maintenance part of it will suffer. Especially for ssbrowser which has really long text. The worry is that the text can get out of sync - there is no ensuring that the raw text in noscript element matches (e.g. without whitespaces) the text in internal.py. The internal.py string is really difficult to read. Maybe a test which would check it could prevent such issues in the future.

I'm not a translator, but I wonder how difficult is to translate such a string as the one in internal.py. Would it be more friendly if the long one was split into multiple parts?

stanislavlevin:
@pvoborni , 
1) yes, actually text of ssbrowser is very long. But there are many other pieces of text from different plugins. For example (this is a screenshot from zanata where actual translation is performed ):

screenshot_20180717_222011

This message takes ~100 lines. Next example shows text mixed with html tags, readability is pretty enough.

screenshot_20180717_230320

2) recently i faced with a similar problem (text in internal.py was mismatched from widget's one #2123 )

3) i worry about an expected behavior, that's why i put text into '\<noscript\>' - even without JS enabled client can see an error. Though without JS nobody can use IPA Web UI. But if it is not the case...

pvoborni:
1. The long example is formatted and the unformatted is short. So they don't seem to prove the point. But I'm fine with it. It was originally just a question.

2. The sync between the texts worries me more. It might be OK to not test short texts as reviews of short text are not that difficult, as mentioned, even there can be issues. With long ones, we are asking for issues. I do not think that we should push this without an automated test to compare the English texts.

3. The approach with <noscript> is a nice one. So +1 to it.

stanislavlevin:
1) translatable text of ssbrowser was reformatted
2) tests were added (tested on firefox)
```
ipa-run-tests-3 -v test_webui/test_translation.py
===================================== test session starts =====================================
platform linux -- Python 3.6.5, pytest-3.4.2, py-1.5.3, pluggy-0.6.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/test, inifile:
plugins: sourceorder-0.5.1, multihost-3.0
collected 2 items                                                                             

test_translation.py::TestConfigPage::test_long_text_of_ssbrowser_page <- ../../usr/lib/python3/site-packages/ipatests/test_webui/test_translation.py PASSED [ 50%]
test_translation.py::TestConfigPage::test_long_text_of_unauthorized_page <- ../../usr/lib/python3/site-packages/ipatests/test_webui/test_translation.py PASSED [100%]

================================== 2 passed in 13.84 seconds ==================================
```

ipatests/test_webui/test_translation.py
```diff
@@ -0,0 +1,113 @@
...
+class TestConfigPage(UI_driver):
+    """
+    Test translation of pages which are located at /ipa/config/
+    """
+
+    def get_config_url(self):
+        """
+        Get FreeIPA Web UI config url
+        """
+        host = self.config.get('ipa_server')
```
serg-cymbaluk: Why wouldn't override get_base_url and keep all magic under the hood in UI_driver?

ipatests/test_webui/test_translation.py
```diff
@@ -0,0 +1,113 @@
...
+
+from ipatests.test_webui.ui_driver import UI_driver
+from ipatests.test_webui.ui_driver import screenshot
+
+try:
+    from selenium.webdriver.common.by import By
+    from selenium.webdriver.support.wait import WebDriverWait
+    NO_SELENIUM = False
+except ImportError:
+    NO_SELENIUM = True
```
serg-cymbaluk: No reason to write this try... except statement in each test file.

ipaserver/plugins/internal.py
```diff
@@ -953,6 +970,175 @@ class i18n_messages(Command):
             "truncated": _("Query returned more results than the configured size limit. Displaying the first ${counter} results."),
             "unselect_all": _("Unselect All"),
         },
+        "ssbrowser": {
+            "ssbrowser-msg": _(
+                "<h1>Browser Kerberos Setup</h1>\n"
+                "\n"
```
serg-cymbaluk: Is it possible to use some markdown? I think there shouldn't be too much HTML-tags in translations.

serg-cymbaluk: Also it is not good idea to use this text as gettext key.
So it can be for example _('SSBrowser-page-content').
And in .po file we can put this conent:
msgid "SSBrowser-page-content"
msgstr "Some raw text with tags..."

ipatests/test_webui/test_translation.py
stanislavlevin: As it was found out 'try/except' is used to satisfy pylint.
But "NO_SELENIUM" is no used variable.
Fixed.

ipatests/test_webui/test_translation.py
stanislavlevin: I use the magic of ui_driver as much as i can :)
But some ( almost all? ) of this magic is related to the IPA web app and is not suited to use it outside of. 

I didn't override 'get_base_url' because it wasn't necessary. There was no benefit.
'get_config_url' is used in method 'load_url', which is not the same as ui_driver's 'load' one.

ipaserver/plugins/internal.py
stanislavlevin: No problem, as you wish.

ipatests/test_webui/test_translation.py
stanislavlevin: Please, take a look at the latest version.

stanislavlevin: Should i make something similar for #2109 ?

serg-cymbaluk: Yes, if there is some common initialization.

stanislavlevin:
Ping.

ipaserver/plugins/internal.py
serg-cymbaluk: About markdown... it can be done later. But still there is needed to make compact gettext key (example is above) and provide translations in po-files.

stanislavlevin: "But still there is needed to make compact gettext key (example is above) and provide translations in po-files."

I don't understand you.
The Gettext key is here already.
After pushing an updated pot file to zanata, translators will fix outdated translations.
Then ipa release person will merge to.

Isn't it?

serg-cymbaluk: Please
> make **compact** gettext key

There are several options how to do that. I have proposed one above.
Other option, you can split ssbrowser-page-content translation into 4 pieces:
* title
* firefox section
* chrome section
* ie section

What you choose is up to you.

stanislavlevin: So, as you proposed I split out long keys.

serg-cymbaluk:
Please rebase your branch to fix Travis checks.

ipaserver/plugins/internal.py
serg-cymbaluk: Ok, thank you, looks good.

stanislavlevin:
Rebased, thank you.

serg-cymbaluk:
master:

* 6f386f2e241294497db18284a1321335afa77a63 Fix translation of "unauthorized.html" Web page
* c66cdf0b601d1ca0f7dd07ff10e8d4bbd7072584 Fix translation of "ssbrowser.html" Web page
* 4b83227ac7428341e6b1c9be78f52b5d4be758a1 Add basic tests to web pages which are located at /ipa/config/

PR #2152, opened 2018-07-13
Teach pylint how our api works

tiran:
pylint 2.0 is more strict and complains about several aspects of
ipalib.api. It turns out that AstroidBuilder.string_build() can be used
to easily teach pylint about object attributes and attribute values.
Although the assignment wouldn't work with the actual implementation,
the string builder assignments shows pylint the names and values of
members. It works without additional transformation.

See: pagure.io/freeipa/issue/7614

### DIFF
--- a/ipalib/config.py
+++ b/ipalib/config.py
@@ -245,10 +245,12 @@ class Env(object):
                 SET_ERROR % (self.__class__.__name__, key, value)
             )
         check_name(key)
+        # pylint: disable=no-member
         if key in self.__d:
             raise AttributeError(OVERRIDE_ERROR %
                 (self.__class__.__name__, key, self.__d[key], value)
             )
+        # pylint: enable=no-member
         assert not hasattr(self, key)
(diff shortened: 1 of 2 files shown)


netoarmando:
Should this address violation `W1306(missing-format-attribute)` in cases [like the following](https://github.com/freeipa/freeipa/blob/master/ipaserver/install/server/replicainstall.py#L875)?
```python
kinit_keytab(
    'host/{env.host}@{env.realm}'.format(env=api.env),
    paths.KRB5_KEYTAB,
    ccache,
)
```

tiran:
@netoarmando yes, it addresses the issue.

netoarmando:
@tiran LGTM

tiran:
master:

* 4fc7f72648df0ae03ac90f701425e297db72e9d6 Teach pylint how our api works
* aacf185ff8eb1bf6b54e0ba893ca4e750cb69564 Add pylint ignore to magic config.Env attributes

PR #2156, opened 2018-07-14
[Backport][ipa-4-6] Fix Pylint 2.0 violations

tiran:
Manual backport of PR #2149 to 4.6 branch

Fix the following violations aiming to support Pylint 2.0

- `unneeded-not` (C0113):
  Consider changing "not item in items" to "item not in items" used
  when a boolean expression contains an unneeded negation.

- `useless-import-alias` (C0414):
  Import alias does not rename original package Used when an import
  alias is same as original package.e.g using import numpy as numpy
  instead of import numpy as np

- `raising-format-tuple` (W0715):
  Exception arguments suggest string formatting might be intended Used
  when passing multiple arguments to an exception constructor, the
  first of them a string literal containing what appears to be
  placeholders intended for formatting

- `bad-continuation` (C0330):
  This was already included on the disable list, although with current
  version of pylint (2.0.0.dev2) violations at the end of the files
  are not being ignored.
  See: https://github.com/PyCQA/pylint/issues/2278

- `try-except-raise` (E0705):
  The except handler raises immediately Used when an except handler
  uses raise as its first or only operator. This is useless because it
  raises back the exception immediately. Remove the raise operator or
  the entire try-except-raise block!

- `consider-using-set-comprehension` (R1718):
  Consider using a set comprehension Although there is nothing
  syntactically wrong with this code, it is hard to read and can be
  simplified to a set comprehension.Also it is faster since you don't
  need to create another transient list

- `dict-keys-not-iterating` (W1655):
  dict.keys referenced when not iterating Used when dict.keys is
  referenced in a non-iterating context (returns an iterator in
  Python 3)

- `comprehension-escape` (W1662):
  Using a variable that was bound inside a comprehension Emitted when
  using a variable, that was bound in a comprehension handler, outside
  of the comprehension itself. On Python 3 these variables will be
  deleted outside of the comprehension.

### DIFF
--- a/ipaclient/plugins/automember.py
+++ b/ipaclient/plugins/automember.py
@@ -28,8 +28,9 @@ register = Registry()
 @register(override=True, no_fail=True)
 class automember_add_condition(MethodOverride):
     has_output_params = (
-        Str('failed',
-        label=_('Failed to add'),
-        flags=['suppress_empty'],
+        Str(
+            'failed',
+            label=_('Failed to add'),
+            flags=['suppress_empty'],
         ),
     )
--- a/ipaclient/plugins/vault.py
+++ b/ipaclient/plugins/vault.py
@@ -661,9 +661,8 @@ class ModVaultData(Local):
 
         name = self.name + '_internal'
         try:
+            # ipalib.errors.NotFound exception can be propagated
             return self.api.Command[name](*args, **options)
-        except errors.NotFound:
-            raise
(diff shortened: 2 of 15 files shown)


tiran:
auto ACK backport patch

tiran:
ipa-4-6:

* a49e933eb6aef3531ce4f2e705d2d525cbfe4329 Fix Pylint 2.0 violations

PR #2160, opened 2018-07-17
Fix KRA replica installation from CA master

tiran:
ipa-replica-install --kra-install can fail when the topology already has
a KRA, but replica is installed from a master with just CA. In that
case, Custodia may pick a machine that doesn't have the KRA auditing and
signing certs in its NSSDB.

Example:
 * master with CA
 * replica1 with CA and KRA
 * new replica gets installed from master

The replica installer now always picks a KRA peer.

The change fixes test scenario TestInstallWithCA1::()::test_replica2_ipa_dns_install

Fixes: https://pagure.io/freeipa/issue/7518

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -1485,7 +1485,10 @@ def install(installer):
     otpd.create_instance('OTPD', config.host_name,
                          ipautil.realm_to_suffix(config.realm_name))
 
-    if ca_enabled:
+    if kra_enabled:
+        # A KRA peer always provides a CA, too.
+        mode = custodiainstance.CustodiaModes.KRA_PEER
+    elif ca_enabled:
         mode = custodiainstance.CustodiaModes.CA_PEER
     else:
         mode = custodiainstance.CustodiaModes.MASTER_PEER


.freeipa-pr-ci.yaml
```diff
@@ -39,206 +43,14 @@ jobs:
...
-  fedora-28/caless:
+  fedora-28/test_replica2_with_ca_kra_install:
     requires: [fedora-28/build]
     priority: 50
     job:
       class: RunPytest
       args:
         build_url: '{fedora-28/build_url}'
-        test_suite: test_integration/test_caless.py::TestServerReplicaCALessToCAFull
+        test_suite: test_integration/test_installation.py::TestInstallWithCA1::()::test_replica2_with_ca_kra_install
```
Rezney: The parenthesis in this pytest definition are some hidden trick or is it a typo?

tiran: I just copy and pasted the test case name. It means that test framework has instantiated the class TestInstallWithCA1 and then executed the case test_replica2_with_ca_kra_install.

PR-CI doesn't like it, because it passes the string through bash without additional quoting. I'm removing it.

ipatests/test_integration/test_installation.py
```diff
@@ -163,7 +163,6 @@ def install(cls, mh):
     def test_replica1_ipa_kra_install(self):
         super(TestInstallWithCA1, self).test_replica1_ipa_kra_install()
 
-    @pytest.mark.xfail(reason="FreeIPA ticket 7008", strict=True)
```
rcritten: Is this related to this patch?

ipaserver/install/server/replicainstall.py
```diff
@@ -1485,7 +1485,10 @@ def install(installer):
     otpd.create_instance('OTPD', config.host_name,
                          ipautil.realm_to_suffix(config.realm_name))
 
-    if ca_enabled:
+    if kra_enabled:
+        # A KRA peer always provides a CA, too.
+        mode = custodiainstance.CustodiaModes.KRA_PEER
+    elif ca_enabled:
         mode = custodiainstance.CustodiaModes.CA_PEER
```
rcritten: This bit LGTM

ipatests/test_integration/test_installation.py
tiran: That should have been in the temp commit. I'll remove it together with the temp commit.

tiran:
Tests pass and the PR is ready for merge.

tiran:
master:

* 530da69eadf5b73e4ca83252e3a370ed70354a39 Fix KRA replica installation from CA master

PR #2163, opened 2018-07-17
Make ipaclient.csrgen optional

tiran:
ipaclient's csrgen plugin has been turned into an optional dependency.
The ipaclient plugin, helper modules like ipaclient.csrgen and templates
are shipped conditionally.

The ipaclient cert plugin and ipatests handle missing csrgen gracefully.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -26,6 +26,13 @@
     %global with_ipatests_option --without-ipatests
 %endif
 
+# Include ipaclient.csrgen feature
+%if 0%{?rhel}
+    %global with_csrgen 0
+%else
+    %{!?with_csrgen:%global with_csrgen 1}
+%endif
+
 # Python 2/3 packages and default Python interpreter
 %if 0%{?rhel} > 7
     %global with_default_python 3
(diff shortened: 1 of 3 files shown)


tiran:
With the patch, all references to csrgen are gone except for the conditional import in ``ipaclient.plugins.cert`` and source file information in egg-info directory. We can safely ignore the latter.

```
$ for F in *.rpm; do rpm2cpio $F | cpio -idmv; done
$ grep -R csrgen etc/ usr/ var/
Binary file usr/lib/python3.6/site-packages/ipaclient/plugins/__pycache__/cert.cpython-36.opt-1.pyc matches
Binary file usr/lib/python3.6/site-packages/ipaclient/plugins/__pycache__/cert.cpython-36.pyc matches
usr/lib/python3.6/site-packages/ipaclient/plugins/cert.py:            # Deferred import, ipaclient.csrgen is expensive to load.
usr/lib/python3.6/site-packages/ipaclient/plugins/cert.py:            from ipaclient import csrgen
usr/lib/python3.6/site-packages/ipaclient/plugins/cert.py:                adaptor = csrgen.NSSAdaptor(database, password_file)
usr/lib/python3.6/site-packages/ipaclient/plugins/cert.py:                adaptor = csrgen.OpenSSLAdaptor(
usr/lib/python3.6/site-packages/ipaclient-4.6.90rc3-py3.6.egg-info/...
```

rcritten:
I don't think we want this change in upstream. It is a downstream decision to disable parts of the code.

abbra:
Original intent was driven by two goals:
- make it easier to generate CSRs for users who struggle with openssl and other libraries
- make it possible to generate CSRs from web UI by calling appropriate APIs.

Admittedly, we still have little improvements on both points but csrgen wasn't used to improve them. So I'm with a split mind here.

PR #2164, opened 2018-07-17
Update 4.7 translations

rcritten:
Sanity-checking only.


rcritten:
@stanislavlevin Can you double-check that this doesn't negatively impact the UI translation work you've been doing?

stanislavlevin:
@rcritten , 
Unfortunately, actual translations are always in follower role:

developer add code -> at some (random? :-) ) point one updates po template in zanata -> translator translates -> release person adds translations...

_In case of approving of translation patch sets somebody have to update pot again_. But i continue to work on improvement of translations .

This PR doesn't impact on my work.
The updating of po files is good deal. So far the translation of UI is not perfect enough.

tiran:
master:

* 759e8355c8b0e4c665e038f3e5bc44b0ca9e279d Update 4.7 translations

PR #2165, opened 2018-07-17
Fix regression: Handle unicode where str is expected

netoarmando:
Regression caused by 947ac4bc1f6f4016cf5baf2ecb4577e893bc3948 when trying to fix a similar issue for clients running Python 3. However, that fix broke Python 2 clients.

Issue: https://pagure.io/freeipa/issue/7626


tiran:
master:

* 717d59e2fe3cc0318b50f4ee51f4a4e0b963d31d Fix regression: Handle unicode where str is expected

PR #2173, opened 2018-07-20
Fix $-style format string in ipa_ldap_init (util/ipa_ldap.c)

t-woerner:
The second argument was not used, but the first one was used twice.


rcritten:
master:

* f3faecbbb49704c182b310fb43cec9171c08ccf7 Fix $-style format string in ipa_ldap_init (util/ipa_ldap.c)

PR #2181, opened 2018-07-24
In IPA 4.4 when updating userpassword with ldapmodify does not update…

tbordaz:
… krbPasswordExpiration nor krbLastPwdChange

When making ipa-pwd-extop TXN aware, some callbacks are call twice.
Particularily
	ipapwd_pre_add is called during PRE_ADD and TXN_PRE_ADD
	ipapwd_pre_mod is called during PRE_MOD and TXN_PRE_MOD
	ipapwd_post_modadd is called during POST_ADD and TXN_POST_ADD
	ipapwd_post_modadd is called during POST_MOD and TXN_POST_MOD
It is not the expected behavior and it results on some skipped updates krbPasswordExpiration
and krbLastPwdChange

https://pagure.io/freeipa/issue/7601

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
+++ b/daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
@@ -1488,8 +1488,6 @@ int ipapwd_pre_init(Slapi_PBlock *pb)
     ret = slapi_pblock_set(pb, SLAPI_PLUGIN_VERSION, SLAPI_PLUGIN_VERSION_01);
     if (!ret) ret = slapi_pblock_set(pb, SLAPI_PLUGIN_DESCRIPTION, (void *)&ipapwd_plugin_desc);
     if (!ret) ret = slapi_pblock_set(pb, SLAPI_PLUGIN_PRE_BIND_FN, (void *)ipapwd_pre_bind);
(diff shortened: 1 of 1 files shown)


rcritten:
This looks ok. Do we have any practical ways to test this to prevent regressions?

There is an integration test, ipatests/test_integration/test_commands.py, that does some ldappasswd changes. Can this be extended to also do other password change types to exercise this, and to check LDAP attributes afterwards?

tbordaz:
Thanks @rcritten for the review

Yes it should be easily tested with ldappasswd and ldapmodify(userpassword) and checking that on successful update krbPasswordExpiration and krbLastPwdChange are updated.
CUrrently the test should fail in 4.x

flo-renaud:
Tests added in PR #2200.

I tested this PR and it works for me (install ipa server, create user, modify user password, read krblastpwdchange and krbpasswordexpiration, ldapmodify userpassword as dir mgr and check that the krblastpwdchange/krbpasswordexpiration attributes have been updated). Hence giving ACK.

Tiboris:
master:

* a71729cc01b83916b5e362a38ce3d65e129b1815 In IPA 4.4 when updating userpassword with ldapmodify does not update krbPasswordExpiration nor krbLastPwdChange

PR #2182, opened 2018-07-24
ipaserver/plugins/cert.py: Added reason to raise of errors.NotFound

t-woerner:
In the case that enabledService is not found ipaConfigString kdc entry, a
NotFound error was raised without setting the reason. This resulted in a
traceback.


rcritten:
One last request. Can you create a pagure ticket to link this to? This will make it easier to ensure it lands as a patch in all the necessary downstream releases.

t-woerner:
@rcritten Here is the pagure ticket: https://pagure.io/freeipa/issue/7652

rcritten:
This was not merged properly. The merge button was is not to be used as github is not the upstream of the IPA source.

You'll need to resubmit this PR. Please include the pagure ticket in the (single) commit message.

PR #2185, opened 2018-07-26
Do not set ca_host when --setup-ca is used.

Tiboris:
Setting ca_host caused replication failures on DL0
because it was trying to connect to wrong CA host.
Trying to avoid corner-case in ipaserver/plugins/dogtag.py
when api.env.host nor api.env.ca_host had not CA configured
and there was ca_host set to api.env.ca_host variable.

Reverts: https://github.com/freeipa/freeipa/pull/2048
Resolves: https://pagure.io/freeipa/issue/7629

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -461,6 +461,11 @@ class CAInstance(DogtagInstance):
                 self.step("updating IPA configuration", update_ipa_conf)
                 self.step("enabling CA instance", self.__enable_instance)
                 if not promote:
+                    if self.clone:
+                        # DL0 workaround; see docstring of __expose_ca_in_ldap
+                        self.step("exposing CA instance on LDAP",
+                                  self.__expose_ca_in_ldap)
+
                     self.step("migrating certificate profiles to LDAP",
                               migrate_profiles_to_ldap)
                     self.step("importing IPA certificate profiles",
@@ -1276,6 +1281,25 @@ class CAInstance(DogtagInstance):
         else:
             config = []
         self.ldap_configure('CA', self.fqdn, None, basedn, config)
+
+    def __expose_ca_in_ldap(self):
+        """
+        In a case when replica is created on DL0 we need to make
+        sure that query for CA service record of this replica in
+        ldap will succeed in time of installation.
+        This method is needed for sucessfull replica installation
+        on DL0 and should be removed alongside with code for DL0.
+
+        To suppress deprecation warning message this method is
+        not invoking ldap_enable() but _ldap_enable() method.
+        """
+
+        basedn = ipautil.realm_to_suffix(self.realm)
+        if not self.clone:
+            config = ['caRenewalMaster']
+        else:
+            config = []
+        self._ldap_enable(u'enabledService', "CA", self.fqdn, basedn, config)
 
     def setup_lightweight_ca_key_retrieval(self):
         if sysupgrade.get_upgrade_state('dogtag', 'setup_lwca_key_retrieval'):
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -241,9 +241,12 @@ def create_ipa_conf(fstore, config, ca_enabled, master=None):
         gopts.extend([
             ipaconf.setOption('enable_ra', 'True'),
             ipaconf.setOption('ra_plugin', 'dogtag'),
-            ipaconf.setOption('dogtag_version', '10'),
-            ipaconf.setOption('ca_host', config.ca_host_name)
+            ipaconf.setOption('dogtag_version', '10')
         ])
+
+        if not config.setup_ca:
+            gopts.append(ipaconf.setOption('ca_host', config.ca_host_name))
+
     else:
         gopts.extend([
             ipaconf.setOption('enable_ra', 'False'),


rcritten:
I think it might be better to just revert the other patch completely.

Baking ca_host into default.conf is likely not the right thing to do. It means that some host is ALWAYS ca_host for the lifetime of the installation, or until someone changes the value. If that host went away at some point then things would break in mysterious ways.

I wonder what @abbra and @tiran think.

Tiboris:
Thanks @rcritten, 

noted and I will change it to only revert mentioned PR.
Let us then wait for others to give their opinion :)

abbra:
@Tiboris, I'd prefer your patch to reverting the origina because it is actually makes the code clearer.
Could you please remove "Reverts #2048" from the commit message?

abbra:
@rcritten if we want to avoid baking ca_host into `default.conf`, we have to change and update certmonger to provide an alternative. I don't think it is a viable solution either.

rcritten:
@abbra ca_host isn't used by certmonger.

By basically hardcoding ca_host this would mean that even when the local host has a CA it would always forward all CA-related requests to that other host.

@tiboris can you add the analysis to the ticket that has been done related to the original patch which causes DL0 installations to fail (or is it specifically DL0 connecting between 3.x and 4.x?)

I'm assuming that the workaround of only setting ca_host when not setting up a local CA is related to my above assertion.

Tiboris:
Hello @rcritten here is cited comment:
> I have tried using remote ports of ipa-3.0 master including:
9443
9444
9445
9446
using uri to rest api login and all attempts failed with 404.

> Method dogtag.https_request() was running requests against replica's hostname.
All requests and installation in the end finished successfuly when i used replica's hostname in step [23/27]: migrating certificate profiles to LDAP of configuring pki in dogtag.enter(self) method. I was struggling with finding a cause for it and Standa helped https://bugzilla.redhat.com/show_bug.cgi?id=1598662#c7 

https://pagure.io/freeipa/issue/7629#comment-523842

abbra:
@rcritten ah, right, so this is a local optimization to make sure which CA we contact.
Perhaps, by the end of install we can remove `ca_host`?

rcritten:
It would need to be removed once the local CA is up. This is needed because otherwise the dogtag version is set to 10 (because the local master is that version) but the remote version may not be which will cause the wrong APIs to be used (the installer will use the REST api which isn't available in dogtag 9).

rcritten:
I think we need @tiran input on this to know where ca_host needs to be set in an install.

I'd be inclined to add a ca_host parameter to create_ipa_conf() and set it only when needed and unset as soon as it is not but I don't know what that range is.

flo-renaud:
@rcritten:
with the current proposal, ca_host is set in /etc/ipa/default.conf only if the replica does not provide the CA service. The behavior is similar to what existed before @tiran's [patch](https://pagure.io/freeipa/c/8c3ff0308c95793a6809b8e0a3ed2a145ad3c8ea?branch=master) (Always set CA host when installing replica).
I agree that hard-coding a specific CA is not a good solution but it is a different issue that has been present for a long time. Maybe we should open a different ticket for this problem and handle it separately.
@Tiboris 
the patch does not fix the issue with DL0 replica installation. When the installer needs to contact Dogtag, it calls ca_host method to find which master is providing the CA service. With your patch, it will first try with the replica, but since the [patch](https://pagure.io/freeipa/c/7284097eedef70dd556270732e6ab8e23501ce09?branch=master) 'Delay enabling services until end of installer', the ipaConfig string does not contain enabledService at this point (the service is configured but not enabled). This means that the ca_host will be found using select_any_master and pick the only CA instance which is the ipa 3.0 master, and we still see the same issue.
A solution could be to set enabledService before this call to Dogtag, but I guess it will re-open a time window where the CA service is advertised as enabled on this replica (and may lead to concurrent replica installation issues). @tiran , what is your opinion on this? Did the concurrent replica installation issues also happen in DL0?

ipaserver/install/cainstance.py
```diff
@@ -1277,6 +1279,15 @@ def __enable_instance(self):
             config = []
         self.ldap_configure('CA', self.fqdn, None, basedn, config)
 
+
+    def __expose_ca_in_ldap(self):
```
tiran: Could you please add a comment that this is a hack for DL0 and will be removed once we remove DL0 code?

tiran:
The code looks good. I'll give an ACK after you have updated the function's doc string.

ipaserver/install/cainstance.py
Tiboris: Sure

ipaserver/install/cainstance.py
```diff
@@ -461,6 +461,8 @@ def configure_instance(self, host_name, dm_password, admin_password,
                 self.step("updating IPA configuration", update_ipa_conf)
                 self.step("enabling CA instance", self.__enable_instance)
                 if not promote:
+                    self.step("exposing CA instance on LDAP",
```
flo-renaud: I think we need to call this step only when installing a replica (otherwise it will also get executed for first master installation). Not sure which variable to check, though...

Tiboris: Ok, @tiran I removed ACK and added WIP.
Let us try to figure out something.

tiran: There is only one possible path to run into the situation: Install a primary master in DL0 mode. But DL0 installation is deprecated and only used in testing and not supported.

flo-renaud: Even in DL1 the master install will go through this code path. I just tried to install a master in DL1 and was able to see in the output:
```
  [23/29]: enabling CA instance
  [24/29]: exposing CA instance on LDAP
/usr/lib/python3.6/site-packages/ipaserver/install/cainstance.py:1297: DeprecationWarning: ldap_enable is deprecated, use ldap_configure instead.
  self.ldap_enable('CA', self.fqdn, None, basedn, config)
  [25/29]: migrating certificate profiles to LDAP
```

flo-renaud:
Adding re-run because fedora-28/test_forced_client_enrolment  failed

flo-renaud:
Hi @Tiboris 
changes look good to me.
I tested server + replica installation with --setup-ca in domain-level 1, works well, does not go through the "exposing CA instance on LDAP" code path.
Then to test on DL0, I choose to cherry-pick the change in ipa-4-6 (the master branch does not support any more the DL0 replica installation). I installed an IPA master 3.0, then installed the replica with your patch in domain level 0 with --setup-ca, works well. So unless you plan other modifications, I am ok with the code changes.

Tiboris:
master:

* 6175672e8e11a5fb0a813ea11513efffb704a672 Do not set ca_host when --setup-ca is used

PR #2189, opened 2018-07-27
Honor no-host-dns when creating client host in replica install

rcritten:
--no-host-dns is supposed to avoid all DNS lookups so pass
this as the force value when creating the host in a replica
installation.

https://pagure.io/freeipa/issue/7656


Tiboris:
Works for me, 
thanks for the patch @rcritten.

tiran:
master:

* df8bffd9acb01fb8c355dd0c11a351d6f524aac0 Honor no-host-dns when creating client host in replica install

PR #2190, opened 2018-07-27
Convert members into types in sudorule-*-option

rcritten:
The indirect members need to be calculated and the member
attributes converted. This is normally done in
baseldap::LDAPRetrieve but these methods provide their
own execute() in order to handle the option values.

Update sudorule_add|remove_option tests to include check
that converted user/group exists in the proper format.

https://pagure.io/freeipa/issue/7649


Tiboris:
Works as described, thanks for the patch!

serg-cymbaluk:
master:

* f68dca24be11b7326dd221d3b5a0a4e366e0358f Convert members into types in sudorule-*-option

PR #2194, opened 2018-07-31
Check if user permssions and umask 0022 is set after ipa-restore

mrizwan93:
This test checks if the access rights for user/group
is set to 644 on /var/lib/dirsrv/slapd-TESTRELM-TEST/ldif/*
and umask 0022 set while restoring.

related ticket: https://pagure.io/freeipa/issue/6844


rcritten:
@tiran may have a different opinion but the use of a global doesn't feel right here. You can achieve the same thing and set it locally by overriding the install method like:
```
    @classmethod
    def install(cls, mh):
        super(TestUserrootFilesOwnershipPermission, cls).install(mh)
        cls.bashrc_file = cls.master.get_file_contents('/root/.bashrc')
```

And then use self.bashrc_file in the test and teardown methods.

ipatests/prci_definitions/gating.yaml
```diff
@@ -27,218 +23,14 @@ jobs:
...
-  fedora-28/dnssec:
+  fedora-28/backup_restore:
     requires: [fedora-28/build]
     priority: 50
     job:
       class: RunPytest
       args:
         build_url: '{fedora-28/build_url}'
-        test_suite: test_integration/test_dnssec.py::TestInstallDNSSECFirst
+        test_suite: test_integration/test_backup_and_restore.py::TestUserrootFilesOwnershipPermission
```
flo-renaud: Hi @mrizwan93 
the test is failing because of a typo. The test suite in this yaml file should be TestUser**R**ootFilesOwnershipPermission with uppercase R.

ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -37,6 +38,7 @@
 
 logger = logging.getLogger(__name__)
 
+bashrc_file = None
```
flo-renaud: Is this variable still used?

rcritten:
Looks good.

Tiboris:
@mrizwan93 please remove temporary commit.

Tiboris:
Ok I removed temporary commit.

Tiboris:
Hello @mrizwan93 please create manual backport to branch ipa-4-6. You will need to resolve some conflicts probably.

PR #2199, opened 2018-08-02
Rename pytest_plugins to ipatests.pytest_ipa

tiran:
pytest 3.7.0 doesn't like ipatests.pytest_plugins package. The string
"pytest_plugins" is used as marker to load plugins. By populare vote and
to avoid future conflicts, we decided to rename the directory to pytest_ipa.

Fixes: https://pagure.io/freeipa/issue/7663


Tiboris:
Thanks for the fix @tiran.

Tiboris:
master:

* 64145093434e2601b8352e9ce95b3316d7ac524d Rename pytest_plugins to ipatests.pytest_ipa

PR #2200, opened 2018-08-02
Tests: add integration test for 7601

flo-renaud:
Add a test for issue 7601:
- add a user, perform kinit user to modify the password, read krblastpwdchange
and krbpasswordexpiration.
- perform a ldapmodify on the password as dir mgr
- make sure that krblastpwdchange and krbpasswordexpiration have been modified

Related to:
https://pagure.io/freeipa/issue/7601

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -6,9 +6,11 @@
 from __future__ import absolute_import
 
 import base64
+import re
 import ssl
 from tempfile import NamedTemporaryFile
 import textwrap
+import time
 
 from ipaplatform.paths import paths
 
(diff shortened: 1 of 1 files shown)


flo-renaud:
Note: for the test to be successful, PR #2181 is needed.

flo-renaud:
Adding labels 4.5, 4.6 and 4.7 as the bug fix is proposed for these versions.

ipatests/test_integration/test_commands.py
```diff
@@ -134,6 +135,75 @@ def test_change_sysaccount_password_issue7561(self):
...
+        # Retrieve krblastpwdchange and krbpasswordexpiration
+        output = master.run_command(search_cmd).stdout_text.lower()
+        # extract krblastpwdchange and krbpasswordexpiration
+        newkrblastpwdchange = re.findall(krbchg_pattern, output)[0]
+        newkrbexp = re.findall(krbexp_pattern, output)[0]
+
+        # both should have changed
+        assert newkrblastpwdchange != krblastpwdchange
+        assert newkrbexp != krbexp
+
```
tbordaz: The patch looks good. Would it be possible to run an additional similar test (create a user + update userpasswd + retrieve lastpwdchange+expiration) but using ldappasswd (instead of ldapmodify) and check values have changed.
Basically we want DM to be able to user ldapmodify or ldapmodify and get the changes of lastpwdchange+expiration

flo-renaud: @tbordaz 
thank you for your comments. I updated the test with your suggestion, and also added a sleep before modification of the password (the krb* attributes have at most 1sec precision).

flo-renaud:
Rebased to include PR #2181

rcritten:
I have just a nit from the commit message. Can you make the first line more meaningful than mentioning the ticket number?

Maybe something like Tests: add integration test for password changes

flo-renaud:
Commit msg updated as suggested.

flo-renaud:
As @tbordaz provided the ACK in a comment and @rcritten concern was addressed, I am adding the ACK label.

Tiboris:
master:

* c1a0c3c556e936077e6d5b386621a36113af86fe Tests: add integration test for password changes by dir mgr

PR #2214, opened 2018-08-06
Generate CSR dynamically for UI tests

mrizwan93:
There were hardcoded csr in test_cert.py. 'host_csr_path'
and 'service_csr_path' needs to be generated before executing
UI tests. This fix removes the hardcoded values as well as need
of generating csr before executing UI testsuit.

### DIFF
--- a/ipatests/test_webui/test_service.py
+++ b/ipatests/test_webui/test_service.py
@@ -56,7 +56,8 @@ class sevice_tasks(UI_driver):
         }
 
     def load_file(self, path):
-        # ENHANCEMENT: generate csr dynamically
+        hostname = self.config.get('ipa_server')
+        self.generate_csr(hostname, path)
         with open(path, 'r') as file_d:
             content = file_d.read()
         return content
--- a/ipatests/test_webui/ui_driver.py
+++ b/ipatests/test_webui/ui_driver.py
@@ -31,6 +31,12 @@ import os
 from functools import wraps
 import unittest
 import paramiko
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization, hashes
+from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography import x509
+from cryptography.x509.oid import NameOID
+
 
 # pylint: disable=import-error
 from six.moves.urllib.error import URLError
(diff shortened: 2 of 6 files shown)


ipatests/test_webui/ui_driver.py
```diff
@@ -174,6 +175,15 @@ def load_config(self):
         if 'type' not in c:
             c['type'] = DEFAULT_TYPE
 
+    def generate_csr(self, cn, csr_file):
```
Rezney: Hi @mrizwan93 thanks for your patch. I would recommend to use python-cryptography for this.

Please refer to :
https://cryptography.io/en/latest/x509/tutorial/

Then you can create a module which could be re-usable for all the tests in the future.

mrizwan93: @Rezney 
Thanks for the review. I am trying to understand if python-cryptography have advantage over using openssl. 
In case of openssl we are not importing any package to the code but for python-cryptography we need to import some of them.
Thanks for the tutorial as well, it will help to replace the code if we agrees upon it.

ipatests/test_webui/test_service.py
```diff
@@ -56,7 +56,7 @@ def prep_data(self):
         }
 
     def load_file(self, path):
-        # ENHANCEMENT: generate csr dynamically
+        self.generate_csr(hostname, path)
```
Tiboris: You did mistake here.

ipatests/test_webui/ui_driver.py
Rezney: No worries, it is your call. Just suggestion. Not blocker...

Tiboris: Hello @mrizwan93. I think that it would be better NOT to use some sub-processes as you do there but to use available libraries and create modul as @Rezney proposed. What do you think @flo-renaud ?

mrizwan93: replacing subprocess makes sense. Thanks @Rezney @Tiboris

ipatests/test_webui/ui_driver.py
```diff
@@ -174,6 +180,35 @@ def load_config(self):
...
+        """ generate a csr with cn and store in csr_file"""
+        key = rsa.generate_private_key(
+            public_exponent=65537,
+            key_size=2048,
+            backend=default_backend())
+
+        with open("/root/key.pem", "wb") as f:
```
tiran: You cannot assume that you can write to ``/root``. I'd change the function to return the key and the CSR as PEM-encoded bytes.

mrizwan93:
Since the issues have been fixed in https://github.com/freeipa/freeipa/pull/2369 . closing this PR.

PR #2222, opened 2018-08-07
Re-open the ldif file to prevent error message

Tiboris:
There was an issue with ipa-server-upgrade and it was
showing an error while upgrading:
DN... does not exists or haven't been updated.

Resolves: https://pagure.io/freeipa/issue/7644

### DIFF
--- a/ipaserver/install/upgradeinstance.py
+++ b/ipaserver/install/upgradeinstance.py
@@ -237,17 +237,22 @@ class IPAUpgrade(service.Service):
 
     def __disable_schema_compat(self):
         ldif_outfile = "%s.modified.out" % self.filename
+
+        with open(self.filename, "r") as in_file:
+            parser = GetEntryFromLDIF(in_file, entries_dn=[COMPAT_DN])
+            parser.parse()
+
+        try:
+            compat_entry = parser.get_results()[COMPAT_DN]
+        except KeyError:
+            return
+
+        if not compat_entry.get('nsslapd-pluginEnabled'):
+            return
+
         with open(ldif_outfile, "w") as out_file:
             with open(self.filename, "r") as in_file:
-                parser = GetEntryFromLDIF(in_file, entries_dn=[COMPAT_DN])
-                parser.parse()
-                try:
-                    compat_entry = parser.get_results()[COMPAT_DN]
-                except KeyError:
-                    return
                 parser = installutils.ModifyLDIF(in_file, out_file)
-                if not compat_entry.get('nsslapd-pluginEnabled'):
-                    return
                 parser.remove_value(COMPAT_DN, "nsslapd-pluginEnabled")
                 parser.remove_value(COMPAT_DN, "nsslapd-pluginenabled")
                 parser.add_value(COMPAT_DN, "nsslapd-pluginEnabled",
--- a/ipatests/test_integration/test_upgrade.py
+++ b/ipatests/test_integration/test_upgrade.py
@@ -18,4 +18,6 @@ class TestUpgrade(IntegrationTest):
     def test_invoke_upgrader(self):
         cmd = self.master.run_command(['ipa-server-upgrade'],
                                       raiseonerr=False)
+        assert ("DN: cn=Schema Compatibility,cn=plugins,cn=config does not \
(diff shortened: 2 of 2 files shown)


ipaserver/install/upgradeinstance.py
```diff
@@ -241,6 +241,9 @@ def __disable_schema_compat(self):
             with open(self.filename, "r") as in_file:
                 parser = GetEntryFromLDIF(in_file, entries_dn=[COMPAT_DN])
                 parser.parse()
+
+                # move to file begining because parser is reading same file
```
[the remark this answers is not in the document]
Tiboris: Of course, thanks for explanation.

flo-renaud:
Hi @Tiboris 
thank you for the patch, works for me. I tested with fresh install + ipa-server-upgrade and the error message was not present any more.
As there is already a test case in `ipatests/test_integration/test_upgrade.py`, you can easily add enhance the existing test by checking that the string `DN: cn=Schema Compatibility,cn=plugins,cn=config does not exists or haven't been updated` is not in cmd.stdout_text (or result.stderr_text).

Tiboris:
Thanks @flo-renaud will do.

tiran:
IMHO the method would benefit from some refactorization. For example it creates ``ldif_outfile`` even if ``COMPAT_DN`` is missing from the infile.

```
        with open(self.filename, "r") as in_file:
            parser = GetEntryFromLDIF(in_file, entries_dn=[COMPAT_DN])
            parser.parse()

        try:
            compat_entry = parser.get_results()[COMPAT_DN]
        except KeyError:
            return

        with open(ldif_outfile, "w") as out_file:
            with open(self.filename, "r") as in_file:
                parser = installutils.ModifyLDIF(in_file, out_file)
                ...
```

Tiboris:
Thanks for the tip @tiran!

ipaserver/install/upgradeinstance.py
```diff
@@ -237,14 +237,18 @@ def __disable_listeners(self):
...
         with open(ldif_outfile, "w") as out_file:
             with open(self.filename, "r") as in_file:
-                parser = GetEntryFromLDIF(in_file, entries_dn=[COMPAT_DN])
-                parser.parse()
-                try:
-                    compat_entry = parser.get_results()[COMPAT_DN]
-                except KeyError:
-                    return
                 parser = installutils.ModifyLDIF(in_file, out_file)
                 if not compat_entry.get('nsslapd-pluginEnabled'):
```
flo-renaud: You can probably move this check before open(ldif_outfile,...). This way, if the input file does not contain the entry nssldap-pluginEnabled, the outfile doesn't get created and we exit quickly.

Tiboris: I will try!

Tiboris: Thanks @flo-renaud, 
ipa-server-upgrade is working for me with this change.

rcritten:
You can fix this as a one-liner with:

in_file.seek(0) before the second invocation of parser = installutils.ModifyLDIF(in_file, out_file)

Tiboris:
Hello @rcritten,

thanks for input. Original pull request used your approach. On the other side I agree with @tiran, as he showed me that this code could look better and this PR created a opportunity to do so. Yes it is maybe opening same file twice and maybe this method could have been more lightweight calling `open()` only once. Idea of improving code that it does not create any artifacts got me but I will not insist on the shape of PR as it is.

rcritten:
Ok, it LGTM.

Tiboris:
Maybe some changes in `ModifyLDIF()` could create better image and resolve this issue with re-reading file.

rcritten:
IMHO it is fine as-is. This is only executed during installs and on upgrades so re-reading the file has limited impact.

flo-renaud:
Hi @Tiboris 
thanks for the patch and the test, works for me.
Tested by fresh install + ipa-server-upgrade, the msg does not appear any more.

flo-renaud:
master:

* 6fa1e6f18ef1798fa8cd5030807c81699fdfbdc6 Re-open the ldif file to prevent error message
* 89799a14ce674f73bdfb8310256e7c8c67866a34 Add assert to check output of upgrade

PR #2228, opened 2018-08-08
do not use RC4 in FIPS mode

abbra:
ipasam: do not use RC4 in FIPS mode
    
    When creating Kerberos keys for trusted domain object account, ipasam
    module requests to generate keys using a series of well-known encryption
    types. In FIPS mode it is not possible to generate RC4-HMAC key:
    MIT Kerberos is using openssl crypto backend and openssl does not allow
    use of RC4 in FIPS mode.
    
    Thus, we have to filter out RC4-HMAC encryption type when running in
    FIPS mode. A side-effect is that a trust to Active Directory running
    with Windows Server 2003 will not be possible anymore in FIPS mode.
    
    Resolves: https://pagure.io/freeipa/issue/7659

### DIFF
--- a/daemons/ipa-sam/ipa_sam.c
+++ b/daemons/ipa-sam/ipa_sam.c
@@ -213,6 +213,7 @@ struct ipasam_private {
 	char *client_princ;
 	struct sss_idmap_ctx *idmap_ctx;
 	uint32_t supported_enctypes;
+	bool fips_enabled;
 };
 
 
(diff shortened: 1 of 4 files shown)


Tiboris:
Giving ACK due to @frozencemetery's approval.

Tiboris:
master:

* de8f969f2d40722b590f43ab9bb31eada58ec4b3 Move fips_enabled to a common library to share across different plugins
* 6907a0cef7f22293c16df17aa486f7ec2d8a0899 ipasam: do not use RC4 in FIPS mode

PR #2234, opened 2018-08-09
Disable DL0 specific tests

tiran:
Disable tests that use domain level 0. Fail early to catch additional
tests that depend on DL0.

See: https://pagure.io/freeipa/issue/7669

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -32,6 +32,7 @@ import time
 
 import dns
 from ldif import LDIFWriter
+import pytest
 from SSSDConfig import SSSDConfig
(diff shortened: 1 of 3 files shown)


t-woerner:
What do you think about using MIN_DOMAIN_LEVEL instead of DOMAIN_LEVEL_0 to make this also work automatically if we move to domain level 2 at some point. It might also be good to test against max domain level.
If this should be independent of MIN/MAX_DOMAIN_LEVEL in constants then how about adding pytest global settings that could be used and adapted easily without the need to change all this again.

t-woerner:
master:

* 13000e2f19401549ec3797b071671087d7fc37e6 Disable DL0 specific tests

PR #2235, opened 2018-08-09
ipaclient: Remove --no-sssd and --noac options

t-woerner:
Client installation with --no-sssd option has already beeen deprecated
with https://pagure.io/freeipa/issue/5860. Authconfig support has been
removed, therefore --no-ac option can be removed also.

ipatests/test_integration/test_authselect.py: Skip no_sssd and no_ac tests.

See: https://pagure.io/freeipa/issue/7671

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -3709,11 +3709,7 @@ class ClientInstall(ClientInstallInterface,
     def prompt_password(self):
         return self.interactive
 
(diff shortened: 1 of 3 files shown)


tiran:
master:

* c5cdd5a5f01306b3a70354d34079efe64565aa69 ipaclient: Remove --no-sssd and --no-ac options

PR #2237, opened 2018-08-10
Replace logo images with new one (version 4.7)

serg-cymbaluk:
Replace logo images with new one (version 4.7)

Resolves: https://pagure.io/freeipa/issue/7362
new-login-screen


Tiboris:
Finally!

Tiboris:
master:

* c2e1cdf8a3b9330018c4d6ea2f32c23be5fb4c28 Replace logo images with new one (version 4.7)

pvoborni:
One note unreleated to this PR. Looking at other images, there is also login-screen-logo.png which is defined in CSS(less) styles. It's hidden by default upstream - so I think we can safely remove it. But some other distributions like RHEL still use it (with different image) in their branding patches.

Tiboris:
So downstream patches will not be affected critically by removal?

pvoborni:
Or maybe disregard my coment about safe removal. I see that ''images/login-screen-logo.png'' it is defined in LoginScreenBase widget so just removal of the file might cause some issues, though they will be unvisible   (error in log).

Tiboris:
Thank you @pvoborni.
I will ask @serg-cymbaluk to look at it.

serg-cymbaluk:
@pvoborni 

> Or maybe disregard my coment about safe removal. I see that ''images/login-screen-logo.png'' it is defined in LoginScreenBase widget so just removal of the file might cause some issues, though they will be unvisible (error in log).

I will check how and where LoginScreenBase uses this logo and, in case the widgets really shows it somewhere, I will replace it.

PR #2241, opened 2018-08-11
ipa-otptoken-import: support import from CSV file

abbra:
Allow import from a CSV file where fields have a fixed meaning:

"serial", "key", "type", "digits", "interval"

Fixes: https://pagure.io/freeipa/issue/7673

Pair-programmed-with: Nathaniel McCallum <npmccallum@redhat.com>


rcritten:
I think you should include a test with one or more invalid CSV.

Did you look at the python built-in csv module for doing the parsing?

abbra:
Thanks, I'll look at using a built-in csv module.

abbra:
I've updated the code to use csv parser and added a test to make sure we test reading invalid tokens. Note that I do not test reading invalid CSV format but there are checks for the possible exceptions thrown by that.

ipaserver/install/ipa_otptoken_import.py
```diff
@@ -506,6 +508,92 @@ def save(self, dest):
...
+        # Always return None as we have no key names in CSV format
+        return None
+
+    def __init__(self, filename):
+        self.__keypackages = []
+
+        with open(filename, "r") as f:
+            reader = csv.reader(f)
+            for row in reader:
+                try:
```
flo-renaud: You may want to add a check on the number of items in each row (the code is expecting 5), otherwise it is possible to add a CSVToken that is missing some elements.

ipaserver/install/ipa_otptoken_import.py
```diff
@@ -506,6 +508,95 @@ def save(self, dest):
...
+        ('interval', 'ipatokentotptimestep', lambda v, o: v),
+    )
+
+    @property
+    def keyname(self):
+        # Always return None as we have no key names in CSV format
+        return None
+
+    def __init__(self, filename):
+        self.__keypackages = []
```
tiran: Please don't use double under. Private attribute use a single under. Double under use additional name mangling magic.

ipaserver/install/ipa_otptoken_import.py
```diff
@@ -506,6 +508,95 @@ def save(self, dest):
...
+
+        with open(filename, "r") as f:
+            reader = csv.reader(f)
```
tiran: You can use a DictReader https://docs.python.org/3/library/csv.html#csv.DictReader to make the code more readable and stable. A dict reader will map undefined fields to None, so be prepared to raise an error when you see None.

ipaserver/install/ipa_otptoken_import.py
```diff
@@ -506,6 +508,95 @@ def save(self, dest):
...
+            for row in reader:
+                if len(row) != len(self._FIELDS):
+                    raise ValidationError(
+                        'Provided token has incorrect number of fields')
+                try:
+                    vals = dict(zip(self._FIELDS, row))
```
tiran: This isn't necessary with a dict reader.

ipaserver/install/ipa_otptoken_import.py
```diff
@@ -506,6 +508,95 @@ def save(self, dest):
...
+
+    @property
+    def options(self):
+        return self.__options
+
+    @property
+    def removed(self):
+        return self.__removed
+
+    def __init__(self, id, options):
```
tiran: ``__init__`` should be defined at the top of the class definition.

ipaserver/install/ipa_otptoken_import.py
```diff
@@ -506,6 +508,95 @@ def save(self, dest):
...
+        ('digits', 'ipatokenotpdigits', lambda v, o: v),
+        ('interval', 'ipatokentotptimestep', lambda v, o: v),
+    )
+
+    @property
+    def keyname(self):
+        # Always return None as we have no key names in CSV format
+        return None
+
+    def __init__(self, filename):
```
tiran: Please move ``__init__`` before ``keyname``.

ipaserver/install/ipa_otptoken_import.py
```diff
@@ -506,6 +508,95 @@ def save(self, dest):
...
+                        CSVToken(options['ipatokenserial'], options))
+                except csv.Error as e:
+                    raise ValidationError(
+                        'Parsing CSV failed: {e}'.format(e=str(e)))
+                except (TypeError, binascii.Error) as e:
+                    raise ValidationError(
+                        'Provided token key is not properly encoded: {e}'.
+                        format(e=str(e)))
+
+    def getKeyPackages(self):
```
tiran: PEP8 -- ``get_key_packages``

ipaserver/install/ipa_otptoken_import.py
```diff
@@ -506,6 +508,95 @@ def save(self, dest):
...
+        for kp in self.__keypackages:
+            yield kp
+
+    def save(self, dest):
+        with open(dest, "w") as f:
+            writer = csv.writer(f)
```
tiran: A DictWriter also simplified this code a lot.

ipaserver/install/ipa_otptoken_import.py
```diff
@@ -506,6 +508,95 @@ def save(self, dest):
...
+        self.__id = id
+        self.__options = options
+        self.__removed = False
+
+    def remove(self):
+        self.__removed = True
+
+
+class CSVDocument(object):
+    _FIELDS = ("serial", "key", "type", "digits", "interval")
```
tiran: Why did you define different field names from IPA field names? This trick should make the code faster, smaller, and easier to understand:

```
_MAP = collections.OrderedDict([
    ('ipatokenserial', lambda v, o: v.strip()),
    ...
])

with open(...) as f:
    reader = csv.DictReader(f, list(self._MAP), restval='')
    for row in reader:
        for k, v in row.items():
            row[k] = self._MAP[k](v, options)
```

abbra:
I still want this to be completed but had no time yet.

PR #2246, opened 2018-08-13
Test if WSGI worker process count is set to 4

mrizwan93:
Test if WSGI worker process count is set to 4 in  /etc/httpd/conf.d/ipa.conf

cat /etc/httpd/conf.d/ipa.conf | grep processes
WSGIDaemonProcess ipa processes=4 threads=1 maximum-requests=500 \

related ticket : https://pagure.io/freeipa/issue/7587

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -12,6 +12,7 @@ from __future__ import absolute_import
 import os
 import pytest
 from ipalib.constants import DOMAIN_LEVEL_0
+from ipaplatform.constants import constants
 from ipaplatform.paths import paths
 from ipatests.pytest_ipa.integration.env_config import get_global_config
 from ipatests.test_integration.base import IntegrationTest
@@ -387,6 +388,21 @@ class TestInstallMaster(IntegrationTest):
 
     def test_install_dns(self):
         tasks.install_dns(self.master)
+
+    def test_WSGI_worker_process(self):
+        """ Test if WSGI worker process count is set to 4
+
+        related ticket : https://pagure.io/freeipa/issue/7587
+        """
+        # check process count in httpd conf file i.e expected string
+        exp = b'WSGIDaemonProcess ipa processes=%d' % constants.WSGI_PROCESSES
+        httpd_conf = self.master.get_file_contents(paths.HTTPD_IPA_CONF)
+        assert exp in httpd_conf
+
+        # check the process count
+        cmd = self.master.run_command('ps -eF')
+        wsgi_count = cmd.stdout_text.count('wsgi:ipa')
+        assert constants.WSGI_PROCESSES == wsgi_count
 
 
 class TestInstallMasterKRA(IntegrationTest):


ipatests/test_integration/test_installation.py
```diff
@@ -389,6 +389,16 @@ def test_install_dns(self):
         tasks.install_dns(self.master)
 
 
+    def test_WSGI_worker_process(self):
+        """ Test if WSGI worker process count is set to 4
+
+        related ticket : https://pagure.io/freeipa/issue/7587
+        """
+        master = self.master
```
Tiboris: You do not need this variable, you are using self.master only once.

And BTW this file will not exists if FreeIPA is not installed:
http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ecab09d6-9efe-11e8-a848-fa163e15adc6/runner.log.gz

pvoborni:
A question: the bug was about increasing the number WSGI worker processes. In this case from 2 to 4. Shouldn't the code test the number of processes instead of value in a configuration file? Hypothetical situation: What if the configuration used in #7587 doesn't actually change the number of processes?

As an example, wouldn't checking the following work better?

```
# ps -eF | grep wsgi
root     14224 13934  0  3016  1004   0 10:43 pts/0    00:00:00 grep --color=auto wsgi
kdcproxy 20293 20289  0 152190 28844  1 Aug06 ?        00:00:35 (wsgi:kdcproxy) -DFOREGROUND
kdcproxy 20295 20289  0 168574 28836  0 Aug06 ?        00:00:35 (wsgi:kdcproxy) -DFOREGROUND
ipaapi   20296 20289  0 265880 122252 1 Aug06 ?        00:00:40 (wsgi:ipa)      -DFOREGROUND
ipaapi   20297 20289  0 265048 107528 0 Aug06 ?        00:00:39 (wsgi:ipa)      -DFOREGROUND
ipaapi   20298 20289  0 265304 109316 1 Aug06 ?        00:00:39 (wsgi:ipa)      -DFOREGROUND
ipaapi   20300 20289  0 266332 110252 1 Aug06 ?        00:00:40 (wsgi:ipa)      -DFOREGROUND
```
(what we are looking for are those: `wsgi:ipa`)

```
# ps -eF | grep wsgi:ipa | grep -v grep | wc -l
4
```

Tiboris:
Thanks @pvoborni, 
@mrizwan93 i like fact that you added checking both cases.

rcritten:
I'm just spit-balling here, I doubt this will change much if ever, but it would be more dynamic by not hardcoding 4 as the number of processes and doing something like:

from ipaplatform.constants.constants import WSGI_PROCESSES
...
assert b'WSGIDaemonProcess ipa processes=%d' % WSGI_PROCESSES in httpd_conf
...
assert WSGI_PROCESSES in cmd.stdout_text

rcritten:
Changes LGTM. Want to give @Tiboris the final say.

ipatests/test_integration/test_installation.py
```diff
@@ -388,6 +389,22 @@ def test_install_kra(self):
     def test_install_dns(self):
         tasks.install_dns(self.master)
 
+    def test_WSGI_worker_process(self):
+        """ Test if WSGI worker process count is set to 4
+
+        related ticket : https://pagure.io/freeipa/issue/7587
+        """
+        # check process count in httpd conf file
+        exp_str = b'WSGIDaemonProcess ipa processes={}'.format(
```
flo-renaud: In python3 this will raise an Exception (AttributeError: 'bytes' object has no attribute 'format')

flo-renaud:
Removing the ACK label because the test will fail in python3.
Can you fix and add a temp commit ipatests/prci_definitions/gating.yaml in order to launch the test?

tiran:
@mrizwan93 Good work!

Please remove the temp commit

mrizwan93:
@tiran Thanks :)
Done

rcritten:
master:

* f0c3a3592818d132bd2b61e5fbe9e5b51b429a96 Test if WSGI worker process count is set to 4

rcritten:
@mrizwan93 a manual backport to ipa-4-6 is needed.

PR #2248, opened 2018-08-13
[ipa-4-5] Fix ipa-restore: create /var/run/ipa files

flo-renaud:
In ipa 4.5.4, ipa-restore fails because the file /etc/tmpfiles.d/ipa.conf
is not restored => /var/run/ipa and /var/run/ipa/ccaches directories
are not created.

The fix creates these directories in ipa-restore and creates ipa.conf.
With this approach, the fix allows to restore a backup done with 4.5.4
prior to the fix.

Note: the fix is specific to ipa-4-5, in ipa-4-6 and above version the file
/etc/tmpfiles.d/ipa.conf is created at package install time and not at
ipa server install time.

Fixes: https://pagure.io/freeipa/issue/7571

### DIFF
--- a/ipaserver/install/ipa_restore.py
+++ b/ipaserver/install/ipa_restore.py
@@ -32,6 +32,7 @@ from six.moves.configparser import SafeConfigParser
 from ipaclient.install.client import update_ipa_nssdb
 from ipalib import api, errors
 from ipalib.constants import FQDN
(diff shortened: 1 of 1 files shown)


Tiboris:
ipa-4-5:

* 9083f9b5167c5e56fc5c4a16e385e50a9043661e Fix ipa-restore: create /var/run/ipa files

PR #2249, opened 2018-08-13
[Backport][ipa-4-5] Check if replication agreement exist before enable/disable it

flo-renaud:
If the replication agreement does not exist, a custom exception is
raised explaining the problem.

https://pagure.io/freeipa/issue/7201


Tiboris:
ipa-4-5:

* 25f28b02cc959fee13be550f2c10f14fd75a9f6c Check if replication agreement exist before enable/disable it

PR #2263, opened 2018-08-21
DS replication settings: fix regression with <3.3 master

flo-renaud:
Commit 811b0fdb4620938963f1a29d3fdd22257327562c introduced a regression
when configuring replication with a master < 3.3
Even if 389-ds schema is extended with nsds5ReplicaReleaseTimeout,
nsds5ReplicaBackoffMax and nsDS5ReplicaBindDnGroupCheckInterval
attributes, it will return UNWILLING_TO_PERFORM when a mod
operation is performed on the cn=replica entry.

This patch ignores the error and logs a debug msg.

See: https://pagure.io/freeipa/issue/7617

### DIFF
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -22,6 +22,7 @@ from __future__ import print_function, absolute_import
 import logging
 import itertools
 
+import re
 import six
 import time
 import datetime
@@ -600,7 +601,20 @@ class ReplicationManager(object):
             r_conn.simple_bind(r_binddn, r_bindpw)
         else:
             r_conn.gssapi_bind()
-        self._finalize_replica_settings(r_conn)
+        # If the remote server has 389-ds < 1.3, it does not
+        # support the attributes we are trying to set.
+        # Find which 389-ds is installed
+        rootdse = r_conn.get_entry(DN(''), ['vendorVersion'])
+        version = rootdse.single_value.get('vendorVersion')
+        mo = re.search(r'(\d+)\.(\d+)\.(\d+)[\.\d]*', version)
+        vendor_version = tuple(int(v) for v in mo.groups())
+        if vendor_version >= (1, 3, 0):
+            # 389-ds understands the replication attributes,
+            # we can safely modify them
+            self._finalize_replica_settings(r_conn)
+        else:
+            logger.debug("replication attributes not supported "
+                         "on remote master, skipping update.")
         r_conn.close()
 
     def setup_chaining_backend(self, conn):


ipaserver/install/replication.py
```diff
@@ -600,7 +600,18 @@ def finalize_replica_config(self, r_hostname, r_binddn=None,
...
+        except errors.DatabaseError as e:
+            # On FreeIPA < 3.3 masters lacking support for the attributes
+            # defined in REPLICA_FINAL_SETTINGS,
+            # the update will return Unwilling to perform
+            # Ignore the error
+            if str(e).startswith('Server is unwilling to perform'):
+                logger.debug("replication attribute not supported "
+                             "on remote master (%s)", e)
+            else:
+                raise e
```
tbordaz: The changes looks okay. 
An other option is to apply the final settings (in _finalize_replica_settings) only  if the 'vendorVersion' >= 1.3* (search -b "" -s base). If version >= 1.3* and we get err=53, then it is a real issue.

tiran: I'd prefer @tbordaz suggestion and check the root DSE version information:

```
>>> dse = api.Backend.ldap2.conn.read_rootdse_s()
>>> vendor_version_str = dse['vendorVersion'][0].decode('utf-8')
>>> vendor_version_str
'389-Directory/1.4.0.13 B2018.201.1334'
>>> mo = re.search('(\d+)\.(\d+)\.(\d+)[\.\d]*', vendor_version_str)
>>> vendor_version = tuple(int(v) for v in mo.groups())
>>> vendor_version
(1, 4, 0)
```

flo-renaud:
Hi @tiran 
thanks for your suggestion, I implemented a check for version before trying to update the replication attributes on the remote master.

ipaserver/install/replication.py
```diff
@@ -600,7 +601,20 @@ def finalize_replica_config(self, r_hostname, r_binddn=None,
             r_conn.simple_bind(r_binddn, r_bindpw)
         else:
             r_conn.gssapi_bind()
-        self._finalize_replica_settings(r_conn)
+        # If the remote server has 389-ds < 1.3, it does not
+        # support the attributes we are trying to set.
+        # Find which 389-ds is installed
+        rootdse = r_conn.get_entry(DN(''), ['vendorVersion'])
+        version = rootdse.single_value.get('vendorVersion')
```
tiran: Does ``get_entry`` return bytes or text? The regular expression will fail under Python 3 if the value is bytes.

flo-renaud: get_entry returns an object `ipapython.ipaldap.LDAPEntry` and single_value.get() returns a str in python3, a unicode in python2, so the reg exp should not have any issue.

flo-renaud:
master:

* 02f4a7a2d6f6b5f5030d2c53df0085b8e699b6fd DS replication settings: fix regression with <3.3 master

PR #2265, opened 2018-08-21
uninstall -v: remove Tracebacks

flo-renaud:
ipa-server-install --uninstall -v -U prints Traceback in its log file.
This issue happens because it calls subprocess.Popen with close_fds=True
(which closes all file descriptors in the child process)
but it is trying to use the file logger in the child process
(preexec_fn is called in the child just before the child is executed).
The fix is using the logger only in the parent process.

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1480502

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -491,20 +491,21 @@ def run(args, stdin=None, raiseonerr=True, nolog=(), env=None,
     logger.debug('Starting external process')
     logger.debug('args=%s', arg_string)
 
+    if runas is not None:
+        pent = pwd.getpwnam(runas)
+
+        suplementary_gids = [
+            grp.getgrnam(sgroup).gr_gid for sgroup in suplementary_groups
+        ]
+
+        logger.debug('runas=%s (UID %d, GID %s)', runas,
+                     pent.pw_uid, pent.pw_gid)
+        if suplementary_groups:
+            for group, gid in zip(suplementary_groups, suplementary_gids):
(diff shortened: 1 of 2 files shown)


ipapython/ipautil.py
```diff
@@ -491,20 +491,21 @@ def run(args, stdin=None, raiseonerr=True, nolog=(), env=None,
...
-            pent = pwd.getpwnam(runas)
+    if runas is not None:
+        pent = pwd.getpwnam(runas)
 
-            suplementary_gids = [
-                grp.getgrnam(sgroup).gr_gid for sgroup in suplementary_groups
-            ]
+        suplementary_gids = [
+            grp.getgrnam(sgroup).gr_gid for sgroup in suplementary_groups
+        ]
```
rcritten: Is this guaranteed to be in-scope when preexec_fn() is called?

rcritten:
Is this testable using a standalone script that runs as root? Perhaps a trivial thing to call id to ensure that the uid/gid is changed?

flo-renaud:
Hi @rcritten 
thanks for your suggestion, I added a new test for run(runas=...).

flo-renaud:
Updated the commit msg with the pagure ticket instead of the BZ.

tiran:
master:

* f0228fa6497f56200ce649c10d5619700bcc5fca uninstall -v: remove Tracebacks
* 79fb0cc6634c7ab76e8181d1c55ac92a123142d8 ipautil.run: add test for runas parameter

PR #2266, opened 2018-08-21
Fix the uninstall test, execute in the nightly runs

rcritten:
I'm not sure what changed that caused the test to start failing. We didn't notice until now because the test wasn't executed in the nightlies.

Rather than only trying to stop dirsrv when it was running just brute-force always try to shut it down.


Rezney:
Thanks for the patch. Can we run "test_uninstallation" suite as part of a temp commit to see the test is now passing?

I tried to run it and am getting:

```
2018-08-22 11:19:03,501    DEBUG  ------------------ generated html file: /vagrant/report.html -------------------
2018-08-22 11:19:03,501    DEBUG  =================================== FAILURES ===================================
2018-08-22 11:19:03,501    DEBUG  ___________________ TestUninstallBase.test_failed_uninstall ____________________
2018-08-22 11:19:03,501    DEBUG  
2018-08-22 11:19:03,501    DEBUG  self = <ipatests.test_integration.test_uninstallation.TestUninstallBase object at 0x7f75bda74550>
2018-08-22 11:19:03,506    DEBUG  
2018-08-22 11:19:03,506    DEBUG      def test_failed_uninstall(self):
2018-08-22 11:19:03,506    DEBUG          self.master.run_command(['ipactl', 'stop'])
2018-08-22 11:19:03,506    DEBUG      
2018-08-22 11:19:03,506    DEBUG          serverid = realm_to_serverid(self.master.domain.realm)
2018-08-22 11:19:03,506    DEBUG          instance_name = ''.join([dsinstance.DS_INSTANCE_PREFIX, serverid])
2018-08-22 11:19:03,506    DEBUG      
2018-08-22 11:19:03,506    DEBUG          try:
2018-08-22 11:19:03,506    DEBUG              # Moving the DS instance out of the way will cause the
2018-08-22 11:19:03,511    DEBUG              # uninstaller to raise an exception and return with a
2018-08-22 11:19:03,511    DEBUG              # non-zero return code.
2018-08-22 11:19:03,511    DEBUG              self.master.run_command([
2018-08-22 11:19:03,511    DEBUG                  '/usr/bin/mv',
2018-08-22 11:19:03,511    DEBUG                  '%s/%s' % (paths.ETC_DIRSRV, instance_name),
2018-08-22 11:19:03,511    DEBUG                  '%s/%s.test' % (paths.ETC_DIRSRV, instance_name)
2018-08-22 11:19:03,511    DEBUG              ])
2018-08-22 11:19:03,511    DEBUG      
2018-08-22 11:19:03,511    DEBUG              cmd = self.master.run_command([
2018-08-22 11:19:03,511    DEBUG                  'ipa-server-install',
2018-08-22 11:19:03,511    DEBUG                  '--uninstall', '-U'],
2018-08-22 11:19:03,511    DEBUG                  raiseonerr=False
2018-08-22 11:19:03,511    DEBUG              )
2018-08-22 11:19:03,512    DEBUG              assert cmd.returncode == 1
2018-08-22 11:19:03,512    DEBUG          finally:
2018-08-22 11:19:03,512    DEBUG              # Be paranoid. If something really went wrong then DS may
2018-08-22 11:19:03,512    DEBUG              # be marked as uninstalled so server cert will still be
2018-08-22 11:19:03,512    DEBUG              # tracked and the instances may remain. This can cause
2018-08-22 11:19:03,512    DEBUG              # subsequent installations to fail so be thorough.
2018-08-22 11:19:03,512    DEBUG              dashed_domain = self.master.domain.realm.replace(".", '-')
2018-08-22 11:19:03,512    DEBUG              dirsrv_service = "dirsrv@%s.service" % dashed_domain
2018-08-22 11:19:03,512    DEBUG              self.master.run_command(['systemctl', 'stop', dirsrv_service])
2018-08-22 11:19:03,512    DEBUG      
2018-08-22 11:19:03,512    DEBUG              # Moving it back should allow the uninstall to finish
2018-08-22 11:19:03,512    DEBUG              # successfully.
2018-08-22 11:19:03,512    DEBUG              self.master.run_command([
2018-08-22 11:19:03,512    DEBUG                  '/usr/bin/mv',
2018-08-22 11:19:03,512    DEBUG                  '%s/%s.test' % (paths.ETC_DIRSRV, instance_name),
2018-08-22 11:19:03,512    DEBUG                  '%s/%s' % (paths.ETC_DIRSRV, instance_name)
2018-08-22 11:19:03,512    DEBUG              ])
2018-08-22 11:19:03,512    DEBUG      
2018-08-22 11:19:03,512    DEBUG              # DS has been marked as uninstalled so force the issue
2018-08-22 11:19:03,512    DEBUG              ds = dsinstance.DsInstance()
2018-08-22 11:19:03,513    DEBUG  >           ds.stop_tracking_certificates(serverid)
2018-08-22 11:19:03,513    DEBUG  
2018-08-22 11:19:03,513    DEBUG  /usr/lib/python3.6/site-packages/ipatests/test_integration/test_uninstallation.py:93: 
2018-08-22 11:19:03,513    DEBUG  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
2018-08-22 11:19:03,513    DEBUG  /usr/lib/python3.6/site-packages/ipaserver/install/dsinstance.py:1136: in stop_tracking_certificates
2018-08-22 11:19:03,513    DEBUG      nickname = self.get_server_cert_nickname(serverid)
2018-08-22 11:19:03,513    DEBUG  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
2018-08-22 11:19:03,513    DEBUG  
2018-08-22 11:19:03,513    DEBUG  self = <ipaserver.install.dsinstance.DsInstance object at 0x7f75c0910208>
2018-08-22 11:19:03,513    DEBUG  serverid = 'IPA-TEST'
2018-08-22 11:19:03,513    DEBUG  
2018-08-22 11:19:03,513    DEBUG      def get_server_cert_nickname(self, serverid=None):
2018-08-22 11:19:03,513    DEBUG          """
2018-08-22 11:19:03,516    DEBUG              Retrieve the nickname of the server cert used by dirsrv.
2018-08-22 11:19:03,516    DEBUG      
2018-08-22 11:19:03,516    DEBUG              The method directly reads the dse.ldif to find the attribute
2018-08-22 11:19:03,516    DEBUG              nsSSLPersonalitySSL of cn=RSA,cn=encryption,cn=config because
2018-08-22 11:19:03,516    DEBUG              LDAP is not always accessible when we need to get the nickname
2018-08-22 11:19:03,516    DEBUG              (for instance during uninstall).
2018-08-22 11:19:03,517    DEBUG              """
2018-08-22 11:19:03,517    DEBUG          if serverid is None:
2018-08-22 11:19:03,517    DEBUG              serverid = self.get_state("serverid")
2018-08-22 11:19:03,517    DEBUG          if serverid is not None:
2018-08-22 11:19:03,517    DEBUG              dirname = config_dirname(serverid)
2018-08-22 11:19:03,517    DEBUG              config_file = os.path.join(dirname, "dse.ldif")
2018-08-22 11:19:03,517    DEBUG              rsa_dn = "cn=RSA,cn=encryption,cn=config"
2018-08-22 11:19:03,517    DEBUG  >           with open(config_file, "r") as in_file:
2018-08-22 11:19:03,517    DEBUG  E           FileNotFoundError: [Errno 2] No such file or directory: '/etc/dirsrv/slapd-IPA-TEST/dse.ldif'
2018-08-22 11:19:03,517    DEBUG  
2018-08-22 11:19:03,517    DEBUG  /usr/lib/python3.6/site-packages/ipaserver/install/dsinstance.py:1116: FileNotFoundError
2018-08-22 11:19:03,517    DEBUG  ------------------------------ Captured log call -------------------------------
```

http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/c64da08e-a5fa-11e8-abed-fa163ef71872/runner.log.gz

rcritten:
Test fixed, ready for review.

ipatests/test_integration/test_uninstallation.py
```diff
@@ -89,17 +93,14 @@ def test_failed_uninstall(self):
...
-
             cmd = self.master.run_command([
                 'ipa-server-install',
                 '--uninstall', '-U'],
                 raiseonerr=False
             )
             assert cmd.returncode == 0
+
+            self.master.run_command([
+                paths.REMOVE_DS_PL,
```
tiran: Why is it necessary to run the 389-DS uninstaller again? In case ``ipa-server-install --uninstall`` is successful, the 389-DS instance is removed. In case it fails, the assert will trigger and the uninstaller is not executed.

ipatests/test_integration/test_uninstallation.py
```diff
@@ -55,6 +55,11 @@ def test_failed_uninstall(self):
         serverid = realm_to_serverid(self.master.domain.realm)
         instance_name = ''.join([dsinstance.DS_INSTANCE_PREFIX, serverid])
 
+        # Stop tracking the DS cert now to make it easier to clean up
+        # later.
+        ds = dsinstance.DsInstance()
+        ds.stop_tracking_certificates(None)
```
tiran: This code will be executed with the local NSSDB of a local 389-DS instance, not the remote instance. That doesn't make much sense.

ipatests/test_integration/test_uninstallation.py
rcritten: It's belt-and-suspenders in case something goes wrong to not leave an ns-slapd process running.

The original purpose of the test is to ensure that on failure a non-zero error is returned. This is forced by moving the DS instance out of the way.

In the DS uninstaller (and others) use restore_state() which removes the state from the stored file so if something blows up later then it may not be possible to know that there are leftovers. While this looks bad it hasn't been a common problems in my experience. But it does mean that if remove-ds.pl fails then DS is already marked as uninstalled therefore needs cleanup manually.

ipatests/test_integration/test_uninstallation.py
rcritten: Sigh, yes, this was me trying to be too clever by half. I removed a master invocation of ipa-getcert to replace this :-(  mostly because of hardcoding Server-Cert. Given it is a controlled environment I guess that's not a big deal.

ipatests/test_integration/test_uninstallation.py
tiran: But ``remove-ds.pl`` will never get executed when ``ipa-server-install`` fails. In error case, assert fails and raises an exception. I think this should do the trick (assuming that either ipa-server-install --uninstall cleans up everything or the other two commands do).

```
            uninstall = self.master.run_command([
                'ipa-server-install',
                '--uninstall', '-U'],
                raiseonerr=False
            )
            if uninstall.returncode != 0:
                self.master.run_command([
                    paths.IPA_GETCERT,
                    'stop-tracking',
                    '-d', '%s/%s' % (paths.ETC_DIRSRV, instance_name),
                    '-n', 'Server-Cert',
                ], raiseonerr=False)

                self.master.run_command([
                    paths.REMOVE_DS_PL,
                    '-i', instance_name
                ])
                pytest.fail("ipa-server-install --uninstall has failed")

```

tiran:
But ``remove-ds.pl`` will never get executed when ``ipa-server-install`` fails. In error case, assert fails and raises an exception. I think this should do the trick (assuming that either ipa-server-install --uninstall cleans up everything or the other two commands do).

```
            uninstall = self.master.run_command([
                'ipa-server-install',
                '--uninstall', '-U'],
                raiseonerr=False
            )
            if uninstall.returncode != 0:
                self.master.run_command([
                    paths.IPA_GETCERT,
                    'stop-tracking',
                    '-d', '%s/%s' % (paths.ETC_DIRSRV, instance_name),
                    '-n', 'Server-Cert',
                ], raiseonerr=False)

                self.master.run_command([
                    paths.REMOVE_DS_PL,
                    '-i', instance_name
                ])
                pytest.fail("ipa-server-install --uninstall has failed")

```

rcritten:
I'm less concerned about the second invocation of the uninstallation. There is no reason for it to fail. If it does it means something deeper is broken I suppose. If you insist I can add this. In any case I'll move the getcert change as it does need to be in the finally block.

rcritten:
Oh, I already moved getcert.

The whole point of renaming slapd-INSTANCE is to force a failure. The uninstaller is generally forgiving of problems and this is pretty much the only way I could reliably for an error such that I would get a non-zero return code.

flo-renaud:
Hi @rcritten 
looks good to me. The stop-tracking and remove-ds.pl calls are indeed needed because the 2nd time uninstall is executed, it fails to start dirsrv (`ipapwd_start - [file ipa_pwd_extop.c, line 1860]: Failed to get default realm?!` because krb5.conf has already been cleaned and doesn't contain the default realm any more). As a consequence, DS is not properly removed, and the test really needs to cleanup the cert tracking and ds instance.

Can you remove the temp commit so that we can push, and add the test to the nightly suite?

rcritten:
done

flo-renaud:
Adding re-run label as `fedora-28/test_authconfig` failed with P`rocess "vagrant up --no-provision --parallel" exited with error code 1`

flo-renaud:
All tests now green, ACK.

PR #2268, opened 2018-08-22
ipa commands: print 'IPA is not configured' when ipa is not setup

flo-renaud:
Some commands print tracebacks or unclear error message when they are called on a machine where ipa packages are installed but IPA is not configured.
Consistently report 'IPA is not configured on this system' in this case.

Also add a test.
    
Related to https://pagure.io/freeipa/issue/6261

### DIFF
--- a/install/tools/ipa-compat-manage
+++ b/install/tools/ipa-compat-manage
@@ -82,6 +82,8 @@ def main():
     retval = 0
     files = [paths.SCHEMA_COMPAT_ULDIF]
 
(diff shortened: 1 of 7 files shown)


tiran:
master:

* a5a619abc19af23cfb25d5ce35deb4338cee7822 ipa commands: print 'IPA is not configured' when ipa is not setup
* 10c6258929510cb99aa162c95a57b26184a0faca Test: test ipa-* commands when IPA is not configured

PR #2270, opened 2018-08-22
Fix render validation items on keypress event at login form

stanislavlevin:
There are many no needed render callings which are performed
on each "keypress" event at login form. It is enough to update
validation items on "CapsLock" state change.

Fixes: https://pagure.io/freeipa/issue/7679

### DIFF
--- a/install/ui/src/freeipa/widgets/LoginScreenBase.js
+++ b/install/ui/src/freeipa/widgets/LoginScreenBase.js
@@ -285,13 +285,17 @@ define(['dojo/_base/declare',
          */
         display_caps_warning: function(display) {
 
-            this.caps_warning = display;
             var val_summary = this.get_widget('validation');
             if (display) {
-                val_summary.add_warning('caps', this.caps_warning_msg);
+                if (!this.caps_warning) {
+                    val_summary.add_warning('caps', this.caps_warning_msg);
+                }
             } else {
-                val_summary.remove('caps');
+                if (this.caps_warning) {
+                    val_summary.remove('caps');
+                }
             }
+            this.caps_warning = display;
         },
 
         bind_validation: function(summary, field) {


install/ui/src/freeipa/widgets/LoginScreenBase.js
```diff
@@ -285,13 +285,17 @@ define(['dojo/_base/declare',
...
         display_caps_warning: function(display) {
 
-            this.caps_warning = display;
             var val_summary = this.get_widget('validation');
             if (display) {
-                val_summary.add_warning('caps', this.caps_warning_msg);
+                if (!this.caps_warning) {
+                    val_summary.add_warning('caps', this.caps_warning_msg);
+                }
             } else {
```
serg-cymbaluk: It is enough to add just one condition in the first line of the function:
`if (this.caps_warning == display) return;`
So function will do nothing if the state isn't changed.

stanislavlevin: It is not exactly same.
about 'caps_warning': 
```js
        /**                                                                     
         * Indicates that CAPS LOCK warning is on. Null indicates that we don't 
         * know the state.                                                      
         * @property {boolean|null}                                             
         */                                                                     
        caps_warning: null, 
```
But 'display' is boolean.
If we just use 'if (this.caps_warning == display) return;' under 'display=false' and 'caps_warning=null' then removing of 'caps' validation will be triggered. This is not i want.

serg-cymbaluk: So
`if (!!this.caps_warning === display) return;`

stanislavlevin: Nice trick :-)
But in this case CapsStatus ('caps_warning') will be catched on 'caps on -> keypress' only.
An expected (current) behavior - on any keypress regardless whether caps is on or off.
Reason of error:
'caps_warning' should become false/true on very first calling of function 'display_caps_warning'.

serg-cymbaluk: Ok, I agree.

PR #2283, opened 2018-08-24
Clear next field when returnining list elements in queue.c

frozencemetery:
The ipa-otpd code occasionally removes elements from one queue,
inspects and modifies them, and then inserts them into
another (possibly identical, possibly different) queue.  When the next
pointer isn't cleared, this can result in element membership in both
queues, leading to double frees, or even self-referential elements,
causing infinite loops at traversal time.

Rather than eliminating the pattern, make it safe by clearing the next
field any time an element enters or exits a queue.

Related https://pagure.io/freeipa/issue/7262

### DIFF
--- a/daemons/ipa-otpd/queue.c
+++ b/daemons/ipa-otpd/queue.c
@@ -111,12 +111,16 @@ void otpd_queue_push(struct otpd_queue *q, struct otpd_queue_item *item)
         q->head = q->tail = item;
     else
         q->tail = q->tail->next = item;
+
+    item->next = NULL;
 }
 
 void otpd_queue_push_head(struct otpd_queue *q, struct otpd_queue_item *item)
 {
     if (item == NULL)
         return;
+
+    item->next = NULL;
 
     if (q->head == NULL)
         q->tail = q->head = item;
@@ -145,6 +149,8 @@ struct otpd_queue_item *otpd_queue_pop(struct otpd_queue *q)
     if (q->head == NULL)
         q->tail = NULL;
 
+    if (item != NULL)
+        item->next = NULL;
     return item;
 }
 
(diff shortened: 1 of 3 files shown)


Tiboris:
Could you please add line:
`Related https://pagure.io/freeipa/issue/7262`
into commit message?

Tiboris:
@frozencemetery nevermind I did it myself :)

spoore1:
FYI I tested a 4.6 version of this and it seems to fix the crash.  I've run tests similar to the pagure 7262 description for many hours (12 last night, multiple tests before).  I've seen no new crashes since applying this patch.

rcritten:
It would be really great to have a unit test to test the queueing code.

daemons/ipa-otpd/queue.c
```diff
@@ -111,6 +111,8 @@ void otpd_queue_push(struct otpd_queue *q, struct otpd_queue_item *item)
         q->head = q->tail = item;
     else
         q->tail = q->tail->next = item;
+
+    item->next = NULL;
 }
 
 void otpd_queue_push_head(struct otpd_queue *q, struct otpd_queue_item *item)
```
flo-renaud: In otpd_queue_push_head, we should also set item->next = NULL if the queue is currently empty.

frozencemetery: Yes it should; good catch!  Fixed.

frozencemetery:
Also added unit tests.

Tiboris:
Hello @frozencemetery,
Thank you for the patch! 
Sorry for nitpicking, unfortunately line with ticket number in commit message disappeared.
Could you please add it?

flo-renaud:
Hi @frozencemetery 
thank you for the patch and the tests, works for me. Waiting for the updated commit msg with a ref to the ticket, then I will ACK.

frozencemetery:
Heh, sorry @Tiboris!  Fixed now.

flo-renaud:
master:

* fe650087b0dd35b3dd23347e98647fdc526614f8 Clear next field when returnining list elements in queue.c
* ab636682c4a0274d411c702ca62a6d26ab8b7869 Add cmocka unit tests for ipa otpd queue code

PR #2291, opened 2018-08-27
Update CA signing certificate default key size to 3072

abbra:
Todo:
- [ ] Figure out what is wrong with external CA re-signing test (a different public key is expected, ipa-cacert-manage renew fails)
- [ ] Decide what to do with server certs that explicitly use 2048 bit key length in p11helper and other places


ipatests/create_external_ca.py
```diff
@@ -24,12 +24,18 @@
...
+import ipalib.constants
 
 import datetime
 import six
 
 ISSUER_CN = 'example.test'
 
+MAP_SIGNING_ALGORITHM = {
+    'SHA256withRSA': hashes.SHA256,
+    'SHA512withRSA': hashes.SHA512,
```
tiran: ``SHA512withRSA`` makes only sense for RSA keys with a very large key size. For comparison a 15360bit RSA key has about the same strength as a 256bit AES key.

``SHA348`` and RSASSA-PSS algorithms are missing.

ipalib/constants.py
```diff
@@ -279,6 +279,8 @@
 DOMAIN_SUFFIX_NAME = 'domain'
 CA_SUFFIX_NAME = 'ca'
 PKI_GSSAPI_SERVICE_NAME = 'dogtag'
+PKI_CA_SIGNING_KEY_SIZE = '4096'
```
tiran: 4096 is way too large as default and goes way beyond all recommendations. CA and OCSP should be either 2048 or 3072, EE certs 2048bit.

ipatests/create_external_ca.py
```diff
@@ -45,7 +51,7 @@ def create_ca(self, cn=ISSUER_CN):
         """
         self.ca_key = rsa.generate_private_key(
             public_exponent=65537,
-            key_size=2048,
+            key_size=int(ipalib.constants.PKI_CA_SIGNING_KEY_SIZE),
```
tiran: The ``int`` call is not necessary.

ipatests/pytest_ipa/integration/create_caless_pki.py
```diff
@@ -285,7 +287,7 @@ def profile_kdc(builder, ca_nick, ca,
 def gen_cert(profile, nick_base, subject, ca=None, **kwargs):
     key = rsa.generate_private_key(
         public_exponent=65537,
-        key_size=2048,
+        key_size=int(ipalib.constants.PKI_CA_SIGNING_KEY_SIZE),
```
tiran: Please remove ``int``

ipatests/test_xmlrpc/test_caacl_profile_enforcement.py
```diff
@@ -410,10 +412,11 @@ def santest_csr(request, santest_host_1, santest_host_2):
     backend = default_backend()
     pkey = rsa.generate_private_key(
         public_exponent=65537,
-        key_size=2048,
+        key_size=int(ipalib.constants.PKI_CA_SIGNING_KEY_SIZE),
```
tiran: ``int``

ipatests/create_external_ca.py
abbra: The idea is that we need to have a mapping place for names to hash algorithm objects. I can add more here, no problem, as long as we keep them in sync with what could be set as a constant in `ipalib.constants`.

abbra: Specific choice for the default is irrelevant in this place.

ipalib/constants.py
abbra: We can change this when the actual agreement will be in place. For now, it is enough to demonstrate that certain code paths are failing because of the key size not being 2048. I'd rather concentrate on fixing those ones rather than discussing the defaults. Note that I specifically made a default-changing commit a separate one to not overlap with the real effort that needs to be done.

simo5: For a long term CA key, 4096 is correct at this point, given the rapid progress in QC.
2048 is considered ok for shorter lived certificates though.

ipatests/create_external_ca.py
abbra: It *is* necessary:
```
# python3 ./create_external_ca.py --csr /var/lib/ipa/ca.csr
  File "./create_external_ca.py", line 210, in <module>
    main()
  File "./create_external_ca.py", line 196, in main
    root_ca = external_ca.create_ca()
  File "./create_external_ca.py", line 55, in create_ca
    backend=default_backend(),
  File "/usr/lib64/python3.6/site-packages/cryptography/hazmat/primitives/asymmetric/rsa.py", line 118, in generate_private_key
    _verify_rsa_parameters(public_exponent, key_size)
  File "/usr/lib64/python3.6/site-packages/cryptography/hazmat/primitives/asymmetric/rsa.py", line 129, in _verify_rsa_parameters
    if key_size < 512:
TypeError: '<' not supported between instances of 'str' and 'int'

```

tiran: You have to define ``PKI_CA_SIGNING_KEY_SIZE`` as number, too.

ipalib/constants.py
tiran: 3072 is good enough for CA keys. 4096 is really slow, especially on low powered IoT devices.

ipatests/create_external_ca.py
abbra: It is really a conflict between RawConfigParser.set() and python cryptography code. Either I should do `str(foo)` in `config.set()` or `int(foo)` in these calls.

abbra:
I implemented @tiran requests and changed a default to 3072 bit key size right now.

ipatests/create_external_ca.py
```diff
@@ -24,12 +24,19 @@
...
+import ipalib.constants
 
 import datetime
 import six
 
 ISSUER_CN = 'example.test'
 
+MAP_SIGNING_ALGORITHM = {
+    'SHA256withRSA': hashes.SHA256,
+    'SHA384withRSA': hashes.SHA384,
```
rcritten: I don't believe dogtag supports SHA384withRSA by default. I don't see it in defaultSigningAlgsAllowed in CS.cfg.

abbra: Support for SHA384withRSA was added in Dogtag in June 2016 with https://github.com/dogtagpki/pki/commit/158bb22a87832ff2be07ac4b75c8f2927caefd55 so caCACert.cfg profile allows to use SHA384withRSA.

It is not yet allowed in CS.cfg by default, indeed. I need to check with the dogtag team why.

abbra: I checked with the Dogtag team and it looks like this is a simple omission that will be fixed in dogtag. So we can keep SHA256withRSA for time being.
I verified that updating CS.cfg and defaulting to SHA384withRSA creates working IPA CA with 3072 bit key size and SHA384withRSA signing algorithm.

abbra:
Yes, that's a known risk. I want to handle all HSM-related activities separately and allow to provide an overlay CA configuration file to installer to allow amend our defaults. So when that would be done, you'd provide a configuration overlay that sets right key size and other parameters.

abbra:
Looks like we do not pass explicitly required key size when generating a CSR for external CA when renewing the CA certificate. As result, certmonger picks up a default size for RSA and it is 2048. So while the certificate itself is having 3072 bit key size, the CSR is issued for 2048 bit key size.

abbra:
I fixed the resubmission of an external CA signing cert request.

flo-renaud:
@abbra 
I did not manage either to force a re-run of the tests. Could you try to rebase the PR, let's see if the new commit hashes trigger the tests.

abbra:
All tests passed.

tiran:
@abbra The PR has two open TODOs.

Did you talk to Tomas or Nikos from the crypto team to get their feedback on default key size?

abbra:
Closing this one. I created a separate ticket for dogtag to handle SHA384withRSA signing algorithm in CS.cfg and the rest from this PR is already handled in master.

PR #2299, opened 2018-08-28
Retrieve certificate subject base directly instead of ipa-join

rcritten:
The subject base is used as a fallback to find the available
CA certificates during client enrollment if the LDAP connection
fails (e.g. due to new client connecting to very old server) and
for constructing the subject if a certificate is requested.

raw=True is passed to config-show in order to avoid parsing
the server roles which will fail because the services aren't
marked as enabled until after the client installation is
successful on a master.

ipa-join providing the subject base via stderr was fragile and
would cause client enrollment to fail if any other output was
included in stderr.

https://pagure.io/freeipa/issue/7674

Testing is handled by existing integration tests, both in plain installations and --request-cert to ensure there is a subject base.

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -2684,13 +2684,6 @@ def _install(options):
             else:
                 logger.info("Enrolled in IPA realm %s", cli_realm)
 
-            start = stderr.find('Certificate subject base is: ')
-            if start >= 0:
-                start = start + 29
-                subject_base = stderr[start:]
-                subject_base = subject_base.strip()
-                subject_base = DN(subject_base)
-
             if options.principal is not None:
                 run([paths.KDESTROY], raiseonerr=False, env=env)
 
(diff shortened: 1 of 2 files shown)


ipaclient/install/client.py
```diff
@@ -2856,6 +2849,19 @@ def _install(options):
...
         disable_ra()
 
+    try:
+        result = api.Backend.rpcclient.forward(
+            'config_show',
+            raw=True,  # so that servroles are not queried
+            version=u'2.0'
+        )
+    except Exception as e:
+        raise ScriptError(
```
tiran: Could you please add a ``log.debug('msg', exc_info=True)`` line, so the exception traceback gets logged in the debug log?

rcritten: Done

Tiboris:
master:

* 8af6accfa5734a7e9a7c92fcf38d5440482413d4 Retrieve certificate subject base directly instead of ipa-join

PR #2301, opened 2018-08-29
Fix SELinux violation in fedora-domainname.service detection

tiran:
## Refactor os-release and platform information

Move the /etc/os-release parser and platform detection code out of the
private _importhook module. The ipaplatform module now contains an
osinfo module that provides distribution, os, and vendor information.

See: https://www.freedesktop.org/software/systemd/man/os-release.html
See: https://pagure.io/freeipa/issue/7661

## Don't check for systemd service

ipaplatform no longer checks for the presence of a systemd service file
to detect the name of the domainname service. Instead it uses osinfo's
version to use the old name on Fedora 28 and the new name on Fedora 29.

This fixes a SELinux violation that prevented httpd from listing systemd
service files.

Fixes: https://pagure.io/freeipa/issue/7661

### DIFF
--- a/ipaplatform/__init__.py
+++ b/ipaplatform/__init__.py
@@ -8,4 +8,4 @@ ignore.
 """
 __import__('pkg_resources').declare_namespace(__name__)
 
-NAME = None  # initialized by IpaMetaImporter
+NAME = None  # initialized by ipaplatform.osinfo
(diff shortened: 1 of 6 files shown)


rcritten:
Works, SELinux exceptions are gone.

Tiboris:
master:

* b8528da5a8e8cf4fdeabb77022cb511043544e9f Refactor os-release and platform information
* 1c03181e78b8f43e7bfd32e52c5b9d161c326fd6 Don't check for systemd service

PR #2302, opened 2018-08-29
Fix external CA next step example output

roman-mueller:
Currently it displays the option `--external-cert-file` twice.
The option should be `--external_cert_file` but only once and the second option should be `--external_ca_file`.


flo-renaud:
Hi,
thank you for your proposal, but I disagree with it based on the project history :)

This [commit](https://pagure.io/freeipa/c/3aa0731fc660ea3d111a44926ab5dea71dc510e7) shows that --external_cert_file and --external_ca_file options were used a while ago but replaced with the single option --external-cert-file. The PR would revert to the previous behavior which was modified on purpose, in order to simplify the options.

I propose to close this PR.

roman-mueller:
You are right, I looked up an old man page online which still had the old options in it. My bad!

PR #2303, opened 2018-08-29
Use $(hostname -f) to determine hostname in mod_ssl password script

rcritten:
https://pagure.io/freeipa/issue/7528

### DIFF
--- a/install/tools/ipa-httpd-pwdreader
+++ b/install/tools/ipa-httpd-pwdreader
@@ -5,32 +5,21 @@
 # see the documentation of the aforementioned directive of the mod_ssl module.
 
 USAGE="./ipa-pwdreader host:port RSA|DSA|ECC|number"
-ERR_UNKNOWN_KEY="\
-ERROR: You seem to be running a non-standard IPA installation.
-Please extend the /var/libexec/ipa/ipa-pwdreader script to cover your case."
 
-if [ ! "$#" -eq 2 ]; then
+if [ "$#" -ne 2 ]; then
     echo "Wrong number of arguments!" 1>&2
     echo "$USAGE" 1>&2
     exit 1
 fi
 
+fname=${1/:/-}-$2
+pwdpath=/var/lib/ipa/passwds/$fname
 
-case "$1" in
-    "${HOSTNAME}:443" )
-        # Read IPA password
-        # IPA expects the password filename format to be
-        # <hostname>-<port>-<ecryption_algorithm>
-        IPA_PASSWD_PATH="/var/lib/ipa/passwds/${1/:/-}-$2"
-        cat $IPA_PASSWD_PATH
-        ;;
-# ================
-# Extend for more virtual hosts with
-#    <vhostname>:<vhost_port> )
-#        your_code
-#        ;;
-# ================
-    *)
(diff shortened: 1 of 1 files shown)


abbra:
I had to do a bit of investigation of how actually mod_ssl is running this helper. For `exec:` type of filter it ends up using `fork()` + `execve()` with an empty environment. So this explains why `HOSTNAME` is undefined when the script is run by Apache.

I think using `hostname -f` is sensible here to solve the FQDN problem. This approach assumes our primary hostname is the one vhost is configured to use, though. If you'd have a system with multiple vhosts using the same certificate with SANs, it would not work, though.

May be we should not do this `HOSTNAME` comparison at all? Instead, resolve `${1/:/-}-$2` via `realpath -e --relative-to=/var/lib/ipa/passwds`,  check that it has no `..` or `/` in the result, do a stat on a file `/var/lib/ipa/passwds/$realpath` and if it exists and readable, return its content.

rcritten:
$HOSTNAME should be set by bash AFAICT. It definitely isn't empty since the password script works 99% of the time.

The current script is typical of what an Apache password reader looks like. It may be that Standa left it this way in order to allow for customization.

I'll look into your suggestion, it is an interesting idea. I think I'd prefer to except .. and / from $1 instead as these may be valid characters in a password. This script will run as root so we definitely need to be careful about what it is allowed to do.

rcritten:
I took a slightly different tact. I compare the calculated pwdfile path to that of the output of realpath. In any normal case this should match.

Theoretically this could blow up if some user has a symlink to /var but really...

Note that the way we generate the filename it is going to have a dash in it which will prevent most tricking of filenames anyway.

/usr/libexec/ipa/ipa-httpd-pwdreader ../../../../etc/shadow ''
resolves to
/etc/shadow-

abbra:
@adelton, can you help us here: regarding 'symlink to /var' above -- do you think this could cause issues to freeipa-container?

install/tools/ipa-httpd-pwdreader
```diff
@@ -4,33 +4,28 @@
 # If you'd like to write your custom binary providing passwords to mod_ssl,
 # see the documentation of the aforementioned directive of the mod_ssl module.
 
+hostname=$(hostname -f)
```
abbra: `hostname` variable is not used anymore

adelton:
If the symlinks don't use `..`s and are mostly absolute, things should work fine for freeipa-container.

abbra:
ACK. Rob, I think you want to rebase to get Travis CI status cleared. It is all green but somehow github shows Travis CI run is 'pending'.

flo-renaud:
I relaunched the travis-CI tests and the status got properly updated. Removing the needs_rebase label as it is not required.

flo-renaud:
@rcritten not sure what happened but your commit includes po files, can you fix it?

rcritten:
Gah! Thanks for noticing, I have no idea how those got in there!

flo-renaud:
fedora-28/test_service_permissions and fedora-28/test_topologies failed, adding re-run label

rcritten:
master:

* e3820682c797c8dca86767dc347290f54f872a7c Try to resolve the name passed into the password reader to a file

PR #2306, opened 2018-08-30
Test error when yubikey hardware not present

mrizwan93:
In order to work with IPA and Yubikey, libyubikey is required.
Before the fix, if yubikey added without having packages, it used to
result in traceback. Now it the exception is handeled properly.
It needs Yubikey hardware to make command successfull. This test
just check of proper error thrown when hardware is not attached.

related ticket : https://pagure.io/freeipa/issue/6979

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -407,6 +407,24 @@ class TestInstallMaster(IntegrationTest):
         wsgi_count = cmd.stdout_text.count('wsgi:ipa')
         assert constants.WSGI_PROCESSES == wsgi_count
 
+    def test_error_for_yubikey(self):
+        """ Test error when yubikey hardware not present
+
+        In order to work with IPA and Yubikey, libyubikey is required.
+        Before the fix, if yubikey added without having packages, it used to
+        result in traceback. Now it the exception is handeled properly.
+        It needs Yubikey hardware to make command successfull. This test
+        just check of proper error thrown when hardware is not attached.
+
+        related ticket : https://pagure.io/freeipa/issue/6979
+        """
+        # try to add yubikey to the user
+        args = ['ipa', 'otptoken-add-yubikey', '--owner=admin']
+        cmd = self.master.run_command(args, raiseonerr=False)
+        assert cmd.returncode != 0
+        exp_str = ("ipa: ERROR: No YubiKey found")
+        assert exp_str in cmd.stderr_text
+
 
 class TestInstallMasterKRA(IntegrationTest):
 


rcritten:
False-negatives are likely possible if libyubikey and/or libusb are already installed. I'm not sure what to suggest, perhaps a test skip if one or both of those packages are installed.

mrizwan93:
@rcritten Thanks for looking into it. 
What if:
1. uninstall theses packages first if installed already
2. test
3. re-install packages if needed by other testsuites

rcritten:
That seems reasonable.

rcritten:
You need to move the reinstallation of the packages either into a try/finally or a cleanup in order to restore the files in case of failure (the assert will cause the test to exit).

You conditionally remove packages and unconditionally re-add them. I'd pick one. I don't know enough about how the images are created to know whether uncondition is the way to go, maybe @Rezney can comment.

I'm not sure whether this is failing due to the test expecting the wrong thing or if the added code is not sufficient to correct the issue.

mrizwan93:
@rcritten @abbra 
Does the package name changed for yubikey or it is the error message which I am trying to assert?

```2018-12-03 10:57:38,628    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.ParamikoTransport] RUN yum remove -y libyubikey libusb
2018-12-03 10:57:38,628    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd52] RUN yum remove -y libyubikey libusb
2018-12-03 10:57:38,628    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd52] No match for argument: libyubikey
2018-12-03 10:57:38,628    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd52] No match for argument: libusb
2018-12-03 10:57:38,628    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd52] Error: No packages marked for removal.
2018-12-03 10:57:38,628    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd52] Exit code: 1
2018-12-03 10:57:38,628    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.ParamikoTransport] RUN ['ipa', 'otptoken-add-yubikey', '--owner=admin']
2018-12-03 10:57:38,628    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd53] RUN ['ipa', 'otptoken-add-yubikey', '--owner=admin']
2018-12-03 10:57:38,628    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd53] ipa: ERROR: No YubiKey found
2018-12-03 10:57:38,628    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd53] Exit code: 2

[..]

2018-12-03 10:58:55,493    DEBUG          # try to add yubikey to the user
2018-12-03 10:58:55,493    DEBUG          args = ['ipa', 'otptoken-add-yubikey', '--owner=admin']
2018-12-03 10:58:55,493    DEBUG          cmd = self.master.run_command(args, raiseonerr=False)
2018-12-03 10:58:55,499    DEBUG          assert cmd.returncode != 0
2018-12-03 10:58:55,499    DEBUG          exp_str = ("ipa: ERROR: No backend available. Please install"
2018-12-03 10:58:55,499    DEBUG                     "'libyubikey' and 'libusb' packages first.")
2018-12-03 10:58:55,499    DEBUG  >       assert exp_str in cmd.stderr_text
2018-12-03 10:58:55,499    DEBUG  E       assert "ipa: ERROR: No backend available. Please install'libyubikey' and 'libusb' packages first." in 'ipa: ERROR: No YubiKey found\n'
2018-12-03 10:58:55,499    DEBUG  E        +  where 'ipa: ERROR: No YubiKey found\n' = <pytest_multihost.transport.SSHCommand object at 0x7f8ddbb8e710>.stderr_text```

**Comment by abbra** [comment]
First, there is no package installed so nothing to remove. Second, there is no yubikey attached so you get a different error returned and as result a different text message. That different text message is what breaks your text as you did not expect it.

```
        try:
            yk = yubico.find_yubikey()
        except usb.core.USBError as e:
            raise NotFound(reason="No YubiKey found: %s" % e.strerror)
        except yubico.yubikey.YubiKeyError as e:
            raise NotFound(reason=e.reason)
        except ValueError as e:
            raise NotFound(reason=str(e) + ". Please install 'libyubikey' "
                           "and 'libusb' packages first.")
```

You need to make sure you are testing the right case.

The package `python3-yubico-1.3.2-10.fc29.noarch` is installed according to the test logs.

mrizwan93:
`yum remove -y python3-yubico` is removing all the freeipa packages (as dependency) as well. So uninstalling `python3-yubico` is not the solution here.

@abbra , could you guide further?

abbra:
@mrizwan93 I don't think you'd be able to test the other exception in PR CI because it is only shown if the hardware is plugged but we have no packages to support it. Otherwise, we always get a generalized `YubiKeyError` exception that simply says that there is no hardware plugged in.

From FreeIPA point of view, either message is preventing `ipa otptoken-add-yubikey` to proceed. It means a failure of `ipa otptoken-add-yubikey` is out test, so if you invert condition and ensure that the test always fails, that would be enough.

abbra:
The test looks good now. Please remove the temporary commit.

rcritten:
This is tagged for backporting to only 4.7 but the related BZ was against 4.6. Should this also be applied to 4.6?

abbra:
Yes, I think it can be applied to ipa-4-6 too -- I checked that both a spec file and the client plugin side have the same yubico package references as in newer branches.

Tiboris:
master:

* e7581cc0d88c3f7d83a551a3decd054589b1d86e Test error when yubikey hardware not present

PR #2308, opened 2018-08-30
Delete empty keytab during client installation

netoarmando:
Client installation fails if '/etc/krb5.keytab' exists as a zero-length file. Deleting empty keytab before proceeding with the installation fixes the problem.

https://pagure.io/freeipa/issue/7625

**Note to reviewers:**
Function implemented instead of an in-place solution, so it could be easily tested, if this solution is overkill I don't mind to remove the test case and move the check-and-delete procedure to where the function is being called.

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -2126,6 +2126,16 @@ def install_check(options):
         logger.warning("Option 'force-join' has no additional effect "
                        "when used with together with option 'keytab'.")
 
+    # Remove invalid keytab file
+    try:
+        gssapi.Credentials(
+            store={'keytab': paths.KRB5_KEYTAB},
+            usage='accept',
+        )
+    except gssapi.exceptions.GSSError:
+        logger.debug("Deleting invalid keytab: '%s'.", paths.KRB5_KEYTAB)
+        remove_file(paths.KRB5_KEYTAB)
+
     # Check if old certificate exist and show warning
     if (
         not options.ca_cert_file and


Tiboris:
Thanks for the patch @netoarmando!
IMHO testing python functionality is not what we want to do.
Probably adding the test for this particular case that creates empty keytab and then performs client installation would be the test case which will actually catch the issue.

frozencemetery:
Why does a 0-length file get created at /etc/krb5.keytab?  That's not a valid keytab file (missing header, footer, etc.).

rcritten:
We don't know but rather than failing cryptically we figured we could detect the condition and work around it.

netoarmando:
@Tiboris thanks for the input! I've changed an existing test case to save resources, what do you think?

@frozencemetery Do you think we should validate other properties of the keytab (header, footer, etc.) as well?

frozencemetery:
@netoarmando If we're going to do this at all, I'd just check klist.  If `klist -kt /etc/krb5.keytab` fails (and it will for empty files), either fail out installation or remove it.

ipaclient/install/client.py
```diff
@@ -2126,6 +2126,13 @@ def install_check(options):
         logger.warning("Option 'force-join' has no additional effect "
                        "when used with together with option 'keytab'.")
 
+    # Remove empty keytab file
+    if (
```
Tiboris: Could you please use `\` instead of `()` ?using  brackets do feel troublesome to me.
```python
if os.path.isfile(paths.KRB5_KEYTAB) and \
        os.path.getsize(paths.KRB5_KEYTAB) == 0 :
    remove_file(paths.KRB5_KEYTAB)
```

flo-renaud:
Hi @netoarmando 
it's a good idea to plug your test into an existing test in order to save resources, but I would rather use test_installation.py as a base (it's more natural to look for a client installation test in test_installation.py rather than in test_uninstallation.py).
Also, the test is failing but I'm not sure about the test status without your change (test_uninstallation.py is not part of PRCI-triggered tests or nightly tests, meaning we have no reference for its result).

ipaclient/install/client.py
netoarmando: Tibor, I think tests are failing for unrelated reasons, like Flo mentioned later, this test is not part of nightly bundle.

About breaking lines, I'm using [black](https://github.com/ambv/black) as reference, it suggests the use of parentheses in cases like this. However, after my changes this is not necessary anymore, so we are good! :-)

netoarmando:
Thank you, @flo-renaud! Tests already defined in `test_installation.py` leverage the fact that replica installation promotes clients, testing a client specific behavior using the current test cases seemed awkward to me, that's why I chose `test_uninstallation.py`. 

But, If I understood you correctly, in this case is okay to add a new test case to `test_installation.py`?

Unrelated question: Why `test_uninstallation.py` is not included in the nightly tests?

rcritten:
@netoarmando it was an oversight that it wasn't included in nightly.

Rather than forking out to klist you may be able to use the gssapi module to test a keytab (not sure).

netoarmando:
@rcritten,
> Rather than forking out to klist you may be able to use the gssapi module to test a keytab (not sure).

You are right, trying to initialize a `gssapi.Credentials` object with an invalid keytab, an exception is raised: 
```
GSSError: Major (851968): Unspecified GSS failure.  \
Minor code may provide more information, Minor (2529639093): \ 
Keytab FILE:/etc/krb5.notfound is nonexistent or empty
```

However, I'm still not sure the best place to test this.

tiran:
master:

* cf1301fb064fc230c780c4bc5eeccb723899f7b6 Delete empty keytab during client installation

PR #2312, opened 2018-08-31
Support the 1.4.x 389-DS Python installer

tiran:
Continuation and re-submit of https://github.com/freeipa/freeipa/pull/1563 to support DS python installer.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -48,9 +48,11 @@
 %global selinux_policy_version 3.14.1-14
 %global slapi_nis_version 0.56.1-4
 %global python_ldap_version 3.1.0-1
+# python3-lib389
 # Fix for "Installation fails: Replica Busy"
-# https://bugzilla.redhat.com/show_bug.cgi?id=1598478
-%global ds_version 1.3.8.4-15
+# https://pagure.io/389-ds-base/issue/49818
+%global ds_version 1.4.0.16-1
+
 %else
 # Fedora
 %global package_name freeipa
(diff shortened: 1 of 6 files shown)


stlaz:
So the last thing that I did not have time to do is the verbosity of lib389. With the logs now being properly not-doubled, python-lib389 codebase still contains chunks of code doing
```python
if self.verbose:
    self.log.info("neener-neener")
```
instead of setting up it's logger's logging level correctly, which results in lib389 being too verbose during the (un)installation (we set `verbose=True` to get the output).

You'll probably want to fix this to still get the log into the installer log-file while not having it printed on the output.

tiran:
@Firstyear @stlaz Simon has opened a ticket for me, https://pagure.io/389-ds-base/issue/49938

tiran:
``test_service_permissions`` keeps failing with vagrant provisioning errors. Looks like a build host issue.

Rezney:
@tiran will try to take a look what is going on...

tiran:
``fedora-28/test_commands`` is failing because OOM killer is killing Dogtag.

tiran:
``vagrant up`` failed, re-run

rcritten:
vagrant up failed, re-run

abbra:
ACK. Time has come to merge.

tiran:
master:

* f7511edbfb8e33015c9ea3ab340ab98136794cc8 Support the 1.4.x python installer tools in 389-ds
* 8f9b0fc1dec501652815853e36d02b7d250ced0c Remove some basic pystyle and pylint errors
* 383311a17fcfccc8d4e48e19dacd3f022c79c06b Don't try legacy installs
* ed955d14d3bc79dbbb90590d87f7e38777602808 Move lib389 imports to module scope
* 56f8e200131a0972e03b15c3cc3ae1dc30bf09b0 DS uninstall: fix serverid missing in state restore
* 104ef413ed9c3119531bb4a811958b270b31d3cc DS install: fix DS asking for NSS pin during install
* e8342d414d6ebec569d2f7c4126169f743f51d22 DS install: don't fail if SSL already configured
* 8fb63966cd0622d37309934ce8cbb1e65cf6ae81 Use the newer way of removing the DS instance
* a606b44f3ad091743dcdf32357271273cd1e49e0 Drop dependency on 389-ds-base-legacy-tools
* d39bb65a2f6e322b9870f2550bb977ffd4b4cad6 Remove DS perl paths from debian platform

PR #2327, opened 2018-09-04
Fix certificate type error when exporting to file

netoarmando:
Commands `ipa ca-show` and `ipa cert-show` share the same code, this commit updates the former, closing the gap between them.

Reflecting the changes done in 5a44ca6.

https://pagure.io/freeipa/issue/7628

To be more specific, `ipa cert-show` code is: https://github.com/freeipa/freeipa/blob/5a44ca638310913ab6b0c239374f4b0ddeeedeb3/ipaclient/plugins/cert.py#L51-L73

### DIFF
--- a/ipaclient/plugins/ca.py
+++ b/ipaclient/plugins/ca.py
@@ -1,6 +1,7 @@
 #
 # Copyright (C) 2016  FreeIPA Contributors see COPYING for license
 #
+import base64
 
 from ipaclient.frontend import MethodOverride
 from ipalib import errors, util, x509, Str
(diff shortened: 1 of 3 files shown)


ipaclient/plugins/ca.py
```diff
@@ -34,13 +35,11 @@ def forward(self, *keys, **options):
...
-                certs = (x509.load_der_x509_certificate(c)
-                         for c in result['result']['certificate_chain'])
+                certs = result['result']['certificate_chain']
             else:
-                certs = [
-                    x509.load_der_x509_certificate(
-                        result['result']['certificate'])
-                ]
+                certs = [result['result']['certificate']]
+            certs = (x509.load_der_x509_certificate(base64.b64decode(cert))
```
tiran: Tricky! This approach is a bit of a hack. In Python 3, ``base64.b64decode()`` accepts both ASCII text and ASCII bytes, and returns bytes. It works for me.

tiran:
master:

* c7064494e5801d5fd4670e6aab1e07c65d7a0731 Fix certificate type error when exporting to file

PR #2328, opened 2018-09-04
Add client uninstallation tests to nightly

netoarmando:
This PR adds the client uninstallation tests (back) to nightly PR-CI list.

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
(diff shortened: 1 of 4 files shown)


tiran:
Test is failing with

```
 def service_instance(self, instance_name, operation=None):
if self.lib_path_exists is None:
self.lib_path_exists = os.path.exists(self.lib_path)

elements = self.systemd_name.split("@")

# Make sure the correct DS instance is returned
if elements[0] == 'dirsrv' and not instance_name:

return ('dirsrv@%s.service'
> % str(self.api.env.realm.replace('.', '-')))
E           AttributeError: 'Env' object has no attribute 'realm'
```

The ``realm`` attribute and other attributes of ``api.env`` aren't set when the API object is not initialized or ``/etc/ipa/default.conf`` is gone.

netoarmando:
Closing this in favor of #2266.

PR #2331, opened 2018-09-05
tests: Installation of replica against a specific server

mrizwan93:
Test to check replica install against specific server. It uses master and
replica1 without CA and having custodia service stopped. Then try to
install replica2 from replica1 and expect it to get fail as specified server
is not providing all the services.

related ticket: https://pagure.io/freeipa/issue/7566

### DIFF
--- a/ipatests/prci_definitions/nightly_master_testing.yaml
+++ b/ipatests/prci_definitions/nightly_master_testing.yaml
@@ -480,6 +480,19 @@ jobs:
         timeout: 10800
         topology: *master_1repl
 
+  fedora-30/test_installation_TestInstallReplicaAgainstSpecificServer:
+    requires: [fedora-30/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-30/build_url}'
+        update_packages: True
+        test_suite: test_integration/test_installation.py::TestInstallReplicaAgainstSpecificServer
+        template: *testing-master-f30
+        timeout: 10800
+        topology: *master_2repl_1client
+
   fedora-30/test_idviews:
     requires: [fedora-30/build]
     priority: 50
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -847,3 +847,95 @@ class TestInstallMasterReplica(IntegrationTest):
 
         assert_text = 'Deleted IPA server "%s"' % self.replicas[0].hostname
         assert assert_text in cmd_output.stdout_text
+
+
+class TestInstallReplicaAgainstSpecificServer(IntegrationTest):
+    """Installation of replica against a specific server
+
+    Test to check replica install against specific server. It uses master and
+    replica1 without CA and having custodia service stopped. Then try to
+    install replica2 from replica1 and expect it to get fail as specified
+    server is not providing all the services.
+
+    related ticket: https://pagure.io/freeipa/issue/7566
+    """
+
+    num_replicas = 2
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master, setup_kra=True)
+
+        # install replica1 without CA
+        cmd = tasks.install_replica(cls.master, cls.replicas[0],
+                                    setup_ca=False, setup_dns=True,
+                                    promote=False)
+
+        # check for warning that CA is not installed on server
+        warn = 'WARNING: The CA service is only installed on one server'
+        assert warn in cmd.stderr_text
+
+    def test_replica_install_against_server_without_ca(self):
+        """Replica install will fail complaining about CA role
+        and exit code 4"""
+
+        # stop custodia service on replica1
+        self.replicas[0].run_command('systemctl stop ipa-custodia.service')
+
+        # check if custodia service is stopped
+        cmd = self.replicas[0].run_command('ipactl status')
+        assert 'ipa-custodia Service: STOPPED' in cmd.stdout_text
+
+        try:
+            # install replica2 against replica1, as CA is not installed on
+            # replica1, installation on replica2 should fail
+            cmd = tasks.install_replica(self.replicas[0], self.replicas[1],
+                                        promote=False, raiseonerr=False)
+            assert cmd.returncode == 4
+            error = "please provide a server with the CA role"
+            assert error in cmd.stderr_text
+
+        finally:
+            tasks.uninstall_master(self.replicas[1],
+                                   ignore_topology_disconnect=True,
+                                   ignore_last_of_role=True)
+
+    def test_replica_install_against_server_without_kra(self):
+        """Replica install will fail complaining about KRA role
+        and exit code 4"""
+
+        # install ca on replica1
+        tasks.install_ca(self.replicas[0])
+        try:
+            # install replica2 against replica1, as KRA is not installed on
+            # replica1(CA installed), installation should fail on replica2
+            cmd = tasks.install_replica(self.replicas[0], self.replicas[1],
+                                        promote=False, setup_kra=True,
+                                        raiseonerr=False)
+            assert cmd.returncode == 4
+            error = "please provide a server with the KRA role"
+            assert error in cmd.stderr_text
+
+        finally:
+            tasks.uninstall_master(self.replicas[1],
+                                   ignore_topology_disconnect=True,
+                                   ignore_last_of_role=True)
+
+    def test_replica_install_against_server(self):
+        """Replica install should succeed if specified server provide all
+        the services"""
+
+        tasks.install_replica(self.master, self.replicas[1],
+                              setup_dns=True, promote=False)
+
+        # check if replication agreement stablished between master
+        # and replica2 only.
+        cmd = self.replicas[1].run_command(['ipa-replica-manage', 'list',
+                                            self.replicas[0].hostname])
+        assert self.replicas[0].hostname not in cmd.stdout_text
+
+        dirman_password = self.master.config.dirman_password
+        cmd = self.replicas[1].run_command(['ipa-csreplica-manage', 'list',
+                                            self.replicas[0].hostname],
+                                           stdin_text=dirman_password)
+        assert self.replicas[0].hostname not in cmd.stdout_text
--- a/ipatests/prci_definitions/nightly_f29.yaml
+++ b/ipatests/prci_definitions/nightly_f29.yaml
@@ -447,6 +447,18 @@ jobs:
         timeout: 10800
         topology: *master_1repl
 
+  fedora-29/test_installation_TestInstallReplicaAgainstSpecificServer:
+    requires: [fedora-29/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-29/build_url}'
+        test_suite: test_integration/test_installation.py::TestInstallReplicaAgainstSpecificServer
+        template: *ci-master-f29
+        timeout: 10800
+        topology: *master_2repl_1client
+
   fedora-29/test_idviews:
     requires: [fedora-29/build]
     priority: 50
--- a/ipatests/prci_definitions/nightly_master.yaml
+++ b/ipatests/prci_definitions/nightly_master.yaml
@@ -447,6 +447,18 @@ jobs:
         timeout: 10800
         topology: *master_1repl
 
+  fedora-30/test_installation_TestInstallReplicaAgainstSpecificServer:
+    requires: [fedora-30/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-30/build_url}'
+        test_suite: test_integration/test_installation.py::TestInstallReplicaAgainstSpecificServer
+        template: *ci-master-f30
+        timeout: 10800
+        topology: *master_2repl_1client
+
   fedora-30/test_idviews:
     requires: [fedora-30/build]
     priority: 50
--- a/ipatests/prci_definitions/nightly_rawhide.yaml
+++ b/ipatests/prci_definitions/nightly_rawhide.yaml
@@ -447,6 +447,18 @@ jobs:
         timeout: 10800
         topology: *master_1repl
 
+  fedora-rawhide/test_installation_TestInstallReplicaAgainstSpecificServer:
+    requires: [fedora-rawhide/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-rawhide/build_url}'
+        test_suite: test_integration/test_installation.py::TestInstallReplicaAgainstSpecificServer
+        template: *ci-master-frawhide
+        timeout: 10800
+        topology: *master_2repl_1client
+
   fedora-rawhide/test_idviews:
     requires: [fedora-rawhide/build]
     priority: 50


mrizwan93:
Seems like firewalld is not present on f-28

ipatests/test_integration/test_installation.py
```diff
@@ -567,3 +567,55 @@ def test_KRA_install_after_cert_renew(self):
...
         self.master.run_command(['systemctl', 'start', 'chronyd'])
+
+
+class TestInstallReplicaAgainstSpecificServer(IntegrationTest):
+    """Installation of replica against a specific server
+
+    Test to check replica install against specific server. It uses master and
+    replica1 without CA and having custodia service stopped. Then try to
+    install replica2 from replica1 so that replica2 will fetch secrets from
+    master as custodia service is not running on replica1.
```
flo-renaud: It's not clear to me what should be the expected behavior in this case. Replica2 / master communications are prevented by the firewall, so should we expect the test to fail?

mrizwan93: Please see https://github.com/freeipa/freeipa/pull/2331#issuecomment-419455394

flo-renaud: Sorry but I still don't get your point:
- replica2 is installed with setup_ca = True (default value in tasks.install_replica) but there is no CA on replica1 => the installer should use master because it needs a CA.
- replica2 / master communications are blocked
So the test should be expecting a failure.

mrizwan93:
Added family to rich-rule as it was throwing error:
```
2019-03-14 06:39:44,651    DEBUG  transport.py               513 DEBUG    RUN ['firewall-cmd', '--add-rich-rule', 'rule source address=192.168.121.134 accept']
2019-03-14 06:39:44,651    DEBUG  transport.py               558 DEBUG    Error: MISSING_FAMILY
```

flo-renaud:
I am forcing a re-run to check the test result after the commit from PR #2939.

ipatests/test_integration/test_installation.py
```diff
@@ -847,3 +847,34 @@ def test_replicamanage_del(self):
...
+    related ticket: https://pagure.io/freeipa/issue/7566
+    """
+
+    num_replicas = 2
+
+    def test_replica_install_against_server(self):
+        tasks.install_master(self.master)
+
+        # install replica1 without CA
+        tasks.install_replica(self.master, self.replicas[0], setup_ca=False)
```
flo-renaud: I would install the replica with DNS, in order to make sure that it's the lack of CA that forces the use of master, and not the lack of DNS.

flo-renaud:
Hi @mrizwan93 
after more thinking I am not sure that setting --server during the ipa-client-install will force the use of the server during replica promotion. Could you install the replica with one-step installation instead of client + promotion? In this case, if the issue 7566 is properly fixed we should see only references to a single master used for the repl install in the logs.

mrizwan93:
Hello @flo-renaud 
The replica is installed in single step only. I have used `task.install_replica()` for it.

ipatests/test_integration/test_installation.py
```diff
@@ -847,3 +847,36 @@ def test_replicamanage_del(self):
...
+        # stop custodia service on replica1
+        self.replicas[0].run_command('systemctl stop ipa-custodia.service')
+
+        # check if custodia service is stopped
+        cmd = self.replicas[0].run_command('ipactl status')
+        assert 'ipa-custodia Service: STOPPED' in cmd.stdout_text
+
+        # install replica2 against replica1, as custodia service is stopped
+        # on replica1, replica2 will fetch secrets from master
+        tasks.install_replica(self.replicas[0], self.replicas[1])
```
fcami: -  tasks.install_replica() defaults to promote=True which launches ipa-client-install first then ipa-replica-install. However promoting a client makes ipa-replica-install ignore the --server argument:
```
transport.py               558 DEBUG    IPA client is already configured on this system, ignoring the --domain, --server, --realm, --hostname, --password and --keytab options.
```

- The first master is installed with the default setup_dns=True but that second replica uses (again, the default) setup_dns=False so it is not a complete replica.

- With the custodia service stopped on replica0 I would _expect_ tasks.install_replica(self.replicas[0], self.replicas[1]) to fail.

ipatests/test_integration/test_installation.py
```diff
@@ -847,3 +847,36 @@ def test_replicamanage_del(self):
...
+    related ticket: https://pagure.io/freeipa/issue/7566
+    """
+
+    num_replicas = 2
+
+    def test_replica_install_against_server(self):
+        tasks.install_master(self.master)
+
+        # install replica1 without CA
+        tasks.install_replica(self.master, self.replicas[0],
```
fcami: Please add setup_kra=True to master, replica0, replica1 install stanza for better coverage.

ipatests/test_integration/test_installation.py
mrizwan93: Replica1 is supposed to fetch the secrets from master in that case.

ipatests/prci_definitions/gating.yaml
```diff
@@ -7,10 +7,10 @@ topologies:
     name: master_1repl
     cpu: 4
     memory: 6450
-  master_1repl_1client: &master_1repl_1client
-    name: master_1repl_1client
+  master_2repl_1client: &master_2repl_1client
```
fcami: Please use the symlink and ipatests/prci_definitions/temp_commit.yaml instead

fcami: The symlink should be at the root of the repository btw.

flo-renaud:
Hi @mrizwan93,
from the test logs, I suspect that the issue 7566 is not completely fixed because the installer is using the first replica for some operations (like running conn check and replication conf) and the master for others (custodia). This means that the test does not correctly detect the problem.

Extract from the ipa-replica-install logs:
the check for the master hostname is done against the replica (ok because it's the value provided in --server):
```
2019-08-20T12:56:32Z DEBUG Check if replica1.ipa.test is a primary hostname for localhost
2019-08-20T12:56:32Z DEBUG Primary hostname for localhost: replica1.ipa.test
2019-08-20T12:56:32Z DEBUG Search DNS for replica1.ipa.test
2019-08-20T12:56:32Z DEBUG Check if replica1.ipa.test is not a CNAME
2019-08-20T12:56:32Z DEBUG Check reverse address of 192.168.121.158
2019-08-20T12:56:32Z DEBUG Found reverse name: replica1.ipa.test
```
the remote_api is bootstrapped against the replica, used for most of the initial checks:
```
2019-08-20T12:56:33Z DEBUG trying https://replica1.ipa.test/ipa/json
```
ipa-replica-conncheck is done against the replica:
```
2019-08-20T12:56:34Z DEBUG args=['/usr/sbin/ipa-replica-conncheck', '--master', 'replica1.ipa.test', '--auto-master-check', '--realm', 'IPA.TEST', '--hostname', 'replica0.ipa.test', '--principal', 'admin', '--password', XXXXXXXX, '--ca-cert-file', '/etc/ipa/ca.crt']
```
The ds keytab is obtained from the replica:
```
2019-08-20T12:57:01Z DEBUG args=['/usr/sbin/ipa-getkeytab', '-k', '/etc/dirsrv/ds.keytab', '-p', 'ldap/replica0.ipa.test@IPA.TEST', '-H', 'ldaps://replica1.ipa.test']
```

The replication is setup with the replica:
```
2019-08-20T12:57:05Z DEBUG Waiting for replication (ldap://replica1.ipa.test:389) cn=meToreplica0.ipa.test,cn=replica,cn=dc\=ipa\,dc\=test,cn=mapping tree,cn=config (objectclass=*)
```
but later we see that the master is used for custodia:
```
2019-08-20T12:57:47Z DEBUG Custodia client for '<CustodiaModes.CA_PEER: 'Custodia CA peer'>' with promotion yes.
2019-08-20T12:57:47Z INFO Custodia uses 'master.ipa.test' as master peer.
```
and
```
2019-08-20T12:57:56Z DEBUG Waiting up to 300 seconds to see our keys appear on host ldap://master.ipa.test
```

In the test I would add a check on the replication agreements. When a replica is installed against a specific server, I would expect all the agreements to be setup against this given server.

When 7566 is completely fixed, I would also expect your test to fail during the second replica install because we specifically ask to install using the first replica but it is impossible given that it does not provide a CA instance.

flo-renaud:
Hi @mrizwan93 
Travis-CI detected a PEP-8 error:
```
./ipatests/test_integration/test_installation.py:888:31: E128 continuation line under-indented for visual indent
```

ipatests/prci_definitions/gating.yaml
flo-renaud: The correct way to use the temp commit is the following:
- don't modify gating.yaml
- edit ipatests/prci_definition/temp_commit.yaml and add the test case you want to run
- from the root of your repository, run `ln -sf ipatests/prci_definitions/temp_commit.yaml .freeipa-pr-ci.yaml`
- `git add ipatests/prci_definitions/temp_commit.yaml  .freeipa-pr-ci.yaml`
- `git commit`
With these steps, .freeip-pr-ci.yaml will point to the temp_commit.yaml file. PRCI is extracting the list of tests from .freeipa-pr-ci.yaml, and will execute only the ones specified in temp_commit.yaml.

ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -49,7 +45,7 @@ jobs:
         git_refspec: '{git_refspec}'
         template: &ci-master-f30
           name: freeipa/ci-master-f30
-          version: 0.0.4
+          version: 0.0.3
```
flo-renaud: Please keep version: 0.0.4 as it's the most recent one.

ipatests/test_integration/test_installation.py
flo-renaud: Please update the test description. The installation is expected to fail as replica1 does not provide a CA and was forced with --server param.

ipatests/test_integration/test_installation.py
```diff
@@ -847,3 +847,43 @@ def test_replicamanage_del(self):
...
+        tasks.install_master(self.master)
+
+        # install replica1 without CA
+        cmd = tasks.install_replica(self.master, self.replicas[0],
+                                    setup_ca=False, setup_dns=True,
+                                    promote=False)
+
+        # check for warning that CA is not installed on server
+        warn = 'WARNING: The CA service is only installed on one server'
+        assert warn in cmd.stdout_text
```
flo-renaud: The test execution failed here, you can try checking cmd.stderr_text instead.

ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -60,7 +60,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-30/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
+        test_suite: test_integration/test_installation.py::TestInstallReplicaAgainstSpecificServer
         template: *ci-master-f30
         timeout: 3600
         topology: *master_1repl_1client
```
flo-renaud: The topology should be master_2repl_1client otherwise the test won't be executed (it requires num_replicas = 2).

ipatests/test_integration/test_installation.py
```diff
@@ -847,3 +847,43 @@ def test_replicamanage_del(self):
...
+
+        # check if custodia service is stopped
+        cmd = self.replicas[0].run_command('ipactl status')
+        assert 'ipa-custodia Service: STOPPED' in cmd.stdout_text
+
+        # install replica2 against replica1, as custodia service is stopped
+        # on replica1, installation on replica2 should fail
+        cmd = tasks.install_replica(self.replicas[0], self.replicas[1],
+                                    promote=False, raiseonerr=False)
+        assert cmd.returncode != 0
```
flo-renaud: With PR #3664, `ipa-replica-install --server ...` will return 4 when the provided remote master does not have CA/KRA. You can also check for the error message `please provide a server with the CA role`.

You can also enhance this scenario:
- install KRA on the first master (add the param `setup_kra=True` to `tasks.install_master` on line 867
- add a new test function that will install the CA on replicas[0], and try to call `ipa-replica-install --setup-ca --setup-kra --server replicas[0]` on replicas[1], should be refused because KRA not present on replicas[0] and produce error message `please provide a server with the KRA role`
- add a new test function that calls `ipa-replica-install --server master --setup-ca --setup-kra`, should succeed. Check that the replica agreement are established with master only.

ipatests/test_integration/test_installation.py
```diff
@@ -847,3 +847,43 @@ def test_replicamanage_del(self):
...
+    replica1 without CA and having custodia service stopped. Then try to
+    install replica2 from replica1 and expect it to get fail as specified
+    server is not providing all the services.
+
+    related ticket: https://pagure.io/freeipa/issue/7566
+    """
+
+    num_replicas = 2
+
+    @pytest.mark.xfail(reason='freeipa ticket 7566', strict=True)
```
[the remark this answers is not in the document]
mrizwan93: Okay, I'll wait for the fix and then enhance the PR with suggested scenarios.

ipatests/test_integration/test_installation.py
```diff
@@ -847,3 +847,91 @@ def test_replicamanage_del(self):
...
+            tasks.uninstall_master(self.replicas[1],
+                                   ignore_topology_disconnect=True,
+                                   ignore_last_of_role=True)
+
+    def test_replica_install_against_server_without_kra(self):
+        """Replica install will fail complaining about KRA role
+        and exit code 4"""
+
+       # install ca on replica1
+       tasks.install_ca(self.replca[0])
```
flo-renaud: Hi @mrizwan93 
the indentation is missing one space in lines 907 and 908.

ipatests/test_integration/test_installation.py
```diff
@@ -847,3 +847,91 @@ def test_replicamanage_del(self):
...
+        # stop custodia service on replica1
+        self.replicas[0].run_command('systemctl stop ipa-custodia.service')
+
+        # check if custodia service is stopped
+        cmd = self.replicas[0].run_command('ipactl status')
+        assert 'ipa-custodia Service: STOPPED' in cmd.stdout_text
+
+        try:
+            # install replica2 against replica1, as custodia service is
+            # stopped on replica1, installation on replica2 should fail
```
flo-renaud: The comment should rather be about replica1 not providing CA.

flo-renaud:
Adding re-run as build failed

ipatests/test_integration/test_installation.py
```diff
@@ -847,3 +847,88 @@ def test_replicamanage_del(self):
...
+    def test_replica_install_against_server(self):
+        """Replica install should succeed if specified server provide all
+        the services"""
+
+        tasks.install_replica(self.master, self.replicas[1],
+                              setup_dns=True, promote=False)
+
+        # check if replication agreement stablished between master
+        # and replica2 only.
+        cmd = self.replicas[1].run_command(['ipa-replica-manage', 'list'])
```
flo-renaud: This command lists all the replicas, it is expected to find master, replica0 and replica1.
You should use instead `ipa-replica-manage list replica1` as it will display only the list of repl agreements for replica1.
You can also do the same command with `ipa-csreplica-manage list replica` to display the repl agreements for the CA suffix.

ipatests/prci_definitions/nightly_master_testing.yaml
```diff
@@ -480,6 +480,18 @@ jobs:
         timeout: 10800
         topology: *master_1repl
 
+  fedora-30/test_installation_TestInstallReplicaAgainstSpecificServer:
+    requires: [fedora-30/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-30/build_url}'
```
flo-renaud: Need to add
```
        update_packages: True
```
for this nightly def (this nightly is doing a dnf update on all packages to install the packages available from updates-testing repo).

flo-renaud:
Re-running `fedora-30/test_smb ` as it failed on timeout. Otherwise, changes LGTM.

netoarmando:
@mrizwan93 could you please rebase your PR to include changes from `master` branch?

flo-renaud:
master:

* c2c1000e2d5481d4be377feb12588fdb09d12de0 Installation of replica against a specific server
* c77bbe7899577cb14b42625953f1b9a868e6f237 Add test to nightly yamls

flo-renaud:
Hi @mrizwan93 
the PR needs a manual backport to ipa-4-8, ipa-4-7 and ipa-4-6. Can you handle it? Thanks.

PR #2332, opened 2018-09-05
 ipa-replica-install: properly use the file store 

flo-renaud:
### ipa-replica-install: properly use the file store
In ipa-replica-install, many components use their own instance of the FileStore to backup configuration files to the pre-install state. This causes issues when the calls are mixed, like for instance:
ds.do_task1_that_backups_file (using ds.filestore)
http.do_task2_that_backups_file (using http.filestore)
ds.do_task3_that_backups_file (using ds.filestore)

because the list of files managed by ds.filestore does not include the files managed by http.filestore, and the 3rd call would remove any file added on 2nd call.

The symptom of this bug is that ipa-replica-install does not save /etc/httpd/conf.d/ssl.conf and subsequent uninstallation does not restore the file, leading to a line referring to ipa-rewrite.conf
that prevents httpd startup.

The installer should consistently use the same filestore.

Fixes https://pagure.io/freeipa/issue/7684

### Test: scenario replica install/uninstall should restore ssl.conf

Test that the scenario ipa-replica-install/ uninstall correctly restores the file /etc/httpd/conf.d/ssl.conf

Related to https://pagure.io/freeipa/issue/7684

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -79,7 +79,7 @@ def make_pkcs12_info(directory, cert_name, password_name):
 
 
 def install_replica_ds(config, options, ca_is_configured, remote_api,
-                       ca_file, promote=False, pkcs12_info=None):
+                       ca_file, promote=False, pkcs12_info=None, fstore=None):
     dsinstance.check_ports()
 
     # if we have a pkcs12 file, create the cert db from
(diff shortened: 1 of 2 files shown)


rcritten:
We ran into something similar in the past. It was fixed in 082c55fb9cf87263f1f585a1adeda464a9d7328a

Is that the same issue?

flo-renaud:
@rcritten 

> We ran into something similar in the past. It was fixed in 082c55f
> 
> Is that the same issue?
082c55f was about the class StateFile (usually linked to sysrestore.state which contains sections and key=value), but this one is related to the class FileStore (sysrestore.index which contains a list of backup files with their permissions uid and gid).
Do you think I should rather follow the same strategy for the fix, i.e. reload the list of files at the beginning of every public method?

@tiran 
thanks for the review, I added the fstore to the bindinstance call.

tiran:
master:

* 6ad11d86d881880a4e7ada7932c2e0401dc2cda7 ipa-replica-install: properly use the file store
* b2ce20c6a7c8c046911d13b9ba5b73f9d5046a27 Test: scenario replica install/uninstall should restore ssl.conf

PR #2345, opened 2018-09-06
Update required version of dogtag to detect when FIPS is available

rcritten:
When it was checking for FIPS it assumed that /proc/sys/crypto
existed which it doesn't in some containers and on Ubuntu.

This was updated in dogtag, this change is just to pull in the
fix.

https://pagure.io/freeipa/issue/7608

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -72,9 +72,9 @@
 
 %endif  # Fedora
 
-# Require Dogtag PKI 10.6.1 with Python 3 and SQL NSSDB fixes for external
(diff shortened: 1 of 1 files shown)


flo-renaud:
@rcritten thanks for the PR, LGTM. Note that I tested that the freeipa-* package require pki-* version 10.6.6, but did not test the install on a system where `/proc/sys/crypto` is absent.

flo-renaud:
master:

* 7da50cf46182cd771e45420a30b7147a78577e69 Update required version of dogtag to detect when FIPS is available

PR #2347, opened 2018-09-07
ipa-server-upgrade: fix typo in setup_lightweight_ca_key_retrieval

flo-renaud:
The method setup_lightweight_ca_key_retrieval is called on
server upgrade and checks first if it needs to be executed or if
a previous upgrade already did the required steps.
The issue is that it looks for setup_lwca_key_retrieval in sysupgrade.state
but writes setup_lwca_key_retieval (with a missing r).

Fixes https://pagure.io/freeipa/issue/7688

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -1301,7 +1301,10 @@ class CAInstance(DogtagInstance):
         self._ldap_enable(u'enabledService', "CA", self.fqdn, basedn, config)
 
     def setup_lightweight_ca_key_retrieval(self):
-        if sysupgrade.get_upgrade_state('dogtag', 'setup_lwca_key_retrieval'):
+        # Important: there is a typo in the below string, which is known
+        # and should not be fixed as existing installations already use it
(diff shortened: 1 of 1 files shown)


rcritten:
This is currently the only place it is set but if it gets referenced in ipa-server-upgrade in the future both values will need to be considered. It has been in IPA since 4.4.0.

Do you have any ideas on where we can note that?

That or an upgrade task to replace the bad value?

flo-renaud:
@rcritten 
thanks for your comment. I changed the fix so that only one label is used (the one which may already exist in sysupgrade.state), meaning we keep the typo `setup_lwca_key_retieval` but consistency ensures that the method will not be called when it's not needed.

frasertweedale:
Hi @flo-renaud, the fix is fine but perhaps you can add a comment in the code stating that the typo is known, and should not be fixed.

tiran:
master:

* 0659ed35025bc0555e42fff71c305c1cd830cf9a ipa-server-upgrade: fix inconsistency in setup_lightweight_ca_key_retrieval

PR #2353, opened 2018-09-11
Incompatible IPA versions, pausing replication

tbordaz:
Bug Description:
	The bug is not systematic. Symptoms are replication broken and this message
	ERR - repl_version_plugin_recv_acquire_cb - [file ipa_repl_version.c, line 119]: Incompatible IPA versions, pausing replication
	IPA Version Replication is a plugin that checks replication version (payload).
	It does the checking in the early phase of replication protocol (acquire_replica).
	The plugin must be enabled on both side, or disabled on both side.
	If we have a mixed setting (i.e. disabled on Master and enabled on Replica), there is
	a risk that the Master sends an incomplete payload. The incomplete payload will be rejected
	by the Replica and replication in both direction will fail.
	The risk is that on the master, the plugin got initialized but not started.
	Initialization of the plugin registers a broker API that allows a plugin (replication) to call
	a callback even if the plugin is not started.
	The problem is that to build valid payload the 'IPA Version Replication' needs to be started
	(data_version to be initialized).

Fix Description:
	The fix consist in:
		- define data_version during plugin initialization
		- Make plugin callback as noop if the plugin is not started. This is just for perf.
		- Prevent to enable the 'IPA Version Replication' when a replica is created.

https://pagure.io/freeipa/issue/7690

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-version/ipa_repl_version.c
+++ b/daemons/ipa-slapi-plugins/ipa-version/ipa_repl_version.c
@@ -45,6 +45,7 @@
 
 #define IPA_PLUGIN_NAME "ipa_replication_version"
 static char *data_version = NULL;
+static PRBool plugin_started = PR_FALSE;
 
 /*
  * Plugin identifiers
@@ -74,6 +75,9 @@ static int
 repl_version_plugin_pre_acquire_cb(void *cookie, const Slapi_DN *repl_subtree,
                                         int is_total, char **data_guid, struct berval **data)
 {
+    if (!plugin_started) {
+        return 0;
+    }
     LOG("repl_version_plugin_pre_acquire_cb() called for suffix \"%s\", "
         "is_total: \"%s\".\n", slapi_sdn_get_ndn(repl_subtree),
         is_total ? "TRUE" : "FALSE");
@@ -106,6 +110,9 @@ static int
 repl_version_plugin_recv_acquire_cb(const char *repl_subtree, int is_total,
                                          const char *data_guid, const struct berval *data)
 {
+    if (!plugin_started) {
+        return 0;
+    }
     LOG("test_repl_session_plugin_recv_acquire_cb() called for suffix \"%s\", is_total: \"%s\".\n",
         repl_subtree, is_total ? "TRUE" : "FALSE");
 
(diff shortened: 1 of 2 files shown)


flo-renaud:
Added the re-run label to trigger PR-CI as tbordaz is not in the whitelist yet.

rcritten:
The test_vault case is failing to install a replica with: [Error (49)  - LDAP error: Invalid credentials]

ipaserver/install/ipa_replica_prepare.py
```diff
@@ -321,9 +320,6 @@ def run(self):
 
         logger.info("Preparing replica for %s from %s",
                     self.replica_fqdn, api.env.host)
-        enable_replication_version_checking(
-            api.env.realm,
-            self.dirman_password)
```
rcritten: The whole point of this is to provide a relief valve in the case where we make some breaking architectural change in IPA so that one version can't replicate with another. If the check is not done at install time this is pointless, right?

tbordaz: Currently the mechanism checks at each replication session, that versions are compatible. So it is not only at install. In addition I am not sure this mechanism would prevent install between incompatible versions.
If we want to check compatible version we can also add a version entry/attribute in the DIT instead of breaking replication

rcritten: I don't mean to imply it is only at install but if it isn't checked at install time then the installation could be successful only to fail at the first replication event right?

We don't have to stick with this replication version but it was, as I understood it, the best or only way to do this years ago.

tbordaz: according to https://bugzilla.redhat.com/show_bug.cgi?id=1623761#c12, the problem is dynamic and transient. So it is quite possible that even if the problem occurs during install, replication retries and install script does not detect the error. During runtime, retry occurs and replication looks just fine.

Repl_version is a nice piece of code and I think the intention was to prevent incompatible versions to pollute each other. So it is enforcing compatibility in a stronger way than uniq/regular version checking from DIT entry. Now being a plugin there is always the risk that it is disabled on one/both side and version not checked. 

If we want to keep this ability to prevent a new version to break an old one (or the opposite), we should keep this plugin enabled.

tbordaz: @rcritten , I have a mixed feeling regarding the repl_version checking.

This plugin was created to prevent incompatible versions to break each other. So it is addressing rare condition that never occurred so far and now we are willing to disable it by default.

Now if we want to keep such safety mechanism . It should be enabled on most or all instances. incompatibility checking requires that the both sides have it enabled.

In short, currently repl_version is disabled on Master and enabled on Replica. This patch is disabling it on Replica. I wonder if in the opposite it should not enable it as well on master.

flo-renaud:
Adding re-run as `fedora-28/dnssec — RunPytest Failed to publish artifacts`

PR #2359, opened 2018-09-13
ipa-advise: configure pam_cert_auth=True for smart card on client

flo-renaud:
ipa-advise config-client-for-smart-card-auth is now using authselect
instead of authconfig, but authselect enable-feature with-smartcard
does not set pam_cert_auth=True in /etc/sssd/sssd.conf.
As a result, smart card auth on a client fails.
The fix adds a step in ipa-advise to configure pam_cert_auth=True.

The fix also forces the use of python3 interpreter.

Fixes https://pagure.io/freeipa/issue/7532

### DIFF
--- a/ipaserver/advise/plugins/smart_card_auth.py
+++ b/ipaserver/advise/plugins/smart_card_auth.py
@@ -52,6 +52,9 @@ class common_smart_card_auth_config(Advice):
             )
 
     def upload_smartcard_ca_certificates_to_systemwide_db(self):
+        # Newer version of sssd use OpenSSL and read the CA certs
+        # from /etc/sssd/pki/sssd_auth_ca_db.pem
+        self.log.command('mkdir -p /etc/sssd/pki')
         with self.log.for_loop(
                 self.single_ca_cert_variable_name,
                 '${}'.format(self.smart_card_ca_certs_variable_name)):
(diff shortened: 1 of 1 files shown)


sumit-bose:
Hi Florence,

I just realized that there is another issue with newer versions of SSSD which are build with OpenSSL and not NSS.

The CA certificates to validate the Smartcard certificates are not read from /etc/pki/nssdb anymore, since OpenSSL cannot read the nss database but are expected in /etc/sssd/pki/sssd_auth_ca_db.pem or whereever pam_cert_db_path points to (see man sssd.conf for details).

A small change like 
```
diff --git a/ipaserver/advise/plugins/smart_card_auth.py b/ipaserver/advise/plugins/smart_card_auth.py
index e8b2bb6..c60af2c 100644
--- a/ipaserver/advise/plugins/smart_card_auth.py
+++ b/ipaserver/advise/plugins/smart_card_auth.py
@@ -52,6 +52,7 @@ class common_smart_card_auth_config(Advice):
             )
 
     def upload_smartcard_ca_certificates_to_systemwide_db(self):
+        self.log.command('mkdir -p /etc/sssd/pki')
         with self.log.for_loop(
                 self.single_ca_cert_variable_name,
                 '${}'.format(self.smart_card_ca_certs_variable_name)):
@@ -60,6 +61,10 @@ class common_smart_card_auth_config(Advice):
                 '-t CT,C,C'.format(
                     self.systemwide_nssdb, self.single_ca_cert_variable_name
                 )
+            self.log.command(
+                'cat {} >>  /etc/sssd/pki/sssd_auth_ca_db.pem'.format(
+                    self.single_ca_cert_variable_name
+                )
             )
 
     def install_smart_card_signing_ca_certs(self):
```

might already help. Shall I open a new ticket for this or do you prefer to add it here?

flo-renaud:
Hi @sumit-bose ,
thanks for the heads-up, I will include the changes in this PR.

rcritten:
Works well, confirmed the new sssd CA directory behavior.

flo-renaud:
master:

* 7729bb73b477cc04d44aba3094f72e301ec11d94 ipa-advise: configure pam_cert_auth=True for smart card on client

PR #2363, opened 2018-09-14
authselect: harden uninstallation of ipa client

flo-renaud:
When ipa client is uninstalled, the content of sysrestore.state is read to restore the previous authselect profile and features.
The code should properly handle the case where sysrestore.state contains the header for the authselect section, but the key=value for profile and features are missing.

The PR also adds an integration test.

Fixes https://pagure.io/freeipa/issue/7657

### DIFF
--- a/ipaplatform/redhat/authconfig.py
+++ b/ipaplatform/redhat/authconfig.py
@@ -162,8 +162,10 @@ class RedHatAuthSelect(RedHatAuthToolBase):
             profile = 'sssd'
             features = ''
         else:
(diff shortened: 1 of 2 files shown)


rcritten:
Nice, simple fix. Works great.

flo-renaud:
master:

* d0173c9548857f855d95ba4bb5b068841f356049 authselect: harden uninstallation of ipa client
* 1a7e4b0ec1a7fcd7242b4ae76582a272c4f63657 tests: add test for uninstall with incomplete sysrestore.state

PR #2373, opened 2018-09-21
Support Samba 4.9

abbra:
Samba 4.9 became a bit more strict about creating a local NT token and a
failure to resolve or create BUILTIN\Guests group will cause a rejection
of the connection for a successfully authenticated one.

Add a default mapping of the nobody group to BUILTIN\Guests.

BUILTIN\Guests is a special group SID that is added to the NT token for
authenticated users.

For real guests there is 'guest account' option in smb.conf which
defaults to 'nobody' user.

This was implicit behavior before as 'guest account = nobody' by
default would pick up 'nobody' group as well.

Fixes: https://pagure.io/freeipa/issue/7705

### DIFF
--- a/ipaserver/install/plugins/adtrust.py
+++ b/ipaserver/install/plugins/adtrust.py
@@ -23,7 +23,8 @@ from ipalib import Registry, errors
 from ipalib import Updater
 from ipapython.dn import DN
 from ipaserver.install import sysupgrade
-from ipaserver.install.adtrustinstance import ADTRUSTInstance
+from ipaserver.install.adtrustinstance import (
+    ADTRUSTInstance, map_Guests_to_nobody)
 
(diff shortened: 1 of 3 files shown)


ipaserver/install/plugins/adtrust.py
```diff
@@ -382,3 +383,20 @@ def execute(self, **options):
             return False, ()
 
         return False, ()
+
+
+@register()
+class update_mapping_Guests_to_nobody(Updater):
```
rcritten: This is never invoked AFAICT.

abbra:
Thanks, @rcritten, forgot that these plugins need an explicit call out. Added that.

tiran:
master:

* 703497532abe4189835d0a02b32f9919c889bc1c Support Samba 4.9

PR #2375, opened 2018-09-21
ipatests: mark known failures as xfail

flo-renaud:
The tests in test_integration/test_installation.py
that inherit from InstallTestBase2 all fail in
test_replica2_ipa_kra_install because of ticket
7654: ipa-kra-install fails on DL1

This is an issue linked to dogtag (see
https://pagure.io/dogtagpki/issue/3055), where the
installation of a KRA clone creates a range depletion
when multiple clones are created from the same master.

Marking the tests as known failure, waiting for dogtag's
fix.

Related to https://pagure.io/freeipa/issue/7654

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -122,6 +122,7 @@ class InstallTestBase2(IntegrationTest):
     def test_replica2_ipa_ca_install(self):
         tasks.install_ca(self.replicas[2])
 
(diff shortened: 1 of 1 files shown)


tiran:
@frasertweedale PR https://github.com/dogtagpki/pki/pull/40 is work in progress

tiran:
master:

* 2488813260a407477c7516b33ce4238b69c8dd8d ipatests: mark known failures as xfail

PR #2380, opened 2018-09-24
Sprinkle raw strings across the code base

tiran:
tox / pytest is complaining about lots and lots of invalid escape
sequences in our code base. Sprinkle raw strings or backslash escapes
across the code base to fix most occurences of:

```
  DeprecationWarning: invalid escape sequence
```

There is still one warning that keeps repeating, though:

```
  source:264: DeprecationWarning: invalid escape sequence \d
```


flo-renaud:
@tiran 
thanks for the PR, LGTM. It was not a fun review and I guess your work was even more tedious :)

PR #2381, opened 2018-09-24
Workaround for pyasn1 0.4

tiran:
pyasn1 0.4 changed handling of ANY containers in a backwards
incompatible way. Try to decode ASN.1 octet streams explicitly for
pyasn1 0.3, then fall back to 0.4 behavior.

Tested with pyasn1 0.3.7 and 0.4.4.

Fixes: https://pagure.io/freeipa/issue/7685

### DIFF
--- a/ipalib/x509.py
+++ b/ipalib/x509.py
@@ -46,9 +46,9 @@ from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.serialization import (
     Encoding, PublicFormat, PrivateFormat, load_pem_private_key
 )
+import pyasn1
 from pyasn1.type import univ, char, namedtype, tag
 from pyasn1.codec.der import decoder, encoder
-# from pyasn1.codec.native import decoder, encoder
 from pyasn1_modules import rfc2315, rfc2459
 import six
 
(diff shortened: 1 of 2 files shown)


frasertweedale:
@tiran does anything need to be done on the encoding side (see `IPACertificate.__encode_extension`)?

tiran:
@frasertweedale good catch! Of course it's also broken :rotating_light:

ipalib/x509.py
```diff
@@ -285,7 +289,7 @@ def extended_key_usage_bytes(self):
             return None
 
         ekurfc = rfc2459.ExtKeyUsageSyntax()
-        for i, oid in enumerate(eku):
+        for i, oid in enumerate(sorted(eku)):
```
frasertweedale: I'm curious: what's the reason for this change?

tiran: EKU is a set of OIDs. Without sort, the sorting of EKU items depends on Python's hash randomziation. The change makes the sort order of EKUs stable. It simplifies testing and makes the output predictable.

tiran:
master:

* e1a30d3ce32aef2492689de8eff09360f5480a99 Workaround for pyasn1 0.4

frasertweedale:
On Wed, Sep 26, 2018 at 09:55:06PM -0700, Christian Heimes wrote:
> tiran commented on this pull request.
> 
> 
> 
> > @@ -285,7 +289,7 @@ def extended_key_usage_bytes(self):
>              return None
>  
>          ekurfc = rfc2459.ExtKeyUsageSyntax()
> -        for i, oid in enumerate(eku):
> +        for i, oid in enumerate(sorted(eku)):
> 
> EKU is a set of OIDs. Without sort, the sorting of EKU items depends on Python's hash randomziation. The change makes the sort order of EKUs stable. It simplifies testing and makes the output predictable.
>
Fair enough.

PR #2383, opened 2018-09-24
Fix script loading errors in ssbrowser.html and unauthorized.html

serg-cymbaluk:
Related issue: https://pagure.io/freeipa/issue/7703

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1081,6 +1081,8 @@ fi
 %dir %{_usr}/share/ipa/ui/js/freeipa
 %{_usr}/share/ipa/ui/js/freeipa/app.js
 %{_usr}/share/ipa/ui/js/freeipa/core.js
+%{_usr}/share/ipa/ui/js/freeipa/ssbrowser.js
+%{_usr}/share/ipa/ui/js/freeipa/unauthorized.js
 %dir %{_usr}/share/ipa/ui/js/plugins
 %dir %{_usr}/share/ipa/ui/images
 %{_usr}/share/ipa/ui/images/*.jpg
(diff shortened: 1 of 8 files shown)


rcritten:
So the fix was just to move the javascript out of the html file? Can you describe the reasoning for the fix in the commit message?

stanislavlevin:
Hello, @serg-cymbaluk 
My apologies, this was my mistake (checked at dev mode only).

Please take a look at last commit: 
https://github.com/stanislavlevin/freeipa/commits/fix_js_at_production

It completely fixes the issue.

serg-cymbaluk:
@rcritten 

> So the fix was just to move the javascript out of the html file? Can you describe the reasoning for the fix in the commit message?

We can't import modules in HTML-file because it requires  js file in the runtime. But when we import module from other module this dependency is resolved in compile-time.
So idea is that all pages have own compiled js file with all dependencies inside.
But at this moment I am inclined to accept @stanislavlevin's [patch](https://github.com/stanislavlevin/freeipa/commit/21acc955be46b4090c4ed0b96467fd28c4a0a464) because it's simpler one.

serg-cymbaluk:
> Please take a look at last commit:
> https://github.com/stanislavlevin/freeipa/commits/fix_js_at_production
@stanislavlevin, please create new PR if it is final edits.

stanislavlevin:
#2387 opened.
Excuse me one more time.

tiran:
@stanislavlevin This PR needs rebase, too.

stanislavlevin:
> @stanislavlevin This PR needs rebase, too.

It does not belong to me ;-)
The issue was fixed at #2387 (already pushed).

About this PR, in my opinion, it is no longer needed.

tiran:
Sorry @stanislavlevin 

@serg-cymbalu, what about this PR? Is it still necessary?

serg-cymbaluk:
@tiran 

> what about this PR? Is it still necessary?

It is already not actual.

PR #2387, opened 2018-09-25
Fix loading 'freeipa/text' at production mode

stanislavlevin:
As for now 'ssbrowser.html' and 'unauthorized.html' pages are
loaded without JS error at development mode only.

There is no standalone 'freeipa/text' module as a source at
production mode. Thus 'core' one has to be loaded first and
then 'text'.

Related issue: https://pagure.io/freeipa/issue/7703

### DIFF
--- a/install/html/ssbrowser.html
+++ b/install/html/ssbrowser.html
@@ -25,20 +25,26 @@
             ]
         };
         (function() {
+            var icons = [
+                '../ui/favicon.ico',
+            ];
             var styles = [
                 '../ui/css/patternfly.css',
                 '../ui/css/ipa.css'
             ];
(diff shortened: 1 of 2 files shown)


stanislavlevin:
There is no problem:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

tiran:
Meh, it's only JSON that annoys people with lack of trailing commas. Thanks :100:

serg-cymbaluk:
master:

* f658a3d107e0c1cb6a316b9e99e75027591b22f9 Fix loading 'freeipa/text' at production mode

PR #2385, opened 2018-09-24
Fix translation of 'Add dialog for 'association' facet

stanislavlevin:
As for now the 'Add' dialog title, which is initialized within the details of the entity, contains translatable concatenated texts, like:
```
'Add ${other_entity} into ${entity} ${primary_key}'
```
where 'other_entity' and 'entity' are also translatable texts.

This construction is used via method 'show_add_dialog' of association_facet as well as 'create_add_dialog' method of 'association_table' widget for the all 'Add' actions within the details of entities.

The concatenation leads to a bad quality translation and should be changed to an entire sentence.
    
From now a mentioned title is taken from a spec and should be specified explicitly.
    
Fixes: https://pagure.io/freeipa/issue/7712
Fixes: https://pagure.io/freeipa/issue/7714

PR #2389, opened 2018-09-25
Fix translation of 'add' dialog's title for association_table widget

stanislavlevin:
As for now the default title of add dialogs, which are initialized from 'association_table' widget, is set to something like:
```python
 'Add ${other_entity} into ${entity} ${primary_key}'
```

where 'other_entity' and 'entity' are also translatable texts. This construction is used via 'create_add_dialog' method of 'association_table' widget for the all 'Add' actions within
the details of entities.
    
Such concatenation leads to a bad quality translation and should be replaced with an entire sentence.
    
From now a mentioned title is taken from a spec and should be specified explicitly.
    
Fixes: https://pagure.io/freeipa/issue/7714

### DIFF
--- a/install/ui/src/freeipa/association.js
+++ b/install/ui/src/freeipa/association.js
@@ -184,7 +184,7 @@ IPA.attribute_adder_dialog = function(spec) {
             required: true
         }
     ];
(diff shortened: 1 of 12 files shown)


serg-cymbaluk:
@stanislavlevin please rebase this branch.

serg-cymbaluk:
#2385 already has included this changes

PR #2392, opened 2018-09-26
Py3: Remove subclassing from object

tiran:
Python 2 had old style and new style classes. Python 3 has only new
style classes. There is no point to subclass from object any more.

See: https://pagure.io/freeipa/issue/7715

### DIFF
--- a/contrib/lite-server.py
+++ b/contrib/lite-server.py
@@ -106,7 +106,7 @@ def get_ccname():
     return krb5_unparse_ccache(scheme, location)
 
 
-class KRBCheater(object):
+class KRBCheater:
     """Add KRB5CCNAME to WSGI environ
     """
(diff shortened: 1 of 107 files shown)


flo-renaud:
Hi @tiran 
thanks for the PR. You may also want to modify the following files:
- install/tools/ipa-custodia-check.in
- install/tools/ipa-httpd-kdcproxy.in
- install/tools/ipa-replica-conncheck.in

tiran:
@flo-renaud good catch! I fixed the three files, too.

flo-renaud:
@tiran thanks for the update, ack

tiran:
master:

* b431e9b684df11c811892bd9d2a5711355f0076e Py3: Remove subclassing from object

PR #2397, opened 2018-09-26
Add jslint check to PR CI tests

stanislavlevin:
For now, from all possible lint checks, pylint applies only.
jslint can prevent JavaScript errors at WebUI.

Fixes: https://pagure.io/freeipa/issue/7717


tiran:
Fix is now part of PR #2398

PR #2401, opened 2018-09-27
Py3: Replace six.moves imports

tiran:
Replace six.moves and six.StringIO/BytesIO imports with cannonical
Python 3 packages.

Note: six.moves.input behaves differently than builtin input function.
Therefore I left six.moves.input for now.

See: https://pagure.io/freeipa/issue/7715


flo-renaud:
master:

* 15d5e44ee89ce4c0f1f366b0d7150166f1a26844 Py3: Replace six.moves imports

PR #2405, opened 2018-09-27
Require sssd-ipa instead of sssd meta pkg

tiran:
The sssd meta package pulls in additional dependencies that are not
required by IPA clients. Only depend on sssd-ipa.

See: https://pagure.io/freeipa/issue/7710

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -80,6 +80,10 @@
 # https://bugzilla.redhat.com/show_bug.cgi?id=1568271
 %global nss_version 3.36.1-1.1
 
+# One-Way Trust authenticated by trust secret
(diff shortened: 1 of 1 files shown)


abbra:
Please update SSSD version to 2.0

abbra:
1.16.3 is in F28, let's use that one. It includes fixes for the Samba AD DC compatibility I will soon need.

tiran:
@abbra 1.16.3-2 it is!

tiran:
master:

* 78c722d4c3b939d94d5dcfb46ede52f242ebaeb2 Require sssd-ipa instead of sssd meta pkg

PR #2406, opened 2018-09-27
[Backport][ipa-4-7] Require sssd-ipa instead of sssd meta pkg

tiran:
Manual backport of PR #2405 

The sssd meta package pulls in additional dependencies that are not
required by IPA clients. Only depend on sssd-ipa.

Also update SSSD to 1.16.3-2 with fixes with support for One-Way Trust
authenticated by trust secret.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1345975
See: https://pagure.io/freeipa/issue/7710
Reviewed-By: Alexander Bokovoy <abokovoy@redhat.com>

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -102,6 +102,10 @@
 # https://bugzilla.redhat.com/show_bug.cgi?id=1568271
 %global nss_version 3.36.1-1.1
 
+# One-Way Trust authenticated by trust secret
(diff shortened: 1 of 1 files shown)


tiran:
ipa-4-7:

* 5a25dc5375c20f961ef6006f39070fddb3a11908 Require sssd-ipa instead of sssd meta pkg

PR #2407, opened 2018-09-28
[Py3] Fix zonemgr encoding issue

tiran:
The zonemgr validator and handler performs additional encodings for IDNA
support. In Python 3, the extra steps are no longer necessary because
arguments are already proper text and stderr can handle text correctly.

This also fixes 'b' prefix in error messages like:

    option zonemgr: b'empty DNS label'

Fixes: https://pagure.io/freeipa/issue/7711

### DIFF
--- a/ipaserver/install/bindinstance.py
+++ b/ipaserver/install/bindinstance.py
@@ -437,26 +437,32 @@ def zonemgr_callback(option, opt_str, value, parser):
     """
     if value is not None:
         # validate the value first
-        try:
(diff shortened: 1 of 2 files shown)


abbra:
LGTM. The encoding issue for py2 has no other solution so we have to take it.

rcritten:
master:

* 341a12054a1363d1a202da64b50b6d62b16c0436 Fix zonemgr encoding issue

PR #2408, opened 2018-09-28
Lint yaml and RPM spec

tiran:
Let's catch broken YAML files (Travis, PR-CI) and spec file early.

- Use rpmlint to detect syntax errors in spec file early
- Attempt to parse all YAML files with PyYAML

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -169,7 +169,7 @@ endif WITH_PYLINT
 if WITH_JSLINT
 JSLINT_TARGET = jslint
 endif WITH_JSLINT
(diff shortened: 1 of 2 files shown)


flo-renaud:
Hi @tiran 
thanks for the patch, works for me.
Note that yamllint will not detect when the test definition file contains an error like: `topology: master_1repl` instead of `topology: *master_1repl`. I opened an issue against freeipa-pr-ci to detect that type of problem: [issue 227](https://github.com/freeipa/freeipa-pr-ci/issues/227)

tiran:
master:

* 4b60bc38148a28caa0ba0ab705d88d16de3322d9 Lint yaml and RPM spec

PR #2410, opened 2018-09-30
Update translations from Zanata

abbra:
Update translations from Zanata to allow testing new translation IDs in Web UI


rcritten:
The agreement in the past is to simply accept the updates from Zanata. It goes through a PR because that's the process.

tiran:
I'm slightly worried that it bloats our git repo and renders history useless. For example Chinese has ``8,667 additions, 8,770 deletions``, Russian even ``24,575 additions, 23,500 deletions``.

rcritten:
Then bring it up on the devel list for broader discussion. This has been the policy for years.

abbra:
@tiran this particular update is huge because we did change translation IDs in Web UI en-masse. Thus, there are changes. I did run `make strip-po` as required by our release page: https://www.freeipa.org/page/Release, so the content is already reduced to what is absolutely necessary.

If you want to review, please pull the branch locally.

tiran:
@abbra The diffs are so humongous because the sorting of all entries has completely changed. Should we try to make the sorting stable to keep future diffs smaller?

abbra:
We do not have any influence over the sorting order in the POs generated by Zanata. So I hoped Zanata would keep the order stable.

abbra:
After discussing with @tiran on IRC, we decided to change the base by sorting in `make strip-po` target first, then importing from Zanata. 

Thus, there is a commit that adds `-s` option `msgattrib` call. Then a commit that establishes a predictable sorting for the base translations we have. Finally, we import Zanata updates with `make strip-po` target run to sort them too.

tiran:
@abbra is going to perform a manual backport of the translation to 4.7 branch.

tiran:
master:

* ef57be6163e74a8aa21707ae7daf73d16ed18884 When stripping PO files, sort the output
* 6f6cac04594a6b2d4a82a07873a16666628f89d6 Re-sort the translations before importing new ones from Zanata
* 60cb8979a5843e1913dd6c02953d6c20c884df90 Import updated translations from Zanata

PR #2414, opened 2018-10-01
UI-Test Automate: ID Range

serg-cymbaluk:
Web UI tests for "ID Ranges" functionality.

This PR doesn't include cases which require AD server for running. So task hasn't completed yet.

Ticket: https://pagure.io/freeipa/issue/7709

### DIFF
--- a/ipatests/test_webui/task_range.py
+++ b/ipatests/test_webui/task_range.py
@@ -41,6 +41,11 @@ class range_tasks(UI_driver):
         'New primary rid range overlaps with existing primary rid range.'
     )
 
+    SECONDARY_RID_RANGE_OVERLAPS_ERROR = (
+        'Constraint violation: '
+        'New secondary rid range overlaps with existing secondary rid range.'
+    )
+
     def get_shifts(self):
         result = self.execute_api_from_ui('idrange_find', [], {})
         idranges = result['result']['result']
@@ -104,6 +109,8 @@ class range_tasks(UI_driver):
         """
         Generate RangeAddFormData instance with initial data based on existing
         ID ranges.
+        :param kwargs: overrides default fields values (base_id, base_rid,
+                       secondary_base_rid)
         """
 
         shift = 100
--- a/ipatests/test_webui/task_range.py
+++ b/ipatests/test_webui/task_range.py
@@ -100,13 +100,19 @@ class range_tasks(UI_driver):
         base_id = kwargs.get('base_id', self.max_id + shift)
         self.max_id = base_id + size
 
-        base_rid = self.max_rid + shift
-        self.max_rid = base_rid + size
+        if 'base_rid' in kwargs:
+            base_rid = kwargs['base_rid']
+        else:
+            base_rid = self.max_rid + shift
+            self.max_rid = base_rid + size
 
         secondary_base_rid = None
         if not domain:
-            secondary_base_rid = base_rid + size + shift
-            self.max_rid = secondary_base_rid + size
+            if 'secondary_base_rid' in kwargs:
+                secondary_base_rid = kwargs['secondary_base_rid']
+            else:
+                secondary_base_rid = self.max_rid + shift
+                self.max_rid = secondary_base_rid + size
 
         return RangeAddFormData(
             pkey, base_id, base_rid,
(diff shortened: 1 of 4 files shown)


mrizwan93:
Hi @serg-cymbaluk 
It would be good if you also include test_webui/test_trust.py to your pr-ci template as there is change in that file too.

serg-cymbaluk:
@mrizwan93 

> Hi @serg-cymbaluk
> It would be good if you also include test_webui/test_trust.py to your pr-ci template as there is change in that file too.
```python
class test_trust(trust_tasks):
     def setup(self, *args, **kwargs):
        super(test_trust, self).setup(*args, **kwargs)
        if not self.has_trusts():
            self.skip('Trusts not configured')
```

Unfortunately **has_trusts** option is disabled on PR CI because there is no configured AD server, so all test_trust.py will be skipped.

ipatests/test_webui/task_range.py
```diff
@@ -67,50 +95,139 @@ def get_domain(self):
...
+    def get_add_form_data(self, pkey, range_type=LOCAL_ID_RANGE, size=50,
+                          domain=None, **kwargs):
+        """
+        Generate RangeAddFormData instance with initial data based on existing
+        ID ranges.
+        """
 
-        base_id = self.max_id + shift
+        shift = 100
+        base_id = kwargs.get('base_id', self.max_id + shift)
```
pvoborni: Does the flow of `base_id` and `secondary_base_rid` need to be controlled by kwargs? Usage of kwargs makes it harder to analyze method's behavior from a signature. In this case simple `is not None` check with defaults will make the same job. Same for occurrences bellow.

ipatests/test_webui/ui_driver.py
```diff
@@ -769,7 +769,8 @@ def fill_text(self, selector, value, parent=None):
             parent = self.get_form()
         tb = self.find(selector, By.CSS_SELECTOR, parent, strict=True)
         try:
-            tb.clear()
+            tb.send_keys(Keys.CONTROL + 'a')
+            tb.send_keys(Keys.DELETE)
```
pvoborni: Not sure if it might break other tests. It's probably fine. Might be worth to run not just range test suite in the temp commit.

ipatests/test_webui/task_range.py
serg-cymbaluk: There are three reasons for that:
1) `base_rid` and `secondary_base_rid` are not so common arguments
2) kwargs allow to forward arguments to other methods easily
3) kwargs allow to increase number of options (field values in this case) without breaking signature.
But for sure I need to describe kwargs options for `get_add_form_data` in docstring

ipatests/test_webui/ui_driver.py
serg-cymbaluk: Agree. I've added all WebUI tests to PR CI config, so we will see if something has broken.

serg-cymbaluk:
@pvoborni 

> "UI tests for "ID Range": fix PEP8 errors" commit should be squashed in the new commits which introduced those issues.

It is done.

> Not sure if it might break other tests

I am now re-running some tests which failed last run. Waiting for results. I guess it is just timeout issues but not regressions.

ipatests/test_webui/ui_driver.py
serg-cymbaluk: @pvoborni 
No regressions found.

pvoborni:
Changes look good to me, tests passed. Thus seems good  to me. Temp commit can be removed. After that final ack is on @mrizwan93 .

serg-cymbaluk:
master:

* e55d17d60980305797f373e4ae199f935033d587 UI tests for "ID Range": check range name and base ID duplication
* 369fb23ed9edaddea34ebceb7e1d7410c3b432ba UI tests for "ID Range": check adding range without primary and secondary RID bases
* b180991aca2f05277c753909227a15df214f7a8f UI tests for "ID Range": check primary RID base duplication
* dde4d19f7b40bfd96720945f374cbc18eb8b484b UI tests for "ID Range": check adding range with overlapping of existing local range
* 70f51c0db2948db8b65dfa9d3da7b3d4c1f9ad87 UI tests for "ID Range": check modifying ID range with invalid or missing values
* dd590e7ea7dfc7ea84580eb8368297c98adc8056 UI tests for "ID Range": - check creating ID range with special characters in name - check modifying ID range with existing secondary RID base
* 6595949e1d48ea9988cc6f99236c2a477ad048a3 UI tests for "ID Range": check creating ID Range with overlapping of primary and secondary RID base
* 2d845cc70ab71ffed8f48ec1730e4c6236bdf494 UI tests for "ID Range": check deleting primary local range

PR #2421, opened 2018-10-03
Update list of contributors 

abbra:
Another part of the preparation for the 4.7 release: update list of contributors.

I tried to sort the list of translators, thus there is a bit more churn in that part.

### DIFF
--- a/Contributors.txt
+++ b/Contributors.txt
@@ -100,6 +100,7 @@ Developers:
 	Jan Pazdziora
 	W. Michael Petullo
 	Pavel Picka
+	Orion Poplawski
 	Gowrishankar Rajaiyan
 	realsobek
 	Michal Reznik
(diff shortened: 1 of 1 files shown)


abbra:
Note that pagure's hook to push to github seems to be broken again, so you can see more commits in this PR than it actually has.

Contributors.txt
```diff
@@ -108,25 +109,29 @@ Developers:
...
 	Lenka Ryznarova
+	Alexander Scheel
 	Thorsten Scherf
 	shanyin
 	Kaleemullah Siddiqui
 	Michael Simacek
 	Lars Sjostrom
 	Filip Skola
 	Aleksei Slaikovskii
+	Alexey Slaykovsky
```
rcritten: Should we ask him which name he prefers?

abbra: @slaykovsky which name do you prefer to use in the Contributors list?

abbra: Ok, let me just drop the new addition for @slaykovsky  and keep using the old one right now.

flo-renaud:
master:

* 753264069f29e47bf222e50e95a7ec5849a7f6cb Update list of contributors

PR #2422, opened 2018-10-03
Test KRA installtion after ca agent cert renewal

mrizwan93:
KRA installtion was failing after ca-agent cert gets renewed.
This test check if the syptoms no longer exists.

related ticket: https://pagure.io/freeipa/issue/7288

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -10,6 +10,8 @@ installed.
 from __future__ import absolute_import
 
 import os
+from datetime import datetime, timedelta
+import time
 import pytest
 from ipalib.constants import DOMAIN_LEVEL_0
 from ipaplatform.constants import constants
@@ -18,6 +20,7 @@ from ipatests.pytest_ipa.integration.env_config import get_global_config
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
 from ipatests.test_integration.test_caless import CALessBase, ipa_certs_cleanup
+from ipalib import x509
 
 config = get_global_config()
 
@@ -486,3 +489,63 @@ class TestInstallMasterReservedIPasForwarder(IntegrationTest):
         exp_str = ("Invalid IP Address 0.0.0.0: cannot use IANA reserved "
                    "IP address 0.0.0.0")
         assert exp_str in cmd.stdout_text
+
+
+class TestKRAinstallAfterCertRenew(IntegrationTest):
+    """ Test KRA installtion after ca agent cert renewal
+
+    KRA installation was failing after ca-agent cert gets renewed.
+    This test checks if the symptoms no longer exist.
+
+    related ticket: https://pagure.io/freeipa/issue/7288
+    """
+
+    def test_KRA_install_after_cert_renew(self):
+
+        tasks.install_master(self.master)
+
+        # get ca-agent cert and load as pem
+        dm_pass = self.master.config.dirman_password
+        admin_pass = self.master.config.admin_password
+        args = [paths.OPENSSL, "pkcs12", "-in",
+                paths.DOGTAG_ADMIN_P12, "-nodes",
+                "-passin", "pass:{}".format(dm_pass)]
+        cmd = self.master.run_command(args)
+
+        certs = x509.load_certificate_list(cmd.stdout_text.encode('utf-8'))
+
+        # get expiry date of agent cert
+        cert_expiry = certs[0].not_valid_after
+
+        # move date to grace period so that certs get renewed
+        self.master.run_command(['systemctl', 'stop', 'chronyd'])
+        grace_date = cert_expiry - timedelta(days=10)
+        grace_date = datetime.strftime(grace_date, "%Y-%m-%d %H:%M:%S")
+        self.master.run_command(['date', '-s', grace_date])
+
+        # get the count of certs track by certmonger
+        cmd = self.master.run_command(['getcert', 'list'])
+        cert_count = cmd.stdout_text.count('Request ID')
+        timeout = 600
+        count = 0
+        start = time.time()
+        # wait sometime for cert renewal
+        while time.time() - start < timeout:
+            cmd = self.master.run_command(['getcert', 'list'])
+            count = cmd.stdout_text.count('status: MONITORING')
+            if count == cert_count:
+                break
+            time.sleep(100)
+        else:
+            # timeout
+            raise AssertionError('TimeOut: Failed to renew all the certs')
+
+        # move date after 3 days of actual expiry
+        cert_expiry = cert_expiry + timedelta(days=3)
+        cert_expiry = datetime.strftime(cert_expiry, "%Y-%m-%d %H:%M:%S")
+        self.master.run_command(['date', '-s', cert_expiry])
+
+        passwd = "{passwd}\n{passwd}\n{passwd}".format(passwd=admin_pass)
+        self.master.run_command(['kinit', 'admin'], stdin_text=passwd)
+        cmd = self.master.run_command(['ipa-kra-install', '-p', dm_pass, '-U'])
+        self.master.run_command(['systemctl', 'start', 'chronyd'])


celestian:
@mrizwan93 
If delay is needed, please, use time.sleep(100). Or... and this should be better, is there any way how to check, if certs are ready? If yes, add loop which check it and when pass, test can continue.

ipatests/test_integration/test_installation.py
```diff
@@ -479,3 +481,51 @@ def test_reserved_ip_as_forwarder(self):
...
+                                       p12.get_certificate())
+
+        # get expiry date of agent cert
+        x509 = crypto.load_certificate(crypto.FILETYPE_PEM, cert)
+        cert_expiry = x509.get_notAfter()
+
+        # remove the 'Z' from the date string
+        cert_expiry = cert_expiry[:-1].decode('utf-8')
+        py_date_obj = datetime.strptime(cert_expiry, "%Y%m%d%H%M%S")
+
```
celestian: Is this needed? I know, Z could look strange, but it is zero, it means UTC. I searched for it and I found out:
https://stackoverflow.com/questions/19654578/python-utc-datetime-objects-iso-format-doesnt-include-z-zulu-or-zero-offset

ipatests/test_integration/test_installation.py
```diff
@@ -479,3 +481,51 @@ def test_reserved_ip_as_forwarder(self):
...
+        grace_date = py_date_obj - timedelta(days=20)
+        grace_date = datetime.strftime(grace_date, "%Y-%m-%d %H:%M:%S")
+        self.master.run_command(['date', '-s', grace_date])
+
+        # check if all certs renewed
+        count = 0
+        while (count != 9):
+            cmd = self.master.run_command(['getcert', 'list'])
+            count = cmd.stdout_text.count('status: MONITORING')
+
```
celestian: I don't understand, why is this while set precisely to count !=9, please?

mrizwan93: This is to check if all certs got renewed. Cert count is 9, that's why it is checking the status count for 9. If count is 9, it means certs got renewed and we are good go next step.

ipatests/test_integration/test_installation.py
mrizwan93: The expiry date was having Z in the string and was failing my side. That's why this workaround.

celestian: @mrizwan93  What about this :-)
```
py_date_obj = datetime.strptime(date, "%Y%m%d%H%M%SZ")
```
Just add Z to the format string, then you can remove line '512' and '513'.

There is an example that it is working:
```
date = "20131029093841Z"
py_date_obj = datetime.strptime(date, "%Y%m%d%H%M%SZ")
print(py_date_obj)
```

ipatests/test_integration/test_installation.py
```diff
@@ -479,3 +481,51 @@ def test_reserved_ip_as_forwarder(self):
...
+        p12_content = self.master.get_file_contents(paths.DOGTAG_ADMIN_P12)
+        admin_pass = self.master.config.admin_password
+
+        # convert cert to pem format
+        p12 = crypto.load_pkcs12(p12_content, admin_pass)
+        cert = crypto.dump_certificate(crypto.FILETYPE_PEM,
+                                       p12.get_certificate())
+
+        # get expiry date of agent cert
+        x509 = crypto.load_certificate(crypto.FILETYPE_PEM, cert)
```
stlaz: We have python-cryptography for that.

ipatests/test_integration/test_installation.py
stlaz: It does not matter, use python-cryptography and you don't have to do any of this at all.

ipatests/test_integration/test_installation.py
```diff
@@ -479,3 +481,51 @@ def test_reserved_ip_as_forwarder(self):
...
+        cert_expiry = x509.get_notAfter()
+
+        # remove the 'Z' from the date string
+        cert_expiry = cert_expiry[:-1].decode('utf-8')
+        py_date_obj = datetime.strptime(cert_expiry, "%Y%m%d%H%M%S")
+
+        # move date to grace period so that certs get renewed
+        grace_date = py_date_obj - timedelta(days=20)
```
stlaz: 20 is a magical number.

ipatests/test_integration/test_installation.py
```diff
@@ -479,3 +481,51 @@ def test_reserved_ip_as_forwarder(self):
...
+
+    related ticket: https://pagure.io/freeipa/issue/7288
+    """
+
+    def test_KRA_install_after_cert_renew(self):
+
+        tasks.install_master(self.master)
+
+        # get ca-agent cert contents
+        p12_content = self.master.get_file_contents(paths.DOGTAG_ADMIN_P12)
```
stlaz: See `ipaserver/install/certs.py.pkcs12_to_certkeys()` to get the command you need to call on master, get its output and convert it to cryptography objects in the same way. Remove `OpenSSL` library import.

ipatests/test_integration/test_installation.py
```diff
@@ -479,3 +481,51 @@ def test_reserved_ip_as_forwarder(self):
...
+        py_date_obj = datetime.strptime(cert_expiry, "%Y%m%d%H%M%S")
+
+        # move date to grace period so that certs get renewed
+        grace_date = py_date_obj - timedelta(days=20)
+        grace_date = datetime.strftime(grace_date, "%Y-%m-%d %H:%M:%S")
+        self.master.run_command(['date', '-s', grace_date])
+
+        # check if all certs renewed
+        count = 0
+        while (count != 9):
```
stlaz: Needs an upper wait bound.
9 is a magical number.

ipatests/test_integration/test_installation.py
```diff
@@ -479,3 +481,51 @@ def test_reserved_ip_as_forwarder(self):
...
+    related ticket: https://pagure.io/freeipa/issue/7288
+    """
+
+    def test_KRA_install_after_cert_renew(self):
+
+        tasks.install_master(self.master)
+
+        # get ca-agent cert contents
+        p12_content = self.master.get_file_contents(paths.DOGTAG_ADMIN_P12)
+        admin_pass = self.master.config.admin_password
```
stlaz: Are you sure it's admin password which protects the file? Isn't it the DM password by any chance?

ipatests/test_integration/test_installation.py
```diff
@@ -479,3 +483,43 @@ def test_reserved_ip_as_forwarder(self):
...
+
+    def test_KRA_install_after_cert_renew(self):
+
+        tasks.install_master(self.master)
+
+        # get ca-agent cert contents
+        p12_content = self.master.get_file_contents(paths.DOGTAG_ADMIN_P12)
+
+        # load cert
+        cert = x509.load_der_x509_certificate(p12_content,
```
stlaz: You can't stick pkcs12 data in this method, it requires DER-formatted certificate. This is the code I was referring to before:
https://github.com/freeipa/freeipa/blob/1ef0fe8bb824282c2f48417efda3a60e7c1bf580/ipaserver/install/certs.py#L106-L115

ipatests/test_integration/test_installation.py
```diff
@@ -479,3 +483,43 @@ def test_reserved_ip_as_forwarder(self):
...
+        # get expiry date of agent cert
+        cert_expiry = cert.not_valid_after()
+
+        # move date to grace period so that certs get renewed
+        grace_date = cert_expiry - timedelta(days=20)
+        grace_date = datetime.strftime(grace_date, "%Y-%m-%d %H:%M:%S")
+        self.master.run_command(['date', '-s', grace_date])
+
+        # wait sometime for cert renewal
+        sleep(200)
```
stlaz: This brings non-determinism to the test - how can you be sure that the certificates are renewed after 200 seconds? You need to check that, otherwise you don't know whether you're testing what you actually wanted to test in the first place.

ipatests/test_integration/test_installation.py
```diff
@@ -479,3 +483,43 @@ def test_reserved_ip_as_forwarder(self):
...
+
+        # move date after 3 days of actual expiry
+        cert_expiry = cert_expiry + timedelta(days=3)
+        cert_expiry = datetime.strftime(cert_expiry, "%Y-%m-%d %H:%M:%S")
+        self.master.run_command(['date', '-s', cert_expiry])
```
stlaz: I wonder what happens in Kerberos with all those time moves :thinking: Maybe something to look at should the test fail.

ipatests/test_integration/test_installation.py
```diff
@@ -479,3 +483,43 @@ def test_reserved_ip_as_forwarder(self):
...
+
+        dm_pass = self.master.config.dirman_password
+        self.master.run_command(['ipa-kra-install', '-p', dm_pass])
```
stlaz: Shouldn't it be run in unattended mode?

ipatests/test_integration/test_installation.py
celestian: @stlaz Could you please provide some hint, how to check it?

celestian: @stlaz  Sorry, I was confused by github UI. This code is not recent. So I close this thread.

celestian: @stlaz  More confused then I thought. This is still valid, because IT IS CURRENT code. So, @stlaz  could you please provide any hint here?

stlaz: @celestian I guess a check that the expiration period of the certificates in question changed?

ipatests/test_integration/test_installation.py
stlaz: Also, I just remembered that changing time in the RHEV lab VM machines might brick your access to the them (at least it was doing that back when I was playing with time-based rules). Keep that in mind (and possibly check that).

ipatests/test_integration/test_installation.py
celestian: @stlaz  good point, @mrizwan93 if you agree, could you please implemented it?

celestian:
@mrizwan93 is this patch set correct? I see the parts which were fixed already.

mrizwan93:
@celestian I didn't get you.

celestian:
@mrizwan93 I am afraid, that you force pushed old version of your patch set. I still see the old code. For example, there is still:

    while(count != 9):

which was discussed.

mrizwan93:
Yes, 9 is a magical number. But that is the no. of certs we get after installation. And I needed something to make sure that all the certs got renewed. ``time.sleep()` was not serving the purpose, that's why I re-introduced these lines.

celestian:
@mrizwan93 So, the certificate is a unique entity. They should differentiate by something: type, name, ... Please, is possible to check them by this unique category or not? Because if it is possible, it will be definitely better than magic number.
@stlaz, @pvoborni could you please help us find the way how to avoid magic number in this test?

Tiboris:
@mrizwan93, what was the problem with using python-cryptography?

mrizwan93:
@Tiboris Nothing. `ipalib.x509` which uses python-cryptography. 

Problem here is that we want to avoid using magical no. 9, which is the no. of certs created by ipa-server installation.

abbra:
Perhaps, it is better to replace this magical number by a verification that specific locations are tracked in certmonger. You can do that with `getcert list -d /path/to/nssdb` or `getcert list -f /path/to/cert.crt`.
So I'd suggest you convert the code into a series of checks that expect the locations to resolve via those commands:

```
locations = (('PKINIT', '-f', '/var/kerberos/krb5kdc/kdc.crt'),
                    ('LDAP', '-d', '/etc/dirsrv/slapd-H-VDA-LI'),
                    ('HTTP', '-f', '/var/lib/ipa/certs/httpd.crt'),
                    ...
                    ('CA', '-d', '/etc/pki/pki-tomcat/alias'))
```
This is just a sketch, you will probably need to account for the cases where we use the same NSS database to store multiple certificates too.

celestian:
Hello @mrizwan93  any progress here?

ipatests/test_integration/test_installation.py
```diff
@@ -461,3 +464,54 @@ def test_reserved_ip_as_forwarder(self):
...
+
+        certs = x509.load_certificate_list(cmd.stdout_text.encode('utf-8'))
+
+        # get expiry date of agent cert
+        cert_expiry = certs[0].not_valid_after
+
+        # move date to grace period so that certs get renewed
+        grace_date = cert_expiry - timedelta(days=10)
+        grace_date = datetime.strftime(grace_date, "%Y-%m-%d %H:%M:%S")
+        self.master.run_command(['date', '-s', grace_date])
```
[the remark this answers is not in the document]
flo-renaud: The logs for [ipaserver-install.log](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/8a047738-f2df-11e8-908e-fa163e3bb4ee/test_installation.py-TestKRAinstallAfterCertRenew--test_KRA_install_after_cert_renew/master.ipa.test/var/log/ipaserver-install.log.gz) show that chrony is setup. I really think the test should stop chrony before playing with the date.

flo-renaud:
Hi @mrizwan93 
thanks for the updated PR, looks good to me. Can you remove the temp commit, and then I will provide the ACK label?

mrizwan93:
@flo-renaud Thanks, temp commit removed.

flo-renaud:
master:

* b7ae9f7a3f577a61c97953da7e65d09349053380 Test KRA installtion after ca agent cert renewal

PR #12, opened 2016-08-23
Tests: Duplicate declaration on variables in ID views tests

mirielka:
In ipatests/test_xmlrpc/test_idviews_plugin several variables are declared
twice, while never using the first declaration. The duplicate declaration is
hereby removed.

https://fedorahosted.org/freeipa/ticket/6246

PR #2423, opened 2018-10-03
certdb: provide meaningful err msg for wrong PIN

flo-renaud:
ipa-server-install or ipa-replica-install do not provide a meaningful error message in CA-less mode when the install fails because of a wrong PIN.
Update the err msg so that it provides a hint to the user.

Update the associated tests with the correct message and remove the pytest.mark.xfail annotation.

Fixes https://pagure.io/freeipa/issue/5378

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -221,6 +221,24 @@ KEY_RE = re.compile(
 )
 
 
+class Pkcs12ImportIncorrectPasswordError(RuntimeError):
+    """ Raised when import_pkcs12 fails because of a wrong password.
+    """
+    pass
+
+
+class Pkcs12ImportOpenError(RuntimeError):
+    """ Raised when import_pkcs12 fails trying to open the file.
+    """
+    pass
+
+
+class Pkcs12ImportUnknownError(RuntimeError):
+    """ Raised when import_pkcs12 fails because of an unknown error.
+    """
+    pass
+
+
 class NSSDatabase:
     """A general-purpose wrapper around a NSS cert database
 
@@ -578,13 +596,15 @@ class NSSDatabase:
         try:
             self.run_pk12util(args)
         except ipautil.CalledProcessError as e:
-            if e.returncode == 17:
-                raise RuntimeError("incorrect password for pkcs#12 file %s" %
-                    pkcs12_filename)
+            if e.returncode == 17 or e.returncode == 18:
+                raise Pkcs12ImportIncorrectPasswordError(
+                    "incorrect password for pkcs#12 file %s" % pkcs12_filename)
             elif e.returncode == 10:
-                raise RuntimeError("Failed to open %s" % pkcs12_filename)
+                raise Pkcs12ImportOpenError(
+                    "Failed to open %s" % pkcs12_filename)
             else:
-                raise RuntimeError("unknown error import pkcs#12 file %s" %
+                raise Pkcs12ImportUnknownError(
+                    "unknown error import pkcs#12 file %s" %
                     pkcs12_filename)
         finally:
             if pkcs12_password_file is not None:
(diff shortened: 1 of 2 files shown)


ipapython/certdb.py
```diff
@@ -749,7 +750,8 @@ def import_files(self, files, import_keys=False, key_password=None,
 
                     continue
 
-            raise RuntimeError("Failed to load %s" % filename)
+            raise RuntimeError(
+                "Failed to load %s, please check key password" % filename)
```
frasertweedale: It seems that we can reach here when:
- the file is a PKCS #12 file but the password is wrong, or
- the file was not a valid PKCS #12 (or any of the preceding formats that were tried).

Ideally we should distinguish these cases.  It would not be helpful to suggest that the problem might be an invalid passphrase, when in fact the file is an unrecognised format.  `import_pkcs12` internals already distinguish the invalid password case from other cases (based on `pk12util` exit status).  So, would it be better to allow the exceptions thrown by `import_pkcs12` to bubble up, instead of suppressing it (line 727)?  The messages of the exceptions thrown by `import_pkcs12` can be tweaked for consistency with the other errors.

flo-renaud: Thanks for your comments, I had missed that point.

ipapython/certdb.py
```diff
@@ -722,8 +722,13 @@ def import_files(self, files, import_keys=False, key_password=None,
             if import_keys:
                 try:
                     self.import_pkcs12(filename, key_password)
-                except RuntimeError:
-                    pass
+                except RuntimeError as e:
+                    if "unknown error import pkcs#12 file" in str(e):
```
frasertweedale: @flo-renaud sorry to be a nit-pick, but I'm not comfortable with this "stringly-typed" approach.  Later when someone updates one string they will easily miss the other, and break the program.  We could create new exception subclasses to convey the errors from `import_pkcs12`, and use different `except` blocks to handle the different cases.

Apart from that the logic now seems fine to me, and I am glad you added comments about it :)

flo-renaud: Thanks for your comment. You are right, it is cleaner to raise subclasses of RuntimeError and will be easier to maintain. I just pushed a new version.

ipapython/certdb.py
```diff
@@ -578,7 +599,7 @@ def import_pkcs12(self, pkcs12_filename, pkcs12_passwd=None):
         try:
             self.run_pk12util(args)
         except ipautil.CalledProcessError as e:
-            if e.returncode == 17:
+            if e.returncode == 17 or e.returncode == 18:
```
frasertweedale: The import code (~L746) is calling `import_pkcs12` (this method), so don't we need to raise the new exceptions from this method also?

flo-renaud: Pff, I am probably tired at the end of the week... I intended to modify import_pkcs12, not export_pkcs12. Thanks for noticing my error.

frasertweedale:
@flo-renaud OK I've picked enough nits.  You have my ACKs!

frasertweedale:
I'm not sure what the ACK protocol is when there are temp commits that twiddle the tests... consider the first two commits ACKed anyhow.

flo-renaud:
Hi @frasertweedale 
thanks for the review. I removed the temp commit, let's wait for the tests to finish.

flo-renaud:
Tests are green, setting the ACK label as discussed in [above comment](https://github.com/freeipa/freeipa/pull/2423#issuecomment-427789352).

flo-renaud:
master:

* 9e1c26c755dd4ddf390d775a743105a8f6a2c460 certdb: provide meaningful err msg for wrong PIN
* 6650d1b5631f77c0f07dd572bc27cdb98378997d ipa tests: CA less

PR #2433, opened 2018-10-08
Bump requires 389-ds-base

flo-renaud:
ipa-replica-install sometimes fails with
```
[28/41]: setting up initial replication
Starting replication, please wait until this has completed.
[ldap://master.ipa.test:389] reports: Replica Busy! Status: [Error (1) Replication error acquiring replica: replica busy]
 [error] RuntimeError: Failed to start replication
```
which is caused by a 389-ds issue
(https://pagure.io/389-ds-base/issue/49818)
Bump requires to include the fix.

Fixes: https://pagure.io/freeipa/issue/7642

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -48,7 +48,9 @@
 %global selinux_policy_version 3.14.1-14
 %global slapi_nis_version 0.56.1-4
 %global python_ldap_version 3.1.0-1
-%global ds_version 1.4.0.8-1
+# Fix for "Installation fails: Replica Busy"
+# https://bugzilla.redhat.com/show_bug.cgi?id=1598478
+%global ds_version 1.3.8.4-15
 %else
 # Fedora
 %global package_name freeipa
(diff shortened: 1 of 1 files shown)


rcritten:
This is going backwards a version. In the master branch of 389-ds-base it is first fixed in tag 1.4.0.12.

tiran:
> This is going backwards a version. In the master branch of 389-ds-base it is first fixed in tag 1.4.0.12.

I was confused for a second, too. Then I realized that the ``1.3.8.4-15`` is for RHEL 7.

flo-renaud:
Yes, I'm trying to set the versions for fedora and rhel as it help during RHEL builds. Thanks for the review.

tiran:
Since the issue affects RHEL 7, I'm also backporting the commit to 4.6.

tiran:
master:

* 3eac88a433744f6b1597b54528c729976b1349d4 Bump requires 389-ds-base

PR #2434, opened 2018-10-08
UI-Test Automate: Automember

serg-cymbaluk:
Web UI tests for "Automember" functionality.

Ticket: https://pagure.io/freeipa/issue/7721

### DIFF
--- a/ipatests/test_webui/test_automember.py
+++ b/ipatests/test_webui/test_automember.py
@@ -49,6 +49,10 @@ HOST_GROUP_DATA = {
     ],
 }
 
+# Condition types
+INCLUSIVE = 'inclusive'
+EXCLUSIVE = 'exclusive'
+
 
 @pytest.mark.tier1
 class TestAutomember(UI_driver):
@@ -122,6 +126,60 @@ class TestAutomember(UI_driver):
         self.delete(ENTITY, [{'pkey': pkey} for pkey in pkeys],
                     facet='searchhostgroup')
 
+    def add_conditions(self, conditions, condition_type):
+        """
+        Add conditions to a rule
+
+        :param conditions: list of conditions where condition is a pair
+                           (attribute, expression)
+        :param condition_type: can be 'inclusive' or 'exclusive'
+        """
+
+        name = 'automember{}regex'.format(condition_type)
+
+        attribute, expression = conditions[0]
+        another_conditions = conditions[1:]
+        another_conditions.reverse()
+
+        self.add_table_record(name, {'fields': [
+            ('selectbox', 'key', attribute),
+            ('textbox', name, expression)
+        ]}, add_another=bool(another_conditions))
+
+        while another_conditions:
+            attribute, expression = another_conditions.pop()
+            self.add_another_table_record(
+                {'fields': [
+                    ('selectbox', 'key', attribute),
+                    ('textbox', name, expression)
+                ]},
+                add_another=bool(another_conditions)
+            )
+
+    def delete_conditions(self, conditions, condition_type):
+        """
+        Delete rule conditions
+
+        :param conditions: list of conditions where condition is a pair
+                           (attribute, expression)
+        :param condition_type: can be 'inclusive' or 'exclusive'
+        """
+
+        self.delete_record(
+            ['{}={}'.format(attr, exp) for attr, exp in conditions],
+            parent=self.get_facet(),
+            table_name='automember{}regex'.format(condition_type)
+        )
+
+    def open_new_condition_dialog(self, condition_type):
+        table = self.find_by_selector(
+            "table[name='automember{}regex'].table".format(condition_type),
+            strict=True
+        )
+        btn = self.find_by_selector(".btn[name=add]", table, strict=True)
+        btn.click()
+        self.wait()
+
     @screenshot
     def test_crud(self):
         """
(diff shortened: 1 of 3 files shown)


ipatests/test_webui/test_automember.py
```diff
@@ -267,6 +354,303 @@ def test_rebuild_membership_users(self):
...
+    def test_search_user_group_rule(self):
+        """
+        Test searching user group rules using filter
+        """
+
+        pkey = 'search123'
+        self.add_user_group(pkey, '')
+        self.add_user_group_rules(pkey)
+
+        for text in ['search123', 'search', ' search ', 'SEARCH', '123']:
```
mrizwan93: does it make sense to split `' search '` into `' search'` and `'search '` i.e separate check for leading and trailing space?
IMO, in case of `' search '` only one condition will be checked (i.e either leading or trailing may be).

I would like to check behaviour with special char too like `#$2` etc.

ipatests/test_webui/test_automember.py
```diff
@@ -50,16 +49,157 @@
...
+            ENTITY,
+            [{
+                'pkey': pkey,
+                'add': [('combobox', 'cn', pkey)],
+            } for pkey in pkeys],
+            facet='searchhostgroup',
+            **kwargs
+        )
+
+    def add_user(self, pkey, name, surname):
```
mrizwan93: Does it make sense to move this and similar method from this file to `ui_driver.py`?

ipatests/test_webui/test_automember.py
serg-cymbaluk: Yes, sure. I'll add some additional cases.

ipatests/test_webui/test_automember.py
serg-cymbaluk: I'd create separated class for such methods. But currently it is impossible without changing design of base class. UI_driver should be rewritten and I try to not extend it yet.

ipatests/test_webui/test_automember.py
serg-cymbaluk: Done.

ipatests/test_webui/test_automember.py
```diff
@@ -267,6 +377,313 @@ def test_rebuild_membership_users(self):
...
+        self.delete_users('dev1', 'dev2')
+        self.delete_user_group_rules('devel')
+
+    @screenshot
+    def test_add_multiple_user_group_rules(self):
+        """
+        Test creating and deleting multiple user group rules
+        """
+
+        groups = ['group1', 'group2', 'group3']
```
mrizwan93: could you consider adding negative tests for adding user group rule and host group rule? like you added for search feature.

serg-cymbaluk: There are several negative test cases for group rules:
test_add_user_group_rule_with_no_group
test_add_host_group_rule_with_no_group
test_add_user_group_rules_for_same_group
test_add_host_group_rules_for_same_group

serg-cymbaluk: It is impossible because I need to add group with such name. But it is invalid name. So we have got no group and nothing to test.

mrizwan93: That is what I would like to see in negative tests i.e invalid name should not be accepted as group name. We can add additional tests for them.

serg-cymbaluk: But it is not "Groups" tests, it is "Automember" tests.
For "Groups" we already have https://github.com/freeipa/freeipa/blob/master/ipatests/test_webui/test_group.py

mrizwan93: ahh, sorry. I missed that.

serg-cymbaluk:
@mrizwan93 
Do you have some other remarks?

mrizwan93:
LGTM. But someone else should review too.

ipatests/test_webui/test_automember.py
```diff
@@ -267,6 +377,313 @@ def test_rebuild_membership_users(self):
...
+        self.add_user_group_rules(pkey)
+        self.navigate_to_record(pkey)
+        self.mod_record(ENTITY, {'mod': [
+            ('textarea', 'description', SEARCH_CASES['description']),
+        ]}, facet='usergrouprule')
+        self.navigate_to_entity(ENTITY, facet='searchgroup')
+
+        for text in SEARCH_CASES['positive']:
+            self.apply_search_filter(text)
+            self.wait(0.5)
```
pvoborni: shouldn't it wait for something more specific, e.g. hiding of activity indicator (`wait_for_request()`)? What if the tested machine is slower for some reason and the search takes longer than 0.5s? Same in other occurances below.

serg-cymbaluk: Fixed

serg-cymbaluk:
master:

* 49df1ab1c847b342ded475c2df6545e4e9b76793 UI tests for "Automember": check search filter
* 63cbf6294eb2c201b49678d211528c020317e71e UI tests for "Automember": check creating and deleting of multiple rules
* a68035dcae239644eb7c75bfdcf0c5832c1eeb21 UI tests for "Automember": check creating and deleting of automember rule conditions
* 33a74fb2c45dc23e4792a70c586286941e37989a UI tests for "Automember": check setting default user/host group
* cd795257cc7b88e0d93201271d03184a33fa9653 UI tests for "Automember": Negative cases
* 8949aa64d991305007a589bc84fd59a484d93301 UI tests for "Automember": Extend search cases

PR #2437, opened 2018-10-09
ipatests: mark known failures as xfail 

flo-renaud:
Commit 5dbcc1a marked the base class method `test_replica0_with_ca_kra_dns_install` as known failure, but this does not work with inherited classes. The child class methods need to be marked
themselves as known failures with `@pytest.mark.xfail`.

Furthermore, `TestInstallWithCA_KRA2` and `TestInstallWithCA_KRA_DNS2` tests should succeed because the master is installed with KRA (issue 7651 is related to replica install with --setup-kra
when it is the first KRA instance).

Related to https://pagure.io/freeipa/issue/7651

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -104,7 +104,6 @@ class InstallTestBase2(IntegrationTest):
     def install(cls, mh):
         tasks.install_master(cls.master, setup_dns=False)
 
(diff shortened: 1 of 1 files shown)


flo-renaud:
Removed the temp commit.

flo-renaud:
master:

* f6793043ecc0e0d04408616c856200b1462d0c0a ipatests: mark known failures as xfail

PR #2446, opened 2018-10-11
Fix misleading errors during client install rollback

rcritten:
Some incorrect errors are possible if a client installation
fails and a configuration rollback is required.

These include:

1. Unconfigured automount client failed: CalledProcessError(Command
['/usr/sbin/ipa-client-automount', '--uninstall', '--debug']
returned non-zero exit status 1: '')

Caused by check_client_configuration() not returning the correct
return value (2).

2. WARNING: Unable to revert to the pre-installation state ('authconfig'
tool has been deprecated in favor of 'authselect'). The default sssd
profile will be used instead.
The authconfig arguments would have been: authconfig --disableldap
--disablekrb5 --disablesssdauth --disablemkhomedir

If installation fails before SSSD is configured there is no state
to roll back to. Detect this condition.

3. An error occurred while removing SSSD's cache.Please remove the
cache manually by executing sssctl cache-remove -o.

Again, if SSSD is not configured yet then there is no cache to
remove. Also correct the missing space after the period.

https://pagure.io/freeipa/issue/7729

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -3282,13 +3282,14 @@ def uninstall(options):
     remove_file(paths.SSSD_MC_GROUP)
     remove_file(paths.SSSD_MC_PASSWD)
 
-    try:
-        run([paths.SSSCTL, "cache-remove", "-o", "--stop", "--start"])
-    except Exception:
-        logger.info(
-            "An error occurred while removing SSSD's cache."
-            "Please remove the cache manually by executing "
-            "sssctl cache-remove -o.")
+    if was_sssd_installed:
+        try:
+            run([paths.SSSCTL, "cache-remove", "-o", "--stop", "--start"])
+        except Exception:
+            logger.info(
+                "An error occurred while removing SSSD's cache."
+                "Please remove the cache manually by executing "
+                "sssctl cache-remove -o.")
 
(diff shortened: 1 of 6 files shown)


tiran:
Two tests are failing

* test_replica_promotion_TestWrongClientDomain fails with error code ``3`` instead of expected ``1``
* test_install_client_no_preconfigured_profile expects a different error message

rcritten:
Yeah, it's a case of "works for me locally". Error code 3 means (or should mean) already configured which makes no sense since I added in an explicit uninstall (which returns 2, not configured). A head scratcher.

rcritten:
Added commit to capture uninstall logs. This is a good idea anyway but in this case I hope it will show me what is going on.

tiran:
rerun, ``fedora-28/test_http_kdc_proxy`` has failed with ``RunPytest Failed to publish artifacts ``

tiran:
master:

* 235b5bd643947eee4e31c0cfed1729b7d945719c Remove the authselect profile warning if sssd was not configured.
* e59ee6099f065c2155e3be13025459108daf7900 Fix misleading errors during client install rollback
* 552777293d7c0741e7f9302088ebccf1ee7ba27e Collect the client and server uninstall logs in tests

PR #2447, opened 2018-10-11
Enable LDAP debug output in client to display TLS errors in join

rcritten:
If ipa-join fails due to a TLS connection error when doing an
LDAP-based enroll then nothing is logged by default except an
Invalid Password error which is misleading (because the failure
occurs during the bind).

The only way that debugging would have been sufficient is if
the user passed --debug to ipa-client-install which is not great.

This log level is otherwise very quiet and only logs one or two
lines on errors which is perfect.

https://pagure.io/freeipa/issue/7728

### DIFF
--- a/client/ipa-join.c
+++ b/client/ipa-join.c
@@ -197,33 +197,31 @@ callRPC(char * user_agent,
 
 /* The caller is responsible for unbinding the connection if ld is not NULL */
 static LDAP *
-connect_ldap(const char *hostname, const char *binddn, const char *bindpw) {
+connect_ldap(const char *hostname, const char *binddn, const char *bindpw,
+             int *ret) {
     LDAP *ld = NULL;
-    int ret;
-    int ldapdebug = 0;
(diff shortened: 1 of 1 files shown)


tiran:
Note to self, ``2`` is ``LDAP_DEBUG_PACKETS``. The man page https://linux.die.net/man/3/ldap_set_option mentions multiple ``LDAP_DEBUG_`` constants, however the constants are not available in any public header file. ``ldap_log.h`` isn't installed to ``/usr/include``.

rcritten:
Yes, I found exactly the same thing. I had set the 2 "magic" value many years ago. I can't explain it now other than to say in my testing it wasn't spitting out extraneous things in any failure or success modes that I tried. It may be because of the operations being performed.

rcritten:
master:

* 1e76f100a596cfd22c84f873e832f22752f31be0 Enable LDAP debug output in client to display TLS errors in join

PR #2448, opened 2018-10-12
ipa-advise: update url of cacerdir_rehash tool

pvoborni:
On legacy systems which don't have cacerdir_rehash tool (provided by authconfig)
the generated advise script downloads this tool from project page and uses it.

After decommision of Fedorahosted and move of authconfig project to Pagure,
this url was not updated in FreeIPA project.

This patch updates the url.

https://pagure.io/freeipa/issue/7731

### DIFF
--- a/ipaserver/advise/plugins/legacy_clients.py
+++ b/ipaserver/advise/plugins/legacy_clients.py
@@ -28,6 +28,9 @@ from ipapython.ipautil import template_file
 
 register = Registry()
 
+CACERTDIR_REHASH_URL = ('https://pagure.io/authconfig/raw/master/f/'
+                        'cacertdir_rehash')
+
 
 class config_base_legacy_client(Advice):
     def get_uri_and_base(self):
@@ -50,8 +53,6 @@ class config_base_legacy_client(Advice):
                          'location. If this value is different on your system '
                          'the script needs to be modified accordingly.\n')
 
-        cacertdir_rehash = ('https://fedorahosted.org/authconfig/browser/'
-                            'cacertdir_rehash?format=txt')
         self.log.comment('Download the CA certificate of the IPA server')
         self.log.command('mkdir -p -m 755 /etc/openldap/cacerts')
         self.log.command('curl http://%s/ipa/config/ca.crt -o '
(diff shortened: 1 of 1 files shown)


pvoborni:
The patch doesn't include a test. Though it is written in a way that a test can be created easily by importing the url, downloading the file and checking it's content (e.g. if it bash script). It was not done because it would introduce a dependency of the test on internet access (not sure if it is somewhere). From this perspective, it might be better to not automate.

pvoborni:
Though it doesn't have tests, the new behavior is visible in test_advise output. http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/a41b5ffe-ce1f-11e8-8e7f-fa163ee50fe4/runner.log.gz

tiran:
Why does the script have to download an external file? Cert directory rehashing can be performed with ``openssl rehash`` with recent versions of OpenSSL or ``c_rehash`` program with older versions.

pvoborni:
The explanation when it was introduced in fc3f3c90b94ef6d4d8f1e109655177412d52f36e was:


    Add ipa-advise plugins for legacy clients
    
    Old versions of SSSD do not directly support cross-realm trusts between IPA
    and AD. This patch introduces plugins for the ipa-advise tool, which should
    help with configuring an old version of SSSD (1.5-1.8) to gain access to
    resources in trusted domain.
    
    Since the configuration steps differ depending on whether the platform includes
    the authconfig tool, two plugins are needed:
    
    * config-redhat-sssd-before-1-9 - provides configuration for Red Hat based
      systems, as these system include the autconfig utility
    * config-generic-sssd-before-1-9 - provides configuration for other platforms
    
    https://fedorahosted.org/freeipa/ticket/3671
    https://fedorahosted.org/freeipa/ticket/3672

Review from that time ( https://www.redhat.com/archives/freeipa-devel/2013-August/msg00006.html )mentions that it was heavily inspired by: https://fedoraproject.org/wiki/QA:Testcase_freeipa_use_legacy_sssd_to_give_access_to_trusted_domain_users 

By brief look I don't see the downloading part mentioned. The obvious guess is that it tried to follow the test case and make it work even if authconfig is not installed. OpenSSL was probably not considered.

I don't know if it can be ensured that c_rehash will be present in all circumstances.

My take (assumption) is that this advise script is used rarely that it is not worth to investigate replacing the download part with an alternative.

PR #2449, opened 2018-10-12
Remove tests which install KRA on replica w/o KRA on master

rcritten:
The KRA installation code explicity quits if trying to
install a KRA during a replica installation if there is not
already a KRA in the topology.

A KRA can be added afterward.

https://pagure.io/freeipa/issue/7651


tiran:
Fantastic!

tiran:
master:

* 5466eca048530ccc26deb31a5b30005f550c37b5 Remove tests which install KRA on replica w/o KRA on master

PR #2450, opened 2018-10-12
Move ipa's systemd tmpfiles from /var/run to /run

stanislavlevin:
systemd 239 complains about the legacy of ipa's tmpfiles which
are located on /var/run.

But this can break compatibility.

Fixes: https://pagure.io/freeipa/issue/7732


tiran:
master:

* 531eca43086915a7ac4b83f7b5c77a6b5735376e Move ipa's systemd tmpfiles from /var/run to /run

PR #2460, opened 2018-10-16
ipatests: fix path in expected error message

flo-renaud:
The test is putting server.p12 / replica.p12 in the test_dir directory,
and the error message is printing the file name with its full path.

Related to https://pagure.io/freeipa/issue/5378


flo-renaud:
Temp commit removed.

flo-renaud:
master:

* fbcb79af1398c6ef484c4003a39fa45ca2bb7546 ipatests: fix path in expected error message

PR #2464, opened 2018-10-18
Support interactive prompt for ntp options

Tiboris:
As the FreeIPA server is no longer a NTP service
providing instance its clients and replicas
configuration of time service can not be handled
as it was before change to chrony. Configuration
using master FQDN or autodiscovery for DNS record
would make no difference because every FreeIPA
instance is only chrony client now and does not
update DNS _ntp._udp record.

FreeIPA now asks user for NTP source server
or pool address in interactive mode if there is
no server nor pool specified and autodiscovery
has not found any NTP source in DNS records.

Resolves: https://pagure.io/freeipa/issue/7747

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -2470,18 +2470,21 @@ def sync_time(options, fstore, statestore):
     # disable other time&date services first
     timeconf.force_chrony(statestore)
 
-    logger.info('Synchronizing time')
-
-    if not options.ntp_servers:
+    if not options.ntp_servers and not options.ntp_pool:
         ds = discovery.IPADiscovery()
         ntp_servers = ds.ipadns_search_srv(cli_domain, '_ntp._udp',
                                            None, break_on_first=False)
-    else:
-        ntp_servers = options.ntp_servers
+        if not ntp_servers and not options.unattended:
+            options.ntp_servers, options.ntp_pool = timeconf.get_time_source()
+        else:
+            options.ntp_servers = ntp_servers
+
+    logger.info('Synchronizing time')
 
     configured = False
-    if ntp_servers or options.ntp_pool:
-        configured = timeconf.configure_chrony(ntp_servers, options.ntp_pool,
+    if options.ntp_servers or options.ntp_pool:
+        configured = timeconf.configure_chrony(options.ntp_servers,
+                                               options.ntp_pool,
                                                fstore, statestore)
     else:
         logger.warning("No SRV records of NTP servers found and no NTP server "
--- a/ipaclient/install/timeconf.py
+++ b/ipaclient/install/timeconf.py
@@ -28,6 +28,7 @@ from ipapython import ipautil
 from ipaplatform.tasks import tasks
 from ipaplatform import services
 from ipaplatform.paths import paths
+from ipapython.ipautil import user_input
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


rcritten:
Was this functionality requested somewhere?

Tiboris:
Hi @rcritten,

this functionality was not requested in any upstream ticket but I can create it and link it to this PR.

The purpose of this PR is to be able to configure time server even in interactive mode, and I have some explanation why is this IMHO nice to have.

As the FreeIPA server is no longer a NTP service providing instance its clients and replicas configuration of time service can not be configured easily as it was before. Before change to chrony every new instance could be configured using master FQDN or autodiscovery for DNS record. This record is not created by default and using FQDN would make no difference because every FreeIPA instance is a chrony client now.

Tiboris:
Hi @rcritten,

this functionality was not requested in any upstream ticket but I can create it and link it to this PR.

The purpose of this PR is to be able to configure time server even in interactive mode, and I have some explanation why is this IMHO nice to have.

As the FreeIPA server is no longer a NTP service providing instance its clients and replicas configuration of time service can not be configured easily as it was before. Before change to chrony every new instance could be configured using master FQDN or autodiscovery for DNS record. This record is not created by default and using FQDN would make no difference because every FreeIPA instance is a chrony client now.

rcritten:
Can you:

1. File an RFE ticket and associate it with this PR? This will make the new feature stand out in the release notes
2. Add your justification to the RFE and (perhaps amended) to the commit message?

This is a behavioral change to the installer that is going to be a surprise to anyone that isn't in on this review.

Tiboris:
Sure, @rcritten.

ipaclient/install/timeconf.py
```diff
@@ -39,6 +40,28 @@ def __backup_config(path, fstore=None):
         shutil.copy(path, "%s.ipasave" % (path))
 
 
+def get_time_source(options):
+    """
+    While in interactive installation user has to specify NTP server or pool
+    to be used in chrony configuration. This method asks user input on these
+    values in case that they were not specified before installation start.
+    """
+    if ipautil.user_input("Do you want to configure chrony "
```
tiran: You aren't using i18n marker here. Is there no translation support for the installer?

ipaclient/install/timeconf.py
```diff
@@ -39,6 +40,28 @@ def __backup_config(path, fstore=None):
...
+                          "with NTP server or pool address?", False):
+        servers = user_input("NTP source server addresses", allow_empty=True)
```
tiran: The prompt does neither mention that multiple servers can be separated by ``,`` nor that an empty server list is supported.

ipaclient/install/timeconf.py
```diff
@@ -39,6 +40,28 @@ def __backup_config(path, fstore=None):
...
+        if servers:  # if user input is not '' (empty)
+            logger.debug("User provided NTP server(s):")
+            options.ntp_servers = servers.split(',')
```
tiran: Users tend to separate servers by ``, `` (with space after comma). How about ``[server.strip for server in servers.split(',') if server.strip()]`` ?

ipaclient/install/timeconf.py
Tiboris: I should probably update man page ... or is it acceptable to have these information in prompt?

ipaclient/install/timeconf.py
Tiboris: oh, right

ipaclient/install/timeconf.py
Tiboris: Sure, can do.

Tiboris:
NTPoptions tests are fine:
http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/611b0458-5acd-11e9-958c-fa163e43d042/

ipaclient/install/client.py
```diff
@@ -2470,18 +2470,21 @@ def sync_time(options, fstore, statestore):
...
-
-    if not options.ntp_servers:
+    if not options.ntp_servers and not options.ntp_pool:
         ds = discovery.IPADiscovery()
         ntp_servers = ds.ipadns_search_srv(cli_domain, '_ntp._udp',
                                            None, break_on_first=False)
-    else:
-        ntp_servers = options.ntp_servers
+        if not ntp_servers and not options.unattended:
+            timeconf.get_time_source(options)
```
tiran: I don't like functions that modify an argument as a side effect. It's too magic and hard to understand. How do you like:

```
options.ntp_servers, options.ntp_pool = timeconf.get_time_source()
```

It's explicit and the reader immediately understands what the function is doing.

Tiboris: Understand.

flo-renaud:
master:

* 536e7da6d2f5617a5fa9adb2d359c8af1737604c Support interactive prompt for ntp options

flo-renaud:
@Tiboris 
A manual backport to ipa-4-7 is required, can you handle it please?

Tiboris:
Sure @flo-renaud .

PR #2471, opened 2018-10-23
ipa-replica-install --setup-adtrust: check for package ipa-server-tru…

flo-renaud:
…st-ad

When adding the option --setup-adtrust to ipa-replica-install,
we need to check that the package freeipa-server-trust-ad is
installed.
To avoid relying on OS-specific commands like yum, the check is instead
ensuring that the file /usr/share/ipa/smb.conf.empty is present
(this file is delivered by the package).
When the check is unsuccessful, ipa-replica-install exits with an error
message.

Fixes: https://pagure.io/freeipa/issue/7602


tiran:
rerun, `` RunPytest Failed to publish artifacts ``

flo-renaud:
Rebased to master to pick the annobin fix.

PR #2474, opened 2018-10-23
Update annobin to fix continuous-integration/travis-ci/pr issues

t-woerner:
gcc is updated with the dnf builddep line, but annobin is not. Therefore
configure fails with "C compiler cannot create executables".

This is related to https://bugzilla.redhat.com/show_bug.cgi?id=1574478

See: https://pagure.io/freeipa/issue/7740


t-woerner:
master:

* 5c0885d2667c396844150af2ce696049e6015cb9 Update annobin to fix continuous-integration/travis-ci/pr issues

PR #2485, opened 2018-10-24
Replace hard-coded interpreter with sys.executable

tiran:
Instead of hard-coding python3, the smart card advise script now uses
the current executable path from sys.executable as interpreter.

Fixes: https://pagure.io/freeipa/issue/7741


tiran:
master:

* 4de97f496dd29d7d510759d7fc875ab720d5e90a Replace hard-coded interpreter with sys.executable

PR #2436, opened 2018-10-09
Restore KRA clone installation integration test

frasertweedale:
This Dogtag issue that caused KRA clone installation failure in some
scenarios has been fixed (https://pagure.io/dogtagpki/issue/3055).
This reverts commit 2488813260a407477c7516b33ce4238b69c8dd8d and
bumps the pki-core dependency.

Fixes: https://pagure.io/freeipa/issue/7654

PR #2491, opened 2018-10-25
Require Dogtag 10.6.7 with KRA DL1 fix

tiran:
Dogtag PKI >= 10.6.7 comes with a fix for number range depletion when
multiple clones created from same master.

pki-ca-10.6.7 updates are pending. For the time being, FreeIPA's master
COPR has builds.

See: https://pagure.io/freeipa/issue/7654
See: https://pagure.io/dogtagpki/issue/3055

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -76,9 +76,9 @@
 
 %endif  # Fedora
 
-# Require Dogtag PKI 10.6.6 to detect when fips is available,
-# https://pagure.io/freeipa/issue/7608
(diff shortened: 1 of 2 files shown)


tiran:
Pending updates:

* https://bodhi.fedoraproject.org/updates/FEDORA-2018-fd68715f9c
* https://bodhi.fedoraproject.org/updates/FEDORA-2018-83e180a755

tiran:
Duplicate of #2436

PR #2494, opened 2018-10-26
ipatests: add test for ipa-advise for enabling sudo for admins group

wladich:
Test that
1) sudo is not enabled for members of admins group by default
2) sudo is enabled for them after execution of script provided
by ipa-advise enable_admins_sudo

Related to https://pagure.io/freeipa/issue/7538

### DIFF
--- a/ipatests/test_integration/test_sudo.py
+++ b/ipatests/test_integration/test_sudo.py
@@ -80,6 +80,11 @@ class TestSudo(IntegrationTest):
                                 'defaults',
                                 '--sudooption', "!authenticate"])
 
+        # Create test user -- member of group admins
+        cls.master.run_command(['ipa', 'user-add', 'admin2',
+                                '--first', 'Admin', '--last', 'Second'])
+        cls.master.run_command(['ipa', 'group-add-member', 'admins',
+                                '--users', 'admin2'])
 
     @classmethod
     def uninstall(cls, mh):
@@ -116,6 +121,34 @@ class TestSudo(IntegrationTest):
                                           raiseonerr=False)
 
         return result
+
+    # testcases test_admins_group_does_not_have_sudo_permission and
+    # test_advise_script_enable_sudo_admins must be run before any other sudo
+    # rules are applied
+    def test_admins_group_does_not_have_sudo_permission(self):
+        result = self.list_sudo_commands('admin2', raiseonerr=False)
+        assert result.returncode == 1
+        assert "Sorry, user admin2 may not run sudo on {}.".format(
+            self.clientname) in result.stderr_text
+
+    def test_advise_script_enable_sudo_admins(self):
+        """
+            Test for advise scipt to add sudo permissions for admin users
+            https://pagure.io/freeipa/issue/7538
+        """
+        result = self.master.run_command('ipa-advise enable_admins_sudo')
+        script = result.stdout_text
+        self.master.run_command('bash', stdin_text=script)
+        try:
+            result = self.list_sudo_commands('admin2')
+            assert '(root) ALL' in result.stdout_text
+        finally:
+            result1 = self.master.run_command(
+                ['ipa', 'sudorule-del', 'admins_all'], raiseonerr=False)
+            result2 = self.master.run_command(
+                ['ipa', 'hbacrule-del', 'admins_sudo'], raiseonerr=False)
+            assert result1.returncode == 0 and result2.returncode == 0,\
+                'rules cleanup failed'
 
     def test_nisdomainname(self):
         result = self.client.run_command('nisdomainname')


rcritten:
Just a question but can these classes be merged? It would save another full server installation.

Or do you think it is safer to have its own clear IPA Master to ensure the changes are applied properly?

ipatests/test_integration/test_sudo.py
```diff
@@ -24,22 +24,43 @@
...
-    http://www.freeipa.org/page/V4/Sudo_Integration#Test_Plan
-    """
+class TestSudoBase(IntegrationTest):
     num_clients = 1
     topology = 'line'
 
     @classmethod
     def install(cls, mh):
-        super(TestSudo, cls).install(mh)
+        super(TestSudoBase, cls).install(mh)
```
tiran: Master is Python 3-only. Please use ``super().install(mh)``.

ipatests/test_integration/test_sudo.py
```diff
@@ -24,22 +24,43 @@
...
+        clear_sssd_cache(self.client)
+        sudo_args = '-ll' if verbose else '-l'
+        stdin_text = None
+        if password is None:
+            sudo_args += ' -n'
+        else:
+            sudo_args += ' -S'
+            stdin_text = password + '\n'
+        return self.client.run_command(
+            'su -c "sudo %s" %s' % (sudo_args, user),
```
tiran: Please pass a list of arguments to ``run_command``. IIRC a string will trigger unsafe ``shell=True`` semantic.

ipatests/test_integration/test_sudo.py
```diff
@@ -680,3 +693,45 @@ def test_category_all_validation_runasgroup(self):
                                           '--groups', 'testgroup2'],
                                           raiseonerr=False)
         assert result.returncode != 0
+
+
+class TestSudoAdvise(TestSudoBase):
```
tiran: A new test class means that the test runner also has to install a new server just for these tests. This will prolongate testing by more than 10 minutes.

wladich:
>  You could and should make the test idempotent by removing all sudo rule related rules and settings after the test. You can combine test_install_advise_script and test_admins_have_sudo_permission in one function, wrap the function in a fixture or try/finally and remove all created rules at the end.

I followed style of existing sudo tests, where rule setup  and check are explicitly split to different methods, no cleanup is done and tests rely on each other. On the other hand there are tests like test_dnssec.py::TestMigrateDNSSECMaster where single method contains setup and several scenarios checks. I don't like either of those styles but decided to be consistent with existing code. Or maybe there is  a written style guide for ipatests?

ipatests/test_integration/test_sudo.py
wladich: I explained why I did it  in  https://github.com/freeipa/freeipa/pull/2494#issuecomment-433840963
I have to deal with the fact, that existing tests do not make cleanup. So to avoid creating new class I will have to put tests at the top of testclas  and to ask in comments not to add anything before them, which looks a little bit ugly to me.

tiran:
Yeah, the test style of ipatests is not the best style for testing. Good tests should be idempotent and executable in any random order. I don't like to introduce new integration test classes, because it results in another ``ipa-server-install`` and uninstall. So any subclass of integration test adds another 6 to 8 minutes to our tests without doing any additional testing.

ipatests/test_integration/test_sudo.py
```diff
@@ -24,22 +24,43 @@
     clear_sssd_cache, get_host_ip_with_hostmask, modify_sssd_conf)
 
 
-class TestSudo(IntegrationTest):
-    """
-    Test Sudo
-    http://www.freeipa.org/page/V4/Sudo_Integration#Test_Plan
-    """
+class TestSudoBase(IntegrationTest):
```
tiran: It's a good idea to *not* name a base class for tests as ``TestSomething``. pytest considers anything starting with ``test`` or ``Test`` as a test. For an abstract base class, we don't want pytest to discover the class. How about ``SudoBaseTest`` or ``BaseSudoTest`` instead?

ipatests/test_integration/test_sudo.py
```diff
@@ -24,22 +24,43 @@
...
+        clear_sssd_cache(self.client)
+        sudo_args = '-ll' if verbose else '-l'
+        stdin_text = None
+        if password is None:
+            sudo_args += ' -n'
+        else:
+            sudo_args += ' -S'
+            stdin_text = password + '\n'
+        return self.client.run_command(
+            ['su', '-', user, '-c', 'sudo %s' % sudo_args],
```
tiran: You are still construction a string here. Did you try to pass in a list?

I'm picky here because this kind of code is a common security problem. It's a potential shell injection vector, something like a SQL injection but way worse. This code is not yet vulnerable, because you control all input variables. However the next person may change it inadvertently.

```
args = ['su', '-', user, '-c', 'sudo']
if verbose:
    args.append('-ll')
else:
    args.append('-l')
if password is not None:
    args.append('-S')
    stdin_text = password + '\n'
else:
    args.append('-n')
    stdin_text = None
...
```

ipatests/test_integration/test_sudo.py
tiran: Technically speaking, existing tests perform a cleanup. They use a tactical nuke to tear down the whole cluster with ``ipa-server-install --uninstall`` instead of a surgical scalpel to remove some data. :rofl:

ipatests/test_integration/test_sudo.py
wladich: @tiran 
from `man su`:
`
 -c, --command=command
              Pass command to the shell with the -c option.
`
"command" must be a single parameter, it does not accept additional arguments.
"-ll", "-S", "-n" are arguments for "sudo" bu in your example strings  will be consumed but "su"

wladich:
@tiran 
I moved new tests to existing class and added cleanup of added rules

tiran:
master:

* 324da5c379847fbb92e6eb71433a2aec9898cde9 ipatests: add test for ipa-advise for enabling sudo for admins group

PR #2500, opened 2018-10-29
Check omitted tests

tiran:
## Add missing tests to nighly runs
    
Run test_customized_ds_config_install and test_dns_locations in nightly
runs.

## Speed up test_customized_ds_config_install
    
Reuse master instance when installing replica with custom DS config.
This avoids one extra ipa-server-install and also tests replica
installation from a master with custom DS config.

See: https://pagure.io/freeipa/issue/7743

### DIFF
--- a/ipatests/prci_definitions/nightly_master.yaml
+++ b/ipatests/prci_definitions/nightly_master.yaml
@@ -639,3 +639,27 @@ jobs:
         template: *ci-master-f28
         timeout: 19600
         topology: *ipaserver
+
+  fedora-28/customized_ds_config_install:
+    requires: [fedora-28/build]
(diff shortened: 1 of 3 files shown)


rcritten:
Shouldn't you include a temporary commit so these two tests run during gating for easier review?

flo-renaud:
@tiran 
thanks for the PR, works for me. Please remove temp commit before pushing.

tiran:
master:

* e64ae1d043fb00f04e1c8c77cd214a806ce308b0 Add missing tests to nighly runs
* 4cede86673e33bd2e8c9b6dd5eec2ccdb98c3a1a Speed up test_customized_ds_config_install

PR #2502, opened 2018-10-30
Fix behaviour when the random node is chosen during the replica

shaman007:
installation, not the one that is defined in the --server option


rcritten:
Please expand the commit message to explain better what it is you are trying to fix. It references a replica but, for example, doesn't mention that it is a CA replica.

Add a reference to https://pagure.io/freeipa/issue/7744 per https://www.freeipa.org/page/Contribute/Code#Create_pull_request_on_Github

This needs to work also in the case where the remote master doesn't have a CA installed on it.

PR #2515, opened 2018-11-06
Allow ipaapi user to access SSSD's info pipe

tiran:
For smart card authentication, ipaapi must be able to access to sss-ifp.
During installation and upgrade, the ipaapi user is now added to
[ifp]allowed_uids.

The commit also fixes two related issues:

* The server upgrade code now enables ifp service in sssd.conf. The
  existing code modified sssd.conf but never wrote the changes to disk.
* sssd_enable_service() no longer fails after it has detected an
  unrecognized service.

Fixes: https://pagure.io/freeipa/issue/7751

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -23,7 +23,7 @@ import SSSDConfig
 import ipalib.util
 import ipalib.errors
 from ipaclient.install import timeconf
-from ipaclient.install.client import sssd_enable_service
+from ipaclient.install.client import sssd_enable_ifp
 from ipaplatform import services
 from ipaplatform.tasks import tasks
 from ipapython import ipautil, version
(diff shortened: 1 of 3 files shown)


ipaserver/install/server/upgrade.py
```diff
@@ -2017,15 +2033,8 @@ def upgrade_configuration():
...
-    set_sssd_domain_option('ipa_server_mode', 'True')
-    set_sssd_domain_option('ipa_server', api.env.host)
 
-    sssdconfig = SSSDConfig.SSSDConfig()
-    sssdconfig.import_config()
-    sssd_enable_service(sssdconfig, 'ifp')
-
-    sssd = services.service('sssd', api)
-    sssd.restart()
+    sssd_updgrade()
```
rcritten: Typo?

ipaserver/install/server/upgrade.py
```diff
@@ -1408,6 +1408,22 @@ def set_sssd_domain_option(option, value):
     sssdconfig.write(paths.SSSD_CONF)
 
 
+def sssd_updgrade():
```
rcritten: updgrade?

ipaserver/install/server/upgrade.py
tiran: Yeah, I couldn't decide on either update or upgrade, so I wrote updgrade.

tiran:
master:

* 8b0f3595fd94a31a0e2d5a83dc74a430c973325b Allow ipaapi user to access SSSD's info pipe

PR #2522, opened 2018-11-07
ipaserver.install.adtrust: fix CID 323644

abbra:
Fix Coverity finding CID 323644: logically dead code path

The code to determine whether NetBIOS name was already set or need to be
set after deriving it from a domain or asking a user for an interactive
input, was refactored at some point to avoid retrieving the whole LDAP
entry. Instead, it was provided with the actual NetBIOS name retrieved.

As result, a part of the code got neglected and was never executed.

Fix this code and provide a test that tries to test predefined,
interactively provided and automatically derived NetBIOS name depending
on how the installer is being run.

We mock up the actual execution so that no access to LDAP or Samba is
needed.

Fixes: https://pagure.io/freeipa/issue/7753

### DIFF
--- a/ipatests/test_ipaserver/test_adtrust_mockup.py
+++ b/ipatests/test_ipaserver/test_adtrust_mockup.py
@@ -1,0 +1,54 @@
+# Copyright (C) 2018  FreeIPA Project Contributors - see LICENSE file
+
+from __future__ import print_function
+import ipaserver.install.adtrust as adtr
+from ipaserver.install.adtrust import set_and_check_netbios_name
+from collections import namedtuple
+from unittest import TestCase, mock
+from io import StringIO
+
+
+class ApiMockup:
+    Backend = namedtuple('Backend', 'ldap2')
+    Calls = namedtuple('Callbacks', 'retrieve_netbios_name')
+    env = namedtuple('Environment', 'domain')
+
+
+class TestNetbiosName(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        api = ApiMockup()
+        ldap2 = namedtuple('LDAP', 'isconnected')
+        ldap2.isconnected = mock.MagicMock(return_value=True)
+        api.Backend.ldap2 = ldap2
+        api.Calls.retrieve_netbios_name = adtr.retrieve_netbios_name
+        adtr.retrieve_netbios_name = mock.MagicMock(return_value=None)
+        cls.api = api
+
+    @classmethod
+    def tearDownClass(cls):
+        adtr.retrieve_netbios_name = cls.api.Calls.retrieve_netbios_name
+
+    def test_NetbiosName(self):
+        """
+        Test set_and_check_netbios_name() using permutation of two inputs:
+        - predefined and not defined NetBIOS name
+        - unattended and interactive run
+        As result, the function has to return expected NetBIOS name in
+        all cases. For interactive run we override input to force what
+        we expect.
+        """
+        self.api.env.domain = 'example.com'
+        expected_nname = 'EXAMPLE'
+        # NetBIOS name, unattended, should set the name?
+        tests = ((expected_nname, True, False),
+                 (None, True, True),
+                 (None, False, True),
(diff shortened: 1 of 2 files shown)


ipatests/test_ipaserver/test_adtrust_mockup.py
```diff
@@ -0,0 +1,59 @@
...
+from unittest import TestCase
+from unittest.mock import MagicMock
+import sys
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
+
+
+class ApiMockup:
```
tiran: Nice! I love it!

In the long run, I'd like to see this more generalized and moved to a place that can be used by other tests. For now, keep it there.

ipatests/test_ipaserver/test_adtrust_mockup.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (C) 2018  FreeIPA Project Contributors - see LICENSE file
+
+from __future__ import print_function
+import ipaserver.install.adtrust as adtr
+from ipaserver.install.adtrust import set_and_check_netbios_name
+from collections import namedtuple
+from unittest import TestCase
+from unittest.mock import MagicMock
```
tiran: FYI, this only works on Python 3. For Python 2.7, you have to use the external package ``mock``.

ipatests/test_ipaserver/test_adtrust_mockup.py
```diff
@@ -0,0 +1,59 @@
...
+        we expect.
+        """
+        self.api.env.domain = 'example.com'
+        expected_nname = 'EXAMPLE'
+        # NetBIOS name, unattended, should set the name?
+        tests = ((expected_nname, True, False),
+                 (None, True, True),
+                 (None, False, True),
+                 (expected_nname, False, False))
+        sys.stdin = StringIO(expected_nname + '\r')
```
tiran: Since you already started to use the mock module, how about ``mock.patch`` here? You can use it either as a context manager or a function decorator.

```
>>> import sys
>>> import unittest.mock
>>> from io import StringIO
>>> with unittest.mock.patch('sys.stdin', new_callable=StringIO):
...     print(sys.stdin)
... 
<_io.StringIO object at 0x7f438ac23ca8>
>>> print(sys.stdin)
<_io.TextIOWrapper name='<stdin>' mode='r' encoding='UTF-8'>
>>> @unittest.mock.patch('sys.stdin', new_callable=StringIO)
... def example(m_stdin):
...     print(sys.stdin)
... 
>>> example()
<_io.StringIO object at 0x7f438a9e5708>
```

ipatests/test_ipaserver/test_adtrust_mockup.py
```diff
@@ -0,0 +1,54 @@
...
+        we expect.
+        """
+        self.api.env.domain = 'example.com'
+        expected_nname = 'EXAMPLE'
+        # NetBIOS name, unattended, should set the name?
+        tests = ((expected_nname, True, False),
+                 (None, True, True),
+                 (None, False, True),
+                 (expected_nname, False, False))
+        with mock.patch('sys.stdin', new_callable=StringIO) as input:
```
tiran: ``input`` is the name of a builtin function. I'm puzzled that none of our linters are complaining... How about ``stdin``? Or don't bind the name at all and just use ``sys.stdin``.

abbra:
I renamed `input` to `stdin`. Since we are not importing `sys` at all, this should be safe. I'd prefer to avoid referencing `sys.stdin` directly.

tiran:
PS: I have removed ``ipa-next`` label. It's a marker for *just land the changes in master*.

tiran:
re-run, ``kadmin.service`` failed to start.

tiran:
Another ``kadmin.service`` startup failure. ``rpcbind`` is really hating this PR.

tiran:
master:

* 82af034023b03ae64f005c8160b9e961e7b9fd55 ipaserver.install.adtrust: fix CID 323644

PR #2523, opened 2018-11-07
Fix test_cli_fsencoding on Python 3.7

tiran:
Starting with Python 3.7, PEP 538 addresses the locale issue. Python now
supports UTF-8 file system encoding with non-UTF-8 C locale.

See: https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-pep538
See: https://pagure.io/freeipa/issue/5887

### DIFF
--- a/ipatests/test_cmdline/test_cli.py
+++ b/ipatests/test_cmdline/test_cli.py
@@ -340,8 +340,14 @@ def test_cli_fsencoding():
         env=env,
     )
     out, err = p.communicate()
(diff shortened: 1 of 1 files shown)


abbra:
LGTM. This should fix a particular problem with Python 3.7.1 on Fedora 29.

tiran:
master:

* 0a5a7bdef7c300cb8f8a8128ce6cf5b115683cbe Fix test_cli_fsencoding on Python 3.7

PR #2530, opened 2018-11-07
Fix ipa user-add --radius=radiusserver

flo-renaud:
## ipa user-add: add optional objectclass for radius-username
    
The command `ipa user-add --radius-username` fails with `ipa: ERROR: attribute "ipatokenRadiusUserName" not allowed` because it does not add the objectclass ipatokenradiusproxyuser that is required by the attribute ipatokenradiususername.

The issue happens with `ipa user-add / stageuser-add / user-mod / stageuser-mod`.
    
The fix adds the objectclass when needed in the pre_common_callback method of baseuser_add and baseuser_mod (ensuring that user and stageuser commands are fixed).

Also add a xmlrpc unit test for `ipa user-add|mod --radius-username`

Fixes https://pagure.io/freeipa/issue/7569

## radiusproxy: add permission for reading radius proxy servers
    
 A non-admin user which has the `User Administrator` role cannot add a user with `ipa user-add --radius=<proxy>` because the call needs to read the radius proxy server entries.

The fix adds a System permission for reading radius proxy server entries (all attributes except the ipatokenradiussecret). This permission is added to the already existing privileges `User Administrators` and `Stage User Administrators`, so that the role `User Administrator` can call `ipa [stage]user-add|mod --radius=<proxy>`

Also add an integration test for the following scenario:
- create a user with the "User Administrator" role
- as this user, create a user with a --radius=<radius_proxy_server>

Fixes: https://pagure.io/freeipa/issue/7570


ipaserver/plugins/baseuser.py
```diff
@@ -573,8 +576,11 @@ def check_userpassword(self, entry_attrs, **options):
             setattr(context, 'randompassword', entry_attrs['userpassword'])
 
     def check_objectclass(self, ldap, dn, entry_attrs):
-        if ('ipasshpubkey' in entry_attrs or 'ipauserauthtype' in entry_attrs
-            or 'userclass' in entry_attrs or 'ipatokenradiusconfiglink' in entry_attrs):
+        if 'ipasshpubkey' in entry_attrs or \
```
tiran: The check can be simplified with a ``set()`` operation:

```
if {'ipasshpubkey', 'ipauserauthtype', 'userclass', 'ipatokenradiusconfiglink', 'ipatokenradiususername'}.intersection(entry_attrs):
```

It's easier to read and faster. Dict's keys views are set-like objects.

flo-renaud: @tiran 
Thanks, I updated the PR with your suggestion.

tiran:
Both patches looks good to me. Thanks for the additional tests, too!

@rcritten @abbra Please take a look, too.

fcami:
Reviewed changes to:
ACI.txt
radiusproxy.py
test_user_permissions.py
baseuser.py
all look good to me.

ipaserver/plugins/baseuser.py
```diff
@@ -602,6 +607,11 @@ def check_objectclass(self, ldap, dn, entry_attrs):
                     answer = self.api.Object['radiusproxy'].get_dn_if_exists(cl)
                     entry_attrs['ipatokenradiusconfiglink'] = answer
 
+            if 'ipatokenradiususername' in entry_attrs:
+                if 'ipatokenradiusproxyuser' not in obj_classes:
+                    entry_attrs['objectclass'].append(
+                        'ipatokenradiusproxyuser')
+
```
rcritten: Isn't this functionally equivalent to lines 488-490, calling add_missing_object_class()?

tiran: Yes, it is. IMHO both places should look like this:

```
if 'ipatokenradiususername' in entry_attrs:
    add_missing_object_class(ldap, u'ipatokenradiusproxyuser', dn,
                             entry_attrs, update=False)
```

flo-renaud: Lines 488-490 are called for user-add, while these lines are called for user-mod.
In user-add (new code), I chose to call the existing function add_missing_object_class, but here I kept the same style as the surrounding lines. The rationale was that obj_classes was already fetched and lowercase'd in line 590, so it looked more efficient. But I can use add_missing_object_class here, too, if you think it's better.

abbra:
I think we can actually go with @flo-renaud variant. Indeed, it would be nice to unify the flows but given we already have the data fetched, not using it would be foolish. May be there could be a comment in the code pointing out that we already have the data fetched?

flo-renaud:
Hi @abbra 
I added a comment as you suggested.

abbra:
ACK, apart from the temp commit. Can it be merged into some other test task? We definitely need to run this test but not as a separate one.

flo-renaud:
The test_user_permissions.py will be run in the nightly thanks to commit [60a31d3](https://github.com/freeipa/freeipa/commit/60a31d3f0ebca54ae21edaeb86b8480ffafe6e23)

flo-renaud:
Temp commit removed

tiran:
master:

* 19cd9603876845ea068b7fd0778e2f0d5ac4fc70 ipa user-add: add optional objectclass for radius-username
* 1c2c2ee6f815babc0ac59040e03e3075d7319e59 tests: add xmlrpc test for ipa user-add --radius-username
* 5d603fce5d87a39c0a12bbed880a286b00128f34 radiusproxy: add permission for reading radius proxy servers
* da4c12c3e6ac978afc1a365c3aed87eae5832a96 ipatests: add integration test for "Read radius servers" perm

PR #2531, opened 2018-11-07
Replace messagebus with modern name dbus

tiran:
"messagebus" is an old, archaic name for dbus. Upstream dbus has started
to move away from the old name. Let's use the modern term in FreeIPA,
too.

Fixes: https://pagure.io/freeipa/issue/7754


tjaalton:
+1 with my Debian hat on

tiran:
RHEL 7.6 and Fedora 28+ also have ``dbus.service``.

flo-renaud:
Hi @tiran 
thanks for the patch. Tested on fedora 28 with no problem.

tiran:
master:

* 4124743862cacd4a9b1c6569e19888302a7736f4 Replace messagebus with modern name dbus

PR #2532, opened 2018-11-07
Fix test_cli_fsencoding on Python 3.7, take 2

tiran:
0a5a7bdef7c300cb8f8a8128ce6cf5b115683cbe introduced another problem. The
test is now failing on systems without a full IPA client or server
installation. Use IPA_CONFDIR env var to override location of
default.conf, so that the command always fails.


tiran:
master:

* e569afb04eddc7eb0b287d9046de05ed8de4772d Fix test_cli_fsencoding on Python 3.7, take 2

PR #2535, opened 2018-11-08
Copy-paste error in permssions plugin, CID 323649

tiran:
Address a bug in the code block for attributeLevelRights for old clients.
The backward compatibility code for deprecated options was not triggered,
because the new name was checked against wrong dict.

Coverity Scan issue 323649, Copy-paste error

   The copied code will not have its intended effect.
   In postprocess_result: A copied piece of code is inconsistent with the
   original (CWE-398)

See: Fixes: https://pagure.io/freeipa/issue/7753

### DIFF
--- a/ipaserver/plugins/permission.py
+++ b/ipaserver/plugins/permission.py
@@ -486,7 +486,7 @@ class permission(baseldap.LDAPObject):
 
             if old_client:
                 for old_name, new_name in _DEPRECATED_OPTION_ALIASES.items():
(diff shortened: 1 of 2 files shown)


abbra:
LGTM. I did look into the same report yesterday and was convinced this is the right fix.

tiran:
master:

* 816783a1b9790dedb360e76883c295bc142e980c Copy-paste error in permssions plugin, CID 323649

PR #2538, opened 2018-11-08
Split Web UI test suite in nightly PR CI configuration

serg-cymbaluk:
Ticket: https://pagure.io/freeipa/issue/7756

### DIFF
--- a/ipatests/prci_definitions/nightly_master.yaml
+++ b/ipatests/prci_definitions/nightly_master.yaml
@@ -640,16 +640,139 @@ jobs:
         timeout: 3600
         topology: *master_1repl_1client
 
-  fedora-28/test_webui:
+  fedora-28/test_webui_cert:
     requires: [fedora-28/build]
     priority: 50
     job:
       class: RunWebuiTests
       args:
         build_url: '{fedora-28/build_url}'
-        test_suite: test_webui/
+        test_suite: test_webui/test_cert.py
         template: *ci-master-f28
-        timeout: 19600
+        timeout: 1800
+        topology: *ipaserver
+
+  fedora-28/test_webui_general:
+    requires: [fedora-28/build]
+    priority: 50
+    job:
+      class: RunWebuiTests
+      args:
+        build_url: '{fedora-28/build_url}'
+        test_suite: >
+          test_webui/test_loginscreen.py
+          test_webui/test_misc_cases.py
+          test_webui/test_navigation.py
+          test_webui/test_translation.py
+        template: *ci-master-f28
+        timeout: 3600
+        topology: *ipaserver
+
+  fedora-28/test_webui_hosts:
+    requires: [fedora-28/build]
+    priority: 50
+    job:
+      class: RunWebuiTests
+      args:
+        build_url: '{fedora-28/build_url}'
+        test_suite: >
+          test_webui/test_host.py
+          test_webui/test_hostgroup.py
+          test_webui/test_netgroup.py
+          test_webui/test_service.py
+        template: *ci-master-f28
+        timeout: 3600
+        topology: *ipaserver
+
+  fedora-28/test_webui_identity:
+    requires: [fedora-28/build]
+    priority: 50
+    job:
+      class: RunWebuiTests
+      args:
+        build_url: '{fedora-28/build_url}'
+        test_suite: >
+          test_webui/test_automember.py
+          test_webui/test_idviews.py
+        template: *ci-master-f28
+        timeout: 3600
+        topology: *ipaserver
+
+  fedora-28/test_webui_network:
+    requires: [fedora-28/build]
+    priority: 50
+    job:
+      class: RunWebuiTests
+      args:
+        build_url: '{fedora-28/build_url}'
+        test_suite: >
+          test_webui/test_automount.py
+          test_webui/test_dns.py
+          test_webui/test_vault.py
+        template: *ci-master-f28
+        timeout: 3600
+        topology: *ipaserver
+
+  fedora-28/test_webui_policy:
+    requires: [fedora-28/build]
+    priority: 50
+    job:
+      class: RunWebuiTests
+      args:
+        build_url: '{fedora-28/build_url}'
+        test_suite: >
+          test_webui/test_hbac.py
+          test_webui/test_krbtpolicy.py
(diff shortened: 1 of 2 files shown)


tiran:
I'm not a particular fan of another layer of indirection. The additional files in ``suites`` subdirectory make it harder to understand what is going on. You can use YAML's multiline syntax to make the lines easier on the eye. Both grouping by file name or by topic is fine, though. In Travis CI, we simply cluster tests by file name. If related topics is easier for you, then go for it.

## Example
```
  fedora-28/test_webui_ah:
    requires: [fedora-28/build]
    priority: 50
    job:
      class: RunWebuiTests
      args:
        test_suite: >
          test_webui/test_automember.py
          test_webui/test_automount.py
          test_webui/test_cert.py
          test_webui/test_config.py
          test_webui/test_delegation.py
          test_webui/test_dns.py
          test_webui/test_group.py
          test_webui/test_hbac.py
          test_webui/test_host.py
          test_webui/test_hostgroup.py
        template: *ci-master-f28
        timeout: 3600
        topology: *ipaserver
```

tiran:
@f-trivino Please take a look, too. Is it sensible to split the webui tests into so many jobs or would it be better for scaling to have less but larger jobs in PR-CI?

tiran:
By the way, would it make sense to decrease memory limits again? da70f397cbcc7fd464f0f165881e7f223f599d6b has double memory. I guess the system doesn't need that much memory when tests are split across multiple jobs.

f-trivino:
Regarding the memory limits, it would be nice to know the reason because we need to increase the memory. If the reason is that we have included a new single test which mem reqs are higher, then it will not help to decrease the limit again. On the other hand, if the reason is due to the number of tests being executed, then we are leaking memory...

f-trivino:
> @f-trivino Please take a look, too. Is it sensible to split the webui tests into so many jobs or would it be better for scaling to have less but larger jobs in PR-CI?

It depends on the topology. Test_webui runs on ipaserver topology: 2 cpus and 4800 mem, this means that we have about 35 instances where test_webui can be executed ... So, I think for this case it makes sense. A different story would be that test_webui can only be executed in a few instances...

ipatests/prci_definitions/nightly_master.yaml
```diff
@@ -628,14 +628,137 @@ jobs:
...
     priority: 50
     job:
       class: RunWebuiTests
       args:
         build_url: '{fedora-28/build_url}'
-        test_suite: test_webui/
+        test_suite: test_webui/test_cert.py
         template: *ci-master-f28
-        timeout: 19600
+        timeout: 1800
```
f-trivino: I guess we can adjust/verify timeouts by looking into the last NR run of the test_webui execution logs.

f-trivino:
> By the way, would it make sense to decrease memory limits again? [da70f39](https://github.com/freeipa/freeipa/commit/da70f397cbcc7fd464f0f165881e7f223f599d6b) has double memory. I guess the system doesn't need that much memory when tests are split across multiple jobs.

Just a minor comment: tons of firefox processes with Sl status (waiting for an event) are eating the memory, I think the proper cleanup is not done when a test (from test_webui suite) is finished, so we run out of memory. Duplicating the mem makes the job to don't fail but it doesn't solve the issue.

pvoborni:
I support splitting into multiple jobs as the WebUI tests take too much time in comparison to other tests. This is especially annoying if a re-run is required.

As for changing the memory. Change in the topology section of the the pr-ci.yaml  will have no effect on the vm, i.e. the vm will not get more memory. Memory and CPU in the pr-ci.yaml serves as an information for the PR-CI runners to know how beefy the runner must be to handle the task. But the real memory definition is in Vagrantfile template which matches the topology name, e.g.: https://github.com/freeipa/freeipa-pr-ci/blob/master/templates/vagrantfiles/Vagrantfile.ipaserver So only a change there will do the effect, e.g.: https://github.com/freeipa/freeipa-pr-ci/pull/165 . There is an attempt to generate these Vagrantfiles dynamically and thus also support the memory and cpu change in the pr-ci.yaml, but this is not ready yet and this PR should not wait for its completion. https://github.com/freeipa/freeipa-pr-ci/pull/226

f-trivino:
> I support splitting into multiple jobs as the WebUI tests take too much time in comparison to other tests. This is especially annoying if a re-run is required.
> 
> As for changing the memory. Change in the topology section of the the pr-ci.yaml will have no effect on the vm, i.e. the vm will not get more memory. Memory and CPU in the pr-ci.yaml serves as an information for the PR-CI runners to know how beefy the runner must be to handle the task. But the real memory definition is in Vagrantfile template which matches the topology name, e.g.: https://github.com/freeipa/freeipa-pr-ci/blob/master/templates/vagrantfiles/Vagrantfile.ipaserver So only a change there will do the effect, e.g.: [freeipa/freeipa-pr-ci#165](https://github.com/freeipa/freeipa-pr-ci/pull/165) . There is an attempt to generate these Vagrantfiles dynamically and thus also support the memory and cpu change in the pr-ci.yaml, but this is not ready yet and this PR should not wait for its completion. [freeipa/freeipa-pr-ci#226](https://github.com/freeipa/freeipa-pr-ci/pull/226)

Thanks Petr for clarifying. In my opinion we should merge this commit and see if any of the new WebUI jobs requires more memory.

ipatests/prci_definitions/nightly_master.yaml
serg-cymbaluk: test_cert ~ 400s
other tests ~ 1000-2000s
So we can try with these timeouts (1800 and 3600). But for some tests they can be increased if necessary.

serg-cymbaluk:
PR is rebased.

@f-trivino please put "ack" if no changes are required.

f-trivino:
ack

f-trivino:
@serg-cymbaluk  could we maybe change back the mem reqs from limits again? da70f39 it doesn't take any effect.

f-trivino:
ack

serg-cymbaluk:
@f-trivino thanks

serg-cymbaluk:
master:

* 87474cc1a3a2782088a78f56c9267109f1fd5b8c Split Web UI test suite in nightly PR CI configuration

PR #2540, opened 2018-11-08
Nightly test definition: remove redundant and add missing tests

flo-renaud:
## Nightly tests: remove tests already executed in gating
Remove tests from nightly_master and nightly_rawhide as they are already executed in gating:
- test_caless.py::TestServerReplicaCALessToCAFull
- test_replica_promotion.py::TestSubCAkeyReplication

## Nightly tests: add test_user_permissions.py
Run the above test in the nightly test suites
    
Related to https://pagure.io/freeipa/issue/7743

### DIFF
--- a/ipatests/prci_definitions/nightly_master.yaml
+++ b/ipatests/prci_definitions/nightly_master.yaml
@@ -628,6 +628,18 @@ jobs:
         timeout: 7200
         topology: *master_1repl_1client
 
+  fedora-28/test_user_permissions:
+    requires: [fedora-28/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-28/build_url}'
+        test_suite: test_integration/test_user_permissions.py
+        template: *ci-master-f28
+        timeout: 3600
+        topology: *master_1repl_1client
+
   fedora-28/test_webui:
     requires: [fedora-28/build]
     priority: 50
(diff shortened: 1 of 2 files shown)


tiran:
I'd rather go the other way and add all missing gating tests to nightly tests. Gating tests are only executed on Fedora 28. Eventually nightly tests should be executed on both Fedora 28 and Fedora 29.

flo-renaud:
Hi @tiran 
- the second commit is adding test in nightly, so I guess your comment does not apply to this one.
- the first commit is removing duplicate tests from nightlies (they are currently executing both in gating and in nightly). The choice was made earlier in the project to run them on each PR to ensure critical functionalities were tested for each commit. Do you mean that you want to re-consider this decision and execute them only in the nightly tests? Or did I get you wrong?

tiran:
Hi @flo-renaud 

1. correct, I'm all in favor of adding *Nightly tests: add test_user_permissions.py*.
2. Simply speaking, I would prefer if gating tests are a subset of nightly tests instead of a disjunct set. All gating tests should also be executed during nightly runs.

Under ideal conditions we would run all tests for every PR. However due to time and resource constraints, we cannot run all tests to gate every PR. Therefore gating tests are limited to a handful of most common and critical scenarios.

The gating tests are also limited to one version of Fedora, e.g. F28 for now. Now that Fedora 29 has been released, we should also execute all tests on F29. If we make gating and nightly disjunct, then we have no means to run the base set of common, critical scenarios on F29.

tiran:
Thanks @flo-renaud LGTM

tiran:
master:

* 60a31d3f0ebca54ae21edaeb86b8480ffafe6e23 Nightly tests: add test_user_permissions.py

PR #2542, opened 2018-11-09
Enable firewall in the tests

t-woerner:
Currently the firewall is not enabled in the tests. The goal is to enable the file firewall using firewalld in the tests.

All uses of iptables and ip6tables commands in the tests need to be replaced.

PR #2545, opened 2018-11-09
ipatests: update tests for ipa-server-certinstall

flo-renaud:
1. The test test_http_intermediate_ca was expecting success when it should expect a failure. Scenario:
- install IPA ca-less with certs signed by rootCA
- call ipa-server-certinstall with a cert signed by a subCA to replace http cert.
In this case, the command should refuse changing the cert (otherwise the clients won't be able any more to use ipa * commands as the subca is not installed in /etc/ipa/nssdb or in /etc/ipa/ca.crt).

The commit fixes the test expectation and marks the test as xfail (see ticket 7759).
    
2. The test test_ds_intermediate_ca was expecting success when it should expect a failure. Same scenario as above, but for the ldap server cert.

The commit fixes the test expectation and removes the xfail (ticket 6959 was closed as invalid).

Note:
The behavior differs for ldap and http cert because LDAP server is using a NSSDB and http server is using openssl, hence ipa-server-certinstall follows 2 different code paths when changing the server cert.
    
Related to https://pagure.io/freeipa/issue/7759
Related to https://pagure.io/freeipa/issue/6959

### DIFF
--- a/ipatests/prci_definitions/gating.yaml
+++ b/ipatests/prci_definitions/gating.yaml
@@ -242,3 +242,15 @@ jobs:
         template: *ci-master-f28
         timeout: 3600
         topology: *master_1repl
+
+  fedora-28/test_caless_TestCertInstall:
+    requires: [fedora-28/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-28/build_url}'
+        test_suite: test_integration/test_caless.py::TestCertInstall
+        template: *ci-master-f28
+        timeout: 5400
+        topology: *master_1repl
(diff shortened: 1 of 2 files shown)


flo-renaud:
Hi @tiran 
I don't understand your comment. `assert_error` is not logging, it's checking that the command failed and that stderr contains the text provided as argument. Can you clarify?

tiran:
err, wrong window. I meant to add the comment to https://github.com/freeipa/freeipa/pull/2542 Sorry for the confusion.

abbra:
ACK. Looks good. Temp commit removal is needed to proceed.

tiran:
master:

* 8b7e17aa16fa20b2917ee6bb2343b2658306b548 ipatests: update tests for ipa-server-certinstall
* 5483f9f6bb268f42b70eef227e268f8e28922f01 temp commit: run test_integration/test_caless.py::TestCertInstall

PR #2548, opened 2018-11-12
certdb: ensure non-empty Subject Key Identifier

frasertweedale:
Installation or IPA CA renewal with externally-signed CA accepts an
IPA CA certificate with empty Subject Key Identifier. This is
technically legal in X.509, but is an operational issue.
Furthermore, due to an extant bug in Dogtag
(https://pagure.io/dogtagpki/issue/3079) it will cause Dogtag
startup failure.

Reject CA certificates with empty Subject Key Identifier.

Fixes: https://pagure.io/freeipa/issue/7762

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -924,10 +924,13 @@ class NSSDatabase:
             raise ValueError("not a CA certificate")
 
         try:
-            cert.extensions.get_extension_for_class(
+            ski = cert.extensions.get_extension_for_class(
                     cryptography.x509.SubjectKeyIdentifier)
         except cryptography.x509.ExtensionNotFound:
             raise ValueError("missing subject key identifier extension")
+        else:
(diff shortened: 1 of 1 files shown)


fcami:
"Expected behavior: Certificate with empty or absent Subject Key Identifier should be rejected."

The test:

tiran:
master:

* d731f6fd74b76e7d5e2b86a74ebdcca8dcb05062 certdb: ensure non-empty Subject Key Identifier

frasertweedale:
@tiran oh and there's also https://pagure.io/freeipa/issue/7761 which I'm currently working on.
(Trying to work out why `certutil -V` succeeds for 1024-bit key; I would prefer not to to hardcode the check but rely on NSS instead, i.e. so the check reflects whatever the system crypto policy is).

PR #2550, opened 2018-11-12
ipatests: fix CA less expectations

flo-renaud:
The test TestServerInstall::test_ca_2_certs has a wrong expectation. Scenario:
install a CA-less master with `ipa-server-install --ca-cert-file root.pem` where root.pem contains the CA that signed the http and ldap certificates + an additional (unneeded) CA cert.

The test was expecting a failure, but this scenario is not problematic as long as the unneeded CA cert is not added.

Related to https://pagure.io/freeipa/issue/6289 which has been closed as won't fix

### DIFF
--- a/ipatests/test_integration/test_caless.py
+++ b/ipatests/test_integration/test_caless.py
@@ -446,7 +446,6 @@ class TestServerInstall(CALessBase):
                      'The full certificate chain is not present in '
                      '%s/server.p12' % self.master.config.test_dir)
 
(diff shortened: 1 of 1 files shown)


frasertweedale:
ACK (sans temp commit).

flo-renaud:
@frasertweedale 

tiran:
master:

* 1e6a77a6da7ee2b65477b8e843f64d4f5a5ff8e0 ipatests: fix CA less expectations

PR #2561, opened 2018-11-13
Update translations from Zanata

abbra:
I pushed update to Zanata and pulled some changes from it for master. IPA-4-7 will come as a separate pull request.


tiran:
master:

* ee52ceb98cdc8746e8b0a9b834afad536b282d85 Update translations from Zanata

PR #2565, opened 2018-11-14
freeipa.spec.in: add BuildRequires for python3-lib389

flo-renaud:
freeipa.spec.in is missing BuildRequires for python3-lib389. The
consequence is that make fasttest is failing.

Fixes https://pagure.io/freeipa/issue/7767


flo-renaud:
@tiran 
thanks for the suggestion, PR updated.

PR #2574, opened 2018-11-16
ipatests: fix test_replica_uninstall_deletes_ruvs

flo-renaud:
test_topology.py is failing because of a wrong scenario.
Currently, test_replica_uninstall_deletes_ruvs does:
- install master + replica with CA
- `ipa-replica-manage list-ruv` to check that the repl is propery setup
- `ipa-replica-manage del $replica`
- (on replica) `ipa-server-install --uninstall -U`
- `ipa-replica-manage list-ruv` to check that replica does not appear any more in the RUV list

When ipa-replica-manage del is run, the topology plugin creates 2 tasks cleanallruvs (one for the domain, one for the ca) and they are run asynchronously. This means that the ruvs may still be present when the test moves forward and calls list-ruv.

The test should wait for the cleanallruvs tasks to finish before checking that list-ruv does not display replica anymore.

Fixes https://pagure.io/freeipa/issue/7545

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -1145,6 +1145,40 @@ def wait_for_replication(ldap, timeout=30):
         logger.error('Giving up wait for replication to finish')
 
 
+def wait_for_cleanallruv_tasks(ldap, timeout=30):
+    """Wait until cleanallruv tasks are finished
+    """
+    logger.debug('Waiting for cleanallruv tasks to finish')
+    success_status = 'Successfully cleaned rid'
+    for i in range(timeout):
+        status_attr = 'nstaskstatus'
+        try:
+            entries = ldap.get_entries(
+                DN(('cn', 'cleanallruv'), ('cn', 'tasks'), ('cn', 'config')),
+                scope=ldap.SCOPE_ONELEVEL,
+                attrs_list=[status_attr])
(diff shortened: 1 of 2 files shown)


ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -1145,6 +1145,36 @@ def wait_for_replication(ldap, timeout=30):
...
+            for e in entries
+        ):
+            logger.debug("All cleanallruv tasks finished successfully")
+            break
+        else:
+            logger.debug("cleanallruv task in progress, (waited %s/%ss)",
+                         i, timeout)
+        time.sleep(1)
+    else:
+        logger.error('Giving up waiting for cleanallruv to finish')
```
tbordaz: The patch looks good to me. 
I think if it gives up waiting, it would be interesting (for diagnose) to log the last task status of the still running tasks

flo-renaud: Hi @tbordaz 
thnaks, that's a nice suggestion. PR updated.

rcritten:
LGTM

flo-renaud:
Removed temp commit.

flo-renaud:
master:

* 55c0a93271cd7ea6111ac188a5c43615bae4d010 ipatests: fix test_replica_uninstall_deletes_ruvs

PR #2575, opened 2018-11-16
ipa-kdb: reduce LDAP operations timeout to 30 seconds

abbra:
Since LDAP operations used by ipa-kdb driver are synchronous, the
timeout specified here is blocking entire KDC. It is worth reducing the
timeout and since AS REQ processing timeout in KDC is 1 minute, reducing
the timeout for LDAP operations down to 30 seconds allows KDC to
respond promptly in worst case scenario as well.

Fixes: https://pagure.io/freeipa/issue/7217

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb_common.c
+++ b/daemons/ipa-kdb/ipa_kdb_common.c
@@ -23,7 +23,7 @@
 #include "ipa_kdb.h"
 #include <unicase.h>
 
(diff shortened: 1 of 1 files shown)


rcritten:
master:

* 122f968ceebbe80a47eca842a98bb39876264535 ipa-kdb: reduce LDAP operations timeout to 30 seconds

PR #2580, opened 2018-11-17
Issue #7744: Use api.env.server for config.ca_host_name

jaredledvina:
Bug: https://pagure.io/freeipa/issue/7744

This is a really quick attempt to fix this bug. Currently, the CA server is always chosen from LDAP. As a result, when installing the CA but, passing in `--server`, the replica install will use the server passed in but the CA install can replicate from another. This leads to lots of confusion when reviewing the replication topology and can lead to replicating the entire CA domain from a high latency link. 

This really bites us in production where new replicas will replicate initially from a server on the other side of the world and that link has high latency causing failures midway through the install. 

I think this change will work and am happy to test it out. I'd love to see this back ported to 4.5.4 but, understand if that's not possible.

### DIFF
--- a/install/tools/ipa-ca-install.in
+++ b/install/tools/ipa-ca-install.in
@@ -165,7 +165,7 @@ def install_replica(safe_options, options):
     api.Backend.ldap2.connect()
 
     config = ReplicaConfig()
-    config.ca_host_name = None
+    config.ca_host_name = api.env.server
     config.realm_name = api.env.realm
     config.host_name = api.env.host
     config.domain_name = api.env.domain


rcritten:
Thanks for the patch, and sorry for the delay. I'm looking at a related PR, https://github.com/freeipa/freeipa/pull/2144, to see if it will do this in a more comprehensive way.

rcritten:
Ok this is not solved by the other PR. It passes along the same host values as before.

tiran:
Yes, Rob is right. PR #2144 does not address the issue. It's a cleanup and refactorization of the detection code.

I think your proposed change is good, but I need to do some testing first. Could you please rebase your branch on latest master? I'll kick off the large test suite afterwards.

jaredledvina:
@tiran - Sounds good to me, rebased and pushed!

abbra:
I kicked off full PRCI run.

abbra:
CA-less test failed: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/840b468a-fd35-11e8-8660-fa163e0c3a31/report.html (remove report.html from the URI if you want logs)

jaredledvina:
hmmmmm, interesting, I'll take a look and see if I can figure out where the issue is. If you or anyone sees the issue, feel free to let me know.

jaredledvina:
Hey @abbra  - So I tried reviewing the test output and logs but, I can't seem to make heads or tails of how this failed. I've actually never done the CA less to CA full process so, I'm not familiar with that either. `404 Client Error: Not Found for url:` is strange to me. I also can't seem to find where `ipa-ca-install` would have stored it's debug logs. I thought they'd be in the ipa*-install logs but, the replica and client logs don't have what am looking for. 

Does anything stand out to you here?

flo-renaud:
Hi @jaredledvina 
A fix similar to the one you proposed has been committed (c0fd5e3 replica install: set the same master as preferred source for domain and CA), so I think we can close this PR if you agree that it fixes your issue.

jaredledvina:
Hey @flo-renaud,

Ah, nice! Yep, that's exactly what I was after here. Thanks for linking that, if it works as the commit describes, that's the exact solution I was hoping for. Thanks!

PR #2581, opened 2018-11-19
PR-CI: Restart rpcbind when it blocks kadmin port

tiran:
Every now and then, a PR-CI job fails because rpcbind blocks the kadmin
port 749/UDP and kadmin.service fails to start. When NFS secure port is
configured, rpcbind reserves a random low port.

A new workaround detects a blocked port and restart rpcbind.service.

See: https://pagure.io/freeipa/issue/7769

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -160,15 +160,32 @@ def prepare_host(host):
         host.put_file_contents(env_filename, env_to_script(host.to_env()))
 
 
-def allow_sync_ptr(host):
-    kinit_admin(host)
-    host.run_command(["ipa", "dnsconfig-mod", "--allow-sync-ptr=true"],
-                     raiseonerr=False)
+def rpcbind_kadmin_workaround(host):
+    """Restart rpcbind in case it blocks 749/UDP
+
+    See https://pagure.io/freeipa/issue/7769
+    See https://bugzilla.redhat.com/show_bug.cgi?id=1592883
+    """
+    for _i in range(5):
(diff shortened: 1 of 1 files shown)


ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -160,15 +160,27 @@ def prepare_host(host):
...
-                     raiseonerr=False)
+def rpcbind_kadmin_workaround(host):
+    """Restart rpcbind in case it blocks 749/UDP
+
+    See https://pagure.io/freeipa/issue/7769
+    See https://bugzilla.redhat.com/show_bug.cgi?id=1592883
+    """
+    result = host.run_command(['ss', '-ulnp', 'sport', '=', '749'])
+    if 'rpcbind' in result.stdout_text:
+        host.run_command(['systemctl', 'restart', 'rpcbind.service'])
```
rcritten: Is a restart guaranteed not to rebind to port 749? Should this be in a loop?

tiran: In my experiments, rpcbind always incremented the port number by a small, random value. Nevertheless a loop doesn't hurt.

t-woerner: I think it would be good to use the wait option to make sure the the restart is finished before trying another restart.

t-woerner: But I am not sure that it is available in old systemd versions.

tiran:
master:

* 7434a3299d25d1b96ac13ad33fb9ec96eb3da85e PR-CI: Restart rpcbind when it blocks kadmin port

PR #2584, opened 2018-11-20

tiran:
The PR was originally PR @abbra's PR https://github.com/freeipa/freeipa/pull/2106. PR-CI was broken for that PR. I also squashed some intermediate commits.

## Original PR message

`cacert.p12`.

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
+++ b/daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
@@ -366,7 +366,10 @@ static int ipapwd_pre_add(Slapi_PBlock *pb)
     pwdop->pwdata.target = e;
 
     ret = ipapwd_CheckPolicy(&pwdop->pwdata);
-    if (ret) {
+    /* For accounts created by cn=Directory Manager or a passsync
+     * managers, ignore result of a policy check */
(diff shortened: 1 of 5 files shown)


daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
```diff
@@ -846,11 +849,13 @@ static int ipapwd_pre_mod(Slapi_PBlock *pb)
...
-     * policies have been properly checked already, so we check them only
-     * if no krb keys are available */
+     * policies have been properly checked already. We check them only if no
+     * krb keys are available and raise error if the change is not done by a
+     * cn=Directory Manager or one of passsync managers */
     if (has_krb_keys == 0) {
         ret = ipapwd_CheckPolicy(&pwdop->pwdata);
-        if (ret) {
+        if ((pwdop->pwdata.changetype != IPA_CHANGETYPE_DSMGR) &&
+	    (ret != 0)) {
```
tbordaz: My understanding is that a passwd set by DM/passsync is checked (max life and min lenght). This for krb or non_krb.
Where is it skipped for krb entries ?

abbra:
@tiran did you address https://github.com/freeipa/freeipa/pull/2106#pullrequestreview-145282543 or should I continue looking into fixing that issue?

PR #2585, opened 2018-11-20
ipa-replica-install: password and admin-password options mutually exclusive

flo-renaud:
## ipa-replica-install: make password and admin-password options mutually exclusive
Currently it is possible to run ipa-replica-install in one step, and provide --password and --admin-password simultaneously.
This is confusing as --password is intended for one-time pwd when the ipa-replica-install command is delegated to a user who doesn't know the admin password.

The fix makes --password and --admin-password options mutually exclusive.
    
Fixes https://pagure.io/freeipa/issue/6353

## ipatests: add test for ipa-replica-install options
Add a test checking that --password and --admin-password options are mutually exclusive.

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -771,6 +771,10 @@ def promote_check(installer):
 
     client_fstore = sysrestore.FileStore(paths.IPA_CLIENT_SYSRESTORE)
     if not client_fstore.has_files():
+        # One-step replica installation
+        if options.password and options.admin_password:
+            raise ScriptError("--password and --admin-password options are "
+                              "mutually exclusive")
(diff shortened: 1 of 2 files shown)


tiran:
Thanks for the patch and for the really fast test -- just 20 seconds.

flo-renaud:
Also adding ipa-4-6 label as a backport will be required.
Temp commit removed.

Tiboris:
master:

* 23306a28c9181065ab3f2828ab2df56c152e2c62 ipa-replica-install: password and admin-password options mutually exclusive
* 26e35dca413ecfa68df0b6115bed17ea4421fe31 ipatests: add test for ipa-replica-install options

Tiboris:

PR #2598, opened 2018-11-23
Fix ipa user-find --certificate

flo-renaud:
## ipaldap.py: fix method creating a ldap filter for IPACertificate
    
`ipa user-find --certificate` and `ipa host-find --certificate` fail to return matching entries, because the method transforming the attribute into a LDAP filter does not properly handle IPACertificate objects.
Directory Server logs show a filter with `(usercertificate=ipalib.x509.IPACertificate object at 0x7fc0a5575b90>)`

When the attribute contains a IPACertificate, the method needs to extract the public bytes instead of calling str(value).

Fixes https://pagure.io/freeipa/issue/7770

## ipatests: add xmlrpc test for user|host-find --certificate
    
There were no xmlrpc tests for `ipa user-find --certificate` or `ipa host-find --certificate`. The commit adds tests for these commands.

Related to https://pagure.io/freeipa/issue/7770

### DIFF
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -33,6 +33,7 @@ from urllib.parse import urlparse
 import warnings
 
 from cryptography import x509 as crypto_x509
+from cryptography.hazmat.primitives import serialization
 
 import ldap
 import ldap.sasl
@@ -1289,6 +1290,8 @@ class LDAPClient:
             ]
             return cls.combine_filters(flts, rules)
         elif value is not None:
+            if isinstance(value, crypto_x509.Certificate):
+                value = value.public_bytes(serialization.Encoding.DER)
             if isinstance(value, bytes):
                 value = binascii.hexlify(value).decode('ascii')
                 # value[-2:0] is empty string for the initial '\\'
--- a/ipatests/test_xmlrpc/test_host_plugin.py
+++ b/ipatests/test_xmlrpc/test_host_plugin.py
@@ -251,6 +251,11 @@ class TestCRUD(XMLRPC_test):
                         valid_not_after=fuzzy_date,
                     ))
         host.retrieve()
+        # test host-find with --certificate
+        command = host.make_find_command(
+            fqdn=host.fqdn, usercertificate=host_cert)
+        res = command()['result']
+        assert len(res) == 1
 
     def test_try_rename(self, host):
         host.ensure_exists()
--- a/ipatests/test_xmlrpc/test_user_plugin.py
+++ b/ipatests/test_xmlrpc/test_user_plugin.py
@@ -25,6 +25,7 @@ Test the `ipaserver/plugins/user.py` module.
 """
 
 import pytest
+import base64
 import datetime
 import ldap
 import re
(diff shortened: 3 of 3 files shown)


celestian:
As QE: test LGTM, code LGTM and commit messages are understandable.
Any other review?

fcami:
Issue reproduced locally on master.
Issue is fixed after upgrading instance to this PR (built on f29).
Code LGTM. Only issue I have is with the test as it's not strictly necessary to specify a user id and my understanding is that this specifies the user name instead (?):
```
        command = user.make_find_command(uid=u'%s' % user.name,
                                         usercertificate=user_cert)
```
for instance:
```
# ipa user-find --certificate="$(base64 cert.der)"
--------------
1 user matched
--------------
  User login: test-cert
(...)
```
Please don't merge yet.

ipatests/test_xmlrpc/test_host_plugin.py
```diff
@@ -251,6 +251,11 @@ def test_update_simple(self, host):
                         valid_not_after=fuzzy_date,
                     ))
         host.retrieve()
+        # test host-find with --certificate
+        command = host.make_find_command(
+            fqdn=u'%s' % host.fqdn, usercertificate=host_cert)
```
tiran: Why do you use ``u'%s' % host.fqdn`` instead of simple ``host.fqdn``?

flo-renaud: you are right, no need to use this as host.fqdn is already built with `self.fqdn = u'%s.%s' % (name, self.api.env.domain)` in HostTracker.

flo-renaud:
Hi @fcami 
thanks for the review. The `make_find_command` specifies uid and usercertificate to make sure that the LDAP search filter would return at most one entry: (&(uid=user.name)(usercertificate=...)).
Without this narrow filter, the test would be dependent on previous tests run and the content of potentially existing user entries with the same cert. As `uid` is guaranteed to be unique, the test is sure to return at most one entry.
The user is created with `UserTracker(name=u'user1', ...)` which used name as uid.

flo-renaud:
Rebased as PR #2602 has been pushed.

fcami:
(checking/reviewing)

ipapython/ipaldap.py
```diff
@@ -1289,6 +1291,8 @@ def make_filter_from_attr(
             ]
             return cls.combine_filters(flts, rules)
         elif value is not None:
+            if isinstance(value, IPACertificate):
```
tiran: I just noticed https://pagure.io/freeipa/issue/7775 and remember that this PR uses IPACertificate. Please use cryptography's certificate class here. It's more generic:

```
>>> from ipalib.x509 import IPACertificate
>>> from cryptography.x509 import Certificate
>>> issubclass(IPACertificate, Certificate)
True
```

flo-renaud: Hi @tiran
Thanks for the suggestion, implemented in the updated PR.

rcritten:
master:

* 372c2fc99021c037f00a6819baa44b64900de479 ipaldap.py: fix method creating a ldap filter for IPACertificate
* d2fa2ecb4be1e00a2b5f45ed0129175fc5f4dbb1 ipatests: add xmlrpc test for user|host-find --certificate

rcritten:
@flo-renaud manual backports for ipa-4-6 and ipa-4-7 are needed:

Aplying to ipa-4-7: ipaldap.py: fix method creating a ldap filter for IPACertificate
Cleaning up
Aplying to ipa-4-6: ipaldap.py: fix method creating a ldap filter for IPACertificate
Cleaning up

PR #2600, opened 2018-11-26
Fix crashing WebUI identity tests in Nightly PR

serg-cymbaluk:
Clear default user/host group before deleting it in "Automember" tests.

### DIFF
--- a/ipatests/test_webui/test_automember.py
+++ b/ipatests/test_webui/test_automember.py
@@ -595,6 +595,10 @@ class TestAutomember(UI_driver):
         self.switch_to_facet('memberof_group')
         self.assert_record(pkey)
 
+        # Clear default user group
+        self.navigate_by_menu('identity/automember/amgroup')
+        self.select_combobox('automemberdefaultgroup', '')
+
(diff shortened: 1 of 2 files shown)


ipatests/prci_definitions/gating.yaml
```diff
@@ -3,14 +3,10 @@ topologies:
     name: build
     cpu: 2
     memory: 3800
-  master_1repl: &master_1repl
```
tiran: Tip: We have a special template for temp commits, which is easier to use. You can activate the template with a single command:

``ln -sf ipatests/prci_definitions/temp_commit.yaml .freeipa-pr-ci.yaml``

serg-cymbaluk: Ok, thank you

serg-cymbaluk:
@tiran 
The PR is rebased, checks have passed.

serg-cymbaluk:
master:

* 07c163ca9214b1d6bb2446dd3c93ad04e1382535 Fix "ID views" tests fail after running "Automember" tests

PR #2610, opened 2018-11-28
Fix authselect invocations to work with 1.0.2

AdamWill:
Since authselect 1.0.2, invoking an authselect command sequence
like this:

['authselect', 'sssd', '', '--force']

does not work: authselect barfs on the empty string arg and
errors out. We must only pass a features arg if we actually have
some text to go in it.

This broke uninstallation.

I don't know for sure whether the fix in `restore` is needed -
it may be that `cfg[1]` is never an empty sequence - but it
seems sensible to do it just in case.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/ipaplatform/redhat/authconfig.py
+++ b/ipaplatform/redhat/authconfig.py
@@ -158,15 +158,26 @@ class RedHatAuthSelect(RedHatAuthToolBase):
                 " ".join(args))
 
             profile = 'sssd'
-            features = ''
+            features = []
         else:
-            profile = \
-                statestore.restore_state('authselect', 'profile') or 'sssd'
-            features = \
-                statestore.restore_state('authselect', 'features_list') or ''
+            profile = statestore.restore_state('authselect', 'profile')
+            if not profile:
+                profile = 'sssd'
+            features_state = statestore.restore_state(
+                'authselect', 'features_list'
+            )
             statestore.delete_state('authselect', 'mkhomedir')
+            # only non-empty features, https://pagure.io/freeipa/issue/7776
+            if features_state is not None:
+                features = [
+                    f.strip() for f in features_state.split(' ') if f.strip()
+                ]
+            else:
+                features = []
 
-        cmd = [paths.AUTHSELECT, "select", profile, features, "--force"]
+        cmd = [paths.AUTHSELECT, "select", profile]
+        cmd.extend(features)
+        cmd.append("--force")
         ipautil.run(cmd)
 
     def backup(self, path):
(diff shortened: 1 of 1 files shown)


AdamWill:
@edewata also ran into this in https://bugzilla.redhat.com/show_bug.cgi?id=1645708#c10 .

ipaplatform/redhat/authconfig.py
```diff
@@ -187,9 +190,10 @@ def restore(self, path):
         if cfg:
             profile = cfg[0]
 
-            cmd = [
-                paths.AUTHSELECT, "select", profile,
-                " ".join(cfg[1]), "--force"]
+            cmd = [paths.AUTHSELECT, "select", profile]
+            if cfg[1]:
+                cmd.append(" ".join(cfg[1]))
```
tiran: I think this is also wrong and should ``cmd.extend(cfg[1])``. authselect doesn't complain about the syntax but it behaves differently:

```
# authselect current --raw
sssd with-fingerprint with-silent-lastlog
# authselect select sssd 'with-fingerprint with-silent-lastlog' --force
Backup stored at /var/lib/authselect/backups/2018-11-28-07-54-05.0uFTrd
Profile "sssd" was selected.
The following nsswitch maps are overwritten by the profile:
- passwd
- group
- netgroup
- automount
- services

Make sure that SSSD service is configured and enabled. See SSSD documentation for more information.

# authselect select sssd with-fingerprint with-silent-lastlog --force
Backup stored at /var/lib/authselect/backups/2018-11-28-07-54-29.iNPMzl
Profile "sssd" was selected.
The following nsswitch maps are overwritten by the profile:
- passwd
- group
- netgroup
- automount
- services

Make sure that SSSD service is configured and enabled. See SSSD documentation for more information.
 
- with-fingerprint is selected, make sure fprintd service is configured and enabled

# 
```

Please notice that the second variant prints ``with-fingerprint is selected, make sure fprintd service is configured and enabled`` while the first does not. The second variant is equivalent to ``cmd.extend(cfg[1])``.

tiran:
@AdamWill since the new version of authselect breaks FreeIPA on Fedora 28 and 29, I want to accelerate this fix. I took the liberty to force push the fix:

* I changed ``if cfg[1]: cmd.append(" ".join(cfg[1]))`` to ``cmd.extend(cfg[1])``
* I slightly edited the commit message and included a link to https://pagure.io/freeipa/issue/7776

AdamWill:
Oh, good catch. I just concentrated on staying as similar as possible to the existing code, I didn't notice it was actually doing the wrong thing there. That's really a separate fix, though - it was wrong before as well.

AdamWill:
I was already preparing updates for F28 and F29, note, but I'll revise them to include your adjusted patch.

tiran:
argh, it's even wrong in both cases! Brace yourself for another fix.

AdamWill:
yes. it needs to be fixed *both* ways.

AdamWill:
I guess that works, sure.

tiran:
@AdamWill is not in the list of trusted developers. I added the re-run label to kick off a PR-CI run.

AdamWill:
As Groucho Marx so nearly said, I would never want to work on a project which would have *me* as a trusted developer...:P

abbra:
LGTM and ACK.

tiran:
master:

* 7aceca2da4f4b68b942c796d6152706f12135502 Fix authselect invocations to work with 1.0.2

PR #2611, opened 2018-11-28
[Backport][ipa-4-7] Fix authselect invocations to work with 1.0.2

tiran:
Manual backport of PR #2610.

Since authselect 1.0.2, invoking an authselect command sequence
like this:

['authselect', 'sssd', '', '--force']

does not work: authselect barfs on the empty string arg and
errors out. We must only pass a features arg if we actually have
some text to go in it.

This broke uninstallation.

In all cases, features are now passed as separate arguments instead of one
argument separated by space.

Fixes: https://pagure.io/freeipa/issue/7776
Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/ipaplatform/redhat/authconfig.py
+++ b/ipaplatform/redhat/authconfig.py
@@ -158,15 +158,26 @@ class RedHatAuthSelect(RedHatAuthToolBase):
                 " ".join(args))
 
             profile = 'sssd'
-            features = ''
+            features = []
         else:
-            profile = \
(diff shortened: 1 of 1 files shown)


Tiboris:
Rebased

abbra:
@Tiboris I think the code is OK here, do you have any additional notes (you removed the ack)

tiran:
ipa-4-7:

* 05c5be1b1c5ae63c5547d248d926b3411bff2733 Fix authselect invocations to work with 1.0.2

PR #2618, opened 2018-11-29
ipa upgrade: handle double-encoded certificates

flo-renaud:
## ipa upgrade: handle double-encoded certificates

Issue is linked to the ticket #3477 LDAP upload CA cert sometimes double-encodes the value
In old FreeIPA releases (< 3.2), the upgrade plugin was encoding twice the value of the certificate in `cn=cacert,cn=ipa,cn=etc,$BASEDN`.
The fix for 3477 is only partial as it prevents double-encoding when a new cert is uploaded but does not fix wrong values already present in LDAP.
With this commit, the code first tries to read a der cert. If it fails, it logs a debug message and re-writes the value caCertificate;binary to repair the entry.

Fixes https://pagure.io/freeipa/issue/7775

## ipatests: add upgrade test for double-encoded cacert
 
Create a test for upgrade with the following scenario:
- install master
- write a double-encoded cert in the entry `cn=cacert,,cn=ipa,cn=etc,$basedn` to simulate bug 7775
- call ipa-server-upgrade
- check that the upgrade fixed the value

The upgrade should finish successfully and repair the double-encoded cert.

Related to https://pagure.io/freeipa/issue/7775

### DIFF
--- a/ipaserver/install/plugins/upload_cacrt.py
+++ b/ipaserver/install/plugins/upload_cacrt.py
@@ -115,7 +115,18 @@ class update_upload_cacrt(Updater):
                 entry.single_value['cACertificate;binary'] = ca_cert
                 ldap.add_entry(entry)
             else:
-                if b'' in entry['cACertificate;binary']:
+                force_write = False
+                try:
+                    _cert_bin = entry['cACertificate;binary']
+                except ValueError:
+                    # BZ 1644874
+                    # sometimes the cert is badly stored, twice encoded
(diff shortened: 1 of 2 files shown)


tiran:
master:

* 800f2690f5b513e7e27ad739ccbebf7e420140b5 ipa upgrade: handle double-encoded certificates
* 93e3fc4d8930d205be16c1bc00768f607b6d850f ipatests: add upgrade test for double-encoded cacert

PR #2628, opened 2018-11-30
[WIP] Refactor LDAPClient

tiran:
**WORK IN PROGRESS**

Simplify LDAPClient construction and ensure that LDAP connections are always secure (LDAP + STARTTLS, LDAPI, LDAPS) unless explicitly stated otherwise.

### DIFF
--- a/install/migration/migration.py
+++ b/install/migration/migration.py
@@ -55,7 +55,7 @@ def bind(ldap_uri, base_dn, username, password):
         raise IOError(errno.EIO, 'Cannot get Base DN')
     bind_dn = DN(('uid', username), ('cn', 'users'), ('cn', 'accounts'), base_dn)
     try:
-        conn = ipaldap.LDAPClient(ldap_uri)
+        conn = ipaldap.LDAPClient(ldap_uri, start_tls=True)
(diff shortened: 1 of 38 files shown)


rcritten:
Just a note that moving functions around is exactly that kind of thing that breaks external users of the API. It makes logic sense but is is really needed?

tiran:
@rcritten That makes sense. I have added aliases with deprecation warnings to ``installutils``. The previous function locations still work, but users will be informed that they are using a deprecated name.

rcritten:
Was this replaced by https://github.com/freeipa/freeipa/pull/2697 ?

tiran:
Mostly, PR #2697 doesn't have https://github.com/freeipa/freeipa/pull/2628/commits/d1041ab3d0296825f86112de4befe2312f831cea and https://github.com/freeipa/freeipa/pull/2628/commits/ca7e5c87a55bb25aa2a66dc16293754198cc77f3. The commits add ``nsslapd-require-secure-binds: on`` and strong SSF.

tiran:
PRs #2812 and #2813 replace this PR.

PR #2630, opened 2018-11-30
PKINIT: fix ipa-pkinit-manage enable|disable

flo-renaud:
## PKINIT: fix ipa-pkinit-manage enable|disable

The command ipa-pkinit-manage enable|disable is reporting success even though the PKINIT cert is not re-issued.
The command triggers the request of a new certificate (signed by IPA CA when state=enable, selfsigned when disabled), but as the cert file is still present, certmonger does not create a new request and the existing certificate is kept.
The fix consists in deleting the cert and key file before calling certmonger to request a new cert.

There was also an issue in the is_pkinit_enabled() function: if no tracking request was found for the PKINIT cert, is_pkinit_enabled() was returning True while it should not.

Fixes https://pagure.io/freeipa/issue/7200

## ipatest: add test for ipa-pkinit-manage enable|disable
Add a test for ipa-pkinit-manage with the following scenario:
- install master with option --no-pkinit
- call ipa-pkinit-manage enable
- call ipa-pkinit-manage disable
- call ipa-pkinit-manage enable
    
At each step, check that the PKINIT cert is consistent with the expectations: when pkinit is enabled, the cert is signed by IPA CA and tracked by 'IPA' ca helper, but when pkinit is disabled, the cert is self-signed and tracked by 'SelfSign' CA helper.

Related to https://pagure.io/freeipa/issue/7200

### DIFF
--- a/ipaserver/install/ipa_pkinit_manage.py
+++ b/ipaserver/install/ipa_pkinit_manage.py
@@ -72,6 +72,8 @@ class PKINITManage(AdminTool):
                 if ca_enabled:
                     logger.warning(
                         "Failed to stop tracking certificates: %s", e)
+            # remove the cert and key
+            krb.delete_pkinit_cert()
 
             krb.enable_ssl()
 
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -77,7 +77,7 @@ def is_pkinit_enabled():
     if os.path.exists(paths.KDC_CERT):
         pkinit_request_ca = get_pkinit_request_ca()
 
-        if pkinit_request_ca != "SelfSign":
+        if pkinit_request_ca and pkinit_request_ca != "SelfSign":
             return True
 
     return False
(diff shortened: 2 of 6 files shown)


abbra:
The changes look good. One question I need to answer myself is what to do with the externally-provided KDC certificate. In other words, what should `ipa-pkinit-manage status` report in such case and how we should treat the certificate and its key in such case if somebody runs `ipa-pkinit-manage disable`.

flo-renaud:
I also added the new test in the nightly definitions (f28, master and rawhide).

abbra:
I think it makes better to handle that question separately -- looks like we get more and more users stumbling upon this bug. So, ACK.

abbra:
Please remove the temp commit and I'll add ACK.

flo-renaud:
@abbra 
temp commit removed.

flo-renaud:
master:

* 52c3c908756b5af3e09e8d359d11c0279f393245 ipatest: add test for ipa-pkinit-manage enable|disable
* a2301538370dc33797959bc1d7bab0c908fccf3e PKINIT: fix ipa-pkinit-manage enable|disable

PR #2404, opened 2018-09-27
Added automation for NTP options test scenarios

varunmylaraiah:
ipatests: add tests for NTP options usage on server, replica, and client
    
    The following tests are added in test_ntp_options.py :: TestNTPoptions
      - test_server_and_client_install_without_option_n
      - test_server_and_client_install_with_option_n
      - test_server_and_client_install_with_multiple_ntp_server
      - test_server_replica_and_client_install_with_ntp_pool_and_ntp_server
      - test_server_and_client_install_with_mixed_options
      - test_two_step_replica_install_using_ntp_options
      - test_two_step_replica_install_without_ntp_options
    
    Details in the ticket: https://pagure.io/freeipa/issue/7719
    and https://pagure.io/freeipa/issue/7723

PR #2638, opened 2018-12-04
Fix test_ntp_options

Tiboris:
Fix test_ntp_options to use tasks' methods
    
**Instead of custom methods use methods:**
- tasks.replica_install()
- tasks.client_install()

Move ntp_pool/server to class scope.
Every test has block finally to cleanup.
Edit tasks.client_install to return result of installation.
**Refactor install_replica task:**
- Add promote parameter to install_replica task.
- Add ntp_args to install_client call and remove from replica installation from tasks.install_replica while promoting.
- Use case while not promoting has to have user allowed to enroll a replica and server to contact in case autodiscovery does not work.

Related: https://pagure.io/freeipa/issue/7719

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -425,7 +425,11 @@ def config_replica_resolvconf_with_master_data(master, replica):
 def install_replica(master, replica, setup_ca=True, setup_dns=False,
                     setup_kra=False, setup_adtrust=False, extra_args=(),
                     domain_level=None, unattended=True, stdin_text=None,
-                    raiseonerr=True):
+                    raiseonerr=True, promote=True):
+    """
+    This task installs client and then promote it to the replica
+    """
+    replica_args = list(extra_args)  # needed for client's ntp options
     if domain_level is None:
         domain_level = domainlevel(master)
     check_domain_level(domain_level)
@@ -437,8 +441,25 @@ def install_replica(master, replica, setup_ca=True, setup_dns=False,
     # Otherwise ipa-client-install would not create a PTR
     # and replica installation would fail
     args = ['ipa-replica-install',
-            '-p', replica.config.dirman_password,
-            '-w', replica.config.admin_password]
+            '--admin-password', replica.config.admin_password]
+
+    if promote:  # while promoting we use directory manager password
+        args.extend(['--password', replica.config.dirman_password])
+        # install client on a replica machine and then promote it to replica
+        # to configure ntp options we have to pass them to client installation
+        # because promotion does not support NTP options
+        ntp_args = [arg for arg in replica_args if "-ntp" in arg]
+
+        for ntp_arg in ntp_args:
+            replica_args.remove(ntp_arg)
+
+        install_client(master, replica, extra_args=ntp_args)
+    else:
+        # for one step installation of replica we need authorized user
+        # to enroll a replica and master server to contact
+        args.extend(['--principal', replica.config.admin_name,
+                     '--server', master.hostname])
+
     if unattended:
         args.append('-U')
     if setup_ca:
@@ -458,12 +479,11 @@ def install_replica(master, replica, setup_ca=True, setup_dns=False,
     if master_authoritative_for_client_domain(master, replica):
         args.extend(['--ip-address', replica.ip])
 
-    args.extend(extra_args)
+    args.extend(replica_args)  # append extra arguments to installation
 
-    # install client on a replica machine and then promote it to replica
-    install_client(master, replica)
     fix_apache_semaphores(replica)
-    args.extend(['-r', replica.domain.realm])
+    args.extend(['--realm', replica.domain.realm,
+                 '--domain', replica.domain.name])
     fw.enable_services(fw_services)
 
     result = replica.run_command(args, raiseonerr=raiseonerr,
@@ -495,16 +515,19 @@ def install_client(master, client, extra_args=(),
     if password is None:
         password = client.config.admin_password
 
-    client.run_command(['ipa-client-install', '-U',
-                        '--domain', client.domain.name,
-                        '--realm', client.domain.realm,
-                        '-p', user,
-                        '-w', password,
-                        '--server', master.hostname]
-                       + list(extra_args))
+    result = client.run_command([
+        'ipa-client-install', '-U',
+        '--domain', client.domain.name,
+        '--realm', client.domain.realm,
+        '-p', user,
+        '-w', password,
+        '--server', master.hostname] + list(extra_args)
+    )
 
     setup_sssd_debugging(client)
     kinit_admin(client)
+
+    return result
 
 
 def install_adtrust(host):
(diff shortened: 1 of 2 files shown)


Tiboris:
Related: https://github.com/freeipa/freeipa/pull/2404

ipatests/test_integration/test_ntp_options.py
```diff
@@ -1,7 +1,7 @@
 #
 # Copyright (C) 2018  FreeIPA Contributors see COPYING for license
 #
-
+from __future__ import absolute_import
```
tiran: Note, this line is no longer necessary on master. You can leave it in if that is easier for a backport. Let's remove the absolute imports from all files together.

ipatests/test_integration/test_ntp_options.py
```diff
@@ -50,7 +39,8 @@ def test_server_client_install_without_options(self):
         assert expected_msg1 in server_install.stderr_text
         assert expected_msg2 in server_install.stdout_text
 
-        client_install = self.install_client()
+        client_install = tasks.install_client(self.master, self.client)
```
tiran: ``tasks.install_client()`` returns None. You have to modify the helper function to return the result object, too.

ipatests/test_integration/test_ntp_options.py
Tiboris: Ok i will remove it and leave it in PR for 4-7 branch

ipatests/test_integration/test_ntp_options.py
Tiboris: Will do.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -500,17 +500,20 @@ def install_client(master, client, extra_args=(),
...
-                        '-w', password,
-                        '--server', master.hostname]
-                       + list(extra_args))
+    result = client.run_command([
+        'ipa-client-install', '-U',
+        '--domain', client.domain.name,
+        '--realm', client.domain.realm,
+        '-p', user,
+        '-w', password,
+        '--server', master.Hostname] + list(extra_args)
```
flo-renaud: master.**H**ostname should be master.**hostname** instead.

Tiboris: My mistake i will fix it, THANKS!

ipatests/test_integration/test_ntp_options.py
```diff
@@ -106,89 +95,85 @@ def test_server_replica_client_install_with_pool_and_srv(self):
...
         assert expected_msg in server_install.stderr_text
         cmd = self.master.run_command(['cat', paths.CHRONY_CONF])
-        assert ntp_pool in cmd.stdout_text
-        assert ntp_server in cmd.stdout_text
+        assert self.ntp_pool in cmd.stdout_text
+        assert self.ntp_server1 in cmd.stdout_text
 
-        replica_install = self.install_replica('--ntp-pool=%s' % ntp_pool,
-                                               '--ntp-server=%s' % ntp_server)
+        replica_install = tasks.install_replica(self.master, self.client,
```
varunmylaraiah: @Tiboris Here we need one step replica install to test ntp options.
but in this case, it is trying to use Replica promotion with ntp options.

Tiboris: OK

tiran:
Test is failing with ``NTP configuration cannot be updated during promotion``

Tiboris:
@tiran install_replica method is written in a way that it promotes a client so NTP options could not be used with it. I have rewritten the method to support NTP options fir install_replica method in tasks.

wladich:
method names are wrong in commit message: not tasks.replica_install but tasks.install_replica

ipatests/test_integration/test_ntp_options.py
```diff
@@ -65,131 +55,136 @@ def test_server_client_install_no_ntp(self):
...
-        self.cleanup()
+        try:
+            server_install = tasks.install_master(self.master, setup_dns=False,
+                                                  extra_args=args)
+            assert expected_msg in server_install.stderr_text
+            cmd = self.master.run_command(['cat', paths.CHRONY_CONF])
+            assert self.ntp_pool in cmd.stdout_text
+            assert self.ntp_server1 in cmd.stdout_text
+
+            tasks.install_replica(self.master, self.replica, extra_args=args)
```
wladich: see previous comment by @varunmylaraiah -- you are replacing one-step replica installation with client + replica installation.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -458,10 +462,16 @@ def install_replica(master, replica, setup_ca=True, setup_dns=False,
     if master_authoritative_for_client_domain(master, replica):
         args.extend(['--ip-address', replica.ip])
 
-    args.extend(extra_args)
+    ntp_args = [arg for arg in replica_args if "-ntp" in arg]
```
wladich: I do not like this added logic in a simple utility method. It is very unobvious, that client and replica options should be joined in one list and that this function will mangle this list to make it "work good". In our tests sometimes we do not want it to work, we want it to fail and in this case options must be passed exactly as specified. 
So if you absolutely need to pass ntp options to client installation, I suggest adding another parameter like extra_client_args. 
But as this new feature is used only in one test (test_server_client_install_without_options)  which has outstanding issue that should be resolved, maybe this change will not be needed.

Tiboris: With changes for chrony we introduced this behavior:

We can not update chrony configuration while promoting client to replica therefore we have to something with it.

As our test task to install replica is doing the same this task should handle such situation which users of FreeIPA would have to.

I agree that it might be a good idea to have separate extra argument for client but having task to "mangle" argument and have it properly commented inline and documented we may spare some struggles while forgetting that replica promotion will not proceed with NTP options.

I am not only trying to fix test failure but to reflect behavioral changes to our tasks.

ipatests/test_integration/test_ntp_options.py
Tiboris: And what would you propose ?
should I create new task for installation of replica without promoting client or handle this situation inline in this one tescase?

Tiboris:
I have added test:
```python
def test_server_promoted_replica_client_install_with_srv(self):
```

varunmylaraiah:
@Tiboris, Looks good to me, Please remove Temp commit.

ipatests/test_integration/test_ntp_options.py
```diff
@@ -47,15 +34,18 @@ def test_server_client_install_without_options(self):
...
-        client_install = self.install_client()
-        assert expected_msg1 in client_install.stderr_text
-        assert expected_msg2 in client_install.stdout_text
+            client_install = tasks.install_client(self.master, self.client)
 
-        self.cleanup()
+            assert expected_msg1 in client_install.stderr_text
+            assert expected_msg2 in client_install.stdout_text
+
+        finally:
```
tiran: Please replace ``try/finally: self.cleanup`` with ``teardown_method`` hook: https://docs.pytest.org/en/latest/xunit_setup.html#method-and-function-level-setup-teardown

abbra:
LGTM

Tiboris:
master:

* 692cbc5d3f15720e59030c4a24912519bf6ce2bf Fix test_ntp_options to use tasks' methods

PR #2640, opened 2018-12-04
Handle service_del with bad service name

tiran:
The command 'ipa service-del badservice' used to fail with an internal
server error, because check_required_principal() could not handle a
principal that is not a service principal. All del commands have less
strict error checking of primary keys so they can reference any stored
key, even illegal ones.

check_required_principal() skips required principal check if the
principal is not a service principal. A non-service principal can never
be a required principal.

Fixes: https://pagure.io/freeipa/issue/7793

### DIFF
--- a/ipaserver/plugins/service.py
+++ b/ipaserver/plugins/service.py
@@ -276,9 +276,13 @@ def set_certificate_attrs(entry_attrs):
 
 def check_required_principal(ldap, principal):
     """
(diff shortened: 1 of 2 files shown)


abbra:
LGTM.

tiran:
master:

* c09927d16a507456802db95560b5422b921bfb4f Handle service_del with bad service name

PR #2641, opened 2018-12-05
Require Dogtag PKI 10.6.8-3

tiran:
pki-core 10.6.7 was unpushed and never landed in Fedora stable. The
latest release is 10.6.8-3 with additional fixes. The new versions are
in testing and FreeIPA's master COPR.

Also remove dependency on JSS. The dependency was originally added as a
workaround. The pki-core package already requires a newer version of JSS.

Fixes: https://pagure.io/freeipa/issue/7654

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -76,10 +76,11 @@
 
 %endif  # Fedora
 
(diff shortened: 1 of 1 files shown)


abbra:
LGTM. I did not add this to Fedora packages but we can move one with the requires there later.

PR #2648, opened 2018-12-05
Resolve user/group names in idoverride*-find

tiran:
## Resolve user/group names in idoverride*-find

ipa idoverrideuser-find and ...group-find have an --anchor argument. The
anchor argument used to support only anchor UUIDs like
':IPA:domain:UUID' or ':SID:S-sid'. The find commands now detect regular
user or group names and translate them to anchors.

## Add integration tests for idviews

Add several tests to verify new anchor override and general idview
override functionality.

## Run idviews integration tests in nightly 

Fixes: pagure.io/freeipa/issue/6594

### DIFF
--- a/ipaserver/plugins/idviews.py
+++ b/ipaserver/plugins/idviews.py
@@ -766,6 +766,40 @@ class baseidoverride(LDAPObject):
                     error=_('Default Trust View cannot contain IPA users')
                     )
 
+    def filter_for_anchor(self, ldap, filter, options, obj_type):
+        """Modify filter to support user and group names
+
+        Allow users to pass in an IPA user/group name and resolve it to an
+        anchor name.
+
+        :param ldap: ldap connection
+        :param filter: pre_callback filter
+        :param options: option dict
+        :param obj_type: 'user' or 'group'
+        :return: modified or same filter
+        """
+        anchor = options.get('ipaanchoruuid', None)
+        # return original filter if anchor is absent or correct
+        if anchor is None or ANCHOR_REGEX.match(anchor):
+            return filter
+        try:
+            resolved_anchor = resolve_object_to_anchor(
+                ldap, obj_type, anchor,
+                options.get('fallback_to_ldap', False)
+            )
+        except (errors.NotFound, errors.ValidationError):
+            # anchor cannot be resolved, let it pass through
+            return filter
+        else:
+            return ldap.make_filter(
+                {
+                    'objectClass': self.object_class,
+                    'ipaanchoruuid': resolved_anchor,
+                },
+                rules=ldap.MATCH_ALL
+            )
+
+
 class baseidoverride_add(LDAPCreate):
     __doc__ = _('Add a new ID override.')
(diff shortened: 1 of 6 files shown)


tiran:
Re-run, COPR was down for maintenance.

ipaserver/plugins/idviews.py
```diff
@@ -1128,6 +1162,15 @@ class idoverrideuser_find(baseidoverride_find):
...
                            '%(count)d User ID overrides matched', 0)
 
+    def pre_callback(self, ldap, filter, attrs_list, base_dn, scope, *args,
+                     **options):
+        result = super(idoverrideuser_find, self).pre_callback(
+            ldap, filter, attrs_list, base_dn, scope, *args, **options
+        )
+        filter, base_dn, scope = result
+        filter = self.obj.filter_for_anchor(ldap, filter, options, 'user')
+        return filter, base_dn, scope
```
rcritten: Should we start moving away from using filter as a variable name since it shadows a built-in? This is spread all over and there have been various attempts to work around this (sfilter, search_filter, filterstr, etc).

ipatests/test_integration/test_idviews.py
```diff
@@ -238,3 +239,133 @@ def test_rules_with_service_principals(self):
...
+    def test_useroverride_original_uid(self):
+        # the original UID is still valid, but getent returns the new UID
+        # I'm not sure if this behavior is a bug or a feature.
+        result = self.clients[0].run_command(
+            ['getent', 'passwd', str(self.user1_uid)]
+        )
+        expected = '{}:*:{}:{}'.format(
+            self.user1, self.user1_uid_override, self.user1_gid_override
+        )
+        assert expected in result.stdout_text
```
rcritten: Interesting. Should we check with SSSD if this is expected behavior?

rcritten:
Is very close, just a few questions (nits).

ipatests/test_integration/test_idviews.py
tiran: I talked to Sumit, it's an expected behavior.

ipaserver/plugins/idviews.py
tiran: That would be good, however I don't know if any code expects a specific argument name. Is filter always passed by position or sometimes passed by name?

rcritten: I just meant in this case. It isn't a show-stopper, it matches existing behavior.

rcritten:
ACK. Since you have to re-push to drop the temp commit can you add a comment about old uid still working when overridden at the same time?

tiran:
@rcritten I updated the comment as requested.

tiran:
master:

* 97e9e009bf21215be65676d55e19c11d5745aca7 Resolve user/group names in idoverride*-find
* 5a2a7151d505398e8a366f15109818e3772b79ac Add integration tests for idviews
* f52e0e31f7c76a3cd6b9b51aeba120c4ba3f38c9 Run idviews integration tests in nightly

PR #2649, opened 2018-12-05
Add more LDAP indices

tiran:
## Add more LDAP indices

An index is used to optimize an LDAP operation. Without an index, 389-DS
has to perform a partial or even full table scan. A full database scan can
easily take 10 seconds or more in a large installation.

* automountMapKey: eq, pres (was: eq)
* autoMountMapName: eq
* ipaConfigString: eq
* ipaEnabledFlag: eq
* ipaKrbAuthzData: eq, sub
* accessRuleType: eq
* hostCategory: eq

automountMapKey and autoMountMapName filters are used for automount.

Installation and service discovery (CA, KRA) use ipaConfigString to find
active services and CA renewal master.

SSSD filters with ipaEnabledFlag, accessRuleType, and hostCategory to
find and cache HBAC rules for each host.

ipaKrbAuthzData is used by ipa host-del. The framework performs a
``*arg*`` query, therefore a sub index is required, too.

Fixes: https://pagure.io/freeipa/issue/7786
Fixes: https://pagure.io/freeipa/issue/7787
Fixes: https://pagure.io/freeipa/issue/7790
Fixes: https://pagure.io/freeipa/issue/7792

## LDAPUpdate: Batch index tasks

The LDAPUpdate framework now keeps record of all changed/added indices
and batches all changed attribute in a single index task. It makes
updates much faster when multiple indices are added or modified.

##  Create reindex task for ipaca DB

pkispawn sometimes does not run its indextasks. This leads to slow
unindexed filters on attributes such as description, which is used
to log in with a certificate. Explicitly reindex attribute that
should have been reindexed by CA's indextasks.ldif.

See: https://pagure.io/dogtagpki/issue/3083

# NOTE

IP service performance issue is now tracked in ticket https://pagure.io/freeipa/issue/7797

### DIFF
--- a/install/share/indices.ldif
+++ b/install/share/indices.ldif
@@ -216,6 +216,40 @@ ObjectClass: top
 ObjectClass: nsIndex
 nsSystemIndex: false
 nsIndexType: eq
+nsIndexType: pres
+
+dn: cn=automountMapName,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+changetype: add
+cn: automountMapName
+ObjectClass: top
+ObjectClass: nsIndex
+nsSystemIndex: false
+nsIndexType: eq
+
+dn: cn=ipaConfigString,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+changetype: add
+cn: ipaConfigString
+objectClass:top
+objectClass:nsIndex
+nsSystemIndex: false
+nsIndexType: eq
+
+dn: cn=ipaEnabledFlag,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+changetype: add
+cn: ipaEnabledFlag
+objectClass:top
+objectClass:nsIndex
+nsSystemIndex: false
+nsIndexType: eq
+
+dn: cn=ipaKrbAuthzData,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+changetype: add
+cn: ipaKrbAuthzData
+objectClass: top
+objectClass: nsIndex
+nsSystemIndex: false
+nsIndexType: eq
+nsIndexType: sub
 
 dn: cn=ipakrbprincipalalias,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
 changetype: add
@@ -341,5 +375,21 @@ changetype: add
 cn: ipServicePort
 objectClass: top
 objectClass: nsIndex
+nsSystemIndex: false
+nsIndexType: eq
+
+dn: cn=accessRuleType,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+changetype: add
+cn: accessRuleType
+objectClass:top
+objectClass:nsIndex
+nsSystemIndex: false
+nsIndexType: eq
+
+dn: cn=hostCategory,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+changetype: add
+cn: hostCategory
+objectClass:top
+objectClass:nsIndex
 nsSystemIndex: false
 nsIndexType: eq
--- a/install/updates/20-indices.update
+++ b/install/updates/20-indices.update
@@ -151,6 +151,36 @@ default:ObjectClass: top
 default:ObjectClass: nsIndex
 default:nsSystemIndex: false
 default:nsIndexType: eq
+add:nsIndexType: pres
+
+dn: cn=automountMapName,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+default: cn: automountMapName
+default: ObjectClass: top
+default: ObjectClass: nsIndex
+default: nsSystemIndex: false
+default: nsIndexType: eq
+
+dn: cn=ipaConfigString,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+default: cn: ipaConfigString
+default: objectClass:top
+default: objectClass:nsIndex
+default: nsSystemIndex: false
+default: nsIndexType: eq
+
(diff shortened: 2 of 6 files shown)


install/share/indices.ldif
```diff
@@ -271,6 +305,22 @@ nsSystemIndex: false
...
+nsSystemIndex: false
+nsIndexType: eq
+
+dn: cn=ipServiceProtocol,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+changetype: add
+cn: ipServiceProtocol
+objectClass:top
+objectClass:nsIndex
+nsSystemIndex: false
+nsIndexType: eq
```
rcritten: IPA doesn't provide either ipServicePort or ipServiceProtocol as far as I can find. So why add an index for attributes that don't exist (by default anyway)?

tiran: 389-DS doesn't know that IPA doesn't have any entries with these attributes. When SSSD's nsswitch plugin looks for services, the LDAP server has to scan the database.

Example from https://pagure.io/freeipa/issue/7786

```
  Unindexed Component #2 (notes=U)
  -  Date/Time:             14/Nov/2018:18:31:11
  -  Connection Number:     3866581
  -  Operation Number:      27
  -  Etime:                 1.1235918834
  -  Nentries:              0
  -  IP Address:            Unknown_Host
  -  Search Base:           cn=accounts,dc=example,dc=local
  -  Search Scope:          2 (subtree)
  -  Search Filter:         (&(ipserviceport=57672)(ipserviceprotocol=tcp)(objectclass=ipservice))
```

It took over 1.1 seconds to figure out that there is no entry with an ipServicePort and ipServiceProtocol.

rcritten: Right so we should ask SSSD to stop doing this, or explain why they are.

ipaserver/install/ldapupdate.py
```diff
@@ -903,13 +907,20 @@ def create_connection(self):
             raise RuntimeError("Offline updates are not supported.")
 
     def _run_updates(self, all_updates):
+        reindex_attributes = set()
```
rcritten: nit but I'm pretty sure updates only add indices so index_attributes is probably a clearer name.

Can you add a comment block somewhere about how the values are batched up, like you have in the commit message?

tiran: I think updates can also modify an index (e.g. I'm adding a ``pres`` index to an existing index with ``eq``).

Sure, I'll add a comment.

install/share/indices.ldif
tiran: I also wonder if the order of filters matters, too. Perhaps ``(&(objectclass=ipservice)(ipserviceport=57672)(ipserviceprotocol=tcp))`` is faster than `` (&(ipserviceport=57672)(ipserviceprotocol=tcp)(objectclass=ipservice))``. I know some SQL servers have a query optimizer that reorder the terms to benefits from index with better selectability.

tiran: Also a subtree query on ``cn=accounts,dc=example,dc=local`` is pretty costly. The accounts subtree can easily contain tenths of thousands of entries.

tiran: The ipService performance issue is now tracked in ticket https://pagure.io/freeipa/issue/7797 . I've removed the indices from this PR.

install/updates/20-indices.update
```diff
@@ -151,6 +151,36 @@ default:ObjectClass: top
 default:ObjectClass: nsIndex
 default:nsSystemIndex: false
 default:nsIndexType: eq
+default:nsIndexType: pres
```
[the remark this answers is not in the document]
tiran: You are obsoletely right. Thanks for this catch.

I did some additional testing and figure out that ``add:nsIndexType: pres`` in the same ``dn`` block works, too. It adds the index when the entry already exists and does not fail when the attribute value is present.

install/share/indices.ldif
```diff
@@ -216,6 +216,40 @@ ObjectClass: top
 ObjectClass: nsIndex
 nsSystemIndex: false
 nsIndexType: eq
+nsIndexType: pres
+
+dn: cn=automountMapName,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+changetype: add
+cn: automountMapNam
```
rcritten: typo, automountMapNam

ipaserver/install/ldapupdate.py
```diff
@@ -525,19 +532,21 @@ def create_index_task(self, attribute):
...
             objectClass=['top', 'extensibleObject'],
             cn=[cn],
             nsInstance=['userRoot'],
-            nsIndexAttribute=[attribute],
+            nsIndexAttribute=list(attributes),
         )
 
-        logger.debug("Creating task to index attribute: %s", attribute)
-        logger.debug("Task id: %s", dn)
+        logger.info(
```
rcritten: IMHO this should remain at debug level because it clutters up the installer-visible output:

  [7/10]: upgrading server
Creating task cn=indextask_137634898455854420_9549,cn=index,cn=tasks,cn=config to index attributes: ipaAnchorUUID, ipaOriginalUid
Indexing finished
Creating task cn=indextask_137634898518146100_9549,cn=index,cn=tasks,cn=config to index attributes: ipaallowedtarget, ipaassignedidview, ipatokenradiusconfiglink, member, memberuid, owner, seeAlso, uniquemember
Indexing finished

rcritten:
Functionally it works ok. I tested new installs and upgrades and the indexes were created properly.

abbra:
LGTM.

tiran:
master:

* 0fb87bfe827b10af25fcd35c327b33977053e0e3 LDAPUpdate: Batch index tasks
* ed436e4b62c9d3ec02111d486ba77bd09e9838fc Add more LDAP indices
* a34d92d25c97feb778d4c222704d8ac32e146768 Create reindex task for ipaca DB

PR #2657, opened 2018-12-07
[NFS] ipa-client-automount and NFS unit name changes

fcami:
NFS unit names were renamed.
Compatibility was maintained with older unit names through symlinks. When these symlinks are removed only new unit names work, so changing to using non-symlink unit names is required in redhat/services.py.

Fixes: https://pagure.io/freeipa/issue/7783

ipa-client-client-automount:
- remove nfs-idmapd from units we enable & start as:
   - it is not used on NFS clients anymore
   - it is a static unit
 - remove rpc-gssd as well as it is a static unit
 - restart nfs-utils and rpc-gssd as needed
 - manage systemctl-related exceptions during uninstall
    
Fixes: https://pagure.io/freeipa/issue/7780
Fixes: https://pagure.io/freeipa/issue/7781

### DIFF
--- a/ipaplatform/redhat/services.py
+++ b/ipaplatform/redhat/services.py
@@ -45,8 +45,8 @@ logger = logging.getLogger(__name__)
 redhat_system_units = dict((x, "%s.service" % x)
                            for x in base_services.wellknownservices)
 
-redhat_system_units['rpcgssd'] = 'nfs-secure.service'
-redhat_system_units['rpcidmapd'] = 'nfs-idmap.service'
+redhat_system_units['rpcgssd'] = 'rpc-gssd.service'
+redhat_system_units['rpcidmapd'] = 'nfs-idmapd.service'
 redhat_system_units['domainname'] = 'nis-domainname.service'
 
 # Rewrite dirsrv and pki-tomcatd services as they support instances via separate
--- a/ipatests/test_integration/test_automount_locations.py
+++ b/ipatests/test_integration/test_automount_locations.py
@@ -1,0 +1,84 @@
+#
+# Copyright (C) 2018  FreeIPA Contributors see COPYING for license
+#
+
+"""This module provides tests for the automount location feature.
+"""
+
+from __future__ import absolute_import
+
+import time
+import re
+
+from ipatests.test_integration.base import IntegrationTest
+from ipatests.pytest_ipa.integration import tasks
+
+# give some time for units to stabilize
+# otherwise we get transient errors
+WAIT_AFTER_INSTALL = 5
+WAIT_AFTER_UNINSTALL = WAIT_AFTER_INSTALL
+
+
+class TestAutomountInstallUninstall(IntegrationTest):
+    """
+    Test if ipa-client-automount behaves as expected
+    """
+
+    num_replicas = 1
+    topology = 'star'
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master, setup_dns=False)
+        client = cls.replicas[0]
+        tasks.install_client(cls.master, client)
+
+    def test_use_automount_location(self):
+
+        client = self.replicas[0]
+
+        self.master.run_command([
+            "ipa", "automountlocation-add", "baltimore"
+        ])
+
+        self.master.run_command([
+            "ipa", "host-mod", client.hostname,
+            "--location", "baltimore"
+        ])
+
+        # systemctl non-fatal errors will only be displayed
+        # if ipa-client-automount is launched with --debug
+        result1 = client.run_command([
+            'ipa-client-automount', '--location', 'baltimore',
+            '-U', '--debug'
+        ])
+
+        # systemctl non-fatal errors will show up like this:
+        # stderr=Failed to restart nfs-secure.service: \
+        #        Unit nfs-secure.service not found.
+        # normal output:
+        # stderr=
+        m1 = re.search(r'(?<=stderr\=Failed).+', result1.stderr_text)
+        # maybe re-use m1.group(0) if it exists.
+        assert m1 is None
+
+        time.sleep(WAIT_AFTER_INSTALL)
+
+        result2 = client.run_command([
+            'ipa-client-automount', '--uninstall',
+            '-U', '--debug'
+        ])
+
+        m2 = re.search(r'(?<=stderr\=Failed).+', result2.stderr_text)
+        assert m2 is None
+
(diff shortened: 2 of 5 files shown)


ipaplatform/redhat/services.py
```diff
@@ -45,8 +45,8 @@
 redhat_system_units = dict((x, "%s.service" % x)
                            for x in base_services.wellknownservices)
 
-redhat_system_units['rpcgssd'] = 'nfs-secure.service'
-redhat_system_units['rpcidmapd'] = 'nfs-idmap.service'
+redhat_system_units['rpcgssd'] = 'rpc-gssd.service'
+redhat_system_units['rpcidmapd'] = 'nfs-idmapd.service'
```
tiran: Debian/Ubuntu has ``rpc-gssd.service`` but the ID mapper service is still called ``nfs-idmapd.service``. Please update the Debian service definition, too.

fcami: Debian already has in ipaplatform/debian/services.py:
~~~
debian_system_units['rpcgssd'] = 'rpc-gssd.service'
debian_system_units['rpcidmapd'] = 'nfs-idmapd.service'
~~~

ipatests/test_integration/test_automount_locations.py
```diff
@@ -0,0 +1,75 @@
+#
+# Copyright (C) 2018  FreeIPA Contributors see COPYING for license
+#
+
+"""
+This module provides tests for the automount location feature.
```
tiran: Style nit pick, the subject belongs on the first line:

```
"""This module provides tests for the automount location feature.
"""
```

ipatests/test_integration/test_automount_locations.py
```diff
@@ -0,0 +1,75 @@
...
+
+# give some time for units to stabilize
+# otherwise we get transient errors
+WAIT_AFTER_INSTALL = 5
+WAIT_AFTER_UNINSTALL = WAIT_AFTER_INSTALL
+
+
+class TestAutomountInstallUninstall(IntegrationTest):
+    """
+    Test if vault feature behaves as expected, when KRA is installed or not
```
tiran: The doc string doesn't match. Did you copy the vault test by any chance?

ipatests/prci_definitions/gating.yaml
```diff
@@ -243,3 +243,15 @@ jobs:
         timeout: 3600
         topology: *master_1repl
 
+  fedora-29/test_automount_locations:
```
tiran: automount is neither super critical nor likely to fail. IMHO it's sufficient to test the feature in the nightly tests.

ipatests/test_integration/test_automount_locations.py
```diff
@@ -0,0 +1,75 @@
...
+            "ipa", "host-mod", client.hostname,
+            "--location", "baltimore"
+        ])
+
+        result = client.run_command([
+            'ipa-client-automount', '--location', 'baltimore',
+            '-U', '--debug'
+        ])
+
+        assert result.returncode == 0
```
tiran: You don't need to check the result code unless you use ``raiseonerr=True`` and expect an error.

ipatests/test_integration/test_automount_locations.py
```diff
@@ -0,0 +1,75 @@
...
+
+        time.sleep(WAIT_AFTER_INSTALL)
+
+        result = client.run_command([
+            'ipa-client-automount', '--uninstall',
+            '-U', '--debug'
+        ])
+
+        assert result.returncode == 0
```
tiran: Same here

ipatests/test_integration/test_automount_locations.py
```diff
@@ -0,0 +1,75 @@
...
+WAIT_AFTER_UNINSTALL = WAIT_AFTER_INSTALL
+
+
+class TestAutomountInstallUninstall(IntegrationTest):
+    """
+    Test if vault feature behaves as expected, when KRA is installed or not
+    installed on replica
+    """
+
+    num_replicas = 1
```
tiran: I think you can use ``num_replicas = 0`` and ``num_clients = 1`` here, too. That would allow you to avoid one expensive ipa-replica-install and just use a faster client installation. Of course you need to change topology to ``topology: *master_1repl_1client``.

Perhaps it's time to introduce a ``topology: *master_1client``. I have several other tests that just need a master and one client, too.

fcami: We will need two clients in an upcoming version of that test. I'll see if there a topology that fits better.

ipatests/test_integration/test_automount_locations.py
fcami: Yes obviously and I missed that. Will fix.

ipatests/test_integration/test_automount_locations.py
fcami: Actually what I would like for the next class (which will not be part of the current PR but comes in next sprint) is master_2clients but we don't seem to have that. Please let me know if we can resolve this now.

ipatests/test_integration/test_automount_locations.py
fcami: Fixed, thanks.

ipatests/test_integration/test_automount_locations.py
fcami: Thanks!
I've replaced that with a regular expression operation that checks for stderr=.+ in the debug output. It will detect non-fatal systemctl errors e.g. when systemctl fails but still returns 0. 
I do expect it to fail first and I'll update the test afterwards.

ipatests/test_integration/test_automount_locations.py
fcami: Same fix as above.

fcami:
re-run

abbra:
@fcami does this work also for F28?

fcami:
@abbra I tested an earlier version and it did work on F28. I can rebuild and test on F28 if you like. As stated earlier this PR is not a complete fix for NFS configuration though.

fcami:
Temp commit removed.

abbra:
ACK.

tiran:
master:

* 97f87513ef345908c74f27c71e7765c16d3abe8c Fix NFS unit names
* 6a56aa6d4987bc4856997351a413c014e14abdd6 ipa-client-automount: use nfs-utils unit
* 7c0b0f34716cabbdc48edf959935b37a21883238 ipatests: add a test for ipa-client-automount

PR #2659, opened 2018-12-07
ipatests: fix test_full_backup_and_restore

flo-renaud:
The test is failing when calling (on the replica) `ipa-replica-manage re-initialize --from <master>` because the tool needs to resolve master.
The test does not set `/etc/resolv.conf` on the replica, as a consequence it relies on whatever DNS server is configured in your test environment prior to launching the test, and makes the test unreliable.
In PR-CI env, `/etc/resolv.conf` points to the machine hosting the replica vm, which is unable to resolve master.ipa.test.

The fix is modifying the replica's /etc/resolv.conf to use the master as DNS.

Fixes https://pagure.io/freeipa/issue/7778

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -418,7 +418,7 @@ def master_authoritative_for_client_domain(master, client):
     return result.returncode == 0
 
 
-def _config_replica_resolvconf_with_master_data(master, replica):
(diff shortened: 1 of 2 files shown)


flo-renaud:
Temp commit removed and rebased to master.

tiran:
master:

* a81ea9af19c284280f5edcc0f9c75af8f6a7fa95 ipatests: fix test_full_backup_and_restore

PR #2661, opened 2018-12-07
Add index and container for RFC 2307 IP services

tiran:
IPA doesn't officially support RFC 2307 IP services. However SSSD has a
nsswitch plugin to provide service lookups. The subtree search for
(&(ipserviceport=$PORT)(ipserviceprotocol=$SRV)(objectclass=ipservice)) in
cn=accounts,$SUFFIX has caused performance issues on large
installations.

This patch introduced a dedicated container
cn=ipservices,cn=accounts,$SUFFIX for IP services for future use or 3rd
party extensions. SSSD will be change its search base in an upcoming
release, too.

A new ipServicePort index is added to optimize searches for an IP
service by port. There is no index on ipServiceProtocol because the index
would have poor selectivity. An ipService entry has either 'tcp' or 'udp'
as protocol.

Fixes: https://pagure.io/freeipa/issue/7797
See: https://pagure.io/freeipa/issue/7786

Also see #2649

### DIFF
--- a/install/share/bootstrap-template.ldif
+++ b/install/share/bootstrap-template.ldif
@@ -34,6 +34,12 @@ objectClass: top
 objectClass: nsContainer
 cn: hostgroups
 
+dn: cn=ipservices,cn=accounts,$SUFFIX
+changetype: add
+objectClass: top
+objectClass: nsContainer
+cn: ipservices
+
 dn: cn=alt,$SUFFIX
 changetype: add
 objectClass: nsContainer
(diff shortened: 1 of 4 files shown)


abbra:
LGTM. Since there is no code in IPA to add objects to RFC2307 IP services, we do not test them.

tiran:
master:

* 39eaf2fab5e27bd12edfb2a24c439a8ea5fb26f0 Add index and container for RFC 2307 IP services

PR #2664, opened 2018-12-10
Always collect test logs

tiran:
mh.install() is the default multi host installer. Most integration test
classes use it to install master, replicas, and clients. In case of a
failed installation, the test collector step is skipped.

Guard log collection with a try/finally block so logs are always
collected.


tiran:
The temp commit deliberately breaks installation to verify that a failed ipa-server-install makes installation and systemd logs available.

tiran:
master:

* 18f610caeadf6fd80c31f951cb7971cec82e4fa1 Always collect test logs

PR #2670, opened 2018-12-11
Add index on idnsName

tiran:
The data structures for the internal DNS server use the attribute idnsName
instead of cn in the DN. It's also used to search for entries when entries
are added, modified, or removed.

The new index speeds up dnsrecord and dnszone related commands as well
as commands like host-add and host-del --updatedns.

Fixes: https://pagure.io/freeipa/issue/7803

Note, this PR contain changes from https://github.com/freeipa/freeipa/pull/2649 to a avoid merge conflict.

### DIFF
--- a/install/share/indices.ldif
+++ b/install/share/indices.ldif
@@ -393,3 +393,11 @@ objectClass:top
 objectClass:nsIndex
 nsSystemIndex: false
 nsIndexType: eq
+
+dn: cn=idnsName,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+changetype: add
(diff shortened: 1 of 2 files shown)


tiran:
``kadmin.service`` failed, re-run

flo-renaud:
@tiran 
thanks for the PR, works for me. I tested with
- upgrade: the index is properly added and the upgrade also creates an index task to create the missing index
- fresh install: the index is properly configured.

tiran:
master:

* 7688808d1f1574f1f77fae90d95cee60ca0badf2 Add index on idnsName

PR #2672, opened 2018-12-11
Replace nss.conf with zero-length file instead of removing

xxblx:
Empty nss.conf avoids recreation of nss.conf in case `mod_nss` package is reinstalled. It is needed because by default (e.g. recreated) nss.conf has `Listen 8443` while this port is also used by dogtag. 

Fixes: https://pagure.io/freeipa/issue/7745

### DIFF
--- a/ipaserver/install/ipa_restore.py
+++ b/ipaserver/install/ipa_restore.py
@@ -144,8 +144,12 @@ class Restore(admintool.AdminTool):
         paths.DNSSEC_TOKENS_DIR,
     ]
 
-    FILES_TO_BE_REMOVED = [
-        paths.HTTPD_NSS_CONF,
+    FILES_TO_BE_REMOVED = []
+
+    # files listed here cannot be removed and these files will be
+    # replaced with zero-length files
+    FILES_TO_BE_CLEARED = [
+        paths.HTTPD_NSS_CONF
     ]
 
     def __init__(self, options, args):
@@ -404,6 +408,7 @@ class Restore(admintool.AdminTool):
             # We do either a full file restore or we restore data.
             if restore_type == 'FULL':
                 self.remove_old_files()
+                self.clear_old_files()
                 self.cert_restore_prepare()
                 self.file_restore(options.no_logs)
                 self.cert_restore()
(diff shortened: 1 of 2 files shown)


tiran:
The backup restore tool has a list of files that get removed on restore:

https://github.com/freeipa/freeipa/blob/f330c59dd8946b649c0d44c17d86bce2b836e993/ipaserver/install/ipa_restore.py#L147-L149

ipaserver/install/ipa_restore.py
```diff
@@ -144,8 +144,10 @@ class Restore(admintool.AdminTool):
         paths.DNSSEC_TOKENS_DIR,
     ]
 
-    FILES_TO_BE_REMOVED = [
```
tiran: Please leave ``FILES_TO_BE_REMOVED`` and just add ``FILES_TO_BE_CLEARED`` as additional list. We might need the remove feature in the future.

ipaserver/install/ipa_restore.py
```diff
@@ -714,12 +717,16 @@ def remove_old_files(self):
...
+        before backup
+        """
+        for f in self.FILES_TO_BE_CLEARED:
             try:
-                os.remove(f)
-            except OSError as e:
-                if e.errno != 2:  # 2: file does not exist
-                    logger.warning("Could not remove file: %s (%s)", f, e)
+                open(f, 'w').close()
+            except IOError as e:
```
tiran: ``os.remove`` is an alias for ``os.unlink``. The unlink function will never raise IOError.

I think the logic is wrong here, at least for the special case of ``nss.conf``. It should be cleared no matter what its previous state was. The function only clears it when it already exists but doesn't clear it when it is missing. Shouldn't it be created as empty file on both update and restore?

xxblx: > `os.remove` is an alias for `os.unlink`. The unlink function will never raise IOError.

`os.remove` with try-except is a part of original code, it isn't mine. 

I added `OSError` for logging situations when it is not possible to clear file (open it for writing), e.g. no permissions for writing. Actually python3 has `PermissionError` exception (based on `OSError`) for that, but because we are supprting python2 at 4.7 I have used `OSError` 

> Shouldn't it be created as empty file on both update and restore?
Yes, I think you are right. Thanks for an advice.

xxblx:
master:

* f28a8177a92938ecbbc5d8fa8f21ed995f00788f Replace nss.conf with zero-length file instead of removing

PR #2676, opened 2018-12-13
ipa-getkeytab: resolve symlink

tiran:
Resolve one level of symbolic links to support a dangling symlink as
keytab target. To prevent symlink attacks, only resolve symlink when the
symlink is owned by the current effective user and group.

Fixes: https://pagure.io/freeipa/issue/4607

### DIFF
--- a/client/ipa-getkeytab.c
+++ b/client/ipa-getkeytab.c
@@ -672,6 +672,56 @@ int read_ipa_config(struct ipa_config **ipacfg)
     return 0;
 }
 
+static int resolve_ktname(const char *keytab, char **ktname, char **err_msg)
+{
+	char keytab_resolved[PATH_MAX + 1];
+	struct stat st;
+	struct stat lst;
+	int ret;
+
+	*err_msg = NULL;
+
+    /* Resolve keytab symlink to support dangling symlinks, see
+     * https://pagure.io/freeipa/issue/4607. To prevent symlink attacks,
+     * the symlink is only resolved owned by the current user or by
+     * root. For simplicity, only one level if indirection is resolved.
+     */
+    if ((stat(keytab, &st) == -1) &&
+            (errno = ENOENT) &&
+            (lstat(keytab, &lst) == 0) &&
+            (S_ISLNK(lst.st_mode))) {
+        /* keytab is a dangling symlink. */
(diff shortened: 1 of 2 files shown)


rcritten:
I asked a question in the issue: would it be possible to override the directory values by treating container as a platform? If so it could eliminate the need to deal with symlinks at all.

rcritten:
LGTM but it would be nice to get another set of eyes on it.

abbra:
So now if our keytab is a dangling symlink but not owned by the same user, we simply try to write to that symlink and fail later as it was originally? Wouldn't we instead refuse to operate on a symlink and exit?

abbra:
I'm mostly thinking along the lines of separating euid/egid check inside the main if() so that we handle symlink case right in the first branch and never deal with it in the 'else'. In that first branch you'd exit the program if euid/egid check is not passing.

tiran:
There are six different cases

1. file exists
2. file does not exist
3. symlink to an existing, symlink is owned by current user
4. symlink to an existing, symlink is owned by different user
5. dangling symlink, symlink is owned by current user
6. dangling symlink, symlink is owned by different user

Cases 1 to 5 are currently covered. To cover case 6 without causing trouble for case 4, I need change the logic.

abbra:
we use `kadmin.local` to create named.keytab, I don't think it is affected.

tiran:
We could also consider symlinks owned by ``root:root`` as trusted:

```
        if (((st.st_uid == geteuid()) && (st.st_gid == getegid())) ||
                ((st.st_uid == 0) && (st.st_gid == 0))) {
            /* Either root or current user own symlink, resolve symlink and
             * return the resolved symlink. */
```

abbra:
right

tiran:
@rcritten The extra lookup is only triggered for dangling symlinks (stat returns ENOENT, lstat returns a symlink). In case of a refresh, keytab file already exists and the owner check is not executed.

rcritten:
Right, of course. So I'm ok with the patch as-is, adding the escape for root I think would be a nice change given that if the user is root then they can play whatever games they want anyway.

tiran:
master:

* 53e0b2255d92c9c21c19306cf37cc8de0476dc9c ipa-getkeytab: resolve symlink

PR #2677, opened 2018-12-13
Don't configure KEYRING ccache in containers

tiran:
environment.


rcritten:
I think I'd like @Tiboris to take a look since it affects containers. This looks fine to me but I don't know all the corner cases.

tiran:
@Tiboris ping, please review.

Tiboris:
master:

* 165a941109a9a5f7ac8f85bdda93b4132875a7b1 Don't configure KEYRING ccache in containers

PR #2681, opened 2018-12-13
WebUI: Temporary fix for UnexpectedAlertPresentException

serg-cymbaluk:
It is regression in Firefox 55
Fixed in Firefox 65: https://bugzilla.mozilla.org/show_bug.cgi?id=1503015

Ticket: https://pagure.io/freeipa/issue/7809

### DIFF
--- a/ipatests/test_webui/ui_driver.py
+++ b/ipatests/test_webui/ui_driver.py
@@ -40,6 +40,7 @@ try:
     from selenium.common.exceptions import NoSuchElementException
     from selenium.common.exceptions import InvalidElementStateException
     from selenium.common.exceptions import StaleElementReferenceException
+    from selenium.common.exceptions import UnexpectedAlertPresentException
     from selenium.common.exceptions import WebDriverException
     from selenium.common.exceptions import ElementClickInterceptedException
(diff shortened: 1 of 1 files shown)


ipatests/test_webui/ui_driver.py
```diff
@@ -37,16 +37,20 @@
 
 try:
     from selenium import webdriver
-    from selenium.common.exceptions import NoSuchElementException
-    from selenium.common.exceptions import InvalidElementStateException
-    from selenium.common.exceptions import StaleElementReferenceException
-    from selenium.common.exceptions import WebDriverException
-    from selenium.common.exceptions import ElementClickInterceptedException
+    from selenium.common.exceptions import (
```
wladich: It is a good point of replacing all those imports with one import expression
But this commit will be most likely reverted when bug is fixed in firefox, so this refactoring effort will be lost.

serg-cymbaluk: Agree, fixed

serg-cymbaluk:
master:

* 18cac460928bac6bc21de1faa2dbb088ae414086 WebUI: Temporary fix for UnexpectedAlertPresentException

PR #2682, opened 2018-12-13
replication: check remote ds version before editing attributes

flo-renaud:
When the remote server has an old DS version, update of the
replication attributes nsds5ReplicaReleaseTimeout nsds5ReplicaBackoffMax
and nsDS5ReplicaBindDnGroupCheckInterval fails even if the remote
schema has been updated.

Check first the remote server version and update the attributes only if
the version is high enough.
A previous fix was already performing this check (commit 02f4a7a),
but not in all the cases. This fix also handles when the remote server
already has a cn=replica entry (for instance because it has already
established replication with another host).

Fixes https://pagure.io/freeipa/issue/7796

### DIFF
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -215,6 +215,22 @@ def wait_for_entry(connection, dn, timeout, attr=None, attrvalue='*',
             time.sleep(1)
 
 
+def get_ds_version(conn):
+    """Returns the DS version
+
+    Retrieves the DS version from the vendorVersion attribute stored in LDAP.
+    :param conn: LDAP connection established and authenticated to the server
+                 for which we need the version
+    :return: a tuple containing the DS version
+    """
+    # Find which 389-ds is installed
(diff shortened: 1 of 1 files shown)


tiran:
Thanks for the fix!

flo-renaud:
Setting WIP as I would like to test when deploying a new replica with an old master (for instance rhel 6.8 master).

flo-renaud:
Removing WIP. I did a build with the same patch on ipa-4-6, installed 2 replicated RHEL 6.8 masters and created a replica file for my new 4.6 replica with ipa-replica-prepare. Install went successfully.
(Note: I backported to ipa-4-6 in order to have domain level 0 support).

tiran:
Thanks for fixing my mess, Flo! :100:

tiran:
master:

* faa122a8b8fc9a6850811ed1466166a7bff13be3 replication: check remote ds version before editing attributes

PR #2691, opened 2018-12-14
Fix test_advise in nightly runs

tiran:
test_advise now needs one client, too.

See: https://pagure.io/freeipa/issue/7751


tiran:
master:

* caffe2e8dd14a9fa8887db72bfbfbab11f5e5c0e Fix test_advise in nightly runs

PR #2693, opened 2018-12-14
ipa-sidgen: make internal fetch_attr helper really internal

abbra:
With 389-ds landing a change for
https://pagure.io/389-ds-base/issue/49950, fetch_attr() helper function
is exposed in slapi-plugin.h. However, in order to be able to build
FreeIPA plugins against older 389-ds versions, prefer using a local
variant of it.

Rename fetch_attr() to ipa_sidgen_fetch_attr() so that it doesn't
conflict at all.

Fixes: https://pagure.io/freeipa/issue/7811


tiran:
master:

* e62eb0007b86e839aba08af5cb673707100be5de ipa-sidgen: make internal fetch_attr helper really internal

PR #2697, opened 2018-12-14
LDAPClient: Prefer LDAPI connections in installer and simplify API

tiran:
On several occasions the installer uses LDAP + simple_bind rather than LDAPI with EXTERNAL bind. LDAPI (LDAP over Unix socket) is more secure, a tiny bit faster, and is immune to some network/DNS related issues.

The changeset simplifies the LDAPClient API by introducing additional constructors. The new constructors are not only easier to use but also prevent some common pitfalls like LDAP connection without StartTSL. Several code paths now use ``LDAPClient.from_realm(realm_name)`` with external bind as root instead of LDAP with DM password.

### Add constructors to ldap client

Add LDAPClient.from_realm(), LDAPClient.from_hostname_secure(), and
LDAPClient.from_hostname_plain() constructors.

The simple_bind() method now also refuses to transmit a password over a
plain, unencrypted line.

LDAPClient.from_hostname_secure() uses start_tls and FreeIPA's CA cert
by default. The constructor also automatically disables start_tls for
ldaps and ldapi connections.

### Use new LDAPClient constructors

Replace get_ldap_uri() + LDAPClient() with new LDAPClient constructors
like LDAPClient.from_realm().

Some places now use LDAPI with external bind instead of LDAP with simple
bind. Although the FQDN *should* resolve to 127.0.0.1 / [::1], there is
no hard guarantee. The draft
tools.ietf.org/html/draft-west-let-localhost-be-localhost-04#section-5.1
specifies that applications must verify that the resulting IP is a
loopback API. LDAPI is always local and a bit more efficient, too.

The simple_bind() method also prevents the caller from sending a
password over an insecure line.

### Let 389-DS configure LDAPI for us
    
The new lib389 installer configures LDAPI with correct socket path by
default. Use LDAPI to boot strap the IPA domain and autobind.

### Use LDAPS when installing CA on replica

On a replica, 389-DS is already configured for secure connections when
the CA is installed.

### Move realm_to_serverid/ldap_uri to ipaldap

The helper function realm_to_serverid() and realm_to_ldap_uri() are
useful outside the server installation framework. They are now in
ipapython.ipaldap along other helpers for LDAP handling in FreeIPA.


install/migration/migration.py
```diff
@@ -55,7 +55,7 @@ def bind(ldap_uri, base_dn, username, password):
         raise IOError(errno.EIO, 'Cannot get Base DN')
     bind_dn = DN(('uid', username), ('cn', 'users'), ('cn', 'accounts'), base_dn)
     try:
-        conn = ipaldap.LDAPClient(ldap_uri)
+        conn = ipaldap.LDAPClient(ldap_uri, start_tls=True)
```
rcritten: So this may get confusing in the future. The ldap_uri as currently called is api.env.ldap_uri which is an ldapi value (by default) so start_tls doesn't make sense. I believe there are similar changes elsewhere.

Should start_tls be True by default or is that not a backwards-compatible change?

ipaserver/install/dsinstance.py
```diff
@@ -199,6 +199,7 @@ def get_all_external_schema_files(root):
 RootDNPwd= $PASSWORD
 InstallLdifFile= /var/lib/dirsrv/boot.ldif
 inst_dir=   /var/lib/dirsrv/scripts-$SERVERID
+ldapifilepath=/var/run/slapd-$SERVERID.socket
```
rcritten: This should use paths.SLAPD_INSTANCE_SOCKET_TEMPLATE.

ipaserver/install/dsinstance.py
```diff
@@ -584,9 +584,20 @@ def __create_instance(self):
...
+            password=self.dm_password
+        )
+
+        # local_simple_allocate() configures LDAPI but doesn't set up the
+        # DirSrv object to use LDAPI.
+        inst.ldapi_enabled = 'on'
+        inst.ldapi_socket = paths.SLAPD_INSTANCE_SOCKET_TEMPLATE % (
+            self.serverid
+        )
+        inst.ldapi_autobind = 'off'
```
rcritten: autobind is enabled in the next installation step. Is there a reason to start with it disabled?

ipapython/ipaldap.py
```diff
@@ -778,15 +777,9 @@ def __init__(self, ldap_uri, start_tls=False, force_schema_updates=False,
             syntax.
         """
         if ldap_uri is not None:
+            # special case for ldap2 sever plugin
```
rcritten: nit: typo sever -> server

ipapython/ipaldap.py
```diff
@@ -1156,7 +1199,8 @@ def _connect(self):
             if not self._sasl_nocanon:
                 conn.set_option(ldap.OPT_X_SASL_NOCANON, ldap.OPT_OFF)
 
-            if self._start_tls:
+            if self._start_tls and self.protocol == 'ldap':
+                # STARTTLS applies only to ldap:// connections
```
rcritten: Should this be an assert in the constructor instead to prevent this sort of bogus combination?

ipapython/ipaldap.py
```diff
@@ -1166,6 +1210,9 @@ def simple_bind(self, bind_dn, bind_password, server_controls=None,
         """
         Perform simple bind operation.
         """
+        if self.protocol == 'ldap' and not self._start_tls and bind_password:
+            # non-empty bind must use a secure connection
+            raise ValueError('simple_bind over insecure LDAP connection')
```
rcritten: Similar question. Is it better to catch this at construction or at bind? If protocol == ldap and not start_tls then regardless of password this is bad, right?

ipaserver/install/ipa_backup.py
```diff
@@ -400,7 +400,7 @@ def get_connection(self):
             self._conn.external_bind()
         except Exception as e:
             logger.error("Unable to bind to LDAP server %s: %s",
-                         self._conn.host, e)
+                         self._conn.ldap_uri, e)
```
rcritten: Having the full uri might be handy because it includes the protocol used so one can distinguish a network (ldap/ldaps) connection vs a local socket (ldapi).

ipaserver/dnssec/ldapkeydb.py
```diff
@@ -467,7 +467,7 @@ def zone_keypairs(self):
 
     # LDAP initialization
     dns_dn = DN(ipalib.api.env.container_dns, ipalib.api.env.basedn)
-    ldap = ipaldap.LDAPClient(ipalib.api.env.ldap_uri)
+    ldap = ipaldap.LDAPClient(ipalib.api.env.ldap_uri, start_tls=True)
     logger.debug('Connecting to LDAP')
```
rcritten: A similar point. This one explicitly uses ldap_uri which IPA sets as ldapi. Is this to prevent users who might muck around with things from making an insecure connection? Because otherwise one might consider start_tls to conflict with ldapi (at best it is a no-op).

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -403,6 +403,7 @@ def test_replica_install_with_existing_entry(self):
             realm=replica.domain.name.upper())
         master.put_file_contents(ldif_file, entry_ldif)
         arg = ['ldapmodify',
+               '-ZZ',
```
rcritten: Perhaps out-of-scope but would it make sense to have ldapsearch and ldapmodify helpers to avoid all this duplication?

install/migration/migration.py
tiran: I see where it can lead to confusion.

The PR also changes when STARTTLS is performed. The option is now ignored when the protocol is LDAPI or LDAPS.

``start_tls=True`` as default is not fully backward compatible. Windows doesn't support GSSAPI and STARTTLS at the same time.

ipaserver/install/dsinstance.py
tiran: ACK

ipaserver/install/dsinstance.py
tiran: This line doesn't change the setting in LDAP. It merely tells the 389-DS installer that ldap_autobind is not available yet and that it has to use password bind. I'll add a comment.

ipapython/ipaldap.py
tiran: No, it is not necessarily bad. GSSAPI + SASL also provide confidentiality and strong authentication over a plain LDAP connection. IPA uses plain LDAP + GSSAPI to fetch the CA certificate. Only password binds over plain LDAP are bad, because the combination transmits the password in plain text.

ipatests/test_integration/test_replica_promotion.py
tiran: +1

ipaserver/dnssec/ldapkeydb.py
tiran: I removed the ``start_tls=True`` argument again. The LDAP key DB for DNSSEC uses GSSAPI and is therefore not vulnerable.

f-trivino:
ci-large-17 is ok now, consider to re-run if you need it.

tiran:
@rcritten Please review again. I have addressed your remarks.

ipaserver/install/custodiainstance.py
```diff
@@ -5,6 +5,7 @@
 import enum
 import logging
 
+import ipapython.ipaldap
```
rcritten: Not a deal breaker but you're using a different import style in different files. In some cases you do 'from ipapython import ipaldap' and in others 'import ipapython.ipaldap'. Is there a reason?

install/migration/migration.py
```diff
@@ -54,8 +54,11 @@ def bind(ldap_uri, base_dn, username, password):
         logger.error('migration unable to get base dn')
         raise IOError(errno.EIO, 'Cannot get Base DN')
     bind_dn = DN(('uid', username), ('cn', 'users'), ('cn', 'accounts'), base_dn)
+    # ldap_uri should be ldapi:// in all common cases. Enforce start_tls just
+    # in case it's a plain LDAP connection.
+    start_tls = ldap_uri.startswith('ldap://')
     try:
-        conn = ipaldap.LDAPClient(ldap_uri)
+        conn = ipaldap.LDAPClient(ldap_uri, start_tls=start_tls)
```
rcritten: ldap_uri in this case comes from api.env.ldap_uri which is going to be an ldapi uri. start_tls will be True. I don't think this will cause problems but I haven't tested it, have you?

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -1613,3 +1608,48 @@ def group_add(host, groupname, extra_args=()):
     ]
     cmd.extend(extra_args)
     return host.run_command(cmd)
+
+
+def ldapmodify_dm(host, ldif_text, **kwargs):
+    """Run ldapmodify as Directory Manager
+
+    :param host: host object
+    :param ldap_text: ldap string
```
rcritten: nit: ldif_text vs ldap_text

rcritten:
Overall this is looking excellent, just a couple of minor quibbles.

ipaserver/install/custodiainstance.py
tiran: Please refresh, you are referring to an older version of the PR. The current PR uses ``from ipapython import ipaldap``.

install/migration/migration.py
tiran: Are you also seeing an older version of the patch here? The new code is ``start_tls = ldap_uri.startswith('ldap://')``, so only ``start_tls=True`` for ``ldap`` URI but not for ``ldapi`` and ``ldaps``.

ipaserver/install/custodiainstance.py
rcritten: I see. It was added in one patch in ipaserver/install/custodiainstance.py and removed in a subsequent one.

install/migration/migration.py
rcritten: You're right, my mistake.

rcritten:
My issues have been addressed, LGTM. ack if the tests pass after removing the temp commit.

rcritten:
master:

* d5d9233b7c94eb33d6c7341899543224c14d2287 Move realm_to_serverid/ldap_uri to ipaldap
* 5be9341fbabaf7bcb396a2ce40f17e1ccfa54b77 Add constructors to ldap client
* a3934a211d5fb601d2ac501fcecf05039ef4c10b Use new LDAPClient constructors
* 1a2ceb155759e6640ed8e07fdaa0da02ac1b60e9 Use secure LDAP connection in tests
* e4fff90074797cdc368f3d52169b8f20440ecd57 Use LDAPS when installing CA on replica
* d87a3b01e0dd370beb502fecacf672c5d3033996 Let 389-DS configure LDAPI for us
* 73bc11a20b904a1203201f5525f147fe1737e960 Add ldapmodify/search helper functions

PR #2698, opened 2018-12-14
replica install: set the same master as preferred source for domain a…

flo-renaud:
…nd CA

During ipa-replica-install, the installer creates a ReplicaConfig
object that contains a config.ca_host_name attribute, built from
api.env.ca_host.
This attribute is used as preferred source when asking the DNS for a CA
master from which to initialize the CA instance
(see commit 8decef33 for master selection and preferred host).

In most of the cases, /etc/ipa/default.conf does not contain any
definition for ca_host. In this case, api.env.ca_host is set to
the local hostname.
As a consequence, replica install is trying to use the local host
as preferred source (which does not have any CA yet), and the method
to find the CA source randomly picks the CA in the DNS.

With the fix, the master picked for domain replication is also used as
preferred source for CA/KRA.

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -812,7 +812,8 @@ def promote_check(installer):
     config.host_name = api.env.host
     config.domain_name = api.env.domain
     config.master_host_name = api.env.server
-    config.ca_host_name = api.env.ca_host
+    # Try to use same master for CA install
+    config.ca_host_name = api.env.server
     config.kra_host_name = config.ca_host_name
     config.ca_ds_port = 389
     config.setup_ca = options.setup_ca


flo-renaud:
Note: this PR should also fix the failing test test_server_del.py::TestServerDel::test_removal_of_replica1_disconnects_domain_topology.

Test was failing because the topology obtained after installation did not match the expectations (the second replica was picking a random server for CA replication, while the tests expected to always pick the master).

tiran:
Does your fix also address the same problem as described in https://github.com/freeipa/freeipa/pull/2580 / https://pagure.io/freeipa/issue/7744 ?

flo-renaud:
> Does your fix also address the same problem as described in #2580 / https://pagure.io/freeipa/issue/7744 ?

Yes, it should also address the issue.

tiran:
The server del tests are passing again. Good work, @flo-renaud! :+1:

flo-renaud:
@tiran 
thanks for the review. I removed the temp commit, rebased and included the upstream ticket 7744 in the commit msg

flo-renaud:
As commit 8decef3 (Unify and simplify LDAP service discovery) was only pushed to the master branch, I am also setting ipa-next label for this PR.

tiran:
master:

* c0fd5e39c726ef4dc12e87a2f9c08ebb32ed27fe replica install: set the same master as preferred source for domain and CA

PR #2706, opened 2018-12-20
pkinit enable: use local dogtag only if host has CA

flo-renaud:
## pkinit enable: use local dogtag only if host has CA

`ipa-pkinit-manage enable` is failing if called on a master that does not have a CA instance, because it is trying to contact dogtag on the localhost.
The command should rather use certmonger in this case, and let certmonger contact the right master to request the KDC certificate.

Fixes: https://pagure.io/freeipa/issue/7795

## ipatests: add integration test for pkinit enable on replica

`ipa-pkinit-manage` enable was failing when run on a replica without a CA instance.
Add a test with the following scenario:
- install a replica with `--no-pkinit`
- check that the KDC cert is self signed
- call `ipa-pkinit-manage enable`
- check that the KDC cert is signed by IPA CA

Related to https://pagure.io/freeipa/issue/7795

### DIFF
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -45,6 +45,7 @@ from ipaserver.install import replication
 from ipaserver.install import ldapupdate
 
 from ipaserver.install import certs
+from ipaserver.masters import find_providing_servers
 from ipaplatform.constants import constants
 from ipaplatform.tasks import tasks
 from ipaplatform.paths import paths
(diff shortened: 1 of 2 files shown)


fcami:
I'm working on the review.

fcami:
- code quality - OK, pylint 10/10
- PR-CI successful - OK for test_pkinit_manage
- test_pkinit_manage tests if  ipa-pkinit-manage can be run on a replica without a CA - OK
... and I can't imagine a simpler/better way to fix the issue.
Could you please remove the temp commit so that we get a full PR-CI run?
Other than that, this is approved.

flo-renaud:
@fcami thanks for the review, temp commit removed and PR rebased.

flo-renaud:
master:

* 778521053336a4ba09923b4b1f9cac0dff72f634 pkinit enable: use local dogtag only if host has CA
* 64be3141da07f459fafa93aec53bbd31683de551 ipatests: add integration test for pkinit enable on replica

PR #2726, opened 2019-01-04
Remove 389-ds templates now that lib389 is used for installs

rcritten:
The templates created the inf files for calling the 389-ds
installer setup-ds.pl. Now that lib389 is being used for installation
these are no longer necessary.

Related: https://pagure.io/freeipa/issue/4491


tiran:
master:

* 9b90ebf4c33e5c4ff4068d557521f8718ec62a36 Remove 389-ds templates now that lib389 is used for installs

PR #2729, opened 2019-01-06
Don't use Python dependency generator yet

tiran:
Fedora 30 started to have python_enable_dependency_generator by default.
Some packages like python3-dbus don't have the new dist names yet. This
fix enables testing on rawhide.
    
https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/


abbra:
LGTM

tiran:
master:

* 8276caf865fa3f8900a08b42a56bba0cf3972184 Don't use Python dependency generator yet

PR #2732, opened 2019-01-08
Dcerpc fix conflict resolution

abbra:
This pull request attempts to fix an issue found by Red Hat QE in our trust code. When we try to automatically handle a topology collision during creation of a forest trust, we might get back a topology that includes a number of records of a type we don't expect (see more details in the commit message). The solution is to make sure we use address those structure's members by their proper names.

A second change is fixing a failing 'fastcheck' for me.


abbra:
I'm working on a mockup test for the whole collision resolution flow but it is not an easy thing to implement. Thus, it has to be tested manually -- once there is a PR CI build of this pull request, I'll ask Red Hat QE to re-try with those packages.

abbra:
I've added ipa-4-6 backport label as I realized this may have been a problem in 4.6 too.

abbra:
@varunmylaraiah  found that we still return 'internal server error' even though processing is now not crashing. I added another condition that should return a proper error message to the client that we were unable to fix the collision automatically.

PR #2745, opened 2019-01-11
[WebUI] Don't use cross-origin request

stanislavlevin:
'Origin' for IPA login page is 'httpS://xxx'. But 'configured' link
has URL like 'http://xxx/ssbrowser.html'.

Since IPA web server doesn't use any kind of Access-Control-Allow-Origin
rules Mozilla Firefox blocks Cross-Origin request due to the Same Origin
policy violation.

So, just follow the Same Origin policy.

Fixes: https://pagure.io/freeipa/issue/7832

### DIFF
--- a/install/ui/src/freeipa/widgets/LoginScreen.js
+++ b/install/ui/src/freeipa/widgets/LoginScreen.js
@@ -60,7 +60,7 @@ define(['dojo/_base/declare',
         kerberos_msg: "<i class=\"fa fa-info-circle\"></i> To log in with " +
             "<strong>Kerberos</strong>, please make sure you" +
             " have valid tickets (obtainable via kinit) and <a href=" +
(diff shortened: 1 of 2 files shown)


stanislavlevin:
Actually, http is always redirected to https.
But your proposal is dynamic-based and I prefer to not use hardcoded values.
Thank you.

stanislavlevin:
@rcritten  thank you for the explanation.

As I see the actual change has been done here: 696fce5c8d4e480c6a731686c8952a4e7ace575f
and it's related to installation of some (no longer used?) firefox extenstion.

stanislavlevin:
Of course, I imply redirection for login page :wink:

serg-cymbaluk:
master:

* e3f37960a00b3f1fb70cd7bdde4daf1a266edc1e Don't use cross-origin request

PR #2746, opened 2019-01-11
Create systemd-user HBAC service and rule

tiran:
authselect changed pam_systemd session from optional to required. When
the HBAC rule allow_all is disabled and replaced with more fine grained
rules, loginsi now to fail, because systemd's user@.service is able to
create a systemd session.

Add systemd-user HBAC service and a HBAC rule that allows systemd-user
to run on all hosts for all users by default. ipa-server-upgrade creates
the service and rule, too. In case the service already exists, no
attempt is made to create the rule. This allows admins to delete the
rule permanently.

### DIFF
--- a/install/share/bootstrap-template.ldif
+++ b/install/share/bootstrap-template.ldif
@@ -346,6 +346,14 @@ cn: sudo-i
 description: sudo-i
 ipauniqueid:autogenerate
 
+dn: cn=systemd-user,cn=hbacservices,cn=hbac,$SUFFIX
+changetype: add
+objectclass: ipahbacservice
+objectclass: ipaobject
+cn: systemd-user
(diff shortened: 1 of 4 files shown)


abbra:
In general, it looks good. I tried to experiment with systemd-user service to exploit it but wasn't able to find a way to do that. So LGTM.

tiran:
I found a bug in my initial implementation: the upgrade didn't specify user/host category ``all``. It's now fixed.

abbra:
We need to backport to ipa-4-6 because newer clients (F29+ and RHEL8) could be enrolled to the older environments.

tiran:
master:

* 2ef6e14c5a87724a3b37dd5f0817af48c4411e03 Create systemd-user HBAC service and rule

PR #2748, opened 2019-01-11
Fix certificate revocation tests for Web UI

serg-cymbaluk:
- correct revocation date before search
- increase timeouts

https://pagure.io/freeipa/issue/7834

### DIFF
--- a/ipatests/test_webui/test_cert.py
+++ b/ipatests/test_webui/test_cert.py
@@ -44,26 +44,6 @@ def search_pkey(self, pkey):
     self.wait_for_request(n=2)
 
 
-def add_cert(self, principal, csr):
-    self.facet_button_click('request_cert')
-    self.fill_textbox('principal', principal)
-    self.check_option('add', 'checked')
-    self.fill_textarea('csr', csr)
-    self.dialog_button_click('issue')
-    self.assert_notification(assert_text='Certificate requested')
-    self.navigate_to_entity(ENTITY)
-    rows = self.get_rows()
-    return rows[-1]
-
-
-def revoke_cert(self, record, reason):
-    self.navigate_to_entity(ENTITY)
-    self.navigate_to_row_record(record)
-    self.action_list_action('revoke_cert', False)
-    self.select('select[name=revocation_reason]', reason)
-    self.dialog_button_click('ok')
-
-
 def check_option_negative(self, date, option):
     self.navigate_to_entity(ENTITY)
     self.select('select[name=search_option]', option)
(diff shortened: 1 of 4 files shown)


ipatests/test_webui/test_service.py
```diff
@@ -218,7 +218,7 @@ def test_certificates(self):
         self.wait()
         self.select('select', '6')
         self.dialog_button_click('ok')
-        self.wait_for_request(n=2, d=3)
+        self.wait_for_request(1, n=3, d=3)
```
pvoborni: Is `wait_for_request` the right call(wait) here? It checks for `.global-activity-indicator` wouldn't it be better to check for the spinner (`.working-widget`) in the cert_widget - as we can see that it is still spinning in the [screenshot](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/c78d72fa-15b0-11e9-8714-fa163ea17f49/2019-01-11T15:24:40.110578_test_service_test_certificates.png)

serg-cymbaluk: Your right. I'll change this wait function.

serg-cymbaluk: @pvoborni
please check if the new method is ok

ipatests/test_webui/ui_driver.py
```diff
@@ -362,6 +362,19 @@ def wait_for_request(self, implicit=0.2, n=1, d=0):
...
+        Wait while working widget active
+        """
+
+        working_widget = self.find('.working-widget', By.CSS_SELECTOR, widget)
+
+        self.wait(implicit)
+        WebDriverWait(self.driver, self.request_timeout).until_not(
+            lambda d: working_widget.is_displayed()
+        )
+        self.wait(0.5)
```
pvoborni: Isn't 0.5s too much? At that point we know that the working widget is gone and thus the parent widget should be re-rendered. To be sure for the re-rendering, small wait like 0.1 can be there but 0.5 seems as unnecessary slowing down of the tests. Or is there a reason for it which I miss?

To be consistent with `wait_for_request`, it can be changed  to `self.wait(d)` and have `d` in method signature.

Otherwise it looks good. Btw, it would be good if somebody else would also look at this PR. I'm not doing any testing or more proper validation.

serg-cymbaluk: With 0.2s it is quite unstable. So I decide set more. It's no so much compared with original timeout 0.2s + 3s (includes AJAX request).
Also this method is used for similar widgets, so there no sense to add parameter until more complicated case comes.

pvoborni: Ok then, makes sense.

ipatests/test_webui/ui_driver.py
```diff
@@ -362,6 +362,19 @@ def wait_for_request(self, implicit=0.2, n=1, d=0):
             self.wait()
         self.wait(d)
 
+    def wait_while_working(self, widget, implicit=0.2):
+        """
+        Wait while working widget active
+        """
+
+        working_widget = self.find('.working-widget', By.CSS_SELECTOR, widget)
```
netoarmando: Could this `wait_while_working()` method be more generic? Something like receiving the CSS selector as an argument? Not sure if this is reusable or not. If it's too specific I believe we are good then.

ipatests/test_webui/test_cert.py
```diff
@@ -162,16 +164,10 @@ def test_search_revocation_reason(self):
...
-        csr = generate_csr(hostname)
-        record = add_cert(self, 'HTTP/{}'.format(hostname), csr)
-
-        # revoke added cert
-        revoke_cert(self, record, '1')
+        # revoke new certificate
+        self._add_and_revoke_cert()
 
         # search cert by revocation reason
-        self.navigate_to_entity(ENTITY)
```
netoarmando: Why this needs to be removed, while similar calls before revoking the certificates were kept intact?

serg-cymbaluk: Yes, calls before should be also moved to the function. I'll fix it.

ipatests/test_webui/ui_driver.py
serg-cymbaluk: There is also different condition for exit the loop. So I wouldn't make over parameterized methods because it will be hard to debug.

ipatests/test_webui/test_cert.py
serg-cymbaluk: @netoarmando 
Fixed, tests have passed.

netoarmando:
Looks good to me, please remove the temp commit.

serg-cymbaluk:
master:

* 88795fb9a930e205ee5ae183e68f4229af0a5a97 Fix certificate revocation tests for Web UI

PR #2751, opened 2019-01-11
[Backport][ipa-4-6] Create systemd-user HBAC service and rule

tiran:
Manual backport of PR #2746 

authselect changed pam_systemd session from optional to required. When
the HBAC rule allow_all is disabled and replaced with more fine grained
rules, loginsi now to fail, because systemd's user@.service is able to
create a systemd session.

Add systemd-user HBAC service and a HBAC rule that allows systemd-user
to run on all hosts for all users by default. ipa-server-upgrade creates
the service and rule, too. In case the service already exists, no
attempt is made to create the rule. This allows admins to delete the
rule permanently.

### DIFF
--- a/install/share/bootstrap-template.ldif
+++ b/install/share/bootstrap-template.ldif
@@ -346,6 +346,14 @@ cn: sudo-i
 description: sudo-i
 ipauniqueid:autogenerate
 
+dn: cn=systemd-user,cn=hbacservices,cn=hbac,$SUFFIX
+changetype: add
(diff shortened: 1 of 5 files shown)


tiran:
ipa-4-6:

* 3b9997348d300e36a96730fb703e96ca46f0f4cb Create systemd-user HBAC service and rule
* 43548d2d735fc90cf3a5a18cc25f43c608312a18 Make conftest compatible with pytest 4.x

PR #2755, opened 2019-01-14
Use expanduser instead of HOME env var

tiran:
The HOME directory may not be available in containers. It's also the
wrong variable on some platforms. Use os.path.expanduser() instead of
HOME.

Also adds a workaround for lib389 bug https://pagure.io/389-ds-base/issue/50152

Fixes: https://pagure.io/freeipa/issue/7837

### DIFF
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -225,7 +225,7 @@ DEFAULT_CONFIG = (
     ('site_packages', object),  # The directory contaning ipalib
     ('script', object),  # sys.argv[0]
     ('bin', object),  # The directory containing the script
-    ('home', object),  # $HOME
+    ('home', object),  # os.path.expanduser('~')
 
     # Vars set in Env._bootstrap():
     ('in_tree', object),  # Whether or not running in-tree (bool)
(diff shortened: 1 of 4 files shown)


xxblx:
Looks good. Patch fixes docker installation of FreeIPA from rpm packages built on recent master commits.

ipalib/constants.py
```diff
@@ -313,13 +313,7 @@
...
 USER_CACHE_PATH = (
     os.environ.get('XDG_CACHE_HOME') or
-    os.path.join(
-        os.environ.get(
-            'HOME',
-            os.path.expanduser('~')
-        ),
-        '.cache'
-    )
+    os.path.expanduser('~/.cache')
```
rcritten: Just one question. In config.py you test that os.path.expanduser('~') doesn't return ~. In the USER_CACHE_PATH case the return value could be the string '~/.cache'. Is that a problem? Should it instead fall back to trying $HOME?

abbra:
Given semantics of `expanduser`, looks good.

tiran:
master:

* 19b1eb1f5a727640b2e6ee2732bc78c0bd5812da Use expanduser instead of HOME env var
* 2ba969da07f1120bc392603ec7bf4bd2a9f967a4 Add workaround for lib389 HOME bug

PR #2759, opened 2019-01-15
Fix systemd-user HBAC rule

tiran:
2ef6e14c5a87724a3b37dd5f0817af48c4411e03 added an invalid HBAC rule that
encoded the service wrongly.


rcritten:
Ok. I'd prefer that the test be in xmlrpc but this fixes the issue in the short-term.

rcritten:
master:

* 965181362a901bb49ee2cfda8aa261a3717213cb Fix systemd-user HBAC rule

PR #2766, opened 2019-01-17
Fix ipatests test_ntp_options to configure the firewall where needed

t-woerner:
The firewall has not been configured before replicas and after masters
have been installed using the command line tools and not the available
methods from tasks. This resulted in failed tests as for example in
test_replica_promotion_without_ntp.

Configuration of the firewall has been added to the methods
install_replica, test_replica_promotion_with_ntp_options and
test_replica_promotion_without_ntp.

In the case of testing command line options that may not be used together
for replica installation, a comment has been added that it is theoretical
no need to configure the firewall in this case, because the installation
will not succeed. But it has been added to make sure that the test will
not fail because of the missing firewall configuration if the options will
not conflict. This is the method test_replica_promotion_with_ntp_options.

In the case where conflicting command line options are tested for the
server installation a comment has been added that it is not needed to
configure the firewall as it is configured after successful server
installation normally. This is the method
test_server_client_install_mixed_options.

tasks.uninstall_master is used to cleanup replica installations and test
remains, therefore the firewall configuration is reverted afterwards again.

See: https://pagure.io/freeipa/issue/7719

### DIFF
--- a/ipatests/test_integration/test_ntp_options.py
+++ b/ipatests/test_integration/test_ntp_options.py
@@ -4,6 +4,7 @@
 
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
+from ipatests.pytest_ipa.integration.firewall import Firewall
 from ipaplatform.paths import paths
 
 
@@ -35,6 +36,8 @@ class TestNTPoptions(IntegrationTest):
         cmd = ['ipa-replica-install', '-w', self.master.config.admin_password,
                '-n', self.master.domain.name, '-r', self.master.domain.realm,
                '--server', self.master.hostname, '-U', *args]
+        Firewall(self.replica).enable_services(["freeipa-ldap",
+                                                "freeipa-ldaps"])
         return self.replica.run_command(cmd, raiseonerr=False)
 
     def test_server_client_install_without_options(self):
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_ntp_options.py
```diff
@@ -175,6 +181,13 @@ def test_replica_promotion_with_ntp_options(self):
         tasks.install_master(self.master, setup_dns=False)
         tasks.install_client(self.master, self.replica)
 
+        # Testing command line options that may not be used together, there
+        # is theoretical no need to configure the firewall in this case,
+        # because the installation will not succeed. But it has been added
+        # to make sure that the test will not fail because of the missing
+        # firewall configuration if the options will not conflict.
+        Firewall(self.replica).enable_services(["freeipa-ldap",
+                                                "freeipa-ldaps"])
```
rcritten: So in other words you want the test to fail because of the conflicting options and not because the firewall wasn't configured?

tiran:
Please rebase and add a temp commit to verify, that this patch fixes the issue.

You aren't disabling the services after the tests. You could use pytest's xunit hooks https://docs.pytest.org/en/2.8.7/xunit_setup.html#method-and-function-level-setup-teardown to disable the firewall afterwards.

ipatests/test_integration/test_ntp_options.py
t-woerner: Yes, just in the case the conflict is not there any more.. then the test would fail because of the missing firewall configuration. But the test should fail because the replica was installed successfully by ignoring the conflict.

flo-renaud:
@t-woerner 
this is just a heads-up in case you missed tiran's [comments](https://github.com/freeipa/freeipa/pull/2766#issuecomment-455492999).

flo-renaud:
@t-woerner 
Is this PR still needed? It seems that test_ntp_options has already been fixed.

t-woerner:
@flo-renaud I think we can close it.

flo-renaud:
Closing this PR as the test_ntp_options are green.

PR #2767, opened 2019-01-17
replica installation: add master record only if in managed zone

flo-renaud:
### replica installation: add master record only if in managed zone

Scenario: install a replica with DNS, whose IP address is part of a forward zone.
Currently, the replica installation fails because the installer is trying to add a A/AAAA record for the replica in the zone when setting up the bind instance, and addition of records in a forward zone is forbidden.

The bind installer should check if the IP address is in a master zone (i.e. a DNS zone managed by IdM, not a forward zone), and avoid creating the record if it's not the case.

During uninstallation, perform the same check before removing the DNS record (if in a forward zone, no need to call dnsrecord-del).

Fixes: https://pagure.io/freeipa/issue/7369

### ipatests: add test for replica in forward zone

Scenario: install a replica with DNS, with the replica part of a forward zone.
The replica installation should proceed successfully and avoid trying to add a DNS record for the replica in the forward zone, as the forward zone is not managed by IPA DNS.

Test added to nightly definitions.

Related to https://pagure.io/freeipa/issue/7369

### DIFF
--- a/ipaserver/install/bindinstance.py
+++ b/ipaserver/install/bindinstance.py
@@ -880,10 +880,13 @@ class BindInstance(service.Service):
 
         # Add forward and reverse records to self
         for addr in addrs:
-            try:
+            # Check first if the zone is a master zone
+            # (if it is a forward zone, dns_zone_exists will return False)
+            if dns_zone_exists(zone, api=self.api):
                 add_fwd_rr(zone, host, addr, self.api)
-            except errors.NotFound:
-                pass
+            else:
+                logger.debug("Skip adding record %s to a zone %s "
+                             "not managed by IPA", addr, zone)
 
             reverse_zone = find_reverse_zone(addr, self.api)
             if reverse_zone:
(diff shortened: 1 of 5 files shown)


fcami:
Reviewing

fcami:
- the proposed fix seems better than the one proposed in https://lists.fedoraproject.org/archives/list/freeipa-users@lists.fedorahosted.org/thread/7A2I475DZFE235QRJRXMRXTL3DVT46IN/ as it does not disable exception checking in add_rr
- it handles the same situation at uninstall time

flo-renaud:
@fcami 
Thanks for the review. I removed the temp commit and tests are green.

fcami:
master:

* 63fa87a36e3026868c49d603762a730948db9643 replica installation: add master record only if in managed zone
* a91e645a147191e48ef939196019c7f4c798eb24 ipatests: add test for replica in forward zone

PR #2770, opened 2019-01-18
[Backport][ipa-4-7] Don't configure KEYRING ccache in containers

tiran:
Manual backport of PR #2677 

environment.


Tiboris:
ipa-4-7:

* b149fff80675b07d280bd0ca8e11a69dc25d0e34 Don't configure KEYRING ccache in containers

PR #2773, opened 2019-01-21
Check if issuer DN is updated after external-ca > self-signed

mrizwan93:
This test checks if issuer DN is updated properly after CA is
renewed back from external-ca to self-signed

related ticket : https://pagure.io/freeipa/issue/7762


flo-renaud:
Hi @mrizwan93 
Thanks for the update, LGTM. Please remove the temp commit and I will ACK.

flo-renaud:
Added the ipa-4-7 label as the fix was also backported to ipa-4-7

flo-renaud:
Tests are green, ACK.

PR #2774, opened 2019-01-21
Split test_webui_hosts PRCI tests

serg-cymbaluk:
Web UI test_host is too heavy and causes timeout errors during night runs, so it is moved to separate configuration.


serg-cymbaluk:
master:

* 76437dc86c1f797ff3ee51250d0ab843f791074b Split test_webui_hosts PRCI tests

PR #2777, opened 2019-01-22
ipatests: check if username are not optimized out in semanage context

mrizwan93:
ipa users having default semanage context were optimized out.
This test checks if those users are listed.

related ticket : https://pagure.io/SSSD/sssd/issue/3819

### DIFF
--- a/ipatests/test_integration/test_user_permissions.py
+++ b/ipatests/test_integration/test_user_permissions.py
@@ -4,11 +4,10 @@
 
 from __future__ import absolute_import
 
-
 from ipaplatform.paths import paths
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
-
+import paramiko
 
 class TestUserPermissions(IntegrationTest):
     topology = 'star'
@@ -67,6 +66,47 @@ class TestUserPermissions(IntegrationTest):
 
         # call ipa user-del --preserve
         self.master.run_command(['ipa', 'user-del', '--preserve', testuser])
+
+    def test_selinux_user_optimized(self):
+        """
+        Check that SELinux login context is set on first login for the
+        user, even if the user is not mapped to a specific SELinux user.
+
+        Related ticket https://pagure.io/SSSD/sssd/issue/3819.
+        """
+        # Scenario: add an IPA user with non-default home dir, login through
+        # ssh as this user and check that there is a SELinux user mapping
+        # for the user with `semanage login -l`.
+
+        # kinit admin
+        tasks.kinit_admin(self.master)
+
+        testuser = 'testuser_selinux'
+        password = 'Secret123'
+        testuser_password_confirmation = "%s\n%s\n" % (password,
+                                                       password)
+        self.master.run_command(['ipa', 'user-add', testuser,
+                                 '--first', testuser,
+                                 '--last', testuser,
+                                 '--password',
+                                 '--homedir',
+                                 '/root/{}'.format(testuser)],
+                                stdin_text=testuser_password_confirmation)
+
+        # login to the system
+        client = paramiko.SSHClient()
+        client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+        client.connect(self.master.hostname,
+                       username=testuser,
+                       password=password)
+        client.close()
+
+        # check if user listed in output
+        cmd = self.master.run_command(['semanage', 'login', '-l'])
+        assert testuser in cmd.stdout_text
+
+        # call ipa user-del
+        self.master.run_command(['ipa', 'user-del', testuser])
 
     def test_stageuser_show_as_alternate_admin(self):
         """


ipatests/test_integration/test_user_permissions.py
```diff
@@ -65,6 +65,13 @@ def test_delete_preserve_as_alternate_admin(self):
         self.master.run_command(['kinit', self.altadmin],
                                 stdin_text=self.master.config.admin_password)
 
+        # check if username in selinux are not optimized out
+        # related ticket https://pagure.io/SSSD/sssd/issue/3819
+        self.master.run_command(['ssh', '{}@{}'.format(self.altadmin,
```
pvoborni: I assume this command is here to have the `semanage` command run as altadmin. I don't think it will work because commands executed by pytests-multihost by default use authentication by ssh key for root user (e.g., look for IPA_ROOT_SSH_KEY ).  If command should run as different user then either the auth part of multihost config needs to be changed  or you can ssh as the user with the right command as param in single call, e.g. https://github.com/freeipa/freeipa/blob/master/ipatests/test_integration/test_legacy_clients.py#L185

On the contrary, I suspect that presence of this command could be the reason for the test timeouts.

In addition, this test is not connected with `test_delete_preserve_as_alternate_admin` in any regard. Adding this test in the middle of the other will make it harder to maintain. It will be better to create a new test method right after `test_delete_preserve_as_alternate_admin` with the intended commands and description.

ipatests/prci_definitions/gating.yaml
```diff
@@ -27,219 +27,14 @@ jobs:
...
-  fedora-29/simple_replication:
+  fedora-29/sssd_selinux:
     requires: [fedora-29/build]
     priority: 50
     job:
       class: RunPytest
       args:
         build_url: '{fedora-29/build_url}'
-        test_suite: test_integration/test_simple_replication.py
+        test_suite: test_integration/test_user_permissions.py::TestUserPermissions::test_delete_preserve_as_alternate_admin
```
flo-renaud: The test name needs to be replaced with `test_selinux_user_optimized`.

ipatests/test_integration/test_user_permissions.py
```diff
@@ -68,6 +68,40 @@ def test_delete_preserve_as_alternate_admin(self):
...
+        self.master.run_command(['ipa', 'user-add', testuser,
+                                 '--first', testuser,
+                                 '--last', testuser,
+                                 '--password',
+                                 '--homedir',
+                                 '/root/{}'.format(testuser)],
+                                stdin_text=testuser_password_confirmation)
+
+        self.master.run_command(
+            'sshpass -p %s '
```
flo-renaud: sshpass may not be installed, you should test first if the file /usr/bin/sshpass is present or skip the test. Please see `ipatests/test_integration/test_legacy_clients.py` for an example.

flo-renaud:
Hi @mrizwan93 
the test was skipped because sshpass is not installed. You need either to work with PR-CI team to have it installed on the test machines or use a different method to login to the machine. Maybe the paramiko.SSHClient() API can be used here.

ipatests/test_integration/test_user_permissions.py
```diff
@@ -68,6 +67,40 @@ def test_delete_preserve_as_alternate_admin(self):
...
+        self.master.run_command(['ipa', 'user-add', testuser,
+                                 '--first', testuser,
+                                 '--last', testuser,
+                                 '--password',
+                                 '--homedir',
+                                 '/root/{}'.format(testuser)],
+                                stdin_text=testuser_password_confirmation)
+
+        # login to the system
+        client = paramiko.SSHClient()
```
flo-renaud: If you add 
```
client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
```
the test will pass.

ipatests/test_integration/test_user_permissions.py
```diff
@@ -68,6 +67,41 @@ def test_delete_preserve_as_alternate_admin(self):
         # call ipa user-del --preserve
         self.master.run_command(['ipa', 'user-del', '--preserve', testuser])
 
+    def test_selinux_user_optimized(self):
+        """
+        check if username in selinux are not optimized out
+        related ticket https://pagure.io/SSSD/sssd/issue/3819
```
flo-renaud: The description is not very explanatory. I would replace with:
```
Check that SELinux login context is set on first login for the user, even if the user is not mapped to a specific SELinux user.
Related ticket https://pagure.io/SSSD/sssd/issue/3819.
Scenario: add a IPA user with non-default home dir, login through ssh as this user and check that there is a SELinux user mapping for the user with `semanage login -l`.
```

flo-renaud:
@mrizwan93 
Thanks for the update, works for me. Tests are green, ACK.

flo-renaud:
master:

* 144a9c74d6f98fcb0f3d82395fe0c9207201b1e9 ipatests: check if username are not optimized out in semanage context

PR #2778, opened 2019-01-22
ipatests: fix ldap server url in ldappasswd_sysaccount_change()

wladich:
master.external_hostname was used to construct ldap server name
which caused ldappasswd utility to exit with error due to host name
mismatch in client certificate. master.hostname should be used instead
as this name is used to generate certificate.

Fixes https://pagure.io/freeipa/issue/7844

Signed-off-by: Sergey Orlov <sorlov@redhat.com>

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -1488,7 +1488,7 @@ def ldappasswd_user_change(user, oldpw, newpw, master):
     basedn = master.domain.basedn
 
     userdn = "uid={},{},{}".format(user, container_user, basedn)
-    master_ldap_uri = "ldap://{}".format(master.external_hostname)
+    master_ldap_uri = "ldap://{}".format(master.hostname)
 
     args = [paths.LDAPPASSWD, '-D', userdn, '-w', oldpw, '-a', oldpw,
             '-s', newpw, '-x', '-ZZ', '-H', master_ldap_uri]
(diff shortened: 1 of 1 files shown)


rcritten:
This variable is also used in ldappasswd_user_change().

It's unclear to me why this change is needed. Did something break? Did this every work?

external_hostname isn't really defined anywhere that I can find. The PR-Ci sample has it defined the same as the hostname. It looks like it is auto-generated using the ipa.test domain if it is not provided.

wladich:
Thanks for pointing usage of external_hostname in ldappasswd_user_change, I added fix also for this function as well.
About the reason for fixing it -- as I pointed in https://pagure.io/freeipa/issue/7844, tests are failing when "name" and "external_hostname" are not equal in test configuration. The configuration itself is described in https://www.freeipa.org/page/Integration_testing_configuration#Using_YAML.2FJSON_configuration_file

rcritten:
Right but that config file just has an example, it doesn't explain why a hostname and external hostname are needed.

I'm just curious what changed that caused the test to start failing.

wladich:
I think this test was always failing when name and external_hostname did not match.

rcritten:
I dug in the code a little and it seems like external_hostname is related to testing in certain environments, perhaps outside of PR-CI. I don't mean to block this one change due to pre-existing missing docs I'm just trying to understand under what conditions the external name is important. Is it the sort of AWS case where a host knows its own name but is know as something else externally?

I, perhaps naively, do not have external_hostname configured in my test yaml. I wonder what happens in this case, does the test blow up, fall back to hostname, something else? It appears that it is an expected value in the config and I guess I've just been lucky so far.

wladich:
As per my understanding "external_hostname"  is used only to ssh from controller machine to test machine, it is defined by DNS server of testing infrastructure. And "name" is a hostname known to ipa components, it is defined by DNS which is part of IPA system or by some other server if IPA master does not have one.
Those names can be different in some testing environments.
For example, when we provision VMs for tests they get names like host-45-34.inetrnal.our-company.com but for tests we define name like  "client1.ipa.test" and latter name is managed by DNS of IPA server installation.
external_hostname can be missing in config and be replaced with ip_address.

flo-renaud:
Hi @wladich 
Thanks for the PR. You are right, PRCI does not detect the problem because the external_hostname and hostname are identical but you found a real test issue, which is properly fixed by your PR.

tiran:
master:

* 20d8286b60b6f3f597912d16afd04ef29ad248c8 ipatests: fix ldap server url

PR #2785, opened 2019-01-25
Add workaround for slow host/service del

tiran:
host-del and service-del are slow because cert revokation is implemented
inefficiently. The internal cert_find() call retrieves all certificates
from Dogtag.

The workaround special cases service and host find without additional RA
search options. A search for service and host certs limits the scope to
certificate with matching subject common name.
 
See: https://pagure.io/freeipa/issue/7835
See: https://bugzilla.redhat.com/show_bug.cgi?id=1669012

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -1466,6 +1466,22 @@ class cert_find(Search, CertMethod):
         result = collections.OrderedDict()
         complete = bool(ra_options)
 
+        # workaround for RHBZ#1669012
+        # Improve performance for service and host case by also searching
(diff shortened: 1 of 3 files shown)


abbra:
LGTM.

abbra:
There are existing tests for host-del/service-del related to certificates, thus ACK.

tiran:
master:

* ae74d348c3da580264c56441c136af3fc6ae58df Add workaround for slow host/service del
* 6cd37542e6c563ddd8595d488712935a4d6b17bf Optimize cert remove case

PR #2788, opened 2019-01-29
[Backport][ipa-4-7] Use expanduser instead of HOME env var

tiran:
Manual backport of first commit from PR #2755 

The HOME directory may not be available in containers. It's also the
wrong variable on some platforms. Use os.path.expanduser() instead of
HOME.

Fixes: https://pagure.io/freeipa/issue/7837


rcritten:
ipa-4-7:

* b8522da3606c70230fd0156a357c991820fcede8 Use expanduser instead of HOME env var

PR #2798, opened 2019-01-30
tests: Don't provide explicit hostname to ldapmodify

rcritten:
Manual revert of bbac233b5ee487ab0e035cf0b861144769a0b738

The assumption was that ldap.conf was hosed and it couldn't
tell what hostname to use so one was hardcoded. This code
doesn't explicitly test that ldap.conf is sane but it is
a nice side-effect I suppose.

https://pagure.io/freeipa/issue/5880

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -306,7 +306,7 @@ def enable_replication_debugging(host, log_level=0):
     host.run_command(['ldapmodify', '-x',
                       '-D', str(host.config.dirman_dn),
                       '-w', host.config.dirman_password,
-                      '-h', host.hostname],
+                      ],
                      stdin_text=logging_ldif)
 
 


rcritten:
So I guess green means good in this case.

flo-renaud:
LGTM.

flo-renaud:
@rcritten 
Ticket [5580](https://pagure.io/freeipa/issue/5880) has the milestone " FreeIPA 4.5 backlog". Do you think a backport to ipa-4-7 is required or is master sufficient?

rcritten:
Oh good question, I didn't think of that when I dropped the WIP flag. I think 4-7 would be good, I'll add the flag. By relying on ldap.conf we do some indirect testing that we are managing that file correctly.

rcritten:
master:

* fd4b84d6d23227a30a83086b6ad0b6966fcba9f3 tests: Don't provide explicit hostname to ldapmodify

PR #2803, opened 2019-02-01
[Backport][ipa-4-7] Check if issuer DN is updated after external-ca > self-signed

flo-renaud:
This PR was opened automatically because PR #2773 was pushed to master and backport to ipa-4-7 is required.


rcritten:
ipa-4-7:

* fde6ef5bbf579d605fddc0c2e33e2629a202eb8b Check if issuer DN is updated after external-ca > self-signed

PR #2805, opened 2019-02-01
Update mod_nss cipher list so there is overlap with a 4.x master

rcritten:
dogtag updated its cipher list, disabling a lot of ciphers, which
causes an overlap problem with a RHEL 6.x IPA master.

This update script adds the two available ciphers to the nss.conf
so that creating a CA replica is possible.

### DIFF
--- a/contrib/copy-schema-to-ca-RHEL6.py
+++ b/contrib/copy-schema-to-ca-RHEL6.py
@@ -31,6 +31,12 @@ from ipaserver.install.dsinstance import DS_USER
 from ipaserver.install.cainstance import PKI_USER
 from ipapython import services
 
+# for mod_nss
+from ipaserver.install.httpinstance import NSS_CONF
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @rcritten 
Thanks for the patch. I did the following test:
- copy the script on a RHEL 6.8 master
- [rhel6.8]$ python copy-schema-to-ca-RHEL6.py
- [rhel6.8]$ ipa-replica-prepare rhel7.example.com --ip-address 10....
- [rhel7.6]$ ipa-replica-install --setup-ca /path/to/replica_file.gpg
and the test succeeded.

I will have a closer look at the code monday morning but so far it looks good to me.

flo-renaud:
master:

* ed74b89878ccf62d22b9188c6f57adab65b66c2b Update mod_nss cipher list so there is overlap with a 4.x master

PR #2812, opened 2019-02-05
Require secure-binds for password login

tiran:
nsslapd-require-secure-binds restricts password based simple binds to
secure connections. It does not prevent a careless user from
transmitting a password in plain text. But it makes it obvious that he
did something bad. Password based bind attempts over an insecure
connections are refused with:

  Confidentiality required: Operation requires a secure connection

Secure connections are:

* LDAP connections on port 389 with STARTTLS
* LDAPS connections in port 636
* LDAPI connections to a local Unix sockets

Anonymous bind (simple_bind with empty DN and password) and GSSAPI
bind operations are not affected.

nsslapd-require-secure-binds is enabled after 389-DS is configured for
TLS/SSL.

**NOTE** The change may cause compatibility issues with applications that don't perform secure binds.

### DIFF
--- a/install/share/Makefile.am
+++ b/install/share/Makefile.am
@@ -81,6 +81,7 @@ dist_app_DATA =				\
 	replica-automember.ldif		\
 	sasl-mapping-fallback.ldif	\
 	schema-update.ldif		\
+	secure-binds.ldif		\
 	vault.ldif			\
 	kdcproxy-enable.uldif		\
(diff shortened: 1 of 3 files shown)


rcritten:
This would not affect upgrades and could result in a mix of security between old and new masters.

tiran:
The change breaks backwards compatibility. Let's handle this in a dedicated hardening patch.

PR #2813, opened 2019-02-05
Require a minimum SASL security factor of 56

tiran:
SSF_MINX 56 level ensures data integrity and confidentiality for SASL
GSSAPI and SASL GSS SPNEGO connections.

Although at least AES128 is enforced pretty much everywhere, 56 is required
for backwards compatibility with systems that announce wrong SSF.

### DIFF
--- a/install/share/min-ssf.ldif
+++ b/install/share/min-ssf.ldif
@@ -1,0 +1,14 @@
+# config
+# pretend SSF for LDAPI connections
+# nsslapd-localssf must be equal to or greater than nsslapd-minssf
+dn: cn=config
+changetype: modify
+replace: nsslapd-localssf
+nsslapd-localssf: 256
+
+# minimum security strength factor for SASL and TLS
+# 56 is considered weak, but some old clients announce wrong SSF.
+dn: cn=config
+changetype: modify
+replace: nsslapd-minssf
+nsslapd-minssf: 56
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -43,7 +43,9 @@ import six
 
 # pylint: disable=ipa-forbidden-import
 from ipalib import errors, x509, _
-from ipalib.constants import LDAP_GENERALIZED_TIME_FORMAT
+from ipalib.constants import (
+    LDAP_GENERALIZED_TIME_FORMAT, LDAP_SSF_MIN_THRESHOLD
+)
 # pylint: enable=ipa-forbidden-import
 from ipaplatform.paths import paths
 from ipapython.ipautil import format_netloc, CIDict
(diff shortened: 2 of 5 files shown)


rcritten:
This probably resolves one or both of https://pagure.io/freeipa/issue/7140 and https://pagure.io/freeipa/issue/4580

install/share/min-ssf.ldif
```diff
@@ -0,0 +1,14 @@
+# config
+# pretended SSF for LDAPI connections
```
rcritten: nit: pretended -> pretend

rcritten:
This otherwise LGTM.

rcritten:
Upon reflection. given that this doesn't affect upgrades, it would mean that different masters would be configured with different security levels which might be undesirable and cause unpredictable results.

rcritten:
I have similar concerns to ending up with a mix of config with no upgrades I have in https://github.com/freeipa/freeipa/pull/2813

Otherwise it seems to work ok. I was able to use a RHEL 6.10 (so 3.0.0) to connect and work ok with the master. I tested the ipa tool and ldapsearch -x and ldapsearch -Y GSSAPI.

ipapython/ipaldap.py
```diff
@@ -103,7 +103,7 @@ def realm_to_ldapi_uri(realm_name):
     return 'ldapi://' + ldapurl.ldapUrlEscape(socketname)
 
 
-def ldap_initialize(uri, cacertfile=None):
+def ldap_initialize(uri, cacertfile=None, ssf_min_threshold=56):
```
Tiboris: Hi @tiran, 

Just asking, could we use ipalib/constants.py to contain this minimal threshold value?

ipapython/ipaldap.py
```diff
@@ -114,13 +114,23 @@ def ldap_initialize(uri, cacertfile=None):
...
 
     # Do not perform reverse DNS lookups to canonicalize SASL host names
     conn.set_option(ldap.OPT_X_SASL_NOCANON, ldap.OPT_ON)
 
     if not uri.startswith('ldapi://'):
+        # require a minimum SSF for TCP connections, but don't lower SSF_MIN
+        # if the current value is already larger.
+        cur_min_ssf = conn.get_option(ldap.OPT_X_SASL_SSF_MIN)
+        if cur_min_ssf < ssf_min_threshold:
+            conn.set_option(ldap.OPT_X_SASL_SSF_MIN, ssf_min_threshold)
```
Tiboris: Maybe not needed, but have you considered to log this behavior of ssf change?

ipapython/ipaldap.py
tiran: Done

ipapython/ipaldap.py
tiran: At the moment the default SSF is 0. This will always trigger.

Tiboris: Aha, Okay.

Tiboris:
master:

* 350954589774499d99bf87cb5631c664bb0707c4 Require a minimum SASL security factor of 56

PR #2814, opened 2019-02-05
Tests: fix option name for dsctl

flo-renaud:
389-ds-base has modified one option name in dsctl, and our test `test_uninstallation.py::TestUninstallBase::()::test_failed_uninstall` is still using the old option (--doit) instead of the new one (--do-it).

Fixes: https://pagure.io/freeipa/issue/7856

### DIFF
--- a/ipatests/test_integration/test_uninstallation.py
+++ b/ipatests/test_integration/test_uninstallation.py
@@ -115,5 +115,5 @@ class TestUninstallBase(IntegrationTest):
             ])
 
             self.master.run_command([
-                paths.DSCTL, serverid, 'remove', '--doit'
+                paths.DSCTL, serverid, 'remove', '--do-it'
             ])


flo-renaud:
For future reference, the commit in 389-ds is the following: [4fd73c5](https://pagure.io/389-ds-base/c/4fd73c5d9a431dec0ac8189d6dee6e592088e3c6)

rcritten:
LGTM, remove test commit and I'll ack.

flo-renaud:
@rcritten 
thanks for the review. Temp commit removed.

rcritten:
master:

* 0376d7043c143fda14244d9d17dad2df043f80c9 Tests: fix option name for dsctl

rcritten:
A manual backport is needed for ipa-4-7. It may be worthwhile to include the same temporary commit to ensure that the basics continue to work.

flo-renaud:
Hi @rcritten,
IIRC the use of lib389 has been introduced in the master branch only (see ticket #[4491](https://pagure.io/freeipa/issue/4491)), meaning we don't need a backport to ipa-4-7. Can you confirm?

rcritten:
This confused me as well. The ipa-4-7 spec file has a requirement on lib389 but it isn't invoked by us anywhere. It sure does seem that a backport is not required though given that. The dependency may be for other reasons. So yeah, no backport needed.

flo-renaud:
Ok, thanks for the confirmation. I will close 7856 as fixed in this case.

PR #2816, opened 2019-02-06
Compile IPA modules with C11 extensions

tiran:
- define ``__STDC_WANT_LIB_EXT1__`` to get C11 extensions like ``memset_s()`` for
  Samba's ``ZERO_STRUCT()`` macro, see
  https://en.cppreference.com/w/c/string/byte/memset
- ``_DEFAULT_SOURCE`` enables features like htole16() from endian.h, see
  http://man7.org/linux/man-pages/man3/endian.3.html
- ``_POSIX_C_SOURCE >= 200809`` enables features like ``strndup()`` from ``string.h``,
  see http://man7.org/linux/man-pages/man3/strndup.3.html
- ``time_t`` is no longer implicitly defined, include ``time.h``
- ``typeof()`` is only available as GNU extension. Use explicit types
  instead of generic ``__typeof__()``.

Fixes: https://pagure.io/freeipa/issue/7858

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -18,14 +18,20 @@ AC_CONFIG_HEADERS([config.h])
 AM_INIT_AUTOMAKE([foreign 1.9 tar-pax])
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES])
 
+dnl enable C11 extensions for features like memset_s()
+CFLAGS+=" -D__STDC_WANT_LIB_EXT1__=1"
+dnl enable features like htole16()
+CFLAGS+=" -D_DEFAULT_SOURCE=1"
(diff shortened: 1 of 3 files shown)


tiran:
I haven't tested ``__STDC_WANT_LIB_EXT1__`` on a F30 machine, yet.

abbra:
LGTM. It still does not compile on F30 due to missing memset_s() (too new!) but as we found in the PR#2820, we don't need that ZERO_STRUCT at all.

tiran:
master:

* 93fb037d8409d9d46606c31d8a240e3963b72651 Compile IPA modules with C11 extensions

PR #2820, opened 2019-02-07
Remove ZERO_STRUCT() call

tiran:
ipa_sam uses Samba's macro ZERO_STRUCT() to safely zero out a block in
memory. On F30 ZERO_STRUCT() is currently broken, because it uses the
undefined C11 function memset_s().

During investigation of the bug, it turned out that
ZERO_STRUCT(td->security_identifier) is not needed. The whole td struct
is allocated with talloc_zero(), so td->security_identifier is already
zeroed.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1672231


abbra:
LGTM.

tiran:
master:

* 272837f1c07729392cdbc88b99a221390d01e70d Remove ZERO_STRUCT() call

PR #2825, opened 2019-02-08
Test --external-ca-type=ms-cs

tiran:
Verify that ipa-server-install with external CA and CA type ms-cs adds
the correct extension to the CSR.

Fixes: https://pagure.io/freeipa/issue/7548

### DIFF
--- a/ipatests/test_integration/test_external_ca.py
+++ b/ipatests/test_integration/test_external_ca.py
@@ -22,12 +22,16 @@ import re
 import time
 import tempfile
 
+from cryptography import x509
+from cryptography.hazmat.backends import default_backend
+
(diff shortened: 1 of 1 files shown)


frasertweedale:
I'm happy for this to go as-is.  If we want to test everything properly there are a bunch of other scenarios (which the pagure ticket alludes to: https://pagure.io/freeipa/issue/7548).  So I think we should merge this, but not close the ticket yet.  I'll find some time this week to implement tests for the rest of the scenarios.

tiran:
Sounds good to me!

tiran:
master:

* 5c8c00a4b759b084b0c3bbbea2193af56d8c867a Test --external-ca-type=ms-cs

PR #2827, opened 2019-02-11
ipatests: add test for correct modlist when value encoding differs

wladich:
See: https://pagure.io/freeipa/issue/7750


abbra:
LGTM

wladich:
master:

* 425dff1e0d2b360557075ea322309411b485d58f ipatests: add test for correct modlist when value encoding differs

PR #2828, opened 2019-02-11
Disable dependency on dogtag-pki PyPI package

tiran:
The dependency on 'dogtag-pki' PyPI package causes problems.

For one it's not the full pki package. It only provides the client part,
but ipaserver also needs the pki.server subpackage with pkispawn command.

The Fedora package dependency generator turns the requirement into a
package requirement, but python3-pki does not provide the package name
python3.7dist(dogtag-pki).

### DIFF
--- a/ipaserver/setup.py
+++ b/ipaserver/setup.py
@@ -48,7 +48,9 @@ if __name__ == '__main__':
             "custodia",
             "dbus-python",
             "dnspython",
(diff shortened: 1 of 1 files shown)


tiran:
master:

* 3ec29a7f6a28a98776845d0ec561b633d07896b1 Disable dependency on dogtag-pki PyPI package

PR #2838, opened 2019-02-14
Web UI: Increase timeouts for UI tests in Nightly PR configuration

serg-cymbaluk:
Some test suites for WebUI in Nightly PR configuration have timeouts without any reserve.
So these tests fails randomly.

Timeout values for these test was increased to {real duration} + ~30%

https://pagure.io/freeipa/issue/7864


pvoborni:
The patch looks good to me. I wouldn't be afraid of using 5400 (1.5h) everywhere as it was a design aspect of the CI.  Larger values also give us space to add more new tests in existing suites.

serg-cymbaluk:
master:

* 5e7f82db3c102ab90c24f19786e4254c09d89a1c Web UI: Increase timeouts for UI tests in Nightly PR configuration

PR #2839, opened 2019-02-14
tests: mark xfail for test_selinux_user_optimized on fed<=28

flo-renaud:
The test TestUserPermissions::test_selinux_user_optimized is
testing the fix for SSSD issue 3819, but the fix is not
available in fedora 28. Hence mark the test as xfail when
executed on fedora <=28 (our nightly tests also run on fed 28).

For full ref: fixed in sssd 1.16.4, Fedora 28 provides
1.16.3-2.fc28 only, while Fedora 29 provides 2.0.0-3.fc29.

related ticket : https://pagure.io/SSSD/sssd/issue/3819

### DIFF
--- a/ipatests/test_integration/test_user_permissions.py
+++ b/ipatests/test_integration/test_user_permissions.py
@@ -4,10 +4,13 @@
 
 from __future__ import absolute_import
 
+import paramiko
+import pytest
+
+from ipaplatform.osinfo import osinfo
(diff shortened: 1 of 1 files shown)


flo-renaud:
master:

* e63c6b202a7aa7b2df40cbffe40e673e7d08f3ed tests: mark xfail for test_selinux_user_optimized on fed<=28

PR #2841, opened 2019-02-15
tests: fix failure in test_topology_TestTopologyOptions:test_add_remove_segment

flo-renaud:
The test is performing topology changes on the master, then waits for replication to replicate the changes and checks the expected outcome on replica1.

The issue is that wait_for_replication was called on replica1, but should be called on the master. This method is reliable only if it is executed on the host where the modification was done.

Fixes https://pagure.io/freeipa/issue/7865


fcami:
Will review when CI finishes

flo-renaud:
Hi @fcami 
thanks for the review. Temp commit removed.

fcami:
master:

* cb14883acc94c20c6330f34890998535ce29934a tests: fix failure in test_topology_TestTopologyOptions:test_add_remove_segment

PR #2843, opened 2019-02-18
Chown pki-tomcat's deployment.cfg to pkiuser after writing it.

fcami:
This fixes ipa-server-install when fs.protected_regular=1
Fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=1677027

Fixes: https://pagure.io/freeipa/issue/7866

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -483,7 +483,6 @@ class CAInstance(DogtagInstance):
         (cfg_fd, cfg_file) = tempfile.mkstemp()
         os.close(cfg_fd)
         pent = pwd.getpwnam(self.service_user)
-        os.chown(cfg_file, pent.pw_uid, pent.pw_gid)
 
         # Create CA configuration
         config = RawConfigParser()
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @fcami 
thanks for the patch. I would replace the commit comment with something more descriptive, for instance:
```
/tmp is a sticky directory. When the OS is configured with fs.protected_regular=1, this means that O_CREATE open is forbidden for files in /tmp if the calling user is not owner of the file, except if the file is owned by owner of the directory.

The installer (executed as root) creates a file in /tmp, then modifies its owner to pkiuser and finally writes the config in the file. With fs.protected_regular=1, the write is denied because root is not owner of the file at this point.
The fix performs the ownership change after the file has been written.
```

fcami:
Hi @flo
Thanks for the suggestion, I've switched to your commit message.

fcami:
master:

* 5525322817c736d8851661a5bfedfdd5c794e5c8 ipa-server-install: fix ca setup when fs.protected_regular=1

PR #2846, opened 2019-02-19
ipa_sam: remove dependency to talloc_strackframe.h

sumit-bose:
Recent Samba versions removed some header files which did include
non-public APIs. As a result talloc_strackframe.h and memory.h (for
SAFE_FREE) are not available anymore. This patch replaces the use of the
non-public APIs with public ones.

### DIFF
--- a/daemons/ipa-sam/ipa_sam.c
+++ b/daemons/ipa-sam/ipa_sam.c
@@ -19,7 +19,6 @@
 #include <util/data_blob.h>
 #include <util/time.h>
 #include <util/debug.h>
-#include <util/talloc_stack.h>
 
 #ifndef _SAMBA_UTIL_H_
 bool trim_string(char *s, const char *front, const char *back);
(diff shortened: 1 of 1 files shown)


abbra:
Please use long-term ipa context (`priv` in the functions affected) instead of `talloc(NULL, ..)`

sumit-bose:
> Please use long-term ipa context (`priv` in the functions affected) instead of `talloc(NULL, ..)`

Thanks, the latest version uses priv.

fcami:
LGTM, waiting for CI

rcritten:
I'm not sure why @abbra asked to use priv instead of NULL. This is a temporary context anyway right?

fcami:
master:

* d1f5ed64e16d65b9df45cc0eac7d2724dcae7b67 ipa_sam: remove dependency to talloc_strackframe.h

abbra:
The pull request is merged, @rcritten, but I'll answer here anyway. ;) We try to avoid using NULL context on talloc because internally talloc library is not thread-safe. There are nuances on using NULL context in multithreaded applications, like covered in this talloc tutorial: https://talloc.samba.org/talloc/doc/html/libtalloc__threads.html

Our practical issue here is that ipasam module is loaded into any Samba process that initializes PASSDB API. It means we cannot guarantee a particular thread state and where NULL context was initialized. We have, however, our own private context that is created off the talloc context our whole pdb_method is tied to. This context is create per each module initialization and thus is guaranteed to be unique.

PR #2847, opened 2019-02-19
pkinit setup: fix regression on master install

flo-renaud:
## pkinit setup: fix regression on master install
The commit 7785210 intended to fix ipa-pkinit-manage enable on a replica without any CA but introduced a regression: ipa-server-install fails to configure pkinit with the fix.

This commit provides a proper fix without the regression: pkinit needs to contact Dogtag directly only in case there is no CA instance yet (for ex. because we are installing the first master).

Fixes: https://pagure.io/freeipa/issue/7795

## test: add non-reg test checking pkinit after server install
Add a test with the following scenario:
ipa-server-install (with ca and pkinit enabled)
check that pkinit is properly enabled:
ipa-pkinit-manage status must return "enabled"
the KDC cert must be signed by IPA CA

Related to: https://pagure.io/freeipa/issue/7795

### DIFF
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -429,13 +429,14 @@ class KrbInstance(service.Service):
             prev_helper = None
             # on the first CA-ful master without '--no-pkinit', we issue the
             # certificate by contacting Dogtag directly
-            localhost_has_ca = self.fqdn in find_providing_servers(
+            ca_instances = find_providing_servers(
                 'CA', conn=self.api.Backend.ldap2, api=self.api)
+
             use_dogtag_submit = all(
                 [self.master_fqdn is None,
                  self.pkcs12_info is None,
                  self.config_pkinit,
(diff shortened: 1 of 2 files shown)


flo-renaud:
Hi @fcami 
since you reviewed the first fix, could you have a look at this one?

fcami:
hi @flo-renaud yes of course, doin' so now.

flo-renaud:
@fcami 
thanks, I just pushed a temp commit to show the behavior of the new test.

fcami:
@flo-renaud could you please remove the temp commit? I'll merge when CI is done.

flo-renaud:
Hi @fcami 

fcami:
master:

* 658983794ccfd56afeb376a2f2d7d9a382d25a0c pkinit setup: fix regression on master install
* a25de958e4819d344d15964c7344b2a9ba15c204 test: add non-reg test checking pkinit after server install

PR #2852, opened 2019-02-20
Add more tests for --external-ca-profile handling

frasertweedale:
Add tests for remaining untested scenarios of --external-ca-profile
handling in ipa-server-install.

ipa-ca-install and ipa-cacert-manage remain untested at present.

Fixes: https://pagure.io/freeipa/issue/7548


flo-renaud:
Hi @frasertweedale 
Thanks for the PR. I have 2 comments:
- the new tests are neither part of the gating tests nor of the nightly tests. You need to add them in `ipatests/prci_definitions/*.yaml` (as they should be quick, I propose to include them in all the test definitions. The current rule is that any test in gating must also be in the nightlies).
- it would be nice to factorize some code, for instance:
```
def check_install_step1(host, profile, success=True, msg=None):
         result = install_server_external_ca_step1(
            host,
            extra_args=[
                '--external-ca-type=ms-cs',
                '--external-ca-profile={}'.format(profile),
            ]
        )
        if success:
            assert result.returncode == 0
        else:
            assert result.returncode != 0
            if msg:
                assert msg in result.stderr_text
```
and call this function from the tests.

frasertweedale:
Thanks @flo-renaud.  Now they are running, I have to fix my broken tests :D

ipatests/test_integration/test_external_ca.py
```diff
@@ -411,3 +411,93 @@ def test_master_install_ca2(self):
...
+        master,
+        extra_args=['--external-ca-type=ms-cs', f'--external-ca-profile={s}'],
+        raiseonerr=False,
+    )
+
+
+def _test_invalid_profile(master, profile):
+    result = _step1_profile(master, profile)
+    assert result.returncode != 0
+    assert '--external-ca-profile' in result.stderr_text
```
rcritten: Wouldn't it be better to expect certain output? In at least the case of "NotAnOid:1" it returns a non-localized error message so comparing all or part of that would be more complete.

frasertweedale: It tests that the output refers to the `--external-ca-profile` option.  This is good enough IMO.  Otherwise tests become more brittle in the face of trivial string changes or when localisation is implemented.

frasertweedale:
Resurrecting this PR.  Had to move a bunch of classes out of `ipaserver` and into `ipalib` along the way, to handle the fact that `ipatests` can no longer import `ipaserver`.  Expect CI to fail (and I am logging off for the day) so putting the WIP label.

frasertweedale:
Removing `ipa-4-7` tag and adding `ipa-4-8`.  This PR now includes fixes to regressions that are only present in `master` and `ipa-4-8`.

frasertweedale:
@tiran found a couple of bugs / regressions related to the pkispawn template changes.  There are separate commits in this PR (it should be clear which ones).

frasertweedale:
Note: we can drop the commit that adds it to gating and leave it as nightly-only, if folks feel that is best.  But since there were actually regressions caught by these tests, maybe it is best to keep it in gating.

ipatests/prci_definitions/nightly_master.yaml
```diff
@@ -75,6 +75,18 @@ jobs:
...
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-30/build_url}'
+        test_suite: test_integration/test_external_ca.py::TestExternalCAProfileScenarios
+        template: *ci-master-f30
+        timeout: 3600
+        topology: *master_1repl
+
```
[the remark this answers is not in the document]
frasertweedale: @flo-renaud no worries, PR updated.  I included `nightly_master_pki` too, because this feature does depend on some less-frequently-used `pki_default.cfg(5)` knobs.

flo-renaud:
@frasertweedale 
thanks for the updated patch, works for me.

abbra:
master:

* 130e1dc3433977f7a80aed139d29d21c5d30d558 move MSCSTemplate classes to ipalib
* 21a9a7107a2028354c0fc15540b8f15d279e5fce install: fix --external-ca-profile option
* 7171142aaf91d6079798d015af9862d6e5474c8c Fix use of incorrect variable
* b15bd50e6db919279e25a8bed796df4836b845a8 Add more tests for --external-ca-profile handling
* 80e76f094c234710ff58917e7ce4661d823c4de7 Collapse --external-ca-profile tests into single class
* 2c8352fe8536be0e630fdb910dc90831847ad119 ci: add --external-ca-profile tests to nightly
* 33f39d88bf0bb871e3868c827aeb335eb40f5ae3 ci: add --external-ca-profile tests to gating

PR #2857, opened 2019-02-22
Fix installation when CA subject DN has escapes

frasertweedale:
There were several bugs across several projects preventing
installation when the CA subject DN contains characters that need
escaping in the string representation, e.g.

    CN=Certificate Authority,O=Acme\, Inc.,ST=Massachusetts,C=US

The package versions containing relevant fixes are:

- 389-ds-base 1.4.0.20 (we already require >= 1.4.0.21)
- pki-core 10.5.5 (we already require >= 10.6.8)
- certmonger 0.79.7 (this commit bumps the dependency)

With this change, installation will now work.  Integration tests are
left for a subsequent commit.

Fixes: https://pagure.io/freeipa/issue/7347

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -91,6 +91,9 @@
 # and https://pagure.io/dogtagpki/issue/3073
 %global pki_version 10.6.8-3
 
+# https://pagure.io/certmonger/issue/90
+%global certmonger_version 0.79.7-1
+
 # NSS release with fix for p11-kit-proxy issue, affects F28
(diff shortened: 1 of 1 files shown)


frasertweedale:
Hm, looks like we have to wait until `certmonger-0.79.7-1` to hit `updates`.

tiran:
master:

* 98b7fbec5fa56e6904c40cc22e484cc1f47891b2 Fix installation when CA subject DN has escapes

PR #2858, opened 2019-02-22
[Backport][ipa-4-6] Processing of server roles should ignore errors.EmptyResult

rcritten:
MANUAL BACKPORT.

This was not backported when the original PR was pushed. The milestone was pushed back afterward.

ipa-4-6 still defines:
WSGI_PROCESSES=constants.WSGI_PROCESSES,

---------
When non-admin user issues a command that utilizes
api.Object.config.show_servroles_attributes(), some server roles might
return errors.EmptyResult, indicating that a role is not visible to this
identity.

Most of the callers to api.Object.config.show_servroles_attributes() do
not process errors.EmptyResult so it goes up to an API caller. In case
of Web UI it breaks retrieval of the initial configuration due to ipa
config-show failing completely rather than avoiding to show available
server roles.

Fixes: https://pagure.io/freeipa/issue/7452
Signed-off-by: Alexander Bokovoy abokovoy@redhat.com

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -1782,6 +1782,8 @@ def upgrade_configuration():
         DOGTAG_PORT=8009,
         CLONE='#',
         WSGI_PROCESSES=constants.WSGI_PROCESSES,
+        IPA_CCACHES=paths.IPA_CCACHES,
+        IPA_CUSTODIA_SOCKET=paths.IPA_CUSTODIA_SOCKET
     )
 
     subject_base = find_subject_base()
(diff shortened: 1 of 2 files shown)


flo-renaud:
Hi @rcritten 
I was not able to reproduce the issue on ipa-4-6 branch, but the fix is harmless so ACk from me.

rcritten:
ipa-4-6:

* 56e55e040bad1d4485ad13dc28dd0537009078b5 Processing of server roles should ignore errors.EmptyResult
* 0ad97028167705d44e9e64b43b312c516fbff7de Update template directory with new variables when upgrading ipa.conf.template

PR #2872, opened 2019-03-01
too-restrictive mask checks

fcami:
ipatests: add too-restritive mask tests
ipa-{server,replica}-install: add too-restritive mask detection

If the mask used during the installation is "too restrictive", ie.0027,
installing FreeIPA results in a broken server or replica.
Check for too-restrictive mask at install time and error out.
    
Fixes: https://pagure.io/freeipa/issue/7193

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -1389,3 +1389,14 @@ def default_subject_base(realm_name):
 
 def default_ca_subject_dn(subject_base):
     return DN(('CN', 'Certificate Authority'), subject_base)
+
+
+def validate_mask():
+    try:
+        mask = os.umask(0)
+    finally:
+        os.umask(mask)
+    mask_str = None
+    if mask & 0b111101101 > 0:
+        mask_str = "{:04o}".format(mask)
+    return mask_str
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -567,3 +567,49 @@ class TestKRAinstallAfterCertRenew(IntegrationTest):
         self.master.run_command(['kinit', 'admin'], stdin_text=passwd)
         cmd = self.master.run_command(['ipa-kra-install', '-p', dm_pass, '-U'])
         self.master.run_command(['systemctl', 'start', 'chronyd'])
+
+
+class TestMaskInstall(IntegrationTest):
+    """ Test master and replica installation with wrong mask
+
+    This test checks that master/replica installation fails (expectedly) if
+    mask > 022.
+
+    related ticket: https://pagure.io/freeipa/issue/7193
+    """
+
+    num_replicas = 0
+
+    @classmethod
+    def install(cls, mh):
+        super(TestMaskInstall, cls).install(mh)
+        cls.bashrc_file = cls.master.get_file_contents('/root/.bashrc')
+
+    def test_install_master(self):
+        self.master.run_command('echo "umask 0027" >> /root/.bashrc')
+        result = self.master.run_command(['umask'])
+        assert '0027' in result.stdout_text
+
+        cmd = tasks.install_master(
+            self.master, setup_dns=False, raiseonerr=False
+        )
+        exp_str = ("Unexpected system mask")
+        assert (exp_str in cmd.stderr_text and cmd.returncode != 0)
+
+    def test_install_replica(self):
+        result = self.master.run_command(['umask'])
+        assert '0027' in result.stdout_text
+
+        cmd = self.master.run_command([
+            'ipa-replica-install', '-w', self.master.config.admin_password,
+            '-n', self.master.domain.name, '-r', self.master.domain.realm,
+            '--server', 'dummy_master.%s' % self.master.domain.name,
+            '-U'], raiseonerr=False
+        )
+        exp_str = ("Unexpected system mask")
+        assert (exp_str in cmd.stderr_text and cmd.returncode != 0)
+
+    def test_files_ownership_and_permission_teardown(self):
+        """ Method to restore the default bashrc contents"""
+        if self.bashrc_file is not None:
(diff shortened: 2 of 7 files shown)


ipatests/test_integration/test_installation.py
```diff
@@ -567,3 +567,51 @@ def test_KRA_install_after_cert_renew(self):
...
+    """
+
+    num_replicas = 0
+
+    @classmethod
+    def install(cls, mh):
+        pass
+
+    def test_install_master(self):
+        self.master.run_command(['umask', '0027'])
```
flo-renaud: The test needs to write the umask in /root/.bashrc, otherwise it won't execute as expected.
`self.master.run_command` makes a ssh connection to the master, starts a shell, calls umask then disconnects. Because of that, the umask setting is lost as soon as the `run_command` call returns.

ipatests/test_integration/test_installation.py
```diff
@@ -567,3 +567,51 @@ def test_KRA_install_after_cert_renew(self):
...
+
+    @classmethod
+    def install(cls, mh):
+        pass
+
+    def test_install_master(self):
+        tasks.install_master(self.master, setup_dns=False)
+
+    def test_install_replica(self):
+        self.replicas[0].run_command(['umask', '0027'])
```
flo-renaud: Same comment as above, the test needs to write the umask in /root/.bashrc so that it gets permanent and does not get lost as soon as `run_command` returns.

fcami: Thanks for the input! Change is done.

ipatests/test_integration/test_installation.py
fcami: Thanks for the input! Change is done.

fcami:
Too-restrictive mask detection must be better than "if mask != 18:" before merging.

ipaserver/install/installutils.py
```diff
@@ -1389,3 +1389,14 @@ def default_subject_base(realm_name):
...
     return DN(('CN', 'Certificate Authority'), subject_base)
+
+
+def validate_mask():
+    try:
+        mask = os.umask(0)
+    finally:
+        os.umask(mask)
+    mask_str = None
+    if mask & 0b101101 > 0:
```
flo-renaud: Hi @fcami 
I am not sure I understand your binary mask here. If umask is set to 033, we still have read access but the validate_mask function will complain. Shouldn't we use 0b100100100 instead (to make sure we always keep read access)?

ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -45,14 +45,38 @@ jobs:
...
-  fedora-29/temp_commit:
+  fedora-29/test_automount_locations:
     requires: [fedora-29/build]
     priority: 50
     job:
       class: RunPytest
       args:
         build_url: '{fedora-29/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
+        test_suite: test_integration/test_automount_locations.py
```
flo-renaud: This test is probably a copy/paste error from another PR. Not a big deal as temp_commit will be removed, though.

fcami: That was on purpose.
That test installs a server and a client with a non-modified mask and still work with this change.

fcami: Hi @flo
Thanks for the comments, I've updated the PR to add the new tests to f28+f29+rawhide.

ipaserver/install/installutils.py
fcami: Hi @flo-renaud 
Did the explanation seem valid to you? So that we can resolve this.

flo-renaud: Yes, so we accept masks composed of 0 or 2 but nothing else.

ipatests/test_integration/test_installation.py
```diff
@@ -567,3 +567,71 @@ def test_KRA_install_after_cert_renew(self):
...
+    def test_install_replica(self):
+        self.replicas[0].run_command('echo "umask 0027" >> /root/.bashrc')
+        result = self.replicas[0].run_command(['umask'])
+        assert '0027' in result.stdout_text
+
+        cmd = tasks.install_replica(
+            self.master, self.replicas[0],
+            setup_dns=False, setup_ca=False, raiseonerr=False)
+        exp_str = ("Unexpected system mask")
+        assert (exp_str in cmd.stderr_text and cmd.returncode != 0)
```
flo-renaud: The test doesn't need to install a master to call ipa-replica-install (as the command should exit even before the master is contacted). Just use 
self.master.run_command(['ipa-replica-install', '-w',
                                     self.master.config.admin_password,
                                     '-n', self.master.domain.name,
                                     '-r', self.master.domain.realm,
                                     '--server', 'dummy_master.%s'.format(self.master.domain.name),
                                     '-U'])
and set num_replicas=0, this will use only the minimal number of resources.

fcami:
hi @flo-renaud 
The prci_definitions have been updated. Please let me know if I can remove the temp commit.

ipatests/test_integration/test_installation.py
fcami: Thanks, I'm merging both test cases into one with your suggestion.

fcami:
Hi @flo-renaud 
The tests have been added to the CI templates.
Please let me know if I can remove the temp commit.

flo-renaud:
@fcami 
Thanks for the modifications, looks great. You can remove the temp commit and I will ACK

fcami:
Thanks @flo-renaud 
I've removed the temp commit.

fcami:
master:

* f2e7c3f68b691f180acd201a81fe10c7c1491071 ipatests: add too-restritive mask tests
* f90a4b9554656c984c8ac47e3503d83fb52d0b1f ipa-{server,replica}-install: add too-restritive mask detection

PR #2873, opened 2019-03-01
ipatests: fix host name for ssh connection from controller to master

wladich:
Use master.external_hostname instead of master.hostname for ssh connection
from controller machine to master. If hostname and external_hostname in
test_config.yml do no match then trying to establish ssh connection
was failing with "[Errno -2] Name or service not known".

Fixes https://pagure.io/freeipa/issue/7874

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -289,7 +289,8 @@ class TestIPACommand(IntegrationTest):
         """
 
         test_user = 'test-ssh'
-        master = self.master.hostname
+        external_master_hostname = \
+            self.master.external_hostname  # pylint: disable=no-member
 
         pub_keys = []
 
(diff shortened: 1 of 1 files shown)


rcritten:
ipatests/test_integration/test_commands.py:292: [E1101(no-member), TestIPACommand.test_ssh_key_connection] Class 'master' has no 'external_hostname' member)

wladich:
I am confused -- self.master  is instance of Host from pytest-multihost and it has external_hostname attribute. fedora-29/test_commands in CI is green. I also ran pylint manually with same options as Trasis does -- without any error messages.

flo-renaud:
@wladich 
you can add `# pylint: disable=no-member` at the end of the line 292 to ignore the error. I already saw this issue in other tests and did not understand its root cause.

flo-renaud:
Hi @wladich 
Works for me, ACK.

flo-renaud:
master:

* 24c77bc4f778f3caa6272de8659da04abe06b67e ipatests: fix host name for ssh connection from controller to master

wladich:
@flo-renaud Thanks for the note about pylint

PR #2878, opened 2019-03-05
domainlevel_get: return errors.NotFound when domain level is missing

abbra:
If the server doesn't have domain level set, callers in replica
installer expect errors.NotFound but never get it.

Return the right exception here and change the other caller to follow
the same convention.

Use DOMAIN_LEVEL_0 constant instead of 0 as used by the rest of the code.

Fixes: https://pagure.io/freeipa/issue/7876

### DIFF
--- a/ipaserver/plugins/domainlevel.py
+++ b/ipaserver/plugins/domainlevel.py
@@ -10,6 +10,7 @@ from ipalib import errors
 from ipalib import output
 from ipalib.parameters import Int
 from ipalib.plugable import Registry
+from ipalib.constants import DOMAIN_LEVEL_0, MIN_DOMAIN_LEVEL
 
 from ipapython.dn import DN
 
@@ -45,7 +46,7 @@ def get_domainlevel_range(master_entry):
             int(master_entry['ipaMaxDomainLevel'][0])
         )
     except KeyError:
-        return DomainLevelRange(0, 0)
+        return DomainLevelRange(DOMAIN_LEVEL_0, DOMAIN_LEVEL_0)
 
 
 def check_conflict_entries(ldap, api, desired_value):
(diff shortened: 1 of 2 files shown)


ipaserver/plugins/domainlevel.py
```diff
@@ -120,7 +129,7 @@ class domainlevel_set(Command):
         Int('ipadomainlevel',
             cli_name='level',
             label=_('Domain Level'),
-            minvalue=0,
+            minvalue=DOMAIN_LEVEL_0,
```
tiran: Shouldn't this be ``DOMAIN_LEVEL_1`` in master?

ipaserver/plugins/topology.py
```diff
@@ -82,7 +82,12 @@
 
 
 def validate_domain_level(api):
-    current = int(api.Command.domainlevel_get()['result'])
+    current = DOMAIN_LEVEL_0
```
tiran: Same here, master (4.8) no no supports DL 0 and has DL 1 as minimum level.

ipaserver/plugins/topology.py
```diff
@@ -82,7 +82,12 @@
...
+    try:
+        current = int(api.Command.domainlevel_get()['result'])
+    except errors.NotFound:
+        pass
```
tiran: I don't like the default + override style and rather prefer retrieve / except default style. It's also one line shorter:
```
try:
    current = ...
except ...:
    current = DOMAIN_LEVEL_1
```

ipaserver/plugins/topology.py
abbra: I added `MIN_DOMAIN_LEVEL` instead of the explicit DL 0.

abbra:
@frasertweedale Travis CI is failing in IP SAN tests. Could you please check? It looks like there is a problem with the recent code you authored.

frasertweedale:
@abbra unfortunately the log archival failed for that job, so it's a bit hard to see what went wrong.  My theory is that the existing DNS or network configuration of the Travis worker is interfering with the test somehow.  I've restarted the job.  If it happens again hopefully the logs will be archived properly so I can investigate more thoroughly.

frasertweedale:
@abbra PR https://github.com/freeipa/freeipa/pull/2884 should hopefully dispel this problem.

abbra:
@tiran please re-review, I believe all requests are addressed.

tiran:
master:

* b9dc975777f9f84dff20c5a79915d380e6aaaf0f domainlevel-get: fix various issues when running as non-admin

PR #2884, opened 2019-03-07
cert-request: handle missing zone

frasertweedale:
SAN IP address validation, while determining the zone for a DNS name
or IP address, does not handle missing zones.  The resulting
dns.resolver.NoNameservers exception is not caught.  As a result,
InternalError is returned to client.

Update cert-request IP address name validation to handle this case.

Part of: https://pagure.io/freeipa/issue/7451

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -1173,8 +1173,12 @@ def _san_ip_update_reachable(reachable, dnsname, cname_depth):
 
     """
     fqdn = dnsutil.DNSName(dnsname).make_absolute()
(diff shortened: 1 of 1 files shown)


frasertweedale:
Setting `ipa-next` label because the manual backports of the main SAN IP pull request will incorporate this change.

abbra:
LGTM.

frasertweedale:
master:

* [8327e11b6](https://pagure.io/freeipa/c/8327e11b6be6c675a2eb7b453a5e7fb68d0c12c2) cert-request: handle missing zone

PR #2890, opened 2019-03-11
Web UI (topology graph): Show FQDN for nodes if they have no common DNS zone

serg-cymbaluk:
It allows to avoid confusion with identical short hostnames.

There are two cases implemented:
- no common DNS zone: graph shows FQDN for all nodes
- all nodes have one common DNS zone: graph shows DN relatively to the common zone

https://pagure.io/freeipa/issue/7206

### DIFF
--- a/install/ui/test/topology_tests.js
+++ b/install/ui/test/topology_tests.js
@@ -1,0 +1,158 @@
+/**
+ * Copyright (C) 2019  FreeIPA Contributors see COPYING for license
+ */
+
+define([
+        'freeipa/ipa',
+        'freeipa/topology',
+        'freeipa/jquery'],
+    function(IPA, topology, $) {
+        return function() {
+
+var widget;
+
+function inject_data(widget, data) {
+    widget._get_data = function() {
+        return data;
+    };
+}
+
+QUnit.module('topology', {
+    beforeEach: function(assert) {
+        widget = new topology.TopologyGraphWidget(
+            topology.topology_graph_facet_spec
+        );
+        widget.render();
+    }
+});
+
+QUnit.test('Testing TopoGraph nodes', function(assert) {
+    var nodes = [
+        { id: 'master.ipa.test' },
+        { id: 'replica.ipa.test' }
+    ];
+
+    var suffixes = [
+        { cn: ['ca'] },
+        { cn: ['domain'] }
+    ];
+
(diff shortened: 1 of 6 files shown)


install/ui/test/topology_tests.js
```diff
@@ -0,0 +1,147 @@
...
+    marker = rendered_links.last().css('marker-start');
+    assert.ok(marker && marker !== 'none',
+        'Checking left direction link has proper marker');
+});
+
+QUnit.test('Testing TopoGraph for multiple DNS zones', function(assert) {
+    var nodes = [
+        { id: 'master.zone1' },
+        { id: 'replica.zone1' },
+        { id: 'master.zone2' },
```
flo-renaud: FreeIPA does not allow single label domains, you could use instead master.sub.zone1 replica.sub.zone1, master.sub.zone2.

serg-cymbaluk:
@flo-renaud 
I've moved "common domain" logic from TopoGraph to TopologyGraphWidget and so it allows me to add this domain name to the page title. Tests are also updated.

install/ui/test/topology_tests.js
serg-cymbaluk: Fixed

flo-renaud:
Hi @serg-cymbaluk 
I noticed the following behavior. If I install 2 nodes with the same domain, the graph properly shows "Topology graph (domain.com)" and only the short names are displayed below the nodes. Then I add a 3rd node with a completely different domain replica.example.test and hit the "refresh" button. The "Topology graph (domain.com)" label does not get updated, whereas the FQDN are displayed for the nodes.
If I reload the page with Ctrl-R, the topology graph label gets updated ("Topology graph" without any domain name).
So it seems that refresh button does not refresh the "Topology graph" label.

serg-cymbaluk:
@flo-renaud 
> The "Topology graph (domain.com)" label does not get updated, whereas the FQDN are displayed for the nodes.

This behavior has been fixed.

flo-renaud:
Hi @serg-cymbaluk 
the fix works for me, thanks. ACK

serg-cymbaluk:
master:

* 133b199f61a6756723a9a5869f7ff7e415bea455 Web UI (topology graph): Show FQDN for nodes if they have no common DNS zone

PR #2891, opened 2019-03-11
ipa-extdom-exop: add instance counter and limit

sumit-bose:
The user and group lookups done by the extdom plugin might need some
time depending on the state of the service (typically SSSD) handling the
requests.

To avoid that all worker threads are busy waiting on a connect or a
reply from SSSD and no other request can be handled this patch adds an
instance counter and an instance limit for the extdom plugin.

By default the limit will be around 80% of the number of worker threads.
It can be tuned further with the plugin option ipaExtdomMaxInstances
which must in set in ipaextdommaxinstances and should have an integer
value larger than 0 and lesser than the number of worker threads.

If the instance limit is reached the extdom plugin will return LDAP_BUSY
for every new request until the number of instance is again below the
limit.

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom.h
+++ b/daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom.h
@@ -157,6 +157,8 @@ struct ipa_extdom_ctx {
     char *base_dn;
     size_t max_nss_buf_size;
     struct nss_ops_ctx *nss_ctx;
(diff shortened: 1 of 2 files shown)


abbra:
One test failed due to conflict on kadmind port with nfs secure daemon. Re-run.

rcritten:
master:

* 33af8c75b334b847c59ee67f33f9777f58eba41f ipa-extdom-exop: add instance counter and limit

PR #2896, opened 2019-03-14
 WebUI test: Fix automember tests according to new behavior 

serg-cymbaluk:
After deleting user/host from group "rebuild" task is triggered,
so the entity returns to the group. And we check if it exists.

Also the order of cleaning test resources are changed:
groups are being deleted only after corresponding rules.

New automembership design description:

Ticket: https://pagure.io/freeipa/issue/7881

### DIFF
--- a/ipatests/test_webui/test_automember.py
+++ b/ipatests/test_webui/test_automember.py
@@ -296,11 +296,15 @@ class TestAutomember(UI_driver):
         # Remove host from hostgroup
         self.delete_record(host1)
 
-        # Assert that host is not a member of hostgroup
+        # Assert that host was re-added to the group.
+        # The behavior is expected with the plugin default setting:
+        # the entry cn=Auto Membership Plugin,cn=plugins,cn=config has
+        # a default value autoMemberProcessModifyOps: on
+        #
+        # See https://www.port389.org/docs/389ds/design/automember-postop-modify-design.html # noqa: E501
         self.facet_button_click('refresh')
         self.wait_for_request()
-        self.assert_record(host1, negative=True)
-        self.assert_record(host2, negative=True)
+        self.assert_record(host1)
 
         # Rebuild membership for all hosts, using action on hosts search facet
         self.navigate_by_menu('identity/host')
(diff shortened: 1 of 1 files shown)


ipatests/test_webui/test_automember.py
```diff
@@ -296,11 +296,11 @@ def test_rebuild_membership_hosts(self):
         # Remove host from hostgroup
         self.delete_record(host1)
 
-        # Assert that host is not a member of hostgroup
+        # Assert that host was re-added to the group
+        # See https://www.port389.org/docs/389ds/design/automember-postop-modify-design.html
```
flo-renaud: Please add a comment stating that this behavior is expected with the plugin default setting: the entry `cn=Auto Membership Plugin,cn=plugins,cn=config` has a default value `autoMemberProcessModifyOps: on` which means that the plugin gets triggered and re-adds the entry.

ipatests/test_webui/test_automember.py
```diff
@@ -359,11 +359,11 @@ def test_rebuild_membership_users(self):
         # Remove user from group
         self.delete_record('dev1')
 
-        # Assert that user is not a member of group
+        # Assert that user was re-added to the group
+        # See https://www.port389.org/docs/389ds/design/automember-postop-modify-design.html
```
flo-renaud: Please add a comment stating that this behavior is expected with the plugin default setting: the entry cn=Auto Membership Plugin,cn=plugins,cn=config has a default value autoMemberProcessModifyOps: on which means that the plugin gets triggered and re-adds the entry.

ipatests/test_webui/test_automember.py
serg-cymbaluk: Fixed

ipatests/test_webui/test_automember.py
serg-cymbaluk: Fixed

serg-cymbaluk:
@flo-renaud 
PEP8 errors are fixed. Comments are added.

flo-renaud:
Hi @serg-cymbaluk 
thanks for the update, all my comments were addressed and test is green. You can remove the temp commit and I will give the ACK once PRCI finishes.

serg-cymbaluk:
Thanks @flo-renaud

pvoborni:
Thanks, @flo-renaud for doing the review. I promised @serg-cymbaluk that I'll do it last week but did not get to it. Just by reading the code it looks good to me as well.

serg-cymbaluk:
master:

* e12e3e84a9ca746c90be34e25cf374648f30c2b5 WebUI test: Fix automember tests according to new behavior

PR #2902, opened 2019-03-15
XML RPC test: fix test_automember_plugin

flo-renaud:
With 389-DS 1.4.0.21, automember plugin also gets triggered on modify ops
by default. This means that a member manually removed gets automatically
re-added by the plugin.
This behavior can be disabled by setting autoMemberProcessModifyOps=off in
the entry cn=Auto Membership Plugin,cn=plugins,cn=config.

Before 389-DS 1.4.0.21, it was possible to remove a member and the member
did not get re-added (unless automember-rebuild was called). This former
behavior can be forced by setting autoMemberProcessModifyOps=off.

This commit fixes the test and checks the behavior when
autoMemberProcessModifyOps=off and when autoMemberProcessModifyOps=on.

Fixes: https://pagure.io/freeipa/issue/7855

### DIFF
--- a/ipatests/test_xmlrpc/test_automember_plugin.py
+++ b/ipatests/test_xmlrpc/test_automember_plugin.py
@@ -28,10 +28,19 @@ from ipatests.test_xmlrpc.tracker.hostgroup_plugin import HostGroupTracker
 from ipatests.test_xmlrpc.tracker.automember_plugin import AutomemberTracker
 from ipalib import api, errors
 from ipapython.dn import DN
+from ipapython.ipautil import run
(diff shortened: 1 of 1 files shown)


flo-renaud:
master:

* 333784034ecee29e2d141b2dad65f31b15266144 XML RPC test: fix test_automember_plugin

flo-renaud:
Manual backport to ipa-4-6 is needed

flo-renaud:
Correction: no need to backport to ipa-4-6. The 4-6 branch is supported on fedora 27 only and the version of 389-ds in Fedora 27 is 389-ds-base-1.3.9.0-1.fc27.x86_64, which doesn't provide the automember plugin processing on modify operations.

PR #2906, opened 2019-03-18
Send only the path and not the full URI to httplib.request

rcritten:
1. 

Sending the full uri was causing httplib to send requests as:

POST http://ipa.example.com/ca/admin/ca/getStatus HTTP/1.1

https://pagure.io/freeipa/issue/7883

### DIFF
--- a/ipapython/dogtag.py
+++ b/ipapython/dogtag.py
@@ -227,7 +227,7 @@ def _httplib_request(
 
     try:
         conn = connection_factory(host, port, **connection_options)
-        conn.request(method, uri, body=request_body, headers=headers)
+        conn.request(method, path, body=request_body, headers=headers)
         res = conn.getresponse()
 
         http_status = res.status


frasertweedale:
@rcritten et al: do we know how / when this regressed?

rcritten:
From what I can tell tomcat changed its URL handling due to a CVE and tightened things up. This has been wrong in freeipa since inception, d6fbbd530ee94bc4cdd00c9106fd789e50fb81cf , but tomcat handled it fine so we didn't notice.

frasertweedale:
@rcritten thanks for the analysis.  I'm always curious how these things happen.  Might be a good idea to put that info the commit msg too.

rcritten:
Good idea, updated the commit msg.

flo-renaud:
Hi @rcritten 
thanks for the PR, works for me. And I can confirm the issue is linked to tomcat version.

rcritten:
master:

* 6defe320551d745b4feebda79bc9b6e5a17f5350 Send only the path and not the full URI to httplib.request

PR #2912, opened 2019-03-21
[Backport][ipa-4-7] WebUI test: Fix automember tests according to new behavior 

serg-cymbaluk:
Manual backport of #2896 

After deleting user/host from group "rebuild" task is triggered,
so the entity returns to the group. And we check if it exists.

Also the order of cleaning test resources are changed:
groups are being deleted only after corresponding rules.

New automembership design description:

Ticket: https://pagure.io/freeipa/issue/7881

### DIFF
--- a/ipatests/test_webui/test_automember.py
+++ b/ipatests/test_webui/test_automember.py
@@ -152,11 +152,15 @@ class test_automember(UI_driver):
         # Remove host from hostgroup
         self.delete_record(host1)
 
-        # Assert that host is not a member of hostgroup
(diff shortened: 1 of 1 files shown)


flo-renaud:
@serg-cymbaluk 
thanks for the backport, please remove the temp commit.

serg-cymbaluk:
ipa-4-7:

* c2043506bd29c476990c7a4b1b8dbc09366f5cce WebUI test: Fix automember tests according to new behavior

PR #2913, opened 2019-03-21
Coverity: fix issue in ipa_extdom_extop.c

flo-renaud:
Coverity found the following issue:
Error: BAD_COMPARE (CWE-697): [#def1]
freeipa-4.6.5/daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_extop.c:121: null_misuse: Comparing pointer "threadnumber" against "NULL" using anything besides "==" or "!=" is likely to be incorrect.

The comparison is using the pointer while it should use the pointed value.

Fixes: https://pagure.io/freeipa/issue/7884


flo-renaud:
master:

* 3ae38973c5961c17bd95b6831ab7a8469a4bfc5c Coverity: fix issue in ipa_extdom_extop.c

PR #2920, opened 2019-03-22
Completely drop /var/cache/ipa/sessions

stanislavlevin:
This directory has been already dropped in 6d66e826c,
but not entirely.


tiran:
master:

* e5244fbeda60e2e324d2c3bad669f48980635e38 Completely drop /var/cache/ipa/sessions

PR #2926, opened 2019-03-22
support  shared secret trust established from Active Directory domain controller side

abbra:
FreeIPA does support trust to an Active Directory forest. The trust can be
established using administrative credentials from the forest root domain or
using a so-called shared secret. In the latter case no administrative access is
given to the remote side of the trust and each administrator performs their
configuration separately: FreeIPA administrator configures IPA side, Active
Directory administrator adds IPA forest as a trusted one on the Active
Directory side.

For trust to be active, one needs to validate it. Validation process includes a
sequences of DCE RPC calls that force a domain controller on the trusted side
to establish a so-called "secure channel" to a remote domain controller in the
trusting domain. This is an administrative operation and requires
administrative privileges to activate. If trust was established using a shared
secret, IPA side will lack ability to initiate a validation process.

At the same time, FreeIPA 4.6 or earlier versions do not include functionality
to allow a remote validation from Active Directory to happen before trust
objects are created and SSSD can retrieve information from the Active Directory
side. Unfortunately, the latter is not possible until trust is validated.

The purpose of this design is to extend FreeIPA setup to allow trust validation
to be initiated from Windows UI in case a shared secret is used to create a
trust agreement.

TODO: while this code is useful as it is, there is no way to force AD DC to pull forest trust information from IPA DC side. This is due to the fact that Samba does not implement a required RPC call in the mode used by IPA. We need to implement a command that would allow pushing the forest trust topology from IPA side but this command requires administrative privileges on AD side which makes the situation a bit weird for a shared-secret trust.

A lack of forest trust topology update does not prevent resolving users through the trust link. It only prevents single sign-on from Windows side to IPA resources using GSSAPI. Password-based authentication works fine.


daemons/ipa-sam/ipa_sam.c
```diff
@@ -1974,43 +1980,82 @@ static bool handle_cross_realm_princs(struct ipasam_private *ipasam_state,
...
+				/* Second: <OUR FLATNAME$>@<REMOTE REALM> is only used
+				 * for SSSD to be able to talk to AD DCs but it has to
+				 * have canonical name set to <OUR FLATNAME>$ because
+				 * this is the salt used by AD DCs when using this
+				 * principal, otherwise authentication will fail.
+				 *
+				 * *disable* use of this principal on our side as it is
+				 * only used to retrieve trusted domain credentials by
+				 * AD Trust Agents across the IPA topology */
+				ok |= set_krb_princ(ipasam_state, tmp_ctx, r_tdo_alias, princ_r_tdo,
```
tiran: This trick doesn't work. bool false is ``0`` and bool true is ``1``. The boolean or operator ``|`` won't reset ``ok`` to false when one of the ``set_krb_princ`` calls fail.

daemons/ipa-sam/ipa_sam.c
```diff
@@ -2624,13 +2688,20 @@ static NTSTATUS ipasam_set_trusted_domain(struct pdb_methods *methods,
...
-		memset(trustpw, 0, strlen(trustpw));
+		status = get_trust_pwd(tmp_ctx, &td->trust_auth_outgoing,
+				       &trustpw_outgoing, NULL);
+		if (!NT_STATUS_IS_OK(status)) {
+			goto done;
+		}
+		res = set_cross_realm_princs(ipasam_state, td->domain_name, td->netbios_name,
+					     trustpw_incoming, trustpw_outgoing,
+					     td->trust_direction);
+		memset(trustpw_incoming, 0, strlen(trustpw_incoming));
```
tiran: The compiler may optimize the ``memset()`` call away. You need to use C11 ``memset_s`` or some hack with volatile pointers.

ipaserver/install/plugins/adtrust.py
```diff
@@ -1,30 +1,26 @@
...
+
+try:
+    from samba.ndr import ndr_unpack
+    from samba.dcerpc import lsa, drsblobs
+except ImportError:
+    # If samba.ndr is not available, this machine is not provisioned
+    # for serving a trust to Active Directory. As result, it does
+    # not matter what ndr_unpack does but we save on pylint checks
+    def ndr_unpack(x):
+        return x
```
tiran: please ``raise NotImplementedError`` here

abbra:
@tiran, thanks, I changed boolean to counting and added an explicit check to make clear what we expect.
I decided to avoid using memset_s() as it is not really implemented yet anywhere. I used krb5 commit https://github.com/krb5/krb5/commit/c163275f899b201dc2807b3ff2949d5e2ee7d838 for inspiration

abbra:
`kadmind` stumbled on a port used up by NFS again

daemons/ipa-sam/ipa_sam.c
```diff
@@ -1974,46 +1981,111 @@ static bool handle_cross_realm_princs(struct ipasam_private *ipasam_state,
...
+				 *
+				 * A local account for the outbound trust must have
+				 * POSIX and SMB identities associated with our domain but we associate
+				 * them with the trust domain object itself */
+				failed += !set_krb_princ(ipasam_state, tmp_ctx,
+							 princ_l_tdo, l_tdo_alias,
+							 pwd_incoming, trusted_dn,
+							 KRB_PRINC_CREATE_DEFAULT);
+
+	                        ok = (failed == 0);
```
tiran: The indention of this line looks wrong.

daemons/ipa-sam/ipa_sam.c
```diff
@@ -1974,46 +1981,111 @@ static bool handle_cross_realm_princs(struct ipasam_private *ipasam_state,
...
 			}
-			if (!ok) {
-				goto done;
+
+			failed = 0;
+			if ((trust_direction & LSA_TRUST_DIRECTION_OUTBOUND) != 0) {
+				failed += !del_krb_princ(ipasam_state, tmp_ctx, princ_l, trusted_dn);
+				failed += !del_krb_princ(ipasam_state, tmp_ctx, princ_l_tdo, trusted_dn);
+
+	                        ok = (failed == 0);
```
tiran: Same here, wrong indention

ipaserver/install/plugins/adtrust.py
```diff
@@ -325,6 +321,28 @@ def execute(self, **options):
...
 
+def get_gidNumber(ldap, env):
+    # Read the gidnumber of the fallback group
+    dn = DN(('cn', ADTRUSTInstance.FALLBACK_GROUP_NAME),
+            env.container_group,
+            env.basedn)
+
+    try:
+        entry = ldap.get_entry(dn, ['gidnumber'])
+        gidNumber = entry.get('gidnumber')
```
tiran: ``LDAPEntry.get()`` always returns a list. You must use ``entry.single_value.get('gidNumber')`` if you want the GID number.

```
>>> entry = conn.get_entry(DN(('cn', 'editors'), api.env.container_group, api.env.basedn))
>>> entry
LDAPEntry(ipapython.dn.DN('cn=editors,cn=groups,cn=accounts,dc=ipa,dc=example'), {'objectClass': [b'top', b'groupofnames', b'posixgroup', b'ipausergroup', b'ipaobject', b'nestedGroup'], 'gidNumber': [b'487600002'], 'description': [b'Limited admins who can edit other users'], 'cn': [b'editors'], 'ipaUniqueID': [b'f455854e-f88b-11e8-99ba-001a4a231543']})
>>> entry.get('gidNumber')
['487600002']
>>> entry.single_value.get('gidNumber')
'487600002'
```

ipaserver/install/plugins/adtrust.py
```diff
@@ -340,43 +358,55 @@ def execute(self, **options):
...
-                            "Failed to add gidnumber to %s", tdo.dn)
+                    tdo['gidnumber'] = gidNumber
+
+                # Generate uidNumber and ipaNTSecurityIdentifier if
+                # uidNumber is missing. We rely on sidgen plugin here
+                # to generate ipaNTSecurityIdentifier.
+                if not tdo.get('uidnumber'):
+                    tdo['uidnumber'] = ['-1']
+
+                if not ('posixAccount' in tdo.get('objectclass')):
```
tiran: ``if not ('posixAccount' in tdo.get('objectclass'))`` should be ``if 'posixAccount' not in tdo['objectclass']``.

daemons/ipa-sam/ipa_sam.c
abbra: Please ignore it. I have a patch that unifies indentation everywhere in ipasam but it also breaks git history as the patch is touching almost every line of code.

ipaserver/install/plugins/adtrust.py
abbra: Actually, I will not change this place because the 'gidNumber' returned from `get_gidNumber` is directly assigned to an attribute on other `LDAPEntry` objects. This change means a need to do `[]` wrapping in other places for nothing because no other operation is done with this value.

ipaserver/install/plugins/adtrust.py
abbra: Changed, thanks.

tiran:
master:

* 2a8176ee03318f208a86d25bd9320f885b16d112 Add design page for one-way trust to AD with shared secret
* 120bab0d9c9970979d109583329a13fc3e4949f3 trust: allow trust agents to read POSIX identities of trust
* dc8f074cc7ba647644b96785590f1645e5661a93 trusts: add support for one-way shared secret trust
* 18cb30d4638c0fecf5f02735f2b4794be5d97b67 upgrade: upgrade existing trust agreements to new layout
* dca901c05ef6dde69145963d3ea51b994b568740 upgrade: add trust upgrade to actual upgrade code

PR #2923, opened 2019-03-22
Add hidden replica feature

tiran:
A hidden replica is a replica that does not advertise its services via
DNS SRV records, ipa-ca DNS entry, or LDAP. Clients do not auto-select a
hidden replica, but are still free to explicitly connect to it.

- [x] replica install from hidden replica
- [x] DNS install on hidden replica: DNS role has hidden state
- [x] CA install on hidden replica: CA role has hidden state
- [x] KRA install on hidden replica: KRA role has hidden state
- [x] prevent hidden state for CA renewal master ``ca_renewal_master_server``
- [x] prevent hidden state for DNSSEC key master ``dnssec_key_master_server``
- [x] require one non-hidden server for all services
- [x] create replication agreement between hidden and enabled server

Co-authored-by: Francois Cami <fcami@redhat.com>:
Signed-off-by: Christian Heimes <cheimes@redhat.com>

PR #2927, opened 2019-03-22
Hidden replica test

fcami:
Depends on https://github.com/freeipa/freeipa/pull/2923

ipatests: Exercise hidden replica feature
    
A hidden replica is a replica that does not advertise its services via
DNS SRV records, ipa-ca DNS entry, or LDAP. Clients do not auto-select a
hidden replica, but are still free to explicitly connect to it.

Signed-off-by: Francois Cami <fcami@redhat.com>


tiran:
PR #2923 has landed.

PR #2935, opened 2019-03-25
Switch nsslapd-unhashed-pw-switch to nolog

tbordaz:
389-ds will change the default value of nsslapd-unhashed-pw-switch from 'on' to 'off'
For new IPA instance this attribute is set to 'nolog'.
For upgraded IPA instance, it replaces 'off' with 'nolog'.

In case of winsync deployement:
For new IPA instance where the attribute have been set to 'nolog', it is changed to 'on'
when a replication agreement is created toward AD.
For upgrade IPA instance, the attribute was already set to 'on'. Upgrade does not change it

https://pagure.io/freeipa/issue/4812

### DIFF
--- a/ipaserver/install/dsinstance.py
+++ b/ipaserver/install/dsinstance.py
@@ -232,6 +232,7 @@ class DsInstance(service.Service):
         self.step("adding default schema", self.__add_default_schemas)
         self.step("enabling memberof plugin", self.__add_memberof_module)
         self.step("enabling winsync plugin", self.__add_winsync_module)
+        self.step("configure password logging", self.__password_logging)
         self.step("configuring replication version plugin", self.__config_version_module)
         self.step("enabling IPA enrollment plugin", self.__add_enrollment_module)
         self.step("configuring uniqueness plugin", self.__set_unique_attrs)
@@ -730,6 +731,9 @@ class DsInstance(service.Service):
 
     def __add_winsync_module(self):
         self._ldap_mod("ipa-winsync-conf.ldif")
+
+    def __password_logging(self):
+        self._ldap_mod("pw-logging-conf.ldif")
 
     def __config_version_module(self):
         self._ldap_mod("version-conf.ldif")
--- a/ipaserver/install/plugins/update_unhashed_password.py
+++ b/ipaserver/install/plugins/update_unhashed_password.py
@@ -1,0 +1,116 @@
+# Authors:
+#   Thierry Bordaz <tbordaz@redhat.com>
+#
+# Copyright (C) 2019  Red Hat
+# see file 'COPYING' for use and warranty information
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+
+from ipalib import Registry, errors
+from ipalib import Updater
+from ipapython.dn import DN
+
+logger = logging.getLogger(__name__)
+
+register = Registry()
+
+
+@register()
+class update_unhashed_password(Updater):
+    """
+    DS
+    """
+    def __remove_update(self, update, key, value):
+        statement = dict(action='remove', attr=key, value=value)
+        update.setdefault('updates', []).append(statement)
+
+    def __add_update(self, update, key, value):
+        statement = dict(action='add', attr=key, value=value)
+        update.setdefault('updates', []).append(statement)
+
+    def execute(self, **options):
+        logger.debug("Upgrading unhashed password configuration")
+        ldap = self.api.Backend.ldap2
+        base_config = DN(('cn', 'config'))
+        try:
+            entry = ldap.get_entry(base_config,
+                                   ['nsslapd-unhashed-pw-switch'])
+        except errors.NotFound:
+            logger.error("Unhashed password configuration not found")
+            return False, []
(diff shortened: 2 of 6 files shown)


tiran:
Linter is failing:

```
************* Module ipaserver.install.plugins.update_unhashed_password
ipaserver/install/plugins/update_unhashed_password.py:24: [W0611(unused-import), ] Unused certstore imported from ipalib.install)
ipaserver/install/plugins/update_unhashed_password.py:26: [W0611(unused-import), ] Unused get_ca_nickname imported from ipapython.certdb)
```

ipaserver/install/plugins/update_unhashed_password.py
```diff
@@ -0,0 +1,111 @@
...
+            entry = ldap.get_entry(base_config,
+                                   ['nsslapd-unhashed-pw-switch'])
+        except errors.NotFound:
+            logger.error("Unhashed password configuration not found")
+            return False, []
+
+        config_dn = entry.dn
+
+        toggle = entry.single_value.get("nsslapd-unhashed-pw-switch")
+        if toggle.lower() not in ['off', 'on', 'nolog']:
```
flo-renaud: Is nsslapd-unhashed-pw-switch a mandatory attribute? Otherwise the code needs to handle the case where toggle=None.

ipaserver/install/plugins/update_unhashed_password.py
```diff
@@ -0,0 +1,111 @@
...
+        if len(winsync_agmts) > 0:
+            # We are running in a winsync environment
+            # Log a warning that changelog will contain sensitive data
+            try:
+                cldb_e = ldap.get_entry(
+                    DN(('cn', 'changelog5'),
+                       ('cn', 'config')),
+                    ['nsslapd-changelogdir'])
+                cldb = cldb_e.single_value.get("nsslapd-changelogdir")
+                logger.warning("This winsync configuration may imply "
```
flo-renaud: I would replace the message with something more affirmative, like "As this server is configured for winsync, the changelog files under %s may contain clear text passwords. Please protect these files with appropriate read permissions."

ipaserver/install/plugins/update_unhashed_password.py
```diff
@@ -0,0 +1,111 @@
...
+                               "passwords.\n"
+                               "Please ensure that these files can be accessed"
+                               " only by trusted accounts."
+                               "log files are under %s", cldb)
+            except errors.NotFound:
+                logger.warning("This winsync configuration may imply "
+                               "that the log file contains clear text "
+                               "passwords.\n"
+                               "Please ensure that these files can be accessed"
+                               " only by trusted accounts.")
```
flo-renaud: Same comment as above re. the message.

ipaserver/install/plugins/update_unhashed_password.py
```diff
@@ -0,0 +1,111 @@
...
+        except errors.NotFound:
+            logger.debug("Unhashed password this is not a winsync deployment")
+            winsync_agmts = []
+
+        update = {
+            'dn': config_dn,
+            'updates': [],
+        }
+        if len(winsync_agmts) > 0:
+            # We are running in a winsync environment
```
flo-renaud: if toggle.lower() == 'on' you can skip this code and set updates = []

ipaserver/install/replication.py
```diff
@@ -1199,6 +1199,46 @@ def setup_replication(self, r_hostname, r_port=389, r_sslport=636,
...
+                    DN(('cn', 'changelog5'),
+                       ('cn', 'config')),
+                    ['nsslapd-changelogdir'])
+                cldb = entry.single_value.get("nsslapd-changelogdir")
+                logger.warning("This configuration (\"--winsync\") may imply "
+                               "that the log file contains clear text "
+                               "passwords.\n"
+                               "Please ensure that these files can be accessed"
+                               " only by trusted accounts."
+                               "log files are under %s", cldb)
```
flo-renaud: Same comment as above regarding the message.

ipaserver/install/dsinstance.py
```diff
@@ -421,6 +421,7 @@ def __setup_replica(self):
             self.dm_password)
 
         repl, bind_dn, bind_pw = self._get_replication_manager()
+        repl.unhashed_password_log(repl.conn, 'nolog')
```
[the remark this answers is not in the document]
tbordaz: Indeed if upgrade scripts are called on fresh install (server or replica) I agree it looks better to let the upgrade script decide which value to set. I will make the change

ipaserver/install/plugins/update_unhashed_password.py
tbordaz: It is a mandatory attribute, even if it is not listed in dse.ldif it has a default value. So it can only be 'on', 'off' or 'nolog'

ipaserver/install/plugins/update_unhashed_password.py
tbordaz: Right. I will change this

ipaserver/install/plugins/update_unhashed_password.py
tbordaz: I am unsure. If for some reason we have a old DS version that set the toggle to 'on'. Would not it be better if the IPA code make the config more secure if possible.
Also this code displays the warning.
To avoid useless update, would it be okay in case of winsync agreements to return updates=[] after the warning if "toggle.lower='on' "?

flo-renaud: Ok for me to print the message in this case.

flo-renaud:
Hi @tbordaz 
thanks for the latest update, LGTM. I tested:

- [x] upgrade from a single master or a replicated master: pw switch set to nolog as expected
- [x] upgrade from an instance iwth a winsync agreement: pw switch set to on as expected
- [x] fresh install: pw switch set to nolog as expected
- [x] fresh install + config of a winsync agreement: pw switch set to on as expected

flo-renaud:
Also tested winsync + pass sync, replication works in both directions.
ACK

flo-renaud:
master:

* 67490acb045697302cc16d34cac3a81bc1f909ba Switch nsslapd-unhashed-pw-switch to nolog

PR #2943, opened 2019-03-27
ipa_sam: remove dependency to sid_string_talloc() and sid_string_dbg()

mastersin:
ipa_sam uses Samba's sid to string construction functions. Recent Samba
versions removed this functions from libsmbconf.so and replace new functions
to internal libsamba-security-samba4.so

Thus, we reimplement new style function dom_sid_str_buf() from Samba-4.10.

Resolves: https://pagure.io/freeipa/issue/7893

### DIFF
--- a/daemons/ipa-sam/ipa_sam.c
+++ b/daemons/ipa-sam/ipa_sam.c
@@ -4,9 +4,12 @@
 #include "config.h"
 #include <stdbool.h>
 #include <stdint.h>
+#include <inttypes.h>
+#include <string.h>
 #include <stdio.h>
 #include <unistd.h>
 #include <sys/types.h>
+#include <sys/param.h>
 #include <pwd.h>
 #include <errno.h>
 #include <ldap.h>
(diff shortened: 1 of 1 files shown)


daemons/ipa-sam/ipa_sam.c
```diff
@@ -100,12 +101,15 @@ struct unixid {
 	enum id_type type;
 }/* [public] */;
 
+/* from samba/libcli/security/dom_sid.h */
+#define MAX(a,b) ((a)>(b)?(a):(b))
```
abbra: You can get `MAX` from `#include <sys/param.h>`.

daemons/ipa-sam/ipa_sam.c
```diff
@@ -283,6 +287,62 @@ static bool is_null_sid(const struct dom_sid *sid)
 	return true;
 }
 
+static int dom_sid_string_buf(const struct dom_sid *sid, char *buf, int buflen)
```
abbra: Please rename to `_domain_sid_string_buf` to avoid possible conflicts when loading in the same process.

daemons/ipa-sam/ipa_sam.c
```diff
@@ -283,6 +287,62 @@ static bool is_null_sid(const struct dom_sid *sid)
...
+			sid->sub_auths[i]);
+		if (ret < 0) {
+			return ret;
+		}
+		ofs += ret;
+	}
+	return ofs;
+}
+
+static char *dom_sid_str_buf(const struct dom_sid *sid, struct dom_sid_buf *dst)
```
abbra: Same here, please rename to `_domain_sid_str_buf`

abbra:
Actually, I have a better idea. We are linking anyway to SSSD's `libsss_idmap` library which has all the same code already. See calls to `sss_idmap_sid_to_smb_sid()` -- this converts stringified SID version to `struct dom_sid`. There is a reverse function too: `sss_idmap_smb_sid_to_sid()`. It takes idmap context which we already have initialized in the ipasam state and we use talloc to allocate resources. This would make it consistent with other places we have.

mastersin:
Fix and rebase current variant. Thinks about better idea as next step.

abbra:
@mastersin let me know if you are able to work on the proposed approach. If not, I'll do it myself next week.

mastersin:
@abbra I try it and understand, that I don't know enough clearly how works talloc contexts in this code, and also I don't have time to debug it yet. I afraid that I may add memory leaks in proposed approach.

abbra:
@mastersin I submitted https://github.com/freeipa/freeipa/pull/2966 that should solve the issue and obsoletes this PR. Please review.

tiran:
PR #2966 with an alternative fix has landed.

tiran:
Thanks for your patch any way! :1st_place_medal:

PR #2951, opened 2019-03-28
Extend test for orphan automember rules (issue/6476)

t-woerner:
The test was not executing ipa automember-rebuild --type hostgroup.

The test has been extended to execute it twice: Once when it needs to fail
because there is an orphan automember rule. Also after this orphan
automember rule has been removed. Here the test needs to succeed.

Fixes: https://pagure.io/freeipa/issue/7891

### DIFF
--- a/ipatests/test_xmlrpc/test_automember_plugin.py
+++ b/ipatests/test_xmlrpc/test_automember_plugin.py
@@ -837,6 +837,15 @@ class TestAutomemberFindOrphans(XMLRPC_test):
 
         hostgroup1.ensure_missing()
 
+        # Test rebuild (is failing)
+        try:
+            api.Command['automember_rebuild'](type=u'hostgroup')
+        except errors.DatabaseError:
+            pass
+        else:
+            pytest.fail("automember_rebuild was not failing with "
+                        "an orphan automember rule")
+
         # Find obsolete automember rules
         result = api.Command['automember_find_orphans'](type=u'hostgroup')
         assert result['count'] == 1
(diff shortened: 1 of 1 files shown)


ipatests/test_xmlrpc/test_automember_plugin.py
```diff
@@ -837,6 +837,14 @@ def test_find_orphan_automember_rules(self, hostgroup1):
 
         hostgroup1.ensure_missing()
 
+        # Test rebuild (is failing)
+        try:
+            api.Command['automember_rebuild'](type=u'hostgroup')
+        except errors.DatabaseError:
+            pass
+        else:
+            assert False
```
tiran: You can use ``pytest.fail(msg)`` to raise a test failure with human readable message.

ipatests/test_xmlrpc/test_automember_plugin.py
```diff
@@ -850,6 +858,14 @@ def test_find_orphan_automember_rules(self, hostgroup1):
         result = api.Command['automember_find_orphans'](type=u'hostgroup')
         assert result['count'] == 0
 
+        # Test rebuild (may not be failing)
+        try:
+            api.Command['automember_rebuild'](type=u'hostgroup')
+        except errors.DatabaseError:
+            assert False
+        else:
```
tiran: ``else: pass`` is not required here.

ipatests/test_xmlrpc/test_automember_plugin.py
t-woerner: Fixed

ipatests/test_xmlrpc/test_automember_plugin.py
t-woerner: Fixed

tiran:
master:

* 19db292e6d97dfa2e55a71829172e0900da2a514 Extend test for orphan automember rules (issue/6476)

PR #2966, opened 2019-03-31
ipasam: use SID formatting calls to libsss_idmap

abbra:
Samba 4.10 moved away to private libraries two functions we used to
convert a binary SID structre to strings:
 - sid_talloc_string()
 - sid_string_dbg()

We already used libsss_idmap to convert textual representation of SIDs
to a binary one, use the reverse function too.

libsss_idmap code operates on talloc structures, so we need to adopt a
bit a place where sid_string_dbg() was used because it assumed a static
buffer was provided by sid_string_dbg().

Finally, sid_talloc_string()'s replacement moves allocated memory to the
right context so that a memory will be freed earlier. Our SSSD idmap
context is a long-living one while in all cases where we were using
sid_talloc_string() we free the context much earlier.

Resolves: https://pagure.io/freeipa/issue/7893

### DIFF
--- a/daemons/ipa-sam/ipa_sam.c
+++ b/daemons/ipa-sam/ipa_sam.c
@@ -104,8 +104,6 @@ enum ndr_err_code ndr_pull_trustAuthInOutBlob(struct ndr_pull *ndr, int ndr_flag
 bool sid_check_is_builtin(const struct dom_sid *sid); /* available in libpdb.so */
 /* available in libpdb.so, renamed from sid_check_is_domain() in c43505b621725c9a754f0ee98318d451b093f2ed */
 bool sid_linearize(char *outbuf, size_t len, const struct dom_sid *sid); /* available in libsmbconf.so */
-char *sid_string_talloc(TALLOC_CTX *mem_ctx, const struct dom_sid *sid); /* available in libsmbconf.so */
-char *sid_string_dbg(const struct dom_sid *sid); /* available in libsmbconf.so */
 char *escape_ldap_string(TALLOC_CTX *mem_ctx, const char *s); /* available in libsmbconf.so */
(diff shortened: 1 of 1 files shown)


daemons/ipa-sam/ipa_sam.c
```diff
@@ -519,8 +529,18 @@ static bool ldapsam_extract_rid_from_entry(LDAP *ldap_struct,
 	}
 
 	if (dom_sid_compare_domain(sid, domain_sid) != 0) {
-		DEBUG(10, ("SID %s is not in expected domain %s\n",
-			   str, sid_string_dbg(domain_sid)));
+		char *debug_domain_sid = NULL;
+		err = sss_idmap_smb_sid_to_sid(idmap_ctx,
+					       discard_const(domain_sid),
+					       &debug_domain_sid);
```
sumit-bose: Is there a reason why you call sss_idmap_smb_sid_to_sid() here directly instead of sid_talloc_string()? You can use (although not nice) str or sid as memory context since those are freed in the end. Or by adding a tmp_ctx which could also be used to remove the talloc on NULL in get_single_attribute().

I'm asking because having at least sss_idmap_smb_sid_to_sid() only in one place would help to make the transition into the Samba source tree more easy since I don't expect that you want to introduce a dependency on libsss_idmap into Samba.

abbra: I considered that and decided to keep it as it is. We have `sss_idmap_sid_to_smb_sid()` already in use so this part of the code will anyway be rewritten. I'm willing to keep this one oddity.

tiran:
master:

* ffcbb835089c8fb71eadd48230e3c034d7ecda96 ipasam: use SID formatting calls to libsss_idmap

PR #2967, opened 2019-04-01
ipa-server-upgrade: fix add_systemd_user_hbac

flo-renaud:
During upgrade, the method add_systemd_user_hbac is creating
a hbacsvc and a hbacrule, but fails in python2 because of
unicode conversion errors.
The arguments should be defined as u'value'.

Fixes: https://pagure.io/freeipa/issue/7896


tiran:
master:

* d60122f9fba9fd9f74721763c4d5cf9f9f93986c ipa-server-upgrade: fix add_systemd_user_hbac

PR #2968, opened 2019-04-01
ipa-setup-kra: fix python2 parameter

flo-renaud:
ipa-setup-kra is failing in python2 with
invalid 'role_servrole': must be Unicode text
because of a unicode conversion error.

The method api.Command.server_role_find is called with the parameter
role_servrole='IPA master' but it should rather be
role_servrole=u'IPA master'

Fixes: https://pagure.io/freeipa/issue/7897

### DIFF
--- a/ipaserver/install/service.py
+++ b/ipaserver/install/service.py
@@ -213,7 +213,7 @@ def sync_services_state(fqdn):
     """
     result = api.Command.server_role_find(
         server_server=fqdn,
-        role_servrole='IPA master',
+        role_servrole=u'IPA master',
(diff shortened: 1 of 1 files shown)


tiran:
Thanks for the fix!

Note: The patch is only required for Python 2 compatibility on RHEL 7. The master branch no longer supports Python 2. It doesn't hurt to have the change in master, though.

tiran:
master:

* 23ae171d97198de0c959201871eb81ad8b83405b ipa-setup-kra: fix python2 parameter

PR #2970, opened 2019-04-01
Disable flaky hidden replica backup test

tiran:
The test case for hidden replica restore is flaky and sometimes fails.
The general issues is covered by upstream bug 7894.

See: https://pagure.io/freeipa/issue/7894


fcami:
As the backup/restore part still has issues we can disable the DNS part.
@tiran shouldn't we mark it as xfail instead?

tiran:
Only the DNS part of the test is flaky. It sometimes passes and sometimes fails. I expect it to become stable once we have fixed the backup and restore case.

fcami:
OK, makes sense. ACK.

fcami:
master:

* c3fc551c2ab844c02f8fcb3eeb1905129f2ae274 Disable flaky hidden replica backup test

PR #2976, opened 2019-04-01
Add PKI config override option

tiran:
Add an option to override CA and KRA settings passed to pkispawn. The feature allows users to change key size, signature algorithm, and other parameters. It's a prerequisite for HSM support.

The patchset also simplifies and improves how IPA creates the pki.ini files that gets passed to pkispawn.

See https://pagure.io/freeipa/issue/5608

Split of PR #2307

### DIFF
--- a/ipatests/test_integration/test_pki_config_override.py
+++ b/ipatests/test_integration/test_pki_config_override.py
@@ -1,0 +1,37 @@
+#
+# Copyright (C) 2019  FreeIPA Contributors see COPYING for license
+#
+"""Test cases for PKI config overrides
+"""
+from __future__ import absolute_import
+
+from ipalib.x509 import load_pem_x509_certificate
+from ipaplatform.paths import paths
+from ipatests.test_integration.base import IntegrationTest
+from ipatests.pytest_ipa.integration import tasks
+
+
+KEY_OVERRIDE = """
+[DEFAULT]
+ipa_key_size=4096
+"""
+
+
+class TestPKIConfigOverride(IntegrationTest):
+    @classmethod
+    def install(cls, mh):
+        pki_ini = tasks.upload_temp_contents(cls.master, KEY_OVERRIDE)
+        extra_args = [
+            '--pki-config-override', pki_ini,
+        ]
+        tasks.install_master(
+            cls.master, setup_dns=False, extra_args=extra_args
+        )
+        cls.master.run_command(['rm', '-f', pki_ini])
+
+    def test_cert_rsa4096(self):
+        ca_pem = self.master.get_file_contents(
+            paths.IPA_CA_CRT, encoding=None
+        )
+        cert = load_pem_x509_certificate(ca_pem)
+        assert cert.public_key().key_size == 4096
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1082,6 +1082,7 @@ fi
 # END
 %{_usr}/share/ipa/wsgi.py*
 %{_usr}/share/ipa/kdcproxy.wsgi
+%{_usr}/share/ipa/ipaca*.ini
 %{_usr}/share/ipa/*.ldif
 %{_usr}/share/ipa/*.uldif
 %{_usr}/share/ipa/*.template
--- a/install/share/Makefile.am
+++ b/install/share/Makefile.am
@@ -96,6 +96,8 @@ dist_app_DATA =				\
 	ipa-rewrite.conf.template	\
 	min-ssf.ldif			\
 	ipaca_default.ini		\
+	ipaca_customize.ini		\
+	ipaca_softhsm2.ini		\
 	$(NULL)
 
 kdcproxyconfdir = $(IPA_SYSCONF_DIR)/kdcproxy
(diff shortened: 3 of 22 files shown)


frasertweedale:
@tiran Thanks for extracting this as a separate PR.  I shall review it in the next day or so!

ipatests/test_integration/test_pki_config_override.py
```diff
@@ -0,0 +1,37 @@
...
+        pki_ini = tasks.upload_temp_contents(cls.master, KEY_OVERRIDE)
+        extra_args = [
+            '--pki-config-override', pki_ini,
+        ]
+        tasks.install_master(
+            cls.master, setup_dns=False, extra_args=extra_args
+        )
+        cls.master.run_command(['rm', '-f', pki_ini])
+
+    def test_cert_rsa4096_sha389(self):
```
abbra: Please fix the typo: `_sha389` -> `_sha384`

tiran:
http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/e3b26a8e-5530-11e9-934d-fa163e7c6875/ is the temp commit with new test case.

abbra:
@tiran looks good. However, I wonder if we could add a small summary to the installer output that says we are using pki override and quote most important values from there? Like key length for root CA, signing algorithm and similar information.

abbra:
There are also few more problems to solve in a separate ticket as we have to change our communication with certmonger to make sure we pass key lengths to the requests. certmonger defaults to 2048 if no key length is passed in the request and if we did install with larger key sizes, any renewal will break that.
It should be noted already in my pull request https://github.com/freeipa/freeipa/pull/2291 in the commit https://github.com/freeipa/freeipa/pull/2291/commits/a3b4275e8b0e8af8d436c93aa1446dfcfe9568ed for root CA but it should be taken into account for any cert we issue.

tiran:
> @tiran looks good. However, I wonder if we could add a small summary to the installer output that says we are using pki override and quote most important values from there? Like key length for root CA, signing algorithm and similar information.

This sounds like a good idea, but can be handled independently of this patch. It would also be useful for external CA installation. I have opened ticket https://pagure.io/freeipa/issue/7899 .

> There are also few more problems to solve in a separate ticket as we have to change our communication with certmonger to make sure we pass key lengths to the requests. certmonger defaults to 2048 if no key length is passed in the request and if we did install with larger key sizes, any renewal will break that.
> It should be noted already in my pull request #2291 in the commit [a3b4275](https://github.com/freeipa/freeipa/commit/a3b4275e8b0e8af8d436c93aa1446dfcfe9568ed) for root CA but it should be taken into account for any cert we issue.

The issue in https://github.com/freeipa/freeipa/commit/a3b4275e8b0e8af8d436c93aa1446dfcfe9568ed should be fixed in a different way. We cannot assume any fixed key size. It's possible to have different key sizes for CA, and subsystem certs. The current key size of a request can be retrieved from certmonger with ``request.prop_if.Get(DBUS_CM_REQUEST_IF, 'key-size')``. I'm not sure this works for a pending request, though.

Also Certmonger should not assume any key size when renewing a certificate. Instead it should inspect the certificate and use the same key size and signature algorithm as the existing certificate.

abbra:
> Also Certmonger should not assume any key size when renewing a certificate. Instead it should inspect the certificate and use the same key size and signature algorithm as the existing certificate.
I think it actually does this but not for the CA cert that we request to renew when doing external CA switch because we don't have a request for that.

frasertweedale:
A few comments / questions:

1. `ipa-server-install` and `ipa-ca-install` man pages need to document the `--pki-config-override` option.

2. We introduce several new configs for PKI configuration, which can be used in the override file (`ipa_key_size`, `ipa_signing_algorithm`, et al).  I think we should properly document these somewhere (e.g. a man page).  Doesn't have to be in this PR but what are your thoughts?

Apart from (1) it's an ACK.

tiran:
@frasertweedale I have added short entries to the man pages. The documentation is obviously not yet helpful. I'll improve it step by step once we have figured out which options we are going to endorse and support.

frasertweedale:
@tiran thanks, here have an ACK :)

tiran:
master:

* 8c4d75fd2e2972495a0693f1f86a64f09fc929f2 Add current default.cfg from Dogtag
* 0a2b02fc620ef16af87be5ec2ba47cd6831059f5 Simplify and slim down ipaca_default.ini
* 70beccada2e439b9d8d32dbeab369cdb7f0a8383 Add IPA specific vars to ipaca_default.ini
* f847d7756f8d15e47930a709374cf3d95e4b4d8a Use new pki_ipaca.ini to spawn instances
* dd47cfc75a69618f486abefb70f2649ebf8264e7 Add pki.ini override option
* 94937424b12da3e7294ebbcb256b9423ee2d150e Simplify and consolidate ipaca.ini
* dba89712c6a67390c616d5950efab39a73dd8c16 Verify pki ini override early
* 42efdc7bb1c208ac02d45865d68df959f87d5148 Add test case for pki config override
* 2b2c5d6c931660e73cc1c9e75aa3055ab20f386b Add --pki-config-override to man pages

PR #2977, opened 2019-04-01
[Backport ipa-4-6] Trust pass args and options

abbra:
Refactor com.redhat.idm.trust-fetch.domains oddjob helper to allow
passing administrative credentials and a domain controller to talk to.

This approach allows to avoid rediscovering a domain controller in case
a user actually specified the domain controller when establishing trust.

It also allows to pass through admin credentials if user decides to do
so. The latter will be used later to allow updating trust topology in a
similar oddjob helper.

Resolves: https://pagure.io/freeipa/issue/7895
(cherry picked from commit de4a9875d410c68ae4f9602b70c753a11034b31b)

### DIFF
--- a/API.txt
+++ b/API.txt
@@ -5765,10 +5765,12 @@ output: Output('result', type=[<type 'dict'>])
 output: Output('summary', type=[<type 'unicode'>, <type 'NoneType'>])
 output: ListOfPrimaryKeys('value')
 command: trust_fetch_domains/1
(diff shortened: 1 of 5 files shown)


Tiboris:
ipa-4-6:

* 916d4f3cd0bf26e96f696903bd6fb19693437712 oddjob: allow to pass options to trust-fetch-domains

PR #2982, opened 2019-04-02
Move DS's Kerberos env vars to unit file

tiran:
The IPA specific env vars KRB5_KTNAME and KRB5CCNAME are now defined in a instance specific ipa-env.conf unit file.
    
Fixes: https://pagure.io/freeipa/issue/7860

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -1104,13 +1104,16 @@ def reverse_record_exists(ip_address):
     return True
 
 
-def config_replace_variables(filepath, replacevars=dict(), appendvars=dict()):
+def config_replace_variables(filepath, replacevars=dict(), appendvars=dict(),
+                             removevars=None):
     """
     Take a key=value based configuration file, and write new version
-    with certain values replaced or appended
+    with certain values replaced, appended, or removed.
 
     All (key,value) pairs from replacevars and appendvars that were not found
     in the configuration file, will be added there.
+
+    All entries in set removevars are removed.
 
     It is responsibility of a caller to ensure that replacevars and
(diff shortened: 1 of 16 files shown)


ipapython/ipautil.py
```diff
@@ -1104,14 +1104,17 @@ def reverse_record_exists(ip_address):
     return True
 
 
-def config_replace_variables(filepath, replacevars=dict(), appendvars=dict()):
+def config_replace_variables(filepath, replacevars=dict(), appendvars=dict(),
+                             removevars=None):
     """
     Take a key=value based configuration file, and write new version
-    with certain values replaced or appended
+    with certain values replaced, appended, ir removed
```
abbra: Please fix the typo `ir removed`

tiran: - Restore from a different version is not supported. ipa-server-upgrade should fix the cause, though.
- I have added ``systemd_daemon_reload``.

tiran:
- Restore from a different version is not supported. ipa-server-upgrade should fix the cause, though.
- I have added ``systemd_daemon_reload``.

abbra:
LGTM. I'll wait for the remaining CI test before adding labels.

tiran:
master:

* 3cb1ccb3b0a4f361d4d3ef667c60d5b6f4a23333 Add option to remove lines from a file
* e4621f1204c8ffc3e1279cea2a897292d560a48c Add tasks.systemd_daemon_reload()
* 1dfac4f5b7e9986e1217d6979b854b94de094aae Move DS's Kerberos env vars to unit file

PR #2988, opened 2019-04-03
Samba 4.9 related backport to ipa-4-6

abbra:
When backporting Samba related patches for ticket https://pagure.io/freeipa/issue/7705, we missed two patches on top of those in ipa-4-6 backport.

As result, FreeIPA 4.6.5 does not work with Samba 4.9 because we aren't able to set up group mapping for BUILTIN\Guests properly.


flo-renaud:
Since it's a simple backport, ACK.

flo-renaud:
ipa-4-6:

* 999b0dc30082fef0e327c25c9577ec8b69449fa4 adtrust: define Guests mapping after creating cifs/ principal
* e9f48f623e432c5ab439ac9a5187c480344852a3 net groupmap: force using empty config when mapping Guests

PR #2989, opened 2019-04-03
Verify external CA's basic constraint pathlen

tiran:
IPA no verifies that intermediate certs of external CAs have a basic constraint path len of at least 1 and increasing.
    
Fixes: https://pagure.io/freeipa/issue/7877

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -893,9 +893,12 @@ def load_pkcs12(cert_files, key_password, key_nickname, ca_cert_files,
                 "The full certificate chain is not present in %s" %
                 (", ".join(cert_files)))
 
-        for nickname in trust_chain[1:]:
+        # verify CA validity and pathlen. The trust_chain list is in reverse
+        # order. trust_chain[1] is the first intermediate CA cert and must
+        # have pathlen >= 0.
+        for minpathlen, nickname in enumerate(trust_chain[1:], start=0):
             try:
-                nssdb.verify_ca_cert_validity(nickname)
+                nssdb.verify_ca_cert_validity(nickname, minpathlen)
             except ValueError as e:
                 raise ScriptError(
                     "CA certificate %s in %s is not valid: %s" %
@@ -1038,9 +1041,12 @@ def load_external_cert(files, ca_subject):
                 "missing certificate with subject '%s'" %
                 (", ".join(files), issuer))
 
-        for nickname in trust_chain:
+        # verify CA validity and pathlen. The trust_chain list is in reverse
+        # order. The first entry is the signed IPA-CA and must have a
+        # pathlen of >= 0.
+        for minpathlen, nickname in enumerate(trust_chain, start=0):
             try:
-                nssdb.verify_ca_cert_validity(nickname)
+                nssdb.verify_ca_cert_validity(nickname, minpathlen)
             except ValueError as e:
                 cert, subject, issuer = cache[nickname]
                 raise ScriptError(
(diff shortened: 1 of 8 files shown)


tiran:
http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/9f344eb0-55fb-11e9-bdab-fa163e55d08e/report.html is the log file from the temp commit. Installation with invalid pathlen restriction failed with

```
A certificate CN=example.test in /ipatests/ipa_ca.crt, /ipatests/root_ca.crt is not valid: basic contraint pathlen 0, must be at least 2
The ipa-server-install command failed. See /var/log/ipaserver-install.log for more information
```

tiran:
The check was too strict. I have updated the PR to allow a pathlen of 0 for IPA-CA. It's still ok for customers that do not use sub CAs.

ipaserver/install/installutils.py
```diff
@@ -893,9 +893,12 @@ def load_pkcs12(cert_files, key_password, key_nickname, ca_cert_files,
                 "The full certificate chain is not present in %s" %
                 (", ".join(cert_files)))
 
-        for nickname in trust_chain[1:]:
+        # verify CA validity and pathlen. The trust_chain list is in reverse
+        # order. The first entry is the signed IPA-CA and must have a
```
frasertweedale: The comment is wrong - this is the method for checking 3rd party service certs right?
Should say "The first entry is the end entity certificate..."

frasertweedale:
@tiran looks pretty good! One comment needs fixing.  Also, the negative test is fine but should we have a positive test, i.e. one that checks that external CA chain containing *valid* pathLenConstraints actually succeeds?

ipaserver/install/installutils.py
tiran: The list is in reverse order (``trust_chain = list(reversed(nssdb.get_trust_chain(key_nickname)))``) and starts with the first intermediate certificate.

tiran:
We have plenty of tests that install IPA with an external cert chain. Do we need another test?

ipaserver/install/installutils.py
tiran: I have updated the comment.

frasertweedale:
> We have plenty of tests that install IPA with an external cert chain. Do we need another test?

Do those chains have pathLenConstraints in them?

ipaserver/install/installutils.py
frasertweedale: Great, thanks!

abbra:
Yes, we do have that supported, though there is only `path_length=1` in almost all tests:
```
$ git grep sign_csr ipatests 
ipatests/create_external_ca.py:    def sign_csr(self, ipa_csr, path_length=1):
ipatests/create_external_ca.py:    ipa_ca = external_ca.sign_csr(ipa_csr)
ipatests/pytest_ipa/integration/tasks.py:    ipa_ca = external_ca.sign_csr(ipa_csr)
ipatests/test_integration/test_caless.py:        ipa_ca = external_ca.sign_csr(ipa_csr)
ipatests/test_integration/test_external_ca.py:        ipa_ca = external_ca.sign_csr(ipa_csr)
ipatests/test_integration/test_external_ca.py:        ipa_ca = external_ca.sign_csr(ipa_csr)
ipatests/test_ipaclient/test_csrgen.py:        csr_der = adaptor.sign_csr(reqinfo)
```
But support is there and for externally signed IPA CA we have explicit `path_length=1`.

abbra:
Another case of `kadmind` hitting NFS server port reservation.

frasertweedale:
@abbra OK, thanks.  @tiran you have my ACK!

tiran:
Thanks @frasertweedale :+1:

tiran:
master:

* 3c354e74f389da6bfd8f444a58cf3900fdf97434 Verify external CA's basic constraint pathlen

PR #2990, opened 2019-04-03
Adapt cert-find performance workaround for users

tiran:
**experimental patch**

ipa cert-find --users=NAME was slow on system with lots of certificates.
User certificates have CN=$username, therefore the performance tweak
from ticket 7835 also works for user certificates.

Related: https://pagure.io/freeipa/issue/7835
Fixes: https://pagure.io/freeipa/issue/7901

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -1643,21 +1643,26 @@ class cert_find(Search, CertMethod):
         result = collections.OrderedDict()
         complete = bool(ra_options)
 
-        # workaround for RHBZ#1669012
-        # Improve performance for service and host case by also searching
-        # for subject. This limits the amount of certificate retrieved from
-        # Dogtag. The special case is only used, when no ra_options are set
-        # and exactly one service or host is supplied.
-        # The complete flag is left to False.
+        # workaround for RHBZ#1669012 and RHBZ#1695685
+        # Improve performance for service, host and user case by also
+        # searching for subject. This limits the amount of certificate
+        # retrieved from Dogtag. The special case is only used, when
+        # no ra_options are set and exactly one service, host, or user is
+        # supplied.
+        # IPA enforces that subject CN is either a hostname or a username.
+        # The complete flag is left to False to catch overrides.
         if not ra_options:
             services = options.get('service', ())
             hosts = options.get('host', ())
-            if len(services) == 1 and not hosts:
-                principal = kerberos.Principal(options['service'][0])
+            users = options.get('user', ())
+            if len(services) == 1 and not hosts and not users:
+                principal = kerberos.Principal(services[0])
                 if principal.is_service:
                     ra_options['subject'] = principal.hostname
-            elif len(hosts) == 1 and not services:
-                ra_options['subject'] = options['host'][0]
+            elif len(hosts) == 1 and not services and not users:
+                ra_options['subject'] = hosts[0]
+            elif len(users) == 1 and not services and not hosts:
+                ra_options['subject'] = users[0]
 
         try:
             ca_enabled_check(self.api)


rcritten:
This LGTM.

ipaserver/plugins/cert.py
```diff
@@ -1643,21 +1643,26 @@ def _ca_search(self, raw, pkey_only, exactly, **options):
...
-        # and exactly one service or host is supplied.
-        # The complete flag is left to False.
+        # workaround for RHBZ#1669012 and RHBZ#1695685
+        # Improve performance for service, host and user case by also
+        # searching for subject. This limits the amount of certificate
+        # retrieved from Dogtag. The special case is only used, when
+        # no ra_options are set and exactly one service, host, or user is
+        # supplied.
+        # IPA enforces that subject CN is either a hostname or a username.
+        # The complete flag is left to False to catch overrides.
```
frasertweedale: This code (including the existing workaround) makes some assumptions that I'm not 100% sure about - in particular that the hostname/username is used as the CN in the issued cert (a custom profile may violate this assumption).  I need to work through it in more detail, and have run out of week, so I'll resume early next week.

tiran: I'm confident that the restriction 
https://github.com/freeipa/freeipa/blob/350954589774499d99bf87cb5631c664bb0707c4/ipaserver/plugins/cert.py#L747-L753 won't allow a custom profile to create certs that do have a different CN.

rcritten: Agreed. I had the same concerns as @frasertweedale but after looking at the code I don't see a way for an IPA-issued cert to not have cn=uid in the subject.

frasertweedale: @tiran @rcritten it is actually pretty easy - you just import a profile configuration with parameter like:
```
policyset.serverCertSet.1.default.params.name=UID=$request.req_subject_name.cn$, O=EXMAPLE.COM
```

The CSR comes in with `CN=alice`, and comes out with `UID=alice` (but of course how you configure the subject DN to look is arbitrary).

tiran: @frasertweedale Oh no :/

We can either decide to ignore this edge case or somebody has to redesign and rewrite the entire ``cert_find`` API. The current implementation doesn't scale at all. Several common cases download **all** certificates from Dogtag, which puts a lot of stress on 389-DS, Dogtag, and the Python API. The BZ mentions > 100k certs.

I have an idea how to optimize the most critical cases for cert revocation. We can use the fact that host certs, service certs, and user certs are always stored on the host, service, or user LDAP entry. For simple host, service, and user cert search, perform ``_ldap_search`` before ``_ca_search``. In ``_ca_search`` only retrieve certificates with subject CNs or serial numbers of certs as returned by ``_ldap_search``.

There is one catch: It looks like Dogtag does not support list of serial numbers. :/

ipaserver/plugins/cert.py
```diff
@@ -1643,21 +1643,26 @@ def _ca_search(self, raw, pkey_only, exactly, **options):
...
         if not ra_options:
             services = options.get('service', ())
             hosts = options.get('host', ())
-            if len(services) == 1 and not hosts:
-                principal = kerberos.Principal(options['service'][0])
+            users = options.get('users', ())
```
frasertweedale: This should be `options.get('user', ())` (the key in `options` is `'user'`).

However with this correction in place, it is clear that the code regresses in the subject DN corner case I [mentioned above](https://github.com/freeipa/freeipa/pull/2990/files#r272901851).

So... I am considering NAKing this.  We might need to go back to the drawing board on `cert-find`.

OTOH the scenario that breaks this code is a corner case.  It's not the end of the world if we ship this patch as-is.  `cert-find` has a lot of limitations as it is, and this patch may help in the common case.

ipaserver/plugins/cert.py
frasertweedale: We can add support on the Dogtag side for that.  I'm inclined to ACK this, given the problem only occurs on an edge case, and file tickets to rework the whole thing.

frasertweedale: ... once you fix the `'users'` / `'user'` issue that is :)

ipaserver/plugins/cert.py
tiran: Thanks for the view and good catch! I have updated the patch.

Do we already have a ticket to track the comments? GH has a tendency to eat comments on rebased PRs.

frasertweedale: @tiran I filed https://pagure.io/freeipa/issue/7904 to capture the comments and discuss re-working cert-find and friends.

frasertweedale:
I'm gonna ACK this and we can address the corner case later.
See https://pagure.io/freeipa/issue/7904

tiran:
master:

* 8a5dc1b375db94c4e722fa725f48eb16d032f1aa Adapt cert-find performance workaround for users

PR #2992, opened 2019-04-04
PoC for stable, minimal Python API

tiran:


JAORMX:
From a consumer point of view, this covers most of our needs in OpenStack (talking about the novajoin project). We do have some extra stuff that's not covered here, but we could start refactoring those usages to get rid of them. So, @tiran , we would certainly appreciate if you and/or @rcritten would guide us in getting rid of those instances for novajoin. If you have time, of course.

PR #2996, opened 2019-04-05
Use Network Manager to configure resolv.conf

tiran:
IPA used to write a custom /etc/resolv.conf. On Fedora and RHEL,
NetworkManager is typically maintaining resolv.conf. On reboot or
restart of the service, NM overwrites the custom settings.

The DNS server installer now drops a config file into NM's global config
directory and delegates resolv.conf to NM.

Fixes: https://pagure.io/freeipa/issue/7900


t-woerner:
LGTM

tiran:
master:

* 80928ba6f5872aab47a3dd5eb361336568dfcc60 Use Network Manager to configure resolv.conf

PR #2999, opened 2019-04-06
Remove DsInstance.request_service_keytab as it is not needed anymore

abbra:
DsInstance.request_service_keytab() used to configure
/etc/sysconfig/dirsrv which is not needed anymore with 389-ds-base
1.4.1.2. Thus, the method became indistinguishable from the parent and
can be removed completely.

### DIFF
--- a/ipaserver/install/dsinstance.py
+++ b/ipaserver/install/dsinstance.py
@@ -1352,14 +1352,6 @@ class DsInstance(service.Service):
         if self.domainlevel is not None:
             self._ldap_mod("domainlevel.ldif", self.sub_dict)
 
-    def request_service_keytab(self):
-        super(DsInstance, self).request_service_keytab()
-
-        # Configure DS to use the keytab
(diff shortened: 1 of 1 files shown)


abbra:
Note that without this patch Fedora 30 is broken because `/etc/sysconfig/dirsrv` already removed in 389-ds-base 1.4.1.2 and we simply fail in `DsInstance.request_service_keytab()` when trying to write down the values to `/etc/sysconfig/dirsrv`. Since we already write a systemd snippet with required information before this step, we actually don't need to access `/etc/sysconfig/dirsrv` at all. Thus, the function removal should work fine on both old and new 389-ds-base.

tiran:
LGTM,

This function should have been removed in commit 1dfac4f5b7e9986e1217d6979b854b94de094aae. Looks like I forgot to adjust the code.

tiran:
master:

* 0c50cc3956cafaece7f00279ddd91d2f9f145827 Remove DsInstance.request_service_keytab as it is not needed anymore

PR #3001, opened 2019-04-08
Make netifaces optional

tiran:
netifaces is a binary Python extension. Outside of the installer, it's
only used by CheckedIPAddress.get_matching_interface, which is only
called from installer code.

Make the import of netifaces optional to reduce the amount of
dependencies for PyPI package use case. Binary extensions are especially
annoying, because they depend on shared libraries, compiler, and header
files to be present.

Related: https://pagure.io/freeipa/issue/6468

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -34,7 +34,6 @@ import socket
 import re
 import datetime
 import netaddr
-import netifaces
 import time
 import pwd
 import grp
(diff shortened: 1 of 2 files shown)


tiran:
Build was failing with ``fatal: unable to access 'https://github.com/freeipa/freeipa/': Could not resolve host: github.com``

tiran:
Thanks for the review!

tiran:
master:

* a5213140c3abf67a341b04d6edaeea6e0e1de62e Make netifaces optional

PR #3002, opened 2019-04-08
automount: rmtree temp directory

tiran:
ipa-client-automount uses the host keytab to acquire a TGT. The script
sets up a temporary directory for its ccache. At the end of the script
it removes the ccache and temp directory again.

In case of a failed kinit, the ccache is not created and the removal of
the ccache causes an exception. The automount installer now uses
shutil.rmtree() to remove the temporary directory and all its content.

Fixes: https://pagure.io/freeipa/issue/7862

### DIFF
--- a/client/ipa-client-automount.in
+++ b/client/ipa-client-automount.in
@@ -26,6 +26,7 @@ from __future__ import print_function
 import logging
 import sys
 import os
+import shutil
 import time
 import tempfile
 import gssapi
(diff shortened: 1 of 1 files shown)


fcami:
Hi @tiran Good catch, LGTM. Will run and provide review.

fcami:
ACK.

fcami:
master:

* 6fed17003d7824dda27253b461616529de07c333 automount: rmtree temp directory

fcami:
Removing ipa-4-6 label as there is no value and backporting would be manual.

tiran:
It's not necesary to backport this to 4.6. There is no related RHEL BZ.

PR #3004, opened 2019-04-08
Skip automember rebuild orphan tests for 389-ds with unmembering feature

abbra:
Fixes: https://pagure.io/freeipa/issue/7902

### DIFF
--- a/ipatests/test_xmlrpc/test_automember_plugin.py
+++ b/ipatests/test_xmlrpc/test_automember_plugin.py
@@ -34,6 +34,7 @@ from ipatests.util import assert_deepequal
 
 import pytest
 import unittest
+from lib389.utils import ds_is_older
 
 try:
     from ipaserver.plugins.ldap2 import ldap2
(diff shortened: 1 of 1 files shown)


ipatests/test_xmlrpc/test_automember_plugin.py
```diff
@@ -34,6 +34,7 @@
 
 import pytest
 import unittest
+from lib389.utils import ds_is_older
```
tiran: We have to backport the patch to IPA 4.7 but we cannot use ``lib389`` in 4.7 because the branch still supports Python 2 and ``lib389`` is not available for Python 2.

Also this will check the version that is installed on the test runner host, not the version on the host that executes the command. You can use ``ipaserver.install.replication.get_ds_version`` to fetch the version number from the Root DSE.

ipatests/test_xmlrpc/test_automember_plugin.py
```diff
@@ -838,22 +839,26 @@ def test_find_orphan_automember_rules(self, hostgroup1):
...
-        assert result['count'] == 1
-
-        # Find and remove obsolete automember rules
-        result = api.Command['automember_find_orphans'](type=u'hostgroup',
-                                                        remove=True)
-        assert result['count'] == 1
+        # Only run this test for 389-ds older than 1.4.0.22 where unmembering
+        # feature was implemented: https://pagure.io/389-ds-base/issue/50077
+        # As of today (2019-04-08), 1.3.9 is not released with this change yet
+        if ds_is_older('1.4.0.22'):
```
tiran: Does it makes sense to verify that ``api.Command['automember_rebuild'](type=u'hostgroup')`` does not fail with >= 1.4.0.22?

tiran:
Do we need to check the version at all? You could bump up requirement to 389-ds-base >= 1.4.0.22.

ipatests/test_xmlrpc/test_automember_plugin.py
t-woerner: I think this is required to make sure that there is really no orphan automember rule after skipping the removal test.

tiran:
Test is now failing in:

```
        # Find obsolete automember rules
        result = api.Command['automember_find_orphans'](type=u'hostgroup')
>       assert result['count'] == 0
E       assert 1 == 0
```

t-woerner:
That means that orphan automember rules are not removed automatically. https://pagure.io/389-ds-base/issue/50077 gives the impression that that this would happen.

tiran:
The issue at hand needs more engineering resources. I have created https://github.com/freeipa/freeipa/pull/3005 to skip the test for now. Please remove the skip in this PR.

flo-renaud:
PR superseded by PR #3413

PR #3005, opened 2019-04-08
Skip orphan automember rule test

tiran:
389-DS 1.4.0.22 was pushed to Fedora over the weekend. The new versin
breaks test_find_orphan_automember_rules. Skip the test case for now
until we have more time to investigate the issue.

Related: https://pagure.io/freeipa/issue/7902

### DIFF
--- a/ipatests/test_xmlrpc/test_automember_plugin.py
+++ b/ipatests/test_xmlrpc/test_automember_plugin.py
@@ -831,6 +831,7 @@ class TestAutomemberFindOrphans(XMLRPC_test):
 
         hostgroup1.retrieve()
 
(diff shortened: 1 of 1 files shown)


t-woerner:
LGTM
But it surely needs to be a temporary solution.

pvoborni:
shoudn't we use `pytest.mark.xfail` instead of skip ?

tiran:
@pvoborni This PR is just a temporary workaround to unblock the PR pipeline. I want to keep it as simple as possible. The test only fails with most recent version of 389-DS.

PR #3004 is still open and will remove the skip once we have figured out the problem.

tiran:
master:

* 3fdbf48f3df52fb503ae91321520c83ac4b55f0d Skip orphan automember rule test

PR #3008, opened 2019-04-08
[Backport ipa-4-6] Bypass D-BUS interface definition deficiences for trust-fetch-domains

abbra:
In oddjobd it is possible to pass arguments as command line or on the
stdin. We use command line to pass them but the way oddjobd registers
the D-BUS method signatures is by specifying all arguments as mandatory.

Internally, oddjobd simply ignores if you passed less arguments than
specified in the D-BUS defition. Unfortunately, it is not possible to
specify less than maximum due to D-BUS seeing all arguments in the
list (30 is defined for the trust-fetch-domains).

To pass options, have to pad a list of arguments to maximum with empty
strings and then filter out unneeded ones in the script. Option parser
already removes all options from the list of arguments so all we need to
do is to take our actual arguments. In case of trust-fetch-domains, it
is the name of the domain so we can only care about args[0].

Fixes: https://pagure.io/freeipa/issue/7903
(cherry picked from commit add6180ae5c5771b0b0f1c743df069ece4256512)

### DIFF
--- a/install/oddjob/com.redhat.idm.trust-fetch-domains
+++ b/install/oddjob/com.redhat.idm.trust-fetch-domains
@@ -5,6 +5,7 @@ from ipaserver.install.installutils import is_ipa_configured, ScriptError
 from ipapython import config, ipautil
 from ipalib import api
 from ipapython.dn import DN
+from ipapython.dnsutil import DNSName
 from ipaplatform.constants import constants
 from ipaplatform.paths import paths
 import sys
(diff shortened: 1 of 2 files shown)


tiran:
ipa-4-6:

* 51fdbf2ba1beb38eb8c1d54b02b956aedad06782 Bypass D-BUS interface definition deficiences for trust-fetch-domains

PR #3014, opened 2019-04-09
[Backport][ipa-4-7] Make netifaces optional

tiran:
Manual backport of PR #3001 

netifaces is a binary Python extension. Outside of the installer, it's
only used by CheckedIPAddress.get_matching_interface, which is only
called from installer code.

Make the import of netifaces optional to reduce the amount of
dependencies for PyPI package use case. Binary extensions are especially
annoying, because they depend on shared libraries, compiler, and header
files to be present.

Related: https://pagure.io/freeipa/issue/6468
Reviewed-By: Oleg Kozlov <okozlov@redhat.com>

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -34,7 +34,6 @@ import socket
 import re
 import datetime
 import netaddr
-import netifaces
 import time
 import pwd
 import grp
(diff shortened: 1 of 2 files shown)


tiran:
ipa-4-7:

* 12bfb86f0dd260f8f0df1a596bddf781e8e72465 Make netifaces optional

PR #3021, opened 2019-04-10
ipa-client-install: autodiscovery must refuse single-label domains

flo-renaud:
Since commit 905ab93, ipa-server-install refuses single-label domains,
but older IPA server versions could be installed with a single-label
domain/realm.
ipa-client-install is already refusing single-label domain/realm when
provided to the CLI with --domain / --realm but does not perform the same
check when the domain is discovered.
This commit adds a check to domain names automatically discovered and skips
single-label domains.

Fixes: https://pagure.io/freeipa/issue/7598

### DIFF
--- a/ipaclient/discovery.py
+++ b/ipaclient/discovery.py
@@ -27,6 +27,7 @@ import six
 from dns import resolver, rdatatype
 from dns.exception import DNSException
 from ipalib import errors
+from ipalib.util import validate_domain_name
 from ipapython.dnsutil import query_srv
 from ipapython import ipaldap
 from ipaplatform.paths import paths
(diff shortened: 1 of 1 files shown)


flo-renaud:
Note: I did not add any test since it needs to be manually tested (in order to have a master with single-label domain, one needs to install IPA <= 4.6.4, for instance RHEL 7.6).
Manual test:

- install IPA master with 4.6.4 with single-label domain and single-label realm
- configure the client machine with the same single-label domain, using the master DNS, with the patch: ipa-client-install --principal admin --password XXXX -d (do not specify domain/realm, so that autodiscovery is triggered)
- the client should output in the debug: `Skipping invalid domain 'mydomain' (single label domains are not supported)` and `DNS discovery failed to determine your DNS domain`, and prompt for the name of the IPA domain

rcritten:
Single-value realm is still allowed.

flo-renaud:
Hi @rcritten 
thanks for the comment, I forgot to add checks for the realm name. PR updated with the additional checks.

rcritten:
That fixed it.

2019-04-10T14:46:18Z DEBUG Skipping invalid realm 'EXAMPLE' (single label realms are not supported)

Patch LGTM, will add ack once CI is done.

rcritten:
master:

* f2826da24bb4ddcc55a3c8d6739e918b0089b246 ipa-client-install: autodiscovery must refuse single-label domains

rcritten:
Manual backports are needed for ipa-4-6 and ipa-4-7 due to moving and renaming of discovery.py

PR #3022, opened 2019-04-10
ipaserver/install/krainstance.py: chown after write

fcami:
When fs.protected_regular=1 root cannot open temp files that
are owned by other users read-write.
So move os.chown after write.

Fixes: https://pagure.io/freeipa/issue/7906

### DIFF
--- a/ipaserver/install/krainstance.py
+++ b/ipaserver/install/krainstance.py
@@ -146,11 +146,6 @@ class KRAInstance(DogtagInstance):
         parameters and passes it to the base class to call pkispawn
         """
 
-        # Create an empty and secured file
-        (cfg_fd, cfg_file) = tempfile.mkstemp()
-        os.close(cfg_fd)
-        pent = pwd.getpwnam(self.service_user)
-        os.chown(cfg_file, pent.pw_uid, pent.pw_gid)
         self.tmp_agent_db = tempfile.mkdtemp(
                 prefix="tmp-", dir=paths.VAR_LIB_IPA)
         tmp_agent_pwd = ipautil.ipa_generate_password()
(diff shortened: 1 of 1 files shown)


tiran:
May I propose that you use the opportunity and replace the ugly code with a nice solution?

```
        # Generate configuration file
        pent = pwd.getpwnam(self.service_user)
        with tempfile.NamedTemporaryFile('w', delete=False) as f:
            config.write(f)
            os.fchown(f.fileno(), pent.pw_uid, pent.pw_gid)
            cfg_file = f.name
```

fcami:
@tiran thanks for the proposal, I've removed the temp file creation from the top of the method and switched to yours.

fcami:
This fix has to be backported to ipa-4-7 so that f29/f28 containers running on f30 hosts do not fail.

ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -45,14 +45,14 @@ jobs:
...
-  fedora-29/temp_commit:
+  fedora-f29/test_vault:
     requires: [fedora-29/build]
     priority: 50
     job:
       class: RunPytest
       args:
-        build_url: '{fedora-29/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
+        build_url: '{fedora-f29/build_url}'
```
netoarmando: ` fedora-29/build`

netoarmando: PR-CI stuck reporting: 
```
Apr 10 20:27:35 ci-large-16 bash[6762]: ERROR: __main__: 'NoneType' object is not subscriptable
```

fcami: Thanks @netoarmando much appreciated.

fcami:
Link to vault test: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/8eb3a816-5c17-11e9-bd14-fa163ee50fe4/
Removing temp commit.

fcami:
master:

* cf42dc1f2930cba3fca144ad4b7c1c01e9ed9163 ipaserver/install/krainstance.py: chown after write

PR #3024, opened 2019-04-10
ipaserver/install/cainstance.py: unlink before creating new file in /tmp

fcami:
When fs.protected_regular=1 root cannot open temp files that
are owned by other users read-write.
So unlink temporary file before shutil.copy to it.

Fixes: https://pagure.io/freeipa/issue/7907
Signed-off-by: François Cami fcami@redhat.com

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -69,6 +69,7 @@ from ipaserver.install import sysupgrade
 from ipaserver.install.dogtaginstance import DogtagInstance
 from ipaserver.plugins import ldap2
 from ipaserver.masters import ENABLED_SERVICE
+from ipaserver.install.installutils import remove_file
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


fcami:
This fix has to be backported to ipa-4-7 so that f29/f28 containers running on f30 hosts do not fail.

ipaserver/install/cainstance.py
```diff
@@ -514,6 +515,15 @@ def __spawn_instance(self):
...
             cafile = self.pkcs12_info[0]
+
+            # if paths.TMP_CA_P12 exists and is not owned by root,
+            # shutil.copy will fail if when fs.protected_regular=1
+            # so remove the file first
+            try:
+                os.unlink(paths.TMP_CA_P12)
+            except OSError as e:
+                if e.errno != errno.ENOENT:
+                    raise
```
rcritten: You can use remove_file() from ipaserver.install.installutils instead.

PR #3025, opened 2019-04-10
Fix unnecessary usrmerge assumptions

frozencemetery:
On non-usrmerge systems (e.g., Debian), bash, mv, cp, cat, tail,
keyctl, and gzip live in /bin, not /usr/bin.

On usrmerge systems, /bin is a symlink to /usr/bin (or vice versa), so
this has no effect.

### DIFF
--- a/ipaplatform/base/paths.py
+++ b/ipaplatform/base/paths.py
@@ -26,7 +26,7 @@ class BasePathNamespace:
     BASH = "/bin/bash"
     BIN_HOSTNAMECTL = "/bin/hostnamectl"
     ECHO = "/bin/echo"
-    GZIP = "/usr/bin/gzip"
+    GZIP = "/bin/gzip"
     LS = "/bin/ls"
     SH = "/bin/sh"
     SYSTEMCTL = "/bin/systemctl"
@@ -390,7 +390,7 @@ class BasePathNamespace:
     BAK2DB = '/usr/sbin/bak2db'
     DB2BAK = '/usr/sbin/db2bak'
     IPA_SERVER_UPGRADE = '/usr/sbin/ipa-server-upgrade'
-    KEYCTL = '/usr/bin/keyctl'
+    KEYCTL = '/bin/keyctl'
     GETENT = '/usr/bin/getent'
     SSHD = '/usr/sbin/sshd'
     SSSCTL = '/usr/sbin/sssctl'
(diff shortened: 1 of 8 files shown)


rcritten:
Either the duplicates should be removed from ipaplatform/debian/paths.py or all of the overrides added to it instead.

Also makes me wonder why mv, cp, cat and tail aren't also handled via ipaplatform.

tiran:
Test is failing on Fedora:

```
_________________________ test_get_pager[cat-/bin/cat] _________________________
pager = 'cat', expected_result = '/bin/cat'
    @pytest.mark.parametrize('pager,expected_result', [
        # Valid values
        ('cat', '/bin/cat'),
        ('/bin/cat', '/bin/cat'),
        # Invalid values (wrong command, package is not installed, etc)
        ('cat_', None),
        ('', None)
    ])
    def test_get_pager(pager, expected_result):
        with mock.patch.dict(os.environ, {'PAGER': pager}):
>           assert get_pager() == expected_result
E           AssertionError: assert '/usr/bin/cat' == '/bin/cat'
E             - /usr/bin/cat
E             ? ----
E             + /bin/cat
```

frozencemetery:
> Either the duplicates should be removed from ipaplatform/debian/paths.py or all of the overrides added to it instead.

That only affects KEYCTL (updated); GZIP wasn't handled in debian/paths.py (probably a bug waiting to happen).

> Test is failing on Fedora:

Should be fixed.  I still don't understand why you're testing the return value of a dict lookup, though.

rcritten:
Test is failing in a new way:

    def test_get_pager(pager, expected_result):
        with mock.patch.dict(os.environ, {'PAGER': pager}):
>           assert get_pager().endswith(expected_result)
E           AttributeError: 'NoneType' object has no attribute 'endswith'

frozencemetery:
Yay None.  Updated.

tiran:
Almost :)

```
_________________________ test_get_pager[cat-/bin/cat] _________________________
pager = '/usr/bin/cat', expected_result = '/bin/cat'
    @pytest.mark.parametrize('pager,expected_result', [
        # Valid values
        ('cat', '/bin/cat'),
        ('/bin/cat', '/bin/cat'),
        # Invalid values (wrong command, package is not installed, etc)
        ('cat_', None),
        ('', None)
    ])
    def test_get_pager(pager, expected_result):
        with mock.patch.dict(os.environ, {'PAGER': pager}):
            pager = get_pager()
>           assert(pager == expected_result or get_pager.endswith(expected_result))
E           AttributeError: 'function' object has no attribute 'endswith'
lib/python3.6/site-packages/ipatests/test_ipalib/test_util.py:26: AttributeError
```

frozencemetery:
Tests green now.  I'm gonna go back to writing C :)

tiran:
master:

* 5ecede781b4eba11a537afd104fc9df9518958c4 Fix unnecessary usrmerge assumptions

PR #3035, opened 2019-04-12
Fix wrong evaluation of attributes in check_repl_update

flo-renaud:
The method check_repl_update in ipaserver/install/replication.py badly
handles the attributes nsds5ReplicaLastUpdateStart and
nsds5ReplicaLastUpdateEnd as it expects them to contain an int.
These attributes are defined as GeneralizedTime
(OID 1.3.6.1.4.1.1466.115.121.1.24, for instance
nsds5ReplicaLastUpdateEnd='20190412122523Z') but older versions of 389-ds can
also return the value 0 for uninitialized values (see 389-ds ticket 47836).
The code must be able to handle the generalized time format or the 0 value.

The fix removes the 'Z' from the GeneralizedTime and converts to an int,
or assigns 0.

Fixes: https://pagure.io/freeipa/issue/7909

### DIFF
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -1070,11 +1070,23 @@ class ReplicationManager:
             inprogress = entry.single_value.get('nsds5replicaUpdateInProgress')
             status = entry.single_value.get('nsds5ReplicaLastUpdateStatus')
             try:
-                start = int(entry.single_value['nsds5ReplicaLastUpdateStart'])
+                # nsds5ReplicaLastUpdateStart is either a GMT time
+                # ending with Z or 0 (see 389-ds ticket 47836)
+                # Remove the Z and convert to int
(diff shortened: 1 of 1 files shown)


fcami:
Hi @flo-renaud 
Good catch!

Affected attributes by the referenced DS issue::

- nsds5ReplicaLastUpdateStart and nsds5ReplicaLastUpdateEnd 
- nsds5replicaLastInitStart and nsds5replicaLastInitEnd 

Only nsds5ReplicaLastUpdateStart and nsds5ReplicaLastUpdateEnd are used in FreeIPA.
Fix LGTM, ACK.

fcami:
master:

* b79ea6a6e92d224859c2d4533b99b3c15d3711ca Fix wrong evaluation of attributes in check_repl_update

PR #3036, opened 2019-04-12
New command logger logcm as a replacement of print calls (updated)

t-woerner:
New command logger logcm as a replacement of print calls

The use of print calls in the installer parts are resulting in the need to
redirect stdout using redirect_stdout in ansible-freeipa. redirect_stdout
is creating a new scope.

The CommandOutput type is a wrapper for a logger instance. It marks log
calls to be printed out on stdout using logger.INFO level. The extra
argument { "cmd_output": True } is set to make sure that a filter could
be used to filter this out for the installer log files.

Usage:

    import logging
    from ipapython.ipa_log_manager import CommandOutput

    logger = logging.getLogger(__name__)
    logcm = CommandOutput(logger)

    def install():
        logcm("Installing ...")

Design note
-----------

The command logger instance is called ``logcm``, because the name has the
same length as the ``print`` function. It may not be the prettiest name,
but it avoids code reformatting and therefore merge conflicts.

At first we considered to add a ``log_cmd`` method to all loggers.
The ``logging.setLoggerClass`` is the public function to set a new logger
class. To make it work, is has to be called before the first logger
objects is instantiated. It's not appropriate to modify the logger
early in ``ipalib``, because ``ipalib`` is also designed to get embedded
into applications. And monkey-patching the ``logging.Logger`` class is
a hack.

The log format for CommandOutput is defined in LOGGING_FORMAT_CMD_OUTPUT
and is applied with the extended Formatter class for the log level INFO
only.

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -9,7 +9,6 @@ Provides methods for installation, uninstallation of IPA client
 """
 
 from __future__ import (
-    print_function,
     absolute_import,
 )
 
@@ -60,6 +59,7 @@ from ipapython.ipautil import (
     run,
     user_input,
 )
+from ipapython.ipa_log_manager import CommandOutput
 from ipapython.ssh import SSHPublicKey
 from ipapython import version
 
@@ -70,6 +70,7 @@ from .ipachangeconf import IPAChangeConf
 NoneType = type(None)
 
 logger = logging.getLogger(__name__)
+logcm = CommandOutput(logger)
 
 SUCCESS = 0
 CLIENT_INSTALL_ERROR = 1
@@ -2055,9 +2056,9 @@ def install_check(options):
     global client_domain
     global cli_basedn
 
-    print("This program will set up FreeIPA client.")
-    print("Version {}".format(version.VERSION))
-    print("")
+    logcm("This program will set up FreeIPA client.")
+    logcm("Version %s", version.VERSION)
+    logcm("")
 
     cli_domain_source = 'Unknown source'
     cli_server_source = 'Unknown source'
@@ -2084,10 +2085,10 @@ def install_check(options):
         try:
             timeconf.check_timedate_services()
         except timeconf.NTPConflictingService as e:
-            print("WARNING: conflicting time&date synchronization service '{}'"
-                  " will be disabled".format(e.conflicting_service))
-            print("in favor of chronyd")
-            print("")
+            logcm("WARNING: conflicting time&date synchronization service '%s'"
+                  " will be disabled", e.conflicting_service)
+            logcm("in favor of chronyd")
+            logcm("")
         except timeconf.NTPConfigurationError:
             pass
 
@@ -2406,14 +2407,14 @@ def install_check(options):
                 is_ipaddr = False
 
         if is_ipaddr:
-            print()
+            logcm("")
             logger.warning(
                 "It seems that you are using an IP address "
                 "instead of FQDN as an argument to --server. The "
                 "installation may fail.")
             break
 
-    print()
+    logcm("")
     if not options.unattended and not user_input(
             "Continue to configure the system with these values?", False):
         raise ScriptError(rval=CLIENT_INSTALL_ERROR)
@@ -2491,7 +2492,7 @@ def sync_time(options, fstore, statestore):
                        "or pool address was provided.")
 
     if not configured:
-        print("Using default chrony configuration.")
+        logcm("Using default chrony configuration.")
 
     return timeconf.sync_chrony()
 
--- a/ipapython/ipa_log_manager.py
+++ b/ipapython/ipa_log_manager.py
@@ -22,7 +22,7 @@ import re
 import time
 
 # Module exports
-__all__ = ['standard_logging_setup',
+__all__ = ['standard_logging_setup', 'CommandOutput',
            'ISO8601_UTC_DATETIME_FMT',
            'LOGGING_FORMAT_STDERR', 'LOGGING_FORMAT_STDOUT', 'LOGGING_FORMAT_FILE']
 
@@ -52,6 +52,55 @@ LOGGING_FORMAT_STANDARD_CONSOLE = '%(name)-12s: %(levelname)-8s %(message)s'
 # Used by standard_logging_setup() for file message
 LOGGING_FORMAT_STANDARD_FILE = '%(asctime)s %(levelname)s %(message)s'
 
+# Used by standard_logging_setup() for command line output
+LOGGING_FORMAT_CMD_OUTPUT = '%(message)s'
+
+
+class CommandOutput:
+    """Command output -- log and print to user
+
+    The CommandOutput type is a wrapper for a logger instance. It
+    marks log calls to be printed out on stdout.
+
+    Usage:
+
+        import logging
+        from ipapython.ipa_log_manager import CommandOutput
+
+        logger = logging.getLogger(__name__)
+        logcm = CommandOutput(logger)
+
+        def install():
+            logcm("Installing ...")
+
+    Design note
+    -----------
+
+    The command logger instance is called ``logcm``, because the name has the
+    same length as the ``print`` function. It may not be the prettiest name,
+    but it avoids code reformatting and therefore merge conflicts.
+
+    At first we considered to add a ``log_cmd`` method to all loggers.
+    The ``logging.setLoggerClass`` is the public function to set a new logger
+    class. To make it work, is has to be called before the first logger
+    objects is instantiated. It's not appropriate to modify the logger
+    early in ``ipalib``, because ``ipalib`` is also designed to get embedded
+    into applications. And monkey-patching the ``logging.Logger`` class is
+    a hack.
+    """
+    __slots__ = ('_logger',)
+
+    def __init__(self, logger):
+        self._logger = logger
+
+    def __call__(self, msg, *args, **kwargs):
+        if 'extra' not in kwargs:
+            kwargs["extra"] = {"cmd_output": True}
+        else:
+            kwargs["extra"]["cmd_output"] = True
+        level = kwargs.pop("level", logging.INFO)
+        self._logger.log(level, msg, *args, **kwargs)
+
 
 class Filter:
     def __init__(self, regexp, level):
@@ -68,6 +117,29 @@ class Formatter(logging.Formatter):
             self, fmt=LOGGING_FORMAT_STDOUT, datefmt=ISO8601_UTC_DATETIME_FMT):
         super(Formatter, self).__init__(fmt, datefmt)
         self.converter = time.gmtime
+
+
+class ConsoleFormatter(logging.Formatter):
+    def __init__(
+            self, fmt=LOGGING_FORMAT_STDOUT, datefmt=ISO8601_UTC_DATETIME_FMT):
+        super().__init__(fmt, datefmt)
+        self.converter = time.gmtime
+        self.cmd_output_formatter = logging.Formatter(
+            LOGGING_FORMAT_CMD_OUTPUT)
+
+    def format(self, record):
+        """
+        Format the specified record as text.
+
+        Uses LOGGING_FORMAT_CMD_OUTPUT for info_cmd, otherwise normal logging
+        format. For info_cmd record.__dict__["cmd_output"] is set to True.
+        """
+        if record.levelno >= logging.INFO and \
+           getattr(record, "cmd_output", False):
+            return self.cmd_output_formatter.format(record)
+
+        # Return normal format
+        return super().format(record)
 
 
 def standard_logging_setup(filename=None, verbose=False, debug=False,
@@ -97,7 +169,7 @@ def standard_logging_setup(filename=None, verbose=False, debug=False,
 
     console_handler = logging.StreamHandler()
     console_handler.setLevel(level)
-    console_handler.setFormatter(Formatter(console_format))
+    console_handler.setFormatter(ConsoleFormatter(console_format))
     root_logger.addHandler(console_handler)
 
 
--- a/ipaserver/install/adtrust.py
+++ b/ipaserver/install/adtrust.py
@@ -6,7 +6,7 @@
 AD trust installer module
 """
 
-from __future__ import print_function, absolute_import
+from __future__ import absolute_import
 
 import logging
 import os
@@ -21,6 +21,7 @@ from ipaplatform.paths import paths
 from ipapython.admintool import ScriptError
 from ipapython import ipaldap, ipautil
 from ipapython.dn import DN
+from ipapython.ipa_log_manager import CommandOutput
 from ipapython.install.core import group, knob
 from ipaserver.install import adtrustinstance
 from ipaserver.install import service
@@ -30,6 +31,7 @@ if six.PY3:
     unicode = str
 
 logger = logging.getLogger(__name__)
+logcm = CommandOutput(logger)
 
 netbios_name = None
 reset_netbios_name = False
@@ -45,18 +47,18 @@ def netbios_name_error(name):
 def read_netbios_name(netbios_default):
     netbios_name = ""
 
-    print("Enter the NetBIOS name for the IPA domain.")
-    print("Only up to 15 uppercase ASCII letters, digits "
+    logcm("Enter the NetBIOS name for the IPA domain.")
+    logcm("Only up to 15 uppercase ASCII letters, digits "
           "and dashes are allowed.")
-    print("Example: EXAMPLE.")
-    print("")
-    print("")
+    logcm("Example: EXAMPLE.")
+    logcm("")
+    logcm("")
     if not netbios_default:
         netbios_default = "EXAMPLE"
     while True:
         netbios_name = ipautil.user_input(
             "NetBIOS domain name", netbios_default, allow_empty=False)
-        print("")
+        logcm("")
         if adtrustinstance.check_netbios_name(netbios_name):
             break
 
@@ -108,16 +110,16 @@ def set_and_check_netbios_name(netbios_name, unattended, api):
         reset_netbios_name = False
     elif cur_netbios_name and cur_netbios_name != netbios_name:
         # change the NetBIOS name
-        print("Current NetBIOS domain name is %s, new name is %s.\n"
+        logcm("Current NetBIOS domain name is %s, new name is %s.\n"
               % (cur_netbios_name, netbios_name))
-        print("Please note that changing the NetBIOS name might "
+        logcm("Please note that changing the NetBIOS name might "
               "break existing trust relationships.")
         if unattended:
             reset_netbios_name = True
-            print("NetBIOS domain name will be changed to %s.\n"
+            logcm("NetBIOS domain name will be changed to %s.\n"
                   % netbios_name)
         else:
-            print("Say 'yes' if the NetBIOS shall be changed and "
+            logcm("Say 'yes' if the NetBIOS shall be changed and "
                   "'no' if the old one shall be kept.")
             reset_netbios_name = ipautil.user_input(
                             'Do you want to reset the NetBIOS domain name?',
@@ -134,7 +136,7 @@ def set_and_check_netbios_name(netbios_name, unattended, api):
 
         if gen_netbios_name is not None:
             # Fix existing trust configuration
-            print("Trust is configured but no NetBIOS domain name found, "
+            logcm("Trust is configured but no NetBIOS domain name found, "
                   "setting it now.")
             reset_netbios_name = True
         else:
@@ -163,16 +165,16 @@ def set_and_check_netbios_name(netbios_name, unattended, api):
 
 
 def enable_compat_tree():
-    print("Do you want to enable support for trusted domains in Schema "
+    logcm("Do you want to enable support for trusted domains in Schema "
           "Compatibility plugin?")
-    print("This will allow clients older than SSSD 1.9 and non-Linux "
+    logcm("This will allow clients older than SSSD 1.9 and non-Linux "
           "clients to work with trusted users.")
-    print("")
+    logcm("")
     enable_compat = ipautil.user_input(
         "Enable trusted domains support in slapi-nis?",
         default=False,
         allow_empty=False)
-    print("")
+    logcm("")
     return enable_compat
 
 
@@ -221,21 +223,21 @@ def retrieve_and_ask_about_sids(api, options):
 
     object_count = len(entries)
     if object_count > 0:
-        print("")
-        print("WARNING: %d existing users or groups do not have "
+        logcm("")
+        logcm("WARNING: %d existing users or groups do not have "
               "a SID identifier assigned." % len(entries))
-        print("Installer can run a task to have ipa-sidgen "
+        logcm("Installer can run a task to have ipa-sidgen "
               "Directory Server plugin generate")
-        print("the SID identifier for all these users. Please note, "
+        logcm("the SID identifier for all these users. Please note, "
               "in case of a high")
-        print("number of users and groups, the operation might "
+        logcm("number of users and groups, the operation might "
               "lead to high replication")
-        print("traffic and performance degradation. Refer to "
+        logcm("traffic and performance degradation. Refer to "
               "ipa-adtrust-install(1) man page")
-        print("for details.")
-        print("")
+        logcm("for details.")
+        logcm("")
         if options.unattended:
-            print("Unattended mode was selected, installer will "
+            logcm("Unattended mode was selected, installer will "
                   "NOT run ipa-sidgen task!")
         else:
             if ipautil.user_input(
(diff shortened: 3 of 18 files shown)


abbra:
@t-woerner thanks for the patch. I have one small suggestion: in order to simplify maintenance backports, please split commits that convert individual files to separate commits. This way, for example, a commit that changes `ipaserver/install/server/upgrade.py` wouldn't create a conflict for changes in `ipaserver/install/dsinstance.py` and so on. With smaller commits backports will be automatic.

ipapython/ipa_log_manager.py
```diff
@@ -197,6 +250,12 @@ def convert_log_level(value):
     return level
 
 
+# Add info_cmd to logger and root logger
+if not hasattr(logging.getLoggerClass(), "info_cmd"):
+    logging.getLoggerClass().info_cmd = log_info_cmd
```
tiran: You are suppose to use [logging.setLoggerClass](https://docs.python.org/3/library/logging.html#logging.setLoggerClass) to define your own logger subclass.

ipapython/ipa_log_manager.py
```diff
@@ -54,6 +54,20 @@
 # Used by standard_logging_setup() for file message
 LOGGING_FORMAT_STANDARD_FILE = '%(asctime)s %(levelname)s %(message)s'
 
+# Used by standard_logging_setup() for command line output
+LOGGING_FORMAT_INFO_CMD = '%(message)s'
+
+
+# Additional info_log level methods
+def log_info_cmd(obj, msg, *args, **kwargs):
```
tiran: Please use self to indicate it's going to be a method:

``def log_info_cmd(self, msg, *args, **kwargs)``

ipapython/ipa_log_manager.py
```diff
@@ -54,6 +54,20 @@
...
+    kwargs.setdefault("extra", {})["cmd_output"] = True
+    obj._log(logging.INFO, msg, args, **kwargs)
```
tiran: Should the level be configurable with a default value of ``INFO``?

ipapython/ipa_log_manager.py
```diff
@@ -54,6 +54,20 @@
...
+
+
+def log_info_cmd_root(msg, *args, **kwargs):
```
tiran: We shouldn't use the root logger at all. If some module uses the root logger, then please fix the module instead.

ipapython/ipa_log_manager.py
```diff
@@ -141,12 +160,45 @@ def filter(self, record):
...
 
 
+class InfoCommandFilter(logging.Filter):
+    """
+    The InfoCommandFilter filters out info_cmd log messages for the log file.
+    info_cmd log messages should only be visble on the command line.
+    """
+
+    def filter(self, record):
+        if hasattr(record, "cmd_output") and record.__dict__["cmd_output"]:
```
tiran: This can be simplified to ``return getattr(record, "cmd_output", False)``. All extras are available as attributes on the record object.

ipapython/ipa_log_manager.py
```diff
@@ -141,12 +160,45 @@ def filter(self, record):
...
         self.converter = time.gmtime
 
+    def format(self, record):
+        """
+        Format the specified record as text.
+
+        Uses LOGGING_FORMAT_INFO_CMD for info_cmd, otherwise normal logging
+        format. For info_cmd record.__dict__["cmd_output"] is set to True.
+        """
+        if record.levelno == logging.INFO and \
```
tiran: IMHO we want to support cmd_output for all log levels >= INFO.

``if record.levelno >= logging.INFO and getattr(record, "cmd_output", False):``

ipapython/ipa_log_manager.py
```diff
@@ -141,12 +160,45 @@ def filter(self, record):
...
+           hasattr(record, "cmd_output") and record.__dict__["cmd_output"]:
+            # Save original logging format
+            orig_fmt = self._fmt
+            # Temporarily use LOGGING_FORMAT_INFO_CMD
+            self._fmt = LOGGING_FORMAT_INFO_CMD
+            # Create result
+            result = logging.Formatter.format(self, record)
```
tiran:
```
try:
    return super(Formatter, self).format(record)
finally:
    self._fmt = orig_fmt
```

ipapython/ipa_log_manager.py
```diff
@@ -141,12 +160,45 @@ def filter(self, record):
...
+
+        Uses LOGGING_FORMAT_INFO_CMD for info_cmd, otherwise normal logging
+        format. For info_cmd record.__dict__["cmd_output"] is set to True.
+        """
+        if record.levelno == logging.INFO and \
+           hasattr(record, "cmd_output") and record.__dict__["cmd_output"]:
+            # Save original logging format
+            orig_fmt = self._fmt
+            # Temporarily use LOGGING_FORMAT_INFO_CMD
+            self._fmt = LOGGING_FORMAT_INFO_CMD
```
tiran: Why is this hack necessary? Formatters are defined on the output handler. We should have a different output handler for stdout and for log files.

Also this is not thread safe. Handlers are shared across threads and multiple threads can log at the same time.

ipapython/ipa_log_manager.py
```diff
@@ -197,6 +250,12 @@ def convert_log_level(value):
     return level
 
 
+# Add info_cmd to logger and root logger
+if not hasattr(logging.getLoggerClass(), "info_cmd"):
+    logging.getLoggerClass().info_cmd = log_info_cmd
+if not hasattr(logging, "info_cmd"):
+    logging.info_cmd = log_info_cmd_root
```
tiran: This should not be necessary.

ipaclient/install/client.py
```diff
@@ -2055,9 +2055,9 @@ def install_check(options):
     global client_domain
     global cli_basedn
 
-    print("This program will set up FreeIPA client.")
-    print("Version {}".format(version.VERSION))
-    print("")
+    logger.info_cmd("This program will set up FreeIPA client.")
+    logger.info_cmd("Version {}".format(version.VERSION))
```
tiran: pylint is going to complain about this style. You have to use old-style formatting and replace this kind of calls with ``logger.info_cmd("Version %s", version.VERSION)``.

ipalib/cli.py
```diff
@@ -200,7 +200,7 @@ def print_plain(self, string):
         """
         Print exactly like ``print`` statement would.
         """
-        print(unicode(string))
+        logger.info_cmd(unicode(string))
```
tiran: While you are on it, kill ``unicode()`` in ``logger.info_cmd()`` commands

ipalib/messages.py
```diff
@@ -502,8 +505,8 @@ def iter_messages(variables, base):
 
 def print_report(label, classes):
     for cls in classes:
-        print('%d\t%s' % (cls.errno, cls.__name__))
-    print('(%d %s)' % (len(classes), label))
+        logger.info_cmd('%d\t%s' % (cls.errno, cls.__name__))
```
tiran: format

ipaserver/install/adtrust.py
```diff
@@ -358,10 +362,11 @@ def install_check(standalone, options, api):
     realm_not_matching_domain = (api.env.domain.upper() != api.env.realm)
 
     if realm_not_matching_domain:
-        print("WARNING: Realm name does not match the domain name.\n"
-              "You will not be able to establish trusts with Active "
-              "Directory unless\nthe realm name of the IPA server matches its "
-              "domain name.\n\n")
+        logger.info_cmd(
```
tiran: For this kind of stuff it should be possible to increase the log level to WARNING.

tiran (changes requested):

ipapython/ipa_log_manager.py
tiran: Update: This does not work with Python 3. The logging module uses a different approach with style formatters and ``formatMessage(self, record)``

tiran:
> @t-woerner thanks for the patch. I have one small suggestion: in order to simplify maintenance backports, please split commits that convert individual files to separate commits. This way, for example, a commit that changes `ipaserver/install/server/upgrade.py` wouldn't create a conflict for changes in `ipaserver/install/dsinstance.py` and so on. With smaller commits backports will be automatic.

I don't understand how this approach is going to reduce the amount of merge conflicts in backports. How do multiple commits reduce the amount of conflicts when none of the commits are cherry-picked into the backport branch?

tiran:
May I propose a slightly different approach that avoids monkey patching of classes and reduces the amount of code reformatting?

```
import logging

class CommandOutput:
    """Command output -- log and print to user

    The CommandOutput type is a wrapper for a logger instance. It
    marks log calls to be printed out on stdout.

    Usage:

        import logging
        from ipapython.ipa_log_manager import CommandOutput

        logger = logging.getLogger(__name__)
        logcm = CommandOutput(logger)

        def install():
            logcm("Installing ...")

    Design note
    -----------

    The command logger instance is called ``logcm``, because the name has the
    same length as the ``print`` function. It may not be the prettiest name,
    but it avoids code reformatting and therefore merge conflicts.

    At first we considered to add a ``log_cmd`` method to all loggers.
    The ``logging.setLoggerClass`` is the public function to set a new logger
    class. To make it work, is has to be called before the first logger
    objects is instantiated. It's not appropriate to modify the logger
    early in ``ipalib``, because ``ipalib`` is also designed to get embedded
    into applications. And monkey-patching the ``logging.Logger`` class is
    a hack.
    """
    __slots__ = ('_logger',)

    def __init__(self, logger):
        self._logger = logger

    def __call__(self, msg, *args, **kwargs):
        if 'extra' not in kwargs:
            kwargs["extra"] = {"cmd_output": True}
        else:
            kwargs["extra"]["cmd_output"] = True
        level = kwargs.pop("level", logging.INFO)
        self._logger.log(level, msg, *args, **kwargs)
```

ipalib/cli.py
tiran: ``ipalib.run`` implements the ``ipa`` command line tool. You have to keep all ``print()`` calls.

tiran:
Please only modify the installer code and unpatch the following files.

```
ipalib/cli.py
ipalib/messages.py
ipaserver/dnssec/ldapkeydb.py
ipaserver/dnssec/localhsm.py
```

``ipalib.cli`` implements the ``ipa`` command. With the current state, ``ipa`` does not print anything. The other modules use ``print()`` in ``__main__`` blocks, which are only used for development and debugging.

abbra:
> I don't understand how this approach is going to reduce the amount of merge conflicts in backports. How do multiple commits reduce the amount of conflicts when none of the commits are cherry-picked into the backport branch?

Suppose you have to backport a fix that was committed after these changes landed, in a proximity of the changes. When the changes split into smaller commits, a backport would result in porting a smaller base commit that adds logging support and the small patch from the affected area. If this would have been a single commit that touches multiple files, you would have to create a manual backport that will involve quite a lot of effort to make sure all pieces are not conflicting with something else.

ipapython/ipa_log_manager.py
t-woerner: This is not needed with the latest version anymore.

ipapython/ipa_log_manager.py
t-woerner: This is used any more with the latest version.

ipapython/ipa_log_manager.py
t-woerner: This is used any more with the latest version.

ipapython/ipa_log_manager.py
t-woerner: This is used any more with the latest version.

ipapython/ipa_log_manager.py
t-woerner: This is used any more with the latest version.

ipaclient/install/client.py
t-woerner: Fixed

ipalib/cli.py
t-woerner: cli.py is not patched anymore

ipalib/messages.py
t-woerner: messages.py is not patched anymore

ipaserver/install/adtrust.py
t-woerner: I think this should be an extra commit.

ipalib/cli.py
t-woerner: Fixed

ipapython/ipa_log_manager.py
t-woerner: It can be simplified to `return not getattr(record, "cmd_output", False)`.

tiran (changes requested):
There should be no no calls to ``.format()`` or `` % `` in ``logcm()``. I marked some of the calls, but there are many more.

ipapython/ipa_log_manager.py
t-woerner: Fixed

ipapython/ipa_log_manager.py
t-woerner: Solved in a different way with an special internal Formatter for cmd_output.

ipapython/ipa_log_manager.py
t-woerner: Solved by using an internal Formatter for cmd_output.

ipaclient/install/client.py
```diff
@@ -2055,9 +2056,9 @@ def install_check(options):
     global client_domain
     global cli_basedn
 
-    print("This program will set up FreeIPA client.")
-    print("Version {}".format(version.VERSION))
-    print("")
+    logcm("This program will set up FreeIPA client.")
+    logcm("Version %s" % version.VERSION)
```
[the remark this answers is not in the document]
t-woerner: Done

ipaserver/install/custodiainstance.py
```diff
@@ -227,10 +229,9 @@ def _wait_keys(self):
                 if saved_e is None:
                     # FIXME: Change once there's better way to show this
                     # message in installer output,
-                    print(
-                        "  Waiting for keys to appear on host: {}, please "
-                        "wait until this has completed.".format(
-                            self.ldap_uri)
+                    logcm(
+                        "  Waiting for keys to appear on host: %s, please "
+                        "wait until this has completed." % self.ldap_uri
```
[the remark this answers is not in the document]
t-woerner: Done

ipaserver/install/dns.py
```diff
@@ -216,7 +222,7 @@ def install_check(standalone, api, replica, options, hostname):
         dnssec_masters = ods.get_masters()
         # we can reinstall current server if it is dnssec master
         if dnssec_masters and api.env.host not in dnssec_masters:
-            print("DNSSEC key master(s):", u','.join(dnssec_masters))
+            logcm("DNSSEC key master(s): %s" % u','.join(dnssec_masters))
```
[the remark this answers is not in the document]
t-woerner: Done

ipaclient/install/client.py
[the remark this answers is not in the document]
t-woerner: Done

ipaserver/install/custodiainstance.py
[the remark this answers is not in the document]
t-woerner: Done

ipaserver/install/dns.py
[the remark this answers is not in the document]
t-woerner: Done

ipaserver/install/dsinstance.py
```diff
@@ -1052,7 +1054,7 @@ def change_admin_password(self, password):
                 ipautil.run(args, env=env)
                 logger.debug("ldappasswd done")
             except ipautil.CalledProcessError as e:
-                print("Unable to set admin password", e)
+                logcm("Unable to set admin password %s" % e)
```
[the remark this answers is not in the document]
t-woerner: Done

ipaserver/install/installutils.py
```diff
@@ -175,7 +177,7 @@ def verify_fqdn(host_name, no_host_dns=False, local_hostname=True):
                 e.errno, e.strerror)  # pylint: disable=no-member
 
     if no_host_dns:
-        print("Warning: skipping DNS resolution of host", host_name)
+        logcm("Warning: skipping DNS resolution of host %s" % host_name)
```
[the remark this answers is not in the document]
t-woerner: Done

ipaserver/install/installutils.py
```diff
@@ -278,7 +280,7 @@ def read_ip_addresses():
         try:
             ip_parsed = ipautil.CheckedIPAddress(ip)
         except Exception as e:
-            print("Error: Invalid IP Address %s: %s" % (ip, e))
+            logcm("Error: Invalid IP Address %s: %s" % (ip, e))
```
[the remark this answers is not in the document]
t-woerner: Done

ipaclient/install/client.py
```diff
@@ -2055,9 +2056,9 @@ def install_check(options):
     global client_domain
     global cli_basedn
 
-    print("This program will set up FreeIPA client.")
-    print("Version {}".format(version.VERSION))
-    print("")
+    logcm("This program will set up FreeIPA client.")
+    logcm("Version %s", version.VERSION)
```
abbra: Wrong conversion here, you need `%`, not `,`

ipaclient/install/client.py
```diff
@@ -2084,10 +2085,10 @@ def install_check(options):
         try:
             timeconf.check_timedate_services()
         except timeconf.NTPConflictingService as e:
-            print("WARNING: conflicting time&date synchronization service '{}'"
-                  " will be disabled".format(e.conflicting_service))
-            print("in favor of chronyd")
-            print("")
+            logcm("WARNING: conflicting time&date synchronization service '%s'"
+                  " will be disabled", e.conflicting_service)
```
abbra: Same here

ipaclient/install/client.py
tiran: no, that's the correct form.

ipaclient/install/client.py
tiran: No, Thomas is using the recommended form here. See https://docs.python.org/3/library/logging.html#logging.debug

ipaserver/install/custodiainstance.py
```diff
@@ -227,10 +229,9 @@ def _wait_keys(self):
                 if saved_e is None:
                     # FIXME: Change once there's better way to show this
                     # message in installer output,
-                    print(
-                        "  Waiting for keys to appear on host: {}, please "
-                        "wait until this has completed.".format(
-                            self.ldap_uri)
+                    logcm(
+                        "  Waiting for keys to appear on host: %s, please "
+                        "wait until this has completed.", self.ldap_uri
```
abbra: Same here

ipaserver/install/dsinstance.py
```diff
@@ -1052,7 +1054,7 @@ def change_admin_password(self, password):
                 ipautil.run(args, env=env)
                 logger.debug("ldappasswd done")
             except ipautil.CalledProcessError as e:
-                print("Unable to set admin password", e)
+                logcm("Unable to set admin password %s", e)
```
abbra: Same here

ipaserver/install/installutils.py
```diff
@@ -254,7 +256,7 @@ def record_in_hosts(ip, host_name=None, conf_file=paths.HOSTS):
                     return None
             return (hosts_ip, names)
         except IndexError:
-            print("Warning: Erroneous line '%s' in %s" % (line, conf_file))
+            logcm("Warning: Erroneous line '%s' in %s", line, conf_file)
```
abbra: same here

ipaserver/install/installutils.py
```diff
@@ -278,7 +280,7 @@ def read_ip_addresses():
         try:
             ip_parsed = ipautil.CheckedIPAddress(ip)
         except Exception as e:
-            print("Error: Invalid IP Address %s: %s" % (ip, e))
+            logcm("Error: Invalid IP Address %s: %s", ip, e)
```
abbra: Same here

ipaserver/install/installutils.py
```diff
@@ -288,12 +290,12 @@ def read_ip_addresses():
 def read_dns_forwarders():
     addrs = []
     if ipautil.user_input("Do you want to configure DNS forwarders?", True):
-        print("Following DNS servers are configured in /etc/resolv.conf: %s" %
+        logcm("Following DNS servers are configured in /etc/resolv.conf: %s",
```
abbra: same here

ipaserver/install/installutils.py
```diff
@@ -303,15 +305,15 @@ def read_dns_forwarders():
             try:
                 ip_parsed = ipautil.CheckedIPAddress(ip, parse_netmask=False)
             except Exception as e:
-                print("Error: Invalid IP Address %s: %s" % (ip, e))
-                print("DNS forwarder %s not added." % ip)
+                logcm("Error: Invalid IP Address %s: %s", ip, e)
+                logcm("DNS forwarder %s not added.", ip)
```
abbra: And these two lines

ipaserver/install/installutils.py
```diff
@@ -303,15 +305,15 @@ def read_dns_forwarders():
...
                 continue
 
-            print("DNS forwarder %s added. You may add another." % ip)
+            logcm("DNS forwarder %s added. You may add another.", ip)
```
abbra: Same here

ipaserver/install/installutils.py
```diff
@@ -393,7 +395,7 @@ def update_file(filename, orig, subst):
         os.chown(filename, st.st_uid, st.st_gid) # reset perms
         return 0
     else:
-        print("File %s doesn't exist." % filename)
+        logcm("File %s doesn't exist.", filename)
```
abbra: Same here

ipaserver/install/installutils.py
```diff
@@ -544,8 +546,8 @@ def update_hosts_file(ip_addresses, host_name, fstore):
     for ip_address in ip_addresses:
         if record_in_hosts(str(ip_address)):
             continue
-        print("Adding [{address!s} {name}] to your /etc/hosts file".format(
-            address=ip_address, name=host_name))
+        logcm("Adding [%s %s] to your /etc/hosts file", ip_address,
```
abbra: Same here

abbra:
@t-woerner I found a bunch of same type conversion errors that need to be fixed.

tiran:
> @t-woerner I found a bunch of same type conversion errors that need to be fixed.

The code is correct and uses the preferred format for logging. The logger does not perform string formatting by itself. The format message and arguments are passed to handlers and then formatted inside the handler's log formatter.

abbra:
@t-woerner @tiran my fault (too late after work, you know..) So please ignore it. ;) LGTM

ipapython/ipa_log_manager.py
```diff
@@ -141,12 +190,37 @@ def filter(self, record):
...
+            LOGGING_FORMAT_CMD_OUTPUT)
+
+    def format(self, record):
+        """
+        Format the specified record as text.
+
+        Uses LOGGING_FORMAT_CMD_OUTPUT for info_cmd, otherwise normal logging
+        format. For info_cmd record.__dict__["cmd_output"] is set to True.
+        """
+        if record.levelno >= logging.INFO and \
```
tiran: This logic should not be in the formatter but rather in the handler's formatter configuration. With this approach the debug logs will miss context and time information for ``logcm`` events. Please use a different formatter for log files and stdout logging here:

https://github.com/freeipa/freeipa/blob/b79ea6a6e92d224859c2d4533b99b3c15d3711ca/ipapython/ipa_log_manager.py#L166-L179

https://github.com/freeipa/freeipa/blob/b79ea6a6e92d224859c2d4533b99b3c15d3711ca/ipalib/plugable.py#L450-L522

ipapython/ipa_log_manager.py
```diff
@@ -164,6 +238,7 @@ def standard_logging_setup(filename=None, verbose=False, debug=False,
         finally:
             os.umask(umask)
         file_handler.setLevel(logging.DEBUG)
+        file_handler.addFilter(InfoCommandFilter())
```
tiran: Why do you add the filter to the file logger instead of the console logger? Isn't the filter suppose to filter which events end up in stdout?

tiran (changes requested):

tiran:
``ipapython.admintool`` also performs logger configuration. I think you need to address ``setup_logging``, too.

ipapython/ipa_log_manager.py
t-woerner: Adding the filter to the installation log fil e handler makes sure that the print output is not in the log file (like it has been up to now) and is by default in the normal installer (log) output. This makes sure that an existing application using FreeIPA parts does not need to enable this explicitly.

tiran: I think we have some kind of misunderstanding. Having the output in the log files is one of the benefits for me. It makes it easier to correlate debug logs with installer console output. 

I proposed the ``cmd_output`` flag to mark log entries that **also** land on stdout (console logger).

t-woerner: This is behavior change that needs to be verified also in all the tests.

t-woerner:
> `ipapython.admintool` also performs logger configuration. I think you need to address `setup_logging`, too.

ipapython.admintool._setup_logging, which is used in ipapython.admintool.setup_logging is using standard_logging_setup. Therefore it should not be needed to do further changes here.

ipapython/ipa_log_manager.py
tiran: This is the wrong approach, see [flow diagram](https://docs.python.org/3/howto/logging.html#logging-flow). The logic should be in the handler configuration. You have to use multiple handlers:

1. One stdout handler with a filter for ``cmd_output`` records, a filter for ``level < WARNING`` records, and a ``cmd_output_format`` formatter.
2. One stderr handler with a filter for ``cmd_output`` records, a filter for ``level >= WARNING`` records, and a ``cmd_output_format`` formatter.
3. One stdout handler with a filter for non-``cmd_output`` records and ``console_format`` formatter.

Handler 3 could be split into two separate handlers. I think it would be a good idea to not print a full traceback in ``cmd_output_formatter`` (see ``formatException``).

tiran (changes requested):
There are still a bunch of print() calls in the installer code.

``logcm("WARNING: ...")`` and ``print(file=sys.stderr)`` should be replaced with ``logcm(..., level=logging.WARNING)`` and a proper formatter/handler combination. The formatter could prepend ``WARN``.

duzda:
Hi @t-woerner, are you still planning to work on this or can this safely be closed?

duzda:
Hi, closing due to inactivity, if this ever becomes relevant again, please reopen

PR #3037, opened 2019-04-12
Remove deprecated object logger

tiran:
The object logger methods been deprecated for about two years since release
4.6.0. The log manager used to moneky-patch additional log methods like
info(), warning(), and error() into API plugin objects. The methods have
been replaced by calls to module logger objects in 4.6.0.

Remove monkey-patch logger methods, log manager, and its root logger from
ipapython.ipa_log_manager.

### DIFF
--- a/ipapython/ipa_log_manager.py
+++ b/ipapython/ipa_log_manager.py
@@ -20,11 +20,9 @@ import logging
 import os
 import re
 import time
-import warnings
-import sys
 
 # Module exports
-__all__ = ['log_mgr', 'root_logger', 'standard_logging_setup',
+__all__ = ['standard_logging_setup',
            'ISO8601_UTC_DATETIME_FMT',
            'LOGGING_FORMAT_STDERR', 'LOGGING_FORMAT_STDOUT', 'LOGGING_FORMAT_FILE']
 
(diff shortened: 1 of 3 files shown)


ipapython/ipa_log_manager.py
```diff
@@ -20,11 +20,10 @@
 import os
 import re
 import time
-import warnings
 import sys
 
 # Module exports
-__all__ = ['log_mgr', 'root_logger', 'standard_logging_setup',
+__all__ = ['log_mgr', 'standard_logging_setup',
```
t-woerner: log_mgr is also not needed any more with this change.

ipapython/ipa_log_manager.py
```diff
@@ -199,6 +122,3 @@ def convert_log_level(value):
 
 # Single shared instance of log manager
 log_mgr = sys.modules[__name__]
```
t-woerner: Same here

t-woerner:
Please add abit more information to the commit message. Like for example that root_logger and log_mgr have finally been removed.

ipapython/ipa_log_manager.py
```diff
@@ -20,11 +20,10 @@
 import os
 import re
 import time
-import warnings
 import sys
```
t-woerner: sys is not needed now.

tiran:
@t-woerner Please review again

rcritten:
LGTM, will give @t-woerner final say.

tiran:
master:

* 36c65c4aaa0dd73969c2e99a91da8ccf7d73e737 Remove deprecated object logger

PR #3039, opened 2019-04-15
Add temporary directory manager

tiran:
The temporary directory manager simplifies the handling of temporary
files that are shared with other processes or kept through out the life
time of the current process. It should only be used in case
tempfile.NamedTemporaryFile is not up for the task.

The manager creates a new temporary directory for each user. The
directory and all its files are accessible by the target user and the
root group ($uid:root / 0o770 / 0o660) to avoid DAC override capability.
The temporary directory is automatically removed on process exit.

Related: https://pagure.io/freeipa/issue/7911

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -19,6 +19,7 @@
 
 from __future__ import print_function
 
+import atexit
 import codecs
 import logging
 import string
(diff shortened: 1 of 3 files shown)


ipapython/ipautil.py
```diff
@@ -1571,3 +1572,107 @@ def major(self):
...
         return self[1]
+
+
+class _TemporaryDirectoryManager:
+    """File manager for shared temporary files
+
+    The temporary directory manager simplifies the handling of temporary
+    files that are shared with other processes or kept through out the life
+    time of the current process. It should only be used in case
+    tempfile.NamedTemporaryFile is not up for the task.
```
rcritten: I think it would be helpful for future developers to expand on a few specific cases where tempfile is not up to the task.

fcami:
hi @tiran could you please rebase this PR so that CI can pass?
Thanks

t-woerner:
What do you think about using a different name like _TemporaryDirectoryHandler? I expected something different within the context of IPA.

PR #3023, opened 2019-04-10
[Backport][ipa-4-7] Add PKI config override option

tiran:
Manual backport of PR #2976 

Add an option to override CA and KRA settings passed to pkispawn. The feature allows users to change key size, signature algorithm, and other parameters. It's a prerequisite for HSM support.

The patchset also simplifies and improves how IPA creates the pki.ini files that gets passed to pkispawn.

See pagure.io/freeipa/issue/5608

PR #3040, opened 2019-04-15
 Increase CA key size to 3072

tiran:
Fixes: https://pagure.io/freeipa/issue/6790

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -12,7 +12,10 @@ from __future__ import absolute_import
 import os
 from datetime import datetime, timedelta
 import time
+
+from cryptography.hazmat.primitives import hashes
 import pytest
+
 from ipalib.constants import DOMAIN_LEVEL_0
 from ipaplatform.constants import constants
 from ipaplatform.paths import paths
@@ -427,6 +430,43 @@ class TestInstallMaster(IntegrationTest):
         assert cmd.returncode != 0
         exp_str = ("ipa: ERROR: No YubiKey found")
         assert exp_str in cmd.stderr_text
+
+    def test_pki_certs(self):
+        certs, keys = tasks.certutil_certs_keys(
+            self.master,
+            paths.PKI_TOMCAT_ALIAS_DIR,
+            paths.PKI_TOMCAT_ALIAS_PWDFILE_TXT
+        )
+
+        expected_certs = {
+            # CA
+            'caSigningCert cert-pki-ca': 'CTu,Cu,Cu',
+            'ocspSigningCert cert-pki-ca': 'u,u,u',
+            'subsystemCert cert-pki-ca': 'u,u,u',
+            'auditSigningCert cert-pki-ca': 'u,u,Pu',  # why P?
+            # KRA
+            'transportCert cert-pki-kra': 'u,u,u',
+            'storageCert cert-pki-kra': 'u,u,u',
+            'auditSigningCert cert-pki-kra': 'u,u,Pu',
+            # server
+            'Server-Cert cert-pki-ca': 'u,u,u',
+        }
+        assert certs == expected_certs
+        assert len(certs) == len(keys)
+
+        for nickname in sorted(certs):
+            cert = tasks.certutil_fetch_cert(
+                self.master,
+                paths.PKI_TOMCAT_ALIAS_DIR,
+                paths.PKI_TOMCAT_ALIAS_PWDFILE_TXT,
+                nickname
+            )
+            key_size = cert.public_key().key_size
+            if nickname == 'caSigningCert cert-pki-ca':
+                assert key_size == 3072
+            else:
+                assert key_size == 2048
+            assert cert.signature_hash_algorithm.name == hashes.SHA256.name
 
 
 class TestInstallMasterKRA(IntegrationTest):
--- a/install/share/ipaca_customize.ini
+++ b/install/share/ipaca_customize.ini
@@ -29,7 +29,7 @@ ipa_signing_algorithm=SHA256withRSA
 # Used for IPA CA
 # signing algorithm can be overriden on command line
 ipa_ca_signing_algorithm=%(ipa_key_algorithm)s
-ipa_ca_key_size=%(ipa_key_size)s
+ipa_ca_key_size=3072
 ipa_ca_key_type=%(ipa_key_type)s
 
 # HSM support
--- a/ipalib/install/certmonger.py
+++ b/ipalib/install/certmonger.py
@@ -590,6 +590,18 @@ def resubmit_request(
             update['template-is-ca'] = True
             update['template-ca-path-length'] = -1  # no path length
 
+        # TODO: certmonger assumes some hard-coded defaults like RSA 2048.
+        # Try to fetch values from current cert rather.
+        for key, convert in [('key-size', int), ('key-type', str)]:
+            try:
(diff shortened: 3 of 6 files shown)


abbra:
LGTM. Could you please add a commit message that says from now on we expect default CA key size to be 3072 but normal certs to default to 2048 bit and that it is enforced in a specific test. Any change of the default key sizes need to adopt the test in question.

abbra:
Also, you might want to pull from my original pull request a fix for making sure we pass the default certificate key size for external CA renewal case like I did in a3b4275e8b0e8af8d436c93aa1446dfcfe9568ed

ipatests/test_integration/test_installation.py
```diff
@@ -428,6 +431,43 @@ def test_error_for_yubikey(self):
...
+                paths.PKI_TOMCAT_ALIAS_DIR,
+                paths.PKI_TOMCAT_ALIAS_PWDFILE_TXT,
+                nickname
+            )
+            key_size = cert.public_key().key_size
+            if nickname == 'caSigningCert cert-pki-ca':
+                assert key_size == 3072
+            else:
+                assert key_size == 2048
+            assert cert.signature_hash_algorithm == hashes.SHA256()
```
abbra: Actually, it is better to compare names here.
```
assert cert.signature_hash_algorithm.name == hashes.SHA256().name
```

tiran:
> LGTM. Could you please add a commit message that says from now on we expect default CA key size to be 3072 but normal certs to default to 2048 bit and that it is enforced in a specific test. Any change of the default key sizes need to adopt the test in question.

The PR is not marked as ready for review for a reason. I'm still testing. :)

> Also, you might want to pull from my original pull request a fix for making sure we pass the default certificate key size for external CA renewal case like I did in [a3b4275](https://github.com/freeipa/freeipa/commit/a3b4275e8b0e8af8d436c93aa1446dfcfe9568ed)

I don't want to commit that workaround for multiple reasons. For one a user can override the setting in the first step of external CA installation. We cannot access the value in the second step. The problem has be solved in a better way.

abbra:
> I don't want to commit that workaround for multiple reasons. For one a user can override the setting in the first step of external CA installation. We cannot access the value in the second step. The problem has be solved in a better way.
Sure. I wanted you to not forget about the actual problem there. As to how to solve it, we can use system store to record values like that and recover them later. We can peek into existing LDAP copy of a CA cert and pull the value from there too -- either at the moment we need it or to store it in a system store.

tiran:
It's renewal, so we should have access to the original cert. Just grab the setting from the existing cert and reuse it.

abbra:
please note that there are tests with hardcoded key size and algorithm type: https://github.com/freeipa/freeipa/commit/c568e297c095aeed2ea5ec249327399bb4424d72

frozencemetery:
Maybe I'm missing something, but wouldn't it make more sense to bump to 4096 rather than (probably) having to bump again in the future?

tiran:
> Maybe I'm missing something, but wouldn't it make more sense to bump to 4096 (or larger) rather than (probably) having to bump again in the future?

Performance, even on a modern, high-end system 4096 bit RSA verify is nearly three times slower than RSA 2048. 3072 is already have the performance. Signing is much worse. 4096 bit RSA is only marginal better than 3072. 3072 bit RSA is equivalent to 128 bit symmetric. 4096 is equivalent to around 145.

Here are some speed tests on an Intel i7-6820HQ. The test is single threaded and runs with 3.6 GHz Turbo Boost. 

```
$ openssl speed rsa2048 rsa3072 rsa4096
Doing 2048 bits private rsa's for 10s: 15169 2048 bits private RSA's in 9.98s
Doing 2048 bits public rsa's for 10s: 511245 2048 bits public RSA's in 9.98s
Doing 3072 bits private rsa's for 10s: 5025 3072 bits private RSA's in 9.98s
Doing 3072 bits public rsa's for 10s: 250935 3072 bits public RSA's in 9.98s
Doing 4096 bits private rsa's for 10s: 2260 4096 bits private RSA's in 9.98s
Doing 4096 bits public rsa's for 10s: 146799 4096 bits public RSA's in 9.98s
OpenSSL 1.1.1b FIPS  26 Feb 2019
built on: Fri Mar 15 16:10:28 2019 UTC
options:bn(64,64) md2(char) rc4(16x,int) des(int) aes(partial) idea(int) blowfish(ptr) 
compiler: gcc -fPIC -pthread -m64 -Wa,--noexecstack -Wall -O3 -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -Wa,--noexecstack -Wa,--generate-missing-build-notes=yes -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -DOPENSSL_USE_NODELETE -DL_ENDIAN -DOPENSSL_PIC -DOPENSSL_CPUID_OBJ -DOPENSSL_IA32_SSE2 -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_MONT5 -DOPENSSL_BN_ASM_GF2m -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DKECCAK1600_ASM -DRC4_ASM -DMD5_ASM -DAES_ASM -DVPAES_ASM -DBSAES_ASM -DGHASH_ASM -DECP_NISTZ256_ASM -DX25519_ASM -DPADLOCK_ASM -DPOLY1305_ASM -DZLIB -DNDEBUG -DPURIFY -DDEVRANDOM="\"/dev/urandom\"" -DSYSTEM_CIPHERS_FILE="/etc/crypto-policies/back-ends/openssl.config"
                  sign    verify    sign/s verify/s
rsa 2048 bits 0.000658s 0.000020s   1519.9  51227.0
rsa 3072 bits 0.001986s 0.000040s    503.5  25143.8
rsa 4096 bits 0.004416s 0.000068s    226.5  14709.3
```

tiran:
> please note that there are tests with hardcoded key size and algorithm type: [c568e29](https://github.com/freeipa/freeipa/commit/c568e297c095aeed2ea5ec249327399bb4424d72)

I like to keep this test as it stands. It shows that IPA CA can have a larger key size than its external signing CA. I'll add a comment to explain why it's still 2048 bit key.

abbra:
LGTM. We now pick up key-size and key algorithm from the existing certmonger's request so it should work with any size/algo combination that is permitted.

tiran:
@abbra I don't think the new default key size should land in 4.7. Nor can I land the change easily. The backport https://github.com/freeipa/freeipa/pull/3023 did not land in 4.7.

abbra:
ACKed. I'm fine with landing this in 4.8 only.

tiran:
master:

* 45b8cc1d83cde31af53d2cffd3913becc591ad34 Increase default key size for CA to 3072 bits
* 00a7868fa4011a0639bf96d23225c4203980bfb3 Reuse key type and size in certmonger resubmit
* 2bad9fd0df540dd93c7802cc06fbcf1f72f83965 Explain why tests still use 2048bit external CA

frozencemetery:
@tiran got it, thanks for explaining

PR #3045, opened 2019-04-16
CA instances: make sure that an instance actually usable after a restart

abbra:
Restarting CA instance might take time.

Dogtag doesn't report status of its own services to init system.
However, in most cases where we expect CA to start, we actually want to
use its HTTP or HTTPS end-points.

Wait until CA status can be retrieved over HTTP end-point. This allows
us to operate in both provisioned and pre-provision environments when
certificates aren't available yet so we cannot rely on TLS.

Fixes: https://github.com/freeipa/ansible-freeipa/issues/72

PR #3047, opened 2019-04-17
Add ExecStartPost hook to wait for Dogtag PKI

tiran:
Dogtag PKI typically takes around 10 seconds to start and respond to
requests. Dogtag uses a simple systemd service, which means systemd is
unable to detect when Dogtag is ready. Commands like ``systemctl start``
and ``systemctl restart`` don't block and wait until the CA is up. There
have been various workarounds in Dogtag and IPA.

Systemd has an ExecStartPost hook to run programs after the main service
is started. The post hook blocks systemctl start and restart until all
post hooks report ready, too. The new ipa-pki-wait-running script polls
on port 8080 and waits until the CA subsystem returns ``running``.

The patch also changes the service timeout to 120s as documented in man default.conf(5). Even 120s won't wait 120s, because systemd uses a smaller default timeout of 90s for service startup.

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -399,6 +399,7 @@ class CAInstance(DogtagInstance):
                 self.step("creating installation admin user", self.setup_admin)
             self.step("configuring certificate server instance",
                       self.__spawn_instance)
+            self.step("Add ipa-pki-wait-running", self.add_ipa_wait)
             self.step("reindex attributes", self.reindex_task)
             self.step("exporting Dogtag certificate store pin",
                       self.create_certstore_passwdfile)
@@ -603,6 +604,18 @@ class CAInstance(DogtagInstance):
                         paths.CACERT_P12)
 
         logger.debug("completed creating ca instance")
+
+    def add_ipa_wait(self):
+        """Add ipa-pki-wait-running to pki-tomcatd service
+        """
+        conf = paths.SYSTEMD_PKI_TOMCAT_IPA_CONF
+        directory = os.path.dirname(conf)
+        if not os.path.isdir(directory):
+            os.mkdir(directory)
+        with open(conf, 'w') as f:
+            f.write('[Service]\n')
+            f.write('ExecStartPost={}\n'.format(paths.IPA_PKI_WAIT_RUNNING))
+        tasks.systemd_daemon_reload()
 
     def safe_backup_config(self):
         """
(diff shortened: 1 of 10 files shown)


tiran:
Alternative approach to #3045

abbra:
The commit title includes `ExecStartPre` while you are doing `ExecStartPost`

ipaserver/install/cainstance.py
```diff
@@ -604,6 +605,18 @@ def __spawn_instance(self):
...
         logger.debug("completed creating ca instance")
 
+    def add_ipa_wait(self):
+        """Add ipa-pki-wait-running to pki-tomcatd service
+        """
+        conf = paths.SYSTEMD_PKI_TOMCAT_IPA_CONF
+        directory = os.path.dirname(conf)
+        if not os.path.isdir(directory):
+            os.mkdir(directory)
+        with open(conf) as f:
```
abbra: You need to open file to write, as reading mode is default. The tests are failing because of that:
```
2019-04-17T06:16:52Z DEBUG   [2/30]: Add ipa-pki-wait-running
2019-04-17T06:16:52Z DEBUG Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/ipaserver/install/service.py", line 603, in start_creation
    run_step(full_msg, method)
  File "/usr/lib/python3.7/site-packages/ipaserver/install/service.py", line 589, in run_step
    method()
  File "/usr/lib/python3.7/site-packages/ipaserver/install/cainstance.py", line 615, in add_ipa_wait
    with open(conf) as f:
FileNotFoundError: [Errno 2] No such file or directory: '/etc/systemd/system/pki-tomcatd@pki-tomcat.service.d/ipa.conf'
```

tiran: ``EMORECOFFEE`` ...

rcritten:
I don't understand why a startup checker needs to have a context of installer. Is it to read a custom installer.conf (e.g. for setting a timeout)? What happens post-install? It will fail to use the custom settings and potentially fail.

tiran:
Yes, that's the idea here. The installer rewrites ``default.conf``. I figured out that ``installer.conf`` is a good place to tune installation defaults. ``api.bootstrap(context='installer')`` ignores a missing ``installer.conf`` file.

I don't care much about the context. Which one of the following options would you prefer?

1. keep ``installer`` context and document why the script uses the context
2. use a new ``service`` or ``pki`` context
3. don't set an explicit context at all

tiran:
I have removed the ``context='installer'`` argument and documented the general discussion with Rob in ticket https://pagure.io/freeipa/issue/7919.

tiran:
``vagrant up`` is still failing, rerunning.

abbra:
LGTM

tiran:
master:

* 3f9e23f12573b1b5dbd5284620e86393018aa548 Add ExecStartPost hook to wait for Dogtag PKI
* 5452eb6e77543ccfbb681d907bf94663c47c1d0c Reduce startup_timeout to 120sec as documented

PR #3048, opened 2019-04-17
Upstream various fixes from Ubuntu package

tiran:
* use ``lesscpy.__main__`` instead of hard-coded name for py3-lesscpy
* skip authselect migration on Debian
* add font paths for open-sans and font-awesome 
* Fixes for OpenDNSSEC 2.0 on Debian
* Use different paths for KDC cert and key
* Wait for pki-tomcatd
* Replace hard-coded paths in oddconf configuration

The gzip patch is no longer required thanks to 5ecede781b4eba11a537afd104fc9df9518958c4

@tjaalton please take a look.


ipaserver/install/opendnssecinstance.py
```diff
@@ -274,6 +279,11 @@ def __setup_dnssec(self):
         if not self.fstore.has_file(paths.OPENDNSSEC_KASP_DB):
             self.fstore.backup_file(paths.OPENDNSSEC_KASP_DB)
 
+        if paths.ODS_ENFORCER is not None:
+            ods_cmd = paths.ODS_ENFORCER
+        else:
+            ods_cmd = paths.ODS_KSMUTIL
+
```
rcritten: Not to make things more complex but wouldn't it be better to put this difference into ipaplatform and use a generic name like ODS_MANAGER or something?

rcritten:
Just one comment, otherwise LGTM.

tjaalton:
Heh, thanks for sending this MR.. I haven't tested any of these but at least they look good. The OpenDNSSEC 2.0 setup probably was still broken, but at least the package itself should be in fine shape now thanks to a new maintainer. Again, it's something to test with the current set of packages to be sure.. but I guess it's not worse than what's in there now?

ipaserver/install/opendnssecinstance.py
tiran: Excellent idea! I move the calls into ``ipaplatform.tasks``. At first I had separate code for Debian and Fedora. But then I came to the conclusion that Fedora will eventually update to OpenDNSSEC 2.0, too. Now the abstraction is in the base tasks module.

tiran:
@tjaalton I pushed two more changes. The first addresses a bad path on Debian that breaks key replication of sub CAs. The second one generates oddjob configs with correct paths.

The first change does not fix existing installations with a bad path. You need to fix this in a pre-update hook in the Debian/Ubuntu packages.

install/ui/util/make-css.sh
```diff
@@ -22,5 +22,5 @@ set -o errexit
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-py3-lesscpy -x $DIR/../less/ipa.less > $DIR/../css/ipa.css
+python3 -m lesscpy -x $DIR/../less/ipa.less > $DIR/../css/ipa.css
```
abbra: A small comment here: on RHEL 8.0 beta we have platform python and it is set as PYTHON:
```
export PYTHON=%{__python3}
```
Since build environment in RHEL 8.0 beta has no python3 but rather uses `platform-python`, this change will break build there. Perhaps, it is better to check if PYTHON is set and use that one instead?

```
${PYTHON:-python3} -m lesscpy -x ...
```

tiran: Thanks, I forgot about platform-python. In the end it was simpler to replace the shell script with a one-liner in make: ``$(PYTHON) -m lesscpy -x ../less/ipa.less > $@``.

abbra:
LGTM

tiran:
master:

* d796c37f17c67cf91693b8f8f96f0f2681a4989a Debian: use -m lesscpy instead of hard-coded name
* ff41a09e363c90e510697c6c7ae8d853d6a98b21 Debian doesn't have authselect
* da2cf1c513ea42dbb38952273421fefdd92d4d0b Debian: Add paths for open-sans and font-awesome
* 750e658dba1244ad6a386ab70fe309fd910ab419 Debian: Add fixes for OpenDNSSEC 2.0
* 8592603ea36375cc0670df1e0b1efb486bafcc82 Debian: Use different paths for KDC cert and key
* edaea8865f42f8b6e786f6648ac4cb5c4ddd4f35 Add ODS manager abstraction to ipaplatform
* d703f3dbcb992b0f29c5c71c35ec112eadccf2ed Debian: Fix replicatio of light weight sub CAs
* 81d0108afc8f08b0e424d8d1fdf83f1e0ccc7765 Debian: auto-generate config files for oddjobd
* 0fa692a5265f4764e3e43fda65eced50e7f42448 Debian: Use RedHatCAService for pki-tomcatd

PR #3052, opened 2019-04-23
Pass token_name to certmonger

tiran:
For HSM support, IPA has to pass the token name for CA and subsystem
certificates to certmonger. For now, only the default 'internal' token is
supported.

Related: https://pagure.io/freeipa/issue/5608

### DIFF
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -98,6 +98,10 @@ class DogtagInstance(service.Service):
     tracking_reqs = None
     server_cert_name = None
 
+    # token for CA and subsystem certificates. For now, only internal token
+    # is supported.
+    token_name = "internal"
+
     ipaca_groups = DN(('ou', 'groups'), ('o', 'ipaca'))
     ipaca_people = DN(('ou', 'people'), ('o', 'ipaca'))
     groups_aci = (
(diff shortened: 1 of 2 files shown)


ipaserver/install/dogtaginstance.py
```diff
@@ -294,24 +304,26 @@ def configure_certmonger_renewal(self):
                     # Give dogtag extra time to generate cert
                     timeout=CA_DBUS_TIMEOUT)
 
-    def __get_pin(self):
+    def __get_pin(self, token_name=None):
         try:
-            return certmonger.get_pin('internal')
+            return certmonger.get_pin(token_name)
```
rcritten: Rather than defaulting token_name to None why not default to "internal" then you can remove the test for None in get_pin() and other places? Theoretically internal:Server-Cert is the equivalent to Server-Cert so there shouldn't be any harm in always using internal unless you are trying to introduce as few changes as possible (it would certainly exercise some code in new ways).

abbra:
LGTM

ipaserver/install/dogtaginstance.py
tiran: That was my initial implementation, too. But I ran into some issues with Dogtag. I may have identified the problem and may have found a workaround. Let's see if tests are now passing.

tiran:
master:

* 8686cd3b4b69f725aee05c9cdd3034d7436055d3 Pass token_name to certmonger

PR #3053, opened 2019-04-23
Deprecate ipa-client-install --request-cert

tiran:
Mark the --request-cert option for ipa-client-install as deprecated.
Users are encouraged to request a PEM certificate with certmonger
instead. The option and /etc/ipa/nssdb will be removed in a future
version.

Related: https://pagure.io/freeipa/issue/7492

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -3691,6 +3691,7 @@ class ClientInstallInterface(hostname_.HostNameInstallInterface,
 
     request_cert = knob(
         None,
+        deprecated=True,
         description="request certificate for the machine",
     )
     request_cert = prepare_only(request_cert)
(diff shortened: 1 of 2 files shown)


ipaclient/install/client.py
```diff
@@ -3713,6 +3714,9 @@ def __init__(self, **kwargs):
             raise RuntimeError(
                 "--ntp-pool cannot be used together with --no-ntp")
 
+        if self.request_cert:
+            logger.warning("Option --request-cert has been deprecated")
```
rcritten: Can you add "and will be removed in a future release" here as well?

tiran: sure! It's done.

tiran:
master:

* 411e6c37fbcb6ddc4c60b240ecaaa418401ac359 Deprecate ipa-client-install --request-cert

PR #3062, opened 2019-04-24
Disable global p11-kit configuration for NSS

tiran:
The p11-kit configuration injects p11-kit-proxy into all NSS databases.
Amongst other p11-kit loads SoftHSM2 PKCS#11 provider. This interferes
with 389-DS, certmonger, Dogtag and other services. For example certmonger
tries to open OpenDNSSEC's SoftHSM2 token, although it doesn't use it at
all. It also breaks Dogtag HSM support testing with SoftHSM2.

IPA server does neither need nor use p11-kit.

### DIFF
--- a/ipaplatform/base/paths.py
+++ b/ipaplatform/base/paths.py
@@ -366,6 +366,8 @@ class BasePathNamespace:
     AUTHCONFIG = None
     AUTHSELECT = None
     SYSCONF_NETWORK = None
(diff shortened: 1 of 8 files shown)


tiran:
PR #3063 is a different approach to solve the same issue.

tiran:
Closing this in favor of #3063

PR #3063, opened 2019-04-24
Globally disable softhsm2 in p11-kit-proxy

tiran:
The p11-kit configuration injects p11-kit-proxy into all NSS databases.
Amongst other p11-kit loads SoftHSM2 PKCS#11 provider. This interferes
with 389-DS, certmonger, Dogtag and other services. For example certmonger
tries to open OpenDNSSEC's SoftHSM2 token, although it doesn't use it at
all. It also breaks Dogtag HSM support testing with SoftHSM2.

IPA server does neither need nor use SoftHSM2 proxied by p11-kit.


tiran:
The change needs SELinux type ``pkcs11_modules_conf_t`` from 
https://github.com/fedora-selinux/selinux-policy/commit/50cc590ecdd0797c653a7685b559b5c48e9c3d30 to work correctly. Otherwise there are AVCs for ns-slapd.

abbra:
LTGM. It is unfortunate but we have to disable it.

tiran:
master:

* 74e09087ed0ebf75936d5045eaaad69d71c678d6 Globally disable softhsm2 in p11-kit-proxy

hcoin:
Please take a look at how this patch breaks dnssec, detailed here:  https://pagure.io/freeipa/issue/8757

PR #3064, opened 2019-04-24
fix test test_integration/test_commands.py::test_ssh_key_connection

wladich:
In testcase test_ssh_key_connection the value provided by tmpdir fixture
is used as regular string, i.e. passed as an argument to os.path.join.
tmpdir is an instance of py.path.local, and in old versions of package
python-py it does not have methods of string object,
which causes this test to fail in distributions where such version of
python-py is installed.
Fixed by explicitly changing type of tmpdir to str.


tiran:
ipa-4-6:

* 8b59475dffbe06a7dc1905ef0858a51ec0356566 ipatests: coerce tmpdir to string

PR #3066, opened 2019-04-24
Check have packages for extra features been installed before restoring backup

xxblx:
Check have packages for extra features been installed before restoring backup 

`iparestore --full` should check that packages for extra features such as dns and adtrust are installed in the system before restoring a backup in case the backup includes content for these features. If the packages are not installed full backup should be refused and an error message with suggestions should be showed.

If corresponding packages for these features are not installed before the backup restoring, it may cause a situation when the packages are going to be installed after the restoring. In that case configuration files restored by `ipa-restore` will be replaced by default configuration files if the files are tracked by `rpm`. E.g. if `freeipa-server-trust-ad` is not installed before `ipa-restore --full` running, when the package will be installed it also will bring `samba` package according to the dependencies. At `samba` installation step exist correct `/etc/samba/smb.conf` is going to be replaced by the default one from the `samba` package.

Fixes: https://pagure.io/freeipa/issue/7630

### DIFF
--- a/ipaserver/install/ipa_restore.py
+++ b/ipaserver/install/ipa_restore.py
@@ -324,6 +324,28 @@ class Restore(admintool.AdminTool):
                     not user_input("Continue to restore?", False)):
                 raise admintool.ScriptError("Aborted")
 
+        # Check have optional dependencies been installed for extra
+        # features from backup
+        if restore_type == 'FULL':
+            if 'ADTRUST' in self.backup_services:
+                if not adtrustinstance or not adtrustinstance.check_inst():
+                    raise admintool.ScriptError(
+                        "Backup includes AD trust feature, it requires '{}' "
+                        "package. Please install the package and "
+                        "run ipa-restore again.".format(
+                            constants.IPA_ADTRUST_PACKAGE_NAME
+                        )
+                    )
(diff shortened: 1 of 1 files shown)


xxblx:
> Could you please change the PR to use `format()` instead of `%`? And please end each sentence with full stop `.`.

Sure. Done. 

But what is the difference for the output in that case? Is it a kind of internal standard to use `format()` instead of `%`? I ask because a lot of constructions in the same method use `%` as well.

tiran:
The modulo format operator ``%`` is mostly used in old code or in logging calls. We can't switch the logging code yet, but we can use the modern ``format()`` method in new code.

tiran:
``ipaserver/install/ipa_restore.py:330: [E1101(no-member), Restore.run] Instance of 'Restore' has no 'backup_service' member; maybe 'backup_services'?)``

tiran:
pep8 checker doesn't like one line:

```
PEP-8 errors:
./ipaserver/install/ipa_restore.py:343:80: E501 line too long (80 > 79 characters)
```

xxblx:
master:

* 23d5c05232bcedda31ed520a487db0f6317ed92e Check have packages for extra features been installed before restoring backup

PR #3068, opened 2019-04-24
Update samba configuration defaults

abbra:
Long overdue, we need to fix Samba configuration to follow testparm recommendations.
A second fix is to make sure smbd by default is limited in its capability to handle concurrent client requests to prevent SMBLoris attack.

### DIFF
--- a/ipaserver/install/adtrustinstance.py
+++ b/ipaserver/install/adtrustinstance.py
@@ -130,6 +130,15 @@ def map_Guests_to_nobody():
     logger.debug("Map BUILTIN\\Guests to a group 'nobody'")
     ipautil.run(args, env=env, raiseonerr=False, capture_error=True)
 
+
+def get_idmap_range(realm):
+    idrange = api.Command.idrange_show('{}_id_range'.format(realm))['result']
+    range_start = int(idrange['ipabaseid'][0])
+    range_size = int(idrange['ipaidrangesize'][0])
+    range_fmt = '{} - {}'.format(range_start, range_start + range_size)
+    return range_fmt
+
+
 class ADTRUSTInstance(service.Service):
 
     ATTR_SID = "ipaNTSecurityIdentifier"
@@ -838,12 +847,18 @@ class ADTRUSTInstance(service.Service):
         )
         api.Backend.ldap2.add_entry(entry)
 
+    def __retrieve_local_range(self):
+        """Retrieves local IPA ID range to make sure
+        """
+        self.sub_dict['IPA_LOCAL_RANGE'] = get_idmap_range(self.realm)
+
     def create_instance(self):
         self.step("validate server hostname",
                   self.__validate_server_hostname)
         self.step("stopping smbd", self.__stop)
         self.step("creating samba domain object", \
                   self.__create_samba_domain_object)
+        self.step("retrieve local idmap range", self.__retrieve_local_range)
         self.step("creating samba config registry", self.__write_smb_registry)
         self.step("writing samba config file", self.__write_smb_conf)
         self.step("adding cifs Kerberos principal",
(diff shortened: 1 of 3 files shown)


ipaserver/install/adtrustinstance.py
```diff
@@ -130,6 +130,15 @@ def map_Guests_to_nobody():
     logger.debug("Map BUILTIN\\Guests to a group 'nobody'")
     ipautil.run(args, env=env, raiseonerr=False, capture_error=True)
 
+
+def get_idmap_range(realm):
+    range = api.Command.idrange_show('%s_id_range' % realm)['result']
```
tiran: ``range`` is the name of a builtin function. pylint should complain about this. Could you pick another name, please?

Also let's use ``format()`` everywhere. Bonus points for f-strings if the change doesn't need backporting.

ipaserver/install/adtrustinstance.py
```diff
@@ -130,6 +130,15 @@ def map_Guests_to_nobody():
...
+    range_start = int(range['ipabaseid'][0])
```
tiran: Is the ``int()`` call actually necessary? The API should return a proper integer here.

ipaserver/install/adtrustinstance.py
```diff
@@ -130,6 +130,15 @@ def map_Guests_to_nobody():
...
+    range_size = int(range['ipaidrangesize'][0])
+    range_fmt = '%s - %s' % range_start, range_start + range_size
```
tiran: ``TypeError: not enough arguments for format string``

ipaserver/install/adtrustinstance.py
```diff
@@ -838,12 +847,19 @@ def find_local_id_range(self):
         )
         api.Backend.ldap2.add_entry(entry)
 
+    def __retrieve_local_range(self):
+        """Retrieves local IPA ID range to make sure
+        """
+        self.sub_dict['IPA_LOCAL_RANGE'] = get_idmap_range(self.realm)
+
+
```
tiran: Too many empty lines

ipaserver/install/adtrustinstance.py
abbra: It will need a backport, possibly to 4.6. I changed the rest.

ipaserver/install/adtrustinstance.py
abbra: Yes, it needs it here because it is a simple dictionary:
```
>>> range = api.Command.idrange_show('%s_id_range' % api.env.realm)['result']
>>> range
{'ipaidrangesize': ['200000'], 'cn': ['EXAMPLE.COM_id_range'], 'ipabaserid': ['1000'], 'ipasecondarybaserid': ['100000000'], 'ipabaseid': ['1536000000'], 'iparangetype': ['local domain range'], 'iparangetyperaw': ['ipa-local'], 'dn': ipapython.dn.DN('cn=EXAMPLE.COM_id_range,cn=ranges,cn=etc,dc=example,dc=com')}
>>> range.get('ipabaseid')
['1536000000']
>>> range.__class__
<class 'dict'>
```

ipaserver/install/adtrustinstance.py
abbra: Fixed by using format()

ipaserver/install/adtrustinstance.py
abbra: `make fastcheck` did not highlight any of these errors on F30

ipaserver/install/adtrustinstance.py
tiran: Ah, the values are strings, not ints. I was expecting ints here.

ipaserver/install/adtrustinstance.py
tiran: I think the pep8 checker would have complained. Thanks!

rcritten:
master:

* 4ba888694bc31972740d52322a6b11006adaddc1 Set idmap config for Samba to follow IPA ranges and use SSSD
* b2c5691e73b7f6f38abed727a23b904290fc64cc Enforce SMBLoris attack protection in default Samba configuration

PR #3076, opened 2019-04-25
ipa console: catch proper exception when history file can not be open

wladich:
In python 2, when file can not be open, IOError is raised, but only OSEror
was catched. This caused "ipa console" command to crash when history file
did not exist.
In python3 OSError is raised in this situation, but as OSError is IOError
in python3, we can safely catch only IOError.

### DIFF
--- a/ipalib/cli.py
+++ b/ipalib/cli.py
@@ -971,7 +971,7 @@ class console(frontend.Command):
         history = os.path.join(api.env.dot_ipa, "console.history")
         try:
             readline.read_history_file(history)
-        except OSError:
+        except IOError:
             pass
 
         def save_history():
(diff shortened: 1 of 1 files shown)


rcritten:
Looks good, just a bit of nit-picking.

Please add a link to the ticket to the commit message, https://pagure.io/freeipa/issue/7922

A few minor corrections to the commit message:

In Python 3 IOError has been merged into OSError but in Python 2 they are still distinct exceptions. This caused "ipa console" in Python 2 to fail  when trying to initialize readline and the history file did not exist because the exception was not caught and ignored.

tiran:
ipa-4-7:

* 818f94d8d9ab08e72bac9ba61b6fd2a733df5ebf ipa console: catch proper exception when history file can not be open

PR #3084, opened 2019-04-26
ipactl restart: fix wrong logic when checking service list

flo-renaud:
ipactl is building a list of currently running services from
the content of /var/run/ipa/services.list, and a list of expected services
from the services configured in LDAP.

Because CA and KRA both correspond to the same pki-tomcatd service, the
lists may contain duplicates. The code handling these duplicates is called
at the wrong place, and may result in a wrong list of services to
stop / restart / start.
The fix removes the duplicates before returning the lists, hence making sure
that there is no error when building the list of services to stop / restart
/ start.

Fixes: https://pagure.io/freeipa/issue/7927


tiran:
master:

* 161008d5caea953973b3f70e1f9fc502bc211c68 ipactl restart: fix wrong logic when checking service list

PR #3085, opened 2019-04-26
tasks.py :: prepare_reverse_zone : Using '--skip-overlap-check' option during dnszone creation

ndehadrai:
Currently when the test_replica_promotion tests are run in CI, it returns warning message 'ipa: WARNING: ipa: ERROR: DNS zone with name <xyz> already exists'. Using the above option would force dnszone-add creation.

### DIFF
--- a/pylint_plugins.py
+++ b/pylint_plugins.py
@@ -499,6 +499,7 @@ AstroidBuilder(MANAGER).string_build(textwrap.dedent(
     api.env.force_schema_check = False
     api.env.home = ''  # object
     api.env.host = ''
+    api.env.host_princ = ''
     api.env.http_timeout = 0
     api.env.in_server = False  # object
     api.env.in_tree = False  # object
@@ -524,6 +525,7 @@ AstroidBuilder(MANAGER).string_build(textwrap.dedent(
     api.env.script = ''  # object
     api.env.site_packages = ''  # object
     api.env.skip_version_check = False
+    api.env.smb_princ = ''
     api.env.startup_timeout = 0
     api.env.startup_traceback = False
     api.env.tls_ca_cert = ''  # object


ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -141,7 +141,8 @@ def prepare_reverse_zone(host, ip):
     zone = get_reverse_zone_default(ip)
     result = host.run_command(["ipa",
                       "dnszone-add",
-                      zone], raiseonerr=False)
+                      zone,
+                      "--skip-overlap-check"], raiseonerr=False) # To force dnszone-add
```
tiran: The place of the comment is confusing. It's unclear if the comment refers to ``raiseonerror=False`` or ``--skip-overlap-check``.

Please reformat the code as
```
    result = host.run_command(
        [
            "ipa",
            "dnszone-add",
            "--skip-overlap-check",  # to force dnszone-add
            zone
        ],
        raiseonerr=False
    )
```
or perhaps replace the call with ``add_dns_zone(host, zone, skip_overlap_check=True)``.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -139,9 +139,7 @@ def wrapped(*args, **kwargs):
 
 def prepare_reverse_zone(host, ip):
     zone = get_reverse_zone_default(ip)
-    result = host.run_command(["ipa",
-                      "dnszone-add",
-                      zone], raiseonerr=False)
+    result = add_dns_zone(host, zone, skip_overlap_check=True)
```
tiran: ``add_dns_zone()`` returns None and fails when the zone doesn't exist and can be added.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -139,9 +139,14 @@ def wrapped(*args, **kwargs):
...
-                      "dnszone-add",
-                      zone], raiseonerr=False)
+    result = host.run_command(
+        [
+            "ipa",
+            "dnszone-add",
+            "--skip-overlap-check",  # to force dnszone-add
+            zone
+        ],
+        raiseonerr=False)
```
flo-renaud: Hi @ndehadrai 
I don't think this diff would fix the issue. If you look for instance at the logs for PR#3103:
- during the setup for `test_replica_promotion.py::TestUnprivilegedUserPermissions`, `ipa-server-install` is called with `--auto-reverse` and it creates the reverse zone 121.168.192.in-addr.arpa. (see [log](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/9768c652-6c67-11e9-b4ec-fa163ef01e1e/runner.log.gz)) because the server has IP@ 192.168.121.74
- during the test `test_replica_promotion.py::TestUnprivilegedUserPermissions::()::test_replica_promotion_by_unprivileged_user`, the test calls `ipa dnszone-add 121.168.192.in-addr.arpa.` (see [log](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/9768c652-6c67-11e9-b4ec-fa163ef01e1e/report.html)) because the client has IP@ 192.168.121.116
So the test is trying to create a reverse zone that already exists. The option `--skip-overlap-check` would force the creation even if the zone was handled *by another DNS server*, not if it is already handled by IPA.
A correct fix would rather be to check first if the zone is already defined in IPA, and try to add only if it's not the case, or ignore the error `DNS zone with name "xxx" already exists`.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -139,12 +139,25 @@ def wrapped(*args, **kwargs):
...
+            [
+                "ipa",
+                "dnszone-add",
+                "--skip-overlap-check",  # to force dnszone-add
+                zone
+            ],
+            raiseonerr=False)
+        if result.returncode > 0:
+            logger.warning("%s", result.stderr_text)
+        return zone, result.returncode
```
flo-renaud: If the ipa dnszone-find command succeeds (validate.returncode == 0), the function prepare_reverse_zone exits without returning any value, while it should return a Tuple (string, int). This will cause unpacking exception if the caller tries to read the returned value.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -137,15 +137,22 @@ def wrapped(*args, **kwargs):
...
+                               "dnszone-add",
+                               zone], raiseonerr=False)
+    if option.returncode > 0:
+        # Force DNSZone addition
+        cmd = ["ipa", "dnszone-add", "--skip-overlap-check", zone]
+        result = host.run_command(cmd, raiseonerr=False)
+        if result.returncode > 0:
+            logger.warning("%s", result.stderr_text)
+        return zone, result.returncode
     return zone, result.returncode
```
flo-renaud: In this codepath (option.returncode ==0), the variable `result` has not been assigned. I guess that you should return option.returncode instead.

PR #3102, opened 2019-05-01
Workaround for AJP to bind on IPv6 localhost

tiran:
Tomcat's AJP connector binds to IPv4 localhost only. This causes issues
with IPv6-only environment. The installer now detects if localhost6 is
available and working, then configures Dogtag's Tomcat to use
localhost6.


rcritten:
What about ipa-pki-proxy.conf? It configures things as:

ProxyPassReverse ajp://localhost:8009

PR #3105, opened 2019-05-02
Revert "Require a minimum SASL security factor of 56"

abbra:
This reverts commit 350954589774499d99bf87cb5631c664bb0707c4.

We cannot force increase in minimum SASL security factor until our
consumers are ready to deal with it. Unfortunately, realmd uses
anonymous connection for discovery and validation of IPA LDAP server.

The way it is done is fragile (it doesn't take into account an
advertised IPA version, only checks that 'IPA' string exists in the info
field) but since bumping of minimum SSF prevents reading IPA info field
using anonymous connection, client enrollment fails.

We should get back to bumping minimum SSF after realmd and other
potential consumers are fixed.


fcami:
LGTM, ACK.

fcami:
master:

* 294aa3a33375dc246b2a733fce3cbd09a39071a0 Revert "Require a minimum SASL security factor of 56"

tiran:
The original patch accomplished two things

1. It configured 389-DS to only accept secure connections with strong SASL SSF.
2. It configured our LDAP client to only allow connections to an LDAP server that allows strong SASL SSF.

I would prefer to keep the patch in and only comment out the call ``self.step("require minimal SSF", self.__min_ssf)`` for now.

abbra:
@tiran We can get it back later, once realmd is fixed. We also started to see it with ipa-client-install -- like in http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ecc0f7b0-6cca-11e9-b113-fa163eea5c7f/report.html

abbra:
@tiran could you please split the original commit into two parts?

PR #3106, opened 2019-05-02
Moving prompt for NTP options to install_check

Tiboris:
In a interactive installation of freeipa server a promt asks for NTP related
options after install_check has been called. As it may cause confusion to users
moving to install_check methods where the prompt for other options is being done.

Resolves: https://pagure.io/freeipa/issue/7930

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -2084,10 +2084,13 @@ def install_check(options):
         try:
             timeconf.check_timedate_services()
         except timeconf.NTPConflictingService as e:
-            print("WARNING: conflicting time&date synchronization service '{}'"
-                  " will be disabled".format(e.conflicting_service))
-            print("in favor of chronyd")
-            print("")
+            print(
+                "WARNING: conflicting time&date synchronization service "
+                "'{}' will be disabled in favor of chronyd\n".format(
+                    e.conflicting_service
+                )
+            )
+
         except timeconf.NTPConfigurationError:
             pass
 
@@ -2366,6 +2369,11 @@ def install_check(options):
                 "Proceed with fixed values and no DNS discovery?", False):
             raise ScriptError(rval=CLIENT_INSTALL_ERROR)
 
+    if options.conf_ntp:
+        if not options.on_master and not options.unattended and not (
+                options.ntp_servers or options.ntp_pool):
+            options.ntp_servers, options.ntp_pool = timeconf.get_time_source()
+
     cli_realm = ds.realm
     cli_realm_source = ds.realm_source
     logger.debug("will use discovered realm: %s", cli_realm)
@@ -2392,6 +2400,14 @@ def install_check(options):
     logger.debug("IPA Server source: %s", cli_server_source)
     logger.info("BaseDN: %s", cli_basedn)
     logger.debug("BaseDN source: %s", cli_basedn_source)
+
+    if not options.on_master:
+        if options.ntp_servers:
+            for server in options.ntp_servers:
+                logger.info("NTP server: %s", server)
+
+        if options.ntp_pool:
+            logger.info("NTP pool: %s", options.ntp_pool)
 
     # ipa-join would fail with IP address instead of a FQDN
     for srv in cli_server:
@@ -2458,7 +2474,7 @@ def update_ipa_nssdb():
                                    (nickname, sys_db.secdir, e))
 
 
-def sync_time(options, fstore, statestore):
+def sync_time(ntp_servers, ntp_pool, fstore, statestore):
     """
     Will disable any other time synchronization service and configure chrony
     with given ntp(chrony) server and/or pool using Augeas.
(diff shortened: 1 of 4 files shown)


xxblx:
@Tiboris tests fail with 
```
2019-05-02 09:45:27,627    DEBUG  transport.py               558 DEBUG    Error checking LDAP: Server is unwilling to perform: Minimum SSF not met.
```
Please do a rebase on https://github.com/freeipa/freeipa/commit/294aa3a33375dc246b2a733fce3cbd09a39071a0. That should fix PR CI failures.

ipaserver/install/server/install.py
```diff
@@ -709,6 +714,13 @@ def install_check(installer):
                   "Directory unless\nthe realm name of the IPA server matches "
                   "its domain name.\n\n")
 
+    if options.ntp_servers or options.ntp_pool:
+        for server in options.ntp_servers:
+            print("NTP server:       %s" % server)
```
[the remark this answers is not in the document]
Tiboris: Sure, will do.

ipaserver/install/server/install.py
```diff
@@ -427,13 +428,13 @@ def install_check(installer):
 
     if not options.no_ntp:
         try:
-            ipaclient.install.timeconf.check_timedate_services()
-        except ipaclient.install.timeconf.NTPConflictingService as e:
+            timeconf.check_timedate_services()
+        except timeconf.NTPConflictingService as e:
             print("WARNING: conflicting time&date synchronization service '{}'"
```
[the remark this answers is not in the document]
Tiboris: Okay!

ipaclient/install/client.py
```diff
@@ -2472,12 +2486,9 @@ def sync_time(options, fstore, statestore):
 
     if not options.ntp_servers and not options.ntp_pool:
         ds = discovery.IPADiscovery()
-        ntp_servers = ds.ipadns_search_srv(cli_domain, '_ntp._udp',
-                                           None, break_on_first=False)
-        if not ntp_servers and not options.unattended:
-            options.ntp_servers, options.ntp_pool = timeconf.get_time_source()
-        else:
-            options.ntp_servers = ntp_servers
+        options.ntp_servers = ds.ipadns_search_srv(
```
rcritten: Is it bad form to update the options passed in vs those things detected? Could this make debugging harder later?

Tiboris: Hello @rcritten, 

yes I agree, actual behaviour could make debugging harder however i think approach described below is good and I will add another logging message and comments to the code itself.
```python
if installation is interactive and no ntp options had been passed:
    ask user for ntp options interactively
if no ntp options are passesd to installation:
    try autodiscovery for DNS record
if installation has no ntp options:
    skip configuration and use default one

synchronize time
```
Would that be acceptable alongside with update of design document?

rcritten: I agree with the steps and the purpose. I found it quite jarring to be prompted for NTP after answering yes to proceed with the installation, so I'm glad you're addressing this!
Is adding a new set of variables, ntp_server/ntp_pool so bad? It seems wrong to twiddle with options.*. We do this other places in the code as well, e.g. hostname, cli_domain, cli_server,...

Tiboris: Sorry I have misunderstood. Updating PR.

Tiboris: https://github.com/freeipa/freeipa/blob/254ba5240e897da04eea06b4f02dea579a9a6c54/ipaclient/install/client.py#L2476

ipatests/test_integration/test_external_ca.py
```diff
@@ -83,7 +83,7 @@ def install_server_external_ca_step1(host, extra_args=()):
 def install_server_external_ca_step2(host, ipa_ca_cert, root_ca_cert,
                                      raiseonerr=True):
     """Step 2 to install the ipa server with external ca"""
-    args = ['ipa-server-install',
+    args = ['ipa-server-install', '-U', '-r', host.domain.realm,
             '-a', host.config.admin_password,
```
rcritten: Is this change related?

Tiboris: Yes @rcritten,

I found out that it is the external CA second step that was  causing a timeout for PRCI (see so many re-runs). Because of that I added an unattended option to this task. Without it it got stuck at new NTP prompt and then timeouted.

However it may happen that more nightly tests which rely on a fact that if all necessary options were passes an unattended option is not needed.

I see missing unattended option here as well because test is passing stdin text:

https://github.com/freeipa/freeipa/blob/master/ipatests/test_integration/test_installation.py#L587-L572

for this one a stdin input needs to be updated to have a ntp ptompt stdin text passed to installer.

Tiboris:
master:

* e3f35843dc33fa6a3fb3071c5c89ecde9db1ce5e Moving prompt for NTP options to install_check
* be39d3a99bfd676cd1661af7e8b2553337b4e1c8 ipatests: Add Unattended option to external ca task

PR #3110, opened 2019-05-03
Index certmap attributes

abbra:
During an investigation into filter optimisation in 389DS it was
discovered that two attributes of the certmap query are unindexed.
Due to the nature of LDAP filters, if any member of an OR query is
unindexed, the entire OR becomes unindexed.

This is then basically a full-table scan, which applies the filter test
to the contained members.

Fixes: https://pagure.io/freeipa/issue/7932
Fixes: https://pagure.io/freeipa/issue/7933

### DIFF
--- a/install/updates/20-aci.update
+++ b/install/updates/20-aci.update
@@ -25,6 +25,10 @@ add:aci:(targetfilter="(objectclass=domain)")(targetattr="objectclass || dc || i
 dn: $SUFFIX
 add:aci:(targetattr="parentid")(version 3.0; acl "Anonymous read access to parentID information"; allow(read, search, compare) userdn = "ldap:///anyone";)
 
+# Read access to altSecurityIdentities to allow filter optimizations in 389-ds
+dn: $SUFFIX
+add:aci:(targetattr="altSecurityIdentities")(version 3.0; acl "Authenticated read access to altSecurityIdentities information"; allow(read, search, compare) userdn = "ldap:///all";)
+
 # Read access to containers
 dn: $SUFFIX
 add:aci:(targetfilter="(&(objectclass=nsContainer)(!(objectclass=krbPwdPolicy)))")(target!="ldap:///cn=masters,cn=ipa,cn=etc,$SUFFIX")(targetattr="objectclass || cn")(version 3.0; acl "Anonymous read access to containers"; allow(read, search, compare) userdn = "ldap:///anyone";)
--- a/ipaserver/plugins/certmap.py
+++ b/ipaserver/plugins/certmap.py
@@ -91,6 +91,67 @@ EXAMPLES:
 register = Registry()
 
 
+def check_maprule_is_for_trusted_domain(api_inst, options, entry_attrs):
+    """
+    Check that the certmap rule references altSecurityIdentities and
+    associateddomain in options is a trusted domain.
+
+    :param api_inst: API instance
+    :param options: options passed to the command
+                    at least ipacertmapmaprule and
+                    associateddomain options are expected for the check
+
+    :raises: ValidationError if altSecurityIdentities is present in the
+             rule but no Active Directory trusted domain listed in
+             associated domains
+    """
+    is_trusted_domain_required = False
+
+    # If no explicit option passed, fallback to the content of the entry.
+    # This helps to catch cases when an associated domain value is removed
+    # while the rule requires its presence.
+    #
+    # In certmap-mod pre-callback we pass LDAPEntry instance instead of a dict
+    # LDAEntry.get() returns lists, even for single valued attrs. Using
+    # LDAPEntry.single_value would give us a single-valued result instead.
+    maprule_entry = entry_attrs
+    if not isinstance(maprule_entry, dict):
+        maprule_entry = entry_attrs.single_value
+
+    maprule = options.get('ipacertmapmaprule',
+                          maprule_entry.get('ipacertmapmaprule'))
+
+    if maprule:
+        if 'altsecurityidentities' in maprule.lower():
+            is_trusted_domain_required = True
+
+    if is_trusted_domain_required:
+        domains = options.get('associateddomain',
+                              entry_attrs.get('associateddomain'))
+        if domains:
+            trusted_domains = api_inst.Object.config.gather_trusted_domains()
+            trust_suffix_namespace = {dom_name.lower() for dom_name in
+                                      trusted_domains}
+
+            candidates = {str(dom).lower() for dom in domains}
+            invalid = candidates - trust_suffix_namespace
+
+            if invalid == candidates or len(trust_suffix_namespace) == 0:
+                raise errors.ValidationError(
+                    name=_('domain'),
+                    error=_('The domain(s) "%s" cannot be used to apply '
+                            'altSecurityIdentities check.') %
+                    ", ".join(list(invalid))
+                )
+        else:
+            raise errors.ValidationError(
+                name=_('domain'),
+                error=_('The mapping rule with altSecurityIdentities '
+                        'should be applied to a trusted Active Directory '
+                        'domain but no domain was associated with the rule.')
+            )
+
+
 def check_associateddomain_is_trusted(api_inst, options):
     """
     Check that the associateddomain in options are either IPA domain or
@@ -299,6 +360,7 @@ class certmaprule_add(LDAPCreate):
     def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys,
                      **options):
         check_associateddomain_is_trusted(self.api, options)
+        check_maprule_is_for_trusted_domain(self.api, options, entry_attrs)
         return dn
 
 
@@ -311,6 +373,17 @@ class certmaprule_mod(LDAPUpdate):
     def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys,
                      **options):
         check_associateddomain_is_trusted(self.api, options)
+        # For update of the existing certmaprule we need to retrieve
+        # content of the LDAP entry because modification may affect two cases:
+        # - altSecurityIdentities might be removed by the modification
+        # - trusted domains may be removed by the modification while they
+        #   should be in place
+        #
+        # For both these cases we need to know actual content of the entry but
+        # LDAPUpdate.execute() provides us with entry_attrs built from the
+        # options, not the original content.
+        entry = ldap.get_entry(dn)
+        check_maprule_is_for_trusted_domain(self.api, options, entry)
         return dn
 
 
--- a/ipatests/test_xmlrpc/test_certmap_plugin.py
+++ b/ipatests/test_xmlrpc/test_certmap_plugin.py
@@ -27,6 +27,17 @@ certmaprule_create_params = {
         u'ipacertmappriority': u'1',
 }
 
+certmaprule_create_trusted_params = {
+    u'cn': u'test_trusted_rule',
+    u'description': u'Certificate mapping and matching rule for test '
+                    u'purposes for trusted domain',
+    u'ipacertmapmaprule': u'altsecurityidentities=X509:<some map>',
+    u'ipacertmapmatchrule': u'arbitrary free-form matching rule defined '
+                            u'and consumed by SSSD',
+    u'associateddomain': api.env.domain,
+    u'ipacertmappriority': u'1',
+}
+
 certmaprule_update_params = {
         u'description': u'Changed description',
         u'ipacertmapmaprule': u'changed arbitrary mapping rule',
@@ -77,6 +88,12 @@ def certmap_rule(request):
 
 
 @pytest.fixture(scope='class')
+def certmap_rule_trusted_domain(request):
+    tracker = CertmapruleTracker(**certmaprule_create_trusted_params)
+    return tracker.make_fixture(request)
+
+
+@pytest.fixture(scope='class')
 def certmap_config(request):
     tracker = CertmapconfigTracker()
     return tracker.make_fixture(request)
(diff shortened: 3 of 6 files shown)


abbra:
This PR now features following operations:
- add `altSecurityIdentities` attribute to IPA LDAP schema
- allow authenticated binds to read altSecurityIdentities attribute
- index both `ipacertmapdata` and `altsecurityidentities`
- add a check to `ipa certmaprule-{add,mod}` that will enforce adding `altSecurityIdentities` filter only to trusted Active Directory domains.

While we aren't going to use `altSecurityIdentities` attribute ourselves, existing rules need to be processed properly, thus the addition of the attribute to schema. ACI is needed to allow 389-ds to perform filter optimizations. Finally, the check in `ipa certmaprule-{add,mod}` commands will help to associate `altSecurityIdentities` only with Active Directory domains to allow SSSD to avoid hammering 389-ds with 'wrong filter'.

ipaserver/plugins/certmap.py
```diff
@@ -91,6 +91,49 @@
...
+                    associateddomain options are expected for the check
+
+    :raises: ValidationError if altSecurityIdentities is present in the
+             rule but no Active Directory trusted domain listed in
+             associated domains
+    """
+    is_trusted_domain_required = False
+    maprules = options.get('ipacertmapmaprule')
+    if maprules:
+        for maprule in maprules:
```
flo-renaud: ipacertmapmaprule is an optional single-valued option => maprules is a string, not a list of string. No need to use this loop, you can just use `if 'altsecurityidentities' in maprules.lower():`
lower() is needed to make a case-insensitive comparison.

install/updates/20-aci.update
```diff
@@ -25,6 +25,10 @@ add:aci:(targetfilter="(objectclass=domain)")(targetattr="objectclass || dc || i
 dn: $SUFFIX
 add:aci:(targetattr="parentid")(version 3.0; acl "Anonymous read access to parentID information"; allow(read, search, compare) userdn = "ldap:///anyone";)
 
+# Read access to altSecurityIdentities to allow filter optimizations in 389-ds
+dn: $SUFFIX
```
rcritten: Does certmap only apply to users? If so it would be better to put this into the user container otherwise it would get evaluated with every request.

ipaserver/plugins/certmap.py
```diff
@@ -91,6 +91,49 @@
...
+            if 'altSecurityIdentities' in maprule:
```
rcritten: Just curios, should this be case-insensitive?

install/updates/20-aci.update
abbra: Certmap rules can be associated with ID Overrides too, I think. For altSecurityIdentities, these, technically, can be present in any bindable object.

ipaserver/plugins/certmap.py
abbra: yes, the search should be case-insensitive. Good catch.

ipaserver/plugins/certmap.py
abbra: Thanks, I pushed an update that treats `options.get('ipacertmapmaprule')` as a sting.

ipatests/test_xmlrpc/test_certmap_plugin.py
```diff
@@ -122,6 +139,23 @@ def test_delete(self, certmap_rule):
...
+        ]),
+        ids=dontfill_idfn,
+    )
+    def test_failed_create(self, dont_fill, certmap_rule_trusted_domain):
+        certmap_rule_trusted_domain.ensure_missing()
+        try:
+            certmap_rule_trusted_domain.create(dont_fill)
+        except errors.ValidationError:
+            pass
+        finally:
```
flo-renaud: The try/except clause is missing a else: statement. When the certmaprule-create does not raise the expected ValidationError, the test should exit on error.

abbra: Thanks. I added raising an assertion error for this case.

abbra:
I pushed an update that moved `finally:` part of the test to `else:`. In case test passes unexpectedly we remove the rule. However, if test generates an exception, the rule is not there and it cannot be removed so `finally:` part actually produced an exception.

ipatests/test_xmlrpc/test_certmap_plugin.py
```diff
@@ -122,6 +139,25 @@ def test_delete(self, certmap_rule):
         certmap_rule.ensure_exists()
         certmap_rule.delete()
 
+    @pytest.mark.parametrize(
```
flo-renaud: With this parametrize annotation, the test will be run with all combinations of missing params. When all params except ipacertmapmaprule are provided, the test won't fail while the code expects it to fail.
The test can probably be simplified (no need to test all combinations of parameters, we are simply interested in checking that altsecurityidentites in the maprule triggers a validation error).

abbra: Yes, I removed all that complexity. A new test is running -- I need to understand why ensure_missing() tracker calls are causing exceptions as well.

abbra:
The test seems to work:
```
Ran command: ipaserver.plugins.certmap.certmaprule_add()('test_trusted_rule', description='Certificate mapping and matching rule for test purposes for trusted domain', ipacertmapmaprule='altsecurityidentities=X509:<some map>', ipacertmapmatchrule='arbitrary free-form matching rule defined and consumed by SSSD', associateddomain='ipa.test', ipacertmappriority='1', version='2.231'): ValidationError: invalid 'domain': The domain ipa.test is not a trusted domain to apply altSecurityIdentities check.
```
I still need to figure out why cleanup of the test still fails (it runs `ipa certmaprule-del` and doesn't handle situation when rule does not exist).

Firstyear:
Thanks for this change, I think it will really help IPA performance too with this feature.

abbra:
@flo-renaud thanks, I added a fallback to pick up the values from the `entry_attrs`. This should cover the case when AD domain is being removed from the rule but the filter still references altSecurityIdentities.

ipaserver/plugins/certmap.py
```diff
@@ -91,6 +91,53 @@
...
+    # while the rule requires its presence
+    maprule = options.get('ipacertmapmaprule',
+                          entry_attrs.get('ipacertmapmaprule'))
+    if maprule:
+        if 'altsecurityidentities' in maprule.lower():
+            is_trusted_domain_required = True
+
+    if is_trusted_domain_required:
+        domains = options.get('associateddomain',
+                              entry_attrs.get('ipacertmapmaprule'))
```
flo-renaud: Should be entry_attrs.get('associateddomain')

abbra: @flo-renaud thanks, silly copy/paste mistake. Fixed.

ipaserver/plugins/certmap.py
```diff
@@ -91,6 +91,53 @@
...
+            is_trusted_domain_required = True
+
+    if is_trusted_domain_required:
+        domains = options.get('associateddomain',
+                              entry_attrs.get('associateddomain'))
+        if domains:
+            trusted_domains = api_inst.Object.config.gather_trusted_domains()
+            trust_suffix_namespace = {dom_name.lower() for dom_name in
+                                      trusted_domains}
+            for dom in domains:
```
flo-renaud: The check should rather be that at least one of the domains is in trust_suffix_namespace. For instance, if `--domain ipa.domain.com --domain ad.domain.com` are specified, a maprule with altsecurityIdentities should be accepted (as in the example 23.4 of [Examples of Identity Mapping Rules](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/linux_domain_identity_authentication_and_policy_guide/smart-cards#sc-id-mapping-rule-examples) where a single rule is used for all the domains).

abbra: @flo-renaud Thanks. I think new code covers it.

flo-renaud:
Hi @abbra 
with the latest pr, it is still possible to add a certmaprule with IPA domain while it should be refused. Shouldn't the line 
```invalid = trust_suffix_namespace - candidates```
be replaced with 
``` invalid = candidates - trust_suffix_namespace```?
Same comment about replacing the domain with a value that is not a trusted domain.

**Comment by abbra** [comment]
@flo-renaud thanks. As discussed on IRC, I changed the code to your suggestion. I also looked into what we actually receive in the case of modification and apparently, we need to retrieve the actual entry content to do validation in the case when someone adds `altSecurityIdentities` but not a trusted domain reference or removes trusted domain reference while keeping `altSecurityIdentities` in the rule.
Since modification of a rule is a rare operation, we can afford this behavior.

File: ipaserver/plugins/certmap.py [hunk 8]
```diff
@@ -91,6 +91,57 @@
...
+    :raises: ValidationError if altSecurityIdentities is present in the
+             rule but no Active Directory trusted domain listed in
+             associated domains
+    """
+    is_trusted_domain_required = False
+    # If no explicit option passed, fallback to the content of the entry
+    # This helps to catch cases when an associated domain value is removed
+    # while the rule requires its presence
+    maprule = options.get('ipacertmapmaprule',
+                          entry_attrs.get('ipacertmapmaprule'))
```
flo-renaud: should be entry_attrs.single_value.get('ipacertmaprule') otherwise a list is returned and the call to maprule.lower() fails.

abbra:
Two tests failing are related to the pytest issues @stanislavlevin identified. Namely, there is out of order teardown execution and related lack of an DNS record object that prevents creating a host in an unrelated test. Hopefully, we can get pytest fixes in soon.

stanislavlevin:
@abbra, https://pagure.io/freeipa/issue/7989 issue is related to pytest4+.

But I could easily check this one, thanks to Azure container artifact :wink: 

Concerning the mentioned issue, I'm waiting for any feedback or thoughts.

stanislavlevin:
Actually, the second test ( `test_xmlrpc/test_host_plugin.py::TestHostDNS::test_add_ipv6_host_from_ip` ) passed.

But since setup plan for this test class is as follow:
```
ipa-run-tests -v test_xmlrpc/test_host_plugin.py::TestHostDNS --setup-plan

usr/lib/python3.7/site-packages/ipatests/test_xmlrpc/test_host_plugin.py::TestHostDNS::test_add_ipv6only_host 
      SETUP    C host
      SETUP    C dns_setup (fixtures used: host)
      SETUP    C ipv6only_host
        usr/lib/python3.7/site-packages/ipatests/test_xmlrpc/test_host_plugin.py::TestHostDNS::()::test_add_ipv6only_host (fixtures used: dns_setup, host, ipv6only_host)
usr/lib/python3.7/site-packages/ipatests/test_xmlrpc/test_host_plugin.py::TestHostDNS::test_add_ipv4only_host 
      TEARDOWN C dns_setup
      TEARDOWN C host
      SETUP    C host
      SETUP    C dns_setup (fixtures used: host)
      SETUP    C ipv4only_host
        usr/lib/python3.7/site-packages/ipatests/test_xmlrpc/test_host_plugin.py::TestHostDNS::()::test_add_ipv4only_host (fixtures used: dns_setup, host, ipv4only_host)
usr/lib/python3.7/site-packages/ipatests/test_xmlrpc/test_host_plugin.py::TestHostDNS::test_add_ipv46both_host 
      TEARDOWN C dns_setup
      TEARDOWN C host
      SETUP    C host
      SETUP    C dns_setup (fixtures used: host)
      SETUP    C ipv46both_host
        usr/lib/python3.7/site-packages/ipatests/test_xmlrpc/test_host_plugin.py::TestHostDNS::()::test_add_ipv46both_host (fixtures used: dns_setup, host, ipv46both_host)
usr/lib/python3.7/site-packages/ipatests/test_xmlrpc/test_host_plugin.py::TestHostDNS::test_add_ipv4_host_from_ip 
      TEARDOWN C dns_setup
      TEARDOWN C host
      SETUP    C host
      SETUP    C dns_setup (fixtures used: host)
      SETUP    C ipv4_fromip_host
        usr/lib/python3.7/site-packages/ipatests/test_xmlrpc/test_host_plugin.py::TestHostDNS::()::test_add_ipv4_host_from_ip (fixtures used: dns_setup, host, ipv4_fromip_host)
usr/lib/python3.7/site-packages/ipatests/test_xmlrpc/test_host_plugin.py::TestHostDNS::test_add_ipv6_host_from_ip 
      TEARDOWN C dns_setup
      TEARDOWN C host
      SETUP    C host
      SETUP    C dns_setup (fixtures used: host)
      SETUP    C ipv6_fromip_host
        usr/lib/python3.7/site-packages/ipatests/test_xmlrpc/test_host_plugin.py::TestHostDNS::()::test_add_ipv6_host_from_ip (fixtures used: dns_setup, host, ipv6_fromip_host)
      TEARDOWN C ipv6_fromip_host
      TEARDOWN C dns_setup
      TEARDOWN C host
      TEARDOWN C ipv4_fromip_host
      TEARDOWN C ipv46both_host
      TEARDOWN C ipv4only_host
      TEARDOWN C ipv6only_host
```
The teardown of `ipv46both_host` fixture failed right after the last test method in this class ( resources deallocation for class scope happens here).
That's why we see `test_xmlrpc/test_host_plugin.py::TestHostDNS::test_add_ipv6_host_from_ip [31mERROR`
Actually, only ` test_xmlrpc/test_host_plugin.py::TestHostDNS::test_add_ipv46both_host` failed.
For some reason, during host adding the dns query reported missing DNS A/AAAA record, though they have been added before.
It may be for an external reason (host system), since python dns reads '/etc/resolvconf' and caches results (default lifetime 5min).

flo-renaud:
Hi @abbra 
thanks for the update. Works for me, and the tests are now green. ACK.

abbra:
master:

* 5a83eea20bdd0c1c6302eada774ef3d8cfa04e36 Add altSecurityIdentities attribute from MS-WSPP schema definition
* 725899595f42b607989db96d849e1b4f01de7faa Create indexes for altSecurityIdentities and ipaCertmapData attributes
* 41ca4d484e7492bf469d73c6e627cc6df46eebf8 certmap rules: altSecurityIdentities should only be used for trusted domains
* 95c2b34c4b468a9b766b24172e34ff2d8b6c9530 certmaprule: add negative test for altSecurityIdentities

PR #3111, opened 2019-05-03
ipa-client-automount: add knob to configure NFSv4 Domain (idmapd.conf)

fcami:
ipa-client-automount assumes the NFS domain to be the same as the IPA domain.
This is not always the case.
This commit adds a --idmap-domain knob with the following behavior:
- if not present, default to IDM domain (current behavior)
- if equal to DNS (magic value), set nothing and let idmapd autodetect domain
- otherwise set Domain in idmap.conf to the value passed by this parameter
    
Fixes: https://pagure.io/freeipa/issue/7918

### DIFF
--- a/ipatests/test_integration/test_nfs.py
+++ b/ipatests/test_integration/test_nfs.py
@@ -9,8 +9,7 @@
    * add automount direct and indirect maps
    * add automount /home for the "seattle" location only
    * validate it is not available in another location
-   * krb5 /home for IdM users in test_krb5_nfs_manual_configuration
-   * krb5 /home for IdM users in test_automount_location
+   * krb5 /home for IdM users in test_automount
    * store nfs configuration in a single place
 """
 
@@ -189,9 +188,7 @@ class TestNFS(TestInit):
             "%s:/exports/home" % nfssrv.hostname, "/home", "-v"
         ])
 
-        # TODO leverage users
-
-    def test_automount_location(self):
+    def test_automount(self):
         """
         Test if ipa-client-automount behaves as expected
         """
@@ -236,6 +233,12 @@ class TestNFS(TestInit):
         # maybe re-use m1.group(0) if it exists.
         assert m1 is None
 
+        # https://pagure.io/freeipa/issue/7918
+        # check whether idmapd.conf was setup using the IPA domain
+        automntclt.run_command([
+            "grep", "Domain = %s" % self.master.domain.name, "/etc/idmapd.conf"
+        ])
+
         time.sleep(WAIT_AFTER_INSTALL)
 
         automntclt.run_command([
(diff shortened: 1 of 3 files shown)


ipatests/test_integration/test_nfs.py
```diff
@@ -236,22 +233,61 @@ def test_automount_location(self):
         # maybe re-use m1.group(0) if it exists.
         assert m1 is None
 
+        # https://pagure.io/freeipa/issue/7918
+        # check whether idmapd.conf was setup using the IPA domain
+        automntclt.run_command([
+            "grep", "Domain = %s" % self.master.domain.name, "/etc/idmapd.conf"
+        ])
```
rcritten: So you're relying on raising an exception?

rcritten: If it fails will there still be cleanup?

ipatests/test_integration/test_nfs.py
```diff
@@ -236,22 +233,61 @@ def test_automount_location(self):
...
+        nfs_domain = "exampledomain.net"
+        automntclt.run_command([
+            'ipa-client-automount', '--location', 'default',
+            '-U', '--debug', "--idmap-domain", nfs_domain
+        ])
+        # check whether idmapd.conf was setup using nfs_domain
+        automntclt.run_command([
+            "grep", "Domain = %s" % nfs_domain, "/etc/idmapd.conf"
+        ])
+
```
rcritten: Similar here, on failure will it leave things installed?

rcritten:
Needs a man page update to describe the new option and the three possible settings: use IPA domain (default), DNS string to not set anything or user-provided.

ipatests/test_integration/test_nfs.py
fcami: It should break execution if the return code is not 0.

ipatests/test_integration/test_nfs.py
fcami: It should and it's probably better for investigation I guess. Would you like anything changed?

fcami:
Thanks for catching the man page issue. PR updated.

fcami:
@rcrit let me know if I should remove the temp commit or if you want the test handled differently. I'm afraid the whole NFS test suite behaves the same way.

ipatests/test_integration/test_nfs.py
rcritten: It's just that most tests clean themselves up even on failures.

rcritten:
Other tests don't seem to clean up on failure either so ack.

fcami:
master:

* 660c4984c6281055752d74dc276fc39a07da324a ipa-client-automount: add knob to configure NFSv4 Domain (idmapd.conf)
* d76737e4c6131ae6b4f550b2b7c1a19304f1e3a8 ipatests: add tests for the new NFSv4 domain option of ipa-client-automount

PR #3112, opened 2019-05-03
ipa-backup: better error message if ENOSPC

fcami:
When the destination directory cannot store the complete backup
ipa-backup fails but does not explain why.
This commit adds error-checking to db2ldif(), db2bak() and
finalize_backup() and enhances the error message.

Fixes: https://pagure.io/freeipa/issue/7647

### DIFF
--- a/ipaserver/install/ipa_backup.py
+++ b/ipaserver/install/ipa_backup.py
@@ -294,8 +294,7 @@ class Backup(admintool.AdminTool):
         os.chown(self.top_dir, pent.pw_uid, pent.pw_gid)
         os.chmod(self.top_dir, 0o750)
         self.dir = os.path.join(self.top_dir, "ipa")
-        os.mkdir(self.dir)
-        os.chmod(self.dir, 0o750)
+        os.mkdir(self.dir, mode=0o750)
         os.chown(self.dir, pent.pw_uid, pent.pw_gid)
         self.tarfile = None
 
@@ -442,11 +441,16 @@ class Backup(admintool.AdminTool):
             try:
                 conn.add_entry(ent)
             except Exception as e:
-                raise admintool.ScriptError('Unable to add LDIF task: %s'
-                    % e)
+                raise admintool.ScriptError(
+                    'Unable to add LDIF task: %s' % e
+                )
 
             logger.info("Waiting for LDIF to finish")
-            wait_for_task(conn, dn)
+            if (wait_for_task(conn, dn) != 0):
+                raise admintool.ScriptError(
+                    'BAK online task failed. Check file systems\' free space.'
+                )
+
         else:
             args = [paths.DB2LDIF,
                     '-Z', instance,
@@ -455,10 +459,25 @@ class Backup(admintool.AdminTool):
                     '-a', ldiffile]
             result = run(args, raiseonerr=False)
             if result.returncode != 0:
-                logger.critical('db2ldif failed: %s', result.error_log)
+                raise admintool.ScriptError(
+                    'db2ldif failed: %s '
+                    'Check if destination directory %s has enough space.'
+                    % (result.error_log, os.path.dirname(ldiffile))
+                )
 
         # Move the LDIF backup to our location
-        shutil.move(ldiffile, os.path.join(self.dir, ldifname))
+        try:
+            shutil.move(ldiffile, os.path.join(self.dir, ldifname))
+        except (IOError, OSError) as e:
+            raise admintool.ScriptError(
+                'Unable to move LDIF: %s '
+                'Check if destination directory %s has enough space.'
+                % (e, os.path.dirname(ldiffile))
+            )
+        except Exception as e:
+            raise admintool.ScriptError(
+                'Unexpected error: %s' % e
+            )
 
 
     def db2bak(self, instance, online=True):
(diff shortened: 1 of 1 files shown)


fcami:
Depending on the amount of space left ipa-backup could error out when moving a tar file around or when creating the LDIFs. It is unclear how to properly test and account for all situations.

fcami:
Hi @frenaud good catch! 
Do you have the traceback handy? I was able to make online backup crash while doing shutil.move but nowhere else.

ipaserver/install/ipa_backup.py
```diff
@@ -455,10 +460,21 @@ def db2ldif(self, instance, backend, online=True):
...
+                    'db2ldif failed: %s '
+                    'Check if destination directory %s has enough space.'
+                    % (result.error_log, os.path.dirname(ldiffile))
+                )
 
         # Move the LDIF backup to our location
-        shutil.move(ldiffile, os.path.join(self.dir, ldifname))
+        try:
+            shutil.move(ldiffile, os.path.join(self.dir, ldifname))
+        except Exception as e:
```
Tiboris: Just comment, do we really want to have regular Exception here? Would not be better not to catch all of them?
Maaaybe OSError or IOError would be enough, or even shutil.Error.

ipaserver/install/ipa_backup.py
```diff
@@ -619,33 +652,45 @@ def finalize_backup(self, data_only=False, encrypt=False, keyring=None):
...
-                '-czf',
-                filename,
-                '.'
-               ]
-        result = run(args, raiseonerr=False)
-        if result.returncode != 0:
+        try:
+            os.mkdir(backup_dir)
+            os.chmod(backup_dir, 0o700)
+        except Exception as e:
```
Tiboris: Similar as here: https://github.com/freeipa/freeipa/pull/3112#pullrequestreview-233924039

ipaserver/install/ipa_backup.py
```diff
@@ -489,18 +505,33 @@ def db2bak(self, instance, online=True):
...
-
-        shutil.move(bakdir, self.dir)
+                raise admintool.ScriptError(
+                    'db2bak failed: %s '
+                    'Check if destination directory %s has enough space.'
+                    % (result.error_log, bakdir)
+                )
+        try:
+            shutil.move(bakdir, self.dir)
+        except Exception as e:
```
Tiboris: Similar as here: https://github.com/freeipa/freeipa/pull/3112#pullrequestreview-233924039

ipaserver/install/ipa_backup.py
fcami: shutil would fail with FileNotFoundError in copyfile. FileNotFoundError is a subclass of OSError.

ipaserver/install/ipa_backup.py
fcami: os.mkdir would fail with OSError I think.

Tiboris: Of course @fcami I only meant to specify Exception type but thanks for correction :)

fcami: In fact that's what I'm concerned about @Tiboris - Python 2.7 would behave significantly differently if I remember correctly (IOError vs OSError).
As the try statement only protects OS-specific calls I'm not sure what would be gained by catching specific exceptions there. Also ipa_backup.py already behaves that way everywhere (see exception handling after LDAP connections).

ipaserver/install/ipa_backup.py
t-woerner: Are generic errors handled and visible on the command line? If not, then please add the specific and also the generic error handling.

fcami: hi @t-woerner good idea! The last revision should address @Tiboris concern to your liking.

ipaserver/install/ipa_backup.py
fcami: Fixed, thanks!

ipaserver/install/ipa_backup.py
fcami: Resolved, see above. Thanks!

ipaserver/install/ipa_backup.py
```diff
@@ -619,33 +660,52 @@ def finalize_backup(self, data_only=False, encrypt=False, keyring=None):
...
+            backup_dir = os.path.join(
+                paths.IPA_BACKUP_DIR,
+                time.strftime('ipa-full-%Y-%m-%d-%H-%M-%S')
+            )
             filename = os.path.join(backup_dir, "ipa-full.tar")
 
-        os.mkdir(backup_dir)
-        os.chmod(backup_dir, 0o700)
+        try:
+            os.mkdir(backup_dir)
```
tiran: Please use the mode argument of os.mkdir instead of a separate chmod.

fcami: Fixed, thanks @tiran

flo-renaud:
Hi @fcami 
the PR looks good to me. Let's wait for PRCI to finish, then you can remove the temp commit.

flo-renaud:
@fcami 
Temp tests are green, you can remove the temp commit.

fcami:
Passes CI ( http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/4cdbe066-701c-11e9-a493-fa163e0737bc/ ). Removing temp commit.

flo-renaud:
Tests are green, ACK.

fcami:
Thanks @flo-renaud

fcami:
master:

* e6415ec3210cda0808867642cb19f0df1ddd3c45 ipa-backup: better error message if ENOSPC

PR #3123, opened 2019-05-08
Fix `build_requestinfo` in OpenSSL1.1.0+ environments

stanislavlevin:
Since OpenSSL 1.1.0 the `req_info` field of X509_REQ structure is
no longer a pointer to X509_REQ_INFO. This results in a crash of
`build_requestinfo` in environments having OpenSSL1.1.0+ (libcrypto).

With this patch, the X509_REQ definition becomes the version dependent.
Both OpenSSL1.0.x and OpenSSL1.1.x are supported.

Fixes: https://pagure.io/freeipa/issue/7937

### DIFF
--- a/ipaclient/csrgen_ffi.py
+++ b/ipaclient/csrgen_ffi.py
@@ -6,6 +6,29 @@ from ipalib import errors
 _ffi = FFI()
 
 _ffi.cdef('''
+unsigned long OpenSSL_version_num(void);
+unsigned long SSLeay(void);
+''')
+
+_libcrypto = _ffi.dlopen(ctypes.util.find_library('crypto'))
+
+# SSLeay has been renamed with OpenSSL_version_num in OpenSSL 1.1.0
+try:
+    OpenSSL_version_num = _libcrypto.OpenSSL_version_num
+except AttributeError:
+    OpenSSL_version_num = _libcrypto.SSLeay
+
+# OpenSSL_version_num()/SSLeay() returns the value of OPENSSL_VERSION_NUMBER
+#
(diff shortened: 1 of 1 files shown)


ipaclient/csrgen_ffi.py
```diff
@@ -118,10 +149,7 @@
 char *ERR_error_string(unsigned long e, char *buf);
 ''')  # noqa: E501
 
-_libcrypto = _ffi.dlopen(ctypes.util.find_library('crypto'))
```
abbra: Why are you removing this line? The code will break because we reference `_libcrypto` below.

stanislavlevin: @abbra, thank you for looking at this. 
Actually, we need to know OpenSSL version ( `OpenSSL_version_num` / `SSLeay`) before parsing the related C code (`cdef`). Thus, the loading libcrypto in ABI mode happened earlier (see line 13, please).

abbra: Ah, thanks. I missed that part. It looks OK to me but fedora-29/caless failed due to a timeout. I restarted the test.

abbra:
Tests completed. @tiran, please help with the review too.

abbra:
I had to pull this commit into my tree https://github.com/abbra/freeipa/pull/11 to make it possible to run Tox tests on Fedora 30 (that pull request is against my own fork and is running tests against Azure Pipelines runners). So ACK.

fcami:
master:

* ac6568dcf58ec8d06df5493d14a28aa41845d4ef Fix `build_requestinfo` in OpenSSL1.1.0+ environments

stanislavlevin:
@abbra, thank you!

PR #3130, opened 2019-05-11
upgrade: adtrust - catch empty result when retrieving list of trusts

abbra:
Upgrade failure when ipa-server-upgrade is being run on a system with no
trust established but trust configured

Fixes: https://pagure.io/freeipa/issue/7939


abbra:
`fedora-29/test_kerberos_flags` stumbled on secure NFS taking over kadmind port. Rerunning.

fcami:
master:

* 98b4c710d90f289322ebda457fdb84c2dd34aace upgrade: adtrust - catch empty result when retrieving list of trusts

PR #2716, opened 2018-12-27
[FileStore] Add support for the stacked backup

stanislavlevin:
The current implementation doesn't ensure the order of backup/restoration. There are use cases requiring the multiple backups of a file with same name but different content. And for these situations, FileStore is not predictable.
    
* to ensure the order usual dict storage is replaced with OrderedDict one
* all search operations are performed in a reversed order
* add tests
* fix pylint errors
    
Fixes: https://pagure.io/freeipa/issue/7820

PR #3137, opened 2019-05-13
Azure pipelines setup

abbra:
Sets up a pipeline to run FreeIPA build and tests in Azure Pipelines.
    
Azure Pipelines provides 10 parallel free runners for open source projects.
Use them to run following jobs:
    
- Build: build RPMs and Fedora 30 container with them
- Lint: run linting of the source code
- Tox: run py36,pypi,pylint tests using Tox
- Web UI unit tests: run Web UI unit tests with Grunt/QUnit/PhantomJS
- XMLRPC tests: install FreeIPA server and run XMLRPC tests against it
    
All jobs are running in Fedora 30 containers. Build, Lint, Tox, and Web UI unit tests run inside f30/fedora-toolbox container. Build job generates a container with pre-installed FreeIPA packages using official fedora:30 container. All containers are picked up from the registry.fedoraproject.org.
    
Artifacts from the build job are pushed to a pipeline storage and reused in the XMLRPC tests. They also are accessible in the 'Summary' tab to download.
    
XUnit and QUnit outputs from the tests that produce it are reported in the 'Tests' tab.
    
Logs from individual steps from each job are available for review in the 'Logs' tab. They also can be downloaded.

### DIFF
--- a/ipatests/azure/azure-run-tests.sh
+++ b/ipatests/azure/azure-run-tests.sh
@@ -1,0 +1,43 @@
+#!/bin/bash -ex
+server_realm=EXAMPLE.TEST
+server_domain=example.test
+server_password=Secret123
+
+# Expand list of tests into -k...  -k... -k... .. sequence
+tests_to_run=$(eval "eval echo -k{$(echo $TESTS_TO_RUN | sed -e 's/[ \t]+*/,/g')}")
+
+systemctl --now enable firewalld
+ipa-server-install -U --domain ${server_domain} --realm ${server_realm} -p ${server_password} -a ${server_password} --setup-dns --setup-kra --auto-forwarders
+sed -ri "s/mode = production/mode = development/" /etc/ipa/default.conf
+systemctl restart httpd.service
+firewall-cmd --add-service={freeipa-ldap,freeipa-ldaps,dns}
+
+
+cd /freeipa
+
+echo ${server_password} | kinit admin && ipa ping
+cp -r /etc/ipa/* ~/.ipa/
+echo ${server_password} > ~/.ipa/.dmpw
+echo 'wait_for_dns=5' >> ~/.ipa/default.conf
+ipa-test-config --help
+ipa-test-task --help
+ipa-run-tests --with-xunit -k-{test_integration,test_webui,test_ipapython/test_keyring.py,test_dns_soa} -v ${tests_to_run}
+grep -n -C5 BytesWarning /var/log/httpd/error_log
+ipa-server-install --uninstall -U
+# second uninstall to verify that --uninstall without installation works
+ipa-server-install --uninstall -U
+firewall-cmd --remove-service={freeipa-ldap,freeipa-ldaps,dns}
+
+mkdir -p /freeipa/logs
+cd /freeipa/logs
+journalctl -b --no-pager > systemd_journal.log
+tar --ignore-failed-read -cvf var_log.tar \
+    /var/log/dirsrv \
+    /var/log/httpd \
+    /var/log/ipa* \
+    /var/log/krb5kdc.log \
+    /var/log/pki \
+    systemd_journal.log
+
+ls -laZ /etc/dirsrv/slapd-*/ /etc/httpd/alias/ /etc/pki/pki-tomcat/alias/ || true
+
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -170,7 +170,7 @@ BuildRequires:  libuuid-devel
 BuildRequires:  libsss_idmap-devel
 BuildRequires:  libsss_certmap-devel
 BuildRequires:  libsss_nss_idmap-devel >= %{sssd_version}
-BuildRequires:  nodejs
+BuildRequires:  nodejs(abi) < 11
 BuildRequires:  uglify-js
 BuildRequires:  libverto-devel
 BuildRequires:  libunistring-devel
--- a/ipatests/azure/Dockerfile.build-container
+++ b/ipatests/azure/Dockerfile.build-container
@@ -1,0 +1,17 @@
+FROM fedora:30
+MAINTAINER [FreeIPA Developers freeipa-devel@lists.fedorahosted.org]
+ENV container=docker LANG=en_US.utf8 LANGUAGE=en_US.utf8 LC_ALL=en_US.utf8
+
+ADD dist /root
+RUN echo 'deltarpm = false' >> /etc/dnf/dnf.conf \
+    && dnf update -y dnf \
+    && dnf install -y dnf-plugins-core sudo wget systemd firewalld nss-tools iptables \
+    && dnf config-manager '*modular*' --set-disabled \
+    && dnf install -y /root/rpms/*.rpm \
+    && dnf clean all && rm -rf /root/rpms /root/srpms \
+    && for i in /usr/lib/systemd/system/*-domainname.service; \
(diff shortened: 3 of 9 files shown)


abbra:
This pull request only adds configuration files but doesn't connect to Azure Pipelines yet. I created https://dev.azure.com/freeipa/freeipa/ and once the files are in git master, I'll connect Azure application from Github marketplace to it so that pull requests and master branch will be testable in Azure Pipelines.

You can see an example run at my personal fork https://github.com/abbra/freeipa/pull/11/ (need to be logged in to github to see results of tests on the pull request page). Actual test results in Azure DevOps are publicly visible: https://dev.azure.com/abbra1freeipa/freeipa/_build/results?buildId=161&view=logs.

The test runs already found few bugs in our tests and IPA code. I filed https://pagure.io/freeipa/issue/7940, https://pagure.io/freeipa/issue/7941, and https://pagure.io/freeipa/issue/7942 to track them. Once all these issues fixed, we can expect `ipa-run-tests` to actually run XMLRPC tests.

stanislavlevin:
Just for interest, what is the main reason to switch from Travis or add a new CI?

abbra:
Travis CI has quite a few limitations we are still trying to get through. The most important one, we cannot actually share build artifacts between test stages because that feature doesn't work properly for pull requests (only for proper branches) and it is not possible to store logs.

We also had a lot of stability issues with Travis CI through the last two years.

stanislavlevin:
it's promising. Thanks!

abbra:
Added `ip6_tables` module to the test machine job to work around an issue in the firewalld where it drops IPv4 rules if IPv6 tables cannot be updated as well. This is going to be fixed in firewalld upstream.

ipatests/azure/azure-run-tests.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash -x
```
tiran: The script does not set ``set -e`` to abort and fail the script when one of the commands fail.

rcritten:
It looks ok to me and seems low-risk. Worst case it blows up reviews and we disable it.

tiran:
> It looks ok to me and seems low-risk. Worst case it blows up reviews and we disable it.

It's possible to mark checks as non-required. In case Azure tests are unstable, it's possible to mark them as optional and don't have them block the pipeline.

ipatests/azure/azure-run-tests.sh
abbra: Fixed.

rcritten:
master:

* b7533d9c5fd66441699a764df3b225a3a62d6fab Use nodejs 1.10 to avoid current issues with nodejs 1.11 in Fedora 30
* 58fe6fac619200873c299b158c639d88f9b9b7ce Set up CI with Azure Pipelines

stanislavlevin:
@abbra,
I have different test results for the same PR ( Azure - PASS vs Travis - FAIL ):
https://dev.azure.com/freeipa/freeipa/_build/results?buildId=36
https://travis-ci.org/freeipa/freeipa/jobs/533680470

The difference is that Travis (unlike Azure) runs `pycodestyle` on PR:
https://github.com/freeipa/freeipa/blob/6662e99e173a16059271528e1fa58e37a3028924/.travis_run_task.sh#L34
```
if [[ "$TASK_TO_RUN" == "lint" ]]
then
    if [[ "$TRAVIS_EVENT_TYPE" == "pull_request" ]]
    then
        git diff origin/$TRAVIS_BRANCH -U0 | \
            pycodestyle --ignore=W504 --diff &> $PEP8_ERROR_LOG ||:
    fi
fi
```
Is it expected? It seems Azure has the same predefined variables:
https://docs.microsoft.com/en-us/azure/devops/pipelines/migrate/from-travis?view=azure-devops

abbra:
We don't run that script in Azure Pipelines. I created https://github.com/abbra/freeipa/pull/14 which should add a `make fastlint` run after `configure` step to catch this. `make fastlint` runs the same sequence (and a bit more).

Can you please pick up those two commits into your branch and retry? These also enable more tests to run in Base tests (and many more failed now due to that!)

stanislavlevin:
I added this here: https://github.com/freeipa/freeipa/pull/2716

abbra:
Looks like I need to tune some other bits in that pull request (ipa-run-tests is failing to start base tests)

abbra:
@stanislavlevin my final set of fixes for Azure pipelines is in the https://github.com/freeipa/freeipa/pull/3163:

```
pick 33f015133 ipa-run-tests: add support of globs for test targets and ignores
pick 2f10b5922 Azure Pipelines: simplify test job definitions
pick f9faf87c1 Azure Pipelines: run fast linter in case of a pull request build
```

stanislavlevin:
Unfortunately, the result is False negative.
https://dev.azure.com/freeipa/freeipa/_build/results?buildId=51

stanislavlevin:
Another problem is that I (only me?) can't download system logs (artifacts):
```json
{

    "$id": "1",
    "innerException": null,
    "message": "TF400813: The user 'Windows Live ID\\slev@altlinux.org' is not authorized to access this resource.",
    "typeName": "Microsoft.TeamFoundation.Framework.Server.UnauthorizedRequestException, Microsoft.TeamFoundation.Framework.Server",
    "typeKey": "UnauthorizedRequestException",
    "errorCode": 0,
    "eventId": 3000

}
```

abbra:
@stanislavlevin I sent you an invite.

stanislavlevin:
@abbra, thank you! I just hoped this is available for everyone.

abbra:
@stanislavlevin I hoped too. I thought it is the same situation as with Travis -- you have to login to Travis to see organizational logs. Apparently, it is a bit more complicated with Azure ACLs.

stanislavlevin:
The reason for the difference between pycodestyle results(Azure vs Travis) is in the manner of `pycodestyle` launching. 
```
The project options are read from the [pycodestyle] section of the
tox.ini file or the setup.cfg file located in any parent folder of the
path(s) being processed.
```

tox.ini contains pycodestyle section:
```
[pycodestyle]                                                                   
# E402 module level import not at top of file                                   
# W504 line break after binary operator                                         
ignore = E402, W504
```

Pycodestyle being run in Azure respects this and ignores properly.
But Travis runs it as:
```
pycodestyle --ignore=W504 --diff &> $PEP8_ERROR_LOG ||:
```

```
Please note that if the option –ignore=errors is used, the default configuration will be overridden and ignore only the check(s) you skip
```
So, all it needs to do:
``` diff
diff --git a/.travis_run_task.sh b/.travis_run_task.sh
index 85a79f2d3..02d639184 100755
--- a/.travis_run_task.sh
+++ b/.travis_run_task.sh
@@ -34,7 +34,7 @@ then
     if [[ "$TRAVIS_EVENT_TYPE" == "pull_request" ]]
     then
         git diff origin/$TRAVIS_BRANCH -U0 | \
-            pycodestyle --ignore=W504 --diff &> $PEP8_ERROR_LOG ||:
+            pycodestyle --diff &> $PEP8_ERROR_LOG ||:
     fi
 fi
```
In this case `pycodestyle` will always use config settings.

stanislavlevin:
@abbra, should I open another PR to eliminate the difference between `fastlint` and TRAVIS `lint`? (please, see the comment above)

abbra:
@stanislavlevin yes, please open a separate pull request to correct the travis script.

PR #3138, opened 2019-05-13
Fix occasional 'whoami.data is undefined' error in FreeIPA web UI

serg-cymbaluk:
'Metadata' phase (Web UI initialization flow) doesn't wait "whoami" response.
It causes the error when on the next phase "whoami" data is undefined.
To avoid this "whoami" request now has flag async = false,
so init_metadata waits until it will be completed.

Ticket: https://pagure.io/freeipa/issue/7917

### DIFF
--- a/install/ui/src/freeipa/ipa.js
+++ b/install/ui/src/freeipa/ipa.js
@@ -251,7 +251,10 @@ var IPA = function () {
             on_success: function(data, text_status, xhr) {
                 that.whoami.metadata = data.result || data;
                 var wa_data = that.whoami.metadata;
-
+                // This AJAX request has no synchronization point,
+                // so we set async = false to make sure that init_metadata
+                // doesn't start before we get whoami response.
+                $.ajaxSetup({async: false});
                 rpc.command({
                     method: wa_data.details || wa_data.command,
                     args: wa_data.arguments,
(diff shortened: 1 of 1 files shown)


install/ui/src/freeipa/ipa.js
```diff
@@ -251,7 +251,10 @@ var IPA = function () {
             on_success: function(data, text_status, xhr) {
                 that.whoami.metadata = data.result || data;
                 var wa_data = that.whoami.metadata;
-
+                // This AJAX request has no synchronization point,
+                // so we set async = false to make sure that init_metadata
+                // doesn't start before we get whoami response.
+                $.ajaxSetup({async: false});
```
rcritten: You don't need to set content or datatype and others here as well, ala install/ui/src/freeipa/config.js:91

serg-cymbaluk: @rcritten 
There is a global ajaxSetup (freeipa/ipa.js:131), so this local setup overrides only one option and keeps other values which were specified previously.

rcritten: Ok, I wasn't sure if it replaced or augmented the settings. LGTM.

serg-cymbaluk:
master:

* 6a9c20a87c2ae3eaa941120073cad34c328cb7df Fix occasional 'whoami.data is undefined' error in FreeIPA web UI

PR #3139, opened 2019-05-13
Load libldap_r-*.so.2

tiran:
libldap_r.so is only available in the OpenLDAP development packages. The
openldap package provides libldap_r-*.so.2.

Fixes: https://pagure.io/freeipa/issue/7941

### DIFF
--- a/ipapython/dn_ctypes.py
+++ b/ipapython/dn_ctypes.py
@@ -12,8 +12,8 @@ import six
 
 __all__ = ("str2dn", "dn2str", "DECODING_ERROR", "LDAPError")
 
-# load reentrant libldap
-ldap_r_lib = ctypes.util.find_library("ldap_r")
+# load reentrant ldap client library (libldap_r-*.so.2)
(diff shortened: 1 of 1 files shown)


ipapython/dn_ctypes.py
```diff
@@ -12,8 +12,8 @@
 
 __all__ = ("str2dn", "dn2str", "DECODING_ERROR", "LDAPError")
 
-# load reentrant libldap
-ldap_r_lib = ctypes.util.find_library("ldap_r")
+# load reentrant ldap clinet library (libldap_r-*.so.2)
```
abbra: clinet -> client

abbra:
Other than typo in a comment, looks good.

abbra:
Since 4.7 is in Fedora 29, we need to backport this patch to 4.7 too.

tiran:
There is no need for a backup. ``ipapython.dn_ctypes`` is not in 4.7

tiran:
master:

* 64dc92ccb4f057ec49070b323f3b827d8642cece Load libldap_r-*.so.2

PR #3140, opened 2019-05-13
Use PKCS#8 instead of traditional privkey format

tiran:
The modern PKCS#8 private key format supports better encryption standard
and is preferable over traditional, weak PKCS#1 key format.

Fixes: https://pagure.io/freeipa/issue/7943


tiran:
master:

* 2042b5a0d2236d27ce36a139b4e9e89a055f28b5 Use PKCS#8 instead of traditional privkey format

PR #3141, opened 2019-05-13
Fix `build_requestinfo` in LibreSSL environments

stanislavlevin:
`build_requestinfo` was broken in ac6568dcf.
In this case LibreSSL behavior is the same as OpenSSL < 1.1.x.
Thus, an additional check for SSL implementation was added.

Fixes: https://pagure.io/freeipa/issue/7937

### DIFF
--- a/ipaclient/csrgen_ffi.py
+++ b/ipaclient/csrgen_ffi.py
@@ -6,13 +6,32 @@ from ipalib import errors
 _ffi = FFI()
 
 _ffi.cdef('''
+/* libcrypto/crypto.h */
 unsigned long OpenSSL_version_num(void);
 unsigned long SSLeay(void);
+const char * OpenSSL_version(int t);
+const char * SSLeay_version(int t);
+
+#define OPENSSL_VERSION 0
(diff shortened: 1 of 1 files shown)


stanislavlevin:
One of my tasks is the setup of the IPA CI for my distro.
I'm going to start from tests (`make fasttest`) during RPM build ( every day running ).

The very first fail I faced is this issue :-)
Actually, `test_ipaclient/test_csrgen.py::test_CSRGenerator::test_works_with_lowercase_attr_type_shortname` catches excellent (crashes).

To sum up, I checked this PR locally with:
- OpenSSL-1.1.1b
- OpenSSL-1.0.2r
- LibreSSL-2.9.1
- LibreSSL-2.6.4

tiran:
I found the reason why our tests never crashed, see https://pagure.io/freeipa/issue/7947

tiran:
master:

* 7b8a2af2197381058ca532d1ae206defb16fac88 Fix `build_requestinfo` in LibreSSL environments

PR #3154, opened 2019-05-16
[Backport][ipa-4-7] Limit hostname length to 64 to be in line with MAXHOSTNAMELEN

rcritten:
Manual backport of PR #3101 to ipa-4-7 branch. Fixes one minor merge issue in ipaserver/install/installutils.py.

On Linux systems the length limit for hostnames is hardcoded
at 64 in MAXHOSTNAMELEN

Solaris, for example, allows 255 characters, and DNS allows the
total length to be up to 255 (with each label < 64).

Add a knob to allow configuring the maximum hostname length (FQDN)

The same validators are used between hosts and DNS to apply
the knob only when dealing with a FQDN as a hostname.

The maxlen option is included so installers can limit the length
of allowed hostnames when the --hostname option is used.

https://pagure.io/freeipa/issue/2018


flo-renaud:
Hi @rcritten 
thanks for the backport, works for me.
- [x] Tested new install and upgrade, new def for attribute/objectclass are present and ACIs properly set.
- [x] `ipa config-mod` allows only values >=64, can be executed by any member of the admins group.
- [x] `ipa config-show` can be executed by anyone, properly displays the value
- [x] `ipa host-add` enforces the configured value
- [x] ipa-server-install, ipa-replica-install and ipa-client-install enforce the 64 max hostname len.

flo-renaud:
ipa-4-7:

* 319b920e19bafab9e2c1555969c11bbf133707e6 Add knob to limit hostname length

PR #3158, opened 2019-05-17
admintool: don't display log file on errors unless logging is setup

rcritten:
The admintool will display the message when something goes wrong:

See %s for more information" % self.log_file_name

This is handy except when finally logging setup is not done
yet so the log file doesn't actually get written to.

This can happen if validation catches and raises an exception.

The workaround is to save off and remove the log_file_name value
before calling validation, then restore it if successful. This will
suppress the above error message.

Fixes: https://pagure.io/freeipa/issue/7952

### DIFF
--- a/ipapython/admintool.py
+++ b/ipapython/admintool.py
@@ -161,6 +161,7 @@ class AdminTool:
     def __init__(self, options, args):
         self.options = options
         self.args = args
+        self.log_file_initialized = False
         self.safe_options = self.option_parser.get_safe_opts(options)
 
     def execute(self):
@@ -247,12 +248,15 @@ class AdminTool:
                 break
 
         self._setup_logging(log_file_mode=log_file_mode)
+        if self.log_file_name:
+            self.log_file_initialized = True
 
     def _setup_logging(self, log_file_mode='w', no_file=False):
         if no_file:
             log_file_name = None
         elif self.options.log_file:
             log_file_name = self.options.log_file
+            self.log_file_name = log_file_name
         else:
             log_file_name = self.log_file_name
         if self.options.verbose:
@@ -313,7 +317,7 @@ class AdminTool:
             # ipa-server-install.
             return
         message = "The %s command failed." % self.command_name
-        if self.log_file_name and return_value != 2:
+        if self.log_file_initialized and return_value != 2:
             # magic value because this is common between server and client
             # but imports are not straigthforward
             message += " See %s for more information" % self.log_file_name


flo-renaud:
Hi @rcritten 
thanks for the patch.
There is still an issue if a command is called with `--log-file /dir/does/not/exist/file` because the output will display:
```
# ipa-backup   --log-file /tmp2/foo
[Errno 2] No such file or directory: '/tmp2/foo'
The ipa-backup command failed. See /var/log/ipabackup.log for more information
```
I would rather add an attribute `self.log_file_initialized = False` in `AdminTool.__init` and set it to True/False in `AdminTool._setup_logging` depending on the value of log_file_name. Then in `AdminTool.log_failure`, replace the check `if self.log_file_name and return_value != 2` with `if self.log_file_initialized and return_value != 2`.

rcritten:
Right, great idea, implemented.

ipapython/admintool.py
```diff
@@ -247,6 +248,8 @@ def setup_logging(self, log_file_mode='w'):
                 root_logger.removeHandler(handler)
                 break
 
+        if self.log_file_name:
+            self.log_file_initialized = True
```
flo-renaud: The attribute `self.log_file_initialized` should be set to True only after `standard_logging_setup` has been called successfully, otherwise we'll still get an error message referring to a non-existent file such as
```
# ipa-backup   --log-file /tmp2/foo
[Errno 2] No such file or directory: '/tmp2/foo'
The ipa-backup command failed. See /var/log/ipabackup.log for more information
```

rcritten: I tested that specifically and it worked fine for me but I'll take another look.

rcritten: Move the init past the __setup_logging.

ipapython/admintool.py
```diff
@@ -309,7 +312,7 @@ def log_failure(self, error_message, return_value, exception, backtrace):
         if error_message:
             logger.error('%s', error_message)
         message = "The %s command failed." % self.command_name
-        if self.log_file_name and return_value != 2:
+        if self.log_file_initialized and return_value != 2:
             # magic value because this is common between server and client
             # but imports are not straigthforward
             message += " See %s for more information" % self.log_file_name
```
flo-renaud: I tested CTRL-C during the backup launched with `ipa-backup --log-file /tmp/foo`, and the error message always refers to `See /var/log/ipabackup.log for more information` even if a different log file was used. The var self.log_file_name is never assigned self.options.log_file, this should probably be done in __setup_logging.

rcritten: Good catch, this seems to be a long-standing bug on its own!

ipapython/admintool.py
```diff
@@ -256,6 +259,7 @@ def _setup_logging(self, log_file_mode='w', no_file=False):
             log_file_name = self.options.log_file
         else:
             log_file_name = self.log_file_name
+        self.log_file_name = log_file_name
```
flo-renaud: This line should be moved in the `elif self.options.log_file:` block, otherwise the first call to _setup_logging(no_file=True) will reset self.log_file_name to None instead of keeping the default value /var/log/ipabackup.log

rcritten: You're right, nice catch.

flo-renaud:
Hi @rcritten 
thanks for the updated PR, works perfectly. Tested with the following:
- [x] if ipa-backup fails during the validate checks, no log file name is displayed, we just get the message `The ipa-backup command failed.`
- [x] if ipa-backup fails during the execution (for instance with CTRL-C), the correct file name is displayed (the one from --log-file or the default /var/log/ipabackup.log): `The ipa-backup command failed. See /tmp/foo for more information` and the log is written in the correct file
- [x] if ipa-backup succeeds, the log is written in the file specified by --log-file or the default /var/log/ipabackup.log

flo-renaud:
master:

* 10b721d118a295b62a4c1df52ae47d8b106464a2 admintool: don't display log file on errors unless logging is setup

PR #3161, opened 2019-05-20
WebUI: Add possibility to run caless UI tests separately

serg-cymbaluk:
Add pytest mark "caless" to all related UI test suites.

Ticket: https://pagure.io/freeipa/issue/7954

### DIFF
--- a/ipatests/test_webui/test_host.py
+++ b/ipatests/test_webui/test_host.py
@@ -312,6 +312,7 @@ class test_host(host_tasks):
         self.delete_record(self.pkey, self.data.get('del'))
 
     @screenshot
+    @pytest.mark.caless
     def test_ca_less(self):
         """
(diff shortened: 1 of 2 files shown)


f-trivino:
This is green but looks like PRCI is not running here... we need PRCI working before merging this PR.

rcritten:
Does this need to be included in ipatests/conftest.py or not because it is WebUI-specific?

serg-cymbaluk:
@rcritten 
It is just for WebUI tests. I don't expect that it will be used for more general purpose.

rcritten:
I guess I don't see where these are hooked up to be executed as part of any of the tests. The code looks ok but I'm missing the bigger picture.

serg-cymbaluk:
These changes are not required for now, because caless tests can be easily run using `-k ca_less` option with pytest.

PR #3163, opened 2019-05-20
Getkeytab fixes

abbra:
This is a set of initial patches needed to support a Samba server running on IPA domain member.

These patches by themselves do not enable Samba server yet. They fix assorted bugs I found when adding the support for SMB service and tests for them. Some patches were omitted due to the complexity to untangle.

Most of `test_xmlrpc` tests that rely on a comparison of the user/group results cannot tolerate running in an environment where trust to Active Directory support is enabled. Originally, I started adding fixes for these problems but it quickly escalated to several hundred failed tests. The main issue is a dictionary comparison feature that doesn't allow to specify an attribute to be optional. When trust to AD is enabled, all POSIX users and groups in IPA will get `ipaNTSecurityIdentifier` attribute and `ipantuserattrs`/`ipantgroupattrs` object class added at a creation time.

Dictionary comparison in tests will fail:
- a tracker-maintained dictionary would not have `ipantuserattrs`/`ipantgroupattrs` objectclasses
- a result returned from `user-add` / `group-add` commands might contain the objectclasses and `ipaNTSecurityIdentifier` or might not, depending on how powerful is the test machine so that `sidgen` plugin is able to add the attributes before `user-add` / `group-add` command would retrieve the entry
- if we would add `ipaNTSecurityIdentifier` into the tracker-maintained dictionary, the result comparison code will have no logic to understand that if `ipaNTSecurityIdentifier` is missing in the result, it is not a failure: `fuzzy_sid` class allows for `None` value.
- for group-finding tests, it is impossible to get just a subset of required groups. This prevents from any additional configuration on the test machines. An attempt to search by a wildcard with `ipa group-find --description="Test desc*"` fails because we escape "*" and that filter doesn't match. We need to refactor search code to allow a wildcard search to reduce the number of expected groups in such tests.

I started fixing these issues but they will evidently require a major refactoring of the test code. Meanwhile, these problems are tracked with https://pagure.io/freeipa/issue/7955

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-pwd-extop/ipa_pwd_extop.c
+++ b/daemons/ipa-slapi-plugins/ipa-pwd-extop/ipa_pwd_extop.c
@@ -97,7 +97,8 @@ void *ipapwd_get_plugin_id(void)
 }
 
 static void filter_keys(struct ipapwd_krbcfg *krbcfg,
-                        struct ipapwd_keyset *kset)
+                        struct ipapwd_keyset *kset,
+                        bool allow_nthash)
 {
     int i, j;
 
@@ -108,6 +109,14 @@ static void filter_keys(struct ipapwd_krbcfg *krbcfg,
                 break;
             }
         }
+
+        /* if requested by the caller, allow arcfour-hmac even
+         * if it is missing in the list of supported enctypes. */
+        if (allow_nthash &&
+            (ENCTYPE_ARCFOUR_HMAC == kset->keys[i].key_data_type[0])) {
+            break;
+        }
+
         if (j == krbcfg->num_supp_encsalts) { /* not valid */
 
             /* free key */
@@ -130,7 +139,8 @@ static void filter_keys(struct ipapwd_krbcfg *krbcfg,
 
 static void filter_enctypes(struct ipapwd_krbcfg *krbcfg,
                             krb5_key_salt_tuple *kenctypes,
-                            int *num_kenctypes)
+                            int *num_kenctypes,
+                            bool allow_nthash)
 {
     /* first filter for duplicates */
     for (int i = 0; i + 1 < *num_kenctypes; i++) {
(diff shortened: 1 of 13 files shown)


tiran:
Thanks! The patch set looks ok to me. I'll do a proper review tomorrow.

abbra:
rebased against master to be able to kick PR CI tests.

abbra:
Looks like some DNS-related tests are failing due to timing. I've added `/var/named/data` to the list of log locations to pull from the test system in Azure Pipelines so that we can investigate what happened at a DNS server side.

abbra:
@tiran all tests passed.

abbra:
master:

* 6a2c356da0c603f209f2310c013ef085e0c11eda ipa-run-tests: add support of globs for test targets and ignores
* c8ef093e56e8eaa9c53abb31bed51ff30c9c64ef Azure Pipelines: simplify test job definitions
* 5230e2a12ddbed0e27b107d7174d08e038f3bc1f Azure Pipelines: run fast linter in case of a pull request build
* 53a0fa9130493d383d2542622ccdbdd483650bad LDAPCreate: allow callers to override objectclasses
* ef67dece522335fd26fcdddc822beb4bcc8e6b17 ldap2.can_read: fix py3 compatibility
* 6163cbc16658930f49794ebecd5a6ac14ba8cfd4 test_ipagetkeytab: allow testing LDAP connection beyond bind operation
* 0f891c6a3fef7b75cb4b0a125e26e6e80b0bdcf3 test_ipagetkeytab: factor out DM password reader
* b5fbbd1957873207afc2542a9d32246bccca0556 Keytab retrieval: allow requesting arcfour-hmac for SMB services
* 46234f0cb91ad892b7420eb061e758e26c64e3c7 test_ipagetkeytab: test retrieval of explicit encryption types
* a9bcf531a69b8a39ee2df86b4eb783023b33928e ipa-pwd-extop: do not remove MagicRegen mod, replace it

PR #3164, opened 2019-05-20
Respect TMPDIR, TEMP or TMP environment variables during testing

stanislavlevin:
The FreeIPA uses its own classes for managing temp files and
directories for tests. One of its underlying low-level functions
is `mkdtemp`.

According to documentation for `mkdtemp`:
```
If dir is not None, the file will be created in that directory; otherwise, a
default directory is used. The default directory is chosen from a
platform-dependent list, but the user of the application can control the
directory location by setting the TMPDIR, TEMP or TMP environment variables.
```

It's actually the truth,
/usr/lib64/python3.7/tempfile.py:
```
def _candidate_tempdir_list():
    """Generate a list of candidate temporary directories which
    _get_default_tempdir will try."""

    dirlist = []

    # First, try the environment.
    for envname in 'TMPDIR', 'TEMP', 'TMP':
        dirname = _os.getenv(envname)
        if dirname: dirlist.append(dirname)

    # Failing that, try OS-specific locations.
    if _os.name == 'nt':
        dirlist.extend([ _os.path.expanduser(r'~\AppData\Local\Temp'),
                         _os.path.expandvars(r'%SYSTEMROOT%\Temp'),
                         r'c:\temp', r'c:\tmp', r'\temp', r'\tmp' ])
    else:
        dirlist.extend([ '/tmp', '/var/tmp', '/usr/tmp' ])
```

For now, there is a hardcoded assertion of a temp directory (`/tmp`) in
IPA tests. But some systems use the mentioned environment variables
(for example, pam_mktemp https://www.openhub.net/p/pam_mktemp).
It's easy to check an actual temp dir via `gettempdir`.

Fixes: https://pagure.io/freeipa/issue/7956

### DIFF
--- a/ipatests/util.py
+++ b/ipatests/util.py
@@ -106,7 +106,8 @@ class TempDir:
 
     def __get_path(self):
         assert path.abspath(self.__path) == self.__path
-        assert self.__path.startswith('/tmp/ipa.tests.')
+        assert self.__path.startswith(path.join(tempfile.gettempdir(),
+                                                'ipa.tests.'))
         assert path.isdir(self.__path) and not path.islink(self.__path)
         return self.__path
     path = property(__get_path)


rcritten:
master:

* 5263c36c1b3a0e1cbc8f3f368c81706a46a75687 Respect TMPDIR, TEMP or TMP environment variables during testing

PR #3166, opened 2019-05-21
ipatest : Test if ipactl restart restarts the pki-tomcatd properly

mrizwan93:
Wrong logic was triggering the start instead of restart
for pki-tomcatd. This test validates that restart
called on pki-tomcat properly.

related ticket : https://pagure.io/freeipa/issue/7927

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -380,6 +380,16 @@ class TestADTrustInstallWithDNS_KRA_ADTrust(ADTrustInstallTestBase):
         self.install_replica(self.replicas[1], setup_ca=True, setup_kra=True)
 
 
+def get_pki_tomcatd_pid(host):
+    pid = ''
+    cmd = host.run_command(['systemctl', 'status', 'pki-tomcatd@pki-tomcat'])
+    for line in cmd.stdout_text.split('\n'):
+        if "Main PID" in line:
+            pid = line.split()[2]
+            break
+    return(pid)
+
+
 ##
 # Rest of master installation tests
 ##
@@ -403,6 +413,34 @@ class TestInstallMaster(IntegrationTest):
             self.master,
             extra_args=['--dnssec-master', '--no-dnssec-validation']
         )
+
+    def test_ipactl_restart_pki_tomcat(self):
+        """ Test if ipactl restart restarts the pki-tomcatd
+
+        Wrong logic was triggering the start instead of restart
+        for pki-tomcatd. This test validates that restart
+        called on pki-tomcat properly.
+
+        related ticket : https://pagure.io/freeipa/issue/7927
+        """
+        # get process id of pki-tomcatd
+        pki_pid = get_pki_tomcatd_pid(self.master)
+
+        # check if pki-tomcad restarted
+        cmd = self.master.run_command(['ipactl', 'restart'])
+        assert "Restarting pki-tomcatd Service" in cmd.stdout_text
+
+        # check if pid for pki-tomcad changed
+        pki_pid_after_restart = get_pki_tomcatd_pid(self.master)
+        assert pki_pid != pki_pid_after_restart
+
+        # check if pki-tomcad restarted
+        cmd = self.master.run_command(['ipactl', 'restart'])
+        assert "Restarting pki-tomcatd Service" in cmd.stdout_text
+
+        # check if pid for pki-tomcad changed
+        pki_pid_after_restart_2 = get_pki_tomcatd_pid(self.master)
+        assert pki_pid_after_restart != pki_pid_after_restart_2
 
     def test_WSGI_worker_process(self):
         """ Test if WSGI worker process count is set to 4


ipatests/test_integration/test_installation.py
```diff
@@ -404,6 +404,46 @@ def test_install_dns(self):
...
+
+        related ticket : https://pagure.io/freeipa/issue/7927
+        """
+        splitted = []
+        # get process id of pki-tomcatd
+        cmd = self.master.run_command(['ps', '-ef'])
+        for line in cmd.stdout_text.split('\n'):
+            if "pki-tomcatd" in line:
+                splitted = line.split()
+        pki_pid = splitted[1]
```
rcritten: Why not use systemctl status pki-tomcatd@pki-tomcat and look for "Main PID:"?

pidof java would work too but could be fragile I suppose if we ended up with another java program (perhaps not likely)

ipatests/test_integration/test_installation.py
```diff
@@ -404,6 +404,46 @@ def test_install_dns(self):
...
+        # check if pki-tomcad restarted
+        cmd = self.master.run_command(['ipactl', 'restart'])
+        assert "Restarting pki-tomcatd Service" in cmd.stdout_text
+
+        # check if pid for pki-tomcad changed
+        cmd = self.master.run_command(['ps', '-ef'])
+        for line in cmd.stdout_text.split('\n'):
+            if "pki-tomcatd" in line:
+                splitted = line.split()
+        assert pki_pid not in splitted
```
rcritten: Rather than duplicating this same block three times it would be better in a separate function.

ipatests/test_integration/test_installation.py
mrizwan93: I tried with systemctl but was getting:
`SyntaxError: Non-ASCII character '\xe2' in file /tmp/temp.py on line 1, but no encoding declared; see http://www.python.org/peps/pep-0263.html for details`

`systemctl status` output have some ASCII characters, and need to use an encoding. I am not sure if using encoding will break other things. So I continued with `ps -ef`

ipatests/test_integration/test_installation.py
```diff
@@ -404,6 +414,38 @@ def test_install_dns(self):
...
+        # check if pki-tomcad restarted
+        cmd = self.master.run_command(['ipactl', 'restart'])
+        assert "Restarting pki-tomcatd Service" in cmd.stdout_text
+
+        # check if pid for pki-tomcad changed
+        cmd = self.master.run_command(['ps', '-ef'])
+        if pki_pid:
+            assert pki_pid not in cmd.stdout_text
+
+        pki_pid = get_pki_tomcatd_pid(self.master)
```
flo-renaud: You can simplify the code and replace
```
        # check if pid for pki-tomcad changed
        cmd = self.master.run_command(['ps', '-ef'])
        if pki_pid:
            assert pki_pid not in cmd.stdout_text

        pki_pid = get_pki_tomcatd_pid(self.master)
```
with
```
        pki_pid_after_restart = get_pki_tomcatd_pid(self.master)
        assert pki_pid != pki_pid_after_restart
```

ipatests/test_integration/test_installation.py
```diff
@@ -404,6 +414,38 @@ def test_install_dns(self):
...
+
+        # check if pki-tomcad restarted
+        cmd = self.master.run_command(['ipactl', 'restart'])
+        assert "Restarting pki-tomcatd Service" in cmd.stdout_text
+
+        # check if pid for pki-tomcad changed
+        cmd = self.master.run_command(['ps', '-ef'])
+        if pki_pid:
+            assert pki_pid not in cmd.stdout_text
+
```
flo-renaud: You can also simplify the code here with
```
       pki_pid_after_restart_2 = get_pki_tomcatd_pid(self.master)
       assert pki_pid_after_restart != pki_pid_after_restart_2
```

ipatests/test_integration/test_installation.py
```diff
@@ -380,6 +380,17 @@ def test_replica1_all_components_adtrust(self):
         self.install_replica(self.replicas[1], setup_ca=True, setup_kra=True)
 
 
+def get_pki_tomcatd_pid(host):
+    pid = ''
+    cmd = host.run_command(['ps', '-ef'])
```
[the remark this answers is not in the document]
rcritten: pidof java is probably the easiest and fastest way to get it. It seems a bit odd but given dogtag is a java app, and the ONLY java app, it works.

flo-renaud:
Hi @mrizwan93 
thanks for the PR, LGTM. Can you remove the temp commit? I will provide the ACK as soon as the test suite is green.

flo-renaud:
master:

* 581b7148f4560cd0595b2c771b6eef2fac5dc93b Test if ipactl restart restarts the pki-tomcatd

flo-renaud:
Hi @mrizwan93 
A manual backport is required for ipa-4-6 branch, can you handle this? Thanks.

PR #3167, opened 2019-05-21
ipatests: new tests for establishing one-way AD trust with shared secret

wladich:
#2926 Introduced support for new type of trust, adding tests for it

Tests added for two scenarios:
1) adding one-way external trust, trust on Windows side is created using
   netdom utility.
2) adding one-way forest trust, trust on Windows side is created using
   powershell bindings to .Net functions
Tests verify that specified trusts can be established, trust domains can
be  fetched and AD user data can be queried by IPA client.

Relates: https://pagure.io/freeipa/issue/6077

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -644,7 +644,17 @@ def unconfigure_dns_for_trust(master, ad):
         restore_dnssec_validation(master)
 
 
-def establish_trust_with_ad(master, ad_domain, extra_args=()):
+def configure_windows_dns_for_trust(ad, master):
+    ad.run_command(['dnscmd', '/zoneadd', master.domain.name,
+                    '/Forwarder', master.ip])
+
+
+def unconfigure_windows_dns_for_trust(ad, master):
+    ad.run_command(['dnscmd', '/zonedelete', master.domain.name, '/f'])
+
+
+def establish_trust_with_ad(master, ad_domain, extra_args=(),
+                            shared_secret=None):
     """
     Establishes trust with Active Directory. Trust type is detected depending
     on the presence of SfU (Services for Unix) support on the AD.
@@ -654,6 +664,7 @@ def establish_trust_with_ad(master, ad_domain, extra_args=()):
     """
 
     # Force KDC to reload MS-PAC info by trying to get TGT for HTTP
+    extra_args = list(extra_args)
     master.run_command(['kinit', '-kt', paths.HTTP_KEYTAB,
                         'HTTP/%s' % master.hostname])
     master.run_command(['systemctl', 'restart', 'krb5kdc.service'])
@@ -663,12 +674,15 @@ def establish_trust_with_ad(master, ad_domain, extra_args=()):
     master.run_command(['klist'])
     master.run_command(['smbcontrol', 'all', 'debug', '100'])
 
-    run_repeatedly(master,
-                   ['ipa', 'trust-add',
-                    '--type', 'ad', ad_domain,
-                    '--admin', 'Administrator',
-                    '--password'] + list(extra_args),
-                   stdin_text=master.config.ad_admin_password)
+    if shared_secret:
+        extra_args += ['--trust-secret']
+        stdin_text = shared_secret
+    else:
+        extra_args += ['--admin', 'Administrator', '--password']
+        stdin_text = master.config.ad_admin_password
+    run_repeatedly(
+        master, ['ipa', 'trust-add', '--type', 'ad', ad_domain] + extra_args,
+        stdin_text=stdin_text)
     master.run_command(['smbcontrol', 'all', 'debug', '1'])
     clear_sssd_cache(master)
     master.run_command(['systemctl', 'restart', 'krb5kdc.service'])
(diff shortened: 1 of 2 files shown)


ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -663,12 +674,15 @@ def establish_trust_with_ad(master, ad_domain, extra_args=()):
...
-                   ['ipa', 'trust-add',
-                    '--type', 'ad', ad_domain,
-                    '--admin', 'Administrator',
-                    '--password'] + list(extra_args),
-                   stdin_text=master.config.ad_admin_password)
+    if shared_secret:
+        extra_args += ['--trust-secret']
+        stdin_text = shared_secret
+    else:
+        extra_args += ['--admin', 'Administrator', '--password']
```
kaleemsiddiqu: should not we leave this as default option like earlier in run_repeatedly as this is not extra_args, its mandatory part during trust-add.
we can just add extra_args for --trust-secret and stdin_text in if section and else section can have just stdin_text for password.

ipatests/test_integration/test_trust.py
```diff
@@ -344,3 +355,121 @@ def test_trustdomains_found_in_external_rootdomain_trust(self):
...
+            '::getCurrentForest()'
+            '.CreateLocalSideOfTrustRelationship("{}", 1, "{}")'.format(
+                self.master.domain.name, self.shared_secret))
+        self.ad.run_command(['powershell', '-c', ps_cmd])
+
+        # create ipa side of trust
+        tasks.establish_trust_with_ad(
+            self.master, self.ad_domain, shared_secret=self.shared_secret)
+
+    def test_trustdomains_found_in_forest_trust_with_shared_secret(self):
```
kaleemsiddiqu: Should not we make this part of test case test_establish_forest_trust_with_shared_secret as validation task after trust-add?

ipatests/test_integration/test_trust.py
```diff
@@ -344,3 +355,121 @@ def test_trustdomains_found_in_external_rootdomain_trust(self):
...
+        result = self.master.run_command(
+            ['ipa', 'trust-fetch-domains', self.ad.domain.name],
+            raiseonerr=False)
+        assert result.returncode == 1
+        self.check_trustdomains(
+            self.ad_domain, [self.ad_domain, self.ad_subdomain])
+
+    def test_user_gid_uid_resolution_in_forest_trust_with_shared_secret(self):
```
kaleemsiddiqu: Should not we make this part of test case test_establish_forest_trust_with_shared_secret as validation task after trust-add also?

ipatests/test_integration/test_trust.py
```diff
@@ -344,3 +355,121 @@ def test_trustdomains_found_in_external_rootdomain_trust(self):
...
+
+        # This regex checks that Test User does not have UID 10042 nor belongs
+        # to the group with GID 10047
+        testuser_regex = r"^testuser@%s:\*:(?!10042)(\d+):(?!10047)(\d+):"\
+                         r"Test User:/home/%s/testuser:/bin/sh$"\
+                         % (re.escape(self.ad_domain),
+                            re.escape(self.ad_domain))
+
+        assert re.search(
+            testuser_regex, result.stdout_text), result.stdout_text
```
kaleemsiddiqu: i think 
assert re.search( testuser_regex, result.stdout_text) is sufficient and another result.stdout_text not needed..

ipatests/test_integration/test_trust.py
wladich: No, I think we should separately check that trust is established and that it is operating, like it is done for other types of trust. Best practice for writing tests is to make one assertion in each test case.

ipatests/test_integration/test_trust.py
wladich: No, I think we should separately check that trust is established and that it is operating, like it is done for other types of trust. Best practice for writing tests is to make one assertion in each test case.

ipatests/test_integration/test_trust.py
wladich: result.stdout_text after comma will be displayed when asserttion fails. This makes debugging easier, as you see output in test logs. Otherwise we would need to manually reproduce test scenario to find out what went wrong

ipatests/test_integration/test_trust.py
kaleemsiddiqu: What if check is not established ? This will get executed and fail.
IMO one def test_* should cover the complete test case possibly including setup/cleanup.

ipatests/pytest_ipa/integration/tasks.py
wladich: shared secret should not be used together with --admin option

ipatests/test_integration/test_trust.py
wladich: We are doing sanity check of new way for  trust establishments. Test cases here are:
1) trust establishment
2) checking that domains can be retreived
3) checking that users can be retreived

Here trust establishment should not be treated as setup but as a testcase

ipatests/test_integration/test_trust.py
wladich: see above

abbra:
master:

* 3f02fc945e6306e8fb656db680d5a553683fabc4 ipatests: new tests for establishing one-way AD trust with shared secret

abbra:
@wladich please provide manual backports:

```
Aplying to ipa-4-7: ipatests: new tests for establishing one-way AD trust with shared secret

Aplying to ipa-4-6: ipatests: new tests for establishing one-way AD trust with shared secret
```

PR #3178, opened 2019-05-23
For Fedora and RHEL use system-wide crypto policy for mod_ssl

rcritten:
Drop the SSLProtocol directive for Fedora and RHEL systems. mod_ssl
will use crypto policies for the set of protocols.

For Debian systems configure a similar set of protocols for what
was previously configured, but do it in a different way. Rather than
iterating the allowed protocols just include the ones not allowed.

Fixes: https://pagure.io/freeipa/issue/7667

### DIFF
--- a/ipaplatform/base/tasks.py
+++ b/ipaplatform/base/tasks.py
@@ -246,6 +246,10 @@ class BaseTaskNamespace:
         """Configure WSGI for correct Python version"""
         raise NotImplementedError()
 
+    def configure_httpd_protocol(self):
+        """Configure TLS protocols in Apache"""
+        raise NotImplementedError()
+
     def is_fips_enabled(self):
         return False
 
--- a/ipaplatform/debian/tasks.py
+++ b/ipaplatform/debian/tasks.py
@@ -10,7 +10,9 @@ from __future__ import absolute_import
 
 from ipaplatform.base.tasks import BaseTaskNamespace
 from ipaplatform.redhat.tasks import RedHatTaskNamespace
+from ipaplatform.paths import paths
 
+from ipapython import directivesetter
 from ipapython import ipautil
 
 class DebianTaskNamespace(RedHatTaskNamespace):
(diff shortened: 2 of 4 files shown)


rcritten:
I haven't set  labels yet. This will work with the 4.7 branch, I don't know if we want to backport this or not. I'd be ok with 4.8+.

rcritten:
Setting WIP since this is breaking all over because connections can't be made now that TLS 1.3 is enabled in Apache (because it is allowed by crypto policy).

KRA installation also fails, https://bugzilla.redhat.com/show_bug.cgi?id=1713451

rcritten:
The problem, I think, is outlined in https://bugzilla.redhat.com/show_bug.cgi?id=1651623

TLSv1.3 uses a Post-Handshake Authentication to do client auth and it is apparently not working yet.

I have two things to investigate:
1. What crypto library is being used to communicate to dogtag. It used to be NSS but I need to confirm. It makes sense since the above BZ says that Firefox doesn't support it
2. Try to find a workaround. If I can't then this is going to have to wait and we'll simply hardcode TLSv1.2 as the protocol.

tiran:
I know how to enable PHA for Python requests (used by pki client) and our HTTPS code. I'll come up with a patch tomorrow.

rcritten:
cipherboy was telling me that he thinks the JDK may not (fully) support TLS 1.3 so that could be part of the issue as well.

tiran:
OpenSSL has PHA disabled by default, but HTTPS over TLS 1.3 requires PHA in certain cases. It's not required if the server always requires client auth. However in case only some paths require TLS client auth, the client must send PHA TLS extension during the handshake.

rcritten:
@tiran is working on a related PR to enable PHA, https://github.com/freeipa/freeipa/pull/3236

flo-renaud:
PR #3350 includes this code change and has been pushed, hence I'm closing this PR

PR #3136, opened 2019-05-13
Add ipa-cert-fix tool

frasertweedale:
Forward-port of https://github.com/freeipa/freeipa/pull/2919 to `master`
branch.

```
da189c9cc (Fraser Tweedale, 3 days ago)
   require Dogtag 10.7.0-1

   Dogtag 10.7 includes the 'pki-server cert-fix' enhancements required by
   ipa-cert-fix.  Bump the dep min bound.

   Part of: https://pagure.io/freeipa/issue/7885

54e967095 (Fraser Tweedale, 6 weeks ago)
   ipa-cert-fix: use customary exit statuses

   It is customary to return 2 when IPA is not configured, and 1 when other
   required bits are not installed or configured.  Update ipa-cert-fix exit
   statuses accordingly.

   Part of: https://pagure.io/freeipa/issue/7885

7a836fb8c (Fraser Tweedale, 7 weeks ago)
   ipa-cert-fix: add man page

   Part of: https://pagure.io/freeipa/issue/7885

6c023f265 (Fraser Tweedale, 7 weeks ago)
   Add ipa-cert-fix tool

   The ipa-cert-fix tool wraps `pki-server cert-fix`, performing additional
   certificate requests for non-Dogtag IPA certificates and performing
   additional actions.  In particular:

   - Run cert-fix with arguments particular to the IPA deployment.

   - Update IPA RA certificate in the ipara user entry (if renewed).

   - Add shared certificates (if renewed) to the ca_renewal LDAP
    container for replication.

   - Become the CA renewal master if shared certificates were renewed.
    This ensures other CA replicas, including the previous CA renewal
    master if not the current host, pick up those new certificates
    when Certmonger attempts to renew them.

   Fixes: https://pagure.io/freeipa/issue/7885

3d1ff725e (Fraser Tweedale, 7 weeks ago)
   constants: add ca_renewal container

   Part of: https://pagure.io/freeipa/issue/7885

c956bc658 (Fraser Tweedale, 7 weeks ago)
   cainstance: add function to determine ca_renewal nickname

   The ipa-cert-fix program needs to know where to put shared certificates. 
   Extract the logic that computes the nickname from dogtag-ipa-ca-renew-agent
   to new subroutine cainstance.get_ca_renewal_nickname().

   Part of: https://pagure.io/freeipa/issue/7885

4f4c32cc6 (Fraser Tweedale, 7 weeks ago)
   Extract ca_renewal cert update subroutine

   When the CA renewal master renews certificates that are shared across CA
[...]

PR #3180, opened 2019-05-24
[ipa-4-6] ipa-cert-fix fixes

frasertweedale:
Cherry-pick a couple of fixes from master-bound PR #3136.

```
2ff6e1ac5 (Fraser Tweedale, 67 minutes ago)
   ipa-cert-fix: fix spurious renewal master change

   We only want to become the renewal master if we actually renewed a shared
   certificate.  But there is a bug in the logic; even if the only Dogtag
   certificate to be renewed is the 'sslserver' (a non-shared certificate),
   the renewal master will be reset.  Fix the bug.

   A static type system would have excluded this bug.

   Part of: https://pagure.io/freeipa/issue/7885

ee889d2d7 (Fraser Tweedale, 2 days ago)
   ipa-cert-fix: handle 'pki-server cert-fix' failure

   When DS cert is expired, 'pki-server cert-fix' will fail at the final step
   (restart).  When this case arises, ignore the CalledProcessError and
   continue.

   We can't know for sure if the error was due to failure of final restart, or
   something going wrong earlier.  But if it was a more serious failure, the
   next step (installing the renewed IPA-specific certificates) will fail.

   Part of: https://pagure.io/freeipa/issue/7885
```

### DIFF
--- a/ipaserver/install/ipa_cert_fix.py
+++ b/ipaserver/install/ipa_cert_fix.py
@@ -113,7 +113,17 @@ class IPACertFix(AdminTool):
             return 0
         print("Proceeding.")
 
-        run_cert_fix(certs, extra_certs)
+        try:
+            run_cert_fix(certs, extra_certs)
+        except ipautil.CalledProcessError:
+            if any(x[0] is IPACertType.LDAPS for x in extra_certs):
+                # The DS cert was expired.  This will cause
+                # 'pki-server cert-fix' to fail at the final
+                # restart.  Therefore ignore the CalledProcessError
+                # and proceed to installing the IPA-specific certs.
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @frasertweedale 
thanks for the backport, LGTM.

frasertweedale:
ipa-4-6:

* 1ee6bb2a019e6b042de7536a9b8e179d375b8b9e ipa-cert-fix: handle 'pki-server cert-fix' failure
* 4c25a83c30b63228f19ffa7e1e1f6d86fe08fe90 (HEAD) ipa-cert-fix: fix spurious renewal master change

PR #3183, opened 2019-05-24
Require a SASL SSF of >= 56 on client side

tiran:
SSF_MINX 56 level ensures data integrity and confidentiality for SASL
GSSAPI and SASL GSS SPNEGO connections. Although at least AES128 is enforced
pretty much everywhere, 56 is required.

The origianl commit 350954589774499d99bf87cb5631c664bb0707c4 added minimum
SSF on LDAP client and LDAP server. Some LDAP consumers like realmd are
not compatible with strong SSF yet.

Related: https://pagure.io/freeipa/issue/7140
Related: https://pagure.io/freeipa/issue/4580

### DIFF
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -314,6 +314,9 @@ TLS_VERSIONS = [
 ]
 TLS_VERSION_MINIMAL = "tls1.0"
 
+# minimum SASL secure strength factor for LDAP connections
+# 56 provides backwards compatibility with old libraries.
+LDAP_SSF_MIN_THRESHOLD = 56
 
 # Use cache path
 USER_CACHE_PATH = (
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -43,7 +43,9 @@ import six
 
 # pylint: disable=ipa-forbidden-import
 from ipalib import errors, x509, _
-from ipalib.constants import LDAP_GENERALIZED_TIME_FORMAT
+from ipalib.constants import (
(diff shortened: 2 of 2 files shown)


frozencemetery:
Is this > 56 or >= 56?  I believe 56 permits single-DES, which is probably undesirable.

tiran:
It's 56 because some servers don't announce the correct SSF. @simo5 told me that the issue has been fixed just recently. Old machines may still consider AES 128 as 56.

simo5:
Old sasl couldn't ask GSSAPI mechanisms what was the security strenght factor, so it always unconditionally ruterned 56 which was the weakest (DES) cipher used in krb5.
Since last year we can now provide the correct SSF via SASL, so any new client can (and probably should) require >= 128. Note that thes efactor is determined locally, so it doesn't matter if other older machines in the network are not updated.

simo5:
(This change does need to be gated on the availablility of the correct cyrus-sasl and libkrb5 versions that implement the new retporting though, of course).

fcami:
@simo5 would flagging the change for ipa-4-8+ only be enough wrt correct cyrus-sasl and libkrb5 versions?

simo5:
I am not sure what "flagging the change for ipa-4-8+" means, whether you can use this cyrus-sasl feature does not depend on ipa versions, but depends on what version of the library is available on the system you compile that version against.

tiran:
Also causing too much issues with backwards compatibility. Let's do this in a separate hardening effort.

PR #3191, opened 2019-05-27
ipatests: add integration test for ipa-replica-manage list

flo-renaud:
The command
    ipa-replica-manage list -v <node>
can display:
    last init ended: 1970-01-01 00:00:00+00:00
    last init status: None
when called on a node that never had total update.
The fix for 7716 modifies the command so that it doesn't print
those lines when there is no last init status.

This commit adds a new test checking the output of
ipa-replica-manage list -v <node>.

Related to: https://pagure.io/freeipa/issue/7716


flo-renaud:
Note: this PR depends on PR #2415.

rcritten:
master:

* 0b21e2ab9f6f0cc08067d8b5a98efbe4de763904 ipatests: add integration test for ipa-replica-manage list

PR #3194, opened 2019-05-28
dn: handle multi-valued RDNs in Name conversion

frasertweedale:
When applying DN to a cryptography.x509.Name, multi-valued RDNs get
"flattened" into separate RDNs.  Update the constructor to correctly
handle Name values with multi-valued RDNs.

Fixes: https://pagure.io/freeipa/issue/7963


abbra:
LGTM.

abbra:
master:

* 891d54e46f9c237493f7985d4d8ea19d4d051d09 dn: handle multi-valued RDNs in Name conversion

PR #3212, opened 2019-05-29
dn: sort AVAs when converting from x509.Name

frasertweedale:
Equal DNs with multi-valued RDNs can compare inequal if one (or
both) is constructed from a cryptography.x509.Name, because the AVAs
in the multi-valued RDNs are not being sorted.

Sort the AVAs when constructing from Name and add test cases for
equality checks on multi-valued RDNs constructed from inputs with
permuted AVA order.

Part of: https://pagure.io/freeipa/issue/7963

### DIFF
--- a/ipapython/dn.py
+++ b/ipapython/dn.py
@@ -1141,6 +1141,8 @@ class DN:
                     ava.value) for ava in rdn]
                 for rdn in value.rdns
             ]))
+            for rdn in rdns:
+                sort_avas(rdn)
         else:
             raise TypeError(
                 "must be str, unicode, tuple, Name, RDN or DN, got %s instead"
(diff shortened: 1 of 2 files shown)


flo-renaud:
Hi @frasertweedale 
thanks for the PR, works for me. If you rebase, PRCI should pick the new f29 templates and proceed without any issue.

rcritten:
+1 from me as well. Tests out fine.

frasertweedale:
Two +1s make an ACK, I guess.

frasertweedale:
master:

* ad7472970305f7be2d3afc65fda1e86296d118dd (HEAD) dn: sort AVAs when converting from x509.Name

PR #3210, opened 2019-05-29
Add Fedora 30 test definitions and bump PR-CI template version

netoarmando:
- Upgrade template version required by new version of PR-CI.
- Use Fedora 30 as main target for testing.

PR #3216, opened 2019-05-30
fix LWCA key retrieval on f30

frasertweedale:
This PR includes fixes for LWCA key retrieval on f30 and fixes for handling of
missing LWCA keys in the ca_find and ca_show commands.

It is based upon https://github.com/freeipa/freeipa/pull/3210 which updates
PR-CI to f30.  (This PR revealed the issue on f30; the tests are not passing
hence it has not been merged yet.)

```
f029a6e3b (Fraser Tweedale, 7 hours ago)
   ipa-pki-retrieve-key: set KRB5CCNAME

   On Fedora 30, for some reason LDAP GSS-API bind now fails in the 
   ipa-pki-retrieve-key program.  The Dogtag keytab credential acquisition
   does succeed, but those credentials are not used for the LDAP bind.

   Update CustodiaClient to support setting KRB5CCNAME when it creates 
   credentials.  This behaviour is optional and disabled by default (no 
   behavioural change for other use cases).  But enable this behaviour in
   ipa-pki-retrieve-key so the Dogtag credentials are used for the LDAP bind.

   Fixes: https://pagure.io/freeipa/issue/7964

fff5119cd (Fraser Tweedale, 85 minutes ago)
   Handle missing LWCA certificate or chain

   If lightweight CA key replication has not completed, requests for the
   certificate or chain will return 404**.  This can occur in normal
   operation, and should be a temporary condition.  Detect this case and
   handle it by simply omitting the 'certificate' and/or
   'certificate_out' fields in the response, and add a warning message to the
   response.

   Also update the client-side plugin that handles the
   --certificate-out option.  Because the CLI will automatically print the
   warning message, if the expected field is missing from the response, just
   ignore it and continue processing.

   ** after the Dogtag NullPointerException gets fixed!

   Part of: https://pagure.io/freeipa/issue/7964

b59c49351 (Armando Neto, 2 days ago)
   Add Fedora 30 test definitions and bump template version

```

### DIFF
--- a/ipaserver/secrets/client.py
+++ b/ipaserver/secrets/client.py
@@ -9,6 +9,7 @@ from jwcrypto.jwk import JWK
 from ipaserver.secrets.kem import IPAKEMKeys
 from ipaserver.secrets.store import iSecStore
 from ipaplatform.paths import paths
+from ipapython.ipautil import private_ccache
 from base64 import b64encode
 import ldapurl
 import gssapi
@@ -53,7 +54,7 @@ class CustodiaClient:
         # Init creds immediately to make sure they are valid.  Creds
         # can also be re-inited by _auth_header to avoid expiry.
         #
-        self.creds = self.init_creds()
+        self.creds, self.ccname = self.init_creds()
 
         self.service_name = gssapi.Name('HTTP@%s' % (server,),
                                         gssapi.NameType.hostbased_service)
@@ -61,8 +62,11 @@ class CustodiaClient:
 
         self.ikk = IPAKEMKeys({'server_keys': keyfile, 'ldap_uri': ldap_uri})
 
-        self.kemcli = KEMClient(self._server_keys(server, realm),
-                                self._client_keys())
+        # Use the acquired credentials to look up keys in LDAP by
+        # temporarily setting KRB5CCNAME.
+        with private_ccache(self.ccname):
+            self.kemcli = KEMClient(self._server_keys(server, realm),
+                                    self._client_keys())
 
         self.keystore = self._keystore(realm, ldap_uri, auth_type)
 
@@ -73,14 +77,15 @@ class CustodiaClient:
     def init_creds(self):
         name = gssapi.Name(self.client_service,
                            gssapi.NameType.hostbased_service)
-        store = {'client_keytab': self.keytab,
-                 'ccache': 'MEMORY:Custodia_%s' % b64encode(
-                     os.urandom(8)).decode('ascii')}
-        return gssapi.Credentials(name=name, store=store, usage='initiate')
+        ccname = "MEMORY:Custodia_{}".format(
+            b64encode(os.urandom(8)).decode('ascii'))
+        store = {'client_keytab': self.keytab, 'ccache': ccname}
+        creds = gssapi.Credentials(name=name, store=store, usage='initiate')
+        return creds, ccname
 
     def _auth_header(self):
         if not self.creds or self.creds.lifetime < 300:
-            self.creds = self.init_creds()
+            self.creds, self.ccname = self.init_creds()
         ctx = gssapi.SecurityContext(name=self.service_name, creds=self.creds)
         authtok = ctx.step()
         return {'Authorization': 'Negotiate %s' % b64encode(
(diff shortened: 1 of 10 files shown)


ipaserver/secrets/client.py
```diff
@@ -73,10 +77,14 @@ def __init__(
...
-                 'ccache': 'MEMORY:Custodia_%s' % b64encode(
-                     os.urandom(8)).decode('ascii')}
-        return gssapi.Credentials(name=name, store=store, usage='initiate')
+        ccname = "MEMORY:Custodia_{}".format(
+            b64encode(os.urandom(8)).decode('ascii'))
+        store = {'client_keytab': self.keytab, 'ccache': ccname}
+        creds = gssapi.Credentials(name=name, store=store, usage='initiate')
+        if self.set_krb5ccname:
+            # do this _after_ 'creds' was successfully constructed
+            os.environ['KRB5CCNAME'] = ccname
```
abbra: I would have preferred if you could refactor this use of CustodiaClient.init_creds as a context manager like I did with ipatests/test_cmdline/test_ipagetkeytab.py. So far, I can see that most of CustodiaClient instance consumers simply call fetch_key() or just call a constructor. I would prefer do switch that to a code that does something like

```
client = CustodiaClient(..)
with client.init_creds() as cli:
    cli.fetch_key()
```
This would make the context under which execution happens clearly defined and will reduce potential time we change `KRBCCNAME`.

frasertweedale:
Well, it did not fix it; LWCA key replication is still busted, even with the change (which worked locally >:[ ). So there will be an opportunity to address your requets too, @abbra.

frasertweedale:
We will probably need to update PR-CI to use the @pki/master COPR so we can get Dogtag master builds into the environment.  I have a pr-ci PR: https://github.com/freeipa/freeipa-pr-ci/pull/276.

The bug itself is not in Dogtag but unfortunately Dogtag is currently swallowing subprocess output so we have little info about where and how the `ipa-pki-retrieve-key` program is failing.  There's a Dogtag PR to improve that situation: https://github.com/dogtagpki/pki/pull/215.

frasertweedale:
@abbra I'm stumped.

We know it is kerb/gss related.  Looking at `krb5kdc.log` the `AS_REQ` succeeds but there is no subsequent `TGS_REQ` for the service ticket for LDAP.  It seemingly cannot see the credentials.  Even though we have just set `KRB5CCNAME`.

I can't reproduce it on my local f30 deployment (i.e. the patch resolves the issue on my local deployment, but not on CI).  I'd love to hear any suggestions why...

We've got a build in the `@pki/master` COPR that ensure `ipa-pki-retrieve-key` stderr goes to parent stdout (it will end up in journal).  But CI images need to be updated to use that COPR.  Until then we can't tell exactly where `ipa-pki-retrieve-key` is failing.

abbra:
@frasertweedale you cannot use MEMORY: type ccache for applications that call out for other processes because those other processes have no access to the memory of _your_ process. You have to use FILE: or DIR: (or KEYRING:/KCM:) for that. I'd suggest you would use DIR: ccache and do move ipaserver/install/server/upgrade.py:empty_ccache to some common place for reuse, then take that context manager to use in custodia. May be accept a prefix and a dir as arguments to empty_ccache. THis will, basically, give you a temporary ccache to reuse that can be safely used for any app sharing the context.

frasertweedale:
@abbra but it is all within the same process.  Dogtag invokes `ipa-pki-retrieve-key`, and all the setting of `KRB5CCACHE`, and connections to LDAP and the custodia server, happen within that process.

abbra:
@frasertweedale @tiran did change the code to run out of the process to address privilege separation in master.

tiran:
@frasertweedale I changed the clients to fix various issues with privilege seperation, wrong privileges for SoftHSM2 keys, and DAC override. See https://pagure.io/freeipa/issue/6888

Now all handlers are executed in a separate process and most of them also run as a different, less privileged user account.

frasertweedale:
@tiran @abbra thanks for the additional info; I'm busy with RH362 this week but will try and squeeze out a proper fix soon!

frasertweedale:
Closing in favour of https://github.com/freeipa/freeipa/pull/3262.

PR #3217, opened 2019-05-30
ipatests: allow to relax security of LDAP connection from controller to IPA host

wladich:
The Host.ldap_connect() method uses LDAPClient from ipapython package.
In a3934a21 we started to use secure connection from tests controller to
ipa server. And also 5be9341f changed the LDAPClient.simple_bind method
to forbid password based authentiction over insecure connection.
This makes it imposible to establish ldap connection in some test
configurations where hostnames known to ipa server do not match ones known
to tests controller (i.e. when host.hostname != host.external_hostname)
because TLS certificate is issued for host.hostname and test controller
tries to verify it against host.external_hostname.

A sublass of LDAPClient is provided which allows to skip certificate check.

Fixes: https://pagure.io/freeipa/issue/7960

### DIFF
--- a/ipatests/pytest_ipa/integration/host.py
+++ b/ipatests/pytest_ipa/integration/host.py
@@ -21,10 +21,38 @@
 import subprocess
 import tempfile
 
+import ldap
 import pytest_multihost.host
 
 from ipaplatform.paths import paths
 from ipapython import ipaldap
+
+
+class LDAPClientWithoutCertCheck(ipaldap.LDAPClient):
+    """Adds an option to disable certificate check for TLS connection
+
+    To disable certificate validity check create client with added option
+    no_certificate_check:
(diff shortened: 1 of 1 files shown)


wladich:
@flo-renaud, good point about class name, changed it.
Also changed:
* added class docstring
* renamed new parameter ignore_certificate_errors -> no_certificate_check
* use constant instead of value in call to set_option

flo-renaud:
Hi @wladich 
I tested on my env with external hostname different from hostname by running test_integration/test_simple_replication.py, works well.
If you rebase, PRCI should pick the latest fed 29 images and proceed smoothly, then I'll ACK when the test results are green.

flo-renaud:
As tests are now green, ACK.

wladich:
master:

* cd2b2443c5ba2f4885e3d85f98332b2fa8b30432 ipatests: allow to relax security of LDAP connection from controller to IPA host

PR #3226, opened 2019-06-03
Stop using 389-ds legacy backup and restoration utilities

rcritten:
Use dsctl instead, the modern replacement for ldif2db, db2ldif,
bak2db and db2bak.

https://pagure.io/freeipa/issue/7965

### DIFF
--- a/ipaserver/install/ipa_restore.py
+++ b/ipaserver/install/ipa_restore.py
@@ -634,10 +634,11 @@ class Restore(admintool.AdminTool):
             # Restore SELinux context of template_dir
             tasks.restore_context(template_dir)
 
-            args = [paths.LDIF2DB,
-                    '-Z', instance,
-                    '-i', ldiffile,
-                    '-n', backend]
+            args = [paths.DSCTL,
+                    instance,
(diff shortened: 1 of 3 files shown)


ipaserver/install/ipa_restore.py
```diff
@@ -687,8 +688,9 @@ def bak2db(self, instance, backend, online=True):
             logger.info("Waiting for restore to finish")
             wait_for_task(conn, dn)
         else:
-            args = [paths.BAK2DB,
-                    '-Z', instance,
+            args = [paths.DSCTL,
```
tiran: ``dsctl`` is not compatible with ``if backend is not None: args.append('-n')``

rcritten: You're right. Not caught because bak2db isn't used anywhere. If I remember correctly I added it as a parallel to the db2bak call (belt and suspenders with backup). I'll fix it up.

rcritten:
master:

* f606d82024d3a74df2aeb5448944aa015e96cce2 Stop using 389-ds legacy backup and restoration utilities

PR #3228, opened 2019-06-04
When reading SSH pub key don't assume last character is newline

rcritten:
The code was attempting to strip off any trailing newline and then
calling lstrip() on the rest.

This assumes that the key has a trailing newline. At best this
can cause the last character of the comment to be lost. If there
is no comment it will fail to load the key because it is invalid.

Patch by Félix-Antoine Fortin <felix-antoine.fortin@calculquebec.ca>

https://pagure.io/freeipa/issue/7959

There is a lot of very good analysis within the ticket.

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -1562,12 +1562,13 @@ def update_ssh_keys(hostname, ssh_dir, create_sshfp):
             continue
 
         for line in f:
-            line = line[:-1].lstrip()
+            line = line.strip()
             if not line or line.startswith('#'):
                 continue
             try:
                 pubkey = SSHPublicKey(line)
-            except (ValueError, UnicodeDecodeError):
+            except (ValueError, UnicodeDecodeError) as e:
+                logger.debug("Decoding line '%s' failed: %s", line, e)
                 continue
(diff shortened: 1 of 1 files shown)


netoarmando:
This needs to be rebased, `master` branch has updated pr-ci definitions required to execute gating tests.

flo-renaud:
Hi @rcritten 
thanks for the PR, works for me. I tested with a file /etc/ssh/ssh_host_rsa_key.pub not containing \r or \n at the end of the line, by calling `ipa-client-install --ssh-trust-dns` and verified that the corresponding SSHFP record was properly added.
A nitpick: we could add a debug msg or a warning when the line cannot be decoded (lines 1565-1571) instead of ignoring the issue:
```
            try:
                pubkey = SSHPublicKey(line)
            except (ValueError, UnicodeDecodeError):
                # here log a msg
                logger.warning("Failed to add SSH public key from %s", filename)
                continue
```

flo-renaud:
Hi @rcritten 
thanks for the updated PR, works perfectly.

rcritten:
master:

* 21777e4ba04594c09a2a20ae92ab8fe2cd908fad When reading SSH pub key don't assume last character is newline

PR #3229, opened 2019-06-04
Better handling for not configured when uninstalling client

rcritten:
Second crack at this.

When uninstalling the client and it isn't configured we should return "not configured". When uninstall the client in context of the server we shouldn't error out.

So in the case of on-master raise the exception with a SUCCESS error so it doesn't cause things to blow up, and revert the previous attempt.

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -3175,9 +3175,13 @@ def uninstall_check(options):
     fstore = sysrestore.FileStore(paths.IPA_CLIENT_SYSRESTORE)
 
     if not is_ipa_client_installed(fstore):
+        if options.on_master:
+            rval = SUCCESS
+        else:
+            rval = CLIENT_NOT_CONFIGURED
         raise ScriptError(
             "IPA client is not configured on this system.",
-            rval=CLIENT_NOT_CONFIGURED)
+            rval=rval)
 
     server_fstore = sysrestore.FileStore(paths.SYSRESTORE)
     if server_fstore.has_files() and not options.on_master:
--- a/ipaclient/install/ipa_client_install.py
+++ b/ipaclient/install/ipa_client_install.py
@@ -62,7 +62,6 @@ ClientInstall = cli.install_tool(
     verbose=True,
     console_format='%(message)s',
     uninstall_log_file_name=paths.IPACLIENT_UNINSTALL_LOG,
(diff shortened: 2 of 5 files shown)


flo-renaud:
Hi @rcritten 
thanks for the PR. I tested the following:
- [x] ipa-client uninstall on a not installed node: retcode=2 + msg `IPA client is not configured on this system`
- [x] ipa-server uninstall on a not installed node: retcode=0 + msg showing it's trying to remove the config anyway:
```Removing IPA client configuration
The ipa-client-install command was successful
The ipa-server-install command was successful
```
I think the result is as expected because we want to force cleanup for a partially uninstalled master.
- [x] ipa-server uninstall on a client: retcode=0 + same message as above, expected because we want to force cleanup for a partially uninstalled master.
- [x] ipa-client uninstall on a master: retcode=2 + `IPA client is configured as a part of IPA server on this system. Refer to ipa-server-install for uninstallation.`

But I noticed that ipatests/test_cmdline/test_cli.py::test_command_ipa_not_installed is failing for the commands ipa-certupdate and ipa-client-automount (probably related to a previous commit), and the failure is not detected by travis-ci (maybe because the tests are executed by a non-root user?).
Could you leverage this PR to fix the test?
```
        (['ipa-certupdate'], 1, None, IPA_CLIENT_NOT_CONFIGURED),
        (['ipa-client-automount'], 1, IPA_CLIENT_NOT_CONFIGURED, None),
```
should be replaced with
```
        (['ipa-certupdate'], 2, None, IPA_CLIENT_NOT_CONFIGURED),
        (['ipa-client-automount'], 2, IPA_CLIENT_NOT_CONFIGURED, None),
```

rcritten:
These tests are skipped because the server is installed.

I added a patch to fix up these values, verified with some manual testing. I'll open a ticket to properly fix this:

1. Pull this test out into a separate file because the other tests rely on an installed server
2. Run this new file separately in CI

flo-renaud:
Hi @rcritten 
thanks for the updated PR, works for me.

flo-renaud:
master:

* 1284bf1588877d3549ddc98e4762038b8740c72e Drop list of return values to be ignored in AdminTool
* c1c50650a7f359aa9fd77d4348c31169ca878003 Return 0 on uninstall when on_master for case of not installed
* cef4edd384baeda913e309ebcc9a9dd8753dc744 Fix expected return code in tests when server is uninstalled

PR #3231, opened 2019-06-05
azure tests: make sure /etc/docker folder exists

abbra:
Azure tests fail because we couldn't configure docker for IPv6 anymore.
This happened because we weren't able to copy our configuration file to
/etc/docker -- looks like the docker directory does not exist.


abbra:
master:

* 8f4ca3957c84e3f9339bcfb7c511f80553c58e00 azure tests: make sure /etc/docker folder exists

PR #3232, opened 2019-06-05
ipatests: add test for sudo with runAsUser and domain resolution order.

wladich:
Running commands with sudo as specific user should succeed
when sudo rule has ipasudorunas field defined with value of that user
and domain-resolution-order is defined in ipa config.

Relates to https://pagure.io/SSSD/sssd/issue/3957

### DIFF
--- a/ipatests/test_integration/test_sudo.py
+++ b/ipatests/test_integration/test_sudo.py
@@ -19,6 +19,7 @@
 
 import pytest
 
+from ipaplatform.osinfo import osinfo
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration.tasks import (
     clear_sssd_cache, get_host_ip_with_hostmask, modify_sssd_conf)
@@ -714,6 +715,9 @@ class TestSudo(IntegrationTest):
                                           raiseonerr=False)
         assert result.returncode != 0
 
+    @pytest.mark.xfail(
+        osinfo.id == 'fedora' and osinfo.version_number < (30,),
+        reason="https://pagure.io/SSSD/sssd/issue/3957", strict=True)
     def test_domain_resolution_order(self):
         """Test sudo with runAsUser and domain resolution order.
 
--- a/ipatests/test_integration/test_sudo.py
+++ b/ipatests/test_integration/test_sudo.py
@@ -713,3 +713,40 @@ class TestSudo(IntegrationTest):
                                           '--groups', 'testgroup2'],
                                           raiseonerr=False)
         assert result.returncode != 0
+
+    def test_domain_resolution_order(self):
+        """Test sudo with runAsUser and domain resolution order.
+
+        Regression test for bug https://pagure.io/SSSD/sssd/issue/3957.
+        Running commands with sudo as specific user should succeed
+        when sudo rule has ipasudorunas field defined with value of that user
+        and domain-resolution-order is defined in ipa config.
+        """
+        self.master.run_command(
+            ['ipa', 'config-mod', '--domain-resolution-order',
+             self.domain.name])  # pylint: disable=no-member
+        try:
+            # prepare the sudo rule: set only one user for ipasudorunas
+            self.reset_rule_categories()
+            self.master.run_command(
+                ['ipa', 'sudorule-mod', 'testrule',
+                 '--runasgroupcat=', '--runasusercat='],
+                raiseonerr=False
+            )
+            self.master.run_command(
+                ['ipa', 'sudorule-add-runasuser', 'testrule',
+                 '--users', 'testuser2'])
+
+            # check that testuser1 is allowed to run commands as testuser2
+            # according to listing of allowed commands
+            result = self.list_sudo_commands('testuser1')
+            expected_rule = ('(testuser2@%s) NOPASSWD: ALL'
+                             % self.domain.name)  # pylint: disable=no-member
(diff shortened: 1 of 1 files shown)


wladich:
Update: ignore pylint no-member errors

rcritten:
The test looks good to me but this doesn't execute it. Do you know the plans on when the updated SSSD with this change will be released? Can we change the skip to a check for < some version of SSSD?

ipatests/test_integration/test_sudo.py
```diff
@@ -713,3 +713,42 @@ def test_category_all_validation_runasgroup(self):
...
+                raiseonerr=False
+            )
+            self.master.run_command(
+                ['ipa', 'sudorule-add-runasuser', 'testrule',
+                 '--users', 'testuser2'])
+
+            # check that testuser1 is allowed to run commands as testuser2
+            # according to listing of allowed commands
+            result = self.list_sudo_commands('testuser1')
+            expected_rule = ('(testuser2@%s) NOPASSWD: ALL'
```
flo-renaud: The rule expects the qualified name but with the SSSD version used in PRCI, the non qualified name is returned:
```
transport.py 513 DEBUG RUN su -c "sudo -l -n" testuser1
transport.py 558 DEBUG Matching Defaults entries for testuser1@ipa.test on client0:
transport.py 558 DEBUG !visiblepw, env_reset, env_keep="COLORS DISPLAY HOSTNAME HISTSIZE KDEDIR LS_COLORS", env_keep+="MAIL PS1 PS2 QTDIR USERNAME LANG LC_ADDRESS LC_CTYPE", env_keep+="LC_COLLATE LC_IDENTIFICATION LC_MEASUREMENT LC_MESSAGES", env_keep+="LC_MONETARY LC_NAME LC_NUMERIC LC_PAPER LC_TELEPHONE", env_keep+="LC_TIME LC_ALL LANGUAGE LINGUAS _XKB_CHARSET XAUTHORITY", secure_path=/usr/local/sbin\:/usr/local/bin\:/usr/sbin\:/usr/bin\:/sbin\:/bin, !authenticate
transport.py 558 DEBUG
transport.py 558 DEBUG User testuser1@ipa.test may run the following commands on client0:
transport.py 558 DEBUG (testuser2) NOPASSWD: ALL
transport.py 217 DEBUG Exit code: 0
```
If you want to test the next command, you can relax the expected rule and use a matching reg exp instead.

wladich: I consider this non-qualified name in sudo output to be another symptom of bug from issue 3954.

wladich:
@rcritten 
@jhrozek said he is planning to release sssd about June 17, but I do not know for which versions of Fedoras.
About sssd version -- we do not have sssd installed on machine where tests are executed. So to get the sssd version at ipa master we need the multihost object which is not available in decorators.

jhrozek:
We're probably going to release a new version tomorrow. Where do you grab packages from? Maybe we can patch that resource in the meantime.

rcritten:
I guess I'd rather see an xfail than a skip. Having some detection whether the test should be executable on a given platform would be ideal.

wladich:
I do not know what will be the version number of sssd where this bug will be fixed. If the fixed version will  not be released today, I will remove "needs review" label and get back to this PR on June 24 when I return from vacations...

wladich:
Postponing PR till June 24

wladich:
@jhrozek 
Have the fix for https://pagure.io/SSSD/sssd/issue/3957 landed Fedora? If yes, then in which releases of Fedora?

jhrozek:
I think yes but check with @mzidek-rh he maintains fedora

mzidek-rh:
@wladich It is in rawhide and Fedora 30. In Fedora 29 it does not have enough karma yet.

ipatests/test_integration/test_sudo.py
```diff
@@ -713,3 +714,43 @@ def test_category_all_validation_runasgroup(self):
                                           '--groups', 'testgroup2'],
                                           raiseonerr=False)
         assert result.returncode != 0
+
+    @pytest.mark.xfail(
+        osinfo.id == 'fedora' and osinfo.version_number <= (30,),
```
flo-renaud: Hi @wladich 
According to [bodhi](https://bodhi.fedoraproject.org/updates/FEDORA-2019-1575ccd38c), the fix is available in sssd 2.2.0 in fedora 30 since June 25. Shouldn't the check be `osinfo.version_number < (30,)` rather than `<=` ?

wladich: good point, thank you

wladich:
Test is failing because Fedora 30 images contain an old version of sssd.

netoarmando:
@wladich new template available: https://github.com/freeipa/freeipa/pull/3359

flo-renaud:
Hi @wladich 
The test LGTM.
We may need to monitor [bodhi](https://bodhi.fedoraproject.org/updates/FEDORA-2019-c1aa6029a3) for the SSSD version in fedora 29. If sssd 2.2.0 makes it into fedora-updates, we will need to change the xfail.

wladich:
master:

* 0d15eb78d4178bea6057e4d32b234c592ca80fa7 ipatests: add test for sudo with runAsUser and domain resolution order.
* 4740655260caa7baee4473b0d35840a4c0da3dac ipatests: mark test_domain_resolution_order as expectedly failing

PR #3240, opened 2019-06-07
ipatests: fix test_caless.py

flo-renaud:
Commit e3f3584 introduced an additional prompt in ipa-server-install "Do you want to configure chrony with NTP server or pool address?".

The test is building a string passed to stdin in interactive mode but this string has not been updated with the additional answer for this new question.

This commit answers 'no' to the question and allows to proceed with the ipa server installation.

Fixes: https://pagure.io/freeipa/issue/7969


flo-renaud:
@Tiboris 

Tiboris:
master:

* 07f7e3eaecf781b45c972c4348a9db5794e3fcaf ipatests: fix test_caless.py

PR #3242, opened 2019-06-07
Work around autoremember-build timing issue test failures

rcritten:
tests: Wait for automember rebuild --no-wait tasks to finish
    
The behavior of automember changed with the design
such that members are "cleaned up" first then re-added. This has 
the effect of removing members that no longer apply to a rule.
    
This was breaking the automember rebuild tests because sometimes
the tests were faster than 389-ds causing memberships to be missed.
    
This does a client-side wait for the task to finish up so still
exercises the rebuild code.
    
https://pagure.io/freeipa/issue/7972
    
### DIFF
--- a/ipatests/test_xmlrpc/test_automember_plugin.py
+++ b/ipatests/test_xmlrpc/test_automember_plugin.py
@@ -31,7 +31,9 @@ from ipapython.dn import DN
 from ipapython.ipautil import run
 from ipatests.test_xmlrpc.xmlrpc_test import XMLRPC_test, raises_exact
 from ipatests.util import assert_deepequal
+from ipaserver.plugins.automember import REBUILD_TASK_CONTAINER
 
+import time
 import pytest
 import unittest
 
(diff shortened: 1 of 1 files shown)


rcritten:
It might be instructive to force a re-run of the xmlrpc tests to demonstrate the tests actually pass

rcritten:
Working in BZ to try to come up with a reproducible case.

rcritten:
I think I have it worked out now, dropping WIP and adding 4-7 backport

ipatests/test_xmlrpc/test_automember_plugin.py
```diff
@@ -275,6 +277,52 @@ def set_automember_process_modify_ops(value):
...
+    try:
+        task_entries, unused = ldap.find_entries(
+            base_dn=REBUILD_TASK_CONTAINER,
+            filter='(&(!(nstaskexitcode=0))(scope=*))')
+    except errors.NotFound:
+        # either it's done already or it never started
+        return
+
+    # we run these serially so there should be only one at a time
+    assert len(task_entries) == 1
```
flo-renaud: In theory we could have more than one task with the filter !(nstaskexitcode=0), for instance one on-going task with no nstaskexitcode yet and a previous task that failed somehow (nstaskexitcode = 1). But I agree that this case has low probability during the tests and it would probably be overkill to retrieve the task id from automember-rebuild --nowait and look for this id in this method.

rcritten:
master:

* 7ec0976cce61f829cb53dc96581dea2c569706d4 tests: Wait for automember rebuild --no-wait tasks to finish

PR #3250, opened 2019-06-11
ipatest : POSIX attributes, are no longer overwritten or missing.

amore17:
During SSSD lookups, SSSD incorrectly considers the attributes to be removed from the server, and removes them from the SSSD cache as well.
This test validates that POSIX attributes, such as shell or home directory, are no longer overwritten or missing.
Related Ticket : https://pagure.io/SSSD/sssd/issue/2474

### DIFF
--- a/ipatests/test_integration/test_trust.py
+++ b/ipatests/test_integration/test_trust.py
@@ -12,8 +12,6 @@ from ipatests.pytest_ipa.integration import tasks
 class TestTrust(IntegrationTest):
     topology = 'line'
     num_ad_domains = 1
-    num_ad_subdomains = 1
-    num_ad_treedomains = 1
 
     upn_suffix = 'UPNsuffix.com'
     upn_username = 'upnuser'
@@ -33,18 +31,6 @@ class TestTrust(IntegrationTest):
         cls.ad_domain = cls.ad.domain.name
         tasks.install_adtrust(cls.master)
         cls.check_sid_generation()
-
-        cls.child_ad = cls.ad_subdomains[0]  # pylint: disable=no-member
-        cls.ad_subdomain = cls.child_ad.domain.name
-        cls.tree_ad = cls.ad_treedomains[0]  # pylint: disable=no-member
-        cls.ad_treedomain = cls.tree_ad.domain.name
-
-        # values used in workaround for
-        # https://bugzilla.redhat.com/show_bug.cgi?id=1711958
-        cls.srv_gc_record_name = \
-            '_ldap._tcp.Default-First-Site-Name._sites.gc._msdcs'
-        cls.srv_gc_record_value = '0 100 389 {}.'.format(cls.master.hostname)
-
 
     @classmethod
     def check_sid_generation(cls):
@@ -98,7 +84,7 @@ class TestTrust(IntegrationTest):
 
     def test_trustdomains_found_in_nonposix_trust(self):
         self.check_trustdomains(
-            self.ad_domain, [self.ad_domain, self.ad_subdomain])
+            self.ad_domain, [self.ad_domain])
 
     def test_range_properties_in_nonposix_trust(self):
         self.check_range_properties(self.ad_domain, 'ipa-ad-trust', 200000)
@@ -180,7 +166,7 @@ class TestTrust(IntegrationTest):
     def test_trustdomains_found_in_posix_trust(self):
         """Tests that all trustdomains can be found."""
         self.check_trustdomains(
-            self.ad_domain, [self.ad_domain, self.ad_subdomain])
+            self.ad_domain, [self.ad_domain])
 
     def test_range_properties_in_posix_trust(self):
         """Check the properties of the created range"""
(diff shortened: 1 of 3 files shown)


wladich:
@amore17 
```
The linter is failing:
************* Module ipatests.test_integration.test_trust
ipatests/test_integration/test_trust.py:260: [E1101(no-member), TestTrust.test_establish_forest_trust_with_shared_secret] Instance of 'TestTrust' has no 'srv_gc_record_name' member)
ipatests/test_integration/test_trust.py:261: [E1101(no-member), TestTrust.test_establish_forest_trust_with_shared_secret] Instance of 'TestTrust' has no 'srv_gc_record_value' member)
ipatests/test_integration/test_trust.py:282: [E1101(no-member), TestTrust.test_trustdomains_found_in_forest_trust_with_shared_secret] Instance of 'TestTrust' has no 'ad_subdomain' member)
ipatests/test_integration/test_trust.py:314: [E1101(no-member), TestTrust.test_remove_forest_trust_with_shared_secret] Instance of 'TestTrust' has no 'srv_gc_record_name' member)
ipatests/test_integration/test_trust.py:315: [E1101(no-member), TestTrust.test_remove_forest_trust_with_shared_secret] Instance of 'TestTrust' has no 'srv_gc_record_value' member)
```

wladich:
Please provide meaningful commit message (not the PR comment) with link to pagure ticket .

wladich:
Are you sure that this PR must  be merged to ipa-4-7 but not to master?

ipatests/test_integration/test_trust.py
```diff
@@ -463,6 +357,50 @@ def test_user_uid_resolution_in_external_trust_with_shared_secret(self):
         assert re.search(
             testuser_regex, result.stdout_text), result.stdout_text
 
+    def test_override_homedir(self):
+        """
```
wladich: As per https://www.freeipa.org/page/Python_Coding_Style docstrings should comply PEP257 (https://www.python.org/dev/peps/pep-0257/#id17):
 * docstring should start on same line as quotes
 * first line should shortly describe the method
 * 'The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".'
 * There must be an empty line between first line and remaining lines.

ipatests/test_integration/test_trust.py
```diff
@@ -463,6 +357,50 @@ def test_user_uid_resolution_in_external_trust_with_shared_secret(self):
...
+
+        assert "/home/testuser" in result2.stdout_text
+
+        log_file = "/var/log/sssd/sssd_{0}.log" .format(self.master.domain.name)
+        sssd_debug_log = self.master.get_file_contents(
+            log_file, encoding='utf-8'
+        )
+        assert 'get_subdomain_homedir_of_user failed' not in sssd_debug_log
+
+        tasks.restore_files(self.master)
```
wladich: The cleanup will not be executed if test fails at some assertion. To make it run regardless of test success you can put the test in "try" block and cleanup statements in "finally" block

wladich: If you setup trust in the testcase, you should also remove it in teardown

ipatests/test_integration/test_trust.py
```diff
@@ -463,6 +357,50 @@ def test_user_uid_resolution_in_external_trust_with_shared_secret(self):
         assert re.search(
             testuser_regex, result.stdout_text), result.stdout_text
 
+    def test_override_homedir(self):
```
wladich: You inserted the test case between steps of the test for external trust with shared secret, and at the same time you setup trust in your testcase, which is completely wrong. I see two options:
1) add this test case as a step of "Tests for posix AD trust". Do not re-establish and remove trust in this case
2) leave this test case as a self-contained scenario with trust setup and teardown but place it after all other tests.

ipatests/test_integration/test_trust.py
```diff
@@ -463,6 +357,50 @@ def test_user_uid_resolution_in_external_trust_with_shared_secret(self):
...
+        """
+        https://pagure.io/SSSD/sssd/issue/2474
+        POSIX attributes, such as shell or home directory, should not be overwritten or missing.
+        """
+        tasks.establish_trust_with_ad(self.master, self.ad_domain,
+                                      extra_args=['--range-type', 'ipa-ad-trust'])
```
wladich: The bugzilla says that to reproduce the original bug you need "IPA-AD trust with POSIX attributes". Here you establish non-POSIX trust. To establish POSIX trust range type should be "ipa-ad-trust-posix"

ipatests/test_integration/test_trust.py
```diff
@@ -463,6 +357,50 @@ def test_user_uid_resolution_in_external_trust_with_shared_secret(self):
...
+            self.master,
+            domain.name,
+            {
+                'subdomain_homedir': '%o'
+            }
+        )
+        sssd_conf = self.master.get_file_contents(
+            paths.SSSD_CONF, encoding='utf-8'
+        )
+        assert "subdomain_homedir" in sssd_conf
```
wladich: This check is redundant, we do not need to verify that utility functions are correct

ipatests/test_integration/test_trust.py
```diff
@@ -463,6 +357,50 @@ def test_user_uid_resolution_in_external_trust_with_shared_secret(self):
...
+        tasks.clear_sssd_cache(self.master)
+
+        run_cmd = '"import os; os.initgroups({0}, 1234)"' .format(testuser)
```
wladich: This looks ugly to me. Could you please ask Jakub Hrozek if there is some other way to trigger initgroups? Maybe something like "sudo -u testuser" will work.

PR #3261, opened 2019-06-12
WebUI: Fix 'user not found' traceback on user ID override details page

serg-cymbaluk:
Disable link to user page from user ID override in case it is in 'Default Trust View'

Ticket: https://pagure.io/freeipa/issue/7139

### DIFF
--- a/install/ui/src/freeipa/idviews.js
+++ b/install/ui/src/freeipa/idviews.js
@@ -472,6 +472,17 @@ idviews.id_override_user_details_facet = function(spec) {
         that.refresh();
     };
 
+    that.load = function(data) {
+        var is_trust_view = that.get_pkeys()[0] === idviews.DEFAULT_TRUST_VIEW;
+        var widget = that.fields.get_field('ipaanchoruuid').widget;
+
+        // Disable link for AD users
+        widget.no_check = is_trust_view;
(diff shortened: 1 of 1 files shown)


serg-cymbaluk:
@flo-renaud thank you for the feedback

@pvoborni could you please check the code?

pvoborni:
The Web UI code looks good to me. I did not check behavior, but Flo did that. From my PoV it can get an ACK.

For the linked, upstream ticket. Shouldn't we also made a change in the server itself to not throw traceback for ad users? (in different PR) IMHO this patch only hides the issue when it the bug is triggered from Web UI. This patch is still good because we will save an unnecessary call.

serg-cymbaluk:
> Shouldn't we also made a change in the server itself to not throw traceback for ad users?

@pvoborni I think server works correct. It returns only one specific entity type, so it is expected behavior. One API call shouldn't solve multiple different problems.

flo-renaud:
@pvoborni 
ipa user-show aduser@ad.domain.com should not find any user, so IMHO the behavior is correct. ACK'ing the PR.

serg-cymbaluk:
master:

* 881ec5a31775a5643ca3c6a37b651fd2104d25b9 WebUI: Fix 'user not found' traceback on user ID override details page

PR #3280, opened 2019-06-15
Exit on fail in azure multiline script

stanislavlevin:
By default, the `last` exit code returned from Azure script will be
checked and, if non-zero, treated as a step failure.  Luckily,
for Linux script is a shortcut for Bash. Hence errexit/e option
could be applied. But Azure pipelines doesn't set it by default:
https://github.com/microsoft/azure-pipelines-agent/issues/1803

For multiline script this is a problem unless otherwise designed.
Some of the benefits of checking the result of each subcommand:
- preventing subsequent issues (broken packages, container images, etc.)
- time saving (next steps will not run)
- good diagnostics (tells which part of the script fails)

### DIFF
--- a/ipatests/azure/azure-pipelines.yml
+++ b/ipatests/azure/azure-pipelines.yml
@@ -16,19 +16,23 @@ jobs:
   steps:
     - template: templates/prepare-build.yml
     - script: |
+        set -e
         echo "Running autoconf generator"
         ./autogen.sh
       displayName: Configure the project
     - script: |
+        set -e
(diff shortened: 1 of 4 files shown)


abbra:
LGTM.

abbra:
Let us run full test too.

abbra:
master:

* b5bb436e9fece23f76b35055eddf638f7b8251fe Exit on fail in azure multiline script

stanislavlevin:
Thank you!

PR #3282, opened 2019-06-16
Make use of the single configuration point for the default shells

stanislavlevin:
For now, all the default shells of users and admin are hardcoded in
different parts of the project. This makes it impossible to run the
test suite against the setup, which has the default shell differed
from '/bin/sh'.

The single configuration point for the shell of users and admin is
added to overcome this limitation.

Fixes: https://pagure.io/freeipa/issue/7978

### DIFF
--- a/install/share/bootstrap-template.ldif
+++ b/install/share/bootstrap-template.ldif
@@ -237,7 +237,7 @@ sn: Administrator
 uidNumber: $IDSTART
 gidNumber: $IDSTART
 homeDirectory: /home/admin
-loginShell: /bin/bash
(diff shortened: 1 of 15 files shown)


frasertweedale:
PR-CI, where are you?

tiran:
@frasertweedale, @stanislavlevin does not have commit rights. You have to trigger CI for non-contributors manually.

frasertweedale:
@cheimes, ah, thanks for explaining.

tiran:
master:

* d86b57c05764f771b1568409efd31c9e7b302919 Make use of the single configuration point for the default shells

PR #3284, opened 2019-06-17
Fix Pytest4.x warnings

stanislavlevin:
This fixes:
- "message" parameter of pytest.raises is deprecated since Pytest4.1
https://docs.pytest.org/en/latest/deprecations.html#message-parameter-of-pytest-raises
- pytest.config global is deprecated since Pytest4.1
https://docs.pytest.org/en/latest/deprecations.html#pytest-config-global

### DIFF
--- a/ipatests/conftest.py
+++ b/ipatests/conftest.py
@@ -13,6 +13,7 @@ import pytest
 
 from ipalib import api
 from ipalib.cli import cli_plugins
+import ipatests.util
 
 try:
     import ipaplatform  # pylint: disable=unused-import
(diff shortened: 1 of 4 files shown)


ipatests/conftest.py
```diff
@@ -83,6 +83,8 @@ def pytest_configure(config):
 
     # always run doc tests
     config.option.doctestmodules = True
+    # pytest.config global is deprecated
+    pytest.ipa_pytestconfig = config
```
tiran: This is too much of a hack.

stanislavlevin: I Just want to know, what are the disadvantages of attribute setting of third party module?
Anyway, I fixed as you said. Thank you!

tiran: IMHO the first approach felt like it's against the spirit of pytests' deprecation.

frasertweedale:
master:

* d16dd2fd6258520d83eb6350b8aebc9bfdbb11f3 Fix Pytest4.1+ warnings about pytest.config
* 9836511a2b6d7cf48b1a54cb3158e5eac674081a (HEAD) Fix Pytest4.x warning about `message`

PR #3288, opened 2019-06-18
dnsrecord-mod: allow to modify ttl without passing the record

flo-renaud:
### dnsrecord-mod: allow to modify ttl without passing the record
The command `ipa dnsrecord-mod <zone> <record> --ttl ` requires to provide at least one record to modify. When none is specified, it prompts by proposing each of the existing records, for instance:
```
$ ipa dnsrecord-mod ZZZZZ.org ns11 --ttl=86400
No option to modify specific record provided.
Current DNS record contents:

A record: xxx.xxx.xxx.xxx
AAAA record: xxxx:xx

Modify A record 'xxxx.xxxx.xxxx.xxxx'? Yes/No (default No):
Modify AAAA record 'xxxx:xx'? Yes/No (default No):
ipa: ERROR: No options to modify a specific record provided.
```
The admin should be able to modify the TTL value without re-entering the record information. The issue happens because of an internal check that forgot to consider 'dnsttl' as a valid standalone modification.

Fixes: https://pagure.io/freeipa/issue/7982

### XMLRPC tests: add new test for ipa dsnrecord-mod $ZONE $RECORD --ttl
    
The test suite did not have any test for modification of the TTL of an existing DNS record.
Related: https://pagure.io/freeipa/issue/7982

### DIFF
--- a/ipalib/dns.py
+++ b/ipalib/dns.py
@@ -55,7 +55,7 @@ def get_extra_rrtype(name):
 
 
 def has_cli_options(cmd, options, no_option_msg, allow_empty_attrs=False):
-    sufficient = ('setattr', 'addattr', 'delattr', 'rename')
+    sufficient = ('setattr', 'addattr', 'delattr', 'rename', 'dnsttl')
     if any(k in options for k in sufficient):
         return
 
--- a/ipatests/test_xmlrpc/test_dns_plugin.py
+++ b/ipatests/test_xmlrpc/test_dns_plugin.py
@@ -1948,6 +1948,34 @@ class test_dns(Declarative):
             },
         ),
 
+        dict(
+            desc='Modify ttl of record %r in zone %r' % (name1, zone1),
+            command=('dnsrecord_mod', [zone1, name1], {'dnsttl': 500}),
+            expected={
+                'value': name1_dnsname,
+                'summary': None,
+                'result': {
+                    'idnsname': [name1_dnsname],
+                    'dnsttl': ['500'],
+                    'arecord': [revname2_ip],
+                },
+            },
+        ),
+
+
+        dict(
+            desc='Delete ttl of record %r in zone %r' % (name1, zone1),
+            command=('dnsrecord_mod', [zone1, name1], {'dnsttl': None}),
+            expected={
+                'value': name1_dnsname,
+                'summary': None,
+                'result': {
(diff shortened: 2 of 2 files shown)


flo-renaud:
Rebased on master branch.

flo-renaud:
Re-started the azure pipeline. Test failed during installation because the install of ds created a temporary password starting with - and the next call to pwdhash <temp pwd> wrongly interprets the - as an option provided through CLI. This is a known issue in 389-ds, fixed with [commit](https://pagure.io/389-ds-base/c/223846dff976de076fdddc52f7adc55b432b0582) which prevents password_generate() from using - in the allowed characters.

rcritten:
An Azure XML-RPC test is failing due to a DNS lookup problem. Seems the host record doesn't have an A record. Completely unrelated code, not sure why it failed.

flo-renaud:
Adding re-run label as PRCI tests were not triggered.

rcritten:
The change works but it is quite a difference in behavior to go from prompting for all entries to only updating the A record. I'm not a heavy user of DNS so this seems ok to me, I think other opinions would be good too.

flo-renaud:
Hi @rcritten 
thanks for the review. My understanding is that the TTL value is shared by all the type of records of a given DNS record, as it is an attribute of the DNS record entry. For instance,
```
dn: idnsname=test,idnsname=domain.com.,cn=dns,dc=domain,dc=com
objectClass: top
objectClass: idnsrecord
aRecord: 10.37.170.195
idnsName: test
aAAARecord: 2620:52:0:25aa:21a:4aff:fe23:1311
dNSTTL: 86400
```
Hence the previously existing prompt did not make any sense (the value was changed for all the types of records even if only one was specified).

> it is quite a difference in behavior to go from prompting for all entries to only updating the A record

the PR doesn't change the behavior, the TTL is updated for the A record as well as the AAAA record if they both exist.

abbra:
LGTM.

flo-renaud:
master:

* bb91fcabee352fcaa71c8e66998a25e01cfaf9f7 dnsrecord-mod: allow to modify ttl without passing the record
* f25a7c2e962e3165e54a01479067ca9a900db6a2 XMLRPC tests: add new test for ipa dsnrecord-mod $ZONE $RECORD --ttl

PR #3289, opened 2019-06-18
stageuser-find: fix search with non-posix user

flo-renaud:
### stageuser-find: fix search with non-posix user
    
`ipa stageuser-find` fails to return a staged user if it does not contain the posixaccount objectclass.
The code is replacing the search filter (objectclass=posixaccount) with (|(objectclass=posixaccount)(objectclass=inetorgperson)) so it should work in theory.
The issue is that on python2 the filter has been hexlified before reaching the stageuser plugin, hence filter.replace does not recognize the pattern (objectclass=posixaccount).
The fix consists in creating the filter with a call to ldap.make_filter_from_attr() that will hexlify too, if needed.

Fixes: https://pagure.io/freeipa/issue/7983

### ipatests: add a test for stageuser-find with non-posix account

Add a new XMLRPC test with the following scenario:
- ldapadd a user without the posixaccount objectclass
- call ipa stageuser-find <user>
- check that 1 entry is returned

Related: https://pagure.io/freeipa/issue/7983

### DIFF
--- a/ipaserver/plugins/stageuser.py
+++ b/ipaserver/plugins/stageuser.py
@@ -449,7 +449,8 @@ class stageuser_find(baseuser_find):
         self.pre_common_callback(ldap, filter, attrs_list, base_dn, scope,
                                  *keys, **options)
 
-        container_filter = "(objectclass=posixaccount)"
+        container_filter = ldap.make_filter_from_attr(
+            'objectclass', 'posixaccount')
         # provisioning system can create non posixaccount stage user
         # but then they have to create inetOrgPerson stage user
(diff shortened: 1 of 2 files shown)


rcritten:
A very subtle change, good catch. Took me a while to get my head around what this was doing.

rcritten:
master:

* e9c4dcdb8532297f4247b930f74a8d2e00db43d9 stageuser-find: fix search with non-posix user
* 0294ad213388a208079ad29ef4f6ce1c47d8866d ipatests: add a test for stageuser-find with non-posix account

PR #3297, opened 2019-06-20
Log the raised message when DNS check_zone_overlap fails

rcritten:
The check can fail for a lot of other reasons than there is
overlap so the error should be logged.

This causes confusion when --auto-reverse is requested and
some lookup fails causing the reverse to not be created.


tiran:
master:

* 0184e967e58fb7663219eacffeae894031794af0 Log the raised message when DNS check_zone_overlap fails

PR #3299, opened 2019-06-21
Replace PYTHONSHEBANG with valid shebang

tiran:
Replace the @PYTHONSHEBANG@ substitution with a valid #!/usr/bin/python3
shebang. This turns Python .in files into valid Python files. The files
can now be checked with pylint and IDEs recognize the files as Python
files.

The shebang is still replaced with "#!$(PYTHON) -E" to support
platform-python.

Related: https://pagure.io/freeipa/issue/7984


tiran:
master:

* 6d02eddd3ef7f65c1a922125371fcbb83e35d7f8 Replace PYTHONSHEBANG with valid shebang

PR #3304, opened 2019-06-24
Make use of `named` well-known service

stanislavlevin:
The systemd unit name of `named`(which is actually used) is platform-dependent:
debian - bind9-pkcs11.service
fedora - named-pkcs11.service
redhat - named-pkcs11.service

Other systems may have their own name of `bind` service.
But the default one (`named-pkcs11`) is assumed in many tests.
Of course, these tests fail on such platforms.

This can be easily fixed.
All platforms define well-known service `named`, which is linked to
the actually utilized one.

Fixes: https://pagure.io/freeipa/issue/7990

### DIFF
--- a/ipatests/test_xmlrpc/test_location_plugin.py
+++ b/ipatests/test_xmlrpc/test_location_plugin.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import
 import pytest
 
 from ipalib import errors, api
+from ipaplatform.services import knownservices
 from ipatests.test_xmlrpc.tracker.location_plugin import LocationTracker
 from ipatests.test_xmlrpc.tracker.server_plugin import ServerTracker
 from ipatests.test_xmlrpc.xmlrpc_test import (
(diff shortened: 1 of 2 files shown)


ipatests/test_xmlrpc/test_location_plugin.py
```diff
@@ -126,11 +127,12 @@ def test_delete_location(self, location):
     not api.Command.dns_is_enabled()['result'], reason='DNS not configured')
 class TestLocationsServer(XMLRPC_test):
     messages = [{
-        u'data': {u'service': u'named-pkcs11.service',
+        u'data': {u'service': u'%s' % knownservices.named.systemd_name,
```
tiran: The ``%`` format operation is not necessary here:
``u'service': knownservices.named.systemd_name,``

stanislavlevin: Ah, yes, it is.
Pushed.

stanislavlevin:
@tiran Thank you for the detail review ( as usual ) :+1:  :smile:

rcritten:
master:

* 8f7d33356503794463611eb48317c9998187eddb Make use of `named` well-known service

PR #3310, opened 2019-06-24
ipatests: tests to verify that POSIX attributes, are no longer overwritten or missing

amore17:
Added test which validates that POSIX attributes, such
as shell or home directory, are no longer overwritten or missing.

Related Ticket : https://pagure.io/SSSD/sssd/issue/2474
Bugzilla : https://bugzilla.redhat.com/show_bug.cgi?id=1194345

### DIFF
--- a/ipatests/test_integration/test_trust.py
+++ b/ipatests/test_integration/test_trust.py
@@ -6,6 +6,7 @@ import re
 import unittest
 
 from ipaplatform.constants import constants as platformconstants
+from ipaplatform.paths import paths
 
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
@@ -47,7 +48,6 @@ class BaseTestTrust(IntegrationTest):
         cls.srv_gc_record_name = \
             '_ldap._tcp.Default-First-Site-Name._sites.gc._msdcs'
         cls.srv_gc_record_value = '0 100 389 {}.'.format(cls.master.hostname)
-
 
     @classmethod
     def check_sid_generation(cls):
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_trust.py
```diff
@@ -223,6 +224,37 @@ def test_user_without_posix_attributes_not_visible(self):
...
+            tasks.clear_sssd_cache(self.master)
+            self.master.run_command(['getent', 'initgroups', '-s', 'sss', testuser])
+
+            result = self.master.run_command(['getent', 'passwd', testuser])
+            assert "/home/testuser" in result.stdout_text
+            log_file = "/var/log/sssd/sssd_{0}.log" .format(self.master.domain.name)
+            sssd_debug_log = self.master.get_file_contents(
+                log_file, encoding='utf-8'
+            )
+            assert 'get_subdomain_homedir_of_user failed' not in sssd_debug_log
```
wladich: There is little problem in this check: the log file could contain this string BEFORE the testscase is executed. So we need to check that this line is not found in the part of log file which was written during execution of this testcase.
To do it you need to check the size of the log file before the test and then skip the initial part.
You could do it right in the testcase like
```
logsize = len(self.master.get_file_contents(log_file))
....
log = self.master.get_file_contents(log_file)[logsize:]
assert '...' not in log
```
but It would be nice to create utility in tasks.py, something like:
```
class LogFile(object):
    def __init__(self, host, path):
        self.host = host
        self.path = path
        self._initial_pos = len(host.get_file_contents(path))

    def new_text(self, encoding=None):
        s = self.host.get_file_contents(self.path)[self._initial_pos:]
        if encoding is not None:
            s = s.decode(encoding)
        return s
```
If you use this snippet without changes, please ask some else to review it.

wladich:
There are errors from pep8 and pylint:
```
./ipatests/test_integration/test_trust.py:245:80: E501 line too long (84 > 79 characters)
./ipatests/test_integration/test_trust.py:249:80: E501 line too long (84 > 79 characters)

./freeipa.spec:121: W: mixed-use-of-spaces-and-tabs (spaces: line 5, tab: line 121)
```

ipatests/test_integration/test_trust.py
```diff
@@ -223,6 +223,50 @@ def test_user_without_posix_attributes_not_visible(self):
...
+        including the home directory set in the AD LDAP and in sssd.conf
+        subdomain_homedir = %o is added after initgroup call home directory
+        should be correct and do not report in logs like,
+        'get_subdomain_homedir_of_user failed: * [Home directory is NULL]'
+        """
+        tasks.backup_file(self.master, paths.SSSD_CONF)
+        log_file = "/var/log/sssd/sssd_%s.log" % self.master.domain.name
+        sssd_debug_log1 = self.master.get_file_contents(
+            log_file, encoding='utf-8')
+        assert 'get_subdomain_homedir_of_user failed' not in sssd_debug_log1
```
wladich: The test should not fail if this string is found in sssd.log BEFORE the test. It can be there due to previous tests or any other events, not related to the test.
As I already proposed, instead you should check only part of the log which is written during test execution.

ipatests/test_integration/test_trust.py
```diff
@@ -223,6 +223,50 @@ def test_user_without_posix_attributes_not_visible(self):
...
+        Regression test for bug https://pagure.io/SSSD/sssd/issue/2474
+
+        When there is IPA-AD trust with POSIX attributes,
+        including the home directory set in the AD LDAP and in sssd.conf
+        subdomain_homedir = %o is added after initgroup call home directory
+        should be correct and do not report in logs like,
+        'get_subdomain_homedir_of_user failed: * [Home directory is NULL]'
+        """
+        tasks.backup_file(self.master, paths.SSSD_CONF)
+        log_file = "/var/log/sssd/sssd_%s.log" % self.master.domain.name
```
wladich: Please use constant paths.VAR_LOG_SSSD_DIR from

wladich:
@amore17 
test_override_homedir is green.
Now remove the temp commit, wait for standard check to pass and I will ACK the PR

PR #3312, opened 2019-06-25
adtrust upgrade: fix wrong primary principal name

abbra:
Upgrade code had Kerberos principal names mixed up: instead of creating
krbtgt/LOCAL-FLAT@REMOTE and marking LOCAL-FLAT$@REMOTE as an alias to
it, it created LOCAL-FLAT$@REMOTE Kerberos principal and marked
krbtgt/LOCAL-FLAT@REMOTE as an alias.

This differs from what Active Directory expects and what is created by
ipasam plugin when trust is established. When upgrading such deployment,
an upgrade code then unexpectedly failed.

Resolves: https://pagure.io/freeipa/issue/7992

### DIFF
--- a/daemons/ipa-sam/ipa_sam.c
+++ b/daemons/ipa-sam/ipa_sam.c
@@ -2039,11 +2039,13 @@ static bool handle_cross_realm_princs(struct ipasam_private *ipasam_state,
 							 pwd_outgoing, trusted_dn,
 							 KRB_PRINC_CREATE_DEFAULT);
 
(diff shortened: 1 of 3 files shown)


abbra:
It was my fault when documenting all these arrangements -- I made a wrong comment in `daemons/ipa-sam/ipa_sam.c` and then used it as a basis for the actual upgrade code.

tiran:
master:

* 34bfffd1bed4ecd11fbcfe54904e0ac7e3f609e7 adtrust upgrade: fix wrong primary principal name

PR #3316, opened 2019-06-26
Profile-based system cert renewal

frasertweedale:
https://pagure.io/freeipa/issue/7991

Early PR to see what CI has to say about it...

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -864,7 +864,7 @@ class CAInstance(DogtagInstance):
                 principal='host/%s' % self.fqdn,
                 subject=str(DN(('CN', 'IPA RA'), self.subject_base)),
                 ca=ipalib.constants.RENEWAL_CA_NAME,
-                profile='caServerCert',
+                profile=ipalib.constants.RA_AGENT_PROFILE,
                 pre_command='renew_ra_cert_pre',
                 post_command='renew_ra_cert',
                 storage="FILE",
@@ -991,7 +991,7 @@ class CAInstance(DogtagInstance):
                              '/org/fedorahosted/certmonger')
         iface = dbus.Interface(obj, 'org.fedorahosted.certmonger')
         for suffix in ['', '-reuse']:
-            name = 'dogtag-ipa-ca-renew-agent' + suffix
+            name = ipalib.constants.RENEWAL_CA_NAME + suffix
             path = iface.find_ca_by_nickname(name)
             if path:
                 iface.remove_known_ca(path)
@@ -1046,7 +1046,7 @@ class CAInstance(DogtagInstance):
         try:
             certmonger.start_tracking(
                 certpath=(paths.RA_AGENT_PEM, paths.RA_AGENT_KEY),
-                ca='dogtag-ipa-ca-renew-agent',
+                ca=ipalib.constants.RENEWAL_CA_NAME,
                 profile=ipalib.constants.RA_AGENT_PROFILE,
                 pre_command='renew_ra_cert_pre',
                 post_command='renew_ra_cert',
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -1047,6 +1047,7 @@ class CAInstance(DogtagInstance):
             certmonger.start_tracking(
                 certpath=(paths.RA_AGENT_PEM, paths.RA_AGENT_KEY),
                 ca='dogtag-ipa-ca-renew-agent',
+                profile=ipalib.constants.RA_AGENT_PROFILE,
                 pre_command='renew_ra_cert_pre',
                 post_command='renew_ra_cert',
                 storage='FILE')
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -267,6 +267,8 @@ class CAInstance(DogtagInstance):
        2 = have signed cert, continue installation
     """
 
+    server_cert_name = 'Server-Cert cert-pki-ca'
+
     # Mapping of nicknames for tracking requests, and the profile to
     # use for that certificate.  'configure_renewal()' reads this
     # dict.  The profile MUST be specified.
@@ -275,8 +277,12 @@ class CAInstance(DogtagInstance):
         'ocspSigningCert cert-pki-ca': 'caOCSPCert',
         'subsystemCert cert-pki-ca': 'caSubsystemCert',
         'caSigningCert cert-pki-ca': 'caCACert',
+        server_cert_name: 'caServerCert',
     }
-    server_cert_name = 'Server-Cert cert-pki-ca'
+    token_names = {
+        server_cert_name: 'internal',  # Server-Cert always on internal token
+    }
+
     # The following must be aligned with the RewriteRule defined in
     # install/share/ipa-pki-proxy.conf.template
     crl_rewrite_pattern = r"^\s*(RewriteRule\s+\^/ipa/crl/MasterCRL.bin\s.*)$"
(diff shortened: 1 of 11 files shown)


frasertweedale:
OK, I'm removing the WIP and asking for review now.  All the regular CI tests passed.
The temp commit runs some tests that perform ipa-server-upgrade in CA-ful and CA-less
deployments, and also the `ipatests.test_integration.test_installation.TestKRAinstallAfterCertRenew` test, which is the only upstream integration test I can see that does any renewal activity.  It is failing,  during installation (this test is not included in PR CI or nightly CI), so I'd say the failure is not related to my change, but I ran out of time to investigate before PTO.

frasertweedale:
@rcritten suggesting you as reviewer because you reviewed the earlier POC branch.  Feel free to pass the buck to anyone else though :)

ipaserver/install/cainstance.py
```diff
@@ -278,10 +278,12 @@ class CAInstance(DogtagInstance):
        2 = have signed cert, continue installation
     """
 
-    tracking_reqs = ('auditSigningCert cert-pki-ca',
-                     'ocspSigningCert cert-pki-ca',
-                     'subsystemCert cert-pki-ca',
-                     'caSigningCert cert-pki-ca')
+    tracking_reqs = {
```
rcritten: Would it be reasonable to add a comment line a description of the key/value pairs being the nickname and profile name?

ipaserver/install/krainstance.py
```diff
@@ -60,9 +60,11 @@ class KRAInstance(DogtagInstance):
     be the same for both the CA and KRA.
     """
 
-    tracking_reqs = ('auditSigningCert cert-pki-kra',
-                     'transportCert cert-pki-kra',
-                     'storageCert cert-pki-kra')
+    tracking_reqs = {
```
rcritten: Same, description of key/value pair might be handy.

ipaserver/install/server/upgrade.py
```diff
@@ -972,56 +972,35 @@ def certificate_renewal_update(ca, ds, http):
...
             'ca-name': 'dogtag-ipa-ca-renew-agent',
             'cert-presave-command': template % 'stop_pkicad',
             'cert-postsave-command':
-                (template % 'renew_ca_cert "Server-Cert cert-pki-ca"'),
-        },
+                (template % 'renew_ca_cert "{}"'.format(nick)),
+        }
+        profile = cainstance.CAInstance.tracking_reqs.get(nick)
+        if profile:
+            req['template-profile'] = profile
```
rcritten: The conditional is because for some certs (server-cert) we can fall back to the default right? Is this comment-worthy somewhere, perhaps block comment in method?

rcritten:
On upgrade the Apache and 389-ds certs didn't get a template_profile.

I suppose I should do an upstream certmonger release soon so profile is displayed on output?

rcritten:
This changes the CA certmonger tracking to add a post-command:

/usr/libexec/ipa/certmonger/renew_ca_cert "caSigningCert cert-pki-ca"

Is this intended? There is no post command currently for the signing cert.

ipaserver/install/server/upgrade.py
frasertweedale: The conditional was removed in a subsequent commit (profile is now always explicitly specified)

frasertweedale:
> On upgrade the Apache and 389-ds certs didn't get a template_profile.

That's fine, because they use the `IPA` CA, i.e. request certificates via `ipa cert-request` command.  They did not use renewal requests but "fresh enrolment" each time.

> I suppose I should do an upstream certmonger release soon so profile is displayed on output?

It's not critical for this feature, but yeah, the sooner the better IMO :)

frasertweedale:
> This changes the CA certmonger tracking to add a post-command:
> 
> /usr/libexec/ipa/certmonger/renew_ca_cert "caSigningCert cert-pki-ca"
> 
> Is this intended? There is no post command currently for the signing cert.

Hm, I don't think that's correct.  There should have been (and on all my installations, there is) a post-save command for the CA signing cert tracking request.

rcritten:
> Hm, I don't think that's correct. There should have been (and on all my installations, there is) a post-save command for the CA signing cert tracking request.

You're right, I'm not sure what I was seeing. This is ok.

rcritten:
I haven't narrowed down the cause yet but on upgrade the httpd cert loses its pinfile which makes it non-renewable. This may have been a pre-existing problem that wasn't exposed because of renewal by serial.

rcritten:
This fixed it for me. I pushed the date two years into the future and the certs renewed successfully:

```
diff --git a/ipaserver/install/httpinstance.py b/ipaserver/install/httpinstan
ce.py
index c3e7f4983..a0e888e3f 100644
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -585,11 +585,14 @@ class HTTPInstance(service.Service):
                          str(e))
 
     def start_tracking_certificates(self):
+        key_passwd_file = paths.HTTPD_PASSWD_FILE_FMT.format(host=api
.env.host)
         cert = x509.load_certificate_from_file(paths.HTTPD_CERT_FILE)
         if certs.is_ipa_issued_cert(api, cert):
             request_id = certmonger.start_tracking(
                 certpath=(paths.HTTPD_CERT_FILE, paths.HTTPD_KEY_FILE),
-                post_command='restart_httpd', storage='FILE'
+                post_command='restart_httpd', storage='FILE',
+                profile=dogtag.DEFAULT_PROFILE,
+                pinfile=key_passwd_file,
             )
             subject = str(DN(cert.subject))
             certmonger.add_principal(request_id, self.principal)

```

frasertweedale:
@rcritten thanks for the patch, I committed it (with you as author) and added to the PR.
AFAICT the bug already existed, but it is triggered by the tracking request remove/re-add on upgrade.

rcritten:
Kind of a nit I guess but installing without the patch results in the 389-ds cert showing the profile caIPAserviceCert. After the upgrade this is removed (because it is the default). All other certs have a profile except this one.

frasertweedale:
> Kind of a nit I guess but installing without the patch results in the 389-ds cert showing the profile caIPAserviceCert. After the upgrade this is removed (because it is the default). All other certs have a profile except this one.

Right, it seems to be another existing deficiency in the existing re-track code.  I'll fix it.

frasertweedale:
@rcritten I added a commit to ensure the profile gets added back to the DS cert tracking request.

frasertweedale:
@rcritten thanks for the ack!

frasertweedale:
master:

* 3c388f5a228b767dfd92bd824dfced166acda143 dogtaginstance: add profile to tracking requests
* f6f6f83dca22fb23ee2a7dd1b2925a74fe395afe upgrade: add profile to Dogtag tracking requests
* 858ef59948f5c8edd511f4db0b5fd69a1169f19b certmonger: use long options when invoking dogtag-ipa-renew-agent
* 1fb6fda01f2aca5fadc3707b7c7a3f3c9e647f85 dogtag-ipa-ca-renew-agent: always use profile-based renewal
* 588f1ddce2f69fd5e80d3271c9c8d80d312d350f dogtaginstance: avoid special cases for Server-Cert
* 4f4e2f96b022a056777fed59583f204edb797a2e upgrade: always add profile to tracking requests
* 482866e47e20520fbe39ef05439badbb3069bef6 upgrade: update KRA tracking requests
* 2d22f568a1043b0cb0b69314d3951d39721becc7 upgrade: log missing/misconfigured tracking requests
* fa5675582caef6d0708c8ae392e170b44d9daf5b upgrade: fix spurious certmonger re-tracking
* 1bf008a64f15e617b0dde93555f5dca1cbdf990a cainstance: add profile to IPA RA tracking request
* bb779baadf0df3333bd853599a21e81907c8c116 Use RENEWAL_CA_NAME and RA_AGENT_PROFILE constants
* 65d9a9be520ea9402143d9f06878eff53647925a ipatests: test ipa-server-upgrade in CA-less deployment
* f5822e3a25c0b1b35678ba3d8bcca6cbc16bff5b httpinstance: add pinfile when tracking certificate
* b7ad11572d3060e64252c4366d7c8afff1bc15e9 (HEAD) dsinstance: add proflie when tracking certificate

PR #3317, opened 2019-06-26
ipatest: check if /var/log is not present in error message of ipa backup

miskopo:
Pagure#6843 - Implemented test for ipa backup fail message mentioning log even in case
no log was created. After ipa-server package is installed, ipa backup is
executed without prior server-install. ipa backup fails and error
message should not contain mention of /var/log as no log was created.

### DIFF
--- a/ipatests/test_ipaclient/test_cli_ipa_not_configured.py
+++ b/ipatests/test_ipaclient/test_cli_ipa_not_configured.py
@@ -1,5 +1,9 @@
 import subprocess
 
+import pytest
+
+from ipapython.admintool import SERVER_NOT_CONFIGURED
+
 
 class TestIPANotConfigured:
     """
@@ -24,6 +28,7 @@ class TestIPANotConfigured:
         if unexpected_error:
             assert unexpected_error not in err
 
+    @pytest.mark.xfail(reason="Mismatch in return code, SERVER_NOT_CONFIGURED is 2, but ipa backup returns 1")
     def test_var_log_message_with_ipa_backup(self):
         """
         Test for BZ1428690: ipa-backup does not create log file at /var/log
@@ -31,4 +36,4 @@ class TestIPANotConfigured:
         As the server is not configured yet, command should fail and stderr should not
         contain link to /var/log, as no such log is created
         """
-        self.run_command("ipa backup", 1, None, "not configured on this system", "/var/log")
+        self.run_command("ipa backup", SERVER_NOT_CONFIGURED, None, "not configured on this system", "/var/log")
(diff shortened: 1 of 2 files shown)


flo-renaud:
Hi @miskopo 
There is a PEP8 error in the PR:
```
PEP-8 errors:
./ipatests/test_cmdline/test_help.py:191:1: W391 blank line at end of file
```
You can run `make fastlint` locally (please see [Testing#Fast_test](https://www.freeipa.org/page/Testing#Fast_test) for instructions)

ipatests/test_cmdline/test_help.py
```diff
@@ -174,3 +174,18 @@ def test_multiline_description():
...
+
+
+def test_bz1428690():
+    """
+    Test for BZ#1428690 - ipa-backup does not create log file at /var/log/
+    :return: None
+    :raises: AssertionError if the test fails
+    """
+    with CLITestContext(exception=SystemExit) as ctx:
+        api.Backend.cli.run(['backup'])
```
flo-renaud: The command should be `ipa-backup`.

ipatests/test_cmdline/test_help.py
```diff
@@ -174,3 +174,18 @@ def test_multiline_description():
         api.Backend.cli.run(['trust-add', '-h'])
 
     assert unicode(api.Command.trust_add.doc).strip() in help_ctx.stdout
+
+
+def test_bz1428690():
+    """
+    Test for BZ#1428690 - ipa-backup does not create log file at /var/log/
+    :return: None
```
flo-renaud: Please add a self-explanatory comment, for instance:
When ipa-backup is executed on a node where IPA is not configured, the error message should not refer to logs at /var/log/ because the log file has not even been created. The command should print that IPA is not configured.

flo-renaud:
Adding ipa-4-6 and ipa-4-7 labels as the issue was fixed in ipa-4-6 branch, hence the test makes sense in these branches, too.

ipatests/test_cmdline/test_help.py
abbra: And it means you cannot use here `api.Backend.cli.run()` because that function is only for running methods registered from plugins loaded into the client. The full list of methods is available in `api.Backend.Command`. `ipa-backup` is a separate script that is only available on the server side.

miskopo:
> Hi @miskopo
> There is a PEP8 error in the PR:
> 
> ```
> PEP-8 errors:
> ./ipatests/test_cmdline/test_help.py:191:1: W391 blank line at end of file
> ```
> 
> You can run `make fastlint` locally (please see [Testing#Fast_test](https://www.freeipa.org/page/Testing#Fast_test) for instructions)

Flo, I tested with PEP8 and hexdump and there is no blank line (\n\n) at the end of my file. There is only one \n, which comforts W292

PR #3326, opened 2019-06-27
adtrust upgrade: fix wrong primary principal name, part 2

abbra:
Second part of the trust principals upgrade

For existing LOCAL-FLAT$@REMOTE object, convert it to
krbtgt/LOCAL-FLAT@REMOTE and add LOCAL-FLAT$@REMOTE as an alias. To do
so we need to modify an entry content a bit so it is better to remove
the old entry and create a new one instead of renaming.

Resolves: https://pagure.io/freeipa/issue/7992

### DIFF
--- a/ipaserver/install/plugins/adtrust.py
+++ b/ipaserver/install/plugins/adtrust.py
@@ -512,16 +512,19 @@ class update_tdo_to_new_layout(Updater):
 
         if isinstance(principals, (list, tuple)):
             trust_principal = principals[0]
-            aliases = principals[1:]
+            alias = principals[1]
         else:
             trust_principal = principals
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @abbra 
thanks for the PR, LGTM. Tested the following scenarii:
- [x] install 4.7.2, etup trust, upgrade: OK
- [x] install with patch, setup trust: OK
In both cases I can see 4 krbprincipalname entries below the trust entry, with expected content.

abbra:
master:

* 7af4c7d4720227b5bff037a128061bf616e27096 adtrust upgrade: fix wrong primary principal name, part 2

PR #3331, opened 2019-06-27
Make use of single configuration point for SELinux

stanislavlevin:
For now, FreeIPA supports SELinux things as they are in RedHat/Fedora.
But different distributions may have their own SELinux customizations.
    
This moves SELinux configuration out to platform constants:
    - SELINUX_MCS_MAX
    - SELINUX_MCS_REGEX
    - SELINUX_MLS_MAX
    - SELINUX_MLS_REGEX
    - SELINUX_USER_REGEX
    - SELINUX_USERMAP_DEFAULT
    - SELINUX_USERMAP_ORDER
    
and applies corresponding changes to the test code.

Fixes: https://pagure.io/freeipa/issue/7996

### DIFF
--- a/install/share/bootstrap-template.ldif
+++ b/install/share/bootstrap-template.ldif
@@ -426,8 +426,8 @@ ipaDefaultEmailDomain: $DOMAIN
 ipaMigrationEnabled: FALSE
 ipaConfigString: AllowNThash
 ipaConfigString: KDC:Disable Last Success
-ipaSELinuxUserMapOrder: guest_u:s0$$xguest_u:s0$$user_u:s0$$staff_u:s0-s0:c0.c1023$$sysadm_u:s0-s0:c0.c1023$$unconfined_u:s0-s0:c0.c1023
-ipaSELinuxUserMapDefault: unconfined_u:s0-s0:c0.c1023
+ipaSELinuxUserMapOrder: $SELINUX_USERMAP_ORDER
+ipaSELinuxUserMapDefault: $SELINUX_USERMAP_DEFAULT
 
 dn: cn=cosTemplates,cn=accounts,$SUFFIX
 changetype: add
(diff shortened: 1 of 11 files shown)


stanislavlevin:
Unfortunately, I don't know - why is SELinux user name regex  `^[a-zA-Z][a-zA-Z_\.]*$` (doesn't include digits). I can create one on Fedora30:
```
semanage user -a test_user111 --roles user_r

semanage user -l | grep test_user111
test_user111    user       s0         s0                             user_r
```

stanislavlevin:
For instance, I set pattern `^[a-zA-Z][a-zA-Z0-9_\.]*$` for ALTLinux.

@rcritten , do you remember? Why are there no digits?

rcritten:
I don't remember. I do recall it very difficult to find authoritative sources on what was allowed in the various components of a SELinux name. It could be that numbers weren't allowed when I originally implemented this (2011) and is now. I'm not opposed to expanding the regex if numerics are indeed legal now.

abbra:
@stanislavlevin please rebase. The change looks good to me and I'd like to get it in for 4.8.0.

stanislavlevin:
@abbra, thanks! Rebased.

stanislavlevin:
@rcritten, current implementation requires nothing (set name as is)
https://github.com/SELinuxProject/selinux/blob/3abeb7c654ed0c8205cc5e666236f1705af32972/libsepol/src/user_record.c#L104

```
semanage user -a '$#@*!&^()/\|#$@.~#$' --roles user_r

semanage user -l
...
$#@*!&^()/\|#$@.~#$ user       s0         s0                             user_r
...
```

abbra:
LGTM. I think we want to have this in master only (e.g. FreeIPA 4.8.0)

abbra:
master:

* 215e8f768c47261d78043dcea23add3da12f364e Fix a typo in `replace` rule of 50-ipaconfig.update
* b2acd65013348c0f5d91582f240714b0f3357678 Make use of single configuration point for SELinux

rcritten:
@stanislavlevin for the record if you want to expand the user regex I'm ok with that. Just may have to exclude $ since IIRC that is what I use as a separator. I don't know that we've ever gotten push-back against the current regex though.

PR #3335, opened 2019-06-28
Makefile.am: add .in files to fastlint target

fcami:
Previously fastlint would only process .py files.
Detect which .in files are in fact Python files and add
them to the list of files to process during the fastlint
Makefile target.

Fixes: https://pagure.io/freeipa/issue/7984

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -219,8 +219,19 @@ endif
 	@echo "Fast linting with $(PYTHON) from branch '$(GIT_BRANCH)'"
 
 	@MERGEBASE=$$(git merge-base --fork-point $(GIT_BRANCH)); \
-	FILES=$$(git diff --name-only --diff-filter=d $${MERGEBASE} \
-	    | grep -E '\.py$$'); \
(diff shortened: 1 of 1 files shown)


rcritten:
Maybe my tree isn't entirely clean I'm not sure but running make fastcheck after importing the patch fails:

************* Module ipatests/test_xmlrpc/test_stageuser_plugin.py\nclient/ipa-certupdate.in
ipatests/test_xmlrpc/test_stageuser_plugin.py\nclient/ipa-certupdate.in:1: [F0001(fatal), ] No module named ipatests/test_xmlrpc/test_stageuser_plugin.py\nclient/ipa-certupdate.in)

fcami:
hi @rcritten good catch - this should be fixed now, thanks.

abbra:
LGTM.

abbra:
master:

* 6b2efdfae5302427543ef42686c35db737da6261 Makefile.am: add .in files to fastlint target

PR #3337, opened 2019-06-28
[Backport][ipa-4-6] trust-fetch-domains: make sure we use right KDC when --server is spec…

abbra:
…ified

Since we are authenticating against AD DC before talking to it (by using
trusted domain object's credentials), we need to override krb5.conf
configuration in case --server option is specified.

The context is a helper which is launched out of process with the help
of oddjobd. The helper takes existing trusted domain object, uses its
credentials to authenticate and then runs LSA RPC calls against that
trusted domain's domain controller. Previous code directed Samba
bindings to use the correct domain controller. However, if a DC visible
to MIT Kerberos is not reachable, we would not be able to obtain TGT and
the whole process will fail.

trust_add.execute() was calling out to the D-Bus helper without passing
the options (e.g. --server) so there was no chance to get that option
visible by the oddjob helper.

Also we need to make errors in the oddjob helper more visible to
error_log. Thus, move error reporting for a normal communication up from
the exception catching.

Resolves: https://pagure.io/freeipa/issue/7895

### DIFF
--- a/install/oddjob/com.redhat.idm.trust-fetch-domains
+++ b/install/oddjob/com.redhat.idm.trust-fetch-domains
@@ -8,9 +8,12 @@ from ipapython.dn import DN
 from ipapython.dnsutil import DNSName
 from ipaplatform.constants import constants
 from ipaplatform.paths import paths
+import io
 import sys
 import os
 import pwd
+import tempfile
+import textwrap
 
 import six
 import gssapi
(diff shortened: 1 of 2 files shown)


flo-renaud:
ACk since it's a simple backport. `ipa trust-add` tested with/without `--server` option and with/without `--two-way=true`

flo-renaud:
ipa-4-6:

* 076cb69f00db734a378a74f50f5b4b74321c0fa4 trust-fetch-domains: make sure we use right KDC when --server is specified

PR #2983, opened 2019-04-02
Add user trust attributes plugin

Tiboris:
Add user trust attributes plugin
https://github.com/abbra/freeipa-user-trust-attributes
Author: @abbra Alexander Bokovoy <abokovoy@redhat.com>

PR #3339, opened 2019-06-29
SMB user attributes

abbra:
This is a variation of the pull request #2983, now that SMB service support is merged. I fixed a number of issues there and spent some time on rethinking how this set of attributes should be called.
I added an experimental code to hide SMB user attributes if there is no `ipaNTUserAttrs` object class in the list of object classes for the user account. You cannot create them without adding the object class and you cannot add this object class without adding `ipaNTSecurityIdentifier`. Right now `ipaNTSecurityIdentifier` added on user creation if `ipa-adtrust-install` was run at least on one IPA master. Alternatively, it is added with `ipa-adtrust-install --add-sids` run.


abbra:
I got CLI and Web UI working now. SMB attributes will not be accessible in `user-add` due to the way how SID generation works in IPA: `ipaNTSecurityIdentifier` attribute is only added when:
* object has POSIX attributes `uidNumber` and `gidNumber`
* the values of those attributes are within 32-bit unsigned integer
* object has any of the following object classes: `ipaIDObject`, `posixAccount`, `posixGroup`
* has no `ipaNTSecurityIdentifier` already

On the other hand, all of SMB attributes require the presence of `ipaNTUserAttrs` object class in the user object LDAP entry. This object class cannot be added without `ipaNTSecurityIdentifier`. We do not want to add it manually as it has to consume values from a range suitable for SIDs and this logic is in `sidgen` plugin. We cannot add `ipaNTSecurityIdentifier` with a "magic" value either, as this is not supported by the `sidgen` plugin.

So, it is better to have the use of these attributes moved to the `ipa user-mod` command instead.

Web UI part now looks like this: when `ipaNTUserAttrs` is present in the user's object classes.

When a user has no `ipaNTUserAttrs`, web UI looks like this:. I wanted to make this section completely invisible but `show_cond` doesn't work for sections, only for fields inside section, it seems.

Next, I'm going to extend the design document for domain controller with these details.

abbra:
Somehow, Azure CI is not working and its tests are missing. Since they duplicate xmlrpc tests ran at Travis CI, all tests were completed.

abbra:
@Tiboris please review this pull request.

abbra:
tests passed.

Tiboris:
master:

* 339771b0d87e787a496c7d193edf5c9c4a50195d Remove unreachable code
* c18ee9b641ddc1e6b52d0413caa1fb98ac13785d Add SMB attributes for users

PR #3343, opened 2019-06-30
Make dnf more robust and faster

fcami:
Sometimes the prepare-build step of azure pipelines fails with download errors:
"configure: error: Package requirements (nspr) were not met:"
This can be due to fastestmirror not being used to check
mirror availability and sometimes speed. Combined with a
too-low default number of retries, and a high timeout this
can lead to download failures that could be avoided.
Activate fastestmirror, add more download workers, and tune
timeout/retries to make dnf more reliable.
    
Fixes: https://pagure.io/freeipa/issue/7999

### DIFF
--- a/ipatests/azure/templates/prepare-build.yml
+++ b/ipatests/azure/templates/prepare-build.yml
@@ -2,6 +2,13 @@ steps:
 - script: |
     set -e
     sudo rm -rf /var/cache/dnf/*
+    echo "dnf.conf: enable fastestmirror, use 8 download workers, lower timeout to fail faster, and add more retries"
+    sudo tee -a /etc/dnf/dnf.conf <<EOF > /dev/null
+    fastestmirror = True
+    max_parallel_downloads = 8
+    timeout = 8
+    retries = 20
+    EOF
     echo 'Disable modular repositories'
     sudo dnf config-manager '*modular*' --set-disabled
     sudo dnf makecache || :


fcami:
It looks like dnf.conf cannot be written to, so it might belong to the container host.
Another avenue to avoid the download failures we see is to create fedora repositories pointing to a subset of known hosts close to our running location.

ipatests/azure/templates/prepare-build.yml
```diff
@@ -2,6 +2,14 @@ steps:
 - script: |
     set -e
     sudo rm -rf /var/cache/dnf/*
+    echo "dnf.conf: enable fastestmirror"
+    sudo echo "fastestmirror = True" >> /etc/dnf/dnf.conf
```
abbra: This line runs echo under sudo and then tries to redirect its output under the user the outer script runs on.

abbra: I'd rather added this as a config in `ipatests/azure/` and then copied it over with `sudo cp .. ..`

tiran: A file ``/etc/dnf/vars/fastestmirror`` with content ``enabled`` should do the trick, too.

fcami:
With the change in syntax I get:
```
determining the fastest mirror (132 hosts).. done.
```
so fastestmirror is enabled.

fcami:
@abbra @tiran I've switched to appending a heredoc to /etc/dnf/dnf.conf with 'tee -a'.
According to the prepare-build logs it works.

tiran:
master:

* 7027f7914e32bcf0942931f8915fe602e27dd7aa Make dnf more robust and faster

PR #3349, opened 2019-07-01
ipatests: filter_users should be applied correctly if SSSD starts offline

amore17:
Added regression tests which validates that filter_users is applied correctly
when SSSD starts in offline mode, which checks that no look up
should be in data provider.

Related Tickets:
https://pagure.io/SSSD/sssd/issue/3983
https://pagure.io/SSSD/sssd/issue/3978

### DIFF
--- a/ipatests/test_integration/test_sssd.py
+++ b/ipatests/test_integration/test_sssd.py
@@ -13,14 +13,11 @@ import pytest
 
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
+from ipaplatform.osinfo import osinfo
 from ipaplatform.paths import paths
 
 
-class TestSSSDAuthCache(IntegrationTest):
-    """Regression tests for cached_auth_timeout option
-
-       https://bugzilla.redhat.com/show_bug.cgi?id=1685581
-   """
+class TestSSSDWithAdTrust(IntegrationTest):
 
     topology = 'star'
     num_ad_domains = 1
@@ -34,6 +31,9 @@ class TestSSSDAuthCache(IntegrationTest):
             'name_tmpl': 'testuser@{domain}',
             'password': 'Secret123'
         },
+        'fakeuser': {
+            'name': 'some_user@some.domain'
+        },
     }
     ipa_user = 'user1'
     ipa_user_password = 'SecretUser1'
@@ -43,7 +43,7 @@ class TestSSSDAuthCache(IntegrationTest):
 
     @classmethod
     def install(cls, mh):
-        super(TestSSSDAuthCache, cls).install(mh)
+        super(TestSSSDWithAdTrust, cls).install(mh)
 
         cls.ad = cls.ads[0]  # pylint: disable=no-member
 
@@ -53,6 +53,9 @@ class TestSSSDAuthCache(IntegrationTest):
 
         cls.users['ad']['name'] = cls.users['ad']['name_tmpl'].format(
             domain=cls.ad.domain.name)
+
+        # Regression tests for cached_auth_timeout option
+        # https://bugzilla.redhat.com/show_bug.cgi?id=1685581
         tasks.user_add(cls.master, cls.intermed_user)
         tasks.create_active_user(cls.master, cls.ipa_user,
                                  cls.ipa_user_password)
@@ -108,3 +111,40 @@ class TestSSSDAuthCache(IntegrationTest):
             assert self.is_auth_cached(self.users[user])
             time.sleep(10)
             assert not self.is_auth_cached(self.users[user])
+
+    @contextmanager
+    def filter_user_setup(self, user):
+        sssd_conf_backup = tasks.FileBackup(self.master, paths.SSSD_CONF)
+        filter_user = {'filter_users': self.users[user]['name']}
+        try:
+            tasks.modify_sssd_conf(self.master, self.master.domain.name,
+                                   filter_user)
+            tasks.clear_sssd_cache(self.master)
+            yield
+        finally:
+            sssd_conf_backup.restore()
+            tasks.clear_sssd_cache(self.master)
+
+    @pytest.mark.xfail(
+        osinfo.id == 'fedora' and osinfo.version_number <= (28,),
+        reason='https://pagure.io/SSSD/sssd/issue/3978')
+    @pytest.mark.parametrize('user', ['ad', 'fakeuser'])
+    def test_is_user_filtered(self, user):
+        """No lookup in data provider from 'filter_users' config option.
+
+        Test for https://bugzilla.redhat.com/show_bug.cgi?id=1685472
+        https://bugzilla.redhat.com/show_bug.cgi?id=1724088
+
+        When there are users in filter_users in domain section then no look
+        up should be in data provider.
+        """
+        with self.filter_user_setup(user=user):
+            log_file = '{0}/sssd_nss.log'.format(paths.VAR_LOG_SSSD_DIR)
+            logsize = tasks.get_logsize(self.master, log_file)
+            self.master.run_command(
+                ['getent', 'passwd', self.users[user]['name']],
+                ok_returncode=2)
+            sssd_log = self.master.get_file_contents(log_file)[logsize:]
+            dp_req = ("Looking up [{0}] in data provider".format(
+                self.users[user]['name']))
+            assert not dp_req.encode() in sssd_log
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -1969,3 +1969,9 @@ def remote_ini_file(host, filename):
 def is_selinux_enabled(host):
     res = host.run_command('selinuxenabled', ok_returncode=(0, 1))
     return res.returncode == 0
+
+
+def get_logsize(host, logfile):
+    """ get current logsize"""
+    logsize = len(host.get_file_contents(logfile))
+    return logsize


ipatests/test_integration/test_trust.py
```diff
@@ -176,6 +176,70 @@ def test_upn_user_authentication_in_nonposix_trust(self):
...
+
+        When there are users in filter_users for nss provider then no
+        look up should be in data provider and NCE/USER/ipa_domain/user
+        should be added to negative cache.
+        """
+
+        aduser = 'testuser@%s' % self.ad_domain
+        fakeuser = 'some_user@some.domain'
+
+        exp_txt1 = 'Looking up [%s] in data provider' % fakeuser
```
celestian: Please, could we be unit on using format() function?

netoarmando:
@amore17 Please, re-base your base to pick recent changes from master (including new version of `ci-master-f30` image).

netoarmando:
Error during the provisioning of `ad-root` host:
```
Jul 08 22:19:58 ad-runner-1 bash[26264]: DEBUG: root: PLAY [ad-root] *****************************************************************
Jul 08 22:19:58 ad-runner-1 bash[26264]: DEBUG: root:
Jul 08 22:19:58 ad-runner-1 bash[26264]: DEBUG: root: TASK [set administrator password] **********************************************
Jul 08 22:20:20 ad-runner-1 bash[26264]: DEBUG: root: An exception occurred during task execution. To see the full traceback, use -vvv. The error was: InvalidCredentialsError: the specified credentials were rejected by the server
Jul 08 22:20:20 ad-runner-1 bash[26264]: DEBUG: root: fatal: [ad-root]: FAILED! => {"msg": "Unexpected failure during module execution.", "stdout": ""}
Jul 08 22:20:20 ad-runner-1 bash[26264]: DEBUG: root: ...ignoring
Jul 08 22:20:20 ad-runner-1 bash[26264]: DEBUG: root:
Jul 08 22:20:20 ad-runner-1 bash[26264]: DEBUG: root: msg: Unexpected failure during module execution.
Jul 08 22:20:20 ad-runner-1 bash[26264]: DEBUG: root:
```
It needs further investigation.

amore17:
Successful Run : http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/770bd998-b290-11e9-8559-525400471dd3/report.html

ipatests/test_integration/test_trust.py
```diff
@@ -177,6 +177,73 @@ def test_upn_user_authentication_in_nonposix_trust(self):
...
+
+        Regression test for https://pagure.io/SSSD/sssd/issue/3983
+        https://pagure.io/SSSD/sssd/issue/3978
+
+        When there are users in filter_users for nss provider then no
+        look up should be in data provider and NCE/USER/ipa_domain/user
+        should be added to negative cache.
+        """
+        tasks.kinit_admin(self.master)
+        aduser = 'testuser@{0}' .format(self.ad_domain)
```
wladich: Please remove the space before .format

ipatests/test_integration/test_trust.py
```diff
@@ -177,6 +177,73 @@ def test_upn_user_authentication_in_nonposix_trust(self):
...
+        fakeuser = 'some_user@some.domain'
+
+        exp_txt1 = 'Looking up [{0}] in data provider' .format(fakeuser)
```
wladich: Please remove the space before .format

ipatests/test_integration/test_trust.py
```diff
@@ -177,6 +160,73 @@ def test_upn_user_authentication_in_nonposix_trust(self):
...
+        exp_txt2 = 'Adding [NCE/USER/{0}/@{1}] to negative cache permanently'\
```
wladich: Please use  braces instead of backslash.
Refer to PEP-8 style guide: https://www.python.org/dev/peps/pep-0008/

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

ipatests/test_integration/test_trust.py
```diff
@@ -177,6 +160,73 @@ def test_upn_user_authentication_in_nonposix_trust(self):
...
+            .format(self.master.domain.name, aduser)
+        exp_txt3 = 'NCE/USER/{0}/{1}] to negative cache permanently'\
```
wladich: Same as above

ipatests/test_integration/test_trust.py
```diff
@@ -177,6 +160,73 @@ def test_upn_user_authentication_in_nonposix_trust(self):
...
+                   .format(self.ad_domain, aduser)
+
+        resolution = '{0}:{1}'.format(self.ad_domain, self.master.domain.name)
```
wladich: variable name is misleading. It holds not resolution, but simply domains. So name like domains is better.

ipatests/test_integration/test_trust.py
```diff
@@ -177,6 +160,102 @@ def test_upn_user_authentication_in_nonposix_trust(self):
...
+        logsize = tasks.get_logsize(self.master, log_file)
+        try:
+            tasks.kinit_admin(self.master)
+            user1 = 'some_user@some.domain'
+            aduser = 'testuser@{0}'.format(self.ad_domain)
+            domain = self.master.domain
+            tasks.modify_sssd_conf(
+                self.master, domain.name,
+                {
+                    'filter_users': '{0}'.format(user1, aduser)
```
netoarmando: Azure's Build and Lint phases are complaining about this.

ipatests/test_integration/test_trust.py
```diff
@@ -177,6 +161,108 @@ def test_upn_user_authentication_in_nonposix_trust(self):
...
+            self.master.run_command(['ipa', 'config-mod',
+                                     '--domain-resolution-order={0}'
+                                    .format(domains)])
+            user1 = 'some_user@some.domain'
+            aduser = 'testuser@{0}'.format(self.ad_domain)
+            domain = self.master.domain
+            tasks.modify_sssd_conf(
+                self.master, domain.name,
+                {
+                    'filter_users': '{0},{1}'.format(user1, aduser)
```
wladich: according to `man sssd.conf` the `filter_users` option should be in [nss] section, not in the domain one

wladich: I was wrong, it can be used in domain section.

amore17: Adding in  [nss] is broader, [domain/...] would only filter for the given domain

wladich:
First thing I think we should do is move those testcases to https://github.com/freeipa/freeipa/blob/master/ipatests/test_integration/test_sssd.py. There are two reasons for that: 1) they are related more to sssd than to trust 2) test_sssd is executed in PR-CI as it requires only one AD machine.
This will require slight changes in test_sssd.py:
 * rename the class to something like TestSSSDWithAdTrust
 * replace the class docstring with simple comment (it will describe the following tests, ones which already exists there)

ipatests/test_integration/test_trust.py
```diff
@@ -175,6 +220,114 @@ def test_upn_user_authentication_in_nonposix_trust(self):
...
+        """Negative cache should use values from 'filter_users' config option.
+
+        Regression test for https://pagure.io/SSSD/sssd/issue/3978
+
+        When there are users in filter_users for nss provider then no
+        look up should be in data provider and added in Negative cache
+        if sssd starts when ipa is offline and domain order to prefer AD.
+        """
+        domains = '{0}:{1}'.format(self.ad_domain, self.master.domain.name)
+        tasks.set_domain_resolution(self.master, order=domains)
```
wladich: Are you sure we need to set domain resolution order in those tests? I do not see it being mentioned in related BZ. And we are using principal names here.

ipatests/test_integration/test_trust.py
```diff
@@ -175,6 +220,114 @@ def test_upn_user_authentication_in_nonposix_trust(self):
...
+        try:
+            aduser = 'testuser@{0}'.format(self.ad_domain)
+            self.check_filter_users(self.ad_domain, aduser,
+                                    provider='nss', offline=True)
```
wladich: Please explain what changes do you expect in sssd.conf with provider='nss'? And also please explain why you decided to add this testcase, it was not mentioned in related BZ.

ipatests/test_integration/test_trust.py
```diff
@@ -90,6 +92,49 @@ def remove_trust(self, ad):
         tasks.remove_trust_with_ad(self.master, ad.domain.name)
         tasks.clear_sssd_cache(self.master)
 
+    def check_filter_users(self, realm, aduser, provider=None, offline=False):
+        master = self.master
+        user1 = 'some_user@some.domain'
+        tasks.backup_file(self.master, paths.SSSD_CONF)
+        filter_users = {'filter_users': '{0},{1}'.format(user1, aduser)}
+        log_file = '{0}/sssd_nss.log'.format(paths.VAR_LOG_SSSD_DIR)
+        if provider:
```
wladich: modify_sssd_conf has default value 'ipa'
So instead of writing these 4 lines  you can simply do:
```
def check_filter_users(...., provider='ipa',....):
    ....
    tasks.modify_sssd_conf(master, master.domain.name, filter_users, provider)
    ....
```

ipatests/test_integration/test_trust.py
```diff
@@ -90,6 +92,49 @@ def remove_trust(self, ad):
...
+            tasks.modify_sssd_conf(master, master.domain.name, filter_users,
+                                   provider)
+        else:
+            tasks.modify_sssd_conf(master, master.domain.name, filter_users)
+        logsize = tasks.get_logsize(master, log_file)
+        if offline:
+            master.run_command(['ipactl', 'stop'])
+        else:
```
wladich: "else" branch is redundant as ipa server is expected to be running here.

ipatests/test_integration/test_trust.py
```diff
@@ -90,6 +92,49 @@ def remove_trust(self, ad):
         tasks.remove_trust_with_ad(self.master, ad.domain.name)
         tasks.clear_sssd_cache(self.master)
 
+    def check_filter_users(self, realm, aduser, provider=None, offline=False):
+        master = self.master
+        user1 = 'some_user@some.domain'
+        tasks.backup_file(self.master, paths.SSSD_CONF)
```
wladich: It is better to use tasks.FileBackup for backup restore since tasks.restore_files restores all files that have been backuped during the test run, which can include the files modified in class setup (usually done in install()).

tasks.FileBackup  on the other hand preserves copy of exactly one file and also preserves selinux context.

wladich:
I have read the bugzillas thoroughly and it seems to me they describe following test cases:
https://bugzilla.redhat.com/show_bug.cgi?id=1685472
TC1
 * setup trust with AD_DOMAIN
 * filter_users=user@AD_DOMAIN
 * start sssd with ipa offile and clean cache
 * start ipa
 * wait for sssctl domain-list
 * run id user@AD_DOMAIN -> not found
 * in sssd logs there is message about adding to negative cache
 * in sssd logs there is no message  "Looking up [user@AD_DOMAIN] in data provider"
TC2
    same as TC1, but for unknown domain 


https://bugzilla.redhat.com/show_bug.cgi?id=1679173
According to BZ description:
TC3
 * setup trust with AD_DOMAIN
 * set domain resolution order=AD_DOMAIN
 * do not change sssd.conf, which implies filter_users=root
 * call 'id root'
 * in sssd logs there is no message "Looking up [root@AD_DOMAIN] in data provider"
 
 But later in BZ1679173 the verification steps are changed to exactly what is described in TC1. So I am confused, why do we use same scenario to verify different bugs?

amore17:
@wladich Thanks for the review,
In TC3
According to [steps](https://bugzilla.redhat.com/show_bug.cgi?id=1679173#c6) given by jakub from ipa perspective it is tested by checking request is added in negative cache.

wladich:
On the test scenario: the mentioned bug https://bugzilla.redhat.com/show_bug.cgi?id=1685472 does not need stopping/starting ipa server for reproducing.

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +114,57 @@ def test_auth_cache_enabled_when_configured(self, user):
...
             assert not self.is_auth_cached(self.users[user])
+
+    @contextmanager
+    def filter_user_setup(self, user, provider='ipa'):
+        tasks.kinit_admin(self.master)
+        master = self.master
+        sssd_conf_backup = tasks.FileBackup(master, paths.SSSD_CONF)
+        if user['name'] == 'fakeuser':
+            provider = 'nss'
+        filter = {'filter_users': '{0}'.format(user['name'])}
```
wladich: `filter` is a python builtin function, do not override it

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +114,57 @@ def test_auth_cache_enabled_when_configured(self, user):
...
+            assert test_id.returncode != 0
+        finally:
+            sssd_log = self.master.get_file_contents(log_file)[logsize:]
+
+        dp_req = ("Looking up [{0}] in data provider".format(user['name']))
+        return dp_req.encode() in sssd_log
+
+    @pytest.mark.xfail(
+        osinfo.id == 'fedora' and osinfo.version_number <= (30,),
+        reason='https://pagure.io/SSSD/sssd/issue/3978', strict=False)
```
wladich: Please do not use xfail with strict=False unless you have a serious reason for that.
Doing so you simply ignore test result, so that we shall not notice if it is finally fixed in Fedora 30 and older

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +114,57 @@ def test_auth_cache_enabled_when_configured(self, user):
             assert self.is_auth_cached(self.users[user])
             time.sleep(10)
             assert not self.is_auth_cached(self.users[user])
+
+    @contextmanager
+    def filter_user_setup(self, user, provider='ipa'):
```
wladich: You are not passing value for `provider` when calling the function, so no need to sepcify it

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +114,57 @@ def test_auth_cache_enabled_when_configured(self, user):
...
+        tasks.kinit_admin(self.master)
+        master = self.master
+        sssd_conf_backup = tasks.FileBackup(master, paths.SSSD_CONF)
+        if user['name'] == 'fakeuser':
```
wladich: "fakeuser" is a value passed to `test_filter_users_start_sssd_offline()`

user['name'] is never equal to 'fakeuser'

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +114,57 @@ def test_auth_cache_enabled_when_configured(self, user):
...
+        should be in data provider.
+        1: domain section for AD user
+        2: for other domain added user in nss section.
+        """
+        with self.filter_user_setup(user=self.users[user]):
+            tasks.clear_sssd_cache(self.master)
+            list_domains = ['sssctl', 'domain-list']
+            test_list = re.escape(self.ad.domain.name)
+            tasks.run_repeatedly(self.master, list_domains, timeout=60,
+                                 test=lambda x: re.search(test_list, x))
```
wladich: what is the reason for those actions in current test scenario?

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +114,57 @@ def test_auth_cache_enabled_when_configured(self, user):
...
+        filter = {'filter_users': '{0}'.format(user['name'])}
+        try:
+            tasks.modify_sssd_conf(
+                master, master.domain.name, filter, provider)
+            yield
+        finally:
+            sssd_conf_backup.restore()
+            tasks.clear_sssd_cache(self.master)
+
+    def is_user_filtered(self, user):
```
wladich: This function should be eliminated and its contents should be moved to test method, based on following:
 * it is used only in one place
 * it contains whole test scenario except setup
 * test method contains only invocation of setup and this function

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +114,57 @@ def test_auth_cache_enabled_when_configured(self, user):
...
+        log_file = '{0}/sssd_nss.log'.format(paths.VAR_LOG_SSSD_DIR)
+        logsize = tasks.get_logsize(self.master, log_file)
+        try:
+            self.master.run_command(['getent', 'passwd', user['name']],
+                                    raiseonerr=False)
+            test_id = self.master.run_command(['id', user['name']],
+                                              raiseonerr=False)
+            assert test_id.returncode != 0
```
wladich: 1) Why call both `id` and `getent`?
2) As we are sure the command should return "not found",  we can specify expected return code in `ok_returncode` parameter.

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +114,57 @@ def test_auth_cache_enabled_when_configured(self, user):
...
+        finally:
```
wladich: if previous check have failed (i.e. `id` or `getent` returned success), we do not need to contonue the test, so try-except is not needed here

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +113,45 @@ def test_auth_cache_enabled_when_configured(self, user):
...
+        should be in data provider.
+        1: domain section for AD user
+        2: for other domain added user in nss section.
+        """
+        with self.filter_user_setup(user=user):
+            log_file = '{0}/sssd_nss.log'.format(paths.VAR_LOG_SSSD_DIR)
+            logsize = tasks.get_logsize(self.master, log_file)
+            self.master.run_command(
+                ['getent', 'passwd', self.users[user]['name']],
+                raiseonerr=False, ok_returncode=[0, 2])
```
wladich: 1. It is a mistake to use both `raiseonerr` and `ok_returncode`. With `raiseonerr=False` the function will not raise exception with any return code.
2. We expect that user is in negative cache, so there is only one valid return code -- 2

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +113,45 @@ def test_auth_cache_enabled_when_configured(self, user):
             assert self.is_auth_cached(self.users[user])
             time.sleep(10)
             assert not self.is_auth_cached(self.users[user])
+
+    @contextmanager
+    def filter_user_setup(self, user):
+        sssd_conf_backup = tasks.FileBackup(self.master, paths.SSSD_CONF)
+        filter_user = {'filter_users': '{0}'.format(self.users[user]['name'])}
```
wladich: `'{0}'.format(self.users[user]['name'])` can be replaced with `self.users[user]['name']`

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +113,45 @@ def test_auth_cache_enabled_when_configured(self, user):
...
+                provider = 'ipa'
+            tasks.modify_sssd_conf(self.master, self.master.domain.name,
+                                   filter_user, provider=provider)
+            tasks.clear_sssd_cache(self.master)
+            yield
+        finally:
+            sssd_conf_backup.restore()
+
+    @pytest.mark.xfail(
+        osinfo.id == 'fedora' and osinfo.version_number <= (30,),
```
wladich: The related fix is in sssd-2.2.1
Fedora 29 and up have sssd >= 2.2.2

The test should execute successfully on all fedoras >= 29.

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +113,45 @@ def test_auth_cache_enabled_when_configured(self, user):
             assert self.is_auth_cached(self.users[user])
             time.sleep(10)
             assert not self.is_auth_cached(self.users[user])
+
+    @contextmanager
+    def filter_user_setup(self, user):
```
wladich: Do you think setup should be in separate function. It looks rather simple and can be incorporated in test function.

amore17: yes i think it should be in separate function, there will be need of this function where we want to test user without domain.

ipatests/test_integration/test_sssd.py
```diff
@@ -108,3 +111,45 @@ def test_auth_cache_enabled_when_configured(self, user):
...
+            if user == 'fakeuser':
+                provider = 'nss'
+            else:
+                provider = 'ipa'
+            tasks.modify_sssd_conf(self.master, self.master.domain.name,
+                                   filter_user, provider=provider)
+            tasks.clear_sssd_cache(self.master)
+            yield
+        finally:
+            sssd_conf_backup.restore()
```
wladich: you need to restart sssd to apply changes to config file

wladich:
On previous comment: I used wrong version which was partially fixed. 
So I re-ran the test wit older one (1.16.2) and have following observations:
* Both testcases are failing in expected way (good)
* "some_user@some.domain" is added to domain section, not to nss one (bad)

Previously in my experiment I failed to reproduce the issue with unknown domain in domain section of sssd.conf which made me think that we need to put `filter_users` in nss section. Now I see that most probably I had incomplete logs.
Did you try to reproduce the issue yourself adding `filter_user=some_user@some.domain` in domain section of sssd.conf? If that is a correct reproducer, then we do not need even try to specify nss section in test setup.

amore17:
@wladich As per your last comment tested scenario by adding filter_user=some_user@some.domain in domain section of sssd.conf
And yes in that way also we i can reproduce  
So removing nss section from test setup.
[test_with_domain.txt](https://github.com/freeipa/freeipa/files/3938662/test_with_domain.txt)

wladich:
LGTM. Please remove temp commit and I will ACK the PR when checks are green

flo-renaud:
master:

* 0162f3aafd63c07f100ee8146a39dab05acbae7d ipatests: filter_users should be applied correctly.

flo-renaud:
@amore17 a manual backport is required to ipa-4-6

PR #3350, opened 2019-07-01
Use only TLS 1.2 by default

tiran:
TLS 1.3 is causing some trouble with client cert authentication.
Conditional client cert authentication requires post-handshake
authentication extension on TLS 1.3. The new feature is not fully
implemented yet.

TLS 1.0 and 1.1 are no longer state of the art and now disabled by
default.

TLS 1.2 works everywhere and supports PFS.

Related: pagure.io/freeipa/issue/7667

### DIFF
--- a/ipalib/config.py
+++ b/ipalib/config.py
@@ -44,7 +44,7 @@ from ipalib.base import check_name
 from ipalib.constants import (
     CONFIG_SECTION,
     OVERRIDE_ERROR, SET_ERROR, DEL_ERROR,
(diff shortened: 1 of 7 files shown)


tiran:
master:

* c484d79ecfa1cc284b47b88377a4c2da23b9db2f For Fedora and RHEL use system-wide crypto policy for mod_ssl
* b57c818fab3bb9627a8c287766cdb5bd8071c837 Use only TLS 1.2 by default

PR #3356, opened 2019-07-01
ipatests: fix ipatests/test_xmlrpc/test_dns_plugin.py

flo-renaud:
The test is calling dnsrecord-mod --ttl and should expect a unicode
value in order to be python2/python3 compatible.

Related: https://pagure.io/freeipa/issue/7982


abbra:
master:

* 71884176478917aa75a4277e580ed785f818b57d ipatests: fix ipatests/test_xmlrpc/test_dns_plugin.py

PR #3357, opened 2019-07-01
Don't configure disabled krb5 enctypes in FIPS mode

rcritten:
The only permitted ciphers are the AES family (called aes, which
is the combination of: aes256-cts-hmac-sha1-96,
aes128-cts-hmac-sha1-96, aes256-cts-hmac-sha384-192, and
aes128-cts-hmac-sha256-128).

DES, RC4, and Camellia are not permitted in FIPS mode.  While 3DES
is permitted, the KDF used for it in krb5 is not, and Microsoft
doesn't implement 3DES anyway.

This is only applied on new installations because we don't
allow converting a non-FIPS install into a FIPS one.


abbra:
Looks good. Since we have SMB test in the gating set now, I wonder if we can simply disable all outdated ciphers (except camelia in non-FIPS) completely and try PR CI? If test_smb test would pass, that means we don't really need to enable all those.

rcritten:
Pushed a second patch on to completely remove 3DES and RC4. We'll see what happens.

frozencemetery:
Well, I'm excited by that second commit; hopefully it doesn't break any trust stuff.  (3DES should be fine; it's gone in Fedora anyway.)

abbra:
@frozencemetery to have it fully working in FIPS mode, I suspect we'd need your help in krb5 to allow creating `krb5_c_string_to_key()` to accept arcfour-hmac (`krb5_c_string_to_key()` will use `EVP_md4()` in openssl backend in FIPS mode and will be blocked from using it).

I'm not sure krb5 in Fedora has this enablement for non crypto use.

frozencemetery:
It doesn't, but I can make it appear, provided the test build worked.  (If this is hard to check, let me know - I can just apply to rawhide and revert if needed.)

abbra:
master:

* a43100badc43a2b22efd472ff4108b9821631311 Don't configure disabled krb5 enctypes in FIPS mode
* dd9fd0971f352e4bed50b79b8b3147d6a94956b7 Remove DES3 and RC4 enctypes from Kerberos
* 258cacb11607426364e92ae17726e7582af5c3d7 Add test_smb to night Fedora 30 test suite

PR #3358, opened 2019-07-01
Added new authentication indicators

tengcm2015:
Added default indicators for SPAKE, PKINIT and FAST. Also added those as default options in webUI

Resolves: https://pagure.io/freeipa/issue/8001
Signed-off-by: Changmin Teng <cteng@redhat.com>


abbra:
@tengcm2015 thanks for the pull request. Could you please open a ticket at https://pagure.io/freeipa/issues and add the following to your commit messages:
```
Resolves: https://pagure/freeipa/issue/<number>
Signed-off-by: ...
```

install/ui/src/freeipa/serverconfig.js
```diff
@@ -85,7 +85,10 @@ return {
                                 { label: '@i18n:authtype.type_disabled', value: 'disabled' },
                                 { label: '@i18n:authtype.type_password', value: 'password' },
                                 { label: '@i18n:authtype.type_radius', value: 'radius' },
-                                { label: '@i18n:authtype.type_otp', value: 'otp' }
+                                { label: '@i18n:authtype.type_otp', value: 'otp' },
+                                { label: '@i18n:authtype.type_pkinit', value: 'pkinit' },
+                                { label: '@i18n:authtype.type_spake', value: 'spake' },
+                                { label: '@i18n:authtype.type_fast', value: 'fast' }
```
abbra: You are introducing here i18n labels for each new type but don't add them in `ipaserver/plugins/internal.py`. Please add them there.

install/ui/src/freeipa/host.js
```diff
@@ -130,6 +130,18 @@ return {
...
+                                    label: 'pkinit',
+                                    value: 'pkinit'
+                                },
+                                {
+                                    label: 'spake',
+                                    value: 'spake'
+                                },
+                                {
+                                    label: 'fast',
+                                    value: 'fast'
```
abbra: And here all the labels should be changed to the proper ones as `@i18n:authtype.type_<method>` for each method, even for the original ones.

frozencemetery:
We had previously talked about how to handle `encrypted_challenge_indicator` which I'll briefly attempt to summarize here.  Per [kdc.conf(5)](https://manpages.debian.org/unstable/krb5-kdc/kdc.conf.5.en.html), the value of `encrypted_challenge_indicator` is attached only when encrypted challenge preauth is performed along a FAST channel.  (So the name makes sense.)  However, there's a subtlety here: it's possible to perform FAST *without* encrypted challenge (e.g., by using SPAKE).  This means that referring to the indicator as "FAST" is potentially confusing.

@tengcm2015, I think we'd discussed four possible outcomes here:

1. Treat FAST and SPAKE the same - that is, give them the same indicator value.
2. Document that this is purely encrypted challenge FAST (probably by naming it something else).
3. Don't expose an indicator for FAST.
4. Decide that the problem is intractable, and call it FAST anyway.

My impression was that you were planning to do (3), but you appear to have done (4).  Could you elaborate on the rationale for that?

tengcm2015:
@frozencemetery Actually I didn't know what exactly was happening behind this behavior until now... My thinking was that by default any preauth mechanism was allowed, so let admins decide how to deal with FAST problems, but this is actually more complicated than I thought... 

Now I start to think that 1. might be the way to go, since I can't think of a reason to force clients not to use SPAKE while using FAST. Still trying to come up with a name for this hybrid solution though.

tengcm2015:
> I have one big issue with this patch: it tries to modify web UI to provide values which will not be accepted by the IPA server side. The reason for that is this part of the server configuration code in `ipaserver/plugins/config.py`:
> 
> ```
>         StrEnum('ipauserauthtype*',
>             cli_name='user_auth_type',
>             label=_('Default user authentication types'),
>             doc=_('Default types of supported user authentication'),
>             values=(u'password', u'radius', u'otp', u'disabled'),
>         ),
> ```
> 
> Only values specified in the parameter definition will be accepted. Same for `ipaserver/plugins/baseuser.py` where per-user parameter is defiend for the same attribute.
> 
> Additionally, ipa KDB driver doesn't have support to recognize these new values too. That looks like not a big issue but at least we need to verify that.

@abbra I'm not very familiar with IPA server internal, but can we add new auth-indicator vlue there?

frozencemetery:
The two failing jobs bailed with message "/bin/sh: rpmbuild: command not found" - that seems like it's CI infrastructure having trouble?

abbra:
@tengcm2015 @frozencemetery if you see failures like that, rebase and push again. Sometimes there are issues accessing Fedora Project repositories from both Azure and Travis clouds. We pushed recently a fix that will attempt to choose the closest mirror. Hope this will make it more reliable.

abbra:
> @abbra I'm not very familiar with IPA server internal, but can we add new auth-indicator value there?
"there" -- in KDB code? Depends on what we expect to do with it from the code. I have a task this and next week to investigate what we can do in the new KDC policy plugin and I'll look at this too.

For IPA framework, see the code I referenced. Extending the list of supported mechanisms there will formally break our API (the call stays the same, the older values are supported but old IPA masters will reject the changes), so we need to think what to do here, too.
Until that analysis is done, I don't think we can proceed with this pull request.

tengcm2015:
Update: Added design document for kerberos ticket policy

abbra:
@tengcm2015 thanks for the design document.

So, we have to add following changes to proceed with this pull request:
- modify `StrEnum('ipauserauthtype*')` for user object to add `reinforced` auth indicator:
```
       StrEnum('ipauserauthtype*',
            cli_name='user_auth_type',
            label=_('User authentication types'),
            doc=_('Types of supported user authentication'),
            values=(u'password', u'radius', u'otp'),
        ),
```
- convert `Str('krbprincipalauthind*')` for service object to the `StrEnum()` similar to the above because we should have the same values in both. Use of `Str()` instead of `StrEnum()` was wrong in past and needs to be fixed
```
        Str('krbprincipalauthind*',
            cli_name='auth_ind',
            label=_('Authentication Indicators'),
            doc=_("Defines a whitelist for Authentication Indicators."
                  " Use 'otp' to allow OTP-based 2FA authentications."
                  " Use 'radius' to allow RADIUS-based 2FA authentications."
                  " Other values may be used for custom configurations."),
        ),
```
- add tests that ensure that new values are supported. See `ipatests/test_xmlrpc/test_service_plugin.py` and `ipatests/test_xmlrpc/test_user_plugin.py` for existing tests touching the same parameters, they would need to be updated.

Make sure to run `makeapi` after the changes so that `API.txt` is updated. You'd also need to change the minor version of API in `VERSION.m4`. See, for example, how it is done with 5a44ca638310913ab6b0c239374f4b0ddeeedeb3 which introduced `Certificate()` param instead of using `Bytes()`.

abbra:
Thanks, @tengcm2015. I added 're-run' label to make sure PRCI will run whole battery of tests.
I think this is ready for merge once tests pass.
To make clear: it only implements adding new auth indicators, KDC policies for lifetime policy and lifetime jitter will come in separate pull requests, right?

tengcm2015:
@abbra Yes, this PR only adds new auth indicators.
Actually, there is still a problem with this PR: for user authentication policy, even though we can set and modify the auth method policy for each user, the KDC is not enforcing that policy for pkinit and reinforced method. Please give me some time to figure out what to do with this.

simo5:
So looking with fresh eyes at the code this morning I found out that we may be able to actually perform authentication policy in the KDC already.

In MIT code the finish_process_as_req() will call check_kdcpolicy_as() passing in the request and a bunch of auth_indicators retrieved from the client principal.

What matters to us is not much the indicators that we can retrieve on our own in the KDB anyway (although it comes handy we do not have to), but that this function is pluggable and krb5_kdc_req is being passed in.

So if we add in the KDC support for implementing a kdcpolicy plugin and register it, then we can implement krb5_kdcpolicy_check_as_fn(), in there we can parse (again, sigh) the padata with the preauthentication data and figure out what preauth type the client sent, perhaps modify it if there are things we do not want to allow, or return an error if the preauth mechanism presented do not match the ones we restricted the user to.

It will require a bit of experimentation and checking, but it is the most promising way to perform proper preauth type restriction per user, and should help us remove the hacks we have currently to force non-password auth.

simo5:
Note, error reporting to clients may not be great as usual ... however it seem we can return a string with the status in the function, so there is hope this string may actually find it's way back to the user ...

simo5:
Aaand it seem we already have a skeleton policy plugin PR: https://github.com/freeipa/freeipa/pull/2147

frozencemetery:
krb5 has parsed the pa_data already; we just have to inspect the `const krb5_kdc_req *request` parameter.  We intentionally can't modify the incoming request structure, and the outgoing request isn't exposed.  So the only recourse is to deny the request if it's not approved.

Which raises another question: how do we know if it's approved?  Does that end up being another LDAP request?  Unfortunate, but I don't see a way around that even with a hypothetical KDB interface.

Finally, assuming we go the route of using kdcpolicy to enforce that, how do you want the code organized?  Should we rebase this work to include the policy features as well, or should that be a follow-on PR?  (We've already been working on features with kdcpolicy as part of the rest of the project proposal.)

simo5:
@frozencemetery not sure why would we need a new LDAP request ?
you have the list of allowed auth indicators for the user (I think ?) and the preauth methods, should be just a matching ?

Also we can preload additional data if need in the client pricinpal as in client->e_data we have a pointer back to oru internal client principal representation, so the kdcpolicy plugin should have all the needed data already and just need to make a decision, no additional lookups

The problem is that we may need a way to tell the KDC to disregard a preauth method we do not want to allow if multiple preauth mehotds are sent by the client? Though that is probably unlikley, and perhaps we can just return error if a client sends multiple one and one of them is unacceptable.

frozencemetery:
I don't believe I have the list of allowed auth indicators for a user.  That's an IPA-ism, which is why we're having this problem :)  We need to get the list from IPA somehow - reading it out of `e_data` sounds better, if we can do that.

I can look into providing such a hook if it becomes important - probably we'd just bump the kdcpolicy plugin minor version and add support for it, if we end up needing that.

simo5:
I do not think we need any hook, client is already passed into the callback, so client->e_data will already be available ...

tengcm2015:
I ended up implementing user pre-auth enforcement with kdcpolicy plugin, and pulling out user pre-auth policy set via IPA from LDAP.

simo5:
I see you pulled in Robbie's PR, maybe we should just depend on it and let Robbie complete it?
I also do not see additional commits on top of it to implement policy in the new plugin.

frozencemetery:
>I see you pulled in Robbie's PR, maybe we should just depend on it and let Robbie complete it?

No worries, I asked him to do that.  The last review that #2147 received was back in April and I addressed it; I'm happy to do other things as needed, but the impression I had gotten was that there wasn't interest in merging it without actual policy using it, hence this work.

>I also do not see additional commits on top of it to implement policy in the new plugin.

Top commit in the series looks like the policy in question to me, since it's modifying `ipa_kdcpolicy_check_as()`.

tengcm2015:
Yes, the top commit in the series is the one. Looks like GitHub is showing commits ordered by time created instead of actual commit order.

abbra:
@tengcm2015 thanks for the new code. However, one existing test failed: test_xmlrpc/test_kerberos_principal_aliases.py:test_authenticate_with_user_alias

If you'd sign to Azure with your github account,  https://dev.azure.com/freeipa/freeipa/_build/results?buildId=830&view=ms.vss-test-web.build-test-results-tab&runId=5456&resultId=101098&paneView=debug should show the details. On 'Summary' tab you can retrieve all logs (this was in BASE and XMLRPC tests 3).

simo5:
> Yes, the top commit in the series is the one. Looks like GitHub is showing commits ordered by time created instead of actual commit order.

I always forget how annoying is github's interface in this regard...

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +57,112 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+        goto done;
+    }
+
+    /* Filter out anonymous PKINIT principal since it doesn't have a LDAP entry */
+    if (strncmp(principal, "WELLKNOWN/ANONYMOUS@WELLKNOWN:ANONYMOUS", 40) == 0) {
+        goto done;
+    }
+
+    /* Get LDAP entry for the principal */
+    kerr = ipadb_fetch_principals(ipactx, KRB5_KDB_FLAG_ALIAS_OK, principal, &result);
```
simo5: before going down the path of searching LDAP once more which will considerably slow down the KDC you should check if request->client->e_data does not already contain a struct ipadb_e_data
If you need more data in there please change the default fetching function and store it in that structure so it is readily available without extra searches.

Note that e_Data may not always be available as sometimes the KDC may be constructing the client data instead of getting it from a KDC search, but it is worth checking if e_data is there because it will speed up the operation considerably.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +57,112 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+    if (*auth_indicators == NULL || strlen(*auth_indicators) == 0) {
+        if (!(ua & IPADB_USER_AUTH_PASSWORD)) {
+            kerr = KRB5KDC_ERR_POLICY;
+            *status = "Reinforced pre-authentication not allowed for this user.";
+        }
+        goto done;
+    }
+
+    /* For each auth indicator, see if it is allowed for that user */
+    if (strstr(*auth_indicators, "otp") != NULL) {
```
simo5: This is not ok, it will match an arbitrary authenticator named "nototp" as well, you should tokenize first and then check individual tokens in their entirety.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +57,112 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+    if (kerr != 0) {
+        krb5_klog_syslog(LOG_ERR, "IPA kdcpolicy: ipadb_find_principal failed.");
+        goto done;
+    }
+
+    /* Get User Auth configuration. */
+    ua = ipadb_get_user_auth(ipactx, lentry);
+
+    /* Make sure string *auth_indicators has a length of 
+     * at least 1 so that strstr() won't broke */
```
simo5: broke -> break

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +57,112 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+    if (*auth_indicators == NULL || strlen(*auth_indicators) == 0) {
+        if (!(ua & IPADB_USER_AUTH_PASSWORD)) {
+            kerr = KRB5KDC_ERR_POLICY;
+            *status = "Reinforced pre-authentication not allowed for this user.";
```
simo5: NOt sure this message is clear, as OTP is not considered reinforced ...

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +57,112 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+            goto done;
+        }
+    }
+
+    if (strstr(*auth_indicators, "reinforced") != NULL) {
+        valid_indicators_set++;
+        /* allow SPAKE/FAST even if only password is allowed */
+        if (!(ua & (IPADB_USER_AUTH_PASSWORD | IPADB_USER_AUTH_REINFORCED))) {
+            kerr = KRB5KDC_ERR_POLICY;
+            *status = "Password authentication not allowed for this user.";
```
simo5: Ah I guess you mixed this error status with the one on top?
Also why do you check for IPADB_USER_AUTH_PASSWORD as well?
We may want to be  forcing the user to use only SPAKE but not pre-auth timestamp, and this check prevents that from working afaics.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +57,112 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+            goto done;
+        }
+    }
+
+    if (!(ua & IPADB_USER_AUTH_PASSWORD) && !valid_indicators_set) {
+        kerr = KRB5KDC_ERR_POLICY;
+        *status = "Non-hardened password authentication not allowed for this user.";
```
simo5: I do not understand this, why is it not ok to have no indicators ?
So far absence of indicators always meant we allow any auth, we just do not enforce any specific pre-auth, are we changing that? (Upgrades may have issues with this)

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -63,8 +175,6 @@ ipa_kdcpolicy_check_tgs(krb5_context context, krb5_kdcpolicy_moddata moddata,
                         krb5_deltat *renew_lifetime_out)
 {
     *status = NULL;
-    *lifetime_out = 0;
-    *renew_lifetime_out = 0;
```
simo5: why removing these ?

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +57,112 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+        krb5_klog_syslog(LOG_ERR, "IPA kdcpolicy: ipadb_fetch_principals failed.");
+        goto done;
+    }
+
+    kerr = ipadb_find_principal(context, KRB5_KDB_FLAG_ALIAS_OK, result, &principal, &lentry);
+    if (kerr != 0) {
+        krb5_klog_syslog(LOG_ERR, "IPA kdcpolicy: ipadb_find_principal failed.");
+        goto done;
+    }
+
```
simo5: everything above here is equivalent to ipadb_get_principal(), why not use that ?

daemons/ipa-kdb/ipa_kdb_principals.c
```diff
@@ -327,7 +327,7 @@ static void ipadb_validate_password(struct ipadb_context *ipactx,
         *ua |= IPADB_USER_AUTH_PASSWORD;
 }
 
-static enum ipadb_user_auth ipadb_get_user_auth(struct ipadb_context *ipactx,
+enum ipadb_user_auth ipadb_get_user_auth(struct ipadb_context *ipactx,
```
simo5: you won't have to change this if we always unconditionally add this data to struct ipadb_e_data and use ipadb_get_principal() as mentioned above

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
tengcm2015: Put user_auth flags in e_data and using that now

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
tengcm2015: Just reallized auth_indicators is an array and fixed accordingly

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
tengcm2015: In ipadb_get_user_auth() only IPADB_USER_AUTH_PASSWORD flag is set when no flag was set in LDAP entry. But yes this is confusing so I made ipadb_get_user_auth() set all flags and do the validation there.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
tengcm2015: I was thinking that this might be confusing since we are not changing lifteme and renewtime right now.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
tengcm2015: agreed

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +45,81 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+    int valid_auth_indicators = 0;
+
     *status = NULL;
-    *lifetime_out = 0;
-    *renew_lifetime_out = 0;
 
     krb5_klog_syslog(LOG_INFO, "IPA kdcpolicy: checking AS-REQ.");
 
-    return 0;
+    ied = (struct ipadb_e_data *)client->e_data;
```
simo5: I think client->e_data may be NULL and the if below may have to check it before dereferencing.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +45,81 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+                kerr = KRB5KDC_ERR_POLICY;
+                *status = "OTP pre-authentication not allowed for this user.";
+                goto done;
+            }
+
+        } else if (strcmp(*auth_indicators, "radius") == 0) {
+            valid_auth_indicators++;
+            if (!(ua & IPADB_USER_AUTH_RADIUS)) {
+                kerr = KRB5KDC_ERR_POLICY;
+                *status = "RADIUS pre-authentication not allowed for this user.";
```
simo5: Radius is an internal detail of the way OTP can be configured, I think we should still return "OTP pre-authentication ..." as the status error that is sent to the client ? Or is this error only sent to logs ?

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +45,81 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+                *status = "Reinforced pre-authentication not not allowed for this user.";
+                goto done;
+            }
+        }
+    }
+
+    if (!valid_auth_indicators) {
+        if (!(ua & IPADB_USER_AUTH_PASSWORD)) {
+            kerr = KRB5KDC_ERR_POLICY;
+            *status = "Non-hardened password authentication not allowed for this user.";
```
simo5: You are using hardened here, but reinforced above, should we just agree on one term? (Hardened sounds better than reinforced to me, although both are somewhat vague).

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
simo5: In krb5 sometimes functions need to explicitly clear out passin arguments, so I would first check in the code if these are actually required, might be uninitialized in the KDC calling function...

daemons/ipa-kdb/ipa_kdb_principals.c
```diff
@@ -351,10 +342,19 @@ static enum ipadb_user_auth ipadb_get_user_auth(struct ipadb_context *ipactx,
...
+    /* If no mechanisms are set, allow every auth method for now. */
+    if (ua == IPADB_USER_AUTH_NONE) {
+        ua = IPADB_USER_AUTH_PASSWORD
+           | IPADB_USER_AUTH_RADIUS
+           | IPADB_USER_AUTH_OTP
+           | IPADB_USER_AUTH_PKINIT
+           | IPADB_USER_AUTH_REINFORCED
+           ;
+    }
+
```
simo5: This is prone to error on following changes, I think we have two options here:
1) do not do this and simply return kerr=0 if ua == IPADB_USER_AUTH_NONE in ipa_kdcpolicy_check_as() befoire we even go through auth_indicators array and save some computing time (preferred).
2) add a macro IPADB_USER_AUTH_ALL in the headers where all other IPADB_USER_AUTH_* defines are so that if someone adds a new one they will see it is needed in there and not need to check all places in the code to figure out there is one where all mechanisms need to be explicitly listed.

I favor (2) less because it is still error prone and it will always be the case that if no indicators are set we accept any pre-auth mechanism so we should just bail out of policy checking right away, and not cycle through each and every one we defined (also it will be faster that way).

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
frozencemetery: I'm fine with either.  We need something that reflects "password-based, brute-force resistant" (but is much shorter than that); those are the best we've come up with.

daemons/ipa-kdb/ipa_kdb_principals.c
frozencemetery: Can you clarify your preference?  You state that you favor 2, but 1 is preferred?

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
tengcm2015: I didn't see the message in krb5 packet returned to client, so I guess this is only showing up in the logs.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
frozencemetery: They are actually required.  I've posted a PR to change that [to krb5](https://github.com/krb5/krb5/pull/967), but it probably should stay for compat.

daemons/ipa-kdb/ipa_kdb_principals.c
tengcm2015: Went for (1)

simo5: err sorry I meant I disfavor (2) ... sigh... language, thanks @tengcm2015 for reading my mind :-)

abbra:
@tengcm2015 Please check failed tests:
* http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ca049de2-bc22-11e9-a611-fa163eae1a35/report.html
* http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/9e87ba00-bc22-11e9-9dc5-fa163e7e8887/report.html
Use the same URLs without `report.html` to get access to actual logs and details.

tengcm2015:
@abbra The reason for [this test failure](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ca049de2-bc22-11e9-a611-fa163eae1a35/report.html) is that in current state, when password is allowed but not reinforced, SPAKE hardened password will get rejected (which is enabled by default). Do you think we should also allow "reinforced" pre-auth even if only password is allowed?

frozencemetery:
Yes, we should - it's strictly better.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
tengcm2015: Decided to use "hardened" anyway since the word has a wider range of use in descriptions etc. Switched to this term all over the PR.

tengcm2015:
So the test is failing on DNS, which I'm not sure if is caused by this PR

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +45,90 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+
     *status = NULL;
     *lifetime_out = 0;
     *renew_lifetime_out = 0;
 
     krb5_klog_syslog(LOG_INFO, "IPA kdcpolicy: checking AS-REQ.");
 
-    return 0;
+    ied = (struct ipadb_e_data *)client->e_data;
+    if ((ied == NULL) || (ied->magic != IPA_E_DATA_MAGIC)) {
```
frozencemetery: No parentheses.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +45,90 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+        /* e-data is not availble, getting user auth from LDAP */
+        krb5_klog_syslog(LOG_INFO, "IPA kdcpolicy: client e_data not availble. Try fetching...");
+        kerr = ipadb_get_principal(context, request->client, KRB5_KDB_FLAG_ALIAS_OK, &client);
+        if (kerr != 0) {
+            krb5_klog_syslog(LOG_ERR, "IPA kdcpolicy: ipadb_find_principal failed.");
+            goto done;
+        }
+
+        ied = (struct ipadb_e_data *)client->e_data;
+        if ((ied == NULL) && (ied->magic != IPA_E_DATA_MAGIC)) {
```
frozencemetery: Here either.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +45,90 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+        const char *auth_indicator = auth_indicators[i];
+
+        if (strcmp(*auth_indicators, "otp") == 0) {
+            valid_auth_indicators++;
+            if (!(ua & IPADB_USER_AUTH_OTP)) {
+                kerr = KRB5KDC_ERR_POLICY;
+                *status = "OTP pre-authentication not allowed for this user.";
+                goto done;
+            }
+
```
frozencemetery: Whitespace.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +45,90 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+        } else if (strcmp(*auth_indicators, "radius") == 0) {
+            valid_auth_indicators++;
+            if (!(ua & IPADB_USER_AUTH_RADIUS)) {
+                kerr = KRB5KDC_ERR_POLICY;
+                *status = "OTP pre-authentication not allowed for this user.";
+                goto done;
+            }
+
```
frozencemetery: Extra whitespace.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +45,90 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+        } else if (strcmp(*auth_indicators, "pkinit") == 0) {
+            valid_auth_indicators++;
+            if (!(ua & IPADB_USER_AUTH_PKINIT)) {
+                kerr = KRB5KDC_ERR_POLICY;
+                *status = "PKINIT pre-authentication not allowed for this user.";
+                goto done;
+            }
+
```
frozencemetery: Here too.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -44,13 +45,90 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
...
+    }
+
+    if (!valid_auth_indicators) {
+        if (!(ua & IPADB_USER_AUTH_PASSWORD)) {
+            kerr = KRB5KDC_ERR_POLICY;
+            *status = "Non-hardened password authentication not allowed for this user.";
+        }
+    }
+
+done:
```
frozencemetery: If all that's happening in this label is a `return`, don't bother with the label - there's no cleanup, so it's easier to follow if you just inline the return.

ipaclient/install/client.py
```diff
@@ -663,20 +663,24 @@ def hardcode_ldap_server(cli_server):
...
-def configure_krb5_conf(
-        cli_realm, cli_domain, cli_server, cli_kdc, dnsok,
-        filename, client_domain, client_hostname, force=False,
-        configure_sssd=True):
-
-    # First, write a snippet to krb5.conf.d.  Currently this doesn't support
-    # templating, but that could be changed in the future.
+# Currently this doesn't support templating, but that could be changed in the
+# future.  Note that this function is also called from %post.
+def configure_krb5_snippet():
```
abbra: I think we should also add `tasks.restore_context()` for the file added here. I found out that SELinux policy doesn't explicitly label `/etc/krb5.conf.d/*` other than `etc_t` but this will change in the future.

ipaserver/install/server/upgrade.py
```diff
@@ -1643,36 +1643,15 @@ def setup_spake(krb):
...
-                    logger.error('augeas: %s', aug.get(error_path))
-                raise
-
-            if krb.is_running():
-                krb.stop()
-            krb.start()
-    finally:
-        aug.close()
+# Currently, this doesn't support templating.
+def enable_server_snippet():
```
abbra: Same here. Please use `tasks.restore_context(...)`

tengcm2015:
Added restore_context() for newly created files.

abbra:
@tengcm2015, so what we miss here is per-indicator maxlifetime KDC policy. I'm going to look at it next week.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -0,0 +1,159 @@
...
+#include <errno.h>
+#include <syslog.h>
+#include <krb5/kdcpolicy_plugin.h>
+
+#include "ipa_krb5.h"
+#include "ipa_kdb.h"
+
+struct krb5_kdcpolicy_moddata_st {
+    /* Storage struct for kdcpolicy callback state */
+    ;
```
simo5: why an empty struct ?
is it actually required to have a pointer at all ?

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -0,0 +1,159 @@
...
+    /* If no mechanisms are set, allow every auth method */
+    if (ua == IPADB_USER_AUTH_NONE) {
+        return 0;
+    }
+
+    /* For each auth indicator, see if it is allowed for that user */
+    for (int i = 0; auth_indicators[i] != NULL; i++) {
+        const char *auth_indicator = auth_indicators[i];
+
+        if (strcmp(*auth_indicators, "otp") == 0) {
```
simo5: *auth_indicators -> auth_indicator and everywhere below

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -0,0 +1,159 @@
...
+                return KRB5KDC_ERR_POLICY;
+            }
+        } else if (strcmp(*auth_indicators, "hardened") == 0) {
+            valid_auth_indicators++;
+            /* Allow hardened even if only password pre-auth is allowed */
+            if (!(ua & (IPADB_USER_AUTH_HARDENED | IPADB_USER_AUTH_PASSWORD))) {
+                *status = "Password pre-authentication not not allowed for this user.";
+                return KRB5KDC_ERR_POLICY;
+            }
+        }
```
simo5: So I assume we do not have an indicator for regular password based pre-auth, so I would add a comment to that effect in the code, and some handling if ua == IPADB_USER_AUTH_PASSWORD for some reason, otherwise this code will prevent a user with just the password ua and not indicators from being able to proceed at all.

simo5: Ah I see that IPADB_USER_AUTH_PASSWORD is handled below as a special case ... uhmmm storngly advise some comment here that explains the reason, otherwise it will be hard to figure it out down the road once we forgot why.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
frozencemetery: Nope.  We could delete it entirely if we don't need any state here.  Historically, it's only in the commit at all because it made sense to expose as much in the skeleton commit as I could.

frozencemetery: Nope.  We could delete it entirely if we don't need any state here.  Historically, it's only in the commit at all because it made sense to expose as much in the skeleton commit as I could.

frozencemetery: Technically this does mean that `ipa_kdcpolicy_init` and `ipa_kdcpolicy_fini` aren't needed either - we can just make them NULLs in the vtable.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
tengcm2015: Comment added.

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
tengcm2015: I know why this worked before, but I'm still surprised that it actually did...

frozencemetery:
Hi, what are the next steps toward merging this?

abbra:
@tengcm2015, @frozencemetery, first, let me thank you for this outstanding body of work. It is almost ready to be merged.
In https://github.com/abbra/freeipa/tree/tengcm2015-pre-auth-policy I did following:
- rebased to current master
- added authentication indicator-specific Kerberos ticket policy support

I reserved two new LDAP attributes for max ticket life and max renewable age, similar to existing ones, in Red Hat's OID registry. The difference these attributes have with existing ones are the following:
- they are multi-valued
- they aren't supposed to be used in their 'naked' form. Rather than using `krbAuthIndMaxTicketLife: 604800`, we are storing the values in option-based attributes: `krbAuthIndMaxTicketLife;pkinit: 604800`.

LDAP attributes have the concept of options appended to the attribute name. They are used for both server-side and client-side specific handling of the attribute value. For example, `userCertificate;binary` says that the value of `userCertificate` attribute is binary-encoded rather than a text. In FreeIPA we are using this approach already for restricting access to Kerberos keys of principals with `ipaAllowedToPerform;read_keys` or `ipaAllowedToPerform;write_keys`, so this is not a new use.

Sadly, the old Kerberos ticket policy attributes were defined as single valued, so I couldn't use them directly as all options represent themselves as multiple values of the same attribute.

Please look at my branch and feel free to pick the top commit which contains all values. I have not tested it yet myself, only compiled and linted.

We need to add a test that exercises new options to `ipa krbtpolicy-mod` and see how to add the same support to Web UI.

simo5:
What is the reason to change ticket life time based on the authentication method used ?

abbra:
@simo5 this is one of use cases defined for the kdc policy in the first place. I think it should be covered in the design document, but the idea was to allow differentiating what admins consider preferred method by the ticket policy.

simo5:
Sorry I do not get it, but my question was very direct:
in what cases it makes sense to have a different ticket lifetime based on which authentication method you used ?
if we do not have clear uses cases we should stay away from adding more custom fields imo

frozencemetery:
>  in what cases it makes sense to have a different ticket lifetime based on which authentication method you used ?

One is two-factor.  Today, for instance, one could define OTP creds to be longer-lived than those without.  (In the future, this would be true of other labeled second factors as well.)

It was also explicitly requested for PKINIT (for smart cards).

frozencemetery:
(@abbra, do you have anything to add to the above?)

simo5:
> > in what cases it makes sense to have a different ticket lifetime based on which authentication method you used ?
> 
> One is two-factor. Today, for instance, one could define OTP creds to be longer-lived than those without. (In the future, this would be true of other labeled second factors as well.)

I guess I simply do not understand why the auth method makes any difference in how long creds last, to me the lifetime of tickets is a matter of defending from potential theft of forgotten tickets on some machine, that risk, though has nothing to do with the auth method used to get them, only with the security of the ccache storage which is orthogonal.

> It was also explicitly requested for PKINIT (for smart cards).

Is there here some assumption that some methods will give higher privileged TGTs than others auth methods? So the risk analysis of losing the TGT changes ?

abbra:
Considering how authentication indicators were supposed to be used, yes, it was implied that something like PKINIT-issued ticket would be higher privileged and might require a different ticket life. It might be shorter or longer, it would still be different and driven by a policy rather than a client's request.

simo5:
I am not 100% convinced you really need this level of fine tuning for 99% of deployment, so I see it as "bloat", but if you rally have people asking you this explicitly I won't complain.
I just would not like to see a feature implemented for the sake of having an additional knob to tune.

abbra:
In order to preserve a history of discussion in PR, I'm going to submit my additional commit as a separate pull request and accept this one. We just need to make sure we aren't releasing new version before accepting that new PR.

LGTM.

abbra:
master:

* 9c0a35f1e79033585786c2567aedcb3b4c10a6b6 Add new authentication indicators in kdc.conf.template
* d0570404ef5a79dfc08d7959d21e9e4843973faf Extend the list of supported pre-auth mechanisms in IPA server API
* c7b938a1d5c20df24a2d8a62019c5341e0f26c63 Enable krb5 snippet updates on client update
* 39e3704a0679d117f5145044e73726175a8f600b Move certauth configuration into a server krb5.conf template
* 179c8f4009adc30b9b3c497855f15927016c84db Add a skeleton kdcpolicy plugin
* 15ff9c8fecdff5556e27b7c9eebd45d327044bc0 Implement user pre-authentication control with kdcpolicy plugin
* b66e8a1ee295e11e286979a0bd9ce303cbab2aa5 Modify webUI to adhere to new IPA server API
* 952dd2a50fe3c3dc099b14af23cd050048235fa3 Add design document

abbra:
I removed backport to ipa-4-7 since we didn't decide how to get API versioning for the backports and 4.7 is only in Fedora 29 which get outdated soon.

PR #3360, opened 2019-07-02
Use system-wide crypto policy for TLS ciphers

tiran:
IPA now uses the system-wide crypto policy for TLS ciphers on RHEL. It's
also now possible to keep the default policy by setting TLS_HIGH_CIPHERS
to None.

Fixes: https://pagure.io/freeipa/issue/7998

### DIFF
--- a/ipalib/util.py
+++ b/ipalib/util.py
@@ -339,9 +339,9 @@ def create_https_connection(
         ssl.OP_SINGLE_ECDH_USE
     )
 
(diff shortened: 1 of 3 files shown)


abbra:
LGTM.

tiran:
master:

* b55344888478e2b05dc01200bed87e551aa7d00a Use system-wide crypto policy for TLS ciphers

PR #3364, opened 2019-07-03
Use nis-domainname.service on all RH platforms

tiran:
RHEL 8 and Fedora >= 29 use "nis-domainname.service" as service name for
domainname service. Remove special code in ipaplatform.rhel and for Fedora
< 28. Only Fedora 29+ is supported by IPA 4.8.

Fixes: https://pagure.io/freeipa/issue/8004


flo-renaud:
master:

* d2c929270c6184022b58799b61bb640cdcdf10a8 Use nis-domainname.service on all RH platforms

PR #3365, opened 2019-07-03
[ipa-4-6] tox: force pytest version to the 4.6.4

netoarmando:
As of today, latest version of pytest (5.0.0), released on June 28th is freezing tox on py36 right after the test session starts.

Version 4.6.4 was released on the same date.


rcritten:
ipa-4-6:

* 9a4f9a470986d2c5629f0f3c7935c49849677a73 tox: force pytest version to the 4.6.4

PR #3366, opened 2019-07-03
ipatests/azure: display actual dnf repo URLs

fcami:
Display which dnf repositories were available at the prepare-build step via metalink.
Also display the fastestmirror cache.
    

fcami:
This should not make builds slower as the makecache step is now much faster due to the added repolist step.

flo-renaud:
Since the change does not impact the build time and it could help us troubleshoot in case of future issues, ACK.

fcami:
master:

* be7f54d40c33d147fdaf0d3cd65b56452a4f001a ipatests/azure: display actual dnf repo URLs

PR #3372, opened 2019-07-05
Fix `test_webui.test_selinuxusermap`

stanislavlevin:
A previous refactoring of SELinux tests has had a wrong
assumption about the user field separator within
ipaSELinuxUserMapOrder. That was '$$', but should be just '$'.

Actually, '.ldif' and '.update' files are passed through
Python template string substitution:

> $$ is an escape; it is replaced with a single $.
> $identifier names a substitution placeholder matching
> a mapping key of "identifier"

This means that the text to be substituted on should not be escaped.
The wrong ipaSELinuxUserMapOrder previously set on a fresh install will be replaced on
upgrade.

Fixes: https://pagure.io/freeipa/issue/8005

### DIFF
--- a/install/updates/50-ipaconfig.update
+++ b/install/updates/50-ipaconfig.update
@@ -1,4 +1,5 @@
 dn: cn=ipaConfig,cn=etc,$SUFFIX
+replace: ipaSELinuxUserMapOrder: guest_u:s0$$$$xguest_u:s0$$$$user_u:s0$$$$staff_u:s0-s0:c0.c1023$$$$sysadm_u:s0-s0:c0.c1023$$$$unconfined_u:s0-s0:c0.c1023::$SELINUX_USERMAP_ORDER
 replace: ipaSELinuxUserMapOrder: ipaSELinuxUserMapOrder: guest_u:s0$$xguest_u:s0$$user_u:s0$$staff_u:s0-s0:c0.c1023$$unconfined_u:s0-s0:c0.c1023::guest_u:s0$$xguest_u:s0$$user_u:s0$$staff_u:s0-s0:c0.c1023$$unconfined_u:s0-s0:c0.c1023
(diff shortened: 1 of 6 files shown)


flo-renaud:
Hi @stanislavlevin 
thanks for the PR. Could you add a temporary additional commit in order to trigger the tests that you fixed? You just need to modify `ipatests/prci_definitions/gating.yaml` and add the test suite. You can find the exact syntax in `ipatests/prci_definitions/nightly_master.yaml` (gating defines the list of tests run on each PR, while nightly_master defines the list of tests run in the nightly execution). test_webui_policy.py is normally launched only during the nightly.

Once the run is green, someone will review your PR and if OK provide an ACK and ask you to remove the temp commit.

stanislavlevin:
@flo-renaud, thanks for the explanation. Waiting for "someone" :smile:

flo-renaud:
I am adding the re-run label as test `fedora-30/dnssec` failed but the issue seems unrelated (see [results](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/e8975128-9f85-11e9-b8b8-5254000972fa/)).

flo-renaud:
I am setting the ipa-4-8 label as the initial fix (commit b2acd65013348c0f5d91582f240714b0f3357678) was included in 4.8.0 and requires this new patch:
```
$ git tag --contains b2acd65013348c0f5d91582f240714b0f3357678
release-4-8-0
```

flo-renaud:
@stanislavlevin 
Could you also add
Fixes: https://pagure.io/freeipa/issue/7996
to your commit? It will be easier for us to track the patches for this issue.

stanislavlevin:
@flo-renaud, thank you so much for the review!
I'm sorry, somehow miss your messages.

Temp commit was removed.
Ticket link was attached.

flo-renaud:
Adding re-run label as "Process "vagrant provision" exited with error code 1"

flo-renaud:
Hi @stanislavlevin 
thanks for your patch, everything looks good, ready to be merged.

abbra:
master:

* ac1ea0ec6710f40adffc3f04b39422f3043c6554 Fix `test_webui.test_selinuxusermap`

PR #3373, opened 2019-07-05
[ipa-4-6] Update the ciphers list

flo-renaud:
The previous list of ciphers was allowing weak algorithms.
The fix removes:
- kECDH: cipher suites using fixed ECDH key agreement signed by CAs with RSA
and ECDSA keys or either respectively.
- kDH: cipher suites using DH key agreement and DH certificates signed by
CAs with RSA and DSS keys or either respectively.

Fixes: https://pagure.io/freeipa/issue/8000

### DIFF
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -55,7 +55,7 @@ TLS_VERSION_DEFAULT_MAX = "tls1.2"
 
 # high ciphers without RC4, MD5, TripleDES, pre-shared key
 # and secure remote password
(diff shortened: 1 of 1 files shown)


flo-renaud:
Note: this change is required only on ipa-4-6 branch as newer branches use ciphers defined at the system level

abbra:
LGTM.
@tiran any opinion here?

tiran:
Go for it!

abbra:
ipa-4-6:

* 91641ac1ad5053487557b3b5a78f54fa157bfb38 Update the ciphers list

PR #3380, opened 2019-07-08
upgrade: remove ipaCert and key from /etc/httpd/alias

flo-renaud:
With ipa 4.5+, the RA cert is stored in files in
/var/lib/ipa/ra-agent.{key|pem}. The upgrade code handles
the move from /etc/httpd/alias to the files but does not remove
the private key from /etc/httpd/alias.

The fix calls certutil -F -n ipaCert to remove cert and key,
instead of -D -n ipaCert which removes only the cert.

Fixes: https://pagure.io/freeipa/issue/7329

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -892,6 +892,32 @@ class NSSDatabase:
     def delete_cert(self, nick):
         self.run_certutil(["-D", "-n", nick])
 
+    def delete_key_only(self, nick):
+        """Delete the key with provided nick
+
+        This commands removes the key but leaves the cert in the DB.
+        """
+        keys = self.list_keys()
+        # keys is a list of tuple(slot, algo, keyid, nickname)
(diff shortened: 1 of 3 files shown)


flo-renaud:
Note: a stalled PR #1441 was trying to fix this issue but was left untouched since April.

frasertweedale:
ACK from me.  I was wondering "does `certutil` even work like that?" but the tests have put to rest my concerns :)

abbra:
@flo-renaud what branches this change needs to go into? Right now there is none marked.

flo-renaud:
@abbra 
The ticket has milestone = 4.7.2, so I would say ipa-4-7.

abbra:
master:

* ef39e1b02a2ef965997d38fc7b72d5ee1542d44b upgrade: remove ipaCert and key from /etc/httpd/alias

PR #3384, opened 2019-07-11
[WIP] Test to check certmonger does not support insecure cryptography for S…

ssidhaye:
Test to check certmonger does not support insecure cryptography for SCEP enrollment

Problem:
Presently, when enrolling a certificate, scep-submit will only use DES and MD5.
The current SCEP draft implementation notes that both AES and SHA256 are mandatory
to implement functionality per
https://tools.ietf.org/html/draft-gutmann-scep-06#section-2.8
and that implementations should not support DES or MD5.

Solution:
In order to address the above problem,
A fix was added to add additional cipher and digests per a more recent SCEP specification

### DIFF
--- a/ipatests/test_integration/test_insecure_crypto_SCEP.py
+++ b/ipatests/test_integration/test_insecure_crypto_SCEP.py
@@ -1,0 +1,162 @@
+#
+# Copyright (C) 2019  FreeIPA Contributors see COPYING for license
+#
+
+import glob
+
+from ipaplatform.paths import paths
+
+from ipatests.test_integration.base import IntegrationTest
+
+AUTHORIZATION_FILE = '/var/lib/pki/pki-tomcat/ca/conf/flatfile.txt'
+CERTMONGER_CONFIG = '/etc/sysconfig/certmonger'
+
+
+def enable_SCEP(host):
+    """Enable SCEP in Certificate Authority's CS.cfg
+    by setting ca.scep.enable=true
+    """
+    content = host.get_file_contents(paths.CA_CS_CFG_PATH,
+                                     encoding='utf-8')
+    new_lines = []
+    input_line = "auths.instance.flatFileAuth.deferOnFailure=false"
+    for line in content.split('\n'):
+        if line.startswith('auths.instance.flatFileAuth.deferOnFailure'):
+            new_lines.append(input_line)
+        elif line.startswith('ca.scep.enable'):
+            new_lines.append("ca.scep.enable=true")
+        else:
+            new_lines.append(line)
+    host.put_file_contents(paths.CA_CS_CFG_PATH, '\n'.join(new_lines))
+
+
+def add_local_ip_to_authfile(host):
+    """Add the local IP to the authorization file,
+    /var/lib/pki/pki-tomcat/ca/conf/flatfile.txt
+    """
+    content = host.get_file_contents(AUTHORIZATION_FILE,
+                                     encoding='utf-8')
+    new_lines = []
+    for line in content.split('\n'):
+        if line.startswith('#UID:'):
+            new_lines.append("UID:%s" % host.ip)
+        elif line.startswith("#PWD:"):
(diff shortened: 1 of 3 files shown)


fcami:
Fails due to https://github.com/freeipa/freeipa-pr-ci/issues/281

ipatests/test_integration/test_insecure_crypto_SCEP.py
```diff
@@ -0,0 +1,141 @@
...
+            new_lines.append(line)
+    host.put_file_contents(CERTMONGER_CONFIG, '\n'.join(new_lines))
+
+
+def set_scep_cipher(host):
+    """hard code the default SCEP cipher to DES3 since there
+    is no CLI to do this by adding scep_cipher=DES3
+    """
+    input_file = None
+    for file in glob.glob('/var/lib/certmonger/cas/*'):
```
tiran: ``file`` was a reserved name in Python 2. We still avoid the name. Please use ``filename`` instead.

The function assumes that the controller machine has certmonger installed. This may not be the case. Please use ``host.run_command(['find', '/var/lib/certmonger/cas', '-type', 'f'])`` to get a list of file names, then retrieve the file from the host.

ipatests/test_integration/test_insecure_crypto_SCEP.py
```diff
@@ -0,0 +1,141 @@
...
+    host.put_file_contents(input_file, content)
+
+
+class TestInsecureCryptoSCEP(IntegrationTest):
+    """This test checks that certmonger does not support
+    insecure cryptography for SCEP enrollment
+    """
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master)
```
tiran: You don't need the install method of you add ``topology = 'star'`` class attribute.

ipatests/test_integration/test_insecure_crypto_SCEP.py
```diff
@@ -0,0 +1,141 @@
...
+        if line.startswith('#UID:'):
+            new_lines.append("UID:%s" % host.ip)
+        elif line.startswith("#PWD:"):
+            new_lines.append("PWD:1234")
+        else:
+            new_lines.append(line)
+    host.put_file_contents(AUTHORIZATION_FILE, '\n'.join(new_lines))
+
+
+def edit_certmonger_config(host):
```
tiran: ``certmonger.service`` only uses ``OPTS``. You don't have to retrieve and modify the file. Just upload a new file with content ``OPTS=-d3\n``

ipatests/test_integration/test_insecure_crypto_SCEP.py
```diff
@@ -0,0 +1,141 @@
...
+        3. Increase cermonger debug output
+        4. Add the SCEP CA to certmonger
+        5. Hard code the default SCEP cipher to DES3
+           since there is no CLI to do this
+        6. Try a request making changes to certmonger
+        7.
+        8.
+        """
+        # Stop CA
+        self.master.run_command(['systemctl', 'stop',
```
tiran: Please use a different indention style for multi-line lists. The code gets more readable and usually shorter if you indent the body of the list like this:

```
        self.master.run_command([
            'systemctl', 'stop', 'pki-tomcatd@pki-tomcat.service'
        ])
``

rcritten:
Does this test belong in IPA CI? IPA doesn't normally enable SCEP.

ipatests/test_integration/test_insecure_crypto_SCEP.py
ssidhaye: added the topology

ipatests/test_integration/test_insecure_crypto_SCEP.py
ssidhaye: Fixed it

rcritten:
SCEP is broken in dogtag, the password authentication always fails, so I don't know that this PR will work until that is fixed.

PR #3385, opened 2019-07-11
Make use of Azure Pipeline slicing

stanislavlevin:
The unit tests execution time within Azure Pipelines(AP) is not balanced.
One test job(Base) takes ~13min, while another(XMLRPC) ~28min.
Fortunately, AP supports slicing:
    
> An agent job can be used to run a suite of tests in parallel. For
    example, you can run a large suite of 1000 tests on a single agent.
    Or, you can use two agents and run 500 tests on each one in parallel.
    To leverage slicing, the tasks in the job should be smart enough to
    understand the slice they belong to.
    
>The step that runs the tests in a job needs to know which test slice
    should be run. The variables System.JobPositionInPhase and
    System.TotalJobsInPhase can be used for this purpose.
    
Thus, to support this pytest should know how to split the test suite into groups(slices).
For this, a new internal pytest plugin was added.

The number of groups becomes 3 (was 2). With applying of such slicing the AP time is reduced from ~43 min to ~33 min.

The same could be applied to Travis, though the profit would be about 3 min.

This PR includes other fixes required for correct working.

Concerning new `101 warnings` (`RemovedInPytest4Warning: usage of Session.XXX is deprecated, please use pytest.XXX instead`):
https://github.com/pytest-dev/pytest/issues/2701
```
pytest 3.10.0 (2018-11-03)
...
Bug Fixes

    #2701: Fix false RemovedInPytest4Warning: usage of Session... is deprecated, please use pytest warnings.
```

Fixes: https://pagure.io/freeipa/issue/8007
Fixes: https://pagure.io/freeipa/issue/8008
Fixes: https://pagure.io/freeipa/issue/8009

### DIFF
--- a/.tox-install.sh
+++ b/.tox-install.sh
@@ -4,8 +4,9 @@ set -ex
 FLAVOR="$1"
 ENVPYTHON="$(realpath "$2")"
 ENVSITEPACKAGESDIR="$(realpath "$3")"
-# 3...end are package requirements
-shift 3
+ENVDIR="$4"
+# 4...end are package requirements
+shift 4
 
 TOXINIDIR="$(cd "$(dirname "$0")" && pwd)"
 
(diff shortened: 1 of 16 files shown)


abbra:
Looks good. PR CI failed to provision a build system so I applied 're-run' label to restart it. Azure and Travis tests succeeded.

abbra:
master:

* 2312b38a678d30b73c08dae19db7f8b74361a957 Simplify ipa-run-tests script
* a2d4e2a61f46cf6337786b2cb031e3a9f394a708 Make use of Azure Pipeline slicing
* c7500220c443c6051178211eadd64392ca7bd08a Avoid use of '/tmp' for pip operations

PR #3403, opened 2019-07-16
Azure Pipelines: allow test re-runs

abbra:
Modify the way we name saved artifacts to allow re-runs of individual pipeline jobs.

This PR also includes changes required to allow Rawhide runs but doesn't change the code to run on Rawhide.

A separate commit doing that on Azure can be seen here: https://github.com/abbra/freeipa/pull/15/commits/950dd764e159fe5e25bba8f184808d6daa2ba8c7

### DIFF
--- a/.test_runner_config.yaml
+++ b/.test_runner_config.yaml
@@ -28,6 +28,7 @@ steps:
   builddep:
   - rm -rf /var/cache/dnf/*
   - "dnf makecache || :"
+  - dnf -y module enable nodejs:12
   - dnf builddep -y ${builddep_opts} -D "with_wheels 1" --spec freeipa.spec.in --best --allowerasing --setopt=install_weak_deps=False
   - dnf install -y gdb
   - dnf update -y annobin
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -169,7 +169,7 @@ BuildRequires:  libuuid-devel
 BuildRequires:  libsss_idmap-devel
 BuildRequires:  libsss_certmap-devel
 BuildRequires:  libsss_nss_idmap-devel >= %{sssd_version}
-BuildRequires:  nodejs(abi) < 11
+BuildRequires:  nodejs(abi)
 BuildRequires:  uglify-js
 BuildRequires:  libverto-devel
 BuildRequires:  libunistring-devel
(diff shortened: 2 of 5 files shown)


abbra:
All tests completed correctly. @flo-renaud @fcami please review.

stanislavlevin:
@abbra, hello.
Could you please provide a link to docs ( if any) on System.StageAttempt and System.PhaseAttempt? How do they influence tests rerun? Which interface is used to rerun? (Azure Pipelines, github pane, etc.). In other words, what can I see applying this? 
Thank you.

abbra:
@stanislavlevin it is not really documented yet but it is part of the pipeline configuration exposed by the azure agent: https://github.com/microsoft/azure-pipelines-agent/search?q=StageAttempt&unscoped_q=StageAttempt. I guess it will get eventually documented at https://docs.microsoft.com/en-us/dotnet/api/microsoft.teamfoundation.distributedtask.pipelines.runtime?view=azure-devops-dotnet-preview

There are other places where this can be found (like VsTest2 which reads the values from the pipeline configuration) but the agent code looks like a core of it.

abbra:
As to the second part of your question, you can trigger re-run via github pane, it works from there. You can also trigger re-run on Azure Pipeline itself but there it is a bit more complicated -- you might get a restart of a whole build instead of a single job or phase.

Once a job is restarted, you'll see the phase name changing to "Attempt # <number>" and there will be a drop-down menu allowing to choose previous attempts to see historical logs data.

The change for artifacts naming shows that attempt value is increasing with reruns, so we never overwrite the artifact from a previous attempt, thus uploading artifacts succeeds.

stanislavlevin:
@abbra, thank you for the explanation. I will look at vstest.
As I understand correctly, this PR about artifacts naming on reruns.

abbra:
@stanislavlevin yes, this PR is about allowing to hit 're-run failed tests' button. You could do it already but then uploading artifacts will fail because there is already an artifact named the same.

It is one of few remaining changes to allow us to drop Travis CI.

flo-renaud:
Hi @abbra 
changes looks good to me.

abbra:
master:

* 5530911f20a7a6d2831c9ce35948d63b226f212f Fix rpmlint errors for Rawhide
* d55c9b6d7f0eefa5648fec5c0fb05ca475acd4fd Use any nodejs version instead of forcing a version before nodejs 11
* 0187a746c9197d5447e08ce2be49e3e021aae49b Use stage and phase attempt counters when saving test artifacts

PR #3413, opened 2019-07-18
test_xmlrpc: fix TestAutomemberFindOrphans.test_find_orphan_automembe…

flo-renaud:
…r_rules

Test scenario:
- create a hostgroup
- create a host
- create an automember rule for the hostgroup with a condition fulfilled
by the host
- delete the hostgroup
- call automember-rebuild (1)
- call automember-find-orphans to remove the orphan automember group
- call automember-rebuild(2)

The test was expecting the first rebuild command to fail but this
assumption is not true if the DS version is >= 1.4.0.22 because of the
fix for https://pagure.io/389-ds-base/issue/50077

Modify the test so that it expects failure only when DS is older.

Fixes: https://pagure.io/freeipa/issue/7902

### DIFF
--- a/ipatests/test_xmlrpc/test_automember_plugin.py
+++ b/ipatests/test_xmlrpc/test_automember_plugin.py
@@ -35,7 +35,9 @@ from ipaserver.plugins.automember import REBUILD_TASK_CONTAINER
 
 import time
 import pytest
+import re
 import unittest
+from pkg_resources import parse_version
 
 try:
     from ipaserver.plugins.ldap2 import ldap2
@@ -885,7 +887,6 @@ class TestAutomemberFindOrphans(XMLRPC_test):
 
         hostgroup1.retrieve()
 
-    @pytest.mark.skip(reason="Fails with 389-DS 1.4.0.22, see issue 7902")
     def test_find_orphan_automember_rules(self, hostgroup1):
         """ Remove hostgroup1, find and remove obsolete automember rules. """
         # Remove hostgroup1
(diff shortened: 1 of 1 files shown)


flo-renaud:
The test was skipped on ipa-4-7, ipa-4-8 and mastre branches, hence setting the corresponding labels.
This PR replaces PR #3004 with a version check not using lib389.

ipatests/test_xmlrpc/test_automember_plugin.py
```diff
@@ -885,20 +886,33 @@ def test_create_deps_for_find_orphans(self, hostgroup1, host1,
...
+        # feature was implemented: https://pagure.io/389-ds-base/issue/50077
+        if not have_ldap2:
+            raise unittest.SkipTest('server plugin not available')
+        ldap = ldap2(api)
+        ldap.connect()
+        rootdse = ldap.get_entry(DN(''), ['vendorVersion'])
+        version = rootdse.single_value.get('vendorVersion')
+        mo = re.search(r'(\d+)\.(\d+)\.(\d+)\.(\d+)*', version)
+        vendor_version = tuple(int(v) for v in mo.groups())
+        expected_failure = vendor_version < (1, 4, 0, 22)
```
rcritten: Would it be safer to use pkg_resources.parse_version instead of assuming that this will always be made up of ints?

flo-renaud: vendorVersion has the following format in the rootDSE: 389-Directory/1.3.8.4 B2019.037.1535

To use pkg_resources, I would need to trim the leading `389-Directory`/ and trailing `Byyyy.ddd.hhhh`, and in my mind it's not safer because there would also be an assumption on the format. But I can change if you prefer.

rcritten: The re already assumes a format of 4 ints.

I was mostly thinking it was a more python-ish way to compare version numbers.

I'm not super worried about this but if the format changes it could provoke a false positive failure I suppose.

flo-renaud: Ok, updated the PR with pkg_resource.parse_version to handle the comparison.

tiran:
master:

* 11e40336c5f0c5b086bbb7a98fb5cd22c80b4df3 test_xmlrpc: fix TestAutomemberFindOrphans.test_find_orphan_automember_rules

Tiboris:
@tiran I just noticed that you probably forgot to create backport to ipa-4-7

PR #3414, opened 2019-07-18
Nightly test definition: add missing tests

flo-renaud:
The following test was missing from all nightlies:
- test_integration/test_crlgen_manage.py

The following tests were missing from nightly_f28 and nightly_f29:
- test_integration/test_nfs.py
- test_integration/test_smb.py

The following test was missing from nightly_rawhide:
- test_integration/test_smb.py

### DIFF
--- a/ipatests/prci_definitions/nightly_f29.yaml
+++ b/ipatests/prci_definitions/nightly_f29.yaml
@@ -231,6 +231,18 @@ jobs:
         timeout: 4800
         topology: *master_1repl_1client
 
+  fedora-29/test_smb:
+    requires: [fedora-29/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-29/build_url}'
+        test_suite: test_integration/test_smb.py
+        template: *ci-master-f29
+        timeout: 4800
+        topology: *master_1repl_1client
+
   fedora-29/test_server_del:
     requires: [fedora-29/build]
     priority: 50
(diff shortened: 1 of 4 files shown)


abbra:
@flo-renaud I think we shouldn't add test_smb.py to Fedora 28 tests. I haven't tested that, actually, and there is no real need to work on F28 as it is not supported anymore.

tiran:
How about we just drop F28 testing completely from master and 4.8?

abbra:
Yes, we need to drop F28 completely. We cannot influence updates to F28 because they are not possible anymore, so anything that would be requiring changes there will not be fixable anyway.

flo-renaud:
@tiran @abbra 
thanks for the review. You are right, we don't schedule anymore tests on fedora 28, meaning the nightly_f28.yaml file doesn't need to be updated.

ipatests/prci_definitions/nightly_f29.yaml
```diff
@@ -1229,6 +1241,18 @@ jobs:
...
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-29/build_url}'
+        test_suite: test_integration/test_nfs.py::TestNFS
+        template: *ci-master-f29
+        timeout: 9000
+        topology: *master_2repl_1client
+
```
rcritten: This already exists in the file on line https://github.com/freeipa/freeipa/blob/master/ipatests/prci_definitions/nightly_f29.yaml#L1220

flo-renaud: @rcritten 
thanks for the review. You're right, I somehow didn't catch this one. PR updated.

abbra:
master:

* 41e5d4653a637bb3b97cf651382fb7668f41e226 Nightly test definition: add missing tests

PR #3418, opened 2019-07-19
Remove posixAccount from service_find search filter

rcritten:
This will allow cifs principals to be found. They were suppressed
because they include objectclass=posixAccount.

This is a bit of a historical anomaly. This was included in the
filter from the initial commit (though it was person, not
posixAccount). I believe it was a mistake from the beginning but
it wasn't noticed because it didn't cause any obvious issues.

https://pagure.io/freeipa/issue/8013


abbra:
LGTM.

rcritten:
Adding ack flag per @abbra

rcritten:
master:

* e771fa59ff65545ff1e84f1cd30e06556fabcee3 Remove posixAccount from service_find search filter

PR #3434, opened 2019-07-23
ipa-server-install: guarantees that FQDN is at beginning on /etc/hosts

mulatinho:
On Fedora 30, and also on another machines that put aliases
first than FQDN address on /etc/hosts, ipa-server-install fails.

#### Steps to Reproduce
1. with a machine without freeipa-server,
2. using the same pattern of fedora 30 in /etc/hosts [1]
3. run **ipa-server-install --setup-dns** make happen an installation bug even if it seems OK;

[1] /etc/hosts
```
127.0.0.1 localhost localhost.localdomain localhost4 localhost4.localdomain4 
127.0.0.1 localhost localhost.localdomain localhost4 localhost4.localdomain4
10.0.0.1 serveripa serveripa.localdomain
```

Instead of checking the first index of ex_name, this code checks if
FQDN is on tuple or not. If not raises error, if it is continues the
installation.

Fixes: https://pagure.io/freeipa/issue/8018

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -104,6 +104,10 @@ class UpgradeMissingVersionError(UpgradeVersionError):
     pass
 
 
+class OrderHostsError(Exception):
+    pass
+
+
 class ReplicaConfig:
     def __init__(self, top_dir=None):
         self.realm_name = ""
@@ -162,11 +166,18 @@ def verify_fqdn(host_name, no_host_dns=False, local_hostname=True):
         try:
             logger.debug('Check if %s is a primary hostname for localhost',
                          host_name)
-            ex_name = socket.gethostbyaddr(host_name)
+            ex_name, aliases = socket.gethostbyaddr(host_name)[:2]
             logger.debug('Primary hostname for localhost: %s', ex_name[0])
-            if host_name != ex_name[0]:
-                raise HostLookupError("The host name %s does not match the primary host name %s. "\
-                        "Please check /etc/hosts or DNS name resolution" % (host_name, ex_name[0]))
+            # if FQDN is not the first item on the /etc/hosts file
+            if host_name != ex_name:
+                if host_name in aliases:
+                    # correct the order of hosts file to user
+                    order_hosts_file(host_name)
+                else:
+                    raise HostLookupError(
+                        "The host name %s does not match the primary host "
+                        "name %s. Please check /etc/hosts or DNS name "
+                        "resolution" % (host_name, ex_name[0]))
         except socket.gaierror:
             pass
         except socket.error as e:
@@ -258,6 +269,35 @@ def record_in_hosts(ip, host_name=None, conf_file=paths.HOSTS):
             continue
 
     return None
+
+
+def order_hosts_file(host_name):
+    """Read and overwrite /etc/hosts with FQDN at beginning, issue #8018"""
+    normalize_host = ''.join(host_name)
+    hostsfile = '/etc/hosts'
+    newfile = tempfile.NamedTemporaryFile(delete=False)
+
+    with open(hostsfile, 'r') as oldfile:
+        for line in oldfile:
+            if normalize_host in line:
+                restring = re.sub(r'%s( |$)' % normalize_host, '',
+                                  line.rstrip())
+                line = re.sub(r'(.*?) (.*)', r'\1 %s \2\n' % normalize_host,
+                              restring.rstrip())
+
+            newfile.write(line.encode())
+        newfile.close()
+
+    try:
+        shutil.copyfile(hostsfile, '%s.orig' % hostsfile)
+        shutil.copystat(hostsfile, newfile.name)
+        shutil.move(newfile.name, hostsfile)
+        tasks.restore_context(hostsfile, force=True)
+    except OSError:
+        raise OrderHostsError("Error while trying putting FQDN after IP "
+                              "Address on '/etc/hosts' file (e.g.: 10.0.0.1 "
+                              "server.example.com server).")
+
 
 def add_record_to_hosts(ip, host_name, conf_file=paths.HOSTS):
     hosts_fd = open(conf_file, 'r+')


abbra:
@mulatinho one problem I am concerned about here is that after deployment you'll still get unqualified hostname resolution. It means, that for applications that construct their Kerberos principals based on what OS returns, there will be use of a principal that is not known by the KDC.

E.g., if hostname resolves to `some-host` instead of `some-host.example.test`, constructing HTTP/.. principal would use `HTTP/some-host` instead of `HTTP/some-host.example.test` and the service ticket acquisition request will fail.

rcritten:
Exactly. Both Kerberos and TLS are dependent upon hostnames. Using the FQDN is a predictable way to ensure that things always work.
Look in the git history and there are dozens of corner cases where the installer was tightened up to catching things like this.

abbra:
Note that CI will not catch this case as we are enforcing use of fully qualified host names in a test setup

abbra:
Travis CI issues are known: pylint problems in https://pagure.io/freeipa/issue/8016

mulatinho:
@rcritten perhaps I was misinterpreted or make a wrong assumption, this change would not allow a ipa-server-install run without a FQDN, it just checks /etc/hosts correctly on 2nd option as commented in https://github.com/freeipa/freeipa/pull/3434#issuecomment-514301697

ipaserver/install/installutils.py
```diff
@@ -164,7 +164,7 @@ def verify_fqdn(host_name, no_host_dns=False, local_hostname=True):
                          host_name)
             ex_name = socket.gethostbyaddr(host_name)
             logger.debug('Primary hostname for localhost: %s', ex_name[0])
-            if host_name != ex_name[0]:
+            if host_name not in ex_name and [host_name] not in ex_name:
```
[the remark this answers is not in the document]
mulatinho: This is exactly the problem. Thanks for reaching out, but if I change to your example, it will raise the same error if the `/etc/hosts` was like `server aliasserver server.ipadomain aliasserver.ipadomain` using **is*" to check if is in tuple would validate dispit the index of FQDN. I think I tried that, reproduce this is very easy:-)

mulatinho:
@tiran , @abbra Please check my comment about the tiran's review, I can change if you all really think that this wouldnt happen in real world. (I mean, the same IP to two or more FQDN/aliases on same line at /etc/hosts)

mulatinho:
Sorry for the late, I've tested against two another samples of `/etc/hosts`

1st sample:
```
192.168.10.1 ipaserver ipa-ad ipaserver.ipadomain ipa-ad.ipadomain
```

2nd sample:
```
192.168.10.1 ipaserver ipaserver.ipadomain ipa-ad ipa-ad.ipadomain
```

Passing FQDN of **ipaserver.ipadomain** OR **ipa-ad.ipadomain** both works fine with the change https://github.com/freeipa/freeipa/pull/3434/commits/4e7bb6f1f697339ad2db2a3e6bca7e95d4928b4f, but it raises error with the review changes proposed by @tiran. Basically because of the tuple returned by `gethostbyaddr()`. 

I will wait for your definition to rewrite the code line or leave like that. Thank you.

abbra:
@mulatinho The use case I'm trying to point to happens after the installation. 

The second field in `/etc/hosts` is called 'canonical hostname` and is used by the GNU C library in `getaddrinfo()` to return the official name of the host, as stated in `getaddrinfo()` man page. Another call that is important here is `gethostname()` which is returning the official node name as known by the Linux kernel on this machine.

The combination of these calls is used by MIT Kerberos library in a lot of places where a local hostname is required to construct a Kerberos principal. It means that, for example, host to realm translation will use `getaddrinfo()` to forward canonicalize a returned hostname from `gethostname()`. If `/etc/hosts` doesn't have FQDN set in the canonical hostname field, you will get an unexpected result:

```
[root@foobar /]# tail -1 /etc/hosts
10.88.0.108	foobar foobar.example.test
[root@foobar /]# python3
Python 3.7.2 (default, Jan 16 2019, 19:49:22) 
[GCC 8.2.1 20181215 (Red Hat 8.2.1-6)] on linux
>>> import socket
>>> socket.getaddrinfo('foobar', 0, socket.AF_UNSPEC, socket.SOCK_STREAM, 0, socket.AI_CANONNAME | socket.AI_ADDRCONFIG)
[(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, 'foobar', ('10.88.0.108', 0))]
>>> 
```

So, the host principal name will be based off `foobar`, not `foobar.example.test`. If we swap them, you'd get it working properly:

```
[root@foobar /]# tail -1 /etc/hosts
10.88.0.108	foobar.example.test foobar
[root@foobar /]# python3
Python 3.7.2 (default, Jan 16 2019, 19:49:22) 
[GCC 8.2.1 20181215 (Red Hat 8.2.1-6)] on linux
>>> import socket
>>> socket.getaddrinfo('foobar', 0, socket.AF_UNSPEC, socket.SOCK_STREAM, 0, socket.AI_CANONNAME | socket.AI_ADDRCONFIG)
[(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, 'foobar.example.test', ('10.88.0.108', 0))]
>>> 
```

Setting canonical hostname to a non-FQDN in `/etc/hosts` simply breaks a number of existing assumptions in FreeIPA:
```
$ git grep gethostname
daemons/ipa-otpd/main.c:    if (gethostname(hostname, sizeof(hostname)) < 0) {
daemons/ipa-sam/ipa_sam.c:      ret = gethostname(hostname, sizeof(hostname));
daemons/ipa-sam/ipa_sam.c:              DEBUG(1, ("gethostname failed.\n"));
daemons/ipa-sam/ipa_sam.c:      ret = gethostname(hostname, sizeof(hostname));
daemons/ipa-sam/ipa_sam.c:              DEBUG(1, ("gethostname failed.\n"));
daemons/ipa-slapi-plugins/ipa-cldap/ipa_cldap_netlogon.c:    ret = gethostname(hostname, MAXHOSTNAMELEN);
```

Same with the kernel representation of the hostname:

```
$ git grep uname
client/ipa-join.c:    uname(&uinfo);
client/ipa-join.c:        uname(&uinfo);
client/ipa-join.c:        uname(&uinfo);
client/man/ipa-join.1:The hostname of this server (FQDN). By default of nodename from uname(2) is used.
daemons/ipa-kdb/ipa_kdb.c:    struct utsname uname_data;
daemons/ipa-kdb/ipa_kdb.c:    ret = uname(&uname_data);
daemons/ipa-kdb/ipa_kdb.c:    ipactx->kdc_hostname = strdup(uname_data.nodename);
```

mulatinho:
@abbra well, I think this is enough to abandon this change... unfortunately I wont saw that and made you work on this, sorry. but what do you think about my use case? user / sysadmin tries to install ipa-server but the ipa-server-install tell that he cant because either /etc/hosts file have a problem or that is some problem in named/bind, I lost a few hours to detect that my canonical name was the problem here.

I even tried to create a zone for BIND that will respond to my host but without success; perhaps it will need to raise a better message to user/sysadmin. and why fc30 or lower are putting canonical names firstly than FQDN names?

Feel free to close this issue if you like. Thank you very much.

abbra:
@mulatinho what about detecting the case of misconfigured canonical name and suggesting a change?
We already have code that modifies `/etc/hosts` by adding the hostname there. We could also modify `/etc/hosts` for the 'wrong' canonical name.

mulatinho:
@abbra I could make this change, I will rollback this PR and just suggest the user/sysadmin to fix their /etc/hosts according to this "bug", I will try soon. Thanks.

mulatinho:
Updating PR. 

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1364139
Resolves: https://pagure.io/freeipa/issue/6210

I started to work on that change a few minutes, but I would not able to found any function that modifies `/etc/hosts` unless in one test: `./ipatests/test_integration/test_replica_promotion.py`. 

So I think I have ended with two solutions and I would love if you all could tell for me what is the best:

1st. Just change the alert text to user and make him change `/etc/hosts` accordly
```python
# check if FQDN is the first item returned by gethostbyaddr()
if host_name != ex_name[0]:
    raise HostLookupError("Please correct your /etc/hosts file using your primary FQDN (fully quallified domain name) at beginning and your aliases and canonical host names at last on that order.")
```
* Perhaps anyone could give me a better text, but I think this is more clear than what we have now.

2nd. As discussed on last comments, I will create another method on `installutils.py` that automatically correct the order of `/etc/hosts` file before the `gethostbyaddr()` with basically a code like that:

```python
# if FQDN is not the first item on the
if host_name != ex_name[0]:
    if host_name in ex_name and [host_name] in ex_name:
        # correct the order of hosts file to user
        order_hosts_file(host_name)
    else:
        raise HostLookupError("The host name %s does not match the primary host of the machine. "\
                        "Please check /etc/hosts or DNS name resolution" % host_name)
```

and the method would be something like:
```python
def order_hosts_file(host_name):
    """Read and overwrite /etc/hosts with FQDN at beginning, issue #8018"""
    normalize_host = ''.join(host_name)
    hostsfile = '/etc/hosts'
    newfile = '/tmp/hosts'

    with open(hostsfile, 'r') as hostsfp, open(newfile, 'w') as outfp:
        for line in hostsfp:
            if normalize_host in line:
                restring = re.sub(r'%s( |$)' % normalize_host, '', line.rstrip())
                line = re.sub(r'(.*?) (.*)', r'\1 %s \2\n' % normalize_host, restring.rstrip())

            outfp.write(line)

    try:
        shutil.copy(hostsfile, '%s.orig' % hostsfile)
        shutil.move(newfile, hostsfile)
    except:
        raise HostLookupError("Can't change /etc/hosts for putting FQDN after IP Address.")
```

If you all agree I will rebase the branch with one of these changes and commit & push. Thank you.

abbra:
We have `add_record_to_hosts`:

```
$ git grep add_record_to_hosts
ipaserver/install/bindinstance.py:                installutils.add_record_to_hosts(str(ip_address), self.fqdn)
ipaserver/install/installutils.py:def add_record_to_hosts(ip, host_name, conf_file=paths.HOSTS):
ipaserver/install/installutils.py:        add_record_to_hosts(str(ip_address), host_name)
```

abbra:
So, you need to modify `update_hosts_file` to add logic to modify it in case an entry is already there (right now it just skips it).

mulatinho:
So, I finally finished this issue, I made a lot of tests to guarantee that everything is OK.

I was looking into what Alexander told me and I dont think that modifying (perhaps, sorry if I am wrong) `installutils.py:update_hosts_file()` function would help with that issue, because we would need to call that function on **verify_fqdn()** and this file has no fstore, even worst at `update_hosts_file()` if the IP is at **/etc/hosts** it will continue without modify the order of file (and by modifying this function could make some problem for other places like install/server/install.py that calls **update_hosts_file()** in a different manner), I think thats not what it needs to solve that issue, what it needs is get the FQDN and IP already placed on **/etc/hosts** and rearrange them in the right order like `'10.0.0.1 server1.fqdn.com'.`

Reproduced tests, /etc/hosts samples:

sample01:
```
$ cat hosts.smp1
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6

192.168.10.105 server1 server1.ipaserver.ipadomain
```

sample02
```$ cat hosts.smp2
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6

192.168.10.105 ipa ipa.ipaserver.ipadomain ipa-ad ipa-ad.ipaserver.ipadomain
```

WITHOUT this change AND without SETTING `hostname --fqdn`, or to be more clear, by only setting `/etc/hosts` file in "wrong order", ipa-server-install WILL FAIL. I think this problem also happened on older issue at https://pagure.io/freeipa/issue/6210.

By applying this commit, original `/etc/hosts` file will be copied and the contents of hosts file will be REORDERED in a way that `ipa-server-install` WONT fail.

ipaserver/install/installutils.py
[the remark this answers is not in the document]
mulatinho: please see the last comments, thanks!

mulatinho:
Hmm, it seems that the linter was warning about long lines, but there are few other long lines on code that I dont changed, perhaps the linter was not configured by that time, may I need to fix those warnings? 

Seems diferently from what I get while running `python -m pylint ipaserver/install/installutils.py`:
https://pastebin.com/y7n6Bwqj

But I think @abba already answered that on this comment https://github.com/freeipa/freeipa/pull/3434#issuecomment-514301111

rcritten:
Linting was added when the project was already mature. Rather than changing a lot of lines just to make them compatible with the linter they were left alone and only the diffs are considered. Existing lint errors can pop up if changes are made "near" to them such that the show up in the diff. I'm guessing that's what happened in this case.

abbra:
For the record, below are errors from the linting process:
```
pycodestyle
-----------
./ipaserver/install/installutils.py:101:1: E302 expected 2 blank lines, found 1
./ipaserver/install/installutils.py:173:80: E501 line too long (85 > 79 characters)
./ipaserver/install/installutils.py:173:85: E502 the backslash is redundant between brackets
./ipaserver/install/installutils.py:267:1: E302 expected 2 blank lines, found 1
./ipaserver/install/installutils.py:276:80: E501 line too long (81 > 79 characters)
./ipaserver/install/installutils.py:277:80: E501 line too long (95 > 79 characters)
./ipaserver/install/installutils.py:285:80: E501 line too long (146 > 79 characters)
```

They need to be addressed.

mulatinho:
@rcritten thanks for the history background, I think I got now.
@abbra I fixed the whole file with according with PEP8 style, I was not using pycodestyle as much of CI. thanks for the feedback.

there was 52 warnings at total (https://pastebin.com/UaWJZDmr)

but now, it stops on pylint issues :-) I will try to deal with that too.

ipaserver/install/installutils.py
```diff
@@ -259,6 +291,32 @@ def record_in_hosts(ip, host_name=None, conf_file=paths.HOSTS):
...
+                restring = re.sub(r'%s( |$)' % normalize_host, '',
+                                  line.rstrip())
+                line = re.sub(r'(.*?) (.*)', r'\1 %s \2\n' % normalize_host,
+                              restring.rstrip())
+
+            outfp.write(line)
+
+    try:
+        shutil.copy(hostsfile, '%s.orig' % hostsfile)
+        shutil.move(newfile, hostsfile)
```
abbra: After this you need to run `tasks.restore_context(hostsfile)`, to restore SELinux context. You also need to set permissions on the file. See, for example. `ipapython.ipautil.config_replace_variables()`.

ipaserver/install/installutils.py
```diff
@@ -259,6 +291,32 @@ def record_in_hosts(ip, host_name=None, conf_file=paths.HOSTS):
 
     return None
 
+
+def order_hosts_file(host_name):
+    """Read and overwrite /etc/hosts with FQDN at beginning, issue #8018"""
+    normalize_host = ''.join(host_name)
+    hostsfile = '/etc/hosts'
+    newfile = '/tmp/hosts'
```
abbra: Please use a temporary file instead, no predicted `newfile`.

ipaserver/install/installutils.py
```diff
@@ -71,15 +71,19 @@
 class BadHostError(Exception):
     pass
 
+
```
tiran: Please don't reformat unrelated code lines. This is causing additional noise and make backports harder.

Remove the additional new lines.

ipaserver/install/installutils.py
```diff
@@ -144,10 +154,12 @@ def verify_fqdn(host_name, no_host_dns=False, local_hostname=True):
     :param local_hostname: If true, run additional checks for local hostnames
     """
     if len(host_name.split(".")) < 2 or host_name == "localhost.localdomain":
-        raise BadHostError("Invalid hostname '%s', must be fully-qualified." % host_name)
+        raise BadHostError("Invalid hostname '%s', must be fully-qualified."
```
tiran: Same here, don't reformat lines that you don't touch.

ipaserver/install/installutils.py
```diff
@@ -156,17 +168,26 @@ def verify_fqdn(host_name, no_host_dns=False, local_hostname=True):
...
         try:
             logger.debug('Check if %s is a primary hostname for localhost',
                          host_name)
             ex_name = socket.gethostbyaddr(host_name)
             logger.debug('Primary hostname for localhost: %s', ex_name[0])
+            # if FQDN is not the first item on the /etc/hosts file
             if host_name != ex_name[0]:
-                raise HostLookupError("The host name %s does not match the primary host name %s. "\
-                        "Please check /etc/hosts or DNS name resolution" % (host_name, ex_name[0]))
+                print(ex_name)
```
tiran: Please remove the debugging print statement.

ipaserver/install/installutils.py
```diff
@@ -156,17 +168,26 @@ def verify_fqdn(host_name, no_host_dns=False, local_hostname=True):
...
+                if host_name in ex_name or [host_name] in ex_name:
```
tiran: This line is hard to understand.

Please rewrite the assignment as ``ex_name, aliases = socket.gethostbyaddr(host_name)[:2]`` and then check if the hostname is either equal to ex_name or in the aliases list.

ipaserver/install/installutils.py
tiran: And create the temporary file on the same file system as the ``/etc/hosts`` so the rename can be atomic.

ipaserver/install/installutils.py
```diff
@@ -259,6 +291,32 @@ def record_in_hosts(ip, host_name=None, conf_file=paths.HOSTS):
...
     return None
 
+
+def order_hosts_file(host_name):
+    """Read and overwrite /etc/hosts with FQDN at beginning, issue #8018"""
+    normalize_host = ''.join(host_name)
+    hostsfile = '/etc/hosts'
+    newfile = '/tmp/hosts'
+
+    with open(hostsfile, 'r') as hostsfp, open(newfile, 'w') as outfp:
```
tiran: In Python we don't use ``fp`` suffix for files like in C. It's not a file pointer, it's a file object. I like ``old`` and ``new`` as names.

ipaserver/install/installutils.py
tiran: Use ``shutil.copystat()`` to copy ownership.

ipaserver/install/installutils.py
```diff
@@ -471,16 +536,21 @@ def get_host_name(no_host_dns):
...
     hostaddr = resolve_ip_addresses_nss(host_name)
     if hostaddr.intersection(
             {ipautil.UnsafeIPAddress(ip) for ip in ['127.0.0.1', '::1']}):
-        print("The hostname resolves to the localhost address (127.0.0.1/::1)", file=sys.stderr)
-        print("Please change your /etc/hosts file so that the hostname", file=sys.stderr)
-        print("resolves to the ip address of your network interface.", file=sys.stderr)
-        print("The KDC service does not listen on localhost", file=sys.stderr)
-        print("", file=sys.stderr)
-        print("Please fix your /etc/hosts file and restart the setup program", file=sys.stderr)
+        err_msg = (
```
tiran: Unrelated change, please revert. This will create conflicts with another pending patch.

ipaserver/install/installutils.py
```diff
@@ -504,11 +574,15 @@ def get_server_ip_address(host_name, unattended, setup_dns, ip_addresses):
             if set(ip_addresses) <= set(ips):
                 ips = ip_addresses
             else:
-                print("Error: the hostname resolves to IP address(es) that are different", file=sys.stderr)
```
tiran: Same here, please revert

mulatinho:
@abbra , @tiran thanks for the reviews, I will fix that in a moment, I only changed the whole file to fix PEP8 rules 'cause the CI was warning about them at the lines changed "near" as rcritten said. I tought: "why not fix the whole file instead?" I think I was wrong.

ipaserver/install/installutils.py
mulatinho: done! copystat and also restore_context

ipaserver/install/installutils.py
mulatinho: done, using tempfile.NamedTemporaryFile() now

ipaserver/install/installutils.py
mulatinho: done, i've discard all chages of PEP8 that I was making. fixing just in the lines I've changed.

ipaserver/install/installutils.py
mulatinho: done, i've discard all chages of PEP8 that I was making. fixing just in the lines I've changed.

ipaserver/install/installutils.py
mulatinho: sorry about that

ipaserver/install/installutils.py
mulatinho: thanks for the tip sir. :-)

ipaserver/install/installutils.py
mulatinho: no problem, I changed the names, in the end they all are file descriptors. thanks.

ipaserver/install/installutils.py
mulatinho: done, i've discard all chages of PEP8 that I was making. fixing just in the lines I've changed.

mulatinho:
Well, I discarded the changes to fix PEP8 problems in the file, fixing only the lines that I changed as requested by @cheimes, I've also used `tempfile.NamedTemporaryFile()`, `shutil.copystat()` and `restore_context()` as requested by @abbra and @tiran .

I made a force push again to discard the fix in file of pep8 issues, I think the CI problem was not related, but who knows? only God.. this PR is beginning to be a test of willpower hehehe :)

mulatinho:
@abbra , @tiran .. just passing to remember you that I made the changes you requested, please when you have some time review them :)

ipaserver/install/installutils.py
```diff
@@ -259,6 +270,35 @@ def record_in_hosts(ip, host_name=None, conf_file=paths.HOSTS):
 
     return None
 
+
+def order_hosts_file(host_name):
+    """Read and overwrite /etc/hosts with FQDN at beginning, issue #8018"""
```
rcritten: This seems to cry out for using ipalib.install.sysrestore, a handle just isn't currently available KRA use of the ReplicaInstall class. This would permanently (perhaps) update the file which we try not to do. It may be straightforward to use sysrestore in the KRA installer, or it may already have a handle somewhere, but this function could be simplified to use sysrestore to handle the backup and then just update the file.

mulatinho: Hmmm, so can you explain to me what you are thinking here exactly? you dont like the idea of reorder hosts file by replacing it? This thread is getting extense and probably my use case is not so relevant. Do you think that I should close this issue?

rcritten: IPA tries to be able to restore a system to the previous state. See ipaserver/install/installutils.py::update_hosts_file() for an example of updating the hosts file with a backup (in fact, perhaps this is a better place to do the re-ordering).

rcritten: We appreciate the contribution and realize that sometimes the review process can some onerous.

mulatinho: I think we (or I, lol) discuss that here: https://github.com/freeipa/freeipa/pull/3434#issuecomment-516069629. As I said there, perhaps am I wrong, you tell me.

> We appreciate the contribution and realize that sometimes the review process can some onerous.
 I am here to learn and contribute with something, I really appreciate all your feedback and learn a lot from them. :)

PR #3435, opened 2019-07-23
ipa-client-samba: remove and restore smb.conf only on first uninstall

fcami:
Fixes: https://pagure.io/freeipa/issue/8019

### DIFF
--- a/ipaclient/install/ipa_client_samba.py
+++ b/ipaclient/install/ipa_client_samba.py
@@ -433,8 +433,9 @@ def uninstall(fstore, statestore, options):
     ipautil.remove_ccache(ccache_path=paths.KRB5CC_SAMBA)
 
     # Remove samba's configuration file
-    ipautil.remove_file(paths.SMB_CONF)
-    fstore.restore_file(paths.SMB_CONF)
+    if fstore.has_file(paths.SMB_CONF):
+        ipautil.remove_file(paths.SMB_CONF)
+        fstore.restore_file(paths.SMB_CONF)
 
     # Remove samba's persistent and temporary tdb files
     tdb_files = [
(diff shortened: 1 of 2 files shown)


abbra:
@fcami thanks for the patch. It looks good to me but there is one more issue to fix:

```
    if options.uninstall:
        if statestore.has_state("domain_member"):
            uninstall(fstore, statestore, options)
....
    if statestore.has_state("domain_member") and not options.force:
        print("Samba domain member is already configured")
        return CLIENT_ALREADY_CONFIGURED
```

in the code above we call uninstallation process but don't remove the `domain_member` state from the state store. It means we are not able to re-run `ipa-client-samba -U` after the uninstallation as it would still think we are installed.

Would you please add the fix to remove the state after `uninstall()` succeeded?

fcami:
hi @abbra thanks for the comment and pointer. I'm waiting for CI to finish (and hopefully fail the test) before pushing a commit with the fix.

fcami:
Failed run without the 2nd fix: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/fb8d5c04-adf0-11e9-869f-fa163eb4190c/

fcami:
@sorlov-rh could you please review the test changes and make sure they match the bugzillas?

wladich:
I ran tests from bugzilla and they passed.

fcami:
For the record the only change is an iteration through a loop instead of multiple calls to statestore.delete_state()

fcami:
master:

* 68b85703d8ab2ede823d96e317dd106de27b108b ipatests: test multiple invocations of ipa-client-samba --uninstall
* 5b65551b3107e46853afcc4901a60ea6e661a511 ipa-client-samba: remove and restore smb.conf only on first uninstall
* ed6ee90c547cd27a3731f177dc9917176a714b49 ipatests: test ipa-client-samba after --uninstall
* cd2cbaecfce6b7b607619b34503b62c8afbbe594 ipa-client-samba: remove state on uninstall

PR #3437, opened 2019-07-24
Azure pipeline: report failure in prepare-build step

flo-renaud:
The azure pipeline defines a "prepare build" step that
installs the FreeIPA development dependencies but the
step does not report failures of the dnf builddep command.

As a consequence, subsequent steps may fail (for instance
because of components not installed such as tox) but are
hard to diagnose.

The fix reports the command failure.

Fixes: https://pagure.io/freeipa/issue/8022

### DIFF
--- a/ipatests/azure/templates/prepare-build.yml
+++ b/ipatests/azure/templates/prepare-build.yml
@@ -16,7 +16,7 @@ steps:
     sudo dnf -y module enable nodejs:12
     sudo dnf makecache || :
     echo "Installing base development environment"
-    sudo dnf install -y gdb make autoconf rpm-build gettext-devel automake libtool docker python3-paramiko python3-pyyaml || :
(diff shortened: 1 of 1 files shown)


abbra:
@flo-renaud may be we should drop all `|| :` in the `prepare-build`?

flo-renaud:
@abbra 
> may be we should drop all || : in the prepare-build?

We probably can keep it for dnf makecache and remove for dnf install. If you agree I'll update the PR in this direction.

abbra:
@flo-renaud sounds good.

flo-renaud:
@abbra 
thanks for the feedback, PR updated.

rcritten:
Changes made per @abbra , looks good to me, ack.

fcami:
master:

* 5e97e80069349d4b6a8c221f724b8f459fc3aa89 Azure pipeline: report failure in prepare-build step

PR #3448, opened 2019-07-26
user-stage: transfer all attributes from preserved to stage user

flo-renaud:
### user-stage: transfer all attributes from preserved to stage user

The user-stage command is internally implemented as:
- user_show(all=True) in order to read the user attributes
- loop on the attributes defined as possible to add using stageuser-add and
transform them into new options for stageuser_add (for instance stageuser-add
provides the option --shell for the attribute loginshell, but there is no
option for the attribute businesscategory).
- call stageuser_add in order to create a new entry in the active users subtree
- user-del to remove the previous entry in the staged users subtree
    
The issue is in the 2nd step. Only the attributes with a stageuser-add option
are processed.
The logic of the code should be slightly modified, so that all the attributes
read in the first step are processed:
- if they correspond to an option of stageuser-add, process them like it's
currently done. For instance if the entry contains displayname, then it
should be processed as --displayName=value in the stageuser-add cmd
- if they do not correspond to an option of stageuser-add, add them with
--setattr=<attrname>=<attrvalue>

Note that some attributes may need to be filtered, for instance user-show
returns has_password or has_keytab, which do not correspond to attributes
in the LDAP entry.

Fixes: https://pagure.io/freeipa/issue/7597

### xmlrpc test: add test for preserved > stage user
    
When moving a preserved user to the stage area, check that the
custom attributes are not lost ( = the attr for which there is
no specific user_stage option).
    
Test scenario:
- add a stage user with --setattr "businesscategory=value"
- activate the user, check that businesscategory is still present
- delete (preserve) the user, check that attr is still present
- stage the user, check that attr is still present
    
Related: https://pagure.io/freeipa/issue/7597

### DIFF
--- a/ipaserver/plugins/user.py
+++ b/ipaserver/plugins/user.py
@@ -921,7 +921,29 @@ class user_stage(LDAPMultiQuery):
     has_output = output.standard_multi_delete
     msg_summary = _('Staged user account "%(value)s"')
 
+    # when moving from preserved to stage, some attributes may be
+    # present in the preserved entry but cannot be provided to
+    # stageuser_add
+    # For instance: dn and uid are derived from LOGIN argument
+    #    has_keytab, has_password, preserved are virtual attributes
+    #    ipauniqueid, krbcanonicalname, sshpubkeyfp, krbextradata
+    #    are automatically generated
+    #    ipacertmapdata can only be provided with user_add_certmapdata
+    ignore_attrs = [u'dn', u'uid',
+                    u'has_keytab', u'has_password', u'preserved',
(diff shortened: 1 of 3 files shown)


rcritten:
Set labels to match ticket milestone.

abbra:
LGTM

abbra:
master:

* 27baf3501389cc3c8a12e06686da13a874d3f9da user-stage: transfer all attributes from preserved to stage user
* 8ebbb271a556ace5d1f68c404a786aef6d56ba97 xmlrpc test: add test for preserved > stage user

PR #3455, opened 2019-07-29
[WIP] Test nightly pki with updates_packages=True

flo-renaud:


flo-renaud:
Conclusion: adding `update_packages: True` in the job args correctly forces dnf update (in the task  [machine/provision : update packages]). The issue is that the vagrant image does not enable the copr repo @pki/master.

PR #3457, opened 2019-07-29
WebUI tests: Fix request timeout for test_trust

serg-cymbaluk:
Because of intergration with AD server response can take time more then 1 minute.
So request_timeout is increased to 120s.

Ticket: https://pagure.io/freeipa/issue/8024


serg-cymbaluk:
master:

* f16ea8e652a3a3b555087646b9887bc77934d175 WebUI tests: Fix request timeout for test_trust

PR #3459, opened 2019-07-29
ipatests: refactor and extend tests for IPA-Samba integration

wladich:
Relates: https://pagure.io/freeipa/issue/3999


fcami:
hi @wladich you can switch from gating.yaml to temp_commit.yaml by changing the symlink as described at the top of `prci_definitions/temp_commit.yaml` rather than editing gating.yaml directly

PR #3483, opened 2019-08-01
Support AES wrapping in LWCA key replication

frasertweedale:
The PR enhances the NSSWrappedCertDB custodia store to accept an optional
symmetric encryption algorithm OID to use for encrypting the key.  Also update
the ipa-pki-retrieve-key program to request AES wrapping.

For backwards compatibility when older servers request a key, default to 3DES
(which is what the older server supports).

For backwards compatibility when retrieving a key from an older server, try AES
first, and on HTTP 404 retry without the algorithm OID.

This change depends on Dogtag PR https://github.com/dogtagpki/pki/pull/232, and
new Dogtag release containing the change (so that we can bump the dep min
bound in FreeIPA).

Changes:

```
4afb3c3fa (Fraser Tweedale, 21 hours ago)
   ipa-pki-retrieve-key: request AES encryption (with fallback)

   Update the ipa-pki-retrieve-key client to issue a request that specifies
   that AES encryption should be used.  Fall back to a simple request (which
   will use default export algorithm) if the server returns 404.  The 404
   indicates that either:

     - It is an old server that does not support extra key arguments

     - It is a new server but the key does not exist, in which case the
      fallback request will also fail with 404.

   Fixes: https://pagure.io/freeipa/issue/8020

c5d150a39 (Fraser Tweedale, 8 days ago)
   NSSWrappedCertDB: accept optional symmetric algorithm

   Add support for specifying the desired symmetric encryption algorithm for
   exporting wrapped key (for LWCA key replication).  If not specified,
   defaults to DES-EDE3-CBC for backwards compatibility.

   Client-side changes will occur in a subsequent commit.

   Part of: https://pagure.io/freeipa/issue/8020

86ba401cc (Fraser Tweedale, 8 days ago)
   IPASecStore: support extra key arguments

   To support lightweight CA key replication using AES, while retaining 
   backwards compatibility with old servers, it is necessary to signal support
   for AES.  Whereas we currently request a key with the path:

     /keys/ca_wrapped/<nickname>

   and whereas paths with > 3 components are unsupported, add support for
   handlers to signal that they support extra arguments (defaulting to False),
   those arguments being conveyed as additional path components, e.g.:

     # 2.16.840.1.101.3.4.1.2 = aes128-cbc
    /keys/ca_wrapped/<nickname>/2.16.840.1.101.3.4.1.2

   This commit only adds the Custodia support for extra handler arguments. 
   Work to support LWCA key replication with AES wrapping will continue in
   subsequent commits.

   Part of: https://pagure.io/freeipa/issue/8020
```

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -86,9 +86,9 @@
 # Fedora
 %endif
 
-# 10.7 includes 'pki-server cert-fix' enhancements required
-# by ipa-cert-fix: https://pagure.io/freeipa/issue/7885
-%global pki_version 10.7.0-1
+# 10.7.3 supports LWCA key replication using AES
+# https://pagure.io/freeipa/issue/8020
+%global pki_version 10.7.3-1
 
 # https://pagure.io/certmonger/issue/90
 %global certmonger_version 0.79.7-1
--- a/install/tools/ipa-pki-retrieve-key.in
+++ b/install/tools/ipa-pki-retrieve-key.in
@@ -5,6 +5,8 @@ from __future__ import print_function
 import argparse
 import os
 
+from requests import HTTPError
+
 from ipalib import constants
 from ipalib.config import Env
 from ipaplatform.paths import paths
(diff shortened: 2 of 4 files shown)


frasertweedale:
This change depends on Dogtag changes, hence CI will probably fail and I will add WIP label until there is a release and I can bump the dependency in specfile.

fcami:
Adding re-run as the test_netgroup is vagrant-related. However replica_promotion indeed failed in CA.

frasertweedale:
@tiran @abbra heads up on this Custodia change.  CI is red until we have the Dogtag PR merged and a new upstream release; but I'd appreciate someone eyeballing the patch, when you have time.

simo5:
@frasertweedale I t would be nice to use a named constant to represent the OIDs in code instead of using comments, it would make the code mode readable Imo.

For example something like:
```
OID-AES128-CBC = '2.16.840.1.101.3.4.1.2'
[...]
path = f'{keyname}/{OID-AES128-CBC}'
[...]
```

frasertweedale:
@simo5 yep, that's reasonable, I'll update the PR soon.

frasertweedale:
Bumped Dogtag version in spec file, so some of the failures relate to that.  Will have to wait for new builds to appear in `updates` (currently they are in `updates-testing`)

frasertweedale:
Everything green now that new pki builds hit fedora stable.  Removing WIP label and adding needs-review.

abbra:
@frasertweedale the code looks good for me but I was confused why we have an arbitrary OID support in the NSSWrappedCert and only AES128 on the server side.

frasertweedale:
@abbra `ipa-pki-retrieve-key` is part of the client - it explicitly requests AES-128, but if the server doesn't understand the request (because it is older version) it falls back to not specifying an algorithm and will take whatever the server sends (3DES, because it is older server).

Conversely, the server has to handle older clients, so it defaults to 3DES if no algorithm was specified.  Hope that clarifies it all for you.

frasertweedale:
@abbra pushed updated commits with more explanation in comments and commit messages.  Alongside the task I have in current sprint to update Custodia design docs, is it enough for the ACK? :)

frasertweedale:
@abbra thank you!

frasertweedale:
master:

* 7e92e65190c04cc3c4699dbba329cbd6696576f1 IPASecStore: support extra key arguments
* 8fbcc3353467026e181f24cd527fe4e30f63aab4 NSSWrappedCertDB: accept optional symmetric algorithm
* bfead9ce66fb7552f9afaea18fb72a09fbe5f1c2 ipa-pki-retrieve-key: request AES encryption (with fallback)
* 7c7a827c0973669c55e4badcb6ab4f02993852a6 (HEAD) Bump Dogtag min version to 10.7.3

PR #3484, opened 2019-08-01
Restore SELinux context for p11-kit config overrides

abbra:
When 74e09087 started disabling softshm2 module in p11-kit-proxy,
we missed to restore SELinux context on the configuration override
creation.

We don't need an explicit restore_context() when removing the override
because restore_file() already calls restore_context().

### DIFF
--- a/ipaplatform/redhat/tasks.py
+++ b/ipaplatform/redhat/tasks.py
@@ -717,6 +717,7 @@ class RedHatTaskNamespace(BaseTaskNamespace):
                 # see man(5) pkcs11.conf
                 f.write("disable-in: {}\n".format(", ".join(disabled_in)))
                 os.fchmod(f.fileno(), 0o644)
+            self.restore_context(filename)
             logger.debug("Created PKCS#11 module config '%s'.", filename)
             filenames.append(filename)
 


rcritten:
So this patch seems to work ok but I discovered that restoration of an existing file doesn't work: the file is not restored. I haven't yet tracked down the reason but by the time the uninstaller gets to calling the task restore_pkcs11_modules() fstore is empty so nothing is restored.
I did confirm that the file is in the fstore after the installation so I'm not sure what happened to it.

rcritten:
I figured it out. The file is first restored by fstore.restore_all_files(). Then the task is executed and the first thing it does is unlink() an existing file. Since the copy is not in the fstore anymore the restored file is just removed.

So either the call to restore_pkcs11_modules() needs to happen before fstore.restore_all_files() or perhaps it can be dropped altogether.

tiran:
@rcritten Is there anything left to do to address the restore bug? Or can we handle the issue in another PR?

rcritten:
I guess it's not a blocker to this change. I opened https://pagure.io/freeipa/issue/8034 to track.

rcritten:
I dropped the ipa-4-7 backport because the change that added configure_pkcs11_modules() is only in 4.8:

$ git tag --contains 74e09087
rc_4-8-0-1
release-4-8-0

rcritten:
master:

* 8f969a5929e8a2f54fd732ea0c8d0a4931250d23 Restore SELinux context for p11-kit config overrides

PR #3485, opened 2019-08-01
configure.ac: don't rely on bashisms

hermannch:
93fb037d8409d9d46606c31d8a240e3963b72651 introduced unportable shell
syntax, which led to erros like
./configure: 3179: ./configure: CFLAGS+= -D__STDC_WANT_LIB_EXT1__=1: not found
in case a posix shell is used.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -19,13 +19,13 @@ AM_INIT_AUTOMAKE([foreign 1.9 tar-pax])
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES])
 
 dnl enable C11 extensions for features like memset_s()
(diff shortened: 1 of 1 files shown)


abbra:
LGTM

abbra:
Needs a full PR CI run but somehow PR CI doesn't pick this pull request.

tiran:
master:

* 9a440ae8855a5bc7b49ffe43c4a6654b17eed0a9 configure.ac: don't rely on bashisms

hermannch:
@tiran I think it's worth cherry-picking it to the 4-7 and 4-6 branches as well...

PR #3489, opened 2019-08-01
Don't return SSH keys with ipa host-find --pkey-only

rcritten:
This was introduced in 14ee02dcbd6cbb6c221ac7526e471a9fc58fcc82

https://pagure.io/freeipa/issue/8029


tiran:
master:

* 73c32dbfeb9a67a476243540dc724784b1a007aa Don't return SSH keys with ipa host-find --pkey-only

PR #3491, opened 2019-08-02
ipaserver/plugins/dns.py: add "Dynamic Update" and "Bind update policy" to default dnszone* output

fcami:
Displaying "Dynamic Update" and "Bind update policy" by default
when 'ipa dnszone-show/find' are used would make client dns update
failures easier to diagnose, so display them.

Fixes: https://pagure.io/freeipa/issue/7938


fcami:
Closing, superseded by https://github.com/freeipa/freeipa/pull/4027

PR #3500, opened 2019-08-05
Change RA agent certificate profile to caSubsystemCert

abbra:
Currently, RA agent certificate is issued using caServerCert profile.
This has unfortunate side effect of asserting id-pk-serverAuth EKU which
is not really needed for RA agent. If IPA CA certificate adds SAN DNS
constraints into issued certificates, presence of id-pk-serverAuth EKU
forces NSS (and other crypto libraries) to validate CN value with
regards to SAN DNS constraints, due to historical use of CN bearing DNS
name.

Since RA agent certificate has 'CN=IPA RA', it is guaranteed to fail
the check.

Default IPA CA configuration does *not* add SAN DNS constraints into RA
agent certificate. However, it is better to be prepared to such
behavior.

Related: https://bugzilla.redhat.com/1670239

### DIFF
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -306,6 +306,9 @@ IPA_CA_RECORD = "ipa-ca"
 IPA_CA_NICKNAME = 'caSigningCert cert-pki-ca'
 RENEWAL_CA_NAME = 'dogtag-ipa-ca-renew-agent'
 RENEWAL_REUSE_CA_NAME = 'dogtag-ipa-ca-renew-agent-reuse'
+# The RA agent cert is used for client cert authentication. In the past IPA
+# used caServerCert profile, which adds clientAuth and serverAuth EKU. The
+# serverAuth EKU caused trouble with NamedConstraints, see RHBZ#1670239.
 RA_AGENT_PROFILE = 'caSubsystemCert'
 # How long dbus clients should wait for CA certificate RPCs [seconds]
 CA_DBUS_TIMEOUT = 120
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -306,7 +306,7 @@ IPA_CA_RECORD = "ipa-ca"
 IPA_CA_NICKNAME = 'caSigningCert cert-pki-ca'
 RENEWAL_CA_NAME = 'dogtag-ipa-ca-renew-agent'
 RENEWAL_REUSE_CA_NAME = 'dogtag-ipa-ca-renew-agent-reuse'
-RA_AGENT_PROFILE = 'caServerCert'
+RA_AGENT_PROFILE = 'caSubsystemCert'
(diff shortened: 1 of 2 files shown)


abbra:
as @frasertweedale noted, caRAagentCert profile uses different authenticator. We need to create a separate profile to handle RA agent case like we did before.

abbra:
Pushed an update with caSubsystemCert which seems to be a closest one to what is needed.

ipalib/constants.py
```diff
@@ -306,7 +306,7 @@
 IPA_CA_NICKNAME = 'caSigningCert cert-pki-ca'
 RENEWAL_CA_NAME = 'dogtag-ipa-ca-renew-agent'
 RENEWAL_REUSE_CA_NAME = 'dogtag-ipa-ca-renew-agent-reuse'
-RA_AGENT_PROFILE = 'caServerCert'
+RA_AGENT_PROFILE = 'caSubsystemCert'
```
frasertweedale: I ask for two small things - some commentary here about why use `caSubsystemCert` (i.e. it doesn't have additional authenticator, and it has the desired EKU assertions), and also to remove the comment in `ipaserver/install/caintsance.py:~860` that mentions `caServerCert`.  With those changes, it is ACK from me.

frasertweedale:
The backport of this to ipa-4-7 and ipa-4-6 will depend on https://github.com/freeipa/freeipa/pull/3508 and https://github.com/freeipa/freeipa/pull/3509 (the profile-based renewal patchset, which so far is only on master and ipa-4-8).

tiran:
I have added the requested comments for @abbra.

abbra:
Do we have anything else to do here? Should we handle upgrades explicitly or just let it wait until the upgrade happens on renewal?

tiran:
I suggest to do the upgrade in another PR or just delay it until auto-renewal happens. Upgrade on renewal is good enough for almost everybody, too. Less code == less stuff that can break.

frasertweedale:
@abbra the upgrade procedure should automatically detect that the tracking request is not correct, and recreate the tracking request with correct profile.  But it will not renew the cert, only update the tracking request.  I think that is fine.  Adminstrator can just issue a `getcert resubmit` to get a new IPA RA cert with the correct profile/EKU.

frasertweedale:
And my ACK!

tiran:
master:

* 802a54bfc8b698ca962b383beb3684a4b1af1865 Change RA agent certificate profile to caSubsystemCert
* 3c82585e52c9cdcf6317bbc029935f8e339bffc6 Update comments to explain caSubsystemCert switch

rcritten:
Sadly this makes healthcheck a lot more complicated. Now there are two possible, legal profiles for a single certificate.

PR #3531, opened 2019-08-13
Don't move keys when key backup is disabled

tiran:
The CA_BACKUP_KEYS_P12 file is not enabled when pki_backup_keys is
set to False. It's the case when FreeIPA is configured with HSM support.

Related: https://pagure.io/freeipa/issue/7677


tiran:
master:

* 17c2e31fdc77183b49edc728cc775c4078bf782e Don't move keys when key backup is disabled

PR #3534, opened 2019-08-13
ipatests: add a test to install a replica when a master is unavailable

fcami:
Fixes: https://pagure.io/freeipa/issue/8039


mrizwan93:
@fcami Hello, Please see https://github.com/freeipa/freeipa/pull/2331/. I guess what you were trying to here is covered in the PR.
Can you close this PR, If not, can we move this PR further.

PR #3535, opened 2019-08-13
Allow insecure binds for migration

tiran:
Commit 5be9341fbabaf7bcb396a2ce40f17e1ccfa54b77 disallowed simple bind
over an insecure connection. Password logins were only allowed over LDAPS
or LDAP+STARTTLS. The restriction broke 'ipa migrate-ds' in some cases.

This commit lifts the restriction and permits insecure binds over plain
LDAP. It also makes the migrate-ds plugin use STARTTLS when a CA
certificate is configured with a plain LDAP connection.

Fixes: https://pagure.io/freeipa/issue/8040

### DIFF
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -1206,12 +1206,14 @@ class LDAPClient:
         return conn
 
     def simple_bind(self, bind_dn, bind_password, server_controls=None,
-                    client_controls=None):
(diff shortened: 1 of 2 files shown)


t-woerner:
LGTM

tiran:
master:

* a36556e1064900af7a75ca6f07aba66212cf321a Allow insecure binds for migration

PR #3542, opened 2019-08-15
extdom: unify error code handling especially LDAP_NO_SUCH_OBJECT

sumit-bose:
A return code LDAP_NO_SUCH_OBJECT will tell SSSD on the IPA client to
remove the searched object from the cache. As a consequence
LDAP_NO_SUCH_OBJECT should only be returned if the object really does
not exists otherwise the data of existing objects might be removed form
the cache of the clients causing unexpected behaviour like
authentication errors.

Currently some code-paths use LDAP_NO_SUCH_OBJECT as default error code.
With this patch LDAP_NO_SUCH_OBJECT is only returned if the related
lookup functions return ENOENT. Timeout related error code will lead to
LDAP_TIMELIMIT_EXCEEDED and LDAP_OPERATIONS_ERROR is used as default
error code.

Related to https://pagure.io/freeipa/issue/8044

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
+++ b/daemons/ipa-slapi-plugins/ipa-extdom-extop/back_extdom_sss_idmap.c
@@ -62,10 +62,10 @@ static enum nss_status __convert_sss_nss2nss_status(int errcode) {
         return NSS_STATUS_SUCCESS;
     case ENOENT:
         return NSS_STATUS_NOTFOUND;
-    case ETIME:
-        /* fall-through */
     case ERANGE:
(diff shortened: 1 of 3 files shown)


sumit-bose:
@abbra, please have a look at the change in `back_extdom_sss_idmap.c`, I wonder if your original fall-through for ETIME to NSS_STATUS_TRYAGAIN was intentional?

abbra:
@sumit-bose the intent for a fall-through was to re-try in case of a timeout. I think your change simply moves the re-try logic to the client which is also a valid approach.

abbra:
Changes look good but I'd really like to add a test for this.

abbra:
PR 3554 was merged, so this one can be closed.

PR #3544, opened 2019-08-16
[WIP] ipa-join: allowing call with jsonrpc into freeipa API

mulatinho:
- Adding JSON-C and LibCURL library into configure.ac and Makefile.am
- Creating a API call with option '-j' or '--jsonrpc' to make host join on FreeIPA with JSONRPC and libCURL.

TODO: unenroll process with JSONRPC.

To test the call:
```
# kinit admin
# ipa-join -s server.freeipa.ipadomain -j
```

Debug:
```
# ipa-join -s server.freeipa.ipadomain -j -d
```
Results: https://pastebin.com/0h3eQXHV

Related: https://pagure.io/freeipa/issue/7966

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -179,6 +179,42 @@ if test x"$try_xmlrpc_fallback" = xtrue; then
 fi
 
 dnl ---------------------------------------------------------------------------
+dnl - Check for LIBCURL
+dnl ---------------------------------------------------------------------------
+PKG_CHECK_MODULES([LIBCURL], [libcurl], [],
+                  [try_curl_fallback=true])
+if test x"$try_curl_fallback" = xtrue; then
+    CURL_LIBS=
+    AC_CHECK_HEADER([curl/curl.h], [],
+                    [AC_MSG_ERROR([curl.h not found])])
+
+    AC_CHECK_LIB([curl], [curl_easy_perform],
+                 [CURL_LIBS="-lcurl"])
+    if test "x$CURL_LIBS" = "x" ; then
+        AC_MSG_ERROR([libcurl not found])
+    fi
+    AC_SUBST(LIBCURL_LIBS)
+fi
+
+dnl ---------------------------------------------------------------------------
+dnl - Check for JSON-C
+dnl ---------------------------------------------------------------------------
+PKG_CHECK_MODULES([JSON], [json-c], [],
+                  [try_json_fallback=true])
+if test x"$try_json_fallback" = xtrue; then
+    JSON_LIBS=
+    AC_CHECK_HEADER([json-c/json_object.h], [],
+                    [AC_MSG_ERROR([json-c/json_object.h not found])])
+
+    AC_CHECK_LIB([json-c], [json_object_get],
+                 [JSON_LIBS="-ljson-c"])
+    if test "x$JSON_LIBS" = "x" ; then
+        AC_MSG_ERROR([json-c not found])
+    fi
+    AC_SUBST(JSON_LIBS)
+fi
+
+dnl ---------------------------------------------------------------------------
 dnl - Check for libintl
 dnl ---------------------------------------------------------------------------
 SAVE_LIBS="$LIBS"
--- a/client/Makefile.am
+++ b/client/Makefile.am
@@ -86,6 +86,8 @@ ipa_join_LDADD = 		\
 	$(LDAP_LIBS)		\
 	$(SASL_LIBS)		\
 	$(XMLRPC_LIBS)		\
+	$(JSON_LIBS)		\
+	$(LIBCURL_LIBS)		\
 	$(POPT_LIBS)		\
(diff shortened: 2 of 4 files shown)


abbra:
This looks promising, thank you, @mulatinho. Few comments.

* please add build dependencies for json-c into freeipa.spec.in
* I think we should not introduce the additional command line switches. Instead, we should be trying json-rpc first, then xml-rpc if the latter failed.
* error handling is missing in many places.
* Eventually we should remove xml-rpc support. I checked that ipa-3-0 already supports json-rpc calls on the server side.

tiran:
Good start!

* Please don't set any TLS/SSL related option. The current patch disables cert validation and sets TLS 1.2. We want to enforce cert checks and use the system policy for TLS/SSL related options.
* To simplify error checking I suggest to use macros and ``goto`` with a cleanup block. You could e.g. define a macro ``CURL_SETOPT(curl, opt, val)`` that performs setting, error checking, error reporting, and goto cleanup.
* print debug output to ``stderr``

mulatinho:
thanks for the feedback @abbra and @tiran. christian, I put TLSv1.2 fixed there mainly because it was on the xmlrpc version, but this is not a problem, I will avoid the code part which skip cert checks and the TLSv1.2 version, thanks for the tip of cleanup, I will try for sure :)

my life is not so easy now, my wife is on late pregnancy and I have a kid of two years old that loves to play :D (thanks God for all that). But I will finish this today or tomorrow and I will make the commit rebase with the changes you asked.

mulatinho:
Hello! Back again :D So to fix this issue the best is to try call with JSON RPC first and if we got an error try XML RPC? 

> Eventually we should remove xml-rpc support. I checked that ipa-3-0 already supports json-rpc calls on the server side.

Or do I need to replace the XMLRPC content?

abbra:
> Hello! Back again :D So to fix this issue the best is to try call with JSON RPC first and if we got an error try XML RPC?
> 
> > Eventually we should remove xml-rpc support. I checked that ipa-3-0 already supports json-rpc calls on the server side.
> 
> Or do I need to replace the XMLRPC content?

My thinking was to enable both (JSON RPC first to call, then XMLRPC if failed). However, make it a configure option and at some point later we can drop the option from build. And even further in future we remove XMLRPC support completely.

abbra:
I still like to get this PR merged. @mulatinho, I can help with this, if do not mind.

mulatinho:
Hey guys! Sorry for the late reply (classic), how are you doing all? I
really hope this get merged someday, I am a fan of this project and would
love to be part of it with more contributions, I will try take a look on
this, like I said before, these days I am little busy with my kids and now
a new job :)


To be honest I was almost forgot of this PR :P I would love to hear some
ideas from you all, this is working but I need to implement some checks,
and finish one other part.

tiran:
Hey,

in case you are still interested to work on the patch, then I suggest to replace the JSON parser with jansson. Its high level ``json_pack`` and ``json_unpack`` methods are such much nicer and easier to use than libjson:

```
json_t *root;
json_error_t error;
int res;
const char *host, *princ;
size_t *hostlen, *princlen;

root = json_loads(s, 0, &error);
if (root == NULL) { ... }
int = json_unpack_ex(root, &error, JSON_STRICT, "[s%, s%]", &host, &hostlen, &princ, &princlen);
if (res == -1) { ... }

strndup(host, hostlen);
strndup(princ, princlen);

json_decref(root);
```

configure.ac
```diff
@@ -178,6 +178,42 @@ if test x"$try_xmlrpc_fallback" = xtrue; then
     AC_SUBST(XMLRPC_LIBS)
 fi
 
+dnl ---------------------------------------------------------------------------
+dnl - Check for LIBCURL
+dnl ---------------------------------------------------------------------------
+PKG_CHECK_MODULES([LIBCURL], [libcurl], [],
```
tiran, line 184: You can keep it simple and assume that platforms have proper pkg-config support for necessary libraries:

```
dnl ---------------------------------------------------------------------------
dnl - Check for jansson and libcurl for ipa-join
dnl ---------------------------------------------------------------------------
PKG_CHECK_MODULES([JANSSON], [jansson])
PKG_CHECK_MODULES([LIBCURL], [libcurl])
```

Carbenium:
I gave it a shot and tried to implement the comments from @abbra and @tiran.
See https://github.com/freeipa/freeipa/compare/master...Carbenium:ipa-join-json-rpc

If this is acceptable so far and @mulatinho agrees, I could try to implement the unenrollment as well.

abbra:
@Carbenium sounds good. Please submit as a pull request so that we can discuss the proposed variant on its own.

tiran:
Thank you for your pull request. https://github.com/freeipa/freeipa/pull/4724 now has landed in master. I'm closing this PR as duplicate.

PR #3529, opened 2019-08-12
ipa-client-automount: always restore nsswitch.conf at uninstall time

fcami:
ipa-client-automount: always restore nsswitch.conf at uninstall time
    
ipa-client-automount used to only restore nsswitch.conf when sssd was not
used. However authselect's default profile is now sssd. So always
restore nsswitch.conf. This restores the behavior seen before commit:
a0e846f56c8de3b549d1d284087131da13135e34
    
Fixes: https://pagure.io/freeipa/issue/8038

PR #3547, opened 2019-08-16
Fix automount behavior with authselect

rcritten:
ipa-client-automount: always restore nsswitch.conf at uninstall time

ipa-client-automount used to only restore nsswitch.conf when sssd was not
used. However authselect's default profile is now sssd. So always
restore nsswitch.conf. This restores the behavior seen before commit:
a0e846f

Fixes: https://pagure.io/freeipa/issue/8038

This is my take on the changes required. I didn't feel comfortable pushing such large changes into someone else's branch so submitting this as a separate PR instead of #3529


rcritten:
There is a bug in authselect where the default user-nsswitch.conf has doesn't have sss as the default for automount which is otherwise the system default these days. Not sure if we want to try to work around that or wait for a fix.

fcami:
hi @rcritten with the latest code changes the upstream issue is fixed.
The test can be marked xfail with https://pagure.io/freeipa/issue/8054 which is a separate issue.

fcami:
Code fix LGTM, fully tested. 
Integration test covers both sssd and non-sssd scenarios.
One (related, but not from original issue) bug noted in ipa-client-install: https://pagure.io/freeipa/issue/8054

rcritten:
Removing temp commit

rcritten:
master:

* 405dcc6becfca504dce7e06c6f0849a9c06df4c6 ipatests: check that ipa-client-automount restores nsswitch.conf at uninstall time
* b27ad6e9f956a2485eee09b647b45c4901a1f928 ipa-client-automount: always restore nsswitch.conf at uninstall time
* e5af8c19a9e40fb3b96c56ace081f79980437fc2 Move ipachangeconf from ipaclient.install to ipapython
* 41ef8fba31ddbb32e2e5b7cccdc9b582a0809111 Use tasks to configure automount nsswitch settings

PR #3562, opened 2019-08-19
[ipa-4-7] prci: Update box used in branch ipa-4-7

netoarmando:
Replace template box used, from ci-master-f29 to ci-ipa-4-7-f29.

This affects the cached packages since it's using ipa-4-7 .spec file.

Based on the changes made in https://github.com/freeipa/freeipa-pr-ci/pull/304.


abbra:
One of sudo rules 'failed' -- it is passed while actually was supposed to fail. This is because SSSD was finally fixed. I guess we need to update the test definition:

```
    @pytest.mark.xfail(
        osinfo.id == 'fedora' and osinfo.version_number < (30,),
        reason="https://pagure.io/SSSD/sssd/issue/3957", strict=True)
    def test_domain_resolution_order(self):
        """Test sudo with runAsUser and domain resolution order.

```

In the previous image we had sssd-2.1.0-2.fc29, new image has sssd-2.2.0-3.fc29.

abbra:
I've pushed an additional commit that removes xfail mark.

abbra:
ipa-4-7:

* db305621fcaf46606cd8a9078f96dd19048140d1 prci: Update box used in branch ipa-4-7
* ce2dcd7b6188416c5c7024786e258a524eb98288 Update sudo test as SSSD 2.2.0 is available in the test image

PR #3563, opened 2019-08-19
 WebUI: Make 'Unlock' option is available only on locked user page 

serg-cymbaluk:
The implementation includes checking password policy for selected user.
'Unlock' option is available only in case user reached a limit of login failures.

Ticket: https://pagure.io/freeipa/issue/5062


pvoborni:
So far read only the code. It looks mostly good. Was is tested as a almost normal user which has some arbitrary role assigned (thus does not see self-service but see admin interface)? I wonder if it will still work in case when a user cannot read password policies or some other attrs.

serg-cymbaluk:
> So far read only the code. It looks mostly good. Was is tested as a almost normal user which has some arbitrary role assigned (thus does not see self-service but see admin interface)? I wonder if it will still work in case when a user cannot read password policies or some other attrs.

@pvoborni

I've add password policy and other attributes checking. Also I've changed the implementation to use already fetched data.

netoarmando:
Looks good to me, I've tested your build locally, it behaved as expected and I believe @pvoborni's concerns were addressed. 
Please remove the temp commit and I'll 'ack' this.

serg-cymbaluk:
@netoarmando
Thanks for the review. Temp commit is removed.

flo-renaud:
master:

* 123c93f92c575eeb4344f918b86112937b8cb4e8 WebUI: Make 'Unlock' option is available only on locked user page

flo-renaud:
Hi @serg-cymbaluk 
A manual backport is required on ipa-4-7 branch, could you handle the backport? Thanks

PR #3589, opened 2019-08-28
ipatests: fix wrong xfail in test_domain_resolution_order

flo-renaud:
The test is written for a SSSD fix delivered in 2.2.0, but has a xfail
based on fedora version < 30.
SSSD 2.2.0 was originally available only on fedora 30 but is now also
available on fedora 29, and recent runs on f29 started to succeed
(because the fix is now present) but with a strict xfail.

The fix considers the sssd version instead of the fedora version to
mark xfail.

Fixes: https://pagure.io/freeipa/issue/8052

### DIFF
--- a/ipatests/test_integration/test_sudo.py
+++ b/ipatests/test_integration/test_sudo.py
@@ -19,7 +19,6 @@
 
 import pytest
 
-from ipaplatform.osinfo import osinfo
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration.tasks import (
     clear_sssd_cache, get_host_ip_with_hostmask, modify_sssd_conf)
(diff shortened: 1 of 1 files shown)


flo-renaud:
Marking for 4.7 and 4.8 as the test was committed to both branches (see PR #3232).

wladich:
I have a concern about this PR: the xfail now is not "strict", and even not a real xfail -- the test is not even executed if sssd version is not suitable, acting like a "skip".

flo-renaud:
@wladich 
I pushed a new version which completely removes the xfail marker, since the master branch is supported on fedora 29+ which contains the SSSD fix.
The fix needs does not need a backport to ipa-4-7 as it is already present in this branch, please see commit [ce2dcd7](https://pagure.io/freeipa/c/ce2dcd7b6188416c5c7024786e258a524eb98288?branch=ipa-4-7), but can be backported to ipa-4-8.

wladich:
LGTM

flo-renaud:
master:

* b48fe19fe58f139e6c0702cd0e0eda65ae0e231d ipatests: fix wrong xfail in test_domain_resolution_order

PR #3601, opened 2019-08-30
Replace %{_libdir} macro in BuildRequires

tiran:
The %{_libdir} macro is architecture dependend and therefore does not
correctly work across different platforms. In the past the SRPM was
created on a platform with /usr/lib64. Recent SRPMs have /usr/lib, which
breaks dnf builddep.

Depend on krb5-server directly rather than a file in krb5-server
package:

$ rpm -qf /usr/lib64/krb5/plugins/kdb/db2.so
krb5-server-1.16.1-25.fc29.x86_64

Fixes: https://pagure.io/freeipa/issue/8056

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -264,7 +264,7 @@ BuildRequires:  python3-yubico
 %if ! %{ONLY_CLIENT}
 BuildRequires:  libcmocka-devel
 # Required by ipa_kdb_tests
-BuildRequires:  %{_libdir}/krb5/plugins/kdb/db2.so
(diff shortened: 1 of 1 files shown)


frozencemetery:
In the future I would like to do away with `%{krb5_version}`, but this is fine for now.

fcami:
Acking after checking the change and reading the associated bz.

fcami:
master:

* 51836c0594d9f97ace8392c03d47aae883b87724 Replace %{_libdir} macro in BuildRequires

PR #3602, opened 2019-08-30
ipatests: add tests for cached_auth_timeout in sssd.conf

wladich:
The tests check that auth cache
    * is disabled by default
    * is working when enabled
    * expires after specified time
    * is inherited by trusted domain
    
    Related to: https://bugzilla.redhat.com/1685581

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -31,6 +31,8 @@ import itertools
 import tempfile
 import time
 from pipes import quote
+import configparser
+from contextlib import contextmanager
 
 import dns
 from ldif import LDIFWriter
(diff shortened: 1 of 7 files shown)


flo-renaud:
@wladich 
I am adding the postponed label to this PR to temporarily prevent it from triggering PRCI.
It seems that PRCI is unable to handle the RunADTests class and crashes because of this.

ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -53,14 +57,14 @@ jobs:
         timeout: 1800
         topology: *build
 
-  fedora-30/temp_commit:
+  fedora-30/test_sssd:
     requires: [fedora-30/build]
     priority: 50
     job:
-      class: RunPytest
+      class: RunADtests
```
flo-renaud: should be:
```
     class: RunADTests
```
with an uppercase T after AD. Otherwise PRCI crashes

wladich: Thanks for comment

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -1872,3 +1873,57 @@ def run_command_as_user(host, user, command, *args, **kwargs):
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.restore()
+
+
+class RemoteIniFile(configparser.RawConfigParser):
+    """Wrap RawConfigParser for modifying .ini files on remote hosts.
```
flo-renaud: This class does not really inherit the RawConfigParser class. For instance read() and read_file() methods probably won't work unless you override them.
I am wondering whether object composition wouldn't be a better fit here.

wladich:
Changes done in new commit:
* updated nightly_master_testing.yaml with new test suite
* replaced tasks.RemoteIniFile with more simple remote_ini_file contexyt manager
* updated docstring for tasks.FileBackup
* added --preserve=context in FileBackup constructor and restore()

ipatests/pytest_ipa/integration/tasks.py
wladich: I replaced this class with a simple context manager.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -1872,3 +1873,57 @@ def run_command_as_user(host, user, command, *args, **kwargs):
...
+    def __init__(self, host, filename):
+        """Create file backup."""
+        self._host = host
+        self._filename = filename
+        self._backup = create_temp_file(host)
+        host.run_command(['cp', filename, self._backup])
+
+    def restore(self):
+        """Restore file. Can be called multiple times."""
+        self._host.run_command(['cp', self._backup, self._filename])
```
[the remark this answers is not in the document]
wladich: Good catch, fixed.

flo-renaud:
Hi @wladich 
Thanks for the PR, LGTM. You can remove the temp commit and I will ACK once PRCI completes.

wladich:
I have noticed that latest changes to tasks.py were added to the wrong commit. 
I rearranged changes by commits, overall diff did not change.

flo-renaud:
Tests are green, ACK.

wladich:
master:

* 7dde3a422067c82b67eab197a0da3d6f149c1b3c ipatests: add new utilities for file management
* 4ea9aead5cab96ed9454facdb6d22bdfe514a09b ipatests: refactoring: use library function to check if selinux is enabled
* 4ab2842b76ffc9cffd931bf652d211ca34b4d1d8 ipatests: add tests for cached_auth_timeout in sssd.conf

PR #3610, opened 2019-09-02
WebUI: Fix new test initialization on "HBAC Test" page

serg-cymbaluk:
"New Test" action cleared only information about selected options but kept
radio buttons checked. It confused users and caused an error on validation step.

New behaviour is:
- tables forget all selected values after "New Test" click;
- first table record is checked initially in case the option is mandatory;
- all records is unchecked initially in case the option is not mandatory.

Ticket: https://pagure.io/freeipa/issue/8031

### DIFF
--- a/install/ui/src/freeipa/hbactest.js
+++ b/install/ui/src/freeipa/hbactest.js
@@ -263,12 +263,28 @@ IPA.hbac.test_select_facet = function(spec) {
         that.table.multivalued = false;
 
         that.table.set_values = function(values) {
-            if (values && values.length && values[0] === '__external__') {
-                if (that.external_radio) that.external_radio.prop('checked', true);
-            } else {
-                that.table.table_set_values(values);
+            var records = that.table.records;
+            var has_values = values && values.length;
+
+            if (has_values && values[0] === '__external__') {
+                if (that.external_radio) {
+                    that.external_radio.prop('checked', true);
+                }
+                return;
             }
+
+            that.table.table_set_values(values);
         };
+    };
+
+    that.show = function() {
+        that.facet_show();
+        that._make_sure_value_selected();
+    };
+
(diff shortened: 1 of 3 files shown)


install/ui/src/freeipa/hbactest.js
```diff
@@ -263,14 +263,30 @@ IPA.hbac.test_select_facet = function(spec) {
...
+            that.table.table_set_values(values);
         };
     };
 
+    that.show = function() {
+        that.facet_show();
+        that._make_sure_value_selected();
+    };
+
+    that.load_records = function(records) {
```
netoarmando: These two methods, `show` and `load_records`, don't appear to be called explicitly. Are they used implicitly anywhere? Or are they just helpers to be used in the future??

netoarmando:
@serg-cymbaluk Besides my comment, everything worked as you described and I think it makes sense in the UI. 

Do you think a new test scenario is required after your changes?

netoarmando:
> I fixed existing test. But I found this feature is weakly covered, so, yes, I'd write new test scenario for it.

Are you planning to include new tests into this PR?

serg-cymbaluk:
@netoarmando 

> 
> Are you planning to include new tests into this PR?

I going to finish this in the scope of some task for automation, so I wouldn't implement new tests there.

netoarmando:
> I going to finish this in the scope of some task for automation, so I wouldn't implement new tests there.

OK! Thanks, please remove the temp commit and I will ack it.

serg-cymbaluk:
@netoarmando 
Thanks for the review.
Temp commit is removed.

netoarmando:
master:

* 4dbc6926b16b44146eef36d2f56403e2347f12c8 WebUI: Fix new test initialization on "HBAC Test" page

PR #3616, opened 2019-09-03
extdom: plugin doesn't allow @ in group name

thalman:
Old implementation handles username and group names with
one common call. Character @ is used in the call to detect UPN.

Group name can legaly contain this character and therefore the
common approach doesn't work in such case.

Also the original call is less efficient because it tries to resolv
username allways then it fallback to group resolution.

Here we implement two new separate calls for resolving users and
groups.

Fixes: https://bugzilla.redhat.com/1746951

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
+++ b/daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
@@ -114,6 +114,13 @@ int __nss_to_err(enum nss_status errcode)
     }
 }
 
+static int get_timeout(struct ipa_extdom_ctx *ctx) {
+    if (ctx == NULL || ctx->nss_ctx == NULL) {
+        return DEFAULT_MAX_NSS_TIMEOUT;
+    }
+    return back_extdom_get_timeout(ctx->nss_ctx);
+}
+
 int getpwnam_r_wrapper(struct ipa_extdom_ctx *ctx, const char *name,
                        struct passwd *pwd, char **buf, size_t *buf_len)
 {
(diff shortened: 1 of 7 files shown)


rcritten:
Add re-run flag to execute full CI

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -271,7 +271,9 @@ int parse_request_data(struct berval *req_val, struct extdom_req **_req)
  *        sid (1),
  *        name (2),
  *        posix uid (3),
- *        posix gid (3)
+ *        posix gid (4),
```
abbra: Could you please also add a comment that explains difference between V0/V1/V2 versions of the sequence. We need to have it documented somewhere. May be even add a document in `doc/designs/extdom-plugin.md`? Nothing formal, just explain what it does and how various versions of the control are structured.

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
```diff
@@ -1189,104 +1230,156 @@ static int handle_name_request(struct ipa_extdom_ctx *ctx,
...
-             * error codes which can indicate that the user was not found. To
-             * be on the safe side we fail back to the group lookup on all
-             * errors. */
-            ret = getgrnam_r_wrapper(ctx, fq_name, &grp, &buf, &buf_len);
-            if (ret != 0) {
+    ret = getpwnam_r_wrapper(ctx, fq_name, &pwd, &buf, &buf_len);
+    switch(ret) {
+    case 0:
+        if (request_type == REQ_FULL_WITH_GROUPS) {
+            ret = sss_nss_getorigbyname(pwd.pw_name, &kv_list, &id_type);
```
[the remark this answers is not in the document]
thalman: @abbra, thank you for the review. I updated this call and as you can see in 840c583adc4e88167211bada5f77d0cdcf965b1a there are many more.

daemons/ipa-slapi-plugins/ipa-extdom-extop/ipa_extdom_common.c
thalman: Added doc/designs/extdom-plugin-protocol.md. @sumit-bose since you are the original author, could you be so kind and check it?

abbra:
Looks good. I had to rerun Azure tests several times as unrelated DNS test was producing a failure.

I think the documentation is OK for now and the rest of fixes is fine too.

abbra:
master:

* e5f04258b5b3fb6c04c28ddd38ae251c822e80bc extdom: plugin doesn't allow @ in group name
* 5f898c3c614f4165f0eb15c3aad2157689fbbcfe extdom: plugin doesn't use timeout in blocking call
* 84b6c0f53b9ebdd4c01181898499bb6992aa9e8a extdom: use sss_nss_*_timeout calls
* bddf64b9da2df21a14022109ae989bd5408bf14b extdom: add extdom protocol documentation

PR #3619, opened 2019-09-04
WebUI: Fix changing category on HBAC/Sudo/etc Rule pages

serg-cymbaluk:
No object can be added to a rule when object category is 'all'.
So while editing rule there is needed to save actual category value before adding related objects.

Ticket: https://pagure.io/freeipa/issue/7961

### DIFF
--- a/ipatests/test_webui/test_selinuxusermap.py
+++ b/ipatests/test_webui/test_selinuxusermap.py
@@ -53,6 +53,7 @@ INVALID_MLS = ("invalid 'selinuxuser': Invalid MLS value, must match {}, "
 HBAC_DEL_ERR = ('{} cannot be deleted because SELinux User Map {} requires '
                 'it')
 HBAC_MEMBER_ERR = 'HBAC rule and local members cannot both be set'
+BATCH_ERR = 'Some operations failed.'
 
 
 @pytest.mark.tier1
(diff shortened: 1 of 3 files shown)


ipatests/test_webui/test_selinuxusermap.py
```diff
@@ -53,6 +53,7 @@
 HBAC_DEL_ERR = ('{} cannot be deleted because SELinux User Map {} requires '
                 'it')
 HBAC_MEMBER_ERR = 'HBAC rule and local members cannot both be set'
+BATCH_ERR = 'Some operations failed.'
```
netoarmando: Every other message doesn't end with a period.

serg-cymbaluk:
@netoarmando 

> You have updated one existing test case; however, I think a test case is missing, for when 'all' is the current selection and the user switches that and tries to add an item to the list. Am I right?

Yes, we need to cover this issue, but not in the scope of the PR.

netoarmando:
Ok! Looks good to me. I tested locally, the issue is gone and the new behaviour was clear from the user's perspective.
Please, remove the temp commit and I'll ack it.

serg-cymbaluk:
@netoarmando 
Thanks for the review.
Temp commit is removed.

serg-cymbaluk:
master:

* 755154318ac6b6a57eee2bdec76debd25e97fd45 WebUI: Fix changing category on HBAC/Sudo/etc Rule pages

PR #3622, opened 2019-09-04
Fix two error paths around ipadb_get_global_config()

frozencemetery:


rcritten:
master:

* 9414b038e714043d87dc0646dd9af3933896a75a Log INFO message when LDAP connection fails on startup
* 45b4f5377bf3921406271148e18a3b99acfee03b Fix NULL pointer dereference in maybe_require_preauth()

PR #3624, opened 2019-09-05
ipatests: extdom plugin should not return error (32)/'No such object'

amore17:
Added test which validates :
If there is a time-out during a request to SSSD the extdom plugin
should not return error 'No such object' and the existing user should
not be removed from the cache on the client.

Related Ticket:  https://pagure.io/freeipa/issue/8044


ipatests/test_integration/test_trust.py
```diff
@@ -268,6 +252,64 @@ def test_override_homedir(self):
...
             tasks.clear_sssd_cache(self.master)
 
+    def test_extdom_plugin(self):
+        """Extdom plugin should not return error (32)/'No such object'
+
+        Regression test for https://pagure.io/freeipa/issue/8044
+
+        If there is a timeout during a request to SSSD the extdom plugin
+        should not return error 'No such object' and the existing user should
+        not be removed from the cache on the client.
```
wladich: According to BZ, the user was added to negative cache, not removed.

wladich:
Commit message is also misleading about removing user from cache.

ipatests/test_integration/test_trust.py
```diff
@@ -268,6 +253,65 @@ def test_override_homedir(self):
...
+        tasks.backup_file(self.master, paths.SSSD_CONF)
+        log_file = '{0}/sssd_{1}.log' .format(paths.VAR_LOG_SSSD_DIR,
+                                              self.master.domain.name)
+        logsize = len(self.master.get_file_contents(log_file))
+        res = self.master.run_command(['pidof', 'sssd_be'])
+        pid = res.stdout_text.strip()
+        try:
+            tasks.kinit_admin(self.master)
+            # set the extdom plugin timeout to 1s (1000) by default it is 10s
+            entry_ldif = textwrap.dedent("""
```
wladich: The code will look cleaner if you use master.ldap_connect and methods get_entry/update_entry to read and modify data. You can find example in method disable_autounmembership_in_ldap() in  test_automember.py

ipatests/test_integration/test_trust.py
```diff
@@ -268,6 +253,65 @@ def test_override_homedir(self):
...
+            tasks.ldapmodify_dm(self.master, entry_ldif)
+            domain = self.master.domain
+            tasks.modify_sssd_conf(
+                self.master,
+                domain.name,
+                {
+                    'timeout': '999999'
+                }
+            )
+            test_id = 'id testuser@%s' % self.ad_domain
```
wladich: Is there really a reason to use AD user for the test?

ipatests/test_integration/test_trust.py
```diff
@@ -268,6 +253,65 @@ def test_override_homedir(self):
...
+            client.run_command(test_id)
+            remove_cache = 'sss_cache -E'
+            self.master.run_command(remove_cache)
+            client.run_command(remove_cache)
+
+            self.master.run_command('kill -STOP %s' % pid)
```
wladich: The test scenario and especially this part with kill -STOP/CONT is not trivial. Please provide comments or expand docstring with explanations what and why you are doing.

ipatests/test_integration/test_trust.py
amore17: IdM servers use the ipa-extdom to receive information about AD users and groups and forwards them to the requesting client. 
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/windows_integration_guide/troubleshooting-cross-forest-trusts#troubleshooting-the-ipa-extdom-plug-in

ipatests/test_integration/test_trust.py
```diff
@@ -95,6 +90,11 @@ def remove_trust(self, ad):
 
 class TestTrust(BaseTestTrust):
 
+    dn = DN(
```
wladich: 1. The variable name tells nothing about its contents.
2. It does not need to be class member as it is used in a single test case

ipatests/test_integration/test_trust.py
```diff
@@ -268,6 +259,65 @@ def test_override_homedir(self):
...
+
+        Regression test for https://pagure.io/freeipa/issue/8044
+
+        If there is a timeout during a request to SSSD the extdom plugin
+        should not return error 'No such object' and the existing user should
+        not be added to negative cache on the client.
+        """
+        client = self.clients[0]
+        tasks.backup_file(self.master, paths.SSSD_CONF)
+        log_file = '{0}/sssd_{1}.log' .format(paths.VAR_LOG_SSSD_DIR,
```
wladich: Please do not add space before dot

ipatests/test_integration/test_trust.py
```diff
@@ -268,6 +259,65 @@ def test_override_homedir(self):
...
+            client.run_command(test_id)
+
+            error = 'ldap_extended_operation result: No such object(32)'
+            sssd_log2 = self.master.get_file_contents(log_file)[logsize:]
+            assert error.encode() not in sssd_log2
+        finally:
+            if stop_sssd_be.returncode == 0:
+                self.master.run_command('kill -CONT %s' % pid)
+            # set back to default extdom plugin
+            entry_ldif = textwrap.dedent("""
```
wladich: The code will look cleaner and more consistent if you use same approach here as in test setup (get_entry/update_entry). 
Another question: previously you were updating field ipaExtdomMaxNssTimeout with value 10000 here. Now you delete it.Can you please explain why you changed it?

ipatests/test_integration/test_trust.py
```diff
@@ -268,6 +259,65 @@ def test_override_homedir(self):
...
+        tasks.backup_file(self.master, paths.SSSD_CONF)
+        log_file = '{0}/sssd_{1}.log' .format(paths.VAR_LOG_SSSD_DIR,
+                                              self.master.domain.name)
+        logsize = len(self.master.get_file_contents(log_file))
+        res = self.master.run_command(['pidof', 'sssd_be'])
+        pid = res.stdout_text.strip()
+        try:
+            test_id = 'id testuser@%s' % self.ad_domain
+            self.master.run_command(test_id)
+            tasks.kinit_admin(self.master)
```
wladich: I am not quite sure but it looks useless as you are not executing any `ipa` commands.

ipatests/test_integration/test_trust.py
```diff
@@ -268,6 +259,65 @@ def test_override_homedir(self):
...
+            self.master.run_command(['ipactl', 'restart'])
+            domain = self.master.domain
+            tasks.modify_sssd_conf(
+                self.master,
+                domain.name,
+                {
+                    'timeout': '999999'
+                }
+            )
+            client.run_command(test_id)
```
wladich: This is useless as you have already checked it and it is in sssd cache.

ipatests/test_integration/test_trust.py
amore17: Yes as per discussion with @sumit-bose  to set it default time-out this approach can be also used .

wladich: The test cleanup should revert the system to original state, not to "original-equivalent". We want to minimize influence of this test on the following ones. Even if it is equivalent now it does not mean it will be always so - due to regression or intentional change. So if this value is defined before the test, we should keep it an restore here. If it is not defined before the test, we should remove it.

ipatests/test_integration/test_trust.py
```diff
@@ -268,6 +253,69 @@ def test_override_homedir(self):
...
+    def test_extdom_plugin(self):
+        """Extdom plugin should not return error (32)/'No such object'
+
+        Regression test for https://pagure.io/freeipa/issue/8044
+
+        If there is a timeout during a request to SSSD the extdom plugin
+        should not return error 'No such object' and the existing user should
+        not be added to negative cache on the client.
+        """
+        search_dn = DN(
```
wladich: Please try to avoid generic names like this one. Try to name variables according to their contents or function.
In this particular case it can be called "extdom_config_dn"

ipatests/test_integration/test_trust.py
```diff
@@ -268,6 +253,69 @@ def test_override_homedir(self):
...
+                                             client.domain.name)
+        logsize = len(client.get_file_contents(log_file))
+        res = self.master.run_command(['pidof', 'sssd_be'])
+        pid = res.stdout_text.strip()
+        test_id = 'id testuser@%s' % self.ad_domain
+        client.run_command(test_id)
+
+        conn = self.master.ldap_connect()
+        entry = conn.get_entry(search_dn)  # pylint: disable=no-member
+        default = entry.single_value.get('ipaextdommaxnsstimeout')
```
wladich: Same here: the variable contains original value of the timeout, so a better name would be "orig_extdom_timeout"

amore17:
Result : http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/52b86c8e-da11-11e9-b52a-fa163eb4190c/report.html

PR #3625, opened 2019-09-05
[ipa-4-6 only] check for single-label domains only during server install

flo-renaud:
The fix for https://pagure.io/freeipa/issue/7207 and
https://pagure.io/freeipa/issue/7598 added checks against single-label
domains in client, server and replica installs. This prevents client
enrollment to existing topologies with single-label domain.

This commit removes those fixes on ipa-4-6 branch. Server installation
with single-label domain will still be refused, but client enrollment
will succeed.

Fixes: https://pagure.io/freeipa/issue/8058

### DIFF
--- a/ipalib/util.py
+++ b/ipalib/util.py
@@ -406,14 +406,15 @@ def validate_dns_label(dns_label, allow_underscore=False, allow_slash=False):
 
 def validate_domain_name(
     domain_name, allow_underscore=False,
-    allow_slash=False, entity='domain'
+    allow_slash=False, entity='domain',
+    check_sld=False
 ):
     if domain_name.endswith('.'):
         domain_name = domain_name[:-1]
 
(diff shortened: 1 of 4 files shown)


fcami:
With this build, enrolling a client whose hostname is: f27-ipa0.ipa against an older (ipa-server-4.6.4-10.el7_6.3) ipa-server with autodetected domain works.

fcami:
Enrolling clients using --server, --domain, --realm parameters works too.

fcami:
ipa-server-install fails with:
```
ipapython.admintool: ERROR    Invalid domain name: single label domains are not supported
```
. and that's expected

rcritten:
ipa-4-6:

* 8ae6c1af1e6ef25fdfbbf7e72265372366e6b106 check for single-label domains only during server install

PR #3635, opened 2019-09-09
adtrust: avoid using timestamp in klist output

abbra:
When parsing a keytab to copy keys to a different keytab, we don't need
the timestamp, so don't ask klist to output it. In some locales (en_IN,
for example), the timestamp is output in a single field without a space
between date and time. In other locales it can be represented with date
and time separated by a space.

Fixes: https://pagure.io/freeipa/issue/8066

### DIFF
--- a/ipaserver/install/plugins/adtrust.py
+++ b/ipaserver/install/plugins/adtrust.py
@@ -721,7 +721,7 @@ class update_tdo_to_new_layout(Updater):
 
 
 KeyEntry = namedtuple('KeyEntry',
-                      ['kvno', 'date', 'time', 'principal', 'etype', 'key'])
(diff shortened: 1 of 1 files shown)


abbra:
I found out that MIT Kerberos when formatting a timestamp attempts to find a format string that produces the shortest output given a buffer length. A list of format strings includes two ISO8601 variants which get triggered in en_IN locale. See https://pagure.io/freeipa/issue/8066 for more details.

t-woerner:
LGTM

abbra:
master:

* 80e4c18b75af989d5839aba7e6a1efc06da16f81 adtrust: avoid using timestamp in klist output

PR #3646, opened 2019-09-10
Add container environment check to replicainstall

Tiboris:
Inside the container environment master's IP address does not resolve to its name.

Resolves: https://pagure.io/freeipa/issue/6210

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -904,7 +904,11 @@ def promote_check(installer):
                            "certificate")
 
     installutils.verify_fqdn(config.host_name, options.no_host_dns)
-    installutils.verify_fqdn(config.master_host_name, options.no_host_dns)
+    # Inside the container environment master's IP address does not
+    # resolve to its name. See https://pagure.io/freeipa/issue/6210
+    container_environment = tasks.detect_container() is not None
+    installutils.verify_fqdn(config.master_host_name, options.no_host_dns,
+                             local_hostname=not container_environment)
 
     ccache = os.environ['KRB5CCNAME']
     kinit_keytab('host/{env.host}@{env.realm}'.format(env=api.env),


ipaserver/install/server/replicainstall.py
```diff
@@ -904,7 +904,11 @@ def promote_check(installer):
                            "certificate")
 
     installutils.verify_fqdn(config.host_name, options.no_host_dns)
-    installutils.verify_fqdn(config.master_host_name, options.no_host_dns)
+    # Inside the container environment master's IP address does not
+    # resolve to its name. See bz #1364139
+    container_environment = tasks.detect_container()
+    installutils.verify_fqdn(config.master_host_name, options.no_host_dns,
+                             local_hostname=not container_environment)
```
rcritten: I'm not sure the test is right. detect_container() returns None if not a container but a string value if it does (e.g. container-other). So perhaps:

container_environment = tasks.detect_container() is not None

Tiboris: Oh yes this or `bool(tasks.detect_container())` You decide.

ipaserver/install/server/replicainstall.py
```diff
@@ -904,7 +904,11 @@ def promote_check(installer):
                            "certificate")
 
     installutils.verify_fqdn(config.host_name, options.no_host_dns)
-    installutils.verify_fqdn(config.master_host_name, options.no_host_dns)
+    # Inside the container environment master's IP address does not
+    # resolve to its name. See bz #1364139  z
```
pvoborni: Might be better to refer to https://pagure.io/freeipa/issue/6210 As the BZ is not available for most people. Also, the z at the end should not be here.

rcritten:
Looks good.

Tiboris:
master:

* f1e20b45c5deeb25989c87a2d717bda5a31bb084 Add container environment check to replicainstall

PR #3649, opened 2019-09-11
[Backport][ipa-4-6] Don't configure KEYRING ccache in containers

Tiboris:
Manual backport of PR #2677

environment.

Signed-off-by: Christian Heimes cheimes@redhat.com

### DIFF
--- a/ipatests/test_ipaplatform/test_tasks.py
+++ b/ipatests/test_ipaplatform/test_tasks.py
@@ -1,0 +1,32 @@
+#
+# Copyright (C) 2017  FreeIPA Contributors see COPYING for license
+#
+from __future__ import absolute_import
+
+import os
+
+from ipaplatform.tasks import tasks
+
+
+def test_detect_container():
+    container = None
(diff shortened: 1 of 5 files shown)


ipatests/test_ipaplatform/test_tasks.py
```diff
@@ -0,0 +1,56 @@
...
+# Copyright (C) 2017  FreeIPA Contributors see COPYING for license
+#
+from __future__ import absolute_import
+
+import os
+
+from ipaplatform.tasks import tasks
+
+
+def test_ipa_version():
```
flo-renaud: This test is not needed in the backport, it has no relationship with the fix. I think you can simply remove it.

Tiboris: Sure!

Tiboris:
ipa-4-6:

* 91e54057f130f0c2d9da8506e34c3cadc9cd9c6e Don't configure KEYRING ccache in containers

PR #3664, opened 2019-09-13
replica install: enforce --server arg

flo-renaud:
When the --server option is provided to ipa-replica-install (1-step
install), make sure that the server offers all the required roles
(CA, KRA). If it's not the case, refuse the installation.

Note that the --server option is ignored when promoting from client to
replica (2-step install with ipa-client-install and ipa-replica-install),
meaning that the existing behavior is not changed in this use case:
by default the host specified in default.conf as server is used for
enrollment, but if it does not provide a required role, another host can
be picked for CA or KRA setup.

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -789,6 +789,8 @@ def promote_check(installer):
             print("IPA client is already configured on this system, ignoring "
                   "the --domain, --server, --realm, --hostname, --password "
                   "and --keytab options.")
+            # Make sure options.server is not used
+            options.server = None
 
         # The NTP configuration can not be touched on pre-installed client:
         if options.no_ntp or options.ntp_servers or options.ntp_pool:
@@ -1039,8 +1041,15 @@ def promote_check(installer):
             config.subject_base = DN(subject_base)
 
         # Find any server with a CA
+        # The order of preference is
+        # 1. the first server specified in --server, if any
+        # 2. the server specified in the config file
+        # 3. any other
+        preferred_cas = [config.ca_host_name]
+        if options.server:
+            preferred_cas.insert(0, options.server)
         ca_host = find_providing_server(
-            'CA', conn, [config.ca_host_name]
+            'CA', conn, preferred_cas
         )
         if ca_host is not None:
             config.ca_host_name = ca_host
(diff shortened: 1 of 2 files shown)


ipaserver/install/server/replicainstall.py
```diff
@@ -1049,6 +1058,14 @@ def promote_check(installer):
...
                              "CA is present on some master.")
                 raise ScriptError(rval=3)
+            if options.setup_ca and options.server and \
+               ca_host != options.server:
+                # Installer was provided with a specific master
+                # but this one doesn't provide CA
+                logger.error("The specified --server %s does not provide CA, "
+                             "please provide a server with the CA role",
+                             options.server)
+                raise ScriptError(rval=3)
```
rcritten: 3 is documented to mean:

```
3 if the host exists in the IPA server or a replication agreement to the remote
master already exists
```
A new return value should probably be defined.

rcritten:
Otherwise this looks good.

ipaserver/install/server/replicainstall.py
flo-renaud: Hi @rcritten 
thanks for the suggestion. Exit code 4 will be used for this specific error, man page also updated.

ipaserver/install/server/replicainstall.py
```diff
@@ -1049,6 +1058,14 @@ def promote_check(installer):
                 logger.error("Certificates could not be provided when "
                              "CA is present on some master.")
                 raise ScriptError(rval=3)
+            if options.setup_ca and options.server and \
+               ca_host != options.server:
+                # Installer was provided with a specific master
+                # but this one doesn't provide CA
+                logger.error("The specified --server %s does not provide CA, "
+                             "please provide a server with the CA role",
```
mrizwan93: Hello @flo-renaud 
Can we use some file to place the stderr/stdout messages? something like `ipaplatform.paths`, so that testsuite can import this file and use the stderr/stdout from it.
I have seen failures in test due to change in stdout/stderr messages(downstream observation though).

flo-renaud: Hi @mrizwan93 
The error messages are not something we commit to (as opposed to the exit codes for instance, or the APIs), and I am reluctant to define constants for them. This would lead to a nightmare finding a different constant name for each message.
If you prefer having a test that doesn't rely on the expected message, I would suggest checking the exit code instead (the value 4 will be documented for this specific case).

rcritten:
I'm inclined to ack this unless @mrizwan93 has any more concerns.

ipaserver/install/server/replicainstall.py
mrizwan93: Okay, relying on returncode make sense.

flo-renaud:
master:

* 802e54dd0e33be6015b22853767fc37a9ec02f39 replica install: enforce --server arg

PR #3672, opened 2019-09-16
Make sure to have storage space for tag

simo5:
ber_scanf expects a pointer to a ber_tag_t to return the tag pointed at
by "t", if that is not provided the pointer will be store in whatever
memory location is pointed by the stack at that time causeing a crash.

Note that this is effectively unused code because in ipa-kdb the only
party that can write a key_data structure to be stored is te kdb_driver
itself and we never encode these s2kparam data.

But we need to handle this for future proofing.

Fixes: https://pagure.io/freeipa/issue/8071


tiran:
master:

* 4abd2f76d76c4c1a1ec5087ec447f4515b63c2c6 Make sure to have storage space for tag

PR #3675, opened 2019-09-17
Re-order tasks.restore_pkcs11_modules() to run earlier

rcritten:
It was executed after restore_all_files() so PKCS11_MODULES was
already restored so that part was a no-op, but the redhat
restore_pkcs11_modules() also calls unlink() on each restored
file so basically the file would be restored, unlinked, then
since it was already restored, skipped.

By moving the call to restore_pkcs11_modules() earlier it can
do the expected restoration properly.

https://pagure.io/freeipa/issue/8034

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -1145,6 +1145,7 @@ def uninstall(installer):
     custodiainstance.CustodiaInstance(realm='REALM.INVALID').uninstall()
     otpdinstance.OtpdInstance().uninstall()
     tasks.restore_hostname(fstore, sstore)
+    tasks.restore_pkcs11_modules(fstore)
     fstore.restore_all_files()
     try:
         os.remove(paths.ROOT_IPA_CACHE)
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @rcritten 
thanks for the PR, works for me:
- create a file /etc/pkcs11/modules/softhsm2.module
- install server
- the file is overwritten
- uninstall server
- the original file is restored.

While testing I discovered that ipa-backup/ipa-restore does not backup the file /etc/pkcs11/modules/softhsm2.module. Issue tracked as [8073](https://pagure.io/freeipa/issue/8073), which can be handled in a different fix.

flo-renaud:
Adding the ipa-4-7 label as the module files were added also in ipa-4-7 branch (commit 74e0908 + backported to ipa-4-7 with ae5831b)

rcritten:
master:

* ffb4b624fc8bfd870958976ec25b74677e740841 Re-order tasks.restore_pkcs11_modules() to run earlier

PR #3677, opened 2019-09-17
Don't log host passwords when they are set/modified

rcritten:
The host password was defined as a Str type so would be
logged in cleartext in the Apache log.

A new class, HostPassword, was defined to only override
safe_value() so it always returns an obfuscated value.

The Password class cannot be used because it has special treatment
in the frontend to manage prompting and specifically doesn't
allow a value to be passed into it. This breaks backwards
compatibility with older clients. Since this class is derived
from Str old clients treat it as a plain string value.

https://pagure.io/freeipa/issue/8017


flo-renaud:
Hi @rcritten 
thanks for the PR, works for me. I tested
- [x] ipa host-add --password PWD
- [x] ipa host-mod --password PWD
and both commands log the sanitized version of the password in debug and info log in /var/log/httpd/error_log (`userpassword='********'`)

flo-renaud:
Note that I also checked that an older client is still able to call ipa host-add / host-mod --password.

rcritten:
master:

* 48a3f4af46517c7dbaddad7e2c5d4e82a6ef2f33 Don't log host passwords when they are set/modified

PR #3693, opened 2019-09-19
Report if a certmonger CA is missing

rcritten:
If a certmonger CA is not defined but is referenced within
a request (so was removed sometime after a request was
created) then anything that pulls all certmonger requests would
fail with the cryptic error:

"Failed to get request: bus, object_path and dbus_interface
must not be None."

This was often seen during upgrades.

Catch this specific condition and report a more specific error
so the user will have some bread crumb to know how to address
the issue.

https://pagure.io/freeipa/issue/7870

### DIFF
--- a/ipalib/install/certmonger.py
+++ b/ipalib/install/certmonger.py
@@ -179,6 +179,9 @@ def _get_requests(criteria=dict()):
         for criterion in criteria:
             if criterion == 'ca-name':
                 ca_path = request.obj_if.get_ca()
+                if ca_path is None:
+                    raise RuntimeError("certmonger CA '%s' is not defined" %
+                                       criteria.get('ca-name'))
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @rcritten 
thanks for the PR, works for me. I tested the following scenario:
- remove the ca 'IPA' with `getcert remove-ca -c IPA`
- run upgrade with `ipa-server-upgrade`
- the upgrade fails as expected and print a hint at the end:
```[Update certmonger certificate renewal configuration]
Failed to get request: certmonger CA 'IPA' is not defined
IPA server upgrade failed: Inspect /var/log/ipaupgrade.log and run command ipa-server-upgrade manually.
certmonger CA 'IPA' is not defined
The ipa-server-upgrade command failed. See /var/log/ipaupgrade.log for more information
```

rcritten:
master:

* 5b28c458b91c545d089da6271da0927e7c91ccdd Report if a certmonger CA is missing

PR #3695, opened 2019-09-20
[Backport][ipa-4-7] Extdom plugin should not return error (32)/'No su…

amore17:
…ch object'

This is manual backport of : https://github.com/freeipa/freeipa/pull/3624

Regression test for https://pagure.io/freeipa/issue/8044

If there is a timeout during a request to SSSD the extdom plugin
should not return error 'No such object' and the existing
user should not be added to negative cache on the client.


wladich:
ipa-4-7:

* 83e3f5d105a8d52328fbd966bf82404760c51668 Extdom plugin should not return error (32)/'No such object'

PR #3696, opened 2019-09-20
[Backport][ipa-4-6] Extdom plugin should not return error (32)/'No su…

amore17:
…ch object'

This is manual backport of : #3624

Regression test for https://pagure.io/freeipa/issue/8044

If there is a timeout during a request to SSSD the extdom plugin
should not return error 'No such object' and the existing
user should not be added to negative cache on the client.


wladich:
ipa-4-6:

* 17536af58b5a2d1ae1adf7e741dade7b3f84179a Extdom plugin should not return error (32)/'No such object'

PR #3721, opened 2019-09-23
Don't create log files from help scripts

tiran:
Helper scripts now use api.bootstrap(log=None) to avoid the creation of
log files. Helper scripts are typically executed from daemons which
perform their own logging. The helpers still log to stderr/stdout.

This also gets rid of some SELinux AVCs when the script tries to write
to /root/.ipa/.

Fixes: https://pagure.io/freeipa/issue/8075

### DIFF
--- a/install/certmonger/dogtag-ipa-ca-renew-agent-submit.in
+++ b/install/certmonger/dogtag-ipa-ca-renew-agent-submit.in
@@ -477,7 +477,9 @@ def main():
     if operation not in ('SUBMIT', 'POLL'):
         return OPERATION_NOT_SUPPORTED_BY_HELPER
 
-    api.bootstrap(in_server=True, context='renew', confdir=paths.ETC_IPA)
+    api.bootstrap(
+        in_server=True, context='renew', confdir=paths.ETC_IPA, log=None
+    )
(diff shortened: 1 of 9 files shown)


rcritten:
To expand on this, is there ever a reason to have log enabled when in_server=True? Perhaps we can disable the log by default in that context.

tiran:
I think so. Some server scripts like update, backup, and restore use in_server and have logging enabled.

rcritten:
That's true but I think this is different. logging has changed so many times over the years I can't recall though and haven't yet traced it. The log in question is the dot_ipa log which we don't want in a server mode (and perhaps others). I think this is configured separately from command-specific logging.

tiran:
Ah yes, there is some special code for logging in a server environment. ``in_server=True`` also changes the location of the log directory (api.env.logdir) from ``~/.ipa`` to ``/var/log/ipa``.

abbra:
So, for `in_server=True` cases we don't need a change here.
Otherwise, LGTM.

tiran:
master:

* 90f72324549f2bceba3e051efb2a1b43c467ff8a Don't create log files from help scripts

PR #3724, opened 2019-09-24
 ipatests: Added testcase to check logrotate is added to ipahealthcheck tool

menonsudhir:
ipatests: Added testcase to check logrotate is added to ipahealthcheck tool
 Issue: freeipa/freeipa-healthcheck#35

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -101,6 +101,7 @@ DEFAULT_PKI_CA_CERTS = [
     "auditSigningCert cert-pki-ca",
     "Server-Cert cert-pki-ca",
 ]
+
 DEFAULT_PKI_KRA_CERTS = [
     "transportCert cert-pki-kra",
     "storageCert cert-pki-kra",
@@ -401,6 +402,28 @@ class TestIpaHealthCheck(IntegrationTest):
         assert returncode == 0
         for check in data:
             assert check["result"] == "SUCCESS"
+
+    def test_ipa_healthcheck_log_rotate_file_exist_issue35(self):
+        """
+        This test checks if log rotation has been added
+        for ipa-healthcheck tool so that logs are rotated
+        in /var/log/ipa/healthcheck folder.
+        The test also checks that the logrotate configuration
+        file is syntactically correct by calling logrotate --debug
+        This is a testcase for below pagure issue
+        https://github.com/freeipa/freeipa-healthcheck/issues/35
+        """
+        msg = "error: {}:".format(HEALTHCHECK_LOG_ROTATE_CONF)
+        tasks.uninstall_packages(self.master, HEALTHCHECK_PKG)
+        assert not self.master.transport.file_exists(
+            HEALTHCHECK_LOG_ROTATE_CONF
+        )
+        tasks.install_packages(self.master, HEALTHCHECK_PKG)
+        assert self.master.transport.file_exists(HEALTHCHECK_LOG_ROTATE_CONF)
+        cmd = self.master.run_command(
+            ['logrotate', '--debug', HEALTHCHECK_LOG_ROTATE_CONF]
+        )
+        assert msg not in cmd.stdout_text
 
     def test_ipa_healthcheck_remove(self):
         """


ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -53,14 +53,14 @@ jobs:
...
     job:
       class: RunPytest
       args:
         build_url: '{fedora-30/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
+        test_suite: test_integration/test_commands.py
         template: *ci-master-f30
         timeout: 3600
-        topology: *master_1repl_1client
+        topology: *master_1client
```
flo-renaud: This topology does not exist, you should use master_1repl instead.
Full explanation: the topology specified in the yaml file with `topology: *<topology_name>` must be defined in the same yaml file in the `topologies:` section. The char & is used to denote an anchor and the char * is used to denote an alias, ie avoid typing twice the same content.

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +679,51 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+        3. Start ipa-healthcheck.timer service
+        4. Check /var/log/ipa/healthcheck directory for log file
+        """
+        log_file = '/var/log/ipa/healthcheck/healthcheck.log'
+        systemd_healthcheck_file = '/etc/systemd/system/' \
+                                   'multi-user.target.wants/' \
+                                   'ipa-healthcheck.timer'
+        # Install ipa-healthcheck tool
+        tasks.kinit_admin(self.master)
+        result = self.master.run_command(['dnf', 'install', '-y',
```
flo-renaud: dnf command will not be available on RHEL 7.x. You can use `which dnf >/dev/null` and check the return code.
I would advise to add a new helper function in ipatests/pytest_ipa/integration/tasks.py, for instance
`def install_package(host, pkgs)`.

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +679,51 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+                                          'freeipa-healthcheck'],
+                                         raiseonerr=False)
+        assert result.returncode == 0
```
flo-renaud: You can simplify the code by removing raiseonerr=False and `assert result.returncode == 0`.
When raiseonerr=True, a failing command will raise an exception and stop the test.

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +679,51 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+                                raiseonerr=False)
+        if os.path.islink(systemd_healthcheck_file):
+            print("Symlink ", systemd_healthcheck_file, " exists")
+        else:
+            print("Symlink ", systemd_healthcheck_file, " missing")
+        # Start ipa-healthcheck.timer service
+        result = self.master.run_command(['systemctl', 'start',
+                                          'ipa-healthcheck.timer'],
+                                         raiseonerr=False)
+        assert result.returncode == 0
```
flo-renaud: You can simplify by removing raiseonerr=False.

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +679,51 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+        tasks.kinit_admin(self.master)
+        result = self.master.run_command(['dnf', 'install', '-y',
+                                          'freeipa-healthcheck'],
+                                         raiseonerr=False)
+        assert result.returncode == 0
+        # Enable ipa-healthcheck.timer service
+        self.master.run_command(['systemctl', 'enable',
+                                 'ipa-healthcheck.timer'],
+                                raiseonerr=False)
+        if os.path.islink(systemd_healthcheck_file):
```
flo-renaud: This command would be executed on the controller, you need to execute on the master instead.

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +679,51 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+                                          'ipa-healthcheck.timer'],
+                                         raiseonerr=False)
+        assert result.returncode == 0
+        result = self.master.run_command(['systemctl', 'status',
+                                          'ipa-healthcheck.timer'],
+                                         raiseonerr=False)
+        assert 'Started Execute IPA Healthcheck every day at 4AM' \
+               in result.stdout_text
+        # Check log file name in /var/log/ipa/healthcheck directory
+        if os.path.exists(log_file):
```
flo-renaud: This command would be executed on the controller, you need to run on the master instead.

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +679,51 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+            print("LogFile ", log_file, " exists")
+        else:
+            print("LogFile ", log_file, " doesnot exist")
```
flo-renaud: The test does not raise any exception when the log file is missing?

ipatests/prci_definitions/temp_commit.yaml
menonsudhir: topology has been modified to  topology: *master_1repl

ipatests/test_integration/test_commands.py
```diff
@@ -919,3 +919,30 @@ def test_cacert_manage(self):
...
+        2. Check logrotate file is present for healthcheck tool
+        3. Check contents of the ipahealthcheck file
+        """
+        HEALTHCHECK_PKG = ['freeipa-healthcheck']
+        log_file = '/var/log/ipa/healthcheck/healthcheck.log'
+        logrotate_file = '/etc/logrotate.d/ipahealthcheck'
+        tasks.install_packages(self.master, HEALTHCHECK_PKG)
+        if self.master.transport.file_exists(logrotate_file):
+            self.master.get_file_contents(logrotate_file)
+            with open(logrotate_file, 'r') as f:
```
mrizwan93: you can replace this with:
```
                contents = self.master.get_file_contents(logrotate_file)
                assert log_file in contents
                assert 'create' in contents
                assert 'daily' in contents
                assert 'dateext' in contents
                assert 'misingok' in contents
                assert 'rotate 30' in contents
```

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -413,6 +413,30 @@ def test_source_ipa_roles_check_crlmanager(self):
...
+            contents = self.master.get_file_contents(
+                HEALTHCHECK_LOG_ROTATE_CONF, encoding="utf-8"
+            )
+            assert HEALTHCHECK_LOG in contents
+            assert "create" in contents
+            assert "daily" in contents
+            assert "dateext" in contents
+            assert "missingok" in contents
+            assert "rotate 30" in contents
+
```
amore17: Instead of multiple assert statements, you can do like this:
check = ['create', 'daily', 'dateext', 'missingok', 'rotate 30']
for i in check:
   assert i in contents

rcritten:
What is the intention of checking the config file? To prevent a bad config from getting created or to require this very specific one?

To check for syntax errors you could run:
# logrotate --debug  /etc/logrotate.d/ipahealthcheck

and look for error: in the output.

For example:
error: ipahealthcheck:6 unknown option 'foo' -- ignoring line

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -402,6 +403,60 @@ def test_ipa_healthcheck_dna_plugin_returns_warning_pagure_issue_60(self):
...
+        """
+        tasks.uninstall_packages(self.master, HEALTHCHECK_PKG)
+        assert (
+            self.master.transport.file_exists(HEALTHCHECK_LOG_ROTATE_CONF)
+            is False
+        )
+        tasks.install_packages(self.master, HEALTHCHECK_PKG)
+        assert (
+            self.master.transport.file_exists(HEALTHCHECK_LOG_ROTATE_CONF)
+            is True
```
mrizwan93: No need of `is True` here. Assert statement will take care of it.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -402,6 +403,60 @@ def test_ipa_healthcheck_dna_plugin_returns_warning_pagure_issue_60(self):
...
+        This test checks if log rotation has been added
+        for ipa-healthcheck tool so that logs are rotated
+        in /var/log/ipa/healthcheck folder.
+        This is a testcase for below pagure issue
+        https://github.com/freeipa/freeipa-healthcheck/issues/35
+        """
+        tasks.uninstall_packages(self.master, HEALTHCHECK_PKG)
+        assert (
+            self.master.transport.file_exists(HEALTHCHECK_LOG_ROTATE_CONF)
+            is False
```
flo-renaud: This can be simplified as 
`assert not self.master.transport.file_exists(HEALTHCHECK_LOG_ROTATE_CONF)`

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -402,6 +403,60 @@ def test_ipa_healthcheck_dna_plugin_returns_warning_pagure_issue_60(self):
...
+        )
+        tasks.install_packages(self.master, HEALTHCHECK_PKG)
+        assert (
+            self.master.transport.file_exists(HEALTHCHECK_LOG_ROTATE_CONF)
+            is True
+        )
+        contents = self.master.get_file_contents(
+            HEALTHCHECK_LOG_ROTATE_CONF, encoding="utf-8"
+        )
+        start_pattern = re.compile(
```
flo-renaud: Missing `import re`

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -402,6 +403,27 @@ def test_ipa_healthcheck_dna_plugin_returns_warning_pagure_issue_60(self):
...
+        tasks.uninstall_packages(self.master, HEALTHCHECK_PKG)
+        assert not self.master.transport.file_exists(
+            HEALTHCHECK_LOG_ROTATE_CONF
+        )
+        tasks.install_packages(self.master, HEALTHCHECK_PKG)
+        assert self.master.transport.file_exists(HEALTHCHECK_LOG_ROTATE_CONF)
+        cmd = self.master.run_command(
+            ["logrotate", "--debug", HEALTHCHECK_LOG_ROTATE_CONF]
+        )
+        assert cmd.returncode == 0
```
flo-renaud: This line is not needed as run_command is using raiseonerr=True by default and would raise an exception if the returncode is not 0.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -402,6 +403,27 @@ def test_ipa_healthcheck_dna_plugin_returns_warning_pagure_issue_60(self):
...
 
+    def test_ipa_healthcheck_log_rotate_file_exist_issue35(self):
+        """
+        This test checks if log rotation has been added
+        for ipa-healthcheck tool so that logs are rotated
+        in /var/log/ipa/healthcheck folder.
+        This is a testcase for below pagure issue
+        https://github.com/freeipa/freeipa-healthcheck/issues/35
+        """
+        msg = "error: /etc/logrotate.d/ipahealthcheck:"
```
flo-renaud: You can use the constant HEALTHCHECK_LOG_ROTATE_CONF:
msg = "error: {}:".format(HEALTHCHECK_LOG_ROTATE_CONF)

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -402,6 +403,27 @@ def test_ipa_healthcheck_dna_plugin_returns_warning_pagure_issue_60(self):
         for check in data:
             assert check["result"] == "SUCCESS"
 
+    def test_ipa_healthcheck_log_rotate_file_exist_issue35(self):
+        """
+        This test checks if log rotation has been added
+        for ipa-healthcheck tool so that logs are rotated
+        in /var/log/ipa/healthcheck folder.
```
flo-renaud: You can add "It also ensures that the logrotate configuration file is syntactically correct by calling logrotate --debug".

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: Replacing single quotes with double quotes was not done by pylint. Revered the change.
Also removed assert statement here.

flo-renaud:
Hi @menonsudhir 
thanks for the PR, LGTM. You can remove the temp commit and I will ACK when PRCI completes.

flo-renaud:
@menonsudhir 
do we need to backport this PR to older branches? Please set the corresponding labels if needed.

flo-renaud:
master:

* c77f4213e9ae78e76c5eee88ffc1cbffbb5556e8 ipatests: Added testcase to check logrotate is added for healthcheck tool

PR #3727, opened 2019-09-24
Fix errors newly exposed by pylint 2.4.0

stanislavlevin:
Newest Pylint introduced additional checks [1]:

- import-outside-toplevel [2]

> This check warns when modules are imported from places other
than a module toplevel, e.g. inside a function or a class.

- no-else-continue [3]

> These checks highlight unnecessary else and elif blocks after
break and continue statements.

- unnecessary-comprehension [4]

> This check is emitted when pylint finds list-, set- or
dict-comprehensions, that are unnecessary and can be rewritten
with the list-, set- or dict-constructors.

[1] https://github.com/PyCQA/pylint/blob/pylint-2.4.0/doc/whatsnew/2.4.rst
[2] https://github.com/PyCQA/pylint/issues/3067
[3] https://github.com/PyCQA/pylint/issues/2327
[4] https://github.com/PyCQA/pylint/issues/2905

Fixes: https://pagure.io/freeipa/issue/8077
Signed-off-by: Stanislav Levin <slev@altlinux.org>

PR #3728, opened 2019-09-24
Disable dogtag cert publishing

rcritten:
Dogtag had only one switch, ca.publish.enable, for both CRLs and certs.

Since cert publishing is not used in IPA it should be disabled to
avoid false positives in the logs.

https://pagure.io/freeipa/issue/7522


frasertweedale:
LGTM; the CI failures seem related to pylint so we have to wait for https://github.com/freeipa/freeipa/pull/3727, then rebase.

abbra:
LGTM.

rcritten:
master:

* c8b0d1d6a036d55f4e3bcac996c14371614105d8 Disable dogtag cert publishing

PR #3287, opened 2019-06-18
[Backport][ipa-4-7] Wrap CustodiaClient in context manager

frasertweedale:
Manual backport of #3262

PR #3731, opened 2019-09-25
[Backport][ipa-4-6] Wrap CustodiaClient in context manager

frasertweedale:
Backport of https://github.com/freeipa/freeipa/pull/3287.

A CustodiaClient object has to the process environment a bit, e.g. set
up GSSAPI credentials. To reuse the credentials in libldap connections,
it is also necessary to set up a custom ccache store and configure
KRBCCNAME.

Credential initialization and ccache is now handled in a context
manager.

https://pagure.io/freeipa/issue/7964
https://bugzilla.redhat.com/show_bug.cgi?id=1755223

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -49,8 +49,8 @@
 
 %global alt_name ipa
 %if 0%{?rhel}
-# 1.15.1-7: certauth (http://krbdev.mit.edu/rt/Ticket/Display.html?id=8561)
-%global krb5_version 1.15.1-4
+# 1.15.1-36: https://bugzilla.redhat.com/show_bug.cgi?id=1755223
+%global krb5_version 1.15.1-36
 # 0.7.16: https://github.com/drkjam/netaddr/issues/71
 %global python_netaddr_version 0.7.5-8
 # Require 4.7.0 which brings Python 3 bindings
--- a/install/tools/ipa-pki-retrieve-key
+++ b/install/tools/ipa-pki-retrieve-key
@@ -2,9 +2,8 @@
 
 from __future__ import print_function
 
+import argparse
 import os
-import sys
-import traceback
 
 from ipalib import constants
 from ipalib.config import Env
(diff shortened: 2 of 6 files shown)


frasertweedale:
Adding needs-review but this was a trivial backport... I just don't want to auto-ACK my own changes...

rcritten:
The commits in the PR look a bit strange. It includes the pylint fix plus one titled "CustodiaClient: fix IPASecStore config on ipa-4-7" which is confusing on the 4.6 branch.

The krb5 version is also a bit confusing since 1.15.1-36 is RHEL-specific AFAICT. That would still pull in 1.15.2-something but it would be confusing if anyone went searching for a release.

frasertweedale:
@rcritten thanks for reviewing.

Would you prefer I limited the version bump to the `%if 0%{?rhel}` block, and leave the `%else` alone?

The ipa-4-7 mention in commit message is because the original backport that branch (before the backport to ipa-4-6 became necessary) needed some extra work.  I could amend the commit message.  But because it is a clean backport I would prefer to leave it as-is.

And yeah, the pylint fixes.  I might extract that to a separate PR.

rcritten:
You convinced me on the rhel conditional, it's fine. Everything looks a-ok, ack.

frasertweedale:
ipa-4-6:

* 82a9fe7e655115befbdde10907a5aa7669c35fde Handle missing LWCA certificate or chain
* 436214aea7fd5893525292cb03b3c28cdbc249f2 Fix CustodiaClient ccache handling
* 1f455867f82407c0dfab0b9f123c75ca0d1a0090 CustodiaClient: use ldapi when ldap_uri not specified
* c9d0ba0c355c433ae883cafa3c1e99fea1a85220 CustodiaClient: fix IPASecStore config on ipa-4-7
* e686949dcdc46486061d23d5e18f21e2a2038f58 (HEAD) Bump krb5 min version

PR #3732, opened 2019-09-25
ipatests: File ownership and permission for dirsrv log instance after restore

mrizwan93:
Check if file ownership and permission is set to dirsrv:dirsrv
and 770 on /var/log/dirsrv/slapd-<instance> after ipa-restore.

related ticket : https://pagure.io/freeipa/issue/7725

### DIFF
--- a/ipatests/test_integration/test_backup_and_restore.py
+++ b/ipatests/test_integration/test_backup_and_restore.py
@@ -28,6 +28,7 @@ import pytest
 from ipaplatform.constants import constants
 from ipaplatform.paths import paths
 from ipaplatform.tasks import tasks as platformtasks
+from ipapython.ipaldap import realm_to_serverid
 from ipapython.dn import DN
 from ipapython import ipautil
 from ipatests.test_integration.base import IntegrationTest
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -217,6 +217,15 @@ def test_full_backup_and_restore(self):
             self.master.run_command(['ipa-restore', backup_path],
                                     stdin_text=dirman_password + '\nyes')
 
+            # check the file permssion and ownership is set to 770 and
+            # dirsrv:dirsrv after restore on /var/log/dirsrv/slapd-<instance>
+            # related ticket : https://pagure.io/freeipa/issue/7725
+            dashed_domain = self.master.domain.realm.replace(".", '-')
```
rcritten: There is a function to convert a realm into the dirsrv instance name, please use that instead:

from ipapython.ipaldap import realm_to_serverid
...
instance = realm_to_serverid(self.master.domain.realm)
log_path = paths.VAR_LOG_DIRSRV_INSTANCE_TEMPLATE % instance
...

rcritten:
Thanks, looks good. Please remove the temporary commit and I'll add the ack.

rcritten:
master:

* 7aec6f10374129d75cd99a4012980516ca535551 Check file ownership and permission for dirsrv log instance

PR #3734, opened 2019-09-25
ipatests: fix fedora29 nightly definition

flo-renaud:
test_sssd is using a wrong dependency (fedora30 build instead
of fedora29 build). As a result, this test is not triggered
by PRCI because it's waiting forever for a dependency.
(See the status: fedora-30/test_sssd Pending — unassigned)

Fix the version in the fedora 29 nightly definition.


rcritten:
master:

* 6127aa0eac5bba37decc02fbffb3da891c81a9d1 ipatests: fix fedora29 nightly definition

PR #3739, opened 2019-09-26
ipa user_add: do not check group if UPG is disabled

flo-renaud:
### ipa user_add: do not check group if UPG is disabled

The UPG plugin is used to create a user private group when a new IPA user is created, with the same name as the user. When this plugin is enabled, the user creation must ensure that no group exists with the same name.

When the UPG plugin is disabled, or when the user is created with the --noprivate option, there is no need to perform this check as the private group will not get created.

Currently, the --noprivate option correctly skips the test, but a disabled UPG plugin does not skip the test. The fix ensures that UPG plugin status is checked.
    
Fixes: https://pagure.io/freeipa/issue/4972

### ipatests: add XMLRPC test for user-add when UPG plugin is disabled
    
Add a new XMLRPC test in test_user_plugin:
- disable the UPG plugin
- create a user without the --gid parameter
  as the default group for new users is not POSIX (ipausers), the command is expected to fail
- create a user with the --gid parameter
  The provided gid is used and command is expected to succeed
- create a user with the same name as an existing group
  As the UPG plugin is disabled, the user creation will not trigger the creation of a group with the same name, and command is expected to succeed
- re-enable the UPG plugin for other tests
    
Related to: https://pagure.io/freeipa/issue/4972

### DIFF
--- a/ipatests/test_xmlrpc/test_user_plugin.py
+++ b/ipatests/test_xmlrpc/test_user_plugin.py
@@ -32,6 +32,7 @@ import re
 
 from ipalib import api, errors
 from ipaplatform.constants import constants as platformconstants
+from ipapython import ipautil
 from ipatests.test_xmlrpc import objectclasses
 from ipatests.util import (
     assert_deepequal, assert_equal, assert_not_equal, raises)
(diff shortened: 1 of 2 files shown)


ipatests/test_xmlrpc/test_user_plugin.py
```diff
@@ -786,6 +787,86 @@ def test_change_default_user_group_back(self, user_npg, user_npg2):
...
+        del testuser.attrs['mepmanagedentry']
+        testuser.attrs.update(gidnumber=[u'1000'])
+        testuser.attrs.update(
+            description=[],
+            objectclass=add_oc(objectclasses.user_base, u'ipantuserattrs')
+        )
+        command = testuser.make_create_command()
+        result = command()
+        testuser.check_create(result, [u'description'])
+        testuser.delete()
```
rcritten: Do you need to remove the group as well or is that cleaned up automatically?

rcritten:
Looks good, just one in-line question.

ipatests/test_xmlrpc/test_user_plugin.py
flo-renaud: The group fixture is defined with `scope=class`, meaning that the group will be deleted as soon as the last test of TestUserWithUPGDisabled is complete.

flo-renaud:
master:

* b2a2d7f4e4c63bf60c276b1fb210c72f1ebc4b9f ipa user_add: do not check group if UPG is disabled
* 387ee6e60f398efcdaf91c8097794ed24179150d ipatests: add XMLRPC test for user-add when UPG plugin is disabled

PR #3760, opened 2019-10-01
[ipa-4-8] prci: increase timeout for jobs that required AD

netoarmando:
Vagrant retries to provision hosts if something happens, it was introduced
in PR-CI after https://github.com/freeipa/freeipa-pr-ci/commit/380c8b8c78a1ce277b7c1a327bda9d123c117c4d.

This takes time, some jobs are killed during test execution, so this
increases the time-out parameter from 1 hour and 20 minutes to 2 hours.


netoarmando:
ipa-4-8:

* 2c7ad3c3df3decfcb61b0a88c13a80282ef5561b prci: increase timeout for jobs that required AD

PR #3759, opened 2019-10-01
prci: increase timeout for jobs that required AD

netoarmando:
Vagrant retries to provision hosts if something happens, it was introduced
in PR-CI after https://github.com/freeipa/freeipa-pr-ci/commit/380c8b8c78a1ce277b7c1a327bda9d123c117c4d.

This takes time, some jobs are killed during test execution, so this
increases the time-out parameter from 1 hour and 20 minutes to 2 hours.

Signed-off-by: Armando Neto <abiagion@redhat.com>

PR #3761, opened 2019-10-01
[ipa-4-7] prci: increase timeout for jobs that required AD

netoarmando:
Vagrant retries to provision hosts if something happens, it was introduced
in PR-CI after https://github.com/freeipa/freeipa-pr-ci/commit/380c8b8c78a1ce277b7c1a327bda9d123c117c4d.

This takes time, some jobs are killed during test execution, so this
increases the time-out parameter from 1 hour to 2 hours.


netoarmando:
`test_sssd.py` doesn't require this change, this changed in #3759, thus this should be closed.

PR #3768, opened 2019-10-07
Use AES for KRA archival

tiran:

### DIFF
--- a/ipaserver/plugins/vault.py
+++ b/ipaserver/plugins/vault.py
@@ -23,6 +23,10 @@ from ipalib.frontend import Command, Object
 from ipalib import api, errors
 from ipalib import Bytes, Flag, Str, StrEnum
 from ipalib import output
+from ipalib.constants import (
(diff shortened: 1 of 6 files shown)


ipaserver/plugins/vault.py
```diff
@@ -42,9 +42,10 @@
     # pylint: disable=no-member
     try:
         # pki >= 10.4.0
-        from pki.crypto import DES_EDE3_CBC_OID
+        from pki.crypto import DES_EDE3_CBC_OID, AES_128_CBC_OID
     except ImportError:
         DES_EDE3_CBC_OID = pki.key.KeyClient.DES_EDE3_CBC_OID
+        AES_128_CBC_OID = None
```
rcritten: I think this could be dangerous. If DES_EDE3_CBC_OID ever fails import it will disable AES.

rcritten:
The commit message is a bit thin.

ipaserver/plugins/vault.py
tiran: This was just experimental pathc. IPA 4.8 requires Dogtag with these constants always defined. I have simplified the code.

PR #3769, opened 2019-10-07
Don't install a preexec_fn by default

tiran:
ipautil.run() now only installs a preexec_fn when it is actually needed.
This addresses a compatibility issue with mod_wsgi subinterpreters under
Python 3.8.

See: https://bugs.python.org/issue37951
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1759290

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -515,14 +515,18 @@ def run(args, stdin=None, raiseonerr=True, nolog=(), env=None,
             for group, gid in zip(suplementary_groups, suplementary_gids):
                 logger.debug('suplementary_group=%s (GID %d)', group, gid)
 
-    def preexec_fn():
-        if runas is not None:
-            os.setgroups(suplementary_gids)
-            os.setregid(pent.pw_gid, pent.pw_gid)
-            os.setreuid(pent.pw_uid, pent.pw_uid)
+    if runas is not None or umask is not None:
+        # preexec function is not supported in WSGI environment
+        def preexec_fn():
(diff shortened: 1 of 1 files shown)


ipapython/ipautil.py
```diff
@@ -515,14 +515,18 @@ def run(args, stdin=None, raiseonerr=True, nolog=(), env=None,
...
-            os.setregid(pent.pw_gid, pent.pw_gid)
-            os.setreuid(pent.pw_uid, pent.pw_uid)
-
-        if umask:
-            os.umask(umask)
+    if runas is not None or umask is not None:
+        # preexec function is not supported in WSGI environment
+        def preexec_fn():
+            if runas is not None:
+                os.setgroups(suplementary_gids)
```
gpshead: also worth noting: in Python 3.9 the subprocess module gains support for doing this without preexec_fn:
  https://bugs.python.org/issue36046

i don't believe we have umask support yet, but that one is easier to work around via a child process shell that calls umask before exec'ing the target.  For anything you find a preexec_fn being used for, feel free to file subprocess feature requests with upstream CPython.

abbra:
LGTM. Please file requests for adding umask support to subprocess for future Python versions.

gpshead:
https://bugs.python.org/issue38417 filed to track adding umask support in the future.

tiran:
master:

* 0b8c81a5bc194ffac7ea713930d6b922d10285d4 Don't install a preexec_fn by default

PR #3794, opened 2019-10-16
ipa-backup: fix python2 issue with os.mkdir

flo-renaud:
Python2 and python3 have different interfaces for os.mkdir:
python2: os.mkdir(path[, mode])
python3: os.mkdir(path, mode=0o777, *, dir_fd=None)

ipa-backup is using the python3 format, which breaks deployments using
python2. The fix consists in using os.mkdir(path, 0o700) instead of
os.mkdir(path, mode=0o700).

Fixes: https://pagure.io/freeipa/issue/8099


flo-renaud:
Adding labels ipa-4-8, ipa-4-7 and ipa-4-6 as the issue was introduced in ipa-4-6 branch, and the fix is needed for RHEL7.8

abbra:
LGTM.

PR #3811, opened 2019-10-18
 ipatests:  'sss_ssh_authorizedkeys user' should return ssh key

amore17:
Added regression tests which validates that when there is ssh key and certificate is added for IPA user then 'sss_ssh_authorizedkeys user' should return ssh key.

Related Ticket:
https://pagure.io/SSSD/sssd/issue/3937

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -710,3 +710,56 @@ class TestIPACommand(IntegrationTest):
             ['ipa-adtrust-install', '-a', 'Secret123', '--add-sids', '-U'])
         res = self.master.run_command(['testparm', '-s'])
         assert 'ERROR' not in (res.stdout_text + res.stderr_text)
+
+    def test_sss_ssh_authorizedkeys(self):
+        """Login via Ssh using private-key for ipa-user should work.
+
+        Test for : https://pagure.io/SSSD/sssd/issue/3937
+        Steps:
+        1) setup user with ssh-key and certificate stored in ipaserver
+        2) simulate p11_child timeout
+        3) try to login via ssh using private key.
+        """
+        user = 'testsshuser'
+        passwd = 'Secret123'
+        user_key = tasks.create_temp_file(self.master, create_file=False)
+        pem_file = tasks.create_temp_file(self.master)
+        # Create a user with a password
+        tasks.create_active_user(self.master, user, passwd)
+        tasks.kinit_admin(self.master)
+        tasks.run_command_as_user(
+            self.master, user, ['ssh-keygen', '-N', '',
+                                '-f', user_key])
+        ssh_pub_key = self.master.get_file_contents('{}.pub'.format(
+            user_key), encoding='utf-8')
+        openssl_cmd = [
+            'openssl', 'req', '-x509', '-newkey', 'rsa:2048', '-days', '365',
+            '-nodes', '-out', pem_file, '-subj', '/CN=' + user]
+        self.master.run_command(openssl_cmd)
+        cert_b64 = self.get_cert_base64(self.master, pem_file)
+        sssd_p11_child = '/usr/libexec/sssd/p11_child'
+        backup = tasks.FileBackup(self.master, sssd_p11_child)
+        try:
+            content = '#!/bin/bash\nsleep 999999'
+            # added sleep to simulate the timeout for p11_child
+            self.master.put_file_contents(sssd_p11_child, content)
+            self.master.run_command(
+                ['ipa', 'user-mod', user, '--ssh', ssh_pub_key])
+            self.master.run_command([
+                'ipa', 'user-add-cert', user, '--certificate', cert_b64])
+            # clear cache to avoid SSSD to check the user in old lookup
+            tasks.clear_sssd_cache(self.master)
+            result = self.master.run_command(
+                [paths.SSS_SSH_AUTHORIZEDKEYS, user])
+            assert ssh_pub_key in result.stdout_text
+            # login to the system
+            self.master.run_command(
+                ['ssh', '-o', 'PasswordAuthentication=no',
+                 '-o', 'IdentitiesOnly=yes', '-o', 'StrictHostKeyChecking=no',
+                 '-l', user, '-i', user_key, self.master.hostname, 'true'])
+        finally:
+            # cleanup
+            self.master.run_command(['ipa', 'user-del', user])
+            backup.restore()
+            self.master.run_command(['rm', '-f', pem_file, user_key,
+                                     '{}.pub'.format(user_key)])
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -1865,9 +1865,11 @@ def ldapsearch_dm(host, base, ldap_args, scope='sub', **kwargs):
     return host.run_command(args, **kwargs)
 
 
-def create_temp_file(host, directory=None):
+def create_temp_file(host, directory=None, create_file=True):
     """Creates temproray file using mktemp."""
     cmd = ['mktemp']
+    if create_file is False:
+        cmd += ['--dry-run']
     if directory is not None:
         cmd += ['-p', directory]
     return host.run_command(cmd).stdout_text.strip()


wladich:
I see some issues with this PR:
1.
In https://bugzilla.redhat.com/show_bug.cgi?id=1659478 you are using the following scenario:
* setup ipa user with
   * ssh key
   * certificate
* replace /usr/libexec/sssd/p11_child with a dummy script
* check sss_ssh_authorizedkeys

But in this PR you are proposing different one:
* setup ipa user with only ssh keys
* check sss_ssh_authorizedkeys

2.
Original issue was about user not beeing able to login using ssh key stored in IPA when he also has certificate. Though sss_ssh_authorizedkeys should do the check, we should also check that original issue is gone, i.e. do ssh login using the stored keys.

3. test_commands test_suite is generally for testing IPA. There is a dedicated suite test_sssd for sssd-related tests.

amore17:
@wladich Thanks for the review.
1. As per [comment](https://bugzilla.redhat.com/show_bug.cgi?id=1659478#c17) 
After fix ssh keys from the certificates should be skipped as they are skipping so i have not added certificate to the ipauser.
2. As the expected is it should return ssh-key and there should be no error as 'Error looking up public keys'
3. As this needs only master and it also tested with ipauser thats why added in test_commands.py

wladich:
1. But without certificate this test would pass on unfixed version. As this is a regression test, you should reproduce the original scenario: user has certificate + ssh key, p11_child times out
2. Again, this is regression test for original problem: user is not able to login. Checking sss_ssh_authorizedkeys output and logs is a good additional check, but we must verify that the original problem is really gone.

wladich:
3. We try to organize test cases by suites basing on tested feature or component. Combining sssd tests in one suite will allow us to execute all tests related to sssd in one run.

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +680,58 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+            self.master.put_file_contents(binary, content)
+            tasks.kinit_admin(self.master)
+            password = 'Secret123'
+            testuser_password_confirmation = "%s\n%s\n" % (password,
+                                                           password)
+            self.master.run_command(['ipa', 'user-add', user,
+                                     '--first', user,
+                                     '--last', user,
+                                     '--password'],
+                                    stdin_text=testuser_password_confirmation)
```
wladich: You can use tasks.user_add() for this

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +680,58 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+        pem_file = tempfile.mkdtemp(suffix='cert.pem', dir=paths.TMP)
+        key = tempfile.mkdtemp(suffix='private.key', dir=paths.TMP)
+        openssl_cmd = [
+            'openssl', 'req', '-x509', '-newkey', 'rsa:2048', '-days', '365',
+            '-nodes', '-keyout', key, '-out', pem_file, '-subj', '/CN=' + user]
+        self.master.run_command(openssl_cmd)
+        cert_b64 = self.get_cert_base64(self.master, pem_file)
+        try:
+            binary = '/usr/libexec/sssd/p11_child'
+            backup = tasks.FileBackup(self.master, binary)
```
wladich: If this line fails then  `backup.restore()` in `finally` block will fail. It is better to place this line before of try-finally block

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +680,58 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+        2. call sss_ssh_authorizedkeys and ssh key should be returned
+        """
+        user = 'testsshuser'
+        ssh_key_pair = tasks.generate_ssh_keypair()
+        pem_file = tempfile.mkdtemp(suffix='cert.pem', dir=paths.TMP)
+        key = tempfile.mkdtemp(suffix='private.key', dir=paths.TMP)
+        openssl_cmd = [
+            'openssl', 'req', '-x509', '-newkey', 'rsa:2048', '-days', '365',
+            '-nodes', '-keyout', key, '-out', pem_file, '-subj', '/CN=' + user]
+        self.master.run_command(openssl_cmd)
```
wladich: There are two issues here: 
1. mkdtemp creates a directory but openssl needs a file name
2. tempfile.mkdtemp operates on local filesystem (i.e. on test runner) but openssl is executed remotely on ipa master

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +680,58 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+        cert_b64 = self.get_cert_base64(self.master, pem_file)
+        try:
+            binary = '/usr/libexec/sssd/p11_child'
```
wladich: It is better to use more descriptive variable name here

wladich: And please provide a comment explaining the reason for replacing p11_child with this dummy script

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +680,58 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+            backup = tasks.FileBackup(self.master, binary)
+            content = '#!/bin/bash\nsleep 999999'
+            self.master.put_file_contents(binary, content)
```
wladich: have you checked if file mode is preserved when writing the contents this way?

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +680,58 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+            password = 'Secret123'
+            testuser_password_confirmation = "%s\n%s\n" % (password,
+                                                           password)
+            self.master.run_command(['ipa', 'user-add', user,
+                                     '--first', user,
+                                     '--last', user,
+                                     '--password'],
+                                    stdin_text=testuser_password_confirmation)
+            cmd = 'ipa user-mod {} --ssh "{}"'.format(user, ssh_key_pair[1])
+            self.master.run_command(cmd)
```
wladich: There is not need to use `cmd` variable, simply put the value to arguments

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +680,58 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+            self.master.run_command([
+                'ipa', 'user-add-cert', user, "--certificate=%s" % cert_b64])
+            self.master.run_command(['sss_cache', '-E'])
```
wladich: Are you sure we really need to invalidate the cache? We have not requested any data for the user yet.

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +680,58 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+            self.master.run_command(['getent', 'passwd', user])
```
wladich: What is the purpose of this command for the test scenario?

ipatests/test_integration/test_commands.py
```diff
@@ -679,3 +680,58 @@ def test_user_mod_change_capitalization_issue5879(self):
...
+            result = self.master.run_command(
+                [paths.SSS_SSH_AUTHORIZEDKEYS, user])
+            assert ssh_key_pair[1] in result.stdout_text
+            # login to the system
+            client = paramiko.SSHClient()
+            client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+            client.connect(self.master.hostname,
+                           username=user,
+                           password=password)
+            client.close()
```
wladich: You are establishing ssh connection from tests runner machine which in general case does not have sssd installed and is not enrolled to ipa domain.

ipatests/test_integration/test_commands.py
amore17: Yes it is needed to invalidate cache.

ipatests/test_integration/test_commands.py
```diff
@@ -699,3 +699,61 @@ def is_tls_version_enabled(tls_version):
...
+        user = 'testsshuser'
+        passwd = 'Secret123'
+        user_key = '/tmp/ssh-key'
+        # Create a user with a password
+        tasks.kinit_admin(self.master)
+        tasks.user_add(self.master, user, password=passwd)
+        user_kinit_stdin_text = "{old}\n%{new}\n%{new}\n".format(
+            old=passwd, new=passwd)
+        self.master.run_command(['kinit', user],
+                                stdin_text=user_kinit_stdin_text)
```
wladich: There is a helper function tasks.create_active_user() which does exactly this

ipatests/test_integration/test_commands.py
```diff
@@ -699,3 +699,61 @@ def is_tls_version_enabled(tls_version):
...
+        tasks.kinit_admin(self.master)
+        tasks.run_command_as_user(
```
wladich: Please try avoid interactive communication using stdin. If you  add `-N ""` option then ssh-keygen will not ask for passphrase.

ipatests/test_integration/test_commands.py
```diff
@@ -699,3 +699,61 @@ def is_tls_version_enabled(tls_version):
         assert not is_tls_version_enabled('tls1')
         assert not is_tls_version_enabled('tls1_1')
         assert is_tls_version_enabled('tls1_2')
+
+    def test_sss_ssh_authorizedkeys(self):
+        """
```
wladich: Issues with the docstring:
1. It does not reflect the actual test case. Actual steps are: 1) setup user with ssh-key and certificate stored in ipaserver 2) simulate p11_child timeout 3) try to login via ssh using private key.
And the description is something like "Check private ssh key is retrieved when user has certificate and p11_child times out." - I guess that is the original bug summary.
2. As per PEP 257 [1], there should be no new line between opening quotes and first line.

[1] https://www.python.org/dev/peps/pep-0257/

ipatests/test_integration/test_commands.py
```diff
@@ -699,3 +699,61 @@ def is_tls_version_enabled(tls_version):
...
+        self.master.run_command(['kinit', user],
+                                stdin_text=user_kinit_stdin_text)
+        tasks.kinit_admin(self.master)
+        tasks.run_command_as_user(
+            self.master, user, ['ssh-keygen', '-f', user_key],
+            stdin_text='\n\n')
+        ssh_pub_key = self.master.get_file_contents('{}.pub'.format(
+            user_key), encoding='utf-8')
+        pem_file = '/tmp/cert.pem'
+        key = '/tmp/private.key'
```
wladich: You are not using the key further in the test so there is no reason to save it. You can simply omit `-keyout`  option in openssl invocation

ipatests/test_integration/test_commands.py
```diff
@@ -699,3 +699,61 @@ def is_tls_version_enabled(tls_version):
...
+        Test for : https://pagure.io/SSSD/sssd/issue/3937
+
+        Steps:
+        1. add ipa user with ssh-key.
+        2. call sss_ssh_authorizedkeys and ssh key should be returned
+        3. check user is able to ssh or not.
+        """
+        user = 'testsshuser'
+        passwd = 'Secret123'
+        user_key = '/tmp/ssh-key'
```
wladich: It is better to create temp file using mktemp to avoid collisions with existing files.
There is a helper `tasks.create_temp_file` to create temp file on remote host.

Same applies to pem_file variable

ipatests/test_integration/test_commands.py
```diff
@@ -699,3 +699,61 @@ def is_tls_version_enabled(tls_version):
...
+            self.master.run_command(['sss_cache', '-E'])
+            result = self.master.run_command(
+                [paths.SSS_SSH_AUTHORIZEDKEYS, user])
+            assert ssh_pub_key in result.stdout_text
+            # login to the system
+            self.master.run_command(
+                ['ssh', '-o=PasswordAuthentication=no', '-l', user, '-i',
+                 user_key, self.master.hostname, 'true'], stdin_text='yes')
+        finally:
+            # cleanup
```
wladich: Try to avoid leaving junk after test run - remove the created files in /tmp

ipatests/test_integration/test_commands.py
```diff
@@ -699,3 +699,61 @@ def is_tls_version_enabled(tls_version):
...
+            self.master.run_command('ipa user-mod {} --ssh "{}"'.format(
+                user, ssh_pub_key))
+            self.master.run_command([
+                'ipa', 'user-add-cert', user, "--certificate=%s" % cert_b64])
+            self.master.run_command(['sss_cache', '-E'])
+            result = self.master.run_command(
+                [paths.SSS_SSH_AUTHORIZEDKEYS, user])
+            assert ssh_pub_key in result.stdout_text
+            # login to the system
+            self.master.run_command(
```
wladich: Please try avoid interactive communication using stdin. 
Use the following option to ssh:  * `-o StrictHostKeyChecking=no` - do not ask for confirmation if new host is seen and do not fail if host key has changed

ipatests/test_integration/test_commands.py
wladich: Please can you be a little more specific?

ipatests/test_integration/test_commands.py
amore17: For ssh-key using it as ssh-key{0}".format(random.randint(1, 99)) 
If i use tasks.create_temp_file it will throw error for ssh-keygen that file already exist / or you want to override. 
so tasks.create_temp_file is using only for  pem_file

wladich:
Please add a more descriptive commit message.

ipatests/test_integration/test_commands.py
```diff
@@ -699,3 +699,55 @@ def is_tls_version_enabled(tls_version):
...
+        sssd_p11_child = '/usr/libexec/sssd/p11_child'
+        backup = tasks.FileBackup(self.master, sssd_p11_child)
+        try:
+            content = '#!/bin/bash\nsleep 999999'
+            # added sleep to simulate the timeout for p11_child
+            self.master.put_file_contents(sssd_p11_child, content)
+            self.master.run_command('ipa user-mod {} --ssh "{}"'.format(
+                user, ssh_pub_key))
+            self.master.run_command([
+                'ipa', 'user-add-cert', user, "--certificate=%s" % cert_b64])
```
wladich: Please be consistent 
a) in using quotes vs double quotes - the preferred is single quotes for strings.
b) in using format styles - the preferred is using .format() method

Also I am pretty sure that `--certificate` and value of `cert_b64` can be provided as separate arguments, without `=` between them

ipatests/test_integration/test_commands.py
```diff
@@ -710,3 +710,58 @@ def test_samba_config_file(self):
...
+        Test for : https://pagure.io/SSSD/sssd/issue/3937
+        Steps:
+        1) setup user with ssh-key and certificate stored in ipaserver
+        2) simulate p11_child timeout
+        3) try to login via ssh using private key.
+        """
+        user = 'testsshuser'
+        passwd = 'Secret123'
+        user_key = tasks.create_temp_file(self.master, create_file=False)
+        pem_file = tasks.create_temp_file(self.master, directory='/tmp')
```
wladich: `directory` argument is not needed here, the default directory for `mktemp` is /tmp

ipatests/test_integration/test_commands.py
```diff
@@ -710,3 +710,58 @@ def test_samba_config_file(self):
...
+            self.master.run_command(
+                ['ssh', '-o', 'PasswordAuthentication=no',
+                 '-o', 'IdentitiesOnly=yes', '-o', 'StrictHostKeyChecking=no',
+                 '-l', user, '-i', user_key, self.master.hostname, 'true'])
+        finally:
+            # cleanup
+            self.master.run_command(['ipa', 'user-del', user])
+            backup.restore()
+            self.master.run_command(['rm', '-f', pem_file, user_key,
+                                     '%s.pub' % user_key])
```
wladich: Please try to use one style of string formatting. The preferred style for new code is with `.format()`

ipatests/test_integration/test_commands.py
wladich: Does this command really need to be run in user account? It seems to me we can use simple `self.master.run_command`.

ipatests/test_integration/test_commands.py
```diff
@@ -710,3 +710,58 @@ def test_samba_config_file(self):
...
+        self.master.run_command(openssl_cmd)
+        cert_b64 = self.get_cert_base64(self.master, pem_file)
+        sssd_p11_child = '/usr/libexec/sssd/p11_child'
+        backup = tasks.FileBackup(self.master, sssd_p11_child)
+        try:
+            content = '#!/bin/bash\nsleep 999999'
+            # added sleep to simulate the timeout for p11_child
+            self.master.put_file_contents(sssd_p11_child, content)
+            self.master.run_command(
+                ['ipa', 'user-mod', user, '--ssh', "{}".format(ssh_pub_key)])
```
wladich: `"{}".format(ssh_pub_key)` can be replaces with `ssh_pub_key`

ipatests/test_integration/test_commands.py
```diff
@@ -710,3 +710,58 @@ def test_samba_config_file(self):
...
+            self.master.run_command([
+                'ipa', 'user-add-cert', user, '--certificate',
+                '"{0}"'.format(cert_b64)])
```
wladich: 1. Double quotes are not needed here as run_command is surrounding each argument in single quotes.
2. ` '{0}'.format(cert_b64)` can be replaced with `cert_b64`

ipatests/test_integration/test_commands.py
```diff
@@ -710,3 +710,58 @@ def test_samba_config_file(self):
...
+            # clear cache to avoid SSSD to check the user in old lookup
+            tasks.clear_sssd_cache(self.master)
+            result = self.master.run_command(
+                [paths.SSS_SSH_AUTHORIZEDKEYS, user])
+            assert 'Error looking up public keys' not in (result.stdout_text +
```
wladich: This type of check is very fragile: if a single character changes in error message we will not be able to detect a regression. It is better to check if command returns the proper user key as was the case in one of the previous version.

wladich:
LGTM, except one unresolved comment from previous review: https://github.com/freeipa/freeipa/pull/3811/files#r352656052

ipatests/test_integration/test_commands.py
amore17: Yes It needs to be run in user account so that in key you will see user@domain 
And if we run it as master.run_command then it will generate keys for root user or any other user.

wladich: `user@domain` in public key is used for reference only. It works perfectly fine when created by root user.
But well, lets leave it as it is...

flo-renaud:
Hi @amore17 
on ipa-4-6 branch, the tests are run on Fedora 27 and this version ships with sssd-1.16.3-2.fc27 which does not contain the fix. This means that the nightly test will fail on the ipa-4-6 branch.
You will need to edit the backport to ipa-4-6 and add a skip annotation.

flo-renaud:
master:

* 836b90f65244a0407e74627b550297d0962d882e ipatests : Login via ssh using private-key for ipa-user should work.

flo-renaud:
@amore17 
manual backports are required to ipa-4-7 and ipa-4-6.

wladich:
@amore17 
About skipping test on ipa-4-6 mentioned by @flo-renaud:
1) Please use xfail with strict=True
2) add it with a separate commit

PR #3812, opened 2019-10-18
ipatests: add tests to check that only secure TLS protocols are enabled in Apache

wladich:
Related to: https://pagure.io/freeipa/issue/7995

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -679,3 +679,23 @@ class TestIPACommand(IntegrationTest):
         assert 'Modified user "%s"' % (ipauser) in cmd.stdout_text
         assert 'First name: %s' % (modfirst) in cmd.stdout_text
         assert 'Last name: %s' % (modlast) in cmd.stdout_text
+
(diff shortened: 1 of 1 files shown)


rcritten:
Failures are due to new version of pylint.

wladich:
@

> Failures are due to new version of pylint.

I do not understand what I can do about it. It found issues in files I did not touch.

rcritten:
Sorry, I should have mentioned that. This is being looked at upstream, there isn't anything you can do about it.

rcritten:
The commit message line wraps.
More details than "secure" are needed. I'd just say:

ipatests: add test to check that only TLS 1.2 is enabled in Apache

Maybe it's a style thing but since is_tls_version_enabled() is only used within one test you should be able to embed that function inside it.

rcritten:
master:

* 14be2715334e16a2d3f07a6b64bcd6d068ce89c1 ipatests: add test to check that only TLS 1.2 is enabled in Apache

rcritten:
@wladich can you do manual backports to the ipa-4-7 and ipa-4-6 branches?

PR #3813, opened 2019-10-18
ipatests: check adding sudorule option does not produce false failure message

wladich:
Related to: https://pagure.io/freeipa/issue/7649

### DIFF
--- a/ipatests/test_integration/test_sudo.py
+++ b/ipatests/test_integration/test_sudo.py
@@ -750,3 +750,26 @@ class TestSudo(IntegrationTest):
         finally:
             self.master.run_command(
                 ['ipa', 'config-mod', '--domain-resolution-order='])
+
+    def test_sudooption_not_failing(self):
+        """Check there is no false error message when adding option to sudorule
+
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_sudo.py
```diff
@@ -750,3 +750,26 @@ def test_domain_resolution_order(self):
         finally:
             self.master.run_command(
                 ['ipa', 'config-mod', '--domain-resolution-order='])
+
+    def test_sudooption_not_failing(self):
```
rcritten: I'd replace "not_failing" with "display_members" since the bug was about not converting member DNs into more user-friendly strings.

rcritten:
The test itself seems fine and I guess it might augment the xmlrpc test added with the original code change. Belt and suspenders.

In the git commit the message wraps. It would be nice to keep it limited to a single line.

And also mention more specifically what it is testing: calculating the indirect members and converting the member attributes.

wladich:
Closing PR since the xmlrpc test covers the issue.

PR #3823, opened 2019-10-24
Fixes debian path for IPA_CUSTODIA_HANDLER

olsonse:
Debian installs into a different directory for libexec files.  This patch
fixes the path to the custodia files for debian.

Signed-off-by: Spencer E. Olson <olsonse@umich.edu>


rcritten:
I have no way of testing this but it looks ok to me and matches the pattern of other custom debian files.

@tjaalton does this look ok to you?

tiran:
The path fix looks good to me.

tjaalton:
yep, looks fine

rcritten:
Adding re-run to run the full CI. I don't expect any failures, just procedure. I'll ack and push once that passes.

rcritten:
master:

* 73796c779758fdd8f3c2d8ede3ef24d5b7d025e7 Fixes debian path for IPA_CUSTODIA_HANDLER

PR #3825, opened 2019-10-25
DSU: add Design for Disable Stale Users

fcami:
Add disable-stale-users.md: feature document for the upcoming DSU feature.

Fixes: https://pagure.io/freeipa/issue/8104

### DIFF
--- a/doc/designs/disable-stale-users.md
+++ b/doc/designs/disable-stale-users.md
@@ -1,0 +1,303 @@
+# Disable Stale Users
+
+**DESIGN STAGE**
+
+## Overview
+
+Some regulations call for accounts that have not been used for a specific
+amount of time be automatically disabled in the identity management system.
+
+FreeIPA does not keep track of successful authentication events.
+The krbLastSuccessfulAuth attribute is not updated by default since
+https://pagure.io/freeipa/issue/5313 for reliability reasons.
+Therefore distinguishing between stale and active accounts cannot be done with
+existing logon data.
+
+
+## Possible approaches
+
+### Environments with mandatory, periodic password changes
+
+The problem is easy to solve in environments where user passwords must be
+changed regularly. A user account whose associated password is past its due
+change date can be considered stale and therefore be disabled by the system.
+In this case, a grace period must be provided between password expiration and
+the actual account locking event.
+Total writes to LDAP using this approach due to the disable stale users tooling
+are minimal (only nsAccountLock). However, password-change related writes are
+needed every 90 days (by default).
+Each password change updates the krbLastPwdChange, krbPasswordExpiration,
+userPassword, krbPrincipalKey, and krbExtraData attributes.
+
+The Disable Stale Users (DSU) tooling proposes to make sure no non-locked
+account has a password expired since N days, N being 7 by default.
+
+### Environments with no mandatory password changes
+
+For environments where user passwords never expire, a new mechanism must be
+added to FreeIPA plugins ipa-kdb and ipa-lockout to account for each user
+account authentication events. Since updating a LDAP user account attribute on
+each authentication event must be avoided as the update frequency could easily
+be high, (see https://pagure.io/freeipa/issue/5313 for background
+information), a coarse update mechanism is proposed.
+
+An attribute containing a timestamp is added to each user:
+coarseSuccessfulAuthTime. It will be added at account creation time (set to
+current date).
+The timestamp contained therein is not precise. It is updated by IPA plugins
+whenever a successful authentication event happens, but only if the current
+timestamp is sufficiently in the past. The update interval is configurable, but
+to avoid replication storms, it is randomized. To achieve this, a randomized
+value is added to a fixed time interval.
+
+Some attributes must be added to the cn=ipaConfig schema:
+* the first one to activate the mechanism (default: not present):
+```
+ipaConfigString: DSU:RecordSuccessfulAuth
+```
+* the second attribute contains the fixed time period (default: 28 days):
+```
+ipaDSURecordSucessfulAuthTimeInterval: 2419200
+```
+* the third one contains a time period to be added. The actual value used is
+  randomized between 0 and the attribute's value (default: 7 days) whenever
+  a user successfully authenticates. 
+```
+ipaDSURecordSucessfulAuthRandomizedTimeInterval: 604800
+```
+
+Updating this (replicated) timestamp attribute every ~30 days requires less
+LDAP writes than mandating password changes every 90 days as the latter would
+imply updating five (replicated) attributes at least once in that 90-day
+period.
+
+The Disable Stale Users (DSU) tooling, when the above mechanism is active,
+proposes to make sure no non-locked account has a timestamp attribute 90 days
+(by default) in the past.
+
+On a successful authentication event:
+
+```
+current_date = datetime.now(timezone.utc)
+last_update = coarseSuccessfulAuthTime
+update_internal = ipaDSURecordSucessfulAuthTimeInterval
+                  + randint(0, ipaDSURecordSucessfulAuthRandomizedTimeInterval)
+if current_date - last_update > update_internal:
+    coarseSuccessfulAuthTime = current_date
+```
+
+As a rule of thumb, if the requirement is to disable users not seen for 90 days
+maximum, updating the attribute roughly every 35 days is good enough.
+At worst, the user would have authenticated the first day and the last day of
+the 35-day maximum update interval and the randomized time interval would have
+been 7 days. 
+The attribute would therefore contain the date from the first day, as day #35
+would have been considered too close to the existing timestamp to warrant
+updating it. DSU would then proceed to disable the user account 90 days after
+the existing coarseSuccessfulAuthTime timestamp - this is only after 55 day
+(90-35) of real inactivity, which fits the requirement (albeit loosely).
+
+Having a 90-day inactive user disabling requirement would by default disable
+accounts really inactive for 55 (90-35) to 90 days.
+
+In short:
+
+```
+if (current_date - 90d) > coarseSuccessfulAuthTime:
+    disable_account()
+```
+
+The 90-day default value is configurable (see appropriate section below).
+
+
+### Notes
+
+Installing the DSU tooling will not change the behavior of existing clusters.
+DSU has to be explicitly configured and enabled to start disabling stale
+accounts.
+
+Accounts are locked out by setting the nsAccountLock attribute
+(OID: 2.16.840.1.113730.3.1.610) to True.
+
+FreeeIPA is not directly involved when a user logs in using an ssh key so it
+cannot make decisions on whether an account is "active" or not. However,
+setting nsaccountlock = TRUE will block all logins including those using ssh
+key authentication.
+For passwordless environments, S4U2Self is proposed to be added to SSSD to
+account for successful user logon events through ipa-kdb:
+https://pagure.io/SSSD/sssd/issue/4077
+The mechanism laid out in "Environments with no mandatory password changes"
+would then be applied as-is.
+
+There is always the possibility that a long-inactive user account will be
+disabled (nsAccountLock'd) by DSU even if that user managed to logon the
+previous few seconds before or during the DSU window due to:
+* the time it takes for the timestamp attribute to be replicated to the DSU
+runner
+* the time it takes for nsAccountLock to be replicated from the DSU runner
+to the replica this user uses.
+
+There is no way to prevent that from happening with the present design, but
+the probability of that happening is low.
+
+
+## Example
+
+### Environments with mandatory, periodic password changes
+
+Given the following requirements and facts:
+* stale accounts must be disabled after 90 days' inactivity
+* grace period is decided to be 7 days
+* the tool runs every day
+* replication is expected to complete in 2h maximum
+
+setting the maximum password lifetime to 80 days or less would meet the
+guidelines as 80+7+1 (give or take 2h) is less than 90.
+
+### Environments with no mandatory password changes
+
+If the requirement is to disable users not seen for 90 days maximum, updating
+the attribute every 35 days is good enough.
+At worst, the user would have authenticated the first day and at least once
+during this 35 day period, say day #29. The attribute would therefore contain
+the date from the first day, as day #29 would have been considered too close to
+the existing timestamp to warrant updating it. The user account would then be
+disabled after 61 days of inactivity which fits (loosely) the requirement.
+In short, having a 90-day inactive user disabling requirement would by default
+disable accounts inactive for 55 (90-35) to 90 days.
+
+
+## Limitations
+
+### Environments with mandatory, periodic password changes
+
+If regulations call for stale/inactive accounts to be disabled within X
+number of days, then the lifetime of the passwords for these accounts must be
+set to a value lower than:
+* X,
+* minus the expected grace period mentioned above,
+* minus the maximum amount of time between two invocations of the DSU tool,
+* minus the expected maximum replication delay from the replica the DSU tool is
+installed on to the farthest point of the cluster.
+
+### Environments with no mandatory password changes
+
+The longer the time period, the less update activity for the timestamp
+attribute. But using a longer period runs the risk of disabling users who have
+been gone for too short a time. For instance, setting the update period to 65
+days would disable accounts inactive for the last 25 to 90 days. Whether this
+fits the site's requirements is up to the administrators to determine.
+
+Sites requiring a better accuracy must take care not to set the update period
+too low for performance reasons. It is strongly recommended to make sure the
+update period is longer than the expected maximum PTO or simply "away from IPA"
+period for most user accounts to avoid replication storms when users come back.
+This is especially true when these away periods are in sync, like in schools.
+
+
+## Multiple runners
+
+DSU can be installed on more than a single replica. This provides redundancy.
+Having multiple DSU instances active at the same time provides no advantage,
+so the systemd timer helper script includes a randomized time wait period.
+
+On very busy clusters, using dedicated replicas (hidden replicas) for DSU is
+recommended.
+
+
+## How to Use DSU
+
+### Installation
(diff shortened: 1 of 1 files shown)


doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,147 @@
...
+Therefore distinguishing between stale and active accounts cannot be done with
+existing logon data.
+
+However, user passwords have to be changed regularly in these environments as
+well. A user who failed to change his password in time can be considered stale
+and therefore be disabled by the system.
+In this case, a grace period should be provided between password expiration and
+the actual account locking event.
+
+The DisableStaleUsers (DSU) tooling is provided to make sure no non-locked-out
```
rcritten: Disable Stale Users

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,147 @@
...
+
+```
+[dsu]
+active = False
+grace_period = 7d
+ignore_groups = admins
+log = True
+```
+
+Switch active to True for the DSU tool to disable (lock) accounts.
```
rcritten: active sounds more like "it's running". disable_accounts seems clearer to me.

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,147 @@
...
+
+Activate the DSU tool to run every night using the provided systemd timer:
+``systemctl enable ipa-dsu.timer``
+
+
+### Logs
+
+The DSU tool logs each action in journald. Sample output:
+```
+ipa-dsu: disabled account uid=alice,cn=users,cn=accounts,dc=example,dc=com - password was expired since 20191001.
```
rcritten: perhaps: password expired at" instead.

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,147 @@
...
+The DSU tool is not a UNIX daemon. It is a CLI tool that can be triggered
+by a systemd timer. 
+
+Running ``ipa-dsu`` is enough to start the tool. If active is set to True in
+the above-mentioned configuration file, it will disable stale accounts
+appopriately.
+
+The dry-run mode ``ipa-dsu --dry-run`` does not disable/lock accounts in any
+way. Rather, it shows what would be done, essentially acting as a verbose mode
+with "active" set to False. The output is JSON.
```
rcritten: Would a JSON mock-up be a good idea or is that going to be done in the future?

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,147 @@
...
+
+setting the maximum password lifetime to 80 days or less would meet the
+guidelines as 80+8+1 is less than 90.
+
+## Multiple runners
+
+DSU can be installed on more than a single replica. This provides redundancy.
+However, care must be taken that DSU is not triggered both:
+* at the same time
+* on the same accounts
```
rcritten: What is the risk of running at the same time? At worst it would disable the same account at the same time right?

doc/designs/disable-stale-users.md
rcritten: Would it be better to store the config in LDAP so it doesn't have to be updated on all masters running the service?

doc/designs/disable-stale-users.md
fcami: nsAccountLock is not present until a user account has been locked once. I'd rather not have two ADD operations on the same entry + attribute at the same time, though the end value is going to be the same.

doc/designs/disable-stale-users.md
fcami: I'd rather do it after the fact. I'm undecided on how it's going to look like exactly.

doc/designs/disable-stale-users.md
fcami: fixed, thanks!

doc/designs/disable-stale-users.md
fcami: Fixed, thanks for the proposal!

doc/designs/disable-stale-users.md
fcami: Fixed, thanks

rcritten:
I'd also highlight that this solves the ssh key login problem. IPA is not directly involved when a user logs in using an ssh key so it can't make decisions on whether an account is "active" or not. By using the password expiration instead this is worked around. nsaccountlock = TRUE will block all logins, including ssh key.

fcami:
Now with the other mechanism to deal with sites with no password changes. Not 100% complete, will push a better version in a bit.

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,281 @@
...
+The problem is easy to solve in environments where user passwords must be
+changed regularly. A user account whose associated password is past its due
+change date can be considered stale and therefore be disabled by the system.
+In this case, a grace period should be provided between password expiration and
+the actual account locking event.
+Total writes to LDAP using this approach due to the disable stale users tooling
+are minimal (only nsAccountLock). However, password-change related writes are
+needed every 90 day (by default).
+Each password change updates the krbLastPwdChange, krbPasswordExpiration, and
+krbExtraData attributes.
```
rcritten: being pedantic, a password change also updates userPassword and krbPrincipalKey.

fcami: Fixed, thanks!

fcami:
Now with a note explaining that configuration might be done in LDAP instead of the proposed configuration file format.

doc/designs/disable-stale-users.md
fcami: As you suggested, I've added a randomized timer which should be enough to alleviate issues, plus there would not be any collision anyway.

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,287 @@
...
+uses configuration files but this is not set in stone yet.
+
+#### Environments with mandatory, periodic password changes
+
+The configuration file for DSU contains the following variables:
+
+```
+[password-based-dsu]
+disable_accounts = False
+grace_period = 7d
```
rcritten: disable_accounts and grace_period were moved to LDAP? Or just grace_period?

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,281 @@
...
+account authentication events. Since updating a LDAP user account attribute on
+each authentication event must be avoided as the update frequency could easily
+be high, (see https://pagure.io/freeipa/issue/5313 for background
+information), a coarse update mechanism is proposed.
+
+An attribute containing a timestamp is added to each user.
+The timestamp is updated by IPA plugins whenever a successfull authentication
+event happens, but only if the previous timestamp is sufficiently in the past.
+
+Three ipaConfigString attributes must be added to the schema:
```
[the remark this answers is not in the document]
fcami: Yes, I've updated the doc.

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,281 @@
...
+### Environments with no mandatory password changes
+
+For environments where user passwords never expire, a new mechanism must be
+added to FreeIPA plugins ipa-kdb and ipa-lockout to account for each user
+account authentication events. Since updating a LDAP user account attribute on
+each authentication event must be avoided as the update frequency could easily
+be high, (see https://pagure.io/freeipa/issue/5313 for background
+information), a coarse update mechanism is proposed.
+
+An attribute containing a timestamp is added to each user.
```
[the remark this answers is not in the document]
fcami: coarseSuccessfulAuthTime will do for now, thanks! Doc updated.

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,291 @@
...
+### Environments with mandatory, periodic password changes
+
+The problem is easy to solve in environments where user passwords must be
+changed regularly. A user account whose associated password is past its due
+change date can be considered stale and therefore be disabled by the system.
+In this case, a grace period must be provided between password expiration and
+the actual account locking event.
+Total writes to LDAP using this approach due to the disable stale users tooling
+are minimal (only nsAccountLock). However, password-change related writes are
+needed every 90 day (by default).
```
rcritten: nit: day -> days

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,291 @@
...
+account authentication events. Since updating a LDAP user account attribute on
+each authentication event must be avoided as the update frequency could easily
+be high, (see https://pagure.io/freeipa/issue/5313 for background
+information), a coarse update mechanism is proposed.
+
+An attribute containing a timestamp is added to each user:
+coarseSuccessfulAuthTime. The timestamp contained therein is not precise.
+It will be added at account creation time (set to current date).
+The timestamp is updated by IPA plugins whenever a successful
+authentication event happens, but only if the previous timestamp is
```
rcritten: previous -> current (or last).

Just to clarify that the if the value currently in LDAP is not sufficiently in the past then no update is made.

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,291 @@
...
+sufficiently in the past.
+
+Three ipaConfigString attributes must be added to the cn=ipaConfig schema:
+* the first one to activate the mechanism (default: False):
+  DSU:RecordSuccessfulAuth
+* the second one contains the fixed time period (default: 28 days):
+  DSU:RecordSucessfulAuth-TimeInterval
+* the third one contains a random time period to be added (default: 7 days):
+  DSU:RecordSucessfulAuth-RandomizedTimeInterval
```
rcritten: We'd need to see whether a dashed attribute will work as an IPA option, I don't think off the top of my head it will as python will interpret it as subtraction.

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,291 @@
...
+during this 35 day period, say day #29. The attribute would therefore contain
+the date from the first day, as day #29 would have been considered too close to
+the existing timestamp to warrant updating it. The user account would then be
+disabled after 61 days of inactivity which fits (loosely) the requirement.
+In short, having a 90-day inactive user disabling requirement would by default
+disable accounts inactive for 55 (90-35) to 90 days. 
+
+Updating this (replicated) timestamp attribute every ~30 days is in fact better
+than mandating password changes every 90 days as the latter would imply
+updating five (replicated) attributes at least once in that 90-day period.
```
rcritten: I'm not sure I'd characterize it as better because password changes may be due to compliance so it out of the hands of the user.

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,291 @@
...
+
+The Disable Stale Users (DSU) tooling, when the above mechanism is active,
+proposes to make sure no non-locked account has a timestamp attribute 90 days
+(by default) in the past.
+
+
+### Passwordless environments
```
rcritten: Should it be environments or authentications? I know what you're talking about, I'm not sure someone looking at this for the first time will immediately understand what it means.

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,291 @@
...
+the probability of that happening is low.
+
+
+## Example
+
+### Environments with mandatory, periodic password changes
+
+Given the following requirements and facts:
+* stale accounts must be disabled after 90 days' inactivity
+* grace period is decided to be 8 days
```
rcritten: If the default grace period is 7 days why not use that for the calculation?

doc/designs/disable-stale-users.md
rcritten: I think RandomizedTimeInterval needs to be expanded on as it is fixed at 7 by default but random?

doc/designs/disable-stale-users.md
fcami: Both of them now, thanks!

doc/designs/disable-stale-users.md
fcami: good catch, thanks

doc/designs/disable-stale-users.md
fcami: thanks, I had a hard time conveying this.

doc/designs/disable-stale-users.md
fcami: I removed the dash, for many reasons it was a bad idea.
I'm expanding the paragraph.

doc/designs/disable-stale-users.md
fcami: Fixed, thanks.

doc/designs/disable-stale-users.md
fcami: Merged into the Notes paragraph for clarity.

doc/designs/disable-stale-users.md
fcami: Fixed, thanks

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,273 @@
...
+each authentication event must be avoided as the update frequency could easily
+be high, (see https://pagure.io/freeipa/issue/5313 for background
+information), a coarse update mechanism is proposed.
+
+An attribute containing a timestamp is added to each user:
+coarseSuccessfulAuthTime. The timestamp contained therein is not precise.
+It will be added at account creation time (set to current date).
+The timestamp is updated by IPA plugins whenever a successful
+authentication event happens, but only if the current timestamp is
+sufficiently in the past. The actual
```
rcritten: Did you mean to add more or drop "The actual"?

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,273 @@
...
+* the first one to activate the mechanism (default: not present):
+  ipaConfigString: DSU:RecordSuccessfulAuth
+* the second attribute contains the fixed time period (default: 28 days):
+  ipaDSURecordSucessfulAuthTimeInterval: 2419200
+* the third one contains a random time period to be added. The actual value
+  used is randomized between 0 and the attribute's value (default: 7 days)
+  whenever a user successfully authenticates. 
+  ipaDSURecordSucessfulAuthRandomizedTimeInterval: 604800
+
+The time period to take into account is 28+7 so 35 days.
```
rcritten: It's range of 28-35 days, right?

It's a bit unclear how to formulate this as both 61 and 55 days are mentioned. 55 caps at 90 days so i assume that's what you mean. But this means some accounts could be disabled early.

28 + 55 = 83
29 + 55 = 84
30 + 55 = 85
...
35 + 55 = 90

So the range is 83-90 based on the randomized time interval. The user could configure that shorter or longer at the cost of potentially more accounts needing updates at the same time. The bigger the spread the more spread out writes are.

28 + 61 = 89
29 + 61 = 90
...
35 + 61 = 96

The text is reasonable but for quick reads it would help IMHO to include the formula for determining the expiration.

The days constant is 90 - coarseSuccessfulAuthTime + ipaDSURecordSucessfulAuthRandomizedTimeInterval

You also need an attribute to store the days-until-expiration value (90 by default).

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,273 @@
...
+period.
+
+The Disable Stale Users (DSU) tooling, when the above mechanism is active,
+proposes to make sure no non-locked account has a timestamp attribute 90 days
+(by default) in the past.
+
+
+### Notes
+
+Installing the DSU tooling will not change the behavior of existing clusters.
```
rcritten: nit: Enabling instead of Installing?

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,273 @@
...
+DSU has to be explicitely configured and enabled to start disabling stale
```
rcritten: explicitely -> explicitly

doc/designs/disable-stale-users.md
```diff
@@ -0,0 +1,273 @@
...
+attribute. But using a longer period runs the risk of disabling users who have
+been gone for too short a time. For instance, setting the update period to 65
+days would disable accounts inactive for the last 25 to 90 days. Whether this
+fits the site's requirements is up to the administrators to determine.
+
+Sites requiring a better accuracy must take care not to set the update period
+too low for performance reasons. It is strongly recommended to make sure the
+update period is longer than the expected maximum PTO or simply "away from IPA"
+period for most user accounts. This is especially true when these away periods
+are in sync, like in schools.
```
rcritten: Implicitly because then all passwords will be reset within a short time frame right? Hence lots of writes which == lesser perf?

doc/designs/disable-stale-users.md
fcami: Installing. This means that upgrading a cluster from a non-DSU-aware one to DSU nothing will change.

doc/designs/disable-stale-users.md
fcami: No, I meant to explain more. Doin' so.

doc/designs/disable-stale-users.md
fcami: re: "You also need an attribute to store the days-until-expiration value (90 by default)."
=> This is explained later, see TimeStampBasedDSUInactivePeriod.

fcami: Q: "It's range of 28-35 days, right?"
A: yes. I've added "Maximum" to the above sentence

fcami: I'm adding a formula.

doc/designs/disable-stale-users.md
fcami: Yes. Do you think this needs +v?

doc/designs/disable-stale-users.md
rcritten: OK

rcritten:
My concerns have been addressed, ack.

doc/designs/disable-stale-users.md
fcami: verbose mode added.

fcami:
master:

* 438094f868e1a96444e148d9a2481b559cab9a11 DSU: add Design for Disable Stale Users

PR #3826, opened 2019-10-25
conditionally restart certmonger during ipa-client-install

rcritten:
Conditionally restart certmonger after client installation

If certmonger is running prior to client installation then its 
IPA CA configuration will be incomplete and missing the CA chain.
    
If a certificate is subsequently requested with -F to store the 
CA chain in a file or NSS db it may not be available yet. A
conditional restart of certmonger will pick up the new IPA 
configuration and complete the IPA CA configuration in certmonger.
    
A pure restart and service activation is not done since certmonger
is not required unless --request-cert was passed ipa-client-install.
    
https://pagure.io/freeipa/issue/8105

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -779,7 +779,16 @@ def configure_krb5_conf(
 def configure_certmonger(
         fstore, subject_base, cli_realm, hostname, options, ca_enabled):
 
+    cmonger = services.knownservices.certmonger
     if not options.request_cert:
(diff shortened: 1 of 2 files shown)


rcritten:
master:

* 1e3de172696a98914f77e50fd64bc8b63360cd27 Add conditional restart (try-restart) capability to services
* 3593e5362206fd9287e97fe2cee229a1117811ff Conditionally restart certmonger after client installation

PR #3841, opened 2019-11-01
Enable AES SHA 256 and 384-bit enctypes in Kerberos

rcritten:
https://pagure.io/freeipa/issue/8110


tiran:
LGTM

The constants are the short aliases for RFC 8009 enc types

rcritten:
master:

* 09d5b938c128d8bb01ae40b5d736a266c6075b39 Enable AES SHA 256 and 384-bit enctypes in Kerberos

PR #3846, opened 2019-11-04
Block camellia in krbenctypes update in FIPS

tiran:
Add FIPS conditional to updates to prevent updater from adding camellia
encsalttypes.

Fixes: https://pagure.io/freeipa/issue/8111

### DIFF
--- a/install/updates/50-krbenctypes.update
+++ b/install/updates/50-krbenctypes.update
@@ -1,8 +1,8 @@
 dn: cn=$REALM,cn=kerberos,$SUFFIX
-add: krbSupportedEncSaltTypes: camellia128-cts-cmac:normal
-add: krbSupportedEncSaltTypes: camellia128-cts-cmac:special
-add: krbSupportedEncSaltTypes: camellia256-cts-cmac:normal
-add: krbSupportedEncSaltTypes: camellia256-cts-cmac:special
+${FIPS}add: krbSupportedEncSaltTypes: camellia128-cts-cmac:normal
+${FIPS}add: krbSupportedEncSaltTypes: camellia128-cts-cmac:special
+${FIPS}add: krbSupportedEncSaltTypes: camellia256-cts-cmac:normal
+${FIPS}add: krbSupportedEncSaltTypes: camellia256-cts-cmac:special
 add: krbSupportedEncSaltTypes: aes128-sha2:normal
 add: krbSupportedEncSaltTypes: aes128-sha2:special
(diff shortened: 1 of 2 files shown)


rcritten:
LGTM, will ack once CI is green.

install/updates/50-krbenctypes.update
```diff
@@ -1,8 +1,8 @@
 dn: cn=$REALM,cn=kerberos,$SUFFIX
-add: krbSupportedEncSaltTypes: camellia128-cts-cmac:normal
-add: krbSupportedEncSaltTypes: camellia128-cts-cmac:special
-add: krbSupportedEncSaltTypes: camellia256-cts-cmac:normal
-add: krbSupportedEncSaltTypes: camellia256-cts-cmac:special
+${FIPS}add: krbSupportedEncSaltTypes: camellia128-cts-cmac:normal
+${FIPS}add: krbSupportedEncSaltTypes: camellia128-cts-cmac:special
+${FIPS}add: krbSupportedEncSaltTypes: camellia256-cts-cmac:normal
+${FIPS}add: krbSupportedEncSaltTypes: camellia256-cts-cmac:special
 add: krbSupportedEncSaltTypes: aes128-sha2:normal
```
rcritten: This isn't going to work. It doesn't need the {} wrapper and either way it is going to fail because this translates into # add and the skip for leading comments has already passed so it tries to evaluate the command '# add' which is unknown so the update fails. Additional changes in the updater will be necessary.

rcritten:
I think this will do it. Templatize the line to see if it is a comment early on. Shouldn't be too costly and we throw away the result.

diff --git a/ipaserver/install/ldapupdate.py b/ipaserver/install/ldapupdate.p
y
index 667ebacfb..bd7800569 100644
--- a/ipaserver/install/ldapupdate.py
+++ b/ipaserver/install/ldapupdate.py
@@ -479,7 +479,9 @@ class LDAPUpdate:
             source_line = source_line.rstrip()
 
             # skip comments and empty lines
-            if source_line.startswith('#') or source_line == '':
+            sub_line = self._template_str(source_line)
+            if (source_line.startswith('#') or source_line == '' or
+                    sub_line.startswith('#')):
                 continue
 
             state = None

rcritten:
master:

* bc56642bf9e59821322903e7470be6ec1d857034 Block camellia in krbenctypes update in FIPS
* 560acf37482fe5c7df1abd2d087e33a18b1cbe38 Skip commented lines after substitution

PR #3857, opened 2019-11-05
[WIP] WebUI tests: Fix ElementClickInterceptedException after adding an entity

serg-cymbaluk:
When some entities are being added notification appears and intercepts buttons
on the facet.
The fix allows to click button one more time in such situation.

Ticket: https://pagure.io/freeipa/issue/8112

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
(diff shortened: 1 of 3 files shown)


pvoborni:
It will do the job but is it a good thing to have the code for closing notifications in every click call? It makes the behavior less clear and it might make impossible to test some scenarios. 

Why not use the method which was used previously - if we expect that there can be a notifications from previous calls then let's close them explicitly.  Alternatively we can change web ui to not have the notification stretched across whole width of the page.

serg-cymbaluk:
@pvoborni @wladich 
The issue has been fixed in other way: https://github.com/freeipa/freeipa/pull/3904
Closing the PR...

PR #3864, opened 2019-11-06
smartcard: make the ipa-advise script compatible with authselect/auth…

flo-renaud:
…config

"ipa-advise config-client-for-smart-card-auth" is run on a server and
creates a script that needs to be copied and executed on a client.
The client may be of a different version and use authconfig instead of
authselect. The generated script must be able to handle both cases
(client using authselect or client using authconfig).

The patch checks whether authselect is available and calls the proper
configuration command (authselect or authconfig) depending on its
availability on the client.

Fixes: https://pagure.io/freeipa/issue/8113

### DIFF
--- a/ipaserver/advise/plugins/smart_card_auth.py
+++ b/ipaserver/advise/plugins/smart_card_auth.py
@@ -319,8 +319,20 @@ class config_client_for_smart_card_auth(common_smart_card_auth_config):
         )
 
     def run_authselect_to_configure_smart_card_auth(self):
+        # In order to be compatible with all clients, we check first
+        # if the client supports authselect.
+        # Otherwise authconfig will be used.
+        self.log.comment('Use either authselect or authconfig to enable '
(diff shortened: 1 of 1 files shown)


rcritten:
I think a dependency on which may be required.

flo-renaud:
@rcritten 
thanks for the suggestion. I replace the `if which authselect` check with `if [ -f /usr/bin/authselect ]`.

rcritten:
I hadn't really thought this through last night but I guess these scripts are already broken on non-redhat/fedora bases anyway since they use authselect/authconfig directly so this is probably ok. On non-redhat platforms paths.AUTHSELECT == None.

LGTM.

flo-renaud:
master:

* 87c24ebd34d4524b73d878f1298232547cafc5ba smartcard: make the ipa-advise script compatible with authselect/authconfig

PR #3868, opened 2019-11-07
 ipatests: replace ad hoc backup with FileBackup helper

wladich:
Test test_smb_mount_and_access_by_different_users was failing with message
    ```
    kdestroy: Permission denied while initializing krb5
    ```
    
    This happened because the previous test
    `test_smb_access_for_ad_user_at_ipa_client` was calling the fixture
    `enable_smb_client_dns_lookup_kdc` which was doing backup of krb5.conf
    in a wrong way:
    - mktemp (to create a temp file)
    - cp /etc/krb5.conf to the temp file
    - ...
    - mv tempfile /etc/krb5.conf
    
    This flow looses the file permissions, because mktemp creates a file
    using the default umask, which results in -rw------- permissions.
    The copy does not modify the permissions, and the mv keeps the
    permissions from the source => /etc/krb5.conf now has -rw-------.
    
    Fixes: https://pagure.io/freeipa/issue/8115

### DIFF
--- a/ipatests/prci_definitions/gating.yaml
+++ b/ipatests/prci_definitions/gating.yaml
@@ -262,3 +262,15 @@ jobs:
         template: *ci-master-f30
         timeout: 1800
         topology: *ipaserver
+
+  fedora-30/test_smb:
+    requires: [fedora-30/build]
+    priority: 50
+    job:
+      class: RunADTests
+      args:
+        build_url: '{fedora-30/build_url}'
+        test_suite: test_integration/test_smb.py
+        template: *ci-master-f30
+        timeout: 7200
+        topology: *ad_master_2client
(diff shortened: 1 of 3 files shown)


flo-renaud:
@wladich 
Thanks for the fix, LGTM. If you could fix the typo noted by @frozencemetery in the commit msg that would be great.

wladich:
I was sure I fixed it yesterday, but something went wrong. Fixed now.

wladich:
Fix for FileBackup is backported to ipa-4-7 in #3883

abbra:
LGTM.

wladich:
master:

* 72540c42337cacab1dcbf0a4b2b03d59512e7956 ipatests: refactor FileBackup helper
* c2b230ce64bb7a29572cbfabbd23dd0900c7fce3 ipatests: replace ad hoc backup with FileBackup helper
* f58fb573d18808343e30a59304c5d3862b4e7e42 ipatests: enable test_smb.py in gating.yaml

PR #3883, opened 2019-11-11
[Backport][ipa-4-7] ipatests: refactor FileBackup helper

wladich:
This is a partial backport of #3868 to fix file attributes issues in FileBackup

* `cp` now preserves all attributes of original file, there is no reason
  to select only some of them
* backup is now restored with `mv` instead of `cp` to avoid leaving junk


wladich:
dnssec has a known failure: https://pagure.io/freeipa/issue/7985

wladich:
ACKing PR as it is a backport

wladich:
ipa-4-7:

* f92c28da7bbc03c2ef47f0617e8dabbdfd0bf65e ipatests: refactor FileBackup helper

PR #3888, opened 2019-11-12
Update selinux-policy minimal requirement

cjeanner:
Since 6c2710446718828e6840ac34ea6fc704ae6790db we need a new selinux
policy in order to ensure /etc/named directory content has the correct
selinux flags.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -45,7 +45,7 @@
 %global python_netaddr_version 0.7.16
 # Require 4.7.0 which brings Python 3 bindings
 %global samba_version 4.7.0
-%global selinux_policy_version 3.14.1-14
+%global selinux_policy_version 3.14.3-21
 %global slapi_nis_version 0.56.1-4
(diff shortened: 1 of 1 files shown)


abbra:
Thanks, @cjeanner. This pull request should be pushed only after PR CI templates get updated to F31.

abbra:
Azure Pipelines run failed as expected because it runs on F30 where selinux-policy 3.14.3-52 is not yet in stable updates. Looks like this change has to wait until it is pushed. Same for F31 ;)

cjeanner:
@abbra any news on that one? :)

tiran:
selinux-policy-3.14.3-52.fc30 is in stable. @cjeanner Could you please rebase and push your commit again to trigger a rebuild?

rcritten:
master:

* ae256fa52440da70c9c0943601a2d62caf5bb292 Update selinux-policy minimal requirement

PR #3893, opened 2019-11-12
DNParam: raise Exception when multiple values provided to a 1-val param

flo-renaud:
### DNParam: raise Exception when multiple values provided to a 1-val param

When ipa user-add-certmapdata is called with multiple --subject or
multiple --issuer, the DNParam's _convert_scalar method is called with
a tuple containing all the params and should raise an exception as the
--subject and --issuer are single-value params.

The DNParam _convert_scalar method internally calls the DN init method,
and the DN init method is able to create a DN from a tuple of RDNs.
As such, it won't raise exception if a tuple/list is provided.

Check that _convert_scalar is only provided a single element.

Fixes: https://pagure.io/freeipa/issue/8097

### XMLRPCtest: add a test for add-certmapdata with multiple subject/issuer
    
ipa user-add-certmapdata defines --issuer and --subject as single valued.
Add a test checking that this is enforced.

Related: https://pagure.io/freeipa/issue/8097

### DIFF
--- a/ipalib/parameters.py
+++ b/ipalib/parameters.py
@@ -2015,6 +2015,10 @@ class DNParam(Param):
         if type(value) in self.allowed_types:
             return value
 
+        if type(value) in (tuple, list):
+            raise ConversionError(name=self.name,
+                                  error=ugettext(self.scalar_error))
+
         try:
             dn = DN(value)
         except Exception as e:
--- a/ipatests/test_ipalib/test_parameters.py
+++ b/ipatests/test_ipalib/test_parameters.py
@@ -46,6 +46,7 @@ from ipalib import parameters, text, errors, config, x509
 from ipalib.constants import TYPE_ERROR, CALLABLE_ERROR
 from ipalib.errors import ValidationError, ConversionError
 from ipalib import _
+from ipapython.dn import DN
 
 if six.PY3:
     unicode = str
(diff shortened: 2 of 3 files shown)


flo-renaud:
The lint failure is linked to 2 commits not synchronized (c62bd16 prci: rename definitions files and jobs to change how fedora releases are referenced and f58fb57  ipatests: enable test_smb.py in gating.yaml). PR #3894 should fix the issue.

flo-renaud:
Rebased after PR #3898 has been merged (prci: bump template version and fix test_smb gating definition)

rcritten:
The change looks ok but I think the test would be better/clearer in ipatests/test_ipalib/test_parameters

DNParam is used in a lot more places than just certmap.

flo-renaud:
Hi @rcritten 
thanks for the review. I added a new test in test_ipalib/test_parameters.py

rcritten:
Looks good, ack.

flo-renaud:
master:

* e08a6de6afc49aabf817b11b878f6765f662e1ff DNParam: raise Exception when multiple values provided to a 1-val param
* ecdd7dae19e84b33c65bf6203a2cd906a9eebc4e XMLRPCtest: add a test for add-certmapdata with multiple subject/issuer
* 7893fb9cb1fc7c78c78079b23756bca53268caa1 test_ipalib: add test for DNParam class

PR #3783, opened 2019-10-11
Migrate xunit/unittest/nose tests to Pytest

stanislavlevin:
Even though Pytest supports xunit style setups, unittest and nose
tests, this support is limited and may be dropped in the future
releases. Worst of all is that the mixing of various test
frameworks results in weird conflicts and of course, is not widely
tested.

This is a part of the work to remove the mixing of test idioms in the
IPA's test suite:

- replace xunit style
- employ the fixtures' interdependencies
- replace unittest.TestCase subclasses
- replace unittest test controls (SkipTest, fail, etc.)
- replace unittest assertions

Fixes: https://pagure.io/freeipa/issue/7989

PR #3896, opened 2019-11-13
Fix azure pipelines to use Fedora 31

abbra:


abbra:
Should work now. Ready for review.

stanislavlevin:
pytest-4.6.6 ...

abbra:
Any suggestion, @stanislavlevin ?

stanislavlevin:
@abbra, there are several options in my opinion:
1) https://github.com/freeipa/freeipa/pull/3783 (need to rebase and recheck)
2) if #3783 is not suitable for some reason, I think it's possible to write some workaround
to address only spotted issues
3) It looks like integration tests passed, this means we could use PyPI packages installed via `pip` on azure as a temporary workaround for F31.

abbra:
@stanislavlevin I'd like option (1). Could you  please rebase it and add commits from this one? I think it is better to have your PR with my two commits than the other way around.

stanislavlevin:
> @stanislavlevin I'd like option (1). Could you please rebase it and add commits from this one? I think it is better to have your PR with my two commits than the other way around.

Sure.
PR relay race  :)

stanislavlevin:
Azure tests passed in https://github.com/freeipa/freeipa/pull/3783

abbra:
This PR can now be closed because an alternative one which included parts of this PR was merged already.

PR #3901, opened 2019-11-14
Enable test_winsyncmigrate in nighlty tests

wladich:


wladich:
The tests fail due to error in PR-CI: vagrant fails to set hostname in Windows VMs due to WinRM issue, as a result the certificate file name does not meet test expectations.

PR #3907, opened 2019-11-16
Add config that maintains existing content to ipa-client-install manpage

miztake:
If --no-ssh and --no-sshd are not specified in ipa-client-install,
/etc/ssh/{ssh, sshd}_config is updated and existing content is maintained.

### DIFF
--- a/client/man/ipa-client-install.1
+++ b/client/man/ipa-client-install.1
@@ -272,6 +272,15 @@ Files updated, existing content is maintained:
 .br
 /etc/sysconfig/network
 
+.TP
+File updated, existing content is maintained if ssh is configured (default):
+
+/etc/ssh/ssh_config
+.TP
+File updated, existing content is maintained if sshd is configured (default):
+
+/etc/ssh/sshd_config
+
 .SH "DEPRECATED OPTIONS"
 .TP
 \fB\-\-request\-cert\fR


rcritten:
This seems to make sense. The ssh and sshd config files are currently constants in ipaplatform so conceivably they could vary by platform so we'll need to keep an eye on that but AFAICT nobody has overridden it yet, at least not upstream.

Adding re-run to execute full CI.

stanislavlevin:
@rcritten, you are right, I override ssh directory as:
```
    SSH_CONFIG_DIR = "/etc/openssh"                                             
    SSH_CONFIG = "/etc/openssh/ssh_config"                                      
    SSHD_CONFIG = "/etc/openssh/sshd_config"
```

rcritten:
@stanislavlevin ok is that something you want to upstream or are you ok patching this man page as well?

stanislavlevin:
@rcritten , I'll be happy if the man page is patched too. But I realize that not all the platform-specific things could be implemented. And if it is difficult for some reason to make such a substitution I'm OK with changes.

stanislavlevin:
I mean I could patch this for myself in downstream.

rcritten:
Ok, I'm going to approve this. It seems like overkill to add templating to man pages that use ipaplatform for just this use-case.

tiran:
master:

* 650db3b80d5f1ec68131d0e17993fb31f9488d7c Add config that maintains existing content to ipa-client-install manpage

PR #3929, opened 2019-11-22
Nightly definition: use right template for krbtpolicy

flo-renaud:
The ipaserver template triggers the installation of IPA server
before the tests are launched and should not be used for
test_integration tests

Switch to master_1repl template.

Related: https://pagure.io/freeipa/issue/8001


flo-renaud:
Manual backport in #3941.

PR #3943, opened 2019-11-26
ipatests: generic uninstall should call ipa server-del

flo-renaud:
### ipatests: generic uninstall should call ipa server-del

At the end of any integration test, the method uninstall is called and
uninstalls master, replicas and clients.
Usually the master is the CA renewal master and DNSSec master, and
uninstallation may fail.
This commits modifies the uninstall method in order to:
- call 'ipa server-del replica' before running uninstall on a replica
- uninstall the replicas before uninstalling the master

Fixes: https://pagure.io/freeipa/issue/7985

### ipatests: fix teardown

The uninstall method of some tests can be skipped as the cleanup is
already done before.

### ipatests: fix test_crlgen_manage

The goal of the last test in test_crlgen_manage is to ensure that
ipa-server-install --uninstall can proceed if the server is the last one
in the topology, even if it is the CRL generation master.

The current code is wrong because it tries to uninstall the master
(which has already been uninstalled in the prev test), It should rather
uninstall replicas[0].

### DIFF
--- a/ipatests/test_integration/base.py
+++ b/ipatests/test_integration/base.py
@@ -20,6 +20,7 @@
 """Base class for FreeIPA integration tests"""
 
 import pytest
+import subprocess
 
 from ipatests.pytest_ipa.integration import tasks
 from pytest_sourceorder import ordered
@@ -97,9 +98,17 @@ class IntegrationTest:
 
     @classmethod
     def uninstall(cls, mh):
-        tasks.uninstall_master(cls.master)
         for replica in cls.replicas:
+            try:
+                tasks.run_server_del(
+                    cls.master, replica.hostname, force=True,
+                    ignore_topology_disconnect=True, ignore_last_of_role=True)
+            except subprocess.CalledProcessError:
+                # If the master has already been uninstalled,
+                # this call may fail
+                pass
             tasks.uninstall_master(replica)
+        tasks.uninstall_master(cls.master)
         for client in cls.clients:
             tasks.uninstall_client(client)
         if cls.fips_mode:
--- a/ipatests/test_integration/test_crlgen_manage.py
+++ b/ipatests/test_integration/test_crlgen_manage.py
@@ -290,6 +290,10 @@ class TestCRLGenManage(IntegrationTest):
         expected_msg = "Deleting this server will leave your installation " \
                        "without a CRL generation master"
         assert expected_msg in result.stdout_text
+        tasks.run_server_del(
+            self.replicas[0], self.master.hostname, force=True,
+            ignore_topology_disconnect=True,
+            ignore_last_of_role=True)
 
     def test_uninstall_last_master_does_not_require_ignore_last_of_role(self):
         """Test uninstallation of the last master
@@ -301,4 +305,5 @@ class TestCRLGenManage(IntegrationTest):
         # Make sure CRL gen is enabled on the replica
         check_crlgen_enable(self.replicas[0])
         # call uninstall without --ignore-last-of-role, should be OK
-        self.master.run_command(['ipa-server-install', '--uninstall', '-U'])
+        self.replicas[0].run_command(
+            ['ipa-server-install', '--uninstall', '-U'])
--- a/ipatests/test_integration/test_ntp_options.py
+++ b/ipatests/test_integration/test_ntp_options.py
@@ -388,3 +388,8 @@ class TestNTPoptions(IntegrationTest):
         tasks.uninstall_client(self.client)
         tasks.uninstall_master(self.replica)
         tasks.uninstall_master(self.master)
+
+    @classmethod
+    def uninstall(cls, mh):
+        # Cleanup already done in teardown_method
+        pass


flo-renaud:
A successful run with all the nightly tests can be seen [here](https://github.com/freeipa-pr-ci2/freeipa/pull/78).

tiran:
Do you want to backport the fix to 4.8, too?

ipatests/test_integration/test_ntp_options.py
```diff
@@ -388,3 +388,8 @@ def teardown_method(self, method):
         tasks.uninstall_client(self.client)
         tasks.uninstall_master(self.replica)
         tasks.uninstall_master(self.master)
+
+    @classmethod
```
wladich: There is nothing special in this `teardown_method` and it is not a part of test suite, so maybe it is better to remove it and let the base `uninstall` do the job.

ipatests/test_integration/test_crlgen_manage.py
```diff
@@ -302,3 +302,8 @@ def test_uninstall_last_master_does_not_require_ignore_last_of_role(self):
         check_crlgen_enable(self.replicas[0])
         # call uninstall without --ignore-last-of-role, should be OK
         self.master.run_command(['ipa-server-install', '--uninstall', '-U'])
+
```
wladich: `test_uninstall_last_master_does_not_require_ignore_last_of_role` is a part of testsuite, so we expect that once it can fail (i.e. if uninstallation without `--ignore-last-of-role` does not work). 
In this case proper uninstallation will be needed

ipatests/test_integration/test_uninstallation.py
```diff
@@ -119,3 +119,8 @@ def test_failed_uninstall(self):
             self.master.run_command([
                 paths.DSCTL, serverid, 'remove', '--do-it'
             ])
+
+    @classmethod
+    def uninstall(cls, mh):
```
wladich: Replica and master uninstallation are parts of testsuite, so we expect that once they can fail. In this case proper uninstallation will be needed.

wladich:
Hi @flo-renaud 
Please see the inline comments.

ipatests/test_integration/test_ntp_options.py
flo-renaud: This method is called after each test inside the class and must be kept (the uninstall base method is called only at the end of the class, not at the end of each test).

ipatests/test_integration/test_uninstallation.py
flo-renaud: The issue is the following: if a test uninstalls the master, the base `uninstall` method is likely to fail as it tries to run ipa server-del on the master node. That's why I came up with this solution.
If you have a solution in mind, I would be happy to implement it.

flo-renaud:
@tiran 
thanks for the review. Yes, a backport to ipa-4-8 would be nice, too.

ipatests/test_integration/test_ntp_options.py
wladich: You are right, I should me more attentive.

ipatests/test_integration/test_uninstallation.py
wladich: Maybe add raiseonerr=False to invocation of  server-del or specify two valid codes in ok_returncode?

flo-renaud:
Hi @wladich 
I slightly modified the base uninstall method to ignore exceptions on ipa server-del. With this modif, no need to override the uninstall method of test_crlgen_manage.py and test_ntp_options.py.
To validate the fix, I added a temp commit with the above tests.

flo-renaud:
@wladich 

ipatests/test_integration/base.py
```diff
@@ -97,9 +98,17 @@ def teardown_class(cls):
...
-        tasks.uninstall_master(cls.master)
         for replica in cls.replicas:
+            try:
+                tasks.run_server_del(
+                    cls.master, replica.hostname, force=True,
+                    ignore_topology_disconnect=True, ignore_last_of_role=True)
+            except subprocess.CalledProcessError:
+                # If the master has already been uninstalled,
+                # this call may fail
+                pass
```
stanislavlevin: @flo-renaud, hi.
Is it possible to narrow down an exception range to the re-uninstallation error only?

flo-renaud: Hi @stanislavlevin 
`tasks.run_server_del` will try to execute `kinit admin` on the master. If the master was already uninstalled, kinit fails and returns CalledProcessError.

stanislavlevin: Since the current implementation of `run_server_del` just tries to perform `ipa server-del`, would it be better to pass `raiseonerr=False` down to `kinit_admin` too?

For example,

```diff
ipatests/pytest_ipa/integration/tasks.py:
 def run_server_del(host, server_to_delete, force=False,
                    ignore_topology_disconnect=False,
                    ignore_last_of_role=False):
-    kinit_admin(host)
+    kinit_admin(host, raiseonerr=False)
     args = ['ipa', 'server-del', server_to_delete]
     if force:

ipatests/test_integration/base.py:
-            try:
-                tasks.run_server_del(
-                    cls.master, replica.hostname, force=True,
-                    ignore_topology_disconnect=True, ignore_last_of_role=True)
-            except subprocess.CalledProcessError:
-                # If the master has already been uninstalled,
-                # this call may fail
-                pass
+            tasks.run_server_del(
+                cls.master, replica.hostname, force=True,
+                ignore_topology_disconnect=True, ignore_last_of_role=True)
```

tiran:
master:

* 7dfc6e004be1dd1b3e2eee51786f3b9f888f3494 ipatests: generic uninstall should call ipa server-del
* 8cf4271aaeb466aa73a44f6c162ddbe9eac7fc88 ipatests: fix teardown
* b3d650370cfa181a68c7ce332a0013ef2e23ad21 ipatests: fix test_crlgen_manage

ipatests/test_integration/base.py
flo-renaud: Hi @stanislavlevin 
with your suggestion, the behavior of run_server_del is modified when called on an uninstalled node: currently it's raising a CalledProcessError exception, with the change it would return the result code 1. That's why I preferred to handle the exception in the function base.py:uninstall.

PR #3951, opened 2019-11-27
Integration tests for OTP token

tiran:
Add integration tests to verify HOTP, TOTP, service with OTP auth indicator, and OTP token sync.

Fix otptoken_sync plugin

The plugin had two bugs:

For one it did not work under Python 3 because urlencode() returns a string but HTTPSHandler expects bytes as data argument.

The primary key field name is not available in client plugins. Just pass the token name and let server code convert the name to DN.

Fixes: https://pagure.io/freeipa/issue/7804

### DIFF
--- a/ipaclient/plugins/otptoken.py
+++ b/ipaclient/plugins/otptoken.py
@@ -26,7 +26,6 @@ from ipalib.messages import add_message, ResultFormattingError
 from ipalib.plugable import Registry
 from ipalib.frontend import Local
 from ipalib.util import create_https_connection
-from ipapython.dn import DN
 from ipapython.version import API_VERSION
 
 import locale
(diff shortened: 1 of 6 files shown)


tiran:
Successful temp commit with four tests: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/795d92de-11ce-11ea-a355-fa163ef3ea43

abbra:
LGTM.

tiran:
master:

* 095d3f9bc94e5d43f5ebb89b88ee87b330af5c1d Add test case for OTP login
* e8b98555fcbdc76c2e728d0c93cbbabeebad9a26 Fix otptoken_sync plugin

PR #3954, opened 2019-11-28
[Backport][ipa-4-8] Integration tests for OTP token

tiran:
Manual backport of PR #3951 

Add integration tests to verify HOTP, TOTP, service with OTP auth indicator, and OTP token sync.

Fix otptoken_sync plugin

The plugin had two bugs:

For one it did not work under Python 3 because urlencode() returns a string but HTTPSHandler expects bytes as data argument.

The primary key field name is not available in client plugins. Just pass the token name and let server code convert the name to DN.

Fixes: https://pagure.io/freeipa/issue/7804

### DIFF
--- a/ipaclient/plugins/otptoken.py
+++ b/ipaclient/plugins/otptoken.py
@@ -26,7 +26,6 @@ from ipalib.messages import add_message, ResultFormattingError
 from ipalib.plugable import Registry
 from ipalib.frontend import Local
 from ipalib.util import create_https_connection
-from ipapython.dn import DN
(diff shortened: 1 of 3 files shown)


tiran:
backport ack

tiran:
ipa-4-8:

* 90f2866163b6fb4ec7f533e8d4820089b804d707 Fix otptoken_sync plugin
* dfa356e3d60468fab9eb873adfa0b08a7f8fd842 Add test case for OTP login

PR #3956, opened 2019-11-28
Covscan fixes

abbra:
Memory leaks found by latest covscan runs.

### DIFF
--- a/client/ipa-getkeytab.c
+++ b/client/ipa-getkeytab.c
@@ -512,6 +512,7 @@ static int ldap_get_keytab(krb5_context krbctx, bool generate, char *password,
     if (enctypes) {
         ret = ipa_string_to_enctypes(enctypes, &es, &num_es, err_msg);
         if (ret || num_es == 0) {
+            free(es);
             return LDAP_OPERATIONS_ERROR;
         }
     }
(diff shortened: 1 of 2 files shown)


flo-renaud:
Hi @abbra 
thanks for the PR, LGTM.

tiran:
master:

* e3ad78538e1dd2f63f171ef1c2b470a1a4f47a8c covscan: free encryption types in case there is an error
* e9dd757763c76402e07f533f19e269eeebc554fa covscan: free ucs2-encoded password copy when generating NTLM hash

kdudka:
@abbra There seems to be also a typo detected by Covscan, which I think was waived incorrectly.  Please consider including this patch:
```
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -778,8 +778,8 @@ class CIDict(dict):
     def viewkeys(self):
         raise NotImplementedError('CIDict.viewkeys is not implemented')

-    def viewvvalues(self):
-        raise NotImplementedError('CIDict.viewvvalues is not implemented')
+    def viewvalues(self):
+        raise NotImplementedError('CIDict.viewvalues is not implemented')


 class GeneralizedTimeZone(datetime.tzinfo):
```

PR #3969, opened 2019-12-02
Fix logic of check_client_configuration

tiran:
The helper function ipalib.util.check_client_configuration() now
considers a client configured when either:

* confdir is overridden (e.g. with IPA_CONFDIR) and the conf_default
  file exists.
* confdir is /etc/ipa, /etc/ipa/default.conf exists and client
  sysrestore state exists.

The check for sysrestore state is faster than checking for the presence
of the directory and presence of files in the directory. The sysrestore
state is always presence. sysrestore.index may be missing if no files
were backed up.

Fixes: https://pagure.io/freeipa/issue/8133

### DIFF
--- a/ipalib/util.py
+++ b/ipalib/util.py
@@ -1164,12 +1164,31 @@ def check_client_configuration(env=None):
 
     Hardcode return code to avoid recursive imports
     """
-    if ((env is not None and not os.path.isfile(env.conf_default)) or
-       (not os.path.isfile(paths.IPA_DEFAULT_CONF) or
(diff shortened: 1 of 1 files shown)


rcritten:
nit in commit message: 

overwritten -> overridden
exists -> exist

I wonder if there is a valid configuration where ipa-client-install doesn't change any files therefore there is no sysrrestore.index. Would it be overkill to check sysrestore.state as well? Does this code  need to be performant (e.g. one extra stat)?

tiran:
master:

* cf9f9bb326a2efae2e459a15be5d17878cb40da0 Fix logic of check_client_configuration

PR #3970, opened 2019-12-02
Fix lite-server to work with GSS_NAME

tiran:
The lite-server does no longer work correctly since rpcserver is also
using GSS_NAME. Set up GSS_NAME from ccache.

### DIFF
--- a/contrib/lite-server.py
+++ b/contrib/lite-server.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # Copyright (C) 2017 FreeIPA Contributors see COPYING for license
 #
@@ -35,8 +35,7 @@ You may also have to enable a development COPR.
 
     $ sudo dnf install -y dnf-plugins-core
     $ sudo dnf builddep --spec freeipa.spec.in
-    $ sudo dnf install -y python-werkzeug python2-watchdog \
-        python3-werkzeug python3-watchdog
+    $ sudo dnf install -y python3-werkzeug python3-watchdog
     $ ./autogen.sh
 
 For more information see
@@ -61,6 +60,7 @@ from ipalib.errors import NetworkError
 from ipalib.krb_utils import krb5_parse_ccache
 from ipalib.krb_utils import krb5_unparse_ccache
 
+import gssapi
 # pylint: disable=import-error
 from werkzeug.contrib.profiler import ProfilerMiddleware
 from werkzeug.exceptions import NotFound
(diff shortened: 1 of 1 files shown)


contrib/lite-setup.py
```diff
@@ -0,0 +1,142 @@
...
+    args.ca_crt = os.path.join(args.dot_ipa, "ca.crt")
+    args.krb5_conf = os.path.join(args.dot_ipa, "krb5.conf")
+    args.ccache = os.path.join(args.dot_ipa, "ccache")
+    args.activate = os.path.join(args.dot_ipa, "activate.sh")
+
+    if not os.path.isdir(args.dot_ipa):
+        os.makedirs(args.dot_ipa, mode=0o750)
+
+    with urlopen(f"http://{args.servername}/ipa/config/ca.crt") as req:
+        ca_data = req.read()
```
rcritten: Why not just copy from /etc/ipa/ca.crt? It has to run locally because localhost is specified as the xmlrpc_url so we know that the file is there.

contrib/lite-setup.py
```diff
@@ -0,0 +1,142 @@
...
+}}
+
+[domain_realm]
+ .ipa.example = {args.realm}
+ ipa.example = {args.realm}
+ {args.servername} = {args.realm}
+"""
+
+ACTIVATE = """\
+deactivate_ipaenv () {{
```
rcritten: Would it be worthwhile to use deactivate like virtual-env? Would someone want to run this within a venv?

contrib/lite-setup.py
```diff
@@ -0,0 +1,142 @@
...
+[global]
+host = {args.hostname}
+server = {args.servername}
+basedn = {args.basedn}
+realm = {args.realm}
+domain = {args.domain}
+# xmlrpc_uri = https://{args.servername}/ipa/xml
+xmlrpc_uri = http://localhost:8888/ipa/xml
+ldap_uri = ldap://{args.servername}
+mode = production
```
rcritten: We don't do significant plugin creation these days but running in development mode is stricter than production mode.

That wouldn't be useful for performance testing though.

rcritten:
I've always run lite-server using two terminals: one for the server and one for the client. This is because I usually end up embedding pdb into server plugins to watch what is happening.

It would be great if the script had some sort of "client" mode which just set the environment variables (though perhaps allowing a separate ccache for the requesting user).

contrib/lite-setup.py
tiran: I have split up the PR and moved the new lite-setup to a different PR.

The script does not assume local configuration. The ``xmlrpc_uri`` is ``http://localhost:8888``, which is the hostname and port of a local lite-server setup.

contrib/lite-setup.py
tiran: I selected another name to avoid conflicts with virtual environments.

tiran:
master:

* dbfb011da7d79b1eb4ace7d8cb45aa8a86f3eaba Fix lite-server to work with GSS_NAME

PR #3971, opened 2019-12-02
Optimize user-add by caching ldap2.has_upg()

tiran:
The method has_upg returns if user private groups are enabled or
disabled. has_upg() is called three times by user-add. The setting is
now cached on the request local variable context to speed up batch
processing of user imports.

context is cleared after every request.

Related: https://pagure.io/freeipa/issue/8134


tiran:
master:

* dcb33e44e7f3125347341f6003bf1fff6055b433 Optimize user-add by caching ldap2.has_upg()

PR #3976, opened 2019-12-03
locations.py: fix divide by 0

fcami:
If all IPA servers (or the first server) of a location have
service weights == 0, the sum of the weights is zero leading
to a division by zero in relative weight calculation.
Detect the situation and set relative weight to 0 in this case.

Fixes: https://pagure.io/freeipa/issue/8135

### DIFF
--- a/ipaserver/plugins/location.py
+++ b/ipaserver/plugins/location.py
@@ -225,10 +225,17 @@ class location_show(LDAPRetrieve):
                     dns_servers.append(s_name)
 
             for server in servers_additional_info.values():
-                server['service_relative_weight'] = [
-                    u'{:.1f}%'.format(
-                        int(server['ipaserviceweight'][0])*100.0/weight_sum)
-                ]
+                if weight_sum != 0:
+                    server['service_relative_weight'] = [
+                        u'{:.1f}%'.format(
+                            int(server['ipaserviceweight'][0]) * 100.0 /
+                            weight_sum
+                        )
+                    ]
+                else:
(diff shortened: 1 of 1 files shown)


tiran:
1. Is 0% the correct answer? Should the sum of all ``ipaserviceweight`` be 100%?
2. There is no test case for weight of 0 yet.

fcami:
Hi @tiran I am working on the test case. As you said it is not clear if the correct ipaserviceweight is 0 or an even percentage split in this case.

fcami:
Hi @tiran I think you are right, an even percentage split (e.g. the sum of all ipaserviceweights = 100) sounds better to me.

ipaserver/plugins/location.py
```diff
@@ -225,10 +225,17 @@ def execute(self, *keys, **options):
...
+                if weight_sum != 0:
+                    server['service_relative_weight'] = [
+                        u'{:.1f}%'.format(
+                            int(server['ipaserviceweight'][0]) * 100.0 /
+                            weight_sum
+                        )
+                    ]
+                else:
+                    server['service_relative_weight'] = [
+                        u'{:.1f}%'.format(int(100.0 / len(servers)))
```
tiran: It doesn't make sense to combine ``int()`` with float formatting. ``.1f`` formats the variable as float with one item after the dot, e.g. ``33.3%`` for 3 servers.

rcritten:
Closing this, resolved in PR #4278

PR #3977, opened 2019-12-03
trust upgrade: ensure that host is member of adtrust agents

flo-renaud:
After an upgrade, the group cn=adtrust agents may be missing some members.
Each ad trust controller must appear twice as member:
- krbprincipalname=cifs/hostname@realm,cn=services,cn=accounts,basedn
- fqdn=hostname,cn=computers,cn=accounts,basedn

Add an upgrade plugin that builds a list of hostnames from the cifs
principals and adds if needed fqdn=hostname...

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1778777

### DIFF
--- a/install/updates/90-post_upgrade_plugins.update
+++ b/install/updates/90-post_upgrade_plugins.update
@@ -14,6 +14,7 @@ plugin: update_tdo_gidnumber
 plugin: update_tdo_to_new_layout
 plugin: update_host_cifs_keytabs
 plugin: update_tdo_default_read_keys_permissions
+plugin: update_adtrust_agents_members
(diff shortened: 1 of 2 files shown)


abbra:
LGTM. For adtrust controller, we add both cifs principal and host principal already in `ipaserver/install/adtrustinstance.py:ADTRUSTInstance.__setup_group_membership()`, so this should be enough to fix situations where host principal is missing.

flo-renaud:
master:

* 2c9b212cf08e9f0e6814b2e7a0922079b3929634 trust upgrade: ensure that host is member of adtrust agents

PR #3978, opened 2019-12-03
ipatests: fix TestMigrateDNSSECMaster teardown

flo-renaud:
The test is installing master +DNSSEC, then replica and migrates the DNSSEC
to the replica.
During teardown, the replica is removed with ipa server-del. This operation
deletes the entries cn=DNS and cn=DNSSEC on the master, but if the
replication is stopped before the operations are replicated on the replica,
the replica may end up with a dangling cn=DNSSEC entry and no cn=DNS entry.
In this case ipa-server-install --uninstall on the replica will fail.
    
The fix: uninstall the DNSSec master as the last step of teardown
    
Related: https://pagure.io/freeipa/issue/7985

### DIFF
--- a/ipatests/test_integration/test_dnssec.py
+++ b/ipatests/test_integration/test_dnssec.py
@@ -5,6 +5,8 @@
 from __future__ import absolute_import
 
 import logging
+import re
+import subprocess
 import time
 
 import dns.dnssec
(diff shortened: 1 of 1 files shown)


stanislavlevin:
Hi @flo-renaud, thanks for PR.
I've run  `test_dnssec` 10 times against your commit (really on `ipa-4-7` branch).
There are no more fails related to `invalid 'DNSSec key master'`.
https://dev.azure.com/slev0400/slev/_build/results?buildId=894&view=results
I could re-run to be sure.

tiran:
ACK, please remove the temp commit.

flo-renaud:
@stanislavlevin and @tiran 
thanks for the reviews. Temp commit removed.

tiran:
master:

* c1272e48dfaa6ccf15342f1888cba556026e6cf2 ipatests: fix TestMigrateDNSSECMaster teardown

PR #3999, opened 2019-12-09
Fix typo in idrange.py docstring

smartsammler:
In my opinion the sentence makes more sense if one replaces this `id` with a `is`.
If this is not what was meant, then I would be pleased if the sentence could be rephrased, so that I get it right. Thank you.

### DIFF
--- a/ipaclient/remote_plugins/2_114/idrange.py
+++ b/ipaclient/remote_plugins/2_114/idrange.py
@@ -120,7 +120,7 @@ for a different domain.
 
 (*) The RID is typically the last integer of a user or group SID which follows
 the domain SID. E.g. if the domain SID is S-1-5-21-123-456-789 and a user from
-this domain has the SID S-1-5-21-123-456-789-1010 then 1010 id the RID of the
+this domain has the SID S-1-5-21-123-456-789-1010 then 1010 is the RID of the
 user. RIDs are unique in a domain, 32bit values and are used for users and
 groups.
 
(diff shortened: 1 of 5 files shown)


abbra:
LGTM but please use correct email for contribution.

flo-renaud:
Hi @smartsammler 
Shouldn't the modification be done in ./ipaserver/plugins/idrange.py?

abbra:
@flo-renaud is correct -- we have it in the server plugin side. There are more places where it needs to be modified, btw:
```
ipaclient/remote_plugins/2_114/idrange.py:this domain has the SID S-1-5-21-123-456-789-1010 then 1010 id the RID of the
ipaclient/remote_plugins/2_156/idrange.py:this domain has the SID S-1-5-21-123-456-789-1010 then 1010 id the RID of the
ipaclient/remote_plugins/2_164/idrange.py:this domain has the SID S-1-5-21-123-456-789-1010 then 1010 id the RID of the
ipaclient/remote_plugins/2_49/idrange.py:this domain has the SID S-1-5-21-123-456-789-1010 then 1010 id the RID of the
ipaserver/plugins/idrange.py:this domain has the SID S-1-5-21-123-456-789-1010 then 1010 id the RID of the
````
(and translation strings in PO files but those will need a special handling during transition from Zanata to Weblate anyway)

smartsammler:
Sorry for the delay and thanks for the input.

abbra:
master:

* 273ff2708cb99dded5e7ef51504cc8de9a4496db Fix typo in idrange.py docstring

PR #4000, opened 2019-12-09
ipatests: Test if ipa-backup throws error if /var/lib/ipa/ runs out of space

mrizwan93:
ipa-backup throws error when /var/lib/ipa runs out of space. Earlier
the error was not so clear. Fix mentions about insufficient space.

related : https://pagure.io/freeipa/issue/7647

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
--- a/ipatests/prci_definitions/temp_commit.yaml
+++ b/ipatests/prci_definitions/temp_commit.yaml
@@ -64,7 +64,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-latest/build_url}'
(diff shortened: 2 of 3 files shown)


fcami:
hi @mrizwan93 
The relevant commit:
https://pagure.io/freeipa/c/e6415ec3210cda0808867642cb19f0df1ddd3c45
has multiple error strings depending on when the backup fails. I think all situations should be tested, not only the one where the filesystem is so full that backup cannot start at all.
Also could you please fix the commit message so that it conforms to https://www.freeipa.org/page/Contribute/Code#Commit_message_requirements
Thank you!
François

rcritten:
Some additional places that can cause disk full:

The backup is first done to os.path.join(paths.SLAPD_INSTANCE_LDIF_DIR_TEMPLATE % instance, ldifname)

That can fill up.

Then it is moved to a temporary directory, result of tempfile.mkdtemp("ipa"). That can fail.

Then it is tarred and that can fail.

There are also specific checks for online vs offline.

So you need to do the current test, and separately fill /tmp and /var/lib/dirsrv/ and do them online and offline.

mrizwan93:
Closing the PR as not able to reproduce all the error messages scenarios.

PR #4001, opened 2019-12-09
DNS install check: allow overlapping zone to be from the master itself

abbra:
When re-running `ipa-server-install --setup-dns` on already installed
server, we do not get to the check of being already installed because
DNS zone overlap forces us to fail earlier.

Change exception returned for this case from check_zone_overlap() to
return structured information that allows to understand whether we are
finding a conflict with ourselves.

Use the returned information to only fail DNS check at this point if DNS
zone overlap is generated by a different name server than ourselves.

### DIFF
--- a/ipapython/dnsutil.py
+++ b/ipapython/dnsutil.py
@@ -39,6 +39,12 @@ if six.PY3:
 logger = logging.getLogger(__name__)
 
 
+class DNSZoneAlreadyExists(dns.exception.DNSException):
+    supp_kwargs = {'zone', 'ns'}
+    fmt = (u"DNS zone {zone} already exists in DNS "
+           "and is handled by server(s): {ns}")
+
+
 @six.python_2_unicode_compatible
 class DNSName(dns.name.Name):
     labels = None  # make pylint happy
@@ -374,10 +380,7 @@ def check_zone_overlap(zone, raise_on_error=True):
                          zone, e)
             ns = []
 
-        msg = u"DNS zone {0} already exists in DNS".format(zone)
-        if ns:
-            msg += u" and is handled by server(s): {0}".format(', '.join(ns))
-        raise ValueError(msg)
+        raise DNSZoneAlreadyExists(zone=zone.to_text(), ns=ns)
 
 
 def _mix_weight(records):
--- a/ipaserver/install/bindinstance.py
+++ b/ipaserver/install/bindinstance.py
@@ -327,7 +327,7 @@ def read_reverse_zone(default, ip_address, allow_zone_overlap=False):
         if not allow_zone_overlap:
             try:
                 dnsutil.check_zone_overlap(zone, raise_on_error=False)
-            except ValueError as e:
+            except dnsutil.DNSZoneAlreadyExists as e:
                 logger.error("Reverse zone %s will not be used: %s",
                              zone, e)
                 continue
--- a/ipaserver/install/dns.py
+++ b/ipaserver/install/dns.py
@@ -130,13 +130,15 @@ def install_check(standalone, api, replica, options, hostname):
         domain = dnsutil.DNSName(util.normalize_zone(api.env.domain))
         try:
             dnsutil.check_zone_overlap(domain, raise_on_error=False)
-        except ValueError as e:
+        except dnsutil.DNSZoneAlreadyExists as e:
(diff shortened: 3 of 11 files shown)


abbra:
@t-woerner is going to add a test of re-running `ipa-server-install` to make sure we see `already installed` message.

t-woerner:
> @t-woerner is going to add a test of re-running `ipa-server-install` to make sure we see `already installed` message.

This is only happening if domain is set explicitly in the command line installer as check_zone_overlap is used in the domain_name validator.

abbra:
Tests succeeded, I removed temp commit.

ipapython/dnsutil.py
```diff
@@ -374,10 +380,7 @@ def check_zone_overlap(zone, raise_on_error=True):
                          zone, e)
             ns = []
 
-        msg = u"DNS zone {0} already exists in DNS".format(zone)
-        if ns:
-            msg += u" and is handled by server(s): {0}".format(', '.join(ns))
-        raise ValueError(msg)
+        raise DNSZoneAlreadyExists(zone.to_text(), ns)
```
tiran: This does not result in a proper error message. You must pass the arguments as keyword arguments:

```
>>> import dns.exception
>>> class DNSZoneAlreadyExists(dns.exception.DNSException):
...     supp_kwargs = {'zone', 'ns'}
...     fmt = (u"DNS zone {zone} already exists in DNS "
...            "and is handled by server(s): {ns}")
... 
>>> DNSZoneAlreadyExists("zone", "ns")
DNSZoneAlreadyExists('zone', 'ns')
>>> str(DNSZoneAlreadyExists("zone", "ns"))
"('zone', 'ns')"
>>> DNSZoneAlreadyExists("zone", "ns").msg is None
True
```

ipaserver/install/bindinstance.py
```diff
@@ -327,9 +327,9 @@ def read_reverse_zone(default, ip_address, allow_zone_overlap=False):
         if not allow_zone_overlap:
             try:
                 dnsutil.check_zone_overlap(zone, raise_on_error=False)
-            except ValueError as e:
+            except dnsutil.DNSZoneAlreadyExists as e:
                 logger.error("Reverse zone %s will not be used: %s",
-                             zone, e)
+                             zone, e.msg)
```
tiran: ``.msg`` is deprecated. ``DNSException`` sets ``self.msg = str(self)`` if keyword args are passed. ``e`` is good enough since the ``%s`` will call ``str(e)`` any way.

ipaserver/install/dns.py
```diff
@@ -130,13 +130,15 @@ def install_check(standalone, api, replica, options, hostname):
...
             if options.force or options.allow_zone_overlap:
                 logger.warning("%s Please make sure that the domain is "
                                "properly delegated to this IPA server.",
                                e)
-            else:
-                raise e
+
+            hst = dnsutil.DNSName(hostname).make_absolute().to_text()
+            if hst not in e.kwargs['ns']:
+                raise ValueError(e.msg)
```
tiran: ``raise ValueError(str(e))``

ipaserver/plugins/dns.py
```diff
@@ -2153,8 +2153,8 @@ def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys, **options):
         if not options['skip_overlap_check']:
             try:
                 check_zone_overlap(keys[-1], raise_on_error=False)
-            except ValueError as e:
-                raise errors.InvocationError(str(e))
+            except DNSZoneAlreadyExists as e:
+                raise errors.InvocationError(e.msg)
```
tiran: ``raise errors.InvocationError(str(e))``

tiran:
master:

* dd7fdaa77d00042d44bec23291f0d0be36bead5e DNS install check: allow overlapping zone to be from the master itself
* d070c5957791d19ece22082dbd69a319975a4b54 Test repeated installation of the primary with DNS enabled and domain set
* a2820bbbc3789cde2ecb9ea3eae91c1bd6f248ee Enable TestInstallMasterDNSRepeatedly in prci_definitions

AdamWill:
This was [missing a bit](https://bodhi.fedoraproject.org/updates/FEDORA-2019-ef8b8ce122#comment-1169537).

PR #4003, opened 2019-12-10
Require idstart to be larger than UID_MAX

tiran:
ipa-server-install fails if idstart is set to 0. There might be
additional issues when idstart overlaps with local users. Ensure that
idstart is larger than UID_MAX or GID_MAX from /etc/login.defs.

Fixes: https://pagure.io/freeipa/issue/8137

### DIFF
--- a/ipaserver/install/server/__init__.py
+++ b/ipaserver/install/server/__init__.py
@@ -5,10 +5,6 @@
 """
 Server installer module
 """
-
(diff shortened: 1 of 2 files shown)


abbra:
LGTM

tiran:
master:

* 44b3791bc3a14ede0708538347954871df6496fa Require idstart to be larger than UID_MAX

PR #4004, opened 2019-12-10
Check valid before/after of external certs

tiran:
verify_server_cert_validity() and verify_ca_cert_validity() now check
the validity time range of external certificates. The check fails if the
certificate is not valid yet or will expire in less than an hour.

Fixes: https://pagure.io/freeipa/issue/8142

### DIFF
--- a/ipatests/test_ipapython/test_certdb.py
+++ b/ipatests/test_ipapython/test_certdb.py
@@ -9,6 +9,7 @@ from ipapython import ipautil
 from ipaplatform.osinfo import osinfo
 
 CERTNICK = 'testcert'
+CERTSAN = 'testcert.certdb.test'
 
 if osinfo.id == 'fedora':
     if osinfo.version_number >= (28,):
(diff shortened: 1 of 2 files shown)


abbra:
LGTM.

tiran:
master:

* d30dd52920a06ded11b9dfe136108343bb2896a3 Check valid before/after of external certs

ipatests/test_ipapython/test_certdb.py
```diff
@@ -242,3 +244,14 @@ def test_delete_cert_and_key():
             assert len(nssdb.list_certs()) == 1
     finally:
         os.unlink(p12file)
+
+
+def test_check_validity():
```
frasertweedale: @tiran thanks for the PR.  My only comment is that this test (thanks for adding it) does not exercise the new NotBefore/NotAfter check.  Ideally we test that here, or in an integration test, or both.

PR #4005, opened 2019-12-10
Fix service ldap_disable()

tiran:
Fix comparison bug that prevents ldap_disable to actually disable a
service.

Fixes: https://pagure.io/freeipa/issue/8143


abbra:
LGTM.

tiran:
master:

* 3cae7f4ee6fbfed99930a87e78d7e37bdb740e43 Fix service ldap_disable()

PR #4012, opened 2019-12-10
[ipatests] Improve automount tests

stanislavlevin:
* added automount direct and indirect maps
* added automount home directory for IPA users
* added mount options check
* added nfs idmapping check


stanislavlevin:
I've run these tests on PRCI (vagrant based) and Azure (Docker based). The results are stable.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -1989,3 +1993,30 @@ def get_logsize(host, logfile):
     """ get current logsize"""
     logsize = len(host.get_file_contents(logfile))
     return logsize
+
+
+def run_ssh_command_as_user(host, username, password, cmd, raiseonerr=True):
+    """
```
flo-renaud: Another method to get a TGT is to call:
```
cmd = ['su', '-l', username, '-c', cmd]
tasks.run_command_as_user(host, username, cmd, stdin_text=password + '\n')
```
as this will run `su -l username -c 'su -l username -c cmd'`

stanislavlevin: Of course, there are several ways to obtain a TGT.
But I think this could be useful since it exercises a different PAM.

abbra: @stanislavlevin well, in this particular case I don't see why we should be using ssh connection to the `nfs_autofs` host, there is simply no need to go that way.

If you want to add a test that tests SSH login with password, please create a new test. However, for NFS test I don't think it is warrant to add a new way of accessing the host.

Note that actual access to the host from test controller might be done differently and you simply might not be able to reach it via ssh in some environments. I'd prefer to keep this abstraction to a paramiko/test environment setup.

stanislavlevin: @abbra, I see how it is made in `ipatests/test_integration/test_smb.py` and I can't say that it fits my needs.

```
    @classmethod                                                                
    def install(cls, mh):                                                       
        if cls.domain_level is not None:                                        
            domain_level = cls.domain_level                                     
        else:                                                                   
            domain_level = cls.master.config.domain_level                       
        tasks.install_topo(cls.topology,                                        
                           cls.master, cls.replicas,                            
                           cls.clients, domain_level,                           
                           clients_extra_args=('--mkhomedir',))
```
Why:
1) The proposed test checks the auto mounting user home directory located on NFS share. The mixing of the ways to provide a home directory for a user is not that I want to check here. In other words, I don't want something else touches the home dir, even if the autofs hides an original one.
2) The proposed test checks a real user's interaction with his environment and in particular,  his home dir. In my opinion, this is a cleaner way rather than `su ... su ...`.

So, `--mkhomedir` and `su` are just another one workaround.

Note: without local `mkhomedir` `su` fails (at least on ALT):
```
INFO     ipatests.pytest_ipa.integration.host.Host.client3.ParamikoTransport:transport.py:318 RUN ['su', '-l', 'athena', '-c', 'su -l athena -c klist']
DEBUG    ipatests.pytest_ipa.integration.host.Host.client3.cmd65:transport.py:513 RUN ['su', '-l', 'athena', '-c', 'su -l athena -c klist']
DEBUG    ipatests.pytest_ipa.integration.host.Host.client3.cmd65:transport.py:558 /home/athena not available; exiting
DEBUG    ipatests.pytest_ipa.integration.host.Host.client3.cmd65:transport.py:217 Exit code: 1
ERROR    ipatests.pytest_ipa.integration.host.Host.client3.cmd65:host.py:74 stderr: /home/athena not available; exiting
```


> Note that actual access to the host from test controller might be done differently and you simply might not be able to reach it via ssh in some environments. I'd prefer to keep this abstraction to a paramiko/test environment setup.

Really? not SSH? Could I see such examples?
What is the kind of restrictions?
I change user `root -> test_user` and authentication type (pub key -> interactive password).

All of this is not something I'm going to dispute, just my view.

abbra: For `test_smb` tests, no automount test is being done, so autocreating home directory is OK there. Samba needs the home directory to exist when connecting as the user and autocreation cannot happen on connection to Samba itself, it needs to be created in advance.

As for 'no SSH', we should, in general, not assume particular details about the transport used in test to access test hosts. Today it is native ssh or Paramiko implementation, tomorrow it might be another method. Think of Kubernetes, for example -- if these tests would at some point be running in k8s environment, getting access to the specific instance would be using a different method than ssh. I don't want us to re-do all the tests due to this change.

stanislavlevin: I replaced my hack for ssh with `su`.

stanislavlevin:
@abbra, @flo-renaud.
Just after-the-holidays remind.

fcami:
hi @stanislavlevin apologies for the delay, I'll review this today.

stanislavlevin:
@fcami, hi. No problem. Thank you in advance!

ipatests/test_integration/test_nfs.py
```diff
@@ -19,249 +16,473 @@
...
+        'export_options': '*(sec=krb5p,no_subtree_check,rw)',
+        'export_rpath': os.path.join('krbnfs_home', IPA_USER),
+        'exposed_path': os.path.join(os.sep, 'krbnfs_home', IPA_USER),
+        'mount_path': os.path.join('/home', IPA_USER),
+    },
+}
+
+
+@pytest.fixture(scope="class")
+def krb5_nfs_client(request, mh):
```
fcami: TestNFS::test_automount used to test ipa-client-automount on unmodified systems.
This is important as testing ipa-client-automount on unmodified systems caught quite a few bugs.
This is not done anymore as krb5_nfs_client is called before ipa-client-automount.

ipatests/test_integration/test_nfs.py
```diff
@@ -19,249 +16,473 @@
...
+            paths.REQUEST_KEY_CONF
+        ])
+        client.run_command(["cat", paths.REQUEST_KEY_CONF])
+
+        # verbosity for rpc-gssd
+        client.run_command(
+            ["sed", "-i.sav",
+             "-e", r"s/^\(# \)\?verbosity=0.*$/verbosity=4/g",
+             "-e", r"s/^\(# \)\?rpc-verbosity=0.*$/rpc-verbosity=4/g",
+             # we don't want gssproxy for NFS clients (default is '=0')
```
fcami: Could you please point out what the motivation for this is?

fcami:
Hi @stanislavlevin please see the review comments above. Would it be possible to split the "Improve automount tests" in multiple commits so that they can be independently reviewed? The change is so large that I am afraid we might miss something and lose coverage. Thank you.

PR #4014, opened 2019-12-11
[Py3] Fix get_trusted_domain_object_from_sid()

tiran:
DomainValidator.get_trusted_domain_object_from_sid() was using
escape_filter_chars() with bytes. The function only works with text.
This caused idview to fail under some circumstances. Reimplement
backslash hex quoting for bytes.

Fixes: https://pagure.io/freeipa/issue/7958

### DIFF
--- a/ipaserver/dcerpc.py
+++ b/ipaserver/dcerpc.py
@@ -63,7 +63,6 @@ import pysss
 import six
 from ipaplatform.paths import paths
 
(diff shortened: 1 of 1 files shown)


abbra:
LGTM

tiran:
master:

* c30a0c22681ec4a5d6cdf341a69d144f6ab3a0a8 Fix get_trusted_domain_object_from_sid()

PR #4015, opened 2019-12-11
ipa-kdb: Remove keys if password auth is disabled

sumit-bose:
With commit 15ff9c8 a check was removed and as a result Kerberos keys
are unconditionally added to the user entry struct if they are
available. As a result the password related pre-authentication methods
PA-ENC-TIMESTAMP and PA-ETYPE-INFO2 are advertised in the NEEDED_PREAUTH
reply to an AS_REQ.

With respect to the KDC policies this does not matter much because if
password authentication is disabled for the given principal the policy
will reject the AS_REQ if the user tries password authentication. This
is possible because with commit 15ff9c8 kinit will ask for a password if
called without any additional options (e.g. armor ticket or PKINIT
identity). Before 15ff9c8 was committed it just failed with 'kinit:
Pre-authentication failed: Invalid argument while getting initial
credentials' because no suitable pre-authentication method was
available. This is the same behavior as if no password was set for the
given principal.

But with this change SSSD fails to detect the available authentication
types for the given principal properly. As described in
https://docs.pagure.org/SSSD.sssd/design_pages/prompting_for_multiple_authentication_types.html
SSSD uses the MIT Kerberos responder interface to determine the
available authentication methods for the principal and does not check
the ipaUserAuthType LDAP attribute. As a result if a user has 2FA (otp)
authentication configured, which implies that a password is set as the
first factor, the responder interface will always indicate that password
authentication is available even if only opt is enabled for the user.

In this case SSSD will use a prompting which indicates that the second
factor might be optional. Additionally if prompting the user directly is
not possible (e.g. ssh with ChallengeResponseAuthentication /
KbdInteractiveAuthentication disabled) the single string entered by the
user will always be assumed as a password and not as a combination of
password and otp-token value. As a consequence authentication will
always fail because password authentication is disabled for the user and
since SSSD does not do try-and-error 2FA is not tried.

This patch add back the check so that if password authentication is not
available for the principal the Kerberos will not be added to the entry
struct and the KDC will not advertise PA-ENC-TIMESTAMP or
PA-ETYPE-INFO2. If you think this is wrong and the behavior added by
15ff9c8 should be preferred SSSD handing of the available authentication
types must be extended to read ipaUserAuthType as well to restore the
user experience with respect to 2FA prompting and ssh behavior.

Related to https://pagure.io/freeipa/issue/8001

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb_principals.c
+++ b/daemons/ipa-kdb/ipa_kdb_principals.c
@@ -774,6 +774,22 @@ static krb5_error_code ipadb_parse_ldap_entry(krb5_context kcontext,
                                       &res_key_data, &result, &mkvno);
     switch (ret) {
     case 0:
+        /* Only set a principal's key if password auth can be used. Otherwise
+         * the KDC would add pre-authentication methods to the NEEDED_PREAUTH
+         * reply for AS-REQs which indicate the password authentication is
+         * available. This might confuse applications like e.g. SSSD which try
+         * to determine suitable authentication methods and corresponding
+         * prompts with the help of MIT Kerberos' responder interface which
+         * acts on the returned pre-authentication methods. A typical example
+         * is enforced OTP authentication where of course keys are available
+         * for the first factor but password authentication should not be
+         * advertised by the KDC. */
+        if (!(ua & IPADB_USER_AUTH_PASSWORD) && (ua != IPADB_USER_AUTH_NONE)) {
+            /* This is the same behavior as ENOENT below. */
+            ipa_krb5_free_key_data(res_key_data, result);
+            break;
+        }
+
         entry->key_data = res_key_data;
         entry->n_key_data = result;
         if (mkvno) {


daemons/ipa-kdb/ipa_kdb_principals.c
```diff
@@ -774,6 +774,15 @@ static krb5_error_code ipadb_parse_ldap_entry(krb5_context kcontext,
                                       &res_key_data, &result, &mkvno);
     switch (ret) {
     case 0:
+        /* Only set a principal's key if password auth should be used to make
+         * sure that the KDC relies do not advertise password auth if not
+         * available. */
```
frozencemetery: Thank you for expanding this comment from its previous version.  Could you perhaps mention more of what your commit message contained about the responder interface and OTP?  It's still not clear from the comment what will break if this conditional were to be removed again (though of course it is from your commit message).

abbra:
@sumit-bose thank you for the patch. I have no additional comments on top of @frozencemetery request.

sumit-bose:
Hi,

I extended the comment in the latest version.

bye,
Sumit

abbra:
master:

* f0d12b7f1b06af8ad369041b64dc763441a1a44a ipa-kdb: Remove keys if password auth is disabled

PR #4034, opened 2019-12-13
Adding auto COPR builds

SilleBille:
This patch adds a script which can be used to auto-build freeipa in COPR.

## Usage
Currently, PKI runs certain cert related tests from `test_xmlrpc`. It pulls `freeipa` packages from stable fedora repo. This patch can help catch bugs as quickly as possible (for every commit made to either repositories)

## Personal Demo
I setup a [personal COPR repo](https://copr.fedorainfracloud.org/coprs/dmoluguw/freeipa-master/package/freeipa/) to demonstrate that this patch works

## Next steps
Once this patch gets merged, following action items need to be performed by the admin:

1. Create a new package `freeipa` inside the [official COPR repo](https://copr.fedorainfracloud.org/coprs/g/freeipa/freeipa-master/packages/)
2. Setup auto build with `srpm` option
3. Setup github hook to trigger the COPR builds

Thank you, @abbra for accepting the idea.

`Signed-off-by: Dinesh Prasanth M K <dmoluguw@redhat.com>`

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -1,0 +1,17 @@
+srpm:
+	# Setup development environment
+	echo "Installing base development environment"
+	dnf install -y dnf-plugins-core
+	echo "Installing FreeIPA development dependencies"
+	dnf builddep -y --skip-broken --spec freeipa.spec.in --best --allowerasing --setopt=install_weak_deps=False
+
+	# Run autoconf
+	autoreconf -i
+	./configure --enable-silent-rules
+
+	# Generate SRPMS
+	make srpms
+
+	if [[ "${outdir}" != "" ]]; then \
+		mv rpmbuild/SRPMS/* ${outdir}; \
+	fi


abbra:
@SilleBille thanks for the proposal! I have few comments. Please use the procedure we have in `ipatests/azure/templates/prepare-build.yml` when preparing the build. It is better to have it unified.

.copr/Makefile
```diff
@@ -0,0 +1,17 @@
+srpm:
+	# Setup development environment
+	echo "Installing base development environment"
+	dnf install -y gdb make autoconf rpm-build gettext-devel automake libtool docker python3-paramiko python3-pyyaml
```
tiran: Why does COPR auto-build need paramiko, docker, and pyyaml?

.copr/Makefile
```diff
@@ -0,0 +1,17 @@
...
+	echo "Installing FreeIPA development dependencies"
+	dnf builddep -y --skip-broken -D "with_wheels 1" -D "with_lint 1" --spec freeipa.spec.in --best --allowerasing --setopt=install_weak_deps=False
```
tiran: You don't need ``with_wheels`` or ``with_lint`` for RPM builds. The dependencies are only required for development and testing.

.copr/Makefile
tiran: You also need ``dnf-plugins-core`` for ``dnf builddep`` command.

SilleBille: As per @abbra 's comment, I tried to reuse the same template as `ipatests/azure/templates/prepare-build.yml` to keep things uniform.

.copr/Makefile
SilleBille: Done

.copr/Makefile
SilleBille: Removed unnecessary packages in my latest commit

abbra:
I think we need this also in ipa-4-8 to setup 'latest ipa-4-8 copr' automation

abbra:
master:

* df59f09eed698b76a9089614a170c771c9dfb7b6 Adding auto COPR builds

PR #4044, opened 2019-12-16
DNS install check: Fix overlapping DNS zone from the master itself

t-woerner:
The change to allow overlapping zone to be from the master itself has
introduced two issues: The check for the master itself should only executed
if options.force and options.allow_zone_overlap are both false and the
reverse zone check later on was still handling ValueError instead of
dnsutil.DNSZoneAlreadyExists.

Both issues have been fixed and the deployment with existing name servers
is properly working again.

Fixes: https://pagure.io/freeipa/issue/8150

### DIFF
--- a/ipaserver/install/dns.py
+++ b/ipaserver/install/dns.py
@@ -135,15 +135,15 @@ def install_check(standalone, api, replica, options, hostname):
                 logger.warning("%s Please make sure that the domain is "
                                "properly delegated to this IPA server.",
                                e)
-
-            hst = dnsutil.DNSName(hostname).make_absolute().to_text()
-            if hst not in e.kwargs['ns']:
-                raise ValueError(str(e))
+            else:
+                hst = dnsutil.DNSName(hostname).make_absolute().to_text()
+                if hst not in e.kwargs['ns']:
+                    raise ValueError(str(e))
 
     for reverse_zone in options.reverse_zones:
         try:
             dnsutil.check_zone_overlap(reverse_zone)
-        except ValueError as e:
+        except dnsutil.DNSZoneAlreadyExists as e:
             if options.force or options.allow_zone_overlap:
                 logger.warning('%s', str(e))
             else:


pvoborni:
The patch looks good to me. 

Though I'd expect in the previous commit that the DNS special case would have a small comment as it will not be obvious for the future code readers why it is there.

flo-renaud:
@t-woerner 
thanks for the PR, looks good to me. I tested the following:

- [x] calling `ipa-server-install --allow-zone-overlap` on an already installed master: properly get the message that IPA server is already installed
- [x] calling `ipa-server-install ` without `--allow-zone-overlap` on an already installed master: properly get the message that IPA server is already installed
- [x] calling `ipa-server-install --allow-zone-overlap` with an overlap in the zone or reverse zone: install goes on and displays the warning about DNS zone already managed
- [x] calling `ipa-server-install` without `--allow-zone-overlap` with an overlap in the zone or reverse zone: install does not proceed, error message about the DNS zone already managed
- [x] calling `ipa dnszone-add --skip-overlap-check` with an overlap in the zone: the command succeeds
- [x] calling `ipa dnszone-add` without `--skip-overlap-check` with an overlap in the zone: the command does not proceed, error message about the zone already managed

Waiting for the last test result before I provide the ACK.

f-trivino:
@flo-renaud does this issue ring your bell?

test_commands.py--TestIPACommand-test_user_mod_change_capitalization_issue5879/master.ipa.test is stuck with:

SELINUX_getpeercon failed [92][Protocol not available].

2019-12-16 15:17:04 [profileChangeMonitor] WARNING: Profile change monitor: failed to get LDAPConnection. Retrying in 1 second.

f-trivino:
> @flo-renaud does this issue ring your bell?
> 
> test_commands.py--TestIPACommand-test_user_mod_change_capitalization_issue5879/master.ipa.test is stuck with:
> 
> SELINUX_getpeercon failed [92][Protocol not available].
> 
> 2019-12-16 15:17:04 [profileChangeMonitor] WARNING: Profile change monitor: failed to get LDAPConnection. Retrying in 1 second.
> 
> I see more than 1500 retries:
> 
> grep "LDAPConnection. Retrying in 1 second" * -R | wc -l
> 1505

running again in big-runner-3

flo-renaud:
Hi @f-trivino 

> does this issue ring your bell?
> 
> test_commands.py--TestIPACommand-test_user_mod_change_capitalization_issue5879/master.ipa.test is stuck with:
> 
> SELINUX_getpeercon failed [92][Protocol not available].
> 

No, never saw this problem.
Since the tests are green now, ACK

flo-renaud:
This PR depends on PR #4001 which is currently backported to ipa-4-8 only, hence removing labels for ipa-4-6 and ipa-4-8

t-woerner:
master:

* f80a6548ad8d0d21219ce26d3e819f7ad5f3663e DNS install check: Fix overlapping DNS zone from the master itself

PR #4057, opened 2019-12-18
Reset per-indicator Kerberos policy

abbra:
When 'ipa krbtpolicy-reset' is called, we need to reset all policy
settings, including per-indicator ones. Per-indicator policy uses
subtyped attributes (foo;bar), the current krbtpolicy-reset code does
not deal with those.

Add support for per-indicator policy reset. It is a bit tricky, as we
need to drop the values to defaults but avoid adding non-per-indicator
variants of the same attributes.

Add test to check that policy has been resetted by observing a new
Kerberos TGT for the user after its policy reset.

Fixes: https://pagure.io/freeipa/issue/8153

### DIFF
--- a/ipaserver/plugins/krbtpolicy.py
+++ b/ipaserver/plugins/krbtpolicy.py
@@ -68,6 +68,8 @@ register = Registry()
 _default_values = {
     'krbmaxticketlife': 86400,
     'krbmaxrenewableage': 604800,
+    'krbauthindmaxticketlife': 86400,
+    'krbauthindmaxrenewableage': 604800,
 }
 
 # These attributes never have non-optional values, so they should be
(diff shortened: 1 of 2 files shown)


abbra:
Test succeeded: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/b83862d4-2189-11ea-8d2a-fa163e1c6062/report.html

ipaserver/plugins/krbtpolicy.py
```diff
@@ -311,9 +314,26 @@ def execute(self, uid=None, **options):
                 def_values[a] = None
         # if reseting global policy - set values to default
         else:
-            def_values = _default_values
+            def_values = deepcopy(_default_values)
```
tiran: The default values dict contains just strings and ints. ``_default_values.copy()`` does the trick, too.

abbra: updated, thanks.

tiran:
master:

* 2ed5eca762e136a1db9d3c20ec068d562b215699 Reset per-indicator Kerberos policy

PR #3924, opened 2019-11-21
ipatests: add test for sssd behavior with disabled trustdomains

wladich:
When a trusted subdomain is disabled in ipa, users from this domain should not be able to access resources ipa resources.
    
Related to: https://pagure.io/SSSD/sssd/issue/4078

This PR also adds 
 * utility function  for  getting sssd version on remore host
 * a context manager for declaring part of test as xfail

PR #4078, opened 2019-12-24
ipatests: User and group with same name should not break reading AD user data.

amore17:
Regression test resolving trusted users and groups should be
successful when there is a user in IPA with the
same name as a group name.
Related : https://pagure.io/SSSD/sssd/issue/4073

### DIFF
--- a/ipatests/test_integration/test_sssd.py
+++ b/ipatests/test_integration/test_sssd.py
@@ -286,3 +286,39 @@ class TestSSSDWithAdTrust(IntegrationTest):
         finally:
             sssd_conf_backup.restore()
             tasks.clear_sssd_cache(self.master)
+
+    def test_ext_grp_with_ldap(self):
+        """User and group with same name should not break reading AD user data.
+
+        Regression test for https://pagure.io/SSSD/sssd/issue/4073
+
+        When aduser is added in extrnal group and this group is added
+        in group with same name of nonprivate ipa user and possix id, then
+        lookup of aduser and group should be successful when cache is empty.
+        """
+        client = self.clients[0]
+        user = 'ipatest'
+        userid = '100996'
+        ext_group = 'ext-ipatest'
+        tasks.kinit_admin(self.master)
+        # add user with same uid and gidnumber
+        tasks.user_add(self.master, user, extra_args=[
+            '--noprivate', '--uid', userid, '--gidnumber', userid])
+        # add group with same as user_name and user_id.
+        tasks.group_add(self.master, user, extra_args=['--gid', userid])
+        tasks.group_add(self.master, ext_group, extra_args=['--external'])
+        self.master.run_command(
+            ['ipa', 'group-add-member', '--group', ext_group, user])
+        self.master.run_command([
+            'ipa', 'group-add-member', '--external',
+            self.users['ad']['name'], ext_group,
+            '--users=', '--groups='])
+        tasks.clear_sssd_cache(self.master)
+        tasks.clear_sssd_cache(client)
+        try:
+            result = client.run_command(['id', self.users['ad']['name']])
+            assert '{uid}({name})'.format(uid=userid,
+                                          name=user) in result.stdout_text
+        finally:
+            self.master.run_command(['ipa', 'user-del', user])
+            self.master.run_command(['ipa', 'group-del', user, ext_group])
--- a/ipatests/test_integration/test_sssd.py
+++ b/ipatests/test_integration/test_sssd.py
@@ -15,6 +15,7 @@ import textwrap
 
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
+from ipaplatform.tasks import tasks as platform_tasks
 from ipaplatform.osinfo import osinfo
 from ipaplatform.paths import paths
 from ipapython.dn import DN
@@ -296,6 +297,12 @@ class TestSSSDWithAdTrust(IntegrationTest):
         in group with same name of nonprivate ipa user and possix id, then
         lookup of aduser and group should be successful when cache is empty.
         """
+        cmd = self.master.run_command(['sssd', '--version'])
+        sssd_version = platform_tasks.parse_ipa_version(
+            cmd.stdout_text.strip())
+        if sssd_version <= platform_tasks.parse_ipa_version('2.2.2'):
+            pytest.skip("Fix for https://pagure.io/SSSD/sssd/issue/4073 "
+                        "unavailable with sssd-2.2.2")
         client = self.clients[0]
         user = 'ipatest'
         userid = '100996'


wladich:
@amore17 
As this test is related to changes solely in sssd, is not it better to add it to test_sssd.py suite?

amore17:
> @amore17
> As this test is related to changes solely in sssd, is not it better to add it to test_sssd.py suite?

Sure Thanks for review
i'll add it to test_sssd.py

ipatests/test_integration/test_sssd.py
```diff
@@ -148,3 +149,38 @@ def test_is_user_filtered(self, user):
...
+        tasks.group_add(self.master, ext_group, extra_args=['--external'])
+        self.master.run_command(
+            ['ipa', 'group-add-member', '--group', ext_group, user])
+        self.master.run_command([
+            'ipa', 'group-add-member', '--external',
+            self.users['ad']['name'], ext_group,
+            '--users=', '--groups='])
+        self.master.run_command(['sss_cache', '-E'])
+        tasks.clear_sssd_cache(self.master)
+        client.run_command(['sss_cache', '-E'])
```
wladich: There is no reason to invalidate cache as you completely remove it in the following line

ipatests/test_integration/test_sssd.py
```diff
@@ -148,3 +149,38 @@ def test_is_user_filtered(self, user):
...
+        tasks.clear_sssd_cache(client)
+        try:
+            result = client.run_command(['id', self.users['ad']['name']])
+            assert id and user in result.stdout_text
```
wladich: This is equivalent to:
```
assert boolean(id) and (user in result.stdout_text)
```

ipatests/test_integration/test_sssd.py
```diff
@@ -148,3 +149,38 @@ def test_is_user_filtered(self, user):
...
+        finally:
+            self.master.run_command(['ipa', 'user-del', user])
+            self.master.run_command(['ipa', 'group-del', user, ext_group])
```
wladich: The test creates two groups but cleanup removes only one

amore17: It adds two groups : user, ext_group
        # add group with same name as user and id.
        tasks.group_add(self.master, user, extra_args=['--gid', id])
        tasks.group_add(self.master, ext_group, extra_args=['--external'])

wladich: ah, sorry, did not notice that you delete two groups in one command

wladich:
The PR needs to be rebased.

Another thing: the commit message and the doc string read "IPA objects should read via LDAP when resolving trusted users".
The proposed test does not check the mechanism of the requests. It does check that having user and group with same name does not break reading AD user data. 

The original BZ described two problems:
1. Requests are sent via extdom plugin (which can cause excessive load on extdom plugin)
2. AD user resolution does not work when we have user and group with same name.

The tests checks only item 2 while stating that it checks item 1. 
This is a regression tests which is expected to fail once the stated regression appears again.
The problems were interconnected and were fixed together, but we can not guarantee that regression 1 will be accompanied with 2.

ipatests/test_integration/test_sssd.py
```diff
@@ -233,3 +234,44 @@ def test_external_group_paging(self):
...
+        Regression test for https://pagure.io/SSSD/sssd/issue/4073
+
+        When aduser is added in extrnal group and this is group is added
+        in group with same name of nonprivate ipa user and possix id, then
+        lookup of aduser and group should be successful when cache is empty.
+        """
+        cmd = self.master.run_command(['sssd', '--version'])
+        sssd_version = platform_tasks.parse_ipa_version(
+            cmd.stdout_text.strip())
+        if sssd_version.version <= '2.2.2':
```
wladich: You are using parse_ipa_version in awrong way: `sssd_version.version` is a string. So you simply compare string with a string. 
And for strings following is true: '2.2.10' < '2.2.2'

ipatests/test_integration/test_sssd.py
```diff
@@ -233,3 +234,44 @@ def test_external_group_paging(self):
...
+            pytest.skip("Fix for sssd issue:4073 unavailable in 2.2.2")
+        client = self.clients[0]
+        user = 'ipatest'
+        id = '100996'
```
wladich: Please try to avoid shadowing builtin objects (`id` is a builtin finction). It makes code confusing.

ipatests/test_integration/test_sssd.py
```diff
@@ -233,3 +234,44 @@ def test_external_group_paging(self):
...
+            'ipa', 'group-add-member', '--external',
+            self.users['ad']['name'], ext_group,
+            '--users=', '--groups='])
+        self.master.run_command(['getent', 'group', user])
+        tasks.clear_sssd_cache(self.master)
+        tasks.clear_sssd_cache(client)
+        try:
+            result = client.run_command(['id', self.users['ad']['name']])
+            assert id in result.stdout_text
+            assert user in result.stdout_text
```
wladich: It is better to use more specific check for the group name and id to be sure that false positive is not produced by presence of similar string elsewhere in the output.
You can use regexp like this one (needs checking):
```
exp_re = r'groups=[^=]*\b{id}\({group}\)'.format(id=id, group=group)
```

flo-renaud:
> We need a green run in CI for the new test to verify this PR. Now the test is skipped as sssd 2.2.3 is not available in any Fedora version.
> @flo-renaud Can you suggest how to deal with it?
> 

This is tricky, as in the 2 sssd branches (master and 1-16) the fix is available but not released yet in fedora. We could add the sssd package to our copr repo @freeipa/freeipa-pr-ci but this should be discussed globally as we encounter this situation more and more often (writing tests for sssd, with fixes not yet available).

ipatests/test_integration/test_sssd.py
```diff
@@ -233,3 +233,41 @@ def test_external_group_paging(self):
...
+        self.master.run_command(
+            ['ipa', 'group-add-member', '--group', ext_group, user])
+        self.master.run_command([
+            'ipa', 'group-add-member', '--external',
+            self.users['ad']['name'], ext_group,
+            '--users=', '--groups='])
+        tasks.clear_sssd_cache(self.master)
+        tasks.clear_sssd_cache(client)
+        self.master.run_command(['id', user])
+        self.master.run_command(['getent', 'group', user])
```
wladich: Please provide information why those lines were added?
When I was reviewing last time they were not here: https://github.com/freeipa/freeipa/commit/3286044712728d7393bc8d5440cb662bcf0d9a73
Also this step is not mentioned in original reproducer in bugzilla.

amore17: Thanks for review I'll remove that.
These checks were added for testing purpose as without latest sssd package test was failing here.

amore17:
> As the sssd-2.2.2 is only available in Fedora 31 but not Fedora 30, this test will fail in nightly-previous test run. You need to add a conditional xfail based on sssd version. Please see the test results for Fedora 30: [wladich#14](https://github.com/wladich/freeipa/pull/14)

Yes adding xfail was pending as i wanted to test with latest package for green run.
I'll add xfail for sssd-2.2.2

amore17:
Test results using sssd-ipa-2.2.3-13.fc31.x86_64 : http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/e282872c-6464-11ea-a9bf-fa163e53827b/

ipatests/test_integration/test_sssd.py
```diff
@@ -286,3 +287,45 @@ def get_cache_update_time(obj_kind, obj_name):
...
+
+        When aduser is added in extrnal group and this group is added
+        in group with same name of nonprivate ipa user and possix id, then
+        lookup of aduser and group should be successful when cache is empty.
+        """
+        cmd = self.master.run_command(['sssd', '--version'])
+        sssd_version = platform_tasks.parse_ipa_version(
+            cmd.stdout_text.strip())
+        if sssd_version <= platform_tasks.parse_ipa_version('2.2.2'):
+            pytest.xfail("Fix for https://pagure.io/SSSD/sssd/issue/4073 "
```
wladich: Let me explain the difference between skip and xfail:

>  A **skip** means that you expect your test to pass only if some conditions are met, otherwise pytest should skip running the test altogether. [...]
A **xfail** means that you expect a test to fail for some reason. A common example is a test for a feature not yet implemented, or a bug not yet fixed. When a test passes despite being expected to fail (marked with pytest.mark.xfail), it’s an xpass and will be reported in the test summary.
-- http://doc.pytest.org/en/latest/skipping.html

We use xfails to to double check ourselves that we properly understand the conditions that must be met for test to pass. Here we could use it to verify that we have properly identified minimum sssd version which has the bug fixed.

But the way it is written now, the test has no chance to succeed and to be eventually reported as "XPASS".

I suggest to do one of the following:
* Replace xfail with skip.
* properly use xfail inverting success/fail of the assertion.
* wait until #3924 is merged and use the xfail context manager introduced in that PR

amore17: > But the way it is written now, the test has no chance to succeed with older version of sssd and to be eventually reported as "XPASS".
with older version i am getting xfail
http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/49484806-64fc-11ea-a229-fa163e97fbb7/report.html
which is expected to fail there with 2.2.2

wladich: "xfail" means we are making additional check for correctness of condition. Imagine, the fix is somehow added to the 2.2.2. Or maybe the assumption that 2.2.3 is needed was wrong and 2.2.2 was working from the beginning. With the current implementation we will always have "expected failure" with 2.2.2 even if it  does not contain the bug (or the bug is fixed in the future). The proper implementation should produce "unexpected success" in that case so that we can review the test and  adapt the condition. That current behavior is exactly what "skip" does so calling it "xfails" is misleading.

> with older version i am getting xfail

The problem is that you will always get xfail with older version, even if this older version is fixed, which is not the expected behavior of "xfail". If you do not care about it, please use "skip" to avoid misleading. Otherwise, xfail should invert the logic of the assertion.

wladich:
LGTM. 
Please remove the temp commit and I will ack the PR.

amore17:
Results 
 fedora-latest/test_sssd_ipa_2_2_2 : http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/63ce488c-675a-11ea-a8cf-fa163e401a00
 fedora-latest/test_sssd_ipa_2_2_3 : http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/5a763e2a-675a-11ea-bb8d-fa163e0a82d3
 fedora-previous/test_sssd : http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/978309ce-675a-11ea-81c1-fa163e19313f

abbra:
master:

* b2ab2863ca121d94e7519cd53a9fdbb43b6c5d27 ipatests: User and group with same name should not break reading AD user data.
* 6018ccaa8dfeccf27494369a8ba9a6c670779e2e Mark test to skip sssd-2.2.2

PR #4089, opened 2020-01-02
ipatests: Add xmlrpc test for valid/invalid inputs for kerberos ticket policy.

amore17:
This checks that valid/invalid inputs for subtypes of 
authentication indicator kerberos ticket policy options.
Related: https://pagure.io/freeipa/issue/8001

### DIFF
--- a/ipatests/test_xmlrpc/test_krbtpolicy.py
+++ b/ipatests/test_xmlrpc/test_krbtpolicy.py
@@ -28,6 +28,33 @@ import pytest
 
 user1 = u'tuser1'
 
+invalid_values = [(u'abc123', 'must be an integer'),
+                  (u'2147483648', 'can be at most 2147483647'),
+                  (u'0', 'must be at least 1'),
+                  (u'-1', 'must be at least 1')]
+parameters = [('krbauthindmaxrenewableage_radius', 'radius_maxrenew'),
+              ('krbauthindmaxticketlife_radius', 'radius_maxlife'),
+              ('krbauthindmaxrenewableage_pkinit', 'pkinit_maxrenew'),
+              ('krbauthindmaxticketlife_pkinit', 'pkinit_maxlife'),
+              ('krbauthindmaxrenewableage_otp', 'otp_maxrenew'),
+              ('krbauthindmaxticketlife_otp', 'otp_maxlife'),
+              ('krbauthindmaxrenewableage_hardened', 'hardened_maxrenew'),
+              ('krbauthindmaxticketlife_hardened', 'hardened_maxlife'),
+              ]
+
+
+def create_dict(desc, param, param_name, value, error):
+    cmd_args = dict()
+    cmd_args[param] = value
+    if value != u'abc123':
+        return dict(desc=desc, command=('krbtpolicy_mod', [user1], cmd_args),
+                    expected=errors.ValidationError(name=param_name,
+                                                    error=error))
+    else:
+        return dict(desc=desc, command=('krbtpolicy_mod', [user1], cmd_args),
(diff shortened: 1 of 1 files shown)


ipatests/test_xmlrpc/test_krbtpolicy.py
```diff
@@ -116,6 +116,35 @@ class test_krbtpolicy(Declarative):
...
+            ),
+            expected=dict(
+                value=user1,
+                summary=None,
+                result=dict(
+                    krbmaxticketlife=[u'86400'],
+                    krbmaxrenewableage=[u'604800'],
+                ),
+            ),
+        ),
```
flo-renaud: I would remove the lines 119 to 147 (Reset global policy and Reset user policy) in order to keep the same strategy throughout the test: define new attr that is incrementally added to the set of policies. Also keep the line krbmaxticketlife=[u'3600'] in the next expected result, after line 156 and in all the subsequent expected results.
You may also use a slightly different value for each maxticketlife to ensure they are not mixed (for instance 3600 for maxticketlife, 3700 for otpmaxticketlife etc...)

ipatests/test_xmlrpc/test_krbtpolicy.py
```diff
@@ -125,12 +154,93 @@ class test_krbtpolicy(Declarative):
...
+
+        dict(
+            desc='Try updating invalid value auth indicator hardened',
+            command=('krbtpolicy_mod', [user1],
+                     dict(krbauthindmaxticketlife_hardened=u'abc123')
+                     ),
+            expected=errors.ConversionError(
+                name='hardened_maxlife', error='must be an integer'),
+        ),
+
```
flo-renaud: Each validation test (min val, max val, invalid format) can be done for all the maxticketlife (otp/radius/pkinit/hardened)

ipatests/test_xmlrpc/test_krbtpolicy.py
```diff
@@ -116,21 +116,407 @@ class test_krbtpolicy(Declarative):
...
+                name='radius_maxrenew', error='can be at most 2147483647'),
+        ),
+        dict(
+            desc='Try updating invalid value auth indicator radius',
+            command=('krbtpolicy_mod', [user1],
+                     dict(krbauthindmaxrenewableage_radius=u'abc123')
+                     ),
+            expected=errors.ConversionError(
+                name='radius_maxrenew', error='must be an integer'),
+        ),
```
flo-renaud: The above code (tests for invalid args) can be simplified with a double loop similar to:
```
invalid_values = [ (u'abc123', 'must be an integer'), (u'2147483648', 'can be at most 2147483647'), (u'0', 'must be at least 1'), (u'-1', 'must be at least 1')]
parameters = [ ('krbauthindmaxrenewableage_radius', 'radius_maxrenew'), ('krbauthindmaxticketlife_radius', 'radius_maxlife'), ...]
for (value,error) in invalid_values:
  for (param, param_name) in parameters:
    tests.append(create_dict(desc='Try updating invalid %s'.format(param_name),param=param, value=value,name=param_name, error=error)
```
And you need to create a function
```
def create_dict(desc, param, param_name, value, error):
    cmd_args=dict()
    cmd_args[param]=value
    return dict(desc=desc, command=('krbtpolicy_mod', [user1],cmd_args),expected=errors.ConversionError(name=name, error=error))
```

flo-renaud:
Hi @amore17 
Thanks for the update, LGTM.

flo-renaud:
master:

* f35738ef22aac54c09d64e3a7985298faf9ac1eb Add xmlrpc test with input validation check for kerberos ticket policy.

PR #4102, opened 2020-01-06
Debian: write out only one CA certificate per file

yrro:
ca-certificates populates /etc/ssl/certs with symlinks to its input
files and then runs 'openssl rehash' to create the symlinks that libssl
uses to look up a CA certificate to see if it is trused.

'openssl rehash' ignores any files that contain more than one
certificate: <https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=945274>.

With this change, we write out trusted CA certificates to
/usr/local/share/ipa-ca, one certificate per file.

The code currently uses f-strings and pathlib which I presume are
newer than the oldest version of Python that FreeIPA supports. I'll
change the code to avoid using these features if you're happy with
the approach I've taken.

Fixes: https://pagure.io/freeipa/issue/8106

### DIFF
--- a/ipaplatform/base/tasks.py
+++ b/ipaplatform/base/tasks.py
@@ -73,6 +73,23 @@ class BaseTaskNamespace:
         Returns True if the operation succeeded, False otherwise.
         """
 
+        try:
+            if self.platform_insert_ca_certs(ca_certs):
+                return self.reload_systemwide_ca_store()
+        except Exception:
+            logger.exception('Could not populate systemwide CA store')
+
+        return False
+
+    def platform_insert_ca_certs(self, ca_certs):
+        """
+        Platform implementations override this method to implement
+        population of the systemwide CA store.
+
+        Returns True if changes were made to the CA store, False otherwise.
+
+        Raises Exception if something went wrong.
+        """
         raise NotImplementedError()
 
     def remove_ca_certs_from_systemwide_ca_store(self):
@@ -83,6 +100,25 @@ class BaseTaskNamespace:
         Returns True if the operation succeeded, False otherwise.
         """
 
+        try:
+            if self.platform_remove_ca_certs():
+                return self.reload_systemwide_ca_store()
+        except Exception:
+            logger.exception(
+                'Could not remove certificates from systemwide CA store'
+            )
+
+        return False
+
+    def platform_remove_ca_certs(self):
+        """
+        Platform implementations override this method to implement
+        removal of certificates from the systemwide CA store.
+
+        Returns True if changes were made to the CA store, False otherwise.
+
+        Raises Exception if something went wrong.
+        """
         raise NotImplementedError()
 
     def get_svc_list_file(self):
--- a/ipaplatform/debian/tasks.py
+++ b/ipaplatform/debian/tasks.py
@@ -8,12 +8,21 @@ This module contains default Debian-specific implementations of system tasks.
 
 from __future__ import absolute_import
 
+import logging
+import os
+import shutil
+from pathlib import Path
+
 from ipaplatform.base.tasks import BaseTaskNamespace
 from ipaplatform.redhat.tasks import RedHatTaskNamespace
 from ipaplatform.paths import paths
 
 from ipapython import directivesetter
 from ipapython import ipautil
+from ipapython.dn import DN
+
+logger = logging.getLogger(__name__)
+
 
 class DebianTaskNamespace(RedHatTaskNamespace):
     @staticmethod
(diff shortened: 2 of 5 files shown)


rcritten:
There are remaining references to paths.IPA_P11_KIT which is being removed in the patch.

ipaplatform/debian/tasks.py
```diff
@@ -8,13 +8,23 @@
 
 from __future__ import absolute_import
 
+import errno
+import logging
+import os
+from pathlib import Path
```
tiran: IPA doesn't use pathlib yet. Should we start using the stdlib module?

ipaplatform/debian/tasks.py
```diff
@@ -88,4 +98,99 @@ def configure_pkcs11_modules(self, fstore):
...
+        ca_certificates_dir_path = Path(paths.CA_CERTIFICATES_DIR)
+        try:
+            ca_certificates_dir_path.mkdir(mode=0o755, exist_ok=True)
+        except OSError as e:
+            logger.error(
+                'Could not create %s: %s', ca_certificates_dir_path, e)
+            return False
+
+        for i, ca_cert in enumerate(ca_certs):
+            cert, nickname, trusted, _ext_key_usage = *ca_cert,
```
tiran: Simple tuple unpacking works here, too:
``cert, nickname, trusted, _ext_key_usage = ca_cert``

ipaplatform/debian/tasks.py
```diff
@@ -88,4 +98,99 @@ def configure_pkcs11_modules(self, fstore):
...
+
+            if not trusted:
+                continue
+
+            try:
+                # XXX is rfc4514_string really displaying the DN
+                # elements in the wrong order?
```
tiran: LDAP and X.509 have different opinions how to display a DN. LDAP (RFC 4514) uses leaf to root order, X.509 root to leaf order.

ipaplatform/debian/tasks.py
```diff
@@ -88,4 +98,99 @@ def configure_pkcs11_modules(self, fstore):
...
+                subject = cert.subject.rfc4514_string()
```
tiran: ``rfc4514_string`` is a new API. IPA targets older python-cryptography versions without the API.

ipaplatform/debian/tasks.py
```diff
@@ -88,4 +98,99 @@ def configure_pkcs11_modules(self, fstore):
...
+                    "Failed to decode certificate \"%s\": %s", nickname, e)
+                continue
+
+            ca_path = ca_certificates_dir_path / f'ipa-ca-{i}.crt'
+
+            try:
+                with open(ca_path, 'w') as f:
+                    os.fchmod(f.fileno(), 0o644)
+
+                    f.write(f'''\
```
tiran: style nit-pick: we prefer double triple quotes.

ipaplatform/debian/tasks.py
yrro: If you're happy with Python 3.4 as the minimum?

ipaplatform/debian/tasks.py
yrro: Right... too much staring at the output of gnutls vs openssl x509 vs ldapsearch all day long... :)

ipaplatform/debian/tasks.py
yrro: OK -- i'll use `hasattr` to see if `rfc4514_string` exists and only print the Subject/Issuer if so. Users with old versions of the library can use `openssl x509` by hand :)

ipaplatform/debian/tasks.py
yrro: OK

tjaalton:
thanks for working on this, but could you make it do both, write a file for pki-kit and ca-certificates?

ipaplatform/debian/tasks.py
yrro: Hold on... if I load a certificate with `cryptography` and look at `cert.subject.rdns` then I get:

```
[<RelativeDistinguishedName(O=IPADOMAIN)>,
 <RelativeDistinguishedName(CN=Certificate Authority)>]
```

So I'd expect `rfc_4514_string` to show them in the opposite order--but it doesn't, it's outputting `O=IPADOMAIN,CN=Certificate Authority`!

abbra: My understanding is that `cryptography` doesn't really implement explicit ordering. It simply traverses all attributes in the list of attributes for the name as they are. And that list is constructed by using the same order as it was passed on the name instantiation. See https://cryptography.io/en/latest/_modules/cryptography/x509/name/#Name.rfc4514_string and around.

ipaplatform/debian/tasks.py
yrro: Actually I just realised that cert is an `IPACertificate`, so I can use the methods there to format the DNs instead...

yrro:
> thanks for working on this, but could you make it do both, write a file for pki-kit and ca-certificates?

No problem. I'll move the code that actually writes out both files to `BaseTaskNamespace` and call the methods from the platform-specific classes.

ipaplatform/debian/tasks.py
yrro: I'm getting the impression that I should avoid using module in future! :)

ipaplatform/base/tasks.py
```diff
@@ -73,6 +73,23 @@ def insert_ca_certs_into_systemwide_ca_store(self, ca_certs):
         Returns True if the operation succeeded, False otherwise.
         """
 
+        try:
```
yrro: An explanation of the error handling here: before `platform_insert_ca_certs` throws an exception, it logs the context that lets the user figure out exactly what caused the exception to be thrown. That way, the utility functions that manipulate the CA store can exit simply any early when something goes wrong; the logic that handles the failure of those utility functions is in one place (here); and the user gets a useful message such as:

```
ipaplatform.base.tasks: ERROR: Could not populate systemwide CA store
  File "/usr/lib/python3/dist-packages/ipaplatform/base/tasks.py", line 77, in insert_ca_certs_into_systemwide_ca_store
    if self.platform_insert_ca_certs(ca_certs):
  File "/usr/lib/python3/dist-packages/ipaplatform/debian/tasks.py", line 106, in platform_insert_ca_certs
    self.write_p11kit_certs(paths.IPA_P11_KIT, ca_certs),
  File "/usr/lib/python3/dist-packages/ipaplatform/redhat/tasks.py", line 316, in write_p11kit_certs
    f = open(path, 'w')
FileNotFoundError: [Errno 2] No such file or directory: '/etc/pki/ca-trust/source/ipa.p11-kit'
```
The first log line was produced by the inner utility function before it re-threw the exception.

ipaplatform/base/tasks.py
```diff
@@ -73,6 +73,23 @@ def insert_ca_certs_into_systemwide_ca_store(self, ca_certs):
         Returns True if the operation succeeded, False otherwise.
```
yrro: The clients of these functions don't actually check the return code...

yrro:
This is ready for another review.

 * Certificate filenames are now built from the Subject DN and serial number,
 * `/usr/local/share/ca-certificates/ipa.p11-kit` is created (for information only since `p11-kit` in Debian does not use it)

tjaalton:
./ipaplatform/debian/tasks.py:152:80: E501 line too long (80 > 79 characters)
this makes the build fail
also, I think the commits should be squashed

yrro:
Yup, I haven't squashed yet because it makes the PR more difficult to review

abbra:
@yrro please squash fixup comments so that we can do actual review. This is how we do all the reviews and development.

ipaplatform/debian/tasks.py
abbra: Note that we are not supporting anything below 3.5 already since FreeIPA 4.7.

abbra:
I triggered PR CI for a final run. Hopefully, we'll get multi-container tests support in Azure Pipelines merged soon and then we can run some of Debian GNU/Linux tests in CI too.

abbra:
@tjaalton could you please run a test on Debian manually?

tjaalton:
Hi, I haven't been able to test this because there's a new breakage caused by an updated component in the distro.. but anything is better than what's there now, so I think this should be merged anyway and fixed afterwards if there's something missing.

tjaalton:
Ok, I've tested on Ubuntu 19.10 and backported ipa 4.8.5 plus this patch there, and client install goes fine and the CA store is updated properly. So +1 from me!

abbra:
OK, I'm going to push this in. @tjaalton has been working on CI testing for Debian so I hope this will help to move forward too.

abbra:
master:

* 3985183d73ffa4be0f1a29a846d0de04d838f62a Debian: write out only one CA certificate per file

PR #4106, opened 2020-01-08
ipatests: fix compatibility with python2 (import ConfigParser)

wladich:
Patch ba4aaa73 backported remote_ini_file utility which requires
`configparser` module. But in python2 this module was named ConfigParser.
As we need to support both python2 and python3 in branch ipa-4-6,
I have added try-except block to allow fallback to python2 verstion import.

Related to: https://pagure.io/SSSD/sssd/issue/3937

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -30,7 +30,7 @@ import itertools
 import tempfile
 import time
 from pipes import quote
-import configparser
+from six.moves import configparser
 from contextlib import contextmanager
 
 import dns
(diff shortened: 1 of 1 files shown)


ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -30,7 +30,10 @@
 import tempfile
 import time
 from pipes import quote
-import configparser
+try:
+    import configparser
+except ImportError:
+    import ConfigParser
```
tiran: ``import ConfigParser as configparser``

rcritten: or from configparser/ConfigParser import RawConfigParser

then change configparser.RawConfigParser to just RawConfigParser

flo-renaud: That can be simplified with:
from six.moves import configparser

wladich: fixed

wladich:
@flo-renaud 
I have fixed import, also added check to use methods of RawConfigParser available in each version of Python.

rcritten:
This LGTM, will give @tiran the final say.

wladich:
@tiran Please have a look at the PR

wladich:
ipa-4-6:

* 0ad66fc17db76187fb869983ded2b2c60e40d4a3 ipatests: fix compatibility with python2 (import ConfigParser)

PR #4108, opened 2020-01-09
Tier-1 test for ipa-healthcheck tool

menonsudhir:
Signed-off-by: sumenon <sumenon@redhat.com>

### DIFF
--- a/ipatests/prci_definitions/nightly_latest.yaml
+++ b/ipatests/prci_definitions/nightly_latest.yaml
@@ -1265,6 +1265,18 @@ jobs:
         timeout: 3600
         topology: *master_1repl
 
+  fedora-latest/test_ipahealthcheck:
+    requires: [fedora-latest/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-latest/build_url}'
+        test_suite: test_integration/test_ipahealthcheck.py
+        template: *ci-master-latest
+        timeout: 3600
+        topology: *master_1repl
+
   fedora-latest/test_ntp_options:
     requires: [fedora-latest/build]
     priority: 50
--- a/ipatests/prci_definitions/nightly_latest_pki.yaml
+++ b/ipatests/prci_definitions/nightly_latest_pki.yaml
@@ -783,6 +783,19 @@ jobs:
         timeout: 3600
         topology: *master_1repl
 
+  pki-fedora/test_ipahealthcheck:
+    requires: [pki-fedora/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{pki-fedora/build_url}'
+        update_packages: True
+        test_suite: test_integration/test_ipahealthcheck.py
+        template: *pki-master-latest
+        timeout: 3600
+        topology: *master_1repl
+
   pki-fedora/test_pkinit_manage:
     requires: [pki-fedora/build]
     priority: 50
--- a/ipatests/prci_definitions/nightly_rawhide.yaml
+++ b/ipatests/prci_definitions/nightly_rawhide.yaml
@@ -1364,6 +1364,19 @@ jobs:
         timeout: 3600
         topology: *master_1repl
 
+  fedora-rawhide/test_ipahealthcheck:
+    requires: [fedora-rawhide/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-rawhide/build_url}'
+        update_packages: True
+        test_suite: test_integration/test_ipahealthcheck.py
+        template: *ci-master-frawhide
+        timeout: 3600
+        topology: *master_1repl
+
   fedora-rawhide/test_ntp_options:
     requires: [fedora-rawhide/build]
     priority: 50
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -1994,3 +1994,40 @@ def get_logsize(host, logfile):
     """ get current logsize"""
     logsize = len(host.get_file_contents(logfile))
     return logsize
+
+
+def get_platform(host):
+    result = host.run_command([
+        'python3', '-c',
+        'from ipaplatform.osinfo import OSInfo; print(OSInfo().platform)'
+    ], raiseonerr=False)
+    assert result.returncode == 0
+    return result.stdout_text.strip()
+
+
+def install_packages(host, pkgs):
+    """Install packages on a remote host.
+    :param host: the host where the installation takes place
+    :param pkgs: packages to install, provided as a list of strings
+    """
+    platform = get_platform(host)
+    # Only supports RHEL 8+ and Fedora for now
+    if platform in ('rhel', 'fedora'):
+        install_cmd = ['/usr/bin/dnf', 'install', '-y']
+    else:
+        raise ValueError('install_packages: unknown platform %s' % platform)
+    host.run_command(install_cmd + pkgs)
+
+
+def uninstall_packages(host, pkgs):
+    """Uninstall packages on a remote host.
+    :param host: the host where the uninstallation takes place
+    :param pkgs: packages to uninstall, provided as a list of strings
+    """
+    platform = get_platform(host)
+    # Only supports RHEL 8+ and Fedora for now
+    if platform in ('rhel', 'fedora'):
+        install_cmd = ['/usr/bin/dnf', 'remove', '-y']
+    else:
+        raise ValueError('install_packages: unknown platform %s' % platform)
+    host.run_command(install_cmd + pkgs)
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -1,0 +1,496 @@
+# Copyright (C) 2020  FreeIPA Contributors see COPYING for license
+#
+
+"""
+Tests to verify that the ipa-healthcheck scenarios
+"""
+
+from __future__ import absolute_import
+
+from ipatests.test_integration.base import IntegrationTest
+from ipatests.pytest_ipa.integration import tasks
+from ipalib import api
+from ipapython.ipaldap import realm_to_serverid
+
+import json
+
+HEALTHCHECK_LOG = "/var/log/ipa/healthcheck/healthcheck.log"
+HEALTHCHECK_SYSTEMD_FILE = (
+    "/etc/systemd/system/multi-user.target.wants/ipa-healthcheck.timer"
+)
+HEALTHCHECK_LOG_ROTATE_CONF = "/etc/logrotate.d/ipahealthcheck"
+HEALTHCHECK_LOG_DIR = "/var/log/ipa/healthcheck"
+HEALTHCHECK_OUTPUT_FILE = "/tmp/output.json"
+HEALTHCHECK_PKG = ['freeipa-healthcheck']
+TOMCAT_CFG = "/var/lib/pki/pki-tomcat/conf/ca/CS.cfg"
+
+
+sources = [
+    "ipahealthcheck.dogtag.ca",
+    "ipahealthcheck.ds.replication",
+    "ipahealthcheck.dogtag.ca",
+    "ipahealthcheck.ipa.certs",
+    "ipahealthcheck.ipa.dna",
+    "ipahealthcheck.ipa.idns",
+    "ipahealthcheck.ipa.files",
+    "ipahealthcheck.ipa.host",
+    "ipahealthcheck.ipa.roles",
+    "ipahealthcheck.ipa.topology",
+    "ipahealthcheck.ipa.trust",
+    "ipahealthcheck.meta.services",
+]
+
+ipa_cert_checks = [
+    "IPACertmongerExpirationCheck",
+    "IPACertfileExpirationCheck",
+    "IPACertTracking",
+    "IPACertNSSTrust",
+    "IPANSSChainValidation",
+    "IPAOpenSSLChainValidation",
+    "IPARAAgent",
+    "IPACertRevocation",
+    "IPACertmongerCA",
+    "IPACAChainExpirationCheck",
+]
+
+ipatrust_checks = [
+    "IPATrustAgentCheck",
+    "IPATrustDomainsCheck",
+    "IPADomainCheck",
+    "IPATrustCatalogCheck",
+    "IPAsidgenpluginCheck",
+    "IPATrustAgentMemberCheck",
+    "IPATrustControllerPrincipalCheck",
+    "IPATrustControllerServiceCheck",
+    "IPATrustControllerConfCheck",
+    "IPATrustControllerGroupSIDCheck",
+    "IPATrustPackageCheck",
+]
+
+metaservices_checks = [
+    "certmonger",
+    "dirsrv",
+    "gssproxy",
+    "httpd",
+    "ipa_custodia",
+    "ipa_dnskeysyncd",
+    "ipa_otpd",
+    "kadmin",
+    "krb5kdc",
+    "named",
+    "pki_tomcatd",
+    "sssd",
+]
+
+ipafiles_checks = ["IPAFileNSSDBCheck", "IPAFileCheck", "TomcatFileCheck"]
+dogtag_checks = ["DogtagCertsConfigCheck", "DogtagCertsConnectivityCheck"]
+iparoles_checks = ["IPACRLManagerCheck", "IPARenewalMasterCheck"]
+replication_checks = ["ReplicationConflictCheck"]
+ruv_checks = ["RUVCheck"]
+dna_checks = ["IPADNARangeCheck"]
+idns_checks = ["IPADNSSystemRecordsCheck"]
+ipahost_checks = ["IPAHostKeytab"]
+ipatopology_checks = ["IPATopologyDomainCheck"]
+filesystem_checks = ["FileSystemSpaceCheck"]
+metacore_checks = ["MetaCheck"]
+
+DEFAULT_PKI_CA_CERTS = [
+    "caSigningCert cert-pki-ca",
+    "ocspSigningCert cert-pki-ca",
+    "subsystemCert cert-pki-ca",
+    "auditSigningCert cert-pki-ca",
+    "Server-Cert cert-pki-ca",
+]
+DEFAULT_PKI_KRA_CERTS = [
+    "transportCert cert-pki-kra",
+    "storageCert cert-pki-kra",
+    "auditSigningCert cert-pki-kra",
+]
+
+
+class TestIpaHealthCheck(IntegrationTest):
+    """
+    Tier-1 test for ipa-healthcheck tool with IPA Master setup with
+    dns and IPA Replica with dns enabled
+    """
+
+    num_replicas = 1
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master, setup_dns=True)
+        tasks.install_replica(cls.master, cls.replicas[0], setup_dns=True)
+
+    def test_ipa_healthcheck_install_on_master(self):
+        """
+        Testcase to check healthcheck package is installed
+        succesfully on IPA master.
+        """
+        tasks.install_packages(self.master, HEALTHCHECK_PKG)
+
+    def test_ipa_healthcheck_install_on_replica(self):
+        """
+        Testcase to check healthcheck package is installed
+        succesfully on IPA replica.
+        """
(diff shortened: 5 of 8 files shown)


stanislavlevin:
Hello, @menonsudhir.
Apparently, non-dnf users will not be happy.

stanislavlevin:
Is it required to install something during testing? Should it be included in test images instead?

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,168 @@
...
+ipatopology_checks = ["IPATopologyDomainCheck"]
+filesystem_checks = ["FileSystemSpaceCheck"]
+metacore_checks = ["MetaCheck"]
+
+
+class TestIpaHealthCheck(IntegrationTest):
+    """
+    Check that ipa healthcheck tool return WARNING if no range is set.
+    https://github.com/freeipa/freeipa-healthcheck/issues/60
+    """
```
rcritten: The comment doesn't match the content. This is just verifying that the default set of sources and checks are available.

rcritten:
I agree that the package should either be in the image or you'll need to abstract something out into tasks to do per-platform installation of a package. I'm not sure if freeipa-healthcheck is available on non-Fedora platforms yet (I wasn't able to find it).

stanislavlevin:
@rcritten, eventually, ipa-healthcheck will be everywhere :wink:

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,446 @@
...
+            [
+                "ipa-healthcheck",
+                "--source",
+                "ipahealthcheck.meta.services",
+                "--check",
+                "sssd",
+                "--output-file",
+                HEALTHCHECK_OUTPUT_FILE,
+            ],
+            raiseonerr=False,
```
mrizwan93: same here. Consider using scenario in case of fail/pass of command.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,446 @@
...
+            [
+                "ipa-healthcheck",
+                "--source",
+                "ipahealthcheck.dogtag.ca",
+                "--check",
+                "DogtagCertsConfigCheck",
+                "--output-file",
+                HEALTHCHECK_OUTPUT_FILE,
+            ],
+            raiseonerr=False,
```
mrizwan93: consider using pass/fail scenario here.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,446 @@
...
+        )
+        contents = self.master.get_file_contents(
+            HEALTHCHECK_OUTPUT_FILE, encoding="utf-8"
+        )
+        data = json.loads(contents)
+        assert data[0]["result"] == "SUCCESS"
+        if osinfo.id == "fedora" and osinfo.version_number == (31):
```
mrizwan93: This will probably always fail on non-fedora system.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,446 @@
...
+            "Failed to obtain host TGT: Major (851968): "
+            "Unspecified GSS failure.  "
+            "Minor code may provide more information, "
+            "Minor (2529638972): Generic error (see e-text)"
+        )
+        dashed_domain = self.master.domain.realm.replace(".", "-")
+        dirsrv_service = "dirsrv@%s.service" % dashed_domain
+        self.master.run_command(["systemctl", "stop", dirsrv_service])
+        result = self.master.run_command(
+            ["systemctl", "status", dirsrv_service], raiseonerr=False
```
mrizwan93: You might want to check the status of dirsrv here and probably want to make use of `result` variable :)

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,446 @@
...
+
+    def test_source_ipa_roles_check_crlmanager(self):
+        """
+        This testcase checks the status of healthcheck tool
+        reflects correct information when crlgen is disabled
+        using ipa-crl-manage disable
+        """
+        msg = "The ipa-crlgen-enable command was successful"
+        cmd = self.master.run_command(["ipa-crl-manage", "disable"])
+        assert cmd.stdout_text == msg
```
mrizwan93: I am wondering how `ipa-crl-manage disable` will return message `The ipa-crlgen-enable command was successful`!

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,446 @@
...
+            assert check["kw"]["crlgen_enabled"] == "false"
+
+    def test_ipa_healthcheck_remove(self):
+        """
+        Testcase to check removal of healthcheck tool
+        """
+        cmd = self.master.run_command(
+            ["dnf", "remove", "-y", "freeipa-healthcheck"]
+        )
+        assert cmd.returncode == 0
```
mrizwan93: No need of this. `self.master.run_command(["dnf", "remove", "-y", "freeipa-healthcheck"])` will do.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,446 @@
...
+    Test for ipa-healthcheck tool with IPA Master and
+    replica without DNS installed
+    """
+
+    num_replicas = 1
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master, setup_dns=True)
+        cls.replica = cls.replicas[0]
```
mrizwan93: No need of this, can be removed.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,446 @@
...
+
+HEALTHCHECK_LOG = "/var/log/ipa/healthcheck/healthcheck.log"
+HEALTHCHECK_SYSTEMD_FILE = (
+    "/etc/systemd/system/multi-user.target.wants/ipa-healthcheck.timer"
+)
+HEALTHCHECK_LOG_ROTATE_CONF = "/etc/logrotate.d/ipahealthcheck"
+HEALTHCHECK_LOG_DIR = "/var/log/ipa/healthcheck"
+HEALTHCHECK_OUTPUT_FILE = "/tmp/output.json"
+HEALTHCHECK_PKG = "freeipa-healthcheck"
+TOMCAT_CFG = "/var/lib/pki/pki-tomcat/conf/ca/CS.cfg"
```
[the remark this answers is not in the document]
menonsudhir: @mrizwan93, not adding these healthcheck constants to ipaplatform since it is only used by the tests and not by IPA itself as required in another PR.

rcritten:
What is the purpose for having test_ipahealthcheck.py and test_ipahealthcheck3.py?

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -1994,3 +1994,13 @@ def get_logsize(host, logfile):
...
+
+
+def install_package(host, pkgs):
+    dnf_cmd = '/usr/bin/dnf'
+    yum_cmd = '/usr/bin/yum'
+    result = host.run_command(['which', 'dnf'], raiseonerr=False)
+    if result.returncode == 0:
+        host.run_command([dnf_cmd, 'install', '-y', pkgs])
+    else:
+        host.run_command([yum_cmd, 'install', '-y', pkgs])
```
rcritten: I'm working on a candidate patch which will make this more generic. I'll provide it once I can get the tests to pass locally.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,446 @@
...
+        succesfully on IPA replica.
+        """
+        tasks.install_package(self.replicas[0], HEALTHCHECK_PKG)
+
+    def test_run_ipahealthcheck_list_source(self):
+        """
+        Testcase to verify sources available in healthcheck tool.
+        """
+        result = self.master.run_command(["ipa-healthcheck", "--list-sources"])
+        assert result.returncode == 0
```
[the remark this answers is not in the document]
rcritten: Agreed, because an exception will be raised.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,446 @@
...
+            [
+                "ipa-healthcheck",
+                "--source",
+                "ipahealthcheck.meta.services",
+                "--check",
+                "sssd",
+                "--output-file",
+                HEALTHCHECK_OUTPUT_FILE,
+            ],
+            raiseonerr=False,
```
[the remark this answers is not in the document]
rcritten: IMHO this correct usage. This is expected to fail because he just stopped sssd.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,507 @@
...
+        contents = self.master.get_file_contents(
+            HEALTHCHECK_OUTPUT_FILE, encoding="utf-8"
+        )
+        data = json.loads(contents)
+        for check in data:
+            assert check[0]["result"] == "ERROR"
+            assert check[0]["kw"]["msg"] == "sssd: not running"
+            assert check[0]["kw"]["status"] == "False"
+        self.master.run_command(["systemctl", "start", "sssd"],
+                                raiseonerr=False)
```
rcritten: Might want to drop the raiseonerr here since we do want things to blow up if sssd fails to start.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,507 @@
...
+        )
+        data = json.loads(contents)
+        assert data[0]["result"] == "SUCCESS"
+        if osinfo.id == "fedora" and osinfo.version_number == (30):
+            assert data[0]["kw"]["ipa_version"] == "4.8.3"
+            assert data[0]["kw"]["ipa_api_version"] == "2.235"
+        else:
+            assert data[0]["kw"]["ipa_version"] == "4.9"
+            assert data[0]["kw"]["ipa_api_version"] == "2.235"
+
```
rcritten: This one test is going to be difficult to maintain because with every new release it will need to be updated as well. It's not a bad test but I'd expect failures. Perhaps instead you could compare that with either the scraped output of ipa --version or by:

result = self.master.run_command(['python3', '-c', 'from ipapython import version; print("%s\t%s" % (version.VERSION, version.API_VERSION))']

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,507 @@
...
+        running on IPA master
+        """
+        msg = (
+            "Failed to obtain host TGT: Major (851968): "
+            "Unspecified GSS failure.  "
+            "Minor code may provide more information, "
+            "Minor (2529638972): Generic error (see e-text)"
+        )
+        dashed_domain = self.master.domain.realm.replace(".", "-")
+        dirsrv_service = "dirsrv@%s.service" % dashed_domain
```
rcritten: Instead of dashed_domain:

from ipapython.ipaldap import realm_to_serverid
serverid = realm_to_serverid(api.env.realm)
dirsrv_service = "dirsrv@%s.service" % serverid

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,507 @@
...
+            raiseonerr=False,
+        )
+        contents = self.master.get_file_contents(
+            HEALTHCHECK_OUTPUT_FILE, encoding="utf-8"
+        )
+        data = json.loads(contents)
+        for check in data:
+            assert check["result"] == "SUCCESS"
+            assert check["kw"]["key"] == "crl_manager"
+            assert check["kw"]["crlgen_enabled"] == "True"
```
rcritten: Arguably if you are disabling this master as the CRL generator then healthcheck should return False here.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,507 @@
...
+
+    def test_ipa_healthcheck_remove(self):
+        """
+        This testcase checks the removal of of healthcheck tool
+        on replica and master
+        """
+        self.master.run_command(
+            ["dnf", "remove", "-y", "freeipa-healthcheck"])
+        self.replicas[0].run_command(
+            ["dnf", "remove", "-y", "freeipa-healthcheck"])
```
rcritten: The patch I'm going to propose will handle this in a generic way as well in tasks.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,507 @@
...
+        )
+        data = json.loads(contents)
+        for check in data:
+            assert check["result"] == "WARNING"
+            assert check["kw"]["msg"] == msg1 or \
+                check["kw"]["msg"] == msg2
+        tasks.uninstall_master(self.master)
+
+
+class TestIpaHealthCheckWithKRA(IntegrationTest):
```
rcritten: This is already an extended test due to three separate server installations plus a replica. You could save an install by moving this into the first invocation and running ipa-kra-install instead as part of this test case.

rcritten:
@menonsudhir I sent you two patches out-of-band. One changes the task for installing/uninstall packages to be more flexible for other distros (hopefully) and the second fixes up the test cases to pass (minus the version one which I commented out in my testing).

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -0,0 +1,497 @@
+# Copyright (C) 2018  FreeIPA Contributors see COPYING for license
```
rcritten: nit: 2018 - > 2020

rcritten:
Beyond the one little nit this looks good to me.
Let's do this:
- fix the date
- drop the temp commit, no need to test again for changing a comment
- then I'll ack and push

We need this in ipa-4-8 as well right? Can you add that tag if so?

ipatests/prci_definitions/nightly_latest_pki.yaml
```diff
@@ -783,6 +783,18 @@ jobs:
         timeout: 3600
         topology: *master_1repl
 
+  pki-fedora/test_ipahealthcheck_TestIpaHealthCheck:
+    requires: [fedora-latest/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{pki-fedora/build_url}'
```
flo-renaud: 1. Within this file the job definition needs to have `update_packages: True` to make sure to use the nightly build of pki.
2. Does this test really need to be executed for dogtag regression testing?

ipatests/prci_definitions/nightly_rawhide.yaml
```diff
@@ -1364,6 +1364,18 @@ jobs:
...
+    requires: [fedora-rawhide/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-rawhide/build_url}'
+        update_packages: True
+        test_suite: test_integration/test_ipahealthcheck::TestIpaHealthCheck
+        template: *ci-master-frawhide
+        timeout: 3600
```
flo-renaud: The job definition is missing `topology: *master_1repl`

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -1994,3 +1994,32 @@ def get_logsize(host, logfile):
...
+    else:
+        raise ValueError('install_packages: unknown platform %s' % platform)
+    host.run_command(install_cmd + pkgs)
+
+
+def uninstall_packages(host, pkgs):
+    platform = get_platform(host)
+    # Only supports RHEL 8+ and Fedora for now
+    if platform in ('rhel', 'fedora'):
+        install_cmd = ['/usr/bin/dnf', 'erase', '-y']
```
flo-renaud: `dnf erase` is a deprecated alias for `dnf remove`. I would rather use the remove command.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -1994,3 +1994,32 @@ def get_logsize(host, logfile):
...
+def get_platform(host):
+    result = host.run_command([
+        'python3', '-c',
+        'from ipaplatform.osinfo import OSInfo; print(OSInfo().platform)'
+    ], raiseonerr=False)
+    assert result.returncode == 0
+    return result.stdout_text.strip()
+
+
+def install_packages(host, pkgs):
```
flo-renaud: I would add a docstring stating that a list is expected for pkgs.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -1994,3 +1994,32 @@ def get_logsize(host, logfile):
...
+    platform = get_platform(host)
+    # Only supports RHEL 8+ and Fedora for now
+    if platform in ('rhel', 'fedora'):
+        install_cmd = ['/usr/bin/dnf', 'install', '-y']
+    else:
+        raise ValueError('install_packages: unknown platform %s' % platform)
+    host.run_command(install_cmd + pkgs)
+
+
+def uninstall_packages(host, pkgs):
```
flo-renaud: I would add a docstring stating that a list is expected for pkgs.

ipatests/prci_definitions/nightly_latest.yaml
```diff
@@ -1265,6 +1265,18 @@ jobs:
...
         topology: *master_1repl
 
+  fedora-latest/test_ipahealthcheck_TestIpaHealthCheck:
+    requires: [fedora-latest/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-latest/build_url}'
+        test_suite: test_integration/test_ipahealthcheck.py::TestIpaHealthCheck
```
flo-renaud: As test_ipahealthcheck.py currently contains a single test class, you can also define `test_suite: test_integration/test_ipahealthcheck.py`. This way, if the py file is later extended with another test class you won't need to modify the job definition.
Up to you, but if you implement my suggestion, change the test name into `fedora-latest/test_ipahealthcheck`.

ipatests/test_integration/test_ipahealthcheck.py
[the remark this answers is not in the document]
menonsudhir: @mrizwan93 @rcritten removed the assert statement.

ipatests/prci_definitions/nightly_latest.yaml
menonsudhir: @flo-renaud added the above mentioned changes.

ipatests/prci_definitions/nightly_latest_pki.yaml
```diff
@@ -783,6 +783,19 @@ jobs:
         timeout: 3600
         topology: *master_1repl
 
+  pki-fedora/test_ipahealthcheck:
+    requires: [fedora-latest/build]
```
flo-renaud: `requires` should be `[pki-fedora/build]`

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -1994,3 +1994,38 @@ def get_logsize(host, logfile):
...
+        'from ipaplatform.osinfo import OSInfo; print(OSInfo().platform)'
+    ], raiseonerr=False)
+    assert result.returncode == 0
+    return result.stdout_text.strip()
+
+
+def install_packages(host, pkgs):
+    '''
+    For install_packages function list is expected for pkgs argument
+    '''
```
flo-renaud: It's good practice to use a consistent format for the doc strings inside the file:
```
"""Install packages on a remote host.

:param host: the host where the installation takes place
:param pkgs: packages to install, provided as a list of strings
"""
```

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -1994,3 +1994,38 @@ def get_logsize(host, logfile):
...
+    else:
+        raise ValueError('install_packages: unknown platform %s' % platform)
+    host.run_command(install_cmd + pkgs)
+
+
+def uninstall_packages(host, pkgs):
+    '''
+    For uninstall_packages function list is expected for pkgs argument
+    '''
+    platform = get_platform(host)
```
flo-renaud: Please use a consistent format for the docstring.

flo-renaud:
@menonsudhir 
thanks for the PR, looks good to me. I'll leave the final ACK to @rcritten  as he did most of the review.

rcritten:
master:

* b5c8efa33c9460337d9ae92fd1ec8215074667a5 Tier-1 test for ipa-healthcheck tool
* 000703c89f18a190e5cac17ad7113a2ed067c35b Nightly definition for ipa-healthcheck tool

PR #4110, opened 2020-01-09
AD user without override receive InternalServerError with API

flo-renaud:
When ipa commands are used by an Active Directory user that
does not have any idoverride-user set, they return the
following error message which can be misleading:
$ kinit aduser@ADDOMAIN.COM
$ ipa ping
ipa: ERROR: cannot connect to 'https://master.ipa.com/ipa/json': Internal Server Error

The fix properly handles ACIError exception received when
creating the context, and now the following message can be seen:

$ kinit aduser@ADDOMAIN.COM
$ ipa ping
ipa: ERROR: cannot connect to 'https://master.ipa.com/ipa/json': Unauthorized

with the following log in /var/log/httpd/error_log:
ipa: INFO: 401 Unauthorized: Insufficient access:  Invalid credentials

Fixes: https://pagure.io/freeipa/issue/8163

### DIFF
--- a/ipaserver/rpcserver.py
+++ b/ipaserver/rpcserver.py
@@ -694,7 +694,7 @@ class KerberosWSGIExecutioner(WSGIExecutioner, KerberosSession):
             status = HTTP_STATUS_SUCCESS
             response = status.encode('utf-8')
             start_response(status, self.headers)
-            return self.marshal(None, e)
+            return [self.marshal(None, e)]
         finally:
             destroy_context()
         return response


abbra:
LGTM

abbra:
Should this be backported down to ipa-4-6 too?

flo-renaud:
On rhel 7.8 the behavior is different (probably because the server is using python2) and httpd's error log already displays `ipa: INFO: 401 Unauthorized: Insufficient access:  Invalid credentials`, meaning I don't think we need a backport to ipa-4-6. Adding label for ipa-4-8 only.

abbra:
Ok. I added ipa-4-7 too as we haven't closed the branch yet. Please push.

flo-renaud:
master:

* e2d69380fbae6dab244fce3fd5afe5fdc28a2663 AD user without override receive InternalServerError with API

PR #4114, opened 2020-01-10
Print LDAP diagnostic messages on error

tiran:
ipa_ldap_init(), ipa_tls_ssl_init(), and the bind operations of ipa-join
and ipa-getkeytab now print LDAP error string and LDAP diagonstic messages
to stderr.

### DIFF
--- a/util/ipa_ldap.c
+++ b/util/ipa_ldap.c
@@ -24,6 +24,43 @@
 
 #include "ipa_ldap.h"
 
+
+/** Print LDAP error message to stderr
+ *
+ * The help function prints custom error message, LDAP error string,
+ * diagnostic message (if available) to stderr.
+ *
+ * @param ld      LDAP connection
+ * @param errnum  error code from LDAP operation
+ * @param msg     Additional custom error message (must include trailing
+ *                new line)
+ */
+void ipa_ldap_error(LDAP *ld, int errnum, char *msg)
+{
+    const char *errstring;
+    char *diagnostic = NULL;
+    int ret = 0;
+    int has_diagnostic = 0;
+
+    /* print custom message msg first. All msg strings have trailing newline
+     */
+    fprintf(stderr, "%s", msg);
+
+    /* Get human readable error string and diagnostic message */
+    errstring = ldap_err2string(errnum);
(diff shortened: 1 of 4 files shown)


util/ipa_ldap.c
```diff
@@ -24,6 +24,35 @@
...
+    ret = ldap_get_option(ld, LDAP_OPT_DIAGNOSTIC_MESSAGE, (void*)&diagnostic);
+    has_diagnostic = ((ret == LDAP_SUCCESS) && diagnostic && *diagnostic);
+
+    if (errstring && has_diagnostic) {
+        fprintf(stderr, "%s: %s\n", errstring, diagnostic);
+    } else if (errstring) {
+        fprintf(stderr, "%s\n", errstring);
+    } else if (has_diagnostic) {
+        fprintf(stderr, "%i: %s", errnum, diagnostic);
+    }
```
rcritten: Can you add a doc-string that msg needs to be new-line terminated otherwise the output will be smashed together?

Should the corner case of errstring=NULL and has_diagnostic=0 be handled?

tiran:
Example error message:

```
# ./ipa-getkeytab -p test/invalid -k /tmp/keytab -s untrusted.server.example
Unable to initialize STARTTLS session
Connect error: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed (self signed certificate in certificate chain)
Failed to bind to server!
Retrying with pre-4.0 keytab retrieval method...
Unable to initialize STARTTLS session
Connect error: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed (self signed certificate in certificate chain)
Failed to bind to server!
Failed to get keytab
````

util/ipa_ldap.c
rcritten: Arguably there is nothing to print in that case. Perhaps I wasn't quite awake yet :/ The docstring might be nice though.

tiran: I have added a doc string to the new function.

rcritten:
This LGTM.

tiran:
master:

* e107b8e4fe1f205cc1bf992aae2e46a7ab02ebb6 Print LDAP diagnostic messages on error

PR #3842, opened 2019-11-01
[KDB] various code hygiene fixes

frozencemetery:

PR #4126, opened 2020-01-13
Support KDB DAL version 8.0

frozencemetery:
A krb5 build containing the 1.18 beta - which has this KDB version - can be found here for testing: https://koji.fedoraproject.org/koji/taskinfo?taskID=40498937

Note that samba isn't ready to go with KDB 8.0, so I don't think this can actually build.  I've reached out to @iboukris to hopefully coordinate that.

(This is based on #3842).

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb.c
+++ b/daemons/ipa-kdb/ipa_kdb.c
@@ -751,8 +751,67 @@ kdb_vftabl kdb_function_table = {
 };
 #endif
 
+#if (KRB5_KDB_DAL_MAJOR_VERSION == 8)
+/* Version 8 adds several arguments here.  However, if we want to actually use
+ * them in mspac, we really ought to drop support for older DAL versions. */
+static inline krb5_error_code
+stub_sign_authdata(krb5_context context, unsigned int flags,
+                   krb5_const_principal client_princ,
+                   krb5_const_principal server_princ, krb5_db_entry *client,
+                   krb5_db_entry *server, krb5_db_entry *header_server,
+                   krb5_db_entry *local_tgt, krb5_keyblock *client_key,
+                   krb5_keyblock *server_key, krb5_keyblock *header_key,
+                   krb5_keyblock *local_tgt_key, krb5_keyblock *session_key,
+                   krb5_timestamp authtime, krb5_authdata **tgt_auth_data,
+                   void *ad_info, krb5_data ***auth_indicators,
+                   krb5_authdata ***signed_auth_data)
+{
+    krb5_db_entry *krbtgt = header_server ? header_server : server;
+    krb5_keyblock *krbtgt_key = header_key ? header_key : server_key;
+
+    return ipadb_sign_authdata(context, flags, client_princ, client, server,
+                               krbtgt, client_key, server_key, krbtgt_key,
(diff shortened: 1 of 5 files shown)


abbra:
Thanks, @frozencemetery. I think this all looks mostly good. Once we have Samba fixes as well, we can add both new krb5 and new samba builds to freeipa-master (https://copr.fedorainfracloud.org/coprs/g/freeipa/freeipa-master/) COPR repo and merge this PR.

The question I have is whether we should merge https://github.com/freeipa/freeipa/pull/3842 on its own and then drop the patches you included in this PR. I'm inclined to do that as PR#3842 did succeed.

frozencemetery:
Rebased; thanks for merging the other PR!

abbra:
Thanks to @frozencemetery and @iboukris work, I was able to submit an updated FreeIPA with krb5 1.18beta support in Rawhide: https://bodhi.fedoraproject.org/updates/FEDORA-2020-d5d4c21e87.

It needs a bit more tests and to see that actual code works fine and then we can merge this PR. However, it needs a rebase -- somehow tests are failing on both Azure and PR CI.

frozencemetery:
Rebased.

If I understand correctly, the tests aren't going to pass on Fedora until the update you created is landed (because the RPMs won't build).

abbra:
They wouldn't pass in any case because they are running on F30, not Rawhide.
However, we need to be able to build on F30 too, so if you would be able to support building with 'old' and 'new' API, then we can merge.
The latter (ability to work with 7.0 KDB API) was implied, right? We only dropped 6.0 API.

frozencemetery:
Yes, but we can only build for one at a time.  I've added a conditional to the spec file.

abbra:
Linter actually failed with spec file changes:
```
./freeipa.spec: E: specfile-error warning: line 154: Possible unexpanded macro in: BuildRequires:  krb5-devel >= %{krb5_version}
./freeipa.spec: E: specfile-error warning: line 280: Possible unexpanded macro in: BuildRequires:  krb5-server >= %{krb5_version}
./freeipa.spec: E: specfile-error warning: line 306: Possible unexpanded macro in: Requires(post): krb5-server >= %{krb5_version}
./freeipa.spec: E: specfile-error warning: line 308: Possible unexpanded macro in: Requires: krb5-pkinit-openssl >= %{krb5_version}
./freeipa.spec: E: specfile-error warning: line 506: Possible unexpanded macro in: Requires: krb5-workstation >= %{krb5_version}
```

You removed `%{krb5_version}` definition for fedora but it is still in use. I guess you can just undo the removal part for this macro.

frozencemetery:
Fixed.  I'd really like us to stop using that logic at all, but I agree that discussion is not in scope for this PR.

abbra:
LGTM.

flo-renaud:
master:

* 1c787cc36c42f05fe945c4a8ae8551832e0e2c1d Handle the removal of KRB5_KDB_FLAG_ALIAS_OK
* ff10f3fa18948971912bab9ad5bfbc3d36133a3b Support DAL version 8.0
* 93e81cfd0c1d4d759985ed1ce3f4aa540dcf438e Drop support for DAL version 5.0

PR #4127, opened 2020-01-14
ipa tests: fix backup and restore

flo-renaud:
The tests for backup_and_restore check that the ipa-backup command
compresses the tar file AFTER restarting IPA services by reading the
output and looking for a pattern with "gzip" before "Starting IPA service."

As the tar file name is randomly created, it sometimes happen that the
name contains gzip and in this case the test wrongly assumes that
the gzip cmd was called.

The fix makes a stricter comparison, looking for /bin/gzip.

Fixes: https://pagure.io/freeipa/issue/8170

### DIFF
--- a/ipatests/test_integration/test_backup_and_restore.py
+++ b/ipatests/test_integration/test_backup_and_restore.py
@@ -165,7 +165,7 @@ def backup(host):
 
     # Test for ticket 7632: check that services are restarted
     # before the backup is compressed
-    pattern = r'.*gzip.*Starting IPA service.*'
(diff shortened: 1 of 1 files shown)


flo-renaud:
Adding labels starting from ipa-4-6 as the test issue is present in all the branches.

rcritten:
LGTM, remove the temp commit and I'll add the ack.

flo-renaud:
Hi @rcritten 

rcritten:
master:

* ae140ae40692154ce692c03901f6e17c0e227498 ipatests: fix backup and restore

PR #4135, opened 2020-01-17
lite-setup: configure lite-server test env

tiran:
Introduce a script that configures a local testing environment
with ipa default.conf, krb5.conf, and ca.crt from a server hostname.

The lite server configuration allows easy and convenient testing of 
IPA server and client code. It uses an existing 389-DS and KRB5 KDC 
server on another machine:

    $ contrib/lite-setup.py master.ipa.example
    $ source ~/.ipa/activate.sh
    (ipaenv) $ kinit username
    (ipaenv) $ make lite-server

IPA server UI is available on http://localhost:8888/ipa/

### DIFF
--- a/contrib/README.md
+++ b/contrib/README.md
@@ -1,0 +1,118 @@
+# In-tree development debugging and testing
+
+lite-server and lite-client enable fast development, debugging, and
+performance analysis of server or client code from an in-tree source
+directory. The lite-server runs a local web server that uses a remote
+LDAP and KRB5 server.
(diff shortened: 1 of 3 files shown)


rcritten:
master:

* 0a55e82d916b38472bd38032946ea9b249a765f2 Add tracemalloc support to profile memory usage
* e9ae7c4b898e570440245714cf20f03d8ee35159 lite-setup: configure lite-server test env

PR #4138, opened 2020-01-17
Explain the effect of OPT_X_TLS_PROTOCOL_MIN

tiran:
OpenLDAP 2.4 sets minimum version with SSL_CTX_set_options(). The
system-wide crypto-policies for TLS minimum version are applied
with SSL_CTX_set_min_proto_version(). The set_option() call cannot
not enable lower versions than allowed by crypto-policy, e.g.
openssl.cnf MinProtocol=TLS1.2 + OPT_X_TLS_PROTOCOL_MIN=TLS1.0
result in TLS 1.2 as minimum protocol version.

### DIFF
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -127,6 +127,12 @@ def ldap_initialize(uri, cacertfile=None):
             conn.set_option(ldap.OPT_X_TLS_CACERTFILE, cacertfile)
 
         # SSLv3 and SSLv2 are insecure
+        # OpenLDAP 2.4 sets minimum version with SSL_CTX_set_options(). The
+        # system-wide crypto-policies for TLS minimum version are applied
+        # with SSL_CTX_set_min_proto_version(). The set_option() call cannot
+        # enable lower versions than allowed by crypto-policy, e.g.
+        # openssl.cnf MinProtocol=TLS1.2 + OPT_X_TLS_PROTOCOL_MIN=TLS1.0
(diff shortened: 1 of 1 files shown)


ipapython/ipaldap.py
```diff
@@ -127,6 +127,12 @@ def ldap_initialize(uri, cacertfile=None):
             conn.set_option(ldap.OPT_X_TLS_CACERTFILE, cacertfile)
 
         # SSLv3 and SSLv2 are insecure
+        # OpenLDAP 2.4 sets minimum version with SSL_CTX_set_options(). The
+        # system-wide crypto-policies for TLS minimum version are applied
+        # with SSL_CTX_set_min_proto_version(). The set_option() call cannot
+        # not enable lower versions than allowed by crypto-policy, e.g.
```
rcritten: "cannot not" is a double negative in English so this reads "the set_option_call() can enable lower versions than allowed by crypto-policy. I think you need to drop the separate "not" based on the reset of the example, where one cannot override global policy within a library.

abbra: @tiran could you please update the text based on @rcritten comments?

abbra:
This PR seems to stuck.

flo-renaud:
Tests are green and the text was updated as suggested, ACK.

tiran:
master:

* f3e117156436bbf715232bea730a9172c14a8c1f Explain the effect of OPT_X_TLS_PROTOCOL_MIN

PR #4142, opened 2020-01-20
Add delete option to ipa-cacert-manage to remove CA certificates

rcritten:
Before removing a CA re-verify all the other CAs to ensure that
the chain is not broken. Provide a force option to handle cases
where the CA is expired or verification fails for some other
reason, or you really just want them gone.

https://pagure.io/freeipa/issue/8124

### DIFF
--- a/ipaserver/install/ipa_cacert_manage.py
+++ b/ipaserver/install/ipa_cacert_manage.py
@@ -30,7 +30,8 @@ from ipapython import admintool, ipautil
 from ipapython.certdb import (EMPTY_TRUST_FLAGS,
                               EXTERNAL_CA_TRUST_FLAGS,
                               TrustFlags,
-                              parse_trust_flags)
+                              parse_trust_flags,
+                              get_ca_nickname)
 from ipapython.dn import DN
 from ipaplatform.paths import paths
 from ipalib import api, errors, x509
@@ -42,7 +43,8 @@ logger = logging.getLogger(__name__)
 class CACertManage(admintool.AdminTool):
     command_name = 'ipa-cacert-manage'
 
-    usage = "%prog renew [options]\n%prog install [options] CERTFILE"
+    usage = "%prog renew [options]\n%prog install [options] CERTFILE\n" \
+            "%prog delete [options] NICKNAME\n%prog list"
 
     description = "Manage CA certificates."
 
@@ -93,6 +95,12 @@ class CACertManage(admintool.AdminTool):
             help="Trust flags for the certificate in certutil format")
         parser.add_option_group(install_group)
 
+        delete_group = OptionGroup(parser, "Delete options")
+        delete_group.add_option(
+            "-f", "--force", action='store_true',
+            help="Force removing the CA even if chain validation fails")
+        parser.add_option_group(delete_group)
+
     def validate_options(self):
         super(CACertManage, self).validate_options(needs_root=True)
 
@@ -112,6 +120,9 @@ class CACertManage(admintool.AdminTool):
                 parser.error("certificate file name not provided")
         elif command == 'list':
             pass
+        elif command == 'delete':
+            if len(self.args) < 2:
+                parser.error("nickname not provided")
         else:
             parser.error("unknown command \"%s\"" % command)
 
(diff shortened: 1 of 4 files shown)


ipaserver/install/ipa_cacert_manage.py
```diff
@@ -112,6 +120,9 @@ def validate_options(self):
                 parser.error("certificate file name not provided")
         elif command == 'list':
             pass
+        elif command == 'delete':
+            if len(self.args) < 1:
+                parser.error("subject not provided")
```
flo-renaud: The error message should rather mention `nickame not provided` as the help mentions NICKNAME and not subject.

ipaserver/install/ipa_cacert_manage.py
```diff
@@ -112,6 +120,9 @@ def validate_options(self):
                 parser.error("certificate file name not provided")
         elif command == 'list':
             pass
+        elif command == 'delete':
+            if len(self.args) < 1:
```
flo-renaud: Should be `if len(self.args) <2:` as args[0]='delete'

rcritten:
For ipa-certupdate I modified update_db() to drop all CA certs from the NSS database before re-adding.

This allows for the trust to be modifed.

The other option is to skip certs that are in the list to avoid some update churn. Given that this isn't expected to be a frequent operation I think we can live with a bit of brute-force.

ipaserver/install/ipa_cacert_manage.py
```diff
@@ -466,6 +479,66 @@ def list(self):
...
+            tmpdb.create_db()
+            for ca_cert, ca_nickname, ca_trust_flags in ca_certs:
+                tmpdb.add_cert(ca_cert, ca_nickname, ca_trust_flags)
+            loaded = tmpdb.list_certs()
+            logger.debug("loaded raw certs '%s'", loaded)
+
+            tmpdb.delete_cert(nickname)
+
+            for ca_nickname, _trust_flags in loaded:
+                if ca_nickname == nickname:
```
flo-renaud: This check should be done after the check for ipa_ca_nickname == nickname (otherwise it's possible to remove IPA CA).

rcritten: Strange, I had at test for that at some point but lost it somehow. Anyway, added.

I suspect there may be valid reasons to do this so I set the error to be "cannot be deleted with this tool". Too dangerous.

flo-renaud:
Hi @rcritten 
With this patch, deleting an unkown CA produces a wrong message:
```
# ipa-cacert-manage delete wrong
The IPA CA cannot be removed with this tool
The ipa-cacert-manage command failed.
```
Maybe you could add a quick integration test to check the behavior for
- deletion without nickname 
- deletion with an unkown nickname
- deletion of IPA CA
- deletion of a root CA needed by a subCA, with/without -f option
- deletion of a subca

rcritten:
Tests added.

ipatests/test_integration/test_commands.py
```diff
@@ -783,3 +859,65 @@ def test_sss_ssh_authorizedkeys(self):
...
+            raiseonerr=False
+        )
+        assert result.returncode != 0
+        assert 'The IPA CA cannot be removed with this tool' in \
+               result.stderr_text
+
+        # Install 3rd party CA's, Let's Encrypt in this case
+        for cert in (isrgrootx1, letsencryptauthorityx3):
+            with tempfile.NamedTemporaryFile() as f:
+                f.write(cert)
```
flo-renaud: This command is run on the controller, not on the master. As a result, the file exists only on the controller and ipa-cacert-manage does not find the file. The method self.master.put_file_contents() can be used instead.

rcritten:
My test runner is the controller so I didn't notice :/

flo-renaud:
Hi @rcritten 
thanks for the tests, the code LGTM. All my previous tests passed, and I also checked that the command works in a CA-less environment.
Note: in CAless env, the command allows to remove the CA that signed the HTTP/LDAP certs but IMO it's OK.

rcritten:
master:

* acfb6191a1b2b35f85201ae013eaccb9474e0c7d Add delete option to ipa-cacert-manage to remove CA certificates
* 6cb4f4bd50fe25f6ad66a4da1cec9dedfb281f12 ipa-certupdate removes all CA certs from db before adding new ones
* 8e71605c54c31dfa5c9ed9fc0dd640047fa96bf6 Add tests for ipa-cacert-manage delete command

PR #4144, opened 2020-01-21
Make assert_error compatible with Python 3.6

tiran:
The re.Pattern class was introduced in Python 3.7. Use duck-typing to
distinguish between str and re pattern object.

Fixes: https://pagure.io/freeipa/issue/8179


wladich:
LGTM

rcritten:
master:

* 10b62ad6bc967e022202d657b7493440fb5cf736 Make assert_error compatible with Python 3.6

PR #4149, opened 2020-01-22
ipatests: Allow zero-length arguments

stanislavlevin:
Currently, such arguments are eaten by 'ipa-run-tests' script as they
are not quoted.

For example, ```running ipa-run-tests -k ''```
results in the actual invocation would be like as:
```
['/bin/sh',
 '--norc',
 '--noprofile',
 '-c',
 '--',
 "/usr/bin/python3 -c 'import sys,pytest;sys.exit(pytest.main())' -o "
 'cache_dir=/tmp/pytest-of-root/pytest-12/test_ipa_run_tests_empty_expression0/.pytest_cache '
 '--confcutdir=/usr/lib64/python3/site-packages/ipatests -k ']
```

Note: expressions or marks could be empty as a result of the building
of command line args by more high-level tools, scripts, etc.

So, a short-termed solution is the quoting of zero-length arguments.

Fixes: https://pagure.io/freeipa/issue/8173

### DIFF
--- a/ipatests/ipa-run-tests
+++ b/ipatests/ipa-run-tests
@@ -60,7 +60,10 @@ pyt_args = [sys.executable, "-c",
             "import sys,pytest;sys.exit(pytest.main())"] + sys.argv[1:]
 # shell is needed to perform globbing
 sh_args = ["/bin/sh", "--norc", "--noprofile", "-c", "--"]
-pyt_args_esc = ["'{}'".format(x) if " " in x else x for x in pyt_args]
+pyt_args_esc = [
+    f"'{x}'" if not x or " " in x else x
+    for x in pyt_args
+]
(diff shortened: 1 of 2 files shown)


abbra:
Do we need to backport this to ipa-4-6? If yes, we shouldn't use f-strings because they do not exist in py2.

stanislavlevin:
@abbra, the change mentioned in the linked issue landed only at `ipa-4-8` ( Azure Pipelines is leveraged as CI runner since FreeIPA-4.8). So, the target branch is `ipa-4-8`.

abbra:
LGTM. ACK

abbra:
master:

* 902821e8aa47936acac713023bff9a4009d8dfb7 ipatests: Allow zero-length arguments

PR #4150, opened 2020-01-22
ipatest: Test if certmonger reads the token in HSM

mrizwan93:
This is to ensure added HSM support for FreeIPA. This test adds
certificate with sofhsm token and checks if certmonger is tracking
it.

related : https://pagure.io/certmonger/issue/125

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -25,6 +25,7 @@ from ipaplatform.osinfo import osinfo
 from ipaplatform.paths import paths
 from ipaplatform.tasks import tasks as platformtasks
 from ipapython import ipautil
+from ipalib.install.certmonger import wait_for_request
 from ipatests.pytest_ipa.integration import tasks
 from ipatests.pytest_ipa.integration.env_config import get_global_config
 from ipatests.test_integration.base import IntegrationTest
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_installation.py
```diff
@@ -254,6 +254,85 @@ def test_replica_ca_install_with_skip_schema_check(self):
...
+        related : https://pagure.io/certmonger/issue/125
+        """
+        test_service='test/%s'%self.master.hostname
+        pkcs_passwd='Secret123'
+        pin='123456'
+        noise_contents=("sjdhkjhdkjahdkjsahdhkjsahdkjsahdkjHDJSADHSADASDJHAA"
+                        "ADAJSDKJSAHDJSAHDJASHDJHSADKJAHSDKJAHDSKJAHSDKJHASA"
+                        "ASJDHASJDHASJDHJASHDJASHDKJASHDJSAHDJSAHDJSAHDJSWSH")
+        noise_file='/tmp/noise'
+        self.master.master.put_file_contents(noise_file, noise_contents)
```
rcritten: test_ipapython/test_certdb.py has an example of a better way to generate a noise file.

ipatests/test_integration/test_installation.py
```diff
@@ -254,6 +254,85 @@ def test_replica_ca_install_with_skip_schema_check(self):
...
+        cmd_args=['ipa-getcert', 'start-tracking', '-d', paths.NSS_DB_DIR,
+                  '-n', 'test', '-t', 'test', '-P', pin]
+        result = self.master.run_command(cmd_args)
+        request_id = re.findall(r'\d+', result.stdout_text)
+
+        # check if certificate ifs tracked by certmonger
+        cmd_args=['getcert', 'list', '-i', request_id[0]]
+        result = self.master.run_command(cmd_args)
+        assert "MONITORING" in result.stdout_text
+
```
rcritten: This test case is certainly valid because at one point certmonger wasn't able to associate an imported cert and key inside a token. To ensure that the token is usable I'd suggest adding getcert resubmit -i request_id as well to really exercise things.

Note that private key is a exercised by the verification step above because it includes the -e flag but by forcing it to get a new cert we can be sure that this is renewable.

ipatests/test_integration/test_installation.py
```diff
@@ -254,6 +254,93 @@ def test_replica_ca_install_with_skip_schema_check(self):
...
+        tasks.kinit_admin(self.master)
+        tasks.install_dns(self.master)
+        self.master.run_command(['ipa', 'service-add', test_service])
+
+        # create a csr
+        cmd_args = ['certutil', '-d', paths.NSS_DB_DIR, '-R', '-a',
+                    '-o', '/root/ipa.csr',
+                    '-s', "CN=%s" % self.master.hostname]
+        self.master.run_command(cmd_args)
```
rcritten: This will never finish because -z noisefile isn't included in the command-line so it is waiting for keyboard input to randomize the key.

ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-latest/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
+        test_suite: test_integration/test_installation.py::TestInstallCA::test_certmonger_reads_token_HSM
         template: *ci-master-latest
         timeout: 3600
-        topology: *master_1repl_1client
+        topology: *master_2repl_1client
```
rcritten: Are two replicas really necessary here?

ipatests/test_integration/test_installation.py
```diff
@@ -254,6 +255,94 @@ def test_replica_ca_install_with_skip_schema_check(self):
...
+        result = self.master.run_command(cmd_args)
+        assert "MONITORING" in result.stdout_text
+
+        # ensure if key and token are re-usable
+        cmd_args = ['getcert', 'resubmit', '-i', request_id[0]]
+        self.master.run_command(cmd_args)
+
+        cmd_args = ['getcert', 'list', '-i', request_id[0]]
+        result = self.master.run_command(cmd_args)
+        assert "MONITORING" in result.stdout_text
```
rcritten: I think you're going to need a loop here with some sleep and a limited number of loops. The cert should renew pretty quickly but it may take a few seconds. Might be worth looping 300 times with a 1-second sleep (5 min max) waiting for the MONITORING.

ipatests/prci_definitions/temp_commit.yaml
mrizwan93: Yes. The other tests in the class requires 2 replicas.

rcritten:
I think it might be failing because the noisefile is being written on the controller. Perhaps use self.master.put_file_contents() to put the file onto the runner (I just ran into this myself).

rcritten:
The certutil verify is failing though I don't see anything obviously wrong with your invocation. A workaround would be to call self.master.put_file_contents('/root/pinfile', pin) and add '-f', '/root/pinfile' to the certutil invocation.

rcritten:
This needs certmonger-0.79.9 in order to work. I've tested it manually and this will address the issue of the tracking getting stuck in NEWLY_ADDED_NEED_KEYINFO_READ_TOKEN

ipatests/test_integration/test_installation.py
```diff
@@ -254,6 +254,99 @@ def test_replica_ca_install_with_skip_schema_check(self):
...
+        time.sleep(30)
+        cmd_args = ['getcert', 'list', '-i', request_id[0]]
+        result = self.master.run_command(cmd_args)
+        assert "MONITORING" in result.stdout_text
+
+        # ensure if key and token are re-usable
+        cmd_args = ['getcert', 'resubmit', '-i', request_id[0]]
+        self.master.run_command(cmd_args)
+
+        for i in range(300):
```
rcritten: Honestly I'd stick this code into a function and call it both for the start-tracking and the renewal.

This code currently will not raise an error on failure.

rcritten:
This test passes for me with the updated certmonger and this patch:

```
diff --git a/ipatests/test_integration/test_installation.py b/ipatests/test_integration/test_installation.py
index 5d90d6659..120b1fd52 100644
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -25,6 +25,7 @@ from ipaplatform.osinfo import osinfo
 from ipaplatform.tasks import tasks as platformtasks
 from ipapython import ipautil
+from ipalib.install.certmonger import wait_for_request
 from ipatests.pytest_ipa.integration import tasks
 from ipatests.pytest_ipa.integration.env_config import get_global_config
@@ -325,27 +326,21 @@ class TestInstallCA(IntegrationTest):
 
         # add certificate tracking to certmonger
         cmd_args = ['ipa-getcert', 'start-tracking', '-d', paths.NSS_DB_DIR,
-                    '-n', 'test', '-t', 'test', '-P', pin]
+                    '-n', 'test', '-t', 'test', '-P', pin,
+                    '-K', test_service]
         result = self.master.run_command(cmd_args)
         request_id = re.findall(r'\d+', result.stdout_text)
 
         # check if certificate is tracked by certmonger
-        time.sleep(30)
-        cmd_args = ['getcert', 'list', '-i', request_id[0]]
-        result = self.master.run_command(cmd_args)
-        assert "MONITORING" in result.stdout_text
+        status = wait_for_request(request_id[0])
+        assert status == "MONITORING"
 
         # ensure if key and token are re-usable
         cmd_args = ['getcert', 'resubmit', '-i', request_id[0]]
         self.master.run_command(cmd_args)
 
-        for i in range(300):
-            time.sleep(1)
-            cmd_args = ['getcert', 'list', '-i', request_id[0]]
-            result = self.master.run_command(cmd_args)
-            if "MONITORING" in result.stdout_text:
-                break
-            i += 1
+        status = wait_for_request(request_id[0], 300)
+        assert status == "MONITORING"
 
 
 class TestInstallWithCA_KRA1(InstallTestBase1): 
```

rcritten:
The updated certmonger was pushed to the stable repos tonight, not sure if we need to update the image to pick it up or not.

netoarmando:
> The updated certmonger was pushed to the stable repos tonight, not sure if we need to update the image to pick it up or not.

Yes, we need. But for testing this you can add `update_packages: True` to the job definition to force PR-CI to update the packages.

rcritten:
There doesn't seem to be a token named "test". I looked through the journal and didn't see anything unusual other than the missing token. It may be worth adding this to gather the output after creating the token:

softhsm2-util --show-slots

I don't see any issues with the code, clearly the token is being created, but for some reason certmonger can't see it. I also checked for AVCs and nothing jumped out at me there either.

rcritten:
Ok, so clearly the token is there, and is confirmed that NSS can see it, but for some reason it isn't visible to certmonger. Let me try to reproduce locally.

rcritten:
I can't reproduce this locally. Christian had noted some oddness in another aspect of certmonger when it isn't restarted after an IPA installation. I added a patch to do a restart before this to see if it is similarly affected.

mrizwan93:
This is strange, loop is not breaking even the state being `MONITORING`.
`certmonger request is in state %r MONITORING`

rcritten:
The string probably contains a newline. I added a strip() and pushed a new patch.

rcritten:
@mrizwan93 tests are green now. Feel free to rebase my patch into yours if you'd prefer. The patch set needs to be rebased anyway.

rcritten:
Looks good, remove the temp commit and I'll add ack.

rcritten:
master:

* fe21094c8eca948673c6ff0aac337ee3151d4be1 Test if certmonger reads the token in HSM
* 806795422934c3220856bcaaf8810cdebd782fd4 Add certmonger wait_for_request that uses run_command

PR #4178, opened 2020-01-30
"Unspecified GSS failure" should not be in log when directory is mounted.

amore17:
This is an integration test for: https://bugzilla.redhat.com/show_bug.cgi?id=1759665

When there is directory mounted on the ipa-client,
Then no "Unspecified GSS failure" should be there in logs.

### DIFF
--- a/ipatests/test_integration/test_nfs.py
+++ b/ipatests/test_integration/test_nfs.py
@@ -127,6 +127,8 @@ class TestNFS(IntegrationTest):
         nfssrv = self.clients[0]
         nfsclt = self.clients[1]
 
+        # for journalctl --since
+        since = time.strftime('%H:%M:%S')
         nfsclt.run_command(["systemctl", "restart", "rpc-gssd"])
         time.sleep(WAIT_AFTER_INSTALL)
         mountpoints = ("/mnt/krb", "/mnt/std", "/home")
@@ -146,6 +148,11 @@ class TestNFS(IntegrationTest):
             "mount", "-t", "nfs4", "-o", "sec=krb5p,vers=4.0",
             "%s:/exports/home" % nfssrv.hostname, "/home", "-v"
         ])
+        error = "Unspecified GSS failure"
+        check_log = [
+            'journalctl', '-u', 'gssproxy', '--since={}'.format(since)]
+        result = nfsclt.run_command(check_log)
+        assert error not in (result.stdout_text, result.stderr_text)
 
     def test_automount(self):
         """


amore17:
Added bugzilla link only as there is no pagure issue reported for gssproxy.

ssidhaye:
LGTM

ipatests/test_integration/test_nfs.py
```diff
@@ -146,6 +147,19 @@ def test_krb5_nfs_manual_configuration(self):
...
         ])
+        # for journalctl --since
+        since = time.strftime('%H:%M:%S')
+        # mount with krb5
+        nfsclt.run_command([
+            "mount", "-o", "sec=krb5",
+            "%s:/exports/krb5nfs" % nfssrv.hostname, "/mnt/krb5", "-v"
+        ])
+        nfsclt.run_command(['umount', '/mnt/krb5'])
+        error_meassage = "Unspecified GSS failure"
```
fcami: Small typo here, please change to error_message

ipatests/test_integration/test_nfs.py
```diff
@@ -146,6 +147,19 @@ def test_krb5_nfs_manual_configuration(self):
...
+        check_log = [
+            'journalctl', '-u', 'gssproxy', '--since={}'.format(since)]
+        result = nfsclt.run_command(check_log)
+        assert error_meassage not in (result.stdout_text, result.stderr_text)
```
fcami: Same as above

ipatests/test_integration/test_nfs.py
```diff
@@ -106,6 +106,7 @@ def test_krb5_nfsd(self):
         basedir = "exports"
         exports = {
             "krbnfs": "*(sec=krb5p,rw)",
+            "krb5nfs": "*(sec=krb5,rw)",
```
fcami: Are we sure we need another mountpoint?
I don't see why the problem would not exist with the existing krb5p mount.

ipatests/test_integration/test_nfs.py
```diff
@@ -129,7 +130,7 @@ def test_krb5_nfs_manual_configuration(self):
 
         nfsclt.run_command(["systemctl", "restart", "rpc-gssd"])
         time.sleep(WAIT_AFTER_INSTALL)
-        mountpoints = ("/mnt/krb", "/mnt/std", "/home")
+        mountpoints = ("/mnt/krb", "/mnt/std", "/home", "/mnt/krb5")
```
fcami: See the mountpoint command above. I'm not convinced we need another one.

ipatests/test_integration/test_nfs.py
amore17: @fcami Thanks for the review.
It was added as it was in bugzilla description but as it can be reproducible  with krb5p also, so removed another mount-point.

fcami:
LGTM, please remove the temp commit and I'll ACK it.

flo-renaud:
master:

* ab1999deb687f77fcad6c466ab3d472c8417d6ee After mounting "Unspecified GSS failure" should not be in logs.

PR #4179, opened 2020-01-30
ipatests: update packages for rawhide and updates-testing nightlies

flo-renaud:
The nightly tests for rawhide and updates_testing are expected
to set
        update_packages: True
in all the job definitions to make sure that dnf/yum update is called
before starting the tests.

This tag was missing for some jobs, this commit fixes the issue.


flo-renaud:
master:

* 60f746d9983b54e71f9ba98632f30e79bdead86e ipatests: update packages for rawhide and updates-testing nightlies

PR #4181, opened 2020-01-31
Added a test to check if ipa host-find --pkey-only does not return SSH public key

ssidhaye:
Added a test to check if ipa host-find --pkey-only 
does not return SSH public key

It check if 'SSH public key fingerprint' is
not present in the output of the command

Pagure Link : https://pagure.io/freeipa/issue/8029

### DIFF
--- a/ipatests/test_xmlrpc/test_host_plugin.py
+++ b/ipatests/test_xmlrpc/test_host_plugin.py
@@ -132,6 +132,12 @@ def host4(request):
 
 
 @pytest.fixture(scope='class')
+def host5(request):
+    tracker = HostTracker(name=u'testhost5')
+    return tracker.make_fixture(request)
+
+
+@pytest.fixture(scope='class')
 def lab_host(request):
     name = u'testhost1'
     tracker = HostTracker(name=name,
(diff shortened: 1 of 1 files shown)


ipatests/test_xmlrpc/test_host_plugin.py
```diff
@@ -261,6 +261,13 @@ def test_update_simple(self, host):
         res = command()['result']
         assert len(res) == 1
 
+        # test host-find with --pkey-only
+        command = host.make_find_command(
+            fqdn=host.fqdn, pkey_only=True)
```
flo-renaud: I tried to run the test on a system without the patch, and the test does not detect the failure. 
1- the host entry does not contain any ssh key, you should rather test with the hostname of the master
2- in the output, you need to check if there is a result[u'result'][u'ipasshpubkey'] value.

ssidhaye: I have made the changes as per review comments

flo-renaud:
Hi @ssidhaye 
thanks for the PR, LGTM. I tested with an older version that does not have the fix and the test properly detected the issue.

flo-renaud:
Adding ipa-4-6+ labels as the issue was fixed in all the branches.

flo-renaud:
master:

* f0f2c2645ef5aa0a065e2b5a2063d57d85f74d56 Added a test to check if ipa host-find --pkey-only does not return SSH public key

ssidhaye:
Thanks @flo-renaud !

PR #4182, opened 2020-01-31
Test ipa-join using different options

jayesh-garg:
test ipa-join on client with passing differnt options
like unenroll,hostname,bindpw.

Signed-off-by: Jayesh Garg <jgarg@redhat.com>

### DIFF
--- a/ipatests/prci_definitions/nightly_latest.yaml
+++ b/ipatests/prci_definitions/nightly_latest.yaml
@@ -495,6 +495,18 @@ jobs:
         timeout: 10800
         topology: *master_2repl_1client
 
+  fedora-latest/test_installation_TestIpaJoin:
+    requires: [fedora-latest/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-latest/build_url}'
+        test_suite: test_integration/test_installation.py::TestIpaJoin
+        template: *ci-master-latest
+        timeout: 10800
+        topology: *ad_master_2client
+
   fedora-latest/test_idviews:
     requires: [fedora-latest/build]
     priority: 50
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -1075,3 +1075,60 @@ class TestInstallReplicaAgainstSpecificServer(IntegrationTest):
                                             self.replicas[0].hostname],
                                            stdin_text=dirman_password)
         assert self.replicas[0].hostname not in cmd.stdout_text
+
+
+class TestIpaJoin(IntegrationTest):
+
+    num_clients = 2
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master, setup_dns=False)
+
+    def test_unenroll_option(self):
+        """
+        Test ipa-join Using unenroll option
+        """
+        # install client and kinit admin
+        tasks.install_client(self.master, self.clients[0])
+        tasks.kinit_admin(self.clients[0])
+
+        # checking unenroll option
+        result = self.clients[0].run_command(['ipa-join', '-u',
+                                             self.clients[0].hostname])
+        assert "Unenrollment successful." in result.stderr_text
+
+        # checking hostname option
+        result = self.clients[0].run_command(['ipa-join', '-h',
+                                              self.clients[0].hostname])
+        assert "Keytab successfully retrieved" in result.stderr_text
+
+    def test_hostname_with_ldap_password(self):
+        """
+        Test ipa-join Using options hostname and bindpw
+        """
+        # install client and kinit admin
+        tasks.install_client(self.master, self.clients[1])
+        tasks.kinit_admin(self.clients[1])
+
+        # Using options hostname and bindpw with invalid pwd
+        invalid_password = 'abcdefgh'
+        result = self.clients[1].run_command(['ipa-join',
+                                              '-h', self.clients[1].hostname,
+                                              '-w', invalid_password],
+                                             raiseonerr=False)
+        assert "Bind failed:" in result.stderr_text
+
+        # Using options hostname and bindpw with valid pwd
+        valid_password = ''
+        self.master.run_command(['ipa', 'host-del', self.clients[1].hostname])
+        cmd = self.master.run_command(['ipa', 'host-add', '--random'],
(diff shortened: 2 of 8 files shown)


ipatests/prci_definitions/nightly_latest.yaml
```diff
@@ -495,6 +495,18 @@ jobs:
...
+    requires: [fedora-latest/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-latest/build_url}'
+        test_suite: test_integration/test_installation.py::TestIpaJoin
+        template: *ci-master-latest
+        timeout: 10800
+        topology: *master_2client
```
mrizwan93: We don't have such topology. Please use `master_3client`.

ipatests/prci_definitions/nightly_latest.yaml
```diff
@@ -495,6 +495,18 @@ jobs:
...
+    requires: [fedora-latest/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-latest/build_url}'
+        test_suite: test_integration/test_installation.py::TestIpaJoin
+        template: *ci-master-latest
+        timeout: 10800
+        topology: *ad_master_2client
```
mrizwan93: This test does not involve AD. Please change the topology.

ipatests/test_integration/test_installation.py
```diff
@@ -1075,3 +1075,60 @@ def test_replica_install_against_server(self):
                                             self.replicas[0].hostname],
                                            stdin_text=dirman_password)
         assert self.replicas[0].hostname not in cmd.stdout_text
+
+
+class TestIpaJoin(IntegrationTest):
+
+    num_clients = 2
```
mrizwan93: Can we not use single client here? 
First un-enrolling the client then enrolling the same with bad password and correct password?

ipatests/test_integration/test_installation.py
```diff
@@ -1075,3 +1075,60 @@ def test_replica_install_against_server(self):
...
+        # Using options hostname and bindpw with valid pwd
+        valid_password = ''
+        self.master.run_command(['ipa', 'host-del', self.clients[1].hostname])
+        cmd = self.master.run_command(['ipa', 'host-add', '--random'],
+                                      stdin_text=self.clients[1].hostname)
+        for line in cmd.stdout_text.split('\n'):
+            if "Random password" in line:
+                valid_password = line.split()[2]
+                break
+        result = self.clients[0].run_command(['ipa-join',
```
mrizwan93: `self.clients[0]` should be `self.clients[1]`.  Please consider use of single client.

ipatests/test_integration/test_installation.py
```diff
@@ -1075,3 +1075,60 @@ def test_replica_install_against_server(self):
...
+        Test ipa-join Using unenroll option
+        """
+        # install client and kinit admin
+        tasks.install_client(self.master, self.clients[0])
+        tasks.kinit_admin(self.clients[0])
+
+        # checking unenroll option
+        result = self.clients[0].run_command(['ipa-join', '-u',
+                                             self.clients[0].hostname])
+        assert "Unenrollment successful." in result.stderr_text
```
rcritten: This verifies that the command returns success, not that it did what it said it did.

ipa-join -u does a kinit using the host principal then calls the ipa host-disable <fqdn> command. You could, for example, do a host-show to ensure that it has no kerberos keys. To really stretch it any services on the host should also be disabled.

It is probably worth calling join -u twice, expecting success the first and failure the second..

ipatests/test_integration/test_installation.py
```diff
@@ -1075,3 +1075,60 @@ def test_replica_install_against_server(self):
...
+
+        # checking hostname option
+        result = self.clients[0].run_command(['ipa-join', '-h',
+                                              self.clients[0].hostname])
+        assert "Keytab successfully retrieved" in result.stderr_text
```
rcritten: The keytab should be verified. Perhaps this would be sufficient:

self.clients[0].run_command(['kinit', '-kt', '/etc/krb5.keytab'])
self.clients[0].run_command(['ipa', 'host-show', self.clients[0].hostname)
# ensure that has_keytab is Yes.

ipatests/test_integration/test_installation.py
```diff
@@ -1075,3 +1075,60 @@ def test_replica_install_against_server(self):
...
+        tasks.install_client(self.master, self.clients[1])
+        tasks.kinit_admin(self.clients[1])
+
+        # Using options hostname and bindpw with invalid pwd
+        invalid_password = 'abcdefgh'
+        result = self.clients[1].run_command(['ipa-join',
+                                              '-h', self.clients[1].hostname,
+                                              '-w', invalid_password],
+                                             raiseonerr=False)
+        assert "Bind failed:" in result.stderr_text
```
rcritten: I'd call ipa-join -u first to un-enroll the client like you did with the previous test.

You only use invalid_password once AFAICT so there is no need to have it in a variable IMHO.

ipatests/test_integration/test_installation.py
```diff
@@ -1075,3 +1075,60 @@ def test_replica_install_against_server(self):
...
+
+        # Using options hostname and bindpw with valid pwd
+        valid_password = ''
+        self.master.run_command(['ipa', 'host-del', self.clients[1].hostname])
+        cmd = self.master.run_command(['ipa', 'host-add', '--random'],
+                                      stdin_text=self.clients[1].hostname)
```
rcritten: This would be simpler if you set a fixed password instead of using a random one. You are just testing that ipa-join works with a password, not that random passwords work.

ipatests/test_integration/test_installation.py
```diff
@@ -1075,3 +1075,60 @@ def test_replica_install_against_server(self):
...
+        for line in cmd.stdout_text.split('\n'):
+            if "Random password" in line:
+                valid_password = line.split()[2]
+                break
+        result = self.clients[0].run_command(['ipa-join',
+                                              '-h', self.clients[1].hostname,
+                                              '-w', valid_password])
+        assert "Keytab successfully retrieved" in result.stderr_text
```
rcritten: I'd add similar keytab validation code as above, in a separate function would be best I suppose.

PR #4189, opened 2020-02-02
WebUI tests: Fix 'Button is not displayed' exception

serg-cymbaluk:
Add a small timeout (up to 1 second) which allows to prevent exceptions when
WebDriver attempts to click a button before it is rendered.

Ticket: https://pagure.io/freeipa/issue/8169

### DIFF
--- a/ipatests/test_webui/ui_driver.py
+++ b/ipatests/test_webui/ui_driver.py
@@ -731,9 +731,16 @@ class UI_driver:
 
     def _button_click(self, selector, parent, name=''):
         btn = self.find(selector, By.CSS_SELECTOR, parent, strict=True)
+
+        # The small timeout (up to 5 seconds) allows to prevent exceptions when
+        # driver attempts to click a button before it is rendered.
+        WebDriverWait(self.driver, 5, 0.2).until(
+            lambda d: btn.is_displayed(),
+            'Button is not displayed: %s' % (name or selector)
+        )
         self.move_to_element_in_page(btn)
+
         disabled = btn.get_attribute("disabled")
-        assert btn.is_displayed(), 'Button is not displayed: %s' % name
         assert not disabled, 'Invalid button state: disabled. Button: %s' % name
         btn.click()
         self.wait_for_request()


ipatests/test_webui/ui_driver.py
```diff
@@ -732,8 +732,15 @@ def button_click(self, name, parent=None,
     def _button_click(self, selector, parent, name=''):
         btn = self.find(selector, By.CSS_SELECTOR, parent, strict=True)
         self.move_to_element_in_page(btn)
+
+        # The small timeout (up to 1 second) allows to prevent exceptions when
+        # driver attempts to click a button before it is rendered.
+        WebDriverWait(self.driver, 1, 0.2).until(
```
wladich: I suggest to increase max wait time to 5 seconds - we execute tests in VMs on heavily loaded hosts and I feel that browser can get stuck for second or two.

serg-cymbaluk: Yes, agree, fixed.

ipatests/test_webui/ui_driver.py
```diff
@@ -732,8 +732,15 @@ def button_click(self, name, parent=None,
     def _button_click(self, selector, parent, name=''):
         btn = self.find(selector, By.CSS_SELECTOR, parent, strict=True)
         self.move_to_element_in_page(btn)
```
wladich: I see an issue here: `scrollIntoView` does not have any effect when called for hidden element. So this line should be moved below WebDriverWait invocation

ipatests/test_webui/ui_driver.py
```diff
@@ -732,8 +732,15 @@ def button_click(self, name, parent=None,
...
+
+        # The small timeout (up to 5 seconds) allows to prevent exceptions when
+        # driver attempts to click a button before it is rendered.
+        WebDriverWait(self.driver, 5, 0.2).until(
+            lambda d: btn.is_displayed(),
+            'Button is not displayed: %s' % name
```
wladich: As the name can be empty, I suggest using selector instead

ipatests/test_webui/ui_driver.py
serg-cymbaluk: Fixed, waiting for test results...

ipatests/test_webui/ui_driver.py
serg-cymbaluk: Fixed.

wladich:
LGTM.
Please remove temp commit and I will ACK the PR

wladich:
ACK

serg-cymbaluk:
master:

* 9418042ee700b8c82f289ce649ab2b80c60d7822 WebUI tests: Fix 'Button is not displayed' exception

PR #4190, opened 2020-02-03
Removed hard-coded default profile subsystem class name

edewata:
Previously in order to enable the LDAP profile subsystem
the ca_enable_ldap_profile_subsystem() would check the
current value of the profile subsystem class parameter in
CS.cfg. If the parameter was still set to the default value
(i.e. ProfileSubsystem), the code would change it to
LDAPProfileSubsystem.

There is a effort in PKI to clean up the profile subsystem
classes which may require changing the default value for
this parameter. However, this improvement is blocked since
the ca_enable_ldap_profile_subsystem() is implicitly assuming
that the default value will always be ProfileSubsystem.

This patch updated the code such that instead of checking
for a specific value that needs to be changed, it will check
whether it has the desired value already. This mechanism
will reduce potential conflicts with future PKI improvements.

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -386,15 +386,34 @@ def ca_enable_ldap_profile_subsystem(ca):
     needs_update = False
     directive = None
     try:
-        for i in range(15):
+        i = 0
+        while True:
+            # find profile subsystem
+            directive = "subsystem.{}.id".format(i)
+            value = directivesetter.get_directive(
+                paths.CA_CS_CFG_PATH,
+                directive,
+                separator='=')
+            if not value:
+                logger.error('Unable to find profile subsystem in %s',
+                             paths.CA_CS_CFG_PATH)
+                return False
+            if value != 'profile':
+                i = i + 1
+                continue
+
+            # check profile subsystem class name
             directive = "subsystem.{}.class".format(i)
             value = directivesetter.get_directive(
                 paths.CA_CS_CFG_PATH,
                 directive,
                 separator='=')
-            if value == 'com.netscape.cmscore.profile.ProfileSubsystem':
+            if value != 'com.netscape.cmscore.profile.LDAPProfileSubsystem':
                 needs_update = True
-                break
+
+            # break after finding profile subsystem
+            break
+
     except OSError as e:
         logger.error('Cannot read CA configuration file "%s": %s',
                      paths.CA_CS_CFG_PATH, e)


abbra:
Linter failed:
```
************* Module ipaserver.install.server.upgrade
ipaserver/install/server/upgrade.py:22: [E0611(no-name-in-module), ] No name 'server' in module 'pki')
ipaserver/install/server/upgrade.py:22: [E0401(import-error), ] Unable to import 'pki.server.instance')
ipaserver/install/server/upgrade.py:387: [E1101(no-member), ca_enable_ldap_profile_subsystem] Module 'pki' has no 'server' member)
ipaserver/install/server/upgrade.py:408: [W0715(raising-format-tuple), ca_enable_ldap_profile_subsystem] Exception arguments suggest string formatting might be intended)
ipaserver/install/server/upgrade.py:423: [E0602(undefined-variable), ca_enable_ldap_profile_subsystem] Undefined variable 'needs_update')

```

Please run `make lint` locally before submitting an update

edewata:
I tried to follow [the build doc](https://github.com/freeipa/freeipa/blob/master/BUILD.txt#L18) but I'm getting this error:
```
$ make lint
CDPATH="${ZSH_VERSION+.}:" && cd . && /bin/sh /home/edewata/Projects/freeipa/missing aclocal-1.15 -I m4
/home/edewata/Projects/freeipa/missing: line 81: aclocal-1.15: command not found
WARNING: 'aclocal-1.15' is missing on your system.
         You should only need it if you modified 'acinclude.m4' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'aclocal' program is part of the GNU Automake package:
         <http://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <http://www.gnu.org/software/autoconf>
         <http://www.gnu.org/software/m4/>
         <http://www.perl.org/>
make: *** [Makefile:554: aclocal.m4] Error 127
```
Is the doc up to date?

edewata:
This PR requires PKI 10.8.0 which has not been released on Fedora. Hopefully that can be done soon.

ipaserver/install/server/upgrade.py
```diff
@@ -383,31 +384,42 @@ def ca_enable_ldap_profile_subsystem(ca):
...
     needs_update = False
-    directive = None
-    try:
-        for i in range(15):
-            directive = "subsystem.{}.class".format(i)
-            value = directivesetter.get_directive(
-                paths.CA_CS_CFG_PATH,
-                directive,
-                separator='=')
-            if value == 'com.netscape.cmscore.profile.ProfileSubsystem':
```
frasertweedale: This can be accomplished with a much smaller change - a one-liner in fact, and it will not depend on a newer version of pki-core.  Just change this comparision from
```python
             if value == 'com.netscape.cmscore.profile.ProfileSubsystem':
```
to
```python
             if value != 'com.netscape.cmscore.profile.LDAPProfileSubsystem':
```

Is there any reason not to keep it simple like this?

edewata: That might only work if the profile subsystem is the first one in the list. If the first subsystem is not a profile subsystem, the code will set `needs_update` to `True` immediately and break from the loop.

I think we need to make sure that `subsystem.<n>.id=profile` first before comparing the class name. That probably can be done without importing `pki.server` module. What do you think?

edewata: Also the code imposes an artificial limit on the number of subsystems it's checking:
```
for i in range(15):
```
While it works now, it could break under certain circumstances.

abbra:
@edewata your issue work aclocal looks like an environment not initialized. Run autogen.sh first.
However, we need a code that works for both of and new pki version. Always depending on a new one is not acceptable. I'm fact, FreeIPA PR CI currently runs on f30, so updating pki in rawhide wouldn't help. We need to handle all variants at the same time. Please consider this while refactoring.

ipaserver/install/server/upgrade.py
frasertweedale: Of course, you are right.  But just add a couple more lines to first check if  `subsystem.<n>.id == 'profile'`.  So it is closer to a 10 line patch than a 1-liner :)

I accept your point about the limit of 15.  Well, maybe we just increase it (to 100, 1000, whatever), and ensure that we break after we have encountered the profile subsystem configuration.  I accept it is not perfect but IMO the risk is so low it hardly seems worth a more complicated patch that depends on unreleased Dogtag version.

edewata:
OK, `make lint` works now. I've updated the patch to no longer depend on PKI API, and I also tried to replace the `range(15)` limitation. Please have a look, thanks!

In the future it's preferred to manage the PKI instance through PKI API or tools instead of direct modifications. That will ensure the changes to the PKI instance are well defined and supported. It will hide implementation details in PKI from IPA, and it will also help the PKI team see how the PKI instance is set up and used in IPA in case there are issues.

frasertweedale:
master:

* edfe95b120edd4794c7d1e1290c99116fac3b194 (HEAD) Removed hard-coded default profile subsystem class name

PR #4192, opened 2020-02-03
WebUI tests: Fix broken reference to parent facet in table record check

serg-cymbaluk:
Add decorator to has_record method which repeats the check when an active facet is changed (catch StaleElementReferenceException).

### DIFF
--- a/ipatests/test_webui/ui_driver.py
+++ b/ipatests/test_webui/ui_driver.py
@@ -138,6 +138,23 @@ def dismiss_unexpected_alert(fn):
     return wrapped
 
 
+def repeat_on_stale_parent_reference(fn):
+    """
+    The decorator repeats a function once when StaleElementReferenceException
+    is caught.
+    It is not applicable if a parent reference is created outside a function.
+    """
+    @wraps(fn)
+    def wrapped(*args, **kwargs):
+        if ('parent' in kwargs) and kwargs['parent'] is None:
+            try:
+                return fn(*args, **kwargs)
+            except StaleElementReferenceException:
+                pass
+        return fn(*args, **kwargs)
+    return wrapped
+
+
 class UI_driver:
     """
     Base class for all UI integration tests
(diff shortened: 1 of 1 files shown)


serg-cymbaluk:
> This was reported as an error affecting rawhide, did you execute this on rawhide too?

Let me check it.

serg-cymbaluk:
@netoarmando 

> Why repeating the function call solves the issue?

If a parent reference is broken, it means that the parent was a previous facet and it is unavailable after navigation, or it was a closed modal form. In these both variants **has_record** method calls `parent = self.get_form()` and the parent reference overwrites with an actual one.

Repetition of the error is impossible because it is caused by previous action.

1. navigation action (old reference is actual)
2. has_record
   2.1. parent = get_form (reference is still valid because of some browser delay)
   2.2. find in parent (reference is gone -> StaleElementReferenceException)
3. repeat has_record
   3.1. parent = get_form (reference is actual since there were no navigation actions between p1 and p3.1)
   3.2. find in parent (return an element)

ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -61,10 +61,36 @@ jobs:
...
+        timeout: 1800
+        topology: *build
+
+  fedora-rawhide/temp_commit:
+    requires: [fedora-rawhide/build]
+    priority: 50
+    job:
+      class: RunWebuiTests
+      args:
+        build_url: '{fedora-frawhide/build_url}'
```
netoarmando: Remove the `f`:
```
build_url: '{fedora-frawhide/build_url}
--------------------^
```
Also, you need to rebase because of https://github.com/freeipa/freeipa/pull/4108/files#diff-46f33c2694f0a52173ef4722db670b48

serg-cymbaluk:
master:

* a4634a59c98ae922708f9211786b67505dada736 WebUI tests: Fix broken reference to parent facet in table record check

PR #4193, opened 2020-02-03
[ACME] Add ipa-ca.$DOMAIN alias to IPA server HTTP certificates

frasertweedale:
First PR for the ACME effort.  This is needed so ACME clients can reach IPA
ACME service via the ``ipa-ca.$DOMAIN`` DNS name (ACME requires TLS).

This change is also reasonable, independent of the ACME effort.

https://pagure.io/freeipa/issue/8186

```
83a5a3aa6 (Fraser Tweedale, 11 hours ago)
   upgrade: add ipa-ca.$DOMAIN alias to HTTP certificate

   If the HTTP certificate does not have the ipa-ca.$DOMAIN dNSName, resubmit
   the certificate request to add the name.  This action is performed after
   the tracking request has already been updated.

   Note: due to https://pagure.io/certmonger/issue/143 the resubmitted 
   request, if it does not immediately succeed and if the notAfter date of the
   current certificate is still far off, the request could get stuck in state
   CA_UNREACHABLE until a Certmonger restart.  There is not much we can do
   about that in the middle of ipa-server-upgrade.

efe071539 (Fraser Tweedale, 12 hours ago)
   httpinstance: add ipa-ca.$DOMAIN alias in initial request

   For new server/replica installation, issue the HTTP server certificate with
   the 'ipa-ca.$DOMAIN' SAN dNSName.  This is accomplished by adding the name
   to the Certmonger tracking request.

feea49420 (Fraser Tweedale, 3 days ago)
   cert-request: allow ipa-ca.$DOMAIN dNSName for IPA servers

   ACME support requires TLS and we want ACME clients to access the service
   via the ipa-ca.$DOMAIN DNS name.  So we need to add the ipa-ca.$DOMAIN
   dNSName to IPA servers' HTTP certificates.  To facilitiate this, add a
   special case to the cert-request command processing.  The rule is:

   - if the dnsName being validated is "ipa-ca.$DOMAIN"
   - and the subject principal is an "HTTP/..." service
   - and the subject principal's hostname is an IPA server

   Then that name (i.e. "ipa-ca.$DOMAIN") is immediately allowed. Otherwise
   continue with the usual dnsName validation.

62129a44a (Fraser Tweedale, 3 days ago)
   httpinstance: add fqdn and ipa-ca alias to Certmonger request

   When (re-)tracking the HTTP certificate, explicitly add the server FQDN and
   ipa-ca.$DOMAIN DNS names to the Certmonger tracking request.

fe3489cf4 (Fraser Tweedale, 4 days ago)
   certmonger: support dnsname as request search criterion

   We need to be able to filter Certmonger tracking requests by the DNS names
   defined for the request.  The goal is to add the
   'ipa-ca.$DOMAIN' alias to the HTTP certificate tracking requests, so we
   will use that name as a search criterion.  Implement support for this.

   As a result of this commit it will be easy to add support for subset match
   of other Certmonger request list properties.  Just add the property name to
   the ARRAY_PROPERTIES list (and update the
   'criteria' description in the module docstring!)

ea6d31bdf (Fraser Tweedale, 4 days ago)
   certmonger: move 'criteria' description to module docstring

   The 'criteria' parameter is used by several subroutines in the 
   ipalib.install.certmonger module.  It has incomplete documentation spread
   across several of these subroutines.  Move the documentation to the module
   docstring and reference it where appropriate.

aa7b88ad6 (Fraser Tweedale, 4 days ago)
   certmonger: avoid mutable default argument

   certmonger._get_requests has a mutable default argument.  Although at the
   present time it is never modified, this is an antipattern to be avoided.

```

### DIFF
--- a/ipalib/install/certmonger.py
+++ b/ipalib/install/certmonger.py
@@ -479,7 +479,7 @@ def request_cert(
 def start_tracking(
         certpath, ca='IPA', nickname=None, pin=None, pinfile=None,
         pre_command=None, post_command=None, profile=None, storage="NSSDB",
-        token_name=None):
+        token_name=None, dns=None):
     """
     Tell certmonger to track the given certificate in either a file or an NSS
     database. The certificate access can be protected by a password_file.
@@ -514,6 +514,8 @@ def start_tracking(
         Which certificate profile should be used.
     :param token_name:
         Hardware token name for HSM support
+    :param dns:
+        List of DNS names
     :returns: certificate tracking nickname.
     """
     if storage == 'FILE':
@@ -558,6 +560,8 @@ def start_tracking(
         # only pass token names for external tokens (e.g. HSM)
         params['key-token'] = token_name
         params['cert-token'] = token_name
+    if dns is not None and len(dns) > 0:
+        params['DNS'] = dns
 
     result = cm.obj_if.add_request(params)
     try:
--- a/ipalib/install/certmonger.py
+++ b/ipalib/install/certmonger.py
@@ -47,6 +47,29 @@ DBUS_CM_CA_IF = 'org.fedorahosted.certmonger.ca'
 DBUS_PROPERTY_IF = 'org.freedesktop.DBus.Properties'
 
 
+"""
+Certmonger helper routines.
+
+Search criteria
+---------------
+
+Functions that look up requests take a ``dict`` of search criteria.
+In general, the key is a name of a property in the request property
+interface.  But there are some special cases with different
+behaviour:
+
+``nickname``
+  a.k.a. "request ID".  If given, only the specified request is
+  retrieved (if it exists), and it is still tested against other
+  criteria.
+
+``ca-name``
+  Test equality against the nickname of the CA (a.k.a. request
+  helper) object for the request.
+
+"""
+
+
 class _cm_dbus_object:
     """
     Auxiliary class for convenient DBus object handling.
@@ -159,6 +182,9 @@ class _certmonger(_cm_dbus_object):
 def _get_requests(criteria):
     """
     Get all requests that matches the provided criteria.
+
+    :param criteria: dict of criteria; see module doc for details
+
     """
     if not isinstance(criteria, dict):
         raise TypeError('"criteria" must be dict.')
@@ -197,11 +223,11 @@ def _get_requests(criteria):
 
 def _get_request(criteria):
     """
-    Find request that matches criteria.
-    If 'nickname' is specified other criteria are ignored because 'nickname'
-    uniquely identify single request.
-    When multiple or none request matches specified criteria RuntimeError is
-    raised.
+    Find request that matches criteria.  Return ``None`` if no match.
+    Raise ``RuntimeError`` if there is more than one matching request.
+
+    :param criteria: dict of criteria; see module doc for details
+
     """
     requests = _get_requests(criteria)
     if len(requests) == 0:
@@ -239,11 +265,11 @@ def get_request_id(criteria):
     If you don't know the certmonger request_id then try to find it by looking
     through all the requests.
 
-    criteria is a tuple of key/value to search for. The more specific
-    the better. An error is raised if multiple request_ids are returned for
-    the same criteria.
+    Return ``None`` if no match.  Raise ``RuntimeError`` if there is
+    more than one matching request.
 
-    None is returned if none of the criteria match.
+    :param criteria: dict of criteria; see module doc for details
+
     """
     try:
         request = _get_request(criteria)
--- a/ipalib/install/certmonger.py
+++ b/ipalib/install/certmonger.py
@@ -46,6 +46,10 @@ DBUS_CM_REQUEST_IF = 'org.fedorahosted.certmonger.request'
 DBUS_CM_CA_IF = 'org.fedorahosted.certmonger.ca'
 DBUS_PROPERTY_IF = 'org.freedesktop.DBus.Properties'
 
+# These properties, if encountered in search criteria, result in a
+# subset test instead of equality test.
+ARRAY_PROPERTIES = ['template-hostname']
+
 
 """
 Certmonger helper routines.
@@ -66,6 +70,10 @@ behaviour:
 ``ca-name``
   Test equality against the nickname of the CA (a.k.a. request
   helper) object for the request.
+
+``template-hostname``
+  Must be an iterable of DNS names.  Tests that the given values
+  are a subset of the values defined on the Certmonger request.
 
 """
 
@@ -210,11 +218,18 @@ def _get_requests(criteria):
                                        criteria.get('ca-name'))
                 ca = _cm_dbus_object(cm.bus, cm, ca_path, DBUS_CM_CA_IF,
                                      DBUS_CM_IF)
-                value = ca.obj_if.get_nickname()
+                if criteria[criterion] != ca.obj_if.get_nickname():
+                    break
+            elif criterion in ARRAY_PROPERTIES:
+                # perform subset test
+                expect = set(criteria[criterion])
+                got = request.prop_if.Get(DBUS_CM_REQUEST_IF, criterion)
+                if not expect.issubset(got):
+                    break
             else:
                 value = request.prop_if.Get(DBUS_CM_REQUEST_IF, criterion)
-            if value != criteria[criterion]:
-                break
+                if criteria[criterion] != value:
+                    break
         else:
             requests.append(request)
 
--- a/ipalib/install/certmonger.py
+++ b/ipalib/install/certmonger.py
@@ -156,7 +156,7 @@ class _certmonger(_cm_dbus_object):
                                           DBUS_CM_IF)
 
 
-def _get_requests(criteria=dict()):
+def _get_requests(criteria):
     """
     Get all requests that matches the provided criteria.
     """
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -44,7 +44,7 @@ from ipapython.dn import DN
 import ipapython.errors
 from ipaserver.install import sysupgrade
 from ipalib import api, x509
-from ipalib.constants import IPAAPI_USER, MOD_SSL_VERIFY_DEPTH
+from ipalib.constants import IPAAPI_USER, MOD_SSL_VERIFY_DEPTH, IPA_CA_RECORD
 from ipaplatform.constants import constants
 from ipaplatform.tasks import tasks
 from ipaplatform.paths import paths
@@ -593,6 +593,7 @@ class HTTPInstance(service.Service):
                 post_command='restart_httpd', storage='FILE',
                 profile=dogtag.DEFAULT_PROFILE,
                 pinfile=key_passwd_file,
+                dns=[self.fqdn, f'{IPA_CA_RECORD}.{api.env.domain}'],
             )
             subject = str(DN(cert.subject))
             certmonger.add_principal(request_id, self.principal)
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -381,7 +381,7 @@ class HTTPInstance(service.Service):
                     subject=str(DN(('CN', self.fqdn), self.subject_base)),
                     ca='IPA',
                     profile=dogtag.DEFAULT_PROFILE,
-                    dns=[self.fqdn],
+                    dns=[self.fqdn, f'{IPA_CA_RECORD}.{api.env.domain}'],
                     post_command='restart_httpd',
                     storage='FILE',
                     passwd_fname=key_passwd_file,
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -13,6 +13,7 @@ import glob
 import shutil
 import pwd
 import fileinput
+import ssl
 import stat
 import sys
 import tempfile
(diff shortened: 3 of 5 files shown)


frasertweedale:
slapping a WIP label on this until CI goes green and initial feedback is received.

ipalib/install/certmonger.py
```diff
@@ -156,10 +156,11 @@ def __init__(self):
                                           DBUS_CM_IF)
 
 
-def _get_requests(criteria=dict()):
+def _get_requests(criteria=None):
```
rcritten: I guess why have a default at all? Why not just _get_request(criteria) and enforce that criteria is dict?

Since you're poking at it probably worth documenting that:

:param criteria: dict of the search criteria

ipalib/install/certmonger.py
```diff
@@ -198,11 +224,11 @@ def _get_requests(criteria=None):
...
     """
-    Find request that matches criteria.
-    If 'nickname' is specified other criteria are ignored because 'nickname'
-    uniquely identify single request.
-    When multiple or none request matches specified criteria RuntimeError is
-    raised.
+    Find request that matches criteria.  Return ``None`` if no match.
+    Raise ``RuntimeError`` if there is more than one matching request.
+
+    See module doc for description of critieria parameter.
```
rcritten: nit: critieria -> criteria

ipaserver/install/server/upgrade.py
```diff
@@ -2183,6 +2245,10 @@ def upgrade_configuration():
     setup_pkinit(krb)
     enable_server_snippet()
 
+    # Must be executed after certificate_renewal_update
+    # (see function docstring for details)
+    http_certificate_ensure_ipa_ca_dnsname(http)
```
rcritten: The return value is ignored.

frasertweedale: This was a vestige of an earlier attempt to get around the certmonger 7-day delay thing.  e.g. ipa-server-install could restart certmonger if the request was resubmitted.  I decided not to go down that path because it could not be made reliable and in fact, in my estimation represents a bigger risk of something very bad happening.  But I left the return True/False in there.

I removed it in the next update.

ipalib/install/certmonger.py
frasertweedale: Yep, i just removed the default arg, because it is always called with arg.

rcritten:
My general issue with ipa-ca as a SAN is that it implies that  the CA endpoints should/can be accessed over HTTPS but for both CRL and OCSP this is bad if not impossible right (e.g. validate the connected web server cert over OCSP by contacting...the web server). So maybe a different name?

Or update the commit message. It says what the patch is doing but not why.

Is ACME to not be confused with LE? Should those users requesting certs over LE also include ipa-ca (or something) as a SAN?

frasertweedale:
@rcritten good questions.

For OCSP, the AIA extension will always have `http://` URL, so clients will not use HTTPS.
So operationally there is no issue.

For CRL, there is no issue, because it is not an "inline" check.  You are just retrieving the "next" version of CRL.  And you can continue to use non-TLS HTTP for that.  And indeed, as with OCSP, this is what is indicated in the certificate extensions (CRL Distribution Point in case of CRL, Authority Info Access (AIA) for OCSP).

The commit message for feea49420 does explain that we are doing this because ACME requires TLS, and that we want ACME clients to use a single DNS name across the deployment.  All the  commit messages refer to the Pagure ticket https://pagure.io/freeipa/issue/8186.  What else do you want me to explain?

I was keen to use `ipa-ca.$DOMAIN` because it is already defined (internal DNS) or required (external DNS) for IPA CA servers.  I don't think it is a problem to reuse it for ACME.  In fact I think it is preferable to reuse it, compared to introducing another name that we, or administrator in case of external DNS, must manage.

I am interested to hear other perspectives.  @abbra @flo-renaud any thoughts?

frasertweedale:
@rcritten indeed ACME is not to be confused with LE.  This change is all about providing an ACME CA in IPA.

But when it comes to 3rd party certs in general, yes, it is something I overlooked; I need to add a sanity check for 3rd-party-issued IPA server HTTP certificates on CA-ful deployments, to ensure that they have the `ipa-ca.$DOMAIN` or at least warn if they do not.  (Lack of ipa-ca dnsName is acceptable and warranted for CA-less deployments, of course).  I will file another ticket for that.

rcritten:
I still don't think there is enough in the commit message explaining why ipa-ca is being overridden. The commit reads in part "we want ACME clients to access the
service via the ipa-ca.$DOMAIN DNS name." WHY do we want to use the ipa-ca alias and not something like acme? Because it's already there? Fine, but please add that reasoning to the commit, the reasoning in your response for example: it's already there, additional maintenance, etc. The upstream ticket has none of this.

Note that currently only masters that are a CA are in the ipa-ca namespace. This is going to add aliases to hosts that are not CAs nor are in DNS which means they won't be found anyway.

My concern about OCSP/CRL is that currently an HTTPS request to ipa-ca will fail with something like: no alternative certificate subject name matches target host name 'ipa-ca.example.test'

And now that will succeed and has the potential to introduce confusion. We get a ticket every other year or so asking to switch OCSP to https.

frasertweedale:
@rcritten ok, fair cop, I'll add some more explanation to the commit message(s).

frasertweedale:
@rcritten I added detailed commentary to commit message https://github.com/freeipa/freeipa/pull/4193/commits/2237992e9354b29aa0281c8d96972e95ac3140df.  Subsequent commits' commit messages explicitly say that for background, motivation and details refer back to `$THIS_COMMIT~<n>` for the correct value of `<n>`.

(I cannot reference the commit hash because that will change when we merge.
Although if I rebase and push myself instead pressing "merge" button on GitHub, I
could mention the exact commit SHA).

frasertweedale:
Reflecting the whole text of the commit message here, in case I rebase the branch and lose the above link to the commit.

```
BACKGROUND:

We are implementing ACME support in FreeIPA (umbrella ticket:
https://pagure.io/freeipa/issue/4751).  ACME is defined in RFC 8555.
HTTPS is REQUIRED (https://tools.ietf.org/html/rfc8555#section-6.1).
Therefore, every FreeIPA server that provides the ACME service
capability must be reachable by HTTPS.

RFC 8555 does not say anything about which port to use for ACME.
The default HTTPS port of 443 is implied.  Therefore, the FreeIPA
ACME service will be reached via the Apache httpd server, which will
be the TLS server endpoint.

As a usability affordance for ACME clients, and as a maintainability
consideration i.e. to allow the topology to change without having to
reconfigure ACME clients, there should be a a single DNS name used
to reach the IPA ACME service.

The question then, is which DNS name to use.

REQUIREMENTS:

Each FreeIPA server that is also an ACME server must:

1. Be reachable via a common DNS name

2. Have an HTTP service certificate with that DNS name as a SAN
   dNSName value

DESIGN CONSIDERATION - WHAT DNS NAME TO USE?:

Some unrelated FreeIPA ACME design decisions provide important
context for the DNS name decision:

- The ACME service will be automatically and unconditionally
  deployed (but not necessarily *enabled*) on all CA servers.

- Enabling or disabling the ACME service will have topology-wide
  effect, i.e. the ACME service is either enabled on all CA
  servers, or disabled on all CA servers.

In a CA-ful FreeIPA deployment there is already a DNS name that
resolves to all CA servers: ``ipa-ca.$DOMAIN``, e.g.
``ipa-ca.example.com``.  It is expected to point to all CA servers
in the deployment, and *only* to CA servers.  If internal DNS is
deployed, the DNS records for ``ipa-ca.$DOMAIN`` are created and
updated automatically.  If internal DNS is not deployed,
administrators are required to maintain these DNS records
themselves.

The ``ipa-ca.$DOMAIN`` alias is currently used for OCSP and CRL
access.  TLS is not required for these applications (and it can
actually be problematic for OCSP).  Enabling TLS for this name
presents some risk of confusion for operators.  For example, if they
see that TLS is available and alter the certificate profiles to
include an HTTPS OCSP URL in the Authority Information Access (AIA)
extension, OCSP-using clients may fail to validate such
certificates.  But it is possible for administrators to make such a
change to the profile, whether or not HTTPS is available.

One big advantage to using the ``ipa-ca.$DOMAIN`` DNS name is that
there are no new DNS records to manage, either in the FreeIPA
implementation or for administrators in external DNS systems.

The alternative approach is to define a new DNS name, e.g.
``ipa-acme.$DOMAIN``, that ACME clients would use.  For internal
DNS, this means the FreeIPA implementation must manage the DNS
records.  This is straightforward; whenever we add or remove an
``ipa-ca.$DOMAIN`` record, also add/remove the ``ipa-acme.$DOMAIN``
record.  But for CA-ful deployments using external DNS, it is
additional work for adminstrators and, unless automated, additional
room for error.

An advantage of using a different DNS name is ``ipa-ca.$DOMAIN`` can
remain inaccessible over HTTPS.  This possibly reduces the risk of
administrator confusion or creation of invalid AIA configuration in
certificate profiles.

Weighing up the advantages and disadvantages, I decided to use the
``ipa-ca.$DOMAIN`` DNS name.

DESIGN CONSIDERATION - CA SERVERS, OR ALL SERVERS?:

A separate decision from which name to use is whether to include it
on the HTTP service certificate for ACME servers (i.e. CA servers)
only, or on all IPA servers.

Combined with the assumption that the chosen DNS name points to CA
servers *only*, there does not seem to be any harm in adding it to
the certificates on all IPA servers.

The alternative is to only include the chosen DNS name on the HTTP
service certificates of CA servers.  This approach entails some
additional complexity:

- If a non-CA replica gets promoted to CA replica (i.e. via
  ``ipa-ca-install``), its HTTP certificate must be re-issued with
  the relevant name.

- ipa-server-upgrade code must consider whether the server is a CA
  replica when validating (and if necessary re-creating) Certmonger
  tracking requests

- IPA Health Check must be made aware of this factor when checking
  certificates and Certmonger tracking requests.

Weighing up the options, I decided to add the common DNS name to the
HTTP service certificate on all IPA servers.  This avoids the
implementation complexity discussed above.

CHANGES IN THIS COMMIT

When (re-)tracking the HTTP certificate, explicitly add the server
FQDN and ipa-ca.$DOMAIN DNS names to the Certmonger tracking request.

Related changes follow in subsequent commits.

```

ipalib/install/certmonger.py
```diff
@@ -47,6 +47,29 @@
...
+
+``nickname``
+  a.k.a. "request ID".  If given, only the specified request is
+  retrieved (if it exists), and it is still tested against other
+  criteria.
+
+``ca-name``
+  Test equality against the nickname of the CA (a.k.a. request
+  helper) object for the request.
+
```
rcritten: Not at all a show-stopper but is it worth mentioning that all criteria names are available in certmonger/src/tdbus.h? Those prefixed with CM_DBUS_PROP_

rcritten:
The patch overall is in good shape, minus one little nit/suggestion (not required).

Is there a way this can be tested using an integration test? Can we at least check that on new installs the HTTP cert has the ipa-ca SAN? To avoid regressions? Is there anything else we might be able to poke at?

rcritten:
I set ipa-next on this since the ticket doesn't have a milestone. If this is correct feel free to add the appropriate branches for backporting.

frasertweedale:
@rcritten thanks, I'll add a test.  Yes, ipa-next (master only) is correct.

ipaserver/install/httpinstance.py
```diff
@@ -593,6 +593,7 @@ def start_tracking_certificates(self):
                 post_command='restart_httpd', storage='FILE',
                 profile=dogtag.DEFAULT_PROFILE,
                 pinfile=key_passwd_file,
+                dns=[self.fqdn, f'{IPA_CA_RECORD}.{api.env.domain}'],
```
tiran: The upgrader code uses ``f'{ipautil.format_netloc(api.env.domain)}'`` instead of ``f'{api.env.domain}'``.

frasertweedale: @tiran that is a good pickup. AFAICT it should not make a difference, because (in theory) it must be a hostname and not an IPv6 address.

I looked at the git blame and can't perceive a clear reason why it was written this way in `upgrade.py`, except perhaps an abundance of caution in case it was somehow an IPv6 address.

Do you perfer to remove the application of `ipautil.format_netloc` and just use plain `f{api.env.doman}`.  Or add `ipautil.format_netloc(...)` to the other places?  Or just ignore the difference?

frasertweedale:
@rcritten: test added (tip commit)

frasertweedale:
@rcritten thanks for the ACK.  Trivial rebase on latest `master` to run the tests again.  Will merge when green.

frasertweedale:
master:

* 0711c4a0d45f9e28459596552ca751890b13c265 certmonger: avoid mutable default argument
* e0fb3816f656dc4b324fc02c8d4506dadcfe544e certmonger: move 'criteria' description to module docstring
* 18ebd1116d706d9f2a3a9c2dca16681a4f138362 certmonger: support dnsname as request search criterion
* 4cf9c8689fc4a791eee17e0cfcb38e837a53b3f0 httpinstance: add fqdn and ipa-ca alias to Certmonger request
* f7c45641fe356364175899399b51b07987b44a1b cert-request: allow ipa-ca.$DOMAIN dNSName for IPA servers
* 4d5b5a9024c1237cb128688c7c4ac796135d50d5 httpinstance: add ipa-ca.$DOMAIN alias in initial request
* cf4c2c64b0bb1e4555a48bc5079aeff101fd9894 upgrade: add ipa-ca.$DOMAIN alias to HTTP certificate
* 45b5384b6ef83aaf742bf7906d846e07db874ef8 (HEAD) ipatests: check HTTP certificate contains ipa-ca.$DOMAIN dnsname

PR #4206, opened 2020-02-06
[Backport][ipa-4-7] WebUI tests: Fix broken reference to parent facet in table record check

serg-cymbaluk:
Manual backport of https://github.com/freeipa/freeipa/pull/4192

Add decorator to has_record method which repeats the check when an active facet is changed
(catch StaleElementReferenceException).

Ticket: https://pagure.io/freeipa/issue/8157


abbra:
ipa-4-7:

* a45679047f82a3fbb98422a023a782766df0ca25 WebUI tests: Fix broken reference to parent facet in table record check

PR #4217, opened 2020-02-10
ipatests: fix modify_sssd_conf()

flo-renaud:
The method modify_sssd_conf() is copying a remote sssd.conf file
to the test controller then uses sssd python API to modify the
config file.
When the test controller does not have sssd-common package installed,
SSSDConfig() call fails because the API needs sssd schema in order
to properly parse the config file, and the schema files are provided
by sssd-common pkg.
The fix also downloads the files representing sssd schema and calls
SSSDConfig() with those files.

Note: we currently don't see any issue in the nightly tests because
the test controller is installed with sssd-common package but if you
run the tests as specified in https://www.freeipa.org/page/Testing
with a controller missing sssd-common, you will see the issue.

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -28,6 +28,7 @@ import textwrap
 import re
 import collections
 import itertools
+import shutil
 import tempfile
 import time
 from pipes import quote
@@ -820,7 +821,30 @@ def modify_sssd_conf(host, domain, mod_dict, provider='ipa',
         with open(temp_config_file, 'wb') as f:
             f.write(current_config)
 
-        sssd_config = SSSDConfig()
+        # In order to use SSSDConfig() locally we need to import the schema
+        # Create a tar file with /usr/share/sssd.api.conf and
+        # /usr/share/sssd/sssd.api.d
+        tmpname = create_temp_file(host)
+        host.run_command(
+            ['tar', 'cJvf', tmpname,
+             'sssd.api.conf',
+             'sssd.api.d'],
+            log_stdout=False, cwd="/usr/share/sssd")
+        # fetch tar file
+        tar_dir = tempfile.mkdtemp()
+        tarname = os.path.join(tar_dir, "sssd_schema.tar.xz")
+        with open(tarname, 'wb') as f:
+            f.write(host.get_file_contents(tmpname))
+        # delete from remote
+        host.run_command(['rm', '-f', tmpname])
+        # Unpack on the local side
+        ipautil.run([paths.TAR, 'xJvf', tarname], cwd=tar_dir)
+        os.unlink(tarname)
+
+        # Use the imported schema
+        sssd_config = SSSDConfig(
+            schemafile=os.path.join(tar_dir, "sssd.api.conf"),
+            schemaplugindir=os.path.join(tar_dir, "sssd.api.d"))
         sssd_config.import_config(temp_config_file)
         sssd_domain = sssd_config.get_domain(domain)
 
@@ -837,6 +861,7 @@ def modify_sssd_conf(host, domain, mod_dict, provider='ipa',
     finally:
         try:
             os.remove(temp_config_file)
+            shutil.rmtree(tar_dir)
         except OSError:
             pass
 


flo-renaud:
test_sudo is using this method, results can be seen [here](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/7db77476-4be9-11ea-af05-fa163e7f018b)

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -820,7 +821,32 @@ def modify_sssd_conf(host, domain, mod_dict, provider='ipa',
...
             f.write(current_config)
 
-        sssd_config = SSSDConfig()
+        # In order to use SSSDConfig() locally we need to import the schema
+        # Create a tar file with /usr/share/sssd.api.conf and
+        # /usr/share/sssd/sssd.api.d
+        cmd = host.run_command(['mktemp'])
+        tmpname = cmd.stdout_text.strip()
+        cmd = host.run_command(
+            ['tar', 'cJvf', tmpname, '--ignore-failed-read',
```
wladich: Why do you ignore errors here? I think we want to detect problems immediately when they appear.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -820,7 +821,32 @@ def modify_sssd_conf(host, domain, mod_dict, provider='ipa',
...
+        # fetch tar file
+        tar_dir = tempfile.mkdtemp()
+        tarname = os.path.join(tar_dir, "sssd_schema.tar.xz")
+        with open(tarname, 'wb') as f:
+            f.write(host.get_file_contents(tmpname))
+        # delete from remote
+        host.run_command(['rm', '-f', tmpname])
+        # Unpack on the local side
+        ipautil.run([paths.TAR, 'xJvf', tarname], cwd=tar_dir,
+                    raiseonerr=False)
```
wladich: Why do you ignore errors here? I think we want to detect problems immediately when they appear.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -820,7 +821,32 @@ def modify_sssd_conf(host, domain, mod_dict, provider='ipa',
...
+            log_stdout=False, raiseonerr=False, cwd="/usr/share/sssd")
+        # fetch tar file
+        tar_dir = tempfile.mkdtemp()
+        tarname = os.path.join(tar_dir, "sssd_schema.tar.xz")
+        with open(tarname, 'wb') as f:
+            f.write(host.get_file_contents(tmpname))
+        # delete from remote
+        host.run_command(['rm', '-f', tmpname])
+        # Unpack on the local side
+        ipautil.run([paths.TAR, 'xJvf', tarname], cwd=tar_dir,
```
wladich: Can you tell what is the benefit of using ipautil.run instead of subprocess.call/check_call?
(Not a review comment, just interested)

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -820,7 +821,32 @@ def modify_sssd_conf(host, domain, mod_dict, provider='ipa',
         with open(temp_config_file, 'wb') as f:
             f.write(current_config)
 
-        sssd_config = SSSDConfig()
+        # In order to use SSSDConfig() locally we need to import the schema
+        # Create a tar file with /usr/share/sssd.api.conf and
+        # /usr/share/sssd/sssd.api.d
```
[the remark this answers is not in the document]
flo-renaud: Honestly I think this time is negligible, especially compared to the time spent in log collections that is done between each test.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -820,7 +821,32 @@ def modify_sssd_conf(host, domain, mod_dict, provider='ipa',
...
+        cmd = host.run_command(['mktemp'])
+        tmpname = cmd.stdout_text.strip()
```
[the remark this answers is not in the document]
flo-renaud: thanks for the tip

ipatests/pytest_ipa/integration/tasks.py
flo-renaud: Good catch, I blindly copied/pasted other code

ipatests/pytest_ipa/integration/tasks.py
flo-renaud: Good catch, I blindly copied/pasted other code

ipatests/pytest_ipa/integration/tasks.py
flo-renaud: Just out of habit because this wrapper also manages the encoding.

wladich:
ACK

flo-renaud:
master:

* cec1ddc39ecd2163270bb00df46bfd2469c2c25a ipatests: fix modify_sssd_conf()

PR #4220, opened 2020-02-11
ipatests: Test AES SHA 256 and 384 Kerberos enctypes enabled

mrizwan93:
AES SHA 256 and 384-bit enctypes supported by MIT kerberos but
was not enabled in IPA. This test is to check if these types are
enabled.

related: https://pagure.io/freeipa/issue/8110

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -33,6 +33,7 @@ from ipatests.pytest_ipa.integration import tasks
 from ipaplatform.tasks import tasks as platform_tasks
 from ipatests.create_external_ca import ExternalCA
 from ipatests.test_ipalib.test_x509 import good_pkcs7, badcert
+from ipapython.ipautil import realm_to_suffix
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


fcami:
Hi @mrizwan93 could you please explain why this test case should go into this particular class? TestInstallWithCA_KRA2 executes the tests defined in InstallTestBase2 so far.

mrizwan93:
@fcami moved to `TestInstallMaster`

fcami:
Hi @mrizwan93 I still fail to see how this test is installation-related and I see including this in any non-related test suite quite confusion. Maybe a new test suite test_sanity would do.

mrizwan93:
HI @fcami , Sorry I didn't got your point in previous iteration. 

The test is to check if  `aes128-sha2:normal`  `aes128-sha2:special`  `aes256-sha2:normal` and `aes256-sha2:special` are included in `krbSupportedEncSaltTypes` on newly installed master. IMO, it is relevant to this suite.

fcami:
hi @mrizwan93 It's not exactly installation-related, more like features-related.
We have a [test suite](https://github.com/freeipa/freeipa/blob/master/ipatests/test_integration/test_commands.py#L120) where this would fit. Could you please move the test to that? Thanks.

ipatests/test_integration/test_commands.py
```diff
@@ -130,6 +131,27 @@ def get_cert_base64(self, host, path):
...
+        AES SHA 256 and 384-bit enctypes supported by MIT kerberos but
+        was not enabled in IPA. This test is to check if these types are
+        enabled.
+
+        related: https://pagure.io/freeipa/issue/8110
+        """
+        tasks.kinit_admin(self.master)
+        realm = self.master.domain.realm
+        suffix = realm_to_suffix(realm)
+        dn = DN(("cn", realm), ("cn", "kerberos")) + suffix
```
rcritten: IMHO the extra variables aren't needed and you don't need to add the DNs:

dn = DN(("cn", realm), ("cn", "kerberos"), realm_to_suffix(self.master.domain.realm))

rcritten:

rcritten:
master:

* b0d57d99e5719099ed0be0476985ab394b131cb0 Test AES SHA 256 and 384 Kerberos enctypes enabled

PR #4221, opened 2020-02-11
adtrust: print DNS records for external DNS case after role is enabled

abbra:
We cannot gather information about required DNS records before "ADTrust
Controller" role is enabled on this server. As result, we need to call
the step to add DNS records after the role was enabled.

Fixes: https://pagure.io/freeipa/issue/8192

### DIFF
--- a/ipaserver/install/adtrust.py
+++ b/ipaserver/install/adtrust.py
@@ -26,6 +26,8 @@ from ipaserver.install import installutils
 from ipaserver.install import adtrustinstance
 from ipaserver.install import service
 from ipaserver.install.plugins.adtrust import update_host_cifs_keytabs
+from ipaserver.install.bindinstance import dns_zone_exists
+from ipaserver.dns_data_management import IPASystemRecords
 
 
 if six.PY3:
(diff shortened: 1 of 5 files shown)


ipaserver/install/adtrust.py
```diff
@@ -435,6 +435,9 @@ def install(standalone, options, fstore, api):
         # and propose them to be added to the list
         add_new_adtrust_agents(api, options)
 
+    # Enable configured services and update DNS SRV records
+    service.sync_services_state(api.env.host)
```
tiran: The ``adtrust.install()`` is also called from the ``ipa-server-install`` script. This change may enable the server a bit too early if the installer is called with ``--setup-adtrust``.

abbra: .. and also from `ipa-replica-install` too. I'll check these paths.

flo-renaud:
Hi @abbra 
overall the change looks good but when ipa-adtrust-install is run on a system with DNS enabled, the DNS records should get automatically added and it's not happening any more.

abbra:
@flo-renaud thanks for the review. I see that I removed one unconditional DNS record updates. I replaced it now with a call to update it only if no DNS help needs to be printed. This simulates the behavior of the original code.

flo-renaud:
Hi @abbra 
thanks for the update, LGTM. Tested the following:
-[x] ipa-server-install without DNS: prints all the missing records in a file
-[x] ipa-adtrust-install without DNS: prints the missing trust-related records on stdout
-[x] ipa-replica-install with AD trust without DNS: prints the missing trust-related records on stdout
-[x] ipa-server-install with AD trust without DNS: prints the missing trust-related records on stdout + all records in a file
When DNS is setup, there is no message about missing records, as expected.

abbra:
master:

* b3dbb36867ebf52483834ceb247050a24cf74d7c adtrust: print DNS records for external DNS case after role is enabled

PR #4232, opened 2020-02-12
[Backport][ipa-4-6] install/updates: move external members past schema compat update

abbra:
There is an ordering discrepancy because the base compat tree
configuration is in install/updates/80-schema_compat.update so it is ran
after 50-externalmembers.update. And since at that point
cn=groups,cn=Schema ... does not exist yet, external members
configuration is not applied.

Move it around to make sure it is applied after Schema Compatibility
plugin configuration is created.

Fixes: https://pagure.io/freeipa/issue/8193
Reviewed-By: Florence Blanc-Renaud <flo@redhat.com>


flo-renaud:
ipa-4-6:

* a5a201fc008b19841f98bb70d44ede7d04ef1126 install/updates: move external members past schema compat update

PR #4234, opened 2020-02-13
dnsrecord: Treat empty list arguments correctly

tiran:
dnsrecord_del fails when one of the record arguments is an empty list:

    AttrValueNotFound("AAAA record does not contain 'None'",)

The problem is caused by the fact that LDAPEntry.__getitem__ returns None
for empty lists. The code in the plugin considers None as a single entry
and maps it to vals = [None].

The patch maps None to empty list.

Fixes: https://pagure.io/freeipa/issue/8196


abbra:
ACK.

abbra:
master:

* 856fdbc183b31ea7432749ff3acb3ff5b73937e6 dnsrecord: Treat empty list arguments correctly

PR #4235, opened 2020-02-13
Remove dependency on custodia package

tiran:
ipa-server no longer use any files and features from the custodia
package. The python3-custodia package provides all Custodia features for
ipa-custodia.service.


abbra:
ACK.

abbra:
master:

* 209e0ac8a6a70dda2f3db7108991187683f0db73 Remove dependency on custodia package

PR #4249, opened 2020-02-17
kdb: make sure audit_as_req callback signature change is preserved

abbra:
audit_as_req() callback has changed its signature with MIT krb5 commit
20991d55efbe1f987c1dbc1065f2d58c8f34031b in 2017, we should preserve the
change for any newer DAL versions. Otherwise audit_as_req() callback
would reference wrong data and we might crash.

Fixes: https://pagure.io/freeipa/issue/8200


abbra:
master:

* e5291963a148c34582d90f2aa269e4a2214f9f6f kdb: make sure audit_as_req callback signature change is preserved
* ba904672a21add3b0ab1808c5640ac786f194aad Azure Pipelines: re-enable nodejs:12 stream for Fedora 31+

PR #4252, opened 2020-02-17
Don't fully quality the FQDN in ssbrowser.html for Chrome

rcritten:
The trailing dot causes it to not function as expected, remove
it from the example.

https://pagure.io/freeipa/issue/8201


rcritten:
master:

* e4966f9c3f9566603be24e09f4803b3183898272 Don't fully quality the FQDN in ssbrowser.html for Chrome

PR #4260, opened 2020-02-18
ipatests: add test for warning text displayed after AD trust agents setup

wladich:
This PR created new suite for testing ipa-adtrust-install utility.
It adds test for proper warning text after setting up AD trust agents and moves one existing from test_commands test suite.

### DIFF
--- a/ipatests/test_integration/test_adtrust_install.py
+++ b/ipatests/test_integration/test_adtrust_install.py
@@ -1,0 +1,48 @@
+#
+# Copyright (C) 2020  FreeIPA Contributors see COPYING for license
+#
+
+"""This module provides tests for ipa-adtrust-install utility"""
+
+import re
+
+from ipatests.test_integration.base import IntegrationTest
+
+
+class TestIpaAdTrustInstall(IntegrationTest):
+    topology = 'line'
+    num_replicas = 1
+
+    def test_install_ad_trust_controller(self):
+        self.master.run_command(['ipa-adtrust-install', '-U'])
+
+    def test_warnings_after_ad_trust_agents_setup(self):
+        """ Check that warning about ipa and sssd restart is displayed.
+
+        Regression test for https://bugzilla.redhat.com/1782658
+        """
+        cmd_input = (
+            # admin password:
+            self.master.config.admin_password + '\n' +
+            # WARNING: The smb.conf already exists. Running ipa-adtrust-install
+            # will break your existing samba configuration.
+            # Do you wish to continue? [no]:
+            'yes\n'
+            # Enable trusted domains support in slapi-nis? [no]:
+            '\n' +
+            # Do you want to run the ipa-sidgen task? [no]:
+            '\n'+
+            # WARNING: 1 IPA masters are not yet able to serve information
+            # about users from trusted forests.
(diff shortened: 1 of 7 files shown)


ipatests/test_integration/test_adtrust_install.py
```diff
@@ -0,0 +1,54 @@
...
+    num_replicas = 1
+
+    def test_samba_config_file(self):
+        """Check that ipa-adtrust-install generates sane smb.conf
+
+        This is regression test for issue
+        https://pagure.io/freeipa/issue/6951
+        """
+        self.master.run_command(
+            ['ipa-adtrust-install', '-a', 'Secret123', '--add-sids', '-U'])
```
amore17: Instead of hard-coded password you can use self.master.config.admin_password

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -0,0 +1,54 @@
...
+        res = self.master.run_command(['testparm', '-s'])
+        assert 'ERROR' not in (res.stdout_text + res.stderr_text)
+
+    def test_warnings_after_ad_trust_agents_setup(self):
+        """ Check that warning about ipa and sssd restart is displayed.
+
+        Regression test for https://bugzilla.redhat.com/1782658
```
amore17: Instead of bugzilla link could you please add related pagure issue.

amore17:
@abbra Could you also suggest negative test for replica.
If sssd is not restarted what will fail and what is expected to pass after sssd restart.

abbra:
@amore17 it is a bit more complex than just SSSD restart.
SSSD needs to know it has to access the trusted domain controllers for performing a search for trusted users and groups. To do so, it needs to fetch a keytab with a key representing credentials for the special account in a trusted domain. Without access rights to that information (e.g. replica being part of a special group in IPA) that is not possible.

Now, SSSD has two points when it might attempt to fetch the keytab. One: after initial start (or restart, as that is the same here). Another: when switching from offline to online state for a domain.

If SSSD has no credentials to access a trusted domain controller, it will consider the domain in question in offline mode. If we don't restart SSSD, it still might try to switch a domain state to online occasionally which will involve an attempt to fetch a keytab first.

So here we get a particular situation that might succeed: if the replica host was added to that special group in IPA, granting access to the trusted domain access, but SSSD wasn't restarted and it decided to try to switch from offline to online for the trusted domain in question.

If we consider the fact that SSSD is unable to resolve a trusted domain user before a restart, our negative test might actually fail (e.g. resolution would succeed) if SSSD decides to switch from offline to online mode when we ask it to resolve that user after the replica was already added to the special group. So we might not have a negative test case here.

The way how this happens heavily depends on time passed between two unsuccessful user/group resolution requests for that domain. SSSD caches negative results for some time to not cause a flurry of checks in short time. So a negative test would probably require to try several tests in a row,  note a failure and restart sssd, then try again.

abbra:
Another thing is to not consider a failure the fact that resolving a user/group before SSSD restart succeeds. That's OK, it is possible we stumbled on the attempt to refresh the domain state when we asked for a user/group info. Coincidence is fine.

wladich:
Closing in favor of #4277

PR #4077, opened 2019-12-24
ipatests: SSSD should fetch external groups without any limit.

amore17:
When there are more external groups than default limit, then
SSSD should fetch all groups.

Regression test for : https://pagure.io/SSSD/sssd/issue/4058

PR #4265, opened 2020-02-20
[Backport][ipa-4-6] ipatests: SSSD should fetch external groups without any limit.

amore17:
This is manual back-port of : #4077

### DIFF
--- a/ipatests/test_integration/test_sssd.py
+++ b/ipatests/test_integration/test_sssd.py
@@ -14,6 +14,7 @@ import textwrap
 
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
+from ipaplatform.tasks import tasks as platform_tasks
 from ipaplatform.paths import paths
 from ipapython.dn import DN
 
@@ -124,6 +125,12 @@ class TestSSSDWithAdTrust(IntegrationTest):
         2: Run the command id aduser@ADDOMAIN.COM
         3: sssd should retrieve all the external groups.
         """
+        cmd = self.master.run_command(['sssd', '--version'])
+        sssd_version = platform_tasks.parse_ipa_version(
+            cmd.stdout_text.strip())
+        if sssd_version <= platform_tasks.parse_ipa_version('1.16.3'):
+            pytest.xfail("Fix for https://pagure.io/SSSD/sssd/issue/4058 "
+                         "unavailable with sssd-1.16.3")
         new_limit = 50
         master = self.master
         conn = master.ldap_connect()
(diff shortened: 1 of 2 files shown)


ipatests/test_integration/test_sssd.py
```diff
@@ -113,3 +117,60 @@ def test_extdom_group(self):
...
+
+        Regression test for https://pagure.io/SSSD/sssd/issue/4058
+        1: Add external groups more than limit.
+        2: Run the command id aduser@ADDOMAIN.COM
+        3: sssd should retrieve all the external groups.
+        """
+        cmd = self.master.run_command(['sssd', '--version'])
+        sssd_version = platform_tasks.parse_ipa_version(
+            cmd.stdout_text.strip())
+        if LooseVersion(sssd_version.version) <= LooseVersion('1.16.3'):
```
flo-renaud: Do you really need LooseVersion or would it work with:
```
if sssd_version <= platform_tasks.parse_ipa_version('1.16.3'):
```

ipatests/test_integration/test_sssd.py
```diff
@@ -113,3 +117,60 @@ def test_extdom_group(self):
...
+            pytest.xfail("Fix for https://pagure.io/SSSD/sssd/issue/4058"
```
flo-renaud: Need to add a space before the closing "

ipatests/test_integration/test_sssd.py
amore17: sssd_version <= platform_tasks.parse_ipa_version('1.16.3'): this will compare string with a string.
That's why it needs LooseVersion, as sssd_version.version with type string.

flo-renaud: Hi @amore17 

> sssd_version <= platform_tasks.parse_ipa_version('1.16.3'): this will compare string with a string.

I'm afraid I disagree here. `parse_ipa_version` is using `pkg_resources.parse_version()` and this [method](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#parsing-utilities) returns an object that can be compared/sorted.
Using `sssd_version.version` means using the string representation and the benefit of the parsing is lost.

flo-renaud:
Hi @amore17 
thanks fo rthe backport, LGTM and the test is properly skipped because of SSSD version. You can remove the temp commit and I will ACK when PRCI completes.

flo-renaud:
ipa-4-6:

* fd74fcf75606ded2987753337161c163e8ae9a44 ipatests: SSSD should fetch external groups without any limit.
* 48ecb92afdbd577fbb4fe05ea15cfaf44e504f89 Cherry-picked only ldapmodify_dm()
* 0c828dad4cfd3df9db8056b2497543c022c7680a Mark xfail for sssd-version 1.16.3

PR #4272, opened 2020-02-24
Added tests for 2FA prompting configurations

amore17:
Added test when 2FA prompting configurations is set.
Test1:       
When [prompting/2fa/sshd] with single_prompt = True is set
then during ssh it should be prompted with given massage
for first and second factor at once.
Test2:
When [prompting/2fa/sshd] with single_prompt = False is set
then during ssh it should be prompted with given massage
for first factor and then for second factor.

related : https://pagure.io/SSSD/sssd/issue/3264

### DIFF
--- a/ipatests/test_integration/test_otp.py
+++ b/ipatests/test_integration/test_otp.py
@@ -4,8 +4,11 @@
 """OTP token tests
 """
 import base64
+import logging
+import paramiko
 import re
 import time
+import textwrap
 from urllib.parse import urlparse, parse_qs
 
 from cryptography.hazmat.backends import default_backend
@@ -14,12 +17,14 @@ from cryptography.hazmat.primitives.twofactor.hotp import HOTP
 from cryptography.hazmat.primitives.twofactor.totp import TOTP
 
 from ipatests.test_integration.base import IntegrationTest
+from ipaplatform.paths import paths
 from ipatests.pytest_ipa.integration import tasks
 
 
 PASSWORD = "DummyPassword123"
 USER = "opttestuser"
 ARMOR = "/tmp/armor"
+logger = logging.getLogger(__name__)
 
 
 def add_otptoken(host, owner, *, otptype="hotp", digits=6, algo="sha1"):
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @amore17 
I relaunched the azure tests as they had issue pulling the container.

ipatests/test_integration/test_otp.py
```diff
@@ -183,3 +212,82 @@ def test_otptoken_sync(self):
...
 
         del_otptoken(master, otpuid)
+
+    def test_2fa_enable_single_prompt(self):
+        """Test ssh with 2FA when single prompt is enabled.
+
+        Test for : https://pagure.io/SSSD/sssd/issue/3264
+
+        When [prompting/2fa/sshd] with single_prompt = True is set
+        then during ssh it should be prompted with given massage
```
abbra: A typo: massage instead of message

ipatests/test_integration/test_otp.py
```diff
@@ -183,3 +212,82 @@ def test_otptoken_sync(self):
...
+            self.master.run_command(['semanage', 'login', '-D'])
+            sssd_conf_backup.restore()
+
+    def test_2fa_disable_single_prompt(self):
+        """Test ssh with 2FA when single prompt is disabled.
+
+        Test for : https://pagure.io/SSSD/sssd/issue/3264
+
+        When [prompting/2fa/sshd] with single_prompt = False is set
+        then during ssh it should be prompted with given massage
```
abbra: Same here: massage instead of message

ipatests/test_integration/test_otp.py
```diff
@@ -74,6 +77,32 @@ def kinit_otp(host, user, *, password, otp, success=True):
...
+        for prmpt in prompt_list:
+            prmpt_str = prmpt[0].strip()
+            resp.append(answers_dict[prmpt_str])
+        return resp
+    trans = paramiko.Transport((hostname, port))
+    trans.connect()
+    trans.auth_interactive_dumb(username, answer_handler)
+
+
+def set_sssd_conf(host, add_contents):
```
abbra: I'm not entirely satisfied with this approach where you add a section to existing `sssd.conf` file. There is no code to remove the added section after the test, so it would affect other tests if they would be added here later.
Since SSSD is able to load configuration file fragments from `/etc/sssd/conf.d` directory, may be better would be to create a file in `/etc/sssd/conf.d/` and remove it afterwards?

amore17: > I'm not entirely satisfied with this approach where you add a section to existing `sssd.conf` file. There is no code to remove the added section after the test, so it would affect other tests if they would be added here later.
@abbra Thanks for the review.
In finally block it is restored with original using 
`sssd_conf_backup.restore()`

abbra: Ah, thanks, you are right. I think this should be OK.

abbra:
ACK. For the record, now we have actual prompt and response in the logs:
http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/f212da32-5c6e-11ea-ae68-fa163e0a82d3/report.html, the snippet below is from `test_integration/test_otp.py::TestOTPToken::test_2fa_disable_single_prompt` test:
```
INFO     paramiko.transport:transport.py:1819 Connected (version 2.0, client OpenSSH_8.1)
INFO     ipatests.test_integration.test_otp:test_otp.py:99 Prompt is: 'Enter first factor:'
INFO     ipatests.test_integration.test_otp:test_otp.py:101 Answer to ssh prompt is: 'DummyPassword123'
INFO     ipatests.test_integration.test_otp:test_otp.py:99 Prompt is: 'Enter second factor:'
INFO     ipatests.test_integration.test_otp:test_otp.py:101 Answer to ssh prompt is: '886982'
INFO     paramiko.transport:transport.py:1819 Authentication (keyboard-interactive) successful!
```

amore17:
result : http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/f212da32-5c6e-11ea-ae68-fa163e0a82d3/report.html

abbra:
master:

* 8007cec855fedc89593b0af92be282aa0e65e5c6 ipatests: Added test when 2FA prompting configurations is set.

abbra:
Manual backports to ipa-4-7 and ipa-4-6 are needed.

PR #4274, opened 2020-02-24
ipatests: Added check that no Traceback found in idviews operations

varunmylaraiah:
https://pagure.io/freeipa/issue/7958

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -2232,3 +2232,17 @@ def get_sssd_version(host):
     """Get sssd version on remote host."""
     version = host.run_command('sssd --version').stdout_text.strip()
     return parse_version(version)
+
+
+def assert_string(host, string_to_check, file_name, assert_ok=True):
+    """ Assert the given string in specified file
+    :param host: host where the check takes place
+    :param string_to_check: string to be check
+    :param file_name: file on which check is to be performed.
+    :param assert_ok: If true assert in file else assert not
+    """
+    file_contents = host.get_file_contents(file_name, encoding='utf-8')
+    if assert_ok:
+        assert string_to_check in file_contents
+    else:
+        assert string_to_check not in file_contents
--- a/ipatests/test_integration/test_idviews.py
+++ b/ipatests/test_integration/test_idviews.py
@@ -28,6 +28,8 @@ class TestCertsInIDOverrides(IntegrationTest):
     def uninstall(cls, mh):
         super(TestCertsInIDOverrides, cls).uninstall(mh)
         cls.master.run_command(['rm', '-rf', cls.reqdir], raiseonerr=False)
+        tasks.assert_string(cls.master, "Traceback",
+                            paths.VAR_LOG_HTTPD_ERROR, assert_ok=False)
 
     @classmethod
     def install(cls, mh):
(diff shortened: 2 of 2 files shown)


varunmylaraiah:
@fcami Could you please review?

ipatests/test_integration/test_idviews.py
```diff
@@ -28,6 +28,9 @@ class TestCertsInIDOverrides(IntegrationTest):
     def uninstall(cls, mh):
         super(TestCertsInIDOverrides, cls).uninstall(mh)
         cls.master.run_command(['rm', '-rf', cls.reqdir], raiseonerr=False)
+        # Checking Traceback https://pagure.io/freeipa/issue/7958
+        http_logs = cls.master.get_file_contents(paths.VAR_LOG_HTTPD_ERROR)
```
kaleemsiddiqu: @varunmylaraiah 
This check is just applicable to TestCertsInIDOverrides test class and not applied to TestIDViews test class.
Better would be to move this to pytest_ipa/integration/tasks.py as utility function and use that in every test class's e.g in this case TestCertsInIDOverrides and TestIDViews.

Also test out these changes with help of temp_commit.yaml for these two test classes by including these two specific tests in temp_commit.yaml.

amore17:
For failures in test_integration/test_idviews.py::TestCertsInIDOverrides::test_certs_in_idoverrides_ad_users::setup

1. You need to define path for adcert1_file and adcert2_file same as 'cls.pwname = os.path.join(cls.reqdir, "pwd")' 
2. Also  
tasks.run_certutil(master, ['-L', '-n', cls.adcert1, '-a', '>', cls.adcert1_file], cls.reqdir)
tasks.run_certutil(master, ['-L', '-n', cls.adcert2, '-a', '>', cls.adcert2_file], cls.reqdir)
Here instead of redirecting output you can use put_file_contents()

amore17:
LGTM you can remove temp_commit.
Also please add in commit message related pagure ticket.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -2196,3 +2196,12 @@ def wait_for_sssd_domain_status_online(host, timeout=120):
...
         raise RuntimeError("SSSD still offline")
+
+
+def check_traceback(host):
+    """Checking Traceback  on a remote host.
+    :param host: the host where the check takes place
+    """
+    http_logs = host.get_file_contents(paths.VAR_LOG_HTTPD_ERROR,
+                                       encoding='utf-8')
+    assert 'Traceback' not in http_logs
```
mrizwan93, line 2207: This method will not be used anywhere else than `test_idviews.py` testsuite. IMO it should be moved to `test_idviews.py` itself. 
OR 
Make it more generalize here in `tasks.py`. May be  something like:
```
def assert_string(host, string_to_check, file_name):
    """ Check the given string in specified file
    : param host: host where the check takes place
    : param string_to_check: string to be check
    : param file_name: file on which check is to be performed.
    """
    file_contents = host.get_file_contents(file_name)
    assert string_to_check in file_contents
```

varunmylaraiah: @mrizwan93 I think we can create separate PR for refactoring work.

varunmylaraiah:
Test run with Temp commit 
http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/fb8c738a-77ed-11ea-a54a-fa163e43b58c/

flo-renaud:
Hi @varunmylaraiah 
I don't really see the point in adding this test. The BZ is marked as sanityOnly, meaning we don't have a reproducer.

varunmylaraiah:
@kaleemsiddiqu I don't remember why this was asked to automate? Can you please clarify this to Flo.

flo-renaud:
@varunmylaraiah 
A suggestion: the check is not really related to the idviews tests. It could be added at the end of all tests. In order to do that, please have a look at freeipa/ipatests/pytest_ipa/integration/__init__.py which defines test fixtures and could be enhanced.

varunmylaraiah:
Sure Flo, will try.

PR #3487, opened 2019-08-01
Allow rename of a host group

abbra:
Fixes: https://pagure.io/freeipa/issue/6783
Signed-off-by: Alexander Bokovoy <abokovoy@redhat.com>

PR #4277, opened 2020-02-24
ipa-adtrust-install: run remote configuration for new agents

flo-renaud:
### ipa-adtrust-install: run remote configuration for new agents

When ipa-adtrust-install is run, the tool detects masters that are
not enabled as trust agents and propose to configure them. With the
current code, the Schema Compat plugin is not enabled on these new
trust agents and a manual restart of LDAP server + SSSD is required.

With this commit, ipa-adtrust-install now calls remote code on the new
agents through JSON RPC api, in order to configure the missing parts.
On the remote agent, the command is using DBus and oddjob to launch
a new command,
/usr/sbin/ipa-trust-enable-agent [--enable-compat]
This command configures the Schema Compat plugin if --enable-compat is
provided, then restarts LDAP server and SSSD.

If the remote agent is an older version and does not support remote
enablement, or if the remote server is not responding, the tool
ipa-adtrust-install prints a WARNING explaining the steps that need
to be manually executed in order to complete the installation, and
exits successfully (keeping the current behavior).

Fixes: https://pagure.io/freeipa/issue/7600

### ipatests: add test for ipa-adtrust-install --add-agents
    
Add tests checking the behavior of ipa-adtrust-install when
adding trust agents:
- try adding a trust agent when the remote node is stopped.
The installer must detect that he's not able to run the remote
commands and print a WARNING.
- try adding a trust agent when the remote node is running.
The WARNING must not be printed as the remote configuration is done.
- try adding a trust agent with --enable-compat.
The WARNING must not be printed and the Schema Compatibility plugin
must be enabled (the entries
cn=users/groups,cn=Schema Compatibility,cn=plugins,cn=config
must contain a new attribute schema-compat-lookup-nsswitch
(=user/group).

Notes:
- this PR will have to be rebased after PR #4260 is merged as it's modifying the same test suite
- TBD: add a SELinux policy rule allowing to call the remote command when #4251 is merged

### DIFF
--- a/ipaserver/install/adtrust.py
+++ b/ipaserver/install/adtrust.py
@@ -14,6 +14,7 @@ import os
 import six
 
 from ipalib.constants import MIN_DOMAIN_LEVEL
+from ipalib import create_api, rpc
 from ipalib import errors
 from ipalib.install.service import ServiceAdminInstallInterface
 from ipalib.install.service import replica_install_only
@@ -344,13 +345,68 @@ def add_new_adtrust_agents(api, options):
     if new_agents:
         add_hosts_to_adtrust_agents(api, new_agents)
 
-        print("""
+        # The method trust_enable_agent was added on API version 2.236
+        # Specifically request this version in the remote call
+        kwargs = {u'version': u'2.236',
+                  u'enable_compat': options.enable_compat}
+        failed_agents = []
+        for agent in new_agents:
+            # Try to run the ipa-trust-enable-agent script on the agent
+            # If the agent is too old and does not support this,
+            # print a msg
+            logger.info("Execute trust_enable_agent on remote server %s",
+                        agent)
+            client = None
+            try:
+                xmlrpc_uri = 'https://{}/ipa/xml'.format(
+                    ipautil.format_netloc(agent))
+                remote_api = create_api(mode=None)
+                remote_api.bootstrap(context='installer',
+                                     confdir=paths.ETC_IPA,
+                                     xmlrpc_uri=xmlrpc_uri,
+                                     fallback=False)
+                client = rpc.jsonclient(remote_api)
+                client.finalize()
+                client.connect()
+                result = client.forward(
+                    u'trust_enable_agent',
+                    ipautil.fsdecode(agent),
+                    **kwargs)
+            except errors.CommandError as e:
+                logger.debug(
+                    "Remote server %s does not support agent enablement "
+                    "over RPC: %s", agent, e)
+                failed_agents.append(agent)
+            except (errors.PublicError, ConnectionRefusedError) as e:
+                logger.debug(
+                    "Remote call to trust_enable_agent failed on server %s: "
+                    "%s", agent, e)
+                failed_agents.append(agent)
+            else:
+                for message in result.get('messages'):
+                    logger.debug('%s', message['message'])
+                if not int(result['result']):
+                    logger.debug(
+                        "ipa-trust-enable-agent returned non-zero exit code "
+                        " on server %s", agent)
+                    failed_agents.append(agent)
+            finally:
+                if client and client.isconnected():
+                    client.disconnect()
+
+        # if enablement failed on some agents, print a WARNING:
+        if failed_agents:
+            if options.enable_compat:
+                print("""
+WARNING: you MUST manually enable the Schema compatibility Plugin and """)
+            print("""
 WARNING: you MUST restart (both "ipactl restart" and "systemctl restart sssd")
 the following IPA masters in order to activate them to serve information about
 users from trusted forests:
 """)
-        for x in new_agents:
-            print(x)
+
+            for x in failed_agents:
+                print(x)
 
 
 def install_check(standalone, options, api):
--- a/ipatests/test_integration/test_adtrust_install.py
+++ b/ipatests/test_integration/test_adtrust_install.py
@@ -1,0 +1,201 @@
+#
+# Copyright (C) 2020  FreeIPA Contributors see COPYING for license
+#
+
+"""This module provides tests for ipa-adtrust-install utility"""
+
+import re
+import textwrap
+
+from ipaplatform.paths import paths
+from ipapython.dn import DN
+from ipatests.pytest_ipa.integration import tasks
+from ipatests.test_integration.base import IntegrationTest
+
+
+class TestIpaAdTrustInstall(IntegrationTest):
+    topology = 'line'
+    num_replicas = 1
+
+    def unconfigure_replica_as_agent(self, host):
+        """ Remove a replica from the list of agents.
+
+        cn=adtrust agents,cn=sysaccounts,cn=etc,$BASEDN contains a list
+        of members representing the agents. Remove the replica principal
+        from this list.
+        This is a hack allowing to run multiple times
+        ipa-adtrust-install --add-agents
+        (otherwise if the replica is in the list of agents, it won't be seen
+        as a possible agent to be added).
+        """
+        remove_agent_ldif = textwrap.dedent("""
+             dn: cn=adtrust agents,cn=sysaccounts,cn=etc,{base_dn}
+             changetype: modify
+             delete: member
+             member: fqdn={hostname},cn=computers,cn=accounts,{base_dn}
+             """.format(base_dn=host.domain.basedn, hostname=host.hostname))
+        # ok_returncode =16 if the attribute is not present
+        tasks.ldapmodify_dm(self.master, remove_agent_ldif,
+                            ok_returncode=[0, 16])
+
+    def test_samba_config_file(self):
+        """Check that ipa-adtrust-install generates sane smb.conf
+        This is regression test for issue
+        https://pagure.io/freeipa/issue/6951
+        """
+        self.master.run_command(
+            ['ipa-adtrust-install', '-a', self.master.config.admin_password,
+             '--add-sids', '-U'])
+        res = self.master.run_command(['testparm', '-s'])
+        assert 'ERROR' not in (res.stdout_text + res.stderr_text)
+
+    def test_add_agent_not_allowed(self):
+        """Check that add-agents can be run only by Admins."""
+        user = "nonadmin"
+        passwd = "Secret123"
+        host = self.replicas[0].hostname
+        data_fmt = '{{"method":"trust_enable_agent","params":[["{}"],{{}}]}}'
+
+        try:
+            # Create a nonadmin user that will be used by curl
+            tasks.create_active_user(self.master, user, passwd,
+                                     first=user, last=user)
+            tasks.kinit_as_user(self.master, user, passwd)
+            # curl --negotiate -u : is using GSS-API i.e. nonadmin user
+            cmd_args = [
+                paths.BIN_CURL,
+                '-H', 'referer:https://{}/ipa'.format(host),
+                '-H', 'Content-Type:application/json',
+                '-H', 'Accept:applicaton/json',
+                '--negotiate', '-u', ':',
+                '--cacert', paths.IPA_CA_CRT,
+                '-d', data_fmt.format(host),
+                '-X', 'POST', 'https://{}/ipa/json'.format(host)]
+            res = self.master.run_command(cmd_args)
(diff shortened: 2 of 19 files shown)


abbra:
Thank you for this PR, @flo-renaud. I haven't startted reviewing it yet but I'd like to ask you to move `ipa-trust-enable-agent` tool into `/usr/libexec/ipa`. We shouldn't really prompt it to run directly. Additionally, if it is part of `/usr/sbin`, it needs to have a man page. Recently, we have removed dependency on custodia package which only provided the same thing -- a `/usr/sbin`-based binary while we relied on internal helpers placed elsewhere.

abbra:
Pylint errors:
```
./ipatests/test_integration/test_adtrust_install.py:34:12: E128 continuation line under-indented for visual indent
./ipatests/test_integration/test_adtrust_install.py:34:28: E231 missing whitespace after ','
```

and also errors in `makeapi`:
```
cd .; /usr/bin/python ./makeapi --validate
Command trust_enable_agent/1 in ipalib, not in API

There are one or more new commands defined.
Update API.txt and increment the minor version in VERSION.
make: *** [Makefile:1259: apilint] Error 2
```

install/oddjob/org.freeipa.server.trust-enable-agent
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec /usr/libexec/ipa/ipa-trust-enable-agent "$1" 2>&1
```
abbra: What about moving the whole helper into this place instead of using, effectively, a double wrapper?

flo-renaud:
Hi @abbra 
thanks for the review. I modified install/oddjob/org.freeipa.server.trust-enable-agent per your suggestion (it is now the python script that was previously in /usr/libexec/ipa/ipa-trust-enable-agent).

abbra:
@flo-renaud thanks. Tests look ok to me.

Few comments to the logic itself:
- typically we keep these internal commands as not available to CLI with `NO_CLI = True` in the class. I think we should treat this one the same way.
- access controls are missing. Since this code modifies entries in `cn=config`, we should have a check whether a particular privilege is allowed for the user calling it. Please see `server_conncheck` implementation in `ipaserver/plugins/server.py`, it also has `NO_CLI = True` example.

abbra:
@flo-renaud sorry, I missed that there is already `NO_CLI = True`! Please ignore that part. The access control is still needed.

abbra:
And then a check for unprivileged user trying to access the command should be added to a test.

abbra:
Looks good to me now. Please remove the temp commit.

@rcritten could you please double check.

The only issue I see (not in this patchset but in general) -- how we should handle the API version for ipa-4-6 backport. This is the same issue like with https://github.com/freeipa/freeipa/pull/3487 (hostgroup renames)...

ipaserver/install/adtrust.py
```diff
@@ -344,13 +345,65 @@ def add_new_adtrust_agents(api, options):
     if new_agents:
         add_hosts_to_adtrust_agents(api, new_agents)
 
-        print("""
+        kwargs = {u'version': u'2.236',
+                  u'enable_compat': options.enable_compat}
```
rcritten: Should you add a comment that this particular version is being used because this is the version the command was added in? It was obvious once I looked at API.txt but I had the whole diff in front of me. I wonder if this will become some magic value in the future?

ipaserver/install/adtrust.py
```diff
@@ -344,13 +345,65 @@ def add_new_adtrust_agents(api, options):
...
+        failed_agents = []
+        for agent in new_agents:
+            # Try to run the ipa-trust-enable-agent script on the agent
+            # If the agent is too old and does not support this,
+            # print a msg
+            logger.info("Execute trust_enable_agent on remote server")
```
rcritten: Would it be clearer to report which agent (server) is being connected to? This same message will print for each one, right?

flo-renaud:
Hi @abbra @rcritten 
thanks for the reviews. I added a comment explaining the requested api version and also logged the remote node name. A successful run of the test can be seen [here](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/c3128d66-5937-11ea-ba13-fa163e00915b/report.html).
Now removing the temp commit.

abbra:
ACK

flo-renaud:
I am removing the ack so that this does not get pushed until this PR is synchronized with @wladich 's PR #4260

flo-renaud:
@wladich could you have a look at the tests?

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -0,0 +1,190 @@
...
+        https://pagure.io/freeipa/issue/6951
+        """
+        self.master.run_command(
+            ['ipa-adtrust-install', '-a', self.master.config.admin_password,
+             '--add-sids', '-U'])
+        res = self.master.run_command(['testparm', '-s'])
+        assert 'ERROR' not in (res.stdout_text + res.stderr_text)
+
+    def test_add_agent_not_allowed(self):
+        """Check that add-agents can be run only by Admins
```
wladich: For single line docstrings closing quotes should be on the same line

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        user = "nonadmin"
+        passwd = "Secret123"
+        tasks.create_active_user(self.master, user, passwd,
```
wladich: Created user should be removed in teardown.
Also could you please explain why it is needed? Does `curl` somehow use the kerberos ticket? If yes, please add the comment as it is hacky and non-obvious

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -0,0 +1,190 @@
...
+    num_replicas = 1
+
+    def unconfigure_replica_as_agent(self, host):
+        """ Remove a replica from the list of agents.
+
+        cn=adtrust agents,cn=sysaccounts,cn=etc,$BASEDN contains a list
+        of members representing the agents. Remove the replica principal
+        from this list.
+        """
+        remove_agent_ldif = textwrap.dedent("""
```
wladich: host has an ldap_connect method which returns LDAPClient object which allows to do LDAP modifications in a more elegant way without formatted diffs. Can it be used here?

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -0,0 +1,190 @@
...
+            '-H', 'Content-Type:application/json',
+            '-H', 'Accept:applicaton/json',
+            '--negotiate', '-u', ':',
+            '--cacert', paths.IPA_CA_CRT,
+            '-d', data_fmt.format(host),
+            '-X', 'POST', 'https://{}/ipa/json'.format(host)]
+        res = self.master.run_command(cmd_args)
+        tasks.kinit_admin(self.master)
+        expected_re = 'Insufficient access: not allowed to remotely add agent'
+        assert re.search(expected_re, res.stdout_text)
```
wladich: No need to use re.search here as the string does not contain any regexp operators

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -0,0 +1,190 @@
...
+from ipapython.dn import DN
+from ipatests.pytest_ipa.integration import tasks
+from ipatests.test_integration.base import IntegrationTest
+
+
+class TestIpaAdTrustInstall(IntegrationTest):
+    topology = 'line'
+    num_replicas = 1
+
+    def unconfigure_replica_as_agent(self, host):
```
wladich: Is this method intended as a "cheap" cleanup instead of topology reinstall? If yes, please add a phrase in docstring, as  it the method looks "hacky" an not obvious.

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -0,0 +1,190 @@
...
+            # access information about trusts.
+            # If you choose to do so, you also need to restart LDAP service on
+            # those masters.
+            # Refer to ipa-adtrust-install(1) man page for details.
+            # IPA master[replica1.testrelm.test]?[no]:
+            'yes\n'
+        )
+        res = self.master.run_command(['ipa-adtrust-install', '--add-agents'],
+                                      stdin_text=cmd_input)
+        self.replicas[0].run_command(['ipactl', 'start'])
```
wladich: if `ipa-adtrust-install` fails then replica will not be started and next tests will fail. It is better to place it in the `finally` block

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -0,0 +1,190 @@
...
+            # those masters.
+            # Refer to ipa-adtrust-install(1) man page for details.
+            # IPA master[replica1.testrelm.test]?[no]:
+            'yes\n'
+        )
+        expected_re = '"ipactl restart".+"systemctl restart sssd"'
+        res = self.master.run_command(['ipa-adtrust-install', '--add-agents'],
+                                      stdin_text=cmd_input)
+        # The replica must have been restarted automatically, no msg required
+        assert not re.search(expected_re, res.stdout_text, re.DOTALL)
```
wladich: I suggest to reduce search string to "ipactl restart", otherwise if we have regression of https://pagure.io/freeipa/issue/8148 then this test will always be green.

(Note: if search string is simplified, re.search can be replaced with `in`)

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -0,0 +1,190 @@
...
+            # about users from trusted forests.
+            # Installer can add them to the list of IPA masters allowed to
+            # access information about trusts.
+            # If you choose to do so, you also need to restart LDAP service on
+            # those masters.
+            # Refer to ipa-adtrust-install(1) man page for details.
+            # IPA master[replica1.testrelm.test]?[no]:
+            'yes\n'
+        )
+        expected_re = '"ipactl restart".+"systemctl restart sssd"'
```
wladich: I suggest to reduce search string to "ipactl restart", otherwise if we have regression of https://pagure.io/freeipa/issue/8148 then this test will always be green.

(Note: if search string is simplified, re.search can be replaced with in)

ipatests/test_integration/test_adtrust_install.py
flo-renaud: OK, updated.

ipatests/test_integration/test_adtrust_install.py
flo-renaud: OK, added a comment and also try/finally block.

ipatests/test_integration/test_adtrust_install.py
flo-renaud: The code is more compact with tasks.ldapmodify_dm IMO.

ipatests/test_integration/test_adtrust_install.py
flo-renaud: OK, updated.

ipatests/test_integration/test_adtrust_install.py
flo-renaud: OK, added a comment

ipatests/test_integration/test_adtrust_install.py
flo-renaud: OK, updated.

ipatests/test_integration/test_adtrust_install.py
flo-renaud: OK, good idea.

ipatests/test_integration/test_adtrust_install.py
flo-renaud: OK, good idea.

wladich:
LGTM.
Waiting for tests to complete.

wladich:
ACK

flo-renaud:
master:

* 68c72e344a29e27416af428f6dbf5300c85e66e1 Privilege: add a helper checking if a principal has a given privilege
* 911992b8bf33b40f650ec406444ca8b9b847b54e ipa-adtrust-install: run remote configuration for new agents
* fc4c3ac795e3af48fcfd8dd51085f5ff98047f1e ipatests: add test for ipa-adtrust-install --add-agents

PR #4278, opened 2020-02-24
Fix divide-by-zero when svc weight is 0 for all masters in location

rcritten:
Fix div-by-zero when svc weight is 0 for all masters in location
    
The relative service weight output tries to show the relative
chance that any given master in a locaiton will be picked. This
didn't account for all masters having a weight of 0 which would
result in a divide-by-zero error.
    
Implement the following rules:
1. If all masters have weight == 0 then all are equally
   weighted.
2. If any masters have weight == 0 then they have an
   extremely small chance of being chosen, percentage is
   0.1.
3. Otherwise it's percentage change is based on the sum of
   the weights of non-zero masters.
    
https://pagure.io/freeipa/issue/8135

### DIFF
--- a/ipaserver/plugins/location.py
+++ b/ipaserver/plugins/location.py
@@ -224,10 +224,30 @@ class location_show(LDAPRetrieve):
                 if 'DNS server' in s_roles:
                     dns_servers.append(s_name)
 
+            # 1. If all masters have weight == 0 then all are equally
+            #    weighted.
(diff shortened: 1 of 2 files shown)


abbra:

abbra:
ACKed.

rcritten:
master:

* f589a8952c33a25794e577077bb3c0bda740667c Fix div-by-zero when svc weight is 0 for all masters in location

PR #4292, opened 2020-03-02
ipatests: Test if server installer lock Bind9 recursion

mrizwan93:
This test is to check if recursion can be configured.
It checks if newly added file /etc/named/ipa-ext.conf
exists and /etc/named.conf should not have
'allow-recursion { any; };'. It also checks if ipa-backup
command backup the /etc/named/ipa-ext.conf file as well

related : https://pagure.io/freeipa/issue/8079

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -842,6 +842,30 @@ class TestInstallMasterDNS(IntegrationTest):
             extra_args=['--zonemgr', 'me@example.org'],
         )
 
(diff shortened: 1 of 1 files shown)


rcritten:
Looks good. please remove the temp commit and I'll ack.

I'm adding the ipa-4-8 label for backport, I assume we want this there. If not please remove it and add the ipa-next label instead.

rcritten:
master:

* 1556f3f767c3fab7634b46a9193e363887558db0 Test if server installer lock Bind9 recursion

PR #4299, opened 2020-03-03
Test for output being indented by default.

miskopo:
Test for output being indented by default value if not stated implicitly.
     
Test checks whether output json-line string is indented by default value
if this value is not stated implicitly. Test compares healthcheck
produced json-like string with manually indented one.
     
Automates: 02272ff39d76f1412483c5e3289564c93d196a03

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -7,12 +7,12 @@ Tests to verify that the ipa-healthcheck scenarios
 
 from __future__ import absolute_import
 
-from ipatests.test_integration.base import IntegrationTest
-from ipatests.pytest_ipa.integration import tasks
+import json
+
 from ipalib import api
 from ipapython.ipaldap import realm_to_serverid
-
-import json
+from ipatests.pytest_ipa.integration import tasks
+from ipatests.test_integration.base import IntegrationTest
 
 HEALTHCHECK_LOG = "/var/log/ipa/healthcheck/healthcheck.log"
 HEALTHCHECK_SYSTEMD_FILE = (
@@ -107,6 +107,7 @@ DEFAULT_PKI_KRA_CERTS = [
     "storageCert cert-pki-kra",
     "auditSigningCert cert-pki-kra",
 ]
+
 
 def run_healthcheck(host, source=None, check=None, output_type="json"):
     """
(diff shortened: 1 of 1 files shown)


rcritten:
I like your thinking on this but I don't see why you are splitting on comma. I think you can achieve what you want with:

output = self.master.run_command(["ipa-healthcheck"], raiseonerr=False)
json_out = output.stdout_text.strip()  # strip off trailing \n from output
assert json_out == json.dumps(json.loads(json_out), indent=2)

miskopo:
@rcritten  Thank you for your comment. I added split because the output is encapsulated in "[...]", thus needed to parse singular jsons from it. Now I see I forgot to take into account there could sometimes be only one json structure inside "[...]"

rcritten:
The whole string is in fact json. The output is a list of dictionaries. Re-parsing and re-generating the output to ensure that the indention is as expected is the correct thing to do (if inefficient).

You can also limit this to a single source to make the execution a bit faster and the output smaller, like --source ipahealthcheck.meta.services.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -421,6 +421,20 @@ def test_ipa_healthcheck_remove(self):
...
+        to output without it being implicitly stated
+        """
+        output = self.master.run_command(["ipa-healthcheck"], raiseonerr=False)
+        output_list_of_jsons = output.stdout_text.split(",")
+        if not len(output_list_of_jsons):  # there is only one json in list
+            output_list_of_jsons = [output.stdout_text.strip("[]")]
+        for output_json in output_list_of_jsons:
+            if output_json:
+                assert str(json.dumps(output_json.strip(),
+                                      indent=2)) == output_json.strip()
```
rcritten: The output of healthcheck is a single JSON object. There is no need to each entry separately.

It is not possible for ipa-healthcheck to return an empty list in the default case.

This should use the newish run_healthcheck method. I suggested previously to add a raw option so the unprocessed JSON can be returned.

miskopo: @rcritten so, if I understand correctly, as the output is begins with "[" and ends with "]", I should add "{" and "}" and read it as json?

rcritten:
The changes look good but this latest patch accidentally removes a lot of existing tests.

I tested this on my local system after restoring the lost tests and it passed so restore the tests, let CI pass it, then we can remove temp commit and ack.

miskopo:
CI passed, removing temp commit.
Run available at [AWS ](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/2eea8cd2-757b-11ea-8409-fa163e97fbb7/)

miskopo:
> The changes look good but this latest patch accidentally removes a lot of existing tests.
> 
> I tested this on my local system after restoring the lost tests and it passed so restore the tests, let CI pass it, then we can remove temp commit and ack.

I have pretty old heater at my place, it's possible that CO leakage is causing me short-time memory loss, because I have no memory of removing other tests.    \s

rcritten:
master:

* f5f960ed2ab08208b52260ab2dbe99da500ee04e Test for output being indented by default value if not stated implicitly.

PR #4301, opened 2020-03-04
ipatests: fix TestSubCAkeyReplication

flo-renaud:
The test is using the output of openssl to compare the SubCA issuer name
with the expected value.
Depending on the version of openssl, the issuer can be displayed
differently (with/without space around the = character). On RHEL 7.x,
there is no space by default while on Fedora the space is used.
Calling openssl with -nameopt space_eq forces a consistent output, always
adding space around =.


menonsudhir:
@flo-renaud LGTM.

flo-renaud:
master:

* 9ee8657c2a76a418e56baee50b07f6ec089e622e ipatests: fix TestSubCAkeyReplication

PR #4302, opened 2020-03-04
ipatest:Test if getcert creates cacert file with -F option

mrizwan93:
It took longer to create the cacert file in older version.
restarting the certmonger service creates the file at the location
specified by -F option. This fix is to check that cacert file
creates immediately after certificate goes into MONITORING state.

related: https://pagure.io/freeipa/issue/8105

### DIFF
--- a/ipatests/test_integration/test_cert.py
+++ b/ipatests/test_integration/test_cert.py
@@ -1,0 +1,48 @@
+#
+# Copyright (C) 2020  FreeIPA Contributors see COPYING for license
+#
+
+"""
+Module provides tests which testing ability of various certificate
+related scenarios.
+"""
+
+from __future__ import absolute_import
+
+import re
+
+from ipatests.pytest_ipa.integration import tasks
+from ipatests.test_integration.base import IntegrationTest
+
+
+class TestInstallMasterClient(IntegrationTest):
+    num_clients = 1
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master)
+        tasks.install_client(cls.master, cls.clients[0])
+
+    def test_cacert_file_appear_with_option_F(self):
+        """Test if getcert creates cacert file with -F option
+
+        It took longer to create the cacert file in older version.
+        restarting the certmonger service creates the file at the location
+        specified by -F option. This fix is to check that cacert file
+        creates immediately after certificate goes into MONITORING state.
+
+        related: https://pagure.io/freeipa/issue/8105
+        """
+        cmd_arg = ['ipa-getcert', 'request',
+                   '-f', '/etc/pki/tls/certs/test.pem',
+                   '-k', '/etc/pki/tls/private/test.key',
+                   '-K', 'test/%s' % self.clients[0].hostname,
+                   '-F', '/etc/pki/tls/test.CA']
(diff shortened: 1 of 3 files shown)


ipatests/test_integration/test_installation.py
```diff
@@ -1208,3 +1208,36 @@ def test_replica_install_against_server(self):
...
+        cmd_arg = ['ipa-getcert request',
+                   '-f', '/etc/pki/tls/certs/test.pem',
+                   '-k', '/etc/pki/tls/private/test.key',
+                   '-K', 'test/`hostname`', '-F',
+                   '/etc/pki/tls/test.CA']
+        result = self.clients[0].run_command([cmd_arg])
+        request_id = re.findall(r'\d+', result.stdout_text)
+
+        # check if certificate is in MONITORING state
+        status = wait_for_request(self.master, request_id[0], 50)
```
rcritten: self.master -> self.clients[0]

ipatests/test_integration/test_installation.py
```diff
@@ -1208,3 +1208,36 @@ def test_replica_install_against_server(self):
...
+        """Test if getcert creates cacert file with -F option
+
+        It took longer to create the cacert file in older version.
+        restarting the certmonger service creates the file at the location
+        specified by -F option. This fix is to check that cacert file
+        creates immediately after certificate goes into MONITORING state.
+
+        related: https://pagure.io/freeipa/issue/8105
+        """
+        cmd_arg = ['ipa-getcert request',
```
rcritten: 'ipa-getcert', 'request',

ipatests/test_integration/test_installation.py
```diff
@@ -1208,3 +1208,36 @@ def test_replica_install_against_server(self):
                                             self.replicas[0].hostname],
                                            stdin_text=dirman_password)
         assert self.replicas[0].hostname not in cmd.stdout_text
+
+
+class TestInstallMasterClient(IntegrationTest):
```
kaleemsiddiqu: This scenario is not related with installation, so IMO should not be part of test_installation.py. We can add a new test file test_cert.py which can contain cert related tests.

mrizwan93: done

ipatests/test_integration/test_cert.py
```diff
@@ -0,0 +1,48 @@
...
+        restarting the certmonger service creates the file at the location
+        specified by -F option. This fix is to check that cacert file
+        creates immediately after certificate goes into MONITORING state.
+
+        related: https://pagure.io/freeipa/issue/8105
+        """
+        cmd_arg = ['ipa-getcert', 'request',
+                   '-f', '/etc/pki/tls/certs/test.pem',
+                   '-k', '/etc/pki/tls/private/test.key',
+                   '-K', 'test/`hostname`', '-F',
```
rcritten: You'll need to set the hostname using self.clients[0].hostname

rcritten:
I'd prefer it if you completely remove the test_installation test in the first patch so it doesn't have to be removed in the unrelated "move wait_for_requests" patch.

rcritten:
Looks good, thanks. Please remove the temp commit and I'll ack.

PR #124, opened 2016-09-27
Fix: find OSCP certificate test

mbasti-rh:
Test should check if any OSCP certificate has been returned

https://fedorahosted.org/freeipa/ticket/6359

PR #4303, opened 2020-03-04
Test for healthcheck being run on replica with stopped master

miskopo:
Test checks whether healthcheck completes successfully from replica even
when IPA services on master are stopped.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1727900

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -7,13 +7,15 @@ Tests to verify that the ipa-healthcheck scenarios
 
 from __future__ import absolute_import
 
+import json
+import re
+
+import pytest
+
 from ipalib import api
 from ipapython.ipaldap import realm_to_serverid
 from ipatests.pytest_ipa.integration import tasks
 from ipatests.test_integration.base import IntegrationTest
-
-import json
-import pytest
 
 HEALTHCHECK_LOG = "/var/log/ipa/healthcheck/healthcheck.log"
 HEALTHCHECK_SYSTEMD_FILE = (
@@ -542,6 +544,27 @@ class TestIpaHealthCheck(IntegrationTest):
         output_str = cmd.stdout_text
         output_json = json.loads(output_str)
         assert output_str == "{}\n".format(json.dumps(output_json, indent=2))
+
+    @pytest.fixture
+    def ipactl(self):
+        """Stop and start IPA during test"""
+        self.master.run_command(["ipactl", "stop"])
+        yield
+        self.master.run_command(["ipactl", "start"])
+
+    def test_run_with_stopped_master(self, ipactl):
+        """
+        Test output of healthcheck where master IPA services are stopped
+        contains only errors regarding master being stopped and no other false
+        positives.
+        """
+        returncode, output = run_healthcheck(
+            self.master,
+            output_type="human",
+            failures_only=True)
+        assert returncode == 1
+        errors = re.findall("ERROR: .*: not running", output)
+        assert len(errors) == len(output.split('\n'))
 
     def test_ipa_healthcheck_remove(self):
         """


ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -494,3 +494,19 @@ def test_ipa_certs_check_ipacertnsstrust(self):
...
+class TestIpaHealthCheckOnReplica(TestIpaHealthCheck):
+
+    def __init__(self):
+        super(TestIpaHealthCheckOnReplica, self).__init__()
+
+    def test_run_with_stopped_master(self):
+        """
+        Test output of healthcheck where master IPA services are stopped.
+        """
+        self.master.run_command(["ipactl stop"], raiseonerr=True)
```
rcritten: This should probably be ["ipactl", "stop"]

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -494,3 +494,19 @@ def test_ipa_certs_check_ipacertnsstrust(self):
...
+        self.replicas[0].run_command(["ipa-healthcheck",
+                                      "--failures-only",
+                                      "--output-type human"],
+                                     raiseonerr=True)
```
rcritten: I think the sequence should be:

ipa-healthcheck --failures-only
ipactl stop
ipa-healthcheck --failures-only

This will catch false positives to show that the health is bad before even starting the test and it isn't a false positive.

Or if you want to get fancier you can run:

ipa-healthcheck --failures-only --source ipahealthcheck.meta.services --output-type human | grep -c 'not running'

And then assert that the value is the number of expected services (9).

raiseonerr=True is not technically necessary since it's the default but it's not wrong to include it here since you are specifically expecting an exception to be raised, you just aren't catching that exception and ensuring it is what you expct.

miskopo: so, if I understand correctly, I could do:

```
self.master.run_command(["ipactl stop"])
out = self.master.run_command(["ipa-healthcheck", "--failures-only", "--source", "ipahealthcheck.meta.services",  "--output-type",  "human" , "|",  "grep", " -c", " 'not running'"])
assert out.stdout == "9"
```

is 9 constant? If not, is it possible to obtain it?

rcritten: Good point, it depends on what optional services are installed.

Since are are just ensuring that healthcheck will only execute the status commands and not throw a ton of false positives a regex to ensure that all lines are something like:

^ERROR: .*: not running$

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -494,3 +495,28 @@ def test_ipa_certs_check_ipacertnsstrust(self):
...
+        tasks.install_master(cls.master, setup_dns=True)
+        tasks.install_replica(cls.master, cls.replicas[0], setup_dns=True)
+
+    def test_run_with_stopped_master(self):
+        """
+        Test output of healthcheck where master IPA services are stopped
+        contains only errors regarding master being stopped and no other false
+        positives.
+        """
+        self.master.run_command(["ipactl", "stop"])
```
rcritten, line 515: This will leave IPA stopped once the test is done. Can you add a fixture to stop then restart IPA?

rcritten:
Please remove the references to commits in the commit message and add a reference to the BZ that this is testing.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -494,3 +495,28 @@ def test_ipa_certs_check_ipacertnsstrust(self):
...
+        result = self.replicas[0].run_command(["ipa-healthcheck",
+                                               "--failures-only",
+                                               "--output-type",
+                                               "human"],
+                                              raiseonerr=True)
```
rcritten, line 520: This invocation can soon be replaced with:

    returncode, output = run_healthcheck(
                self.master,
                output_type="human",
                failures_only = True
            )

This change depends on a change added in https://github.com/freeipa/freeipa/pull/4323

miskopo: Regarding output-type, I run into following issue: [#124](https://github.com/freeipa/freeipa-healthcheck/issues/124 )

rcritten: I can't reproduce this.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -580,3 +581,28 @@ def test_ipa_certs_check_ipacertnsstrust(self):
...
+        positives.
+        """
+        self.master.run_command(["ipactl", "stop"])
+        returncode, output = run_healthcheck(
+            self.master,
+            output_type="human",
+            failures_only=True
+        )
+        errors = re.findall("^ERROR: .*: not running$", output)
+        assert len(errors) == len(output.split('\n'))
```
rcritten, line 608: Working on a proposal. I think the current patch just needs some tweaking to pass. What I'm doing is:

- dropping as a separate class and integrating into TestIpaHealthcheck
- adding a fixture to restart IPA once the test is complete
- dropping the ^ and $ from the output and replacing \n with '' so the re will succeed

rcritten:
Here is what I came up with after rebasing. It passes for me.

```
diff --git a/ipatests/test_integration/test_ipahealthcheck.py b/ipatests/test
_integration/test_ipahealthcheck.py
index 11e7a997a..2d83af639 100644
@@ -14,6 +14,7 @@ from ipatests.test_integration.base import Integration
Test
 
 import json
 import pytest
+import re
 
 HEALTHCHECK_LOG = "/var/log/ipa/healthcheck/healthcheck.log"
 HEALTHCHECK_SYSTEMD_FILE = (
@@ -543,6 +544,28 @@ class TestIpaHealthCheck(IntegrationTest):
         output_json = json.loads(output_str)
         assert output_str == "{}\n".format(json.dumps(output_json, indent=2))
 
+    @pytest.fixture
+    def ipactl(self):
+        """Stop and start IPA during test"""
+        self.master.run_command(["ipactl", "stop"])
+        yield
+        self.master.run_command(["ipactl", "start"])
+
+    def test_run_with_stopped_master(self, ipactl):
+        """
+        Test output of healthcheck where master IPA services are stop
ped
+        contains only errors regarding master being stopped and no ot
her false
+        positives.
+        """
+        returncode, output = run_healthcheck(
+            self.master,
+            output_type="human",
+            failures_only=True
+        )
+        assert returncode == 1
+        errors = re.findall("ERROR: .*: not running", output)
+        assert len(errors) == len(output.split('\n'))
+
     def test_ipa_healthcheck_remove(self):
         """
         This testcase checks the removal of of healthcheck tool
@@ -604,19 +627,3 @@ class TestIpaHealthCheckWithoutDNS(IntegrationTest)
:
                 check["kw"]["key"] in DEFAULT_PKI_KRA_CERTS
             )
         tasks.uninstall_master(self.master)
-
-
-class TestIpaHealthCheckOnReplica(TestIpaHealthCheck):
-
-    def __init__(self):
-        super(TestIpaHealthCheckOnReplica, self).__init__()
-
-    def test_run_with_stopped_master(self):
-        """
-        Test output of healthcheck where master IPA services are stopped.
-        """
-        self.master.run_command(["ipactl stop"], raiseonerr=True)
-        self.replicas[0].run_command(["ipa-healthcheck",
-                                      "--failures-only",
-                                      "--output-type human"],
-                                     raiseonerr=True)
```

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -604,3 +606,34 @@ def test_ipa_certs_check_ipacertnsstrust(self):
                 check["kw"]["key"] in DEFAULT_PKI_KRA_CERTS
             )
         tasks.uninstall_master(self.master)
+
+
+class TestIpaHealthCheckOnReplica(IntegrationTest):
```
rcritten, line 611: I don't think this needs a separate class (and accompanying server installs). I think this can be added to TestIpaHealthCheck before test_ipa_healthcheck_remove.

Otherwise looks good.

netoarmando:
`temp_commit` job is timing out (and failing to report it to github), here's one execution output: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/80d821f6-801d-11ea-bda8-fa163e07b2f0/runner.log.gz
Please, increase the timeout argument.

rcritten:
Increased to 4800.

rcritten:
LGTM, tests are green, please remove the temporary commit.

miskopo:
Removing temp commit, run available at [AWS](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/6c2ce2e4-8029-11ea-8171-fa163e2d2908/)

miskopo:
@rcritten Ready for ACKing

rcritten:
master:

* 3a64ac08e1b3202bb469fcf0f2a17bd5114bab8f Test for healthcheck being run on replica with stopped master

PR #4317, opened 2020-03-06
Test that ipa-healthcheck human output translates error strings

rcritten:
The code rather than the string was being displayed in human
output for non-SUCCESS messages. Verify that in case of an error
the right output will be present.
    
https://bugzilla.redhat.com/show_bug.cgi?id=1752849

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -108,11 +108,15 @@ DEFAULT_PKI_KRA_CERTS = [
 ]
 
 
-def run_healthcheck(host, source=None, check=None):
+def run_healthcheck(host, source=None, check=None, output_type="json"):
     """
     Run ipa-healthcheck on the remote host and return the result
 
-    Returns: the tuple returncode, json (if any)
+    Returns: the tuple returncode, output
+
+    output is:
+        json data if output_type == "json"
+        stdout if output_type == "human"
     """
     data = None
     cmd = ["ipa-healthcheck"]
@@ -124,10 +128,16 @@ def run_healthcheck(host, source=None, check=None):
             cmd.append("--check")
             cmd.append(check)
 
+    cmd.append("--output-type")
+    cmd.append(output_type)
+
     result = host.run_command(cmd, raiseonerr=False)
 
     if result.stdout_text:
-        data = json.loads(result.stdout_text)
+        if output_type == "json":
+            data = json.loads(result.stdout_text)
+        else:
+            data = result.stdout_text.strip()
 
     return result.returncode, data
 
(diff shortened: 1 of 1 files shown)


rcritten:
Is that an ack? Can you add the flag if so?

ssidhaye:
ACK. Changes look good.

flo-renaud:
@rcritten 
please remove the temp commit

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -143,6 +143,31 @@ def test_run_ipahealthcheck_list_source(self):
...
             assert source in result.stdout_text
 
+    def test_human_output(self):
+        """
+        Test that in human output the severity value is correct
+
+        Only the SUCCESS (0) value was being translated, otherwise
+        the numeric value was being shown (BZ 1752849)
+        """
+        self.master.run_command(["systemctl", "stop", "sssd"])
```
stanislavlevin: @rcritten.
May be too late... sorry.
If this test fails, then the next ones which rely on running `sssd` will fail too.
I suggest using the pytest's `fixture` here, something like:

```python
import pytest

@pytest.fixture
def sssd_stop():
    print('sssd stop')
    yield
    print('sssd start')

class TestMy:
    def test_1(self, sssd_stop):
        print('test_1')
```

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -143,6 +143,31 @@ def test_run_ipahealthcheck_list_source(self):
...
+                "--check",
+                "sssd",
+                "--output-type",
+                "human",
+            ],
+            raiseonerr=False,
+        )
+        assert result.returncode == 1
+        assert result.stdout_text.strip() == \
+            "ERROR: ipahealthcheck.meta.services.sssd: sssd: not running"
```
stanislavlevin: IMO, this assertion is very brittle, because the error message may be changed(?) in `ipa-healthcheck` unless it's guaranteed to be so. May be regular exprs?

rcritten: It's a sister project so hopefully we'll communicate. It isn't impossible for the string to change but is not very likely.

ipatests/test_integration/test_ipahealthcheck.py
rcritten: That's a good idea, I'll look into it.

rcritten:
Removing ack so this doesn't get pushed while I look into adding the fixture.

ipatests/test_integration/test_ipahealthcheck.py
rcritten: I need to pass in the master value in order to execute the stop/start. I could do that via a context but it seems like it would be clearer if I just moved the restart up so that it executes right after ipa-healthcheck which can fail but won't raise an exception so sssd should always be restarted.

stanislavlevin: alternatively, a fixture may be defined as a method:
```python
class TestMy:
    master='Master'

    @pytest.fixture
    def sssd_stop(self):
        print('sssd stop on ', self.master)
        yield
        print('sssd start on ', self.master)

    def test_1(self, sssd_stop):
        print('test_1')
```
But I'm happy with the current version.

stanislavlevin: @rcritten, the exception could be other than `CalledProcessError`. That's why the fixture is preferred.

rcritten: Yeah, I guess so, but it is just calling exec so I'd find it extremely unlikely to be anything else.
And anything that might return something else would be so catastrophically bad that not being able to restart sssd for this test case would be the least of our issues.

tiran: You could also wrap the call into ``try: run_healthcheck(); finally: start sssd`` block.

rcritten: added try/except/else/finally

stanislavlevin: as @tiran proposed it's enough `try->finally` (without except/else).

stanislavlevin:
I'm ok with the changes.
I rerun Azure because of fails are not related to this PR.

stanislavlevin:
@rcritten, please, don't forget to remove a temp commit.
Thank you.

flo-renaud:
master:

* 4a3b7baed785f7492e8404dac3eee8a8ce9fd937 Test that ipa-healthcheck human output translates error strings

PR #3645, opened 2019-09-10
ipatests: Added testcase to check that ipa-healthcheck tool displays warning if no range is set

menonsudhir:
ipatests: Added testcase to check that ipa-healthcheck tool displays warning if no range is set.
Issue: https://github.com/freeipa/freeipa-healthcheck/issues/60

PR #4323, opened 2020-03-08
Perform baseline healthcheck

rcritten:
Run healthcheck on a default installation and ensure that there
are no failures. This test ensures that a fresh IPA installation
will pass healthcheck.

https://bugzilla.redhat.com/show_bug.cgi?id=1774032

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -7,12 +7,13 @@ Tests to verify that the ipa-healthcheck scenarios
 
 from __future__ import absolute_import
 
-import json
-
 from ipalib import api
 from ipapython.ipaldap import realm_to_serverid
 from ipatests.pytest_ipa.integration import tasks
 from ipatests.test_integration.base import IntegrationTest
+
+import json
+import pytest
 
 HEALTHCHECK_LOG = "/var/log/ipa/healthcheck/healthcheck.log"
 HEALTHCHECK_SYSTEMD_FILE = (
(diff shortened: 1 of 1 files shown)


ssidhaye:
@rcritten Thanks for the test. Should we check if the output contains
Missing tracking for cert-database or Unknown certmonger id.
Otherwise the patch looks good.

rcritten:
I don't think we need to explicitly check for this specific error. This baseline test will ensure that no errors at all are returned on a default installation which also covers the specific problem raised in the BZ.

flo-renaud:
Hi @rcritten 
tests are green, please remove the temp commit.

rcritten:
This doesn't require a rebase but PR https://github.com/freeipa/freeipa/pull/3645 added a new way to execute the test so I'm going to migrate to that method. I need to get PR https://github.com/freeipa/freeipa/pull/4317 merged first.

Removing ack for now.

rcritten:
Tests passing after much flag issues. Adding back ack and pushing.

rcritten:
Hang on, probably worth re-review since I rebased and made another slight change, adding back needs_review.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -387,6 +392,20 @@ def test_source_ipa_roles_check_crlmanager(self):
             assert check["result"] == "SUCCESS"
             assert check["kw"]["key"] == "crl_manager"
             assert check["kw"]["crlgen_enabled"] is False
+        self.master.run_command(["ipa-crlgen-manage", "enable"])
```
flo-renaud, line 395: I would put this command in a `finally:` block to make sure it is called even if the test fails.

rcritten:
I implemented it as a fixture

flo-renaud:
Hi @rcritten 
thanks for the updated patch, LGTM and tests are green. Please remove the temp commit and I will ack.

rcritten:
master:

* 3022bb5fd200ea3b22fb1600401e95aaee2006ea Perform baseline healthcheck

PR #4324, opened 2020-03-09
ipatests:Test if schema-compat-entry-attribute is set

mrizwan93:
This is to ensure if said entry is set after installation.

related: https://pagure.io/freeipa/issue/8193

### DIFF
--- a/ipatests/test_integration/test_adtrust_install.py
+++ b/ipatests/test_integration/test_adtrust_install.py
@@ -199,3 +199,18 @@ class TestIpaAdTrustInstall(IntegrationTest):
         entry = conn.get_entry(DN(
             "cn=groups,cn=Schema Compatibility,cn=plugins,cn=config"))
         assert entry.single_value['schema-compat-lookup-nsswitch'] == "group"
+
+    def test_schema_compat_attribute(self):
+        """Test if schema-compat-entry-attribute is set
+
+        This is to ensure if said entry is set after installation with AD.
+
+        related: https://pagure.io/freeipa/issue/8193
+        """
+        conn = self.replicas[0].ldap_connect()
+        entry = conn.get_entry(DN(
+            "cn=groups,cn=Schema Compatibility,cn=plugins,cn=config"))
+        entry_list = list(entry['schema-compat-entry-attribute'])
+        value = (r'ipaexternalmember=%deref_r('
+                 '"member","ipaexternalmember")')
+        assert value in entry_list
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -20,6 +20,7 @@ from cryptography.hazmat.primitives import hashes
 
 from ipalib import x509
 from ipalib.constants import DOMAIN_LEVEL_0
+from ipapython.dn import DN
 from ipaplatform.constants import constants
 from ipaplatform.osinfo import osinfo
 from ipaplatform.paths import paths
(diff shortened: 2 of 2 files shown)


ipatests/test_integration/test_adtrust_install.py
```diff
@@ -199,3 +199,14 @@ def test_add_agent_on_running_replica_with_compat(self):
...
             "cn=groups,cn=Schema Compatibility,cn=plugins,cn=config"))
         assert entry.single_value['schema-compat-lookup-nsswitch'] == "group"
+
+        """Test if schema-compat-entry-attribute is set
+
+        This is to ensure if said entry is set after installation.
+
+        related: https://pagure.io/freeipa/issue/8193
+        """
+        entry_list = list(entry['schema-compat-entry-attribute'])
```
kaleemsiddiqu: @mrizwan93 
Check for compat tree enablement is missing.
Also this only perform check for installation with AD . We need to check without AD install as well.

mrizwan93: The check is already covered at https://github.com/freeipa/freeipa/pull/4324/files#diff-295adcfce4b5bafad3de8b3a88ca533dR201

I'll submit another commit for non AD environment.

ipatests/test_integration/test_installation.py
```diff
@@ -559,6 +560,21 @@ def install(cls, mh):
...
+        """Test if schema-compat-entry-attribute is set
+
+        This is to ensure if said entry is set after installation.
+
+        related: https://pagure.io/freeipa/issue/8193
+        """
+        conn = self.master.ldap_connect()
+        entry = conn.get_entry(DN(             # pylint: disable=no-member
+            "cn=groups,cn=Schema Compatibility,cn=plugins,cn=config"))
+
```
kaleemsiddiqu: @mrizwan93 
Check for compat tree disablement is missing here.

kaleemsiddiqu: Ignore, its down there.

ipatests/test_integration/test_installation.py
```diff
@@ -559,6 +560,21 @@ def install(cls, mh):
 
     def test_install_master(self):
         tasks.install_master(self.master, setup_dns=False)
+        """Test if schema-compat-entry-attribute is set
```
fcami: Could you please move this test into a dedicated test function? It makes it easier to xfail it if necessary.

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -199,3 +199,14 @@ def test_add_agent_on_running_replica_with_compat(self):
         entry = conn.get_entry(DN(
             "cn=groups,cn=Schema Compatibility,cn=plugins,cn=config"))
         assert entry.single_value['schema-compat-lookup-nsswitch'] == "group"
+
+        """Test if schema-compat-entry-attribute is set
+
```
fcami: Same remark as the non-AD case. Please move this to a dedicated test.

fcami: Also it'd be clearer if the test function names matched what the test did. This is not the case here with the additional test.

mrizwan93: @fcami Hello, 
I moved test as suggested for non AD installation however IMO, AD one should remain like this as we are interested in only one attribute that the `entry` variable providing already.

mrizwan93: Moved both as suggested.

kaleemsiddiqu:
LGTM.

fcami:
LGTM from me as well. No objection to ACK above.

tiran:
master:

* 9120d65e881e582cc37ee8445fb66a54f6fb4d75 Test if schema-compat-entry-attribute is set
* 312d00df90fc40c1c4c934945c3c73053b6ea18a Test if schema-compat-entry-attribute is set

PR #4331, opened 2020-03-10
Fixes for ipa-adtrust-install --add-agents

flo-renaud:
### ipa-adtrust-install: remote command fails if ipa-server-trust-ad pkg missing

When the command ipa-adtrust-install --add-agents is run, it executes
remotely the command trust_enable_agent. This command does not require
the package ipa-server-trust-ad to be installed on the remote node, but
fails if it's not the case because dbus is not imported.
Need to move the "import dbus" outside of the try/except related to
dcerpc import.

Related: https://pagure.io/freeipa/issue/7600

### selinux policy: add the right context for org.freeipa.server.trust-enable-agent

This commit sets the system_u:object_r:ipa_helper_exec_t:s0 context to the
oddjob script org.freeipa.server.trust-enable-agent.
Without this context, oddjob cannot launch the command
/usr/libexec/ipa/oddjob/org.freeipa.server.trust-enable-agent
when ipa-adtrust-install --add-agents is run with SElinux enforcing.

Related: https://pagure.io/freeipa/issue/7600

### DIFF
--- a/ipaserver/plugins/trust.py
+++ b/ipaserver/plugins/trust.py
@@ -18,6 +18,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import dbus
+import dbus.mainloop.glib
 import logging
 
 import six
@@ -74,8 +76,6 @@ except Exception as e:
 if api.env.in_server and api.env.context in ['lite', 'server']:
     try:
         import ipaserver.dcerpc
-        import dbus
-        import dbus.mainloop.glib
         _bindings_installed = True
     except ImportError:
         _bindings_installed = False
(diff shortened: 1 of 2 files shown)


flo-renaud:
The selinux policy fix needs to be backport to ipa-4-8 only but the other fix is required on all branches starting from ipa-4-6.

tiran:
The rearrange of ``import dbus`` should not cause any issues. Other packages like ``ipaserver.plugins.server`` already import dbus on top level.

abbra:
I guess `ipa-4-6` backport will be manual.

flo-renaud:
master:

* 233a18b2a24d814518a119bd3f1688a0eb361917 ipa-adtrust-install: remote command fails if ipa-server-trust-ad pkg missing
* 1fbc4e01ea5be004f7c57cb71df2d37659271d68 selinux policy: add the right context for org.freeipa.server.trust-enable-agent

PR #4343, opened 2020-03-11
certstore.get_ca_certs(): order CA certs

fcami:
Currently, get_ca_certs() returns a non-ordered list of CA
certificates. ipa-certupdate then writes that list to ca.crt.
However, ldapsearch and other tools expect the first certificate
in ca.crt to be valid. This is not the case if the first
cACertificate attribute contains an expired certificate.

get_ca_certs() will now insert in front of the list the current
certificates, and append the ones which are either not yet valid
or not valid anymore, making sure the first certificate in front
of the list is current.

Fixes: https://pagure.io/freeipa/issue/8223

### DIFF
--- a/ipalib/install/certstore.py
+++ b/ipalib/install/certstore.py
@@ -22,6 +22,8 @@
 LDAP shared certificate store.
 """
 
+import datetime
+
 from pyasn1.error import PyAsn1Error
 
 from ipapython.dn import DN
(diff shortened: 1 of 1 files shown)


fcami:
For the record, this is probably the wrong approach as it would require updating all older clients' CA cert retrieval codepath too. See the upstream issue for discussion.

fcami:
Closing, see upstream issue for reasoning.

PR #4349, opened 2020-03-12
ipa-restore: restart services at the end

fcami:
When IPA was not installed on the restore target host, and
when httpd was already running, "ipactl stop" does not stop
httpd. "ipactl start" at the end of the restore tool will
therefore not restart httpd either.
Calling "ipactl restart" at the end of the restore fixes the
issue, and as an added bonus, makes sure IPA can restart itself
properly.

Fixes: https://pagure.io/freeipa/issue/8226

### DIFF
--- a/ipaserver/install/ipa_restore.py
+++ b/ipaserver/install/ipa_restore.py
@@ -474,6 +474,14 @@ class Restore(admintool.AdminTool):
                 http.remove_httpd_ccaches()
                 # have the daemons pick up their restored configs
                 tasks.systemd_daemon_reload()
+                # Restart IPA a final time.
+                # Starting then restarting is necessary to make sure some
+                # daemons like httpd are restarted
+                # (https://pagure.io/freeipa/issue/8226).
+                logger.info('Restarting IPA services')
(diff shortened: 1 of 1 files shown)


tiran:
@fcami Please remove the temp commit

flo-renaud:
Hi @fcami 
another possible solution would be to force httpd stop when ipactl stop returns 6 (IPA not configured on this system) [at this line](https://github.com/freeipa/freeipa/blob/682b59c8e801f06b3c6e4aa97c1ab33390c9edf8/ipaserver/install/ipa_restore.py#L424) in ipa-restore. This would prevent calling twice ipactl restart.

fcami:
hi @flo I'd rather have the restart done. It makes sure IPA can restart properly. If any other service behaves the same way as httpd does, stopping only httpd will break the restore again.

flo-renaud:
Hi @fcami 
no problem with your solution, I would have removed the ACK if I was against it :)

fcami:
master:

* 1eb6a9bf16e02b16328091cb98b13358cf372ee1 ipa-restore: restart services at the end

PR #4355, opened 2020-03-13
Allow hosts to read DNS records for IP SAN

tiran:
For SAN IPAddress extension the cert plugin verifies that the IP address
matches the host entry. Certmonger uses the host principal to
authenticate and retrieve certificates. But the host principal did not
have permission to read DNS entries from LDAP.

Allow all hosts to read some entries from active DNS records.

Fixes: https://pagure.io/freeipa/issue/8098


abbra:
LGTM

rcritten:
You add another SAN for the host but it is currently unused. Is that some future change? Should it be removed since it is unused?

ipatests/test_integration/test_cert.py
```diff
@@ -46,3 +63,43 @@ def test_cacert_file_appear_with_option_F(self):
...
+            certdata, default_backend()
+        )
+        ext = cert.extensions.get_extension_for_oid(
+            ExtensionOID.SUBJECT_ALTERNATIVE_NAME
+        )
+        dnsnames = set(ext.value.get_values_for_type(x509.DNSName))
+        # assert dnsnames == {self.clients[0].hostname, self.altname}
+        assert dnsnames == {self.clients[0].hostname}
+        ipaddrs = set(ext.value.get_values_for_type(x509.IPAddress))
+        assert ipaddrs == {self.clients[0].ip}
```
abbra: The assert for `ipaddrs` fails because you get `str` in the set on the right side and `IPv4Address` instance on the lest side:
```
>       assert ipaddrs == {self.clients[0].ip}
E       AssertionError: assert {IPv4Address(...168.122.193')} == {'192.168.122.193'}
E         Extra items in the left set:
E         IPv4Address('192.168.122.193')
E         Extra items in the right set:
E         '192.168.122.193'
E         Full diff:
E         - {IPv4Address('192.168.122.193')}
E         + {'192.168.122.193'}
```
I think you either need to convert `self.clients[0].ip` to an IPAddress instance:
```
from ipaddress import ip_address
...
assert ipaddrs == {ip_address(self.clients[0].ip)}
```

abbra:
LGTM

tiran:
master:

* 7a9ac1f58694b28c96b8bdb22eaf506f7019ac65 Allow hosts to read DNS records for IP SAN

PR #4359, opened 2020-03-13
spec: Take the ownership over '/usr/libexec/ipa/custodia'

stanislavlevin:
Ideally, an every file on system has to have an owner.

'/usr/libexec/ipa/custodia' directory was added recently, but:

```
[root@dc ~]# LANG=C rpm -qf /usr/libexec/ipa/custodia/ipa-custodia-dmldap
freeipa-server-4.8.4-2.fc31.x86_64
[root@dc ~]# LANG=C rpm -qf /usr/libexec/ipa/custodia
file /usr/libexec/ipa/custodia is not owned by any package
```

ALTLinux build system warns about files or directories which were
'created' during a package installation but haven't an owner. So,
after the resyncing spec file to upstream's one my build fails.

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1109,10 +1109,6 @@ fi
 %{_sbindir}/ipa-cert-fix
 %{_libexecdir}/certmonger/dogtag-ipa-ca-renew-agent-submit
 %{_libexecdir}/certmonger/ipa-server-guard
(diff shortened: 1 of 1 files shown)


rcritten:
master:

* 9c9c6a70632cdb8f55b65a50c9a808007523666b spec: Take the ownership over '/usr/libexec/ipa/custodia'

PR #4373, opened 2020-03-16
Test to check if Certmonger tracks certs in between reboots/interruptions

ssidhaye:
Test to check if Certmonger tracks certs in between
reboots/interruptions and while in "CA_WORKING" state
    
When a resubmit request is submitted an "invalid cookie"
error message is no longer shown
    
Earlier an "invlaid cookie" error message was shown when getcert list was called.
    
The fix allows an empty cookie in dogtag-ipa-ca-renew-agent-submit
    
Pagure Issue: https://pagure.io/freeipa/issue/8164

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -2135,11 +2135,39 @@ def wait_for_request(host, request_id, timeout=120):
         )
 
         state = result.stdout_text.strip()
-        print("certmonger request is in state %r", state)
+        logger.info("certmonger request is in state %s", state)
         if state in ('CA_REJECTED', 'CA_UNREACHABLE', 'CA_UNCONFIGURED',
                      'NEED_GUIDANCE', 'NEED_CA', 'MONITORING'):
             break
         time.sleep(5)
+    else:
+        raise RuntimeError("request timed out")
+
+    return state
+
+
+def wait_for_certmonger_status(host, status, request_id, timeout=120):
+    """Aggressively wait for a specific certmonger status.
+
+       This checks the status every second in order to attempt to
+       catch transient states like SUBMITTED. There are no guarantees.
+
+       :param host: the host where the uninstallation takes place
+       :param status: tuple of statuses to look for
+       :param request_id: request_id of request to check status on
+       :param timeout: max time in seconds to wait for the status
+    """
+    for _i in range(0, timeout, 1):
+        result = host.run_command(
+            "getcert list -i %s | grep status: | awk '{ print $2 }'" %
+            request_id
+        )
+
+        state = result.stdout_text.strip()
+        logger.info("certmonger request is in state %s", state)
+        if state in status:
+            break
+        time.sleep(1)
     else:
         raise RuntimeError("request timed out")
 
--- a/ipatests/test_integration/test_cert.py
+++ b/ipatests/test_integration/test_cert.py
@@ -7,14 +7,50 @@ Module provides tests which testing ability of various certificate
 related scenarios.
 """
 import ipaddress
+import pytest
 import re
 
+from ipaplatform.paths import paths
 from cryptography import x509
 from cryptography.x509.oid import ExtensionOID
 from cryptography.hazmat.backends import default_backend
 
 from ipatests.pytest_ipa.integration import tasks
 from ipatests.test_integration.base import IntegrationTest
+
+DEFAULT_RA_AGENT_SUBMITTED_VAL = '19700101000000'
+
+
+def get_certmonger_fs_id(input_str):
+    """Get certmonger FS ID
+    from the `getcert list -f /var/lib/ipa/ra-agent.pem` output
+    command output
+
+    :return request ID string
+    """
+    request_id = re.findall(r'\d+', input_str)
+    return request_id[1]
+
+
+def get_certmonger_request_value(host, requestid, state):
+    """Get certmonger submitted value from
+    /var/lib/certmonger/requests/<timestamp>
+
+    :return submitted timestamp value
+    """
+    result = host.run_command(
+        ['grep', '-rl', 'id={0}'.format(requestid),
+         paths.CERTMONGER_REQUESTS_DIR]
+    )
+    assert result.stdout_text is not None
+    filename = result.stdout_text.strip()
+    request_file = host.get_file_contents(filename, encoding='utf-8')
+    val = None
+    for line in request_file.split('\n'):
+        if line.startswith('%s=' % state):
+            _unused, val = line.partition("=")[::2]
+            break
+    return val
 
 
 class TestInstallMasterClient(IntegrationTest):
@@ -90,3 +126,94 @@ class TestInstallMasterClient(IntegrationTest):
         assert dnsnames == [self.clients[0].hostname]
         ipaddrs = ext.value.get_values_for_type(x509.IPAddress)
         assert ipaddrs == [ipaddress.ip_address(self.clients[0].ip)]
+
+
+class TestCertmongerInterruption(IntegrationTest):
+    num_replicas = 1
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master)
+        tasks.install_replica(cls.master, cls.replicas[0])
+
+    def test_certmomger_tracks_renewed_certs_during_interruptions(self):
+        """Test that CA renewal handles early CA_WORKING and restarts
+
+        A non-renewal master CA might submit a renewal request before
+        the renewal master actually updating the certs. This is expected.
+        The tracking request will result in CA_WORKING.
+
+        This would trigger a different path within the IPA renewal
+        scripts which differentiate between a SUBMIT (new request) and
+        a POLL (resume request). The script was requiring a cookie
+        value for POLL requests which wasn't available and was
+        erroring out unrecoverably without restarting certmonger.
+
+        Submit a request for renewal early and wait for it to go into
+        CA_WORKING. Resubmit the request to ensure that the request
+        remains in CA_WORKING without reporting any ca_error like
+        Invalid cookie: ''
+
+        Use the submitted value in the certmonger request to validate
+        that the request was resubmitted and not rely on catching
+        the states directly.
+
+        Pagure Issue: https://pagure.io/freeipa/issue/8164
+        """
+        cmd = ['getcert', 'list', '-f', paths.RA_AGENT_PEM]
+        result = self.replicas[0].run_command(cmd)
+
+        # Get Request ID and Submitted Values
+        request_id = get_certmonger_fs_id(result.stdout_text)
+        start_val = get_certmonger_request_value(self.replicas[0],
+                                                 request_id, "submitted")
+
+        # at this point submitted value for RA agent cert should be
+        # 19700101000000 since it has never been submitted for renewal.
+        assert start_val == DEFAULT_RA_AGENT_SUBMITTED_VAL
+
+        cmd = ['getcert', 'resubmit', '-f', paths.RA_AGENT_PEM]
+        self.replicas[0].run_command(cmd)
+
+        tasks.wait_for_certmonger_status(self.replicas[0],
+                                         ('CA_WORKING', 'MONITORING'),
+                                         request_id)
+
+        resubmit_val = get_certmonger_request_value(self.replicas[0],
+                                                    request_id,
+                                                    "submitted")
+
+        if resubmit_val == DEFAULT_RA_AGENT_SUBMITTED_VAL:
+            pytest.fail("Request was not resubmitted")
+
+        ca_error = get_certmonger_request_value(self.replicas[0],
+                                                request_id, "ca_error")
+        state = get_certmonger_request_value(self.replicas[0],
+                                             request_id, "state")
+
+        assert ca_error is None
+        assert state == 'CA_WORKING'
+
+        cmd = ['getcert', 'resubmit', '-f', paths.RA_AGENT_PEM]
+        self.replicas[0].run_command(cmd)
+
+        tasks.wait_for_certmonger_status(self.replicas[0],
+                                         ('CA_WORKING', 'MONITORING'),
+                                         request_id)
+
+        resubmit2_val = get_certmonger_request_value(self.replicas[0],
+                                                     request_id,
+                                                     "submitted")
+
+        if resubmit_val == DEFAULT_RA_AGENT_SUBMITTED_VAL:
+            pytest.fail("Request was not resubmitted")
+
+        assert resubmit2_val > resubmit_val
+
+        ca_error = get_certmonger_request_value(self.replicas[0],
+                                                request_id, "ca_error")
+        state = get_certmonger_request_value(self.replicas[0],
+                                             request_id, "state")
+
+        assert ca_error is None
+        assert state == 'CA_WORKING'


ipatests/test_integration/test_cert.py
```diff
@@ -46,3 +47,50 @@ def test_cacert_file_appear_with_option_F(self):
...
         self.clients[0].run_command(['ls', '-l', '/etc/pki/tls/test.CA'])
+
+
+class TestInstallMasterReplica(IntegrationTest):
+    num_replicas = 1
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master)
+        tasks.install_replica(self.master, self.replicas[2], setup_ca=True,
```
mrizwan93: Only one replica is defined for the class. It should be `cls.replicas[0]`. Also you need to change `self.master` to `cls.master`

ipatests/test_integration/test_cert.py
```diff
@@ -46,3 +47,50 @@ def test_cacert_file_appear_with_option_F(self):
...
+                              setup_kra=False)
```
mrizwan93: No need `setup_ca=True` and `setup_kra=False` as these are default values anyways in the method definition.

ipatests/test_integration/test_cert.py
```diff
@@ -46,3 +47,50 @@ def test_cacert_file_appear_with_option_F(self):
...
+        the renewal master actually updating the certs. This is expected.
+
+        But, if certmonger on this node is stopped/interrupted
+        for any reason while the resubmit request is in "CA_WORKING" state
+        it will not be able to track changes to this cert anymore.
+
+        Pagure Issue: https://pagure.io/freeipa/issue/8164
+        """
+        cmd = ['getcert', 'resubmit', '-f',
+               '/var/lib/ipa/ra-agent.pem']
```
mrizwan93: `paths.RA_AGENT_PEM` can be used here.

ipatests/test_integration/test_cert.py
```diff
@@ -46,3 +47,50 @@ def test_cacert_file_appear_with_option_F(self):
...
+        result = self.replicas[0].run_command(cmd)
+        # assert 'CA_WORKING' in result.stdout_text
+        request_id = re.findall(r'\d+', result.stdout_text)
+
+        # check if certificate is in CA_WORKING state
+        status = tasks.wait_for_request(self.clients[0], request_id[0], 50)
```
mrizwan93: You might need to change `self.clients[0]` to `self.replicas[0]`

ipatests/test_integration/test_cert.py
```diff
@@ -46,3 +47,50 @@ def test_cacert_file_appear_with_option_F(self):
...
+        assert status == "CA_WORKING"
+
+        # restart certmonger service on replica
+        self.replicas[0].run_command(['systemctl', 'restart', 'certmonger'])
+        time.sleep(15)
+
+        cmd = ['getcert', 'list', '-f', '/var/lib/ipa/ra-agent.pem']
```
mrizwan93: `paths.RA_AGENT_PEM` can be used here.

ipatests/test_integration/test_cert.py
```diff
@@ -46,3 +47,50 @@ def test_cacert_file_appear_with_option_F(self):
...
+        result = self.replicas[0].run_command(cmd)
+        # check if certificate is in SUBMITTING state
+        status = tasks.wait_for_request(self.clients[0], request_id[0], 50)
```
mrizwan93: Replace it with `self.replicas[0]`

ipatests/test_integration/test_cert.py
```diff
@@ -46,3 +47,50 @@ def test_cacert_file_appear_with_option_F(self):
...
+        assert status == "SUBMITTING"
+
+        cmd = ['grep', '-P', "'certmonger|dogtag'", '/var/log/messages*']
```
mrizwan93: I would rather do it like this:
```
file_contents = self.replicas[0].get_file_contents('var/log/messages')
assert "Invalid cookie: u''" not in file_contents
```

ipatests/test_integration/test_cert.py
ssidhaye: Just fixed it. Realized it after checking it in. Thanks

ipatests/test_integration/test_cert.py
ssidhaye: OK, will fix it.

ipatests/test_integration/test_cert.py
ssidhaye: Fixed it

ipatests/test_integration/test_cert.py
ssidhaye: Fixed it

ipatests/test_integration/test_cert.py
ssidhaye: Fixed it

ipatests/test_integration/test_cert.py
ssidhaye: Valid point. Fixed it

ipatests/test_integration/test_cert.py
ssidhaye: Fixed it

.freeipa-pr-ci.yaml
```diff
@@ -1 +1 @@
-ipatests/prci_definitions/gating.yaml
```
fcami: Please don't mix the temp commit and actual code changes in a single commit.

ipatests/test_integration/test_cert.py
```diff
@@ -46,3 +47,50 @@ def test_cacert_file_appear_with_option_F(self):
         assert status == "MONITORING"
 
         self.clients[0].run_command(['ls', '-l', '/etc/pki/tls/test.CA'])
+
+
+class TestInstallMasterReplica(IntegrationTest):
```
fcami: This class name is already taken and sounds like this is an installation test. Could you change it?

ipatests/test_integration/test_cert.py
fcami: see tasks.wait_for_request():
`    for _i in range(0, timeout, 5):
(...)
        if state in ('CA_REJECTED', 'CA_UNREACHABLE', 'CA_UNCONFIGURED',
                     'NEED_GUIDANCE', 'NEED_CA', 'MONITORING'):
            break
        time.sleep(5)
`
I don't think there is any chance of status being SUBMITTING there.

.freeipa-pr-ci.yaml
ssidhaye: I had amended the previous commit, but the last amend messed things up. Will keep them separate henceforth

ipatests/test_integration/test_cert.py
ssidhaye: Will fix it.

ipatests/test_integration/test_cert.py
ssidhaye: I guess I'll just need to assert it then

fcami: or use a similar algorithm to wait for it, yes.

rcritten:
The commit message needs some re-wording. The issue isn't with certmonger but with the IPA renewal scripts and the handling/requirement for a "cookie" to maintain state.

The IPA scripts incorrectly required a cookie on every request.

I explained the details in https://pagure.io/freeipa/c/b5b9efeb57c010443c33c6f14f831abdbd804e78

ssidhaye:
@rcritten I have modified the commit message.

ipatests/test_integration/test_cert.py
ssidhaye: For now I have changed to ```assert``` instead of ```wait_for_request```

ipatests/test_integration/test_cert.py
```diff
@@ -90,3 +92,48 @@ def test_ipa_getcert_san_aci(self):
...
+        But, if certmonger on this node is stopped/interrupted
+        for any reason while the resubmit request is in "CA_WORKING" state
+        it will not be able to track changes to this cert anymore.
+
+        Pagure Issue: https://pagure.io/freeipa/issue/8164
+        """
+        cmd = ['getcert', 'resubmit', '-f',
+               paths.RA_AGENT_PEM]
+        result = self.replicas[0].run_command(cmd)
+        time.sleep(15)
```
fcami: I suppose this is to wait for the request to be in CA_WORKING. Rather than sleeping for 15s, it might be more reliable to poll until a timeout is reached.

rcritten:
Ok, this is going to seem a bit hacky but it will definitively tell you when and whether certmonger has resubmitted a certificate.
There is a "submitted" value in a certmonger tracking request of the form 20200319143229. It is not visible using any of the certmonger tools.
This could be obtained over dbus but I think it would be even uglier than what I'm about to propose... Use whichever unix-isms you prefer (awk, sed, cut, etc).

1. Get the request nickname: getcert list -f paths.RA_AGENT_PEM
the value is Request ID 'XYZ':

2. Using that nickname find the certmonger requests file:

id=grep XYZ /var/lib/certmonger/requests/* | cut -d: -f1

3. Find submitted in that file:

get submitted= /var/lib/certmonger/requests/<id>

So get submitted prior to doing anything. It should be 1970... since it was never submitted by certmonger but was added via start-tracking.

getcert resubmit -f ....

submitted now should be current time.

systemctl restart certmonger

Keep reading for submitted until it is > than the previous to know for sure that the request was resubmitted

Check the status of the cert and look for ca-error. Should be CA_WORKING and no error.

I'd write a helper within this test file to get a certmonger id from a FS. Then another helper to get the value of submitted.

ssidhaye:
@rcritten thanks! I'll make changes.

ipatests/test_integration/test_cert.py
```diff
@@ -7,15 +7,54 @@
...
+
+    :return submitted timestamp value
+    """
+    result = host.run_command(
+        'grep id=\'%s\' /var/lib/certmonger/requests/*' % requestid
+    )
+    print(result.stdout_text)
+    print(result.stderr_text)
+    assert result.stdout_text is not None
+    filename = result.stdout_text.split(':')[0]
```
amore17: This can be removed as grep with -l option will give filename only
#grep -l id=20200324074843 /var/lib/certmonger/requests/*
/var/lib/certmonger/requests/20200324074844

ipatests/test_integration/test_cert.py
```diff
@@ -7,15 +7,54 @@
...
+    """Get certmonger submitted value from
+    /var/lib/certmonger/requests/<timestamp>
+
+    :return submitted timestamp value
+    """
+    result = host.run_command(
+        'grep id=\'%s\' /var/lib/certmonger/requests/*' % requestid
+    )
+    print(result.stdout_text)
+    print(result.stderr_text)
```
amore17: Remove these print statements.
stdout_text and stderr_text captured with run_command() also.

ipatests/test_integration/test_cert.py
```diff
@@ -7,15 +7,54 @@
...
+
+    :
+    :return request ID string
+    """
+    request_id = re.findall(r'\d+', input_str)
+    print(request_id)
+    return request_id[1]
+
+
+def get_certmonger_request_value(host, requestid, directive):
```
amore17: Instead of directive you can use state.

ipatests/test_integration/test_cert.py
```diff
@@ -7,15 +7,54 @@
...
+    """Get certmonger submitted value from
+    /var/lib/certmonger/requests/<timestamp>
+
+    :return submitted timestamp value
+    """
+    result = host.run_command(
+        'grep id=\'%s\' /var/lib/certmonger/requests/*' % requestid
```
amore17: please use .format() :
host.run_command(['grep', 'id={}'.format(requestid), '/var/lib/certmonger/requests/*'])

ipatests/test_integration/test_cert.py
ssidhaye: fixed it

ipatests/test_integration/test_cert.py
ssidhaye: fixed it

ipatests/test_integration/test_cert.py
ssidhaye: @rcritten 's patch addresses this

ipatests/test_integration/test_cert.py
ssidhaye: @rcritten 's patch addresses this

ipatests/test_integration/test_cert.py
```diff
@@ -7,15 +7,53 @@
...
+
+def get_certmonger_request_value(host, requestid, state):
+    """Get certmonger submitted value from
+    /var/lib/certmonger/requests/<timestamp>
+
+    :return submitted timestamp value
+    """
+    result = host.run_command(
+        ['grep', '-rl', 'id={0}'.format(requestid),
+         '/var/lib/certmonger/requests/']
```
amore17: this can be replaced with paths.CERTMONGER_REQUESTS_DIR

ipatests/test_integration/test_cert.py
```diff
@@ -90,3 +128,94 @@ def test_ipa_getcert_san_aci(self):
...
+            pytest.fail("Request was not resubmitted")
+
+        assert resubmit2_val > resubmit_val
+
+        ca_error = get_certmonger_request_value(self.replicas[0],
+                                                request_id, "ca_error")
+        state = get_certmonger_request_value(self.replicas[0],
+                                             request_id, "state")
+
+        assert ca_error is None
```
amore17: These checks can be replaced with: 
add at the start of test >>   
since = time.strftime('%H:%M:%S')
and add at the end of test >>  
check_log = host.run_command(['journalctl', '-u', 'certmonger', '--since={}'.format(since)])
                                                 assert error not in (check_log.stdout_text, check_log.stderr_text)

rcritten: Why consult the journal when you can check the tracking request itself?

amore17: Yes Its removed.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -2121,7 +2121,7 @@ def wait_for_request(host, request_id, timeout=120):
         )
 
         state = result.stdout_text.strip()
-        print("certmonger request is in state %r", state)
+        print("certmonger request is in state %r" % state)
```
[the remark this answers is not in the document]
fcami: @ssidhaye I think logger.info() is really better in this case.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -2146,6 +2146,34 @@ def wait_for_request(host, request_id, timeout=120):
...
+       :param timeout: max time in seconds to wait for the status
+    """
+    for _i in range(0, timeout, 1):
+        result = host.run_command(
+            "getcert list -i %s | grep status: | awk '{ print $2 }'" %
+            request_id
+        )
+
+        state = result.stdout_text.strip()
+        print("certmonger request is in state %r" % state)
```
fcami: Same as above.

ipatests/pytest_ipa/integration/tasks.py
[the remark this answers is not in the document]
ssidhaye: fixed it

ipatests/pytest_ipa/integration/tasks.py
ssidhaye: fixed it

ipatests/test_integration/test_cert.py
```diff
@@ -7,15 +7,53 @@
...
 
+DEFAULT_RA_AGENT_SUBMITTED_VAL = '19700101000000'
+
+
+def get_certmonger_fs_id(input_str):
+    """Get certmonger FS ID
+    from the `getcert list -f /var/lib/ipa/ra-agent.pem` output
+    command output
+
+    :
```
mrizwan93: nitpick : can we remove this extra `:` ?

ipatests/test_integration/test_cert.py
```diff
@@ -7,15 +7,53 @@
...
+    :return request ID string
+    """
+    request_id = re.findall(r'\d+', input_str)
+    print(request_id)
```
mrizwan93: Do we need this print statement?

ssidhaye: removed it

ipatests/test_integration/test_cert.py
ssidhaye: removed it

rcritten:
Ack, please remove the temp commit.

rcritten:
master:

* 58ad7b74eb4136ff8cd10ad6caf463df7403f5b3 Test to check if Certmonger tracks certs in between reboots/interruptions and while in "CA_WORKING" state

PR #4383, opened 2020-03-18
ipatests: wait for SSSD to become online in backup/restore tests

flo-renaud:
The backup/restore tests are calling 'id admin' after restore
to make sure that the user name can be resolved after a restore.
The test should wait for SSSD backend to become online before
doing any check, otherwise there is a risk that the call to
'id admin' fails.

Fixes: https://pagure.io/freeipa/issue/8228

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -2130,3 +2130,27 @@ def wait_for_request(host, request_id, timeout=120):
         raise RuntimeError("request timed out")
 
     return state
+
+
+def wait_for_sssd_domain_status_online(host, timeout=120):
+    """Wait up to timeout (in seconds) for sssd domain status to become Online
+
+    The method is checking the Online Status of the domain as displayed by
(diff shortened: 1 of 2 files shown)


flo-renaud:
Adding ipa-4-7 and ipa-4-8 labels as the issue was discovered in relation to PR #4349 that was applied on the ipa-4-7 / ipa-4-8 branches.

wladich:
The PR looks good to me. 
I would only ask you is to make the name of the new function in tasks module more descriptive  and describe intended use in docstring: it should be clear that it is waiting for domain status and should be used specifically after starting 389-ds without restarting sssd. Now I had to read comments in pagure ticket to find this information.
Providing better name and documentation will help to avoid misusing this function for checking sssd status in other cases, for example after sssd restart.

abbra:

flo-renaud:
Thanks for the heads-up, temp commit removed.

abbra:
master:

* 3753862401b8500b7498d2723ccb727b5dd6263e ipatests: wait for SSSD to become online in backup/restore tests

abbra:

PR #4342, opened 2020-03-11
[Backport][ipa-4-6] Add secret to ajp connector

abbra:
This PR was opened manually because PR #4337 was pushed to master and backport to ipa-4-6 is required.

PR #4384, opened 2020-03-18
pki-proxy: Don't rely on running apache until it's configured

stanislavlevin:
This partially restores the pre- ec73de969f state of `http_proxy`,
which fails to restart the apache service during master
installation. The failure happens because of apache is not
configured yet on 'pki-tomcatd' installation phase. The mentioned
code and proposed one relies on the installer which bootstraps the
master.

Fixes: https://pagure.io/freeipa/issue/8233

### DIFF
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -363,12 +363,6 @@ class DogtagInstance(service.Service):
         with open(paths.HTTPD_IPA_PKI_PROXY_CONF, "w") as fd:
             fd.write(template)
             os.fchmod(fd.fileno(), 0o640)
-        # Restart httpd
-        http_service = services.knownservices.httpd
-        logger.debug("Restarting %s to apply AJP changes",
-                     http_service.service_name)
-        http_service.restart()
-        logger.debug("%s successfully restarted", http_service.service_name)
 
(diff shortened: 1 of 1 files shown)


abbra:
Originally I thought to reuse `http_proxy()` in both (installer and upgrade code) places and I made upgrade plugin that was separate from the current code. Unfortunately, after implementing it I realized the separate plugin is a no go as it was called within the sequence when 389-ds server is configured to not respond over 636 which made any PKI changes/restart not possible.
So that restart part is effectively a leftover of the original upgrade design which is not needed anymore.

abbra:
LGTM

stanislavlevin:
With this change and https://svn.apache.org/viewvc?view=revision&revision=1874456 (httpd) installation passed on ALTLinux.

abbra:
master:

* 14c9cf99889fec73c46b33f94b0dc483c4542044 pki-proxy: Don't rely on running apache until it's configured

stanislavlevin:
@abbra, the backport of this PR for `ipa-4-6` _is not required_ because of the manual backport [0]  doesn't contain the changes to be reverted.

[0] https://github.com/freeipa/freeipa/pull/4342/commits/13bbb922ce0c1111da6b5689c8b7c9b4264ddb92

PR #4374, opened 2020-03-16
Prevent adding IPA objects as external members of external groups

abbra:
The purpose of external groups in FreeIPA is to be able to reference
objects only existing in trusted domains. These members get resolved
through SSSD interfaces but there is nothing that prevents SSSD from
resolving any IPA user or group if they have security identifiers
associated.

Enforce a check that a SID returned by SSSD does not belong to IPA
domain and raise a validation error if this is the case. This would
prevent adding IPA users or groups as external members of an external
group.

Fixes: ...
Signed-off-by: Alexander Bokovoy <abokovoy@redhat.com>

PR #4395, opened 2020-03-19
[Backport][ipa-4-6] Prevent adding IPA objects as external members of external groups

abbra:
The purpose of external groups in FreeIPA is to be able to reference
objects only existing in trusted domains. These members get resolved
through SSSD interfaces but there is nothing that prevents SSSD from
resolving any IPA user or group if they have security identifiers
associated.

Enforce a check that a SID returned by SSSD does not belong to IPA
domain and raise a validation error if this is the case. This would
prevent adding IPA users or groups as external members of an external
group.

RN: Command 'ipa group-add-member' allowed to specify any user or group
RN: for '--external' option. A stricter check is added to verify that
RN: a group or user to be added as an external member does not come
RN: from IPA domain.

Fixes: https://pagure.io/freeipa/issue/8236
Reviewed-By: Florence Blanc-Renaud <flo@redhat.com>

### DIFF
--- a/ipaserver/dcerpc.py
+++ b/ipaserver/dcerpc.py
@@ -418,7 +418,12 @@ class DomainValidator(object):
         if object_name in result and \
            (pysss_nss_idmap.SID_KEY in result[object_name]):
             object_sid = result[object_name][pysss_nss_idmap.SID_KEY]
-            return object_sid
+            if self.is_trusted_sid_valid(object_sid):
+                return object_sid
+            else:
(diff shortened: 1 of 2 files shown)


abbra:
This PR was opened automatically because PR #4374 was pushed to master and backport to ipa-4-6 is required.

abbra:
ipa-4-6:

* c14e385141ea05f2709364b6f0fca844578a7652 Prevent adding IPA objects as external members of external groups

PR #4398, opened 2020-03-20
Update SELinux policy

tiran:
## Move ``ipa_custodia`` policy from upstream into IPA's policy

The SELinux policy defines file contexts that are also used by clients, e.g. /var/log/ipa/. Make freeipa-selinux a dependency of freeipa-common.

## Make freeipa-selinux a dependency of freeipa-common 

ipa-custodia is an internal service for IPA. The upstream SELinux policy has a separate module for ipa_custodia. Fold the current policy from Fedora rawhide into ipa's SELinux policy.

Related: https://pagure.io/freeipa/issue/6891

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,7 @@ ACLOCAL_AMFLAGS = -I m4
 
 if ENABLE_SERVER
     IPASERVER_SUBDIRS = ipaserver
-    SERVER_SUBDIRS = daemons init install selinux
+    SERVER_SUBDIRS = daemons init install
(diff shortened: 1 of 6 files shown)


abbra:
LGTM

tiran:
master:

* a55a722237b2d32c55eddddfb9c753c06e1eb5ee Integrate ipa_custodia policy
* d23322434f71c02505b4b85e85e184d274eaeb2d Move freeipa-selinux dependency to freeipa-common

PR #4407, opened 2020-03-21
Fix /doc/workshop subtree merge

tiran:
Something went wrong with git subtree merge of the external
freeipa-workshop repository. A couple of files accidently ended up
in / instead of /doc/workshop/.


abbra:
LGTM. ipatool couldn't even retrieve the patches from the original pull request so I had to do manual merge and add few more changes to move files around. I don't think subtree information is recorded anywhere in the tree.

tiran:
Ah, I see. AFAIK it is necessary to ``git remote add`` the source repo first.

abbra:
The question is how we can do that.

abbra:
LGTM. I still need to understand how to handle import of subtrees.

tiran:
master:

* d34db063774952814e0cae7e4499c7462ef04687 Fix /doc/workshop subtree merge

PR #4417, opened 2020-03-23

rcritten:

cacert.p12.

Signed-off-by: Alexander Bokovoy abokovoy@redhat.com

Replaces https://github.com/freeipa/freeipa/pull/2106

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-pwd-extop/common.c
+++ b/daemons/ipa-slapi-plugins/ipa-pwd-extop/common.c
@@ -422,6 +422,7 @@ done:
 
 int ipapwd_entry_checks(Slapi_PBlock *pb, struct slapi_entry *e,
                         int *is_root, int *is_krb, int *is_smb, int *is_ipant,
+                        int *is_memberof,
                         char *attr, int acc)
 {
     Slapi_Value *sval;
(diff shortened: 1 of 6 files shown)


abbra:
LGTM. All tests run as expected.

tiran:
master:

* 38204856fd72e5191bae07a47907314aacf43d1a Fix indentation levels
* 8c191ddf6d75090c80f567dd665bdacc44ef8883 ipatests: allow changing sysaccount passwords as cn=Directory Manager
* a620ac0f6f2505b6e9242f21dc0314c45747336e ipatests: test sysaccount password change with a password policy applied
* 527f30be76f0da6d4453745ace25f64948c0504f ipa-pwd-extop: use SLAPI_BIND_TARGET_SDN
* d9c41df6fd39b8d6bc879a9321b3f76eb1847b06 ipa-pwd-extop: don't check password policy for non-Kerberos account set by DM or a passsync manager
* 132a0f8771edd3a741bc60e6bc4c9b56cb172e4a Don't save password history on non-Kerberos accounts
* ff6984e2eea0f54851db796c8b3ad29c54a4e325 Add ability to change a user password as the Directory Manager
* 89066892151e1da5c2557d8eb76c2b04dbd61979 Test that pwpolicy only applied on Kerberos entries

PR #4418, opened 2020-03-23
selinux: disable ipa_custodia when installing custom policy

vmojzis:
Since ipa_custodia got integrated into ipa policy package, the upstream policy
module needs to be disabled before ipa module installation (in order to be able
to make changes to the ipa_custodia policy definitions).
Upstream ipa module gets overridden automatically because of higher priority of
the custom module, but there is no mechanism to automatically disable
ipa_custodia.

Related: https://pagure.io/freeipa/issue/6891

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1037,11 +1037,13 @@ fi
 %selinux_relabel_pre -s %{selinuxtype}
 
 %post selinux
+semodule -d ipa_custodia &> /dev/null || true;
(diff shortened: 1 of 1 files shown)


abbra:
LGTM, I started PR CI tests.

abbra:
master:

* 3aad16a75ef06d20d111773268df30ee04274db7 selinux: disable ipa_custodia when installing custom policy

PR #4448, opened 2020-03-25
po: fix LINGUAS to use whitespace separation

tiran:
The po/LINGUAS file contains a list of all avilable translations.
According to the GNU gettext documentation it's is a whitespace
separated list. Our LINGUAS file used newline separated list with inline
comments. This conflicts with weblate automation.

Fixes: https://pagure.io/freeipa/issue/8159
See: https://www.gnu.org/software/gettext/manual/html_node/po_002fLINGUAS.html


abbra:
LGTM. I verified that after runing `autogen.sh` the generated makefiles are usable for `make strip-po`

tiran:
master:

* ac5cb4260320a4509c840f679633eb36a79f6f99 po: fix LINGUAS to use whitespace separation

PR #4454, opened 2020-03-25
Don't configure ntpd with -x

rcritten:
slew mode (-x) may break ntpd from starting if time slew is too
great between the system and hardware clock. Slew mode is an
unstable configuration choice and has many known drawbacks.

https://pagure.io/freeipa/issue/8242

This applies **only** to the ipa-4-6 branch.

### DIFF
--- a/ipaclient/install/ntpconf.py
+++ b/ipaclient/install/ntpconf.py
@@ -80,7 +80,7 @@ keys /etc/ntp/keys
 #controlkey 8
 """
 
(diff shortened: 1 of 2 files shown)


tiran:
ACK, the man page for ntpd also advises to be careful with ``-x``.

tiran:
ipa-4-6:

* 2c1495460fcb0d58d27579bfbd6aba63b91bf985 Don't configure ntpd with -x

PR #4455, opened 2020-03-25
Set lookup_family_order = ipv6_first on IPv6-only clients

rcritten:
Look at the IPs configured on the interface that the client connects
to IPA with. If this interface has only IPv6 addresses configured
the set the option lookup_family_order to ipv6_first. This will
prefer IPv6 DNS names and will allow an IPv6-only client to be able
to connect to a mixed IPv4/IPv6 IPA installation.

https://pagure.io/freeipa/issue/8243

**NOTE:** I'm being conservative here and not using ipv6_only. This is an attempt to ensure that things will continue to work if the underlying network changes int he future.

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -956,6 +956,24 @@ def configure_sssd_conf(
         nss_service.set_option('memcache_timeout', 600)
         sssdconfig.save_service(nss_service)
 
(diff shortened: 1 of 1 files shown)


rcritten:
Dropping this PR, it is better done in sssd.

PR #4459, opened 2020-03-26
Test for ipahealtcheck.ipa.idns check when integrated DNS is setup

menonsudhir:
This testcase ensures that the ipahealthcheck.ipa.dns check displays correct result
when ipa master/replica is setup with integrated dns.

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -21,7 +21,7 @@ HEALTHCHECK_SYSTEMD_FILE = (
 HEALTHCHECK_LOG_ROTATE_CONF = "/etc/logrotate.d/ipahealthcheck"
 HEALTHCHECK_LOG_DIR = "/var/log/ipa/healthcheck"
 HEALTHCHECK_OUTPUT_FILE = "/tmp/output.json"
-HEALTHCHECK_PKG = ['freeipa-healthcheck']
+HEALTHCHECK_PKG = ["*ipa-healthcheck"]
 TOMCAT_CFG = "/var/lib/pki/pki-tomcat/conf/ca/CS.cfg"
 
 
@@ -107,7 +107,6 @@ DEFAULT_PKI_KRA_CERTS = [
     "storageCert cert-pki-kra",
     "auditSigningCert cert-pki-kra",
 ]
-
 
 def run_healthcheck(host, source=None, check=None, output_type="json"):
     """
@@ -456,6 +455,55 @@ class TestIpaHealthCheck(IntegrationTest):
             ['logrotate', '--debug', HEALTHCHECK_LOG_ROTATE_CONF]
         )
         assert msg not in cmd.stdout_text
+
+    def test_ipa_dns_systemrecords_check(self):
+        """
+        This test ensures that the ipahealthcheck.ipa.idns check
+        displays the correct result when master and replica is setup
+        with integrated DNS.
+        """
+        SRV_RECORDS = [
+            "_ldap._tcp." + self.replicas[0].domain.name + ".:" +
+            self.replicas[0].hostname + ".",
+            "_ldap._tcp." + self.master.domain.name + ".:" +
+            self.master.hostname + ".",
+            "_kerberos._tcp." + self.replicas[0].domain.name + ".:" +
+            self.replicas[0].hostname + ".",
+            "_kerberos._tcp." + self.master.domain.name + ".:" +
+            self.master.hostname + ".",
+            "_kerberos._udp." + self.replicas[0].domain.name + ".:" +
+            self.replicas[0].hostname + ".",
+            "_kerberos._udp." + self.master.domain.name + ".:" +
+            self.master.hostname + ".",
+            "_kerberos-master._tcp." + self.replicas[0].domain.name +
+            ".:" + self.replicas[0].hostname + ".",
+            "_kerberos-master._tcp." + self.master.domain.name + ".:" +
+            self.master.hostname + ".",
+            "_kerberos-master._udp." + self.replicas[0].domain.name +
+            ".:" + self.replicas[0].hostname + ".",
+            "_kerberos-master._udp." + self.master.domain.name + ".:" +
+            self.master.hostname + ".",
+            "_kpasswd._tcp." + self.replicas[0].domain.name + ".:" +
+            self.replicas[0].hostname + ".",
+            "_kpasswd._tcp." + self.master.domain.name + ".:" +
+            self.master.hostname + ".",
+            "_kpasswd._udp." + self.replicas[0].domain.name + ".:" +
+            self.replicas[0].hostname + ".",
+            "_kpasswd._udp." + self.master.domain.name + ".:" +
+            self.master.hostname + ".",
+            "\"" + self.master.domain.realm.upper() + "\"",
+            self.master.ip,
+            self.replicas[0].ip
+        ]
+        returncode, data = run_healthcheck(
+            self.master,
+            "ipahealthcheck.ipa.idns",
+            "IPADNSSystemRecordsCheck"
+        )
+        assert returncode == 0
+        for check in data:
+            assert check["result"] == "SUCCESS"
+            assert check["kw"]["key"] in SRV_RECORDS
 
     def test_ipa_healthcheck_remove(self):
         """


ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -252,9 +250,7 @@ def test_source_ipahealthcheck_meta_services_check_sssd(self):
         """
         self.master.run_command(["systemctl", "stop", "sssd"])
         returncode, data = run_healthcheck(
-            self.master,
-            "ipahealthcheck.meta.services",
-            "sssd",
+            self.master, "ipahealthcheck.meta.services", "sssd",
         )
```
rcritten: Is there a reason this unrelated reformatting is included?

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -453,10 +430,51 @@ def test_ipa_healthcheck_log_rotate_file_exist_issue35(self):
...
+            self.replicas[0].ip,
+        ]
+        returncode, data = run_healthcheck(
+            self.master, "ipahealthcheck.ipa.idns",
+        )
+        assert returncode == 0
+        for check in data:
+            assert check["result"] == "SUCCESS"
+            assert check["kw"]["key"] in SRV_MASTER_RECORDS
+            assert check["kw"]["key"] in SRV_REPLICA_RECORDS
```
rcritten: It may work if you change to an or instead. Any given key will only be for one of the records.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: @rcritten actually this was done by the tool i had used to format, will revert the changes.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: @rcritten have made the change

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -453,10 +432,63 @@ def test_ipa_healthcheck_log_rotate_file_exist_issue35(self):
...
+            self.replicas[0].hostname + ".",
+            "_kerberos-replica._tcp." + self.replicas[0].domain.name +
+            ".:" + self.replicas[0].hostname + ".",
+            "_kerberos-replica._udp." + self.replicas[0].domain.name +
+            ".:" + self.replicas[0].hostname + ".",
+            "_kpasswd._tcp." + self.replicas[0].domain.name + ".:" +
+            self.replicas[0].hostname + ".",
+            "_kpasswd._udp." + self.replicas[0].domain.name + ".:" +
+            self.replicas[0].hostname + ".",
+            '"' + self.replicas[0].domain.name.upper() + '"',
```
rcritten: The full key is \"EXAMPLE.TEST\" so you need those escape strings in there.

Also please use domain.realm.upper instead of domain.name.upper. They just happen to match for your case.

rcritten:
This will fix it. It passes on my local systems:

```
diff --git a/ipatests/test_integration/test_ipahealthcheck.py b/ipatests/test
_integration/test_ipahealthcheck.py
index 04f507c8b..60664ae10 100644
@@ -472,7 +472,8 @@ class TestIpaHealthCheck(IntegrationTest):
             "_kpasswd._udp." + self.master.domain.name + ".:" +
             self.master.hostname + ".",
             "\"" + self.master.domain.realm.upper() + "\"",
-            "\"" + self.master.ip + "\""
+            self.master.ip,
+            self.replicas[0].ip
         ]
         returncode, data = run_healthcheck(
             self.master, "ipahealthcheck.ipa.idns",

```

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -21,7 +21,7 @@
 HEALTHCHECK_LOG_ROTATE_CONF = "/etc/logrotate.d/ipahealthcheck"
 HEALTHCHECK_LOG_DIR = "/var/log/ipa/healthcheck"
 HEALTHCHECK_OUTPUT_FILE = "/tmp/output.json"
-HEALTHCHECK_PKG = ['freeipa-healthcheck']
+HEALTHCHECK_PKG = ["freeipa-healthcheck"]
```
mrizwan93: Can you update it  to  `HEALTHCHECK_PKG = ["*ipa-healthcheck"]` ?

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -333,17 +323,14 @@ def test_source_ipahealthcheck_ipa_host_check_ipahostkeytab(self):
...
         )
-        dirsrv_ipactl_status = 'Directory Service: STOPPED'
+        dirsrv_ipactl_status = "Directory Service: STOPPED"
         api.env.realm = self.master.domain.name
         serverid = (realm_to_serverid(api.env.realm)).upper()
         dirsrv_service = "dirsrv@%s.service" % serverid
         self.master.run_command(["systemctl", "stop", dirsrv_service])
-        result = self.master.run_command(
-            ["ipactl", "status"])
+        result = self.master.run_command(["ipactl", "status"])
```
mrizwan93: Are we using `result` anywhere else?

menonsudhir: @mrizwan93 removed the result since we are not using it.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: @mrizwan93 have made this change as well to ensure it works for other distros.

rcritten:
The patch looks ok to me but then unrelated changes need to be backed out.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -423,13 +426,11 @@ def test_ipa_healthcheck_dna_plugin_returns_warning_pagure_issue_60(self):
...
         tasks.user_add(
-            self.replicas[0], 'ipauser1', first='Test', last='User',
+            self.replicas[0], "ipauser1", first="Test", last="User",
         )
         # Now run the ipa-healthcheck command again
         returncode, data = run_healthcheck(
-            self.replicas[0],
-            "ipahealthcheck.ipa.dna",
-            "IPADNARangeCheck",
+            self.replicas[0], "ipahealthcheck.ipa.dna", "IPADNARangeCheck",
```
rcritten: fixing formatting not relevant here.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -423,13 +426,11 @@ def test_ipa_healthcheck_dna_plugin_returns_warning_pagure_issue_60(self):
         # DNA configuration.
         tasks.kinit_admin(self.replicas[0])
         tasks.user_add(
-            self.replicas[0], 'ipauser1', first='Test', last='User',
+            self.replicas[0], "ipauser1", first="Test", last="User",
```
rcritten: Changing quotes not needed.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -285,17 +284,21 @@ def test_source_ipahealthcheck_dogtag_ca_dogtagcertsconfigcheck(self):
...
             "ipahealthcheck.dogtag.ca",
             "DogtagCertsConfigCheck",
         )
         assert returncode == 1
         assert data[0]["result"] == "CRITICAL"
-        self.master.run_command(["mv", TOMCAT_CFG + ".old", TOMCAT_CFG])
+        self.master.run_command(
+            ["mv", paths.CA_CS_CFG_PATH + ".old", paths.CA_CS_CFG_PATH]
+        )
         self.master.run_command(["ipactl", "restart"])
```
rcritten: Use a constant is probably a good idea but should be done in a separate patch IMHO.

rcritten:
Can you add more details to the commit message? A BZ or Pagure link?

Additional text like:

Add test to compare the expected service records vs what is returned by healthcheck.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: @rcritten ok will remove this and add in a seperate patch.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: @rcritten the changes have been reverted.

menonsudhir:
> Can you add more details to the commit message? A BZ or Pagure link?
> 
> Additional text like:
> 
> Add test to compare the expected service records vs what is returned by healthcheck.

@rcritten have added the description and bz link.

rcritten:
Perfect, please remove the temp commit.

rcritten:
master:

* fd9f1b3d5bf35f8286505d1ed0971a42349a476b Test for ipahealthcheck.ipa.idns check when integrated DNS is setup

PR #4461, opened 2020-03-26
ipatests: Test if slew mode is not set while configuring ntpd

mrizwan93:
This is to ensure that slew mode (-x) is not configured while
installing ipa-server.

related: https://pagure.io/freeipa/issue/8242

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -62,6 +62,11 @@ def server_install_setup(func):
     return wrapped
 
 
+def check_slew_mode(host):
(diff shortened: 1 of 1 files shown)


rcritten:
The client also configured ntpd with -x. Can you do the same check on clients?

rcritten:
Nice catch on promoting a client to a master.

rcritten:
Please remove the temp commit and I'll get it pushed.

rcritten:
ipa-4-6:

* 81b859795c72f6c96b27137cc24d6df327ca8471 ipatests: Test if slew mode is not set while configuring ntpd

PR #4484, opened 2020-04-02
Azure: Allow distros to install Python they want

stanislavlevin:
The platforms may have different Pythons. But due to \[0\] the Python installed via the `UsePythonVersion@0` task should be compatible with the container's `libpythonxx.so`.

`AZURE_PYTHON_VERSION` platform variable is introduced to cover this. So, if your distro has Python3.8, set the mentioned variable to '3.8', later, this version will be installed by the `UsePythonVersion@0` Azure task for `WebUI_Unit_Tests` and `Tox`
jobs.

To allow tox to run any Python3 environment the 'py3' one is used. 'py3' is the well-known Tox's environment, which utilizes 'python3' executable.

\[0\]: https://github.com/microsoft/azure-pipelines-tasks/issues/11070

Fixes: https://pagure.io/freeipa/issue/8254

### DIFF
--- a/ipatests/azure/azure-pipelines.yml
+++ b/ipatests/azure/azure-pipelines.yml
@@ -93,7 +93,7 @@ jobs:
     - template: templates/${{ variables.PREPARE_BUILD_TEMPLATE }}
     - task: UsePythonVersion@0
       inputs:
-        versionSpec: 3.7
+        versionSpec: ${{ variables.AZURE_PYTHON_VERSION }}
         architecture: x64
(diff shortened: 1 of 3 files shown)


stanislavlevin:
```
tasks.common.PopenException: Process "vagrant provision" exited with error code 1
```

stanislavlevin:
```tasks.common.PopenException: Process "vagrant provision" exited with error code 1```

abbra:
LGTM

rcritten:
master:

* aa5a3336a80b381a4a49ad61accdde2f984203cb Azure: Allow distros to install Python they want

PR #4502, opened 2020-04-07
[WIP] Upgrade rfc2307 schema

abbra:
Update nisDomain and nisDomainObject to follow actual RFC2307bis schema
    
When RFC2307bis-based schema was added to FreeIPA and Fedora Directory in 2008, wrong OIDs were used for nisDomain attribute and nisDomainObject compared to the actual RFC2307bis schema. FreeIPA installed own schema version by default as rfc2307bis.ldif first and then as 15rfc2307bis.ldif. Fedora Directory (later 389-ds) kept it as 60nis.ldif in optional content and never installed into a working instance.
    
Recently, 389-ds decided to unify various RFC 2307-related schema files and install them by default. As result, FreeIPA-provided nisDomain attribute and nisDomainObject objectclass started to conflict with the ones installed by default by 389-ds.
    
All other attributes from 15rfc2307bis.ldif are not in use by FreeIPA. Since they are provided in 389-ds starting with 1.4.3.5 anyway, it makes no sense to keep them shipped.
    
This pull request updates 15rfc2307bis.ldif to only contain two entries which FreeIPA depends on.
    
 - attribute 'nisDomain'
 - objectclass 'nisDomainObject'
    
The definition of 'nisDomain' attribute and 'nisDomainObject' objectclass were updated to be compatible with 389-ds 1.4.3.5+.

Fixes: http://pagure.io/freeipa/issue/8258

This issue needs resolution from 389-ds side as well, we aren't agreed fully yet how to solve multiple-master replication problem. For details please see https://pagure.io/389-ds-base/issue/50933, https://pagure.io/389-ds-base/pull-request/51009, https://pagure.io/389-ds-base/pull-request/50934#comment-115218, and two Fedora 32 bugs: https://bugzilla.redhat.com/show_bug.cgi?id=1821548 and https://bugzilla.redhat.com/show_bug.cgi?id=1820176


abbra:
This PR is a work in progress. I'm running it against Rawhide because we unpushed 389-ds 1.4.3.5 from Fedora 32 to prevent the damage. However, testing on Rawhide currently needs modifications to how we run some services (chronyd and nis-domainname), thus there are patches that handle this part as well.

nis-domainname.service is broken for Debian as well, according to @tjaalton, so I guess this will be a general fix. We can simply remove the test but I'd prefer to keep it because the change for nis-domainname is needed to even being able to install FreeIPA server on some platforms (it doesn't work without this fix in rootless podman containers, for example).

install/share/15rfc2307bis.ldif
```diff
@@ -3,14 +3,5 @@
...
-attributeTypes: ( 1.3.6.1.4.1.42.2.27.1.1.13 NAME 'nisNetIdGroup' DESC 'nisNetIdGroup' EQUALITY caseExactIA5Match SYNTAX 1.3.6.1.4.1.1466.115.121.1.26 X-ORIGIN 'RFC2307bis' )
-attributeTypes: ( 1.3.6.1.4.1.42.2.27.1.1.14 NAME 'nisNetIdHost' DESC 'nisNetIdHost' EQUALITY caseExactIA5Match SYNTAX 1.3.6.1.4.1.1466.115.121.1.26 X-ORIGIN 'RFC2307bis' )
-objectClasses: ( 1.3.6.1.1.1.2.14 NAME 'nisKeyObject' DESC 'nisKeyObject' SUP top MUST ( cn $ nisPublickey $ nisSecretkey ) MAY ( uidNumber $ description ) )
-objectClasses: ( 1.3.1.6.1.1.1.2.15 NAME 'nisDomainObject' DESC 'nisDomainObject' SUP top AUXILIARY MUST nisDomain )
-objectClasses: ( 2.16.840.1.113730.3.2.4 NAME 'mailGroup' DESC 'mailGroup' SUP top MUST mail MAY ( cn $ mgrpRFC822MailMember ) )
-objectClasses: ( 1.3.6.1.4.1.42.2.27.1.2.6 NAME 'nisNetId' DESC 'nisNetId' SUP top MUST cn MAY ( nisNetIdUser $ nisNetIdGroup $ nisNetIdHost ) )
+attributeTypes: ( 1.3.6.1.1.1.1.30 NAME 'nisDomain' DESC 'NIS domain' EQUALITY caseIgnoreIA5Match SUBSTR caseIgnoreIA5SubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.26 )
+objectClasses: ( 1.3.6.1.1.1.2.15 NAME 'nisDomainObject' DESC 'nisDomainObject' SUP top AUXILIARY MUST nisDomain )
```
rcritten, line 7: Why retain these at all since they are provided by 389-ds? We may be able to do away with this entire file. The only remaining objectclass would be nisNetId which is unused.

ipatests/test_integration/test_sudo.py
```diff
@@ -153,8 +153,15 @@ def test_advise_script_enable_sudo_admins(self):
...
-        assert self.client.domain.name in result.stdout_text
+        """
+            Test that the client has been set up with proper NIS domain name
+            Also support containers where NIS domain name cannot be used
+            in runc environments
+        """
+        result = self.client.run_command('nisdomainname', raiseonerr=False)
+        assert (self.client.domain.name in result.stdout_text or
+                (("Local domain name not set" in result.stdout_text) and
+                 result.returncode == 1))
```
rcritten, line 164: I'd prefer if we detect that this is a container and only allow "Local domain not set" in that case otherwise this could regress and we wouldn't know.

install/share/15rfc2307bis.ldif
abbra: They are not provided by 389-ds in versions before 1.4.3.5. Additionally, we need to update existing installations which have `15rfc2307bis.ldif` installed. Since we only depend on these two entries in FreeIPA, we can remove the rest and it will work for both old and new 389-ds versions. In new versions we'll have the same definitions provided by the 389-ds itself as part of its core schema, in older versions FreeIPA would provide those. As long as those definitions are the same, 389-ds knows to ignore the second one.

ipatests/test_integration/test_sudo.py
abbra: Correct, I'll add another check. Or maybe move to skip the test in container.

rcritten:
This LGTM. Do you intend to backport this to ipa-4-8 as well?

abbra:
Not yet. This is not enough -- I have a larger fix that also accounts for removal of attributes/objectclasses from 99user.ldif. I need to collect all changes and update this PR.

abbra:
Azure pipeline fails because of new 389-ds-base 1.4.4.0 in Rawhide which fails to set itself up in Docker container due to a fix to https://pagure.io/389-ds-base/issue/49731 which enabled database default home directory to be on `/dev/shm/dirsrv` while `ns-slapd` couldn't write there.

abbra:
This PR is currently on hold. 389-ds team decided to revert RFC2307compat schema promotion due to issues with upgrade in multi-master replication.

PR #4508, opened 2020-04-08
Fix test ipa commands run as aduser assert

fcami:
test_trust.py::test_ipa_commands_run_as_aduser expects
'ipa: ERROR: Insufficient access: SASL(-14):'
'authorization failure: Invalid credentials'
but the correct string is:
'ipa: ERROR: Insufficient access:  Invalid credentials'
Change assert.
    
Fixes: https://pagure.io/freeipa/issue/8263


fcami:
Closing. See upstream ticket for reason.

PR #4525, opened 2020-04-14
CVE-2020-1722: prevent use of too long passwords

abbra:
NIST SP 800-63-3B sets a recommendation to have password length upper bound limited in A.2:

https://pages.nist.gov/800-63-3/sp800-63b.html#appA

	Users should be encouraged to make their passwords as lengthy as they
	want, within reason. Since the size of a hashed password is independent
	of its length, there is no reason not to permit the use of lengthy
	passwords (or pass phrases) if the user wishes. Extremely long passwords
	(perhaps megabytes in length) could conceivably require excessive
	processing time to hash, so it is reasonable to have some limit.

FreeIPA already applied 256 characters limit for non-random passwords
set through ipa-getkeytab tool. The limit was not, however, enforced in
other places.

MIT Kerberos limits the length of the password to 1024 characters in its
tools. However, these tools (kpasswd and 'cpw' command of kadmin) do not
differentiate between a password larger than 1024 and a password of 1024
characters. As a result, longer passwords are silently cut off.

To prevent silent cut off for user passwords, use limit of 1000
characters.

Thus, this patch enforces common limit of 1000 characters everywhere:
 - LDAP-based password changes
   - LDAP password change control
   - LDAP ADD and MOD operations on clear-text userPassword
   - Keytab setting with ipa-getkeytab
 - Kerberos password setting and changing

Reviewed-by: Simo Sorce <ssorce@redhat.com>

### DIFF
--- a/client/ipa-getkeytab.c
+++ b/client/ipa-getkeytab.c
@@ -562,6 +562,11 @@ done:
  * set match=true to enforce that the two entered passwords match.
  *
  * To prompt for an existing password provide prompt1 and set match=false.
+ *
+ * Implementation details:
+ * krb5_prompter_posix() does not differentiate between too long entry or
+ * an entry exactly the size of a buffer. Thus, allocate a bigger buffer
+ * and do the check for a too long password afterwards.
  */
 static char *ask_password(krb5_context krbctx, char *prompt1, char *prompt2,
                           bool match)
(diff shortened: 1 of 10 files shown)


abbra:
ACK was added because this fix was pre-approved during several offline review sessions.

kaleemsiddiqu:
@abbra 
Should we add tests for mod*(user-mod, host-mod) operations as well with long password?

abbra:
Forgot to update the test after the error message was changed to include 'clear-text password is too long'.

abbra:
> Should we add tests for mod*(user-mod, host-mod) operations as well with long password?
no, we don't need them because they are doing LDAP MODIFY operations and those tested already. The test suite covers all actual ways of modifying passwords from the LDAP server and KDC server perspectives.

abbra:
master:

* dbf5df4a66b68f62a9e063c43a30b46e539c603b CVE-2020-1722: prevent use of too long passwords

PR #4528, opened 2020-04-14
ipatests: Collect all logs on all Unix hosts

stanislavlevin:
Each integration test entity sets up its own list of logfiles. This is made by calling the callback of host's 'collect_log', which knows nothing about the context of execution: whether it's the test class scope or the test method one. Of course, in this case one-time collection of test method log is not supported because the logs tracker collects only test class logs.
In the meantime, almost all the entities (except 'client') collect identical logs. Besides, due to the IPA roles transformation an each IPA host can become master, replica or client, all of these, in turn, can have subroles. So, the most common case is the collection of all the possible logs from all the IPA (Unix) hosts. However, the customization of a logfiles collection is possible.
The collection is performed with the help of 'integration_logs' fixture. For example, to add a log file to a list of logs on a test completion one should add the dependency on this fixture and call its 'collect_method_log' method.
 ```
class TestFoo(IntegrationTest):
    def test_foo(self):
        pass

    def test_bar(self, integration_logs):
        integration_logs.collect_method_log(self.master, '/logfile')
```
Collected logs:
1) 'test_foo' - default logs
2) 'test_bar' - default logs + /logfile
3) 'TestFoo' - default logs

Also, this updated the default list of log files (added missing ones).

Fixes: https://pagure.io/freeipa/issue/8265

### DIFF
--- a/ipaplatform/base/paths.py
+++ b/ipaplatform/base/paths.py
@@ -407,6 +407,7 @@ class BasePathNamespace:
     DIRSRV_LOCK_DIR = "/run/lock/dirsrv"
     ALL_SLAPD_INSTANCE_SOCKETS = "/run/slapd-*.socket"
     VAR_LOG_DIRSRV_INSTANCE_TEMPLATE = "/var/log/dirsrv/slapd-%s"
+    VAR_LOG_DIRSRV = "/var/log/dirsrv/"
     SLAPD_INSTANCE_ACCESS_LOG_TEMPLATE = "/var/log/dirsrv/slapd-%s/access"
     SLAPD_INSTANCE_ERROR_LOG_TEMPLATE = "/var/log/dirsrv/slapd-%s/errors"
     SLAPD_INSTANCE_AUDIT_LOG_TEMPLATE = "/var/log/dirsrv/slapd-%s/audit"
--- a/ipatests/azure/Dockerfiles/docker-compose.yml
+++ b/ipatests/azure/Dockerfiles/docker-compose.yml
@@ -7,7 +7,7 @@ services:
     - ALL
     security_opt:
     - apparmor:unconfined
-    mem_limit: 1610612736
+    mem_limit: 1900m
     volumes:
     - /sys/fs/cgroup/systemd:/sys/fs/cgroup/systemd
     - ./ipa-test-config.yaml:/root/.ipa/ipa-test-config.yaml:ro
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by stanislavlevin** [comment]
@fcami, thank you for trying, but it was beaten by the original issue. I've added a commit to fix that.

**Comment by fcami** [comment]
The changes LGTM.

**Comment by abbra** [comment]
ACK

**Comment by rcritten** [comment]
Removing ack, this doesn't collect the dirsrv logs.

An example is:
fedora-latest/simple_replication -> test_integration-test_simple_replication.py-TestSimpleReplication -> master.ipa.test -> var -> log -> dirsrv -> empty

**Comment by stanislavlevin** [comment]
@rcritten, thank you for looking at this PR.

This is the expected behavior because `class_integration_logs` gathers IPA 'uninstall' logs (some logs are removed during that process).
But looking at it I realize that I don't collect 'install' logs.
So, it was fixed, thank you.

Simple diff of the fixture setup process:
``` diff
      SETUP    C class_integration_logs

      SETUP    C mh (fixtures used: class_integration_logs)
        SETUP    F integration_logs (fixtures used: class_integration_logs)
        test_integration/test_file.py::TestClass::test_m (fixtures used: class_integration_logs, integration_logs, mh, request)
-         TEARDOWN F integration_logs <= Collecting class logs
+         TEARDOWN F integration_logs <= Collecting method logs

      TEARDOWN C mh <= Collecting 'install' logs here
-       TEARDOWN C class_integration_logs
+       TEARDOWN C class_integration_logs <= Collecting 'uninstall' logs
```

Screenshot_20200424_142301

The logs collected on TestClass's install are in '{nodeid}-install'.
The logs collected on TestClass's uninstall are in '{nodeid}-uninstall'.

tiran:
master:

* 43ac2d9ab36518e510d621813ed73fb40e7cd2de ipatests: Cleanup 'collect_logs' decorator
* 5da309ee11f2aaba687d12a8c517f64070e22ae8 ipatests: Pretty print multihost config
* 63747bc0c041382536e2cad121bb591e42df3a2f ipatests: Collect all logs on all Unix hosts
* 87408ee7557cff04c1dd9c5de052e49f22c34f0f Azure: Increase memory limit

tiran:

stanislavlevin:
https://github.com/freeipa/freeipa/pull/4611

PR #4530, opened 2020-04-14
Include man pages in Sphinx documentation

tiran:


doc/conf.py
```diff
@@ -83,3 +83,80 @@
...
+     "Install new SSL server certificates", None, 1),
+    ("man/ipa-server-install.1", "ipa-server-install",
+     "Configure an IPA server", None, 1),
+    ("man/ipa-server-upgrade.1", "ipa-server-upgrade",
+     "upgrade IPA server", None, 1),
+    ("man/ipa-winsync-migrate.1", "ipa-winsync-migrate",
+     "Seamless migration of AD users created by winsync", None, 1),
+    ("man/ipa.1", "ipa", "IPA command-line interface", None, 1),
+    ("man/ipactl.8", "ipactl", "IPA Server Control Interface", None, 8),
+]
```
rcritten, line 162: Is there a way to auto-discover these? I fear it will get missed if/when a new man page is added (heck, we often forget the man page updates too).

tiran: There is no easy way to detect this. :( I could add a comment to each Makefile so it gets detected in code review.

Please note that this data structure is used for ``make man`` to turn the restructured text files back to groff man pages. In the long run I hope that we can generate man pages from RST instead of the other way around.

rcritten:
There seems to be a missing dependency on pandoc.

rcritten:
And after installing build fails with:
Unknown reader: man

tiran:
You should not need pandoc. It's only needed for initial import of man pages. I kept the ``gen_man.py`` script to keep it for further hacking.

rcritten:
Ah ok, I thought it would be used to re-build the man pages upon other changes, the same reason for the requestion about discovering the man pages to translate.

tiran:
> Ah ok, I thought it would be used to re-build the man pages upon other changes, the same reason for the requestion about discovering the man pages to translate.

I was able to get rid of all sphinx warnings and improve layout by modifying the man pages. This make re-building RST from man pages feasible.

rcritten:
This is looking a lot better. ack on the man page changes. What's the plan moving forward? Do you want to tidy this up further?

rcritten:
I used restview to view the new files in a browser and ran into this. I don't know if it is an issue with the changes or restview:

System Message: ERROR/3 (doc/man/ipa-rmkeytab.1.rst, line 43)

Unknown directive type "option".

.. option:: -p principal-name

   The non-realm part of the full principal name.

rcritten:
The ipa-ldap-updater page is a bit unreadable in readview too:

`* default: the starting value * add: add a value to an attribute * remove: remove a value from an attribute * only: set an attribute to this * onlyifexist: set an attribute to this only if the entry exists * deleteentry: remove the entry * replace: replace an existing value, format is old::new * addifnew: add a new attribute and value only if the attribute doesn't already exist. Only works with single-value attributes. * addifexist: add a new attribute and value only if the entry exists. This is used to update optional entries.`

tiran:
> The ipa-ldap-updater page is a bit unreadable in readview too:

The groff markup is incorrect here. Other places like https://manpages.debian.org/unstable/freeipa-server/ipa-ldap-updater.1.en.html have the same issue with the markup.

PR #4533, opened 2020-04-15
Use /run and /run/lock instead of /var

tiran:
Fixes: https://pagure.io/freeipa/issue/8272


stanislavlevin:
Hi, @tiran.
What is about:
```
$ git grep 'localstatedir.*/run'

daemons/dnssec/ipa-ods-exporter.socket.in:ListenStream=@localstatedir@/run/opendnssec/engine.sock
daemons/ipa-kdb/Makefile.am:    -DLDAPIDIR=\""$(localstatedir)/run"\"           \
daemons/ipa-sam/Makefile.am:    -DLDAPIDIR=\""$(localstatedir)/run"\"           \
server.m4:krb5rundir="${localstatedir}/run/krb5kdc"
```

tiran:
@stanislavlevin Good catch! ``runstatedir`` is not yet available in our autoconf versions. I have added it to ``configure.ac``

abbra:
LGTM. Thanks for fixing this old-standing issue. ;)

tiran:
master:

* bdf11371693936e4ff6cc4d0e245b19493df564c Use /run and /run/lock instead of /var

PR #4540, opened 2020-04-16
ipatests: Test to check password leak in apache error log

mrizwan93:
Host enrollment with OTP used to log the password in cleartext
to apache error log. This test ensures that the password should
not be a cleartext.

related: https://pagure.io/freeipa/issue/8017

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -523,6 +523,28 @@ class TestIPACommand(IntegrationTest):
                                     raiseonerr=False)
         assert result.returncode == 0
 
+    def test_cleartext_password_httpd_log(self):
+        """Test to check password leak in apache error log
+
+        Host enrollment with OTP used to log the password in cleartext
+        to apache error log. This test ensures that the password should
+        not be log in cleartext.
+
+        related: https://pagure.io/freeipa/issue/8017
+        """
+        hostname = 'test.{}'.format(self.master.domain.name)
+        passwd = 'Secret123'
+
+        self.master.run_command(['ipa', 'host-add', '--force',
+                                 hostname, '--password', passwd])
+
+        # remove added host i.e cleanup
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_commands.py
```diff
@@ -523,6 +523,28 @@ def test_huge_password(self):
...
+
+        self.master.run_command(['ipa', 'host-add', '--force',
+                                 hostname, '--password', passwd])
+
+        result = self.master.run_command(['grep', 'host_add',
+                                          paths.VAR_LOG_HTTPD_ERROR])
+        assert (hostname in result.stdout_text and
+                passwd not in result.stdout_text)
+
+        self.master.run_command(['ipa', 'host-del', hostname])
```
rcritten, line 546: Please move the host-del before the assert so that things will always be cleaned up, even in case the assert fails.

The grep is rather broad and may return multiple, unrelated entries, but in any case the password should never be logged so this should be fine.

mrizwan93: Changed the grep to grab the entries containing host only. This should minimize the result.

rcritten:
The build is failing, 
*** Error compiling '/builddir/build/BUILDROOT/freeipa-4.9.0.dev-0.fc31.x86_64/usr/lib/python3.7/site-packages/ipatests/test_integration/test_commands.py'...
  File "/usr/lib/python3.7/site-packages/ipatests/test_integration/test_commands.py", line 546
    assert passwd not in result.stdout_text)

rcritten:
Looks good, thanks. Please remove the temporary commit and I'll add the ack.

mrizwan93:
@rcritten Thanks, temp commit removed.

rcritten:
master:

* 2c54609d73d691874ff8d5512a2d1a1284f5e77e ipatests: Test to check password leak in apache error log

PR #4544, opened 2020-04-17
Test for removing a subgroup

ssidhaye:
Problem description:
Removing an IPA sub-group should NOT remove the members
from indirect parent that also belong to other subgroups

The test:
A user and three groups are created a,b,c
'c' should be a child of 'b' so that you have a->b->c

user is direct member of 'a' and as a result member of 'b'
and 'c'. Now when one adds a direct membership to 'b' nothing will
change.

If one removes the direct membership to 'b' again,
nothing should change as well

Pagure Link: https://pagure.io/SSSD/sssd/issue/3636

Signed-off-by: Sumedh Sidhaye ssidhaye@redhat.com

### DIFF
--- a/ipatests/test_integration/test_sssd.py
+++ b/ipatests/test_integration/test_sssd.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import
 import time
 from contextlib import contextmanager
 import re
+import os
 
 import pytest
 import subprocess
@@ -16,6 +17,7 @@ import textwrap
 
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
+from ipatests.pytest_ipa.integration.tasks import clear_sssd_cache
 from ipatests.util import xfail_context
 from ipaplatform.tasks import tasks as platform_tasks
 from ipaplatform.osinfo import osinfo
@@ -403,3 +405,178 @@ class TestSSSDWithAdTrust(IntegrationTest):
                 assert 'no such user' in res.stderr_text
         # verify the user can be retrieved after re-enabling trustdomain
         self.master.run_command(['id', user])
+
+
+class TestNestedMembers(IntegrationTest):
+    num_clients = 1
+    username = "testuser001"
+    userpasswd = 'Secret123'
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master)
+        tasks.install_client(cls.master, cls.clients[0])
+
+    @pytest.fixture
+    def nested_group_setup(self, tmpdir):
+        """Setup and Clean up groups and user created"""
+        master = self.master
+        client = self.clients[0]
+
+        # add a user and set password
+        tasks.create_active_user(master, self.username, self.userpasswd)
+        tasks.kinit_admin(master)
+
+        privkey, pubkey = tasks.generate_ssh_keypair()
+        with open(os.path.join(
+                tmpdir, 'ssh_priv_key'), 'w') as fp:
+            fp.write(privkey)
+
+        master.run_command([
+            'ipa', 'user-mod', self.username, '--ssh', "{}".format(pubkey)
+        ])
+
+        master.put_file_contents('/tmp/user_ssh_priv_key', privkey)
+        master.run_command(['chmod', '600', '/tmp/user_ssh_priv_key'])
+
+        # add group groupa
+        cmd_output = master.run_command(['ipa', 'group-add', 'groupa'])
+        assert 'Added group "groupa"' in cmd_output.stdout_text
+
+        # add group groupb
+        cmd_output = master.run_command(['ipa', 'group-add', 'groupb'])
+        assert 'Added group "groupb"' in cmd_output.stdout_text
+
+        # add group groupc
+        cmd_output = master.run_command(['ipa', 'group-add', 'groupc'])
+        assert 'Added group "groupc"' in cmd_output.stdout_text
+
+        client.put_file_contents('/tmp/user_ssh_priv_key',
+                                 privkey)
+        client.run_command(['chmod', '600', '/tmp/user_ssh_priv_key'])
+        yield
+        # test cleanup
+        for group in ['groupa', 'groupb', 'groupc']:
+            self.master.run_command(['ipa', 'group-del', group, '--continue'])
+        self.master.run_command(['ipa', 'user-del', self.username,
+                                 '--no-preserve', '--continue'])
+        tasks.kdestroy_all(self.master)
+        tasks.kdestroy_all(self.clients[0])
+
+    def test_nested_group_members(self, tmpdir, nested_group_setup):
+        """Nested group memberships should be honoured
+
+        "groupc" should be a child of "groupb"
+        so that parent child relationship is as follows:
+        "groupa"->"groupb"->"groupc"
+
+        testuser001 is direct member of "groupa" and as a result
+        member of "groupb" and "groupc"".
+        Now if one adds a direct membership to "groupb"
+        nothing will change.
+
+        Now if one removes the direct membership to "groupb"
+        nothing should change, the memberships should be honored
+        Linked Issue: https://pagure.io/SSSD/sssd/issue/3636
+        """
+        master = self.master
+        client = self.clients[0]
+
+        # add group members
+        cmd_output = master.run_command(['ipa', 'group-add-member',
+                                         'groupb', '--groups', 'groupa'])
+        assert 'Group name: groupb' in cmd_output.stdout_text
+        assert 'Member groups: groupa' in cmd_output.stdout_text
+        assert 'Number of members added 1' in cmd_output.stdout_text
+
+        cmd_output = master.run_command(['ipa', 'group-add-member',
+                                         'groupc', '--groups', 'groupb'])
+        assert 'Group name: groupc' in cmd_output.stdout_text
+        assert 'Member groups: groupb' in cmd_output.stdout_text
+        assert 'Indirect Member groups: groupa' in cmd_output.stdout_text
+
+        # add user to group 'groupa'
+        cmd_output = master.run_command(['ipa', 'group-add-member',
+                                         'groupa', '--users', self.username])
+        assert 'Group name: groupa' in cmd_output.stdout_text
+        assert_str = 'Member users: {}'.format(self.username)
+        assert assert_str in cmd_output.stdout_text
+        assert 'Member of groups: groupb' in cmd_output.stdout_text
+        assert 'Indirect Member of group: groupc' in cmd_output.stdout_text
+
+        # clear sssd_cache
+        clear_sssd_cache(master)
+
+        # user lookup
+        # at this point, testuser001 has the following group memberships
+        # Member of groups: groupa, ipausers
+        # Indirect Member of group: groupb, groupc
+        cmd_output = master.run_command(['ipa', 'user-show', self.username])
+        assert 'groupa' in cmd_output.stdout_text
+        assert 'ipausers' in cmd_output.stdout_text
+        assert 'groupb' in cmd_output.stdout_text
+        assert 'groupc' in cmd_output.stdout_text
+
+        clear_sssd_cache(client)
+
+        cmd = ['ssh', '-i', '/tmp/user_ssh_priv_key',
+               '-q', '{}@{}'.format(self.username, client.hostname),
+               'groups']
+        cmd_output = master.run_command(cmd)
+        assert self.username in cmd_output.stdout_text
+        assert 'groupa' in cmd_output.stdout_text
+        assert 'groupb' in cmd_output.stdout_text
+        assert 'groupc' in cmd_output.stdout_text
+
+        # add member
+        cmd_output = master.run_command(['ipa', 'group-add-member',
+                                         'groupb', '--users', self.username])
+        assert 'Group name: groupb' in cmd_output.stdout_text
+        assert_str = 'Member users: {}'.format(self.username)
+        assert assert_str in cmd_output.stdout_text
+        assert 'Member groups: groupa' in cmd_output.stdout_text
+        assert 'Member of groups: groupc' in cmd_output.stdout_text
+        assert 'Number of members added 1' in cmd_output.stdout_text
+
+        # now check ssh on the client
+        clear_sssd_cache(client)
+
+        # after adding testuser001 to b group
+        # testuser001 will have the following memberships
+        # Member of groups: groupa, ipausers, groupb
+        # Indirect Member of group: groupc
+        cmd = ['ssh', '-i', '/tmp/user_ssh_priv_key',
+               '-q', '{}@{}'.format(self.username, client.hostname),
+               'groups']
+        cmd_output = client.run_command(cmd)
+        assert self.username in cmd_output.stdout_text
+        assert 'groupa' in cmd_output.stdout_text
+        assert 'groupb' in cmd_output.stdout_text
+        assert 'groupc' in cmd_output.stdout_text
+
+        # now back to server to remove member
+        cmd_output = master.run_command(['ipa', 'group-remove-member',
+                                         'groupb', '--users', self.username])
+        assert_str = 'Indirect Member users: {}'.format(self.username)
+        assert 'Group name: groupb' in cmd_output.stdout_text
+        assert 'Member groups: groupa' in cmd_output.stdout_text
+        assert 'Member of groups: groupc' in cmd_output.stdout_text
+        assert assert_str in cmd_output.stdout_text
+        assert 'Number of members removed 1' in cmd_output.stdout_text
+
+        clear_sssd_cache(master)
+
+        # now check ssh on the client again
+        # after removing testuser001 from b group
+        # testuser001 will have the following memberships
+        # Member of groups: groupa, ipausers
+        # Indirect Member of group: groupb, groupc
+        clear_sssd_cache(client)
+        cmd = ['ssh', '-i', '/tmp/user_ssh_priv_key',
+               '-q', '{}@{}'.format(self.username, client.hostname),
+               'groups']
+        cmd_output = client.run_command(cmd)
+        assert self.username in cmd_output.stdout_text
+        assert 'groupa' in cmd_output.stdout_text
+        assert 'groupb' in cmd_output.stdout_text
+        assert 'groupc' in cmd_output.stdout_text


ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,145 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        cmd_output = master.run_command(cmd)
+        assert 'Added user "%s"' % username in cmd_output.stdout_text
+
+        cmd_output = master.run_command(['ipa', 'passwd', username],
+                                        stdin_text=input_passwd)
+        assert 'Changed password' in cmd_output.stdout_text
+
+        # add group a
+        cmd_output = master.run_command(['ipa', 'group-add', 'a'])
+        assert 'Added group "child1"' in cmd_output.stdout_text
```
mrizwan93, line 459: Adding group `a` would result in `Added group "child1"` ?

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,145 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+
+        # add group b
+        cmd_output = master.run_command(['ipa', 'group-add', 'b'])
+        assert 'Added group "child2"' in cmd_output.stdout_text
+
+        # add group parent
+        cmd_output = master.run_command(['ipa', 'group-add', 'c'])
+        assert 'Added group "parent"' in cmd_output.stdout_text
```
mrizwan93, line 467: Adding group `c` would result in `Added group "parent"` ?

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,145 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        assert 'Group name: a' in cmd_output.stdout_text
+        assert 'Member users: {}'.format(username) in cmd_output.stdout_text
+        assert 'Member of groups: b' in cmd_output.stdout_text
+        assert 'Indirect Member of group: c' in cmd_output.stdout_text
+
+        # clear sssd_cache
+        clear_sssd_cache(master)
+
+        # user lookup
+        cmd_output = master.run_command(['ipa', 'user-show', 'u1',
```
mrizwan93, line 494: Do we have added user `u1 ` yet?

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,145 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+                            'any_to_any'])
+        # add a user
+        username = "testuser" + str(random.randint(200000, 9999999))
+        userpasswd = 'Secret123'
+        cmd = ['ipa', 'user-add',
+               '--first', username,
+               '--last', username,
+               '--password', username]
+        input_passwd = '{upasswd}\n{upasswd}\n'.format(upasswd=userpasswd)
+        cmd_output = master.run_command(cmd)
```
mrizwan93, line 450: You can use `tasks.user_add()` here.

ipatests/test_integration/test_sssd.py
ssidhaye: This is a typo. Will fix it.

ipatests/test_integration/test_sssd.py
ssidhaye: Another typo. I had earlier used parent, child1 and child2 but later changed it to a, b, c.
Will fix this

ipatests/test_integration/test_sssd.py
ssidhaye: nother typo. I had earlier used parent, child1 and child2 but later changed it to a, b, c.
Will fix this

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,145 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        master.run_command(['ipa', 'hbacrule-add', '--usercat', 'all',
+                            '--hostcat', 'all', '--servicecat', 'all',
+                            'any_to_any'])
+        # add a user
+        username = "testuser" + str(random.randint(200000, 9999999))
+        userpasswd = 'Secret123'
+        cmd = ['ipa', 'user-add',
+               '--first', username,
+               '--last', username,
+               '--password', username]
```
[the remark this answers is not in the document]
ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: thanks for the suggestion. fixed it.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,141 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        master = self.master
+        client = self.clients[0]
+        tasks.kinit_admin(master)
+
+        # add hbacrule
+        master.run_command(['ipa', 'hbacrule-add', '--usercat', 'all',
+                            '--hostcat', 'all', '--servicecat', 'all',
+                            'any_to_any'])
+        # add a user
+        username = "testuser" + str(random.randint(200000, 9999999))
```
mrizwan93, line 443: As we are not using more user here, `testuser001` or something like that would be enough.
Make sure to remove the user/group at the end of test as a cleanup task.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,141 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        input_passwd = '{upasswd}\n{upasswd}\n'.format(upasswd=userpasswd)
+        tasks.user_add(master, username, password=userpasswd)
+
+        cmd_output = master.run_command(['ipa', 'passwd', username],
+                                        stdin_text=input_passwd)
+        assert 'Changed password' in cmd_output.stdout_text
+
+        # add group a
+        cmd_output = master.run_command(['ipa', 'group-add', 'a'])
+        assert 'Added group "a"' in cmd_output.stdout_text
```
mrizwan93, line 455: It would be nice to have group name like `parent` `child1` and `child2`.
OR
`grandparent` >  `parent` > `child`  (a > b >c).

It would make easy to understand who is who's subgroup.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,141 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        assert 'Group name: a' in cmd_output.stdout_text
+        assert 'Member users: {}'.format(username) in cmd_output.stdout_text
+        assert 'Member of groups: b' in cmd_output.stdout_text
+        assert 'Indirect Member of group: c' in cmd_output.stdout_text
+
+        # clear sssd_cache
+        clear_sssd_cache(master)
+
+        # user lookup
+        cmd = 'ipa user-show {} | grep group'.format(username)
```
mrizwan93, line 490: We don't really need `|` here. We are printing only 1 user details and that would be much lines. Please remove it.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,141 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        cmd_output = master.run_command(cmd)
+        assert_str = 'Member of groups: a, ipausers'
+        assert assert_str in cmd_output.stdout_text
+        assert_str = 'Indirect Member of group: b, c'
+        assert assert_str in cmd_output.stdout_text
+
+        clear_sssd_cache(master)
+
+        # ssh to client using the user created above from master
+        input_passwd = '{upasswd}\n{upasswd}\n{upasswd}\n{upasswd}\n'\
```
mrizwan93, line 500: We already changes password for the user. Do we need to specify password 4 times again?

ipatests/test_integration/test_sssd.py
ssidhaye: I only want to view and then assert on the  group info, rest of the info is not of use here.

ipatests/test_integration/test_sssd.py
ssidhaye: When a new user is created and you try to use that user for ssh,
it expires the password and prompts you for a new one

ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: Fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +404,138 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        # add a user
+        username = "testuser001"
+        userpasswd = 'Secret123'
+
+        input_passwd = '{upasswd}\n{upasswd}\n'.format(upasswd=userpasswd)
+        tasks.user_add(master, username, password=userpasswd)
+
+        cmd_output = master.run_command(['ipa', 'passwd', username],
+                                        stdin_text=input_passwd)
+        assert 'Changed password' in cmd_output.stdout_text
```
flo-renaud, line 450: You can use tasks.create_active_user to create the user + change the password.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +404,138 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        Linked Issue: https://pagure.io/SSSD/sssd/issue/3636
+        """
+        master = self.master
+        client = self.clients[0]
+        tasks.kinit_admin(master)
+
+        # add hbacrule
+        master.run_command(['ipa', 'hbacrule-add', '--usercat', 'all',
+                            '--hostcat', 'all', '--servicecat', 'all',
+                            'any_to_any'])
```
flo-renaud, line 440: By default IPA install has a HBAC rule allow_all with the same settings. Are you sure you need to create that one?

ssidhaye: removed hbacrule-add

ipatests/test_integration/test_sssd.py
ssidhaye: used tasks.create_active_user instead of ```ipa user-add``` + ```ipa passwd```

ipatests/test_integration/test_sssd.py
ssidhaye: changed the group names to parent, child1 and child2

ssidhaye:
> Hi @ssidhaye
> thanks for the PR. Please find inline comments. Also, I am not sure that clear_sssd_cache is needed on the master, but I guess it can't hurt...

@flo-renaud  This was added to eliminate entries picked up from cache, this was suggested by Sumit.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +404,135 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+
+        # now check ssh on the client
+        clear_sssd_cache(client)
+        cmd_output = client.run_command(['ssh', '-q',
+                                        '%s@%s' % (username, client.hostname),
+                                         'groups'], stdin_text=userpasswd)
+        assert '{}'.format(username) in cmd_output.stdout_text
+        assert 'a' in cmd_output.stdout_text
+        assert 'b' in cmd_output.stdout_text
+        assert 'c' in cmd_output.stdout_text
```
mrizwan93, line 515: Need to replace a, b and c with parent, child1 and child2

ssidhaye: replaced it

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +404,135 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        Now if one adds a direct membership to "child1"
+        nothing will change.
+
+        Now if one removes the direct membership to "child1"
+        nothing should change, the memberships should be honored
+        Linked Issue: https://pagure.io/SSSD/sssd/issue/3636
+        """
+        master = self.master
+        client = self.clients[0]
+        tasks.kinit_admin(master)
```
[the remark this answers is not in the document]
ssidhaye: removed it

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +407,173 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        # since ssh does not work well paramiko
+        tmpdir = tempfile.mkdtemp(prefix="tmp-")
+
+        # add a user and set password
+        tasks.create_active_user(master, username, userpasswd)
+        tasks.kinit_admin(master)
+
+        pub_keys = []
+
+        for i in range(40):
```
flo-renaud, line 452: I don't understand this loop. Why does the test add 40 keys?
Wouldn't a single key be enough for the test?

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +407,173 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        assert 'Member groups: parent' in cmd_output.stdout_text
+        assert 'Member of groups: child2' in cmd_output.stdout_text
+        assert 'Number of members added 1' in cmd_output.stdout_text
+
+        # now check ssh on the client
+        clear_sssd_cache(client)
+        try:
+            cmd = 'ssh -i {} -q {}@{} groups'.format(user_ssh_priv_key,
+                                                     username, client.hostname)
+            cmd_output = client.run_command(cmd)
```
flo-renaud, line 543: Also need to copy the user_ssh_priv_key file to the client.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +407,173 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+
+        clear_sssd_cache(master)
+
+        sshcon = paramiko.SSHClient()
+        sshcon.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+
+        try:
+            cmd = 'ssh -i {} -q {}@{} groups'.format(user_ssh_priv_key,
+                                                     username, client.hostname)
+            cmd_output = master.run_command(cmd)
```
flo-renaud, line 520: The command is run on the master but the file user_ssh_priv_key was generated on the controller and never copied to the master. This is why you get `Warning: Identity file /tmp/tmp-9ywpcf6o/ssh_priv_1 not accessible: No such file or directory.`

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +407,173 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+    def install(cls, mh):
+        tasks.install_master(cls.master)
+        tasks.install_client(cls.master, cls.clients[0])
+
+    def test_nested_group_members(self):
+        """Nested group memberships should be honoured
+
+        "child2" should be a child of "child1"
+        so that parent child relationship is as follows:
+        "parent"->"child1"->"child2"
```
frasertweedale, line 425: I find the naming confusing.  "parent" is a member (subset) of "child1", and "child1" is a member (subset) of "child2".
It's a nit-pick, but even just calling them 'A' 'B' and 'C' would help, because those names don't have contradictory meanings.

ipatests/test_integration/test_sssd.py
ssidhaye: removed the loop.

ipatests/test_integration/test_sssd.py
ssidhaye: added the copy to master and client part

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: that's primary reason why I had used a,b,c but there was another suggestion to use grandparent, parent and child from @mrizwan93 

> It would be nice to have group name like `parent` `child1` and `child2`.
> OR
> `grandparent` > `parent` > `child` (a > b >c).
> 
> It would make easy to understand who is who's subgroup.

but I guess that would be also confusing.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +407,188 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        cmd_output = master.run_command(['ipa', 'user-show', username])
+        assert 'parent' in cmd_output.stdout_text
+        assert 'ipausers' in cmd_output.stdout_text
+        assert 'child1' in cmd_output.stdout_text
+        assert 'child2' in cmd_output.stdout_text
+
+        clear_sssd_cache(master)
+
+        sshcon = paramiko.SSHClient()
+        sshcon.set_missing_host_key_policy(paramiko.AutoAddPolicy())
```
frasertweedale, line 521: Are we even using the paramiko library now?  I see no further reference to `sshcon`.

Is it the case the the ssh connections are being made from client to server, and not directly from this test code,
via `run_command` on line 553?  If so, we can remove the paramiko import and all references to paramiko.
If so we can

ssidhaye: Fixed it. Forgot to remove those two lines.

ipatests/test_integration/test_sssd.py
```diff
@@ -9,13 +9,17 @@
 import time
 from contextlib import contextmanager
 import re
+import os
+import paramiko
```
frasertweedale, line 13: So we shouldn't need this import any more either?

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +407,185 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        try:
+            cmd = 'ssh -i {} -q {}@{} groups'.format('/tmp/user_ssh_priv_key',
+                                                     username, client.hostname)
+            cmd_output = master.run_command(cmd)
+            assert '{}'.format(username) in cmd_output.stdout_text
+            assert 'parent' in cmd_output.stdout_text
+            assert 'child1' in cmd_output.stdout_text
+            assert 'child2' in cmd_output.stdout_text
+        except (paramiko.AuthenticationException,
+                paramiko.SSHException) as e:
```
frasertweedale, line 529: or this except block catching paramiko exceptions (we are not using paramiko)

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +407,185 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        try:
+            cmd = 'ssh -i {} -q {}@{} groups'.format('/tmp/user_ssh_priv_key',
+                                                     username, client.hostname)
+            cmd_output = client.run_command(cmd)
+            assert '{}'.format(username) in cmd_output.stdout_text
+            assert 'parent' in cmd_output.stdout_text
+            assert 'child1' in cmd_output.stdout_text
+            assert 'child2' in cmd_output.stdout_text
+        except (paramiko.AuthenticationException,
+                paramiko.SSHException) as e:
```
frasertweedale, line 556: also here

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +407,185 @@ def test_trustdomain_disable_disables_subdomain(self):
...
```
frasertweedale, line 582: and here.

ipatests/test_integration/test_sssd.py
ssidhaye: Removed the import

ipatests/test_integration/test_sssd.py
ssidhaye: Removed it

ipatests/test_integration/test_sssd.py
ssidhaye: Removed it

ipatests/test_integration/test_sssd.py
ssidhaye: Removed it

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +406,182 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        # now check ssh on the client
+        clear_sssd_cache(client)
+        client.put_file_contents('/tmp/user_ssh_priv_key',
+                                 ssh_key_file_contents)
+        client.run_command('chmod 600 /tmp/user_ssh_priv_key')
+
+        try:
+            cmd = 'ssh -i {} -q {}@{} groups'.format('/tmp/user_ssh_priv_key',
+                                                     username, client.hostname)
+            cmd_output = client.run_command(cmd)
```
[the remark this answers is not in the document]
ssidhaye: After the group-add command on the master, the about command checks it on the client. So the ssh command from and to client is on purpose, due to the nature of the test

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +406,182 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        clear_sssd_cache(master)
+
+        # user lookup
+        cmd_output = master.run_command(['ipa', 'user-show', username])
+        assert 'parent' in cmd_output.stdout_text
+        assert 'ipausers' in cmd_output.stdout_text
+        assert 'child1' in cmd_output.stdout_text
+        assert 'child2' in cmd_output.stdout_text
+
+        clear_sssd_cache(master)
```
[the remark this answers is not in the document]
ssidhaye: Fixed it

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +406,182 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        assert 'Member of groups: child2' in cmd_output.stdout_text
+        assert 'Number of members added 1' in cmd_output.stdout_text
+
+        # now check ssh on the client
+        clear_sssd_cache(client)
+        client.put_file_contents('/tmp/user_ssh_priv_key',
+                                 ssh_key_file_contents)
+        client.run_command('chmod 600 /tmp/user_ssh_priv_key')
+
+        try:
```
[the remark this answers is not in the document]
ssidhaye: Added a comment

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +406,182 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        assert 'Member of groups: child2' in cmd_output.stdout_text
+        assert assert_str in cmd_output.stdout_text
+        assert 'Number of members removed 1' in cmd_output.stdout_text
+
+        clear_sssd_cache(master)
+
+        # now check ssh on the client again
+        clear_sssd_cache(client)
+        try:
+            cmd = 'ssh -i {} -q {}@{} groups'.format('/tmp/user_ssh_priv_key',
```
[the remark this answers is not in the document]
ssidhaye: added a comment

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +406,182 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+
+        # user lookup
+        cmd_output = master.run_command(['ipa', 'user-show', username])
+        assert 'parent' in cmd_output.stdout_text
+        assert 'ipausers' in cmd_output.stdout_text
+        assert 'child1' in cmd_output.stdout_text
+        assert 'child2' in cmd_output.stdout_text
+
+        clear_sssd_cache(master)
+
```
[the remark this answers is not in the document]
ssidhaye: added a comment

ipatests/test_integration/test_sssd.py
flo-renaud: Please don't mark a comment as resolved unless you either apply the proposed change or argument why you disagree. As @frasertweedale I find the naming confusing, and think that subgroup / group / supergroup or even A / B / C would be better.

ssidhaye:
@flo-renaud I think it would be even more confusing if I use supergroup, subgroup and group instead of parent, child1 and child2

ipatests/test_integration/test_sssd.py
ssidhaye: changed the group names to a,b,c as requested.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +406,193 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        """
+        master = self.master
+        client = self.clients[0]
+
+        username = "testuser001"
+        userpasswd = 'Secret123'
+
+        # tmpdir for creating ssh keys for user
+        # since ssh does not work well paramiko
+        tmpdir = tempfile.mkdtemp(prefix="tmp-")
```
tiran, line 443: Please use the [tmpdir](https://docs.pytest.org/en/latest/tmpdir.html#the-tmpdir-fixture) fixture from pytest instead. This will take care of cleanup, too.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +406,193 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+
+        # add a user and set password
+        tasks.create_active_user(master, username, userpasswd)
+        tasks.kinit_admin(master)
+
+        pub_keys = []
+
+        ssh_key_pair = tasks.generate_ssh_keypair()
```
tiran, line 451: ``ssh_key_pair`` is cryptic, why not something like ``privkey, pubkey = tasks.generate_ssh_keypair()``?

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +406,193 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        pub_keys.append(ssh_key_pair[1])
+        with open(os.path.join(
+                tmpdir, 'ssh_priv_key'), 'w') as fp:
+            fp.write(ssh_key_pair[0])
+
+        keys_opts = ' '.join(['--ssh "{}"'.format(k) for k in pub_keys])
```
tiran, line 457: Don't construct commands as string. Always use a list, please.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +406,193 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        cmd = 'ipa user-mod {} {}'.format(username, keys_opts)
+        master.run_command(cmd)
+
+        # connect with first SSH key
+        user_ssh_priv_key = os.path.join(tmpdir, 'ssh_priv_key')
+
+        # change private key permission to comply with SS rules
```
tiran, line 464: What's ``SS`` rule? Why do you even write the key to a local file on the controller machine? Where does the controller use the local copy?

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +406,193 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        assert 'Member users: {}'.format(username) in cmd_output.stdout_text
+        assert 'Member groups: a' in cmd_output.stdout_text
+        assert 'Member of groups: c' in cmd_output.stdout_text
+        assert 'Number of members added 1' in cmd_output.stdout_text
+
+        # now check ssh on the client
+        clear_sssd_cache(client)
+        client.put_file_contents('/tmp/user_ssh_priv_key',
+                                 ssh_key_file_contents)
+        client.run_command('chmod 600 /tmp/user_ssh_priv_key')
```
tiran, line 546: ``['chmod', '600', '/tmp/user_ssh_priv_key']``

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +406,193 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+
+        # after adding testuser001 to b group
+        # testuser001 will have the following memberships
+        # Member of groups: a, ipausers, b
+        # Indirect Member of group: c
+        try:
+            cmd = 'ssh -i {} -q {}@{} groups'.format('/tmp/user_ssh_priv_key',
```
tiran, line 553: please use a list

ipatests/test_integration/test_sssd.py
ssidhaye: removed it

ipatests/test_integration/test_sssd.py
ssidhaye: simplified it

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: used a list

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,186 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        assert 'a' in cmd_output.stdout_text
+        assert 'ipausers' in cmd_output.stdout_text
+        assert 'b' in cmd_output.stdout_text
+        assert 'c' in cmd_output.stdout_text
+
+        clear_sssd_cache(client)
+
+        try:
+            cmd = ['ssh', '-i', '/tmp/user_ssh_priv_key',
+                   '-q', '{}@{}'.format(username, client.hostname)]
```
flo-renaud, line 516: You need to add a command at the end of ssh, for instance:
```
            cmd = ['ssh', '-i', '/tmp/user_ssh_priv_key',
                   '-q', '{}@{}'.format(username, client.hostname), 'true']
```
otherwise the ssh command will not return. Same comment for all the places calling ssh.

ssidhaye: thanks for pointing it out

ipatests/test_integration/test_sssd.py
ssidhaye: modified the command to use a list instead of strings

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,186 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        client = self.clients[0]
+
+        username = "testuser001"
+        userpasswd = 'Secret123'
+
+        # add a user and set password
+        tasks.create_active_user(master, username, userpasswd)
+        tasks.kinit_admin(master)
+
+        pub_keys = []
```
flo-renaud, line 444: The list pub_keys has no use, you can simplify the code and use only pubkey instead of pub_keys[0].

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,186 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+
+        privkey, pubkey = tasks.generate_ssh_keypair()
+        pub_keys.append(pubkey)
+        with open(os.path.join(
+                tmpdir, 'ssh_priv_key'), 'w') as fp:
+            fp.write(privkey)
+
+        master.run_command([
+            'ipa', 'user-mod', username, '--ssh', "{}".format(pub_keys[0])
```
flo-renaud, line 453: You can simplify as `'ipa', 'user-mod', username, '--ssh', pubkey)`

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,186 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        ])
+
+        # connect with first SSH key
+        user_ssh_priv_key = os.path.join(tmpdir, 'ssh_priv_key')
+
+        ssh_key_file_contents = ''
+        with open(user_ssh_priv_key, 'r') as infile:
+            ssh_key_file_contents = infile.read()
```
flo-renaud, line 461: No need to open + read the file, as you already have the content in the privkey variable. You can simply do
`master.put_file_contents('/tmp/user_ssh_priv_key', privkey)

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,186 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+
+        # clear sssd_cache
+        clear_sssd_cache(master)
+
+        # user lookup
+        # at this point, testuser001 has the following group memberships
+        # Member of groups: a, ipausers
+        # Indirect Member of group: b, c
+        cmd_output = master.run_command(['ipa', 'user-show', username])
+        assert 'a' in cmd_output.stdout_text
```
flo-renaud, line 507: The assert is too vague. It will match for instance the line "First n**a**me: testuser001". I would also advise to name the groups groupa, groupb and groupc to avoid this type of issue instead of a,b,c.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,186 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        # after removing testuser001 from b group
+        # testuser001 will have the following memberships
+        # Member of groups: a, ipausers
+        # Indirect Member of group: b, c
+        clear_sssd_cache(client)
+        try:
+            cmd = ['ssh', '-i', '/tmp/user_ssh_priv_key',
+                   '-q', '{}@{}'.format(username, client.hostname), 'groups']
+            cmd_output = client.run_command(cmd)
+            assert '{}'.format(username) in cmd_output.stdout_text
```
flo-renaud, line 577: Simplify with `assert username in cmd_output.stdout_text`

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,186 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+
+        # after adding testuser001 to b group
+        # testuser001 will have the following memberships
+        # Member of groups: a, ipausers, b
+        # Indirect Member of group: c
+        try:
+            cmd = ['ssh', '-i', '/tmp/user_ssh_priv_key',
+                   '-q', '{}@{}'.format(username, client.hostname), 'groups']
+            cmd_output = client.run_command(cmd)
+            assert '{}'.format(username) in cmd_output.stdout_text
```
flo-renaud, line 548: Simplify with `assert username in cmd_output.stdout_text`

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,186 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+            assert 'a' in cmd_output.stdout_text
+            assert 'b' in cmd_output.stdout_text
+            assert 'c' in cmd_output.stdout_text
+        except Exception as e:
+            pytest.fail('Authentication using SSH key not successful', e)
+
+        # test cleanup
```
flo-renaud, line 584: The cleanup should be executed even if the test fails. You can use a fixture or try/except.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,186 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        assert 'b' in cmd_output.stdout_text
+        assert 'c' in cmd_output.stdout_text
+
+        clear_sssd_cache(client)
+
+        try:
+            cmd = ['ssh', '-i', '/tmp/user_ssh_priv_key',
+                   '-q', '{}@{}'.format(username, client.hostname), 'groups']
+            cmd_output = master.run_command(cmd)
+            assert '{}'.format(username) in cmd_output.stdout_text
```
flo-renaud, line 518: Simplify with `assert username in cmd_output.stdout_text`

flo-renaud:
@ssidhaye 
I also forgot to mention that the sssd fix is not available in fedora27 but present in rhel 7.8+. You will need to add a xfail when backporting to the ipa-4-6 branch, based on the OS version.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,186 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+    @pytest.fixture
+    def nested_group_cleanup(self):
+        """Clean up groups and user created"""
+        # test cleanup
+        for group in ['groupa', 'groupb', 'groupc']:
+            self.master.run_command(['ipa', 'group-del', group, '--continue'])
+        self.master.run_command(['ipa', 'user-del', self.username,
+                                 '--no-preserve', '--continue'])
+        tasks.kdestroy_all(self.master)
+        tasks.kdestroy_all(self.client)
```
flo-renaud, line 429: Should be `self.clients[0]`

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,186 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+    userpasswd = 'Secret123'
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master)
+        tasks.install_client(cls.master, cls.clients[0])
+
+    @pytest.fixture
+    def nested_group_cleanup(self):
+        """Clean up groups and user created"""
```
flo-renaud, line 422: The fixture can be written like this:
```
@pytest.fixture
def nested_group_cleanup(self):
    # Do setup tasks, user creation, group creation etc...
    yield
    # Do cleanup tasks
```
so that the test gets executed after the setup tasks have been called and before the clean up tasks.

ipatests/test_integration/test_sssd.py
ssidhaye: removed and used pubkey instead

ipatests/test_integration/test_sssd.py
ssidhaye: simplified it

ipatests/test_integration/test_sssd.py
ssidhaye: used groupa, groupb and groupc

ipatests/test_integration/test_sssd.py
ssidhaye: removed the read file part

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: added cleanup as a fixture

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: modified as per suggestion

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,184 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        try:
+            cmd = ['ssh', '-i', '/tmp/user_ssh_priv_key',
+                   '-q', '{}@{}'.format(self.username, client.hostname),
+                   'groups']
+            cmd_output = master.run_command(cmd)
+            assert self.username in cmd_output.stdout_text
+            assert 'groupa' in cmd_output.stdout_text
+            assert 'groupb' in cmd_output.stdout_text
+            assert 'groupc' in cmd_output.stdout_text
+        except Exception as e:
```
flo-renaud, line 525: The try/except clause doesn't have much value here, you can simplify the code and remove it. If master.run_command fails, the test will fail anyway.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,184 @@ def test_trustdomain_disable_disables_subdomain(self):
...
+        try:
+            cmd = ['ssh', '-i', '/tmp/user_ssh_priv_key',
+                   '-q', '{}@{}'.format(self.username, client.hostname),
+                   'groups']
+            cmd_output = client.run_command(cmd)
+            assert self.username in cmd_output.stdout_text
+            assert 'groupa' in cmd_output.stdout_text
+            assert 'groupb' in cmd_output.stdout_text
+            assert 'groupc' in cmd_output.stdout_text
+        except Exception as e:
```
flo-renaud, line 557: Same comment as above, the try/catch clause doesn't add any value here.

ipatests/test_integration/test_sssd.py
```diff
@@ -403,3 +405,184 @@ def test_trustdomain_disable_disables_subdomain(self):
...
```
flo-renaud, line 587: Same comment as above, the try/catch clause doesn't add any value here.

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

flo-renaud:
Hi @ssidhaye 
tests are green and code LGTM. Please remove the temp commit and I will ack.

ssidhaye:
> Hi @ssidhaye
> tests are green and code LGTM. Please remove the temp commit and I will ack.

Thanks @flo-renaud 
Removed the temp commit

flo-renaud:
Tests are green and code already reviewed, ACK.

flo-renaud:
master:

* 47bddf4f458699f46b398d0969d5e3a348feed52 Test for removing a subgroup

ssidhaye:
#4705 Manual backport for ipa-4-6

PR #4545, opened 2020-04-17
Define default password policy for sysaccounts

tiran:
cn=sysaccounts,cn=etc now has a default password policy to permit system
accounts with krbPrincipalAux object class. This allows system accounts
to have a keytab that does not expire.

Fixes: https://pagure.io/freeipa/issue/8276

### DIFF
--- a/install/updates/20-default_password_policy.update
+++ b/install/updates/20-default_password_policy.update
@@ -1,6 +1,9 @@
-# Default password policies for hosts, services and Kerberos services
-# Setting all attributes to zero effectively disables any password policy
-# We can do this because hosts and services uses keytabs instead of passwords
+# Default password policies for hosts, services, system accounts, and
+# Kerberos services
+# Setting all attributes to zero effectively disables any password policy.
+# We can do this because hosts and services uses keytabs instead of
+# passwords. System accounts with krbPrincipalAux objectClass also use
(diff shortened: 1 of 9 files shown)


abbra:
@tiran I was thinking about it and I think we should create separate default policy for system accounts.  In the way similar to the discussion about application-specific passwords, I think we should define minimal password length to passwords of system accounts-based Kerberos principals. Since they are supposed to be used for keytabs but there is nothing that would limit them to set passwords via LDAP modifications, applying a policy that forces minimal password length is preferred.

Setting minimal password length will not affect random key generation in `ipa-getkeytab` but will prevent from using small passwords in both LDAP and `ipa-getkeytab -P` (setkeytab) operations.

tiran:
@abbra I have updated the policy ``cn=Default System Accounts Password Policy,cn=sysaccounts,cn=etc,$SUFFIX`` with ``default:krbPwdMinLength: 8``

abbra:
LGTM

tiran:
master:

* bb24641e8ffdc1083565b9de9496606cd458cf8d Use api.env.container_sysaccounts
* ca6d6781c741fe5aedc68f8de592b9806ebe21d7 Define default password policy for sysaccounts

PR #4547, opened 2020-04-17
Fixes to allow running the testsuite on Debian/Ubuntu

tjaalton:
These make running the testsuite on azure easier, and while it's not fully there yet, this bunch should be safe to apply upstream.

### DIFF
--- a/ipaplatform/debian/tasks.py
+++ b/ipaplatform/debian/tasks.py
@@ -206,4 +206,12 @@ Serial Number (hex): {cert.serial_number:#x}
 
         return True
 
+    # Debian doesn't use authselect, so call enable/disable_ldap_automount
+    # from BaseTaskNamespace.
+    def enable_ldap_automount(self, statestore):
+        return BaseTaskNamespace.enable_ldap_automount(self, statestore)
+
+    def disable_ldap_automount(self, statestore):
+        return BaseTaskNamespace.disable_ldap_automount(self, statestore)
+
 tasks = DebianTaskNamespace()
--- a/ipaplatform/debian/tasks.py
+++ b/ipaplatform/debian/tasks.py
@@ -72,10 +72,6 @@ class DebianTaskNamespace(RedHatTaskNamespace):
         # Debian doesn't have authselect
         return True
 
-    @staticmethod
-    def parse_ipa_version(version):
-        return BaseTaskNamespace.parse_ipa_version(version)
-
     def configure_httpd_wsgi_conf(self):
         # Debian doesn't require special mod_wsgi configuration
         pass
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -906,11 +906,10 @@ class TestIPACommand(IntegrationTest):
         if self.master.is_fips_mode:  # pylint: disable=no-member
             pytest.skip("paramiko is not compatible with FIPS mode")
 
-        sssd_version = ''
-        cmd_output = self.master.run_command(['sssd', '--version'])
-        sssd_version = platform_tasks.\
-            parse_ipa_version(cmd_output.stdout_text.strip())
-        if sssd_version.version < '2.2.0':
+        cmd = self.master.run_command(['sssd', '--version'])
+        sssd_version = platform_tasks.parse_ipa_version(
+            cmd.stdout_text.strip())
+        if sssd_version < platform_tasks.parse_ipa_version('2.2.0'):
             pytest.xfail(reason="sssd 2.2.0 unavailable in F29 nightly")
 
         username = "testuser" + str(random.randint(200000, 9999999))
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -607,24 +607,25 @@ class TestInstallMaster(IntegrationTest):
         other hand in case of restart it will change.
         """
         # listing all services
-        services = [
+        ipa_services_name = [
             "Directory", "krb5kdc", "kadmin", "named", "httpd", "ipa-custodia",
             "pki-tomcatd", "ipa-otpd", "ipa-dnskeysyncd"
         ]
 
         # checking the service status
         cmd = self.master.run_command(['ipactl', 'status'])
-        for service in services:
+        for service in ipa_services_name:
             assert f"{service} Service: RUNNING" in cmd.stdout_text
 
         # stopping few services
         service_stop = ["krb5kdc", "kadmin", "httpd"]
         for service in service_stop:
-            self.master.run_command(['systemctl', 'stop', service])
+            service_name = services.knownservices[service].systemd_name
+            self.master.run_command(['systemctl', 'stop', service_name])
 
         # checking service status
         service_start = [
-            svcs for svcs in services if svcs not in service_stop
+            svcs for svcs in ipa_services_name if svcs not in service_stop
         ]
         cmd = self.master.run_command(['ipactl', 'status'])
         for service in service_start:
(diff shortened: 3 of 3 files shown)


ipaplatform/debian/paths.py
```diff
@@ -58,7 +58,7 @@ class DebianPathNamespace(BasePathNamespace):
     SYSCONFIG_IPA_DNSKEYSYNCD = "/etc/default/ipa-dnskeysyncd"
     SYSCONFIG_IPA_ODS_EXPORTER = "/etc/default/ipa-ods-exporter"
     SYSCONFIG_KRB5KDC_DIR = "/etc/default/krb5-kdc"
-    SYSCONFIG_NAMED = "/etc/default/bind9"
+    SYSCONFIG_NAMED = "/etc/default/named"
```
stanislavlevin, line 61: Could it be conditional?
I mean this change could break the systems having BIND < 9.16 if any.

stanislavlevin:
Most likely, `TestIPACommand.test_hbac_systemd_user` will be fixed in https://github.com/freeipa/freeipa/pull/4528. You could rerun this job by manual or wait for the merging of that PR.

ipaplatform/ubuntu
```diff
@@ -0,0 +1 @@
+debian
```
tiran, line 1: Is this a symlink? GH UI doesn't clearly indicate the type.

It should not be necessary to add a symlink here. IPA tries ``ID_LIKE`` and Ubuntu should define ``ID_LIKE=debian`` in ``/etc/os-release``.

ipatests/test_integration/test_advise.py
```diff
@@ -141,7 +141,7 @@ def test_advice_enable_admins_sudo(self):
 
     def test_advice_config_server_for_smart_card_auth(self):
         advice_id = 'config_server_for_smart_card_auth'
-        advice_regex = r"\#\!\/bin\/sh.*"
+        advice_regex = r"\#\!\/bin\/bash.*"
```
stanislavlevin, line 144: Why do you change this?
`/bin/sh` is the expected value, while as I see, the actual one comes from https://github.com/freeipa/freeipa/blob/master/ipaserver/advise/base.py#L495 :
```
    def print_header(self, header, print_shell=False):
        header_size = len(header)

        prefix = ''
        if print_shell:
            prefix = '# '
            print('#!/bin/sh')
```
and it also should be `/bin/sh`.

ipaplatform/debian/paths.py
tjaalton: I don't think that's necessary, then again this could be dropped until there's actual support for 9.16

ipatests/test_integration/test_commands.py
```diff
@@ -884,11 +884,10 @@ def test_ssh_from_controller(self):
...
 
-        sssd_version = ''
-        cmd_output = self.master.run_command(['sssd', '--version'])
-        sssd_version = platform_tasks.\
-            parse_ipa_version(cmd_output.stdout_text.strip())
-        if sssd_version.version < '2.2.0':
+        cmd = self.master.run_command(['sssd', '--version'])
+        sssd_version = platform_tasks.parse_ipa_version(
+            cmd.stdout_text.strip())
+        if sssd_version <= platform_tasks.parse_ipa_version('2.2.0'):
```
stanislavlevin, line 891: Shouldn't it be like this?
```diff
- if sssd_version <= platform_tasks.parse_ipa_version('2.2.0'):
+ if sssd_version < platform_tasks.parse_ipa_version('2.2.0'):
```
Original comparison is 'if less than'.

tiran:
The platform specific skip code is rather ugly. How about a proper mark to skip tests on platforms?

```
index 4105c0b94..8d56efe15 100644
--- a/ipatests/conftest.py
+++ b/ipatests/conftest.py
@@ -19,6 +19,10 @@ try:
     import ipaplatform  # pylint: disable=unused-import
 except ImportError:
     ipaplatform = None
+    PLATFORM_IDS = set()
+else:
+    from ipaplatform.osinfo import osinfo
+    PLATFORM_IDS = set(osinfo.platform_ids)
 
 
 HERE = os.path.dirname(os.path.abspath(__file__))
@@ -42,6 +46,7 @@ MARKERS = [
     'ds_acceptance: Acceptance test suite for 389 Directory Server',
     'skip_ipaclient_unittest: Skip in ipaclient unittest mode',
     'needs_ipaapi: Test needs IPA API',
+    'skip_if_platform(platform, reason): Skip test on platform (ID and ID_LIKE)'
 ]
 
 
@@ -152,6 +157,12 @@ def pytest_runtest_setup(item):
             if item.config.option.skip_ipaapi:
                 pytest.skip("Skip tests that needs an IPA API")
 
+        for mark in item.iter_markers(name="skip_if_platform"):
+            platform = mark.kwargs["platform"]
+            reason = mark.kwargs["reason"]
+            if platform in PLATFORM_IDS:
+                pytest.skip(f"Skip test on platform {platform}: {reason}")
+
```

then you can do ``@pytest.mark.skip_if_platform(platform="debian", reason="Debian does not support crypto policy")``

ipaplatform/ubuntu
tjaalton: it is, and the failure came from pylint IIRC, build etc all would work fine without this

though it could be probably fixed in some other way too?

tjaalton:
> The platform specific skip code is rather ugly. How about a proper mark to skip tests on platforms?
> 

Yeah that looks good.

ipatests/test_integration/test_commands.py
tjaalton: oh right.. I must've messed it up with this revision by copying from the other place instead of my old 'hacks.diff'

stanislavlevin:
@tiran, this helps for skipping tests if the test's controller has the same os as the test's target. For runtime skipping based on OS, one should make a query to the remote host.

ipatests/test_integration/test_advise.py
tjaalton: ha, this might've been a mistake in the first place, unlike the other one which fixed actual tests

ipaplatform/debian/paths.py
tjaalton: I've dropped this from the set, there are other changes needed but I'll let master use 9.11 until the test suite passes

ipatests/test_integration/test_advise.py
tjaalton: dropped

ipaplatform/ubuntu
tjaalton: I've dropped it for now, need to check where it actually failed and then send separately to see if it can be fixed in some other way

tjaalton:
so I've kept the test skipping as they were for now

ipaplatform/debian/tasks.py
```diff
@@ -206,4 +202,10 @@ def remove_ca_certificates_dir(self, directory):
 
         return True
 
+    def enable_ldap_automount(self, statestore):
+        return BaseTaskNamespace.enable_ldap_automount(self, statestore)
```
tiran, line 206: Could you please add a comment here to explain why you are calling the methods from the base namespace?

ipaplatform/ubuntu
tiran: It still shows up.

tiran:
Latest pylint is causing issues. You'll have to rebase as soon as https://github.com/freeipa/freeipa/pull/4597 has been merged.

tiran:
master:

* 7d657c6faa4f090d5f672502f9f62d64e74c355d Debian: Use enable/disable_ldap_automount() from base
* 7ed5374c91dd7e69b2ac0222cf7dc5d48df75c19 Debian: Use parse_ipa_version from redhat.
* 158257c4b3f0eb58db73ac667aae4feaef21b58b ipatests/test_commands: Check sssd version like on test_sssd
* 2e85b4809afb4757fd7b81868709ced4f211d7b4 ipatests/test_installation: Use knownservices to map the service name.

PR #4560, opened 2020-04-20
[ipa-4-6] ipatests: fix teardown method

flo-renaud:
The teardown method of the integration tests is calling
  tasks.uninstall_master(replica)
which in turn does ipa server-del + ipa-server-install --uninstall

The uninstallation is called by default with --ignore-last-of-role
and --ignore-topology-disconnect excepted if the domain-level is 0,
in which case these options are not supported by uninstall.
    
The issue is that the call to get domain level fails because the
server has already been removed, and the code handles that failure by
assuming domain level =0, resulting in missing --ignore-last-of-role.

To workaround the issue, the call to tasks.uninstall should be done
with the actual domain level:
  tasks.uninstall_master(replica, domain_level=...)

Fixes: https://pagure.io/freeipa/issue/8280

### DIFF
--- a/ipatests/test_integration/base.py
+++ b/ipatests/test_integration/base.py
@@ -79,6 +79,10 @@ class IntegrationTest(object):
 
     @classmethod
     def uninstall(cls, mh):
+        if cls.domain_level is not None:
+            domain_level = cls.domain_level
+        else:
+            domain_level = cls.master.config.domain_level
         for replica in cls.replicas:
             try:
                 tasks.run_server_del(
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @wladich 
could you review this PR? Thanks

wladich:
@flo-renaud 
Thank you for the fix.

Do you have an idea why the failure in issue 8280 is intermittent?

flo-renaud:
@wladich 
> Do you have an idea why the failure in issue 8280 is intermittent?

I suppose that has to do with replication having time to replicate the server-del operation or not, but it's only a wild guess.

flo-renaud:
Temp commit removed.

kaleemsiddiqu:
@flo-renaud 
Should not we do the same domain level changes for run_server_del function definition and its call?

flo-renaud:
@kaleemsiddiqu 
> Should not we do the same domain level changes for run_server_del function definition and its call?
The server-del command always accept the ignore-last-of-role and ignore-topogy-disconnect parameters, I don't think any change is needed.

kaleemsiddiqu:
> @kaleemsiddiqu
> 
> > Should not we do the same domain level changes for run_server_del function definition and its call?
> 
> The server-del command always accept the ignore-last-of-role and ignore-topogy-disconnect parameters, I don't think any change is needed.

OK. then LGTM

flo-renaud:
ipa-4-6:

* bbec79845e060d9b0a523a9502316f200f4302fa ipatests: fix teardown method

PR #4564, opened 2020-04-21
Makefile.am: add doclint to fastcheck

fcami:
Add doclint to fastcheck so that documentation syntax issues
are caught sooner (before they hit CI).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -327,6 +327,16 @@ yamllint:
 	done; \
 	echo "-----------"
 
+# Build & lint documentation.
+#
+.PHONY: doclint
+doclint:
+	@echo -e "\nBuild and lint documentation";
+	@echo "-----------";
+	@$(MAKE) -C $(srcdir)/doc/ clean;
+	@$(MAKE) -C $(srcdir)/doc/ html; 
+	@echo "-----------"
+
 # Run pylint for all python files. Finds all python files/packages, skips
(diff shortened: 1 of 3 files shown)


Makefile.am
```diff
@@ -327,6 +327,15 @@ yamllint:
 	done; \
 	echo "-----------"
 
+# Build & lint documentation.
+#
+.PHONY: doclint
+doclint:
+	echo -e "\nBuild and lint documentation"; \
+	echo "-----------"; \
+	$(MAKE) -C $(srcdir)/doc/ html SPHINXOPTS="-W"; \
```
tiran, line 336: This won't report any errors if the user has built documentation before. You have to clean first, then build.

I'm not sure it's a good idea to have this part of fastcheck. As soon as we have more documentation, it can easily take a minute or two to rebuild the documentation.

fcami: OK, I've added the clean step and removed it from fastcheck.

fcami:
hi @tiran thanks for the review, I've removed the doclint target from fastcheck and added -W to sphinx-build by default.

tiran:
master only
4.8 doesn't have sphinx build system

fcami:
master:

* 51d15176a4434b9c47aa2b99476f3137a361d30c Makefile.am: add doclint to fastcheck
* 7558e1413dca240a77893352b5bb62bd21103b46 doc/Makefile: use sphinx-build -W by default

PR #4573, opened 2020-04-22
ipatests: Test to check IPADomainCheck without trust setup.

menonsudhir:
This testcase checks the behaviour of IPADomainCheck within the healthcheck tool
when there is no trust setup between IPA and AD.

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -565,6 +565,22 @@ class TestIpaHealthCheck(IntegrationTest):
             ruvs.remove(check["kw"]["ruv"])
         assert not ruvs
 
+    def test_ipa_healthcheck_without_trust_setup(self):
+        """
+        This testcase checks that when trust isn't setup between IPA
+        server and Windows AD, IPADomainCheck displays key value as
+        domain-check and result is SUCCESS
+        """
+        returncode, data = run_healthcheck(
+            self.master,
+            "ipahealthcheck.ipa.trust",
+            "IPADomainCheck"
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -532,6 +532,26 @@ def test_ipa_dns_systemrecords_check(self):
...
+        between IPA server and Windows AD, IPADomainCheck
+        displays key value as domain-check and result is SUCCESS
+        """
+        msg = '0 trusts matched'
+        result = self.master.run_command(
+            ["ipa", "trust-find"], raiseonerr=False
+        )
+        assert result.returncode == 1
+        assert msg in result.stdout_text
+        returncode, data = run_healthcheck(
```
rcritten, line 547: You can probably remove the trust-find bit since this check runs whether trust is enabled or not.

This check ensures that the provide in sssd.conf for the IPA domain is "ipa".

rcritten:
Other than that this LGTM.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: @rcritten removed the trust-find command from the code.

rcritten:
ipatests/test_integration/test_ipahealthcheck.py:559: [E0001(syntax-error), ] invalid syntax (<unknown>, line 559))

rcritten:
master:

* ba213aa448a9b568da4ba9920a4b513406bd3964 ipatests: Test for ipahealthcheck tool for IPADomainCheck.

PR #4574, opened 2020-04-22
Test for check of HostKeyAlgorithms option in ssh_config

kaleemsiddiqu:
Test checks that HostKeyAlgorithms is not present in
/etc/ssh/ssh_config after client install with option
-ssh-trust-dns.

https://pagure.io/freeipa/issue/8082

Signed-off-by: Kaleemullah Siddiqui <ksiddiqu@redhat.com>

### DIFF
--- a/ipatests/test_integration/test_installation_client.py
+++ b/ipatests/test_integration/test_installation_client.py
@@ -1,0 +1,36 @@
+#
+# Copyright (C) 2020  FreeIPA Contributors see COPYING for license
+#
+
+"""
+Module provides tests for various options of ipa-client-install.
+"""
+
+from __future__ import absolute_import
+
+from ipatests.test_integration.base import IntegrationTest
+from ipatests.pytest_ipa.integration import tasks
+
+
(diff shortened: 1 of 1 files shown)


kaleemsiddiqu:
@mrizwan93 
> 2 observations:
> 
> 1. Can we test it having `fips_mode = True`?
This should work in fips/non-fips both env thats why i did not make it specific to fips env. that will be taken care in downstream.
> 2. Can't it be part of `test_installation.py` or `test_fips.py`?
test_installation.py is having Server/replica related tests. This will have tests specific installs for client scenarios only. for fips same justificaiton as above.

mrizwan93:
Okay. LGTM.

rcritten:
Does this need to be backported to any branches?

kaleemsiddiqu:
> Does this need to be backported to any branches?

Yes . Thx for adding those labels.

rcritten:
master:

* bba41dc85c8427992b5626b5a9daaf86c3b2a812 Test for check of HostKeyAlgorithms option in ssh_config

PR #4577, opened 2020-04-23
ipatests: Test for ipahealthcheck ReplicationConflictCheck

menonsudhir:
This test checks for particular objectclass (&(!(objectclass=nstombstone))(nsds5ReplConflict=*)) entry in IPA Master ldap database and when the entry is not found, it checks that ipahealthcheck tool with
ReplicationConflictCheck displays the result as SUCCESS.

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
--- a/ipatests/prci_definitions/temp_commit.yaml
+++ b/ipatests/prci_definitions/temp_commit.yaml
@@ -68,7 +68,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-latest/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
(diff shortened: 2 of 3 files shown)


rcritten:
```
ipatests/test_integration/test_ipahealthcheck.py:346: [E0602(undefined-variable), T```
estIpaHealthCheck.test_source_ipahealthcheck_ipa_host_check_ipahostkeytab] Undefined variable 'realm_to_serverid')
ipatests/test_integration/test_ipahealthcheck.py:579: [E1101(no-member), TestIpaHealthCheck.test_ipa_replication_conflict_check] Instance of 'ldap_connect' has no 'get_entries' member)
ipatests/test_integration/test_ipahealthcheck.py:579: [W0612(unused-variable), TestIpaHealthCheck.test_ipa_replication_conflict_check] Unused variable 'entries')
ipatests/test_integration/test_ipahealthcheck.py:17: [W0611(unused-import), ] Unused ipautil imported from ipapython)
```
I'm thinking this approach isn't going to work. It's more than likely that you'll get no results so the test will effectively just skip. In the case you do get results, which mean there is a conflict, then healthcheck won't return SUCCESS.

I found if quite difficult to force replication errors into DS and had to resort to a lot of starting and stopping, running scripts and swearing a lot. I've never tried it but perhaps try loading a conflict via ldif would work.

menonsudhir:
@rcritten thanks for the review, will try loading a conflict via ldif.

PR #4589, opened 2020-04-25
Rework named config upgrade (Issue 8287)

Carbenium:
https://pagure.io/freeipa/issue/8287

### DIFF
--- a/install/share/bind.named.conf.template
+++ b/install/share/bind.named.conf.template
@@ -18,9 +18,6 @@ options {
 
 	dnssec-enable yes;
 
-	/* Path to ISC DLV key */
-	bindkeys-file "$BINDKEYS_FILE";
-
 	managed-keys-directory "$MANAGED_KEYS_DIR";
 
 	/* crypto policy snippet on platforms with system-wide policy. */
--- a/install/share/bind.named.conf.template
+++ b/install/share/bind.named.conf.template
@@ -12,8 +12,9 @@ options {
 	// "allow-recursion" are set to "localnets; localhost;".
 	// If either "allow-query-cache" or "allow-recursion" is set,
 	// the other would be set the same value.
-	// Please refer to $CUSTOM_CONFIG
+	// Please refer to $CUSTOM_OPTIONS_CONFIG
 	// for more informations
+	include "$CUSTOM_OPTIONS_CONFIG";
 
 	tkey-gssapi-keytab "$NAMED_KEYTAB";
 	pid-file "$NAMED_PID";
--- a/install/share/bind.named.conf.template
+++ b/install/share/bind.named.conf.template
@@ -1,7 +1,4 @@
 options {
-	// turns on IPv6 for port 53, IPv4 is on by default for all ifaces
-	listen-on-v6 {any;};
-
 	// Put files that named is allowed to write in the data/ directory:
 	directory "$NAMED_VAR_DIR"; // the default
 	dump-file		"${NAMED_DATA_DIR}cache_dump.db";
@@ -20,7 +17,6 @@ options {
 	pid-file "$NAMED_PID";
 
 	dnssec-enable yes;
-	dnssec-validation yes;
 
 	/* Path to ISC DLV key */
 	bindkeys-file "$BINDKEYS_FILE";
--- a/install/share/bind.named.conf.template
+++ b/install/share/bind.named.conf.template
@@ -1,3 +1,6 @@
+/* WARNING: This part of the config file is IPA-managed.
+ * Modifications may break IPA setup or upgrades.
+ */
 options {
 	// Put files that named is allowed to write in the data/ directory:
 	directory "$NAMED_VAR_DIR"; // the default
@@ -45,9 +48,6 @@ include "$ROOT_KEY";
 /* custom configuration snippet */
 include "$CUSTOM_CONFIG";
 
-/* WARNING: This part of the config file is IPA-managed.
- * Modifications may break IPA setup or upgrades.
- */
 dyndb "ipa" "$BIND_LDAP_SO" {
 	uri "ldapi://%2fvar%2frun%2fslapd-$SERVER_ID.socket";
 	base "cn=dns, $SUFFIX";
--- a/install/share/bind.named.conf.template
+++ b/install/share/bind.named.conf.template
@@ -16,8 +16,6 @@ options {
 	tkey-gssapi-keytab "$NAMED_KEYTAB";
 	pid-file "$NAMED_PID";
 
-	dnssec-enable yes;
-
 	managed-keys-directory "$MANAGED_KEYS_DIR";
 
 	/* crypto policy snippet on platforms with system-wide policy. */
--- a/ipatests/test_ipaserver/test_install/test_bindinstance.py
+++ b/ipatests/test_ipaserver/test_install/test_bindinstance.py
@@ -2,95 +2,89 @@
 # Copyright (C) 2018  FreeIPA Contributors.  See COPYING for license
 #
 
-from __future__ import absolute_import
-
+import os
 import tempfile
+from pathlib import Path
+from unittest.mock import patch
 
 import pytest
 
+from ipalib import api
 from ipaplatform.paths import paths
-from ipaserver.install.server.upgrade import named_add_crypto_policy
+from ipaserver.install import bindinstance as bind_i
+from ipaserver.install.server.upgrade import named_add_ipa_ext_conf_file
 
-try:
-    from unittest.mock import patch  # pylint: disable=import-error
-except ImportError:
-    from mock import patch  # pylint: disable=import-error
-
-
-TEST_CONFIG = """
-options {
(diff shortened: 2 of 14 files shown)


abbra:
@Carbenium do you have some time to improve this PR as requested? If not, we can do that for you.

Carbenium:
Working on it right now. Changes you requested are done. Currently trying to come up with some test(s).

abbra:
@Carbenium thanks. Please also follow @tiran requests and don't remove existing tests but rather add new ones. I did reproduce the issue you are seeing too and I mostly agree with the split of include files because `options {}` cannot be defined multiple times.

install/share/bind.named.conf.template
```diff
@@ -12,14 +9,14 @@ options {
...
-	// Please refer to $CUSTOM_CONFIG
+	// Please refer to $CUSTOM_OPTIONS_CONFIG
 	// for more informations
+	include "$CUSTOM_OPTIONS_CONFIG";
 
 	tkey-gssapi-keytab "$NAMED_KEYTAB";
 	pid-file "$NAMED_PID";
 
 	dnssec-enable yes;
-	dnssec-validation yes;
```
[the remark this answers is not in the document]
Carbenium: Done.

Carbenium:
Pushed out a new version.

I tried to come up with some meaningful tests. They should at least catch if somebody adds a new variable to the config file and forgets to add it in the code.
The unit test which were removed are related to code which is also removed with this patch.

ipatests/test_ipaserver/test_install/test_bindinstance.py
```diff
@@ -2,95 +2,84 @@
...
 from ipaplatform.paths import paths
-from ipaserver.install.server.upgrade import named_add_crypto_policy
+from ipaserver.install import bindinstance as bind_i
+from ipaserver.install.server.upgrade import named_add_ipa_ext_conf_file
 
-try:
-    from unittest.mock import patch  # pylint: disable=import-error
-except ImportError:
-    from mock import patch  # pylint: disable=import-error
+share_dir = Path(__file__).parent.parent.parent.parent / 'install/share'
```
abbra, line 16: This doesn't work in CI. You should never reference in-tree part because in CI we install ipatests as a package and expects files to exist in their proper locations.

abbra:
Tests are failing:

```
test_ipaserver/test_install/test_bindinstance.py:65: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../ipaserver/install/bindinstance.py:310: in named_add_ext_conf_file
    ipa_ext_txt = ipautil.template_file(src, t_params)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

infilename = PosixPath('/usr/lib/python3.7/site-packages/install/share/bind.ipa-options-ext.conf.template')
vars = {'NAMED_DNSSEC_VALIDATION': 'yes'}

    def template_file(infilename, vars):
        """Read a file and perform template substitutions"""
>       with open(infilename) as f:
E       FileNotFoundError: [Errno 2] No such file or directory: '/usr/lib/python3.7/site-packages/install/share/bind.ipa-options-ext.conf.template'

../ipapython/ipautil.py:320: FileNotFoundError
```

this is because you should never refer paths in source tree in actual tests -- in CI they got installed as a normal package, so you should always reference proper paths.

abbra:
Existing tests around named conf should not be removed. They test, among various things, that crypto policies are in place.

ipatests/test_ipaserver/test_install/test_bindinstance.py
Carbenium: Right, wasn't sure about that one.
Unfortunately, with proper paths local testing (e.g ```make fastest```) fails. Is there something in place to allow for both testing methods to succeed?

Carbenium:
As I mentioned before, the removed unit tests reference code which is removed with this patch (```named_add_crypto_policy```). So I'm not sure how to keep them.

The only remaining code regarding crypto policies is
```python
        if paths.NAMED_CRYPTO_POLICY_FILE is not None:
            crypto_policy = 'include "{}";'.format(
                paths.NAMED_CRYPTO_POLICY_FILE
            )
        else:
            crypto_policy = "// not available"
```
and the template substitution of course.

ipatests/test_ipaserver/test_install/test_bindinstance.py
abbra: you can always test in a container like our Azure Pipelines CI does -- build packages with 'make rpms', then instantiate a container and install packages in it. With podman it is pretty simple and similar to
```
podman run -ti --rm --name my-test --hostname master.ipa.test -v /path/to/sources:/source f31/fedora-toolbox /sbin/init
(in a separate console):
podman exec -ti my-test /bin/bash
# cd /source
# ...
```

Carbenium: Alright. So you don't have a hard requirement that ```make (fast)check``` passes?

abbra: `make fast{lint,check}` always pass for me locally in the tree before I submit. However, any of the changes submitted must pass CI tests as well. There are several targets (we run 23 test suites in a pull request CI, 126 test suites in nightly tests), and they got called separately as well:

- `make fastlint` is called in Azure Pipelines' `Build' job
- `make lint` and `make -C doc/ html` called in `Lint` job
- there is separate Tox job which runs tox tests for client code
- web UI unit tests run in `WebUI_Unit_Tests` job

`make fastcheck` runs bunch of linting jobs that are already covered above and so-called `fasttest`. `make fasttest` runs very basic tests that do not require presence of installed IPA server. The latter will get covered by the `BASE_XMLRPC` test jobs in Azure Pipelines CI.
The other PR CI tests run separate machines (either containers in Azure Pipelines or VMs in PR CI) so they can cover functionality where full installation is required.

So yes, we cover most of the same tests in CI as you can call locally. You'd waste CI capacity with failures at worst but none of the tests should require presence of the source tree content the way how your code does. If you want to do unit testing of the code that replaces named.conf content, then I'd suggest to generate a local configuration file / templates and use them for unit testing. See, for example, `ipatests/test_ipaclient/test_ldapconf.py` how to do that.

Carbenium:
@abbra @tiran 
Unfortunately, I'm a bit stuck with the unit testing.
Let's summarize: The tests I wrote depend (IMHO they should) on the config file templates we ship. This is to ensure the unit tests not only catch code errors but also changes in the shipped templates.

**The problem**: Path handling has to accommodate for in-tree testing (```make ...```) and testing with packages installed (```ipa-run-tests```).
I tried to base this distinction on ```api.env.in_tree```.  Unfortunately ```api.env.in_tree``` is hard-coded to ```True``` in [ipatests/conftest.py](https://github.com/freeipa/freeipa/blob/master/ipatests/conftest.py#L119) for any test run.
In the last commit of this PR i tried to make it configurable. This works nicely for the simple unit tests (no API), but breaks for the others in ```in_tree=False``` case.
I could back-track it to [remote_plugins/__init__](https://github.com/freeipa/freeipa/blob/master/ipaclient/remote_plugins/__init__.py#L120) being the cause of trouble. Some more debugging revealed the accepted params by plugins are not the same when loaded via ```from ipaserver import plugins``` respectively as remote plugin.

e.g accepted params  for ```hbacrule-add``` in the in-tree case:
```
cn
accessruletype
all
raw
no_members
usercategory
hostcategory
sourcehostcategory
servicecategory
description
ipaenabledflag
externalhost
setattr
addattr
version
```
accepted params in the remote plugin case:
```
cn
all
raw
no_members
usercategory
hostcategory
servicecategory
description
setattr
addattr
version
```
(```no_option``` flag is the difference)

The unit tests which are failing in the out-of-tree case are depending on params which are not available in the remote plugin.

**Questions**:

- Is the way how I expect the unit tests to work wrong?
- Did I misunderstand the purpose of the ```in_tree``` flag?
- Is the difference in parameters for local loading and remote loading intended?

Thanks for the help so far - learned a lot!
If this doesn't fit into your schedule feel free to take any code you seem suitable from this PR and integrate in a way you seem fitting.

abbra:
`in_tree` is mostly used to ignore global IPA configuration in case your system is enrolled in IPA domain (like mine).
We are investigating what to do. @tiran found that few other places have broken logic too.
At worst, we can set `abs_top_srcdir` in the environment executed from `make fastcheck` or similar, to trigger changes in `ipa-run-tests`.

flo-renaud:
Hi @Carbenium 
can we close this PR since commit https://pagure.io/freeipa/c/f52a15b808aa4d827c1f77f807611d49c9563d03?branch=master was pushed?

Carbenium:
Ah, yes. That commit should cover it.

PR #4597, opened 2020-04-27
Fix new pylint warnings

tiran:
### Fix exception escape warning
    
``W1661(exception-escape), RPCClient.forward]
Using an exception object that was bound by an except handler)``

### Fix APIVersion.__getnewargs__

``E0312(invalid-getnewargs-returned), APIVersion.__getnewargs__]     __getnewargs__ does not return a tuple)``

### DIFF
--- a/ipalib/rpc.py
+++ b/ipalib/rpc.py
@@ -1170,11 +1170,11 @@ class RPCClient(Connectible):
                     try:
                         principal = getattr(context, 'principal', None)
                         delete_persistent_client_session_data(principal)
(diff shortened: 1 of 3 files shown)


abbra:
LGTM

rcritten:
master:

* 49f909c9d3abb08a9f6492fbdfeb33534889e37f Fix APIVersion.__getnewargs__
* 24cc13db898e07523227d04af9a15ea90eeacc3c Fix exception escape warning

PR #4605, opened 2020-04-28
Address issues found by new pylint 2.5.0

tiran:
* fix multiple exception-escape
* fix function signatures of DsInstance start/stop/restart
* silence f-string-without-interpolation
* silence import error in Azure script
* silence too-many-function-args

Fixes: https://pagure.io/freeipa/issue/8297

### DIFF
--- a/ipaserver/install/dsinstance.py
+++ b/ipaserver/install/dsinstance.py
@@ -652,21 +652,27 @@ class DsInstance(service.Service):
             # Does not apply with newer DS releases
             pass
 
-    def start(self, *args, **kwargs):
-        super(DsInstance, self).start(*args, **kwargs)
+    def start(self, instance_name="", capture_output=True, wait=True):
+        super(DsInstance, self).start(
+            instance_name, capture_output=capture_output, wait=wait
+        )
         api.Backend.ldap2.connect()
 
-    def stop(self, *args, **kwargs):
+    def stop(self, instance_name="", capture_output=True):
         if api.Backend.ldap2.isconnected():
             api.Backend.ldap2.disconnect()
 
-        super(DsInstance, self).stop(*args, **kwargs)
+        super(DsInstance, self).stop(
+            instance_name, capture_output=capture_output
+        )
 
-    def restart(self, instance=''):
+    def restart(self, instance_name="", capture_output=True, wait=True):
         api.Backend.ldap2.disconnect()
         try:
-            super(DsInstance, self).restart(instance)
-            if not is_ds_running(instance):
+            super(DsInstance, self).restart(
+                instance_name, capture_output=capture_output, wait=wait
+            )
+            if not is_ds_running(instance_name):
                 logger.critical("Failed to restart the directory server. "
                                 "See the installation log for details.")
                 raise ScriptError()
--- a/pylintrc
+++ b/pylintrc
@@ -102,6 +102,7 @@ disable=
     consider-using-enumerate,  # pylint 2.1, clean up tests later
     no-else-raise, # python 2.4.0
     import-outside-toplevel, # pylint 2.4.2
+    f-string-without-interpolation,  # pylint 2.5.0, bare f-strings are ok
 
 [REPORTS]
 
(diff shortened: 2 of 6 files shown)


ipaserver/install/dsinstance.py
```diff
@@ -652,21 +652,23 @@ def __add_default_schemas(self):
             # Does not apply with newer DS releases
             pass
 
-    def start(self, *args, **kwargs):
-        super(DsInstance, self).start(*args, **kwargs)
+    def start(self, instance_name="", capture_output=True, wait=True):
+        super().start(instance_name, capture_output=capture_output, wait=wait)
```
stanislavlevin, line 656: I see `ipa-4-6` label, this assumes Python2. `super()` is Python3 feature [0]

[0]: https://docs.python.org/3/whatsnew/3.0.html#builtins

tiran: good call, fixed!

pylintrc
```diff
@@ -102,6 +102,8 @@ disable=
     consider-using-enumerate,  # pylint 2.1, clean up tests later
     no-else-raise, # python 2.4.0
     import-outside-toplevel, # pylint 2.4.2
+    f-string-without-interpolation,  # pylint 2.5.0, bare f-strings are ok
```
stanislavlevin, line 105: Yes, bare f-strings are ok in general, but they could hide bugs.

For example,
```python
subst = "WORLD"
s = f"HELLO,subst"
```
What if I just forgot to add braces? Reviewers (people) don't see that, I think.
There is only one error of this type in code.

pylintrc
```diff
@@ -102,6 +102,8 @@ disable=
...
+    too-many-function-args,  # pylint 2.5.0, should be fixed
```
stanislavlevin, line 106: This can be fixed with:
```diff
--- a/ipatests/pytest_ipa/integration/host.py
+++ b/ipatests/pytest_ipa/integration/host.py
@@ -134,7 +134,7 @@ class Host(pytest_multihost.host.Host):
         else:
             cls = Host
 
-        return cls(domain, hostname, role, ip, external_hostname)
+        return cls(domain, hostname, role, ip=ip, external_hostname=external_hostname)
 
     def ldap_connect(self):
         """Return an LDAPClient authenticated to this host as directory manager

```

pylintrc
tiran: What about 
```
subst = "WORLD"
s = "Hello {subst}"
```
Is the second line a bug or not?

IPA is going to use more f-strings over time. CPython upstream is considering to make all strings f-strings and introduce a new prefix for plain strings. We (CPython core) are encouraging people to use f-strings everywhere.

ipatests/azure/scripts/setup_containers.py
```diff
@@ -2,7 +2,7 @@
 import os
 import subprocess
 
-import docker
+import docker  # pylint: disable=import-error
```
stanislavlevin, line 5: `python3-docker` is the 'lint' dependency from the spec.

stanislavlevin:
@tiran, please, see inline comments.

ipatests/azure/scripts/setup_containers.py
tiran: Ah, I see. I've removed the change.

pylintrc
tiran: Good idea, fixed!

pylintrc
stanislavlevin: Of course, two examples can be buggy: missing `f` or `{}`, and neither default type of string helps.
The difference is that `pylint` can prevent the bug in my example, but can't in yours.

stanislavlevin:
I've verified this PR locally using Pylint 2.5.0.
There are no warnings/errors.

pylintrc
stanislavlevin: @tiran, I like 'f-strings'.
It seems that my argument is too weak for CPython core :)

stanislavlevin:
ACK.

fcami:
master:

* 9941c9ee955ea52a48d9dc4d991c2292a839ae52 Address issues found by new pylint 2.5.0

fcami:
@tiran both automatic backports failed. Could you please create the backport PRs manually?

PR #4609, opened 2020-04-28
WIP: test fix for KDC crash of host principal aliases

abbra:


abbra:
Forgot to commit actual test instead of a template. ;)

daemons/ipa-kdb/ipa_kdb_mspac.c
```diff
@@ -1798,8 +1798,12 @@ static krb5_error_code ipadb_verify_pac(krb5_context context,
         priv_key = krbtgt_key;
     }
 
-    kerr = krb5_pac_verify(context, old_pac, authtime,
-                            client_princ, srv_key, priv_key);
+    /* only pass with_realm TRUE when it is cross-realm ticket and S4U2Self
```
iboukris, line 1801: Technically, cross-realm S4U as this also applies to cross-realm S4U2Proxy (aka RBCD).

daemons/ipa-kdb/ipa_kdb_mspac.c
```diff
@@ -2188,6 +2201,12 @@ krb5_error_code ipadb_sign_authdata(krb5_context context,
         make_ad = true;
     }
 
+#ifdef KRB5_KDB_FLAG_ISSUING_REFERRAL
```
iboukris, line 2204: this could use a comment to clarify being specific to S4U2Self path.

daemons/ipa-kdb/ipa_kdb_mspac.c
```diff
@@ -2188,6 +2201,12 @@ krb5_error_code ipadb_sign_authdata(krb5_context context,
...
+    if (!is_as_req && (make_ad && !(flags & KRB5_KDB_FLAG_ISSUING_REFERRAL))) {
```
iboukris, line 2205: this should actually be KRB5_KDB_FLAG_CROSS_REALM, and no need of ifdef.

daemons/ipa-kdb/ipa_kdb_mspac.c
```diff
@@ -1964,8 +1970,15 @@ static krb5_error_code ipadb_sign_pac(krb5_context context,
...
-    kerr = krb5_pac_sign(context, pac, authtime, client_princ,
-                         server_key, right_krbtgt_signing_key, pac_data);
+#ifdef KRB5_KDB_FLAG_ISSUING_REFERRAL
+    is_issuing_referral = (flags & KRB5_KDB_FLAG_ISSUING_REFERRAL) != 0;
+#endif
+
+    /* only pass with_realm TRUE when it is cross-realm ticket and S4U2Self
+     * was requested */
+    kerr = krb5_pac_sign_ext(context, pac, authtime, client_princ, server_key,
+                             right_krbtgt_signing_key, is_issuing_referral,
```
iboukris, line 1980: this is missing the condition on S4U, like: is_issuing_referral && (flags & KRB5_KDB_FLAG_PROTOCOL_TRANSITION)

daemons/ipa-kdb/ipa_kdb_mspac.c
```diff
@@ -2188,6 +2201,11 @@ krb5_error_code ipadb_sign_authdata(krb5_context context,
         make_ad = true;
     }
 
+    /* Only applies to S4U2Self over TGS request over cross-realm path */
+    if (!is_as_req && (make_ad && (flags & KRB5_KDB_FLAG_CROSS_REALM))) {
```
iboukris, line 2205: i'd just stuff if above, to make it easier to follow:
```c
    if ((is_as_req && (flags & KRB5_KDB_FLAG_INCLUDE_PAC)) ||
        ((flags & KRB5_KDB_FLAG_PROTOCOL_TRANSITION)
          && !(flags & KRB5_KDB_FLAG_CROSS_REALM)) {
        make_ad = true;
    }

iboukris:
Hmm, what version of MIT is this testing, we need to be careful with KRB5_KDB_FLAG_CROSS_REALM, we might need updated library version, see upstream commit (and message) e12e890f063f41bf8aef45e44a3ee329f64139d2.
Maybe actually to use the current is_cross_realm flag instead..

iboukris:
I mean just copy the other is_cross_realm code, instead of using KRB5_KDB_FLAG_CROSS_REALM, from: https://github.com/freeipa/freeipa/blob/45b4f5377bf3921406271148e18a3b99acfee03b/daemons/ipa-kdb/ipa_kdb_mspac.c#L1790

iboukris:
As discussed, we can't used CROSS_REALM flag nor is_cross_realm_krbtgt(), instead we can check if client==NULL. With these additional two patches cross-realm S4U2Self works both ways with PACs!
https://github.com/iboukris/freeipa/commits/iboukris-ab-xrealm-s4u2self

PR #4614, opened 2020-04-29
ipatests: add test for kdcproxy handling reply split to several TCP packets

wladich:
This is a regression test for the bug in python-kdcproxy mentioned in
    https://github.com/latchset/kdcproxy/pull/44
      When the reply from AD is split into several TCP packets the kdc
      proxy software cannot handle it and returns a false error message
      indicating it cannot contact the KDC server.
    
    This could be observed as login failures of AD user on IPA clients
    when:
    * IPA client was configured to use kdcproxy to communicate with AD
    * kdcproxy used TCP to communicate with AD
    * response from AD to kdcproxy was split into several packets
    
    This patch also refactors and improves existing tests:
    * switch to using pytest fixtures for test setup and cleanup steps to make
      them isolated and reusable
    * simulate a much more restricted network environment: instead of blocking
      single 88 port we now block all outgoing traffic except few essential
      ports
    * add basic tests for using kdcproxy to communicate between IPA client
      and AD DC.

### DIFF
--- a/ipatests/test_integration/test_http_kdc_proxy.py
+++ b/ipatests/test_integration/test_http_kdc_proxy.py
@@ -12,6 +12,7 @@ import pytest
 from ipatests.pytest_ipa.integration import tasks
 from ipatests.pytest_ipa.integration.firewall import Firewall
 from ipatests.test_integration.base import IntegrationTest
+from ipatests.util import xfail_context
 from ipaplatform.paths import paths
 
 
@@ -196,4 +197,9 @@ class TestHttpKdcProxy(IntegrationTest):
         tasks.clear_sssd_cache(self.master)
         user = users['ad']
         with self.configure_kdc_proxy_for_ad_trust(use_tcp=True):
-            tasks.kinit_as_user(self.client, user['name'], user['password'])
+            kdcproxy_version = tasks.get_python_package_version(
+                self.master, 'kdcproxy')
+            with xfail_context(kdcproxy_version < tasks.parse_version('0.4.2'),
+                               'https://github.com/latchset/kdcproxy/pull/44'):
+                tasks.kinit_as_user(
+                    self.client, user['name'], user['password'])
--- a/ipatests/test_integration/test_http_kdc_proxy.py
+++ b/ipatests/test_integration/test_http_kdc_proxy.py
@@ -4,53 +4,196 @@
 
 from __future__ import absolute_import
 
-import six
+import re
+from contextlib import contextmanager
+
+import pytest
+
 from ipatests.pytest_ipa.integration import tasks
 from ipatests.pytest_ipa.integration.firewall import Firewall
 from ipatests.test_integration.base import IntegrationTest
 from ipaplatform.paths import paths
 
 
-if six.PY3:
-    unicode = str
-
-
+@pytest.mark.usefixtures('ipa_ad_trust')
 class TestHttpKdcProxy(IntegrationTest):
     topology = "line"
     num_clients = 1
-    # Firewall rules without --append/-A, --delete/-D, .. First entry of
-    # each rule is the chain name, the argument to add or delete the rule
-    # will be added by the used Firewall method. See firewall.py for more
-    # information.
-    fw_rules = [['OUTPUT', '-p', 'tcp', '--dport', '88', '-j', 'DROP'],
-                ['OUTPUT', '-p', 'udp', '--dport', '88', '-j', 'DROP']]
+    num_ad_domains = 1
 
     @classmethod
     def install(cls, mh):
+        cls.client = cls.clients[0]
+        cls.ad = cls.ads[0]
         super(TestHttpKdcProxy, cls).install(mh)
-        # Block access from client to master's port 88
-        Firewall(cls.clients[0]).prepend_passthrough_rules(cls.fw_rules)
-        # configure client
-        cls.clients[0].run_command(
-            r"sed -i 's/ kdc = .*$/ kdc = https:\/\/%s\/KdcProxy/' %s" % (
-                cls.master.hostname, paths.KRB5_CONF)
-            )
-        cls.clients[0].run_command(
-            r"sed -i 's/master_kdc = .*$/master_kdc"
-            r" = https:\/\/%s\/KdcProxy/' %s" % (
-                cls.master.hostname, paths.KRB5_CONF)
-            )
-        # Workaround for https://fedorahosted.org/freeipa/ticket/6443
-        cls.clients[0].run_command(['systemctl', 'restart', 'sssd.service'])
-        # End of workaround
 
-    @classmethod
-    def uninstall(cls, mh):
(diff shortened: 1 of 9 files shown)


ipatests/pytest_ipa/integration/fixtures.py
```diff
@@ -0,0 +1,30 @@
+#
```
tiran, line 1: pytest fixtures are usually stored in a ``conftest.py``, which is auto-imported by pytest.

ipatests/test_integration/test_http_kdc_proxy.py
```diff
@@ -4,53 +4,202 @@
...
+        mh.master.run_command(['ipa', 'user-del', users['ipa']['name']])
+
+    @pytest.fixture()
+    def restrict_network_for_client(self, mh):
+        fw_rules_allow = [
+            ['OUTPUT', '-p', 'udp', '--dport', '53', '-j', 'ACCEPT'],
+            ['OUTPUT', '-p', 'tcp', '--dport', '80', '-j', 'ACCEPT'],
+            ['OUTPUT', '-p', 'tcp', '--dport', '443', '-j', 'ACCEPT'],
+            ['OUTPUT', '-p', 'tcp', '--sport', '22', '-j', 'ACCEPT']]
+        Firewall(self.client).prepend_passthrough_rules(fw_rules_allow)
```
tiran, line 60: Is it necessary to create a new firewall object for every function call?

ipatests/test_integration/test_http_kdc_proxy.py
```diff
@@ -4,53 +4,202 @@
...
+        result = tasks.kinit_as_user(
+            self.client, user['name'], user['password'], raiseonerr=False)
+        expected_error = (
+            "Cannot contact any KDC for realm '{}' while getting initial "
+            "credentials".format(user['domain'].realm))
+        assert result.returncode == 1 and expected_error in result.stderr_text
+
+    @pytest.mark.usefixtures('restrict_network_for_client',
+                             'client_use_kdcproxy')
+    def test_ipa_user_login_on_client_with_kdcproxy(self, users):
```
tiran, line 137: Please use follow black's and PEP-8's formatting style:
```

    @pytest.mark.usefixtures(
        'restrict_network_for_client',
        'client_use_kdcproxy'
    )
    def test_ipa_user_login_on_client_with_kdcproxy(self, users):
```

ipatests/pytest_ipa/integration/fixtures.py
wladich: fixed

ipatests/test_integration/test_http_kdc_proxy.py
wladich: definitely not, fixing.

ipatests/test_integration/test_http_kdc_proxy.py
wladich: Could you please point which pep-8 style rule you are referring to? 
And https://www.freeipa.org/page/Python_Coding_Style does not seem to mention any other rules...

tiran: It's black code style, https://github.com/psf/black#the-black-code-style

netoarmando:
@wladich, please, increase the timeout argument, I suspect that the reason jobs are picked by multiple runners is related to that.

netoarmando:
@wladich, please, increase the timeout argument, I suspect that the reason jobs are picked by multiple runners is related to that.

ipatests/test_integration/test_http_kdc_proxy.py
wladich: I tried to use black utility with new code and it heavily reformatted the code, not only the mentioned lines. One of the changes is making lines longer then 79 chars which violates current  code style.
But even with "-l 79" the changes are huge.
So could you please tell me, is the use of black formater now agreed and enforced for the FreeIPA project? I see it is not mentioned in https://www.freeipa.org/page/Python_Coding_Style, neither it is being checked by PR-CI.

wladich:
All tests [pass](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ca4f1e28-8e1b-11ea-877b-fa163e5dc42b) in rawhide and one test expectedly [fails](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/c113e56e-8e1b-11ea-adf7-fa163eea4ea4) in Fedora 31

ipatests/pytest_ipa/integration/fixtures.py
kaleemsiddiqu: This approach is applicable for fixtures which are applicable for all tests. Here this fixture is specific to a particular set of tests, so IMO its better to have this in that test class's install class method.

wladich: This is ipa_ad_trust fixture. It can be used to establish trust between IPA and AD when no special options are required. So this fixture could be used in this module, in test_smb, test_sssd and any new test suite which needs to setup trust.

kaleemsiddiqu: @wladich 
Still this fixture servers the trust scenario only.
I think we can move this fixture to a shared file(e.g pytest_ipa/integreation/tasks.py) and can import/use where ever necessary.

wladich: @kaleemsiddiqu Please see comment https://github.com/freeipa/freeipa/pull/4614#discussion_r417246592

Personally I do not have any preference where to store this small fixture - in a separate file or in conftest or somewhere else.

kaleemsiddiqu: I only want to say that scope of fixtures defined in conftest.py is very large and applies to almost every tests and thats when we should define it. IMO ipa_trust_ad does not fit that case. Its upto you and @tiran to decide, i just shared my opinion.

ipatests/conftest.py
```diff
@@ -169,3 +173,24 @@ def fin():
 
     request.addfinalizer(fin)
     return tempdir
+
+
+@pytest.fixture(scope='class')
+def ipa_ad_trust(mh):
```
tiran, line 179: This fixture only applies to integration tests and depends on the multihost ``mh`` fixture. Please move it to ``ipatests/pytest_ipa/integration/__init__.py``.

wladich:
Tests fail at ipa-server-install: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/d7be85da-9942-11ea-b3a3-fa163e168469

ipatests/conftest.py
wladich: @tiran 
fixed

wladich:
Rerunning the tests to check that they still pass after https://github.com/freeipa/freeipa-pr-ci/commit/7c6f67e3caa2088572accdd80c3a045e18f68505

PR #4615, opened 2020-04-29
tox.ini: switch from W503 to W504

fcami:
PEP8 recently changed from W503 to W504.
Line breaks should therefore come before operators.
See: https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator


tiran:
master:

* 1632827caf4fe34419ed0f14f35d5959e013aadd tox.ini: switch from W503 to W504

PR #4621, opened 2020-04-30
Black code style

tiran:


stanislavlevin:
@tiran, what about feature request to 'black' to run black over `git diff -U0`, like it is made in ```pycodestyle --diff```.

tiran:
> @tiran, what about feature request to 'black' to run black over `git diff -U0`, like it is made in `pycodestyle --diff`.

I'm pretty sure that this feature request would be rejected. It's also very hard to impossible to implement. Black operates on low-level grammar, parser, and tokenizer level.

And even if black would support this, I would still want to reformat all code at once. We have been struggling with code style issues for a long time. Let's have one commit to fix our code style once and for all.

doc/designs/black-codestyle.rst
```diff
@@ -0,0 +1,159 @@
...
+FreIPA.
+
+* Mass-changes of code is going to make backports harder because there will
+  be a lot of merge conflicts. To reduce merge conflicts all active branches
+  have to be auto-formatted with |black|.
+* Pull requests will also be affected by merge conflicts. Therefore we should
+  try to merge or close as many pull requests as possible before the code is
+  formatted.
+* Auto formatting is going to interfere with git history and make it harder
+  to find find issues. Git has an option to ignore revisions with
```
stanislavlevin, line 86: just typo `find find`.

Makefile.am
```diff
@@ -291,6 +291,18 @@ api: $(GENERATED_PYTHON_FILES)
...
+.PHONY: black
+black: $(GENERATED_PYTHON_FILES)
+	$(PYTHON) -m black \
+	    --config=$(top_srcdir)/pyproject.toml \
+	    $(top_srcdir)
+
+.PHONY: blacklint
+blacklint: $(GENERATED_PYTHON_FILES)
+	$(PYTHON) -m black \
+	    --config=$(top_srcdir)/pyproject.toml --fast --check \
```
stanislavlevin, line 303: no clues in the 'check' mode:
```
would reformat test.py
Oh no! 💥 💔 💥
1 file would be reformatted.
```
I guess not all the contributors run the linting locally. Would it be useful to see the diff ( via `--diff`)?
Of course, it's simpler to run `black` over the 'failed' files.

stanislavlevin:
@tiran, I see your point of view and agree with you.
But I'm afraid that not all the members like the idea to change the code without a really good reason.

doc/designs/index.rst
```diff
@@ -7,6 +7,7 @@ FreeIPA design documentation
    adtrust/oneway-trust-with-shared-secret.md
    adtrust/samba-domain-controller.md
    adtrust/samba-domain-member.md
+   black-codestyle
```
fcami, line 10: I think you want that as black-codestyle.md

fcami:
In theory, I'd like the whole project to be run through black. But I'm not sure how it would affect the maintenance part of our work. 
Can we limit that to new files at first?
- if a file is new, add it to a list of files to be run through black
- if a PR has a new file in, fail if that file is not present in the above database
- run black on each file mentioned in the database for all PRs

doc/designs/index.rst
tiran: No, it's a RST file and RST files don't need a file extension when including in a TOC tree.

fcami: Oy, ok

Makefile.am
tiran: Contributors should use ``make black`` or integrate black in their IDE instead. The ``blacklint`` target is for CI.

stanislavlevin: Yes, for CI. And I would like to see the exact reason of CI failure (in particular, the output of this tool to don't run manually on a local machine to get that info).

tiran: I could add ``--diff` to the argument list. However I don't think it is useful. The correct solution for black violation is always **run black again**.

tiran:
Preview PR https://github.com/tiran/freeipa/pull/new/black_preview

doc/designs/black-codestyle.rst
```diff
@@ -0,0 +1,254 @@
...
+   * Create ``pyproject.toml`` to configure |black| and include Python code
+     that does not have a ``.py`` file extension.
+   * Exclude auto-generated plugin code in ``ipaclient/remote_plugins/2_???``
+     from black. It's legacy code and no developer is going to touch the code
+     any more.
+
+2. Address remaining :ref:`pycodestyle issues <black-pycodestyle>` by either
+   fixing the issue or ignoring the warning locally or globally.
+   ``python3 -m pycodestyle .`` should pass without any error.
+3. Backport changes from (1) and (2) to ipa-4-6 and ipa-4-6 branches.
```
frasertweedale, line 240: Mentions `ipa-4-6` branch twice.

frasertweedale:
```
Showing 613 changed files with 79,999 additions and 71,319 deletions
```

Cool, cool.

I appreciate a lot of thought has gone into the design.

I appreciate that Git has a feature to ignore certain revisions in Git blame, but some empirical evidence that it will work well in this project---not only right after the big bang but after many more edits---would be helpful.  Otherwise we are at the mercy of that feature working correctly, and it would not be good to discover down the track that it does not work as well as we hoped.

Overall I feel that this is not needed.  We already have code style checking.  Comments about style in PRs usually occur because the linter picked something up; rarely because there is an actual dispute.

If we want to evolve the target style, in pursuit of smaller diffs or to eliminate quarrels or whatever reason, why can't we do that, instead of a big-bang rewrite?

Other than saying my piece here (ok and also over in https://pagure.io/freeipa/issue/8306), I'm not going to fight these changes.  Other day, another proof that the paradigm of "programming as manipulating strings" is rubbish.

/me goes back to manipulating strings in a text editor :)

abbra:
I don't like the fact that black style tends to waste vertical space, making very hard to work with long logical blocks. We have plenty of those and splitting into smaller pieces is not really an option in all of those. To me this is one of major issues as it forces to break out of code analysis more often. The waste of my brain energy becomes noticeable in such case, unfortunately.
In the end, it is all about emotions and feelings. If I have to spend more time to get my task done due to the tools, I'd tend to change tools.

rcritten:
I don't understand the genesis for this change. Was there some big argument recently about code style? Is this related to the recent pylint changes that caused issues?

Why was this major change introduced as a PR and then a ticket created? Should a change of this magnitude require a conversation in advance? Or was the feeling that showing the changes would provide a benefit?

Are we going to assume that because CI passes that the changes are all fine? Who can review a hundred thousand+ changes?

fcami:
@rcritten replying to "Are we going to assume that because CI passes that the changes are all fine?" only, quoting from Black docs: "Black will check that the reformatted code still produces a valid AST that is equivalent to the original." (unless you use --fast, which we won't). It all comes down to trusting this.

doc/designs/black-codestyle.rst
```diff
@@ -0,0 +1,308 @@
...
+RHEL 7.8 was released March 31, 2020. RHEL 7.9 will be the last RHEL 7
+y-stream release. RHEL 7 has reached end of full support in August 2019 and
+is going to reach end of *Maintenance Support 1* in August 2020. During the
+*Maintenance Support 2* phase critical and import security fixes and select
+urgent bug fixes are provided until June 2024.
+
+I expect that the amount of backports to 4.6 will soon diminish to a few
+commits a year.
+
+Q: When can we start to limit backports to urgent CVEs only?
```
fcami, line 148: According to the "RHEL Life Cycle" document above, "End of Maintenance Support or Maintenance Support 2" for RHEL7 happens on June 30, 2024. Maintenance Support 2 includes "Red Hat-defined Critical and Important impact Security Advisories (RHSAs) and selected (at Red Hat discretion) Urgent Priority Bug Fix Advisories (RHBAs)".

So there is no date that limits the changes to RHEL7 to urgent CVEs during the 10-year RHEL7 life cycle.

doc/designs/black-codestyle.rst
```diff
@@ -0,0 +1,308 @@
...
+New code style
+==============
+
+The new code style is *whatever black does*.
+
+The rules for i18n strings and unused variables from the |FreeIPA|
+`Python code style`_ still apply, though.
+
+|Black| increases the permitted maximum line length from 79/80 characters to
+88 characters. According to `black project`_'s documentation
```
fcami, line 170: In short: whether we approve this change or not, I would appreciate moving to 88 characters (or more (*)).

(*) Using longer lines would make code accessibility lower: longer lines are only readable on high-resolution screens and/or when using small characters. So a middle-ground has to be found, and 88c is at least better than 80c.

doc/designs/black-codestyle.rst
```diff
@@ -0,0 +1,308 @@
...
+-----------
+
+*E203* whitespace before ':'
+   *E203* is not :pep:`8` conform. |Black| treats slice ``:`` as binary
+   operator and enforces whitespace in slices, for example ``ham[1 + 1 :]``.
+*E231* missing whitespace after ','
+   |Black| always adds a comma after all arguments, e.g. ``func(a,)``.
+*W503* line break before binary operator
+   *W503* is not :pep:`8` conform.
+*W504* line break after binary operator
```
fcami, line 266: This is unfortunate. But OK, if it happens rarely.

fcami:
Also: a gating run will not be enough for this. For now, while we discuss design, please use a temp commit within nothing defined, but if we agree on merging it, please change the PR-CI symlink to ipatests/prci_definitions/nightly_latest.yaml

doc/designs/black-codestyle.rst
```diff
@@ -0,0 +1,308 @@
...
+
+2. Address remaining :ref:`pycodestyle issues <black-pycodestyle>` by either
+   fixing the issue or ignoring the warning locally or globally.
+   ``python3 -m pycodestyle .`` should pass without any error.
+3. Backport changes from (1) and (2) to ipa-4-8 branch.
+4. Run ``make black`` in master + ipa-4-8 branch and merge the changes.
+5. Create ``.gitignorerevs`` file with commit hashes of |Black| run from all
+   active branches``.
+6. Enable ``blacklint`` for ``fastlint`` and ``lint`` targets so
+   local linting and linting on Azure check for black violations.
```
rcritten, line 297: rather than running blacklint in CI, if we know all the code is black-compliant, then add a git commit check to require black formatting so a PR can't be submitted at all w/o complying with formatting. This way users wouldn't have to wait for CI to find out they have a formatting issue. I don't always run fastlint myself before submitting a PR.

PR #4625, opened 2020-04-30
Updated ipa-pki-proxy.conf.template

edewata:
pkispawn is being modified to use PKI CLI for installation.
The `ipa-pki-proxy.conf.template` has been modified to allow
PKI CLI to run properly.

It would be nicer if we can use wildcards (e.g. `/pki/*`, `/ca/*`,
`/kra/*`) instead so we don't have to update this for every changes
in PKI.


tiran:
``LocationMatch`` is a prefix match, see https://github.com/freeipa/freeipa/pull/4641

edewata:
Yes, that'll be great. Thanks! Closing this one.

PR #4634, opened 2020-05-03
Use httpd 2.4 syntax for access control

tiran:
The httpd options Allow, Deny, Order, and Satisfy are deprecated in
Apache httpd 2.4. These options are provided by the mod_access_compat
module and should no longer be used.

Replace "Allow from all" with "Require all granted".

Removal of "Satisfy Any" needs more investigation.

See: https://httpd.apache.org/docs/2.4/upgrading.html
See: https://httpd.apache.org/docs/2.4/mod/mod_access_compat.html
Fixes: https://pagure.io/freeipa/issue/8305

### DIFF
--- a/install/share/ipa.conf.template
+++ b/install/share/ipa.conf.template
@@ -1,5 +1,5 @@
 #
-# VERSION 30 - DO NOT REMOVE THIS LINE
+# VERSION 31 - DO NOT REMOVE THIS LINE
 #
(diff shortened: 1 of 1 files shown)


rcritten:
Looks good, ack.

rcritten:
master:

* 2bfe5ff689163ab1b9763c86c2b0355a07b2c33c Use httpd 2.4 syntax for access control

PR #4641, opened 2020-05-04
Simplify pki proxy conf

tiran:
``pkispawn`` is being modified to use PKI CLI for installation.

Add ``/pki/rest`` to proxied routes and simplify location matching with
a prefix regular expression.

### DIFF
--- a/install/share/ipa-pki-proxy.conf.template
+++ b/install/share/ipa-pki-proxy.conf.template
@@ -1,4 +1,4 @@
-# VERSION 14 - DO NOT REMOVE THIS LINE
+# VERSION 15 - DO NOT REMOVE THIS LINE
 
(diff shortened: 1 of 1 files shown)


tiran:
@edewata This changeset may also expose additional routes to users. Are there any REST endpoints that must be protected or filtered out?

edewata:
The PKI REST endpoints that need to be protected are already protected by PKI. I don't think we need to add more protection through ipa-pki-proxy.conf.

abbra:
LGTM

tiran:
master:

* 19ea1b97a1559609014e0784c6f973c0e073878a Simplify pki proxy conf

edewata:
Thanks!

PR #4643, opened 2020-05-04
Make ipaplatform a regular top-level package

tiran:
ipaplatform was made a namespace package so that 3rd party OS
distributors can easily define their own distribution subpackage. Since
major distributions have contributed to FreeIPA project and no 3rd party
ipaplatform subpackage was uploaded to PyPI, it doesn't make much sense
to keep ipaplatform a namespace package.

The ipaplatform-*-nspkg.pth file for namespace package definition is
causing trouble with local testing on developer boxes.


abbra:
LGTM. This fixed long standing issue I had with `PYTHONPATH` failing to prefer in-tree modules.

tiran:
master:

* 490682ac3cab8978635655bdcc880f28227e900b Make ipaplatform a regular top-level package

PR #4647, opened 2020-05-05
ipa-advise: fallback to /usr/libexec/platform-python if python3 not f…

flo-renaud:
…ound

when ipa-advise generates a script to configure a client for smart card
auth, the script calls python3 to configure SSSD. The issue happens
if the server (when ipa-advise is run) and the client do not have
the same path for python3 command.

By default, try to use python3 but if the command is not found, fallback
to /usr/libexec/platform-python (which is the python3 path on RHEL8).

Fixes: https://pagure.io/freeipa/issue/8311


tiran:
master:

* edcfba6010c0b6a81841ca6e1e478835ce76191b ipa-advise: fallback to /usr/libexec/platform-python if python3 not found

PR #4656, opened 2020-05-05
Fix detection logic for api.env.in_tree

tiran:
The logic to detect in-tree builds was broken and ipatests/conftest.py
had hard-coded in_tree=True.

IPA now considers an environment as in-tree when the parent directory of
the ``ipalib`` package contains ``ipasetup.py.in``. This file is only
present in source and never installed.

API bootstrap() does not use ```self.site_packages in site.getsitepackages()``
because the function call can be expensive and would require path
normalization, too.

Fixes: https://pagure.io/freeipa/issue/8312

### DIFF
--- a/ipalib/config.py
+++ b/ipalib/config.py
@@ -457,11 +457,11 @@ class Env:
         # Merge in overrides:
         self._merge(**overrides)
 
-        # Determine if running in source tree:
+        # Determine if running in source tree. The root directory of
+        # IPA source directory contains ipasetup.py.in.
         if 'in_tree' not in self:
-            self.in_tree = (
-                self.bin == self.site_packages
-                and path.isfile(path.join(self.bin, 'setup.py'))
+            self.in_tree = os.path.isfile(
+                os.path.join(self.site_packages, "ipasetup.py.in")
             )
         if self.in_tree and 'mode' not in self:
             self.mode = 'developer'
--- a/ipatests/conftest.py
+++ b/ipatests/conftest.py
@@ -118,7 +118,7 @@ def pytest_addoption(parser):
 
 def pytest_cmdline_main(config):
     kwargs = dict(
-        context=u'cli', in_server=False, in_tree=True, fallback=False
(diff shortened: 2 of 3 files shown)


Carbenium:
Test failures are the same I observed and mentioned in https://github.com/freeipa/freeipa/pull/4589#issuecomment-623528468

tiran:
DNS tests keep failing with ``ipalib.errors.OptionError: Unknown option: force`` in ``permission_del``.

Carbenium:
@tiran I am pretty sure the failure you mentioned are related to the fact that the ```force``` option on ``permission_del`` has the ``no_option`` flag set: https://github.com/freeipa/freeipa/blob/340a50b7e78a4b63e9d2abd86766c3b6c4f47099/ipaserver/plugins/permission.py#L1060

Combine this with the different plugin import logic depending on ``in_tree`` https://github.com/freeipa/freeipa/blob/340a50b7e78a4b63e9d2abd86766c3b6c4f47099/ipaclient/remote_plugins/__init__.py#L120
we end up in the position that all options which are flagged with ``no_option`` are unavailable in remote plugins.
At least that's what I came up with while stepping through https://github.com/freeipa/freeipa/blob/340a50b7e78a4b63e9d2abd86766c3b6c4f47099/ipalib/frontend.py#L558

abbra:
There are many other failures in Azure regarding sudo operations. They seem to be genuine and may be related to some resource racing. I suggest to rebase this on top of F32 switch to see if we get rid of them that way.

tiran:
Blocked by https://pagure.io/freeipa/issue/8317

tiran:
I have temporary re-add ``in_true=True`` to make the tests pass. Let's address the issue with https://pagure.io/freeipa/issue/8317

tiran:
master:

* 13c3997baa97b7974870cb2353caa4ced8f134d1 Fix detection logic for api.env.in_tree
* 0fa31ef123a544378f04a2177524c9f72181a812 Hard-code in_tree=True for tests

PR #4666, opened 2020-05-06
Fix kdc memory handling

abbra:
1 .kdb: fix memory handling in ipadb_find_principal
    
BER structure representing a string might not have termination '\0' character, thus we should use length-bound functions to operate on it.
    
Memory handling of LDAP values was leaving previous vals over iteration. Also, when freeing vals, we need to explicitly set it to NULL.
   
2. kdb: initialize flags in ipadb_delete_principal()

Minor fix after kdb module refactoring some time ago.

3. test_smb: test that we can auth as NetBIOS alias
    
cifs/... principal on SMB server side has NetBIOS name of the SMB server as its alias. Test that we can actually initialize credentials using this alias. We don't need to use it anywhere in Samba, just verify that alias works. This is a test for the main fix.

Fixes: https://pagure.io/freeipa/issue/8291

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb_principals.c
+++ b/daemons/ipa-kdb/ipa_kdb_principals.c
@@ -1039,6 +1039,7 @@ krb5_error_code ipadb_find_principal(krb5_context kcontext,
     struct berval **vals = NULL;
     int result;
     krb5_error_code ret;
+    size_t princ_len = 0;
 
     ipactx = ipadb_get_context(kcontext);
     if (!ipactx) {
@@ -1046,6 +1047,7 @@ krb5_error_code ipadb_find_principal(krb5_context kcontext,
         goto done;
     }
 
+    princ_len = strlen(*principal);
     for (le = ldap_first_entry(ipactx->lcontext, res); le != NULL;
          le = ldap_next_entry(ipactx->lcontext, le)) {
         vals = ldap_get_values_len(ipactx->lcontext, le, "krbprincipalname");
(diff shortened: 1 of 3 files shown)


abbra:
The fix and test were successfully verified by a separate PR already but since that PR contains many unrelated changes, I'm submitting it separately. Currently test_smb is flacky, so it was removed from gating.
The fix works according to the original reporter.

daemons/ipa-kdb/ipa_kdb_principals.c
```diff
@@ -1057,7 +1059,8 @@ krb5_error_code ipadb_find_principal(krb5_context kcontext,
         for (int i = 0; vals[i]; i++) {
 #ifdef KRB5_KDB_FLAG_ALIAS_OK
             if ((flags & KRB5_KDB_FLAG_ALIAS_OK) == 0) {
-                found = strcmp(vals[i]->bv_val, *principal) == 0;
+                found = ((vals[i]->bv_len == princ_len) &&
+                         strncmp(vals[i]->bv_val, *principal, vals[i]->bv_len) == 0);
```
tiran, line 1063: Is ``strncmp`` safe? How about a custom function that first compares length of BerValue with strlen(char*), then performs memcmp() iff both are the same size.

abbra: it is a string but not '\0'-terminated. The principal cannot contain any NUL byte.

abbra: `strncmp` is fine, this is what krb5 itself uses.

tiran: Thanks!

abbra:
ipa-4-6 backport will be partial as we do not have test_smb in 4.6 branch

abbra:
master:

* 1b9233615e611f8ec52986a57ca50dfceb558c1a kdb: initialize flags in ipadb_delete_principal()
* 999af8e2ef5330bfb488a20e2b7c669064b73d69 kdb: fix memory handling in ipadb_find_principal
* 6fc213d10deacf107f5ca225a0095bdf215dc73b test_smb: test that we can auth as NetBIOS alias

PR #4674, opened 2020-05-08
Disable password schema update on LDAP bind

tiran:
389-DS 1.4.1+ attempts to update passwords to new schema on LDAP bind. IPA
blocks hashed password updates and requires password changes to go through
proper APIs. This option disables password hashing schema updates on bind.

See: https://pagure.io/freeipa/issue/8315
See: https://bugzilla.redhat.com/show_bug.cgi?id=1833266
See: https://pagure.io/389-ds-base/issue/49421

### DIFF
--- a/install/updates/10-config.update
+++ b/install/updates/10-config.update
@@ -66,3 +66,10 @@ only:nsslapd-allow-hashed-passwords:on
 # Decrease default value for IO blocking to prevent server unresponsiveness
 dn: cn=config
 only:nsslapd-ioblocktimeout:10000
+
+# 389-DS 1.4.1.6+ attempts to update passwords to new schema on LDAP bind.
+# IPa blocks hashed password updates and requires password changes to go
+# through proper APIs. This option disables password hashing schema updates
+# on LDAP bind, see https://pagure.io/freeipa/issue/8315
+dn: cn=config
+only: nsslapd-enable-upgrade-hash:off
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -824,6 +824,17 @@ class TestInstallMaster(IntegrationTest):
             msg = "rpm -V found group issues for the following files: {}"
             assert group_warnings == [], msg.format(group_warnings)
 
+    def test_ds_disable_upgrade_hash(self):
+        # Test case for https://pagure.io/freeipa/issue/8315
+        # Disable password schema migration on LDAP bind
+        result = tasks.ldapsearch_dm(
+            self.master,
+            "cn=config",
+            ldap_args=["nsslapd-enable-upgrade-hash"],
+            scope="base"
+        )
+        assert "nsslapd-enable-upgrade-hash: off" in result.stdout_text
+
 
 class TestInstallMasterKRA(IntegrationTest):
 
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -752,6 +752,7 @@ class LDAPClient:
         'nsslapd-idlistscanlimit': True,
         'nsslapd-anonlimitsdn': True,
         'nsslapd-minssf-exclude-rootdse': True,
+        'nsslapd-enable-upgrade-hash': True,
     })
 
     time_limit = -1.0   # unlimited


install/updates/10-config.update
```diff
@@ -66,3 +66,11 @@ only:nsslapd-allow-hashed-passwords:on
 # Decrease default value for IO blocking to prevent server unresponsiveness
 dn: cn=config
 only:nsslapd-ioblocktimeout:10000
+
+# 389-DS 1.4.1+ attempts to update passwords to new schema on LDAP bind. IPA
```
abbra, line 70: the exact version is 1.4.1.6+

ipatests/test_integration/test_installation.py
```diff
@@ -824,6 +824,15 @@ def test_file_permissions(self):
...
             assert group_warnings == [], msg.format(group_warnings)
 
+    def test_ds_disable_upgrade_hash(self):
+        result = tasks.ldapsearch_dm(
+            self.master,
+            "cn=config",
+            ldap_args=["nsslapd-enable-upgrade-hash"],
+            scope="base"
+        )
+        assert "nsslapd-enable-upgrade-hash:off" in result.stdout_text
```
abbra, line 834: the output would be with a space between the `argument: value` pair:

```
# ldapsearch -Y EXTERNAL -H `cat /etc/ipa/default.conf |grep ldap_uri|cut -d= -f2` -b cn=config -s base nsslapd-enable-upgrade-hash
SASL/EXTERNAL authentication started
SASL username: gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth
SASL SSF: 0
# extended LDIF
#
# LDAPv3
# base <cn=config> with scope baseObject
# filter: (objectclass=*)
# requesting: nsslapd-enable-upgrade-hash 
#

# config
dn: cn=config
nsslapd-enable-upgrade-hash: on

# search result
search: 2
result: 0 Success

# numResponses: 2
# numEntries: 1
```

tiran:
Thanks for the feedback.

ipatests/test_integration/test_installation.py
```diff
@@ -824,6 +824,15 @@ def test_file_permissions(self):
             msg = "rpm -V found group issues for the following files: {}"
             assert group_warnings == [], msg.format(group_warnings)
 
+    def test_ds_disable_upgrade_hash(self):
```
mrizwan93, line 827: It would be nice to have some lines describing test and pagure issue in doc string.

abbra:
I think internal updater stumbles upon the issue in `LDAPEntry.generate_modlist()` where if an attribute existed in the entry before (our case) it gets replaced only if we sense a single value attribute through the schema:
```
            # We used to convert to sets and use difference to calculate
            # the changes but this did not preserve order which is important
            # particularly for schema
            adds = [value for value in new if value not in old]
            dels = [value for value in old if value not in new]
            if adds and self.conn.get_attribute_single_value(name):
                if len(adds) > 1:
                    raise errors.OnlyOneValueAllowed(attr=name)
                modlist.append((ldap.MOD_REPLACE, name, adds))
            else:
                # dels before adds, in case the same value occurs in
                # both due to encoding differences
                # (https://pagure.io/freeipa/issue/7750)
                if dels:
                    modlist.append((ldap.MOD_DELETE, name, dels))
                if adds:
                    modlist.append((ldap.MOD_ADD, name, adds))

```

I verified that `nsslapd-enable-upgrade-hash` is not returned by the LDAP schema.

abbra:
```
# ldapsearch -Q -Y EXTERNAL -H `cat /etc/ipa/default.conf |grep ldap_uri|cut -d= -f2` -o ldif-wrap=1000 -b cn=schema -s base attributetypes | grep -q nsslapd-enable-upgrade-hash || echo 'Not found'
Not found

```

tiran:
After further investigation Mark and Alexander figured out that the issue is related to LDAP schema. 389-ds doesn't list the attribute in its schema and therefore our ipaldap API treats the attribute as multi-valued attribute. I have added the attribute to ``_SINGLE_VALUE_OVERRIDE`` to work around the problem.

abbra:
LGTM.

tiran:
master:

* aa341020c883b3d8e6117f6f515f5ed0e19c04e7 Disable password schema update on LDAP bind

PR #4677, opened 2020-05-11
Support server referrals

abbra:
Implement minimal server referrals support for enterprise principals as defined in RFC 6806.

Use krb5_pac_verify_ext() and krb5_pac_sign_ext() to support cross-realm S4U extensions. We have to verify/sign PAC and take the realm into account for S4U in these cases.

The use of extended functions require krb5 1.17+.

For PAC verification, we have to filter existing PAC CLIENT-INFO structure in cross-realm S4U case because otherwise old CLIENT-INFO would change the PAC principal due to adding or ommiting the realm in transition.  Since a new PAC CLIENT-INFO will be provided by
k5_insert_client_info() anyway, we can filter it in all cases.

Generate PAC only for the first S4U2Self request to the client realm (client != NULL). Otherwise, use the PAC from the cross-realm ticket. The latter PAC belongs to the impersonated user.

Finally, in PAC signing we have to take the realm into account as well for S4U2Self cross-realm operation. This does not work when compiling against krb5 1.17 at the moment because sign_authdata() callback does not know whether we are dealing with an issuing referral or not. In 1.18 there is a special flag set for the client that signifies this.

Fixes: https://pagure.io/freeipa/issue/8319

### DIFF
--- a/daemons/ipa-kdb/ipa-print-pac.c
+++ b/daemons/ipa-kdb/ipa-print-pac.c
@@ -1,0 +1,723 @@
+/*
+ * Copyright (C) 2020  FreeIPA Contributors see COPYING for license
+ */
+#include <gen_ndr/ndr_krb5pac.h>
+#include <gssapi/gssapi_ext.h>
+#include <gssapi/gssapi_krb5.h>
+#include <ndr.h>
+#include <popt.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h>
+#include <strings.h>
+
+#define IPAPWD_PASSWORD_MAX_LEN 1024
+
+typedef enum {
+    OP_SERVICE_TICKET,
+    OP_IMPERSONATE
+} pac_operation_t;
+
+pac_operation_t operation = OP_SERVICE_TICKET;
+char *keytab_path         = NULL;
+char *ccache_path         = NULL;
+bool init_tgt             = true;
+const gss_OID *import_name_oid   = &GSS_C_NT_USER_NAME;
+
+TALLOC_CTX *frame = NULL;
+
+gss_OID_desc mech_krb5 = {9, "\052\206\110\206\367\022\001\002\002"};
+
+/* NDR printing interface passes flags but the actual public print function
+ * does not accept flags. Generated ndr helpers actually have a small wrapper
+ * but since it is a static to the generated C code unit, we have to reimplement
+ * it here.
+ */
+static void
+print_flags_PAC_DATA(struct ndr_print *ndr,
+                     const char *name,
+                     int unused,
+                     const struct PAC_DATA *r)
+{
+    ndr_print_PAC_DATA(ndr, name, r);
+}
+
+/*
+ * Print content of a PAC buffer, annotated by the libndr helpers
+ */
+static void
+print_pac(gss_buffer_desc *pac, gss_buffer_desc *display)
+{
+    struct ndr_print *ndr = NULL;
+    DATA_BLOB blob;
+    struct ndr_pull *ndr_pull = NULL;
+    void *st                  = NULL;
+    int flags                 = NDR_SCALARS | NDR_BUFFERS;
+    enum ndr_err_code ndr_err;
+    struct ndr_interface_call ndr_call = {
+        .name        = "PAC_DATA",
+        .struct_size = sizeof(struct PAC_DATA),
+        .ndr_push    = (ndr_push_flags_fn_t)ndr_push_PAC_DATA,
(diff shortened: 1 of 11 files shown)


abbra:
Current state: tests in both `test_smb` and `test_commands` demonstrate that changes do work. Unfortunately, we are putting too much pressure on parallel execution in Azure so there is an issue with 389-ds in containers not recognizing we have enough swap available and denying memory allocations. This has been happening before but because the last test in `test_commands`, `test_integration/test_commands.py::TestIPACommand::test_ipa_adtrust_install_with_locale_issue8066`, is using `raiseonerr=False`, the failure is not reported and the whole test suite is considered successful. When I added PAC tests after it (we need to have system configured with ipa-adtrust-install before that), they exposed the issue which existed previously.

Failures in `test_smb` are also unrelated to my changes since they happened before. There seems to be some general issue with `cifs.ko` operations that need to be investigated separately.

At this point I am going to remove `test_commands` from Azure and only run them in PR CI.

For the reference, successful runs of PAC tests in PR CI:
- `test_commands`: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/02b0ed6a-9ad2-11ea-b4d0-fa163e2dc507/report.html, tests `test_integration/test_commands.py::TestIPACommand::test_ipa_user_pac` and `test_integration/test_commands.py::TestIPACommand::test_ipa_user_s4u2self_pac`
- `test_smb`: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/c71e5ef0-9acb-11ea-aaa6-fa163e0a82d3/report.html, test `test_integration/test_smb.py::TestSMB::test_smb_service_s4u2self`

I also need to update help message in `ipa-print-pac` to make it corresponding to code logic. I'm going to remove temp commit with that.

abbra:
@flo-renaud `ipa-print-pac` tests are run as part of `test_adtrust_install` test suite. For example, http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/bdb87cac-9f7a-11ea-baec-fa163e968362/report.html demonstrates two modes of `ipa-print-pac` (last two tests), they have PAC data published. The difference can be seen by the SIDs included in sids array:
- normal ticket acquisition operation should have
```
  sid                      : S-1-18-1
```
- impersonation with S4U2Self request should have
```
sid                      : S-1-18-2
```

daemons/ipa-kdb/ipa-print-pac.c
```diff
@@ -0,0 +1,723 @@
...
+                       "user (S4U2Self).\n\n"
+                       "Resulting service ticket can be stored in the credential cache file "
+                       "specified by '-c file' option.\n\n";
+    fprintf(stderr, help, name);
+    poptPrintHelp(pc, stderr, 0);
+}
+
+static char *
+ask_password(TALLOC_CTX *context, char *prompt1, char *prompt2, bool match)
+{
```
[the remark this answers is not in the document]
abbra: I intentionally left it this way because this tool will eventually other operations and I wanted to keep the method the same as in `ipa-getkeytab`.

abbra:
@flo-renaud thank you for the review -- I fixed comments and help strings.

flo-renaud:
@abbra 
my bad, I probably forgot the `--add-sids` option during `ipa-adtrust-install`. Now able to see the PAC data, and ipa-print-pac behaves as described.

daemons/ipa-kdb/ipa_kdb_principals.c
```diff
@@ -1295,163 +1298,203 @@ krb5_error_code ipadb_get_principal(krb5_context kcontext,
...
+
+    *entry = NULL;
+
+    ipactx = ipadb_get_context(kcontext);
+    if (!ipactx) {
+        return KRB5_KDB_DBNOTINITED;
+    }
+
+    if (!is_request_for_us(kcontext, ipactx->local_tgs, search_for)) {
+        return KRB5_KDB_NOENTRY;
```
iboukris, line 1486: If we have any log facilities, it would be nice to log this.

abbra: Added an info log entry.

flo-renaud:
All the comments were addressed, tests are green, ACK

abbra:
Thank you. I pushed but somehow ipatool timed out and didn't complete the required changes to the tickets and PR. I'll do those (and backport) manually.

abbra:
The fixes are in pagure git master but not synced to github yet.

PR #4685, opened 2020-05-13
WebUI: Add confirmation dialog for changing default user/host group

serg-cymbaluk:
Changing default group on automember rules page is too easy.
Add a confirmation dialog to avoid misclick in the case.

Ticket: https://pagure.io/freeipa/issue/8322

### DIFF
--- a/install/ui/src/freeipa/automember.js
+++ b/install/ui/src/freeipa/automember.js
@@ -601,11 +601,22 @@ IPA.automember.default_group_widget = function(spec) {
 
         if (group === that.group) return;
 
-        if (group === '') {
-            that.remove_default_group();
-        } else {
-            that.set_default_group(group);
-        }
+        var dialog = IPA.confirm_dialog({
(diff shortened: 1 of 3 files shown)


flo-renaud:
Hi @serg-cymbaluk 
thanks for the PR, works well for me.
I tested with Automember > User group rules and Automember > Host group rules, both are OK for changing the default group or cancelling the change.

flo-renaud:
Please remove the temp commit to allow pushing.

serg-cymbaluk:
@flo-renaud 
Thanks for your review.
Temp commit is removed. Tests passed.

serg-cymbaluk:
master:

* 33ca0745585ae3fcd466631a069e244bea74be88 WebUI: Add confirmation dialog for changing default user/host group
* 3645854c11b0489215a27077a633755638f00268 WebUI tests: Add confirmation step after changing default group in automember tests

PR #4686, opened 2020-05-13
service delegation: allow to add and remove host principals

abbra:
Service delegation rules and targets deal with Kerberos principals.
As FreeIPA has separate service objects for hosts and Kerberos services,
it is not possible to specify host principal in the service delegation
rule or a target because the code assumes it always operates on Kerberos
service objects.

Simplify the code to add and remove members from delegation rules and
targets. New code looks up a name of the principal in cn=accounts,$BASEDN
as a krbPrincipalName attribute of an object with krbPrincipalAux object
class. This search path is optimized already for Kerberos KDC driver.

To support host principals, the specified principal name is checked to
have only one component (a host name). Service principals have more than
one component, typically service name and a host name, separated by '/'
sign. If the principal name has only one component, the name is
prepended with 'host/' to be able to find a host principal.

The logic described above allows to capture also aliases of both
Kerberos service and host principals. Additional check was added to
allow specifying single-component aliases ending with '$' sign. These
are typically used for Active Directory-related services like databases
or file services.

RN: service delegation rules and targets now allow to specify hosts as
RN: a rule or a target's member principal.

Fixes: https://pagure.io/freeipa/issue/8289

### DIFF
--- a/ipaserver/plugins/servicedelegation.py
+++ b/ipaserver/plugins/servicedelegation.py
@@ -15,10 +15,10 @@ from .baseldap import (
     LDAPDelete,
     LDAPSearch,
     LDAPRetrieve)
-from .service import normalize_principal
 from ipalib import _, ngettext
 from ipalib import errors
 from ipapython.dn import DN
+from ipapython import kerberos
 
 if six.PY3:
     unicode = str
@@ -46,6 +46,13 @@ A target consists of a list of principals that can be delegated.
 In English, a rule says that this principal can delegate as this
 list of principals, as defined by these targets.
 
+In both a rule and a target Kerberos principals may be specified
+by their name or an alias and the realm can be omitted. Additionally,
+hosts can be specified by their names. If Kerberos principal specified
+has a single component and does not end with '$' sign, it will be treated
+as a host name. Kerberos principal names ending with '$' are typically
+used as aliases for Active Directory-related services.
+
 EXAMPLES:
 
  Add a new constrained delegation rule:
@@ -56,6 +63,10 @@ EXAMPLES:
 
  Add a principal to the rule:
    ipa servicedelegationrule-add-member --principals=ftp/ipa.example.com \
+      ftp-delegation
+
+ Add a host principal of the host 'ipa.example.com' to the rule:
+   ipa servicedelegationrule-add-member --principals=ipa.example.com \
       ftp-delegation
 
  Add our target to the rule:
@@ -169,6 +180,21 @@ class servicedelegation(LDAPObject):
     )
 
 
+def normalize_principal_name(name, realm):
+    try:
+        princ = kerberos.Principal(name, realm=realm)
+    except ValueError as e:
+        raise errors.ValidationError(
+            name='principal',
+            reason=_("Malformed principal: %(error)s") % dict(error=str(e)))
+
+    if len(princ.components) == 1 and not princ.components[0].endswith('$'):
+        nprinc = 'host/' + unicode(princ)
+    else:
+        nprinc = unicode(princ)
+    return nprinc
+
+
 class servicedelegation_add_member(LDAPAddMember):
     __doc__ = _('Add target to a named service delegation.')
     member_attrs = ['memberprincipal']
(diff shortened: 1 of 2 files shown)


ipaserver/plugins/servicedelegation.py
```diff
@@ -212,22 +223,37 @@ def post_callback(self, ldap, completed, failed, dn, entry_attrs,
...
                 if not name:
                     continue
-                name = normalize_principal(name)
-                obj_dn = ldap_obj.get_dn(name)
+
+                try:
+                    n = kerberos.Principal(name, realm=self.api.env.realm)
+                except ValueError:
+                    raise errors.ValidationError(
+                        name='principal', reason=_("Malformed principal"))
```
rcritten, line 236: I think you should catch the error and include it in the reason. The error would be something like "Principal is not in <name>@<realm> format"

ipaserver/plugins/servicedelegation.py
```diff
@@ -212,22 +223,37 @@ def post_callback(self, ldap, completed, failed, dn, entry_attrs,
...
+
+                if len(n.components) == 1 and (n.components[0][-1] != '$'):
+                    nprincipal = 'host/' + unicode(n)
+                else:
+                    nprincipal = unicode(n)
```
rcritten, line 241: I'm not sure the unicode bit is necessary, this is going to be python3-only right? Or are you being considerate of backports?

ipaserver/plugins/servicedelegation.py
abbra: We do accept non-qualified principals (`hostname` or `ldap/hostname` are just fine), so that wouldn't be correct way to report an error. The error here would be from `kerberos.parse_princ_name_and_realm()` which throws `ValueError` if more than one unescaped `@` is in the original principal, e.g. we were passed a malformed enterprise principal:
```
>>> kerberos.Principal('foo@bar@realm')
  File "<console>", line 1, in <module>
  File "/usr/lib/python3.8/site-packages/ipapython/kerberos.py", line 76, in __init__
    self.components, self.realm = self._parse_from_text(
  File "/usr/lib/python3.8/site-packages/ipapython/kerberos.py", line 125, in _parse_from_text
    principal_name, parsed_realm = parse_princ_name_and_realm(
  File "/usr/lib/python3.8/site-packages/ipapython/kerberos.py", line 32, in parse_princ_name_and_realm
    raise ValueError(
ValueError: Principal is not in <name>@<realm> format
>>> kerberos.Principal('foo\@bar@realm')
ipapython.kerberos.Principal('foo\@bar@realm')
```

I think in this case it is OK to throw an error here and not proceed further because chances are the whole request is invalid.

ipaserver/plugins/servicedelegation.py
abbra: Yes, I want to not preclude this possibility. Additionally, the rest of the module is py2-friendly and I am keeping the status quo here. When we'd decide to clean everything of the py2 compatibility, it will need to be done for the whole file anyway.

ipaserver/plugins/servicedelegation.py
rcritten: Exactly, I'm not suggesting you ignore an error, I"m asking to report the details of it. Something like:

```
except ValueError as e:
    raise errors.ValidationError(
        name='principal', reason=_("Malformed principal: %(error)s") % dict(error=str(e))
```
This would be a value-add while the code around it is changing.

ipaserver/plugins/servicedelegation.py
rcritten: +1

ipaserver/plugins/servicedelegation.py
abbra: Fixed.

ipaserver/plugins/servicedelegation.py
```diff
@@ -212,22 +223,39 @@ def post_callback(self, ldap, completed, failed, dn, entry_attrs,
...
                 except errors.NotFound as e:
                     failed[self.principal_failedattr][
                         self.principal_attr].append((name, unicode(e)))
                     continue
                 try:
-                    if name not in entry_attrs.get(self.principal_attr, []):
-                        members.append(name)
+                    mprinc = [p for p in e_attrs.get('krbprincipalname')
+                              if nprincipal.lower() == unicode(p).lower()]
+                    mprinc = unicode(mprinc.pop())
```
tiran, line 256: This block is hard to understand. It looks like you are looking for the last entry (or any entry?) that has matches the principal in lower case. A for loop would be easier to understand:

```
for p in e_attrs.get('krbprincipalname'):
    p = unicode(p)
    if p.lower() == nprincipal.lower():
        mprinc = p
        break
```

ipaserver/plugins/servicedelegation.py
```diff
@@ -314,10 +342,21 @@ def post_callback(self, ldap, completed, failed, dn, entry_attrs,
...
+                        reason=_("Malformed principal: %(error)s") % dict(
+                            error=str(e)))
+
+                if len(n.components) == 1 and (n.components[0][-1] != '$'):
+                    nprincipal = 'host/' + unicode(n)
+                else:
+                    nprincipal = unicode(n)
+                try:
+                    if nprincipal in entry_attrs.get(self.principal_attr, []):
+                        entry_attrs[self.principal_attr].remove(nprincipal)
```
tiran, line 359: This block looks very similar to the modified block below. How about moving the code into a shared helper function?

ipaserver/plugins/servicedelegation.py
```diff
@@ -212,22 +223,39 @@ def post_callback(self, ldap, completed, failed, dn, entry_attrs,
...
                 try:
-                    ldap.get_entry(obj_dn, ['krbprincipalname'])
+                    n = kerberos.Principal(name, realm=self.api.env.realm)
+                except ValueError:
+                    raise errors.ValidationError(
+                        name='principal',
+                        reason=_("Malformed principal: %(error)s") % dict(
+                            error=str(e)))
+
+                if len(n.components) == 1 and (n.components[0][-1] != '$'):
```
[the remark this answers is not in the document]
abbra: Please read the commit message.

abbra:
@tiran I moved the normalization part into a separate helper function and fixed most of remaining comments. There is no need to move actual attribute value check into the helper because it differs across the two callees.

ipaserver/plugins/servicedelegation.py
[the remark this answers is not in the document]
abbra: You cannot use `#is_user`, it makes no sense in the context we have here. You cannot distinguish host names and users by using that method because it only checks we have a single component.

We don't need to support users in service delegation rules, we have to add support for host principal instead. But since host principal is not a service (it is a host object) you either accept that the principal has one component (hostname itself) or do more expensive checks for existence of the host via LDAP queries. The latter is not really needed.

The only thing that prevents using this logic unconditionally are aliases for Windows-style machine accounts which consist of NetBIOS name of a machine with `$` sign appended. There are at least two users of this method: Samba and SQL Server. In case they need the delegation (S4U2Proxy is OK for them both, if needed), their service delegation rule has to contain Kerberos principal alias, not the canonical name (`cifs/...` or `MSSQLSvc/...`). In that case we must not prepend `host/..` to the single-component principal name.

abbra:
master:

* 1f82d281ccb4eb823ec5086f107410d7f8100c7d service delegation: allow to add and remove host principals

PR #4688, opened 2020-05-14
make: serialize strip-po / strip-pot

tiran:
The strip-po target modifies files in place. This sometimes creates
conflicts with other make targets when make is run in parallel mode.

* split strip-po into strip-po and strip-pot
* move strip-po[t] from dependency to explicit, serial execution
* declare dependencies on POT/POFILES
* don't run strip on clean

Fixes: https://pagure.io/freeipa/issue/8323

### DIFF
--- a/po/Makefile.hack.in
+++ b/po/Makefile.hack.in
@@ -9,10 +9,20 @@ MSGATTRIB = @MSGATTRIB@
 PYTHON = @PYTHON@
 GIT_BRANCH = @GIT_BRANCH@
 
-.PHONY: strip-po
-strip-po:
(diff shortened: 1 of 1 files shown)


stanislavlevin:
It seems that I'm too late.
https://github.com/stanislavlevin/freeipa/commit/934f10a5242bd126d4d87bc5f7305415dd6c0e78

abbra:
LGTM

tiran:
master:

* d20cda218960960fa4d1ab790f31edd4cac49ccc make: serialize strip-po / strip-pot

PR #4699, opened 2020-05-15
Don't serialize local methods

tiran:
Omit methods that server local, internal commands.

See: https://pagure.io/freeipa/issue/8324

### DIFF
--- a/ipalib/frontend.py
+++ b/ipalib/frontend.py
@@ -1298,7 +1298,10 @@ class Object(HasParam):
         )
         if self.primary_key:
             json_dict['primary_key'] = self.primary_key.name
-        json_dict['methods'] = list(self.methods)
+        json_dict['methods'] = [
+            name for name, method in self.methods.__todict__().items()
+            if not isinstance(method, Local)
+        ]
         return json_dict
 
 


ipalib/frontend.py
```diff
@@ -1298,7 +1298,10 @@ def __json__(self):
         )
         if self.primary_key:
             json_dict['primary_key'] = self.primary_key.name
-        json_dict['methods'] = list(self.methods)
+        json_dict['methods'] = [
+            name for name, method in self.methods.__todict__().items()
+            if not isinstance(method, Local)
```
rcritten, line 1303: Interestingly ipaserver/plugins/internal.py::json_metadata is already trying to filter these out.
I suppose if this is done at a lower level then the instance checking can be removed from json_metadata().

tiran: I think we still need it in higher levels when processing Method objects. This just filters out method names.

rcritten:
Maybe it was my procedure but I'm not seeing any difference between the before and after of this patch other than some minor reordering of values.

rcritten:
What I did was:
1. run the curl in the ticket, output to a file
2. apply the patch
3. run the same curl, output to new file
4. the file sizes are the same
5. load each into python json module and print with indent=2, dnsserver is still in updated version.

rcritten:
Tried writing an xmlrpc test for this but on my local system running the xmlrpc unit tests is failing. Could be a an environment issue, I dunno.

tiran:
PR didn't work and I have no time to investigate.

PR #4693, opened 2020-05-14
WebUI: Add authentication indicator specific fields to "Kerberos Ticket Policy" page

serg-cymbaluk:
Ticket: https://pagure.io/freeipa/issue/8207

PR #4703, opened 2020-05-19
baseldap: de-duplicate passed attributes when checking for limits

abbra:
LDAP attribute options aren't enforced in the schema, thus we strip them
when checking attribute conformance with the schema. This, however, can
leave us with a situation when multiple base LDAP attribute names are
present in the list of attribute names to check.

Use set of attribute names to deduplicate the list.

Fixes: https://pagure.io/freeipa/issue/8328

### DIFF
--- a/ipatests/test_xmlrpc/test_krbtpolicy.py
+++ b/ipatests/test_xmlrpc/test_krbtpolicy.py
@@ -290,7 +290,8 @@ class test_krbtpolicy(Declarative):
             command=(
                 'krbtpolicy_mod', [user1], dict(setattr=u'givenname=Pete')
             ),
-            expected=errors.ObjectclassViolation(info='attribute "givenname" not allowed'),
+            expected=errors.ObjectclassViolation(
+                info='these attributes are not allowed: givenname'),
         ),
     ]
     for (value, error) in invalid_values:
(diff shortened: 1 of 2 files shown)


abbra:
Solves problem noted in https://github.com/freeipa/freeipa/pull/4693

ipatests/test_xmlrpc/test_krbtpolicy.py
```diff
@@ -290,7 +290,8 @@ class test_krbtpolicy(Declarative):
             command=(
                 'krbtpolicy_mod', [user1], dict(setattr=u'givenname=Pete')
             ),
-            expected=errors.ObjectclassViolation(info='attribute "givenname" not allowed'),
+            expected=errors.ObjectclassViolation(
+                info='following attributes are not allowed: givenname'),
```
rcritten, line 294: This reads awkwardly in English because it's missing the article "the following...". Perhaps "these" instead of "following":

"these attributes are not allowed:"

or

"the following attributes are not allowed: "

rcritten:
master:

* 32c6b02eedf1243aa3ae716e7526118fe14fae59 baseldap: de-duplicate passed attributes when checking for limits

PR #4705, opened 2020-05-19
[Backport][ipa-4-6] Test for removing a subgroup

ssidhaye:
Problem description:
Removing an IPA sub-group should NOT remove the members
from indirect parent that also belong to other subgroups

The test:
A user and three groups are created a,b,c
'c' should be a child of 'b' so that you have a->b->c

user is direct member of 'a' and as a result member of 'b'
and 'c'. Now when one adds a direct membership to 'b' nothing will
change.

If one removes the direct membership to 'b' again,
nothing should change as well

Pagure Link: https://pagure.io/SSSD/sssd/issue/3636

Manual backport of #4544

### DIFF
--- a/ipatests/test_integration/test_sssd.py
+++ b/ipatests/test_integration/test_sssd.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import
 from contextlib import contextmanager
 import re
 import time
+import os
 
 import ipaplatform
 import pytest
@@ -17,6 +18,7 @@ import textwrap
 
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
+from ipatests.pytest_ipa.integration.tasks import clear_sssd_cache
 from ipatests.util import xfail_context
 from ipaplatform.tasks import tasks as platform_tasks
 from ipaplatform.paths import paths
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_sssd.py
```diff
@@ -348,3 +350,179 @@ def get_cache_update_time(obj_kind, obj_name):
         finally:
             sssd_conf_backup.restore()
             tasks.clear_sssd_cache(self.master)
+
+
+@pytest.skipif("SSSD fix is not available on Fedora 27")
```
flo-renaud, line 355: The annotation should be `@pytest.mark.skip(reason=...")` or `@pytest.mark.skipif(condition, reason="...")`

flo-renaud:
Backport LGTM.

@wladich do you prefer to have the @pytest.mark.skip annotation in a separate commit?

wladich:
@flo-renaud 
Yes, separate commit is highly preferred as it will help to avoid manual editing of the commit during backporting.

ssidhaye:
> @flo-renaud
> Yes, separate commit is highly preferred as it will help to avoid manual editing of the commit during backporting.

ok @wladich I'll move it to separate commit.

ipatests/test_integration/test_sssd.py
ssidhaye: fixed it

wladich:
@ssidhaye 
LGTM

rcritten:
LGTM, please remove the temporary commit.

rcritten:
Sorry, just noticed the commit message still doesn't match. It should look like:

Test for removing a subgroup

Problem description: 
Removing ...

I"m not trying to be pedantic but it is sometimes important when looking for changes between branches for the commit messages to match up.

No need to re-add a temporary commit as this just affects the message.

flo-renaud:
ipa-4-6:

* 5f5174b0c1d058d0d7926bcad7689e2166a5776f Test for removing a subgroup
* 6a1dd26dc6a62e37b501068634f1a6804a495ab5 Added pytest.mark.skip as SSSD fix for https://pagure.io/SSSD/sssd/issue/3636 is not available in Fedora 27

PR #4709, opened 2020-05-21
Allow leading and trailing space in user passwords

mrizwan93:
IPA doesn't allow leading and trainling spaces in user passwords.
This fix allows to have it.

Fixes: https://pagure.io/freeipa/issue/7599


rcritten:
I think you can redefine kwargs as:

kwargs = Data.kwargs

There are no patterns in passwords.

You'll also need to consider whether to update the HostPassword class in ipaserver/plugins/host.py so that host OTP passwords can begin/end with spaces. I'm not sure it's relevant here.

PR #4710, opened 2020-05-22
Azure: Make dnf repos consistent

stanislavlevin:
Build container(image registry.fedoraproject.org/f32/fedora-toolbox)
has two more dnf repos enabled compared to Tests container(image
fedora:32). This results in the packages built within the Build
container can have dependencies which are unresolvable(missing)
within Tests container.

This enables updates-testing and updates-testing-modular,
disables fedora-cisco-openh264 for Tests container.

Fixes: https://pagure.io/freeipa/issue/8330


abbra:
LGTM

abbra:
master:

* 26f96595b010396f055c7d11a645be57fc637863 Azure: Make dnf repos consistent

PR #4713, opened 2020-05-28
WebUI: Fix "IPA Error 3007: RequirmentError" while adding idoverrideuser association

serg-cymbaluk:
Add builder for association adder dialog which allows to override behavior of the component.
Replace default implementation with a custom one for idoverrideuser.
Replace text filter with 'ID view' select box in the idoverrideuser dialog.

Ticket: https://pagure.io/freeipa/issue/8335

### DIFF
--- a/install/ui/src/freeipa/association.js
+++ b/install/ui/src/freeipa/association.js
@@ -25,6 +25,7 @@
 define([
     'dojo/_base/lang',
     'dojo/Deferred',
+    './builder',
     './metadata',
     './ipa',
     './jquery',
@@ -38,7 +39,7 @@ define([
     './facet',
     './search',
     './dialog'],
-        function(lang, Deferred, metadata_provider, IPA, $, metadata,
+        function(lang, Deferred, builder, metadata_provider, IPA, $, metadata,
                  navigation, phases, reg, rpc, su, text) {
 
 /**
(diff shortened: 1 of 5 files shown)


serg-cymbaluk:
@abbra @pvoborni 

The patch fixes only adder dialog and, as result, allows to add a association with idoverrideuser.

But in other hand it disclose a couple of issues which I couldn't solve:

1) After adding idoverrideuser to group, it is not possible to navigate to the item.
The reason is I have only `ipaanchoruuid` values in `member_idoverrideuser` field (`group-show` call). But it is not enough to even get the object, because I need to have specify an ID view.

2) The anchor field is presented as `:SID:S-***-***-***`
As a result it is not possible to remove the idoverrideuser member due to a server error.

```
# ipa group-remove-member admins --idoverrideusers=Administrator@win.test
  Group name: admins
  Description: Account administrators group
  GID: 559600000
  Member users: admin
---------------------------
Number of members removed 1
---------------------------
```

but

```
# ipa group-remove-member admins --idoverrideusers=':SID:S-1-5-21-4064286453-263327960-3213440627-500'
```

pvoborni:
What is in httpd error log for the internal error?

abbra:
Please look at https://github.com/freeipa/freeipa/pull/4728 where I tried to merge all the changes required to make idoverrideuser membership working. I think with that one changes on Web UI side should be limited -- at least, I have no problem discussed here in my tests anymore.

serg-cymbaluk:
@abbra thanks
I checked my changes together with your patch and it works good. User ID override is displayed properly and 'delete' action is fixed.

@pvoborni 
I didn't find solution for the navigation issue, so for now I'm keeping plain text in the table instead of links.

abbra:
Please remove temp commit.

serg-cymbaluk:
@abbra 

Temp commit is removed.

serg-cymbaluk:
master:

* c2ba333b9681d008d9c528a79dbdd76ce11a3ecd WebUI: Fix "IPA Error 3007: RequirmentError" while adding idoverrideuser association

PR #4714, opened 2020-05-28
Test for PG7961 - WebUI requires to click on save changes after change

miskopo:
Test adds entry into sudorules and checks whether save button is
disabled.

Related: https://pagure.io/freeipa/issue/7961

### DIFF
--- a/ipatests/setup.py
+++ b/ipatests/setup.py
@@ -75,7 +75,7 @@ if __name__ == '__main__':
             "six",
         ],
         extras_require={
-            "integration": ["dbus-python", "pyyaml", "ipaserver"],
+            "integration": ["dbus-python", "pyyaml", "ipaserver", "homoglyphs"],
             "ipaserver": ["ipaserver"],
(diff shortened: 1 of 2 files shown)


ipatests/test_webui/test_sudo.py
```diff
@@ -133,3 +133,13 @@ def test_actions(self):
...
+
+    @screenshot
+    def test_autosave(self):
+        """
+        Test autosave is active and "Save" button is disabled
+        """
+        self.init_app()
+        self.add_record(sudo.RULE_ENTITY, sudo.RULE_DATA)
+        save_btn = self.find("save", by="name")
+        assert save_btn.get_attribute("disabled")
```
mrizwan93, line 145: I would delete the record at the end of the test as a cleanup so that it won't affect subsequent test.

miskopo: Thank you, mrizwan93, you are absolutely right

PR #4715, opened 2020-05-28
Fix design bug in named ipa-ext.conf

tiran:
## Split named custom config to allow changes in options stanza

Upgrade path to add additional include to named.conf is not handled.

Remove bindkeys-file directive from named config
The ISC DVL service was shut down (isc.org/bind-keys).
BIND versions since April 2017 (i.e. 9.9.10, 9.10.5, 9.11.1 and later)
include a hard-coded copy of the root KSK which gets updates automatically
according to RFC 5011.

Move dnssec-enable directive to custom named config

Move comment named config being managed by FreeIPA to the top

Move settings which could be changed by administrators to
ipa-options-ext.conf. Settings defined there are sole responsibility of the
administrator. We do not check if they might collide with our settings in
named.conf.

Fixes: https://pagure.io/freeipa/issue/8287
Co-authored-by: Peter Keresztes Schmidt <carbenium@outlook.com>

## backup named configuration


abbra:
LGTM. A potential issue is when we upgrade and ipa-ext.conf already exists, we don't overwrite it, so the whole configuration might be broken. Unfortunately, I don't see good way of handling this other than parsing the snippet and moving "options { .. };" part to the second customization file.

tiran:
This is still a WIP draft. It's missing tests and migration.

abbra:
I'm going to ACK this part. @tiran please create a separate ticket for the remaining testing and upgrade part.

abbra:
master:

* a5cbdb57e50cfc62f61affda19ce878b2abd33de Split named custom config to allow changes in options stanza
* 6e5d40e2d2a53e44270fb18a7be3b0d7e346f46d Include named config files in backup

PR #4718, opened 2020-05-28
Allow dnsrecord-add --force on clients

tiran:
See: https://pagure.io/freeipa/issue/8317

### DIFF
--- a/ipaserver/plugins/dns.py
+++ b/ipaserver/plugins/dns.py
@@ -3550,7 +3550,6 @@ class dnsrecord_add(LDAPCreate):
     takes_options = LDAPCreate.takes_options + (
         Flag('force',
              label=_('Force'),
-             flags=['no_option', 'no_output'],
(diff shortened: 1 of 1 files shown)


abbra:
Another part that will fail is the test `test_permission_flags` which calls `_make_permission_flag_tests()` with `SYSTEM` permission flag. The test creates and then deletes the permission and without `--force` one cannot delete the system permissions.

I think we should actually move these tests to a normal test that can clean up resulting system permissions via direct LDAP modification. Perhaps, this could be a combined test of permission and upgrade plugin? Because upgrade plugin can return DNs of entries to remove.

tiran:
I've reduced the PR to only allow force for dnsrecord-add.

abbra:
LGTM.

flo-renaud:
master:

* ad8e0af0770caf039ff6c431a3182238c1952f8f Allow dnsrecord-add --force on clients

PR #4719, opened 2020-05-28
ipatests: Test ipa user login with wrong password

mrizwan93:
When ipa user login to machine using wrong password, it
should log proper message in /var/log/secure

related: https://github.com/SSSD/sssd/issues/5139

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -1192,6 +1192,11 @@ class TestIPACommand(IntegrationTest):
         related: https://github.com/SSSD/sssd/issues/5139
         """
         # try to login with wrong password
+        sssd_version = tasks.get_sssd_version(self.master)
+        if (sssd_version < tasks.parse_version('2.3.0')):
+            pytest.xfail('Fix is part of sssd 2.3.0 and is'
+                         ' available from fedora32 onwards')
+
         sshconn = paramiko.SSHClient()
         sshconn.set_missing_host_key_policy(paramiko.AutoAddPolicy())
         since = time.strftime('%H:%M:%S')
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -1170,3 +1170,44 @@ class TestIPACommand(IntegrationTest):
             assert msg2 not in result.stderr_text
         finally:
             bashrc_backup.restore()
+
+    @pytest.fixture
+    def user_creation_deletion(self):
+        # create user
+        self.testuser = 'testuser'
+        tasks.create_active_user(self.master, self.testuser, 'Secret123')
+
+        yield
+
+        # cleanup
+        tasks.kinit_admin(self.master)
+        self.master.run_command(['ipa', 'user-del', self.testuser])
+
+    def test_login_wrong_password(self, user_creation_deletion):
+        """Test ipa user login with wrong password
+
+        When ipa user login to machine using wrong password, it
+        should log proper message
+
+        related: https://github.com/SSSD/sssd/issues/5139
+        """
+        # try to login with wrong password
+        sshconn = paramiko.SSHClient()
+        sshconn.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+        since = time.strftime('%H:%M:%S')
+        try:
+            sshconn.connect(self.master.hostname,
+                            username=self.testuser,
+                            password='WrongPassword')
+        except paramiko.AuthenticationException:
+            pass
+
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_commands.py
```diff
@@ -1170,3 +1170,35 @@ def test_ipa_adtrust_install_with_locale_issue8066(self):
...
+                            password='WrongPassword')
+        except paramiko.AuthenticationException:
+            pass
+
+        sshconn.close()
+
+        # check if proper message logged
+        exp_msg = ("pam_sss(sshd:auth): received for user {}: 7"
+                   "(Authentication failure)".format(testuser))
+        result = self.master.run_command(['tail', '/var/log/secure'])
```
amore17, line 1201: As per looking https://github.com/SSSD/sssd/issues/5139
It should be running on client right ?

flo-renaud: @amore17 
the test is doing ssh from controller to master, meaning the sshd logs must be checked on the master.
@mrizwan93 
on Fedora sshd is logging to the journal, not to /var/log/secure as on RHEL (because rsyslog is not shipped by default). The test will have to be adapted.

flo-renaud:
Hi @mrizwan93 
A comment related to SSSD version:
- On sssd 1-16 branch, the fix has not been incorporated in any fedora build yet. This sssd 1-16 branch is used for fedora <= 28, meaning that our PRCI upstream tests on ipa-4-6 will fail.
- RHEL 7.9 will be shipping 1.16.5-6 with the fix. Downstream tests on ipa-4-6 will succeed.
- The issue is fixed in sssd master's branch for SSSD 2.3.0+, which is currently available from fedora 32. This means that the test will fail in fedora 31, which is used in [testing_master_previous].

Summary: the backport is needed for ipa-4-6 and ipa-4-8, but you need to add a xfail. You can either:
- test sssd version and add a xfail based on that version in a separate commit (inside this PR), and notify ipa-tests maintainer to avoid picking the commit with the xfail for the RHEL7 build
- or check the os version

ipatests/test_integration/test_commands.py
```diff
@@ -1170,3 +1170,35 @@ def test_ipa_adtrust_install_with_locale_issue8066(self):
             assert msg2 not in result.stderr_text
         finally:
             bashrc_backup.restore()
+
+    def test_login_wrong_password(self):
+        """Test ipa user login with wrong password
+
+        When ipa user login to machine using wrong password, it
+        should log proper message in /var/log/secure
```
flo-renaud, line 1178: Please update the comment, the message is not checked in /var/log/secure on fedora. I would simply write "it should log proper message".

ipatests/test_integration/test_commands.py
```diff
@@ -1170,3 +1170,35 @@ def test_ipa_adtrust_install_with_locale_issue8066(self):
...
+                            password='WrongPassword')
+        except paramiko.AuthenticationException:
+            pass
+
+        sshconn.close()
+
+        # check if proper message logged
+        exp_msg = ("pam_sss(sshd:auth): received for user {}: 7"
+                   " (Authentication failure)".format(testuser))
+        result = self.master.run_command(['journalctl', '-u', 'sshd'])
```
flo-renaud, line 1205: Instead of checking the whole journal, I would add the option `'--since={}'.format(since)` and define
`since = time.strftime('%H:%M:%S')` at the beginning of the test. Otherwise the error may be triggered by another test.

ipatests/test_integration/test_commands.py
```diff
@@ -1170,3 +1170,35 @@ def test_ipa_adtrust_install_with_locale_issue8066(self):
...
+        assert exp_msg in result.stdout_text
+
+        tasks.kinit_admin(self.master)
+        self.master.run_command(['ipa', 'user-del', testuser])
```
flo-renaud, line 1209: The test would be more robust if you use a fixture to do user creation/user deletion.

flo-renaud:
Hi @mrizwan93 
tests are green and code LGTM. Please remove the temp commit and I will ACK when PRCI completes.

flo-renaud:
master:

* e33ffab534c9ecce26fbd069956d15da182ce6ed ipatests: Test ipa user login with wrong password
* 32a5359ece5b449b1cedb700d79cfa2744a288cb Xfail test for sssd < 2.3.0

flo-renaud:
@mrizwan93 
the automated backport failed for ipa-4-6 branch, can you do a manual backport? Thanks

PR #4725, opened 2020-06-01
Remove remains of unused config options

Carbenium:
Options removed:
 * ca_ee_port
 * ca_agent_install_port
 * ca_ee_install_port

Closes: https://pagure.io/freeipa/issue/6708

### DIFF
--- a/client/man/default.conf.5
+++ b/client/man/default.conf.5
@@ -68,9 +68,6 @@ Specifies the base DN to use when performing LDAP operations. The base must be i
 .B ca_agent_port <port>
 Specifies the secure CA agent port. The default is 8443.
 .TP
-.B ca_ee_port <port>
-Specifies the secure CA end user port. The default is 8443.
-.TP
 .B ca_host <hostname>
 Specifies the hostname of the dogtag CA server. The default is the hostname of the IPA server.
 .TP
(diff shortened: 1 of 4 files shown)


Carbenium:
Although the issue mentions deprecation notices should be added, I opted not to.
Since removing the options won't break the existing configs, it might be enough to mention it in the changelogs.
If this is not acceptable, I'll add the notices.

rcritten:
Even if these are defined in a configuration file it won't cause issues because they are referenced anywhere within IPA other than defaults. The unknown values will be represented in api.env but otherwise won't be used.

For example:
```
$ cat /etc/ipa/default.conf
[global]
...
ca_ee_port = 443
foobar = baz
$ ipa console
(Custom IPA interactive Python console)
    api: IPA API object
    pp: pretty printer
>>> api.env.ca_ee_port
443
>>> api.env.foobar
'baz'
>>> api.env.unknown
  File "/usr/lib64/python3.7/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
AttributeError: 'Env' object has no attribute 'unknown'
```

So if a user had ca_ee_port set previously it was a no-op and will continue to be a no-op with this patch.

Carbenium:
@rcritten Yep. Exactly what I meant.

_Side-note:_ Good to know there is `ipa console`. Didn't know about that.

tiran:
LGTM, thanks!

I'm marking this for 4.9 to further reduce the chance to break stuff.

tiran:
master:

* 1606174457a3db5f8300fca2230f211d5f72dfa1 Remove remains of unused config options

PR #4726, opened 2020-06-02
Allow permissions with 'self' bindruletype

tiran:
Make it possible to create a managed permission with
ipapermbindruletype="self". The ACI will have bind rule
'(userdn = "ldap:///self")'.

Example
-------

Allow users to modify their own fasTimezone and fasIRCNick attributes:

```
managed_permissions = {
    "System: Self-Modify FAS user attributes": {
        "ipapermright": {"write"},
        "ipapermtargetfilter": ["(objectclass=fasuser)"],
        "ipapermbindruletype": "self",
        "ipapermdefaultattr": ["fasTimezone", "fasIRCNick"],
    }
}
```

Fixes: https://pagure.io/freeipa/issue/8348


rcritten:
There is no ticket outlining the use case(s) and reasoning for this.

ipaserver/install/plugins/update_managed_permissions.py
```diff
@@ -635,6 +635,10 @@ def update_entry(self, obj, entry, template,
 
         # Attributes from template
         bindruletype = template.pop('ipapermbindruletype', 'permission')
+        if bindruletype not in {"all", "anonymous", "self", "permission"}:
```
tiran, line 638: I added this check to make the plugin fail earlier and to give a better error message.

rcritten:
The commit message references the PR number and not the ticket, https://pagure.io/freeipa/issue/8348

ipatests/test_xmlrpc/test_permission_plugin.py
```diff
@@ -4345,3 +4345,48 @@ class test_autoadd_operational_attrs(Declarative):
...
+        ),
+
+        verify_permission_aci(
+            permission1, DN('cn=accounts', api.env.basedn),
+            '(targetattr = "ObjectClass || createtimestamp || entryusn || '
+            + 'modifytimestamp")'
+            + '(version 3.0;acl "permission:%s";' % permission1
+            + 'allow (read) userdn = "ldap:///self";)',
+        ),
+    ]
```
rcritten, line 4392: A bit of a nit but we try to keep attributes lower-case in the python code and ObjectClass really stands out in any case.

rcritten: Can you expand in the commit message that this is so a plugin can provide its own self-service rule and not rely on the global one? This is quite useful in particular for the external use case you found in the FAS plugin.

tiran: Good point, I copy pasta the code from another test and didn't bother to change any non-related strings.

rcritten:
Looks good, will add ack once CI passes

ipaserver/plugins/permission.py
```diff
@@ -279,7 +279,7 @@ class permission(baseldap.LDAPObject):
             label=_('Bind rule type'),
             doc=_('Bind rule type'),
             autofill=True,
-            values=(u'permission', u'all', u'anonymous'),
+            values=(u'permission', u'all', u'anonymous', 'self'),
```
abbra, line 282: the other values explicitly marked as unicode, this one is not. Should this be fixed?

tiran: It's Python 3, the ``u''`` prefix is just for show and doesn't affect the type.

tiran: I left it out deliberately.

abbra: It looks weird to have it all mixed on the same line. Either remove u'' or have it everywhere in that line.

rcritten:
LGTM

abbra:
master:

* 9dda004f270f321685597be276ca51e6b32fcd59 Allow permissions with 'self' bindruletype

PR #4737, opened 2020-06-03
ipa-backup/restore: remove remaining chdir calls

Carbenium:
Closes: https://pagure.io/freeipa/issue/7416

### DIFF
--- a/ipaserver/install/ipa_backup.py
+++ b/ipaserver/install/ipa_backup.py
@@ -306,7 +306,6 @@ class Backup(admintool.AdminTool):
 
         self.header = os.path.join(self.top_dir, 'header')
 
-        cwd = os.getcwd()
         try:
             dirsrv = services.knownservices.dirsrv
 
@@ -359,10 +358,6 @@ class Backup(admintool.AdminTool):
                                  options.gpg_keyring)
 
         finally:
-            try:
-                os.chdir(cwd)
-            except Exception as e:
-                logger.error('Cannot change directory to %s: %s', cwd, e)
             shutil.rmtree(self.top_dir)
 
     def check_roles(self, raiseonerr=True):
@@ -761,11 +756,10 @@ class Backup(admintool.AdminTool):
                 'Unexpected error: %s' % e
             )
 
-        os.chdir(self.dir)
         args = [
             'tar', '--xattrs', '--selinux', '-czf', filename, '.'
         ]
-        result = run(args, raiseonerr=False)
+        result = run(args, raiseonerr=False, cwd=self.dir)
         if result.returncode != 0:
             raise admintool.ScriptError(
                 'tar returned non-zero code %s: %s' %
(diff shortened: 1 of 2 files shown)


Carbenium:
CI failure seems unrelated. CA timeout?

abbra:
You can re-run failed jobs if you go to Azure Pipelines page and press 'rerun failed jobs' there. Or you can do '/azp run' in the comment.

ipaserver/install/ipa_backup.py
```diff
@@ -357,10 +356,11 @@ def run(self):
                                  options.gpg_keyring)
 
         finally:
+            tmp_dir = tempfile.gettempdir()
             try:
-                os.chdir(cwd)
+                os.chdir(tmp_dir)
             except Exception as e:
-                logger.error('Cannot change directory to %s: %s', cwd, e)
+                logger.error('Cannot change directory to %s: %s', tmp_dir, e)
```
rcritten, line 363: I'm not sure another temporary directory is needed here, or perhaps any chdir at all. The previous chdir was needed because it did a chdir(self.dir) so in order to clean up one needed to chdir out. Since chdir(self.dir) isn't done any more there is no need to change to somewhere else.

ipaserver/install/ipa_restore.py
```diff
@@ -483,10 +481,11 @@ def run(self):
                 if result.returncode != 0:
                     logger.error('Restarting IPA failed: %s', result.error_log)
         finally:
+            tmp_dir = tempfile.gettempdir()
             try:
-                os.chdir(cwd)
+                os.chdir(tmp_dir)
             except Exception as e:
-                logger.error('Cannot change directory to %s: %s', cwd, e)
+                logger.error('Cannot change directory to %s: %s', tmp_dir, e)
```
rcritten, line 488: Same as in backup. I  think the change to run() is enough in itself.

ipaserver/install/ipa_backup.py
Carbenium: You might be right. I left it in since I didn't play enough close attention what self.top_dir actually is. I thought it might be possible to start the script from inside it and then the `chdir` would be necessary. Since it is a `tempfile.mkdtemp` this definitely isn't possible. I'll test the change and push an update.

Carbenium: Removed and tested. Seems to work just fine.

ipaserver/install/ipa_restore.py
Carbenium: Removed and tested. Seems to work just fine.

rcritten:
@Carbenium can you rebase this? The vagrant templates have changed. I tested this and it looks ok, just need it to pass CI and then I'll ack it.

Carbenium:
@rcritten Of course. Here you go.

rcritten:
master:

* cf8ef6fd2d275ea50758770c72db7faf4cedf0e3 ipa-backup/restore: remove remaining chdir calls

PR #4738, opened 2020-06-03
[ipa-4-6] ipa-adtrust-install: avoid failure when replica is offline

flo-renaud:
When ipa-adtrust-install --add-agents fails to connect to the
replica (for instance because services are stopped on the replica),
it should not exit on error but rather print an error message.

The issue happens because of python2/python3 differences in the
returned exceptions. As ipa-4-6 branch also supports python2,
a socket.error exception is raised when the connection fails and
must be properly caught.

Fixes: https://pagure.io/freeipa/issue/8345

### DIFF
--- a/ipaserver/install/adtrust.py
+++ b/ipaserver/install/adtrust.py
@@ -8,7 +8,6 @@ AD trust installer module
 
 from __future__ import print_function, absolute_import
 
-import errno
 import logging
 import os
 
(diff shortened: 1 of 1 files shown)


fcami:
ACK. 
As a side note, I think the updated code will also behave better in case of Timeouterror. This would help with silent firewalls (essentially iptables -j DROP rather than -j REJECT).

flo-renaud:
ipa-4-6:

* dad858a70c551a00fa1cf8f1c7ac3f6122b6df24 ipa-adtrust-install: avoid failure when replica is offline

PR #4739, opened 2020-06-03
Run pytest without color in Azure

tiran:
The raw log files get littered with ANSI escape sequences when pytest colors the output

### DIFF
--- a/ipatests/azure/scripts/azure-run-base-tests.sh
+++ b/ipatests/azure/scripts/azure-run-base-tests.sh
@@ -45,6 +45,7 @@ if [ "$install_result" -eq 0 ] ; then
         --logging-level=debug \
         --logfile-dir="$IPA_TESTS_LOGSDIR" \
         --verbose \
+        --color=no \
         --with-xunit \
         '-k not test_dns_soa' \
         $IPA_TESTS_TO_IGNORE \
--- a/ipatests/azure/scripts/azure-run-integration-tests.sh
+++ b/ipatests/azure/scripts/azure-run-integration-tests.sh
@@ -14,8 +14,9 @@ tests_result=1
     ipa-run-tests \
     --logging-level=debug \
     --logfile-dir="$IPA_TESTS_LOGSDIR" \
-    --with-xunit \
     --verbose \
+    --color=no \
+    --with-xunit \
     $IPA_TESTS_TO_IGNORE \
     $IPA_TESTS_TO_RUN && tests_result=0 ; } || \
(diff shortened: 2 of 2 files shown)


stanislavlevin:
I'm sorry what is your problem?
For example, I use the next logs viewers:
1) a web browser on Azure
2) less -r/-R runner.log
Using coloring helps me greatly to parse results by eyes.

tiran:
The problem is that the raw logs are littered with ANSI escape sequences and therefore hard to read.

* Page up / down in HTML log view like [this](https://dev.azure.com/freeipa/freeipa/_build/results?buildId=4507&view=logs&j=c3b3beac-3142-545b-181b-8ae72407437a&t=4475fbc0-1119-5fa5-c1d0-2f0cb6cb4700) is not working for me in Firefox
* The [raw output](https://dev.azure.com/freeipa/dd9bf4ec-d636-4558-8f6e-ee17af87f33b/_apis/build/builds/4507/logs/121) is hard to read.

stanislavlevin:
I give up :-) I'm outnumbered.

rcritten:
I'll hold off adding the ack for a bit longer to see if anyone else has any input on it.

stanislavlevin:
That's is a matter of opinion. People do use different tools/editors/viewers.
We cannot please everyone. The majority should win.

rcritten:
Agreed, that's why I want to hold off, in order to be sure this doesn't come as a surprise to those that haven't yet had an opportunity to look at this. The majority right now is no color, that could change with more input.

stanislavlevin:
Logs in question are Pytest _runner_ logs (stdout + stderr).
However, it is possible to collect logging records coming from tests into specified paths or even include them into XML report.

stanislavlevin:
Another colored feature I used is indicating the overall status of a runner.

Screenshot_20200605_151121

It's helpful for time-consuming tests since I haven't to wait for all tests to finish.

tiran:
FWIW I don't care how the problem is fixed. I just want to be able to browser and read the logs efficiently. In Firefox mouse wheel scrolling of HTML is super slow, page up/down doesn't work for me at all, and raw logs are littered with ANSI escape sequences. If you get me another way to view raw logs inside my default browser easily, then I'm happy to close the PR.

stanislavlevin:
I have no objections.
It seems there will be no more input.

fcami:
@tiran could you please rebase? the CI issue is known and was fixed with more recent images.

PR #4740, opened 2020-06-03
ipatests: Test to check warning state for TomcatFileCheck in ipahealthcheck.ipa.files

menonsudhir:
This testcase changes the ownership of the tomcat config files on an IPA Master and then checks if healthcheck tools reports the status as WARNING


rcritten:
Looks good, please remove the temp commit and I'll add the ack.

rcritten:
master:

* 0d0dc73ae1fd640841a4b15ed83edc44096859e3 ipatests: Test to check warning state for TomcatFileCheck in ipahealthcheck.ipa.files

PR #4749, opened 2020-06-05
Handle DatabaseError in RPC-Server connect()

tiran:
DatabaseError exceptions with 'account inactivated' message are turned
into 401 Unauthorized errors. The problem occurs when a user is disabled
but has a valid cookie.

Other DatabaseErrors are turned into 503 Service Unavailable. They
usually occur when LDAP server is not available or broken.

Fixes: https://pagure.io/freeipa/issue/8352


abbra:
LGTM

abbra:
Should this be added to ipa-4-6 as well? I would imagine it affecting quite old code too.

abbra:
master:

* d79a7a9696ff634141bdff3423157d08513da310 Handle DatabaseError in RPC-Server connect()

PR #4750, opened 2020-06-05
Cleanup GCC warning flags

Carbenium:
Related: https://pagure.io/freeipa/issue/8351

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -460,6 +460,7 @@ if test "$GCC" = "yes" -a "$set_more_warnings" != "no"; then
         -Wchar-subscripts -Wmissing-declarations -Wmissing-prototypes \
         -Wnested-externs -Wpointer-arith \
         -Wcast-align -Wsign-compare \
+        -Wshadow -Wstrict-prototypes \
         $CFLAGS"
 
         for option in -Wno-strict-aliasing -Wno-sign-compare; do
--- a/configure.ac
+++ b/configure.ac
@@ -446,7 +446,7 @@ AC_ARG_ENABLE(more-warnings,
 	[AC_HELP_STRING([--enable-more-warnings],
 	[Maximum compiler warnings])],
 	set_more_warnings="$enableval",[
-        	if test -d $srcdir/../.hg; then
+        	if test -d $srcdir/.git; then
         		set_more_warnings=yes
               	else
                   	set_more_warnings=no
--- a/configure.ac
+++ b/configure.ac
@@ -462,22 +462,6 @@ if test "$GCC" = "yes" -a "$set_more_warnings" != "no"; then
         -Wcast-align -Wsign-compare \
         -Wshadow -Wstrict-prototypes \
         $CFLAGS"
-
-        for option in -Wno-strict-aliasing -Wno-sign-compare; do
-                SAVE_CFLAGS="$CFLAGS"
-                CFLAGS="$CFLAGS $option"
-                AC_MSG_CHECKING([whether gcc understands $option])
-                AC_TRY_COMPILE([], [],
-                        has_option=yes,
-                        has_option=no,)
-                if test $has_option = no; then
-                        CFLAGS="$SAVE_CFLAGS"
-                fi
-                AC_MSG_RESULT($has_option)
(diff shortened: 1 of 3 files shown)


Carbenium:
@rcritten Maybe you could take a look at this, since you were the original author of the touched code

tiran:
In general we try to support all C99 compatible compilers on all POSIX platforms, not just GCC on Linux. At least the client package should compile with older clang and perhaps in AIX and Solaris, too.

configure.ac
```diff
@@ -435,26 +435,11 @@ if test "$GCC" = "yes" -a "$set_more_warnings" != "no"; then
...
-                        has_option=yes,
-                        has_option=no,)
-                if test $has_option = no; then
-                        CFLAGS="$SAVE_CFLAGS"
-                fi
-                AC_MSG_RESULT($has_option)
-                unset has_option
-                unset SAVE_CFLAGS
-        done
-        unset option
```
Carbenium, line 457: Reasoning for this removal:
 * `-Wsign-compare` is explicitly set, we shouldn't set `-Wno-sign-compare`
* GCC enables strict-aliasing rules by default. Disabling the corresponding warings with `-Wno-strict-aliasing` seems evil

rcritten: This code dates back to 2009 so we're talking Fedora 9/10-ish depending on what release I was running at the time. RHEL-5 for sure. The context was the nearly full re-write of the tree while moving from v1.x to 2.0.0. That's likely how the mercurial directory slipped through.

Perhaps -Wchar-subscripts should remain to be explicit and not rely on compiler defaults (which can always change). I think the rest of the changes are ok. It is not enable by default in clang, for example, per https://clang.llvm.org/docs/DiagnosticsReference.html

Beyond being hidden behind an option that everyone probably forgot long ago it actually turns OFF some warnings. It may have made sense at the time to ignore false positives.

Right now there are more than 20 -Wsign-compare warnings, all post-dating this change but with lots of code moving around I didn't follow every one. It doesn't look like disabling this was in reaction to any current code.

By my count there are 111 warnings in the code vs 1 when enable-more-warnings is not set.

I have a vague memory that we intended to add a raw -Werror at some point in order to fail on all errors. It is currently restricted to implicit-function-declaration.

Having the gcc-specific CFLAGS behind the option HAVE_GCC is probably the safest route forward but it makes sense IMHO to centralize CFLAGS in $top/Makefile.am and not distribute it.

For example, it seems that clang has -Wpointer-arithmatic and gcc uses -Wpointer-arith.

There are still plenty of Solaris and AIX users for sure and they should be able to still leverage at least ipa-join if not ipa-getkeytab as well. I don't know if they are.

Carbenium: I added the explicit `-Wchar-subscripts` back. I'd be glad if somebody else more skilled/experienced could do the `Makefile.am` conversion. I don't wan't to break anything I might not be able to test for.

abbra:
I'll keep it here for a week, we need to get other PRs pushed first for the 4.8.7 release. When we are ready, I'll do PR CI run for this.

fcami:
@Carbenium Apologies for the delay. Could you rebase to master? Our CI now depends on newer images than the ones that were available when you created the PR.

Carbenium:
@fcami No worries. Here you go.

rcritten:
I think this looks ok.

I'm not sure if the needs-rebase flag is stale or not.

Carbenium:
@rcritten Sorry for the delay. Just to be sure I rebased it.

Carbenium:
@rcritten Is this something you would still consider merging?

rcritten:
Yes.

abbra:
Let's merge it.

flo-renaud:
master:

* 54b42f72fb1cd70c28f52feb5c8eaf2bca9e32cb configure: Fix source tree detection to enable more warnings
* a355646c37f8b36549d79b2f1cbaeb428d2c98d8 build: Unify compiler warning flags used
* f9357cb986dda76dbee795e7b602662ffb4602b6 configure: Do not set -Wno-strict-aliasing -Wno-sign-compare

PR #4751, opened 2020-06-05
 ldapupdate: Use proper error mappings provided by ipaldap

Carbenium:
Closes: https://pagure.io/freeipa/issue/7610

### DIFF
--- a/ipaserver/install/ldapupdate.py
+++ b/ipaserver/install/ldapupdate.py
@@ -32,7 +32,6 @@ import os
 import pwd
 import fnmatch
 
-import ldap
 import six
 
 from ipaserver.install import installutils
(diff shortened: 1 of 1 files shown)


ipaserver/install/ldapupdate.py
```diff
@@ -32,7 +32,6 @@
 import pwd
```
rcritten, line 32: Can you change the title to something like "Use ipaldap exceptions rather than ldap error codes in LDAP updater"

and add to the commit message this line from the ticket:

The code in ipaldap got changed with df4ed77 but ldapupdate was never updated. 

I think this will clarify in the title the reason for the change (for release notes) and add additional context if someone comes across this in the future, w/o having to dereference the ticket.

ipaserver/install/ldapupdate.py
```diff
@@ -63,10 +62,7 @@ def connect(ldapi=False, realm=None, fqdn=None, dm_password=None):
             fqdn, decode_attrs=False
         )
     try:
-        if dm_password:
-            conn.simple_bind(bind_dn=ipaldap.DIRMAN_DN,
-                             bind_password=dm_password)
-        elif os.getegid() == 0:
```
tiran, line 69: Why did you remove this code?

Carbenium: See comment by @rcritten in the ticket:
>Near as I can tell only schema_update() uses ldapupdate.connect() and never passes in a DM password so looks like simple bind can be dropped altogether.

tiran: Any removed feature means that we cannot backport this PR to 4.8 branch.

ipaserver/install/ldapupdate.py
Carbenium: Done

ipaserver/install/ldapupdate.py
Carbenium: I removed the commit from PR. If requested I can submit it separately.

rcritten: Right, ipa-4-8 still uses connect with dm_password all over the place. A separate PR for "next" branch works for me.

Carbenium: :+1:

abbra:
LGTM

abbra:
master:

* e66036481477ec8efc27c697e8a0d7adaf8a691d Use ipaldap exceptions rather than ldap error codes in LDAP updater

PR #4754, opened 2020-06-05
Specify min and max values for TTL of a DNS record

Carbenium:
Fixes: https://pagure.io/freeipa/issue/8358


abbra:
master:

* 373f8cdce707e250e88f43cd59aeae3e1b5d9d12 Specify min and max values for TTL of a DNS record

PR #4772, opened 2020-06-08
ipatests: Test for ipa-nis-manage CLI tool

menonsudhir:
The testcases check the various options of ipa-nis-manage CLI tool as below

1. ipa-nis-mange enable
2. ipa-nis-manage disable
3. Enabling NIS plugin with invalid admin password

### DIFF
--- a/ipatests/test_integration/test_adtrust_install.py
+++ b/ipatests/test_integration/test_adtrust_install.py
@@ -7,12 +7,16 @@
 from __future__ import absolute_import
 
 import re
+import time
 import textwrap
 
 from ipaplatform.paths import paths
+from ipaserver.install.installutils import realm_to_serverid
 from ipapython.dn import DN
 from ipatests.pytest_ipa.integration import tasks
 from ipatests.test_integration.base import IntegrationTest
+
+DIRSRV_SLEEP = 5
 
 
 class TestIpaAdTrustInstall(IntegrationTest):
@@ -216,3 +220,81 @@ class TestIpaAdTrustInstall(IntegrationTest):
         value = (r'ipaexternalmember=%deref_r('
                  '"member","ipaexternalmember")')
         assert value in entry_list
+
+
+class TestIpaNisManage(IntegrationTest):
+    """
+    Tests for ipa-nis-manage CLI tool
+    """
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master)
+
+    def get_dirsrv_id(self):
+        serverid = realm_to_serverid(self.master.domain.realm)
+        return("dirsrv@%s.service" % serverid)
+
+    def test_ipa_nis_manage_enable(self):
+        """
+        This testcase checks if ipa-nis-manage enable
+        command enables plugin on an IPA master
+        """
+        dirsrv_service = self.get_dirsrv_id()
+        console_msg = (
+            "Enabling plugin\n"
+            "This setting will not take effect until you restart Directory Server.\n"
+            "The rpcbind service may need to be started"
+        )
+        status_msg = "Plugin is enabled"
+        tasks.kinit_admin(self.master)
+        result = self.master.run_command(
+            ["ipa-nis-manage", "enable"],
+            stdin_text=self.master.config.admin_password,
+        )
+        assert console_msg in result.stdout_text
+        self.master.run_command(["systemctl", "restart", dirsrv_service])
+        time.sleep(DIRSRV_SLEEP)
+        self.master.run_command(["systemctl", "restart", "rpcbind"])
+        result = self.master.run_command(
+            ["ipa-nis-manage", "status"],
+            stdin_text=self.master.config.admin_password,
+        )
+        assert status_msg in result.stdout_text
+
+    def test_ipa_nis_manage_disable(self):
+        """
+        This testcase checks if ipa-nis-manage disable
+        command disable plugin on an IPA Master
+        """
+        dirsrv_service = self.get_dirsrv_id()
+        msg = "This setting will not take effect until you restart Directory Server."
+        status_msg = "Plugin is not enabled"
+        tasks.kinit_admin(self.master)
+        result = self.master.run_command(
+            ["ipa-nis-manage", "disable"],
+            stdin_text=self.master.config.admin_password,
+        )
+        assert msg in result.stdout_text
+        self.master.run_command(["systemctl", "restart", dirsrv_service])
+        time.sleep(DIRSRV_SLEEP)
+        result = self.master.run_command(
+            ["ipa-nis-manage", "status"],
+            stdin_text=self.master.config.admin_password,
+            raiseonerr=False,
+        )
+        assert result.returncode == 4
+        assert status_msg in result.stdout_text
+
+    def test_ipa_nis_manage_enable_incorrect_password(self):
+        """
+        This testcase checks if ipa-nis-manage enable
+        command throws error on console for invalid DS admin password
+        """
+        msg = "Insufficient access:  Invalid credentials\n"
+        result = self.master.run_command(
+            ["ipa-nis-manage", "enable"],
+            stdin_text='Invalid_pwd',
+            raiseonerr=False,
+        )
+        assert result.returncode == 1
+        assert msg in result.stderr_text
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
--- a/ipatests/prci_definitions/temp_commit.yaml
+++ b/ipatests/prci_definitions/temp_commit.yaml
@@ -69,7 +69,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-27/build_url}'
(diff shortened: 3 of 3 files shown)


ipatests/test_integration/test_adtrust_install.py
```diff
@@ -216,3 +221,107 @@ def test_schema_compat_attribute(self):
...
+
+    def test_ipa_nis_manage_no_map_message(self):
+        """
+        This testcase checks that when ipa-nis-manage is enabled
+        and search for user is done on compat tree, there
+        is no message as 'map_unlock: old way MAP_MONITOR_DISABLED'
+        in dirsrv error logs
+        """
+        serverid = (realm_to_serverid(api.env.realm)).upper()
+        dirsrv_service = "dirsrv@%s.service" % serverid
```
mrizwan93, line 311: Please use from `ipapython.ipaldap import realm_to_serverid` here.

menonsudhir:
@mrizwan93 thank you for your review comments, all the requested changes was done.

mrizwan93:
`from ipapython.ipaldap import realm_to_serverid` is not available on `ipa-4-6`. Sorry for the inconvenience. You can revert it.

Please resolve these errors too:
```
************* Module ipatests.test_integration.test_adtrust_install

ipatests/test_integration/test_adtrust_install.py:248: [C0303(trailing-whitespace), ] Trailing whitespace)

ipatests/test_integration/test_adtrust_install.py:14: [E0611(no-name-in-module), ] No name 'realm_to_serverid' in module 'ipapython.ipaldap')

ipatests/test_integration/test_adtrust_install.py:245: [W0612(unused-variable), TestIpaNisManage.test_ipa_nis_manage_enable] Unused variable 'msg')

ipatests/test_integration/test_adtrust_install.py:307: [C0113(unneeded-not), TestIpaNisManage.test_ipa_nis_manage_no_map_message] Consider changing "not msg in cmd.stdout_text" to "msg not in cmd.stdout_text")

ipatests/test_integration/test_adtrust_install.py:298: [W0612(unused-variable), TestIpaNisManage.test_ipa_nis_manage_no_map_message] Unused variable 'result')

------------------------------------
```

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -216,3 +220,88 @@ def test_schema_compat_attribute(self):
...
+        console_msg = (
+            "Enabling plugin\n"
+            "This setting will not take effect until you restart Directory Server.\n"
+            "The rpcbind service may need to be started"
+        )
+        tasks.kinit_admin(self.master)
+        result = self.master.run_command(
+            ["ipa-nis-manage", "enable"],
+            stdin_text=self.master.config.admin_password,
+        )
```
mrizwan93, line 249: Would be a good idea to check status after enabling the plugin and see if it is enabled.
`ipa-nis-manage status`

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -216,3 +220,88 @@ def test_schema_compat_attribute(self):
...
+        command disable plugin on an IPA Master
+        """
+        serverid = realm_to_serverid(self.master.domain.realm)
+        dirsrv_service = "dirsrv@%s.service" % serverid
+        msg = "This setting will not take effect until you restart Directory Server."
+        tasks.kinit_admin(self.master)
+        result = self.master.run_command(
+            ["ipa-nis-manage", "disable"],
+            stdin_text=self.master.config.admin_password,
+        )
```
mrizwan93, line 267: Would be a good idea to check status after disabling the plugin and see if it is disabled.
`ipa-nis-manage status`

abbra:
nis plugin is not a schema-compat, these are two separate plugins. If you are looking for a lockup message, you should be using different prefix corresponding to the plugin.

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -216,3 +220,88 @@ def test_schema_compat_attribute(self):
...
+    def install(cls, mh):
+        tasks.install_master(cls.master)
+
+    def test_ipa_nis_manage_enable(self):
+        """
+        This testcase checks if ipa-nis-manage enable
+        command enables plugin on an IPA master
+        """
+        serverid = realm_to_serverid(self.master.domain.realm)
+        dirsrv_service = "dirsrv@%s.service" % serverid
```
mrizwan93, line 239: This can be moved to `install()` method.
```
        self.serverid = realm_to_serverid(self.master.domain.realm)
        self.dirsrv_service = "dirsrv@%s.service" % self.serverid
```
 By doing so you will not have to define it in each test. You can simply use `self.master.run_command(["systemctl", "restart", self.dirsrv_service])` further in testcases.

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -216,3 +220,88 @@ def test_schema_compat_attribute(self):
...
+        tasks.kinit_admin(self.master)
+        result = self.master.run_command(
+            ["ipa-nis-manage", "enable"],
+            stdin_text=self.master.config.admin_password,
+        )
+        assert result.returncode == 0
+        self.master.run_command(["systemctl", "restart", dirsrv_service])
+        time.sleep(DIRSRV_SLEEP)
+        instance = realm_to_serverid(self.master.domain.realm)
+        log_path = paths.SLAPD_INSTANCE_ERROR_LOG_TEMPLATE % instance
```
mrizwan93, line 305: After moving the said code to `install()`, you can use:
`log_path = paths.SLAPD_INSTANCE_ERROR_LOG_TEMPLATE % self.serverid`

menonsudhir:
> nis plugin is not a schema-compat, these are two separate plugins. If you are looking for a lockup message, you should be using different prefix corresponding to the plugin.

agreed that nis/schema-compat plugin are seperate, but the message "schema-compat - map_unlock: old way MAP_MONITOR_DISABLED"  was displayed only when the nis plugin was enabled/disabled on the IPA server.
Also the search for users/groups on compat tree did not log the lock message in dirsrv error log
As a result, ended up adding tests under ipa-nis-manage rather than for ipa-compat-manage.

abbra:
I see. The code in slapi-nis package actually uses the same implementation in both plugins but the code that reports a log message has this marker hard-coded, so it is always "schema-compat". However, the phrase you are looking for is not there anymore. Instead, it has "schema-compat - using former locking strategy". So I am not sure what this test is going to test as it will be missing the right message in newer versions all the time.

menonsudhir:
> I see. The code in slapi-nis package actually uses the same implementation in both plugins but the code that reports a log message has this marker hard-coded, so it is always "schema-compat". However, the phrase you are looking for is not there anymore. Instead, it has "schema-compat - using former locking strategy". So I am not sure what this test is going to test as it will be missing the right message in newer versions all the time.

The intention behind the test was to check that when nis plugin is enabled on system with newer slapi-nis versions we no more see the message "the schema-compat - map_unlock: old way MAP_MONITOR_DISABLED" in dirsrv error logs, which was the case earlier.

abbra:
I understand intent but it is not going to work because there is no such phrase in slapi-nis modules anymore. See https://pagure.io/slapi-nis/c/c7938c444f1c88e6f12ab9306fad558a0c25490e?branch=master -- this commit is in all slapi-nis updates in the current Fedora and RHEL versions.

Adding a test that tests something that does not exist anymore does not make a test valid.

menonsudhir:
> I understand intent but it is not going to work because there is no such phrase in slapi-nis modules anymore. See https://pagure.io/slapi-nis/c/c7938c444f1c88e6f12ab9306fad558a0c25490e?branch=master -- this commit is in all slapi-nis updates in the current Fedora and RHEL versions.
> 
> Adding a test that tests something that does not exist anymore does not make a test valid.
Ok. so are you suggesting to have this specific testcase removed or the entire PR to be dropped.
I will definitely go ahead and do what you suggest.

abbra:
if you want to add NIS test, please go ahead. However, the grep for this message is not needed.

abbra:
@menonsudhir should this be added to master too, not just ipa-4-6? I'd prefer if we'd have these tests everywhere.

ipatests/test_integration/test_adtrust_install.py
```diff
@@ -216,3 +220,79 @@ def test_schema_compat_attribute(self):
...
+        )
+        assert status_msg in result.stdout_text
+
+    def test_ipa_nis_manage_disable(self):
+        """
+        This testcase checks if ipa-nis-manage disable
+        command disable plugin on an IPA Master
+        """
+        serverid = realm_to_serverid(self.master.domain.realm)
+        dirsrv_service = "dirsrv@%s.service" % serverid
```
mrizwan93, line 267: These 2 lines are redundant of line 238, 239. Would be good idea to move this to method:
```
def get_dirsrv_id(realm):
    serverid = realm_to_serverid(realm)
    return("dirsrv@%s.service" % serverid)
```

menonsudhir: @mrizwan93 made the changes suggested.

menonsudhir:
> @menonsudhir should this be added to master too, not just ipa-4-6? I'd prefer if we'd have these tests everywhere.

@abbra, yes will surely have this tests added everywhere

rcritten:
If you want it in master then this patch needs to be rebased against master and use the flags to indicate what versions it needs to be backported to.

PR #4728, opened 2020-06-02
idoverrides: allow AD users to be members of IPA groups and roles

abbra:
Merge freeipa-adusers-admins plugin code from
https://github.com/abbra/freeipa-adusers-admins

This extension of ID overrides allows to add them to IPA groups and
roles to give them access to IPA management.

As result, Active Directory users can manage FreeIPA resources if these
groups are part of appropriate roles. For example, adding an Active
Directory user ID override as a member of 'admins' group would make it
equivalent to built-in FreeIPA 'admin' user.

The code works by allowing user ID overrides from the Default Trust View
in FreeIPA to be members of IdM groups and roles. User ID overrides in
the Default Trust View can only be created for trusted Active Directory
users. When Active Directory user authenticates with GSSAPI against the
FreeIPA LDAP server, its Kerberos principal is automatically mapped to
the user's ID override in the Default Trust View. LDAP server's access
control plugin uses membership information of the corresponding LDAP
entry to decide how access can be allowed.

Web UI part of this commit allows ID overridden users to see full
management console instead of a self-service. It also allows to add ID
overrides to the groups.

Fixes: https://pagure.io/freeipa/issue/7255
See also: https://bugzilla.redhat.com/show_bug.cgi?id=1651577

Signed-off-by: Alexander Bokovoy <abokovoy@redhat.com>

PR #4784, opened 2020-06-08
[Backport][ipa-4-8] idoverrides: allow AD users to be members of IPA groups and roles

rcritten:
Manual backport of PR of #4728 was pushed to master and backport to ipa-4-8 is required.

Adding ack automatically because the only reason the backport failed is due to the design document which exists only in current master.


rcritten:
ipa-4-8:

* 6abade3f8daed8dfa024936114209d19319c4f12 kdb: handle enterprise principal lookup in AS_REQ
* afe9191f99e034bcf52475b57996d81609de6837 support using trust-related operations in the server console
* 2ffb4fd18fceb509773951ce4f02aa0c5e2f851a idviews: handle unqualified ID override lookups from Web UI
* 8cce2bb31ab96f6ce6edba95f54575576f2b1a40 Support adding user ID overrides as group and role members
* 5e8df37e4cca155bf58aa4e61b9fa3f28eddd526 tests: account for ID overrides as members of groups and roles
* 99e613e478f7925d0f470a04d4de5a2f93385b7a Web UI: allow users from trusted Active Directory forest manage IPA
* 6b0f8f3617378da41ead8640e194e5b9415a38b1 ipatests: test that adding Active Directory user to a role makes it an administrator

PR #4786, opened 2020-06-08
baseuser: fix ipanthomedirectorydrive option name

pvoborni:
It should be ipanthomedirectorydrive and not ipanthomedirectoryrive.

This fixes showing the field in Web UI and also should fix CLI as it
probably never worked.


abbra:
LGTM. We don't need to rise API version here as this is really a typo fix.

pvoborni:
I wonder what are the next steps in this PR. It is approved by @abbra, but it doesn't have ACK label. Do we need another reviewer?

abbra:
I added an ACK.

rcritten:
master:

* 3912e8e6739058ef8fcb2964fda0edc3ee23fc1e baseuser: fix ipanthomedirectorydrive option name

PR #4787, opened 2020-06-08
webui: hide user attributes for SMB services section if empty

pvoborni:
This section should be hidded if user object hasn't ipantuserattrs
object class. I.e. when trusts are not enabled.

Web UI framework already supports hidding of sections if the
section contains no visible field. So to achieve it we simply needs
to hide the fields. Given that attributelevelrights
contains rights only for attributes of current object classes, all
of these are regarded as not writable.

We can leverage feature of input_widget that it gets hidden
when the attribute is not writable and has no value and widget's
"hidden_if_empty" is set to true. Thus doing it here.

For this to work, it is also required to fix an issue with
"ipanthomedirectorydrive" which is optional (in API) but Web UI
doesn't offer "empty" ("") value. Adding it here.

fixes: https://pagure.io/freeipa/issue/8336

### DIFF
--- a/install/ui/src/freeipa/user.js
+++ b/install/ui/src/freeipa/user.js
@@ -367,36 +367,39 @@ return {
                 {
                     name: 'smb_attributes',
                     label: '@i18n:objects.smb_attributes.title',
-                    show_cond: ['oc_ipantuserattrs'],
                     fields: [{
                             name: 'ipantlogonscript',
                             tooltip: {
                                 title: '@i18n:objects.smb_attributes.ipantlogonscript_tooltip'
-                            }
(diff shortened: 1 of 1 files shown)


pvoborni:
Related: https://github.com/freeipa/freeipa/pull/4786

abbra:
LGTM.

serg-cymbaluk:
I tested two cases without trust and with trust enabled. It works properly.

abbra:
master:

* f6707a71dcefd4b54d7909ad1ba9ba84f57c430a webui: hide user attributes for SMB services section if empty

PR #4789, opened 2020-06-08
Restructure Params validation to make it more flexible

Carbenium:
**This is a rough PoC to get input whether this idea should be developed further or not**

The current Param validation system is based on the the definition of `_rule_*` methods in the respective Param types which are activated by passing the respective kwarg to the Param constructor.

The current way is IMHO not very easily extendable to new validation rules (esp. by plugins). A prominent example would be the validation of email related fields (see https://pagure.io/freeipa/issue/3692). Basic validation can of course be performed by using the `pattern` rule, but there are cases where a simple regex is not enough. Adding a new validation rule by a plugin would require to subclass the appropriate Param class and define it's own `_rule_*` method there.

The basic idea behind this patchset is to revive and extend `ipavalidate` by introducing a common validation interface through the `Validator` class.
A new `validators` kwarg is introduced for the `Param` class which accepts:
  - a `Validator` instance
  - a list of `Validator` instances
 (- a dict `{'ctx1': [Validator(), ...], ...}`) 

If needed (idea came while trying to solve https://pagure.io/freeipa/issue/3226) the notion of validation contexts can be introduced. This would allow to specify different sets of validation rules during Param definition depending on the source of the validation request (e.g internal or external object). Whether this is needed is debatable.

For the WebUI this would mean that every `Validator` which should also be executed client-side will have to be implemented as well in the UI code. `Param` serialization has to adapted to serialize validators and their arguments correctly.

In general backwards compatibility for the `Param` API should be retained.


rcritten:
I'm not sure the scope of the change merits the requirements. I suspect we can implement the member change and e-mail domain changes without massive intervention.

The email domain has zero restrictions on it. There is no validator at all. We could likely re-use the DNS validators for this.

For adding/removing members it may be simpler to not run the pattern validators on them because either the user/group/member exists or they don't (minus external, which is taken on faith).

Carbenium:
Regarding the user validation: It will be solved probably in a different way anyway.
Regarding the mail validation: Agreed. If the goal is to solve only that one validation problem, then this isn't necessary.

The main advantage or rather my motivator is to provide a user/plugin-extendable validation system just like the `IPA.validator` WebUI counterpart.
I don't see an easy way to achieve that with the current system, although that might be due to my limited knowledge of the codebase.

Carbenium:
I have to revise the last sentence in my previous comment a bit.
User-extendability can be achieved using the `*rules` arg of `Params`, but there is currently no way to share some information about rules passed through that argument with the WebUI.

PR #4791, opened 2020-06-09
httpinstance: retry request without ipa-ca.$DOMAIN dnsName on failure

frasertweedale:
In the migration case of replica installation, if the CA server is an older
version it may not support the ipa-ca.$DOMAIN dnsName in the HTTP cert (it
is a special case in the cert_request command). Therefore if the request
fails, try it again without the ipa-ca.$DOMAIN dnsName.


abbra:
LGTM.

PR #4794, opened 2020-06-09
EPN: configure auth, starttls and SSL for tests, do live testing, add mailtest option

rcritten:
Wrap up the core of EPN functionality:

* minor fixups for starttls and SSL client support
* add live tests to verify deliver and template formatting using auth, starttls and SSL
* add --mail-test to test sending mail
* add smtp_delay option to limit velocity of e-mails


ipatests/test_integration/test_epn.py
```diff
@@ -239,7 +242,7 @@ def cleanupusers(self):
     @pytest.fixture
     def cleanupmail(self):
         """Cleanup any existing mail that has been sent."""
-        for i in self.notify_ttls:
+        for i in range(30):
```
fcami, line 249: Wouldn't max(self.notify_ttls) be more appropriate?

rcritten: That would work too I suppose and be more future proof but we don't twiddle with the configuration and that's all this is doing, ensuring that all old mail gets cleaned up.

rcritten: It's hardcoded there primarily because 30 users are created. This is to ensure that no stray mailboxes are left.

fcami: Yes, I understand. I wish it was not hardcoded, but it's a simple fix to do if ever we need it. Marking resolved.

fcami:
ACK from me. Please remove the temp commit and let gating run.

PR #4795, opened 2020-06-09
Fix named.conf update bug NAMED_DNSSEC_VALIDATION

tiran:
Commit a5cbdb57e50cfc62f61affda19ce878b2abd33de introduced a bug when
updating IPA from 4.8.6 to 4.8.7. NAMED_DNSSEC_VALIDATION template
variable was not declared.

Fixes: https://pagure.io/freeipa/issue/8363


ipaserver/install/server/upgrade.py
```diff
@@ -594,14 +594,26 @@ def named_add_ipa_ext_conf_file():
...
             paths.NAMED_CUSTOM_CFG_SRC,
             paths.NAMED_CUSTOM_CONFIG
         ),
         bindinstance.named_add_ext_conf_file(
             paths.NAMED_CUSTOM_OPTIONS_CFG_SRC,
-            paths.NAMED_CUSTOM_OPTIONS_CONFIG
+            paths.NAMED_CUSTOM_OPTIONS_CONFIG,
+            dict(
+                NAMED_DNSSEC_VALIDATION=dnssec_validation
+            )
```
flo-renaud, line 616: The upgrade will create the new file in /etc/named/ipa-options-ext.conf but it will leave the existing option dnssec-validation in /etc/named.conf.
I don't know how bind handles the same option defined in 2 different files but it may be confusing for the admin and lead to errors. The code should probably remove the existing option from /etc/named.conf.

tiran: There is no upgrade path from old to new configuration yet. I'm working on upgrade in another (not yet published) PR.

flo-renaud: There are also other issues:
- each time the ipa-server-upgrade tool is executed, a line with `include "/etc/named/ipa-ext.conf";` is added to /etc/ipa/named.conf. The sysupgrade state file contains a section for [named.conf] that can be leveraged.
- the line `include "/etc/named/ipa-options-ext.conf";` is added on clean install but not on upgrade

tiran: Thanks Flo! I didn't know about the first issue.
The second one is to be expected. There is no upgrade code yet.

ipatests/test_integration/test_upgrade.py
```diff
@@ -5,13 +5,96 @@
...
+
+include "$RFC1912_ZONES";
+include "$ROOT_KEY";
+
+/* WARNING: This part of the config file is IPA-managed.
+ * Modifications may break IPA setup or upgrades.
+ */
+dyndb "ipa" "$BIND_LDAP_SO" {
+        uri "ldapi://%2fvar%2frun%2fslapd-$SERVER_ID.socket";
+        base "cn=dns, $SUFFIX";
```
abbra, line 58: You removed space in the 'base' in the template. Do you want to remove it here as well?

abbra:
I think it is good enough on a first look. I have a test system where I'm going to try an upgrade.

PR #4796, opened 2020-06-09
Use DogTag's two step installation for CA only

fcami:
Some configuration items of Dogtag are only available using a 2-step
installation method described at:
https://www.dogtagpki.org/wiki/Two-Step_Installation

dogtaginstance.py: enable two-step Dogtag instance installation

cainstance.py: use Dogtag's two-step installation method
    
Use Dogtag's two-step installation method for CAs to be
able to customize some parameters like CA lifetime.
    
    
Related: https://pagure.io/freeipa/issue/7827
Fixes: https://pagure.io/freeipa/issue/8438

### DIFF
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -181,16 +181,23 @@ class DogtagInstance(service.Service):
         return os.path.exists(os.path.join(
             paths.VAR_LIB_PKI_TOMCAT_DIR, self.subsystem.lower()))
 
-    def spawn_instance(self, cfg_file, nolog_list=()):
+    def spawn_instance(
+        self, cfg_file, nolog_list=(),
+        skip_configuration=False, skip_installation=False
+    ):
         """
         Create and configure a new Dogtag instance using pkispawn.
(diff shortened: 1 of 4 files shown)


fcami:
2-step pkispawn: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/4280ddfe-aa8d-11ea-85d8-fa163e07b2f0/

ipaserver/install/dogtaginstance.py
```diff
@@ -181,24 +181,41 @@ def is_installed(self):
...
+                args = list(args)
+                args.append("--skip-configuration")
+                ipautil.run(args, nolog=nolog_list)
+            elif skip_installation is True:
+                args = list(args)
+                args.append("--skip-installation")
+                ipautil.run(args, nolog=nolog_list)
+            else:
+                args = list(args)
+                ipautil.run(args, nolog=nolog_list)
```
rcritten, line 218: Why make args a tuple only to convert it into a list, especially when it is **always** converted?

PR #4801, opened 2020-06-10
ipatests: Update the pki-master-f32 image version

flo-renaud:
There is a new Vagrant image for pki-master-f32, that contains
jss 4.7.0-0 instead of jss 4.7.0-1.
This change is required because the copr repo @pki/master initially
provided 4.7.0-1 but went backwards in the version number, and
critical fixes are available in 4.7.0-0.
    
Without this change, the vagrant image is using 4.7.0-1 and tries to
update (not downgrade), hence does not install the most recent version
with the fixes.

### DIFF
--- a/ipatests/prci_definitions/nightly_latest_pki.yaml
+++ b/ipatests/prci_definitions/nightly_latest_pki.yaml
@@ -35,7 +35,7 @@ jobs:
         git_refspec: '{git_refspec}'
         template: &pki-master-latest
           name: freeipa/pki-master-f32
-          version: 0.0.1
+          version: 0.0.2
         timeout: 1800
         topology: *build
 


frasertweedale:
Are the Azure pipelines failures expected?

abbra:
Fedora infra works with reduced capacity, failures in resolving repositories and container registries are expected. I re-started the jobs.

frasertweedale:
@flo-renaud we can ignore the healthcheck error count for now.  There is a plan in motion to address this issue but it will require a new release of *freeipa-healthcheck*.  See https://github.com/freeipa/freeipa-healthcheck/pull/128; when this has landed we can specify to only run tests in the *ipahealthcheck* module namespace and ignore the *pki* namespace.

flo-renaud:
The run with the temp commit can be seen [here](https://github.com/freeipa/freeipa/commit/7e2ce10618110f58a79705764f3cb7e6b71364f5), with [ipa-healthcheck](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/8af1b256-ab33-11ea-8427-fa163e2b3f63/report.html) failure that can safely be ignored.
Temp commit removed.

flo-renaud:
master:

* c358f8dbb5b2532f5f1e9f9ff823bd3c069b6e6e ipatests: Update the pki-master-f32 image version

PR #4808, opened 2020-06-11
Move ipa-epn systemd files and run RPM hooks

tiran:
The init/systemd directory is for server only and not part of
CLIENT_ONLY builds.

It's necesary to run pre/post installation hooks to make systemd aware
of new files.


rcritten:
@fcami FYI

abbra:
LGTM. @fcami please review.

rcritten:
Tested and this works fine.

I had attempted to avoid creating a client-specific directory for systemd but alas there is no other way without creating an ugly mess.

fcami:
Tested, works. ACK.

fcami:
master:

* a18d406b5631f80134e202ae3310580683937f92 Move ipa-epn systemd files and run RPM hooks

PR #4810, opened 2020-06-11
fix iPAddress cert issuance for >1 host/service

frasertweedale:
The 'cert_request' command accumulates DNS names from the CSR,
before checking that all IP addresses in the CSR are reachable from
those DNS names.  Before adding a DNS name to the set, we check that
that it corresponds to the FQDN of a known host/service principal
(including principal aliases).  When a DNS name maps to a
"alternative" principal (i.e.  not the one given via the 'principal'
argument), this check was not being performed correctly.
Specifically, we were looking for the 'krbprincipalname' field on
the RPC response object directly, instead of its 'result' field.

To resolve the issue, dereference the RPC response to its 'result'
field before invoking the '_dns_name_matches_principal' subroutine.

Fixes: https://pagure.io/freeipa/issue/8368

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -827,13 +827,13 @@ class cert_request(Create, BaseCertMethod, VirtualCommand):
                 try:
                     if principal_type == HOST:
                         alt_principal_obj = api.Command['host_show'](
-                            name, all=True)
+                            name, all=True)['result']
                     elif principal_type == KRBTGT:
                         alt_principal = kerberos.Principal(
                             (u'host', name), principal.realm)
                     elif principal_type == SERVICE:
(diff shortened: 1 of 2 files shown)


frasertweedale:
RHEL 8 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1846352

RHEL 7 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1846349

Customer issue was reported against RHEL 7, but I suppose we should wait for ACKs before backporting.  It's a fairly obscure use case.

frasertweedale:
master:

* 68ada5f2040bb207a5402220fb93e360bae215fe (HEAD) fix iPAddress cert issuance for >1 host/service

frasertweedale:
Thanks for the review @rcritten.

PR #4818, opened 2020-06-15
Populate nshardwareplatform and nsosversion during join operation

Carbenium:
Fixes: https://pagure.io/freeipa/issue/8370

### DIFF
--- a/ipaserver/plugins/host.py
+++ b/ipaserver/plugins/host.py
@@ -361,7 +361,9 @@ class host(LDAPObject):
         },
         'System: Enroll a Host': {
             'ipapermright': {'write'},
-            'ipapermdefaultattr': {'objectclass', 'enrolledby'},
+            'ipapermdefaultattr': {
+                'objectclass', 'enrolledby', 'nshardwareplatform', 'nsosversion'
+            },
             'replaces': [
                 '(targetattr = "objectclass")(target = "ldap:///fqdn=*,cn=computers,cn=accounts,$SUFFIX")(version 3.0;acl "permission:Enroll a host";allow (write) groupdn = "ldap:///cn=Enroll a host,cn=permissions,cn=pbac,$SUFFIX";)',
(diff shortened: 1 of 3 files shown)


rcritten:
These attributes need to be writeable for the enrollment role.

ipaserver/plugins/host.py
```diff
@@ -361,7 +361,8 @@ class host(LDAPObject):
         },
         'System: Enroll a Host': {
             'ipapermright': {'write'},
-            'ipapermdefaultattr': {'objectclass', 'enrolledby'},
+            'ipapermdefaultattr': {'objectclass', 'enrolledby',
+                                   'nshardwareplatform', 'nsosversion'},
```
tiran, line 365:
```suggestion
            'ipapermdefaultattr': {
                'objectclass', 'enrolledby', 'nshardwareplatform', 'nsosversion'
            },
```

ipaserver/plugins/join.py
```diff
@@ -85,7 +85,7 @@ class join(Command):
     has_output = tuple()
     use_output_validation = False
 
-    def execute(self, hostname, **kw):
+    def execute(self, hostname, **kwargs):
```
tiran, line 88: There is nothing wrong with ``kw`` here.

ipaserver/plugins/join.py
```diff
@@ -94,13 +94,21 @@ def execute(self, hostname, **kw):
...
+        assert 'cn' not in kwargs
         ldap = self.api.Backend.ldap2
 
+        try:
+            # We don't use the realm parameter at the moment. Since we pass
+            # kwargs directly to host_{add,mod} and 'realm' is not a supported
+            # parameter, we remove it.
+            del kwargs['realm']
+        except KeyError:
+            pass
```
tiran, line 106:
```suggestion
        # realm parameter is not supported by host_{add,mod}
        kwargs.pop("realm", None)
```

Carbenium: I even considered this one :wink:

ipaserver/plugins/join.py
Carbenium: True that.

abbra:
LGTM

abbra:
master:

* 8f640f8672f2bec7b560a2838e2451a190abf6cf Populate nshardwareplatform and nsosversion during join operation

PR #4819, opened 2020-06-15
Use 256 bit AJP secret 

cipherboy:
This PR introduces two commits:

 - A documentation / cleanup commit that changes to xpath query to search for AJP adapters, not ports.
 - Switch IPA to use a 256-bit AJP secret. IPA seems to like 256-bit passwords. I don't believe PKI's default password is 256-bits.

I believe @abbra's earlier commit handled the upgrade path correctly. However, this one ensures that _new_ `ipa-server-install` passwords are 256-bits by default. Is this a concern we care about? Do we care about users setting their own password?

Also, this isn't strictly required IMO; it is a nice enhancement though. If we don't want to, we don't need to backport to RHEL 8.3. 

Fixes: https://pagure.io/freeipa/issue/8372

### DIFF
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -840,7 +840,9 @@ class PKIIniLoader:
             pki_subsystem_type=subsystem.lower(),
             home_dir=os.path.expanduser("~"),
             # for softhsm2 testing
-            softhsm2_so=paths.LIBSOFTHSM2_SO
+            softhsm2_so=paths.LIBSOFTHSM2_SO,
+            # Configure a more secure AJP password by default
+            ipa_ajp_secret=ipautil.ipa_generate_password(special=None)
         )
 
     @classmethod
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -308,11 +308,12 @@ class DogtagInstance(service.Service):
         doc = server_xml.getroot()
 
         # no AJP connector means no need to update anything
-        connectors = doc.xpath('//Connector[@port="8009"]')
+        connectors = doc.xpath('//Connector[@protocol="AJP/1.3"]')
         if len(connectors) == 0:
             return
 
-        # AJP connector is set on port 8009. Use non-greedy search to find it
+        # AJP protocol is at version 1.3. Assume there is only one as
+        # Dogtag only provisions one.
         connector = connectors[0]
 
         # Detect tomcat version and choose the right option name
@@ -331,11 +332,24 @@ class DogtagInstance(service.Service):
             rewrite = False
         else:
             if oldattr in connector.attrib:
+                # Sufficiently new Dogtag versions (10.9.0-a2) handle the
+                # upgrade for us; we need only to ensure that we're not both
+                # attempting to upgrade server.xml at the same time.
+                # Hopefully this is guaranteed for us.
                 self.ajp_secret = connector.attrib[oldattr]
                 connector.attrib[secretattr] = self.ajp_secret
                 del connector.attrib[oldattr]
             else:
-                # Generate password, don't use special chars to not break XML
+                # Generate password, don't use special chars to not break XML.
+                #
+                # If we hit this case, pkispawn was run on an older Dogtag
+                # version and we're stuck migrating, choosing a password
+                # ourselves. Dogtag can't generate one randomly because a
+                # Dogtag administrator might've configured AJP and might
+                # not be using IPA.
+                #
+                # Newer Dogtag versions will generate a random password
+                # during pkispawn.
                 self.ajp_secret = ipautil.ipa_generate_password(special=None)
                 connector.attrib[secretattr] = self.ajp_secret
 
--- a/install/share/ipaca_customize.ini
+++ b/install/share/ipaca_customize.ini
@@ -12,6 +12,7 @@
 #
 # Predefined variables
 #  - ipa_ca_subject
+#  - ipa_ajp_secret
 #  - ipa_fqdn
 #  - ipa_subject_base
 #  - pki_admin_password
--- a/install/share/ipaca_default.ini
+++ b/install/share/ipaca_default.ini
@@ -12,6 +12,7 @@ ipa_ca_pem_file=/etc/ipa/ca.crt
 
 ## dynamic values
 # ipa_ca_subject=
+# ipa_ajp_secret=
 # ipa_subject_base=
 # ipa_fqdn=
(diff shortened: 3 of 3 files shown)


rcritten:
Is it important to reconfigure existing instances on upgrade to use this as well?

cipherboy:
If there's no secret on an existing install, @abbra's existing changes ensure IPA provides a 256-bit secret. This is one upgrade scenario.

However...

On new installs with Dogtag 10.9 (RHEL 8.3 and Fedora 31-rawhide currently in `updates-testing`...), without this patch, PKI provides a [`log2(76^12) =~ 75`](https://github.com/dogtagpki/pki/blob/737fc097afb42c0b48fe362e970591099dfed2c7/base/common/python/pki/__init__.py#L217-L218)-bit secret for IPA. Because @abbra wanted PKI to fix it and didn't want more code in IPA, we used our existing password generation code. This is fine, we did it correctly so IPA can now set a password in the configuration when calling `pkispawn`.

If IPA cares about 256-bit password strength, PKI has provided a mechanism for IPA to set it for both of us, on new installs. It involves changing only a couple of lines in IPA's installer; that's what the last commit is. IPA will continue to handle upgrades.

So you can either:
 - Ignore this patch and close the PR. Perhaps 256-bit AJP password doesn't matter for IPA.
 - Decide you are fine shipping RHEL 8.3 with a 75-bit password and only merge this to master. According to what Mozilla uses, this is FIPS compliant enough for NSS. 
 - Decide you want to fix this in 8.3 too, and backport this change (probably: last commit only) to RHEL 8.3 branch.

Only in the middle case would you need to write an upgrade script to fix our 75-bit password in RHEL 8.4. But if you backport it now, as a bugfix and patch RHEL 8.3, you don't need to write said upgrade script because RHEL 8.3 hasn't shipped yet and RHEL 8.2 didn't have this issue. Any _new_ upgrade script would thus only be run for new installs for servers on RHEL 8.3, when they upgrade to RHEL 8.4+.

I opened this PR because I can't answer the question of "does IPA necessarily require a 256-bit AJP password?" -- that's an IPA question. :) You tell me if you need an upgrade script. :)

rcritten:
Thanks for the background, this PR just came out of the blue without an associated ticket outlining the reasoning and the first patch has a one-line commit message. 

From what I can tell this is related to commit ec73de969f55b7a005b6401029f87fe6a225a417 and was done due to a change in tomcat, https://tomcat.apache.org/migration-9.html#Upgrading_9.0.x

There are people with installs that originated in F20 so we have to be careful with them. My F25 install still has the port 8009 connector. I'll leave it to @abbra to evaluate whether both need to be handled since he knows this code better.

cipherboy:
"My F25 install still has the port 8009 connector." -- I'm not sure I follow this. Tomcat isn't port based; it is protocol based. Your 8009 port connector still aught to say AJP protocol... Is that assumption not correct? 

The first commit is just a documentation and clean up commit (assuming above assumption holds). Does IPA require BZs to clarify in-code documentation? I can file one.

The second commit is the real reason this PR was updated. The associated BZ is [rh-bz#1845447](https://bugzilla.redhat.com/show_bug.cgi?id=1845447). This is a `pki-core` BZ because @abbra wanted PKI to fix it. We did. But you're getting a PKI password, not an IPA password. I can expand the commit message if you'd like to include this additional context. :)

Thanks @rcritten for looking at this!

abbra:
@cipherboy thank you for the pull request. The code changes look OK. We only used 256-bit because that's the default password strength in our generator. Anything that is good enough for FIPS is good enough here.

Please do the following:
- add a ticket at https://pagure.io/freeipa/new_issue that has the explanation of a problem
- reference the ticket in the commit messages with 'Fixes: url' line

I'll kick full re-run of PR CI once this is done.

cipherboy:
@abbra Ok, updated. Issue linked in second commit message: https://pagure.io/freeipa/issue/8372

cipherboy:
To summarize some discussion with @abbra on this topic:

 - AJP/1.3 is a fully plaintext protocol
 - The secret is sent from client -> server on the wire; thus it is important that it is only done over a `localhost` boundary.
 - This is sent as an "attribute" in plaintext
    - spec: https://tomcat.apache.org/connectors-doc-archive/jk2/common/AJPv13.html
    - Python Impl: https://github.com/cipherboy/AJPy/blob/master/ajpy/ajp.py#L200
    - Tomcat Java impl: https://github.com/apache/tomcat/blob/master/java/org/apache/coyote/ajp/AjpProcessor.java#L857-L866

Any security derived here comes from:

 1) Secret not being brute-forcable,
 2) Secret not being interceptable (i.e., localhost needs to be protected commmunication, no pcaps).

It'd be cool to get this going over a local (Unix, not TCP) socket, but Java lacks JDK support for those. We'd have to provide support for that elsewhere and see if we can get Tomcat to work with it.

ipaserver/install/dogtaginstance.py
```diff
@@ -308,11 +308,12 @@ def secure_ajp_connector(self):
         doc = server_xml.getroot()
 
         # no AJP connector means no need to update anything
-        connectors = doc.xpath('//Connector[@port="8009"]')
+        connectors = doc.xpath('//Connector[@protocol="AJP/1.3"]')
```
rcritten, line 311: Please explain why you are changing the search criteria. I assume this is because it is more specific to this version of AJP and relying only on the port could lead to future problems but this is only a wild guess.

cipherboy: AJP is a protocol a connector can have. Specifically it has `AJP/1.3` in the `protocol=` field. Most commonly it listens on port 8009. However, you can run AJP on any port you desire. The comment above this block of code says "AJP connector", so we should update the AJP connector, not whatever is listening on port 8009.

rcritten: Yes, exactly! It should be in the commit message.

cipherboy: ACK, adding.

cipherboy:
@abbra / @tiran: I've filed https://bugzilla.redhat.com/show_bug.cgi?id=1849146 for the IPA component in 8.3 ; feel free to approve or deny w.r.t. IPA 4.8 backport. Lemme know early next week if this needs anything else :)

abbra:
LGTM.

tiran:
master:

* c5e9bd61d6a0a8fc8f06592b9cf14bd576404d51 Clarify AJP connector creation process
* 3ecea7800ab7d44f47ef26264dfb8c33aadeca49 Configure PKI AJP Secret with 256-bit secret

PR #4821, opened 2020-06-16
fix cert-find errors in CA-less deployment

frasertweedale:
Under some search conditions (in particular, when user is
specified), the CA sub-search of cert-find command throws an error
on CA-less deployments.  Do not execute the CA sub-search on CA-less
deployments.

Fixes: https://pagure.io/freeipa/issue/8369


rcritten:
master:

* 19544d53aeada733eb63d596be0a8576b17b9d04 fix cert-find errors in CA-less deployment

PR #4831, opened 2020-06-17
WebUI: Disable certificate loading for CA-less installation

serg-cymbaluk:
- prevent parsing data in certs_field if certificates are disabled
- fix the corresponding issue on user, host and service details pages

Ticket: https://pagure.io/freeipa/issue/8203

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
--- a/install/ui/src/freeipa/field.js
+++ b/install/ui/src/freeipa/field.js
@@ -1392,6 +1392,8 @@ field.certs_field = IPA.certs_field = function(spec) {
     that.acl_result_index = spec.acl_result_index;
 
     that.load = function(data) {
+        if (!IPA.cert.is_enabled()) return;
+
         var value = that.adapter.load(data);
         var parsed = util.parse(that.data_parser, value, "Parse error:"+that.name);
         value = parsed.value;
--- a/install/ui/src/freeipa/host.js
+++ b/install/ui/src/freeipa/host.js
@@ -520,18 +520,20 @@ IPA.host.details_facet = function(spec, no_init) {
         var host_command = that.details_facet_create_refresh_command();
         batch.add_command(host_command);
 
-        var certificates = rpc.command({
-            entity: 'cert',
-            method: 'find',
-            retry: false,
-            options: {
-                host: [ pkey ],
-                sizelimit: 0,
-                all: true
-            }
-        });
+        if (IPA.cert.is_enabled()) {
+            var certificates = rpc.command({
+                entity: 'cert',
+                method: 'find',
+                retry: false,
+                options: {
+                    host: [ pkey ],
+                    sizelimit: 0,
+                    all: true
+                }
+            });
 
-        batch.add_command(certificates);
+            batch.add_command(certificates);
+        }
 
         return batch;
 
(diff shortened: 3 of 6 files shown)


rcritten:
My only hesitation with this is that it is perfectly fine to have a CA-less installation and have user certificates provided by another CA. This will hide the certificates altogether. What about certificates stored in the user/host/service entry directly?

flo-renaud:
I agree with Rob, the certs can be stored directly in the entity's entry even in CA-less env. The correct fix would be to retrieve the cert from the entity's entry (ipa  user-show / host-show / service-show will return the cert in base64).

serg-cymbaluk:
@rcritten @flo-renaud 

I tested the solution with using base64 certificate representation. But there are a couple of issues with that:
1) I can't show any certificate attributes on the page because decoding the raw data on browser side is quite tricky.
It makes impossible to recognize certificates visually.
2) There is an error when you are trying to remove a certificate: 'CA is not configured'
Meanwhile the command removes the certificate successfully.
Reproducible even in CLI.

pvoborni:
Quite a while ago,  `cert-find` (I think) was enhanced to be able to parse arbitrary cert and return the data to Web UI so that Web UI doesn't need to do the parsing.  We needs something like that to work even in CA-less mode. 

Maybe the real fix is to modify cert-find to return the right data in CA-less mode. I wonder if this never worked, or there was some behavior change.

rcritten:
cert-find for CA-less was fixed in 32c64a78cef499c05e4ffa71d033692571297dca

This only shows the CA chain though.

The service certs are available directly in their respective records, HTTP/fqdn and ldap/fqdn and a show will include the relevant information:

```
$ ipa service-show HTTP/`hostname`
  Principal name: HTTP/ipa.example.test@EXAMPLE.TEST
  Principal alias: HTTP/ipa.example.test@EXAMPLE.TEST
  Certificate: MIIE ... 9hhw=
  Subject: CN=ipa.example.test,O=Certificate Shack Ltd
  Serial Number: 4096
  Serial Number (hex): 0x1000
  Issuer: CN=Certificate Shack Intermediate CA,O=Certificate Shack Ltd
  Not Before: Fri Jun 12 22:27:50 2020 UTC
  Not After: Tue Jun 22 22:27:50 2021 UTC
  Fingerprint (SHA1): fc:3b:42:08:32:36:f7:0a:c8:43:4d:a0:7a:41:6f:9b:8f:cf:b3:5a
  Fingerprint (SHA256): 1c:9e:a3:19:53:63:cb:da:92:8b:1c:e7:6d:b3:93:fe:1b:ac:4f:71:89:64:46:2f:db:e9:4f:d4:96:19:eb:95
  Keytab: True
  Managed by: ipa.example.test
```

serg-cymbaluk:
@rcritten hi,

Many thanks for the link.
I've tested the patch and it works for me. Given that I'm closing the PR. There are no changes needed on UI side.

pvoborni:
I don't understand one thing. Do we have an issue in showing the info in Web UI or not? The original issue this PR was about  was reported 5 months ago with ipa-4-8. But the patch 32c64a7 was done 3 years ago in ipa-4-5+. So if 32c64a7 is the fix then we should not have the error. So what was reported? Is it possible that there is some different issue?

flo-renaud:
@pvoborni
There was an additional fix done for https://pagure.io/freeipa/issue/8369.
- 19544d53aeada733eb63d596be0a8576b17b9d04 (on the master branch, included in release rc_4-9-0-1)
- 60a58eac02ed2c0741fbce3c1fb3acceef61fdbc (on ipa-4-8 branch, included in releases release-4-8-10 and release-4-8-9)
- eb8cb4d941ed8d4d76772a7a7d65c018497cb615 (on ipa-4-6 branch, not included in any official release).

PR #4834, opened 2020-06-17
Use old uglifyjs on RHEL 8

tiran:
RHEL 8 buildroot does not have python3-rjsmin yet. Fall back to
uglifyjs.

See: https://pagure.io/freeipa/issue/8300


tiran:
The PR solves one issue that prevents me to build IPA RPMs on RHEL from the upstream spec file. That's useful to verify a change quickly without going through a full build with IPA module spec.

abbra:
LGTM.

flo-renaud:
master:

* 6e3346f0a7d820cb3643044884657fa60a80f918 Use old uglifyjs on RHEL 8

PR #4836, opened 2020-06-18
Enable CA nonce protection

cipherboy:
Disabling CA nonce protection prohibits Dogtag CA from performing CSRF
protection, making some XSS vulnerabilities trivially reflected. We need
to ensure this is enabled by default to limit the scope of these types
of vulnerabilities in the future.

Fixes: https://pagure.io/freeipa/issue/8376

`Signed-off-by: Alexander Scheel <ascheel@redhat.com>`

----

@tiran -- I'm not on the approved contributors list. Mind kicking off PR-CI for me?

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -408,7 +408,6 @@ class CAInstance(DogtagInstance):
             self.step("stopping certificate server instance to update CS.cfg",
                       self.stop_instance)
             self.step("backing up CS.cfg", self.safe_backup_config)
-            self.step("disabling nonces", self.__disable_nonce)
             self.step("set up CRL publishing", self.__enable_crl_publish)
             self.step("enable PKIX certificate path discovery and validation",
                       self.enable_pkix)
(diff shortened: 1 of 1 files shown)


cipherboy:
OK so this approach isn't viable right now. We'd need to update certmonger to read and respect nonces (and eventually Tomcat CSRF tokens) before this'd work.

@rcritten -- we're thinking about enabling Tomcat's default CSRF protection profile in the next release. Is this something you'd have cycles to work on, or would you prefer someone else take a crack at it? 

/cc @edewata and @SilleBille

rcritten:
I'd need to know a bit more about how the protection is implemented but sure, it's possible depending on the timeline.

duzda:
Hi @cipherboy, is this still planned or can be safely closed?

cipherboy:
@duzda I haven't worked on the project since 2021, but a quick code search suggests yes: https://github.com/freeipa/freeipa/blob/9054f0d819fd38f6e8b50dfcdbd6981c8e3c45bf/ipaserver/install/cainstance.py#L860

duzda:
Thanks for the swift reply, I felt better asking than just going and closing everything :)

cipherboy:
@duzda Apologies, I meant that the work still needed to be done theoretically.

PR #4838, opened 2020-06-19
Test that override_gid is working for subdomain

amore17:
When override_gid is set in sssd.conf in IPA domain section
Then it should also work for subdomain.  
Related: https://pagure.io/SSSD/sssd/issue/4061

### DIFF
--- a/ipatests/test_integration/test_sssd.py
+++ b/ipatests/test_integration/test_sssd.py
@@ -481,6 +481,34 @@ class TestSSSDWithAdTrust(IntegrationTest):
         for text in verify_in_stdout:
             assert text in second_res.stdout_text
 
+    @contextmanager
+    def override_gid_setup(self, gid):
+        sssd_conf_backup = tasks.FileBackup(self.master, paths.SSSD_CONF)
+        try:
+            with tasks.remote_sssd_config(self.master) as sssd_conf:
+                sssd_conf.edit_domain(self.master.domain,
+                                      'override_gid', gid)
+            tasks.clear_sssd_cache(self.master)
+            yield
+        finally:
+            sssd_conf_backup.restore()
+            tasks.clear_sssd_cache(self.master)
+
+    def test_override_gid_subdomain(self):
+        """Test that override_gid is working for subdomain
+
+        This is a regression test for sssd bug:
+        https://pagure.io/SSSD/sssd/issue/4061
+        """
+        tasks.clear_sssd_cache(self.master)
+        user = self.users['child_ad']['name']
+        gid = 10264
+        # verify the user can be retrieved initially
+        self.master.run_command(['id', user])
+        with self.override_gid_setup(gid):
+            test_gid = self.master.run_command(['id', user])
+            assert 'gid={id}'.format(id=gid) in test_gid.stdout_text
+
 
 class TestNestedMembers(IntegrationTest):
     num_clients = 1
--- a/ipatests/test_integration/test_sssd.py
+++ b/ipatests/test_integration/test_sssd.py
@@ -471,15 +471,19 @@ class TestSSSDWithAdTrust(IntegrationTest):
             "--matchrule='<ISSUER>{}'".format(cert_subject),
             "--domain={}".format(self.master.domain.name)
         ])
-        tasks.clear_sssd_cache(self.master)
-
-        # verify the user can be retrieved after the certmaprule is added
-        second_res = self.master.run_command(['id', self.users['ad']['name']])
-
-        assert first_res.stdout_text == second_res.stdout_text
-        verify_in_stdout = ['gid', 'uid', 'groups', self.users['ad']['name']]
-        for text in verify_in_stdout:
-            assert text in second_res.stdout_text
+        try:
+            tasks.clear_sssd_cache(self.master)
+            # verify the user can be retrieved after the certmaprule is added
+            second_res = self.master.run_command(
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @amore17 
I believe the test failure is linked to the previous test `test_subdomain_lookup_with_certmaprule_containing_dn` which does not perform cleanup and leaves a certmap-rule.
You can either move your test before `test_subdomain_lookup_with_certmaprule_containing_dn` or fix by adding cleanup code that removes the certmap-rule.

flo-renaud:
Hi @amore17 
there are pycodestyle issues:
```
pycodestyle
-----------
./ipatests/test_integration/test_sssd.py:445:81: E501 line too long (82 > 80 characters)
./ipatests/test_integration/test_sssd.py:447:81: E501 line too long (81 > 80 characters)
```

ipatests/test_integration/test_sssd.py
```diff
@@ -75,6 +75,14 @@ def install(cls, mh):
...
             cls.users['child_ad']['name_tmpl'].format(
                 domain=cls.child_ad.domain.name))
+
+        # workaround for https://pagure.io/freeipa/issue/8294
+        # enable AES128 and AES256 encryption on AD
+        cmd = ("ksetup /setenctypeattr {0} AES128-CTS-HMAC-SHA1-96 "
+               "AES256-CTS-HMAC-SHA1-96")
+        cls.child_ad.run_command(cmd.format(cls.ad.domain.name))
+        cls.ad.run_command(cmd.format(cls.child_ad.domain.name))
+
```
flo-renaud, line 85: These steps will probably be needed by other tests using ad domain + child domain, I would put them in the tasks module.
The commit message can also be changed from "Related: http..." to "Fixes: http..."

ipatests/test_integration/test_sssd.py
```diff
@@ -432,15 +432,47 @@ def test_subdomain_lookup_with_certmaprule_containing_dn(self):
...
+            second_res = self.master.run_command(
+                ['id', self.users['ad']['name']])
+            assert first_res.stdout_text == second_res.stdout_text
+            verify_in_stdout = ['gid', 'uid', 'groups',
+                                self.users['ad']['name']]
+            for text in verify_in_stdout:
+                assert text in second_res.stdout_text
+        finally:
+            self.master.run_command(
+                ['ipa', 'certmaprule-del', "'{}'".format(cert_subject)])
```
flo-renaud, line 455: Can you extract the fix for this test `test_subdomain_lookup_with_certmaprule_containing_dn` in a separate commit and refer to https://pagure.io/SSSD/sssd/issue/3721 ?

flo-renaud:
Hi @amore17 
LGTM, you can remove the temp commit.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -2297,3 +2297,14 @@ def get_sssd_version(host):
     """Get sssd version on remote host."""
     version = host.run_command('sssd --version').stdout_text.strip()
     return parse_version(version)
+
+
+def enable_encryption(host, domain_name):
```
kaleemsiddiqu, line 2302: @amore17 , @flo-renaud
Can we add ansible play for this in https://github.com/freeipa/freeipa-pr-ci/tree/master/ansible/roles/windows/common/tasks to enable AES on ad vms ?
This way we can avoid calling this every test suite

flo-renaud: The step is needed only when there is a subdomain, in this case https://github.com/freeipa/freeipa-pr-ci/blob/master/ansible/roles/windows/ad-subdomain/tasks/main.yml would be a better fit IMO.

kaleemsiddiqu: I think we need to enable AES krb5 communication on every AD vms as it has affected kinit with every, so we should add in common task.  We have added this on every AD vms in downstream.

amore17: Added : https://github.com/freeipa/freeipa-pr-ci/pull/380

ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-latest/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
+        test_suite: test_integration/test_sssd.py::TestSSSDWithAdTrust
         template: *ci-master-latest
         timeout: 3600
         topology: *master_1repl_1client
```
netoarmando, line 74: @amore17 `test_sssd.py` requires `adroot_adchild_adtree_master_1client` as the topology.

flo-renaud:
Hi @amore17 
thanks for the PR, LGTM. Please remove the temp commit and I will ACK.

fcami:
master:

* 0dfb44c3c793623686258794c3c8fc8ec6216e29 ipatests : Test to verify override_gid works with subdomain.
* 4247fb9c73e940f459ab62dcd32157e6148c319e ipatests: xfail test with older versions of sssd
* ea7b8d66536e388152619c99f8cc83e3fd7e92b8 ipatests: cleanup in test_subdomain_lookup_with_certmaprule_containing_dn

fcami:
hi @amore17 both automatic backports failed, could you open backport PRs to ipa-4-6 and ipa-4-8 please?

PR #4839, opened 2020-06-19
ipa-client-install: use sshd drop-in configuration

flo-renaud:
### ipa-client-install: use sshd drop-in configuration

sshd 8.2+ now supports the "Include" keyword in sshd_config and
ships by default /etc/ssh/sshd_config with
"Include /etc/ssh/sshd_config.d/*"

As fedora 32 provides a config file in that directory (05-redhat.conf) with
ChallengeResponseAuthentication no
that is conflicting with IPA client config, ipa-client-install now needs
to make its config changes in a drop-in file read before 05-redhat.conf
(the files are read in lexicographic order and the first setting wins).

There is no need to handle upgrades from sshd < 8.2: if openssh-server
detects a customisation in /etc/ssh/sshd_config, it will not update
the file but create /etc/ssh/sshd_config.rpmnew and ask the admin
to manually handle the config upgrade.
    
Fixes: https://pagure.io/freeipa/issue/8304

### client install: fix broken sshd config

If ipa client was installed with openssh-server >= 8.2, the
configuration parameters for sshd were put in /etc/ssh/sshd_config
instead of in a snippet in /etc/ssh/sshd_config.d.
Upgrade to this new ipa version fixes the sshd conf by
moving the params to the snippet.

Related: https://pagure.io/freeipa/issue/8304

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1069,7 +1069,7 @@ fi
 %endif
 
 
-%triggerin client -- openssh-server
+%triggerin client -- openssh-server < 8.2
 # Has the client been configured?
 restore=0
 test -f '/var/lib/ipa-client/sysrestore/sysrestore.index' && restore=$(wc -l '/var/lib/ipa-client/sysrestore/sysrestore.index' | awk '{print $1}')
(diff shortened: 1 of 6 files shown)


abbra:
LGTM. Please remove the temporary commit.

flo-renaud:
@abbra 
thanks for the review. Temp commit removed and tests are green.

flo-renaud:
master:

* 3cf9979aec109e01ee470c22fbc7ca55bd8c2eb0 ipa-client-install: use sshd drop-in configuration
* 511f5194dcf12a64fc8a6c2791edc9c6cda2d926 client install: fix broken sshd config

PR #4840, opened 2020-06-20
handle Y2038 in timestamp to datetime conversions

abbra:
According to datetime.utcfromtimestamp() method documentation[1],
this and similar methods fail for dates past 2038 and can be replaced by
the following expression on the POSIX compliant systems:

  datetime(1970, 1, 1, tzinfo=timezone.utc) + timedelta(seconds=timestamp)

Make sure to use a method that at least allows to import the timestamps
properly to datetime objects on 32-bit platforms.

[1] https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp

Fixes: to-be-filed

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -1647,3 +1647,27 @@ def rmtree(path):
             shutil.rmtree(path)
     except Exception as e:
         logger.error('Error removing %s: %s', path, str(e))
+
+
+def datetime_from_utctimestamp(t, units=1):
+    """
+    Convert a timestamp or a time.struct_time to a datetime.datetime
+    object down to seconds, with UTC timezone
+
+    The conversion is safe for year 2038 problem
+
+    :param t: int or float timestamp in (milli)seconds since UNIX epoch
+              or time.struct_time
+    :param units: normalizing factor for the timestamp
+                  (1 for seconds, 1000 for milliseconds)
+                  defaults to 1
+    :return: datetime.datetime object in UTC timezone
+    """
+    if isinstance(t, time.struct_time):
+        v = int(time.mktime(t))
+    elif isinstance(t, (float, int)):
+        v = int(t)
+    else:
+        raise TypeError(t)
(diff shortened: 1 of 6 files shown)


ipapython/ipautil.py
```diff
@@ -1647,3 +1647,14 @@ def rmtree(path):
             shutil.rmtree(path)
     except Exception as e:
         logger.error('Error removing %s: %s', path, str(e))
+
+
+def datetime_from_timestamp(t, units=1000):
```
tiran, line 1652: Could you please add a doc string that explains what ``t`` and ``units`` mean? The function name does not reflect that the function returns an UTC time stamp.

ipapython/ipautil.py
```diff
@@ -1647,3 +1647,14 @@ def rmtree(path):
...
+    if isinstance(t, time.struct_time):
+        v = time.mktime(t)
+    else:
+        v = t
+    value = datetime.datetime(1970, 1, 1,
+                              tzinfo=datetime.timezone.utc
+                              ) + datetime.timedelta(seconds=(int(v) // units))
+    return value
```
tiran, line 1660:
```suggestion
    if isinstance(t, time.struct_time):
        v = int(time.mktime(t))
    elif isinstance(t, (float, int)):
        v = int(t)
    else:
        raise TypeError(t)
    epoch = datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone.utc)
    return epoch + datetime.timedelta(seconds=v // units)
```

ipapython/ipautil.py
abbra: it returns 'datetime.datetime' object, not a timestamp. I'll add docstring.

rcritten:
Just a nit but should the function include "uct" in it somewhere (prefix/suffix) since it handles UTC times? That would match the other python time function/method name patterns.

abbra:
@rcritten I renamed the method to `datetime_from_utctimestamp()`. All tests passed.

PR #4854, opened 2020-06-24
ipatests: Tests for ipahealthcheck tool with IPA external CA step1 installation only.

menonsudhir:
This testsuite checks whether the healthcheck tool reports correct status in a scenario where only Step1 installation
is performed for an IPA Server with external CA option. Below are some of the checks which report the status accordingly.

DogtagCertsConnectivityCheck, 
IPADomainCheck
IPACRLManagerCheck
IPAFileCheck
IPADNARangeCheck
IPACAChainExpirationCheck
IPACertRevocation
IPARAAgent
IPACertTracking
IPACertNSSTrust
IPANSSChainValidation
IPAOpenSSLChainValidation
IPACertmongerCA
IPAOpenSSLChainValidation

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
--- a/ipatests/prci_definitions/temp_commit.yaml
+++ b/ipatests/prci_definitions/temp_commit.yaml
@@ -68,7 +68,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-latest/build_url}'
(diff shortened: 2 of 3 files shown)


rcritten:
Do we expect people to run healthcheck in between steps?

I wonder if it should instead just skip everything that isn't completely configured.

menonsudhir:
> Do we expect people to run healthcheck in between steps?
The intention behind writing such test was to ensure that healthcheck tool catches critical/error message that could easily be seen in an incomplete setup. 
However as per design if we do not allow healthcheck tool to run on an incomplete setup, the tool should basically detect and display a warning message, saying that 'IPA is not configured'.

menonsudhir:
Closing this PR since the healthcheck tool is expected to be run on a fully configured system even in the case of an external CA.

PR #4856, opened 2020-06-25
ipatests: Creating tmp files for systemd on behalf of certmonger

menonsudhir:
This testcase checks that error related to certmonger is not displayed on the console when the command 'systemd-tmpfiles --create' is run.


flo-renaud:
@menonsudhir 
IMO the issue is not worth automating (and it has TestCaseNotNeeded keyword). The message logged is only informational and happens because /var/run/certmonger needs to be replaced by /run/certmonger (/var/run is a link to /run).

menonsudhir:
@flo-renaud thanks for the review, will close the PR.

PR #4861, opened 2020-06-26
certupdate: update config when deployment becomes CA-ful

frasertweedale:
*Another aged-in-oak branch I've had sitting around for years.*

When a deployment gets promoted from CA-less to CA-ful other
replicas still have enable_ra=False in default.conf, and do not have
the ra-agent key and certificate.  Enhance ipa-certupdate to detect
when the deployment has become CA-ful; retrieve the ra-agent
credential and update default.conf.

The rationale for adding this behaviour to ipa-certupdate is that it
is already necessary to use this command to update local trust
stores with the new CA certificate(s).  So by using ipa-certupdate
we avoid introducing additional steps for administrators.

It is necessary to choose a CA master to use as the ca_host.  We use
the first server returned by LDAP.  A better heuristic might be to
choose a master in the same location but this is just left as a
comment unless or until the need is proven.

This change also addresses the case of a CA server being removed
from the topology, i.e. ipa-certupdate detects when non-CA replicas
are pointing at the removed server, and chooses a new ca_host.

HOW TO TEST:

1. Install a CA-less server (first server).

2. Install a CA-less replica.

3. Run 'ipa-ca-install' on first server, promoting deployment from
   CA-less to CA-ful.

4. Run 'ipa-certupdate' on second server.

5. Exceute 'ipa cert-show 5' on second server.  Should succeed,
   because ra-agent credential was retrieve and default.conf
   updated at step #4.

Fixes: https://pagure.io/freeipa/issue/7188

### DIFF
--- a/ipaclient/install/ipa_certupdate.py
+++ b/ipaclient/install/ipa_certupdate.py
@@ -106,10 +106,17 @@ def run_with_args(api):
 
     server_fstore = sysrestore.FileStore(paths.SYSRESTORE)
     if server_fstore.has_files():
+        # look up CA servers before service restarts
+        resp = api.Command.server_role_find(
+            role_servrole=u'CA server',
+            status='enabled',
+        )
+        ca_servers = [server['server_server'] for server in resp['result']]
+
         update_server(certs)
 
         # pylint: disable=import-error,ipa-forbidden-import
-        from ipaserver.install import cainstance
+        from ipaserver.install import cainstance, custodiainstance
         # pylint: enable=import-error,ipa-forbidden-import
 
         # Add LWCA tracking requests.  Only execute if *this server*
@@ -120,6 +127,19 @@ def run_with_args(api):
             except Exception:
                 logger.exception(
                     "Failed to add lightweight CA tracking requests")
+
+        try:
+            update_server_ra_config(
+                cainstance, custodiainstance,
+                api.env.enable_ra, api.env.ca_host, ca_servers,
+            )
+        except Exception:
+            logger.exception("Failed to update RA config")
+
+        # update_server_ra_config possibly updated default.conf;
+        # restart httpd to pick up changes.
+        if services.knownservices.httpd.is_running():
+            services.knownservices.httpd.restart()
 
     update_client(certs)
 
@@ -153,9 +173,6 @@ def update_server(certs):
     update_db(paths.ETC_DIRSRV_SLAPD_INSTANCE_TEMPLATE % instance, certs)
     if services.knownservices.dirsrv.is_running():
         services.knownservices.dirsrv.restart(instance)
-
-    if services.knownservices.httpd.is_running():
-        services.knownservices.httpd.restart()
 
     criteria = {
         'cert-database': paths.PKI_TOMCAT_ALIAS_DIR,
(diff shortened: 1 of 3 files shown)


frasertweedale:
Don't merge this yet.  I think the next problem is that IPA RA cert is not present on the CA-less replicas after promoting deployment from CA-less to CA-ful.  Need to investigate further, but out of time today.

frasertweedale:
Ready for review now.  Test steps in PR description.

frasertweedale:
I'm suggesting ipa-next to be conservative, because there is a substantial new behaviour in `ipa-certupdate`.  I think a shakedown in master only is a good idea before backports are considered.

frasertweedale:
@flo-renaud thanks, I am developing on f31 and it worked.  I'll investigate on f32 next week.  Cheers!

ipaclient/install/ipa_certupdate.py
```diff
@@ -106,15 +106,31 @@ def run_with_args(api):
 
     server_fstore = sysrestore.FileStore(paths.SYSRESTORE)
     if server_fstore.has_files():
+        # look up CA servers before service restarts
+        resp = api.Command.server_role_find(
+            role_servrole=u'CA server',
+            status='enabled',
+        )
+        ca_servers = [server['server_server'] for server in resp['result']]
```
rcritten, line 114: This is also going to require that the current principal has read access to roles. Today this works:
```
kinit someuser
ipa-certupdate
```
but ca_servers will be an empty list and I don't know what impact that will have.

rcritten: Oh, I see it will skip things. Still perhaps not expected but at least it won't break anything.

frasertweedale: @rcritten yes, empty result is handled.  But also, unprivileged users and read serverrole objects with default ACLs.

rcritten: Yeah, I must have run ipa server-find instead or something to verify. So yeah, this is fine.

frasertweedale:
I still need to investigate the failures described by @flo-renaud.

frasertweedale:
> Hi @frasertweedale
> Thanks for the PR. I have a few comments:
> 
>     * if the server where the CA is installed is in SElinux enforcing mode, ipa-certupdate fails on the replica in fetch_key. Probably an AVC on the CA server while getting the key with /ipa/keys/ra/ipaCert.
> 
>     * after the /etc/ipa/default.conf file has been updated, a restart of httpd is required, otherwise ipa cert-show commands fail.

@flo-renaud I was unable to reproduce this behaviour on f32.  I encountered no issues with SELinux, and `cert-show` command worked immediately after `ipa-certupdate` (no explicit restart was required).

If you are still experiencing the issue could you please provide more details about your environment, transcript of steps, and any relevant logs.

Thanks for reviewing and testing this change!

flo-renaud:
Hi @frasertweedale 
here is my procedure to reproduce the 2nd issue:
- (server) ipa-server-install CA-less
- (replica) ipa-client-install - kinit admin - ipa-replica-install CA-less
- (server) ipa-ca-install
- (replica) ipa-certupdate
- (replica) ipa cert-show 5 => error
- (replica) systemctl restart httpd; ipa cert-show 5 => success
Both server and replica are installed with fedora32, updated with latest pkgs from updates repo.
I wasn't able to reproduce the selinux issue, probably went away when I updated my machines.

Logs for the cert-show issue: on (replica)
```
ipa: ERROR: non-public: AttributeError: ra
  File "/usr/lib/python3.8/site-packages/ipalib/plugable.py", line 350, in __getattr__
    return self[key]
  File "/usr/lib/python3.8/site-packages/ipalib/plugable.py", line 342, in __getitem__
    plugin = self.get_plugin(key) 
  File "/usr/lib/python3.8/site-packages/ipalib/plugable.py", line 339, in get_plugin
    return self.__plugins_by_key[key]
KeyError: 'ra'

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.8/site-packages/ipaserver/rpcserver.py", line 395, in wsgi_execute
    result = command(*args, **options)
  File "/usr/lib/python3.8/site-packages/ipalib/frontend.py", line 471, in __call__
    return self.__do_call(*args, **options)
  File "/usr/lib/python3.8/site-packages/ipalib/frontend.py", line 499, in __do_call
    ret = self.run(*args, **options)
  File "/usr/lib/python3.8/site-packages/ipalib/frontend.py", line 821, in run
    return self.execute(*args, **options)
  File "/usr/lib/python3.8/site-packages/ipaserver/plugins/cert.py", line 1373, in execute
    result = self.Backend.ra.get_certificate(str(serial_number))
  File "/usr/lib/python3.8/site-packages/ipalib/plugable.py", line 352, in __getattr__
    raise AttributeError(key)
AttributeError: ra
ipa: INFO: [jsonserver_session] admin@IPA.TEST: cert_show/1('1', version='2.239'): InternalError
```

frasertweedale:
@flo-renaud thank you, I have reproduced the issue.

frasertweedale:
@flo-renaud I pushed an update that should avoid having to explicitly restart `httpd`.  Thanks again for review and testing this change!

flo-renaud:
Hi @frasertweedale 
I re-ran my procedure and it's now working with your update, thanks.
Also tested the uninstallation of a CA server (topology with server1: CA, server2: CA, server3: no CA), and the replica without CA is properly reconfigured to use the other CA.

frasertweedale:
master:

* 2fcc260cae4bd1186a7312c61a63144e8d656cd0 cainstance.update_ipa_conf: allow specifying ca_host
* a1b3b34b906808c2deb69a3838edd0cf4739b467 cainstance: extract function import_ra_key
* 53d472b490ac7a14fc78516b448d4aa312b79b7f (HEAD) certupdate: update config after deployment becomes CA-ful

frasertweedale:
@flo-renaud thank you so much for your thorough testing.

PR #4866, opened 2020-06-29
Add __signature__ to plugins

tiran:
Auto-generate inspect.Signature from plugin arguments and options. The
signature is used by (amongst others) pydoc / help.

```
$ ipa console
>>> help(api.Command.group_add)
Help on group_add in module ipaserver.plugins.group object:

class group_add(ipaserver.plugins.baseldap.LDAPCreate)
 |  group_add(cn: str, *, description: str = None, gidnumber: int = None, setattr: List[str] = None, addattr: List[str] = None, nonposix: bool, external: bool, all: bool, raw: bool, version: str = None, no_members: bool) -> Dict[str, Any]
```

Fixes: https://pagure.io/freeipa/issue/8388

### DIFF
--- a/ipalib/cli.py
+++ b/ipalib/cli.py
@@ -949,6 +949,14 @@ class show_mappings(frontend.Command):
             print(to_cli(item[0]).ljust(mcl)+' : '+item[1])
 
 
+class IPACompleter(rlcompleter.Completer):
+    def _callable_postfix(self, val, word):
+        # Don't add '(' postfix for callable API objects
+        if isinstance(val, (plugable.APINameSpace, plugable.API)):
+            return word
+        return super()._callable_postfix(val, word)
+
+
 class console(frontend.Command):
     """Start the IPA interactive Python console, or run a script.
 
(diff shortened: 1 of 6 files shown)


rcritten:
Might this be confusing since this shows the python equivalent of the IPA data types?

tiran:
> Might this be confusing since this shows the python equivalent of the IPA data types?

I'm using allowed_types here. Unless I'm mistaken it shows what types are accepted by the argument or option. Here is another example:

```
usercertificate: List[Union[cryptography.x509.base.Certificate, bytes, str]] = None
```

It's a list of items where each item has to be a certificate, bytes, or str.

abbra:
All runs succeeded but Azure was unable to get the status of a run back to github due to an outage earlier.

How can we test this code?

tiran:
> How can we test this code?

I have added several simple tests that mimic ``help()`` and tab completion in the console.

abbra:
LGTM.

tiran:
master:

* 069f41a01e65f07367000c6d9605167242a88736 Add __signature__ to plugins
* 80794f6b5e263566a1da9685aba82762208a6219 Make tab completion in console more useful

PR #4867, opened 2020-06-29
ipatests: test_epn - generate large batch of users

miskopo:
Prepare large (10k) batch of users with various attributes, namely
without passwords, gradually expiring passwords, unicode names and
unicode e-mails.

### DIFF
--- a/ipatests/test_integration/test_epn.py
+++ b/ipatests/test_integration/test_epn.py
@@ -844,3 +844,8 @@ class TestEPN(IntegrationTest):
             # benchmarks show that + concat is the fastest for large strings
             useradd_ldif += str(generate_user_ldif(user))
         tasks.ldap_add(self.master, useradd_ldif)
+
+    def test_temporary_batch_invocation(self):
+        users = self.generate_user_batch()
+        # RIP runner
+        self.add_user_batch_to_ldap(users)
--- a/ipatests/test_integration/test_epn.py
+++ b/ipatests/test_integration/test_epn.py
@@ -27,16 +27,19 @@ from __future__ import print_function, absolute_import
 import base64
 import datetime
 import email
+import io
 import json
 import logging
 import os
-import pytest
+import random
 import textwrap
-
+from collections import deque
 from subprocess import CalledProcessError
 
-from ipatests.test_integration.base import IntegrationTest
+import pytest
+
 from ipatests.pytest_ipa.integration import tasks
+from ipatests.test_integration.base import IntegrationTest
 
 logger = logging.getLogger(__name__)
 
@@ -58,6 +61,86 @@ def datetime_to_generalized_time(dt):
 def postconf(host, option):
     host.run_command(r"postconf -e '%s'" % option)
 
+
+def generate_user(host, uid, **kwargs):
+    user_defaults = {
+        'objectClass': [
+            'ipaobject',
+            'person',
+            'top',
+            'ipasshuser',
+            'inetorgperson',
+            'organizationalperson',
+            'krbticketpolicyaux',
+            'krbprincipalaux',
+            'inetuser',
+            'posixaccount',
+            'ipaSshGroupOfPubKeys',
+            'mepOriginEntry'],
+        'ipaUniqueID': ['autogenerate'],
+        'loginShell': ['/bin/zsh'],
+        'uidNumber': ['-1'],
+        'gidNumber': ['-1'],
+    }
+
+    user = dict(user_defaults)
+    user['dn'] = 'uid={uid},cn=users,cn=accounts,{suffix}'.format(
+        uid=uid,
+        suffix=host.hostname,
+    )
+    user['uid'] = [uid]
+    user['displayName'] = ['{0} {0}'.format(uid)]
+    user['initials'] = ['{}{}'.format(uid[1], uid[-1])]
+    user['gecos'] = user['displayName']
+    user['sn'] = [uid]
+    user['homeDirectory'] = ['/other-home/{}'.format(uid)]
+    user['mail'] = ['{uid}@{domain}'.format(
+        uid=uid, domain=host.domain)]
+    user['krbPrincipalName'] = ['{uid}@{realm}'.format(
+        uid=uid, realm=host.domain.realm)]
+    user['krbCanonicalName'] = ['{uid}@{realm}'.format(
+        uid=uid, realm=host.domain.realm)]
+    user['givenName'] = [uid]
+    user['cn'] = ['{0} {0}'.format(uid)]
+
+    if "krbPasswordExpiration" in kwargs.keys():
+        user['krbPasswordExpiration'] = kwargs['krbPasswordExpiration']
+
+    if "userPassword" in kwargs.keys():
+        user['userPassword'] = kwargs['userPassword']
+
+    if "Mail2" in kwargs.keys():
+        user['mail'].append(kwargs['Mail2'])
+
+    if "Mail3" in kwargs.keys():
+        user['mail'].append(kwargs['Mail3'])
+
+    return user
+
+
+def generate_user_ldif(user_dict):
+    user_ldif = io.StringIO()
+    print("dn: {}".format(user_dict['dn']), file=user_ldif)
+    for k, values in user_dict.items():
+        if k == 'dn':
+            continue
+        for v in values:
+            print("{}: {}".format(k, v), file=user_ldif)
+    return user_ldif.getvalue()
+
+
+def generate_random_unicode_string(length=6):
+    alphabets = {
+        # taken alphabet samples from Wikipedia
+        'armenian': "աբգդեզէըթժիլխծկհձղճմ ",
+        'latin_ext': "aâêbcçddheèëfggjhijľščťřžýáíéôpqrsśtuúvxyzź",
+        'arabic': "غظضذخثتشرقصفعسنملكيطحزوهدجبأ",
+        'cyrillic': "абвгдеёжзийклмнопрстуфхцчшщъыьэюя",
+        'hebrew': "אבגדהוזחטיכמנסעפצק"
+    }
+    # pylint doesn't like when .keys() is not used with iteration
+    alphabet_type = random.choice(list(alphabets))
+    return random.sample(alphabets[alphabet_type], length)
 
 def configure_postfix(host, realm):
     """Configure postfix for:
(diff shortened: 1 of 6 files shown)


ipatests/test_integration/test_epn.py
```diff
@@ -170,12 +173,12 @@ class TestEPN(IntegrationTest):
     notify_ttls = (28, 14, 7, 3, 1)
 
     def _check_epn_output(
-        self,
-        host,
-        dry_run=False,
-        from_nbdays=None,
-        to_nbdays=None,
-        raiseonerr=True,
+            self,
```
miskopo, line 176: These are false positives, even in hexdump I see them being the same

ipatests/test_integration/test_epn.py
```diff
@@ -494,3 +497,78 @@ def test_EPN_delay_config(self, cleanupmail):
...
+        """
+        Prepare large (10k) batch of users with various attributes, namely
+        without passwords, gradually expiring passwords, unicode names and
+        unicode e-mails
+        :return: None
+        """
+        # number of groups: 5 (one is split in two)
+        # describe list format in deque
+        USERS_IN_GROUP = 2000
+        # inactive users without passwords
```
miskopo, line 511: I am thinking about merging all the deques into one, which would simplify removal and checking should be unaffected, as both username and expiry date are stored

miskopo:
As per discussion with @fcami, for the sake of performance and those poor little electrons running in CPUs, I shall generate ldif and feed it directly to ldap.

ipatests/test_integration/test_epn.py
miskopo: I merged it into one big deque. But worry not, deques are nice and effective.

ipatests/test_integration/test_epn.py
```diff
@@ -494,3 +497,126 @@ def test_EPN_delay_config(self, cleanupmail):
...
+            username = "user_to_exp_ascii_%d" % i
+            email = "%s@%s" % (username, self.master.domain)
+            tasks.user_add(
+                self.master,
+                username,
+                password="Secret123",
+                extra_args=[
+                    "--password-expiration",
+                    krbpasswordexpiration,
+                   "--email",
```
fcami, line 598: This fails:
```
./ipatests/test_integration/test_epn.py:598:20: E131 continuation line unaligned for hanging indent
```

fcami:
hi @miskopo could you please add a temp commit so that only test_epn.py is run?

ipatests/test_integration/test_epn.py
miskopo: That's funny. I wonder how's this part of PEP8 implemented. I use well-known professional tier IDE which I shall not name and is **sometimes** misaligns things during automatic alignment. I guess I should fill a bug report. Anyway, thanks for report.

ipatests/test_integration/test_epn.py
rcritten: I think it's a style thing, pushing the indention deeper so it doesn't line up with result =

It doesn't seem to be a necessary change.

ipatests/test_integration/test_epn.py
```diff
@@ -494,3 +497,126 @@ def test_EPN_delay_config(self, cleanupmail):
...
+            username = "user_wo_pass_%d" % i
+            tasks.user_add(self.master, username)
+            users.append((username, -1))
+
+        # Users with passwords gradually expiring
+        for i in range(USERS_IN_GROUP):
+            username = "user_with_pass_%d" % i
+            # TODO: Compute
+            krbpasswordexpiration = datetime_to_generalized_time(
+                datetime.datetime.utcnow() + datetime.timedelta(days=7)
```
rcritten, line 522: Assuming this runs really really fast the date is going to very only slightly. Would it be better to move this out of the loop and save all the sys calls to get time? I don't think a few seconds will make a difference in the test.

ipatests/test_integration/test_epn.py
```diff
@@ -494,3 +497,126 @@ def test_EPN_delay_config(self, cleanupmail):
...
+                            krbpasswordexpiration],
+            )
+            users.append((username, krbpasswordexpiration))
+
+        # Users with wide unicode in usernames and with passwords to expire
+        all_langs = hg.Languages.get_all()
+        users_per_lang = USERS_IN_GROUP // len(all_langs)
+        remainder = USERS_IN_GROUP % len(all_langs)
+        krbpasswordexpiration = datetime_to_generalized_time(
+            datetime.datetime.utcnow() + datetime.timedelta(days=7)
```
rcritten, line 538: Similar to above where this is outside the loop but you probably still only need to compute the time once since we're checking in full days. Plus depending on the TOD that this test executes it could have wonky results if some users were pre midnight and some were post.

rcritten:
This generally seems like a good approach to generate a bunch of different types of user accounts. I wonder if you want to be able to vary the expiration dates.

For example if you could take in an argument for number of users to create you could use similar math to what you do with the non-ASCII naming and divide all the user types somewhat evenly with a configurable expiration date.

So if I wanted to create 1000 users expiring in 7 days, 1000 in 8 days, 1000 in 9 days, etc. I could do it pretty easily with a few calls. And that way you could build up to huge number of diverse users.

netoarmando:
@miskopo `homoglyphs` package is not installed in the runner because:
- PRCI builds freeipa packages
- PRCI installs all rpms available here: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/be325a7e-e221-11ea-9473-fa163ee3e098/rpms/
- `ipatests` dependencies are defined in freeipa's spec file. 

So the `extras_require` packages are simply not taken into account, the other packages listed there are probably being installed during some other stage.

Since `homoglyphs` is not available as a rpm package, we could pre-install it via `pip` into the PR-CI templates, but ideally we should find a better way (https://fedoraproject.org/wiki/Changes/PythonExtras). I tried this https://github.com/freeipa-pr-ci2/freeipa/pull/357/commits/fcc73101b910cafa5dbe935ecfbca24596d5035c without success.

miskopo:
@rcritten @fcami I think this is ready for review. Suggestions are welcome, nitpicking as well :)

ipatests/test_integration/test_epn.py
```diff
@@ -59,6 +62,79 @@ def postconf(host, option):
...
+            'top',
+            'ipasshuser',
+            'inetorgperson',
+            'organizationalperson',
+            'krbticketpolicyaux',
+            'krbprincipalaux',
+            'inetuser',
+            'posixaccount',
+            'ipaSshGroupOfPubKeys',
+            'mepOriginEntry'],
```
rcritten, line 79: Not a deal breaker but I wonder if the list of objectclasses can be obtained via ipa config-show. This would future-proof the test.

ipatests/test_integration/test_epn.py
```diff
@@ -59,6 +62,79 @@ def postconf(host, option):
...
+        user['krbPasswordExpiration'] = kwargs['krbPasswordExpiration']
+
+    if "userPassword" in kwargs.keys():
+        user['userPassword'] = kwargs['userPassword']
+
+    if "Mail2" in kwargs.keys():
+        user['Mail2'] = kwargs['Mail2']
+
+    if "Mail3" in kwargs.keys():
+        user['Mail3'] = kwargs['Mail3']
```
rcritten, line 116: Mail2 and Mail3 are not valid attributes. Are you trying to make it multi-valued?

ipatests/util.py
```diff
@@ -25,15 +25,15 @@
...
 import re
+import shutil
+import tempfile
 import uuid
-import pytest
 from contextlib import contextmanager
+from os import path
 from pprint import pformat
 
+import pytest
```
rcritten, line 36: I assume these import changes aren't needed anymore?

ipatests/test_integration/test_epn.py
```diff
@@ -59,6 +62,79 @@ def postconf(host, option):
...
+
+    return user
+
+
+def generate_user_ldif(user_dict):
+    return ldif.CreateLDIF(dn=user_dict['dn'],record=user_dict)
```
rcritten, line 122: CreateLDIF is deprecated. Can you use ldif.LDIFWriter() instead?

Also this requires bytes, not strings, so it will fail. At least in Fedora.

ipatests/test_integration/test_epn.py
```diff
@@ -674,3 +750,92 @@ def test_EPN_reinstall(self):
...
+                                 userPassword=user_password,
+                                 Mail2=email_1,
+                                 Mail3=email_2)
+            users.append(user)
+
+            return users
+
+    def add_user_batch_to_ldap(self, users):
+        for user in users:
+            tasks.ldap_add(self.master, str(generate_user_ldif(user)))
```
rcritten, line 836: It will be far more efficient to generate a single humongous LDIF and import that rather than sshing # users times.

ipatests/test_integration/test_epn.py
miskopo: I am trying to add multiple emails.
Dummy `ldapsearch` showed, that multiple values are stored like this (this is supposedly `LDIF` format):

```
krbPrincipalName: jozkomrkvicka@never.gonna.give.you.up
mail: jozko@mrkvicka.com
mail: igor@hnizdo.com
mail: mehehe@foo.bar
```

ipatests/test_integration/test_epn.py
miskopo: It would be probably better. I guess I could leave it as it is right now and add comment for future revisions

ipatests/util.py
miskopo: It is unused in the code, but I am not sure any more. Nothing is certain. Entropy is over 900.

ipatests/test_integration/test_epn.py
miskopo: Great point.

ipatests/test_integration/test_epn.py
miskopo: Great point

ipatests/test_integration/test_epn.py
rcritten: The value of mail should be a list and each unique address appended to that and the LDIFWriter will handle the rest.

Otherwise it will happily write out the Mail2 and Mail3 attributes but the the result will not be importable since those are not valid attribute names.

miskopo:
@rcritten I am frankly stuck with current error. I tried using `keys` and `values` as string, as bytes, then only `keys` as string and `values` as bytes, but it still fails.

ipatests/test_integration/test_epn.py
```diff
@@ -674,3 +757,95 @@ def test_EPN_reinstall(self):
...
+            email_2 = "%s_2@%s" % (username, self.master.domain)
+
+            user = generate_user(self.master, username,
+                                 krbPasswordExpiration=expiration,
+                                 userPassword=user_password,
+                                 Mail2=email_1,
+                                 Mail3=email_2)
+            users.append(user)
+
+            return users
```
flo-renaud, line 839: The indentation seems wrong here (return inside the for loop).

ipatests/test_integration/test_epn.py
```diff
@@ -59,6 +63,85 @@ def postconf(host, option):
...
+
+
+def generate_user_ldif(user_dict):
+    writer = ldif.LDIFWriter(io.StringIO(), None, 76, '\n')
+    user_dict_bytes = {
+        key: val.encode('utf-8') if not isinstance(val, list)
+        else [item.encode('utf-8') for item in val]
+        for key, val in user_dict.items()
+    }
+    return writer.unparse(dn=user_dict['dn'], record=user_dict_bytes)
```
flo-renaud, line 129: unparse is not a function that returns data. It should be used like this:
```
io = StringIO()
writer = ldif.LDIFWriter(io)
user_dict_bytes = {
    key: val.encode('utf-8') if not isinstance(val, list)
    else [item.encode('utf-8') for item in val]
    for key, val in user_dict.items()
}
writer.unparse(dn=user_dict['dn'], record=user_dict_bytes)
return io.getvalue()
```

miskopo: Thank you Flo!

ipatests/test_integration/test_epn.py
```diff
@@ -59,6 +62,86 @@ def postconf(host, option):
...
+        'ipaUniqueID': ['autogenerate'],
+        'loginShell': ['/bin/zsh'],
+        'uidNumber': ['-1'],
+        'gidNumber': ['-1'],
+    }
+
+    user = dict(user_defaults)
+    user['dn'] = 'uid={uid},cn=users,cn=accounts,{suffix}'.format(
+        uid=uid,
+        suffix=host.hostname,
```
rcritten, line 89: suffix should be ipautil.realm_to_suffix(host.domain.realm) and not hostname.

ipatests/test_integration/test_epn.py
```diff
@@ -59,6 +62,86 @@ def postconf(host, option):
...
+
+def generate_user_ldif(user_dict):
+    user_ldif = io.StringIO()
+    print("dn: {}".format(user_dict['dn']), file=user_ldif)
+    for k, values in user_dict.items():
+        if k == 'dn':
+            continue
+        for v in values:
+            print("{}: {}".format(k, v), file=user_ldif)
+    return user_ldif.getvalue()
```
rcritten, line 129: If you wanted to go back to using LDIFWriter this would fix it:
```
    output = io.StringIO()
    writer = ldif.LDIFWriter(output, None, 76, '\n')
    user_dict_bytes = {
        key: val.encode('utf-8') if not isinstance(val, list)
        else [item.encode('utf-8') for item in val]
        for key, val in user_dict.items()
    }
    user_dict_bytes.pop('dn')
    writer.unparse(dn=user_dict['dn'], record=user_dict_bytes)
    output.seek(0)
    return output.read()
```

ipatests/test_integration/test_epn.py
```diff
@@ -674,3 +757,95 @@ def test_EPN_reinstall(self):
...
+                                 Mail3=email_2)
+            users.append(user)
+
+        return users
+
+    def add_user_batch_to_ldap(self, users):
+        useradd_ldif = ""
+        for user in users:
+            # benchmarks show that + concat is the fastest for large strings
+            useradd_ldif += str(generate_user_ldif(user))
```
rcritten, line 845: Alternatively if you create all the users as a list and then loop over that list you can let LDIFWriter generate the huge string for you:

```
def generate_users_ldif(users):
    output = io.StringIO()
    writer = ldif.LDIFWriter(output, None, 76, '\n')
    for user_dict in users:
        user_dict_bytes = {
            key: val.encode('utf-8') if not isinstance(val, list)
            else [item.encode('utf-8') for item in val]
            for key, val in user_dict.items()
        }
        user_dict_bytes.pop('dn')
        writer.unparse(dn=user_dict['dn'], record=user_dict_bytes)
    output.seek(0)
    return output.read()
```

abbra:
Reopen the PR, hopefully this will be rebased and updated soon.

PR #4868, opened 2020-06-30
ca-del: require CA to already be disabled

frasertweedale:
*Another aged-in-oak commit from the IPA->Dogtag GSS-API effort.
This one should be ipa-next and have a release note attached.*

Currently ca-del disables the target CA before deleting it.
Conceptually, this involves two separate permissions: modify and
delete.  A user with delete permission does not necessarily have
modify permission.

As we head toward enforcing IPA permissions in Dogtag, it is
necessary to decouple disablement from deletion, otherwise the
disable operation shall fail if the user does not have modify
permission.  Although it introduces an additional step for
administrators, the process is consistent, required permissions map
1:1 to the operations, and the error messages make it clear what
needs to happen (i.e. disable first).

### DIFF
--- a/ipaserver/plugins/ca.py
+++ b/ipaserver/plugins/ca.py
@@ -342,7 +342,12 @@ class ca_del(LDAPDelete):
 
         ca_id = self.api.Command.ca_show(keys[0])['result']['ipacaid'][0]
         with self.api.Backend.ra_lightweight_ca as ca_api:
-            ca_api.disable_ca(ca_id)
+            data = ca_api.read_ca(ca_id)
+            if data['enabled']:
+                raise errors.ProtectedEntryError(
+                    label=_("CA"),
+                    key=keys[0],
+                    reason=_("Must be disabled first"))
             ca_api.delete_ca(ca_id)
 
         return dn
--- a/ipaserver/plugins/ca.py
+++ b/ipaserver/plugins/ca.py
@@ -35,20 +35,22 @@ certificate to be revoked and its private key deleted.
 """) + _("""
 EXAMPLES:
 """) + _("""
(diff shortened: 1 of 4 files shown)


rcritten:
The patch looks good and does require a disable before deleting but delegating disablement may not be clear.

The disable command is granted via System: Modify CA.

There is no separate Disable CA option. It's likely possible to do so with the current model but it would require a different test. Not sure if that's in the scope for this.

I never noticed before that the CAQuery class is quite similar to the VirtualCommand class.

frasertweedale:
@rcritten I can sprinkle some comments to make it clear that enable/disable is granted by the "Modify" permission.

frasertweedale:
@rcritten plugin doc updated on latest push (I added another commit just for the doc changes).

rcritten:
Changes looks good. You dropped the ipa-next label, can you add any backports you'd like? I'm assuming you want this in ipa-4-8 as well but don't want to guess.

I'm acking this patch but not setting the label yet to avoid confusion since there is an open question. When you set the backports feel free to add ack label as well and merge.

frasertweedale:
@rcritten whoops, didn't mean to remove ipa-next.  IMO this should not be backported.

frasertweedale:
Test failures are introduced by this PR.  Pursuing solution.

rcritten:
Test changes LGTM.

rcritten:
master:

* 5ab24ddf8a74c4c59c5c28bc79da150633530363 ca-del: require CA to already be disabled
* 6da63e3be44033d7d623a574787937ab4ec9ed0a ca plugin: improve doc
* 6a0901f6fdf744023dc90ec5141aa9ebe4d806fd tests: fix cleanup for CATracker

PR #4869, opened 2020-06-30
ra.get_certificate: use REST API

frasertweedale:
*Another patch dragged kicking and screaming into the new decade.*

Update ra.get_certificate to use the Dogtag REST API.  This change
is being done as part of the Dogtag GSS-API authentication effort
because the servlet-based method expects an internal Dogtag user.
It is less intrusive to just change FreeIPA to call the REST API
instead (which is also part of an existing ticket).

Depends on https://pagure.io/dogtagpki/issue/2601 (which was merged
and released long ago).

### DIFF
--- a/ipaserver/plugins/dogtag.py
+++ b/ipaserver/plugins/dogtag.py
@@ -250,7 +250,7 @@ import contextlib
 
 import six
 
-from ipalib import Backend, api
+from ipalib import Backend, api, x509
 from ipapython.dn import DN
 import ipapython.cookie
 from ipapython import dogtag, ipautil, certdb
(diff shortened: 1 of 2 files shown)


ipaserver/plugins/dogtag.py
```diff
@@ -1575,49 +1460,41 @@ def get_certificate(self, serial_number):
...
-        serial_number = int(serial_number, 0)
-
         # Call CMS
-        http_status, _http_headers, http_body = (
-            self._sslget('/ca/agent/ca/displayBySerial',
-                         self.env.ca_agent_port,
-                         serialNumber=str(serial_number),
-                         xml='true')
+        path = 'certs/{}'.format(serial_number)
+        _http_status, _http_headers, http_body = self._ssldo(
```
tiran, line 1465: Have you considered to replace ``_ssldo()`` with Dogtag's API ``pki.certs.CertClient.get_cert()``? I would prefer to slowly get rid of our own API wrapper and use pki client library instead.

frasertweedale: It's a fine idea.  Maybe for another commit?  I don't plan on spending more than the minimum time to get these patches working, reviewed and merged.  That means it's either take this as-is (subject to whatever minor changes are asked in review), or it will be abandoned.  I don't really mind either way - I just wanted to get all my old patches "out there" for consideration.

tiran: SGTM! Let's get rid of ``_ssldo`` in another commit.

ipaserver/plugins/dogtag.py
```diff
@@ -1575,49 +1460,41 @@ def get_certificate(self, serial_number):
...
         cmd_result = {}
 
-        if 'certificate' in parse_result:
-            cmd_result['certificate'] = parse_result['certificate']
+        if 'Encoded' in resp:
+            s = resp['Encoded']
+            match = x509.PEM_CERT_REGEX.search(s.encode('utf-8'))
+            if match:
+                s = match.group(2).decode('utf-8')
+            cmd_result['certificate'] = s.strip()
```
tiran, line 1486: I don't quite understand why the regexp is necessary here. Could you please add a comment to educate your fellow developers what is going on here?

frasertweedale: sure, it's just stripping the PEM headers; the old method to do that was removed.  I'll add a comment.

ipaserver/plugins/dogtag.py
frasertweedale: Cool, thanks :)

tiran:
master:

* d7f3a0b2d31fe3a02a19286d590cf1a343458fd8 ra.get_certificate: use REST API

PR #4870, opened 2020-06-30
extract virtual operation access check subroutine

frasertweedale:
*Another GSS-API related patch that could be merged with minimal risk or
disruption.*

Outside of virtual commands themselves there is no way to evaluate
access to perform a virtual operation.  Such a capability will be
needed for Dogtag-based certificate request validation using
Kerberos proxy credentials.

Add the 'check_operation_access' method for explicit virtual
operation access checks.  Refactor 'VirtualCommand.check_access()'
to use it.

Part of: https://pagure.io/freeipa/issue/6423

### DIFF
--- a/ipaserver/plugins/virtual.py
+++ b/ipaserver/plugins/virtual.py
@@ -59,16 +59,28 @@ class VirtualCommand(Command):
         if operation is None:
             operation = self.operation
 
-        ldap = self.api.Backend.ldap2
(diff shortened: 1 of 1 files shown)


tiran:
master:

* 0c0061babdd9a6584ea30742648a200d6aa4d3ba extract virtual operation access check subroutine

frasertweedale:
Thanks @tiran!

PR #4820, opened 2020-06-16
Specify cert_paths when calling PKIConnection

cipherboy:
`PKIConnection` now defaults to specifying `verify=True`. We've introduced
a new parameter, `cert_paths`, to specify additional paths (directories or
files) to load as certificates. Specify the IPA CA certificate file so
we can guarantee connections succeed.

`Signed-off-by: Alexander Scheel <ascheel@redhat.com>`

---

This is blocked on dogtagpki/pki#443. Once that is merged, this can be updated and merged. What's this project's preference on requiring attributes from a new Dogtag API?

PR #4873, opened 2020-07-01
cainstance.is_crlgen_enabled: handle missing ipa-pki-proxy.conf

frasertweedale:
A failed ipa-ca-install left my installation in an inconsistent
state.  Then, 'ipa-server-install --uninstall' also failed when
is_crlgen_enabled() tried to read ipa-pki-proxy.conf, which was
missing.

Update is_crlgen_enabled() to handle missing ipa-pki-proxy.conf so
that uninstallation works.

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -1402,7 +1402,7 @@ class CAInstance(DogtagInstance):
                         rewriteRuleDisabled = False
                         break
         except IOError:
-            raise RuntimeError(
+            raise InconsistentCRLGenConfigException(
                 "Unable to read {}".format(paths.HTTPD_IPA_PKI_PROXY_CONF))
 
         # if enableCRLUpdates and rewriteRuleDisabled are different, the config


frasertweedale:
Side-note: the issue that caused the installation failure is discussed at https://github.com/freeipa/freeipa/pull/4820#issuecomment-648729659.  It is fixed in pki commit 1b4558bf7ebac2d1fcf5d431ac12b0c6a15dadb9 which was not in my Dogtag build.

ipaserver/install/cainstance.py
```diff
@@ -1396,11 +1396,18 @@ def is_crlgen_enabled(self):
...
+                with open(paths.HTTPD_IPA_PKI_PROXY_CONF) as f:
+                    for line in f.readlines():
+                        if p.search(line):
+                            rewriteRuleDisabled = False
+                            break
+            else:
+                # Installation seems to be in inconsistent state.
+                # Throwing an exception could e.g. inhibit uninstall
+                # (preventing cleanup), so let's not do that.
+                rewriteRuleDisabled = False
```
flo-renaud, line 1409: If the code throws an InconsistentCRLGenConfigException, uninstall will catch the exception and proceed either with a prompt "are you sure, deleting your CRL gen master" or abort uninstall in non-interactive mode, unless --ignore-last-of-role is provided.

frasertweedale: It was throwing `IOError` due to missing file, causing whole `ipa-server-install` program to terminate.

flo-renaud: Yes, but my point is that the code could be modified by replacing the `raise RuntimeError` with a `raise InconsistentCRLGenConfigException` after `except IOError` in line 1405.

frasertweedale: Ah I see, thanks for clarifying.  I'll apply your suggestion, thanks @flo-renaud .

flo-renaud:
Hi @frasertweedale 
Thanks for the update, LGTM. Tested the following scenarios:
- [x] install server with CA, replica without CA, delete ipa-pki-proxy.conf on server, uninstall server1 with --ignore-last-of-role => success
- [x] install server with CA, delete ipa-pki-proxy.conf, run ipa-crlgen-manage status: exits with ret code =1 and prints error message about missing file.

flo-renaud:
Which branches (if any) should the fix be backported to?

frasertweedale:
Thanks @flo-renaud.  Hm I think backport to `ipa-4-8` only, if it is conflict-free, otherwise can be `ipa-next` only.

flo-renaud:
> Thanks @flo-renaud. Hm I think backport to ipa-4-8 only, if it is conflict-free, otherwise can be ipa-next only.

Agreed, I am adding the ipa-4-8 label.

abbra:
master:

* 51d5ec17570ef6911bded6381c77041f4a5c2182 cainstance.is_crlgen_enabled: handle missing ipa-pki-proxy.conf

PR #4879, opened 2020-07-01
Test with pki-core 10.9.0-0.4

tiran:


frasertweedale:
ACK. We've already tested extensively with 10.9.0 in #4723 (which is still unmerged).  Azure piplines are still busted due to certmonger segfault.

tiran:
> ACK. We've already tested extensively with 10.9.0 in #4723 (which is still unmerged). Azure piplines are still busted due to certmonger segfault.

This is a draft PR to verify that tests pass without #4820. Let's bump requirements with #4820.

PR #4881, opened 2020-07-02
ipa cert-show: fix the code setting revocation reason

flo-renaud:
ipa cert-show wrongly displays all certs as Revoked.
The dogtag plugin code is checking if the JSON data received
from dogtag contains a RevocationReason with:
        if 'RevocationReason' in resp:
but the value can be None.
Replace the check with
        if resp.get('RevocationReason'):
as this will execute the code only if there is a value
and it is not None.
    
Fixes: https://pagure.io/freeipa/issue/8394

### DIFF
--- a/ipaserver/plugins/dogtag.py
+++ b/ipaserver/plugins/dogtag.py
@@ -1462,12 +1462,18 @@ class ra(rabase.rabase, RestClient):
 
         # Call CMS
         path = 'certs/{}'.format(serial_number)
-        _http_status, _http_headers, http_body = self._ssldo(
-            'GET', path, use_session=False,
-            headers={
-                'Accept': 'application/json',
-            },
-        )
+        try:
+            _http_status, _http_headers, http_body = self._ssldo(
+                'GET', path, use_session=False,
+                headers={
+                    'Accept': 'application/json',
+                },
+            )
+        except errors.HTTPRequestError as e:
+            self.raise_certificate_operation_error(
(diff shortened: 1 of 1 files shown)


frasertweedale:
Thanks for the fix @flo-renaud.  ACK from me when (a) temp_commit CI is green and (b) temp_commit gets removed and gating CI is green.

flo-renaud:
I had to slightly change the fix as `resp['RevocationReason']` can be equal to 0 and in this case the code setting `cmd_result['revocation_reason']` did not get executed.

frasertweedale:
> I had to slightly change the fix as `resp['RevocationReason']` can be equal to 0 and in this case the code setting `cmd_result['revocation_reason']` did not get executed.

Good pickup @flo-renaud!

flo-renaud:
The failure in test_ipahealthcheck is a known issue ([8386](https://pagure.io/freeipa/issue/8386)). Apart from this known failure, the temp commit tests are green.

flo-renaud:
For future ref, the temp commit results can be seen [here](https://github.com/freeipa/freeipa/pull/4881/commits/ee945473063dd932a2072e728b7093434790e7ac) by clicking on the red x left to "Temp commit".
Removing the temp commit.

fcami:
master:

* dcdcd1ce88a6d5ed5997f50758dc6fd025df5f41 ipa cert-show: fix the code setting revocation reason

PR #4882, opened 2020-07-02
selinux: allow oddjobd to set up ipa_helper_t context for execution

abbra:
On Fedora 32+ and RHEL 8.3.0+ SELinux policy requires explicit process
transition from httpd_t context. In addition, a setup of a helper
execution needs permission to use 'noatsecure', 'rlimitinh', and
'siginh'. These operations invoked during execve() setup by glibc.

Fixes: https://pagure.io/freeipa/issue/8395

### DIFF
--- a/selinux/ipa.te
+++ b/selinux/ipa.te
@@ -390,3 +390,7 @@ optional_policy(`
 	sssd_search_lib(ipa_custodia_t)
 	sssd_stream_connect(ipa_custodia_t)
 ')
+
+optional_policy(`
+       systemd_private_tmp(ipa_custodia_tmp_t)
+')
--- a/selinux/ipa.te
+++ b/selinux/ipa.te
@@ -115,6 +115,7 @@ optional_policy(`
 
 
 allow ipa_helper_t self:capability { net_admin dac_read_search dac_override chown };
+seutil_read_config(ipa_helper_t);
 
 #kernel bug
 dontaudit ipa_helper_t self:capability2  block_suspend;
--- a/selinux/ipa.if
+++ b/selinux/ipa.if
@@ -419,3 +419,21 @@ ifndef(`dirsrv_systemctl',`
         ps_process_pattern($1, dirsrv_t)
     ')
 ')
+
+
+########################################
+## <summary>
+##	Allow ipa_helper noatsecure
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`ipa_helper_noatsecure',`
+    gen_require(`
+	type ipa_helper_t;
+    ')
+    allow $1 ipa_helper_t:process { noatsecure };
+')


abbra:
Please see http://blog.siphos.be/2011/04/selinux-and-noatsecure-or-why-portage-complains-about-ld_preload-and-libsandbox-so/ for explanation about noatsecure appearance.

This fix should make fetching of trust topology information working again. The denials are part of don't audit rules and thus were not visible otherwise, causing weird issue with Samba Python bindings not being able to discover proper current kerberos configuration.

As a result, the code went with assumption that the current principal (trusted domain object's credential from AD realm) in a credentials cache belongs to the default realm (IPA). This caused an attempt to request a cross-realm referral to AD domain controller and since the trust is one-way by default (and into an opposite direction), it was rejected by KDC.

selinux/ipa.te
```diff
@@ -108,6 +108,19 @@ optional_policy(`
...
+#
+require {
+	type httpd_t;
+	type oddjob_t;
+	class process { noatsecure rlimitinh siginh transition };
+}
+
+allow oddjob_t httpd_t:process transition;
+allow oddjob_t ipa_helper_t:process { noatsecure rlimitinh siginh };
+
```
tiran, line 123: A SELinux domain should not use external types directly like that. FreeIPA has to use interfaces and wrap them in an optional policy block.

abbra: We are discussing what should be in selinux-policy and what should be in ipa right now with SELinux maintainers.

abbra:
After discussing with @zpytela and others, we came to the following conclusion:
- we need to extend oddjob policy interface to provide a way to tell `ipa_helper_t` is allowed to use `noatsecure` when when transitioning from `oddjob_t`
- we need to add a call to this interface to ipa policy.

Additionally I found that there is an attempt to load selinux policy from `ipa-getkeytab`, most likely due to libkrb5 initialization. This is denied for `ipa_helper_exec_t` context and needs to be extended.

Full set of AVCs on a fresh F32 is the following. We don't need to allow `rlimitinh` and `siginh`, they aren't really required, according to my testing with a  custom policy:

```
type=AVC msg=audit(1593698019.970:2915): avc:  denied  { noatsecure } for  pid=33476 comm="oddjobd" scontext=system_u:system_r:oddjob_t:s0-s0:c0.c1023 tcontext=system_u:system_r:ipa_helper_t:s0 tclass=process permissive=1
type=AVC msg=audit(1593698019.971:2916): avc:  denied  { rlimitinh } for  pid=33476 comm="com.redhat.idm." scontext=system_u:system_r:oddjob_t:s0-s0:c0.c1023 tcontext=system_u:system_r:ipa_helper_t:s0 tclass=process permissive=1
type=AVC msg=audit(1593698019.971:2917): avc:  denied  { siginh } for  pid=33476 comm="com.redhat.idm." scontext=system_u:system_r:oddjob_t:s0-s0:c0.c1023 tcontext=system_u:system_r:ipa_helper_t:s0 tclass=process permissive=1
type=AVC msg=audit(1593698022.000:2918): avc:  denied  { read } for  pid=33479 comm="ipa-getkeytab" name="config" dev="vda1" ino=156614 scontext=system_u:system_r:ipa_helper_t:s0 tcontext=system_u:object_r:selinux_config_t:s0 tclass=file permissive=1
type=AVC msg=audit(1593698022.001:2919): avc:  denied  { open } for  pid=33479 comm="ipa-getkeytab" path="/etc/selinux/config" dev="vda1" ino=156614 scontext=system_u:system_r:ipa_helper_t:s0 tcontext=system_u:object_r:selinux_config_t:s0 tclass=file permissive=1
type=AVC msg=audit(1593698022.001:2920): avc:  denied  { getattr } for  pid=33479 comm="ipa-getkeytab" path="/etc/selinux/config" dev="vda1" ino=156614 scontext=system_u:system_r:ipa_helper_t:s0 tcontext=system_u:object_r:selinux_config_t:s0 tclass=file permissive=1
```

abbra:
SELinux policy change: https://github.com/fedora-selinux/selinux-policy-contrib/pull/292

zpytela:
selinux-policy counterpart:
https://github.com/fedora-selinux/selinux-policy-contrib/pull/293

abbra:
master:

* f6055e6c9f6e88f8174c3bc920ff46583a14822f selinux: allow oddjobd to set up ipa_helper_t context for execution
* 91713f4f0a1033943ccf949fb3bfed2447e3ea0f selinux: support running ipa-custodia with PrivateTmp=yes

PR #4883, opened 2020-07-02
ipatests: Tests for ipahealthcheck tool with IPA external CA

menonsudhir:
This testsuite checks whether the healthcheck tool reports correct status in a scenario when IPA server is setup with
external self-signed CA. Below are the checks covered.
    
IPACRLManagerCheck
IPACertmongerCA
IPAOpenSSLChainValidation
IPANSSChainValidation
IPARAAgent

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -16,6 +16,7 @@ import uuid
 import pytest
 
 from ipalib import x509
+from ipapython.dn import DN
 from ipapython.ipaldap import realm_to_serverid
 from ipapython.certdb import NSS_SQL_FILES
 from ipatests.pytest_ipa.integration import tasks
@@ -23,6 +24,11 @@ from ipaplatform.paths import paths
 from ipaplatform.osinfo import osinfo
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.test_integration.test_cert import get_certmonger_fs_id
+from ipatests.test_integration.test_external_ca import (
+    install_server_external_ca_step1,
+    install_server_external_ca_step2,
+    ISSUER_CN,
+)
 
 HEALTHCHECK_LOG = "/var/log/ipa/healthcheck/healthcheck.log"
 HEALTHCHECK_SYSTEMD_FILE = (
@@ -32,8 +38,9 @@ HEALTHCHECK_LOG_ROTATE_CONF = "/etc/logrotate.d/ipahealthcheck"
 HEALTHCHECK_LOG_DIR = "/var/log/ipa/healthcheck"
 HEALTHCHECK_OUTPUT_FILE = "/tmp/output.json"
 HEALTHCHECK_PKG = ["*ipa-healthcheck"]
-TOMCAT_CFG = "/var/lib/pki/pki-tomcat/conf/ca/CS.cfg"
 
+IPA_CA = "ipa_ca.crt"
+ROOT_CA = "root_ca.crt"
 
 sources = [
     "ipahealthcheck.dogtag.ca",
@@ -346,9 +353,10 @@ class TestIpaHealthCheck(IntegrationTest):
         assert returncode == 0
         for check in data:
             assert check["result"] == "SUCCESS"
-            assert check["kw"]["configfile"] == TOMCAT_CFG
+            assert check["kw"]["configfile"] == paths.CA_CS_CFG_PATH
             assert check["kw"]["key"] in DEFAULT_PKI_CA_CERTS
-        self.master.run_command(["mv", TOMCAT_CFG, TOMCAT_CFG + ".old"])
+        self.master.run_command(["mv", paths.CA_CS_CFG_PATH,
+                                 "%s.old" % paths.CA_CS_CFG_PATH])
         returncode, data = run_healthcheck(
             self.master,
             "ipahealthcheck.dogtag.ca",
@@ -356,7 +364,8 @@ class TestIpaHealthCheck(IntegrationTest):
         )
         assert returncode == 1
         assert data[0]["result"] == "CRITICAL"
-        self.master.run_command(["mv", TOMCAT_CFG + ".old", TOMCAT_CFG])
+        self.master.run_command(["mv", "%s.old" % paths.CA_CS_CFG_PATH,
+                                 paths.CA_CS_CFG_PATH])
         self.master.run_command(["ipactl", "restart"])
 
     @pytest.fixture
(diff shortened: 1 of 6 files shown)


ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +904,101 @@ def test_ipahealthcheck_trust_agent_member_check(self):
         for check in data:
             assert check["result"] == "SUCCESS"
             assert check["kw"]["key"] == self.master.hostname
+
+
+class TestIpaHealthCheckWithExternalSelfSignedCA(IntegrationTest):
+    """
+    Tests to run and check whether ipa-healthcheck tool reports correct status when
+    IPA Master has external self signed CA.
```
mrizwan93, line 912: Shouldn't it be external CA only?

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +904,101 @@ def test_ipahealthcheck_trust_agent_member_check(self):
...
+    """
+
+    @classmethod
+    def install(cls, mh):
+        result = install_server_external_ca_step1(cls.master)
```
mrizwan93, line 917: You need to import this method from `test_integration/test_external_ca.py`.
OR
You can move these method (`test_external_ca.install_server_external_ca_step1()`) to `tasks.py`

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +904,101 @@ def test_ipahealthcheck_trust_agent_member_check(self):
...
+        assert result.returncode == 0
+        root_ca_fname, ipa_ca_fname = tasks.sign_ca_and_transport(
+            cls.master, paths.ROOT_IPA_CSR, ROOT_CA, IPA_CA
+        )
+
+        result = install_server_external_ca_step2(
```
mrizwan93, line 923: Same here, you need to import the method

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +904,101 @@ def test_ipahealthcheck_trust_agent_member_check(self):
...
+            cls.master, ipa_ca_fname, root_ca_fname
+        )
+        assert result.returncode == 0
```
mrizwan93, line 926: You don't need this line. `raiseonerr=True` is set by default for the method.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: @mrizwan93 modified the comment here.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: missing import statement was added.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: Removed the assert statement

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +910,101 @@ def test_ipahealthcheck_trust_agent_member_check(self):
...
+        """
+        Test for IPARAAgent
+        """
+        returncode, data = run_healthcheck(
+            self.master, "ipahealthcheck.ipa.certs", "IPARAAgent"
+        )
+        assert returncode == 0
+        for check in data:
+            assert check["result"] == "SUCCESS"
+            assert check["kw"] == {}
```
rcritten, line 1010: For completion sake I guess this is ok but this is enforced by the baseline test as well so if that fails this will fail too.

This test has a lot of room to test things that can go wrong:

1. paths.RA_AGENT_PEM is missing or mal-formatted
2. The entry doesn't exist in LDAP at all
3. The description doesn't match expectations (like differing serial #)
4. The userCertificate value is different or missing

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +910,101 @@ def test_ipahealthcheck_trust_agent_member_check(self):
...
+        assert returncode == 0
+        for check in data:
+            if check["kw"]["dbdir"] == (paths.PKI_TOMCAT_ALIAS_DIR):
+                assert check["result"] == "SUCCESS"
+                assert check["kw"]["nickname"] == "Server-Cert cert-pki-ca"
+            elif check["kw"]["dbdir"] == (
+                paths.ETC_DIRSRV_SLAPD_INSTANCE_TEMPLATE % instance
+            ):
+                assert check["result"] == "SUCCESS"
+                assert check["kw"]["nickname"] == "Server-Cert"
```
rcritten, line 998: To test failures you could, for example, delete the IPA cert from the database.

I suspect running ipa-certupdate afterwards and running ipactl restart in a fixture would restore things. NSS doesn't often like databases changing while they are running.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +910,101 @@ def test_ipahealthcheck_trust_agent_member_check(self):
...
+            self.master, "ipahealthcheck.ipa.certs", "IPACertmongerCA",
+        )
+        assert returncode == 0
+        for check in data:
+            assert check["result"] == "SUCCESS"
+            assert (
+                check["kw"]["key"] == "IPA"
+                or check["kw"]["key"] == "dogtag-ipa-ca-renew-agent"
+                or check["kw"]["key"] == "dogtag-ipa-ca-renew-agent-reuse"
+            )
```
rcritten, line 962: This will fail to catch if for some crazy reason ipa-healthcheck doesn't return one of the values. You might add a counter to ensure that all 3 are picked up.

rcritten: And moreso if you want to delete one or more of the CA's with getcert remove-ca -c CA_NAME and re-run healthcheck to confirm it is gone that would be even better. With a fixture to re-add.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +910,194 @@ def test_ipahealthcheck_trust_agent_member_check(self):
...
+        self.master.run_command(
+            ["mv", paths.HTTPD_CERT_FILE, paths.HTTPD_CERT_FILE + ".old"]
+        )
+        yield
+        self.master.run_command(
+            ["mv", paths.HTTPD_CERT_FILE + ".old", paths.HTTPD_CERT_FILE]
+        )
+        returncode, data = run_healthcheck(
+            self.master, "ipahealthcheck.ipa.certs", "IPAOpenSSLChainValidation"
+        )
```
rcritten, line 998: This doesn't need to be in the fixture.

menonsudhir: removed the above mentioned code from fixture

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +910,201 @@ def test_ipahealthcheck_trust_agent_member_check(self):
...
+            self.master, "ipahealthcheck.ipa.certs", "IPARAAgent"
+        )
+        assert returncode == 1
+        for check in data:
+            assert check["result"] == "ERROR"
+            assert (
+                check["kw"]["msg"]
+                == "Unable to load RA cert: [Errno 2] No such file or "
+                "directory: '{}'".format(paths.RA_AGENT_PEM)
+            )
```
rcritten, line 1110: This is fine but the RA agent checker does quite a bit more. It compares the cert in the LDAP people entry to the cert on disk and verifies that description matches the expectation: serial number, issue, subject.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +910,201 @@ def test_ipahealthcheck_trust_agent_member_check(self):
...
+            "PR_FILE_NOT_FOUND_ERROR: File not found\n"
+        )
+        returncode, data = run_healthcheck(
+            self.master, "ipahealthcheck.ipa.certs", "IPANSSChainValidation",
+        )
+        assert returncode == 1
+        for check in data:
+            if check["kw"]["nickname"] == "Server-Cert":
+                assert check["result"] == "ERROR"
+                assert check["kw"]["reason"] == error_msg
```
rcritten, line 1081: This is ok I guess but without Server-Cert the service won't start at all so so it would be pretty obvious. I think you delete the CA cert before this that would better exercise things. You could run ipa-certupdate afterward, I think, to re-add the CA cert.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +910,201 @@ def test_ipahealthcheck_trust_agent_member_check(self):
...
+            self.master,
+            "ipahealthcheck.ipa.certs",
+            "IPAOpenSSLChainValidation",
+        )
+        assert returncode == 1
+        for check in data:
+            if check["kw"]["key"] == paths.HTTPD_CERT_FILE:
+                assert check["result"] == "ERROR"
+                assert error_msg in check["kw"]["reason"]
+                assert verification_msg in check["kw"]["msg"]
```
rcritten, line 1015: I'd also test that if /etc/ipa/ca.crt is some random, otherwise valid CA (since external you could try to exclude part of the chain) and confirm that verification fails.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: > For completion sake I guess this is ok but this is enforced by the baseline test as well so if that fails this will fail too.
> 
> This test has a lot of room to test things that can go wrong:
> 
>     1. paths.RA_AGENT_PEM is missing or mal-formatted
> 
>     2. The entry doesn't exist in LDAP at all
> 
>     3. The description doesn't match expectations (like differing serial #)
> 
>     4. The userCertificate value is different or missing
have replaced the existing code and  implemented one of the scenario to modify the description of the RA cert in ldap and the tool reports the status as "RA agent not found in LDAP."

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: Removed the existing test and tried deleting the 'dogtag-ipa-ca-renew-agent' cert using getcert and an error msg is displayed 'Certmonger CA '{key}' missing'

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: Removed the existing test and tried deleting the 'dogtag-ipa-ca-renew-agent' cert using getcert and an error msg is displayed 'Certmonger CA '{key}' missing'

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -904,3 +911,266 @@ def test_ipahealthcheck_trust_agent_member_check(self):
...
+        dn = DN(("uid", "ipara"), ("ou", "People"), ("o", "ipaca"))
+        entry = ldap.get_entry(dn)  # pylint: disable=no-member
+        ldap_cert_desc = entry.single_value.get("description")
+        desc = '200,12,CN=abc,CN=test'
+        entry.single_value['description'] = desc
+        ldap.update_entry(entry)  # pylint: disable=no-member
+        yield
+        ldap = self.master.ldap_connect()
+        dn = DN(("uid", "ipara"), ("ou", "People"), ("o", "ipaca"))
+        entry = ldap.get_entry(dn)  # pylint: disable=no-member
```
rcritten, line 1158: You don't need to reconnect LDAP or re-do dn, they are still available from before the yield.

I think this can be expanded as a fixture into something more generic. Let me see what I can do, that way multiple errors can be detected using the same fixture.

rcritten: This will additionally test that a difference in just the serial # will pop up the right error:

```
diff --git a/ipatests/test_integration/test_ipahealthcheck.py b/ipatests/test_integration/test_ipahealthcheck.py
index 5bc473d63..7c1f7e8d0 100644
@@ -1149,12 +1149,14 @@ class TestIpaHealthCheckWithExternalCA(IntegrationTest):
         dn = DN(("uid", "ipara"), ("ou", "People"), ("o", "ipaca"))
         entry = ldap.get_entry(dn)  # pylint: disable=no-member
         ldap_cert_desc = entry.single_value.get("description")
-        desc = '200,12,CN=abc,CN=test'
-        entry.single_value['description'] = desc
-        ldap.update_entry(entry)  # pylint: disable=no-member
-        yield
-        ldap = self.master.ldap_connect()
-        dn = DN(("uid", "ipara"), ("ou", "People"), ("o", "ipaca"))
+
+        def _update_entry(description):
+            entry = ldap.get_entry(dn)  # pylint: disable=no-member
+            entry.single_value['description'] = description
+            ldap.update_entry(entry)  # pylint: disable=no-member
+
+        yield _update_entry
+
         entry = ldap.get_entry(dn)  # pylint: disable=no-member
         entry.single_value['description'] = ldap_cert_desc
         ldap.update_entry(entry)  # pylint: disable=no-member
@@ -1165,6 +1167,26 @@ class TestIpaHealthCheckWithExternalCA(IntegrationTest):
         is modified, healthcheck tool reports the correct message
         """
         error_msg = 'RA agent not found in LDAP'
+        update_ra_cert_desc('200,12,CN=abc,CN=test')
+
+            self.master,
+            "ipahealthcheck.ipa.certs",
+            "IPARAAgent",
+        )
+        assert returncode == 1
+        for check in data:
+            assert check["result"] == "ERROR"
+            assert check["kw"]["msg"] == error_msg
+
+    def test_ipahealthcheck_iparaagent_bad_serial(self, update_ra_cert_desc):
+        error_msg = 'RA agent description does not match. Found {got} ' \
+                    'in LDAP and expected {expected}'
+        update_ra_cert_desc(
+            '2;16;CN=Certificate Authority,O=%s;CN=IPA RA,O=%s' %
+            (self.master.domain.realm, self.master.domain.realm)
+        )
+
         returncode, data = run_healthcheck(
             self.master,
             "ipahealthcheck.ipa.certs",
@@ -1173,4 +1195,10 @@ class TestIpaHealthCheckWithExternalCA(IntegrationTest):
         assert returncode == 1
         for check in data:
             assert check["result"] == "ERROR"
+            assert check["kw"]["expected"] == '2;6;' \
+                'CN=Certificate Authority,O=%s;CN=IPA RA,' \
+                'O=%s' % (self.master.domain.realm, self.master.domain.realm)
+            assert check["kw"]["got"] == '2;16;' \
+                'CN=Certificate Authority,O=%s;CN=IPA RA,' \
+                'O=%s' % (self.master.domain.realm, self.master.domain.realm)
             assert check["kw"]["msg"] == error_msg
```

rcritten:
The modify_permissions fixture was indented. It is not a class method.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -1411,3 +1416,293 @@ def test_ipa_filesystemspace_check(self, create_jumbo_file):
...
+        Test for IPACRLManagerCheck
+        """
+        returncode, data = run_healthcheck(
+            self.master, "ipahealthcheck.ipa.roles", "IPACRLManagerCheck"
+        )
+        assert returncode == 0
+        for check in data:
+            assert check["result"] == "SUCCESS"
+            assert check["kw"]["key"] == "crl_manager"
+            assert check["kw"]["crlgen_enabled"] is True
```
rcritten, line 1451: I think this test should install a replica as well to ensure that the crlmanagercheck returns crlgen_enabled False.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -1411,3 +1416,293 @@ def test_ipa_filesystemspace_check(self, create_jumbo_file):
...
+                "-c",
+                "dogtag-ipa-ca-renew-agent",
+                "-e",
+                paths.DOGTAG_IPA_CA_RENEW_AGENT_SUBMIT,
+            ]
+        )
+
+    def test_ipahealthcheck_certmongerca(self, getcert_ca):
+        """
+        Test for IPACertmongerCA
```
rcritten, line 1475: Can you change this comment to something like: Test that healthcheck detects that a certmonger-defined CA is missing.

menonsudhir: Modified the description.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -32,6 +38,8 @@
 HEALTHCHECK_PKG = ["*ipa-healthcheck"]
 TOMCAT_CFG = "/var/lib/pki/pki-tomcat/conf/ca/CS.cfg"
```
rcritten, line 39: nit: you can instead uses paths.CA_CS_CFG_PATH

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -1021,12 +1028,10 @@ def test_ipahealthcheck_trust_agent_member_check(self):
             assert check["result"] == "SUCCESS"
             assert check["kw"]["key"] == self.master.hostname
 
-
 @pytest.fixture
 def modify_permissions():
     """Fixture to change owner, group and/or mode
-
-       This can run against multiple files at once but only one host.
+    This can run against multiple files at once but only one host.
```
rcritten, line 1034: I got over eager when removing spaces, can you indent this +4 to line up with block?

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: made the changes to use paths.CA_CS_CFG_PATH

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: intended to +4 to lineup with block

rcritten:
Tests are green, LGTM. @mrizwan93 what do you think?

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -344,17 +350,19 @@ def test_source_ipahealthcheck_dogtag_ca_dogtagcertsconfigcheck(self):
...
         returncode, data = run_healthcheck(
             self.master,
             "ipahealthcheck.dogtag.ca",
             "DogtagCertsConfigCheck",
         )
         assert returncode == 1
         assert data[0]["result"] == "CRITICAL"
-        self.master.run_command(["mv", TOMCAT_CFG + ".old", TOMCAT_CFG])
+        self.master.run_command(["mv", paths.CA_CS_CFG_PATH,
+                                 "%s.old" % paths.CA_CS_CFG_PATH])
```
mrizwan93, line 365: Shouldn't it be other way around?
```
self.master.run_command(["mv", "%s.old" % paths.CA_CS_CFG_PATH,  paths.CA_CS_CFG_PATH])
```

menonsudhir: @mrizwan93 yes i have made the change

rcritten:
Rebased against master

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -1450,3 +1459,374 @@ def test_ipa_filesystemspace_check(self, create_jumbo_file):
...
+        for check in data:
+            if check["kw"]["key"] == paths.HTTPD_CERT_FILE:
+                assert check["result"] == "ERROR"
+                assert error_msg in check["kw"]["reason"]
+            elif check["kw"]["key"] == paths.RA_AGENT_PEM:
+                assert check["result"] == "ERROR"
+                assert error_msg in check["kw"]["reason"]
+
+    @pytest.fixture()
+    def replace_ipa_chain(self):
```
flo-renaud, line 1595: Is this fixture used anywhere?

rcritten: It is supposed to be the fixture for test_opensslchainvalidation_ipa_ca_cert().

The fixture rename_ipa_cert() should be dropped or a separate test for this case added.

replace_ipa_chain() leaves the file there as readable but with an incomplete chain which better replicates what is seen in the user deployments. It is not likely that /etc/ipa/ca.crt will be removed or missing but it having the wrong chain is something we've seen.

menonsudhir: dropped rename_ipa_cert() fixture.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -1603,3 +1612,361 @@ def test_append_arguments_to_list_sources(self):
...
+             "-o", paths.IPA_CA_CRT]
+        )
+        yield
+        self.master.run_command(
+            ["mv", "%s.old" % paths.IPA_CA_CRT, paths.IPA_CA_CRT]
+        )
+
+    def test_opensslchainvalidation_ipa_ca_cert(self, replace_ipa_chain):
+        """
+        Test for IPAOpenSSLChainValidation when IPA CA cert is renamed
```
flo-renaud, line 1738: The comment needs to be fixed: when /etc/ipa/ca.crt contains IPA CA cert but not the external CA.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -1603,3 +1612,361 @@ def test_append_arguments_to_list_sources(self):
...
+        returncode, data = run_healthcheck(
+            self.master, "ipahealthcheck.ipa.certs", "IPACertmongerCA",
+        )
+        assert returncode == 1
+        for check in data:
+            if check["kw"]["key"] == "dogtag-ipa-ca-renew-agent":
+                assert check["result"] == "ERROR"
+                assert (
+                    check["kw"]["msg"]
+                    == "Certmonger CA '{key}' missing"
```
flo-renaud, line 1683: @rcritten The error msg is using {key}, shouldn't it be replaced with the actual helper name?

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -1603,3 +1612,361 @@ def test_append_arguments_to_list_sources(self):
...
+        for check in data:
+            assert check["result"] == "ERROR"
+            assert check["kw"]["msg"] == error_msg
+
+    def test_ipahealthcheck_iparaagent_bad_serial(self, update_ra_cert_desc):
+        """
+        Test to check cert description doesnt match the expected
+        """
+        error_msg = 'RA agent description does not match. Found {got} ' \
+                    'in LDAP and expected {expected}'
```
flo-renaud, line 1953: @rcritten another example of variables not substituted in the error message.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: Comment has been fixed, also the yaml files have been updated to include the ExternalCA test in nightly_* yaml files

menonsudhir:
> Hi @menonsudhir
> the test was added to the nightly definitions in the temp commit, please move the diff to another commit that won't be removed.
> I also noticed one comment from Rob that wasn't addressed:
> 
> > I think this test should install a replica as well to ensure that the crlmanagercheck returns crlgen_enabled False.



> Hi @menonsudhir
> the test was added to the nightly definitions in the temp commit, please move the diff to another commit that won't be removed.
> I also noticed one comment from Rob that wasn't addressed:
> 
> > I think this test should install a replica as well to ensure that the crlmanagercheck returns crlgen_enabled False.
@flo-renaud , i have added the code back which was removed while rebase was done.

flo-renaud:
@menonsudhir 
pycodestyle reported an error:
```
-----------
./ipatests/test_integration/test_ipahealthcheck.py:1654:1: W293 blank line contains whitespace
```

ipatests/test_integration/test_ipahealthcheck.py
rcritten: No, it's a placeholder within the string. This is to make future i18n of the error messages possible.

ipatests/test_integration/test_ipahealthcheck.py
rcritten: Same, it's expected.

rcritten:
This looks ok to me with one minor thing. I think the timeout for the nightly tests will need to be bumped up to accommodate the new test class, maybe try 7200. IMHO this can be done at the same time that the temp commit is dropped to avoid another run cycle.

@flo-renaud do you agree?

flo-renaud:
> This looks ok to me with one minor thing. I think the timeout for the nightly tests will need to be bumped up to accommodate the new test class, maybe try 7200. IMHO this can be done at the same time that the temp commit is dropped to avoid another run cycle.
> 
> @flo-renaud do you agree?

Yes, the new test takes ~30min and test-Ipahealthcheck used to need 1h, so 7200 should be OK.

flo-renaud:
Tests are green, all comments were addressed, ACK.

flo-renaud:
master:

* 400ef3aa2c1ea74da22547aa7ca1c20fdc6cb211 ipatests: Tests for ipahealthcheck tool with IPA external
* 7642ce35828025cd292f0e7785b93455ae4684d2 Modified nightly YAML files to include ipa-healthcheck ExternalCA Tests

flo-renaud:
@menonsudhir 
A manual backport is required for ipa-4-8 branch, can you do it? thanks

menonsudhir:
> @menonsudhir
> A manual backport is required for ipa-4-8 branch, can you do it? thanks

@flo-renaud have raised a manual backport.
https://github.com/freeipa/freeipa/pull/5034

PR #4888, opened 2020-07-06
ipatests: Tests to check profile and issue field is displayed for getcert request

menonsudhir:
test_getcert_list_profile
This test checks that the cert request generated using  getcert utility which is placed in /var/lib/certmonger/requests
directory displays profile name and issuer fields
    
test_getcert_list_profile_using_subca
This test checks that the cert request generated with -X as  subca and -T <profilename> displays correct profilename

### DIFF
--- a/ipatests/test_integration/test_cert.py
+++ b/ipatests/test_integration/test_cert.py
@@ -127,6 +127,95 @@ class TestInstallMasterClient(IntegrationTest):
         ipaddrs = ext.value.get_values_for_type(x509.IPAddress)
         assert ipaddrs == [ipaddress.ip_address(self.clients[0].ip)]
 
+    def test_getcert_list_profile(self):
+        """
+        Test that getcert list command displays the profile
+        for the cert
+        """
+        result = self.master.run_command(
+            ["getcert", "list", "-f", paths.HTTPD_CERT_FILE]
+        )
+        assert "profile: caIPAserviceCert" in result.stdout_text
+        result = self.master.run_command(
+            ["getcert", "list", "-n", "Server-Cert cert-pki-ca"]
+        )
+        assert "profile: caServerCert" in result.stdout_text
+
+    @pytest.fixture
+    def test_subca_certs(self):
+        """
+        Fixture to add subca, stop tracking request,
+        followed by removing SUB CA along with
+        cert keys
+        """
+        sub_name = "CN=SUBCA"
+        tasks.kinit_admin(self.master)
+        self.master.run_command(
+            ["ipa", "ca-add", "mysubca", "--subject={}".format(sub_name)]
+        )
+        self.master.run_command(
+            [
+                "ipa",
+                "caacl-add-ca",
+                "hosts_services_caIPAserviceCert",
+                "--cas=mysubca",
+            ]
+        )
+        yield
+        self.master.run_command(
+            ["getcert", "stop-tracking", "-i", "test-request"]
+        )
+        self.master.run_command(["ipa", "ca-disable", "mysubca"])
+        self.master.run_command(["ipa", "ca-del", "mysubca"])
+        self.master.run_command(
+            ["rm", "-fv", "/etc/pki/tls/private/test.key"]
+        )
+        self.master.run_command(["rm", "-fv", "/etc/pki/tls/certs/test.pem"])
+
+    def test_getcert_list_profile_using_subca(self, test_subca_certs):
+        """
+        Test that getcert list command displays the profile
+        for the cert requests generated, with a SubCA configured
+        on the IPA server.
+        """
+        cmd_arg = [
+            "getcert",
+            "request",
+            "-c",
+            "ipa",
+            "-I",
+            "test-request",
+            "-k",
+            "/etc/pki/tls/private/test.key",
+            "-f",
+            "/etc/pki/tls/certs/test.pem",
+            "-D",
+            self.master.hostname,
+            "-K",
+            "host/%s" % self.master.hostname,
+            "-N",
+            "CN={}".format(self.master.hostname),
+            "-U",
+            "id-kp-clientAuth",
+            "-X",
+            "mysubca",
+            "-T",
+            "caIPAserviceCert",
+        ]
+        result = self.master.run_command(cmd_arg)
+        assert (
+            'New signing request "test-request" added.\n' in result.stdout_text
+        )
+        status = tasks.wait_for_request(self.master, "test-request", 50)
+        if status == "MONITORING":
+            result = self.master.run_command(
+                ["getcert", "list", "-i", "test-request"]
+            )
+            assert "profile: caIPAserviceCert" in result.stdout_text
+        else:
+            raise AssertionError("certmonger request is "
+                                 "in state {}". format(status))
+
 
 class TestCertmongerInterruption(IntegrationTest):
     num_replicas = 1


ipatests/test_integration/test_cert.py
```diff
@@ -127,6 +128,84 @@ def test_ipa_getcert_san_aci(self):
...
+            )
+            assert "profile: caSubsystemCert" in result.stdout_text
+            assert "profile: caSignedLogCert" in result.stdout_text
+            assert "profile: caOCSPCert" in result.stdout_text
+            assert "profile: caSubsystemCert" in result.stdout_text
+            assert "profile: caCACert" in result.stdout_text
+            assert "profile: caServerCert" in result.stdout_text
+            assert "profile: caIPAserviceCert" in result.stdout_text
+            assert "profile: caIPAserviceCert" in result.stdout_text
+            assert "profile: KDCs_PKINIT_Certs" in result.stdout_text
```
rcritten, line 148: A couple of things. If this is really cycling through the requests then only one profile will be in each execution.

And it is not reliable that the filename will match the request id.

We're testing here whether certmonger displays the profile not that the profile is set for all IPA certs so I think you can just pick one and use that, something like:

getcert list -f HTTPD_CERT_FILE

It will be caIPAserviceCert

If you want to check them all then you'll need to collect something to query them on. You could collect them in a similar way that healthcheck does but that isn't very black box-like.

Otherwise, I think, you'll have to walk through CERTMONGER_REQUESTS_DIR collecting the id and then iterate on those and do a profile in [list of possibilities].

ipatests/test_integration/test_cert.py
```diff
@@ -127,6 +128,84 @@ def test_ipa_getcert_san_aci(self):
...
+        result = self.master.run_command(cmd_arg)
+        assert (
+            'New signing request "test-request" added.\n' in result.stdout_text
+        )
+        status = tasks.wait_for_request(self.master, test-request, 50)
+        if status == "MONITORING":
+            result = self.master.run_command(
+                ["getcert", "list", "-i", "test-request"]
+            )
+            assert "profile: caIPAserviceCert" in result.stdout_text
```
rcritten, line 201: if status is not MONITORING then the test will pass. You'll need to do an assert at the end of the test to raise an issue if not.

If this test fails then it is going to leave the request and the CA behind which could cause other tests to fail. I suspect this is a good case for a fixture.

ipatests/test_integration/test_cert.py
```diff
@@ -127,6 +128,84 @@ def test_ipa_getcert_san_aci(self):
         ipaddrs = ext.value.get_values_for_type(x509.IPAddress)
         assert ipaddrs == [ipaddress.ip_address(self.clients[0].ip)]
 
+    def test_getcert_list_profile(self):
+        """
+        Test that getcert list command displays the profile
+        for the requests
+        """
+        for requests in os.listdir(paths.CERTMONGER_REQUESTS_DIR):
```
mrizwan93, line 136: `os.listdir()` will be run on runner and not on `master`.

ipatests/test_integration/test_cert.py
```diff
@@ -127,6 +128,84 @@ def test_ipa_getcert_san_aci(self):
...
+
+    def test_getcert_list_profile_using_subca(self):
+        """
+        Test that getcert list command displays the profile
+        for the request with a SUBCA configured on IPA server
+        """
+        sub_name = "CN=SUBCA"
+        tasks.kinit_admin(self.master)
+        self.master.run_command(
+            ["ipa", "ca-add", "mysubca", "--subject=" + sub_name]
```
mrizwan93, line 158: It would be nice to use `"%s" % (var)` instead of concatenation.

ipatests/test_integration/test_cert.py
menonsudhir: Modified to check only specific cert e.g HTTPD_CERT_FILE

menonsudhir:
@rcritten i don't see the profile attribute for the certificate request being monitored, the certmonger installed is 
certmonger-0.79.9-1.fc32.x86_64. 
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 Request ID 'test-request':
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	status: MONITORING
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	stuck: no
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	key pair storage: type=FILE,location='/etc/pki/tls/private/test.key'
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	certificate: type=FILE,location='/etc/pki/tls/certs/test.pem'
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	CA: IPA
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	issuer: CN=SUBCA
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	subject: CN=master.ipa.test,O=IPA.TEST
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	expires: 2022-07-10 14:54:28 UTC
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	dns: master.ipa.test
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	principal name: host/master.ipa.test@IPA.TEST
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	key usage: digitalSignature,nonRepudiation,keyEncipherment,dataEncipherment
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	eku: id-kp-serverAuth,id-kp-clientAuth
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	pre-save command: 
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	post-save command: 
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	track: yes
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd59:transport.py:557 	auto-renew: yes

ipatests/test_integration/test_cert.py
rcritten: In fact you can just do [..., "--subject", sub_name]

ipatests/test_integration/test_cert.py
```diff
@@ -127,6 +128,91 @@ def test_ipa_getcert_san_aci(self):
...
+            ]
+        )
+        yield
+        self.master.run_command(
+            ["getcert", "stop-tracking", "-i", "test-request"]
+        )
+        self.master.run_command(["ipa", "ca-disable", "mysubca"])
+        self.master.run_command(["ipa", "ca-del", "mysubca"])
+        self.master.run_command(
+            ["rm", "-frv", "/etc/pki/tls/private/test.key"]
```
rcritten, line 172: You don't need -r since you're removing a single file. Same below.

ipatests/test_integration/test_cert.py
```diff
@@ -127,6 +128,91 @@ def test_ipa_getcert_san_aci(self):
...
+            "-k",
+            "/etc/pki/tls/private/test.key",
+            "-f",
+            "/etc/pki/tls/certs/test.pem",
+            "-D",
+            self.master.hostname,
+            "-K",
+            "host/%s" % self.master.hostname,
+            "-N",
+            "CN="+self.master.hostname,
```
rcritten, line 197: Please use a %s format at least instead of concatenation.

ipatests/test_integration/test_cert.py
rcritten: You still need an else: <force failure via some sort of assert>

rcritten:
The test is failing because it isn't picking up the latest certmonger release. It is still using certmonger-0.79.9-1.fc32.x86_64 which lacks the updated package. I don't know if this is an image problem or not. The package was pushed to stable today so maybe this will clear itself up shortly.

ipatests/test_integration/test_cert.py
menonsudhir: removed the -r option

ipatests/test_integration/test_cert.py
menonsudhir: @mrizwan93 the code was removed.

ipatests/test_integration/test_cert.py
menonsudhir: Removed the concatenation

ipatests/test_integration/test_cert.py
menonsudhir: @mrizwan93 have made the change.

ipatests/test_integration/test_cert.py
menonsudhir: Added the else statement as well.

rcritten:
certmonger is included with the template. Once an updated template is available this can be rebased and it should pass. Azure passes because it pulls from updates.

menonsudhir:
@rcritten seems that the template isn't updated.

flo-renaud:
@menonsudhir 
I just pushed the PR #4911 that updates the template with certmonger-0.79.11-2. You can rebase and it will pick the good template.

menonsudhir:
> @menonsudhir
> I just pushed the PR #4911 that updates the template with certmonger-0.79.11-2. You can rebase and it will pick the good template.

@flo-renaud thank you  have rebased.

rcritten:
Looks good, please remove the temporary commit.

flo-renaud:
Tests are green, ACK.

flo-renaud:
master:

* 8e05a8a8da6eea7dcad48ec5b77631e4495e4879 ipatests: Tests to check profile is displayed for getcert request.

PR #4904, opened 2020-07-09
Add ipaplatforms for containers

tiran:
Fixes: https://pagure.io/freeipa/issue/8401

###  Allow to override ipaplatform with env var

The ipaplatform provider module can now be overriden by setting
IPAPLATFORM_OVERRIDE environment variable.

### Add ipaplatform for Fedora and RHEL container 

Container platforms for Fedora and RHEL simplify FreeIPA container
effort. Paths are based on patches from https://github.com/freeipa/freeipa-container

### Write state dir to smb.conf

smb.conf now sets state and cache directory, then includes the registry.
This also allows us to write the final smb.conf before importing
remaining settings into the Samba registry.


abbra:
```
[ipatests.pytest_ipa.integration.host.Host.master.cmd55]   [error] FileNotFoundError: [Errno 2] No such file or directory: '/usr/share/ipa/smb.conf.registry.template'

```

abbra:
`Makefile.am` needs to be modified to include the new template file.

tiran:
> ```
> [ipatests.pytest_ipa.integration.host.Host.master.cmd55]   [error] FileNotFoundError: [Errno 2] No such file or directory: '/usr/share/ipa/smb.conf.registry.template'
> ```

Already fixed
I forgot to add the new template to ``Makefile.am``

rcritten:
This looks sane to me but my container usage is limited. Perhaps @Tiboris can look at it.

pvoborni:
We had a meeting with @tiran and Alejandro about how to handle PRs like this. The agreement was that:
* FreeIPA core team will review it from perspective that the PR is good for long-term maintenance and works well also with traditional workflows
* other OCP team member will review the PR from perspective if it is good for the container workflow.

Having ACKs form both parties means it is good to go.

fcami:
After review: ACK from me (IPA side).

frasertweedale:
I will review from container side.

frasertweedale:
@tiran I just want to confirm that there are no upgrade considerations resulting from the samba config-related changes.  If you could confirm that, then it's an ACK from me.

abbra:
We don't do upgrade of `smb.conf` -- we assume it is configured already and only work with the registry-based configuration afterwards. This means that containers which were done without the fix in freeipa-container will not work with Samba anyway, so that is not a supported situation. Containers that have the fix we are upstreaming here will work transparently because their registry configuration will be accessible to the upgrade code.

abbra:
ACK

tiran:
master:

* eec5c9d8206f8ca612677369fdd3fce39205fa32 Allow to override ipaplatform with env var
* 02986ff42b97dceb689abb32cf937da993880ddd Add ipaplatform for Fedora and RHEL container
* 64b20aad283c6a0a9557e77993b09679a72efa88 Write state dir to smb.conf
* 664007e031bec3b1c6f4a9518f427e139b026a1e Explicitly pass keytab to ipa-join
* 8f6502db03576c92b25f4460fb98d75e0fe68f28 Convert ipa-httpd-pwdreader into Python script

PR #4907, opened 2020-07-09
Don't delegate the TGT in ipa-join

rcritten:
Pre 3.0.0 IPA delegated the TGT to enforce access control in
389-ds. At the point that S4U2Proxy support was added there
were still IPA 2.0.x servers in use so this delegation was
left in place in ipa-join so that enrollment would work.

Those days are long gone, remove that support in the XML and
JSON RPC requests.

https://pagure.io/freeipa/issue/8405


flo-renaud:
Hi @rcritten 
thanks for the patch, LGTM.

rcritten:
master:

* 28caa22a8e5b1fd402692cdc41fa4dd4a73f6698 Don't delegate the TGT in ipa-join

rcritten:
Removed ipa-4-8 label, the JSON change is going into ipa-next only. No backport needed.

PR #4908, opened 2020-07-10
Support changelog integrated into main database

vashirov:
389-ds will integrate changelog into main database:
https://pagure.io/389-ds-base/issue/49562
http://www.port389.org/docs/389ds/design/integrate-changelog-database-and-backend-database.html

It also changes the location of the configuration from a single entry
`cn=changelog5,cn=config` to an entry for each backend in
`cn=changelog,<backend>,cn=ldbm database,cn=plugins,cn=config`

This breaks IPA in a few places, in particular:

* `setup_changelog()`
Here the expected failure is err=68, it would be trivial to add err=53.
A better way would be to reuse lib389 for setting up changelog,
not sure how much refactoring that would involve.

* `update_unhashed_password()`
Since the entry no longer exists, log message would be affected.

* `update scripts`
nsslapd-changelogmaxage now needs to be set for each backend.

This fix updates the locations of a new changelog. 

Related: https://pagure.io/freeipa/issue/8407
Signed-off-by: Viktor Ashirov <vashirov@redhat.com>

### DIFF
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -584,7 +584,7 @@ class ReplicationManager:
         )
         try:
             conn.add_entry(entry)
-        except errors.DuplicateEntry:
+        except (errors.DuplicateEntry, errors.DatabaseError):
             return
 
     def _finalize_replica_settings(self, conn):
--- a/install/updates/20-replication.update
+++ b/install/updates/20-replication.update
@@ -64,5 +64,8 @@ default: nsslapd-pluginVendor: none
 default: nsslapd-pluginDescription: none
 
 # Set replication changelog limit (#5086)
-dn: cn=changelog5,cn=config
+dn: cn=changelog,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+addifnew: nsslapd-changelogmaxage: 7d
+
+dn: cn=changelog,cn=ipaca,cn=ldbm database,cn=plugins,cn=config
 addifnew: nsslapd-changelogmaxage: 7d
--- a/ipaserver/install/plugins/update_unhashed_password.py
+++ b/ipaserver/install/plugins/update_unhashed_password.py
@@ -79,7 +79,10 @@ class update_unhashed_password(Updater):
             # Log a warning that changelog will contain sensitive data
             try:
                 cldb_e = ldap.get_entry(
-                    DN(('cn', 'changelog5'),
+                    DN(('cn', 'changelog'),
+                       ('cn', 'userRoot'),
+                       ('cn', 'ldbm database'),
+                       ('cn', 'plugins'),
                        ('cn', 'config')),
                     ['nsslapd-changelogdir'])
                 cldb = cldb_e.single_value.get("nsslapd-changelogdir")


vashirov:
I'd need some help to test this with 389-ds-base build from https://pagure.io/389-ds-base/pull-request/51181 as this will most certainly fail on a latest stable build.
I'm also looking for comments on how to handle this migration better. Perhaps add checks for 389-ds-base versions to support both old and new changelog locations.
Thanks.

ipaserver/install/replication.py
```diff
@@ -584,7 +584,7 @@ def setup_changelog(self, conn):
         )
         try:
             conn.add_entry(entry)
-        except errors.DuplicateEntry:
+        except (errors.DuplicateEntry, errors.DatabaseError):
```
rcritten, line 587: What would cause a DataBaseError? This would ignore the error and try to continue.

rcritten:
In this case setting the Requires to the new version of 389-ds would prevent new installs using the old location. There would be nothing to protect existing users who only upgrade [free]ipa-* packages I think.

ipaserver/install/replication.py
vashirov: Previously adding `cn=changelog5,cn=config` would return err=68 if it already exist, but with the change in https://pagure.io/389-ds-base/pull-request/51181#_15__191 it will return err=53, which is `errors.DatabaseError`, because separate changelog is no longer supported.
See also discussion in https://pagure.io/389-ds-base/pull-request/51101#comment-122648

tbordaz:
Note that err=53 is the RC of DS when trying to ADD 'cn=changelog5,cn=config'. DS returns unwilling_to_perform that is translated into databaseError. The failure is not related to a problem with the DS database layer and that DS is out of order.

rcritten:
What about existing installations that currently use cn=changelog5,cn=config? Do those need to be moved?

rcritten:
Also, what can I do to verify that the changes aren't breaking things? Just check that replication is working?

tbordaz:
> What about existing installations that currently use cn=changelog5,cn=config? Do those need to be moved?

During upgrade, when the existed instance is restarted with the new software version, the old changelog is upgraded.

When the new version of replication plugins starts it detects it exists an old config ('cn=changelog5,cn=config') it then upgrades the config (a config per backend), moves the changelog database file to the appropriate new location and finally removes old config and old changelog location.

When upgrading the config, all old config options (common to all backends) are propagated to the new configs (per backends). All except nsslapd-changelogdir that is now using the same location as the backend database directory.

tbordaz:
> Also, what can I do to verify that the changes aren't breaking things? Just check that replication is working?

From freeipa pov, one test is setting up server/replica and checking that replication both way are working fine.
I think some administrative operations should be good also like:

- Stop master, update replica, restart master (and the opposite)
- Check that backup/restore works and that after backup/restore the server can replicate (in/out)
- export the changelog (if this option exists in IPA).

Note that backups taken before the upgrade can not be restored after the upgrade

rcritten:
The backup is important to highlight, thanks for raising it. Conceivably a user could update only 389 (say to pick up some specific fix), backup, then restore on a new machine without the updated 389 (users always find a way). This would present some risks to IPA, right? I wonder if we'll eventually need a 389 Conflicts for IPA < some version.

I'll do some additional replication testing, otherwise this seems fine since it seems that 389 is doing most of the heavy lifting.

tbordaz:
> The backup is important to highlight, thanks for raising it. Conceivably a user could update only 389 (say to pick up some specific fix), backup, then restore on a new machine without the updated 389 (users always find a way). This would present some risks to IPA, right? I wonder if we'll eventually need a 389 Conflicts for IPA < some version.
> 
> I'll do some additional replication testing, otherwise this seems fine since it seems that 389 is doing most of the heavy lifting.

That are good news that both 389 and freeipa patches are working fine so far.

Regarding the backup something is not clear to me. a 389 backup is basically a copy of the database (db files and txn logs) and is used to restore the instance where the backup was taken. Are Freeipa backups used to restore others instances ?
I can imagine that an admin who wants to upgrade a server, takes a backup, then upgrade DS. If something gets wrong the admin hopes to use the backup to restore the instance. I think this procedure may not work as config has been upgraded and will not match the backup. I think we should suggest an export (with replicated data) before upgrading a server.

rcritten:
IPA does ldif dump and import for backup and does require that on restore the IPA version matches the backup, so chances are we'll be ok as long as we have the Requires setup properly (will need a note somewhere to remember to bump things).

It seems like we need Requires on both sides to ensure that mixing of versions doesn't happen. It shouldn't be a problem as long as we coordinate Fedora releases.

This PR is set for ipa-next will means the next major release of IPA which is TBD. It will not appear in any IPA 4.8.x release. Is that ok?

marcus2376:
> IPA does ldif dump and import for backup and does require that on restore the IPA version matches the backup, so chances are we'll be ok as long as we have the Requires setup properly (will need a note somewhere to remember to bump things).
> 
> It seems like we need Requires on both sides to ensure that mixing of versions doesn't happen. It shouldn't be a problem as long as we coordinate Fedora releases.
> 
> This PR is set for ipa-next will means the next major release of IPA which is TBD. It will not appear in any IPA 4.8.x release. Is that ok?

Whatever lands in Fedora 33 is where we need this merged

marcus2376:
@rcritten if an update in 20-replication.update fails does it break anything or does it continue? Not sure how and where these files are processed.  This is the only challenge remaining to make this PR work in both versions of DS :-)

abbra:
@marcus2376 I suspect you need to replace an update file change with a dynamic plugin that decides which entry to insert depending on the configuration it discovers.

E.g., something like this:
* remove changelog entry from `20-replication.update`
* add an update plugin to `ipaserver/install/plugins/` that handles choice of the changelog entry content
* add name of the plugin class to `install/updates/05-pre_upgrade_plugins.update` so that it is run before normal `NN-*.update` files. You can add to any position, as this doesn't matter where it should be, I guess.

For examples of the plugins see `ipaserver/install/plugins/`. Perhaps the simplest is `update_pacs.py` which looks up an entry and updates an attribute there.

vashirov:
Closing this PR as @marcus2376 will open a new one.

PR #4909, opened 2020-07-10
Add unauthenticated nsupdate

fcami:
ipa-client-install: update sssd.conf if nsupdate requires -g
    
    If dynamic DNS updates are selected, sssd will use GSS-TSIG
    by default for nsupdate.
    When ipa-client-install notices that plain nsupdate is required,
    switch sssd to use no authentication for dynamic updates too.
    
    Fixes: https://pagure.io/freeipa/issue/8402

+

    ipa-client-install: invoke nsupdate twice (GSS-TSIG, plain)
    
    ipa-client-install invokes nsupdate with GSS-TSIG at client
    enrollment time. If that fails, no retry is done.
    Change that behavior to try again without GSS-TSIG.
    
    Fixes: https://pagure.io/freeipa/issue/8402

####
This is purely WIP as it needs a proper test, however in my manual testing it behaves like it should (sssd configuration included).

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -1370,6 +1370,11 @@ def do_nsupdate(update_txt):
         result = True
     except CalledProcessError as e:
         logger.debug('nsupdate failed: %s', str(e))
+        try:
+            ipautil.run([paths.NSUPDATE, UPDATE_FILE])
+            result = True
+        except CalledProcessError as e:
+            logger.debug('Unauthenticated nsupdate failed: %s', str(e))
 
     try:
         os.remove(UPDATE_FILE)
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -1369,9 +1369,35 @@ def do_nsupdate(update_txt):
         ipautil.run([paths.NSUPDATE, '-g', UPDATE_FILE])
         result = True
     except CalledProcessError as e:
-        logger.debug('nsupdate failed: %s', str(e))
+        logger.debug('nsupdate (GSS-TSIG) failed: %s', str(e))
         try:
             ipautil.run([paths.NSUPDATE, UPDATE_FILE])
+            try:
+                sssdconfig = SSSDConfig.SSSDConfig()
+                sssdconfig.import_config()
+                domains = sssdconfig.list_active_domains()
+                for name in domains:
+                    domain = sssdconfig.get_domain(name)
+                    try:
+                        provider = domain.get_option('id_provider')
+                    except SSSDConfig.NoOptionError:
+                        continue
+                    if name == api.env.domain and provider == "ipa":
+                        try:
+                            if domain.get_option('dyndns_update') is True:
+                                domain.set_option('dyndns_auth', 'none')
+                                sssdconfig.save_domain(domain)
+                                sssdconfig.write(paths.SSSD_CONF)
+                                break
+                        except SSSDConfig.NoOptionError:
+                            break
+            except Exception as e:
+                logger.debug('Unable to update SSSD configuration: %s', str(e))
+                logger.info(
+                    'Failed to configure SSSD for unauthenticated DNS '
+                    'dynamic updates. SSSD might be unable to update DNS '
+                    'entries for this host.'
+                )
             result = True
         except CalledProcessError as e:
             logger.debug('Unauthenticated nsupdate failed: %s', str(e))
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -1363,7 +1363,6 @@ def do_nsupdate(update_txt):
 
     with open(UPDATE_FILE, "w") as f:
         f.write(update_txt)
-        ipautil.flush_sync(f)
 
     result = False
     try:
--- a/client/man/ipa-client-install.1
+++ b/client/man/ipa-client-install.1
@@ -205,7 +205,11 @@ Create DNS A/AAAA record for each IP address on this host.
 Configure SSSD to permit all access. Otherwise the machine will be controlled by the Host\-based Access Controls (HBAC) on the IPA server.
 .TP
 \fB\-\-enable\-dns\-updates\fR
-This option tells SSSD to automatically update DNS with the IP address of this client.
+This option tells SSSD to automatically update DNS with the IP address of this
+client.
+The default is to use GSS-TSIG. However, if using GSS-TSIG fails for any reason
+at install time, \fBipa\-client\-install\fR will configure SSSD to use
+unauthenticated nsupdates instead.
 .TP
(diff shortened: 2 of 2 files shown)


ipaclient/install/client.py
```diff
@@ -1337,9 +1337,29 @@ def do_nsupdate(update_txt):
...
+                    domain = sssdconfig.get_domain(name)
+                    try:
+                        provider = domain.get_option('id_provider')
+                    except SSSDConfig.NoOptionError:
+                        continue
+                    if name == api.env.domain and provider == "ipa":
+                        if domain.get_option('dyndns_update') is True:
+                            # dyndns_update is set by configure_sssd_conf()
+                            domain.set_option('dyndns_auth', 'none')
+                            sssdconfig.save_domain(domain)
```
rcritten, line 1358: You can add a break here. Once the ipa domain is found there is no need to process any others.

ipaclient/install/client.py
```diff
@@ -1337,9 +1337,29 @@ def do_nsupdate(update_txt):
...
+                sssdconfig.write(paths.SSSD_CONF)
+                sssd.restart()
+            except Exception:
+                logger.debug('Unable to update SSSD configuration: %s', str(e))
```
rcritten, line 1362: There should be a message to the user here too because if the SSSD config isn't updated then all subsequent DNS updates are going to fail.

ipaclient/install/client.py
fcami: Done, thanks

ipaclient/install/client.py
fcami: Added, thanks

fcami:
Note to self:
- use the master_3client topology
- install IPA with DNS
- configure BIND on one client with:
```
zone "ipa.test" {
     type master;
     file "ipa.test.db";
     allow-update { <network_range>/24;};
};
```
- use "ipa dns-update-system-records --all" to find all necessary records to add to the zone file
- add the first server's A record to the zone file
- start named (so that's an external DNS)
- point the two other clients' resolv.conf to the external DNS
- on those clients, run ipa-client-install, with and without --enable-dns-updates
- double-check logs on both the DNS and clients
- double-check sssd configuration on both clients

abbra:
@fcami what is a progress here? Is it ready for review?

fcami:
@abbra not at all, there are no associated tests.

fcami:
In fact testing this is quite cumbersome as it requires a separate BIND instance to manage the IPA DNS.
The change is small anyway.
Flagging as ready for review.

ssidhaye:
Since this requires elaborate env setup which cannot be run in PRCI, I am OK with having this change without a test. A test would be added downstream

ipaclient/install/client.py
```diff
@@ -1370,9 +1370,35 @@ def do_nsupdate(update_txt):
...
+                    if name == api.env.domain and provider == "ipa":
+                        try:
+                            if domain.get_option('dyndns_update') is True:
+                                domain.set_option('dyndns_auth', 'none')
+                                sssdconfig.save_domain(domain)
+                                sssdconfig.write(paths.SSSD_CONF)
+                                break
+                        except SSSDConfig.NoOptionError:
+                            break
+            except Exception:
```
rcritten, line 1395: e in this case is the previous exception, not the one most recently raised.

ipaclient/install/client.py
```diff
@@ -1370,9 +1370,35 @@ def do_nsupdate(update_txt):
         ipautil.run([paths.NSUPDATE, '-g', UPDATE_FILE])
         result = True
     except CalledProcessError as e:
-        logger.debug('nsupdate failed: %s', str(e))
+        logger.debug('nsupdate (GSS-TSIG) failed: %s', str(e))
         try:
             ipautil.run([paths.NSUPDATE, UPDATE_FILE])
+            try:
+                sssdconfig = SSSDConfig.SSSDConfig()
```
rcritten, line 1377: It almost feels like this whole sssdconfig part should be in an else:

ipaclient/install/client.py
```diff
@@ -1364,7 +1364,6 @@ def do_nsupdate(update_txt):
 
     with open(UPDATE_FILE, "w") as f:
         f.write(update_txt)
-        ipautil.flush_sync(f)
```
rcritten, line 1367: Doesn't this ensure that the update file is written to disk and not cached so that it is properly passed to nsupdate?

fcami: The flush is unnecessary because the file descriptor will be closed after the compound statement.
The fsync is unnecessary too: it would be necessary to ensure the file's content after a hard reboot, but we do not care about that one ; if the box is forcefully rebooted during this stage, the enrollment will have failed anyway.

ipaclient/install/client.py
fcami: Fixed, I guess.

ipaclient/install/client.py
fcami: or part of it, but I'm wary of changing the logic there.

flo-renaud (commented):
Hi @fcami 
the change LGTM. My only comment is related to the documentation of this new behavior. I think we should at least expand a bit the ipa-client-install man page (there is a section related to the `--enable-dns-updates` option, where we could say that the default is to use GSS-TSIG but we fallback to unauthenticated nsupdates, both during the install but also in whenever the local IP @ changes and sssd updates the DNS records).
The documentation will also need to be updated, for instance https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/installing_identity_management/index#installing-client-basic-installing-a-client-by-using-user-credentials-interactive-installation:
```
 Add the --enable-dns-updates option to update the DNS records with the IP address of the client system if either of the following conditions applies:

    The IdM server the client will be enrolled with was installed with integrated DNS
    The DNS server on the network accepts DNS entry updates with the GSS-TSIG protocol 
```

flo-renaud:
Hi @fcami,
thanks for the update, the man page diff lgtm.

rcritten:
LGTM

abbra:
We seem to have missed this PR. It is acked and ready for push. Is that correct?

fcami:
@abbra from my PoV this is correct yes.

abbra:
master:

* 72f44b5596027a42fc6038b5f8dcd9e6c3854f73 ipa-client-install: remove fsync in do_nsupdate()
* 20c7bd5eba6a8be36abf5dfd03daf235ee113d1d ipa-client-install: invoke nsupdate twice (GSS-TSIG, plain)
* 2e31e8479a94e75ad33cf50f6eee106283b44b0d ipa-client-install: update sssd.conf if nsupdate requires -g

PR #4910, opened 2020-07-10
 ipa-join: handle JSON-RPC error codes 

Carbenium:
Handling of IPA error codes was missing from the JSON-RPC code paths.
Add parsing support for the `error` object and handle `ACIError` explicitly to match XML-RPC behaviour.

Fixes: https://pagure.io/freeipa/issue/8408

### DIFF
--- a/client/ipa-join.c
+++ b/client/ipa-join.c
@@ -811,12 +811,47 @@ cleanup:
 }
 
 static int
+jsonrpc_parse_error(json_t *j_error_obj) {
+    int rval = 0;
+
+    json_error_t j_error;
+
+    int error_code = 0;
+    char *error_message = NULL;
+    if (json_unpack_ex(j_error_obj, &j_error, 0, "{s:i, s:s}",
+                       "code", &error_code,
+                       "message", &error_message) != 0) {
+        if (debug)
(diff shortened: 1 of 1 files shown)


rcritten:
fedora-latest/test_advise failed due to timeout, fedora-latest/test_replica_promotion_TestSubCAkeyReplication with CA spawn issue. Neither related to join. Curious they both failed twice.

flo-renaud:
Hi @Carbenium 
thanks for the fix, works for me. I manually ran test_integration/test_replica_promotion.py::TestUnprivilegedUserPermissions and it succeeded, also tested manually:
- [x] ipa-client-install with non-admin user produces the error msg `Joining realm failed: No permission to join this host to the IPA domain.`
- [x] promotion using ipa-replica-install with non-admin user produces the error msg `Insufficient privileges to promote the server.`

flo-renaud:
master:

* 4696644f3fc996da2a88d91d29af11a56ecae0c3 ipa-join: extract common JSON-RPC response parsing to common function
* 6dfefc97459376a2ac4fb079a19b49e70056fd14 ipa-join: handle JSON-RPC error codes

PR #4913, opened 2020-07-10
vault.py: flush the certificate after renaming

fcami:
store_cert() does f.write, flush_sync, f.close, os.rename(f.name)
in that order. The directory is not flushed.
* get rid of f.close as this is taken care of by the nested block
* rename the file first then use flush_sync(f)


abbra:
@fcami this PR seems to have stuck in PR CI. Could you please rebase and re-push?

PR #4914, opened 2020-07-12
replica: Ensure that ipaapi user is allowed to access ifp

jsf9k:
`ipa-server-install` executes `ipa-client-install` with the `--on-master` flag set, which causes the `ipaclient.install.client.sssd_enable_ifp()` function to be called.  This function configures `sssd` so that the `ipaapi` user is allowed to access `ifp`.  Any FreeIPA replica should also have `sssd` configured like this, but in that case we cannot simply pass the `--on-master` flag to `ipa-client-install` because it has other side effects.  The solution is to call the `ipaclient.install.client.sssd_enable_ifp()` function from inside the `ipaserver.install.server.replicainstall.promote_sssd()` function.

https://pagure.io/freeipa/issue/8403

### DIFF
--- a/ipatests/test_integration/test_replica_promotion.py
+++ b/ipatests/test_integration/test_replica_promotion.py
@@ -47,6 +47,21 @@ class ReplicaPromotionBase(IntegrationTest):
         assert(found > 0), result2.stdout_text
 
 
+def sssd_config_allows_ipaapi_access_to_ifp(host):
+    """Checks that the sssd configuration allows the ipaapi user to access
+    ifp
+
+    :param host the machine on which to check that sssd allows ipaapi
+    access to ifp
+    """
+    with tasks.remote_sssd_config(host) as sssd_conf:
+        ifp = sssd_conf.get_service('ifp')
+        uids = [
+            uid.strip() for uid in ifp.get_option('allowed_uids').split(',')
+        ]
+        assert 'ipaapi' in uids
+
+
 class TestReplicaPromotionLevel1(ReplicaPromotionBase):
     """
     TestCase: http://www.freeipa.org/page/V4/Replica_Promotion/Test_plan#
@@ -99,6 +114,16 @@ class TestReplicaPromotionLevel1(ReplicaPromotionBase):
         # Ensure that pkinit is properly configured, test for 7566
         result = self.replicas[0].run_command(['ipa-pkinit-manage', 'status'])
         assert "PKINIT is enabled" in result.stdout_text
+
+    @replicas_cleanup
+    def test_sssd_config_allows_ipaapi_access_to_ifp(self):
+        """Verify that the sssd configuration allows the ipaapi user to
+        access ifp
+
+        Test for ticket 8403.
+        """
+        for replica in self.replicas:
+            sssd_config_allows_ipaapi_access_to_ifp(replica)
 
 
 class TestUnprivilegedUserPermissions(IntegrationTest):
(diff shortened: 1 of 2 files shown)


jsf9k:
I'm happy to try and create a test case if someone can get me pointed in the right direction.

rcritten:
It would be one or more integration tests. They live in ipatests/test_integration.

Considering we're talking about replica installations test_replica_promotion.py is a candidate for at least one test. I think that the TestReplicaPromotionLevel1 could add a new test function to fetch sssd.conf via:
```
contents = host.get_file_contents(paths.SSSD_CONF, encoding='utf-8')
```
and examine the results to ensure that the ifp settings were done.

TestUnprivilegedUserPermissions could be used to do the same ifp check as this is a client -> server promotion.

In fact, the ifp test could be some generic function that given a host fetch the file, load using SSSDConfig and verify the ifp settings.

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -47,6 +48,32 @@ def test_kra_install_master(self):
         assert(found > 0), result2.stdout_text
 
 
+def test_sssd_config_allows_ipaapi_access_to_ifp(host):
```
rcritten, line 51: Can you drop the test_ prefix? It makes it look like this is a test rather than a standalone function.

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -47,6 +48,32 @@ def test_kra_install_master(self):
...
+
+    # Verify that the ifp section contains an allow_uids item
+    ifp_items = config.items('ifp')
+    ifp_items_dict = {item[0]: item[1] for item in ifp_items}
+    assert 'allowed_uids' in ifp_items_dict
+
+    # Verify that the allow_uids item's value contains ipaapi
+    uids = [uid.strip() for uid in ifp_items_dict['allowed_uids'].split(',')]
+    assert 'ipaapi' in uids
+
```
rcritten, line 74: Only two blank lines are necessary.

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -47,6 +48,32 @@ def test_kra_install_master(self):
...
+def test_sssd_config_allows_ipaapi_access_to_ifp(host):
+    '''Tests that the sssd configuration allows the ipaapi user to access
+    ifp
+
+    :param host the machine on which to test that sssd allows ipaapi
+    access to ifp
+    '''
+    contents = host.get_file_contents(paths.SSSD_CONF, encoding='utf-8')
+    # Parse the sssd config as a config file
+    config = ConfigParser()
```
rcritten, line 60: Strictly speaking, yes, the config is an INI-style configuration file. There is also an SSSD python API and that may be a better long-term solution. Maybe something approaching:

```
        sssdconfig = SSSDConfig.SSSDConfig()
        sssdconfig.import_config()

        sssdconfig.get_service('ifp')

        uids = ifp.get_option('allowed_uids').split(',')
        assert 'ipaapi' in uids
```

jsf9k: I believe this is addressed in commit adc5e44.

ipatests/test_integration/test_replica_promotion.py
jsf9k: I believe this is addressed in commit adc5e44.

ipatests/test_integration/test_replica_promotion.py
jsf9k: I believe this is addressed in commit adc5e44.

rcritten:
The PR looks outstanding, thanks for the contribution. Triggering full CI. I don't see any other issues. Kudos on finding remote_sssd_config(), I had forgotten about that.

jsf9k:
> The PR looks outstanding, thanks for the contribution. Triggering full CI. I don't see any other issues. Kudos on finding remote_sssd_config(), I had forgotten about that.

Thanks @rcritten.  I couldn't figure out how to use `SSSDConfig()` remotely at first either, which is why I used `ConfigParser` the first time around.

rcritten:
master:

* 12529d7ef184846ca3852821b813c725b22e11c6 replica: Ensure the ipaapi user is allowed to access ifp on replicas
* 2ff1d6b4509e0214f8147a80a6ff80b429b680df replica: Add tests to ensure the ipaapi user is allowed access to ifp on replicas

jsf9k:
Thanks again for all your help @rcritten!

PR #4061, opened 2019-12-19
doc/designs: Add a design page for application-specific passwords

RichardKalinec:
This design page describes a new enhancement: application-specific
passwords and HBAC rules for them. Users will be able to have additional
passwords besides the primary one, and administrators will be able to
set HBAC rules for them specifying what systems and services will each
application-specific password have access to. Application-specific
passwords will not be usable for account management. They will also be
supported by ipa-kdb for Kerberos authentication.

https://pagure.io/freeipa/issue/4510

PR #4923, opened 2020-07-15
Add support for app passwords

RichardKalinec:
Users will be able to have additional passwords besides the primary one - app passwords. They will be usable for accessing all systems and services that his/her FreeIPA account is used for, but not to manage the account (including configuring the app passwords).

Resolves: https://pagure.io/freeipa/issue/4510
Design page and its discussion: https://github.com/freeipa/freeipa/pull/4061

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
+++ b/daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
@@ -459,7 +459,6 @@ done:
     if (pwdop) pwdop->pwdata.target = NULL;
     free_ipapwd_krbcfg(&krbcfg);
     slapi_ch_free_string(&userpw);
-    slapi_ch_array_free(rdns);
     if (rc != LDAP_SUCCESS) {
         slapi_send_ldap_result(pb, rc, NULL, errMesg, 0, NULL);
         return -1;
@@ -1456,6 +1455,12 @@ static int ipapwd_pre_bind(Slapi_PBlock *pb)
     Slapi_PBlock* search_pb;
     Slapi_Entry** search_apppw_entries = NULL;
 
+    ret = ipapwd_gen_checks(pb, &errMesg, &krbcfg, 0);
+    if (ret != 0) {
+        LOG_FATAL("ipapwd_gen_checks failed!?\n");
+        return 0;
+    }
+
     /* get BIND parameters */
     ret |= slapi_pblock_get(pb, SLAPI_BIND_TARGET_SDN, &target_sdn);
     ret |= slapi_pblock_get(pb, SLAPI_BIND_METHOD, &method);
@@ -1546,7 +1551,6 @@ static int ipapwd_pre_bind(Slapi_PBlock *pb)
         LOG_OOM();
         goto invalid_creds;
     }
-    krbcfg = ipapwd_getConfig();
     apppws_dn = slapi_sdn_get_dn(krbcfg->apppws_sdn);
 
     /* Also compare against app passwords of the user. If some of them matches,
--- a/daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
+++ b/daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
@@ -68,6 +68,12 @@
 #define IPAPWD_OP_ADD 1
 #define IPAPWD_OP_MOD 2
 
+/* Partial DN of the container for app passwords */
+#define APPPW_CONTAINER_DN "cn=apps,cn=accounts"
+
+/* Search filter to use when searching for app passwords */
+#define SEARCH_FILTER "objectclass=account"
+
 extern Slapi_PluginDesc ipapwd_plugin_desc;
 extern void *ipapwd_plugin_id;
 extern const char *ipa_realm_tree;
@@ -218,6 +224,12 @@ static int ipapwd_pre_add(Slapi_PBlock *pb)
     char *userpw = NULL;
     Slapi_DN *sdn = NULL;
     struct ipapwd_operation *pwdop = NULL;
+    const char** rdns = NULL;
+    size_t suffix_start;
+    size_t rdn_count;
+    bool suffix_found = False;
+    bool cn_apps = False;
+    bool cn_accounts = False;
     void *op;
     int is_repl_op, is_root, is_krb, is_smb, is_ipant, is_memberof;
     int ret;
@@ -379,14 +391,33 @@ static int ipapwd_pre_add(Slapi_PBlock *pb)
     pwdop->pwdata.timeNow = time(NULL);
     pwdop->pwdata.target = e;
 
-    ret = ipapwd_CheckPolicy(&pwdop->pwdata);
-    /* For accounts created by cn=Directory Manager or a passsync
-     * managers, ignore result of a policy check */
-    if ((pwdop->pwdata.changetype != IPA_CHANGETYPE_DSMGR) &&
-        (ret != 0) ) {
-        errMesg = ipapwd_error2string(ret);
-        rc = LDAP_CONSTRAINT_VIOLATION;
+    rdns = slapi_ldap_explode_dn(pwdop->pwdata.dn, 0);
+    if (!rdns) {
+        LOG_OOM();
         goto done;
+    }
+    rdn_count = sizeof(rdns) / sizeof(rdns[0]);
+    for (suffix_start = 0; suffix_start < rdn_count; suffix_start++) {
+        if (strstr(rdns[suffix_start], 'dc=') == rdns[suffix_start]) {
+            break;
+        }
+    }
+    suffix_found = suffix_start < rdn_count;
+    cn_apps = strcmp(rdns[suffix_start - 2], 'cn=apps') == 0;
+    cn_accounts = strcmp(rdns[suffix_start - 1], 'cn=accounts') == 0;
+    /* If the entry represents an app password, no password policy checks
+     * are performed
+     */
+    if (suffix_found && cn_apps && cn_accounts) {
+        ret = ipapwd_CheckPolicy(&pwdop->pwdata);
+        /* For accounts created by cn=Directory Manager or a passsync
+         * managers, ignore result of a policy check */
+        if ((pwdop->pwdata.changetype != IPA_CHANGETYPE_DSMGR) &&
+            (ret != 0)) {
+            errMesg = ipapwd_error2string(ret);
+            rc = LDAP_CONSTRAINT_VIOLATION;
+            goto done;
+        }
     }
 
     if (is_krb || is_smb || is_ipant) {
@@ -448,6 +479,7 @@ done:
     if (pwdop) pwdop->pwdata.target = NULL;
     free_ipapwd_krbcfg(&krbcfg);
     slapi_ch_free_string(&userpw);
+    slapi_ch_array_free(rdns);
     if (rc != LDAP_SUCCESS) {
         slapi_send_ldap_result(pb, rc, NULL, errMesg, 0, NULL);
         return -1;
@@ -1420,6 +1452,10 @@ static int ipapwd_pre_bind(Slapi_PBlock *pb)
         "krbPasswordExpiration", "krblastpwchange",
         NULL
     };
+    static const char* apppw_attrs_list[] = {
+        SLAPI_USERPWD_ATTR, "uid", "objectclass",
+        NULL
+    };
     struct berval *credentials = NULL;
     Slapi_Entry *entry = NULL;
     Slapi_DN *target_sdn = NULL;
@@ -1435,6 +1471,17 @@ static int ipapwd_pre_bind(Slapi_PBlock *pb)
     struct tm expire_tm;
     int rc = LDAP_INVALID_CREDENTIALS;
     char *errMesg = NULL;
+    const char** rdns = NULL;
+    size_t rdn_count;
+    size_t suffix_start;
+    size_t suffix_len = 0;
+    size_t copied = 0;
+    char *suffix = NULL;
+    char *apppw_dn = NULL;
+    size_t apppw_count = 0;
+    size_t i = 0;
+    Slapi_PBlock* search_pb;
+    Slapi_Entry** search_apppw_entries = NULL;
 
     /* get BIND parameters */
     ret |= slapi_pblock_get(pb, SLAPI_BIND_TARGET_SDN, &target_sdn);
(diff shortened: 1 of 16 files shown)


RichardKalinec:
This implementation is not yet fully finished and functional, but I put it here, so that the particular code can be discussed. I will get back to this after submitting my thesis on Sunday.

The following has yet to be implemented:
* limiting effective permissions of sessions authenticated using an app password to:
  * full access to the `apppw` object with the app password that was used for authentication
  * read-only access to the user's primary entry
* enforcement of the password generation at the LDAP level
* adding ACIs limiting the allowed values of:
  * the `uid` attribute to numbers 0 - 99 (regex pattern: `'(?!^0+[0-9]$)^[0-9]{1,2}$'`)
  * the `ou` attribute (app name) to the regex pattern `'^[a-zA-Z0-9_-]*[a-zA-Z0-9_$-]?$'`

The following may also be a problem:
* definition of the managed permissions in the `apppw` object, especially I am, for now, unable to precisely specify their bind rule type so that they would mirror the respective new ACIs in the LDAP database
* it must be tested whether race conditions can arise when renaming a user and the entry grouping his app passwords due to how targets of the new ACIs are specified

The installation of FreeIPA with this extension currently fails with the following error:
``` Update failed: "Unknown template keyword 'dn'" ```
I appended the installation log. Of course, this must be also fixed.

File: daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c [hunk 1]
```diff
@@ -379,14 +385,19 @@ static int ipapwd_pre_add(Slapi_PBlock *pb)
...
-     * managers, ignore result of a policy check */
-    if ((pwdop->pwdata.changetype != IPA_CHANGETYPE_DSMGR) &&
-        (ret != 0) ) {
-        errMesg = ipapwd_error2string(ret);
-        rc = LDAP_CONSTRAINT_VIOLATION;
-        goto done;
+    /* If the entry represents an app password, no password policy checks
+     * are performed
+     */
+    if (strstr(pwdop->pwdata.dn, 'cn=apps,cn=accounts,dc=') == NULL) {
```
rcritten, line 391: A string compare of a DN is probably not the best way to do this.

daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
```diff
@@ -1520,8 +1545,48 @@ static int ipapwd_pre_bind(Slapi_PBlock *pb)
...
     ipapwd_write_krb_keys(pb, discard_const(dn), entry, credentials);
 
+    /* Get the (user's) uid and $SUFFIX parts of the DN. */
+    uid_len = (size_t)(strchr(dn, ',') - dn);
+    strncpy(uid, dn, uid_len);
+    uid[uid_len] = '\0';
+    suffix_start = strstr(dn, ",dc=");
+    suffix_len = strlen(dn) - (size_t)(suffix_start - dn);
+    suffix = malloc(suffix_len + 1);
+    strncpy(suffix, suffix_start, suffix_len);
```
rcritten, line 1555: A more robust way to do this would be slapi_ldap_explode_dn() to break the DN into RDNs and operate on those.

daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
```diff
@@ -1520,8 +1545,48 @@ static int ipapwd_pre_bind(Slapi_PBlock *pb)
...
+
+    /* Also compare against app passwords of the user. If some of them matches,
+     * the user is authenticated using that password and he/she will have
+     * read-only access to his primary entry. No operations with OTP or
+     * Kerberos are performed for app passwords.
+     */
+    for (i = 0; i < APPPW_MAX_COUNT; i++) {
+        sprintf(apppw_dn, "%d,%s,%s%s", i, uid, APPPW_CONTAINER_DN, suffix);
```
rcritten, line 1563: slapi_ch_smprintf() or asprintf() is preferred. and check that an OOM didn't happen.

daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
```diff
@@ -68,6 +68,12 @@
 #define IPAPWD_OP_ADD 1
 #define IPAPWD_OP_MOD 2
 
+/* Partial DN of the container for app passwords */
+#define APPPW_CONTAINER_DN "cn=apps,cn=accounts"
+
+/* Maximum number of allowed app passwords */
+#define APPPW_MAX_COUNT 100
```
rcritten, line 75: Is 100 arbitrary or only a temporary setting to get the basic functionality working?

daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
RichardKalinec: I refactored it to use slapi_ldap_explode_dn() function. It is now much more complex both comutationally and w.r.t. code, but it should hopefully be more robust. I now also check for OOM.

daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
RichardKalinec: I heeded this advice here as well, the same holds as mentioned [above](https://github.com/freeipa/freeipa/pull/4923#discussion_r460436550).

daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
RichardKalinec: I decided to use slapi_ch_smprintf(). I now check for OOM throughout entire new code.

daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
RichardKalinec: Yes, in practice. I did not find other way to implement this, as I did not find any function of the slapi plugin to search entries in the LDAP database - therefore, I am not able to easily know what `apppw` objects are already stored in the LDAP database. I picked the number 100, as it is certainly sufficient quantity of app passwords for a single user, while it can be easily enforced by a regex, and it seemed to me as reasonably limiting the performance impact of the new code - but this is where I may be wrong.

RichardKalinec:
On top of the issues mentioned in the previous comments, I also fixed a gross logical error in the new code - I came to understand that the ipapwd_authenticate() return 0 on success and non-zero on failure, so I modified the code accordingly. For now, I also assume that it is not a problem if I attempt to free an object that is actually not allocated - if it is a problem, I have to rewrite the code again, and it will be more complicated.

daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
abbra: To do internal search for entries, use a new `Slapi_PBlock` and do `slapi_search_internal_set_pb()` / `slapi_search_internal_pb()` combination. See, for example, `ipa_cldap_get_domain_entry()` implementation in `daemon/ipa-slapi-plugins/ipa-cldap/ipa_cldap_netlogon.c`

RichardKalinec:
I have rewritten the code in `ipapwd_pre_bind()` to search for entries, as advised by @abbra . A potential problem is how I currently calculate the length of the array of entries returned by the search:
```
apppw_count = sizeof(search_apppw_entries) / sizeof(search_apppw_entries[0]);
```
This is correct if entries occupy the same amount of memory, what is a questionable assumption at best. Is there a more appropriate way how can this be performed?

Another thing that is left to be done is somehow getting the value for the `Slapi_ComponentId *plugin_identity` parameter. I have not found a function to do this. In the file referenced by @abbra for inspiration, it is available in the context structure that is passed into the function, but that is not the case here, and I cannot change the function signature by adding a parameter for it.

abbra:
You should be using `slapi_pblock_get(search_pblock, SLAPI_NENTRIES, &entry_count);`

abbra:
as for the `plugin_identity`, in `ipa-pwd-extop` it is available as
```
extern void *ipapwd_plugin_id;
```

see `daemons/ipa-slapi-plugins/ipa-pwd-extop/common.c` for details.

RichardKalinec:
All right, I have solved this two issues using your advice. Now I need to figure out why the installation of FreeIPA with all these changes crashes, as [described at the end of the first post above](https://github.com/freeipa/freeipa/pull/4923#issuecomment-658994740). The problem is in [this file](https://github.com/freeipa/freeipa/pull/4923/files#diff-b28f3ef1fcf9cbe60da53d21677eca56).

Originally, I had to modify it, as it was:
```
dn: cn=apps,cn=accounts,$SUFFIX
changetype: add
objectclass: top
objectClass: nsContainer
cn: apps
aci: (target = "ldap:///uid=*,cn=($dn),cn=apps,cn=accounts,$SUFFIX")(targetattr = "uid || description || ou || userPassword")(targetfilter=(objectClass=account))(version 3.0; acl "System: Allow users to add or remove an app password for themselves"; allow (add, delete) userdn = "ldap:///uid=($dn),cn=users,cn=accounts,$SUFFIX";)
aci: (target = "ldap:///uid=*,cn=($dn),cn=apps,cn=accounts,$SUFFIX")(targetattr = "uid || description || ou")(targetfilter=(objectClass=account))(version 3.0; acl "System: Allow users to search for their app passwords"; allow (search, read) userdn = "ldap:///uid=($dn),cn=users,cn=accounts,$SUFFIX";)
```
thus - changetype explicitly specified in the beginning, not as a prefix of all added attributes. Back then, it complained about syntax error on 'changetype'. So I reworked it to the current form, following the structure used in other .update files. Now, I have no cue why I still end up with syntax error, this time on 'dn'. What am I doing wrong?

abbra:
Can you explain what do you want to define with `cn=($dn),..` ?

RichardKalinec:
By this, I define that these ACIs grant access only to `apppw` objects stored under the entry with the same `cn` as is the `uid` of the current user.

abbra:
> By this, I define that these ACIs grant access only to `apppw` objects stored under the entry with the same `cn` as is the `uid` of the current user.

Then you should be using `$$dn` because `$foo` is expanded by the template engine and you want to have `$` in the actual ACI string parsed by 389-ds.

RichardKalinec:
All right, hopefully, this problem is sorted out.

abbra:
Please fix linting issues:
```
2020-08-10T18:40:44.3346465Z pycodestyle
2020-08-10T18:40:44.3347735Z -----------
2020-08-10T18:40:44.8471802Z ./ipaserver/plugins/user.py:857:1: W293 blank line contains whitespace
2020-08-10T18:40:44.9457902Z make: *** [Makefile:1230: fastlint] Error 1
```

RichardKalinec:
I have fixed that issue, now we need to sort out things that remain to be implemented - I opened the first one above.

abbra:
pycodestyle is OK but lint is failing:
```
2020-08-11T09:14:18.4895786Z pycodestyle
2020-08-11T09:14:18.4896702Z -----------
2020-08-11T09:14:19.2476568Z 
2020-08-11T09:14:19.2504971Z pylint
2020-08-11T09:14:19.2506406Z ------
2020-08-11T09:14:29.3890125Z [7;33m************* Module ipatests.test_xmlrpc.test_apppw_plugin[0m
2020-08-11T09:14:29.3895118Z ipatests/test_xmlrpc/test_apppw_plugin.py:23: [W0611([35munused-import[0m), ] [35mUnused import ldap[0m)
2020-08-11T09:14:29.3896632Z ipatests/test_xmlrpc/test_apppw_plugin.py:25: [W0611([35munused-import[0m), ] [35mUnused api imported from ipalib[0m)
2020-08-11T09:14:29.3897873Z ipatests/test_xmlrpc/test_apppw_plugin.py:26: [W0611([35munused-import[0m), ] [35mUnused raises imported from ipatests.util[0m)
2020-08-11T09:14:29.3899072Z ipatests/test_xmlrpc/test_apppw_plugin.py:30: [W0611([35munused-import[0m), ] [35mUnused ldap_initialize imported from ipapython.ipaldap[0m)
2020-08-11T09:14:29.3899701Z 
2020-08-11T09:14:29.3900808Z -----------------------------------
```

abbra:
In our tests in Azure CI we don't even reach testing IPA or build it until both pycodestyle and pylint reports are clean, so they need to be fixed. You should be using `make fastlint` or `make lint` locally before pushing to make sure you have clean code.

RichardKalinec:
This is very interesting, because I did so, and since then, no changes were made in the places of the code where these tests complain now. That means pycodestyle and pylint did not detect locally, what they detect now in the cloud. It can be that something was not tested due to errors elsewhere that were fixed later, or I have outdated something w.r.t. these tests. I do not see other explanation. I will fix this right away.

RichardKalinec:
All right, I have also rebased on top of the current master branch, and as fixup commits already started to accumulate too much, while many of them were only minor fixes, I also squashed them all into the initial commit. I hope that this does not cause any problems and makes the implementation easier to navigate.

abbra:
Azure CI runs against up to date Fedora 32 and uses pylint/tox from pypi. It means we always test against very recent pylint while you might not have those locally.

RichardKalinec:
As I updated everything in my testing envirnoment (running on Fedora 32 VM), I tried to build FreeIPA with my modifications, and got a bunch of errors and warnings in `daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c`. I started fixing them, but got stuck at this one:
```
prepost.c:399:30: warning: division 'sizeof (const char **) / sizeof (const char *)' does not compute the number of array elements [-Wsizeof-pointer-div]
  399 |     rdn_count = sizeof(rdns) / sizeof(rdns[0]);
      |                              ^
```
I realized that I have no reliable way to determine how many elements are in the `rdns` array returned by the `slapi_ldap_explode_dn()` function, and I don't see it returning anything about the number of elements, just the pointer to the array of elements themselves. What will I do with this? Without this information, I am unable to reliably determine the end of the array...

abbra:
The returned array is NULL-terminated, so if you want to calculate the RDNs, loop through it. The elements count is typically low enough to allow that. For a user dn (`uid=user,cn=users,cn=accounts,$BASEDN`) the length of an array would be 3+number of DC elements in the `$BASEDN`, which is typically 2-4.

However, I'd question a need to do it this way. You want to find out whether `pwdop->pwdata.dn` is belonging to either `cn=apps,..` subtree or `cn=accounts` subtree. So instead of the manual expansion and testing, use existing SLAPI functions, `slapi_sdn_scope_test()` or any of `slapi_sdn_issuffix()`,  `slapi_sdn_isparent()`, and `slapi_sdn_isgrandparent()`. You can build your test targets in `ipapwd_getConfig()` and store them in `struct ipapwd_krbcfg` instance which is already retrieved by the time you are doing these checks in `prepost.c`.

RichardKalinec:
I would definitely rather do it using existing SLAPI functions. After thinking about it, it would be optimal to create a Slapi_DN* variable in the `struct ipapwd_krbcfg` instance in the `ipapwd_getConfig()` function that would hold full DN of the container for app passwords (including the suffix), and this would be used:
* in the `ipapwd_pre_add()` function to test if the new entry is in this subtree,
* in the `ipapwd_pre_bind()` function to specify where to search for app passwords.
There is one important thing to solve here: how to retrieve the suffix in the `ipapwd_getConfig()` function. Isn't it stored in the `ipa_realm_dn` variable? Or do I need to do it similarly as I was attempting do do it so far in the `ipapwd_pre_add()` and `ipapwd_pre_bind()` functions?

RichardKalinec:
I have committed the refactoring, and fixed the remaining errors. Now the code compiled, fastcheck passed as well as possible, and I once again attempted to install the modified FreeIPA server. The installation failed at the same point as the last time, but this time my file does not seem to directly be a problem:
```
Upgrading IPA:. Estimated time: 1 minute 30 seconds
...
  [7/10]: upgrading server
Error retrieving: cn=changelog5,cn=config
Upgrade failed with no such entry
  [error] RuntimeError: no such entry
...
Update failed: no such entry
```
I made a quick search for what this means, and it seems to be caused by the recent change in 389-DS:
https://www.port389.org/docs/389ds/design/integrate-changelog-database-and-backend-database.html
That missing entry is no longer created for new installations of 389-DS, new solution is used instead. Is FreeIPA already adapted to this change? But it is interesting that this problem occurs precisely when performing updates added by me - but what could be wrong with them? I do not specify anything related to changelog.

abbra:
@RichardKalinec this is going to be corrected with https://github.com/freeipa/freeipa/pull/5023.

daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
```diff
@@ -1505,29 +1552,127 @@ static int ipapwd_pre_bind(Slapi_PBlock *pb)
...
+        /* Authenticate the user using the app password. */
+        ret = ipapwd_authenticate(slapi_entry_get_dn(search_apppw_entries[i]),
+            search_apppw_entries[i], credentials);
+
+        /* Limit effective permissions of the new session to:
+         * 1. full access to the apppw object with the app password that was
+         *    used for authentication
+         * 2. read-only access to the user's primary entry
+         */
+
```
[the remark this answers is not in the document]
RichardKalinec: I apologize, I created this review a week ago, but I did not know I must also then click "Submit review", which I mistakenly thought was just to add more comments.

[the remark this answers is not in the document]
RichardKalinec: I attempted to search on the internet, but did not find anything. I am clueless how to accomplish limiting effective permissions of the new session if the user is authenticated using an app password.

daemons/ipa-slapi-plugins/ipa-pwd-extop/prepost.c
```diff
@@ -379,14 +385,19 @@ static int ipapwd_pre_add(Slapi_PBlock *pb)
...
+        ret = ipapwd_CheckPolicy(&pwdop->pwdata);
+        /* For accounts created by cn=Directory Manager or a passsync
+         * managers, ignore result of a policy check */
+        if ((pwdop->pwdata.changetype != IPA_CHANGETYPE_DSMGR) &&
+            (ret != 0)) {
+            errMesg = ipapwd_error2string(ret);
+            rc = LDAP_CONSTRAINT_VIOLATION;
+            goto done;
+        }
     }
```
RichardKalinec, line 401: As for enforcement of app password generation on the LDAP level, it seems to me that it will be best to add an `else` branch here, where I will generate the app password and ensure it will then be:
1. stored into the LDAP database - this will be accomplished by simply putting the generated password into the entry to be stored (in the `e` variable).
2. somehow retrieved in the POST callback of the `apppw-add` method for displaying to the user - and here I do not know how to do this, as that is in a completely different file written in Python, not C (`ipaserver/plugins/apppw.py`). Currently, it is done in that file, but that is easy, as the app password is generated in the PRE callback, and then just saved and retrieved for display using `setattr(context, 'randompassword', ...)` and `getattr(context, 'randompassword')`.

abbra:
Let us keep this opened. @RichardKalinec, do you have plans on completing this work?

RichardKalinec:
Yes, definitely. I have finished my studies in the meantime and got a job. These days, I am busy, but I definitely will finish this - if I will fare well, I should manage to get back to this in few weeks. Sorry for delays. As you can see right above, I started thinking about two of the reamining issues and asked about them, but as of now, I have no feedback. These issues seem challenging, but I definitely will finish this.

RichardKalinec:
I am gradually getting better at my new job, but I am still too busy to move forward with this. Realistically, I do not expect that I would be able to work on this before February. However, I remain determined to finish this.

As already hinted above, I would greatly appreciate any feedback or advice w.r.t. two of the challenges that remain to be resolved, which I started to think about above [here](https://github.com/freeipa/freeipa/pull/4923#discussion_r468106857) and [here](https://github.com/freeipa/freeipa/pull/4923#discussion_r475766984). But I understand that you are almost certainly very busy, not to mention the general disruption caused by the COVID-19 pandemic. Do you know anybody who might be able to effectively give us needed knowledge and advice?

RichardKalinec:
I am still busy.

yakatz:
I will see if I can add this to my todo list - this would be a great feature...

RichardKalinec:
See [the post](https://github.com/freeipa/freeipa/pull/4061#issuecomment-841711041) I have written in the issue #4061.

PR #4943, opened 2020-07-24
Fix for two failing healthcheck tests

rcritten:
The release of freeipa-healthcheck 0.6 fixes a problem and introduces a new one.

It adds namespacing so we can limit the checks and avoid spurious errors from the PKI healthchecks that miscount the number of errors when all services are stopped.

It also pulls in checks provided by lib389. Change the naming of Replication error checking.


rcritten:
This requires the pending freeipa-healthcheck 0.6-2

rcritten:
A compatibility issue was found with the previous test version with pki-healthcheck. A new set of updates have been submitted,:

F31: https://bodhi.fedoraproject.org/updates/FEDORA-2020-8c469ba568
F32: https://bodhi.fedoraproject.org/updates/FEDORA-2020-60509a76d8

fcami:
@rcritten you could switch to having updates-testing enabled in the temp_commit if you'd like faster testing here.
Check out ipatests/prci_definitions/nightly_latest_testing.yaml
You will need the build and test_ipahealthcheck job definitions as the template differs.

rcritten:
I thought about that but then if it merged before they got pushed to stable the nightlies would break again and it'd be considered a regression.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -89,7 +89,7 @@
 ipafiles_checks = ["IPAFileNSSDBCheck", "IPAFileCheck", "TomcatFileCheck"]
 dogtag_checks = ["DogtagCertsConfigCheck", "DogtagCertsConnectivityCheck"]
 iparoles_checks = ["IPACRLManagerCheck", "IPARenewalMasterCheck"]
-replication_checks = ["ReplicationConflictCheck"]
+replication_checks = ["ReplicationCheck"]
```
fcami, line 92: OK

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -717,6 +717,7 @@ def test_run_with_stopped_master(self, ipactl):
         """
         returncode, output = run_healthcheck(
             self.master,
+            "ipahealthcheck.meta",
```
fcami, line 720: OK

fcami:
The test_advise failure is unrelated ( https://pagure.io/freeipa/issue/8279 ). Re-running.

rcritten:
master:

* d238fb4f2bb6525367a73c42a8a238843d5c8b43 ipatests: lib389 is now providing healthchecks, update naming
* 61c71e4a62347cd7188e407acc69ce8362857cb4 ipatests: Use healthcheck namespacing in stopped server test

PR #4949, opened 2020-07-28
ipatests: Test that healthcheck detects and reports expiration

rcritten:
ipatests: Test that healthcheck detects and reports expiration
    
Set the date forward to while the certificates are still valid and 
run healthcheck to confirm that an appropriate warning is made.
    
This validates two separate checks, one that relies on certmonger
to report expiration and one that relies on the data on disk to
determine expiration in case certmonger is out-of-date for some
reason (belt and suspenders).

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -7,6 +7,7 @@ Tests to verify that the ipa-healthcheck scenarios
 
 from __future__ import absolute_import
 
+from datetime import datetime, timedelta
 import json
 import os
 import re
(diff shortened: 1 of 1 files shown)


rcritten:
I think it's UTC vs local time that is the issue.

flo-renaud:
Hi @rcritten 
thanks for the PR, works for me. Please remove the temp commit and I will ack when PRCI completes.

rcritten:
Rebased and dropped temp commit.

flo-renaud:
Tests are green, ACK

flo-renaud:
master:

* c84b1db80954c4a963f268c62924b69d8ab5c1a2 ipatests: Test that healthcheck detects and reports expiration

PR #4953, opened 2020-07-29
ipatests: Don't turn Pytest IPA deprecation warnings into errors

stanislavlevin:
With new Pytest 6.0 \[0\]:

> PytestDeprecationWarning are now errors by default.
Following our plan to remove deprecated features with as little disruption as
possible, all warnings of type PytestDeprecationWarning now generate errors
instead of warning messages.

PytestWarnings are no longer marked as the part of public API, but as
internal warnings. It's unsafe to use bare PytestDeprecationWarning,
which is turned into the error on major releases.

\[0\]: https://github.com/pytest-dev/pytest/releases/tag/6.0.0

Fixes: https://pagure.io/freeipa/issue/8435

Added:
Note: this fixes only `pypi` task. The whole IPA test framework should be run against the new Pytest 6.0.

### DIFF
--- a/ipatests/conftest.py
+++ b/ipatests/conftest.py
@@ -26,6 +26,10 @@ else:
 
 HERE = os.path.dirname(os.path.abspath(__file__))
 
+
+class PytestIPADeprecationWarning(pytest.PytestWarning, DeprecationWarning):
+    """Warning class for features that will be removed in a future version."""
+
 pytest_plugins = [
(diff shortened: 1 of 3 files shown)


rcritten:
master:

* 55b7787ef5e71200632bcd2751f228b57312fcb0 ipatests: Don't turn Pytest IPA deprecation warnings into errors

PR #4956, opened 2020-07-29
Replace SSLCertVerificationError with SSLCertificateError for py36

rcritten:
This exception was added in python 3.7. Use SSLCertificateError
instead which is an alias and will work with older python releases.

https://bugzilla.redhat.com/show_bug.cgi?id=1858318

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -686,7 +686,7 @@ def http_certificate_ensure_ipa_ca_dnsname(http):
 
     try:
         cert.match_hostname(expect)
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @rcritten 
thanks for the PR, LGTM. The tox failure is a known issue (https://pagure.io/freeipa/issue/8435) that should be fixed with PR #4953

flo-renaud:
Tests are green, ack.

flo-renaud:
master:

* 5dd566951198c3bcf0e5860deea4e76a9b8a6dc0 Replace SSLCertVerificationError with CertificateError for py36

PR #4964, opened 2020-07-30
[Backport][ipa-4-8] Remove paramiko usage from ipatests

fcami:
MANUAL BACKPORT

Paramiko is not compatible with FIPS.
Migrate all tests using paramiko to the OpenSSH CLI SSH(1).

Fixes: https://pagure.io/freeipa/issue/8129

Note that https://pagure.io/freeipa/issue/8431 was filed as sshpass would need an enhancement to cover all OTP tests.

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -30,6 +30,7 @@ import collections
 import itertools
 import shutil
 import copy
+import subprocess
 import tempfile
 import time
 from pipes import quote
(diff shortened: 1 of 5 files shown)


fcami:
@miskopo could you please double-check that the result looks sane compared to the original PR: https://github.com/freeipa/freeipa/pull/4938

miskopo:
> @miskopo could you please double-check that the result looks sane compared to the original PR: #4938

Affirmative, diff is the same.

fcami:
ACKing based on @miskopo comments - this is a clean cherry-pick.

PR #4969, opened 2020-07-31
Issue 8407 - Support changelog integration into main database

marcus2376:
Description: Add support for both the old and new replication changelogs.
             First try to get and update the new entry, if it's not found
             then we know we need to update the old global changelog entry.

https://pagure.io/freeipa/issue/8407

Reviewed by: ?

### DIFF
--- a/install/updates/05-pre_upgrade_plugins.update
+++ b/install/updates/05-pre_upgrade_plugins.update
@@ -1,5 +1,6 @@
 # first
 plugin: update_managed_post_first
+plugin: update_changelog_maxage
 
 # middle
 plugin: update_replica_attribute_lists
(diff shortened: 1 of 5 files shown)


marcus2376:
I was not able to do an extensive testing besides a ipa-server-install.  We just need to sanity check that replication still works, and it should also be tested against this 389-ds-base build:  https://copr.fedorainfracloud.org/coprs/build/1567068

abbra:
@marcus2376 thank you for the PR. It looks good and I have just few comments about markup, not the code:

- Please remove `Reviewed-by: ?` from the commit message, it will be added automatically by a script that will be used to push the PR
- Please add `Signed-off-by: name <email>` for yourself
- Please replace the long copyright header by a shorter one:
```
#
# Copyright (C) 2020  FreeIPA Contributors see COPYING for license
#
```

marcus2376:
@abbra thanks for the review, changes made!

abbra:
@marcus2376 thanks. Sorry, I forgot one more: we use `Fixes: <url>`, not just the `<url>`. This is used by the commit tools.

marcus2376:
Added "Fixed", we use that too in DS/Pagure, but I wasn't sure about IPA so I skipped it.  Good to know for next time.

abbra:
master:

* 44259e8e68582e7116f8d5bee6e2c454254de69f Issue 8407 - Support changelog integration into main database

PR #4986, opened 2020-08-04
ipatests: collect IPA_RENEWAL_LOCK file

flo-renaud:
In order to troubleshoot certmonger timeouts, collect the
file /run/ipa/renewal.lock that is used as cross-process lock
by ipa-server-guard.

### DIFF
--- a/ipatests/pytest_ipa/integration/__init__.py
+++ b/ipatests/pytest_ipa/integration/__init__.py
@@ -76,6 +76,8 @@ CLASS_LOGFILES = [
     # system
     paths.RESOLV_CONF,
     paths.HOSTS,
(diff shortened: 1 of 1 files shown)


rcritten:
This makes sense to me. The lock acquires for an hour so it would totally blow up in the tests if not released.

abbra:
LGTM

rcritten:
master:

* 606f1abd05e0cb89ae7bba26272065bc51ea3d1c ipatests: collect IPA_RENEWAL_LOCK file

rcritten:
Pushed only to master since the failures we're trying to catch are in nightly_latest which runs against master.

PR #4990, opened 2020-08-05
ipatests: Test healthcheck revocation checker

rcritten:
Revoke the Apache certificate and ensure that healthcheck properly
reports the problem.


abbra:
Please remove temp commit. @mrizwan93 we typically do not add ACK before all PR CI tests passed.

mrizwan93:
@abbra Ahh, sorry for that.

rcritten:
Temp patch removed.

rcritten:
Since the board is green again after removing the temp commit I'll add back the ack.

rcritten:
master:

* 61db3527e31b1d7c7cda15c034c666cc95c17775 ipatests: Test healthcheck revocation checker

PR #4999, opened 2020-08-06
Don't configure authselect in containers

tiran:
freeipa-container images come with authselect pre-configured. There is
no need to configure, migrate, or restore authselect. The --mkhomedir
option is not supported, too.

Related: https://pagure.io/freeipa/issue/8401

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -2160,6 +2160,14 @@ def install_check(options):
             "authentication resources",
             rval=CLIENT_INSTALL_ERROR)
 
+    # --mkhomedir is not supported by fedora_container and rhel_container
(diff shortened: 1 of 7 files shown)


tiran:
See https://github.com/freeipa/freeipa-container/blob/40802b63aecc0208aac9b60a6f257a371723450b/Dockerfile.fedora-32#L68-L69 and https://github.com/freeipa/freeipa-container/blob/master/authselect

abbra:
LGTM. Thanks also for the readme part.

frasertweedale:
Coool, nice work @tiran!

tiran:
master:

* 999485909a0bc654570f05cd2127d86d14d6f25a Don't configure authselect in containers

PR #5002, opened 2020-08-06
ipatests: Test certmonger rekey command works fine

mrizwan93:
Certmonger's rekey command was throwing an error as
unrecognized command. Test is to check if it is working fine.

related: https://bugzilla.redhat.com/show_bug.cgi?id=1249165

### DIFF
--- a/ipatests/test_integration/test_cert.py
+++ b/ipatests/test_integration/test_cert.py
@@ -8,7 +8,10 @@ related scenarios.
 """
 import ipaddress
 import pytest
+import random
 import re
+import string
+import time
 
 from ipaplatform.paths import paths
 from cryptography import x509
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_cert.py
```diff
@@ -216,6 +216,52 @@ def test_getcert_list_profile_using_subca(self, test_subca_certs):
...
+        unrecognized command. Test is to check if it is working fine.
+
+        related: https://bugzilla.redhat.com/show_bug.cgi?id=1249165
+        """
+        result = self.master.run_command([
+            'ipa-getcert', 'request',
+            '-f', '/etc/pki/tls/certs/test_rekey.pem',
+            '-k', '/etc/pki/tls/private/test.key',
+            '-K', 'test/{}'.format(self.master.hostname)])
+        request_id = re.findall(r'\d+', result.stdout_text)
```
rcritten, line 232: You can use get_certmonger_fs_id() from ipatests.test_integration.test_certs for this regex instead.

mrizwan93: Initially I used it but it returns `request_id[1]` and we are interested in `request_id[0]` here

kaleemsiddiqu:
@mrizwan93 
I think we should more tests covering this e.g rekey with different key-type with option -G, rekey without any option etc. Also couple of scenarios from Certificate handling settings getcert-rekey .

ipatests/test_integration/test_cert.py
rcritten: Ok, right, you're using the output of getcert. Another way to do this is to use -I <name> to specify the nickname you want to use for this cert, then you don't have to regex the output to find the certmonger-generated id. What you are doing is fine.

rcritten:
The IPA CA will only issue RSA certs so this is going to limit your key type checking.

You can change other things in the request add/remove a SAN for example. I guess that would confirm that the other cli options also work.

kaleemsiddiqu:
> The IPA CA will only issue RSA certs so this is going to limit your key type checking.
@rcritten , Do we have any documentation for this? Do we need an RFE for this for other key types?
> 
> You can change other things in the request add/remove a SAN for example. I guess that would confirm that the other cli options also work.

rcritten:
I think there is an RFE to issue EC certs, I don't think there is one for DSA. I don't know how useful they would be.

ipatests/test_integration/test_cert.py
```diff
@@ -217,6 +218,135 @@ def test_getcert_list_profile_using_subca(self, test_subca_certs):
...
+class TestCertmongerRekey(IntegrationTest):
+
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master, setup_dns=True)
+
+    @pytest.fixture
+    def request_cert(self):
+        """Fixture to request and remove a certificate"""
+        self.request_id = 'test'
```
kaleemsiddiqu, line 230: IMO we should randomize the request_id on each request.

ipatests/test_integration/test_cert.py
```diff
@@ -217,6 +218,135 @@ def test_getcert_list_profile_using_subca(self, test_subca_certs):
...
+            '-I', self.request_id,
+            '-K', 'test/{}'.format(self.master.hostname)])
+        status = tasks.wait_for_request(self.master, self.request_id, 100)
+        assert status == "MONITORING"
+
+        yield
+
+        self.master.run_command(['getcert', 'stop-tracking',
+                                 '-i', self.request_id])
+        self.master.run_command([
```
kaleemsiddiqu, line 244: We also need to clean up the key as well, right?

ipatests/test_integration/test_cert.py
```diff
@@ -217,6 +218,135 @@ def test_getcert_list_profile_using_subca(self, test_subca_certs):
...
+        result = self.master.run_command(['getcert', 'rekey',
+                                          '-i', self.request_id,
+                                          '-I', new_req_id])
+        # above command output: Resubmitting "newtest" to "IPA".
+        assert new_req_id in result.stdout_text
+
+        # rename back the request id for fixture to delete it
+        self.master.run_command(['getcert', 'rekey',
+                                 '-i', new_req_id,
+                                 '-I', self.request_id])
```
kaleemsiddiqu, line 316: Should not we check the new request_id in getcerrt list?

kaleemsiddiqu: you are not capturing result.stdout_text

ipatests/test_integration/test_cert.py
```diff
@@ -217,6 +220,142 @@ def test_getcert_list_profile_using_subca(self, test_subca_certs):
...
+                                          '-i', self.request_id,
+                                          '-I', new_req_id])
+        # above command output: Resubmitting "newtest" to "IPA".
+        assert new_req_id in result.stdout_text
+
+        # rename back the request id for fixture to delete it
+        result = self.master.run_command(['getcert', 'rekey',
+                                          '-i', new_req_id,
+                                          '-I', self.request_id])
+        assert self.request_id in result.stdout_text
```
kaleemsiddiqu, line 325: self.request_id should be self.new_req_id

ipatests/test_integration/test_cert.py
rcritten: I think using a fixed request id is fine as long as it is unique to the test (or random). It avoids the requirement to scrape the output of getcert.

kaleemsiddiqu:
@mrizwan93 , saw following pycodestyle errors which needs to be fixed.

pycodestyle
-----------
./ipatests/test_integration/test_cert.py:233:17: E126 continuation line over-indented for hanging indent
./ipatests/test_integration/test_cert.py:235:21: E123 closing bracket does not match indentation of opening bracket's line
./ipatests/test_integration/test_cert.py:236:17: E123 closing bracket does not match indentation of opening bracket's line
./ipatests/test_integration/test_cert.py:254:13: E123 closing bracket does not match indentation of opening bracket's line
./ipatests/test_integration/test_cert.py:259:13: E123 closing bracket does not match indentation of opening bracket's line
make: *** [Makefile:1230: fastlint] Error 1

kaleemsiddiqu:
@mrizwan93 

rcritten:
master:

* abd0cbfcfdb913e4576721bc9c5acd43bdd2bae2 ipatests: Test certmonger rekey command works fine

PR #5006, opened 2020-08-06
IPA-EPN: use entry.get() to retrieve attributes

fcami:
Use entry.get() to retrieve attributes to avoid tripping on missing attrs.

Fixes: TBD


fcami:
Note WIP: the fix is not complete.

PR #5007, opened 2020-08-06
ipatests: fix TestIpaHealthCheckWithoutDNS failure

flo-renaud:
TestIpaHealthCheckWithoutDNS is launched after
TestIpaHealthCheck::test_ipa_healthcheck_expiring that is playing with
the date. At the end of test_ipa_healthcheck_expiring, the date is
reset using systemctl start chronyd but the date may need time to adjust
and the subsequent tests may be launched with a system date set in the
future.

When this happens, dnf install fails because the certificate for
the package repo is seen as expired, and TestIpaHealthCheckWithoutDNS
fails.

In order to avoid this issue, call chronyc waitsync to make sure the
date was adjusted back.

Fixes: https://pagure.io/freeipa/issue/8447

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -860,22 +860,31 @@ class TestIpaHealthCheck(IntegrationTest):
                     else:
                         assert check["kw"]["days"] == 10
 
+        # Store the current date to restore at the end of the test
+        now = datetime.utcnow()
+        now_str = datetime.strftime(now, "%Y-%m-%d %H:%M:%S Z")
+
         # Pick a cert to find the upcoming expiration
         certfile = self.master.get_file_contents(paths.RA_AGENT_PEM)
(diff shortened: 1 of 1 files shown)


rcritten:
The failure does seem to suggest your hypothesis. Perhaps increase the number of tries to 6 to try over a minute and add the -d option for more output.

It would probably also be a good idea to grab datetime.datetime.utcnow() at the beginning of the test and set to that value when the test is done before the waitsync and restarting chronyd to get more in the ballpark of the current time.

It's very strange, I must have run this code a dozen times locally without a failure :-(

rcritten:
Patch looks good, please remote the temp commit.

flo-renaud:
Rebased to include the azure pipeline fix.

flo-renaud:
master:

* f2711321679f99be9a8521f07c81e715bad74b06 ipatests: fix TestIpaHealthCheckWithoutDNS failure

PR #5011, opened 2020-08-07
CLI Validation of ipa-healtcheck tool for list-sources option

menonsudhir:
Test ensure that error message is displayed on the console when arguments are specified to --list-sources option
to ipa-healthcheck command.


menonsudhir:
Closing the PR, since the test has been added to original PR4996

PR #5013, opened 2020-08-07
[Azure] Switch to dockerhub

stanislavlevin:
`registry.fedoraproject.org/f32/fedora-toolbox` image is used to build packages on Azure Pipelines.
registry.fedoraproject.org experiences an availability problem and makes unstable FreeIPA CI.

Fedora also distributes its official images on https://hub.docker.com/_/fedora.
`fedora:32` is already used by FreeIPA CI to build the image for tests.

### DIFF
--- a/ipatests/azure/Dockerfiles/Dockerfile.build.fedora
+++ b/ipatests/azure/Dockerfiles/Dockerfile.build.fedora
@@ -4,10 +4,6 @@ ENV container=docker LANG=en_US.utf8 LANGUAGE=en_US.utf8 LC_ALL=en_US.utf8
 
 ADD dist /root
 RUN echo 'deltarpm = false' >> /etc/dnf/dnf.conf \
-    && dnf install -y dnf-plugins-core \
-    && dnf config-manager --set-enabled updates-testing \
-    && dnf config-manager --set-enabled updates-testing-modular \
-    && dnf config-manager --set-disabled fedora-cisco-openh264 \
     && dnf update -y dnf \
     && sed -i 's/%_install_langs \(.*\)/\0:fr/g' /etc/rpm/macros.image-language-conf \
     && dnf install -y systemd \
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -262,6 +262,7 @@ BuildRequires: python3-m2r
 BuildRequires:  jsl
 BuildRequires:  rpmlint
 BuildRequires:  softhsm
+BuildRequires:  keyutils
 BuildRequires:  python3-augeas
 BuildRequires:  python3-cffi
 BuildRequires:  python3-cryptography >= 1.6
--- a/ipatests/azure/azure-pipelines.yml
+++ b/ipatests/azure/azure-pipelines.yml
@@ -12,7 +12,7 @@ jobs:
     vmImage: $(VM_IMAGE)
   container:
     image: $(DOCKER_BUILD_IMAGE)
-    options: --cap-add=SYS_PTRACE --security-opt seccomp=unconfined --privileged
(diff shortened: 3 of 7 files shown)


ipatests/azure/Dockerfiles/Dockerfile.build.fedora
```diff
@@ -4,10 +4,6 @@ ENV container=docker LANG=en_US.utf8 LANGUAGE=en_US.utf8 LC_ALL=en_US.utf8
 
 ADD dist /root
 RUN echo 'deltarpm = false' >> /etc/dnf/dnf.conf \
-    && dnf install -y dnf-plugins-core \
-    && dnf config-manager --set-enabled updates-testing \
-    && dnf config-manager --set-enabled updates-testing-modular \
-    && dnf config-manager --set-disabled fedora-cisco-openh264 \
```
fcami, line 10: Why is this change necessary? Is this because the images differ?

abbra: this seems to be unnecessary with the fedora:32 image anymore.

abbra:
Azure CI pipeline: https://dev.azure.com/freeipa/freeipa/_build/results?buildId=5674&view=results
It moved past build and lint stages into running IPA tests so we can consider this change successful.

stanislavlevin:
Please, wait a bit, the keyrings are not skipped. Let me check why.

ipatests/azure/Dockerfiles/Dockerfile.build.fedora
stanislavlevin: That was added by me in 26f96595b010396f055c7d11a645be57fc637863 :
> Azure: Make dnf repos consistent
    Build container(image registry.fedoraproject.org/f32/fedora-toolbox)
    has two more dnf repos enabled compared to Tests container(image
    fedora:32). This results in the packages built within the Build
    container can have dependencies which are unresolvable(missing)
    within Tests container.

With this change `build` stage and `tests` one will use the same parent image.

stanislavlevin:
heh, container within VM is not detected for some reason:
```
[root@97684d414c4b /]# /usr/bin/systemd-detect-virt --container
none
[root@97684d414c4b /]# echo $?
1
```

stanislavlevin:
Ah, https://github.com/systemd/systemd/blob/master/src/basic/virt.c#L522
```
        /* Fallback for cases where PID 1 was not systemd (for example, cases where init=/bin/sh is used. */
        r = getenv_for_pid(1, "container", &m);
        if (r > 0) {
                e = m;
                goto translate_name;
        }
```

ipatests/azure/Dockerfiles/Dockerfile.build.fedora
fcami: ack for this, thanks for the explanation.

abbra:
master:

* 2b85bfb03052b56591b06abe62c71e36170dc9a0 Azure: Switch to dockerhub provider
* e5c09675f3c7c2ecb105df4c4ecf846a2d3ffdc4 ipatests: Skip keyring tests on containerized platforms

stanislavlevin:
So, to let `systemd-detect-virt` to work correctly one should run this util as a privileged user ( to be able to read
`/proc/1/environ`), otherwise read `/run/systemd/container` \[0\], which is populated by `systemd` on startup.
But Azure run tasks within containers as an unprivileged user and the `build` image, in turn, comes without pre-installed `systemd` package. Thus, to let `systemd` to be the PID 1 we have to re-build container image, which takes additional time if made during CI or pre-made which is additional maintenance duty.

Thus, the best option is to leave as-is :-)

\[0\]: https://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/

PR #5023, opened 2020-08-10
Issue 8456 - Add new aci's for the new replication changelog entries

marcus2376:
Description:  

We need a read and a write aci for the new changelog location, which was moved from cn=changelog5,cn=config to cn=changelog,cn=BACKEND,cn=ldbm database,cn=plugins,cn=config

The read aci allows the replica hostgroup entry to find and read the changelog configuration, and the write allows the replica to update the changelog with a proper trimming settings.

Fixes: https://pagure.io/freeipa/issue/8456

Signed-off-by: Mark Reynolds <mreynolds@redhat.com>

### DIFF
--- a/install/updates/40-delegation.update
+++ b/install/updates/40-delegation.update
@@ -212,6 +212,28 @@ default:ipapermissiontype: SYSTEM
 dn: cn=config
 add:aci: (targetattr = "cn || createtimestamp || entryusn || modifytimestamp || objectclass || passsyncmanagersdns*")(target = "ldap:///cn=ipa_pwd_extop,cn=plugins,cn=config")(version 3.0;acl "permission:Read PassSync Managers Configuration";allow (compare,read,search) groupdn = "ldap:///cn=Read PassSync Managers Configuration,cn=permissions,cn=pbac,$SUFFIX";)
 
+dn: cn=Read Replication Changelog Configuration,cn=permissions,cn=pbac,$SUFFIX
+default:objectClass: groupofnames
+default:objectClass: ipapermission
+default:objectClass: top
+default:cn: Read Replication Changelog Configuration
+default:member: cn=Replication Administrators,cn=privileges,cn=pbac,$SUFFIX
+default:ipapermissiontype: SYSTEM
+
+dn: cn=config
+add:aci: (targetattr = "cn || objectclass || nsslapd-changelogmaxentries || nsslapd-changelogmaxage || nsslapd-changelogtrim-interval || nsslapd-encryptionalgorithm || nsSymmetricKey")(targetfilter = "cn=changelog")(target = "ldap:///cn=ldbm database,cn=plugins,cn=config")(version 3.0; acl "permission:Read Replication Changelog Configuration"; allow (read,search) groupdn = "ldap:///cn=Read Replication Changelog Configuration,cn=permissions,cn=pbac,$SUFFIX";)
+
+dn: cn=Write Replication Changelog Configuration,cn=permissions,cn=pbac,$SUFFIX
+default:objectClass: groupofnames
+default:objectClass: ipapermission
+default:objectClass: top
+default:cn: Write Replication Changelog Configuration
+default:member: cn=Replication Administrators,cn=privileges,cn=pbac,$SUFFIX
+default:ipapermissiontype: SYSTEM
+
+dn: cn=config
+add:aci: (targetattr = "nsslapd-changelogmaxentries || nsslapd-changelogmaxage || nsslapd-changelogtrim-interval || nsslapd-encryptionalgorithm || nsSymmetricKey")(targetfilter = "cn=changelog")(target = "ldap:///cn=ldbm database,cn=plugins,cn=config")(version 3.0; acl "permission:Write Replication Changelog Configuration"; allow (write) groupdn = "ldap:///cn=Write Replication Changelog Configuration,cn=permissions,cn=pbac,$SUFFIX";)
+
 dn: cn=Modify PassSync Managers Configuration,cn=permissions,cn=pbac,$SUFFIX
 default:objectClass: groupofnames
 default:objectClass: ipapermission
--- a/ipaserver/install/plugins/update_changelog_maxage.py
+++ b/ipaserver/install/plugins/update_changelog_maxage.py
@@ -45,7 +45,7 @@ class update_changelog_maxage(Updater):
                     cl_entry = ldap.get_entry(dn, ['nsslapd-changelogmaxage'])
                     self.update_entry(cl_entry, ldap)
                 except errors.NotFound:
-                    logger.warning('Error retrieving: %s', str(dn))
+                    logger.debug('Error retrieving: %s', str(dn))
                 return False, []
 
         return False, []
--- a/ipaserver/install/replication.py
+++ b/ipaserver/install/replication.py
@@ -619,7 +619,11 @@ class ReplicationManager:
         else:
             # Set the changelog trimming
             cl_entry['nsslapd-changelogmaxage'] = '7d'
-            conn.update_entry(cl_entry)
+            try:
+                conn.update_entry(cl_entry)
+            except errors.EmptyModlist:
+                # not a problem since the trimming is already set
+                pass
 
     def _finalize_replica_settings(self, conn):
         """Change replica settings to final values


install/updates/40-delegation.update
```diff
@@ -211,6 +211,8 @@ default:ipapermissiontype: SYSTEM
...
+add:aci: (targetattr = "cn || objectclass || nsslapd-changelogmaxentries || nsslapd-changelogmaxage || nsslapd-changelogtrim-interval || nsslapd-encryptionalgorithm || nsSymmetricKey")(targetfilter = "cn=changelog")(target = "ldap:///cn=ldbm database,cn=plugins,cn=config")(version 3.0; acl "Replication Admin read access to replication changelog"; allow (read,search) groupdn = "ldap:///cn=Replication Administrators,cn=privileges,cn=pbac,$SUFFIX";)
+add:aci: (targetattr = "nsslapd-changelogmaxentries || nsslapd-changelogmaxage || nsslapd-changelogtrim-interval || nsslapd-encryptionalgorithm || nsSymmetricKey")(targetfilter = "cn=changelog")(target = "ldap:///cn=ldbm database,cn=plugins,cn=config")(version 3.0; acl "Replication Admin write access to replication changelog"; allow (write) groupdn = "ldap:///cn=Replication Administrators,cn=privileges,cn=pbac,$SUFFIX";)
```
rcritten, line 215: The group dn need to be in cn=permissions, not cn=privileges.

The ACI name needs the prefix "permission" ala acl "permission:Replication Admin read access to replication changelog"

Is targetfilter necessary? It isn't really a filter is it? Can't cn=changelog be merged with target?

marcus2376: > The group dn need to be in cn=permissions, not cn=privileges.

Well it needs to be one of these groups:

memberOf: cn=Replication Administrators,cn=privileges,cn=pbac,dc=ipa,dc=test
memberOf: cn=Add Replication Agreements,cn=permissions,cn=pbac,dc=ipa,dc=test
memberOf: cn=Modify Replication Agreements,cn=permissions,cn=pbac,dc=ipa,dc=te
 st
memberOf: cn=Read Replication Agreements,cn=permissions,cn=pbac,dc=ipa,dc=test
memberOf: cn=Remove Replication Agreements,cn=permissions,cn=pbac,dc=ipa,dc=te
 st
memberOf: cn=Modify DNA Range,cn=permissions,cn=pbac,dc=ipa,dc=test
memberOf: cn=Read PassSync Managers Configuration,cn=permissions,cn=pbac,dc=ip
 a,dc=test
memberOf: cn=Modify PassSync Managers Configuration,cn=permissions,cn=pbac,dc=
 ipa,dc=test
memberOf: cn=Read LDBM Database Configuration,cn=permissions,cn=pbac,dc=ipa,dc
 =test
memberOf: cn=Add Configuration Sub-Entries,cn=permissions,cn=pbac,dc=ipa,dc=te
 st
memberOf: cn=Read DNA Range,cn=permissions,cn=pbac,dc=ipa,dc=test

Which group would you prefer?

> 
> The ACI name needs the prefix "permission" ala acl "permission:Replication Admin read access to replication changelog"
> 
> Is targetfilter necessary? It isn't really a filter is it? Can't cn=changelog be merged with target?

It's to single out each changelog entry for each backend

cn=changelog,cn=userroot,cn=ldbm database,cn=plugins,cn=config
cn=changelog,cn=ipaca,cn=ldbm database,cn=plugins,cn=config

So we need a filter it so we only target the changelog entry from each backend, and not the backend entry itself:  cn=userroot,cn=ldbm database,cn=plugins,cn=config.  So the ACI targets cn=ldbm database,cn=plugins,cn=config, then it only applies to "cn=changelog" entries under that target.

marcus2376: > Which group would you prefer?

I left out this group which is another option:    cn=ipaservers,cn=hostgroups,cn=accounts,$SUFFIX

rcritten: There is a 1-1 mapping of ACI to permission in IPA. A privilege represents an operation and is generally a single permission/ACI but there are times where multiple are required. Examples are adding a user and setting a password are two permissions in the same privilege, and some of the replication management is centralized in a single privilege.

For your purposes, see Read PassSync Managers Configuration in the same update file as an example. The permissions entry is created with a member entry for the Replication Administrators privilege. Then the ACI points to this permission as the bind rule. So you'd need to generate two permissions entries, one for each ACI.

marcus2376:
Changes made - thanks for the info Rob!  

Just out of curiosity is the ACI name used by the UI or the CLI somehow?

flo-renaud:
> Just out of curiosity is the ACI name used by the UI or the CLI somehow?

Yes, in the GUI you can browse IPA Server> Role-Based Access Control > Permissions and you will see the new permission "Read Replication Changelog Configuration", or in the CLI with ipa permission-find/permission-show.

rcritten:
Having trouble testing this. The F31 389-ds-base builds don't have the moved changelog and the current nightly fails to install with IPA (CA doesn't start). I tried an upgrade too and 389 fails to start due to attribute conflicts:

```
[12/Aug/2020:12:48:05.923712551 -0400] - ERR - dse_read_one_file - The entry cn=schema in file /usr/share/dirsrv/schema/10rfc2307compat.ldif (lineno: 1) is invalid, error code 20 (Type or value exists) - object class nisMap: The name does not match the OID "1.3.6.1.1.1.2.9". Another object class is already using the name or OID.
[12/Aug/2020:12:48:06.076486268 -0400] - ERR - setup_internal_backends - Please edit the file to correct the reported problems and then restart the server.
[12/Aug/2020:12:48:06.259816451 -0400] - ERR - dse_read_one_file - The entry cn=schema in file /usr/share/dirsrv/schema/10rfc2307compat.ldif (lineno: 1) is invalid, error code 20 (Type or value exists) - object class nisMap: The name does not match the OID "1.3.6.1.1.1.2.9". Another object class is already using the name or OID.
[12/Aug/2020:12:48:06.337037278 -0400] - ERR - setup_internal_backends - Please edit the file to correct the reported problems and then restart the server.
```

marcus2376:
Funny, I just fixed this schema issue upstream like 5 minutes ago.

rcritten:
@marcus2376 provided me a build to test with. I wasn't able to get it working with the CA I was able to get a successful CA-less installation. The ACI's and permissions work fine. One side-effect, just as a note, not a black mark, is that with most permissions you can do ipa permission-show --all --raw <name> and the raw ACI will be visible. That doesn't cover those permissions that are under cn=config because users can't read that (for good reason). Just something to remember.

I pulled the build from https://copr.fedorainfracloud.org/coprs/build/1607102

Adding ipa-4-8 label since the previous changelog patch landed there.

rcritten:
I don't see a problem merging this now since it grants permission to part of a tree that currently doesn't exist. Waiting on further opinions though in case I'm missing something.

abbra:
LGTM too. Nothing stands out as wrong to me. We already have a number of ACIs that reference 'unknown' trees, whole compat tree does not exist at the time when acl plugin parses those ACIs.

flo-renaud:
There are still a few issues with the new replication changelog.
- On a fresh master install, a warning is printed to the console: `Error retrieving cn=changelog5,cn=config`. The upgrade plugin `update_changelog_maxage` is trying to update the changelog entry (at the new location, or at the old location) but the entry doesn't exist before a replica is setup. The warning can probably be moved to a debug level.
- First replica installation works well, except for the same error message for the changelog on o=ipaca.
- Second replica installation fails with EmptyModList in ipaserver/install/replication.py, line 622. The code should ignore EmptyModList exceptions which can happen if the changelog max age is also set to the expected value.

marcus2376:
> There are still a few issues with the new replication changelog.
> 
>     * On a fresh master install, a warning is printed to the console: `Error retrieving cn=changelog5,cn=config`. The upgrade plugin `update_changelog_maxage` is trying to update the changelog entry (at the new location, or at the old location) but the entry doesn't exist before a replica is setup. The warning can probably be moved to a debug level.
> 
>     * First replica installation works well, except for the same error message for the changelog on o=ipaca.
> 
>     * Second replica installation fails with EmptyModList in ipaserver/install/replication.py, line 622. The code should ignore EmptyModList exceptions which can happen if the changelog max age is also set to the expected value.

I'm looking into these issues now...

marcus2376:
This revision fixes all issues found by @flo-renaud ...

abbra:
LGTM.

flo-renaud:
master:

* b9ae7c45b8b99a3c8aefc0c962e4563c43e51e99 Issue 8456 - Add new aci's for the new replication changelog entries

PR #5025, opened 2020-08-10
IPA-EPN: Test that EPN can be installed, uninstalled and re-installed

rcritten:
Verify that no cruft is left over that will prevent reinstallation
if it is uninstalled.

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -37,6 +37,7 @@ from pipes import quote
 import configparser
 from contextlib import contextmanager
 from pkg_resources import parse_version
+import uuid
 
 import dns
 from ldif import LDIFWriter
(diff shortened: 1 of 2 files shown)


rcritten:
Ok, so the package I download during install is deleted in the tests themselves by...something? The directory is still there but not the package.

fcami:
You might want to copy it somewhere safe (/root/whatever) and use it from there. Any successful dnf invocation will clean the package cache for instance.

rcritten:
Saving a copy of the downloaded package to /tmp works. Did some cleanup, added comments and squashed all changes.

abbra:
The test works, please remove the temp commit.

rcritten:
Thanks, temp commit removed.

flo-renaud:
The temp commit has been removed, the tests are green, ACK.

flo-renaud:
master:

* af5138c2aa80c85b9994eb2891d2f1ec1e2cad01 IPA-EPN: Test that EPN can be install, uninstalled and re-installed

PR #5026, opened 2020-08-11
rpm-spec: Don't fail on missing /etc/ssh/ssh_config

stanislavlevin:
`openssh-clients` is not a strict requirement of `freeipa-client`
package and if it's missing then this case should be handled in
post scriptlet of `freeipa-client` package. Otherwise, the remaining
part of that scriptlet will not be run at all.

Fixes: https://pagure.io/freeipa/issue/8459

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1078,14 +1078,11 @@ if [ $1 -gt 1 ] ; then
         fi
 
         %{__python3} -c 'from ipaclient.install.client import configure_krb5_snippet; configure_krb5_snippet()' >>/var/log/ipaupgrade.log 2>&1
-    fi
-
-    if [ $restore -ge 2 ]; then
         %{__python3} -c 'from ipaclient.install.client import update_ipa_nssdb; update_ipa_nssdb()' >>/var/log/ipaupgrade.log 2>&1
-    fi
-
-    if [ $restore -ge 2 ]; then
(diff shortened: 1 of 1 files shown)


rcritten:
This looks ok to me and it works, I guess I'd prefer to keep the spacing between the individual statements for readability given they all overlap lines, but it isn't a deal breaker.

What releases did you have in mind, ipa-4-8 backport or master only?

flo-renaud:
The issue is present on ipa-4-6, ipa-4-8 and master but we are lucky as the failure happens on the last command run in the post scriptlet => there is no consequence apart from the message logged to the console. So no strong opinion on my side regarding the backports...

abbra:
I think we should backport down to all supported branches. Whether downstreams consider this worth using in a specific package/distribution is up to them.

abbra:
LGTM

flo-renaud:
master:

* 777147e051d71ebfbac23269cf78fcb23a3b1f43 rpm-spec: Don't fail on missing /etc/ssh/ssh_config

flo-renaud:
@stanislavlevin a manual backport is needed for ipa-4-6 branch, can you do it? thanks

stanislavlevin:
@flo-renaud, sure.

flo-renaud:
For full ref, the backport to ipa-4-6 branch is PR #5037

PR #5027, opened 2020-08-11
Support older servers in is_ipa_configured() call

rcritten:
is_ipa_configured() incorrectly returned False if the installation section didn't exist. Fall back to the original detection method if the section doesn't exist.

Add a test for this case.

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -21,6 +21,7 @@ from cryptography import x509 as crypto_x509
 
 from ipalib import x509
 from ipalib.constants import DOMAIN_LEVEL_0
+from ipalib.sysrestore import SYSRESTORE_STATEFILE, SYSRESTORE_INDEXFILE
 from ipapython.dn import DN
 from ipaplatform.constants import constants
 from ipaplatform.osinfo import osinfo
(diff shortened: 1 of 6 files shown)


rcritten:
@abbra heads up, this is needed to allow upgrades from existing versions.

abbra:
@rcritten thanks, I noticed the issue myself yesterday and am holding off the release until we get it sorted out.

flo-renaud:
Hi @rcritten 
thanks for the patch, LGTM. Tested with an old install upgrade, works well, and the test is correctly detecting issues. Please remove the temp commit and I will ack when pr-ci completes.

ipatests/test_integration/test_installation.py
```diff
@@ -357,6 +358,53 @@ def test_ipa_ca_crt_permissions(self):
         assert owner == "root"
         assert group == "root"
 
+    def test_is_ipa_configured(self):
```
stanislavlevin, line 361: There is the FileBackup which could be used as a context manager in Pytest fixture or directly.
I mean if some of the assertions fail then `SYSRESTORE_STATEFILE` ends up in an unpredictable state.

stanislavlevin:
Modified `is_ipa_configured` never returns `None`.


https://github.com/freeipa/freeipa/blob/143dea18fe9bc2ebc65a8dbdcfa44d331adc161c/ipaserver/install/server/upgrade.py#L1489

https://github.com/freeipa/freeipa/blob/143dea18fe9bc2ebc65a8dbdcfa44d331adc161c/ipaserver/install/server/upgrade.py#L1490 is unreachable.

ipatests/test_integration/test_installation.py
```diff
@@ -357,6 +358,53 @@ def test_ipa_ca_crt_permissions(self):
...
+
+           If not then it will fall back to looking for configured
+           services and files and use that for determination.
+        """
+        cmd = ['python3', '-c',
+               'import sys; from ipalib import facts; sys.exit(0 '
+               'if facts.is_ipa_configured() else 1);']
+
+        # Use new method which should return 0 since successful install
+        self.master.run_command(cmd)
```
stanislavlevin, line 374: In this case, we want to be sure that `is_ipa_configured` correctly detects the new state (`installation.complete`), but at this point, it's an unknown state. I suggest making sure that `sstore.get_state("installation", "complete") is True`.

rcritten: The intention was to never return None anymore so it's complete or it isn't. I'll have to remove the is None comparisions for this.

rcritten:
Updated the test in the upgrader so that if there is no installation state then call ipa_is_configured() and rely on old behavior to know whether the installation is complete or not and set that as the state.

ipatests/test_integration/test_installation.py
rcritten: Subsequent tests should confirm that sstore is updated. I wanted to specifically test that the code in the spec file worked as it used to.

ipatests/test_integration/test_installation.py
rcritten: It could but since the installation falls back to the old method it should be able to cleanup properly. If it doesn't then that's a good thing and demonstrates a bug.

ipatests/test_integration/test_installation.py
stanislavlevin: I mean that the initial state of `installation.complete` is unknown (it may be missing at all). In this case, the positive scenario of the old method will be checked twice.

method | scenario | checked
-|-|-
old | positive | <ul><li>           [x] </li></ul>
old | negative | <ul><li>           [ ] </li></ul>
new | positive |               ?
new | negative | <ul><li>           [x] </li></ul>

stanislavlevin:
@rcritten, awesome! thank you.

stanislavlevin:
That means ACK from my side ;-)

abbra:
So we need to remove a temp commit and do a full test run.

rcritten:
Temp commit removed, rebased to master

flo-renaud:
Temp commit has been removed, tests are green and review comments have been addressed. ACK.

flo-renaud:
master:

* a8d5e6bbfe03070925c85793ac898299d543a3d4 Fall back to old server installation detection when needed
* 2bdb18d56fd65e0058ee302502f68a37decc55e8 Use is_ipa_configured from ipalib.facts
* 25e042d3d1e04972d4536b1fa793811e6b047559 ipatests: Add test for is_ipa_configured

flo-renaud:
@rcritten a manual backport is needed for ipa-4-8 branch, can you do it? Thanks

PR #5030, opened 2020-08-12
WIP: Troubleshoot Azure

netoarmando:
Azure is displaying the output of the first of the five tests, this PR tells Azure to only run `test_forced_client_enrolment`.


rcritten:
This is caused by a broken krb5 package in F32, https://bodhi.fedoraproject.org/updates/FEDORA-2020-e33d26ea4e

We need a new stable build to address this. It's already in stable so too late to claw it back.

abbra:
Closing the PR as it is not needed anymore.

PR #5040, opened 2020-08-17
ipatests: stop the CA during healthcheck expiration test

rcritten:
ipatests: stop the CA during healthcheck expiration test
    
Time is moved during the test to ensure that ipa-healthcheck
finds expired certificates. It's possible that certmonger will also
wake up and renew the certificates before ipa-healthcheck can 
execute so shut down the CA during the test so that certificates
can't be issued.
    
https://pagure.io/freeipa/issue/8463

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -830,7 +830,7 @@ class TestIpaHealthCheck(IntegrationTest):
                 assert check["kw"]["dbdir"] == paths.PKI_TOMCAT_ALIAS_DIR
                 assert check["kw"]["msg"] == error_msg
 
-    def test_ipa_healthcheck_expiring(self):
+    def test_ipa_healthcheck_expiring(self, restart_service):
         """
         There are two overlapping tests for expiring certs, check both.
         """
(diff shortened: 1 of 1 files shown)


rcritten:
This could be an interesting problem since ipa-healthcheck determines its expiring dates from certmonger itself so shutting it down won't do anything, and in fact could make things worse, since on startup it refreshes all certs.

rcritten:
Stop the CA instead. No CA == no renewal. ipa-healthcheck doesn't care about the request status.

flo-renaud:
> Stop the CA instead. No CA == no renewal. ipa-healthcheck doesn't care about the request status.

I came to the same conclusion, especially since running `getcert list` would restart certmonger. The fix LGTM and the [test](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/945152b2-e0b6-11ea-b9fc-fa163e07b2f0) is green, I am just waiting for azure pipeline to complete.

flo-renaud:
@rcritten you can remove the temp commit and I will ACK.

kaleemsiddiqu:
@rcritten 
Please update the commit message as we are now stopping the CA server instead of certmonger

rcritten:
Temp commit dropped. The commit message is ok, I just hadn't updated the description in the PR.

kaleemsiddiqu:
> Temp commit dropped. The commit message is ok, I just hadn't updated the description in the PR.

Ah my bad.. yes thats correct.

flo-renaud:
Temp commit was removed, tests are green and commit msg is OK, ACK.

rcritten:
master:

* 454e023ad8f0821e8a7c2400485a4d270cf4d3ea ipatests: stop the CA during healthcheck expiration test

PR #5044, opened 2020-08-18
pylint: Fix several warnings on dev branch

stanislavlevin:
- teach pylint about more RRs types and dns flags
- fix warning W0612(unused-variable)

Fixes: https://pagure.io/freeipa/issue/8468


abbra:
LGTM. Do we need to keep the temporary commit for the latest pylint?

stanislavlevin:
@abbra, this is proof of fixes.
We don't need pypi's pylint, but of course, new commits may introduce new pylint warnings/errors which may not be caught by current(F32) pylint.

stanislavlevin:
The temp commit was removed.

stanislavlevin:
The dns part could be applied to `ipa-4-6` branch if `dnspython` 2.0+ is planned for landing for that distros.

abbra:
At this point there are no plans to have dnspython update in RHEL 7. I think we can skip `ipa-4-6`.

PR #5045, opened 2020-08-18
Improve performance of ipa-server-guard

rcritten:
* Drop support for python 2
* Only import traceback and syslog when needed
* Only import ipaserver.install.certs when the lock is needed
* Only import ipautil when run is needed

For the unsupported operations case this improves performance by
95%

For the supported operations that don't require a lock the
improvement is about 50%.

For the supported operations that require a lock the improvement
is about 20%

When configuring a CA certmonger calls its helper with the
following operations:

IDENTIFY
FETCH-ROOTS
GET-SUPPORTED-TEMPLATES
GET-DEFAULT-TEMPLATE
GET-NEW-REQUEST-REQUIREMENTS
GET-RENEW-REQUEST-REQUIREMENTS
FETCH-SCEP-CA-CAPS
FETCH-SCEP-CA-CERTS

Only IDENTIFY, FETCH-ROOTS and GET-NEW-REQUEST-REQUIREMENTS are
supported by ipa-submit, along with the request options SUBMIT and
POLL.

Which means every time the IPA CA in certmonger is updated
eight calls to ipa-server-guard are made so the savings are
cumulative.

The savings when executing these eight operations is a 73% decrease
(.7 sec vs 2.5 sec).

https://pagure.io/freeipa/issue/8425

### DIFF
--- a/install/certmonger/ipa-server-guard.in
+++ b/install/certmonger/ipa-server-guard.in
@@ -19,25 +19,29 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from __future__ import print_function
-
 import os
 # Prevent garbage from readline on standard output
 # (see https://fedorahosted.org/freeipa/ticket/4064)
 if not os.isatty(1):
     os.environ['TERM'] = 'dumb'
 import sys
-import syslog
-import traceback
-
-import six
-
-from ipapython import ipautil
-from ipaserver.install import certs
 
 # Return codes. Names of the constants are taken from
 # https://git.fedorahosted.org/cgit/certmonger.git/tree/src/submit-e.h
 OPERATION_NOT_SUPPORTED_BY_HELPER = 6
+
+
+def run_operation(cmd):
+    from ipapython import ipautil
+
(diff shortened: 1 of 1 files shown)


rcritten:
I'm not sure if I'm being too clever by half with the imports but this paired with a certmonger that doesn't actively kill its children is faster than without the changes.

abbra:
LGTM

abbra:
I don't see anything that could be bad here. PR CI also succeeds with older certmonger. Let's merge?

flo-renaud:
@rcritten 
thanks for the fix, I agree with the strategy to take the lock only for specific operations. I tested by running in loop `getcert modify-ca -c IPA -e '/usr/libexec/certmonger/ipa-server-guard /usr/libexec/certmonger/ipa-submit'; cat /run/ipa/renewal.lock`, and the issue happens less often than without the fix.
In order to completely fix the problem we probably also need certmonger's fix (no SIGKILL) but this PR is already a step in the right direction and can be pushed IMO.

rcritten:
master:

* 18a8a4158051cf6b5b809605dd88dd8090f3485f Improve performance of ipa-server-guard

PR #5052, opened 2020-08-21
Fix resolv conf on smbserver test

abbra:
test_smb test suite sets up IPA master, AD forest, and two clients. The clients are used as an SMB server and an SMB client and they need to resolve and authenticate AD users with Kerberos.

Previously, the test only configured SMB client to use IPA master as its DNS server. SMB server wasn't using IPA master and thus any attempt to resolve SRV records from AD DNS zone was failing.

Make sure that both SMB client's and SMB server's DNS resolution is set up in the same way.

Fixes: https://pagure.io/freeipa/issue/8344


fcami:
LGTM. Waiting for CI.

wladich:
master:

* 0a01f576e0b32b38bfd57498d154a061f84ff728 test_smb: make sure both smbserver and smbclient use IPA master for DNS

PR #5029, opened 2020-08-12
[WIP] Add new krbtpolicy for lifetime jitter 

rebeccc:
I'm having difficulty setting up my test environment and want to check if my tests work against the PR tests. I'll update and clean up this PR when I confirm my tests work.

PR #5055, opened 2020-08-21
Add krbtpolicy for jittered lifetime

rebeccc:
Continuation of #5029 

This KDC extension will create a jittered lifetime for services with a lifetime greater than 60 minutes. The lifetime will be set to a random number between max_life - 60 minutes and max_life.


rcritten:
5044 was merged so if you rebase your PR may pass testing.

doc/designs/ticket-lifetime-jitter.md
```diff
@@ -0,0 +1,14 @@
+# Ticket Lifetime Jitter
+
+## Overview
+
+Ticket lifetimes can be jittered so that renewals / re-issues do not overwhelm the KDC at a certain moment. 
+The feature is enabled automatically so we can avoid triggering an LDAP query on every AS_REQ and TGS_REQ.
```
frozencemetery, line 6: I'm not sure this is quite right anymore.  The spirit is correct (i.e., that it's more expensive to check if we should jitter rather than just jitter), but I think in our discussion we decided it doesn't save an LDAP query.

ipatests/test_integration/test_krbtpolicy.py
```diff
@@ -177,3 +177,54 @@ def test_krbtpolicy_otp(self):
...
+        master.run_command(['kinit', USER1], stdin_text=PASSWORD + '\n')
+        result = master.run_command('klist | grep krbtgt')
+
+        assert maxlife_within_policy(result.stdout_text, MAXLIFE,
+                                     slush=60) is True
+
+        reset_to_default_policy(master, USER1)
+        master.run_command(['ipa', 'config-mod', '--user-auth-type='])
+
+        """ Test jitter lifetime for pkinit """
```
frozencemetery, line 213: Probably should be a normal comment (or separate function, depending)?

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
```diff
@@ -56,6 +56,8 @@ ipa_kdcpolicy_check_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
     
     /* If no mechanisms are set, allow every auth method */
     if (ua == IPADB_USER_AUTH_NONE) {
+	/* Set lifetime_out to a random value in the last hour of lifetime, between 23 and 24 hours */
+        *lifetime_out = (86400 - IPA_KRB_JITTER_WINDOW) + (rand() % IPA_KRB_JITTER_WINDOW);
```
frozencemetery, line 60: Might be worth using getrandom() or OSSL_rand() here?  rand() is fine for this application, but static analyzers are going to complain later :)

doc/designs/ticket-lifetime-jitter.md
rebeccc: True, I hadn't read this document in a while! I removed the line in the upcoming commit.

ipatests/test_integration/test_krbtpolicy.py
rebeccc: Moved to its own function, but it's very redundant since it's exactly the same as the otp function. Should I only have one function, or should I keep both?

daemons/ipa-kdb/ipa_kdb_kdcpolicy.c
rebeccc: I used getrandom(), but can change to RAND_bytes() if anyone thinks that's better.

frozencemetery:
Hi @rebeccc, I see replies to comments but no code change.  Can you push the changes when they're ready?

rcritten:
Per @frozencemetery he is going to submit a new PR with additional changes before EOY. Closing this PR in advance of that.

PR #5063, opened 2020-08-24
When parsing options require name/value pairs

rcritten:
If single-option values are combined together with invalid options
an exception would be raised.
    
For example -verbose was treated as -v -e rbose. Since rbose isn't
a name/value pair things would blow up. This is now caught and
a somewhat more reable error returned. The -v and -e are consumed,
not much we can do about that, but at least a more usable error is
returned.
    
https://pagure.io/freeipa/issue/6115

### DIFF
--- a/ipalib/plugable.py
+++ b/ipalib/plugable.py
@@ -600,13 +600,18 @@ class API(ReadOnly):
             assert type(options.env) is list
             for item in options.env:
                 try:
-                    (key, value) = item.split('=', 1)
+                    values = item.split('=', 1)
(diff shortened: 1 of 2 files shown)


abbra:
LGTM. ACK

rcritten:
master:

* 4a89da5352314cad4417ee42eac2cd8ad7cdc057 ipatests: Add option/arg parsing tests for the cli
* 12dfb0fc96ad6db3aa2cd5a56b6cbc8cd092751a cli: When parsing options require name/value pairs

PR #5067, opened 2020-08-27
    Require at least 2Gb of available RAM to install the server

rcritten:
Require at least 2Gb of available RAM to install the server
    
Verify that there is at least 2Gb of usable RAM on the system. Swap
is not considered. While swap would allow a user to minimally install
IPA it would not be a great experience.

Using any proc-based method to check for available RAM does not
work in containers unless /proc is re-mounted so use cgroups
instead. This also handles the case if the container has memory
constraints on it (-m).
    
Add a switch to skip this memory test if the user is sure they
know what they are doing.
    
https://pagure.io/freeipa/issue/8404

NOTE: I'm not aware of a pythonic "is this in a contaner" so I rolled my own.

IPA is actually installable, with a CA in < 2GB of RAM but things are very tight
and the end result isn't all that useful. 2GB should be considered the absolute
minimum. This is not including swap.

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -18,6 +18,7 @@
 #
 
 from __future__ import absolute_import
+from __future__ import division
 from __future__ import print_function
 
 import logging
@@ -28,6 +29,7 @@ import ldif
 import os
 import re
 import fileinput
+import psutil
 import sys
 import tempfile
 import shutil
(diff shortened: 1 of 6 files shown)


rcritten:
Azure's containers have 1.73GB of memory available so the install step is failing. We can either reduce the required limit or pass in the skip flag.

ipaserver/install/installutils.py
```diff
@@ -972,6 +974,71 @@ def check_entropy():
...
+    """Determine if we're running in a container.
+
+       virt-what will return the underlying machine information so
+       isn't usable here.
+
+       systemd-detect-virt requires the whole systemd subsystem which
+       isn't a reasonable require in a container.
+    """
+    with open('/proc/1/sched', 'r') as sched:
+        data_sched = sched.readline()
```
stanislavlevin, line 987: There are envs which mount 'proc' with enabled hidepid option [1]:
> hidepid=n (since Linux 3.3)
      This option controls who can access the information in
      /proc/[pid] directories
    
[1]: https://man7.org/linux/man-pages/man5/proc.5.html

stanislavlevin: Usually I detect this as:
```python
def is_hidepid():
    """Determine if /proc is mounted with hidepid=1/2 option"""
    try:
        os.lstat('/proc/1/stat')
    except (FileNotFoundError, PermissionError):
        return True
    return False
```

stanislavlevin:
In case of Azure, each master reports its memory statistics after the tests completion like:
```
[2020-08-27 15:15:35] + cat /sys/fs/cgroup/memory/memory.memsw.failcnt
[2020-08-27 15:15:35] 0
[2020-08-27 15:15:35] + cat /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes
[2020-08-27 15:15:35] 3984588800
[2020-08-27 15:15:35] + cat /sys/fs/cgroup/memory/memory.memsw.max_usage_in_bytes
[2020-08-27 15:15:35] 1586233344
[2020-08-27 15:15:35] + cat /sys/fs/cgroup/memory/memory.failcnt
[2020-08-27 15:15:35] 0
[2020-08-27 15:15:35] + cat /sys/fs/cgroup/memory/memory.max_usage_in_bytes
[2020-08-27 15:15:35] 1577467904
[2020-08-27 15:15:35] + cat /sys/fs/cgroup/memory/memory.limit_in_bytes
[2020-08-27 15:15:35] 1992294400
```

rcritten:
Thanks. I incorporated that function and dropped the check to 1.8Gb. I'll still report as 2Gb. This is supposed to be a guideline to prevent users from wasting a lot of time installing on underpowered systems. I don't want it to be too much of a hammer.

rcritten:
I squashed the patches and set the advertised minimum to 1.6GB.

abbra:
LGTM

abbra:
master:

* cfad7af35dd5a2cdd4081d1e9ac7c245f47f1dce Require at least 1.6Gb of available RAM to install the server
* fc271a55bb6c942ec5268f190dca72dbb6cb8387 ipatests: Add tests for checking available memory

PR #5080, opened 2020-09-03
ipatests: fix bind service name

flo-renaud:
With the commit 721435c
the service name for bind is now 'named' instead of
'named-pkcs11' on fedora. The ipa-healthcheck test was hardcoding
the service name but it should instead use the name stored in
knownservices.named.systemd_name as it varies depending on
the OS.

Fixes: https://pagure.io/freeipa/issue/8482


mrizwan93 (approved):
LGTM

abbra:

flo-renaud:
Temp commit removed and tests are green

PR #5082, opened 2020-09-08
Run KDC as non-privileged user

tiran:

### DIFF
--- a/install/share/kdc.conf.template
+++ b/install/share/kdc.conf.template
@@ -21,3 +21,9 @@
   spake_preauth_indicator = hardened
   encrypted_challenge_indicator = hardened
  }
+
+[logging]
+ # disable default log file and only use custom log files
+ default = DEVICE:/dev/null
+ admin_server = FILE:$KADMIND_LOG
+ kdc = FILE:$KRB5KDC_LOG
--- a/daemons/ipa-otpd/ipa-otpd.socket.in
+++ b/daemons/ipa-otpd/ipa-otpd.socket.in
@@ -5,6 +5,8 @@ Description=ipa-otpd socket
 ListenStream=@krb5rundir@/DEFAULT.socket
 RemoveOnStop=true
 SocketMode=0600
+SocketUser=krb5kdc
+SocketGroup=krb5kdc
 Accept=true
 
 [Install]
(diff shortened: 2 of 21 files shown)


abbra:
sysaccount for krb5kdc is replicated but UID/GID values in it could be different on replicas. How are you going to handle this?

tiran:
> sysaccount for krb5kdc is replicated but UID/GID values in it could be different on replicas. How are you going to handle this?

See design doc. tl;dr we need to acquire a reserved uid and gid from FPC for the KDC.

install/share/kdc.conf.template
```diff
@@ -21,3 +21,8 @@
   spake_preauth_indicator = hardened
   encrypted_challenge_indicator = hardened
  }
+
+[logging]
+ default = DEVICE:/dev/null
```
frozencemetery, line 26: This probably needs changed, unless I'm missing something?

tiran: This disables the default logger. Logger configuration adds additional log locations and does not remove the default log files ``/var/log/krb5kdc.log``. Since krb5kdc is no longer allowed to write to ``/var/log`` I had to disable the default logger and only enable custom log locations.

frozencemetery: Per man page this shouldn't be necessary - default "Specifies how either daemon performs logging in the absence of relations specific to the daemon".

frozencemetery: In general I don't think I like that the log files move.  This is going to mess with admin expectations, if nothing else, and will not be the "usual" place for them.

stanislavlevin:
> > sysaccount for krb5kdc is replicated but UID/GID values in it could be different on replicas. How are you going to handle this?
> 
> See design doc. tl;dr we need to acquire a reserved uid and gid from FPC for the KDC.

If I understand correctly this can limit the cross-distroness. Right now it looks as Fedora/RHEL specific.

install/share/kdc.conf.template
abbra: We can define 

```
[logging]
kdc = STDERR
admin_server = STDERR
```
then both krb5kdc and kadmind will automatically log into the journal.

abbra:
> > > sysaccount for krb5kdc is replicated but UID/GID values in it could be different on replicas. How are you going to handle this?
> > 
> > 
> > See design doc. tl;dr we need to acquire a reserved uid and gid from FPC for the KDC.
> 
> If I understand correctly this can limit the cross-distroness. Right now it looks as Fedora/RHEL specific.

We can move the sysaccount into cn=config. This would make the object local to a server and can be different on different replicas. As result, it will not be an issue for distributions.

duzda:
Hi @tiran, I assume this is not going to be further worked on. Is this still relevant for the future or this falls more into nice-to-have category?

abbra:
It is still relevant.

PR #5083, opened 2020-09-08
SELinux Policy: let custodia replicate keys

fcami:
Enhance the SELinux policy so that custodia can replicate sub-CA keys and certificates:

    allow ipa_custodia_t self:tcp_socket { bind create };
    allow ipa_custodia_t node_t:tcp_socket node_bind;
    allow ipa_custodia_t pki_tomcat_cert_t:dir remove_name;
    allow ipa_custodia_t pki_tomcat_cert_t:file create;
    allow ipa_custodia_t pki_tomcat_cert_t:file unlink;
    allow ipa_custodia_t self:process execmem;
    
Found by: test_replica_promotion::TestSubCAkeyReplication
    
Fixes: https://pagure.io/freeipa/issue/8488

### DIFF
--- a/selinux/ipa.te
+++ b/selinux/ipa.te
@@ -72,6 +72,9 @@ logging_log_file(ipa_custodia_log_t)
 type ipa_custodia_tmp_t;
 files_tmp_file(ipa_custodia_tmp_t)
 
+type pki_tomcat_cert_t;
+type node_t;
+
 ########################################
 #
 # ipa_otpd local policy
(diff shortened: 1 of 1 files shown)


fcami:
Failed test without the fix: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ea5f1404-f1dd-11ea-b70d-fa163e53827b/

fcami:
successful test with the fix: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/a4c68fec-f1e9-11ea-ab26-fa163eef7a64/

fcami:
Azure errors are expected considering "setenforce 1" will not work there.

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -568,6 +568,18 @@ def check_pki_error(self, host):
         # check for cert/key import error message
         assert self.ERR_MESS not in pki_debug_log
 
+    def switch_selinux_to_enabled(self, host):
+        """Candidate to move to tasks.py
+        """
+        cmd = ["setenforce", "1"]
+        status = host.run_command(cmd)
+        assert "usage:  setenforce" not in status.stderr_text
```
rcritten, line 576: I don't think the assert is needed. If setenforce returns 1 then something went wrong and it will raise an exception in that case.

rcritten:
> Azure errors are expected considering "setenforce 1" will not work there.

It is probably better to add setenforce to ipaplatform. If a platform doesn't support it then you can test whether paths.SETENFORCE Is None or something and make it a no-op rather than failing.

fcami:
@rcritten I do not intend to keep the "setenforce 1" python method. These will be superseded by https://github.com/freeipa/freeipa-pr-ci/issues/391 which is already implemented by @netoarmando at https://github.com/freeipa/freeipa-pr-ci/pull/392
My intent is to have only the "SELinux Policy: let custodia replicate keys" commit merged here. Does that make any sense?

rcritten:
Ok, let's remove the temp commits, this seems reasonable.

rcritten:
master:

* 68328299c881df497dab32b145157d6e0e42d6c6 SELinux Policy: let custodia replicate keys

PR #5084, opened 2020-09-08
Set the certmonger subject with a string, not an object

rcritten:
ipa-server-certinstall goes through a slightly different code path
if the replacement certificate is issued by IPA. This was setting
the subject using cert.subject which is a Name object and not the
string representation of that object. This was failing in the
dbus call to certmonger.

https://pagure.io/freeipa/issue/8204

### DIFF
--- a/ipaserver/install/ipa_server_certinstall.py
+++ b/ipaserver/install/ipa_server_certinstall.py
@@ -170,7 +170,7 @@ class ServerCertInstall(admintool.AdminTool):
         if req_id is not None:
             certmonger.add_principal(
                 req_id, 'HTTP/{host}'.format(host=api.env.host))
-            certmonger.add_subject(req_id, cert.subject)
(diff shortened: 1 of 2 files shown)


frasertweedale:
ACK from me, when temp commit removed and CI green.

PRCI temp_commit job (succeeded): http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/f6503fc0-f225-11ea-8d4d-fa163e19313f

rcritten:
Thanks. Temp commit removed.

flo-renaud:
master:

* 040d48fa61eb4902f43dfe7fa10257de1e3818b6 ipatests: test ipa_server_certinstall with an IPA-issued cert
* f249c51bf4342341378e63953386e9ac4070be54 Set the certmonger subject with a string, not an object

PR #5091, opened 2020-09-10
Don't allow both a zone name and --name-from-ip to be provided

rcritten:
Don't allow both a zone name and --name-from-ip to be provided

--name-from-ip will generate a zone name so there is no point in
the user providing one. If one is provided and doesn't match the
generated name then a validation exception is raised.

https://pagure.io/freeipa/issue/8446


rjeffman (approved):
LGTM.

abbra:
LGTM. ACK.

abbra:
master:

* 2265cb86cf71f7cbdcb969b790991c26491c753c Don't allow both a zone name and --name-from-ip to be provided
* e92a4ba4aeee60b80ac361b9c4858cd54484ace8 ipatests: test that a zone name and name-from-ip will be rejected

PR #5092, opened 2020-09-11
Add user and group wrapers

tiran:
New classes for user and group names provide a convenient way to access
the uid and primary gid of a user / gid of a group. The classes also
provide chown() and chgrp() methods to simplify common operations.

The wrappers are subclasses of builtin str type and behave like ordinary
strings with additional features. The pwd and grp structs are retrieved
once and then cached.

### DIFF
--- a/ipatests/test_ipaplatform/test_constants.py
+++ b/ipatests/test_ipaplatform/test_constants.py
@@ -1,0 +1,53 @@
+#
+# Copyright (C) 2020  FreeIPA Contributors see COPYING for license
+#
+import pytest
+
+from ipaplatform.base.constants import User, Group
+
+
+def test_user_root():
+    user = User("root")
+    assert user == "root"
+    assert str(user) == "root"
+    assert repr(user) == '<User "root">'
+    assert user.uid == 0
+    assert user.pgid == 0
+    assert user.entity.pw_uid == 0
+
+
+def test_user_invalid():
+    invalid = User("invalid")
+    with pytest.raises(ValueError) as e:
+        assert invalid.uid
+    assert str(e.value) == "user 'invalid' not found"
+
+
+def test_group():
+    group = Group("root")
+    assert group == "root"
+    assert str(group) == "root"
+    assert repr(group) == '<Group "root">'
+    assert group.gid == 0
+    assert group.entity.gr_gid == 0
+
+
+def test_group_invalid():
+    invalid = Group("invalid")
+    with pytest.raises(ValueError) as e:
+        assert invalid.gid
+    assert str(e.value) == "group 'invalid' not found"
+
+
+@pytest.mark.skip_if_platform("debian", reason="test is Fedora specific")
+@pytest.mark.skip_if_platform("suse", reason="test is Fedora specific")
+def test_user_group_daemon():
+    # daemon user / group are always defined
+    user = User("daemon")
+    assert user == "daemon"
+    assert user.uid == 2
+    assert user.pgid == 2
+
+    group = Group("daemon")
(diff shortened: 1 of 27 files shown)


rcritten:
nit: typo in commit message, wrapers -> wrappers

ipatests/test_ipaplatform/test_constants.py
```diff
@@ -0,0 +1,24 @@
...
+    assert user.entity.pw_uid == 0
+
+
+def test_group():
+    group = Group("root")
+    assert group == "root"
+    assert str(group) == "root"
+    assert repr(group) == '<Group "root">'
+    assert group.gid == 0
+    assert group.entity.gr_gid == 0
```
rcritten, line 24: This might be distro-specific but in Fedora we can predict the uid of other users, like httpd and dirsrv. Would it be better to check less trivial users as well? Maybe as a separate test with a skip?

rcritten: And a test for a non-existent user and group is needed. It doesn't currently work very nicely. It may be preferable to raise an exception when a user can't be located to make debugging easier.

```
>>> from ipaplatform.base.constants import BaseConstantsNamespace, User, Group
>>> foo = User("notfound")
>>> foo.uid
  File "/home/rcrit/redhat/freeipa/ipaplatform/base/constants.py", line 46, in uid
    return self.entity.pw_uid
  File "/home/rcrit/redhat/freeipa/ipaplatform/base/constants.py", line 39, in entity
    self._entity = entity = pwd.getpwnam(self)
KeyError: 'getpwnam(): name not found: <User "notfound">'
```

rcritten:
I haven't done an exhaustive search so maybe this is a one-off but in ipaserver/install/bindinstance.py::uninstall() there is a call to:

```ipautil.remove_ccache(run_as=self.service_user)```

By default service_user is ```constants.NAMED_USER``` but there is no enforcement that this is a class and not a string (there are no uses that call the class with service_user currently). I suspect it would blow up mightily somewhere in development/testing if a string were passed in so perhaps not a big deal.

Anyway, service_user is passed into ipautil.run() runas=<value> and I think it would fail with the class.

tiran:
> I haven't done an exhaustive search so maybe this is a one-off but in ipaserver/install/bindinstance.py::uninstall() there is a call to:
> 
> `ipautil.remove_ccache(run_as=self.service_user)`
> 
> By default service_user is `constants.NAMED_USER` but there is no enforcement that this is a class and not a string (there are no uses that call the class with service_user currently). I suspect it would blow up mightily somewhere in development/testing if a string were passed in so perhaps not a big deal.
> 
> Anyway, service_user is passed into ipautil.run() runas= and I think it would fail with the class.

This would have worked because a User object is a subclass of str and pwd.getpwnam() accepts subclasses of str. I took the opportunity to teach both cases how to deal with User and Group objects. User and Group typles now return an identical object, e.g. ``User(userobj) is userobj``.

rcritten:
The azure failures seem to be repo-related. I'm not sure if rebasing will help or if it's an issue with azure (or ubuntu).

rcritten:
LGTM, what's the branch target?

tiran:
4.9 is sufficient.

rcritten:
master:

* 72fb4e60c8d8eec854b4986e00037ff3a85e4389 Add user and group wrappers
* b19d20e2db9441334ef412dcdefa06bd5c790216 Use new classes for run_command and Service
* bc128cae471dae6e070953333d5d76f211aa7688 Add User and Group to all ipaplatform.constants

PR #5093, opened 2020-09-11
Only restart DS when duplicate cacert was found

tiran:
The update_fix_duplicate_cacrt_in_ldap plugin no longer restarts DS when
CA is disabled or no duplicate cacert entry was dedected.

Related: https://pagure.io/freeipa/issue/7125


flo-renaud:
@tiran 
thanks for the fix, LGTM.

rcritten:
master:

* 0a2b6ca6ee4d62e5d69cadd9eca21fb95d243c8d Only restart DS when duplicate cacrt was found

PR #5094, opened 2020-09-11
De-duplicate ACI attributes and permissions

rcritten:
De-duplicate ACI attributes and permissions

Ensure uniqueuess in attributes and permissions in the ACI class.

A set() is not used because it doesn't guarantee order which ends up
causing cascading and unpredictable test failures. Since all we
really need is de-duplication and not a true mathematical set iterating
through the list is sufficiently fast, particularly since the number
of elements will always be low.

https://pagure.io/freeipa/issue/8443

### DIFF
--- a/ipalib/aci.py
+++ b/ipalib/aci.py
@@ -83,7 +83,7 @@ class ACI:
             op = v['operator']
             if type(v['expression']) in (tuple, list):
                 target = ""
-                for l in v['expression']:
+                for l in self._unique_list(v['expression']):
                     target = target + l + " || "
                 target = target[:-4]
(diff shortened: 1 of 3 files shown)


abbra:
LGTM. ACK.

abbra:
master:

* cdf830af189ceac4b0e8606fb3b538d17d90aaf6 De-duplicate ACI attributes and permissions
* 2656c4687b09f8eb2fe847e5e9901df1dbb3335e Use ACI class set_permissions() method to set permissions
* 2e4431af7066ce298008973b6caddc170645c98d ipatests: Add test for ACI attribute and permission uniqueness

PR #5096, opened 2020-09-11
Add index for more trust-related attributes

rcritten:
Add index for ipaNTTrustPartner, ipaNTSecurityIdentifier and
krbprincipalname

https://pagure.io/freeipa/issue/8491

### DIFF
--- a/install/updates/20-indices.update
+++ b/install/updates/20-indices.update
@@ -395,3 +395,24 @@ default: objectClass: top
 default: objectClass: nsIndex
 default: nsSystemIndex: false
 default: nsIndexType: eq
+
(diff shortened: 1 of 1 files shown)


rcritten:
WIP since there is still one un-handled indexed search in the ticket.

abbra:
LGTM.

abbra:
We can ignore the remaining unindexed search because indexing `cn'` attribute is not going to make things fast, as you pointed out. So it is OK to go just with this PR.

abbra:
master:

* 20b55f4017ab42113f1ced829a4b4afa17839b55 Add index for more trust-related attributes

PR #5102, opened 2020-09-14
Add ldap_update() helper to service class

tiran:
The new _ldap_update() helper methods makes it easier to apply LDAP
update files from a service instance.

### DIFF
--- a/ipaserver/install/krainstance.py
+++ b/ipaserver/install/krainstance.py
@@ -34,7 +34,6 @@ from ipapython import ipautil
 from ipapython.dn import DN
 from ipaserver.install import cainstance
 from ipaserver.install import installutils
-from ipaserver.install import ldapupdate
 from ipaserver.install.dogtaginstance import DogtagInstance
 from ipaserver.plugins import ldap2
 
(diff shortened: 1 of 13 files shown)


ipaserver/install/krainstance.py
```diff
@@ -274,13 +273,7 @@ def __add_vault_container(self):
...
 
     def __apply_updates(self):
-        sub_dict = {
-            'SUFFIX': self.suffix,
-        }
-
-        ld = ldapupdate.LDAPUpdate(dm_password=self.dm_password,
-                                   sub_dict=sub_dict)
-        ld.update([os.path.join(paths.UPDATES_DIR, '40-vault.update')])
+        self._ldap_update(['40-vault.update'])
```
rcritten, line 276: Was using DM here a relic of the past? It looks like dm_password in the class isn't used at all any more and it assumes ldapi=True because that's the URI that is generated. Would it make sense to clean that up as well?

tiran: I came to the same conclusion. The argument is passed to the class and assigned to an instance var but the var is never used. I can remove the code.

rcritten:
These changes look good to me but it changes the API of the LDAPUpdate class. Do we think that will cause issues with any 3rd party who may be using it? Should we go the Deprecation route instead?

rcritten:
LGTM, what's the target release for the changes?

tiran:
Let's apply the patch to 4.9 only. I don't think that I'm going to need the change in 4.8.

rcritten:
master:

* 87cf2a3c789929337abc3ae38192ee64cfd6c9e3 Add ldap_update() helper to service class
* 99a40cbbe95b629ccfd49cb3c809889731148779 Simplify LDAPUpdater

PR #5107, opened 2020-09-15
[Container] Unify access to FQDN

tiran:
FreeIPA's Python and C code used different approaches to get the FQDN of
the host. Some places assumed that gethostname() returns a FQDN. Other
code paths used glibc's resolver to resolve the current node name to a
FQDN.

Python code now uses the ipalib.constants.FQDN where a fully qualified
domain name is expected. The variable is initialized only once and avoids
potential DNS lookups.

C code uses a new helper function ipa_gethostfqdn() in util package. The
function implements similar logic as gethostfqdn() except it uses more
modern getaddrinfo(). The result is cached as well.

Fixes: https://pagure.io/freeipa/issue/8501

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb.c
+++ b/daemons/ipa-kdb/ipa_kdb.c
@@ -25,6 +25,7 @@
 
 #include "ipa_kdb.h"
 #include "ipa_krb5.h"
+#include "ipa_hostname.h"
 
 #define IPADB_GLOBAL_CONFIG_CACHE_TIME 60
 
@@ -535,9 +536,9 @@ static krb5_error_code ipadb_init_module(krb5_context kcontext,
 {
     struct ipadb_context *ipactx;
     krb5_error_code kerr;
+    char hostname[IPA_HOST_NAME_LEN];
     int ret;
     int i;
-    struct utsname uname_data;
 
     /* make sure the context is freed to avoid leaking it */
     ipactx = ipadb_get_context(kcontext);
@@ -603,13 +604,12 @@ static krb5_error_code ipadb_init_module(krb5_context kcontext,
         goto fail;
     }
 
-    ret = uname(&uname_data);
-    if (ret) {
-        ret = EINVAL;
+    ret = ipa_gethostfqdn(hostname);
+    if (ret != 0) {
+        ret = errno;
         goto fail;
     }
-
-    ipactx->kdc_hostname = strdup(uname_data.nodename);
+    ipactx->kdc_hostname = strdup(hostname);
     if (!ipactx->kdc_hostname) {
         ret = ENOMEM;
         goto fail;
--- a/daemons/ipa-kdb/ipa_kdb.c
+++ b/daemons/ipa-kdb/ipa_kdb.c
@@ -53,7 +53,7 @@ static void ipadb_context_free(krb5_context kcontext,
         free((*ctx)->base);
         free((*ctx)->realm_base);
         free((*ctx)->accounts_base);
-        free((*ctx)->kdc_hostname);
+        /* kdc_hostname is a pointer to a static char[] */
         /* ldap free lcontext */
         if ((*ctx)->lcontext) {
             ldap_unbind_ext_s((*ctx)->lcontext, NULL, NULL);
@@ -536,7 +536,6 @@ static krb5_error_code ipadb_init_module(krb5_context kcontext,
 {
     struct ipadb_context *ipactx;
     krb5_error_code kerr;
-    char hostname[IPA_HOST_NAME_LEN];
     int ret;
     int i;
 
@@ -604,14 +603,9 @@ static krb5_error_code ipadb_init_module(krb5_context kcontext,
         goto fail;
     }
 
-    ret = ipa_gethostfqdn(hostname);
-    if (ret != 0) {
-        ret = errno;
-        goto fail;
-    }
-    ipactx->kdc_hostname = strdup(hostname);
+    ipactx->kdc_hostname = ipa_gethostfqdn();
     if (!ipactx->kdc_hostname) {
-        ret = ENOMEM;
+        ret = errno;
         goto fail;
     }
 
--- a/daemons/ipa-sam/ipa_sam.c
+++ b/daemons/ipa-sam/ipa_sam.c
@@ -4441,8 +4441,7 @@ static char *sec_key(TALLOC_CTX *mem_ctx, const char *d)
 
 static NTSTATUS save_sid_to_secret(struct ipasam_private *ipasam_state)
 {
-	char hostname[IPA_HOST_NAME_LEN];
-	int ret;
+	const char *hostname;
 	char *p;
 	TALLOC_CTX *tmp_ctx;
 	NTSTATUS status = NT_STATUS_UNSUCCESSFUL;
@@ -4467,8 +4466,8 @@ static NTSTATUS save_sid_to_secret(struct ipasam_private *ipasam_state)
 		goto done;
 	}
 
-	ret = ipa_gethostfqdn(hostname);
-	if (ret == -1) {
+	hostname = ipa_gethostfqdn();
+	if (hostname == NULL) {
 		DEBUG(1, ("ipa_gethostfqdn failed.\n"));
 		status = NT_STATUS_UNSUCCESSFUL;
 		goto done;
@@ -4721,10 +4720,9 @@ static int bind_callback(LDAP *ldap_struct, struct smbldap_state *ldap_state, vo
 static NTSTATUS ipasam_generate_principals(struct ipasam_private *ipasam_state) {
 
 	krb5_error_code rc;
-	int ret;
 	krb5_context context;
 	NTSTATUS status = NT_STATUS_UNSUCCESSFUL;
-	char hostname[IPA_HOST_NAME_LEN];
+	const char *hostname;
 	char *default_realm = NULL;
 
 	if (!ipasam_state) {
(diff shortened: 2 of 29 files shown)


rcritten:
```client/ipa-join.c``` and ```daemons/ipa-kdb/ipa_kdb.c``` call ```uname(2)``` to get the hostname.

In the case of ipa-join the man page will need to be updated as well since it is explicitly notes that uname() is used.

**Comment by tiran** [comment]
> `client/ipa-join.c` and `daemons/ipa-kdb/ipa_kdb.c` call `uname(2)` to get the hostname.
> 
> In the case of ipa-join the man page will need to be updated as well since it is explicitly notes that uname() is used.

Thanks! I missed the fact that some code uses uname()'s nodename. I only grepped for ``gethostname()`` and ``getfqdn``. I have updated the PR and simplified ``ipa-join``.

File: ipapython/fqdn.py [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+"""
+import socket
+
+
+def gethostfqdn():
+    hostname = socket.gethostname()
+
+    # optional optimization, consider hostname with dot as FQDN
+    if "." in hostname:
+        return hostname
```
frasertweedale, line 14: Hmm... is this a safe assumption?

frasertweedale (commented):
Initial comments, but I haven't looked at the whole diff yet (let alone tested it).  So there will be more later.

ipapython/fqdn.py
tiran: No, it's actually not safe. hostname may be partly qualified. In that case glibc/NSS can return a FQDN canonical name.

daemons/ipa-sam/ipa_sam.c
```diff
@@ -4466,13 +4467,12 @@ static NTSTATUS save_sid_to_secret(struct ipasam_private *ipasam_state)
 		goto done;
 	}
 
-	ret = gethostname(hostname, sizeof(hostname));
+	ret = ipa_gethostfqdn(hostname);
 	if (ret == -1) {
 		DEBUG(1, ("gethostname failed.\n"));
```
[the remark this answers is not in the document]
tiran: Yes, good catch!

daemons/ipa-sam/ipa_sam.c
```diff
@@ -4736,12 +4736,11 @@ static NTSTATUS ipasam_generate_principals(struct ipasam_private *ipasam_state)
 		return status;
 	}
 
-	ret = gethostname(hostname, sizeof(hostname));
+	ret = ipa_gethostfqdn(hostname);
 	if (ret == -1) {
 		DEBUG(1, ("gethostname failed.\n"));
```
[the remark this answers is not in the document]
tiran: Yes, good catch!

frasertweedale (commented):
Finished eyeballing the C code; looking at Python code now.

ipaserver/install/installutils.py
```diff
@@ -118,16 +118,16 @@ def __init__(self, top_dir=None):
...
-    except Exception:
-        try:
-            fqdn = socket.gethostname()
-        except Exception:
-            fqdn = ""
-    return fqdn
+    """Get fully qualified domain name of current host
+
+    :note: used by ansible_freeipa
+    :deprecated: use ipalib.constants.FQDN
```
frasertweedale, line 126: :+1:

ipapython/fqdn.py
```diff
@@ -0,0 +1,29 @@
...
+        None,  # service/port is irrelevant
+        family=socket.AF_UNSPEC,  # IPv4 or IPv6
+        type=socket.SOCK_DGRAM,  # optimization, TCP/RAW gives same result
+        # include canonical name in first addrinfo struct
+        # only use address family when at least one non-local interface
+        # is configured with that address family
+        flags=socket.AI_CANONNAME | socket.AI_ADDRCONFIG
+    )
+    # first addrinfo struct, fourth field is canonical name
+    return gai[0][3]
```
frasertweedale, line 29: We should check `len(gai) > 0` and either fail with a more descriptive exception than `IndexError`, or return `hostname` (even though it may not be fully qualified).  If nssswitch is misconfigured, we should at least try and be helpful instead of failing with a cryptic error :)

frasertweedale (dismissed):
Finished reviewing the patch; haven't tested yet.

ipapython/fqdn.py
tiran: I don't see how ``gai`` can be empty at this point. ``getaddrinfo()`` either returns at least one result or it fails with an error. All C code I have read recently always assumes that the result struct contains valid data on success.

util/ipa_hostname.c
```diff
@@ -0,0 +1,111 @@
...
+{
+    static char cached_fqdn[IPA_HOST_NAME_LEN] = {0};
+
+    if (!cached_fqdn) {
+        int res = _get_fqdn(cached_fqdn);
+        if (res != 0) {
+            return -1;
+        }
+    }
+    strcpy(name, cached_fqdn);
```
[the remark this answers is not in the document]
tiran: It's a bit more radical than my initial proposal but I like it :+1:. The library is not used by any code outside FreeIPA so we are good to modify the code in the future w/o ABI issues.

ipapython/fqdn.py
frasertweedale: OK.  I prefer to be defensive ("if the return type admits it, I must handle it"), but I will leave it up to you.

frasertweedale:
I have tested server install (not in container) and it works.  Not much testing, but something.  I will not have time to do more thorough testing before PTO.  We should test regression test trusts and, ipa-otpd->RADIUS before merging (if not adequately covered by PR-CI).  We should also test these features in a shortname environment.  I'm primarily concerned about avoiding regressions, but unless we confirm it achieves our objectives we should hold off merging it.

util/ipa_hostname.c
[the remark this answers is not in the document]
tiran: I have changed the function signature to ``const char* ipa_gethostfqdn(void)``.

util/ipa_hostname.c
```diff
@@ -0,0 +1,111 @@
...
+        freeaddrinfo(ai);
+    }
+    return -1;
+}
+
+int ipa_gethostfqdn(char *name)
+{
+    static char cached_fqdn[IPA_HOST_NAME_LEN] = {0};
+
+    if (!cached_fqdn) {
```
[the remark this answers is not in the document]
tiran: I also wonder why this has worked ... anyhow it's now fixed. Thanks!

daemons/ipa-kdb/ipa_kdb.c
```diff
@@ -603,13 +604,12 @@ static krb5_error_code ipadb_init_module(krb5_context kcontext,
...
-    if (ret) {
-        ret = EINVAL;
+    ret = ipa_gethostfqdn(hostname);
+    if (ret != 0) {
+        ret = errno;
         goto fail;
     }
-
-    ipactx->kdc_hostname = strdup(uname_data.nodename);
+    ipactx->kdc_hostname = strdup(hostname);
```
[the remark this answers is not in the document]
tiran: Function signature is now ``const char* ipa_gethostfqdn(void)`` and ``ipadb_context_free`` no longer frees the hostname.

ipapython/fqdn.py
tiran: I have added a comment that explains why this line cannot fail.

frasertweedale:
@tiran 

The tests that were previously failing in #5201 now seem to be happy with this following patch.
Could you please add it to your PR (you can fix-up the tip commit "Easier to use ipa_gethostfqdn()", or keep it as a separate patch, whatever you prefer).

```patch
commit 031f62aa7d04d46f15d3b7aafb804e63632d1e93
Author: Fraser Tweedale <ftweedal@redhat.com>
Date:   Fri Oct 23 01:06:15 2020 +1100

    ipa_sam: do not clobber static buffer holding fqdn

diff --git a/daemons/ipa-sam/ipa_sam.c b/daemons/ipa-sam/ipa_sam.c
index 6993152e7..0ce858a6f 100644
--- a/daemons/ipa-sam/ipa_sam.c
+++ b/daemons/ipa-sam/ipa_sam.c
@@ -4441,7 +4441,7 @@ static char *sec_key(TALLOC_CTX *mem_ctx, const char *d)
 
 static NTSTATUS save_sid_to_secret(struct ipasam_private *ipasam_state)
 {
-       const char *hostname;
+       char hostname[IPA_HOST_FQDN_LEN];
        char *p;
        TALLOC_CTX *tmp_ctx;
        NTSTATUS status = NT_STATUS_UNSUCCESSFUL;
@@ -4466,12 +4466,13 @@ static NTSTATUS save_sid_to_secret(struct ipasam_private *ipasam_state)
                goto done;
        }
 
-       hostname = ipa_gethostfqdn();
-       if (hostname == NULL) {
+       const char *fqdn = ipa_gethostfqdn();
+       if (fqdn == NULL) {
                DEBUG(1, ("ipa_gethostfqdn failed.\n"));
                status = NT_STATUS_UNSUCCESSFUL;
                goto done;
        }
+       strncpy(hostname, fqdn, sizeof(hostname));
        p = strchr(hostname, '.');
        if (p != NULL) {
                *p = '\0';
```

frasertweedale:
master:

* e28ec76898a0240e416910b9cfea24e82e7d91ba Unify access to FQDN
* 5d4ed65b83f6ed8bb7f3da05561698cbea692bb3 Replace nodename with ipa_gethostfqdn()
* b66b961fdd3d0b7f3c37bd245decf700e720708c Remove problematic optimization from gethostfqdn()
* 3d796a7e5102eded524e4847f8f9dd26f508ae94 Update debug strings to reflect new calls
* 727a2ffb9336a050c26c790989edaf41864c6911 Easier to use ipa_gethostfqdn()
* 3f59118ffcce07cbd74360f9ce8047e0b35960be (HEAD) ipa_sam: do not modify static buffer holding fqdn

PR #5111, opened 2020-09-16
Don't create DS SSCA and self-signed cert

tiran:
Instruct lib389 to not create its self-signed CA and temporary
self-signed certificate. FreeIPA uses local connections and Unix socket
for bootstrapping.

Fixes: https://pagure.io/freeipa/issue/8502

### DIFF
--- a/ipaserver/install/dsinstance.py
+++ b/ipaserver/install/dsinstance.py
@@ -544,6 +544,7 @@ class DsInstance(service.Service):
         slapd_options = Slapd2Base(logger)
         slapd_options.set('instance_name', self.serverid)
         slapd_options.set('root_password', self.dm_password)
+        slapd_options.set('self_sign_cert', False)
         slapd_options.verify()
         slapd = slapd_options.collect()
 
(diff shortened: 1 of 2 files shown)


ipaserver/install/dsinstance.py
```diff
@@ -544,6 +544,8 @@ def __create_instance(self):
         slapd_options = Slapd2Base(logger)
         slapd_options.set('instance_name', self.serverid)
         slapd_options.set('root_password', self.dm_password)
+        # don't create SSCA and temporary self-signed certificate
+        slapd_options.set('self_sign_cert', False)
```
rcritten, line 548: It's fine I guess but do we need a comment here? Is the option self-explanatory enough?

rcritten:
LGTM, test passes. Nit/question about an in-line comment.

tiran:
I have removed the comment and changed backup label to 4.9. The change is not strictly necessary in 4.8 and 4.9 is scheduled to be released soonish.

rcritten:
Thanks, ack.

rcritten:
master:

* 3c86baf0ade9c9045cc0f3b3c7fb0e7c580e29a4 Don't create DS SSCA and self-signed cert

PR #5116, opened 2020-09-17
Add timings to install logs

tiran:
The logging manager now adds timings for installation steps to the
installer logs. The information can be extracted and dumped to a CSV
file with a simple grep command:

    grep -Po 'TIMING: \K.*' /var/log/ipaserver-install.log > ipaserver.csv

### DIFF
--- a/install/tools/ipa-ca-install.in
+++ b/install/tools/ipa-ca-install.in
@@ -316,8 +316,9 @@ def main():
     api.Backend.ldap2.disconnect()
 
     # execute ipactl to refresh services status
+    logger.info("Starting IPA services")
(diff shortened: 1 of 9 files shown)


abbra:
We already have something similar for each installer step. In Azure logs we even collect this information:
```
    echo "Installation complete. Performance of individual steps:"
    grep 'service duration:' /var/log/ipaserver-install.log | sed -e 's/DEBUG //g'


```

rcritten:
A rebase is needed to pick up the updated dnf package for Azure to pass.

abbra:
Not stale but a rebase is needed.

PR #5118, opened 2020-09-18
test_smb: mark test_smb_service_s4u2self as xfail for fed31 

flo-renaud:
The test test_integration/test_smb.py::TestSMB::test_smb_service_s4u2self
is expected to fail in Fedora <= 31 as it requires krb >= 1.18
that is shipped from fedora 32 only

Mark the test as conditionally failing depending on the fedora version.

Fixes: https://pagure.io/freeipa/issue/8505

### DIFF
--- a/ipatests/test_integration/test_smb.py
+++ b/ipatests/test_integration/test_smb.py
@@ -17,6 +17,7 @@ import pytest
 
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
+from ipaplatform.osinfo import osinfo
 from ipaplatform.paths import paths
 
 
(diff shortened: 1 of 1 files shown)


rcritten:
This looks fine to me, you can remove the temporary commit.

I just wonder if we should do xfail or skip in this case. If we know it is going to fail, and will always fail because F31 is near EOL, would skip be better?

rcritten:
Not a deal breaker, I'm fine wither with xfail or skip, but skip saves a few CPU cycles by not running the test at all.

flo-renaud:
@rcritten I followed your advice and replace xfail with skipif. For future reference, tests results are available here:
- [fedora-latest/test_smb](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/21c784fc-fbdf-11ea-af11-fa163e168469): test_smb_service_s4u2self passed
- [fedora-previous/test_smb](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ead81920-fbde-11ea-9cb0-fa163ebb6b2b): test_smb_service_s4u2self skipped

Now removing the temp commit.

rcritten:
master:

* 8ba15027d4fdca6aabda00494d29a528e1ce53d4 test_smb: skip test_smb_service_s4u2self for fed31

PR #5119, opened 2020-09-18
Require an ipa-ca SAN on 3rd party certs if ACME is enabled

rcritten:
Require an ipa-ca SAN on 3rd party certs if ACME is enabled

ACME requires an ipa-ca SAN to have a fixed URL to connect to.
If the Apache certificate is replaced by a 3rd party cert then
it must provide this SAN otherwise it will break ACME.

Add a status option to ipa-acme-manage.

https://pagure.io/freeipa/issue/8498

Marking as ipa-next since I'm sure yet if ACME is going to be backported to ipa-4-8.

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -31,6 +31,7 @@ import ldap
 import os
 import re
 import shutil
+import ssl
 import sys
 import syslog
 import time
@@ -2269,6 +2270,30 @@ def import_ra_key(custodia):
     custodia.import_ra_key()
     CAInstance._set_ra_cert_perms()
     CAInstance.configure_agent_renewal()
+
+
+def check_ipa_ca_san(cert):
+    """
+    Test whether the certificate has an ipa-ca SAN
+
+    :param cert: x509.IPACertificate
+
+    This SAN is necessary for ACME.
+
+    The caller is responsible for initializing the api.
+
+    On success returns None, on failure raises ValidationError
+    """
+    expect = f'{ipalib.constants.IPA_CA_RECORD}.' \
+             f'{ipautil.format_netloc(api.env.domain)}'
+
+    try:
+        cert.match_hostname(expect)
+    except ssl.CertificateError:
+        raise errors.ValidationError(
+            name='certificate',
+            error='Does not have a \'{}\' SAN'.format(expect)
+        )
 
 
 if __name__ == "__main__":
--- a/ipatests/test_integration/test_acme.py
+++ b/ipatests/test_integration/test_acme.py
@@ -11,7 +11,9 @@ import pytest
 from ipalib.constants import IPA_CA_RECORD
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
+from ipatests.test_integration.test_caless import CALessBase, ipa_certs_cleanup
 from ipaplatform.osinfo import osinfo
+from ipaplatform.paths import paths
 from ipaserver.install import cainstance
 
 
@@ -42,9 +44,19 @@ def check_acme_status(host, exp_status, timeout=60):
     return status
 
 
+def server_install_teardown(func):
+    def wrapped(*args):
+        master = args[0].master
+        try:
+            func(*args)
+        finally:
+            ipa_certs_cleanup(master)
+    return wrapped
+
+
 @pytest.mark.skipif(not cainstance.minimum_acme_support(),
                     reason="does not provide ACME")
-class TestACME(IntegrationTest):
+class TestACME(CALessBase):
     """
     Test the FreeIPA ACME service by using ACME clients on a FreeIPA client.
 
@@ -73,15 +85,17 @@ class TestACME(IntegrationTest):
 
     @classmethod
     def install(cls, mh):
+        super(TestACME, cls).install(mh)
+
         # cache the acme service uri
         acme_host = f'{IPA_CA_RECORD}.{cls.master.domain.name}'
         cls.acme_server = f'https://{acme_host}/acme/directory'
 
         # install packages before client install in case of IPA DNS problems
         if not skip_certbot_tests:
-            cls.clients[0].run_command(['dnf', 'install', '-y', 'certbot'])
+            tasks.install_packages(cls.clients[0], ['certbot'])
         if not skip_mod_md_tests:
-            cls.clients[0].run_command(['dnf', 'install', '-y', 'mod_md'])
+            tasks.install_packages(cls.clients[0], ['mod_md'])
 
         tasks.install_master(cls.master, setup_dns=True)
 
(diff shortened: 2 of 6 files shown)


ipatests/test_integration/test_acme.py
```diff
@@ -93,6 +131,9 @@ def test_acme_service_not_yet_enabled(self):
             ['curl', '--fail', self.acme_server],
             ok_returncode=22,
         )
+        result = self.master.run_command(['ipa-acme-manage', 'status'])
+        assert result.returncode == 0
```
mrizwan93, line 135: This check can be removed., the return code for status command will be `0` anyways.

ipatests/test_integration/test_acme.py
```diff
@@ -111,6 +152,10 @@ def test_enable_acme_service(self):
         else:
             raise exc
 
+        result = self.master.run_command(['ipa-acme-manage', 'status'])
+        assert result.returncode == 0
```
mrizwan93, line 156: This check can be removed., the return code for status command will be 0 anyways.

ipatests/test_integration/test_acme.py
```diff
@@ -252,3 +298,55 @@ def test_disable_acme_service(self):
             ['curl', '--fail', self.acme_server],
             ok_returncode=22,
         )
+        result = self.master.run_command(['ipa-acme-manage', 'status'])
+        assert result.returncode == 0
```
mrizwan93, line 302: This check can be removed., the return code for status command will be 0 anyways.

ipatests/test_integration/test_acme.py
```diff
@@ -252,3 +298,55 @@ def test_disable_acme_service(self):
...
+        # which should be ok since ACME is disabled.
+        result = self.certinstall()
+        print(result.stdout_text)
+        print(result.stderr_text)
+        assert result.returncode == 0
+
+        # Enable ACME which should fail since the Apache cert lacks the SAN
+        result = self.master.run_command(['ipa-acme-manage', 'enable'],
+                                         raiseonerr=False)
+        assert result.returncode == 1
```
mrizwan93, line 351: Shall we make `ipa-acme-manage enable` command more robust in this case? cause return code 1 implies `"CA is not installed on this server."`

mrizwan93 (commented):
Hello @rcritten ,
 Please find comments inline.

ipatests/test_integration/test_acme.py
rcritten: Yes, I suppose any exception is sufficient in case something goes wrong.

ipatests/test_integration/test_acme.py
rcritten: I'll look into a new return value for this and ipa-server-certinstall for invalid certificates.

rcritten: I think I'm going to adjust the return values in ipa-acme-manage 0 is success, 1 is an error, and 2 and 3 replace the current 1/2. ipa-server-certinstall only returns 0 == ok, 1 == error. Since ACME hasn't shipped in an official release IMHO it's fine to change this.

ipatests/test_integration/test_acme.py
```diff
@@ -252,3 +296,50 @@ def test_disable_acme_service(self):
...
+    ######################
+    # Require ipa-ca SAN
+    ######################
+
+    @server_install_teardown
+    def test_third_party_certs(self):
+        """Require ipa-ca SAN on replacement web certificates"""
+
+        result = self.master.run_command(['ipa-acme-manage', 'enable'])
+        assert result.returncode == 0
```
mrizwan93, line 311: This can be removed as `raiseonerr=True`  is set by default.

ipatests/test_integration/test_acme.py
```diff
@@ -252,3 +296,50 @@ def test_disable_acme_service(self):
...
+        )
+        assert result.returncode == 0
+
+        # Install using a 3rd party cert with a missing SAN for ipa-ca
+        # which should be rejected.
+        result = self.certinstall()
+        assert result.returncode == 1
+
+        result = self.master.run_command(['ipa-acme-manage', 'disable'])
+        assert result.returncode == 0
```
mrizwan93, line 334: This can be removed as `raiseonerr=True`  is set by default.

ipatests/test_integration/test_acme.py
rcritten: done

ipatests/test_integration/test_acme.py
rcritten: done

ipatests/test_integration/test_acme.py
```diff
@@ -252,3 +296,48 @@ def test_disable_acme_service(self):
...
+
+    ######################
+    # Require ipa-ca SAN
+    ######################
+
+    @server_install_teardown
+    def test_third_party_certs(self):
+        """Require ipa-ca SAN on replacement web certificates"""
+
+        result = self.master.run_command(['ipa-acme-manage', 'enable'])
```
mrizwan93, line 310: variable `result` is not used.

ipatests/test_integration/test_acme.py
```diff
@@ -252,3 +296,48 @@ def test_disable_acme_service(self):
...
+            pin=pin
+        )
+        assert result.returncode == 0
+
+        # Install using a 3rd party cert with a missing SAN for ipa-ca
+        # which should be rejected.
+        result = self.certinstall()
+        assert result.returncode == 1
+
+        result = self.master.run_command(['ipa-acme-manage', 'disable'])
```
mrizwan93, line 332: variable `result` is not used.

mrizwan93 (commented):
Just a nitpick, else LGTM.

ipatests/test_integration/test_acme.py
rcritten: removed

rcritten:
@frasertweedale do you have any opionions?

rcritten:
Marking this as WIP as the status option will be dependent upon https://github.com/freeipa/freeipa/pull/5170

ipaserver/install/cainstance.py
```diff
@@ -2208,6 +2209,44 @@ def import_ra_key(custodia):
     CAInstance.configure_agent_renewal()
 
 
+def has_ipa_ca_san(cert):
```
frasertweedale, line 2212: Bit of a nit, but "has_x" sounds like it would return `bool`.  If you're going to raise on error, maybe "check_x" or "ensure_x" would be better?

ipatests/test_integration/test_acme.py
frasertweedale: I'm fine with that change also.

frasertweedale (commented):
Haven't tested but the approach seems good to me.

ipaserver/install/cainstance.py
rcritten: Good point, switched to check_

rcritten:
Rebased on top of PR #5170 so we can potentially merge this at or near the same time.

flo-renaud:
ACK'ing based on @frasertweedale and @mrizwan93 comments

rcritten:
master:

* 2768b0dbaf0e07bc632a4867b13ef4c5ac875372 Require an ipa-ca SAN on 3rd party certs if ACME is enabled
* e0ff82c884356a6c9fcd0fef66580de30ec5af87 Change the return codes of ipa-acme-manage
* c8f13cd85503f2713c616eccfd7c37e52792c7ef ipatests: Add tests for requiring ipa-ca SAN when ACME is enabled

PR #5124, opened 2020-09-21
Clean up entire /run/ipa/ccaches directory not just files

rcritten:
Remove entire /run/ipa/ccaches directory not just files

If there are any sub-directories in the ccaches directory
then cleaning it up will fail.

Instead remove the whole directory and allow systemd-tmpfiles
to re-create it.

https://pagure.io/freeipa/issue/8248

This seems rather strange, deleting the whole thing then re-creating it, but walking the directory deleting discrete files and directories just seems slower and is quite a bit more code.

### DIFF
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -26,6 +26,7 @@ import glob
 import errno
 import shlex
 import pipes
+import shutil
 import tempfile
 
 from augeas import Augeas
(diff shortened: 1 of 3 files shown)


ipaserver/install/httpinstance.py
```diff
@@ -172,8 +173,8 @@ def remove_httpd_ccaches(self):
         # Make sure that empty env is passed to avoid passing KRB5CCNAME from
         # current env
         ipautil.remove_file(paths.HTTP_CCACHE)
-        for f in os.listdir(paths.IPA_CCACHES):
-            os.remove(os.path.join(paths.IPA_CCACHES, f))
+        shutil.rmtree(paths.IPA_CCACHES)
+        ipautil.run([paths.SYSTEMD_TMPFILES, '--create'])
```
tiran, line 177: You could use systemd-tmpfiles prefix option to only create the directory and any subdirectory. The option is documented as

> Only apply rules with paths that start with the specified prefix.

I verified that ``[paths.SYSTEMD_TMPFILES, '--create', '--prefix', paths.IPA_CCACHES]`` creates a missing paths.IPA_CCACHES directory.

rcritten: Nice catch, I'll add that, thanks.

tiran (approved):
LGTM!

rcritten:
Thanks. Dropped temp commit and rebased to master.

tiran:
master:

* cc5d9a8c9de3db863ca181348bfb01506f74302d Clean up entire /run/ipa/ccaches directory not just files
* 9f9dcfe88acb27c6ac734af18dc4015e8cb5c327 Test that ccaches are cleaned up during installation

PR #5126, opened 2020-09-22
Make git a build requirement

tiran:
FreeIPA uses git in its build process. In the past git was automatically
pulled in. On Fedora 33 builds are failing because git is missing.


rcritten:
IIRC git is only required if IPA_VERSION_IS_GIT_SNAPSHOT is yes.

Production builds of IPA shouldn't require git as a build requirement.

abbra:
LGTM

tiran:
master:

* 644bd0e46b405fde889952eecb2e8ab34913bbc3 Make git a build requirement

PR #5129, opened 2020-09-22
Require a matching server package for the selinux subpackage

rcritten:
Ensure that the selinux subpackage is upgraded along with the
rest of IPA if it is built.

https://pagure.io/freeipa/issue/8511


fcami:
LGTM. ACK for when CI passes.

rcritten:
master:

* 84bbf68781715e2e574d28d7e6d2c6d75afee59e Require a matching server package for the selinux subpackage

PR #5132, opened 2020-09-23
Delay import of psutil to avoid AVC

tiran:
Commit cfad7af35dd5a2cdd4081d1e9ac7c245f47f1dce added a check to ensure a
system has sufficient amount of memory. The feature uses psutil to get
available memory. On import psutil opens files in /proc which can result in
an SELinux violations and Python exception.

     PermissionError: [Errno 13] Permission denied: '/proc/stat'

Fixes: https://pagure.io/freeipa/issue/8512

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -29,7 +29,6 @@ import ldif
 import os
 import re
 import fileinput
-import psutil
 import sys
 import tempfile
 import shutil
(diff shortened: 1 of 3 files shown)


abbra:
LGTM

tiran:
master:

* 80fca8d7010ab8ed2190ef8fb57d882c61e89723 Delay import of psutil to avoid AVC

PR #5144, opened 2020-09-24
Reduce runtime of server installer by about 90 seconds or more

tiran:
This experimental patch speeds up installer by tightening poll/sleep loops, reducing timeouts for DNS and NTP to a sensible value, avoiding duplicate work.

## Reduce CA record DNS timeout to 10s

30 seconds is still a lot of time for a DNS query. Clients typically
do not wait that long. OpenSSH uses 10 seconds for reverse DNS lookup.
That's considered a long timeout already. It's unlikely that a DNS query
is going to succeed after 10 seconds of failed lookups.

At this point during the installer IPA's BIND DNS instance has been
running long enough to be fully available, too.

The changeset reduces installation time by 40 seconds when ipa-ca DNS
has not been created yet.

See: https://pagure.io/freeipa/issue/6176

## Skip duplicate import of cert profiles
    
All supported Dogtag versions import the cert profiles during pkispawn
when using the LDAP profile backend.

This reduces the installation time by 9 to 14 seconds

## Add timings to install logs
    
The logging manager now adds timings for installation steps to the
installer logs. The information can be extracted and dumped to a CSV
file with a simple grep command:

    grep -a -Po 'TIMING: \K.*' /var/log/ipaserver-install.log > ipaserver.csv

### DIFF
--- a/install/tools/ipa-ca-install.in
+++ b/install/tools/ipa-ca-install.in
@@ -316,8 +316,9 @@ def main():
     api.Backend.ldap2.disconnect()
 
     # execute ipactl to refresh services status
+    logger.info("Starting IPA services")
     ipautil.run([paths.IPACTL, 'start', '--ignore-service-failures'],
-                raiseonerr=False)
+                raiseonerr=False, redirect_output=True)
 
 
 fail_message = '''
--- a/install/tools/ipa-dns-install.in
+++ b/install/tools/ipa-dns-install.in
@@ -143,11 +143,12 @@ def main():
     dns_installer.install(True, False, options)
     # Services are enabled in dns_installer.install()
 
-    # execute ipactl to refresh services status
-    ipautil.run([paths.IPACTL, 'start', '--ignore-service-failures'],
-                raiseonerr=False)
-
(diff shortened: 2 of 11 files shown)


tiran:
Comparison between [simple_replication PR 5116](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/75777810-fe68-11ea-aa42-fa163e43b58c/) and [simple_replication](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/97af8a58-fe4a-11ea-a718-fa163e3ad6c5/) from this PR:

ipa-server-install before: 453s
ipa-server-install faster: 391s
ipa-replica-install before: 514s
ipa-replica-install faster: 447s

rcritten:
I'd expect the update file to be slower than a raw ldif import but I guess it makes sense from a maintenance perspective.

There is a bit of mixing of default, only and add.  Things that came originally from the ldif and had no other updates can be done via default. I looked at memberHost as an example.

tiran:
I have split up the PR into multiple smaller PRs.

PR #5151, opened 2020-09-25
IPA-EPN: Make ipa-epn.timer a configuration file

fcami:
The time at which ipa-epn runs using the timer should be configurable.
Currently, ipa-epn.timer is not marked as a config file, resulting in
overwriting the file at each update.
Add %config(noreplace) so that customisation can persist.

Fixes: https://pagure.io/freeipa/issue/8517

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1446,7 +1446,7 @@ fi
 %{_mandir}/man1/ipa-epn.1*
 %{_mandir}/man5/epn.conf.5*
 %attr(644,root,root) %{_unitdir}/ipa-epn.service
-%attr(644,root,root) %{_unitdir}/ipa-epn.timer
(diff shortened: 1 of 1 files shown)


tiran:
-1

The systemd files for ipa-epn are
```
/usr/lib/systemd/system/ipa-epn.service
/usr/lib/systemd/system/ipa-epn.timer
```

These are global, package-owned files. Users must not modify these files. Instead users have to use systemd tools like ``systemctl edit``, manually create a drop-in, or copy the files to to ``/etc/systemd/system`` to override the file.

abbra:
@tiran is correct. There is no need to explicitly handle this in IPA. Normal way is exactly to create systemd-level override, whether manually or with systemd-provided tools.

See https://www.freedesktop.org/software/systemd/man/systemd.unit.html -- note that it applies to timers as well.

fcami:
OK, closing.

PR #5153, opened 2020-09-26
Pre-populate IP addresses for the name server upgrades

abbra:
Setting up resolv.conf in BIND instance expects IP addresses of the
server to be provided. This is done wiht BindInstance.setup() method
call. However, when reusing resolver setup during upgrade BIND instance
has no IP addresses configured and fails with an assert in
tasks.configure_dns_resolver().

Pass through the server's IP addresses during upgrade.

Fixes: https://pagure.io/freeipa/issue/8518


abbra:
F32 to F33 upgrade in OpenQA succeeds with this PR: https://openqa.fedoraproject.org/tests/overview?distri=fedora&version=33&build=Update-FEDORA-2020-e9e815177e where previous version (without this PR) failed.

tiran (approved):
Thanks for the fix!

tiran:
master:

* 2c393c09e0ca6b180841ab89b5259c623de91e15 Pre-populate IP addresses for the name server upgrades

PR #5156, opened 2020-09-28
Ensure that resolved.conf.d is accessible

tiran:
systemd-resolved runs as user systemd-resolve. Ensure that
resolved.conf.d drop-in directory is accessible when installer runs with
restricted umask. Also ensure the file and directory has correct SELinux
context.

The parent directory /etc/systemd exists on all platforms.

Also backup DNS config drop-ins

Fixes: https://pagure.io/freeipa/issue/8275


abbra:
LGTM.

tiran:
master:

* 34e47778b47748e94792455699b35e8c126c5db7 Ensure that resolved.conf.d is accessible
* ced1dcb1d9343a6fd0b021551eaf0b2c1d92a2c9 Also backup DNS config drop-ins

PR #5160, opened 2020-09-28
Add libpwquality checking to IPA password policy

rcritten:
This adds support for some of the libpwquality password checking features:

* palindromes (automatic)
* maximum number of repeats in a row
* maximum number of monotonic sequences (abcde, 1234, etc)
* check for username in the password
* dict check via cracklib

I attempted to retain backwards compatibility so didn't enable the character class evaluations. We could totally do this but it add six more knobs.

I didn't enable the gecos check to avoid an nss lookup which would pass through a lot of libraries only to end up back at IPA :-)

Note that pwquality has a minimum character limit of six which is different than IPA so a limit of six is enforced if any of the pwqualtiy values are set.

I suspect the SELinux policy I wrote isn't awesome.


tiran:
The SELinux policy extension does not reference any IPA parts. In the long run it should be integrated into the upstream policy for KDC and then removed from IPA.

rcritten:
I went back and forth on this myself and decided to include it in IPA since ionly IPA is using it because of the addition of libpwquality.

abbra:
Provisional ACK. Could you please remote a temp commit and see if existing passwords in tests would survive against pwquality? ;)

rcritten:
The existing policy tests are passing. The only change to them was the addition of the new objectclass.

fcami:
test_acme timed out but the runner hit ENOSPC - re-running

rcritten:
design document added

flo-renaud:
`test_external_ca_TestExternalCA` [failure](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/af50b226-0995-11eb-9a4c-fa163e041af5) is not related to this PR but rather to [8533](https://pagure.io/freeipa/issue/8533), hence adding re-run label.

abbra:
LGTM. Are the OIDs final? I don't see them reserved in RHANANA. If they are, please commit to RHANANA repository and then we can ACK this one.

I am also fine with ipa-4-8 backport.

rcritten:
The schema changes have been merged into RHANANA.

rcritten:
What do we want to do with the integration test? Integrate it into an existing set of tests? If standalone we need to figure out when to run it: part of gating or just as nightly.

IMHO nightly is probably fine.

abbra:
Yes, nightly is fine.

rcritten:
Before pushing I wanted to let you look at the testing config, top patch.

abbra:
The test definitions look good to me -- at least, I don't see any obvious issue. @netoarmando anything from your side?

netoarmando:
> The test definitions look good to me -- at least, I don't see any obvious issue. @netoarmando anything from your side?

Nope, everything looks good to me as well, all arguments present where they need to be.

rcritten:
master:

* 41021c278ae572ff5b1b3dea828a7dd93fe1ffff Add LDAP schema for new libpwquality attributes
* 6b452e54045bb957e6f787209b4498eefc5df779 Extend IPA pwquality plugin to include libpwquality support
* c03b4862b84d52ddc91c5a3fb885b0ebf753d8f2 Add new pwpolicy objectclass to test_xmprpc/objectclasses.py
* 3fc2eda4e15e9592132062036d70acad3bab401c Require libpwolicy and configure it in the build system
* c4cca53e88e78bfe512ebe59898ede0f94ec24ff Extend password policy to evaluate passwords using libpwpolicy
* 46d0096218488a961125b6d97a9210b68e5434e5 Add a unit test for libpwquality-based password policy
* 6da070e655c5d084a825607ed3be604c809b12f0 Pass the user to the password policy check in the kdb driver
* be2efc12d37018794200fee874f27d83e0442ea4 Add a raiseonerr option to ldappasswd_user_change
* fe44835970eca197543eb3c908c51a240204d846 ipatests: add test for password policies
* 68aa7c05542422aca05bec4967133be09a32496e Add SELinux policy so kadmind can read the crackdb dictionary
* f602da4b28fcf8822225b80df241eed6b624bf8e Requirements and design for libpwquality integration
* 5155280bb4a92eb3dfdee5ca3f3a332f0159d568 ipatests: Add test_pwpolicy to nightly runs

PR #5161, opened 2020-09-29
Add missing fedora_container platform members

tiran:
The fedora_container platform was missing User and Group members.

Add test case to verify that all known platforms define correct module
API.

Fixes: https://pagure.io/freeipa/issue/8519


tiran:
master:

* 1684b0f2c3de8f6dd3d15945c87bfe7e887f966c Add missing fedora_container platform members

PR #5164, opened 2020-09-29
Speed up DS related installer steps

tiran:
## Remove root-autobind configuration
    
The new lib389-based installer configured 389-DS with LDAPI support and
autobind for root.

cn=root-autobind,cn=config entry is no longer needed.

## Skip offline dse.ldif patching by default
    
The installer now stop and patches dse.ldif only when the option
--dirsrv-config-file is used. LDBM nsslapd-db-locks are increased in a
new step. This speeds up installer by 4 or more seconds on a fast system.

## Remove magic sleep from create_index_task
    
11 years ago 5ad91a0781 added a magic sleep to work around a rare deadlock
bug in memberOf plugin. Thierry is not aware of any outstanding issues
with memberOf plugin that could lead to a deadlock.

Total speedup: ~10s


ipaserver/install/dsinstance.py
```diff
@@ -566,17 +565,16 @@ def __create_instance(self):
...
             ldapuri=ipaldap.get_ldap_uri(realm=self.realm, protocol='ldapi'),
-            password=self.dm_password
         )
 
         # local_simple_allocate() configures LDAPI but doesn't set up the
         # DirSrv object to use LDAPI. Modify the DirSrv() object to use
-        # LDAPI with password bind. autobind is not available, yet.
+        # LDAPI with with autobind.
         inst.ldapi_enabled = 'on'
+        inst.ldapi_autobind = 'on'
```
rcritten, line 577: It seems a strange API but after setting the ldapi URL you can call init.setup_ldapi() to set ldapi_enabled and ldapi_autobind to "on". One is not better than the other, I just wonder if they'll refactor this at some point.

install/share/root-autobind.ldif
```diff
@@ -1,19 +0,0 @@
...
-
-dn: cn=config
-changetype: modify
-replace: nsslapd-ldapiautobind
-nsslapd-ldapiautobind: on
-
-dn: cn=config
-changetype: modify
-replace: nsslapd-ldapimaptoentries
-nsslapd-ldapimaptoentries: on
```
rcritten, line 18: Not sure if this is important but you remove this from new installs but leave it on existing ones.

tiran: I removed the lines because lib389 configures ``nsslapd-ldapimaptoentries: on`` for us. The feature is ``on`` on existing and new systems.

ipaserver/install/dsinstance.py
tiran: Ah, the function even works without setting the LDAPI url manually.

install/share/root-autobind.ldif
rcritten: I'll have to double-check. I started with an existing install and the value remained on. I then did a fresh install and it was off. I tested by doing an ldapsearch of cn=config for each. The names work out to current version, upgraded version, fresh install with upgraded version:

```
/tmp/config:nsslapd-ldapimaptoentries: on
/tmp/config2:nsslapd-ldapimaptoentries: on
/tmp/config3:nsslapd-ldapimaptoentries: off
```

tiran: Nice catch, Rob!

tiran:
master:

* 37a0af6a8cc716acd589b6295039af8ae20720fa Remove root-autobind configuration
* daec8049610ad7bbe0718276d2e7d6ffc1a7f66e Remove magic sleep from create_index_task
* 9eccaf62697f5281729c7d0f38265c66a7264e25 Skip offline dse.ldif patching by default

PR #5165, opened 2020-09-29
Reduce long sleeps in certmonger wait_for_request()

tiran:
## Add helper for poll/sleep loops with timeout
    
The Sleeper class is a helper that makes poll/sleep loops with timeout
easier to write. It takes care of edge cases and does not oversleep
timeout deadline.

## Faster certmonger wait_for_request()
    
wait_for_request() now waits 0.5 instead of 5 seconds. This shoves off
15 to 20 seconds from ipa-server-install while marginally increased
load on the system.

### DIFF
--- a/ipapython/ipautil.py
+++ b/ipapython/ipautil.py
@@ -1672,15 +1672,10 @@ class Sleeper:
 
     sleep = Sleeper(sleep=0.5, timeout=60)
     while True:
-        try:
-            do_something
-        except Exception:
-            # sleep duration can be extended
-            sleep(10)
-        else:
-            if not sleep():
-                log.info("timeout")
-                break
+        do_something
+        if not sleep():
+            log.info("timeout")
+            break
 
     longsleep = Sleeper(sleep=1, timeout=sys.maxsize)
     """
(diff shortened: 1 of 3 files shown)


ipapython/ipautil.py
```diff
@@ -1654,3 +1654,66 @@ def datetime_from_utctimestamp(t, units=1):
...
+        if now >= self.deadline:
+            if self.raises is not None:
+                raise self.raises
+            else:
+                return False
+
+        # caller can instruct sleeper to sleep longer
+        dur = self.sleep if extended_sleep is None else extended_sleep
+        # but don't sleep over deadline
+        dur = min(self.deadline - now, dur)
```
rcritten, line 1715: What's the use-case for having variable sleep periods within the same duration?

tiran: In certmonger D-Bus code there are places with longer sleeps when something went wrong. For example there is a ``sleep(10)`` when a requests times out or gets resubmitted.

The feature is not used at the moment. I can remove it. The use-case is not super important.

rcritten: If there is a future use-case I'm ok with it. Can you add a note to the block comment that it doesn't extend the duration? I paused on it longer than I'd like to admit as I absorbed the code.

Otherwise LGTM.

tiran: Let's remove the feature and re-add it when it's required.

tiran:
master:

* aa67177f4cbea0f51574e890c3a8543ff603d5d2 Add helper for poll/sleep loops with timeout
* b79191f7107fbc1890631f4c65be48b646798131 Faster certmonger wait_for_request()
* 1921d33d41f5b0e1632f7876a0acbde5462ea8be Drop unused extended sleep feature from Sleeper

PR #5166, opened 2020-09-29
Retry chronyc waitsync only once

tiran:
It's unlikely that a third chrony synchronization attempt is going to
succeed after the the first two attempts have failed. Perform more
retries with smaller timeout.

This speed up installer by 11 seconds on systems without fully
configured chronyd or no chronyd (e.g. containers).

### DIFF
--- a/ipaclient/install/timeconf.py
+++ b/ipaclient/install/timeconf.py
@@ -82,13 +82,16 @@ def sync_chrony():
     # Restart chronyd
     services.knownservices.chronyd.restart()
 
-    sync_attempt_count = 3
-    # chrony attempt count to sync with configiured servers
-    # each next attempt is tried after 10seconds of timeot
-    # 3 attempts means: if first immidiate attempt fails
-    # there is 10s delay between next attempts
-
-    args = [paths.CHRONYC, 'waitsync', str(sync_attempt_count), '-d']
+    # chrony attempt count to sync with configured servers. Each attempt is
+    # retried after $interval seconds.
+    # 4 attempts with 3s interval result in a maximum delay of 9 seconds.
+    sync_attempt_count = 4
+    sync_attempt_interval = 3
+    args = [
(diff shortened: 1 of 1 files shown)


ipaclient/install/timeconf.py
```diff
@@ -82,12 +82,10 @@ def sync_chrony():
...
-    sync_attempt_count = 3
+    sync_attempt_count = 2
     # chrony attempt count to sync with configiured servers
-    # each next attempt is tried after 10seconds of timeot
-    # 3 attempts means: if first immidiate attempt fails
-    # there is 10s delay between next attempts
-
+    # each next attempt is tried after 10 seconds of timeout
+    # 2 attempts means: if first attempt fails there is a 10s delay
     args = [paths.CHRONYC, 'waitsync', str(sync_attempt_count), '-d']
```
fcami, line 89: The change makes sense. However, how about keeping 3-4 attempts but with a much lower interval?
From what I understand this would allow choosing another host to sync from but I only have had a cursory look at chronyc sources. 
The full syntax for selecting an interval is:
chronyc waitsync [max-tries [max-correction [max-skew [interval]]]]

The lowest interval chronyc accepts is 0.1.

fcami (changes requested):

ipaclient/install/timeconf.py
tiran: Sounds like a good idea. ``chronyc waitsync 6 0 0 2`` would perform 6 reqests (5 retries) with a 2 second interval and default correction / skew values. Total runtime is little more than 10 seconds.

tiran: I have changed the values to 4 attempts with 3 second interval.

fcami (approved):

fcami:
LGTM.

tiran:
master:

* 3ab3ed5ff29bb389c28feb1d5398e76d4b04926c Retry chronyc waitsync only once

PR #5167, opened 2020-09-29
Speed up PKI installer steps

tiran:
## Dogtag: Remove set_audit_renewal step

The step set_audit_renewal modifies Dogtag's caSignedLogCert.cfg to bump
renewal to 2 years. The problem was fixed in Dogtag upstream in 2012 before
Dogtag 10.0 came out, see https://github.com/dogtagpki/pki/commit/f5b8ea5b087f642a0208c228dce6f700cd7d91c1

The update step would also no longer work. Profiles have been migrated
to LDAP several FreeIPA releases ago. pkispawn populates LDAP with all
of Dogtag's default profiles. FreeIPA does not overwrite any existing
profiles.

Win: 11 to 50 seconds

## Spawn PKI: Execute more steps early

Move several steps to an earlier phase of CA spawn. RA and ACME agent
ACLs are now configured while the server is down. This avoids yet
another restart and saves between 11 and 50 seconds per installation.

Total: ~20s to ~75s 

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -399,22 +399,35 @@ class CAInstance(DogtagInstance):
                 self.step("creating installation admin user", self.setup_admin)
             self.step("configuring certificate server instance",
                       self.__spawn_instance)
+            # Config file and ACL modifications require either restart or
+            # offline update of Dogtag.
+            self.step("stopping certificate server instance to update CS.cfg",
+                      self.stop_instance)
+            self.step("backing up CS.cfg", self.safe_backup_config)
             self.step("Add ipa-pki-wait-running", self.add_ipa_wait)
             self.step("secure AJP connector", self.secure_ajp_connector)
             self.step("reindex attributes", self.reindex_task)
             self.step("exporting Dogtag certificate store pin",
                       self.create_certstore_passwdfile)
-            self.step("stopping certificate server instance to update CS.cfg",
-                      self.stop_instance)
-            self.step("backing up CS.cfg", self.safe_backup_config)
             self.step("disabling nonces", self.__disable_nonce)
             self.step("set up CRL publishing", self.__enable_crl_publish)
             self.step("enable PKIX certificate path discovery and validation",
                       self.enable_pkix)
+            self.step("authorizing RA to modify profiles",
+                      configure_profiles_acl)
+            self.step("authorizing RA to manage lightweight CAs",
+                      configure_lightweight_ca_acls)
+            self.step("Ensure lightweight CAs container exists",
+                      ensure_lightweight_cas_container)
+            if self.clone and not promote:
+                self.step(
+                    "Ensuring backward compatibility",
+                    self.__dogtag10_migration)
             if promote:
                 self.step("destroying installation admin user",
                           self.teardown_admin)
             self.step("deploying ACME service", self.setup_acme)
+            # Materialize config changes and new ACLs
             self.step("starting certificate server instance",
                       self.start_instance)
             if promote:
@@ -433,26 +446,14 @@ class CAInstance(DogtagInstance):
                 else:
                     self.step("importing RA certificate from PKCS #12 file",
                               self.__import_ra_cert)
-
             if not ra_only:
                 if not self.clone:
                     self.step("publishing the CA certificate",
                               self.__export_ca_chain)
                     self.step("adding RA agent as a trusted user", self.__create_ca_agent)
-                self.step("authorizing RA to modify profiles", configure_profiles_acl)
-                self.step("authorizing RA to manage lightweight CAs",
-                          configure_lightweight_ca_acls)
-                self.step("Ensure lightweight CAs container exists",
-                          ensure_lightweight_cas_container)
-                if self.clone and not promote:
-                    self.step(
-                        "Ensuring backward compatibility",
-                        self.__dogtag10_migration)
                 self.step("configure certificate renewals", self.configure_renewal)
                 self.step("Configure HTTP to proxy connections",
                           self.http_proxy)
-                # This restart is needed for ACL reload in CA, do not remove it
-                self.step("restarting certificate server", self.restart_instance)
                 self.step("updating IPA configuration", update_ipa_conf)
                 self.step("enabling CA instance", self.__enable_instance)
                 if not promote:
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -435,8 +435,6 @@ class CAInstance(DogtagInstance):
                               self.__import_ra_cert)
 
             if not ra_only:
-                self.step("setting audit signing renewal to 2 years", self.set_audit_renewal)
-                self.step("restarting certificate server", self.restart_instance)
                 if not self.clone:
                     self.step("publishing the CA certificate",
                               self.__export_ca_chain)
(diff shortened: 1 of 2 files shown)


ipaserver/install/cainstance.py
```diff
@@ -463,8 +463,6 @@ def configure_instance(self, host_name, dm_password, admin_password,
                         self.step("exposing CA instance on LDAP",
                                   self.__expose_ca_in_ldap)
 
-                    self.step("migrating certificate profiles to LDAP",
-                              migrate_profiles_to_ldap)
                     self.step("importing IPA certificate profiles",
```
rcritten, line 462: Is migrate_profiles_ldap() needed at all then? It's executed as part of the upgrade. Speeding that up would be nice too.

On my hardware I got a 40-second reduction in install time.

ipaserver/install/cainstance.py
```diff
@@ -433,26 +443,14 @@ def configure_instance(self, host_name, dm_password, admin_password,
...
-                self.step("Ensure lightweight CAs container exists",
-                          ensure_lightweight_cas_container)
-                if self.clone and not promote:
-                    self.step(
-                        "Ensuring backward compatibility",
-                        self.__dogtag10_migration)
                 self.step("configure certificate renewals", self.configure_renewal)
                 self.step("Configure HTTP to proxy connections",
                           self.http_proxy)
-                # This restart is needed for ACL reload in CA, do not remove it
```
rcritten, line 456: This restart is to avoid a CVE. It looks like a restart happens later but I didn't spend timing finding out exactly where. I don't know if this was simply to ensure a restart or if it had to be done at this point.

frasertweedale: Not quite right - this comment was added during the secure-ajp-connection commit, but only to explain why this restart was needed.  The restart already occurred here for reason stated in the comment.  The restart was not introduced during and is not needed for the secure-ajp-connection change because a restart already happens after `self.secure_ajp_connection` gets executed.

So yes the restart was needed, no it was not for the CVE. But with this change, the ACL tweaks are now being performed before the server is restarted.  So it is OK to remove this restart.

@tiran please add a comment near steps that tweak the Dogtag ACLs, to explain that the server must be (re)started after those steps.

ipaserver/install/cainstance.py
```diff
@@ -399,18 +399,28 @@ def configure_instance(self, host_name, dm_password, admin_password,
...
-                      self.stop_instance)
-            self.step("backing up CS.cfg", self.safe_backup_config)
             self.step("disabling nonces", self.__disable_nonce)
             self.step("set up CRL publishing", self.__enable_crl_publish)
             self.step("enable PKIX certificate path discovery and validation",
                       self.enable_pkix)
+            self.step("authorizing RA to modify profiles",
+                      configure_profiles_acl)
+            self.step("authorizing RA to manage lightweight CAs",
+                      configure_lightweight_ca_acls)
```
frasertweedale, line 417: Here we should keep a comment to explain that restart must occur after these steps.

frasertweedale:
You have my ACK.  (Can I still ack a freeipa PR?)

ipaserver/install/cainstance.py
abbra: I remember we had an issue that the full upgrade code didn't run at install time, only LDAP upgrades.

tiran: The server upgrader is on my TODO list. For now I'm focusing on the installer. There are several upgrade steps that may no longer be necessary or that should be skipped if the step has been executed by another server in the cluster already.

I think the upgrade step is required when updating a server from an old Dogtag version or upgrading a cluster.

frasertweedale: @abbra w.r.t. profile migration, Dogtag itself imports LDAP profiles upon start up (when using the LDAPProfileSubsystem, which in FreeIPA it is).

Wondering about why this migration code is called in the installer, perhaps it was required at some point (during development) after the `LDAPProfileSubsystem` was implemented, but before the PKI installer learned the import behaviour.  I cannot remember the exact reasons, but this is my working theory.

abbra: It may well be the reason, indeed.  The code comes from you ;)
```
commit b24fe0eb733c68af4042cdd78fca6f609efe843b
Author: Fraser Tweedale <ftweedal@redhat.com>
Date:   Mon May 11 21:17:48 2015 -0400
```

frasertweedale: The commit where it was added is 620036d26e98fdcefff00168e9e5463a8257d49c.

Profiles are not being imported during *replica installation*, apart from this step.  So I think we still need it but only for replica installation, and we can skip it if we see that profiles are already in LDAP.

frasertweedale:
ACK revoked.  We need to discuss that LDAP profile import code a bit more: https://github.com/freeipa/freeipa/pull/5167#discussion_r497328345

ipaserver/install/cainstance.py
tiran: > @abbra w.r.t. profile migration, Dogtag itself imports LDAP profiles upon start up (when using the LDAPProfileSubsystem, which in FreeIPA it is).

Could you please clarify. Does Dogtag import missing profiles on every startup of the pki-tomcat service or does Dogtag only import profiles during pkispawn?

frasertweedale: @tiran only during pkispawn, and *not* for cloning (replica install).

frasertweedale: @tiran stand by, I want to confirm that.

tiran: The code block with ``migrate_profiles_to_ldap`` is guarded by ``if not promote``. Replica installer calls the function with ``promote=True``.

frasertweedale: @tiran confirmed.  From `base/server/python/pki/server/deployment/scriptlets/configuration.py`~912:
```python
        if not clone and subsystem.type == 'CA':
            subsystem.import_profiles( 
                input_folder='/usr/share/pki/ca/profiles/ca') 
```

(Sorry that took a while to work out.  The import mechanism has completely changed since I wrote it 6 years ago).

ipaserver/install/cainstance.py
```diff
@@ -399,22 +399,35 @@ def configure_instance(self, host_name, dm_password, admin_password,
                 self.step("creating installation admin user", self.setup_admin)
             self.step("configuring certificate server instance",
                       self.__spawn_instance)
+            # Config file and ACL modifications require either restart or
+            # offline update of Dogtag.
```
frasertweedale, line 403: :+1:

ipaserver/install/cainstance.py
frasertweedale: @tiran wrote:
> The code block with migrate_profiles_to_ldap is guarded by if not promote. Replica installer calls the function with promote=True.

Let's think this through.

We no longer support DL0 replica installation right?  That means that the peer server is at a version that has LDAP profiles.  Therefore replica installation never needs to import profiles.  Dogtag CA cloning procedure does not import profiles to LDAP, but they are already in LDAP so that is not a problem.

Does the logic work out?  If it does, then we can indeed delete the profile import step for all installation scenarios.

tiran: I confirmed removal of DL0 support with Martin. IPA 4.8 / RHEL 8 no longer support DL0 IdM clusters. DL0 was deprecated in [RHEL 7.9](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/8.0_release_notes/index). We can safely assume that a 4.9 replica requires a DL1 cluster that has been migrated to have LDAP profiles.

frasertweedale: OK :+1:

*hands back the ack*.

tiran:
I have changed to PR to handle LDAP cert profile migration differently. It's still not entirely clear if we can drop this migration. It may be required under some circumstances when updating an old cluster to DL1. The migration code now assumes that profiles are migrated when the RA agent cert profile is in LDAP.

tiran:
I have dropped the commit that removes ``migrate_profiles_to_ldap()``. There is an ongoing discussion how to handle LDAP profile updates and new profiles.

tiran:
master:

* 8882680ee167c535e7539fe0452892aae617aac3 Dogtag: Remove set_audit_renewal step
* 942fe07eb53dbbacb33352bbb67039abf73b2544 Spawn PKI: Execute more steps early

PR #5170, opened 2020-10-02
Centralize enable/disable of the ACME service

rcritten:
Centralize enable/disable of the ACME service

The initial implementation of ACME in dogtag and IPA required
that ACME be manually enabled on each CA.

dogtag added a REST API that can be access directly or through
the `pki acme` CLI tool to enable or disable the service.

It also abstracted the database connection and introduced the
concept of a realm which defines the DIT for ACME users and
groups, the URL and the identity. This is configured in realm.conf.

A new group was created, Enterprise ACME Administrators, that
controls the users allowed to modify ACME configuration.

The IPA RA is added to this group for the ipa-acme-manage tool
to authenticate to the API to enable/disable ACME.

Two ACME configuration templates were removed so that the dogtag
defaults would be used, configsources.conf and engine.conf.

Related dogtag installation documentation:
https://github.com/dogtagpki/pki/blob/master/docs/installation/acme/Configuring_ACME_Database.md
https://github.com/dogtagpki/pki/blob/master/docs/installation/acme/Configuring_ACME_Realm.md
https://github.com/dogtagpki/pki/blob/master/docs/installation/acme/Installing_PKI_ACME_Responder.md

ACME REST API:
https://github.com/dogtagpki/pki/wiki/PKI-ACME-Enable-REST-API

https://pagure.io/freeipa/issue/8524


ipaserver/install/ipa_acme_manage.py
```diff
@@ -52,28 +98,17 @@ def run(self):
...
-        # Work around a limitation in PKI ACME service file watching
-        # where renames (what DirectiveSetter does) are not detected.
-        # It will be fixed, but keeping the workaround will do no harm.
-        pathlib.Path(paths.PKI_ACME_ENGINE_CONF).touch()
-
-        # Nothing else to do; the Dogtag ACME service monitors engine.conf
-        # for updates and reconfigures itself as required.
+        api.bootstrap(in_server=True, confdir=paths.ETC_IPA)
+        api.finalize()
+        api.Backend.ldap2.connect()
```
tiran, line 103: Why do Dogtag plugins create a new LDAP connection instead of using the existing LDAP connection from the default API? I have seen this before in other plugins.

ipaserver/install/service.py
```diff
@@ -347,7 +347,10 @@ def _ldap_mod(self, ldif, sub_dict=None, raise_on_err=True,
                   ldap_uri=None, dm_password=None):
         pw_name = None
         fd = None
-        path = os.path.join(paths.USR_SHARE_IPA_DIR, ldif)
+        if not os.path.isabs(ldif):
+            path = os.path.join(paths.USR_SHARE_IPA_DIR, ldif)
+        else:
+            path = ldif
```
tiran, line 353: This check is not needed:

> https://docs.python.org/3/library/os.path.html#os.path.join
> If a component is an absolute path, all previous components are thrown away and joining continues from the absolute path component.

ipaserver/plugins/dogtag.py
```diff
@@ -256,7 +256,7 @@
 from ipapython import dogtag, ipautil, certdb
 from ipaserver.masters import find_providing_server
 
-if api.env.in_server:
+if api.isdone('finalize') and api.env.in_server:
```
tiran, line 259: Is this import check still necessary? ``freeipa-server`` depends on ``python3-pki``, so the import should never fail.

ipaserver/install/service.py
rcritten: The check is there because the existing code always adds /usr/share/ipa to the path.

ipaserver/install/ipa_acme_manage.py
rcritten: I don't understand the question. This isn't a plugin, it's using code that I think expects to be run inside the server and assumes that the API is initialized and an LDAP connection exists.

ipaserver/plugins/dogtag.py
rcritten: I suspect this is legacy code from when the plugins were included in the client. I'll remove it.

rcritten:
FWIW the Azure failure is expected since it doesn't have the necessary pki builds.

rcritten:
With pki-acme-10.10.0-0.1.alpha1.20201009195936UTC.9332e4f6 the global enable/disable works.

This is ready for review.

abbra:
LTGM. I have, however, a concern about use of port 8443 reference in installers' code:

```
$ git grep 8443
client/man/default.conf.5:Specifies the secure CA agent port. The default is 8443.
install/share/pki-acme-issuer.conf.template:url=https://$FQDN:8443
ipaplatform/base/services.py:    'pki-tomcatd@pki-tomcat.service': [8080, 8443],
ipaplatform/base/services.py:    'pki-tomcat': [8080, 8443],
ipaplatform/base/services.py:    'pki-tomcatd': [8080, 8443],  # used if the incoming instance name is blank
ipaserver/install/ca.py:                "IPA requires ports 8080 and 8443 for PKI, but one or more "
ipaserver/install/cainstance.py:    """Check that dogtag ports (8080, 8443) are available.
ipaserver/install/cainstance.py:    return all([ipautil.check_port_bindable(8443),
ipaserver/install/cainstance.py:                      'https://%s:8443/ca/agent/ca/' % self.fqdn,
ipaserver/install/cainstance.py:            api.Backend.ra.override_port = 8443
ipaserver/install/cainstance.py:    api.Backend.ra_certprofile.override_port = 8443
ipaserver/install/cainstance.py:    api.Backend.ra_certprofile.override_port = 8443
ipaserver/install/cainstance.py:    api.Backend.ra_certprofile.override_port = 8443
ipaserver/install/cainstance.py:    api.Backend.ra_lightweight_ca.override_port = 8443
ipaserver/install/certs.py:            self.host_name, 8443,
ipaserver/install/dogtaginstance.py:        port='8443',
ipaserver/install/dsinstance.py:                    '--ee-url', 'https://%s:8443/ca/ee/ca' % self.fqdn,
ipaserver/install/httpinstance.py:                    '--ee-url', 'https://%s:8443/ca/ee/ca' % self.fqdn,
ipaserver/install/krbinstance.py:                    '--ee-url', 'https://%s:8443/ca/ee/ca' % self.fqdn,
```

This is unrelated to ACME support but it looks fragile this way because we allow to configure `ca_agent_port` in `/etc/ipa/default.conf` (or in its per-user variant) and at the same time time ignore it almost everywhere in the installers:

```
$ git grep ca_agent_port
client/man/default.conf.5:.B ca_agent_port <port>
install/ui/test/data/ipa_init.json:               "ca_agent_port" : 443,
ipalib/constants.py:    ('ca_agent_port', 443),
ipaserver/plugins/dogtag.py:            self.ca_host, self.override_port or self.env.ca_agent_port,
ipaserver/plugins/dogtag.py:            self.ca_host, self.override_port or self.env.ca_agent_port,
ipaserver/plugins/dogtag.py:            self.ca_host, self.override_port or self.env.ca_agent_port,
ipaserver/plugins/dogtag.py:                         self.env.ca_agent_port,
ipaserver/plugins/dogtag.py:                         self.env.ca_agent_port,
ipaserver/plugins/dogtag.py:                         self.override_port or self.env.ca_agent_port,
pylint_plugins.py:    api.env.ca_agent_port = 0
```

I think this needs to be improved in a separate PR.

abbra:
One more note: Azure Pipelines failed as there is no `pki-acme` subpackage installed there.

mrizwan93:
Test is failing because `ipa-acme-manage` doesn't have `status` option. It will be available with https://github.com/freeipa/freeipa/pull/5119. Can we move PR#5119 first and then this PR?

rcritten:
If you believe that #5119 is acceptable then feel free to replace the WIP tag with ack.

rcritten:
> LTGM. I have, however, a concern about use of port 8443 reference in installers' code:
> 

This is done in cases where communication is done directly to the CA and not proxied.  I didn't look at all cases but at least some of them are due to the CA being configured before the proxy.

rcritten:
If it is determined that ACME will ship in the ipa-4-8 branch then I'll need time to backport and test the patches.

If necessary I can add a switch which disables ACME with pki < 10.10 so we can merge this code and remove it once the pki builds are available.

The testing of globle enable/disable isn't going to star working as the mechanism for enable/disable changed completely. So even if we merge #5119 the test is not going to pass without a fair bit of rework.

rcritten:
Added two patches:

1) add status option to ipa-acme-status to try to get CI passing again
2) skip ACME if global enable part is missing. This may be a temporary patch, added to get green on Azure.

ipaserver/install/cainstance.py
```diff
@@ -769,6 +770,11 @@ def __create_ca_agent(self):
             self.basedn)
         conn.add_entry_to_group(user_dn, group_dn, 'uniqueMember')
 
+        if minimum_acme_support():
+            group_dn = DN(('cn', ACME_AGENT_GROUP), ('ou', 'groups'),
+                          self.basedn)
+            conn.add_entry_to_group(user_dn, group_dn, 'uniqueMember')
```
abbra, line 776: `ldap2.add_entry_to_group()` can throw `errors.AlreadyGroupMember` error, so it makes sense to intercept it here for a replica case.

abbra (changes requested):
I added few comments. One remaining is whether we should have an explicit dependency to a newer dogtag version in the spec file. As discussed with Dogtag team, they are going to release 10.10 so that we can move forward with ACME support.

install/tools/man/ipa-acme-manage.1
```diff
@@ -24,6 +24,9 @@ Enable the ACME service on this host.
 .TP
 \fBdisable\fR
 Disable the ACME service on this host.
+.TP
+\fBdisable\fR
```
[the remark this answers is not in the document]
rcritten: sigh, fixed.

ipaserver/install/cainstance.py
rcritten: This is called in __create_ca_agent() which only executes if not a clone.

ipatests/test_integration/test_acme.py
```diff
@@ -266,3 +266,78 @@ def test_centralize_acme_disable(self):
...
+
+        # check status of acme on replica, result: CA is not installed
+        result = self.replicas[0].run_command(['ipa-acme-manage', 'status'],
+                                              raiseonerr=False)
+        assert result.returncode == 1
+
+        # Make CA on replica
+        tasks.install_ca(self.replicas[0])
+
+        time.sleep(30)
```
rcritten, line 339: What is the hardcoded sleep for?

ipatests/test_integration/test_acme.py
```diff
@@ -266,3 +266,78 @@ def test_centralize_acme_disable(self):
...
+        for _i in range(5):
+            time.sleep(2)
+            try:
+                self.clients[0].run_command(
+                    ['curl', '--fail', self.acme_server])
+                break
+            except Exception as e:
+                exc = e
+        else:
+            raise exc
```
rcritten, line 325: I don't understand why this block comes before checking the status, which does the same thing.

From a client perspective you don't know which master you're getting. Since ipa-ca should be a CNAME it's likely that you're pinging the master and replica in round-robin.

abbra:
https://bodhi.fedoraproject.org/updates/FEDORA-2020-9765318e8b was pushed to stable and should now unblock Azure Pipelines CI.

rcritten:
CA installation failed, re-running.

rcritten:
pki-spawn is failing on the replica installation. Pinged @edewata and @cipherboy

edewata:
Just FYI, the fix for PKI cloning has been merged:
https://github.com/dogtagpki/pki/commit/61744481ed346d16649359dbbdd13c52b1d893a5
and the build is available in @pki/master repo:
https://copr.fedorainfracloud.org/coprs/g/pki/master/builds/
so feel free to remove my repo. Thanks.

rcritten:
The test_certbot_dns test is failing when a replica with a CA is installed.

This is what we get back from the ACME server:

{"status":"invalid","expires":"2020-10-23T22:56:31Z","identifier":{"type":"dns","value":"client0.ipa.test"},"challenges":[{"type":"dns-01","url":"https://ipa-ca.ipa.test/acme/chall/ZeSTt5V9Nl","token":"E2IeF4ioSxfeABMOVLzcAg","status":"invalid"}]}

rcritten:
The dns challenge works ok for me locally. I started with one master and one client. I registered the client and obtained a cert over dns with:

certbot --server https://ipa-ca.example.test/acme/directory register -m nobody@example.test --agree-tos --no-eff-email

certbot --server https://ipa-ca.example.test/acme/directory certonly --non-interactive --domain `hostname` --preferred-challenges dns --manual --manual-public-ip-logging-ok --manual-auth-hook /usr/libexec/ipa/acme/certbot-dns-ipa --manual-cleanup-hook /usr/libexec/ipa/acme/certbot-dns-ipa

Then I cleaned up with rm -rf /etc/letsencrypt/* and installed another server with a CA and tried again.

Succeeded again. So the failure only happens in CI for some reason.

rcritten:
The replica resolv.conf needed to be pointed to the IPA master. I merged those two calls into the newly added tests and that should make it pass.

ipatests/test_integration/test_acme.py
```diff
@@ -114,6 +115,11 @@ def test_enable_acme_service(self):
         else:
             raise exc
 
+    def test_centralize_acme_enable(self):
+        """Test if ACME enable on replica if enabled on master"""
+        result = self.master.run_command(['ipa-acme-manage', 'status'])
+        assert 'enabled' in result.stdout_text
```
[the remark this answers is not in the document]
frasertweedale: Seems like it is checking the replica now.

I think we may want some retry behavoiur here in case there is a lag in the replica receiving and responding to the new configuration.  (Same for the disable check).  What do you think @mrizwan93 @rcritten ?

ipatests/test_integration/test_acme.py
```diff
@@ -252,3 +264,72 @@ def test_disable_acme_service(self):
...
+        tasks.config_host_resolvconf_with_master_data(
+            cls.master, cls.replicas[0]
+        )
+
+    def test_cafull_centralize_acme_enable(self):
+        """Test if ACME is enabled after CA-less to CA-full deployment
+
+        Deployment where one server is deployed as CA-less, when converted
+        to CA full, should have ACME enabled by default.
+
```
frasertweedale, line 313: I think this comment is wrong.  ACME service should be available after CA installation, but not be immediately enabled.  The test procedure executes `ipa-acme-manage enable` which supports this.

ipatests/test_integration/test_acme.py
```diff
@@ -252,3 +264,72 @@ def test_disable_acme_service(self):
...
+    def install(cls, mh):
+        # cache the acme service uri
+        acme_host = f'{IPA_CA_RECORD}.{cls.master.domain.name}'
+        cls.acme_server = f'https://{acme_host}/acme/directory'
+
+        # install packages before client install in case of IPA DNS problems
+        if not skip_certbot_tests:
+            cls.clients[0].run_command(['dnf', 'install', '-y', 'certbot'])
+        if not skip_mod_md_tests:
+            cls.clients[0].run_command(['dnf', 'install', '-y', 'mod_md'])
```
frasertweedale, line 295: AFAICT this test does not perform certbot or mod_md tests.  We can delete these installation steps and also set
`num_clients = 0` and `num_replicas = 0`.

frasertweedale (changes requested):
No red flags but a few comments.  Haven't tested - but that's what the integration tests are for :)

ipatests/test_integration/test_acme.py
mrizwan93: My understanding is that in topology if we enable ACME on any server,  it should be enabled on all the servers.

So, IMO in topology where one server is CA-less, when converted to CA-full, ACME should be by default enabled.

ipatests/test_integration/test_acme.py
mrizwan93: removed. But `num_replicas` remains 1 as it is needed for test.

ipatests/test_integration/test_acme.py
frasertweedale: @mrizwan93 there are a few different scenarios here:

1. No CA in deployment at all.  `ipa-ca-install` will install the CA, promoting deployment from CA-less to CA-ful.  ACME service is initially disabled and can be enabled by `ipa-acme-manage`.

2. CA-ful deployment with CA-less replica, ACME enabled.  In this case, `ipa-ca-install` will install the CA role on the previously CA-less replica, and ACME service should be active on that replica automatically.

3. CA-ful deployment with CA-less replica, ACME not enabled.  In this case `ipa-ca-install` will install the CA role on the previously CA-less replica, and ACME service remains disabled on all CA replicas, until an adminstrator enables it.

Do you agree?

mrizwan93: - [ ] 1, what do you mean by 'NO CA in deployment'? AFAIK, We cannot install ipa master without CA. `ipa-server-install` itself installs CA.

- [ ] 2 is what I tried to implement here

- [ ] 3 agree

frasertweedale: @mrizwan93 ok, thanks for clarifying.

1. You can indeed install a FreeIPA deployment (first replica via `ipa-server-install`) without a CA.  This is a "CA-less deployment".  This is why I was confused about your comment, where you refer to "CA-less deployment".  When there is a CA-ful deloyment (that is, a deployment which has at least one CA replica), we would refer to "CA-less *replica*", but the deployment is "CA-ful".  I do not think we need a specific test for what happens with ACME when installing first CA server into a CA-less deployment.  But it would not hurt to test it.

2. OK, that sounds good.  So I think the comment / test description needs clarification, to say that (a) deployment is CA-ful, (b) ACME is enabled, (c) promoting a CA-less replica to CA-ful results in ACME immediately enabled on that replica.

3. :+1:

ipatests/test_integration/test_acme.py
frasertweedale: @mrizwan93 yes thank you, you are correct!  Cheers.

ipaserver/install/cainstance.py
```diff
@@ -1718,6 +1727,19 @@ def ensure_lightweight_cas_container():
...
+    """
+    ACME with global enable/disable is required.
+
+    This first shipped in dogtag version 10.10.0.
+
+    We could test for minimum version by calling pki --version
+    but the file realm.conf was introduced in this and is less
+    costly to test for.
+    """
+    return os.path.exists('/usr/share/pki/acme/conf/realm.conf')
```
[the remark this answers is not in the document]
rcritten: It's not going to be merged. I added this so Azure would pass so the PR would be green.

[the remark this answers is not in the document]
frasertweedale: OK, I missed the memo :)  That's fine then.

rcritten:
Fixed the linter issues and replaced logger.info with print.

ipaserver/install/cainstance.py
```diff
@@ -1727,6 +1728,33 @@ def ensure_lightweight_cas_container():
...
+    groups = re.match('.*\nSpecification-Version: ([\d+\.]*)\n.*', s)
+    if groups:
+        version_string = groups.groups(0)[0]
+        minimum_version = parse_version('10.10.0')
+
+        return parse_version(version_string) >= minimum_version
+    else:
+        logger.debug('Unable to parse version from {}'.format(data))
+        return False
+
```
rcritten, line 1756: @frasertweedale change in plans. I think we may ship with this minimum version test so we can merge this without breaking until pki 10.10.0 lands. I changed the file test into a version parser which should be less fragile.

rcritten:
I think we're good to go with this PR now. I've reviewed @mrizwan93 changes and they look good to me.

Can I get a final review and ack on this by end of day Friday the 30th?

This is safe to push to master because of the minimum pki version check. Once 10.10 is widely available ACME will magically start working (for new installs). Currently ACME is skipped altogether if < 10.10. Arguably 10.9.4 has a working ACME system it just lacks global enablement. I can loosen the requirements if desired but only 10.10.0+ is my preference.

We will need a future PR to remove the minimum check and add a Requires on 10.10.0.

frasertweedale (approved):

frasertweedale:
The temp commit is happy and I am happy.  Can we remove the temp commit, make sure gating CI goes green?  Then we can merge!

rcritten:
Adding ack per Fraser's last comment. I also reviewed and approve Rizwan's tests.

rcritten:
master:

* 2ef53196c6a012ad7d02aed5672d69fbcb5d0a4e Enable importing LDIF files not shipped by IPA
* e13d058a066bdbd8a81b794b6f18ca8eca1f31c8 Let dogtag.py be imported if the api is not initialized
* c0d55ce6de5e41a98b1e37e23e8bdb339e772c0f Centralize enable/disable of the ACME service
* 92c3ea4e293a4fca58269265b7fbf511024dab59 Don't install ACME if full support is not available
* 69ae48c8b614a23f530ec6ed33c9019e0b491e50 Add a status option to ipa-acme-manage
* e7fd791579eca8b7a1c30b4f17bfac7c4fafe2a7 ipatests: Check if ACME is enabled on all CA servers
* d4ef64b229541200564131e927cd0b4b32662fe2 ipatests: Collect the let's encrypt log

PR #5173, opened 2020-10-06
Use separate install logs for AD and DNS instance

tiran:
ipa-dns-install and ipa-adtrust-install no longer overwrite
ipaserver-install.log. Instead they use a separate log file.

Add AD-Trust, DNS, KRA, and replica log files to backups.

Fixes: https://pagure.io/freeipa/issue/8528


fcami:
LGTM

abbra:
LGTM. I think we also need to update sosreport to include new files into the reports (optionally).

tiran:
master:

* 6860c6376087cf4a3b099444232baf2a6a4d3945 Use separate install logs for AD and DNS instance

PR #5174, opened 2020-10-06
Lookup ipa-ca record with NSS

tiran:
DNS data management now uses NSS's getaddrinfo() instead of direct DNS
queries to resolve the ipa-ca record. This fixes missing ipa-ca records
when the current hostname is not resolvable in DNS but has correct
records in /etc/hosts.

Reduce timeout to 15 seconds and tighten timeout loop.

The changeset can speed up installation by almost 60 seconds.
ipa-server-install without built-in DNS calls into DNS data management
twice with a timeout of 30 seconds for each call.

Fixes: https://pagure.io/freeipa/issue/8529
Related: https://pagure.io/freeipa/issue/8501

### DIFF
--- a/ipaserver/dns_data_management.py
+++ b/ipaserver/dns_data_management.py
@@ -15,13 +15,13 @@ from dns import (
     rdatatype,
     zone,
 )
-from dns.exception import DNSException
 
 from time import sleep, time
 
 from ipalib import errors
 from ipalib.dns import record_name_format
-from ipapython.dnsutil import DNSName, resolve_rrsets
+from ipapython.dnsutil import DNSName
+from ipaserver.install import installutils
 
 if six.PY3:
     unicode=str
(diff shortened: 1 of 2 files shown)


abbra:
```
[ipatests.pytest_ipa.integration.host.Host.master.cmd29] 'UnsafeIPAddress' object is not iterable

```

ipaserver/dns_data_management.py
```diff
@@ -139,16 +139,13 @@ def __add_srv_records(
...
-            try:
-                rrsets = resolve_rrsets(hostname, (rdatatype.A, rdatatype.AAAA))
-            except DNSException:  # logging is done inside resolve_rrsets
-                pass
+            # logging is done inside function
+            rrsets = installutils.resolve_rrsets_nss(hostname)
             if rrsets:
                 break
-            sleep(5)
+            sleep(4)
```
rcritten, line 148: Shouldn't the sleep value be a factor of CA_RECORDS_DNS_TIMEOUT?

tiran: I change sleep to 3sec with 15sec timeout.

abbra:
LGTM

tiran:
master:

* 731c5b211041ee18940082b29e6a21b3f084fc57 Lookup ipa-ca record with NSS

PR #5176, opened 2020-10-06
freeipa.spec.in: client: depend on libsss_sudo

fcami:
On 10.10+ releases of Dogtag, the PKI installer will not depend on sudo anymore. This opens the possibility of creating IPA servers without a properly configured sudo.
In fact, even IPA clients should have sudo and libsss_sudo installed in most cases, so: add a weak dependency on libsss_sudo to freeipa-client.

Fixes: https://pagure.io/freeipa/issue/8530

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -24,6 +24,7 @@ import re
 import SSSDConfig
 import shutil
 import socket
+import subprocess
 import sys
 import tempfile
 import textwrap
@@ -2198,7 +2199,18 @@ def install_check(options):
             "authentication resources",
             rval=CLIENT_INSTALL_ERROR)
 
-    # when installing with '--no-sssd' option, check whether nss-ldap is
+    # When installing without the "--no-sudo" option, check whether sudo is
+    # available.
+    if options.conf_sudo:
+        try:
+            subprocess.Popen(['sudo -V'])
+        except FileNotFoundError:
+            logger.info(
+                "The sudo binary does not seem to be present on this "
+                "system. Please consider installing sudo if required."
+            )
+
+    # when installing with the '--no-sssd' option, check whether nss-ldap is
     # installed
     if not options.sssd:
         if not os.path.exists(paths.PAM_KRB5_SO):
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -29,6 +29,7 @@ import re
 import collections
 import itertools
 import shutil
+import shlex
 import copy
 import subprocess
 import tempfile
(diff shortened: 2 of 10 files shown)


fcami:
@tiran, could you please have a look? Adding a hard dependency on libsss_sudo did not seem wise for the FreeIPA container use-case, hence my using Recommends: here.

fcami:
Please hold any ACK.

fcami:
Re-tagging WIP. This requires client installer changes and an associated test.

abbra:
http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/231631a8-0a37-11eb-a24f-fa163ee3281e/ demostrated some weird dogtag exceptions during installation. I re-started the test for `test_installation_TestInstallMaster`.

fcami:
The [http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/231631a8-0a37-11eb-a24f-fa163ee3281e/test_integration-test_installation.py-TestInstallMaster-test_install_kra/master.ipa.test/var/log/pki/pki-tomcat/pki/debug.2020-10-09.log.gz](log) has:
```
Caused by: java.security.AccessControlException: access denied ("java.io.FilePermission" "/usr/share/pki/server/webapps/pki/WEB-INF/lib/slf4j-jdk14.jar" "read")
```
This is an unrelated, known intermittent Dogtag issue.

fcami:
hi @netoarmando could you please have a look at the template changes?

netoarmando:
> hi @netoarmando could you please have a look at the template changes?

They look fine!

ipaclient/install/client.py
```diff
@@ -2165,7 +2166,18 @@ def install_check(options):
             "authentication resources",
             rval=CLIENT_INSTALL_ERROR)
 
-    # when installing with '--no-sssd' option, check whether nss-ldap is
+    # When installing without the "--no-sudo" option, check whether sudo is
+    # available.
+    if options.conf_sudo:
+        try:
+            subprocess.Popen(['sudo'])
```
abbra, line 2173: can you please run at least 'sudo -V' here? This would not start sudo in authentication mode but would at least allow to verify the tool exists.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -2585,3 +2585,23 @@ def run_ssh_cmd(
...
+
+def remove_package(host, pkg, nodeps=False, raiseonerr=True):
+    cmd = ['rpm', '-e', pkg]
+    if nodeps:
+        cmd.append('--nodeps')
+    result = host.run_command(cmd, raiseonerr=raiseonerr)
+    return result
+
+
+def install_package(host, pkg):
```
abbra, line 2605: We already have `ipatests/pytest_ipa/integration/tasks.py::install_packages()` and `uninstall_packages()`, why do we need another pair? This one also doesn't support Debian platform, unlike the previous one.

abbra:
Review comments need to be addressed first.

ipaclient/install/client.py
fcami: Thanks for the hint. Fixed.

ipatests/pytest_ipa/integration/tasks.py
fcami: I've adapted the existing code so that it handles --nodeps situations including on Debian. thanks!

ipatests/test_integration/test_installation.py
```diff
@@ -1536,3 +1536,69 @@ def test_replica_install_against_server(self):
...
+            self.master, ['sudo', 'libsss_sudo'], nodeps=True
+        )
+        self.master.run_command(
+            ['ipactl', 'restart']
+        )
+
+    def test_install_sudo_on_client(self):
+        """ Check that installing sudo pulls libsss_sudo in"""
+        for pkg in ('sudo', 'libsss_sudo'):
+            assert tasks.is_package_installed(self.clients[0], pkg) is False
```
miskopo, line 1600: `assert not tasks.is_package_installed(self.clients[0], pkg)` would be preferred

miskopo:
:duck: as in "quack", which roughly translates to "QA ACK"

netoarmando:
It's ACK'd but the temp commit must be removed before pushing it.

fcami:
@netoarmando and @miskopo thanks! I've removed the temp commit.

flo-renaud:
removing the ipa-4-8 label as this branch is not maintained any more

flo-renaud:
master:

* fcfefe0a2686363d49da8ca5ce8ab137465bd7fd freeipa.spec: client: depend on libsss_sudo and sudo
* a912bbbe0e91aede0c0abb8acef31ec516826c04 ipa-client-install: output a warning if sudo is not present
* ed8a9e4e6a79285f9e7c50e86c3835ffe2f0fd9f ipatests: tasks: handle uninstalling packages with nodeps
* 82216bfa3df82d425ecb44faf8b177b21c304624 ipatests: add TestInstallWithoutSudo

flo-renaud:
@fcami A manual backport is required for ipa-4-9 branch

PR #5180, opened 2020-10-07
Temp commit to try to reproduce uninstall issue

rcritten:
Stanislav found that there are some test failures with the
updated certmonger package both in ALTLinux and F32.

This is an attempt to reproduce using one of the tests that is
failing for him.

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
--- a/client/sysconfig/certmonger
+++ b/client/sysconfig/certmonger
@@ -1,1 +1,1 @@
-OPTS=-d2
+OPTS=-d9
--- a/ipatests/prci_definitions/temp_commit.yaml
+++ b/ipatests/prci_definitions/temp_commit.yaml
@@ -82,10 +82,10 @@ jobs:
       args:
         build_url: '{fedora-latest/build_url}'
         update_packages: True
-        test_suite: test_integration/test_external_ca.py::TestExternalCA test_integration/test_external_ca.py::TestExternalCAConstraints test_integration/test_installation.py::TestInstallCA::test_certmonger_reads_token_HSM
+        test_suite: test_integration/test_external_ca.py::TestExternalCA test_integration/test_external_ca.py::TestExternalCAConstraints
         template: *testing-master-latest
         timeout: 4800
-        topology: *master_2repl_1client
+        topology: *master_1repl_1client
 
   fedora-latest/temp_commit_3:
     requires: [fedora-latest/build]
@@ -95,10 +95,10 @@ jobs:
       args:
         build_url: '{fedora-latest/build_url}'
         update_packages: True
-        test_suite: test_integration/test_external_ca.py::TestExternalCA test_integration/test_external_ca.py::TestExternalCAConstraints test_integration/test_installation.py::TestInstallCA::test_certmonger_reads_token_HSM
+        test_suite: test_integration/test_external_ca.py::TestExternalCA test_integration/test_external_ca.py::TestExternalCAConstraints
         template: *testing-master-latest
         timeout: 4800
-        topology: *master_2repl_1client
+        topology: *master_1repl_1client
 
   fedora-latest/temp_commit_4:
     requires: [fedora-latest/build]
--- a/ipatests/prci_definitions/temp_commit.yaml
+++ b/ipatests/prci_definitions/temp_commit.yaml
@@ -69,7 +69,7 @@ jobs:
       args:
         build_url: '{fedora-latest/build_url}'
         update_packages: True
-        test_suite: test_integration/test_external_ca.py::TestExternalCA test_integration/test_external_ca.py::TestExternalCAConstraints test_integration/test_installation.py::TestInstallCA::test_certmonger_reads_token_HSM
+        test_suite: test_integration/test_external_ca.py::TestExternalCA test_integration/test_installation.py::TestInstallCA::test_certmonger_reads_token_HSM
         template: *testing-master-latest
         timeout: 4800
         topology: *master_2repl_1client
@@ -95,7 +95,7 @@ jobs:
       args:
         build_url: '{fedora-latest/build_url}'
         update_packages: True
-        test_suite: test_integration/test_external_ca.py::TestExternalCA test_integration/test_external_ca.py::TestExternalCAConstraints
+        test_suite: test_integration/test_external_ca.py::TestExternalCA
         template: *testing-master-latest
         timeout: 4800
         topology: *master_1repl_1client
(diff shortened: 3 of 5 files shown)


rcritten:
@stanislavlevin any ideas on how I can reproduce the certmonger failure? I tried on my own systems and picked one of the failing tests on F32 + updates-testing and it's still passing.

stanislavlevin:
@rcritten, this means that there is a difference between Docker image and vagrant one. I'm checking.

flo-renaud:
@rcritten 
Is it the same issue as this [one](https://pagure.io/freeipa/issue/8506) ? It happened on rawhide with certmonger-0.79.11-4.fc34.x86_64

rcritten:
Yes. There are now two behaviors seen:

* It blows up during uninstallation
* It blows up starting tracking on the CA certificates

It looks to me like the uninstall logs should be kept but it doesn't seem to be the case in practice so I'm focusing on the installation error. dbus is checked as running two steps prior to the start_tracking call. I'm bumping up the debug level so maybe we can see what if anything is going on inside of certmonger.

rcritten:
Ok, so I got somewhere on the uninstall failure.

It's at the step where it's trying to remove the dogtag-ipa-ca-renew-agent certmonger CA. The request for it begins at about 2020-10-08T19:09:54Z based on ipaserver-uninstall.log (its the last thing before the stack trace).

The removal begins in ipaserver/install/cainstance.py:1004 where it confirms that the nickname exists:

Oct 08 19:09:56 master.ipa.test certmonger[32626]: 2020-10-08 19:09:56 [32626] message 0x55a6cb2dcc60(method_call)->:1.75:/org/fedorahosted/certmonger:org.fedorahosted.certmonger.find_ca_by_nickname

That is successful which means dbus is working ok at that point. It takes less than a second for that.

Then certmonger logs the remove_ca request:

Oct 08 19:09:56 master.ipa.test certmonger[32626]: 2020-10-08 19:09:56 [32626] message 0x55a6cb2dcc60(method_call)->:1.75:/org/fedorahosted/certmonger:org.fedorahosted.certmonger.remove_known_ca

So it took 2 seconds between the last code logged in the installer and the logging within certmonger. That seems somewhat reasonable. certmonger was just restarted at this point so it had to run through all its CAs and tracked certs.

The next logging is related to the KDC being unavailable, after a 13 second delay:

Oct 08 19:10:09 master.ipa.test certmonger[32635]: Error obtaining initial credentials: Cannot contact any KDC for requested realm.
Oct 08 19:10:09 master.ipa.test certmonger[32635]: Error setting up ccache at the client: Cannot contact any KDC for requested realm.

There is another 6 second delay within certmonger. It seems it the request got put back into the queue on the failure.

After a bit the request is finally executed and the CA removed:

Oct 08 19:10:23 master.ipa.test certmonger[32626]: 2020-10-08 19:10:23 [32626] Removed file "/var/lib/certmonger/cas/20201008185202".

It's timing out after around 25 seconds and certmonger seems to be taking about 27 due to the KDC timeout.

I don't know what the default timeout is for a dbus request but perhaps it needs to be extended.

fcami:
A quick search yielded "25 seconds is indeed the _DBUS_DEFAULT_TIMEOUT_VALUE in libdbus, the C reference implementation."

stanislavlevin:
debug level 9:
https://dev.azure.com/slev0400/slev/_build/results?buildId=2101&view=results

Indeed, it seems like a timeout issue.

stanislavlevin:
Let me check with your changes.

stanislavlevin:
I'm afraid that `get_system_bus` wouldn't help.
This is a timeout workaround for `get_object`, but it doesn't set the default Reply timeout for dbus method call.
At first glance there is no such facility. Every method call have to provide `timeout` arg like:
`path = iface.find_ca_by_nickname("dogtag-ipa-ca-renew-agent", timeout=NSeconds)`

stanislavlevin:
based on this pr run:
https://dev.azure.com/slev0400/slev/_build/results?buildId=2113&view=logs&j=9bf0c493-406d-5b8e-d189-f64f53270fd1

rcritten:
The dbus code is quite dense and spread between two packages. I had been looking for the least intrusive change. Apparently this isn't it.

I'm curious what the root cause is as certmonger hasn't changed for months. I can see about making some changes, like avoiding the 10+ second delay if the kinit fails, but there are other delays that still opaque. Increasing the timeout may work but it would just be a bandaid.

rcritten:
I tested with a fixed certmonger package that resolves the NSS shutdown error. It was caused by a resource leak. It doesn't affect the speed though, as I suspected.

In the latest failure there is quite a long delay in authenticating with the token of 14 seconds:

Oct 12 13:51:55 replica0.ipa.test certmonger[31355]: 2020-10-12 13:51:55 [31355] Need login to token NSS Certificate DB
Oct 12 13:52:09 replica0.ipa.test certmonger[31355]: 2020-10-12 13:52:09 [31355] Looking for auditSigningCert cert-pki-ca

Theare are only two NSS calls in between these lines is:

PK11_Authenticate(sle->slot, PR_TRUE, &cb_data);

certs = PK11_ListCertsInSlot(sle->slot);

Since I have my private build working I'll see about adding additional debug logging to try to narrow down what is taking so long.

rcritten:
For some reason this call is sometimes taking 13 seconds.

certs = PK11_ListCertsInSlot(sle->slot);

rcritten:
The latest failure is in test_integration-test_external_ca.py-TestExternalCA-test_client_installation_with_otp during teardown.

certmonger can't get a ticket:

Oct 12 23:33:23 master.ipa.test certmonger[30500]: Error obtaining initial credentials: Generic error (see e-text).
Oct 12 23:33:23 master.ipa.test certmonger[30500]: Error setting up ccache at the client: Generic error (see e-text).

And the KDC confirms it:

Oct 12 23:33:23 master.ipa.test krb5kdc[30509](info): AS_REQ (6 etypes {aes256-cts-hmac-sha1-96(18), aes256-cts-hmac-sha384-192(20), camellia256-cts-cmac(26), aes128-cts-hmac-sha1-96(17), aes128-cts-hmac-sha256-128(19), camellia128-cts-cmac(25)}) 192.168.122.56: LOOKING_UP_CLIENT: host/master.ipa.test@IPA.TEST for krbtgt/IPA.TEST@IPA.TEST, Server error

389-ds seems to have been restarting at the time, with 389 disabled!?

[12/Oct/2020:23:33:20.365546759 +0000] - INFO - main - slapd stopped.
[12/Oct/2020:23:33:22.021960144 +0000] - NOTICE - config_set_port - Non-Secure Port Disabled
[12/Oct/2020:23:33:22.344948656 +0000] - INFO - main - 389-Directory/1.4.3.13 B2020.279.0000 starting up
[12/Oct/2020:23:33:22.355252920 +0000] - INFO - main - Setting the maximum file descriptor limit to: 524288

The full logs are are:

http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/e27601e4-0ce0-11eb-bd43-fa163e988597/test_integration-test_external_ca.py-TestExternalCA-test_client_installation_with_otp/master.ipa.test/var/log/

rcritten:
Opened https://github.com/freeipa/freeipa/pull/5197 to address the uninstall issue.

rcritten:
Ok, CI seems happier with my replacement of PK11_ListCertsInSlot() and scan with PK11_FindCertFromNickname(). I'll review the changes a bit more closely and maybe submit a PR in upstream certmonger.

rcritten:
I dropped the d-bus timeout changes as they 1) didn't work and 2) wouldn't have fixed the underlying problem

rcritten:
Closing this PR. certmonger-0.79.13 was released upstream to address the start-tracking problem and the uninstall issue is handled by 139d60d74706d2ba9855db5faefee322b23b0ba5

PR #5182, opened 2020-10-08
Speed up cainstance.migrate_profiles_to_ldap

tiran:
The ra_certprofile API is slow. It takes ~200ms to migrate and enable a
profile even when the profile already available. The migration step
slows down the installer and upgrader by about 12 to 15 seconds.

Skip all profiles that have been imported by Dogtag already.

Related: https://pagure.io/freeipa/issue/8522

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -23,6 +23,7 @@ from __future__ import print_function, absolute_import
 
 import base64
 import binascii
+import enum
 import logging
 
 import dbus
@@ -72,6 +73,8 @@ ADMIN_GROUPS = [
 ]
 
 ACME_AGENT_GROUP = 'ACME Agents'
+
+PROFILES_DN = DN(('ou', 'certificateProfiles'), ('ou', 'ca'), ('o', 'ipaca'))
 
 
 def check_ports():
@@ -1713,7 +1716,7 @@ def update_ca_renewal_entry(conn, nickname, cert):
 
 def ensure_ldap_profiles_container():
     ensure_entry(
-        DN(('ou', 'certificateProfiles'), ('ou', 'ca'), ('o', 'ipaca')),
+        PROFILES_DN,
         objectclass=['top', 'organizationalUnit'],
         ou=['certificateProfiles'],
     )
@@ -1898,7 +1901,6 @@ def import_included_profiles():
             api.env.container_certprofile, api.env.basedn)
         try:
             conn.get_entry(dn)
-            continue  # the profile is present
         except errors.NotFound:
             # profile not found; add it
             entry = conn.make_entry(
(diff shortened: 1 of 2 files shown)


ipaserver/install/cainstance.py
```diff
@@ -1982,6 +1988,15 @@ def migrate_profiles_to_ldap():
     profile_ids = match.group(1).split(',')
```
frasertweedale, line 1988: Commentary related to the profile upgrade / new profile discussion.
Revisiting this code again, we see that:
- Any Dogtag upgrade scripts that *add* new profiles (including referencing them in CS.cfg `profile.list`) will cause FreeIPA to also import that profile to LDAP.  This is fine in general.  If there is a conflict with custom profile name, the custom profile will be left alone.
- Any *updated* Dogtag profiles in the instance profile dir will not be applied in LDAP.

I don't have any comment whether this is good or bad or how it could be improved.  I just wanted to record my understanding of what currently happens in these scenarios.

frasertweedale:
Haven't tested but looks good.  We might need to further discuss whether to reconcile the enabled/disable state of profiles.  But I think this is OK as-is.

Another *potential* optimisation would be to ldapsearch all profiles at once and construct a map, avoiding multiple LDAP searches.  But that might be too much complexity for too little payoff.  Still, I would be curious if it makes a difference.

ipaserver/install/cainstance.py
tiran: I added more doc strings to the function that explain how migration behaves.

Regarding profile patching / updates: I think updates should be handled by Dogtag. Dogtag knows the previous state of a profile and should know how and when to update a profile.

ipaserver/install/cainstance.py
```diff
@@ -1982,6 +1988,15 @@ def migrate_profiles_to_ldap():
...
 
     for profile_id in profile_ids:
+        state = _check_ldap_profile_state(profile_id)
+        if state in (ProfileState.ENABLED, ProfileState.DISABLED):
+            logger.info(
+                "Profile '%s' is already in LDAP and %s; skipping",
+                profile_id, state.value
+            )
```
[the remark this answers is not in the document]
tiran: Good point! I changed the condition to ``if state != ProfileState.MISSING`` and added a comment.

frasertweedale:
LGTM.  Built, installed and smoke-tested `freeipa-server-upgrade`.  Adding needs-review tag for someone from FreeIPA team to look at it and give approval.

abbra:
LGTM

tiran:
master:

* a9d34c8e66590c6ddb18a227fa0201cacdc7d72d Speed up cainstance.migrate_profiles_to_ldap
* fa58071221bfb37159db6d6dc4f3bcfd088d80b1 Reuse main LDAP connection

PR #5183, opened 2020-10-08
ipatests: various enhancement to hidden replica tests

fcami:
+
    ipatests: add wait_for_ipa_to_start
    
    wait_for_ipa_to_start(host) waits for ipactl to return RUNNING for all
    IPA services on the specified host.

    Related: https://pagure.io/freeipa/issue/8534
+
    ipatests: hiddenreplica:  use wait_for_ipa_to_start after restore
    
    Use wait_for_ipa_to_start to wait until the restored replica is online.

    Related: https://pagure.io/freeipa/issue/8534
+
    ipatests: use wait_for_replication for hidden replica checks
    
    Previously, hidden replica checks were run without waiting for replication
    to complete, potentially leading to unstable behavior.
    Use wait_for_replication.

    Fixes: https://pagure.io/freeipa/issue/8534
+
    ipatests: hidden replica: misc fixes
    
    Split a test in two and add additional fixes.
    
    Related: https://pagure.io/freeipa/issue/8534

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -46,6 +46,7 @@ from cryptography import x509
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import rsa
 from cryptography.hazmat.backends import default_backend
+from datetime import datetime, timedelta
 
 from ipapython import certdb
 from ipapython import ipautil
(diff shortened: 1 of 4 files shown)


mrizwan93 (approved):
LGTM

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -2455,6 +2456,37 @@ def get_healthcheck_version(host):
     return healthcheck_version
 
 
+def wait_for_ipa_to_start(host, timeout=60):
```
rcritten, line 2459: The default constants.startup_timeout is configured in IPA as 120 seconds, and that's per-service. The CA can be slow to start so 60 seconds may be too short.

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -1024,11 +1036,15 @@ def test_hidden_replica_backup_and_restore(self):
             stdin_text=dirman_password + '\nyes'
         )
 
-        # give replication some time
+        # wait for the replica to be available
         time.sleep(5)
+        tasks.wait_for_ipa_to_start(self.replicas[0])
```
rcritten, line 1041: Is the sleep necessary ehre since wait_for_ipa_to_start() will also sleep?

abbra:
@fcami this PR seems to stuck -- could you please update it according to the Rob's comments?

abbra:
@fcami when you have time, please apply @rcritten's suggestions.

PR #5184, opened 2020-10-08
ipatests: run freeipa-healthcheck on hidden replica

fcami:
Run freeipa-healthcheck on a FreeIPA clusters with a hidden replica to make sure a hidden replica is considered fully healthy.
    
Fixes: https://pagure.io/freeipa/issue/8536

### DIFF
--- a/ipatests/test_integration/test_replica_promotion.py
+++ b/ipatests/test_integration/test_replica_promotion.py
@@ -11,13 +11,18 @@ import textwrap
 import pytest
 
 from ipatests.test_integration.base import IntegrationTest
+from ipatests.test_integration.test_ipahealthcheck import (
+    run_healthcheck, HEALTHCHECK_PKG
+)
 from ipatests.pytest_ipa.integration import tasks
 from ipatests.pytest_ipa.integration.tasks import (
-    assert_error, replicas_cleanup)
+    assert_error, replicas_cleanup
+)
 from ipatests.pytest_ipa.integration.firewall import Firewall
 from ipatests.pytest_ipa.integration.env_config import get_global_config
 from ipalib.constants import (
-    DOMAIN_LEVEL_1, IPA_CA_NICKNAME, CA_SUFFIX_NAME)
+    DOMAIN_LEVEL_1, IPA_CA_NICKNAME, CA_SUFFIX_NAME
+)
 from ipaplatform.paths import paths
 from ipapython import certdb
 from ipatests.test_integration.test_dns_locations import (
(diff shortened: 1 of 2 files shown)


ipatests/test_integration/test_replica_promotion.py
```diff
@@ -887,12 +890,32 @@ def _check_config(self, enabled=(), hidden=()):
...
+        """Ensure that ipa-healthcheck runs successfully on all members
+        of an IPA cluster that includes a hidden replica.
+        """
+        # verify state
+        self._check_config([self.master], [self.replicas[0]])
+        # A DNA range is needed on the replica for ipa-healthcheck to work.
+        # Create a user so that the replica gets a range.
+        tasks.user_add(self.replicas[0], 'testuser')
+        tasks.user_del(self.replicas[0], 'testuser')
+        returncode0, _unused = run_healthcheck(
```
flo-renaud, line 908: Prior to running healthcheck, the test should call tasks.install_packages() to install ipa-healthcheck. Forgetting to do so is not an issue when the test is executed inside PRCI because our Vagrant images contain the package by default, but if the test is launched outside of PRCI using for instance the steps in https://www.freeipa.org/page/Testing#Integration_tests, it may fail because of the missing pkg.

fcami: Good catch, thanks, will fix!

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -887,12 +892,29 @@ def _check_config(self, enabled=(), hidden=()):
...
+        of an IPA cluster that includes a hidden replica.
+        """
+        # verify state
+        self._check_config([self.master], [self.replicas[0]])
+        # A DNA range is needed on the replica for ipa-healthcheck to work.
+        # Create a user so that the replica gets a range.
+        tasks.user_add(self.replicas[0], 'testuser')
+        tasks.user_del(self.replicas[0], 'testuser')
+        for srv in (self.master, self.replicas[0]):
+            tasks.install_packages(srv, HEALTHCHECK_PKG)
```
kaleemsiddiqu, line 911: Can we move this to test class level so if we add more tests in future, this pkg install is applied for those tests as well?

kaleemsiddiqu:
@fcami , LGTM . Please remove the temp commit.

fcami:
Thanks for the review @flo-renaud and @kaleemsiddiqu - temp commit removed.

flo-renaud:
master:

* 7bbb997150fedf8ac5d8ef106414a0ffcf9e921b ipatests: tasks: add user_del
* 5de67028d07b998da0adf5d13ac70bef7c03e7a7 ipatests: run freeipa-healthcheck on hidden replica

PR #5194, opened 2020-10-13
ipatests: properly handle journalctl return code 

flo-renaud:
The test test_installation.py::TestInstallMaster::test_selinux_avcs
is failing when no AVCs are detected because it is calling
journalctl --full --grep=AVC--since=yesterday
and the command exits with return code 1.

Call the command with raiseonerr=False to support this case.

Fixes: https://pagure.io/freeipa/issue/8541

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -902,7 +902,7 @@ class TestInstallMaster(IntegrationTest):
         # installed by default and journalctl gives us all AVCs.
         result = self.master.run_command([
             "journalctl", "--full", "--grep=AVC", "--since=yesterday"
-        ])
+        ], raiseonerr=False)
         avcs = list(
             line.strip() for line in result.stdout_text.split('\n')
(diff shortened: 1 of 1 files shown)


mrizwan93 (approved):
LGTM.
Please remove temp commit after azure run completes.

flo-renaud:
Temp commit removed, all tests were green.

ipatests/test_integration/test_installation.py
```diff
@@ -902,7 +902,7 @@ def test_selinux_avcs(self):
         # installed by default and journalctl gives us all AVCs.
         result = self.master.run_command([
             "journalctl", "--full", "--grep=AVC", "--since=yesterday"
-        ])
```
fcami, line 905: ACK for when CI passes

fcami (approved):

flo-renaud:
master:

* cb7d09642205cba07ef614065d106f7f84d8921b ipatests: properly handle journalctl return code

PR #5197, opened 2020-10-14
Don't restart certmonger after stopping tracking in uninstall

rcritten:
certmonger was later restarted to remove the custom CA entries
and the startup delay sometimes caused uninstallation to fail.

certmonger is stopped in cainstance.py::uninstall() so it will
still be stopped post-install.

https://pagure.io/freeipa/issue/8533

There is no real test for this. The uninstall failure happens about once out of every five PR-CI runs in https://github.com/freeipa/freeipa/pull/5180

With this patch (and a few unrelated others) there have been 10 successful runs in that PR.

### DIFF
--- a/ipaserver/install/ca.py
+++ b/ipaserver/install/ca.py
@@ -425,7 +425,7 @@ def install_step_1(standalone, replica_config, options, custodia):
 
 def uninstall():
     ca_instance = cainstance.CAInstance(api.env.realm)
-    ca_instance.stop_tracking_certificates()
+    ca_instance.stop_tracking_certificates(stop_certmonger=False)
     ipautil.remove_file(paths.RA_AGENT_PEM)
     ipautil.remove_file(paths.RA_AGENT_KEY)
     if ca_instance.is_configured():


ipaserver/install/ca.py
```diff
@@ -425,7 +425,7 @@ def install_step_1(standalone, replica_config, options, custodia):
 
 def uninstall():
     ca_instance = cainstance.CAInstance(api.env.realm)
-    ca_instance.stop_tracking_certificates()
+    ca_instance.stop_tracking_certificates(False)
```
frasertweedale, line 428: [nit] could you write the whole kwarg `stop_certmonger=False` to give meaning to the bool, here on the caller side?

frasertweedale (approved):
LGTM.  One nit for readability.  If you agree to change it, make it so, and you have my ACK.  If you don't want to change it, you still have my ACK :)

ipaserver/install/ca.py
rcritten: Yeah, sorry for being lazy. Fixed.

abbra:
LGTM

rcritten:
master:

* 139d60d74706d2ba9855db5faefee322b23b0ba5 Don't restart certmonger after stopping tracking in uninstall

PR #5198, opened 2020-10-15
tox.ini: Extend max-line-length from 80 to 88+

fcami:
Change tox.ini's 80c character limit to 88.
- 88 is the limit for a font size of 14 on a FHD (1920x1200) screen
  with two editors side-by-side.
- A too-high number can become an issue for potential contributors
  with eyesight problems. So we want to avoid that.

Fixes: https://pagure.io/freeipa/issue/8546

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -56,6 +56,6 @@ commands=
 # E731 do not assign a lambda expression
 # E741 ambiguous variable name 'l'
 ignore = E203, E402, E231, W503, E731, E741
-max-line-length = 80
+max-line-length = 88
 # exclude auto-generated remote plugins
 exclude=.git,.venv,build,_build,rpmbuild,2_49,2_114,2_156,2_164


simo5:
Not that it matters much as I am not contributing much anymore, but I am against raising above 80.

The 80 rule is mostly for legibility issues and it is not a matter of TTY sizes or screen resolutions.

Also PEP-8

Also https://baymard.com/blog/line-length-readability

Also https://brainwagon.org/2007/07/07/are-80-columns-enough/

As for side-by-side considerations, when I edit code I use 4 vim buffers side-by-side on a ultra-wide screen, they just match 80 columns each, a larger limit would force me to get down to 3 buffers :-D

tiran:
PEP-8 explicitly allows [longer line length](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters.

frozencemetery:
I'm not a high contributor either, but I'm also not in favor of raising above 80.  On a 1080p screen, I get 4 buffers side-by-side.  While it's true that black defauls to 88, that's one of the few things it allows customization in; similarly flake8 defaults to 79, but that's also customizable.

While I don't advocate all aspects of Linux's code style, they do make the point that if the line length is a problem, it likely indicates that what is being attempted is too complex and should refactored, possibly breaking out into more functions.

rcritten:
I'm fine with 80 as well. I don't do vim buffers but I do have multiple terminals open at once. I can easily arrange six 24x80 terminals on my desktop with no overlap.

Where we generally run into trouble is with splitting error messages in really awkward ways.

abbra:
I am more struggling with a (relative) lack of vertical space and a bit of ugliness with error message split in Python code limited by 80 characters. If we can find a different solution to those than to widen the lines, so be it. ;)

abbra:
Looks like majority of commenters are against breaking 80 limit. Should we close this PR?

fcami:
@abbra most of the +1s from the meeting are not showing here.

abbra:
Ok, so what is the tally? :)

fcami:
On 2020-10-14, we had an internal meeting about $subject which resulted in the following votes:
[~100] 3 votes  
[we need to increase, do not know/do not care how much] 4 votes
[I do not care] 1 vote

Votes from this PR:
[do not increase] 3 votes

PR #5203, opened 2020-10-22
On password reset also set krbLastAdminUnlock to unlock account

rcritten:
On password reset also set krbLastAdminUnlock to unlock account

This fixes the case where an account is locked on one or more servers
and the password is reset by an administrator. The account would
remain locked on those servers for the duration of the lockout.

This is done by setting krbLastAdminUnlock to the current date and
time. The lockout plugin will see this and unlock the account. Since
the value should be replicated along with the password any server
that has the new password will also be unlocked.

This does incur an additional attribute that must be replicated,
whether it is needed or not, but since lockout is computed
per-server this is the only guaranteed way to be sure that the
account will be unlocked everywhere.

https://pagure.io/freeipa/issue/8551

### DIFF
--- a/daemons/ipa-slapi-plugins/ipa-pwd-extop/common.c
+++ b/daemons/ipa-slapi-plugins/ipa-pwd-extop/common.c
@@ -855,6 +855,20 @@ int ipapwd_SetPassword(struct ipapwd_krbcfg *krbcfg,
                                  data->expireTime, (data->expireTime == 0));
             if (ret != LDAP_SUCCESS)
                 goto free_and_return;
+
+            switch(data->changetype) {
+                case IPA_CHANGETYPE_DSMGR:
+                case IPA_CHANGETYPE_ADMIN:
+                    /* Mark as administratively reset which will unlock acct */
+                    ret = ipapwd_setdate(data->target, smods, 
+                                         "krbLastAdminUnlock",
+                                         data->timeNow, false);
+                    if (ret != LDAP_SUCCESS)
+                        goto free_and_return;
+                    break;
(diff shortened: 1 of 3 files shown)


rcritten:
Some additional thinking on my part. I can add this to the commit message if desired.

Since lockout is computed and not stored trying to detect the condition on every server on every password change could result in more replication and the chance of conflicts. My original thought was to grab password replication events and detect whether the user was locked out and unlock them, but on any given server you can only know if the user is locked out on that server by computing it. I could move some of the lockout code from the ipa-lockout plugin to more easily compute that.

To unlock a user we'd need to update krbLastAdminUnlock which would be replicated, possibly on multiple servers leading to the chance of conflicts.

We could set this attribute on user-driven password changes as well but the original ask and my thinking are that if you forgot your password and got locked out, how can you change it yourself? Upon reflection I guess a user could fat-finger it a bunch of times against one IPA server then have a revelation and log in against a different server. So they would still be locked out for the duration on the first one. I'm not sure the extra replication is worth it for user-generated password changes.

abbra:
Please update the commit message. Other than that, looks good.

rcritten:
I revised the commit message, slightly differently than my proposal, and dropped the temporary commit.

abbra:
ACK.

abbra:
master:

* 3ab3578b36bc7b8ce777e38ba764649d1b684ce5 On password reset also set krbLastAdminUnlock to unlock account
* ca6fc689ba92ee2db6945f37429ee7fe8d75a4b6 ipatests: Test that password reset unlocks users too

PR #5206, opened 2020-10-23
ipa-kdb: support subordinate/superior UPN suffixes

abbra:
[MS-ADTS] 6.1.6.9.3.2 requires msDS-TrustForestTrustInfo attribute of
trusted domain information in Active Directory to conform certain rules.
One side-effect of those rules is that list of UPN suffixes reported
through the netr_DsRGetForestTrustInformation function is dynamically
filtered to deduplicate subordinate suffixes.

It means that if list of UPN suffixes contains the following top level
names (TLNs):

  fabrikam.com
  sub.fabrikam.com

then netr_DsRGetForestTrustInformation would only return 'fabrikam.com'
as the TLN, fully filtering 'sub.fabrikam.com'.

IPA KDB driver used exact comparison of the UPN suffixes so any
subordinate had to be specified exactly.

Modify logic so that if exact check does not succeed, we validate a
realm to test being a subordinate of the known UPN suffixes. The
subordinate check is done by making sure UPN suffix is at the end of the
test realm and is immediately preceded with a dot.

Because the function to check suffixes potentially called for every
Kerberos principal, precalculate and cache length for each UPN suffix at
the time we retrieve the list of them.

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb_mspac.c
+++ b/daemons/ipa-kdb/ipa_kdb_mspac.c
@@ -2393,6 +2393,7 @@ void ipadb_mspac_struct_free(struct ipadb_mspac **mspac)
                     free((*mspac)->trusts[i].upn_suffixes[j]);
                 }
                 free((*mspac)->trusts[i].upn_suffixes);
+                free((*mspac)->trusts[i].upn_suffixes_len);
             }
         }
         free((*mspac)->trusts);
@@ -2600,6 +2601,24 @@ krb5_error_code ipadb_mspac_get_trusted_domains(struct ipadb_context *ipactx)
             } else {
                 ret = EINVAL;
                 goto done;
+            }
+        }
+
+        t[n].upn_suffixes_len = NULL;
+        if (t[n].upn_suffixes != NULL) {
+            size_t len = 0;
+
+            for (; t[n].upn_suffixes[len] != NULL; len++);
+
+            if (len != 0) {
+                t[n].upn_suffixes_len = calloc(n, sizeof(size_t));
+                if (t[n].upn_suffixes_len == NULL) {
+                    ret = ENOMEM;
+                    goto done;
+                }
+                for (i = 0; i < len; i++) {
+                    t[n].upn_suffixes_len[i] = strlen(t[n].upn_suffixes[i]);
+                }
             }
         }
 
@@ -2972,6 +2991,17 @@ krb5_error_code ipadb_is_princ_from_trusted_realm(krb5_context kcontext,
 				result = strncasecmp(test_realm,
 						     ipactx->mspac->trusts[i].upn_suffixes[j],
 						     size) == 0;
+				if (!result) {
+					/* if UPN suffix did not match exactly, find if it is
+					 * superior to the test_realm, e.g. if test_realm ends
+					 * with the UPN suffix prefixed with dot*/
+					size_t len = ipactx->mspac->trusts[i].upn_suffixes_len[j];
+					if ((size > len) && (test_realm[size - len - 1] == '.')) {
+						result = strncasecmp(test_realm + (size - len),
+								     ipactx->mspac->trusts[i].upn_suffixes[j],
+								     len) == 0;
+					}
+				}
 				if (result)
 					break;
 			}
--- a/daemons/ipa-kdb/ipa_kdb_mspac_private.h
+++ b/daemons/ipa-kdb/ipa_kdb_mspac_private.h
@@ -48,6 +48,7 @@ struct ipadb_adtrusts {
     struct ipadb_adtrusts *parent;
     char *parent_name;
     char **upn_suffixes;
+    size_t *upn_suffixes_len;
 };
 
 int string_to_sid(const char *str, struct dom_sid *sid);


abbra:
The algorithm used in the code can be tested with this small program:
```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>

int main(int argc, const char *argv[]) {
	char **upn_suffixes = NULL;
	size_t *upn_suffixes_len = NULL;
	size_t dlen = 0;
	char *domain = NULL;

	if (argc < 3) {
		printf("%s: domain-to-check UPN-suffix UPN-suffix ...\n", argv[0]);
		return 1;
	}

	upn_suffixes = calloc(argc - 1, sizeof(char *));
	upn_suffixes_len = calloc(argc - 2, sizeof(size_t));

	for (int i = 0; i < argc - 2 ; i++) {
		upn_suffixes[i] = strdup(argv[2 + i]);
		upn_suffixes_len[i] = strlen(upn_suffixes[i]);
	}
	
	domain = strdup(argv[1]);
	dlen = strlen(domain);

	for (int i = 0; upn_suffixes[i] != NULL; i++) {
		int result = strncasecmp(domain, upn_suffixes[i], dlen);
		if (result != 0) {
                        size_t len = upn_suffixes_len[i];
                        if ((dlen > len) && (domain[dlen - len - 1] == '.')) {
                                result = strncasecmp(domain + (dlen - len),
                                                     upn_suffixes[i],
                                                     len);
                        }
		}
		if (result == 0) {
			printf("'%s' matched domain '%s'\n", upn_suffixes[i], domain);
			break;
		}
	}

	return 0;
	
}

```

Here is an example:
```
$ gcc -o subordinate subordinate.c
$ ./subordinate sub.fabrikam.com fabrikam.com test1.lan test2.lan
'fabrikam.com' matched domain 'sub.fabrikam.com'
```

daemons/ipa-kdb/ipa_kdb_mspac.c
```diff
@@ -2603,6 +2604,24 @@ krb5_error_code ipadb_mspac_get_trusted_domains(struct ipadb_context *ipactx)
...
+        t[n].upn_suffixes_len = NULL;
+        if (t[n].upn_suffixes != NULL) {
+            size_t len = 0;
+
+            for (; t[n].upn_suffixes[len] != NULL; len++);
+
+            if (len != 0) {
+                t[n].upn_suffixes_len = calloc(n, sizeof(size_t));
+                if (t[n].upn_suffixes_len == NULL) {
+                    ret = EINVAL;
```
rcritten, line 2616: Should it return ENOMEM instead of EINVAL?

rcritten:
Just the one question, otherwise LGTM.

daemons/ipa-kdb/ipa_kdb_mspac.c
abbra: Fixed, thanks.

abbra:
Fixed the return code and added a link to the ticket https://pagure.io/freeipa/issue/8554

frozencemetery (approved):
No problems detected.

rcritten:
master:

* 8b6d1ab854387840f7526d6d59ddc7102231957f ipa-kdb: support subordinate/superior UPN suffixes

rcritten:
ab a manual backport is necessary, can you handle it please?

sumit-bose:
Hi @abbra,

@menonsudhir, asked me about a test failure he is seeing. The trusted forest root is `ad.test` and there is a upn suffix `new.ad.test` defined. It looks like `new.ad.test` is not reported back by AD explicitly since it is a sub-domain of `ad.test`. If I understand the patch correctly it currently only checks if the given realm is a sub-domain of found upn suffixes but not if it is a sub-domain of the trusted domain itself. As a result a `kinit -E Administrator@new.ad.test` fails with `kinit: Client 'Administrator\@new.ad.test@IPA.TEST' not found in Kerberos database while getting initial credentials`.

bye,
Sumit

abbra:
@sumit-bose I think you are right. What about the following patch?
```
diff --git a/daemons/ipa-kdb/ipa_kdb_mspac.c b/daemons/ipa-kdb/ipa_kdb_mspac.c
index bb9d85c10..07c433e14 100644
--- a/daemons/ipa-kdb/ipa_kdb_mspac.c
+++ b/daemons/ipa-kdb/ipa_kdb_mspac.c
@@ -2975,10 +2975,20 @@ krb5_error_code ipadb_is_princ_from_trusted_realm(krb5_context kcontext,
 
        /* Iterate through list of trusts and check if input realm belongs to any of the trust */
        for(i = 0 ; i < ipactx->mspac->num_trusts ; i++) {
+               size_t len = 0;
                result = strncasecmp(test_realm,
                                     ipactx->mspac->trusts[i].domain_name,
                                     size) == 0;
 
+               if (!result) {
+                       len = strlen(ipactx->mspac->trusts[i].domain_name);
+                       if ((size > len) && (test_realm[size - len - 1] == '.')) {
+                               result = strncasecmp(test_realm + (size - len),
+                                                    ipactx->mspac->trusts[i].domain_name,
+                                                    len) == 0;
+                       }
+               }
+
                 if (!result && (ipactx->mspac->trusts[i].flat_name != NULL)) {
                        result = strncasecmp(test_realm,
                                             ipactx->mspac->trusts[i].flat_name,
@@ -2994,7 +3004,7 @@ krb5_error_code ipadb_is_princ_from_trusted_realm(krb5_context kcontext,
                                        /* if UPN suffix did not match exactly, find if it is
                                         * superior to the test_realm, e.g. if test_realm ends
                                         * with the UPN suffix prefixed with dot*/
-                                       size_t len = ipactx->mspac->trusts[i].upn_suffixes_len[j];
+                                       len = ipactx->mspac->trusts[i].upn_suffixes_len[j];
                                        if ((size > len) && (test_realm[size - len - 1] == '.')) {
                                                result = strncasecmp(test_realm + (size - len),
                                                                     ipactx->mspac->trusts[i].upn_suffixes[j],

```

sumit-bose:
Hi,

yes, looks good. Since you now do the same check in two places it might be worth to add a `is_subdomain_of()` function to make the code more readable?

bye,
Sumit

abbra:
Well, I don't think it is needed here -- I am thinking about backports mostly ;)

abbra:
@menonsudhir please pick up a fix from https://github.com/freeipa/freeipa/pull/5287 into your test branch.

menonsudhir:
@sumit-bose thank you for looking in to the test to confirm its not sssd issue.
@abbra will pick up the fix from #5827, thank you for the patch.

PR #5207, opened 2020-10-23
Use a state to determine if a 389-ds upgrade is in progress

rcritten:
When applying update files to 389 the listeners are disabled.
There is a large try/except around this so that if a failure
happens then the configuration should be automatically
restored.

We've seen multiple cases where this doesn't occur. Best guess
is that users are killing or ^C breaking out of the script.

What happens in that case is that when the next upgrade is run
the configuration is backed up again overwriting the original
values. This leaves dirsrv with no listener on 389.

Add a new state, upgrade-in-progress, so that the backup of the
config information can be skipped when the upgrader is executed
again after a failure.

The idea behind using a new state value is that if additional
attributes are ever backed up we don't need to remember to update
the list of possible saved values to check to decide if the
upgrade is in progress.

https://pagure.io/freeipa/issue/7534

**NOTE**: automated testing is very difficult because getting the timing right would be nearly impossible. I manually tested in two ways:

* Used the pdb module so that the upgrade stopped in the middle and I killed the upgrade process
* Spammed the keyboard with ^C in the middle of the DS upgrade process

Look for the last three elements in ```/var/lib/ipa/sysrestore/sysrestore.state```
```
[dirsrv]
serverid = EXAMPLE-TEST
enabled = True
upgrade-in-progress = True
nsslapd-port = 389
nsslapd-security = on
schema_compat_enabled = on
```
If these are left over killing the upgrader then re-running it should debug log that the values aren't being saved again.

### DIFF
--- a/ipaserver/install/upgradeinstance.py
+++ b/ipaserver/install/upgradeinstance.py
@@ -134,6 +134,11 @@ class IPAUpgrade(service.Service):
 
     def __save_config(self):
         shutil.copy2(self.filename, self.savefilename)
+        if self.get_state('upgrade-in-progress') is not None:
+            logger.debug('Previous upgrade in process, not saving config')
+            return
+        else:
+            self.backup_state('upgrade-in-progress', True)
         with open(self.filename, "r") as in_file:
             parser = GetEntryFromLDIF(in_file, entries_dn=["cn=config"])
             parser.parse()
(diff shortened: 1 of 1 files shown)


abbra:
LGTM

rcritten:
master:

* 2f8eb73f588c96b391be57020ecf0b247c646d19 Use a state to determine if a 389-ds upgrade is in progress

PR #5212, opened 2020-10-28
New validation efforts in 389-ds-base require that the backend entry for

mreynolds389:
a database be created before the mapping tree entry. This enforces that
the mapping tree entry (the suffix) actually belongs to an existing backend.

For IPA we simply need to reverse the order of the backend vs mapping tree
creation in cainstance.py -> __create_ds_db()

Fixes: https://pagure.io/freeipa/issue/8558

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -1162,17 +1162,6 @@ class CAInstance(DogtagInstance):
         backend = 'ipaca'
         suffix = DN(('o', 'ipaca'))
 
-        # replication
-        dn = DN(('cn', str(suffix)), ('cn', 'mapping tree'), ('cn', 'config'))
-        entry = api.Backend.ldap2.make_entry(
-            dn,
-            objectclass=["top", "extensibleObject", "nsMappingTree"],
-            cn=[suffix],
-        )
-        entry['nsslapd-state'] = ['Backend']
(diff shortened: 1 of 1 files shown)


rcritten:
The code change LGTM.

nit: the first line of commit should be readable on its own so git shortlog is useful.

Suggestion: Reorder creation of the CA mapping tree and database backend

Just stick this as the top line of the commit with the existing message intact, eg:

```
Reorder creation of the CA mapping tree and database backend

New validation efforts in 389-ds-base require that the backend entry for

a database be created before the mapping tree entry. This enforces that
the mapping tree entry (the suffix) actually belongs to an existing backend.

For IPA we simply need to reverse the order of the backend vs mapping tree
creation in cainstance.py -> __create_ds_db()

Fixes: https://pagure.io/freeipa/issue/8558
```

mreynolds389:
Hi Rob,

Yeah I just started noticing the qwerks of github as the DS team just moved to it.  I will reformat the commit message now...

abbra:
I added ack but didn't notice that whole test suite didn't run. Removing ACK and asking PR CI for the full run.

abbra:
LGTM

rcritten:
master:

* 9c4785f042feb1fddbe579d07ff591a532e4f24d Reorder creation of the CA mapping tree and database backend

PR #5213, opened 2020-10-28
rpcserver: fallback to non-armored kinit in case of trusted domains

abbra:
MIT Kerberos implements FAST negotiation as specified in RFC 6806
section 11. The implementation relies on the caller to provide a hint
whether FAST armoring must be used.

FAST armor can only be used when both client and KDC have a shared
secret. When KDC is from a trusted domain, there is no way to have a
shared secret between a generic Kerberos client and that KDC.

[MS-KILE] section 3.2.5.4 'Using FAST When the Realm Supports FAST'
allows KILE clients (Kerberos clients) to have local settings that
direct it to enforce use of FAST. This is equal to the current
implementation of 'kinit' utility in MIT Kerberos requiring to use FAST
if armor cache (option '-T') is provided.

[MS-KILE] section 3.3.5.7.4 defines a way for a computer from a
different realm to use compound identity TGS-REQ to create FAST TGS-REQ
explicitly armored with the computer's TGT. However, this method is not
available to IPA framework as we don't have access to the IPA server's
host key. In addition, 'kinit' utility does not support this method.

Active Directory has a policy to force use of FAST when client
advertizes its use. Since we cannot know in advance whether a principal
to obtain initial credentials for belongs to our realm or to a trusted
one due to enterprise principal canonicalization, we have to try to
kinit. Right now we fail unconditionally if FAST couldn't be used and
libkrb5 communication with a KDC from the user realm (e.g. from a
trusted forest) causes enforcement of a FAST.

In the latter case, as we cannot use FAST anyway, try to kinit again
without advertizing FAST. This works even in the situations when FAST
enforcement is enabled on Active Directory side: if client doesn't
advertize FAST capability, it is not required. Additionally, FAST cannot
be used for any practical need for a trusted domain's users yet.

### DIFF
--- a/ipalib/errors.py
+++ b/ipalib/errors.py
@@ -245,6 +245,12 @@ class PluginModuleError(PrivateError):
     format = '%(name)s is not a valid plugin module'
 
 
+class KrbPrincipalWrongFAST(PrivateError):
+    """
+    Raised when it is not possible to use our FAST armor for kinit
+    """
+    format = '%(principal) cannot use Anonymous PKINIT as a FAST armor'
+
 ##############################################################################
 # Public errors:
 
--- a/ipaserver/rpcserver.py
+++ b/ipaserver/rpcserver.py
@@ -972,7 +972,7 @@ class login_password(Backend, KerberosSession):
 
         try:
             query_dict = parse_qs(query_string)
-        except Exception as e:
+        except Exception:
             return self.bad_request(environ, start_response, "cannot parse query data")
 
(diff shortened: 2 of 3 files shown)


rcritten:
master:

* 9ba0494c266a3c918dba5d3c8ead8e06be819f3b pylint: remove unused variable
* b8b46779dc239cf5b007ad11b5b8c9cf51e34ed0 rpcserver: fallback to non-armored kinit in case of trusted domains

PR #5214, opened 2020-10-29
ipatests: ipa-healthcheck test for DS BackendsCheck

menonsudhir:
This testcase checks that the BackendsCheck reports the CRITICAL status when dse.ldif present in the
DS instance directory is renamed/moved.

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -967,6 +967,45 @@ class TestIpaHealthCheck(IntegrationTest):
             # synced. Help chrony by resetting the date
             self.master.run_command(['date', '-s', now_str])
 
+    @pytest.fixture
+    def rename_ldif(self):
+        """Fixture to rename dse.ldif file and revert after test"""
+        instance = realm_to_serverid(self.master.domain.realm)
+        self.master.run_command(
+            [
+                "mv",
+                paths.ETC_DIRSRV_SLAPD_INSTANCE_TEMPLATE % instance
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -967,6 +967,45 @@ def execute_expiring_check(check):
...
+    def rename_ldif(self):
+        """Fixture to rename dse.ldif file and revert after test"""
+        instance = realm_to_serverid(self.master.domain.realm)
+        self.master.run_command(
+            [
+                "mv",
+                paths.ETC_DIRSRV_SLAPD_INSTANCE_TEMPLATE % instance
+                + "/dse.ldif",
+                paths.ETC_DIRSRV_SLAPD_INSTANCE_TEMPLATE % instance
+                + "/_dse.ldif",
```
rcritten, line 980: The _dse seems a bit subtle. Can you name it something like dse.ldif.rename or something? I had to stare at this a bit to see the single-character difference.

rcritten (changes requested):
Otherwise it looks good to me.

ipatests/test_integration/test_ipahealthcheck.py
menonsudhir: @rcritten, the file has been changed to dse.ldif.renamed.

rcritten (approved):

rcritten:
master:

* 9a41966a250a97e7df204b4e66a4e7e9546d78aa ipatests: ipa-healthcheck test for DS BackendsCheck

PR #5215, opened 2020-10-30
ipatests: ipa-healthcheck test for DS EncryptionCheck

menonsudhir:
This testcase checks that EncryptionCheck reports ERROR status when DS tls version is modified to TLS1.0.


rcritten:
Looks good, please remove the temp commit.

rcritten:
master:

* b8a2a0f3013139574ac42ae69eb3e37c9ff56fba ipatests: ipa-healthcheck test for EncryptionCheck

PR #5223, opened 2020-11-02
ipatests: curl outputs the cookie in stderr and not in sdtout

flo-renaud:
The integration test test_trust.py::TestTrust::test_password_login_as_aduser
is expecting curl to output the cookie obtained after password login
in stdout but should use stderr instead.

Fixes: https://pagure.io/freeipa/issue/8559


rcritten:

abbra:
LGTM, thank you for spotting this!

abbra:
master:

* c053b5e0d6a06ad17b1d427a8345bca792981b47 ipatests: curl outputs the cookie in stderr and not in sdtout

PR #5230, opened 2020-11-06
wgi/plugins.py: ignore empty plugin directories

abbra:
Dynamic plugin registry returns as a plugin any folder within the
plugins directory. Web UI then attempts to load for each plugin 'foo' a
JavaScript file named 'foo/foo.js'. The problem is that if 'foo/foo.js'
does not exist, Web UI breaks and it is impossible to recover until the
empty folder is removed or 'foo/foo.js' (even empty) is created at the
server side.

Check that 'foo/foo.js' actual exists when including a plugin into the
registry.

Test the registry generator by creating fake plugins and removing them
during the test.

### DIFF
--- a/ipatests/test_ipaserver/test_jsplugins.py
+++ b/ipatests/test_ipaserver/test_jsplugins.py
@@ -1,0 +1,68 @@
+# Copyright (C) 2020  FreeIPA Contributors see COPYING for license
+
+import os
+import pytest
+
+from ipatests.test_ipaserver.httptest import Unauthorized_HTTP_test
+from ipatests.util import assert_equal, assert_not_equal
+from ipaplatform.paths import paths
+
+
+@pytest.mark.tier1
+class test_jsplugins(Unauthorized_HTTP_test):
+    app_uri = '/ipa/ui/js/freeipa/plugins.js'
+    jsplugins = (('foo', 'foo.js'), ('bar', ''))
+    content_type = 'application/javascript'
+
+    def test_jsplugins(self):
+        empty_response = "define([],function(){return[];});"
+
+        # Step 1: make sure default response has no additional plugins
+        response = self.send_request(method='GET')
+        assert_equal(response.status, 200)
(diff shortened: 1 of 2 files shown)


abbra:
This is a work in progress. I also need to open a ticket. Orignal bugzilla is https://bugzilla.redhat.com/show_bug.cgi?id=1894800

abbra:
The test succeeded:

```
2020-11-06T10:25:27.6897630Z 1	[2020-11-06 10:25:27] test_ipaserver/test_jsplugins.py::test_jsplugins::test_jsplugins [32mPASSED[0m[36m  [ 59%][0m
```

ipatests/test_ipaserver/test_jsplugins.py
```diff
@@ -0,0 +1,68 @@
...
+                dir = os.path.join(paths.IPA_JS_PLUGINS_DIR, d)
+                file = os.path.join(dir, f)
+                if f and os.path.exists(file):
+                    os.unlink(file)
+                if os.path.exists(dir):
+                    os.rmdir(dir)
+        except OSError:
+            pass
+
+        # Step 5: make sure default response has no additonal plugins
```
flo-renaud, line 64: Nitpick: an i is missing in addit*i*onal

flo-renaud (commented):
Hi @abbra 
thanks for the PR, LGTM. I also ran the test against an instance not updated and the issue was correctly detected.
I will ACK when the commit msg is updated with ticket number https://pagure.io/freeipa/issue/8567

ipatests/test_ipaserver/test_jsplugins.py
abbra: Fixed, thanks.

abbra:
Updated the ticket reference.

rcritten:
master:

* 91706690e0c894864c93716c4fbecb285722e77d wgi/plugins.py: ignore empty plugin directories

PR #5232, opened 2020-11-06
ipa-kdb: fix crash in MS-PAC cache init code

abbra:
When initializing UPN suffixes, we calculate their sizes and didn't use
the right variable to allocate their size. This affects us if there are
more than one UPN suffix available for a trust due to memory corruption
while filling in sizes.

Add unit test for multiple UPN suffixes.

Fixes: https://pagure.io/freeipa/issue/8566

### DIFF
--- a/daemons/ipa-kdb/tests/ipa_kdb_tests.c
+++ b/daemons/ipa-kdb/tests/ipa_kdb_tests.c
@@ -71,6 +71,10 @@ struct test_ctx {
 #define DOM_SID "S-1-5-21-1-2-3"
 #define DOM_SID_TRUST "S-1-5-21-4-5-6"
 #define BLOCKLIST_SID "S-1-5-1"
+#define NUM_SUFFIXES 10
+#define SUFFIX_TEMPLATE "d%0d" DOMAIN_NAME
+#define TEST_REALM_TEMPLATE "some." SUFFIX_TEMPLATE
+#define EXTERNAL_REALM "WRONG.DOMAIN"
 
 static int setup(void **state)
 {
@@ -91,6 +95,9 @@ static int setup(void **state)
 
     ipa_ctx = calloc(1, sizeof(struct ipadb_context));
     assert_non_null(ipa_ctx);
+
+    kerr = krb5_get_default_realm(krb5_ctx, &ipa_ctx->realm);
+    assert_int_equal(kerr, 0);
 
     ipa_ctx->mspac = calloc(1, sizeof(struct ipadb_mspac));
     assert_non_null(ipa_ctx->mspac);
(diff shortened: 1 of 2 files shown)


daemons/ipa-kdb/tests/ipa_kdb_tests.c
```diff
@@ -478,6 +494,44 @@ void test_dom_sid_string(void **state)
...
+{
+    struct test_ctx *test_ctx;
+
+    test_ctx = (struct test_ctx *) *state;
+
+    for(size_t i = 0; i < NUM_SUFFIXES; i++) {
+        char *test_realm = NULL;
+        char *trusted_realm = NULL;
+        asprintf(&test_realm, TEST_REALM_TEMPLATE, i);
+	krb5_error_code kerr = 0;
```
rcritten, line 507: visual seems off, tabs vs spaces?

daemons/ipa-kdb/tests/ipa_kdb_tests.c
```diff
@@ -478,6 +494,44 @@ void test_dom_sid_string(void **state)
...
+
+        if (test_realm) {
```
rcritten, line 509: I think this implies that if there is low memory and the asprintf fails (test_realm==NULL) then the test won't be executed.

daemons/ipa-kdb/tests/ipa_kdb_tests.c
```diff
@@ -478,6 +494,44 @@ void test_dom_sid_string(void **state)
...
+                       test_ctx->krb5_ctx,
+                       test_realm,
+                       strlen(test_realm),
+                       &trusted_realm);
+            assert_int_equal(kerr, 0);
+            free(test_realm);
+            free(trusted_realm);
+        }
+
+        asprintf(&test_realm, EXTERNAL_REALM, i);
```
rcritten, line 520: The variable i is unused because the are no replacements in the EXTERNAL_REALM template.

daemons/ipa-kdb/tests/ipa_kdb_tests.c
abbra: Just to keep a good hygiene here.

daemons/ipa-kdb/tests/ipa_kdb_tests.c
abbra: I changed this part to simply use external realm reference directly. Also moved it out of the loop as there is no need to build the variable anymore.

rcritten:
This looks good.

rcritten:
Seems like something has changed at a low level. The dnssec tests are failing comparing a string IP address to an object:

'172.20.0.3' = <ipatests.pytest_ipa.integration.host.Host replica1.ipa.test (replica)>.ip

I haven't found any changes that would have caused this yet.

rcritten:
Never mind I guess, it's magically passing now.

frozencemetery (approved):

rcritten:
master:

* 81cbee4e3ff2e667946e0d41097b402257608b7e ipa-kdb: fix crash in MS-PAC cache init code

PR #5239, opened 2020-11-09
dns: allow PTR records in arbitrary zones

frasertweedale:
PTR records in zones other than in-addr.arpa and in6.arpa are legal,
e.g. DNS-SD [RFC6763] uses such records.  If in a reverse zone
proceed with the existing checks, otherwise just accept the record.

Fixes: https://pagure.io/freeipa/issue/5566

### DIFF
--- a/ipaserver/plugins/dns.py
+++ b/ipaserver/plugins/dns.py
@@ -3111,10 +3111,11 @@ class dnsrecord(LDAPObject):
                 zone_len = REVERSE_DNS_ZONES[valid_zone]
 
         if not zone_len:
-            allowed_zones = ', '.join([unicode(revzone) for revzone in
-                                       REVERSE_DNS_ZONES])
-            raise errors.ValidationError(name='ptrrecord',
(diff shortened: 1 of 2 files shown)


abbra:
I wonder if we should have a test that adds PTR records in in-addr.arpa, to make sure they are served correctly by bind-dyndb-ldap.

frasertweedale:
@abbra `test_dns_plugin.py` does have tests for adding DNS records including PTR (in `in-addr.arpa` but .  But it doesn't then actually query them via DNS to verify.  But I think that is a more general testing improvement that could be addressed separately.

There aren't, however, any tests for creating PTR records *not* in the `in-addr.arpa` or `ip6.arpa` zones so I should probably add one.

abbra:
LGTM

abbra:
master:

* b153b23c75a8ddd4170c3ba47774ced219bf20aa dns: allow PTR records in arbitrary zones

PR #5245, opened 2020-11-10
ipa-kdb: implement AS-REQ lifetime jitter

frozencemetery:
Jitter is always enabled, so there is no additional configuration.
Update design docs to fix rationale and include information about
current behavior.

An earlier version of this patch was authored by Becky Shanley.

[Supersedes #5055]

### DIFF
--- a/ipatests/test_integration/test_krbtpolicy.py
+++ b/ipatests/test_integration/test_krbtpolicy.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2019  FreeIPA Contributors see COPYING for license
+# Copyright (C) 2019,2020  FreeIPA Contributors see COPYING for license
 #
 
 """
@@ -8,6 +8,7 @@ Module provides tests for Kerberos ticket policy options
 
 from __future__ import absolute_import
 
+import pytest
 import time
 from datetime import datetime
 
@@ -44,8 +45,15 @@ def reset_to_default_policy(host, user):
     """Reset default user authentication and user authentication type"""
     tasks.kinit_admin(host)
     host.run_command(['ipa', 'user-mod', user, '--user-auth-type='])
-    host.run_command(['ipa', 'krbtpolicy-reset'])
     host.run_command(['ipa', 'krbtpolicy-reset', user])
+
+
+def kinit_check_life(master, user):
+    """Acquire a TGT and check if it's within the lifetime window"""
+    master.run_command(["kinit", user], stdin_text=f"{PASSWORD}\n")
+    result = master.run_command("klist | grep krbtgt")
+    assert maxlife_within_policy(result.stdout_text, MAXLIFE,
+                                 slush=60) is True
 
 
 class TestPWPolicy(IntegrationTest):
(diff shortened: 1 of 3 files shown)


rcritten:
If this will be backported to the ipa-4-8 branch it would be a lot easier to backport if the doc change is a separate commit as this directory doesn't existing below current master.

abbra:
Could you please add 

```
Fixes: https://pagure.io/freeipa/issue/8010
```
to the commit message? I think this is remaining part of that ticket.

frozencemetery:
Done and done!

ipatests/test_integration/test_krbtpolicy.py
```diff
@@ -177,3 +185,32 @@ def test_krbtpolicy_otp(self):
...
+
+    def test_krbtpolicy_jitter_otp(self):
+        """Test jitter lifetime with OTP"""
+        tasks.kinit_admin(self.master)
+        self.master.run_command(["ipa", "user-mod", USER1,
+                                 "--user-auth-type", "otp"])
+        self.master.run_command(["ipa", "config-mod",
+                                 "--user-auth-type", "otp"])
+        kinit_check_life(self.master, USER1)
+        self.master.run_command(["ipa", "config-mod", "--user-auth-type="])
```
rcritten, line 203: You can move resetting the global auth type back to empty using a pytest fixture:

```
@pytest.fixture
def reset_policy(self):
    yield
    self.master.run_command(["ipa", "config-mod", "--user-auth-type="])
    tasks.kdestroy_all(self.master)
```

And define each test ala:
```
def test_krb5policy_foo(self, reset_policy)
```

frozencemetery: Done, thanks for the suggestion!

abbra:
Linter says there are issues:

```
************* Module ipatests.test_integration.test_krbtpolicy
ipatests/test_integration/test_krbtpolicy.py:70: [E0602(undefined-variable), TestPWPolicy.reset_policy] Undefined variable 'pytest')
```
You can run `make fastlint` locally. ;)

frozencemetery:
Updated and rebased.

ipatests/test_integration/test_krbtpolicy.py
```diff
@@ -176,4 +183,27 @@ def test_krbtpolicy_otp(self):
...
+
+    def test_krbtpolicy_jitter_pkinit(self, reset_policy):
+        """Test jitter lifetime with PKINIT"""
+        tasks.kinit_admin(self.master)
+        self.master.run_command(["ipa", "user-mod", USER2,
+                                 "--user-auth-type", "pkinit"])
+        self.master.run_command(["ipa", "config-mod",
+                                 "--user-auth-type", "pkinit"])
+        kinit_check_life(self.master, USER2)
+        reset_to_default_policy(self.master, USER2)
```
rcritten, line 209: In test_krbtpolicy_default() and test_krbtpolicy_jitter() the global auth type config isn't being modified so strictly speaking you don't need reset_policy as a fixture.

Is changing botht eh user and the global auth type necessary to check the jitter?

abbra:
LGTM. The two tests that failed in Azure are unrelated:
- test_reenroll_with_both_force_join_and_keytab seems to be related to the DNS issues (failed once in last month)
- test_chain_of_trust is due to syncrepl issues currently being investigated

abbra:
master:

* 249097c62414abc99256af9dc622c284745081e4 Update kdcpolicy design doc for jitter implementation
* 0d67180f7d2d0c6b5856db7061c44521f6a13c23 ipa-kdb: implement AS-REQ lifetime jitter

PR #5246, opened 2020-11-10
Use host keytab to obtain credentials needed for ipa-certupdate

rcritten:
This command doesn't require any permissions that a host doesn't
already have and it simplifies overall credential handling.

It also corrects the case where the server API cache is out of
date and there are no credentials available to refresh it which
can lead to a confusing error message.

Also switch to MEMORY-based ccache rather than file to avoid
having to clean up the filesystem.

https://pagure.io/freeipa/issue/8531

### DIFF
--- a/ipaclient/install/ipa_certupdate.py
+++ b/ipaclient/install/ipa_certupdate.py
@@ -21,14 +21,11 @@ from __future__ import absolute_import
 
 import logging
 import os
-import tempfile
-import shutil
 
 from urllib.parse import urlsplit
 
 from ipalib.install import certmonger, certstore
 from ipalib.facts import is_ipa_configured
-from ipalib.install.kinit import kinit_keytab
 from ipapython import admintool, certdb, ipaldap, ipautil
 from ipaplatform import services
 from ipaplatform.paths import paths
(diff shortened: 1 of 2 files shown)


frasertweedale:
Tested, works.  But CI job `fedora-latest/test_commands` is grumpy about *something*.

abbra:
Looks like test_commands suite doesn't even get the server installed at all.

rcritten:
So yeah, it failed because it wasn't installing. I had overridden the install and uninstall classmethods as pass during development and forgot to drop those bits when I pushed it :/

frasertweedale:
ACK when green.

abbra:
@rcritten please rebase to pickup the changes in gating tests

rcritten:
Rebased the test.

rcritten:
Green. Adding ack per Fraser.

abbra:
master:

* d9b259dd584eb65e2c4796968f18542a41efb42c Use host keytab to obtain credentials needed for ipa-certupdate
* d30939ebfdcd820c7e26c0709f48185cc88431fc ipatests: Test that ipa-certupdate can run without credentials

PR #5250, opened 2020-11-11
ipatests: Tests for 'NssCheck' in ipahealthcheck.ds.nss_ssl

menonsudhir:
The Server-Cert in IPA is valid for 2 year time. 
This patch changes the system year on which IPA is installed by using
the 'date -s +-2/3 Years' so that we simulate the expiry or about to expire scenario of the Server-Cert.
The modified year of the IPA server is reverted back to the actual year and ipa service is restarted 
using two different fixtures so that the system is healthy again.
This patch also checks that the 'ipahealthcheck.ds.nss_ssl' check in healthcheck tool reports the 
correct status for the "Server-Cert"  about to expire and already expired respectively.

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -1138,10 +1138,57 @@ class TestIpaHealthCheck(IntegrationTest):
                 assert "cn=config" in check["kw"]["items"]
                 assert error_msg in check["kw"]["msg"]
 
+    @pytest.fixture
+    def expire_cert_critical(self):
+        """
+        Fixture to expire the cert by moving the system date using
+        date -s command and revert it back
+        """
+        self.master.run_command(['date','-s', '+3Years'])
+        yield
+        self.master.run_command(['date','-s', '-3Years'])
+        self.master.run_command(['ipactl', 'restart'])
+
+    def test_nsscheck_cert_expired(self, expire_cert_critical):
+        """
+        This test checks that critical message is displayed
+        for NssCheck when Server-Cert has expired
+        """
+        msg = "The certificate (Server-Cert) has expired"
+        returncode, data = run_healthcheck(
+            self.master, "ipahealthcheck.ds.nss_ssl", "NssCheck",
+        )
+        assert returncode == 1
+        for check in data:
+            assert check["result"] == "CRITICAL"
+            assert check["kw"]["key"] == "DSCERTLE0002"
+            assert "Expired Certificate" in check["kw"]["items"]
+            assert check["kw"]["msg"] == msg
+
+
     def test_ipa_healthcheck_expiring(self, restart_service):
         """
         There are two overlapping tests for expiring certs, check both.
         """
+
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -1138,6 +1138,63 @@ def test_ds_configcheck_passwordstorage(self, modify_pwdstoragescheme):
...
+    def test_nsscheck_cert_expiring(self, expire_cert_error):
+        """
+        This test checks that error message is displayed
+        for NssCheck when Server-Cert is about to expire
+        """
+        msg = (
+            "The certificate (Server-Cert) will "
+            "expire in less than 30 days"
+        )
+        returncode, data = run_healthcheck(
```
flo-renaud, line 1188: If certmonger renews this cert while this method is executing, the next test (`test_ipa_healthcheck_expiring`) is likely to fail or be disturbed.
I would rather do the following:
- remove the fixture `expire_cert_error`
- move `test_nsscheck_cert_expiring` inside the test `test_ipa_healthcheck_expiring` as an internal method (`def execute_nsscheck_cert_expiring():`)
- inside `test_ipahealthcheck_expiring`, call `execute_nsscheck_cert_expiring` right after the loop `for check in ("IPACertmongerExpirationCheck", "IPACertfileExpirationCheck",):`

This way, there is only one place where the date is moved just before expiration and all the related healthcheck checks are done inside the same place.

menonsudhir: @flo-renaud have removed the expire_cert_error fixture.
moved test_nsscheck_cert_expiring inside the test test_ipa_healthcheck_expiring as an internal method.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -1197,8 +1244,10 @@ def execute_expiring_check(check):
             self.master.run_command(['date', '-s', grace_date])
 
             for check in ("IPACertmongerExpirationCheck",
-                          "IPACertfileExpirationCheck",):
+                          "IPACertfileExpirationCheck",
+                          "NssCheck"):
```
flo-renaud, line 1248: This should rather read like:
```
            for check in ("IPACertmongerExpirationCheck",
                          "IPACertfileExpirationCheck",):
                execute_expiring_check(check)

            execute_nsscheck_cert_expiring(check)
```

menonsudhir: @flo-renaud have made the said change.

flo-renaud:
@menonsudhir thanks for the update, LGTM and the temp test is [green](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/c97c3300-2a3d-11eb-a5ad-fa163e132092). You can remove the temp commit and I will ack when PRCI completes.

flo-renaud:
master:

* cd9b4efa91925978c4823f29bff49c24462f33c3 ipatests: Tests for ipahealthcheck.ds.nss_ssl

PR #5252, opened 2020-11-11
ipatests: Add debug option to healthcheck_run

menonsudhir:
Currently the run_healthcheck doesnot display the debug ouput while running, hence added the --debug option 
everytime healthcheck_run is called

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
--- a/ipatests/prci_definitions/temp_commit.yaml
+++ b/ipatests/prci_definitions/temp_commit.yaml
@@ -68,7 +68,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-latest/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
+        test_suite: test_integration/test_ipahealthcheck.py::TestIpaHealthCheck
         template: *ci-master-latest
         timeout: 3600
-        topology: *master_1repl_1client
+        topology: *master_1repl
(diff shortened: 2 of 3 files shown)


flo-renaud (commented):
Hi @menonsudhir 
I would add debug as an optional argument with default value=True:
```
def run_healthcheck(host, source=None, check=None, output_type="json",
                    failures_only=False, debug=True):
```
and provide --debug only when requested:
```
cmd = ["ipa-healthcheck"]
if debug:
    cmd.append("--debug")
```

menonsudhir:
> Hi @menonsudhir
> I would add debug as an optional argument with default value=True:
> 
> ```
> def run_healthcheck(host, source=None, check=None, output_type="json",
>                     failures_only=False, debug=True):
> ```
> 
> and provide --debug only when requested:
> 
> ```
> cmd = ["ipa-healthcheck"]
> if debug:
>     cmd.append("--debug")
> ```

menonsudhir:
Reopening the PR since it was closed by mistake.

menonsudhir:
The test failed is due to known issue: https://pagure.io/freeipa/issue/8506

menonsudhir:
@flo-renaud the test has now passed in rerun, seems to be intermittent issue.

flo-renaud:
`ipa-healthcheck` command is sometimes called directly with `self.master.run_command` and these calls don't benefit from the `--debug` option. You may want to replace them with `run_healthcheck(self.master, source=...)`. I noted for instance the following tests: test_dogtag_ca_check_exists, test_replication_check_exists, test_ipa_cert_check_exists, test_ipa_trust_check_exists, test_ipa_healthcheck_output_indent, test_ipahealthcheck_topology_with_ipactl_stop.

rcritten:
It does seem to make sense to extend run_healthcheck() to handle the other possible options: indent, severity, input-file, etc.

abbra:
There are suggestions from review that weren't implemented in past two weeks. I am turning WIP label back here.

PR #5256, opened 2020-11-12
Handle multiple AJP adapters during upgrade

cipherboy:
In this patch, we ensure we upgrade all AJP adapters with the same
secret value if any are missing. This ensures that both IPv4 and IPv6
adapters have the same secret value, so whichever httpd connects to
will be in sync. This is consistent with what Dogtag does when
provisioning them.

Notably missing from this patch is handling of multiple unrelated AJP
adapters. In an IPA scenario (and default PKI scenario) this shouldn't
be necessary. However, with external load balancing, this might happen.

This patch benefits IPA in the scenario when:

 1. `pkispawn` runs on an older PKI version (pre-AJP secret, so ~8.2?)
 2. pki gets upgraded to 10.10.1 before IPA can provision a secret,
    resulting in split IPv4/IPv6 adapters -- this would only happen
    on a direct migration from 8.2 -> 8.4
 3. ipa upgrade script then runs to provision an AJP secret value for
    use with both Dogtag and IPA.

Without this patch, only the first (IPv4) adapter would have a secret
value provisioned in the above scenario.

`Signed-off-by: Alexander Scheel <ascheel@redhat.com>`

---

Is this scenario likely? Or will IPA have provisioned a secret anyway on older PKI versions? I do not know.

### DIFF
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -326,9 +326,9 @@ class DogtagInstance(service.Service):
         if len(connectors) == 0:
             return
 
-        # AJP protocol is at version 1.3. Assume there is only one as
-        # Dogtag only provisions one.
-        connector = connectors[0]
+        # Whether or not we should rewrite the tomcat server.xml file with
+        # our changes.
+        rewrite = False
 
         # Detect tomcat version and choose the right option name
         # pre-9.0.31.0 uses 'requiredSecret'
@@ -338,34 +338,66 @@ class DogtagInstance(service.Service):
         if self.__is_newer_tomcat_version('9.0.31.0'):
             secretattr = 'secret'
 
-        rewrite = True
-        if secretattr in connector.attrib:
-            # secret is already in place
-            # Perhaps, we need to synchronize it with Apache configuration
-            self.ajp_secret = connector.attrib[secretattr]
-            rewrite = False
-        else:
-            if oldattr in connector.attrib:
+        # AJP protocol is at version 1.3. With IPv4/IPv6 split, there might
+        # be multiple AJP adapters; update them all.
+        #
+        # First, iterate through all adapters and see if any of them have a
+        # secret value set.
+        for connector in connectors:
+            if secretattr in connector.attrib or oldattr in connector.attrib:
+                # secret is already in place
+                #
+                # Perhaps, we need to synchronize it with Apache configuration
+                # or other AJP connector entries. Save it so we know we've
+                # found at least one. Because in our next loop we update the
+                # config value if incorrect, it is safe to overwrite
+                # self.ajp_adapter -- in the worst case, we'll create an
+                # entirely new value if this element happened to have an
+                # empty secret value. Plus, IPA is in charge of managing the
+                # value for the httpd side of the AJP connection as well
+                # which needs to happen after this call.
+                #
+                # The first secret found wins.
+                self.ajp_secret = connector.attrib.get(secretattr) or \
+                    connector.attrib.get(oldattr)
+                break
+
+        # If no secret value was detected, create a single unique value.
+        if not self.ajp_secret:
+            # Generate password, don't use special chars to not break XML.
+            self.ajp_secret = ipautil.ipa_generate_password(special=None)
+
+        # Finally, iterate through them all again, upgrading adapter attribute
+        # and setting the secret value if missing or incorrect.
+        for connector in connectors:
+            if oldattr != secretattr and oldattr in connector.attrib:
                 # Sufficiently new Dogtag versions (10.9.0-a2) handle the
                 # upgrade for us; we need only to ensure that we're not both
                 # attempting to upgrade server.xml at the same time.
                 # Hopefully this is guaranteed for us.
-                self.ajp_secret = connector.attrib[oldattr]
                 connector.attrib[secretattr] = self.ajp_secret
                 del connector.attrib[oldattr]
-            else:
-                # Generate password, don't use special chars to not break XML.
+                rewrite = True
+            if (secretattr not in connector.attrib
+                    or connector.attrib[secretattr] != self.ajp_secret):
+                # We hit this either when:
                 #
-                # If we hit this case, pkispawn was run on an older Dogtag
-                # version and we're stuck migrating, choosing a password
-                # ourselves. Dogtag can't generate one randomly because a
-                # Dogtag administrator might've configured AJP and might
-                # not be using IPA.
+                #   1. pkispawn was run on an older Dogtag version, or
+                #   2. there were multiple AJP adapters with mismatched
+                #      secrets.
                 #
                 # Newer Dogtag versions will generate a random password
-                # during pkispawn.
-                self.ajp_secret = ipautil.ipa_generate_password(special=None)
+                # during pkispawn. In the former scenario, it is always
+                # safe to change the AJP secret value. In the latter
+                # scenario we should always ensure the AJP connector is
+                # the one we use use with httpd, as we don't officially
+                # support multiple AJP adapters for non-IPA uses.
+                #
+                # In new Dogtag versions, Dogtag deploys separate IPv4 and
+                # IPv6 localhost adapters, which we should ensure have the
+                # same AJP secret for httpd's use.
                 connector.attrib[secretattr] = self.ajp_secret
+                rewrite = True
 
         if rewrite:
             with open(paths.PKI_TOMCAT_SERVER_XML, "wb") as fd:
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -294,7 +294,7 @@ class DogtagInstance(service.Service):
             logger.critical("failed to uninstall %s instance %s",
                             self.subsystem, e)
 
-    def __is_newer_tomcat_version(self, default=None):
+    def _is_newer_tomcat_version(self, default=None):
         try:
             result = ipautil.run([paths.BIN_TOMCAT, "version"],
                                  capture_output=True)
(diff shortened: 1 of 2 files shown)


cipherboy:
Related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1694664

cipherboy:
@rcritten Are these failing on other PRs? It appears to be DNS Zone issues:

>     ipalib.errors.InvocationError: DNS zone zoneadd.ipa.test. already exists in DNS and is handled by server(s): master1.ipa.test.

>     ipalib.errors.InvocationError: DNS zone zoneadd.ipa.test. already exists in DNS and is handled by server(s): master1.ipa.test.

>     AssertionError: Zone example.test. is not signed (master) assert False  +  where False = wait_until_record_is_signed('172.22.0.3', 'example.test.', timeout=100)  +    where '172.22.0.3' = <ipatests.pytest_ipa.integration.host.Host master1.ipa.test (master)>.ip  +      where <ipatests.pytest_ipa.integration.host.Host master1.ipa.test (master)> = <ipatests.test_integration.test_dnssec.TestInstallDNSSECFirst object at 0x7f8a72d97e20>.master

>     ipalib.errors.NotFound: Nameserver 'ns2.zone3.test.' does not have a corresponding A/AAAA record

-- not sure how I would've caused those with this PR.

rcritten:
@cipherboy yes, its a yet-to-be-determined issue in Azure. You can re-run the failed tests by clicking on AzurePipeline.freeipa and selecting re-run.

rcritten:
It seems complex enough to have a unit test to verify that the right secret is stored in the various new conditions this adds. The test could go into ipatests/test_ipaserver. I suspect mock could be used to override the value of paths.PKI_TOMCAT_SERVER_XML to pull scenarios from that data subdirectory.

cipherboy:
@rcritten Could someone more familiar with the IPA project take over this PR then? I do not have the cycles to learn IPA's unit testing framework for this PR.

rcritten:
Can you provide the sample files to test against? Assuming the xml loader will handle it they can consist just of the necessary bits to validate the fix.

cipherboy:
These test cases assume Tomcat >= 9.0.31.0:

1.a. Before -- single adapter gets upgrade:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Server port="1234" shutdown="SHUTDOWN">
  <Service name="Catalina">
    <Connector port="9000" protocol="AJP/1.3" redirectPort="443" address="localhost" requiredSecret="testing_ajp_secret" />
  </Service>
</Server>
```
1.b. After:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Server port="1234" shutdown="SHUTDOWN">
  <Service name="Catalina">
    <Connector port="9000" protocol="AJP/1.3" redirectPort="443" address="localhost" secret="testing_ajp_secret" />
  </Service>
</Server>
```

2.a. Before -- no change to existing valid adapter (same as 1.b. After)
2.b. After (same as 2.a. Before)


3.a. Before -- two adapters with different values get upgraded to the same value:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Server port="1234" shutdown="SHUTDOWN">
  <Service name="Catalina">
    <Connector port="9000" protocol="AJP/1.3" redirectPort="443" address="localhost4" requiredSecret="testing_ajp_secret" />
    <Connector port="9000" protocol="AJP/1.3" redirectPort="443" address="localhost6" requiredSecret="other_secret" />
  </Service>
</Server>
```
3.b. After:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Server port="1234" shutdown="SHUTDOWN">
  <Service name="Catalina">
    <Connector port="9000" protocol="AJP/1.3" redirectPort="443" address="localhost4" secret="$SOME_VALUE" />
    <Connector port="9000" protocol="AJP/1.3" redirectPort="443" address="localhost6" secret="$SOME_VALUE" />
  </Service>
</Server>
```
(Note here, I don't define `$SOME_VALUE`, as it doesn't really matter if first or second is chosen -- but test should check that it is the same across adapters).

4.a. Before -- two values with valid adapters don't change (same as 3.b After above)
4.a. After (same as 4.a Before above)

5.a. Before -- tow adapters missing secrets get them:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Server port="1234" shutdown="SHUTDOWN">
  <Service name="Catalina">
    <Connector port="9000" protocol="AJP/1.3" redirectPort="443" address="localhost4" />
    <Connector port="9000" protocol="AJP/1.3" redirectPort="443" address="localhost6" />
  </Service>
</Server>
```
5.b. After:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Server port="1234" shutdown="SHUTDOWN">
  <Service name="Catalina">
    <Connector port="9000" protocol="AJP/1.3" redirectPort="443" address="localhost4" secret="$SOME_VALUE" />
    <Connector port="9000" protocol="AJP/1.3" redirectPort="443" address="localhost6" secret="$SOME_VALUE" />
  </Service>
</Server>
```
(Note here, I don't define `$SOME_VALUE`, as it doesn't really matter if first or second is chosen -- but test should check that it is the same across adapters).

HTH,

Alex

ipaserver/install/dogtaginstance.py
```diff
@@ -320,9 +320,9 @@ def secure_ajp_connector(self):
         if len(connectors) == 0:
             return
 
-        # AJP protocol is at version 1.3. Assume there is only one as
-        # Dogtag only provisions one.
-        connector = connectors[0]
+        # Wether or not we should rewrite the tomcat server.xml file with
```
rcritten, line 323: nit: Wether -> Whether

ipaserver/install/dogtaginstance.py
```diff
@@ -332,13 +332,33 @@ def secure_ajp_connector(self):
...
+                #
+                # Perhaps, we need to synchronize it with Apache configuration
+                # or other AJP connector entries. Save it so we know we've
+                # found at least one. Because in our next loop we update the
+                # config value if incorrect, it is safe to overwrite
+                # self.ajp_adapter -- in the worst case, we'll create an
+                # entirely new value if this element happened to have an
+                # empty secret value. Plus, IPA is in charge of managing the
+                # value for the httpd side of the AJP connection as well.
+                self.ajp_secret = connector.attrib[secretattr]
```
rcritten, line 352: For test case 1a where requiredSecret="testing_ajp_secret". Because __is_newer() is True then secretattr  = 'secret'. The conditional on line 341 looks for both the old and new attr in connector.attrib. Since requiredSecret is there this is satisfied but since 'secret' is not there this blows on up line 352 with a KeyError.

rcritten (changes requested):
This seems more a code issue than a test case issue.

rcritten:
FYI I'm also planning on changing __is_newer_tomcat_version() to _is_newer_tomcat_version() so it can be overridden in a subclass so I can mock it out. I have an outline of a test, I just need to validate the written update.

rcritten:
@cipherboy I think I have the broad test worked out but this has the real chance of changing or setting a secret. Should there be an equivalent update to httpd.conf? Otherwise a working, but potentially insecure, configuration would suddenly stop working. From what I can tell there is no error that states that the secret doesn't work. In fact there is apparently no logging on the CA side at all.

ipaserver/install/dogtaginstance.py
cipherboy: Ah we probably want an `if`/`elseif` here instead, gotcha.

rcritten:
I'm fine if merging my code changes into the original patch make sense.

rcritten:
New test passing in 
https://dev.azure.com/freeipa/freeipa/_build/results?buildId=7652&view=logs&jobId=9bf0c493-406d-5b8e-d189-f64f53270fd1&j=c3b3beac-3142-545b-181b-8ae72407437a&t=4475fbc0-1119-5fa5-c1d0-2f0cb6cb4700

rcritten:
dbus uninstall failure, re-running.

ipaserver/install/dogtaginstance.py
cipherboy:
```suggestion
        # Whether or not we should rewrite the tomcat server.xml file with
```

cipherboy: @rcritten Feel free to commit suggestion if you want to fix this prior to push.

ipaserver/install/dogtaginstance.py
```diff
@@ -326,46 +326,78 @@ def secure_ajp_connector(self):
...
-                # Dogtag administrator might've configured AJP and might
-                # not be using IPA.
+                #   1. pkispawn was run on an older Dogtag version, or
+                #   2. there were multiple AJP adapters with mismatched
+                #      secrets.
                 #
                 # Newer Dogtag versions will generate a random password
-                # during pkispawn.
-                self.ajp_secret = ipautil.ipa_generate_password(special=None)
+                # during pkispawn. In the former safe, it is always safe to
```
cipherboy, line 390:
```suggestion
                # during pkispawn. In the former scenario, it is always safe to
```

cipherboy (commented):
I saw one other nit in the comments. Otherwise, I think the tests look good. Thanks @rcritten!

ipaserver/install/dogtaginstance.py
rcritten: Fixed

rcritten:
Ok if this passes CI I'm going to merge the last two patches into your original patch (so there will be three patches: code changes, make __is_newer_tomcat_version overridable and the test. I kept my other changes separate to make it easier for you to review.

rcritten:
Ok, the test pass. I squashed my non-test changes into the original patch. Adding needs-review to get another set of IPA eyes on this.

flo-renaud:
Hi @rcritten 
thanks for the patch, LGTM. There is one test scenario that is not covered, but I'm not sure it can happen in real life: upgrade 2 connectors with new "secret" attribute and different secrets. I would expect to force the use of the same secret in this case and I think the code properly handles it.

rcritten:
I rebased the code to see if the test failure is resolved. It was another uninstall DBus-related failure.

flo-renaud:
master:

* 8a715feb57a1f43c6a72c59f69c30b03b398a7d9 Handle multiple AJP adapters during upgrade
* fd66a74a5910f03c6c05a32d76dd8a196cd5314d Allow overriding is_newer_tomcat_version()
* 7ee30aa0920498ecb504de0e9b1d673ac7265ae4 ipatests: Test secure_ajp_connector works with multiple connectors

PR #5257, opened 2020-11-13
[EPN] SMTP client enhancements

stanislavlevin:
- don't downgrade security
- enable SMTP server certificate validation and hostname checking
- allow authentication by SMTP client's certificate

https://pagure.io/freeipa/issue/8578
https://pagure.io/freeipa/issue/8579
https://pagure.io/freeipa/issue/8580

PR #5262, opened 2020-11-16
Fix TLS/SSL related issues in EPN

tiran:
Prevent downgrade attack when admin requests STARTTLS but server does
not return STARTTLS in EHLO message. The SMTP.starttls() methods checks
for has_extn("starttls") and fails with SMTPNotSupportedError when the
extension is not available. SMTPNotSupportedError is a subclass of
SMTPException.

Verify cert chain and hostname in starttls() and SMTP_SSL. The smtplib
module defaults to unverified connection.

Fixes: https://pagure.io/freeipa/issue/8578
Fixes: https://pagure.io/freeipa/issue/8579

### DIFF
--- a/ipaclient/install/ipa_epn.py
+++ b/ipaclient/install/ipa_epn.py
@@ -29,6 +29,7 @@ import os
 import pwd
 import logging
 import smtplib
+import ssl
 import time
 
 from collections import deque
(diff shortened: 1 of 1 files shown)


abbra:
This fix is already included in more comprehensive https://github.com/freeipa/freeipa/pull/5257, shouldn't we merge that instead?

abbra:
I think we can close this PR since #5257 already merged

PR #5264, opened 2020-11-16
[ipatests] Raise log level of 389-ds replication

stanislavlevin:
- change log level for replication debugging
  According to the docs:
  ```
  default level of logging(16384) used for critical errors and other messages that are always
  written to the error log. Messages at this level are always included in the error log, regardless
  of the log level setting.
   ```
    
- always flush the access logs to filesystem
  During the testing access logs may be written with a significant delay, this results in logs are not collected by this test node, but for example,  the next one.
    
- as of now, the changes on `cn=config` are made after the installation of a server or replica. If an error occurs during these stages, then the actual log level will be the default and not as expected.

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -148,6 +148,29 @@ def apply_common_fixes(host):
     rpcbind_kadmin_workaround(host)
 
 
+def prepare_dse_changes(host, log_level=8192):
+    """Put custom changes for dse.ldif on the host
+    """
+    ipatests_dse_path = os.path.join(host.config.test_dir, "ipatests_dse.ldif")
+    ldif = textwrap.dedent(
+        """\
+        # replication debugging
+        dn: cn=config
+        changetype: modify
+        replace: nsslapd-errorlog-level
+        nsslapd-errorlog-level: {log_level}
+
+        # server writes all access log entries directly to disk
+        dn: cn=config
+        changetype: modify
+        replace: nsslapd-accesslog-logbuffering
+        nsslapd-accesslog-logbuffering: off
+        """
+    ).format(log_level=log_level)
+    host.put_file_contents(ipatests_dse_path, ldif)
+    return ipatests_dse_path
+
+
 def allow_sync_ptr(host):
     kinit_admin(host)
     host.run_command(["ipa", "dnsconfig-mod", "--allow-sync-ptr=true"],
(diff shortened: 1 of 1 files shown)


rcritten:
From the commit message, I think what you're saying is to set the log level to replication debugging (8192) and it isn't necessary to set it to 24576 (8192 + 16384) because critical errors are always logged?

rcritten:
The patch looks ok, I just wonder what the original motivation for it is. Certainly having replication logging is good I just wonder if some error caused you to add this or you just saw it as a good idea.

stanislavlevin:
> From the commit message, I think what you're saying is to set the log level to replication debugging (8192) and it isn't necessary to set it to 24576 (8192 + 16384) because critical errors are always logged?

Yes, 8192 is enough to see general + replication errors.

- in past, I had issues with replication during the installation of replica and at that moment I rebuilt 389-ds with default 8192 log level. There is a bunch of recent issues(test_dnssec, test_dns_locations) related to replication. I hope this helps to see what happens.
- 389-ds log entries are not always contained in that CI artifact I expected by the time of execution but in the next one.

rcritten:
This seems ok to me but it is going to use disk quite a bit faster. I spot-checked the simple replication test and looked at the master log in test_integration-test_simple_replication.py-TestSimpleReplication-test_user_replication_to_master. The errors log was 111288 B compressed vs 14172 previously. So not immense but given the number of installations this may add up to quite a bit. I don't know what our storage capacity is.

rcritten:
@netoarmando What to you think about this change? Would it push our storage limits?

netoarmando:
> @netoarmando What to you think about this change? Would it push our storage limits?

I think we are okay, runners are cleaned frequently, that new limit won't hurt. 
Also, we keep a long history of artifacts, eventually we need to delete old stuff, we checked the usage recently and the conclusion was that we still don't need to worry about it. In conclusion, the impact here is minimum.

@rcritten I wonder if there is a precedent for using DEBUG as default log level elsewhere in the code base.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -148,6 +148,29 @@ def apply_common_fixes(host):
     rpcbind_kadmin_workaround(host)
 
 
+def prepare_dse_changes(host):
```
netoarmando, line 151: Any problem to keep `log_level` as an argument like previous function? 

It would be easy to spot the following and think that we can change the log level in this function. 
```python
def prepare_dse_changes(host, log_level=8192):
    ...
```

stanislavlevin:
@netoarmando, thank you, added `log_level`.

abbra:
LGTM.

stanislavlevin:
Please, wait. There is the temp commit.

stanislavlevin:
- rebased
- removed temp commit

flo-renaud:
master:

* 9fb222467b08edc69b382733443e52a33e8c650e ipatests: Raise log level of 389-ds replication

PR #5270, opened 2020-11-17
Merge fixes from rawhide experiments

abbra:
- 11f8ec48c spec: use pkgconf to find out krb5 version
- eecf4dbb3 Drop upper bound on krb5 version in freeipa.spec
- aaf0232f0 Azure CI: use PPA to provide newer libseccomp version
- dde3528ee Azure CI: use Ubuntu-20.04 image by default
- 70175f61b ipa-acme-manage: user a cookie created for the communication with dogtag REST endpoints
- ed3258563 ipa-otpd: fix gcc complaints in Rawhide
- 95bd6402e ipa-sam: fix gcc complaints on Rawhide
- 300961a13 ipa-kdb: fix gcc complaints in kdb tests
- bb2369329 ipa-kdb: fix gcc complaints

PR #5258, opened 2020-11-13
WIP: test against Rawhide

abbra:
We target FreeIPA 4.9.0 to Fedora 34 (Rawhide now).

Rawhide version of 389-ds-base has all required fixes for https://pagure.io/freeipa/issue/8538

Let's test that and also fix Rawhide issues we see.

PR #5271, opened 2020-11-17
Pass the cookie obtained on login to the dogtag REST API

rcritten:
https://pagure.io/freeipa/issue/8584


rcritten:
Duplicate of https://github.com/freeipa/freeipa/pull/5258/commits/a249c4d14a1c4a9fe87bb1e3436cb3d065de606d

Difference is in the local function defining **kw vs headers. I'm ok with with either version.

abbra:
Let's then save on the CI runs? ;) PR #5270 is needed anyway.

PR #5281, opened 2020-11-19
Fix scope on pytest fixture in krbtpolicy tests

frozencemetery:
0d67180f7d2d0c6b5856db7061c44521f6a13c23 introduced the with_admin
fixture using class scope, which caused test failures as pytest
instantiated it before the multihost fixture.

Resolves: https://pagure.io/freeipa/issue/8590

### DIFF
--- a/ipatests/test_integration/test_krbtpolicy.py
+++ b/ipatests/test_integration/test_krbtpolicy.py
@@ -22,7 +22,7 @@ USER2 = "testuser2"
 MAXLIFE = 86400
 
 
-def maxlife_within_policy(input, maxlife, slush=5):
+def maxlife_within_policy(input, maxlife, slush=3600):
     """Given klist output of the TGT verify that it is within policy
 
        Ensure that the validity period is somewhere within the
@@ -32,6 +32,8 @@ def maxlife_within_policy(input, maxlife, slush=5):
 
        Input should be a string like:
        11/19/2019 16:37:40  11/20/2019 16:37:39  krbtgt/...
+
+       slush defaults to 1 * 60 * 60 matching the jitter window.
     """
     data = input.split()
     start = datetime.strptime(data[0] + ' ' + data[1], '%m/%d/%Y %H:%M:%S')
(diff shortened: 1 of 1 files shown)


rcritten:
Not all tests are executed as part of gating, some are only run in the nightlies. To invoke it:

* edit ```ipatests/prci_definition/temp_commit.yaml``` and add the test case you want to run (in this case ```test_krbtpolicy.py``` replacing ```test_REPLACEME.py```)
* run:  ```ln -sf ipatests/prci_definitions/temp_commit.yaml .freeipa-pr-ci.yaml```
* ```git add ipatests/prci_definitions/temp_commit.yaml .freeipa-pr-ci.yaml```
* ```git commit  -a -m "Temp commit"```

It is usually a good idea to look in ```ipatests/prci_definitions``` to find where your test is executed to see if it needs an extended timeout. In this case it uses the default 3600 so you're fine.

PRCI extracts the list of tests from .freeipa-pr-ci.yaml, and will execute only the ones specified (Azure will still run its tests).

rcritten:
I haven't been able to trace anything down yet but running this on my local machines hangs after the uninstall.

rcritten:
I think the problem is jitter and the slush amount. In some cases the slush (allowed difference between policy max life and issued max life) is set and in others defaults to 5. With jitter always enabled  this needs to be tweaked.

IIRC the jitter can vary the issuance by up to an hour so that needs to be accounted for.

rcritten:
Added candidate patch.

rcritten:
I don't know if the pkinit failure is related to the jitter changes or not.

frozencemetery:
The investigation is very appreciated.  I've squished (what I hope are) all the fixes into a single commit, but please let me know if you'd prefer the separation (or feel free to chop it up any other way, for that matter).  There's a strong argument that having your name on the fix is more honest anyway :)

(Historical note: the tests were copied from the original PR.  Due to lack of familiarity with the gating/PRCI system as well as other things being broken at the time, we incorrectly assumed they were running and working.)

frozencemetery:
Removing temp commit as requested.  Passing PRCI run is http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/7da8607e-2ec6-11eb-a5ad-fa163e132092/

rcritten:
LGTM. I asked Robbie to remove the temp commit on IRC.

rcritten:
master:

* 9e0084ba981a274d3a7df82ca020a7c9f8c3e106 Fix krbtpolicy tests

PR #5289, opened 2020-11-24
ipaserver: Check memory requirement also for CA

t-woerner:
There is no --setup-ca option for the server installer but options.setup_ca
has been used as an option for installutils.check_available_memory.
Therefore it always was False. options.setup_ca has been set later on
depending on options.dirsrv_cert_files.

This means that the check always ignoring the needed memory for the CA.

### DIFF
--- a/ipaserver/install/server/install.py
+++ b/ipaserver/install/server/install.py
@@ -346,6 +346,13 @@ def install_check(installer):
     dirsrv_ca_cert = None
     pkinit_ca_cert = None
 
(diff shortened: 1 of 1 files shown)


rcritten:
In my testing it is actually True as the default. Where this is a problem is installing CA-less as the additional memory requirement will still apply.

rcritten (changes requested):
I'm ok with the change, please just update the commit message that the default is True and needs to be False in the case of CA-less.

I don't know if this is worthy of a ticket but if we want it included into downstream it may need one.

PR #5292, opened 2020-11-25
Always define the path DNSSEC_OPENSSL_CONF

flo-renaud:
The variable was None by default and set to /etc/ipa/dnssec/openssl.cnf
for fedora only because the code is specific to the support of pkcs11
engine for bind. As a consequence ipa-backup had a "None" value in the
list of files to backup and failed on Exception.

ipa-backup code is able to handle missing files, and the code using
the pkcs11 engine is called only when NAMED_OPENSSL_ENGINE is set
(only in fedora so far). It is safe to always define a value for
DNSSEC_OPENSSL_CONF even on os where it does not exist.

Fixes: https://pagure.io/freeipa/issue/8597

### DIFF
--- a/ipaplatform/base/paths.py
+++ b/ipaplatform/base/paths.py
@@ -70,7 +70,7 @@ class BasePathNamespace:
     IPA_DEFAULT_CONF = "/etc/ipa/default.conf"
     IPA_DNSKEYSYNCD_KEYTAB = "/etc/ipa/dnssec/ipa-dnskeysyncd.keytab"
     IPA_ODS_EXPORTER_KEYTAB = "/etc/ipa/dnssec/ipa-ods-exporter.keytab"
-    DNSSEC_OPENSSL_CONF = None
+    DNSSEC_OPENSSL_CONF = "/etc/ipa/dnssec/openssl.cnf"
     DNSSEC_SOFTHSM2_CONF = "/etc/ipa/dnssec/softhsm2.conf"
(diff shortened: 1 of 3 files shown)


flo-renaud:
Only ipa-4-9 label is needed as the code was committed to the master branch only, and ipa-4-9 inherited from the issue as it branched after the commit.

stanislavlevin:
Hi, @flo-renaud. Thank's for the fix.
I think it would be better to handle `None`, as paths can be so.
For example, with:
```
list(filter(None, files))
```
and similar for dirs and logs.

abbra:
I think we need both. ;) For example, Fedora ELN (Rawhide right now = RHEL 9 base) already has this and would need to have support for openssl engine. Also, RHEL 8.4 gets all the same bits, including bind-dyndb-ldap, so it can do the same if we want.

flo-renaud:
@stanislavlevin thanks for your suggestion. I added a check for non-None item in verify_directories, that should protect us from other constants set to None. Also added a temp commit to launch ipa-backup tests.

abbra:
LGTM. Please remove the temporary commit.

flo-renaud:
@abbra Temp commit removed and tests are green.

flo-renaud:
master:

* 3b007b7bba185989bec34bb071742270d3362e2e Always define the path DNSSEC_OPENSSL_CONF

PR #5303, opened 2020-11-28
ipatests: fix TestTrust::test_subordinate_suffix

flo-renaud:
The test test_subordinate_suffix is failing when configuring the DNS
for the trust, because the dnsforwardzone already exists. It was
configured during the previous test for nonposix trust.

At the end of the tests for nonposix trust, unconfigure the DNS
and the trust before calling the subordinate_suffix test, and add
a test cleaning up subordinate_suffix test.

Fixes: https://pagure.io/freeipa/issue/8601

### DIFF
--- a/ipatests/test_integration/test_trust.py
+++ b/ipatests/test_integration/test_trust.py
@@ -245,6 +245,11 @@ class TestTrust(BaseTestTrust):
         self.master.run_command(['kinit', '-C', '-E', self.upn_principal],
                                 stdin_text=self.upn_password)
 
+    def test_remove_nonposix_trust(self):
+        self.remove_trust(self.ad)
+        tasks.unconfigure_dns_for_trust(self.master, self.ad)
+
+    # Test with AD trust defining subordinate suffixes
     def test_subordinate_suffix(self):
         """Test subordinate UPN Suffixes"""
         tasks.configure_dns_for_trust(self.master, self.ad)
@@ -290,7 +295,7 @@ class TestTrust(BaseTestTrust):
         )
         tasks.kdestroy_all(self.master)
 
-    def test_remove_nonposix_trust(self):
+    def test_remove_subordinate_suffixes_trust(self):
         self.remove_trust(self.ad)
         tasks.unconfigure_dns_for_trust(self.master, self.ad)
 


ipatests/test_integration/test_trust.py
```diff
@@ -245,6 +245,11 @@ def test_upn_user_authentication_in_nonposix_trust(self):
         self.master.run_command(['kinit', '-C', '-E', self.upn_principal],
                                 stdin_text=self.upn_password)
 
+    def test_remove_nonposix_trust(self):
```
mrizwan93, line 248: Can we move these line to fixture and use it in test method?
Else IMO, it (i.e cleanup) should be added as part of test at the end of test method.

flo-renaud: In general I agree with your comment and would also prefer to have a test fixture. But in this specific case, the code is creating the trust in test_establish_noposix_trust, then makes a bunch of other tests in separate test methods (test_trustdomains_found_in_nonposix_trust, test_range_properties_in_nonposix_trust ...) The trust removal can be done only after the last test related to nonposix.

mrizwan93: Can't we use the fixture with last test? and some comments about not to disturb the sequence of test or soemething like.

menonsudhir: @flo-renaud Adding test_remove_nonposix_trust and test_remove_subordinate_suffixes_trust is correct, the trust removal is done here prior to adding a different type of trust.
@mrizwan93 @flo-renaud rest i believe we can skip the part for adding the test fixture.

mrizwan93: @menonsudhir The point I am making here is, IMO having 2 `test_` method for exact same thing doesn't make sense. It's behaving like test cases!

flo-renaud: @mrizwan93 the whole test `test_trust.py` is written with this mindset, you can see many occurrences of `tasks.unconfig_dns_for_trust()` in methods that are in fact cleaners for the previous test methods. You are welcome to re-organize the whole file if you think it's worth it, but my only goal in this commit is to make the test suite behave properly and pass in the nightlies.

abbra:
LGTM. Please remove temp commit

flo-renaud:
Temp commit removed, tests are green.

abbra:
master:

* 97aa70cd7a835e5f061e773ebb446333de3d5f7c ipatests: fix TestTrust::test_subordinate_suffix

PR #5306, opened 2020-11-30
ipatests: Enable httpd_can_network_connect for mod_md test

rcritten:
This is required because the test runs a local Apache instance.

https://pagure.io/freeipa/issue/8514

### DIFF
--- a/ipatests/test_integration/test_acme.py
+++ b/ipatests/test_integration/test_acme.py
@@ -50,6 +50,23 @@ def check_acme_status(host, exp_status, timeout=60):
     return status
 
 
+def get_selinux_status(host):
+    """
+    Return the SELinux enforcing status.
+
+    Return True if enabled and enforcing, otherwise False
+    """
+    result = host.run_command(['/usr/sbin/selinuxenabled'], raiseonerr=False)
+    if result.returncode != 0:
+        return False
+
+    result = host.run_command(['/usr/sbin/getenforce'], raiseonerr=False)
+    if 'Enforcing' in result.stdout_text:
+        return True
+
+    return False
+
+
 def server_install_teardown(func):
     def wrapped(*args):
         master = args[0].master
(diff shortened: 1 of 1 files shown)


flo-renaud:
@rcritten The fix seems good but could you run a temp commit with `selinux_enforcing: True` in the job definition?
```
 fedora-latest/test_acme:
    requires: [fedora-latest/build]
    priority: 50
    job:
      class: RunPytest
      args:
        selinux_enforcing: True
        test_suite: test_integration/test_acme.py
        template: *ci-master-latest
        timeout: 7200
```

rcritten:
It's failing writing to the configured md storage directory.

type=AVC msg=audit(1606919627.665:1864): avc:  denied  { add_name } for  pid=20428 comm="httpd" name="md_store.json" scontext=system_u:system_r:httpd_t:s0 tcontext=unconfined_u:object_r:httpd_config_t:s0 tclass=dir permissive=0

At this point I'm not sure if it's best to just drop that directory and see if things pass or work on a policy for it (restorecon may be sufficient).

rcritten:
Fixing this would require us to import and maintain the upstream SELinux policy. This is more effort than we need to expend. We aren't supporting mod_md directly, just testing that multiple clients work. This could be a blackhole of work.

Instead I'll skip the test in SELinux mode.

rcritten:
test_mod_md is properly skipped:

('/usr/lib/python3.8/site-packages/ipatests/test_integration/test_acme.py', 313, 'Skipped: SELinux is enabled, this will fail')

abbra:
LGTM

rcritten:
Adding ack per @abbra LGTM

rcritten:
master:

* df4380c11fa7d8eb08a43bca41fad741c116ad65 Skip the ACME mod_md test when the client is in enforcing mode

PR #5199, opened 2020-10-15
Change KRA profiles in certmonger tracking so they can renew

rcritten:
Change KRA profiles in certmonger tracking so they can renew

Internal profiles were assigned which prevented rewewals.

dogtag is providing a new profile for the audit signing cert,
caAuditSigningCert.

There are existing profiles for the transport (caTransportCert)
and storage (caStorageCert) certificates.

https://pagure.io/freeipa/issue/8545

Signed-off-by: Rob Crittenden <rcritten@redhat.com>

**NOTE**: This is WIP because the necessary profile is only in the pki nightly repo. We want this backported to other supported IPA branches but they may be delayed depending on when pki builds are available.

PR #5311, opened 2020-12-01
[Backport][ipa-4-6] Change KRA profiles in certmonger tracking so they can renew

rcritten:
This PR was opened manually because PR #5199 was pushed to master and backport to ipa-4-6 is required.

This also required backporting fixes from ticket https://pagure.io/freeipa/issue/7991

I'm not sure this is going to pass as I don't believe the required profile is available in downstream and a backport seems unlikely given the Fedora release is EOL.

### DIFF
--- a/install/certmonger/dogtag-ipa-ca-renew-agent-submit
+++ b/install/certmonger/dogtag-ipa-ca-renew-agent-submit
@@ -211,9 +211,9 @@ def request_cert(reuse_existing, **kwargs):
              "--certfile", paths.RA_AGENT_PEM,
              "--keyfile", paths.RA_AGENT_KEY] +
             sys.argv[1:] +
-            ['--submit-option', "requestor_name=IPA"])
-    if os.environ.get('CERTMONGER_CA_PROFILE') == 'caCACert':
-        args += ['--force-new', '--approval-option', 'bypassCAnotafter=true']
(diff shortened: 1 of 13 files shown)


flo-renaud:
@rcritten I'm confused by your comment:

> This also required backporting fixes from ticket https://pagure.io/freeipa/issue/7991

because I thought the renewal failure was a regression introduced by the fix for 7991. Let me check if the issue is present on ipa-4-6 branch first (this backport may not be needed at all).

flo-renaud:
I couldn't reproduce the issue on RHEL 7.9, I think we can simply close this backport and the BZ that was erroneously cloned to RHEL 7: https://bugzilla.redhat.com/show_bug.cgi?id=1872604.

rcritten:
This is confusing to me. What profile is being used to renew the certificates then? Are the renewed certificates consistent with the originally-issued certificates?

rcritten:
Discussed this out-of-band with @flo-renaud. We can close this because it doesn't apply.

In order to do this backport I had to also pull in https://pagure.io/freeipa/issue/7991. That PR specified to certmonger with profile to use for issuing certificates and was the root cause of the KRA renewal failures. It was using internal CA profiles.

Prior to this certmonger would renew by serial number using the existing CSR. dogtag will re-issue a new cert using that CSR so there is no need to use a profile.

Closing.

PR #5312, opened 2020-12-01
Move where the restore state is marked during IPA server upgrade

rcritten:
There is still some exposure to killing in a bad place. This was
reproduced by killing the process in the parser.parse() call within
__restore_config (line 230) so the values were restored from the
backup but the new dse.ldif never written or copied. But the values
had already been restored from the state file.

I'm not sure this can ever be 100% bullet-proof since it can be
externally killed but if rather than calling restore_state() on the
values in __restore_config we use get_state() which will peek at the
values in the state file without removing them. Then the last step
is to pop upgrade-in-progress and then the rest.

If the values have been restored and the new ldif written and copied
then it's only upgrade-in-progress that really matters. The rest will
be overwritten.

https://pagure.io/freeipa/issue/7534

### DIFF
--- a/ipaserver/install/upgradeinstance.py
+++ b/ipaserver/install/upgradeinstance.py
@@ -199,11 +199,11 @@ class IPAUpgrade(service.Service):
         shutil.copy2(ldif_outfile, self.filename)
 
     def __restore_config(self):
-        port = self.restore_state('nsslapd-port')
-        security = self.restore_state('nsslapd-security')
-        global_lock = self.restore_state('nsslapd-global-backend-lock')
-        schema_compat_enabled = self.restore_state('schema_compat_enabled')
-        self.restore_state('upgrade-in-progress')
+        # peek the values during the restoration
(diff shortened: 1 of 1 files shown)


rcritten:
This is basically a manual reproducer, requiring hitting ^C multiple times quickly after step 5 stopping directory server. You have to spam ^C so that the outer try/except won't call __restore_config()

If you manage to kill it without it calling restore you'll have an upgrade-in-progress value in /var/lib/ipa/sysrestore/sysrestore.state in [dirsrv].

If you allow the next upgrade to run it should skip backing the values up again and properly restore things at the end. Or you can kill things again and the state values should be retained (e.g. port remains 389 in the state file and doesn't become 0 or removed).

abbra:
LGTM

abbra:
master:

* ba9b9502e4f4e99d0f11a8cdc40e8e57cb774848 Move where the restore state is marked during IPA server upgrade

PR #5323, opened 2020-12-03
Debian: some path/service name fixes

tjaalton:
The bind thing I'm not fully sure about, since the code doesn't need to support 9.11 anymore, but I've no idea what would happen on upgrade.


flo-renaud:
Adding re-run as @tjaalton is not a member of the allowed list for PRCI.

abbra:
@tjaalton please rebase so that we can re-run full test suite.

abbra:
master:

* 7dbe6e4fe3e0863bf10870590fbabc73cc5b1aa1 Debian: Fix paths and service names for bind 9.16
* 483315868b8e36e312ecd06f092dbceba2a292d3 Debian: Fix chrony service name
* 5783a8822c16a2bc99f1b6629e9d81a9ec2e6f26 ipaplatform: Use gpg instead of gpg2

PR #5328, opened 2020-12-09
Allow mod_auth_gssapi to create and access ccaches in /run/ipa/ccaches

abbra:
With commit c6644b8566f747fa80e2c1925b79bad9f8c92bd7 we default to
create unique credential caches in /run/ipa/ccaches for every client
that connects to IPA with a new session. On F34, mod_auth_gssapi process
running as 'apache' cannot create the ccache in /run/ipa/ccaches because
it has no access rights.

The core of the problem is that we have two different paths to obtaining
a ccache: one where 'apache' running httpd process creates it directly
and one where an internal redirect from 'ipaapi' running httpd process
is happening.

Since /run/ipa/ccaches is a temporary directory created in /run which is
mounted as tmpfs, POSIX ACLs can be used to add 'apache' group
permission to create the ccaches.

When '/run/ipa/ccaches' has SGID bit, created files inside it would
automatically get 'ipaapi' group assigned. So all we need is to mark the
directory as 02770 (SGID bit).

Fixes: https://pagure.io/freeipa/issue/8613

### DIFF
--- a/init/tmpfilesd/ipa.conf.in
+++ b/init/tmpfilesd/ipa.conf.in
@@ -1,2 +1,3 @@
 d /run/ipa 0711 root root
-d /run/ipa/ccaches 0770 ipaapi ipaapi
+d /run/ipa/ccaches 6770 ipaapi ipaapi
+a+ /run/ipa/ccaches - - - - g:apache:rwx
--- a/install/share/ipa.conf.template
+++ b/install/share/ipa.conf.template
@@ -75,7 +75,7 @@ WSGIScriptReloading Off
 
   GssapiImpersonate On
   GssapiDelegCcacheDir $IPA_CCACHES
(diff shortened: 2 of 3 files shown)


abbra:
Azure CI failures are due to https://pagure.io/freeipa/issue/8616

abbra:
@flo-renaud please review

flo-renaud:
@abbra thanks for the PR, LGTM. I tested the installation on rawhide, works well.

abbra:
master:

* ee7dfc3d297feeba88643df77eb29f5087168395 Allow mod_auth_gssapi to create and access ccaches in /run/ipa/ccaches
* 07fa32a0a57a7033cc53a987d15fc365c43e75f3 upgrade: provide DOMAIN to the server upgrade dictionary

PR #5338, opened 2020-12-11
ipa-cert-fix: do not fail when CSR is missing from CS.cfg

flo-renaud:
### ipa-cert-fix: do not fail when CSR is missing from CS.cfg

When the CSR for an expired cert is not found in
/etc/pki/pki-tomcat/{ca|kra}/CS.cfg, ipa-cert-fix fails to
renew the certificate and repair the installation.

The CSR can be found using certmonger as it is stored in
/var/lib/certmonger/requests/<ID> in the "csr" attribute.
Prior to calling pki-server cert-fix, make sure that the
CSR is present in CS.cfg, or update CS.cfg with the content
found using certmonger.

Fixes: https://pagure.io/freeipa/issue/8618

### ipatests: add a test for ipa-cert-fix

Add a new test for ipa-cert-fix issue 8618. When the CSR for one
of the certs to be renewed is missing from /etc/pki/pki-tomcat/{ca|kra}/CS.cfg
ipa-cert-fix fails to renew the certificates.

Test scenario:
move the date in the future to expire PKI system certificates (+3 years)
delete the directive ca.sslserver.certreq from CS.cfg
call ipa-cert-fix and ensure that the CSR was found

Related: https://pagure.io/freeipa/issue/8618

### ipatests: add test_ipa_cert_fix to the nightly definitions

Add the new test test_integration/test_ipa_cert_fix.py to the
nightly definitions.

Related: https://pagure.io/freeipa/issue/8618

### DIFF
--- a/ipaserver/install/ipa_cert_fix.py
+++ b/ipaserver/install/ipa_cert_fix.py
@@ -25,6 +25,9 @@
 
 from __future__ import print_function, absolute_import
 
+import base64
+from cryptography import x509 as crypto_x509
+from cryptography.hazmat.backends import default_backend
 import datetime
 from enum import Enum
 import logging
@@ -33,12 +36,14 @@ import shutil
 from ipalib import api
 from ipalib import x509
 from ipalib.facts import is_ipa_configured
+from ipalib.install import certmonger
 from ipaplatform.paths import paths
 from ipapython.admintool import AdminTool
 from ipapython.certdb import NSSDatabase, EMPTY_TRUST_FLAGS
 from ipapython.dn import DN
 from ipapython.ipaldap import realm_to_serverid
 from ipaserver.install import ca, cainstance, dsinstance
+from ipapython import directivesetter
 from ipapython import ipautil
 
 msg = """
@@ -53,6 +58,16 @@ and keys, is STRONGLY RECOMMENDED.
 
 logger = logging.getLogger(__name__)
 
+
+cert_nicknames = {
+    'sslserver': 'Server-Cert cert-pki-ca',
+    'subsystem': 'subsystemCert cert-pki-ca',
+    'ca_ocsp_signing': 'ocspSigningCert cert-pki-ca',
+    'ca_audit_signing': 'auditSigningCert cert-pki-ca',
+    'kra_transport': 'transportCert cert-pki-kra',
+    'kra_storage': 'storageCert cert-pki-kra',
+    'kra_audit_signing': 'auditSigningCert cert-pki-kra',
+}
 
 class IPACertType(Enum):
     IPARA = "IPA RA"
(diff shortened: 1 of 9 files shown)


abbra:
LGTM. Please remove temp commit.
I am a bit concerned with the unguarded dictionary lookups in 
```
        # Check if the directive is set in the config file
        nickname = cert_nicknames[certid]
        (directive, cfg_path) = directives[nickname]
```
as we already saw several cases when `Server-Cert` was replaced by a specific certificate name in the wild, but I think this has to be addressed in a separate pull request.

ipaserver/install/ipa_cert_fix.py
```diff
@@ -228,6 +236,69 @@ def print_cert_info(context, desc, cert):
...
+    }
+
+    id = certmonger.get_request_id(criteria)
+    if id:
+        csr = certmonger.get_request_value(id, "csr")
+        if csr:
+            csr_obj = crypto_x509.load_pem_x509_csr(
+                csr.encode('ascii'), default_backend())
+            val = base64.b64encode(csr_obj.public_bytes(x509.Encoding.DER))
+            return val.decode('ascii')
```
rcritten, line 258: At first glance this seems like overkill, to load the cert, convert it between formats, etc, but it's a belt-and-suspenders approach. In the end, if there is a CSR in certmonger, we'll know that it is valid.

I think a big try/except would be good around this block though in case the CSR can't be loaded or one of the encode/decode fails.

ipaserver/install/ipa_cert_fix.py
```diff
@@ -228,6 +236,69 @@ def print_cert_info(context, desc, cert):
...
+    # pki-server cert-fix needs to find the CSR in the subsystem config file
+    # otherwise it will fail
+    # For each cert to be fixed, check that the CSR is present or
+    # get it from certmonger
+    for (certid, _cert) in certs:
+        # Check if the directive is set in the config file
+        nickname = cert_nicknames[certid]
+        (directive, cfg_path) = directives[nickname]
+        csr = directivesetter.get_directive(cfg_path, directive, '=')
+        if csr is None:
```
rcritten, line 293: nit: is csr needed here? Would it be simple to do ```if directivesetter.get_directive(...) is None:```

flo-renaud:
@rcritten thanks for your suggestions, added to the commit.

rcritten:
LGTM. Will add ack once tests are green.

flo-renaud:
Temp commit removed

flo-renaud:
master:

* b8ece644e849f8341544bcadba8b19fd888dd90d ipatests: add a test for ipa-cert-fix
* 1a988ba96d7563d66e47db0d6bcc13cad7a85149 ipa-cert-fix: do not fail when CSR is missing from CS.cfg
* 98711e8edf052211e5ccc3c1ac3c2367c6f23aff ipatests: add test_ipa_cert_fix to the nightly definitions

PR #5340, opened 2020-12-14
Change mkdir logic in DNSSEC

tiran:
- Create /var/named/dyndb-ldap/ipa/master/ early
- Assume that /var/named/dyndb-ldap/ipa/master/ exists in BINDMgr.sync()

### DIFF
--- a/ipaserver/dnssec/bindmgr.py
+++ b/ipaserver/dnssec/bindmgr.py
@@ -182,10 +182,9 @@ class BINDMgr:
         zone_path = os.path.join(paths.BIND_LDAP_DNS_ZONE_WORKDIR,
                 self.get_zone_dir_name(zone))
         try:
-            os.makedirs(zone_path)
-        except OSError as e:
-            if e.errno != errno.EEXIST:
-                raise e
+            os.mkdir(zone_path, 0o770)
+        except FileExistsError:
+            pass
 
         # fix HSM permissions
         # TODO: move out
--- a/ipaserver/install/dnskeysyncinstance.py
+++ b/ipaserver/install/dnskeysyncinstance.py
@@ -66,12 +66,19 @@ class DNSKeySyncInstance(service.Service):
         """
         Setting up correct permissions to allow write/read access for daemons
         """
-        if not os.path.exists(paths.BIND_LDAP_DNS_IPA_WORKDIR):
-            os.mkdir(paths.BIND_LDAP_DNS_IPA_WORKDIR, 0o770)
-        # dnssec daemons require to have access into the directory
-        os.chmod(paths.BIND_LDAP_DNS_IPA_WORKDIR, 0o770)
-        os.chown(paths.BIND_LDAP_DNS_IPA_WORKDIR, self.named_uid,
-                 self.named_gid)
+        directories = [
+            paths.BIND_LDAP_DNS_IPA_WORKDIR,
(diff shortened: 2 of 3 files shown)


abbra:
I have built https://bodhi.fedoraproject.org/updates/FEDORA-2020-0ff70f500a with these patches in Rawhide to allow OpenQA to test them at next Rawhide compose. We'll see how it fares against that tomorrow.

abbra:
Quick test by Adam did show that the bug is still there and these changes did not affect it. We are going to look into auditd rules to catch changes to `/var/named/dyndb-ldap`.

abbra:
I found actual issue: bind-dyndb-ldap package does require 'named' group for own directories' permissions. It means bind package has to be installed before bind-dyndb-ldap. I fixed this in rawhide today, so this PR is not needed for that particular problem.
We can merge it for clearing up the code, of course.
For that reason, LGTM.

stanislavlevin:
excuse me, which problem was solved here?

stanislavlevin:
sorry, if it is wrong place to ask.
I'm interesting - how would `Requires(pre):` help?
```
Often package scriptlets need various other packages in order to execute correctly, and sometimes those dependencies aren’t even needed at runtime, only for installation. Such install/erase-time dependencies can be expressed with “Requires(): " notation, for example the following tells rpm that useradd program is needed by the package %pre scriptlet (often the case if a package uses a custom group/username for its files):

        Requires(pre): /usr/sbin/useradd

This has two “side-effects”:

    It ensures that the package providing /usr/sbin/useradd is installed before this package. In presence of dependency loops, scriptlet dependencies are the only way to ensure correct install order.
    If there are no other dependencies on the package providing /usr/sbin/useradd, that package is permitted to be removed from the system after installation(!)
```

abbra:
`bind-dyndb-ldap` has 

```
%dir %attr(770, root, named) %{_localstatedir}/named/dyndb-ldap
```

The group `named` is created in the pre-install script for `bind` package. If `bind` package is not installed before `bind-dyndb-ldap` is installed, RPM will complain that `named` group does not exist and will replace the group ownership for `%{_localstatedir}/named/dyndb-ldap` by `root`. We end up with `root:root` ownership and `bind-dyndb-ldap` driver will not be able to read the content of its work directory.

This happened due to refactoring of `bind-dyndb-ldap` spec file I did to allow rebuilding it against both `bind` and `bind-pkcs11`. Basically, the `%else ..` branch for non-`bind-pkcs11` case did not include *any* requires to `bind` anymore. As a result, RPM did install `bind-dyndb-ldap` but the order of installation with `bind` was not predictable. We still got `bind` through the dependency in `freeipa-server-dns` but when `bind-dyndb-ldap` was installed before `bind`, the ownership failed.

I only noticed this when RPM post-install scriptlet in `bind-dyndb-ldap` failed by not finding `/etc/named.conf`.

stanislavlevin:
Thank you for the explanation, I meant `Requires` will be more suitable in this case. In general, `Requires(pre)` means we need this package only in a pre-scriptlet. I don't aware of dnf specifics, but in non-freeipa cases dnf may remove bind.

abbra:
You are right. I pushed another update to rawhide that adds normal `Requires: bind ...` too.

abbra:
master:

* a297097bc2dbc3a4b5d036a784d5e6442ca467ae Change mkdir logic in DNSSEC

PR #5341, opened 2020-12-15
ipa-client-install: unilaterally set dns_lookup_kdc to True

fcami:
Previously, dns_lookup_kdc was only set to True if DNS
discovery worked or if the KDC was not specified on the
command-line.

Setting dns_lookup_kdc to False would result in a hardcoded
configuration which is less reliable in the long run.
For instance, adding a trust to an Active Directory forest
after clients are enrolled would result in clients not being
able to authenticate AD users. Recycling FreeIPA servers
could prove problematic if the original hostnames are not
reused too.

Change summary:
Always set dns_lookup_kdc to True on client enrollment.
With this change, DNS SRV search will always be performed
before looking into  /etc/krb5.conf realm entries.

Fixes: https://pagure.io/freeipa/issue/6523

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -693,16 +693,15 @@ def configure_krb5_conf(
     if not dnsok or not cli_kdc or force:
         libopts.extend([
             krbconf.setOption('dns_lookup_realm', 'false'),
-            krbconf.setOption('dns_lookup_kdc', 'false')
         ])
     else:
         libopts.extend([
             krbconf.setOption('dns_lookup_realm', 'true'),
-            krbconf.setOption('dns_lookup_kdc', 'true')
         ])
     libopts.extend([
         krbconf.setOption('rdns', 'false'),
         krbconf.setOption('dns_canonicalize_hostname', 'false'),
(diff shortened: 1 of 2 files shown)


fcami:
Flagging WIP - test incoming

fcami:
Both test failures are unrelated, hitting re-run

fcami:
hitting re-run - timeout is a PR-CI stall

fcami:
This PR soft-depends on https://github.com/freeipa/freeipa/pull/5343

fcami:
Expected failure: [test run without the fix](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/5a8fb4a4-3ee1-11eb-984a-fa163e03489e/report.html)

abbra:

abbra:
master:

* 18a1e7ee5c4b99f8c1362e5a549b88c22d8145df ipatests: make sure dns_lookup_kdc is always true
* e7827a6f0cd1785f914ec317f0be19535ea96e83 ipa-client-install: unilaterally set dns_lookup_kdc to True

PR #5367, opened 2020-12-19
Use uglifyjs on CentOS too

carlwgeorge:
Only checking for ID to equal "rhel" causes build failures on CentOS Stream.  Instead check both ID and ID_LIKE.  This should also work later on when rebuilds like CentOS Linux get this update.


abbra:
LGTM. Let's run the whole PR CI.

abbra:
master:

* 6ae1a05ee173caf05c768f892fe2b5f6408d8a78 Use uglifyjs on CentOS too

PR #5376, opened 2020-12-22
Fix uninstall without dns

abbra:
Service constructors are called even when the service itself is not configured. A common pattern in FreeIPA code is to instantiate a service and check whether it is configured, then perform uninstall of the service configuration. This fails if the service constructor does depend on the artifacts only present if other (relevant to the service) packages were installed.

A common pattern is:
```
  svc = SVCClass(..)
  if svc.is_configured(): 
      svc.uninstall()
```

Most of DNS-related service classes do resolution of UID/GIDs for ODS and NAMED in their constructors which breaks uninstallation of a DNS-less FreeIPA deployment because neither 'bind' nor 'opendnssec' packages are not installed and user and group they provide are not available in the system.

Fixes: https://pagure.io/freeipa/issue/8630

### DIFF
--- a/ipaserver/install/dnskeysyncinstance.py
+++ b/ipaserver/install/dnskeysyncinstance.py
@@ -55,10 +55,6 @@ class DNSKeySyncInstance(service.Service):
             keytab=paths.IPA_DNSKEYSYNCD_KEYTAB
         )
         self.extra_config = [u'dnssecVersion 1', ]  # DNSSEC enabled
-        self.named_uid = constants.NAMED_USER.uid
-        self.named_gid = constants.NAMED_GROUP.gid
-        self.ods_uid = constants.ODS_USER.uid
-        self.ods_gid = constants.ODS_GROUP.gid
 
     suffix = ipautil.dn_attribute_property('_suffix')
 
@@ -78,7 +74,7 @@ class DNSKeySyncInstance(service.Service):
             else:
                 os.chmod(directory, 0o770)
             # dnssec daemons require to have access into the directory
-            os.chown(directory, self.named_uid, self.named_gid)
+            constants.NAMED_USER.chown(directory, gid=constants.NAMED_GROUP.gid)
 
     def remove_replica_public_keys(self, replica_fqdn):
         ldap = api.Backend.ldap2
@@ -180,7 +176,7 @@ class DNSKeySyncInstance(service.Service):
             )
             with open(paths.DNSSEC_OPENSSL_CONF, 'w') as f:
                 os.fchmod(f.fileno(), 0o640)
-                os.fchown(f.fileno(), 0, self.named_gid)
+                os.fchown(f.fileno(), 0, gid=constants.NAMED_GROUP.gid)
                 f.write(named_openssl_txt)
 
     def setup_named_sysconfig(self):
@@ -235,7 +231,8 @@ class DNSKeySyncInstance(service.Service):
             os.mkdir(paths.IPA_DNSSEC_DIR)
             os.chmod(paths.IPA_DNSSEC_DIR, 0o770)
             # chown ods:named
-            os.chown(paths.IPA_DNSSEC_DIR, self.ods_uid, self.named_gid)
+            constants.ODS_USER.chown(paths.IPA_DNSSEC_DIR,
+                                     gid=constants.NAMED_GROUP.gid)
 
         # setup softhsm2 config file
         softhsm_conf_txt = ("# SoftHSM v2 configuration file \n"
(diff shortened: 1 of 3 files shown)


abbra:
Note that since PR CI and Azure do install all packages for integrated DNS-based deployment, it is only possible to validate these patches manually.

stanislavlevin:
It's legitimate to uninstall DNS packages on setup.

abbra:
@stanislavlevin that doesn't help -- they do not remove created users/groups in postinstall scripts because those may have already files owned in the file system.
```
[root@m1 ~]# rpm -q --scripts bind|egrep '(useradd|userdel|groupadd|groupdel)'
  /usr/sbin/groupadd -g 25 -f -r named >/dev/null 2>&1 || :;
  /usr/sbin/useradd  -u 25 -r -N -M -g named -s /sbin/nologin -d /var/named -c Named named >/dev/null 2>&1 || :;
[root@m1 ~]# rpm -q bind
bind-9.11.25-2.fc34.x86_64

```

stanislavlevin:
Future improvement for this case may be:
- place packages(rpms) somewhere on the filesystem
- install nothing by default
- install ipatests on controller
- install only required packages on setup from rpms storage

abbra:
Could you please file a ticket to PR CI for this? https://github.com/freeipa/freeipa-pr-ci/issues

ipaserver/install/dnskeysyncinstance.py
```diff
@@ -55,10 +55,10 @@ def __init__(self, fstore=None, logger=logger):
...
         )
         self.extra_config = [u'dnssecVersion 1', ]  # DNSSEC enabled
-        self.named_uid = constants.NAMED_USER.uid
-        self.named_gid = constants.NAMED_GROUP.gid
-        self.ods_uid = constants.ODS_USER.uid
-        self.ods_gid = constants.ODS_GROUP.gid
+        self.named_uid = constants.NAMED_USER
+        self.named_gid = constants.NAMED_GROUP
+        self.ods_uid = constants.ODS_USER
+        self.ods_gid = constants.ODS_GROUP
```
tiran, line 61: How about your remove these instance attributes completely and use ``constants.NAMED_USER.uid`` everwhere instead? IIRC the DNS code is the only code that has user/group instance vars.

ipaserver/install/dnskeysyncinstance.py
```diff
@@ -235,7 +235,7 @@ def __setup_softhsm(self):
             os.mkdir(paths.IPA_DNSSEC_DIR)
             os.chmod(paths.IPA_DNSSEC_DIR, 0o770)
             # chown ods:named
-            os.chown(paths.IPA_DNSSEC_DIR, self.ods_uid, self.named_gid)
+            os.chown(paths.IPA_DNSSEC_DIR, self.ods_uid.uid, self.named_gid.gid)
```
tiran, line 238:
```suggestion
            constants.ODS_USER.chown(paths.IPA_DNSSEC_DIR, gid=constants.NAMED_GROUP)
```

ipaserver/install/dnskeysyncinstance.py
```diff
@@ -272,7 +272,7 @@ def __setup_softhsm(self):
         os.mkdir(paths.DNSSEC_TOKENS_DIR)
         os.chmod(paths.DNSSEC_TOKENS_DIR, 0o770 | stat.S_ISGID)
         # chown to ods:named
-        os.chown(paths.DNSSEC_TOKENS_DIR, self.ods_uid, self.named_gid)
+        os.chown(paths.DNSSEC_TOKENS_DIR, self.ods_uid.uid, self.named_gid.gid)
```
tiran, line 275:
```suggestion
        constants.ODS_USER.chown(paths.DNSSEC_TOKENS_DIR, gid=constants.NAMED_GROUP)
```

ipaserver/install/dnskeysyncinstance.py
```diff
@@ -284,7 +284,7 @@ def __setup_softhsm(self):
         logger.debug("Saving user PIN to %s", paths.DNSSEC_SOFTHSM_PIN)
         with open(paths.DNSSEC_SOFTHSM_PIN, 'w') as f:
             # chown to ods:named
-            os.fchown(f.fileno(), self.ods_uid, self.named_gid)
+            os.fchown(f.fileno(), self.ods_uid.uid, self.named_gid.gid)
```
tiran, line 287:
```suggestion
            constants.ODS_USER.chown(f.fileno(), gid=constants.NAMED_GROUP)
```

tiran (commented):
You could simplify the code by removing the instance attributes and using the ``chown()`` helper method of the ``User`` object.

abbra:
will do, good suggestions.

abbra:
@tiran all tests are green.

tiran (approved):

abbra:
master:

* 8e16a1d169ddd6b6c1fee8f666ba7660b6d8ed6e opendnssecinstance: use late binding for UID/GID resolution
* b6a645338c447f8081c3e3501b7bfeb56178ca1d dnskeysyncinstance: use late binding for UID/GID resolution
* 4c4f16fc9cd08385fda90452669484ab647c47eb odsexporterinstance: use late binding for UID/GID resolution

PR #5377, opened 2020-12-22
ipatests: fix expected error message in test_commands 

flo-renaud:
389ds does not return any more additional information
about a failing bind (to avoid leaking information).

As a consequence, when ipa-nis-manage is provided a
wrong password, the error message contains less info
as in the past and needs to be fixed.

Fixes: https://pagure.io/freeipa/issue/8631

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -1286,17 +1286,16 @@ class TestIPACommand(IntegrationTest):
         This testcase checks if ipa-nis-manage enable
         command throws error on console for invalid DS admin password
         """
-        msg = (
-            "Insufficient access: Invalid credentials "
-            "Invalid credentials\n"
-        )
+        msg1 = "Insufficient access: "
+        msg2 = "Invalid credentials"
         result = self.master.run_command(
             ["ipa-nis-manage", "enable"],
             stdin_text='Invalid_pwd',
             raiseonerr=False,
         )
         assert result.returncode == 1
-        assert msg in result.stderr_text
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_commands.py
```diff
@@ -1286,17 +1286,16 @@ def test_ipa_nis_manage_enable_incorrect_password(self):
         This testcase checks if ipa-nis-manage enable
         command throws error on console for invalid DS admin password
         """
-        msg = (
-            "Insufficient access: Invalid credentials "
-            "Invalid credentials\n"
-        )
+        msg1 = "Insufficient access: "
```
kaleemsiddiqu, line 1289: @flo-renaud ,
Cannot we have just one string to check this? why we need two?

flo-renaud: @kaleemsiddiqu 
The error message when using the @389ds/389-ds-base-nightly is: 
`Insufficient access:  Invalid credentials`
 (with 2 spaces between : and I) instead of
`Insufficient access: Invalid credentials Invalid credentials`
I chose to check for the 2 msgs instead of writing a regexp compatible with one space or two spaces after the colon.

kaleemsiddiqu: ok.

kaleemsiddiqu:
LGTM

flo-renaud:
master:

* e7949e2a0f5874c26d021ac5daa027a789930886 ipatests: fix expected error message in test_commands

PR #5379, opened 2020-12-23
ipatests: remove test_acme from gating

flo-renaud:
test_acme is not stable and often needs to be
launched multiple times. Remove the test from gating
until the issue is fixed

Related: https://pagure.io/freeipa/issue/8602


kaleemsiddiqu:
LGTM

flo-renaud:
master:

* d0a1606e771a29e9f2f057d617dc482ab18a52b1 ipatests: remove test_acme from gating

PR #5386, opened 2020-12-27
Fix ACME installation with pki-core 10.9

abbra:
pki-core 10.9 uses different location for ACME LDAP content sources.
As a result, provisioning CA fails on CentOS 8 Stream.

Use any of the known locations for ACME LDAP schema, the correct
location will be fixed up in the ipa-server-upgrade run afterwards.

Fixes: https://pagure.io/freeipa/issue/8634

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -1481,7 +1481,11 @@ class CAInstance(DogtagInstance):
             logger.debug('ACME service is already deployed')
             return False
 
-        self._ldap_mod('/usr/share/pki/acme/database/ds/schema.ldif')
+        for path in ('/usr/share/pki/acme/database/ds/schema.ldif',
+                     '/usr/share/pki/acme/database/ldap/schema.ldif'):
+            if os.path.exists(path):
+                self._ldap_mod(path)
+                break
 
         configure_acme_acls()
 


rcritten:
Unless it's been backported, pki 10.9 lacks the distributed configuration changes.  I expect that ipa-acme-manage will not work.

abbra:
It is more of a problem on deploying CA at all. Without this patch `ipa-server-install` fails. With the patch it succeeds but `ipa-acme-manage` does not work:
```
[root@m1 ~]# ipa-acme-manage status
Failed to authenticate to CA REST API
The ipa-acme-manage command failed.
[root@m1 ~]# ipa-acme-manage enable
Failed to authenticate to CA REST API
The ipa-acme-manage command failed.
```

I think it is better to make sure we are able to configure CA at all.

abbra:
@rcritten for some reason newer pki-core 10.10 is not yet in CentOS 8 Stream, so we cannot get any IPA build working there without the patch. That's the reason I looked at it.

rcritten:
Sure, that makes sense, it just isn't going to work, as you've seen. The current ipa-acme-status uses the pki REST API to enable/disable ACME. Even if it was able to authenticate that part of the API won't be available so it still won't work. IMHO it isn't worth investigating further than fixing the broken install. We could look into not enabling ACME on pki < 10.10 by reverting 81c97bb9928a88a595b3afe6fa70fcfb267b1440

abbra:
Right, my focus here is to get a fix for a broken install

PR #5387, opened 2020-12-31
Raise RuntimeError when kinit_armor fails

Carbenium:
Adjust the raised error so it matches the expectations
of the calling code.

Closes: https://pagure.io/freeipa/issue/8639

### DIFF
--- a/ipalib/install/kinit.py
+++ b/ipalib/install/kinit.py
@@ -111,7 +111,7 @@ def kinit_armor(ccache_name, pkinit_anchors=None):
         use. Otherwise the value from Kerberos client library configuration is
         used
 
-    :raises: CalledProcessError if the anonymous PKINIT fails
+    :raises: RuntimeError if the anonymous PKINIT fails
     """
     logger.debug("Initializing anonymous ccache")
 
@@ -124,4 +124,6 @@ def kinit_armor(ccache_name, pkinit_anchors=None):
 
     # this workaround enables us to capture stderr and put it
     # into the raised exception in case of unsuccessful authentication
-    run(args, env=env, raiseonerr=True, capture_error=True)
+    result = run(args, env=env, raiseonerr=False, capture_error=True)
+    if result.returncode:
+        raise RuntimeError(result.error_output)


tiran:
The function is documented to raise a ``CalledProcessError`` exception. I'd rather fix the code in ``ipaserver.rpcserver`` to catch the correct exception.

Carbenium:
@tiran I based the changes on how `kinit_password` works and some `git blame`.
* The [comment](https://github.com/freeipa/freeipa/pull/5387/files#diff-fbf97caa320b745212acaa0ff2c35dc8253d08803d9636e6cf0b63a703eb1f7eR125-R126) doesn't match the code.
*  The documentation was added later in https://github.com/freeipa/freeipa/commit/3adb9ca875f8eb99e99a29e17a471a2b6f408a4a which was based on the faulty original code. The calling code in the original commit already expected `RuntimeError` (https://github.com/freeipa/freeipa/commit/b6741d81e187fc84177c12ef8ad900d3b5cda6a4#diff-21d951ac2d07631c0818b056e289cd02d980b05545f9eabb18b407178da0af0cR955-R959)

So I suspect the mistake was introduced in the original commit and the documentation was just based on that.

rcritten:
It was commit 3d2feac0 that changed the ipautil.run from raiseonerr=False to True and then caught the wrong exception.

I have no strong feeling on which exception should be used in this case and the method is called just once.

What is distinguishing is how the exception is raised. For example on an unconfigured system CalledProcessError vs RuntimeError:

Command ['kinit', '-n', '-c', '/tmp/c', '-X', 'X509_anchors=FILE:/etc/ipa/ca.crt'] returned non-zero exit status 1: "kinit: Cannot contact any KDC for realm 'EXAMPLE.TEST' while getting initial credentials\n"

vs

kinit: Cannot contact any KDC for realm 'EXAMPLE.TEST' while getting initial credentials

IMHO the second is much more readable. Is it important to know the command and args? I think the patch is fine as-is.

abbra:
On the other hand, knowing the arguments allows to re-run them independently and we used that multiple times on freeipa-users@ mailing list to investigate people's issues.

rcritten:
Fair point. It should probably continue to raise CalledProcessError and the except fixed, or the args included in the raised exception.

Carbenium:
/unstale

I will come back to this.

PR #5389, opened 2021-01-04
Revert "Remove test for minimum ACME support and rely on package deps"

rcritten:
This reverts commit 81c97bb9928a88a595b3afe6fa70fcfb267b1440.

This is to make IPA installable again with older versions of dogtag
so it will install on CentOS 8 Stream.

ACME will not be deployed but on upgrade, if pki 10.10.x is available
then it will be.

https://pagure.io/freeipa/issue/8634

Alternative PR to #5386

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -37,6 +37,7 @@ import syslog
 import time
 import tempfile
 from configparser import RawConfigParser
+from pkg_resources import parse_version
 
 from ipalib import api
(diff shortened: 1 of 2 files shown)


abbra:
LGTM. Let's take this one over #5386

flo-renaud:
master:

* 85d4f2d9c6f8ef7a9bd9a016d894ad273c58b6d2 Revert "Remove test for minimum ACME support and rely on package deps"

PR #5392, opened 2021-01-04
Add cgroup v2 support to the minimum RAM checker

rcritten:
It is similar to the v1 checking in that there is a set of files to use but if the value of the maximum available is 'max' then fall back to using psutil since all host memory is available.

https://pagure.io/freeipa/issue/8635

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -1072,27 +1072,49 @@ def check_available_memory(ca=False):
 
     Use Kb instead of KiB to leave a bit of slush for the OS
     """
+    available = None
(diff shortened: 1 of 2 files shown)


abbra:
LGTM, please remove temp commit.

rcritten:
Rebased and removed temp commit

abbra:
LGTM

flo-renaud:
master:

* 176fe83a01f249f214e031dbb7029f9e0944a4ad Add support for cgroup v2 to the installer memory checker
* 573d66e1f969143fa8c491a2308a8a893558810b ipatests: test the cgroup v2 memory restrictions

PR #5393, opened 2021-01-04
Include the CA profile when modifying request in ipa_certupdate

rcritten:
The CA tracking request is modified (it calls renew but it doesn't
actually do a renewal) as part of ipa-certupdate and it dropped
the profile. ipa-healthcheck discovered this condition.

https://pagure.io/freeipa/issue/8644

### DIFF
--- a/ipaclient/install/ipa_certupdate.py
+++ b/ipaclient/install/ipa_certupdate.py
@@ -188,7 +188,7 @@ def update_server(certs):
         #
         logger.debug("resubmitting certmonger request '%s'", request_id)
         certmonger.resubmit_request(
-            request_id, ca='dogtag-ipa-ca-renew-agent-reuse', profile='')
+            request_id, ca='dogtag-ipa-ca-renew-agent-reuse')
         try:
             state = certmonger.wait_for_request(request_id, timeout)
         except RuntimeError:
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -284,6 +284,20 @@ class TestIpaHealthCheck(IntegrationTest):
         assert output == \
             "ERROR: ipahealthcheck.meta.services.sssd: sssd: not running"
 
+    def test_ipa_healthcheck_after_certupdate(self):
+        """
+        Verify that ipa-certupdate hasn't messed up tracking
+
+        ipa-certupdate was dropping the profile value from the CA
+        signing cert tracking. ipa-healthcheck discovered this.
+
+        Run ipa-healthcheck after ipa-certupdate to ensure that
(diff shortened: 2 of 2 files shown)


ipaclient/install/ipa_certupdate.py
```diff
@@ -188,7 +188,8 @@ def update_server(certs):
         #
         logger.debug("resubmitting certmonger request '%s'", request_id)
         certmonger.resubmit_request(
-            request_id, ca='dogtag-ipa-ca-renew-agent-reuse', profile='')
+            request_id, ca='dogtag-ipa-ca-renew-agent-reuse',
+            profile='caCACert')
```
frasertweedale, line 192: [nit] Will it still work properly if we omit the `profile` kwarg?

frasertweedale (commented):
Nice that healthcheck is finding issues for us now :)

One minor question (see comments).

Otherwise, ACK after temp commit green and temp commit removed.

ipaclient/install/ipa_certupdate.py
rcritten: It looks like it will. Let me drop that. There is no need to override things here. Nice catch.

rcritten:
The test failed on uninstall with the known dbus issue.

flo-renaud:
@rcritten the test is probably failing because it's executed after `test_ipa_healthcheck_expiring` which is changing the date back and forth. I would move the new test before `test_ipa_healthcheck_expiring`.

rcritten:
I need to move this up further in the list. It failed because of a revoked cert in another test.

flo-renaud:
Adding re-run, the test failed in uninstall (known issue https://pagure.io/freeipa/issue/8506)

rcritten:
Re-run, test timeout. It's possible that with the new test + ipa-certupdate I'll need to extend the timeout.

flo-renaud:
LGTM, you can remove the temp commit

rcritten:
Temp commit removed.

rcritten:
The build failed, http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/16176060-505c-11eb-80e2-fa163e068b94

2021-01-06 20:19:47,169    DEBUG  Package libuv-1.40.0-1.fc32.x86_64.rpm is not signed
2021-01-06 20:19:47,169    DEBUG  Error: GPG check FAILED

flo-renaud:
master:

* 53f4fd91269dac89d83d255b174e6c0344c644f1 Don't change the CA profile when modifying request in ipa_certupdate
* 8e9fecd72ac1db3969583207e549d2bdadc20986 ipatests: test that no errors are reported after ipa-certupdate

PR #5401, opened 2021-01-06
ipatests: See if nologin supports -c before asserting message

rcritten:
Per the ssh_config(5) man page under ProxyCommand:

"The command string extends to the end of the line, and is
executed using the user's shell ‘exec’ directive to avoid a
lingering shell process."

<shell> -c <proxy command>

Some older versions of nologin (RHEL/CentOS) do not support
the -c option so will still fail but since nologin doesn't
actually execute properly it doesn't include the output
'This account is currently not available' so don't assert
in that case. The returncode of 1 is sufficient to know
that the login is denied.

https://pagure.io/freeipa/issue/7676

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -1419,6 +1419,14 @@ class TestIPACommand(IntegrationTest):
              'tuser1@%s' % self.master.hostname, 'cat /etc/hosts'],
         )
 
+        # Some versions of nologin do not support the -c option.
+        # ssh will still fail in a Match properly since it will return
+        # non-zero but we don't get the account failure message.
+        nologin = self.clients[0].run_command(
+            ['nologin', '-c', '/bin/true',],
+            raiseonerr=False
+        )
+
         # ssh as a restricted user to a restricted user should fail
         result = self.clients[0].run_command(
             ['sudo', '-u', restricted_user,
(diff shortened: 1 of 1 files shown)


fcami:
[logs](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/4995c12a-5057-11eb-a371-fa163e3d14f0/) show:
```
2021-01-06 19:45:03,856    DEBUG  Package libuv-1.40.0-1.fc32.x86_64.rpm is not signed
2021-01-06 19:45:03,856    DEBUG  Error: GPG check FAILED
```

rcritten:
It looks like python3-dns has new build-in error checking.

In ipatests.test_ipapython.test_dnsutil.TestSortSRV::test_prio this is failing:

>       hs = mksrv(-1, 0, 443, u"special")

```
cls = <class 'dns.rdtypes.IN.SRV.SRV'>, value = -1

    @classmethod
    def _as_uint16(cls, value):
        if not isinstance(value, int):
            raise ValueError('not an integer')
        if value < 0 or value > 65535:
>           raise ValueError('not a uint16')
E           ValueError: not a uint16

../dns/rdata.py:385: ValueError
```

In python3-dns-2.0.0-1.fc33 it would return:

<DNS IN SRV rdata: -1 0 443 special.>

rcritten:
Azure is using:

Processing /__w/1/s/dist/bundle/dnspython-2.1.0-py3-none-any.whl

rcritten:
Blocked by #5410

flo-renaud:
#5410 has been pushed, re-launching azure tests

abbra:
LGTM

flo-renaud:
master:

* dfa084217ed96c60bd9317de98e74d197f793420 ipatests: See if nologin supports -c before asserting message

PR #5408, opened 2021-01-07
upgrade.py: restart CS for 30 seconds until it is up

fcami:
Restart CS as many times as necessary within a 30-second window
to wait for DS to be ready.
    
Fixes: https://pagure.io/freeipa/issue/8645

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -429,7 +429,10 @@ def ca_enable_ldap_profile_subsystem(ca):
             quotes=False,
             separator='=')
 
-        ca.restart('pki-tomcat')
+        try:
+            ca.restart('pki-tomcat')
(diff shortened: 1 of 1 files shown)


rcritten:
I suspect an ExecPreStart script in the systemd unit is probably a better option to ensuring that LDAP is available. That or a script in the 389 unit to require that LDAP be available before it returns a status after startup.

tiran:
I recommend that you add 

```
[Unit]
Wants=dirsrv@{REALM}.service
After=dirsrv@{REALM}.service
```

to ``SYSTEMD_PKI_TOMCAT_IPA_CONF``. This ensures that systemd postpones dirsrv startup until dirsrv is running. It blocks the start of Dogtag instance and waits until 389-DS sends sd_notify message to systemd.

fcami:
Closing as the original issue is invalid.

PR #5411, opened 2021-01-08
Remove virtual attributes before rolling back a permission

rcritten:
On a failed permission update if the generated ACI is
invalid then the updated permission is rolled back.

Add the virtual relationship attributes to list of attributes
to be ignored when rolling back the entry.

This relies on the current order in the LDAPObject
relationships field where member and memberof are the first
two values.

https://pagure.io/freeipa/issue/8646

### DIFF
--- a/ipaserver/plugins/permission.py
+++ b/ipaserver/plugins/permission.py
@@ -1236,8 +1236,15 @@ class permission_mod(baseldap.LDAPUpdate):
             else:
                 self.obj.update_aci(entry, old_entry.single_value['cn'])
         except Exception:
-            # Don't revert attribute which doesn't exist in LDAP
+            # Don't revert attributes which don't exist in LDAP
             entry.pop('attributelevelrights', None)
+            for attr in list(
+                filter(
+                    lambda x: x not in ["member", "memberof"],
+                    [*self.obj.relationships],
+                )
+            ):
+                entry.pop(attr, None)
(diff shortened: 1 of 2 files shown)


ipaserver/plugins/permission.py
```diff
@@ -1238,6 +1238,9 @@ def post_callback(self, ldap, dn, entry, *keys, **options):
         except Exception:
             # Don't revert attribute which doesn't exist in LDAP
             entry.pop('attributelevelrights', None)
+            # NOTE: this relies on the order in relationships to not change
+            for attr in [*self.obj.relationships][2:]:
```
abbra, line 1242: May be it is better to have it more explicit:
```
>>> list(filter(lambda a: a not in [1,2], [1,2,3,4,5]))
[3, 4, 5]

```

rcritten: Yes, I knew this would be controversial :-) Thanks for the suggestion, implemented.

I also fixed some linting issues in the test.

abbra:
LGTM. ;)

flo-renaud:
master:

* 59d6c4da735af2ead833bcc0288873558a6928c4 Remove virtual attributes before rolling back a permission
* 8dd2eb45c689068874ebfe32162d5a650383301f ipatests: test that modifying a permission attrs handles failure

PR #5418, opened 2021-01-13
Support trust in FIPS mode

abbra:
## ipaserver/dcerpc.py: enforce SMB encryption on LSA pipe if available
    
We want to always use SMB encryption if it is possible on LSA pipe as we are going to pass what accounts to a plain-text content within CreateTrustedDomainEx2 call.
    
The catch is that older Samba version might not have a way to enforce this and we need fall back to work with existing connection then.
    
## ipaserver/dcerpc.py: use Kerberos authentication for discovery
    
In FIPS mode we cannot rely on NTLMSSP at all, so we have ensure Kerberos is used by Samba Python libraries. This is achieved by requiring credentials objects to always use Kerberos authentication.
    
Additionally, we have to normalize the principal used to authenticate. In case it was passed without realm, add forest root domain as a realm. In case it was passed with NetBIOS domain name, remove it and replace with a realm. Since we only know about the forest root domain as a realm, require that for other domains' users a real Kerberos principal is specified.

## ipaserver/dcerpc: use Samba-provided trust helper to establish trust
    
When establishing trust to Active Directory forest, RC4 is used to encrypt trusted domain object credentials as an application-specific material in a secure channel based on AES session key.
    
In FIPS mode it is not possible to use RC4 directly.
    
Samba 4.14 and backports to 4.13 in Fedora 33+ and RHEL 8.4+ now provide a helper that wraps LSA RPC call CreateTrustedDomainEx2.  This helper ensures that in FIPS mode we first check that LSA session key is AES before allowing RC4 use internally in Samba bindings. Thus, it becomes possible to establish trust to Active Directory forest in
FIPS mode.
    
Adopt FreeIPA code to use the helper provided by Samba when it is available. If neither the helper nor unprotected arcfour_encrypt utility is available from Samba bindings, fail import of the ipaserver.dcerpc module.
    
Fixes: https://pagure.io/freeipa/issue/8655

### DIFF
--- a/ipaserver/dcerpc.py
+++ b/ipaserver/dcerpc.py
@@ -1595,6 +1595,13 @@ def fetch_domains(api, mydomain, trustdomain, creds=None, server=None):
     return result
 
 
+def enforce_smb_encryption(creds):
+    try:
+        creds.set_smb_encryption(credentials.SMB_ENCRYPTION_REQUIRED)
+    except AttributeError:
+        pass
+
+
 def retrieve_remote_domain(hostname, local_flatname,
                            realm, realm_server=None,
                            realm_admin=None, realm_passwd=None):
@@ -1639,6 +1646,7 @@ def retrieve_remote_domain(hostname, local_flatname,
                           % (realm_admin, realm_passwd)
             td = get_instance(local_flatname)
             td.creds.set_kerberos_state(credentials.MUST_USE_KERBEROS)
+            enforce_smb_encryption(td.creds)
             td.creds.parse_string(auth_string)
             td.creds.set_workstation(hostname)
             if realm_server is None:
@@ -1678,6 +1686,7 @@ class TrustDomainJoins:
         ld = TrustDomainInstance(self.local_flatname)
         ld.creds = credentials.Credentials()
         ld.creds.set_kerberos_state(credentials.MUST_USE_KERBEROS)
+        enforce_smb_encryption(ld.creds)
         ld.creds.guess(ld.parm)
         ld.creds.set_workstation(ld.hostname)
         ld.retrieve(FQDN)
--- a/ipaserver/dcerpc.py
+++ b/ipaserver/dcerpc.py
@@ -1618,14 +1618,27 @@ def retrieve_remote_domain(hostname, local_flatname,
     rd.read_only = True
     if realm_admin and realm_passwd:
         if 'name' in rd.info:
+            realm_netbios = ""
             names = realm_admin.split('\\')
             if len(names) > 1:
                 # realm admin is in DOMAIN\user format
                 # strip DOMAIN part as we'll enforce the one discovered
                 realm_admin = names[-1]
-            auth_string = r"%s\%s%%%s" \
-                          % (rd.info['name'], realm_admin, realm_passwd)
+                realm_netbios = names[0]
+            names = realm_admin.split('@')
+            if len(names) == 1:
+                if all([len(realm_netbios) != 0,
+                        realm_netbios.lower() != rd.info['name'].lower()]):
+                    raise errors.ValidationError(
+                        name=_('Credentials'),
+                        error=_('Non-Kerberos user name was specified, '
+                                'please provide user@REALM variant instead'))
+                realm_admin = r"%s@%s" % (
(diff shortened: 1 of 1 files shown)


abbra:
Tested on RHEL 8.4 nightly in FIPS mode and without it:
```
[root@master ~]# sysctl crypto.fips_enabled
crypto.fips_enabled = 1
[root@master ~]# ipa trust-add win2016.test --admin Administrator --password 
Active Directory domain administrator's password: 
-----------------------------------------------------
Added Active Directory trust for realm "win2016.test"
-----------------------------------------------------
  Realm name: win2016.test
  Domain NetBIOS name: WIN2016
  Domain Security Identifier: S-1-5-21-3674478510-3489545405-2687489055
  Trust direction: Trusting forest
  Trust type: Active Directory domain
  Trust status: Established and verified
[root@master ~]# systemctl restart sssd
[root@master ~]# sssctl domain-list
implicit_files
ipa.test
win2016.test
sub1.win2016.test
[root@master ~]# sssctl domain-status win2016.test
Online status: Online

Active servers:
AD Global Catalog: not connected
AD Domain Controller: not connected
IPA: master.ipa.test

Discovered AD Global Catalog servers:
None so far.
Discovered AD Domain Controller servers:
None so far.
Discovered IPA servers:
- master.ipa.test

[root@master ~]# id administrator@win2016.test
uid=117200500(administrator@win2016.test) gid=117200500(administrator@win2016.test) groups=117200500(administrator@win2016.test),117200519(enterprise admins@win2016.test),117200518(schema admins@win2016.test),117200513(domain users@win2016.test),117200520(group policy creator owners@win2016.test),117200512(domain admins@win2016.test)
[root@master ~]# sssctl domain-status win2016.test
Online status: Online

Active servers:
AD Global Catalog: ad2.sub1.win2016.test
AD Domain Controller: ad1.win2016.test
IPA: master.ipa.test

Discovered AD Global Catalog servers:
- ad2.sub1.win2016.test
- ad1.win2016.test

Discovered AD Domain Controller servers:
- ad1.win2016.test

Discovered IPA servers:
- master.ipa.test
[root@master ~]# kdestroy
[root@master ~]# kinit Administrator@WIN2016.TEST
Password for Administrator@WIN2016.TEST: 
[root@master ~]# ssh -k -l Administrator@WIN2016.TEST master.ipa.test
Activate the web console with: systemctl enable --now cockpit.socket

This system is not registered to Red Hat Insights. See https://cloud.redhat.com/
To register this system, run: insights-client --register
Could not chdir to home directory /home/win2016.test/administrator: No such file or directory
[administrator@win2016.test@master /]$

```

rcritten:
The code looks ok to me and the commit messages very clear.

I don't mean to bikeshed but the trust_utils naming doesn't seem like it is going to age well. IPA is assuming that because the trust_utils import exists that FIPS-compatible wrapping is supported. Should it instead import CreateTrustedDomainRelax directly? I'm assuming that the trust_utils import was added specifically to solve this problem?

It's also curious that trust_utils has its own arcfour_encrypt. Does that also infer that wrapped-RC4 happens?

abbra:
It is the same `arcfour_encrypt`, we import it from the same location in Samba Python bindings. I added an import guard for missing `arcfour_encrypt` in case it will disappear from Samba Python bindings and will not be directly accessible _and_ we are not able to use CreateTrustedDomainRelax.

Samba code imports `CreateTrustedDomainRelax` from `trust_utils`:

```
$ git grep trust_utils
python/samba/netcmd/domain.py:from samba.trust_utils import CreateTrustedDomainRelax
python/samba/tests/dcerpc/createtrustrelax.py:from samba.trust_utils import CreateTrustedDomainRelax
```
I updated the code. ;)

rcritten:
Perfect, thanks. It is very clear what the code is doing now. ack.

abbra:
master:

* cf17b7af5a8cd835ae20fe23a57fe656481e27a3 ipaserver/dcerpc: use Samba-provided trust helper to establish trust
* e157ea1e14edd239c58847957fe0ae7c7fa5c9ab ipaserver/dcerpc.py: use Kerberos authentication for discovery
* fd15f60216b4098d0423fc39a0ad62293df5d116 ipaserver/dcerpc.py: enforce SMB encryption on LSA pipe if available

PR #5419, opened 2021-01-13
Test that IPA certs are removed on server uninstall 

antoniotorresm:
Test that IPA certs are removed after the server uninstall
process.

Related: https://pagure.io/freeipa/issue/8614

### DIFF
--- a/ipatests/pytest_ipa/integration/tasks.py
+++ b/ipatests/pytest_ipa/integration/tasks.py
@@ -963,6 +963,15 @@ def uninstall_master(host, ignore_topology_disconnect=True,
 
     result = host.run_command(uninstall_cmd)
     assert "Traceback" not in result.stdout_text
+
+    # Check that IPA certs have been deleted after uninstall
+    # Related: https://pagure.io/freeipa/issue/8614
+    assert host.run_command(['test', '-f', paths.IPA_CA_CRT],
+                            raiseonerr=False).returncode == 1
+    assert host.run_command(['test', '-f', paths.IPA_P11_KIT],
+                            raiseonerr=False).returncode == 1
+    assert "IPA CA" not in host.run_command(['trust', 'list']).stdout_text
+
     if clean:
         Firewall(host).disable_services(["freeipa-ldap", "freeipa-ldaps",
                                          "freeipa-trust", "dns"])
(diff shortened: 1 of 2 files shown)


ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -963,6 +963,16 @@ def uninstall_master(host, ignore_topology_disconnect=True,
...
     result = host.run_command(uninstall_cmd)
     assert "Traceback" not in result.stdout_text
+
+    # Check that IPA certs have been deleted after uninstall
+    # Related: https://pagure.io/freeipa/issue/8614
+    assert host.run_command(['test', '-f', paths.IPA_CA_CRT],
+                            raiseonerr=False).returncode == 1
+    assert host.run_command(['test', '-f', paths.IPA_P11_KIT],
+                            raiseonerr=False).returncode == 1
+    assert host.run_command("trust list | grep IPA",
```
fcami, line 973: You can use shlex for this. You can grep in ipatests/ to see how it works.

fcami (changes requested):

ipatests/pytest_ipa/integration/tasks.py
fcami: Or you could fetch the whole output (e.g. not use grep) and use substring matching instead.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -963,6 +963,16 @@ def uninstall_master(host, ignore_topology_disconnect=True,
...
     assert "Traceback" not in result.stdout_text
+
+    # Check that IPA certs have been deleted after uninstall
+    # Related: https://pagure.io/freeipa/issue/8614
+    assert host.run_command(['test', '-f', paths.IPA_CA_CRT],
+                            raiseonerr=False).returncode == 1
+    assert host.run_command(['test', '-f', paths.IPA_P11_KIT],
+                            raiseonerr=False).returncode == 1
+    assert "IPA CA" not in host.run_command(['trust', 'list'],
+                                            raiseonerr=False).stdout_text
```
rcritten, line 974: If this command fails this assert will still pass. Dropping the raiseonerr=False will address this. Or you can assign the result to a variable and assert returncode == 0 along with the 'IPA CA' not in stdout_text.

fcami (commented):
LGTM. Please remove the temp commit and I'll ack the PR.

fcami (approved):
ACK

flo-renaud:
master:

* 6f830ae1ad1ee62a597bcfad81c9d6a4784c24ee Test that IPA certs are removed on server uninstall
* 79d9790c9954ec6ab97e8285074f6cf41e346879 Check that IPA cert is added to trust store after server install

PR #5424, opened 2021-01-14
ipatest: fix test_upgrade.py::TestUpgrade::()::test_kra_detection

flo-renaud:
Modify the test scenario in order to be independant from PKI
behavior. The aim of the test is to ensure that the KRA
detection is not based on the presence of the directory
/var/lib/pki/pki-tomcat/kra/.
Previously the test was calling ipa-server-upgrade but this cmd
may fail even with the kra detection fix because of an issue in
pki (https://github.com/dogtagpki/pki/issues/3397).
Instead of exercising the whole ipa-server-upgrade command, the
test now checks the output of the API kra.is_installed() to validate
KRA detection mechanism.

Fixes: https://pagure.io/freeipa/issue/8653
Related: https://pagure.io/freeipa/issue/8596

### DIFF
--- a/ipatests/test_integration/test_upgrade.py
+++ b/ipatests/test_integration/test_upgrade.py
@@ -285,21 +285,26 @@ class TestUpgrade(IntegrationTest):
         Test for https://pagure.io/freeipa/issue/8596
         When the directory /var/lib/pki/pki-tomcat/kra/ exists, the upgrade
         wrongly assumes that KRA component is installed and crashes.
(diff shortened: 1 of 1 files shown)


fcami (approved):
ACK

flo-renaud:
master:

* 6e0634bd72f60295b86a1561adf94f14d6948009 ipatest: fix test_upgrade.py::TestUpgrade::()::test_kra_detection

PR #5430, opened 2021-01-14
Value stored to 'krberr' is never read in ipa-rmkeytab.c

rcritten:
Clang reports that the return values for the keytab cursor commands in ipa-rmkeytab are not checked consistently.

Also replace the numeric return values with defines for clarity.

### DIFF
--- a/client/ipa-rmkeytab.c
+++ b/client/ipa-rmkeytab.c
@@ -29,6 +29,13 @@
 #include "ipa-client-common.h"
 #include "config.h"
 
+#define KERBEROS_ERROR	1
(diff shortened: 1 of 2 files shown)


abbra:
LGTM. I re-ran failed jobs which aren't related to the fix.

flo-renaud:
master:

* 90eef4f565fc56bf50ff95707ac835f2b1b83538 ipa-rmkeytab: convert numeric return values to #defines
* f3f9672d527008dc741ac90aa465bac842eea08d ipa-rmkeytab: Check return value of krb5_kt_(start|end)_seq_get

PR #5434, opened 2021-01-15
ipatests: use fully qualified name for AD admin when establishing trust

wladich:
Changes in https://pagure.io/freeipa/issue/8655 made it impossible
to use AD admin name without domain part in "ipa trust-add" command to
establish external trust with an AD tree domain.
Also use fully qualified admin name by default in all trust related tests
to reduce abiguity

UPDATE: related to https://pagure.io/freeipa/issue/8664

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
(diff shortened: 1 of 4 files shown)


wladich:
Tests are still failing with same symptoms: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/3f378b70-573a-11eb-8e56-fa163e95e74a/

Re-running with smb debug enabled

wladich:
The observed failures are due to https://pagure.io/freeipa/issue/8664. Waiting until the issue is fixed

wladich:
Merged as part of #5436

PR #5435, opened 2021-01-17
ipatests: fix expected errmsg in TestTrust::test_ipa_commands_run_as_aduser

flo-renaud:
389ds does not return any more additional information
about a failing bind (to avoid leaking information).

As a consequence, when ipa ping is executed with an AD user
the error message contains less info as in the past and needs to be fixed.

Fixes: https://pagure.io/freeipa/issue/8668
Signed-off-by: Florence Blanc-Renaud <flo@redhat.com>

PR #5437, opened 2021-01-18
ipatests: check error message less strictly

wladich:
Error reporting of 389ds was changed in
389ds/389-ds-base#4480 to return no
additional information about a failing bind (to avoid leaking information).

As a consequence, when an unauthorized user tries to perform
administrative task on IPA server the error message contains less info.

The assertion was changed to accept old and new variants of error message.


mrizwan93:
Similar PR raised by @flo-renaud https://github.com/freeipa/freeipa/pull/5435

wladich:
@mrizwan93 

Thanks for info, dropping the PR in favor of #5435

PR #5440, opened 2021-01-18
ipatests: fix expected output for ipahealthcheck.ipa.files

flo-renaud:
With ipa-healthcheck 0.8, the test ipahealthcheck.ipa.files is able
to return a list of possible owners/groups as a comma-separated string
instead of a single owner/group (see commit 930ec5f).

The test output needs to be fixed accordingly.

Fixes: https://pagure.io/freeipa/issue/8662

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -1620,6 +1620,19 @@ class TestIpaHealthCheckFileCheck(IntegrationTest):
         version = tasks.get_healthcheck_version(self.master)
         if parse_version(version) < parse_version("0.6"):
             pytest.skip("Skipping test for 0.4 healthcheck version")
+
+        # ipa-healthcheck 0.8 returns a list of possible owners instead
+        # of a single value
+        if parse_version(version) >= parse_version("0.8"):
+            expected_owner = 'root,systemd-resolve'
+            expected_msg = ("Ownership of %s is admin "
+                            "and should be one of root,systemd-resolve"
+                            % paths.RESOLV_CONF)
+        else:
+            expected_owner = 'root'
+            expected_msg = ("Ownership of %s is admin and should be root"
+                            % paths.RESOLV_CONF)
+
         modify_permissions(self.master, path=paths.RESOLV_CONF, owner="admin")
         returncode, data = run_healthcheck(
             self.master,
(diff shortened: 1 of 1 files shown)


flo-renaud:
Note: this PR needs to wait for ipa-healthcheck PR #[184](https://github.com/freeipa/freeipa-healthcheck/pull/184)

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -1632,18 +1645,27 @@ def test_ipa_filecheck_bad_owner(self, modify_permissions):
...
+            assert check["kw"]["msg"] == expected_msg
 
     def test_ipa_filecheck_bad_group(self, modify_permissions):
         version = tasks.get_healthcheck_version(self.master)
         if parse_version(version) < parse_version("0.6"):
             pytest.skip("Skipping test for 0.4 healthcheck version")
+
+        # ipa-healthcheck 0.8 returns a list of possible groups instead
+        # of a single value
+        if parse_version(version) >= parse_version("0.8"):
```
rcritten, line 1659: I realize I +1 this idea yesterday but in the morning light I don't think this is going to work in general. This behavior is less dependent upon ipa-healthcheck and more dependent on the underlying system (F33+). F32 is going to get 0.8 as well so this test will break again.

I think instead checking if systemd-resolved is active then expect root,systemd-resolve.

flo-renaud: Do you mean that ipa-healthcheck is dynamically checking if systemd-resolve is in use and crafts the possible list of owners/groups based on that? To me it looks like the opposite, the expected owner/group is statically set:
https://github.com/freeipa/freeipa-healthcheck/blob/master/src/ipahealthcheck/ipa/files.py#L95
and only depends on healthcheck version.

rcritten: No, it does not check dynamically (not a bad idea actually).

I think I misread the test code. I was worried about behavior in F-32 but looking again I think it's fine as-is.

flo-renaud:
For the record: successful temp commits:
- `fedora-previous/test_ipahealthcheck_nodns_extca_file`: [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/a6f27b4e-5a64-11eb-bf93-fa163ee3281e)
- `fedora-rawhide/test_ipahealthcheck_nodns_extca_file`: [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/d7abe566-5b1f-11eb-a618-fa163e45d937)

Temp commit removed.

rcritten:
LGTM. What branches do you want this backported to?

flo-renaud:
Adding ipa-4-9 and ipa-4-8 labels for backports.

rcritten:
master:

* 0801d4c0581e55aea6cfc2befbd419522be7b642 ipatests: fix expected output for ipahealthcheck.ipa.files

PR #5442, opened 2021-01-19
ipatests: fix healthcheck test for ipahealthcheck.ds.encryption 

flo-renaud:
389ds is combining the value set in dse.ldif and the current crypto
policy to evaluate the min TLS version that it will be using.
The test needs to change the crypto policy to LEGACY in order to allow
TLS 1.0, because the DEFAULT policy prevents TLS 1.0 on fc33+.

Fixes: https://pagure.io/freeipa/issue/8670

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -1026,6 +1026,10 @@ class TestIpaHealthCheck(IntegrationTest):
         """
         instance = realm_to_serverid(self.master.domain.realm)
         cmd = ["systemctl", "restart", "dirsrv@{}".format(instance)]
+        # The crypto policy must be set to LEGACY otherwise 389ds
+        # combines crypto policy amd minSSLVersion and removes
+        # TLS1.0 on fedora>=33 as the DEFAULT policy forbids TLS1.0
+        self.master.run_command(['update-crypto-policies', '--set', 'LEGACY'])
         self.master.run_command(
             [
                 "dsconf",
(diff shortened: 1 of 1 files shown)


kaleemsiddiqu:
@flo-renaud 
How about a new test case that dsconf does not works for --tls-protocol-min if protocol is not minimum as specified by crypto-policies by default?

flo-renaud:
> @flo-renaud
> How about a new test case that dsconf does not works for --tls-protocol-min if protocol is not minimum as specified by crypto-policies by default?

@kaleemsiddiqu 
It would be more relevant in 389ds source base IMO. And I wouldn't say that it _doesn't work_, because according to https://access.redhat.com/documentation/en-us/red_hat_directory_server/11/html/configuration_command_and_file_reference/core_server_configuration_reference#sslVersionMin:
>  The sslVersionMin parameter sets the minimum version of the TLS protocol Directory Server uses. However, by default, Directory Server sets this parameter automatically based on the system-wide crypto policy. If you set the crypto policy profile in the /etc/crypto-policies/config file to:
>
>    DEFAULT, FUTURE, or FIPS, Directory Server sets sslVersionMin to TLS1.2
>    LEGACY, Directory Server sets sslVersionMin to TLS1.0 
>
> Alternatively, you can manually set sslVersionMin to higher value than the one defined in the crypto policy. 

The doc is clear that only a higher value would be taken into account.

kaleemsiddiqu:
LGTM, temp commit can be removed.

flo-renaud:
Thanks for the review, temp commit removed.

flo-renaud:
master:

* 279d8b7ec49585eab47baab145eb89d177fc3f24 ipatests: fix healthcheck test for ipahealthcheck.ds.encryption

PR #5444, opened 2021-01-19
Various dnssec fixes

flo-renaud:
- dnssec: fix the key type with OpenDNSSEC 2.1
- ipatests: add a test for ZSK/KSK keytype in DNSKEY record 
- OpenDNSSEC: fix timezone in key creation date

- Run a temp commit using @tbordaz's fix for 389ds sync_repl issue

Note: the patch does not handle the upgrade yet (when ZSK was created as a KSK, then upgrade should fix the LDAP entry and make sure that dns re-loads the record).

### DIFF
--- a/ipaserver/dnssec/_ods21.py
+++ b/ipaserver/dnssec/_ods21.py
@@ -3,6 +3,7 @@
 #
 
 import os
+import dateutil.tz
 
 from ipaserver.dnssec._odsbase import AbstractODSDBConnection
 from ipaserver.dnssec._odsbase import AbstractODSSignerConn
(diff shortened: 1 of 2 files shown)


ipaserver/dnssec/_ods21.py
```diff
@@ -47,9 +47,9 @@ def get_keys_for_zone(self, zone_id):
             key['active'] = None
             key['retire'] = None
             key['dead'] = None
-            if row['keyType'] == 2:
+            if row['role'] == 2:
                 key['keytype'] = 256
-            elif row['keyType'] == 1:
+            elif row['role'] == 1:
```
rcritten, line 56: I don't know if it's worth mentioning, I know next to zero about DNSSEC, but the best documentation for the magic numbers 256 and 257 that I was able to dig up was at https://docs.infoblox.com/display/NAG8/DNSKEY+Resource+Records

The keyType -> role change checks out.

rcritten:
The patch looks good to me. The commit messages are excellent and should help the next person that needs to update this.

flo-renaud:
Temp commit removed

flo-renaud:
master:

* cdfc86364ee2eed2aafda2e8b3a484a7a264677e dnssec: fix the key type with OpenDNSSEC 2.1
* 7902c784963c64b2d0fba8e7fa03529e793c221c ipatests: add a test for ZSK/KSK keytype in DNSKEY record
* ca17a81a30cc83568008bad9031db3b8c6b90f2f OpenDNSSEC: fix timezone in key creation date

PR #5452, opened 2021-01-19
Change NameType of custodia client's service

frozencemetery:
Using hostbased_service for initiators hits an unresolved corner case in
Kerberos with respect to realm selection and canonicalization.  However,
since there is only a single entry in the keytab, we do not need to
specify a name at all.
    
Fixes: https://pagure.io/freeipa/issue/8672

### DIFF
--- a/ipaserver/install/custodiainstance.py
+++ b/ipaserver/install/custodiainstance.py
@@ -239,7 +239,6 @@ class CustodiaInstance(SimpleServiceInstance):
         self._wait_keys()
 
         return CustodiaClient(
-            client_service='host@{}'.format(self.fqdn),
             keyfile=self.server_keys, keytab=paths.KRB5_KEYTAB,
             server=self.custodia_peer, realm=self.realm
         )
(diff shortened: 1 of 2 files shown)


tiran:
Can we safely assume that ``/etc/krb5.keytab`` only contains entries for the current host? @simo5

abbra:
@tiran depends on where that happens. For IPA replicas it might be true but on IPA clients in general it might be quite unsafe assumption. Given that clients might get promoted to replicas we might hit the cases with multiple principals in the keytab that do not necessary belong to the same host and also might even be from different realms.

simo5:
In general we could have keys for aliases in the keytab on clients (and on servers one day, for load balancing, or simply aliasing). So, no we cannot assume only keys for a single hostname are present in a keytab.

simo5:
Perhaps using GSS_KRB5_NT_PRINCIPAL_NAME in this case is a better option *if* we need to change code on the freeipa side at all.

PR #5453, opened 2021-01-19
Reduce the differences between upstream and downstream builds

rcritten:
Make it easier to share the same spec file between upstream and downstream (at least CentOS and RHEL) builds.

This includes:

- Replace some FreeIPA references with IPA to be more name-agnostic.
- Drop csrgen which was never feature complete, doesn't work today.
- Tweak the UI builder which was already capable of picking a tool to be more generic in its output.

The spec file itself has already been tweaked to handle picking the right dependencies per distribution.


rcritten:
Downstream builders probably will want to replace the logos and/or update the CSS to match whatever color scheme they want the UI:

install/ui/css/patternfly.css
install/ui/less/brand.less
install/ui/less/patternfly.less

abbra:
LGTM

flo-renaud:
master:

* 6518a600b46bf6db57ec1adc1282999293e1dfb8 Change FreeIPA references to IPA and Identity Management
* 767232b478ffd7f0a8f53618c17c531a7e70d3c6 Remove support for csrgen
* 8021a2066d562aac56359684642ee9cd13b7622b Remove references to rjsmin in UI compile.sh

PR #5458, opened 2021-01-21
ipatests: fix discrepancies in nightly defs

flo-renaud:
- Build is using a prio of 100 while tests use 50, use consistent
values
- fix the requires for test_installation_client

### DIFF
--- a/ipatests/prci_definitions/nightly_latest.yaml
+++ b/ipatests/prci_definitions/nightly_latest.yaml
@@ -1536,7 +1536,7 @@ jobs:
 
   fedora-latest/test_membermanager:
     requires: [fedora-latest/build]
-    priority: 100
+    priority: 50
     job:
       class: RunPytest
       args:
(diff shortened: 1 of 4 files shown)


netoarmando:
@flo-renaud, I have a hunch that the priority values for testing using 100 was because gating is setting them as 100 and build as 150. My suggestion is to use this PR to change gating as well just in case someone wants to copy/paste definitions.

flo-renaud:
> @flo-renaud, I have a hunch that the priority values for testing using 100 was because gating is setting them as 100 and build as 150. My suggestion is to use this PR to change gating as well just in case someone wants to copy/paste definitions.

@netoarmando the higher values in gating were ensuring that PRs were processed before nightly tests (see https://github.com/freeipa/freeipa/commit/991d508a5c9ec6b19e9a937864fd24b765dff7f2). As of today, the nightlies are run on a different set of machines but I would like to keep this priority (for instance I am sometimes using the nightlies infra when testing 389ds patches).

netoarmando (approved):

flo-renaud:
master:

* 9b2c05affc79e61d0698f8dbc3b19ad0d000d31d ipatests: fix discrepancies in nightly defs

PR #5459, opened 2021-01-21
Don't install csrgen extra dependencies

tiran:
See: https://pagure.io/freeipa/issue/8669


rcritten:
Gah, my git-fu must be on the fritz :/

Can you add https://pagure.io/freeipa/issue/8669 to the commit message?

rcritten:
master:

* ce2e41ab6e86433ba2cd3c33019b5e010c0302a9 Don't install csrgen extra dependencies

PR #5472, opened 2021-01-22
Manual backport of 9f0ec27e9f13ed40b8e58162d99bf9b0e8b4afd5.

fcami:
Manual backport of 9f0ec27e9f13ed40b8e58162d99bf9b0e8b4afd5.
Original commit message:
****************************************************
ipaCASubjectDN is used by lightweight sub CA feature.

ipaExternalMember is used by KRB driver to assemble MS-PAC records.

ipaNTSecurityIdentifier was only index for "pres" and was missing an
index on "eq". Samba and ipasam perform queries with SID string.

memberPrincipal is used by S4U2Proxy constrained delegation and by
ipa-custodia.

Also note that dnaHostname, ipServiceProtocol, ipaCertSubject, and
ipaKeyUsage are currently not index because an index would rarely used
or have a poor selectivity.

****************************************************

The ipaNTSecurityIdentifier entry was missing in ipa-4-6 and is
added by this commit.

Fixes: https://pagure.io/freeipa/issue/8677

### DIFF
--- a/install/updates/20-indices.update
+++ b/install/updates/20-indices.update
@@ -380,3 +380,32 @@ default: objectClass: top
 default: objectClass: nsIndex
 default: nsSystemIndex: false
 default: nsIndexType: eq
+
+dn: cn=ipaCASubjectDN,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+only:cn: ipaCASubjectDN
+default:objectClass: nsIndex
+default:objectClass: top
+default:nsSystemIndex: false
+add:nsIndexType: eq
+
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @fcami 
on ipa-4-6 branch the file install/share/indices.ldif is still used. I am not sure whether it is also needed to add the new indices in that file or it is redundant (IIRC the installer calls upgrade at the end but I wanted to double-check).
Otherwise LGTM.

flo-renaud:
Tested with new installation and upgrade from freeipa-server-4.6.4-2.fc27.x86_64. In both case the indexes are properly defined.

flo-renaud:
ipa-4-6:

* 2c802ab10479342e6d0e7e6d1589f0d31a608c1b Add more indices

PR #5475, opened 2021-01-26
Use Fedora 33 for temp commit

wladich:
This is to keep temp commit template in line with "nightly_latest.yaml"


wladich:
Gating tests are still running on F32, hence temp_commit should do the same

PR #5486, opened 2021-01-27
[ipatests][Azure Pipelines] Populate containers with self-AAAA records

stanislavlevin:
- IPA server's AAAA records at embedded DNS mode depend on result of `get_server_ip_address` function(`ipaserver.install.installutils`), which in turn, relies on NSS. In case of Azure Pipelines, there are neither IPv6 records in '/etc/hosts' nor external DNS, which may provide such. This leads to the missing AAAA records for master and missing AAAA records for `ipa-ca` pointing to master in embedded DNS. In particular, tests `test_ipa_healthcheck_no_errors`, `test_ipa_dns_systemrecords_check` fail with:
```
    [
      {
        "source": "ipahealthcheck.ipa.idns",
        "check": "IPADNSSystemRecordsCheck",
        "result": "WARNING",
        "uuid": "b979a88a-6373-4990-bc83-ce724e9730b4",
        "when": "20210120055054Z",
        "duration": "0.032740",
        "kw": {
          "msg": "Got {count} ipa-ca AAAA records, expected {expected}",
          "count": 1,
          "expected": 2
        }
      }
    ]
```
where `ipa-ca` record exists only for replica.
    
Note: since the most of the code in setup_containers was touched it has been reformatted.

- Handle AAAA records in test_ipa_dns_systemrecords_check

Fixes: https://pagure.io/freeipa/issue/8683

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -861,6 +861,8 @@ Requires: python3-pexpect
 # These packages do not exist on RHEL and for ipatests use
 # they are installed on the controller through other means
 Requires: ldns-utils
+# update-crypto-policies
+Requires: crypto-policies-scripts
 Requires: python3-polib
 Requires: python3-pytest >= 3.9.1
 Requires: python3-pytest-multihost >= 0.5
--- a/ipatests/azure/scripts/setup_containers.py
+++ b/ipatests/azure/scripts/setup_containers.py
@@ -5,20 +5,21 @@ import subprocess
 import docker
 from jinja2 import Template
 
-logging.basicConfig(level=logging.INFO, format='%(levelname)s: %(message)s')
+logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
 
-IPA_TESTS_ENV_WORKING_DIR = os.environ.get('IPA_TESTS_ENV_WORKING_DIR')
-IPA_TESTS_ENV_NAME = os.environ.get('IPA_TESTS_ENV_NAME')
-IPA_TESTS_ENV_ID = os.environ.get('IPA_TESTS_ENV_ID', '1')
-IPA_TESTS_CLIENTS = int(os.environ.get('IPA_TESTS_CLIENTS', 0))
-IPA_TESTS_REPLICAS = int(os.environ.get('IPA_TESTS_REPLICAS', 0))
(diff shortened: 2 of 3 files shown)


rcritten:
No specific comments on the core changes but there is a significant number of single to double quote changes that seem to be formatting-only.

stanislavlevin:
I've especially noted that in PR message and the commit's one:

> Note: since the most of the code in setup_containers was touched it has been reformatted.

abbra:
Yep, in that case it makes sense to do some uniformity. Thanks!

stanislavlevin:
@rcritten, if you strongly object to this I can revert extra changes, even though it was convenient for me to use `black` here.

rcritten:
I've never understood the significance of single vs double quotes nor why changing it en-mass is important. I don't really care about it other than polluting git history.

stanislavlevin:
I'm for enforcing code style one or another. In this case, that was easier for me to do 'hack hack black'.
I like the minimal possible diffs too, but I've already changed this file totally. There can't be a small size diff.
This simplifies my future work here - no need to choose hunks of a patch to commit among style changes.

abbra:
I think we can merge this. LGTM.

stanislavlevin:
@abbra, thank you!
- rebased to upstream
- removed temp commit

abbra:
master:

* 3e33e546c3e4fa21e5232a0490093b17ad82515e Azure: Populate containers with self-AAAA records
* 596bb320bef19d2fd2270b9eb797a24a9f107b09 ipatests: Handle AAAA records in test_ipa_dns_systemrecords_check
* 778ef959162a98886495c5b28bb0b7942ea2e946 rpm-spec: Require crypto-policies-scripts

abbra:
Needs a manual backport to `ipa-4-8`, @stanislavlevin could you please handle that?

```
Applying to ipa-4-8: Azure: Populate containers with self-AAAA records
Failure to apply patches: error: Failed to merge in the changes.
hint: Use 'git am --show-current-patch=diff' to see the failed patch

Cleaning up

```

stanislavlevin:
@abbra, yes, sure.

stanislavlevin:
https://github.com/freeipa/freeipa/pull/5555, nice PR number :)

PR #5489, opened 2021-01-27
Better support for unreadable NSS database

rcritten:
Ignore database errors when trying to extract ipaCert on upgrade

If NSSDatabase() throws a ValueError it means we can't open it
to look for an existing ipaCert to migrate. Chances are there is
no certificate to migrate at this point in Fedora so don't let
it blow up the entire installation/upgrade. Warn the user and let
them figure it out.

We have no real path forward on this and by proceeding it could
lead to more errors (like no RA) but it is extremely unlikely and
would require a user to upgrade from very old Fedora to very
new Fedora in one step.

https://pagure.io/freeipa/issue/8675

I added no specific test for this because it a corner case. This code is just to mitigate the problem and make it more understandable.

### DIFF
--- a/ipapython/certdb.py
+++ b/ipapython/certdb.py
@@ -265,7 +265,8 @@ class NSSDatabase:
 
         if dbtype == "dbm" and not nss_supports_dbm():
             raise ValueError(
-                "NSS is built without support of the legacy database(DBM)"
+                f"NSS is built without support of the legacy database(DBM) "
+                f"directory '{nssdir}'",
             )
 
         if nssdir is None:
--- a/ipaserver/install/plugins/update_ra_cert_store.py
+++ b/ipaserver/install/plugins/update_ra_cert_store.py
@@ -33,7 +33,14 @@ class update_ra_cert_store(Updater):
         if not ca_enabled:
             return False, []
 
-        certdb = NSSDatabase(nssdir=paths.HTTPD_ALIAS_DIR)
+        try:
+            certdb = NSSDatabase(nssdir=paths.HTTPD_ALIAS_DIR)
+        except ValueError as e:
+            logger.warning("Problem opening NSS database in "
+                           "%s. Skipping check for existing RA "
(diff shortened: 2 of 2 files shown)


ipapython/certdb.py
```diff
@@ -265,7 +265,8 @@ def __init__(self, nssdir=None, dbtype='auto'):
 
         if dbtype == "dbm" and not nss_supports_dbm():
             raise ValueError(
-                "NSS is built without support of the legacy database(DBM)"
+                f"NSS is built without support of the legacy database(DBM) "
+                "directory '{nssdir}'",
```
stanislavlevin, line 269: The second string should be f-string, otherwise no interpolation occurs.

rcritten: Hmm, I must have pushed the wrong version. I changed this from using interpolation locally. I'll push again.

rcritten: Fixed, thanks.

tiran (approved):

flo-renaud:
master:

* 693ac709644483c8d3a3b3f00ff18937f5229723 Report the NSS database directory if it cannot be opened
* 43222cb85513d70363e721afb3dd43e0a939c7d9 Ignore database errors when trying to extract ipaCert on upgrade

stanislavlevin:
Test `ipatests/test_ipapython/test_certdb.py::test_dbm_raise` checks the error value:
```
 71 @pytest.mark.skipif(                                                            
 72     nss_supports_dbm(),                                                         
 73     reason="NSS is built with support of the legacy database(DBM)",             
 74 )                                                                               
 75 def test_dbm_raise():                                                           
 76     with pytest.raises(ValueError) as e:                                        
 77         NSSDatabase(dbtype="dbm")                                               
 78     assert (                                                                    
 79         str(e.value) == "NSS is built without support of the legacy "           
 80         "database(DBM)"                                                         
 81     )
```
This test will fail on F33+.

PR #5493, opened 2021-01-28
Only build the UI with uglifyjs on RHEL 8

rcritten:
The previous expression tested for RHEL or RHEL-like
systems to use uglifyjs. Tighten that up to only RHEL 8
so future RHEL can use rjsmin.


rcritten:
master:

* 43f344b931db3f72f50e1620443be9f21623e29a Only build the UI with uglifyjs on RHEL 8

PR #5495, opened 2021-01-29
Ensure that KDC cert has SAN DNS entry

tiran:
The dns parameter of request_and_wait_for_cert() must be a string of
hostnames.

* Enforce list/tuple type so that API misuse no longer passes silently.
* Add commonNameToSANDefaultImpl to KDCs_PKINIT_Certs profile
* Explicitly pass hostname for service certs

Fixes: https://pagure.io/freeipa/issue/8685


rcritten:
Looks good.

rcritten:
master:

* 24a5d4d06ba496a936b05d89081a4df0a03708b2 Ensure that KDC cert has SAN DNS entry

PR #5496, opened 2021-01-29
Fix cert_request for KDC cert

tiran:
ca_kdc_check() expects an API object, not an LDAP connection. Issue was
introduced in commit 8f4abf7bc1607fc44f528b8a443b69cb82269e69.

See: https://pagure.io/freeipa/issue/6739
Fixes: https://pagure.io/freeipa/issue/8686


rcritten:
LGTM, this is the only other occurrence of the call that I can find.

rcritten:
master:

* 98a88bd71976919d52e5806daf887e42ef3eb320 Fix cert_request for KDC cert

PR #5505, opened 2021-02-01
Don't stop certmonger during IPA uninstallation

rcritten:
An effort prevent CI failures.

https://pagure.io/freeipa/issue/8533

### DIFF
--- a/ipaserver/install/ca.py
+++ b/ipaserver/install/ca.py
@@ -425,7 +425,7 @@ def install_step_1(standalone, replica_config, options, custodia):
 
 def uninstall():
     ca_instance = cainstance.CAInstance(api.env.realm)
-    ca_instance.stop_tracking_certificates(stop_certmonger=False)
+    ca_instance.stop_tracking_certificates()
     ipautil.remove_file(paths.RA_AGENT_PEM)
     ipautil.remove_file(paths.RA_AGENT_KEY)
     if ca_instance.is_configured():
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -1075,13 +1075,13 @@ class CAInstance(DogtagInstance):
             logger.error(
                 "certmonger failed to start tracking certificate: %s", e)
 
-    def stop_tracking_certificates(self, stop_certmonger=True):
+    def stop_tracking_certificates(self):
         """
         Stop tracking our certificates. Called on uninstall.  Also called
         during upgrade to fix discrepancies.
 
         """
-        super(CAInstance, self).stop_tracking_certificates(False)
+        super(CAInstance, self).stop_tracking_certificates()
 
         # stop tracking lightweight CA signing certs
         for request_id in certmonger.get_requests_for_dir(self.nss_db):
@@ -1094,9 +1094,6 @@ class CAInstance(DogtagInstance):
         except RuntimeError as e:
             logger.error(
                 "certmonger failed to stop tracking certificate: %s", e)
-
-        if stop_certmonger:
-            services.knownservices.certmonger.stop()
 
     def is_renewal_master(self, fqdn=None):
         if fqdn is None:
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -453,7 +453,7 @@ class DogtagInstance(service.Service):
                 logger.error(
                     "certmonger failed to start tracking certificate: %s", e)
 
-    def stop_tracking_certificates(self, stop_certmonger=True):
+    def stop_tracking_certificates(self):
         """
         Stop tracking our certificates. Called on uninstall.  Also called
         during upgrade to fix discrepancies.
@@ -476,9 +476,6 @@ class DogtagInstance(service.Service):
             except RuntimeError as e:
                 logger.error(
                     "certmonger failed to stop tracking certificate: %s", e)
-
-        if stop_certmonger:
-            cmonger.stop()
 
     def update_cert_cs_cfg(self, directive, cert):
         """
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -648,9 +648,9 @@ def certificate_renewal_update(ca, kra, ds, http):
 
     # Ok, now we need to stop tracking, then we can start tracking them
     # again with new configuration:
-    ca.stop_tracking_certificates(stop_certmonger=False)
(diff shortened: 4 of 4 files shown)


rcritten:
It passed, re-running.

rcritten:
Passed five times, re-running.

rcritten:
@flo-renaud do you think there is a better test to exercise this re-ordering of the uninstall?

flo-renaud:
@rcritten test_cert_fix and test_replica_promotion_TestHiddenReplicaPromotion are probably the jobs that fail the most frequently. But I think the right ticket is https://pagure.io/freeipa/issue/8506 for this issue (8533 happens during replica installation while configuring the cert tracking).

rcritten:
Thanks for the suggestion, updating the test. I'll have to remember to go back to fix the ticket number.

rcritten:
Both invocations of test_integration/test_replica_promotion.py::TestHiddenReplicaPromotion failed due to DNS issues. Re-running.

rcritten:
Two more failures, unrelated to dbus.

rcritten:
Try test_integration/test_caless.py::TestReplicaCALessToCAFull to try to provoke failure. If I can't maybe I'll drop the change to see if I can more easily reproduce this and then add the change back in to see if it actually did anything.

rcritten:
Re-based and dropped change to see if one or more of the tests fail.

rcritten:
temp_commit_2 reproduced the uninstall dbus error:

[ipatests.pytest_ipa.integration.host.Host.master.cmd54] org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.

rcritten:
All five executions of test_ipa_cert_fix.py failed with the dbus error and the re-organized code change.

rcritten:
I think I may have it worked out but I've heavily redone the patch so we'll see. Originally I was trying to move the certmonger CA removal ahead of shutting down IPA. I've dropped that and went with the original code in cainstance.py but wrapped it with a loop.

dbus is socket activated and I think that sometimes it takes longer than certmonger is willing to wait to start up. So stick it into a loop and try up to 5 times.

In testing it never took more than 2 tries with no sleep.

rcritten:
One of the dbus failures related to stop tracking of the CA certs in certmonger. We got a DBus timeout. Upon retrying sometimes we would also get a no_such_entry from certmonger which is very odd. It basically means that we asked for a request from certmonger and it returned one, but the path to that request is not found upon requesting stop tracking. I added code to skip in that case. We should see in the uninstall log if there are remaining untracked certs.

rcritten:
What I observed in the last set of logs is the request to stop tracking failed with a DBus timeout but the request was actually successful. Tune up the logging around this so it is more clear what went on.

We would see a message like "Some certificates may still be tracked by certmonger." in the server uninstall log if there were remaining certificates. I haven't seen this in these dbus timeout failure cases.

rcritten:
For the reviewer:

There are two categories of errors here:

1. Stop tracking the CA certificates in certmonger faisl with a DBus.Error.NoReply.

A handled failure looks like:

024-02-11T15:50:57Z DEBUG Attempt 1 to stop tracking for 'auditSigningCert cert-pki-ca in /etc/pki/pki-tomcat/alias'
2024-02-11T15:51:22Z DEBUG org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
2024-02-11T15:51:22Z DEBUG Attempt 2 to stop tracking for 'auditSigningCert cert-pki-ca in /etc/pki/pki-tomcat/alias'
2024-02-11T15:51:43Z DEBUG org.fedorahosted.certmonger.no_such_entry: No matching entry found.
2024-02-11T15:51:43Z DEBUG Tracking no longer exists, stop trying
2024-02-11T15:51:43Z DEBUG Attempt 1 to stop tracking for 'ocspSigningCert cert-pki-ca in /etc/pki/pki-tomcat/alias'
2024-02-11T15:51:43Z DEBUG stop tracking successful

What you see here is the first attempt failing, it being caught, and a second attempt only to find out that the path doesn't exist. This means that the original request was successful but not communicated back.

A success on the first try looks like:

2021-02-10T16:02:37Z DEBUG Attempt 1 to stop tracking for /var/kerberos/krb5kdc/kdc.crt
2021-02-10T16:02:37Z DEBUG stop tracking successful

I tend to see this failure in test_integration-test_ipa_cert_fix.py-TestIpaCertFix-uninstall. You'll find the entries in the uninstall log.

The other error happens when trying to remove the IPA-created CA's in certmonger. Similarly we don't get a response back but the retry succeeds without a not found error.

A success looks like:
2024-02-11T15:51:45Z DEBUG Attempt 1 to remove certmonger CAs
2024-02-11T15:51:45Z DEBUG certmonger CA removal successful

A handled failure looks like:
2021-02-10T16:02:06Z DEBUG Attempt 1 to remove certmonger CAs
2021-02-10T16:02:31Z DEBUG org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
2021-02-10T16:02:31Z DEBUG Attempt 2 to remove certmonger CAs
2021-02-10T16:02:32Z DEBUG certmonger CA removal successful

I tend to see this most in test_integration-test_caless.py-TestReplicaCALessToCAFull-uninstall on the replica in the uninstall log.

rcritten:
I'm going to try to generalize this test loop. https://pagure.io/freeipa/issue/8470 shows this dbus error also on a fresh installation.

rcritten:
So. Maybe it's been luck, but I made another change so that certmonger is not restarted during uninstall and that seems to have helped. I just ran a successful set with just that change and passed. I'll kick it again once Azure is done.

rcritten:
Dropping WIP label, this seems to have fixed the uninstallation dbus-related issue.

abbra:
LGTM. Please remove temp commit.

rcritten:
Dropped the temp commits.

abbra:
ACK.

abbra:
master:

* 71047f68b55ab59d4615c2601b702ec4a84bce4f Remove the option stop_certmonger from stop_tracking_*

PR #5515, opened 2021-02-03
client: synchronize ignored return codes with ipa-rmkeytab

abbra:
Refactoring ipa-rmkeytab with commit
f3f9672d527008dc741ac90aa465bac842eea08d led to new error code 7 when
MIT Kerberos fails to iterate through the keys. It appears now in places
where in past error code 3 was returned.

Related: https://pagure.io/freeipa/issue/8658


rcritten:
master:

* 3231c6fc790a5631b794508d57bcb790117e70ac client: synchronize ignored return codes with ipa-rmkeytab

PR #5516, opened 2021-02-03
configure: ipaplatform falls back to ID_LIKE

tiran:
The auto-detection code for IPAPLATFORM now falls back to ID_LIKE.

CentOS platform will now be treated as rhel-like, Ubuntu platforms as
debian-like automatically.

Fixes: https://pagure.io/freeipa/issue/8689

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -341,19 +341,33 @@ AC_ARG_WITH([ipaplatform],
 	    [IPAPLATFORM=""])
 AC_MSG_CHECKING([supported IPA platform])
 
-if test "x${IPAPLATFORM}" == "x"; then
+if test "x${IPAPLATFORM}" != "x"; then
+	if test ! -d "${srcdir}/ipaplatform/${IPAPLATFORM}"; then
+		AC_MSG_ERROR([IPA platform ${IPAPLATFORM} is not supported])
+	fi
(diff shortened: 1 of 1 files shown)


rcritten:
Tested auto-detection on:

- [x] Fedora
- [x] CentOS Stream

Tested --with-platform with existing platforms
- [x] fedora
- [x] debian

Tested --with-platform without existing platforms (fails with unsupported platform error)
- [x] distro_not_found
- [x] ubuntu
- [x] centos

Works well.

The failure message looks like:

checking supported IPA platform... configure: error: IPA platform centos is not supported

I don't want to over think this but could this be misinterpreted?

I don't know if we want to duplicate the effort of ID_LIKE in our own platform system...

rcritten:
@tiran I'm not demanding changes, just pausing to see if we should handle this other case. It otherwise LGTM.

abbra:
LGTM. I don't think we need to put more into an error message. Should be good?

rcritten:
Sure, we can address it if it comes up.

rcritten:
master:

* 5ca918943c4614f17085a55239fdfd4796a1cb16 configure: ipaplatform falls back to ID_LIKE

PR #5460, opened 2021-01-21
ipatests: use pexpect for controlling interactive command line utilities

wladich:
This PR adds a tests-oriented wrapper for pexpect module
  
The pexpect module can be used for controlling and testing interactive command-line programs. The wrapper adds testing-oriented features like logging and automatic process termination and default check for process exit status.

Invocation of `ktutil` on systems with old version of `readline` is fixed using `pexpect`

Few tests in `test_adtrust_install` were refactored to utilize `pexpect` to improve the tests and provide usage example.

PR #5504, opened 2021-02-01
Test on Fedora 33

tiran:
389-DS on Fedora 32 does not have new DS feature.

Update to temmplate 0.0.5 with updated DS.

Signed-off-by: Christian Heimes <cheimes@redhat.com>

PR #5520, opened 2021-02-03
Fix pylint issues with pexpect wrapper

tiran:
Also move dependencies to extras_require.

See: https://pagure.io/freeipa/issue/8690
See: https://github.com/freeipa/freeipa/pull/5460

### DIFF
--- a/ipatests/pytest_ipa/integration/expect.py
+++ b/ipatests/pytest_ipa/integration/expect.py
@@ -107,18 +107,44 @@ class IpaTestExpect(pexpect.spawn):
         logger.debug('Sending %r', s)
         return super().send(s)
 
-    def expect_list(self, pattern_list, *args, **kwargs):
+    def expect_list(
+        self,
+        pattern_list,
+        timeout=-1,
+        searchwindowsize=-1,
+        async_=False,
+        **kw
+    ):
         """Wrapper to provide logging output string and expected patterns"""
         try:
-            result = super().expect_list(pattern_list, *args, **kwargs)
+            result = super().expect_list(
+                pattern_list,
+                timeout=timeout,
+                searchwindowsize=searchwindowsize,
+                async_=async_,
+                **kw
+            )
         finally:
             self._log_output(pattern_list)
         return result
 
-    def expect_exact(self, pattern_list, *args, **kwargs):
+    def expect_exact(
+        self,
+        pattern_list,
+        timeout=-1,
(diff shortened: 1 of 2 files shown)


ipatests/setup.py
```diff
@@ -73,10 +73,14 @@
             "pytest_multihost",
             "python-ldap",
             "six",
-            "pexpect",
         ],
         extras_require={
-            "integration": ["dbus-python", "pyyaml", "ipaserver"],
+            "integration": [
+                "dbus-python",
+                "pexpect",
```
wladich, line 80: I really was not sure where to put this dependency, can you please clarify why it should be in extras?
For me it looks same as pytest_multihost - it is required only by integration testst and those tests can not even start without it.

ipatests/pytest_ipa/integration/expect.py
```diff
@@ -107,18 +107,44 @@ def send(self, s):
         logger.debug('Sending %r', s)
         return super().send(s)
 
-    def expect_list(self, pattern_list, *args, **kwargs):
+    def expect_list(
```
wladich, line 110: If the signature of original function changes we will need to
a) somehow notice it
b) change the signature in this code.

Also duplicating and passing all parameters looks too verbose compared to `*args, **kwargs` approach.

Is not it better to temporary disable the failing rule in pylintrc or add a `# pylint: disable=W0222` comment on single line?

ipatests/setup.py
tiran: The dependency should not be needed by client-only tests, only for integration tests. In theory any dependency that is only used by ``ipatests.pytest_ipa.integration`` should be in ``extras_require`` for ``integration`` (hence the name).

Something changed since I last touched the code. The PR is failing because something triggers an important of ``ipatests.pytest_ipa.integration``.

wladich: Sounds reasonable. But should not pytest_multihost also be moved to extras?

tiran:
I don't insist on my way to address the issue, I just need a fix for Fedora 33 ASAP.

The lack of Fedora 33 support in PR-CI has been blocking container work for weeks. I *just* figured out how to solve the last DNS-related issue today (https://github.com/freeipa/freeipa/pull/5504), when the merge of https://github.com/freeipa/freeipa/pull/5460 broke Fedora 33. It feels like I'm playing whack-a-mole and loosing constantly...

It's annoying :(

PR #5528, opened 2021-02-05
ipatests: ipactl status now exits with 3 when a service is stopped

flo-renaud:
ipatests: ipactl status now exits with 3 when a service is stopped
    
Some tests are individually stopping a service and call
ipactl status to ensure it is stopped. They need to use
run_command with raiseonerr=False as ipactl status now
exits with 3 when one of the IPA services is down
(since commit 928ab51).
    
Related: https://pagure.io/freeipa/issue/8588

### DIFF
--- a/ipatests/test_integration/test_replica_promotion.py
+++ b/ipatests/test_integration/test_replica_promotion.py
@@ -702,7 +702,7 @@ class TestReplicaInstallCustodia(IntegrationTest):
         tasks.install_replica(master, replica1, setup_ca=False)
         replica1.run_command(['ipactl', 'status'])
         replica1.run_command(['systemctl', 'stop', 'ipa-custodia'])
-        replica1.run_command(['ipactl', 'status'])
+        replica1.run_command(['ipactl', 'status'], raiseonerr=False)
 
         # Install Replica2 with CA with source as Replica1.
         tasks.install_replica(replica1, replica2, setup_ca=True)
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -1481,7 +1481,7 @@ class TestInstallReplicaAgainstSpecificServer(IntegrationTest):
         self.replicas[0].run_command('systemctl stop ipa-custodia.service')
 
         # check if custodia service is stopped
-        cmd = self.replicas[0].run_command('ipactl status')
+        cmd = self.replicas[0].run_command('ipactl status', raiseonerr=False)
         assert 'ipa-custodia Service: STOPPED' in cmd.stdout_text
 
         try:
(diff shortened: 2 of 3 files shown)


flo-renaud:
The test failures can be seen in PR #[692](https://github.com/freeipa-pr-ci2/freeipa/pull/692) [testing_master_previous]:
- test_installation_TestInstallReplicaAgainstSpecificServer: [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/9c79f2f2-674b-11eb-9369-fa163e5185d3/report.html)
- test_replica_promotion_TestReplicaInstallCustodia: [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/d7a4c208-6750-11eb-a776-fa163e6c3b05/report.html)

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -702,7 +702,7 @@ def test_replica_install_for_custodia(self):
         tasks.install_replica(master, replica1, setup_ca=False)
         replica1.run_command(['ipactl', 'status'])
         replica1.run_command(['systemctl', 'stop', 'ipa-custodia'])
-        replica1.run_command(['ipactl', 'status'])
+        replica1.run_command(['ipactl', 'status'], raiseonerr=False)
```
rcritten, line 706: I realize you're just fixing the failure but I'm having a hard time understanding the purpose of this test.

Ostensibly it is testing that keys can be retrieved according to https://github.com/freeipa/freeipa/commit/dcaa62f6a4e0a57de9d0affda584a27539bc4a36

But since this server doesn't have a CA the replica installer will pick the first server as the ca_host and completely avoid the second one. I guess this is reinforced by the fact that custodia is stopped and the install is still successful, but that is a pretty subtle difference.

Either way the ipactl status on lines 703 and 705 aren't really needed and don't really do anything other than check that 1) after an IPA installation all services are running and 2) ipa-custodia is stoppable.

flo-renaud: IIRC during replica installation, the replica could be talking to custodia on 2 different masters if the selected master did not host a CA instance. The test wanted to make sure that all communications went to the same custodia instance.
I agree with you that the ipactl status lines aren't needed, they are just helpful for debugging. I can remove them if you prefer.

rcritten: Yeah, I can see it being helpful during troubleshooting. In that light it's fine with your change.

flo-renaud:
I added a new fix in line [489](https://github.com/freeipa/freeipa/pull/5528/files#diff-a03c8613b33ba79f0d7ba6db263a495631e47bc2e730adecda3e52e0b6aa1993R489) of test_ipahealthcheck.py : same issue with a call to ipactl status when one service is stopped.
=> Removing the ACK and waiting for someone to review this new diff.

fcami (approved):
ACK, the new change makes sense, thanks for the catch.

flo-renaud:
master:

* 610d542c0eba50890aac9dad79d283828644597c ipatests: ipactl status now exits with 3 when a service is stopped

PR #5531, opened 2021-02-08
Set the ACME baseURL in order to pin a client to a single IPA server

rcritten:
ACME uses nonce values to prevent replay attacks. Since the
ipa-ca name can go to any of the IPA servers in order to verify the
nonce the servers need to know the value that was set which
relies on replication. Sometimes the client is faster than
replication so a request can fail.

This change returns the baseURL to the client as the name of the
ACME server during discovery which should pin all requests to this
one IPA server and alleviate the replication issue.

### DIFF
--- a/install/share/pki-acme-configsources.conf.template
+++ b/install/share/pki-acme-configsources.conf.template
@@ -1,2 +1,3 @@
+# VERSION 2 - DO NOT REMOVE THIS LINE
 engine.class=org.dogtagpki.acme.server.ACMEEngineConfigFileSource
 engine.filename=/etc/pki/pki-tomcat/acme/engine.conf
--- a/install/share/pki-acme-database.conf.template
+++ b/install/share/pki-acme-database.conf.template
@@ -1,3 +1,4 @@
+# VERSION 2 - DO NOT REMOVE THIS LINE
 class=org.dogtagpki.acme.database.LDAPDatabase
 basedn=ou=acme,o=ipaca
 configFile=/etc/pki/pki-tomcat/ca/CS.cfg
--- a/install/share/pki-acme-engine.conf.template
+++ b/install/share/pki-acme-engine.conf.template
@@ -9,3 +9,5 @@ enabled=false
 
 # Whether to accept wildcard DNS identifiers:
 policy.wildcard=false
+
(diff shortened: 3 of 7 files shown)


mrizwan93 (commented):
Could you trigger acme test with this change?

flo-renaud:
Hi @rcritten,

the template is propely used during ipa server installation and the $FQDN replaced with the server name in `/etc/pki/pki-tomcat/acme/engine.conf`.
The patch does not handle upgrade, is it on purpose?

rcritten:
Oh yeah, this is a design oversight. We probably need to version all of these files and re-apply on update if necessary like we do with other configuration files. A most excellent catch.

flo-renaud (commented):
Hi @rcritten 
thanks for the updated patch.
- [x] upgrade from 4.9.1-1 properly adds the new setting in `/etc/pki/pki-tomcat/acme/engine.conf`
- [x] new install properly adds the setting
- [x]  certbot --standalone work. Moreover we can see in /var/log/letsencrypt/letsencrypt.log that the acme server name ipa-ca.$DOMAIN is replaced with the actual server name in subsequent POST requests.

flo-renaud:
Note: I did my tests with pki-server taken from @pki/master repo. Do we need to bump pki version to get compatibility with the new `baseURL` setting in `/etc/pki/pki-tomcat/acme/engine.conf`?

flo-renaud:
@rcritten you can go ahead and remove the temp commit. The temp commit was run with an older pki version (10.10.2, see [installed pkgs](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/39ded7d8-6c6f-11eb-bb39-fa163ec3b944/installed_packages/installed_packages_master.log.gz)) and it didn't break.

flo-renaud:
ACK for when PRCI completes.

abbra:
master:

* d2d487bcccf6365ad4302106e13d9588722f7b33 Set the ACME baseURL in order to pin a client to a single IPA server
* b1e72cb12e3b0357edfb37652e564529528271a9 Add versions to the ACME config templates and update on upgrade
* 3d2d06719fccd5d42207acceb686edea392b1899 Add some logging around initial ACME deployment

PR #5532, opened 2021-02-08
ipa-cacert-manage: add prune option

antoniotorresm:
Add prune option to `ipa-cacert-manage`, allowing
to remove all expired certificates from the certificate store.

Related: https://pagure.io/freeipa/issue/7404

### DIFF
--- a/install/tools/man/ipa-cacert-manage.1
+++ b/install/tools/man/ipa-cacert-manage.1
@@ -27,6 +27,8 @@ ipa\-cacert\-manage \- Manage CA certificates in IPA
 \fBipa\-cacert\-manage\fR [\fIOPTIONS\fR...] delete \fINICKNAME\fR
 .br
 \fBipa\-cacert\-manage\fR [\fIOPTIONS\fR...] list
+.br
+\fBipa\-cacert\-manage\fR [\fIOPTIONS\fR...] prune
 .SH "DESCRIPTION"
 \fBipa\-cacert\-manage\fR can be used to manage CA certificates in IPA.
 .SH "COMMANDS"
@@ -71,6 +73,13 @@ Please do not forget to run ipa-certupdate on the master, all the replicas and a
 .sp
 .RS
 Display a list of the nicknames or subjects of the CA certificates that have been installed.
+.RE
+.TP
+\fBprune\fR
+\- Prune the stored CA certificates
+.sp
+.RS
+Removes installed CA certificates that are expired.
 .RE
 .SH "COMMON OPTIONS"
 .TP
--- a/ipaserver/install/ipa_cacert_manage.py
+++ b/ipaserver/install/ipa_cacert_manage.py
@@ -19,6 +19,7 @@
 
 from __future__ import print_function, absolute_import
 
+import datetime
 import logging
 import os
 from optparse import OptionGroup  # pylint: disable=deprecated-module
@@ -46,7 +47,7 @@ class CACertManage(admintool.AdminTool):
     command_name = 'ipa-cacert-manage'
 
     usage = "%prog renew [options]\n%prog install [options] CERTFILE\n" \
-            "%prog delete [options] NICKNAME\n%prog list"
+            "%prog delete [options] NICKNAME\n%prog list\n%prog prune"
 
     description = "Manage CA certificates."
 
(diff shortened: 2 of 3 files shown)


ipaserver/install/ipa_cacert_manage.py
```diff
@@ -145,6 +148,8 @@ def run(self):
                 return self.list()
             elif command == 'delete':
                 return self.delete()
+            elif command == 'prune':
+                return self.prune()
```
rcritten, line 152: I wonder if we have enough commands that than passing we do:
```
if command not in ('renew', 'install', ...):
    parser.error("unknown command \"%s\"" % command)
```

and then do the additional argument enforcing for 'install' and 'delete'.

ipatests/test_integration/test_commands.py
```diff
@@ -1143,6 +1143,29 @@ def test_cacert_manage(self):
...
+
+        # Jump in time to make sure the cert is expired
+        self.master.run_command(['date', '-s', '+15Years'])
+        result = self.master.run_command(
+            [paths.IPA_CACERT_MANAGE, 'prune'], raiseonerr=False
+        ).stdout_text
+        self.master.run_command(['date', '-s', '-15Years'])
+
+        assert isrgrootx1_nick in result
+
```
rcritten, line 1168: I'd try either resetting time back to present or moving this test to the end of the file since it is twiddling with time.

ipaserver/install/ipa_cacert_manage.py
antoniotorresm: OK, I've just updated that. We could also remove all the if's with function calls by using `getattr(self, command)` but I'm not sure if that is safe enough.

ipatests/test_integration/test_commands.py
antoniotorresm: The time is resetted when calling `date -s -15years`. However I agree moving it to the end is probably safer.

install/tools/man/ipa-cacert-manage.1
```diff
@@ -72,6 +74,13 @@ Please do not forget to run ipa-certupdate on the master, all the replicas and a
 .RS
 Display a list of the nicknames or subjects of the CA certificates that have been installed.
 .RE
+.TP
+\fBprune\fR
+\- Prune the stored CA certificates
+.sp
+.RS
+Removes installed CA certificates that are not longer valid.
```
rcritten, line 82: I think it would be clearer to say that it removes CA's that are expired. "not longer valid" could include other reasons like revocation.

ipaserver/install/ipa_cacert_manage.py
```diff
@@ -546,6 +543,31 @@ def delete(self):
                 conn.delete_entry(dn)
                 return
 
+    def delete(self):
+        options = self.options
+        nickname = self.args[1]
+        self._delete_by_nickname(nickname, options)
```
rcritten, line 549: I'd just use self.options here like you did in prune

ipaserver/install/ipa_cacert_manage.py
```diff
@@ -546,6 +543,31 @@ def delete(self):
...
+            if ca_cert.not_valid_after < now:
+                expired_certs.append(ca_nickname)
+                self._delete_by_nickname(ca_nickname, self.options)
+
+        if expired_certs:
+            print("Expired certificates deleted:")
+            for nickname in expired_certs:
+                print(nickname)
+        else:
+            print("No certificates were deleted")
```
rcritten, line 569: Should we suggest users run ipa-certupdate on all enrolled machines afterward to apply the removed certs?

rcritten:
LGTM

rcritten:
The question is what branches do we want to backport this to?

abbra:
Down to ipa-4-6, I think?

rcritten:
> Down to ipa-4-6, I think?

The oldest branch to have the delete option is ipa-4-8 so that's as far back as we can go unless we want to backport list and delete as well to ipa-4-6. IIRC that code is pretty-well centralized so it may be doable but given this is effectively a new feature I don't see it landing in any releases that actually use 4.6 as a base.

I think this is safe to just backport to ipa-4-9.

abbra:
Ok, ipa-4-9 then.

rcritten:
I'm going to delay the merge on this. It is effectively a new feature.

rcritten:
master:

* 5d8cb1dd1fbfbdb7feb7f33c60eb6066f48aa563 ipa-cacert-manage: add prune option
* 8a2e6ec3202ece070e6d9f529727ea89cec460f1 ipatests: add test for ipa-cacert-manage prune

PR #5534, opened 2021-02-08
Add RHEL 9 UI branding patch reference

rcritten:
The UI in RHEL has a different set of logos and different
background colors. Some direct adjustments were made that
are not buildable so apply them as a patch.


abbra:
LGTM

rcritten:
master:

* bf06579895733c4a5fa39520815b31ea6c2f5c0a Add RHEL 9 UI branding patch reference

PR #5538, opened 2021-02-10
 sudocmd: ensure command doesn't contain trailing dot before adding it

antoniotorresm:
Trailing dots aren't permitted in sudo commands, as
enforced explicitly in `get_dn`. Performing this check
before adding the command prevents the user from
entering invalid commands, which would otherwise trigger
errors when accessing them afterwards.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1925410

### DIFF
--- a/ipaserver/plugins/sudocmd.py
+++ b/ipaserver/plugins/sudocmd.py
@@ -49,6 +49,12 @@ register = Registry()
 
 topic = 'sudo'
 
+
+def command_validator(ugettext, value):
+    if value.endswith('.'):
+        return _('must not contain trailing dot: %s') % value
+    return None
+
 @register()
 class sudocmd(LDAPObject):
     """
@@ -112,7 +118,7 @@ class sudocmd(LDAPObject):
     label_singular = _('Sudo Command')
 
     takes_params = (
-        Str('sudocmd',
+        Str('sudocmd', command_validator,
             cli_name='command',
             label=_('Sudo Command'),
             primary_key=True,
@@ -146,7 +152,6 @@ class sudocmd_add(LDAPCreate):
     __doc__ = _('Create new Sudo Command.')
 
     msg_summary = _('Added Sudo Command "%(value)s"')
-
 
 @register()
 class sudocmd_del(LDAPDelete):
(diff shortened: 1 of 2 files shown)


antoniotorresm:
I'm not sure if we should add more checks for validity, but the trailing dot is the only being explicitly enforced in the `get_dn` function.

abbra:
I think it is better to use a validator callback instead of `pre-` callback here. Just define a validator for `Str('cn' ...)` parameter in `sudocmd` class that checks for a presence of the `.` at the end of the parameter. See `hostname_validator` in `ipalib/util.py` for an example (and search for the same validator in `ipaserver/plugins`.

mrizwan93 (commented):
@antoniotorresm Hello, Please consider adding test with this change i.e adding sudo-cmd having trailing space throws an error.

rcritten:
Such a test would probably best go into the xmlrpc test directory.

antoniotorresm:
I have added the test to `ipatests/test_integration/test_sudo.py` since there was already a method for testing command additions.

abbra:
LGTM, the test is correct too.
http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/491eeb0e-6c86-11eb-bd27-fa163e9d1aeb/report.html (`test_add_sudo_commands`):
```
INFO     ipatests.pytest_ipa.integration.host.Host.master.IPAOpenSSHTransport:transport.py:391 RUN ['ipa', 'sudocmd-add', '/bin/cat.']
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd85:transport.py:513 RUN ['ipa', 'sudocmd-add', '/bin/cat.']
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd85:transport.py:557 ipa: ERROR: invalid 'command': must not contain trailing dot: /bin/cat.
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd85:transport.py:217 Exit code: 1
```

Please remove the temp commit and add the release note to the commit message:

```
RN: ipa sudocmd-* commands now validate SUDO command name to not end with a dot.
RN: Previously a trailing dot was stripped away in when addressing a SUDO command's LDAP object. 
RN: As a result, a SUDO command was created but it was not possible to refer to it in other IPA commands.
```

abbra:
master:

* d6c5a9260940c4e132469439a7155e62553c72b7 sudocmd: ensure command doesn't contain trailing dot before adding it
* 89135830cc95e3fefb5106d4680631f782dc7baf ipatests: test addition of invalid sudo command

PR #5540, opened 2021-02-11
WebUI: change FreeIPA naming to IPA in About dialog

antoniotorresm:
As part of the effort for reducing differences between
upstream and dowmstream releases, product naming in WebUI
About dialog is changed from FreeIPA to IPA.

Related: https://pagure.io/freeipa/issue/8669

### DIFF
--- a/install/ui/src/freeipa/dialog.js
+++ b/install/ui/src/freeipa/dialog.js
@@ -1325,7 +1325,7 @@ IPA.about_dialog = function(spec) {
     spec = spec || {};
 
     spec.name = spec.name || 'version_dialog';
-    var product = 'FreeIPA';
+    var product = 'IPA';
     var version = 'Unknown';
     var msg = text.get('@i18n:dialogs.about_message', '${product}, version: ${version}');
     if (IPA.env) {


rcritten:
LGTM

abbra:
One note: we most likely will need to update Fedora OpenQA UI needles with these changes. Needles are screenshots used to recognize where we are in the GUI flow tests. @AdamWill FYI

AdamWill:
Thanks for the headsup, @abbra , but I don't believe we actually have a needle for that text so it'll be fine.

abbra:
@AdamWill we did recently a wholesale refactor of the name FreeIPA to IPA in all user-visible strings to unify RHEL and upstream terminology. If any needle contained FreeIPA string, that would be affected.

AdamWill:
Ah, k. Still don't think it'll be many or any, but again, thanks for letting me know, I'll keep an eye out!

rcritten:
master:

* 9b8810c88bd71f119263bcf9439f14530bd01330 WebUI: change FreeIPA naming to IPA in About dialog

PR #5544, opened 2021-02-12
ipatests: check that external names are resolvable by IPA namesever

wladich:
Related to https://pagure.io/freeipa/issue/8710


wladich:
The test is expected to fail because https://pagure.io/freeipa/issue/8710 is not resolved yet.

wladich:
The problem was caused by firewall configuration of PR-CI runners. Fixed by allowing outbound TCP connections on port 53.

PR #5562, opened 2021-02-16
ipatests: Enable certbot test on rhel

mrizwan93:
With this change, certbot test will be running on rhel.
certbot is not avilable on rhel through repository or epel.
Plan is to install certbot using pip on rhel and increase the
test coverage on rhel

### DIFF
--- a/ipatests/test_integration/test_acme.py
+++ b/ipatests/test_integration/test_acme.py
@@ -27,7 +27,7 @@ ROOT_CA = "root_ca.crt"
 # RHEL does not have certbot.  EPEL's version is broken with
 # python-cryptography-2.3; likewise recent PyPI releases.
 # So for now, on RHEL we suppress tests that use certbot.
-skip_certbot_tests = osinfo.id not in ['fedora', ]
+skip_certbot_tests = osinfo.id not in ['fedora', 'rhel']
 
 # Fedora mod_md package needs some patches before it will work.
 # RHEL version has the patches.


ipatests/test_integration/test_acme.py
```diff
@@ -114,9 +105,9 @@ def prepare_acme_client(cls):
         cls.acme_server = f'https://{acme_host}/acme/directory'
 
         # install acme client packages
-        if not skip_certbot_tests:
-            tasks.install_packages(cls.clients[0], ['certbot'])
-        if not skip_mod_md_tests:
+        if osinfo.id == 'fedora':
+            tasks.install_packages(cls.clients[0], ['certbot', 'mod_md'])
+        if osinfo.id == 'rhel':
```
rcritten, line 110: By dropping the global variables this is going to break the test on non-fedora and RHEL platforms because they will no longer be marked as skip.

I think I'd leave it as it was and add a special case here for rhel since it the package will be provided externally via pip.

rcritten:
Is there a reason to change from using a fixture to skip tests vs skipping directly within the test itself?

mrizwan93:
> Is there a reason to change from using a fixture to skip tests vs skipping directly within the test itself?

Its the same thing. I didn't realized that initially there was a fixture to skip.

ipatests/test_integration/test_acme.py
```diff
@@ -27,7 +27,7 @@
 # RHEL does not have certbot.  EPEL's version is broken with
 # python-cryptography-2.3; likewise recent PyPI releases.
 # So for now, on RHEL we suppress tests that use certbot.
-skip_certbot_tests = osinfo.id not in ['fedora', ]
+skip_certbot_tests = osinfo.id not in ['fedora', 'rhel']
```
amore17, line 30: LGTM

flo-renaud:
@mrizwan93 please remove the temp commit

flo-renaud:
master:

* 19da4a76d516fca1c8bfdf360ff4eb4356d0541e ipatests: Enable certbot test on rhel

PR #5564, opened 2021-02-16
azure: Pin Pylint

stanislavlevin:
Temporarily pin Pylint to 2.6.0, since 2.6.2 has the crash regression.

Related: https://pagure.io/freeipa/issue/8716


stanislavlevin:
The issue was related to Pylint 2.6.1 only.
No longer needed.

PR #5567, opened 2021-02-16
ipaserver: don't ignore zonemgr option on install

antoniotorresm:
Fix zonemgr option in ipaserver install being
ignored because of an incorrect condition.

Fixes: https://pagure.io/freeipa/issue/8718

### DIFF
--- a/ipaserver/install/bindinstance.py
+++ b/ipaserver/install/bindinstance.py
@@ -355,7 +355,7 @@ def add_zone(name, zonemgr=None, dns_backup=None, ns_hostname=None,
         else:
             update_policy = get_dns_forward_zone_update_policy(api.env.realm)
 
-    if zonemgr is None:
+    if not zonemgr:
         zonemgr = 'hostmaster.%s' % name
 
     if ns_hostname:
@@ -682,7 +682,7 @@ class BindInstance(service.Service):
         self.forward_policy = forward_policy
         self.reverse_zones = reverse_zones
 
-        if zonemgr is not None:
+        if not zonemgr:
             self.zonemgr = 'hostmaster.%s' % normalize_zone(self.domain)
         else:
             self.zonemgr = normalize_zonemgr(zonemgr)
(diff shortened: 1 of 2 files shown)


abbra:
Thanks for the PR. Please clone the bug to a Pagure ticket and reference it in the commit message.
Your test failed because there is already master installed by the previous test. Since all you change in that installation is a zone manager email, please merge your test into the previous test instead of creating a separate one.

ipaserver/install/bindinstance.py
```diff
@@ -682,7 +682,7 @@ def setup(self, fqdn, ip_addresses, realm_name, domain_name, forwarders,
         self.forward_policy = forward_policy
         self.reverse_zones = reverse_zones
 
-        if zonemgr is not None:
+        if zonemgr is None:
```
flo-renaud, line 685: I would rather write `if not zonemgr:` as this would also cover the case where the user provided an empty string with `--zonemgr ""`.

antoniotorresm: OK. I also changed another occurrence in a method above.

flo-renaud:
LGTM, you can remove the temp commit. For the record, successful test [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/6f2ff25a-72a0-11eb-b7e6-fa163e45d937/report.html)

Can you also fix the ipatests commit msg in order to mention the pagure ticket instead of the BZ? Thx

flo-renaud:
@antoniotorresm thanks for the update. ACK for when PRCI completes

flo-renaud:
@antoniotorresm the PR doesn't request any backport but I guess the fix should also go to ipa-4-9 branch, right? If that's the intent, please add ipa-4-9 label to mark for a backport.

flo-renaud:
master:

* 04e57e6fd42b5d21905b6a9d79deff72f10acd06 ipaserver: don't ignore zonemgr option on install
* 4fc1e15c7c7bd86355f084fd303a41e8c669402a ipatests: check that zonemgr is set correctly during server install

PR #5568, opened 2021-02-17
ipa passwd: make help for `--otp` option clearer

antoniotorresm:
Update help for the `--otp` option in `ipa passwd`
to actually explain its usage.

Fixes: https://pagure.io/freeipa/issue/8244

### DIFF
--- a/ipaserver/plugins/passwd.py
+++ b/ipaserver/plugins/passwd.py
@@ -47,10 +47,16 @@ is used. This is so the end-user is the only one who knows the password.
 The IPA password policy controls how often a password may be changed,
 what strength requirements exist, and the length of the password history.
 
+If the user authentication method is set to password+OTP, the user should
+pass the --otp option when resetting the password.
+
 EXAMPLES:
 
  To reset your own password:
    ipa passwd
+
+ To reset your own password when password+OTP is set as authentication method:
+   ipa passwd --otp
 
  To change another user's password:
    ipa passwd tuser1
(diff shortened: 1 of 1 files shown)


rcritten:
I realize you took the suggestion of the reporter but I don't really care for it myself.

I guess at least something to should be added to the top-level __doc__ for further explanation about the OTP. The OTP gets combined with the current password when the modify is passed to the LDAP server.

I think the option be something like "The OTP if the user has a token configured" or something like that.

But I could also be bike shedding.

flo-renaud:
+1 to @rcritten 's comment.
`ipa passwd` can be used in 2 different situations: either as admin to reset someone else's password or as any user changing his own password.
In the first case, the tool only prompts for the new password and the --otp option is useless:
```
# klist
...
Default principal: admin@DOMAIN.COM
...
# ipa passwd otheruser
New Password: 
Enter New Password again to verify: 
------------------------------------------
Changed password for "otheruser@DOMAIN.COM"
------------------------------------------
```

In the second case, the tool prompts for the current password and for the new password, and checks the current password is valid by doing a bind. If the user authentication method is set to password+OTP, the user needs to provide password and OTP separately, the OTP being prompted for only if the --otp option is specified:
```
# klist
...
Default principal: otpuser@DOMAIN.COM
...
# ipa passwd --otp 
Current Password: 
New Password: 
Enter New Password again to verify: 
OTP: 
------------------------------------------
Changed password for "otpuser@DOMAIN.COM"
------------------------------------------
```

The best fix would be to add explanations in the __doc__ for this command.

rcritten:
LGTM

abbra:
master:

* e820e2d1a8706b7abf051c19f532756f15449ccc ipa passwd: make help for `--otp` option clearer

PR #5569, opened 2021-02-17
Test for uninstall scenario with ipa kra

kaleemsiddiqu:
ipatests: error message check in uninstall log for KRA

This test checks that there is no error message in uninstall
log for KRA instance when IPA was installed with KRA.

related: https://pagure.io/freeipa/issue/8550

### DIFF
--- a/ipatests/test_integration/test_backup_and_restore.py
+++ b/ipatests/test_integration/test_backup_and_restore.py
@@ -451,9 +451,11 @@ class BaseBackupAndRestoreWithKRA(IntegrationTest):
 
             backup_path = tasks.get_backup_dir(self.master)
 
-            self.master.run_command(['ipa-server-install',
-                                     '--uninstall',
-                                     '-U'])
+            # check that no error message in uninstall log for KRA instance
+            cmd = self.master.run_command(['ipa-server-install',
+                                           '--uninstall',
+                                           '-U'])
+            assert "failed to uninstall KRA" not in cmd.stderr_text
 
             if reinstall:
                 tasks.install_master(self.master, setup_dns=True)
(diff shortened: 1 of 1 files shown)


rcritten:
I wonder if this can be integrated into an existing installation test so we can avoid another install + KRA.

Is this test checking https://pagure.io/freeipa/issue/8550 ?

kaleemsiddiqu:
@rcritten 
> I wonder if this can be integrated into an existing installation test so we can avoid another install + KRA.
Yes makes sense.
Move the test uninstall function in https://github.com/freeipa/freeipa/blob/master/ipatests/test_integration/test_installation.py#L613  ?
> 
> Is this test checking https://pagure.io/freeipa/issue/8550 ?
Yes, should i refer in the commit message?

flo-renaud:
IMO the best option would be to use test_integration/test_backup_and_restore.py::TestBackupReinstallRestoreWithKRA::test_full_backup_reinstall_restore_with_vault and add the check there (the issue was detected in this test).

ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -482,6 +488,21 @@ def test_full_backup_reinstall_restore_with_vault(self):
         """backup, uninstall, reinstall, restore"""
         self._full_backup_restore_with_vault(reinstall=True)
 
+    def test_no_error_message_with_uninstall_ipa_with_kra(self):
```
flo-renaud, line 491: No need to add this test as you already added the check in `_full_backup_restore_with_vault`.

ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -455,6 +455,12 @@ def _full_backup_restore_with_vault(self, reinstall=False):
                                      '--uninstall',
                                      '-U'])
 
+            # check that no error message in uninstall log for KRA instance
+            ipaserver_uninstall_log = self.master.get_file_contents(
```
flo-renaud, line 459: Instead of transferring the file from master to controller, the test can directly check the command output:
```
cmd = self.master.run_command(...)
assert "failed to uninstall KRA" not in cmd.stdout_text
```

flo-renaud (commented):
Hi @kaleemsiddiqu 
thanks for the PR, please my my inline comments below.

ipatests/test_integration/test_backup_and_restore.py
kaleemsiddiqu: Yep it makes sense as it is seen on console and logfile both.

ipatests/test_integration/test_backup_and_restore.py
kaleemsiddiqu: @flo-renaud 
Just want to make sure that after restore, this is not seen again during uninstall.

flo-renaud: ok, makes sense

flo-renaud:
LGTM, tested with 4.9.1 (which doesn't contain the fix) and the test properly reported the issue. Let's wait for PRCI to complete.

flo-renaud:
Adding ipa-4-9 label as the fix has been pushed to this branch.

flo-renaud:
master:

* 5e49910bcf015f14533cfd7a788364aba7a7cbd3 ipatests: error message check in uninstall log for KRA

PR #5570, opened 2021-02-17
ipatests: test Samba mount with NTLM authentication

wladich:
Related to https://pagure.io/freeipa/issue/8636

### DIFF
--- a/ipatests/pytest_ipa/integration/__init__.py
+++ b/ipatests/pytest_ipa/integration/__init__.py
@@ -28,12 +28,14 @@ import os
 import tempfile
 import shutil
 import re
+import functools
 
 import pytest
 from pytest_multihost import make_multihost_fixture
 
 from ipapython import ipautil
 from ipaplatform.paths import paths
+from . import fips
 from .config import Config
 from .env_config import get_global_config
 from . import tasks
@@ -478,3 +480,18 @@ def del_compat_attrs(cls):
         del cls.ad_subdomains
         del cls.ad_treedomains
     del cls.ad_domains
+
+
+def skip_if_fips(reason='Not supported in FIPS mode', host='master'):
+    if callable(reason):
+        raise TypeError('Invalid decorator usage, add "()"')
+
+    def decorator(test_method):
+        @functools.wraps(test_method)
+        def wrapper(instance, *args, **kwargs):
+            if fips.is_fips_enabled(getattr(instance, host)):
+                pytest.skip(reason)
+            else:
+                test_method(instance, *args, **kwargs)
+        return wrapper
+    return decorator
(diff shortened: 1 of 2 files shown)


abbra:
Test looks good but it needs to be wrapped with FIPS check because NTLMSSP will not work in FIPS mode. Otherwise good.

wladich:
Yes, I agree, I am just waiting for test run with FIPS to see it failing.

wladich:
Test were modified:
* added test cases for specifying domain name as a part of UPN
* decorator for skipping tests in FIPS mode is moved to `patests.pytest_ipa.integration` package

wladich:
Succesful test run for test_smb: [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/30e25b36-72c4-11eb-97a0-fa163e607052/report.html)

Removing temp commit

ipatests/pytest_ipa/integration/__init__.py
```diff
@@ -478,3 +480,18 @@ def del_compat_attrs(cls):
         del cls.ad_subdomains
         del cls.ad_treedomains
     del cls.ad_domains
+
+
+def skip_if_fips(reason='Not supported in FIPS mode', host='master'):
```
kaleemsiddiqu, line 485: @wladich 
Why we are not using pyest skipif decorator (https://docs.pytest.org/en/reorganize-docs/new-docs/user/skipping.html) with call for fips.is_fips_enabled? 
May be there could be scenario where we need to skip tests in FIPS mode due to some bug, so in that case it won't fit.

wladich: @kaleemsiddiqu 

* `skipif` requires a boolean (not a callable) as as condition. 
* `is_fips_enabled` requires `host` argument, an instance of Host
*  the decorator is executed when the test class (not the instance) is created. 
*  Host objects are created in `mh` fixture, which is called for the test class instance 

=> When decorator is executed the Host object for test machine does not yet exist. 


I do not fully get the second part of the comment. We actually have a bug here and we skip the test until it is fixed.

kaleemsiddiqu: Second part is about specific reason displayed e.g skipping due to a bz which i think can be used in the current implementation.

amore17 (changes requested):
test_file_client_path also needs to be removed.

wladich:
> test_file_client_path also needs to be removed.

No. `test_file_client_path` is a view of remote file on client side inside the mounted directory.

abbra:
ACK.

wladich:
master:

* 0b60408dab4aee3a02448d0147f35433da446c61 ipatests: test Samba mount with NTLM authentication

wladich:
Not backporting test to ipa-4-6 as it is missing `ipa-client-samba` utility whichis required by `test_smb` test suite.

PR #5572, opened 2021-02-18
Remove a remaining file used with csrgen

rcritten:
This file was missed in the first pass to remove support for
csrgen.

This was never feature complete and currently has issues and
we lack the resources to maintain it.

Drop it for now. It can be revived from git history in the
future if we see the need.


fcami (approved):
ACK for when CI passes

flo-renaud:
master:

* ab13cf52f2619d6b6300e9895882f68dbc95e87a Remove a remaining file used with csrgen

PR #5581, opened 2021-02-19
ipa-kdb: do not use OpenLDAP functions with NULL LDAP context

abbra:
Calling to ipadb_get_connection() will remove LDAP context if any error
happens. This means upper layers must always verify that LDAP context
exists after such calls.

ipadb_get_user_auth() may re-read global configuration and that may fail
and cause IPA context to have NULL LDAP context.

Fixes: https://pagure.io/freeipa/issue/8681

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb_certauth.c
+++ b/daemons/ipa-kdb/ipa_kdb_certauth.c
@@ -71,10 +71,13 @@ struct krb5_certauth_moddata_st {
     time_t valid_until;
 };
 
-void ipa_certmap_debug(void *private,
-                       const char *file, long line,
-                       const char *function,
-                       const char *format, ...)
+krb5_error_code certauth_ipakdb_initvt(krb5_context context,
+                                       int maj_ver, int min_ver,
+                                       krb5_plugin_vtable vtable);
+
+static void ipa_certmap_debug(void *private, const char *file, long line,
+                              const char *function,
+                              const char *format, ...)
 {
     va_list ap;
     char str[255] = { 0 };
@@ -354,12 +357,12 @@ static krb5_error_code ipa_certauth_authorize(krb5_context context,
      * so there is nothing more to add here. */
     auth_inds = calloc(2, sizeof(char *));
     if (auth_inds != NULL) {
-	ret = asprintf(&auth_inds[0], "pkinit");
-	if (ret != -1) {
+        ret = asprintf(&auth_inds[0], "pkinit");
+        if (ret != -1) {
             auth_inds[1] = NULL;
             *authinds_out = auth_inds;
-	} else {
-	    free(auth_inds);
+        } else {
+            free(auth_inds);
         }
     }
 
@@ -404,12 +407,12 @@ static void ipa_certauth_free_indicator(krb5_context context,
     size_t i = 0;
 
     if ((authinds == NULL) || (moddata == NULL)) {
-	return;
+        return;
     }
 
     for(i=0; authinds[i]; i++) {
-	free(authinds[i]);
-	authinds[i] = NULL;
+        free(authinds[i]);
+        authinds[i] = NULL;
     }
 
     free(authinds);
--- a/daemons/ipa-kdb/ipa_kdb_mspac.c
+++ b/daemons/ipa-kdb/ipa_kdb_mspac.c
@@ -2408,9 +2408,10 @@ void ipadb_mspac_struct_free(struct ipadb_mspac **mspac)
     *mspac = NULL;
 }
 
-krb5_error_code ipadb_adtrusts_fill_sid_blacklist(char **source_sid_blacklist,
-                                                  struct dom_sid **result_sids,
-                                                  int *result_length)
+static krb5_error_code
+ipadb_adtrusts_fill_sid_blacklist(char **source_sid_blacklist,
+                                  struct dom_sid **result_sids,
+                                  int *result_length)
 {
     int len, i;
     char **source;
@@ -2441,9 +2442,10 @@ krb5_error_code ipadb_adtrusts_fill_sid_blacklist(char **source_sid_blacklist,
     return 0;
 }
 
-krb5_error_code ipadb_adtrusts_fill_sid_blacklists(struct ipadb_adtrusts *adtrust,
-                                                   char **sid_blocklist_incoming,
-                                                   char **sid_blocklist_outgoing)
+static krb5_error_code
+ipadb_adtrusts_fill_sid_blacklists(struct ipadb_adtrusts *adtrust,
+                                   char **sid_blocklist_incoming,
+                                   char **sid_blocklist_outgoing)
 {
     krb5_error_code kerr;
 
@@ -2464,7 +2466,8 @@ krb5_error_code ipadb_adtrusts_fill_sid_blacklists(struct ipadb_adtrusts *adtrus
     return 0;
 }
 
-krb5_error_code ipadb_mspac_check_trusted_domains(struct ipadb_context *ipactx)
+static krb5_error_code
+ipadb_mspac_check_trusted_domains(struct ipadb_context *ipactx)
 {
     char *attrs[] = { NULL };
     char *filter = "(objectclass=ipaNTTrustedDomain)";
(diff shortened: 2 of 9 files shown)


abbra:
@frozencemetery this is an attempt to address infamous 'krb5kdc crash'. Could you please review it?

daemons/ipa-kdb/ipa_kdb_principals.c
```diff
@@ -610,14 +615,15 @@ static krb5_error_code ipadb_parse_ldap_entry(krb5_context kcontext,
     lcontext = ipactx->lcontext;
     if (!lcontext) {
         krb5_klog_syslog(LOG_INFO,
-                         "No LDAP connection in ipadb_parse_ldap_entry(); retrying...\n");
+                         "No LDAP connection in ipadb_parse_ldap_entry(); retrying [1]...\n");
```
frozencemetery, line 618: We don't need `lcontext` before calling `ipadb_get_user_auth()`.  So I think we can do away with this entire check, and just have the one you've added on line 635.

daemons/ipa-kdb/ipa_kdb_principals.c
```diff
@@ -626,6 +632,20 @@ static krb5_error_code ipadb_parse_ldap_entry(krb5_context kcontext,
     /* Get User Auth configuration. */
     ua = ipadb_get_user_auth(ipactx, lentry);
 
+    lcontext = ipactx->lcontext;
+    /* ipadb_get_user_auth() may call into ipadb_get_global_config() and that might fail */
```
frozencemetery, line 636: Line length.

frozencemetery (changes requested):
I agree that this fixing the traceback in the linked issue is plausible.  Two small changes inline.

abbra:
Thanks, updated the code to reduce duplication.

frozencemetery (approved):
Looks good to me, thanks!

abbra:
Looks like there is a similar problem in other places, I'll add more fixes.

flo-renaud:
FYI the tox failure is unrelated to your patch and should be fixed with PR https://github.com/freeipa/freeipa/pull/5582

frozencemetery (approved):
Makes me wonder if we should just eliminate all local variable assignments of `ipactx->lcontext` entirely.

abbra:
Yes, this is what I am looking at today. There is one remaining part -- `struct ipadb_multires` which stores lcontext internally.

daemons/ipa-kdb/ipa_kdb_principals.c
```diff
@@ -333,6 +333,11 @@ static enum ipadb_user_auth ipadb_get_user_auth(struct ipadb_context *ipactx,
     if (gcfg != NULL)
         gua = gcfg->user_auth;
 
+    /* lcontext == NULL means ipadb_get_global_config() failed to load
+     * global config and cleared the ipactx */
+    if (ipactx->lcontext == NULL)
+        return IPADB_USER_AUTH_NONE;
```
rcritten, line 339: What implications does this have? IIUC if the LDAP connection has dropped then this authentication will happen as password-only. Should it just fail outright instead of falling back to a perhaps less-secure mechanism (that may well fail)?

daemons/ipa-kdb/ipa_kdb_principals.c
```diff
@@ -607,8 +612,16 @@ static krb5_error_code ipadb_parse_ldap_entry(krb5_context kcontext,
...
-    lcontext = ipactx->lcontext;
-    if (!lcontext) {
+
+    entry->magic = KRB5_KDB_MAGIC_NUMBER;
+    entry->len = KRB5_KDB_V1_BASE_LENGTH;
+
+    /* Get User Auth configuration. */
+    ua = ipadb_get_user_auth(ipactx, lentry);
+
+    /* ipadb_get_user_auth() may call into ipadb_get_global_config()
```
rcritten, line 622: nit: ipadb_get_user_auth() always calls ipadb_get_global_config() AFAICT.

daemons/ipa-kdb/ipa_kdb_principals.c
abbra: We default to `IPADB_USER_AUTH_NONE` even in the global auth configuration. See `ipadb_parse_user_auth()` -- in case our global configuration is missing, we use `IPADB_USER_AUTH_NONE`. I think we should keep this.

daemons/ipa-kdb/ipa_kdb_principals.c
abbra: I changed the comment.

daemons/ipa-kdb/ipa_kdb_certauth.c
```diff
@@ -71,6 +71,15 @@ struct krb5_certauth_moddata_st {
...
 };
 
+void ipa_certmap_debug(void *private,
+                       const char *file, long line,
+                       const char *function,
+                       const char *format, ...);
+
+krb5_error_code certauth_ipakdb_initvt(krb5_context context,
+                                          int maj_ver, int min_ver,
+                                          krb5_plugin_vtable vtable);
```
frozencemetery, line 81: Nit: whitespace on these is wrong (possibly tabs/spaces?)

daemons/ipa-kdb/ipa_kdb_mspac.c
```diff
@@ -80,6 +80,21 @@ static char *memberof_pac_attrs[] = {
...
+void get_authz_data_types(krb5_context context, krb5_db_entry *entry,
+                          bool *_with_pac, bool *_with_pad);
+krb5_error_code ipadb_adtrusts_fill_sid_blacklist(char **source_sid_blacklist,
+                                                  struct dom_sid **result_sids,
+                                                  int *result_length);
+krb5_error_code ipadb_adtrusts_fill_sid_blacklists(struct ipadb_adtrusts *adtrust,
+                                                   char **sid_blocklist_incoming,
+                                                   char **sid_blocklist_outgoing);
+krb5_error_code ipadb_mspac_check_trusted_domains(struct ipadb_context *ipactx);
+krb5_error_code ipadb_mspac_get_trusted_domains(struct ipadb_context *ipactx);
```
frozencemetery, line 96: These functions should just be made static - I don't see a need for a prototype.

daemons/ipa-kdb/ipa_kdb_mspac.c
```diff
@@ -80,6 +80,21 @@ static char *memberof_pac_attrs[] = {
 #define AUTHZ_DATA_TYPE_PAD "PAD"
 #define AUTHZ_DATA_TYPE_NONE "NONE"
 
+krb5_error_code filter_logon_info(krb5_context context,
+                                  TALLOC_CTX *memctx,
+                                  krb5_data realm,
+                                  struct PAC_LOGON_INFO_CTR *info);
+void get_authz_data_types(krb5_context context, krb5_db_entry *entry,
```
frozencemetery, line 87: I think these should be declared in ipa_kdb_mspac_private.h, since the only reason they're not static is so that they can be used in ipa_kdb_tests.c

daemons/ipa-kdb/ipa_kdb_certauth.c
```diff
@@ -71,6 +71,15 @@ struct krb5_certauth_moddata_st {
     time_t valid_until;
 };
 
+void ipa_certmap_debug(void *private,
```
frozencemetery, line 74: Nit: this declaration can be re-wrapped.

frozencemetery (changes requested):
Some minor stuff inline on the later two commits.

frozencemetery (approved):
Thanks!

rcritten:
I'll ack once CI is done.

abbra:
@rcritten looks like CI is OK, please ack. ;)

rcritten:
master:

* 47d49aa7b795a31c559ce75e49940f5b5aac8d0f ipa-kdb: do not use OpenLDAP functions with NULL LDAP context
* 1d2a14a2d20ec2c67752d4269b2121e1f4676136 ipa-kdb: fix compiler warnings
* f996c21edb2c55d779b25fd061a1042b01fa8a2e ipa-kdb: add missing prototypes
* 3037de2820a0f4eaa4293a1dab7dcf3fde3bf92b ipa-kdb: reformat ipa_kdb_certauth
* 3fc75b15faea63c2468f70411e5604cc62bb4dce ipa-kdb: mark test functions as static

PR #5584, opened 2021-02-23
platform-python only on RHEL8

rcritten:
cherry-picked from upstream contribution.


rcritten:
Adding ack. The reviewers are myself and Alexander (who merged it upstream).

rcritten:
master:

* d11acaf4d3de39b180f7441af40577a27d7bd55c platform-python only on RHEL8

PR #5587, opened 2021-02-25
Allow multiple permitopen/permitlisten entries in SSH keys

antoniotorresm:
SSH keys allow to have multiple entries for the permitopen and permitlisten options. 
Prior to this change, only one of each could be configured.

Fixes: https://pagure.io/freeipa/issue/8423


ipatests/test_ipapython/test_ssh.py
```diff
@@ -69,6 +69,8 @@
      u'%s this is a comment' % openssh),
     (u'opt3,opt2="\tx ",opt1,opt2="\\"x " %s comment ' % openssh,
      u'opt1,opt2="\\"x ",opt3 %s comment' % openssh),
+    (u'permitopen=\"1.1.1.1:111\",permitopen=\"2.2.2.2:222\" %s' % openssh,
+     u'permitopen=\"1.1.1.1:111\",permitopen=\"2.2.2.2:222\" %s' % openssh),
```
abbra, line 73: Did you mean to use `permitlisten` in the second line?

antoniotorresm: No, that line is the expected output for the test. Before the change, only one `permitopen` entry would be returned. I guess I could add another test input/output for `permitlisten`, but it works exactly the same as `permitopen`.

abbra: Ok, understood.

Please add the `permitlisten` to make it clear we expect it working this way for both commands.

abbra:
master:

* c8b5779e288f31459c8b520e4047df7ee3b89d56 Allow multiple permitopen/permitlisten in SSH keys
* 6cd544d38e96aba529d5075990edbaa10d75d63b ipatests: add test for multiple permitopen entries in SSH keys

PR #5596, opened 2021-03-02
ipatests: use whole date when calling journalctl --since

flo-renaud:
The test test_commands.py::TestIPACommand::test_ssh_key_connection
is checking the content of the journal using journalctl --since ...
but provides only the time, not the whole date with year-month-day.
As a consequence, if the test is executed around midnight it may
find nothing in the journal because it's looking for logs after 11:50PM,
which is a date in the future.

The fix provides a complete date with year-month-day hours:min:sec.

Fixes: https://pagure.io/freeipa/issue/8728

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -642,7 +642,8 @@ class TestIPACommand(IntegrationTest):
         # start to look at logs a bit before "now"
         # https://pagure.io/freeipa/issue/8432
         since = time.strftime(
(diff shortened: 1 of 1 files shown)


fcami (approved):
LGTM

fcami:
ACK for when CI passes. I suppose you want to backport this to ipa-4-9 at least.

flo-renaud:
master:

* 5cd3944bb2d3614b9c1268830acfca8ac645ae6a ipatests: use whole date when calling journalctl --since

PR #5599, opened 2021-03-02
Enhance error message when adding non-posix group with a GID

antoniotorresm:
Enhance error message when adding non-posix group
with a GID so the user knows that a GID should not
be passed when adding a group with the --nonposix option.

Fixes: https://pagure.io/freeipa/issue/8155

### DIFF
--- a/ipaserver/plugins/group.py
+++ b/ipaserver/plugins/group.py
@@ -387,6 +387,15 @@ class group_add(LDAPCreate):
                 entry_attrs['gidnumber'] = baseldap.DNA_MAGIC
         return dn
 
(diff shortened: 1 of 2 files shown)


rcritten:
Looks good, needs a test. Adding to the xmlrpc tests is probably the most straightforward IMHO.

rcritten:
LGTM. What, if any, backports are needed?

abbra:
master:

* c0216fb23531d8ad9d4298cb6a51e2baacba16d9 Enhance error message when adding non-posix group with a GID
* 5984f14426d23552eace4795d659fd3bd43b5e64 ipatests: add test for group creation with GID and nonposix option

PR #5604, opened 2021-03-04
prci: run "gating" and "temp_commit" tests on Fedora 33

wladich:


wladich:
Waiting for @netoarmando to update Fedora-33 images.

PR #5557, opened 2021-02-15
ipatests: Test if ACME renews the issued cert with cerbot

mrizwan93:
This test is to check if ACME certificate renews upon
reaching expiry

related: https://pagure.io/freeipa/issue/4751
Signed-off-by: Mohammad Rizwan <myusuf@redhat.com>

PR #5605, opened 2021-03-04
Add ACME profile on replica install in non-ACME environment

rcritten:
Always import the default certificate profiles
    
The default certificate profiles were not imported for
in a promoted client installation.
    
This could lead to missing profiles, as was seen in the ACME
case, while upgrading from older versions.
    
The sequence was:
    
1. Install a non-ACME IPA server
2. Install an ipa client
3. On the client promote to a server with a build that supports
   ACME.
   
The resulting installation is missing the acmeIPAServerCert
profile because it is provided in INCLUDED_PROFILES which isn't
loaded during promotions.
    
There is already handling to handle issues if the profile already
exists so it is safe to call this on every installation.
    
https://pagure.io/freeipa/issue/8738

### DIFF
--- a/ipaserver/plugins/dogtag.py
+++ b/ipaserver/plugins/dogtag.py
@@ -1100,9 +1100,17 @@ def parse_updateCRL_xml(doc):
 #-------------------------------------------------------------------------------
 
 from ipalib import Registry, errors, SkipPluginModule
-if api.isdone('finalize') and api.env.ra_plugin != 'dogtag':
+
+# We only load the dogtag RA plugin if it is necessary to do so.
+# This is legacy code from when multiple RA backends were supported.
+#
+# If the plugins are loaded by the server then load the RA backend.
+#
+if api.isdone("finalize") and not (
+    api.env.ra_plugin == 'dogtag' or api.env.context == 'installer'
+):
     # In this case, abort loading this plugin module...
-    raise SkipPluginModule(reason='dogtag not selected as RA plugin')
+    raise SkipPluginModule(reason='Not loading dogtag RA plugin')
 import os
 from ipaserver.plugins import rabase
 from ipalib.constants import TYPE_ERROR
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -477,12 +477,17 @@ class CAInstance(DogtagInstance):
                         self.step("exposing CA instance on LDAP",
                                   self.__expose_ca_in_ldap)
 
-                    self.step("migrating certificate profiles to LDAP",
-                              migrate_profiles_to_ldap)
                     self.step("importing IPA certificate profiles",
                               import_included_profiles)
+                    self.step("migrating certificate profiles to LDAP",
+                              migrate_profiles_to_ldap)
                     self.step("adding default CA ACL", ensure_default_caacl)
                     self.step("adding 'ipa' CA entry", ensure_ipa_authority_entry)
+                else:
+                    # Re-import profiles in the promote case to pick up any
+                    # that will only be triggered by an upgrade.
+                    self.step("importing IPA certificate profiles",
+                              import_included_profiles)
 
                 self.step("configuring certmonger renewal for lightweight CAs",
                           self.add_lightweight_ca_tracking_requests)


mrizwan93 (commented):
```
2021-03-05 20:15:16,699    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.IPAOpenSSHTransport] RUN ['certbot', '--server', 'https://ipa-ca.ipa.test/acme/directory', 'certonly', '--domain', 'client0.ipa.test', '--standalone']
2021-03-05 20:15:16,699    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] RUN ['certbot', '--server', 'https://ipa-ca.ipa.test/acme/directory', 'certonly', '--domain', 'client0.ipa.test', '--standalone']
2021-03-05 20:15:16,699    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] Saving debug log to /var/log/letsencrypt/letsencrypt.log
2021-03-05 20:15:16,699    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] Plugins selected: Authenticator standalone, Installer None
2021-03-05 20:15:16,699    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] Requesting a certificate for client0.ipa.test
2021-03-05 20:15:16,699    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] Performing the following challenges:
2021-03-05 20:15:16,700    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] http-01 challenge for client0.ipa.test
2021-03-05 20:15:16,700    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] Waiting for verification...
2021-03-05 20:15:16,700    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] Cleaning up challenges
2021-03-05 20:15:16,700    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] An unexpected error occurred:
2021-03-05 20:15:16,700    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] acme.errors.ClientError: <Response [500]>
2021-03-05 20:15:16,700    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] Please see the logfiles in /var/log/letsencrypt for more details.
2021-03-05 20:15:16,700    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd85] Exit code: 1
2021-03-05 20:15:16,700    DEBUG  ipa: ERROR: stderr: Saving debug log to /var/log/letsencrypt/letsencrypt.log
2021-03-05 20:15:16,700    DEBUG  Plugins selected: Authenticator standalone, Installer None
2021-03-05 20:15:16,700    DEBUG  Performing the following challenges:
2021-03-05 20:15:16,701    DEBUG  http-01 challenge for client0.ipa.test
2021-03-05 20:15:16,701    DEBUG  Waiting for verification...
2021-03-05 20:15:16,701    DEBUG  Cleaning up challenges
2021-03-05 20:15:16,701    DEBUG  An unexpected error occurred:
2021-03-05 20:15:16,701    DEBUG  acme.errors.ClientError: <Response [500]>
2021-03-05 20:15:16,701    DEBUG  Please see the logfiles in /var/log/letsencrypt for more details.
2021-03-05 20:15:16,701    DEBUG  
```

https://github.com/freeipa/freeipa/pull/5562 and https://github.com/freeipa/freeipa/pull/5557 is also hit by the same issue .
It is using `pki-acme-10.10.3-3.fc32.noarch`. Fix for this available with `pki-acme-10.10.4`. 

See https://bugzilla.redhat.com/show_bug.cgi?id=1919282

Might have to use pki copr repo.

rcritten:
pki 10.10.5-1.fc32 is available in updates. Seems like an image problem. I'll try a temp commit to update the packages to see if that helps.

rcritten:
Thanks, rebased.

rcritten:
All tests were green but forcing a re-run to exercise the replica setup change once again.

Removing the WIP flag.

flo-renaud:
Hi @rcritten 
I tried the following scenario:
- install IPA server 4.8.10 on server1 with embedded CA
- install IPA client 4.8.10 on server2
- update server2 with the patch
- `ipa-replica-install --setup-ca` on server2
- `ipa certprofile-find` doesn't show the acme profile
- `ipa-acme-manage enable` on server2: still no acme profile.
- `ipa-server-upgrade` on server2/server1: the acme profile is now visible.

I would expect the profile to be visible right after the ipa-replica-install command. Or did I get it wrong?

rcritten:
Upgrade should not be necessary. Installing the replica should be sufficient. The profile should be available on both servers post-install. I'll take another look.

rcritten:
So yeah, I rebased the important patch out of the mix before pushing apparently. I've re-created it (2-liner) and am re-testing now.

rcritten:
This is what I was afraid of twiddling with ra_plugin. It succeeds locally, will investigate.

rcritten:
I discovered a couple more issues I think only one of which I can resolve with code.

In a mixed environment of ACME incapable and ACME capable the ipa-ca DNS A record must point only to the ACME servers. This only needs to be done if ipa-acme-manage enable is run.

ACME also requires that the Apache cert have ipa-ca as a SAN. I need to investigate why this wasn't done in my test bed.

rcritten:
A special case was added in f7c45641fe for issuing the HTTP service principal if the ipa-ca SAN is requested. This is not present in older installs so the newer server falls back to not requesting one. It will need to be added in manually by the admin.

rcritten:
I'm a bit baffled by the failure in temp_commit_3 (test_integration/test_acme.py) because the changes I made occur *after* this. This error means that the ldap principal for the new replica is unknown to the existing server.

RuntimeError: Certificate issuance failed (CA_REJECTED: Server at https://master.ipa.test/ipa/json denied our request, giving up: 2100 (Insufficient access: Principal 'ldap/replica0.ipa.test@IPA.TEST' is not permitted to use CA 'ipa' with profile 'caIPAserviceCert' for certificate issuance.).)

rcritten:
Rebased

rcritten:
I took a completely different approach and added an upgrade file for the ACME profile. This will add it during the final phase of the replica install.

rcritten:
Obviously I'll squash this all down before merging. It was an attempt to keep the temp commit working, which appears to have failed :-(

rcritten:
Re-running failed test. Adding a profile shouldn't cause a failure like this.

rcritten:
test_integration/test_caless.py::TestServerInstall

rcritten:
test_integration/test_caless.py::TestServerInstall::test_valid_certs has failed twice now and I have no idea why. Adding a new profile shouldn't affect caless at all. Re-running again.

flo-renaud:
Hi @rcritten 
I tried the same scenario as above:
- install IPA server 4.8.10 on server1 with embedded CA
- install IPA client 4.8.10 on server2
- update server2 with the patch
- ipa-replica-install --setup-ca on server2
- ipa certprofile-find doesn't show the acme profile
- ipa-acme-manage enable on server2: still no acme profile.
- ipa-server-upgrade on server2/server1: the acme profile is now visible.

The code importing the cert profiles in the installer (`import_included_profiles`) is called only `if not promote`, meaning that it doesn't get executed in my scenario (install a replica with a clone CA).

rcritten:
Setting WIP flag while I experiement with additional ways to load the dogtag RA plugin in more cases.

rcritten:
Dropping WIP since the new method seems to be working.

flo-renaud:
Hi @rcritten 
I am still seeing the same behavior:
- install old server without acme support
- install replica with the patch
- certprofile doesn't show the new profile
- run ipa-server-upgrade on the replica: the new profile is now visible on server and replica

rcritten:
No wonder it didn't work: I hadn't changed the patch set. I suspect what I did was test new code, got it working, did a rebase to drop the old non-working change and dropped the working one instead.

flo-renaud:
Hi @rcritten 
thanks for the update. I tested the following scenario:
- [x] install 4.8.10 master (fedora33), install replica with patch, the certprofile is installed on both nodes and `ipa-acme-manage enable` works, a client is able to get a cert.
- [ ] install 4.8.6 master (fedora 32), install replica with patch, the certprofile is installed on both nodes but `ipa-acme-manage enable` fails on the replica: `invalid 'certificate': Does not have a 'ipa-ca.domain.com' SAN`
- [ ] install 4.8.6 master, upgrade with patch, the upgrade fails in migrate_profiles_to_ldap (the instruction api.Backend.ra_certprofile.override_port = 8443 fails with AttributeError(key)). Since the context is 'updates', the module  ipaserver/plugins/dogtag.py doesn't get imported.

rcritten:
@flo-renaud the second test is expected. ACME is **not** going to play very nicely in a mixed environment.

It uses the ipa-ca name for discovering the ACME endpoint but that endpoint is not available on all CA's. A workaround is to remove the non-ACME CA's from that CNAME but, yeah, not awesome. The next issue is that the Apache cert on the ACME-capable server will be lacking the ipa-ca SAN. It's pretty straightforward to re-issue this but the error from certbot is not great.

Doing these things I was able to get an ACME client to get a certificate.

In the meantime I'll work on the upgrade issue.

rcritten:
I addressed the issue, and potential future ones, by adding back in the ra_plugin test along with the new context test.

ipaserver/plugins/dogtag.py
```diff
@@ -1100,7 +1100,15 @@ def parse_updateCRL_xml(doc):
...
-if api.isdone('finalize') and api.env.ra_plugin != 'dogtag':
+if api.isdone("finalize") and not (
+    api.env.ra_plugin == "dogtag"
+    or api.env.context
+    in (
+        "installer",
+        "server",
+        "lite",
+    )
+):
```
flo-renaud, line 1111: I am puzzled by this change and fail to understand why it is needed. When the context is installer, server or lite the plugin is always loaded, even if ra_plugin is None. In which situation is it used? Is it because during promotion the ra_plugin is initially None (because we start from a client-only) ?
I need to confess that all the variables in the installer vs the api are confusing for me: ra_only, clone, promote / ra_plugin, enable_ra.

rcritten: The bulk of the dogtag module is not loaded unless certain conditions are met, making the RA interface unavailable. The RA interface is needed to load the ACME profile into the CA.

The root of the problem is that during promotion ra_plugin is None. Trying to set it early caused a whole other series of hard to debug problems so I went with context. I initially tried using _only_ context but that broke upgrades. Given there could be other unknowns in the current patch I try the least-possible change. I should limit the context to only installer since I am no longer dropping the ra_plugin test.

ra_plugin and enable_ra date back to when there was a lot more abstraction around the CA since we supported more than just dogtag. There were originally two ra_plugins, dogtag and selfsign. So except for a CA-less install, any IPA server can communicate with the CA but only some have the CA installed (so they are all RA's but only some are CA's). enable_ra probably made more sense in the past with the abstraction layer.

ra_only is completely confusing and seems to really mean "setup-ca was passed in." 

IIRC at one point a fair bit of code was duplicated in order to have two separate paths for initial installation and replica install. These were merged and then DL1 (promoting a client to server) was added which added additional complexity.

rcritten:
I can reproduce the temp_commit failure locally finally so maybe I'll have a chance to figure out what's going on.

rcritten:
Got it. No profiles were being created on the initial server so the replica install which needs caIPAserviceCert failed. The profiles are used in the caacl processing so no profile, no access.

flo-renaud:
Hi @rcritten 
tested manually various cases, all of them work like a charm:
- [x] install 4.8.10 master (fedora33), install replica with patch, the certprofile is installed on both nodes. If I tweak the ipa-ca DNS records, `ipa-acme-manage enable` works, a client is able to get a cert.
- [x]  install 4.8.10 master, upgrade with patch. The certprofile is installed, `ipa-acme-manage status` succeeds, a client is able to get a cert.
- [x] install 4.8.6 master (fedora 32), install replica with patch, the certprofile is installed on both nodes. `ipa-server-upgrade` on the replica resubmits the http cert with the additional SAN. If I tweak the DNS records for ipa-ca, `ipa-acme-manage enable` succeeds on the replica.
- [x] install 4.8.6 master, upgrade with patch, `ipa-server-upgrade` resubmits the http cert with the additional SAN and `ipa-acme-manage enable` succeeds.
- [x] The fresh install is already tested in PRCI and tests were green, also good.

You can remove the temp commit and I will ACK when PRCI completes.

rcritten:
Dropped temp commit and rebased.

flo-renaud:
Now that tests are green without the temp commit, ACK.

flo-renaud:
master:

* b01547da79c189bfb5b4b32e54b2ef2cb733741b Load dogtag RA plugin in installers so profiles can be loaded

PR #5610, opened 2021-03-05
ipatests: test_installation: add install test scenarios

miskopo:
test_hostname_parameter: Test for issue 2692 ipa-server-install ignores --hostname:
check whether hostname provided in `--hostname` parameter is being taken into account and set as new hostname without prompting for it again
    
test_ad_subpackage_dependency: Test for issue 4011 ipa-server-install crashes when AD subpackage is not installed:
test if ipa-server installation succeeds without `freeipa-ipa-server-trust-ad` installed
    
test_backup_of_cs_cfg_should_be_created: Test for issue 4166 Backup CS.cfg before modifying it:
test if ipa-server installer backs up CS.cfg before modifying it
    
test_installer_wizard_should_prompt_for_DNS: Test for issue 2575 [RFE] Installer wizard should prompt for DNS:
test if installer is asking for DNS setup details if not provided as parameter
    
Related: https://pagure.io/freeipa/issue/2692
Related: https://pagure.io/freeipa/issue/4011
Related: https://pagure.io/freeipa/issue/4166
Related: https://pagure.io/freeipa/issue/2575
    
### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -58,6 +58,17 @@ def server_install_setup(func):
     return wrapped
 
 
+@pytest.fixture
+def server_cleanup(request):
+    """
+    Fixture to uninstall ipa server before and after the test
+    """
+    host = request.cls.master
+    tasks.uninstall_master(host)
+    yield
+    tasks.uninstall_master(host)
+
+
 class InstallTestBase1(IntegrationTest):
 
     num_replicas = 3
@@ -662,7 +673,6 @@ def get_pki_tomcatd_pid(host):
             break
     return(pid)
 
-
 def get_ipa_services_pids(host):
     ipa_services_name = [
         "krb5kdc", "kadmin", "named", "httpd", "ipa-custodia",
@@ -1119,6 +1129,113 @@ class TestInstallMaster(IntegrationTest):
                 expected_stdout=f'href="https://{self.master.hostname}/'
             )
 
+    def test_hostname_parameter(self, server_cleanup):
+        """
+        Test that --hostname parameter is respected in interactive mode.
+
+        https://pagure.io/freeipa/issue/2692
+        """
+        original_hostname = self.master.hostname
+        new_hostname = 'new.' + original_hostname
+        # New hostname is added into /etc/hosts as the installer
+        # is looking for it for resolution. Without it, an installation
+        # fails with `Unable to resolve host name, check /etc/hosts or DNS
+        # name resolution`.
+        hosts = self.master.get_file_contents(paths.HOSTS, encoding='utf-8')
+        new_hosts = hosts.replace(original_hostname, new_hostname)
+        self.master.put_file_contents(paths.HOSTS, new_hosts)
+        try:
+            cmd = ['ipa-server-install', '--hostname', new_hostname]
+            with self.master.spawn_expect(cmd) as e:
+                e.expect_exact('Do you want to configure integrated '
+                               'DNS (BIND)? [no]: ')
+                e.sendline('no')
+                e.expect_exact('Please confirm the domain name [{}]: '.format(
+                    original_hostname  # DN is computed from new hostname
+                ))
+                e.sendline(self.master.domain.name)
+                e.expect_exact('Please provide a realm name [{}]: '.format(
+                    self.master.domain.realm
+                ))
+                e.sendline(self.master.domain.realm)
+                e.expect_exact('Directory Manager password: ')
+                e.sendline(self.master.config.dirman_password)
+                e.expect_exact('Password (confirm): ')
+                e.sendline(self.master.config.dirman_password)
+                e.expect_exact('IPA admin password: ')
+                e.sendline(self.master.config.admin_password)
+                e.expect_exact('Password (confirm): ')
+                e.sendline(self.master.config.admin_password)
+                e.expect_exact('Do you want to configure chrony with '
+                               'NTP server or pool address? [no]: ')
+                e.sendline('no')
+                e.expect_exact('Continue to configure the system '
+                               'with these values? [no]: ')
+                e.sendline('yes')
+                e.expect_exit(ignore_remaining_output=True, timeout=720)
+            hostname = self.master.run_command([
+                'hostname', '-f']).stdout_text.strip()
+            assert hostname == new_hostname
+        finally:
+            # no need to restore the hostname as the installer
+            # does it during uninstallation
+            self.master.put_file_contents(paths.HOSTS, hosts)
+
+    def test_ad_subpackage_dependency(self, server_cleanup):
+        """
+        Test if the installer is not dependant on trust-ad package and
+        succeeds even when AD subpackage is not installed
+
+        https://pagure.io/freeipa/issue/4011
+        """
+        if osinfo.id == 'fedora':
+            package_name = 'freeipa-server-trust-ad'
+        else:
+            package_name = 'ipa-server-trust-ad'
+        reinstall = False
+        if tasks.is_package_installed(self.master, package_name):
+            tasks.uninstall_packages(self.master, [package_name])
+            reinstall = True
+        try:
+            tasks.install_master(self.master)
+        finally:
+            if reinstall:
+                tasks.install_packages(self.master, [package_name])
+
+    def test_backup_of_cs_cfg_is_created(self, server_cleanup):
+        """
+        Test that the installer backs up CS.cfg configuration before it's
+        being modified.
+
+        https://pagure.io/freeipa/issue/4166
+        """
+        bcp_location = paths.CA_CS_CFG_PATH + '.ipabkp'
+        original_cfg_content = None
+        if self.master.transport.file_exists(paths.CA_CS_CFG_PATH):
+            original_cfg_content = self.master.get_file_contents(
+                paths.CA_CS_CFG_PATH, encoding='utf-8')
+
+        time_before_install = int(self.master.run_command(
+            ['date', '+%s']).stdout_text.strip())
+        tasks.install_master(self.master)
+        ipaserver_install_log = self.master.get_file_contents(
+            paths.IPASERVER_INSTALL_LOG, encoding='utf-8')
+        assert 'backing up CS.cfg' in ipaserver_install_log
+        assert self.master.transport.file_exists(bcp_location)
+
+        cfg_mod_time = int(self.master.run_command(
+            ['stat', '-c', '%Y', paths.CA_CS_CFG_PATH]).stdout_text.strip())
+        bcp_mod_time = int(self.master.run_command(
+            ['stat', '-c', '%Y', bcp_location]).stdout_text.strip())
+
+        # check if the backup file was modified/created during the installation
+        # and before the original file was modified
+        assert time_before_install <= bcp_mod_time <= cfg_mod_time
+        bcp_cfg_content = self.master.get_file_contents(
+            paths.CA_CS_CFG_PATH + '.ipabkp', encoding='utf-8')
+        if original_cfg_content is not None:
+            assert original_cfg_content == bcp_cfg_content
+
 
 class TestInstallMasterKRA(IntegrationTest):
 
@@ -1215,6 +1332,60 @@ class TestInstallMasterDNS(IntegrationTest):
 
     def test_install_kra(self):
         tasks.install_kra(self.master, first_instance=True)
+
+    def test_installer_wizard_prompts_for_DNS(self, server_cleanup):
+        """
+        Installer wizard should prompt for DNS even if --setup-dns is not
+        provided as an argument.
+
+        https://pagure.io/freeipa/issue/2575
+        """
+        cmd = ['ipa-server-install']
+        with self.master.spawn_expect(cmd) as e:
+            e.expect_exact('Do you want to configure integrated '
+                           'DNS (BIND)? [no]: ')
+            e.sendline('yes')
+            e.expect_exact('Server host name [{}]: '.format(
+                self.master.hostname))
+            e.sendline(self.master.hostname)
+            e.expect_exact('Please confirm the domain name [{}]: '.format(
+                self.master.domain.name))
+            e.sendline(self.master.domain.name)
+            e.expect_exact('Please provide a realm name [{}]: '.format(
+                self.master.domain.realm
+            ))
+            e.sendline(self.master.domain.realm)
+
+            e.expect_exact('Directory Manager password: ')
+            e.sendline(self.master.config.dirman_password)
+            e.expect_exact('Password (confirm): ')
+            e.sendline(self.master.config.dirman_password)
+
+            e.expect_exact('IPA admin password: ')
+            e.sendline(self.master.config.admin_password)
+            e.expect_exact('Password (confirm): ')
+            e.sendline(self.master.config.admin_password)
+
+            e.expect_exact(
+                'Do you want to configure DNS forwarders? [yes]: ')
+            e.sendline('no')  # irrelevant for this test
+            e.expect_exact('Do you want to search for missing reverse '
+                           'zones? [yes]: ')
+            e.sendline('no')  # irrelevant for this test
+            e.expect_exact('Do you want to configure chrony with NTP '
+                           'server or pool address? [no]: ')
+            e.sendline('no')  # irrelevant for this test
+            e.expect_exact('Continue to configure the system with these '
+                           'values? [no]: ')
+            e.sendline('yes')
+            e.expect_exit(ignore_remaining_output=True, timeout=720)
+
+        tasks.kinit_admin(self.master)
+        result = self.master.run_command(
+            ['ipa', 'dnszone-show', self.master.domain.name]
+        ).stdout_text
+
+        assert 'Active zone: TRUE' in result
 
 
 class TestInstallMasterDNSRepeatedly(IntegrationTest):


ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1108,74 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
             )
 
+    def test_hostname_parameter(self):
+        """
+        Test for #2692 ipa-server-install ignores --hostname
+
+        https://pagure.io/freeipa/issue/2692
+        """
+        original_hostname = self.master.hostname
+        new_hostname = "ipa2692.test"
```
rcritten, line 1118: Single label domain. It is probably better to use the existing domain just a new hostname in it.

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1108,74 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        result = tasks.install_master(self.master)
+        assert result.returncode == 0
+
+    def test_backup_of_cs_cfg_should_be_created(self):
+        """
+        Test for #4166 Backup CS.cfg before modifying it
+
+        https://pagure.io/freeipa/issue/4166
+        """
+        log_location = "/var/lib/pki/pki-tomcat/ca/conf/CS.cfg"
```
rcritten, line 1153: paths.CA_CS_CFG_PATH

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1108,74 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        bcp_location = log_location + ".ipabkp"
+        ipainstall_log_location = "/var/log/ipaserver-install.log"
```
rcritten, line 1155: paths.IPASERVER_INSTALL_LOG

miskopo:
I am rather puzzled by presence of error in `test_install_kra`, saying that KRA is already installed, when no prior KRA installation is being executed. The check follows:
```
2021-03-07T16:48:28Z DEBUG args=['pki-server', 'subsystem-show', 'kra']
2021-03-07T16:48:29Z DEBUG Process finished, return code=0
2021-03-07T16:48:29Z DEBUG stdout=  Subsystem ID: kra
  Instance ID: pki-tomcat
  Enabled: True
```

ipatests/test_integration/test_installation.py
```diff
@@ -670,6 +670,13 @@ def get_ipa_services_pids(host):
 ##
 # Rest of master installation tests
 ##
+def master_unistalled(function):
```
wladich, line 673: As this decorator is actually a fixture, what do you think about defining it as a pytest.fixture?
You also should wrap `function()` in try-finally block (not needed if using pytest.fixture)

ipatests/test_integration/test_installation.py
```diff
@@ -1178,6 +1258,23 @@ def test_install_master(self):
...
         assert "Administrator e-mail address: me.example.org" in result
 
+    @master_unistalled
+    def test_installer_wizard_should_prompt_for_DNS(self):
+        """
+        Test for #2575 [RFE] Installer wizard should prompt for DNS
+
+        https://pagure.io/freeipa/issue/2575
+        """
+        tasks.install_master(self.master,
```
wladich, line 1268: It is not clear if this command really checks anything about installer questions. Also this step can brake and it will be difficult to find out why when the installation scenario (i.e. wizard questions) changes.
I suggest to rewrite this test similar to https://github.com/freeipa/freeipa/blob/886c0dcb07e84e2d92545d84bed4bc55c5c520f7/ipatests/test_integration/test_adtrust_install.py#L108

ipatests/test_integration/test_installation.py
miskopo: Good point, let me rewrite it as a fixture

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1115,79 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
                 expected_stdout=f'href="https://{self.master.hostname}/'
             )
 
+    @master_unistalled
+    def test_hostname_parameter(self):
+        """
+        Test for #2692 ipa-server-install ignores --hostname
```
wladich, line 1121: 1. issue number is duplicated in the link below, so it is not needed here.
2. Please try to follow [PEP257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings), specifically
> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".
3. The docstring should provide enough information about the purpose of the test. Now I need to follow the link to get what the test is trying to check.

What I mean is that docstring should say what test does, not what the bug was. Something like "Check ipa-server-install respects --hostname option in interactive mode."

Same applies to following tests

miskopo:
I've noticed an error in `test_installer_wizard_should_prompt_for_DNS`, I'll resolve with newest update

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1115,79 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        hosts = self.master.get_file_contents('/etc/hosts')
+        new_hosts = str(hosts) + "\n{}        {}".format(
+            ip_address, new_hostname)
+        self.master.put_file_contents('/etc/hosts', new_hosts)
+        result = tasks.install_master(self.master,
+                                      setup_dns=False,
+                                      extra_args=['--hostname',
+                                                  new_hostname])
+        try:
+            assert result.returncode == 0
```
wladich, line 1137: assert is redundant  since `raiseonerr` parameter of `tasks.install_master` is True by default

ipatests/test_integration/test_installation.py
```diff
@@ -1178,6 +1258,23 @@ def test_install_master(self):
...
 
+    @master_unistalled
+    def test_installer_wizard_should_prompt_for_DNS(self):
+        """
+        Test for #2575 [RFE] Installer wizard should prompt for DNS
+
+        https://pagure.io/freeipa/issue/2575
+        """
+        tasks.install_master(self.master,
+                             setup_dns=False,
```
miskopo, line 1269: `unattended` must be set to `False`

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1115,79 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        https://pagure.io/freeipa/issue/2692
+        """
+        original_hostname = self.master.hostname
+        new_hostname = "new." + original_hostname
+        ip_address = self.master.ip
+        hosts = self.master.get_file_contents('/etc/hosts')
+        new_hosts = str(hosts) + "\n{}        {}".format(
+            ip_address, new_hostname)
+        self.master.put_file_contents('/etc/hosts', new_hosts)
+        result = tasks.install_master(self.master,
```
wladich, line 1132: The original [issue](https://pagure.io/freeipa/issue/2692) is describing bug during interactive installation but the test is installing ipa server in unattended mode.

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1115,79 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        """
+        Test for #4011 ipa-server-install crashes
+        when AD subpackage is not installed
+
+        https://pagure.io/freeipa/issue/4011
+        """
+        package_name = "*ipa-server-trust-ad"
+        tasks.uninstall_packages(self.master, [package_name])
+        result = tasks.install_master(self.master)
+        assert result.returncode == 0
```
wladich, line 1156: redundant assertion

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1115,79 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+    @master_unistalled
+    def test_backup_of_cs_cfg_should_be_created(self):
+        """
+        Test for #4166 Backup CS.cfg before modifying it
+
+        https://pagure.io/freeipa/issue/4166
+        """
+        bcp_location = paths.CA_CS_CFG_PATH + ".ipabkp"
+        tasks.install_master(self.master)
+        ipaserver_install_log = str(self.master.get_file_contents(
```
wladich, line 1167: There is another way to get file contents as a string: 
```
self.master.get_file_contents(path, encoding='utf-8')
```

What the current implementation does is calling `__repr__`:
```
str(b'abc')
Out[2]: "b'abc'"
```

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1115,79 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+            paths.IPASERVER_INSTALL_LOG))
+        assert "backing up CS.cfg" in ipaserver_install_log
+        try:
+            self.master.get_file_contents(bcp_location)
+        except IOError:
+            raise AssertionError("Backup of CS.cfg not created")
```
wladich, line 1173: pytest idiomatic way for this is `pytest.fail(reason)`

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1115,79 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+
+        https://pagure.io/freeipa/issue/4166
+        """
+        bcp_location = paths.CA_CS_CFG_PATH + ".ipabkp"
+        tasks.install_master(self.master)
+        ipaserver_install_log = str(self.master.get_file_contents(
+            paths.IPASERVER_INSTALL_LOG))
+        assert "backing up CS.cfg" in ipaserver_install_log
+        try:
+            self.master.get_file_contents(bcp_location)
```
wladich, line 1171: It will be nice to also check that contents of the backup are really equal to what was in the file before the installation.

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1115,79 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+
+    @master_unistalled
+    def test_update_krb5_conf_template_with_keyring_default_ccache_name(self):
+        """
+        Test for #4013 Update krb5.conf.template to support
+        KEYRING default_ccache_name
+
+        https://pagure.io/freeipa/issue/4013
+        """
+        tasks.uninstall_master(self.master)
```
wladich, line 1183: master is already uninstalled by fixture.

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1115,79 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        tasks.install_master(self.master)
+        tasks.kinit_admin(self.master)
+        klist = self.master.run_command(['klist']).stdout_text
+        assert '/tmp/' not in klist
+        krb5conf = str(self.master.get_file_contents('/etc/krb5.conf'))
```
wladich, line 1188: See above about getting file contents as a string

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1115,79 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        assert "default_ccache_name = KEYRING:persistent:%{uid}" in krb5conf
```
wladich, line 1189: This line can be present in file but commented out. We should check if it is really in effect

wladich (commented):
Two new test cases do not deal with the installation process but rather check the rusult: `test_update_krb5_conf_template_with_keyring_default_ccache_name` and 
`test_backup_of_cs_cfg_should_be_created`. Do you think those could be integrated to some other tests to save two ipa-server installations? This could save about 20 for of test execution.

ipatests/test_integration/test_installation.py
wladich: 1. Fixtures do have "scope" parameter
2. Fixtures are applied to each test case individually unless the `autouese` parameter is `True` (False by default)
3. You do need to uninstall ipa server in every test case you have added

ipatests/test_integration/test_installation.py
```diff
@@ -1108,6 +1115,79 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+    def test_hostname_parameter(self):
+        """
+        Test for #2692 ipa-server-install ignores --hostname
+
+        https://pagure.io/freeipa/issue/2692
+        """
+        original_hostname = self.master.hostname
+        new_hostname = "new." + original_hostname
+        ip_address = self.master.ip
+        hosts = self.master.get_file_contents('/etc/hosts')
```
wladich, line 1128: Can you please comment why do we need to edit `/etc/hosts` file?

Anyway you should backup it and restore after the test.

wladich: Sorry, did not notice it is restored in the end.

miskopo: I need to add new hostname to `/etc/hosts`, because it's being looked for there by the installer. If it's not found, installation fails with
```
Unable to resolve host name, check /etc/hosts or DNS name resolution

```


`/etc/hosts/` is being backed-up and restored in `finally` branch.

wladich: Thank you for explanation.  
If you will be adding more changes, please also add this to comments in the code as this preparation step is not obvious.

miskopo:
> Two new test cases do not deal with the installation process but rather check the rusult: `test_update_krb5_conf_template_with_keyring_default_ccache_name` and
> `test_backup_of_cs_cfg_should_be_created`. Do you think those could be integrated to some other tests to save two ipa-server installations? This could save about 20 for of test execution.

That's true, however, each test should test one thing and not be dependent on other test - if I merged these two test case into, e.g. `test_install_master`, I could obfuscate failures in one of the test cases, depending on which is going to be first.

ipatests/test_integration/test_installation.py
miskopo: The file is not available before the installation, so unless I install twice, I can't get the "original" file.

miskopo:
relevant test - `TestInstallMasterDNS::test_installer_wizard_should_prompt_for_DNS` in `temp_commit_DNS ` is passing. The others are known failures.

miskopo:
Except for some unrelated failures in `temp_commit_DNS `, all checks passed. Marking as ready to review.

Temp commit runs:
- [DNS only](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/dc18c89a-9df4-11eb-adfd-52540098fe37/)
- [Install only](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/d48ae716-9df4-11eb-8a52-525400698708/)

ipatests/test_integration/test_installation.py
```diff
@@ -1119,6 +1130,115 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        ipaserver_install_log = str(self.master.get_file_contents(
+            paths.IPASERVER_INSTALL_LOG, encoding='utf-8'))
+        assert "backing up CS.cfg" in ipaserver_install_log
+        try:
+            self.master.get_file_contents(bcp_location, encoding='utf-8')
+        except IOError:
+            pytest.fail("Backup of CS.cfg not created")
+            raise
+
+    @pytest.mark.skip_if_container(
```
wladich, line 1218: It seems to me that this test case can be moved to `TestInstallMaster` class which installs ipa server in the `install()` method. 
By removing `server_cleanup` and `tasks.install_master(self.master)` we can save about 10 minutes without degrading the test.

wladich (changes requested):
The failures in temp_commit_DNS ARE related to the changes.
Existing tests in TestInstallMasterDNS rely on ipa server being installed during `install()` but the new test test_installer_wizard_should_prompt_for_DNS uninstalls it.

ipatests/test_integration/test_installation.py
```diff
@@ -678,10 +693,6 @@ def get_ipa_services_pids(host):
     return pids_of_ipa_services
 
 
-##
```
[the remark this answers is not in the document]
miskopo: Frankly, I am not aware of this change. Reverted

ipatests/test_integration/test_installation.py
```diff
@@ -1119,6 +1130,115 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        Installer should update krb5.conf.template to support
+        KEYRING default_ccache_name.
+
+        https://pagure.io/freeipa/issue/4013
+        """
+        server_cleanup(self.master)
+        tasks.install_master(self.master)
+        tasks.kinit_admin(self.master)
+        klist = self.master.run_command(['klist']).stdout_text
+        assert '/tmp/' not in klist
```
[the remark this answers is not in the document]
miskopo: There is positive check, lines 1240-1245

ipatests/test_integration/test_installation.py
```diff
@@ -1119,6 +1130,115 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        krb5conf = str(self.master.get_file_contents('/etc/krb5.conf'),
+                       encoding='utf-8')
+        for line in krb5conf.split('\n'):
```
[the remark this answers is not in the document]
miskopo: no, because we want to make sure the line isn't commented out, as you suggested previously.

ipatests/test_integration/test_installation.py
```diff
@@ -1119,6 +1130,115 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        hosts = self.master.get_file_contents('/etc/hosts', encoding='utf-8')
+        new_hosts = str(hosts).replace(original_hostname, new_hostname)
+        self.master.put_file_contents('/etc/hosts', new_hosts)
+        try:
+            cmd = ['ipa-server-install', '--hostname', new_hostname]
+            with self.master.spawn_expect(cmd) as e:
+                e.expect_exact('Do you want to configure integrated '
+                               'DNS (BIND)? [no]: ')
+                e.sendline('no')
+                e.expect_exact('Please confirm the domain name [{}]: '.format(
```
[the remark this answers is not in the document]
miskopo: I'll investigate

ipatests/test_integration/test_installation.py
miskopo: I am sorry, but I have to disagree. The tests must not be tight-coupled, even if it would mean extra machine time. Also, I am not sure what you meant, because `TestInstallMaster` has only `pass` in its class install method.

ipatests/test_integration/test_installation.py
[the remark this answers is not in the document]
miskopo: so, if I understand correctly, the installer is correct:
From `ipa-server-install` manpage:
```
--hostname=HOST_NAME
                The fully-qualified DNS name of this server.
```
and 
```
-n DOMAIN_NAME, --domain=DOMAIN_NAME
                 The  primary  DNS  domain  of the IPA deployment, e.g. example.com. This DNS domain should contain the SRV records generated by the IPA server installer. The specified DNS
                 domain must not contain DNS records of any other LDAP or Kerberos based management system (like Active Directory or MIT Kerberos).
   
                 It is strongly recommended to use a lower-cased name of the IPA Kerberos realm name.
   
                 The primary DNS domain name cannot be changed after the installation.
```
which suggests that domain name is computed from hostname's base. Maybe @flo-renaud or @rcritten would be able to confirm or deny this statement.

miskopo:
[`temp commit`](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/c6f2e1a4-a299-11eb-8c41-525400be55b5), [`temp commit DNS`](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/d32e3874-a299-11eb-b06b-52540098fe37)

@wladich, could you please review? Thank you :)

ipatests/test_integration/test_installation.py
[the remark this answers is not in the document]
miskopo: @fcami confirmed this is a correct behaviour.

wladich (commented):
In added code you are using both types of quotes - single and double. According to [PEP 8](https://www.python.org/dev/peps/pep-0008/#string-quotes) it is advisable to stick to one type of quotes. This does not apply do docstrings, they should always use double quotes.

ipatests/test_integration/test_installation.py
```diff
@@ -1119,6 +1129,108 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+                e.sendline('no')
+                e.expect_exact('Continue to configure the system '
+                               'with these values? [no]: ')
+                e.sendline('yes')
+                e.expect_exit(ignore_remaining_output=True, timeout=720)
+            hostname = self.master.run_command([
+                'hostname', '-f']).stdout_text.strip()
+            assert hostname == new_hostname
+        finally:
+            self.master.hostname = original_hostname
```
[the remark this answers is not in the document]
miskopo: In fact, the `hostname` is changed by the installer, however, server is uninstalled at the end and the original `hostname` is restored, so it's really okay to remove.

miskopo:
@wladich I am really sorry, but I am once again asking you for a review. I am, however, eternally grateful for what you do.

ipatests/test_integration/test_installation.py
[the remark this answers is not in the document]
miskopo: Fixed by using `regex`

ipatests/test_integration/test_installation.py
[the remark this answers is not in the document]
miskopo: @wladich I am not quite sure what you mean. What exactly is the difference between `Proper cache is used` and `Improper cache is not used.`, please?

ipatests/test_integration/test_installation.py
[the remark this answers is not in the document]
miskopo: In fact, you are right. It's not being changed in the test configuration, so the restoration is pointless.

ipatests/test_integration/test_installation.py
```diff
@@ -1119,6 +1129,108 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
 
+    def test_hostname_parameter(self, server_cleanup):
+        """
+        Test that --hostname parameter is respected in interactive mode.
+
+        https://pagure.io/freeipa/issue/2692
+        """
+        original_hostname = self.master.hostname
+        new_hostname = "new." + original_hostname
+        """New hostname is added into /etc/hosts as the installer
```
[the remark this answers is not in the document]
miskopo: My bad, sorry. Fixed the code

ipatests/test_integration/test_installation.py
```diff
@@ -1119,6 +1130,115 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+
+        https://pagure.io/freeipa/issue/4166
+        """
+        server_cleanup(self.master)
+        bcp_location = paths.CA_CS_CFG_PATH + ".ipabkp"
+        tasks.install_master(self.master)
+        ipaserver_install_log = str(self.master.get_file_contents(
+            paths.IPASERVER_INSTALL_LOG, encoding='utf-8'))
+        assert "backing up CS.cfg" in ipaserver_install_log
+        try:
```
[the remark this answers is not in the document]
miskopo: Implemented time checks.

miskopo:
as seen in [failed temp commit](http://idm-artifacts.usersys.redhat.com/QEW/identity-management/ipa-tests/1059740/install-server-cli/3445730//runner.log), klist doesn't use KEYRING, but KCM in this particular situation. This, however, has nothing to do with the original ticket.

miskopo:
- [`temp commit for non-DNS tests`](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/afc893bc-ae37-11eb-a201-fa163ed6a638)
- [`temp commit for DNS test`](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/bf99f416-ae37-11eb-9ebd-525400867fd5)

wladich:
@miskopo Please do check if timeouts in prci definitions need to be adjusted.

miskopo:
> @miskopo Please do check if timeouts in prci definitions need to be adjusted.

Both `TestInstallMaster` and `TestInstallMasterDNS` have plenty of extra time, more than 16 minutes in gating and more than 2 hours in nightlies.

ipatests/test_integration/test_installation.py
```diff
@@ -1216,6 +1330,60 @@ def test_server_install_lock_bind_recursion(self):
...
+                           'values? [no]: ')
+            e.sendline('yes')
+            e.expect_exit(ignore_remaining_output=True, timeout=720)
+
+        tasks.kinit_admin(self.master)
+        result = self.master.run_command(
+            ['ipa', 'dnszone-show', self.master.domain.name]
+        ).stdout_text
+
+        assert "Active zone: TRUE" in result
```
wladich, line 1385: Inconsistent usage of double quotes

ipatests/test_integration/test_installation.py
```diff
@@ -1119,6 +1129,110 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+                e.sendline('no')
+                e.expect_exact('Continue to configure the system '
+                               'with these values? [no]: ')
+                e.sendline('yes')
+                e.expect_exit(ignore_remaining_output=True, timeout=720)
+            hostname = self.master.run_command([
+                'hostname', '-f']).stdout_text.strip()
+            assert hostname == new_hostname
+        finally:
+            # no need to restore the hostane as the installer
```
wladich, line 1180: Typo: "hostane" -> "hostname"

ipatests/test_integration/test_installation.py
```diff
@@ -1119,6 +1129,110 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        if tasks.is_package_installed(self.master, package_name):
+            tasks.uninstall_packages(self.master, [package_name])
+            reinstall = True
+        try:
+            tasks.install_master(self.master)
+        finally:
+            if reinstall:
+                tasks.install_packages(self.master, [package_name])
+
+    def test_backup_of_cs_cfg_should_be_created(self, server_cleanup):
```
wladich, line 1202: Same as with `test_installer_wizard_prompts_for_DNS`: proper form of the phrase is "test backup is created" or "test creates backup"

wladich (commented):
Please rewrite commit message to avoid GitHub from interpretting Pagure issues as GH PRs
Please see few minor issues in inline comments. Otherwise LGTM

ipatests/test_integration/test_installation.py
```diff
@@ -1119,6 +1129,110 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+    def test_ad_subpackage_dependency(self, server_cleanup):
+        """
+        Test if the installer is not dependant on trust-ad package and
+        succeeds even when AD subpackage is not installed
+
+        https://pagure.io/freeipa/issue/4011
+        """
+        package_name = '*ipa-server-trust-ad'
+        reinstall = False
+        if tasks.is_package_installed(self.master, package_name):
```
wladich, line 1193: This internaly calls `'rpm' '-q' '*ipa-server-trust-ad'`. Unlike `dnf` the `rpm` does not expand wildcards and this check will allways return False

ipatests/test_integration/test_installation.py
```diff
@@ -1119,6 +1129,110 @@ def run_request(url, expected_stdout=None, expected_stderr=None):
...
+        assert self.master.transport.file_exists(bcp_location)
+
+        cfg_mod_time = self.master.run_command(
+            ['stat', '-c', '%Y', paths.CA_CS_CFG_PATH]).stdout_text.strip()
+        bcp_mod_time = self.master.run_command(
+            ['stat', '-c', '%Y', bcp_location]).stdout_text.strip()
+
+        # check if the backup file was modified/created during the installation
+        # and before the original file was modified
+        assert time_before_install <= bcp_mod_time <= cfg_mod_time
```
wladich, line 1230: You are comparing numbers as strings. This is generally a bad idea, please convert them to numbers. This will  verify that the output of `date` and `stat` is in expected format.

wladich (changes requested):
Removing "LGTM". Looking through the logs I found the incorrect invocation of `rpm`

miskopo:
- [`temp commit TestInstallMaster`](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/6a082de0-ae61-11eb-b8b3-52540010f1df)
- [`temp commit TestInstallMasterDNS`](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/79ff9a76-ae61-11eb-892b-525400aa93bc)

wladich:
LGTM #2
Plese remove the temp commit and I will ACk

miskopo:
master:

* b8ebce7b185f4c934be693a1fbefe5fd2f279e38 ipatests: test_installation: add install test scenarios

PR #5612, opened 2021-03-05
DNSSEC key synchronization issues

flo-renaud:
### dnssec: fix ipa-ods-exporter crash when master key missing

When a master key is missing from the local HSM, ipa-ods-exporter crashes.
This can happen when the DNSSEC master role is moved from one node to
another with the following scenario:
- install server1 with dns + dnssec
- install server2 without dns
- disable dnssec from server1
- install dns + dnssec on server2

With the above scenario, server2 never had the opportunity to get
the master key (this happens only when the replica is already
configured as DNS server and has put its public replica key in LDAP +
the current DNSSEC master wraps its master key with the replica key).

ipa-ods-exporter can only log an error instead of crashing.

Related: https://pagure.io/freeipa/issue/8654

### dnssec: concurrency issue when disabling old replica key

When dnssec role is removed from the local node, the uninstaller
creates a new replica key and marks the older replica keys as disabled
(both in the local HSM and in LDAP).
If ipa-ods-exporter runs in the middle of this operation, the old replica
key may be disabled in the local HSM but not yet in LDAP and
ipa-ods-exporter believes that it is a new replica key that needs to be
imported from LDAP to local hsm. The op fails as there is already the key
in the local HSM.

The error can be ignored, ipa-ods-exporter simply needs to log a warning.

Fixes: https://pagure.io/freeipa/issue/8654

### DIFF
--- a/daemons/dnssec/ipa-ods-exporter.in
+++ b/daemons/dnssec/ipa-ods-exporter.in
@@ -375,7 +375,18 @@ def master2ldap_master_keys_sync(ldapkeydb, localhsm):
                      str_hexlify(mkey_id), hex_set(new_replica_keys))
 
         # wrap master key with new replica keys
-        mkey_local = localhsm.find_keys(id=mkey_id).popitem()[1]
+        try:
+            mkey_local = localhsm.find_keys(id=mkey_id).popitem()[1]
+        except KeyError:
+            # The master key is present in LDAP but could not be found
+            # in the local HSM. Let's hope it's not the active key,
+            # log an error and process the next master key
+            logger.error("master key 0x%s missing in local HSM, "
+                "will not be able to add master key wrapped with "
+                "replica keys",
+                str_hexlify(mkey_id))
+            continue
+
         for replica_key_id in new_replica_keys:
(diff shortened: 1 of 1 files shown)


abbra:
Test results: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ee4b9752-7dc6-11eb-a5bd-fa163e95e74a/report.html

flo-renaud:
@abbra 
thanks for the review. I removed the temp commit and the gating tests are green.

abbra:
ACK added. I also added `ipa-4-9` label. Do you think it should go into earlier branches as well?

flo-renaud:
@abbra since dnssec support is tech preview only in rhel7 I don't think we will ever backport the patch, no need for ipa-4-6 label IMO.

abbra:
Not ipa-4-6 but what about ipa-4-8?

flo-renaud:
master:

* 3b1a56f588b7585bb98440454f2db72360233bb4 dnssec: fix ipa-ods-exporter crash when master key missing
* 73ced07e2eefee137716452198d7cb0d59016c8f dnssec: concurrency issue when disabling old replica key

PR #5617, opened 2021-03-08
Add checks to prevent adding auth indicators to internal IPA services

antoniotorresm:
Authentication indicators should not be enforced against internal
IPA services, since not all users of those services are able to produce
Kerberos tickets with all the auth indicator options. This includes
`host`, `ldap`, `HTTP` and `cifs` in IPA server and `host` and `cifs` in IPA clients.

Fixes: https://pagure.io/freeipa/issue/8206

### DIFF
--- a/ipaserver/install/server/replicainstall.py
+++ b/ipaserver/install/server/replicainstall.py
@@ -770,6 +770,15 @@ def promotion_check_ipa_domain(master_ldap_conn, basedn):
         ))
 
 
+def promotion_check_host_principal_auth_ind(conn, hostdn):
+    entry = conn.get_entry(hostdn, ['krbprincipalauthind'])
+    if 'krbprincipalauthind' in entry:
+        raise RuntimeError(
+            "Client cannot be promoted to a replica if the host principal "
+            "has an authentication indicator set."
+        )
+
+
 @common_cleanup
 @preserve_enrollment_state
 def promote_check(installer):
@@ -956,6 +965,10 @@ def promote_check(installer):
                                      config.master_host_name, None)
 
         promotion_check_ipa_domain(conn, remote_api.env.basedn)
+        hostdn = DN(('fqdn', api.env.host),
+                    api.env.container_host,
+                    api.env.basedn)
+        promotion_check_host_principal_auth_ind(conn, hostdn)
 
         # Make sure that domain fulfills minimal domain level
         # requirement
--- a/ipaserver/plugins/host.py
+++ b/ipaserver/plugins/host.py
@@ -38,7 +38,7 @@ from .baseldap import (LDAPQuery, LDAPObject, LDAPCreate,
                                      LDAPAddAttributeViaOption,
                                      LDAPRemoveAttributeViaOption)
 from .service import (
-    validate_realm, normalize_principal,
+    validate_realm, validate_auth_indicator, normalize_principal,
     set_certificate_attrs, ticket_flags_params, update_krbticketflags,
     set_kerberos_attrs, rename_ipaallowedtoperform_from_ldap,
     rename_ipaallowedtoperform_to_ldap, revoke_certs)
@@ -730,6 +730,8 @@ class host_add(LDAPCreate):
         update_krbticketflags(ldap, entry_attrs, attrs_list, options, False)
         if 'krbticketflags' in entry_attrs:
             entry_attrs['objectclass'].append('krbticketpolicyaux')
+        validate_auth_indicator(entry_attrs)
+
         return dn
 
     def post_callback(self, ldap, dn, entry_attrs, *keys, **options):
@@ -994,6 +996,7 @@ class host_mod(LDAPUpdate):
             if 'krbprincipalaux' not in (item.lower() for item in
                                          entry_attrs['objectclass']):
                 entry_attrs['objectclass'].append('krbprincipalaux')
+            validate_auth_indicator(entry_attrs)
 
         add_sshpubkey_to_attrs_pre(self.context, attrs_list)
 
--- a/ipaserver/plugins/service.py
+++ b/ipaserver/plugins/service.py
@@ -201,6 +201,28 @@ def validate_realm(ugettext, principal):
         raise errors.RealmMismatch()
 
 
+def validate_auth_indicator(entry):
+    new_value = entry.get('krbprincipalauthind', None)
+    if not new_value:
+        return
+    # The following services are considered internal IPA services
+    # and shouldn't be allowed to have auth indicators.
+    # https://pagure.io/freeipa/issue/8206
+    pkey = api.Object['service'].get_primary_key_from_dn(entry.dn)
+    principal = kerberos.Principal(pkey)
+    server = api.Command.server_find(principal.hostname)['result']
+    if server:
+        prefixes = ("host", "cifs", "ldap", "HTTP")
+    else:
+        prefixes = ("cifs",)
+    if principal.service_name in prefixes:
+        raise errors.ValidationError(
+            name='krbprincipalauthind',
+            error=_('authentication indicators not allowed '
+                    'in service "%s"' % principal.service_name)
+        )
+
+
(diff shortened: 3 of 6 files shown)


abbra:
LGTM.

abbra:
Retracting my ACK. I think there is incorrect logic for client side principals. It prevents setting additional authentication indicators on the hosts where one doesn't have Samba configured. E.g. with the change you wouldn't be able to set `host/client.ipa.test` to require an OTP indicator even if this host has nothing to do with hosting Samba or being IPA replica.

Also, I think we need to add a check into replica promotion code. Suppose you have a client enrolled into IPA that is later decided to be promoted to a replica. If this client's host principal was marked as requiring OTP auth indicator, promoting this host to be a replica would break at the replica connectivity check. So I think there need to be more changes.

antoniotorresm:
Actually, I think it is not possible to add auth indicators to host principals at all. [Host principals are not visible in the `service` module](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/identity_management_guide/renaming-machines), which is the only way to add auth indicators.

```
[root@master ~]# ipa service-show host/master.ipa.local
ipa: ERROR: host/master.ipa.local@IPA.LOCAL: service not found
[root@master ~]# ipa service-add host/master.ipa.local
ipa: ERROR: You must enroll a host in order to create a host service
[root@master ~]# ipa service-add host/master.ipa.local --force
ipa: ERROR: service with name "host/master.ipa.local@IPA.LOCAL" already exists
```
If I understand it correctly it is safe to remove the `host` filter from both server and client, since these services can't be accessed by the module at all. What do you think?

ipaserver/plugins/service.py
```diff
@@ -568,6 +568,27 @@ def validate_ipakrbauthzdata(self, entry):
...
+
+        # The following services are considered internal IPA services
+        # and shouldn't be allowed to have auth indicators.
+        # https://pagure.io/freeipa/issue/8206
+        server_services = ('host', 'ldap', 'HTTP', 'cifs')
+        client_services = ('host', 'cifs')
+        services = server_services if api.env.in_server else client_services
+        name = unicode(self.get_primary_key_from_dn(entry.dn))
+        for service in services:
+            if service + '/' in name:
```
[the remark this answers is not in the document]
tiran: Explanation: get_primary_key_from_dn() returns a str (text). str.startswith() also accepts a tuple of strings.

Also ``ipa.env.in_server`` is wrong. Since the code is executed in ipaserver code, the flag is always true. You have to convert entry to a ``ipapython.kerberos.Principal`` object, then check if the hostname is an IPA server with ``host_is_master()`` or ``server_find()``.

rcritten:
It's probably worth rebasing this PR.

The host principal lives in the host. You can use ipa host-mod --auth-ind=<whatever> ipa.example.test 

You can import the function you added in service.py into host for validation.

antoniotorresm:
OK, PR rebased.

I have also added changes regarding your proposals:
* the check is now also performed in `host-add` and `host-mod` (to handle host principals).
* clients are allowed to have auth indicators in host principals
* client to replica promotion now fails if client has an auth indicator set in its host principal
* tests for changes on both plugins and also on promotion integration tests

rcritten:
This is looking good to me, we just need to wait until we can get a full CI run.

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -101,6 +101,40 @@ def test_one_step_install_pwd_and_admin_pwd(self):
...
+                            '--domain', self.master.domain.name,
+                            '--realm', self.master.domain.realm,
+                            '-p', 'admin',
+                            '-w', self.master.config.admin_password,
+                            '--server', self.master.hostname,
+                            '--force-join'])
+        # Configure firewall first
+        Firewall(client).enable_services(["freeipa-ldap",
+                                          "freeipa-ldaps"])
+        client.run_command(['ipa', 'host-mod', '--auth-ind="otp"',
```
rcritten, line 122: You need to add tasks.kinit_admin(client)

antoniotorresm:
Tests are passing correctly now.

ipaserver/plugins/host.py
```diff
@@ -730,6 +730,9 @@ def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys, **options):
         update_krbticketflags(ldap, entry_attrs, attrs_list, options, False)
         if 'krbticketflags' in entry_attrs:
             entry_attrs['objectclass'].append('krbticketpolicyaux')
+        if 'krbprincipalauthind' in entry_attrs:
+            validate_auth_indicator(entry_attrs)
```
rcritten, line 734: I think you can safely always run this validator as the same check is done at lower levels.

ipaserver/plugins/service.py
```diff
@@ -201,6 +201,28 @@ def validate_realm(ugettext, principal):
...
+    # and shouldn't be allowed to have auth indicators.
+    # https://pagure.io/freeipa/issue/8206
+    pkey = api.Object['service'].get_primary_key_from_dn(entry.dn)
+    principal = kerberos.Principal(pkey)
+    servers = (s['cn'][0] for s in api.Command.server_find()['result'])
+    if principal.hostname in servers:
+        prefixes = ("host", "cifs", "ldap", "HTTP")
+    else:
+        prefixes = ("cifs")
+    if principal.service_name.startswith(prefixes):
```
rcritten, line 218: I think we want exact matches because this may catch other random principals that users might create. The service prefixes are basically arbitrary, as you can tell because of different casing, etc

So this instead:
if principal.service_name in prefixes:

antoniotorresm:
Changes applied, tests passing :)

antoniotorresm:
So, is there anything more needed?

ipaserver/install/server/replicainstall.py
```diff
@@ -770,6 +770,15 @@ def promotion_check_ipa_domain(master_ldap_conn, basedn):
         ))
 
 
+def promotion_check_host_principal_auth_ind(master_ldap_conn, hostdn):
+    entry = master_ldap_conn.get_entry(hostdn, ['krbprincipalauthind'])
```
rcritten, line 774: We are trying to get away from the word master. I think just conn is fine.

rcritten: Otherwise this is fine. IMHO you can make this change and drop the temp commit at the same time.

LGTM.

rcritten:
fedora-latest/test_external_ca_TestExternalCA timed out, re-running

ipaserver/plugins/service.py
```diff
@@ -201,6 +201,28 @@ def validate_realm(ugettext, principal):
...
+def validate_auth_indicator(entry):
+    new_value = entry.get('krbprincipalauthind', None)
+    if not new_value:
+        return
+    # The following services are considered internal IPA services
+    # and shouldn't be allowed to have auth indicators.
+    # https://pagure.io/freeipa/issue/8206
+    pkey = api.Object['service'].get_primary_key_from_dn(entry.dn)
+    principal = kerberos.Principal(pkey)
+    servers = (s['cn'][0] for s in api.Command.server_find()['result'])
```
abbra, line 213: I think this has a potential for a performance issue. Instead of doing blind search for all IPA servers, please do use per-name search `ipa server-find --name={name}`.

abbra:
I removed ACK again because there is a need to improve a server search.

rcritten:
Re-run, a couple of tests timed out.

rcritten:
The changes LGTM, @abbra what do you think?

abbra:
Yes, looks fine to me too. Thank you!

flo-renaud:
master:

* 0bdbf11442024715e405b8f1fc27eb2fce4f4a9c Add checks to prevent adding auth indicators to internal IPA services
* da72a576586f40d3378a3cb376ba0af510c74b86 ipatests: ensure auth indicators can't be added to internal IPA services

PR #5632, opened 2021-03-10
ipatests: re-add test_dnssec.py::TestInstallDNSSECFirst in gating

flo-renaud:
The test was temporarily removed because of a known issue
but the issue is now fixed.

Related: https://pagure.io/freeipa/issue/8496

### DIFF
--- a/ipatests/pytest_ipa/integration/resolver.py
+++ b/ipatests/pytest_ipa/integration/resolver.py
@@ -1,6 +1,7 @@
 import os
 import abc
 import logging
+import re
 import textwrap
 import time
 
@@ -131,6 +132,16 @@ class Resolver(abc.ABC):
         :param state: internal state object  specific to subclass implementaion
         """
 
+    def uses_localhost_as_dns(self):
+        """Return true if the localhost is set as DNS server.
+
+        Default implementation checks the content of /etc/resolv.conf
+        """
+        resolvconf = self.host.get_file_contents(paths.RESOLV_CONF, 'utf-8')
+        patterns = [r"^\s*nameserver\s+127\.0\.0\.1\s*$",
+                    r"^\s*nameserver\s+::1\s*$"]
+        return any(re.search(p, resolvconf, re.MULTILINE) for p in patterns)
+
 
 class ResolvedResolver(Resolver):
     RESOLVED_RESOLV_CONF = {
@@ -191,6 +202,17 @@ class ResolvedResolver(Resolver):
             self.host.put_file_contents(
                 self.RESOLVED_CONF_FILE, state['resolved_config'])
         self._restart_resolved()
+
+    def uses_localhost_as_dns(self):
+        """Return true if the localhost is set as DNS server.
+
+        When systemd-resolved is in use, the DNS can be found using
+        the command resolvectldns.
+        """
+        dnsconf = self.host.run_command(['resolvectl', 'dns']).stdout_text
+        patterns = [r"^Global:.*\s+127.0.0.1\s+.*$",
+                    r"^Global:.*\s+::1\s+.*$"]
+        return any(re.search(p, dnsconf, re.MULTILINE) for p in patterns)
 
 
 class PlainFileResolver(Resolver):
--- a/ipatests/azure/azure_definitions/gating-fedora.yml
+++ b/ipatests/azure/azure_definitions/gating-fedora.yml
@@ -26,12 +26,11 @@ vms:
     - test_integration/test_membermanager.py
 
 - vm_jobs:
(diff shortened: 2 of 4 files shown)


flo-renaud:
Also added a fix for https://pagure.io/freeipa/issue/8695. Since it's related to resolvers, @wladich could you have a look?

rcritten:
LGTM. Can you confirm what branches this needs backports for? I see the change in ipa-4-8 and ipa-4-9.

wladich (commented):
@flo-renaud 
Do you think the new method `Resolver.uses_localhost_as_dns` is expected to be of any use outside of the `test_resolvconf` test case, especially in such strict form - checking for localhost only, not any ip address? I prefer not to place code which is not expected to be reused to common libraries like this one.

The type of resolver can be detected in the test with `isinstance()` or we could add a string property "resolver_type" to all Resolver subclasses.

What do you think about changing the test from checking config files contents (which is not perfect in current state, see inline comment) to checking if ipa server is actually using localhost as a resolver? For that we could create a unique DNS record  and query it with `dig` or `host` commands (or even both as they use different algorithms for obtaining system resolver).

Another thing - the name of test does not reflect the actual check as we are not checking the resolv.conf in case of Fedora 33. Maybe it is better to call it according to  test goal - something like "test_localhost_resolver_used".

flo-renaud:
> @flo-renaud
> Do you think the new method `Resolver.uses_localhost_as_dns` is expected to be of any use outside of the `test_resolvconf` test case, especially in such strict form - checking for localhost only, not any ip address? I prefer not to place code which is not expected to be reused to common libraries like this one.

I don't agree with this comment. Resolver is also a clean way for a test to call methods without the hastle of checking which resolver is in place. If I move the logic into the test, it means that the test would need to be updated whenever a new Resolver implementation is available.

> What do you think about changing the test from checking config files contents (which is not perfect in current state, see inline comment) to checking if ipa server is actually using localhost as a resolver? For that we could create a unique DNS record and query it with `dig` or `host` commands (or even both as they use different algorithms for obtaining system resolver).

I updated the current code with a pattern and regexp search. It will probably be way faster than adding a DNS record.

> Another thing - the name of test does not reflect the actual check as we are not checking the resolv.conf in case of Fedora 33. Maybe it is better to call it according to test goal - something like "test_localhost_resolver_used".

Updated.

flo-renaud:
Link to successful run with fedora33: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/4e9af5ae-86f9-11eb-a883-fa163e2b2ac7)
Link to successful run woth fedora32: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/4d342ce8-86fa-11eb-8ec3-fa163ec3b944)
Removing temp commit.

ipatests/pytest_ipa/integration/resolver.py
```diff
@@ -192,6 +203,15 @@ def _apply_state(self, state):
...
         self._restart_resolved()
 
+    def uses_localhost_as_dns(self):
+        """Return true if the localhost is set as DNS server.
+
+        When systemd-resolved is in use, the DNS can be found using
+        the command resolvectldns.
+        """
+        dnsconf = self.host.run_command(['resolvectl', 'dns'])
+        return any(ip in dnsconf.stdout_text for ip in ('127.0.0.1', '::1'))
```
wladich, line 213: Sorry, I did not notice the issue in previous review: this check will match "127.0.0.127".  
Also It will be nice to check that the address is in the "Global" section.

The output of `resolvectl dns` with IPA installed with DNS support is something like
```
# resolvectl dns
Global: 127.0.0.1
Link 2 (eth0): 10.11.5.19 10.5.30.45
```

wladich (commented):
I see your point. You also mentioned that checking resolver with `dig`  can produce a false positive in case system uses some other  nameserver which has a delegation to IPA server. So the changes mostly LGTM, please see the inline comment.

flo-renaud:
The first commit (re-enable test in gating) needs to be backported up to ipa-4-8. The second one can go to ipa-4-9 only (it needs the [DNS Resolver](https://github.com/freeipa/freeipa/pull/5474) commit that was backported to ipa-4-9).

wladich:
Thank you, ACK

flo-renaud:
master:

* a9b4ed4f528b01076403750872ddb899532b6c01 ipatests: re-add test_dnssec.py::TestInstallDNSSECFirst in gating
* fb107b9180ad1d679d2b0842536cc9cb907b96f5 ipatests: fix TestInstalDNSSECFirst::test_resolvconf logic

PR #5633, opened 2021-03-11
Return nsaccountlock in user-add as boolean

antoniotorresm:
The `nsaccountlock` attribute was being returned as a
list of string ("TRUE"/"FALSE") instead of a boolean.
Use the convert function used in `user-find` and `user-mod`
for consistency, since these commands return the parameter as a boolean.

Fixes: https://pagure.io/freeipa/issue/8743

### DIFF
--- a/ipaserver/plugins/baseuser.py
+++ b/ipaserver/plugins/baseuser.py
@@ -545,6 +545,8 @@ class baseuser_add(LDAPCreate):
         self.obj.convert_usercertificate_post(entry_attrs, **options)
         self.obj.get_password_attributes(ldap, dn, entry_attrs)
         convert_sshpubkey_post(entry_attrs)
+        if 'nsaccountlock' in entry_attrs:
+            convert_nsaccountlock(entry_attrs)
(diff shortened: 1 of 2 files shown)


rcritten:
nit: the value was returned as a list of string instead of a boolean.

antoniotorresm:
That's right, commit message updated :)

rcritten:
Some of the xmlrpc tests are failing because they still expect a list and not a boolean.

rcritten:
Test changes look good. Can you add the ticket to that commit message as well?

rcritten:
One last question: Is this for the master branch only or backport to ipa-4-9?

abbra:
The original report was done against Fedora 33, thus ipa-4-9 backport is needed.

abbra:
master:

* 45de2d7f393e22288ae3d804b97ecadf5fc29c9c Return nsaccountlock in user-add as boolean
* 8caac84146cd633f8ccd3501fe220e74a30af466 ipatests: expect boolean type for nsaccountlock in user module

PR #5634, opened 2021-03-11
Extend logging to include execution time

antoniotorresm:
Adding execution time in logs provides useful information
for identifying API operations that impact IPA performance.

Related: https://pagure.io/freeipa/issue/8759

### DIFF
--- a/ipaserver/rpcserver.py
+++ b/ipaserver/rpcserver.py
@@ -28,6 +28,7 @@ from __future__ import absolute_import
 import logging
 from xml.sax.saxutils import escape
 import os
+import time
 import traceback
 from io import BytesIO
 from urllib.parse import parse_qs
@@ -374,6 +375,8 @@ class WSGIExecutioner(Executioner):
             return self.marshal(result, RefererError(referer='missing'), _id)
         if not environ['HTTP_REFERER'].startswith('https://%s/ipa' % self.api.env.host) and not self.env.in_tree:
             return self.marshal(result, RefererError(referer=environ['HTTP_REFERER']), _id)
+        if self.api.env.debug:
+            time_start = time.perf_counter_ns()
         try:
             if ('HTTP_ACCEPT_LANGUAGE' in environ):
                 lang_reg_w_q = environ['HTTP_ACCEPT_LANGUAGE'].split(',')[0]
@@ -414,6 +417,8 @@ class WSGIExecutioner(Executioner):
             try:
                 params = command.args_options_2_params(*args, **options)
             except Exception as e:
+                if self.api.env.debug:
+                    time_end = time.perf_counter_ns()
                 logger.info(
                    'exception %s caught when converting options: %s',
                    e.__class__.__name__, str(e)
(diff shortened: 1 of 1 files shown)


ipaserver/rpcserver.py
```diff
@@ -368,12 +369,15 @@ def wsgi_execute(self, environ):
         args = ()
         options = {}
         command = None
+        time_start = None
+        time_end = None
```
rcritten, line 373: Do these need to be defined here? I guess it could be good practice, but I don't see a way that either could be used without calling time.perf_counter() first.

abbra:
I am concerned by timer calls in the normal code paths. This might turn to be way more expensive than expected. In some situations clock_gettime() which is used by the Python's time.perf_counter() since 2018 will have to do a real syscall and slow things down. This is, in particular, an issue on xen-based AWS instances.

Another part is a change of output format for the logger in one case but not the other one.

rcritten:
Thomas had similar performance concerns. We can wrap it with a conditional if desired, something that can be set in the IPA config either by file or bootstrap.

I'm not sure which formatting cases you're referring to, but if we add a conditional that will also change the format depending on configuration. The value is tacked onto the end of the line as the best we can do. If there are any log parsers as long as they don't have a trailing $ in the regex then its likely that this won't break them.

abbra:
We can have that conditional to print the timing on a separate line. Or add it to the json instead.

rcritten:
A separate line would work but we'd basically have to log a lot of the same information for it to be useful.

I'm not sure what you mean by json. If you mean in the data returned to the client that would make analyzing the data a lot more difficult and less central.

abbra:
That's also true. On the other hand, we have 'raw' logging when `debug=true` is set in the `server` context. May be extend this raw format instead of the main log lines?

antoniotorresm:
I have updated the PR, now time is only recorded when debug mode is enabled, and a second debug line is printed on command success showing exactly the same information + time.
```

[Mon Mar 15 12:32:13.467490 2021] [wsgi:error] [pid 14895:tid 15355] [remote fe80::f816:3eff:fedd:a779%eth0:33040] ipa: INFO: [jsonserver_session] admin@IPA.LOCAL: user_add/1('testuser2', givenname='test', sn='test', version='2.240'): SUCCESS
[Mon Mar 15 12:32:13.467588 2021] [wsgi:error] [pid 14895:tid 15355] [remote fe80::f816:3eff:fedd:a779%eth0:33040] ipa: DEBUG: [jsonserver_session] admin@IPA.LOCAL: user_add/1('testuser2', givenname='test', sn='test', version='2.240'): SUCCESS 0.24110692400063272
```

rcritten:
This works for me.

rcritten:
We like to have upstream tickets associated with commits. Can you create an upstream RFE for this and link it in the commit message?

antoniotorresm:
Done: https://pagure.io/freeipa/issue/8759

abbra:
Thanks for this change. I have only one question left: what about returning time in nanoseconds instead of fractions of a second?  E.g. use `perf_counter_ns()` instead. I can imagine that it will be a tiny bit faster also on systems like Raspberry Pi.

antoniotorresm:
OK, now it's using nanoseconds. Thomas also suggested adding a tag to the number, I added `etime` since that's how it appears in LDAP logs.

rcritten:
Talked to @abbra offline and got his ack.

rcritten:
master:

* 4d716d3fbc69760bc0f7bd1a7c83ab14f1a62f18 Extend logging to include execution time

PR #5637, opened 2021-03-16
ipatests: Test to check that ResponseNotReady error is not displayed when user session cache is deleted

menonsudhir:
This testcase checks that http.client.ResponseNotReady error is not displayed when
user session cache is deleted.

https://pagure.io/freeipa/issue/7752

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -1489,3 +1489,36 @@ class TestIPACommand(IntegrationTest):
         self.master.run_command(['date', '-s', '-15Years'])
 
         assert isrgrootx1_nick in result
+
+
+class TestIPACommandWithoutReplica(IntegrationTest):
+    """
+    Execute tests with scenarios having only single
+    IPA server and no replica
+    """
+    @classmethod
+    def install(cls, mh):
+        tasks.install_master(cls.master, setup_dns=True)
+
+    def test_client_doesnot_throw_responsenotready_error(self):
+        """
+        This testcase checks that ipa command
+        doesn't throw http.client.ResponseNotReady error
+        when current users session is deleted from the cache
+        """
+        user = 'ipauser1'
+        orig_pwd = 'Password123'
+
+        tasks.kinit_admin(self.master)
+        tasks.user_add(self.master, user, password=orig_pwd)
+        # kinit as admin on ipa-server and run ipa command
+        tasks.kinit_admin(self.master, raiseonerr=False)
+        self.master.run_command(['ipa', 'user-show', "ipauser1"])
+        # Delete the current user session cache on IPA server
+        self.master.run_command(
+            "rm -fv /run/ipa/ccaches/admin@{}-*".format(
+                self.master.domain.realm
+            )
+        )
+        # Run the command again after cache is removed
+        self.master.run_command(['ipa', 'user-show', 'ipauser1'])


ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,28 @@ def test_ipa_cacert_manage_prune(self):
...
+        """
+        This testcase checks that ipa command
+        doesn't throw http.client.ResponseNotReady error
+        when current users session is deleted from the cache
+        """
+        error_msg = (
+            "http.client.ResponseNotReady: Request-sent \n"
+            "ipa: ERROR: an internal error has occurred"
+        )
+        domainname = "ipa.test"
```
mrizwan93, line 1503: Use `self.master.domain.name` instead of hardcode

ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,28 @@ def test_ipa_cacert_manage_prune(self):
...
+        del_ccache = "rm -fv /run/ipa/ccaches/admin@{}".format(domainname)
```
mrizwan93, line 1504: Can use f string like `f"rm -fv /run/ipa/ccaches/admin@{domainname}"`

ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,28 @@ def test_ipa_cacert_manage_prune(self):
...
+
+        # kinit as admin on ipa-client and run ipa command
+        tasks.kinit_admin(self.clients[0], raiseonerr=False)
+        self.clients[0].run_command(["ipa", "user-find"])
```
mrizwan93, line 1508: What is the purpose of this, are you expecting something? Can use assert if that's the case.

ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,28 @@ def test_ipa_cacert_manage_prune(self):
...
+        """
+        error_msg = (
+            "http.client.ResponseNotReady: Request-sent \n"
+            "ipa: ERROR: an internal error has occurred"
+        )
+        domainname = "ipa.test"
+        del_ccache = "rm -fv /run/ipa/ccaches/admin@{}".format(domainname)
+
+        # kinit as admin on ipa-client and run ipa command
+        tasks.kinit_admin(self.clients[0], raiseonerr=False)
```
mrizwan93, line 1507: any specific reason to use `raiseonerr=False`?

ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,28 @@ def test_ipa_cacert_manage_prune(self):
...
+        self.clients[0].run_command(["ipa", "user-find"])
+
+        # Delete the current user session cache on IPA server
+        cmd = self.master.run_command([del_ccache])
```
mrizwan93, line 1511: You need to add `' '` to command like `cmd = self.master.run_command(['del_ccache'])`

ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,28 @@ def test_ipa_cacert_manage_prune(self):
...
+
+        # Run the command on the client again after cache is removed
+        cmd = self.clients[0].run_command(["ipa", "user-find"])
+        assert cmd.returncode == 0
```
mrizwan93, line 1515: No need to add this, `raiseonerr=True` is set by default.

mrizwan93 (commented):
Hello @menonsudhir 
Thanks for the patch. Please see inline comments.

ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-latest/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
+        test_suite: test_integration/test_commands.py::TestIPACommand::test_client_doesnot_throw_responsenotready_error
```
amore17, line 71: In temp_commit please use only test_integration/test_commands.py::TestIPACommand

ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,26 @@ def test_ipa_cacert_manage_prune(self):
...
+        tasks.kinit_admin(self.clients[0], raiseonerr=False)
+        self.clients[0].run_command(['ipa', 'user-find'])
+
+        # Delete the current user session cache on IPA server
+        self.master.run_command(['rm', '-fv', '/run/ipa/ccaches/*'])
+
+        # Run the command on the client again after cache is removed
+        cmd = self.clients[0].run_command(['ipa', 'user-find'])
+        assert cmd.returncode == 0
+        assert error_msg not in cmd.stdout_text
```
amore17, line 1514: Please also add check for `stderr_text`
eg: assert error not in (result.stdout_text, result.stderr_text)

amore17 (changes requested):

ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,26 @@ def test_ipa_cacert_manage_prune(self):
...
         assert isrgrootx1_nick in result
+
+    def test_client_doesnot_throw_responsenotready_error(self):
+        """
+        This testcase checks that ipa command
+        doesn't throw http.client.ResponseNotReady error
+        when current users session is deleted from the cache
+        """
+        error_msg = (
+            "http.client.ResponseNotReady: Request-sent \n"
```
flo-renaud, line 1500: Is there really a space at the end of the string when the issue happens?

ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,26 @@ def test_ipa_cacert_manage_prune(self):
...
+        tasks.kinit_admin(self.clients[0], raiseonerr=False)
+        self.clients[0].run_command(['ipa', 'user-find'])
+
+        # Delete the current user session cache on IPA server
+        self.master.run_command(['rm', '-fv', '/run/ipa/ccaches/*'])
+
+        # Run the command on the client again after cache is removed
+        cmd = self.clients[0].run_command(['ipa', 'user-find'])
+        assert cmd.returncode == 0
+        assert error_msg not in (cmd.stdout_text, cmd.stderr_text)
```
flo-renaud, line 1514: should be `not in (cmd.stdout_text + cmd.stderr_text)` otherwise the assert is equivalent to `error_msg != cmd.stdout_text and error_msg != cmd.stderr_text` which is a different test:
```
>>> assert "substring" not in ("substring123", "otherstring")
>>> assert "substring" not in ("substring123" + "otherstring")
AssertionError
```

flo-renaud (commented):
Hi @menonsudhir 
please find inline comments.

ipatests/test_integration/test_commands.py
menonsudhir: @flo-renaud i did see the space at the end of the string.

ipatests/test_integration/test_commands.py
menonsudhir: @flo-renaud have made he requested change

flo-renaud:
Hi @menonsudhir 
if you look at the traces of the temp commit in [report.html](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/eadfe3fc-8ae1-11eb-88eb-fa163ebd809e/report.html), you can see that the rm command does not remove any file in `/run/ipa/ccaches`:
```
INFO     ipatests.pytest_ipa.integration.host.Host.master.IPAOpenSSHTransport:transport.py:391 RUN ['rm', '-fv', '/run/ipa/ccaches/*']
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd509:transport.py:513 RUN ['rm', '-fv', '/run/ipa/ccaches/*']
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd509:transport.py:217 Exit code: 0
```

It should have found at least one file for the admin that just did `ipa user-find`. This means there is an issue with the test and it won't detect any regression.
I also tried to run the test on a set of machines where the fix was reverted, the test wasn't able to detect the issue.

ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,26 @@ def test_ipa_cacert_manage_prune(self):
...
+            "http.client.ResponseNotReady: Request-sent \n"
+            "ipa: ERROR: an internal error has occurred"
+        )
+
+        # kinit as admin on ipa-client and run ipa command
+        tasks.kinit_admin(self.clients[0], raiseonerr=False)
+        self.clients[0].run_command(['ipa', 'user-find'])
+
+        # Delete the current user session cache on IPA server
+        self.master.run_command(['rm', '-fv', '/run/ipa/ccaches/*'])
```
flo-renaud, line 1509: According to https://github.com/encukou/pytest-multihost/blob/master/pytest_multihost/host.py#L262, run_command with a list of args does not provide shell expansion. You can either replace your line with `run_command("rm -fv /run/ipa/ccaches/*")` or do as in https://github.com/freeipa/freeipa/blob/master/ipatests/pytest_ipa/integration/tasks.py#L938

ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,26 @@ def test_ipa_cacert_manage_prune(self):
...
+            "http.client.ResponseNotReady: Request-sent \n"
+            "ipa: ERROR: an internal error has occurred"
+        )
+
+        # kinit as admin on ipa-client and run ipa command
+        tasks.kinit_admin(self.clients[0], raiseonerr=False)
+        self.clients[0].run_command(['ipa', 'user-find'])
+
+        # Delete the current user session cache on IPA server
+        self.master.run_command(['rm -fv /run/ipa/ccaches/*'])
```
flo-renaud, line 1509: Remove the brackets to provide a string instead of a list of strings: `self.master.run_command('rm -fv /run/ipa/ccaches/*')`

flo-renaud:
Hi @menonsudhir 
I launched your new test on an instance where the fix for 7752 was reverted. I expected the test to fail and it didn't, which means that there is an issue with the test.
You can see [here](https://github.com/freeipa-pr-ci2/freeipa/pull/801) the output.

I'm wondering if it's because there is a replica in the test. You can use `ipa -d -vv user-find` to have more information.

menonsudhir:
> ipa -d -vv user-find



> Hi @menonsudhir
> I launched your new test on an instance where the fix for 7752 was reverted. I expected the test to fail and it didn't, which means that there is an issue with the test.
> You can see [here](https://github.com/freeipa-pr-ci2/freeipa/pull/801) the output.
> 
> I'm wondering if it's because there is a replica in the test. You can use `ipa -d -vv user-find` to have more information.

@flo-renaud thank you for the review once again, have made the changes and triggered a run for that specific test.

flo-renaud:
The logs show that the 2nd ipa user-find command tries first with the master then moves to the replica:
```
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557 ipa: DEBUG: New HTTP connection (master.ipa.test)
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557 ipa: DEBUG: HTTP connection destroyed (master.ipa.test)
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557 Traceback (most recent call last):
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557   File "/usr/lib/python3.9/site-packages/ipalib/rpc.py", line 728, in single_request
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557     raise ProtocolError(
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557 xmlrpc.client.ProtocolError: <ProtocolError for master.ipa.test/ipa/session/json: 500 Internal Server Error>
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557 ipa: INFO: Connection to https://master.ipa.test/ipa/session/json failed with <ProtocolError for master.ipa.test/ipa/session/json: 500 Internal Server Error>
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557 ipa: DEBUG: trying https://replica0.ipa.test/ipa/session/json
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557 ipa: INFO: Request: {
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557     "id": 0,
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557     "method": "ping",
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557     "params": [
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557         [],
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557         {}
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557     ]
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557 }
DEBUG    ipatests.pytest_ipa.integration.host.Host.client0.cmd49:transport.py:557 ipa: DEBUG: New HTTP connection (replica0.ipa.test)
```
You will need to modify the test in order to have a single master and no replica, otherwise the test is meaningless and wouldn't reproduce the issue.

flo-renaud:
@menonsudhir 
I am not sure that your test scenario is exactly the same as what is described in the reproducer steps from https://bugzilla.redhat.com/show_bug.cgi?id=1639890#c15:
- there is only a server involved, no client
- the admin cache is deleted, not all the files in /run/ipa/ccaches/

flo-renaud:
Hi @menonsudhir 
the server behavior has changed since this commit: https://pagure.io/freeipa/c/cf80bb33baa79ba93987ee7e63109ece4b37bbd0
Before the commit, a missing session cache in /run/ipa/ccaches/admin@IPA.TEST-xxx did not cause any issue (the server was sending a 401 response) while we now have the server sending a 500 response.

@stanislavlevin or @rcritten can you have a look at the commit https://pagure.io/freeipa/c/cf80bb33baa79ba93987ee7e63109ece4b37bbd0, especially the diff related to the `get_credentials()` method in ipalib/krb_utils.py. I don't have a good understanding of all the consequences of this code change. When the session cache is missing, the client doesn't retry but is it really intended?

flo-renaud:
Removing the stale label.

@rcritten can you have a look at my previous comment?


> Hi @menonsudhir
> the server behavior has changed since this commit: https://pagure.io/freeipa/c/cf80bb33baa79ba93987ee7e63109ece4b37bbd0
> Before the commit, a missing session cache in /run/ipa/ccaches/admin@IPA.TEST-xxx did not cause any issue (the server was sending a 401 response) while we now have the server sending a 500 response.
> 
> @stanislavlevin or @rcritten can you have a look at the commit https://pagure.io/freeipa/c/cf80bb33baa79ba93987ee7e63109ece4b37bbd0, especially the diff related to the `get_credentials()` method in ipalib/krb_utils.py. I don't have a good understanding of all the consequences of this code change. When the session cache is missing, the client doesn't retry but is it really intended?

stanislavlevin:
Sorry, I missed this, will check today.

rcritten:
The 500 is due to an uncaught exception:

```
File "/usr/share/ipa/wsgi.py", line 59, in application
  return api.Backend.wsgi_dispatch(environ, start_response)
File "/usr/lib/python3.9/site-packages/ipaserver/rpcserver.py", line 296, in __call__
  return self.route(environ, start_response)
File "/usr/lib/python3.9/site-packages/ipaserver/rpcserver.py", line 308, in route
  return app(environ, start_response)
File "/usr/lib/python3.9/site-packages/ipaserver/rpcserver.py", line 894, in __call__
  ccache_name = self.get_environ_creds(environ)
File "/usr/lib/python3.9/site-packages/ipaserver/rpcserver.py", line 672, in get_environ_creds
  creds = get_credentials_if_valid(name=gss_name,
File "/usr/lib/python3.9/site-packages/ipalib/krb_utils.py", line 199, in get_credentials_if_valid
  creds = get_credentials(name=name, ccache_name=ccache_name)
File "/usr/lib/python3.9/site-packages/ipalib/krb_utils.py", line 158, in get_credentials
  raise ValueError('"%s", ccache="%s"' % (e, ccache_name))
ValueError: "Major (851968): Unspecified GSS failure.  Minor code may provide more information, Minor (2598845123): No credentials cache found
```

And it looks like even the purposefully raised exception wouldn't be handled. The caller of ```get_environ_creds()``` excepts either a ccache or None.

So the right  call I think is to catch ValueError around the call to ```get_credentials_if_valid()``` and return None if caught.

rcritten:
I filed https://pagure.io/freeipa/issue/8873 and will work on a fix.

rcritten:
https://github.com/freeipa/freeipa/pull/5808

flo-renaud:
@menonsudhir 
Rob fixed the regression and the fix has been merged in master branch. Can you rebase this PR to pick the fix? This way we'll be able to check if the test is OK. Thanks

menonsudhir:
> @menonsudhir
> Rob fixed the regression and the fix has been merged in master branch. Can you rebase this PR to pick the fix? This way we'll be able to check if the test is OK. Thanks

@flo-renaud i have just now rebase the PR

ipatests/test_integration/test_commands.py
flo-renaud: I agree with Rizwan, by default raiseonerr=True and the test will fail before reaching the `assert cmd.returncode == 0` line.
You have 2 choices:
- if you want to use the default raiseonerr=True, just call ipa user-show and delete the assertions on line 1529, 1520 + delete the error_msg on line 1514 as it won't be used. If ipa user-show fails, the test will exit on failure.
- if you want to check the output, add raiseonerr = True. This way you can check that error_msg is not in the output but in this case I would invert the 2 assertions.
IMO no need to look for a specific error message, you can stick to the default raiseonerr=True behavior.

flo-renaud:
@menonsudhir 
not sure whether you saw my  previous comment at https://github.com/freeipa/freeipa/pull/5637/files#r645593252 ...

ipatests/test_integration/test_commands.py
```diff
@@ -1489,3 +1489,37 @@ def test_ipa_cacert_manage_prune(self):
...
+        tasks.kinit_admin(self.master, raiseonerr=False)
+        self.master.run_command(['ipa', 'user-show', "ipauser1"])
+        # Delete the current user session cache on IPA server
+        self.master.run_command(
+            "rm -fv /run/ipa/ccaches/admin@{}-*".format(
+                self.master.domain.realm
+            )
+        )
+        # Run the command again after cache is removed
+        cmd = self.master.run_command(['ipa', 'user-show', 'ipauser1'])
```
flo-renaud, line 1524: Please delete the assert after this line, it will never be reached.

flo-renaud:
Hi @menonsudhir 
I ran your test with a version where the fix for https://pagure.io/freeipa/issue/7752 was reverted and it properly triggers the issue (see https://github.com/freeipa-pr-ci2/freeipa/pull/819).
Azure pipelines are still running but when they are green, please remove the temp commit and I will ACK.

flo-renaud:
ACk for when azure pipelines complete

flo-renaud:
master:

* 0c2530efe5e2540722a37c65a6aae8cc65dcde34 ipatests: Test to check that ResponseNotReady error is not displayed when user session cache is deleted

flo-renaud:
@menonsudhir the automated backport failed, can you do a manual backport to ipa-4-9 branch? Thanks

flo-renaud:
Note: the backport is https://github.com/freeipa/freeipa/pull/5851

PR #5640, opened 2021-03-17
Allowed longer RSA key lengths and PSS signing algorithm in certificate profiles

ysyvachenko:
Allowed longer RSA key lengths and PSS signing algorithm in all profiles to be sync with Dogtag and OpenSSL.

 - Allowed using RSA key length up to 16384, this is what OpenSSL recommends to generate. OpenSSL can generate longer keys but shows a warning that it is not recommended.
 - Allowed using RSA-PSS signing algorithms. Starting from version 10.10.5 Dogtag supports an instruction 'pki_use_pss_rsa_signing_algorithm=True',
   which can be enabled in FreeIPA via --pki-config-override. The instruction enforces using RSA-PSS from PKCS#1 v2.1 instead of PKCS#1 v1.5.


abbra:
Thanks for this PR. Few comments:
 - Could you please point me to Dogtag's upstream change that allows 16384? I do not see any of the profiles in PKI enabling anything above 4096
 - Perhaps, you'd want to open a separate PR to PKI to add 8192, 15360, and 16384 there as well
 - Please create a ticket for this change on https://pagure.io/freeipa/new_issue and reference it in the commit message as `Fixes: <url>`, along with your sign-off
 - if use of RSA-PSS requires PKCS#1 v2.1, do we need to bump any specific dependencies ourselves?

ysyvachenko:
Thanks for your comment. 

Finally, I have created a ticket https://pagure.io/freeipa/issue/8777 and a PR in the Dogtag project https://github.com/dogtagpki/pki/pull/3492

As for RSA-PSS, I have not fully tested this in FreeIPA, but I have installed version 4.9.0 with pki_override.cfg where pki_use_pss_rsa_signing_algorithm=True. Installation is successful and WebUI works fine. I have inspected the issued certificates and they seem correct.

In the scope of this work I am playing with ipalib/install/certmonger.py to allow configuring IPA RA key length, but let it be a separate PR.

rcritten:
Existing installations won't get updated profiles.  Only the *UPDATE* profile changes are applied.

ysyvachenko:
May I rephrase to check that I understood correctly. If we have a profile with suffix 'UPGRADE', like caIPAserviceCert.UPGRADE.cfg, the settings in this file override the settings in the caIPAserviceCert.cfg of the upgrade. If so, for me it looks reasonable that new key sizes become available only for new installations. 

For a fresh installation, Admin can pass --pki-config-override and configure Dogtag appropriately to use long keys, but for existing installations, I am not aware if we have such parameter to re-configure Dogtag automatically. I suppose Admin can change Dogtag and FreeIPA profiles manually for existing systems. And also Admin needs to re-issue CA and other certificates to enable long key usage. :)

If I understand 'UPGRADE' logic correctly it seems I need to keep long keys in caIPAserviceCert.UPGRADE.cfg as well, because in another case on the first upgrade of the freshly installed system it becomes corrupted. But this also means that we partially deliver long keys support for existing systems (that I a bit dislike). 

Do you think we need to add .UPGRAGE files for all existing profiles ?

rcritten:
> Do you think we need to add .UPGRAGE files for all existing profiles ?

I haven't quiet made up my mind on it. Ifthe larger sizes are available, along with your forthcoming change to certmonger support, even an existing installation could increase the in-place key sizes with the exception of the CA. But does it really make sense to have server cert sizes >> the CA? I'm not sure how much, if any, pent up demand there is for that. But deploying a whole new IPA system to take advantage of it can be challenging.

rcritten:
Can I ask what the use case for 16k keys are? It would be beyond slow. When I did RSA and ECC testing quite a number of years ago RSA was in the low single digit ops/sec at 7k keys. I didn't even record 15k keys though honestly I don't remember whether that was because it didn't work at all or something else.

ysyvachenko:
The main use case of 16k keys is the mitigation of risks we have with quantum computations. Personally, I cannot feel any difference between 4096 and 16384 keys for ordinary computers, but for quantum computers situation is different. Both keys are easy to crack in terms of quantum computational complexity, but engineers feel some certainty that for 16384 keys we won't have such a large workable quantum computer for the next decade, while for 4096 we will very likely have.

The mitigation action is to select some temporary cryptographic algorithm for usage, where the number of quantum gates needed is maximized. Some estimates: https://quantumcomputing.stackexchange.com/questions/5048/how-many-logical-qubits-are-needed-to-run-shors-algorithm-efficiently-on-large
ECC needs more gates for the same key size comparing to RSA, but from my point of view, RSA scales better, and even 65536 key should work fine in OpenSSL. From estimates, RSA 16384 is better than ECC 521. I checked a few enterprise solutions and they all do support 16384, while larger keys are not widely supported. 

Personally, I have not faced significant performance issues besides key generation. Key generation is slow, but it is acceptable, I believe. If the Administrator decides to use long keys, slow key generation is the first thing they face, thus, it is going to be a considered decision to use 16K keys.

Another option I investigated, is to use multi-prime RSA: four times 4096. This should eliminate performance issues on modern CPUs, but I found that not all devices support it, and mathematics is tricky there, and there is no clear evidence that quantum computers will need a lot of gates in this case.

I do not know any better approach right now to mitigate risks rather than increasing RSA key length. If you have more information, I would very much appreciate your help as I need to deploy a few systems that should be resilient for such attacks. This issue with RSA keys is just a part (e.g. DH is more painful).

Could we progress with pull requests here and in dogtag project?

rcritten:
Kicking off a full CI run.

tiran:
I'm strong -1 on including larger RSA key sizes. A 16,384 bit RSA key takes about 2 minutes on a modern CPU with 4 GHz turbo boost.

rcritten:
Yeah but it's one-time generation and a lifetime of slow but it's user-choice right? It isn't like we're mandating it. Now if being slow would cause a user to make poor choices to reduce security is one thing, but being slow in itself isn't a problem is it? As long as this doesn't create a slew of race conditions anyway.

tiran:
Each TLS handshake is going to be slow, too. Clients have to verify a much larger signature. If users increase the EE cert size, then Servers have to sign the TLS handshake parameters with a much larger key.

@ysyvachenko Could you please provide performance statistics how the increased key size affects TLS handshake size and performance?

ysyvachenko:
@tiran 
Do you have some procedure for getting TLS handshake size and TLS performance?

PR #5573, opened 2021-02-18
Don't start the CA during healthcheck grace period check so uninstall succeeds

rcritten:
We saw the DBusError.NoReply failure in a downstream test.

Exercise it a bit.

PR #5641, opened 2021-03-17
ipatests: Test to check ipa-healthcheck tool displays warning when run on ipa-client

menonsudhir:
The testcase checks that when ipa-healthcheck tool is run on ipa-client it will display 
warning message "IPA is not configured"

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -234,12 +234,14 @@ class TestIpaHealthCheck(IntegrationTest):
     """
 
     num_replicas = 1
+    num_clients = 1
 
     @classmethod
     def install(cls, mh):
         if not cls.master.transport.file_exists(SOS_CMD):
             tasks.install_packages(cls.master, SOS_PKG)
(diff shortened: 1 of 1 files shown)


rcritten:
It looks ok but this test will break with the next upstream release due to https://github.com/freeipa/freeipa-healthcheck/commit/a6897503d73e4d398dedbb713ee8d90d29a683d2

So I think the you should check for both strings. Something like:

valid_msg = ('IPA is not configured', 'IPA server is not configured',)
...
assert cmd.stdout_text in valid_msg

I assume you want this backported to ipa-4-9?

rcritten:
re-running. It failed with the dbus teardown problem.

menonsudhir:
dbus related error during uninstall is being tested in https://github.com/freeipa/freeipa/pull/5573

flo-renaud:
@menonsudhir #5573 has been pushed, if you rebase to the master branch the issue shouldn't appear anymore.

flo-renaud:
master:

* e86ff48b49d271905528c2ce352bd8665df2b164 ipatests: Test to check ipa-healthcheck tool displays warning when run on ipa-client

PR #5650, opened 2021-03-19
ipatests: allocate pseudo-terminal only for specific command

wladich:
While "ktutil" does require a pseudo-terminal on particular systems to
    operate, majority of programs do not need it.
    At the same time invoking `ssh` with forced pseudo-terminal allocation
    interferes with sessions multiplexing feature and increases connection
    time. The increase can be as large as 10 seconds in certain cases which
    leads to unexpected EOFs of pexpect utility.

This PR is fixing `test_http_kdc_proxy` test suite on Fedora32

### DIFF
--- a/ipatests/pytest_ipa/integration/expect.py
+++ b/ipatests/pytest_ipa/integration/expect.py
@@ -68,6 +68,8 @@ class IpaTestExpect(pexpect.spawn):
         else:
             command = argv[0]
             args = argv[1:]
+        logger.debug('Expect will spawn command "%s" with args %s',
+                     command, args)
         super().__init__(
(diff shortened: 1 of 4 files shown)


wladich:
The temp commit executes all tests which utilize pexpect, both on Fedora 32 and Fedora 33

flo-renaud:
Hi @wladich 
thanks for the fix, LGTM. Please remove the temp commit and I will ack when PRCI completes.

abbra:
master:

* ae533e2998bcd53a465b9647cbe1ebbdc89efa82 ipatests: allocate pseudo-terminal only for specific command
* 947167d892aeb693a9bda058158e1bb2d5760c8b ipatests: log command spawned by pexpect

PR #5651, opened 2021-03-19
Avoid comparing 'max' with 'max\n'.

adelton:
Fixes https://pagure.io/freeipa/issue/8764.


rcritten:
We mock the file opens which is why this wasn't caught by tests. I'm not sure we can automation for this.

flo-renaud:
master:

* 4769c15f3759a5c00458a17ed3b3db140c5eebfe Avoid comparing 'max' with 'max\n'.

PR #5653, opened 2021-03-22
ipatests: test to renew certs on replica using ipa-cert-fix

mrizwan93:
This test checks if ipa-cert-fix renews the certs on replica
after cert renewal on master.
    
related: https://pagure.io/freeipa/issue/7885

### DIFF
--- a/ipatests/test_integration/test_ipa_cert_fix.py
+++ b/ipatests/test_integration/test_ipa_cert_fix.py
@@ -6,6 +6,7 @@
 Module provides tests for ipa-cert-fix CLI.
 """
 import pytest
+import re
 import time
 
 import logging
@@ -48,6 +49,16 @@ def check_status(host, cert_count, state, timeout=600):
     return count
 
 
+def move_date(host, chrony_state, date_str):
+    """Helper method to move the date on given host
+    :param host: The host on which date is to be moved
+    :param chrony_state: State to which chrony service to be moved
+    :param date_str: date string to move the date i.e 2years1month1days
+    """
+    host.run_command(['systemctl', chrony_state, 'chronyd'])
+    host.run_command(['date', '-s', date_str])
+
+
 @pytest.fixture
 def expire_cert_critical():
     """
@@ -64,15 +75,15 @@ def expire_cert_critical():
                              extra_args=['--no-ntp'])
         if setup_kra:
             tasks.install_kra(host)
-        host.run_command(['systemctl', 'stop', 'chronyd'])
-        host.run_command(['date', '-s', '+3Years+1day'])
+
+        # move date to expire certs
+        move_date(host, 'stop', '+3Years+1day')
 
     yield _expire_cert_critical
 
     host = hosts.pop('host')
     tasks.uninstall_master(host)
-    host.run_command(['date', '-s', '-3Years-1day'])
-    host.run_command(['systemctl', 'start', 'chronyd'])
+    move_date(host, 'start', '-3Years-1day')
 
 
 class TestIpaCertFix(IntegrationTest):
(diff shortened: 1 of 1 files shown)


rcritten:
I suppose it would be ideal if move_date was called in a fixture to ensure that the date is always returned to present time.

check_status() seems a bit risky and relies on undocumented knowledge of certificate state. This could blow up mightily if we add additional certs. I realize you didn't write the original method. What you have here looks ok but ideally it would check for all known certs for a given config and confirm their status, and perhaps blow up if there are unknown certs or a cert is not found (or in the wrong state). There are just currently limited helpers to obtain certmonger data from a remote host.

mrizwan93:
> I suppose it would be ideal if move_date was called in a fixture to ensure that the date is always returned to present time.
> 
> check_status() seems a bit risky and relies on undocumented knowledge of certificate state. This could blow up mightily if we add additional certs. I realize you didn't write the original method. What you have here looks ok but ideally it would check for all known certs for a given config and confirm their status, and perhaps blow up if there are unknown certs or a cert is not found (or in the wrong state). There are just currently limited helpers to obtain certmonger data from a remote host.

Rob, Thanks for the feedback.
Here is the scenario:
1. Move date on master, wait for certificate to expire
2. run `ipa-cert-fix` on master to fix the expired certs
3. Move date on replica, HTTPD/LDAP/PKINIT certs will renew automatically.

Here I have to wait for fixing the cert on master and then waiting for these certs to fix on replica. Hence I didn't write `move_date` as fixture(I don't know if I can call same fixture twice at different time and with different params).

Regarding `check_status()` we are providing it number of certs to check in given state. If found it returns else get timedout in case of mismatch of cert state/missing certs.

rcritten:
Good point on the move_date. If it ever fails then all subsquent tests in the class are likely to fail as well, but the system would also be probably hosed so it's about the same thing.

What I worry about with checking the state this way is it isn't asking "are these 5 certs in MONITORING?" it's asking "are any 5 certs in MONITORING?" I guess it works but seems brittle. We expect things to renew in a certain order, particularly since we control the date and ipa-cert-fix only works on so much, but if something else changes (e.g. we add a kra) this will start failing.

But like I said, I don't want to force you to rewrite it that code.

Test seems sane and are green. Please remove the temp commit.

Also, can you add the ticket to refactor commit so we know what drove the change?

mrizwan93:
Refactor is due to the fact that I introduced `move_date()` as method and used it in the `expire_cert_critical` fixture, hence no ticket.

ipatests/test_integration/test_ipa_cert_fix.py
```diff
@@ -260,3 +261,68 @@ def test_renew_expired_cert_with_kra(self, expire_cert_critical):
...
+
+        # wait for cert expiry
+        check_status(self.master, 8, "CA_UNREACHABLE")
+
+        self.master.run_command(['ipa-cert-fix', '-v'], stdin_text='yes\n')
+
+        check_status(self.master, 9, "MONITORING")
+
+        # move system date to expire cert on replica
+        move_date(self.replicas[0], 'stop', '+3years+1days')
```
rcritten, line 299: Are you keeping the replica at a different time so that replication of the updated certs doesn't work so that you can test the state of it in a more controlled way?

mrizwan93: I didn't get that. But I moved the date with exact time frame on master too.
So now, master have updated certs. Replica have same date as master and 3 certs(httpd/ldap/pkinit) got renewed(replicated from master) and remaining needs to be resubmit and RA agent will be fixed by `ipa-cert-fix`

rcritten:
This looks ok to me. Can you squash everything into one commit?

flo-renaud:
master:

* 99e7ad0fd8d7f621f1d3999c3fb7327802293ec6 ipatests: test to renew certs on replica using ipa-cert-fix

flo-renaud:
Hi @mrizwan93 
the automated backport to ipa-4-8 failed, can you do a manual one? thanks

mrizwan93:
ipa-4-8 doesn;t have test and fixture updated in this patch. Hence removing label and closing the PR for ipa-4-8

PR #5658, opened 2021-03-23
configure: Make rpmlint optional

stanislavlevin:
Distributions may want to run comprehensive fastcheck or lint tasks, but rpmlint tool is mandatory for these targets while some platforms don't have it at all.

With this change the rpmlint becomes optional for fastcheck, devcheck and lint make targets.

Note: rpmlint option is `disabled` by default.
To enable: `./configure --enable-rpmlint`
To explicitly disable: `./configure --disable-rpmlint`

Fixes: https://pagure.io/freeipa/issue/8768

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -559,6 +559,23 @@ AS_CASE([$JSLINT],
 AC_SUBST([JSLINT])
 AM_CONDITIONAL([WITH_JSLINT], [test "x${JSLINT}" != "xno"])
 
+
+AC_ARG_ENABLE(
+    [rpmlint],
+    [AC_HELP_STRING([--enable-rpmlint], [Enable rpmlint for the rpm spec])],
+    [case "${enableval}" in
+         yes) ENABLE_RPMLINT=yes ;;
+          no) ENABLE_RPMLINT=no ;;
+           *) AC_MSG_ERROR([bad value ${enableval} for --enable-rpmlint]) ;;
+     esac],
+    [ENABLE_RPMLINT=no])
+AS_IF([test "x$ENABLE_RPMLINT" = "xyes"],
+      [AC_PATH_PROG([RPMLINT], [rpmlint])
+       AS_IF([test "x$RPMLINT" = "x"],
+             [AC_MSG_ERROR([Could not find rpmlint])])],
+)
+AM_CONDITIONAL([WITH_RPMLINT], [test "x$ENABLE_RPMLINT" = "xyes"])
+
 AM_CONDITIONAL([HAVE_UNSHARE],
     [test "x${ac_cv_func_unshare}" = "xyes" -a "x${ac_cv_func_chroot}" = "xyes"])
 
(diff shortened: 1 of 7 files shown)


configure.ac
```diff
@@ -672,6 +689,7 @@ echo "
         Python:                   ${PYTHON} (${PYTHON_VERSION})
         pylint:                   ${PYLINT}
         jslint:                   ${JSLINT}
+        rpmlint:                  $enable_rpmlint
```
rcritten, line 692: nit: shouldn't this follow the same pattern as the others, e.g. upper-case ${ENABLE_RPMLINT} ?

rcritten: Other than the nit the patch LGTM.

stanislavlevin: PYLINT and JSLINT are shell variables in this case as well as enable_rpmlint. There is no common convention in IPA configure.ac regarding upper/low case(both exist). It doesn't matter for me. If you think the upper is more appropriate then I'll change it.

rcritten: That's true about inconsistency but most (I didn't check for all) of the optional/detected features use upper case. They universally use the ${<variable>} format.

stanislavlevin: Thank you, pushed the 'upper' version.

rcritten:
Maybe it's just that I have a tough time understanding how Azure actually works but the template script(s) aren't displayed in the output AFAICT so its difficult to verify that the right template is actually applied. For example the closest I can find to calling autogen.sh is:

2021-03-24T15:20:53.8776078Z [command]/usr/bin/bash --noprofile --norc /__w/_temp/0ed4b940-be80-48fa-be3a-705ef9e693a2.sh
2021-03-24T15:20:53.8776736Z Running autoconf generator
2021-03-24T15:20:54.3655448Z Copying file ABOUT-NLS

But what are the contents of /__w/_temp/0ed4b940-be80-48fa-be3a-705ef9e693a2.sh ?

stanislavlevin:
The content of `script` task is copied into xxx.sh and then executed.
```
  5 - script: |                                                                     
  6     set -e                                                                      
  7     echo "Running autoconf generator"                                           
  8     ./autogen.sh \                                                              
  9         ${{ parameters.options }} \                                             
 10         --enable-rpmlint \
```
The traditional way is `set -x`. Another option is system diagnostic option in Azure Web interface, but this requires manual run/rerun: https://dev.azure.com/freeipa/freeipa/_build/results?buildId=8324&view=results

abbra:
LGTM.

abbra:
master:

* 7fc0c04dc4f46acfd8bc0d73b3116cd07ce32d7a configure: Make rpmlint optional
* 83e16a4e47fcf488aa5f9aab6c0f0424990ab171 Azure: Run rpmlint on Fedora

abbra:
@stanislavlevin  could you please do `ipa-4-8` backport? Automation failed:
```
Applying to ipa-4-8: configure: Make rpmlint optional
Applying to ipa-4-8: Azure: Run rpmlint on Fedora
Failure to apply patches: error: Не удалось слить изменения.
подсказка: Use 'git am --show-current-patch=diff' to see the failed patch

```

stanislavlevin:
@abbra, thank you!


> @stanislavlevin could you please do `ipa-4-8` backport? Automation failed:
> 
> ```
> Applying to ipa-4-8: configure: Make rpmlint optional
> Applying to ipa-4-8: Azure: Run rpmlint on Fedora
> Failure to apply patches: error: Не удалось слить изменения.
> подсказка: Use 'git am --show-current-patch=diff' to see the failed patch
> 
> Failed to apply patches onto origin/ipa-4-8. Manual backport is needed.
> ```

Sure, https://github.com/freeipa/freeipa/pull/5677

PR #5669, opened 2021-03-26
Adapt redhat ipaplatform to RHEL9/ELN

flo-renaud:
On RHEL8, ipa is using named-pkcs11.service but RHEL9 is based on
fedora34 and uses named.service instead. There is already some support
for this distinction in ipaplatform, and the patch relies on the
specific settings that can be configured in ipaplatform/xx/services.py
and ipaplatform/xx/constants.py

On RHEL9 ipa also needs to define NAMED_OPENSSL_ENGINE for named
to use openssl's okcs11 engine.

Fixes: https://pagure.io/freeipa/issue/8753


rcritten:
Looks identical to the working code in Fedora, ack.

abbra:
master:

* c2c533b7655e3b8da966c84874a63fdcf5ff3fe6 Adapt redhat ipaplatform to RHEL9/ELN

PR #5680, opened 2021-03-29
ipatests: Skip test_jsplugins in server-less environments

stanislavlevin:
This test assumes that IPA server is installed and configured.
But test_jsplugins is the subtask of fasttest which is designed
(unittest) to be run in server-less environment.

Note: `needs_ipaapi` is not completely suitable because there is
no direct IPA API usage. `xmlrpc_setup` fixture is also not
suitable because it assumes XMLRPC.

Fixes: https://pagure.io/freeipa/issue/8781

### DIFF
--- a/ipatests/test_ipaserver/test_jsplugins.py
+++ b/ipatests/test_ipaserver/test_jsplugins.py
@@ -3,12 +3,17 @@
 import os
 import pytest
 
+from ipalib import facts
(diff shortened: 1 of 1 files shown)


abbra:
LGTM. ACK

abbra:
master:

* 68a5fe822358887e6d42fc500d5eb5408f90b148 ipatests: Skip test_jsplugins in server less environments

stanislavlevin:
@abbra, thank you for the review!

PR #5681, opened 2021-03-29
 Implement simple LDAP cache layer

rcritten:
Pretty straightforward caching except for the attribute
handling. We probably won't want to cache the entire entry
since that could be expensive (userCertificate, ssh keys, etc)
but not storing it could lead to multiple requests for the
same thing. Also, requesting * vs a set of attributes may not
return all a plugin needs, as it will skip operational attributes.
    
Still, this saves 20-50% of queries in my limited testing.

Mar 29: initial implementation works about 80% of the time. 
I guarantee the xmlrpc tests will fail. We'll see about the integration tests.
There are still corner cases to find and the member modlist stuff
is a mess.

Apr 7: ready for review

### DIFF
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -31,6 +31,8 @@ import os
 import pwd
 import warnings
 
+from collections import OrderedDict
+
 from cryptography import x509 as crypto_x509
 from cryptography.hazmat.primitives import serialization
 
@@ -47,7 +49,7 @@ from ipalib.constants import LDAP_GENERALIZED_TIME_FORMAT
 # pylint: enable=ipa-forbidden-import
 from ipaplatform.paths import paths
 from ipapython.ipautil import format_netloc, CIDict
-from ipapython.dn import DN
+from ipapython.dn import DN, RDN
 from ipapython.dnsutil import DNSName
 from ipapython.kerberos import Principal
 
@@ -1698,6 +1700,7 @@ class LDAPClient:
         modlist = entry.generate_modlist()
         if not modlist:
             raise errors.EmptyModlist()
+        logger.debug("update_entry modlist %s", modlist)
 
         # pass arguments to python-ldap
         with self.error_handler():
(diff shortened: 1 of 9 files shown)


abbra:
You probably need to fix 

```
./ipapython/ipaldap.py:1822:81: E501 line too long (94 > 80 characters)

```

before we would be able to see xmlrpc tests' results ;)

rcritten:
Knowing when to cache is pretty tricky. It may require some direct changes to the plugins, we'll see.

For example, with a DNS record if you set the value to '' then the whole record will be deleted. The problem is knowing that the value has been removed. The sequence is:

get_entry(dn)
update_entry(someattr=None)
get_entry(dn) in super( dnsrecord_mod) because we return the value
If none of our known attributes are in the entry then delete it

So in that get_entry() in LDAPUpdate a cached value is returned including the idnsname which cancels the delete.

One workaround would be to add an option to not return the updated value if we aren't going to use it. This in itself would save a query without relying on the cache. But doing so without modifying the API...

rcritten:
Ok so the XMLRPC tests are failing because I forgot about modes. IPA has strict checking to prevent developers from doing nasty things. This is only checked during development. In production mode that is disabled because we assume the developers already did the right thing.

So the tests are failing because it isn't in production mode and I'm tweaking values in a locked object:

[Wed Mar 31 19:28:38.288182 2021] [wsgi:error] [pid 6331:tid 6668] [remote 2001:db8:1:2::2:33046]     raise AttributeError(
[Wed Mar 31 19:28:38.288185 2021] [wsgi:error] [pid 6331:tid 6668] [remote 2001:db8:1:2::2:33046] AttributeError: locked: cannot set ldap2.misses to 1

Maybe this can live in context instead.

rcritten:
Hmm, so test_sudo and test_membermanager are failing in consistent ways in CI that I can't reproduce locally.

One of the failing xmlrpc tests is checking that a delete was actually performed but deletes aren't cached. I'll check into that as it may be a subtle difference.

rcritten:
I was able to reproduce the Kerberos flags test failure. It is due to kadmin.local modifying the service principal. This is cached so a service-show returns stale information.

This is in part because even though there are discrete ipa commands issued by the test bed they all seem to use the same connection.

rcritten:
re-run timed out test-cert

rcritten:
All tests are green. Adding patch to re-enable caching of exceptions to see what impact that has.

rcritten:
Ok, cache exceptions == bad during install. Lesson learned.

rcritten:
It works, time to analyze if it's worth it.

rcritten:
The xmlrpc tests, excluding vault (I didn't have a kra installed), result in:

Total cache hits 3154
Total cache misses 11074
Savings (hits / (hits + misses)) 0.221676

rcritten:
The biggest cache size is 35 but the average is 1.2 per connection.

rcritten:
The biggest cache impact is visible in these tests:

test_cert_plugin.py 0.100
test_hbactest_plugin.py 0.141
test_caacl_plugin.py 0.148
test_host_plugin.py 0.203
test_netgroup_plugin.py 0.174
est_realmdomains_plugin.py 0.220
test_sudorule_plugin.py 0.228
test_dns_plugin.py 0.229
test_dns_realmdomains_integration.py 0.260
ttest_cert_request_ip_address.py 0.314
test_config_plugin.py 0.625
test_kerberos_principal_aliases.py 0.744
test_location_plugin.py 0.746

Most others are in the 2-6% range.

Whether reflect actual usage or not is another discussion.

A bunch of plugins had no caching at all, including sudocmd and hostgroup. These may be simple enough that they don't re-read entries ever.

TODO: limit the types of exceptions that are cached.

Assuming though that all exceptions are NotFound then hits and misses each represent an LDAP query then this represents a pretty substantial load saving for the LDAP server.

rcritten:
While looking to limit the exceptions cached I discovered that while cached, exceptions weren't counted as a hit. The numbers improve quite a bit:

Total cache hits 4693
Total cache misses 11125
Savings 0.296687

rjeffman:
Tested some DNS plugins (dnszone-add, dnszone-del, dnsrecord-add, dnsrecord-del) against the cache layer.

When running against an unpatched Fedora 33, FreeIPA 4.9.3, the number of queries executed are:
```COMMAND                                  QUERIES
dnszone-add example.test                    23
dnsrecord-add 192.168.122.100               10
dnsrecord-add 192.168.122.101               12
dnsrecord-add 192.168.122.102               12
dnsrecord-add 192.168.122.110               10
dnsrecord-del 192.168.122.100               11
dnsrecord-del --del-all                      9
dnsrecord-del --del-all                      9
dnszone-del example.test                    22
TOTAL: 118
```

When executed the same commands against this branch:
```
COMMAND                                  QUERIES
dnszone-add example.test                    17
dnsrecord-add 192.168.122.100                7
dnsrecord-add 192.168.122.101                9
dnsrecord-add 192.168.122.102                9
dnsrecord-add 192.168.122.110                7
dnsrecord-del 192.168.122.100                7
dnsrecord-del --del-all                      6
dnsrecord-del --del-all                      6
dnszone-del example.test                    18
TOTAL: 86                                          
```

The results in ~27% less queries against the LDAP database, which I find very nice.

rjeffman:
Also run the changes against the ansible-freeipa tests, and all passed.

Although I'd prefer that the backend class (LDAPClient or LDAPCache) were configured (by assignment) rather than inherited, I agree that the way it is done requires less change to the source.

This change looks good to me.

rcritten:
Added a unit test for the cache to confirm that hits and misses work and the cache is cleared on any mod/add/del.

ipapython/ipaldap.py
```diff
@@ -1749,3 +1752,239 @@ def get_ldap_uri(host='', port=389, cacert=None, ldapi=False, realm=None,
...
+                 sasl_nocanon=True, enable_cache=True, cache_size=100):
+
+        self.cache = OrderedDict()
+        self._cache_enabled = enable_cache
+        self._cache_size = cache_size
+        object.__setattr__(self, '_cache_misses', 0)
+        object.__setattr__(self, '_cache_hits', 0)
+        object.__setattr__(self, '_enable_cache', enable_cache)
+
+        object.__setattr__(self, '_cache_size', cache_size)
```
rjeffman, line 1780: Can only one of setting instance attributes be used here? I prefer `self.attr`, as I din't find `__setattr__` to be overriden.
Also, you are creating both `_cache_enabled` and `_enabled_cache`. Are both needed? I believe only one is required.

ipapython/ipaldap.py
```diff
@@ -1749,3 +1752,239 @@ def get_ldap_uri(host='', port=389, cacert=None, ldapi=False, realm=None,
...
+
+class LDAPCache(LDAPClient):
+    """A very basic LRU Cache using an OrderedDict"""
+
+    def __init__(self, ldap_uri, start_tls=False, force_schema_updates=False,
+                 no_schema=False, decode_attrs=True, cacert=None,
+                 sasl_nocanon=True, enable_cache=True, cache_size=100):
+
+        self.cache = OrderedDict()
+        self._cache_enabled = enable_cache
```
rjeffman, line 1774: What do you think about setting `self._cache_enabled`to `enable_cache and cache_size > 0`, as setting `cache_size` to `0` effectively disables cache?

rcritten: Good idea, done.

ipapython/ipaldap.py
rcritten: I renamed the value at some point and didn't thoroughly move over to it, fixed.

rjeffman:
LGTM.

ipapython/ipaldap.py
```diff
@@ -1749,3 +1752,239 @@ def get_ldap_uri(host='', port=389, cacert=None, ldapi=False, realm=None,
...
+            object.__setattr__(self, '_cache_hits', hits)
+            self.emit("HIT: Re-raising %s", entry.exception)
+            self.cache_status('HIT')
+            raise entry.exception
+
+        self.emit("Requested attrs_list %s", attrs_list)
+        if entry:
+            self.emit("Cached attrs_list %s", entry.attrs_list)
+
+        if not attrs_list:
```
rcritten, line 1939: We have a number of places where attrs_list = [''] which should be equivalent to []/['*']. I'm experimenting with this. I think we can save additional searches.

rcritten: So get_dn() calls generally just do that, get the dn of an entry. This is still cacheable but `if entry `returns False so these were being ignored. Explicitly set the attrs_list to ['dn'] and change the method to see if there is a cache entry.

Also add a log parser I wrote to help calculate the cache hits, misses, etc.

rjeffman:
LGTM.

rcritten:
master:

* d1f3ff5506b5ff9bcdc4aa75174cc32f6e4269c6 Call the LDAPClient layer when modifying values
* 8365d5e734be74b108465abf847160f05aadcb6a Unify installer context to be 'installer'
* a4675f6f503b153d5adfcfae8c5c1a7a67bdb692 Implement simple LDAP cache layer
* 3539857ecb156351f3385ddbcab029f22ee4afb4 Add LDAP cache options to the default.conf man page
* 8d21df93522c7ef76ae03e6b5be303da68420f69 Add a unit test for the LDAP cache layer
* 1e9a238f358268f23cf359e57038014026fa50b0 Parse the debugging cache log to determine the read savings

PR #5689, opened 2021-03-31
hbacrule: reduce number of LDAP searches during deletion

antoniotorresm:
The `hbacrule` module performs a call to `selinuxusermap-find`
during entry deletion. This can be optimized by passing pkey_only=True
to the search, skipping the post-callback function. Passing the full
DN of the hbacrule and detecting it in the selinuxusermap find
also saves one call to hbacrule-show, further reducing the searches.

Related: https://pagure.io/freeipa/issue/8784

### DIFF
--- a/ipaserver/plugins/hbacrule.py
+++ b/ipaserver/plugins/hbacrule.py
@@ -317,7 +317,7 @@ class hbacrule_del(LDAPDelete):
 
     def pre_callback(self, ldap, dn, *keys, **options):
         assert isinstance(dn, DN)
-        kw = dict(seealso=keys[0])
+        kw = dict(seealso=str(dn), pkey_only=True)
         _entries = api.Command.selinuxusermap_find(None, **kw)
         if _entries['count']:
             raise errors.DependentEntry(key=keys[0], label=self.api.Object['selinuxusermap'].label_singular, dependent=_entries['result'][0]['cn'][0])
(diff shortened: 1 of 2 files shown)


rcritten:
The problem is moving logic out of one plugin and into another. This could lead to unexpected problems if the source plugin is updated.

Neither this PR not the associated ticket outline what this optimizes.

antoniotorresm:
This is just an existence check. Checking this way instead of calling `selinuxusermap-find` saves one LDAP operation for each HBAC rule deletion. I don't think it is particularly problematic since it's just a query, it isn't modifying anything in an unexpected way. In fact, the `selinuxusermap` plugin does also [retrieve HBAC rules entries directly from LDAP](https://github.com/freeipa/freeipa/blob/7eb50d4e6eaee12e9950de611d71f428567798a4/ipaserver/plugins/selinuxusermap.py#L321).

rcritten:
I think adding pkey_only=True to the find will have the same effect.

The selinuxusermap plugin encapsulates its own logic so knowing that seealso is the dn of the HBAC rule is ok. Using this logic outside the plugin is where we can get into trouble IMHO.

antoniotorresm:
Updated the PR after discussion offline. Now no logic is extracted from the selinuxusermap plugin, and it is able to detect that a full HBAC DN has been passed to the find, saving one call to hbacrule-show. Also, passing `pkey_only=True` to the find saves an additional search.

rcritten:
Looks good. I particularly like catching the right exception (ValueError) when converting to DN rather than Exception.

rcritten:
Do we want to backport this to ipa-4-9 or have it in master branch only?

t-woerner:
I think that a backport to ipa-4-9 would be good.

rcritten:
master:

* 1a539984c719c7142d356dfc71d9a81dd566013c hbacrule: reduce number of LDAP searches during deletion

PR #5690, opened 2021-03-31
Add pkey_only to the service_find calls in the host plugin

rcritten:
This limits the amount of data and post-processing done
in when host_del and host_disable call service_find.
This also saves a presence query for host_keytab.

https://pagure.io/freeipa/issue/8787

### DIFF
--- a/ipaserver/plugins/host.py
+++ b/ipaserver/plugins/host.py
@@ -808,7 +808,7 @@ class host_del(LDAPDelete):
         truncated = True
         while truncated:
             try:
-                ret = api.Command['service_find'](fqdn)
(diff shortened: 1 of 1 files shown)


abbra:
LGTM but it only works because `krbprincipalname` is the primary key here. May be you could extend the commit message explaining that?

rcritten:
Excellent point. Commit message updated.

flo-renaud:
Hi @rcritten 
thanks for the fix, LGTM and @abbra 's comment has been taken into account. ACK

rcritten:
master:

* a3b606c08f621e7e65cac5aea59aaecd5c791101 Add pkey_only to the service_find calls in host del and disable

PR #5691, opened 2021-03-31
Also use uglifyjs on CentOS Stream 8

carlwgeorge:
This conditional was recently changed to match VERSION_ID "8." to only apply to RHEL 8 releases, but it should also match CentOS Stream 8 which has VERSION_ID "8".

https://pagure.io/freeipa/c/43f344b931db3f72f50e1620443be9f21623e29a

Please also cherry-pick this to the ipa-4-9 branch to be included in future RHEL 8 releases.

### DIFF
--- a/install/ui/util/compile.sh
+++ b/install/ui/util/compile.sh
@@ -112,7 +112,7 @@ fi
 echo "Minimizing: $RDIR/$RELEASE/$LAYER.js"
 echo "Target file: $OUTPUT_FILE"
 if [[ ("$ID" == "rhel" || "$ID_LIKE" =~ "rhel")
-      && "$VERSION_ID" =~ "8." ]];
+      && ("$VERSION_ID" =~ "8." || "$VERSION_ID" == "8") ]];
 then
     echo "Minifier: uglifyjs"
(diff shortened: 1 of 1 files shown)


abbra:
Right. This is what happens when there is no regular way to test builds on CentOS 8 Stream. ;) A corresponding commit 5fb0cc43eab329e8cb0020ca96f70a05fa9bb4bd is in FreeIPA 4.9.2 release already.

I added ipa-4-9 label and triggered full CI run.

rcritten:
master:

* 2c83b1e7135640dd55572f1bbe9407c57c459c1e Also use uglifyjs on CentOS Stream 8

carlwgeorge:
> Right. This is what happens when there is no regular way to test builds on CentOS 8 Stream. ;)

This is what happens when the Stream concept is done retroactively on 8, instead as part of the initial bring up process like we're doing on 9.  But that's neither of our faults.  😉

abbra:
I am mostly worried about inability to rebuild FreeIPA on publicly available CentOS images due to missing build requirements. Hopefully, we can solve this with 9.

PR #5698, opened 2021-04-03
ipatests: call server-del before replica uninstall

flo-renaud:
The test test_replica_promotion.py::TestRenewalMaster::
test_automatic_renewal_master_transfer_ondelete is calling
ipa-server-install --uninstall directly without performing first
ipa server-del. This can lead to incomplete uninstallation and
test failures.

Fixes: https://pagure.io/freeipa/issue/8792

### DIFF
--- a/ipatests/test_integration/test_replica_promotion.py
+++ b/ipatests/test_integration/test_replica_promotion.py
@@ -402,7 +402,7 @@ class TestRenewalMaster(IntegrationTest):
     def test_automatic_renewal_master_transfer_ondelete(self):
         # Test that after replica uninstallation, master overtakes the cert
         # renewal master role from replica (which was previously set there)
-        tasks.uninstall_master(self.replicas[0])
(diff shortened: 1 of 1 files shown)


rcritten:
Strictly speaking this method uses ipa-replica-manage to do the delete, which does call ipa server-del. Should the commit message be updated or should we use ipa server-del directly instead of ipa-replica-manage?

flo-renaud:
@rcritten I updated the commit message. `ipa-replica-manage del` is working with DL0 and DL1, that's probably the reason it was used in the tasks module.

rcritten:
Thanks, ack!

rcritten:
Well, ack once you remove the temp commit please :-)

flo-renaud:
@rcritten thanks for the review, temp commit removed

rcritten:
master:

* 56639861a96754ad9f0ddec6d240cf62d750cf41 ipatests: call server-del before replica uninstall

PR #5701, opened 2021-04-06
Only request VALID certs when revoking certs for a host/service

rcritten:
Only request VALID certs when revoking certs for a host/service
    
This utilizes the new status option so that we only retrieve
VALID certificates when revoking certificates issued for a
specific host or service.
    
ae74d348c3da580264c56441c136af3fc6ae58df made a special case
in cert_find when searching for hosts and services so that if only
one host/service was searched on do a subject search. It only
works when there is exactly one option requested.
    
Since we want to restrict to only VALID certificates, pass
the hostname as subject directly when revoking certs when deleting
an entry to limit the number of certificates to consider.
    
### DIFF
--- a/ipaserver/plugins/host.py
+++ b/ipaserver/plugins/host.py
@@ -871,7 +871,9 @@ class host_del(LDAPDelete):
                 )
 
         if self.api.Command.ca_is_enabled()['result']:
-            certs = self.api.Command.cert_find(host=keys)['result']
+            certs = self.api.Command.cert_find(
+                subject=fqdn, status='VALID'
+            )['result']
             revoke_certs(certs)
 
         return dn
--- a/API.txt
+++ b/API.txt
@@ -754,7 +754,7 @@ output: Entry('result')
 output: Output('summary', type=[<type 'unicode'>, <type 'NoneType'>])
 output: PrimaryKey('value')
 command: cert_find/1
-args: 1,29,4
+args: 1,30,4
 arg: Str('criteria?')
 option: Flag('all', autofill=True, cli_name='all', default=False)
 option: Str('cacn?', cli_name='ca')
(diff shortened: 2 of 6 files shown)


frasertweedale:
@rcritten I haven't tested this (and about to go on PTO until mid-April).  Seems like the right approach though.

Just wanted to mention another ticket: https://pagure.io/freeipa/issue/7549.  It looks like this PR also resolves that ticket.

rcritten:
@ftweedal thanks. I had looked at 7549 and it requests that multiple statuses be supported at the same time which AFAICT isn't yet supported by dogtag. I was surprised we didn't support at least one status already to be honest.

frasertweedale:
@rcritten thanks, you're right.  I didn't look closely enough.

ipaserver/plugins/host.py
```diff
@@ -871,7 +871,9 @@ def pre_callback(self, ldap, dn, *keys, **options):
                 )
 
         if self.api.Command.ca_is_enabled()['result']:
-            certs = self.api.Command.cert_find(host=keys)['result']
+            certs = self.api.Command.cert_find(
+                subject=keys[-1], status='VALID'
```
flo-renaud, line 875: If the user called `ipa host-del host1` but the fqdn is `host1.ipa.test`, `keys[-1]` is `host1` and the filter `subject=host1` may match other hosts. I agree that host-del will also remove them but I think it's safer to use a filter based on fqdn that is evaluated a few lines above: cert_find(`subject=fqdn`,...).

flo-renaud (commented):
Hi @rcritten 
thanks for the PR. Please find one inline comment.

ipaserver/plugins/host.py
rcritten: I'll take a look, thanks.

rcritten: Nice catch, PR updated and rebased.

flo-renaud:
@rcritten the PR needs to be rebased because of my [commit](https://pagure.io/freeipa/c/ba7ab24b2962a9526d66ae4e0a5d7053b2577c86?branch=master) for idranges. I hope it will also unblock prci.
Otherwise LGTM.

flo-renaud:
@rcritten 
thanks for the update, tested with host-del and service-del with revoked and valid certs, works well.

flo-renaud:
master:

* 09426f8ed5ebc27d01ecbfd808fabdc0c4430854 Add ability to search on certificate revocation status
* aa1350384ad6a7d6b2f6056d99fbb43c5c5a6be7 Only request VALID certs when revoking certs for a host/service

PR #5702, opened 2021-04-06
Cache the value of ca_is_enabled in the request context

rcritten:
This may be called multiple times in a request and should
be impossible to change.

https://pagure.io/freeipa/issue/8797


flo-renaud:

rcritten:
master:

* 9cb65b4eebb9ec628b2795700a797a0c78e7ad7d Cache the value of ca_is_enabled in the request context

PR #5707, opened 2021-04-09
Retrieve the user objectclasses when checking for existence

rcritten:
This saves at least one search per user-mod because the current
set of objectclasses are verified to ensure they are complete
on each update.

So always retrieve them in get_either_dn(). They are used by
every call but there is negligible overhead in retrieving
this from LDAP.

https://pagure.io/freeipa/issue/8801

### DIFF
--- a/ipaserver/plugins/user.py
+++ b/ipaserver/plugins/user.py
@@ -443,15 +443,16 @@ class user(baseuser):
 
     def get_either_dn(self, *keys, **options):
         '''
-        Returns the DN of a user
+        Returns the DN of a user and their objectclasses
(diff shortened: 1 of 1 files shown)


rcritten:
In the xmlrpc test_user_plugin.py this reduces the number of searches by 27.

flo-renaud:
@rcritten 
thanks for the PR, LGTM. Do you need a backport to ipa-4-9?

rcritten:
Yes, adding the flag.

flo-renaud:
master:

* bf1f66c0c15ceeb65671df495cf66018522ae3cc Retrieve the user objectclasses when checking for existence

PR #5708, opened 2021-04-10
ipaserver/install/dns: handle SERVFAIL when checking reverse zone

abbra:
systemd-resolved in Fedora 34+ returns SERVFAIL for reverse zone that
does not yet exist when we attempt to look it up before installation.
Assume that this is OK -- we are going to create the zone ourselves
during installation.

Fixes: https://pagure.io/freeipa/issue/8794

### DIFF
--- a/ipapython/dnsutil.py
+++ b/ipapython/dnsutil.py
@@ -447,7 +447,13 @@ def check_zone_overlap(zone, raise_on_error=True):
     except dns.exception.DNSException as e:
         msg = ("DNS check for domain %s failed: %s." % (zone, e))
         if raise_on_error:
-            raise ValueError(msg)
+            if isinstance(e, dns.resolver.NoNameservers):
+                # Show warning and continue in case we've got SERVFAIL
+                # because we are supposedly going to create this reverse zone
+                logger.warning('%s', msg)
+                return
+            else:
+                raise ValueError(msg)
         else:
             logger.warning('%s', msg)
             return


abbra:
This is Fedora 34 release blocker: https://bugzilla.redhat.com/show_bug.cgi?id=1948034

ipapython/dnsutil.py
```diff
@@ -447,6 +451,8 @@ def check_zone_overlap(zone, raise_on_error=True):
     except dns.exception.DNSException as e:
         msg = ("DNS check for domain %s failed: %s." % (zone, e))
         if raise_on_error:
+            if isinstance(e, dns.resolver.NoNameservers):
+                raise DNSNoNameservers(**e.args, **e.kwargs)
```
tiran, line 455: ``**args`` looks wrong. You can use ``from None`` to suppress exception chaining (another error occurred while handling ...).

```suggestion
                raise DNSNoNameservers(*e.args, **e.kwargs) from None
```

abbra: Thanks, fixed. This was really an idea to try but I am still to reproduce the original issue myself.

abbra:
I've reproduced it with Fedora Rawhide and got fixed with the updated patch.

flo-renaud (commented):
Hi @abbra 
Thanks for the patch. I have one in-line comment but otherwise LGTM.

ipaserver/install/bindinstance.py
```diff
@@ -312,6 +312,12 @@ def read_reverse_zone(default, ip_address, allow_zone_overlap=False):
                 logger.error("Reverse zone %s will not be used: %s",
                              zone, e)
                 continue
+            except dnsutil.DNSNoNameservers as e:
```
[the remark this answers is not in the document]
abbra: Yep. I'll remove it from here.

abbra:
@AdamWill created a test OpenQA pipeline with my scratch build for F34, it is succeeded in staging environment:
https://openqa.stg.fedoraproject.org/tests/overview.html?distri=fedora&version=34&build=Kojitask-65800229-NOREPORT&groupid=2

flo-renaud (commented):
Hi @abbra 
the new Exception `DNSNoNameservers` is always caught (simply logs a warning), maybe we can avoid changing the calling code everywhere and simply modify `check_zone_overlap` to log the warning and pass.

abbra:
Yes, simplified the logic now. I already pushed an update to Fedora with the previous version, we will update that with the next IPA release.

flo-renaud:
master:

* 48ef179e23925737499fc62ac015bf0844763217 ipaserver/install/dns: handle SERVFAIL when checking reverse zone

PR #5709, opened 2021-04-13
Add max/min safe integer

tiran:
JSON cannot safely handle integers outside range ``-(2**53) - 1`` to
``(2**53) - 1``. Add constants for safe integers and limit the Int
parameter to safe JSON values.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

See: https://pagure.io/freeipa/issue/8802
See: https://pagure.io/freeipa/issue/8361

### DIFF
--- a/ipalib/parameters.py
+++ b/ipalib/parameters.py
@@ -130,6 +130,13 @@ from ipapython.dn import DN
 from ipapython.dnsutil import DNSName
 
 
+MAX_UINT32 = (1 << 32) - 1
+# JavaScript Number.MAX_SAFE_INTEGER / Number.MIN_SAFE_INTEGER
+# JSON cannot safely encode values outside this range as regular number
+MAX_SAFE_INTEGER = (2**53) - 1
+MIN_SAFE_INTEGER = -MAX_SAFE_INTEGER
+
+
 def _is_null(value):
     if value:
         return False
@@ -1093,6 +1100,12 @@ class Int(Number):
     allowed_types = (int,)
     type_error = _('must be an integer')
 
+    MININT = MININT
+    MAXINT = MAXINT
+    MAX_UINT32 = MAX_UINT32
+    MAX_SAFE_INTEGER = MAX_SAFE_INTEGER
+    MIN_SAFE_INTEGER = MIN_SAFE_INTEGER
+
     kwargs = Param.kwargs + (
         ('minvalue', int, int(MININT)),
         ('maxvalue', int, int(MAXINT)),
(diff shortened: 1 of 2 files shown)


rcritten:
JSON in python doesn't seem to have a 2**53 limitation, at least in my ham-handed testing:
```
>>> import json
>>> x = 2**99
>>> d = dict(x=x)
>>> s = json.dumps(d)
>>> json.loads(s)
>>> s
'{"x": 172438684545000970159254868079478095774}'
```

All works fine. Or maybe I'm being unsafe? Or python JSON is more flexible than other libs?

tiran:
Python's JSON library can handle arbitrary length numbers fine. In general JSON and JS cannot. JS uses a mixed int/float type. Numbers >2**53 will have rounding errors.

abbra:
So this would be a reasonable compromise now. For a better solution we should rewrite Web UI to expect the serial number as a string, not an integer and do a similar change to other IPA API consumers. This can be backported relatively easily to both server and client sides.

rcritten:
I just wonder if this is a python json bug since it's out of spec.

I found https://developers.google.com/discovery/v1/type-format which clarifies the 2**53 limit better than the Mozilla page IMHO.

frasertweedale:
@rcritten JSON spec itself does not specify a number/int range: https://tools.ietf.org/html/rfc8259#section-6 . 2**53 limit is suggested for wide interoperability, in particular with JavaScript.  So the Python JSON implementation is compliant (with regard to numbers), but big numbers may cause interop issues with webUI or non-Python clients.

ipalib/parameters.py
```diff
@@ -130,6 +130,13 @@
 from ipapython.dnsutil import DNSName
 
 
+MAX_UINT32 = (1 << 32) - 1
+# JavaScript Number.MAX_SAFE_INTEGER / Number.MIN_SAFE_INTEGER
+# JSON cannot safely encode values outside this range as regular number
```
frasertweedale, line 135: JSON itself does not specify range or precision of the number type.  So it is really JavaScript's implementation (and others that depend on IEEE float implementations) that cannot safely handle such values.

rcritten: Do you want him to clarify the comment?

frasertweedale: That would be nice.  I don't consider it blocker though.

rcritten: Ok I'm going to ack it since this constrains IPA/python to the lowest-common denominator.

rcritten:
master:

* a297ebbb8a277c7c3fcb44f6da182dde71447442 Add max/min safe integer

PR #5710, opened 2021-04-13
client install: do not capture sudo -V stdout

flo-renaud:
### client install: do not capture sudo -V stdout

ipa-client-install is checking if the sudo command is available
by calling 'sudo -V'. The call is currently using subprocess.popen
which redirects the output to the default stdout.
Use ipautil.run instead of subprocess.popen as this does not
capture stdout (the command output is just logged in the debug file).

Fixes: https://pagure.io/freeipa/issue/8767

### ipatests: check that the output of sudo -V is not displayed

During client installation, the installer calls sudo -V
to check if sudo command is installed. The output must not
be displayed in stdout.

Related: https://pagure.io/freeipa/issue/8767

### DIFF
--- a/ipaclient/install/client.py
+++ b/ipaclient/install/client.py
@@ -24,7 +24,6 @@ import re
 import SSSDConfig
 import shutil
 import socket
-import subprocess
 import sys
 import tempfile
 import textwrap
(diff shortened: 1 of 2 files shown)


rcritten:
Works great, please remove the temp commit.

flo-renaud:
@rcritten tahnks for the review. Temp commit removed and tests are green

flo-renaud:
master:

* ec8d72cf7cdf960cb270570dde459dfde3d94fd9 client install: do not capture sudo -V stdout
* 058a9c01c73907ed1d81aeaa0298a4fde269e9cc ipatests: check that the output of sudo -V is not displayed

PR #5712, opened 2021-04-14
Add new auto-private-groups option to trust id ranges

flo-renaud:
The whole description is available in the design doc in doc/designs/adtrust/auto_private_groups.md

In short: add a new option --auto-private-groups=['true','false','hybrid'] to the idrange-add/idrange-mod/idrange-find commands that allows the configuration of auto private groups in a similar way to sssd.conf option.

Fixes: https://pagure.io/freeipa/issue/8807

### DIFF
--- a/install/share/60basev3.ldif
+++ b/install/share/60basev3.ldif
@@ -62,6 +62,7 @@ attributeTypes: (2.16.840.1.113730.3.8.18.2.1 NAME 'ipaVaultType' DESC 'IPA vaul
 attributeTypes: (2.16.840.1.113730.3.8.18.2.2 NAME 'ipaVaultSalt' DESC 'IPA vault salt' EQUALITY octetStringMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.40 X-ORIGIN 'IPA v4.2' )
 # FIXME: https://bugzilla.redhat.com/show_bug.cgi?id=1267782
 attributeTypes: (2.16.840.1.113730.3.8.18.2.3 NAME 'ipaVaultPublicKey' DESC 'IPA vault public key' EQUALITY octetStringMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.40 X-ORIGIN 'IPA v4.2' )
+attributeTypes: (2.16.840.1.113730.3.8.23.6 NAME 'ipaAutoPrivateGroups' DESC 'Auto private groups' EQUALITY caseIgnoreIA5Match SUBSTR caseIgnoreIA5SubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.26 X-ORIGIN 'IPA v4.9' )
 objectClasses: (2.16.840.1.113730.3.8.12.1 NAME 'ipaExternalGroup' SUP top STRUCTURAL MUST ( cn ) MAY ( ipaExternalMember $ memberOf $ description $ owner) X-ORIGIN 'IPA v3' )
 objectClasses: (2.16.840.1.113730.3.8.12.2 NAME 'ipaNTUserAttrs' SUP top AUXILIARY MUST ( ipaNTSecurityIdentifier ) MAY ( ipaNTHash $ ipaNTLogonScript $ ipaNTProfilePath $ ipaNTHomeDirectory $ ipaNTHomeDirectoryDrive ) X-ORIGIN 'IPA v3' )
 objectClasses: (2.16.840.1.113730.3.8.12.3 NAME 'ipaNTGroupAttrs' SUP top AUXILIARY MUST ( ipaNTSecurityIdentifier ) X-ORIGIN 'IPA v3' )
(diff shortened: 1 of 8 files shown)


flo-renaud:
Note to self: need to register the new OID

install/share/60basev3.ldif
```diff
@@ -62,6 +62,7 @@ attributeTypes: (2.16.840.1.113730.3.8.18.2.1 NAME 'ipaVaultType' DESC 'IPA vaul
 attributeTypes: (2.16.840.1.113730.3.8.18.2.2 NAME 'ipaVaultSalt' DESC 'IPA vault salt' EQUALITY octetStringMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.40 X-ORIGIN 'IPA v4.2' )
 # FIXME: https://bugzilla.redhat.com/show_bug.cgi?id=1267782
 attributeTypes: (2.16.840.1.113730.3.8.18.2.3 NAME 'ipaVaultPublicKey' DESC 'IPA vault public key' EQUALITY octetStringMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.40 X-ORIGIN 'IPA v4.2' )
+attributeTypes: (2.16.840.1.113730.3.8.23.12 NAME 'ipaAutoPrivateGroups' DESC 'Auto private groups' EQUALITY caseIgnoreIA5Match SUBSTR caseIgnoreIA5SubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.26 X-ORIGIN 'IPA v4.9' )
```
rcritten, line 65: Where did this OID come from? The last one in the file is 2.16.840.1.113730.3.8.23.5. Is this taking into account the work Christian is doing on subuid range? If so we need to get this added.

ipaserver/plugins/idrange.py
```diff
@@ -482,6 +496,15 @@ def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys, **options):
                             'or ipa-ad-trust-posix when SID of the trusted '
                             'domain is not specified.'))
 
+            # auto private group is possible only for ad trusts
+            if entry_attrs['iparangetype'] == 'ipa-local':
```
rcritten, line 500: For future proofing rather than looking for the odd-one out would this be better:
if entry_attrs['iparangetype'] not in ('ipa-ad-trust', 'ipa-ad-trust-posix'):

Is it safe to pull this directly and not use entry_attrs.get('iparangetype')?

rcritten:
You also need to update IPA_API_VERSION_MINOR in VERSION4.m4

install/share/60basev3.ldif
flo-renaud: Yes, I didn't want to step on his toes.

ipaserver/plugins/idrange.py
flo-renaud: > For future proofing rather than looking for the odd-one out would this be better:
> if entry_attrs['iparangetype'] not in ('ipa-ad-trust', 'ipa-ad-trust-posix'):

Good suggestion, updated.

> Is it safe to pull this directly and not use entry_attrs.get('iparangetype')?
Yes it's safe thanks to these lines just a few lines above:
```
            # Default to ipa-local if no type set
            if 'iparangetype' not in entry_attrs:
                entry_attrs['iparangetype'] = 'ipa-local'
```

flo-renaud:
> You also need to update IPA_API_VERSION_MINOR in VERSION4.m4

Thanks for the reminder, I always forget that part.

flo-renaud:
I synchronized offline with @tiran and he let me have the OID 2.16.840.1.113730.3.8.23.6 for the new attribute. Updated the design doc + schema with that OID.

ipaserver/plugins/idrange.py
```diff
@@ -80,6 +80,9 @@
  - rid-base: the first RID of the corresponding RID range
  - sid: domain SID of the trusted domain
 
+and an ID range of a trusted domain may set
+ - auto-private-groups: [true|false|hybrid] autom creation of private groups
```
rcritten, line 84: is autom cut off? Should this be automatically?

rcritten:
One minor typo I think. I think it's safe to fix this and remove the temp commits at the same time.

The patch works well for me.

flo-renaud:
Typo fixed and the OID has been registered internally.

abbra:
LGTM but let @rcritten to give the final ack.

rcritten:
ack

flo-renaud:
master:

* a0509eb9d06ab2d3a608af0b67d7d1616ffdecc2 Design doc for idrange option "auto-private-groups"
* 3844f5b3c83b460dd4684dd874ae38743bb7f565 LDAP schema: new attribute ipaautoprivategroups
* ba7ab24b2962a9526d66ae4e0a5d7053b2577c86 Trust: add auto private groups option
* f45a79a78331aaa125d6eb8317d35452c9b95d80 xmlrpc tests: add test for idrange auto-private-groups option

PR #5726, opened 2021-04-22
Use PyCA crypto provider for KRAClient

tiran:
The Dogtag KRA backend now uses CryptographyCryptoProvider instead of
NSSCryptoProvider for KRAClient connections. The
CryptographyCryptoProvider uses PyCA cryptography to provide wrapping
and unwrapping. The change will allow Dogtag to remove the
NSSCryptoProvider and drop python-nss as a dependency.

The code in ipaserver.plugins.dogtag creates a Certificate object to
work around a bug in Dogtag. Dogtag supports paths but passes the wrong
type to PyCA cryptography.

Fixes: https://pagure.io/freeipa/issue/8814
See: https://github.com/dogtagpki/pki/issues/3499

### DIFF
--- a/ipaserver/plugins/dogtag.py
+++ b/ipaserver/plugins/dogtag.py
@@ -253,7 +253,7 @@ import six
 from ipalib import Backend, api, x509
 from ipapython.dn import DN
 import ipapython.cookie
-from ipapython import dogtag, ipautil, certdb
+from ipapython import dogtag, ipautil
 from ipaserver.masters import find_providing_server
 
 import pki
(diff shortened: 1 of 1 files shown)


rcritten:
LGTM, please remove temp commit

tiran:
PR-CI didn't trigger

edewata:
I'm fixing dogtagpki/pki#3499 in https://github.com/dogtagpki/pki/pull/3500, but I think it's better to provide a certificate object like in this PR rather than relying on `CryptographyCryptoProvider` to load the file anyway.

flo-renaud:
master:

* 1d80048e056ff0e1aabccaacb453c9d8d63ae5b9 Use PyCA crypto provider for KRAClient

PR #5728, opened 2021-04-22
ipa-otpd: handle LDAP timeout in a better way

abbra:
When LDAP server disconnects ipa-otpd client connection due to a
timeout, ipa-otpd instance would stop and report an issue. This confuses
systemd service reporting, so for these situations we better to shut
down gracefully.

Fixes: https://pagure.io/freeipa/issue/6587

### DIFF
--- a/daemons/ipa-otpd/bind.c
+++ b/daemons/ipa-otpd/bind.c
@@ -85,9 +85,11 @@ static void on_bind_readable(verto_ctx *vctx, verto_ev *ev)
         if (rslt <= 0)
             results = NULL;
         ldap_msgfree(results);
-        otpd_log_err(EIO, "IO error received on bind socket");
+        otpd_log_err(EIO, "IO error received on bind socket: %s", ldap_err2string(rslt));
         verto_break(ctx.vctx);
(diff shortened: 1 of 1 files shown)


abbra:
This is an experimental fix. Sadly, OpenLDAP library does not allow to see an error reported by the underlying LDAP connection (LDAP structure is private and thus `ld->ld_errno` is inaccessible) so the best we can do is a guessing based on how `ldap_result()`'s processing of timeouts / closed connections is done (in `wait2msg()` helper): https://git.openldap.org/openldap/openldap/-/blob/master/libraries/libldap/result.c#L325-333

abbra:
@rcritten @simo5 @npmccallum fyi, please review.

rcritten:
Do we know what the value of rslt is when a timeout occurs? Should we log ldap_err2string(rslt)?

abbra:
I think we should add that, yes. I'll add it tomorrow.

simo5:
LGTM!

abbra:
master:

* 1cc267118262b58a190ec4702c66f237c432adf9 ipa-otpd: handle LDAP timeout in a better way

PR #5733, opened 2021-04-23
Revert the way hosts/services certs are searched for

rcritten:
Revert the way hosts/services certs are searched for
    
This reverts a change that caused all service certs to be
revoked if any other service was deleted.

There is special handling when passing in host, service or
user into cert_find. The _ldap_search method creates
filters for the object which later in cert_find removes
unrelated certificates.
    
### DIFF
--- a/ipatests/test_integration/test_cert.py
+++ b/ipatests/test_integration/test_cert.py
@@ -168,6 +168,42 @@ class TestInstallMasterClient(IntegrationTest):
         ipaddrs = ext.value.get_values_for_type(x509.IPAddress)
         assert ipaddrs == [ipaddress.ip_address(self.clients[0].ip)]
 
+    def test_service_del_revocation(self):
+        """Test that removing a service revokes its certificate
+
+           This removes the service created in the previous test
+           and ensures that other certificates are not revoked when
+           one is.
+        """
+        hostname = self.clients[0].hostname
+        certfile = os.path.join(paths.OPENSSL_CERTS_DIR, "testservice.pem")
+
+        # This assumes that the cert was issued in the previous test
+        certdata = self.clients[0].get_file_contents(certfile)
+        cert = x509.load_pem_x509_certificate(
+            certdata, default_backend()
+        )
+        serial = cert.serial_number
+
+        # Baseline: the cert should not be revoked yet
+        result = self.master.run_command(["ipa", "cert-show", str(serial)])
+        assert "Revoked: False" in result.stdout_text
+
+        # Delete the service and confirm the cert is revoked
+        tasks.kinit_admin(self.master)
+        self.master.run_command(["ipa", "service-del", f"test/{hostname}"])
+        result = self.master.run_command(["ipa", "cert-show", str(serial)])
+        assert "Revoked: True" in result.stdout_text
+
+        # Get the web cert so we can ensure it isn't revoked
+        certdata = self.clients[0].get_file_contents(paths.HTTPD_CERT_FILE)
+        cert = x509.load_pem_x509_certificate(
+            certdata, default_backend()
+        )
+        serial = cert.serial_number
+        result = self.master.run_command(["ipa", "cert-show", str(serial)])
+        assert "Revoked: False" in result.stdout_text
+
     def test_getcert_list_profile(self):
         """
         Test that getcert list command displays the profile
--- a/ipatests/test_integration/test_cert.py
+++ b/ipatests/test_integration/test_cert.py
@@ -171,15 +171,30 @@ class TestInstallMasterClient(IntegrationTest):
     def test_service_del_revocation(self):
         """Test that removing a service revokes its certificate
 
-           This removes the service created in the previous test
-           and ensures that other certificates are not revoked when
-           one is.
+           Create and revoke a new service ticket on an IPA serve
+           and ensure that other certificates are not revoked.
         """
-        hostname = self.clients[0].hostname
+        hostname = self.master.hostname
         certfile = os.path.join(paths.OPENSSL_CERTS_DIR, "testservice.pem")
+        keyfile = os.path.join(paths.OPENSSL_PRIVATE_DIR, "testservice.key")
 
-        # This assumes that the cert was issued in the previous test
-        certdata = self.clients[0].get_file_contents(certfile)
+        tasks.kinit_admin(self.master)
+
+        cmd_arg = [
+            'ipa-getcert', 'request', '-v', '-w',
+            '-f', certfile,
+            '-k', keyfile,
+            '-K', f'test/{hostname}',
+            '-D', hostname,
+        ]
+        result = self.master.run_command(cmd_arg)
+        request_id = re.findall(r'\d+', result.stdout_text)
+
+        # wait until the certificate is in MONITORING state
+        status = tasks.wait_for_request(self.master, request_id[0], 50)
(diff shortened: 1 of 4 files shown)


ipatests/test_integration/test_cert.py
```diff
@@ -168,6 +168,41 @@ def test_ipa_getcert_san_aci(self):
...
+
+        # This assumes that the cert was issued in the previous test
+        certdata = self.clients[0].get_file_contents(certfile)
+        cert = x509.load_pem_x509_certificate(
+            certdata, default_backend()
+        )
+        serial = cert.serial_number
+
+        # Baseline: the cert should not be revoked yet
+        result = self.master.run_command(["ipa", "cert-show", str(serial)])
```
flo-renaud, line 189: There was a kdestroy in the previous test, this test needs to do a `kinit admin`.

flo-renaud:
Hi @rcritten 
with the patch, only the relevant certs are revoked but the output of `ipa cert-find --status REVOKED` displays additional (and unexpected) certs:
```
# ipa cert-find --status REVOKED
----------------------
8 certificates matched
----------------------
  Issuing CA: ipa
  Subject: CN=host1.domain.com,O=DOMAIN.COM
  Issuer: CN=Certificate Authority,O=DOMAIN.COM
  Not Before: Fri May 07 11:00:42 2021 UTC
  Not After: Mon May 08 11:00:42 2023 UTC
  Serial number: 13
  Serial number (hex): 0xD
  Status: REVOKED
  Revoked: True

  Issuing CA: ipa
  Subject: CN=host1.domain.com,O=DOMAIN.COM
  Issuer: CN=Certificate Authority,O=DOMAIN.COM
  Not Before: Fri May 07 11:01:27 2021 UTC
  Not After: Mon May 08 11:01:27 2023 UTC
  Serial number: 14
  Serial number (hex): 0xE
  Status: REVOKED
  Revoked: True

  Serial number: 8
  Serial number (hex): 0x8

  Serial number: 9
  Serial number (hex): 0x9

  Serial number: 15
  Serial number (hex): 0xF

  Serial number: 11
  Serial number (hex): 0xB

  Serial number: 12
  Serial number (hex): 0xC

  Serial number: 16
  Serial number (hex): 0x10
----------------------------
Number of entries returned 8
----------------------------
```

flo-renaud:
Another issue: `ipa host-del shortname` properly deletes the host shortname.domain.com but does not revoke its certificates.

rcritten:
> Hi @rcritten
> with the patch, only the relevant certs are revoked but the output of `ipa cert-find --status REVOKED` displays additional (and unexpected) certs:

This seems to match current behavior but I'll take a look. It seems these semi-certs are those pulled directly out of LDAP.

rcritten:
> This seems to match current behavior but I'll take a look. It seems these semi-certs are those pulled directly out of LDAP.

Or not, apparently I worked on this patch in multiple environments and my "clean" install was actually not so clean.

rcritten:
I added a test case for both of the failures:
- that no extraneous certs show up in output with --status
- that host-del <shortname> revokes the host certificate

When searching on status we should not search the local LDAP database as we have no way currently of verifying the state. Theoretically we could write our own validator to check notbefore/notafter and potentially an OCSP client to check on status but boy would that be slow and difficult. So just ignore them for now.

The previous code would use the results to filter out certs for the same host but different service. This is still not quite working, I re-submitted the code to run the new tests. Not ready for review yet.

rcritten:
I think I have it worked out and the tests should help verify I haven't regressed anything else.

ipatests/test_integration/test_cert.py
```diff
@@ -168,6 +168,119 @@ def test_ipa_getcert_san_aci(self):
...
 
+    def test_service_del_revocation(self):
+        """Test that removing a service revokes its certificate
+
+           This removes the service created in the previous test
+           and ensures that other certificates are not revoked when
+           one is.
+        """
+        hostname = self.clients[0].hostname
+        certfile = os.path.join(paths.OPENSSL_CERTS_DIR, "testservice.pem")
```
flo-renaud, line 179: The file is called test2.pem in the previous test

ipatests/test_integration/test_cert.py
```diff
@@ -168,6 +168,119 @@ def test_ipa_getcert_san_aci(self):
...
+
+           Ensure that the certificate was revoked.
+        """
+        shortname = 'testdel'
+        hostname = '{}.{}'.format(shortname, self.clients[0].domain.name)
+        certfile = os.path.join(paths.OPENSSL_CERTS_DIR, "testdel.pem")
+        keyfile = os.path.join(paths.OPENSSL_PRIVATE_DIR, "testdel.key")
+
+        tasks.kinit_admin(self.master)
+        self.master.run_command(["ipa", "host-add", hostname])
```
flo-renaud, line 218: Either use --force or provide a dummy --ip-addr $ADDR

flo-renaud (commented):
Hi @rcritten 
I only had a look at the test failures so far, please find the related inline comments.

ipatests/test_integration/test_cert.py
rcritten: Gah, I fixed then reverted this for some unknown reason. Fixed again.

ipatests/test_integration/test_cert.py
rcritten: Heh, so I put --force in the host-add-managedby command instead of host-add :-) fixed.

flo-renaud:
Hi @rcritten 
the test failure in `test_adtrust_install` is a random issue, already reported at [8353](https://pagure.io/freeipa/issue/8353) and unrelated to this PR. 
The situation is different for `test_cert` as it is a new test added by this PR. The test requested 2 certs so far: test.pem and test2.pem, both with a principal "test/client_hostname", but there is no  /var/lib/ipa/certs/httpd.crt on the client.

flo-renaud:
Hi @rcritten
the following test case is failing for me:
- [ ] add a service with 2 certificates, one valid and one revoked, search for valid certs only for this service
```
# ipa cert-find --services other/host1.domain.com --status VALID
----------------------
2 certificates matched
----------------------
  Issuing CA: ipa
  Subject: CN=host1.domain.com,O=DOMAIN.COM
  Issuer: CN=Certificate Authority,O=DOMAIN.COM
  Not Before: Thu May 20 09:59:57 2021 UTC
  Not After: Sun May 21 09:59:57 2023 UTC
  Serial number: 24
  Serial number (hex): 0x18
  Status: VALID
  Revoked: False

  Serial number: 25
  Serial number (hex): 0x19
----------------------------
Number of entries returned 2
----------------------------
```
- [ ] add a service with 2 certificates, one valid and one revoked, search for revoked certs only for this service. The output lists certificates revoked for the same host even if they are not for the same service.

rcritten:
These all revolve around the strange subtraction system that is currently employed.

I wonder if we'd be better off reverting my original patch from master and starting over completely :/ The current patch in master as-is is pretty dangerous (all servcies for a host are revoked) and we may soon forget that it's lurking there.

rcritten:
https://github.com/freeipa/freeipa/pull/5784

I'm going to keep this PR open but move it a draft so I don't forget about it.

rcritten:
I was keeping this around as a reminder. Going to shut it down since it is going to require some additional focus I just don't have time for now.

PR #5734, opened 2021-04-23
Design doc to allow LDAP bind using the RADIUS auth type

rcritten:
The RADIUS auth type is only supported with Kerberos currently.
This design proposes a way to make it work with LDAP binds
as well without relying ok workarounds.

### DIFF
--- a/doc/designs/ldap_pam_passthrough.md
+++ b/doc/designs/ldap_pam_passthrough.md
@@ -1,0 +1,330 @@
+**IMPORTANT**: This is a design proposal and is not implemented yet.
+
+# LDAP PAM Passthrough support
+
+## Overview
+
+Many organizations have authentication mechanisms already in place.
+They may not want to have IPA be the central repository for authentication.
+
+RADIUS is a common authentication protocol used for external authentication
+into existing systems. IPA currently has support for verifying credentails
+over RADIUS for Kerberos connections using the radius authentication
+indicator, but this does not work with LDAP authentication.
+
+For this document "PAM Passthrough" is defined as any subsequent
+plugin that handles authentication of the user entry using the PAM stack.
+
+### Expected Workflow
+
+There are a lot of components potentially involved in LDAP authentication
+over RADIUS:
+
+ - 389-ds
+ - PAM
+ - SSSD
+ - KDC
+ - ipa-otpd
+ - the remote RADIUS server
+
+The workflow starts with an LDAP bind.
+
+ - On an LDAP BIND with uid=user,cn=users,cn=accounts,$SUFFIX, the BIND
+   request will get processed by the IPA password plugin preop.
+   If both radius and otp auth types are set and there are no tokens
+   the plugin will return 0, allowing further authentication to happen.
+   389-ds treats this as no authentication decision so allows other
+   plugins to try.
+
+ - At this point another 389-ds plugin can step in to handle the
+   authentication using the PAM stack.
+
+ - As PAM authentication processing happens, if pam_sss.so is present in
+   the PAM stack, it will attempt to perform password-based authentication
+   for the 'user' account using the provided credentials.
+
+ - Since the user account matches the IPA domain, it will be treated as
+   Kerberos authentication against IPA KDC running on the same host, as
+   we are authenticating on a IPA server.
+
+ - The IPA KDB driver in the KDC will notice that the 'user@IPA' principal
+   has the 'radius' pre-authentication method configured with TL data
+   "otp\0[{\"indicators\": [\"radius\"]}]" meaning it should advertise
+   the OTP pre-auth mechanism to the Kerberos client.
+
+ - The client (SSSD) notices the availability of the OTP pre-auth
+   mechanism and uses host principal's TGT as its FAST channel wrapper to
+   proceed with OTP pre-auth.
+
+ - OTP pre-auth mechanism will talk between KDC and the client to ask
+   for additional details (OTP value) via prompting mechanism it has.
+
+ - The client (SSSD) will return the OTP value and the KDC then will
+   issue a RADIUS request "Accept-Request" to a RADIUS server configured
+   in the KDC configuration. The OTP value in this case is the
+   credentials the user provided.
+
+ - ipa-otpd handles OTP requests and will connect to LDAP (the LDAP URI
+   is passed as part of ipa-otpd@.service definition from
+   /etc/ipa/default.conf, so it'll be an LDAPI access). ipa-otpd will
+   parse the RADIUS packet and look up requested user principal entry from
+   LDAP.
+
+ - If the user principal entry has the 'radius' authentication indicator
+   configured (or it is default for IPA deployment) and there is a
+   RADIUS proxy link in the user entry (there is no default so it must
+   be set per user), it will send the same RADIUS packet to the RADIUS
+   server configured as a proxy link with the credentials provided by
+   KDC
+
+ - For a native OTP setup where the user has the 'otp' authentication
+   indicator the process is about the same: instead of sending a RADIUS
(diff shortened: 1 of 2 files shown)


doc/designs/ldap_pam_passthrough.md
```diff
@@ -0,0 +1,315 @@
...
+   the plugin will return 0, allowing further authentication to happen.
+   389-ds treats this as no authentication decision so allows other
+   plugins to try.
+
+ - If the user DN is in the allowed configuration for PAM pass-through
+   then the PAM stack is loaded and processed. One can limit the
+   subtrees that may be authenticated. In the IPA case this is basically
+   moot because of the flat DIT.
+
+ - PAM is invoked an dpam_sss.so and attempts to perform a
```
abbra, line 51: I would suggest to change here a sentence to:
> As PAM authentication processing happens, if `pam_sss.so` is present in the PAM stack, it will attempt to perform password-based authentication for 'user' account using the provided credentials.

doc/designs/ldap_pam_passthrough.md
```diff
@@ -0,0 +1,315 @@
...
+   authentication against IPA KDC running on the same host, as we are
+   authenticating on a IPA server.
+
+ - The IPA KDB driver in the KDC will notice that the 'user@IPA' principal
+   has the 'radius' pre-authentication method configured with TL data
+   "otp\0[{\"indicators\": [\"radius\"]}]" meaning it should advertise
+   the OTP pre-auth mechanism to the Kerberos client.
+
+ - The client (SSSD) notices the availability of the OTP pre-auth
+   mechanism uses host principal's TGT as its FAST channel wrapper to
```
abbra, line 65: the client notices the availability ... _and_ uses ...

doc/designs/ldap_pam_passthrough.md
```diff
@@ -0,0 +1,315 @@
...
+ - Add a RADIUS proxy configuration in IPA
+
+ - Add this proxy to one or more users
+
+ - set default authentication indicator in IPA to 'radius, otp' or on
+   one or more of those users.
+
+ - Do *not* set a password or principal key for those users
+
+ - The user has no OTP tokens
```
abbra, line 113: Previous items on the list use imperatives, may be turn this one into a similar sentence?
> Do *not* define any OTP tokens for those users

doc/designs/ldap_pam_passthrough.md
```diff
@@ -0,0 +1,315 @@
...
+authenticted, if there is one. This is the two-step first check
+OTP, then check the password.
+
+Since a plugin returning 0 will allow subsequent authentication
+for the case of RADIUS we need to enforce the no password and
+radius server requirements because with PAM passthrough enabled it
+becomes a defacto default method. In fact we may want to always ensure
+there is no RADIUS server defined if the RADIUS authentication
+indicator is not set.
+
```
abbra, line 225: An additional thought is to simply bring the PAM pass-through method into ipa-pwd-extop, like we have in slapi-nis plugin. We can copy over the code from slapi-nis.

doc/designs/ldap_pam_passthrough.md
```diff
@@ -0,0 +1,315 @@
...
+In order to test we'll need to setup a RADIUS server to test against.
+The pyrad project provides a sample in
+https://raw.githubusercontent.com/pyradius/pyrad/master/example
+
+It would need to be adapted for our needs to actually do authentication
+and ideally dynamically setup its listening hosts. The passwords
+could be hardcoded with a "good" one that is always accepted.
+
+And the ldapserver PAM service needs to be created and cleaned up.
+Making a copy of system-auth is sufficient.
```
abbra, line 264: If we use internal implementation of PAM pass-through, we don't need to manage 389-ds plugin setup and handle individually its configuration.

abbra (changes requested):
Thanks for the design, @rcritten. I have few stylistic changes and also few suggestions as to how we should work with PAM pass-through. I think we should reuse slapi-nis plugin's implementation of the PAM operations for couple reasons:
- this component is already deployed for authenticating AD users through the compat tree so we know the code works
- we would have no need to fall back to any other plugin in 389-ds, making a logic a bit more clear and predictable
- no need for additional configuration
- PAM pass-through plugin in 389-ds is not really tested upstream

doc/designs/ldap_pam_passthrough.md
rcritten: done

doc/designs/ldap_pam_passthrough.md
rcritten: done

doc/designs/ldap_pam_passthrough.md
rcritten: done

doc/designs/ldap_pam_passthrough.md
rcritten: done

rcritten:
Thanks for the review.

I generalized things a bit as the PAM passthrough plugin is actually an implementation detail and not dependent upon this design.

We can use either slapi-nis or the 389-ds plugin (or something else), I think the only thing that would vary is configuration.

doc/designs/ldap_pam_passthrough.md
```diff
@@ -0,0 +1,327 @@
...
+over RADIUS for Kerberos connections using the radius authentication
+indicator, but this does not work with LDAP authentication.
+
+For this document "PAM Passthrough" is defined as any subsequent
+plugin that handles authentication of the user entry using the PAM stack.
+
+There are two known possibilities:
+
+- The 389-ds PAM Passthrough Plugin
+- The slapi-nis 389-ds plugin
```
abbra, line 19: Sorry for making a confusion here. I did not suggest to specify `slapi-nis` as one of authentication plugins here. What I suggested is that we could incorporate the code that we have in `slapi-nis` into `ipa-pwd-extop` so that it would cover PAM authentication on itself, without failing over to other authentication plugins.

`slapi-nis` plugin only provides PAM authentication for its own Compatibility Tree entries that correspond to AD users, so by design it would not be able to handle DNs in the primary tree.

rcritten: Ok I can clarify something like the way slapi-nis does it, but it basically uses a copy of the 389-ds pam pass through plugin...

I was trying to make this backend agnostic. I don't think it really matters design-wise what does the PAM auth, as long as it succeeds and fails when it should, and we do the right thing in that case.

rcritten:
Re-based patch so CI can pass. Clearing changes requested because I believe I've applied all suggestions.

abbra:
@rcritten it all looks good, I have only one comment. Since any documentation committed to master tree is visible on readthedocs site, it would probably be good to add an explicit statement that the proposal is not implemented yet. We can remove the sentence once we have implemented the proposal.

rcritten:
Good call. I added a bolded statement as the first line of the document.

abbra:
LGTM

abbra:
Added ack label. I think we can backport the design to ipa-4-9 once implementation is done. At the moment it is OK to keep it in master only.

rcritten:
Removing the ack for now. We discovered a related issue in ticket 8820.

rcritten:
I added a reference to the multiple auth mechanism issue in the "IPA password plugin changes" section as this is where it would most likely be triggered in an LDAP authentication case (the purpose of the design). The idea is to maintain parity between Kerberos and LDAP authentication so the design applies to both.

abbra:
LGTM

rcritten:
master:

* 16ab690bf629726df38a35333cc469dc067e1b9c Design doc to allow LDAP bind using the RADIUS auth type

PR #5737, opened 2021-04-25
pylint: Adapt to new Pylint 2.8

stanislavlevin:
- globally ignore `consider-using-with`
- fix `consider-using-max-builtin`
- explicitly enable pylint on project configuration
- unpin Pylint
- added transformation for IntegrationTest attributes(will work
  unless explicitly defined)

Fixes: https://pagure.io/freeipa/issue/8818

### DIFF
--- a/ipatests/azure/templates/autoconf-fedora.yml
+++ b/ipatests/azure/templates/autoconf-fedora.yml
@@ -7,6 +7,7 @@ steps:
     echo "Running autoconf generator"
     ./autogen.sh \
         ${{ parameters.options }} \
+        --enable-pylint \
         --enable-rpmlint \
 
   displayName: Configure the project
--- a/ipatests/azure/templates/prepare-lint-fedora.yml
+++ b/ipatests/azure/templates/prepare-lint-fedora.yml
@@ -2,6 +2,5 @@ steps:
 - script: |
     set -e
     sudo dnf -y install python3-pip
-    # https://pagure.io/freeipa/issue/8771
-    python3 -m pip install --user --upgrade astroid pylint==2.7.1
+    python3 -m pip install --user --ignore-installed pylint
   displayName: Install Lint dependencies
--- a/pylint_plugins.py
+++ b/pylint_plugins.py
@@ -182,50 +182,6 @@ ipa_class_members = {
         'validatedns',
         'normalizedns',
     ],
-    'ipatests.test_integration.base.IntegrationTest': [
-        {'domain': [
-            {'name': dir(str)},
-        ]},
-        {'master': [
-            {'config': [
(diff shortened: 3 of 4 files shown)


stanislavlevin:
re-runned on updated 
- pylint 2.8.1
- astroid 2.5.6

ipapython/dn.py
```diff
@@ -445,13 +445,11 @@ def _adjust_indices(start, end, length):
         end = length
     elif end < 0:
         end += length
-        if end < 0:
-            end = 0
+        end = max(end, 0)
```
tiran, line 448: I find the old code more readable. It's more obvious was is happening here. It might even be faster (not that it matters).

tiran (changes requested):
Could you please update ``.wheelconstraints.in`` and restrict pylint version to < 2.9, too? This avoids future breakage. IMHO we should update and address pylint issues whenever we add support for a new Fedora release, not when upstream releases a new version of pylint

```raw
# .wheelconstraints.in
# keep pylint version in sync with current Fedora release
pylint < 2.9
```

stanislavlevin:
The recent changes to Pylint task of Azure intentionally do the opposite - catch and fix new true warnings/errors (as well as false negative/positive, unfortunately) introduced by upstream's releases,; in other words, check against PyPI instead of any OS distro. Usually such fixes are compatible with previous Pylint versions and can be applied by downstream if required. The only problem is the breakage of CI for 1-2 days dependending on found issues, of course, this can be overcomed by pinning of Pylint version.

In this case the changes are trivial, but unfortunately there was no review until today. I started thinking that I'm from "@umn.edu" :)

abbra:
@stanislavlevin please don't feel offended. There is a lot of parallel work that eats into review times, so many people aren't available as they were earlier. I myself have little time for reviews until SambaXP...

ipapython/dn.py
stanislavlevin: Logically it's the same. We can skip below all together:
```
consider-using-max-builtin (R1731)
Consider using '%s' instead of unnecessary if block Using the max builtin instead of a conditional improves readability and conciseness.

consider-using-min-builtin (R1730)
Consider using '%s' instead of unnecessary if block Using the min builtin instead of a conditional improves readability and conciseness.
```

tiran (approved):
Thanks, looks good :+1:

stanislavlevin:
@tiran, thank you for your time!

@abbra, I see. I just was worried that this was blocking the all users of CI.

flo-renaud:
master:

* bdb26968052d7152de37d01d9314e8f04eb58965 pylint: Adapt to new Pylint 2.8

PR #5740, opened 2021-04-27
Constrain pylint to supported versions 

tiran:
Two, three times a year PR-CI starts failing because tox tests pull in a
newer version of pylint with new warnings. To reduce breakage this
change constraints pylint (and indirectly astroid) to latest tested
minor version. The constraint should be updated when FreeIPA starts to
support a new Fedora version with more recent pylint.

Related: pagure.io/freeipa/issue/8818

### DIFF
--- a/.wheelconstraints.in
+++ b/.wheelconstraints.in
@@ -9,5 +9,7 @@ ipapython == @VERSION@
 ipaserver == @VERSION@
 ipatests == @VERSION@
 
-# upstream pylint 1.7.5 fixed bad python3 import of stat module
-pylint >= 1.7.5
+# keep pylint version in sync with current Fedora release
+# F34 has 2.6
+# https://pagure.io/freeipa/issue/8818 added pylint 2.8 support
+pylint < 2.9
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -347,12 +347,8 @@ BuildRequires:  python3-polib
 BuildRequires:  python3-pyasn1
 BuildRequires:  python3-pyasn1-modules
 BuildRequires:  python3-pycodestyle
-%if 0%{?fedora} || 0%{?rhel} > 8
-# https://bugzilla.redhat.com/show_bug.cgi?id=1648299
-BuildRequires:  python3-pylint >= 2.1.1-2
-%else
-BuildRequires:  python3-pylint >= 1.7
-%endif
+# .wheelconstraints.in limits pylint version in Azure and tox tests
+BuildRequires:  python3-pylint
 BuildRequires:  python3-pytest-multihost
 BuildRequires:  python3-pytest-sourceorder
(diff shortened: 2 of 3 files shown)


stanislavlevin:
@tiran,
tbh, I don't understand why should we bind Pylint to Fedora.
For example, new "old" Pylint landed in Fedora and ... there will be the same, CI stops until somebody fixes "new" errors.
Or is there any internal Fedora CI which blocks new Pylint if that breaks distro's FreeIPA?

My distro experience is:
1) run pylint on FreeIPA RPM build (each build, everyday distro CI)
2) on new pylint PyPI release I trigger FreeIPA RPM build, block Pylint downstream release if required. Feedback to FreeIPA upstream + fixes if required.

stanislavlevin:
How do we know when Fedora starts to support new Pylint?

tiran:
It's the easiest way to prevent out-of-bounce test failures. Each new version of pylint breaks our CI. I chose Fedora because that's what our CI is using as base image.

tiran:
In case you are going to need a more recent version of pylint for your work, then it's totally ok to introduce fixes and bump to constrain file. This should be a deliberate decision and not imposed on us by external release cycles. Technically speaking each minor release of pylint is a breaking change because it breaks our CI. I want to prevent future out-of-bounce breakage.

In my experience the cost of PR-CI breakage is way higher than benefits of new pylint checkers.

stanislavlevin:
Pros:
- more stable CI (bumping supported version of Pylint when _we_ want to do it instead of _Pylint upstream_ wants)
- there is no urgent necessity of fixing linting issues and urgent review

Cons:
- not latest version may not catch some code issues
- one version may be pinned forever (possible downgrade of Fedora's Pylint).

stanislavlevin (approved):

rcritten:
Won't this potentially break 'make lint' if we run it as a developer? Shouldn't the spec have the same constraints as wheel?

stanislavlevin:
If a developer finds errors related to new Pylint on local `make pylint` then it's time to fix it in upstream.

tiran:
> Won't this potentially break 'make lint' if we run it as a developer? Shouldn't the spec have the same constraints as wheel?

That's potentially possible but less likely and less severe. At the moment we always use the latest pylint from PyPI. Linting in tox acts as a very sensitive canary in the coal mine.

With this PR we have more control. You may notice breakage if you are the first engineer that updates to a new Fedora version. Just let @abbra update first. :)

rcritten:
Ok, ack.

flo-renaud:
master:

* 6d0d1c70881c977cb942046a4642b1eeeeba7563 Constrain pylint to supported versions

flo-renaud:
@tiran the PR needs a manual backport to ipa-4-8, can you handle it? Thanks.

PR #5744, opened 2021-04-29
WIP: Convert HMAC to EVP interface

simo5:
I haven't even compiled this yet, but I thought it would be better to post it then forget in my branch

@abbra  @tiran  let me know if you'd like this contribution in this form and I will go through the paces of testing with openssl3. Or we can simply leave it here until we have an OS that carries openssl3 we can run tests against.

### DIFF
--- a/daemons/ipa-slapi-plugins/libotp/hotp.c
+++ b/daemons/ipa-slapi-plugins/libotp/hotp.c
@@ -47,12 +47,20 @@
 #include <string.h>
 #include <syslog.h>
 #include <time.h>
-#include <openssl/hmac.h>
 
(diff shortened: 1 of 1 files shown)


abbra:
I am OK with this approach. As discussed on IRC, we'll also need to internalize MD4 code, so having a similar consumption of the HMAC wrapper is OK.

tiran:
OpenSSL 3.0.0 still has MD4 support. It's in the legacy provider: ``OSSL_PROVIDER_load(NULL, "legacy")``

abbra:
@tiran it is not built when building FIPS provider.

PR #5748, opened 2021-05-04
ipatests: increase timeout for test_commands up to 1.5 hours

wladich:
Normally it takes 50 minutes for test_commands test suite
to complete. But due to fluctuations in virtualized environment
sometimes it exceeds 60 minutes which produces random failures
in nightly runs.

### DIFF
--- a/ipatests/prci_definitions/gating.yaml
+++ b/ipatests/prci_definitions/gating.yaml
@@ -140,7 +140,7 @@ jobs:
         build_url: '{fedora-latest/build_url}'
         test_suite: test_integration/test_commands.py
         template: *ci-master-latest
-        timeout: 3600
+        timeout: 4800
         topology: *master_1repl_1client
 
   fedora-latest/test_kerberos_flags:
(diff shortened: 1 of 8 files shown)


tiran:
IMHO it would make more sense to move slow tests to a separate file. ``test_commands`` was initially designed as a kitchen-sink file for easy, fast testing of simple commands. This should be one of the fastest PR-CI test files.

wladich:
The purpose of this PR is to reduce time spent by FreeIPA team for inspecting nightly tests failures and to reduce risk of missing real failure in the test suite being obscured by time out.
While I agree that keeping the test suite execution time small is a proper goal, I suggest it should be done separately. Could you file a Pagure issue if you feel it is important?

rcritten:
I agree with Christian that we should restructure the test. acking to avoid too many nightly failures while we work that out.

I'll file a ticket to examine this test case to see what, if any tests can be moved around to reduce the time it takes.

rcritten:
https://pagure.io/freeipa/issue/8836 if you want to add any comments.

wladich:
master:

* 47d0ef40c82815135bbe255bff91353ae2589026 ipatests: increase timeout for test_commands up to 1.5 hours

PR #5749, opened 2021-05-05
pkispawn: override the AJP address

flo-renaud:
The /etc/hosts file currently defines the following:
```
127.0.0.1  localhost
::1  localhost
```

but pki defines AJP connectors using localhost4 and localhost6:
```
    <Connector port="8009" protocol="AJP/1.3" redirectPort="8443" address="localhost4" name="Connector1" secret="..."/>
    <Connector port="8009" protocol="AJP/1.3" redirectPort="8443" address="localhost6" name="Connector2" secret="..."/>
```
(since commit 1906afbeb3c8b7140601be7f9bee2f7fef5b0a5e, in order to fix
rhbz#1780082)

Override the definition with pki_ajp_host_ipv4 = 127.0.0.1 and pki_ajp_host_ipv6 = ::1 in the config file provided to pkispawn.

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -537,6 +537,11 @@ class CAInstance(DogtagInstance):
         if os.path.exists(paths.IPA_CA_CRT):
             cfg['pki_cert_chain_path'] = paths.IPA_CA_CRT
 
+        # Use IP address instead of default localhost4 and localhost6
+        # because /etc/hosts does not always define them
+        cfg['pki_ajp_host_ipv4'] = "127.0.0.1"
+        cfg['pki_ajp_host_ipv6'] = "::1"
+
         if self.clone:
             if self.no_db_setup:
                 cfg.update(


rcritten:
Works great. Do we want to fix existing installations as well?

flo-renaud:
> Works great. Do we want to fix existing installations as well?

I don't have a strong opinion on that one. If the install succeeded I guess we can assume that /etc/hosts has the proper definition and we don't really need to patch server.xml, except if we want to have a consistent configuration for new installs/upgrades.

rcritten:
Yes, that sounds reasonable to me. It's fine as-is, ack.

rcritten:
Before I push this do we want a ticket for it? The commit message describes what the change is quite well but do we want the back story associated with the need for this change in a ticket as well?

I'm fine either way, just being cautious. If you decide that no ticket is needed feel free to add the ack back.

flo-renaud:
Created issue https://pagure.io/freeipa/issue/8851 and amended the commit msg to refer to the issue

stanislavlevin:
This is the annoying error in logs.

abbra:
As Rob is on vacation, I'm adding ACK.

abbra:
master:

* 379d5da0ae5e54d4d4b8d73979013fad96d123dc pkispawn: override AJP connector address

stanislavlevin:
@abbra, thank you!
Most important is that these addresses are queried to BIND(via NSS configuration) and who knows what the answer will be.

abbra:
Yes, this was raised in one of freeipa-container issues. Since it can be overridden at runtime, changing to direct IP addresses is OK.

PR #5750, opened 2021-05-05
azure: Re-group test envs

stanislavlevin:
This is the experimental fix to workaround the issue with
PKI component, which is sensitive to slow systems(at least,
appropriate delays and timeouts should be adjusted for such).

Somehow Azure's test envs became slower then they were earlier
(for example, CA subsystem start changed
~(20-30)sec -> ~(45-60)sec). This triggered various issues with
subsystems of PKI.

To release CPU resources(Azure VM provides 2 cores, 1 thread per
core) the extreme usage of swap was reduced to the limits provided
by Azure(7Gb + 4Gb swap). Server/replica container consumes ~2Gb,
client one ~0.5Gb.

Also this increases the number of Azure's Gating jobs from 4 to 6.

Related: https://lists.fedorahosted.org/archives/list/freeipa-devel@lists.fedorahosted.org/thread/AQRKHR56L3AHST2XGSNQGYS22ORYE5MI/

### DIFF
--- a/ipatests/azure/azure_definitions/gating-fedora.yml
+++ b/ipatests/azure/azure_definitions/gating-fedora.yml
@@ -17,14 +17,6 @@ vms:
     tests:
     - test_integration/test_forced_client_reenrollment.py
 
-  - container_job: external_ca_ExternalCAInstall
-    tests:
-    - test_integration/test_external_ca.py::TestExternalCAInstall
-
-  - container_job: membermanager
-    tests:
-    - test_integration/test_membermanager.py
-
 - vm_jobs:
   - container_job: InstallDNSSECFirst
     containers:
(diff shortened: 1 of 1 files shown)


stanislavlevin:
The main issue is the increased number of jobs, with new capacity only 1 run(PR, merge, manual) can be processed simultaneously.
As an alternative solution I can try PKI's `PKISPAWN_STARTUP_TIMEOUT_SECONDS` env variable (if not set timeout is 60sec). This will make startup timeout consistent with IPA's one (startup_timeout=120sec, which in particular, is used in `ipa-pki-wait-running`).

stanislavlevin:
Closed in favour of https://github.com/freeipa/freeipa/pull/5751

PR #5751, opened 2021-05-05
pkispawn: Make timeout consistent with IPA's startup_timeout

stanislavlevin:
This is the experimental fix to workaround the issue with PKI on Azure Pipelines. PKI is sensitive to slow systems (at least, appropriate delays and timeouts should be adjusted for such).
    
Somehow Azure's test envs became slower then they were earlier (for example, CA subsystem start changed ~(20-30)sec -> ~(45-60)sec). This triggered various issues with subsystems of PKI in CI.
    
PKI honors `PKISPAWN_STARTUP_TIMEOUT_SECONDS` env variable (if not set the timeout is 60sec) for pkispawn tool. The default is inconsistent with IPA's one (startup_timeout=120sec), which in particular, is used in ipa-pki-wait-running tool).
 
Related: https://pagure.io/freeipa/issue/8830

### DIFF
--- a/ipaserver/install/dogtaginstance.py
+++ b/ipaserver/install/dogtaginstance.py
@@ -193,6 +193,10 @@ class DogtagInstance(service.Service):
         parameters.
         """
         subsystem = self.subsystem
+        spawn_env = os.environ.copy()
+        timeout = str(api.env.startup_timeout)
+        spawn_env["PKISPAWN_STARTUP_TIMEOUT_SECONDS"] = timeout
+
         args = [paths.PKISPAWN,
                 "-s", subsystem,
                 "-f", cfg_file,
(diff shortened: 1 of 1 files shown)


fcami (changes requested):
Hi @stanislavlevin thanks for the PR, LGTM.
Could you please create an issue at pagure and mention it in the commit message (and reference the email thread in the issue, rather than in the commit)?

stanislavlevin:
Hi, @fcami. My bad, fixed. Thank you!

fcami (approved):
ACK

rcritten:
master:

* 2410fd25a81454aef74dd28cc49a350ada5445b3 pkispawn: Make timeout consistent with IPA's startup_timeout

PR #5752, opened 2021-05-06
Fix update_dna_shared_config to wait for both entries

tiran:
update_dna_shared_config plugin now waits for presence of both
``dnaHostname=fqdn+dnaPortNum=0`` and
``dnaHostname=fqdn+dnaPortNum=389`` entries before it updates the
entries.

The commit also refactors the code of update_dna_shared_config for
subordinate id support.

Fixes: https://pagure.io/freeipa/issue/8831

### DIFF
--- a/ipaserver/install/plugins/update_dna_shared_config.py
+++ b/ipaserver/install/plugins/update_dna_shared_config.py
@@ -15,73 +15,95 @@ logger = logging.getLogger(__name__)
 
 register = Registry()
 
-
 @register()
 class update_dna_shared_config(Updater):
-    def execute(self, **options):
-        method = options.get('method', "SASL/GSSAPI")
-        protocol = options.get('protocol', "LDAP")
+    dna_plugin_names = ('posix IDs',)
 
-        dna_bind_method = "dnaRemoteBindMethod"
-        dna_conn_protocol = "dnaRemoteConnProtocol"
-        dna_plugin = DN(('cn', 'Distributed Numeric Assignment Plugin'),
-                        ('cn', 'plugins'),
-                        ('cn', 'config'))
-        dna_config_base = DN(('cn', 'posix IDs'), dna_plugin)
+    dna_plugin_dn = DN(
+        ('cn', 'Distributed Numeric Assignment Plugin'),
+        ('cn', 'plugins'),
+        ('cn', 'config')
+    )
 
-        conn = self.api.Backend.ldap2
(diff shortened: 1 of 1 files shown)


ipaserver/install/plugins/update_dna_shared_config.py
```diff
@@ -15,73 +15,95 @@
...
+
+        In case the server is not able to create the entry
+        The loop gives a grace period of 60s before logging
+        the failure to update the shared config entry and return
+        """
+
+        method = options.get('method', "SASL/GSSAPI")
+        protocol = options.get('protocol', "LDAP")
+        dna_bind_method = "dnaRemoteBindMethod"
+        dna_conn_protocol = "dnaRemoteConnProtocol"
```
rcritten, line 86: I realize you just preserved what is there but dna_bind_method and dna_conn_protocol are used just once. Do we need variables to store the values?

ipaserver/install/plugins/update_dna_shared_config.py
```diff
@@ -15,73 +15,95 @@
...
-        # In case the server is not able to create the entry
-        # The loop gives a grace period of 60s before logging
-        # the failure to update the shared config entry and return
-        #
         max_wait = 30
+
+        conn = self.api.Backend.ldap2
         fqdn = self.api.env.host
+
         for _i in range(0, max_wait + 1):
```
rcritten, line 93: Again, copied code, but max_wait is assigned then used once, here, and it counts from 0 but adds 1 so this will range 0..30 for 31 iterations, up to 62 seconds.

rcritten:
I tested in a fresh install and never got both entries, just 32 logs of:

Got only one entry. Retry again in 2 sec.

Post-install both are there but this extended install time by a minute.

ipaserver/install/plugins/update_dna_shared_config.py
tiran: good point, let me clean this up.

rcritten:
And now I can't duplicate the wait I saw before. Both entries are available immediately. The patch looks ok to me.

fcami (approved):
ACK

rcritten:
master:

* 45d9a66faafbbe015af8d1d05d969a70f209c1a8 Fix update_dna_shared_config to wait for both entries
* 5c92a718b38ff971cb25a3028d2e61a430d56406 Move constants, document timeout loop

PR #5756, opened 2021-05-06
Only attempt to upgrade ACME configuration files if deployed

rcritten:
This can happen on upgrades from older deployments that lack
an ACME installation and don't meet the minimum requirements
to deploy one automatically.

Also don't consider missing ACME schema a total failure, just
log and skip it.

https://pagure.io/freeipa/issue/8832

TESTING: Testing is difficult because on upgrade IPA will deploy ACME if it isn't already configured. One way to avoid this is to twiddle with /usr/share/pki/VERSION and set the value < 10.10.0.

If ACME is already deployed you can remove it with:

```
# pki-server acme-undeploy
# rm -rf /etc/pki/pki-tomcat/acme
```

### DIFF
--- a/ipaserver/install/server/upgrade.py
+++ b/ipaserver/install/server/upgrade.py
@@ -1122,7 +1122,8 @@ def ca_upgrade_schema(ca):
             acme_schema_ldif = path
             break
     else:
-        raise RuntimeError('ACME schema file not found')
+        logger.info('ACME schema is not available')
+        return False
 
     schema_files=[
         '/usr/share/pki/server/conf/schema-certProfile.ldif',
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @rcritten 
if 8.4 spec file bumps pki_version to 10.10, I believe that this change won't be needed at all.
I tried to upgrade from 8.3 to 8.4, saw the issue with missing issuer.conf file. Then I manually patched the upgrade.py file, `ipactl start` seemed to work (triggered ipa-server-upgrade) but after upgrading pki server, ipa-acme-manage status returns 'Failed to authenticate to CA REST API'.

rcritten:
These aren't guarded by the minimum_acme_support() call so we'd either need to add that, or go with my proposal, or perhaps both (belt and suspenders).

The failure to authenticate is likely misleading. ACME needs ipa-ca.$DOMAIN as a SAN but this isn't added on upgrades in mixed environments.

This whole ACME code shows we need some sort of upgrade and mixed environmetn test in PR-CI.

flo-renaud:
Hi @rcritten 
I'm OK with your changes but please also bump pki version in RHEL 8.4 spec file as it will prevent having a partial update.

rcritten:
master:

* 8dac8ad834164062ef0a49d20c7bfcdf1773fbe5 Only attempt to upgrade ACME configuration files if deployed

PR #5760, opened 2021-05-11
ipatests: Tests for RFE auto-private-groups cli option

menonsudhir:
This patch tests the various values available for the option auto-private-groups for the idrange-mod command and
ensure that the CLI displays the correct result when the value is changed.

### DIFF
--- a/ipatests/prci_definitions/temp_commit.yaml
+++ b/ipatests/prci_definitions/temp_commit.yaml
@@ -68,7 +68,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-latest/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
+        test_suite: test_integration/test_trust.py::TestNonPosixAutoPrivateGroup test_integration/test_trust.py::TestPosixAutoPrivateGroup
         template: *ci-master-latest
-        timeout: 3600
-        topology: *master_1repl_1client
+        timeout: 4800
+        topology: *adroot_adchild_adtree_master_1client
--- a/ipatests/test_integration/test_trust.py
+++ b/ipatests/test_integration/test_trust.py
@@ -6,6 +6,7 @@ import re
 import textwrap
 import time
 import functools
+import subprocess
 
 import pytest
 
@@ -72,6 +73,7 @@ class BaseTestTrust(IntegrationTest):
         cls.srv_gc_record_name = \
             '_ldap._tcp.Default-First-Site-Name._sites.gc._msdcs'
         cls.srv_gc_record_value = '0 100 389 {}.'.format(cls.master.hostname)
+        cls.existing_rangename = cls.ad_domain.upper() + "_id_range"
 
     @classmethod
     def check_sid_generation(cls):
@@ -105,6 +107,47 @@ class BaseTestTrust(IntegrationTest):
         assert expected_text in result.stdout_text
         expected_text = 'iparangetype: %s\n' % expected_type
         assert expected_text in result.stdout_text
+
+    def mod_idrange_auto_private_group(
+        self, auto_priv_grp_val='false'
+    ):
+        """
+        Set the auto-private-group option of the default trusted
+        AD domain range and restart sssd
+        """
+        rangename = self.existing_rangename
+        error_msg = "ipa: ERROR: no modifications to be performed"
+        cmd = ["ipa", "idrange-mod", rangename,
+               "--auto-private-groups", auto_priv_grp_val]
+        try:
+            self.master.run_command(cmd, raiseonerr=False)
+            self.master.run_command(["systemctl", "restart", "sssd"])
+        except subprocess.CalledProcessError as e:
+            # ignore error if the auto-private-group is already set
+            # to the expected value
+            if error_msg not in e.stderr:
+                raise
+
+    def get_user_id(self, host, username):
+        """
+        User uid and gid is parsed from the output
+        of id command and a tuple is returned with (uid,gid)
+        """
+        id_output = host.run_command(["id", username])
+        match = re.match(
+            r"^uid=(?P<uid>\d+).*gid=(?P<gid>\d+)", id_output.stdout_text
+        )
+        uid = match.group('uid')
+        gid = match.group('gid')
+        return(uid, gid)
+
+    def add_trust(self, rangetype):
+        tasks.configure_dns_for_trust(self.master, self.ad)
+        tasks.establish_trust_with_ad(
+            self.master,
+            self.ad_domain,
+            extra_args=['--range-type', rangetype]
+        )
 
     def remove_trust(self, ad):
         tasks.remove_trust_with_ad(self.master,
@@ -993,3 +1036,248 @@ class TestTrust(BaseTestTrust):
             self.master.run_command(['rm', '-f', ad_zone_file])
             tasks.configure_dns_for_trust(self.master, self.ad)
             self.remove_trust(self.ad)
+
+
+class TestNonPosixAutoPrivateGroup(BaseTestTrust):
+    """
+    Tests for auto-private-groups option with non posix AD trust
+    """
+    topology = 'line'
+    num_ad_domains = 1
+    num_clients = 1
+
+    def test_add_nonposix_trust(self):
+        self.add_trust('ipa-ad-trust')
+
+    def test_modify_auto_private_groups_value_default_trusted_range(self):
+        """
+        Modify existing range for default trusted AD domain range
+        with auto-private-groups set as true
+        """
+        domainusers = 'domain users'
+        nonposixuser = "nonposixuser@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("true")
+        result = self.clients[0].run_command(["id", nonposixuser])
+        uid_reg = (
+            r"^uid=(\d+)\({0}\)".format(nonposixuser)
+        )
+        gid_reg = (
+            r"gid=(\d+)\({0}\)".format(nonposixuser)
+        )
+        grp_reg = (
+            r"groups=(\d+)\({0}\),(\d+)\({1}@{2}\)".format
+            (nonposixuser, domainusers, self.ad_domain)
+        )
+        assert re.search(uid_reg, result.stdout_text)
+        assert re.search(gid_reg, result.stdout_text)
+        assert re.search(grp_reg, result.stdout_text)
+
+    def test_modify_auto_private_groups_value_trusted_range_hybrid(self):
+        """
+        Modify existing range for default trusted AD domain range
+        with auto-private-groups set as hybrid
+        """
+        domainusers = 'domain users'
+        nonposixuser = "nonposixuser@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("hybrid")
+        result = self.clients[0].run_command(["id", nonposixuser])
+        uid_reg = (
+            r"^uid=(\d+)\({0}\)".format(nonposixuser)
+        )
+        gid_reg = (
+            r"gid=(\d+)\({0}\)".format(nonposixuser)
+        )
+        grp_reg = (
+            r"groups=(\d+)\({0}\),(\d+)\({1}@{2}\)".format
+            (nonposixuser, domainusers, self.ad_domain)
+        )
+        assert re.search(uid_reg, result.stdout_text)
+        assert re.search(gid_reg, result.stdout_text)
+        assert re.search(grp_reg, result.stdout_text)
+
+    @pytest.mark.xfail(reason='bz2010512', strict=True)
+    def test_idoverride_with_auto_private_group_false(self):
+        """
+        override ad trusted user in default trust view
+        and set auto-private-groups=false
+        and ensure that overridden values takes effect
+        """
+        domainusers = 'domain users'
+        nonposixuser = "nonposixuser@%s" % self.ad_domain
+        self.master.run_command(
+            [
+                "ipa",
+                "idoverrideuser-add",
+                "Default Trust View",
+                "--uid=99999999",
+                "--gid=98888888",
+                nonposixuser,
+            ]
+        )
+        tasks.clear_sssd_cache(self.master)
+        self.mod_idrange_auto_private_group("false")
+        tasks.clear_sssd_cache(self.master)
+        tasks.clear_sssd_cache(self.clients[0])
+        result = self.clients[0].run_command(['id', nonposixuser])
+        uid_reg = (
+            r"^uid=(99999999)\({0}\)".format(nonposixuser)
+        )
+        gid_reg = r"gid=(98888888)"
+        grp_reg = (
+            r"groups=(99999999)\({0}\),(\d+)\({1}@{2}\)".format
+            (nonposixuser, domainusers, self.ad_domain)
+        )
+        assert re.search(uid_reg, result.stdout_text)
+        assert re.search(gid_reg, result.stdout_text)
+        assert re.search(grp_reg, result.stdout_text)
+
+    @pytest.mark.xfail(reason='bz2010512', strict=True)
+    def test_idoverride_with_auto_private_group_true(self):
+        """
+        override ad trusted user in default trust view
+        and set auto-private-groups=true and ensure that
+        overridden values takes effect
+        """
+        domainusers = 'domain users'
+        nonposixuser = "nonposixuser@%s" % self.ad_domain
+        self.master.run_command(
+            [
+                "ipa",
+                "idoverrideuser-mod",
+                "Default Trust View",
+                "--uid=88888888",
+                "--gid=88888888",
+                nonposixuser,
+            ]
+        )
+        tasks.clear_sssd_cache(self.master)
+        self.mod_idrange_auto_private_group("true")
+        tasks.clear_sssd_cache(self.clients[0])
+        result = self.clients[0].run_command(['id', nonposixuser])
+        uid_reg = (
+            r"^uid=(88888888)\({0}\)".format
+            (nonposixuser)
+        )
+        gid_reg = r"gid=(88888888)"
+        grp_reg = (
+            r"groups=(88888888),(\d+)\({0}@{1}\)".format
+            (domainusers, self.ad_domain)
+        )
+        assert re.search(uid_reg, result.stdout_text)
+        assert re.search(gid_reg, result.stdout_text)
+        assert re.search(grp_reg, result.stdout_text)
+
+    @pytest.mark.xfail(reason='bz2010512', strict=True)
+    def test_idoverride_with_auto_private_group_hybrid(self):
+        """
+        override ad trusted user in default trust view
+        and set auto-private-groups=hybrid and ensure that
+        overridden values takes effect.
+        """
+        domainusers = 'domain users'
+        nonposixuser = "nonposixuser@%s" % self.ad_domain
+        self.master.run_command(
+            [
+                "ipa",
+                "idoverrideuser-mod",
+                "Default Trust View",
+                "--uid=78888888",
+                "--gid=79999999",
+                nonposixuser
+            ]
+        )
+        tasks.clear_sssd_cache(self.master)
+        self.mod_idrange_auto_private_group("hybrid")
+        tasks.clear_sssd_cache(self.master)
+        tasks.clear_sssd_cache(self.clients[0])
+        result = self.clients[0].run_command(['id', nonposixuser])
+        uid_reg = (
+            r"^uid=(78888888)\({0}\)".format
+            (nonposixuser)
+        )
+        gid_reg = r"gid=(79999999)"
+        grp_reg = (
+            r"groups=(78888888),(\d+)\({0}@{1}\)".format
+            (domainusers, self.ad_domain)
+        )
+        assert re.search(uid_reg, result.stdout_text)
+        assert re.search(gid_reg, result.stdout_text)
+        assert re.search(grp_reg, result.stdout_text)
+
+    def test_nonposixuser_nondefault_primary_group(self):
+        """
+        Test for non default primary group
+        """
+        nonposixuser = "nonposixuser1@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("true")
+        tasks.clear_sssd_cache(self.master)
+        self.clients[0].run_command(['id', nonposixuser])
+        # Remove trust
+        self.remove_trust(self.ad)
+
+
+class TestPosixAutoPrivateGroup(BaseTestTrust):
+    """
+    Tests for auto-private-groups option with non posix AD trust
+    """
+    topology = 'line'
+    num_ad_domains = 1
+    num_clients = 1
+
+    def test_add_posix_trust(self):
+        self.add_trust('ipa-ad-trust-posix')
+
+    @pytest.mark.xfail(reason='bz2010512', strict=True)
+    def test_gidnumber_not_coresponding_existing_group(self):
+        """
+        Test checks that AD users which contain posix attributes
+        (uidNumber and gidNumber) but there is no group with the
+        corresponding gidNumber.
+        """
+        posixuser = "testuser2@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("true")
+        tasks.clear_sssd_cache(self.master)
+        (uid, gid) = self.get_user_id(self.clients[0], posixuser)
+        assert uid == gid
+
+    def test_uid_number_auto_private_group_false(self):
+        """
+        Test checks that posix user with only uidNumber defined
+        and gidNumber not set, is not resolvable when auto-private-group
+        is set to False
+        """
+        posixuser = "testuser1@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("false")
+        tasks.clear_sssd_cache(self.master)
+        result = self.clients[0].run_command(['id', posixuser],
+                                             raiseonerr=False)
+        assert 'no such user' in result.stderr_text
+
+    def test_uid_number_auto_private_group_true(self):
+        """
+        Test checks that posix user with only uidNumber defined
+        and gidNumber not set, a private group is created with
+        gidNumber = uidNumber
+        """
+        posixuser = "testuser1@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("true")
+        tasks.clear_sssd_cache(self.master)
+        tasks.clear_sssd_cache(self.clients[0])
+        (uid, gid) = self.get_user_id(self.clients[0], posixuser)
+        assert uid == gid
+
+    def test_uid_number_auto_private_group_hybrid(self):
+        """
+        Test checks that posix user with only uidNumber defined
+        and gidNumber not set, is not resolvable when auto-private-group
+        is set to hybrid
+        """
+        posixuser = "testuser1@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("hybrid")
+        tasks.clear_sssd_cache(self.master)
+        tasks.clear_sssd_cache(self.clients[0])
+        result = self.clients[0].run_command(['id', posixuser],
+                                             raiseonerr=False)
+        assert 'no such user' in result.stderr_text
+        # Remove trust
+        self.remove_trust(self.ad)
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml


ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-latest/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
+        test_suite: test_integration/test_trust.py::TestTrustWithAutoPrivateGroup
         template: *ci-master-latest
         timeout: 3600
-        topology: *master_1repl_1client
+        topology: adroot_adchild_adtree_master_1client
```
flo-renaud, line 74: The topology should be *adroot_adchild_adtree_master_1client (the start at the beginning means it's an alias to an anchor defined earlier, at line 44).

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +992,90 @@ def test_server_option_with_unreachable_ad(self):
...
+        tasks.configure_dns_for_trust(self.master, self.ad)
+        tasks.establish_trust_with_ad(
+            self.master, self.ad_domain,
+            extra_args=['--range-type', 'ipa-ad-trust-posix'])
+
+    def test_idrange_mod_auto_private_group_false(self):
+        range_name = self.ad_domain.upper() + '_id_range'
+        self.master.run_command(['ipa', 'idrange-show', range_name])
+        cmd = self.master.run_command(
+            ['ipa','idrange-mod', range_name, '--auto-private-groups=false'])
```
mrizwan93, line 1050: no space between (`ipa` and `idrange-mode`) >  `['ipa', 'idrange-mod', range_name, '--auto-private-groups=false']`
Wondering why pep8 check didn;t catch!

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +992,90 @@ def test_server_option_with_unreachable_ad(self):
...
+    def test_idrange_mod_auto_private_group_hybrid(self):
+        range_name = self.ad_domain.upper() + '_id_range'
+        self.master.run_command(['ipa', 'idrange-show', range_name])
+        cmd = self.master.run_command(
+            ['ipa','idrange-mod', range_name, '--auto-private-groups=hybrid'])
+        assert 'Auto private groups: hybrid' in cmd.stdout_text
+        self.master.run_command(['systemctl', 'restart', 'sssd.service'])
+
+    def test_idrange_mod_auto_private_group_invalid(self):
+        error_msg = (
```
mrizwan93, line 1071: Add this test to `TestNonPosixADTrustWithAutoPrivateGroup` as well.

Also these tests are redundant in both classes. Could you define  a separate method for these and call in here and above class?

```
def idrange_mod_auto_private_group_hybrid(host, ad_domain):
    range_name = ad_domain.upper() + '_id_range'
    host.run_command(['ipa', 'idrange-show', range_name])
    cmd = host.run_command(
          ['ipa','idrange-mod', range_name, '--auto-private-groups=hybrid'])
    assert 'Auto private groups: hybrid' in cmd.stdout_text
    host.run_command(['systemctl', 'restart', 'sssd.service'])
```
and call this as 
```
classTestPosixADTrustWithAutoPrivateGroup:
    def test_idrange_mod_auto_private_group_hybrid(self):
        idrange_mod_auto_private_group_hybrid(self.master, self.ad_domain)

classTestNonPosixADTrustWithAutoPrivateGroup:
    def test_idrange_mod_auto_private_group_hybrid(self):
        idrange_mod_auto_private_group_hybrid(self.master, self.ad_domain)
```

mrizwan93 (changes requested):
Hello @menonsudhir 
Please find comments inline.

flo-renaud (commented):
Hi @menonsudhir 
The tests added are redundant with those already present in https://pagure.io/freeipa/blob/master/f/ipatests/test_xmlrpc/test_range_plugin.py
Unless they are extended with more checks (for instance the output of id $ADUSER or getent $ADUSER) they don't provide additional value.
If the intent is to start laying a skeleton for future tests, I would recommend to factor as much as possible. The code should avoid copy-paste. For instance defined a class `class BaseTestAutoPrivateGroups` and classes inheriting from it, `class TestNonPosixADTrust(BaseTestAutoPrivateGroups)` and `class TestPosixADTrust(BaseTestAutoPrivateGroup)`.
The parent class can define common methods with parameters, that will be called by the children classes.

flo-renaud:
Removing needs_review as the PR is not ready yet.

menonsudhir:
@fcami @mrizwan93 the tests have passed, but AzurePipeline tests are still failing after multiple reruns.
Could you please review the tests meanwhile.

menonsudhir:
TestAdvice.test_advice_config_client_for_smart_card_auth fails due to
[2021-07-26 06:17:28] E           subprocess.CalledProcessError: Command '['sh', '/tmp/tmp.uFa3zV4rMu', '/tmp/tmp.jfj2905xEE']' returned non-zero exit status 1.
Could not install OpenSC package

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +992,270 @@ def test_server_option_with_unreachable_ad(self):
...
+        elif trust_type == "ipa-ad-trust-posix":
+            cmd.extend(("--type", "ipa-ad-trust-posix"))
+
+        if auto_priv_grp_val == "false":
+            cmd.extend(("--auto-private-groups", "false"))
+        elif auto_priv_grp_val == "true":
+            cmd.extend(("--auto-private-groups", "true"))
+        else:
+            cmd.extend(("--auto-private-groups", " hybrid"))
+
```
amore17, line 1058: Remove if elif else condition.
Instead of that directly add  
`cmd.extend(("--auto-private-groups", auto_priv_grp_val))`
Also define default value for `auto_priv_grp_val`

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +992,270 @@ def test_server_option_with_unreachable_ad(self):
...
+            "--dom-name",
+            domainname,
+        ]
+
+        if trust_type == "ipa-adtrust":
+            cmd.extend(("--type", "ipa-ad-trust"))
+        elif trust_type == "ipa-ad-trust-posix":
+            cmd.extend(("--type", "ipa-ad-trust-posix"))
+
+        if auto_priv_grp_val == "false":
```
amore17, line 1052: Remove if elif else conditions.
Instead of that directly add  
`cmd.extend(("--auto-private-groups", auto_priv_grp_val))`
Also define default value for `auto_priv_grp_val`

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +992,270 @@ def test_server_option_with_unreachable_ad(self):
...
+            newbaseid,
+            "--range-size",
+            range,
+            "--rid-base",
+            newbaseid,
+            "--dom-name",
+            domainname,
+        ]
+
+        if trust_type == "ipa-adtrust":
```
amore17, line 1047: Same here as well remove conditions.
Define default value.

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +992,270 @@ def test_server_option_with_unreachable_ad(self):
...
+            cmd.extend(("--type", "ipa-ad-trust-posix"))
+
+        if auto_priv_grp_val == "false":
+            cmd.extend(("--auto-private-groups", "false"))
+        elif auto_priv_grp_val == "true":
+            cmd.extend(("--auto-private-groups", "true"))
+        else:
+            cmd.extend(("--auto-private-groups", " hybrid"))
+
+        self.master.run_command(cmd, raiseonerr=False)
```
amore17, line 1059: Remove raiseonerr=False or if you are expecting it to fail please add comment
with reason.

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +992,270 @@ def test_server_option_with_unreachable_ad(self):
...
+
+    def mod_idrange_auto_private_group(self, rangename, auto_priv_grp_val):
+        """
+        Function to modify existing idrange
+        """
+        valid_msg = "Auto private groups: "
+        error_msg = "ipa: ERROR: no modifications to be performed"
+        rangename = self.ad_domain.upper() + "_id_range"
+        cmd = ["ipa", "idrange-mod", rangename]
+        if auto_priv_grp_val == "false":
```
amore17, line 1070: Same here remove conditions,
define default value for auto_priv_grp_val and accept it from that.
cmd.extend(("--auto-private-groups", auto_priv_grp_val))

amore17 (changes requested):

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +992,264 @@ def test_server_option_with_unreachable_ad(self):
...
+        gid_reg = (
+            r"gid=(\d+)\({0}\)".format(nonposixuser)
+        )
+        grp_reg = (
+            r"groups=(\d+)\({0}\),(\d+)\({1}@{2}\)".format
+            (nonposixuser, domainusers, self.ad_domain)
+        )
+        assert re.search(uid_reg, result.stdout_text)
+        assert re.search(gid_reg, result.stdout_text)
+        assert re.search(grp_reg, result.stdout_text)
```
flo-renaud, line 1109: I fail to understand the rationale for the 3 tests test_modify_auto_private_groups_value_default_trusted_range, test_modify_auto_private_groups_value_trusted_range_hybrid and test_idoverride_with_auto_private_group_false.
They have a different auto_private_group value but expect exactly the same thing (the output of `id username` should contain uid=xxx gid=xxx and groups=xxx).

You need to tailor each test and make clear that the expected value is different.

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +992,264 @@ def test_server_option_with_unreachable_ad(self):
...
+
+    def test_idoverride_with_auto_private_group_hybrid(self):
+        """
+        override ad trusted user in default trust view
+        and set auto-private-groups=hybrid to check the
+        output displayed
+        """
```
flo-renaud, line 1207: Please state in the comment what is the expected behavior (should the override win over the uid/gid values?)

flo-renaud (commented):
Hi @menonsudhir 
The tests still need a good amount of work, please find inline comments.
- Some scenario are already covered in XMLRPC tests and don't need to be written here.
- posix range type tests are missing
- the important thing to check is the actual values returned for uid/gid and groups. The tests need to look for specific values, not simply regexp like "uid must contain numbers".

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +1052,242 @@ def test_server_option_with_unreachable_ad(self):
...
+class TestNonPosixAutoPrivateGroup(BaseTestTrust):
+    """
+    Tests for auto-private-groups option with non posix AD trust
+    """
+    topology = 'line'
+    num_ad_domains = 1
+
+    @classmethod
+    def install(cls, mh):
+        if not cls.master.transport.file_exists('/usr/bin/rpcclient'):
```
flo-renaud, line 1066: Why is the install method overriden instead of re-using the install method from BaseTestTrust? IMO this method definition can be removed. You can move the line `cls.existing_rangename = cls.ad_domain.upper() + "_id_range"` in the BaseTestTrust install method.

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +1052,242 @@ def test_server_option_with_unreachable_ad(self):
...
+
+    """
+    Tests for auto-private-groups option with non posix AD trust
+    """
+    topology = 'line'
+    num_ad_domains = 1
+
+    @classmethod
+    def install(cls, mh):
+        if not cls.master.transport.file_exists('/usr/bin/rpcclient'):
```
flo-renaud, line 1243: Why is the install method overriden instead of re-using the install method from BaseTestTrust? IMO this method definition can be removed.

ipatests/test_integration/test_trust.py
```diff
@@ -106,6 +107,65 @@ def check_range_properties(self, realm, expected_type, expected_size):
         expected_text = 'iparangetype: %s\n' % expected_type
         assert expected_text in result.stdout_text
 
+    def add_idrange_auto_private_group(
```
flo-renaud, line 110: Is the function used anywhere? I believe it can be removed.

ipatests/test_integration/test_trust.py
```diff
@@ -106,6 +107,65 @@ def check_range_properties(self, realm, expected_type, expected_size):
...
+        ]
+
+        cmd.extend(("--type", trust_type))
+        cmd.extend(("--auto-private-groups", auto_priv_grp_val))
+
+        self.master.run_command(cmd)
+        self.master.run_command(["systemctl", "restart", "sssd"])
+
+    def mod_idrange_auto_private_group(
+        self, rangename, auto_priv_grp_val='false'
```
flo-renaud, line 153: The method `mod_idrange_auto_private_group` is always called with `rangename = self.existing_rangename`. You can remove this parameter from the method signature and replace `rangename` inside the method with `self.existing_rangename`:
```
    def mod_idrange_auto_private_group(
        self, auto_priv_grp_val='false'):
```

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +1052,242 @@ def test_server_option_with_unreachable_ad(self):
...
+            raise pytest.skip("Package samba-client not available "
+                              "on {}".format(cls.master.hostname))
+        super(TestNonPosixAutoPrivateGroup, cls).install(mh)
+        cls.ad = cls.ads[0]
+        cls.ad_domain = cls.ad.domain.name
+        tasks.install_adtrust(cls.master)
+        tasks.sync_time(cls.master, cls.ad)
+        cls.existing_rangename = cls.ad_domain.upper() + "_id_range"
+
+    def test_add_nonposix_trust(self):
```
flo-renaud, line 1076: You can factor code and define the following in BaseTestTrust:
```
def add_trust(self, rangetype):
    tasks.configure_dns_for_trust(self.master, self.ad)
    tasks.establish_trust_with_ad(self.master, self.ad_domain, extra_args=['--range-type', rangetype])
```
Then in this class:
```
def test_add_nonposix_trust(self):
    self.add_trust('ipa-ad-trust')
```
and in TestPosixAUtoPrivateGroup:
```
def test_add_posix_trust(self):
    self.add_trust('ipa-ad-trust-posix')
```

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +1052,242 @@ def test_server_option_with_unreachable_ad(self):
...
+        gid_reg = (
+            r"gid=(\d+)\({0}\)".format(nonposixuser)
+        )
+        grp_reg = (
+            r"groups=(\d+)\({0}\),(\d+)\({1}@{2}\)".format
+            (nonposixuser, domainusers, self.ad_domain)
+        )
+        assert re.search(uid_reg, result.stdout_text)
+        assert re.search(gid_reg, result.stdout_text)
+        assert re.search(grp_reg, result.stdout_text)
```
flo-renaud, line 1103: In this test, you check the output for a user which has an existing gid defined in AD. It would be interesting to check with a user which doesn't have any gid defined in AD or for a user which has a gid set but no corresponding group in AD. The same comment applies to all the possible settings for auto-private-group.
It means that you need to modify the ansible playbook that is creating the AD users (see https://github.com/freeipa/freeipa-pr-ci/tree/master/ansible/roles/windows/ipa-ad-data) and also create a user without gid, and a user with gid but no corresponding group.
Same thing for non-posix users: a user with a primaryGroupID that is not the default one.

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +1052,242 @@ def test_server_option_with_unreachable_ad(self):
...
+
+    def test_idoverride_with_auto_private_group_false(self):
```
flo-renaud, line 1128: This test is 99% identical to class TestPosixAutoPrivateGroup::test_idoverride_with_auto_private_group_false.
Please refactor (only the user differs between the 2 tests).

flo-renaud (commented):
Hi @menonsudhir 
please find inline comments. They are mostly related to code factorization and avoiding duplicate code when possible.

ipatests/test_integration/test_trust.py
```diff
@@ -106,6 +107,36 @@ def check_range_properties(self, realm, expected_type, expected_size):
...
+        cmd = ["ipa", "idrange-mod", rangename,
+               "--auto-private-groups", auto_priv_grp_val]
+        result = self.master.run_command(cmd, raiseonerr=False)
+        try:
+            self.master.run_command(cmd)
+            self.master.run_command(["systemctl", "restart", "sssd"])
+        except CalledProcessError as e:
+            # ignore error if the auto-private-group is already set
+            # to the expected value
+        if error_msg not in e.stderr:
```
flo-renaud, line 129: This line and the following need to have +4chars indentation in order to be executed in the except clause.

ipatests/test_integration/test_trust.py
```diff
@@ -106,6 +108,47 @@ def check_range_properties(self, realm, expected_type, expected_size):
...
+        rangename = self.existing_rangename
+        valid_msg = "Auto private groups: "
+        error_msg = "ipa: ERROR: no modifications to be performed"
+        cmd = ["ipa", "idrange-mod", rangename,
+               "--auto-private-groups", auto_priv_grp_val]
+        result = self.master.run_command(cmd, raiseonerr=False)
+        try:
+            self.master.run_command(cmd)
+            self.master.run_command(["systemctl", "restart", "sssd"])
+        except CalledProcessError as e:
```
flo-renaud, line 127: Because ` import subprocess` is used, the exception should be `subprocess.CalledProcessError`.

ipatests/test_integration/test_trust.py
```diff
@@ -106,6 +108,47 @@ def check_range_properties(self, realm, expected_type, expected_size):
...
+        """
+        Set the auto-private-group option of the default trusted
+        AD domain range and restart sssd
+        """
+        rangename = self.existing_rangename
+        valid_msg = "Auto private groups: "
+        error_msg = "ipa: ERROR: no modifications to be performed"
+        cmd = ["ipa", "idrange-mod", rangename,
+               "--auto-private-groups", auto_priv_grp_val]
+        result = self.master.run_command(cmd, raiseonerr=False)
```
flo-renaud, line 123: This call is duplicated in line 125, please remove.

ipatests/test_integration/test_trust.py
```diff
@@ -106,6 +108,47 @@ def check_range_properties(self, realm, expected_type, expected_size):
...
 
+    def mod_idrange_auto_private_group(
+        self, auto_priv_grp_val='false'
+    ):
+        """
+        Set the auto-private-group option of the default trusted
+        AD domain range and restart sssd
+        """
+        rangename = self.existing_rangename
+        valid_msg = "Auto private groups: "
```
flo-renaud, line 119: The variable `valid_msg` is not used, please remove

flo-renaud (commented):
@menonsudhir 
Please find inline comments

ipatests/test_integration/test_trust.py
flo-renaud: I meant to remove this line and keep the one in try/except.

ipatests/test_integration/test_trust.py
```diff
@@ -106,6 +108,45 @@ def check_range_properties(self, realm, expected_type, expected_size):
...
+                raise
+
+    def get_user_id(host, username):
+        """
+        User uid and gid is parsed from the output
+        of id command and a tuple is returned with (uid,gid)
+        """
+        id_output = host.run_command(["id", username])
+        match = re.match("^uid=(?P<uid>\d+).*gid=(?P<gid>\d+)", id_output)
+        uid = match.group(uid)
```
flo-renaud, line 138: This should be instead:
```
uid = match.group('uid')
gid = match.group('gid')
```
Please see https://docs.python.org/3/howto/regex.html#non-capturing-and-named-groups

ipatests/test_integration/test_trust.py
```diff
@@ -106,6 +108,45 @@ def check_range_properties(self, realm, expected_type, expected_size):
...
+        result = self.master.run_command(cmd, raiseonerr=False)
+        try:
+            self.master.run_command(["systemctl", "restart", "sssd"])
+        except subprocess.CalledProcessError as e:
+            # ignore error if the auto-private-group is already set
+            # to the expected value
+            if error_msg not in e.stderr:
+                raise
+
+    def get_user_id(host, username):
```
flo-renaud, line 131: Either the method is defined in this class as a static method, and it needs to be 
```
@staticmethod
def get_user_id(host, username):
```
or it is a function defined outside of any class, and it can be `def get_user_id(host, username):`.

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +1033,239 @@ def test_server_option_with_unreachable_ad(self):
...
+    def test_uid_number_auto_private_group_true(self):
+        """
+        Test checks that posix user with only uidNumber defined
+        and gidNumber not set, a private group is created with
+        gidNumber = uidNumber
+        """
+        posixuser = "testuser1@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("true")
+        tasks.clear_sssd_cache(self.master)
+        result = self.get_user_id(self.clients[0], posixuser)
```
flo-renaud, line 1255: This should be `(uid, gid) = self.get_user_id(self.clients[0], posixuser)`

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +1033,239 @@ def test_server_option_with_unreachable_ad(self):
...
+    def test_gidnumber_not_coresponding_existing_group(self):
+        """
+        Test checks that AD users which contain posix attributes
+        (uidNumber and gidNumber) but there is no group with the
+        corresponding gidNumber.
+        """
+        posixuser = "testuser2@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("true")
+        tasks.clear_sssd_cache(self.master)
+        result = self.get_user_id(self.clients[0], posixuser)
```
flo-renaud, line 1228: This should be `(uid, gid) = self.get_user_id(self.clients[0], posixuser)`

flo-renaud (commented):
@menonsudhir 
please find inline comments

ipatests/test_integration/test_trust.py
flo-renaud: get_user_id is returning a tuple, please use `(uid, gid) = self.get_user_id(self.clients[0], posixuser)`

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +1033,239 @@ def test_server_option_with_unreachable_ad(self):
...
+    def test_uid_number_auto_private_group_hybrid(self):
+        """
+        Test checks that posix user with only uidNumber defined
+        and gidNumber not set, is not resolvable when auto-private-group
+        is set to hybrid
+        """
+        posixuser = "testuser1@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("true")
+        tasks.clear_sssd_cache(self.master)
+        (uid, gid) = self.clients[0].run_command(['id', posixuser],
```
flo-renaud, line 1267: run_command does not return a tuple, please use `result = self.clients[0].run_command(...`

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +1033,239 @@ def test_server_option_with_unreachable_ad(self):
...
+        """
+        Test checks that posix user with only uidNumber defined
+        and gidNumber not set, a private group is created with
+        gidNumber = uidNumber
+        """
+        posixuser = "testuser1@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("true")
+        tasks.clear_sssd_cache(self.master)
+        (uid, gid) = self.get_user_id(self.clients[0], posixuser)
+        assert result[uid] == result[gid]
```
flo-renaud, line 1256: This should be instead `assert uid == gid`

flo-renaud (commented):
Please find inline comments

flo-renaud (commented):
Hi @menonsudhir 
Please find inline comments.

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +1033,237 @@ def test_server_option_with_unreachable_ad(self):
...
+            r"^uid=(99999999)\({0}\)".format(nonposixuser)
+        )
+        gid_reg = r"gid=(98888888)"
+        grp_reg = (
+            r"groups=(99999999)\({0}\),(\d+)\({1}@{2}\)".format
+            (nonposixuser, domainusers, self.ad_domain)
+        )
+        assert re.search(uid_reg, result.stdout_text)
+        assert re.search(gid_reg, result.stdout_text)
+        assert re.search(grp_reg, result.stdout_text)
```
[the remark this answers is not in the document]
menonsudhir: @flo-renaud i did try these tests manually and actually the output of id command is not displayed on the ipaclient.
However have added the sssd clear cache statement in the tests, waiting for the execution result.

ipatests/test_integration/test_trust.py
```diff
@@ -106,6 +108,47 @@ def check_range_properties(self, realm, expected_type, expected_size):
...
+            # to the expected value
+            if error_msg not in e.stderr:
+                raise
+
+    def get_user_id(self, host, username):
+        """
+        User uid and gid is parsed from the output
+        of id command and a tuple is returned with (uid,gid)
+        """
+        id_output.stdout_text = host.run_command(["id", username])
```
flo-renaud, line 136: This line should read
`id_output = host.run_command(["id", username])`

ipatests/test_integration/test_trust.py
flo-renaud: Please address my previous comment: the test should ensure that gid=uid. You can reuse the method get_user_id here.

ipatests/test_integration/test_trust.py
```diff
@@ -993,3 +1036,248 @@ def test_server_option_with_unreachable_ad(self):
...
+            r"gid=(\d+)\({0}\)".format(nonposixuser)
+        )
+        grp_reg = (
+            r"groups=(\d+)\({0}\),(\d+)\({1}@{2}\)".format
+            (nonposixuser, domainusers, self.ad_domain)
+        )
+        assert re.search(uid_reg, result.stdout_text)
+        assert re.search(gid_reg, result.stdout_text)
+        assert re.search(grp_reg, result.stdout_text)
+
```
flo-renaud, line 1074: The test is also missing a test case: with nonposixuser1 (a non posix user with a different primaryGroup), ensure that the user is returned with uid different from gid (the primaryGroup is used for gid).

ipatests/test_integration/test_trust.py
```diff
@@ -992,3 +1019,244 @@ def test_server_option_with_unreachable_ad(self):
...
+        gid_reg = (
+            r"gid=(\d+)\({0}\)".format(nonposixuser)
+        )
+        grp_reg = (
+            r"groups=(\d+)\({0}\),(\d+)\({1}@{2}\)".format
+            (nonposixuser, domainusers, self.ad_domain)
+        )
+        assert re.search(uid_reg, result.stdout_text)
+        assert re.search(gid_reg, result.stdout_text)
+        assert re.search(grp_reg, result.stdout_text)
```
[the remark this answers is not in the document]
flo-renaud: Please address my previous comment. The test needs to be run with both nonposixuser and nonposixuser1, and ensure that in each case, the returned gid = the value from primaryGroup.

We are also missing the same type of test, this time with auto private groups set to false. Same behavior as for hybrid.

ipatests/test_integration/test_trust.py
```diff
@@ -993,3 +1036,248 @@ def test_server_option_with_unreachable_ad(self):
...
+    def test_nonposixuser_nondefault_primary_group(self):
+        """
+        Test for non default primary group
+        """
+        nonposixuser = "nonposixuser1@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("true")
+        tasks.clear_sssd_cache(self.master)
+        self.clients[0].run_command(['id', nonposixuser])
+        # Remove trust
+        self.remove_trust(self.ad)
```
flo-renaud, line 1216: This test can be deleted and incorporated in the previous ones (we need a different expectation for each auto private group value).

ipatests/test_integration/test_trust.py
```diff
@@ -993,3 +1036,248 @@ def test_server_option_with_unreachable_ad(self):
...
+    Tests for auto-private-groups option with non posix AD trust
+    """
+    topology = 'line'
+    num_ad_domains = 1
+    num_clients = 1
+
+    def test_add_posix_trust(self):
+        self.add_trust('ipa-ad-trust-posix')
+
+    @pytest.mark.xfail(reason='bz2010512', strict=True)
```
flo-renaud, line 1230: The default setting for posix trust is "false", hence this test should not be able to resolve the user. Don't mark it as xfail but rather expect an exception "no such user"

ipatests/test_integration/test_trust.py
```diff
@@ -993,3 +1036,248 @@ def test_server_option_with_unreachable_ad(self):
...
+        """
+        Test checks that AD users which contain posix attributes
+        (uidNumber and gidNumber) but there is no group with the
+        corresponding gidNumber.
+        """
+        posixuser = "testuser2@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("true")
+        tasks.clear_sssd_cache(self.master)
+        (uid, gid) = self.get_user_id(self.clients[0], posixuser)
+        assert uid == gid
```
flo-renaud, line 1241: The same test with testuser2 should be added with the other values of auto private group and have different expectations.

ipatests/test_integration/test_trust.py
```diff
@@ -993,3 +1036,248 @@ def test_server_option_with_unreachable_ad(self):
...
+        """
+        posixuser = "testuser1@%s" % self.ad_domain
+        self.mod_idrange_auto_private_group("hybrid")
+        tasks.clear_sssd_cache(self.master)
+        tasks.clear_sssd_cache(self.clients[0])
+        result = self.clients[0].run_command(['id', posixuser],
+                                             raiseonerr=False)
+        assert 'no such user' in result.stderr_text
+        # Remove trust
+        self.remove_trust(self.ad)
```
flo-renaud, line 1283: We are missing tests with a posix user with uid, gid combined with the 3 possible values of auto private groups.

flo-renaud (commented):
@menonsudhir 
please find additional comments. Some test cases described in the test plan are still missing

amore17:
Closing this PR with https://github.com/freeipa/freeipa/pull/6150 is in progress.

PR #5767, opened 2021-05-12
ipatests: kinit on server for test_proxycommand_invalid_shell

rcritten:
We've seen some intermittent failures of this test with warnings
about passwords about to expire. We suspect there may be a time
difference between the client and server so set the passwords
on the server instead to be sure that time is correct.

https://pagure.io/freeipa/issue/8785

### DIFF
--- a/ipatests/test_integration/test_commands.py
+++ b/ipatests/test_integration/test_commands.py
@@ -1426,10 +1426,10 @@ class TestIPACommand(IntegrationTest):
 
         user_kinit = "{password}\n{password}\n{password}\n".format(
             password=password)
-        self.clients[0].run_command([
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @rcritten 
thanks for the patch, LGTM. You can remove the temp commit and I will ACK when PRCI completes.

rcritten:
Rebased and dropped temp commit.

flo-renaud:
Temp commit removed and tests are green, ACK

flo-renaud:
master:

* 4f301306974209086a6800c736cd57b8017db1ad ipatests: kinit on server for test_proxycommand_invalid_shell

PR #5769, opened 2021-05-13
WebUI: Add support of 'ipaautoprivategroups' LDAP attribute on 'ID Ranges' page

serg-cymbaluk:
Add 'Auto private groups' field on 'Add ID range' form with the following options: true, false, hybrid.
The field is optional and can be omitted.
Its value can be also modified on 'Range Settings' page after the range is added.
    
Ticket: https://pagure.io/freeipa/issue/8837

### DIFF
--- a/install/ui/src/freeipa/idrange.js
+++ b/install/ui/src/freeipa/idrange.js
@@ -82,6 +82,13 @@ return {
                             name: 'ipanttrusteddomainsid',
                             label: '@i18n:objects.idrange.ipanttrusteddomainsid',
                             title: '@mo-param:idrange:ipanttrusteddomainsid:label'
+                        },
+                        {
+                            name: 'ipaautoprivategroups',
+                            $type: 'select',
+                            label: '@i18n:objects.idrange.ipaautoprivategroups',
+                            title: '@mo-param:idrange:ipaautoprivategroups:label',
+                            options: IPA.create_options(['', 'true', 'false', 'hybrid'])
                         }
                     ]
                 }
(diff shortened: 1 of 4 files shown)


rcritten:
LGTM but I'm not really well-versed in the UI. It does seem to perform the same way as the cli.

serg-cymbaluk:
@rcritten
Thanks for your review.
Can I merge the changes?

flo-renaud:
@serg-cymbaluk 
I would like to test with an AD trust before ACK'ing the PR, but the infra is not very cooperative today. I'll retry tomorrow and let you know.

netoarmando:
I tried to test this on my lab, after setting "Auto private groups" as 'true' for my 'AD.TEST_id_range' I got an alert saying:

> Service sssd.service requires restart on IPA server AD.TEST_id_range to apply configuration changes.

abbra:
@netoarmando that's correct, we send a warning from the server side.

netoarmando:
> @netoarmando that's correct, we send a warning from the server side.

So, I guess this feature is working fine from WebUI/Server perspective. @flo-renaud was about to test with it with an AD trust, after https://github.com/SSSD/sssd/pull/5613 I think it's possible..

flo-renaud:
Hi @serg-cymbaluk 
thanks for the patch, works for me. I was finally able to test thanks to an AD instance borrowed from @menonsudhir.
Checked that:
- [x] webui allows to set auto private range for AD idranges 
- [x] webui doesn't allow to set auto-private range for Local Domain range
- [x] only supported values are proposed in the drop down list
- [x] webui displays the correct value, compared to what is stored in LDAP

flo-renaud:
master:

* 6395d239476223a42b5add4d12979aed1439d46d WebUI: Add support of 'ipaautoprivategroups' LDAP attribute on 'ID Ranges' page
* d8b8f54bceb585b5c9b14c1d03e77a4c2cad3ae9 WebUI tests: Add test for 'ipaautoprivategroups' field on 'ID Ranges' page

PR #5780, opened 2021-05-19
ipatests: dnssec: Add alternative approach for checking chain of trust

stanislavlevin:
drill is currently broken on F34. Fortunately, there are another tools for checking DNSSEC trust. One of them is `delv`:
    
    delv is a tool for sending DNS queries and validating the results,
    using the same internal resolver and validator logic as named.
    
    delv sends to a specified name server all queries needed to fetch and
    validate the requested data; this includes the original requested query,
    subsequent queries to follow CNAME or DNAME chains, queries for DNSKEY,
    and DS records to establish a chain of trust for DNSSEC validation. It
    does not perform iterative resolution, but simulates the behavior of a
    name server configured for DNSSEC validating and forwarding.
    

- mark as XFAIL `test_chain_of_trust_drill` on F34
- conditionally skip `test_chain_of_trust_delv` (requires delv > 9.15.4 for `+yaml`)

Related: https://pagure.io/freeipa/issue/8793

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -878,7 +878,12 @@ Requires: python3-pexpect
 %if 0%{?fedora}
 # These packages do not exist on RHEL and for ipatests use
 # they are installed on the controller through other means
+# https://pagure.io/freeipa/issue/8793
+%if 0%{?fedora} >= 34
+Requires: ldns-utils >= 1.7.1-4
+%else
 Requires: ldns-utils
+%endif
 # update-crypto-policies
 Requires: crypto-policies-scripts
 Requires: python3-polib
--- a/ipatests/prci_definitions/temp_commit.yaml
+++ b/ipatests/prci_definitions/temp_commit.yaml
@@ -47,7 +47,7 @@ topologies:
     memory: 14500
 
 jobs:
-  fedora-latest/build:
+  fedora-f33/build:
     requires: []
     priority: 100
     job:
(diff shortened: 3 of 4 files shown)


hcoin:
While you're at changing dnssec testing, you will catch further bugs (including likely named crashing) if your tests enable dnssec on several domains at the same time, then test them all.  It's not enough to just turn on dnssec on one domain and test it then call it 'ok'.

stanislavlevin:
I'm sorry if I break your installations. Also I'm aware about libp11 issues and follow their bugtracker and commits.
Thank you for your suggestions. Can you please provide the exact test case?

hcoin:
FC34 workstation, no extra packages beyond the min necessary for freeipa server with dnssec.  
1) With DNSSEC off, create several domains, including at least one that's a subdomain of another.  
2) In a tight loop, turn on dnssec on half the domains.
3) run your favorite test code against named, looking for ds, ns, A and AAAA records.
4) Turn on dnssec for all the domains in a tight loop.

What you will notice:

You'll notice logs filled with named not having permission to access files dnssec creates with permissions enough to make them available to named.    Named/ bind9 will error out the calls though the correct info in in the files.
At some random future point, it is possible the 'permission correcter' aspect of dnssec will jump in.  It will correct some, but not all of the permissions.  It often finishes before all the dnssec files are created.  It sometimes fires up again to catch them, sometimes it never does.

And, unless you include complete thread safety for eng_back.c in lib p11, and not the partial stuff in the patches made by the fellow tasked primarily on other matters entirely, named/bind9 will crash within minutes 9 runs out of 10, and within the hour 10 runs out of 10.  With the thread / context safety routines added, named will not crash (but owing to the dnssec issues will report errors arising from permission problems.

stanislavlevin:
I see, thank you.
This is out of the scope of this PR, but should be tracked.

flo-renaud:
>  This is out of the scope of this PR, but should be tracked.

This is already tracked with https://pagure.io/freeipa/issue/8835. We have gathered the DNSSEC-related issues with a common [milestone](https://pagure.io/freeipa/issues?status=Open&milestone=DNSSEC&close_status=), planning to work on them globally.

stanislavlevin:
The `ldns` issue was fixed with `ldns-1.7.1-4.fc34`, current PR-CI's `ldns-1.7.1-3`.
Works correctly in: https://dev.azure.com/freeipa/freeipa/_build/results?buildId=8940&view=logs&j=a2ec7775-2a99-56c3-7189-a6eea5eabd73&t=be06ddc1-a7cb-5d8c-a47c-1176bedd1de8&l=1813

stanislavlevin:
I suggest to apply `delv` for checking the trust (`drill` checks signature chase, although it also can check trust).

stanislavlevin:
Closed in favour of #5790.

PR #5783, opened 2021-05-20
Add argument for --entry option in ipa-managed-entries.1

miztake:
There are no arguments in the --entry option,
but DN for the managed entry definition must actually be specified.
Therefore, add MANAGED_ENTRY as an argument.


rcritten:
Looks good, thanks for the patch.

rcritten:
master:

* 5b87f587ea74af9943c3527f55d94953d8e65363 Add argument for --entry option in ipa-managed-entries.1

PR #5784, opened 2021-05-20
Revert "Only request VALID certs when revoking certs for a host/service"

rcritten:
This reverts commit aa1350384ad6a7d6b2f6056d99fbb43c5c5a6be7.

The search for certificates is a complex, three-step process,
which filters results in subsequent searches. This filters out
non-relevant certificates when deleting a host or service.

This patch breaks that so deleting one service of a host will
revoke *all* certificates for that host.

Another attempt will be made separately to implement this.

Signed-off-by: Rob Crittenden rcritten@redhat.com

### DIFF
--- a/ipaserver/plugins/cert.py
+++ b/ipaserver/plugins/cert.py
@@ -1687,6 +1687,8 @@ class cert_find(Search, CertMethod):
                 ra_options['subject'] = hosts[0]
             elif len(users) == 1 and not services and not hosts:
                 ra_options['subject'] = users[0]
+        if 'status' in options:
+            ra_options['status'] = options.get('status')
 
         try:
             ca_enabled_check(self.api)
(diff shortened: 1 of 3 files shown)


rcritten:
This has only been merged into master.

flo-renaud:
@rcritten thanks for the PR, works for me.
The commit 09426f8 (Add ability to search on certificate revocation status) is not reverted and currently only present in master branch, do you want to keep it that way? I tested it manually and it's stable, we can push it to ipa-4-9 if you find it useful.

rcritten:
I think that https://github.com/freeipa/freeipa/commit/09426f8ed5ebc27d01ecbfd808fabdc0c4430854 is safe as it only adds the API option but doesn't change the way certificates are searched for internally.

I'll create a PR for the ipa-4-9 backport.

flo-renaud:
Ok, wfm

rcritten:
I created PR https://github.com/freeipa/freeipa/pull/5787 for that

rcritten:
master:

* c5e827401045dbecabcbe7b800add66a3729c80d Revert "Only request VALID certs when revoking certs for a host/service"

PR #5788, opened 2021-05-20
WIP: Bronze Bit KDB fixes

frozencemetery:
A WIP PR for this branch as requested.  It is expected to not pass CI, which is for two reasons.  First, it needs the [krb5 branch](https://github.com/krb5/krb5/pull/1159).  Second, because the krb5 branch isn't all tested good - RBCD doesn't work right, and I'm still trying to get testing against local AD working.  (If anyone is up for providing feedback on that, it would be appreciated.)  And if I'm being honest, this code is probably buggy too, so that's three reasons.


abbra:
@frozencemetery I cannot re-open this branch with github tools. Could you please submit a new PR?

PR #5799, opened 2021-05-31
ipatests: test_installation: move tracking_reqs dependency to ipalib constants

miskopo:
ipatests: test_installation: move tracking_reqs dependency to ipalib
constants
ipaserver: krainstance: utilize moved tracking_reqs dependency
    
KRA instance import depends on lib389 package, which is not always
installed and that results in failure. Furthermore, test_installation
utilizes krainstance import. This fix moves relevant parts from
krainstance to ipalib constants where those are subsequently imported
from.
    
Related: https://pagure.io/freeipa/issue/8795

### DIFF
--- a/ipalib/constants.py
+++ b/ipalib/constants.py
@@ -360,3 +360,11 @@ SUBID_RANGE_MAX = (2 ** 32) - (2 * SUBID_COUNT)
 SUBID_RANGE_SIZE = SUBID_RANGE_MAX - SUBID_RANGE_START
 # threshold before DNA plugin requests a new range
 SUBID_DNA_THRESHOLD = 500
+
+# moved from ipaserver/install/krainstance.py::KRAInstance to avoid duplication
+# as per https://pagure.io/freeipa/issue/8795
+KRA_TRACKING_REQS = {
+    'auditSigningCert cert-pki-kra': 'caAuditSigningCert',
+    'transportCert cert-pki-kra': 'caTransportCert',
+    'storageCert cert-pki-kra': 'caStorageCert',
+}
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -20,7 +20,7 @@ from cryptography.hazmat.primitives import hashes
 from cryptography import x509 as crypto_x509
 
 from ipalib import x509
-from ipalib.constants import DOMAIN_LEVEL_0
+from ipalib.constants import DOMAIN_LEVEL_0, KRA_TRACKING_REQS
 from ipalib.constants import IPA_CA_RECORD
 from ipalib.sysrestore import SYSRESTORE_STATEFILE, SYSRESTORE_INDEXFILE
 from ipapython.dn import DN
@@ -34,7 +34,7 @@ from ipatests.pytest_ipa.integration.env_config import get_global_config
 from ipatests.test_integration.base import IntegrationTest
 from ipatests.test_integration.test_caless import CALessBase, ipa_certs_cleanup
 from ipaplatform import services
-from ipaserver.install import krainstance
+
 
 config = get_global_config()
 
@@ -1282,8 +1282,7 @@ class TestInstallMasterKRA(IntegrationTest):
         """
         Test that the KRA subsystem certificates renew properly
         """
-        kra = krainstance.KRAInstance(self.master.domain.realm)
-        for nickname in kra.tracking_reqs:
+        for nickname in KRA_TRACKING_REQS:
             cert = tasks.certutil_fetch_cert(
                 self.master,
                 paths.PKI_TOMCAT_ALIAS_DIR,
--- a/ipaserver/install/krainstance.py
+++ b/ipaserver/install/krainstance.py
@@ -27,6 +27,7 @@ import base64
 
 from ipalib import api
 from ipalib import x509
+from ipalib.constants import KRA_TRACKING_REQS
 from ipaplatform.paths import paths
 from ipapython import directivesetter
 from ipapython import ipautil
@@ -64,11 +65,7 @@ class KRAInstance(DogtagInstance):
     # Mapping of nicknames for tracking requests, and the profile to
     # use for that certificate.  'configure_renewal()' reads this
     # dict.  The profile MUST be specified.
-    tracking_reqs = {
-        'auditSigningCert cert-pki-kra': 'caAuditSigningCert',
-        'transportCert cert-pki-kra': 'caTransportCert',
-        'storageCert cert-pki-kra': 'caStorageCert',
-    }
+    tracking_reqs = KRA_TRACKING_REQS
 
     def __init__(self, realm):
         super(KRAInstance, self).__init__(


miskopo:
Upon extensive dependency resolution attempt I've found out, that in order to remove external dependency on `lib389` and move relevant parts of the code into common test accessible place, I would have to copy-paste and re-resolve about 24 separate files. These files, however, would have to be manually kept in sync with `lib389` package, which would again create lot of work. Thus,  I've decided to skip **one** test that truly depends on `lib389` in case that cannot be imported.

Regarding the `ipaserver` dependency removal part, I am not fully on board as to why. See discussion on [pagure](https://pagure.io/freeipa/issue/8795)

miskopo:
`lint` is failing on unrelated code sources

```
************* Module ipa-custodia-check
install/tools/ipa-custodia-check:187: [E0702(raising-bad-type), IPACustodiaTester._check_jwk_single] Raising NoneType while only classes or instances are allowed)
install/tools/ipa-custodia-check:196: [E0702(raising-bad-type), IPACustodiaTester._check_jwk_single] Raising NoneType while only classes or instances are allowed)
install/tools/ipa-custodia-check:213: [E0702(raising-bad-type), IPACustodiaTester._check_jwk_single] Raising NoneType while only classes or instances are allowed)
```

ipatests/test_integration/test_installation.py
```diff
@@ -16,25 +16,24 @@
...
 import pytest
-from cryptography.hazmat.primitives import hashes
 from cryptography import x509 as crypto_x509
+from cryptography.hazmat.primitives import hashes
 
 from ipalib import x509
 from ipalib.constants import DOMAIN_LEVEL_0
 from ipalib.constants import IPA_CA_RECORD
 from ipalib.sysrestore import SYSRESTORE_STATEFILE, SYSRESTORE_INDEXFILE
-from ipapython.dn import DN
```
Tiboris, line 26: Is this reorder forced by something?

miskopo: Yes, this order is encouraged by [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) and convention, that imports should be sorted in alphabetical order.

Tiboris: Yes, it should but this change was not needed for this file to pass checks, was it?
I do not see this as an issue just asking.

miskopo: not needed, it just a part of "holding code up to certain standards"

rcritten:
The commit message says what this does but not why lib389 is not always available.

I can't say I'm thrilled with the code duplication.

ipatests/test_integration/test_installation.py
tiran: I recommend to avoid refactoring unless it is required to make code more readable or to pass a linter. It reduces unnecessary noise.

ipatests/test_integration/test_installation.py
```diff
@@ -1255,8 +1254,12 @@ def test_kra_certs_renewal(self):
         """
         Test that the KRA subsystem certificates renew properly
         """
-        kra = krainstance.KRAInstance(self.master.domain.realm)
-        for nickname in kra.tracking_reqs:
+        tracking_reqs = {
+            'auditSigningCert cert-pki-kra': 'caAuditSigningCert',
+            'transportCert cert-pki-kra': 'caTransportCert',
+            'storageCert cert-pki-kra': 'caStorageCert',
+        }
```
tiran, line 1261: I also don't like code duplication. How about you move the information to a common place like ``ipalib.constants`` and import the definition from there?

miskopo: I understand your concern, therefore I've moved aforementioned code to the `constants`.

fcami: Marking resolved as it is resolved.

fcami (commented):
The change LGTM though I have difficulties understanding the reason of the change.

pvoborni:
The reason is simple. The real motivation is to not import from `ipaserver` in integration tests. It makes preparation of runner machine harder. Ideally integration tests should not depend on anything from the project it tests, but limiting it to server packages is a reasonable compromise.  We've been in this state before, but it was broken.

ipalib/constants.py
```diff
@@ -343,3 +343,9 @@
 # Apache's mod_ssl SSLVerifyDepth value (Maximum depth of CA
 # Certificates in Client Certificate verification)
 MOD_SSL_VERIFY_DEPTH = '5'
+
+TRACKING_REQS = {
```
fcami, line 347: Could you please add a comment here stating where this was duplicated from and why so that no refactoring happens down the road?

fcami (changes requested):

fcami:
I'm not that keen on duplicated code but I understand the reasoning now.
I think it's not needed for this change but if there are more changes like this to be done, I would like:
* having them in a separate file than constants - putting all of the duplicated code in a single-well-identified place seems best to me
* some kind of test that checks whether the duplicate stays as an exact duplicate and does not deviate from the original (or rather, the reverse) due to changes.

tiran:
I recommend to move the definition of constants into a common module in ipalib, e.g. ``ipalib.constants``.

fcami:
Indeed, moving the tracking_reqs dictionary from krainstance::KRAInstance ipalib.constants makes even more sense.

miskopo:
Thank you all for your feedback. I've also imported the original `dict` to `krainstance` where it was taken from to avoid duplication and provided comment to `ipalib.constants`.

fcami:
@miskopo thanks for the fix. It LGTM. Could you please change the commit message so that it reflects what the current change is? When this is done I'll ack it.

fcami (changes requested):
The only change I'd request is the rewording of the commit message as noted.

miskopo:
@fcami thank you for your comment. Commit message is edited, feel free to suggest more edits, if you think it's needed.

ipalib/constants.py
```diff
@@ -343,3 +343,11 @@
...
 # Certificates in Client Certificate verification)
 MOD_SSL_VERIFY_DEPTH = '5'
+
+# moved from ipaserver/install/krainstance.py::KRAInstance to avoid duplication
+# as per https://pagure.io/freeipa/issue/8795
+TRACKING_REQS = {
+    'auditSigningCert cert-pki-kra': 'caAuditSigningCert',
+    'transportCert cert-pki-kra': 'caTransportCert',
+    'storageCert cert-pki-kra': 'caStorageCert',
+}
```
flo-renaud, line 353: I'd rather use KRA_TRACKING_REQS instead of TRACKING_REQS to avoid any confusion.

flo-renaud (commented):
Hi @miskopo 
Just a nitpick, please see inline.

ipalib/constants.py
miskopo: @flo-renaud done :)

fcami:
LGTM. Please post a link to the successful CI run then remove the temp commit. I'll ACK afterwards.

miskopo:
[`temp_commit`](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/5d14002a-e09b-11eb-89c4-fa163eda45c5)

fcami:
Adding re-run because of a timeout

miskopo:
I did a rebase and now everything's broken. __sigh__

miskopo:
Seems like automatic rebase broke previous precious commit. I've fixed it manually, should be okay now.

miskopo:
Finally clear run! Could you please `ACK`, @flo-renaud  or @fcami ? I'll push the patch then. Thank you

flo-renaud:
> Finally clear run! Could you please `ACK`, @flo-renaud or @fcami ? I'll push the patch then. Thank you

Done :)

PR #5812, opened 2021-06-04
ipatests: set selinux context for fips mode

flo-renaud:
In order to test FIPS mode, the test is faking a user-space
FIPS environment by creating a file /var/tmp/userspace-fips
and bind-mounting this file as /proc/sys/crypto/fips_enabled

The security context needs to be properly set otherwise
/proc/sys/crypto/fips_enabled inherits the security context
unconfined_u:object_r:user_tmp_t:s0 and cannot be read,
resulting in the test seeing fips_mode=false.

Fixes: https://pagure.io/freeipa/issue/8868

### DIFF
--- a/ipatests/pytest_ipa/integration/fips.py
+++ b/ipatests/pytest_ipa/integration/fips.py
@@ -38,6 +38,9 @@ def enable_userspace_fips(host):
     host.run_command(["mkdir", "-p", FIPS_OVERLAY_DIR])
     host.put_file_contents(FIPS_OVERLAY, "1\n")
     host.run_command(
+        ["chcon", "-t", "sysctl_crypto_t", "-u", "system_u", FIPS_OVERLAY]
+    )
+    host.run_command(
         ["mount", "--bind", FIPS_OVERLAY, paths.PROC_FIPS_ENABLED]
     )
     # set crypto policy to FIPS mode


ipatests/pytest_ipa/integration/fips.py
```diff
@@ -37,6 +37,9 @@ def enable_userspace_fips(host):
     # fake Kernel FIPS mode with bind mount
     host.run_command(["mkdir", "-p", FIPS_OVERLAY_DIR])
     host.put_file_contents(FIPS_OVERLAY, "1\n")
+    host.run_command(
+        ["chcon", "-t", "sysctl_crypto_t", FIPS_OVERLAY]
+    )
```
rcritten, line 42: I think you should probably add the user too, -u system_u. Otherwise it is going to be unconfined_u I think.

flo-renaud: Thanks for the suggestion, PR updated.

rcritten:
Looks good, thanks.

rcritten:
Please remove the temp commit and I'll ack.

flo-renaud:
@rcritten thanks for thé review, temp commit removed and tests are green

rcritten:
master:

* 5c83ec06fa6aa23d8c540e9f5889f31aaf51dfbb ipatests: set selinux context for fips mode

PR #5813, opened 2021-06-04
krb_utils: Simplify get_credentials

stanislavlevin:
Previously, `get_credentials` raises either `ValueError` or re-raises `GSSError`. The former makes the handling of this function more difficult without a good reason.
    
With this change:
- `get_credentials` no longer handles exceptions by itself, but delegates
    this to the callers (which already process GSS errors).
- `get_credentials_if_valid` doesn't raise any expected exceptions, but
    return valid credentials (on the moment of calling) or None. This makes
    it consistent with docs.
    
Related: https://pagure.io/freeipa/issue/8873

tested scenarios:
1. missing creds &&  ipa user-find xxx
2. expired creds && ipa user-find xxx
3. https://pagure.io/freeipa/issue/7752 (removed ccache on server + ipa user-find xxx)
4. ccache sweeper

### DIFF
--- a/ipalib/krb_utils.py
+++ b/ipalib/krb_utils.py
@@ -19,13 +19,9 @@
 import time
 import re
 
-import six
 import gssapi
 
 from ipalib import errors
-
-if six.PY3:
-    unicode = str
 
 #-------------------------------------------------------------------------------
 
@@ -149,14 +145,24 @@ def get_credentials(name=None, ccache_name=None):
     store = None
     if ccache_name:
         store = {'ccache': ccache_name}
-    try:
-        return gssapi.Credentials(usage='initiate', name=name, store=store)
-    except gssapi.exceptions.GSSError as e:
-        if e.min_code in (  # pylint: disable=no-member
-            KRB5_FCC_NOFILE, GSSPROXY_KRB5_FCC_NOFILE, KRB5_CC_NOTFOUND,
-        ):
-            raise ValueError('"%s", ccache="%s"' % (e, ccache_name))
-        raise
+    """
+    https://datatracker.ietf.org/doc/html/rfc2744.html#section-5.2
+    gss_acquire_cred:
+        If credential acquisition is time-consuming for a mechanism, the
+    mechanism may choose to delay the actual acquisition until the
+    credential is required (e.g. by gss_init_sec_context or
+    gss_accept_sec_context). Such mechanism-specific implementation
+    decisions should be invisible to the calling application; thus a call
+    of gss_inquire_cred immediately following the call of
+    gss_acquire_cred must return valid credential data, and may therefore
+    incur the overhead of a deferred credential acquisition.
+
+    So, as gssapi.Credentials() calls only gss_acquire_cred it is not
+    guaranteed to have valid(not expired) returned creds and all the
+    callers of this function have to deal with GSSAPI exceptions by
+    themselves, for example, to handle ExpiredCredentialsError.
+    """
+    return gssapi.Credentials(usage="initiate", name=name, store=store)
 
 def get_principal(ccache_name=None):
     '''
@@ -174,9 +180,9 @@ def get_principal(ccache_name=None):
     '''
     try:
         creds = get_credentials(ccache_name=ccache_name)
-        return unicode(creds.name)
-    except ValueError as e:
-        raise errors.CCacheError(message=unicode(e))
+        return str(creds.name)
+    except gssapi.exceptions.GSSError as e:
+        raise errors.CCacheError(message=str(e))
 
 def get_credentials_if_valid(name=None, ccache_name=None):
     '''
@@ -199,8 +205,6 @@ def get_credentials_if_valid(name=None, ccache_name=None):
         creds = get_credentials(name=name, ccache_name=ccache_name)
         if creds.lifetime > 0:
             return creds
+    except gssapi.exceptions.GSSError:
         return None
-    except gssapi.exceptions.ExpiredCredentialsError:
-        return None
-    except ValueError:
-        return None
+    return None
--- a/install/share/gssproxy.conf.template
+++ b/install/share/gssproxy.conf.template
@@ -11,7 +11,6 @@
 [service/ipa-api]
   mechs = krb5
   cred_store = keytab:$HTTP_KEYTAB
-  cred_store = client_keytab:$HTTP_KEYTAB
   allow_constrained_delegation = true
   allow_client_ccache_sync = true
   cred_usage = initiate
--- a/install/tools/ipa-ccache-sweeper.in
+++ b/install/tools/ipa-ccache-sweeper.in
@@ -43,13 +43,8 @@ def should_delete(fname, t, minlife):
     # is not provided in cred_store the result of gss_acquire_cred_from
     # is KRB5_FCC_NOFILE, which is mapped by gssproxy to
     # 0x04200000 + KRB5_FCC_NOFILE.
-    try:
-        creds = get_credentials_if_valid(ccache_name=fname)
-        return creds is None
-    except ValueError:
-        return True
-
-    return False
+    creds = get_credentials_if_valid(ccache_name=fname)
+    return creds is None
 
 
 if __name__ == "__main__":
--- a/ipaserver/rpcserver.py
+++ b/ipaserver/rpcserver.py
@@ -913,6 +913,9 @@ class jsonserver_session(jsonserver, KerberosSession):
             else:
                 return self.service_unavailable(environ, start_response, msg)
 
+        except CCacheError:
+            return self.need_login(start_response)
+
         try:
             response = super(jsonserver_session, self).__call__(environ, start_response)
         finally:


ipalib/krb_utils.py
```diff
@@ -174,9 +180,9 @@ def get_principal(ccache_name=None):
     '''
     try:
         creds = get_credentials(ccache_name=ccache_name)
-        return unicode(creds.name)
-    except ValueError as e:
-        raise errors.CCacheError(message=unicode(e))
+        return str(creds.name)
+    except gssapi.exceptions.GSSError as e:
+        raise errors.CCacheError(message=str(e))
```
abbra, line 185: There are few places where `krb5_utils.get_principal()` call is not expected to raise an exception. May be they need to be protected?

abbra (commented):
I have one comment (inline), otherwise looks good.

ipalib/krb_utils.py
stanislavlevin: This change restored the logic that was before + refactored ValueError.
Usually catching of exceptions are added for error handling, pretty printing of error for users, etc.
The few places are:
1. ldap2 plugin
2. passwd plugin

No one of these handles  `CCacheError`.

abbra: Right, that's what I am asking improvement for. Since now we raise exception that is not handled, should we improve those two places?

stanislavlevin: I think the next possible scenario will raise `CCacheError`:
- ccache is outdated during the processing of user's request

If this is the case, I can add something for users.

stanislavlevin: > Right, that's what I am asking improvement for. Since now we raise exception that is not handled, should we improve those two places?

So was from original code bdccebbcdb9eb7da476762743121c1e73f95fa10.

abbra: Well, it was in a different location than these two. They indeed might appear when we have a user session that is about to expire and do process a command that takes some time to complete.

For WSGI, `ldap2.create_connection()` is only called at runtime through `Executioner.create_context()` so a failure there would be handled by the `KerberosWSGIExecutioner` code that wraps `PublicError` exceptions. The only place where this would fail is `/json/session` which does not intercept `PublicError` but does so for `ACIError` and `DatabaseError`.

For command line that would fail in `cli.run()` which calls `self.create_context` without any exception handling but then a `PublicError` exception would be logged in `ipalib/cli.py:run()`.

So I think we probably need to enforce `/json/session` handling of `PublicError` here.

stanislavlevin: Regarding 

> 2. passwd plugin
> 

This is only related to server's plugin with name 'passwd'. The minimal reproducer is
```
from ipalib import api

api.bootstrap(in_server=True)
api.finalize()

api.Command['passwd']()
```

```
(Pdb) bt
  //test.py(18)<module>()
-> api.Command['passwd']()
  /usr/lib/python3.9/site-packages/ipalib/frontend.py(471)__call__()
-> return self.__do_call(*args, **options)
  /usr/lib/python3.9/site-packages/ipalib/frontend.py(490)__do_call()
-> params.update(self.get_default(**params))
  /usr/lib/python3.9/site-packages/ipalib/frontend.py(721)get_default()
-> return dict(self.__get_default_iter(_params, kw))
  /usr/lib/python3.9/site-packages/ipalib/frontend.py(759)__get_default_iter()
-> default = param(None, **kw)
  /usr/lib/python3.9/site-packages/ipalib/parameters.py(624)__call__()
-> value = self.get_default(**kw)
  /usr/lib/python3.9/site-packages/ipalib/parameters.py(964)get_default()
-> default = self.default_from(**kw)
> /usr/lib/python3.9/site-packages/ipalib/parameters.py(268)__call__()
-> return self.callback(*vals)
  /usr/lib/python3.9/site-packages/ipaserver/plugins/passwd.py(88)princ_def()->None
-> return kerberos.Principal(krb_utils.get_principal())
  /usr/lib/python3.9/site-packages/ipalib/krb_utils.py(185)get_principal()
-> raise errors.CCacheError(message=str(e))

(Pdb) l 255, 272
255         def __call__(self, **kw):
256             """
257             Call the callback if all keys are present.
258  
259             If all keys are present, the callback is called and its return value is
260             returned.  If any keys are missing, ``None`` is returned.
261  
262             :param kw: The keyword arguments.
263             """
264             vals = tuple(kw.get(k, None) for k in self.keys)
265             if None in vals:
266                 return None
267             try:
268  ->             return self.callback(*vals)
269             except Exception:
270                 pass
271             return None
```
So, any exception is ignored for `default_from` and actual error is hidden.

stanislavlevin: It's meaningless to to handle CCacheError  for passwd.

abbra: Yep, sounds good.

stanislavlevin: Sorry for delay, in GSSPROXY context lifetime of credentials is not currently exposed. So, nothing will be raised until the actual usage of credentials.

stanislavlevin: For example,
```
> /usr/lib/python3.9/site-packages/ipaserver/plugins/ldap2.py(209)create_connection()
-> principal = krb_utils.get_principal(ccache_name=ccache)
(Pdb) ccache
'FILE:/run/ipa/ccaches/admin@IPA.TEST-yThOA2'
(Pdb) krb_utils.get_principal(ccache_name=ccache)
'HTTP/master1.ipa.test@IPA.TEST'
(Pdb) store = {'ccache': ccache}
(Pdb) store
{'ccache': 'FILE:/run/ipa/ccaches/admin@IPA.TEST-yThOA2'}
(Pdb) import gssapi
(Pdb) creds=gssapi.Credentials(usage="initiate", name=None, store=store)
(Pdb) creds
<gssapi.creds.Credentials object at 0x7f9266bf87f0>
(Pdb) creds.lifetime
20

[root@master1 /]# KRB5_KTNAME=/var/lib/ipa/gssproxy/http.keytab /usr/sbin/gssproxy -d --extract-ccache /run/ipa/ccaches/admin@IPA.TEST-yThOA2 --into-ccache ~/decryptedccache
[2021/06/08 09:13:05]: Debug Enabled (level: 1)
[2021/06/08 09:13:05]: Service: Extract Ccache, Keytab: /var/lib/ipa/gssproxy/http.keytab, Enctype: 18
decrypted
Failed with: The referenced credential has expired, Success
```

stanislavlevin: Asked in upstream: https://github.com/gssapi/gssproxy/issues/33

simo5: Just a note, 20 seconds is less than the clock skew of 5 minutes, so it may not give the results you hope for.

stanislavlevin: @simo5, thank you.

So, with `[service/ipa-api]` gssproxy service (in particular, with `cred_store = client_keytab:xxx`) and default GSS name (=None) `get_credentials` _in GSSProxy envs_ re-initiates expired credentials with client's keytab.

In my opinion this is not something we want and all the users of `get_credentials` should use an explicit GSS name or this function should be aware about `GSS_USE_PROXY` usage.

stanislavlevin: I'm not sure why we need `cred_store = client_keytab:xxx` for ipa-api:
```
[service/ipa-api]
  mechs = krb5
  cred_store = keytab:/var/lib/ipa/gssproxy/http.keytab
  cred_store = client_keytab:/var/lib/ipa/gssproxy/http.keytab
  allow_constrained_delegation = true
  allow_client_ccache_sync = true
  cred_usage = initiate
  euid = ipaapi
```
because we already has obtained delegated credentials from apache and stored in `KRB5CCNAME`, what the credentials can be refreshed for? at the same time replacing delegated creds with the initial ones on http's keytab.

@abbra, @simo5, do you remember?

abbra: ipa-api talks to LDAP and smbd, it needs to work as a client in some of those cases.

stanislavlevin: The initiator credentials for ipa-api are obtained from ccache (and can be only `initiate` according to config). `client_keytab` is not required unless creds in ccache are expired or missing.

abbra: Did you try to remove the `client_keytab`? I wonder if we have anything that actually requires to use the client keytab? Schema cache uses anonymous connection...

stanislavlevin: Yes, I tried. I don't see where this can be used. Gating tests don't catch any regressions.

abbra: @simo5 you were the one who added `client_keytab` to `gssproxy.conf.template` when introducing the support for gssproxy in FreeIPA in 2016. Was there a reason for client keytab to be specified? See commits d2f5fc304f1938d23171ae330fa20b213ceed54e and 4fd89833ee5421b05c10329d627d0e0fc8496046

simo5: I do not recall exactly, it may have been because we thought we were going to fully restrict anonymous access at some point. But if nothing is using it, we should just drop it.
It is actually kind of dangerous in retrospect as we may have bugs that end up using the HTTP credential and mask off access by adversaries, so I would keep only if we have a clear use case.

stanislavlevin:
With updated PR expired credentials are no longer auto-refreshed:
```
(Pdb) principal = krb_utils.get_principal(ccache_name=ccache)
*** ipalib.errors.CCacheError: Major (458752): No credentials were supplied, or the credentials were unavailable or inaccessible, Minor (2598845123): No credentials cache found
```
This exception (`CCacheError`) is caught in `jsonserver_session.__call__.create_context` and as `PublicError` in `KerberosWSGIExecutioner.__call__.create_context`.

stanislavlevin:
Re-checked manually original 4 scenarios, no regressions.

stanislavlevin:
run timout:
```
2021-06-11 10:41:08,430    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd38] Upgrading IPA:. Estimated time: 1 minute 30 seconds
2021-06-11 10:41:08,430    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd38]   [1/11]: stopping directory server
2021-06-11 10:41:11,497    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd38]   [2/11]: saving configuration
2021-06-11 10:41:11,761    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd38]   [3/11]: disabling listeners
2021-06-11 10:41:11,844    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd38]   [4/11]: enabling DS global lock
2021-06-11 10:41:11,914    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd38]   [5/11]: disabling Schema Compat
2021-06-11 10:41:11,950    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd38]   [6/11]: starting directory server
2021-06-11 10:41:14,463    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd38]   [7/11]: updating schema
2021-06-11 10:41:16,203    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd38]   [8/11]: upgrading server
2021-06-11 10:41:19,362    DEBUG  [ipatests.pytest_ipa.integration.host.Host.master.cmd38] Add failure missing required attribute "objectclass"
2021-06-11 11:52:56,153 CRITICAL  Terminating execution, runtime exceeded 4800s
```

stanislavlevin:
@abbra, thank you!
it was interesting issue :)

abbra:
master:

* 0a169b1bea997b489979f624f31057facfcfd984 krb_utils: Simplify get_credentials
* 0ebc59c26d54454a0821d64dcad1bbe7621b623c gssproxy: Don't refresh expired delegated credentials

PR #5821, opened 2021-06-09
ipatest: Test ipa-cert-fix fails when startup directive is missing from CS.cfg

mrizwan93:
This test checks that if 'selftests.container.order.startup' directive
is missing from CS.cfg, ipa-cert-fix fails and throw proper error
message. It also checks that underlying command 'pki-server cert-fix'
should fail to renew the cert.

related: https://pagure.io/freeipa/issue/8721

### DIFF
--- a/ipatests/test_integration/test_ipa_cert_fix.py
+++ b/ipatests/test_integration/test_ipa_cert_fix.py
@@ -93,6 +93,17 @@ class TestIpaCertFix(IntegrationTest):
         # the fixture
         pass
 
+    @pytest.fixture
+    def expire_ca_cert(self):
+        tasks.install_master(self.master, setup_dns=False,
+                             extra_args=['--no-ntp'])
+        move_date(self.master, 'stop', '+20Years+1day')
+
+        yield
+
+        tasks.uninstall_master(self.master)
+        move_date(self.master, 'start', '-20Years-1day')
+
     def test_missing_csr(self, expire_cert_critical):
         """
         Test that ipa-cert-fix succeeds when CSR is missing from CS.cfg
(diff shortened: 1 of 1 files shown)


rcritten:
The test looks ok but the comment and commit message I think are too narrowly focused on one failure use-case.

The investigation discovered that we were too optimistic when pki-cert-fix returned an error and continued to try to renew certificates only to end up with an even stranger message. The configuration you're testing matches this report but the underlying fix theoretically would apply to other types of failures.

I guess I'd emphasis more that we are exercising this better checking of general failures than preventing this one specific failure type. This particular failure is just one known way to cause the problem.

mrizwan93:
> The test looks ok but the comment and commit message I think are too narrowly focused on one failure use-case.
> 
> The investigation discovered that we were too optimistic when pki-cert-fix returned an error and continued to try to renew certificates only to end up with an even stranger message. The configuration you're testing matches this report but the underlying fix theoretically would apply to other types of failures.
> 
> I guess I'd emphasis more that we are exercising this better checking of general failures than preventing this one specific failure type. This particular failure is just one known way to cause the problem.

Bug report have expired CA and running ipa-cert-fix. I'll add this scenario.

rcritten:
Better, thanks. Please remove the temp commit.

mrizwan93:
Thanks @rcritten , temp commit removed.

rcritten:
master:

* 16057898af69ee795c9c2871ce5936a49d108e1c ipatest: Test ipa-cert-fix fails when startup directive is missing from CS.cfg

ipatests/test_integration/test_ipa_cert_fix.py
```diff
@@ -192,6 +203,53 @@ def test_ipa_cert_fix_non_ipa(self):
...
+                                 '--ignore-service-failures'])
+
+        result = self.master.run_command(['ipa-cert-fix', '-v'],
+                                         stdin_text='yes\n',
+                                         raiseonerr=False)
+        err_msg1 = "ERROR: 'selftests.container.order.startup'"
+        # check that pki-server cert-fix command fails
+        err_msg2 = ("ERROR: CalledProcessError(Command "
+                    "['pki-server', 'cert-fix'")
+        assert err_msg1 and err_msg2 in result.stderr_text
```
wladich, line 235: This means 
```
assert (err_msg1) and (err_msg2 in result.stderr_text)
```

So it does not check if err_msg1 is conained in output. It only checks that err_msg is not empty.

PR #5822, opened 2021-06-09
[ipa-4-6] CA less installation: non ASCII chars in CA subject

flo-renaud:
### CA less installation: non ASCII chars in CA subject
In CA-less installation, ipa-server-install fails when the CA
certificate contains a subject with non ASCII characters.

ipa-server-install is internally calling ipautil.run(...)
to launch a certutil -n nickname command, and the nickname is
provided as a unicode instead of a string.

The fix  makes sure the nickname is provided as a utf-8
encoded string.

Fixes: https://pagure.io/freeipa/issue/8879

### ipatests: use non-ascii chars in CA-less install

The CA-less installation creates an external CA with the
subject CN=CA,O=Example Organization.
In order to test non-ascii subjects, use
CN=CA,O=Example Organization España
instead.

Related: https://pagure.io/freeipa/issue/8879

### DIFF
--- a/ipalib/install/certstore.py
+++ b/ipalib/install/certstore.py
@@ -294,7 +294,7 @@ def get_ca_certs(ldap, base_dn, compat_realm, compat_ipa_ca,
                         'cACertificate;binary'])
 
         for entry in result:
-            nickname = entry.single_value['cn']
+            nickname = entry.single_value['cn'].encode('utf-8')
             trusted = entry.single_value.get('ipaKeyTrust', 'unknown').lower()
             if trusted == 'trusted':
                 trusted = True
--- a/ipatests/pytest_ipa/integration/create_caless_pki.py
+++ b/ipatests/pytest_ipa/integration/create_caless_pki.py
@@ -550,7 +550,7 @@ def create_pki():
                 x509.NameAttribute(NameOID.COMMON_NAME, server2)
              ])
              )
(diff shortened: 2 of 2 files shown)


flo-renaud:
Note for reviewers: the code is different in RHEL8 (ipautil.run does not call `_log_arg` which is raising an exception in python2 if the argument is a unicode with non-ascii chars), this is why the PR is relevant for ipa-4-6 branch only.
There is also an issue in ipa-4-9/master branch with the same scenario (CA less with non ASCII char in the CA subject), but it is located in a completely different place and I will open a separate ticket for that.

flo-renaud:
> Just a question: did you run the test without the fix? It looks ok to me but I haven't tested it.

@rcritten yes I used a f27 test controller with the patched test + a rhel 7.9 server and the test `test_caless.py::TestServerInstall::test_valid_certs` failed as expected (rhel 7.9 because 7.9 = python2, with a f27 server the issue would not happen).

rcritten:
Ok, thanks for that. This looks good, ack.

rcritten:
Please remove the temp commit and I'll ack for real ;-)

flo-renaud:
ipa-4-6:

* 82b4e8872c4890fd598a6e86924fd396810c845c CA less installation: non ASCII chars in CA subject
* f99067f4a34da4416d1877f63e257a8bec470bf5 ipatests: use non-ascii chars in CA-less install

PR #5825, opened 2021-06-10
ipa-cert-fix man page: add note about certmonger renewal

flo-renaud:
ipa-cert-fix man page needs to explain that certmonger may
trigger a renewal right after ipa-cert-fix completes because
certmonger does not notice the updated certificates.

Fixes: https://pagure.io/freeipa/issue/8702

### DIFF
--- a/install/tools/man/ipa-cert-fix.1
+++ b/install/tools/man/ipa-cert-fix.1
@@ -39,6 +39,13 @@ for shared certificates via \fIgetcert-resubmit(1)\fR (on the other
 CA server).  This is to avoid unnecessary renewal of shared
 certificates.
 
+Important note: the \fIcertmonger\fR daemon does not immediately notice
+the updated certificates and may trigger a renewal after \fIipa-cert-fix\fR
+completes. As a consequence, \fIgetcert list\fR output may display
+that a renewal is in progress even if \fIipa-cert-fix\fR just
+finished. It is recommended to monitor the certmonger-initiated
+renewal and wait for its completion before any other administrative task.
+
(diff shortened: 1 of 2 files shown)


mrizwan93 (commented):
Suggestion:
Can we add note to `ipa-cert-fix` command's output?
```
[..]
Note: Monitor the renewal status of certs and wait for its completion before any other administrative task

ipapython.admintool: INFO: The ipa-cert-fix command was successful.
```

flo-renaud:
Please don't review this PR yet, I am planning to test whether stopping certmonger before renewal and starting it at the end of ipa-cert-fix solves the double-renewal issue. The stop needs to be set at the right time because any call to `getcert list` would restart it.

flo-renaud:
@rcritten 
I did a test with certmonger stopped just before calling `pki-server cert-fix`, and restarting certmonger right after the call to `ipactl restart`. This doesn't fix the double-renewal issue, certmonger starts by reading the `/var/lib/certmonger/requests/` files and the request is marked with `state=CA_UNREACHABLE`. Probably because of this state, certmonger directly tries to renew the cert without checking its actual expiry date.
As a conclusion, there is no fix for double renewal unless certmonger code is modified, and we need this PR to explain the situation in the man page and at the end of the command.

rcritten:
Thanks for testing that. Maybe it's something I can look at in certmonger eventually.

ack

flo-renaud:
master:

* 5509e00a822ed5155cec3346bed554892bd1ff9e ipa-cert-fix man page: add note about certmonger renewal

PR #5826, opened 2021-06-10
host: provide a better error message with updatedns when FQDN is not provided

antoniotorresm:
The --updatedns option for host-mod requires the FQDN to be passed,
otherwise an obscure error message is shown since the domain cannot be
found.

Fixes: https://pagure.io/freeipa/issue/8726


rcritten:
Another approach, that might also fix https://bugzilla.redhat.com/show_bug.cgi?id=1967325, would be to try to resolve the fqdn if one is not provided, in host-*.

antoniotorresm:
Closing this PR as it will be fixed when #5838 is merged.

PR #5827, opened 2021-06-10
ipatests: disable test_nfs.py::TestNFS in nightly runs on Fedora 33

wladich:
Also disable in Fedora 33 as it also has the faulty version of sssd
which produces multi-gigabyte log file

Related to https://pagure.io/freeipa/issue/8877


flo-renaud:
master:

* accc27e663a268b61af826d6714d7a9ea6c70bbc ipatests: disable test_nfs.py::TestNFS in nightly runs on Fedora 33

PR #5829, opened 2021-06-10
WSGI: Control passed environment variables

stanislavlevin:
It's common practice to care about input environment variables and to produce a minimal and safe set of such.
    
Fixes: https://pagure.io/freeipa/issue/8881

### DIFF
--- a/install/share/kdcproxy.wsgi
+++ b/install/share/kdcproxy.wsgi
@@ -1,5 +1,25 @@
 # Copyright (C) 2017  FreeIPA Contributors see COPYING for license
 """WSGI entry point for kdcproxy
 """
+import os
+
 from kdcproxy import application
 
+
+def set_env():
+    """Replace current env with new one"""
+    # KDCPROXY_CONFIG is used by kdcproxy
+    pass_vars = ["KDCPROXY_CONFIG", "LANG", "LC_ALL"]
+    new_env = {}
+    for var in pass_vars:
+        try:
+            new_env[var] = os.environ[var]
+        except KeyError:
+            pass
+
+    new_env["PATH"] = "/usr/bin:/bin"
+    os.environ.clear()
+    os.environ.update(new_env)
+
+
+set_env()
--- a/install/share/wsgi.py
+++ b/install/share/wsgi.py
@@ -23,6 +23,26 @@
 """
 WSGI application for IPA server.
 """
+import os
+
 from ipaserver.wsgi import create_application
+
+
+def set_env():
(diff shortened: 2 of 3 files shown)


stanislavlevin:
Sure.

I'm sorry I wasn't clear.
There are at least 3 entities within one systemd unit "httpd":
- Apache
- IPA framework
- conditional kdcproxy

Each of them require external OS env vars. Setting a required var makes it common for all entities. But some of entities may be broken/confused with common env var, which they don't expect or don't properly handled. In my particular example, I worry about `KRB5CCNAME` set for Apache, bu the other entities don't need it and it may confuse IPA.
I think it's safer to split all the passed vars into their targets:
- `KRB5CCNAME` only for Apache
- `KDCPROXY_CONFIG` only for kdcproxy
- `GSS_USE_PROXY` for Apache and IPA

tiran:
This has never been a problem in the last 6 years. Is there any reason why you want to change the env vars now? Did you run into a specific issue?

The new code also prevents users from setting additional env vars like ``PYTHONBREAKPOINT`` to perform remote debugging.

stanislavlevin:
If there were no problems before, it does not mean that there are no problems.

I don't try to solve any known particular issue here. This is the kind of precaution.

tiran:
I'm strong -1 in regards of removing all env vars. There are valid use cases to pass additional envs vars.

I'm -0 to do ``os.environ.pop("KRB5CCNAME")`` in the WSGI startup scripts.

stanislavlevin:
Do you know these cases? Can I know them, please?

tiran:
* PYTHONBREAKPOINT to set up a custom debugger
* PYTHONFAULTHANDLER to debug crashes
* PYTHONWARNINGS to check for deprecations
* KRB5_TRACE to debug KRB5 issues

and more

stanislavlevin:
If all are related to debugging purposes then I can add some debug var,  the presence of which allows to bypass all the env vars. Would it be solution?

tiran:
No, there are more use cases for env vars. The list is not complete and contains just a few env vars that I can remember using in the past.

stanislavlevin:
I assume that only systemd is supported.
The person who wants to pass in some additional vars for any reason will most likely do it via modifying the unit file or dropping a new one and thus should be a privileged user. So, anyway this person will pass a list of desired vars and adding a new one(to bypass all) is almost zero cost.

stanislavlevin:
Seems we are at a stalemate.
Ok, since nobody else comments here, closing.

PR #5831, opened 2021-06-11
Incorporate custodia into IPA

tiran:
https://pagure.io/freeipa/issue/8882


abbra:
@tiran thanks for PR. Do we need to obsolete python3-custodia? Are you planing to remove it from the distributions too? Perhaps, obsoleting is not needed right now as we are not going to conflict by file paths anyway, right?

abbra:
Ok, next question is whether we should import this in ipa-4-9 branch. Without that I am not sure when this change would hit the distributions. 4.10? 5.0? How bad is it to include it into 4.9.5?

tiran:
I don't see any harm to backport the change to 4.9. You may want to copy some unit tests from upstream into downstream first.

rcritten:
Backporting would let us accelerate the retirement of custodia as a separate package and retire it in rawhide and potentially downstream distributions.

abbra:
Ok, may be we should then add those unit tests to IPA as well?

rcritten:
+1 on incorporating the tests

rcritten:
I filed upstream ticket to explain at least some of the reasoning behind this. Can you add this ticket to the commit msgs? https://pagure.io/freeipa/issue/8882

rcritten:
Take this with a grain of salt, I don't know custodia, but client.py and forwarder.py aren't used (except self-referrentially and client in one test).

Should we add a Provides so that python3-custodia and custodia can be removed on upgrade?

In general this change looks to be in good shape.

tiran:
> Take this with a grain of salt, I don't know custodia, but client.py and forwarder.py aren't used (except self-referrentially and client in one test).

Excellent finding! I removed the modules.

> Should we add a Provides so that python3-custodia and custodia can be removed on upgrade?

The copied code does not provide ``python3-custodia`` and is independent from upstream Custodia. The import paths and plugin entry points are prefixed with ``ipaserver``. You can have both python3-custodia package and freeipa-server package installed without any conflict.

I'll orphan custodia after the next update has landed in Fedora.

rcritten:
The only I asked about the Provides is on upgrades the custodia packages won't be removed. They don't harm anything per-se. Not a show stopper.

LGTM.

rcritten (approved):

rcritten:
master:

* 1e98f310f6e5a33218ae5625c1a0a4f7b02692b7 Add Custodia 0.6.0 to ipaserver package
* d27f01b2fbc9b539e57a5ce04914b76714ba4128 Remove unused Custodia modules
* a4631b7f3f1eca068c3ff6fbfe91563dbe911c4b Fix Custodia imports
* e1abfe0fb7079f47c63e65048b1a54fa32fe258a Fix Custodia pylint issues
* c27233e3a0d4c02479bbb91a21bfa3b608931fcf Remove more unused Custodia code
* 470bb6ed45b9b1b2391c2e12a11be7769807edcd Add Custodia tests
* e6f09c1b5f59ce6487d5d34399926defb4471953 Also drop Custodia client and forwarder

PR #5832, opened 2021-06-11
Allow PKINIT to be enabled when updating from a pre-PKINIT IPA CA server

antoniotorresm:
When upgrading from a server with IPA CA before PKINIT was introduced
(4.5), PKINIT would not be enabled and there wasn't any way to enable it
since upgrade code only issues self-signed certificates when
certificates are missing. With these change there is a way to enable
PKINIT when coming from a IPA server with a pre-PKINIT version (4.4 and
before).

Fixes: https://pagure.io/freeipa/issue/8532

### DIFF
--- a/ipaserver/install/ipa_pkinit_manage.py
+++ b/ipaserver/install/ipa_pkinit_manage.py
@@ -78,6 +78,8 @@ class PKINITManage(AdminTool):
             krb.enable_ssl()
 
         if setup_pkinit:
+            if not is_pkinit_enabled():
+                krb.setup_pkinit()
             krb.pkinit_enable()
         else:
             krb.pkinit_disable()


abbra:
LGTM. Please remove the temp commit

antoniotorresm:
Thank you, temp commit removed.

rcritten:
What branches, if any, do we need to backport this to?

antoniotorresm:
This is useful when upgrading from pre-4.5 release, so I guess it would be good to backport it to at least 4.6.

rcritten:
Sure, why not. I'm not sure if we're going to do another 4.6 release but if we do this would be a nice fix.

rcritten:
master:

* 50306cc636955ff5e478819e98224953cb6346ee Allow PKINIT to be enabled when updating from a pre-PKINIT IPA CA server

PR #5835, opened 2021-06-14
run test_kdc_proxy to test cygwin installation

wladich:


wladich:
The PR was tested on pers
https://github.com/wladich/freeipa/pull/54

PR #5838, opened 2021-06-15
host: try to resolve FQDN before command execution

antoniotorresm:
Trying to resolve the FQDN before command execution (during
pre-callback) helps detect cases where the host specified by the user
does not exist, saving execution time. Aside from this, resolving the
FQDN is useful when only the shortname of the host is passed, as this
would cause issues when trying to update the DNS records during
modification of the entry.

Fixes: https://pagure.io/freeipa/issue/8726
Fixes: https://pagure.io/freeipa/issue/8884

### DIFF
--- a/ipaserver/plugins/host.py
+++ b/ipaserver/plugins/host.py
@@ -252,6 +252,11 @@ def validate_ipaddr(ugettext, ipaddr):
     return None
 
 
+def resolve_fqdn(name):
+    hostentry = api.Command['host_show'](name)['result']
+    return hostentry['fqdn'][0]
+
+
 class HostPassword(Str):
     """
     A data type for host passwords to not log password values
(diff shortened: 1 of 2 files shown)


ipaserver/plugins/host.py
```diff
@@ -252,6 +252,11 @@ def validate_ipaddr(ugettext, ipaddr):
     return None
 
 
+def resolve_fqdn(name):
+    hostentry = api.Command['host_show'](name)['result']
+    return hostentry['fqdn'][0]
+
```
rcritten, line 258: It is probably worth it to retain the host validation. It can save an LDAP read if a FQDN is passed in and I don't think it changes the behavior. Or does it?

rcritten:
I toyed with a couple of other host-* commands and didn't see any need for the lookup there. Did you come to the same conclusion?

ipaserver/plugins/host.py
antoniotorresm: The idea of this is to act as a existence check so the LDAP search will be performed regardless of whether a FQDN is passed or not. I removed it because of this, after all the host search already returns the FQDN 'validated'.

antoniotorresm:
> I toyed with a couple of other host-* commands and didn't see any need for the lookup there. Did you come to the same conclusion?

I just added the search where it was being done before. It acts as an existence check so I think it's good to have it on top of the pre-callback function.

ipaserver/plugins/host.py
rcritten: Yes of course, that's the whole point of this exercise.

rcritten:
This looks good. Can you add a test to ensure that the shortname use-cases work as expected? The xmlrpc tests would be probably the best place.

rcritten:
LGTM.

Can you set the label for backports? I assume ipa-4-9 is sufficient. You can set ack at the same time if you want, or I can add it later.

rcritten:
master:

* 3a4939fa390667e964248bbc3ec4a050910ae6c8 host: try to resolve FQDN before command execution
* 3e77d3132fe44cc2ec579b0e41f92d9a7a558d4c ipatests: test host update using shortname

PR #5841, opened 2021-06-16
dnszone-add: deprecate option for setting SOA serial

antoniotorresm:
Since IPA 3 [1] SOA serial is managed automatically via autoincrement,
and the option of disabling this behavior was deprecated in IPA 3.3.3 [2].
As a result, the option `--serial` during DNS zone addition would be
ignored as it is set during the creation. This commit adds a deprecation
warning if this option is used. Setting the SOA serial manually would still
be useful in certain scenarios, so the option of doing it with dnszone-mod
is kept.

[1] https://www.freeipa.org/page/V3/DNS_SOA_serial_auto-incrementation
[2] https://www.freeipa.org/page/Releases/3.3.3

Fixes: https://pagure.io/freeipa/issue/8227

### DIFF
--- a/ipaserver/plugins/dns.py
+++ b/ipaserver/plugins/dns.py
@@ -2432,6 +2432,7 @@ class dnszone(DNSZoneBase):
             autofill=True,
         ),
         Int('idnssoaserial',
+            # Deprecated
             cli_name='serial',
             label=_('SOA serial'),
             doc=_('SOA record serial number'),
@@ -2439,6 +2440,7 @@ class dnszone(DNSZoneBase):
             maxvalue=4294967295,
             default_from=_create_zone_serial,
             autofill=True,
+            flags=['no_option'],
         ),
         Int('idnssoarefresh',
             cli_name='refresh',
@@ -2776,6 +2778,14 @@ class dnszone_add(DNSZoneBase_add):
                     option='ip-address',
                     additional_info=u"Value will be ignored.")
             )
+        if 'idnssoaserial' in options:
+            messages.add_message(
+                options['version'],
+                result,
+                messages.OptionDeprecatedWarning(
+                    option='idnssoaserial',
+                    additional_info=u"Value will be ignored.")
+            )
 
     def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys, **options):
         assert isinstance(dn, DN)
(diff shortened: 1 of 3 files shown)


rjeffman (commented):
If I understand it correctly, by deprecating idnssoaserial, we should also remove it from the command line options (as is done with `ip-address`).

rcritten:
As in set the no_option flag. We'd need to get this into c9s so that future versions of RHEL do not have the option so its being missing is not considered a regression.

rjeffman:
Why deprecate the soa serial when adding a new zone, but not when modifying a new zone? It would still allow a user to create a zone with a specific serial, but would have to execute two commands to achieve that.

Also, any change to the zone does modify the soa serial to the current UNIX timestamp, so between a "mod" and a "show" the soa serial can be changed (we've seen this in ansible-freeipa tests).

I think that we either disallow setting soa serial at all (which I would prefer), or allow it both in add and mod.

abbra:
We cannot remove the attribute from API. But we can ignore it and issue a warning. Perhaps, this would be a better way?

antoniotorresm:
I can add the same warning for dnszone-mod too. Should we also remove the option from the CLI?

rjeffman:
Sorry if it sounded that I wanted to remove the attribute for API, that's not what I meant. I meant on making user interface consistent.

If we can ignore the attribute in the API and return a warning, and hide it from CLI on both add and mod, it would be great.

antoniotorresm:
I have disabled the option in the CLI and added deprecation warnings to the logger when the option is used.

ipaserver/plugins/dns.py
```diff
@@ -2783,6 +2785,10 @@ def pre_callback(self, ldap, dn, entry_attrs, attrs_list, *keys, **options):
         if options.get('force'):
             options['skip_nameserver_check'] = True
 
+        if 'idnssoaserial' in options:
+            logger.warning("'idnssoaserial' option is deprecated."
+                           "Value will be ignored.")
+
```
rcritten, line 2791: This will log on the server side. I guess it's a heads-up of sorts.  You want to look at messages.add_message() instead.

antoniotorresm:
Added the warning message and updated the tests to expect the new warning.

rcritten:
This change LGTM. I think I'd like @rjeffman to take another look before merging.

rjeffman:
Patch looks good to me.

fcami (approved):
LGTM. ACK even.

antoniotorresm:
master:

* e11a78a29e16bc7f34139e486d759ac8156f0208 dnszone: deprecate option for setting SOA serial
* 715987a82e9270b1ba9812f530f42c64af6aa093 ipatests: expect SOA serial option deprecation warning

PR #5844, opened 2021-06-18
WebUI: Fix certificate serial number representation

serg-cymbaluk:
Big numbers are automatically translated to scientific notation in JavaScript.
It causes an issue with some certificate serial numbers.
The fix normalizes the notation base on original value from serial_number_hex.

The implementation works only for browsers that support BigInt.
It would not work for old browsers like Internet Explorer.

Ticket: https://pagure.io/freeipa/issue/8754

### DIFF
--- a/install/ui/src/freeipa/certificate.js
+++ b/install/ui/src/freeipa/certificate.js
@@ -20,6 +20,7 @@
  */
 
 define([
+    'dojo/_base/declare',
     'dojo/_base/lang',
     'dojo/on',
     './builder',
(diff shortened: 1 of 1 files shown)


serg-cymbaluk:
Hi @flo-renaud 

Could you please take a look at the patch?

flo-renaud:
Hi @serg-cymbaluk 
thanks for the patch. It works well for displaying the certificates in the user/host/service pages (the box containing the certificate with serial number/issued by/valid from/valid to contains the right serial number, and clicking on "View" also displays the right value).
I couldn't test the revoke button with a high serial number but it's working as expected for certs issued by IPA.

There is something that isn't consistent, though. When navigating to Authentication > Certificates > Certificates, the table containing all the certificates still displays the wrong serial number in the column "serial_number".
certificates_table

serg-cymbaluk:
@flo-renaud 
Thanks for the feedback.
I will provide a fix for the 'Certificates' page ASAP.

serg-cymbaluk:
@flo-renaud 

PR updated. Could you please check the changes?

flo-renaud:
Hi @serg-cymbaluk 
thanks for the updated patch, works for me. The certificates are now displayed with the proper serial number in the certificates table.
You can remove the temp commit and I will ack when PRCI completes.
certs_table

serg-cymbaluk:
@flo-renaud 

Temp commit is removed. Pipeline is green.

serg-cymbaluk:
master:

* 1439764c0e8ab9fd046c465e00f065684682f6b0 WebUI: Fix certificate serial number representation

PR #5845, opened 2021-06-18
 installer: Fix installation on hosts without /proc/1/sched. 

pfandl:
My host does not have `/proc/1/sched` and this patch fixes the installation exception.

I have not tested the other code path where `ìs_hidepid` is `True`.

best regards

### DIFF
--- a/ipaserver/install/installutils.py
+++ b/ipaserver/install/installutils.py
@@ -1098,49 +1098,49 @@ def in_container():
     """
     try:
         return tasks.detect_container() is not None
-    except subprocess.CalledProcessError:
+    except (subprocess.CalledProcessError, NotImplementedError):
         pass
-    except NotImplementedError:
+
+    try:
+        with open("/proc/sys/kernel/osrelease") as release:
+            if release in ["Microsoft", "WSL"]:
+                return True
+    except FileNotFoundError:
+        pass
+
+    # TODO: detect proot
+
+    def check_container(container):
+        if container == "oci":
+            # Some images hardcode container=oci, but OCI is not a specific
+            # container manager. Try to detect one based on well-known
+            # files.
+            return detect_container_files()
+        return bool(container)
+
+    for path in ["/run/host/container-manager", "/run/systemd/container"]:
         try:
-            with open("/proc/sys/kernel/osrelease") as release:
-                if release in ["Microsoft", "WSL"]:
-                    return True
+            with open(path) as manager:
+                return check_container(manager.readline())
         except FileNotFoundError:
             pass
 
-        # TODO: detect proot
+    try:
+        with open("/proc/1/environ") as env:
+            for line in env.read().split("\0"):
+                if not line.startswith("container"):
+                    continue
 
-        def check_container(container):
-            if container == "oci":
-                # Some images hardcode container=oci, but OCI is not a specific
-                # container manager. Try to detect one based on well-known
-                # files.
-                return detect_container_files()
-            return bool(container)
+                return check_container(line.split("=")[1])
+    except FileNotFoundError:
+        pass
 
-        for path in ["/run/host/container-manager", "/run/systemd/container"]:
-            try:
-                with open(path) as manager:
-                    return check_container(manager.readline())
-            except FileNotFoundError:
-                pass
+    if detect_container_files():
+        return True
+    if running_in_cgroupns():
+        return True
 
-        try:
-            with open("/proc/1/environ") as env:
-                for line in env.read().split("\0"):
-                    if not line.startswith("container"):
-                        continue
-
-                    return check_container(line.split("=")[1])
-        except FileNotFoundError:
-            pass
-
-        if detect_container_files():
-            return True
-        if running_in_cgroupns():
-            return True
-
-        return False
+    return False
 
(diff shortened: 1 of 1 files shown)


abbra:
Hi, thank you for the patch. There are few issues I'd like to address first:
1. Please use human-identifying Author in a commit message instead of root@<hostname>
2. There are syntax issues reported by the `make fastlint`:
```

2021-06-18T10:28:37.4467463Z pycodestyle
2021-06-18T10:28:37.4468401Z -----------
2021-06-18T10:28:38.0382577Z ./ipaserver/install/installutils.py:1045:9: E722 do not use bare 'except'
2021-06-18T10:28:38.0384175Z ./ipaserver/install/installutils.py:1046:22: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
2021-06-18T10:28:38.0453050Z make: *** [Makefile:1240: fastlint] Error 1
```
3. If you are moving `split()[0]` to inner call, please do the same with `data_cgroup` as well, for consistency.

Finally, just a notification: 389-ds project formally does not support 32-bit architectures (including armv7) and if you are considering to use FreeIPA in production on armv7, it might be not a good idea.

abbra:
Also, please use rebase, not merge. We do not accept merges.

pfandl:
>     1. If you are moving `split()[0]` to inner call, please do the same with `data_cgroup` as well, for consistency.

Alright, will do.

> 
> Finally, just a notification: 389-ds project formally does not support 32-bit architectures (including armv7) and if you are considering to use FreeIPA in production on armv7, it might be not a good idea.

Oh, I didn't know that. I do this for a homelab setup for me to use and maybe some friends. Would you still consider this a bad idea?

> Also, please use rebase, not merge. We do not accept merges.

Not sure yet how to do that, will look into it.

ipaserver/install/installutils.py
```diff
@@ -1037,17 +1037,20 @@ def in_container():
...
     if not is_hidepid():
-        with open('/proc/1/sched', 'r') as sched:
-            data_sched = sched.readline()
-    else:
-        data_sched = []
+        try:
+            with open('/proc/1/sched', 'r') as sched:
+                data_sched = sched.readline().split()[0]
+        except FileNotFoundError:
+            if not os.system('systemd-detect-virt --container'):
```
tiran, line 1046: Don't use ``os.system``. It invokes a shell and is potentially dangerous. Use ``tasks.detect_container()`` to detect containers.

abbra:
> Oh, I didn't know that. I do this for a homelab setup for me to use and maybe some friends. Would you still consider this a bad idea?
> 

I would suggest you to use arm64 if possible. This archtecture is fully supported. FreeIPA on Fedora works just fine on Raspberry Pi4B+, for example. 389-ds had intent to remove any 32-bit support but they missed to mark explicitly armhfp in Fedora.

> > Also, please use rebase, not merge. We do not accept merges.
> 
> Not sure yet how to do that, will look into it.

`git rebase origin/master` should be enough instead of `git merge`

pfandl:
> I would suggest you to use arm64 if possible. This archtecture is fully supported. FreeIPA on Fedora works just fine on Raspberry Pi4B+, for example. 389-ds had intent to remove any 32-bit support but they missed to mark explicitly armhfp in Fedora.

Is the unavailable `/proc/1/sched` specific ty this platform, and if so I guess we can close this pull request anyway?

abbra:
I think it is a kernel-specific thing. According to https://github.com/systemd/systemd/pull/17917, we should not use `/proc/1/sched` for the container detection since kernel 4.14. See also https://github.com/systemd/systemd/pull/17902, so we most likely need to rewrite this detector code anyway.

pfandl:
> I think it is a kernel-specific thing. According to [systemd/systemd#17917](https://github.com/systemd/systemd/pull/17917), we should not use `/proc/1/sched` for the container detection since kernel 4.14. See also [systemd/systemd#17902](https://github.com/systemd/systemd/pull/17902), so we most likely need to rewrite this detector code anyway.

What is the goal, to just try `tasks.detect_container()` and if it is not implemented to implement following steps in the python script?

https://github.com/systemd/systemd/pull/17902#issuecomment-748627568

abbra:
That sounds like a good plan. We don't need to know cgroups v1/v2 difference, just that we are in container or not.

pfandl:
> That sounds like a good plan. We don't need to know cgroups v1/v2 difference, just that we are in container or not.

I think i could replicate the logic, I tested it in docker and podman and outside a container and seems the automatic tests are fine as well.

The "proot" whatever it is is not implemented and I can't test WSL and OpenVZ as of now.

tiran:
I don't like the fact that this PR re-implements ``systemd-virt-check``. This increases our support surface and makes our lives harder. I deliberately used the systemd tool, because it takes care of all edge cases. It's not FreeIPA's job to track how various container and virtualization software work.

If ``systemd-virt-check`` does not work on your platform, then the issue should either be treated as a systemd bug or as a platform bug.

tiran:
``/sys/fs/cgroup/memory/memory.limit_in_bytes`` is cgroups v1-only and does not exist in cgroups v2.

PR #5850, opened 2021-06-22
ipatests: healthcheck: Update IPAHostKeytab assumptions

stanislavlevin:
As of 0.9 freeipa-healthcheck requires running `dirsrv` service
for `IPAHostKeytab` check. So, previous assumption about the
triggering the GSSAPI error no longer works. For example, this can
be achieved by deletion of host's keytab.

Fixes: https://pagure.io/freeipa/issue/8889

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -492,44 +492,29 @@ class TestIpaHealthCheck(IntegrationTest):
         assert data[0]["kw"]["ipa_version"] in result.stdout_text
         assert data[0]["kw"]["ipa_api_version"] in result.stdout_text
 
-    def test_source_ipahealthcheck_ipa_host_check_ipahostkeytab(
-        self, restart_service
-    ):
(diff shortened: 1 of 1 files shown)


rcritten:
Looks ok to me. Another alternative (or addition) would be to use ipa-getkeytab to get new keys in a different location so that the kvno in /etc/krb5.keytab is old. Maybe a future enhancement. You'd get something like:

```
  {
    "source": "ipahealthcheck.ipa.host",
    "check": "IPAHostKeytab",
    "result": "ERROR",
    "uuid": "8272406b-4d5f-4d6e-b9e2-4fe0d6afb213",
    "when": "20210622182721Z",
    "duration": "0.079473",
    "kw": {
      "msg": "Failed to obtain host TGT: Major (851968): Unspecified GSS failure.  Minor code may provide more information, Minor (2529638936): Preauthentication failed"
    }
  }
```

stanislavlevin:
@rcritten, sure. Many ways to achieve the same.

rcritten:
master:

* 3597ab9bac6afbb5f055b2d62cb60f3e2850b31e ipatests: healthcheck: Update IPAHostKeytab assumptions

PR #5852, opened 2021-06-22
rpcserver.py: perf_counter_ns is Python 3.7+

fcami:
perf_counter_ns is only available in Python 3.7 and later.
Define a lambda for 3.6 and lower.

Fixes: https://pagure.io/freeipa/issue/8891

### DIFF
--- a/ipaserver/rpcserver.py
+++ b/ipaserver/rpcserver.py
@@ -31,6 +31,7 @@ import os
 import time
 import traceback
 from io import BytesIO
+from sys import version_info
 from urllib.parse import parse_qs
 from xmlrpc.client import Fault
 
@@ -71,6 +72,10 @@ from requests.auth import AuthBase
 
 if six.PY3:
     unicode = str
+
+# time.perf_counter_ns appeared in Python 3.7.
+if version_info < (3, 7):
+    time.perf_counter_ns = lambda: int(time.perf_counter() * 10**9)
 
 logger = logging.getLogger(__name__)
 
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -703,6 +703,33 @@ class TestInstallMaster(IntegrationTest):
     def test_install_master(self):
         tasks.install_master(self.master, setup_dns=False)
 
+    @pytest.mark.skip_if_platform(
+        "debian", reason="This test hardcodes the httpd service name"
+    )
+    def test_smoke_test_for_debug_mode(self):
+        """Test if an IPA server works in debug mode.
+        Related: https://pagure.io/freeipa/issue/8891
+
+        Note: this test hardcodes the "httpd" service name.
+        """
+
+        target_fname = paths.IPA_SERVER_CONF
+        assert not self.master.transport.file_exists(target_fname)
+
+        # set the IPA server in debug mode
+        server_conf = "[global]\ndebug=True"
+        self.master.put_file_contents(target_fname, server_conf)
+        self.master.run_command(["systemctl", "restart", "httpd"])
+
+        # smoke test in debug mode
+        tasks.kdestroy_all(self.master)
+        tasks.kinit_admin(self.master)
+        self.master.run_command(["ipa", "user-show", "admin"])
+
+        # rollback
+        self.master.run_command(["rm", target_fname])
+        self.master.run_command(["systemctl", "restart", "httpd"])
+
     def test_schema_compat_attribute_and_tree_disable(self):
         """Test if schema-compat-entry-attribute is set
 
--- a/ipaplatform/base/paths.py
+++ b/ipaplatform/base/paths.py
@@ -71,6 +71,7 @@ class BasePathNamespace:
     IPA_DEFAULT_CONF = "/etc/ipa/default.conf"
     IPA_DNSKEYSYNCD_KEYTAB = "/etc/ipa/dnssec/ipa-dnskeysyncd.keytab"
     IPA_ODS_EXPORTER_KEYTAB = "/etc/ipa/dnssec/ipa-ods-exporter.keytab"
+    IPA_SERVER_CONF = "/etc/ipa/server.conf"
     DNSSEC_OPENSSL_CONF = "/etc/ipa/dnssec/openssl.cnf"
     DNSSEC_SOFTHSM2_CONF = "/etc/ipa/dnssec/softhsm2.conf"
     DNSSEC_SOFTHSM_PIN_SO = "/etc/ipa/dnssec/softhsm_pin_so"


fcami:
Note: the to_remove commit is for reviewing purposes and will be removed with the temp commit.

spoore1 (approved):
Looks good to me.   

I did confirm that ipa user-show fails with Internal Server Error on a system without the fix.

I also manually tested the version check with python3.6.

fcami:
Thanks @spoore1 I've removed the temp commit.

spoore1:
ACK for when CI passes

ipaserver/rpcserver.py
```diff
@@ -72,6 +73,11 @@
 if six.PY3:
     unicode = str
 
+# time.perf_counter_ns appeared in Python 3.7.
+if (version_info.major == 3 and version_info.minor < 7) or \
+   (version_info.major == 2):
```
stanislavlevin, line 78: would it be shorter to compare tuples?
```python
if version_info < (3, 7):
```

ipatests/test_integration/test_installation.py
```diff
@@ -703,6 +703,30 @@ def install(cls, mh):
     def test_install_master(self):
         tasks.install_master(self.master, setup_dns=False)
 
+    def test_smoke_test_for_debug_mode(self):
+        """Test if an IPA server works in debug mode.
+        Related: https://pagure.io/freeipa/issue/8891
+        """
+
+        target_fname = paths.IPA_SERVER_CONF
+        assert not self.master.transport.file_exists(target_fname)
```
stanislavlevin, line 712: Is it expected that this test fails if `server.conf` already exists? In future this assumption may be broken.

fcami: Yes, this is expected, so that we can adapt the test.

ipatests/test_integration/test_installation.py
```diff
@@ -703,6 +703,30 @@ def install(cls, mh):
...
+
+        # set the IPA server in debug mode
+        server_conf = "[global]\ndebug=True"
+        self.master.put_file_contents(target_fname, server_conf)
+        assert self.master.transport.file_exists(target_fname)
```
stanislavlevin, line 717: Is it really needed? We put the file in previous action and assume `put_file_contents` does its job.

ipaserver/rpcserver.py
fcami: I kinda like the readability of the first version but the 2nd would fit too...

ipatests/test_integration/test_installation.py
fcami: This is not 100% needed. I can remove it if you like.

ipatests/test_integration/test_installation.py
```diff
@@ -703,6 +703,30 @@ def install(cls, mh):
...
+        self.master.run_command(["systemctl", "restart", "httpd"])
```
stanislavlevin, line 718: Could you please replace the hardcoded service name with `knownservices.httpd.systemd_name`?
in the assumption that tests controller's OS is the same as the worker's one or better to query it from worker's ipaplatform.

ipatests/test_integration/test_installation.py
```diff
@@ -703,6 +703,30 @@ def install(cls, mh):
...
+
+        # smoke test in debug mode
+        tasks.kdestroy_all(self.master)
+        tasks.kinit_admin(self.master)
+        self.master.run_command(["ipa", "user-show", "admin"])
+
+        # rollback
+        self.master.run_command(["rm", target_fname])
+        assert not self.master.transport.file_exists(target_fname)
+        self.master.run_command(["systemctl", "restart", "httpd"])
```
stanislavlevin, line 728: same for the service name.

ipatests/test_integration/test_installation.py
stanislavlevin: For example,
```python3
    result = host.run_command(                                                  
        [                                                                       
            "python3",                                                          
            "-c",                                                               
            (                                                                   
                "from ipaplatform.services import knownservices; "              
                "print(knownservices.named.systemd_name)"                       
            ),                                                                  
        ]                                                                       
    )                                                                           
    service_name = result.stdout_text.strip()                                   
    host.run_command(["systemctl", "restart", service_name])
```

ipatests/test_integration/test_installation.py
stanislavlevin: May be adapt without waiting for breakage? For example, check for existance and edit with `FileBackup` if exists.

ipatests/test_integration/test_installation.py
fcami: Sorry, no. I'll restart ipa instead.

stanislavlevin: It's ok :)

ipaserver/rpcserver.py
fcami: Fixed. Thanks.

ipatests/test_integration/test_installation.py
fcami: Removed, thanks.

ipatests/test_integration/test_installation.py
fcami: I've switched to restarting ipa.

ipatests/test_integration/test_installation.py
```diff
@@ -703,6 +703,29 @@ def install(cls, mh):
...
+        Related: https://pagure.io/freeipa/issue/8891
+        """
+
+        target_fname = paths.IPA_SERVER_CONF
+        assert not self.master.transport.file_exists(target_fname)
+
+        # set the IPA server in debug mode
+        server_conf = "[global]\ndebug=True"
+        self.master.put_file_contents(target_fname, server_conf)
+        self.master.run_command(["paths.IPACTL", "restart"])
```
stanislavlevin, line 717: should be not a string ;-)

fcami:
Temporarily Removing ACK and re-adding WIP

fcami:
hi @stanislavlevin could you please have a second look?

ipaserver/rpcserver.py
```diff
@@ -72,6 +73,10 @@
 if six.PY3:
     unicode = str
 
+# time.perf_counter_ns appeared in Python 3.7.
+if version_info < (3, 7):
+    time.perf_counter_ns = lambda: int(time.perf_counter() * 10**9)
```
stanislavlevin, line 78: Just a note: the version check is not necessary here, the `feature` check is preferred.

ipatests/test_integration/test_installation.py
```diff
@@ -703,6 +703,29 @@ def install(cls, mh):
...
+        Related: https://pagure.io/freeipa/issue/8891
+        """
+
+        target_fname = paths.IPA_SERVER_CONF
+        assert not self.master.transport.file_exists(target_fname)
+
+        # set the IPA server in debug mode
+        server_conf = "[global]\ndebug=True"
+        self.master.put_file_contents(target_fname, server_conf)
+        self.master.run_command([paths.IPACTL, "restart"])
```
stanislavlevin, line 717: The full restart of `ipa` takes ~1-2min, while the restart of `httpd` takes several seconds. We waste here ~2-4min.

ipatests/test_integration/test_installation.py
```diff
@@ -703,6 +703,29 @@ def install(cls, mh):
...
+
+        # smoke test in debug mode
+        tasks.kdestroy_all(self.master)
+        tasks.kinit_admin(self.master)
+        self.master.run_command(["ipa", "user-show", "admin"])
+
+        # rollback
+        self.master.run_command(["rm", target_fname])
+        assert not self.master.transport.file_exists(target_fname)
```
stanislavlevin, line 726: extra check here

stanislavlevin (commented):
Hi, @fcami. Thank you for the fix.
I'm not agree with the unnecessary wasting time: `ipactl restart` vs (query service name + `systemctl restart service`).

ipatests/test_integration/test_installation.py
fcami: Less than that if I can trust the [log](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/8848656a-d58d-11eb-99d1-fa163ed6a638/runner.log.gz) : 2x50s, so 1m40. It also makes sure the whole of ipa comes up in debug mode (bonus!). I find this much better than having to detect the running platform's httpd service (or making sure it uses systemd...).

ipatests/test_integration/test_installation.py
fcami: Removed. Thanks.

ipatests/test_integration/test_installation.py
stanislavlevin: > Less than that if I can trust the [log](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/8848656a-d58d-11eb-99d1-fa163ed6a638/runner.log.gz) : 2x50s, so 1m40. 

Azure log:
https://dev.azure.com/freeipa/freeipa/_build/results?buildId=9187&view=logs&j=fd5ef46c-a89d-5fee-07c3-d9f8082531c5&t=e993e2cc-4ff6-5b17-419c-da6d7496f246&l=844

> It also makes sure the whole of ipa comes up in debug mode (bonus!). I find this much better than having to detect the running platform's httpd service (or making sure it uses systemd...).

IPA doesn't support non-systemd platforms...

> Plus, ipactl is the abstraction for the platform, so it works everytime.

ipatests/test_integration/test_installation.py
fcami: Yeah, I saw that this morning. Thanks for the catch.

ipatests/test_integration/test_installation.py
fcami: > IPA doesn't support non-systemd platforms...

For now? FreeIPA went from non-systemd to systemd, who knows what the future holds. 
Anyway, I'd rather restart than having to detect how the httpd service is named on the targed host.

stanislavlevin: May be https://pagure.io/freeipa/issue/8620 more clearly explains what I'm trying to achieve.
If that PR will be merged then every service name will be resolved on a remote ipaplatform.
I don't see any problem to do it manually in this PR.


> Anyway, I'd rather restart than having to detect how the httpd service is named on the targed host.

Why?

rcritten: I also think that a full restart is overkill and this should revert to systemctl restart httpd.

While I agree in principal that service handling should be abstracted to not be system-dependent I don't think that it should block this individual change, particularly since it is a regression with older python releases This will just represent one more place to fix on that abstraction is in place.

fcami: > May be https://pagure.io/freeipa/issue/8620 more clearly explains what I'm trying to achieve.
> If that PR will be merged then every service name will be resolved on a remote ipaplatform.
> I don't see any problem to do it manually in this PR.

I appreciate what you are trying to achieve, but either it is done globally or not. 

> > Anyway, I'd rather restart than having to detect how the httpd service is named on the targed host.
> 
> Why?

Because "ipactl restart" works on all supported platforms.

fcami: > I also think that a full restart is overkill and this should revert to systemctl restart httpd.
> 
> While I agree in principal that service handling should be abstracted to not be system-dependent I don't think that it should block this individual change, particularly since it is a regression with older python releases This will just represent one more place to fix on that abstraction is in place.

@rcritten @stanislavlevin please let me know what you would like best: "ipactl restart", which is platform-independent and a complete smoke test (e.g. can we start and restart ipa in debug mode) but (I agree) slow, or "systemctl restart httpd" which is much faster but platform-dependent.

stanislavlevin: Neither of these. I proposed and propose resolve service name + systemctl restart resolved_name
The example of code for resolving _remote_ name is above (btw, copy-pasted from current IPA code).

stanislavlevin: In current IPA code there are many places with the same code snippet for resolving remote service name.

fcami: Flagging postponed.

fcami:
I've switched to running the test only if the runner is RHEL or Fedora. I'm open to have this test use platform-detection code in tasks.py or in the host class when available.

stanislavlevin:
Your current code skips the test if the controller's platform is RHEL or Fedora.

stanislavlevin:
There is no reason to pin this test to any platform because nothing special happens.
If you don't want to follow the suggestion for an unknown reason (I really don't understand it), of course, you can leave it as is but please, remove the skip marker, at least, I can catch the error and fix it for me myself in downstream. Thank you.

fcami:
Hi @stanislavlevin My reasoning is that platform detection code does not belong in the tests themselves. I see your existing PR ( https://github.com/freeipa/freeipa/pull/5656/ ) as the right way to do it.
Does skipping only on debian make sense to you?

stanislavlevin:
Hi, @fcami.
Thank you, I see your PoV :)

ofc I can't answer for Debian. But afaik, nobody runs integration tests in Debian.

stanislavlevin:
@fcami, thank you for putting up with me.
ACK from my side.

fcami:
Thanks @stanislavlevin !

rcritten:
LGTM ack

rcritten:
master:

* 675b5d2f0519af68e3455075b39b7c5e2897cc07 paths: add IPA_SERVER_CONF
* 69497f4a026ab0689f6276f6695ca14cdc7164d1 ipatests: smoke test for server debug mode.
* 06791db5fad62f32ddc3641f40997c1035a86f96 rpcserver.py: perf_counter_ns is Python 3.7+

PR #5855, opened 2021-06-24
ipatests: Test ipa-cert-fix warns when startup directive is missing from CS.cfg

mrizwan93:
Earlier it used to fail when startup directive missing from CS.cfg.
With https://github.com/dogtagpki/pki/pull/3466, it changed to display
a warning than failing.

related: https://pagure.io/freeipa/issue/8890

### DIFF
--- a/ipatests/test_integration/test_ipa_cert_fix.py
+++ b/ipatests/test_integration/test_ipa_cert_fix.py
@@ -205,7 +205,7 @@ class TestIpaCertFix(IntegrationTest):
 
     def test_missing_startup(self, expire_cert_critical):
         """
-        Test ipa-cert-fix fails when startup directive is missing from CS.cfg
+        Test ipa-cert-fix fails/warns when startup directive is missing
 
         This test checks that if 'selftests.container.order.startup' directive
         is missing from CS.cfg, ipa-cert-fix fails and throw proper error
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @mrizwan93 
please update the commit message with "Fixes: https://pagure.io/freeipa/issue/8890".

The test needs to have different expectations depending on the version of pki-server that is installed.
If pki-server < 10.11, the current code still applies, but if pki-server>=10.11 your patch applies. You can reuse the function tasks.get_pki_version(host) from [ipatests/pytest_ipa/integration/tasks.py](https://pagure.io/freeipa/blob/master/f/ipatests/pytest_ipa/integration/tasks.py).

ipatests/test_integration/test_ipa_cert_fix.py
```diff
@@ -228,11 +237,20 @@ def test_missing_startup(self, expire_cert_critical):
...
         err_msg2 = ("ERROR: CalledProcessError(Command "
                     "['pki-server', 'cert-fix'")
-        assert err_msg1 and err_msg2 in result.stderr_text
+        warn_msg = ("WARNING: No selftests configured in "
+                    f"{paths.CA_CS_CFG_PATH} "
+                    "(selftests.container.order.startup)")
+
+        if (tasks.get_pki_version(self.master)
+           < tasks.parse_version('10.11.0')):
+            assert err_msg1 and err_msg2 in result.stderr_text
```
flo-renaud, line 251: The above assert is equivalent to assert (err_msg1 is not None) and (err_msg2 in stderr), which is not what the test should check.

flo-renaud:
Hi @mrizwan93 
thanks for the patch, LGTM. You can remove the temp commit.
Should we also backport to ipa-4-9?

flo-renaud:
master:

* ea8f4b69a8f874548c9a6f17d0c99976aa28c291 ipatests: Test ipa-cert-fix warns when startup directive is missing from CS.cfg

flo-renaud:
Hi @mrizwan93 
a manual backport is required for ipa-4-9, can you handle it? Thanks

PR #5857, opened 2021-06-24
plugins: Don't treat keys of API as bytes

stanislavlevin:
The plugin `plugins` iterates over the keys of API instance, __iter__ of which is a generator of class.__name__ from 
(Command, Object, Method, Backend, Updater). So, the allowed type is str, not bytes.

- fixed `plugins` plugin
- added tests for `plugins` plugin
- added tests for `env` plugin
 
Fixes: https://pagure.io/freeipa/issue/8898

### DIFF
--- a/ipatests/test_xmlrpc/test_env_plugin.py
+++ b/ipatests/test_xmlrpc/test_env_plugin.py
@@ -1,0 +1,101 @@
+#
+# Copyright (C) 2021  FreeIPA Contributors see COPYING for license
+#
+"""Test `env` plugin
+"""
+
+import pytest
+
+from ipalib import api, errors
+from ipatests.test_xmlrpc.xmlrpc_test import XMLRPC_test
+
+
+@pytest.mark.tier1
+class TestEnv(XMLRPC_test):
+    """Test `env` plugin
+    """
+    EXPECTED_KEYS = ("result", "count", "total", "summary")
+
+    def run_env(self, *args, **options):
+        cmd = api.Command.env
+        cmd_result = cmd(*args, **options)
+        return cmd_result
+
+    def assert_result(self, cmd_result):
+        assert tuple(cmd_result.keys()) == self.EXPECTED_KEYS
+        result = cmd_result["result"]
+        assert isinstance(result, dict)
+
+        total_count = cmd_result["total"]
+        assert isinstance(total_count, int)
+
+        actual_count = cmd_result["count"]
+        assert isinstance(actual_count, int)
+        assert actual_count <= total_count
+        assert len(result) == actual_count
+
+        if actual_count > 1:
+            assert cmd_result["summary"] == f"{actual_count} variables"
+        else:
+            assert cmd_result["summary"] is None
+
+    @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


ipatests/test_xmlrpc/test_env_plugin.py
```diff
@@ -0,0 +1,101 @@
...
+        kwargs["server"] = server
+        cmd_result = self.run_env(*args, **kwargs)
+        self.assert_result(cmd_result)
+        result = cmd_result["result"]
+        assert result["in_server"] is server
+        assert cmd_result["count"] == 2
+
+    @pytest.mark.parametrize("server", [True, False], ids=["server", "local"])
+    def test_env_with_variables_missing_var(self, server):
+        cmd_result = self.run_env("notexistedvariable", server=server)
```
rcritten, line 93: nit: notexistedvariable -> nonexistentvariable

Similar later, notexistedoption -> nonexistentoption

rcritten:
I assume you noticed that env didn't have a basic test when you discovered that plugin also lacks a test, so added it?

For some future us it might be helpful to expand the commit message to mention this. I go back and forth on this even as I type but since env was introduced in like 2011 it makes sense to me to recognize how long its been missing.

That and a couple of minor nits that are common between the two patches.

Otherwise LGTM, thanks.

ipatests/test_xmlrpc/test_env_plugin.py
stanislavlevin: > I assume you noticed that env didn't have a basic test when you discovered that plugin also lacks a test, so added it?
> 
Unintentionally I see that `ipa plugins` doesn't work. I was looking for a place where I could add tests for `ipalib.misc`,  but nothing helpful was found, only CLI tests for `plugins` in `ipatests/test_cmdline/test_cli.py`. So, tests for `ipalib.misc` module have been added.

> For some future us it might be helpful to expand the commit message to mention this. I go back and forth on this even as I type but since env was introduced in like 2011 it makes sense to me to recognize how long its been missing.
> 
> That and a couple of minor nits that are common between the two patches.
> 
> Otherwise LGTM, thanks.

Thank you!

stanislavlevin:
Fixed spelling.

stanislavlevin:
It seems that `plugins` command is not very popular, no reports so far.

stanislavlevin:
historical notes which I can see from git:
`env` plugin was added in 138305b365307a30d151ff144d5a1e0d95723293(2008), no tests.

stanislavlevin:
`plugins` 3e70c3b56b29dcc9c0f6dd15eee7d4a24945944a(2008), no tests.

abbra:
Adding ACK based on @rcritten review. Thank you!

abbra:
master:

* cf80244375851dff01aaab467f7f24f555d6d778 plugins: Don't treat keys of api as bytes
* cc212e552fc61b5fedb83e6cbc81a958ada138b8 ipatests: Add tests for `plugins` plugin
* fe8635941634f21448123779a41e4d21827a99c9 ipatests: Add tests for `env` plugin

PR #5860, opened 2021-06-25
Don't return more attributes than requested from LDAP cache

rcritten:
On a hit in the LDAP cache the entire cached entry was returned.
This caused problems because more attributes could be returned
than requested.

The automember condition add/remove calls only request the
inclusive/exclusive rule attributes and loop over the returned
values to look for duplicates. This was failing because the queried
entry contains attributes that the candidate entry does not contain.
The automember code is:

    old_entry = ldap.get_entry(dn, [attr])
    for regex in old_entry.keys():
        if not isinstance(entry_attrs[regex], (list, tuple)):

old_entry, returned from the cache, contained objectclass, cn,
description, etc. which don't exist in the candidate entry so
entry_attrs[regex] threw a KeyError.

Instead generate a new entry from the cache consisting of the
requested attributes only if all of the requested attributes are
cached.

Also be more careful when storing the attribbutes in the cache entry.
The returned attributes may not match the requested. So store the
attributes we actually have.

This issue was exposed by Ansible which maintains a larger and
longer-lived cache because commands are executed in the server context
one after another, giving the cache a chance to build up.

https://pagure.io/freeipa/issue/8897

### DIFF
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -1802,6 +1802,22 @@ class LDAPCache(LDAPClient):
         if self._enable_cache:
             logger.debug(msg, *args, **kwargs)
 
+    def copy_entry(self, dn, entry, attrs=[]):
+        new_entry = LDAPEntry(self, DN(dn))
+
+        # Return either the whole entry or only those attrs requested
+        if not attrs:
+            new_entry.raw.update(deepcopy(dict(entry.raw)))
+        else:
+            for attr, original_values in entry.raw.items():
+                if attr.lower() not in attrs:
+                    continue
+                new_entry.raw[attr.lower()] = deepcopy(original_values)
+        new_entry._orig_raw = deepcopy(dict(entry.raw))
+        new_entry.reset_modlist()
+
+        return new_entry
+
     def add_cache_entry(self, dn, attrs_list=None, get_all=False,
                         entry=None, exception=None):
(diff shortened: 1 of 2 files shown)


abbra:
The change LGTM but tests need to be fixed to follow new returned results.

rcritten:
Re-running the failed Azure tests. I tried reproducing a couple of these locally and it worked fine.

stanislavlevin:
These errors are new unknown errors, most likely related to changes.

rcritten:
One is failing on kinit. That one for sure is definitely not related and has failed twice now.

stanislavlevin:
I reproduced locally at least `test_ipapython/test_ldap_cache.py` errors.

rcritten:
That is fixed. The Azure failures in the test tab did not point to this, only to other failures. This test was buried in the logs. Azure output leaves much to be desired.

stanislavlevin:
tests tab:
image

web logs:
image

Would pytest-html's report help?

rcritten:
That's not what it had initially. Originally it had three failed tests including test_ipa_cacert_manage_prune and test_proxycommand_invalid_shell. So this looked more like a false negative than anything wrong with the PR. 

In any case I've discovered the issue and am working on it.

abbra:
master:

* 19d5b3b621dbdfe96b290ac2f7af63008d01aa80 Return a copy of cached entries, only with requested attributes

PR #5868, opened 2021-06-30
Add SHA384withRSA as a certificate signing algorithm

rcritten:
It required support in dogtag which was added in 10.5.0.

This is only easily configurable during installation because
it will set ca.signing.defaultSigningAlgorithm to the
selected algorithm in CS.cfg

The certificate profiles will generally by default set
default.params.signingAlg=- which means use the CA default.

So while an existing installation will technically allow
SHA384withRSA it will require profile changes and/or
changing the defaultSigningAlgorithm in CS.cfg and
restarting (completely untested). And that won't affect
already issued-certificates.

https://pagure.io/freeipa/issue/8906


flo-renaud:
Hi @rcritten 
thanks for the PR, LGTM. I tested the following scenarios:
 - [x] installation of the new package, `ipa-server-install --ca-signing-algorithm SHA384withRSA ...`: install succeeds and the CA cert has the expected signing algo (for instance `certutil -L -d /etc/dirsrv/slapd-IPA-TEST/ -n Server-Cert | grep "Signature Algorithm"` properly displays `SHA-384 With RSA Encryption`)
 - [x]  installation of 4.9.6 with embedded CA, upgrade: no issue. Resubmitted certs still have the same signing algo (SHA-256 With RSA Encryption by default).
 - [x] installation of 4.9.6 CA-less, upgrade, `ipa-ca-install --ca-signing-algorithm SHA384withRSA`: new certs signed by IPA CA properly display the expected signing algo.

rcritten:
master:

* 02e19d0a396286d9de5a516f62f60699b5ee2527 Add SHA384withRSA as a certificate signing algorithm

PR #5871, opened 2021-07-05
stageuser: add ipauserauthtypeclass when required

flo-renaud:
### stageuser: add ipauserauthtypeclass when required
The command
ipa stageuser-add --user-auth-type=xxx
is currently failing because the objectclass ipauserauthtypeclass
is missing from the created entry.

There is code adding the missing objectclass in the
pre_common_callback method of user_add, and this code should
be common to user_add and stageuser_add. In order to avoid code
duplication, it makes more sense to move the existing code to
pre_common_callback of baseuser_add, that is called by both
classes.

Fixes: https://pagure.io/freeipa/issue/8909

### XMLRPC test: add a test for stageuser-add --user-auth-type

Related: https://pagure.io/freeipa/issue/8909

### DIFF
--- a/ipaserver/plugins/baseuser.py
+++ b/ipaserver/plugins/baseuser.py
@@ -539,6 +539,9 @@ class baseuser_add(LDAPCreate):
         if entry_attrs.get('ipatokenradiususername', None):
             add_missing_object_class(ldap, u'ipatokenradiusproxyuser', dn,
                                      entry_attrs, update=False)
+        if entry_attrs.get('ipauserauthtype', None):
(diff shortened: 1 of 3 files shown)


abbra:
LGTM

rcritten:
Me too. ack.

flo-renaud:
master:

* ec8b26b2c42d798926986c2fccb6ec9aa683f63f stageuser: add ipauserauthtypeclass when required
* 05d69c8c325215778645bf8172ce6f8fc746bd1d XMLRPC test: add a test for stageuser-add --user-auth-type

PR #5885, opened 2021-07-11
ipatests: use whole date when calling journalctl --since

flo-renaud:
The test TestSelfExternalSelf::test_switch_back_to_self_signed
is checking the content of the journal using journalctl --since ...
but provides only the time, not the whole date with year-month-day.
As a consequence, if the test is executed around midnight it may
find nothing in the journal because it's looking for logs after 11:50PM,
which is a date in the future.
Fixes: https://pagure.io/freeipa/issue/8918


amore17 (approved):

flo-renaud:
master:

* eb8a6d078235400fdf3ba8c66102217425237196 ipatests: use whole date when calling journalctl --since

PR #5889, opened 2021-07-12
Fall back to krbprincipalname when validating host auth indicators

rcritten:
When adding a new host the principal cannot be determined because it
relies on either:

a) an entry to already exist
b) krbprincipalname be a component of the dn

As a result the full dn is being passed into ipapython.Kerberos
which can't parse it.

Look into the entry in validate_validate_auth_indicator() for
krbprincipalname in this case.

https://pagure.io/freeipa/issue/8206

### DIFF
--- a/ipaserver/plugins/service.py
+++ b/ipaserver/plugins/service.py
@@ -209,6 +209,11 @@ def validate_auth_indicator(entry):
     # and shouldn't be allowed to have auth indicators.
     # https://pagure.io/freeipa/issue/8206
     pkey = api.Object['service'].get_primary_key_from_dn(entry.dn)
+    if pkey == str(entry.dn):
+        # krbcanonicalname may not be set yet if this is a host entry,
+        # try krbprincipalname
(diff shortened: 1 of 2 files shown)


flo-renaud:
Hi @rcritten 
the fix works for me. It should be straightforward to add an XMLRPC test in test_host_plugin.py with creation of a host + auth-indicator.

rcritten:
Right, I don't know why I didn't include the test :-( Coming soon.

rcritten:
Added test test_xmlrpc/test_host_plugin.py::TestProtectedMaster:: test_add_non_master_with_auth_ind

flo-renaud:
Hi @rcritten 
thanks for the test, WFM. The test without the patch fails as expected, and the test + the patch succeeds. ACK for when PRCI completes

rcritten:
master:

* bd0d43745072ec7976207c231cba8411efb41e17 Fall back to krbprincipalname when validating host auth indicators

PR #5902, opened 2021-07-16
webui tests: fix algo for finding available idrange

flo-renaud:
The webui tests for ID range evaluate a potentially free id range
by looking for existing ranges and picking a range = max value
\+ 1 million.

With the addition of subuid range this algorithm produces values
over the limit because the subuid range goes from
2,147,483,648 to 4,294,836,224 and the max base id is 4,294,967,295.

Ignore the subuid range when picking a potential range.
Fixes: https://pagure.io/freeipa/issue/8919

### DIFF
--- a/ipatests/test_webui/task_range.py
+++ b/ipatests/test_webui/task_range.py
@@ -64,6 +64,12 @@ class range_tasks(UI_driver):
         max_rid = 0
 
         for idrange in idranges:
+            # IPA.TEST_subid_range is automatically created near the end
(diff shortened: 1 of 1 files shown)


flo-renaud:
For future ref, link to successful run: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/0731555e-e63b-11eb-ab93-fa163e390eb7)
Removing temp commit

rcritten:
master:

* 319aa7a4b1fd1640d21f721fb8cf8534b64b6428 webui tests: fix algo for finding available idrange

PR #5923, opened 2021-07-26
ipa-advise: if p11-kit provides opensc, don't add to NSS db

rcritten:
p11-kit-proxy in newer distributions handles loading the OpenSC
PKCS#11 library so don't try to add it to the NSS database in
/etc/pki/nssdb if it is already available in order to avoid a
potentially confusing error message.
    
https://pagure.io/freeipa/issue/8934
    
NOTE:

1. A second patch fixes script error I noticed while working on the ticket.
2. There are no tests because it relies completely on the OS release and state of the client and we don't have that capacity in CI. I propose manual tests.

### DIFF
--- a/ipaserver/advise/plugins/smart_card_auth.py
+++ b/ipaserver/advise/plugins/smart_card_auth.py
@@ -149,7 +149,7 @@ class config_server_for_smart_card_auth(common_smart_card_auth_config):
         self.log.exit_on_predicate(
             '[ -z "$ipaca_records" ]',
             [
-                'Can not resolve ipa-ca records for ${domain_name}',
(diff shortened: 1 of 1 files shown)


flo-renaud:
@rcritten thanks for the patch, works for me. Tested on f34 with opensc module present or absent, both cases work.
I agree that manual testing is enough.

rcritten:
master:

* e13bd7ed0e62b9812c4189163105beca4391d4aa ipa-advise: if p11-kit provides opensc, don't add to NSS db
* 9d1ee6443487d0200c57541b29111f91af3651fd ipa-advise: Define the domain used when looking up ipa-ca

PR #5933, opened 2021-07-29
Add index for sudoorder

rcritten:
sudorule-mod <rule> --order=<num> does a search for an existing
order and this search is unindexed.

https://pagure.io/freeipa/issue/8939


flo-renaud:
@rcritten thanks for the patch, works for me:
- upgrade correctly adds the index
- fresh install also configures the index

flo-renaud:
master:

* 4785d87ea9092c7cf3c2136d0803d8125e54d459 Add index for sudoorder

PR #5935, opened 2021-07-30
Don't assume that plugin attributes and objectclasses are lowercase

rcritten:
Don't assume that plugin attributes and objectclasses are lowercase

A user wrote their own plugin to add custom attributes which was
failing with an incorrect error that the attribute wasn't allowed.

It wasn't allowed because it wasn't being treated as case-insensitive
so wasn't being found in the schema.

https://pagure.io/freeipa/issue/8415

**NOTE**: the new test file is not yet integrated into PR-CI. I think this is fine testing as part of the nightlies and not as gating but I'm open to suggestions.

### DIFF
--- a/ipaserver/plugins/config.py
+++ b/ipaserver/plugins/config.py
@@ -534,14 +534,14 @@ class config_mod(LDAPUpdate):
                     checked_attrs = checked_attrs + [self.api.Object[obj].uuid_attribute]
                 for obj_attr in checked_attrs:
                     obj_attr, _unused1, _unused2 = obj_attr.partition(';')
-                    if obj_attr in OPERATIONAL_ATTRIBUTES:
+                    if obj_attr.lower() in OPERATIONAL_ATTRIBUTES:
                         continue
-                    if obj_attr in self.api.Object[obj].params and \
+                    if obj_attr.lower() in self.api.Object[obj].params and \
                       'virtual_attribute' in \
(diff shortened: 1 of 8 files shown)


flo-renaud:
Hi @rcritten 
the test and the fix lgtm (the test correctly detects the issue when executed on a server without the fix, and test is green with the fix).
I agree with you, no need to add the new test into gating, but can you include it in ipatests/prci_definitions/ in the nightly defs for nightly_latest_selinux.yaml, nightly_latest_testing_selinux.yaml, nightly_latest_testing.yaml, nightly_latest.yaml, nightly_previous.yaml, and nightly_rawhide.yaml?

rcritten:
@flo-renaud Ok cool, glad we agree. I've dropped the temp commit and added the new test to the nightly runs.

flo-renaud:
@rcritten the yaml definitions LGTM, thanks

flo-renaud:
master:

* 4932a9c98264d177b5811caa618c04e7490b249b Don't assume that plugin attributes and objectclasses are lowercase
* 7b77fef75a5010aa9ae37be6a1355885fb73692c ipatests: add suite for testing custom plugins
* e3304ff3aab20001ec8c0dafcdd57ac5053b102f pr-ci definitions: add custom plugin-related jobs

flo-renaud:
@rcritten a manual backport is required for ipa-4-9 (because of the test definitions), can you handle it? thanks

PR #5936, opened 2021-07-30
Provide more information in ipa-certupdate on ccache failure

rcritten:
ipa-certupdate obtains host credentials to operate. If this
fails with a ccache error this can be confusing if the user
executing it already has admin credentails.

Include the principal being retrieved and the keytab being
used.

This basically intercepts the exception to log additional
information and lets the exception be handled at a higher
level.

https://pagure.io/freeipa/issue/8257

I've only reproduced this with contrived use-cases but when I did run into it during development debugging the root cause was difficult so providing this information can be useful. As for test automation, I think it's overkill.

### DIFF
--- a/ipaclient/install/ipa_certupdate.py
+++ b/ipaclient/install/ipa_certupdate.py
@@ -31,7 +31,7 @@ from ipaplatform import services
 from ipaplatform.paths import paths
 from ipaplatform.tasks import tasks
 from ipalib import api, errors, x509
-from ipalib.constants import IPA_CA_NICKNAME, RENEWAL_CA_NAME
+from ipalib.constants import FQDN, IPA_CA_NICKNAME, RENEWAL_CA_NAME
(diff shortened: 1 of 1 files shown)


flo-renaud:
@rcritten 
thanks for the PR, LGTM. Tested the following scenario:
```
[root@fedora test]# kinit admin
Password for admin@IPA.TEST: 
[root@fedora test]# mv /etc/krb5.keytab /etc/krb5.keytab.sav
[root@fedora test]# ipa-certupdate
Unable to obtain credentials for server.ipa.test from /etc/krb5.keytab
did not receive Kerberos credentials
The ipa-certupdate command failed.
```
The command now displays the principal and keytab location, it will be really helpful when debugging.

rcritten:
master:

* 3568a3d1e7d8931286fdf77edefbd2f751d937d0 Provide more information in ipa-certupdate on ccache failure

PR #5939, opened 2021-08-02
ipaserver: disable resolved' stub resolver

fcami:
Disable systemd-resolved stub resolver at install time.
Use systemd-resolved' maintained list of upstream DNS servers instead.

Rationale: systemd-resolved always resolves the FQDN to the local IP
and vice-versa. This breaks DNS zone detection and especially reverse
zone detection. This results in --auto-reverse being broken.

On systemd-resolved enabled systems, there are four ways to
configure resolv.conf:
* a symlink to /run/systemd/resolve/stub-resolv.conf
    This is the default and uses both the 127.0.0.53 DNS stub plus
    the search domains.
* a symlink to /usr/lib/systemd/resolv.conf
    This only contains the 127.0.0.53 DNS stub.
* a symlink to /run/systemd/resolve/resolv.conf
    This contains the upstream DNS IPs and bypasses systemd-resolved.
* Create and maintain /etc/resolv.conf directly.

Solutions #1 and #2 break DNS zone detection.
Solution #4 is not straightforward.
Combine Solution3 and 4: copy the file containing the upstream DNS IPs
so that it is properly maintained by NetworkManager after installation.

Fixes: https://pagure.io/freeipa/issue/8700

+

ipatests: xfail TestInstallWithCA_DNS3
    
TestInstallWithCA_DNS3 cannot succeed in PR-CI due to libvirt's
dnsmasq auto-maintaining the reverse zone for the networks it belongs
to. Mark the test as xfail on Fedora.
    
Related: https://pagure.io/freeipa/issue/8700

### DIFF
--- a/ipaplatform/redhat/tasks.py
+++ b/ipaplatform/redhat/tasks.py
@@ -29,6 +29,7 @@ import ctypes
 import logging
 import os
 from pathlib import Path
+import shutil
 import socket
 import traceback
 import errno
@@ -774,5 +775,60 @@ class RedHatTaskNamespace(BaseTaskNamespace):
         authselect_cmd = [paths.AUTHSELECT, "disable-feature",
                           "with-custom-automount"]
         ipautil.run(authselect_cmd)
+
+    def disable_resolved(self, sstore):
+        """
+        https://pagure.io/freeipa/issue/8700
+        https://www.freedesktop.org/software/systemd/man/systemd-resolved.service.html
+        Combine the third and fourth methods to manage /etc/resolv.conf:
+        Copy the resolv.conf file containing the upstream DNS servers
+        into /etc/resolv.conf so that systemd-resolved does not interfere
+        with reverse zone detection (--auto-reverse).
+        """
+        if os.path.islink(paths.RESOLV_CONF):
+            resolv_conf_link_target = os.readlink(
+                paths.RESOLV_CONF
+            )
+            # the target can be any of:
+            # /run/systemd/resolve/stub-resolv.conf
+            # ../run/systemd/resolve/stub-resolv.conf
+            if paths.RESOLV_CONF_STUB_RESOLVED in resolv_conf_link_target:
+                sstore.backup_state(
+                    "installation",
+                    "resolv-conf-links-to-stub-resolved",
+                    resolv_conf_link_target
+                )
+                logger.info(
+                    'Disabling systemd-resolved.'
+                )
+                os.unlink(paths.RESOLV_CONF)
+                shutil.copyfile(
+                    paths.RESOLV_CONF_RESOLVED,
+                    paths.RESOLV_CONF
+                )
+
+    def reenable_resolved(self, sstore):
+        """
+        The reverse of disable_resolved above.
+        """
+        resolv_conf_link_target = sstore.get_state(
+            "installation",
+            "resolv-conf-links-to-stub-resolved"
+        )
+        if resolv_conf_link_target:
+            if os.path.isfile(paths.RESOLV_CONF):
+                logger.info(
+                    'Re-enabling systemd-resolved.'
+                )
+                os.unlink(paths.RESOLV_CONF)
+                os.symlink(
+                    resolv_conf_link_target,
+                    paths.RESOLV_CONF
+                )
+            sstore.delete_state(
+                "installation",
+                "resolv-conf-links-to-stub-resolved"
+            )
+
 
 tasks = RedHatTaskNamespace()
--- a/ipaserver/install/dns.py
+++ b/ipaserver/install/dns.py
@@ -42,6 +42,8 @@ from ipaserver.install import dnskeysyncinstance
 from ipaserver.install import odsexporterinstance
 from ipaserver.install import opendnssecinstance
 from ipaserver.install import service
+from ipaplatform.tasks import tasks
+
 
 if six.PY3:
     unicode = str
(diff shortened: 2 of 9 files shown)


fcami:
From the log:
```
2021-08-02T17:56:33Z DEBUG IP address 192.168.122.140 belongs to a private range, using forward policy only
2021-08-02T17:56:33Z DEBUG Checking DNS server: 192.168.122.1
2021-08-02T17:56:34Z DEBUG will use DNS forwarders: [CheckedIPAddressLoopback('192.168.122.1')]

2021-08-02T17:56:34Z INFO Reverse record for IP address 192.168.122.140 already exists
```
Using libvirt's 192.168.122.1 will always succeed (e.g. there will always be a reverse entry) I think.
@netoarmando do we have a way out in PR-CI?

netoarmando:
> @netoarmando do we have a way out in PR-CI?

For the record, PR-CI used to set a custom  `/etc/resolv.conf`. It was removed at https://github.com/freeipa/freeipa-pr-ci/commit/4b034af13ec7c0fee4b842dc1a59ce6f90b057fb.

Maybe @wladich has some input on this.

fcami:
I would like the resolv.conf to point to something we can control, like a BIND instance on the runner, configured with appropriate forwarders... 192.168.122.1 is not representative of what to expect in real envs :/

fcami:
The libvirt-managed dnsmasq on the host builds the reverse zone from DHCP leases. So by default it will manage that reverse zone, always breaking --auto-reverse.

The only way out I see is to configure either BIND or dnsmasq on the runner, configure forwarders, and use that as upstream resolver on the test instances. Perhaps using https://galaxy.ansible.com/bertvv/bind - it has all the features needed.

ipaserver/install/dns.py
```diff
@@ -143,6 +145,16 @@ def install_check(standalone, api, replica, options, hostname):
...
 
+    # https://pagure.io/freeipa/issue/8700
+    # https://www.freedesktop.org/software/systemd/man/systemd-resolved.service.html
+    # Temporary copy the resolv.conf file containing the upstream DNS servers
+    # into /etc/resolv.conf so that systemd-resolved does not interfere with
+    # reverse zone detection.
+    if os.path.islink(paths.RESOLV_CONF):
+        if os.readlink(paths.RESOLV_CONF) == paths.RESOLV_CONF_STUB_RESOLVED:
+            os.unlink(paths.RESOLV_CONF)
+            shutil.copyfile(paths.RESOLV_CONF_RESOLVED, paths.RESOLV_CONF)
```
antoniotorresm, line 156: Would it be possible to update the link to `paths.RESOLV_CONF_RESOLVED` instead of removing it and copying the file?

fcami: That was my first idea and is the 3rd way to configure systemd-resolved.

However, systemd-resolved then merges the upstream DNS list with 127.0.0.1 (the FreeIPA DNS provided via a NetworkManager configuration snippet).
The resulting resolv.conf contains something like:
```
nameserver 127.0.0.1
nameserver UPSTREAM_DNS_1
nameserver UPSTREAM_DNS_2
domain FOO
```
which is somewhat brittle and misconfigured.
This is why I decided on the fourth way e.g. provide a resolv.conf that works for us during install.

antoniotorresm:
LGTM.

flo-renaud:
pylint detected the following issues:
```
************* Module ipaserver.install.dns
ipaserver/install/dns.py:124: [E1101(no-member), install_check] Instance of 'BasePathNamespace' has no 'SYSRESTORE_DIR_PATH' member)
ipaserver/install/dns.py:149: [E1101(no-member), install_check] Module 'ipaplatform.tasks' has no 'disable_resolved' member)
************* Module ipaserver.install.server.install
ipaserver/install/server/install.py:1201: [E1101(no-member), uninstall] Instance of 'BaseTaskNamespace' has no 'restore_resolv_conf' member)
```

fcami:
@antoniotorresm @flo-renaud just FYI my goal is now to refactor the change into ipaplatform and make sure the system is properly returned to its original state after uninstall.

fcami:
Tested with:
* a standalone server installed with DNS
* a standalone server installed without DNS, then DNS was added
* a replica with DNS
Gating is green with the patches from https://github.com/freeipa/freeipa/pull/5983 .

fcami:
hi @antoniotorresm and @rcritten 
I've pushed a new version that handles all types of symlinks (relative and absolute).

For testing, this requires using a DNS that replies NXDOMAIN to the tested VM's IP.
Google DNS: 8.8.8.8 / 8.8.4.4 work fine for that purpose.

ipaserver/install/server/install.py
```diff
@@ -1198,6 +1198,7 @@ def uninstall(installer):
     tasks.restore_hostname(fstore, sstore)
     tasks.restore_pkcs11_modules(fstore)
     fstore.restore_all_files()
+    tasks.reenable_resolved(sstore)
```
rcritten, line 1201: restore_all_files() should be called last IIRC

ipaplatform/redhat/tasks.py
```diff
@@ -775,4 +776,59 @@ def disable_ldap_automount(self, statestore):
...
+        with reverse zone detection (--auto-reverse).
+        """
+        if os.path.islink(paths.RESOLV_CONF):
+            resolv_conf_lnk_tgt = os.readlink(
+                paths.RESOLV_CONF
+            )
+            # the target can be any of:
+            # /run/systemd/resolve/stub-resolv.conf
+            # ../run/systemd/resolve/stub-resolv.conf
+            if paths.RESOLV_CONF_STUB_RESOLVED in resolv_conf_lnk_tgt:
```
rcritten, line 795: bike shedding: lnk_tgt -> link_target

tgt makes my brain think TGT

fcami: Fixed, ty.

ipaserver/install/server/install.py
fcami: Fixed, ty.

fcami:
Note to self: this should also be done on upgrades.

fcami:
Closing, the latest systemd updates seem to have fixed this:
http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/0ce093e6-49f7-11ec-9ecc-fa163e3ef39f/report.html

PR #5951, opened 2021-08-04
Test semi-generic way to pull in external repo

rcritten:
Testing in the context of https://github.com/freeipa/freeipa/pull/5932

I'm not yet sure that setup_class() is the best way to do this and IIRC DNS resolution can be tricky but let's see what happens.

### DIFF
--- a/ipatests/test_integration/base.py
+++ b/ipatests/test_integration/base.py
@@ -24,6 +24,9 @@ import subprocess
 
 from ipatests.pytest_ipa.integration import tasks
 from pytest_sourceorder import ordered
+from ipatests.pytest_ipa.integration.env_config import get_global_config
+
+config = get_global_config()
 
 
 @ordered
@@ -39,6 +42,18 @@ class IntegrationTest:
     topology = None
     domain_level = None
     fips_mode = None
+
+    @classmethod
+    def setup_class(cls):
+        hosts = config.get_all_hosts()
+        for host in hosts:
+            if host.role not in ('master', 'replica'):
+                continue
+
+            host.run_command(
+                ['dnf', '-y', 'copr', 'enable', 'ckelley/pki']
+            )
+            host.run_command(['dnf', '-y', 'update', 'pki-ca'])
 
     @classmethod
     def host_by_role(cls, role):
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
--- a/install/tools/ipa-pki-wait-running.in
+++ b/install/tools/ipa-pki-wait-running.in
@@ -13,6 +13,7 @@ import logging
 import sys
 import time
 from xml.etree import ElementTree
+import json
 
 from ipalib import api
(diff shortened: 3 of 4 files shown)


ipatests/test_integration/base.py
```diff
@@ -40,6 +40,19 @@ class IntegrationTest:
     domain_level = None
     fips_mode = None
 
+    @classmethod
+    def setup_class(cls):
+        cls.master.run_command(
+            ['dnf', '-y', 'copr', 'enable', 'ckelley/pki']
+        )
+        cls.master.run_command(['dnf', '-y', 'update', 'pki-core'])
+
```
fcami, line 49: @rcritten @ckelleyRH Note for later: if this depends on a current 389DS it might be best to update it here too.
The current images are up-to-date at the moment so that should not be an issue now but it could.

rcritten: While a general solution would be nice, I'm specifically just trying to hack this so we can test both ways.

Unfortunately the host configuration (cls.master, cls.replicas, etc) are done after setup_class() is called.

rcritten:
Ok my approach seems super hacky but...

More importantly, we have a passing test @ckelleyRH . How do we tell whether this worked because of the JSON or your fallback to XML?

ckelleyRH:
@rcritten - nice! There is a new log message [here](https://github.com/freeipa/freeipa/pull/5932/files#diff-698fa8f795049ff756b137a90a0d3b85748bbfdbada72fc092920d5f998bd428R86), I tried to find it in your test logs but have no idea where to start!

ckelleyRH:
@rcritten ...and we shouldn't be able to find it if you are including the patch in these CI runs. We should be able to see it in the CI for https://github.com/freeipa/freeipa/pull/5932 though right?

rcritten:
The message logs at DEBUG but the logging level is hardcoded at INFO so this will never appear in the journal. Temporarily bump it to DEBUG so we can see the message.

We may want to make it info level instead long-term.

rcritten:
Running with update pki sources shows that the JSON output is properly handled, and confirmed that it isn't falling back (via lack of message) from http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/782e8584-f5e9-11eb-9526-525400b889ed/test_integration-test_installation.py-TestInstallWithCA1-install/master.ipa.test/journal.gz

Aug 05 12:43:04 master.ipa.test ipa-pki-wait-running[19907]: urllib3.connectionpool: Starting new HTTP connection (10): master.ipa.test:8080
Aug 05 12:43:05 master.ipa.test ipa-pki-wait-running[19907]: urllib3.connectionpool: http://master.ipa.test:8080 "GET /ca/admin/ca/getStatus HTTP/1.1" 200 129
Aug 05 12:43:05 master.ipa.test ipa-pki-wait-running[19907]: ipa-pki-wait-running: Got status 'running', error 'None'
Aug 05 12:43:05 master.ipa.test ipa-pki-wait-running[19907]: ipa-pki-wait-running: Success, subsystem ca is running!

I'm going to close this PR as it demonstrates that with the patch IPA will interoperate with the new CA code once it lands. In the related PR we just need to show the fallback message, perhaps via a similar temporary commit, and I think we can merge.

PR #5955, opened 2021-08-05
Add index for dnahostname

fcami:
There are 60+ searches for:
Filter:   (dnahostname=FQDN)
at startup.

Fixes: https://pagure.io/freeipa/issue/8945

### DIFF
--- a/install/updates/20-indices.update
+++ b/install/updates/20-indices.update
@@ -107,6 +107,13 @@ default:nsSystemIndex: false
 add:nsIndexType: eq
 add:nsIndexType: sub
 
+dn: cn=dnahostname,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
+only:cn: dnahostname
+default:objectClass: nsIndex
+default:objectClass: top
+default:nsSystemIndex: false
+add:nsIndexType: eq
+
 dn: cn=fqdn,cn=index,cn=userRoot,cn=ldbm database,cn=plugins,cn=config
 only:cn: fqdn
 default:objectClass: nsIndex


fcami:
@rcritten and @tiran could you please have a look?

rcritten:
LGTM. Want to get @tiran 's input before acking since he knows the use-case better.

tiran:
Is the missing index a performance issue? Last time I checked the attribute was only queried during installation, update, and removal of a server.

https://github.com/freeipa/freeipa/blob/e3304ff3aab20001ec8c0dafcdd57ac5053b102f/install/updates/20-indices.update#L37-L38

rcritten:
The investigation of this started with an error displayed during the upgrade step at the end of installation:

Could not get dnaHostname entries in 60 seconds

In the server install log:

2021-08-05T17:29:16Z DEBUG Found DNA config cn=Subordinate IDs,cn=Distributed Numeric Assignment Plugin,cn=plugins,cn=config
2021-08-05T17:29:16Z DEBUG dnaSharedCfgDN: cn=subordinate-ids,cn=dna,cn=ipa,cn=etc,dc=example,dc=test
2021-08-05T17:29:16Z DEBUG Got only one entry. Retry again in 2 sec.
2021-08-05T17:29:18Z DEBUG Got only one entry. Retry again in 2 sec.
...
etc.

tiran:
I wonder if the issue is related to a missing index. The DNA plugin creates two entries for each DNA instance and host. In my (limited) experience the entries are created shortly after DS is restarted. I can't say why it takes so long sometimes and if the changeset is going to fix the problem.

@mreynolds389, what do you think?

mreynolds389:
> I wonder if the issue is related to a missing index. The DNA plugin creates two entries for each DNA instance and host. In my (limited) experience the entries are created shortly after DS is restarted. I can't say why it takes so long sometimes and if the changeset is going to fix the problem.
> 
> @mreynolds389, what do you think?

This sort of reminds me of an issue where if a plugin adds an index, that index is not "enabled" right away.  It takes a db2index task, or a server restart, for it to start working and updating the index files.  A quick test would be to run "dbscan" on the index file and see if it has entry keys & id's that match the existing entries.  You can also turn on plugin logging and internal operation log level to see the DNA operations and their results:

    nsslapd-plugin-logging: on
    nsslapd-accesslog-level: 260

Now in newer versions of DS if you set the systemIndex attribute to true then the index is enabled immediately:

    nsSystemIndex: true

Anyway if your case is reproducible then try reindexing dnaHostname before the DNA stuff, or restarting the server after adding the new index.  If that is all correct, and it's still not working as expected then you should grab that logging I mentioned above as it might provide more insight into the issue.

fcami:
@mreynolds389 the issue was present without the index. I haven't been able to reproduce it with this index but since it was intermittent I'm not 100% certain this was really fixed.
I think @tiran was asking whether this was really an indexing issue (that would be fixed by adding the index in this current PR) or something else.

mreynolds389:
> @mreynolds389 the issue was present without the index. I haven't been able to reproduce it with this index but since it was intermittent I'm not 100% certain this was really fixed.
> I think @tiran was asking whether this was really an indexing issue (that would be fixed by adding the index in this current PR) or something else.

There is just not enough info in this ticket to determine that.  I would be very interested in seeing the logging I suggested above.  It will at least rule out indexing, and provide more info as to what the plugin is doing.  

You could even add "plugin" error logging (nsslapd-errorlog-level: 65536), but that will absolutely slow the server down, and if the problem is random then that verbose logging is probably useless.

PR #5956, opened 2021-08-05
[Backport][ipa-4-9] Don't assume that plugin attributes and objectclasses are lowercase

rcritten:
Manual backport of #5935

Only the nightly PR-CI configuration needed to be updated.

Don't assume that plugin attributes and objectclasses are lowercase

A user wrote their own plugin to add custom attributes which was
failing with an incorrect error that the attribute wasn't allowed.

It wasn't allowed because it wasn't being treated as case-insensitive
so wasn't being found in the schema.

https://pagure.io/freeipa/issue/8415

Signed-off-by: Rob Crittenden rcritten@redhat.com

### DIFF
--- a/ipatests/prci_definitions/nightly_ipa-4-9_previous.yaml
+++ b/ipatests/prci_definitions/nightly_ipa-4-9_previous.yaml
@@ -1713,3 +1713,15 @@ jobs:
         template: *ci-ipa-4-9-previous
         timeout: 3600
         topology: *master_1repl
+
+  fedora-previous-ipa-4-9/test_custom_plugins:
+    requires: [fedora-previous-ipa-4-9/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-previous-ipa-4-9/build_url}'
+        test_suite: test_integration/test_custom_plugins.py
+        template: *ci-ipa-4-9-previous
+        timeout: 3600
+        topology: *master_1repl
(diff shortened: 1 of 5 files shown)


ipatests/prci_definitions/nightly_ipa-4-9_previous.yaml
```diff
@@ -1713,3 +1713,16 @@ jobs:
...
         topology: *master_1repl
+
+  fedora-previous-ipa-4-9/test_custom_plugins:
+    requires: [fedora-previous-ipa-4-9/build]
+    priority: 50
+    job:
+      class: RunPytest
+      args:
+        build_url: '{fedora-previous-ipa-4-9/build_url}'
+        selinux_enforcing: True
```
flo-renaud, line 1724: The other tests for 4-9-previous don't run in selinux enforcing mode, this line should be removed.

rcritten: Fixed, too quick copy-pasta.

flo-renaud:
Thanks, LGTM

rcritten:
ipa-4-9:

* 97a2a925348d3bd732e582108feb02d644ba011a Don't assume that plugin attributes and objectclasses are lowercase
* e28e45402c7edb007e356a59cf09ed8e10cd14d9 ipatests: add suite for testing custom plugins
* 78c48199782743e619463cefa7411817f4fe4a14 pr-ci definitions: add custom plugin-related jobs

PR #5959, opened 2021-08-06
Fix string check in uninstall helper

fcami:
The install helpers used an invalid string check. ``('ubuntu')`` is
not a tuple. It's a string with superfluous parenthesis. A single-item
tuple would be ``('ubuntu',)``. It's recommended to use set literals to
avoid such mistakes.

Also check for 'debian' platform.

Fixes: https://pagure.io/freeipa/issue/8937


flo-renaud:
master:

* c75be14bee9ded79b92cb0f92eb6c903b7b6c6dc Fix string check in uninstall helper

PR #5964, opened 2021-08-10
ipatests: Fix TestAJPSecretUpgrade tests on systems without pkiuser

stanislavlevin:
Tests in `test_ipaserver.test_secure_ajp_connector' assume that there
is pkiuser in OS, but this is not always true (for example, in systems
having minimum installed dependencies, in particular, without pki-server
RPM package). Since the tests already use the mock and pkiuser entity is
not the subject of testing the pwd.getpwnam has been mocked.

Fixes: https://pagure.io/freeipa/issue/8942

### DIFF
--- a/ipatests/test_ipaserver/test_secure_ajp_connector.py
+++ b/ipatests/test_ipaserver/test_secure_ajp_connector.py
@@ -1,5 +1,6 @@
 # Copyright (C) 2021  FreeIPA Project Contributors - see LICENSE file
 
+from collections import namedtuple
 from io import BytesIO
 from lxml.etree import parse as myparse  # pylint: disable=no-name-in-module
 import pytest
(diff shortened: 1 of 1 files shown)


rcritten:
test_certmonger_ipa_responder_jsonrpc failed to verify that the certmonger request used JSON. I have the feeling the "since" value was too late. Full report http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/8d71187a-f9b9-11eb-9fb5-5254007adcf0/report.html

Since unrelated to this PR re-running.

rcritten:
@flo-renaud what do you think about the --since failure? Does this sound related to the recent change?

flo-renaud:
@rcritten 
> test_certmonger_ipa_responder_jsonrpc failed to verify that the certmonger request used JSON. I have the feeling the "since" value was too late. Full report http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/8d71187a-f9b9-11eb-9fb5-5254007adcf0/report.html
> 
> Since unrelated to this PR re-running.

The `since` value contains only the time, not the whole date, and this sometimes leads to issues (`cls.since = time.strftime('%H:%M:%S')`) for instance when since = "23:59:00" and the journal is checked a few minutes later (--since 23:59:00 would be a time in the future if the current time is 00:02:00). I don't think it was the case in this test (--since=09:12:51 is not around midnight).

The second potential issue is the time synchronization. The `since` time is taken from the controller, not from the master. If they are not in sync we may see a small delay and miss a few entries.

Third point, many tests have a comment about processes not flushing immediately the logs into the journal and they add a small delay or a loop before checking for the journal's content. I think it's probably the root cause here.

So anyway, the failure in `test_certmonger_ipa_responder_jsonrpc` seems completely unrelated to this change, and the re-run produced a clean status.

rcritten:
I opened https://pagure.io/freeipa/issue/8953 to track the test failure.

rcritten:
LGTM. @stanislavlevin do you need this backported to ipa-4-9 or is the master branch sufficient?

stanislavlevin:
Thank you, @rcritten. I'm sorry, forgot to tag on.

rcritten:
master:

* bb628aebaf714421c4b54ce87b559e8f20731a7e ipatests: Fix TestAJPSecretUpgrade tests on systems without pkiuser

PR #5970, opened 2021-08-11
ipatests: test_ipahealthcheck: Test if check for FIPS is present

miskopo:
Test if the check for FIPS mode status is present in ipahealthcheck.meta.core
   
Resolves: https://pagure.io/freeipa/issue/8951
    
### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -504,6 +504,26 @@ class TestIpaHealthCheck(IntegrationTest):
         assert data[0]["kw"]["ipa_version"] in result.stdout_text
         assert data[0]["kw"]["ipa_api_version"] in result.stdout_text
 
+    def test_fips_mode_check_exists(self):
+        """
+        Testcase to verify whether FIPS mode check exists
+        in ipahealthcheck.meta.core
+        """
+        returncode, data = run_healthcheck(self.master,
+                                           "ipahealthcheck.meta.core",
+                                           "MetaCheck")
+        assert returncode == 0
+
+        cmd = self.master.run_command(['fips-mode-setup', '--is-enabled'],
+                                      raiseonerr=False)
+        fips_enabled = cmd.returncode == 0
+        for check in data:
+            if check["kw"].get("fips", None):
+                if fips_enabled:
+                    assert check["kw"]["fips"] == "enabled"
+                else:
+                    assert check["kw"]["fips"] == "disabled"
+
     def test_source_ipahealthcheck_ipa_host_check_ipahostkeytab(self):
         """
         Testcase checks behaviour of check IPAHostKeytab in source
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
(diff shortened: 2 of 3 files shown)


ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -504,6 +504,18 @@ def test_source_ipahealthcheck_meta_core_metacheck(self):
         assert data[0]["kw"]["ipa_version"] in result.stdout_text
         assert data[0]["kw"]["ipa_api_version"] in result.stdout_text
 
+    def test_fips_mode_check_exists(self):
+        """
+        Testcase to verify whether FIPS mode check exists
+        in ipahealthcheck.meta.core
+        """
+        returncode, data = run_healthcheck(self.master,
+                                           "ipahealthcheck.meta.core")
```
flo-renaud, line 513: Currently there is only one check in "ipahealthcheck.meta.core" but this could change in the future. You can restrict the checks to "MetaCheck" only.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -504,6 +504,18 @@ def test_source_ipahealthcheck_meta_core_metacheck(self):
...
+        assert returncode == 0
+        for check in data:
+            if check["kw"].get("fips", None):
```
flo-renaud, line 516: If the check doesn't return any value for "fips", the test will be successful. Is it really what you want? I would rather check explicitly that there is a value, and also check that the returned value is consistent with the system.
There are multiple ways to check if the system is fips-enabled:
- the command fips-mode-setup --is-enabled
- the content of the file /proc/sys/crypto/fips_enabled
- the output of `sysctl crypto.fips_enabled`

flo-renaud (commented):
Hi @miskopo 
please find inline comments

miskopo:
@flo-renaud thank you, I've addressed your comments.

ipatests/test_integration/test_ipahealthcheck.py
rcritten: I suspect it would be worthwhile to also execute this test with fips_mode = True

miskopo: I suspect you are right, Rob. I'll extend the test.

flo-renaud: @miskopo if you want to run in fips mode, you can have a look at the test in `test_integration/test_fips.py` that is faking fips mode.

miskopo: To be perfectly candid, I am not sure how to proceed. From my understanding, the base class in `test_fips` only fakes FIPS mode by returning `true` to `fips_enabled` queries.

rcritten: Right. It might be handy for pr-ci to have a fips option in the YAML but for now I think, since this is a small and very-confined test, to duplicate it in test_fips. The FIPS version can be simplified since you can assert that healthcheck returns fips=True. You may need to do additional setup to ensure the healthcheck package is installed. This is normally handled in the healthcheck test class. I'd add a cross-reference to each one so we know that there is some duplication, and why.

miskopo: Okay, thank you, Rob. I'll try to duplicate relevant bits and see what can be done. Also, with the FIPS mode - I think I may be able to implement FIPS param support to PR-CI.

miskopo:
Abandoned in favour of https://github.com/freeipa/freeipa/pull/6368

PR #5975, opened 2021-08-13
schema plugin: Fix commands without metaobject argument

stanislavlevin:
- schema plugin: fix commands without metaobject arg
- command_defaults: don't crash on nonexistent command
- test_schema_plugin: drop dependency on Tracker
- test_schema_plugin: add missing tests for command, class and topic commands

Fixes: https://pagure.io/freeipa/issue/8954


stanislavlevin:
rebased.

stanislavlevin:
Ready for review, if somebody wants.

flo-renaud:
@stanislavlevin Thanks for the PR. This is the type of review where one scratches his head and finally finds it was obvious when all the pieces fall into place...
command_find, topic_find, class_find, param_find, output_find all inherit from BaseMetaSearch but only param_find or output_find need to get the command as additional argument in the execute() method. So the right place for the execute() function  is indeed BaseParamSearch (only output_find and param_find inherit from this class). Kudos!

stanislavlevin:
@flo-renaud, I'm sorry if I wasn't descriptive enough.
Thank you for the review!

flo-renaud:
master:

* 7706dfaf09ac1c9acc1e711c67a1ea42d5113b6d schema plugin: Fix commands without metaobject arg
* 210739c0ed7731a6509b90021660e403b5338c72 command_defaults: Don't crash on nonexistent command
* d4a3ed9dd4313a8dd13e615b6d03622fcc4d2bbc test_schema_plugin: Drop dependency on Tracker
* c0a2b377917a165cb7f8e248b8f5ed2b8386897f test_schema_plugin: Add missing tests for command, class and topic commands

PR #5979, opened 2021-08-16
schema plugin: Generate stable fingerprint

stanislavlevin:
If some Param defines several values for `exclude` or `include` attributes then API schema hash will be unstable.
    
First, these Param's attributes are converted to frozenset (`ipalib.parameters`), then `ipaserver.plugins.schema` plugin converts `exclude` and `include` attrs to list. Set/frozenset in turn, is unordered collection \[0\]. So, the end order of values is undefined.

But due to the nature of sets:
> two sets are equal if and only if every element of each set is contained in the other (each is a subset of the other)
    
the order of values can be ignored.
    
Note: other Param's attrs with type frozenset are not affected because they are not processed by the schema plugin.
    
\[0\]: https://docs.python.org/3/library/stdtypes.html#set-types-set-frozenset
    
Fixes: https://pagure.io/freeipa/issue/8955

### DIFF
--- a/ipaserver/plugins/schema.py
+++ b/ipaserver/plugins/schema.py
@@ -839,7 +839,7 @@ class schema(Command):
         langs = "".join(getattr(context, "languages", []))
 
         if getattr(self.api, "_schema", None) is None:
-            setattr(self.api, "_schema", {})
+            object.__setattr__(self.api, "_schema", {})
 
         schema = self.api._schema.get(langs)
         if schema is None:
--- a/ipaserver/plugins/schema.py
+++ b/ipaserver/plugins/schema.py
@@ -592,7 +592,7 @@ class param(BaseParam):
                 obj[key] = unicode(value)
             elif key in ('exclude',
                          'include'):
-                obj[key] = list(unicode(v) for v in value)
+                obj[key] = sorted(list(unicode(v) for v in value))
             if isinstance(metaobj, Command):
                 if key == 'alwaysask':
                     obj.setdefault(key, value)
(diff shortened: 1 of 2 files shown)


stanislavlevin:
forgot, dev mode.

ipaserver/plugins/schema.py
```diff
@@ -839,7 +839,7 @@ def execute(self, *args, **kwargs):
         langs = "".join(getattr(context, "languages", []))
 
         if getattr(self.api, "_schema", None) is None:
-            setattr(self.api, "_schema", {})
+            object.__setattr__(self.api, "_schema", {})
```
flo-renaud, line 842: Any reason for using `object.__setattr__` instead of `setattr` ? `self.api` is an instance of `ipalib.API` which inherits from `ReadOnly`, it means that the method  `ReadOnly.__setattr__` is skipped if I understand correctly (the method checks if the attribute is locked).

stanislavlevin: Xmlrpc tests are run in `development` mode, while API is locked only for this mode(for performance reasons). So, no chance to pass as it was.
`object.__setattr__` is a legitimate way to say API: `I know what I do`.
`api._schema` is a cache, it cannot be read-only.

flo-renaud: Thanks for the explanation, makes sense.

rcritten: Exactly this. The whole ReadOnly bit was to prevent developers from being too clever by half and twiddling with things in the API in a running server. We left the setattr method alone since sometimes it is necessary and stands out like a sore thumb so its **obvious** when we are twiddling things :-)

We used to do this enforcement in "production" mode too but the startup cost is too high.

flo-renaud:
master:

* 3eb0759a6d501348cf581e433f195aba3e548f5a schema plugin: Generate stable fingerprint
* 77e7ac3592c0a708aa496275fa21e18bd6d01c1d ipatests: Add tests for `schema` Command

stanislavlevin:
Thank you, @flo-renaud!

PR #5983, opened 2021-08-19
Azure: Run pycodestyle check in Lint job

stanislavlevin:
- previously, fastlint make's target includes both the Pylint task
and pycodestyle one. The purpose of this target is a fast checking
only for changed Python files. This makes sense for pycodestyle, but
limits Pylint due to a context(file) checking. The clients which
call the code being linted are not checked at all. In Azure Pylint
(for the whole codebase) is run in the Lint task, this makes fastlint
extra for Azure.

- `Quick code style check` task used distro's Pylint, while `Lint`
task PyPI's one. This may cause different results and confuse a
user.

- `Build` task takes time longer than `Lint` one, so this change
doesn't lead to increased CI time.

- all Azure tests depend on Build and Lint tasks. Mostly it's no need
to run tests due to a probably broken code.

Fixes: https://pagure.io/freeipa/issue/8961

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -230,6 +230,8 @@ fasttest: $(GENERATED_PYTHON_FILES) ipasetup.py
 	    --ignore $(abspath $(top_srcdir))/ipatests/test_xmlrpc
 
 fastcodestyle: $(GENERATED_PYTHON_FILES) ipasetup.py
+	@ # keep Python files in sync to pycodestyle configuration in
+	@ # tox.ini(filename=)
 	@echo "Fast code style checking with $(PYTHON) from branch '$(GIT_BRANCH)'"
 
 	@MERGEBASE=$$(git merge-base --fork-point $(GIT_BRANCH)); \
@@ -250,8 +252,8 @@ fastcodestyle: $(GENERATED_PYTHON_FILES) ipasetup.py
 	    echo -e "Fast code style checking for files:\n$${FILES}\n"; \
 	    echo "pycodestyle"; \
 	    echo "-----------"; \
-	    git diff -U0 $${MERGEBASE} | \
-	        $(PYTHON) -m pycodestyle --diff || exit $$?; \
+	    git diff -U0 $${MERGEBASE} -- $${FILES} | \
+	        $(PYTHON) -m pycodestyle -v --diff || exit $$?; \
 	else \
 	    echo "No modified Python files found"; \
 	fi
--- a/Makefile.am
+++ b/Makefile.am
@@ -216,7 +216,7 @@ endif
 	$(MAKE) $(AM_MAKEFLAGS) acilint apilint polint pylint jslint $(RPMLINT_TARGET) yamllint check
 	@echo "All tests passed."
 
-.PHONY: fastcheck fasttest fastlint
+.PHONY: fastcheck fasttest fastlint fastcodestyle
 fastcheck:
 	@$(MAKE) -j1 $(AM_MAKEFLAGS) fastlint $(RPMLINT_TARGET) yamllint fasttest apilint acilint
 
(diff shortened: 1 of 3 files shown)


stanislavlevin:
- added Pycodestyle check for Python '.in' scripts

Makefile.am
```diff
@@ -216,7 +216,7 @@ endif
 	$(MAKE) $(AM_MAKEFLAGS) acilint apilint polint pylint jslint $(RPMLINT_TARGET) yamllint check
 	@echo "All tests passed."
 
-.PHONY: fastcheck fasttest fastlint
+.PHONY: fastcheck fasttest fastlint fastcodestyle
 fastcheck:
 	@$(MAKE) -j1 $(AM_MAKEFLAGS) fastlint $(RPMLINT_TARGET) yamllint fasttest apilint acilint
```
rcritten, line 221: Shouldn't fastcheck run fastcodestyle instead of fastlint as well?

rcritten:
If I understand this correctly this is moving CI away from using pylint because it's been proven time and time again to be annoying because of bugs and differences in release cadences.

Do we need/want the fastlint target at all anymore?

stanislavlevin:
Current Azure has fastlint(pylint + pycodestyle over modified Python files) + lint tasks.
fastlint one utilizes Pylint from distro(Fedora), while lint task - PyPI.
This PR proposes to get rid of fastlint check in Azure. The Lint task is still there.

stanislavlevin:
As for fastlint target. I can't say for all, maybe someone uses it locally.

stanislavlevin:
My point is leaving fastlint as is, it's extra only for Azure CI.

rcritten:
Ok, what about fastcheck? It runs some of the unit tests and also some linting. I use this from time to time for sanity checks before submitting. Having this run the codestyle instead of lint would be helpful since linting is of little use, as you point out.

stanislavlevin:
Pylint over modified files is the subset of Pylint over the project, it catches errors too but not all it can. It's a tradeoff between a fast check(several seconds) and a long check(several minutes).

I think that `fastcheck` should include all the possible checks to prevent wasting common CI time. Though, personally, I run my Azure CI before submitting PR.

Another inconvenience with fastlint is that it uses distro's Pylint by default, so, a developer has to have Pylint's version in sync with CI's one by installing from PyPI.

In other words, I don't use fastcheck at all, if you want to replace fastlint with fastcodestyle I'll just follow :)

ipatests/azure/azure-pipelines.yml
```diff
@@ -20,12 +20,6 @@ jobs:
   steps:
     - template: templates/${{ variables.PREPARE_BUILD_TEMPLATE }}
```
flo-renaud, line 21: Wouldn't it work if we only call here:
- template: templates/${{ variables.PREPARE_LINT_TEMPLATE }}
and make no other change? This way we force installation of pylint respecting `--constraint .wheelconstraints.in` and we have the same version of pylint in this pipeline and in the "Lint" pipeline

stanislavlevin: Yes, it would work.
But as I pointed out in https://github.com/freeipa/freeipa/commit/7855c77f93e0bf326e0307bce7072d5c12915659
pip can install everything needed for Pylint with the help of `--ignore-installed` (https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-I) even a dependency has been resolved in system site-packages.

stanislavlevin: I think it's the primary goal of that option and that's ok for our usage.

stanislavlevin: Not all IPA Python scripts use isolated mode, so it's maybe dangerous.

stanislavlevin:
@rcritten, can we leave fastcheck as is for this PR? I'm afraid It may be a disputable decision.

rcritten:
I'm ok with this change. The fastcheck lint has caught me in the past where it passes locally and fails in CI because of differing versions. I'll just have to re-train my fingers to run the codestyle checker instead.

stanislavlevin:
any other questions?

fcami (approved):
hi @stanislavlevin 
The changes LGTM. 
Tested OK with a few changes that are currently broken ( https://github.com/freeipa/freeipa/pull/5939 ). 
@flo-renaud were your concerns fully addressed?

flo-renaud:
>  @flo-renaud were your concerns fully addressed?

Yes, I'm OK with the change.

stanislavlevin:
Thank you all for your review!

fcami:
master:

* 42067e5610174d6457eae90e52234bdd7036f763 Azure: Run pycodestyle check in Lint job
* 3f134fea38a1938d523c5a59370d778139be4458 pycodestyle: Check *.in Python files

PR #5985, opened 2021-08-19
Increase default limit on LDAP searches to 100k

rcritten:
A similar change was attempted years ago in commit
9724251292e4c0797367fcc351a9f16f30c6aefe but it was
never applied because it used the wrong DN and because
nsslapd-timelimit is already present in the entry
the default keyword won't trigger.

Use replace instead to increase the value to 100k from
the default as originally intended.

https://pagure.io/freeipa/issue/8962

### DIFF
--- a/install/updates/10-config.update
+++ b/install/updates/10-config.update
@@ -16,7 +16,7 @@ only: nsslapd-pluginPrecedence: 60
 # Set limits to suite better IPA deployment sizes, defaults are too
 # conservative
 dn: cn=config
-default: nsslapd-sizelimit:100000
+replace: nsslapd-sizelimit:2000::100000
 
 dn: cn=config,cn=ldbm database,cn=plugins,cn=config
 replace: nsslapd-lookthroughlimit:5000::100000
--- a/ipapython/ipaldap.py
+++ b/ipapython/ipaldap.py
@@ -765,6 +765,7 @@ class LDAPClient:
         'nsslapd-logging-hr-timestamps-enabled': True,
         'nsslapd-ldapientrysearchbase': True,
         'nsslapd-ldapidnmappingbase': True,
+        'nsslapd-sizelimit': True,
     })
 
     time_limit = -1.0   # unlimited


install/updates/10-config.update
```diff
@@ -15,10 +15,8 @@ only: nsslapd-pluginPrecedence: 60
 
 # Set limits to suite better IPA deployment sizes, defaults are too
 # conservative
-dn: cn=config
-default: nsslapd-sizelimit:100000
-
-dn: cn=config,cn=ldbm database,cn=plugins,cn=config
```
flo-renaud, line 21: nsslapd-lookthroughlimit and nsslapd-idlistscanlimit are still documented as attributes of cn=config,cn=ldbm database,cn=plugins,cn=config, not of cn=default instance config,cn=chaining database,cn=plugins,cn=config.

rcritten: You're right, I should have left it in cn=config,cn=ldbm database,cn=plugins,cn=config. I got too cute in consolidating the changes

install/updates/10-config.update
```diff
@@ -15,10 +15,8 @@ only: nsslapd-pluginPrecedence: 60
...
+dn: cn=default instance config,cn=chaining database,cn=plugins,cn=config
+replace: nsslapd-sizelimit:2000::100000
```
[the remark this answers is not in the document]
rcritten: It's not, https://access.redhat.com/documentation/en-us/red_hat_directory_server/10/html/administration_guide/finding_directory_entries says it should be in cn=config. I think what I did was look in dse.ldif and found rather than looking in the docs.

rcritten:
I have the feeling I searched for an alternate place to set the size limit because in cn=config trying to set it using ipa-ldap-update results in a LDAP error 16, no such attribute. I'm guessing this has some special treatment by 389. I'll check with them.

flo-renaud:
@rcritten thanks for the new patch, works for me.
I tested a fresh install, the parameter is properly set, as well as an upgrade from 4.9.6, the parameter is also properly set.

rcritten:
master:

* b91c0afd0b5ae9eb4954078f8c76f2914daa1d85 Increase default limit on LDAP searches to 100k

PR #5991, opened 2021-08-25
extdom: return LDAP_NO_SUCH_OBJECT if domains differ

sumit-bose:
If a client sends a request to lookup an object from a given trusted
domain by UID or GID and an object with matching ID is only found in a
different domain the extdom should return LDAP_NO_SUCH_OBJECT to
indicate to the client that the requested ID does not exists in the
given domain.

Resolves: https://pagure.io/freeipa/issue/8965


rcritten:
Temporarily removing the ack while we determine where/if this will land in downstream releases.

rcritten:
master:

* d743219a9ae8a0ec9978dcbdd81eb89b7fd707f4 extdom: return LDAP_NO_SUCH_OBJECT if domains differ

PR #6001, opened 2021-09-02
subid: subid-match: display the owner's ID not DN

fcami:
Previously, the subid-match command would output the full
DN of the owner of the matched range.
With this change, the UID of the owner is displayed, just like
for other subid- commands.

Fixes: https://pagure.io/freeipa/issue/8977


rcritten:
LGTM. I didn't see any other post_callback that was also missing this.

flo-renaud:
master:

* b624179cb18d445c27abef067540fa9e914eacc5 subid: subid-match: display the owner's ID not DN

PR #6007, opened 2021-09-07
azure: Ignore tar errors

stanislavlevin:
Sometimes tar fails on changed in process files:
```
[2021-09-07 11:03:33] + tar --ignore-failed-read -czf ipaserver_install_logs.tar.gz --warning=no-failed-read /var/log/dirsrv /var/log/httpd2 /var/log/ipa /var/log/ipaclient-install.log /var/log/ipa-custodia.audit.log /var/log/ipaserver-install.log /var/log/krb5kdc.log /var/log/pki /var/log/samba /var/lib/bind/data systemd_journal.log
[2021-09-07 11:03:33] tar: Removing leading `/' from member names
[2021-09-07 11:03:33] tar: Removing leading `/' from hard link targets
[2021-09-07 11:03:33] tar: /var/log/dirsrv/slapd-IPA-TEST/access: file changed as we read it
[2021-09-07 11:03:33] + tests_result=1
```

This is expected failure since processes are not stopped during logs
collection and can flush their logs.

Fixes: https://pagure.io/freeipa/issue/8983

### DIFF
--- a/ipatests/azure/scripts/azure-run-base-tests.sh
+++ b/ipatests/azure/scripts/azure-run-base-tests.sh
@@ -22,7 +22,8 @@ function collect_logs() {
         /var/log/pki \
         /var/log/samba \
         "$BIND_DATADIR" \
-        systemd_journal.log
+        systemd_journal.log \
+        ||:
 }
 
 server_password=Secret123
@@ -95,6 +96,8 @@ if [ "$install_result" -eq 0 ] ; then
 else
     echo "ipa-server-install failed with code ${install_result}, skip IPA tests"
 fi
+# let the services gracefully flush their logs
+ipactl stop ||:
 collect_logs ipaserver_install_logs.tar.gz
 
 echo "Potential Python 3 incompatibilities in the IPA framework:"


stanislavlevin:
not related to azure failures, re-run.

abbra:
LGTM. Will add ack label once the remaining test is done.

stanislavlevin:
The failure is already tracked in https://pagure.io/freeipa/issue/8870
re-run

rcritten:
Just a question, I have no real objection to the proposed path. Is a simpler solution to run ipactl stop? I assume at this point that we aren't losing something useful in the buffered tail of the 389-ds access log?

stanislavlevin:
@rcritten, sorry, I missed your question.

I thought about this and didn't add it for uniformity with integration tests.

In Base Task of Azure CI the logs are collected twice (on tests completion and IPA uninstallation), while in integration tests the logs are collected on each method/class completion. A service stop/start takes a long time and in addition, may lead to harmful side effects.

So, in this particular case(for base tests), I think it's possible to _try_ to stop all the processes the logs we are interested in to let them flush their logs (on tests completion) because this happens only once.

rcritten:
@stanislavlevin I'm ok with the patch as-is. If we run into a situation where we miss logging because of a lack of flush we can address it then. I'm not aware of such a situation to-date. Adding the ack per Alexander's review.

stanislavlevin:
btw, for most integration tests environments `nsslapd-accesslog-logbuffering` is set to `off`, but not for `base` ones.

flo-renaud:
master:

* a941e8f602c9171ac9b1a7324f7ace495a712624 azure: Ignore tar errors

PR #6008, opened 2021-09-07
migrate-ds: workaround to detect compat tree

flo-renaud:
Migrate-ds needs to check if compat tree is enabled before
migrating users and groups. The check is doing a base
search on cn=compat,$SUFFIX and considers the compat tree
enabled when the entry exists.

Due to a bug in slapi-nis, the base search may return NotFound
even though the compat tree is enabled. The workaround is to
perform a base search on cn=users,cn=compat,$SUFFIX instead.

Fixes: https://pagure.io/freeipa/issue/8984

### DIFF
--- a/ipaserver/plugins/migration.py
+++ b/ipaserver/plugins/migration.py
@@ -922,7 +922,8 @@ migration process might be incomplete\n''')
         # check whether the compat plugin is enabled
         if not options.get('compat'):
             try:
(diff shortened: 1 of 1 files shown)


abbra:
LGTM

flo-renaud:
master:

* 2ba3028b968f8d3c27980c9921c0c08b29d024e9 migrate-ds: workaround to detect compat tree

PR #6023, opened 2021-09-16
ipatests: fix TestInstallWithCA_DNS3's resolv.conf

fcami:
TestInstallWithCA_DNS3 uses the Resolved resolver class which
does not configure resolv.conf correctly.

Unconfigure NetworkManager first and then reconfigure systemd-resolved.

Fixes: https://pagure.io/freeipa/issue/8993

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -570,6 +570,11 @@ class TestInstallWithCA_DNS3(CALessBase):
     ticket 7239
     """
 
+    @pytest.mark.skipif(
+        osinfo.id == 'fedora',
+        reason='https://pagure.io/freeipa/issue/8700',
+        strict=True
+    )
     @server_install_setup
     def test_number_of_zones(self):
         """There should be two zones: one forward, one reverse"""
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -462,7 +462,7 @@ class TestInstallCA(IntegrationTest):
 
         # Tweak sysrestore.state to drop installation section
         self.master.run_command(
-            ['sed','-i', r's/\[installation\]/\[badinstallation\]/',
+            ['sed', '-i', r's/\[installation\]/\[badinstallation\]/',
              os.path.join(paths.SYSRESTORE, SYSRESTORE_STATEFILE)])
 
         # Re-run installation check and it should fall back to old method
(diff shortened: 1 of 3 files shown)


fcami:
Expected bad result, showing that the resolv.conf configuration is now right:
http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/581717d2-179c-11ec-9aeb-fa163eda45c5/report.html

fcami:
hi @sorlov-rh could you please have a look?

miskopo:
> hi @sorlov-rh could you please have a look?
I think he use @wladich

wladich:
We have discussed the PR with @fcami, topics discussed:

- provide more descriptive commit messages and code comments (what is not working now, how it is fixed)
- move NM and resolved config management to `create_broken_resolv_conf()`
- detect presence of systemd-resolved using tasks.resolver classes
- run tests on top of #5939 without `skip`
- think about cleanup for config changes

ipatests/test_integration/test_installation.py
```diff
@@ -570,22 +570,90 @@ class TestInstallWithCA_DNS3(CALessBase):
...
+            )
+            for netif in result.stdout_text.splitlines():
+                self.master.run_command(
+                    ["resolvconf", "-d", f"{netif}"]
+                )
 
-        result = self.master.run_command([
-            'ipa', 'dnszone-find'])
+            # force using 127.0.0.2
+            self.master.run_command(
```
wladich, line 614: In current version this is already done by lines
```
master.resolver.setup_resolver('127.0.0.2')
```

It does it in a more preferred way by dropping config file in `/etc/systemd/resolved.conf.d/`

wladich:
As far as I understand, what the nm-script and successive commands are doing is preventing NetworkManager from pushing DNS servers obtained via DHCP into systemd-resolved config. According to https://developer-old.gnome.org/NetworkManager/stable/NetworkManager.conf.html this can be achived by utilizing `systemd-resolved` config option:
```
printf "[main]\nsystemd-resolved = false" > /etc/NetworkManager/conf.d/zzzz-ipatests-disable-auto-dns.conf
resolvconf -d 2 # need to get list of interfaces first
systemctl restart NetworkManager
systemctl restart systemd-resolved
```

And the cleanup is
```
rm /etc/NetworkManager/conf.d/zzzz-ipatests-disable-auto-dns.conf
systemctl restart NetworkManager
systemctl restart systemd-resolved
```

What do you think about it?

rcritten:
Can you update the commit message with more details on what isn't being configured correctly and what the result of that misconfiguration is?

fcami:
Closing as not needed.

PR #6028, opened 2021-09-22
selinux: Fix file context definition for /var/run

vmojzis:
There is a file context equivalence rule assigning /run the same
contexts as /var/run. Because of it it's necessary to use /var/run
instead of /run in file context definitions.

See:
https://fedoraproject.org/wiki/SELinux/IndependentPolicy#File_contexts_and_equivalency_rules

Signed-off-by: Vit Mojzis <vmojzis@redhat.com>

### DIFF
--- a/selinux/ipa.fc
+++ b/selinux/ipa.fc
@@ -23,7 +23,7 @@
 
 /var/log/ipareplica-conncheck.log.*	--	gen_context(system_u:object_r:ipa_log_t,s0)
 
-/run/ipa(/.*)?              gen_context(system_u:object_r:ipa_var_run_t,s0)
+/var/run/ipa(/.*)?              gen_context(system_u:object_r:ipa_var_run_t,s0)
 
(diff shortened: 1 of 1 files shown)


fcami:
Adding re-run to kick PR-CI off

rcritten:
@vmojzis  This is equivalent as of what version of selinux-policy? Do we need to set a minimum package version?

vmojzis:
> @vmojzis This is equivalent as of what version of selinux-policy? Do we need to set a minimum package version?

The equivalency was introduced in 2011, selinux-policy-3.9.16-11 (Fedora-16), so no need.

rcritten:
This change looks ok, and basically reverts part of bdf11371693936e4ff6cc4d0e245b19493df564c which is in 4.9.0+. I'm just wondering why we haven't noticed until now.

rcritten:
master:

* cd85b729d3fd03e6acd75ec4f0f916aec4bc9247 selinux: Fix file context definition for /var/run

PR #6031, opened 2021-09-27
Improve sudooption docs, make the option multi-value

rcritten:
I don't know why this wasn't always multi-value but if one wanted
to set multiple options they needed to call add-option multiple
times. The LDAP attribute is already multi-value.

This shouldn't cause API issues as it understood the attribute as
multi-value just didn't expose it. Client output on the CLI will
look a bit different:

Added option "('one', 'two')" to Sudo Rule "test"

or

Added option "(u'one', u'Two')" to Sudo Rule "test"

instead of with this change:

Added option "one,two" to Sudo Rule "test"

https://pagure.io/freeipa/issue/2278

### DIFF
--- a/API.txt
+++ b/API.txt
@@ -5569,7 +5569,7 @@ command: sudorule_add_option/1
 args: 1,5,3
 arg: Str('cn', cli_name='sudorule_name')
 option: Flag('all', autofill=True, cli_name='all', default=False)
-option: Str('ipasudoopt', cli_name='sudooption')
+option: Str('ipasudoopt+', cli_name='sudooption')
 option: Flag('no_members', autofill=True, default=False)
 option: Flag('raw', autofill=True, cli_name='raw', default=False)
 option: Str('version?')
@@ -5724,7 +5724,7 @@ command: sudorule_remove_option/1
 args: 1,5,3
 arg: Str('cn', cli_name='sudorule_name')
 option: Flag('all', autofill=True, cli_name='all', default=False)
-option: Str('ipasudoopt', cli_name='sudooption')
+option: Str('ipasudoopt+', cli_name='sudooption')
 option: Flag('no_members', autofill=True, default=False)
 option: Flag('raw', autofill=True, cli_name='raw', default=False)
 option: Str('version?')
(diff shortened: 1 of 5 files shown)


flo-renaud:
Hi @rcritten 
You will need to bump VERSION.m4 and call makeapi to re-generate the API file.
Works for me but I have a question: the method `ipa sudorule-remove-option` hasn't been modified and doesn't offer the possibility of removing multiple sudooptions in a single command. Was it by choice?

rcritten:
I have no idea why I didn't enable multiple option deletes as well as adds. Patch updated.

flo-renaud:
Hi @rcritten 
The output displayed when only one option is added/removed is wrong:
```
# ipa sudorule-add-option srule --sudooption opt1
-------------------------------------------
Added option "o,p,t,1" to Sudo Rule "srule"                      <<<< HERE the single option is broken into multiple chars
-------------------------------------------
  Rule name: srule
  Enabled: TRUE
  Sudo Option: opt1
```

The LDAP entry is properly updated, though.

rcritten:
It's the join() in the output_for_cli override. In the server plugin the sudooption it is a tuple but in the client, for a single value, it's a simple string. Will need to dig into ipalib to see if this is something we should change globally or restrict to this plugin.

rcritten:
Ok, in `ipalib/frontend.py::__do_call()` this fixes the data type:

`params = self.normalize(**params)`

The client still has the single-value view unless multiple options were passed in. It seems like a heavy lift to change the client view so I'll handle this individually in the `output_for_cl`i methods.

flo-renaud:
@rcritten thanks for the patch update, works for me. Adding or removing a single option now has the expected output:
```
[root@replica ~]# ipa sudorule-add-option srule --sudooption opt4
----------------------------------------
Added option "opt4" to Sudo Rule "srule"
----------------------------------------
  Rule name: srule
  Enabled: TRUE
  Sudo Option: opt2, opt3, opt4
[root@replica ~]# ipa sudorule-remove-option srule --sudooption opt4
--------------------------------------------
Removed option "opt4" from Sudo Rule "srule"
--------------------------------------------
  Rule name: srule
  Enabled: TRUE
  Sudo Option: opt2, opt3
[root@replica ~]# 
```

flo-renaud:
master:

* 47fbe05f7fe81a90e420286dbda22cfc5a21f078 Improve sudooption docs, make the option multi-value

PR #6033, opened 2021-09-28
Design: Integrate SID configuration into base IPA installers

flo-renaud:
Add design doc for the feature.

Related: https://pagure.io/freeipa/issue/8995


doc/designs/adtrust/sidconfig.md
```diff
@@ -0,0 +1,287 @@
...
+### UI
+
+No new UI for server/replica installation.
+
+`IPA server / ID ranges` tab is already available and displays Primary and 
+Secondary RID base.
+
+Open questions:
+- should we display the SID identifier / NetBIOS Name in
+`IPA Server / Configuration` tab instead of `IPA Server / Trusts / Global Trust`
```
abbra, line 150: We can show them in both places, there is no problem with that. It would, actually, make sense to have that information available through both views.

doc/designs/adtrust/sidconfig.md
```diff
@@ -0,0 +1,287 @@
...
+Configuration tab?
+- how can we show that SID generation is enabled / not enabled in the WebUI?
```
abbra, line 152: SID generation is enabled if there are domain SID and NetBIOS name defined for IPA domain. That's the only requirement beyond a user's own SID be present in `ipaNTSecurityIdentifier`.

On user view we show SMB attributes if they exist. Perhaps, we can also add a note if there is no `ipaNTSecurityIdentifier` in the user's entry to warn admins that they need to generate SID for this user?

doc/designs/adtrust/sidconfig.md
```diff
@@ -0,0 +1,287 @@
...
+| Command |	Options |
+| --- | ----- |
+| ipa-server-install | --netbios-name=NETBIOS_NAME --rid-base=RID_BASE --secondary-rid-base=SECONDARY_RID_BASE  |
+| ipa-replica-install | --netbios-name=NETBIOS_NAME --rid-base=RID_BASE --secondary-rid-base=SECONDARY_RID_BASE --add-sids |
+| ipa config-mod | --enable-sid --add-sids |
+
+The `--enable-sid` option turns the feature on, and `--add-sids` triggers
+the SIDgen task in order to immediately generate SID for existing users or
+groups.
+`--add-sids` can be used only when `--enable-sid` is used.
```
abbra, line 165: I think what you meant here is that `--add-sids` option depends on the presence of `--enable-sid` but `--enable-sid` can be used without it. Is that correct?

Perhaps, we can move detection logic to the framework and simply allow to specify either option. If SIDs weren't enabled or `--enable-sid` was not specified in the same request, `--add-sids` would simply fail. If `--enable-sid` was  specified, we enable SID generation. The key is to be able to detect whether SID generation is enabled.

While `ipa idrange-mod` is used to add RID bases, we still need a way to specify NetBIOS name for the domain. Should this be part of `ipa config-mod --enable-sid`? I think this part needs more investigation.

abbra (changes requested):
Thank you for the write up. It is a good one. I have few comments in several places and an open discussion around `ipa config-mod` part.

doc/designs/adtrust/sidconfig.md
```diff
@@ -0,0 +1,287 @@
...
+If the PAC data is not generated for a user, ensure that the user contains a
+SID in its `ipantsecurityidentifier` attribute. If the SID is missing, run
+the SIDgen task in order to generate SID for existing users and groups:
+- Find ipa base DN with: `grep basedn /etc/ipa/default.conf | cut -d= -f2-`
+- Copy `/usr/share/ipa/ipa-sidgen-task-run.ldif` to
+`/tmp/ipa-sidgen-task-run.ldif`
+- Edit the copy `/tmp/ipa-sidgen-task-run.ldif` and replace $SUFFIX with
+IPA base DN
+- As root, launch the task (replace IPA-TEST with the proper value):
+`ldapmodify -H ldapi://%2Frun%2Fslapd-IPA-TEST.socket -f /tmp/ipa-sidgen-task-run.ldif`
```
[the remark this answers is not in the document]
rcritten: This information will be very useful for ipa-healthcheck to verify that the configuration is correct.

PR #6043, opened 2021-10-07
ipatests: Update the subca used in TestIPACommand::test_cacert_manage

flo-renaud:
The above test is installing 2 Let's Encrypt certificates:
the root ISRG Root X1 and a subca. The subca expired Oct 6 and needs to
be replaced with a valid one, otherwise ipa-cacert-manage install
refuses to install it.

Fixes: https://pagure.io/freeipa/issue/9006


flo-renaud:
Removing ipa-4-6 label as the test does not exist on this branch

PR #6048, opened 2021-10-15
seccomp profile: Default to ENOSYS instead of EPERM

stanislavlevin:
This allows an application to detect whether the kernel supports
syscall or not. Previously, an error was unconditionally EPERM.
There are many issues about glibc failed with new syscalls in containerized
environments if their host run on old kernel or not updated seccomp profile.

More about motivation for ENOSYS over EPERM:
https://github.com/opencontainers/runc/issues/2151
https://github.com/opencontainers/runc/pull/2750

See about defaultErrnoRet introduction:
https://github.com/opencontainers/runtime-spec/pull/1087

Previously, FreeIPA profile was vendored from
https://github.com/containers/podman/blob/main/vendor/github.com/containers/common/pkg/seccomp/seccomp.json

Now it is merged directly from
https://github.com/containers/common/blob/main/pkg/seccomp/seccomp.json

Fixes: https://pagure.io/freeipa/issue/9008

### DIFF
--- a/ipatests/azure/Dockerfiles/seccomp.json
+++ b/ipatests/azure/Dockerfiles/seccomp.json
@@ -1,6 +1,7 @@
 {
 	"__defaultAction": "Change defaultAction to SCMP_ACT_LOG and then check Host's journal for SECCOMP",
-        "defaultAction": "SCMP_ACT_ERRNO",
+	"defaultAction": "SCMP_ACT_ERRNO",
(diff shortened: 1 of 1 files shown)


flo-renaud:
master:

* 8f1dda6404b9e75aa6b48c8f9694f2c65efa05e1 seccomp profile: Default to ENOSYS instead of EPERM

PR #6049, opened 2021-10-15
Make the schema cache TTL user-configurable

rcritten:
The API schema is not checked for changes until after a TTL
is expired. A one-hour TTL was hardcoded which makes development
tedious because the only way to force a schema update is to
remember to remove files between invocations.

This adds a new environment variable, schema_ttl, to configure
the TTL returned by the server to schema() calls. This can be
set low to ensure a frequent refresh during development.

If the client is in compat mode, that is if client is working
against a server that doesn't support the schema() command,
then use the client's schema_ttl instead so that the user still
has control.

https://pagure.io/freeipa/issue/8492

### DIFF
--- a/client/man/default.conf.5
+++ b/client/man/default.conf.5
@@ -172,6 +172,9 @@ Specifies the Kerberos realm.
 .B replication_wait_timeout <seconds>
 The time to wait for a new entry to be replicated during replica installation. The default value is 300 seconds.
 .TP
+.B schema_ttl <seconds>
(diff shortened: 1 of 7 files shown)


rcritten:
@stanislavlevin can I interpret that as an ack? :-)

stanislavlevin:
@rcritten, sure :)

flo-renaud:
master:

* b842b825ab4511dc6d56dea860ddf7347f14b4bd Make the schema cache TTL user-configurable

PR #6053, opened 2021-10-19
azure: Don't customize pip's builddir

stanislavlevin:
As of 21.3 pip:

> Remove the --build-dir option and aliases, one last time. (pypa/pip#10485)

https://pip.pypa.io/en/stable/news/#v21-3

Previous versions warn about deprecation.

The builddir is provided to pip via env variable PIP_BUILD in Tox task.
The purpose of changing of default builddir was noexec mount option for
/tmp in Travis (see 17d571c961). Since Travis is no longer used and
Azure lacks this issue the PIP_BUILD can be safely removed.

Note: pip 21.3 just ignores this env variable, which is more than can be
said for the command line option. It's better to clean it up, since the
behaviour may be changed in future.

Fixes: https://pagure.io/freeipa/issue/9011

### DIFF
--- a/.tox-install.sh
+++ b/.tox-install.sh
@@ -4,9 +4,8 @@ set -ex
 FLAVOR="$1"
 ENVPYTHON="$(realpath -s "$2")"
 ENVSITEPACKAGESDIR="$(realpath -s "$3")"
-ENVDIR="$4"
-# 4...end are package requirements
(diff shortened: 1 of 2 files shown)


stanislavlevin:
@rcritten, thank you! Indeed, this should be the revert of 17d571c961.

PR #6054, opened 2021-10-19
ipatests: TestMultipleExternalCA: Create tempfiles on remote host

stanislavlevin:
Previously, `test_master_install_ca1` and `test_master_install_ca2`
attempt to create tempdirs on local host and later write some
content into the returned paths on remote host. This fails if
a remote host is a local one.

The existent `create_temp_file` function has been extended to
support `suffix` option of `mktemp`.

Fixes: https://pagure.io/freeipa/issue/9013


mrizwan93:
Hello @stanislavlevin 
Thanks for the patch. Would you be able to run only changed tests using `ipatests/prci_definitions/temp_commit.yaml`, please ?

PR #6061, opened 2021-10-25
ipatests: use AD domain name from config instead of hardcoded value

wladich:
The test fails when test config contains AD domain value other than one hardcoded in the test code.


wladich:
Green result for test_trust: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/826f56c6-35a0-11ec-af55-fa163ed9fe48

Removing temp commit

wladich:
master:

* f1922fe5017a1e342af7500d875d9a238a48570c ipatests: use AD domain name from config instead of hardcoded value

PR #6077, opened 2021-11-10
ipatests: fix get_user_result method

flo-renaud:
Because the sidgen plugin is a postop plugin, it is not
always triggered before the result of an ADD is returned
and the objectclasses of the user may / may not contain
ipantuserattrs.
Fix the get_user_result method to work in all the cases.

Related: https://pagure.io/freeipa/issue/8995


abbra:
LGTM. ACK.

flo-renaud:
master:

* 2e7396b0e2869e1c5025903cf752e7d90e163a0a ipatests: fix get_user_result method

PR #6080, opened 2021-11-11
Harden PAC processing leftovers

abbra:
Two tests failed in nightlies after PAC hardening was merged. These patches address those changes.

### DIFF
--- a/daemons/ipa-kdb/ipa_kdb_mspac.c
+++ b/daemons/ipa-kdb/ipa_kdb_mspac.c
@@ -653,28 +653,15 @@ static krb5_error_code ipadb_fill_info3(struct ipadb_context *ipactx,
      * clear it after detecting the changes */
     info3->base.acct_flags = ACB_USE_AES_KEYS;
 
-    if ((is_host || is_service)) {
-        /* it is either host or service, so get the hostname first */
-        char *sep = strchr(info3->base.account_name.string, '/');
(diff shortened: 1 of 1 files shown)


abbra:
Tests are green:
test_smb: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/7894e986-42fc-11ec-916b-fa163ea73ebe/report.html
test_trust: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/5992cd64-42fc-11ec-8ad6-fa163e04db38/report.html

abbra:
Andreas actually checked this with me offline and got LGTM mark.

The were three problems:
 - the general logic is that we should prefer SID from the DB entry to our dynamic 'Domain Computers' / 'Domain Controllers' SID assignment
 - for S4U case, if the original client is from the trusted domain, we cannot compare their domain SID against ours domain SID. We have to go and check all trusted domains to see if there is a match and deny otherwise.
 - finally, for such client from trusted domain SID from S4U client cannot be checked against client_princ DB entry

rcritten:
master:

* 4062e7b963bbc23fc6ed908e94c4a0747712905a ipa-kdb: honor SID from the host or service entry
* 0c08faf3c1881661723a03b7ce7abce25147974c ipa-kdb: validate domain SID in incoming PAC for trusted domains for S4U

PR #6084, opened 2021-11-15
fix(webui): create correct PTR record when navigated from host page

pvoborni:
In scenario:
1. make sure that reverse zone doesn't have the desired PTR record
2. open host page of the host with matchnig the A record, e.g.: https://server.pvoborni.test/ipa/ui/#/e/host/details/test2.pvoborni.test
3. click on the "Host name" link, it will bring us to it's DNS record page. E.g., https://server.pvoborni.test/ipa/ui/#/e/dnsrecord/details/pvoborni.test&test2
! notice the missing '.' in the URL after zone name (pvoborni.test)
4. click on the A record , dialog will show up, saying "record not found"
5. click on the "create DNS record"

PTR record created by Web UI doesn't have trailing '.' (is not fully
qualified record) even if the DNS zone is.

This patch is fixing the link to the DNS Record page so that the
page then correctly gets the DNS Zone name and thus creates a correct
fully qualified PTR record.

https://bugzilla.redhat.com/show_bug.cgi?id=2009114
https://pagure.io/freeipa/issue/9036

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -1,3 +1,5 @@
+%define ipa_requires_gt()  %(LC_ALL="C" echo '%*' | xargs -r rpm -q --qf 'Requires: %%{name} >= %%{epoch}:%%{version}-%%{release}\\n' | sed -e 's/ (none):/ /' -e 's/ 0:/ /' | grep -v "is not")
+
 # ipatests enabled by default, can be disabled with --without ipatests
 %bcond_without ipatests
 # default to not use XML-RPC in Rawhide, can be turned around with --with ipa_join_xml
@@ -473,6 +475,8 @@ Requires: gssproxy >= 0.7.0-2
 Requires: sssd-dbus >= %{sssd_version}
 Requires: libpwquality
 Requires: cracklib-dicts
+# NDR libraries are internal in Samba and change with version without changing SONAME
+%ipa_requires_gt samba-client-libs
 
 Provides: %{alt_name}-server = %{version}
 Conflicts: %{alt_name}-server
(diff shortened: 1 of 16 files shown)


pvoborni:
We can also greatly simplify the patch by always adding '.' (and thus skipping the check) if we can be certain, that the dns zone is always fully qualified. But I'm not sure if that is the case.

flo-renaud:
@pvoborni thanks for the patch, the fix works for me. Can you remove the temp commit?
I assume a backport to ipa-4-9 will be needed, right? If that's your intent, please add the ipa-4-9 label.
Thanks

pvoborni:
Removed temp commit and rebased on top of master. But it seems that GitHub master is behind Pagure's master branch by 4 commits.  

WRT backport to 4-9 - up to you. From my PoV it can go there.

flo-renaud:
> Removed temp commit and rebased on top of master. But it seems that GitHub master is behind Pagure's master branch by 4 commits.
> 

This issue was reported at https://pagure.io/pagure/issue/5247 and is now fixed.

> WRT backport to 4-9 - up to you. From my PoV it can go there.

let’s include in 4-9 branch. Added the label.

flo-renaud:
master:

* 36b975804fad6b81dbfddd5ea0b5804542a2ed18 fix(webui): create correct PTR record when navigated from host page
* da4a7d29e9308e77c547c238c31c6c86fca78b3f webui tests: remove unnecessary code in add_record

flo-renaud:
@pvoborni a manual backport to ipa-4-9 is required (probably because of the pagure->github sync issue), can you do it? thanks

PR #6085, opened 2021-11-16
Fix use of comparison functions to avoid GCC bug 95189

abbra:
Due to a bug in GCC 9 and GCC 10 optimizing code, all C library
comparison functions should be used with explicit result comparison in
the code to avoid problems described in

http://r6.ca/blog/20200929T023701Z.html

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189

The code below is affected:

```
    if (strcmp(a, b) || !strcmp(c, d)) ...
```

while the code below is not affected:

```
    if (strcmp(a, b) != 0 || strcmp(c, d)) == 0
```

for all C library cmp functions and related:

 - strcmp(), strncmp()
 - strcasecmp(), strncasecmp()
 - stricmp(), strnicmp()
 - memcmp()

This PR idea is based on the pull request by 'Nicolas Williams <nico@twosigma.com>'
to Heimdal Kerberos: https://github.com/heimdal/heimdal/pull/855

### DIFF
--- a/client/ipa-getkeytab.c
+++ b/client/ipa-getkeytab.c
@@ -769,7 +769,7 @@ static char *ask_password(krb5_context krbctx, char *prompt1, char *prompt2,
                 NULL, NULL,
                 num_prompts, ap_prompts);
 
-    if (match && (strcmp(pw0, pw1))) {
+    if (match && (strcmp(pw0, pw1) != 0)) {
         fprintf(stderr, _("Passwords do not match!\n"));
         return NULL;
     }
--- a/client/ipa-join.c
+++ b/client/ipa-join.c
@@ -297,7 +297,7 @@ check_ipa_server(LDAP *ld, char **ldap_base, struct berval **vals)
 
         entry = ldap_first_entry(ld, res);
         infovals = ldap_get_values_len(ld, entry, info_attrs[0]);
-        if (!strcmp(infovals[0]->bv_val, "IPA V2.0"))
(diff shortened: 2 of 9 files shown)


abbra:
There are two remaining uses without immediate checks which we need to address:
```
$ git grep 'return.*cmp('
daemons/ipa-slapi-plugins/topology/topology_util.c:    return strcasecmp(l->from, r->from);
ipaplatform/redhat/tasks.py:        return self._rpmvercmp(self._bytes, other._bytes) == 0
ipaplatform/redhat/tasks.py:        return self._rpmvercmp(self._bytes, other._bytes) < 0
util/ipa_pwd.c:    return memcmp(*(void * const *)p2, *(void * const *)p1, GENERALIZED_TIME_LENGTH);
```

Python code can be ignored but both `return *cmp(..);` in C code could be affected.

abbra:
On the other hand, the pointers there do not come from constant expressions so gcc would not do analysis of the content of the pointers and probably should be OK.

flo-renaud:
@abbra thanks for the patch, LGTM. I didn't find additional occurrences, I think you covered all of them.

rcritten:
backport or main branch only?

flo-renaud:
master:

* 9043b8d534c8a7be10b7f2f17ac05c581ac19e65 Fix use of comparison functions to avoid GCC bug 95189

flo-renaud:
main branch only: RHEL 8 is using gcc 8.5 (not affected), RHEL 9 is using gcc 11 (not affected), fedora 34+ is using gcc 11.

PR #6090, opened 2021-11-18
ipatests: mark test_installation_TestInstallWithCA_DNS3 as xfail

flo-renaud:
The test failure is a known issue, happening on f33+. Mark as xfail
until 8700 is fixed.

Related: https://pagure.io/freeipa/issue/8700

### DIFF
--- a/ipatests/test_integration/test_installation.py
+++ b/ipatests/test_integration/test_installation.py
@@ -582,6 +582,9 @@ class TestInstallWithCA_DNS3(CALessBase):
     ticket 7239
     """
 
+    @pytest.mark.xfail(
+        osinfo.id == 'fedora' and osinfo.version_number >= (33,),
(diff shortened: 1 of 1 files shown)


fcami (changes requested):
Hi,
Shouldn't this be conditional on the OS version like this:
```
osinfo.id == 'fedora' and osinfo.version_number >= (33,)
```
otherwise this will start failing downstream for no good reason?
Thanks @flo-renaud

flo-renaud:
@fcami thanks for your comment, I added the condition as suggested.

fcami (approved):
ACK, thanks

flo-renaud:
master:

* 680d92b5488b5f71aaaafdd0f3675e4150c2e23f ipatests: mark test_installation_TestInstallWithCA_DNS3 as xfail

PR #6091, opened 2021-11-18
Various fixes for Debian

tjaalton:
This MR fixes some issues that I found when rebasing to current FreeIPA and Debian unstable.


ipaplatform/debian/paths.py
```diff
@@ -17,11 +17,11 @@
 
 class DebianPathNamespace(BasePathNamespace):
     BIN_HOSTNAMECTL = "/usr/bin/hostnamectl"
+    FIPS_MODE_SETUP = "/bin/fips-mode-setup"
+    SYSTEMD_TMPFILES = "/bin/systemd-tmpfiles"
```
tiran, line 21: It's safe to fix the path in ``ipaplatform/base/paths.py`` instead. Fedora-based systems have a unified ``/usr``.

tjaalton: yeah I thought about that.. will amend

ipaplatform/base/services.py
```diff
@@ -539,4 +539,4 @@ def base_service_class_factory(name, api=None):
 
 # System may support more time&date services. FreeIPA supports chrony only.
 # Other services will be disabled during IPA installation
-timedate_services = ['ntpd', 'chronyd']
+timedate_services = ['ntpd', 'chronyd', 'systemd-timesyncd']
```
tiran, line 542: I don't know if this is safe on Fedora and RHEL.

@abbra ?

tjaalton: how could there be both chronyd and systemd-timesyncd running?

rcritten: Right, chronyd is the only time service supported. By adding systemd-timesyncd to the list it will be explicitly disabled and shut down if it is running. So I don't see a problem with it, assuming you're ok with chrony.

ipaplatform/base/services.py
```diff
@@ -58,7 +58,7 @@
     'named', 'ods_enforcerd', 'ods_signerd', 'gssproxy',
     'nfs-utils', 'sssd', 'NetworkManager', 'ipa-custodia',
     'ipa-dnskeysyncd', 'ipa-otpd', 'ipa-ods-exporter',
-    'systemd-resolved',
+    'systemd-resolved', 'systemd-timesyncd',
```
rcritten, line 61: I don't think this is needed here since it isn't a service IPA needs to know, just one that may get shut down in favor of another time daemon.

tjaalton: fixed

rcritten:
Patch LGTM.

Adding re-run to run full test suite.

rcritten:
@tjaalton do you need this backported to the ipa-4-9 branch?

tjaalton:
That would be nice, yes.

rcritten:
master:

* 4738ab3dc74e01c3a838e62160ba02f97d6abc08 ipaplatform/debian: Fix HTTPD_ALIAS_DIR, and drop some obsolete paths.
* de97d83c8c7cbc297bcc9dedd3dd30f3ad79227a ipaplatform: Add support for recognizing systemd-timesyncd
* dfbae694a916b44c8a682664367a7099534dc23c ipaplatform/debian: Fix named keytab name
* e94afdd6f728780dcced9c5150fa4d07e5647989 ipaplatform/debian: Fix ntpd service name
* 52090d39e4fb33d03c9bb4825dc847111e9ca786 ipatests/test_ipaplatform: Skip test_ipa_version on Debian
* 9799b819fb049653a9352d5a5526d97accba2bef ipaplatform: Modify paths to fips-mode-setup and systemd-tmpfiles
* 9eecaddd86f0ef830c6783ef13c2c3761d95000b configure: Use HTTPD_GROUP in init/tmpfiles/ipa.conf.in

PR #6092, opened 2021-11-18
ipatests: Test cases for ipa-replica-conncheck command

mrizwan93:
Following test cases would be checked:
- when called with --principal (it should then prompt for a password)
- when called with --principal / --password
- when called without principal and password but with a kerberos TGT,
  kinit admin done before calling ipa-replica-conncheck
- when called without principal and password, and without any kerberos
  TGT (it should default to principal=admin and prompt for a password)

related: https://pagure.io/freeipa/issue/9047

### DIFF
--- a/ipatests/test_integration/test_replica_promotion.py
+++ b/ipatests/test_integration/test_replica_promotion.py
@@ -138,7 +138,6 @@ class TestReplicaPromotionLevel1(ReplicaPromotionBase):
         assert res.returncode == 1
         assert expected_err in res.stderr_text
 
-
     @replicas_cleanup
     def test_one_command_installation(self):
         """
@@ -150,11 +149,11 @@ class TestReplicaPromotionLevel1(ReplicaPromotionBase):
         Firewall(self.replicas[0]).enable_services(["freeipa-ldap",
                                                     "freeipa-ldaps"])
         self.replicas[0].run_command(['ipa-replica-install', '-w',
-                                     self.master.config.admin_password,
-                                     '-n', self.master.domain.name,
-                                     '-r', self.master.domain.realm,
-                                     '--server', self.master.hostname,
-                                     '-U'])
+                                      self.master.config.admin_password,
+                                      '-n', self.master.domain.name,
+                                      '-r', self.master.domain.realm,
+                                      '--server', self.master.hostname,
+                                      '-U'])
         # Ensure that pkinit is properly configured, test for 7566
         result = self.replicas[0].run_command(['ipa-pkinit-manage', 'status'])
         assert "PKINIT is enabled" in result.stdout_text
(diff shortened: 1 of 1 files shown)


ipatests/test_integration/test_replica_promotion.py
```diff
@@ -437,6 +437,56 @@ def test_renewal_master_with_csreplica_manage(self):
...
+        """Test cases for ipa-replica-conncheck command
+
+        Following test cases would be checked:
+        - when called with --principal (it should then prompt for a password)
+        - when called with --principal / --password
+        - when called without principal and password but with a kerberos TGT,
+          kinit admin done before calling ipa-replica-conncheck
+        - when called without principal and password, and without any kerberos
+          TGT (it should default to principal=admin and prompt for a password)
+        """
```
amore17, line 450: Is there any related issue for this ?

mrizwan93: No.

amore17: Please create upstream issue and link it into commit message.

mrizwan93: done

abbra:
Thanks for the patch. Could you please do assert checks for each `run_command` instance? Right now all of them run but have no checks. This means in case of a failure `run_command` would through an exception but an expectation is left to the caller. It would be great to encode that expectation within the test, e.g. that result is 0. See other tests for examples.

ipatests/test_integration/test_replica_promotion.py
```diff
@@ -437,6 +437,62 @@ def test_renewal_master_with_csreplica_manage(self):
...
+        tasks.kdestroy_all(self.replicas[0])
+        # when called with --principal (it should then prompt for a password)
+        result = self.replicas[0].run_command(
+            ['ipa-replica-conncheck', '--auto-master-check',
+             '--master', self.master.hostname,
+             '-r', self.replicas[0].domain.realm,
+             '-p', self.replicas[0].config.admin_name],
+            stdin_text=self.master.config.admin_password
+        )
+        assert result.returncode == 0
```
amore17, line 462: I would suggest you can also add assert "Connection from [*] to [*] is OK." in result.stdout

amore17 (changes requested):

amore17 (approved):

abbra:
LGTM. Please collect references to the successful tests (in a comment here) and remove the temp commit

mrizwan93:
temp commit logs : http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/39171710-5dab-11ec-8720-fa163e8572d3/

mrizwan93:
Thanks @abbra , temp commit removed.

mrizwan93:
> LGTM. Please collect references to the successful tests (in a comment here) and remove the temp commit
Adding ack per above comment.

flo-renaud:
master:

* 265f2618746134bb7804e569ffe18f491dc879be Test cases for ipa-replica-conncheck command
* df03dd9bfe855a65468da88a859e4fc52021053b PEP8 Fixes

PR #6086, opened 2021-11-17
pwpolicy: change lifetime error message

fcami:
ipa pwpolicy-mod --minlife $min --maxlife $max
accepts $max >= $min, yet the error message says:
"Maximum password life must be greater than minimum."

Change the error message so that it conveys the
actual logic.

Fixes: https://pagure.io/freeipa/issue/9038
Signed-off-by: François Cami <fcami@redhat.com>

PR #6098, opened 2021-11-22
ipatests: Check maxlife error message where minlife > maxlife specified

mrizwan93:
When minlife > maxlife specified on commandline, it says:
"ipa: ERROR: invalid 'maxlife': Maximum password life must be
greater than minimum."

But when minlife == maxlife specfied, It works.
This test check that error message says what exactly it does.

related: https://pagure.io/freeipa/issue/9038

### DIFF
--- a/ipatests/test_xmlrpc/test_pwpolicy_plugin.py
+++ b/ipatests/test_xmlrpc/test_pwpolicy_plugin.py
@@ -31,6 +31,59 @@ from ipatests.test_xmlrpc.xmlrpc_test import (XMLRPC_test, assert_attr_equal,
                                               Declarative)
 
 
+def pwpolicy_cmd(
+        cmd, minlife, maxlife, cospriority=None,
+        pwpolicy_group=None):
+    """Helper method to add or modify the password policy
+
+    :param cmd: Either pwpolicy_add or pwpolicy_mod
+    :param minlife: The minimum amount of time (in hours) that must pass
+                    between two password change operations.
+    :param maxlife: The maximum amount of time(in days) that must pass
+                    between two password change operations.
+    :param cospriority: priority
+    :param pwpolicy_group: password policy group
+    """
+    if cmd == 'pwpolicy_add':
+        if 0 <= minlife <= 24:
+            entry = api.Command[cmd](pwpolicy_group,
+                                     cospriority=cospriority,
+                                     krbminpwdlife=minlife,
+                                     krbmaxpwdlife=maxlife)['result']
+            assert_attr_equal(entry, 'krbminpwdlife', str(minlife))
+        elif minlife < 0:
+            with pytest.raises(errors.ValidationError) as e:
+                entry = api.Command[cmd](pwpolicy_group,
+                                         cospriority=cospriority,
+                                         krbminpwdlife=minlife,
+                                         krbmaxpwdlife=maxlife)['result']
+                assert "invalid 'minlife': must be at least 0" in str(e)
+        else:
+            with pytest.raises(errors.ValidationError) as e:
+                entry = api.Command[cmd](pwpolicy_group,
+                                         cospriority=cospriority,
+                                         krbminpwdlife=minlife,
+                                         krbmaxpwdlife=maxlife)['result']
+                assert ("Maximum password life must be equal to "
+                        "or greater than the minimum.") in str(e)
+    else:
+        if 0 <= minlife <= 24:
+            entry = api.Command[cmd](krbminpwdlife=minlife,
+                                     krbmaxpwdlife=maxlife)['result']
+            assert_attr_equal(entry, 'krbminpwdlife', str(minlife))
+        elif minlife < 0:
+            with pytest.raises(errors.ValidationError) as e:
+                entry = api.Command[cmd](krbminpwdlife=minlife,
+                                         krbmaxpwdlife=maxlife)['result']
+                assert "invalid 'minlife': must be at least 0" in str(e)
+        else:
+            with pytest.raises(errors.ValidationError) as e:
+                entry = api.Command[cmd](krbminpwdlife=minlife,
+                                         krbmaxpwdlife=maxlife)['result']
+                assert ("Maximum password life must be equal to "
+                        "or greater than the minimum.") in str(e)
+
+
 @pytest.mark.tier1
 class test_pwpolicy(XMLRPC_test):
     """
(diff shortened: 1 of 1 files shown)


mrizwan93:
depends on https://github.com/freeipa/freeipa/pull/6086

ipatests/test_integration/test_pwpolicy.py
```diff
@@ -261,3 +261,35 @@ def test_minlength_add(self):
         )
         assert result.returncode != 0
         assert 'minlength' in result.stderr_text
+
+    def test_maxlife(self):
+        """Check maxlife error message where minlife > maxlife specified
```
amore17, line 266: If its checking only error text then i would suggest you to add this test 
in test_xmlrpc/test_pwpolicy_plugin.py

ipatests/test_xmlrpc/test_pwpolicy_plugin.py
```diff
@@ -157,6 +157,36 @@ def test_9_pwpolicy_mod(self):
...
+                                            krbmaxpwdlife=1)['result']
+        assert_attr_equal(entry, 'krbminpwdlife', '20')
+
+        # when minlife(specified in hours) > maxlife(specified in days)
+        try:
+            entry = api.Command['pwpolicy_mod'](krbminpwdlife=25,
+                                                krbmaxpwdlife=1)['result']
+        except errors.ValidationError as e:
+            assert ("Maximum password life must be equal to "
+                    "or greater than the minimum.") in str(e)
```
stanislavlevin, line 188: Is it an expected mandatory exception?
https://docs.pytest.org/en/latest/how-to/assert.html#assertions-about-expected-exceptions

mrizwan93: Yes, see https://github.com/freeipa/freeipa/pull/6086/files#diff-fcbdf956172aafe4f6c1e64b54988af95c6bd35770c9a1e83e056e72e3114d21R492

stanislavlevin: let me rephrase myself.
if this is an expected exception then it should be raised *always*. Currently, the test assumes that the exception is *optional*. For example, `try except else` construction or Pytest's `pytest.raises` (the link above).

mrizwan93: Missed your point :)
Updated.

stanislavlevin: example: https://github.com/freeipa/freeipa/blob/637412acfd20eaa9e741db7a5d548d6daa14aebd/ipatests/test_xmlrpc/test_plugins_plugin.py#L56-L60

ipatests/test_integration/test_pwpolicy.py
rcritten: +1

mrizwan93:
@stanislavlevin @rcritten Could you look at this please?

stanislavlevin:
Hi, @mrizwan93.
- I think it's a good idea to extend your tests to check `validate_lifetime` of `ipaserver/plugins/pwpolicy.py`, not only the one case (`pwpolicy_mod`). `pwpolicy_add` is affected as well
- `validate_lifetime` for `pwpolicy_mod` has additional conditions. Please, cover them with tests.

ipatests/test_xmlrpc/test_pwpolicy_plugin.py
```diff
@@ -154,9 +205,53 @@ def test_9_pwpolicy_mod(self):
...
+        pwpolicy_cmd('pwpolicy_mod', 20)
+        pwpolicy_cmd('pwpolicy_add', 20, 6, test_group102, pwpolicyadd=True)
+
+        # when minlife(specified in hours) > maxlife(specified in days)
+        pwpolicy_cmd('pwpolicy_mod', 25)
+        pwpolicy_cmd('pwpolicy_add', 25, 7, test_group101, pwpolicyadd=True)
+
+        # when minlife is -1
+        pwpolicy_cmd('pwpolicy_mod', -1)
+        pwpolicy_cmd('pwpolicy_add', -1, test_group101, pwpolicyadd=True)
```
flo-renaud, line 253: The cospriority argument is missing: the method expects `cmd, minlife, cospriority=None, pwpolicy_group=None, pwpolicyadd=False`.

flo-renaud (commented):
Hi @mrizwan93 
please find an inline comment

ipatests/test_xmlrpc/test_pwpolicy_plugin.py
```diff
@@ -186,28 +282,32 @@ def test_b_pwpolicy_add(self):
     def test_c_pwpolicy_find(self):
         """Test that password policies are sorted and reported properly"""
         result = api.Command['pwpolicy_find']()['result']
-        assert len(result) == 4
+        assert len(result) == 6
```
flo-renaud, line 285: Instead of modifying `test_c_pwpolicy_find` and `test_c_pwpolicy_find_pkey_only` because there are additional policies, you can simply delete the pwpolicies created in `test_maxlife_pwpolicy` at the end of `test_maxlife_pwpolicy`.

ipatests/test_xmlrpc/test_pwpolicy_plugin.py
```diff
@@ -31,6 +31,49 @@
                                               Declarative)
 
 
+def pwpolicy_cmd(
+        cmd, minlife, cospriority=None,
+        pwpolicy_group=None, pwpolicyadd=False):
```
flo-renaud, line 36: The method arguments are redundant: there is no need for pwpolicyadd=True/False as the `cmd` arg already defines either add or mod. I would remove the `pwpolicyadd` arg and replace `if pwpolicyadd:` with `if cmd == "pwpolicy_add":`

ipatests/test_xmlrpc/test_pwpolicy_plugin.py
```diff
@@ -31,6 +31,49 @@
...
+    if pwpolicyadd:
```
flo-renaud, line 37: Add a method docstring: either create a new pwpolicy with max life = 1 day or modify the existing global pwpolicy.

ipatests/test_xmlrpc/test_pwpolicy_plugin.py
```diff
@@ -31,6 +31,56 @@
                                               Declarative)
 
 
+def pwpolicy_cmd(
+        cmd, minlife, cospriority=None,
+        pwpolicy_group=None):
+    """Helper method to add or modify the password policy
+
+    :param cmd: Either pwpolicy_add or pwpolicy_mod
+    :pram minlife: minimum life for the password to exist
```
flo-renaud, line 40: typo `pram`
minlife: The minimum amount of time (in hours) that must pass between two password change operations.

ipatests/test_xmlrpc/test_pwpolicy_plugin.py
```diff
@@ -31,6 +31,56 @@
                                               Declarative)
 
 
+def pwpolicy_cmd(
+        cmd, minlife, cospriority=None,
+        pwpolicy_group=None):
+    """Helper method to add or modify the password policy
+
```
flo-renaud, line 38: The add method harcodes the max life to 1 day.

flo-renaud (commented):
Hi @mrizwan93 
I only have minor nitpicks, please find them inline.

abbra:
master:

* 5a909cff8df4b37cdba3bc5bfabf97cc11379094 ipatests: Check maxlife error message where minlife > maxlife specified

PR #6101, opened 2021-11-22
Don't limit role-find by hostname when searching for last KRA

rcritten:
The "is this the last KRA" test did a role-find including the
current server. This skewed the result if the server to be
removed has a KRA installed, it would always return "not allowed"
because len(roles) == 1 and the name matched, regardless of
whether other servers also provided a KRA.

https://pagure.io/freeipa/issue/8397

### DIFF
--- a/ipaserver/plugins/server.py
+++ b/ipaserver/plugins/server.py
@@ -509,7 +509,6 @@ class server_del(LDAPDelete):
         if self.api.Command.ca_is_enabled()['result']:
             try:
                 roles = self.api.Command.server_role_find(
-                    server_server=hostname,
                     role_servrole='KRA server',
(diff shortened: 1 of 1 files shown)


flo-renaud:
@rcritten thanks for the PR, works for me:

- [x] server and replica installed with kra, remove replica with ipa server-del: ok
- [x] server with kra, replica without kra, remove replica: ok
- [x] server with kra, replica without kra: remove server: refused
- [x] server with kra, replica without kra: remove server with --ignore-last-of-role: ok

rcritten:
master:

* 3bcbc869f6e1cce441294b98d4fa3688f5917042 Don't limit role-find by hostname when searching for last KRA

PR #6112, opened 2021-11-26
ipatests : local ca is not generated under fips

ssidhaye:
Certmonger uses default OpenSSL encryption algorithms
to generate the PKCS12 object used for the local CA.
This uses operations that are disallowed under fips,
and so the local ca pkcs12 creds file is not generated.

Bugzilla Link:  https://bugzilla.redhat.com/show_bug.cgi?id=1950132

### DIFF
--- a/.freeipa-pr-ci.yaml
+++ b/.freeipa-pr-ci.yaml
@@ -1,1 +1,1 @@
-ipatests/prci_definitions/gating.yaml
+ipatests/prci_definitions/temp_commit.yaml
--- a/ipatests/prci_definitions/temp_commit.yaml
+++ b/ipatests/prci_definitions/temp_commit.yaml
@@ -68,7 +68,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-latest/build_url}'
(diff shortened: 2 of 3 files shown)


fcami (changes requested):
Please change the commit message with a ticket number, not a bz link.
Please explain what the expected behavior is in the commit message and what the previous, buggy behavior was.

ssidhaye:
I have updated the commit message. As for the ticket link all I could find was https://pagure.io/certmonger/pull-request/198 , I could not find a ticket. @rcritten is there a pagure ticket link for this fix ?

rcritten:
I think I can address automation directly in certmonger rather than in IPA. Let me see what is involved.

rcritten:
Here is my proposal to fix this in the certmonger unit tests, https://pagure.io/certmonger/pull-request/232

ssidhaye:
I don't mind including it in certmonger repo itself. Is it tested as part of freeipa CI ?

rcritten:
No but the problem doesn't apply to IPA which doesn't use the local CA helper.

PR #6113, opened 2021-11-27
PAC fixes for Windows Server November 2021 security release

abbra:
### ipa-kdb: issue PAC_REQUESTER_SID only for TGTs
    
MS-KILE 3.3.5.6.4.8 in revision after Windows Server November 2021  security fixes added the following requirement:

- PAC_REQUESTER_SID is only added in TGT case (including referrals and  tickets to RODCs)

### ipa-kdb: fix requester SID check according to MS-KILE and MS-SFU updates
    
New versions of MS-KILE and MS-SFU after Windows Server November 2021 security updates add PAC_REQUESTER_SID buffer check behavior:

- PAC_REQUESTER_SID should only be added for TGT requests
- if PAC_REQUESTER_SID is present, KDC must verify that the cname on the ticket resolves to the account with the same SID as the PAC_REQUESTER_SID. If it doesn't KDC must respond with KDC_ERR_TKT_REVOKED
    
Change requester SID check to skip exact check for non-local PAC_REQUESTER_SID but harden to ensure it comes from the trusted domains we know about.

If requester SID is the same as in PAC, we already do cname vs PAC SID verification.

With these changes FreeIPA works against Windows Server 2019 with November 2021 security fixes in cross-realm S4U2Self operations.
    
Fixes: https://pagure.io/freeipa/issue/9031


daemons/ipa-kdb/ipa_kdb_mspac.c
```diff
@@ -2005,22 +2005,43 @@ static krb5_error_code ipadb_check_logon_info(krb5_context context,
...
+            if (!is_from_trusted_domain) {
+                /* memctx is freed by the caller */
+                char *pac_sid = dom_sid_string(tmpctx, &client_sid);
+                char *req_sid = dom_sid_string(tmpctx, requester_sid);
+                krb5_klog_syslog(LOG_ERR, "PAC issue: PAC has a SID "
+                                        "different from what PAC requester claims. "
+                                        "PAC [%s] vs PAC requester [%s]",
+                                        pac_sid ? pac_sid : "<failed to display>",
+                                        req_sid ? req_sid : "<failed to display>");
+                kerr = KRB5KDC_ERR_POLICY;
```
rcritten, line 2042: Shouldn't this also return KRB5KDC_ERR_TGT_REVOKED?

abbra: May be. Perhaps we should do the same in the other place where we do this check as well

abbra: Updated. The other place already used `KRB5KDC_ERR_TGT_REVOKED`. Basically, this means the original ticket was not valid so we indeed want to force clients to drop it.

abbra:
master:

* bda59110f172231a5b123d1259ade943c9df5428 ipa-kdb: issue PAC_REQUESTER_SID only for TGTs
* 637412acfd20eaa9e741db7a5d548d6daa14aebd ipa-kdb: fix requester SID check according to MS-KILE and MS-SFU updates

PR #6120, opened 2021-12-07
ipatests: Adding --no-dnssec-validation option for ipa-server and replica install

menonsudhir:
healthcheck related tests are failing because of the below issue 
"client @0x7f8ee47c4d48 : servfail cache hit (CD=0)" and as a result healthcheck related packages 
are not downloaded on test system.
Hence adding the --no-dnssec-validation option to install_master and install_replica function

### DIFF
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -241,9 +241,16 @@ class TestIpaHealthCheck(IntegrationTest):
     def install(cls, mh):
         if not cls.master.transport.file_exists(SOS_CMD):
             tasks.install_packages(cls.master, SOS_PKG)
-        tasks.install_master(cls.master, setup_dns=True)
+        tasks.install_master(
+            cls.master, setup_dns=True, extra_args=['--no-dnssec-validation']
+        )
         tasks.install_client(cls.master, cls.clients[0])
-        tasks.install_replica(cls.master, cls.replicas[0], setup_dns=True)
+        tasks.install_replica(
+            cls.master,
+            cls.replicas[0],
+            setup_dns=True,
+            extra_args=['--no-dnssec-validation']
+        )
 
     def test_ipa_healthcheck_install_on_master(self):
         """
@@ -1451,7 +1458,9 @@ class TestIpaHealthCheckWithoutDNS(IntegrationTest):
     def install(cls, mh):
         tasks.uninstall_replica(cls.master, cls.replicas[0])
         tasks.uninstall_master(cls.master)
-        tasks.install_master(cls.master, setup_dns=False)
+        tasks.install_master(
+            cls.master, setup_dns=False, extra_args=['--no-dnssec-validation']
+        )
 
     def test_ipa_dns_systemrecords_check(self):
         """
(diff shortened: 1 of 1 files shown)


ipatests/prci_definitions/temp_commit.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
       class: RunPytest
       args:
         build_url: '{fedora-latest/build_url}'
-        test_suite: test_integration/test_REPLACEME.py
+        test_suite: test_integration/test_ipahealthcheck.py::TestIpaHealthCheck
         template: *ci-master-latest
         timeout: 3600
```
flo-renaud, line 73: This test usually needs more than 1h to execute, see for instance https://pagure.io/freeipa/blob/master/f/ipatests/prci_definitions/nightly_latest.yaml#_1369: `timeout: 5400`

menonsudhir: @flo-renaud added the changes to yaml file, rerunning the job

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -354,7 +354,8 @@ def install_master(host, setup_dns=True, setup_kra=False, setup_adtrust=False,
         args.extend([
             '--setup-dns',
             '--forwarder', host.config.dns_forwarder,
-            '--auto-reverse'
+            '--auto-reverse',
+            '--no-dnssec-validation',
```
flo-renaud, line 358: With this change, **all** the integration tests with DNS will be setup with `--no-dnssec-validation`. This breaks `ipatests.test_integration.test_upgrade.TestUpgrade.test_current_named_conf` for instance (see the azure test output).

If the intent is to change only the healthcheck tests, you can modify the `install` method of the class `TestIpaHealthCheck`, so that it calls `tasks.install_master` with `extra_args=['--no-dnssec-validation']`. This way only this test will be impacted.

flo-renaud (commented):
Hi @menonsudhir 
Is there a ticket for this issue?

ipatests/pytest_ipa/integration/tasks.py
menonsudhir: @flo-renaud i have reverted the changes done in tasks.py, will be making the changes in healthcheck tests

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -354,7 +354,7 @@ def install_master(host, setup_dns=True, setup_kra=False, setup_adtrust=False,
         args.extend([
             '--setup-dns',
             '--forwarder', host.config.dns_forwarder,
-            '--auto-reverse'
+            '--auto-reverse',
```
amore17, line 357: Remove extra `,`

amore17 (changes requested):

abbra:
@menonsudhir any progress with this PR?

menonsudhir:
@abbra right now i am not able to get a good run of above code fix in LTE which is preventing it from getting done, as the results are not getting generated to confirm that all the tests are passing.

ipatests/pytest_ipa/integration/tasks.py
```diff
@@ -355,7 +355,7 @@ def install_master(host, setup_dns=True, setup_kra=False, setup_adtrust=False,
...
```
miskopo, line 358: Please remove the extra `,`.

rcritten:
@menonsudhir  can you add ticket https://pagure.io/freeipa/issue/9151 to your commit message?

Otherwise this change LGTM.

rcritten:
master:

* 92d8077304ad0847c42eca3d11c79d9e22a46335 ipatests: Adding --no-dnssec-validation option for healthcheck

PR #6125, opened 2021-12-14
ipatests: Fix expected object classes

flo-renaud:
Because the sidgen plugin is a postop plugin, it is not
always triggered before the result of an ADD is returned
and the objectclasses of the user may / may not contain
ipantuserattrs.
Fix the expected object classes.

Related: https://pagure.io/freeipa/issue/9062

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -155,6 +155,7 @@ install/custodia/ipa-custodia-dmldap
 install/custodia/ipa-custodia-pki-tomcat
 install/custodia/ipa-custodia-pki-tomcat-wrapped
 install/custodia/ipa-custodia-ra-agent
+install/oddjob/org.freeipa.server.config-enable-sid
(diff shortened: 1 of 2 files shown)


rcritten:
Is there a reason to not convert the sidgen plugin into a transaction plugin so we always get the expected result?

stanislavlevin:
@flo-renaud, any chance that it may be fixed in a universal manner instead of fix of spotted issues? There are still cases that are not covered with this PR.

abbra:
LGTM

flo-renaud:
master:

* 54d32fcd766896dc0036c9a2825a7156ca8d2d14 ipatests: Fix expected object classes
* 21091c2bc779d65d9049e01cd6ac6a7f2d2ef60d gitignore: add install/oddjob/org.freeipa.server.config-enable-sid

PR #6129, opened 2022-01-03
automember default group: remove --desc parameter

flo-renaud:
The automember-default-group commands inherit from
the automember commands but should not provide the
--desc parameter.
Remove 'description' from the list of parameters.

Fixes: https://pagure.io/freeipa/issue/9068


t-woerner (approved):
LGTM

rcritten:
Adding ack. Normally we don't remove options from commands for backwards compatibility but in this case description is not allowed in the entry so we have no chance of breaking anything.

rcritten:
master:

* 189d090d2990c781bad88b251f998617d48382b8 automember default group: remove --desc parameter

PR #6140, opened 2022-01-13
ipa-kdb: refactor KDB driver to prepare for KDB version 9

abbra:
MIT Kerberos 1.20 changes DAL interface around PAC record issuance:
sign_authdata callback is removed and replaced with issue_pac one.
The signatures are different and logic changed as well.

Prepare for KDB version 9 by moving PAC implementation into separate
source files. ipa_kdb_mspac.c is left with most of the common code.

FreeIPA supports sign_authdata callback since KDB version 6, move current
implementation to ipa_kdb_mspac_v6.c.

KDB version 8 actually changed sign_authdata interface and we accounted
to that in ipa_kdb.c with a stub that re-uses v6 version. Keep it as it
is right now.

Finally, add KDB version 9 stub files. Compiling against MIT Kerberos
1.20 does not work yet, thus explicit #error message in ipa_kdb.c. This
will be worked on later.

Related: https://pagure.io/freeipa/issue/9083

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -293,6 +293,9 @@ AM_COND_IF([BUILD_IPA_CERTAUTH_PLUGIN], [
 AM_CONDITIONAL([BUILD_IPA_KDCPOLICY_PLUGIN],
                [test x$have_kdcpolicy_plugin = xyes])
 
+AM_CONDITIONAL([BUILD_IPA_ISSUE_PAC],
+               [test x$have_kdb_issue_pac = xyes])
+
 dnl ---------------------------------------------------------------------------
 dnl - Check for program paths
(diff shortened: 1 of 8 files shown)


rcritten:
This looks ok to me. I had to read the commit message to understand the v6/v9 file naming. At first I thought it was PAC versioning and not DAL versioning. I think it would be helpful to mention this in the top comment block of each file. Otherwise LGTM.

abbra:
I can do that in next update. PAC has no versioning in itself.

rcritten:
master:

* 2dc752018cc452b57bd6018504166131cd33c6c5 ipa-kdb: refactor KDB driver to prepare for KDB version 9

PR #6141, opened 2022-01-13
ipatests: fix expected automount config in nsswitch.conf 

flo-renaud:
The test TestIpaClientAutomountFileRestore expects a
specific order for the automount sources to query
in /etc/nsswitch.conf.

With authselect update 1.3.0, the databases are sorted in
order of likelihood and the following line in seen:
automount: files sss
instead of
automount: sss files

Since the test doesn't care about the order but rather about
the list of sources, ignore the order.

Fixes: https://pagure.io/freeipa/issue/9067

### DIFF
--- a/ipatests/test_integration/test_nfs.py
+++ b/ipatests/test_integration/test_nfs.py
@@ -323,7 +323,9 @@ class TestIpaClientAutomountFileRestore(IntegrationTest):
         if no_sssd:
             assert after_ipa_client_automount == ['files', 'ldap']
         else:
-            assert after_ipa_client_automount == ['sss', 'files']
+            # The default order depends on the authselect version
+            # but we only care about the list of sources, not their order
+            assert sorted(after_ipa_client_automount) == ['files', 'sss']
 
         cmd = self.clients[0].run_command(grep_automount_command)
         assert cmd.stdout_text.split() == after_ipa_client_automount


rcritten:
So your code looks fine but I'm not sure I understand your point on ordering. Are they saying it is more likely to have a file-based automount configuration? Is it because IPA always answers with an automount map so there no fallback to files?

flo-renaud:
> So your code looks fine but I'm not sure I understand your point on ordering. Are they saying it is more likely to have a file-based automount configuration? Is it because IPA always answers with an automount map so there no fallback to files?

With authselect < 1.3.0 and sssd profile, the content of authselect is:
```
automount: sss files
```
but with authselect >= 1.3.0 (available in rawhide):
```
automount: files sss
```
This change happened with authselect commit [3e3a473](https://github.com/authselect/authselect/commit/3e3a473c66c24b621838c1285f1f808149d3967b) and the justification was that having files first would 1. accelerate lookups and 2. be consistent with the other databases (services, netgroup, sudoers).

I could have fixed the test by checking the authselect version and expecting a different output dependending on it, but IMO we are mainly interested in the list of sources, not in the specific order they are queried.

rcritten:
Ok that makes sense, ack.

rcritten:
You can remove the temp commit now.

flo-renaud:
@rcritten thanks for the review, temp commit removed and tests are green.

flo-renaud:
master:

* d97a221f7e085f0c49c9d4faf94b7ddd14226e75 ipatests: fix expected automount config in nsswitch.conf

PR #6149, opened 2022-01-21
Implement LDAP bind grace period

rcritten:
Not ready for review, still hunting side-effects. Want to see what full CI thinks of it.


rcritten:
Marking as ready for review.

rcritten:
Rebased

abbra:
I wonder if it would be better to turn all of those helpers in the plugin's C code to be static and/or inline. They only used in the same source code and should not be exported. Things like `getPluginId()` would probably be common enough that exporting them beyond the compilation unit could cause issues with other plugins.

abbra:
Please rebase as API version did change due to other merged PRs.

abbra:
LGTM. If you are going to push this before RSN, then RSN PR needs rebase for the API version. And vice versa.

abbra:
master:

* f347c3f2302e468b7f92ec0146100b19570e382e Implement LDAP bind grace period 389-ds plugin
* 2d5e6935140c92b4faf0e2866c8f80f6a53c73d9 Remove the replicated attribute constants
* aefa5f22520d565f5accfc2257f48ff31be9b17b Exclude passwordgraceusertime from replication

PR #6154, opened 2022-01-27
Turn off interpolation to avoid issues with special characters

Caligatio:
This fixes https://pagure.io/freeipa/issue/9085

The Python `configparser.ConfigParser` class enables string interpolation by default which throws errors if `%` signs are used in any values that don't interpolate correctly. The majority of the code base uses `configparser.RawConfigParser` which has interpolation disabled by default so `ipalib/sysrestore.py` can match this functionality by setting `interpolation=None` on instantiation.

### DIFF
--- a/ipalib/sysrestore.py
+++ b/ipalib/sysrestore.py
@@ -82,7 +82,7 @@ class FileStore:
 
         self.files = {}
 
-        p = SafeConfigParser()
+        p = SafeConfigParser(interpolation=None)
         p.optionxform = str
         p.read(self._index)
 
@@ -103,7 +103,7 @@ class FileStore:
                 os.remove(self._index)
             return
 
-        p = SafeConfigParser()
+        p = SafeConfigParser(interpolation=None)
         p.optionxform = str
 
         p.add_section('files')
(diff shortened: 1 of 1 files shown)


rcritten:
Thanks for the patch. Can you update the commit message per https://www.freeipa.org/page/Contribute/Code#Commit_message_requirements

I don't think its a bad thing to disable interpolation but I don't know that this is going to make a NIS domain name with a % in it work everywhere. Have you tried netgroups with this change locally? Do they work as expected?

Caligatio:
Commit message updated!

I think I was unclear about the problem, this bug is preventing me from migrating **from** a NIS domain with a `%`. Rough steps to reproduce:

- Don't use FreeIPA
- Use a NIS domain with a percent sign (e.g. `exam%ple`)
- Decide to move to FreeIPA and pick a proper domain name (e.g. `example.com`)
- Run `ipa-client-install` with default options pointing to the FreeIPA server with `example.com` set
    - By default, `ipa-client-install` will configure the NIS domain
    - If an existing NIS domain is detected, (`exam%ple`), it will back-up that value using `ipalib/sysrestore.py:FileStore`
    - `FileStore` uses `SafeConfigParser` which then explodes while trying to  back-up the old/defunct `exam%ple` domain

I'm not trying to continue support for `%` in NIS domains, I'm trying to fix it so I can migrate off of it.

rcritten:
That makes sense. Kicking off full CI.

PR #6157, opened 2022-01-28
Remove override port from import_included_profiles

rcritten:
I can only guess to the original purpose of this override. I
believe it was because this is called in the installer prior
to Apache being set up. The expectation was that this would
only be called locally. It predates the RestClient class.

RestClient will attempt to find a usable service. In this
case, during a CA installation, the local server is not
considered available because it lacks an entry in
cn=masters.

So by overriding the port to 8443 a remote connection on that
port will be attempted and likely fail because we don't
require that the port be open.

Drop the override. The request will go remotely to an existing
CA via the pki proxy which will avoid a connection error.

The motivation for this is trying to install an EL 8.x replica
against an EL 7.9 server. 8.5+ includes the ACME service and
a new profile is needed which doesn't exist in 7. This was
failing because the RestClient determined that the local server
wasn't running a CA so tried the remote one (7.9) on the override
port 8443. Since this port isn't open: failure.

Chances are that adding the profile is still going to fail
because again, 7.9 lacks ACME capabilities, but it will fail in
a way that allows the installation to continue.

I suspect that all of the overrides can similarly be removed
but for the sake of "do no harm" I'm only removing this one for
now.

https://pagure.io/freeipa/issue/9100

### DIFF
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -1953,7 +1953,35 @@ def import_included_profiles():
         cn=['certprofiles'],
     )
 
-    api.Backend.ra_certprofile.override_port = 8443
+    # At this point Apache may or may not be running with a valid
+    # certificate. The local server is not yet recognized as a full
+    # CA yet so it isn't discoverable. So try to do some detection
+    # on what port to use, 443 (remote) or 8443 (local) for importing
+    # the profiles.
+    #
+    # api.Backend.ra_certprofile invokes the RestClient class
+    # which will discover and login to the CA REST API. We can
+    # use this information to detect where to import the profiles.
+    #
+    # If the login is successful (e.g. doesn't raise an exception)
+    # and it returns our hostname (it prefers the local host) then
+    # we override and talk locally.
+    #
+    # Otherwise a NetworkError means we can't connect on 443 (perhaps
+    # a firewall) or we get an HTTP error (valid TLS certificate on
+    # Apache but no CA, login fails with 404) so we override to the
+    # local server.
+    #
+    # When override port was always set to 8443 the RestClient could
+    # pick a remote server and since 8443 isn't in our firewall profile
+    # setting up a new server would fail.
+    try:
+        with api.Backend.ra_certprofile as profile_api:
+            if profile_api.ca_host == api.env.host:
+                api.Backend.ra_certprofile.override_port = 8443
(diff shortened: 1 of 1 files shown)


rcritten:
So apparently I only tested replica installs. On an initial install it must use the override port because there is no web server running yet.

rcritten:
So this failed with a 404:
```
GET https://master.ipa.test:443/ca/rest/account/login
```
Apache is up with a trusted cert and (on the "master") logs:
```
192.168.122.85 - - [28/Jan/2022:22:48:20 +0000] "GET /ca/rest/account/login HTTP/1.1" 404 196
```

I'll update the expect to also switch to 8443 on a RemoteRetrieveError error.

ipaserver/install/cainstance.py
```diff
@@ -1953,7 +1953,14 @@ def import_included_profiles():
         cn=['certprofiles'],
     )
 
-    api.Backend.ra_certprofile.override_port = 8443
+    try:
+        with api.Backend.ra_certprofile as profile_api:
+            print(profile_api.ca_host)
```
flo-renaud, line 1958: Do not forget to remove the above line before commit.

ipaserver/install/cainstance.py
```diff
@@ -1953,7 +1953,14 @@ def import_included_profiles():
...
+            if profile_api.ca_host == api.env.host:
```
flo-renaud, line 1959: It would be nice to add a comment explaining why the port is overriden.

flo-renaud (commented):
Hi @rcritten 
thanks for the PR, works for me. I tested multiple scenarios, all of them with firewalld blocking the 8443 port:
- [x] install 7.9 server with CA, install replica with patch with CA
- [x] install 7.9 server with CA, install replica with patch without CA, install CA clone on replica
- [x] install CA-less server with patch, install CA with ipa-ca-install
- [x] install server with patch with CA, install replica with patch with CA
- [x] install server with patch with CA, install replica with patch without CA, install CA clone on replica
- [x] call ipa-server-upgrade on server with CA instance

ipaserver/install/cainstance.py
rcritten: print() removed and giant comment block added.

rcritten:
master:

* 8c241869dda779d38fee4c23e44c89eb81a2a4d4 Don't always override the port in import_included_profiles

PR #6159, opened 2022-01-31
Remove ipa-join errors from behind the debug option

rcritten:
This brings it inline with the previous XML-RPC output which
only hid the request and response from the output and not
any errors returned.

https://pagure.io/freeipa/issue/9103

### DIFF
--- a/client/ipa-join.c
+++ b/client/ipa-join.c
@@ -743,8 +743,7 @@ jsonrpc_request(const char *ipaserver, const json_t *json, curl_buffer *response
 
     json_str = json_dumps(json, 0);
     if (!json_str) {
-        if (debug)
-            fprintf(stderr, _("json_dumps() failed\n"));
+        fprintf(stderr, _("json_dumps() failed\n"));
 
         rval = 17;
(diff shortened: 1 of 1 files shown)


Carbenium:
Would it make sense to review the use of `quiet` as well? It seems to be used rather inconsistently, e.g. https://github.com/freeipa/freeipa/blob/f3636c6afa75d9822a2fd1a535415b09f3fe1867/client/ipa-join.c#L428-L430

The usage text states `Only errors are displayed`, which is not quite true.

rcritten:
quiet is a fair point but this is PR is trying to regain parity with the XML-RPC implementation. With JSON failures are not displayed by default so it is difficult to troubleshoot why a join failed.

flo-renaud:

rcritten:
master:

* 3c1f4ba9956d920c0251525419d0dff3d8bf1bb3 Remove ipa-join errors from behind the debug option

rcritten:
I filed https://pagure.io/freeipa/issue/9105 to track the usage of --quiet.

PR #6165, opened 2022-02-04
Re-work the quiet option in ipa-join to not suppress errors

rcritten:
The quiet option was supposed to suppress unnecessary output
on successful invocations. Instead it was suppressing most
error messages as well.

Remove all the quiet checks when an error would be displayed.

Pass the quiet option to ipa-getkeytab so it honors the
request as well.

Also tidy up some of the debug output.

https://pagure.io/freeipa/issue/9105

### DIFF
--- a/client/ipa-join.c
+++ b/client/ipa-join.c
@@ -62,8 +62,7 @@ static int debug = 0;
 
 #define ASPRINTF(strp, fmt...) \
     if (asprintf(strp, fmt) == -1) { \
-        if (!quiet) \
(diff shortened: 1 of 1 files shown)


rcritten:
I'm leaving any branch tags off for now as this may warrant a discussion whether these changes could cause problems for existing scripts.

flo-renaud:

rcritten:
master:

* 61650c577f8847063de73218e1b8bc9f55e62fd8 Re-work the quiet option in ipa-join to not suppress errors

PR #6166, opened 2022-02-07
OpenLDAP 2.6+: use only -H option to specify LDAP url

abbra:
OpenLDAP 2.6+ finally deprecated -h and -p options in all its command line tools. They are not allowed anymore and cause ldap* tools to stop hard with 'unknown option' error.

Fix this by always using -H url option instead. Deriving default value for -H url from the configuration file still works, it is only -h and -p that were deprecated.

See also: https://bugs.openldap.org/show_bug.cgi?id=8618
    
Fixes: https://pagure.io/freeipa/issue/9106


flo-renaud:
@abbra thanks for the PR. I didn't spot additional uses of `-h hostname` option in our code, LGTM.

ACK for when PRCI completes.

abbra:
master:

* c93fa491f62ea17375d4f8fc9fe26895536400d2 OpenLDAP 2.6+: use only -H option to specify LDAP url
* 39a4d785274284b286107de3997178988439161c pylint: workaround incorrect pylint detection of a local function
* d64db126084861fc7214f5eaf32fb7396b88f302 translations: regenerate translations after changes in help message in sudorule

PR #6168, opened 2022-02-07
Enable the ccache sweep timer during installation

rcritten:
The timer was only being enabled during package installation
if IPA was configured. So effectively only on upgrade.

Add as a separate installation step after the ccache directory
is configured.

https://pagure.io/freeipa/issue/9107

### DIFF
--- a/ipaserver/install/httpinstance.py
+++ b/ipaserver/install/httpinstance.py
@@ -140,6 +140,8 @@ class HTTPInstance(service.Service):
         self.step("publish CA cert", self.__publish_ca_cert)
         self.step("clean up any existing httpd ccaches",
                   self.remove_httpd_ccaches)
+        self.step("enable ccache sweep",
+                  self.enable_ccache_sweep)
         self.step("configuring SELinux for httpd", self.configure_selinux_for_httpd)
         if not self.is_kdcproxy_configured():
             self.step("create KDC proxy config", self.create_kdcproxy_conf)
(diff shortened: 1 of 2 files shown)


abbra:
LGTM but could you please prefix the ticket with 'Fixed: ' ?

rcritten:
Since when are we using Fixes: as a prefix?

abbra:
I don't know, since ... forever? First commit message with `Fixes: <url>` was done in 2010:

```
commit 7aae58fd2d31e468a7fafd79d0bfb02595c66841 (tag: alpha_5-1-9-0-1)
Author: Simo Sorce <ssorce@redhat.com>
Date:   Wed Nov 10 13:34:16 2010 -0500

    uuid-plugin: Fix control access bug on replication
    
    Fixes: https://fedorahosted.org/freeipa/ticket/468

```

Before that there were `Fixes: <ticket number>` few times.

rcritten:
https://www.freeipa.org/page/Contribute/Code doesn't include it. I don't think I've ever done it.

abbra:
It is roughly 10% of all commits, 20% of commits with ticket references. Ok, may be I am too used to that. The script that generates release notes takes it into account.

rcritten:
Test is green, you can remove the temp commit.

mrizwan93:
temp commit result for reference : http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/48401984-89ad-11ec-8ca9-fa163ed9fe48

abbra:
Looks good to me, the test is concise and captures what is being changed in the installer. So LGTM and ACK after the tests passed.

PR #6173, opened 2022-02-08
Remove the --no-sssd option from ipa-client-automount

rcritten:
This makes automount configurable only using sssd and not LDAP.
The reason is that authselect 1.3 no longer supports
user-nsswitch.conf which is where we made direct changes to the
nss configuration on Fedora/RHEL.

The equivalent option was removed from ipa-client-install in
https://pagure.io/freeipa/issue/7671

Fixes: https://pagure.io/freeipa/issue/9084

### DIFF
--- a/client/man/ipa-client-automount.1
+++ b/client/man/ipa-client-automount.1
@@ -24,14 +24,12 @@ ipa\-client\-automount [\fIOPTION\fR]... <location>
 .SH "DESCRIPTION"
 Configures automount for IPA.
 
-The automount configuration consists of three files:
+The automount configuration consists of two files:
 .PP
 .IP  o
 /etc/nsswitch.conf
 .IP  o
 /etc/sysconfig/autofs
-.IP  o
-/etc/autofs_ldap_auth.conf
 
 .TP
 By default this will use DNS discovery to attempt to determine the IPA server(s) to use. If IPA servers are discovered then the automount client will be configured to use DNS discovery.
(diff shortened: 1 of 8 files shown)


rcritten:
I marked it as ipa-4-9 so we can fix a broken test but it probably should be master only. A rebase into an existing distro with the 4.9.x release will see this as loss of capability.

flo-renaud (commented):
Hi @rcritten 
thanks for the PR. I have a few minor inline comments.
We could also remove the method `tasks.enable_ldap_automount()` from ipaplatform (but we need to keep `disable_ldap_automount` for uninstallations). Not a strong requirement from me as I see the value in keeping it for the sake of enable/disable symmetry... so whatever your choice, I'm ok with it.

abbra:
@rcritten @flo-renaud any progress in this PR?

client/man/ipa-client-automount.1
```diff
@@ -42,9 +40,9 @@ The default automount location is named default. To specify a different one use
 .TP
 The IPA client must already be configured in order to configure automount. The IPA client is configured as part of a server installation.
 .TP
-There are two ways to configure automount. The default is to use sssd to manage the automount maps. Alternatively autofs can configured to bind to LDAP over GSSAPI and authenticate using the machine's host principal.
+SSSD is configured to manage the automount maps.
 .TP
-The nsswitch automount service is configured to use either sss or ldap and files depending on whether SSSD is configured or not.
+The nss automount service is configured to sss and files.
```
flo-renaud, line 45: Typo: is configured to *use* sss and files

client/man/ipa-client-automount.1
```diff
@@ -71,21 +66,11 @@ Restore the automount configuration files.
 
 .SH "FILES"
 .TP
-Files that will be always be configured:
+Files that will be configured:
 
 /etc/nsswitch.conf
```
flo-renaud, line 71: An empty line is needed between the 2 files (otherwise they will be displayed on the same line).

flo-renaud (commented):
Hi @rcritten 
2 nitpicks remaining...

client/man/ipa-client-automount.1
rcritten: Sigh, sorry about that. Both are resolved.

flo-renaud:
@rcritten Thanks for the update, ACK for when prci completes

flo-renaud:
master:

* a283735fbf1034a1ebffba598e47f9de9b394d6c Remove the --no-sssd option from ipa-client-automount

PR #6177, opened 2022-02-09
ipatests: Remove certmonger tracking before uninstall in cert tests

rcritten:
There is some contention between certmonger starting during the
uninstallation process in order to stop the tracking and activity
going on within certmonger helpers.

As near as I can tell certmonger is not running, then IPA is
stopped in order to uninstall, then certmonger is started to stop
the tracking. certmonger checks cert status on startup but since
IPA isn't running it can't get a host ticket. During this time any
request over DBus may time out, causing a test to fail when we're
just trying to clean up.

https://pagure.io/freeipa/issue/8506

### DIFF
--- a/ipatests/test_integration/test_ipa_cert_fix.py
+++ b/ipatests/test_integration/test_ipa_cert_fix.py
@@ -118,6 +118,13 @@ def expire_cert_critical():
     yield _expire_cert_critical
 
     host = hosts.pop('host')
+    # Prior to uninstall remove all the cert tracking to prevent
+    # errors from certmonger trying to check the status of certs
+    # that don't matter because we are uninstalling.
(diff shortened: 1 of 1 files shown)


flo-renaud:
@rcritten 
thanks for patch, LGTM. Since the issue happens randomly, it's difficult to certify that the patch will fix it completely but your analysis makes sense. Tested in this PR with a temp commit: #[1500](https://github.com/freeipa-pr-ci2/freeipa/pull/1500) and the test completed successfully ([Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/d99a7566-8a73-11ec-bb5b-fa163e616a6d/report.html)).

PR #6178, opened 2022-02-10
ipatests: remove additional check for failed units.

amore17:
Related : https://pagure.io/freeipa/issue/9108


stanislavlevin:
weird, it happens only for upstream CI.

PR #6184, opened 2022-02-11
Convert values using _SYNTAX_MAPPING with --delattr

rcritten:
When an entry is loaded the incoming values are converted
into python datatypes automatically based on the _SYNTAX_MAPPING
value in ipaldap.

When using delattr to remove a mapped value it will fail because
the datatypes do not match up. For example date types are
datetime.datetime structions and won't match a generalized time
string.

So try to map the value to delete using _SYNTAX_MAPPING before
trying to remove the value. Fall back to trying to remove the
raw value if the mapping fails.

This won't work for some mapping types, DNs for example. Providing
only the RDN value for a DN-type, manager for example, lacks the
context to know how to construct the DN (RDN and contaner).

Fixes: https://pagure.io/freeipa/issue/9004

### DIFF
--- a/ipaserver/plugins/baseldap.py
+++ b/ipaserver/plugins/baseldap.py
@@ -1135,7 +1135,11 @@ last, after all sets and adds."""),
 
                 for delval in deldict.get(attr, []):
                     try:
-                        entry_attrs[attr].remove(delval)
+                        try:
+                            val = ldap.decode(delval.encode('utf-8'), attr)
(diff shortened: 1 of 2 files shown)


abbra:
LGTM and ACK

abbra:
master:

* fe9be8c4a1aaca2a3bc3d8e5f0849a1716e19ca9 Convert values using _SYNTAX_MAPPING with --delattr

PR #6189, opened 2022-02-15
Set the mode on ipaupgrade.log during RPM %post snipppet

rcritten:
The IPA tools will create /var/log/ipaupgrade.log with mode
0600. If for some reason this file doesn't exist during
upgrade then it will be created by the RPM transaction with
mode 0644 (because of umask).

So always set the mode once the snippets are done. This
will ensure that a newly created log will have the expected
mode and also fix any previous incorrectly set mode.

Fixes: https://pagure.io/freeipa/issue/8899


stanislavlevin:
checked scenarios:
- existing installation (660) + upgrade => 600
- new installation + upgrade => 600

@rcritten, thank you for the fix, works as expected.

stanislavlevin:
re-run pr-ci and azure.

flo-renaud:
master:

* 56708d60703db794e02609f41f81da8e29313ddd Set the mode on ipaupgrade.log during RPM %post snipppet

PR #6192, opened 2022-02-22
ipa-restore: Mark a restored server as enabled

rcritten:
There is no use-case to keep a restored server in a hidden
state. It can be re-marked as hidden once the installation is
recovered from the restore. So mark all restored services as
enabled so they are visible to existing clients during the
remaining recovery.

Fixes: https://pagure.io/freeipa/issue/9095

### DIFF
--- a/ipaserver/install/ipa_restore.py
+++ b/ipaserver/install/ipa_restore.py
@@ -235,6 +235,15 @@ class Restore(admintool.AdminTool):
                 raise admintool.ScriptError(
                     "Directory Manager password required")
 
+    def enable_server(self):
+        """Make sure the current server is marked as enabled"""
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @rcritten 
thanks for the patch, works for me. I tested manually with @389ds/389-ds-base-nightly copr repo:
- install master and replica with DNS and CA
- hide the replica
- backup the replica
- stop master, uninstall replica, restore replica, check replica state
Without your patch, the replica is hidden after restore, with your patch the replica is visible.

The patch doesn't cause any issue if the restore is done on an already visible server.

rcritten:
master:

* 50241b36afe9019543dda0f8387ff25e6c672426 ipa-restore: Mark a restored server as enabled

PR #6197, opened 2022-02-25
KRB instance: make provision to work with crypto policy without SHA-1…

abbra:
… HMAC types

RHEL 9 system-wide crypto policies aim at eventual removal of SHA-1 use.

Due to bootstrapping process, force explicitly supported encryption
types in kdc.conf or we may end up with AES128-CTS and AES256-CTS only
in FIPS mode at bootstrap time which then fails to initialize kadmin
principals requiring use of AES256-SHA2 and AES128-SHA2.

Camellia ciphers must be filtered out in FIPS mode, we do that already
in the kerberos.ldif.

Fixes: https://pagure.io/freeipa/issue/9119

### DIFF
--- a/freeipa.spec.in
+++ b/freeipa.spec.in
@@ -695,6 +695,12 @@ Provides: %{name}-admintools = %{version}-%{release}
 Conflicts: crypto-policies < 20200629-1
 %endif
 
+%if 0%{?rhel} == 9
+# Conflict with crypto-policies < 20220223-1 to get upgraded AD-SUPPORT and
+# AD-SUPPORT-LEGACY policy modules
+Conflicts: crypto-policies < 20220223-1
+%endif
+
 %description client
 IPA is an integrated solution to provide centrally managed Identity (users,
 hosts, services), Authentication (SSO, 2FA), and Authorization
--- a/install/share/kdc.conf.template
+++ b/install/share/kdc.conf.template
@@ -6,7 +6,8 @@
 
 [realms]
  $REALM = {
-  master_key_type = aes256-cts
+  master_key_type = $MASTER_KEY_TYPE
+  supported_enctypes = $SUPPORTED_ENCTYPES
   max_life = 7d
   max_renewable_life = 14d
   acl_file = $KRB5KDC_KADM5_ACL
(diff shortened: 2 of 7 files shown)


abbra:
@f-trivino and I tested this patchset on the current internal RHEL 9 compose and in general it works for base installation and use. We are collecting failure cases to classify them and see what we can fix. So far, most of problems are within test infra setup but we also depend on very new versions of OpenSSL 3.0.0 build, crypto-policies, and krb5 fixes which aren't yet merged to CentOS 9 Stream.

So this PR will need to be updated when we'll have all proper package requirements available. Some of the tests need additional setup to work in FIPS mode (like AD-related tests).

jrisc:
I think it would be nice to rename `aes128-cts` and `aes256-cts` encryption types to `aes128-sha1`/`aes256-sha1`. They all are aliases to `aes*-cts-hmac-sha1-96`, but since `aes128-sha2` and `aes256-sha2` encryption types [are also using CTS](https://web.mit.edu/kerberos/krb5-1.19/doc/admin/conf_files/kdc_conf.html#encryption-types), it may be confusing.

I am not sure if it could cause conflicts with older versions though.

abbra:
Sure, I can do that. Since all of the shorter ones are aliases, they all will be accepted, so `aes256-cts` and `aes256-sha1` are the same.

abbra:
We did a test on RHEL 9 internal CI run and with some arrangement of openssl, crypto-policies, and krb5 versions we've got successful run for trust to Active Directory environment when FIPS mode is relaxed with AES encryption types using SHA-1 HMAC. @f-trivino is going to run a full test suite before giving a final ack here. Since this only affects RHEL 9, we'll add specific version pinning downstream. I will do a follow-up patch for upstream once openssl and crypto-policies updates will be in Fedora (they focus on RHEL 9 / CentOS 9 Stream changes now).

flo-renaud:
master:

* d38dd2680fb828da4b36e8eb4ce8dc8b056491ce KRB instance: make provision to work with crypto policy without SHA-1 HMAC types
* 2eee5931d714ca237290be7dc2fb7233ce747eca tests: ensure AD-SUPPORT subpolicy is active
* 985dffe147501e1b9cba3ab0be378115a23c3564 ipatests: extend AES keyset to SHA2-based ones
* 7d25eead994715b98f5dea5679c8fee0db4c4494 freeipa.spec: bump crypto-policies dependency for CentOS 9 Stream

PR #6202, opened 2022-03-07
ipatests: remove certmonger tracking before uninstall

flo-renaud:
test_ipahealthcheck_expiring is moving the date in the future
in order to check that certmonger properly warns about expiring
certificates, then uninstalls the master.

The uninstallation randomly fails with a DBus error communicating
with certmonger because of a contention between certmonger being
waken up by the call to stop tracking certs and the certmonger
helpers trying to renew the certs.

To avoid the contention, manually remove the tracking before
calling uninstall.

Fixes: https://pagure.io/freeipa/issue/9123

### DIFF
--- a/ipatests/test_integration/test_ipa_cert_fix.py
+++ b/ipatests/test_integration/test_ipa_cert_fix.py
@@ -122,9 +122,10 @@ def expire_cert_critical():
     # errors from certmonger trying to check the status of certs
     # that don't matter because we are uninstalling.
     host.run_command(['systemctl', 'stop', 'certmonger'])
-    host.run_command(
-        ['rm', '-f', paths.CERTMONGER_REQUESTS_DIR + '/*']
-    )
+    # Important: run_command with a str argument is able to
+    # perform shell expansion but run_command with a list of
+    # arguments is not
+    host.run_command('rm -fv ' + paths.CERTMONGER_REQUESTS_DIR + '*')
     tasks.uninstall_master(host)
     tasks.move_date(host, 'start', '-3Years-1day')
 
--- a/ipatests/test_integration/test_ipahealthcheck.py
+++ b/ipatests/test_integration/test_ipahealthcheck.py
@@ -1387,6 +1387,19 @@ class TestIpaHealthCheck(IntegrationTest):
             execute_nsscheck_cert_expiring(check)
 
         finally:
+            # Prior to uninstall remove all the cert tracking to prevent
+            # errors from certmonger trying to check the status of certs
+            # that don't matter because we are uninstalling.
+            self.master.run_command(['systemctl', 'stop', 'certmonger'])
+            # Important: run_command with a str argument is able to
+            # perform shell expansion but run_command with a list of
+            # arguments is not
+            self.master.run_command(
+                "rm -fv " + paths.CERTMONGER_REQUESTS_DIR + "*"
+            )
+            # Delete the renewal lock file to make sure the helpers don't block
+            self.master.run_command("rm -fv " + paths.IPA_RENEWAL_LOCK)
+            self.master.run_command(['systemctl', 'start', 'certmonger'])
             # Uninstall the master here so that the certs don't try
             # to renew after the CA is running again.
             tasks.uninstall_master(self.master)


rcritten:
I wonder if certmonger should be restarted after the removal. It has some startup cost that might be causing the DBus timeout. certmonger is DBus-activated so it will start when an operation is requested.

flo-renaud:
> I wonder if certmonger should be restarted after the removal. It has some startup cost that might be causing the DBus timeout. certmonger is DBus-activated so it will start when an operation is requested.

I tried your suggestion but it doesn't seem to fix the issue. There are probably some helpers running, maybe stuck with the global lock, and they block certmonger. I need to think more on a solution...

rcritten:
It looks like the uninstall is blazing fast so certmonger hasn't had time to start up the CAs yet. I don't see a way to tell whether this is finished in code, only by watching the journal which isn't practical here.

Maybe a sleep() after the startup to give it a chance. Perhaps start with 10 seconds?

rcritten:
So the cheat works but it doesn't feel very awesome.

For the longer term I'll look to see if there is something I can do in certmonger to represent that startup is "done".

I think this patch is fine as long as you don't think the sleep() is too icky. IMHO you can remove the temp commit.

flo-renaud:
> So the cheat works but it doesn't feel very awesome.
> 
> For the longer term I'll look to see if there is something I can do in certmonger to represent that startup is "done".
> 
> I think this patch is fine as long as you don't think the sleep() is too icky. IMHO you can remove the temp commit.

I also delete the renewal lock file, this may be the part that fixes everything and maybe we don’t even need the sleep. I’ll try tomorrow without the sleep and with the rm renewal.lock and see if it’s enough.

flo-renaud:
The tests are successful when the renewal lock file is deleted (no need to add a sleep) + the tracking files are removed. I think we have a proper solution now.
For future ref, links to successful tests:
- `latest/test_cert_fix`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/fb210380-a508-11ec-b407-fa163ea73ebe)
- `latest/test_ipahealthcheck-1`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/5dd31cd0-a508-11ec-9a72-fa163ed6a638)
- `latest/test_ipahealthcheck-2`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/68d62550-a508-11ec-a194-fa163e3cfd8f)
- `latest/test_ipahealthcheck-3`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ccbf7832-a508-11ec-93c6-fa163e68ae97)
- `latest/test_ipahealthcheck-4`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/dd38a51c-a508-11ec-b816-fa163e46d29b)
- `latest/test_ipahealthcheck-5`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ecdcb472-a508-11ec-842d-fa163ed5a785)
- `previous/test_ipahealthcheck-1`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/4605f5b4-a508-11ec-abca-fa163e8ff3eb)
- `previous/test_ipahealthcheck-2`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/55e54b60-a508-11ec-8435-fa163ef6d51c)
- `previous/test_ipahealthcheck-3`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/67934e0c-a508-11ec-aa77-fa163ea49ad6)
- `previous/test_ipahealthcheck-4`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/09fdc398-a509-11ec-b416-fa163ec98a23)
- `previous/test_ipahealthcheck-5`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/fb6315c2-a508-11ec-8ee8-fa163ee24cc1)

Removing the temp commit.

flo-renaud:
@rcritten is the fix ok with you? Since we are not tracking the certs anymore, it feels safe to remove the renewal lock file.

rcritten:
I don't like to argue with success but I don't understand why it works.

Yes, there will be no contention among the now non-existent certificates, but that should have been true before as well.

ipa-server-guard (the certmonger CA IPA) also uses the lock for the FETCH-ROOTS CA operation when certmonger starts. This lock file would be reconstituted when certmonger iterates its CA states at startup.

Which makes me wonder what the state of the lock is at removal time.

flo-renaud:
The test is stopping PKI server, then moves the date in the future. At this point, certmonger is still running (we are testing that getcert list properly warns about near expiration). This means that chances are high that certmonger has enough time to launch the CA helper for renewal, that takes the lock. But since PKI is down, I suspect that the helper remains running for a while and does not release the lock. Then certmonger is stopped, the tracking files are removed, certmonger is restarted.

In the log of a [failing test](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/863c4372-9ef5-11ec-a7bc-fa163ea73ebe/report.html) keeping the log file, we can see the [renewal.lock](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/863c4372-9ef5-11ec-a7bc-fa163ea73ebe/test_integration-test_ipahealthcheck.py-TestIpaHealthCheck-test_ipa_healthcheck_expiring/master.ipa.test/run/ipa/renewal.lock.gz) file:
```
[lock]
locked = 1
owner = ipa-server-guard[41468]
expire = 20240216010028011134
```
and the [journal ](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/863c4372-9ef5-11ec-a7bc-fa163ea73ebe/test_integration-test_ipahealthcheck.py-TestIpaHealthCheck-test_ipa_healthcheck_expiring/master.ipa.test/journal.gz) shows that the pid 41468 corresponds to the helper renewing the KDC cert:
```
Feb 16 00:00:05 master.ipa.test certmonger[41468]: 2024-02-16 00:00:05 [41468] Setting "CERTMONGER_CA_PROFILE" to "KDCs_PKINIT_Certs" for child.
```

rcritten:
Ok then yes, this is a fine workaround.

rcritten:
Can you add description this to the commit message? I think this analysis is worth preserving.

flo-renaud:
Sure thing, I updated the commit msg.

rcritten:
LGTM.  I added an ack since this looks fine but you'll need to note whether this is next or needs an ipa-4-9 backport.

flo-renaud:
master:

* 52ec9cc027d06621a0d1d5d9ad1084825c2a9b3c ipatests: remove certmonger tracking before uninstall
* e32bfd44ee8668237bb5191c794225040b7e07f5 ipatests: Fix a call to run_command with wildcard

PR #6204, opened 2022-03-09
allow overriding systemd-tmpfiles program

frasertweedale:
In some contexts, filesystem mounts may be owned by unmapped users
(e.g. `emptyDir` mounts in Kubernetes / OpenShift when using user
namespaces).  This causes `systemd-tmpfiles(8)` to fail, as a
consequence of systemd's path processing routines which reject this
scenario.  Therefore, in Fedora container context, if the
`IPA_TMPFILES_PROG` environment value is set, use the program
specified by its value instead of `/bin/systemd-tmpfiles`.

### DIFF
--- a/ipaplatform/fedora_container/paths.py
+++ b/ipaplatform/fedora_container/paths.py
@@ -29,5 +29,15 @@ class FedoraContainerPathNamespace(FedoraPathNamespace):
     HTTPD_IPA_WSGI_MODULES_CONF = None
     HTTPD_PASSWD_FILE_FMT = data(FedoraPathNamespace.HTTPD_PASSWD_FILE_FMT)
 
+    # In some contexts, filesystem mounts may be owned by unmapped users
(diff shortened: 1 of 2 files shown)


rcritten:
Can you update the commit with at least your signature?

If there is an upstream ticket, even in your project, related to this that would be good to add as well.

frasertweedale:
Added `signed-off-by` line to commit message.

There is no (public) ticket in our project for this.  I can make on in the FreeIPA issue tracker, if you want.  But it will only contain the same info as the commit message :)

abbra:
We use tickets to track backports and use in RHEL, so please add the ticket.

frasertweedale:
No worries.  Here is the ticket: https://pagure.io/freeipa/issue/9126 . I updated the commit message.

rcritten:
master:

* 4bdecd55e32249be817edd142d4939d9de61ae9c allow overriding systemd-tmpfiles program

PR #6209, opened 2022-03-14
Kerberos instance: default to AES256-SHA2 for master key encryption

abbra:
KDC configuration in /var/kerberos/krb5kdc/kdc.conf is generated from
the template in install/share/kdc.conf.template. Master key encryption
type specified there is used to bootstrap the master key in LDAP
database. Once it is done, actual deployment does not rely on the
master_key_type value anymore. The actual master key(s) get loaded from
LDAP database where they stored in a BER-encoded format, preserving all
parameters, including encryption type.

This means we can safely migrate to AES256-SHA2 as the default master
key encryption type for new installations. Replicas will get their
master key encryption type details from the server they were provisioned
from.

MIT Kerberos supports AES256-SHA2 since 1.15 (2015), meaning RHEL 7.4 is
the earliest supported version as it provides krb5 1.15.1. Current
supported RHEL 7 version is RHEL 7.9. Since RHEL 6 already cannot be
used as a replica to IPA 4.5+ due to a domain level 1 upgrade, this
change does not affect old releases.

Migration from the previously deployed master key encryption type is
described by MIT Kerberos upstream in
http://web.mit.edu/kerberos/krb5-latest/doc/admin/advanced/retiring-des.html#the-database-master-key

One would need to use '-x ipa-setup-override-restrictions' to allow
the `kdb5_util` utility to modify the data over IPA KDB driver.

Fixes: https://pagure.io/freeipa/issue/9119

### DIFF
--- a/ipaserver/install/krbinstance.py
+++ b/ipaserver/install/krbinstance.py
@@ -51,7 +51,7 @@ logger = logging.getLogger(__name__)
 
 PKINIT_ENABLED = 'pkinitEnabled'
 
-MASTER_KEY_TYPE = 'aes256-sha1'
+MASTER_KEY_TYPE = 'aes256-sha2'
 SUPPORTED_ENCTYPES = ('aes256-sha2:special', 'aes128-sha2:special',
                       'aes256-sha2:normal', 'aes128-sha2:normal',
                       'aes256-cts:special', 'aes128-cts:special',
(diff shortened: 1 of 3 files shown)


abbra:
I think we are done here for this round.

f-trivino (approved):
LGTM

abbra:
master:

* a6030f5f53e077a0443605fba23ef104c797f168 Kerberos instance: default to AES256-SHA2 for master key encryption
* 621af275c3564f1d5f1a97c46290b5d166fe8c8a test_otp: do not use paramiko unless it is really needed
* 517ae594e1ad80598a154f1dcacf7b303f420fc8 test_krbtpolicy: skip SPAKE-related tests in FIPS mode

PR #6230, opened 2022-03-30
ipatests: fix the topologysegment-reinitialize command

mrizwan93:
There is no guarantee for the topologysegement name, it could be
master-to-replica or replica-to-master. If it is master-to-replica
then --right should be used with the command else --left.

### DIFF
--- a/ipatests/test_integration/test_backup_and_restore.py
+++ b/ipatests/test_integration/test_backup_and_restore.py
@@ -35,6 +35,7 @@ from ipatests.test_integration.base import IntegrationTest
 from ipatests.pytest_ipa.integration import tasks
 from ipatests.test_integration.test_dnssec import wait_until_record_is_signed
 from ipatests.test_integration.test_simple_replication import check_replication
+from ipatests.test_integration.test_topology import find_segment
 from ipatests.util import assert_deepequal
 from ldap.dn import escape_dn_chars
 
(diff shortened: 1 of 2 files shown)


netoarmando:
I'm still not able to recover PR-CI runners. I accessed the dashboard and restart some runners, apparently it wasn't enough.

ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -862,13 +863,16 @@ def test_replica_install_after_restore(self):
...
+        topo_name = find_segment(master, replica1)
         for topo_suffix in 'domain', 'ca':
             arg = ['ipa',
                    'topologysegment-reinitialize',
                    topo_suffix,
-                   topo_name,
-                   '--left']
+                   topo_name]
+            if topo_name.split('-to-', maxsplit=1)[0] != master.hostname:
+                arg.append(['--left'])
```
flo-renaud, line 873: Append takes an item as parameter, not a list: 
`arg.append('--left')`

ipatests/test_integration/test_backup_and_restore.py
```diff
@@ -862,13 +863,16 @@ def test_replica_install_after_restore(self):
                            stdin_text=dirman_password + '\nyes')
 
         # re-initialize topology after restore.
-        topo_name = "{}-to-{}".format(master.hostname, replica1.hostname)
+        topo_name = find_segment(master, replica1)
```
flo-renaud, line 866: `find_segment` needs to be modified and accept the suffix as parameter (either `domain` or `ca`).
`topo_name` can be different for each suffix and needs to be set inside the `for topo_suffix` loop.

ipatests/test_integration/test_topology.py
```diff
@@ -10,16 +10,15 @@
...
 from ipatests.pytest_ipa.integration.env_config import get_global_config
-from ipalib.constants import DOMAIN_SUFFIX_NAME
 from ipatests.util import assert_deepequal
 
 config = get_global_config()
 reasoning = "Topology plugin disabled due to domain level 0"
 
 
-def find_segment(master, replica):
+def find_segment(master, replica, suffix):
```
flo-renaud, line 19: When an existing function signature is modified, you should provide a default value for the new parameter:
`def find_segment(master, replica, suffix=DOMAIN_SUFFIX_NAME):`

This way the existing code already using the function doesn't have to be modified (you can remove the change on line 138).

flo-renaud:
@mrizwan93 the change LGTM. I opened ticket #[9137](https://pagure.io/freeipa/issue/9137), can you update the commit message with "Fixes: https://pagure.io/freeipa/issue/9137" and remove the temp commit?

flo-renaud:
master:

* 4514e8090ab288a3bddde56c04d8014735fe95d1 ipatests: fix the topologysegment-reinitialize command
* 10046ffb33a0d8015fe8fcf06dd34666a2f0053c ipatests: extend find_segment with suffix param

PR #6231, opened 2022-04-01
healthcheck: add tests for setting cli options in config file

rcritten:
Fixes: https://pagure.io/freeipa/issue/9136


rcritten:
Ah right, I jumped the gun here. This release of freeipa-healthcheck is only in rawhide at the moment. I'll have to hold onto this until F36 is out of beta I think.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -581,6 +585,44 @@ def test_ipa_healthcheck_no_errors(self):
...
+
+        self.master.put_file_contents(
+            config_file,
+            '\n'.join([
+                '[default]',
+                'output_type=human'
+            ])
+        )
+        returncode, output = run_healthcheck(
+            self.master, failures_only=True, config=config_file
```
flo-renaud, line 603: `run_healthcheck` is expecting json output (because the caller doesn't set output_type="human") but receives human output because of the config file. This triggers a JSONDecodeError in the call to `json.loads(result.stdout_text)`.

`run_healthcheck` needs to be adapted and take into account the content of the config file.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -581,6 +585,44 @@ def test_ipa_healthcheck_no_errors(self):
...
+                '[default]',
+                'output_type=human',
+                'output_file=/tmp/hc.log',
+            ])
+        )
+        returncode, _unused = run_healthcheck(
+            self.master, config=config_file
+        )
+        logsize = len(self.master.get_file_contents('/tmp/hc.log'))
+        self.master.run_command(['rm', '-f', '/tmp/c.log'])
```
flo-renaud, line 621: The file name has a typo (`c.log` should be `hc.log`). Probably safer to define a constant for the file name.

ipatests/test_integration/test_ipahealthcheck.py
```diff
@@ -581,6 +585,44 @@ def test_ipa_healthcheck_no_errors(self):
         )
         assert returncode == 0
 
+    def test_ipa_healthcheck_no_errors_with_overrides(self):
+        """
+        Test overriding command-line options in a configuration file.
+        """
```
flo-renaud, line 591: You can add code to retrieve ipa-healthcheck version (we already have `tasks.get_healthcheck_version`) and call `pytest.skip` if the version is < 0.10.
```
        version = tasks.get_healthcheck_version(self.master)
        if parse_version(version) < parse_version("0.10"):
            pytest.skip("Skipping test for 0.10 healthcheck version")
```

flo-renaud (commented):
Hi @rcritten 
please find inline comments

ipatests/test_integration/test_ipahealthcheck.py
rcritten: This is testing a new feature in healthcheck where cli options can be set in configuration. But this release is only in rawhide currently which is why it failed.

ipatests/test_integration/test_ipahealthcheck.py
rcritten: ok

ipatests/test_integration/test_ipahealthcheck.py
rcritten: I will eventually.

ipatests/test_integration/test_ipahealthcheck.py
flo-renaud: I meant that even with the new freeipa-healthcheck version it would fail. `run_healthcheck` has the following code:
```
        if output_type == "json":
            data = json.loads(result.stdout_text)
```
`output_type` is `json` by default (and not set here in line 602), and the code will call json.loads. But since the config file is used, ipa-healthcheck outputs human format and json.loads will complain that it doesn't recognize the output as valid JSON.

ipatests/test_integration/test_ipahealthcheck.py
rcritten: Fixed.

rcritten:
I think I've caught up with the comments. Dropped WIP since the version of healthcheck is either available or guarded by a skip check.

flo-renaud:
@rcritten thanks for the updated patch, LGTM. You can remove the temp commit and I will ack when CI completes.

PR #6234, opened 2022-04-11
ipatests: Test certmonger rekey command fails for DSA key type

mrizwan93:
DSA is disabled in DEFAULT crypto policy in F30+, hence
the command with DSA key type will fail.

related: https://bugzilla.redhat.com/show_bug.cgi?id=2066439


rcritten:
getcert tends to not return 1 when things fail. I think if you remove this expectation the test will pass.

mrizwan93:
Closing in favor of https://github.com/freeipa/freeipa/pull/6236

PR #6241, opened 2022-04-29
EPN: document missing option msg_subject

flo-renaud:
In /etc/ipa/epn.conf it is possible to customize the
e-mail subject by setting msg_subject=<value> but this
setting is not documented in the man page.

Add the options in epn.conf man page and in the template.

Fixes: https://pagure.io/freeipa/issue/9145


rcritten:
The Azure failures seem to revolve around the CA not working (returning 500 error). I think for this PR we can ignore those failures and just merge as this has absolutely nothing to do with that.

rcritten:
master:

* 300f62f72bdc329b7ea5e1f54779d4078dc4f387 EPN: document missing option msg_subject

PR #6246, opened 2022-05-03
web ui: do not provide Remove button in subid page

abbra:
subid range management does not allow to delete ranges

If subid range was allocated, it cannot be removed because there might
be file objects associated with it on one of IPA clients.

In Web UI a button to remove the range should not be shown.

Remove corresponding test from the Web UI test for subid as the button
to remove the subid range is not present anymore.

Fixes: https://pagure.io/freeipa/issue/9150

### DIFF
--- a/install/ui/src/freeipa/subid.js
+++ b/install/ui/src/freeipa/subid.js
@@ -16,12 +16,30 @@ define([
 
 var exp = IPA.subid = {};
 
+exp.search_facet_control_buttons_pre_op = function(spec, context) {
+    spec.control_buttons = [
+        {
+            name: 'add',
+            label: '@i18n:buttons.add',
+            icon: 'fa-plus'
+        },
+        {
+            name: 'refresh',
+            label: '@i18n:buttons.refresh',
+            icon: 'fa-refresh'
+        }
+    ];
+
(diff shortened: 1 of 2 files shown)


install/ui/src/freeipa/subid.js
```diff
@@ -16,12 +16,30 @@ define([
...
+    return spec;
+};
+
 var make_spec = function() {
 return {
     name: 'subid',
     facets: [
         {
             $type: 'search',
+	    $pre_ops: [ exp.search_facet_control_buttons_pre_op ],
```
netoarmando, line 42: I believe this line's indentation is mixing tabs and spaces.

abbra: thanks, I fixed it.

netoarmando (approved):
LGTM. Everthing is working as expected. `spec.no_update` can't be used, so overriding `spec.control_buttons` is the way to achieve this. 

Screenshot from 2022-05-04 22-17-14

abbra:
master:

* 46dc1c76ef9996341b89167c6a0c1f63e44f4381 web ui: do not provide Remove button in subid page

